PR #3711, opened 2019-02-17
update libvirtd

Expire0:
libvirtd must be started before you run minikube. Failure to do so will result in the below error 
Failed to connect socket to '/var/run/libvirt/libvirt-sock'
I believe this should be noted in the documentation to assist other users.  
Added lines 40-43

### DIFF
--- a/docs/drivers.md
+++ b/docs/drivers.md
@@ -25,9 +25,9 @@ To install the KVM2 driver, first install and configure the prereqs:
 ```
 Enable,start, and verify the libvirtd service has started. 
 ```shell
-systemctl enable libvirtd.service
-systemctl start libvirtd.service
-systemctl status libvirtd.service
+sudo systemctl enable libvirtd.service
+sudo systemctl start libvirtd.service
+sudo systemctl status libvirtd.service
 ```
 
 
@@ -44,7 +44,7 @@ Now install the driver:
 ```shell
 curl -LO https://storage.googleapis.com/minikube/releases/latest/docker-machine-driver-kvm2 \
   && sudo install docker-machine-driver-kvm2 /usr/local/bin/
-
+```
 
 
 NOTE: Ubuntu users on a release older than 18.04, or anyone experiencing [#3206: Error creating new host: dial tcp: missing address.](https://github.com/kubernetes/minikube/issues/3206) you will need to build your own driver until [#3689](https://github.com/kubernetes/minikube/issues/3689) is resolved. Building this binary will require [Go v1.11](https://golang.org/dl/) or newer to be installed. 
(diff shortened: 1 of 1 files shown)


afbjorklund:
This probably only applies to the KVM drivers  ("kvm" and "kvm2") on Linux ? Not VirtualBox etc.

Expire0:
> Also, I believe the daemon is started automatically on Debian and derivates like Ubuntu.

I tested this on Ubuntu 18.04 and it does start automatically. The issue may be only related to Centos 7/Fedora 29 /Redhat 7

afbjorklund:
On other hand, it doesn't really hurt to start it again... But the instructions should be up with the rest of the hypervisor install (like adding yourself to the root-equivalent `libvirtd` group), rather than with the driver:


> To install the KVM2 driver, first install and configure the prereqs:

**HERE**: Make sure that the libvirt daemon actually started, after installation

> Then you will need to add yourself to libvirt group (older distributions may use libvirtd instead)

> Then to join the group with your current user session:

> Now install the driver:

Expire0:
I moved the instructions under the install as you suggested.  Good idea

docs/drivers.md
```diff
@@ -35,7 +44,8 @@ Now install the driver:
 ```shell
 curl -LO https://storage.googleapis.com/minikube/releases/latest/docker-machine-driver-kvm2 \
   && sudo install docker-machine-driver-kvm2 /usr/local/bin/
-```
+
```
afbjorklund: What happened here ? Seems like the "shell" end was lost ?

docs/drivers.md
```diff
@@ -22,6 +22,15 @@ To install the KVM2 driver, first install and configure the prereqs:
 * Ubuntu 16.x or older: `sudo apt install libvirt-bin libvirt-daemon-system qemu-kvm`
 * Fedora/CentOS/RHEL: `sudo yum install libvirt-daemon-kvm qemu-kvm`
 
+```
+Enable,start, and verify the libvirtd service has started. 
+```shell
+systemctl enable libvirtd.service
```
afbjorklund: Normally one wants to use "sudo" here, to avoid that annoying popup

docs/drivers.md
```diff
@@ -22,6 +22,15 @@ To install the KVM2 driver, first install and configure the prereqs:
...
+systemctl start libvirtd.service
+systemctl status libvirtd.service
```
afbjorklund: Status seems a bit paranoid ?

Expire0: yeah just a little.  I think it may be good to check before moving to the next step.

docs/drivers.md
Expire0: yeah it was. i cleaned it up .

docs/drivers.md
Expire0: good suggestion. I found that i needed to be root to run through the documentation . If that's the case then sudo is not needed.  I updated the doc to include sudo just in case

PR #3712, opened 2019-02-17
Add Netfilter `xt_socket` module to complete support for Transparent Proxying (TPROXY)

yskopets:
According to Kernel documentation, support for Transparent Proxying (TPROXY)
requires 2 Netfilter modules to be installed: `xt_TPROXY` and `xt_socket`.

So far, Minikube had only `xt_TPROXY`.

See https://www.kernel.org/doc/Documentation/networking/tproxy.txt


yskopets:
I've also opened an Issue for that - https://github.com/kubernetes/minikube/issues/3713

yskopets:
I've verified locally that changes in this PR actually fix the problem in #3713

PR #3714, opened 2019-02-18
Add --image-repository flag so that users can select an alternative repository mirror

### LINKED ISSUE #3166: Allow set `ImageRepository` for kubeadm init
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): FEATURE REQUEST

Please provide the following details:

**Environment**:

```
minikube version: v0.28.2

OS:
cat: /etc/os-release: No such file or directory

VM driver:
    "DriverName": "hyperkit",

ISO version
        "Boot2DockerURL": "file:///path/to/.minikube/cache/iso/minikube-v0.28.1.iso",
```

Can the kubeadm bootstrapper adds a configuration option support for [`ImageRepository`][kubeadmin_init], which enables setting different image base repository. This feature will help developers who are having difficulties to connect to k8s.gcr.io (namely, China), as long as they can set the repository prefix to self managed proxies.

[kubeadmin_init]: https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/#custom-images

laozc:
This PR is aiming to support a new command line argument --image-repository for some users which has difficulty accessing gcr.io.

Closes #3166


tstromberg:
Thank you so much for writing this. This PR could potentially make minikube work much better in China, as well as environments with limited internet connectivity.  I'll give this a closer look tomorrow, but at least initially this seems to be doing the right things.

Do you have a mirror or two to recommend that I could test this PR against?

One potential concern to eventually address is cache-poisoning, which will require minikube addressing images with hashes instead of versions -- but I won't let "perfect be the enemy of good". 

A potential eventual enhancement would be to have an automated fall-back mechanism between a set of known-good image repositories (similar to "apt" or other systems), so that minikube in users in China would not have to manually specify a --image-repository URL unless they wish to.

Thanks again!

tstromberg:
@minikube-bot OK to test

laozc:
You may start with registry.cn-hangzhou.aliyuncs.com/google_containers.
This is one of the well known mirrors for gcr.io maintained by Aliyun guys.
I verified with this mirror so it should work as expected (v1.13.3).

We may use --image-repository as an ultimate workaround
if none of mirrors provided is accessible in user's environment.
I agree that maintaining a mirror list as like the APT / pacman world also helps,
or even trying to guess a reliable mirror by IP address during the setup.
I can work on that later.

Let me know if you have any other concerns.

tstromberg:
How does this new `--image-repository` flag interact with this existing flag:

` --registry-mirror strings           Registry mirrors to pass to the Docker daemon`

I'll admit, I didn't know about this flag up until today.

laozc:
I believe --registry-mirror only impacts those images with no repository prefix - images that come from the Docker official registry.
It won'twork on images from private registries, which is the case like gcr.io/kube-proxy. For private images, docker will still go to the private registry and fetch them.

If you set up minikube with --registry-mirror, it could work for pods/deployments that use ubuntu/18.04, but not for gcr.io/ - you will need to wipe the "gcr.io/" out from the references to make it look like an image from the official registry. For the latter case, it could be achieved using --registry-mirror https://private_server --image-registry [private_server/]google_containers

afbjorklund:
For instance `crio` has a `--registry`flag, but that only works for images without prefix ("unqualified"):

```
   --registry value              registry to be prepended when pulling unqualified images, can be specified multiple times
```
The same goes for the global `/etc/containers/registries.conf`, which lists the default registries:

``` toml
[registries.search]
registries = ['docker.io']

[registries.insecure]
registries = []
```

And `/etc/crio/crio.conf` with CRI-O's:

```  toml
# insecure_registries is used to skip TLS verification when pulling images.
insecure_registries = [
]

# registries is used to specify a comma separated list of registries to be used
# when pulling an unqualified image (e.g. fedora:rawhide).
registries = [
        "docker.io"
]


```

cmd/minikube/cmd/start.go
```diff
@@ -123,6 +124,7 @@ func init() {
 	startCmd.Flags().String(serviceCIDR, pkgutil.DefaultServiceCIDR, "The CIDR to be used for service cluster IPs.")
 	startCmd.Flags().StringSliceVar(&insecureRegistry, "insecure-registry", nil, "Insecure Docker registries to pass to the Docker daemon.  The default service CIDR range will automatically be added.")
 	startCmd.Flags().StringSliceVar(&registryMirror, "registry-mirror", nil, "Registry mirrors to pass to the Docker daemon")
+	startCmd.Flags().String(imageRepository, "", "Image repository to pull down docker images.")
```
tstromberg: The help text isn't fully helpful here. How about: 

"Alternative image repository to pull docker images from."

tstromberg: If there are any publicly available examples of mirrors that work -- especially within China, could you add them as an example?

tstromberg:
/approve

Thank you for the clarification and excellent work here. I've added a few nitpicks, but it's good enough as is. This is really going to help minikube users around the world.

We're going to want to add an integration test for this, but until #3767 lands it's rather painful to add start arguments. We can do that in a follow-up PR.

cmd/minikube/cmd/start.go
laozc: Sure. I'll change the help text with mirror repository example for China users.

tstromberg:
@minikube-bot OK to test

tstromberg:
@minikube-bot OK to test

laozc:
Have some other work based on this PR.
Any chance to have it merged sooner?
Let me know if you have any concerns.

tstromberg:
@minikube-bot OK to test

tstromberg:
I think it may have been my attempt at merging to master, but it appears that "minikube dashboard" is broken with this PR at the moment.

tstromberg:
Do you mind remerging this PR? I'd like to also confirm that `minikube dashboard` still works.

laozc:
I'm checking the case you describe and looking into the PR.

tstromberg:
Thanks for making the changes!

/lgtm /approve

AdnanHodzic:
It says this PR has been merged with master, however if I run `minikube --image-repository` I'll get an error with `unknown flag: --image-repository` message.

Could somebody please clarify how can this flag be used exactly, as I'm in need of one such functionality.

laozc:
This flag only take affect when you execute minikube start.
You may check the output of minikube start --help.

If you don’t see the flag appearing in the help text it might suggest you’re using an old release.
The flag can be used as shown in the following examples:
minikube start --image-repository=registry.cn-hangzhou.aliyuncs.com/google_containers
or
minikube start --image-repository=auto

Regards,
Zhongcheng

AdnanHodzic:
@laozc got it, thank you for the clarification!

PR #3737, opened 2019-02-22
Add --download-only option to start command

### LINKED ISSUE #3736: Add start option, to download only
Make it possible to download files ahead-of-time, without actually install and loading them.

``` console
$ .minikube start --cache-images --download-only
😄  minikube v0.34.1 on linux (amd64)
🤹  Caching images in the background ...
💿  Downloading Minikube ISO ...
 184.30 MB / 184.30 MB [============================================] 100.00% 0s
💾  Downloading kubeadm v1.13.3
💾  Downloading kubelet v1.13.3
⌛  Waiting for image caching to complete ...
✔️  Download complete and in download only mode
```

This will save the files for later in the local minikube cache, before actually starting the host.

afbjorklund:
Make it possible to download files without loading or installing,
and without starting the host... Download: ISO, binaries, images.

Closes #3736


tstromberg:
I feel like the more natural place for this would be "minikube cache --startup-images", but I also recognize that this would be really difficult to implement with the way the code is currently organized.

cmd/minikube/cmd/cache.go
```diff
@@ -67,7 +67,7 @@ var deleteCacheCmd = &cobra.Command{
 }
 
 // LoadCachedImagesInConfigFile loads the images currently in the config file (minikube start)
-func LoadCachedImagesInConfigFile() error {
+func LoadCachedImagesInConfigFile(load bool) error {
```
tstromberg: Having a verb in both the function name and the variable name is a sign that this should be refactored. Given the function name and docstring, I don't understand what this would actually do if load is false.

Would it help to split this into two functions? Perhaps just a "Images" function that returns a list of images, and the caller can decide if they should be loaded or cached?

cmd/minikube/cmd/start.go
```diff
@@ -180,6 +182,22 @@ func runStart(cmd *cobra.Command, args []string) {
...
+			exit.WithError("Failed to cache ISO", err)
+		}
+		if err := bootstrapper.CacheBinaries(nil, config.KubernetesConfig); err != nil {
+			exit.WithError("Failed to cache binaries", err)
+		}
+		waitCacheImages(&cacheGroup)
+		if err := LoadCachedImagesInConfigFile(false); err != nil {
+			exit.WithError("Failed to cache images", err)
+		}
+		console.OutStyle("check", "Download complete and in download only mode")
```
tstromberg: I think just "Download complete!" would suffice as a console message.

pkg/minikube/bootstrapper/bootstrapper.go
```diff
@@ -58,3 +67,63 @@ func GetCachedImageList(version string, bootstrapper string) []string {
...
+	// If it exists, do no verification and continue
+	if err == nil {
+		glog.Infof("Not caching binary, using %s", url)
+		return targetFilepath, nil
+	}
+	if !os.IsNotExist(err) {
+		return "", errors.Wrapf(err, "stat %s version %s at %s", binary, version, targetDir)
+	}
+
+	if err = os.MkdirAll(targetDir, 0777); err != nil {
```
tstromberg: Would 0700 also work?

afbjorklund:
I thought `minikube cache` was mostly used to add extra stuff, not for Kubernetes itself ?
The name of the parameter (and the idea) was mostly modelled after "apt-get upgrade"....

tstromberg:
Just checking if you plan to address the comments so that this will be merged before 1.0?

afbjorklund:
Working on it, think that it needs to be rewritten to match the images better.

cmd/minikube/cmd/cache.go
afbjorklund: Two different functions now, similar to how othe things handle Cache and Load (here: Copy)

pkg/minikube/bootstrapper/bootstrapper.go
afbjorklund: Maybe, but I didn't actually change that line.

tstromberg:
Please merge once rebased. Thanks!

PR #3742, opened 2019-02-24
Add make target for building a rpm file

afbjorklund:
Similar to the deb, it just packages the standard binary...

It does not build from source, and not for any distribution.


afbjorklund:
For this to work, we need the "rpm" package to be installed on the build server.

Issue #3741, opened 2019-02-24
The deb package has the wrong revision


afbjorklund:
Currently we always build major.minor version and then put the third digit into the revision:

``` Makefile
DEB_VERSION ?= $(VERSION_MAJOR).$(VERSION_MINOR)-$(VERSION_BUILD)
```

https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-version
> The version number of a package. The format is: [epoch:]upstream_version[-debian_revision].

----

This means that we have now built 0.34 **twice**, once for `v0.34.0` and once for `v0.34.1`:

https://github.com/kubernetes/minikube/releases/download/v0.34.0/minikube_0.34-0.deb

https://github.com/kubernetes/minikube/releases/download/v0.34.1/minikube_0.34-1.deb

We should change this to match our tags/releases (0.34.0, 0.34.1), and drop revision (or 0):

> The absence of a debian_revision is equivalent to a debian_revision of 0.

Since we don't build anything but just re-package, we don't need a revision.

Issue #3544, opened 2019-01-18
Bad file ownership in minikube deb packages


mgabeler-lee-6rs:
Please provide the following details:

**Environment**:

**Minikube version**: 0.32.0, 0.33.0, probably others
- **OS**: Debian
- **VM Driver**: N/A
- **ISO version**: N/A
- **Install tools**: N/A
- **Others**: N/A

**What happened**: The package has the binary owned by non-root user/group.  On my system this ended up as `usbmix`/`geoclue`, but that is a somewhat random coincidence.  The package shows that it "thinks" it's trying to install the binary owned by `jenkins`/`jenkins`, which is also wrong, but points to the source of the error.

**What you expected to happen**: The minikube binary to be installed with standard permissions, specifically owned by root/root

**How to reproduce it** (as minimally and precisely as possible): Install the package.  `ls -l /usr/bin/minikube`

**Output of `minikube logs` (if applicable)**: N/A

**Anything else do we need to know**: N/A

tstromberg:
Ouch. That's terrible and should be fixed ASAP. I don't know the cause of this yet, but if anyone happens to, I'd be happy to review a PR which addresses this.

Marking as a v1.0 blocker due to its severity.

leifcr:
Ownership is correct when building the debian package under Ubuntu 18.04 locally using the docker image to build. This likely happens in the CI pipeline somewhere, because the the UID is set to 108, and GID is set to 113. dpkg-deb --build uses the exact UID/GID set on the file while building.

It look likes the builds are done by minikube-bot ?

Here are some options worth trying:
* Add ```fakeroot``` before ```dpkg-deb --build```
* Add ```--root-owner-group``` in ```dpkg-deb``` parameters, while building
* Changing to dpkg-buildpackage. (Gives more options, but stricter file structure when building)

I can give it a try if you have some information regarding how the build is performed for releases.

afbjorklund:
Using `fakeroot dpkg-deb --build` seems to be simplest to implement, and fixes the issue.

It seems to be available in the `$(BUILD_IMAGE)`, and should be available on the VM too.

leifcr:
thanks @afbjorklund for fixing, in #3745

PR #3745, opened 2019-02-24
Improve building of deb package (versioning and permissions)

afbjorklund:
#3741 "The deb package has wrong revision"
#3544 "Bad file ownership in minikube deb packages"


afbjorklund:
@balopat : maybe we should change the package maintainer, as well ?

```
 Package: minikube
 Version: 0.34.1
 Section: base
 Priority: optional
 Architecture: amd64
 Recommends: virtualbox
 Maintainer: Aaron Prindle <aaprindle@gmail.com>
```
https://github.com/kubernetes/minikube/blob/master/installers/linux/deb/minikube_deb_template/DEBIAN/control#L7

PR #3758, opened 2019-02-26
Enable tap and vhost-net in minikube iso

### LINKED ISSUE #3762: vhost-net support in minikube iso for kubevirt
i d like to have vhost-net support in minikube iso as it's needed in kubevirt in order to use minikube

karmab:
These are required to use kubevirt within minikube.

Closes #3762


afbjorklund:
Please open an issue describing why this is needed

afbjorklund:
@karmab : Thanks!

PR #3766, opened 2019-02-27
Improve error handling for versions without `v` prefix

dimitropoulos:
I always forget whether the v is required syntax.  When you start minikube without the v, it currently looks like this:

```
$ minikube start --kubernetes-version 1.10.0
Starting local Kubernetes 1.10.0 cluster...
Starting VM...
Getting VM IP address...
Moving files into cluster...
E0227 14:07:52.609622  117027 start.go:302] Error updating cluster:  generating kubeadm cfg: parsing kubernetes version: parsing kubernetes version: strconv.ParseUint: parsing "": invalid syntax


minikube failed :( exiting with error code 1
```

I added a test or two and tried to make the error message clearer, since, after all, it isn't `""` that I passed so I (as the user) don't have much to go on to understand what it's talking about in the current error.

That said.. I would be happy to just fix things up to allow either the `v` or the no-`v` versions accepted for this flag.  I only didn't do so because I was unsure if this was some kind of design choice.  In the two places in the code (both edited in this PR) where the version is used, it is immediately stripped of its `v` anyway.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -427,6 +427,11 @@ func validateKubernetesVersions(old *cfg.Config) string {
 	if nv == "" {
 		nv = constants.DefaultKubernetesVersion
 	}
+
+	if nv[0] != version.VersionPrefix {
+		exit.WithCode(exit.Usage, "the --kubernetes-version flag requires the version to start with a 'v'")
+	}
+
 	nvs, err := semver.Make(strings.TrimPrefix(nv, version.VersionPrefix))
 	if err != nil {
 		exit.WithCode(exit.Data, "Unable to parse %q: %v", nv, err)
--- a/pkg/minikube/bootstrapper/kubeadm/versions.go
+++ b/pkg/minikube/bootstrapper/kubeadm/versions.go
@@ -161,7 +161,7 @@ func Supports(featureName string) bool {
 
 // ParseKubernetesVersion validates the kubernetes version as legitimate
 func ParseKubernetesVersion(input string) (semver.Version, error) {
-	if input[0] != version.VersionPrefix {
+	if !strings.HasPrefix(input, version.VersionPrefix) {
 		return semver.Version{}, fmt.Errorf("version numbers must begin with %v", version.VersionPrefix)
 	}
 
--- a/pkg/minikube/bootstrapper/kubeadm/versions.go
+++ b/pkg/minikube/bootstrapper/kubeadm/versions.go
@@ -25,11 +25,11 @@ import (
 
 	"github.com/blang/semver"
 	"github.com/golang/glog"
-	"github.com/kubernetes/minikube/pkg/version"
 	"github.com/pkg/errors"
 	"k8s.io/kubernetes/cmd/kubeadm/app/features"
 	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/util"
+	"k8s.io/minikube/pkg/version"
 )
 
 // These are the components that can be configured
--- a/pkg/minikube/bootstrapper/kubeadm/versions.go
+++ b/pkg/minikube/bootstrapper/kubeadm/versions.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/blang/semver"
 	"github.com/golang/glog"
+	"github.com/kubernetes/minikube/pkg/version"
 	"github.com/pkg/errors"
 	"k8s.io/kubernetes/cmd/kubeadm/app/features"
 	"k8s.io/minikube/pkg/minikube/constants"
@@ -158,10 +159,13 @@ func Supports(featureName string) bool {
 	return false
 }
 
-// ParseKubernetesVersion parses the kubernetes version
-func ParseKubernetesVersion(version string) (semver.Version, error) {
-	// Strip leading 'v' prefix from version for semver parsing
-	v, err := semver.Make(version[1:])
+// ParseKubernetesVersion validates the kubernetes version as legitimate
+func ParseKubernetesVersion(input string) (semver.Version, error) {
+	if input[0] != version.VersionPrefix {
+		return semver.Version{}, fmt.Errorf("version numbers must begin with %v", version.VersionPrefix)
+	}
+
+	v, err := semver.Make(strings.TrimPrefix(input, version.VersionPrefix))
 	if err != nil {
 		return semver.Version{}, errors.Wrap(err, "invalid version, must begin with 'v'")
 	}
--- a/pkg/minikube/bootstrapper/kubeadm/versions_test.go
+++ b/pkg/minikube/bootstrapper/kubeadm/versions_test.go
@@ -94,12 +94,45 @@ func TestVersionIsBetween(t *testing.T) {
 }
 
 func TestParseKubernetesVersion(t *testing.T) {
-	version, err := ParseKubernetesVersion("v1.8.0-alpha.5")
-	if err != nil {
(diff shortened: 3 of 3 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -390,6 +390,11 @@ func validateKubernetesVersions(old *cfg.Config) string {
 	if nv == "" {
 		nv = constants.DefaultKubernetesVersion
 	}
+
+	if nv[0] != version.VersionPrefix {
+		exit.WithCode(exit.Data, "the --kubernetes-version flag requires the version to start with a 'v'")
```
tstromberg: small nitpick: Do you mind making this exit.Usage instead of exit.Data? Since the input data is from a command-line, it seems to be more correct:

```
EX_USAGE (64)	   The command was used	incorrectly, e.g., with	the
			   wrong number	of arguments, a	bad flag, a bad	syntax
			   in a	parameter, or whatever.
```

Thanks!

pkg/minikube/bootstrapper/kubeadm/versions.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	"github.com/blang/semver"
 	"github.com/golang/glog"
+	"github.com/kubernetes/minikube/pkg/version"
```
tstromberg: Is this PR missing a file?

dimitropoulos: I'm not sure.  could you please rephrase?  I am hoping this file (and the other kubeadm one) isn't from somewhere else.

cmd/minikube/cmd/start.go
dimitropoulos: done. thanks!

tstromberg:
@minikube-bot OK to test

pkg/minikube/bootstrapper/kubeadm/versions.go
tstromberg: The travis test failed with:

```
Running go tests...
pkg/minikube/bootstrapper/kubeadm/versions.go:28:2: cannot find package "github.com/kubernetes/minikube/pkg/version" in any of:
	/home/travis/gopath/src/k8s.io/minikube/vendor/github.com/kubernetes/minikube/pkg/version (vendor tree)
	/home/travis/.gimme/versions/go1.12.linux.amd64/src/github.com/kubernetes/minikube/pkg/version (from $GOROOT)
	/home/travis/gopath/src/github.com/kubernetes/minikube/pkg/version (from $GOPATH)
make: *** [test] Error 1
The command "make test" exited with 2.
```

I'm not sure why this is happening though, because that package does exist as far as I know. Hmm...

dimitropoulos: I mean.. it's not totally necessary because I feel pretty confident the version prefix isn't changing from `v` any time soon.  I just used it because I noticed it was there and figured it could (if nothing else) be clearer instead of looking like some random magic-character, you know?

tstromberg:
Marking as WIP until @dimitropoulos has a chance to rebase.

dimitropoulos:
ok so I rebased it but there seem to be some other changes that attempt to catch the same thing that have been done since I started... I'll leave it to you to decide how you want to proceed.

pkg/minikube/bootstrapper/kubeadm/versions.go
tstromberg: To pass the tests, this should be:

	"k8s.io/minikube/pkg/version"

Thanks!

tstromberg:
A coworker clued me into what is happening with this error message:

`pkg/minikube/bootstrapper/kubeadm/versions.go:28:2: cannot find package "github.com/kubernetes/minikube/pkg/version" in any of:
	/home/travis/gopath/src/k8s.io/minikube/vendor/github.com/kubernetes/minikube/pkg/version (vendor tree)
	/home/travis/.gimme/versions/go1.12.3.linux.amd64/src/github.com/kubernetes/minikube/pkg/version (from $GOROOT)
	/home/travis/gopath/src/github.com/kubernetes/minikube/pkg/version (from $GOPATH)
`

The import is using `github.com/kubernetes` instead of `k8s.io/minikube`. It is admittedly a bit weird to have this alias, but it should fix the broken tests. Do you mind changing it? 

You'll want to move which directory minikube source is stored in:

https://github.com/kubernetes/minikube/blob/master/docs/contributors/build_guide.md#building-from-source

I suspect when we move to Go modules that this will go away.

dimitropoulos:
perhaps I misunderstood your directions (I don't, myself, understand what the actual problem is either) but I tried to update it and it seems to still fail.

tstromberg:
Excellent, it looks like the failure is at least different now:


```pkg/minikube/bootstrapper/kubeadm/versions.go:164:14: cannot convert "k8s.io/minikube/pkg/version".VersionPrefix (type untyped string) to type byte
pkg/minikube/bootstrapper/kubeadm/versions.go:164:14: invalid operation: input[0] != "k8s.io/minikube/pkg/version".VersionPrefix (mismatched types byte and string)
FAIL	k8s.io/minikube/cmd/minikube/cmd [build failed]
```

I think this can be fixed by using this for your comparison:

`if strings.HasPrefix(input, version.VersionPrefix) {`

If you run `make test` locally, do you see the same error?

tstromberg:
Is this PR still being worked on? I believe this may be a 1-line change away from merging.

dimitropoulos:
I made the change you suggested and it appears to still be failing.  when I test it locally I get:

```
cannot convert "k8s.io/minikube/pkg/version".VersionPrefix (type untyped string) to type byte
```

but I can't seem to figure a way to fix it.

cmd/minikube/cmd/start.go
```diff
@@ -427,6 +427,11 @@ func validateKubernetesVersions(old *cfg.Config) string {
 	if nv == "" {
 		nv = constants.DefaultKubernetesVersion
 	}
+
+	if nv[0] != version.VersionPrefix {
```
tstromberg: To avoid the bytes/string error, this should match the check in versions.go:

`if !strings.HasPrefix(nv, version.VersionPrefix) {`

Alternatively, why not just allow a version to be specified without a leading v? We know what the user wants, so instead of raising the error, we could just give it to them and make their life easier:

```go
if !strings.HasPrefix(nv, version.VersionPrefix) {
  nv = version.VersionPrefix + nv
}
```

pkg/minikube/bootstrapper/kubeadm/versions.go
```diff
@@ -158,10 +159,13 @@ func Supports(featureName string) bool {
 	return false
 }
 
-// ParseKubernetesVersion parses the kubernetes version
-func ParseKubernetesVersion(version string) (semver.Version, error) {
-	// Strip leading 'v' prefix from version for semver parsing
-	v, err := semver.Make(version[1:])
+// ParseKubernetesVersion validates the kubernetes version as legitimate
+func ParseKubernetesVersion(input string) (semver.Version, error) {
+	if !strings.HasPrefix(input, version.VersionPrefix) {
```
tstromberg: I think this check can be removed: it isn't necessary to check for the leading 'v' here because it is stripped away immediately afterwards.  Just continue to use strings.TrimPrefix, and it will handle both cases perfectly.

tstromberg:
As I haven't heard anything recently and I bumped into this issue myself, I took inspiration from this PR to create a new PR: #4568

dimitropoulos:
great.  thanks @tstromberg

PR #3773, opened 2019-02-28
Split color and emoji into separate settings

afbjorklund:
Currently it was using the color setting to
actually determine whether or not to use emoji.

Even if those happened to be in black-and-white.
And without actually using any color (text)...

### DIFF
--- a/docs/env_vars.md
+++ b/docs/env_vars.md
@@ -13,7 +13,9 @@ Some features can only be accessed by environment variables, here is a list of t
 
 * **MINIKUBE_HOME** - (string) sets the path for the .minikube directory that minikube uses for state/configuration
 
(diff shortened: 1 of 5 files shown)


tstromberg:
I don't think this is the right way forward, for the following reasons:

- With this PR, MINIKUBE_IN_COLOR does nothing.
- When there is emoji, we have no control over whether the OS displays them in color or not. There is no way to request black & white emoji or color emoji.
- This adds a second configuration knob when one would do.

Instead, I propose renaming this environment variable to basically distinguish between applying advanced styling versus a basic 7-bit ASCII display? 

What do you think about `MINIKUBE_IN_STYLE` or something of that sort?

afbjorklund:
I think renaming `MINIKUBE_IN_COLOR` to `MINIKUBE_IN_STYLE` would do the trick.

Issue #3087, opened 2018-08-21
Missing kernel modules for kube-proxy's IPVS mode


residentsummer:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
FEATURE REQUEST

Please provide the following details:

**Environment**:

minikube version: v0.28.2
OS: MacOS 10.13
VM driver: xhyve
ISO version: ~/.minikube/cache/iso/minikube-v0.28.1.iso
- **Install tools**:
- **Others**:

**What happened**:
kube-proxy won't start in ipvs mode

**What you expected to happen**:
kube-proxy should start in ipvs mode

**How to reproduce it** (as minimally and precisely as possible):

After minikube boots, set `mode: "ipvs"` in `config.conf` in `kube-proxy` ConfigMap:

```bash
kubectl edit -n kube-system configmap/kube-proxy
```

To apply new configuration, delete the old pod and k8s will create a new one,
as required by corresponding DaemonSet:

```bash
$ kc get -n kube-system pods
NAME                                    READY     STATUS    RESTARTS   AGE
...
kube-proxy-49psk                        1/1       Running   0          11h
...
$ kc delete -n kube-system po/kube-proxy-49psk
pod "kube-proxy-49psk" deleted
$ kc get -n kube-system pods
NAME                                    READY     STATUS    RESTARTS   AGE
...
kube-proxy-x7qgq                        1/1       Running   0          7m
...
```

**Output of `minikube logs` (if applicable)**:

```conf
$ kc logs -n kube-system po/kube-proxy-x7qgq
E0805 09:46:12.625751       1 ipset.go:156] Failed to make sure ip set: &{{KUBE-CLUSTER-IP hash:ip,port inet 1024 65536 0-65535 Kubernetes service cluster ip + port for masquerade purpose} map[] 0xc420562080} exist, error: error creating ipset KUBE-CLUSTER-IP, error: exit status 1
E0805 09:46:42.645604       1 ipset.go:156] Failed to make sure ip set: &{{KUBE-LOAD-BALANCER-FW hash:ip,port inet 1024 65536 0-65535 Kubernetes service load balancer ip + port for load balancer with sourceRange} map[] 0xc420562080} exist, error: error creating ipset KUBE-LOAD-BALANCER-FW, error: exit status 1
[...]

PR #3783, opened 2019-03-02
Enable kernel modules necessary for IPVS

mdonkers:
To be able to use Kubernetes' IPVS proxy-mode, certain (IP Set) kernel modules
need to be enabled during kernel build time.
Building a new Minikube image with these modules enabled, will allow
others to switch kube-proxy in IPVS mode and experiment with it inside
Minikube.


mdonkers:
PR for issue: https://github.com/kubernetes/minikube/issues/3087

PR #3787, opened 2019-03-04
Failure to read documentation isn't really a crash

afbjorklund:
Show a friendly link to the documentation instead


tstromberg:
I think this hard-coded approach is a great initial look at the problem. Thanks!

I can easily see this growing into a proper table of solutions for common problems, defined in a YAML file, perhaps even one that could even be referenced & updated between releases.

PR #3796, opened 2019-03-05
Change DefaultMountVersion to 9p2000.L

### LINKED ISSUE #2290: Minikube mount sets incorrect ownership
**Is this a BUG REPORT or FEATURE REQUEST?**: BUG REPORT

**Environment**:

**Minikube version**: v0.24.1
- **OS**: macOS Sierra 10.12.6
- **VM Driver**: virtualbox
- **ISO version**: v0.23.6

**What happened**:
minikube mount sets ownership of mounted files to uid 501, gid 20

minikube mount sets ownership of mounted files to uid 1001, gid 1001

minikube mount /Users/username:/var/mnt/username
minikube ssh 'ls -ld /var/mnt/username'

`drwxrwxr-x 1 501 20 1836 Dec  8 07:35 /var/mnt/username`

The automounted /Users path in minikube is correctly owned by docker/docker.  Under that path, I have not experienced any issues.  Unfortunately, using the default /Users path forces PodSpecs with baked in usernames for volume mounts.  Using minikube mount allows us to avoid coupling deployments to specific user directories, but the permissions appear to be causing IO errors.  When trying to set up for use cases like live code reloading, some of our node projects fail their npm installs with EIO errors.  When using the /Users path instead of a minikube mount path, those npm installs complete without issue.

kron4eg:
9p2000.u is a legacy protocol version. This upgrades us to the most recent version with better permissions handling. From the Linux kernel docs: https://landley.net/kdocs/Documentation/filesystems/9p.txt

```
Around 2003 Plan 9 was open sourced, which led to the development of an
extended version of the protocol, "9p2000.U" (for Unix) providing full posix
semantics .... The current protocol version, "9p2000.L" (for Linux) was developed for v9fs
after a few years of real world experience. It tailors the protocol to Linux
with support for ACLs, file locking, more efficient directory listing, corner
cases in things like file deletion, and so on. 
```

Fixes #2290

### DIFF
--- a/pkg/minikube/constants/constants.go
+++ b/pkg/minikube/constants/constants.go
@@ -161,7 +161,7 @@ const (
 	DefaultUfsDebugLvl   = 0
 	DefaultMountEndpoint = "/minikube-host"
 	DefaultMsize         = 262144
-	DefaultMountVersion  = "9p2000.u"
+	DefaultMountVersion  = "9p2000.L"
 )
 
 func GetKubernetesReleaseURL(binaryName, version string) string {


tstromberg:
@minikube-bot OK to test

tstromberg:
I think this PR is a great idea. Thanks for proposing it! Will approve once I can see that the integration tests pass.

afbjorklund:
Looks OK, but I think we need to also upgrade the server for it to be of most use ?

kron4eg:
@afbjorklund I'm not sure I got you. Do you mean `third_party/go9p` library?

tstromberg:
@minikube-bot OK to test

FWIW, I think updating the server is unrelated to this version change. L has been around for a very long time.

afbjorklund:
@kron4eg yes, I think that is serving ?

afbjorklund:
Afaik, it will negotiate. So we should start with updating this client, but also update server

kron4eg:
Unfortunately I can't find from where `third_party/go9p` originated. But it worked with old minikube releases (oldest tested is from 2 years ago) when I've explicitly specified `--9p-version=9p2000.L`

PR #3810, opened 2019-03-06
Update minikube to be compatible with Kubernetes v1.14.0-beta.1

### LINKED ISSUE #3612: Assert compatibility with kubernetes v1.14.x
v1.14.0-alpha2 now shows up at https://github.com/kubernetes/kubernetes/releases

tstromberg:
Updates to the latest kubeadm config API (`v1beta1`), and removes the `Initializers` admission controller that was removed in v1.14.

Closes #3612


pkg/minikube/bootstrapper/kubeadm/templates.go
```diff
@@ -39,16 +40,16 @@ etcd:
   dataDir: {{.EtcdDataDir}}
 nodeName: {{.NodeName}}
 {{if .CRISocket}}criSocket: {{.CRISocket}}
-{{end}}{{range .ExtraArgs}}{{.Component}}:{{range $i, $val := printMapInOrder .Options ": " }}
+{{end}}{{range .ExtraArgs}}{{.Component}}ExtraArgs:{{range $i, $val := printMapInOrder .Options ": " }}
```
tstromberg: Is this necessary?

tstromberg: Yes - it is. We need the raw component names for v1beta1 template, since v1.14 has a sane YAML layout.

pkg/util/constants.go
```diff
@@ -45,6 +45,9 @@ var DefaultAdmissionControllers = []string{
 	"ResourceQuota",
 }
 
+// DefaultLegacyAdmissionControllers are admission controllers we default to in order Kubernetes releases
```
balopat:
```suggestion
// DefaultLegacyAdmissionControllers are admission controllers we default to in other Kubernetes releases
```

tstromberg: Done. Thanks for flagging that!

PR #3817, opened 2019-03-07
Add missing variable and upload of rpm to github

afbjorklund:


hack/jenkins/release_github_page.sh
```diff
@@ -121,6 +122,7 @@ FILES_TO_UPLOAD=(
     'minikube-windows-amd64.sha256'
```
tstromberg: While you are here, I'm pretty sure this should be minikube-windows-amd64.exe.sha256

afbjorklund: Yes, that looks right.

afbjorklund:
For the `v0.35.0`, I just removed the rpm link from the release text.

hack/jenkins/release_github_page.sh
afbjorklund: Added https://github.com/kubernetes/minikube/pull/3838

PR #3819, opened 2019-03-07
Fix tunnel error logging

calind:
The message when patching services with `minikube tunnel` was wrongfully logged as error.


tstromberg:
Thank you for the fix!

PR #3833, opened 2019-03-08
Cruntime version

afbjorklund:
Output the container runtime version, for weird setups (like `none` custom VMs or older ISO, or similar)

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -479,8 +479,8 @@ func configureRuntimes(h *host.Host, runner bootstrapper.CommandRunner) cruntime
 	if err != nil {
 		exit.WithError("Failed to enable container runtime", err)
 	}
-	version := cr.Version()
-	if version != "" {
+	version, err := cr.Version()
+	if err == nil {
 		console.OutStyle(cr.Name(), "Version of container runtime is %s", version)
 	}
 	return cr
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -479,6 +479,10 @@ func configureRuntimes(h *host.Host, runner bootstrapper.CommandRunner) cruntime
 	if err != nil {
 		exit.WithError("Failed to enable container runtime", err)
 	}
+	version := cr.Version()
+	if version != "" {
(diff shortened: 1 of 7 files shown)


pkg/minikube/cruntime/containerd.go
```diff
@@ -33,6 +34,19 @@ func (r *Containerd) Name() string {
 	return "containerd"
 }
 
+// Version retrieves the current version of this runtime
+func (r *Containerd) Version() string {
+	ver, err := r.Runner.CombinedOutput("containerd --version")
+	// containerd github.com/containerd/containerd v1.2.0 c4446665cb9c30056f4998ed953e6d4ff22c7c39
+	if err == nil {
```
tstromberg: Try to keep the normal code path at a minimal indentation by terminating the exceptional condition early on. For instance:

if err != nil {
  return err
}
<everything else>

pkg/minikube/cruntime/rkt.go
```diff
@@ -33,6 +34,18 @@ func (r *Rkt) Name() string {
 	return "rkt"
 }
 
+// Version retrieves the current version of this runtime
+func (r *Rkt) Version() string {
+	ver, err := r.Runner.CombinedOutput("rkt version")
+	// rkt Version: 1.24.0
+	// appc Version: 0.8.10
+	if err == nil {
+		line := strings.Split(ver, "\n")[0]
```
tstromberg: Same here..

afbjorklund:
Will do, most likely the breakage will be in the version command output (like changing format or that client/server vs CE/EE etc), but there should be some tests on our end.

pkg/minikube/cruntime/containerd.go
afbjorklund: Done, just returning empty string to keep the API simple

cmd/minikube/cmd/start.go
```diff
@@ -479,6 +479,10 @@ func configureRuntimes(h *host.Host, runner bootstrapper.CommandRunner) cruntime
 	if err != nil {
 		exit.WithError("Failed to enable container runtime", err)
 	}
+	version, err := cr.Version()
+	if err == nil {
```
tstromberg: Logging the error might be nice here.

PR #3834, opened 2019-03-08
Automatically propagate proxy environment variables to docker env

### LINKED ISSUE #3242: Propagate HTTP proxy config into VM
**BUG REPORT**

**Minikube version** (use `minikube version`): v0.30.0
- **OS** (e.g. from /etc/os-release): Windows
- **VM Driver** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep DriverName`): Virtualbox
- **ISO version** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep -i ISO` or `minikube ssh cat /etc/VERSION`): v0.30.0

**What happened**:
`minikube start --kubernetes-version v1.10.8 --docker-env HTTP_PROXY=xxx --docker-env HTTPS_PROXY=xxx --docker-env NO_PROXY=xxx` is failing in timeout during kubeadm command execution

```
sudo /usr/bin/kubeadm init --config /var/lib/kubeadm.yaml --ignore-preflight-errors=DirAvailable--etc-kubernetes-manifests --ignore-preflight-errors=DirAvailable--data-minikube --ignore-preflight-errors=Port-10250 --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-kube-scheduler.yaml --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-kube-apiserver.yaml --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-kube-controller-manager.yaml --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-etcd.yaml --ignore-preflight-errors=Swap --ignore-preflight-errors=CRI  &&
sudo /usr/bin/kubeadm alpha phase addon kube-dns

.: Process exited with status 1
```
minikube starts normally

The faulty command is `kubeadm alpha phase addon kube-dns`. I replayed it inside the minikube VM to have more logs : 
```
$ sudo kubeadm alpha phase addon kube-dns
unable to get URL "https://dl.k8s.io/release/stable-1.10.txt": Get https://dl.k8s.io/release/stable-1.10.txt: dial tcp 23.236.58.218:443: getsockopt: connection timed out
```
kubeadm is trying to download a file (that actually contains the exact version of kubernetes) but since HTTP proxy settings are not set in VM (they are only set for docker engine through `--docker-env` parameters), this is failing. I found that if you add `--kubernetes-version`, no download is done and the command succeeds:
```
$ sudo kubeadm alpha phase addon kube-dns --kubernetes-version v1.10.8
[addons] Applied essential addon: kube-dns
```
So a fix could be to propagate the `--kubernetes-version` passed to `minikube start` to `kubeadm alpha phase addon kube-dns` command.

tstromberg:
Currently, our [documentation](https://github.com/kubernetes/minikube/blob/master/docs/http_proxy.md) recommends that users use:

```
minikube start \
--docker-env=HTTP_PROXY=$HTTP_PROXY \
--docker-env HTTPS_PROXY=$HTTPS_PROXY \
 --docker-env NO_PROXY=$NO_PROXY
```

This makes the setting of --docker-env automatic if the environment variables are set, and partially resolves #3242

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -96,6 +96,9 @@ var (
 	apiServerNames   []string
 	apiServerIPs     []net.IP
 	extraOptions     pkgutil.ExtraOptionSlice
+
+	// proxyVars are variables we plumb through to the underlying container runtime
+	proxyVars = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
 )
 
 func init() {
@@ -260,6 +263,17 @@ func generateConfig(cmd *cobra.Command, kVersion string) (cfg.Config, error) {
 		}
 	}
 
+	// Feed Docker our host proxy environment by default, so that it can pull images
+	if _, ok := r.(*cruntime.Docker); ok {
+		if !cmd.Flags().Changed("docker-env") {
+			for _, k := range proxyVars {
+				if v := os.Getenv(k); v != "" {
+					dockerEnv = append(dockerEnv, fmt.Sprintf("%s=%s", k, v))
+				}
+			}
+		}
+	}
+
 	cfg := cfg.Config{
 		MachineConfig: cfg.MachineConfig{
 			MinikubeISO:         viper.GetString(isoURL),
@@ -370,7 +384,7 @@ func validateNetwork(h *host.Host) string {
 	console.OutStyle("connectivity", "%q IP address is %s", cfg.GetMachineName(), ip)
 
 	optSeen := false
-	for _, k := range []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"} {
+	for _, k := range proxyVars {
 		if v := os.Getenv(k); v != "" {
 			if !optSeen {
 				console.OutStyle("internet", "Found network options:")


tstromberg:
Closing in preference to #3835 - which is admittedly a bit more advanced.

tstromberg:
Reopening, since this is a good introductory approach.

afbjorklund:
It would be nice to have something more portable as, well ?

i.e. those proxy variables are also needed for other runtimes

cmd/minikube/cmd/start.go
```diff
@@ -260,6 +263,17 @@ func generateConfig(cmd *cobra.Command, kVersion string) (cfg.Config, error) {
 		}
 	}
 
+	// Feed Docker our host proxy environment by default, so that it can pull images
+	if _, ok := r.(*cruntime.Docker); ok {
+		if !cmd.Flags().Changed("docker-env") {
+			for _, k := range proxyVars {
+				if v := os.Getenv(k); v != "" {
+					dockerEnv = append(dockerEnv, fmt.Sprintf("%s=%s", k, v))
```
tejal29: so, potentially user can have multiple dockerEnv variables set.
In case, user can have NO_PROXY as well as "HTTPS_PROXY" set which shd error out right? or this handled later?

tstromberg: > It would be nice to have something more portable as, well ?

Yes, I agree. This is an introductory PR that only solves the problem for most (99%?) of users who use the defaults. Other runtimes so far don't have a mechanism for introducing environment variables, but I suspect that #3835 should cover them nicely, which adds them to /etc/environment.

tstromberg: If a user manually sets --docker-env, this PR is disabled. 

NO_PROXY and HTTPS_PROXY are complimentary, and thus plumbed into the environment as expected.

tejal29: cool.

PR #3838, opened 2019-03-10
Fix exe suffix and missing checksum for windows binary

afbjorklund:
Windows needs that DOS suffix, so we make a copy


afbjorklund:
The files were built alright, but missing from the GitHub releases page:
https://github.com/kubernetes/minikube/releases/tag/v0.35.0

> **Windows [Experimental]**
> Download the minikube-windows-amd64.exe file, rename it to minikube.exe and add it to your path.

Issue #3805, opened 2019-03-06
[FEATURE] feature-gates via minikube config set ...


Doqnach:
I would love to be able to set defaults for `--feature-gates` just like for `--vm-driver` etc.

```
$ minikube config set feature-gates foo
💣  Set failed: Property name feature-gates not found
```

PR #3861, opened 2019-03-12
feature-gates via minikube config set

RA489:
[FEATURE] feature-gates via minikube config set
fixes https://github.com/kubernetes/minikube/issues/3805


tstromberg:
@minikube-bot OK to test

PR #3879, opened 2019-03-14
--extra-config now work for kubeadm as well

marcosdiez:
This small patch allows us to use `--extra-config` also with `kubeadm`.

My primary motivation is to run `minikube` with the `none` vm driver in any version of docker.

This patch allows one to do so like this: `sudo -E minikube start --vm-driver=none --kubernetes-version v1.11.8 --extra-config kubeadm.ignore-preflight-errors=SystemVerification `

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -135,7 +135,7 @@ func init() {
 	startCmd.Flags().Var(&extraOptions, "extra-config",
 		`A set of key=value pairs that describe configuration that may be passed to different components.
 		The key should be '.' separated, and the first part before the dot is the component to apply the configuration to.
-		Valid components are: kubelet, apiserver, controller-manager, etcd, proxy, scheduler.`)
+		Valid components are: kubelet, kubeadm, apiserver, controller-manager, etcd, proxy, scheduler.`)
 	startCmd.Flags().String(uuid, "", "Provide VM UUID to restore MAC address (only supported with Hyperkit driver).")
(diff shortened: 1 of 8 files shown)


afbjorklund:
This is quite a convoluted syntax for the check, and it seems that it is duplicating the preflight checks in two different places now. Can see the reason for skipping Docker version check (surely they will name the next version something weird and break it again), but supporting generic parameters to kubeadm might be hard to do. Would rather see something more specific to SystemVerification at this point in time.

marcosdiez:
hey, @afbjorklund !

An alternative for my generic approach would be similar to what was implemented here: https://github.com/kubernetes/minikube/pull/3892

The problem with that implementation is that minikube will one day become a mess if we start exposing each and every parameter from other tools to it. Also, my approach is future compatible.

That being said, I am neutral to how this get's implemented as long as I can tell minikube to tell kubeadm not to complain about docker versions.
So if you tell me that you prefer the other way, I can implement that as well.

Please tell me if I can help you in any other way!
Marcos

tstromberg:
If you add an example command-line to test_start_stop_delete.go, I'd be good to merge this.

marcosdiez:
Hi @tstromberg !
I just added the extra line on `start_stop_delete_test.go` to test this.

And since trevis does not do integration tests, here is my passing output: 
https://gist.github.com/marcosdiez/88e029d2068cf349c1daa1fa97585ff9

this is the command I used to run the tests:
`env TEST_ARGS="-minikube-start-args=--vm-driver=none -test.run TestStartStop" sudo -E make integration`

(I must use `--vm-driver=none` because I run linux inside of a VM, so other vm drivers are not an option to me)

tstromberg:
@minikube-bot OK to test

tstromberg:
@minikube-bot OK to test

PR #3892, opened 2019-03-18
Expose ‘—pod-network-cidr’ argument in minikube

### LINKED ISSUE #3865: Expose pod-network-cidr on kubeadm
Getting Calico working in Kubernetes / Minikube can be tricky because you need to set the pod network cidr on multiple components (apiserver, controller-manager, and proxy) individually.

Kubeadm has made this easier by tying all three of those flags to a single "high-level" flag - `--pod-network-cidr`.

Now that minikube is using kubeadm under the hood, could we expose the `--pod-network-cidr` flag on it?

After a brief look at the code, I think we'd just need to template this value: https://github.com/kubernetes/minikube/blob/3714bbd787f977b3d3a8b5460a249aff901e0f64/pkg/minikube/bootstrapper/kubeadm/templates.go#L84

11janci:
Exposes --pod-network-cidr arg in minikube and forwards it to kubeadm so that it is not necessary to configure the cidr on multiple components separately.

Closes #3865

### DIFF
--- a/test/integration/start_stop_delete_test.go
+++ b/test/integration/start_stop_delete_test.go
@@ -19,7 +19,6 @@ limitations under the License.
 package integration
 
 import (
-	"encoding/json"
 	"fmt"
 	"net"
 	"strings"
@@ -33,19 +32,19 @@ import (
 
 func TestStartStop(t *testing.T) {
 	tests := []struct {
-		name         string
-		args         []string
-		assertCustom func(t *testing.T)
+		name string
+		args []string
 	}{
 		{"nocache_oldest", []string{
 			"--cache-images=false",
 			fmt.Sprintf("--kubernetes-version=%s", constants.OldestKubernetesVersion),
-		}, nil},
+		}},
 		{"feature_gates_newest_cni", []string{
 			"--feature-gates",
 			"ServerSideApply=true",
 			"--network-plugin=cni",
 			"--extra-config=kubelet.network-plugin=cni",
+			"--extra-config=kubeadm.pod-network-cidr=192.168.111.111/16",
 			fmt.Sprintf("--kubernetes-version=%s", constants.NewestKubernetesVersion),
 		}},
 		{"containerd_and_non_default_apiserver_port", []string{
@@ -57,10 +56,7 @@ func TestStartStop(t *testing.T) {
 			"--container-runtime=crio",
 			"--extra-config",
 			"kubeadm.ignore-preflight-errors=SystemVerification",
-		}, nil},
-		{"podCidr", []string{
-			"--pod-network-cidr=192.168.111.111/16",
-		}, assertPodCIDR},
+		}},
 	}
 
 	for _, test := range tests {
@@ -75,10 +71,6 @@ func TestStartStop(t *testing.T) {
 			r.CheckStatus(state.None.String())
 			r.Start(test.args...)
 			r.CheckStatus(state.Running.String())
-
-			if test.assertCustom != nil {
-				test.assertCustom(t)
-			}
 
 			ip := r.RunCommand("ip", true)
 			ip = strings.TrimRight(ip, "\n")
@@ -116,26 +108,5 @@ func TestStartStop(t *testing.T) {
 			r.RunCommand("delete", true)
 			r.CheckStatus(state.None.String())
 		})
-	}
-}
-
-func assertPodCIDR(t *testing.T) {
-	kr := util.NewKubectlRunner(t)
-	out, err := kr.RunCommand([]string{"get", "nodes", "-o", "json"})
-	if err != nil {
-		t.Fatalf("Failed to obtain nodes info")
-	}
-
-	var result map[string]interface{}
-	json.Unmarshal([]byte(out), &result)
-
-	items := result["items"].([]interface{})
-	for _, item := range items {
-		spec := item.(map[string]interface{})["spec"]
-		podCidr := spec.(map[string]interface{})["podCIDR"].(string)
-
-		if !strings.HasPrefix(podCidr, "192.168.0.0") {
-			t.Errorf("Unexpected podCIDR: %s", podCidr)
-		}
 	}
 }
--- a/test/integration/start_stop_delete_test.go
+++ b/test/integration/start_stop_delete_test.go
@@ -19,6 +19,7 @@ limitations under the License.
 package integration
 
 import (
+	"encoding/json"
 	"fmt"
 	"net"
 	"strings"
@@ -32,13 +33,14 @@ import (
 
 func TestStartStop(t *testing.T) {
 	tests := []struct {
-		name string
-		args []string
+		name         string
+		args         []string
+		assertCustom func(t *testing.T)
 	}{
 		{"nocache_oldest", []string{
 			"--cache-images=false",
 			fmt.Sprintf("--kubernetes-version=%s", constants.OldestKubernetesVersion),
-		}},
+		}, nil},
 		{"feature_gates_newest_cni", []string{
 			"--feature-gates",
 			"ServerSideApply=true",
@@ -55,7 +57,10 @@ func TestStartStop(t *testing.T) {
 			"--container-runtime=crio",
 			"--extra-config",
 			"kubeadm.ignore-preflight-errors=SystemVerification",
-		}},
+		}, nil},
+		{"podCidr", []string{
+			"--pod-network-cidr=192.168.111.111/16",
+		}, assertPodCIDR},
 	}
 
 	for _, test := range tests {
@@ -70,6 +75,10 @@ func TestStartStop(t *testing.T) {
 			r.CheckStatus(state.None.String())
 			r.Start(test.args...)
 			r.CheckStatus(state.Running.String())
+
+			if test.assertCustom != nil {
+				test.assertCustom(t)
+			}
 
 			ip := r.RunCommand("ip", true)
 			ip = strings.TrimRight(ip, "\n")
@@ -107,5 +116,26 @@ func TestStartStop(t *testing.T) {
 			r.RunCommand("delete", true)
 			r.CheckStatus(state.None.String())
 		})
+	}
+}
+
+func assertPodCIDR(t *testing.T) {
+	kr := util.NewKubectlRunner(t)
+	out, err := kr.RunCommand([]string{"get", "nodes", "-o", "json"})
+	if err != nil {
+		t.Fatalf("Failed to obtain nodes info")
+	}
+
+	var result map[string]interface{}
+	json.Unmarshal([]byte(out), &result)
+
+	items := result["items"].([]interface{})
+	for _, item := range items {
+		spec := item.(map[string]interface{})["spec"]
+		podCidr := spec.(map[string]interface{})["podCIDR"].(string)
+
+		if !strings.HasPrefix(podCidr, "192.168.0.0") {
+			t.Errorf("Unexpected podCIDR: %s", podCidr)
+		}
 	}
 }
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -79,6 +79,7 @@ const (
 	apiServerPort         = "apiserver-port"
 	dnsDomain             = "dns-domain"
 	serviceCIDR           = "service-cluster-ip-range"
+	podSubnet             = "pod-network-cidr"
 	imageRepository       = "image-repository"
 	imageMirrorCountry    = "image-mirror-country"
 	mountString           = "mount-string"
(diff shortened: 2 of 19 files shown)


marcosdiez:
Just for the record I had a similar problem of needing to pass a specific parameter to kubeadm and I implemented a generic approach to solve it: https://github.com/kubernetes/minikube/pull/3879
It has not been merged yet. I am not sure if my approach is better or worse, but I thought it would be relevant to have it mentioned here.

11janci:
I'd be happy to move it under the --extra-config flag if @marcosdiez 's PR gets approved and other reviewers agree with the approach (I personally think it makes sense to move it there).

tstromberg:
#3879 was merged. Should this PR be closed or altered?

11janci:
@tstromberg I'll be happy to alter it. Do you want to just resolve the conflicts and keep --pod-network-cidr as a standalone minikube param or do you want to move it under the --extra-config param?

11janci:
/ptal
@tstromberg 
the existing generic solution (using --extra-conf) done within #3879 does not work for --pod-network-cidr as kubeadm does not allow this parameter to be combined with the --config param. kubeadm allows only following parameters to be used when --config is specified (see kubernetes/cmd/kubeadm/app/apis/kubeadm/validation/validation.go:ValidateMixedArguments):
```
kubeadmcmdoptions.CfgPath,
kubeadmcmdoptions.IgnorePreflightErrors,
kubeadmcmdoptions.DryRun,
kubeadmcmdoptions.KubeconfigPath,
kubeadmcmdoptions.NodeName,
kubeadmcmdoptions.NodeCRISocket,
kubeadmcmdoptions.KubeconfigDir,
kubeadmcmdoptions.UploadCerts,
kubeadmcmdoptions.CertificateKey,
"print-join-command"
"rootfs",
"v"
```
The rest of the kubeadm parameters are not allowed and thus will fail if specified in the --extra-conf param in minikube. 
Here are some ideas how to fix this:

- update kubeadm to allow all parameters with the --config option (too big of a change change so probably not feasible)
- expose parameters which are not allowed to be used in extra-conf as top-level minikube params, i.e. like this PR does. Probably would be better to drop supporting kubeadm params in extra-conf in this case
- add some handling in minikube that would contain a whitelist of supported kubeadm parameters. The ones that are allowed to be combined with --conf would be propagated to kubeadm as parameters, the rest could be manually added to the config file. E.g. the --pod-network-cidr would not be used as a kubeadm param, but manually put into the config file similarly as done in this PR.

If you would like to keep the --extra-conf support for kubeadm, I find the last option the best. It would fail immediately for unsupported params, is extensible in the future to support additional parameters and obviously fixes the problem.

tstromberg:
@11janci - Thanks for the update. Let me give it some thought, as maintaining a whitelist sounds like something that could easily go out of date and cause issues as well. Hmm..

tstromberg:
@11janci - I thought about it some more, and agree that the 3rd item makes the most sense. I don't think that particular whitelist will be too much of burden to maintain.

 That said, I'm flexible and will accept this PR in either direction. Thank you again for contributing it!

11janci:
@tstromberg cool thx, I will look into it over the weekend and hopefully come up with a PR.

tstromberg:
Please check the failing test for `TestGenerateConfig/containerd-pod-network-cidr__default`

11janci:
@tstromberg I added the whitelist, please take another look

PatrickLang:
@11janci  - I built from your branch but am hitting some issues even without using the whitelist. Does this work with one of the other `--vm-driver`s on your machine?

```
.\minikube-windows-amd64.exe start --vm-driver=hyperv --network-plugin=cni --enable-default-cni
o   minikube v1.0.1 on windows (amd64)
$   Downloading Kubernetes v1.14.1 images in the background ...
i   Tip: Use 'minikube start -p <name>' to create a new cluster, or 'minikube delete' to delete this one.
:   Restarting existing hyperv VM for "minikube" ...
:   Waiting for SSH access ...
-   "minikube" IP address is 192.168.1.137
-   Configuring Docker as the container runtime ...
-   Version of container runtime is 18.06.3-ce
:   Waiting for image downloads to complete ...
-   Preparing Kubernetes environment ...
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x18 pc=0x1087811]

goroutine 878 [running]:
k8s.io/minikube/pkg/minikube/assets.(*BinDataAsset).Read(0xc0002fae80, 0xc000736000, 0x8000, 0x8000, 0x0, 0x7550300, 0x1b)
        /go/src/k8s.io/minikube/pkg/minikube/assets/vm_assets.go:230 +0x31
io.copyBuffer(0x5370db8, 0xc00047ed00, 0x7550320, 0xc0002fae80, 0xc000736000, 0x8000, 0x8000, 0x0, 0x0, 0x0)
        /usr/local/go/src/io/io.go:402 +0x129
io.Copy(...)
        /usr/local/go/src/io/io.go:364
k8s.io/minikube/pkg/minikube/bootstrapper.(*SSHRunner).Copy.func1(0xc0003221a0, 0x16b2400, 0xc00047ed00, 0x16e4ba0, 0xc0002fae80)
        /go/src/k8s.io/minikube/pkg/minikube/bootstrapper/ssh_runner.go:181 +0x361
created by k8s.io/minikube/pkg/minikube/bootstrapper.(*SSHRunner).Copy
        /go/src/k8s.io/minikube/pkg/minikube/bootstrapper/ssh_runner.go:176 +0x3da
```

11janci:
Hi @PatrickLang,
I am able to run it on a Mac using the following command: 
`minikube start --vm-driver=virtualbox --network-plugin=cni --enable-default-cni`

It seems quite unlikely that the error you posted is caused by the changes in this PR, shouldn't be related. Did you try to run it from the master branch to see if you're getting same error?
Thx for looking into it :)

PatrickLang:
1.0.1 works but I haven't tried master yet. Will try tomorrow

PatrickLang:
The crash isn't happening today. I built from master, it worked. I did a `git merge` to bring in your changes, built, and no crash. However, I still can't get it working.

So what was the usage you decided on?

I ran `.\minikube-windows-amd64_prmerge.exe start --vm-driver=hyperv --network-plugin=cni --enable-default-cni --extra-config=kubeadm.pod-network-cidr=10.244.0.0/16 --v=7`

The same error was returned:

```
X Error starting cluster: kubeadm init:
sudo /usr/bin/kubeadm init --config /var/lib/kubeadm.yaml --pod-network-cidr=10.244.0.0/16 --ignore-preflight-errors=DirAvailable--etc-kubernetes-manifests --ignore-preflight-errors=DirAvailable--data-minikube --ignore-preflight-errors=Port-10250 --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-kube-scheduler.yaml --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-kube-apiserver.yaml --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-kube-controller-manager.yaml --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-etcd.yaml --ignore-preflight-errors=Swap --ignore-preflight-errors=CRI

can not mix '--config' with arguments [pod-network-cidr]

: Process exited with status 1

* Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
  - https://github.com/kubernetes/minikube/issues/new
```

11janci:
did you rebuilt after the merge? cause that should work and it does on my machine:
Screen Shot 2019-05-08 at 10 45 55 PM
(fyi I rebased off latest master)

PatrickLang:
Hmm, I must have missed a step somewhere. 

Good news, this gets a working flannel setup in host-gw mode:

```
.\minikube-windows-amd64.exe start --vm-driver=hyperv --network-plugin=cni --enable-default-cni --extra-config=kubeadm.pod-network-cidr=10.244.0.0/16 --extra-config=controller-manager.allocate-node-cidrs=true --extra-config=controller-manager.cluster-cidr=10.244.0.0/16
kubectl apply -f https://gist.githubusercontent.com/PatrickLang/02b7c504fecdd08b498d13bdeee88186/raw/b77ab11e02dba40b597ff6614c401cd71827fa81/kube-flannel.yml
```

@ksubrmnn this could make the `kubeadm` adjustments for Windows easier since it's easy to bring up a 1-node cluster with `minikube`. I still haven't tried joining more nodes yet but it seems like it should work.

PatrickLang:
cc @neolit123 ^^

11janci:
good to hear, thank you sir :)

neolit123:
> can not mix '--config' with arguments [pod-network-cidr]

`+`

> update kubeadm to allow all parameters with the --config option (too big of a change change so probably not feasible)

similar flag -> config overrides might work eventually in the long term (requires component config work).  for now we encourage using `--config`. technically the kubeadm config API (e.g. v1beta1) can be used from a go template which is written to a temp directory and passed via `--config`.

@PatrickLang 

> @ksubrmnn this could make the kubeadm adjustments for Windows easier since it's easy to bring up a 1-node cluster with minikube. I still haven't tried joining more nodes yet but it seems like it should work.

:+1:

hswong3i:
With my latest founding we could run minikube + flannel without `kubeadm --pod-network-cidr`, see

* https://github.com/alvistack/ansible-role-minikube/blob/afdf7b937a26e3e52e2cc68beab86cb6909022d4/tasks/main.yml#L90
* https://github.com/alvistack/ansible-role-minikube/blob/afdf7b937a26e3e52e2cc68beab86cb6909022d4/molecule/ubuntu-18.04/playbook.yml#L34

From https://github.com/coreos/flannel/blob/master/Documentation/troubleshooting.md#kubernetes-specific what we really needed are kubelet's `--pod-cidr=<cidr>` and controller-manager's `--allocate-node-cidrs=true --cluster-cidr=<cidr>`, so something like this:

```
minikube start \
    --extra-config=controller-manager.allocate-node-cidrs=true \
    --extra-config=controller-manager.cluster-cidr=10.233.64.0/18 \
    --extra-config=kubeadm.ignore-preflight-errors=FileContent--proc-sys-net-bridge-bridge-nf-call-iptables,SystemVerification \
    --extra-config=kubelet.cgroup-driver=systemd \
    --extra-config=kubelet.network-plugin=cni \
    --extra-config=kubelet.pod-cidr=10.233.64.0/18 \
    --extra-config=kubelet.resolv-conf=/run/systemd/resolve/resolv.conf \
    --kubernetes-version=v1.14.1 \
    --network-plugin=cni \
    --vm-driver=none

curl -skL https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml > /tmp/kube-flannel.yml
sed -i 's/10.244.0.0\/16/10.233.64.0\/18/g' /tmp/kube-flannel.yml
kubectl apply -f /tmp/kube-flannel.yml
```

pkg/util/slice.go
```diff
@@ -0,0 +1,74 @@
...
+
+package util
+
+import (
+	"sort"
+)
+
+// CopyStrings copies the contents of the specified string slice
+// into a new slice.
+func CopyStrings(s []string) []string {
```
tstromberg: Rather than adding a function, just use the built-in Go feature for copying slices through append:

`copy := append([]string{}, s...)`

It's admittedly uglier, but less code for us to test.

pkg/util/slice.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil
+	}
+	c := make([]string, len(s))
+	copy(c, s)
+	return c
+}
+
+// SortStrings sorts the specified string slice in place. It returns the same
+// slice that was provided in order to facilitate method chaining.
+func SortStrings(s []string) []string {
```
tstromberg: Just use sort.Strings(s) instead of adding a new function.

pkg/util/slice.go
```diff
@@ -0,0 +1,74 @@
...
+	sort.Strings(s)
+	return s
+}
+
+// ContainsString checks if a given slice of strings contains the provided string.
+// If a modifier func is provided, it is called with the slice item before the comparation.
+func ContainsString(slice []string, s string, modifier func(s string) string) bool {
```
tstromberg: Omit the `modifier` argument - it isn't used by any callers.

pkg/util/slice.go
```diff
@@ -0,0 +1,74 @@
...
+		if modifier != nil && modifier(item) == s {
+			return true
+		}
+	}
+	return false
+}
+
+// RemoveString returns a newly created []string that contains all items from slice that
+// are not equal to s and modifier(s) in case modifier func is provided.
+func RemoveString(slice []string, s string, modifier func(s string) string) []string {
```
tstromberg: Is this function dead code? I don't see a caller, other than tests.

pkg/util/slice_test.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"reflect"
+	"testing"
+)
+
+func TestCopyStrings(t *testing.T) {
+	var src1 []string
+	dest1 := CopyStrings(src1)
+
+	if !reflect.DeepEqual(src1, dest1) {
```
tstromberg: nit: It's fine for this particular case, but if you want diffs or have more advanced structs, consider using `cmp.Equal` and `cmp.Diff`, as some other tests do: https://godoc.org/github.com/google/go-cmp/cmp

pkg/util/slice_test.go
```diff
@@ -0,0 +1,145 @@
...
+		t.Errorf("%v and %v are not equal", src1, dest1)
+	}
+
+	src2 := []string{}
```
tstromberg: Instead of incrementing variable numbers, this should use table-driven subtests: https://blog.golang.org/subtests

See `pkg/minikube/console/console_test.go` for an example.

test/integration/start_stop_delete_test.go
```diff
@@ -56,6 +56,9 @@ func TestStartStop(t *testing.T) {
 			"--extra-config",
 			"kubeadm.ignore-preflight-errors=SystemVerification",
 		}},
+		{"podCidr", []string{
+			"--extra-config=kubeadm.pod-network-cidr=192.168.111.111/16",
```
tstromberg: Since we're already struggling to keep up with the integration test load, please combine this flag into one of the existing tests here. Perhaps `feature_gates_newest_cni`?

pkg/util/slice.go
11janci: I copied this file from the main k8s repo, this function wasn't even used. I removed the whole file and added the one function used (`ContainsString()`) into utils.go

pkg/util/slice.go
11janci: same as prev (file removed)

pkg/util/slice.go
11janci: same as prev (file removed)

pkg/util/slice_test.go
11janci: same as prev (file removed)

pkg/util/slice_test.go
11janci: same as prev (file removed)

test/integration/start_stop_delete_test.go
11janci: done

pkg/util/slice.go
11janci: same as prev (file removed)

11janci:
Hi @tstromberg ,
thx for your review, PR updated, PTAL.
It is working as expected and is ready to ship.

11janci:
there is a unrelated unit test failing in the build (`TestNewSSHClient`), it's failing in master as well. Needs to be resolved first

PR #3896, opened 2019-03-18
Do not wait for k8s-app pods when starting with CNI

brb:
Previously, minikube was waiting for some pods labeled with "k8s-app" (e.g. "kube-dns") to be ready before indicating that "minikube start" was successful. However, in the case of
"network-plugin=cni", network of such pods is managed by a CNI plugin which is usually started after "minikube start" has returned making the waiting for the pods to fail with:

    Error starting cluster: wait: waiting for k8s-app=kube-dns: timed out
    waiting for the condition

This commit omits waiting for "k8s-app" pods if minikube is started with "network-plugin=cni".

Original bug report: https://github.com/cilium/cilium/issues/7383

### DIFF
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -194,7 +194,7 @@ func (k *KubeadmBootstrapper) StartCluster(k8s config.KubernetesConfig) error {
 		}
 	}
 
-	if err := waitForPods(false); err != nil {
+	if err := waitForPods(k8s, false); err != nil {
 		return errors.Wrap(err, "wait")
 	}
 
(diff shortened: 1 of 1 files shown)


tstromberg:
@minikube-bot OK to test

tstromberg:
Excellent change - thank you for cleaning up after my mess. 

So that we don't break this again, could you also add an integration test for CNI support. There is a handy list of 'start' commands we add to:

https://github.com/kubernetes/minikube/blob/master/test/integration/start_stop_delete_test.go#L39

Alternatively - is it viable to wait for CNI first?

tstromberg:
PR looks good. Can merge once the example CNI command-line is added to `start_stop_delete_test.go`.

Thank you again!

brb:
@tstromberg Thanks for reviewing and merging. I will try to add the integration test next week.

PR #3899, opened 2019-03-18
Use Reference, allow caching images with both Tag and Digest

### LINKED ISSUE #3822: "cache add" with sha256 sum: "Failed to cache... repository can only contain the runes"
Right now the minikube cache command allows to cache only  images with tag, it will be great to support caching images with SHAs

afbjorklund:
Closes #3822


tstromberg:
Any chance of adding an integration test to affirm that we don't break this in the future?

tstromberg:
FWIW, It has just come to my attention that `minikube cache` isn't covered by our integration tests. 

😱

afbjorklund:
Sounds like a good idea, for now I just tested it manually and locally.

tstromberg:
@minikube-bot OK to test

tstromberg:
Thanks for the PR! Please merge when ready.

PR #3905, opened 2019-03-19
The rktlet binary is missing, for the rkt container runtime

afbjorklund:
The "rktenetes" integration has been removed from Kubernetes, since version 1.10

https://github.com/kubernetes-incubator/rktlet/blob/master/docs/getting-started-guide.md

I'm not sure if we should keep `rkt`, but right now it is referring to non-existing binaries.

And we might as well update it to the latest versions, if we are going to spend 100M on it ?


tstromberg:
Seems OK - but personally I would rather just remove all vestiges of rkt altogether.

afbjorklund:
Hmm, seems I forgot the .gitignore - added the binary tarball to the repo by mistake :-(

PR #3832, opened 2019-03-08
Skip "pull" command if using Kubernetes 1.10, which does not support it.

tstromberg:
kubeadm v1.10.x doesn't support it. Here's how it gets displayed with this PR:

```
✨  Preparing Kubernetes environment ...
🚜  Pulling images required by Kubernetes v1.10.0 ...
❌  Unable to pull images, which may be OK: pull command is not supported by kubeadm v1.10.0
🔄  Relaunching Kubernetes v1.10.0 using kubeadm ... 
```

Closes #3824

PR #3914, opened 2019-03-20
Fix 'config' flag to kubeadm.

GeekMeUpScotty:
minikube start complains that --config is unsopprted. Proper flag name is --kubeconfig

### DIFF
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -314,7 +314,7 @@ func (k *KubeadmBootstrapper) DeleteCluster(k8s config.KubernetesConfig) error {
 
 // PullImages downloads images that will be used by RestartCluster
 func (k *KubeadmBootstrapper) PullImages(k8s config.KubernetesConfig) error {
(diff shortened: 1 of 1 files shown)


GeekMeUpScotty:
CLA registration done

Den ons. 20. mar. 2019 kl. 13.26 skrev Kubernetes Prow Robot <
notifications@github.com>:

GeekMeUpScotty:
Actually... I am doubtful the flag is the real issue. From what I can read it looks like 'config' _is_ the proper flag, provided kubeadm is recent enough!? So maybe the problem is that the image pulled is incorrect?

tstromberg:
Thank you for this PR! 

I was trying to solve the same issue with Kubernetes v1.10 with #3832 (just skipping "pull"), but I feel like a combination of our two approaches are better. I didn't realize until now that it was just a renamed flag. 

If you don't mind, I'd like to take your idea and incorporate it into #3832 so that it works across all Kubernetes releases.

tstromberg:
So, it turns out that there is no "pull" command on older Kubernetes releases. I've merged #3832 which now simply warns v1.10 users that pull isn't possible, so this PR should be closeable now as well.

tstromberg:
Closing this PR because I don't think it is doing the right thing after all. Feel free to re-open if you disagree. Thank you for bringing this up!

PR #3929, opened 2019-03-22
Adding test to make sure HEAD can run on VM created by latest release

### LINKED ISSUE #816: Add integration test to assert that HEAD can use state created by latest release
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): Feature request

The 0.12.0 to 0.12.1 release required a recreate of the minikube VM because of localkube DNS changes.  This caused a lot of confusion for users upgrading. (#781) 

Minikube should ideally be able to be upgraded without recreating the VM.  We should add a test during our release job to enforce this behavior.

In the rare case that we can't get around recreating the VM, we should make a clear note of it in the release docs.

sharifelgamal:
Closes #816

### DIFF
--- a/test/integration/version_upgrade_test.go
+++ b/test/integration/version_upgrade_test.go
@@ -1,0 +1,88 @@
+/*
+Copyright 2019 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package integration
+
+import (
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"os/exec"
+	"runtime"
+	"testing"
+
+	"github.com/docker/machine/libmachine/state"
+	"github.com/pkg/errors"
+	"k8s.io/minikube/test/integration/util"
+)
+
+func TestVersionUpgrade(t *testing.T) {
+	currentRunner := NewMinikubeRunner(t)
+	currentRunner.RunCommand("delete", true)
+	currentRunner.CheckStatus(state.None.String())
+
+	isoName := "minikube-linux-amd64"
+	if runtime.GOOS == "darwin" {
+		isoName = "minikube-darwin-amd64"
+	} else if runtime.GOOS == "windows" {
+		isoName = "minikube-windows-amd64.exe"
+	}
+	// Grab latest release binary
+	url := "https://storage.googleapis.com/minikube/releases/latest/" + isoName
+	resp, err := http.Get(url)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


test/integration/version_upgrade_test.go
```diff
@@ -0,0 +1,88 @@
...
+	currentRunner.CheckStatus(state.None.String())
+
+	isoName := "minikube-linux-amd64"
+	if runtime.GOOS == "darwin" {
+		isoName = "minikube-darwin-amd64"
+	} else if runtime.GOOS == "windows" {
+		isoName = "minikube-windows-amd64.exe"
+	}
+	// Grab latest release binary
+	url := "https://storage.googleapis.com/minikube/releases/latest/" + isoName
```
tstromberg: See if util.GetBinaryDownloadURL() is applicable for your use case.

test/integration/version_upgrade_test.go
```diff
@@ -0,0 +1,88 @@
...
+	resp, err := http.Get(url)
+	if err != nil {
+		t.Fatal(errors.Wrap(err, "Failed to get latest release binary"))
+	}
+	defer resp.Body.Close()
+
+	iso, err := ioutil.TempFile("", isoName)
```
tstromberg: nit: I'd just name this "tf". "iso" doesn't have any meaning to me.

test/integration/version_upgrade_test.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		t.Fatal(errors.Wrap(err, "Failed to create binary file"))
+	}
+	defer os.Remove(iso.Name())
+
+	_, err = io.Copy(iso, resp.Body)
```
tstromberg: non-actionable note: other parts of the minikube code base use github.com/jimmidyson/go-download, but I'm not really sure of the benefits of doing so.

test/integration/version_upgrade_test.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		t.Fatal(errors.Wrap(err, "Failed to populate iso file"))
+	}
+	if err := iso.Close(); err != nil {
+		t.Fatal(errors.Wrap(err, "Failed to close iso"))
+	}
+
+	if runtime.GOOS != "windows" {
+		if err := exec.Command("chmod", "+x", iso.Name()).Run(); err != nil {
```
tstromberg: Prefer native functions: use os.Chmod instead.

test/integration/version_upgrade_test.go
```diff
@@ -0,0 +1,88 @@
...
+			t.Fatal(errors.Wrap(err, "Failed to make binary executable."))
+		}
+	}
+
+	latestRunner := util.MinikubeRunner{
```
tstromberg: "latest" and "current" is difficult to differentiate between. How about something like "releaseRunner"?

test/integration/version_upgrade_test.go
sharifelgamal: Done.

test/integration/version_upgrade_test.go
sharifelgamal: Done.

test/integration/version_upgrade_test.go
sharifelgamal: Done.

test/integration/version_upgrade_test.go
sharifelgamal: It most certainly is!

tstromberg:
Thanks!

Issue #3145, opened 2018-09-20
Network pre-flights: Detect network issues and present actionable feedback


tstromberg:
Many incoming issues are due to local network access problems, mostly from proxies, but many also due to HyperV:

https://github.com/kubernetes/minikube/labels/failed%2Flocal-networking

We should have a basic connectivity check to the minikube VM, in a way that detects problems in a user-friendly way, such as:

```shell
* Unable to establish a network connection to the VM: please check http://x
  for possible solutions.
```
And at least until we have good story around offline usage:

```
* Unable to establish a fully functional internet connection: please check http://x 
  for possible solutions.
```

tstromberg:
Problems to warn about:

- HTTP client to internet (storage.googleapis.com)
- HTTP client to VM (such as apiserver)
- DNS client from VM to internet
- HTTP client from VM to internet

These failures should point to a connectivity problems document where users can read up on solutions.

PR #3931, opened 2019-03-22
Add solution catalog to help users who run into known problems

tstromberg:
This extracts the nascent problem detection code from exit.go, and formalizes it in into a package that we can use for the upcoming network preflights (#3145). 

Example output:

```
😄  minikube v0.35.0 on darwin (amd64)
🤹  Downloading Kubernetes v1.13.4 images in the background ...
🔥  Creating virtualbox VM (CPUs=2, Memory=2048MB, Disk=20000MB) ...

💣  Unable to start VM
> Error:         [VBOX_NOT_FOUND] create: precreate: VBoxManage not found. Make sure VirtualBox is installed and VBoxManage is in the path
> Solution:      Install VirtualBox, ensure that VBoxManage is executable and in path, or select an alternative value for --vm-driver
> Documentation: https://www.virtualbox.org/wiki/Downloads
> Related issues:
  - https://github.com/kubernetes/minikube/issue/3784
  - https://github.com/kubernetes/minikube/issue/3776

😿  If the solution does not help, please let us know:
```

This also renames exit.WithProblems to exit.WithLogEntries to avoid the terminology mismatch.

### DIFF
--- a/pkg/minikube/problem/err_map.go
+++ b/pkg/minikube/problem/err_map.go
@@ -9,33 +9,33 @@ func r(s string) *regexp.Regexp {
 
 // vmProblems are VM related problems
 var vmProblems = map[string]match{
-	"VBOX_NOT_FOUND": match{
+	"VBOX_NOT_FOUND": {
 		Regexp:   r(`VBoxManage not found. Make sure VirtualBox is installed and VBoxManage is in the path`),
 		Solution: "Install VirtualBox, ensure that VBoxManage is executable and in path, or select an alternative value for --vm-driver",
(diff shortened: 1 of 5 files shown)


pkg/minikube/problem/err_map.go
```diff
@@ -0,0 +1,102 @@
+package problem
+
+import "regexp"
+
+// r is a shortcut around regexp.MustCompile
+func r(s string) *regexp.Regexp {
```
afbjorklund: maybe call this "re()", or perhaps that is just me mixing it with `r''` raw strings (from [python](https://docs.python.org/2/library/re.html)...)

PR #3934, opened 2019-03-22
Clean up lint and other small report card issues

afbjorklund:
Also return an exit code from golint (why is this not default?)

```
pkg/minikube/cluster/mount_test.go:31:37: exported func NewMockMountHost returns unexported type *cluster.mockMountHost, which can be annoying to use
pkg/minikube/constants/constants.go:237:2: don't use underscores in Go names; var ge_v1_14 should be geV1_14

docs/contributors/build_guide.md: 60: MD001/heading-increment/header-increment Heading levels should only increment by one level at a time [Expected: h3; Actual: h4]
docs/drivers.md: 39: MD009/no-trailing-spaces Trailing spaces [Expected: 0 or 2; Actual: 1]

pkg/minikube/tests/provision_mock.go:87:20: "attemps" is a misspelling of "attempts"
```

### DIFF
--- a/pkg/minikube/cluster/mount_test.go
+++ b/pkg/minikube/cluster/mount_test.go
@@ -28,7 +28,7 @@ type mockMountHost struct {
 	T    *testing.T
 }
 
-func NewMockMountHost(t *testing.T) *mockMountHost {
+func newMockMountHost(t *testing.T) *mockMountHost {
 	return &mockMountHost{
 		T:    t,
 		cmds: []string{},
@@ -86,7 +86,7 @@ func TestMount(t *testing.T) {
 	}
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			h := NewMockMountHost(t)
+			h := newMockMountHost(t)
 			err := Mount(h, tc.source, tc.target, tc.cfg)
 			if err != nil {
 				t.Fatalf("Mount(%s, %s, %+v): %v", tc.source, tc.target, tc.cfg, err)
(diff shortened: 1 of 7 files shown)


pkg/minikube/cluster/mount_test.go
```diff
@@ -23,19 +23,20 @@ import (
...
-type mockMountHost struct {
+// MockMountHost is for testing
+type MockMountHost struct {
 	cmds []string
 	T    *testing.T
 }
 
-func NewMockMountHost(t *testing.T) *mockMountHost {
-	return &mockMountHost{
+func NewMockMountHost(t *testing.T) *MockMountHost {
```
tstromberg: Make this function name private (newMockMountHost) - then the underlying struct can be private too.

There's no need to pollute the public namespace with this minimal implementation.

pkg/minikube/constants/constants.go
```diff
@@ -234,7 +234,7 @@ func GetKubeadmCachedImages(kubernetesVersionStr string) []string {
 		"k8s.gcr.io/kube-apiserver-amd64:" + kubernetesVersionStr,
 	}
 
-	ge_v1_14 := semver.MustParseRange(">=1.14.0")
+	v1_14 := semver.MustParseRange(">=1.14.0")
```
tstromberg: I feel like this name isn't as meaningful any more, but understand why lint picked this up.

Perhaps "newerThan114" or "greaterThan114". "ge114" is probably a valid token too.

Alternatively, give this a max range?

tstromberg:
Minor changes requested. 

/lgtm /approve

pkg/minikube/constants/constants.go
afbjorklund: I improvised. Most likely this whole section should be rewritten anyway ?

pkg/minikube/cluster/mount_test.go
afbjorklund: Right

Issue #3915, opened 2019-03-20
Patching nginx controller deployment results in rollback to former config


leifcr:
Create new minikube cluster
minikube start

Enable ingress addon

```minikube addons enable ingress```

Patch option for args

```
kubectl patch deployment nginx-ingress-controller --type 'json' --namespace kube-system -p '[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--default-ssl-certificate=kube-system/test-tls"}]'

deployment.extensions/nginx-ingress-controller patched
```

Same for setting healthz-port

```
kubectl patch deployment nginx-ingress-controller --type 'json' --namespace kube-system -p '[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--healthz-port 5533"}]'

deployment.extensions/nginx-ingress-controller patched
```

After patching successfuly, configuration is rolled back, and argument is removed, and deployment 'reset' back to former state.

I expected the patch to be persisted. I'm not seeing this on a staging cluster, so I'm trying to figure out if this is minikube the addon that enforces configuration?

**Comment by leifcr** [comment]
Due to minikube running only 1 node, and the ingress using the hostPort, rolling updates will not work for the ingress deployment, as there is no 'available port' when the new replicaset is rolled out. I had to trigger an update + pause deployment, to have time to see the error message.

After patching the ingress to use recreate instead, patching the ingress config works as expected.

Command to set the ingress controller to 'recreate':

```
kubectl patch deployment nginx-ingress-controller --type 'json' --namespace kube-system -p '[{"op": "replace", "path": "/spec/strategy/type", "value": "Recreate"}, {"op": "replace", "path": "/spec/strategy/rollingUpdate", "value": null }]'
```


Command to set debug output logging on the nginx-ingress-controller:

```
kubectl patch deployment nginx-ingress-controller --type 'json' --namespace kube-system -p '[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "-v=5"}]'
```

The ingress controller now has debug log output, and is set to recreate if the config or image changes.

I do think this can be closed, and just be used as a how-to.

leifcr:
For others that might need to change the ingress controller configuration, the options are:
1. Copy the ingress configuration files top .minikube/addons and modify there.
2. Roll out using helm
3. Roll out manually

PR #3935, opened 2019-03-22
Use recreate rollout strategy on ingress controller

leifcr:
Due to issue #3915, I think the recreate strategy fits better in minikube. For a production cluster this shouldn't apply as there should be 'free' space to rollout the nginx ingress controller over several nodes.

### DIFF
--- a/deploy/addons/ingress/ingress-dp.yaml
+++ b/deploy/addons/ingress/ingress-dp.yaml
@@ -69,6 +69,8 @@ metadata:
     addonmanager.kubernetes.io/mode: Reconcile
 spec:
   replicas: 1
+  strategy:
+    type: Recreate
   selector:
     matchLabels:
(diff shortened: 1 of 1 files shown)


leifcr:
It looks like this doesn't help when changing the ingress parameters, as even with recreate, the ports are bound before the new pod is started. Please await merging this until proper fix for changing the ingress parameters has been found/figured out.

tstromberg:
@minikube-bot OK to test

leifcr:
It seems like the addon-manager rolls back to it's known working settings. @tstromberg Can you confirm? If so, this PR should be cancelled, as #3915 won't be fixable, and the best option for configurable ingress would be to not use the addon, and rather use manual deployment or a helm chart.

leifcr:
Closing this, as the only way to change configuration on the ingress is to roll out the ingress manually, through helm or by writing your own addon.

PR #3937, opened 2019-03-23
Select an accessible image repository for some users

laozc:
Now it's possible to fallback to a mirror site for those users
who cannot access the main k8s.gcr.io repository. You may also
set the --image-mirror-country to forcibly use a known local
mirror in the specific location.


tstromberg:
@minikube-bot OK to test

tstromberg:
Looks good so far, though I'll need to give this a more careful review. Thank you!

cmd/minikube/cmd/start.go
```diff
@@ -266,6 +293,60 @@ func showKubectlConnectInfo(kubeconfig *pkgutil.KubeConfigSetup) {
...
+		_, ok := repos[mirrorCountry]
+		if !ok {
+			return false, "", fmt.Errorf("invalid image mirror country code: %s", mirrorCountry)
+		}
+		countries = []string{mirrorCountry, ""}
+
+	} else {
+		// make sure global is preferred
+		countries = []string{""}
+		for k := range repos {
```
tstromberg: I really like this, but I don't feel we should implement alternative mirrors by default until we have a way of validating the contents of the images (#4104). My concern is the case where a business blocks gcr.io and other known image repositories, and we fall back to a 3rd party poisoned repository for them, and deliver them a backdoored image.

In the short term, I don't mind if we have country fallbacks if we have a reasonable suspicion that the user is in that country - but I hesitate to apply it globally until #4104 can be fixed.

pkg/minikube/constants/constants.go
```diff
@@ -211,6 +211,12 @@ const (
 	DefaultMountVersion = "9p2000.L"
 )
 
+// ImageRepositories contains all known image repositories
+var ImageRepositories = map[string][]string{
+	"":   {""}, // global
```
tstromberg: Name the key "global" so that it is self-documenting.

tstromberg:
Since I like this change overall and want to see it merged, what do you think about supporting a flag, such as `--image-repository=auto` to trigger this logic? At least until we can figure out a way to either:

- Only auto-select mirrors for the country the user appears to be in
- Or #4104 is resolved, where mirror auto-selection is guaranteed to be safe

cmd/minikube/cmd/start.go
```diff
@@ -176,6 +181,28 @@ func runStart(cmd *cobra.Command, args []string) {
...
+	if config.KubernetesConfig.ImageRepository == "" {
+		console.OutStyle("connectivity", "checking main repository and mirrors for images")
+		found, repository, err := selectImageRepository(config)
+		if err != nil {
+			exit.WithError("Failed to check main repository and mirrors for images for images", err)
+		}
+
+		if !found {
+			if repository == "" {
+				exit.WithCode(exit.Failure, "None of known repositories is accessible. Consider specifying an alternative image repository with --image-repository flag")
```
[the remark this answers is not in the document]
laozc: Sure. I'll verify the if the offline mode is still working as expected.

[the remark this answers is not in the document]
laozc: Right now it works with cached data as long as the image repository does not change.
Especially in the latest commit this logic only triggers when --image-repository is set to auto, or --image-mirror-country is set, so I believe for most cases it should work as expected.

However, when minikube caches remote content, it takes image repository name + tag not SHA-256. So the local cache may not work with the case when two identical images from different repository. I think this could be addressed along with #4104.

cmd/minikube/cmd/start.go
laozc: This logic is only triggered manually by setting --image-repository to auto or --image-mirror-country is set before #4104 is fixed.

tstromberg:
Merging! I'm hoping to make "auto" the default once #4104 can be addressed. Thank you so much for your hard work here.

PR #3943, opened 2019-03-23
Fix infinite loop when failing on caching

laozc:
waitCacheImages is called after prepareHostEnvironment at the
moment, which may cause
 1. Incompleted image file to be loaded if downloading speed is
    not fast enough
 2. Infinite loop when image file is not cached on the disk due to
    failures as minikube keeps checking the image file

This moves to waitCacheImages ahead of prepareHostEnvironment to
ensure at least image caching completes (either succeeds or fails)
when trying to load the cached images inside the VM.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -199,8 +199,8 @@ func runStart(cmd *cobra.Command, args []string) {
 	}
 
 	cr := configureRuntimes(host, runner)
-	bs := prepareHostEnvironment(m, config.KubernetesConfig)
 	waitCacheImages(&cacheGroup)
+	bs := prepareHostEnvironment(m, config.KubernetesConfig)
 
 	// The kube config must be update must come before bootstrapping, otherwise health checks may use a stale IP
 	kubeconfig := updateKubeConfig(host, &config)
--- a/pkg/minikube/machine/cache_images.go
+++ b/pkg/minikube/machine/cache_images.go
@@ -202,10 +202,8 @@ func getWindowsVolumeNameCmd(d string) (string, error) {
 func LoadFromCacheBlocking(cr bootstrapper.CommandRunner, k8s config.KubernetesConfig, src string) error {
 	glog.Infoln("Loading image from cache at ", src)
 	filename := filepath.Base(src)
-	for {
-		if _, err := os.Stat(src); err == nil {
-			break
-		}
+	if _, err := os.Stat(src); err != nil {
+		return err
 	}
 	dst := path.Join(tempLoadDir, filename)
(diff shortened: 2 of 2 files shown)


tstromberg:
@minikube-bot OK to test

tstromberg:
I'm curious if https://github.com/kubernetes/minikube/pull/3961 fixes the underlying issue here.

I'm not comfortable with this PR yet, since it changes the behavior of LoadFromCacheBlocking() so that the name and intent is no longer applicable. I do not like infinite loops, but I not yet convinced that this PR is doing the correct thing.

I'd rather that every loop have an exit condition.

laozc:
Thanks for reviewing.

The lines calling os.Stat in the loop assumes that will be a local downloaded copy won't work for the case when it fails to connect or read from the remote repository.

LoadFromCacheBlocking doesn't sync the status with image pulling. So it won't distinguish those cases when the pulling is not even started or actually failed. Therefore an infinite loop would take place when the pull fails to produce a local copy. This can be easily reproduced by specifying a wrong repository prefix (like _registry.cn-hangzhou.aliyuncs.com/google_containers1_). Somehow  transient networking problem may do the same thing.

Unless we introduce some sort of sync mechanism between the image pulling and LoadFromCacheBlocking (such as passing cacheGroup to Bootstrap), it's kind of dangerous as it could drain out the CPU quickly while the user knows nothing about what's actually happening due to an infinite without no prompt.

🐳  Version of container runtime is 18.06.2-ce
✨  Preparing Kubernetes environment ...
    ▪ kubelet.volume-plugin-dir=/etc/kubernetes/volumeplugins
(...with long wait and fan noise...)

It is not a perfect fix. Rather than having the user experiencing an infinite loop, it might be better to just break the loop with a small change even the name is mismatch slightly with the behavior for a temporary solution.

I could dig into the codes to fix the code in a more elegant way 😄 if time permits.

tstromberg:
> Unless we introduce some sort of sync mechanism between the image pulling and LoadFromCacheBlocking (such as passing cacheGroup to Bootstrap), it's kind of dangerous as it could drain out the CPU quickly while the user knows nothing about what's actually happening due to an infinite loop with no prompt.

I did not realize this until now, but you are completely right, and I also think that passing a sync mechanism is the correct answer here. That's a great idea. The image caching is already using a sync.ErrGroup object.

Instead of this infinite loop, we should just call .Wait() on this err group object, and then we can safely use this as a non-blocking call as you have implemented here.

I feel that this change without the Wait() call isn't ideal, because it introduces a race condition in the normal case, and may cause users to download the ISO twice simultaneously: one outside the VM, and one inside the VM.

If you called Wait() and renamed the function to not include the word Blocking, I'd approve this PR.

laozc:
Thanks for taking the work.

PR #3961, opened 2019-03-25
Use a temp file, when writing the cached image

### LINKED ISSUE #3960: When interrupting a download, the cache image gets corrupted
> ❌  Unable to load cached images: loading cached images: loading image /home/anders/.minikube/cache/images/k8s.gcr.io/k8s-dns-kube-dns-amd64_1.14.13: Docker load /tmp/k8s-dns-kube-dns-amd64_1.14.13: command failed: docker load -i /tmp/k8s-dns-kube-dns-amd64_1.14.13
stdout: 
stderr: Error processing tar file(exit status 1): unexpected EOF

It works anyway (downloading the image again), but the output is ugly enough.

afbjorklund:
Closes #3960

### DIFF
--- a/pkg/minikube/machine/cache_images.go
+++ b/pkg/minikube/machine/cache_images.go
@@ -314,10 +314,21 @@ func CacheImage(image, dst string) error {
 	}
 
 	glog.Infoln("OPENING: ", dstPath)
-	f, err := os.Create(dstPath)
+	f, err := ioutil.TempFile(filepath.Dir(dstPath), filepath.Base(dstPath)+".*.tmp")
 	if err != nil {
 		return err
 	}
-	defer f.Close()
-	return tarball.Write(tag, img, nil, f)
+	err = tarball.Write(tag, img, nil, f)
+	if err != nil {
+		return err
+	}
+	err = f.Close()
+	if err != nil {
+		return err
+	}
+	err = os.Rename(f.Name(), dstPath)
+	if err != nil {
+		return err
+	}
+	return nil
 }


pkg/minikube/machine/cache_images.go
```diff
@@ -314,10 +314,19 @@ func CacheImage(image, dst string) error {
...
 	if err != nil {
 		return err
 	}
-	defer f.Close()
-	return tarball.Write(tag, img, nil, f)
+	err = tarball.Write(tag, img, nil, f)
+	if err != nil {
+		return err
+	}
+	f.Close()
```
tstromberg: Please check the error code to close!

pkg/minikube/machine/cache_images.go
```diff
@@ -314,10 +314,19 @@ func CacheImage(image, dst string) error {
 	}
 
 	glog.Infoln("OPENING: ", dstPath)
-	f, err := os.Create(dstPath)
+	tmpPath := dstPath + ".tmp"
+	f, err := os.Create(tmpPath)
```
tstromberg: Will this fail if dstPath.tmp exists already? 

I'd personally rather see ioutil.TempFile used here, passing in this same directory as an argument.

afbjorklund: It should truncate, but we can use random file names

pkg/minikube/machine/cache_images.go
afbjorklund: I think we did it with `defer` before... WIll check

Issue #3511, opened 2019-01-08
none on Ubuntu should automatically set --extra-config=kubelet.resolv-conf


tlkh:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):

**BUG REPORT**

Please provide the following details:

**Environment**: Ubuntu 18.04, fresh install

**Minikube version** (use `minikube version`): v0.32.0
- **OS** (e.g. from /etc/os-release): Ubuntu 18.04.1 LTS
- **VM Driver** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep DriverName`): None
- **ISO version** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep -i ISO` or `minikube ssh cat /etc/VERSION`): ?
- **Install tools**: ?
- **Others**: NIL
The above can be generated in one go with the following commands (can be copied and pasted directly into your terminal):
```
minikube version
echo "";
echo "OS:";
cat /etc/os-release
echo "";
echo "VM driver:"; 
grep DriverName ~/.minikube/machines/minikube/config.json
echo "";
echo "ISO version";
grep -i ISO ~/.minikube/machines/minikube/config.json
```

**What happened**:

CoreDNS CrashLoopBackOff. Log shows:

```
[FATAL] plugin/loop: Seen "HINFO IN xxxxxxxxx." more than twice, loop detected
```

Related issue in CoreDNS: https://github.com/coredns/coredns/issues/2087

**What you expected to happen**:

Expected it to work!

**How to reproduce it** (as minimally and precisely as possible):

Deploy minikube on Ubuntu 18.04 with "None" driver

**Output of `minikube logs` (if applicable)**:

**Anything else do we need to know**:

**Solution**:

Add instructions to disable systemd-resolved and use dnsmaq. This worked for me.

```
sudo apt-get install dnsmasq
sudo systemctl stop systemd-resolved
sudo systemctl disable systemd-resolved

sudo nano /etc/NetworkManager/NetworkManager.conf
# add under [main]
# dns=dnsmasq

sudo cp /etc/resolv.conf /etc/resolv.conf.bak
sudo rm /etc/resolv.conf; sudo ln -s /var/run/NetworkManager/resolv.conf /etc/resolv.conf

sudo systemctl start dnsmasq
sudo systemctl restart NetworkManager
```

tstromberg:
[...]

PR #3979, opened 2019-03-26
Doc update: Issue with ubuntu 18.04, systemd-resolve dns and minikube

marcosdiez:
I am basically documenting the solution for https://github.com/kubernetes/minikube/issues/3511 properly.

It stole my whole morning and I don't want anybody else to loose so much time.

If and only if I have time, I might write a proper patch for minikube in the weekend. 

For now, this will do.

Above all, thank you @andrewjjenkins


tstromberg:
Thanks for adding this! My apologies for the time spent :(

We should also see about getting this doc cited when the error occurs, by adding it here:
 https://github.com/kubernetes/minikube/blob/f74751c3b702c6b193cc80183509849ef9ecd829/pkg/minikube/problem/err_map.go#L134

Issue #3972, opened 2019-03-26
Unable to enable dashboard on k8s v1.14.0 (addon-manager misbehaving?)


LarsKumbier:
Current master seems to be unable to enable the dashboard addon with kubernetes v1.14.0.

## Reproduce steps:
* build minikube from master (commit tested: f74751c3b702c6b193cc80183509849ef9ecd829)
* provision a cluster: `minikube start --vm-driver=kvm2 --kubernetes-version v1.14.0`
* try to enable the dashboard: `minimube dashboard --v=7`

```
$ minikube dashboard --v=7
Found binary path at /usr/local/bin/docker-machine-driver-kvm2
Launching plugin server for driver kvm2
Plugin server listening at address 127.0.0.1:41149
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(minikube) Calling .GetState
-   Enabling dashboard ...
Found binary path at /usr/local/bin/docker-machine-driver-kvm2
Launching plugin server for driver kvm2
Plugin server listening at address 127.0.0.1:37685
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(minikube) Calling .GetState
Found binary path at /usr/local/bin/docker-machine-driver-kvm2
Launching plugin server for driver kvm2
Plugin server listening at address 127.0.0.1:40377
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(minikube) Calling .GetSSHHostname
(minikube) Calling .GetSSHPort
(minikube) Calling .GetSSHKeyPath
(minikube) Calling .GetSSHUsername
Making call to close driver server
(minikube) Calling .Close
Successfully made call to close driver server
Making call to close connection to plugin binary
Making call to close driver server
(minikube) Calling .Close
Successfully made call to close driver server
Making call to close connection to plugin binary
-   Verifying dashboard health ... Temporary Error: Error getting service kubernetes-dashboard: services "kubernetes-dashboard" not found
Temporary Error: Error getting service kubernetes-dashboard: services "kubernetes-dashboard" not found
Temporary Error: Error getting service kubernetes-dashboard: services "kubernetes-dashboard" not found                                                                                                      
[...last message repeated a few dozen times...]
```
## Minikube logs
[minikube.log](https://github.com/kubernetes/minikube/files/3007976/minikube.log)

## Versions
* Ubuntu 18.10
[...]

PR #3982, opened 2019-03-26
Add addon-manager, dashboard, and storage-provisioner to minikube logs

tstromberg:
Helpful for debugging issues like #3972 

This also opens the filter up slightly so that we can see non-system pods, such as the dashboard.


tstromberg:
I can also confirm that the command silently doesn't display logs if the pods are not found.

PR #3985, opened 2019-03-26
More reliable unmount w/ SIGINT, particularly on kvm2

tstromberg:
With kvm2, when using mount and hitting Ctrl-C, the unmount fails with:

`read tcp 127.0.0.1:40862->127.0.0.1:37853: read: connection reset by peer`

If you try again, it fails with:

`connection is shut down`

This is because the native Docker ssh implementation dies when SIGINT is encountered. This switches the implementation to use our CommandRunner over ssh, as well as adding a second attempt using "umount -f".

Additionally, I noticed that "findmnt -T /tmp/777" would show the tmpfs mount, so there is now a `grep` command added in.


pkg/minikube/cluster/mount.go
```diff
@@ -100,9 +103,31 @@ func mntCmd(source string, target string, c *MountConfig) string {
 	return fmt.Sprintf("sudo mount -t %s -o %s %s %s", c.Type, strings.Join(opts, ","), source, target)
 }
 
+// umountCmd returns a command for unmounting
+func umountCmd(target string, force bool) string {
```
balopat:
```suggestion
func unmountCmd(target string, force bool) string {
```

balopat: nevermind - it's actually valid as umount as well :)

PR #3990, opened 2019-03-26
Block on cache downloads before using the results

tstromberg:
@laozc brought up some very good points in #3943. This PR just takes the idea and clarifies it.

- Previously, we called prepareHostEnvironment before waitCacheImages. This is bad because prepareHostEnvironment was calling LoadImages!

- This removes the unnecessary infinite loop, because the Wait() call implements this already, and without the high CPU load.

- This renames `LoadFromCacheBlocking` to a more appropriate private function: `loadImageFromCache`


tstromberg:
@minikube-bot OK to test

PR #3997, opened 2019-03-27
Escape systemd special chars in docker-env

### LINKED ISSUE #3812: Proxy settings by --docker-env http_proxy fails with percent encoded chars
If this is a bug report, please include:

- [x] How to replicate the error, including the exact command-lines used.
- [ ] The full output of the command that failed
- [x] The operating system name and version used

11janci:
The --docker-env param is interpolated into systemd unit file for docker engine. Any special characters that systemd provides for use in templates will be interpreted by systemd. Since minikube does not support these special characters in docker-env, we must escape them so that systemd does not interpret them as special chars. 
See examples in the below issue.

Note: this PR deals only with the docker-env param which gets inserted into the Environment option of the systemd unit file. Other parts of the file are omitted.

Closes #3812

### DIFF
--- a/pkg/provision/buildroot.go
+++ b/pkg/provision/buildroot.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"path"
 	"path/filepath"
+	"strings"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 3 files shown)


afbjorklund:
/lgtm /approve

Looks good, thanks for the analysis and test cases!

PR #4026, opened 2019-03-31
Escape ‘%’ in console.OutStyle arguments

### LINKED ISSUE #4025: console.OutStyle() arguments escaping
Hi,
if you send an argument to console.OutStyle() that contains a '%' sign, the sign will get interpreted as a formatting directive instead of being printed out.
Example:
```console.OutStyle("option", "%s", "HTTP_PROXY=http://user%40example.com:password@proxy.com:8080")```
will print out
```HTTP_PROXY=http://user%!e(MISSING)xample.com:password@proxy.com:8080```
instead of
```HTTP_PROXY=http://user%40example.com:password@proxy.com:8080```

### PR DESCRIPTION
**Pull request by 11janci**
Escape arguments to console.OutStyle so that '%' signs get printed out instead of being interpreted as formatting directives and printed out like ```%!e(MISSING)```
Closes #4025

### DIFF
--- a/pkg/minikube/console/console.go
+++ b/pkg/minikube/console/console.go
@@ -70,7 +70,7 @@ func HasStyle(style string) bool {
 
 // OutStyle writes a stylized and formatted message to stdout
 func OutStyle(style, format string, a ...interface{}) error {
-	OutStyle, err := applyStyle(style, useColor, fmt.Sprintf(format, a...))
+	outStyled, err := applyStyle(style, useColor, format, a...)
 	if err != nil {
 		glog.Errorf("applyStyle(%s): %v", style, err)
 		if oerr := OutLn(format, a...); oerr != nil {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by tstromberg** [comment]
Agreed. This will definitely need an addition to console_test.go.

**Comment by tstromberg** [comment]
@minikube-bot OK to test

File: pkg/minikube/console/console.go [hunk 1]
```diff
@@ -70,15 +70,19 @@ func HasStyle(style string) bool {
 
 // OutStyle writes a stylized and formatted message to stdout
 func OutStyle(style, format string, a ...interface{}) error {
-	OutStyle, err := applyStyle(style, useColor, fmt.Sprintf(format, a...))
+	outStyled, err := applyStyle(style, useColor, format, a...)
+	// format the output string and escape any outstanding '%' signs so that they don't
+	// get interpreted as a formatting directive down the line
+	outStyled = strings.Replace(outStyled, "%", "%%", -1)
```
tstromberg: This should probably use `%` instead of "%"

11janci:
@tstromberg @afbjorklund done, thanks for pointing to the test, now I found it was also broken for ErrStyle().

/ptal

pkg/minikube/console/console.go
11janci: I think this is just a github thing, neither IDE nor gofmt complains

PR #4030, opened 2019-04-01
Update Hyper-V daemons

laozc:
Currently only KVP daemons are complied and installed into the minikube ISO.
There are a few more Hyper-V daemons for integration services needed like Fcopy and VSS.
This PR includes new configuration to enable these daemons in minikube.

* Guest service is not enabled by default. Once it's checked in Hyper-V setting, the Fcopy daemon will be brought up by systemd, in which case you may leverage FCopy cmdlet to copy file into the VM


tstromberg:
@minikube-bot OK to test

tstromberg:
Thank you!

Issue #4057, opened 2019-04-04
Verifying component health always fails with non-default --apiserver-port


moduspwnens:
System: macOS 10.14.4
minikube version: 1.0.0
Command: minikube start --apiserver-port=6443

When I do this, it always fails in the "Verifying component health" stage. However, kubectl works as expected and the cluster seems to be in good health.

I'm not super familiar with golang, but it appears that the function being called to verify the apiserver's health is here:
https://github.com/kubernetes/minikube/blob/58a704521a9faf9fe496ccdaed2ac529c1693e58/cmd/minikube/cmd/start.go#L561

It's calling a "GetAPIServerStatus" method and passing in only the IP address of the apiserver. The function itself is here:

https://github.com/kubernetes/minikube/blob/89ab0be3b0c7a60b96793128850a98441966b96f/pkg/minikube/bootstrapper/kubeadm/kubeadm.go#L126

And for the port to use, it looks like it may be referencing a constant declared here as 8443:
https://github.com/kubernetes/minikube/blob/8ba53c6b0b6487e212a86a07f58a6b43ee8cf6e5/pkg/util/constants.go#L27

Presumably it should only reference the constant if the user didn't specify the API server port. Although again, my experience with golang is small.

moduspwnens:
$ minikube start --apiserver-port=6443
😄  minikube v1.0.0 on darwin (amd64)
🤹  Downloading Kubernetes v1.14.0 images in the background ...
🔥  Creating virtualbox VM (CPUs=2, Memory=2048MB, Disk=20000MB) ...
📶  "minikube" IP address is 192.168.99.165
🐳  Configuring Docker as the container runtime ...
🐳  Version of container runtime is 18.06.2-ce
⌛  Waiting for image downloads to complete ...
✨  Preparing Kubernetes environment ...
🚜  Pulling images required by Kubernetes v1.14.0 ...
🚀  Launching Kubernetes v1.14.0 using kubeadm ... 
⌛  Waiting for pods: apiserver proxy etcd scheduler controller dns
🔑  Configuring cluster permissions ...
🤔  Verifying component health ..................................
💣  apiserver checks failed: Temporary Error: apiserver status=Stopped err=<nil>
Temporary Error: apiserver status=Stopped err=<nil>
Temporary Error: apiserver status=Stopped err=<nil>
Temporary Error: apiserver status=Stopped err=<nil>
Temporary Error: apiserver status=Stopped err=<nil>
[...]

PR #4058, opened 2019-04-05
Fixed status checking with non-default apiserver-port.

moduspwnens:
This is a fix for #4057.

Component health checking as part of "start" and "status" now respect non-default apiserver port settings from the command line and kubeconfig file, respectively. The functionality is now verified with an integration test.

### DIFF
--- a/test/integration/start_stop_delete_test.go
+++ b/test/integration/start_stop_delete_test.go
@@ -46,17 +46,15 @@ func TestStartStop(t *testing.T) {
 			"--extra-config=kubelet.network-plugin=cni",
 			fmt.Sprintf("--kubernetes-version=%s", constants.NewestKubernetesVersion),
 		}},
-		{"containerd", []string{
+		{"containerd_and_non_default_apiserver_port", []string{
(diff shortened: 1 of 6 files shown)


test/integration/start_stop_delete_test.go
```diff
@@ -55,6 +55,9 @@ func TestStartStop(t *testing.T) {
 			"--extra-config",
 			"kubeadm.ignore-preflight-errors=SystemVerification",
 		}},
+		{"apiserver_port_non_default", []string{
+			"--apiserver-port=8444",
```
tstromberg: Do you mind combining this with another test entry - such as "containerd"? 

This is less than ideal, but each entry adds about 10 minutes to our integration tests.

tstromberg:
@minikube-bot OK to test

test/integration/start_stop_delete_test.go
moduspwnens: No problem. I've now pushed a new commit that combines the two tests.

tstromberg:
Looks great. Thank you!

Issue #2880, opened 2018-06-07
Add support for linux s390x


salamani:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
FEATURE REQUEST

Please provide the following details:

**Environment**:
Linux s390x

**Minikube version** (use `minikube version`): v0.27.0
- **OS** (e.g. from /etc/os-release):
NAME="Ubuntu"
VERSION="16.04.4 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.4 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial

- **VM Driver** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep DriverName`): none
- **ISO version** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep -i ISO` or `minikube ssh cat /etc/VERSION`): -
- **Install tools**:
- **Others**:
The above can be generated in one go with the following commands (can be copied and pasted directly into your terminal):
```
minikube version
echo "";
echo "OS:";
cat /etc/os-release
echo "";
echo "VM driver": 
grep DriverName ~/.minikube/machines/minikube/config.json
echo "";
echo "ISO version";
grep -i ISO ~/.minikube/machines/minikube/config.json
```

**What happened**: 
minikube-linux-s390x binary is not available.

**What you expected to happen**:
Add support to release minikube-linux-s390x binary for linux/s390x.
kubernetes and kubeadm has support for linux/s390x.
Willing to work and provide linux/s390x support for minikube.
Checked all testcases(./test.sh) and integration-non-driver. All are successful.

**How to reproduce it** (as minimally and precisely as possible):

**Output of `minikube logs` (if applicable)**:

**Anything else do we need to know**:

tstromberg:
salamani: The first step is confirming that minikube works on s390x. I don't have access to an s390 host, so can you confirm what minikube's working state is on the platform? I'd be curious what VM driver you are using.

Due to our current resources, supporting s390x, or any non x86-64 platform, will need to come from the community. It will probably need to support a QEMU or some other x86-64 hypervisor, or perhaps ship an alternative set of ISO's.

[...]

PR #4091, opened 2019-04-13
Adding support for s390x

salamani:
PR for #2880 and recreating #2881 due to multiple conflicts in rebase

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,16 @@ _testmain.go
 /out
 /_gopath
 
+
+#.yaml files specific to deploy's .template extensions
+deploy/addons/addon-manager.yaml                           
+deploy/addons/dashboard/dashboard-dp.yaml                  
+deploy/addons/heapster/heapster-rc.yaml                    
+deploy/addons/heapster/influx-grafana-rc.yaml              
+deploy/addons/ingress/ingress-dp.yaml                      
+deploy/addons/metrics-server/metrics-server-deployment.yaml
+deploy/addons/storage-provisioner/storage-provisioner.yaml 
+
 #iso version file
 deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/etc/VERSION
 
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,25 @@ endif
 ifeq ($(GOOS),windows)
 	IS_EXE = ".exe"
 endif
+
+ifeq ($(GOARCH),amd64)
+	ARCHTAG ?= -amd64
+	ARCHTAG_NONE ?=
+else
+	ARCHTAG ?= -$(GOARCH)
+	ARCHTAG_NONE ?= -$(GOARCH)
+endif
+
+.PHONY: makedeploys
+makedeploys:
+	sed "s|\-ARCHTAG_NONE|$(ARCHTAG_NONE)|g" deploy/addons/addon-manager.template > deploy/addons/addon-manager.yaml
+	sed "s|\-ARCHTAG|$(ARCHTAG)|g" deploy/addons/dashboard/dashboard-dp.template > deploy/addons/dashboard/dashboard-dp.yaml
+	sed "s|\-ARCHTAG|$(ARCHTAG)|g" deploy/addons/heapster/heapster-rc.template > deploy/addons/heapster/heapster-rc.yaml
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -28,6 +28,14 @@ _testmain.go
 /out
 /_gopath
 
+deploy/addons/addon-manager.yaml
```
tstromberg: This feels weird. It needs a comment, and a wildcard so that it doesn't have to be manually maintained every time a new addon is created.

Makefile
```diff
@@ -99,6 +99,25 @@ endif
...
+	IMAGETAG ?= -amd64
+	IMAGETAG_NONE ?=
+else
+	IMAGETAG ?= -$(GOARCH)
+	IMAGETAG_NONE ?= -$(GOARCH)
+endif
+
+.PHONY: makedeploys
+makedeploys:
+	sed "s|\-IMAGETAG_NONE|$(IMAGETAG_NONE)|g" deploy/addons/addon-manager.template > deploy/addons/addon-manager.yaml
```
tstromberg: This adds a new dependency for Windows builders ("sed"), but I'm not really sure what the alternative is here.

deploy/addons/dashboard/dashboard-dp.template
```diff
@@ -37,7 +37,7 @@ spec:
     spec:
       containers:
       - name: kubernetes-dashboard
-        image: {{default "k8s.gcr.io" .ImageRepository}}/kubernetes-dashboard-amd64:v1.10.1
+        image: {{default "k8s.gcr.io" .ImageRepository}}/kubernetes-dashboard-IMAGETAG:v1.10.1
```
tstromberg: The term "tag" is already used for what comes after the colon (v1.10.1).

What do you think about something like ARCH or PLATFORM?

deploy/addons/storage-provisioner/storage-provisioner.template
```diff
@@ -51,7 +51,7 @@ spec:
   hostNetwork: true
   containers:
   - name: storage-provisioner
-    image: {{default "gcr.io/k8s-minikube" .ImageRepository}}/storage-provisioner:v1.8.1
+    image: {{default "gcr.io/k8s-minikube" .ImageRepository}}/storage-provisioner-IMAGETAG_NONE:v1.8.1
```
tstromberg: Why is it necessary to add -IMAGETAG_NONE here? Can't this image name just be left alone?

pkg/minikube/constants/constants.go
```diff
@@ -51,6 +51,15 @@ func GetMinipath() string {
 	return filepath.Join(os.Getenv(MinikubeHome), ".minikube")
 }
 
+// SupportedArchTag retuns the archtag for images
```
tstromberg: How about naming this function just ArchTag?

Also, fix spelling on "returns".

.gitignore
salamani: I added the comments now. I don't see any approach to have a wid card to ignore these files.

Makefile
salamani: Seems to be windows already has this dependency. I can see other sed commands in Makefile

deploy/addons/storage-provisioner/storage-provisioner.template
salamani: Some images used in templates does not have a amd64 archTag
Example: `gcr.io/k8s-minikube/storage-provisioner-IMAGETAG_NONE:v1.8.1` will become `gcr.io/k8s-minikube/storage-provisioner:v1.8.1`

pkg/minikube/constants/constants.go
salamani: YES. I agree with you.  changed function name to ArchTag

deploy/addons/dashboard/dashboard-dp.template
salamani: changes to ARCHTAG.

tstromberg:
@minikube-bot OK to test

salamani:
@here any idea on these CI Failures? It looks good for me on my s390x as well as on amd64 machine.

tstromberg:
They appear to be flakes. Retrying.

@minikube-bot OK to test

tstromberg:
Thank you for this epic contribution, @salamani! My sincere apologies for it taking so long for us to accept it.  

I look forward to this paving the way for properly supporting other architectures.

PR #4095, opened 2019-04-14
Systemd returns error on inactive, so allow that

### LINKED ISSUE #4092: Running systemctl over the new ssh driver fails
Systemd returns an error code (!) for all kinds of status-like commands,
and the new SSH driver doesn't like when the command it executes fails.

This means that we cannot run things like "is-active" anymore:

```
$ minikube docker-env

💣  Error getting service status: ssh command error:
command : sudo systemctl is-active docker
err     : Process exited with status 3
output  : inactive


```

Or at least not when it returns `inactive`, still OK with `active`.

``` man
       is-active PATTERN...
           Check whether any of the specified units are active (i.e. running). Returns an exit code 0 if at least one is active, or non-zero otherwise. Unless
           --quiet is specified, this will also print the current unit state to standard output.


```

afbjorklund:
The SSH infrastructure currently doesn't handle when the command
is _supposed_ to return an error, so allow this case explicitly.

Closes #4092

### DIFF
--- a/cmd/minikube/cmd/env.go
+++ b/cmd/minikube/cmd/env.go
@@ -26,7 +26,9 @@ import (
 	"text/template"
 
 	"github.com/docker/machine/libmachine"
+	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/host"
+	"github.com/docker/machine/libmachine/log"
 	"github.com/docker/machine/libmachine/shell"
 	"github.com/docker/machine/libmachine/state"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/env.go
```diff
@@ -301,13 +303,31 @@ func (EnvNoProxyGetter) GetNoProxyVar() (string, string) {
...
+	}
+
+	log.Debugf("About to run SSH command:\n%s", command)
+	output, err := client.Output(command)
+	log.Debugf("SSH cmd err, output: %v: %s", err, output)
+	return output, err
+}
+
+// same as host.RunSSHCommand, but allows errors
+func runSSHCommand(h *host.Host, command string) (string, error) {
```
tstromberg: Adding a one-line function to call another new function seems weird. Can they be combined?

RA489: +1

PR #4101, opened 2019-04-15
Extend ReasonableStartTime to 20 minutes to workaround against slow emulators

marcosdiez:
I have a fast machine. 
Minikube runs fine here.
Inside of VmWare Player, it runs fine as well.
Inside of VirtualBox, it times out. 

This solves the problem for me. Should help others without any side effects.

### DIFF
--- a/pkg/util/kubernetes.go
+++ b/pkg/util/kubernetes.go
@@ -41,8 +41,8 @@ import (
 var (
 	// ReasonableMutateTime is how long to wait for basic object mutations, such as deletions, to show up
 	ReasonableMutateTime = time.Minute * 1
-	// ReasonableStartTime is how long to wait for pods to start, considering dependency chains & slow networks.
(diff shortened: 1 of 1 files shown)


tstromberg:
The downside to this approach is that this would require users to wait 20 minutes for a pod that is failing to launch. 10 minutes is already much longer than it should be, and I am surprised VirtualBox is running into an issue here.

I'd rather see this number adjusted by a flag, or determined automatically (but how?), than force users to sit around for another 10 minutes because we haven't otherwise come to the determination that the scheduler is dead.

I'm curious - did you run into an issue with this using v1.0.0 or with minikube checked out from head?

marcosdiez:
You are right. 20 minutes seems too much. Maybe the problem is something on my end.
I "solved" it by preloading every single docker image and both `kubeadm` and `kubelet`.
Not ideal, but since my virtualbox the images are created via a script, it does not care.

PR #4110, opened 2019-04-17
Don't load Minikube on boot 

marcosdiez:
I am not sure if this should be an option or not. Again, what's the use case for `--vm-driver=none` ?

Every other app I run in my laptop does not silently reloads upon a reboot. 

I expect Minikube to behave the same.

Thinking about it, it does not do that if we use any other driver, hence this PR makes things consistent!

That's a behavior change, so if you guys want I can add a command line parameter. `--dont-start-on-boot` for example. (suggestions are welcome)

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -496,7 +496,10 @@ func prepareHostEnvironment(api libmachine.API, kc cfg.KubernetesConfig) bootstr
 		console.OutStyle("option", "%s.%s=%s", eo.Component, eo.Key, eo.Value)
 	}
 	// Loads cached images, generates config files, download binaries
-	if err := bs.UpdateCluster(kc); err != nil {
+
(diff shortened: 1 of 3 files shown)


tstromberg:
/ok-to-test

Alternatively - should we ever have this enabled on boot? I'd rather remove the special case on the none driver if possible.

tstromberg:
Looks good. This will also require an update to `cruntime_test.go` so that `make test` passes.

tstromberg:
On second thought - yes, I think this PR would be better if it the logic was applied consistently so that we didn't enable kubelet across reboots for any drivers. VM reboots are for the time-being somewhat undefined - but my preference here is that `minikube start` should start kubelet in these cases.

marcosdiez:
@tstromberg good idea! Not making the kubernetes cluster start upon reboot is a better alternative for minikube. Now my PR is a one liner!

tstromberg:
Thanks!

Issue #4051, opened 2019-04-04
Improve error handling for "VirtualBox won't boot a 64bits VM when Hyper-V is activated"


Dhiraj0007:
Shows the error bellow:
!   Unable to start VM: create: precreate: This computer is running Hyper-V. VirtualBox won't boot a 64bits VM when Hyper-V is activated. Either use Hyper-V as a driver, or disable the Hyper-V hypervisor. (To skip this check, use --virtualbox-no-vtx-check)

RA489:
@Dhiraj0007 are you using virtualbox to create the VM when Hyper-V is activated?

Dhiraj0007:
Yes

RA489:
@Dhiraj0007 Only one hypervisor can run at a time, and since Hyper-V runs all the time as it is enabled virtualbox won't boot a 64bits VM.

RA489:
@Dhiraj0007 try this workaround and lets see this works for you or not

snip300r:
Did we find any solution for this problem?

aningoo:
When you want to use Hyper-V instead of the default vBox.

1) run CMD as admin
2) "minikube start --vm-driver hyperv"

KnutRyagerInmeta:
I was able to deactivate hyper-V. However, now I get an error message from Docker that it requires Hyper-V to work properly, and `minikube start` just stops at `Preparing Kubernetes v1.15.2 on Docker 18.09.8 ...` indefinitely. My understanding is that running Docker containers is the main purpose of Kubernetes? Confused.

codepractisedemo:
I am also facing the similar problem , when you disable the deactivate hyper-V and it wont able to connect the virtual box itself. since i have download the latest version

lukasmrtvy:
@tstromberg Latest VirtualBox 6.1.16 can run 64bits VM with Hyper-V backend

panktrip:
I am running virtualbox 6.1.32 on windows-10. but I still see the above reported issue.

PS C:> minikube start --driver=virtualbox

* minikube v1.25.1 on Microsoft Windows 10 Enterprise 10.0.19044 Build 19044
* Using the virtualbox driver based on user configuration
* Starting control plane node minikube in cluster minikube
* Creating virtualbox VM (CPUs=2, Memory=4000MB, Disk=20000MB) ...
[...]

PR #4112, opened 2019-04-18
VirtualBox won't boot a 64bits VM when Hyper-V is activated

RA489:
VirtualBox won't boot a 64bits VM when Hyper-V is activated
fixes https://github.com/kubernetes/minikube/issues/4051


tstromberg:
Thanks!

PR #4114, opened 2019-04-18
Assert that docker has started rather than explicitly restarting it

mdiez-modus:
I am not sure why we `restart` instead of `start` docker.

I mean, do we expect it to be in a broken state ?

Removing this 16 bits from the code made things start 10 seconds quicker.

I am not sure if `restart` is needed for other drivers.


tstromberg:
Looks good - though I will admit I don't know what side effects to expect from this - such as modifications to --docker-env. Let's see if the tests uncover anything interesting.

This will also require an update to cruntime_test.go so that make test passes.

mdiez-modus:
PR created with my work user. Sorry. I have to open it again.

PR #4116, opened 2019-04-19
Assert that docker has started rather than explicitly restarting it 

marcosdiez:
I am not sure why we restart instead of start docker.

I mean, do we expect it to be in a broken state ?

Removing this 16 bits from the code made things start 10 seconds quicker.

I am not sure if restart is needed for other drivers.

Same as https://github.com/kubernetes/minikube/pull/4114


tstromberg:
Thank you!

Issue #3724, opened 2019-02-20
Emoji/icons are too large on my terminal


jsravn:
![screenshot from 2019-02-20 15-39-17](https://user-images.githubusercontent.com/522123/53103531-b77d1200-3525-11e9-824d-9716e13c0cba.png)

Please fix this, or remove the icons. It makes minikube unusable for me recently. On gnome-terminal 3.30.2.

tstromberg:
Ouch! That looks very painful, and completely unintentional. Terminal applications do not have the ability to control font sizes, but we do offer a way to turn off this feature:

https://github.com/kubernetes/minikube/blob/master/docs/env_vars.md#example-disabling-emoji
What distribution of Linux is this on, and what version of it?

I thought I've seen everything. I did find one mention of someone running arch linux experiencing this in another application:
https://www.reddit.com/r/archlinux/comments/81wyq2/giant_emoji_in_terminal/

afbjorklund:
You can use `MINIKUBE_IN_COLOR=false` to disable color output (including emojis)

Or maybe just increase your font size to 180pt or whatever would match those icons 😀

jsravn:
Haha, thanks guys. It was pretty funny when I saw this the first time. A giant hand showed up on my terminal (the wait icon I think?).

It is indeed the bitmap scaling which I've disabled, since it can make bitmap fonts blurry. I don't think I've used bitmap fonts in a while though, so I may just enable that again.

shanesveller:
Guess I'm setting that environment variable for the rest of time. Not a fan of the new emoji in the output. Would be really, really nice to not sacrifice all color to avoid that particular aesthetic.

afbjorklund:
AFAIK, minikube doesn't use any color in the output ? Maybe we should split the configuration variables, one for color (text) and one for emoji (icons). That way you could have colored text output - without emojis.

Note that on older platforms the emoji will be shown in black-and-white, so there's no color here...
![minikube-start](https://user-images.githubusercontent.com/10364051/53223545-fae49700-3671-11e9-846e-d69da18adc7b.png)

intelfx:
Who got the clever idea of adding emoji into the output, anyway?..

afbjorklund:
@intelfx: There is an ASCII version available as well.
```
$ export MINIKUBE_IN_COLOR=false
$ minikube start --container-runtime=cri-o --cache-images
o   minikube v0.34.1 on linux (amd64)
$   Caching images in the background ...
[...]

PR #4118, opened 2019-04-19
Add environment variable to disable style prefix

afbjorklund:
Some people find the "style" prefix distracting and confusing,
whether it is shown in color emoji or in 7-bit compatible text.

So add a way to opt-out from the console prefix completely, this
will just show the message text starting from the first column.

From discussion in #3724

### DIFF
--- a/docs/env_vars.md
+++ b/docs/env_vars.md
@@ -13,6 +13,8 @@ Some features can only be accessed by environment variables, here is a list of t
 
 * **MINIKUBE_HOME** - (string) sets the path for the .minikube directory that minikube uses for state/configuration
 
+* **MINIKUBE_NO_STYLE** - disable the style prefix completely
+
 * **MINIKUBE_IN_STYLE** - (bool) manually sets whether or not emoji and colors should appear in minikube. Set to false or 0 to disable this feature, true or 1 to force it to be turned on.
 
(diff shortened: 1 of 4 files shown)


tstromberg:
Rather than adding another knob to support, how about improving MINIKUBE_IN_STYLE?

While I have not yet seen complaints about the ASCII mode, it could be a bit more consistent. I think it's useful to support indentation, but perhaps we should just have a single standard prefix character for info "-" and another for error.

Thoughts?

afbjorklund:
It just seemed like a better option than filtering the output afterwards, as was their current approach...
 
But maybe the style output can be tweaked enough to be acceptable ? Or convert it from a boolean ?

afbjorklund:
A fourth approach was printing the style out in full, like they do in the subtitles:

```
😄  minikube v1.0.0 on linux (amd64)
o   minikube v1.0.0 on linux (amd64)
minikube v1.0.0 on linux (amd64)
[happy] minikube v1.0.0 on linux (amd64)
```

But maybe that would work better with categories, than with the current styles.

tstromberg:
I'd rather not support 3 output modes if I could avoid it. Let's just tweak the default output style to be more acceptable.

Here's a straw-man proposal:

- top level
  * second level
! error/warning

afbjorklund:
I'm not going to pursue this further, so probably time to open a new issue for the original requester then.

PR #4129, opened 2019-04-22
Don't attempt to pull docker images on relaunch

marcosdiez:
Hi again!
If we are relaunching a cluster, we already have pulled the images (and if they ever got deleted, it's not the end of the world). That saves startup time!

The fact is, since I am physically far from the k8s.gcr.io servers, a simple `time sudo kubeadm config images pull --config "/var/lib/kubeadm.yaml"` takes around 30 seconds if the images are already there.

From an AWS server in the US the same command takes only 5 seconds, but, well, it's still a performance improvement without any downside :) 

Please don't think that I have the worse connection on earth and hence I am an edge case. Quite the opposite. I am just physically far for `k8s.gcr.io` and therefore the latency is high`.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -170,7 +170,7 @@ func runStart(cmd *cobra.Command, args []string) {
 	if err != nil && !os.IsNotExist(err) {
 		exit.WithCode(exit.Data, "Unable to load config: %v", err)
 	}
-	k8sVersion := validateKubernetesVersions(oldConfig)
+	k8sVersion, isUpgrade := validateKubernetesVersions(oldConfig)
 	config, err := generateConfig(cmd, k8sVersion)
 	if err != nil {
 		exit.WithError("Failed to generate config", err)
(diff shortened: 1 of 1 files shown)


tstromberg:
@minikube-bot OK to test

tstromberg:
@minikube-bot OK to test

RA489:
@marcosdiez please squash the commits to one

cmd/minikube/cmd/start.go
```diff
@@ -574,6 +570,11 @@ func bootstrapCluster(bs bootstrapper.Bootstrapper, r cruntime.Manager, runner b
 			exit.WithLogEntries("Error restarting cluster", err, logs.FindProblems(r, bs, runner))
 		}
 		return
+	} else {
+		console.OutStyle("pulling", "Pulling images required by Kubernetes %s ...", kc.KubernetesVersion)
```
tstromberg: I like this change - but there is a bit more subtlety here that is missing. This will also need to handle the case where the pre-existing cluster is a different version of Kubernetes.

tstromberg:
Any idea on how much time this saves? I want to make sure the added complexity is worth it.

cmd/minikube/cmd/start.go
marcosdiez: Here in Brazil it really saves me 30 seconds. So I will later send you will a PR to check if the server is being upgraded or not.

I read that you aim to make minikube launch in less than a minute. 

With this patch (and all the others you have merged), here it loads in 53 seconds. How long does it take to load for you ?

marcosdiez:
@tstromberg Implementing the check was easier than I thought. And I only had to add 4 lines, one of them being blank. I also rebased, as you requested.

tstromberg:
@minikube-bot OK to test

PR #4139, opened 2019-04-23
BugFix:  ExecRunner.Copy now parses permissions strings as octal

marcosdiez:
I actually fix two things here.
First, `ExecRunner.Copy` was parsing permissions as decimal integers instead of octal integers, which was giving odd results.

Second, for me it seems that `/usr/bin/kubeadm` and `/usr/bin/kubelet` should have `0755` instead of `0641`


tstromberg:
@minikube-bot OK to test

PR #4165, opened 2019-04-26
Do not use separators when printing numbers

sharifelgamal:


Gopkg.lock
```diff
@@ -696,7 +696,7 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:11d290de3457882172ce8d9ffe930999cfb62929b58e486e1ff1b6adcf3c52bc"
+  digest = "1:d3cc042f4e05e3edff799edb30d84c510d0c5f433a1ba5a139e0a88ec5f6b5a9"
```
tstromberg: Since this already includes a Gopkg.lock update, do you mind pinning x/text to a specific version in Gopkg.toml so that we don't have other future breakage?

pkg/minikube/console/style.go
```diff
@@ -131,6 +132,11 @@ func lowPrefix(s style) string {
 // Apply styling to a format string
 func applyStyle(style string, useColor bool, format string, a ...interface{}) (string, error) {
 	p := message.NewPrinter(preferredLanguage)
+	for i, x := range a {
+		if _, ok := x.(int); ok {
+			a[i] = number.Decimal(x, number.NoSeparator())
```
tstromberg: Sneaky. Let's add a test so that this doesn't accidentally get refactored out in a 
 future PR.

Gopkg.lock
sharifelgamal: Done

pkg/minikube/console/style.go
sharifelgamal: Done

tstromberg:
@minikube-bot OK to test

PR #4174, opened 2019-04-29
Remove unused ArchLinux releaser, raise errors if Brew push fails.

tstromberg:
Also renames release_update_installers.sh to release_update_brew.sh It's preferable for each package script to run in its own process space so that steps can't interfere with one another.

This should address some of the wonkiness we saw with the release this morning.


tstromberg:
Yup. We will have to rename the script path called in the Release trigger. I'll make sure to do that.

PR #4188, opened 2019-05-01
Add script to update kubernetes version

sharifelgamal:
Usage: from `hack/kubernetes_version`, run

`go run update_kubernetes_version.go [v]1.14.1`


sharifelgamal:
Either parameterizing this script or adding a separate script for newest seems like the easiest solution there. We'll cross that bridge when we get there.

PR #4194, opened 2019-05-02
Fix the deploy targets, restore default target

afbjorklund:
The default target (out/minikube) was replaced by "makedeploys",
that would always rebuild the deploy .yaml files from .template


tstromberg:
Ouch. I believe this was caused by #4091

Issue #4195, opened 2019-05-02
kvm2: this function is not supported by the connection driver: virNetworkDefineXML


frunzales:
I'm unable to do a 'minikube start' on gentoo ~amd64. Could this be libvirt related?
```
xxx@x230 ~ $ minikube start --vm-driver kvm2 --logtostderr -v 8
😄  minikube v1.0.1 on linux (amd64)
...
💣  Unable to start VM: create: Error creating machine: Error in driver during machine creation: creating network: defining network from xml:
<network>
  <name>minikube-net</name>
  <dns enable='no'/>
  <ip address='192.168.39.1' netmask='255.255.255.0'>
    <dhcp>
      <range start='192.168.39.2' end='192.168.39.254'/>
    </dhcp>
  </ip>
</network>
: virError(Code=3, Domain=19, Message='this function is not supported by the connection driver: virNetworkDefineXML')
```

tstromberg:
I found this Gentoo forum post which seems relevant: https://forums.gentoo.org/viewtopic-t-981692-start-0.html

`You need libvirt built with USE=virt-network too.`

Do you mind checking if that fixes the problem, and if it does, provide the exact steps for rebuilding libvirt appropriately?  I'd like to improve the error message so that it is more helpful to others who run into this problem.

frunzales:
@tstromberg you're right, I rebuild libvirt with USE=virt-network and things seem to be better. I think it's safe to close for now.

Issue #4107, opened 2019-04-17
Staring minikube with virtual box on Mojave fails with 0x80004005


domschoen:

domschoen:
Here the details of the configuration:
- Mac Mini 2018
- OSX: mac OS Mojave 10.14.4
- VirtualBox: 6.0.6 - 130049- OSX
I tried some of the suggested solutions like reinstalling Virtual Box etc but nothing work. That's way I have created a new issued because it may be specific to my configuration

tstromberg:
Thanks for the bug report! Do you mind providing the output of:

`minikube start --alsologtostderr -v=8`

This seems to be a rather general error code with many possibilities:

- https://apple.stackexchange.com/questions/323960/virtualbox-ns-error-failure-0x80004005-mac
- https://forums.virtualbox.org/viewtopic.php?f=8&t=90147
- https://stackoverflow.com/questions/52689672/virtualbox-ns-error-failure-0x80004005-macos
- https://apple.stackexchange.com/questions/343705/virtualbox-app-would-no-longer-runs

The general gist appears that something, either MacOS or Symantec, is preventing VirtualBox from opening. Could you uninstall VirtualBox, then install it again, and then go to "System Preferences -> Security & Privacy -> General" and ensure that it does not say "Some system software was blocked from loading"? See the image here: https://apple.stackexchange.com/questions/343705/virtualbox-app-would-no-longer-runs

This seems entirely fixable, but if it isn't, there is also the hyperkit VM driver we could use. Thank you!

tstromberg:
I'm closing this issue as it hasn't seen activity in awhile, and it's unclear if this issue still exists. If this issue does continue to exist in the most recent release of minikube, please feel free to re-open it. 

Thank you for opening the issue!

ecourtial:
For me the issue was coming from a bad update of VirtualBox on Ubuntu. The driver was too old despite the VB update.
I had to remove virtual box and virtualbox-dkms and reinstall virtual box.

Mohitduklan:
Worked just by reinstalling virtualbox using cli

Issue #4043, opened 2019-04-03
 Unable to start VM: WARNING: The vboxdrv kernel module is not loaded.


jainVish:

balopat:
hi @jainVish, can you try to upgrade your virtualbox installation?

tstromberg:
NOTE: This error is a sign of a failed VirtualBox installation. The solution is to reinstall VirtualBox. We should direct users to do so.

frankmeten:
> NOTE: This error is a sign of a failed VirtualBox installation. The solution is to reinstall VirtualBox. We should direct users to do so. This may help isolate the specific failure as well:
> 
> `sudo modprobe vboxdrv`


CentOS 7 can reference the following link

https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Kernel_Administration_Guide/sect-signing-kernel-modules-for-secure-boot.html

John-Dormevil:
Try to:
- Deactivate Secure Boot from your bios system;
- Then run `sudo /sbin/vboxconfig` in case you install and use virtualbox 6+
- Then `minikube start`

pythonInRelay:
I know this is closed, but I just wanted to say. The problem for me was a few missing packages, and then all I really needed to do was just reboot my machine and everything worked perfectly. I just installed a couple of packages, namely these ones:

`community/virtualbox 6.1.22-3 [installed]
community/virtualbox-ext-vnc 6.1.22-3 [installed]
community/virtualbox-guest-iso 6.1.22-1 [installed]
community/virtualbox-guest-utils 6.1.22-3 [installed]
community/virtualbox-host-dkms 6.1.22-3 [installed]
community/virtualbox-sdk 6.1.22-3 [installed]
chaotic-aur/linux-tkg-muqss 5.12.15-178 [installed]
chaotic-aur/virtualbox-ext-oracle 6.1.22-1 [installed]
chaotic-aur/virtualbox-meta 2-1 [installed]`

Rebooted my machine and no problem.

OkabeRitarou:
Thanks @pythonInRelay

Issue #3974, opened 2019-03-26
Unable to start VM: create: precreate: The character device /dev/vboxdrv does not exist


yrmmout:
minikube v0.35.0 on linux (amd64)
🔥  Creating virtualbox VM (CPUs=2, Memory=2048MB, Disk=20000MB) ...
💣  Unable to start VM: create: precreate: We support Virtualbox starting with version 5. Your VirtualBox install is "WARNING: The character device /dev/vboxdrv does not exist.\n\t Please install the virtualbox-dkms package and the appropriate\n\t headers, most likely linux-headers-generic.\n\n\t You will not be able to start VMs until this problem is fixed.\n5.2.18_Ubuntur123745". Please upgrade at https://www.virtualbox.org

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
👉  https://github.com/kubernetes/minikube/issues/new

Think you

tstromberg:
It sounds like VirtualBox is not completely installed.  Could you try uninstalling it and installing it again?

You may also want to take the time to upgrade VirtualBox to 6.0: https://www.virtualbox.org/wiki/Downloads

balopat:
friendly ping @yrmmout - does reinstalling VirtualBox help?

akanurlan0:
> friendly ping @yrmmout - does reinstalling VirtualBox help?
Faced the same issue, reinstallation didn't help. I am running ubuntu 18.04, tried virtualbox 6.0 and 5.2 both. 

``
Unable to start VM: create: precreate: We support Virtualbox starting with version 5. Your VirtualBox install is "WARNING: The vboxdrv kernel module is not loaded. Either there is no module\n         available for the current kernel (4.15.0-46-generic) or it failed to\n         load. Please recompile the kernel module and install it by\n\n           sudo /sbin/vboxconfig\n\n         You will not be able to start VMs until this problem is fixed.\n6.0.4r128413". Please upgrade at https://www.virtualbox.org
``

tstromberg:
I'm closing this issue as it hasn't seen activity in awhile, and it's unclear if this issue still exists. If this issue does continue to exist in the most recent release of minikube, please feel free to re-open it. 

Thank you for opening the issue!

ajit555db:
I disabled the "secured boot" option on my host machine.

https://askubuntu.com/questions/900118/vboxdrv-sh-failed-modprobe-vboxdrv-failed-please-use-dmesg-to-find-out-why

[...]

PR #4198, opened 2019-05-02
Add solution text for common kvm2 and VirtualBox problems

tstromberg:
Contains solutions for:

- #1994
- #3614
- #3859
- #3864
- #3910
- #3974
- #4043
- #4107
- #4195

### DIFF
--- a/pkg/minikube/problem/err_map.go
+++ b/pkg/minikube/problem/err_map.go
@@ -36,10 +36,39 @@ var vmProblems = map[string]match{
 		Advice: "In some environments, this message is incorrect. Try 'minikube start --no-vtx-check'",
 		Issues: []int{3900},
 	},
-	"VBOX_THIRD_PARTY": {
-		Regexp: re(`The virtual machine * has terminated unexpectedly during startup with exit code 1`),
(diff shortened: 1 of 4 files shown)


pkg/minikube/problem/err_map.go
```diff
@@ -36,11 +36,43 @@ var vmProblems = map[string]match{
...
+		Regexp: re(`The vboxdrv kernel module is not loaded`),
+		Advice: "Run 'sudo modprobe vboxdrv' and reinstall VirtualBox if it fails.",
+		Issues: []int{4043},
+	},
+	"VBOX_DEVICE_MISSING": {
+		Regexp: re(`/dev/vboxdrv does not exist`),
+		Advice: "Run 'sudo modprobe vboxdrv' and reinstall VirtualBox if it fails.",
+		Issues: []int{3974},
+	},
+	"VBOX_HARDENING": {
```
sharifelgamal: This is an improvement over VBOX_THIRD_PARTY above. Go ahead and remove that one.

tstromberg: Good call. I've also added GOOS matching, since the advice is slightly different between Mac & Windows.

PR #4199, opened 2019-05-03
Add automatic locale detection

### LINKED ISSUE #4183: Add code to switch languages appropriately
This is to track the work for adding one additional language to minikube.

sharifelgamal:
Fixes #4183

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -30,6 +30,7 @@ import (
 	"time"
 
 	"github.com/blang/semver"
+	"github.com/cloudfoundry-attic/jibber_jabber"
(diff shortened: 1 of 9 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -691,3 +693,12 @@ func saveConfig(clusterConfig cfg.Config) error {
 	}
 	return nil
 }
+
+func determineLocale() {
```
tstromberg: This should be in `main.go`, so that it affects all commands, rather than just `minikube start`.

sharifelgamal: Makes sense. done.

PR #4216, opened 2019-05-06
Replace ARCH sed macros with Go templates

medyagh:


tstromberg:
@minikube-bot OK to test

pkg/minikube/assets/addons.go
```diff
@@ -382,8 +383,10 @@ func addMinikubeDirToAssets(basedir, vmpath string, assets *[]CopyableFile) erro
 func GenerateTemplateData(cfg config.KubernetesConfig) interface{} {
 	opts := struct {
 		ImageRepository string
+		ArchTag         string
 	}{
 		ImageRepository: cfg.ImageRepository,
+		ArchTag:         runtime.GOARCH,
```
tstromberg: How about just "Arch": it's shorter.

deploy/addons/storage-provisioner/storage-provisioner.template
```diff
@@ -51,7 +51,7 @@ spec:
   hostNetwork: true
   containers:
   - name: storage-provisioner
-    image: {{default "gcr.io/k8s-minikube" .ImageRepository}}/storage-provisioner-ARCHTAG_NONE:v1.8.1
+    image: {{default "gcr.io/k8s-minikube" .ImageRepository}}/storage-provisioner-{{default "ARCHTAG_NONE" .ArchTag}}:v1.8.1
```
tstromberg: ARCHTAG_NONE feels like a hack, but I don't honestly understand it enough. 

I don't want to violate Chesterton's Fence (https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence), but I wonder if this template macro just be safely removed with this new model, in preference to `storage-provisioner:v1.1.1`?

I believe all of the mentions of ARCHTAG in these template files should be replaced with {{ .Arch }}, however.

Makefile
```diff
@@ -100,13 +100,6 @@ ifeq ($(GOOS),windows)
...
 
-ifeq ($(GOARCH),amd64)
-	ARCHTAG ?= -amd64
-	ARCHTAG_NONE ?=
-else
-	ARCHTAG ?= -$(GOARCH)
-	ARCHTAG_NONE ?= -$(GOARCH)
-endif
 
 DEPLOYS=\
```
tstromberg: I believe DEPLOYS should be removable with this change as well. You'll need to rename the .template files back into .yaml though.

tstromberg:
The original pull request by @salamani that added multi-arch support might be useful context:

https://github.com/kubernetes/minikube/pull/4091

tstromberg:
The test failure makes me wonder if this PR needs a rebase, or if I broke head :(

deploy/addons/storage-provisioner/storage-provisioner.template
medyagh: fair enough

Makefile
```diff
@@ -229,7 +210,8 @@ out/test.d: pkg/minikube/assets/assets.go
 test:
 	GOPATH=$(GOPATH) ./test.sh
 
-pkg/minikube/assets/assets.go: $(DEPLOYS)
+# Regenerates assets.go when tempalte files have been updated
```
tstromberg: "template"

tstromberg:
Looks great. Thank you!

Issue #4235, opened 2019-05-10
Browser Log noise in terminal when opening dashboard


medyagh:
opening the dashbord in browser, produces a lot of GTK noise for css. it is better to send those browser related logs not into the terminal

**The exact command to reproduce the issue**:
`minikube dashbord`

**The full output of the command that failed**:
```
$ ./out/minikube dashboard
🔌  Enabling dashboard ...
🤔  Verifying dashboard health ...
🚀  Launching proxy ...
🤔  Verifying proxy health ...
🎉  Opening http://127.0.0.1:40175/api/v1/namespaces/kube-system/services/http:kubernetes-dashboard:/proxy/ in your default browser...

(google-chrome-stable:123045): Gtk-WARNING **: 17:36:01.563: Theme parsing error: gtk-widgets.css:186:14: not a number

(google-chrome-stable:123045): Gtk-WARNING **: 17:36:01.563: Theme parsing error: gtk-widgets.css:186:14: Expected a string.

(google-chrome-stable:123045): Gtk-WARNING **: 17:36:01.567: Theme parsing error: gtk-widgets.css:2749:24: not a number

(google-chrome-stable:123045): Gtk-WARNING **: 17:36:01.567: Theme parsing error: gtk-widgets.css:2749:24: Expected a string.

(google-chrome-stable:123045): Gtk-WARNING **: 17:36:01.568: Theme parsing error: gtk-widgets.css:2940:14: not a number

(google-chrome-stable:123045): Gtk-WARNING **: 17:36:01.568: Theme parsing error: gtk-widgets.css:2940:14: Expected a string.

(google-chrome-stable:123045): Gtk-WARNING **: 17:36:01.568: Theme parsing error: gtk-widgets.css:2946:17: Expected a string.

(google-chrome-stable:123045): Gtk-WARNING **: 17:36:01.569: Theme parsing error: gtk-widgets.css:4083:14: not a number

(google-chrome-stable:123045): Gtk-WARNING **: 17:36:01.569: Theme parsing error: gtk-widgets.css:4083:14: Expected a string.

(google-chrome-stable:123045): Gtk-WARNING **: 17:36:01.569: Theme parsing error: gtk-widgets.css:4088:17: Expected a string.

(google-chrome-stable:123045): Gtk-WARNING **: 17:36:01.570: Theme parsing error: gtk-widgets.css:4729:14: not a number

[...]

Issue #2548, opened 2018-02-14
dashboard as root: Unsupported authorisation protocol / cannot open display: localhost:10.0


farooqte96:
Hi i just installed minikube on a remote linux machine (Ubuntu 16.04 server version). I am able to start minikube but to see running deployments, **minikube dashboard** command is  reporting some errors like unsupported authorization protocol, missing modules (chrome browser, mozilla etc.). Can anyone tell me how to fix this issue so i can see dashboard on my local pc by accessing : http://192.140.127.45:30000

minikube status
minikube: Running
cluster: Running
kubectl: Correctly Configured: pointing to minikube-vm at 192.140.127.45
kubectl cluster-info
Kubernetes master is running at https://192.140.127.45:8443
**root@yx-lab2: minikube dashboard**

Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Waiting, endpoint for service is not ready yet...
Opening kubernetes dashboard in default browser...
MobaXterm X11 proxy: Unsupported authorisation protocol
Failed to connect to Mir: Failed to connect to server socket: No such file or directory
Unable to init server: Broadway display type not supported: localhost:10.0
Error: cannot open display: localhost:10.0
MobaXterm X11 proxy: Unsupported authorisation protocol
Failed to connect to Mir: Failed to connect to server socket: No such file or directory
Unable to init server: Broadway display type not supported: localhost:10.0
Error: cannot open display: localhost:10.0
MobaXterm X11 proxy: Unsupported authorisation protocol
Failed to connect to Mir: Failed to connect to server socket: No such file or directory
Unable to init server: Broadway display type not supported: localhost:10.0
Error: cannot open display: localhost:10.0
/usr/bin/xdg-open: 771: /usr/bin/xdg-open: iceweasel: not found
[...]

PR #4237, opened 2019-05-10
Reroute browser logout from stdout

medyagh:
Reroutes the logs from the browser (for the dashboard addon) not to flood the stdout and stderr

Fixes https://github.com/kubernetes/minikube/issues/4235

### DIFF
--- a/cmd/minikube/cmd/dashboard.go
+++ b/cmd/minikube/cmd/dashboard.go
@@ -18,9 +18,12 @@ package cmd
 
 import (
 	"bufio"
+	"bytes"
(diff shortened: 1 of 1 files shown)


tstromberg:
I'm concerned that this may hide valuable error messages for users to debug why their browser isn't opening. For instance: https://github.com/kubernetes/minikube/issues/2548

Thoughts?

medyagh:
Given that this seems to happen only on certain linux configurations, and the browser output might be helpful. I am okay with closing this PR.

PR #4253, opened 2019-05-14
De-lint everything and replace "go lint" with golangci-lint

tstromberg:
This replaces the "gofmt" check with "golangci-lint", pinned to v1.16.0.

This applies a multitude of changes to make the linter happy in its current configuration. Many error checking bugs died to bring us clean lint output.


pkg/drivers/hyperkit/iso.go
```diff
@@ -45,36 +44,15 @@ func ExtractFile(isoPath, srcPath, destPath string) error {
...
 	if err != nil {
 		return err
 	}
+	defer dst.Close()
 
 	_, err = io.Copy(dst, f.Sys().(io.Reader))
 	return err
 }
 
-func readFile(isoPath, srcPath string) (string, error) {
```
sharifelgamal: Is this just no longer needed?

hack/kubernetes_version/update_kubernetes_version.go
```diff
@@ -56,21 +58,30 @@ func main() {
...
+		if err != nil {
+			return err
 		}
-		return nil
+		if !strings.HasSuffix(path, "default.yaml") {
+			return nil
+		}
+		cf, err = ioutil.ReadFile(path)
+		if err != nil {
+			// Keep going if this errors out
```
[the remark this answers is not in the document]
tstromberg: Nope. Fixed.

pkg/drivers/hyperkit/iso.go
tstromberg: Correct. The linter identified this as an unused function, and it seems legit.

PR #4255, opened 2019-05-14
Return host IP when using vmware as vm driver.

ne0h:
Adds the missing case "vmware" to return the host IP.

### DIFF
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -381,7 +381,15 @@ func GetVMHostIP(host *host.Host) (net.IP, error) {
 	case "xhyve", "hyperkit":
 		return net.ParseIP("192.168.64.1"), nil
 	case "vmware":
-		return net.ParseIP("192.168.4.1"), nil
+		vmIPString, err := host.Driver.GetIP()
+		if err != nil {
+			return []byte{}, errors.Wrap(err, "Error getting VM IP address")
+		}
+		vmIP := net.ParseIP(vmIPString).To4()
+		if vmIP == nil {
+			return []byte{}, errors.Wrap(err, "Error converting VM IP address to IPv4 address")
+		}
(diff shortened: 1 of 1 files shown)


ne0h:
Signed CLA.

tstromberg:
Hey ne0h!

I'm a bit confused by this PR: the title says it implements mounting, but the PR adds an IP. 

Can you elaborate some in the PR description as to how the two are related? Thanks!

tstromberg:
@minikube-bot OK to test

ne0h:
Hey @tstromberg, you are absolutely right. I updated the commit and the description of the PR to be more applicable.

ne0h:
Given that `host.Driver.GetIP()` returns the guest IP address reliably, the switch statement could be removed completely?

tstromberg:
> Given that `host.Driver.GetIP()` returns the guest IP address reliably, the switch statement could be removed completely?

I wonder about this. I'd be happy to merge this PR as is, but if this simplification works, that's great. I'm concerned however that some drivers may have different values for host.Driver.GetIP() and the VM-Host IP, but the integration tests should catch that. Feel free to propose in a different PR if you like.

pkg/minikube/cluster/cluster.go
```diff
@@ -380,6 +380,16 @@ func GetVMHostIP(host *host.Host) (net.IP, error) {
 		return ip, nil
 	case "xhyve", "hyperkit":
 		return net.ParseIP("192.168.64.1"), nil
+	case "vmware":
+		vmIPString, err := host.Driver.GetIP()
+		if err != nil {
+			return []byte{}, errors.Wrap(err, "Error getting VM IP address")
+		}
+		vmIP := net.ParseIP(vmIPString).To4()
```
tstromberg: To be consistent with the others, I think you can just call:

return net.ParseIP(vmIPstring), nil

PR #4270, opened 2019-05-16
Test the upgrade of Kubernetes versions (oldest->newest)

tstromberg:
This combines the release->trunk upgrade test with one that also upgrades Kubernetes versions from oldest->newest.  I'd normally rather not combine tests, but this is done in the interest of keeping the integration test runtime short.


tstromberg:
@minikube-bot OK to test

Issue #4264, opened 2019-05-16
Synchronize VM clocks at start time


tstromberg:
Sometimes VM's may run into issues where their clocks fall behind the times. We should detect and fix when possible, and error when impossible.
 
See #4263 for context.

GavinOsborn:
We run into this issue more often than we'd like. The APIs that we build and test accept various forms of Bearer tokens for authentication. Validation of these tokens often fails due to clock skew.

tstromberg:
Fixed at head.

reegnz:
Still experiencing the issue with minikube 1.1.1. See my comment in #1378

PR #4283, opened 2019-05-17
Sync guest system clock if desynchronized from host

### LINKED ISSUE #1378: hyperkit clock falls behind: failed to write or validate certificate: the certificate is not valid yet
**Minikube version**: v0.18.0

**Environment**:
- **OS**: MacOS Sierra (10.12.4 (16E195))
- **VM Driver** : xhyve
- **ISO version** : v0.18.0

**What happened**:
When I wake my laptop after sleep, the clock of the Minikube VM lags behind. This causes problems with the registry credentials plugin, since AWS rejects credentials requests with an invalid timestamp.

The clock should sync after computer wakeup.

See above.

tstromberg:
Fixes #1378 and #4264


tstromberg:
WIP because I would like to write a decent test for this.

sharifelgamal:
Test failure is unrelated to this change.

pkg/minikube/cluster/cluster.go
```diff
@@ -145,11 +158,60 @@ func waitForSSHAccess(h *host.Host, e *engine.Options) error {
...
+		glog.Infof("guest clock delta is within tolerance: %s", d)
+		return nil
+	}
+	if err := adjustGuestClock(h, time.Now()); err != nil {
+		return errors.Wrap(err, "adjusting system clock")
+	}
+	return nil
+}
+
+// systemClockDelta returns the approximate difference between the host and guest system clock
```
medyagh: wrong func name in the comment "systemClockDelta" should be "guestClockDelta"

PR #4297, opened 2019-05-20
Make the minikube binary depend on cmd/ and pkg/ go files

tstromberg:
This used to be handled by `makedepend.sh`, but it was retired as part of the move to go modules.


tstromberg:
@minikube-bot OK to test

PR #4298, opened 2019-05-20
Add missing testdata update to v1.14.2

tstromberg:


sharifelgamal:
We should add this to the script as well.

PR #4308, opened 2019-05-21
Validate kvm network exists

josedonizetti:
Minikube isn't validating if the kvm network exists, in the case the `default` one isn't created or the user passed a custom one `--kvm-network` and it doesn't exist the code will go on creating the PrivateNetwork to only raise an error at the end. This PR adds validation for this case, so it fails fast instead of moving on with other resources creation.


tstromberg:
Thank you for contributing this! Having validation like this will definitely help minikube users debug what's going on. It's a great step forward to a world where minikube is able to fix kvm network configuration errors.

PR #4310, opened 2019-05-21
Add some CRI-agnostic image build documentation

### LINKED ISSUE #4309: CRI-agnostic image build documentation
Currently everything circles around the `minikube docker-env` command.

We should add some more container runtime agnostic documentation...

This will allow you to build a container image using either of the two provided.

Both `docker` (containerd) and `podman` (cri-o) allow you to use Dockerfile.

afbjorklund:
Closes #4309

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -26,6 +26,8 @@
 
 * **Reusing the Docker Daemon** ([reusing_the_docker_daemon.md](reusing_the_docker_daemon.md)): How to point your docker CLI to the docker daemon running inside minikube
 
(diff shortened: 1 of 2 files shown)


docs/building_on_the_vm.md
```diff
@@ -0,0 +1,31 @@
+# Building on the VM
```
tstromberg: "Building on the VM" could mean a lot of things. What do you think about a more specific title, such as "Building images within the VM"?

afbjorklund: Yeah, that is probably a better title

PR #4321, opened 2019-05-22
Make buildah --no-pivot default, using env var

afbjorklund:
Needed for doing "podman build" (uses buildah) when running on rootfs.

This environment variable has the same affect as DOCKER_RAMDISK does.


afbjorklund:
Not much use doing `sudo -E` when you forget to provide the environment...

deploy/iso/minikube-iso/package/podman/buildah.profile
```diff
@@ -0,0 +1 @@
+export BUILDAH_NOPIVOT=true
```
tstromberg: This is somewhat esoteric, so please add a comment here describing why this is necessary to help future code archaeologists out.

afbjorklund: Will do, the fact that `podman build` uses buildah to do the work might not be obvious... The name of the option comes from `runc run --no-pivot`

PR #4323, opened 2019-05-22
kvm2: Add support for --kvm-network to ensureNetwork

josedonizetti:
kvm `ensureNetwork` is directly checking for the constant `defaultNetworkName` which is a problem when a user is passing a custom network through the option `--kvm-network`.

Kvm drivers pass the `Network` to be the same as the flag `--kvm-network`.
https://github.com/kubernetes/minikube/blob/master/pkg/minikube/drivers/kvm/driver.go#L67
https://github.com/kubernetes/minikube/blob/master/pkg/minikube/drivers/kvm2/driver.go#L67

Even the default Host creation sets up the `Network`:
https://github.com/kubernetes/minikube/blob/master/pkg/drivers/kvm/kvm.go#L107

The PR changes `ensureNetwork` to use the given network instead of hardcoded constant.

### DIFF
--- a/pkg/drivers/kvm/network.go
+++ b/pkg/drivers/kvm/network.go
@@ -91,8 +91,8 @@ func (d *Driver) ensureNetwork() error {
 
 	// Start the default network
 	// It is assumed that the libvirt/kvm installation has already created this network
-	log.Infof("Ensuring network %s is active", defaultNetworkName)
-	if err := setupNetwork(conn, defaultNetworkName); err != nil {
+	log.Infof("Ensuring network %s is active", d.Network)
+	if err := setupNetwork(conn, d.Network); err != nil {
 		return err
 	}
 
--- a/pkg/drivers/kvm/network.go
+++ b/pkg/drivers/kvm/network.go
@@ -89,7 +89,6 @@ func (d *Driver) ensureNetwork() error {
 
 	// network: default
 
-	// Start the default network
 	// It is assumed that the libvirt/kvm installation has already created this network
 	log.Infof("Ensuring network %s is active", d.Network)
 	if err := setupNetwork(conn, d.Network); err != nil {
--- a/test/integration/start_stop_delete_test.go
+++ b/test/integration/start_stop_delete_test.go
@@ -38,6 +38,7 @@ func TestStartStop(t *testing.T) {
 		{"nocache_oldest", []string{
 			"--cache-images=false",
 			fmt.Sprintf("--kubernetes-version=%s", constants.OldestKubernetesVersion),
+			"--kvm-network=default",
 		}},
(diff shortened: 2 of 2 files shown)


pkg/drivers/kvm/network.go
```diff
@@ -91,8 +91,8 @@ func (d *Driver) ensureNetwork() error {
 
 	// Start the default network
```
tstromberg: Remove obsolete comment. The logging messages tell enough of a story.

tstromberg:
@minikube-bot OK to test

josedonizetti:
@tstromberg Let me know if I got the integration test correct (unfortunately they break locally even without my change). Also, do you want me to squash the commits into one?

test/integration/start_stop_delete_test.go
```diff
@@ -38,6 +38,7 @@ func TestStartStop(t *testing.T) {
 		{"nocache_oldest", []string{
 			"--cache-images=false",
 			fmt.Sprintf("--kubernetes-version=%s", constants.OldestKubernetesVersion),
+			"--kvm-network=default",
```
tstromberg: is the value "default" special in any way? If not, I'd prefer it to be something that is unlikely to conflict or appear special, like "test-net".

tstromberg:
> @tstromberg Let me know if I got the integration test correct (unfortunately they break locally even without my change). Also, do you want me to squash the commits into one?

No need to squash. I'm curious though about the local breakage - what are you seeing?

test/integration/start_stop_delete_test.go
josedonizetti: Nope, it's not. I'll change it to `test-net`. :)

josedonizetti:
@tstromberg They are passing now. Seems the tests I was running with the `--kvm-network` left the `./minikube/config.json` in a bad state, which wasn't allowing a start, breaking the tests, not even start it on the cmd line directly, after removing the kvm domain by hand, it passed.  
I'll investigate this a little bit more, and try to address in a separate PR.

test/integration/start_stop_delete_test.go
josedonizetti: Actually, I take this back. As the integration test does start minikube,  the `--kvm-network=net-test` won't exist, it will fail. Unless we create this network on the CI beforehand, but probably we should have integration tests for both cases. Using the default network, using the pre-defined network.

josedonizetti:
@tstromberg Builds for KVM, and VirtualBox never returned? It seems none build is breaking elsewhere as well.

tstromberg:
@minikube-bot OK to test

FYI - the none failure is https://github.com/kubernetes/minikube/issues/4339

PR #4331, opened 2019-05-23
Skip kvm network deletion if private network doesn't exist

josedonizetti:
Not returning an error when deleting a kvm network is optional, because the driver will ignore it:
https://github.com/kubernetes/minikube/blob/master/pkg/drivers/kvm/kvm.go#L435

But I think that skipping with a proper log message if the network doesn't exist, tells a better story on the code, and as well on the logs. 

WDYT?


tstromberg:
@minikube-bot OK to test

tstromberg:
Thank you!

PR #4334, opened 2019-05-23
Refactor kvm lookupIPFromStatusFile

josedonizetti:
There are a few problems with the kvm `lookupIPFromStatusFile`.

1 - when the IP is found, it keeps executing the for instead of returning directly.
2 - it ignores the `json.Unmarshal` error due to a 'possible' empty file

This PR refactor the code to:

1 - return as soon as the IP match was found
2 - not ignore the JSON error, in case of an empty file it returns blank, but if the JSON itself is invalid we report the error because it means a possible error on the KVM installation of the user, which worths notifying.  
3 - add tests to this logic

For the tests, I've used the same convention I saw for the hyperkit tests.


tstromberg:
@minikube-bot OK to test

tstromberg:
@minikube-bot OK to test

Issue #4319, opened 2019-05-22
kvm2 v1.2.0: Failed to delete cluster: Requested operation is not valid: domain is not running


carroarmato0:
There was an existing cluster created by Minikube 1.0.1 and shutdown.
After upgrading minikube to 1.1.0 and the corresponding docker-machine-driver-kvm2, deleting the existing cluster didn't work.

Starting the existing cluster triggered minikube to perform a successful upgrade, although deleting the upgraded cluster afterwards still doesn't work.

Upgrading the cluster:
```
[carroarmato0@neon-flower:~] 70 $ minikube start
😄  minikube v1.1.0 on linux (amd64)
👍  minikube will upgrade the local cluster from Kubernetes 1.14.1 to 1.14.2
💿  Downloading Minikube ISO ...
 131.28 MB / 131.28 MB [============================================] 100.00% 0s
2019/05/22 09:26:30 No matching credentials were found, falling back on anonymous
2019/05/22 09:26:30 No matching credentials were found, falling back on anonymous
2019/05/22 09:26:30 No matching credentials were found, falling back on anonymous
2019/05/22 09:26:30 No matching credentials were found, falling back on anonymous

⚠️  Ignoring --vm-driver=virtualbox, as the existing "minikube" VM was created using the kvm2 driver.
⚠️  To switch drivers, you may create a new VM using `minikube start -p <name> --vm-driver=virtualbox`
⚠️  Alternatively, you may delete the existing VM using `minikube delete -p minikube`

🔄  Restarting existing kvm2 VM for "minikube" ...
⌛  Waiting for SSH access ...
🐳  Configuring environment for Kubernetes v1.14.2 on Docker 18.06.3-ce
💾  Downloading kubeadm v1.14.2
💾  Downloading kubelet v1.14.2
🚜  Pulling images ...
🔄  Relaunching Kubernetes v1.14.2 using kubeadm ... 
⌛  Verifying: apiserver proxy etcd scheduler controller dns
🏄  Done! kubectl is now configured to use "minikube"
[carroarmato0@neon-flower:~] $ minikube stop
✋  Stopping "minikube" in kvm2 ...
🛑  "minikube" stopped.
```

Deleting the cluster:

```
[carroarmato0@neon-flower:~] 70 $ minikube delete -v 10
Found binary path at /home/carroarmato0/.local/bin/docker-machine-driver-kvm2
Launching plugin server for driver kvm2
Plugin server listening at address 127.0.0.1:33831
[...]

PR #4344, opened 2019-05-24
Fix kvm remove when domain is not running

josedonizetti:
Fix for https://github.com/kubernetes/minikube/issues/4319

This PR is considering only the case when the running domain was deleted, with `virsh destroy DOMAIN_NAME`, and not the problem of undefining a domain, I'll create a separate PR for that to simplify review.

### DIFF
--- a/pkg/drivers/kvm/kvm.go
+++ b/pkg/drivers/kvm/kvm.go
@@ -461,7 +461,7 @@ func (d *Driver) Remove() error {
 func (d *Driver) destroyRunningDomain(dom *libvirt.Domain) error {
 	state, reason, err := dom.GetState()
 	if err != nil {
-		return err
+		return errors.Wrap(err, "getting domain state")
 	}
 
 	if state == libvirt.DOMAIN_SHUTOFF && reason == int(libvirt.DOMAIN_SHUTOFF_DESTROYED) {
--- a/pkg/drivers/kvm/kvm.go
+++ b/pkg/drivers/kvm/kvm.go
@@ -443,16 +443,31 @@ func (d *Driver) Remove() error {
 	dom, err := conn.LookupDomainByName(d.MachineName)
 	if err != nil {
 		log.Warnf("Domain %s does not exist, nothing to clean up...", d.MachineName)
+		return nil
 	}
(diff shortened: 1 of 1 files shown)


pkg/drivers/kvm/kvm.go
```diff
@@ -443,16 +443,31 @@ func (d *Driver) Remove() error {
...
-		if err := dom.Undefine(); err != nil {
-			return err
-		}
+
+	log.Infof("Domain %s exists, removing...", d.MachineName)
+	if err := d.destroyRunningDomain(dom); err != nil {
+		return errors.Wrap(err, "destroying running domain")
+	}
+
+	if err := dom.Undefine(); err != nil {
```
tstromberg: Since this function doesn't do anything after handling this error, just use "return dom.Undefine()".

pkg/drivers/kvm/kvm.go
```diff
@@ -443,16 +443,31 @@ func (d *Driver) Remove() error {
...
+		return err
 	}
 
 	return nil
 }
+
+func (d *Driver) destroyRunningDomain(dom *libvirt.Domain) error {
+	state, reason, err := dom.GetState()
+	if err != nil {
+		return err
```
tstromberg: Since this function returns two types of errors, do you mind using errors.Wrap() on at least one of them so we can tell from the error message which function caused the error to occur?

pkg/drivers/kvm/kvm.go
josedonizetti: hey, I haven't changed this because on the other upcoming PR I'm changing this code, should I still change it nonetheless?

pkg/drivers/kvm/kvm.go
josedonizetti: sure!

pkg/drivers/kvm/kvm.go
josedonizetti: ignore, it makes sense to change, so the commit itself is 100% no matter which code came later.

tstromberg:
@minikube-bot OK to test

PR #4355, opened 2019-05-25
Fix kvm remove when domain is not defined

josedonizetti:
This is the second part to fix for #4319

In case someone undefined the domain outside minikube, the deletion would fail, so we check if the domain is actually defined before callding `undefined` on it.


tstromberg:
@minikube-bot OK to test

Issue #4339, opened 2019-05-24
TestProxy/ConsoleWarnning fails with the none driver


tstromberg:
See https://storage.googleapis.com/minikube-builds/logs/4328/Linux-None.txt

```I0523 19:39:31.867653    3045 cluster.go:118] engine options: &{ArbitraryFlags:[] DNS:[] GraphDir: Env:[HTTP_PROXY=localhost:38685] Ipv6:false InsecureRegistry:[10.96.0.0/12] Labels:[] LogLevel: StorageDriver: SelinuxEnabled:false TLSVerify:false RegistryMirror:[] InstallURL:}
            E0523 19:42:31.881144    3045 start.go:529] StartHost: detecting provisioner: Too many retries waiting for SSH to be available.  Last error: Maximum number of retries (60) exceeded
            I0523 19:42:31.881204    3045 utils.go:122] non-retriable error: detecting provisioner: Too many retries waiting for SSH to be available.  Last error: Maximum number of retries (60) exceeded
            W0523 19:42:31.881310    3045 exit.go:100] Unable to start VM: detecting provisioner: Too many retries waiting for SSH to be available.  Last error: Maximum number of retries (60) exceeded
            
            X Unable to start VM: detecting provisioner: Too many retries waiting for SSH to be available.  Last error: Maximum number of retries (60) exceeded
```

I admittedly do not understand what's going on.

medyagh:
@tstromberg I can't replicate this on my linux box, do we have any recent PR with this failure ?

medyagh:
as far as the test logic, it is looking for a warning message to the user that they are using a proxy. but it minikube start doesn't get there, because it is stuck "Waiting for SSH to be available". and then it times out ... I could maybe try to increase the test time out

medyagh:
fixed by https://github.com/kubernetes/minikube/pull/4383

PR #4364, opened 2019-05-29
docker-env: Remove DOCKER_API_VERSION

### LINKED ISSUE #4363: DOCKER_API_VERSION does not match version in VM image
Currently we hardcode DOCKER_API_VERSION show to users who run `minikube docker-env`.

If a user starts an older VM image, we incorrectly advertise the DOCKER_API_VERSION it supports. We shouldn't hardcode the value.

The workaround we recommend at the moment is that users run `minikube delete` in order to create a newer VM version. We may also want to warn users if they start minikube from a VM where the Docker version is older than expected.

See related Skaffold bug: https://github.com/GoogleContainerTools/skaffold/issues/2187

The error users notice is:

`docker build: Error response from daemon: client version 1.39 is too new. Maximum supported API version is 1.38`

tstromberg:
This is no longer required, as Docker clients/servers have been able to auto-negotiate for many years now.

Closes #4363


tstromberg:
None test failure is unrelated issue: https://github.com/kubernetes/minikube/issues/4339

PR #4372, opened 2019-05-29
Increase apiserver wait time from 1 minute to 3 minutes

tstromberg:
This hopefully addresses #4366


tstromberg:
@minikube-bot OK to test

PR #4373, opened 2019-05-29
OWNERS and MAINTAINERS file cleanup

sharifelgamal:
Removing reviewers that are not longer active on minikube and adding new ones.


RA489:
@tstromberg @sharifelgamal Thank you very much for adding me into the approvers list.

PR #4381, opened 2019-05-30
Fix kvm gpu log

josedonizetti:
Fix log for kvm gpu.


sharifelgamal:
@minikube-bot OK to test

PR #4382, opened 2019-05-30
clean vendor on make clean

medyagh:
To fix the build issue when there is there a vendor folder conflicting with gomod


sharifelgamal:
The vendor directory should never exist if you merge master into your branch.

medyagh:
ah you are right, I guess this would be useful only to people who had checked pre-gomodule minikube  and had a global .gitignore for `vendor`.

PR #4383, opened 2019-05-30
Fix TestProxy 

medyagh:
To address this issue https://github.com/kubernetes/minikube/issues/4339. 
and making sure minikube is deleted before/after test  avoid #4132

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,6 +209,7 @@ checksum:
 clean:
 	rm -rf $(BUILD_DIR)
 	rm -f pkg/minikube/assets/assets.go
+	rm -rf ./vendor
 
 .PHONY: gendocs
 gendocs: out/docs/minikube.md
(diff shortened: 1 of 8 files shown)


tstromberg:
@minikube-bot OK to test

medyagh:
@minikube-bot OK to test

medyagh:
@minikube-bot OK to test

medyagh:
Turns out the issue for this test is related to https://github.com/kubernetes/minikube/issues/4132
the root cause is, when the None driver integration test starts an already started minikube, it fails with Waiting SSH ...

the root cause needs to be address in ttps://github.com/kubernetes/minikube/issues/4132 meanwhile we can make sure the TestProxy doesn't start an already started minikube for this test

medyagh:
@minikube-bot OK to test

Issue #4384, opened 2019-05-30
Addon manager presumes port 8443 -- fails when used with different --apiserver-port


moduspwnens:
If minikube is started with the --apiserver-port directive set to a non-default value, the addon-manager pod fails to connect to the apiserver.

The culprit seems to be a hardcoding of port 8443 here:
https://github.com/kubernetes/minikube/blob/8ba53c6b0b6487e212a86a07f58a6b43ee8cf6e5/pkg/minikube/bootstrapper/certs.go#L69

The fix is a one-liner--I'll submit a pull request.

tstromberg:
Oof. Good find! I'm looking forward to the PR for this.

PR #4386, opened 2019-05-30
Fixed addon-manager failing with non-default --apiserver-port

moduspwnens:
This addresses issue #4384.


tstromberg:
@minikube-bot OK to test

Issue #4359, opened 2019-05-27
Segmentation fault when starting dashboard using nonexistent profile


bitfactory-henno-schooljan:
When I try to launch the Kubernetes dashboard using a profile that does not exist, the process segfaults. This also happens without the -p option when no profiles do not yet exist (e.g. after a fresh install) and it tries to use the default `minikube` one. The expected response should be that the profile does not exist, which does happen correctly with a command like `minikube ip`.
 
**The exact command to reproduce the issue**:
```
minikube dashboard -p nonexistent
```

**The full output of the command that failed**:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x178 pc=0x1e00e9c]

goroutine 1 [running]:
k8s.io/minikube/cmd/minikube/cmd.glob..func5(0x2e828c0, 0xc000173520, 0x0, 0x2)
	/app/cmd/minikube/cmd/dashboard.go:62 +0x12c
github.com/spf13/cobra.(*Command).execute(0x2e828c0, 0xc0001734c0, 0x2, 0x2, 0x2e828c0, 0xc0001734c0)
	/go/pkg/mod/github.com/spf13/cobra@v0.0.0-20180228053838-6644d46b81fa/command.go:757 +0x2ae
github.com/spf13/cobra.(*Command).ExecuteC(0x2e83960, 0x1, 0x1ee3ea0, 0xc0000cf948)
	/go/pkg/mod/github.com/spf13/cobra@v0.0.0-20180228053838-6644d46b81fa/command.go:843 +0x2ec
github.com/spf13/cobra.(*Command).Execute(...)
	/go/pkg/mod/github.com/spf13/cobra@v0.0.0-20180228053838-6644d46b81fa/command.go:791
k8s.io/minikube/cmd/minikube/cmd.Execute()
	/app/cmd/minikube/cmd/root.go:103 +0x2e
main.main()
	/app/cmd/minikube/main.go:44 +0xcf
```

**The output of the `minikube logs` command**:
No logs seem to get written for this session, process crashes.

**The operating system version**:
macOS 10.14.5
minikube 1.1.0
VirtualBox 6.0.4

tstromberg:
Thanks for the bug report! Based on the backtrace, this is the line that crashed:

https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/dashboard.go#L62

[...]

PR #4396, opened 2019-05-31
dashboard: detect nonexistent profile instead of causing a panic

fenglixa:
Fix for #4359

My UT result could see the picture:

image

### DIFF
--- a/cmd/minikube/cmd/dashboard.go
+++ b/cmd/minikube/cmd/dashboard.go
@@ -72,12 +72,11 @@ var dashboardCmd = &cobra.Command{
 			exit.WithError("Error getting client", err)
 		}
 
-		_, err = api.Load(pkg_config.GetMachineName())
-		if err != nil {
+		if _, err = api.Load(pkg_config.GetMachineName()); err != nil {
 			switch err := errors.Cause(err).(type) {
 			case mcnerror.ErrHostDoesNotExist:
 				console.OutStyle(console.Meh, "%q cluster does not exist", pkg_config.GetMachineName())
-				os.Exit(0)
+				os.Exit(exit.Unavailable)
 			default:
 				exit.WithError("Error getting cluster", err)
 			}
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/dashboard.go
```diff
@@ -79,6 +71,28 @@ var dashboardCmd = &cobra.Command{
 		if err != nil {
 			exit.WithError("Error getting client", err)
 		}
+
+		_, err = api.Load(pkg_config.GetMachineName())
+		if err != nil {
+			switch err := errors.Cause(err).(type) {
+			case mcnerror.ErrHostDoesNotExist:
+				console.OutStyle(console.Meh, "%q cluster does not exist", pkg_config.GetMachineName())
+				os.Exit(0)
```
tstromberg: I think this should still be a non-zero exit. Perhaps exit.Unavailable?

cmd/minikube/cmd/dashboard.go
```diff
@@ -79,6 +71,28 @@ var dashboardCmd = &cobra.Command{
 		if err != nil {
 			exit.WithError("Error getting client", err)
 		}
+
+		_, err = api.Load(pkg_config.GetMachineName())
+		if err != nil {
```
tstromberg: Join these two expressions:

`if _, err = api.Load(pkg_config.GetMachineName()); err != nil {`

tstromberg:
Thank you for doing this! This is a much better user experience.

cmd/minikube/cmd/dashboard.go
fenglixa: Thanks @tstromberg, done.

cmd/minikube/cmd/dashboard.go
fenglixa: Thanks @tstromberg, done. New UT result:
image

medyagh:
@minikube-bot OK to test

sharifelgamal:
@minikube-bot OK to test

Issue #4387, opened 2019-05-30
Add `--version` to kvm2 driver


josedonizetti:
Lots of the issues related to kvm2 driver are due to old versions of the driver. When a user installs minikube for the first time, the driver is installed by hand, and as the time passes the user only remembers to upgrade minikube, but not the driver itself.

By having a `--version` on the driver it would simplify the communication on reported issues because we would be able to know exactly what version of the driver is running. Also, we could use such a mechanism in minikube to report an error for the user to upgrade the driver.

I'm willing to work on this. What do you think? Does it make sense?

tstromberg:
Yes, this would be great.

Would it be too much trouble to add it for hyperkit as well? It suffers from the same problem.

josedonizetti:
@tstromberg No trouble at all. I’ll add it to both. Are we using specific version for it? Like 0.0.1? Or basing then o minikube version?

tstromberg:
I recommend starting by basing the version on what minikube is at.

Eventually we will push the hyperkit driver back upstream in #3939 - but upstream is currently at v1.0.0 and we shouldn't have a problem preserving the version number when we eventually make that happen.

PR #4398, opened 2019-05-31
Add kvm2 --version

josedonizetti:
FIx https://github.com/kubernetes/minikube/issues/4387

Considerations:

- I've added a version.go to the kvm driver instead of importing `minikube/version`
- As we are only supporting `--version`, I didn't use cobra/viper

Let me know what you think. I'll add it to hyperkit, once this one was approved.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -366,7 +366,7 @@ release-minikube: out/minikube checksum
 out/docker-machine-driver-kvm2:
 	go build 																		\
 		-installsuffix "static" 													\
-		-ldflags "-X k8s.io/minikube/pkg/drivers/kvm/version.VERSION=$(VERSION)" 	\
+		-ldflags "-X k8s.io/minikube/pkg/drivers/kvm.version=$(VERSION)" 	\
 		-tags libvirt.1.3.1 														\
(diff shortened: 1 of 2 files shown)


medyagh:
@minikube-bot OK test

medyagh:
@josedonizetti thank for your contributions, would you please add documentation and integration tests for it ? and how one would use this --version ?

josedonizetti:
@medyagh The usage will be `docker-machine-driver-kvm2 --version`. The idea is to simplify issues communication about the kvm2 driver, the user would have a simple way to know the driver is outdated, and probably the reason it's currently failing. I'll add the documentation and tests. Thanks

blueelvis:
@medyagh @josedonizetti  I think it should be `--kvm-version` instead of just `--version`. That way, people don't get confused if that is the minikube's version or the kvm driver version. What do you think?

josedonizetti:
@blueelvis I think the command `docker-machine-driver-kvm2 --version`  is very explanatory that is not the same as `minikube --version`. Yet, no strong feelings about it, let's see what others think. :)

josedonizetti:
@medyagh Hey, have any pointers on adding integration tests for this?

Issue #4367, opened 2019-05-29
`minikube dashboard` > "Settings" gives 403 forbidden


Doqnach:
**The exact command to reproduce the issue**:

1. `minikube dashboard`
2. In the left menu, navigate to "Settings"

**The full output of the command that failed**:

![image](https://user-images.githubusercontent.com/25537191/58537432-2b991700-81f3-11e9-8621-e0fb2a500aa2.png)

**The operating system version**:

MacOS 10.14.5, minikube v1.1.0, cluster kubernetes v1.11.8, hyperkit

Possibly related? https://github.com/kubernetes/dashboard/issues/2787

medyagh:
are you using a HTTP_PROXY or HTTPS_PROXY ?

medyagh:
could you please paste the output of `minikube start` and also `minikube logs`

tstromberg:
This sounds like https://github.com/kubernetes/dashboard/issues/2694 and/or https://github.com/kubernetes/dashboard/issues/2787

One workaround may be to `--disable-settings-authorizor=true` to the dashboard command line. If anyone tries this and it works, feel free to send a PR out.

fenglixa:
I would like to verify it and provide PR for this ticket. Thanks

Doqnach:
This has been released with v1.1.1, so can be closed as fixed.

PR #4405, opened 2019-06-02
Dashboard: add --disable-settings-authorizer to avoid settings 403 forbidden

fenglixa:
- Fix for #4367

- My UT Results:

Before code change, reproduced the issue: 
image

Code changed, the issue is gone:
image
image

- Build & Test PASS:
image


RA489:
@minikube-bot OK to test

PR #4407, opened 2019-06-02
Stop updating /etc/rkt/net.d config files

afbjorklund:
We have now removed rkt, should be enough with cni


RA489:
@minikube-bot OK to test

Issue #4371, opened 2019-05-29
k8s v1.15-beta1: kubelet: unknown flag: --allow-privileged


tstromberg:
If `minikube start --kubernetes-version=v1.15.0-beta.1` is used, a failure occurs:

```

Unfortunately, an error has occurred:
	timed out waiting for the condition

This error is likely caused by:
	- The kubelet is not running
	- The kubelet is unhealthy due to a misconfiguration of the node in some way (required cgroups disabled)

If you are on a systemd-powered system, you can try to troubleshoot the error with the following commands:
	- 'systemctl status kubelet'
	- 'journalctl -xeu kubelet'

Additionally, a control plane component may have crashed or exited when started by the container runtime.
To troubleshoot, list all containers using your preferred container runtimes CLI, e.g. docker.
Here is one example how you may list all Kubernetes containers running in docker:
	- 'docker ps -a | grep kube | grep -v pause'
	Once you have found the failing container, you can inspect its logs with:
	- 'docker logs CONTAINERID'

: Process exited with status 1

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
👉  https://github.com/kubernetes/minikube/issues/new
❌  Problems detected in "kubelet":
    May 29 18:23:32 k8s115 kubelet[2944]: F0529 18:23:32.984077    2944 server.go:156] unknown flag: --allow-privileged
```

tstromberg:
Deprecation notice for this flag: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#action-required

If anyone wants to fix this, here is where to get started:

https://github.com/kubernetes/minikube/blob/6bc6c3fe876f55534360bc1a8aa30e1722f388ce/pkg/minikube/bootstrapper/kubeadm/versions.go#L245

If Kubernetes v1.15.0-beta1 is working properly, the PR should also update NewestKubernetesVersion so that this release is tested in our integration tests:

https://github.com/kubernetes/minikube/blob/b64f62ccfb5ec34897540b1721fa3c89b5c8c5b5/pkg/minikube/constants/constants.go#L168

blueelvis:
[...]

PR #4412, opened 2019-06-03
minikube support for k8s v1.15

fenglixa:
- Fix for #4371

- Reproduced the issue, before my modifying:
image

- My UT result PASS:

1.  minikube start for k8s v1.15, and check kubelet args:
image

2. minikube start for k8s v1.14.2, and check kubelet args:
image

- Build & Test PASS:
image

### DIFF
--- a/pkg/minikube/bootstrapper/kubeadm/versions.go
+++ b/pkg/minikube/bootstrapper/kubeadm/versions.go
@@ -242,7 +242,14 @@ var versionSpecificOpts = []VersionedExtraOption{
 
 	// System pods args
 	NewUnversionedOption(Kubelet, "pod-manifest-path", "/etc/kubernetes/manifests"),
-	NewUnversionedOption(Kubelet, "allow-privileged", "true"),
+	{
+		Option: util.ExtraOption{
+			Component: Kubelet,
+			Key:       "allow-privileged",
+			Value:     "true",
+		},
+		LessThanOrEqual: semver.MustParse("1.15.0-alpha.3"),
+	},
 
 	// Network args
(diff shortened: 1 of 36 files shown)


pkg/minikube/bootstrapper/kubeadm/versions.go
```diff
@@ -242,7 +242,14 @@ var versionSpecificOpts = []VersionedExtraOption{
...
 	// System pods args
 	NewUnversionedOption(Kubelet, "pod-manifest-path", "/etc/kubernetes/manifests"),
-	NewUnversionedOption(Kubelet, "allow-privileged", "true"),
+	{
+		Option: util.ExtraOption{
+			Component: Kubelet,
+			Key:       "allow-privileged",
+			Value:     "true",
+		},
+		LessThanOrEqual: semver.MustParse("1.14.2"),
```
afbjorklund: Is this still needed in 1.14.3 ? Maybe this should be < 1.15 ?

afbjorklund:
We shouldn't merge this before 1.2.0 (that will have 1.14.3), and probably not before 1.15 is out properly.

fenglixa:
Sorry, squashed error, close this PR. I will create a new one later

pkg/minikube/bootstrapper/kubeadm/versions.go
fenglixa: @afbjorklund I checked change logs of v1.15.0-beta.1. The deprecated Kubelet flag --allow-privileged has been removed since v1.15.0-alpha.3. So here should change to <= 1.15.0-alpha.3.
Sorry for my mistake, and thanks for your reminder.

PR #4419, opened 2019-06-04
Improve type check for driver none

medyagh:


pkg/minikube/cluster/cluster.go
```diff
@@ -72,7 +72,7 @@ func init() {
 
 // CacheISO downloads and caches ISO.
 func CacheISO(config cfg.MachineConfig) error {
-	if config.VMDriver != "none" {
+	if config.VMDriver == constants.DriverNone {
```
afbjorklund: This logic looks like it has been reversed ?

medyagh: ooops good catch. thanks for catching it

sharifelgamal:
@minikube-bot OK to test

pkg/minikube/cluster/cluster.go
medyagh: Done

PR #4410, opened 2019-06-02
Upgrade CNI config version to 0.3.0

afbjorklund:
Was using the default version of 0.2.0, didn't load with 0.4.0

For #4406

PR #4421, opened 2019-06-04
Enable running containers with Podman

### LINKED ISSUE #4406: Running containers with podman doesn't work
We install the `podman` tool into the VM, to do things such as loading or building OCI images.

But currently there are some issues with the CNI configuration, if you try to use it for _running_:

``` console
$ minikube ssh
                         _             _            
            _         _ ( )           ( )           

$ sudo podman run -it busybox
Trying to pull docker.io/library/busybox...Getting image source signatures
Copying blob 53071b97a884 done
Copying config 64f5d945ef done
Writing manifest to image destination
Storing signatures
Error: error parsing CNI plugin result "IP4:{IP:{IP:10.1.0.7 Mask:ffff0000} Gateway:10.1.0.1 Routes:[{Dst:{IP:0.0.0.0 Mask:00000000} GW:10.1.0.1}]}, DNS:{Nameservers:[] Domain: Search:[] Options:[]}": cannot convert version ["" "0.1.0" "0.2.0"] to 0.4.0: cannot convert version ["" "0.1.0" "0.2.0"] to 0.4.0
```

Even though it is not the **primary** use-case (normally use crictl/crio), this should work...

Maybe we can get the standard minikube CNI configuration compatible with Podman ?

afbjorklund:
Running containers with `sudo podman` did not work properly.

1. ~~The CNI version used for crio is too old, so it fails to load from podman (PR #4410)~~
2. We need to upgrade conmon, i.e. provide /usr/libexec/podman/conmon
3. We need to rebuild runc, systemd support doesn't work when built static
4. We need to provide /etc/containers/libpod.conf, with our rootfs setting

Closes #4406


sharifelgamal:
@minikube-bot OK to test

tstromberg:
PR looks good, but needs a better title for our release notes.

afbjorklund:
Hopefully this title is better ? Also reopened the issue, that got closed prematurely.

afbjorklund:
Planning to upgrade from 1.3.1 to 1.4.1, but want to get this landed first.

Maybe the `conmon-master` version needs to be bumped to `v0.3.0`, too.

PR #4431, opened 2019-06-05
fix ShowDriverDeprecationNotification config setting

### LINKED ISSUE #4161: xhyve suggests obsolete WantShowDriverDeprecationNotification setting
minikube v1.0.0

I use `minikube start --vm-driver xhyve`

which tells me

```
⚠️  The xhyve driver is deprecated and support for it will be removed in a future release.
Please consider switching to the hyperkit driver, which is intended to replace the xhyve driver.
See https://github.com/kubernetes/minikube/blob/master/docs/drivers.md#hyperkit-driver for more information.
To disable this message, run [minikube config set WantShowDriverDeprecationNotification false]
```

but 
```
$ minikube config set WantShowDriverDeprecationNotification false

💣  Set failed: Property name WantShowDriverDeprecationNotification not found
```

Pradip-Khakurel:
fix #4161:
WantShowDriverDeprecationNotification property seems like a mistype in the console message. The real setting is ShowDriverDeprecationNotification but it needs to be added in the list of configurable settings (the variable settings in config.go). I could have renamed everything to WantShowDriverDeprecationNotification but it seems easier and still very readable that way.

User test:

### DIFF
--- a/cmd/minikube/cmd/config/config.go
+++ b/cmd/minikube/cmd/config/config.go
@@ -132,6 +132,10 @@ var settings = []Setting{
 		set:  SetString, //TODO(r2d4): more validation here?
 	},
 	{
+		name: config.ShowDriverDeprecationNotification,
+		set:  SetBool,
+	},
+	{
 		name: config.ShowBootstrapperDeprecationNotification,
 		set:  SetBool,
 	},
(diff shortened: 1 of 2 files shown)


afbjorklund:
@fenglixa : removing the deprecated drivers is a different topic

sharifelgamal:
LGTM. We'll merge this once tests pass.

PR #4465, opened 2019-06-10
Automatically add extra options  for none driver on ubuntu

medyagh:
To fix this issue, https://github.com/kubernetes/minikube/issues/3511

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -350,7 +350,7 @@ func validateConfig() {
 		exit.Usage("Sorry, the --hidden feature is currently only supported with --vm-driver=kvm2")
 	}
 
-	err := autoSetOptions()
+	err := autoSetOptions(viper.GetString(vmDriver))
 	if err != nil {
 		glog.Errorf("Error autoSetOptions : %v", err)
 	}
@@ -492,9 +492,9 @@ func generateConfig(cmd *cobra.Command, k8sVersion string) (cfg.Config, error) {
 }
 
 // autoSetOptions sets the options needed for specific configuration automatically.
-func autoSetOptions() error {
+func autoSetOptions(vmDriver string) error {
 	//  options for none driver
-	if viper.GetString(vmDriver) == constants.DriverNone {
+	if vmDriver == constants.DriverNone {
 		if o := autoOptionForNone(); o != "" {
 			return extraOptions.Set(o)
 		}
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -350,6 +350,17 @@ func validateConfig() {
 		exit.Usage("Sorry, the --hidden feature is currently only supported with --vm-driver=kvm2")
 	}
 
+	// Adding extra option (resolv-conf) for none driver on ubutnu to address /issues/3511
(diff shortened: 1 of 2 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -486,6 +497,16 @@ func generateConfig(cmd *cobra.Command, k8sVersion string) (cfg.Config, error) {
 	return cfg, nil
 }
 
+// noneNeedsResolveConf checks extra config option for resolv.conf is needed
+// for none driver. more info https://github.com/kubernetes/minikube/issues/3511
+func noneNeedsResolveConf() bool {
+	filename := "/run/systemd/resolve/resolv.conf"
```
tstromberg: Could this be refactored in a way that doesn't require this filename to be duplicated by the caller?

Perhaps this function could return the extra flag required (kubelet.resolv-conf=xxx), or the value could be shared as a const?

cmd/minikube/cmd/start.go
```diff
@@ -350,6 +350,17 @@ func validateConfig() {
 		exit.Usage("Sorry, the --hidden feature is currently only supported with --vm-driver=kvm2")
 	}
 
+	// Adding extra option (resolv-conf) for none driver on ubutnu to address /issues/3511
+	if viper.GetString(vmDriver) == constants.DriverNone && noneNeedsResolveConf() {
+		o := "kubelet.resolv-conf=/run/systemd/resolve/resolv.conf"
+		err := extraOptions.Set(o)
+		if err == nil {
+			glog.Infof("Sucsessfully set extra option (%s) for none driver on ubuntu-like OS.", o)
```
tstromberg: Spelling: "Successfully". Also, no need to mention ubuntu-like OS. If you need to, it should be capitalized: "Ubuntu".

cmd/minikube/cmd/start.go
```diff
@@ -350,6 +350,17 @@ func validateConfig() {
...
+		} else {
+			glog.Errorf("Error setting extra options (%s) for none driver on ubuntu-like OS: %s", o, err)
```
tstromberg: Shorten if possible. Just: "Error setting options %q: %v"

error types should always be %v instead of %s.

medyagh:
@minikube-bot okay to test.

cmd/minikube/cmd/start.go
```diff
@@ -486,6 +497,16 @@ func generateConfig(cmd *cobra.Command, k8sVersion string) (cfg.Config, error) {
 	return cfg, nil
 }
 
+// noneNeedsResolveConf checks extra config option for resolv.conf is needed
+// for none driver. more info https://github.com/kubernetes/minikube/issues/3511
+func noneNeedsResolveConf() bool {
+	filename := "/run/systemd/resolve/resolv.conf"
+	if _, err := os.Stat(filename); os.IsNotExist(err) {
```
tstromberg: optional: one could argue if there is any error calling os.Stat(), that we shouldn't use this file. I think you can safely remove the os.IsNotExist check.

Alternatively, you could also check for other more exceptional errors and log them.

medyagh:
@minikube-bot okay to test

medyagh:
DNS test failure is related to this issue : https://github.com/kubernetes/minikube/issues/4475

RA489:
@medyagh Pleash squash the commits to one rest LGTM

PR #4497, opened 2019-06-15
WIP: Remove the requirement of a shell to run commands with ExecRunner

### LINKED ISSUE #4122: none driver doesn't start if /bin/bash doesn't exist.
Command:

```
minikube --logtostderr start -p sentinel --vm-driver none
```

Output: https://gist.github.com/alunduil/1e78933095fc69402ec0dc2e0927724f
Minikube logs: https://gist.github.com/alunduil/a05cc1071bb0cf709f875817198d20f8
My os-release: https://gist.github.com/alunduil/f49e9aa43b826de82011eada8dc4cf75

In NixOS, there is no /bin/bash so starting the none driver fails to function.  A simple fix for this error would be to use /usr/bin/env to invoke bash by following the path rather than assuming that it's hard coded, but I'm not sure what other assumptions that might break.  This solution should work because NixOS ensures a symlink exists at /usr/bin/env that points at the executable and most distributions that I'm aware of ship with /usr/bin/env.

If that sounds like a reasonable solution, I can work on figuring out a pull request (unless someone can get to it faster).

palnabarun:
Fixes #4122 

This PR addresses the issue of forking an extra shell when executing commands. This also partly solves the dependency on `bash`. 

#### Changes Made
- Instead of using `/bin/bash` to run the command given to `ExecRunner.Run`, now `ExecRunner.Run` splits the command string into the command and a list of arguments.
- Broke down a command with `&&` operator in `pkg/minikube/bootstrapper/kubeadm/kubeadm.go` into the two constituent commands

#### Testing

I used an environment where `bash` is not available at `/bin/bash` but is available in the `PATH` elsewhere, in my case `$HOME/.local/bin`

#### Limitations:

- `ExecRunner.Run` can't run command strings of the format. Eg: `touch main.go && vi main.go`

#### Further Work:
- After I implemented `splitCommandString`, I found about [shlex](https://github.com/google/shlex) which also handles comments and quotes in the command string. I didn't add this since I was unsure about how a new dependency is added to minikube source code.


afbjorklund:
Is this about optimizing minikube to avoid forking an extra shell, or about support alternate shells ?

It seems to be a little of both, which is making it harder to follow. It started out as `#!/usr/bin/env bash`

palnabarun:
@afbjorklund Sorry if I made it too confusing. It is related to avoid forking an extra shell.

Ref: https://github.com/kubernetes/minikube/issues/4122#issuecomment-485463877

palnabarun:
Adding to that, this PR also partly solves the dependency on `bash`.

pkg/minikube/bootstrapper/exec_runner.go
```diff
@@ -34,10 +35,31 @@ import (
 // It implements the CommandRunner interface.
 type ExecRunner struct{}
 
+// splitCommandString splits a command string like `ls -l` into the command `ls` and `-l`
+func splitCommandString(cmd string) (string, []string) {
```
tstromberg: I'm not yet convinced that this will work without a lot of extra hacks, but https://godoc.org/github.com/google/shlex will work better than implementing our own splitter, as it properly handles quotes.

tstromberg:
@minikube-bot OK to test

pkg/minikube/bootstrapper/exec_runner.go
palnabarun: I agree. `shlex` would be better. I found about that after committing the changes in this PR.

I am going ahead and implementing the same using shlex

medyagh:
@palnabarun I was wondering if you are still working on this ?

palnabarun:
@medyagh 

Yes, I am still working on this. I was out for a few days.

There is a major bottleneck in the approach we are taking here, Ability to run commands with pipes and control operators.

For example, in the following line, there exists a control operator `&&`.

https://github.com/kubernetes/minikube/blob/124b845b601abd52e16a8d723c976a4dcd2bd77f/pkg/minikube/logs/logs.go#L117

This specific instance though is very simple to solve, because we can just split that into two different commands.

However, there is a more complex case where we get the kernel related messages.

https://github.com/kubernetes/minikube/blob/124b845b601abd52e16a8d723c976a4dcd2bd77f/pkg/minikube/bootstrapper/kubeadm/kubeadm.go#L154-L176

I am yet to find a suitable solution for the above case where the command contains pipes. In the above case, the command string gets parsed to the following:

```
Command String: sudo dmesg -PH -L=never --level warn,err,crit,alert,emerg | tail -n 50
Command: sudo
Full Path of command: /usr/bin/sudo
Arguments: ["dmesg", "-PH", "-L=never", "--level", "warn,err,crit,alert,emerg", "|", "tail", "-n", "50"]
```

I was thinking of another possible solution though. Why not just use `/bin/sh` instead of `/bin/bash`? Because, in any system, `sh` is expected to be present and generally links to another shell. For example, in Arch Linux it points to `bash`.

PS: I am pushing the changes where I am using `shlex`

medyagh:
there is a CI test error for unused, error.

palnabarun:
@medyagh Sorry! Forgot to push that change.

medyagh:
I saw this error in none driver integration test for this PR

--- FAIL: TestFunctional (108.60s)
    util.go:248: got state: "Command String: sudo systemctl is-active kubelet, Command: sudo, Full Path: /usr/bin/sudo, Arguments: [systemctl is-active kubelet] \nRunning", expected "Running"
which happens here:

it seems like when it failed to split the output currently and figure out it is running.
https://github.com/kubernetes/minikube/blob/master/test/integration/util/util.go#L237

palnabarun:
> ```
> --- FAIL: TestFunctional (108.60s)
>     util.go:248: got state: "Command String: sudo systemctl is-active kubelet, Command: sudo, Full Path: /usr/bin/sudo, Arguments: [systemctl is-active kubelet] \nRunning", expected "Running"
> which happens here:
> ```

This error was due to accidentally committed debug code. I should have caught this early on. The unit tests passed with the change when running `make test`.

I am trying to figure out how to run the integration tests locally.

afbjorklund:
Still confused...

> Sorry if I made it too confusing. **It is related to avoid forking an extra shell.**

> Why not just use /bin/sh instead of /bin/bash?

I think we already discussed the handful of systems that *don't* have /bin/bash.
Basically it was only NixOS, and using lookpath is probably an easier fix (#4122) ?

The approach here looks complicated and error-prone.

medyagh:
integration test failed for none-driver:

```
           E0702 16:57:15.267193       1 controller.go:682] Error watching for provisioning success, can't provision for claim "default/testpvc": events is forbidden: User "system:serviceaccount:kube-system:storage-provisioner" cannot list resource "events" in API group "" in the namespace "default"

```

palnabarun:
> I think we already discussed the handful of systems that _don't_ have /bin/bash.
> Basically it was only NixOS, and using lookpath is probably an easier fix (#4122) ?

The `LookPath` approach won't work for commands having pipes or control operators. The approach taken here is the `LookPath` one itself with a bit of fiddling around to get the actual command to run.

For example, let us try to see what would happen if we try to run `sudo systemctl start kubelet`.

1. Parse the command string to get the command and the arguments to pass to the command.
Command: `sudo`
Arguments: `[systemctl start kubelet]`
[Reference](https://github.com/palnabarun/minikube/blob/900a8af0ebfc1063ae8fd31390de7740047101cd/pkg/minikube/bootstrapper/exec_runner.go#L43-L49)

2. We do an `exec.LookPath` for the command (`sudo`) and get back the full path to the command. In this case, `/usr/bin/sudo`.
[Reference](https://github.com/palnabarun/minikube/blob/900a8af0ebfc1063ae8fd31390de7740047101cd/pkg/minikube/bootstrapper/exec_runner.go#L55-L58)

3. Then we execute `exec.Command` with the arguments `/usr/bin/sudo` and `[systemctl start kubelet]...`.
[Reference](https://github.com/palnabarun/minikube/blob/900a8af0ebfc1063ae8fd31390de7740047101cd/pkg/minikube/bootstrapper/exec_runner.go#L60-L64)

Now, a question would arise, why are we using `shlex` at all.

This is needed because if we don't split the command string into command and arguments, `exec.LookPath` would fail to resolve the absolute path for the command string. I wrote a sample demonstration [here](https://paste.fedoraproject.org/paste/4BW1ziDchU8MmoYyI1ZZBw).

I also mentioned caveats of the `exec.LookPath` in the comment [above](https://github.com/kubernetes/minikube/pull/4497#issuecomment-507660095)

cc: @afbjorklund

palnabarun:
> integration test failed for none-driver:
> 
> ```
>            E0702 16:57:15.267193       1 controller.go:682] Error watching for provisioning success, can't provision for claim "default/testpvc": events is forbidden: User "system:serviceaccount:kube-system:storage-provisioner" cannot list resource "events" in API group "" in the namespace "default"
> ```

@medyagh Thank you for pointing that out. I am looking into it why that might have happened.

On a different note, I compiled a list of errors occurring due to `exec_runner`
```
...
 16:55:39 | ! : running command: cat /proc/$(pgrep kube-apiserver)/oom_adj
 16:55:39 | ! .: exit status 1
...
 16:57:15 | ! E0702 16:57:15.900201     575 logs.go:134] failed: running command: sudo dmesg -PH -L=never --level warn,err,crit,alert,emerg | tail -n 50
 16:57:15 | ! .: exit status 1
 16:57:15 | ! I0702 16:57:15.900287     575 exec_runner.go:70] Run with output: uptime && uname -a
...
 16:57:15 | ! E0702 16:57:15.903248     575 logs.go:134] failed: running command: uptime && uname -a
 16:57:15 | ! .: exit status 1
...
 16:58:28 | ! : running command: cat /proc/$(pgrep kube-apiserver)/oom_adj
 16:58:28 | ! .: exit status 1
...
 17:02:37 | ! : running command: cat /proc/$(pgrep kube-apiserver)/oom_adj
 17:02:37 | ! .: exit status 1
...
```

afbjorklund:
> The `LookPath` approach won't work for commands having pipes or control operators.

I mean't LookPath for `bash` (instead of /bin/bash). I **don't** want to have a shell parser in minikube.

RA489:
@palnabarun needs rebase.

afbjorklund:
This PR implements something that hasn't been requested. I would actually prefer if it was closed.
> This PR addresses the issue of forking an extra shell when executing commands.

The original issue was just about lacking `/bin/bash`, and going to `/usr/bin/env bash` if missing.
> In NixOS, there is no /bin/bash so starting the none driver fails to function.

palnabarun:
Sorry, I haven't been able to take a look at this for some time.

@afbjorklund Thank you for the guidance. I agree with the suggestions provided by you. The problem was misunderstood by me. I will file a new PR with the relevant changes.

Closing this PR.

afbjorklund:
Thank you for your effort!

PR #4510, opened 2019-06-17
Add 'native-ssh' flag to 'minikube start' and 'minikube ssh'

yinzara:
Docker machine gives an option for sshing into a running container of using the Golang SSH native client or using the external 'ssh' command line executable.

Normally, "docker-machine ssh" uses the "external" client by default and a "--native-ssh" flag must be passed to enable the native client.

Minikube by set the client to native by default with no option to change this behavior.  This causes issues with some VM drivers.

Though the native SSH client for Golang is more efficient than the command line 'ssh' command, there are times in which using the 'ssh' CLI is necessary.

This PR adds a "--no-native-ssh" flag to the `minikube start` command an and associated "no-native-ssh" config option (through `minikube config set no-native-ssh true`.

### DIFF
--- a/cmd/minikube/cmd/config/config.go
+++ b/cmd/minikube/cmd/config/config.go
@@ -265,6 +265,10 @@ var settings = []Setting{
 		name: "embed-certs",
 		set:  SetBool,
 	},
+	{
+		name: "native-ssh",
+		set:  SetBool,
+	},
 }
 
 // ConfigCmd represents the config command
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -35,6 +35,7 @@ import (
 	"github.com/blang/semver"
 	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/host"
+	"github.com/docker/machine/libmachine/ssh"
 	"github.com/golang/glog"
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
@@ -107,6 +108,7 @@ const (
 	waitUntilHealthy      = "wait"
 	force                 = "force"
 	waitTimeout           = "wait-timeout"
+	nativeSSH             = "native-ssh"
 )
 
 var (
@@ -157,6 +159,7 @@ func initMinikubeFlags() {
 	startCmd.Flags().Bool(enableDefaultCNI, false, "Enable the default CNI plugin (/etc/cni/net.d/k8s.conf). Used in conjunction with \"--network-plugin=cni\".")
 	startCmd.Flags().Bool(waitUntilHealthy, true, "Wait until Kubernetes core services are healthy before exiting.")
 	startCmd.Flags().Duration(waitTimeout, 3*time.Minute, "max time to wait per Kubernetes core services to be healthy.")
+	startCmd.Flags().Bool(nativeSSH, true, "Use native Golang SSH client (default true). Set to 'false' to use the command line 'ssh' command when accessing the docker machine. Useful for the machine drivers when they will not start with 'Waiting for SSH'.")
 }
 
 // initKubernetesFlags inits the commandline flags for kubernetes related options
@@ -298,6 +301,12 @@ func runStart(cmd *cobra.Command, args []string) {
 
 	// With "none", images are persistently stored in Docker, so internal caching isn't necessary.
 	skipCache(&config)
+
+	if viper.GetBool(nativeSSH) {
+		ssh.SetDefaultClient(ssh.Native)
+	} else {
+		ssh.SetDefaultClient(ssh.External)
+	}
 
 	// Now that the ISO is downloaded, pull images in the background while the VM boots.
 	var cacheGroup errgroup.Group
--- a/cmd/minikube/cmd/ssh.go
+++ b/cmd/minikube/cmd/ssh.go
@@ -19,7 +19,10 @@ package cmd
 import (
 	"os"
 
+	"github.com/docker/machine/libmachine/ssh"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
 	"k8s.io/minikube/pkg/minikube/cluster"
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/constants"
@@ -46,6 +49,12 @@ var sshCmd = &cobra.Command{
 		if host.Driver.DriverName() == constants.DriverNone {
 			exit.UsageT("'none' driver does not support 'minikube ssh' command")
 		}
+		if viper.GetBool(nativeSSH) {
+			ssh.SetDefaultClient(ssh.Native)
+		} else {
+			ssh.SetDefaultClient(ssh.External)
+		}
+
 		err = cluster.CreateSSHShell(api, args)
 		if err != nil {
 			// This is typically due to a non-zero exit code, so no need for flourish.
@@ -54,4 +63,8 @@ var sshCmd = &cobra.Command{
 			os.Exit(exit.Failure)
 		}
 	},
+}
+
+func init() {
+	sshCmd.Flags().Bool(nativeSSH, true, "Use native Golang SSH client (default true). Set to 'false' to use the command line 'ssh' command when accessing the docker machine. Useful for the machine drivers when they will not start with 'Waiting for SSH'.")
 }
--- a/go.sum
+++ b/go.sum
@@ -8,7 +8,6 @@ dmitri.shuralyov.com/app/changes v0.0.0-20180602232624-0a106ad413e3/go.mod h1:Yl
 dmitri.shuralyov.com/html/belt v0.0.0-20180602232347-f7d459c86be0/go.mod h1:JLBrvjyP0v+ecvNYvCpyZgu5/xkfAUhi6wJj28eUfSU=
 dmitri.shuralyov.com/service/change v0.0.0-20181023043359-a85b471d5412/go.mod h1:a1inKt/atXimZ4Mv927x+r7UpyzRUf4emIoiiSC2TN4=
 dmitri.shuralyov.com/state v0.0.0-20180228185332-28bcc343414c/go.mod h1:0PRwlb0D6DFvNNtx+9ybjezNCa8XF0xaYcETyp6rHWU=
-github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999/go.mod h1:cp2SuWMxlEZw2r+iP2GNCdIi4C1qmUzdZFSVb+bacwQ=
 github.com/Azure/azure-sdk-for-go v21.4.0+incompatible/go.mod h1:9XXNKU+eRnpl9moKnB4QOLf1HestfXbmab5FXxiDBjc=
 github.com/Azure/go-ansiterm v0.0.0-20170929234023-d6e3b3328b78 h1:w+iIsaOQNcT7OZ575w+acHgRric5iCyQh+xv+KJ4HB8=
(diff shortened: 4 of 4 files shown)


pkg/minikube/cluster/cluster.go
```diff
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"github.com/docker/machine/libmachine/ssh"
```
afbjorklund: this change looks odd

cmd/minikube/cmd/config/config.go
```diff
@@ -257,6 +257,18 @@ var settings = []Setting{
...
+		set: SetBool,
+	},
+	{
+		name: "mount",
+		set: SetBool,
+	},
+	{
+		name: "mount-string",
+		set: SetString,
+	},
```
afbjorklund: fail to see the connection between ssh and "mount" ?

yinzara: I could have added this to a second PR if you'd prefer.  They were just not added when the 'mount' and 'mount-string' flags were added and I needed them all in my use case.

pkg/minikube/cluster/cluster.go
yinzara: Agree. That was definitely added in error. I'll fix.

cmd/minikube/cmd/start.go
```diff
@@ -154,6 +157,7 @@ func init() {
 	startCmd.Flags().Bool(gpu, false, "Enable experimental NVIDIA GPU support in minikube (works only with kvm2 driver on Linux)")
 	startCmd.Flags().Bool(hidden, false, "Hide the hypervisor signature from the guest in minikube (works only with kvm2 driver on Linux)")
 	startCmd.Flags().Bool(noVTXCheck, false, "Disable checking for the availability of hardware virtualization before the vm is started (virtualbox)")
+	startCmd.Flags().Bool(noNativeSSH, false, "Disable using the native Golang SSH client and instead uses the command line 'ssh' command when accessing the docker machine. Useful for the machine drivers when they will not start with 'Waiting for SSH'.")
```
medyagh: I would make the description shorter, instead of `Useful for the machine drivers when they will not start with 'Waiting for SSH'. ` I would link to an issue or a docs.md

like see github.com/kubernetes/minikube/docs/drivers#nossh.md

yinzara: There have been NUMEROUS issues related to "Waiting for SSH" for almost every VM driver which is why I specifically mentioned it as a possible fix when you encounter that issue.

The issue is that it isn't "required" by the Hyperkit driver.  It's required if you have a USB network adapter like a USB dongle or a docking station that provides wired network access for a laptop (which is why I made it an optional flag instead of setting it automatically for the hyperkit driver).

medyagh: > There have been NUMEROUS issues related to "Waiting for SSH" for almost every VM driver which is why I specifically mentioned it as a possible fix when you encounter that issue.
Fair enough, I know this issue has happened for sure for none driver, I wonder if it has happened to virtualbox as well ?

cmd/minikube/cmd/config/config.go
yinzara: I moved it out and will open a separate PR.

cmd/minikube/cmd/start.go
yinzara: My only debate with naming in this PR relates to the name of the arg itself.

Should it be "--no-native-ssh" since it mirrors the "--native-ssh" arg from 'docker machine ssh' and effectively affects its opposite or should it be "--external-ssh" (or maybe --ssh-cli) so that it implies what it actually does?

medyagh:
@minikubebot OK to test

RA489:
@minikube-bot OK to test

yinzara:
I need a little help deciphering what went wrong. It seems the Linux-VirtualBox test somehow failed but I can't figure out what's causing it.  Help!

yinzara:
I've rebased against master and force pushed. All builds are "Pending" now, does something need to be done?

yinzara:
Nvm everything is good now :) . Can I get a lgtm?

medyagh:
lets wait for @tstromberg's input

RA489:
@minikube-bot OK to test

yinzara:
I still don't understand how the VirtualBox-Linux test is failing.  My change is so minor and should have no affect unless specifically enabled (which it isn't by default). Is it possible this is unrelated to my changes?  The errors are so cryptic I can't decipher what's going on.

yinzara:
2028 ssh_runner.go:101] SSH: docker load -i /tmp/k8s-dns-kube-dns-amd64_1.14.13


That seems to be the line that's erroring.  All of the `docker load...` lines before it are passing but that one seems to fail.

medyagh:
I believe the vbox failing test was timing out

yinzara:
Yay! All we need is a lgtm :) . @tstromberg  ?

tstromberg:
> Though the native SSH client for Golang is more efficient than the command line 'ssh' command, there are times in which using the 'ssh' CLI is necessary.

First, thank you for this PR!

I want to make sure I understand the technical merits of adding another flag before LGTM'ing, as once we do, we are stuck supporting it for the long haul.  Do you mind elaborating what specific use cases this flag this flag would be useful for in the PR description and help text?

If the minikube VM used an exotic authentication mechanism, I would completely understand why using the external ssh mechanism is useful, but I don't yet understand other reasons for it. I would expect that the packets will be routed similarly.

In general, I would personally would prefer having an automatic fallback/selection mechanism for flags like this, rather than forcing the user to fail first and try again, but I don't understand enough about the problem yet to have a more specific recommendation.

yinzara:
Unfortunately the reason for needing this flag are not always clear.  In my example with the Hyperkit driver, when a USB network adapter was added to the system, the Golang SSH library was unable to connect to the Hyperkit container. It just sat waiting forever at "Waiting for SSH" without ever timing out.  It was literally frozen and didn't time out (so I couldn't have "fallen back automatically" even if I wanted to).

While I would love to have solved the cause for this, there are TONS of issues over the years related to issues with "Waiting for SSH..." that trying this option may have allowed a user to continue. If we know the command line SSH client does work, having a method for enabling this is necessary (as docker-machine ssh already does).  The docker-machine library doesn't even provide more information about the use of this flag other than its existence, how to use it, and that it may provide different functionality.
https://docs.docker.com/machine/reference/ssh/

yinzara:
Looks like this was changed in 2017 as part of:
https://github.com/kubernetes/minikube/commit/eafb4fcdbe935d6dd8dc07526ff869a15a6d6c47

RA489:
@minikube-bot OK to test

cmd/minikube/cmd/start.go
```diff
@@ -160,6 +163,7 @@ func init() {
...
```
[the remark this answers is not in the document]
yinzara: So then should we change the parameter to be "--native-ssh" and the default behavior to use the "External" CLI SSH (as it was before) eafb4fc (and as it is in `docker-machine ssh`) or should I change the parameter to be "--external-ssh" and maintain the current default behavior (even though it is the opposite of `docker-machine ssh`)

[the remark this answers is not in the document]
tstromberg: Correct. Name the flag `--native-ssh`, but do not yet change the default behavior.

tstromberg:
Needs rebase.

yinzara:
lol I think you missed the two options. I can either name the parameter "--native-ssh" and change the default behavior to use External SSH (like docker-machine ssh) by removing the line from the eafb4fc or I can name the parameter "--external-ssh" and leave the default it the way it is currently.  Adding a "--native-ssh" parameter would do nothing unless I change the default behavior.

RA489:
@yinzara Please resolve the conflicts.

sharifelgamal:
Could you change the name to be `--native-ssh` while still defaulting the behavior to `true`? Ideally changing the default behavior would be in a separate PR. This way, it can still be set to `false` when needed.

yinzara:
I don't know how to make myself clearer lol 

If I call it `--native-ssh` while still defaulting the behavior to `true`, this PR will literally do nothing.   That would mean if I call "minikube ssh" it would use native SSH no matter if I use the `--native-ssh` flag or not and there will be no way of using the external CLI ssh (the exact reason for this PR).

There are literally only two options (and a variation on the 1st in naming). 

Call it `--no-native-ssh` or `--external-ssh` and keep the default the same (i.e. native SSH). Specify the flag would cause it to use the "external" CLI SSH.

OR

Call it `--native-ssh` (like `docker-machine ssh`) and change the default to use "external" CLI SSH.

sharifelgamal:
`--native-ssh=false` would do exactly what you want without having to change the parameter name or the default behavior.

yinzara:
INTERESTING! 

I didn't realize that was an option (I can't see any other minikube options like that).  

I guess I could call it `--native-ssh` and make it default to "true" when not specified but if specified as false, follow the value given. Then if we change the default behavior, the CLI flag doesn't change. I didn't realize viper had that ability.

yinzara:
There.  Rebased against master and renamed to `--native-ssh` (where `--native-ssh=false` will enable use of the "external" CLI ssh).  

Also added the flag to the `minikube ssh` command

sharifelgamal:
Excellent, thanks for fixing that!

vmvmpvm:
Does this flag work?

I have minikube version 1.12.1, I still cannot use external ssh, I am running below command on Windows 10 using Hyper-V.

minikube start -v=8 --alsologtostderr --driver=hyperv --force --native-ssh=false

yinzara:
Yup we use it all the time and it works great. What do you mean by "I still cannot use external ssh"?

vmvmpvm:
Sorry, I should have provided more details, it always uses native ssh in my case

I am using minikube version: v1.12.1.

I execute below command

```
minikube start -v8 --alsologtostderr --profile minikube03 --driver hyperv --force --hyperv-virtual-switch "External Switch" --native-ssh=false
```

I see below error

```
I0731 15:17:41.393688    8300 main.go:115] libmachine: Using SSH client type: native
I0731 15:17:41.394690    8300 main.go:115] libmachine: &{{{<nil> 0 [] [] []} docker [0x7b7070] 0x7b7040 <nil>  [] 0s} myvmip 22 <nil> <nil>}
I0731 15:17:41.395689    8300 main.go:115] libmachine: About to run SSH command:
sudo poweroff
I0731 15:17:52.360497    8300 main.go:115] libmachine: Error dialing TCP: dial tcp myvmip:22: connectex: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
I0731 15:18:02.398113    8300 main.go:115] libmachine: Error dialing TCP: dial tcp myvmip:22: connectex: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
I0731 15:18:16.363667    8300 main.go:115] libmachine: Error dialing TCP: dial tcp myvmip:22: connectex: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
I0731 15:18:26.399598    8300 main.go:115] libmachine: Error dialing TCP: dial tcp myvmip:22: connectex: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
....

```

How do I make it detect my External SSH?, Just fyi, when I run minishift it calls my external ssh just fine.

vmvmpvm:
My external ssh config is configured to map myvmip ssh to a forwarded port, I need minikube to use the external ssh config for it to work.

sharifelgamal:
@vmvmpvm I've opened https://github.com/kubernetes/minikube/pull/8907 to fix your issue.

vmvmpvm:
Thanks @sharifelgamal

PR #4512, opened 2019-06-17
WIP: Add option to list all profiles

djmgit:
Created a new command `minikube profiles` to list
all the available profiles

### DIFF
--- a/pkg/util/kubeconfig.go
+++ b/pkg/util/kubeconfig.go
@@ -325,3 +325,16 @@ func UnsetCurrentContext(filename, machineName string) error {
 	}
 	return nil
 }
+
+//Fetch all the profiles and return
+func GetProfiles(filename string) ([]string, error) {
+	config, err := ReadConfigOrNew(filename)
+	if err != nil {
+		return nil, errors.Wrap(err, "Error getting kubeconfig status")
+	}
+	profiles := make([]string, 0, 4)
+	for context, _ := range config.Contexts {
+		profiles = append(profiles, context)
+	}
+	return profiles, nil
+}
--- a/cmd/minikube/cmd/config/profiles.go
+++ b/cmd/minikube/cmd/config/profiles.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"os"
+	"github.com/spf13/cobra"
+	"k8s.io/minikube/pkg/minikube/console"
+	"k8s.io/minikube/pkg/minikube/exit"
+	pkgutil "k8s.io/minikube/pkg/util"
+	"k8s.io/minikube/pkg/minikube/constants"
+)
+
+// ProfileCmd represents the profile command
+var ProfilesCmd = &cobra.Command{
+	Use:   "profiles",
+	Short: "profiles gets the list of all the present profiles",
+	Long:  "profiles displays a list of all the profiles which have been created earlier.",
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) > 1 {
+			exit.Usage("usage: minikube profile [MINIKUBE_PROFILE_NAME]")
+		}
+		profiles, err := pkgutil.GetProfiles(constants.KubeconfigPath)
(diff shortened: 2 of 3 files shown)


medyagh:
I am not sure what is better,
to have a separate command to list profiles , `minikube profiles`
or add it to the existing `profile` (without s) command ?

something like `minikube profile list`

I could be convinced either way, what others think ?

djmgit:
@medyagh I liked the idea of ```minikube profile list`` too, however creating a new command like profiles, does not requiring changing the behavior of existing command.
To make ```minikube profile list``` work we would have to change the behavior of the ```profile``` command. We need to create a new sub command like ```minikube profile set <profile_name>``` or something like that (as you have mentioned too in the issue). So basically its like creating new ```profile``` sub commands which is very much doable. 
However as I mentioned before creating a new command one level up like ```minikube profiles``` will not affect any existing code.
In my pr I have implemented ```minikube profiles``` command, however I can update it if required to implement the first way.

pkg/util/kubeconfig.go
```diff
@@ -325,3 +325,16 @@ func UnsetCurrentContext(filename, machineName string) error {
...
+
+//Fetch all the profiles and return
+func GetProfiles(filename string) ([]string, error) {
+	config, err := ReadConfigOrNew(filename)
+	if err != nil {
+		return nil, errors.Wrap(err, "Error getting kubeconfig status")
+	}
+	profiles := make([]string, 0, 4)
+	for context, _ := range config.Contexts {
+		profiles = append(profiles, context)
```
medyagh: This is getting all kubecontexts... Inclduing the ones not created by minikube.

djmgit:
@medyagh the contexts which I get via the above way and the profiles present in ~/.minikube/profiles directory. Although it contains the profile created by kubernetes (```minikube```), shouldn't it be shown too? Because it is also there under the profiles directory.

pkg/util/kubeconfig.go
medyagh: > @medyagh the contexts which I get via the above way and the profiles present in ~/.minikube/profiles directory. Although it contains the profile created by kubernetes (`minikube`), shouldn't it be shown too? Because it is also there under the profiles directory.

that code will get all the kubectl contexts, so for example if I have a GKE cluster or another k8s cluster, there will be a list of contexts, and when you create minikube profile it will create another context.

if you get list of all contexts and show it as list of profiles, it will show my gke cluster or my other k8s cluster as a minikube profile , so it is wrong

djmgit:
@medyagh got it, so I should instead list the contents of ~/.minikube/profiles . Will that be the right approach?

medyagh:
> @medyagh got it, so I should instead list the contents of ~/.minikube/profiles . Will that be the right approach?

Yes there are a few details :
 - make sure the profiles in `~/.minikube/profiles` are valid profile ( not empty left over profile configs)
 - make sure the machine vm file also exists too

djmgit:
@medyagh Got it.
Just to confirm, the following approach has to be followed : 
- Iterate over the profiles in ~/.minikube/profiles
- Check if they are not empty (contains config.json)
- Check if the corresponding machine with the respective profile name is present in ~/.minikube/machines
- Create a list of such valid profiles and display

Does the above look fine?

medyagh:
> @medyagh Got it.
> Just to confirm, the following approach has to be followed :
> 
> * Iterate over the profiles in ~/.minikube/profiles
> * Check if they are not empty (contains config.json)
> * Check if the corresponding machine with the respective profile name is present in ~/.minikube/machines
> * Create a list of such valid profiles and display
> 
> Does the above look fine?

Yes that sounds greats, only think to validate the profile I wouldn't just check if config.json is not empty, I would check if it is loadable ?

medyagh:
@djmgit this PR needs a rebase, please let me know if you are still working on this PR

djmgit:
@medyagh I am still on it, got caught up with some other stuff, will update and rebase my PR ASAP.

medyagh:
> @medyagh I am still on it, got caught up with some other stuff, will update and rebase my PR ASAP.

no worries, let me know if you need any help on this

pkg/util/kubeconfig.go
```diff
@@ -325,3 +325,16 @@ func UnsetCurrentContext(filename, machineName string) error {
...
 }
+
+//Fetch all the profiles and return
+func GetProfiles(filename string) ([]string, error) {
+	config, err := ReadConfigOrNew(filename)
+	if err != nil {
+		return nil, errors.Wrap(err, "Error getting kubeconfig status")
+	}
+	profiles := make([]string, 0, 4)
+	for context, _ := range config.Contexts {
```
fenglixa: I guess you are mean: "for _, context := ..." here, is it right?

fenglixa: Pls run `make test` before committing the code

tstromberg:
Just checking in - would you like any help with moving this PR forward?

medyagh:
@djmgit  Thanks for trying to make this PR work but have to close it PR in favor this one https://github.com/kubernetes/minikube/pull/4811 as this functionality is needed for the other PR depending on it. 
. I welcome your future contributions.

PR #4538, opened 2019-06-20
jenkins: Increase integration test time limit to 90 minutes

tstromberg:
I've seen a few panics due to tests taking too long recently, such as:

https://storage.googleapis.com/minikube-builds/logs/4455/Linux-VirtualBox.txt

13 "minikube start" runs in, and it ran out of time. Longer term, we need to do some brainstorming as to how to get integration tests under 30 minutes. The test latency is intolerable.


medyagh:
I agree I have been adding prs to make the minikube profile work. Currently it has bugs. I think with profiles we can make it parallel and also we can simply tests and not redo them.

For example in test version upgrade test we also do start stop delete .. which is another test and that alone takes a quite amount of time

Issue #4531, opened 2019-06-19
Add Documentation for registry addon


medyagh:
Related PRs:

https://github.com/kubernetes/minikube/pull/4341
https://github.com/kubernetes/minikube/pull/4529

Maybe could be added under this
https://github.com/kubernetes/minikube/blob/master/docs/reusing_the_docker_daemon.md

PR #4529, opened 2019-06-19
Register registry-proxy.yaml.tmpl with registry addons

kumarom:
Related to #4341

PR #4341, opened 2019-05-24
Enable registry-proxy

kumarom:
As per [this blog](https://blog.hasura.io/sharing-a-local-registry-for-minikube-37c7240d0615) and [this gist](https://gist.github.com/coco98/b750b3debc6d517308596c248daf3bb1), we need to deploy a registry-proxy
which will expose docker registry on the minikube host.

Once this daemon set is deployed on minikube, one can access registry on `$(minikube ip):5000`.
This has been tested with minikube v1.0.1 with none driver. With this, one will not have to use
`kubectl port-forward`. I was able to push a container image to registry using
```
docker push $(minikube ip):5000/test-img
```
And then ran it in minikube using
```
kubectl run -i -t test-img --image=$(minikube ip):5000/test-img --restart=Never
```

Issue #4242, opened 2019-05-11
Document how to use the registry addon - covering usage on none & Windows


robertdumitrescu:
**The operating system version**: Microsoft Windows 10 PRO - Version 10.0.17134 Build 17134

Can we add to the documentation a tutorial of some sort on how to "Use local images by re-using the Docker daemon" or how to have a minikube registry up and running on Windows?

I am really struggling with this one and in the end, after spending 6 hours trying to set up with no success I ended up using Amazon ECR for pushing images and then pulling down to Minikube with skaffold. 

Some sort of step by step tutorial for local development with local registries or other mechanisms on Windows with Minikube would be highly appreciated.

Thank you! Let me know your thoughts guys.

AkarshES:
This might be useful: https://blog.hasura.io/sharing-a-local-registry-for-minikube-37c7240d0615/
Essentially the same thing is supposed to be usable by enabling the registry addon in minikube but there is an issue in the current release which has a fix now in (this commit)[https://github.com/kubernetes/minikube/commit/40da5023d05354e1155942d2aab242411dd10bd7]

kumarom:
The https://blog.hasura.io/sharing-a-local-registry-for-minikube-37c7240d0615/ point to https://gist.github.com/coco98/b750b3debc6d517308596c248daf3bb1 which also deploys a daemon sets. This daemon set is missing from minikube addon registry yaml files. Is this something that needs to be fixed..?

I confirm that once this daemon set is created, I can access registry on localhost:5000 when using **minikube v1.0.1** with **none** driver. This is without needing a kubectl port-forward. I was able to push a container image to registry using
```
docker push localhost:5000/test-img
```
And then ran it in minikube using 
```
kubectl run -i -t test-img --image=localhost:5000/test-img --restart=Never
```

The final state of kubernetes was something like below:
```
ubuntu@ubuntu:~$ kubectl get pods --all-namespaces
NAMESPACE     NAME                                       READY     STATUS      RESTARTS   AGE
default       test-img                                   0/1       Completed   0          4s
kube-system   coredns-78fcdf6894-25hgh                   1/1       Running     0          45m
kube-system   coredns-78fcdf6894-z9rxg                   1/1       Running     0          45m
kube-system   default-http-backend-ddc5ff66b-xnhb4       1/1       Running     0          45m
kube-system   etcd-minikube                              1/1       Running     0          45m
kube-system   kube-addon-manager-minikube                1/1       Running     0          45m
[...]

PR #4539, opened 2019-06-20
Add docs for registry proxy addon

kumarom:
This commit adds documentation on how to use minikube addon registry.
Related to #4341 and #4529
This addresses #4531 and #4242

### DIFF
--- a/docs/reusing_the_docker_daemon.md
+++ b/docs/reusing_the_docker_daemon.md
@@ -1,5 +1,7 @@
 # Reusing the Docker daemon
 
+## Method 1: Without minikube registry addon
+
 When using a single VM of Kubernetes it's really handy to reuse the Docker daemon inside the VM; as this means you don't have to build on your host machine and push the image into a docker registry - you can just build inside the same docker daemon as minikube which speeds up local experiments.
 
 To be able to work with the docker daemon on your mac/linux host use the docker-env command in your shell:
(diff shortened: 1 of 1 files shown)


docs/reusing_the_docker_daemon.md
```diff
@@ -31,3 +37,39 @@ The fix is to update /etc/sysconfig/docker to ensure that minikube's environment
 ```
 
 Remember to turn off the _imagePullPolicy:Always_, as otherwise Kubernetes won't use images you built locally.
+
+Another approach is to enable minikube registry addons and then push images directly into registry. Steps for this approach is as follows:
+
+Ensure that docker is configured to use `192.168.39.0/24` as insecure registry. Refer [here](https://docs.docker.com/registry/insecure/) for instructions.
```
medyagh: is this IP same on windows, mac on different drivers ... ?

docs/reusing_the_docker_daemon.md
```diff
@@ -31,3 +37,39 @@ The fix is to update /etc/sysconfig/docker to ensure that minikube's environment
 ```
 
 Remember to turn off the _imagePullPolicy:Always_, as otherwise Kubernetes won't use images you built locally.
+
+Another approach is to enable minikube registry addons and then push images directly into registry. Steps for this approach is as follows:
```
medyagh: Could you please make a heading so it is clear at the first look that this is the second method ?

maybe have the following headiings:

# Reusing the Docker daemon
## Method 1 : using ...

## method: 2 : using...

docs/reusing_the_docker_daemon.md
kumarom: Will explain how to decide what subnet to use here.

docs/reusing_the_docker_daemon.md
kumarom: Will do..

docs/reusing_the_docker_daemon.md
```diff
@@ -31,3 +39,43 @@ The fix is to update /etc/sysconfig/docker to ensure that minikube's environment
...
+Push docker image to minikube registry:
+
+```shell
+docker push $(minikube ip):5000/test-img
+```
+
+Now run it in minikube:
+
+```shell
+kubectl run test-img --image=$(minikube ip):5000/test-img
```
Coderhypo: kubectl run a Deployment is deprecated, maybe you can use kubectl create or kubectl run a Pod

Coderhypo:
The "registry host" is indeed easy to misunderstand，this PR is very useful，thinks for you work！

Issue #4479, opened 2019-06-12
Build: add -dev suffix for minikube version on master build


medyagh:
Feature request:

currently when we build minikube, the binary `out/minikube version`  would return the latest version 
```
$ ./out/minikube-darwin-amd64 version
minikube version: v1.1.1
```
it would be nice if it lets you know this is a built of master HEAD, maybe have a suffix 1.1.1-snapshot

djmgit:
@medyagh I would like to take this up.
So what needs to done here is that the final binary that is produced via **make** should have a suffix like
**version-snapshot** right?

afbjorklund:
It is normally called "dev" in these contexts, not snapshot.

Probably a good feature would be to add the git commit ?

Compare:

```
$ runc --version
runc version 1.0.0-rc6+dev
commit: 2b18fe1d885ee5083ef9f0838fee39b62d653e30
spec: 1.0.1-dev
```

medyagh:
I like that format you pasted ! yes that would be great.
we want to make sure also it doesn't break anything including download iso ... (not sure if it would act weird if filename has + in it when downloading from GCS on different platforms including windows)

afbjorklund:
Yeah, I would probably use `-` rather than `+`, but that's another story. In that project they use a version file to set the name, so they can use whatever they want. Also note that we have a lot of subprojects that "assume" they are building from their own git, and thus either fail to record a version or use buildroot's.

So this innocent feature has some implications on the release process...

The first is that you _should_ still be able to build from git. That is, just because I build from `master` doesn't mean it's a development release - what if I check out the `v1.1.1` tag and build it ? The alternative would be tagging on a branch, and then you miss nice features (like `git describe`)

The second is that you should be able to **not** build from git. That is, there should be some way to override the information posted - what if you download the tarball and don't even have `git` installed ? Currently we use something like `$(MAKE) GIT_COMMIT=$(THE_COMMIT)` to provide the info.

Maybe we can start with adding a second line with the commit, and take it from there ?

``` console
$ minikube version
minikube version: v1.1.1
commit: ef60a69700170ace1b20b286cd6fd8d3e8d13277
```

----

There's already things in that little minimal output, to make people confused...

* why is "minikube" duplicated ? just in case there's another version (like the "spec" above) ?

* should there be a colon before the version ? lots of other programs don't have it there...

* which gets even worse, when _programs_ start screen-scraping the version output...

[...]

PR #4548, opened 2019-06-21
Display commit id along with version

djmgit:
Issue : #4479. Now git commit id will be displayed
in minukube version command output

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,10 @@ INSTALL_SIZE ?= $(shell du out/minikube-windows-amd64.exe | cut -f1)
 BUILDROOT_BRANCH ?= 2018.05.3
 REGISTRY?=gcr.io/k8s-minikube
 
+# Get git commit id
+COMMIT_NO := $(shell git rev-parse HEAD 2> /dev/null || true)
+COMMIT ?= $(if $(shell git status --porcelain --untracked-files=no),"${COMMIT_NO}-dirty","${COMMIT_NO}")
+
 HYPERKIT_BUILD_IMAGE 	?= karalabe/xgo-1.12.x
 # NOTE: "latest" as of 2019-05-09. kube-cross images aren't updated as often as Kubernetes
 BUILD_IMAGE 	?= k8s.gcr.io/kube-cross:v1.12.5-1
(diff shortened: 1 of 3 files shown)


djmgit:
@medyagh @afbjorklund Added commit id as it was discussed on the issue.

djmgit:
If the current branch contains changes which are not staged, the ```-dirty``` is appended to the commit id. This is the same convention which is followed by runc.

tstromberg:
Looks like `make lint` emits some errors:

```
pkg/version/version.go:34:5: var `gitCommitId` should be `gitCommitID` (golint)
var gitCommitId = ""
    ^
pkg/version/version.go:47:6: func GetGitCommitId should be GetGitCommitID (golint)
func GetGitCommitId() string {
     ^
cmd/minikube/cmd/version.go:35:3: var `gitCommitId` should be `gitCommitID` (golint)
		gitCommitId := version.GetGitCommitId()
		^
make[1]: *** [lint] Error 1
```

pkg/version/version.go
```diff
@@ -30,6 +30,9 @@ const VersionPrefix = "v"
 // version is a private field and should be set when compiling with --ldflags="-X k8s.io/minikube/pkg/version.version=vX.Y.Z"
 var version = "v0.0.0-unset"
 
+// version is a private field and should be set when compiling with --ldflags="-X k8s.io/minikube/pkg/version.gitCommitId=<commit-id>"
+var gitCommitId = ""
```
tstromberg: this should be gitCommitID: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/version/version.go
```diff
@@ -40,6 +43,11 @@ func GetVersion() string {
 	return version
 }
 
+// GetGitCommitId returns the git commit id from which it is being built
+func GetGitCommitId() string {
```
tstromberg: This should be `GetGitCommitID` 

I'd normally say this should be just `GitCommitID()`, but we should be consistent with the other existing function calls.

medyagh: I would also rather GitCommitID even if it is not consistent with others, over time we can improve others

Makefile
```diff
@@ -26,6 +26,10 @@ INSTALL_SIZE ?= $(shell du out/minikube-windows-amd64.exe | cut -f1)
 BUILDROOT_BRANCH ?= 2018.05.3
 REGISTRY?=gcr.io/k8s-minikube
 
+# Get git commit id
+COMMIT_NO := $(shell git rev-parse HEAD 2> /dev/null || true)
+COMMIT ?= $(if $(shell git status --porcelain --untracked-files=no),"${COMMIT_NO}-dirty","${COMMIT_NO}")
+
```
medyagh: I personally rather -dev than -dirty

afbjorklund: The terms have different meaning. The `-dev` means that it has a couple of commits more than the latest official release (git tag). The `-dirty` means that someone has edit the files locally, compared to the commit. That is, the binary does not (fully) reflect what is in the commit. We should make sure that official builds can be built "clean", or otherwise this will not look good.

djmgit:
@afbjorklund done
@medyagh In this PR I have kpt it GetCommitID, I am thinking of creating a separate issue for doing away with the ```Get``` prefix in every function and make the respective changes. This way all of such changes will be in the same commit. I will be opening the issue shortly.

medyagh:
@minikube-bot OK to test

Makefile
```diff
@@ -26,6 +26,10 @@ INSTALL_SIZE ?= $(shell du out/minikube-windows-amd64.exe | cut -f1)
 BUILDROOT_BRANCH ?= 2018.05.3
 REGISTRY?=gcr.io/k8s-minikube
 
+# Get git commit id
+COMMIT_NO := $(shell git rev-parse HEAD 2> /dev/null || true)
```
medyagh: let's limit it to first 6 characters of the commit . the full commit ID visually is too long.

medyagh: lets also call this COMMIT_ID as it is not a number

medyagh:
Thank you for this contribution.

PR #4552, opened 2019-06-21
Suppress error message when translation file doesn't exist

sharifelgamal:


pkg/minikube/translate/translate.go
```diff
@@ -77,13 +77,13 @@ func DetermineLocale() {
 	translationFile := "pkg/minikube/translate/translations/" + preferredLanguage.String() + ".json"
 	t, err := ioutil.ReadFile(translationFile)
 	if err != nil {
-		glog.Errorf("Failed to load transalation file for %s: %s", preferredLanguage.String(), err)
+		glog.Infof("Failed to load translation file for %s: %s", preferredLanguage.String(), err)
```
medyagh: lets change the second %s to %v for err

pkg/minikube/translate/translate.go
```diff
@@ -77,13 +77,13 @@ func DetermineLocale() {
...
 		return
 	}
 
 	err = json.Unmarshal(t, &Translations)
 	if err != nil {
-		glog.Errorf("Failed to populate translation map: %s", err)
+		glog.Infof("Failed to populate translation map: %s", err)
```
medyagh: %v for err here too

pkg/minikube/translate/translate.go
sharifelgamal: Done.

pkg/minikube/translate/translate.go
sharifelgamal: Done.

pkg/minikube/translate/translate.go
```diff
@@ -77,13 +77,13 @@ func DetermineLocale() {
 	translationFile := "pkg/minikube/translate/translations/" + preferredLanguage.String() + ".json"
 	t, err := ioutil.ReadFile(translationFile)
 	if err != nil {
-		glog.Errorf("Failed to load transalation file for %s: %s", preferredLanguage.String(), err)
+		glog.Infof("Failed to load translation file for %s: %v", preferredLanguage.String(), err)
```
tstromberg: I may be incorrect, but I think the call to String() is likely unnecessary,  as %s should call it automatically.

sharifelgamal: You're probably right yeah.

Issue #2982, opened 2018-07-14
Debian package should include the kvm driver (docker-machine-driver-kvm2)


marcominetti:
**Is this a BUG REPORT or FEATURE REQUEST?**: FEATURE REQUEST

The minikube-XXX-.deb package is intended to be used on debian systems. To simplify setup and usage of kvm2 driver, it could be include within the package (maybe provided through private debian repository). Thus the installation and upgrade of minikube will be more convenient.

Please provide the following details:

**Environment**: Linux

**Minikube version**: 0.27 and 0.28
- **OS** : Ubuntu 18.04
- **VM Driver**: KVM2

**What happened**:
I need to install and maintain (upgrade) docker-machine-driver-kvm2 manually as sudoer.

**What you expected to happen**:
Installing/upgrading the minikube through debian package, the docker-machine-driver-kvm2 is installed/upgraded accordingly.

idvoretskyi:
+1, I second this.

nmabhinandan:
+1

afbjorklund:
I suppose removing the .deb altogether is another option ? Then it could be packaged _properly_ downstream instead, using PPA or some similar means. And then wouldn't have to defend the lack of .rpm etc (#402)

afbjorklund:
If we packaged kvm2 as debs, maybe that could avoid versioning issues: #3437

mgabeler-lee-6rs:
/remove-lifecycle rotten

This is still very desirable for Debian-based systems.

afbjorklund:
@mgabeler-lee-6rs : note that virtualbox is still the default driver

Added some .deb packaging now, but it is not a hard dependency

mgabeler-lee-6rs:
Given vbox as the default driver, `Suggests` seems like the right relation, yes.

Thanks for this!

afbjorklund:
Most likely we will need to set up our own apt repository for this to work...

Similar to what is available for kubeadm (but **not** for `bionic`, just yet?)

https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl

```shell
curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
cat <<EOF >/etc/apt/sources.list.d/kubernetes.list
deb https://apt.kubernetes.io/ kubernetes-xenial main
EOF
```

We also need to fix minikube build so that it actually builds it for `xenial`.

[...]

PR #4556, opened 2019-06-22
Add linux packaging for the kvm2 driver binary

afbjorklund:
Just some copy/paste of the minikube packaging, should probably be cleaned up later.

For #2982

### DIFF
--- a/installers/linux/deb/kvm2_deb_template/DEBIAN/control
+++ b/installers/linux/deb/kvm2_deb_template/DEBIAN/control
@@ -5,8 +5,8 @@ Priority: optional
 Architecture: amd64
 Depends: libvirt0 (>= 1.3.1)
 Recommends: minikube
-Maintainer: Aaron Prindle <aaprindle@gmail.com>
+Maintainer: Thomas Strömberg <t+minikube@stromberg.org>
 Description: Machine driver for KVM
- Minikube uses Docker Machine to manage the Kubernetes VM so it benefits
+ minikube uses Docker Machine to manage the Kubernetes VM so it benefits
  from the driver plugin architecture that Docker Machine uses to provide
  a consistent way to manage various VM providers.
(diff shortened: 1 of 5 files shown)


installers/linux/deb/kvm2_deb_template/DEBIAN/control
```diff
@@ -0,0 +1,12 @@
+Package: docker-machine-driver-kvm2
+Version: --VERSION--
+Section: base
+Priority: optional
+Architecture: amd64
+Depends: libvirt0 (>= 1.3.1)
+Recommends: minikube
+Maintainer: Aaron Prindle <aaprindle@gmail.com>
```
tstromberg: This should probably be someone else. If you don't want your name, you may put in my own address: `t+minikube@stromberg.org`

I wonder if Debian accepts mailing lists as an alternative to a single person?

installers/linux/deb/kvm2_deb_template/DEBIAN/control
```diff
@@ -0,0 +1,12 @@
...
+Description: Machine driver for KVM
+ Minikube uses Docker Machine to manage the Kubernetes VM so it benefits
```
tstromberg: I prefer 'minikube' to 'Minikube'. We are mini, after all. :)

installers/linux/deb/kvm2_deb_template/DEBIAN/control
afbjorklund: Okay, will change the maintainer on both packages (minikube and docker-machine-driver-kvm2)

I think that it does, the `docker-ce` package has: `Maintainer: Docker <support@docker.com>`

installers/linux/deb/kvm2_deb_template/DEBIAN/control
afbjorklund: Weird to start a sentence with a lower case letter, but OK

PR #4568, opened 2019-06-24
Allow --kubernetes-version to be specified without the leading v

tstromberg:
I bumped into this issue when preparing for the demo at KubeCon.

With this PR, minikube allows both `--kubernetes-version=v1.15.0` and `--kubernetes-version=1.15.0`. Otherwise, it fails later when it tries to cache images.

This PR was inspired by @dimitropoulos's work in #3766 - but rather than raising an early error if v is omitted, allows both formats.


sharifelgamal:
This looks fine once the lint error is fixed. Apparently there's a `testDataDir` that's not used somewhere.

PR #4579, opened 2019-06-24
Makefile: add platform specific phony targets

### LINKED ISSUE #4458: Add ability in Make to build specific platforms
`make cross`

`Windows 10 Enterprise`

I am doing development on Windows in addition to `Bash for Windows`. As of now, running `make cross` compiles the binaries for every platform i.e. `Linux`, `Mac` & `Windows`. We need to have the ability to build for a targeted platform. For example for me, it doesn't make sense to build the Mac binaries as well while trying to build the Windows binaries to test.

blueelvis:
1. Added platform specific Phony targets (Linux, Windows & Darwin)
2. Added Documentation in the build instructions about this.
3. Changed `cp` to `mv` to remove a duplicate file.

Fixes #4458

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ out/minikube$(IS_EXE): out/minikube-$(GOOS)-$(GOARCH)$(IS_EXE)
 	cp $< $@
 
 out/minikube-windows-amd64.exe: out/minikube-windows-amd64
-	cp out/minikube-windows-amd64 out/minikube-windows-amd64.exe
+	mv out/minikube-windows-amd64 out/minikube-windows-amd64.exe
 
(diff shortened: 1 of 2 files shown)


afbjorklund:
BTW; the change from `cp` to `mv` now causes windows to be rebuilt everytime... Not good.

Would rather have duplicate files, than having to rebuild it because something removed output.

blueelvis:
@afbjorklund - How about we directly output as `.exe`? That would give best of both options.

afbjorklund:
Yeah, but it is harder to write the make rules. So best would be if Windows stopped with the DOS :-)

PR #4581, opened 2019-06-24
 Add translation files into the minikube binary

sharifelgamal:


cmd/extract/extract.go
```diff
@@ -30,7 +30,7 @@ import (
 func main() {
 	paths := []string{"cmd", "pkg"}
 	functions := []string{"translate.T"}
-	output := "pkg/minikube/translate/translations"
+	output := "translations"
```
medyagh: at first look it wasn't clear to me output was a filename, I suggest a better name, may be `f` or `outFile` ?

sharifelgamal: `output` is a directory

medyagh: then maybe `outDir`

PR #4584, opened 2019-06-25
improve code readability cmd package 

medyagh:
- moving long cobra run functions (for stop, delete) to their own funcs for better readability and to match start cmd.
- also adding defer close for api in start cmd


medyagh:
@minikube-bot OK to test

PR #4595, opened 2019-06-25
Added validation for start memory size

### LINKED ISSUE #4437: Add memory size sanity checking

minikube start --disk-size 35g --memory 8 --vm-driver virtualbox

😄  minikube v1.1.0 on darwin (amd64)
🔥  Creating virtualbox VM (CPUs=2, Memory=8MB, Disk=35000MB) ...
^C

I exited the command after around 10 minutes.

💣  command runner: getting ssh client for bootstrapper: Error dialing tcp via ssh client: dial tcp 127.0.0.1:22: connect: connection refused

macOS Mojave 10.14.5

blueelvis:
1. Added memory size checking
2. Reworked the function to parse human readable size
3. Reworked disk size.

As the previous PR went bad on my local, created this new PR. Older PR - #4463 

Fixes #4437


blueelvis:
@medyagh @afbjorklund -- This is the cleaned up PR.

medyagh:
@minikube-bot OK to test

RA489:
@minikube-bot OK to test

blueelvis:
Any idea why the Osx-Virtualbox build step is being marked as failure?

medyagh:
@blueelvis  
Currently if it is under min size : it warns two times. is there an easy way to warn only once ?
```
/minikube start start --vm-driver=hyperkit --memory=12
😄  minikube v1.2.0 on darwin (amd64)
📌  Requested memory allocation (12MB) is less than the default memory allocation of (2000MB). Beware that Minikube might not work correctly or crash unexpectedly.
💡  Requested memory allocation (12MB) is less than the minimum allowed of 1024MB

```

blueelvis:
@medyagh - That should fix the issue :)

pkg/minikube/constants/constants.go
```diff
@@ -151,14 +151,16 @@ const (
...
+	//MinimumMemorySize is the minimum memory size, in megabytes
+	MinimumMemorySize = "1024mb"
 	// DefaultCPUS is the default number of cpus of a host
 	DefaultCPUS = 2
-	// DefaultDiskSize is the default disk image size, parseable
-	DefaultDiskSize = "20g"
-	// MinimumDiskSizeMB is the minimum disk image size, in megabytes
-	MinimumDiskSizeMB = 2000
+	// DefaultDiskSize is the default disk image size, in megabytes
+	DefaultDiskSize = "20000mb"
```
medyagh: was this supposed to be 20000 ?

cmd/minikube/cmd/start.go
```diff
@@ -390,6 +390,14 @@ func validateConfig() {
 		glog.Errorf("Error autoSetOptions : %v", err)
 	}
 
+	memorySizeMB := pkgutil.CalculateSizeInMB(viper.GetString(memory))
+	if memorySizeMB < pkgutil.CalculateSizeInMB(constants.MinimumMemorySize) {
+		exit.Usage("Requested memory allocation (%dMB) is less than the minimum allowed of %dMB", memorySizeMB, pkgutil.CalculateSizeInMB(constants.MinimumMemorySize))
+	}
+	if memorySizeMB < pkgutil.CalculateSizeInMB(constants.DefaultMemorySize) {
+		console.OutStyle(console.Notice, "Requested memory allocation (%dMB) is less than the default memory allocation of (%dMB). Beware that Minikube might not work correctly or crash unexpectedly.", memorySizeMB, pkgutil.CalculateSizeInMB(constants.DefaultMemorySize))
```
medyagh: pleaese change Minikube to minikube (we are mini no need to capitalize :) )

pkg/minikube/constants/constants.go
blueelvis: Yep, the default disk size is currently set to `20g`

cmd/minikube/cmd/start.go
blueelvis: Done.

pkg/util/utils.go
```diff
@@ -53,13 +54,18 @@ type RetriableError struct {
...
+// CalculateSizeInMB returns the number of MB in the human readable string
+func CalculateSizeInMB(humanReadableSize string) int {
+	_, err := strconv.ParseInt(humanReadableSize, 10, 64)
+	if err == nil {
+		humanReadableSize += "mb"
+	}
+	size, err := units.FromHumanSize(humanReadableSize)
 	if err != nil {
-		glog.Errorf("Invalid disk size: %v", err)
+		exit.WithCode(exit.Config, "Invalid size passed in argument: %v", err)
```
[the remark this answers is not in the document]
blueelvis: I agree with you that the most ideal way would be to let the caller handle it. But, before I pushed this commit, I tried to do that and it became pretty messy. Adding loads of `if` conditions didn't make sense when ideally we should be exiting immediately if we are not able to parse the size. It seems to be better for the function itself to exit rather than the chance of caller not handling it properly deep in the code.

This utility function needs to be called every time when we need to parse any memory or disk size. Be it the defaults or input by the user. Do you think it is still adding a lot of `if` conditions to check for errors everywhere?

cmd/minikube/cmd/start.go
```diff
@@ -390,6 +390,14 @@ func validateConfig() {
 		glog.Errorf("Error autoSetOptions : %v", err)
 	}
 
+	memorySizeMB := pkgutil.CalculateSizeInMB(viper.GetString(memory))
+	if memorySizeMB < pkgutil.CalculateSizeInMB(constants.DefaultMemorySize) {
+		console.OutStyle(console.Notice, "Requested memory allocation (%dMB) is less than the default memory allocation of (%dMB). Beware that Minikube might not work correctly or crash unexpectedly.", memorySizeMB, pkgutil.CalculateSizeInMB(constants.DefaultMemorySize))
+	}
+	if memorySizeMB < pkgutil.CalculateSizeInMB(constants.MinimumMemorySize) {
+		exit.Usage("Requested memory allocation (%dMB) is less than the minimum allowed of %dMB", memorySizeMB, pkgutil.CalculateSizeInMB(constants.MinimumMemorySize))
```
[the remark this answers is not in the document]
blueelvis: If that is the case, I can put an empty `exit.Usage("")` command below `console.OutStyle`. Do you agree with it? On the other hand, I think it makes sense to add a function to `error` package similar to `OutStyle` of `Console` package. What do you think?

pkg/util/utils.go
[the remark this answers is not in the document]
medyagh: ok you convinced me this is a case we can tolerate exit because of the nature that if it is not readable we can not for sure continue ! we could be more cautious if more people use this function for other utilities.

PR #4605, opened 2019-06-26
Use glog instead of logrus

dfang:
glog is everywhere in repo, 

try to remove logrus dependency in go.mod but failed,  as docker relies on it (go mod why).

at least by changing logrus.Fatal to glog.Fatalf can make logrus as an indirect dependency


medyagh:
@dfang good catch ! thanks for the contribution !

Issue #3720, opened 2019-02-19
Make virtualbox DNS settings configurable


chancez:
#3453 change the default behavior and as a result, I believe I am seeing breakage as a result. If these items were configurable I believe it would allow me to more easily validate the issue is related, and bypass the current values that are problematic.

As it it is currently, it seems even if I attempt to change the value of this setting, minikube will overwrite it.

taravkov:
Agree, please make them configurable.

cvila84:
Agree, as activating the DNS proxy (through HostDNSResolver=true or DNSProxy=true) on NAT sometimes lead to silent crashes which prevents any DNS requests to be done (https://github.com/kubernetes/minikube/issues/3606#issuecomment-505014348)

PR #4619, opened 2019-06-27
virtualbox: Make DNS settings configurable

cvila84:
Closes https://github.com/kubernetes/minikube/issues/3720

add --dns-proxy and --host-dns-resolver flags to start command for controlling virtualbox dns settings

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -97,6 +97,8 @@ const (
 	embedCerts            = "embed-certs"
 	noVTXCheck            = "no-vtx-check"
 	downloadOnly          = "download-only"
+	dnsProxy              = "dns-proxy"
+	hostDNSResolver       = "host-dns-resolver"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


cvila84:
Hello @medyagh,

These settings would allow people to disable the usage of Virtualbox NAT DNS proxy which can be quite unstable (see [https://www.virtualbox.org/ticket/13899?cversion=0&cnum_hist=3]).

To test it, when --dns-proxy and --host-dns-resolver are set to false, the /etc/resolv.conf of the minikube VM shall contain the same DNS IP as the host. If any of the settings are true, the /etc/resolv.conf of the minikube shall be the default DNS proxy IP which is 10.2.0.3

medyagh:
ok thank you for the explain. does this have any relation with proxy or vpn ?

@minikube-bot OK to test

cvila84:
@medyagh i don't see any relation with proxy. the only relation i see with VPN is related to host network switching (office <-> VPN home) while minikube VM is running. when DNS proxy is deactivated, it won't reflect the new host DNS IP into /etc/resolv.conf but this is an acceptable constraint as sometimes the silent crash of Virtualbox DNS proxy prevents any communication to be done from the VM to the external world (because of DNS resolve timeout)

i see the label needs-ok-to-test is still set. it seems you need to write /ok-to-test on its own line

PR #4644, opened 2019-06-29
Add kvm prefix to flags only used by kvm

josedonizetti:
As both flags `hidden` and `gpu` are kvm only, I feel we should prefix it to make it explicit instead of having it only on the flag documentation.

wdyt?


medyagh:
@minikube-bot OK to test

RA489:
@minikube-bot OK to test

medyagh:
Failing test is not related to this ! lgtm

Issue #4345, opened 2019-05-24
minikube_iso failed: crio-bin-1.14.1: <file> already exists and is not an object file


tstromberg:
As seen at https://storage.googleapis.com/minikube-builds/logs/4340/build.txt

```
/usr/bin/make -C pause
make[3]: Entering directory '/var/lib/jenkins/go/src/k8s.io/minikube/out/buildroot/output/build/crio-bin-v1.14.1/pause'
/var/lib/jenkins/go/src/k8s.io/minikube/out/buildroot/output/host/bin/x86_64-minikube-linux-gnu-gcc -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -O2  -static -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -c -o pause.o pause.c
/var/lib/jenkins/go/src/k8s.io/minikube/out/buildroot/output/host/bin/x86_64-minikube-linux-gnu-gcc -o ../bin/pause pause.o -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -O2  -static 
make[3]: Leaving directory '/var/lib/jenkins/go/src/k8s.io/minikube/out/buildroot/output/build/crio-bin-v1.14.1/pause'
go build -i  -tags "btrfs_noversion exclude_graphdriver_devicemapper libdm_no_deferred_remove exclude_graphdriver_btrfs containers_image_ostree_stub seccomp   containers_image_ostree_stub" -o bin/crio github.com/cri-o/cri-o/cmd/crio
go build github.com/cri-o/cri-o/vendor/github.com/hashicorp/golang-lru/simplelru: build output "/var/lib/jenkins/go/src/k8s.io/minikube/out/buildroot/output/build/crio-bin-v1.14.1/_output/pkg/linux_amd64/github.com/cri-o/cri-o/vendor/github.com/hashicorp/golang-lru/simplelru.a" already exists and is not an object file
Makefile:107: recipe for target 'bin/crio' failed
make[2]: *** [bin/crio] Error 1
make[2]: *** Waiting for unfinished jobs....
( cd conmon && /var/lib/jenkins/go/src/k8s.io/minikube/out/buildroot/output/build/crio-bin-v1.14.1/bin/crio-config )
[...]

PR #4657, opened 2019-07-01
Upgrade Docker, from 18.09.6 to 18.09.7

afbjorklund:
runc v1.0.0-rc8-0-g425e105

    containerd v1.2.6


afbjorklund:
Currently this doesn't build properly with go1.10, so might need to upgrade Buildroot to get go1.12 ?

```
# runtime/internal/sys
../../../../../../../host/lib/go/src/runtime/internal/sys/stubs.go:11:7: const initializer 1 * (1 - GoarchArm64) + 16 * GoarchArm64 is not a constant
../../../../../../../host/lib/go/src/runtime/internal/sys/stubs.go:11:22: undefined: GoarchArm64
Makefile:173: receptet för målet ”bin/containerd-shim” misslyckades

```

afbjorklund:
Hmm, must be something else - because containerd v1.2.5 is also failing, with the same build error...

afbjorklund:
Rebuilding both the host go packages fixed it, so just something about the local out/buildroot...

afbjorklund:
The Jenkins build failed due to #4345

Issue #4592, opened 2019-06-25
Run as many integration tests as possible in parallel using profile


medyagh:
now that we have improved minikube profile, (with https://github.com/kubernetes/minikube/pull/4504 and https://github.com/kubernetes/minikube/pull/4572) we could use as many integration tests as meaningful in parallel using profile to speed up the integration tests.

medyagh:
Done by the PR

PR #4665, opened 2019-07-02
WIP: Run integration tests in profile in parallel 

medyagh:
Closes https://github.com/kubernetes/minikube/issues/4592


test/integration/util/common.go
```diff
@@ -0,0 +1,162 @@
...
+	"testing"
+	"time"
+
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/labels"
+	commonutil "k8s.io/minikube/pkg/util"
+)
+
+// WaitForBusyboxRunning waits until busybox pod to be running
+func WaitForBusyboxRunning(t *testing.T, namespace string, kCtx string) error {
```
tstromberg: nit: since `ctx` is traditionally used for uses of https://golang.org/pkg/context/ - what do you think about naming the var `context`?

It is admittedly weird that Kubernetes contexts mean something completely different than Go contexts.

test/integration/addons_test.go
```diff
@@ -76,7 +76,7 @@ func readLineWithTimeout(b *bufio.Reader, timeout time.Duration) (string, error)
 
 func testDashboard(t *testing.T) {
 	t.Parallel()
-	minikubeRunner := NewMinikubeRunner(t)
+	minikubeRunner := NewMinikubeRunner(t, "")
```
tstromberg: Passing an empty string to a function is a code smell, as the meaning of the string is completely mysterious to readers. Let's not do that. Three alternative ideas come to mind:

- Perhaps `NewMinikubeRunner` should randomly generate the profile name. It gets populated in the return object anyways, so that clients can retrieve it if they need it. This may make test cleanup trickier, but makes parallelism easier.
- NewMinikubeRunner could allow a config struct to be passed, where one of the options is a name.
- There could be two functions: one for a named runner, and one for a default named runner.

test/integration/start_stop_delete_test.go
```diff
@@ -63,17 +64,18 @@ func TestStartStop(t *testing.T) {
...
 	}
 
-	for _, test := range tests {
-		t.Run(test.name, func(t *testing.T) {
-			r := NewMinikubeRunner(t)
-			if !strings.Contains(test.name, "docker") && usingNoneDriver(r) {
-				t.Skipf("skipping %s - incompatible with none driver", test.name)
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			testName := tc.name
```
tstromberg: Just use `tc.name` rather than introducing a second variable representing the test name.

test/integration/util/kubectl_runner.go
```diff
@@ -0,0 +1,130 @@
...
+// NewKubectlRunner creates a new KubectlRunner
+func NewKubectlRunner(t *testing.T, kctx string) *KubectlRunner {
+	p, err := exec.LookPath(kubectlBinary)
+	if err != nil {
+		t.Fatalf("Couldn't find kubectl on path.")
+	}
+	return &KubectlRunner{BinaryPath: p, T: t, KCtx: kctx}
+}
+
+// CurrentContext gets the current kubectl current-context
```
tstromberg: Since `PastContext` makes no sense, can this be named `Context` or `ContextName`. I do understand that this differs from the name of the kubectl command.

test/integration/util/kubectl_runner.go
```diff
@@ -0,0 +1,130 @@
...
+func (k *KubectlRunner) CurrentContext() (stdOut string, err error) {
```
tstromberg: Use `name` for the returned variable. It isn't strictly `stdOut`, since it does get trimmed. Also, readers shouldn't have to know the stdOut format to use this function.

test/integration/util/common.go
medyagh: how about kubeCtx ?

pkg/util/kubernetes.go
```diff
@@ -68,10 +68,15 @@ func (s *PodStore) Stop() {
 	close(s.stopCh)
 }
 
-// GetClient gets the client from config
-func GetClient() (kubernetes.Interface, error) {
+// GetClient gets the client from config for the given kubectl context
+func GetClientWithKContext(kCtx string) (kubernetes.Interface, error) {
```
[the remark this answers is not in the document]
medyagh: fair will rename

medyagh:
closing in favor of https://github.com/kubernetes/minikube/pull/4665

PR #4674, opened 2019-07-03
Switch documentation site to docsy theme, add sample content

tstromberg:
This switches our doc website from whisper to docsy, and includes sample content and sample home page. 

The intent here is to have a site structure committed to the tree that we can build upon - it is not yet ready for prime time.


afbjorklund:
Shouldn't this docs website be in a separate git repository / project ?

tstromberg:
@afbjorklund - In-repo docs are the suggested way for Kubernetes subprojects to operate, so that PR's can atomically update code and documentation.  I'll admit this PR is rather large and unwieldly at the moment, but subsequent PR's should make more sense.

tstromberg:
netlify is working properly with SASS now (see command), so this PR is now ready for review.

afbjorklund:
> In-repo docs are the suggested way for Kubernetes subprojects to operate, so that PR's can atomically update code and documentation.

I fully support this (otherwise the documentation gets out of sync), but I just thought this "doc website" was **separate** from the code ? Similar to the current https://kubernetes.io/docs/tasks/tools/install-minikube/

site/content/en/_index.html
```diff
@@ -0,0 +1,89 @@
...
+{{< /blocks/cover >}}
+
+
+{{% blocks/lead color="primary" %}}
+
+minikube is a local Kubernetes environment focused on application developers.
+
+{{< imgproc "start" Fit "730x239" >}}{{< /imgproc >}}
+
+Kubernetes certified, mother approved.
```
afbjorklund: Kid Tested, Mother Approved ? :-) Does this mean we need to pass the certification test again ?

tstromberg:
> > In-repo docs are the suggested way for Kubernetes subprojects to operate, so that PR's can atomically update code and documentation.
> 
> I fully support this (otherwise the documentation gets out of sync), but I just thought this "doc website" was **separate** from the code ? Similar to the current https://kubernetes.io/docs/tasks/tools/install-minikube/

Correct: The primary Kubernetes documentation site does run in it's own repository, but subproject docs are managed in-tree. This PR is for the minikube-specific documentation, sort of like http://kind.sigs.k8s.io

Once The Kubernetes documentation & minikube documentation are using the same engine, we can  do a better job of synchronizing the documentation between the two.

site/content/en/_index.html
tstromberg: Yes =)

Issue #4658, opened 2019-07-01
Add warn about old driver version (hyperkit, kvm)


josedonizetti:
Now that both kvm, and hyperkit support `version`. It would be good to warn the user if his driver version is too old.

josedonizetti:
@medyagh does this make sense?

medyagh:
Yes that's a great idea! Would be happy to review the PR

josedonizetti:
Ok. I’ll create a pr.

PR #4676, opened 2019-07-03
Add warn if kvm driver version is old

josedonizetti:
issue: https://github.com/kubernetes/minikube/issues/4658

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors All rights reserved.
+Copyright 2016 The Kubernetes Authors All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -870,7 +870,7 @@ func validateDriverVersion(vmDriver string) {
 
 		// if the driver doesn't have return any version, it is really old, we force a upgrade.
 		if len(v) == 0 {
-			exit.WithCode(exit.Failure, "Please upgrade the 'docker-machine-driver-kvm2'.")
+			exit.WithCode(exit.Failure, "Please upgrade the 'docker-machine-driver-kvm2'. %s", constants.KVMDocumentation)
 		}
 
 		vmDriverVersion, err := semver.Make(v)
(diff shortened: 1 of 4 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -850,3 +853,52 @@ func saveConfig(clusterConfig cfg.Config) error {
...
+		if err != nil {
+			console.Warning("Error checking driver version: %v", err)
+			return
+		}
+
+		v := extractVMDriverVersion(string(output))
+
+		// if the driver doesn't have return any version, it is really old, we force a upgrade.
+		if len(v) == 0 {
+			exit.WithCode(exit.Failure, "Please upgrade the 'docker-machine-driver-kvm2'.")
```
medyagh: lets include a link to the docs or where they can download the kvm2 driver

cmd/minikube/cmd/start.go
```diff
@@ -850,3 +853,52 @@ func saveConfig(clusterConfig cfg.Config) error {
...
+		}
+
+		minikubeVersion, err := version.GetSemverVersion()
+		if err != nil {
+			console.Warning("Error parsing minukube version: %v", err)
+			return
+		}
+
+		if vmDriverVersion.LT(minikubeVersion) {
+			console.Warning("The 'docker-machine-driver-kvm2' version is old. Please consider upgrading.")
```
medyagh: same here, we could link to a page where they can easily find the newest kvm2.

( i noticed our docs page is a bit confusion, as part of this PR we could make that doc page for kvm2 super simple.( we have too many ifs in the doc)

cmd/minikube/cmd/start.go
```diff
@@ -850,3 +853,52 @@ func saveConfig(clusterConfig cfg.Config) error {
...
+		}
+	}
+}
+
+func extractVMDriverVersion(s string) string {
```
medyagh: please add a comment with example output so it is clear for the next person what this func is doing

cmd/minikube/cmd/start_test.go
```diff
@@ -0,0 +1,45 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
```
medyagh: 2019

josedonizetti:
@medyagh Let me know what you think about the changes. I've added a constant to the kvm documentation, we could start enforcing it everywhere we link to the documentation, that way if we need to change one day, it's a single change. 

I'll try to simplify the documentation now. Although I'm intrigued the PR https://github.com/kubernetes/minikube/pull/4681 make this one unnecessary.

medyagh:
I think this PR could be used to detect the too old KVM drivers that we want to enforce the upgrade ( before the user waits a few minutes till the error happen) if we know the kvm driver is too old. we should end gracefuly without error.

the other PR is good when minikube crashes with an error. it is meant as an after-happened handling, so we have less support ticket.

so this is PR is still helpful and needed to detect sooner

josedonizetti:
@medyagh I've added an upgrade section to the KVM documentation. I know you mentioned we had too many `ifs`, but it kind of seems necessary to explain about dependencies, and possible problems with libvirtd. Wdyt?

medyagh:
@minikube-bot OK to test

PR #4678, opened 2019-07-04
Delete kubeconfig context when a machine has been deleted

kairen:
The commit added a new function for deleting the kubeconfig context when a machine has been deleted.

### DIFF
--- a/pkg/util/kubeconfig.go
+++ b/pkg/util/kubeconfig.go
@@ -344,3 +344,29 @@ func SetCurrentContext(kubeCfgPath, name string) error {
 	}
 	return nil
 }
+
+// DeleteKubeConfigContext deletes the specified machine's kubeconfig context
+func DeleteKubeConfigContext(kubeCfgPath, machineName string) error {
+	kcfg, err := ReadConfigOrNew(kubeCfgPath)
+	if err != nil {
+		return errors.Wrap(err, "Error getting kubeconfig status")
+	}
+
+	if kcfg == nil || api.IsConfigEmpty(kcfg) {
+		glog.V(2).Info("kubeconfig is empty")
+		return nil
+	}
+
+	delete(kcfg.Clusters, machineName)
+	delete(kcfg.AuthInfos, machineName)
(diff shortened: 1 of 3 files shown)


medyagh:
@minikube-bot OK to test

pkg/util/kubeconfig.go
```diff
@@ -344,3 +344,29 @@ func SetCurrentContext(kubeCfgPath, name string) error {
 	}
 	return nil
 }
+
+// DeleteKubeConfig deletes the specified machine's kubeconfig file
+func DeleteKubeConfig(kubeCfgPath, machineName string) error {
```
medyagh: I think it is better to rename this to DeleteKubeContext, since we are not actually deleting kubeconfig file.

medyagh:
Thank you so much for this contribution ! this makes minikube much more stable and user-friendly ! thank you @kairen

pkg/util/kubeconfig.go
kairen: I agree with you. I am gonna update! Thank you for your suggestion.

kairen: @medyagh Updated. PTAL.

pkg/util/kubeconfig.go
```diff
@@ -344,3 +344,29 @@ func SetCurrentContext(kubeCfgPath, name string) error {
 	}
 	return nil
 }
+
+// DeleteKubeConfigContext deletes the specified machine's kubeconfig context
+func DeleteKubeConfigContext(kubeCfgPath, machineName string) error {
+	kcfg, err := ReadConfigOrNew(kubeCfgPath)
+	if err != nil {
+		return errors.Wrap(err, "Error getting kubeconfig status")
```
josedonizetti: Does it make sense to start enforcing errors not to be capitalized? We currently have both in the minikube codebase, but I think we could start using the same pattern everywhere. WDYT? @kairen @medyagh

kairen:
@minikube-bot OK to test

RA489:
@minikube-bot OK to test

PR #4691, opened 2019-07-05
Warn if hyperkit version is old

josedonizetti:
Fix for https://github.com/kubernetes/minikube/issues/4658

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -955,7 +955,7 @@ func validateDriverVersion(vmDriver string) {
 
 	if vmDriverVersion.LT(minikubeVersion) {
 		out.WarningT(
-			"The '{{.driver_executable}}' version is old. Please consider upgrading. {{.documentation_url}}",
+			"There's a new version for '{{.driver_executable}}'. Please consider upgrading. {{.documentation_url}}",
 			out.V{"driver_executable": driverExecutable, "documentation_url": driverDocumentation},
 		)
 	}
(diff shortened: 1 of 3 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -856,39 +856,54 @@ func saveConfig(clusterConfig cfg.Config) error {
 }
 
 func validateDriverVersion(vmDriver string) {
```
medyagh: is there a reason we don't do this validation before download ISO ?

sharifelgamal:
@minikube-bot OK to test

josedonizetti:
@minikube-bot OK to test

cmd/minikube/cmd/start.go
```diff
@@ -856,39 +855,54 @@ func saveConfig(clusterConfig cfg.Config) error {
...
-			return
-		}
+	var (
+		driverExecutable    string
+		driverDocumentation string
+	)
+
+	switch vmDriver {
+	case constants.DriverKvm2:
+		driverExecutable = "docker-machine-driver-kvm2"
```
tstromberg: Please move these constants somewhere else - preferably as close to the driver as possible. I'm afraid we may  silently break this in the future, should we rename from kvm to kvm2, for instance.

josedonizetti: Let me know if the last change is enough, or if you prefer to actually create a constant inside the `pkg/drivers/kvm|hyperkit`.

tstromberg:
LGTM, but needs rebase.

RA489:
@josedonizetti please resolve the conflicts rest LGTM.

Issue #4684, opened 2019-07-04
discussion:  `minikube kubectl` 


josedonizetti:
The command `minikube kubectl` has a lot of problems. Any flag that you pass to it won't work because `corba` will try to parse it, eg: 

```
minikube kubectl logs -f my-pod
Error: unknown shorthand flag: 'f' in -f
Usage:
  minikube kubectl [flags]

Flags:
  -h, --help   help for kubectl

Global Flags:
      --alsologtostderr                  log to standard error as well as files
  -b, --bootstrapper string              The name of the cluster bootstrapper that will set up the kubernetes cluster. (default "kubeadm")
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory
      --logtostderr                      log to standard error instead of files
  -p, --profile string                   The name of the minikube VM being used. This can be set to allow having multiple instances of minikube independently. (default "minikube")
      --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
  -v, --v Level                          log level for V logs
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
```

Or if you run `--help` on it, it shows the corba help, not kubectl itself.

Would it make sense to deprecate it? Let users use kubectl, as long as we have the `minikube profile` to allow setting the context, or switching the context it should be okay.

wdyt? @medyagh @tstromberg @afbjorklund

afbjorklund:
I wouldn't say _lots_ of problems, you just have to put the `--` in the right place...

``` console
$ minikube kubectl -- --help
kubectl controls the Kubernetes cluster manager
```

See #4333

But maybe we should go for something like `sudo minikube kubectl --install` ?

I would prefer to install it in e.g. the home directory, but it's not always in $PATH.
So could use something like `/usr/local/bin`, and hope _that_ is in the default path.

I just want the user to avoid having to go through **two** horrible installation procedures.

See #3645

Hopefully some documentation could "fix" this ?

I think that we should handle kubectl like kubelet...

josedonizetti:
@afbjorklund Thank you so much for clarifying. That makes a lot of sense, definitely, some documentation would fix it.

afbjorklund:
Thanks for the suggestion, in the current form it wouldn't fly very well with our documentation.

``` console
$ minikube kubectl run hello-minikube --image=k8s.gcr.io/echoserver:1.4 --port=8080
[...]

PR #4692, opened 2019-07-05
Add kubectl examples

josedonizetti:
Fix for https://github.com/kubernetes/minikube/issues/4684

@afbjorklund I've added some examples instead of adding something to `Use`. When changing `Use` the cobra help was displayed wrongly `Run kubectl -- [flags]`.

wdyt?

### DIFF
--- a/cmd/minikube/cmd/kubectl.go
+++ b/cmd/minikube/cmd/kubectl.go
@@ -36,7 +36,10 @@ import (
 var kubectlCmd = &cobra.Command{
 	Use:   "kubectl",
 	Short: "Run kubectl",
-	Long:  `Run the kubernetes client, download it if necessary.`,
+	Long: `Run the kubernetes client, download it if necessary.
+Examples:
+minikube kubectl -- --help
+kubectl get pods --namespace kube-system`,
 	Run: func(cmd *cobra.Command, args []string) {
 		api, err := machine.NewAPIClient()
 		if err != nil {


medyagh:
Good one ! I am curious, how does it look in the terminal output ? would you mind pasting it ?

afbjorklund:
Here is what it looks like:
``` console
$ ./out/minikube kubectl --help
Run the kubernetes client, download it if necessary.
Examples:
minikube kubectl -- --help
minikube kubectl -- -nkube-system get pods

Usage:
  minikube kubectl [flags]

Flags:
  -h, --help   help for kubectl


```

cmd/minikube/cmd/kubectl.go
```diff
@@ -36,7 +36,11 @@ import (
 var kubectlCmd = &cobra.Command{
 	Use:   "kubectl",
 	Short: "Run kubectl",
-	Long:  `Run the kubernetes client, download it if necessary.`,
+	Long: `Run the kubernetes client, download it if necessary.
+Examples:
+minikube kubectl -- --help
+minikube kubectl -- -nkube-system get pods
```
afbjorklund: Seems to be missing a space ? i.e. we normally write `-n kube-system`

afbjorklund: Or even after `get`, more like `kubectl get pods --namespace kube-system`

Issue #4697, opened 2019-07-06
Add command for installing kubectl program


afbjorklund:
Currently we either run `kubectl` for the user, or ask them to find the download themselves.

Since we already have a versioning and caching mechanism, we could also do it for them ?

So the `minikube kubectl` could grow two new options, one `--install` and one `--prefix`.

Where the prefix would default to `/usr/local` (needs sudo), but you _could_ use e.g. `$HOME`

```
sudo minikube kubectl --install
```

```
minikube kubectl --install --prefix=$HOME
```

Note that for the second usage, the user would need to make sure to include it in $PATH.

Some distributions do it by default, while some use `~/bin` and some use `~/.local/bin`...

----

Reference #4684 and #3645

And also https://kubernetes.io/docs/tasks/tools/install-kubectl/ of course, including packages:

```
sudo apt-get install -y kubectl
```

```
su -c "yum install -y kubectl"
```

josedonizetti:
Wonder if instead, we should support a `minikube install`? I really like the idea of removing the complexity of installation, but why stop at `kubectl`? 
There are other dependencies we could also install. Like `docker-driver-machine-kvm2` or `docker-driver-machine-hyperkit`, maybe support a `--upgrade` flag.

afbjorklund:
Mostly because we install those other components on the VM ? And I don't want to install `docker`.

The drivers are a special case, wonder if we shouldn't just bundle them (like the virtualbox one) ?
But yeah, downloading those are needed - but probably automatically, rather than a command.

We should use the same functions underneath, but not sure if we need a `minikube install` cmd.

----

Note that there are benefits to _not_ install kubectl too, such as having multiple versions available.

We could also do some half-way approach, by printing the path to the `kubectl` (but not run it) ?

josedonizetti:
That's true, it wouldn't make sense to install docker or podman. And an `install` sounds too generic, better to stick with improving kubectl,  I like both new flags ideas `--install` and `--prefix`.

afbjorklund:
Typical usage:

``` console
$ minikube kubectl --install
💣  Failed to write kubectl: open /usr/local/bin/kubectl: permission denied
$ sudo minikube kubectl --install
$ 
```

or alternatively:

``` console
$ install $(minikube kubectl --path) /usr/local/bin
install: cannot remove '/usr/local/bin/kubectl': Permission denied
$ sudo install $(minikube kubectl --path) /usr/local/bin
$ 
```

[...]

PR #4698, opened 2019-07-06
Add an install option to kubectl command

afbjorklund:
* Add --install option to minikube kubectl command
* Add --path option to minikube kubectl command

```
Usage:
  minikube kubectl [flags]

Flags:
  -h, --help            help for kubectl
      --install         Install kubectl and exit
      --path            Display kubectl path instead of running it
      --prefix string   Installation prefix (default "/usr/local")
```

For #4697

### DIFF
--- a/cmd/minikube/cmd/kubectl.go
+++ b/cmd/minikube/cmd/kubectl.go
@@ -37,7 +37,7 @@ import (
 var (
 	kubectlInstallMode   bool
 	kubectlInstallPrefix string
-	kubectlPathMode      bool
+	kubectlPrintMode     bool
 )
 
 const defaultPrefix = "/usr/local"
@@ -80,7 +80,7 @@ var kubectlCmd = &cobra.Command{
 			return
 		}
 
-		if kubectlPathMode {
+		if kubectlPrintMode {
 			console.OutLn(path)
 			return
 		}
@@ -131,6 +131,6 @@ func installKubectl(binary, version, path, prefix string) {
 func init() {
 	kubectlCmd.Flags().BoolVar(&kubectlInstallMode, "install", false, "Install kubectl and exit")
 	kubectlCmd.Flags().StringVar(&kubectlInstallPrefix, "prefix", defaultPrefix, "Installation prefix")
-	kubectlCmd.Flags().BoolVar(&kubectlPathMode, "path", false, "Display kubectl path instead of running it")
+	kubectlCmd.Flags().BoolVar(&kubectlPrintMode, "print", false, "Print kubectl path instead of running it")
 	RootCmd.AddCommand(kubectlCmd)
 }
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/kubectl.go
```diff
@@ -84,6 +104,33 @@ var kubectlCmd = &cobra.Command{
...
+			console.Fatal("%s: %v", "Failed to write kubectl", err)
+			os.Exit(exit.Software)
+		}
+		exit.WithError("Failed to write kubectl", err)
+	}
+}
+
 func init() {
+	kubectlCmd.Flags().BoolVar(&kubectlInstallMode, "install", false, "Install kubectl and exit")
+	kubectlCmd.Flags().StringVar(&kubectlInstallPrefix, "prefix", defaultPrefix, "Installation prefix")
```
josedonizetti: won't this default prefix be a problem for windows?

cmd/minikube/cmd/kubectl.go
```diff
@@ -84,6 +104,33 @@ var kubectlCmd = &cobra.Command{
...
+	kubectlCmd.Flags().BoolVar(&kubectlPathMode, "path", false, "Display kubectl path instead of running it")
```
josedonizetti: without reading the help command I would consider that `path` would be the place to install, pretty much like `prefix`.
what do you think about `--which` cos it does the same thing as `which kubectl`, right? probably wouldn't make the same sense for a windows user.

cmd/minikube/cmd/kubectl.go
afbjorklund: Probably, but haven't used anything but MSYS myself.

cmd/minikube/cmd/kubectl.go
afbjorklund: Yeah, that is one idea. I copied `--url` (from dashboard), and then it became `--path`.

cmd/minikube/cmd/kubectl.go
josedonizetti: Yeah, I'm not a windows user as well. It worries me that downloading `kubeclt` currently works on all platforms, and now considering this default prefix it will break in windows. Does it make sense to only use the prefix when in non-windows systems? It basically does the previous behavior for windows, and we add an issue for someone else to implement this support.

afbjorklund: I can add another prefix (already have a suffix...), just not sure what it would be

cmd/minikube/cmd/kubectl.go
afbjorklund: Think `--which` can be misleading, since it is normally associated with PATH (which it is not in). Will go for something like `--print`, and hope that there is no confusion.

afbjorklund:
AFAIK, Windows doesn't **have** the concept of a default PATH ?

You choose `C:\any\random\path` and then set up variables and crap.

So I suppose one would have to use `--prefix` ?

Or another approach, such as running in MinGW.

tstromberg:
I prefer that minikube not be in the business of installing system binaries when possible. I don't mind downloading kubectl to our own folder, but I feel that this duty is better left to a package manager.

afbjorklund:
Okay

Issue #3454, opened 2018-12-14
containerd built for minikube reports it's version as a commit hash from buildroot


ianlewis:
**Minikube version** (use `minikube version`): 0.31.0
- **OS** (e.g. from /etc/os-release): Ubuntu 16.04.5 LTS (Xenial Xerus)
- **VM Driver** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep DriverName`): kvm2
- **ISO version** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep -i ISO` or `minikube ssh cat /etc/VERSION`): 0.31.0

```
$ minikube ssh 'containerd -v'
containerd github.com/containerd/containerd f3d114a1ef f3d114a1effd8a6ef773bee14fe49ea6d8d7c350
```

https://github.com/buildroot/buildroot/commit/f3d114a1ef

afbjorklund:
Seems like it doesn't "support"* building from a tarball, it is supposed to say `v1.2.0-beta.0`
https://github.com/kubernetes/minikube/blob/master/deploy/iso/minikube-iso/package/containerd-bin/containerd-bin.mk#L6

* https://github.com/containerd/containerd/blob/master/Makefile#L23

ianlewis:
I'm sure there are reasons minikube builds from source, but is it out of the question to just use the containerd binary releases?

afbjorklund:
Minikube uses binary releases for other software (like docker), it is a bit random at the moment.

It might even be so that the "-bin" in the package name originally referred to binary installations ?

Also the version seems a bit old, but I guess nobody uses containerd (preferring docker or cri-o)

https://github.com/containerd/containerd/releases

PR #4703, opened 2019-07-07
Upgrade crio to 1.15.0

afbjorklund:
The repository was moved to a separate organization
    
strings.ReplaceAll only work in go1.12, not go1.10
    
Need to create conmon/config.h without tags and git
    
The configuration _needs_ to be updated (yet again)


afbjorklund:
Currently doesn't build anymore, due to https://github.com/cri-o/cri-o/issues/2450

`runtime_vm.go:175:20: undefined: strings.ReplaceAll`

afbjorklund:
Reported the workarounds for conmon/config.h as https://github.com/cri-o/cri-o/issues/2575

medyagh:
I could upgrade the go version on the jenkins nodes, would that be sufficient ? or do we need to update the travis config too?

afbjorklund:
@medyagh : it is the buildroot go, and it will be upgraded when we move from 2018.x to 2019.x
But it was working ok in the previous cri-o (1.14) and it seemed like a petty reason to require it ?

deploy/iso/minikube-iso/package/crio-bin/crio.conf.default
```diff
@@ -202,15 +217,19 @@ gid_mappings = ""
...
+# and manage its lifecycle.
+manage_network_ns_lifecycle = false
+
+# The "crio.runtime.runtimes" table defines a list of OCI compatible runtimes.
+# The runtime to use is picked based on the runtime_handler provided by the CRI.
+# If no runtime_handler is provided, the runtime will be picked based on the level
+# of trust of the workload.
+
+[crio.runtime.runtimes.runc]
+runtime_path = ""
```
saschagrunert: Please be aware that `runc` has to be in `$PATH` to let this work during runtime.

deploy/iso/minikube-iso/package/crio-bin/crio-bin.patch
```diff
@@ -0,0 +1,39 @@
+diff --git a/Makefile b/Makefile
+index de79f63..a9ea13b 100644
+--- a/Makefile
++++ b/Makefile
+@@ -136,7 +136,7 @@ crio.conf: bin/crio
+ release-note: ${RELEASE_TOOL}
+ 	${RELEASE_TOOL} -n $(release)
+ 
+-conmon/config.h: git-vars cmd/crio-config/config.go oci/oci.go
++conmon/config.h: cmd/crio-config/config.go oci/oci.go
```
saschagrunert: So the git executable is not available within your build environment? Because the git-vars target was introduced to actually fix this, but it only checks for the `.git` folder. I think I can improve this :)

afbjorklund: Actually it was worse... We do have git, but that's for our buildroot. Thus it recorded [commits from a different project](https://github.com/kubernetes/minikube/issues/3454). We build all our packages from tarballs, so have to supply the matching commit for each one.

The main problem here is that the "git-vars" target forces the conmon/config.h to be re-generated, even if it already exists (same with crio.conf too)

deploy/iso/minikube-iso/package/crio-bin/crio.conf.default
afbjorklund: Yeah, this is what the crio generated by default. Compare it with our config (above).

    runtime_path = "/usr/bin/runc"

I generated and added the default config, since it changes and breaks _all. the. time_

* plugin_dir, plugin_dirs

deploy/iso/minikube-iso/package/crio-bin/crio-bin.patch
saschagrunert: Ah I see, thanks. Let's see if we can get rid of the git-vars dependency of this target entirely.

afbjorklund:
On the plus side, the new release seem to be working fine with the tweaks made.

Haven't tried it with the new runc yet, need to make a trial run with all the changes.

deploy/iso/minikube-iso/package/crio-bin/crio.conf.default
saschagrunert: I guess it should work fine :)

tstromberg:
@minikube-bot OK to test

PR #4713, opened 2019-07-08
WIP: Add prefix for the kubeconfig context and the VM name

### LINKED ISSUE #4680: profile's kube-context should have a "minikube-" prefix
I have been thinking, if we should make minikube kube contexts have a minikube- prefix to them, so the user knows it was created by minikube ?

let say if I create a minikube profile named dev, or test, 
I might have a GKE cluster or another cluster named test or dev, I wouldn't be sure which kube context belongs to minikube. and would be harder to clean up !

I would love to know others opinions on this.

kairen:
This PR added a prefix `minikube-` for the kubeconfig context and the VM name. In addition, if minikube starts by default, it will not add the prefix.

Fixes #4680.

/cc @medyagh

### DIFF
--- a/pkg/minikube/service/service.go
+++ b/pkg/minikube/service/service.go
@@ -28,7 +28,6 @@ import (
 	"github.com/golang/glog"
 	"github.com/pkg/browser"
 	"github.com/pkg/errors"
-	"github.com/spf13/viper"
 	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -72,11 +71,11 @@ func (k *K8sClientGetter) GetCoreClient() (typed_core.CoreV1Interface, error) {
 // GetClientset returns a clientset
 func (*K8sClientGetter) GetClientset(timeout time.Duration) (*kubernetes.Clientset, error) {
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
-	profile := viper.GetString(config.MachineProfile)
+	contextName := config.GetMachineName()
 	configOverrides := &clientcmd.ConfigOverrides{
 		Context: clientcmdapi.Context{
-			Cluster:  profile,
-			AuthInfo: profile,
+			Cluster:  contextName,
+			AuthInfo: contextName,
 		},
 	}
 	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
--- a/pkg/minikube/config/config.go
+++ b/pkg/minikube/config/config.go
@@ -96,15 +96,18 @@ func decode(r io.Reader) (MinikubeConfig, error) {
 
 // GetMachineName gets the machine name for the VM
 func GetMachineName() string {
-	if viper.GetString(MachineProfile) == "" {
+	if viper.GetString(MachineProfile) == "" || viper.GetString(MachineProfile) == constants.DefaultMachineName {
 		return constants.DefaultMachineName
 	}
-	return viper.GetString(MachineProfile)
+	return fmt.Sprintf("%s-%s", constants.MinikubePrefix, viper.GetString(MachineProfile))
 }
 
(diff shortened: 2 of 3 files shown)


medyagh:
I wonder how hard it would be to write a test for this.... and I wonder what are the edge cases that things could go wrong.

medyagh:
@minikube-bot OK to test

tstromberg:
Code looks great. Conceptually I love this idea, but I also want to make sure we don't break compatibility with VM's that were created on an older release of minikube.

I agree with @medyagh's commen. This suggestion abuses the existing tests a bit, but do you mind adding something like: `-p profiletest` to the arguments in this test:

https://github.com/kubernetes/minikube/blob/bbae91a67762791e442cd3f987d34505c02b3249/test/integration/version_upgrade_test.go#L81

This would ensure that we at least get the non-default profile upgrade case correct.

pkg/minikube/service/service.go
```diff
@@ -72,11 +71,11 @@ func (k *K8sClientGetter) GetCoreClient() (typed_core.CoreV1Interface, error) {
 // GetClientset returns a clientset
 func (*K8sClientGetter) GetClientset(timeout time.Duration) (*kubernetes.Clientset, error) {
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
-	profile := viper.GetString(config.MachineProfile)
+	contextName := config.GetMachineName()
```
tstromberg: I'm not sure that the `context` prefix adds meaningful context to the variable name. What do you think about just `name` as a variable name?

https://github.com/golang/go/wiki/CodeReviewComments#variable-names

RA489:
@minikube-bot Ok to test

kairen:
I also agree with @medyagh. I should consider carefully about compatibility. Actually, this change has some issues in the non-default profile:

1. If users have an existing cluster that runs by the old version. When users upgrade minikube to the new version for this change, it will be unable to manage the existing VM through minikube's command(e.g. stop, ssh, delete). For example:

 ```sh
# start a cluster in the old version
$ minikube start -p dev 

# delete a cluster in this change
$ ./out/minikube ssh -p dev --logtostderr
I0709 16:07:23.567137   48017 notify.go:128] Checking for updates...
W0709 16:07:23.733993   48017 exit.go:100] Error getting host: Machine does not exist for api.Exists(minikube-dev)
```

2. As noted above,  when users deleted an existing cluster, the profile will delete but the kubeconfig's context and the VM will not.

I will update the code to make sure this change doesn't break compatibility.

pkg/minikube/service/service.go
kairen: Yes, Use `name` as a variable name is cleanly. Because of this variable will pass to the Context structure, so we don't add any prefix to emphasize this variable is a `context`.

RA489:
@minikube-bot OK to test

tstromberg:
> But these two results are passed, so I think this test can't ensure the compatibility is correct, I think we need to add a new test for verifying the non-default profile upgrade, e.g. after upgrade to the new version, the test will check `ssh` or `status` command.

Interesting. Do you mind taking your changes and adding them to a `TestVersionUpgradeWithProfile` test? We can then see what the integration tests look like on our end. 

If the PR tests reliably, I'm comfortable with it.

kairen:
@tstromberg Sure, I will update in the weekend. I apologize for the delay in replying because I've been super busy recently.

medyagh:
@kairen are you still working this ?

kairen:
@medyagh Yes, I am still working, sorry would be a bit delayed addressing this, because of some personal commitments and company tasks. I will back in the next week.

tstromberg:
@kairen - Thank you for your work on this. Do you mind if I close the PR until it's ready to be looked at again? I think clearing up the prefixes will be very helpful to address user confusion and make it easier for us to cleanup stale prefixes.

Once you are able to make the changes, please re-open it and we'll make sure to review the PR in a more timely manner. I'm looking forward to it.

tstromberg:
Anyone willing to revive this PR? =)

PR #4494, opened 2019-06-15
Update gvisor runsc version

ianlewis:
- Updates the gvisor addon to use containerd shim v2
- Updates the version of runsc
- Auto-installs a gvisor RuntimeClass

Issue #4482

PR #4717, opened 2019-07-08
Rebuild gvisor image for integration tests

priyawadhwa:
We should rebuild the gvisor image for integration tests, so that if
changes are made to the gvisor image they are tested. I added an
environment variable that, when set, will change the expected gvisor
image repo.

should unblock #4494


medyagh:
@minikube-bot OK to test

pkg/minikube/constants/constants.go
```diff
@@ -430,6 +430,9 @@ const (
 	GvisorContainerdShimURL = "https://github.com/google/gvisor-containerd-shim/releases/download/v0.0.1-rc.0/gvisor-containerd-shim-v0.0.1-rc.0.linux-amd64"
 	// GvisorURL is the url to download gvisor
 	GvisorURL = "https://storage.googleapis.com/gvisor/releases/nightly/2018-12-07/runsc"
+
+	// MinikubeGvisorAddonImageRepo is the environment variable set by integration tests to change the gvisor image repo.
+	MinikubeGvisorAddonImageRepo = "MINIKUBE_GVISOR_ADDON_IMAGE_REPO"
```
medyagh: do we need to add set this evn to hack/jenkins/ bash files?

pkg/minikube/assets/addons.go
```diff
@@ -403,10 +403,12 @@ func GenerateTemplateData(cfg config.KubernetesConfig) interface{} {
 		Arch            string
 		ExoticArch      string
 		ImageRepository string
+		GvisorImageRepo string
 	}{
 		Arch:            a,
 		ExoticArch:      ea,
 		ImageRepository: cfg.ImageRepository,
+		GvisorImageRepo: os.Getenv(constants.MinikubeGvisorAddonImageRepo),
```
ianlewis: Would this not require normal addon users to set this environment variable? Seems like that would require an update to the docs. I'd prefer if it defaulted to cfg.ImageRepository

pkg/minikube/constants/constants.go
priyawadhwa: I don't think so since I set the environment variable programatically & only for the gvisor tests.

pkg/minikube/assets/addons.go
priyawadhwa: Since the [default](https://github.com/kubernetes/minikube/pull/4717/files#diff-751198fe5c4c7b092069faf65234f22aR27) value is `gcr.io/k8s-skaffold`, users should never have to set the environment variable, since it should be empty. This way, they should always be using the latest version of the gvisor image.

I opted out of using cfg.ImageRepository because that repo would be applied to other images minikube pulls as well (storage provisioner, etc), which wouldn't exist. 

This way, if the env variable is set (which should only happen in integration tests), then a non-default gvisor image will be used.

ianlewis: I was thinking some folks would want to override the image repo globally. My thought was to use GvisorImageRepo if it is set, but use ImageRepository if it's not set (i.e. go back to current behavior)

pkg/gvisor/enable.go
```diff
@@ -44,6 +44,7 @@ const (
 //   3. copies necessary containerd config files
 //   4. restarts containerd
 func Enable() error {
+	return fmt.Errorf("local image used correctly")
```
ianlewis: I assume this was for debugging? I would have thought that Go would complain about unreachable code when compiling.

hack/jenkins/common.sh
```diff
@@ -71,7 +71,8 @@ echo ">> Downloading test inputs from ${MINIKUBE_LOCATION} ..."
 gsutil -qm cp \
   "gs://minikube-builds/${MINIKUBE_LOCATION}/minikube-${OS_ARCH}" \
   "gs://minikube-builds/${MINIKUBE_LOCATION}/docker-machine-driver"-* \
-  "gs://minikube-builds/${MINIKUBE_LOCATION}/e2e-${OS_ARCH}" out
+  "gs://minikube-builds/${MINIKUBE_LOCATION}/e2e-${OS_ARCH}" out \
+  "gs://minikube-builds/${MINIKUBE_LOCATION}/gvisor-addon" out
```
ianlewis: I think out needs to be specified at the end. Here you have the destination twice in this command.

pkg/gvisor/enable.go
priyawadhwa: Yup, I want to make sure minikube actually pulls the locally built gvisor image, so tests should fail with this error

hack/jenkins/common.sh
priyawadhwa: thanks for catching that, i'll fix this!

pkg/gvisor/enable.go
ianlewis: Currently it fails with compile error so you may want to remove the code after this line if you are intentionally returning an error.

priyawadhwa: Where are you seeing the compile error? If I run `make out/gvisor-addon` locally the binary does compile, and running the docker image from `make gvisor-addon-image` gives me:

```
minikube$ docker run gcr.io/k8s-minikube/gvisor-addon:latest
2019/07/10 23:20:28 local image used correctly
```

ianlewis: Hmm, I'm seeing the error in Travis. Go's compiler complains if code is unreachable.
https://travis-ci.org/kubernetes/minikube/builds/557049237#L541

ianlewis: My bad. It was go vet the provides the error. In any case, the code afterwards isn't used.

tstromberg:
Looks like this still needs some work:

```
 10:50:38 | SSH: sudo ctr cri load /gvisor-image.tar
 10:50:39 | Output: ctr: failed to load image: rpc error: code = Unknown desc = failed to open file: open /gvisor-image.tar: no such file or directory

--- FAIL: TestFunctionalContainerd (353.19s)
    functional_test.go:79: error loading gvisor addon image: exit status 1 
         ctr: failed to load image: rpc error: code = Unknown desc = failed to open file: open /gvisor-image.tar: no such file or directory
=== RUN   TestPersistence
```

Let me know if there is anything I can do to help. Thank you!

priyawadhwa:
@tstromberg I tried adding the `minikube cache add` command, and I was expecting the tests to fail because I've added a temporary error to `gvisor.go` (wanted to make sure the local image was being pulled correctly).

I'm not sure why this is happening, would you have any suggestions?

ianlewis:
Friendly ping. #4494 is still blocked on this.

priyawadhwa:
hey @ianlewis thanks for the ping! For a quick update: I've found a couple bugs while trying to fix this -- right now this is blocked on #4985, which is needed to make sure we are running the local version of the gvisor image to correctly test changes

ianlewis:
@priyawadhwa Thanks for the update! Sounds like we open a can of worms!

medyagh:
PR looks good, only one test fails, but it might be a flake fail.
```
--- FAIL: TestContainerd (0.00s)
    --- FAIL: TestContainerd/GvisorRestart (1244.80s)
        containerd_test.go:59: waiting for gvisor controller to come up

```
I am gonna restart the KVM test 
for furture reference the error is :  

```
 22:41:54 | ! I0815 22:41:54.934127   24312 ssh_runner.go:101] SSH: sudo podman load -i /tmp/pause_3.1
 22:41:55 | ! W0815 22:41:55.145638   22225 exit.go:99] Wait failed: waiting for k8s-app=kube-proxy: timed out waiting for the condition
 22:41:55 | ! * 
 22:41:55 | ! X Wait failed: waiting for k8s-app=kube-proxy: timed out waiting for the condition
 22:41:55 | ! * 
 22:41:55 | ! * Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
 22:41:55 | !   - https://github.com/kubernetes/minikube/issues/new/choose
 22:41:55 | > * Waiting for: apiserver proxy
```

medyagh:
@priyawadhwa are you okay with merging this PR ?

priyawadhwa:
@medyagh yah I think this should be ok!

medyagh:
awesome thank you for fixing this !

Issue #4710, opened 2019-07-08
Error starting kubernetes cluster using minikube on Mac


zonghaishang:
➜  ~ minikube start
😄  minikube v1.2.0 on darwin (amd64)

⚠️  Ignoring --vm-driver=hyperkit, as the existing "minikube" VM was created using the none driver.
⚠️  To switch drivers, you may create a new VM using `minikube start -p <name> --vm-driver=hyperkit`
⚠️  Alternatively, you may delete the existing VM using `minikube delete -p minikube`

🔄  Restarting existing none VM for "minikube" ...
⌛  Waiting for SSH access ...

💣  Failed to enable container runtime: running command: sudo systemctl start docker: exit status 1

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
👉  https://github.com/kubernetes/minikube/issues/new

➜  ~ uname -a
Darwin localhost 18.2.0 Darwin Kernel Version 18.2.0: Mon Nov 12 20:24:46 PST 2018; root:xnu-4903.231.4~2/RELEASE_X86_64 x86_64

afbjorklund:
The none driver doesn't work on Mac, you will have to delete the VM and start over

Same as #4498

medyagh:
We are working on adding support for vm free minikube on all platforms, please stay tuned !
https://github.com/kubernetes/minikube/issues/4389

afbjorklund:
It will be VM free on Linux at least, but on Mac and Win it will instead use Docker's VM.

zonghaishang:
I have tried `minikube delete -p minikube` & `minikube delete`， It still reports the same error. 

@afbjorklund 
> The none driver doesn't work on Mac, you will have to delete the VM and start over
what should I do ？

afbjorklund:
I was assuming that you wanted to change drivers to hyperkit ?

`minikube start --vm-driver=hyperkit`

If you deleted the old one, there should be no more "none"...

zonghaishang:
I deleted the cache file(`~/.minikube`) and the minikube took effect.

zonghaishang:
`step1` : remove cache file `~/.minikube`
`step2`: minikube start -p minikube --vm-driver=hyperkit

```
➜  ~ minikube start -p minikube --vm-driver=hyperkit
😄  minikube v1.2.0 on darwin (amd64)
💿  Downloading Minikube ISO ...
 129.33 MB / 129.33 MB [============================================] 100.00% 0s
[...]

PR #4720, opened 2019-07-09
Validate driver is supported on current OS

josedonizetti:
Fix for https://github.com/kubernetes/minikube/issues/4710

```
minikube start --vm-driver=none
😄  minikube v1.2.0 on darwin (amd64)
💣  The driver 'none' is not supported on darwin
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -212,6 +212,10 @@ func runStart(cmd *cobra.Command, args []string) {
 	out.T(out.Happy, "minikube {{.version}} on {{.os}} ({{.arch}})", out.V{"version": version.GetVersion(), "os": runtime.GOOS, "arch": runtime.GOARCH})
 
 	validateConfig()
+	vmDriver := viper.GetString(vmDriver)
+	if !validateOSSupportVMDriver(runtime.GOOS, vmDriver) {
+		exit.WithCode(exit.Failure, "The driver '%s' is not supported on %s", vmDriver, runtime.GOOS)
+	}
 	validateUser()
 	validateDriverVersion(viper.GetString(vmDriver))
 
@@ -977,4 +981,15 @@ func extractVMDriverVersion(s string) string {
 
 	v := strings.TrimSpace(matches[1])
 	return strings.TrimPrefix(v, version.VersionPrefix)
+}
+
+func validateOSSupportVMDriver(os, vmDriver string) bool {
+	switch vmDriver {
+	case constants.DriverNone:
+		if os != constants.Linux {
+			return false
+		}
+	}
+
+	return true
 }
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -997,6 +997,10 @@ func validateOSSupportVMDriver(os, vmDriver string) bool {
 		if os != constants.Darwin {
 			return false
 		}
+	case constants.DriverKvm2:
+		if os != constants.Linux {
+			return false
+		}
 	}
 
 	return true
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -211,11 +211,15 @@ var startCmd = &cobra.Command{
 func runStart(cmd *cobra.Command, args []string) {
 	out.T(out.Happy, "minikube {{.version}} on {{.os}} ({{.arch}})", out.V{"version": version.GetVersion(), "os": runtime.GOOS, "arch": runtime.GOARCH})
 
-	validateConfig()
 	vmDriver := viper.GetString(vmDriver)
-	if !validateOSSupportVMDriver(runtime.GOOS, vmDriver) {
-		exit.WithCode(exit.Failure, "The driver '%s' is not supported on %s", vmDriver, runtime.GOOS)
+	if err := cmdcfg.IsValidDriver(runtime.GOOS, vmDriver); err != nil {
+		exit.WithCodeT(
+			exit.Failure,
+			"The driver '{{.driver}}' is not supported on {{.os}}",
+			out.V{"driver": vmDriver, "os": runtime.GOOS},
+		)
 	}
+	validateConfig()
 	validateUser()
 	validateDriverVersion(viper.GetString(vmDriver))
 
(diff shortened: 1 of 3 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -908,3 +913,14 @@ func extractVMDriverVersion(s string) string {
 	v := strings.TrimSpace(matches[1])
 	return strings.TrimPrefix(v, version.VersionPrefix)
 }
+
+func validateOSSupportVMDriver(os, vmDriver string) bool {
+	switch vmDriver {
+	case constants.DriverNone:
+		if os == constants.Windows || os == constants.Darwin {
```
afbjorklund: You should be looking for Linux here. For instance, if FreeBSD is added - it still won't support running

pkg/minikube/constants/constants.go
```diff
@@ -436,3 +436,12 @@ const (
...
 	KVMDocumentation = "https://github.com/kubernetes/minikube/blob/master/docs/drivers.md#kvm2-driver"
 )
+
+const (
+	// Windows OS
+	Windows = "windows"
+	// Darwin OS
+	Darwin = "darwin"
+	// Unix OS
+	Unix = "unix"
```
afbjorklund: Please use Linux here. We don't want to get into the UNIX wars again (and Darwin actually is UNIX, but anyway)

cmd/minikube/cmd/start_test.go
```diff
@@ -43,3 +45,22 @@ func Test_extractVMDriverVersion(t *testing.T) {
...
+	}{
+		{constants.DriverNone, constants.Darwin, false},
+		{constants.DriverNone, constants.Windows, false},
+		{constants.DriverNone, constants.Unix, true},
+	}
+
+	for _, test := range tests {
+		s := validateOSSupportVMDriver(test.OS, test.driver)
+		if test.supports != s {
+			t.Errorf("Expected: %t got: %t for driver %s support on the OS %s", test.supports, s, test.driver, test.OS)
```
afbjorklund: "on the OS Linux" is slightly controversial. Might as well go for "on Linux"

cmd/minikube/cmd/start_test.go
```diff
@@ -43,3 +45,22 @@ func Test_extractVMDriverVersion(t *testing.T) {
...
+
+func Test_validateOSSupportVMDriver(t *testing.T) {
+	tests := []struct {
+		driver   string
+		OS       string
+		supports bool
+	}{
+		{constants.DriverNone, constants.Darwin, false},
+		{constants.DriverNone, constants.Windows, false},
+		{constants.DriverNone, constants.Unix, true},
```
afbjorklund: Linux

cmd/minikube/cmd/start_test.go
```diff
@@ -43,3 +45,22 @@ func Test_extractVMDriverVersion(t *testing.T) {
...
+
+func Test_validateOSSupportVMDriver(t *testing.T) {
+	tests := []struct {
+		driver   string
+		OS       string
+		supports bool
+	}{
+		{constants.DriverNone, constants.Darwin, false},
+		{constants.DriverNone, constants.Windows, false},
+		{constants.DriverNone, constants.Linux, true},
```
afbjorklund: Could add some tests for e.g. VirtualBox here ?

josedonizetti: of course! 

virtualbox is supported on all envs?
hyperkit darwin only?
kvm2 anything but darwin or widnows?

afbjorklund: I think vbox is ok/enough, it's not like we have any logic for the others. basically "not none"

it is supported on all three platforms, yes (if installed/without hyper-v, etc.etc but anyway)

medyagh:
it would be amazing if we have it in a way that we could generate a table out of it to the docs, automatically later.

like generate Supported Drivers doc

afbjorklund:
I already got into trouble for the "recommended" vs. "default" driver, so I think I will stay out of this one...

And "supported" adds another level of connotations, like what kind of support and how was it tested etc.

Basically VirtualBox is default, the "native virtualization" of libvirt/hyperv/hyperkit (will be) recommended.

Explaining "none" needs a page of its own already, and everything else is more like "good luck have fun"

medyagh:
I am prototyping a vm-free model, once that is ready we will have a way better story for none

cmd/minikube/cmd/start_test.go
josedonizetti: @afbjorklund I can add logic to those if it makes sense.

afbjorklund: I suppose we could lock the other three to each platform, basically kvm2 = linux

josedonizetti:
@medyagh Netlify failures all have `No space left on device`. Not related to the PR, correct?

medyagh:
@minikube-bot OK to test

medyagh:
@josedonizetti  yeah I don't think those errors are related to your changes

tstromberg:
With #4738 now merged - I imagine this PR should change in a way that doesn't require hardcoding drivers per platforms? Is this check even necessary now?

josedonizetti:
I still prefer an earlier warning about not supporting a driver. For example, the latest minikube with https://github.com/kubernetes/minikube/pull/4738 prints when trying to start kvm2 driver on OSX.

```
minikube start --vm-driver=kvm2
😄  minikube v1.2.0 on darwin (amd64)
⚠️  Error checking driver version: exec: "docker-machine-driver-kvm2": executable file not found in $PATH
🔥  Creating kvm2 VM (CPUs=2, Memory=2000MB, Disk=20000MB) ...
💡  unsupported driver: kvm2
```

With this PR, it will print:

```
minikube start --vm-driver=kvm2
😄  minikube v1.2.0 on darwin (amd64)
💣  The driver 'kvm2' is not supported on darwin
```

WDYT? @tstromberg 
If you think the first message is enough, go ahead and close this PR, pls.

cmd/minikube/cmd/start.go
```diff
@@ -949,3 +954,26 @@ func extractVMDriverVersion(s string) string {
 	v := strings.TrimSpace(matches[1])
 	return strings.TrimPrefix(v, version.VersionPrefix)
 }
+
+func validateOSSupportVMDriver(os, vmDriver string) bool {
+	switch vmDriver {
+	case constants.DriverNone:
```
tstromberg: Rather than duplicating the platform-to-driver map, could this be adjusted to use the new build-tag specific list?

josedonizetti:
@minikube-bot Ok to test

josedonizetti:
@tstromberg By changing it to use the `IsValidDriver` that cheks `SupportedDrivers` I can no longer write a unit test for it. Wdyt?

medyagh:
@josedonizetti I am ok with merging this PR after the tests pass, I think it adds a great value to detect failure as soon as possible

PR #4724, opened 2019-07-09
Extract Advice strings for localization

sharifelgamal:


pkg/minikube/extract/extract.go
```diff
@@ -173,6 +173,11 @@ func inspectFile(e *state) error {
 		return err
 	}
 
+	if strings.HasSuffix(e.filename, "err_map.go") {
```
tstromberg: This seems brittle in the case of the filename changing. Could you make this a compile-time failure if the path changes?

Two ideas: Use the filename as a constant in err_map, or `reflect.PkgPath()`. 

Thoughts?

pkg/minikube/extract/extract.go
```diff
@@ -173,6 +173,11 @@ func inspectFile(e *state) error {
...
+		extractAdvice(file, e)
+		return nil
```
tstromberg: Maybe just `return extractAdvice(...)`?

pkg/minikube/extract/extract.go
sharifelgamal: I felt it was easier to explicitly check whether the file existed so I went ahead and did that.

pkg/minikube/extract/extract.go
sharifelgamal: yeah. done.

PR #4729, opened 2019-07-10
Show more relevant information, when using none driver

### LINKED ISSUE #4524: The none driver is showing imaginary numbers for the VM
`* Creating none VM (CPUs=2, Memory=2048MB, Disk=20000MB) ...`

These are just from the config of the VM that we did **not** create.

It has nothing to do with the VM that we are actually running on ?

We should either remove this information, or show the live data.

Related implementation: #3574

`github.com/shirou/gopsutil`

afbjorklund:
* Don't lie about creating a VM with none driver
* Show "localhost" information for none driver

Before:
`🔥  Creating none VM (CPUs=2, Memory=2000MB, Disk=20000MB) ...`
After:
`🤹  Running on localhost (CPUs=4, Memory=7800MB, Disk=138379MB) ...`

Closes #4524

### DIFF
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -37,6 +37,9 @@ import (
 	"github.com/docker/machine/libmachine/state"
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
+	"github.com/shirou/gopsutil/cpu"
+	"github.com/shirou/gopsutil/disk"
+	"github.com/shirou/gopsutil/mem"
 	"github.com/spf13/viper"
 	cfg "k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/console"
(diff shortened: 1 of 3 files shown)


pkg/minikube/cluster/cluster.go
```diff
@@ -344,9 +347,21 @@ To disable this message, run [minikube config set ShowDriverDeprecationNotificat
...
 }
 
+func megs(bytes uint64) int {
+	return int(bytes / 1024 / 1024)
+}
+
 func createHost(api libmachine.API, config cfg.MachineConfig) (*host.Host, error) {
 	preCreateHost(&config)
-	console.OutStyle(console.StartingVM, "Creating %s VM (CPUs=%d, Memory=%dMB, Disk=%dMB) ...", config.VMDriver, config.CPUs, config.Memory, config.DiskSize)
+	if config.VMDriver != constants.DriverNone {
```
medyagh: just a nit, but maybe put this logic into a new helper function ? showHostInfo() ?

afbjorklund: Sounds like a good idea!

pkg/minikube/cluster/cluster.go
```diff
@@ -344,9 +347,21 @@ To disable this message, run [minikube config set ShowDriverDeprecationNotificat
...
+		console.OutStyle(console.StartingVM, "Creating %s VM (CPUs=%d, Memory=%dMB, Disk=%dMB) ...", config.VMDriver, config.CPUs, config.Memory, config.DiskSize)
+	} else {
+		i, _ := cpu.Info()
```
josedonizetti: I know it shouldn't be a problem here. But can we not ignore errors?

medyagh: good catch, lets  glog the error for the logs

afbjorklund:
This is also appropriate for the generic driver (#4734)

tstromberg:
Thank you!

Issue #4733, opened 2019-07-10
Support the "generic" driver, for user provided VMs


afbjorklund:
Some users don't want to use minikube to set up their virtual machine,
but prefer to provide their own pre-installed VM (with a standard distro).

We should support this scenario, and offer to install Kubernetes on it...
This way they can still get started locally, and also with a physical server.

These parameters are required: (defaults)
* IP Address
* SSH User (root)
* SSH Key
* SSH Port (22)

The provided user is supposed to have sudo rights, and all OS requirements.

User does **not** have to install docker or kubeadm, this will be done by "start":

* https://github.com/docker/docker-install

* https://github.com/kubernetes/kubernetes

The distribution should be one of the ones supported by machine/kubeadm:

* **Ubuntu 16.04**+
* Debian 9
* **CentOS 7**+
* RHEL 7

The server also needs to meet the minimum HW requirements (2 CPUs, 2000M Memory)

See https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/

----

Update: For now I think that Ubuntu 16.04 and CentOS 7 will be plenty, one deb and one rpm ?

They are also supported by all tools, and are the same platforms that [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-using-native-package-management)/[kubeadm](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl) targets:

"xenial" = Ubuntu 16.04
```
deb https://apt.kubernetes.io/ kubernetes-xenial main
```

"el7" = RHEL/CentOS 7
```
https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64
```

Other Linux distributions _could_ work, but is not something that I want to support downstream.

Eventually we might support Ubuntu 18.04 and CentOS 8, but probably not until later (2020?)

~Note that Fedora 28 could be used for testing, until CentOS 8 has been released. "Close enough".~
(See https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux#Relationship_with_Fedora for details)

* Ubuntu 16.04 LTS (xenial)
* Ubuntu 18.04 LTS (bionic)
* CentOS 7 / Fedora 18
* CentOS 8 / Fedora 28

afbjorklund:
The main use case is when the user's development machine is unable to run the virtual machines...
[...]

PR #4734, opened 2019-07-10

afbjorklund:

Implements #4733


medyagh:
could you provide an example with the PR so we have a context ?

medyagh:
I am thinking if we could use this in the VM-free model too, maybe if the generic driver supports exec-ing into a docker container ?

afbjorklund:
Typical, seems like the generic install (https://get.docker.com) is more strict about security...

```
If you would like to use Docker as a non-root user, you should now consider
adding your user to the "docker" group with something like:

  sudo usermod -aG docker vagrant

Remember that you will have to log out and back in for this to take effect!

WARNING: Adding a user to the "docker" group will grant the ability to run
         containers which can be used to obtain root privileges on the
         docker host.
         Refer to https://docs.docker.com/engine/security/security/#docker-daemon-attack-surface
         for more information.

```

That doesn't really work, since minikube normally assumes that user has group access:

```Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.39/version: dial unix /var/run/docker.sock: connect: permission denied```

afbjorklund:
Added the user to the group, but doesn't work with vagrant since it only uses one IP (i.e. 127.0.0.1)

`The connection to the server 127.0.0.1:8443 was refused - did you specify the right host or port?`

afbjorklund:
I think I will split this PR

afbjorklund:
This PR is not going to be merged (it keeps on breaking!), so mostly using it for testing...

PR #4738, opened 2019-07-11
Move supported drivers per platform

josedonizetti:
This is something `minishift` do and is pretty useful. Wdyt?

### DIFF
--- a/pkg/minikube/constants/constants_windows.go
+++ b/pkg/minikube/constants/constants_windows.go
@@ -1,4 +1,4 @@
-// +build windows
+// +build windows, !gendocs
 
 /*
 Copyright 2016 The Kubernetes Authors All rights reserved.
@@ -23,3 +23,11 @@ import (
 )
 
 var DefaultMountDir = homedir.HomeDir()
+
+// SupportedVMDrivers is a list of supported drivers on Windows.
+var SupportedVMDrivers = [...]string{
+	DriverVirtualbox,
+	DriverVmwareFusion,
+	DriverHyperv,
+	DriverVmware,
+}
--- a/Makefile
+++ b/Makefile
@@ -288,7 +288,7 @@ mdlint:
 	@$(MARKDOWNLINT) $(MINIKUBE_MARKDOWN_FILES)
 
 out/docs/minikube.md: $(shell find cmd) $(shell find pkg/minikube/constants) pkg/minikube/assets/assets.go pkg/minikube/translate/translations.go
-	go run -ldflags="$(MINIKUBE_LDFLAGS)" hack/help_text/gen_help_text.go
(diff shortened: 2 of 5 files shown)


afbjorklund:
To be honest I'm not sure why even bother with plugins, if we are hard-coding the list of "supported" ones anyway ? It's not like you can add your own, since `minikube' will just complain that it is not supported...

But since we are going the way of validating the drivers (#4720), then maybe this makes the code prettier/smaller. And first glance it just looks like duplication, but maybe that can be cleaned up a bit ?

josedonizetti:
@afbjorklund I don't see it the same way. I understood this `supported` drivers as something for the documentation, maybe the naming isn't good, but it is simple to change. I think the documentation should be as clean as possible, listing the KVM driver for a Windows user when he runs `minikube start --help` is misleading, and the idea here is to improve this documentation, not to force the use of only those drivers. That's the case for the PR https://github.com/kubernetes/minikube/pull/4720 as well, it should only validate that if someone tries the none driver in Darwin or Windows he is alerted this won't work, cos it's not for those environments, not that only those drivers are now supported and no one can try to create his one using a plugin.

afbjorklund:
I’m all for filtering out the invalid drivers (i.e. not available on platform)

pkg/minikube/constants/constants_darwin.go
```diff
@@ -19,3 +19,13 @@ limitations under the License.
...
 
 var DefaultMountDir = "/Users"
+
+// SupportedVMDrivers is a list of supported drivers on Darwin.
+var SupportedVMDrivers = [...]string{
+	DriverVirtualbox,
+	DriverParallels,
+	DriverVmwareFusion,
+	DriverXhyve,
+	DriverHyperv,
```
tstromberg: This should be HyperKit, not HyperV.

josedonizetti: great catch!

pkg/minikube/constants/constants_windows.go
```diff
@@ -23,3 +23,12 @@ import (
 )
 
 var DefaultMountDir = homedir.HomeDir()
+
+// SupportedVMDrivers is a list of supported drivers on Windows.
+var SupportedVMDrivers = [...]string{
+	DriverVirtualbox,
+	DriverParallels,
```
blueelvis: I am pretty sure that Parallels is not supported/can be used on Windows. Pretty sure it is only for Mac. Could you please confirm?

josedonizetti: Thanks!

afbjorklund:
Apparently it got cancelled. https://www.parallels.com/blogs/parallels-desktop-for-windows/

PR #4741, opened 2019-07-11
Use print() function in both Python 2 and Python 3

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

/kind bug
/release-note-none
Signed-off-by: cclauss <cclauss@me.com>

### DIFF
--- a/hack/prow/run_tests.py
+++ b/hack/prow/run_tests.py
@@ -20,6 +20,7 @@
    https://k8s-testgrid.appspot.com
 """
 
+from __future__ import print_function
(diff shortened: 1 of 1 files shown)


afbjorklund:
Since the script is run by python (not python3), I’m not sure it is needed ? But on the other hand, it probably doesn’t hurt either...

cclauss:
Python is Python 3.  Python 2 is legacy Python. ;-)

https://pythonclock.org

tstromberg:
Thank you for fixing this!

PR #4743, opened 2019-07-12
Reduce kubectl restart time from 10 seconds to 600ms

tstromberg:
This PR reduces VM start time by up to 9 seconds.

For complicated reasons, the kubelet has to be running before kubeadm, but will kill itself as it has not yet been configured. Inspired by kind, this adjusts the restart delay to fit better with a single-node dev cluster use case.

To go from "Starting kubelet" to "Watching apiserver":

10s (current): 26s
1s (kind): 19s
750ms: 18s
600ms: 17s
250ms: 17s


medyagh:
nice !!!!

Issue #4418, opened 2019-06-03
TestVersionUpgrade on none: error: failed to create listener... bind: address already in use


medyagh:
the TestVersionUpgrade randomly fails sometimes in the integration tests.
with error 

```
error: failed to create listener: failed to listen on 0.0.0.0:8443: listen tcp 0.0.0.0:8443: bind: address already in use
```

```
--- FAIL: TestVersionUpgrade (128.05s)
    util.go:130: Error running command: start --vm-driver=none  --v=10 --logtostderr --bootstrapper=kubeadm --kubernetes-version=v1.14.2 --alsologtostderr --v=2 . Output: * minikube v1.1.0 on linux (amd64)
        * minikube will upgrade the local cluster from Kubernetes 1.10.13 to 1.14.2
        * Tip: Use 'minikube start -p <name>' to create a new cluster, or 'minikube delete' to delete this one.
        * Restarting existing none VM for "minikube" ...
        * Waiting for SSH access ...
        * Configuring environment for Kubernetes v1.14.2 on Docker 18.06.1-ce
        * Pulling images ...
        * Relaunching Kubernetes v1.14.2 using kubeadm ... 
        * Problems detected in "kube-apiserver":
          - error: failed to create listener: failed to listen on 0.0.0.0:8443: listen tcp 0.0.0.0:8443: bind: address already in use
FAIL
```

medyagh:
and this is the random fail I got for OS X Virtualbox integration test:
http://mkproxy.cbf.corp.google.com/job/OSX_Integration_Tests_VirtualBox/5556/consoleFull

```

12:30:06 --- FAIL: TestVersionUpgrade (121.48s)
12:30:06     util.go:130: Error running command: start --vm-driver=virtualbox  --v=10 --logtostderr --bootstrapper=kubeadm --kubernetes-version=v1.10.13 --alsologtostderr --v=2 . Output: * minikube v1.1.0 on darwin (amd64)
12:30:06         2019/06/03 12:28:06 No matching credentials were found, falling back on anonymous
12:30:06         2019/06/03 12:28:06 No matching credentials were found, falling back on anonymous
12:30:06         2019/06/03 12:28:06 No matching credentials were found, falling back on anonymous
12:30:06         2019/06/03 12:28:06 No matching credentials were found, falling back on anonymous
12:30:06         2019/06/03 12:28:06 No matching credentials were found, falling back on anonymous
[...]

PR #4744, opened 2019-07-12
Allow users to skip pod health checks with --wait=false

tstromberg:
This saves about 1 minute of run time, and will be useful for integrating with other applications which may do their own health checking. This flag behaves the same as the similarly named flag in `kind`, with one important difference: we leave the default as `true` for a more predictable start-up sequence, at the expense of startup latency. This was also the behavior in minikube up until v1.0.

This PR also sets `--wait=false` to all tests where it's possible to begin testing without all of the services online, such as CoreDNS or etcd. As a result, this PR decreases integration testing time by ~9 minutes.


tstromberg:
@minikube-bot OK to test

afbjorklund:
If you change your mind (or your cluster is acting weird), can you run the health checks afterwards ?

tstromberg:
> If you change your mind (or your cluster is acting weird), can you run the health checks afterwards ?

Not yet, but you bring a good point. We should extend `status` to have the same model of pod healthiness as `start`. I'd prefer to introduce that as a second PR though, as it does pose some challenges to how the status command is currently implemented.

tstromberg:
@minikube-bot OK to test

The one failure here appears to be #4418 - just looking for more test data:

```
--- FAIL: TestVersionUpgrade (143.72s)
    util.go:130: Error running command: start --vm-driver=none  --v=10 --logtostderr --...
          - error: failed to create listener: failed to listen on 0.0.0.0:8443: listen tcp 0.0.0.0:8443: bind: address already in use
```

tstromberg:
@minikube-bot OK to test

tstromberg:
@minikube-bot OK to test

tstromberg:
PTAL - passes all tests, amazingly.

PR #4745, opened 2019-07-12
WIP: hyperv: Assert that the user has correct permissions

### LINKED ISSUE #4511: hyperv : Hyper-v commands have to be run as an Administrator

```
 Unable to start VM: create: precreate: Hyper-v commands have to be run as an Administrator  * Sorry that minikube crashed. If this was unexpected, we would love to hear from you:

```

blueelvis:
1. This adds a check to ensure that we are running Hyper-V with administrator privileges.
2. Virtualbox seems to work fine without administrator privileges.
3. Moved the `validateUser()` so that it is effective even when cluster is already provisioned.

Fixes #4511

-Pranav


pkg/util/utils.go
```diff
@@ -291,3 +293,31 @@ func ContainsString(slice []string, s string) bool {
 	}
 	return false
 }
+
+// ValidateUser validates minikube is run by the recommended user (privileged or regular)
```
tstromberg: I like the idea, but let's find a better home for it with a more appropriate function signature. What do you think about moving this to pkg/drivers/drivers.go so that the call can be `driver.ValidatePermissions()?

Since this function will no longer be in the `main` package, it should return an error instead of exiting.

pkg/util/utils.go
```diff
@@ -291,3 +293,31 @@ func ContainsString(slice []string, s string) bool {
...
+func ValidateUser(driverName string) {
+	u, err := user.Current()
+	// Check if minikube needs to run with sudo or not.
+	if err == nil {
+		if driverName == constants.DriverNone && u.Name != "root" {
```
tstromberg: Instead of checking for the username `root`, check that `u.Uid != 0` here an elsewhere. The root username can be renamed, and there can be multiple users with uid 0, and thus root privileges.

pkg/util/utils.go
```diff
@@ -291,3 +293,31 @@ func ContainsString(slice []string, s string) bool {
...
+			exit.Usage("Please run with sudo. The vm-driver %q requires sudo.", constants.DriverNone)
+
+		} else if u.Name == "root" && !(driverName == constants.DriverHyperv || driverName == constants.DriverNone) {
```
tstromberg: The username of "root" is meaningless in Windows (as is uid 0), so just simplify this check to not compare against HyperV.

pkg/util/utils.go
```diff
@@ -291,3 +293,31 @@ func ContainsString(slice []string, s string) bool {
...
+			console.OutStyle(console.WarningType, "Please don't run minikube as root or with 'sudo' privileges. It isn't necessary with %s driver.", driverName)
+		}
+
+		// If Hyper-V, ensure we are having Administrator Privilleges.
+		if driverName == constants.DriverHyperv {
+			// This tries opening the first disk attached to the system for a direct I/O operation. This operation requires Administrator privileges and will fail without that.
+			rawDisk, err := os.Open("\\\\.\\PHYSICALDRIVE0")
+			if err != nil {
+				exit.Usage("You need to run minikube with Administrator privileges as interacting with %q requires them.", constants.DriverHyperv)
```
tstromberg: This should be `exit.Permissions` rather than `exit.Usage`

pkg/util/utils.go
```diff
@@ -291,3 +293,31 @@ func ContainsString(slice []string, s string) bool {
...
+			exit.Usage("Please run with sudo. The vm-driver %q requires sudo.", constants.DriverNone)
+
+		} else if u.Name == "root" && !(driverName == constants.DriverHyperv || driverName == constants.DriverNone) {
+			console.OutStyle(console.WarningType, "Please don't run minikube as root or with 'sudo' privileges. It isn't necessary with %s driver.", driverName)
+		}
+
+		// If Hyper-V, ensure we are having Administrator Privilleges.
+		if driverName == constants.DriverHyperv {
+			// This tries opening the first disk attached to the system for a direct I/O operation. This operation requires Administrator privileges and will fail without that.
+			rawDisk, err := os.Open("\\\\.\\PHYSICALDRIVE0")
```
tstromberg: Add `defer rawDisk.Close()` so that the resources are cleaned up regardless of what branch is taken.

pkg/util/utils.go
```diff
@@ -291,3 +293,31 @@ func ContainsString(slice []string, s string) bool {
...
+			if err != nil {
+				exit.Usage("You need to run minikube with Administrator privileges as interacting with %q requires them.", constants.DriverHyperv)
+			} else {
+				console.OutStyle(console.Permissions, "Got administrator privilleges. I can handle Hyper-V now...")
```
tstromberg: I don't feel that this message isn't helpful to the end user. Perhaps make it a logging statement?

pkg/util/utils.go
tstromberg: I just realized that this may break some users: Hyper-V has a separate set of ACL's, and in fact there is support for a Hyper-V administrators group. Rather than requiring full administrative access, I think we should:

- Make this check more specific so that HyperV admins can continue to use minikube. Perhaps there we could call `Get-VM`, `Get-VMConnectAccess` or some other kind of command line instead to see if we have access to Hyper-V?
- Add a `--force` flag so that these checks can be skipped.

We can start with one of those approaches, but both would be ideal in the long term.

tstromberg:
@minikube-bot OK to test

pkg/util/utils.go
blueelvis: I was not aware that multiple users can have the same `User ID`. I always thought that `gid` can be same. I guess this is one of the special cases. Thanks for the info!

pkg/util/utils.go
blueelvis: Oops. Fixed.

pkg/util/utils.go
blueelvis: Do you think we could track this in a separate bug/ work item? I fully agree with this and I think that we could check what other approaches could be in this case instead of running raw Powershell commands. Not sure if we already have a package to execute Powershell commands as part of golang. Pretty sure that there would be a native Windows API to check this.

pkg/util/utils.go
blueelvis: I have moved this to the `drivers` file. I was wondering, do we have any middleware setup? With this type of implementation, I really don't like how in some places, I need to check the commandline arguments and in some places, what is loaded from the config file. If we don't have it, I think it would be best to have a middleware setup which can validate config/permissions after  arguments are saved to config file and we validate the config file only.

That way, in future, it would be easier to add more checks like `validatePermissions`. What do you think?

pkg/util/utils.go
blueelvis: @tstromberg - I was browsing around and I see that what you are asking is already part of the machine library - https://github.com/docker/machine/blob/a555e4f7a8f518a8b1b174824c377e46cbfc4fe2/drivers/hyperv/powershell.go#L71

Let me add that and push a commit.

pkg/drivers/drivers.go
```diff
@@ -136,3 +139,32 @@ func fixPermissions(path string) error {
 	}
 	return nil
 }
+
+// ValidatePermissions validates minikube is run by the recommended user (privileged or regular)
+func ValidatePermissions(driverName string) error {
```
josedonizetti: I don't think this function should be in this package. This is related to what @afbjorklund was taking in the issue https://github.com/kubernetes/minikube/issues/4909. Anything under `pkg/drivers` shouldn't be using minikube code directly, because the idea is that those drivers will be moved upstream. In this case, we have translations here. Drivers code related to minikube should be in `pkg/minikube/drivers`.

wdyt?

blueelvis: In one of the review comments above, Thomas mentioned that it would be better if we could place this function at this path.

pkg/util/utils.go
josedonizetti: Gotcha! No strong feelings about it. Although I think `pkg/drivers` should be drivers only, not using minikube code.

@tstromberg thoughts?

pkg/drivers/drivers.go
josedonizetti: Gotcha! No strong feelings about it. Although I think pkg/drivers should be drivers only, not using minikube code.

@tstromberg thoughts?

pkg/util/utils.go
blueelvis: @tstromberg - I noticed that these checks are private to the package for now. I am in the process of creating a PR to make those checks public and we can revisit this again. As of now, this PR will fix the issues with most of the client users. The very small user base which will be affected would be the ones running Hyper-V on Windows Server environments with that role.

WDYT?

blueelvis:
@tstromberg - ping. The integration tests are failing because now more output is coming instead of just `Running` -

```
 02:28:33 | > Running
--- FAIL: TestDocker (158.46s)
    util.go:248: got state: "! Please don't run minikube as root or with 'sudo' privileges. It isn't necessary with kvm2 driver.\n! Please don't run minikube as root or with 'sudo' privileges. It isn't necessary with kvm2 driver.\n! Please don't run minikube as root or with 'sudo' privileges. It isn't necessary with kvm2 driver.\nRunning", expected "Running"
```

I also cannot run the integration tests on my Windows box :( . Could you please guide me on where I need to make a fix for this?

tstromberg:
> @tstromberg - ping. The integration tests are failing because now more output is coming instead of just `Running` -
> 
> ```
>  02:28:33 | > Running
> --- FAIL: TestDocker (158.46s)
>     util.go:248: got state: "! Please don't run minikube as root or with 'sudo' privileges. It isn't necessary with kvm2 driver.\n! Please don't run minikube as root or with 'sudo' privileges. It isn't necessary with kvm2 driver.\n! Please don't run minikube as root or with 'sudo' privileges. It isn't necessary with kvm2 driver.\nRunning", expected "Running"
> ```
> 
> I also cannot run the integration tests on my Windows box :( . Could you please guide me on where I need to make a fix for this?

Is there a bug about integration tests on Windows? I'd like to make sure this gets sorted out. Presumably this should work even if `make` isn't available, but there are probably other dragons lurking:

`go test -v -test.timeout=60m ./test/integration --tags="integration container_image_ostree_stub containers_image_openpgp" -minikube-start-args='--vm-driver=hyperv' -test.run=TestDocker`

blueelvis:
@tstromberg @medyagh - If I run that command, I am getting the following. This is happening even after cleaning the packages and getting them again.
```
PS D:\dev\minikube> go test -v -test.timeout=60m ./test/integration --tags="integration container_image_ostree_stub containers_image_openpgp" -minikube-start-args='--vm-driver=hyperv --memory=1.3g --hyperv-virtual-switch="Default Switch"' -test.run=TestDocker
can't load package: package k8s.io/minikube: unknown import path "k8s.io/minikube": cannot find module providing package k8s.io/minikube
```

Any other way, this can be fixed?

medyagh:
> @tstromberg @medyagh - If I run that command, I am getting the following. This is happening even after cleaning the packages and getting them again.
> 
> ```
> PS D:\dev\minikube> go test -v -test.timeout=60m ./test/integration --tags="integration container_image_ostree_stub containers_image_openpgp" -minikube-start-args='--vm-driver=hyperv --memory=1.3g --hyperv-virtual-switch="Default Switch"' -test.run=TestDocker
> can't load package: package k8s.io/minikube: unknown import path "k8s.io/minikube": cannot find module providing package k8s.io/minikube
> ```
> 
> Any other way, this can be fixed?

sorry for the late reply, does this help ? https://minikube.sigs.k8s.io/docs/contributing/testing/

 env TEST_ARGS="-minikube-start-args=--vm-driver=hyperv -test.run TestStartStop" make integration

pkg/drivers/drivers.go
tstromberg: SGTM. I think this should go in `pkg/minikube/drivers/drivers.go` then?

blueelvis:
@medyagh - I tried running that in an Admin Git bash prompt but I am still not able to run the integration tests against Hyper-V :(
```
bluee@DESKTOP-8TI9P3N MINGW64 /d/dev/minikube (master)
$ env TEST_ARGS="-minikube-start-args=--vm-driver=hyperv -test.run TestStartStop" make integration
go test -v -test.timeout=60m ./test/integration --tags="integration container_image_ostree_stub containers_image_openpgp" -minikube-start-args=--vm-driver=hyperv -test.run TestStartStop
panic: Failed to define asset deploy/addons/gvisor/gvisor-runtimeclass.yaml: Asset deploy/addons/gvisor/gvisor-runtimeclass.yaml not found

goroutine 1 [running]:
k8s.io/minikube/pkg/minikube/assets.MustBinAsset(0x1a58de7, 0x2d, 0x1a2f85e, 0x16, 0x1a34147, 0x18, 0x1a13f2e, 0x4, 0x100, 0xc000524a80)
        D:/dev/minikube/pkg/minikube/assets/vm_assets.go:156 +0x1b2
k8s.io/minikube/pkg/minikube/assets.init.ializers()
        D:/dev/minikube/pkg/minikube/assets/addons.go:317 +0x23a5
FAIL    k8s.io/minikube/test/integration        0.528s
make: *** [Makefile:220: integration] Error 1
```

I am in the root folder of the project and I can see the file in the console but not sure why this is not able to find it. Any thoughts?

tstromberg:
@sharifelgamal has some experience here.

priyawadhwa:
Hey @blueelvis, we may have merged in a change that solves your issue ~3 weeks ago. Could you resolve your merge conflicts and try again? If you're still seeing issues, please ping @sharifelgamal.

blueelvis:
@priyawadhwa - Are you talking about the integration tests? Let me try again over the weekend and get back on this.

priyawadhwa:
@blueelvis yup I meant the integration tests, sorry for not clarifying! And sound great, thanks.

pkg/drivers/drivers.go
```diff
@@ -136,3 +139,31 @@ func fixPermissions(path string) error {
...
+			return errors.Errorf("Please run with sudo. The vm-driver %q requires sudo.", constants.DriverNone)
+
+		} else if !(driverName == constants.DriverHyperv || driverName == constants.DriverNone) {
+			out.T(out.WarningType, "Please don't run minikube as root or with 'sudo' privileges. It isn't necessary with {{.driver}} driver.", out.V{"driver": driverName})
+		}
+
+		// If Hyper-V, ensure we are having Administrator Privilleges.
+		if driverName == constants.DriverHyperv {
+			// This tries opening the first disk attached to the system for a direct I/O operation. This operation requires Administrator privileges and will fail without that.
+			rawDisk, err := os.Open("\\\\.\\PHYSICALDRIVE0")
```
tstromberg: Alternatively: https://coolaj86.com/articles/golang-and-windows-and-admins-oh-my/

Also, we should perhaps be checking for membership to the hyperv admins group.

blueelvis:
@priyawadhwa @tstromberg - I am closing this Pull Request. I will submit a new one after copying the functions from the machine-drivers fork. That would be really easy to achieve what is mentioned in the article as well.

Discussion - https://github.com/machine-drivers/machine/pull/20

Issue #4699, opened 2019-07-06
Provide tarballs for releases


afbjorklund:
In addition to the currently existing binaries, we should provide tarballs for smaller downloads.

Doesn't have to be anything fancy, just add .tar.gz with the same files. Kubernetes does this.

medyagh:
Do you mean tarball for the source code ?

afbjorklund:
We already have a `cross-tars` target to do this, and it is one tarball per platform.

We could have just one checksum file though (SHA512?), and one entry per line.

afbjorklund:
There's some _bugs_ with the current targets, though. It "forgets" to gzip the files, and includes kvm2 driver.

So I think we want to have a solution for the drivers first. And I have no idea why we include the add-ons ?

tstromberg:
Do you mind elaborating on the use case for the tarballs? I'm hesitant to introduce another download form on our release pages unless there is a clear benefit for users:

- If users want a managed minikube install, they use a packaged form
- If users want an unmanaged minikube install, they use the binary form

The tarball addresses the solution of how to download the kvm2 driver, but I'm not sure it otherwise differentiates itself enough to warrant a 3rd form of download, particularly as it would still need a form of installation. 

For what it's worth, I don't mind the tarball approach: I do in fact prefer it to the current approach of supplying an arbitrary executable (on macOS and Linux). If we adopt the tarball approach, we should discuss what it would take to deprecate the binary download approach.

Thoughts?

afbjorklund:
Some packaging formats prefer tarballs (in fact only support tar), so it might make it easier downstream.
I wasn't planning on deprecating the "raw" binaries, just make minikube more accessible (like deb or rpm)

The two benefits were:
- offering compression (60-70% smaller download), something like .gz or .xz
- offering checksum (verifiable download), something like MD5 or SHA-512

Both of these benefits are already in the packages, but since we can't see to settle on one standard...
The main idea was to make `minikube` as available as `kubectl`, in several packaging formats:

* exe (40M)
* tar (17M .tgz)
* apt (13M .deb)
* yum (11M .rpm)

Note: all of these formats have the exact same `minikube` binary, with the exact same checksum!
Unfortunately all those packaging standards checksum the archive and not the actual contents...

But I **don't** think that we want to remove the executable, since lots of users are using that already.
Another approach would of course be to delete the buggy old `cross-tars` target, rather than fixing it ?

medyagh:
[...]

Issue #4555, opened 2019-06-22
Latest release kvm2 driver still linking to newer libvirt


afbjorklund:
Seems that the driver was **not** built using the Docker image:

``` console
$ wget -q https://github.com/kubernetes/minikube/releases/download/v1.1.1/docker-machine-driver-kvm2
$ ldd docker-machine-driver-kvm2 
./docker-machine-driver-kvm2: /usr/lib/x86_64-linux-gnu/libvirt-lxc.so.0: version `LIBVIRT_LXC_2.0.0' not found (required by ./docker-machine-driver-kvm2)
./docker-machine-driver-kvm2: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_2.2.0' not found (required by ./docker-machine-driver-kvm2)
./docker-machine-driver-kvm2: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_3.0.0' not found (required by ./docker-machine-driver-kvm2)
./docker-machine-driver-kvm2: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_1.3.3' not found (required by ./docker-machine-driver-kvm2)
./docker-machine-driver-kvm2: /usr/lib/x86_64-linux-gnu/libvirt.so.0: version `LIBVIRT_2.0.0' not found (required by ./docker-machine-driver-kvm2)
```

It is supposed to still be using libvirt 1.3.1, for compatibility.

`FROM gcr.io/gcp-runtimes/ubuntu_16_0_4`

afbjorklund:
Related to ~~#3689~~ (regression)

afbjorklund:
Possibly a race condition, between the `make -j16` and the hacky ways of `kvm_in_docker`

medyagh:
I wonder if this is the root cause of all the "minikube stop" failing in KVM integration tests? https://github.com/kubernetes/minikube/issues/4495

I think it is important, I would be happy to review the PR fixing this

josedonizetti:
I'll take a look at this.

afbjorklund:
This issue affects **all** versions after v0.30.0:

```
docker-machine-driver-kvm2-v0.25.0	LIBVIRT_1.2.9
docker-machine-driver-kvm2-v0.25.1	LIBVIRT_1.2.9
docker-machine-driver-kvm2-v0.25.2	LIBVIRT_1.2.9
[...]

PR #4751, opened 2019-07-13
Improve the release tarballs with compression and checksum

afbjorklund:
* Make sure to actually gzip the tarballs (.tar.gz)
* Don't include the add-ons, but do include hyperkit

~* Add proper docker target for building kvm driver~ (--> PR #4779)
~* Improve libvirt linking and check ABI version (#4555)~  (--> PR #4779)

* Checksum the tarballs, ~and include in the release (#4699)~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,8 @@ $(shell mkdir -p $(BUILD_DIR))
 PYTHON := $(shell command -v python || echo "docker run --rm -it -v $(shell pwd):/minikube -w /minikube python python")
 BUILD_OS := $(shell uname -s)
 
+SHA512SUM=$(shell command -v sha512sum || echo "shasum -a 512")
+
 STORAGE_PROVISIONER_TAG := v1.8.1
 
 # Set the version information for the Kubernetes servers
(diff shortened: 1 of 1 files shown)


afbjorklund:
Need some help with pushing the `gcr.io/k8s-minikube/kvm-build-image` image

hack/jenkins/release_build_and_upload.sh
```diff
@@ -40,6 +40,9 @@ cat Makefile | grep "VERSION_BUILD ?=" | grep $VERSION_BUILD
 BUILD_IN_DOCKER=y make -j 16 all out/minikube-installer.exe out/minikube_${DEB_VERSION}.deb out/minikube-${RPM_VERSION}.rpm
```
tstromberg: While we're here - can this rule be made more specific than `all`?

`all` can unexpectedly when multiple rules share the same output name, such as driver rules which can run in docker, or outside of docker. I suspect this may be the root cause of #4555 - but even if it isn't, it will make the result more explicit.

afbjorklund:
I think I will split this PR

tstromberg:
@minikube-bot OK to test

afbjorklund:
This PR probably won't be merged either, at least not if we have to give up on the normal binaries.

medyagh:
@afbjorklund do you know the status of this PR? is this good to be merged ?

afbjorklund:
As it is now, no documentation changes.

The hyperkit changes are quite random, think they were only about make cleanup and target alignment ? (Probably should have been separate PR)

Issue #4673, opened 2019-07-03
Improve Error message when Kubernetes version asked is too old


apupier:
While investigating n issue I encountered, i discovered that i (mistakenly) mistyped a Kubernetes version and it was causing trouble which was not so easy to understand. It would be nice to improve the error message.

The issue encountered was https://github.com/kubernetes/minikube/issues/4670

@tstromberg is suggesting to add a line here to print a warning if the user requests a version (`nv`) less than `constants.OldestKubernetesVersion`:

https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/start.go#L682

to go further:

@tstromberg also suggested:
It doesn’t have to be in the same PR, but we should also check against `constants.NewestKubernetesVersion`. This check is slightly trickier, as we should allow users to specify newer build numbers than we know about (like 1.15.1 rather than 1.15.0), and only check the major.minor version.

medyagh:
Thank you for creating this issue, you pointed out correctly, we should not allow invalid kubernetes versions to too old and fail fast without crashing ! 

I will make sure there will be a PR to handle this, either myself or if you are interested to make a PR, I would be more than happy to review it.

thank you again for helping the user experience on minikube :)

medyagh:
here is a link for supported by kubeadm versions https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/

serhatcetinkaya:
Hi @medyagh, I'd like to work on this issue if that's ok for everyone.

medyagh:
sounds good @serhatcetinkaya I will assign you.

medyagh:
what I would like to see is it shoud exit with error with SUPER old kubernetes versions, and it should warn you with anything which is less than our supported versions.

for example currently kubernetes 1.9 still works with minkube even though it is not supported, but 1.2 does NOT work since it is super old.

tstromberg:
I spoke with @medyagh on this offline, and wanted to add my thoughts here:

- `constants.OldestKubernetesVersion` should reflect the oldest version that is known to pass integration tests. If 1.9 works (I've never tried), the constant should reflect it. Let's keep the code path simple rather than tracking two oldest values.

- All sanity checks should either be a warning, or skippable using a `--force` flag

tstromberg:
This shipped with minikube v1.4.

PR #4759, opened 2019-07-14
Exit if --kubernetes-version is older than the oldest supported version

serhatcetinkaya:
This pull request is intended as a fix for the issue described in #4673. 
If the version specified by the user is less than `constants.OldestKubernetesVersion` we print a warning to the user but continue execution, if the specified version is less than `v1.9.0` we exit with a proper message. 
Can you please review @medyagh :)


cmd/minikube/cmd/start.go
```diff
@@ -679,6 +679,20 @@ func validateKubernetesVersions(old *cfg.Config) (string, bool) {
 		return nv, isUpgrade
 	}
 
+	oldestVersion, err := semver.Make(strings.TrimPrefix(constants.OldestKubernetesVersion, version.VersionPrefix))
+	if err != nil {
+		exit.WithCode(exit.Data, "Unable to parse oldest Kubernetes version from constants: %v", err)
+	}
+
+	superOldVersionLimit := semver.Version{Major: 1, Minor: 9, Patch: 0}
```
tstromberg: Thanks for this PR!

If possible, I would rather not have two effective "oldest" constants. The intent of `constants.OldestKubernetesVersion` was to have a constant that defined the oldest possible version. If v1.9.x works, the variable should reflect that state.

Do you mind updating `constants.OldestKubernetesVersion` to v1.9.0 so that the intent is met and the logic simple? If it happens to fail tests, we can stick with v1.10.x and add a `--force` flag so that users can still bypass the check.

RA489:
@minikube-bot OK to test

cmd/minikube/cmd/start.go
serhatcetinkaya: Hi Thomas, thanks for the review.

As far as I understand this is the behavior requested in the original issue #4673. According to your input we can remove the check for `superOldVersionLimit` and just return a warning (without exit) to the user if the specified version is less than `constants.OldestKubernetesVersion`. 

Can you check this comment to make things clear about what needs to be done https://github.com/kubernetes/minikube/issues/4673#issuecomment-508633818

medyagh: Hi sorry for the confusion, I think I contributed to the confusion ! my fault !

Thomas is right, rethinking this I believe it would be hard to maintain two separate versions of supported kuberentes. 

the original reason we created this issue was someone was trying to start an oddly old kubernetes by mistake (1.2 instead of 1.12)


here is my current new suggestion based on @tstromberg input:

we need to exit on oldest supported kubernets version ( yes we can update the oldest possible one to 1.9) however we need to infrom the user in the exit message, that to override this check please add "--force"

tstromberg: @serhatcetinkaya - Yes, that sounds like the right way to go. I spoke to Medya about this yesterday and he seemed willing to accept it as well. I've added my thoughts to the bug. Thank you!

serhatcetinkaya: Thanks for the clarification @medyagh @tstromberg, I am working on this, I will submit as soon as I finish.

tstromberg: Huzzah! start.go now has support for a --force flag.  If you change the logic to use it - I think we should merge this ASAP. 

```
if viper.GetBool(force) {
  out.WarningT("Kubernetes {{.version}} is not supported by this release of minikube", ...)
} else {
  exit.WithCode(exit.Data, "Sorry, Kubernetes {{.version}} is not supported by this release of minikube")
}
```

Please let me know if you need any help.

serhatcetinkaya:
Hi, sorry for the long delay. I added an if check to control `--force` flag and take action accordingly. However, I noticed a strange thing; when I try to build it fails with following error:
```➜  minikube git:(master) ✗ MINIKUBE_BUILD_IN_DOCKER=y make cross
docker run --rm -e GOCACHE=/app/.cache -e IN_DOCKER=1 --user 501:20 -w /app -v /Users/serhatcetinkaya/workspace/minikube:/app -v /Users/serhatcetinkaya/go:/go --entrypoint /bin/bash k8s.gcr.io/kube-cross:v1.12.9-1 -c '/usr/bin/make out/minikube-linux-amd64'
Unable to find image 'k8s.gcr.io/kube-cross:v1.12.9-1' locally
v1.12.9-1: Pulling from kube-cross
4ae16bd47783: Pull complete
bbab4ec87ac4: Pull complete
2ea1f7804402: Pull complete
96465440c208: Pull complete
16a3d8aca6cd: Pull complete
e0ec5610455a: Pull complete
96d705baf026: Pull complete
28bcdfdf2e97: Pull complete
c0e213211a87: Pull complete
8484dd64d83a: Pull complete
0a2bb18ece41: Pull complete
ce095b3f0e42: Pull complete
c04dacc9a7b7: Pull complete
72a4df4fe9f4: Pull complete
Digest: sha256:290ba56289b9a587ab5d2d53962ef78e3318237b4a9ea8a90b312ca8e9841b12
Status: Downloaded newer image for k8s.gcr.io/kube-cross:v1.12.9-1
GOOS="linux" GOARCH="amd64" go build -tags "container_image_ostree_stub containers_image_openpgp" -ldflags="-X k8s.io/minikube/pkg/version.version=v1.4.0-beta.0 -X k8s.io/minikube/pkg/version.isoVersion=v1.4.0-beta.0 -X k8s.io/minikube/pkg/version.isoPath=minikube/iso -X k8s.io/minikube/pkg/version.gitCommitID="578e32559b5af90eb26e7dbf8ee819c7fed39a70-dirty"" -a -o out/minikube-linux-amd64 k8s.io/minikube/cmd/minikube
go: finding git.apache.org/thrift.git v0.0.0-20180902110319-2566ecd5d999
go: git.apache.org/thrift.git@v0.0.0-20180902110319-2566ecd5d999: git fetch -f https://git.apache.org/thrift.git refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /go/pkg/mod/cache/vcs/83dba939f95a790e497d565fc4418400145a1a514f955fa052f662d56e920c3e: exit status 128:
	fatal: unable to access 'https://git.apache.org/thrift.git/': Failed to connect to git.apache.org port 443: Connection refused
go: error loading module requirements
make: *** [Makefile:145: out/minikube-linux-amd64] Error 1
make: *** [out/minikube-linux-amd64] Error 2```
I guess this is because of a vote to use the Github repos as primary repo and this one is moved to Github. Do you have any solutions for this ?

tstromberg:
Looks great. Thank you!

PR #4762, opened 2019-07-15
Undefined name 'test' in run_tests.py

cclauss:
This issue was hidden by the syntax error in #4741

$ __time flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics__
```
./hack/prow/run_tests.py:28:32: F821 undefined name 'test'
  classname = os.path.basename(test).split('.')[0]
                               ^
1     F821 undefined name 'test'
1
```

### DIFF
--- a/hack/prow/run_tests.py
+++ b/hack/prow/run_tests.py
@@ -25,7 +25,7 @@ import os, sys, json, re, argparse, calendar, time, subprocess, shlex
 
 def get_classname(test_script):
   """ parse out the test classname from the full path of the test script"""
-  classname = os.path.basename(test).split('.')[0]
+  classname = os.path.basename(test_script).split('.')[0]
   return classname
 
 def write_results(outdir, started, finished, test_results):


tstromberg:
good catch! Thank you.

afbjorklund:
Actually it was only "hidden" if you used python3 instead of python, when running flake8.

```
hack/prow/run_tests.py:90:15: E901 SyntaxError: invalid syntax
    print line
              ^
1       E901 SyntaxError: invalid syntax
1
```
When running flake8 with python (i.e. python2), it showed the actual error (here on v1.2.0):

```
hack/prow/run_tests.py:27:32: F821 undefined name 'test'
  classname = os.path.basename(test).split('.')[0]
                               ^
1       F821 undefined name 'test'
1
```

Fix itself is fine.

cclauss:
Yes.  Se the very important note at the top of http://flake8.pycqa.org  Python and legacy Python have different abstract syntax trees and therefor flake8 generates different results on the two versions.  The former issue is a Syntax Error on Python but is perfectly acceptable on legacy Python.  The _undefined name_ is a problem in all versions of Python.

PR #4777, opened 2019-07-16
Add test to make sure translations files are up to date.

sharifelgamal:


pkg/minikube/extract/extract_test.go
```diff
@@ -79,3 +80,53 @@ func TestExtract(t *testing.T) {
...
+		t.Fatalf("Error translating strings: %v", err)
+	}
+
+	dest, err := ioutil.ReadFile(tempfile)
+	if err != nil {
+		t.Fatalf("Reading resulting json file: %v", err)
+	}
+
+	if bytes.Compare(src, dest) != 0 {
+		t.Fatalf("There are changes to the localizable strings in the code base.\n Please run `make extract`.")
```
tstromberg: This should show a diff. `cmp.Diff` may be helpful here: I forget how it handles multi-line strings, but hopefully well.

```
if diff := cmp.Diff(src, dst); diff != "" {
  t.Errorf("Localized string mismatch (-want, +got):\n%s", diff)
}
```

sharifelgamal: yep, cmp.Diff works as long as I unmarshal the file contents back to json so the result is human-readable. fixed.

tstromberg:
It looks like it works - though you'll need to run the extraction to make it pass tests:

```
--- FAIL: TestTranslationsUpToDate (1.32s)
    extract_test.go:143: Localized string mismatch (-want, +got):
        root["Running on localhost (CPUs=%d, Memory=%dMB, Disk=%dMB) ..."]:
        	-: <non-existent>
        	+: ""
```

PR #4790, opened 2019-07-17
Update github issues URL now that we have issue templates

tstromberg:
Now that we support issue templates, https://github.com/kubernetes/minikube/issues/new points to a blank template. This switches the URL to the template chooser.

This PR makes me sad.


afbjorklund:
But I suppose it makes sense that the CLI URL goes to the same place as the "New Issue" web button...

PR #4811, opened 2019-07-18
Add a new command (profile list)

medyagh:
This PR add a new command to minikube, profile list that looks like this
```
 profile list
|---------|------------|----------------|-----------|--------------------|
| Profile | VM Driver  |     NodeIP     | Node Port | Kubernetes Version |
|---------|------------|----------------|-----------|--------------------|
| m1      | hyperkit   | 192.168.64.75  |      8443 | v1.15.0            |
| m4      | virtualbox | 192.168.99.136 |      8443 | v1.15.0            |
|---------|------------|----------------|-----------|--------------------|

```
if there be invalid profiles it will warn this way:
```
medya@minikube$ ./out/minikube profile list
|---------|------------|----------------|-----------|--------------------|
| Profile | VM Driver  |     NodeIP     | Node Port | Kubernetes Version |
|---------|------------|----------------|-----------|--------------------|
| p1      | hyperkit   | 192.168.64.75  |      8443 | v1.15.0            |
| p2      | virtualbox | 192.168.99.136 |      8443 | v1.15.0            |
|---------|------------|----------------|-----------|--------------------|
⚠️  Found 3 invalid profile(s) ! 
         p3_empty
         p4_invalid_file
         p5_partial_config
         $ minikube delete -p p3_empty 
         $ minikube delete -p p4_invalid_file 
         $ minikube delete -p p5_partial_config 
```

it will also print an err to the logs, if there is a not valid config profile (empty dir)


sharifelgamal:
Could you add a unit test of profile.go? Just a basic one that makes sure you can find all the profiles in a given directory?

cmd/minikube/cmd/config/profile_list.go
```diff
@@ -0,0 +1,55 @@
...
+	"k8s.io/minikube/pkg/minikube/config"
+
+	"github.com/olekukonko/tablewriter"
+	"github.com/spf13/cobra"
+)
+
+var profileListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "Lists all minikube profiles",
+	Long:  "Lists all valid minikube profiles",
```
tstromberg: Is the word 'valid' helpful for users? If not, remove it.

cmd/minikube/cmd/config/profile_list.go
```diff
@@ -0,0 +1,55 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+
+		var tData [][]string
+
+		table := tablewriter.NewWriter(os.Stdout)
+		table.SetHeader([]string{"Profile", "VM Driver", "NodeIP", "Node Port", "Kubernetes Version"})
```
tstromberg: Can you tell me more about the usefulness of `Node Port`? I'd be surprised if there were users manually accessing it.

cmd/minikube/cmd/config/profile_list.go
```diff
@@ -0,0 +1,55 @@
...
+		table.SetAutoFormatHeaders(false)
+		table.SetBorders(tablewriter.Border{Left: true, Top: true, Right: true, Bottom: true})
+		table.SetCenterSeparator("|")
+
+		for _, p := range config.AllProfiles() {
+			tData = append(tData, []string{p.Name, p.Config.MachineConfig.VMDriver, p.Config.KubernetesConfig.NodeIP, strconv.Itoa(p.Config.KubernetesConfig.NodePort), p.Config.KubernetesConfig.KubernetesVersion})
+		}
+
+		table.AppendBulk(tData) // Add Bulk Data
```
tstromberg: Remove the comment - it's a "what" rather than "why".

pkg/minikube/config/profile.go
```diff
@@ -0,0 +1,70 @@
...
+func AllProfiles() (profiles []*Profile) {
+
+	pDirs, err := allProfileDirs()
+	if err != nil {
+		glog.Infof("Error getting profile directories %v", err)
+	}
+	for _, n := range pDirs {
+		p, err := loadProfile(n)
+		if err != nil {
+			// TODO warn user to delete this folder or maybe do it for them
```
tstromberg: I don't find this comment useful. I would definitely not expect a list command mutate data.

pkg/minikube/config/profile.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/golang/glog"
+	"k8s.io/minikube/pkg/minikube/constants"
+)
+
+// AllProfiles returns all minikube profiles
+func AllProfiles() (profiles []*Profile) {
```
tstromberg: Unless there is a different function that returns a subset of profiles, I would rather see this function named ListProfiles or Profiles.

pkg/minikube/config/profile.go
```diff
@@ -0,0 +1,70 @@
...
+
+	cfg, err := DefaultLoader.LoadConfigFromFile(n)
+	profile := &Profile{
+		Name:   n,
+		Config: cfg,
+	}
+	return profile, err
+}
+
+// allProfileDirs gets all the folders in the user's profiles folder
```
tstromberg: Consistently use "folder" or "dir" for terminology. I recommend the latter, as Go uses that term.

Also, consider naming this profileDirs(), since there is no subset version of this function.

pkg/minikube/config/profile.go
```diff
@@ -0,0 +1,70 @@
...
+func allProfileDirs() (dirs []string, err error) {
+	root := filepath.Join(constants.GetMinipath(), "profiles")
+	err = filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
```
tstromberg: If it isn't necessary to recurse, consider using ioutil.ReadDir().

pkg/minikube/config/profile.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/golang/glog"
+	"k8s.io/minikube/pkg/minikube/constants"
+)
+
+// AllProfiles returns all minikube profiles
```
tstromberg: All newly added public functions need unit tests.

cmd/minikube/cmd/config/profile_list.go
```diff
@@ -0,0 +1,55 @@
...
+	"os"
+	"strconv"
+
+	"k8s.io/minikube/pkg/minikube/config"
+
+	"github.com/olekukonko/tablewriter"
+	"github.com/spf13/cobra"
+)
+
+var profileListCmd = &cobra.Command{
```
tstromberg: New commands should get integration tests, or be listed as experimental.

pkg/minikube/config/profile.go
```diff
@@ -0,0 +1,70 @@
...
+func AllProfiles() (profiles []*Profile) {
+
+	pDirs, err := allProfileDirs()
+	if err != nil {
+		glog.Infof("Error getting profile directories %v", err)
```
tstromberg: This should probably return an error.

cmd/minikube/cmd/config/profile_list.go
medyagh: I think it is nice, to see, incase someone configure their clusters with different ports. (for testing kuberentes) it comes at a cheap cost, just one column.

pkg/minikube/config/profile.go
medyagh: added

cmd/minikube/cmd/config/profile_list.go
medyagh: will add integration test

sharifelgamal:
The test failure looks to be the unrelated registry failure we've been seeing. LGTM.

PR #4813, opened 2019-07-19
Make cluster_test and utils_test run in isolation

tstromberg:
This PR reduces "make test" latency on Travis from 9.1 minutes to 3.6 minutes, by making the following changes:

- cluster_test: Consistently use MockDriver instead of DefaultVMDriver ("virtualbox")
- api_mock: Use a random profile name to avoid operating against the default ("minikube")
- utils: Remove unused ParseSHAFromURL and it's ill-conceived test that takes >30s to execute


tstromberg:
@minikube-bot OK to test

medyagh:
wow this is amazing ! it  finished in 4 min 24 sec ! it used to be 9.5 mins.  thank you sooo much for this.

pkg/util/utils_test.go
```diff
@@ -88,45 +85,6 @@ func TestRetryNotRetriableError(t *testing.T) {
...
 
-type getTestArgs struct {
-	input         string
-	expected      string
-	expectedError bool
-}
-
-var testSHAString = "test"
-
-func TestParseSHAFromURL(t *testing.T) {
```
sharifelgamal: do we no longer need this test?

tstromberg: I didn't find the test valuable, as the function shouldn't have been public in the first place. I took a second look, and discovered that this function is entirely unused, so I've opted to remove it entirely. Thanks for questioning it!

tstromberg:
NOTE: Next time any of ya'll are tempted to start an HTTP server for tests, please use: https://golang.org/pkg/net/http/httptest/

PR #4825, opened 2019-07-19
Console output cleanup

sharifelgamal:
- Removed all non-templated output functions: OutStyle, ErrStyle, Failure, Fatal, Warning, Success, Usage, WithCode

- Moved package console to out for brevity, renamed: console.OutT -> out.T, console.OutLn -> out.Ln, console.Out -> out.String, console.Arg -> out.V.


cmd/minikube/cmd/ip.go
```diff
@@ -51,7 +51,7 @@ var ipCmd = &cobra.Command{
 		if err != nil {
 			exit.WithError("Error getting IP", err)
 		}
-		console.Out(ip)
+		out.Out(ip)
```
tstromberg: Not your fault, but while you are here, this should actually be `out.Ln`. We accidentally broke compatibility with the translation: something I noticed when refactoring TestRegistry.

cmd/minikube/cmd/config/prompt.go
```diff
@@ -36,7 +36,7 @@ func AskForYesNoConfirmation(s string, posResponses, negResponses []string) bool
 	reader := bufio.NewReader(os.Stdin)
 
 	for {
-		console.Out("%s [y/n]: ", s)
+		out.Out("%s [y/n]: ", s)
```
tstromberg: Find a stutter-free replacement name for out.Out. Perhaps something likue `out.F` or `out.String`?

cmd/minikube/cmd/ip.go
sharifelgamal: Fixed.

cmd/minikube/cmd/config/prompt.go
sharifelgamal: yeah makes sense. I prefer String, since it's not a widely called function.

sharifelgamal:
@minikube-bot OK to test

Issue #4788, opened 2019-07-17
[FEATURE] allow to set container-runtime via config


csantanapr:
I know there is a [flag](https://github.com/kubernetes/minikube/blob/master/docs/alternative_runtimes.md) `minikube start --container-runtime=containerd` for start to use containerd but I would like to save the option to persistent config in hy $HOME directory along with other options I use all the time like cpu, memory, kubenertes-version, and vm-driver

**The exact command to reproduce the issue**:
```
minikube config set container-runtime containerd
```

**The full output of the command that failed**:<details>

💣  Set failed: property name "container-runtime" not found

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
👉  https://github.com/kubernetes/minikube/issues/new

</details>

medyagh:
This is a great suggestions, I am suprised we already do not save the config set for the next run... thank you for creating this issue. I will make sure this gets attention ! I would be happy to review a PR that makes sure the config set actually does save the config for next run !

medyagh:
@csantanapr are you starting your minikube using a  -p profile ? or just the default profile?

csantanapr:
@medyagh 
>I thought the config set would have saved the config set for the next run

I don't understand the term `next run` because it did not make far enough to be written into the `$HOME/.minikube/config/config.json`

The `minkube set config` command is not doing the save because `container-runtime` is not part of the list of settings in https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/config/config.go#L48

csantanapr:
Woot !

PR #4834, opened 2019-07-22
[FEATURE] allow to set container-runtime via config

RA489:
allow to set container-runtime via config
fixes https://github.com/kubernetes/minikube/issues/4788


medyagh:
It seems like it needs to be goimported

RA489:
> please goimport -w the files to pass the lint test

done.

Issue #4629, opened 2019-06-28
[FEATURE] insecure-registry via minikube config set ...


dcecile:
I would love to be able to set defaults for --insecure-registry just like for --vm-driver etc.

```
$ minikube config set insecure-registry foo
💣  Set failed: property name "insecure-registry" not found
````

(Similar to #3805, hopefully same simple & easy fix as #3861?)

PR #4844, opened 2019-07-23
[FEATURE] insecure-registry via minikube config set 

RA489:
insecure-registry via minikube config set 
fixes https://github.com/kubernetes/minikube/issues/4629

### DIFF
--- a/cmd/minikube/cmd/config/config.go
+++ b/cmd/minikube/cmd/config/config.go
@@ -176,6 +176,12 @@ var settings = []Setting{
 		callbacks:   []setFn{EnableOrDisableAddon},
 	},
 	{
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/config/config.go
```diff
@@ -175,6 +175,12 @@ var settings = []Setting{
 		validations: []setFn{IsValidAddon},
 		callbacks:   []setFn{EnableOrDisableAddon},
 	},
+	{
```
kumarom: Making a change here will just expose another addon.. Not a config set. I could be wrong but can you double check.

dcecile:
Thanks @RA489! 🙏

PR #4850, opened 2019-07-23
Limit the number of golangci-lint jobs to one

afbjorklund:
If you want the old behaviour back, use:
e.g. `make lint GOLINT_JOBS=$(nproc)`

Running multiple linters takes a lot of RAM.
And it only saves us like 15 seconds anyway ?

At least on my old laptop:
4 CPU: 16s
2 CPU: 20s
1 CPU: 32s


afbjorklund:
Unfortunately it still ran out of memory, even with one single job (!)

Have to increase the garbage collection, which makes it even slower.

So now it finishes in 1m 41s on the laptop. ☹️

But at least it passes "ok" and doesn't crash...

Issue #4851, opened 2019-07-23
CI: golangci-lint out of memory 


medyagh:
we had this issue yesterday, for two PRs but it was flaky and not consistant error:

today I saw it in travis again : here https://travis-ci.org/kubernetes/minikube/jobs/562528986 
```

goroutine 11121 [runnable]:
github.com/golangci/golangci-lint/pkg/golinters/goanalysis/checker.execAll.func1(0xc0a63ed400)
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/golinters/goanalysis/checker/checker.go:290
created by github.com/golangci/golangci-lint/pkg/golinters/goanalysis/checker.execAll
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/golinters/goanalysis/checker/checker.go:297 +0x11b
goroutine 11206 [semacquire]:
sync.runtime_SemacquireMutex(0xc0a63ed404, 0xc001d0f000)
	/home/travis/.gimme/versions/go1.12.5.linux.amd64/src/runtime/sema.go:71 +0x3d
sync.(*Mutex).Lock(0xc0a63ed400)
	/home/travis/.gimme/versions/go1.12.5.linux.amd64/src/sync/mutex.go:134 +0x109
sync.(*Once).Do(0xc0a63ed400, 0xc0a6500790)
	/home/travis/.gimme/versions/go1.12.5.linux.amd64/src/sync/once.go:40 +0x3b
github.com/golangci/golangci-lint/pkg/golinters/goanalysis/checker.(*action).exec(0xc0a63ed400)
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/golinters/goanalysis/checker/checker.go:303 +0x50
github.com/golangci/golangci-lint/pkg/golinters/goanalysis/checker.execAll.func1(0xc0a63ed400)
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/golinters/goanalysis/checker/checker.go:291 +0x34
created by github.com/golangci/golangci-lint/pkg/golinters/goanalysis/checker.execAll
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/golinters/goanalysis/checker/checker.go:297 +0x11b

```

the error is reported in: it seems to be the root cause of the problem:
https://github.com/golangci/golangci-lint/issues/483

medyagh:
The work arround for this issue is setting the "GOGC" env variable
[...]

PR #4852, opened 2019-07-23
Separate make lint targets for developers vs travis

medyagh:
Hopefully closes https://github.com/kubernetes/minikube/issues/4851 without making the 10 minutes. 
address the golangci-lint out of memory flaky fails

currently it runs at  5 min 57 sec

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,9 +48,9 @@ KERNEL_VERSION ?= 4.16.14
 GO_VERSION ?= $(shell go version | cut -d' ' -f3 | sed -e 's/go//')
 GOLINT_VERSION ?= v1.17.1
 # Limit number of default jobs, to avoid the CI builds running out of memory
-GOLINT_JOBS ?= 1
+GOLINT_JOBS ?= 2
 # see https://github.com/golangci/golangci-lint#memory-usage-of-golangci-lint
-GOLINT_GOGC ?= 8
+GOLINT_GOGC ?= 5
 
 export GO111MODULE := on
 
--- a/Makefile
+++ b/Makefile
@@ -280,6 +280,16 @@ out/linters/golangci-lint:
 
 .PHONY: lint
 lint: pkg/minikube/assets/assets.go pkg/minikube/translate/translations.go out/linters/golangci-lint
+	GOGC=100 ./out/linters/golangci-lint run \
+	  --deadline 4m \
+	  --build-tags "${MINIKUBE_INTEGRATION_BUILD_TAGS}" \
+	  --enable goimports,gocritic,golint,gocyclo,interfacer,misspell,nakedret,stylecheck,unconvert,unparam \
+	  --exclude 'variable on range scope.*in function literal|ifElseChain' \
+	  ./...
+
+# lint-ci is slower than lint and is meant to be used in ci (travis) to avoid out of memory leaks.
+.PHONY: lint-ci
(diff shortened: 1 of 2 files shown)


josedonizetti:
@medyagh did any build broke after @afbjorklund fix https://github.com/kubernetes/minikube/pull/4850 this morning?

medyagh:
> t least locally, only adding GOLINT_GOGC was enough to drop the memory usage, GOLINT_JOBS could be set to the previous value.

it didn't I made this PR before I saw ander's changes but I also noticed it added the test time from 5 to 10, I believe this PR will bring it back to 5 minutes and hopefully still fix the issue

medyagh:
I am thinking to make a lint target just for travis with limited ram, and let the default make lint be memory consumer for contributors, so the contributors dont face slow linting on their machine

Makefile
```diff
@@ -280,6 +280,16 @@ out/linters/golangci-lint:
 
 .PHONY: lint
 lint: pkg/minikube/assets/assets.go pkg/minikube/translate/translations.go out/linters/golangci-lint
+	GOGC=100 ./out/linters/golangci-lint run \
```
afbjorklund: `GOGC=100` is the default, remove

Makefile
```diff
@@ -280,6 +280,16 @@ out/linters/golangci-lint:
...
+	  --deadline 4m \
+	  --build-tags "${MINIKUBE_INTEGRATION_BUILD_TAGS}" \
+	  --enable goimports,gocritic,golint,gocyclo,interfacer,misspell,nakedret,stylecheck,unconvert,unparam \
```
afbjorklund: Maybe a variable for these ?

test.sh
```diff
@@ -23,7 +23,7 @@ go mod download 2>&1 | grep -v "go: finding" || true
 go mod tidy -v && echo ok || ((exitcode+=2))
 
 echo "= make lint ============================================================="
-make -s lint && echo ok || ((exitcode+=4))
+make -s lint-travis && echo ok || ((exitcode+=4))
```
afbjorklund: Think you meant “lint-ci”

afbjorklund:
I don’t really run golangci-lint locally because it is such a beast, I prefer golint. So prefer the “reportcard”, and let CI do the extended checks...

afbjorklund:
> I am down for using a standard lint for all developers, do you mind sharing your lint options. in my case, I like it as pre-commit hook. it is nice to not to commit if it will fail on travis.

Basically I think `make golint` covers it, but it is only `make gofmt` that is *mandatory*.

The authors of golint explicitly say that they do **not** want it to be blocking anything:

> The suggestions made by golint are exactly that: suggestions. Golint is not perfect, and has both false positives and false negatives. Do not treat its output as a gold standard. We will not be adding pragmas or other knobs to suppress specific warnings, so do not expect or require code to be completely "lint-free". In short, this tool is not, and will never be, trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process.

And authors of golangci-lint recommends that it is **not** run locally, but by the system:

> Local installation is not recommended for your CI pipeline. Only install the linter this way in a local development environment.

----

Other than that, I expect **all** commits to pass through `gofmt` and `go vet`,
but also hopefully through golint and gocyclo and some of the other checks.

Adding pre-commit hooks is up to each commiter, but they need to be fast.
As long as `make lint` and `make test` are so heavy as they are now, it's out.


For me they now take 16s and 1m20s respectivelym and use 2-3G of memory.

That's OK to run every once in a while, but not on every single commit or edit.

Issue #4868, opened 2019-07-25
Provide a different way of building images with minikube


afbjorklund:
Currently we delegate all building of images to `docker`, using `minikube docker-env`.

This requires the user to install Docker on their machine, and then learn how to set it up...

https://kubernetes.io/docs/tutorials/hello-minikube/

> [minikube/Dockerfile](https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/minikube/Dockerfile)
```Dockerfile
FROM node:6.14.2
EXPOSE 8080
COPY server.js .
CMD node server.js
```

> [minikube/server.js ](https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/minikube/server.js)
```js
var http = require('http');

var handleRequest = function(request, response) {
  console.log('Received request for URL: ' + request.url);
  response.writeHead(200);
  response.end('Hello World!');
};
var www = http.createServer(handleRequest);
www.listen(8080);
```

> For more information on the docker build command, read the [Docker documentation](https://docs.docker.com/engine/reference/commandline/build/).

If the user doesn't already have a local installation of docker, they can't build the image!

----

We could do better, by providing an abstraction that will simply do the build for them:

```
$ minikube build -- -t hello-node minikube
💾  Downloading docker 18.09.8
Sending build context to Docker daemon  3.072kB
Step 1/4 : FROM node:6.14.2
6.14.2: Pulling from library/node
3d77ce4481b1: Pull complete 
7d2f32934963: Pull complete 
0c5cf711b890: Pull complete 
9593dc852d6b: Pull complete 
4e3b8a1eb914: Pull complete 
ddcf13cc1951: Pull complete 
2e460d114172: Pull complete 
d94b1226fbf2: Pull complete 
Digest: sha256:62b9d88be259a344eb0b4e0dd1b12347acfe41c1bb0f84c3980262f8032acc5a
Status: Downloaded newer image for node:6.14.2
 ---> 00165cd5d0c0
Step 2/4 : EXPOSE 8080
 ---> Running in 2a302085e433
Removing intermediate container 2a302085e433
 ---> 9172f65af846
Step 3/4 : COPY server.js .
 ---> 035625e5e23f
Step 4/4 : CMD node server.js
 ---> Running in 1771091ed23a
Removing intermediate container 1771091ed23a
 ---> 250208286ec5
Successfully built 250208286ec5
Successfully tagged hello-node:latest
```
[...]

PR #4869, opened 2019-07-25
WIP: Add a build command, for building container images

afbjorklund:
From #4868

### DIFF
--- a/cmd/minikube/cmd/build.go
+++ b/cmd/minikube/cmd/build.go
@@ -20,26 +20,25 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"runtime"
 	"syscall"
 
 	"github.com/golang/glog"
 	"github.com/spf13/cobra"
-	pkg_config "k8s.io/minikube/pkg/minikube/config"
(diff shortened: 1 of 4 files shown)


medyagh:
This is reall cool ! I assume the built image will be available inside minikube.  Is there any way they could push that image to outside minikube too ?

Could you provide couple example real world use cases that you have in your mind ?

afbjorklund:
> This is reall cool ! I assume the built image will be available inside minikube. Is there any way they could push that image to outside minikube too ?

Sure, but unfortunately `crictl` only knows "pull" - it doesn't know "tag" or "push".
So just like with the missing "load" command, one would have to talk to the engine.

That is: after building the image, you would use either docker or podman to upload.
Of course, you would also have to sort out the credentials so probably need "login".
 
> Could you provide couple example real world use cases that you have in your mind ?

I prefer to discuss the implementation here, but can do some more examples in #4869

The basic use case is just being able to convert a `Dockerfile` into a container image.

afbjorklund:
I'm going to close this one as a PoC, until we some more buy-in on the concept (#4868)

Issue #4889, opened 2019-07-27
UI: Advice user when minikube profile is deleted


medyagh:
When minikube profile is delete and try to do minkube ssh , it crashes with this error:

it should have a nice Advice message, that says your minikube vm is not running, try minikube start.

```

$ ./out/minikube ssh -- ls

💣  Error getting host: Machine does not exist for api.Exists(lis)

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
👉  https://github.com/kubernetes/minikube/issues/new/choose

```

PR #4896, opened 2019-07-28
 Advice user does command ssh on non existing profile 

RA489:
When minikube profile is delete and try to do minkube ssh , it crashes with error
fixes https://github.com/kubernetes/minikube/issues/4889

### DIFF
--- a/pkg/minikube/problem/err_map.go
+++ b/pkg/minikube/problem/err_map.go
@@ -30,6 +30,11 @@ var vmProblems = map[string]match{
 		Advice: "Please make sure the service you are looking for is deployed or is in the correct namespace.",
 		Issues: []int{4599},
 	},
+	"MACHINE_NOT_FOUND": {
+		Regexp: re(`Machine does not exist for api.Exists`),
+		Advice: "Your minikube vm is not running, try minikube start.",
+		Issues: []int{4889},
+	},
 	"HYPERKIT_NO_IP": {
 		Regexp: re(`IP address never found in dhcp leases file Temporary Error: Could not find an IP address for`),
 		Advice: "Install the latest minikube hyperkit driver, and run 'minikube delete'",


pkg/minikube/problem/err_map.go
```diff
@@ -30,6 +30,11 @@ var vmProblems = map[string]match{
 		Advice: "Please make sure the service you are looking for is deployed or is in the correct namespace.",
 		Issues: []int{4599},
 	},
+	"PROFILE_NOT_FOUND": {
+		Regexp: re(`Could not find the minikube profile`),
```
medyagh: the regex should be,  "Machine does not exist for api.Exists(lis)"

pkg/minikube/problem/err_map.go
```diff
@@ -30,6 +30,11 @@ var vmProblems = map[string]match{
 		Advice: "Please make sure the service you are looking for is deployed or is in the correct namespace.",
 		Issues: []int{4599},
 	},
+	"PROFILE_NOT_FOUND": {
```
medyagh: this should be "MACHINE_NOT_FOUND"

pkg/minikube/problem/err_map.go
RA489: Done!

pkg/minikube/problem/err_map.go
RA489: Done!

RA489:
@medyagh Thanks for your reviews! I have updated the PR as per your suggestions.

pkg/minikube/problem/err_map.go
```diff
@@ -30,6 +30,11 @@ var vmProblems = map[string]match{
 		Advice: "Please make sure the service you are looking for is deployed or is in the correct namespace.",
 		Issues: []int{4599},
 	},
+	"MACHINE_NOT_FOUND": {
+		Regexp: re(`Machine does not exist for api.Exists(lis)`),
```
medyagh: the (lis) is extra ( that was my example profile)
it should be "Machine does not exist for api.Exists"

I recommend attaching the output of the console to try, I recmmend
deleting your minikube then 
then try ./out/minikube ssh -p p1 -- ls -lah

then paste the output in the PR review so we know it worked :)

RA489: sure

RA489: Done!

RA489:
@medyagh attaching the output of the console ```
root@openstack-nti-18:/home/ubuntu/k8sworkspaces/minikube/out# minikube ssh -p p1 -- ls -lah

💣  Error getting host: Machine does not exist for api.Exists(p1)

```

**Comment by tstromberg** [comment]
I don't think this is taking the optimal approach: the error is still opaque.

Rather than advice, I would prefer that ssh to exit with a very clear message, such as:

`There is no active "minikube" cluster to ssh to.`

I took a poke around, and think this could be solved by updating ssh.go to use `api.Exists`, such as:

```
	exists, err := api.Exists(machineName)
	if err != nil {
		return exit.WIthError(...)
	}
	if !exists {
               return exit.WithCode(exit.Data, "crafty error message")
       }
```

Alternatively, I saw that `dashboard.go` does some fancy error type checking, but `api.Exists` seems the clearest.

medyagh:
@RA489 Thank you for this contribution ! this will make user experience better

Issue #4606, opened 2019-06-26
wip: hyperv: Powershell was not found in the path


purushothamb11:
I am installing kubernetes in windows 10 with HyperV as VM-Drive, but when running minikube start it shows error unable to start vm, can any one help me in running minikube.
 
![kubeerror](https://user-images.githubusercontent.com/26902669/60176280-82fec700-9833-11e9-9e32-97e166f6561e.JPG)

sharifelgamal:
Make sure Powershell is in your PATH. It's required to start minikube with HyperV.

tstromberg:
I think the error is pretty clear here, but we could have a better solution message. Marking it as such.

blueelvis:
I wonder how is it possible that Powershell is not in PATH on the system. I wonder if there is something wrong with the system because Powershell by default is always in Path...

PR #4898, opened 2019-07-28
hyperv: Powershell was not found in the path 

RA489:
hyperv: Powershell was not found in the path 
fixes https://github.com/kubernetes/minikube/issues/4606

### DIFF
--- a/pkg/minikube/problem/err_map.go
+++ b/pkg/minikube/problem/err_map.go
@@ -144,6 +144,12 @@ var vmProblems = map[string]match{
 		Advice: "Configure an external network switch following the official documentation, then add `--hyperv-virtual-switch=<switch-name>` to `minikube start`",
 		URL:    "https://docs.docker.com/machine/drivers/hyper-v/",
 	},
+	"HYPERV_POWERSHELL_NOT_FOUND": {
+		Regexp: re(`Powershell was not found in the path`),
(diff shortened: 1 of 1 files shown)


pkg/minikube/problem/err_map.go
```diff
@@ -144,6 +144,12 @@ var vmProblems = map[string]match{
 		Advice: "Configure an external network switch following the official documentation, then add `--hyperv-virtual-switch=<switch-name>` to `minikube start`",
 		URL:    "https://docs.docker.com/machine/drivers/hyper-v/",
 	},
+	"HYPERV_POWERSHELL_NOT_FOUND": {
+		Regexp: re(`Powershell is not found in the path`),
```
medyagh: Thanks for making this PR, please make sure the regex, matches the exact error message,
in this case it is "Powershell was not found in the path"

RA489: s/is/was

RA489:
@medyagh Thanks for your reviews! I have updated as per your comments.

blueelvis:
@medyagh @RA489 - This same set of functionality is available with Docker machine's Hyper-V implementation but these are not public :(

https://github.com/machine-drivers/machine/blob/master/drivers/hyperv/powershell.go

Issue #4895, opened 2019-07-27
CI: install specific version of golang for cross build 


medyagh:
currently the cross build script on jenkins, uses whatever golang available on the master,

to minimize variances,
I recommend we pin down the version to use same golang verison both in travis and jenkins

the cross build script in jenkins should check for go version, if not available install it.
https://github.com/kubernetes/minikube/blob/master/hack/jenkins/minikube_cross_build_and_upload.sh

Currently I set the travis version to 1.12.7
we need to set it in jenkins too

afbjorklund:
So this means we should build more of minikube in docker ? Or did you have another plan here ?

Seems to be using quite different methods now, random Travis-image vs random Jenkins-install :

```yaml
matrix:
  include:
    - go: 1.12.7
```
```shell
declare -rx GOPATH=/var/lib/jenkins/go
```

> currently the cross build script on jenkins, uses whatever golang available on the master,

It seems that most _builds_, are using the version that is in the "kube-cross:v1.12.5" image.

```make
# NOTE: "latest" as of 2019-05-09. kube-cross images aren't updated as often as Kubernetes
BUILD_IMAGE 	?= k8s.gcr.io/kube-cross:v1.12.5-1
```

So far I can see a `gofmt` call, that is not respecting the `BUILD_IN_DOCKER` variable.

afbjorklund:
Maybe it is enough to update the BUILD_IMAGE ?

https://github.com/kubernetes/kubernetes/tree/master/build/build-image/cross

https://github.com/kubernetes/kubernetes/commit/dfa46f8d479920b56807b18eacb4d1bab6817e96

```Dockerfile
FROM golang:1.12.7
```

afbjorklund:
Building the kvm2 driver uses the go version from the host. Maybe that was a bad idea ?

Need to set GO_VERSION, when building in CI at least.

```Makefile
GO_VERSION ?= $(shell go version | cut -d' ' -f3 | sed -e 's/go//')
```

Anyway, that is where the 1.12.5 in the logs is coming from.

`docker build --build-arg "GO_VERSION=1.12.5" -t gcr.io/k8s-minikube/kvm-build-image -f installers/linux/kvm/Dockerfile installers/linux/kvm/`

afbjorklund:
I believe that everything should use the docker version of go now, is there anything else ?

```console
$ docker run k8s.gcr.io/kube-cross:v1.12.7-1 go version
go version go1.12.7 linux/amd64
```

[...]

PR #4907, opened 2019-07-29
Upgrade the build image from 1.12.5 to 1.12.7

afbjorklund:
For #4895

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -373,7 +373,7 @@ out/minikube-installer.exe: out/minikube-windows-amd64.exe
 	mv out/windows_tmp/minikube-installer.exe out/minikube-installer.exe
 	rm -rf out/windows_tmp
 
-out/docker-machine-driver-hyperkit:
+out/docker-machine-driver-hyperkit: pkg/minikube/translate/translations.go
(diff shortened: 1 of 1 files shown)


afbjorklund:
Seems that we are leaking translation support into the drivers now... ☹️

`# k8s.io/minikube/pkg/minikube/translate`
`pkg/minikube/translate/translate.go:77:12: undefined: Asset`

warmchang:
Now should update to go 1.12.8?

xref: 
https://github.com/kubernetes/kubernetes/blob/master/build/build-image/cross/Dockerfile#L18
https://github.com/docker-library/golang/blob/master/1.12/stretch/Dockerfile#L12

afbjorklund:
> Now should update to go 1.12.8?

Should be done in #5087

@medyagh: note that this also updates the underlying distro from "stretch" to "buster"

k8s.gcr.io/kube-cross:v1.12.8-1

```Dockerfile
FROM golang:1.12.8
```

https://hub.docker.com/_/golang
1.12.8, 1.12, 1, latest:
1.12.8-buster

https://github.com/docker-library/golang/blob/master/1.12/buster/Dockerfile

PR #4917, opened 2019-07-30
Pin the go version used for kvm-build-image

afbjorklund:
For #4895

Before it would build the docker image with the same go version as the build host (normally 1.12.5)


afbjorklund:
Note that our kvm-build-image is rather similar to library/golang

Except that we are using Ubuntu 16.04 rather than Debian stretch

And some differences in the package selection: g++ out, libvirt-dev in

Made them similar on purpose, we also needed curl and git packages.

PR #4918, opened 2019-07-30
Travis build status link should go to master

afbjorklund:
Currently it goes to the latest Pull Request instead.
The status widget (image) shows `master` though.

But at least it shows "Default Branch" at the top:

https://travis-ci.org/kubernetes/minikube/branches

----

We have a _lot_ of stale old branches there...
Maybe some of them could be cleaned up ?

Currently have **16 branches**, that's not very CI:
https://github.com/kubernetes/minikube/branches/all

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![BuildStatus Widget]][BuildStatus Result]
 [![GoReport Widget]][GoReport Status]
 
-[BuildStatus Result]: https://travis-ci.org/kubernetes/minikube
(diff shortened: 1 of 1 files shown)


afbjorklund:
Wonder if I spoke too soon. I had a "green" light on the website, and a "yellow" light in travis...
But it might not have been the actual PR, but a _merged_ PR that was not yet built on master.

Will keep an eye on it, and perhaps revert this change if it turns out that "Current" is good enough ?
It would also be excellent if we could get a Jenkins (or Prow?) result summary - from somewhere.

travis

PR #4921, opened 2019-07-30
Use consistent method of getting kvm-build-image

afbjorklund:
First try the local images, then try to pull from the registry...
If not found in either place, then build a tag a new local image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ HYPERKIT_BUILD_IMAGE 	?= karalabe/xgo-1.12.x
 # NOTE: "latest" as of 2019-07-12. kube-cross images aren't updated as often as Kubernetes
 BUILD_IMAGE 	?= k8s.gcr.io/kube-cross:v1.12.7-1
 ISO_BUILD_IMAGE ?= $(REGISTRY)/buildroot-image
-KVM_BUILD_IMAGE ?= $(REGISTRY)/kvm-build-image
+KVM_BUILD_IMAGE ?= $(REGISTRY)/kvm-build-image:$(GO_VERSION)
 
 ISO_BUCKET ?= minikube/iso
 
@@ -441,8 +441,7 @@ release-minikube: out/minikube checksum
 
 out/docker-machine-driver-kvm2: pkg/minikube/translate/translations.go
 ifeq ($(MINIKUBE_BUILD_IN_DOCKER),y)
-	docker inspect -f '{{.Id}} {{.RepoTags}}' $(KVM_BUILD_IMAGE) || \
-	docker pull $(KVM_BUILD_IMAGE) || $(MAKE) $(KVM_BUILD_IMAGE)
+	docker inspect -f '{{.Id}} {{.RepoTags}}' $(KVM_BUILD_IMAGE) || $(MAKE) kvm-image
 	$(call DOCKER,$(KVM_BUILD_IMAGE),/usr/bin/make $@ COMMIT=$(COMMIT))
 	# make extra sure that we are linking with the older version of libvirt (1.3.1)
 	test "`strings $@ | grep '^LIBVIRT_[0-9]' | sort | tail -n 1`" = "LIBVIRT_1.2.9"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -441,6 +441,7 @@ release-minikube: out/minikube checksum
 
 out/docker-machine-driver-kvm2: pkg/minikube/translate/translations.go
 ifeq ($(MINIKUBE_BUILD_IN_DOCKER),y)
+	docker inspect -f '{{.Id}} {{.RepoTags}}' $(KVM_BUILD_IMAGE) || \
```
josedonizetti: The`docker inspect` won't be necessary after we have this image on the registry, correct? Because `docker pull` will only pull if the image doesn't exist. Maybe we can add a `TODO: remove this when ....`

Makefile
```diff
@@ -479,7 +480,8 @@ $(KVM_BUILD_IMAGE): installers/linux/kvm/Dockerfile
 	@echo "$(@) successfully built"
 
 kvm_in_docker:
-	docker inspect -f '{{.Id}} {{.RepoTags}}' $(KVM_BUILD_IMAGE) || $(MAKE) $(KVM_BUILD_IMAGE)
+	docker inspect -f '{{.Id}} {{.RepoTags}}' $(KVM_BUILD_IMAGE) || \
```
josedonizetti:
```
docker inspect -f '{{.Id}} {{.RepoTags}}' $(KVM_BUILD_IMAGE) ||  docker pull $(KVM_BUILD_IMAGE) || $(MAKE) $(KVM_BUILD_IMAGE)
```

As we are doing it in two places, maybe worth extracting a function?

afbjorklund: Normally live by the "one, two, many (1,2,N)" rule

Makefile
afbjorklund: It will pull it every single time (maybe because we're using "latest") but anyway.

```
Using default tag: latest
Error response from daemon: manifest for gcr.io/k8s-minikube/kvm-build-image:latest not found
```

afbjorklund:
> @afbjorklund do you think worth tagging this image with the current minikube version?

No, definitely not the minikube version. We could tag it with the go version, though ?

Makefile
josedonizetti: hmm, for example, if I try to `docker pull nginx` which is on my machine already I get:
```
Using default tag: latest
latest: Pulling from library/nginx
Digest: sha256:eb3320e2f9ca409b7c0aa71aea3cf7ce7d018f03a372564dbdb023646958770b
Status: Image is up to date for nginx:latest
docker.io/library/nginx:latest
```

afbjorklund: It might be possible to use `docker images` instead of `docker inspect`, not sure ?

No `docker image exists`, unfortunately. (like there is with `podman image exists`)

josedonizetti: if you think worth double-checking before doing the pull, I'm cool with it, just that locally at least the second pull is really fast too, it probably just check the digest instead of doing a big download, but definitely not faster than checking with `docker inspect`

josedonizetti:
@afbjorklund do you think worth having a version at all? it seems a bit odd to me to have only the latest tag. But maybe it isn't

Makefile
afbjorklund: The problem is if you build an image locally that _doesn't exist_ on the docker hub.
Then it will fail to pull, and thus fail to recognize that it exists in the local storage..

afbjorklund: You are right of course in that it won't actually have to pull any layers or anything.

afbjorklund:
Basically we only had latest for the ISO_BUILD_IMAGE, which I copied...

But we should probably add the go version since it is a parameter after all.
That would also make it more similar to the kube-cross image versioning.

Now that we have pinned the go version down, could perhaps just use that ?

afbjorklund:
Apparently we are not using the registry _at all_. Might skip the "pull" step altogether.

josedonizetti:
+1 for using the go version and skipping pull

afbjorklund:
For some reason it takes docker 15-30 seconds to determine that the image is missing...

`time docker inspect -f '{{.Id}} {{.RepoTags}}' gcr.io/k8s-minikube/kvm-build-image:1.12.7`

afbjorklund:
The final straw is that make hates having a target with a colon (`:`) in it:

`Makefile:477: *** målmönstret innehåller inget ”%”.  Stannar.`

josedonizetti:
@afbjorklund doesn't take long here.

```
time docker inspect -f '{{.Id}} {{.RepoTags}}' gcr.io/k8s-minikube/kvm-build-image:1.12.7

Error: No such object: gcr.io/k8s-minikube/kvm-build-image:1.12.7

real	0m0.107s
user	0m0.041s
sys	0m0.028s
```

Issue #4891, opened 2019-07-27
Delete profile should Unset the current profile


medyagh:
Delete profile should Unset the current profile

to reproduce:
1- create a minikube with profile p1 and p2
```
./out/minikube start -p p1 --vm-driver=hyperkit
```
```
./out/minikube start -p p2 --vm-driver=hyperkit
```

2- set the current profile  to p1

```
./out/minikube profile p1
```

3- delete profile p1
```
 ./out/minikube delete -p p1
🔥  Deleting "p1" in hyperkit ...
💔  The "p1" cluster has been deleted.
```

4-
the current profile is gone from profile list (correctly) however if u check for current profile it shows the deleted one:
```
medya@ ~/workspace/ppl-minikube (fix-glog-parse-error) $ ./out/minikube profile list
|----------|-----------|----------------|-----------|--------------------|
| Profile  | VM Driver |     NodeIP     | Node Port | Kubernetes Version |
|----------|-----------|----------------|-----------|--------------------|
| minikube | hyperkit  | 192.168.64.102 |      8443 | v1.15.0            |
| p3       | hyperkit  | 192.168.64.118 |      8443 | v1.15.0            |
|----------|-----------|----------------|-----------|--------------------|

```

5- check for current profile
```
$ ./out/minikube profile
p1
```

The fix should, after delete unset the current config.

PR #4922, opened 2019-07-30
Unset profile when it is deleted

josedonizetti:
Fixes https://github.com/kubernetes/minikube/issues/4891


josedonizetti:
@medyagh Didn't understand. A unit test for the `.minikube delete` or for the `Unset`?

medyagh:
> @medyagh Didn't understand. A unit test for the `minikube delete` or for the `Unset` function?

ideally for both, but the Unset function is first ! priority is, all public functions must have unit tests

josedonizetti:
@medyagh `Unset` already have a test. https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/config/unset_test.go

But it can be improved by reading the property after than unset. I'll look into it.

PR #4927, opened 2019-07-30
Update KERNEL_VERSION to match buildroot config

afbjorklund:
We downgraded the kernel (to 4.15) in 2be51dc,
but never updated the Makefile from 4c4cd31...

The default kernel in Buildroot 2018.05.x was
4.16.13, but upgraded in Minikube to 4.16.14

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ MINIKUBE_BUCKET ?= minikube/releases
 MINIKUBE_UPLOAD_LOCATION := gs://${MINIKUBE_BUCKET}
 MINIKUBE_RELEASES_URL=https://github.com/kubernetes/minikube/releases/download
 
-KERNEL_VERSION ?= 4.16.14
+KERNEL_VERSION ?= 4.15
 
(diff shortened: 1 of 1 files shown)


medyagh:
Admitiably I don't know this part of this code but the PR description and the code makes sense to me. I think if you feel confident with it, you are have my +1 too.

afbjorklund:
Basically what it does is that it fixes the `linux-menuconfig` target:

```
$ LANG=C make linux-menuconfig
make -C ./out/buildroot/output/build/linux-4.16.14/ menuconfig
make[1]: *** out/buildroot/output/build/linux-4.16.14/: No such file or directory.  Stop.
Makefile:167: recipe for target 'linux-menuconfig' failed
make: *** [linux-menuconfig] Error 2
```

And it's also useful as documentation. Other than that, it is unused...

afbjorklund:
I don't really know why we still call it the "defconfig" after we have modified it, but...
Normally the "defconfig" would be the _default_ values, and "config" the modified one.

But I suppose it could be interpreted as the default configuration for minikube (or linux)
And the other one is call `alldefconfig` or whatever. Kconfig can be confusing sometimes.

Issue #4909, opened 2019-07-29
Building drivers crashes on missing translations


afbjorklund:
When trying to build either of the machine drivers, the build fails due to missing translations:

```Makefile
.PHONY: drivers
drivers: out/docker-machine-driver-hyperkit out/docker-machine-driver-kvm2
```

```
# k8s.io/minikube/pkg/minikube/translate
pkg/minikube/translate/translate.go:77:12: undefined: Asset
```

It requires this file to be generated (from JSON):
`pkg/minikube/translate/translations.go`

These files should not be needed nor used in the drivers, that are supposed to be lowlevel...

i.e. I don't think that we should have translation support, in the docker machine drivers code

afbjorklund:
@medyagh @sharifelgamal anyone knows what is going here ?
Is `go build` flipping out because of using modules, or what ?

``` console
$ go build -v k8s.io/minikube/cmd/drivers/kvm
k8s.io/minikube/pkg/minikube/translate
# k8s.io/minikube/pkg/minikube/translate
pkg/minikube/translate/translate.go:77:12: undefined: Asset
```

is there a good tool to show dependencies ?

afbjorklund:
Using `go list -json` is not so bad, it shows all the dependencies:

`GOOS=darwin go list -json cmd/drivers/hyperkit/main.go`

`GOOS=linux go list -json cmd/drivers/kvm/main.go`



It is the usage of either of these two imports:

* `k8s.io/minikube/pkg/minikube/config`

* `k8s.io/minikube/pkg/util`

afbjorklund:
I added the translations as a dependency, but don't really like it.

It is definitely going to block any upstreaming... 😢 #3939 #3169

josedonizetti:
@afbjorklund I've just tried to build the kvm driver on master and it worked without problem with both `make out/docker-machine-driver-kvm2` and `go build -v k8s.io/minikube/cmd/drivers/kvm`.
Is this error related to the link PR?

afbjorklund:
It is related so far as I started building the drivers clean, and noticed the issue when the files are gone.

josedonizetti:
Yeah, it fails. Probably related to https://github.com/kubernetes/minikube/blob/master/pkg/drivers/kvm/kvm.go#L28-L30

afbjorklund:
The "constants" are OK, they're just hiding some strings. Will be a problem when upstreaming, though.

josedonizetti:
[...]

PR #4933, opened 2019-07-31
Remove all minikube dependencies from drivers

sharifelgamal:
All the config not related to networking gets overridden in the config passed in from the client, so there's never a need to set any defaults.


pkg/drivers/hyperkit/driver.go
```diff
@@ -121,7 +117,7 @@ func (d *Driver) Create() error {
 
 // DriverName returns the name of the driver
 func (d *Driver) DriverName() string {
-	return constants.DriverHyperkit
+	return "hyperkit"
```
josedonizetti: can we make it a constant inside the driver?

pkg/drivers/kvm/kvm.go
```diff
@@ -224,7 +214,7 @@ func (d *Driver) GetSSHHostname() (string, error) {
 
 // DriverName returns the name of the driver
 func (d *Driver) DriverName() string {
-	return constants.DriverKvm2
+	return "kvm2"
```
josedonizetti: can we make it a constant inside the driver?

pkg/drivers/hyperkit/driver.go
```diff
@@ -268,13 +264,22 @@ func (d *Driver) Start() error {
...
+	}
+
+	// Implement a retry loop without calling any minikube code
+	for i := 0; i < 30; i++ {
+		err = getIP()
+		if err == nil {
+			break
 		}
-		return nil
+		if err.Error() != fmt.Sprintf(IPErrorMessage, mac) {
```
josedonizetti: hey, do you think it's worth having a custom error and testing the type? comparing error strings seems ugly :/

pkg/drivers/kvm/kvm.go
sharifelgamal: yep, done.

pkg/drivers/hyperkit/driver.go
sharifelgamal: yep, done.

pkg/drivers/hyperkit/driver.go
sharifelgamal: yeah I wasn't wild about text parsing either, and the custom error type is easy to add. done.

pkg/drivers/hyperkit/driver.go
```diff
@@ -52,6 +49,7 @@ const (
 	permErr         = "%s needs to run with elevated permissions. " +
 		"Please run the following command, then try again: " +
 		"sudo chown root:wheel %s && sudo chmod u+s %s"
+	driverHyperkit = "hyperkit"
```
afbjorklund: or just use the string?

pkg/drivers/kvm/kvm.go
```diff
@@ -91,6 +87,7 @@ const (
 	qemusystem                = "qemu:///system"
 	defaultPrivateNetworkName = "minikube-net"
 	defaultNetworkName        = "default"
+	driverKvm2 = "kvm2"
```
afbjorklund: or just use the string ?

pkg/drivers/kvm/kvm.go
```diff
@@ -224,7 +215,7 @@ func (d *Driver) GetSSHHostname() (string, error) {
 
 // DriverName returns the name of the driver
 func (d *Driver) DriverName() string {
-	return constants.DriverKvm2
+	return driverKvm2
```
afbjorklund: "kvm2"

pkg/drivers/hyperkit/driver.go
```diff
@@ -121,7 +118,7 @@ func (d *Driver) Create() error {
 
 // DriverName returns the name of the driver
 func (d *Driver) DriverName() string {
-	return constants.DriverHyperkit
+	return driverHyperkit
```
afbjorklund: "hyperkit"

pkg/drivers/hyperkit/network.go
```diff
@@ -78,7 +80,7 @@ func getIPAddressFromFile(mac, path string) (string, error) {
 			return dhcpEntry.IPAddress, nil
 		}
 	}
-	return "", fmt.Errorf("could not find an IP address for %s", mac)
+	return "", fmt.Errorf(IPErrorMessage, mac)
```
afbjorklund: would prefer having the printf inline here

pkg/drivers/hyperkit/driver.go
afbjorklund: I disagree :-P

pkg/drivers/kvm/kvm.go
afbjorklund: But why

afbjorklund:
> All the config not related to networking gets overridden in the config passed in from the client, so there's never a need to set any defaults.

I find this strange, since normally it does have some defaults. But I guess we will see, once it has been tried stand-alone (with `docker-machine`)

Then again, those defaults are usually 1 CPU and 1G RAM since that is what Docker requires.

It is Kubernetes that is increasing that to 2 CPU and 2G RAM, or maybe I should kubeadm does.

pkg/drivers/kvm/kvm.go
josedonizetti: @afbjorklund As those are the driver names I consider they good for a constant. To always have only one place to change it, even though now we don't have many. But as always, no strong feelings. and probably a too simple thing to start a big discussion, so all good to revert it to string only :)

afbjorklund: Well, at least it's not using the constants module. But the string is only used in 1 or 2 places ?

It's the not the most important thing, just thought it was an unnecessary step of indirection.

afbjorklund:
This addresses the dependencies found in #4909, but seems like it doesn't build just yet.

```
# k8s.io/minikube/pkg/drivers/hyperkit
pkg/drivers/hyperkit/driver.go:269:12: undefined: TempError
pkg/drivers/hyperkit/driver.go:280:18: impossible type assertion:
	*tempError does not implement error (missing Error method)
```

sharifelgamal:
@minikube-bot OK to test

sharifelgamal:
This should be ready now.

afbjorklund:
Why revert the libmachine version ?

Moving back to the 0.16.1 version re-introduces the AMD bug (yet again)

sharifelgamal:
It was causing issues with go modules, and specifically with the go proxy: 
`curl https://proxy.golang.org/github.com/docker/machine/@v/v0.16.1-0.20190718054102-a555e4f7a8f5.info
not found: github.com/docker/machine@v0.16.1-0.20190718054102-a555e4f7a8f5: invalid pseudo-version: revision a555e4f7a8f5 is not a descendent of preceding tag (v0.16.0)`

I didn't realize that reverting back to the release tag reintroduces a bug, I'll looking into pinning this correctly.

afbjorklund:
> It was causing issues with go modules, and specifically with the go proxy:
No worries, it was my fault for hacking the version. I'll go back to the weirdo `v0.0.0`

afbjorklund:
@sharifelgamal : why aren't we messing with go proxy (and Travis) in a separate PR ?

sharifelgamal:
You're right, it should be a separate PR. The travis issues were causing tests to fail in this PR so I was testing solutions. I'll open a new PR.

afbjorklund:
I can fix the go.mod, in order to make go proxy happy (have never tried go proxy)

sharifelgamal:
#5018 is the new go proxy PR

PR #4934, opened 2019-07-31
Docs-site: Front-page beautification

tstromberg:
The text is perhaps a bit bombastic, but here is my first stab at getting the feel of the front-page correct. Since the docs haven't moved over, the links are not terribly meaningful yet.


blueelvis:
@tstromberg - Can we add the link to our survey on the home page as well?

Also, what does "mother approved" mean? o.O

site/content/en/_index.html
```diff
@@ -7,83 +7,99 @@
...
 
-{{% blocks/lead color="primary" %}}
+<!-- Sales pitch -->
+{{< blocks/lead >}}
 
-minikube is a local Kubernetes environment focused on application developers.
+<div class="container">
+<h2>Instantly productive.</h2>
+
+A single command away from reproducing your production environment, from the comfort of localhost.
```
afbjorklund: I'm not 100% about "reproducing your production environment", since we don't have any multi-node support.

Would probably be more like "learning Kubernetes" or something, but I suppose for _apps_ it makes sense.

site/content/en/_index.html
```diff
@@ -7,83 +7,99 @@
...
 
 {{< imgproc "start" Fit "730x239" >}}{{< /imgproc >}}
 
-Kubernetes certified, mother approved.
-{{% /blocks/lead %}}
+<h3>Highlights</h3>
+<ul class="fa-ul">
+<li><i class="fas fa-check"></i> Always supports the latest Kubernetes release, as well as previous versions</li>
+<li><i class="fas fa-check"></i> Cross-platform (Linux, macOS, Windows)</li>
+<li><i class="fas fa-check"></i> Provides a Docker API endpoint</li>
```
afbjorklund: I'm not sure if this is a bug or a feature, but it sure seems popular.

afbjorklund:
> Also, what does "mother approved" mean? o.O

https://www.youtube.com/watch?v=amnZoXR5I04

tstromberg:
> @tstromberg - Can we add the link to our survey on the home page as well?
> 
> Also, what does "mother approved" mean? o.O

Just a silly American expression: https://www.youtube.com/watch?v=amnZoXR5I04

site/content/en/_index.html
tstromberg: > I'm not sure if this is a bug or a feature (the Docker API), but it sure seems popular.

A feature, but agreed. I suspect 2 years from know tools will have evolved to where it isn't an important feature.



> We need to talk about VirtualBox some time.

This PR doesn't mention VirtualBox, but agreed. That said, at the moment, it seems that VirtualBox is still the easiest environment to use on all platforms.

> The Docker API feature is a similar story, we have both "Reusable Docker daemon" and "Any container runtime". It would be awesome if we could focus on use cases, like building images or providing a CRI.

Agreed. If you have any suggestions on how to recategorize the features to improve this, I'm all ears.

> I would rather talk about different standards like the OCI, and how it supports your Dockerfile and your regular Kubernetes commands. Not so much about which vendor implementation that we are providing ?

In this section I was trying to highlight ways in which our tool was different than other  similar tools. I've gone ahead and mentioned the multiple runtimes here though.

tstromberg:
I think I've addressed all the feedback. Please take another look.

blueelvis:
@tstromberg - 
1. Sorry for not mentioning this earlier but I think that we should have some place on the website where it is easy to just click and get the latest release. Basically a `Downloads` section.

2. I just clicked on `Norsk` from the dropdown instead of English and there is no content?
image

3. The Twitter link at the bottom of the footer is bad.

tstromberg:
> @tstromberg -
> 
>     1. Sorry for not mentioning this earlier but I think that we should have some place on the website where it is easy to just click and get the latest release. Basically a `Downloads` section.

Added "Get Started" button: for most platforms, we give users a command to execute rather than a download to fetch.

>     2. I just clicked on `Norsk` from the dropdown instead of English and there is no content?
>        image
> 

Removed Norwegian translation.

>     3. The Twitter link at the bottom of the footer is bad.

Removed.

PR #4936, opened 2019-08-01
fix typo: clusters->cluster

xieyanker:


tstromberg:
Thanks for the PR! I don't feel that this is actually a typo, so I feel this PR should be closed.

minikube does allow for managing multiple local clusters (using --profile).

xieyanker:
> Thanks for the PR! I don't feel that this is actually a typo, so I feel this PR should be closed.
> 
> minikube does allow for managing multiple local clusters (using --profile).

@tstromberg Thanks for your reply, it's my mistake, I am sorry!

Issue #4920, opened 2019-07-30
Change the error message when Config file is not available


blueelvis:
**The exact command to reproduce the issue**:
minikube-windows-amd64.exe start --alsologtostderr --v=8

**The full output of the command that failed**:<details>
The full output is correct but this first line kind of pops out -
```
Error reading config file at C:\Users\Pranav.Jituri\.minikube\config\config.json: open C:\Users\Pranav.Jituri\.minikube\config\config.json: The system cannot find the file specified.
```

</details>

**The operating system version**:
Windows 10 Enterprise.

Instead of showing it as an error, this needs a better solution message. Earlier, I used to see this and get confused as to what was wrong. Later on figured that this is rather the external custom config which is used.

afbjorklund:
Should be shown as a **warning**, not as an error ?

```go
glog.Warningf("Error reading config file at %s: %v", configPath, err)
```

The verbose log does have some "scary" messages.

I supposed you _could_ filter out `os.IsNotExist(err)`.

But then it is hard to catch parameters that are wrong.

blueelvis:
My bad. Not an actual error but I mean that it should not state "Error" when clearly it is not an error but just that custom config is not present which needs to be applied.

afbjorklund:
Apparently it will be a `viper.ConfigFileNotFoundError`
So could perhaps use that instead of os.IsNotExist

And seems like glog doesn't show the severity level.
You filter that using the -v flag, where 8 is quite high

tstromberg:
Anyone know if this still happens in v1.4? I'm still confused as to why this output only appears in Windows. I would have expected to see it tucked away in the logs.

blueelvis:
Yep. Still happening on the latest build -
```
PS C:\utilities> .\minikube-windows-amd64.exe mount "E:\:/home/mount" --alsologtostderr --v=9                                                     W0922 02:15:42.239730   11644 root.go:240] Error reading config file at C:\Users\bluee\.minikube\config\config.json: open C:\Users\bluee\.minikube\config\config.json: The system cannot find the file specified.
* Mounting host path E:\ into VM as /home/mount ...
  - Mount type:   <no value>
  - User ID:      docker
  - Group ID:     docker
  - Version:      9p2000.L
  - Message Size: 262144
  - Permissions:  755 (-rwxr-xr-x)
  - Options:      map[]
* Userspace file server: ufs starting
[...]

PR #4940, opened 2019-08-01
WIP: Change the error message when Config file is not available

RA489:
Change the error message when Config file is not available
fixes https://github.com/kubernetes/minikube/issues/4920


cmd/minikube/cmd/root.go
```diff
@@ -145,9 +145,12 @@ func initConfig() {
...
 	viper.SetConfigFile(configPath)
 	viper.SetConfigType("json")
-	err := viper.ReadInConfig()
-	if err != nil {
-		glog.Warningf("Error reading config file at %s: %v", configPath, err)
+	if err := viper.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			glog.Warningf("Config file not found; ignore error if desired")
+		} else {
+			glog.Warningf("Config file was found but another error was produced")
```
josedonizetti: I know this is still WIP, but for the else case it's important to print the error that was returned. :)

afbjorklund: Just keep the original message...

cmd/minikube/cmd/root.go
```diff
@@ -145,9 +145,12 @@ func initConfig() {
 	configPath := constants.ConfigFile
 	viper.SetConfigFile(configPath)
 	viper.SetConfigType("json")
-	err := viper.ReadInConfig()
-	if err != nil {
-		glog.Warningf("Error reading config file at %s: %v", configPath, err)
+	if err := viper.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			glog.Warningf("Config file not found; ignore error if desired")
```
afbjorklund: This should just log a debug message, and continue. No warning, no "ignore error if desired" (it is)

cmd/minikube/cmd/root.go
RA489: sure

cmd/minikube/cmd/root.go
RA489: ok

cmd/minikube/cmd/root.go
RA489: PTAL

cmd/minikube/cmd/root.go
RA489: PTAL

cmd/minikube/cmd/root.go
```diff
@@ -145,9 +145,10 @@ func initConfig() {
 	configPath := constants.ConfigFile
 	viper.SetConfigFile(configPath)
 	viper.SetConfigType("json")
-	err := viper.ReadInConfig()
-	if err != nil {
-		glog.Warningf("Error reading config file at %s: %v", configPath, err)
+	if err := viper.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			glog.Infof("Error reading config file at %s: %v", configPath, err)
+		}
```
tstromberg: This should call `glog.Errorf` or at least `glog.Warningf` if the error is not ConfigFileNotFoundError

tstromberg:
Obsoleted by #5732

PR #4944, opened 2019-08-01
Refactor extract writeconfig

josedonizetti:
- `WriteConfig` should be on the same package as `ReadConfig`
- `WriteConfig` need proper tests
- remove duplication (check `configTestCase`) 
- cleaning up `cmd`

### DIFF
--- a/pkg/minikube/config/config_test.go
+++ b/pkg/minikube/config/config_test.go
@@ -18,6 +18,7 @@ package config
 
 import (
 	"bytes"
+	"os"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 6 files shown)


pkg/minikube/config/config_test.go
```diff
@@ -141,3 +142,43 @@ func Test_readConfig(t *testing.T) {
 		t.Errorf("Did not read config correctly,\n\n wanted %+v, \n\n got %+v", expectedConfig, mkConfig)
 	}
 }
+
+func Test_writeConfig(t *testing.T) {
```
tstromberg: I prefer that unit tests target the public method when possible, though I understand that in this case it isn't easy to do so without breaking the API.

This is definitely an improvement over what we currently have, but two alternative ideas are:

- Have the public WriteConfig function accept an argument of the file path
- Give the `config` package a var containing the default path to use that can be set by tests

Thoughts?

pkg/minikube/config/config_test.go
```diff
@@ -141,3 +142,43 @@ func Test_readConfig(t *testing.T) {
...
+	configFile := "./testdata/configTest"
+	cfg := map[string]interface{}{
+		"vm-driver":            constants.DriverKvm2,
+		"cpus":                 4,
+		"disk-size":            "20g",
+		"show-libmachine-logs": true,
+		"log_dir":              "/etc/hosts",
+	}
+
+	err := writeConfig(configFile, cfg)
```
tstromberg: Can this test instead write to an out-of-tree path by using `ioutil.TempFile`? This appears to update the git tree.

pkg/minikube/config/config_test.go
josedonizetti: If changing an IP is a possibility, let's do it. I don't like methods hiding it's internal like that.

medyagh:
Thank you for improving minikube's code base.

Issue #4932, opened 2019-07-31
`minikube status` should not exit non-zero


medyagh:
if minikube is not running and you do minikube status, it exists 1. I dont think it is an Error if you want your status to be not running !

```
 $ minikube status  --format={{.Host}}
 $ echo $?
1
```

this has caused weird error handling in integration tests

the same story goes if the status is stopped, the status command is returning 1:

$ ./out/minikube status --format={{.Host}}
Stopped
$ echo $?
1

tstromberg:
I feel like 'stopped' should still return an error code, but a well defined one that differentiates between a general failure, 'stopped', and a failed cluster.

Thoughts?

PR #4923, opened 2019-07-30
parallel integration tests

medyagh:
This rebase of two previously worked on PRs (100+) commits: #4665 and https://github.com/kubernetes/minikube/pull/4665

This PR replaces the other two PR that were already reviewed.

- Make parallel optional
- Consistent naming for minikube runner (mk) and kubectl runner (kr) everywhere in the tests
- Logging the ignored stderr when error happens !
- Adding an retriable RunCommand with exponential backoff 
- Remove unused funcs of minikube runner
- Add profile to minikube runner
- Add kube-context to kubectl runner
- Breakdown the giant util file for integration tests into three files (minikube_runner, kubectl_runner and common)
- Change vague helper names verify(t) --> verifyBusybox(t)
- Organize the integration test files, (add fn_ to all functional test related tests)
- Move download binary logic to common.sh for better test setup
- Run some tests in Parallel.
- Decouple isTestNoneDriver from MinikubeRunner for more reusability
- Move containerd tests to its own file
- Changing hard coded path to "testdata" to get from args
- set explicitly the max number of parallel tests to run using for test.parallel
-  added a new minikube runner helper, StartWithStd, that returns sdoout,stderr to simplfy many repetitive codes in the tests
- pass opts to EnsureRunning(opts ...string)
-  added a new configuration for commons.sh PARALLEL_COUNT to configure parrallel tests count per job type
-  add run groups to deal with golang's tear down behavior for parallel runs . golang/go#17791 (comment)
- handle error for minkube status because of https://github.com/kubernetes/minikube/issues/4932
(instead of ignoring it)
- remove duplicate gvisor integration test

PR #4948, opened 2019-08-01
parallel integration tests 4

medyagh:
This rebase of three previously worked on PRs (100+) commits: #4665 and #4665 and #4923

This PR replaces the other three PR that were already reviewed.
- Make parallel optional
- Consistent naming for minikube runner (mk) and kubectl runner (kr) everywhere in the tests
- Logging the ignored stderr when error happens !
- Adding an retriable RunCommand with exponential backoff 
- Remove unused funcs of minikube runner
- Add profile to minikube runner
- Add kube-context to kubectl runner
- Breakdown the giant util file for integration tests into three files (minikube_runner, kubectl_runner and common)
- Change vague helper names verify(t) --> verifyBusybox(t)
- Organize the integration test files, (add fn_ to all functional test related tests)
- Move download binary logic to common.sh for better test setup
- Run some tests in Parallel.
- Decouple isTestNoneDriver from MinikubeRunner for more reusability
- Move containerd tests to its own file
- Changing hard coded path to "testdata" to get from args
- set explicitly the max number of parallel tests to run using for test.parallel
-  added a new minikube runner helper, StartWithStd, that returns sdoout,stderr to simplfy many repetitive codes in the tests
- pass opts to EnsureRunning(opts ...string)
-  added a new configuration for commons.sh PARALLEL_COUNT to configure parrallel tests count per job type
-  add run groups to deal with golang's tear down behavior for parallel runs . golang/go#17791 (comment)
- handle error for minkube status because of https://github.com/kubernetes/minikube/issues/4932
(instead of ignoring it)
- remove duplicate gvisor integration test


.gitignore
```diff
@@ -42,3 +42,5 @@ deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/etc/VERSION
 /.idea
 
 /.vscode
+
+test/integration/testdata/minikube-linux-amd64-latest-stable
```
tstromberg: This doesn't seem right. test/ is a code directory and should be immutable: testdata/ doubly so.

test/integration/a_download_only_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"time"
+
+	"github.com/hashicorp/go-getter"
+	"github.com/pkg/errors"
+	"k8s.io/minikube/pkg/minikube/constants"
+	pkgutil "k8s.io/minikube/pkg/util"
+	"k8s.io/minikube/test/integration/util"
+)
+
+// Note this test runs before all because filename is alpahbetically first
```
tstromberg: spelling of alphabetically.

test/integration/a_download_only_test.go
```diff
@@ -0,0 +1,89 @@
...
+	t.Run("DownloadLatestRelease", func(t *testing.T) {
+		dest := filepath.Join(*testdataDir, fmt.Sprintf("minikube-%s-%s-latest-stable", runtime.GOOS, runtime.GOARCH))
+		err := downloadMinikubeBinary(dest, "latest")
+		if err != nil {
+			t.Errorf("erorr downloading the latest minikube release %v", err)
+		}
+	})
+}
+
+// downloadMinikubeBinary downloads the minikube binary from github used by TestVersionUpgrade
```
tstromberg: Please ensure that TestVersionUpgrade can still function independently of TestDownloadOnly.

test/integration/flags.go
```diff
@@ -31,19 +32,55 @@ func TestMain(m *testing.M) {
...
+	if strings.Contains(p, "/") { // for i.e, TestFunctional/SSH returns TestFunctional
+		p = strings.Split(p, "/")[0]
+	}
+	if p == "TestFunctional" {
+		return "minikube"
+	}
+	return p
+}
+
+// toParallel deterimines if test should run in  parallel or not
```
tstromberg: I couldn't tell what this function was by it's name. Could you find an alternative, such as `shouldRunInParallel()`?

test/integration/containerd_test.go
```diff
@@ -0,0 +1,133 @@
...
+	stdout, stderr, err := mk.Start("--container-runtime=containerd", "--docker-opt containerd=/var/run/containerd/containerd.sock")
+	if err != nil {
+		t.Fatalf("failed to start minikube (for profile %s) failed : %v\nstdout: %s\nstderr: %s", p, err, stdout, stderr)
+	}
+
+	mk.RunCommand("addons enable gvisor", true)
+
+	t.Log("waiting for gvisor controller to come up")
+	if err := waitForGvisorControllerRunning(p); err != nil {
+		t.Fatalf("waiting for gvisor controller to be up: %v", err)
```
tstromberg: Not your fault, but these t.Fatalf calls should be t.Errorf.

Even after a test case encounters a failure, they should keep going for as long as possible in order to print out all of the failed checks in a single run. This way, someone who's fixing the failing test doesn't have to play whac-a-mole, fixing one bug and then re-running the test to find the next bug.

test/integration/flags.go
```diff
@@ -31,19 +32,55 @@ func TestMain(m *testing.M) {
...
+
+// isTestNoneDriver checks if the current test is for none driver
+func isTestNoneDriver() bool {
+	return strings.Contains(*startArgs, "--vm-driver=none")
+}
+
+// profileName chooses a profile name based on the test name
+// to be used in minikube and kubecontext across that test
+func profileName(t *testing.T) string {
+	if isTestNoneDriver() {
```
tstromberg: This function seems simple enough, but all test helpers (functions which accept *testing.T) should call immediately `t.Helper()`, so that Go can attribute failures in the test helper to the line where the helper was called.

test/integration/flags.go
```diff
@@ -31,19 +32,55 @@ func TestMain(m *testing.M) {
...
+		return "minikube"
+	}
+	p := t.Name()
+	if strings.Contains(p, "/") { // for i.e, TestFunctional/SSH returns TestFunctional
```
tstromberg: This if statement isn't necessary. strings.Split on the first index will still succeed whether or not the string contains a /.

test/integration/flags.go
```diff
@@ -31,19 +32,55 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
+var startTimeout = flag.Int("timeout", 25, "number of minutes to wait for minikube start")
 var binaryPath = flag.String("binary", "../../out/minikube", "path to minikube binary")
 var globalArgs = flag.String("minikube-args", "", "Arguments to pass to minikube")
 var startArgs = flag.String("minikube-start-args", "", "Arguments to pass to minikube start")
 var mountArgs = flag.String("minikube-mount-args", "", "Arguments to pass to minikube mount")
 var testdataDir = flag.String("testdata-dir", "testdata", "the directory relative to test/integration where the testdata lives")
+var disableParallel = flag.Bool("disable-parallel", false, "run the tests squentially and disable all parallel runs")
```
tstromberg: Flags should not be named negative, as it can cause confusing combinations, for instance, if someone sets `--disable-parallel=false`. How about just `parallel`, but set it to true by default?

test/integration/util/common.go
```diff
@@ -0,0 +1,82 @@
...
+		err = callback()
+		if err == nil {
+			return nil
+		}
+		time.Sleep(d)
+	}
+	return err
+}
+
+// RetryX is expontential backoff retry
```
tstromberg: This looks good. What do you think about replacing Retry() with this implementation and dropping Retry2()?

test/integration/z_proxy_test.go
```diff
@@ -16,6 +16,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// the name of this file starts with z intentionally to make it run last after all other tests
```
tstromberg: The comment should include "why".

.gitignore
medyagh: this is people who run their integration test on desktop and dont wanna accidentally checkin the file.

I thought about downloading the file into a random temp file, but since I am downloading the file in the TestDownloadOnly, and use it in another tests I rather I have a specific location for it.

test/integration/a_download_only_test.go
medyagh: The reason I made this compromise was, because I had issues when running tests in parallel, for some reason downloading file (because it happens async) was making it imnpossible to download the file in a not-parallell test, so I had to either choose to make trestversionupgrade a not parallel test or do this hack !

the other option was to download this file in our commons.sh in jenkins

medyagh: thats is a good idea ! I will add download if dioesnt exist to test version upgrade !

test/integration/util/common.go
medyagh: I will replace all the retries in minikube in a separate PR, for better PR review

tstromberg:
Please merge when you feel ready &  sort out the retries in a separate PR.

```
“Be empty of worrying.
Think of who created thought!

Why do you live with slow integration tests
When the PR is ready to merge?” 

-- Rumi
```

PR #4966, opened 2019-08-03
logs container status: fallback to docker if crictl fails (none)

tstromberg:
Users of the `none` driver may not have `crictl` installed.


afbjorklund:
Hmm, the presence of `crictl` _should_ have been verified by the `kubeadm` bootstrapper ?

But apparently* Kubernetes lifted that requirement, and still alllow Docker to "bypass" the CRI 😢

* 66282

> "Eventually, yes, it would be great to replace all container runtime calls with crictl."

PR #4969, opened 2019-08-03
Use configured DNS domain in bootstrapper templates.

Eelis:
Tested, by verifying that the generated `ClusterConfiguration` entry in the `kubeadm-config` ConfigMap now gets the custom DNS domain value if one was passed with `--dns-domain` to `minikube start`.


Eelis:
Never mind, way too much hassle dealing with the bot spam and invasive licensing registration crap. I prefer waiting for the maintainers to fix this.

PR #4984, opened 2019-08-05
Switch ISO download impl from go-download to go-getter

### LINKED ISSUE #4983: head: ISO download progress bar display issues: repeat bars
Reported by @medyagh - I can confirm that this affects both macOS and Linux

Repeatable using:

`rm -Rf ~/.minikube/cache/iso`

Appearance:

```
😄  minikube v1.2.0 on Debian rodete
💿  Downloading Minikube ISO ...
 129.33 MiB / 129.33 MiB [=============================================] 100.00%
🔥  Creating virtualbox VM (CPUs=2, Memory=2000MB, Disk=20000MB) ...
 129.33 MiB / 129.33 MiB [=============================================] 100.00%E0805 09:14:54.965460   43836 start.go:723] StartHost: create: precreate: VBoxManage not found. Make sure VirtualBox is installed and VBoxManage is in the path
```

This does not appear to happen in the official v1.2.0 release.

tstromberg:
In 04144ca we took on a new Go download library dependency: `hashicorp/go-getter`. This new dependency inadvertently updated the progressbar library (`cheggaaa/pb`) from v1.0.6 to v1.0.27, which caused display issues for us.

This PR migrates our code base away from `go-download`, which is both little used outside of minikube, and hasn't seen commits in years.

go-getter does require us to provide a progressbar wrapper. I have chosen to adapt the one used from their `main.go`:  https://github.com/hashicorp/go-getter/blob/master/cmd/go-getter/progress_tracking.go

The output differs slightly in that go-getter does not hide the checksum download:

```
😄  minikube v1.3.0 on Darwin 10.14.5
    minikube-v1.3.0.iso.sha256: 65 B / 65 B [=====================] 100.00% 0s
    minikube-v1.3.0.iso: 131.07 MiB / 131.07 MiB [================] 100.00% 3s
🔥  Creating virtualbox VM (CPUs=2, Memory=2000MB, Disk=20000MB) ...
```

This PR closes #4983

### DIFF
--- a/pkg/util/progressbar.go
+++ b/pkg/util/progressbar.go
@@ -21,6 +21,7 @@ limitations under the License.
 package util
 
 import (
+	"fmt"
 	"io"
 	"path/filepath"
 	"sync"
@@ -55,8 +56,8 @@ func (cpb *progressBar) TrackProgress(src string, currentSize, totalSize int64, 
 	p := pb.New64(totalSize)
 	p.Set64(currentSize)
 	p.SetUnits(pb.U_BYTES)
-	p.Prefix(filepath.Base(src))
-
+	p.Prefix(fmt.Sprintf("    %s:", filepath.Base(src)))
+	p.SetWidth(78)
 	cpb.pool.Add(p)
 	reader := p.NewProxyReader(stream)
 
--- a/pkg/util/progressbar.go
+++ b/pkg/util/progressbar.go
@@ -57,6 +57,7 @@ func (cpb *progressBar) TrackProgress(src string, currentSize, totalSize int64, 
 	p.Set64(currentSize)
 	p.SetUnits(pb.U_BYTES)
 	p.Prefix(fmt.Sprintf("    %s:", filepath.Base(src)))
+	// Just a hair less than 80 (standard terminal width) for aesthetics & pasting into docs
 	p.SetWidth(78)
 	cpb.pool.Add(p)
 	reader := p.NewProxyReader(stream)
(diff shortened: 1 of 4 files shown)


pkg/util/progressbar.go
```diff
@@ -0,0 +1,88 @@
...
+type progressBar struct {
+	lock sync.Mutex
+	pool *pb.Pool
+	pbs  int
+}
+
+// TrackProgress instantiates a new progress bar that will
+// display the progress of stream until closed.
+// total can be 0.
+func (cpb *progressBar) TrackProgress(src string, currentSize, totalSize int64, stream io.ReadCloser) io.ReadCloser {
```
medyagh: how do you feel about moving this to "out" package?

tstromberg: I'd like to give download related code a home in a `download` package rather than `util`, but I felt that refactor was too big for a late PR blocker.

`out` is another possibility, but this implementation is specifically tied to go-getter.ProgessTracker, so I'd rather leave it nearer to the caller for convenience.

medyagh: ok fair enough

pkg/util/progressbar.go
```diff
@@ -0,0 +1,89 @@
...
+		if err := cpb.pool.Start(); err != nil {
+			glog.Errorf("pool start: %v", err)
+		}
+	}
+
+	p := pb.New64(totalSize)
+	p.Set64(currentSize)
+	p.SetUnits(pb.U_BYTES)
+	p.Prefix(fmt.Sprintf("    %s:", filepath.Base(src)))
+	p.SetWidth(78)
```
medyagh: a comment here would be nice why 78 ?

PR #4995, opened 2019-08-06
Check for Username instead of Name for root

MonoAnji:
for "--vm-driver=none":
user.Name may be empty as per https://golang.org/pkg/os/user/#User which happens to be true on Manjaro 18.0.4 for root. Actually checking for the Username instead (display-) Name seems to be more reasonable.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -471,9 +471,9 @@ func validateUser() {
 	d := viper.GetString(vmDriver)
 	// Check if minikube needs to run with sudo or not.
 	if err == nil {
-		if d == constants.DriverNone && u.Name != "root" {
+		if d == constants.DriverNone && u.Username != "root" {
 			exit.UsageT(`Please run with sudo. the vm-driver "{{.driver_name}}" requires sudo.`, out.V{"driver_name": constants.DriverNone})
-		} else if u.Name == "root" && !(d == constants.DriverHyperv || d == constants.DriverNone) {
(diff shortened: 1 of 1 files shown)


medyagh:
@MonoAnji Thank you for this contribution ! great catch, this will improve minikube's checks !

cmd/minikube/cmd/start.go
```diff
@@ -471,9 +471,9 @@ func validateUser() {
 	d := viper.GetString(vmDriver)
 	// Check if minikube needs to run with sudo or not.
 	if err == nil {
-		if d == constants.DriverNone && u.Name != "root" {
+		if d == constants.DriverNone && u.Username != "root" {
 			exit.UsageT(`Please run with sudo. the vm-driver "{{.driver_name}}" requires sudo.`, out.V{"driver_name": constants.DriverNone})
-		} else if u.Name == "root" && !(d == constants.DriverHyperv || d == constants.DriverNone) {
+		} else if u.Username == "root" && !(d == constants.DriverHyperv || d == constants.DriverNone) {
```
tstromberg: Thank you! This should really be checking for `u.Uid == 0`, as systems can have multiple root-access accounts with a uid of 0.

The root user can also be renamed.

blueelvis:
@tstromberg @medyagh - Just FYI, this is being implemented as part of this - https://github.com/kubernetes/minikube/pull/4745

RA489:
@minikube-bot OK to test

medyagh:
> @tstromberg @medyagh - Just FYI, this is being implemented as part of this - #4745

I am closing this in favor of  https://github.com/kubernetes/minikube/pull/4745

PR #5001, opened 2019-08-06
Consistently name test targets <driver>_<os>

tstromberg:
The names were inconsistent between Jenkins and Cloud Storage URL's. This makes them consistent.

It also removes crufty old code.


hack/jenkins/linux_integration_tests_kvm.sh
```diff
@@ -27,7 +27,7 @@ set -e
 
 OS_ARCH="linux-amd64"
 VM_DRIVER="kvm2"
-JOB_NAME="Linux-KVM"
+JOB_NAME="Linux_KVM"
```
sharifelgamal: This should be KVM_Linux

tstromberg: Done

Issue #4854, opened 2019-07-23
start_stop_delete_test: `config current-context` run against minikube instead of kubectl


medyagh:
the following code is not right ! it is using a minikube runner to run a kubectl config command !

https://github.com/kubernetes/minikube/blob/9dfe660c64ed9ac5d76650868bde29272211c582/test/integration/start_stop_delete_test.go#L105

This needs to be fixed

```

			// running this command results in error when the current-context is not set
			if err := r.Run("config current-context"); err != nil {
				t.Logf("current-context is not set to minikube")
			}

````

tstromberg:
Was this fixed?

medyagh:
> Was this fixed?

no I will add it to the kubeconfig PR

medyagh:
Done

PR #5003, opened 2019-08-07
Refactor Kubeconfig 

medyagh:
- Move all kubeconfig funcs to its own packge
- Make them more testable by parameterizing config path 
- Get rid of cmd/util package
- Adds Integration Test for KubeConfig Current Context closes this https://github.com/kubernetes/minikube/issues/4854


pkg/minikube/kubeconfig/update.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/pkg/errors"
+	"github.com/spf13/viper"
+	cfg "k8s.io/minikube/pkg/minikube/config"
+	"k8s.io/minikube/pkg/minikube/constants"
+	"k8s.io/minikube/pkg/minikube/exit"
+)
+
+// Update sets up kubeconfig to be used by kubectl
+func Update(h *host.Host, c *cfg.Config) *Setup {
+	addr, err := h.Driver.GetURL()
```
tstromberg: Instead of accepting host.Host, could this could accept a proper URL for ClusterServerAddress?

I feel like making this library depend on a large library like host makes the interface muddy and forces it to handle implementation details it shouldn't need to know. It would make testing nicer too.

pkg/minikube/kubeconfig/update.go
```diff
@@ -0,0 +1,114 @@
...
+// If no CurrentContext is set, the given name will be used.
+func update(cfg *Setup) error {
+	glog.Infoln("Using kubeconfig: ", cfg.fileContent())
+
+	// read existing config or create new if does not exist
+	config, err := readOrNew(cfg.fileContent())
+	if err != nil {
+		return err
+	}
+
```
tstromberg: Perhaps compare values first to see if the write is even necessary?

pkg/minikube/kubeconfig/update.go
```diff
@@ -0,0 +1,114 @@
...
+		addr = strings.Replace(addr, c.KubernetesConfig.NodeIP, c.KubernetesConfig.APIServerName, -1)
+	}
+
+	kcs := &Setup{
+		ClusterName:          cfg.GetMachineName(),
+		ClusterServerAddress: addr,
+		ClientCertificate:    constants.MakeMiniPath("client.crt"),
+		ClientKey:            constants.MakeMiniPath("client.key"),
+		CertificateAuthority: constants.MakeMiniPath("ca.crt"),
+		KeepContext:          viper.GetBool("keep-context"),
```
tstromberg: Rather than relying on global variables, can you make it possible to pass these values in?

pkg/minikube/kubeconfig/update.go
```diff
@@ -0,0 +1,114 @@
...
+		EmbedCerts:           viper.GetBool("embed-certs"),
+	}
+	kcs.setPath(Path())
+	if err := update(kcs); err != nil {
+		exit.WithError("Failed to setup kubeconfig", err)
```
tstromberg: Return an err - libraries shouldn't panic.

pkg/minikube/kubeconfig/setup.go
```diff
@@ -0,0 +1,111 @@
...
+package kubeconfig
+
+import (
+	"io/ioutil"
+	"sync/atomic"
+
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+// Setup is the kubeconfig setup
```
tstromberg: I feel like a struct with these sort of values would typically be called Config... but it's not, because it has functions on it. It doesn't quite feel right to me.

pkg/minikube/kubeconfig/setup.go
```diff
@@ -0,0 +1,111 @@
...
+	ClientKey string
+
+	// Should the current context be kept when setting up this one
+	KeepContext bool
+
+	// Should the certificate files be embedded instead of referenced by path
+	EmbedCerts bool
+
+	// kubeConfigFile is the path where the kube config is stored
+	// Only access this with atomic ops
```
tstromberg: Why?

pkg/minikube/kubeconfig/setup.go
```diff
@@ -0,0 +1,111 @@
...
+	kubeConfigFile atomic.Value
+}
+
+// SetKubeConfigFile sets the kubeconfig file
+func (k *Setup) setPath(kubeConfigFile string) {
```
tstromberg: This seems like overkill to me. What's going on here? :)

pkg/minikube/kubeconfig/update.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// write back to disk
+	if err := writeToFile(config, cfg.fileContent()); err != nil {
+		return errors.Wrap(err, "writing kubeconfig")
+	}
+	return nil
+}
+
+// UpdateIP overwrites the IP stored in kubeconfig with the provided IP.
+func UpdateIP(ip net.IP, filename string, machineName string) (bool, error) {
```
tstromberg: See if there is a more appropriate term to use. AFAIK, kubeconfig shouldn't know what a machine is. It only knows about clusters and contexts: https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/

pkg/minikube/kubeconfig/update.go
```diff
@@ -0,0 +1,114 @@
...
+		return false, err
+	}
+	if kip.Equal(ip) {
+		return false, nil
+	}
+	kport, err := Port(filename, machineName)
+	if err != nil {
+		return false, err
+	}
+	con, err := readOrNew(filename)
```
tstromberg: This seems to be re-reading the kubeconfig file 3 times. For consistency and performance, can you get it down to a single read?

pkg/minikube/kubeconfig/update.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// UpdateIP overwrites the IP stored in kubeconfig with the provided IP.
+func UpdateIP(ip net.IP, filename string, machineName string) (bool, error) {
+	if ip == nil {
+		return false, fmt.Errorf("error, empty ip passed")
+	}
+	kip, err := extractIP(filename, machineName)
+	if err != nil {
+		return false, err
```
tstromberg: There are many ways this function can return an error, so please use errors.Wrap. Future debuggers will thank you.

pkg/minikube/kubeconfig/context.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+// UnsetCurrentContext unsets the current-context from minikube to "" on minikube stop
+func UnsetCurrentContext(filename, machineName string) error {
+	confg, err := readOrNew(filename)
```
tstromberg: readOrNew feels like it does too much here (and elsewhere). For functions which expect a file to exist, I don't think it should go around creating a new file. Thoughts?

pkg/minikube/kubeconfig/context_test.go
```diff
@@ -0,0 +1,148 @@
...
+func TestSetCurrentContext(t *testing.T) {
+	contextName := "minikube"
+
+	kubeConfigFile, err := ioutil.TempFile("/tmp", "kubeconfig")
+	if err != nil {
+		t.Fatalf("Error not expected but got %v", err)
+	}
+	defer os.Remove(kubeConfigFile.Name())
+
+	cfg, err := readOrNew(kubeConfigFile.Name())
```
tstromberg: I don't like "or" functions, especially in tests. Reading this test, I can't trivially determine if you are expecting this test to read a file, or create a file.

pkg/minikube/kubeconfig/context_test.go
```diff
@@ -0,0 +1,148 @@
...
+		t.Fail()
+	}
+
+	if len(cfg.Contexts) != 0 {
+		t.Fail()
+	}
+}
+
+func TestSetCurrentContext(t *testing.T) {
+	contextName := "minikube"
```
tstromberg: Move this variable closer to where it's first used.

pkg/minikube/kubeconfig/context_test.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}()
+
+	cfg, err = readOrNew(kubeConfigFile.Name())
+	if err != nil {
+		t.Fatalf("Error not expected but got %v", err)
+	}
+
+	if cfg.CurrentContext != contextName {
+		t.Errorf("Expected context name %s but got %s", contextName, cfg.CurrentContext)
```
tstromberg: Test outputs should output the actual value that the function returned before printing the value that was expected. A usual format for printing test outputs in Go is `YourFunc(%v) = %v, want %v`

pkg/minikube/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/client-go/tools/clientcmd/api/latest"
+	"k8s.io/minikube/pkg/minikube/constants"
+	pkgutil "k8s.io/minikube/pkg/util"
+)
+
+// VeryifyMachineIP verifies the ip stored in kubeconfig.
```
tstromberg: Use cluster or context instead of machine here too.

tstromberg: Can this be reworded in a way that describes when to expect true vs false? I can't trivially tell.

pkg/minikube/kubeconfig/setup.go
```diff
@@ -0,0 +1,111 @@
...
+		cluster.CertificateAuthorityData, err = ioutil.ReadFile(cfg.CertificateAuthority)
+		if err != nil {
+			return err
+		}
+	} else {
+		cluster.CertificateAuthority = cfg.CertificateAuthority
+	}
+	apiCfg.Clusters[clusterName] = cluster
+
+	// user
```
tstromberg: this comment doesn't seem useful.

PR #5025, opened 2019-08-08
Upgrade cheggaaa/pb from v1.0.27 to v3.0.1

### LINKED ISSUE #4996: Cannot create a new VM on git bash or cygwin
Hello,

When using GIT bash or cygwin, it is no more possible to create a new VM the first time as it fails with the following error (and hangs):

```
$ minikube start
* minikube v1.3.0 on Microsoft Windows 7 Enterprise Service Pack 1 6.1.7601 Build 7601
E0806 14:52:32.005000    3936 progressbar.go:52] pool start: Can't get terminal settings: The handle is invalid.
```

Using winpty works but is not very convenient...

blueelvis:
Fixes #4996 

Tested on the following -
1. Windows CMD
image
2. Windows PowerShell
image
3. Git for Bash Terminal (With default size)
image

Although, as of now, I still don't have any idea why we are locking and unlocking with the mutex. Could someone please explain?

Please try on your local as well.

-Pranav

### DIFF
--- a/pkg/util/progressbar.go
+++ b/pkg/util/progressbar.go
@@ -45,9 +45,8 @@ func (cpb *progressBar) TrackProgress(src string, currentSize, totalSize int64, 
 	if cpb.progress == nil {
 		cpb.progress = pb.New64(totalSize)
 	}
-	tmpl := `{{ string . "filePath" }} {{ bar . "-" "█" (cycle . "↖" "↗" "↘" "↙" ) "." "-"}} {{speed . }} {{percent .}} {{ rtime . " ETA - %s"}} {{ etime . " Elapsed - %s"}}`
-	p := pb.ProgressBarTemplate(tmpl).Start64(totalSize)
-	p.Set("filePath", filepath.Base(src))
+	p := pb.Full.Start64(totalSize)
+	p.Set("prefix", filepath.Base(src+": "))
 	p.SetCurrent(currentSize)
 	p.Set(pb.Bytes, true)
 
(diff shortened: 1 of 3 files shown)


pkg/util/progressbar.go
```diff
@@ -45,38 +42,25 @@ type progressBar struct {
...
-
-	if cpb.pool == nil {
-		cpb.pool = pb.NewPool()
-		if err := cpb.pool.Start(); err != nil {
-			glog.Errorf("pool start: %v", err)
-		}
+	if cpb.progress == nil {
+		cpb.progress = pb.New64(totalSize)
 	}
+	tmpl := `{{ string . "filePath" }} {{ bar . "-" "█" (cycle . "↖" "↗" "↘" "↙" ) "." "-"}} {{speed . }} {{percent .}} {{ rtime . " ETA - %s"}} {{ etime . " Elapsed - %s"}}`
```
tstromberg: Any chance you could change this to keep the same UI as before? 

```
	p := pb.Full.Start64(totalSize)
	p.Set("prefix", filepath.Base(src+": "))
```

tstromberg:
Thank you!!!

cvila84:
Great !

pkg/util/progressbar.go
blueelvis: Fixed. Just thought of adding some new pattern :P 

This should be ready for merging now I believe.

RA489:
@minikube-bot Ok to test

tstromberg:
Thanks!

Issue #4967, opened 2019-08-03
start race: x509: certificate is valid for <ips>, not <ip>


medyagh:
making issue out of  @tstromberg conversations n slack:

as seen in integration tests ( the parallel version)

 https://storage.googleapis.com/minikube-builds/logs/4966/Linux-KVM.txt seems to indicate that tests may be running kubelet commands against the wrong minikube context (race condition?).
```
   Temporary Error: error running command [--context=TestContainerd replace -f testdata/nginx-untrusted.yaml --force]: exit status 1. Stdout: 
             Unable to connect to the server: x509: certificate is valid for 192.168.39.145, 10.96.0.1, 10.0.0.1, not 192.168.39.105
```

more errors

```
        kubectl_runner.go:83: Temporary Error: error running command [--context=TestContainerd replace -f testdata/nginx-untrusted.yaml --force]: exit status 1. Stdout: 
             Unable to connect to the server: x509: certificate is valid for 192.168.39.145, 10.96.0.1, 10.0.0.1, not 192.168.39.105
             
        containerd_test.go:95: creating untrusted nginx resource: Temporary Error: error running command [--context=TestContainerd replace -f testdata/nginx-untrusted.yaml --force]: exit status 1. Stdout: 
             Unable to connect to the server: x509: certificate is valid for 192.168.39.145, 10.96.0.1, 10.0.0.1, not 192.168.39.105
             
            Temporary Error: error running command [--context=TestContainerd replace -f testdata/nginx-untrusted.yaml --force]: exit status 1. Stdout: 
             Unable to connect to the server: x509: certificate is valid for 192.168.39.145, 10.96.0.1, 10.0.0.1, not 192.168.39.105
             

```

I feel like it’s cleanest for certs to be per-profile, to avoid raciness. Apparently in that log, one IP is for TestDocker and the other TestContainerd

The issue appears to be two profiles using the same certs.

tstromberg:
Even with locking in `generateCerts`, this is still an issue. Perhaps because the IP information from a previous unlocked bit of code. Here's a recent failure:

https://storage.googleapis.com/minikube-builds/logs/5341/VirtualBox_Linux.txt

```
        I0913 08:03:48.016591   10758 certs.go:61] Setting up certificates for IP: 192.168.99.103
        I0913 08:03:48.016624   10758 certs.go:135] acquiring lock: {Name:generateCerts Clock:{} Delay:10s Timeout:0s Cancel:<nil>}
[...]

PR #5075, opened 2019-08-14
Improve parallel run reliability by putting lock on files before writing

medyagh:
- Hopefully closes https://github.com/kubernetes/minikube/issues/4967 and maybe https://github.com/kubernetes/minikube/issues/5040
- Provide a decorative wrapper arround ioutil.WriteFile that locks the file before writing  
- Provide more reliable parallel minikube experience by locking write file for following operations:
  -- minikube config
  -- profile config
  -- cert generation
  -- notify write time
  -- kubeconfig update 
   -- mount file 
   -- extract 
- Added Exponential Retry logic for acquiring the lock (retry up to 13 seconds)

- Converted all the retry across minikube codebase to exponential retry

NOTE1: this lock mechanism is only respected by minikube binaries. not by other procs in the OS. 
(which is mostly okay!)


pkg/util/lock/lock.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// WriteWithLock decorates ioutil.WriteFile with a file lock and retry
+func WriteFile(filename string, data []byte, perm os.FileMode) error {
+	lock := fslock.New(filename)
+
+	getLock := func() error {
+		lockErr := lock.TryLock()
+		if lockErr != nil {
+			glog.Infof("temproary error : %v", lockErr.Error())
```
tstromberg: Make this a warningf? Also, check spelling on this and next line.

pkg/util/lock/lock.go
```diff
@@ -0,0 +1,57 @@
...
+	"os"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/juju/fslock"
+	"github.com/pkg/errors"
+	"k8s.io/minikube/pkg/util/retry"
+)
+
+// WriteWithLock decorates ioutil.WriteFile with a file lock and retry
```
tstromberg: This should read WriteFile.

pkg/util/lock/lock.go
```diff
@@ -0,0 +1,57 @@
...
+func WriteFile(filename string, data []byte, perm os.FileMode) error {
```
tstromberg: Do you mind calling log.Infof to say we are about to write to a particular file here? It will come in handy for future debugging.

pkg/util/lock/lock.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+		return nil
+	}
+	err := retry.Expo(getLock, 500*time.Millisecond, 13*time.Second)
+	if err != nil {
+		return errors.Wrapf(err, "acquiring file lock for %s", filename)
+	}
+
+	if err := ioutil.WriteFile(filename, data, perm); err != nil {
+		return errors.Wrapf(err, "error writing file %s", filename)
```
tstromberg: As there is no deferred unlock - what happens to the file lock at this point?

pkg/util/lock/lock.go
```diff
@@ -0,0 +1,57 @@
...
+		lockErr := lock.TryLock()
+		if lockErr != nil {
+			glog.Infof("temproary error : %v", lockErr.Error())
+			return errors.Wrapf(lockErr, "falied to acquire file lock for %s > ", filename)
+		}
+		return nil
+	}
+	err := retry.Expo(getLock, 500*time.Millisecond, 13*time.Second)
+	if err != nil {
+		return errors.Wrapf(err, "acquiring file lock for %s", filename)
```
tstromberg: This can just say "lock" rather than "file lock". It's a unique enough error message that we will find it.

Issue #4391, opened 2019-05-30
Download our own kvm2 driver if installed version is missing or out of date


tstromberg:
Related: #3975 #4387 

Also, mentioned in our roadmap: https://github.com/kubernetes/minikube/blob/master/docs/contributors/roadmap.md

josedonizetti:
@tstromberg can I work on this? It shouldn't be hard after we have https://github.com/kubernetes/minikube/issues/4387

medyagh:
Indeed @josedonizetti

josedonizetti:
Looking at the libmachine code, and it expects the driver to be in the PATH. We can't pass a custom path 
of the driver location.

So the download would have to happen to a folder that is already on the path (`/usr/local/bin`?) but for this, we need `sudo` access, and I would prefer users not doing `sudo minikube start --vm-driver=kvm2` only to have the driver downloaded. 

My idea here would be to make the download optional, not force it, and if the user accepts we ask for `sudo` only at this moment. Something like:
```
minikube start --vm-driver=kvm2
Starting....
There's a new version for the kvm2 driver. Would you like to install it? [Y/n]
(link to doc to install yourself)
[sudo] password for XXX:
```
or

```
minikube start --vm-driver=kvm2
Starting....
You don't have the kvm2 driver installed. Would you like to install it? [Y/n]
(link to doc to install yourself)
[sudo] password for XXX:
```

@medyagh @afbjorklund @tstromberg wdyt?

afbjorklund:
Personally I think that a "solution" message is fine, especially if you installed minikube from package.

It is possible that long-term we want to build kvm2 **into** minikube, like we do with virtualbox today.

The downside is that it would require everyone to install libvirt, just for the library dependency to work...

But maybe we can `dlopen` it ? As in only load when needed. Check `github.com/libvirt/libvirt-go`

afbjorklund:
We could cut down on the dependencies, by "only" loading `libvirt.so.0` and `libvirt-qemu.so.0` ?
> By default the binding will support APIs in libvirt.so, libvirt-qemu.so and libvirt-lxc.so. Coverage for the latter two libraries can be dropped from the build using build tags 'without_qemu' or 'without_lxc' respectively.

https://github.com/coreos/pkg/tree/master/dlopen

tstromberg:
> Looking at the libmachine code, and it expects the driver to be in the PATH. We can't pass a custom path
> of the driver location.
> 
[...]

PR #5085, opened 2019-08-15
Download kvm2 driver

josedonizetti:
fix for https://github.com/kubernetes/minikube/issues/4391

### DIFF
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"runtime"
 	"strings"
 
 	"github.com/docker/machine/libmachine"
@@ -107,6 +108,12 @@ func Execute() {
 	RootCmd.Flags().VisitAll(func(flag *pflag.Flag) {
 		flag.Usage = translate.T(flag.Usage)
 	})
+
+	if runtime.GOOS == "linux" {
+		// add minikube binaries to the path
+		targetDir := constants.MakeMiniPath("bin")
+		addToPath(targetDir)
+	}
 
 	if err := RootCmd.Execute(); err != nil {
 		// Cobra already outputs the error, typically because the user provided an unknown command.
@@ -280,4 +287,9 @@ func getClusterBootstrapper(api libmachine.API, bootstrapperName string) (bootst
 	}
 
 	return b, nil
+}
+
+func addToPath(dir string) {
+	path := os.Getenv("PATH")
+	os.Setenv("PATH", fmt.Sprintf("%s:%s", dir, path))
 }
--- a/pkg/drivers/drivers.go
+++ b/pkg/drivers/drivers.go
@@ -151,8 +151,8 @@ func fixPermissions(path string) error {
 	return nil
 }
 
-// Download downloads driver if it is not present, or if the version is old
-func Download(driver, destination string, minikubeVersion semver.Version) error {
+// InstallOrUpdate downloads driver if it is not present, or updates it if there's a newer version
+func InstallOrUpdate(driver, destination string, minikubeVersion semver.Version) error {
 	_, err := exec.LookPath(driver)
 	// if file driver doesn't exist, download it
 	if err != nil {
--- a/pkg/drivers/drivers.go
+++ b/pkg/drivers/drivers.go
@@ -17,6 +17,7 @@ limitations under the License.
 package drivers
 
 import (
+	"fmt"
 	"io"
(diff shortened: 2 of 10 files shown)


cmd/minikube/cmd/root.go
```diff
@@ -290,3 +295,12 @@ func getClusterBootstrapper(api libmachine.API, bootstrapperName string) (bootst
 
 	return b, nil
 }
+
+func addToPath(dir string) {
+	if runtime.GOOS != "linux" {
```
afbjorklund: Shouldn’t this check be in the caller, rather than hidden in the function?

josedonizetti:
@afbjorklund @medyagh  can I get some feedback on how this is design? and any concerns?
I'm currently looking on writing some kind of integration test for it.

cmd/minikube/cmd/root.go
```diff
@@ -290,3 +297,8 @@ func getClusterBootstrapper(api libmachine.API, bootstrapperName string) (bootst
 
 	return b, nil
 }
+
+func addToPath(dir string) {
+	path := os.Getenv("PATH")
```
medyagh: does this work on windows as well ?

medyagh: @blueelvis do you know if this is okay for windows ?

medyagh:
the overall looks good to me ! my only nit is if it works on windows.

josedonizetti:
@medyagh windows won't be supported. At least for now, it will only be for linux/kvm2 driver, and last osx/hyperkit.

cmd/minikube/cmd/root.go
josedonizetti: I don't think it works for windows. That's why the caller validates the OS before calling it. It was actually validated inside the function before, but after a review from Anders, he asked to change it.

pkg/util/downloader.go
```diff
@@ -69,7 +69,7 @@ func (f DefaultDownloader) CacheMinikubeISOFromURL(url string) error {
 	// Predictable temp destination so that resume can function
 	tmpDst := dst + ".download"
 
-	opts := []getter.ClientOption{getter.WithProgress(defaultProgressBar)}
+	opts := []getter.ClientOption{getter.WithProgress(DefaultProgressBar)}
```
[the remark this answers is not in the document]
josedonizetti: I considered that, but end up not doing to avoid doing two different things in the same PR. I can definitely look into it again though.

pkg/drivers/drivers.go
```diff
@@ -136,3 +150,88 @@ func fixPermissions(path string) error {
 	}
 	return nil
 }
+
+// Download downloads driver if it is not present, or if the version is old
+func Download(driver, destination string, minikubeVersion semver.Version) error {
```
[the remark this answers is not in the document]
josedonizetti: - it is called by the start cmd (https://github.com/kubernetes/minikube/pull/5085/files#diff-ed78cd81495af3964aba5fbc898c33a2R941), we don't add callers in comments for no other code, so I will prefer to avoid that here too.
- great idea about the name

pkg/drivers/drivers.go
```diff
@@ -20,15 +20,29 @@ import (
...
+	"github.com/hashicorp/go-getter"
 	"github.com/pkg/errors"
+	"k8s.io/minikube/pkg/version"
+
+	"k8s.io/minikube/pkg/minikube/out"
+	"k8s.io/minikube/pkg/util"
+)
+
+const (
+	driverKVMDownloadURL = "https://storage.googleapis.com/minikube/releases/latest/docker-machine-driver-kvm2"
```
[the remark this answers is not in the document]
josedonizetti: That's a good idea. I also want to make the whole feature have a toggle configuration. You can opt-out of the automatic downloads if you don't agree with it.

josedonizetti:
@medyagh yup, the HyperV driver comes with the libmachine code, it's not an external binary.

tstromberg:
Looks great: Let me know if I can provide any help to dig to the bottom of this integration test failure:

```
--- FAIL: TestDriverInstallOrUpdate (0.03s)
    driver_install_or_update_test.go:70: Expected driver to be download. test: driver-without-version-support, got: stat /tmp/driver-without-version-support758569209/docker-machine-driver-kvm2: no such file or directory
```

josedonizetti:
@minikube-bot Ok to test

josedonizetti:
@minikube-bot Ok to test

josedonizetti:
@minikube-bot  Ok to test

tstromberg:
w00t!

Issue #5081, opened 2019-08-14
macOS: `brew install` should be `brew cask install`


yacn:
When following the installation guide for macOS, it says to install using `brew` with `brew install minikube`. However, `minikube` is cask now so the correct command is `brew cask install minikube`. Not a big issue since the failed command tells you what to try but seems like it might be worth updating.

#### `brew install minikube`

```
Error: No available formula with the name "minikube"
Found a cask named "minikube" instead. Try
  brew cask install minikube
```

afbjorklund:
Basically "cask" is brew-speak for binary, for things that it cannot (or don't want to) build from source.

medyagh:
This make a good first issue to fix the doc !

Patil2099:
@yacn Can I work on this?

PR #5089, opened 2019-08-15
docs: brew install should be brew cask install

Patil2099:
macOS: `brew install` should be `brew cask install`


Patil2099:
Fixes #5081

medyagh:
thank you for fixing the doc ! This will make user experience better! So they won't have to face an error on their first encounter with minikube!

tstromberg:
Thank you!

PR #5093, opened 2019-08-15
minikube is a brew cask package

danielma82:
`brew install minikube` doesn't work for me at least, on my machine - it's reported as a cask and has been for a while.


afbjorklund:
Same as #5089

tstromberg:
Merged #5089

PR #5104, opened 2019-08-16
Provide default InstallURL in EngineOptions for libmachine

afbjorklund:
**Provide defaults for libmachine.**

This URL is normally `https://get.docker.com`, it is used to provision docker if not available:

```go
	if err := installDockerGeneric(provisioner, provisioner.EngineOptions.InstallURL); err != nil {
		return err
	}
```

```go
func installDockerGeneric(p Provisioner, baseURL string) error {
	// install docker - until cloudinit we use ubuntu everywhere so we
	// just install it using the docker repos
	if output, err := p.SSHCommand(fmt.Sprintf("if ! type docker; then curl -sSL %s | sh -; fi", baseURL)); err != nil {
		return fmt.Errorf("error installing docker: %s", output)
	}

	return nil
}
```

Currently we use "", which just leads to a weird error from `curl` (if trying it on another OS).

```json
        "EngineOptions": {
            "ArbitraryFlags": null,
            "Dns": null,
            "GraphDir": "",
            "Env": null,
            "Ipv6": false,
            "InsecureRegistry": [
                "10.96.0.0/12"
            ],
            "Labels": null,
            "LogLevel": "",
            "StorageDriver": "",
            "SelinuxEnabled": false,
            "TlsVerify": false,
            "RegistryMirror": null,
            "InstallURL": ""
        },
```

```
curl: no URL specified!
curl: try 'curl --help' or 'curl --manual' for more information
```

Cherry-picked from the "generic" driver. (~#4734~)

### DIFF
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -338,6 +338,7 @@ func engineOptions(config cfg.MachineConfig) *engine.Options {
 		InsecureRegistry: append([]string{pkgutil.DefaultServiceCIDR}, config.InsecureRegistry...),
 		RegistryMirror:   config.RegistryMirror,
 		ArbitraryFlags:   config.DockerOpt,
+		InstallURL:       drivers.DefaultEngineInstallURL,
 	}
 	return &o
 }


tstromberg:
Do you mind clarifying "why?" in your PR description. I don't understand what the net effect of this change is.

afbjorklund:
> Do you mind clarifying "why?" in your PR description. I don't understand what the net effect of this change is.

Sure thing!

PR #5112, opened 2019-08-17
cleanup: lock.go error message typo fix

ethan-daocloud:
Signed-off-by: Guangming Wang <guangming.wang@daocloud.io>


medyagh:
great catch ! thank you @ethan-daocloud

PR #5116, opened 2019-08-17
Refactor config.ReadConfig to accept a file

josedonizetti:
This PR is the first from a series of PR where I want to decouple `pkg/minikube/config` from working only with `constants.ConfigFile`. The idea is that minikube configuration should be per profile and not one global for everything. I'd like to set different properties depending on the profile I'm working.

Let me know if this makes sense.

@medyagh @tstromberg


cmd/minikube/cmd/cache.go
```diff
@@ -67,7 +67,7 @@ var deleteCacheCmd = &cobra.Command{
 }
 
 func imagesInConfigFile() ([]string, error) {
-	configFile, err := config.ReadConfig()
+	configFile, err := config.ReadConfig(constants.ConfigFile)
```
medyagh: we want to make sure, regardless of profile we use the image cache, and not store the images separately. I haven't studied the code but have you verified that it still uses a global image cache and not re-download them per profile? ( could be a good integration test added)

pkg/minikube/config/config.go
```diff
@@ -88,11 +88,7 @@ func WriteConfig(configFile string, m MinikubeConfig) error {
 }
 
 // ReadConfig reads in the JSON minikube config
-func ReadConfig() (MinikubeConfig, error) {
-	return readConfig(constants.ConfigFile)
-}
-
-func readConfig(configFile string) (MinikubeConfig, error) {
+func ReadConfig(configFile string) (MinikubeConfig, error) {
```
medyagh: I wonder, could this name be changed to something more obvious that this is not a profile config but global minikube config ?

cmd/minikube/cmd/cache.go
josedonizetti: noted

test/integration/fn_minikube_config.go
```diff
@@ -0,0 +1,47 @@
...
+limitations under the License.
+*/
+
+package integration
+
+import (
+	"testing"
+)
+
+func testMinikubeConfig(t *testing.T) {
```
medyagh: better to called it testConfig, it is implied we are testing minikube.
we try to never ever Capitalize minikube :)

medyagh: same as file name f_config is good

josedonizetti: you mentioned `ReadConfig` could use a better name because it is not a profile test, that's why I named this `testMinikubeConfig`, was actually planning to change `ReadConfig` to `ReadMinikubeConfig`.

test/integration/fn_minikube_config.go
```diff
@@ -0,0 +1,47 @@
...
+	t.Parallel()
+	p := profileName(t)
```
medyagh: I wonder if this test will affect the rest of the tests running in parralell, if so, feel free to give it a different profile name and do the testing without starting it.


that brings a question, doest minikube write to the config file if it is not started?

medyagh: could be given a 
p = profileName(t)+config

medyagh:
Thanks for adding integration tests for it ! I am interested to test it against a Not Started minikube profile

test/integration/fn_minikube_config.go
josedonizetti: yup, it writes to the config even when stopped. What we can do here, is to actually set a default value, this wouldn't affect any other parallel test.

josedonizetti: sure

josedonizetti:
@medyagh currently I'm not calling `Start' on minikubeRunner, is that what you are expecting?

medyagh:
No sorry for confusion the profile you used is already started by the parent test.
 I meant the config test could use it's own profile and not start it at all

medyagh:
interesting fail on the none:

```
     --- FAIL: TestFunctional/group/config (0.12s)
            minikube_runner.go:119: Error running command: -p=minikubeconfig config get cpus . Output:  Stderr: Error: specified key could not be found in config

```

Thanks again for adding this integration tests, this will open our eyes to what minikube REALLY does. and make sure we don't break it in the future.

test/integration/config_test.go
```diff
@@ -0,0 +1,81 @@
...
+			stdout: "",
+			stderr: ""},
+		{
+			cmd:    "config get cpus",
+			stdout: "",
+			stderr: "Error: specified key could not be found in config",
+		},
+	}
+
+	for _, test := range tests {
```
medyagh: rename test to tc (to be more consistent with other tests and not getting confused with go test)

medyagh:
Looks good ! the nits are optional.

tstromberg:
*KVM_Linux*
`*TestStartStop/group/cni*: StartHost: Temporary Error: Error configuring auth on host: OS type not recognized`

*VirtualBox_macOS*
`*TestProxy/ProxyDashboard*: failed to read url: timeout after 3m0s`

*VirtualBox_Linux*
`*TestDocker*:  got state: "", expected "Running`

These all appear to be unrelated flakes.

tstromberg:
Thank you!

Issue #5083, opened 2019-08-14
vbox test: "rm: cannot remove '/var/lib/minikube/certs/ca.crt': Buffer I/O error on device sda1


medyagh:
As seen in:
-  virtualbox mac os  https://storage.googleapis.com/minikube-builds/logs/5075/VirtualBox_macOS.txt
  and https://storage.googleapis.com/minikube-builds/logs/5082/VirtualBox_macOS.txt
-  virtualbox windows  https://github.com/kubernetes/minikube/issues/4342 and also another 

~~this could be side effect of https://github.com/kubernetes/minikube/pull/5075~~
I was wrong this is not side effect of that PR,

it seems to be in both cases it has happend in Virtualbox (windows and macos)

```
--- FAIL: TestPersistence (1502.40s)
        sudo systemctl daemon-reload &&
        sudo systemctl start kubelet
        I0814 11:47:13.448775    1984 certs.go:49] Setting up certificates for IP: 192.168.99.101
        I0814 11:47:13.452133    1984 lock.go:33] attempting to write to file "/Users/jenkins/minikube-integration/darwin-amd64-virtualbox-5075-641-9fad24aa37f48dd18ce5e76a0714a4a642ee4fed/.minikube/client.crt" with filemode -rw-r--r--
        I0814 11:47:13.452383    1984 lock.go:33] attempting to write to file "/Users/jenkins/minikube-integration/darwin-amd64-virtualbox-5075-641-9fad24aa37f48dd18ce5e76a0714a4a642ee4fed/.minikube/client.key" with filemode -rw-------
        I0814 11:47:13.455882    1984 lock.go:33] attempting to write to file "/Users/jenkins/minikube-integration/darwin-amd64-virtualbox-5075-641-9fad24aa37f48dd18ce5e76a0714a4a642ee4fed/.minikube/apiserver.crt" with filemode -rw-r--r--
        I0814 11:47:13.456142    1984 lock.go:33] attempting to write to file "/Users/jenkins/minikube-integration/darwin-amd64-virtualbox-5075-641-9fad24aa37f48dd18ce5e76a0714a4a642ee4fed/.minikube/apiserver.key" with filemode -rw-------
[...]

PR #5121, opened 2019-08-18
Add wait-timeout flag to start command and refactor util/kubernetes

medyagh:
- Removes more than a few unused functions such as :
    - NewPodStore, StartPods, WaitForPodDelete, WaitForEvent , WaitForServiceEndpointsNum, VersionedExtraOption
-  Moves funcs out of `pkg/utils` into `pgk/kube`.
-  Moves `ExtraOptions` type and its funcs from `pkg/util` to `pkg/minikube/config` package. and Rename  func `.ContainsString` to `.ContainsParam`
- Choose better timeout based on [kubernetes consts](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/constants/constants.go#L187)
- Added logs for how long it took for each component and k8s-app to come up ( to be used later to fine-tune our default waiting time) which will appear in the logs like  : 
     ```
      kube.go:103] duration metric: took 1m17.514019465s to wait for component=etcd ...
      kube.go:103] duration metric: took 7.289066ms to wait for component=kube-scheduler ...
    ```

- Added a new flag for start cmd  `wait-timeout` that specifies max `wait per component`.
- Reduced the default `wait per component` from 5 minutes to 3 minutes for end users
- Increase default `wait per component` from 5 minutes to  13 minutes for parallel integration tests. 
- Parameterized integration tests to accept `wait-timeout`
- Fixed the test setup for givsor test which was put after the e2e test run. moved up the script.

Closes https://github.com/kubernetes/minikube/issues/5122
and Hopefully reduces some test flakes due to timeout  

---
topics I like to know the reviewer's opinion on :
- name of the package, "kube" vs "kubernetes"
- path of package,  current: "k8s.io/minikube/pkg/kube" vs "k8s.io/minikube/pkg/util/kube"

--- 
### Golang Parallel Test logging gotcha! :

I was hoping to see the duration metric logs in the tests ! 
```
	elapsed := time.Since(start)
       glog.Infof("duration metric: took %s to wait for %s ...", elapsed, label)
```

but I found out sometimes it doesn't log at all and sometimes logs at most 2 sets of them. 

- KMV logs shows only two sets of duration metrics https://storage.googleapis.com/minikube-builds/logs/5121/KVM_Linux.txt
- Hyperkit logs showed no duration metric at all !!! https://storage.googleapis.com/minikube-builds/logs/5121/HyperKit_macOS.txt   (the logs might be overwritten by the next commit)

That makes me believe golang only outputs the not PAUSED Tests and the paused tests (which their VM still are running and our wait for running func is still working on them) will not output to the logs.  I created an issue in golang to track this : https://github.com/golang/go/issues/33706

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -46,6 +46,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/bootstrapper/kubeadm"
 	"k8s.io/minikube/pkg/minikube/cluster"
 	"k8s.io/minikube/pkg/minikube/command"
+	"k8s.io/minikube/pkg/minikube/config"
 	cfg "k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/minikube/cruntime"
@@ -111,7 +112,7 @@ var (
 	insecureRegistry []string
 	apiServerNames   []string
 	apiServerIPs     []net.IP
-	extraOptions     pkgutil.ExtraOptionSlice
+	extraOptions     config.ExtraOptionSlice
 )
 
 func init() {
@@ -538,8 +539,8 @@ func validateConfig() {
 
 	// check that kubeadm extra args contain only whitelisted parameters
 	for param := range extraOptions.AsMap().Get(kubeadm.Kubeadm) {
-		if !pkgutil.ContainsString(kubeadm.KubeadmExtraArgsWhitelist[kubeadm.KubeadmCmdParam], param) &&
-			!pkgutil.ContainsString(kubeadm.KubeadmExtraArgsWhitelist[kubeadm.KubeadmConfigParam], param) {
+		if !cfg.ContainsParam(kubeadm.KubeadmExtraArgsWhitelist[kubeadm.KubeadmCmdParam], param) &&
+			!cfg.ContainsParam(kubeadm.KubeadmExtraArgsWhitelist[kubeadm.KubeadmConfigParam], param) {
 			exit.UsageT("Sorry, the kubeadm.{{.parameter_name}} parameter is currently not supported by --extra-config", out.V{"parameter_name": param})
 		}
 	}
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -103,6 +103,7 @@ const (
 	dnsProxy              = "dns-proxy"
 	hostDNSResolver       = "host-dns-resolver"
 	waitUntilHealthy      = "wait"
+	waitTimeout           = "wait-timeout"
 )
 
 var (
@@ -149,6 +150,7 @@ func initMinikubeFlags() {
 	startCmd.Flags().String(networkPlugin, "", "The name of the network plugin.")
 	startCmd.Flags().Bool(enableDefaultCNI, false, "Enable the default CNI plugin (/etc/cni/net.d/k8s.conf). Used in conjunction with \"--network-plugin=cni\".")
 	startCmd.Flags().Bool(waitUntilHealthy, true, "Wait until Kubernetes core services are healthy before exiting.")
+	startCmd.Flags().Duration(waitTimeout, 3*time.Minute, "max time to wait for Kubernetes core services to be healthy.")
 }
 
 // initKubernetesFlags inits the commandline flags for kubernetes related options
@@ -322,7 +324,7 @@ func runStart(cmd *cobra.Command, args []string) {
 	// special ops for none driver, like change minikube directory.
 	prepareNone(viper.GetString(vmDriver))
 	if viper.GetBool(waitUntilHealthy) {
-		if err := bs.WaitCluster(config.KubernetesConfig); err != nil {
+		if err := bs.WaitCluster(config.KubernetesConfig, viper.GetDuration(waitTimeout)); err != nil {
 			exit.WithError("Wait failed", err)
 		}
 	}
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -150,7 +150,7 @@ func initMinikubeFlags() {
 	startCmd.Flags().String(networkPlugin, "", "The name of the network plugin.")
 	startCmd.Flags().Bool(enableDefaultCNI, false, "Enable the default CNI plugin (/etc/cni/net.d/k8s.conf). Used in conjunction with \"--network-plugin=cni\".")
 	startCmd.Flags().Bool(waitUntilHealthy, true, "Wait until Kubernetes core services are healthy before exiting.")
-	startCmd.Flags().Duration(waitTimeout, 3*time.Minute, "max time to wait for Kubernetes core services to be healthy.")
+	startCmd.Flags().Duration(waitTimeout, 3*time.Minute, "max time to wait per Kubernetes core services to be healthy.")
 }
 
 // initKubernetesFlags inits the commandline flags for kubernetes related options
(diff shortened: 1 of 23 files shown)


medyagh:
the flaky test is related to virtualbox corrupt issue https://github.com/kubernetes/minikube/issues/5083

cmd/minikube/cmd/start.go
```diff
@@ -148,6 +150,7 @@ func initMinikubeFlags() {
 	startCmd.Flags().String(networkPlugin, "", "The name of the network plugin.")
 	startCmd.Flags().Bool(enableDefaultCNI, false, "Enable the default CNI plugin (/etc/cni/net.d/k8s.conf). Used in conjunction with \"--network-plugin=cni\".")
 	startCmd.Flags().Bool(waitUntilHealthy, true, "Wait until Kubernetes core services are healthy before exiting.")
+	startCmd.Flags().Duration(waitTimeout, 3*time.Minute, "max time to wait for Kubernetes core services to be healthy.")
```
tstromberg: This default seems quite short for certain environments: Previously, it was 5-minutes per pod. 5 minutes overall perhaps?

pkg/kube/kube.go
```diff
@@ -92,50 +68,16 @@ func GetClient(kubectlContext ...string) (kubernetes.Interface, error) {
...
+	start := time.Now()
 	glog.Infof("Waiting for pod with label %q in ns %q ...", ns, label)
 	lastKnownPodNumber := -1
-	return wait.PollImmediate(constants.APICallRetryInterval, ReasonableStartTime, func() (bool, error) {
+	f := func() (bool, error) {
 		listOpts := meta.ListOptions{LabelSelector: label.String()}
 		pods, err := c.CoreV1().Pods(ns).List(listOpts)
 		if err != nil {
-			glog.Infof("error getting Pods with label selector %q [%v]\n", label.String(), err)
+			glog.Infof("temproary error: getting Pods with label selector %q : [%v]\n", label.String(), err)
```
tstromberg: check spelling of temporary. 
Also, no need for the space before the colon.

pkg/kube/kube.go
```diff
@@ -150,45 +92,24 @@ func WaitForPodsWithLabelRunning(c kubernetes.Interface, ns string, label labels
 
 		for _, pod := range pods.Items {
 			if pod.Status.Phase != core.PodRunning {
+				glog.Infof("temporary error: for Pod with label %q expected status to be running but got %s : [%v]\n", label.String(), pod.Status.Phase, err)
```
tstromberg: I'm not sure this is an error, and I also find this difficult to read. Could this be reworded with less words? Perhaps just:

"waiting for pod %q, current state: %s"

pkg/kube/kube.go
```diff
@@ -150,45 +92,24 @@ func WaitForPodsWithLabelRunning(c kubernetes.Interface, ns string, label labels
...
-		}
-		return false, nil
-	})
+	}
+	t := ReasonableStartTime
+	if timeOut != nil {
+		t = timeOut[0]
+	}
+	err := wait.PollImmediate(kconst.APICallRetryInterval, t, f)
+	elapsed := time.Since(start)
```
tstromberg: This value is only used in the log message, so just inline it into the `glog.Infof` by passing `time.Since(start)` in as an  argument directly.

pkg/kube/kube.go
```diff
@@ -150,45 +92,24 @@ func WaitForPodsWithLabelRunning(c kubernetes.Interface, ns string, label labels
...
+	glog.Infof("duration metric: took %s to wait for %s ...", elapsed, label)
+	return err
 }
 
-// WaitForRCToStabilize waits till the RC has a matching generation/replica count between spec and status.
+// WaitForRCToStabilize waits till the RC has a matching generation/replica count between spec and status. used by integration tests
```
tstromberg: Rather than promising something that you cannot guarantee in the future, would it be possible to move this and the next function into the integration tests? Feel free to say no.

pkg/kube/kube.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package util
+package kube
```
tstromberg: I feel like this needs a more obvious package name. Is `kube` kubernetes or `minikube`, 'kubeadm'? It feels like it could be so many things. These are all calls that talk to the Kubernetes API server. 

Some alternative ideas: `kubernetes`, `kapi`, `kubeclient`, 'apiclient', 'kapiclient'?

pkg/minikube/bootstrapper/kubeadm/versions.go
```diff
@@ -158,8 +159,8 @@ func Supports(featureName string) bool {
 	return false
 }
 
-// ParseKubernetesVersion parses the kubernetes version
-func ParseKubernetesVersion(version string) (semver.Version, error) {
+// arseKubernetesVersion parses the kubernetes version
```
tstromberg: typo

pkg/minikube/bootstrapper/kubeadm/versions.go
```diff
@@ -158,8 +159,8 @@ func Supports(featureName string) bool {
...
+func parseKubernetesVersion(version string) (semver.Version, error) {
```
tstromberg: Thank you for making this private.

test/integration/flags.go
```diff
@@ -32,7 +32,7 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
-var startTimeout = flag.Int("timeout", 25, "number of minutes to wait for minikube start")
+var startTimeout = flag.Duration("timeout", 25*time.Minute, "max duration to wait for a full minikube start")
```
tstromberg: 8x the default we give to users is crazy. I can understand 2x, but more than that I feel like we are making ourselves reliable at the expense of users.

hack/jenkins/common.sh
```diff
@@ -241,6 +241,11 @@ export MINIKUBE_HOME="${TEST_HOME}/.minikube"
 export MINIKUBE_WANTREPORTERRORPROMPT=False
 export KUBECONFIG="${TEST_HOME}/kubeconfig"
 
+# Build the gvisor image. This will be copied into minikube and loaded by ctr.
+# Used by TestContainerd for Gvisor Test.
+docker build -t gcr.io/k8s-minikube/gvisor-addon:latest -f testdata/gvisor-addon-Dockerfile out
```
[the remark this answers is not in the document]
medyagh: This was added by Priya to fix the gvisor test ( to be added before the integration tests) but it was added after the minikube clean up ! so it was not being used by the test.

cmd/minikube/cmd/start.go
medyagh: I should had added Per Kubernetes service... ( this is not for all) and it used to be 5 min per componenent

pkg/kube/kube.go
medyagh: I like kapi ! the kubeclient would stutter for the name kbueclient.Client()

test/integration/flags.go
medyagh: sorry for the confusion, the --wait-time out is Per component not for all start.

before this PR the total wait was 5 x Per component
this PR reduces it to 3 min for end user (per component)

so it is not too much more than what we do for the end user. (if we count 5* per component")

but I intend to have another clean up PR in the integration tests ,to get rid of all kind of Retrying Start in parallel. once fixed all the flakes ( certs, and corruptions...)

hack/jenkins/common.sh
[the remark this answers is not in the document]
medyagh: this only moved the command up the script

medyagh:
@tstromberg I believe I solved all the comments

PR #5123, opened 2019-08-18
Altered root user check for vm-driver=none

WurmNL:
Altered current user check to use user.Username as some distro's don't have the "Name" set. This caused me to "add" root to the root user on ArchLinux:

root:x:0:0:**thiscanbeanything**:/root:/bin/bash

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -503,7 +503,7 @@ func validateUser() {
 	d := viper.GetString(vmDriver)
 	// Check if minikube needs to run with sudo or not.
 	if err == nil {
-		if d == constants.DriverNone && u.Name != "root" {
+		if d == constants.DriverNone && u.Username != "root" {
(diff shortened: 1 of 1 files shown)


afbjorklund:
Same as #4995 - not sure why that was "merged" with the Windows changes though ? @bluelvis ?

As per the earlier reports, we might want to check for UID 0 rather than testing the login name...

medyagh:
please sign the CLA https://identity.linuxfoundation.org/projects/cncf

blueelvis:
@afbjorklund - #4995 hasn't been merged. medya closed it in favor of #4745. I am still working on fixing those integration test issues. Details in there. Once that is merged, all of these issues would be fixed.

afbjorklund:
Just saying that I think we should fix this one-liner first, and do those unrelated changes later.

tstromberg:
I'm OK if this gets merged in first, so long as it is checking the correct field (uid).

RA489:
@minikube-bot OK to test

tstromberg:
Obsoleted by #5179 - thank you for bringing this to our attention!

Issue #3512, opened 2019-01-08
Migrate away from rootfs / DOCKER_RAMDISK


tstromberg:
This will allow us to use pivot_root and mitigate security issues which involve escaping containers.

afbjorklund:
The security issues should be addressed now, right ? With https://github.com/opencontainers/runc/commit/28a697cce3e4f905dca700eda81d681a30eef9cd

AkihiroSuda:
Mitigated but `chroot` is still unrecommended

cc @cyphar

cyphar:
Yes, I absolutely recommend against using `chroot`. It's simply not secure, and continuing to use it is a bad idea -- the security you get from `chroot` is incredibly minimal compared to the security you get from `pivot_root`. I'm actually of half a mind to remove `chroot` entirely, that's how bad of an idea it is to use it.

afbjorklund:
OK. Moving away from [Buildroot](https://buildroot.org/) (already from Boot2Docker) probably means some work, though... From what we have seen [so far](https://github.com/minishift/minishift-centos-iso), it will also make the footprint (in terms of the various disk images) larger. So it depends on how important it is to make the **local** VM secure - we don't use `no_pivot_root` for remote host.

i.e. the default user has access to `sudo` (through %wheel) and the default user has a known password...

cyphar:
In theory the fix could be as simple as just having a `tmpfs` for Docker. I don't know how the images are stored (and if they're baked into images that might be difficult) but since `initramfs` is already in-memory there isn't a difference with `tmpfs` other than the fact you need to load it on-boot.

afbjorklund:
Interesting! Today we have "everything" on the rootfs (besides the usual runtime suspects), and then have the docker/containers on a mounted disk (for persistance). Using overlay2/overlay storage drivers.

Something like this: (with the cgroup/bind-mounts/overlay/shm excluded)
```
Filesystem      Size  Used Avail Use% Mounted on
rootfs             0     0     0    - /
devtmpfs        906M     0  906M   0% /dev
sysfs              0     0     0    - /sys
proc               0     0     0    - /proc
tmpfs           996M     0  996M   0% /dev/shm
devpts             0     0     0    - /dev/pts
tmpfs           996M  102M  895M  11% /run
tmpfs           996M     0  996M   0% /sys/fs/cgroup
hugetlbfs          0     0     0    - /dev/hugepages
nfsd               0     0     0    - /proc/fs/nfsd
[...]

PR #5133, opened 2019-08-19
Move root filesystem from rootfs to tmpfs

afbjorklund:
This moves the root filesystem of the ISO, from rootfs to tmpfs.

This allows us to stop having to use `--no-pivot` flag with `runc`.

The Docker configuration is flexible enough to work with both...

Thanks to @kfox1111 for suggesting a much easier way to do it.

Issue #3512


kfox1111:
Looks good to me. :)

pkg/provision/buildroot.go
```diff
@@ -155,6 +168,14 @@ WantedBy=multi-user.target
 	}, nil
 }
 
+func rootFileSystemType(p *BuildrootProvisioner) (string, error) {
+	fs, err := p.SSHCommand("df --output=fstype / | sed 1d")
```
tstromberg: Consider using `tail -n 1`, as it is likely to be more obvious to the mere mortals who are not one with the stream editor.

deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/init
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+mkdir /sysroot
+mount -t tmpfs -o size=90% tmpfs /sysroot
```
tstromberg: Add a comment explaining how you arrived at 90% being the correct choice here.

deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/init
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
tstromberg: For future archeologists, do you mind adding a quick comment describing why this file exists, and what it's intent is?

medyagh: future archeologists hahaha ! love it ! hello future archeologists from 2019 !

pkg/provision/buildroot.go
afbjorklund: Upstream suggests it is the way to go, rather than silly little things like `--without-header` etc. But I could use `tail`...

* https://lists.gnu.org/archive/html/coreutils/2012-12/msg00066.html

deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/init
afbjorklund: That’s easy, I stole it from tinycore (and boot2docker). 😀
I think it needs some space left, so using 100% won’t work. But we could probably get away with 95% since our VM is twice as big ? Can experiment with it

deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/init
afbjorklund: Explaining init(1) ?

Or maybe what is the difference between this init and the normal one would be more appropriate. It doesn’t really help that the final `/sbin/init` is actually **not** init but just Lennart (systemd) posing

leseb:
This should also close https://github.com/kubernetes/minikube/issues/4072. Thanks!

deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/init
tstromberg: That's fine. Just mention that you have copied the number from elsewhere then. I was afraid it might have some magical principle to it.

deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/init
tstromberg: I for one didn't realize this took the place of the system init, though it makes sense given the location. 

It'd be nice to have an explanation of why data is being copied from one directory to another before exec'ing out to another init. =)

pkg/provision/buildroot.go
cyphar: As I mentioned in the other bug report, why not use `stat --file-system --format '%T' /`?

afbjorklund: Good idea!

deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/init
afbjorklund: Settled for adding a reference to `switch_root(8)`

>DESCRIPTION
>       switch_root moves already mounted /proc, /dev, /sys and /run to newroot and makes newroot the new root filesystem and starts init process.
>
> WARNING: switch_root removes recursively all files and directories on the current root filesystem.

deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/init
afbjorklund: I'll stick with 90%, until I know better. Will leave a reference to the original

afbjorklund:
@TravisBuddy : Build failed due to #5176

kfox1111:
Looks like the travis fix may be merged.
/retest

medyagh:
> @TravisBuddy : Build failed due to #5176

@TravisBuddy sorry it was my fault ! should had pulled upstream

pkg/provision/buildroot.go
```diff
@@ -92,6 +92,13 @@ func (p *BuildrootProvisioner) GenerateDockerOptions(dockerPort int) (*provision
 	driverNameLabel := fmt.Sprintf("provider=%s", p.Driver.DriverName())
 	p.EngineOptions.Labels = append(p.EngineOptions.Labels, driverNameLabel)
 
+	noPivot := true
+	// Using pivot_root is not supported on fstype rootfs
+	if fstype, err := rootFileSystemType(p); err == nil {
+		log.Debugf("root file system type: %s", fstype)
+		noPivot = fstype == "rootfs"
+	}
```
cyphar: I would suggest making this a warning -- again, `--no-pivot` is fundamentally insecure.

afbjorklund: Note that switching to stat broke everything, since it has a different return value

afbjorklund: For some reason, stat thinks that the file system is called "ramfs" (not "rootfs")

afbjorklund: Added a warning, not likely to be seen - but only valid for old ISO, not necessary for new ISO

tstromberg:
PR looks good, but I'm concerned about the apparently reliable Gvisor failure:

```
FAIL: TestContainerd/GvisorRestart (613.33s)
...
08:44:04 | ! [kubelet-check] Initial timeout of 40s passed.
```

We'll need to confirm if this issue occurs at master without this PR.

afbjorklund:
The test results look fairly non-obvious to me.

They seem to fail, no matter if the code is broken (7b5f648) or if it is working (edea8f9) ?
So either I will override and go on pushing broken code, or be afraid to submit working code.

That being said, it works here (don't use gvisor)

pkg/provision/buildroot.go
cyphar: `ramfs` === `rootfs` (both tools are fetching the filesystem using the same method -- `statfs(2)` and checking the filesystem magic number against `RAMFS_SUPER_MAGIC`). But I don't have a strong opinion either way. Thanks.

leseb:
What's the target release for this? Thanks

afbjorklund:
v1.4.0

PR #5160, opened 2019-08-21
fix wrong spells in  kapi.go

AllenZMC:
Make the comment more readable


medyagh:
Thanks for catching this

PR #5165, opened 2019-08-21
Replace file lock with mutex

### LINKED ISSUE #5149: Starting minikube on windows is broken at master
` .\minikube-windows-amd64.exe start --vm-driver=hyperv --hyperv-virtual-switch="Default Switch"`

```* minikube v1.3.1 on Microsoft Windows 10 Pro 10.0.18362 Build 18362
* Creating hyperv VM (CPUs=2, Memory=2000MB, Disk=20000MB) ...
E0820 12:32:23.767224   15160 start.go:758] StartHost: create: provisioning: Temporary Error: getting cluster config: unexpected end of JSON input
* Tip: Use 'minikube start -p <name>' to create a new cluster, or 'minikube delete' to delete this one.```

It seems like we're writing an empty `config.json` and trying to read it and failing.

### PR DESCRIPTION
**Pull request by sharifelgamal**
The file lock was breaking some file writing in windows, causing startup to fail silently.

Fixes #5149

### CODE REVIEW

**Comment by medyagh** [comment]
nice !

File: pkg/util/lock/lock.go [hunk 1]
```diff
@@ -17,45 +17,66 @@ limitations under the License.
...
+
+func getMutexName(filename string) string {
+	// Make the mutex name the file name and its parent directory
+	dir, name := filepath.Split(filename)
+
+	// Replace underscores and periods with dashes, the only valid punctuation for mutex name
+	name = strings.ReplaceAll(name, ".", "-")
+	name = strings.ReplaceAll(name, "_", "-")
+
+	profile := strings.ReplaceAll(filepath.Base(dir), ".", "-")
```
medyagh: please change the profile name to something else, anything, 
maybe p

medyagh: by profile did you mean to get minikube's base dir ? how about using constants.GetMiniPath

sharifelgamal: no, I literally just want to use the file name and its direct parent directory name. I'll change the variable name.

pkg/util/lock/lock.go
```diff
@@ -17,45 +17,66 @@ limitations under the License.
...
+	defer releaser.Release()
+
 	if err = ioutil.WriteFile(filename, data, perm); err != nil {
 		return errors.Wrapf(err, "error writing file %s", filename)
 	}
 
 	return err
 }
+
+func getMutexName(filename string) string {
```
josedonizetti: Although this is not a public function, I would prefer to see a test for it.

sharifelgamal: I think that's fair. Added.

sharifelgamal:
@minikube-bot OK to test

PR #5179, opened 2019-08-22
Exit if uid=0, add --force flag to override

### LINKED ISSUE #3138: minikube should fail early if executed as root
This is a bad practice, and could address some awkward connectivity and authentication issues, like #2548

tstromberg:
New behavior:

```
sudo ./out/minikube start
😄  minikube v1.3.1 on Darwin 10.14.6
🛑  The virtualbox driver should not be used with root privileges.
💡  Tip: To remove this root owned cluster, run: sudo minikube delete
💣  Exiting
```

or:

```
sudo ./out/minikube start --force 
😄  minikube v1.3.1 on Darwin 10.14.6
⚠️  WARNING: The virtualbox driver should not be used with root privileges.
```

Closes #3138
Obsoletes #5179

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -102,6 +102,7 @@ const (
 	dnsProxy              = "dns-proxy"
 	hostDNSResolver       = "host-dns-resolver"
 	waitUntilHealthy      = "wait"
+	force                 = "force"
 	waitTimeout           = "wait-timeout"
 )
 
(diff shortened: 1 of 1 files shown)


tstromberg:
@minikube-bot OK to test

cmd/minikube/cmd/start.go
```diff
@@ -502,25 +505,33 @@ func selectImageRepository(mirrorCountry string, k8sVersion string) (bool, strin
...
-		}
+	useForce := viper.GetBool(force)
 
-	} else {
-		glog.Errorf("Error getting the current user: %v", err)
+	if d == constants.DriverNone && u.Uid != "0" && !useForce {
+		exit.WithCodeT(exit.Permissions, `The "{{.driver_name}}" driver requires root privileges. Please run minikube using 'sudo minikube --vm-driver={{.driver_name}}'.`, out.V{"driver_name": d})
 	}
 
+	if d == constants.DriverNone || u.Uid != "0" {
```
afbjorklund: Wouldn't this always exit, for driver none ? Or maybe that is the "happy path" of validateUser

afbjorklund: Never mind, the return will just go the same place as before (the checks were reversed)

afbjorklund:
This might lead to some support issues, since configuration is unreadable (owned by root)

tstromberg:
> This might lead to some support issues, since configuration is unreadable (owned by root)

Good point. Let me add something to detect if there is an existing config and offer up a solution.

PR #5224, opened 2019-08-28
Windows: Use path.Join for VM paths, fall-back to anon image downloads

tstromberg:
With kubeadm, we are generating paths that must be valid inside of the Linux VM.

On Windows, `filepath.Join` uses \ instead of /

The image download fallback was required for unknown reasons on our CI machine.


afbjorklund:
I thought we _wanted_ bizarre paths on Windows, like `C:\Windows` and friends.

PR #5244, opened 2019-09-02
Change minikube config list to minikube config in documentation

dpandhi-git:
Changed minikube config list to minikube config view

### DIFF
--- a/site/content/en/docs/Reference/Configuration/minikube.md
+++ b/site/content/en/docs/Reference/Configuration/minikube.md
@@ -130,7 +130,7 @@ minikube config set vm-driver hyperkit
 
 ### Config variables
 
-minikube supports passing environment variables instead of flags for every value listed in `minikube config list`.  This is done by passing an environment variable with the prefix `MINIKUBE_`.
+minikube supports passing environment variables instead of flags for every value listed in `minikube config`.  This is done by passing an environment variable with the prefix `MINIKUBE_`.
 
 For example the `minikube start --iso-url="$ISO_URL"` flag can also be set by setting the `MINIKUBE_ISO_URL="$ISO_URL"` environment variable.
 
--- a/site/content/en/docs/Reference/environment_variables.md
+++ b/site/content/en/docs/Reference/environment_variables.md
@@ -7,7 +7,7 @@ date: 2019-08-01
 
 ## Config option variables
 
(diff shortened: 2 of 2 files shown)


dpandhi-git:
@josedonizetti  Thanks for info. But when i tested with latest version of minikube I do not see list as an option. Am i missing anything?

`[dpandhi out]$  minikube version
minikube version: v1.3.1
commit: ca60a424ce69a4d79f502650199ca2b52f29e631
[dpandhi out]$  minikube config --help
config modifies minikube config files using subcommands like "minikube config set vm-driver kvm"
Configurable fields: 

 * vm-driver
 * feature-gates
 * v
 * cpus
 * disk-size
 * host-only-cidr
 * memory
 * log_dir
 * kubernetes-version
 * iso-url
 * WantUpdateNotification
 * ReminderWaitPeriodInHours
 * WantReportError
 * WantReportErrorPrompt
 * WantKubectlDownloadMsg
 * WantNoneDriverWarning
 * profile
 * bootstrapper
 * ShowDriverDeprecationNotification
 * ShowBootstrapperDeprecationNotification
 * dashboard
 * addon-manager
 * default-storageclass
 * heapster
 * efk
 * ingress
 * registry
 * registry-creds
 * freshpod
 * default-storageclass
 * storage-provisioner
 * storage-provisioner-gluster
 * metrics-server
 * nvidia-driver-installer
 * nvidia-gpu-device-plugin
 * logviewer
 * gvisor
 * hyperv-virtual-switch
 * disable-driver-mounts
 * cache
 * embed-certs

Available Commands:
  get         Gets the value of PROPERTY_NAME from the minikube config file
  set         Sets an individual value in a minikube config file
  unset       unsets an individual value in a minikube config file
  view        Display values currently set in the minikube config file

Usage:
  minikube config SUBCOMMAND [flags] [options]

Use "minikube <command> --help" for more information about a given command.
Use "minikube config options" for a list of global command-line options (applies to all commands).`

josedonizetti:
@dpandhi-git  You are correct. There's not `minikube config list`, but yet the object here is to display all possible configurations, and not the current set which is what `minikube config view` does.

Do you mind changing it to be `minikube config`, this will list the possible options. Also, there's another place to change this: https://github.com/kubernetes/minikube/blob/5b215eb7521a9dde1ea265089dbebb0d03a9ee09/site/content/en/docs/Reference/Configuration/minikube.md#config-variables

WDYT?

PR #5275, opened 2019-09-05
Add - for ~ replacement in release_build_and_upload

tstromberg:


afbjorklund:
Not all older systems accept tilde (it was not in the LSB), but alternative (“1.3.1.999”) requires more work

PR #5284, opened 2019-09-06
Refactor integration tests for a linear parallel magic-free experience

tstromberg:
Major changes:

- Tests are now designed to be read as a single linear testcase. 
- Tests can now timeout independently
- `minikube logs` is emitted when a test group fails
- Helper functions that masked information, particularly `MinikubeRunner` and `KubectlRunner` have been removed.
- `TestFunctional` shares a single running minikube profile between all tests.
- New flag: `--profile` to run tests against a currently running profile
- New flag: `--cleanup` to prevent cleanup


tstromberg:
Will re-open with better commits.

Issue #3558, opened 2019-01-19
none: minikube service/dashboard should show URL instead of opening browser when run with sudo


Pacheco95:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): BUG REPORT

Please provide the following details:

**Environment**:
```
System:    Host: MICHAEL-PC Kernel: 4.15.0-43-generic x86_64 bits: 64 compiler: gcc v: 7.3.0 
           Desktop: Cinnamon 4.0.9 wm: muffin dm: LightDM Distro: Linux Mint 19.1 Tessa 
           base: Ubuntu 18.04 bionic 
Machine:   Type: Desktop Mobo: ASUSTeK model: PRIME H370M-PLUS v: Rev 1.xx serial: <filter> 
           UEFI: American Megatrends v: 0904 date: 09/14/2018 
CPU:       Topology: 6-Core model: Intel Core i5-8600K bits: 64 type: MCP arch: Kaby Lake rev: A 
           L2 cache: 9216 KiB 
           flags: lm nx pae sse sse2 sse3 sse4_1 sse4_2 ssse3 vmx bogomips: 43200 
           Speed: 800 MHz min/max: 800/4300 MHz Core speeds (MHz): 1: 800 2: 800 3: 800 4: 800 
           5: 800 6: 800 
Graphics:  Device-1: NVIDIA GP104 vendor: ASUSTeK driver: nvidia v: 390.77 bus ID: 01:00.0 
           chip ID: 10de:1b82 
           Display: x11 server: X.Org 1.19.6 driver: nvidia 
           unloaded: fbdev,modesetting,nouveau,vesa resolution: 1360x768~60Hz 
           OpenGL: renderer: GeForce GTX 1070 Ti/PCIe/SSE2 v: 4.6.0 NVIDIA 390.77 
           direct render: Yes 
Audio:     Device-1: Intel vendor: ASUSTeK driver: snd_hda_intel v: kernel bus ID: 00:1f.3 
           chip ID: 8086:a348 
           Device-2: NVIDIA GP104 High Definition Audio vendor: ASUSTeK driver: snd_hda_intel 
           v: kernel bus ID: 01:00.1 chip ID: 10de:10f0 
           Device-3: C-Media CM108 Audio Controller type: USB 
           driver: hid-generic,snd-usb-audio,usbhid bus ID: 1-10:5 chip ID: 0d8c:013c 
           Sound Server: ALSA v: k4.15.0-43-generic 
[...]

PR #5292, opened 2019-09-07
dashboard: When run as root, show URL instead of opening browser

dpandhi-git:
fix in dashboard.go to display dashborad url with root #3558

### DIFF
--- a/cmd/minikube/cmd/dashboard.go
+++ b/cmd/minikube/cmd/dashboard.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"os/user"
 	"regexp"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/dashboard.go
```diff
@@ -128,7 +129,12 @@ var dashboardCmd = &cobra.Command{
 			exit.WithCodeT(exit.Unavailable, "{{.url}} is not accessible: {{.error}}", out.V{"url": url, "error": err})
 		}
 
-		if dashboardURLMode {
+		//check if current user is root
+		user, err := user.Current()
+		if err != nil {
+			exit.WithError("Unable to get current user", err)
+		}
+		if dashboardURLMode && user.Uid == "0" {
```
tstromberg: I think this should be a `||` instead of `&&`, and expect . that this might fail the integration tests as is.

tstromberg:
@minikube-bot OK to test

cmd/minikube/cmd/dashboard.go
dpandhi-git: Thanks @tstromberg for the update. I think I focussed too much on root , will update the commit with ||

tstromberg:
Looks good. Thank you!

PR #5300, opened 2019-09-09
improve problem test coverage

n0npax:


pkg/minikube/problem/problem_test.go
```diff
@@ -17,10 +17,92 @@ limitations under the License.
...
+	var tests = []struct {
+		description string
+		problem     Problem
+		expected    string
+	}{
+		{
+			problem:     Problem{},
+			description: "empty",
+			expected: `
+* Error: [] <no value>
```
tstromberg: "empty" feels like a "change detector" test, do you mind removing it? I consider the output of an empty problem as undefined behavior.

Context: https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html

pkg/minikube/problem/problem_test.go
```diff
@@ -17,10 +17,92 @@ limitations under the License.
...
+* Error: [example] test
+* Suggestion: 
+* Documentation: example.com
+`,
+		},
+		{
+			problem:     Problem{Issues: []int{0, 1}},
+			description: "with 2 issues",
+			expected: `
+* Error: [] <no value>
```
tstromberg: Could these test cases be updated to always have an error associated to them? This way we test real-world behavior, and don't codify the string output of undefined behavior.

n0npax: fixed

pkg/minikube/problem/problem_test.go
n0npax: fixed.

n0npax:
happy to help :)

PR #5304, opened 2019-09-09
Upgrade crio to 1.15.1

afbjorklund:
Drop patch that has now been integrated upstream.

Give up on running crio-config, use pre-generated.


tstromberg:
LGTM. Merge whenever.

PR #5314, opened 2019-09-11
Update NO_PROXY example for Windows

sam-sla:
A wrong IP was used in the example for windows, ending in .1 instead of .0.

### DIFF
--- a/site/content/en/docs/Reference/Networking/proxy.md
+++ b/site/content/en/docs/Reference/Networking/proxy.md
@@ -40,7 +40,7 @@ To make the exported variables permanent, consider adding the declarations to ~/
 ```shell
 set HTTP_PROXY=http://<proxy hostname:port>
 set HTTPS_PROXY=https://<proxy hostname:port>
-set NO_PROXY=localhost,127.0.0.1,10.96.0.0/12,192.168.99.1/24,192.168.39.0/24
+set NO_PROXY=localhost,127.0.0.1,10.96.0.0/12,192.168.99.0/24,192.168.39.0/24
 
 minikube start
 ```


RA489:
@sam-sla  Are you getting any issue with the CLA?

sam-sla:
> 
> 
> @sam-sla Are you getting any issue with the CLA?

Yes... Apparently this is something that does not move fast, frustrating 😠 Maybe we should close this PR, meanwhile someone else can just fix it.

sam-sla:
I signed it

site/content/en/docs/Reference/Networking/proxy.md
```diff
@@ -40,7 +40,7 @@ To make the exported variables permanent, consider adding the declarations to ~/
 ```shell
 set HTTP_PROXY=http://<proxy hostname:port>
 set HTTPS_PROXY=https://<proxy hostname:port>
-set NO_PROXY=localhost,127.0.0.1,10.96.0.0/12,192.168.99.1/24,192.168.39.0/24
```
medyagh: shouldn't that be irrelevant the 0 or 1 because it is /24 ?

medyagh:
> I signed it

yay ! glad you didn't give up ! thankfully this is a one time signing thing and then you won't be bugged again.

medyagh:
@sam-sla thanks for your contribution to make the examples more accurate.

site/content/en/docs/Reference/Networking/proxy.md
sam-sla: Yes I think that is the case, but I thought it would be nicer to align it with the other examples.

Issue #5317, opened 2019-09-11
Sync core service versions for v1.16


tstromberg:
For proper v1.16 support, we need to sync at least the CoreDNS and etcd versions, possibly others between:

https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/constants/constants.go

and:

pkg/minikube/constants/constants.go - `GetKubeadmCachedImages`

PR #5324, opened 2019-09-12
Sync default service versions for kubeadm v1.16

RA489:
Sync default service versions for kubeadm v1.16
fixes https://github.com/kubernetes/minikube/issues/5317


pkg/minikube/constants/constants.go
```diff
@@ -298,8 +298,8 @@ func GetKubeadmCachedImages(imageRepository string, kubernetesVersionStr string)
 			imageRepository + "k8s-dns-kube-dns" + ArchTag(true) + "1.14.13",
 			imageRepository + "k8s-dns-dnsmasq-nanny" + ArchTag(true) + "1.14.13",
 			imageRepository + "k8s-dns-sidecar" + ArchTag(true) + "1.14.13",
-			imageRepository + "etcd" + ArchTag(false) + "3.3.10",
-			imageRepository + "coredns" + ArchTag(false) + "1.3.1",
+			imageRepository + "etcd" + ArchTag(false) + "3.3.15-0",
```
tstromberg: This will need some sort of "if v1_16_plus()` condition, so that we aren't accidentally downloading these newer versions for Kubernetes v1.14 users.

RA489: sure

Issue #5295, opened 2019-09-08
virtualbox: registry: driver not found (and delete doesn't work)


tstromberg:
From #minikube Slack:

```
$ minikube start
😄  minikube v1.3.1 on Darwin 10.14.6
E0908 18:55:41.195263   51385 start.go:723] StartHost: Error loading existing host. Please try running [minikube delete], then run [minikube start] again.: registry: driver not found
```

```
$ minikube delete

💣  Failed to delete cluster: registry: driver not found

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
👉  https://github.com/kubernetes/minikube/issues/new/choose
```

My first thought was that this was xhyve, but it's virtualbox:

```$ minikube start --alsologtostderr -v=8
W0908 19:28:43.892842   52493 root.go:254] Error reading config file at /Users/jonas/.minikube/config/config.json: open /Users/jonas/.minikube/config/config.json: no such file or directory
I0908 19:28:43.894379   52493 notify.go:124] Checking for updates...
I0908 19:28:44.206587   52493 start.go:224] hostinfo: {"hostname":"Jonass-MacBook.local","uptime":1331455,"bootTime":1566632269,"procs":374,"os":"darwin","platform":"darwin","platformFamily":"","platformVersion":"10.14.6","kernelVersion":"18.7.0","virtualizationSystem":"","virtualizationRole":"","hostid":"982f17b3-0252-37fb-9869-88b3b1c77335"}
W0908 19:28:44.206744   52493 start.go:232] gopshost.Virtualization returned error: not implemented yet
😄  minikube v1.3.1 on Darwin 10.14.6
I0908 19:28:44.207409   52493 downloader.go:59] Not caching ISO, using https://storage.googleapis.com/minikube/iso/minikube-v1.3.0.iso
I0908 19:28:44.207544   52493 start.go:922] Saving config:
{
    "MachineConfig": {
        "KeepContext": false,
        "MinikubeISO": "https://storage.googleapis.com/minikube/iso/minikube-v1.3.0.iso",
        "Memory": 2000,
        "CPUs": 2,
        "DiskSize": 20000,
        "VMDriver": "virtualbox",
        "ContainerRuntime": "docker",
        "HyperkitVpnKitSock": "",
        "HyperkitVSockPorts": [],
        "DockerEnv": null,
        "InsecureRegistry": null,
        "RegistryMirror": null,
        "HostOnlyCIDR": "192.168.99.1/24",
[...]

PR #5325, opened 2019-09-12
virtualbox: registry: driver not found (and delete doesn't work)

RA489:
virtualbox: registry: driver not found (and delete doesn't work)
fixes https://github.com/kubernetes/minikube/issues/5295


tstromberg:
Obsoleted by #5384

PR #5346, opened 2019-09-13
WIP: Remove unneccesary test script vars (EXTRA_ARGS, PARALLEL_COUNT, etc)

tstromberg:
Will update this description with any performance differences from the removal of `PARALLEL_COUNT`. 

Normally it's best to let Go scale parallelism by default, but this is a special case, as the hypervisor core management is out of Go's hands.


tstromberg:
@minikube-bot OK to test

PR #5361, opened 2019-09-15
Fix addon config

woodcockjosh:
`constants. AddonsPath` doesn't exist and isn't used by the example efk plugin. `constants.GuestAddonsDir` seems more appropriate. Docs are missing argument for whether or not plugin file is template.


medyagh:
nice catch ! thank you :)

PR #5373, opened 2019-09-16
Download version appropriate driver, add sha256 checksum check

tstromberg:
The driver download currently downloads the driver from the latest official release. 

When one is running a beta version of minikube, this results in constantly downloading an older release of the driver.

Also, while I was here, add sha256 checking, and a lot of logging.


tstromberg:
@minikube-bot OK to test

Issue #5368, opened 2019-09-16
hyperkit integration testing: could not find an IP address / OS type not recognized


tstromberg:
This error has been popping up more frequently:

```
cluster.go:94] Provisioning machine with config: {KeepContext:false EmbedCerts:false MinikubeISO:https://storage.googleapis.com/minikube/iso/minikube-v1.4.0-beta.2.iso Memory:2000 CPUs:2 DiskSize:20000 VMDriver:hyperkit ContainerRuntime:docker HyperkitVpnKitSock: HyperkitVSockPorts:[] DockerEnv:[] InsecureRegistry:[] RegistryMirror:[] HostOnlyCIDR:192.168.99.1/24 HypervVirtualSwitch: KVMNetwork:default KVMQemuURI:qemu:///system KVMGPU:false KVMHidden:false Downloader:{} DockerOpt:[] DisableDriverMounts:false NFSShare:[] NFSSharesRoot:/nfsshares UUID: NoVTXCheck:false DNSProxy:false HostDNSResolver:true}
start.go:838] StartHost: create: Error creating machine: Error in driver during machine creation: IP address never found in dhcp leases file Temporary error: could not find an IP address for 7e:f:4e:c2:cb:dd
cluster.go:98] Skipping create...Using existing machine configuration
cluster.go:117] Machine state:  Running
cluster.go:135] engine options: &{ArbitraryFlags:[] DNS:[] GraphDir: Env:[] Ipv6:false InsecureRegistry:[10.96.0.0/12] Labels:[] LogLevel: StorageDriver: SelinuxEnabled:false TLSVerify:false RegistryMirror:[] InstallURL:https://get.docker.com}
cluster.go:156] configureHost: &{plugin:0xc000800230 heartbeatDoneCh:0xc000040d20 Client:0xc00016f7d0}
cluster.go:178] Configuring auth for driver hyperkit ...
cluster.go:158] configureHost completed within 6m7.361143537s
start.go:838] StartHost: Temporary Error: Error configuring auth on host: OS type not recognized
```

Example: https://storage.googleapis.com/minikube-builds/logs/5354/HyperKit_macOS.txt

I suspect that hyperkit has race conditions when run in parallel. It's difficult to tell at the current level of debugging. It's worth noting that we currently wait up to a minute for the IP to appear, which should be more than enough.

tstromberg:
Still a thing: 

https://storage.googleapis.com/minikube-builds/logs/5251/HyperKit_macOS.txt

Issue #5153, opened 2019-08-21
flake test : TestDocker:   "X Failed to enable container runtime: command failed: sudo systemctl start docker"


medyagh:
as seen in KVM tests:
the error is:
```
       X Failed to enable container runtime: command failed: sudo systemctl start docker
        stdout: 
```
which fails for this code sometimes

```
func (r *Docker) Enable(disOthers bool) error {
	if disOthers {
		if err := disableOthers(r, r.Runner); err != nil {
			glog.Warningf("disableOthers: %v", err)
		}
	}
	return r.Runner.Run("sudo systemctl start docker")
}
```

more detailed log:

```
        LimitNPROC=infinity
        LimitCORE=infinity
        
        # Uncomment TasksMax if your systemd version supports it.
        # Only systemd 226 and above support this version.
        TasksMax=infinity
        TimeoutStartSec=0
        
        # set delegate yes so that systemd does not reset the cgroups of docker containers
        Delegate=yes
        
        # kill only the docker process, not all processes in the cgroup
        KillMode=process
        
        [Install]
        WantedBy=multi-user.target
        " | sudo tee /lib/systemd/system/docker.service
        SSH cmd err, output: <nil>: [Unit]
        Description=Docker Application Container Engine
        Documentation=https://docs.docker.com
        After=network.target  minikube-automount.service docker.socket
        Requires= minikube-automount.service docker.socket 
        
        [Service]
        Type=notify
        
        # DOCKER_RAMDISK disables pivot_root in Docker, using MS_MOVE instead.
        Environment=DOCKER_RAMDISK=yes
        Environment=FOO=BAR
        Environment=BAZ=BAT
        
        
        # This file is a systemd drop-in unit that inherits from the base dockerd configuration.
        # The base configuration already specifies an 'ExecStart=...' command. The first directive
        # here is to clear out that command inherited from the base configuration. Without this,
        # the command from the base configuration and the command specified here are treated as
        # a sequence of commands, which is not the desired behavior, nor is it valid -- systemd
        # will catch this invalid input and refuse to start the service with an error like:
        #  Service has more than one ExecStart= setting, which is only allowed for Type=oneshot services.
        ExecStart=
[...]

PR #5381, opened 2019-09-17
Offset integration tests which start clusters by 5s

tstromberg:
Adds `MaybeSlowParallel`, which offsets parallel test startup by 5 seconds. 

This is used to avoid race conditions in tests which create a cluster. My instinct says that 5 seconds should be enough time to offset sensitive steps, such as manipulating certificates. 

https://github.com/kubernetes/minikube/issues/5368
https://github.com/kubernetes/minikube/issues/5153
https://github.com/kubernetes/minikube/issues/4967

### DIFF
--- a/test/integration/addons_test.go
+++ b/test/integration/addons_test.go
@@ -36,7 +36,7 @@ import (
 
 // TestAddons tests addons that require no special environment -- in parallel
 func TestAddons(t *testing.T) {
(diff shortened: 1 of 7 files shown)


medyagh:
Interesting

tstromberg:
@medyagh - By "interesting", I assume you mean "terribly embarrassing hack".

PR #5403, opened 2019-09-19
Upgrade dashboard to 2.0.0b4

tstromberg:
Required for compatibility with Kubernetes v1.16.

Sadly, there is no non-beta release of the dashboard that works well enough with v1.16 to show the front page.

Tested for backwards compat all the way to v1.11.7 - seems fine.


tstromberg:
This addon now works with "kubectl apply", but need to split file up to make it work with addons.

PR #5405, opened 2019-09-19
Update addon-manager to v9.0.2, improve startup & reconcile latency

tstromberg:
It was tiring to wait 60 seconds while testing addon updates.


tstromberg:
https://storage.googleapis.com/minikube-builds/logs/5405/HyperKit_macOS.txt

PR #5414, opened 2019-09-19
Re-order hypervisor tabs on install page based on best user experience

tstromberg:
Thoughts?


afbjorklund:
I will probably continue with VirtualBox for a while longer, but will consider libvirt when upgrading OS.

As a user I can fully understand why you would want to use the so called "native" virtualization.
As a developer I don't really like targetting 3 APIs, I can see why microk8s came up with Multipass...

But I think this is a good first step, eventually we probably want to phase out VirtualBox completely.

PR #5422, opened 2019-09-20
Improve reportcard

afbjorklund:
* golint
* gocyclo


afbjorklund:
It would be interesting to be able to see the output and current “score” of `golint` and friends, but these should **not** be blocking in the same way as `gofmt` and such...

PR #5425, opened 2019-09-21
improve translations test coverage

n0npax:


pkg/minikube/translate/translate_test.go
```diff
@@ -26,28 +31,190 @@ func TestSetPreferredLanguage(t *testing.T) {
...
+			description: "default polish",
+			langDef:     language.Polish,
+			langEnvVar:  "en_US.UTF-8",
+			expected:    "Setting Language to en",
+		},
+		{
+			description: "preferred polish, no translation",
+			langDef:     language.English,
+			langEnvVar:  "pl_PL.UTF-8",
+			expected:    "Failed to load translation file for pl-PL",
```
tstromberg: This test feels like a change-detector test: https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html

I'd argue the test could probably be removed. The more important thing is to determine success of a task, rather than what error message is emitted if it fails. This also goes along with the idea of "never inspect the output of error.Error":

https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

n0npax: Hi,
I'm not sure if you are talking about lines `148-153` or whole `129-199`.

Assuming `129-199`
I can remove this if you prefer, but I don't see any better option to test this function.
The function is returning nothing and just printing info for user using `glog`.
In theory, we can invoke the function and ensure it won't panic, but it's not useful in my mind.
I can also refactor this test to just check if `preferredLocal` was configured as expected, but It will cover just a single path.

Assuming `148-153`
this is valid use-case in my mind. By default language is always English(https://github.com/kubernetes/minikube/blob/master/pkg/minikube/translate/translate.go#L32) Some people are using national locales, so If I'm using `LC_ALL=pl_PL` I should get the proper error when a translation cannot be loaded and it's tested here.

priyawadhwa: Hey @n0npax, thanks for adding these tests in. For this one specifically (`129-199`), I agree that we probably shouldn't compare error strings. I think it would be fine to either remove it or to just check if `preferredLocal` is configured as expected.

n0npax: Hi guys,
requested changes implemented. Can you review once again please?

priyawadhwa:
Hey @n0npax looks like travis is failing with a linting error:

```
pkg/minikube/translate/translate_test.go:55: File is not `goimports`-ed (goimports)
```

could you run `goimports` on the file? I believe you can test unit tests locally with `make test` once that's been done to make sure it worked.

n0npax:
Hi @priyawadhwa 

I may be wrong but seems like `VirtualBox_Linux` is failing due to some reasons including:
```
error execution phase certs/apiserver-etcd-client: couldn't load CA certificate etcd-ca: failure loading etcd/ca certificate authority: failed to load certificate: the certificate is not valid yet
```
And Windows
```
        * ==> storage-provisioner [58f7068bd656] <==
        * E1012 01:48:02.798373       1 controller.go:682] Error watching for provisioning success, can't provision for claim "default/testpvc": events is forbidden: User "system:serviceaccount:kube-system:storage-provisioner" cannot list resource "events" in API group "" in the namespace "default"

```

Running `goimports` on mentioned file is not causing any change.
```
 n0npax@narrenturm  ~/workspace/minikube/pkg/minikube/translate   improve-translations-test-coverage  (⎈ minikube:default) goimports -w .                
 n0npax@narrenturm  ~/workspace/minikube/pkg/minikube/translate   improve-translations-test-coverage  (⎈ minikube:default) git status 
On branch improve-translations-test-coverage
Your branch is up to date with 'n0npax/improve-translations-test-coverage'.

nothing to commit, working tree clean
```

Can you please point me to the logs where go-imports issue occurs?
I believe issue can connected with #5605

tstromberg:
Thank you!

n0npax:
Thanks!

Issue #3416, opened 2018-12-05
Node InternalIP not matching Host-only address


reductor:
**Environment**:

**Minikube version** v0.30.0
- **OS** Windows 10 64-bit
- **VM Driver**: VirtualBox
- **ISO version**: v0.30.0

- **minikube ip**: 192.168.99.100
- **minikube ssh ifconfig**:
```
docker0   Link encap:Ethernet  HWaddr 02:42:43:CF:20:AF
          inet addr:172.17.0.1  Bcast:172.17.255.255  Mask:255.255.0.0
          UP BROADCAST MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

eth0      Link encap:Ethernet  HWaddr 08:00:27:A4:13:0D
          inet addr:10.0.2.15  Bcast:10.0.2.255  Mask:255.255.255.0
          inet6 addr: fe80::a00:27ff:fea4:130d/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:217809 errors:0 dropped:0 overruns:0 frame:0
          TX packets:14653 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:321672994 (306.7 MiB)  TX bytes:1092520 (1.0 MiB)

eth1      Link encap:Ethernet  HWaddr 08:00:27:84:82:B1
          inet addr:192.168.99.100  Bcast:192.168.99.255  Mask:255.255.255.0
          inet6 addr: fe80::a00:27ff:fe84:82b1/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:953 errors:0 dropped:0 overruns:0 frame:0
          TX packets:912 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:63500 (62.0 KiB)  TX bytes:49816 (48.6 KiB)

lo        Link encap:Local Loopback
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
[...]

PR #5427, opened 2019-09-21
Fix node InternalIP not matching host-only address

woodcockjosh:
This fixes the issue with the node InternalIP not matching the minikube ip


woodcockjosh:
#3416

afbjorklund:
Pretty sure that the config was `kubelet.node-ip` (rather than just "node-ip"), but we will see

/ok-to-test

woodcockjosh:
> Pretty sure that the config was `kubelet.node-ip` (rather than just "node-ip"), but we will see
> 
> /ok-to-test

yeah it is inside the `NewKubeletConfig` function which filters the k8s.ExtraOptions on line 540

afbjorklund:
Ok, makes sense.

tstromberg:
This is a perfect example where a long-standing bug has a simple solution, as the true difficulty is in discovering where it went wrong.

Well done, and Thank You!

PR #5439, opened 2019-09-23
Fix order of parameters to CurrentContext funcs

### LINKED ISSUE #5436: profile <name>: error acquiring lock for x: Name "-x" not valid
Trying Minikube 1.4.0 and seeing this error when setting the profile name:

```
minikube profile operator-minikube
😿  Error while setting kubectl current context :  writing kubeconfig: Error writing file operator-minikube: error acquiring lock for operator-minikube: Name "-operator-minikube" not valid
```

Seems like a bug? I did prefix any `-` into the profile name.

afbjorklund:
Apparently it is easy to get name and path swapped around.

Fixes #5436

Also fixes the issue where "stop" didn't unset the current-context (due to profile name not matching).

### DIFF
--- a/cmd/minikube/cmd/stop.go
+++ b/cmd/minikube/cmd/stop.go
@@ -77,7 +77,7 @@ func runStop(cmd *cobra.Command, args []string) {
 	}
 
 	machineName := pkg_config.GetMachineName()
-	err = kubeconfig.UnsetCurrentContext(constants.KubeconfigPath, machineName)
+	err = kubeconfig.UnsetCurrentContext(machineName, constants.KubeconfigPath)
 	if err != nil {
 		exit.WithError("update config", err)
 	}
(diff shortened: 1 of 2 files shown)


cmd/minikube/cmd/stop.go
```diff
@@ -77,7 +77,7 @@ func runStop(cmd *cobra.Command, args []string) {
 	}
 
 	machineName := pkg_config.GetMachineName()
-	err = kubeconfig.UnsetCurrentContext(constants.KubeconfigPath, machineName)
+	err = kubeconfig.UnsetCurrentContext(machineName, constants.KubeconfigPath)
```
tstromberg: This PR caused me to LOL, but not necessarily in a good way. 

There is a second bug here: `constants.KubeconfigPath` does not respect the `KUBECONFIG` environment variable. If the filename isn't provided, we do the right thing. Here is my suggested change to reduce the number of paths to sadness in our code base:

- Change *SetCurrentContext to either take a second argument, or not take a second argument. Having it support both is a recipe for disaster.
- If you prefer the 1-argument version, update `context_test.go` to set the temp path via an alternative mechanism, such as `KUBECONFIG`
- If you prefer the 2-argument version, update these the `kubeconfig.*Context()` calls to pass  `kubeconfig.PathFromEnv()`.

afbjorklund:
Never got around to updating this PR, but you can of course still improve `SetCurrentContext` more

Issue #4730, opened 2019-07-10
Better documentation needed for virtual desktop


afbjorklund:
More and more users are trying to run minikube on a virtual desktop (VM).
Typically this **fails**, because nested virtualization is not enabled for them...

``` console
vagrant@ubuntu-xenial:~$ egrep --color 'vmx|svm' /proc/cpuinfo
vagrant@ubuntu-xenial:~$ wget -q https://github.com/kubernetes/minikube/releases/download/v1.2.0/minikube_1.2.0.deb
vagrant@ubuntu-xenial:~$ sudo dpkg -i minikube_1.2.0.deb 
(Reading database ... 54233 files and directories currently installed.)
Preparing to unpack minikube_1.2.0.deb ...
Unpacking minikube (1.2.0) ...
Setting up minikube (1.2.0) ...
vagrant@ubuntu-xenial:~$ minikube start
😄  minikube v1.2.0 on linux (amd64)
💿  Downloading Minikube ISO ...
 129.33 MB / 129.33 MB [============================================] 100.00% 0s
🔥  Creating virtualbox VM (CPUs=2, Memory=2048MB, Disk=20000MB) ...
E0711 16:56:33.992057    1778 start.go:559] StartHost: create: precreate: VBoxManage not found. Make sure VirtualBox is installed and VBoxManage is in the path

💣  Unable to start VM
❌  Error:         [VBOX_NOT_FOUND] create: precreate: VBoxManage not found. Make sure VirtualBox is installed and VBoxManage is in the path
💡  Advice:        Install VirtualBox, ensure that VBoxManage is executable and in path, or select an alternative value for --vm-driver
📘  Documentation: https://www.virtualbox.org/wiki/Downloads
⁉️   Related issues:
    ▪ https://github.com/kubernetes/minikube/issues/3784
    ▪ https://github.com/kubernetes/minikube/issues/3776

😿  If the above advice does not help, please let us know: 
👉  https://github.com/kubernetes/minikube/issues/new
vagrant@ubuntu-xenial:~$ sudo apt install -y virtualbox
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following additional packages will be installed:
  binutils cpp cpp-5 dkms fakeroot fontconfig fontconfig-config fonts-dejavu-core gcc gcc-5 libasan2 libasound2 libasound2-data libasyncns0
[...]

PR #5452, opened 2019-09-24
Update roadmap & principles for 2020 (comments & discussion welcome!)

tstromberg:
Q4 is about to begin, so I'm sending this out now so that we can get the comments rolling in well in advance.

We should aim for ratification before KubeCon 2019 (Nov 18).


medyagh:
looks good on first look.

afbjorklund:
As I mentioned elsewhere, I would prefer not treating Docker specific to other runtimes. It is fine to have a default, but I think it should use CRI.

blueelvis:
Looks good! I am really excited about the new features planned!

tstromberg:
Thank you for the feedback! I am particularly curious if anyone has suggestions on:

- Any items that need clarification
- Any items that are not roadmap worthy
- Any important items that are missing from the roadmap

blueelvis:
What does WSL2 support entail in this case? Could you please explain this more?

afbjorklund:
I was thinking we could include running on “generic” virtual machines and even bare metal. As per the issue that I wrote (#4733)

afbjorklund:
Thinking about how we can "expand" these ideas gradually, like how we can move them from principles to roadmap to descriptions. The current roadmap is a bit cryptic, so it would be great if it could gradually refer to a lengthier document where each item has a least a couple of sentences. But I think the length of the current documents (principles/roadmap) is just about right, so keep that!

Maybe it is enough to add a GitHub reference, but those issues tend to be added with history like discussions and revisions (and bots!) so it is harder to keep a generic summary updated there ?

Direct links:

* [Principles](https://github.com/kubernetes/minikube/blob/0836a2d4769fd7d7d2a68ce5bb8b1f10736cf3a0/site/content/en/docs/Concepts/principles.en.md) 0836a2d4769fd7d7d2a68ce5bb8b1f10736cf3a0

* [Roadmap](https://github.com/kubernetes/minikube/blob/0836a2d4769fd7d7d2a68ce5bb8b1f10736cf3a0/site/content/en/docs/Contributing/roadmap.en.md) 0836a2d4769fd7d7d2a68ce5bb8b1f10736cf3a0

Maybe it can be retrofitted with the MEP suggestion ? Or just add yet another markdown file. Maybe it is enough with one sub-heading per roadmap item, and just add some text. Two places to update, but...

site/content/en/docs/Concepts/principles.en.md
```diff
@@ -5,14 +5,13 @@ date: 2019-06-18T15:31:58+08:00
...
-5. High-fidelity
-6. Compatible with all supported Kubernetes releases
-7. Support for all Kubernetes-friendly container runtimes
-8. Stable and easy to debug
+1.  Inclusive and community-driven
+1.  User-friendly
+1.  Support all Kubernetes features
+1.  Cross-platform
+1.  Reliable
+1.  High Performance
```
afbjorklund: I am not sure if "high performance" should be on the principles at all. While I like the solution to be as effecient as possible, I am not sure it is more important than "developer focused" ?

One example: if we replaced `dockerd` with `containerd` it would improve runtime performance (microk8s did this). But it would also make it harder / more cumbersome to use when developing.

site/content/en/docs/Concepts/principles.en.md
```diff
@@ -5,14 +5,13 @@ date: 2019-06-18T15:31:58+08:00
...
-4. Support all Kubernetes features
-5. High-fidelity
-6. Compatible with all supported Kubernetes releases
-7. Support for all Kubernetes-friendly container runtimes
-8. Stable and easy to debug
+1.  Inclusive and community-driven
+1.  User-friendly
+1.  Support all Kubernetes features
+1.  Cross-platform
+1.  Reliable
```
afbjorklund: Do we need this ? It seems to indicate that it would be a stable deployment environment or something. It should of course start without crashing, but as long as it is up during the development session I guess ?

As a comparison, other similar components have writings like "_Boot2Docker is designed and tuned for development. Using it for any kind of production workloads is highly discouraged._" Not so much reliable there.

site/content/en/docs/Concepts/principles.en.md
```diff
@@ -5,14 +5,13 @@ date: 2019-06-18T15:31:58+08:00
 
 The primary goal of minikube is to make it simple to run Kubernetes locally, for day-to-day development workflows and learning purposes. Here are the guiding principles for minikube, in rough priority order:
```
afbjorklund: We're seeing increasing confusion about what "locally" really means. There is a historic context, in where it meant "something running on your laptop". But nowadays we have both the people expecting the control-plane components to run directly on their development host, as well as the people having their desktop environment in a private cloud. This sentence is great for a slogan ("`Run Kubernetes locally`"), but could actually need some expanding elsewhere ?

From #4730

afbjorklund:
>  Graphical User Interface

I hadn't thought about this item in a _long_ time. But when I saw the Kubernetes integration in Visual Studio Code*, I'm starting to think that it is actually quite nice to have, even if it is only "start" and "stop"...

* https://code.visualstudio.com/docs/azure/kubernetes

Started to play with Qt5, and it is lotsa fun! We should definitely at least have a small minikube icon, it doesn't look impossible when looking at https://github.com/therecipe/qt and its small [systray.go](https://github.com/therecipe/qt/blob/master/internal/examples/widgets/systray/systray.go)* ?

* https://doc.qt.io/qt-5/qtwidgets-desktop-systray-example.html

josedonizetti:
@tstromberg I'd like to explore the idea of using minikube for developing Kubernetes itself. Does it sound like a worthy item for the 2020 roadmap? I was actually considering writing an MEP about it.

tstromberg:
@josedonizetti - Sounds great! I'll add it to the roadmap.

@afbjorklund - indeed. Some inspiration for this idea has been seeing IDE integration (Cloud Code), as well as seeing folks using Docker for Desktop. Start/stop/resume/switch context from the menu bar would be killer, and not difficult to implement. There is room for more (addon selection, resource settings), but I would save it for a subsequent iteration.

I haven't explored which frameworks allow for buttons straight from the menubar, but I have noted that there are no less than 3 Go frameworks for menubar/systray based applications.

site/content/en/docs/Concepts/principles.en.md
tstromberg: I feel like "High Performance" should stay, as it provides guidance that we actually do care about performance. That said, we also care about "User-friendly" and "Reliable", so we need to balance principles against one another.

site/content/en/docs/Concepts/principles.en.md
tstromberg: What do you suggest for wording here? 

I believe our primary goal is to run it locally, as in "something on your laptop". Running elsewhere seems secondary.

tstromberg:
Updated: please take another look.

site/content/en/docs/Concepts/principles.en.md
afbjorklund: We should probably should stick with the "run Kubernetes locally". I would like it if it could include having a second computer right next to your usual one, or even something like private cloud - but agree those are not the primary goal.

Nothing wrong with the paragraph in general, though.

site/content/en/docs/Concepts/principles.en.md
afbjorklund: I'm OK with keeping "reliable" (stable) and "high performance" as guiding principles, even if they are not the main goal.

PR #5454, opened 2019-09-24
WIP: Check if viper JSON file exists before parsing

### LINKED ISSUE #4920: Change the error message when Config file is not available

tstromberg:
Fixes #4920

Obsoletes #4940

### DIFF
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -233,10 +233,13 @@ func init() {
 // initConfig reads in config file and ENV variables if set.
 func initConfig() {
 	configPath := constants.ConfigFile
+	_, err := os.Stat(configPath)
+	if os.IsNotExist(err) {
+		return
+	}
 	viper.SetConfigFile(configPath)
 	viper.SetConfigType("json")
-	err := viper.ReadInConfig()
-	if err != nil {
+	if err := viper.ReadInConfig(); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/root.go
```diff
@@ -233,10 +233,13 @@ func init() {
 // initConfig reads in config file and ENV variables if set.
 func initConfig() {
 	configPath := constants.ConfigFile
+	_, err := os.Stat(configPath)
+	if os.IsNotExist(err) {
+		return
```
sharifelgamal: Is it safe to just return silently here?

afbjorklund: In the _other_ PR (#4940), we issued a warning

afbjorklund:
There is an alternative implementation in #4940, that used the viper API instead of checking twice.

PR #5487, opened 2019-09-28
Improve notify test coverage

n0npax:


pkg/minikube/notify/notify_test.go
```diff
@@ -146,39 +152,80 @@ func TestGetLatestVersionFromURLMalformed(t *testing.T) {
...
-
-	MaybePrintUpdateText(server.URL, lastUpdateCheckFilePath)
-	if len(outputBuffer.String()) == 0 {
-		t.Fatalf("Expected MaybePrintUpdateText to output text as the current version is %s and version %s was served from URL but output was [%s]",
-			version.GetVersion(), latestVersionFromURL, outputBuffer.String())
+	viper.Set(config.ReminderWaitPeriodInHours, 24)
+	for _, test := range tc {
+		t.Run(test.description, func(t *testing.T) {
+			viper.Set(config.WantUpdateNotification, test.wantUpdateNotification)
+			lastUpdateCheckFilePath = filepath.Join(tempDir, "last_update_check")
```
josedonizetti: At some point I remember @tstromberg changing calls to `filepath.Join` to use `path.Join` due to problems in windows.

pkg/minikube/notify/notify_test.go
```diff
@@ -146,39 +152,80 @@ func TestGetLatestVersionFromURLMalformed(t *testing.T) {
...
+			latestVersionFromURL := test.latestVersionFromURL
+			handler := &URLHandlerCorrect{
+				releases: []Release{{Name: version.VersionPrefix + latestVersionFromURL}},
+			}
+			server := httptest.NewServer(handler)
+			defer server.Close()
+			if test.url == "" {
+				test.url = server.URL
+			}
+			status := MaybePrintUpdateText(test.url, "/dev/null")
```
josedonizetti: Probably, better to create a tempdir here, and use it's path, because `/dev/null` will be a problem on windows.

pkg/minikube/notify/notify_test.go
n0npax: fixed

pkg/minikube/notify/notify_test.go
n0npax: fixed. Thanks for the review.

pkg/minikube/notify/notify_test.go
```diff
@@ -34,11 +35,17 @@ import (
...
+		t.Fatal("MaybePrintUpdateTextFromGithub() expected to return false for basic setup, bot got true")
+	}
+}
+
 func TestShouldCheckURL(t *testing.T) {
 	tempDir := tests.MakeTempDir()
 	defer os.RemoveAll(tempDir)
 
-	lastUpdateCheckFilePath := filepath.Join(tempDir, "last_update_check")
+	lastUpdateCheckFilePath := path.Join(tempDir, "last_update_check")
```
tstromberg: Why this change? Won't this break on Windows?

tstromberg:
@minikube-bot OK to test

pkg/minikube/notify/notify_test.go
josedonizetti: If that's the case, it is my fault, I thought it was the other way around. Where `filepath` was the problem. Sorry @n0npax 

Thanks @tstromberg

medyagh: this PR looks good after this line is changed back to  `ilepath.Join`

tstromberg: No problem - it is confusing in minikube, but here is the rule of thumb:

- Local filesystem path -> `filepath.Join`
- VM filesystem path -> `path.Join`
- URL: `path.Join`

n0npax: reverted to `filepath` Thanks guys!

n0npax: fixed. Thx!

Issue #5501, opened 2019-09-30
proposal: Remove repetitive addon configurations


woodcockjosh:
There are a lot of repetitive configurations in: 
[addons.go](https://github.com/kubernetes/minikube/blob/668146d10238d7a35e2fe918239cfadb2c413773/pkg/minikube/assets/addons.go)

Instead of providing a list of configurations via`MustBinAsset` we can create a class that is able to generate a list of `MustBinAsset` classes based on the plugin name. We can also include a class that generates the addons so that the map key is always the the same as the plugin name. 
1. If the file ends in .tpl then it is a template
2. The default directory is `vmpath.GuestAddonsDir`
3. The default permissions are `0644`
4. The default plugin directory with the files in it is `deploy/addons/`
5. By default the plugin is disabled. This can be overwritten in the new class to be true.
6. They keyname of the map that its generated is the same as the plugin name. 

This would look something like this: 
```go
var AddOnConfig = new AddOnConfigs();
AddOnConfig.Add(new DefaultAddonConfig("dashboard");
AddOnConfig.Add(new DefaultAddonConfig("default-storageclass", map[string]interface{}{"enabled": true});
// etc, etc, etc
var Addons = AddOnConfig.GetMap();
```

To take this a step further we can eliminate contributors having to create an entry in this file all together if we create an optional file inside each addon directory of `config.json` with a value of
```json
{"enabled": true }
```

tstromberg:
This seems a reasonable approach, but my preference is that we skip this intermediate step and head straight for what you insinuate at the end: make it easier for developers to create addons by eliminating the necessity for them to modify `addons.go` at all.

I have some thoughts around addons (hence why it is coyly called out in the 2020 roadmap), and would love to work with you on a more ambitious proposal. I would like a much richer 3rd party ecosystem, which will require it to be trivial for developers to add/update addons. Basically, something modeled more closely on other open source packaging systems.

- A `minikube addons update` command to pull down recent addon changes
- Support for standard `kubectl apply` YAML files. 
- Support for resource hints (this addon requires 2 CPU's, 4096MB of memory)
- Metadata with description & ownership info

There is certainly more pie-in-the-sky possibilities, like variants and plumbing configuration in, but I'd rather defer to a second iteration.

Thoughts?

woodcockjosh:
[...]

Issue #1701, opened 2017-07-15
ingress addon: "default backend - 404" when HTTPS is used.


veqryn:
**BUG REPORT**

**Minikube version**: v0.20.0

**Environment**:
- **OS**: Windows 10 Pro (Anniversary Edition)
- **VM Driver**: hyperv
- **ISO version**: minikube-v0.20.0.iso

**What happened**:
```
$ kubectl run hello-world --image=tutum/hello-world:latest --port=80
deployment "hello-world" created

$ kubectl expose deployment hello-world --type=NodePort
service "hello-world" exposed

$ curl $(minikube service hello-world --url)
<html>
<head>
        <title>Hello world!</title>
...

$ minikube addons enable ingress
ingress was successfully enabled

$ curl $(minikube ip)
default backend - 404
```

So far so good. Created a file `minikube-ingress.yaml` as so:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: minikube-ingress
  annotations:
spec:
  rules:
  - host: hello.world
    http:
      paths:
      - path: /*
        backend:
          serviceName: hello-world
          servicePort: 80
```

Added my minikube ip to my hosts file: `192.168.0.25    hello.world`
```
$ kubectl apply -f minikube-ingress.yaml
ingress "minikube-ingress" created

$ curl http://hello.world
<html>
<head>
        <title>Hello world!</title>
...

$ curl -k https://hello.world
default backend - 404
```

Ok, so I *should* be able to hit the hello-world service with https.  Supposedly the default configuration of the ingress is SSL Termination, with some auto-generated self-signed certs.
I did try hitting this in my browser as well, where I was asked to accept the cert first.
But no matter what, it still goes to the 'default backend' instead of my hello-world service.

Also tried the following configuration:
```
apiVersion: v1
kind: Secret
metadata:
  name: minikube-ingress-secret
  namespace: default
type: Opaque
data:
  tls.crt: LS0t...
  tls.key: LS0t...

---

apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: minikube-ingress
  annotations:
    ingress.kubernetes.io/rewrite-target: /
    ingress.kubernetes.io/ssl-redirect: "true"
spec:
  tls:
  - secretName: minikube-ingress-secret
  backend:
    serviceName: hello-world
    servicePort: 80
```

The above ingress definition completely ignores whatever I put in "backend" and just sends everything to the "default backend 404".
[...]

PR #5505, opened 2019-09-30
MEP: Add addon host OS hooks

woodcockjosh:
This is to add a set of CLI hooks that would run on the host operating system when an addon is enabled or disabled.

### DIFF
--- a/enhancements/proposed/20190930-addon-host-os-hooks.md
+++ b/enhancements/proposed/20190930-addon-host-os-hooks.md
@@ -20,11 +20,19 @@ In order to provide a richer and more simplified development experience,
 some addons may be better equipped if they can be integrated with capabilities that run on the Host OS
 triggered by lifecycle actions that occur as part of enabling and disabling addons
 
+Addons that can only be applied to a cluster have limited usefulness as those same features can be easily installed 
+with kubectl or helm. If the point of an addon is to extend the functionality of minikube without modifying minikube's 
+core architecture then that extension should apply to the host OS not just the kubernetes cluster itself. Other prominent
+tools used with kubernetes like helm provide hooks on the host OS in order to enable a rich plugin ecosystem.
+
 ### Feature use cases
 - Enabling and starting a background process that monitors ingress resources in a kubernetes cluster and updates host 
   DNS resolver configurations
 - Enabling and starting a background process that monitors the ingress resource in kubernetes cluster and installs 
   SSL certificates for local domains on the host
+- Develop a notification architecture in minikube which can publish events to a pub/sub broker and services
+  started on the host OS as part of a plugin can subscribe to those events and perform additional actions on behalf
+  of the user
 - Start a background process that could notify an IDE of the minikube ips that are running so that the 
 IDE could be used for connecting to services which use the minikube ip address and NodePort to connect to the database
 - Better integrations for tools that currently are wrapping minikube like: 
--- a/enhancements/proposed/20190930-addon-host-os-hooks.md
+++ b/enhancements/proposed/20190930-addon-host-os-hooks.md
@@ -1,0 +1,197 @@
+# Addon Host OS Hooks
+
+First proposed: 2019-09-30
+Authors: Josh Woodcock
+
+## Reviewer Priorities
+
+Please review this proposal with the following priorities:
+
+*   Does this fit with minikube's [principles](https://minikube.sigs.k8s.io/docs/concepts/principles/)?
+*   Are there other approaches to consider?
+*   Could the implementation be made simpler?
+*   Are there usability, reliability, or technical debt concerns?
+
+Please leave the above text in your proposal as instructions to the reader.
+
+## Summary
+
+In order to provide a richer and more simplified development experience, 
+some addons may be better equipped if they can be integrated with capabilities that run on the Host OS
+triggered by lifecycle actions that occur as part of enabling and disabling addons
+
+### Feature use cases
+- Enabling and starting a background process that monitors ingress resources in a kubernetes cluster and updates host 
+  DNS resolver configurations
+- Enabling and starting a background process that monitors the ingress resource in kubernetes cluster and installs 
+  SSL certificates for local domains on the host
+- Start a background process that could notify an IDE of the minikube ips that are running so that the 
+IDE could be used for connecting to services which use the minikube ip address and NodePort to connect to the database
+- Better integrations for tools that currently are wrapping minikube like: 
+  - https://github.com/mrbobbytables/oidckube
+    In this case you have to use this tool to start the minikube instance. What if instead of doing that
+    The tool cool start a background process when a addon is enabled that would enable the same features this tool
+    provides?
+  - https://docs.seldon.io/projects/seldon-core/en/latest/examples/go_example.html
+    This tool has to request the minikube ip from the user. What if it didn't have to 
+    through a addon which could update a seldon-core addon config on the host os?
+  - https://github.com/superbrothers/minikube-ingress-dns/blob/master/README.md
+    This tool could be started and stoped automatically when the addon is enabled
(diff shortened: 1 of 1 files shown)


woodcockjosh:
Implementing this proposal https://github.com/kubernetes/minikube/issues/5501 would make this implementation easier since we would already be loading a config file from the plugins directory.

woodcockjosh:
Can enable addon to install local ssl certificates: https://github.com/kubernetes/minikube/issues/1701

woodcockjosh:
An alternative design to the configuration file would be instead of running an arbitrary command, an exectuable is installed inside the ~/.minikube/plugins/pluginname/pluginname and it is passed arguments for the hook. So for example when `preEnable` hook runs we run `~/.minikube/plugins/pluginname/pluginname pre-enable`. There would have to be a post-post-remove step which deletes the plugin from the plugins directory. and a pre-pre-enable step which copies the executable to the plugins directory.

enhancements/proposed/20190930-addon-host-os-hooks.md
```diff
@@ -0,0 +1,215 @@
...
+If a "preDisable" hook exits with a status code other than 0 then the addon is still disabled with an error message
+printed to the console and logged 
+If a "preDisable" hook exits with a status code other than 0 then the addon is still disabled with an error message
+printed to the console and logged
+
+### Testing Plan 
+Write integration test for an `example` addon which will install an example set of kubernetes configurations and which has all hooks configured.
+The output of each configuration will be tested and should match the expected result based on the Host OS
+
+Include integration test for failed hooks
```
priyawadhwa: This sounds good to me!

enhancements/proposed/20190930-addon-host-os-hooks.md
```diff
@@ -0,0 +1,215 @@
...
+            # (required) A run command is a command to run in order to obtain some information about the Host OS and OS version
+            - getHostOSInfoCommand: uname
+              # (required) A regex pattern to match based on the output from the run command
+              regex: Darwin
+              # (required) The command to run if the regex matches the output of the run command
+              command: echo "I am about to be enabled"
+            - getHostOSInfoCommand: lsb_release -d
+              # (required) A list regex pattern to try to match. Any match will count as a successful match
+              regexAny:
+                - ^.*Ununtu 18.*$
```
priyawadhwa: nit: Ununtu/Ubuntu

enhancements/proposed/20190930-addon-host-os-hooks.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Background service that monitors addon status
+Write a wrapper program which runs as a background service and continuously monitors which addons are enabled or disabled by running
+minikube commands on the cli
+
+#### Cons of doing this: 
+1. Any action that should be taken as hook by said background service and which also must have sudo permission 
+   must always have sudo permission. This decreases user security since you will basically have to give a program 
+   permission to run as an admin in the backgroundwhich means it can literally do anything it wants to do.
+2. The preEnable and preDisable hooks could not exist which means that depending on what the application is doing
```
priyawadhwa: Do you mean that they might not exist in the config?

enhancements/proposed/20190930-addon-host-os-hooks.md
```diff
@@ -0,0 +1,215 @@
...
+it could end up in strange and unintended
+3. The background program will have to be started/enabled/stopped separately from any minikube addons which it might
```
priyawadhwa: I guess if there is a background program running, then stopping it could be part of the postDisable hook?

woodcockjosh: Yes exactly

enhancements/proposed/20190930-addon-host-os-hooks.md
woodcockjosh: I mean a background service rather than a hook, could not intercept an enable or disable command so preEnable and preDisable features could not be developed if a background service is used to respond to the user initiated event of enabling or disabling an addon.

enhancements/proposed/20190930-addon-host-os-hooks.md
```diff
@@ -0,0 +1,215 @@
...
+    host:
+      # (optional) Runs just before the addon is enabled
+      preEnable:
+        # A list of potential commands to be run based on the OS and OS version for unix operating systems
+        # Based on the order, the first regex to match the output of the getHostOSInfoCommand will be run
+        # Only 1 command per hook will be run
+        unix:
+          filteredCommands:
+            # (required) A run command is a command to run in order to obtain some information about the Host OS and OS version
+            - getHostOSInfoCommand: uname
```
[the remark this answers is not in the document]
woodcockjosh: The idea with the hooks is that they are configured by the addon developer and used to extend minikube in a way that the current ecosystem cannot do with kubernetes configurations alone. Having to separately configure a plugin for behavior which could be done automatically would not provide a good user experience as it gives an opportunity to misconfigure something.

enhancements/proposed/20190930-addon-host-os-hooks.md
```diff
@@ -0,0 +1,215 @@
...
+
+* Lifecycle events for addons other than the ones listed above
+* Conditional logic for whether or not to run the addon other than information related to obtaining the Host OS and/or version number
+* Running commands in the background
+* Running commands with administrative privileges eg: sudo without requesting the user to authenticate
+* Changing how a addon is enabled or disabled by default
+
+## Design Details
+
+A configuration file `config.yaml` optionally exists within the `deploy/addons/addonname` directory.
```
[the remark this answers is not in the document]
woodcockjosh: See answer below

enhancements/proposed/20190930-addon-host-os-hooks.md
```diff
@@ -0,0 +1,215 @@
...
+  - https://www.mailgun.com/blog/creating-development-environments-with-kubernetes-devgun
+    This tool could be started and stopped automatically when the addon is enabled
+    rather than building a wrapper for minikube
+  - https://www.telepresence.io/tutorials/minikube-vpn
+    This tool could be enabled as a minikube addon if it can be installed and started on the host OS
+  - https://developers.facebook.com/docs/whatsapp/installation/dev-multiconnect-minikube/
+    Instead of asking the developer to provide the minikube ip what if what's app could just identify it
+    automatically. What if I had a whats app business API addon that could be installed and 
+    seemlessly integrated with a host os hook?
+- A desktop UI could be developed for minikube which could be enabled and installed as an addon which would show
```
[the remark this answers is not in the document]
woodcockjosh: No. It would be more akin to kitematic I believe. Dashboard run in a web browser and only reports information on a single kubernetes instance

enhancements/proposed/20190930-addon-host-os-hooks.md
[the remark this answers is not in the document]
woodcockjosh: I don't mind a different format for the config file as long as we can achieve the goal. My proposal would allow the plugin developer flexibility for doing different things on different types of hosts which could also be done in a script outside of minikube. I believe your proposal is simpler and better.

tstromberg: Do you mind describing in your proposal getHostOSInfoCommand? It seems like just using runtime.GOOS should be enough for most circumstances. if it isn't, I feel like we should look into plumbing distro information in rather than relying on each addon to implement distro detection, but perhaps not for an initial iteration.

tstromberg:
Do you mind adding a proof-of-concept YAML file for one of the use cases you mention? I'm having a difficult time deciding whether or not a YAML-based approach is preferential to allowing addons to have their own Go executable.

tstromberg:
Closing until this proposal can be refreshed and looked at again. 

I suspect we may need to adjust the MEP workflow to allow for faster iteration and approval.

PR #5508, opened 2019-10-01
Adjusted Terminal Style Detection

### LINKED ISSUE #5506: Terminal Style Support Not Properly Detected
**Short Version**:
* minikube uses `$TERM` to determine if the terminal supports color, this is insufficient. `$COLORTERM` should also be checked. 
* `pkg/minikube/out/out.go` should be changed on/around line 163 to correct this issue.

```
echo $TERM
minikube start
#CTRL+C
TERM=xterm-256color
minikube start
```

```
$> echo $TERM
xterm-termite
$> minikube start
* minikube v1.4.0 on Debian parrot
^C
$> TERM=xterm-256color
$> minikube start
😄  minikube v1.4.0 on Debian parrot
^C
$> env | grep COLOR
COLORTERM=truecolor

```
</details>

```
minikube version: v1.4.0
commit: 7969c25a98a018b94ea87d949350f3271e9d64b6
```
**Suggested Change**:

/pkg/minikube/out/out.go Line 163
```
     term := os.Getenv("TERM")
     colorTerm := os.Getenv("COLORTERM")
     // Example: term-256color
     if !strings.Contains(term, "color") && !strings.Contains(colorTerm, "truecolor") && !strings.Contains(colorTerm, "24bit") && !strings.Contains(colorTerm, "yes") {
         glog.Infof("TERM=%s,COLORTERM=%s, which probably does not support color", term,colorTerm)
         return false
     }

### PR DESCRIPTION
**Pull request by pucheGit**
Fixes #5506  Added check for COLORTERM environment variable.

### DIFF
--- a/pkg/minikube/out/out.go
+++ b/pkg/minikube/out/out.go
@@ -164,7 +164,7 @@ func wantsColor(fd uintptr) bool {
 	colorTerm := os.Getenv("COLORTERM")
 	// Example: term-256color
 	if !strings.Contains(term, "color") && !strings.Contains(colorTerm, "truecolor") && !strings.Contains(colorTerm, "24bit") && !strings.Contains(colorTerm, "yes") {
-		glog.Infof("TERM=%s,COLORTERM=%s, which probably does not support color", term,colorTerm)
+		glog.Infof("TERM=%s,COLORTERM=%s, which probably does not support color", term, colorTerm)
 		return false
 	}
 
--- a/pkg/minikube/out/out.go
+++ b/pkg/minikube/out/out.go
@@ -161,9 +161,10 @@ func wantsColor(fd uintptr) bool {
 	}
 
 	term := os.Getenv("TERM")
+	colorTerm := os.Getenv("COLORTERM")
 	// Example: term-256color
-	if !strings.Contains(term, "color") {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: pkg/minikube/out/out.go [hunk 1]
```diff
@@ -161,9 +161,10 @@ func wantsColor(fd uintptr) bool {
 	}
 
 	term := os.Getenv("TERM")
+	colorTerm := os.Getenv("COLORTERM")
 	// Example: term-256color
-	if !strings.Contains(term, "color") {
-		glog.Infof("TERM=%s, which probably does not support color", term)
+	if !strings.Contains(term, "color") && !strings.Contains(colorTerm, "truecolor") && !strings.Contains(colorTerm, "24bit") && !strings.Contains(colorTerm, "yes") {
```
josedonizetti: Thank you for the PR. Is there any documentation we can link here on why the values we are testing are only those 4? 
Those names seem a bit magic to me.

tstromberg: Some context behind COLORTERM, because it was a new discovery for me as well:

https://gist.github.com/XVilka/8346728

tstromberg: Expected values:

https://www.reddit.com/r/archlinux/comments/5hun36/expected_values_for_term_colorterm/

Issue #5494, opened 2019-09-29
proposal: Enable host OS dns resolution of cluster ingress host names


woodcockjosh:
## Background
Kubernetes offers the rich ability to run multiple instances of the same service in the the cluster. For example I can have 2 redis database instances with different service names. Currently there are 2 ways of accessing these services from the host machine when running on minikube: 
1. If the service has a NodePort can use a server's NodePort capability using `minikube service redis-1` then I can access the service via the minikube ip and service's NodePort
2. If the service is a LoadBalancer I can use `minikube tunnel` then access the service via `redis-1.default.svc.cluster.local`

3. If however my services are ClusterIP I can only access these services via an ingress. Thankfully minikube has an ingress plugin which I can enable via `minikube addons enable ingress`. This works great with 1 additional step. I have to add the dns entry configured for the ingress to my hosts `/etc/hosts` file. This is potentially not such a big deal if it is a small number of entries. If however I have a larger number of hosts like 20 for example then it starts to become problematic. This becomes especially more problematic if I have to do either of the 2 things: 

    1. Delete and re-create my minikube instance which will issue a new ip for the cluster. This can be potentially solved with the resolution of this issue: https://github.com/kubernetes/minikube/issues/951

    2. If the hostnames for my cluster are updated or added. If I am a sole developer of a project updating or adding a host name for the cluster with a predictable ip address is somewhat trivial. However if the number of developers is more than a handful then keeping track of and reminding developers to update their hosts is fairly annoying and slows development efforts. 

## Why does hostname resolution especially matter on minikube?
Some capabilities in k8s can only be tested when using a domain name. An example would be if I am using a router like istio or traefik that matches inbound requests to a particular domain name. 

When running a development cluster that has some type of server, I can use [external dns](https://github.com/kubernetes-incubator/external-dns) to automatically configure my DNS to point to the right server. However, when I am using minikube I can only resolve host names locally since the cluster ip is only available to my Host OS. 

## Possible solutions to host ingress dns resolution
[...]

PR #5511, opened 2019-10-01
MEP: Background service to update host dns resolution for ingress resources

woodcockjosh:
This is to create a background service that runs on the host OS and automatically updates resolver configurations based on host names in ingress resources and minikube ip addresses

If this is implemented it will enable a developer to access `redis1.myproject1.test` on the right minikube instance regardless of what the minikube ip is and their only configuration is the ingress resource that is installed in the cluster.

### DIFF
--- a/enhancements/proposed/20191001-background-service-to-update-dns-resolvers-on-host-os-for-ingress-resources.md
+++ b/enhancements/proposed/20191001-background-service-to-update-dns-resolvers-on-host-os-for-ingress-resources.md
@@ -17,7 +17,7 @@ Please leave the above text in your proposal as instructions to the reader.
 ## Summary
 
 Some capabilities in k8s can only be tested when using a domain name. An example would be if a developer is using a router like 
-istio or traefik that matches inbound requests to a particular domain name. When running a development cluster that has 
+Istio or Traefik that matches inbound requests to a particular domain name. When running a development cluster that has 
 some type of server, they can use [external dns](https://github.com/kubernetes-incubator/external-dns) to automatically 
 configure their DNS to point to the right ip address. However, when they are using minikube they can only resolve host names locally 
 since the cluster ip is only available to their Host OS.
@@ -28,7 +28,7 @@ the cluster.
 
 ## Goals
 
-* Enable ingress host names configured in any minikube cluster to be automatically resolved to the right mininkube ip
+* Enable ingress host names configured in any minikube cluster to be automatically resolved to the right minikube ip
 * After a user installs an ingress in minikube they can access the hostname for the ingress on their host web browser
   or command line with no additional steps
 * This capability can be enabled with a maximum of 2 steps and never repeated even while new minikube instances are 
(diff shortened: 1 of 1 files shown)


woodcockjosh:
Partially resolves https://github.com/kubernetes/minikube/issues/5494

medyagh:
@minikube-bot OK to test

enhancements/proposed/20191001-background-service-to-update-dns-resolvers-on-host-os-for-ingress-resources.md
```diff
@@ -0,0 +1,174 @@
...
+configure their DNS to point to the right ip address. However, when they are using minikube they can only resolve host names locally 
+since the cluster ip is only available to their Host OS.
+
+If this proposal is implemented it will enable a developer to access `redis1.myproject1.test` on the right minikube
+instance regardless of what the minikube ip is and their only configuration is the ingress resource that is installed in
+the cluster.
+
+## Goals
+
+* Enable ingress host names configured in any minikube cluster to be automatically resolved to the right mininkube ip
```
tstromberg: Spelling

enhancements/proposed/20191001-background-service-to-update-dns-resolvers-on-host-os-for-ingress-resources.md
```diff
@@ -0,0 +1,174 @@
...
+## Alternatives Considered
+
+### Ways to access services
+* Don't use ingress at all and just use NodePort then match the DNS name using one of the alternatives described below
+  * Some services won't have a node port available. Especially if they are in a helm chart
+  * Now I have to remember the NodePort every time I want to access my service which maybe I can commit that 5 digit 
+    number to my memory for every service I need to access. If I don't have a good memory then this is really terrible
+    for me as I have to keep a reference sheet of where everything is running or I have to look it up with `kubectl`
+  * If the static ip address is somehow configured then I only have to create the DNS entries one time
+* Use `minikube tunnel` and an ingress dns resolver like CoreDNS then add a resolver file for the internal host IP
```
tstromberg: It's worth noting for this proposal that "tunnel" does update the host resolver on macOS.

We've been thinking more and more about background services, as we've been hitting limits with what is possible without one. This is especially true when we start to think about the minikube UI project. Here's some things we would also like to do with a daemon (outside of the scope of this proposal)

- Resource alerts ("kubectl get node" is noticing bad things)
- Time synchronization for hyperkit

Perhaps we should think about having a `minikubed` with this as one of the first features?

woodcockjosh: `minikubed` sounds cool. How would we install the service? Like what would the command be?

tstromberg:
Closing until this proposal can be refreshed and looked at again. 

I suspect we may need to adjust the MEP workflow to allow for faster iteration and approval.

PR #5512, opened 2019-10-01
Improve machine test coverage

n0npax:
To keep PR small It's improving coverage just for a single file.
I'll provide improvements for remaining files later.


pkg/minikube/machine/cache_binaries_test.go
```diff
@@ -0,0 +1,215 @@
...
+		t.Fatalf("error during creating tmp dir: %v", err)
+	}
+	defer os.RemoveAll(noWritePermDir)
+	err = os.Chmod(noWritePermDir, 0000)
+	if err != nil {
+		t.Fatalf("error (%v) during changing permissions of dir %v", err, noWritePermDir)
+	}
+	noPermsDir := "/etc"
+	if runtime.GOOS == "windows" {
+		noPermsDir = "C:\\Windows\\System32"
```
n0npax: I don't have access to any windows machine, so this should work on windows, but It wasn't tested.

pkg/minikube/machine/cache_binaries_test.go
```diff
@@ -0,0 +1,215 @@
...
+	noWritePermDir, err := ioutil.TempDir("/tmp", "")
+	if err != nil {
+		t.Fatalf("error during creating tmp dir: %v", err)
+	}
+	defer os.RemoveAll(noWritePermDir)
+	err = os.Chmod(noWritePermDir, 0000)
+	if err != nil {
+		t.Fatalf("error (%v) during changing permissions of dir %v", err, noWritePermDir)
+	}
+	noPermsDir := "/etc"
```
medyagh: could you add a comment what is NoPermsDir and why ?

n0npax: comment added (https://github.com/kubernetes/minikube/pull/5512/files#diff-ad5f1899d0f1fcc73de6ef1008d741a9R131)

tstromberg: There is no guarantee that the user running the test does not have write access to /etc.

I would either remove this check, or create another directory using `ioutil.TempDir` and set the permissions appropriately for it.

n0npax: removed

n0npax:
`noPermsDir` removed from test. Thanks for previous review.
Can you please re-review?

tstromberg:
Looks good. Thank you!

Issue #5518, opened 2019-10-01
Patching ingress deployment causes pods to never be scheduled


woodcockjosh:
The issue is explained in detail here:
https://github.com/kubernetes/ingress-nginx/issues/4626

The resolution should be to update the deployment rollout settings in the ingress plugin

PR #5519, opened 2019-10-01
Fix pods not being scheduled when ingress deployment is patched

woodcockjosh:
This issue resolves https://github.com/kubernetes/minikube/issues/5518
Also update nginx-ingress-controller to 0.26.1 which resolves some other inconsistency issues


woodcockjosh:
has same change as https://github.com/kubernetes/minikube/pull/5514

medyagh:
@minikube-bot Ok to test.

looks good ! thank you for improving minikube

tstromberg:
Thank you!

PR #5527, opened 2019-10-02
Refactor the code in utils file

abejide001:
Improve the code in utils test file
Improve the code in utils file


abejide001:
> /assign @tstromberg

pkg/util/utils.go
```diff
@@ -180,23 +176,23 @@ func TeePrefix(prefix string, r io.Reader, w io.Writer, logger func(format strin
 
 // ReplaceChars returns a copy of the src slice with each string modified by the replacer
 func ReplaceChars(src []string, replacer *strings.Replacer) []string {
-	ret := make([]string, len(src))
+	returnSlice := make([]string, len(src))
```
tstromberg: Just `ret` is fine: `returnSlice` is not necessarily more readable. 

For variable naming, omit words that are obvious given a variable's type declaration or otherwise clear from the surrounding context.

For go specifically, the length of the variable name should be proportional to it's scope: if the variable scope is only 1 or 2 lines, it should only be 1 or 2 characters. Global constants should have long names as they will require context whenever referenced.

pkg/util/utils_test.go
```diff
@@ -103,10 +102,9 @@ func TestReplaceChars(t *testing.T) {
 
 func TestConcatStrings(t *testing.T) {
 	testData := []struct {
-		src         []string
-		prefix      string
-		postfix     string
-		expectedRes []string
+		src             []string
+		prefix, postfix string
```
tstromberg: I feel like this hurts readability exactly in the place it's necessary the most: tests. Having one doubled up field in the struct makes the test data difficult to scan.

pkg/util/utils.go
```diff
@@ -34,14 +34,10 @@ import (
...
 
-// ErrPrefix notes an error
-const ErrPrefix = "! "
-
-// OutPrefix notes output
-const OutPrefix = "> "
-
 const (
 	downloadURL = "https://storage.googleapis.com/minikube/releases/%s/minikube-%s-amd64%s"
+	OutPrefix   = "> " // OutPrefix notes output
```
tstromberg: For readability, please move the comment to appear on it's own line before the variable.

pkg/util/utils.go
abejide001: alright, fixed

pkg/util/utils_test.go
abejide001: fixed

pkg/util/utils.go
abejide001: fixed

abejide001: alright, fixed.

pkg/util/utils.go
abejide001: fixed.

pkg/util/utils_test.go
abejide001: fixed

tstromberg:
Thank you for this PR!

With a heavy heart, I'm going to close this PR. I don't think it contributes enough to readability to make it worth moving forward with it. I'd be happy to discuss this further on Slack if you would like. 

Thank you again,
Thomas.

PR #5544, opened 2019-10-05
Get rid of python scripts in build process

### LINKED ISSUE #5461: Get rid of python scripts in build process
We have a python script that checks if there is license boilerplate in every file in the repo.
that was initially copied from kuberntes project. 

https://github.com/kubernetes/minikube/blob/master/hack/boilerplate/boilerplate.py

we need to get rid of it, because it would be nicer to have everything in go.
also reduce the dependency installation during each build / PR.  currently we have to spin up a python vm in travis just for this script .

https://github.com/kubernetes/minikube/blob/master/.travis.yml#L10

it would be a good -first issue to translate this python script to a a go run script.

duohedron:
fixes #5461 Get rid of python scripts in build process


duohedron:
I signed it

medyagh:
@duohedron thank for this PR ! this sounds exciting,
please make sure to sign the CLA https://go.k8s.io/bot-commands?repo=kubernetes%2Fminikube

hack/boilerplate/boilerplate.go
```diff
@@ -0,0 +1,152 @@
...
+*/
+func getFileExtension(filename string) string {
+	splitted := strings.Split(filepath.Base(filename), ".")
+	return strings.ToLower(splitted[len(splitted)-1])
+}
+
+/**
+Function to get all the files from the directory that heeds to be checked.
+*/
+func getFileList(rootDir string, extensions map[string][]byte) []string {
```
medyagh: in golang we dont use get in naming,
consider a better name, like "filesToCheck(..."

duohedron: Thanks for the review. I renamed the function.

medyagh:
@duohedron  please rename the funcs starting with get... and also add this script to the `make lint-ci` so that will check for linting in this script also

hack/boilerplate/boilerplate.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// if our test file is smaller than the reference it surely fails!
+	if len(data) < len(ref) {
+		return false
+	}
+
+	data = data[:len(ref)]
+
+	// Search for "Copyright YEAR" which exists in the boilerplate, but shouldn't in the real thing
+	re = regexp.MustCompile(`Copyright YEAR`)
```
tstromberg: This check should be only performed once when loading boilerplate text - in the function currently called `boilerplateRefs`.

hack/boilerplate/boilerplate.go
```diff
@@ -0,0 +1,158 @@
...
+
+var (
+	skippedPaths   = regexp.MustCompile(`Godeps|third_party|_gopath|_output|\.git|cluster/env.sh|vendor|test/e2e/generated/bindata.go|site/themes/docsy`)
+	rootdir        *string
+	boilerplatedir *string
+)
+
+func init() {
+	cwd, _ := os.Getwd()
+	boilerplatedir = flag.String("boilerplate-dir", cwd, "Boilerplate directory for boilerplate files")
```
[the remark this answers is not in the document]
duohedron: I replaced it with "."
I didn't want to use empty value, since the original script used default values as well, and I didn;t want to change the behavior too much

hack/boilerplate/boilerplate.go
```diff
@@ -0,0 +1,158 @@
...
+	for _, file := range files {
+		if !filePasses(file, refs[fileExtension(file)]) {
+			fmt.Println(file)
+		}
+	}
+
+}
+
+/*
+This function is to populate the refs variable with the
```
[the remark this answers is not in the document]
duohedron: Replaced the function name and the comment

hack/boilerplate/boilerplate.go
duohedron: We're not checking the boilerplate here, but check every file if it contains the template YEAR instead of the actual number

hack/boilerplate/boilerplate.go
```diff
@@ -0,0 +1,158 @@
...
+Function to check whether the processed file
+is valid.
+Returning false means that the file does not the
+proper boilerplate template
+*/
+func filePasses(filename string, ref []byte) bool {
+	var re *regexp.Regexp
+	data, err := ioutil.ReadFile(filename)
+	if err != nil {
+		log.Fatal(err)
```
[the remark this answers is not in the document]
duohedron: I replaced log.Fatal from the functions. Should I remove it from main as well?

hack/boilerplate/boilerplate.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+}
+
+// extensionToBoilerplate returns a map of file extension to required boilerplate text.
+func extensionToBoilerplate(dir string) (map[string][]byte, error) {
+	refs := make(map[string][]byte)
+	files, _ := filepath.Glob(dir + "/*.txt")
+	for _, filename := range files {
+		re := regexp.MustCompile(`\.txt`)
```
tstromberg: Rather than compiling this within a for loop, it should be made a top-level `var`, like you did with `skippedPaths`

hack/boilerplate/boilerplate.go
```diff
@@ -0,0 +1,166 @@
...
+		extension := strings.ToLower(filepath.Ext(re.ReplaceAllString(filename, "")))
+		data, err := ioutil.ReadFile(filename)
+		if err != nil {
+			return nil, err
+		}
+		re = regexp.MustCompile(`\r`)
```
tstromberg: Same with this.

duohedron: Moved regex to top-level

tstromberg:
Thank you!

PR #5548, opened 2019-10-06
Added flags to purge configuration with minikube delete

### LINKED ISSUE #5358: delete should be able to delete the `.minikube` folder (w/ a flag)
From https://github.com/kubernetes/minikube/pull/4980#issuecomment-531568673

`minikube delete` should have a switch/flag to delete the `.minikube` folder as well. This should ask for confirmation and display a warning before proceeding ahead.

/assign

blueelvis:
Fixes #5358


cmd/minikube/cmd/delete.go
```diff
@@ -107,6 +122,30 @@ func runDelete(cmd *cobra.Command, args []string) {
...
 	}
+
+	// Delete the .minikube folder if the flags are set
+	if viper.GetBool(purge) {
+		glog.Infof("Purging the '.minikube' directory located at %s", localpath.MiniPath())
+
+		if viper.GetBool(noPrompt) {
+			glog.Infof("Will not prompt for deletion.")
+		} else {
+			out.T(out.Check, "Are you sure you want to delete the directory located at {{.minikubePath}}? This will delete all your configuration data related to minikube. (Y/N)", out.V{"minikubePath": localpath.MiniPath()})
```
medyagh: maybe check if there are more than 1 profile (default profile) and if there are, tell the user

"this will also delete the configurations for X other profiles "

cmd/minikube/cmd/delete.go
```diff
@@ -51,6 +57,15 @@ associated files.`,
 	Run: runDelete,
 }
 
+func init() {
+	deleteCmd.Flags().Bool(purge, false, "Set this flag to delete the '.minikube' folder from your user directory. This will prompt for confirmation.")
+	deleteCmd.Flags().Bool(noPrompt, false, "Set this flag so that there are no prompts.")
```
tstromberg: Do you mind naming this flag to `--interactive` for consistency with `minikube start`?

cmd/minikube/cmd/delete.go
```diff
@@ -107,6 +122,30 @@ func runDelete(cmd *cobra.Command, args []string) {
 	if err := cmdcfg.Unset(pkg_config.MachineProfile); err != nil {
 		exit.WithError("unset minikube profile", err)
 	}
+
+	// Delete the .minikube folder if the flags are set
+	if viper.GetBool(purge) {
+		glog.Infof("Purging the '.minikube' directory located at %s", localpath.MiniPath())
+
+		if viper.GetBool(noPrompt) {
+			glog.Infof("Will not prompt for deletion.")
```
tstromberg: This logging is probably not necessary.

cmd/minikube/cmd/delete.go
```diff
@@ -107,6 +122,30 @@ func runDelete(cmd *cobra.Command, args []string) {
 	if err := cmdcfg.Unset(pkg_config.MachineProfile); err != nil {
 		exit.WithError("unset minikube profile", err)
 	}
+
+	// Delete the .minikube folder if the flags are set
+	if viper.GetBool(purge) {
```
tstromberg: If a user specifies `--purge`, I don't see why we should ask them again? 

My preference in behavior would be:

- If `--purge` is specified, and this is the only profile, just do what the user wanted.
- If `--purge` is specified, and there are more than one minikube profile, error out unless the user specifies `--force`.

Thoughts?

blueelvis: This is fixed with the latest commit. And yes, I agree with what you mentioned! Now, if there is more than 1 profile, `--purge` will check for it and exit if `--all` flag is not specified. I decided to remove the `--force` flag. The combination of `--all --purge` should be enough. What do you think?

cmd/minikube/cmd/delete.go
blueelvis: As per the other feedback, `--interactive` has been removed and instead a combination of switches is now required.

cmd/minikube/cmd/delete.go
blueelvis: Done. Thanks Medya!

PR #5550, opened 2019-10-06
Address some go9p issues adressed by SonarCloud

afbjorklund:
Some "major" issues shown in https://sonarcloud.io/project/issues?id=blueelvis_minikube&languages=go&resolved=false&types=BUG

### DIFF
--- a/third_party/go9p/unpack.go
+++ b/third_party/go9p/unpack.go
@@ -218,7 +218,7 @@ func Unpack(buf []byte, dotu bool) (fc *Fcall, err error, fcsz int) {
 		goto szerror
 	}
 
-	return
+	return //NOSONAR
 
 szerror:
 	return nil, &Error{"invalid size", EINVAL}, 0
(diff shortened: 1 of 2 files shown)


third_party/go9p/unpack.go
```diff
@@ -218,7 +218,7 @@ func Unpack(buf []byte, dotu bool) (fc *Fcall, err error, fcsz int) {
 		goto szerror
 	}
 
-	return
+	return //NOSONAR
 
 szerror:
 	return nil, &Error{"invalid size", EINVAL}, 0
```
blueelvis: Instead of using `NOSONAR`, could you please refactor the code so that it returns the error inside the `szerror` label and remove the usage of Goto completely? I think that would be much better.

afbjorklund: The check is buggy, since the use of `goto` makes it entirely possible to have two exit points.

tstromberg:
@minikube-bot OK to test

afbjorklund:
This was for #5577

PR #5555, opened 2019-10-07
Fix addons list output format option

woodcockjosh:
Fixes option to set the output format for listing addons

### DIFF
--- a/cmd/minikube/cmd/config/addons_list.go
+++ b/cmd/minikube/cmd/config/addons_list.go
@@ -52,7 +52,7 @@ var addonsListCmd = &cobra.Command{
 }
 
 func init() {
-	AddonsCmd.Flags().StringVar(&addonListFormat, "format", defaultAddonListFormat,
+	addonsListCmd.Flags().StringVar(&addonListFormat, "format", defaultAddonListFormat,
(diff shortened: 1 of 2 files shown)


medyagh:
I am not familiar with what this PR is trying to solve, do we have an issue for this ?  or was it side effect of a recently merged PR ?

woodcockjosh:
> I am not familiar with what this PR is trying to solve, do we have an issue for this ? or was it side effect of a recently merged PR ?

@medyagh Apparently this feature doesn't work at all. I didn't open an issue because it was trivial. However, there is a question about whether or not it should be implemented seeing as it doesn't work for anyone. Maybe instead we should implement a `-o json` like I did with the profile list rather than allowing user to specify a custom format.

tstromberg:
Since this may be your first integration test addition, I thought I'd point you to some useful docs for iterating quickly on integration tests:

https://github.com/kubernetes/minikube/blob/master/test/integration/README.md

woodcockjosh:
@tstromberg I will have time to try to add a test tomorrow

woodcockjosh:
@medyagh this is complete

sharifelgamal:
@minikube-bot OK to test

tstromberg:
Thank you!

Issue #5558, opened 2019-10-07
virtualbox integration tests clean up not working


medyagh:
while debugging integration failures in I noticed the clean up is not doing this job and many left over vms are listed. it appears to be side effect of side effect of this PR: https://github.com/kubernetes/minikube/pull/5378

```
13:21:31 >> VirtualBox VM list after clean up (should be empty):
13:21:31 "<inaccessible>" {dd99bb34-4d8f-41d7-9127-a7915e496555}
13:21:31 "<inaccessible>" {5d3a9e76-f062-48d2-b30c-d4e0d04c6d64}
13:21:31 "<inaccessible>" {a0368b23-736c-4e68-9e5b-3e8eed3229dd}
13:21:31 "<inaccessible>" {cea3b20a-d04a-4d5c-9fa5-857c14a49c57}
13:21:31 "<inaccessible>" {1c706be5-7cc4-4c92-8bfd-05ba8331f3bc}
13:21:31 "<inaccessible>" {60407b52-f093-4814-ba82-1cb3a448afd6}
13:21:31 "<inaccessible>" {377b7887-921a-4393-9ac9-db66fdd21fb2}
13:21:31 "<inaccessible>" {28119044-3d54-4b09-bdf3-07c335064075}
```

medyagh:
```
13:45:19 >> Starting at Mon Oct  7 20:45:19 UTC 2019
13:45:19 
13:45:19 arch:      linux-amd64
13:45:19 build:     5559
13:45:19 driver:    virtualbox
13:45:19 job:       VirtualBox_Linux
13:45:19 test home: /home/jenkins/minikube-integration/linux-amd64-virtualbox-5559-4294-ea1c273987120718817b58dac427d6b2bf83331e
13:45:19 sudo:      
13:45:19 kernel:    #1 SMP Debian 4.9.168-1+deb9u4 (2019-07-19)
13:45:19 kubectl:   Client Version: v1.15.1
13:45:20 docker:    18.06.1-ce
13:45:20 vbox:      5.2.32r132073
13:45:20 
13:45:20 
13:45:20 >> Downloading test inputs from 5559 ...
13:45:23 CommandException: No URLs matched: gs://minikube-builds/5559/testdata/*
13:45:23 CommandException: 1 file/object could not be transferred.
13:45:23 Build step 'Execute shell' marked build as failure
13:45:23 [Google Cloud Storage Plugin] Uploading: VirtualBox_Linux.txt
[...]

PR #5559, opened 2019-10-07
clean up inaccessible virtualbox

medyagh:
closes https://github.com/kubernetes/minikube/issues/5558
also closes https://github.com/kubernetes/minikube/issues/5561

### DIFF
--- a/hack/jenkins/common.sh
+++ b/hack/jenkins/common.sh
@@ -254,7 +254,7 @@ cleanup_stale_routes || true
 
 ${SUDO_PREFIX} rm -Rf "${MINIKUBE_HOME}" || true
 ${SUDO_PREFIX} rm -f "${KUBECONFIG}" || true
-rmdir "${TEST_HOME}"
+${SUDO_PREFIX} rmdir "${TEST_HOME}" || true
 echo ">> ${TEST_HOME} completed at $(date)"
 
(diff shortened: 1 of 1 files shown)


hack/jenkins/common.sh
```diff
@@ -133,8 +133,15 @@ if type -P virsh; then
...
   vboxmanage list vms || true
+  # remove inaccessible stale VMs https://github.com/kubernetes/minikube/issues/4872
+  vboxmanage list vms \	
+    | grep inaccessible \	
+    | cut -d'"' -f3 \	
+    | xargs -I {} sh -c "vboxmanage startvm {} --type emergencystop; vboxmanage unregistervm {} --delete" \	
+    || true
   vboxmanage list vms \
     | egrep -o '{.*?}' \
     | xargs -I {} sh -c "vboxmanage startvm {} --type emergencystop; vboxmanage unregistervm {} --delete" \
```
tstromberg: Rather than adding a second more-specific command, let's fix the general command that comes afterwards.  For clarity, I recommend splitting it into two commands:

```
vboxmanage list vms | egrep -o '{.*?}' | xargs -I{} vboxmanage startvm {} --type emergencystop || true
vboxmanage list vms | egrep -o '{.*?}' | xargs -I{} vboxmanage unregistervm {} | true
```

I tested this on macOS and it seems to do the right thing.

PR #5563, opened 2019-10-07
Make error message more human readable

### LINKED ISSUE #5271: Improve error message for service not in default namespace.
1- create a NodePort service in not default namespace:

```
$ kubectlget svc -A
NAMESPACE     NAME                    TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)                  AGE
default       kubernetes              ClusterIP   10.96.0.1        <none>        443/TCP                  15d
dgeek-v1      dgeek-helm-dgeek-helm   NodePort    10.103.186.237   <none>        80:32316/TCP             3m40s
```

2- try to open the service url:
```
$ minikube service dgeek-helm-dgeek-helm

💣  Error opening service
❌  Error: [SERVICE_NOT_FOUND] Could not find finalized endpoint being pointed to by dgeek-helm-dgeek-helm: Temporary Error: Error getting service dgeek-helm-dgeek-helm: services "dgeek-helm-dgeek-helm" not found
Temporary Error: Error getting service dgeek-helm-dgeek-helm: services "dgeek-helm-dgeek-helm" not found
Temporary Error: Error getting service dgeek-helm-dgeek-helm: services "dgeek-helm-dgeek-helm" not found
Temporary Error: Error getting service dgeek-helm-dgeek-helm: services "dgeek-helm-dgeek-helm" not found
💡  Suggestion: Please make sure the service you are looking for is deployed or is in the correct namespace.
⁉️   Related issues:
    ▪ https://github.com/kubernetes/minikube/issues/4599

```

the error message is big and suggestion gets over-shaddowed.

I suggest the error message say

`service X was not found in default namespace , please try with "minikube service X -n Y"`

nanikjava:
fixes #5271 

Changes made:
* service.go - changes to the error string returned
* service_test.go - modify TestWaitAndMaybeOpenService test case to accomodate
for the new changes

### DIFF
--- a/pkg/minikube/service/service.go
+++ b/pkg/minikube/service/service.go
@@ -275,7 +275,7 @@ func WaitAndMaybeOpenService(api libmachine.API, namespace string, service strin
 	chkSVC := func() error { return CheckService(namespace, service) }
 
 	if err := retry.Expo(chkSVC, time.Duration(interval)*time.Second, time.Duration(wait)*time.Second); err != nil {
-		return errors.Wrapf(err, "Service %s was not found in %s namespace , please try with 'minikube service %s -n Y'", service, namespace, service)
+		return errors.Wrapf(err, "Service %s was not found in %q namespace. You may select another namespace by using 'minikube service %s -n <namespace>", service, namespace, service)
 	}
 
 	serviceURL, err := GetServiceURLsForService(api, namespace, service, urlTemplate)
--- a/pkg/minikube/service/service.go
+++ b/pkg/minikube/service/service.go
@@ -275,7 +275,7 @@ func WaitAndMaybeOpenService(api libmachine.API, namespace string, service strin
 	chkSVC := func() error { return CheckService(namespace, service) }
 
 	if err := retry.Expo(chkSVC, time.Duration(interval)*time.Second, time.Duration(wait)*time.Second); err != nil {
-		return errors.Wrapf(err, "Could not find finalized endpoint being pointed to by %s", service)
+		return errors.Wrapf(err, "Service %s was not found in default namespace , please try with 'minikube service %s -n Y'", service, service)
 	}
 
 	serviceURL, err := GetServiceURLsForService(api, namespace, service, urlTemplate)
--- a/pkg/minikube/service/service.go
+++ b/pkg/minikube/service/service.go
@@ -275,7 +275,7 @@ func WaitAndMaybeOpenService(api libmachine.API, namespace string, service strin
 	chkSVC := func() error { return CheckService(namespace, service) }
 
 	if err := retry.Expo(chkSVC, time.Duration(interval)*time.Second, time.Duration(wait)*time.Second); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/minikube/service/service.go
```diff
@@ -275,7 +275,7 @@ func WaitAndMaybeOpenService(api libmachine.API, namespace string, service strin
 	chkSVC := func() error { return CheckService(namespace, service) }
 
 	if err := retry.Expo(chkSVC, time.Duration(interval)*time.Second, time.Duration(wait)*time.Second); err != nil {
-		return errors.Wrapf(err, "Could not find finalized endpoint being pointed to by %s", service)
+		return errors.Wrapf(err, "Service %s was not found in default namespace , please try with 'minikube service %s -n Y'", service, service)
```
tstromberg: I love this text, but it should detect if `namespace` is default, otherwise the error message doesn't make much sense.

pkg/minikube/service/service_test.go
```diff
@@ -824,6 +845,7 @@ func TestWaitAndMaybeOpenService(t *testing.T) {
 		urlMode     bool
 		https       bool
 		err         bool
+		nondefault  bool
```
tstromberg: bools should never be negative or have a name that starts with with `no`: it gets too confusing, particularly for non-english readers, when `nondefault=false`. I also don't understand what this bool is the default of.

Instead of `default` or `nondefault`, I think this would be more readable if each test directly defined which `servicesMap` to use.

nanikjava: Will look into the test again and refactor to make it more readable for code structure

pkg/minikube/service/service.go
nanikjava: Made changes to the string to include the service name.

pkg/minikube/service/service_test.go
nanikjava: Refactor the code to split out the test on it's own to make it more easier to understand for other. Let me know what you think ?. Thanks

pkg/minikube/service/service.go
nanikjava: @tstromberg any feedback ? Thanks

pkg/minikube/service/service.go
```diff
@@ -275,7 +275,7 @@ func WaitAndMaybeOpenService(api libmachine.API, namespace string, service strin
 	chkSVC := func() error { return CheckService(namespace, service) }
 
 	if err := retry.Expo(chkSVC, time.Duration(interval)*time.Second, time.Duration(wait)*time.Second); err != nil {
-		return errors.Wrapf(err, "Could not find finalized endpoint being pointed to by %s", service)
+		return errors.Wrapf(err, "Service %s was not found in %s namespace , please try with 'minikube service %s -n Y'", service, namespace, service)
```
tstromberg: If I understand properly, "Y" doesn't have a special meaning. What do you think about something like this:
```
Service %s was not found in the %q namespace. You may select another namespace by using:

'minikube service %s -n <namespace>'
```

nanikjava: Make sense. PR has been updated. Thanks

Issue #4346, opened 2019-05-24
TestNewSSHClient flake: Error running "foo": wait: remote command exited


tstromberg:
As seen at https://travis-ci.org/kubernetes/minikube/builds/536576254?utm_source=github_status&utm_medium=notification and on other builds:

```
ok  	k8s.io/minikube/pkg/minikube/service	0.028s	coverage: 40.7% of statements
--- FAIL: TestNewSSHClient (0.19s)
    sshutil_test.go:53: Error running command: foo
FAIL
coverage: 75.0% of statements
FAIL	k8s.io/minikube/pkg/minikube/sshutil	0.197s
ok  	k8s.io/minikube/pkg/minikube/tunnel	2.073s	coverage: 64.5% of statements
ok  	k8s.io/minikube/pkg/util	0.847s	coverage: 51.7% of statements
make: *** [test] Error 16
The command "make test" exited with 2.
```

Based on the timing, this may be related by recent updates to sshutil and supporting libraries that fixed lint errors.

tstromberg:
Although we now have better logs, this is still an issue:

```
--- FAIL: TestNewSSHClient (0.61s)
    ssh_mock.go:173: Listening on 127.0.0.1:37039
    ssh_mock.go:83: Accepting...
    ssh_mock.go:83: Accepting...
    ssh_mock.go:132: exec request received: &{Type:exec WantReply:true Payload:[0 0 0 3 102 111 111] ch:0xc00010c0c0 mux:<nil>}
    sshutil_test.go:60: Error running "foo": wait: remote command exited without exit status or exit signal
    ssh_mock.go:189: Stopping
```

PR #5565, opened 2019-10-08
Allow addon enabling and disabling when minikube is not running

### LINKED ISSUE #5510: Cannot enable addons if minikube instance is not running
```bash
minikube addons enable ingress
```

💣  minikube is not running, so the service cannot be accessed
</details>

💣  api load: filestore "minikube": Docker machine "minikube" does not exist. Use "docker-machine ls" to list machines. Use "docker-machine create" to add a new one.

</details>

Mac OS 10.14.6

Maybe this is part of the minikube design but since addons are not tied to any 1 particular machine why does a machine have to be running in order to enable an addon? At least I think the error message should say something like `In order to enable an addon at least 1 machine must be running` however I think this should be fixed so that addons can be enabled without having a machine running.

RickyRajinder:
Fixes #5510
Currently, if you try to enable an addon while minikube is not up, an error is returned saying
>  minikube is not running, so the service cannot be accessed

This PR attempts to support addon disabling/enabling even when minikube is not running. All feedback and suggestions are welcome.

### DIFF
--- a/cmd/minikube/cmd/config/enable_test.go
+++ b/cmd/minikube/cmd/config/enable_test.go
@@ -16,10 +16,24 @@ limitations under the License.
 
 package config
 
-import "testing"
+import (
+	"testing"
+
+	"gotest.tools/assert"
+	pkgConfig "k8s.io/minikube/pkg/minikube/config"
+	"k8s.io/minikube/pkg/minikube/localpath"
+)
 
 func TestEnableUnknownAddon(t *testing.T) {
 	if err := Set("InvalidAddon", "false"); err == nil {
 		t.Fatalf("Enable did not return error for unknown addon")
 	}
+}
+
+func TestEnableAddon(t *testing.T) {
+	if err := Set("ingress", "true"); err != nil {
+		t.Fatalf("Enable returned unexpected error: " + err.Error())
+	}
+	config, _ := pkgConfig.ReadConfig(localpath.ConfigFile)
+	assert.Equal(t, config["ingress"], true)
 }
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -604,12 +604,13 @@ func CreateSSHShell(api libmachine.API, args []string) error {
 
 // EnsureMinikubeRunningOrExit checks that minikube has a status available and that
 // the status is `Running`, otherwise it will exit
-func EnsureMinikubeRunningOrExit(api libmachine.API, exitStatus int) {
+func IsMinikubeRunning(api libmachine.API) bool {
 	s, err := GetHostStatus(api)
 	if err != nil {
-		exit.WithError("Error getting machine status", err)
+		return false
 	}
 	if s != state.Running.String() {
-		exit.WithCodeT(exit.Unavailable, "minikube is not running, so the service cannot be accessed")
+		return false
 	}
+	return true
 }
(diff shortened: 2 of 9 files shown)


RickyRajinder:
It appears the test is failing because there is no existing minikube machine in Travis. Could anyone suggest a workaround to this?

medyagh:
@RickyRajinder  the lint is failing because the files needs to be goimported
you could use
```goimports -w ...(path to the file)```

for the unit test failure, maybe you could use the Mock Machine as seen in the the other uni tests? ( I haven't had chance to study your PR yet to give more precise advice but maybe that is helpful ?)

File: cmd/minikube/cmd/config/enable_test.go [hunk 1]
```diff
@@ -16,10 +16,20 @@ limitations under the License.
...
+)
 
 func TestEnableUnknownAddon(t *testing.T) {
 	if err := Set("InvalidAddon", "false"); err == nil {
 		t.Fatalf("Enable did not return error for unknown addon")
 	}
 }
+func TestEnableAddon(t *testing.T) {
+	if err := Set("ingress", "true"); err != nil {
+		if !strings.Contains(err.Error(), "was already enabled") {
```
tstromberg: I don't understand this test - why would ingress already be enabled when this test is run? Enabling an enabled addon should not cause an error. We should aim for idempotent API's when possible.

Also, never compare error strings in go: https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully


Also, please don't compare error strings, they are not an API.

pkg/minikube/machine/client.go
```diff
@@ -147,8 +147,8 @@ func (api *LocalClient) Close() error {
 }
 
 // CommandRunner returns best available command runner for this host
-func CommandRunner(h *host.Host) (command.Runner, error) {
-	if h.DriverName == constants.DriverMock {
+func CommandRunner(h *host.Host, minikubeRunning bool) (command.Runner, error) {
```
tstromberg: This API change doesn't feel right to me. The caller should first detect that the cluster is not running, and not ask for a runner in that case. Silently passing a mock back can result in unexpected behavior.

pkg/minikube/cluster/cluster.go
```diff
@@ -604,12 +604,13 @@ func CreateSSHShell(api libmachine.API, args []string) error {
 
 // EnsureMinikubeRunningOrExit checks that minikube has a status available and that
 // the status is `Running`, otherwise it will exit
-func EnsureMinikubeRunningOrExit(api libmachine.API, exitStatus int) {
+func IsMinikubeRunning(api libmachine.API) bool {
```
tstromberg: Great cleanup here. =)

tstromberg:
@minikube-bot OK to test

tstromberg:
FYI, flaky ssh test failure was https://github.com/kubernetes/minikube/issues/4346

pkg/minikube/machine/client.go
RickyRajinder: What time of runner should be used?

cmd/minikube/cmd/config/enable_test.go
RickyRajinder: Is there a way to create a mock machine in a directory? I found .minikube/ test data but they only contain profiles.

tstromberg:
Thank you!

PR #5574, opened 2019-10-09
Add libmachine debug logs back

### LINKED ISSUE #5453: logs no longer include libmachine commands, making hyperv impossible to debug
We used to have logs for Hyper-V and Virtualbox machine drivers that showed each command run. Where'd they go?

sharifelgamal:
Fixes #5453 

`--alsologtostderr` will now always print out all libmachine debug logs, which seems like a reasonable choice.

### DIFF
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -37,7 +37,6 @@ import (
 	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/minikube/exit"
 	"k8s.io/minikube/pkg/minikube/localpath"
-	"k8s.io/minikube/pkg/minikube/notify"
 	"k8s.io/minikube/pkg/minikube/translate"
 )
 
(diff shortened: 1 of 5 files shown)


medyagh:
I see werid error in common.sh
```
17:24:20 >> virsh VM list after clean up (should be empty):
17:24:20  Id    Name                           State
17:24:20 ----------------------------------------------------
17:24:20 
17:24:20 ./common.sh: line 141: syntax error near unexpected token `|'
17:24:20 Build step 'Execute shell' marked build as failure
```

cmd/minikube/cmd/root.go
```diff
@@ -76,6 +77,7 @@ var RootCmd = &cobra.Command{
 				exit.WithError("logdir set failed", err)
 			}
 		}
+		notify.MaybePrintUpdateTextFromGithub()
```
tstromberg: This shouldn't be here?

sharifelgamal: Removed

PR #5584, opened 2019-10-09
jenkins: Resilient VirtualBox VM and test directory cleanup

tstromberg:
- Make sure that "list vms" regex only matches guid portion
- Add logging
- unregistervm still called if startvm fails

Related: #5583

### DIFF
--- a/hack/jenkins/common.sh
+++ b/hack/jenkins/common.sh
@@ -134,11 +134,13 @@ if type -P virsh; then
 fi
 
 if type -P vboxmanage; then
-  vboxmanage list vms || true
   vboxmanage list vms \
     | egrep -o '{.*?}' \
-    | xargs -I {} sh -c "vboxmanage startvm {} --type emergencystop; vboxmanage unregistervm {} --delete" \
-    || true
+    | xargs -I{} vboxmanage startvm {} --type emergencystop || true
+
+  vboxmanage list vms \
+    | egrep -o '{.*?}' \
+    | xargs -I{} vboxmanage unregistervm {} || true
 
   echo ">> VirtualBox VM list after clean up (should be empty):"
   vboxmanage list vms || true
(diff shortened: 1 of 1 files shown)


tstromberg:
It seems like our script is eating it on wildcards, though I don't see `set -f` open. Will make the script work in this case:

```
>> Cleaning up after previous test runs ...
* Cleaning stale test root: /Users/jenkins/minikube-integration/*
find: /Users/jenkins/minikube-integration/*: No such file or directory
find: /Users/jenkins/minikube-integration/*: No such file or directory
find: /Users/jenkins/minikube-integration/*: No such file or directory
rmdir: /Users/jenkins/minikube-integration/*: No such file or directory
ls: /Users/jenkins/minikube-integration/*: No such file or directory
Build step 'Execute shell' marked build as failure
```

tstromberg:
This now removes the following out of the jenkins config:

`vboxmanage list -l hostonlyifs | grep -w "Name:" |  awk '{$1=""; print $0}' | xargs -l vboxmanage hostonlyif remove || true`

as it was creating the following error:

```
+ vboxmanage list -l hostonlyifs
+ grep -w Name:
+ awk '{$1=""; print $0}'
+ xargs -l vboxmanage hostonlyif remove
Oracle VM VirtualBox Command Line Management Interface Version 5.2.18
(C) 2005-2018 Oracle Corporation
All rights reserved.

Usage:

VBoxManage hostonlyif       ipconfig <name>
                            [--dhcp |
                            --ip<ipv4> [--netmask<ipv4> (def: 255.255.255.0)] |
                            --ipv6<ipv6> [--netmasklengthv6<length> (def: 64)]]
                            create |
                            remove <name>


Syntax error: No interface name was specified
```

medyagh:
one thing I noitced in vboxmanage list is, the order of columns changes, if the VM is "inaccessible".
that why I had I to do this: https://github.com/kubernetes/minikube/commit/c5111ef3f5120f1f90dcb87456b5de2acd5a4742

 (probably not related to your PR) but when putting regex on vbox manage be mindful of that.

PR #5572, opened 2019-10-08
Bump golangci-lint to v1.20.0

corneliusweig:
This version of golangci-lint comes with significant memory improvements, making some
adjustments for small memory footprint in the CI obsolete.

See https://github.com/golangci/golangci-lint/issues/337

PR #5591, opened 2019-10-10
Remove unused enableUpdateNotification variable

tstromberg:
This is no longer necessary as we now only check for updates during `minikube start`.


tstromberg:
Related PR that inspired this one: #5572

Issue #3329, opened 2018-11-12
Warn if incompatible kubectl version is in use


tstromberg:
Currently, it's possible to use minikube on a host running v1.6.1 against a master running v1.12.1. This isn't a supported configuration, as per https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#supported-releases-and-component-skew

We should have a flight check that validates that ancient version of kubectl aren't in use. This would also serve as an early warning to users about issues such as #3203

tstromberg:
If someone wants to implement this:

* `showKubectlConnectInfo` in `cmd/minikube/cmd/start.go` might be the right place to start
* `kubectl version --client -o json` will return an easily JSON blob containing the current version of the kubectl client.
* If `clientVersion` is more than one minor release away from `k8sVersion`, issue a warning, like:

`out.WarningT("The version of kubectl [{{.kcVersion}}] installed is incompatible with your Kubernetes [{{.k8sVersion}}] deployment"). Please upgrade/downgrade as necessary, or use `minikube kubectl` to connect to your cluster `

tstromberg:
@nanikjava - excellent news! Please let me know if you need any additional help, either here or on the Slack #minikube  channel.

nanikjava:
@tstromberg PR ready https://github.com/kubernetes/minikube/pull/5596

PR #5596, opened 2019-10-11
Warn if incompatible kubectl version is in use

nanikjava:
Fixes: #3329

Modification is done inside start.go where additional checking on the
version returned via the kubectl CLI is checked.

Running 'kubectl version --output=json' will return both client
and server information.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 	"net"
 	"net/url"
@@ -129,6 +130,17 @@ var (
 	apiServerIPs     []net.IP
 	extraOptions     cfg.ExtraOptionSlice
 )
+
+type kubectlversion struct {
+	CVersion ClientVersion `json:"clientVersion"`
+	SVersion ClientVersion `json:"serverVersion"`
+}
+
+type ClientVersion struct {
+	Major      string `json:"major"`
+	Minor      string `json:"minor"`
+	GitVersion string `json:"gitVersion"`
+}
 
 func init() {
 	initMinikubeFlags()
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -477,16 +489,57 @@ func showVersionInfo(k8sVersion string, cr cruntime.Manager) {
...
+					out.T(out.Tip, "The version of kubectl {{.kubectl}} installed is incompatible with your Kubernetes {{.kubernetes}}  deployment. Please upgrade/downgrade as necessary, or use minikube kubectlto connect to your cluster",
+						out.V{"kubectl": clientVersion.CVersion.GitVersion, "kubernetes": serverVersion.SVersion.GitVersion})
+				}
+			}
+		} else {
+			out.T(out.Tip, "For best results, install kubectl: https://kubernetes.io/docs/tasks/tools/install-kubectl/")
+		}
+	}
+
 	if kcs.KeepContext {
```
tstromberg: Move this output to the top of the function. This will allow both the version mismatch error to come at the end, which is more visible, and will allow you to exit early if kubectl does not exist, which will allow for more readable code as the version checking won't have to be indented.

cmd/minikube/cmd/start.go
```diff
@@ -477,16 +489,57 @@ func showVersionInfo(k8sVersion string, cr cruntime.Manager) {
...
+	clientVersion := kubectlversion{}
+	serverVersion := kubectlversion{}
+
+	path, err := exec.LookPath("kubectl")
+
+	if err != nil {
+		out.T(out.Tip, "For best results, install kubectl: https://kubernetes.io/docs/tasks/tools/install-kubectl/")
+	} else {
+		glog.Infof("Kubectl found at the following location %s. Let's execute and get the result", path)
+		output, err = exec.Command(path, "version", "--output=json").Output()
```
tstromberg: This may show the wrong server version, as it queries the active context kubectl is configured for, which may not be minikube.

My suggestion is to use `kubectl version --short --output=json`, and add `k8sVersion` as an argument to the function.

cmd/minikube/cmd/start.go
```diff
@@ -477,16 +489,57 @@ func showVersionInfo(k8sVersion string, cr cruntime.Manager) {
...
+	var output []byte
+	clientVersion := kubectlversion{}
+	serverVersion := kubectlversion{}
+
+	path, err := exec.LookPath("kubectl")
+
+	if err != nil {
+		out.T(out.Tip, "For best results, install kubectl: https://kubernetes.io/docs/tasks/tools/install-kubectl/")
+	} else {
+		glog.Infof("Kubectl found at the following location %s. Let's execute and get the result", path)
```
tstromberg: While good for initial debugging, I think this log message can be removed.

cmd/minikube/cmd/start.go
```diff
@@ -477,16 +489,57 @@ func showVersionInfo(k8sVersion string, cr cruntime.Manager) {
...
+		output, err = exec.Command(path, "version", "--output=json").Output()
+
+		// ...once we get something back do some version checking
+		if err == nil {
```
tstromberg: I mention this later, but try to structure functions so that exceptional conditions are handled first, and exited early. So for istance:

if err != nil {
   .. return ..
}
.. do the rest with one level less indentation
```

File: cmd/minikube/cmd/start.go [hunk 5]
```diff
@@ -477,16 +489,57 @@ func showVersionInfo(k8sVersion string, cr cruntime.Manager) {
...
+			glog.Infof("Received output from kubectl %s", output)
+
+			// unmarshal both the {client}
+			clientjsonErr := json.Unmarshal(output, &clientVersion)
+
+			// .......  and {server} json
+			serverjsonErr := json.Unmarshal(output, &serverVersion)
```
tstromberg: No need to unmarshal the same output twice: the server version is already in `clientVersion.SVersion`

cmd/minikube/cmd/start.go
```diff
@@ -477,16 +489,57 @@ func showVersionInfo(k8sVersion string, cr cruntime.Manager) {
...
+
+			if (clientjsonErr != nil) || (serverjsonErr != nil) {
+				glog.Infof("There was an error processing the json output")
```
tstromberg: here is another opportunity to return early.

cmd/minikube/cmd/start.go
```diff
@@ -477,16 +489,57 @@ func showVersionInfo(k8sVersion string, cr cruntime.Manager) {
...
+			} else {
+				// obtain the minor version for both
+				serverMinor, _ := strconv.Atoi(serverVersion.SVersion.Minor)
+				clientMinor, _ := strconv.Atoi(serverVersion.CVersion.Minor)
+
+				if clientMinor < serverMinor {
```
tstromberg: This is not quite correct yet. https://kubernetes.io/docs/setup/release/version-skew-policy/#kubectl dictates that you are allowed up to one version difference.  One possibility is:

`if math.Abs(float64(clientMinor - serverMinor)) > 1 {`

cmd/minikube/cmd/start.go
```diff
@@ -477,16 +489,57 @@ func showVersionInfo(k8sVersion string, cr cruntime.Manager) {
...
+					out.T(out.Tip, "The version of kubectl {{.kubectl}} installed is incompatible with your Kubernetes {{.kubernetes}}  deployment. Please upgrade/downgrade as necessary, or use minikube kubectlto connect to your cluster",
```
tstromberg: It would be helpful to include the path. How about something like:

```
{{.path}} is version {{.version}}, and is incompatible with your specified Kubernetes version. You will need to update {{.path}} or use 'minikube kubectl' to connect with this cluster
```

PR #5611, opened 2019-10-13
Add json output for status

### LINKED ISSUE #5080: status command should support json output
For upstream tooling, we need JSON output for commands like `minikube status`.

minikube status

cc: @tstromberg

woodcockjosh:
Adds json output for `minikube status` command
Adds shorthand argument for format option

## Before
Command:
```shell
minikube status -o json
```
Output:
```text
Error: unknown shorthand flag: 'o' in -o


Options:
      --format='host: {{.Host}}
kubelet: {{.Kubelet}}
apiserver: {{.APIServer}}
kubectl: {{.Kubeconfig}}
': Go template format string for the status output.  The format for Go templates can be found here: https://golang.org/pkg/text/template/
For the list accessible variables for the template, see the struct values here: https://godoc.org/k8s.io/minikube/cmd/minikube/cmd#Status

Usage:
  minikube status [flags] [options]

Use "minikube status options" for a list of global command-line options (applies to all commands).
```

### Adds json output
Command:
```shell
minikube status -o json
```
Output:
```text
{"APIServer":"Running","Host":"Running","Kubeconfig":{"Correct":true,"IP":"192.168.99.189"},"Kubelet":"Running"}
```

### Adds shorthand for format option
Command:
```shell
minikube status --help
```
Output:
```text
Gets the status of a local kubernetes cluster.
        Exit status contains the status of minikube's VM, cluster and kubernetes encoded on it's bits in this order from right
to left.
        Eg: 7 meaning: 1 (for minikube NOK) + 2 (for cluster NOK) + 4 (for kubernetes NOK)

Options:
  -f, --format='host: {{.Host}}
kubelet: {{.Kubelet}}
apiserver: {{.APIServer}}
kubectl: {{.Kubeconfig}}
': Go template format string for the status output.  The format for Go templates can be found here:
https://golang.org/pkg/text/template/
For the list accessible variables for the template, see the struct values here:
https://godoc.org/k8s.io/minikube/cmd/minikube/cmd#Status
  -o, --output='text': minikube status --output OUTPUT. json, text

Usage:
  minikube status [flags] [options]

Use "minikube status options" for a list of global command-line options (applies to all commands).
```

### Output and Format options cannot both be used at the same time
Command:
```shell
minikube status -o json -f "hello"
```
Output:
```text
💡  Cannot use both --output and --format options
```

Fixes #5080

### DIFF
--- a/cmd/minikube/cmd/status.go
+++ b/cmd/minikube/cmd/status.go
@@ -17,7 +17,10 @@ limitations under the License.
 package cmd
 
 import (
+	"encoding/json"
+	"fmt"
 	"os"
+	"strings"
 	"text/template"
 
 	"github.com/docker/machine/libmachine/state"
@@ -35,13 +38,20 @@ import (
 )
 
 var statusFormat string
+var output string
+
+type KubeconfigStatus struct {
+	Correct bool
+	IP      string
+}
 
 // Status represents the status
 type Status struct {
-	Host       string
-	Kubelet    string
-	APIServer  string
-	Kubeconfig string
+	Host             string
+	Kubelet          string
+	APIServer        string
+	Kubeconfig       string
+	KubeconfigStatus KubeconfigStatus
 }
 
 const (
@@ -63,6 +73,11 @@ var statusCmd = &cobra.Command{
 	Exit status contains the status of minikube's VM, cluster and kubernetes encoded on it's bits in this order from right to left.
 	Eg: 7 meaning: 1 (for minikube NOK) + 2 (for cluster NOK) + 4 (for kubernetes NOK)`,
 	Run: func(cmd *cobra.Command, args []string) {
+
+		if output != "text" && statusFormat != defaultStatusFormat {
+			exit.UsageT("Cannot use both --output and --format options")
+		}
+
 		var returnCode = 0
 		api, err := machine.NewAPIClient()
 		if err != nil {
@@ -78,6 +93,8 @@ var statusCmd = &cobra.Command{
 		kubeletSt := state.None.String()
 		kubeconfigSt := state.None.String()
 		apiserverSt := state.None.String()
+		var ks bool
+		var ipString = ""
 
 		if hostSt == state.Running.String() {
 			clusterBootstrapper, err := getClusterBootstrapper(api, viper.GetString(cmdcfg.Bootstrapper))
@@ -110,12 +127,13 @@ var statusCmd = &cobra.Command{
 				returnCode |= clusterNotRunningStatusFlag
 			}
 
-			ks, err := kubeconfig.IsClusterInConfig(ip, config.GetMachineName())
+			ks, err = kubeconfig.IsClusterInConfig(ip, config.GetMachineName())
 			if err != nil {
 				glog.Errorln("Error kubeconfig status:", err)
 			}
 			if ks {
 				kubeconfigSt = "Correctly Configured: pointing to minikube-vm at " + ip.String()
+				ipString = ip.String()
 			} else {
 				kubeconfigSt = "Misconfigured: pointing to stale minikube-vm." +
 					"\nTo fix the kubectl context, run minikube update-context"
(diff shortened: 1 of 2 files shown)


cmd/minikube/cmd/status.go
```diff
@@ -35,13 +38,20 @@ import (
...
 type Status struct {
-	Host       string
-	Kubelet    string
-	APIServer  string
-	Kubeconfig string
+	Host             string
+	Kubelet          string
+	APIServer        string
+	Kubeconfig       string
+	KubeconfigStatus KubeconfigStatus
```
tstromberg: The `Status.KubeconfigStatus` name stutters. Is there a different naming convention that we could use?

Also, it isn't outwardly apparent what the difference is between `Status.Kubeconfig` and `Status.KubeconfigStatus` is.

cmd/minikube/cmd/status.go
```diff
@@ -35,13 +38,20 @@ import (
 )
 
 var statusFormat string
+var output string
+
+type KubeconfigStatus struct {
```
tstromberg: Could this be a private struct?

If not, please document the use of this struct and its fields.

cmd/minikube/cmd/status.go
```diff
@@ -35,13 +38,20 @@ import (
...
+	Correct bool
```
tstromberg: What is `Correct` in relation to `KubeconfigStatus`? The naming here does not seem intuitive to me.

woodcockjosh: It means that the kubeconfig is correctly configured. I can change this to `CorrectlyConfigured` ?

woodcockjosh:
@tstromberg I please check this update. From my perspective the current output doesn't make sense. 
1. In the code the kubeconfig status is referenced as kubeconfig but in the output it is referenced as kubectl
1. In all the other statuses there is a single word status but in the kubeconfig status it is more of a message.
1. Do we really need the ip address that kubeconfig is configured for? I don't think so. 

For this reason I think it makes sense to modify the current output while adding json output. The JSON output needs to be simple and doesn't need to have user messages. We also want the json output to be somewhat congruent with the default output. I have separated the kubeconfig status from the warning message in case the kubeconfig is misconfigured. 

I think this is cleaner and less confusing.

tstromberg:
PR looks good, but the status integration test is failing:

```
                -- /stdout --
            functional_test.go:181: [out/minikube-linux-amd64 status] failed: exit status 1
            functional_test.go:185: (dbg) Run:  out/minikube-linux-amd64 status -f host:{{.Host}},kublet:{{.Kubelet}},apiserver:{{.APIServer}},kubectl:{{.Kubeconfig}}
            functional_test.go:185: (dbg) Non-zero exit: out/minikube-linux-amd64 status -f host:{{.Host}},kublet:{{.Kubelet}},apiserver:{{.APIServer}},kubectl:{{.Kubeconfig}}: exit status 1 (88.157569ms)
                -- stdout --
                host:,kublet:,apiserver:,kubectl:
                -- /stdout --
            functional_test.go:187: [out/minikube-linux-amd64 status -f host:{{.Host}},kublet:{{.Kubelet}},apiserver:{{.APIServer}},kubectl:{{.Kubeconfig}}] failed: exit status 1
            functional_test.go:191: [out/minikube-linux-amd64 status -f host:{{.Host}},kublet:{{.Kubelet}},apiserver:{{.APIServer}},kubectl:{{.Kubeconfig}}] failed: exit status 1. Output for custom format did not match
            functional_test.go:195: (dbg) Run:  out/minikube-linux-amd64 status -o json
            functional_test.go:195: (dbg) Non-zero exit: out/minikube-linux-amd64 status -o json: exit status 1 (79.102917ms)
```

woodcockjosh:
@tstromberg works on my machine :-D
```shell script
make integration -e TEST_ARGS="-test.parallel=1 -test.run TestFunctional/parallel/StatusCmd --profile=minikube --cleanup=false"
```

```text
go test -v -test.timeout=60m ./test/integration --tags="integration container_image_ostree_stub containers_image_openpgp" -test.parallel=1 -test.run TestFunctional/parallel/StatusCmd --profile=minikube --cleanup=false
=== RUN   TestFunctional
=== PAUSE TestFunctional
=== CONT  TestFunctional
=== RUN   TestFunctional/parallel
=== RUN   TestFunctional/parallel/StatusCmd
=== PAUSE TestFunctional/parallel/StatusCmd
=== CONT  TestFunctional/parallel/StatusCmd
--- PASS: TestFunctional (1.28s)
    --- PASS: TestFunctional/parallel (0.00s)
        --- PASS: TestFunctional/parallel/StatusCmd (1.28s)
            functional_test.go:179: (dbg) Run:  out/minikube status
            functional_test.go:185: (dbg) Run:  out/minikube status -f host:{{.Host}},kublet:{{.Kubelet}},apiserver:{{.APIServer}},kubeconfig:{{.Kubeconfig}}
            functional_test.go:195: (dbg) Run:  out/minikube status -o json
    helpers.go:170: skipping cleanup of minikube (--cleanup=false)
PASS
ok      k8s.io/minikube/test/integration        1.316s
```

test/integration/functional_test.go
```diff
@@ -173,6 +175,46 @@ func validateComponentHealth(ctx context.Context, t *testing.T, profile string)
 	}
 }
 
+func validateStatusCmd(ctx context.Context, t *testing.T, profile string) {
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), "status"))
```
tstromberg: You'll need to pass `"-p", profile` to all of commands you are running, otherwise these tests only run against the `minikube` profile, which doesn't necessarily exist when run in parallel.

woodcockjosh: @tstromberg I have made changes as suggested.

tstromberg:
Looks good. Thank you!

PR #5614, opened 2019-10-14
Performance and security enhancement for ingress-dns addon

woodcockjosh:
This is a performance and security enhancement for the ingress-dns addon. 

1. Resolves issue where ingress-dns pod is evicted when tcp connection cannot be established by ingress nginx controller
1. Removes cluster-admin role for ingress-dns addon pod
1. Removes use of ingress-nginx as passthrough
1. Ingress dns pod uses host network and binds only to Pod IP / Node IP rather than 0.0.0.0
1. Tcp routing removed as it is not used
1. Pod IP / Node IP is used as minikube IP which is the same when pod is using host network rather than querying the cluster IP


tstromberg:
@minikube-bot OK to test

PR #5621, opened 2019-10-14
WIP: Re-organize $MINIKUBE_HOME to be race-proof and OS-idiomatic

### LINKED ISSUE #4109: Respect the XDG Base Directory Specification
Under Linux/Unix systems, there is the [XDG Base Directory Specifiation](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) that states where applications should store user-related data. For example, the `config.json` currently resides in `.minikube/config/config.json`, but it should be in `$XDG_CONFIG_HOME/minikube/config.json` (which would be `.config/minikube/config.json` most of the time).

### LINKED ISSUE #5353: certs should be per-profile
`SetupCerts` is not profile aware, and very racy.

It constantly overwrites the certificates every time you run `minikube start`, which is bad if two are started at the same time. We should make it so that clusters are updated to use per-profile certs to avoid raciness.

tstromberg:
Major changes:

- libmachine stores are now per-profile. Previously, libmachine used $MINIKUBE_HOME as it's store.
- As a side effect, certificates are now per-profile. Fixes #5353 
- $MINIKUBE_HOME now defaults to idiomatic locations based on the operating system.
  * Uses `os.UserConfigDir` for configuration, `os.UserCacheDir` for cache.  
  * On Windows, instead of C:\Users\username.minikube, this uses %AppData%/minikube, for instance. 
  * On macOS, this means ~/Application Support/minikube
  * On Linux, this means ~/.config/minikube
- This PR provides a forward migration path, but not a backwards one if users downgrade.

Closes #4109 and #5353

### DIFF
--- a/cmd/minikube/cmd/cache.go
+++ b/cmd/minikube/cmd/cache.go
@@ -20,7 +20,6 @@ import (
 	"github.com/spf13/cobra"
 	cmdConfig "k8s.io/minikube/cmd/minikube/cmd/config"
 	"k8s.io/minikube/pkg/minikube/config"
-	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/minikube/exit"
 	"k8s.io/minikube/pkg/minikube/localpath"
 	"k8s.io/minikube/pkg/minikube/machine"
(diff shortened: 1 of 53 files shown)


tstromberg:
TODO:

- Add missing function for updating paths within libmachine `config.json` files
- Preserve `--embed-certs` default (save for another PR)

tstromberg:
Will re-open split into a couple of PR's.

PR #5623, opened 2019-10-14
Delegate release notes generation to external tool

### LINKED ISSUE #5252: Expose release_notes helper as go sub-module
The release notes script in https://github.com/kubernetes/minikube/blob/3c0dc9d878aba31c93087ffbcb924008f75b737c/hack/release_notes/listpullreqs.go#L1 is a very useful helper. I found various copies of this on github, but I think it originates from here. Now we would like to use this script in `krew.dev` as well (see https://github.com/kubernetes-sigs/krew/pull/326). However, we can't really maintain this and we would not receive updates if we copy the script into our repository.

Would you therefore be open to expose this helper as a sub-module, so that it can be installed with `go get github.com/kubernetes/minikube/hack/release_notes` as a standalone tool? This would mean to make the `org` and `repo` variable configurable with CLI flags. I'd be happy take that up.

corneliusweig:
This PR extracts the logic to generate the release notes changelog to an external tool. It was discussed in https://github.com/kubernetes/minikube/issues/5252 to expose the `listpullreqs.go` script as a submodule so that other projects can depend on it. Instead, I extracted the script to a repository in my github space, as discussed. I promise to maintain this script and not break the format :)

The main change is in `hack/release-notes.sh`, where the external tool is installed in the `hack` directory if missing.

Close #5252


tstromberg:
Thank you!

Issue #4883, opened 2019-07-26
UI: Advice when user tries to set profile name to "delete" or "start"


medyagh:
to delete a minikube profile you need to do

```minikube delete -p prof_name```

but if you do

```minikube profile delete```

it will actually create a profile named delete. because that command sets  profile name.

we need to at lest add an advice to the user, that says

Did you mean to delete a profile ? and here is the command to delete profile !

I recommend the same approach for other key words such as, start, stop, status, ... to make it less confusing for new users.

we could add this to the Unit tests of profile, so if someone tries to set the profile name to a list of reserved words, it should return an error.

bhanu011:
i would like to pick this up. /assign

medyagh:
its yours ! @bhanu011  let me know if you need help

bhanu011:
@medyagh when user tries to create profiles with reserved keywords(delete, start, stop, status, etc ...) system should restrict the user from creating and throw an error?

medyagh:
I would say exit with an Advice, and yes it should not allow you create a profile in the reserved names.


the reserved names could be the basic commands,
start,stop, status, delete, config

medyagh:
@bhanu011 we had a discussion on this in minikube office hours.

Please sync with @josedonizetti  to handle this issue

bhanu011:
@medyagh ok @josedonizetti any changes to the design discussed in above comments?

medyagh:
@bhanu011 yes we had a change in design, you could still work on it if you like, please sync with @josedonizetti for any help you need with this PR !

josedonizetti:
@bhanu011 Currently if you do `minikube profiler NAME` this will create a new profile, and switch you to it. After discussing this today we thought that profile creation should only happen by the `start` command, and the `profile` should be only used to switch among existing profiles. 

Basically, we need to change `profile` to validate if a profile exists before switching and if not warn the user.

Let me know if you still want to work on this! And if you have any questions. :)

bhanu011:
i am done with code according to previous design i was just writing unit test cases. I would like to continue working on this. Here are few questions I have
1. where can i get all existing profiles? is it from this folder~/.minikube/profiles/ or is there any better way? 
2. What should happen when user trying to create profiles with kind of reserved words like delete, stop etc? Do we need to allow the user to create profile with that name?
3. If we are allowing we should only show some advice like if you are looking for this try with this command ?

josedonizetti:
@bhanu011 

[...]

PR #5624, opened 2019-10-15
Add validation checking for minikube profile

nanikjava:
Fixes : #4883

New function ProfileNameInReservedKeywords(..) has been added inside
pkg/minikube/config/profile.go

The new function perform validation to make sure the profile name is not
in the list of keywords.

Following are the keywords that are used:

> start, stop, status, delete, config, open, profile, addons, cache, logs

The checking is case insensitive to cover combo of upper and lower case


sharifelgamal:
@minikube-bot OK to test

nanikjava:
Did another commit as reported is has conflict

PR #5639, opened 2019-10-16
Speed up RNG initialization on new systemd

tstromberg:
With the v1.5.0-beta.0 kernel, this change reduces sshd startup time after system boot by about 2 minutes on hyperkit. It previously blocked sshd from running with:

`random: crng init done`

Changes to the kernel command line:

- `random.trust_cpu=on`:  Trust the emulated CPU as a source of randomization
- `hw_rng_model=virtio`: Plumb in VirtIO RNG devices if available

For hyperkit:

- Remove `user=docker` argument, which differed from other VM drivers


sharifelgamal:
@minikube-bot OK to test

tstromberg:
@minikube-bot OK to test

PR #5647, opened 2019-10-17
common.sh tune-up: Add load warning, fix profile deletion, fix lint errors and noise

tstromberg:
- Adds output of `uptime`
- Adds load warning (if handy, we can make this a fatal issue later)
- Fixes cleanup, where in a recent refactor I neglected to add the parent directory to the path
- Combines and removes errant `minikube delete --all` call that did not have a MINIKUBE_HOME set.
- VirtualBox hostonlyifs cleanup will no longer try to cleanup an empty list of interfaces
- Fix lint errors mentioned by `shellcheck`

### DIFF
--- a/hack/jenkins/common.sh
+++ b/hack/jenkins/common.sh
@@ -159,7 +159,7 @@ if type -P virsh; then
 fi
 
 if type -P vboxmanage; then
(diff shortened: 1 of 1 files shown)


sharifelgamal:
none driver tests passed, so that's good. the kvm machine looks totally hosed though.

PR #5648, opened 2019-10-17
Automatically install periodic cleanup script on macOS and Linux

tstromberg:
I noticed that some of our newer Linux VM's didn't have the periodic cleanup script installs. This ensures that the cleanup script is updated as part of accepting a new Jenkins job.


medyagh:
Looks good

tstromberg:
@minikube-bot OK to test

PR #5626, opened 2019-10-15
fix "minikube update-context" command fail

govargo:
### What type of PR is this?
/kind bug

### What this PR does / why we need it:
In minikube v1.4.0, `minikube update-context` command always fails.

This PR fixes this bug of the command fail.

### Which issue(s) this PR fixes:
Fixes #5497

### Does this PR introduce a user-facing change?

Yes. this PR fixes bug in order to `minikube update-context` command work successfully.

**Before this PR, `minikube update-context` command always fails**
```
1. minikube update-context(error)
update config: Kubeconfig does not have a record of the machine cluster
```

**After this PR, `minikube update-context` command works**
```
1. minikube update-context(success)
minikube IP was already correctly configured for xxx.xxx.xxx.xxx
(or)
minikube IP has been updated to point at xxx.xxx.xxx.xxx
```

### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:
```
NONE
```

PR #5652, opened 2019-10-17
Add simple integration test for update-context

tstromberg:
To assert that #5626 does not regress


sharifelgamal:
There's a lint error but otherwise LGTM.

medyagh:
@tstromberg needs resolve conflict.

btw traiage party showed this PR as "Pull requests: Approved & Mergeable "

it would be cool if traige party could detect Needs Resolve Conflict

PR #5664, opened 2019-10-19
Upgrade toolbox versions, fedora and systemd

### LINKED ISSUE #5239: Upgrade Toolbox fedora version
Currently we are using `fedora:29`, should upgrade to `fedora:30` (or even `fedora:latest`)

```
TOOLBOX_DOCKER_IMAGE=fedora
TOOLBOX_DOCKER_TAG=29
```
We could also merge rkt/podman/systemd-nspawn with upstream, or at least get the bug fixes:

https://github.com/coreos/toolbox
https://coreos.com/os/docs/latest/install-debugging-tools.html

---

Can do this with #5237 
Similar issue to ~#3058~

_I think we will keep with Fedora 30 and using systemd-nspawn, rather than latest and podman_
https://en.wikipedia.org/wiki/Fedora_version_history (this has the same life-span as the other LTS)

afbjorklund:
* upgrade to latest fedora release (currently 30)
* support newer systemd-nspawn (232 and up)

Compare https://github.com/coreos/toolbox/blob/master/toolbox

Closes #5239

### DIFF
--- a/deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/usr/bin/toolbox
+++ b/deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/usr/bin/toolbox
@@ -8,11 +8,11 @@ machine=$(uname -m)
 case ${machine} in
 	aarch64 )
 		TOOLBOX_DOCKER_IMAGE=arm64v8/fedora
-		TOOLBOX_DOCKER_TAG=29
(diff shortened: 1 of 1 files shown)


tstromberg:
Thank you Anders!

kauedg:
I know it's already and old commit but it is not a good practice to use the "latest" tag of docker images. By doing so unpredictable errors may happen because the newer version was not tested.

PR #5665, opened 2019-10-19
Build minikube for arm64 (aarch64) as well

afbjorklund:
This a pure build target, no test done (yet)
No virtualization supported on this platform

Adopt packaging to handle multiple arch, but
don't package anything extra by default yet.


tstromberg:
Thank you!

Issue #5577, opened 2019-10-09
Can we get Sonarcloud setup for our main repository?


blueelvis:
Can we get Sonarcloud setup for our main repository? This is free for open source projects and provides valuable metrics like code quality issues and what not. It also supports Pull Request validations.

I have set it up on my fork and you can see the results over here - https://sonarcloud.io/dashboard?id=blueelvis_minikube

-Pranav

medyagh:
I don't see why not ! it is a good idea !

afbjorklund:
Note that maintaining SonarQube is like a full-time position, due to the amount of false positives...

Still a good idea, but none of the current complaints seems like it was actually any real issue ?

i.e. it currently lists:
* 12 Bugs
* 234 Vulnerabilities
* 276 Code Smells

All of these would need to be handled.

---

PS. I addressed the 3 Go bugs, the rest were in the website (8 HTML and 1 CSS)

And of the vulnerabilities, it is all about having "hardcoded IP addresses" in the tests.

sharifelgamal:
Is this still something we're interested in supporting?

PR #5670, opened 2019-10-20
Reduce code smell reported by SonarCloud

cueo:
Ref. issue: #5577 

Two changes:
1. Use opposite operator: `a != b` instead of `!(a == b)`
2. Rename couple of methods to match the convention.


cueo:
I signed it

PR #5676, opened 2019-10-21
upgrade go version 1.12.9 -> 1.12.12

tanjunchen:
There is a CVE for Go and this vulnerability is fixed in 1.12.12. We should update it.
See: https://nvd.nist.gov/vuln/detail/CVE-2019-16276

See issue golang/go#34540
Seems this is fixed in 1.12.12. I suggest we should update to 1.12.12 first

/kind bug
/priority important-soon

Fixes#5675

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@
 os: linux
 language: go
 go:
-  - 1.12.9
+  - 1.12.12
 env:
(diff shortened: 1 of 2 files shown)


afbjorklund:
Sounds good

/ok-to-test

PR #5692, opened 2019-10-22
updating for syntax error and being deprecated

asloan7:


medyagh:
@asloan7  do you mind sharing what is the reason for chaning -A to a longer format ?

asloan7:
@medyagh Maybe it is my version of kubectl, but I get: 
kubectl get pod -A
Error: unknown shorthand flag: 'A' in -A

tstromberg:
I don't believe `kubectl get po -A` is being deprecated any time soon.

asloan7:
What I was trying to say is kubectl get po -A does not work with older versions of kubectl, while kubectl get pods --all-namespaces works with all versions.

Issue #5329, opened 2019-09-12
kubelet: The generated config is not loaded (imagefs.available not applied)


maingoh:
Minikube generates a kubelet config file in `/var/lib/kubelet/config.yaml` with eviction thresholds at 0%.
However it doesn't seem to be loaded by kubelet.

**The exact command to reproduce the issue**:

```
$ minikube start --vm-driver=none --extra-config=kubelet.v=20
😄  minikube v1.3.1 on Ubuntu 18.04
🤹  Running on localhost (CPUs=16, Memory=32176MB, Disk=436619MB) ...
ℹ️   OS release is Ubuntu 18.04.1 LTS
🐳  Preparing Kubernetes v1.15.2 on Docker 19.03.2 ...
    ▪ kubelet.resolv-conf=/run/systemd/resolve/resolv.conf
💾  Downloading kubeadm v1.15.2
💾  Downloading kubelet v1.15.2
🚜  Pulling images ...
🚀  Launching Kubernetes ... 
🤹  Configuring local host environment ...

⚠️  The 'none' driver provides limited isolation and may reduce system security and reliability.
⚠️  For more information, see:
👉  https://minikube.sigs.k8s.io/docs/reference/drivers/none/

⌛  Waiting for: apiserver proxy etcd scheduler controller dns
🏄  Done! kubectl is now configured to use "minikube"

$ sudo journalctl -u kubelet --since=today | grep imagefs.available
[...]

PR #5697, opened 2019-10-22
kubelet: Pass --config to use kubeadm generated configuration

nanikjava:
#5329

Kubelet startup parameters does not include the --config flag. This flag pass the
location of the configuration file. During minikube startup process this file
is copied over to the VM. Fix test cases.


nanikjava:
Tested for both kvm2 and vbox. Test done by changing the value inside _pkg/minikube/bootstrapper/kubeadm/templates.go_

> evictionHard:
>   nodefs.available: "0%"
>   nodefs.inodesFree: "0%"
>   imagefs.available: "0%"
> `))

Following is kubelet log when reading the config.yaml (using 0,0,0 as evictionHard value)

`container_manager_linux.go:270] Creating Container Manager object based on Node Config: {RuntimeCgroupsName: SystemCgroupsName: KubeletCgroupsName: ContainerRuntime:docker CgroupsPerQOS:true CgroupRoot:/ CgroupDriver:cgroupfs KubeletRootDir:/var/lib/kubelet ProtectKernelDefaults:false NodeAllocatableConfig:{KubeReservedCgroupName: SystemReservedCgroupName: EnforceNodeAllocatable:map[pods:{}] KubeReserved:map[] SystemReserved:map[] HardEvictionThresholds:[]} QOSReserved:map[] ExperimentalCPUManagerPolicy:none ExperimentalCPUManagerReconcilePeriod:10s ExperimentalPodPidsLimit:-1 EnforceCPULimits:true CPUCFSQuotaPeriod:100ms ExperimentalTopologyManagerPolicy:none}`

Following is kubelet log when reading the config.yaml (using 10,10,10 as evictionHard value)

`container_manager_linux.go:270] Creating Container Manager object based on Node Config: {RuntimeCgroupsName: SystemCgroupsName: KubeletCgroupsName: ContainerRuntime:docker CgroupsPerQOS:true CgroupRoot:/ CgroupDriver:cgroupfs KubeletRootDir:/var/lib/kubelet ProtectKernelDefaults:false NodeAllocatableConfig:{KubeReservedCgroupName: SystemReservedCgroupName: EnforceNodeAllocatable:map[pods:{}] KubeReseator:LessThan Value:{Quantity:<nil> Percentage:0.1} GracePeriod:0s MinReclaim:<nil>} {Signal:nodefs.available Operator:LessThan Value:{Quantity:<nil> Percentage:0.1} GracePeriod:0s MinReclaim:<nil>} {Signal:nodefs.inodesFree Operator:LessThan Value:{Quantity:<nil> Percentage:0.1} GracePeriod:0s MinReclaim:<nil>}]} QOSReserved:map[] ExperimentalCPUManagerPolicy:none ExperimentalCPUManagerReconcilePeriod:10s ExperimentalPodPidsLimit:-1 EnforceCPULimits:true CPUCFSQuotaPeriod:100ms ExperimentalTopologyManagerPolicy:none}`

tstromberg:
@minikube-bot OK to test

thomas-riccardi:
It seems some (if not most) kubelet parameters are now duplicated in kubelet commandline and config.yaml file.
It's probably better to have the values only once: either cmdline or config file.

nanikjava:
> It seems some (if not most) kubelet parameters are now duplicated in kubelet commandline and config.yaml file.
> It's probably better to have the values only once: either cmdline or config file.
> 
> I'm not even sure if all the values are the same:
> 
> * the cmdline is built in `/pkg/minikube/bootstrapper/kubeadm/versions.go`
> * the kubelet config file is defined in `/pkg/minikube/bootstrapper/kubeadm/templates.go`, generated in `/pkg/minikube/bootstrapper/kubeadm/kubeadm.go` as a `kubeadmCfg`, but it's not really kubeadmin config (it's kubelet config), isn't it? (and there is a `kubeletCfg` next to it, but it contains other things... I'm lost :) )

Can you please open a separate ticket issue for the above so we can take a look at it separately so it's easier to track.

Thanks

thomas-riccardi:
> Can you please open a separate ticket issue for the above so we can take a look at it separately so it's easier to track.
> 
> Thanks

I opened #5719, but it may be better to resolve the issue *before* merging this current PR (#5697), as it may introduce regressions if the parameters are different.

nanikjava:
> > Can you please open a separate ticket issue for the above so we can take a look at it separately so it's easier to track.
> > Thanks
> 
> I opened #5719, but it may be better to resolve the issue _before_ merging this current PR (#5697), as it may introduce regressions if the parameters are different.

Thanks @thomas-riccardi 

Will wait for feedback from the reviewers about this.

tstromberg:
> > It seems some (if not most) kubelet parameters are now duplicated in kubelet commandline and config.yaml file.
> > It's probably better to have the values only once: either cmdline or config file.
> > I'm not even sure if all the values are the same:
> > 
> > * the cmdline is built in `/pkg/minikube/bootstrapper/kubeadm/versions.go`
> > * the kubelet config file is defined in `/pkg/minikube/bootstrapper/kubeadm/templates.go`, generated in `/pkg/minikube/bootstrapper/kubeadm/kubeadm.go` as a `kubeadmCfg`, but it's not really kubeadmin config (it's kubelet config), isn't it? (and there is a `kubeletCfg` next to it, but it contains other things... I'm lost :) )
> 
> Can you please open a separate ticket issue for the above so we can take a look at it separately so it's easier to track.
> 
> Thanks

This overlap is intentional, because at first, very few kubelet flags were exposed by kubeadm. Flags always take precedence over configuration files. Here is my guidance:

- If the kubelet configuration option can be set with kubeadm, add them to `kubeadm/templates.go` (the kubeadm API is more stable across versions)
- Otherwise, add it to `kubeadm/versions.go`

Unless someone sees something I'm not, I think the conflict is OK as is. User-provided flags will still take precedence.

tstromberg:
FYI, I think this PR is doing the right thing, and I would like to use it to solve #5751 correctly, by asking kubeadm to program kubelet for setting set file limits.

PR #5705, opened 2019-10-23
Fix comment for go lint

Jarvis-Zhou:
minikube/pkg/minikube/config/profile.go
```
Line 47: warning: comment on exported function ProfileNameInReservedKeywords should be of the form "ProfileNameInReservedKeywords ..." (golint)
```


afbjorklund:
Also in #5679 - fixed the "keywords" typo as well.

PR #5711, opened 2019-10-23
Previous boilerplate fix break make test

nanikjava:
The fix change to boilerplate directory but did not change back
to the original directory.

Use pushd and popd to store the directory within the context of
the boilerplate script

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -35,6 +35,7 @@ then
     echo "= boilerplate ==========================================================="
     readonly ROOT_DIR=$(pwd)
     readonly BDIR="${ROOT_DIR}/hack/boilerplate"
+    pushd . >/dev/null
     cd ${BDIR}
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -35,6 +35,7 @@ then
     echo "= boilerplate ==========================================================="
     readonly ROOT_DIR=$(pwd)
     readonly BDIR="${ROOT_DIR}/hack/boilerplate"
+    pushd . > /dev/null
```
tstromberg: space isn't necessary after >

tstromberg:
Thank you!

medyagh:
Thanks for fixing it, @sharifelgamal was just about to send a very similar PR .  I am okay with the way it is now, I will merge to unblock others.

PR #5718, opened 2019-10-24
Refactor to remove opening browser and just return url(s)

nanikjava:
The WaitAndMaybeOpenService(..) method allow user to open the service
url in a browser when found, this create issue during testing as it's
opening browser and consuming resources.

The fix is to introduce a boolean flag allowing the caller to specify
whether to just print out the url or open a browser window

### DIFF
--- a/cmd/minikube/cmd/service.go
+++ b/cmd/minikube/cmd/service.go
@@ -17,8 +17,13 @@ limitations under the License.
 package cmd
 
 import (
+	"fmt"
 	"os"
 	"text/template"
+
+	"github.com/pkg/browser"
+
+	"k8s.io/minikube/pkg/minikube/out"
 
 	"github.com/spf13/cobra"
 	"k8s.io/minikube/pkg/minikube/cluster"
@@ -68,10 +73,23 @@ var serviceCmd = &cobra.Command{
 		if !cluster.IsMinikubeRunning(api) {
 			os.Exit(1)
 		}
-		err = service.WaitAndMaybeOpenService(api, namespace, svc,
+
+		var urlString []string
+
+		urlString, err = service.WaitForService(api, namespace, svc,
 			serviceURLTemplate, serviceURLMode, https, wait, interval)
 		if err != nil {
 			exit.WithError("Error opening service", err)
+		}
+
+		if len(urlString) != 0 {
+			out.T(out.Celebrate, "Opening kubernetes service  {{.namespace_name}}/{{.service_name}} in default browser...", out.V{"namespace_name": namespace, "service_name": svc})
+
+			for _, url := range urlString {
+				if err := browser.OpenURL(url); err != nil {
+					exit.WithError(fmt.Sprintf("browser failed to open url %s", url), err)
+				}
+			}
 		}
 	},
 }
--- a/pkg/minikube/service/service.go
+++ b/pkg/minikube/service/service.go
@@ -29,7 +29,6 @@ import (
 	"github.com/docker/machine/libmachine"
 	"github.com/golang/glog"
 	"github.com/olekukonko/tablewriter"
-	"github.com/pkg/browser"
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 	core "k8s.io/api/core/v1"
@@ -265,9 +264,11 @@ func PrintServiceList(writer io.Writer, data [][]string) {
 	table.Render()
 }
 
-// WaitAndMaybeOpenService waits for a service, and opens it when running
-func WaitAndMaybeOpenService(api libmachine.API, namespace string, service string, urlTemplate *template.Template, urlMode bool, https bool,
-	wait int, interval int) error {
+// WaitForService waits for a service, and return the urls when available
+func WaitForService(api libmachine.API, namespace string, service string, urlTemplate *template.Template, urlMode bool, https bool,
+	wait int, interval int) ([]string, error) {
+
+	var urlList []string
 	// Convert "Amount of time to wait" and "interval of each check" to attempts
 	if interval == 0 {
 		interval = 1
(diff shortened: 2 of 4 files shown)


pkg/minikube/service/service.go
```diff
@@ -267,7 +267,7 @@ func PrintServiceList(writer io.Writer, data [][]string) {
 
 // WaitAndMaybeOpenService waits for a service, and opens it when running
 func WaitAndMaybeOpenService(api libmachine.API, namespace string, service string, urlTemplate *template.Template, urlMode bool, https bool,
```
tstromberg: WaitAndMaybeOpenService already has too many arguments, let's not add another.

How about removing the `browser.OpenURL` call altogether, and leaving that up to `cmd/service.go` to handle. Then this function could be renamed to a more definite `WaitForService`, and be more testable.

medyagh: Agree with tstromberg

nanikjava: Cool. Will make the changes

pkg/minikube/service/service.go
```diff
@@ -265,22 +264,24 @@ func PrintServiceList(writer io.Writer, data [][]string) {
 	table.Render()
 }
 
-// WaitAndMaybeOpenService waits for a service, and opens it when running
-func WaitAndMaybeOpenService(api libmachine.API, namespace string, service string, urlTemplate *template.Template, urlMode bool, https bool,
-	wait int, interval int) error {
+// WaitForService waits for a service, and opens it when running
```
tstromberg: Make comment accurate again

cmd/minikube/cmd/service.go
```diff
@@ -68,11 +71,16 @@ var serviceCmd = &cobra.Command{
...
 			os.Exit(1)
 		}
-		err = service.WaitAndMaybeOpenService(api, namespace, svc,
+		urlString, err := service.WaitForService(api, namespace, svc,
 			serviceURLTemplate, serviceURLMode, https, wait, interval)
 		if err != nil {
 			exit.WithError("Error opening service", err)
 		}
+
+		if len(urlString) != 0 {
```
tstromberg: `WaitAndMaybeOpenService` had the following logic for opening a URL:

`if urlMode || !isHTTPSchemedURL {`

Make sure your PR preserves the same behavior as before.

pkg/minikube/service/service.go
```diff
@@ -265,22 +264,24 @@ func PrintServiceList(writer io.Writer, data [][]string) {
...
+func WaitForService(api libmachine.API, namespace string, service string, urlTemplate *template.Template, urlMode bool, https bool,
+	wait int, interval int) (string, error) {
+
+	var urlString string
```
tstromberg: Just `url` is fine as a variable name. It generally isn't useful to repeat the type of an object in the object name.

See also https://github.com/golang/go/wiki/CodeReviewComments#variable-names

pkg/minikube/service/service_test.go
```diff
@@ -903,12 +903,12 @@ func TestWaitAndMaybeOpenService(t *testing.T) {
 				servicesMap:  serviceNamespaces,
 				endpointsMap: endpointNamespaces,
 			}
-			err := WaitAndMaybeOpenService(test.api, test.namespace, test.service, defaultTemplate, test.urlMode, test.https, 1, 0)
+			_, err := WaitForService(test.api, test.namespace, test.service, defaultTemplate, test.urlMode, test.https, 1, 0)
```
tstromberg: This returns a URL now - please test the returned value of it.

pkg/util/utils.go
```diff
@@ -200,3 +202,9 @@ func ConcatStrings(src []string, prefix string, postfix string) []string {
 	}
 	return ret
 }
+
+func OpenBrowser(urlString string) {
```
tstromberg: This function doesn't add ta lot of value, so just remove it and inline the code instead. The only thing really duplicated is the error message. 

Also, we're trying to get rid of unscoped packages like the `util` package.

cmd/minikube/cmd/service.go
```diff
@@ -68,11 +71,16 @@ var serviceCmd = &cobra.Command{
...
+			out.T(out.Celebrate, "Opening kubernetes service  {{.namespace_name}}/{{.service_name}} in default browser...", out.V{"namespace_name": namespace, "service_name": svc})
+			util.OpenBrowser(urlString)
```
tstromberg: If opening the browser fails, call `exit.WithError`

pkg/minikube/service/service_test.go
nanikjava: @tstromberg Want to bounce some thoughts. In master branch the current implementation is

```
	for _, bareURLString := range serviceURL.URLs {
		urlString, isHTTPSchemedURL := OptionallyHTTPSFormattedURLString(bareURLString, https)

		if urlMode || !isHTTPSchemedURL {
			out.T(out.Empty, urlString)
		} else {
			out.T(out.Celebrate, "Opening kubernetes service  {{.namespace_name}}/{{.service_name}} in default browser...", out.V{"namespace_name": namespace, "service_name": service})
			if err := browser.OpenURL(urlString); err != nil {
				out.ErrT(out.Empty, "browser failed to open url: {{.error}}", out.V{"error": err})
			}
		}
	}

```
which means that there is possibility that there are more than 1 url, so is it correct to also
implement string array as the returned value ?

cmd/minikube/cmd/service.go
nanikjava: The code that you are referring to is inside pkg/minikube/service/service.go and the only changes made was to send back the array of urls

pkg/minikube/service/service_test.go
nanikjava: After going through the code and the functionality the function will have to return array of url. So have made changes to both the code and test cases.

pkg/minikube/service/service.go
tstromberg: This method no longer opens a URL.

nanikjava: Comment has been updated to reflect returning url instead of opening in browser. PR title has been modified too.

tstromberg:
Thank you for sticking through the review process!

nanikjava:
> Thank you for sticking through the review process!

No worries. Thanks for being patience in reviewing my PRs, still learning :)

Issue #5723, opened 2019-10-24
kvm2 test failures: unable to stop vm, current state "Running"


medyagh:
On master I have been seeing lots and lots of fails on  "TestStartStop/group/crio"

as seen in this PR : https://storage.googleapis.com/minikube-builds/logs/5721/KVM_Linux.txt

here is one from master:

<details>
<pre>
09:20:01 --- FAIL: TestStartStop (923.84s)
09:20:01     --- FAIL: TestStartStop/group (130.00s)
09:20:01         --- FAIL: TestStartStop/group/crio (778.90s)
09:20:01             helpers.go:349: MaybeSlowParallel: Sleeping 40s to avoid start race ...
09:20:01             start_stop_delete_test.go:89: (dbg) Run:  out/minikube-linux-amd64 start -p crio-20191024T160659.448877654-27169 --alsologtostderr -v=3 --container-runtime=crio --disable-driver-mounts --extra-config=kubeadm.ignore-preflight-errors=SystemVerification --vm-driver=kvm2 
09:20:01             start_stop_delete_test.go:89: (dbg) Done: out/minikube-linux-amd64 start -p crio-20191024T160659.448877654-27169 --alsologtostderr -v=3 --container-runtime=crio --disable-driver-mounts --extra-config=kubeadm.ignore-preflight-errors=SystemVerification --vm-driver=kvm2 : (6m35.799116136s)
09:20:01             start_stop_delete_test.go:100: (dbg) Run:  kubectl --context crio-20191024T160659.448877654-27169 create -f testdata/busybox.yaml
09:20:01             start_stop_delete_test.go:100: (dbg) Done: kubectl --context crio-20191024T160659.448877654-27169 create -f testdata/busybox.yaml: (18.026871106s)
09:20:01             start_stop_delete_test.go:105: (dbg) waiting for pods with labels "integration-test=busybox" in namespace "default" ...
09:20:01             helpers.go:244: "busybox" [5de2f98f-d0a2-4d21-8a64-45a6ca5902ca] Pending
09:20:01             helpers.go:244: "busybox" [5de2f98f-d0a2-4d21-8a64-45a6ca5902ca] Pending / Ready:ContainersNotReady (containers with unready status: [busybox]) / ContainersReady:ContainersNotReady (containers with unready status: [busybox])
09:20:01             helpers.go:244: "busybox" [5de2f98f-d0a2-4d21-8a64-45a6ca5902ca] Running
[...]

PR #5732, opened 2019-10-25
Hide innocuous viper ConfigFileNotFoundError

### LINKED ISSUE #4920: Change the error message when Config file is not available

tstromberg:
Fixes #4920
Based on the proposal by @RA489 in #4940 + review feedback

### DIFF
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -235,9 +235,11 @@ func initConfig() {
 	configPath := localpath.ConfigFile
 	viper.SetConfigFile(configPath)
 	viper.SetConfigType("json")
-	err := viper.ReadInConfig()
-	if err != nil {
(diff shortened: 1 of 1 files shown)


medyagh:
KVM fails are related to : https://github.com/kubernetes/minikube/issues/5724 and https://github.com/kubernetes/minikube/issues/5723

Vbox Linux fails are related to https://github.com/kubernetes/minikube/issues/5724

PR #5738, opened 2019-10-25
small travis test changes

sharifelgamal:
1. Consolidating boilerplate and lint tests under one heading
2. Add `make` to travis tests as a sanity check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
     - language: go
       name: Cross Build
       go: 1.12.12
-      script: make cross
+      script: make
 after_success:
   - bash <(curl -s https://codecov.io/bash)
 travisBuddy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
       script: make test
 
     - language: go
-      name: Cross Build
+      name: Build
       go: 1.12.12
       script: make
 after_success:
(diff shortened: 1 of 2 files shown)


afbjorklund:
You could do three different builds (parallel) instead of doing three (four?) builds in sequence (serial)

The "cross" target is basically doing a straight-forward `go build` invocation of each GOOS/GOARCH.

```make
.PHONY: cross
cross: minikube-linux-amd64 minikube-linux-arm64 minikube-darwin-amd64 minikube-windows-amd64.exe
```
This would save a couple of minutes waiting, even if it has to generate some code three times.

Analyzing the modules takes a lot of time, but that cache has to be in place anyway.

## Before 
[#9293](https://travis-ci.org/kubernetes/minikube/builds/603063390) passed
 Ran for 5 min 29 sec
 Total time 12 min 26 sec

## After 
[#9294](https://travis-ci.org/kubernetes/minikube/builds/603063900) passed
 Ran for 7 min 52 sec
 Total time 17 min 58 sec

Also the commit is changing a bunch of other stuff, like lint and buddy. And not only the `make cross`.

afbjorklund:
@sharifelgamal : if we had bigger servers, we could build **all** make cross targets in parallel (`make -j 4 cross`). That way they could share some of the overhead of downloading all the modules and such...

Unfortunately our Travis VMs are really small (2 vCPU), and a take long time (20-50s) to boot up.

https://docs.travis-ci.com/user/reference/overview/

But we _definitely_ need to introduce at least a build step before merge, even if it is not the full cross.

Not rebuilding everything is 10 times faster (#5749)

PR #5749, opened 2019-10-27
Use faster build for the default minikube target

afbjorklund:
When we are doing the "cross" compile targets, we
force rebuilding all packages (up-to-date or not).

For the default native target "out/minikube", we
can do a much faster build (10x!), for development.

Leaving all the cross target to rebuild, as before.

Also the _test.go files are only used for unit tests.

----

On my machine it takes about 50 seconds, to rebuild all the 645 packages...

```
real	0m47.801s
user	2m36.732s
sys	0m8.620s
```

Doing an incremental build takes only 6 seconds, and is **much** more usable.

```
real	0m6.077s
user	0m7.368s
sys	0m0.928s
```

Interesting flags to `go build`:

```
The build flags are shared by the build, clean, get, install, list, run,
and test commands:

	-a
		force rebuilding of packages that are already up-to-date.
	-n
		print the commands but do not run them.
	-v
		print the names of packages as they are compiled.
	-x
		print the commands.
```

I used -v to get the 645 lines.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,9 @@ CMD_SOURCE_DIRS = cmd pkg
 SOURCE_DIRS = $(CMD_SOURCE_DIRS) test
 SOURCE_PACKAGES = ./cmd/... ./pkg/... ./test/...
 
+SOURCE_GENERATED = pkg/minikube/assets/assets.go pkg/minikube/translate/translations.go
+SOURCE_FILES = $(shell find $(CMD_SOURCE_DIRS) -type f -name "*.go" | grep -v _test.go)
+
 # kvm2 ldflags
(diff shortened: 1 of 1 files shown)


tstromberg:
@minikube-bot OK to test

tstromberg:
That's a really impressive improvement. I look forward to merging this :)

medyagh:
the mac os test failures I seen on other PRs, not related to this PR

afbjorklund:
Also had broken the default build in Docker, i.e. `BUILD_IN_DOCKER=y make`

medyagh:
afbjorklund  please merge this when you feel ready !

PR #5760, opened 2019-10-29
integration: Change --postmortem-logs to --postmortem-log-lines, default to 5

tstromberg:
5 is just enough to hint that there was an OOM, without flooding you with crap log lines.

### DIFF
--- a/test/integration/helpers.go
+++ b/test/integration/helpers.go
@@ -175,9 +175,9 @@ func Cleanup(t *testing.T, profile string, cancel context.CancelFunc) {
 // CleanupWithLogs cleans up after a test run, fetching logs and deleting the profile
 func CleanupWithLogs(t *testing.T, profile string, cancel context.CancelFunc) {
 	t.Helper()
-	if t.Failed() && *postMortemLogs {
+	if t.Failed() && *postMortemLogLines > 0 {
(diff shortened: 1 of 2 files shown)


medyagh:
I am curious what is the current number ? is it unlimited ? if yes, how about give it a few more lines, maybe 13?

medyagh:
@minikube-bot OK to test

tstromberg:
> I am curious what is the current number ? is it unlimited ? if yes, how about give it a few more lines, maybe 13?

The current limit is 100. The post-mortem logs in our integration tests are mostly duplicated and useless at this point, but handy to leave around for a few edge cases.

medyagh:
> The current limit is 100. The post-mortem logs in our integration tests are mostly duplicated and useless at this point, but handy to leave around for a few edge cases.

is there a way to grep that 100 lines for some keywords ? such as Fail or Error ?

tstromberg:
> > The current limit is 100. The post-mortem logs in our integration tests are mostly duplicated and useless at this point, but handy to leave around for a few edge cases.
> 
> is there a way to grep that 100 lines for some keywords ? such as Fail or Error ?

Good idea. Will re-propose with that instead.

Issue #5398, opened 2019-09-18
PR bot: Add CPU load impact


tstromberg:
As part of the roadmap item "Reduce guest VM overhead by 50%", I would like to see an integration test that both our CI and developers can run that generates:

- A graph of load average impact for minikube start + 60 seconds of idle afterwards.
- An approximate idle load impact score, for instance: (mean minikube idle average - mean pre-start average). 

An additional enhancement would be generating the statistics in a way that demonstrates VM overhead versus VM+Kubernetes overhead.

tstromberg:
Related: #3207

It would be nice to see this mature to eventually give us feedback on whether or not a PR has a negative or positive performance improvement.

tstromberg:
Also related: #4622

priyawadhwa:
As part of this, we are adding a PR bot to:
- [ ] Compare time for `minikube start` across three runs
- [ ] Compare binary sizes via [build-size](https://github.com/afbjorklund/build-size)

sharifelgamal:
Opened #6095 for the binary size diff.

PR #5773, opened 2019-10-29
Add skeleton for performance analysis service

priyawadhwa:
This service runs in a VM and will be responsible for:

1. Determining which PRs need to be analyzed
2. Running mkcmp against each PR
3. Commenting the results on the PR via minikube-bot

This PR also moves the mkcmp code into its own directory.

Ref #5398


priyawadhwa:
@tstromberg for sure, done!

PR #5782, opened 2019-10-30
Improve parallel start scheduling and autoset parallelism

tstromberg:
* Improve parallel scheduling by guaranteeing a 30 second offset rather than guessing how long to wait based on the number of tests seen. Previously, two clusters might start simultaneously if they called MaybeSlowParallel at T-offset from one another.

* Hard-coding parallelism does not necessarily fit every machine, so let Go determine the right number based on the cores available.

### DIFF
--- a/test/integration/addons_test.go
+++ b/test/integration/addons_test.go
@@ -36,7 +36,8 @@ import (
 
 // TestAddons tests addons that require no special environment -- in parallel
 func TestAddons(t *testing.T) {
-	MaybeSlowParallel(t)
+	MaybeParallel(t)
+	WaitForStartSlot(t)
 	profile := UniqueProfileName("addons")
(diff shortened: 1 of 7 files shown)


medyagh:
on the KVM log it says

"uptime:     05:17:54 up 5 days,  7:16,  0 users,  load average: 1.91, 5.92, 8.45"

how could it be up for 5 days ? I thought we had daily restarts !!!

tstromberg:
> on the KVM log it says
> 
> "uptime: 05:17:54 up 5 days, 7:16, 0 users, load average: 1.91, 5.92, 8.45"
> 
> how could it be up for 5 days ? I thought we had daily restarts !!!

Not all of our nodes had the cleanup script installed. This is being addressed by #5648

PR #5790, opened 2019-10-30
service: fix --url mode, add integration tests

### LINKED ISSUE #5789: minikube service opens in browser when it should not

minikube service --namespace mynamespace --url mynodeportservice

In versions of minikube through 1.3.1, this output the service URL to stdout, and _did not_ open the URL in the default browser.  In 1.5.1, it outputs the URL and the "🎉  Opening kubernetes service ..." message to stdout, and it opens in the default browser.

With --url=false, it does not open in the browser, but now it outputs an ascii-art table instead of the plain URL.
With --url=true, it behaves the same as with --url.

Based on finding #5699, I tried using --format="{{ .IP }}".  That leads to even greater insanity... it emits the message about opening the service in the default browser, but then fails to do so because it tries to open a file in the current directory matching the IP address, and errors out.

Please restore the older behavior so scripts can get the URL for a service, and fix --format to _not_ try to open the resulting string in the default browser.
</details>

</details>

Mojave
minikube 1.5.1 installed via homebrew

tstromberg:
The service command was opening URL's, even if `--url` mode was specified. This was caused by a subtle regression in #5718

Fixes #5789

### DIFF
--- a/cmd/minikube/cmd/service.go
+++ b/cmd/minikube/cmd/service.go
@@ -83,7 +83,7 @@ var serviceCmd = &cobra.Command{
 		for _, u := range urls {
 			_, err := url.Parse(u)
 			if err != nil {
-				glog.Warningf("unable to parse %s: %v (will not open)", u, err)
+				glog.Warningf("failed to parse url %q: %v (will not open)", u, err)
 				out.String(fmt.Sprintf("%s\n", u))
 				continue
 			}
--- a/cmd/minikube/cmd/service.go
+++ b/cmd/minikube/cmd/service.go
@@ -18,17 +18,18 @@ package cmd
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"text/template"
 
+	"github.com/golang/glog"
 	"github.com/pkg/browser"
-
-	"k8s.io/minikube/pkg/minikube/out"
-
 	"github.com/spf13/cobra"
+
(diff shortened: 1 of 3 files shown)


cmd/minikube/cmd/service.go
```diff
@@ -74,21 +75,26 @@ var serviceCmd = &cobra.Command{
...
-			serviceURLTemplate, serviceURLMode, https, wait, interval)
+		urls, err := service.WaitForService(api, namespace, svc, serviceURLTemplate, serviceURLMode, https, wait, interval)
 		if err != nil {
 			exit.WithError("Error opening service", err)
 		}
 
-		if len(urlString) != 0 {
-			out.T(out.Celebrate, "Opening kubernetes service  {{.namespace_name}}/{{.service_name}} in default browser...", out.V{"namespace_name": namespace, "service_name": svc})
+		for _, u := range urls {
+			_, err := url.Parse(u)
```
retracile: If --format has been passed, will `u` contain that formatted value?  If so, this would imply that `--urls --format {{.IP}}` would generate a warning, when it really shouldn't.  If my understanding is correct, I'd think you'd want the `if serviceURLMode` to come before the `url.Parse(u)` call.

retracile: Based on some local testing, apparently that's not an issue

PR #5794, opened 2019-10-30
Fix pod list retry, reset start timer

tstromberg:
Because pod.List was returning an error, it wasn't being retried.

Addresses issue seen on https://storage.googleapis.com/minikube-builds/logs/5790/none_Linux.txt - where the apiserver may be restarting with the none driver when calling start/stop.


tstromberg:
Related: #5795

Issue #2570, opened 2018-02-23
Omission of --kubernetes-version upgrades cluster to the default version


jar349:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): BUG REPORT

Please provide the following details:

**Environment**:
Mac Sierra (10.12.6)

**Minikube version** (use `minikube version`):
- **OS** (e.g. from /etc/os-release): Buildroot 2017.02
- **VM Driver** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep DriverName`): virtualbox
- **ISO version** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep -i ISO` or `minikube ssh cat /etc/VERSION`): minikube-v0.23.6.iso

**What happened**:
minikube upgraded itself from 1.8 to 1.9 when it was stopped and started

**What you expected to happen**:
don't upgrade

**How to reproduce it** (as minimally and precisely as possible):
```
minikube start --kubernetes-version v1.8.0
kubectl version
minikube stop
minikube start
kubectl version
```

r2d4:
You'll either need to supply the flag everytime or use `minikube config set kubernetes-version v1.8.0`

jar349:
Thank you for the workarounds.  If I start a kube, is it unreasonable to expect that it will stay the same until deleted and re-created, or otherwise explicitly re-configured?  If this isn't a bug, I at least consider current behavior to be non-normal.

utdrmac:
Agree with @jar349 in that if I launch minikube specifying 1.13.1, then stop, then start it back up, it should NOT auto-upgrade k8s to latest. No other software I have does this behavior; not at least without prompting/asking me first. This is very non-standard behavior and is causing **LOTS** of problems with out students.

tace:
I can't believe how stupid feature this is. No sw should ever upgrade itself without explicitly asking it! And user should ask it explicitly, no prompt even is not enough.

afbjorklund:
Might as well have opened a **new** ticket for discussing this, but I suppose we can start over here...

Currently minikube only has one single "start" command, that is used for both create and for start.
This means that there is no way to just start an existing VM, without for instance also upgrading it.

It is going to be hard to change this (`minikube start`), but we could at least make it opt-out ?
In the future we would also like to add a suspend/resume feature, if possible (with libmachine).

As stated above, the workaround for both vm-driver and kubernetes-version is using `config`.

afbjorklund:
[...]

PR #5798, opened 2019-10-31
Make Kubernetes version sticky for a cluster instead of auto-upgrading

nanikjava:
#2570 

The fix should do the following 

* Not specifying the --kubernetes-version flag means just use the currently deployed version.
* If an update is available inform the user that they may use --kubernetes-version=<version>.
* When --kubernetes-version is specifically set, upgrade the cluster.

### DIFF
--- a/test/integration/version_upgrade_test.go
+++ b/test/integration/version_upgrade_test.go
@@ -18,6 +18,7 @@ package integration
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -28,6 +29,7 @@ import (
 	"time"
 
 	"github.com/docker/machine/libmachine/state"
+
 	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/util/retry"
 
@@ -66,13 +68,13 @@ func TestVersionUpgrade(t *testing.T) {
 
 	args := append([]string{"start", "-p", profile, fmt.Sprintf("--kubernetes-version=%s", constants.OldestKubernetesVersion), "--alsologtostderr", "-v=1"}, StartArgs()...)
 	rr := &RunResult{}
-	releaseStart := func() error {
+	r := func() error {
 		rr, err = Run(t, exec.CommandContext(ctx, tf.Name(), args...))
 		return err
 	}
 
 	// Retry to allow flakiness for the previous release
-	if err := retry.Expo(releaseStart, 1*time.Second, 30*time.Minute, 3); err != nil {
+	if err := retry.Expo(r, 1*time.Second, 30*time.Minute, 3); err != nil {
 		t.Fatalf("release start failed: %v", err)
 	}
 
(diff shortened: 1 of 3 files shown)


medyagh:
@nanikjava uni tests are failing, plz check the travis logs

medyagh:
@minikube-bot OK to test

test/integration/version_upgrade_test.go
```diff
@@ -35,6 +35,113 @@ import (
...
+// Start with installing from v1.13.0 then to v1.15.0 and then try to downgrade it
+// to v1.13.0
+// Following are the test cases that the function is doing
+// 1. Start with specific version
+//    ./out/minikube start -p vupgrade --kubernetes-version=v1.13.0
+// 2. Upgrade with newer version
+//   ./out/minikube start -p vupgrade --kubernetes-version=v1.13.0
+// 3. Downgrade with older version (this will spit out error and test the error message)
+//   ./out/minikube start -p vupgrade --kubernetes-version=v1.13.0
+func TestDifferentVersionUpgrade(t *testing.T) {
```
tstromberg: In the interest of keeping integration tests fast, I think this function can be safely removed as it does not add much over  `TestVersionUpgrade`.

If you really want to assert that we don't break the `downgrades are always failures` case, I would suggest adding it to `TestVersionUpgrade`, where it won't incur the penalty of creating an additional cluster.

nanikjava: No worries. Will refactor the code. Thanks

nanikjava: > In the interest of keeping integration tests fast, I think this function can be safely removed as it does not add much over `TestVersionUpgrade`.
> 
> If you really want to assert that we don't break the `downgrades are always failures` case, I would suggest adding it to `TestVersionUpgrade`, where it won't incur the penalty of creating an additional cluster.

Gone through the `TestVersionUpgrade(..) `logic and even though in terms of what it is trying to accomplish (checking kubernetes version updrade checking) is similar but the checking that it is performing is less than what `TestDifferentVersionUpgrade(..) `is doing. The `TestDifferentVersionUpgrade(..)` checks the output to make sure that the process does execute properly while `TestVersionUpgrade(..) `does not do any kind of checking as it just 'assume' that the process will start.

My thought behind this is that by having test cases that checks the output will assure that there are no behavioural changes in the code. 

If this is not valid then happy to remove to just integrate the test case (3) for checking the downgrade to lower version to `TestVersionUpgrade(..)`

tstromberg: Rather than implementing the test twice, let's improve `TestVersionUpgrade` so that both tests benefit.  The only thing missing is:

- A version check (see my later comments)
- A downgrade check

The downgrade check can just come at the end of the test. No need to assert the output, just check that start fails if you try to downgrade to the same version that the test started with.

test/integration/version_upgrade_test.go
```diff
@@ -35,6 +35,113 @@ import (
...
+				if err != nil {
+					t.Fatalf("release start failed: %v", err)
+				}
+
+				// grab the Stdout for checking...
+				ss := rr.Stdout.String()
+
+				// fail if it does not contain the string
+				if !strings.Contains(ss, tc.message) {
+					t.Fatalf("test failed as no message %s detected", tc.message)
```
tstromberg: The stdout of 'minikube start' is not designed to be parsed. Do you mind checking for the resulting apiserver version of the cluster that was started instead?

Something like:

```

	j, err := exec.Command(path, "version", "--client", "--output=json").Output()
	if err != nil {
		return errors.Wrap(err, "exec")
	}

	cv := struct {
		ServerVersion struct {
			GitVersion string `json:"gitVersion"`
		} `json:"serverVersion"`
	}{}
	err = json.Unmarshal(j, &cv)
	if err != nil {
		return errors.Wrap(err, "unmarshal")
	}
```

Something like what we do at:

https://github.com/kubernetes/minikube/blob/1c46502f358e791c7a7a1ce708c7ff63480f3697/cmd/minikube/cmd/start.go#L519

test/integration/version_upgrade_test.go
```diff
@@ -35,6 +35,113 @@ import (
...
+		tests := []struct {
+			name                 string
+			message              string
+			kubernetesversion    string
+			expectstartupfailure bool
+		}{
+			{"fresh install kubernetes v1.13.0", "v1.13.0", "v1.13.0", false},
+			{"upgrade kubernetes to v 1.14.0", "v1.14.0", "v1.14.0", false},
+			{"downgrade kubernetes to v 1.12.0", "Non-destructive downgrades are not supported", "v1.12.0", true},
+		}
```
tstromberg: Please don't hardcode particular Kubernetes versions in tests when possible - they get to be a pain to maintain when we drop old versions.

test/integration/version_upgrade_test.go
nanikjava: I've updated refactor the code as per the suggested comment. `TestVersionUpgrade` now contains 2 additional test cases.

PR #5805, opened 2019-10-31
Store kubernetes clientset when waiting for apiserver

priyawadhwa:
While looking at pprof data, I noticed that the line:

```go
client, err = k.client(k8s)
```

takes around .9 seconds to complete. Previously, we were calling it on
every iteration of the waitForApiServer loop. Instead, if we store the clientset once
we have it, we should be able to speed up this function.

During local testing, this optimization reduced time spent waiting for
the apiserver by 13 seconds on average.


medyagh:
wow ! what a great finding !

tstromberg:
Integration test failures appear to be unrelated flakes. Thank you!

Issue #3963, opened 2019-03-25
hyperv: cannot use "fe80::215:5dff:fe45:3110" as the bind address for the API Server (IPv6)


xtracoder:
**Problem**

minikube start fails with
cannot use "fe80::215:5dff:fe45:3110" as the bind address for the API Server

**command-lines**

```
minikube config set vm-driver hyperv && minikube config set hyperv-virtual-switch MyMiniKube
minikube start --alsologtostderr --logtostderr
```

**The full output of the command that failed**

```
C:\Sdk\Cloud\Minikube>minikube config set vm-driver hyperv && minikube config set hyperv-virtual-switch MyMiniKube
!   These changes will take effect upon a minikube delete and then a minikube start

C:\Sdk\Cloud\Minikube>minikube start --alsologtostderr --logtostderr
I0325 23:45:43.917153    5632 notify.go:121] Checking for updates...
o   minikube v0.35.0 on windows (amd64)
I0325 23:45:44.059218    5632 start.go:582] Saving config:
{
    "MachineConfig": {
        "MinikubeISO": "https://storage.googleapis.com/minikube/iso/minikube-v0.35.0.iso",
        "Memory": 2048,
        "CPUs": 2,
        "DiskSize": 20000,
        "VMDriver": "hyperv",
        "ContainerRuntime": "docker",
        "HyperkitVpnKitSock": "",
        "HyperkitVSockPorts": [],
        "XhyveDiskDriver": "ahci-hd",
        "DockerEnv": null,
        "InsecureRegistry": null,
        "RegistryMirror": null,
        "HostOnlyCIDR": "192.168.99.1/24",
        "HypervVirtualSwitch": "MyMiniKube",
        "KvmNetwork": "default",
        "DockerOpt": null,
        "DisableDriverMounts": false,
        "NFSShare": [],
        "NFSSharesRoot": "/nfsshares",
        "UUID": "",
        "GPU": false,
        "NoVTXCheck": false
    },
    "KubernetesConfig": {
        "KubernetesVersion": "v1.13.4",
        "NodeIP": "",
        "NodePort": 8443,
        "NodeName": "minikube",
        "APIServerName": "minikubeCA",
        "APIServerNames": null,
        "APIServerIPs": null,
        "DNSDomain": "cluster.local",
        "ContainerRuntime": "docker",
        "CRISocket": "",
        "NetworkPlugin": "",
        "FeatureGates": "",
        "ServiceCIDR": "10.96.0.0/12",
[...]

PR #5841, opened 2019-11-06
Limit only IPv4 address used for API server with Hyper-V driver

laozc:
Signed-off-by: Zhongcheng Lao <Zhongcheng.Lao@microsoft.com>
Fixes: #3963

This is to fix the issue by only asking the Hyper-V driver to return IPv4 address for minikube.
Note: The change to Hyper-V driver is still under review. Will update the dependency once the change is merged in driver.


tstromberg:
> I am not so sure this should be handled by the driver as it would apply to ALL hypervisors, although you have only seen this now for HyperV as the Default Switch or External Switch can return an IPv6 address in case of an error. This has also been observed for others, like VirtualBox, etc.

Is there an alternative approach you recommend?

tstromberg:
@gbraad - I'm not sure how else to hint to the driver to return an IPv6 network for DHCP. 

@laozc - Sounds reasonable, but I would like to wait to see if this can get merged into the `machine-drivers` fork first. Is there a PR for this change that you can link to for status updates?

laozc:
@tstromberg Thanks. The dependency change is just for quick verification when the machine-drivers change (machine-drivers/machine#23) has not been changed. I'm still working on verifying some of the IPv6 stack behavior in the Hyper-V driver. Once it's done I will push for a new change for both PR.

gbraad:
You can disable IPv6 as DHCP using:

```
PS> Set-NetIPInterface -InterfaceAlias 'vEthernet (Default Switch)' -AddressFamily IPv6 -dhcp disable
```

This way, only IPv4 will be assigned to a host on the Default Switch

gbraad:
> @gbraad - I'm not sure how else to hint to the driver to return an IPv6 network for DHCP.

That is kind of the problem. This issue could occur for any hypervisor that is used, if the network layer is configured to use IPv6. It is a generic issue in how libmachine does not handle multipe IP address assignment to the VM. At the moment it just return the first address from the first interface. The question would be use a workaround in code, to accomodate a possible timing issue or error in the Hyper-V networking stack? Or in this case instruct people to configure the network to prevent his to happen? Or refactor parts of the machine code to allow the use of multiple network addresses. I agree, the work around seems the simplest, but at the moment the PR that is proposed does not match default behaviour and might cause unexpected other issues. So far, disabling the IPv6 assignment on the switch sounds like in line with what Libvirt and Hyperkit do at the moment...

... but, for how long? Would we in the end start putting preferred options in each and every driver we use? It would sounds more logical to solve the overall issue in that case and do the final selection in the consumer.

gbraad:
I took some more time and went over issues for Minikube on Windows, and noticed a pattern of network related issues. For instance Windows 10 Enterprise does not come with a Default Switch option, but this is now assumed as the fallback option in your codebase (as the machine-drivers/machine was made to default to this switch). To avoid this, in CRC we have the option of a named alternative, that is hardcoded by name, and when not found we try to detect the Default Switch. This happens before the VM is created.

Also, to solve #3963, which is caused by the IPv6 configuration, to use the Default Switch, as it might prevent misconfigurations. However, while it might ease the deployment, this is not a worry-free solution. It is based on the ICS Internet Connection Sharing option and so far I have seen issues with unanswered DHCPRequests, re-use not being tolerated as a DHCPOffer is expected (therefore any communication might fail), and the fact that the DHCP addressfamily implementations seem to operate as two different servers, which could cause an issue with IPv6 being assigned first and/or IPv4 never being offered at all. So the code suggested to machine is only partially solving this in case the timing might have offered both but in the wrong order.

tstromberg:
Please re-propose when we can do this without switching to a new forked machine driver.

PR #5848, opened 2019-11-06
Add retry to SSH connectivity check

### LINKED ISSUE #5834: Add retry logic to minikub SSH connectivity check
minikube start

* minikube v1.5.2 on Microsoft Windows 7 Enterprise Service Pack 1 6.1.7601 Build 7601
  - MINIKUBE_VERSION=v1.5.2
* Automatically selected the 'virtualbox' driver
    > minikube-v1.5.1.iso.sha256: 65 B / 65 B [--------------] 100.00% ? p/s 0s
    > minikube-v1.5.1.iso: 143.76 MiB / 143.76 MiB [-] 100.00% 5.96 MiB p/s 24s
* Creating virtualbox VM (CPUs=4, Memory=4096MB, Disk=20000MB) ...
* Found network options:
  - HTTP_PROXY=http://10.43.216.8:8080
  - HTTPS_PROXY=http://10.43.216.8:8080
  - NO_PROXY=localhost,valmante,minikube,.gemalto.com,127.0.0.1,10.10.154.37,192.168.99.100
  - http_proxy=http://10.43.216.8:8080
  - https_proxy=http://10.43.216.8:8080
  - no_proxy=localhost,valmante,minikube,.gemalto.com,127.0.0.1,10.10.154.37,192.168.99.100
X minikube is unable to connect to the VM: dial tcp 192.168.99.100:22: connectex: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.

This is likely due to one of two reasons:

- VPN or firewall interference
- virtualbox network configuration issue

Suggested workarounds:

- Disable your local VPN or firewall software
- Configure your local VPN or firewall to allow access to 192.168.99.100
- Restart or reinstall virtualbox
- Use an alternative --vm-driver
panic: exit status 74
</details>

**The operating system version**: Windows 7

Hello, it seems the trySSH method in validateNetwork in start.go seems a little bit aggressive when testing connectivity to VM. Immediately after the failure, I tried minikube ssh but it was not responding. When I tried 10s later, everything was working well

```
user@FSP4KC2 ~
$ ssh -l docker 192.168.99.100
^C

[wait 10 seconds]

cvila@FSP4KC2 ~
$ ssh -l docker 192.168.99.100
Warning: Permanently added '192.168.99.100' (ECDSA) to the list of known hosts.
docker@192.168.99.100's password:
                         _             _
            _         _ ( )           ( )

$ exit
logout
Connection to 192.168.99.100 closed.
```

hypnoglow:
Added retry to the SSH connectivity check on "start", starting from 1 second and up to a total of 10 seconds of attempts. Also added the 3-second timeout to dialer.

Fixes #5834

---

- Is a 10-second timeout enough for the connectivity check?
- Is a 3-second timeout suitable for dialer timeout?

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1046,8 +1046,19 @@ func validateNetwork(h *host.Host, r command.Runner) string {
 
 func trySSH(h *host.Host, ip string) {
 	sshAddr := fmt.Sprintf("%s:22", ip)
-	conn, err := net.Dial("tcp", sshAddr)
-	if err != nil {
+
+	dial := func() (err error) {
+		d := net.Dialer{Timeout: 3 * time.Second}
+		conn, err := d.Dial("tcp", sshAddr)
+		if err != nil {
+			out.WarningT("Unable to verify SSH connectivity: {{.error}}. Will retry...", out.V{"error": err})
+			return err
+		}
+		_ = conn.Close()
+		return nil
+	}
+
+	if err := retry.Expo(dial, time.Second, 13*time.Second); err != nil {
 		exit.WithCodeT(exit.IO, `minikube is unable to connect to the VM: {{.error}}
 
 This is likely due to one of two reasons:
@@ -1062,7 +1073,6 @@ Suggested workarounds:
 - Restart or reinstall {{.hypervisor}}
 - Use an alternative --vm-driver`, out.V{"error": err, "hypervisor": h.Driver.DriverName(), "ip": ip})
 	}
-	defer conn.Close()
 }
 
 func tryLookup(r command.Runner) {


cmd/minikube/cmd/start.go
```diff
@@ -1030,8 +1030,18 @@ func validateNetwork(h *host.Host, r command.Runner) string {
...
+		dialer := net.Dialer{Timeout: 3 * time.Second}
+		conn, err := dialer.Dial("tcp", sshAddr)
+		if err != nil {
+			return err
+		}
+		_ = conn.Close()
+		return nil
+	}
+
+	if err := retry.Expo(dial, time.Second, 10*time.Second); err != nil {
```
medyagh: on a second thought, lets increase this to 13, since 10 seems to be the cutting time. give it a with of wiggle room.

lets change time.Second to 300 milliseconds.

cmd/minikube/cmd/start.go
```diff
@@ -1030,8 +1030,18 @@ func validateNetwork(h *host.Host, r command.Runner) string {
 
 func trySSH(h *host.Host, ip string) {
 	sshAddr := fmt.Sprintf("%s:22", ip)
-	conn, err := net.Dial("tcp", sshAddr)
-	if err != nil {
+
+	dial := func() (err error) {
+		dialer := net.Dialer{Timeout: 3 * time.Second}
```
medyagh: lets rename this to d, since it is such a short lived variable

cmd/minikube/cmd/start.go
tstromberg: I personally feel 250ms is really aggressive for an issue that should never happen.

cmd/minikube/cmd/start.go
```diff
@@ -1030,8 +1030,18 @@ func validateNetwork(h *host.Host, r command.Runner) string {
...
+		conn, err := dialer.Dial("tcp", sshAddr)
+		if err != nil {
+			return err
```
tstromberg: Do you mind logging an error, perhaps with out.WarningT? This error should never occur, and it might be wise to give the user a hint that VM connectivity is apparently unstable.

cmd/minikube/cmd/start.go
medyagh: I am okay with 1 second, but the timeout time needs to be increased from 10

cmd/minikube/cmd/start.go
hypnoglow: - Should this error really never happen, or may it still occur in some rare cases due to slow startups or some reasons as in #5834? I mean, do we really want a warning here, or is it better to make the retry transparent from the user point of view? If, for example, after one or two retries (so in around 3 seconds) the dial would succeed.

- If we still want to log this, do we use `out.WarningT` or `out.T(out.Resetting, ...` like in `startHost` func? 
  - What log message would describe this clear enough? `Unable to verify SSH connectivity, will retry...` - is that ok? Please correct me if need.

cmd/minikube/cmd/start.go
hypnoglow: Changed timeout to 13 seconds, keeping the interval at 1 second.

cmd/minikube/cmd/start.go
hypnoglow: Done, thanks for tip

cmd/minikube/cmd/start.go
tstromberg: As far as I can tell, this condition should not regularly happen - but certainly could in a rare case. By the time `verifyNetwork` is called, minikube has already successfully ssh'd in, populated certificates, synced time (see `configureHost` in `cluster.go`). The catch here is that some drivers allow ssh access in a way that does not involve the minikube IP or virtual interface to be available.

The error message you mentioned sounds good, as does out.WarningT. I want folks to be very aware that their network connection may be unstable.

hypnoglow: Thanks for the detailed explanation! Much clearer now for me.

I've added error logging.

medyagh:
@hypnoglow this PR seems to have significally added to the start time of minikube !  check https://github.com/kubernetes/minikube/pull/5848#issuecomment-553570792

medyagh:
@hypnoglow I am curious, can we investigate if the Retry Logic, waits First the initial delay time, Before even starting ? or does it do the Wait after first try ? 

(admitibly I refactored that part of code but I haven't had chance to verify it's behaviour)

hypnoglow:
> @hypnoglow this PR seems to have significally added to the start time of minikube ! check #5848 (comment)

How can we be sure that this PR is related? `Preparing Kubernetes` log time did not change, and this happens **after** network validation occurs:

- https://github.com/kubernetes/minikube/blob/33293ed689/cmd/minikube/cmd/start.go#L336 `startMachine` is called, which calls `validateNetwork`, which calls `trySSH`
- https://github.com/kubernetes/minikube/blob/33293ed689/cmd/minikube/cmd/start.go#L340 `showVersionInfo` is called, which prints log `Preparing Kubernetes`...

Am I understanding this correctly?

BTW why there is no log time for `Waiting for: apiserver` in the comment?

tstromberg:
Thank you!

Issue #5828, opened 2019-11-04
Retriable failure: create: precreate: exec: "docker": executable file not found in $PATH


a1exus:
Hello World!

I'm trying to follow [Linux | minikube](https://minikube.sigs.k8s.io/docs/start/linux/), yet running into following issue:

I went w/ "None (bare-metal)" for "Hypervisor Setup":

_start_:
```
# minikube start --vm-driver=none
😄  minikube v1.5.2 on Centos 8.0.1905
🤹  Running on localhost (CPUs=8, Memory=7724MB, Disk=466979MB) ...
🔄  Retriable failure: create: precreate: exec: "docker": executable file not found in $PATH
🤹  Running on localhost (CPUs=8, Memory=7724MB, Disk=466979MB) ...
🔄  Retriable failure: create: precreate: exec: "docker": executable file not found in $PATH
🤹  Running on localhost (CPUs=8, Memory=7724MB, Disk=466979MB) ...
🔄  Retriable failure: create: precreate: exec: "docker": executable file not found in $PATH
🤹  Running on localhost (CPUs=8, Memory=7724MB, Disk=466979MB) ...
🔄  Retriable failure: create: precreate: exec: "docker": executable file not found in $PATH

💣  Unable to start VM: create: precreate: exec: "docker": executable file not found in $PATH

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
👉  https://github.com/kubernetes/minikube/issues/new/choose
# 
```

_logs_:
```
# minikube logs

💣  api load: filestore "minikube": Docker machine "minikube" does not exist. Use "docker-machine ls" to list machines. Use "docker-machine create" to add a new one.

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
👉  https://github.com/kubernetes/minikube/issues/new/choose
# docker-machine ls
bash: docker-machine: command not found...
# 
```

_OS_:
```
# cat /etc/redhat-release 
CentOS Linux release 8.0.1905 (Core) 
# uname -a
Linux noc.uftwf.local 4.18.0-80.11.2.el8_0.x86_64 #1 SMP Tue Sep 24 11:32:19 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
# 
```

Is `docker` and/or `docker-machine` is prerequisites? if it is, it's not clear (or maybe I just missed it?) Also is it possible to add CentOS for KVM (Hypervisor Setup)?

Please advise.

[...]

PR #5854, opened 2019-11-06
Improve message when selected driver is incompatible with existing cluster

tstromberg:
## Before:

```
😄  minikube v1.5.2 on Debian rodete
💥  The existing "minikube" VM that was created using the "kvm2" driver, and is incompatible with the "none" driver.
👉  To proceed, either:
      1) Delete the existing VM using: 'minikube delete'
      or
      2) Restart with the existing driver: 'minikube start --vm-driver=kvm2'
```

## After:

```
😄  minikube v1.5.2 on Debian rodete
💥  The existing "minikube" cluster was created using the "kvm2" driver, and cannot be started using the "none" driver.
👉  To proceed, either:

    1) Delete the existing "minikube" cluster using: 'minikube delete'

    * or *

    2) Start the existing "minikube" cluster using: 'minikube start --vm-driver=kvm2'
	
💣  Exiting.

```

I believe the previous message may have been confusing to users, based on the context in #5828 

/cc @a1exus

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -616,7 +616,7 @@ func selectDriver(oldConfig *cfg.Config) string {
 		return name
 	}
 
-	out.ErrT(out.Conflict, `The existing "{{.profile_name}}" cluster that was created using the "{{.old_driver}}" driver, and cannot be started using the "{{.driver}}" driver.`,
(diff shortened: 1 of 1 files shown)


medyagh:
I like the new format and space between lines. We need. More of this easy to read solution message

Issue #5835, opened 2019-11-05
File sync (scp) not working properly


nanikjava:
Tried the instruction to sync file from local machine to vm

> https://minikube.sigs.k8s.io/docs/tasks/sync/ 

**Test scenario**

Add new entry into resolv.conf inside `$HOME/.minikube/files/etc/resolv.conf `

```
nameserver 1.1.1.1
nameserver 8.8.8.8
```
After running `minikube start` the file /etc/resolv.conf inside the VM is not updated. Tested using kvm2 VM

Also tried adding a new file `$HOME/.minikube/files/etc/` but cannot see it being copied over.

nanikjava:
From investigation scp is using `-t` flag where it should  be using `-r `

medyagh:
thank for you digging in to this feature,

how about we write an integration test for this ? basicly create a file "helloworld.txt" and put in the minkube home files and expect it to be there?

medyagh:
@nanikjava  do you mind sharing on which platofrm/driver and version you tried it ?

I tried it on mac os with hyperkit driver and it seems to be working

here is what I see:
```
medya@~/workspace/minikube (clean_cron) $ mkdir -p ~/.minikube/files/etc
medya@~/workspace/minikube (clean_cron) $ echo nameserver 8.8.8.8 > ~/.minikube/files/etc/helloword.txt
medya@~/workspace/minikube (clean_cron) $ minikube start

(minikube starting ...)
```

checking with minikube ssh 
```
$ minikube ssh
                         _             _            
            _         _ ( )           ( )           
  ___ ___  (_)  ___  (_)| |/')  _   _ | |_      __  
/' _ ` _ `\| |/' _ `\| || , <  ( ) ( )| '_`\  /'__`\
| ( ) ( ) || || ( ) || || |\`\ | (_) || |_) )(  ___/
(_) (_) (_)(_)(_) (_)(_)(_) (_)`\___/'(_,__/'`\____)

$ cat /etc/h
helloword.txt  hostname       hosts          
$ cat /etc/h
helloword.txt  hostname       hosts          
$ cat /etc/helloword.txt 
nameserver 8.8.8.8
```

nanikjava:
@medyagh Used kvm2 for testing and when the the VM was running when doing the minikube start again.

Tested without the -r  option it did not copied the files and when the option is put in it copied the file perfectly.

medyagh:
[...]

PR #5865, opened 2019-11-07
Add test case to test file sync between host and VM.

nanikjava:
#5835  

Executing 'minikube start' when it is currently running
will allow user to sync folder under MINIKUBE_HOME/files
to the VM.

The test case will copy a file to the MINIKUBE_HOME/files/etc
folder to be copied to the VM /etc/ folder.

### DIFF
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -35,8 +35,13 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
+
+	"k8s.io/minikube/pkg/minikube/localpath"
+
 	"github.com/elazarl/goproxy"
 	"github.com/hashicorp/go-retryablehttp"
+	"github.com/otiai10/copy"
 	"github.com/phayes/freeport"
 	"github.com/pkg/errors"
 	"golang.org/x/build/kubernetes/api"
@@ -59,6 +64,7 @@ func TestFunctional(t *testing.T) {
 			name      string
 			validator validateFunc
 		}{
+			{"CopySyncFile", setupFileSync},            // Set file for the file sync test case
 			{"StartWithProxy", validateStartWithProxy}, // Set everything else up for success
 			{"KubeContext", validateKubeContext},       // Racy: must come immediately after "minikube start"
 			{"KubectlGetPods", validateKubectlGetPods}, // Make sure apiserver is up
@@ -96,6 +102,7 @@ func TestFunctional(t *testing.T) {
 			{"TunnelCmd", validateTunnelCmd},
 			{"SSHCmd", validateSSHCmd},
 			{"MySQL", validateMySQL},
+			{"FileSync", validateFileSync},
 		}
 		for _, tc := range tests {
 			tc := tc
@@ -570,6 +577,37 @@ func validateMySQL(ctx context.Context, t *testing.T, profile string) {
 	}
 	if err = retry.Expo(mysql, 5*time.Second, 180*time.Second); err != nil {
 		t.Errorf("mysql failing: %v", err)
+	}
+}
+
+// Copy extra file into minikube home folder for file sync test
+func setupFileSync(ctx context.Context, t *testing.T, profile string) {
+	// 1. copy random file to MINIKUBE_HOME/files/etc
+	f := filepath.Join(localpath.MiniPath(), "/files/etc/sync.test")
+	err := copy.Copy("./testdata/sync.test", f)
+	if err != nil {
+		t.Fatalf("copy: %v", err)
+	}
+}
+
+// validateFileSync to check existence of the test file
+func validateFileSync(ctx context.Context, t *testing.T, profile string) {
+	if NoneDriver() {
+		t.Skipf("skipping: ssh unsupported by none")
+	}
+	// check file existence
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "cat /etc/sync.test"))
+	if err != nil {
+		t.Errorf("%s failed: %v", rr.Args, err)
+	}
+
+	expected, err := ioutil.ReadFile("./testdata/sync.test")
+	if err != nil {
+		t.Errorf("test file not found: %v", err)
+	}
+
+	if diff := cmp.Diff(string(expected), rr.Stdout.String()); diff != "" {
+		t.Errorf("/etc/sync.test content mismatch (-want +got):\n%s", diff)
 	}
 }
 
--- a/test/integration/testdata/sync.test
+++ b/test/integration/testdata/sync.test
@@ -1,0 +1,1 @@
+Test file for checking file sync process


test/integration/functional_test.go
```diff
@@ -564,6 +566,63 @@ func validateMySQL(ctx context.Context, t *testing.T, profile string) {
...
+	minikubeHome, err := ioutil.TempDir(home, "")
+	if err != nil {
+		t.Fatalf("error during creating tmp dir: %v", err)
+	}
+	defer os.RemoveAll(minikubeHome)
+	os.Setenv("MINIKUBE_HOME", minikubeHome)
+	fmt.Println("Setting  MINIKUBE_HOME  ", minikubeHome)
+
+	// 2. start VM
+	startArgs := append([]string{"start", "-p", profile, "--wait=true", "--alsologtostderr", "-v=3"}, StartArgs()...)
```
medyagh: I recommend not to start a VM for this !

what you could do is, add the files before other Serial tests 
and then let other Serial tests start minikube and this one in the end should check if the files are in there.

test/integration/functional_test.go
```diff
@@ -564,6 +566,63 @@ func validateMySQL(ctx context.Context, t *testing.T, profile string) {
...
+	c := exec.CommandContext(ctx, Target(), startArgs...)
+	rr, err := Run(t, c)
+	if err != nil {
+		t.Errorf("%s failed: %v", rr.Args, err)
+	}
+
+	// 3. copy random file to MINIKUBE_HOME/files/etc
+	tempfile := filepath.Join(minikubeHome, ".minikube/files/etc/filesync.txt")
```
medyagh: can you double check if ".minikube" is needed ? and also dont hard code directories. you could use filepath.Join

medyagh: for example from other integration tests I see this ``` fp := filepath.Join(localpath.MiniPath(), "cache", "images", img) ``` ...

test/integration/functional_test.go
medyagh: I recommend moving the test setup to the serial part
```
https://github.com/kubernetes/minikube/blob/5509edb1d38352c9b0e3101fad523186e93e0550/test/integration/functional_test.go#L62
```

we have two kind of integraiton  test serial and parallel, you could put some files in the folder after other serial tests have start a minikube ( for example before proxy test you can put the files, and then after it starts, check for the files to be there)

this will save tons of time in integraiton tests to not spin up a new VM

test/integration/functional_test.go
```diff
@@ -564,6 +566,63 @@ func validateMySQL(ctx context.Context, t *testing.T, profile string) {
...
+	err = copy.Copy("./testdata/nginx-pod-svc.yaml", tempfile)
```
medyagh: lets not make this test depend on other test's test-data, you can may be just echo "hello world testing sync" to a hello-sync.txt file ?

nanikjava: Will create a separate test file for this

test/integration/functional_test.go
```diff
@@ -96,6 +99,7 @@ func TestFunctional(t *testing.T) {
 			{"TunnelCmd", validateTunnelCmd},
 			{"SSHCmd", validateSSHCmd},
 			{"MySQL", validateMySQL},
+			{"FileSync", validateFileSync},
```
medyagh: @nanikjava the problem with this test is it starts a new minikube just for testing file sync. which will add time to the integraiton tests.


what I recommend is moving your test to 

this part of the test
```
	// Serial tests
	t.Run("serial", func(t *testing.T) {
		tests := []struct {
			name      string
			validator validateFunc
		}{
			{"StartWithProxy", validateStartWithProxy}, // Set everything else up for success
			{"KubeContext", validateKubeContext},       // Racy: must come immediately after "minikube start"
			{"KubectlGetPods", validateKubectlGetPods}, // Make sure apiserver is up
			{"CacheCmd", validateCacheCmd},             // Caches images needed for subsequent tests because of proxy
		}
```

in the abve code we have 4 serial tests, but only one of them starts minikube ! (start proxy)

what I recommend is, you have s function that is fileSyncTestSetup
that runs _before_ . validateStartWithProxy

and then you have a test _after "validateStartWithProxy" test in that checks if the file is synced.
(then you can delete the "start" in your test

test/integration/functional_test.go
```diff
@@ -568,6 +572,40 @@ func validateMySQL(ctx context.Context, t *testing.T, profile string) {
 	}
 }
 
+// validateFileSync start up VM and test sync file between host and VM
+func validateFileSync(ctx context.Context, t *testing.T, profile string) {
+	// This sync command will instruct the minikube to start so that it
+	// will trigger the sync process. The minikube HAS already been
+	// started in the serial test above.
+	startArgs := append([]string{"start", "-p", profile, "--alsologtostderr", "-v=3"}, StartArgs()...)
```
medyagh: remove the minikube start and see the comment above.

test/integration/functional_test.go
```diff
@@ -568,6 +572,40 @@ func validateMySQL(ctx context.Context, t *testing.T, profile string) {
...
+
+	// 1. copy random file to MINIKUBE_HOME/files/etc
+	tempfile := filepath.Join(localpath.MiniPath(), "/files/etc/filesync.txt")
+	err := copy.Copy("./testdata/filesync.txt", tempfile)
```
medyagh: move this to a test setup before this test runs, see comment above.

test/integration/functional_test.go
nanikjava: In the normal user scenario, the file sync process in minikube gets triggered only by doing 'minikube start' so in the test case the function 'validateFileSync' is doing exactly that. The 'serial' test starts minikube which is fine but in order for the file sync process to kicks in the test case need to simulate 'minikube start'. That is what I'm trying to simulate in the test case

Correct me if I'm wrong but from your comment you are suggesting to copy the files locally _before_ the 'minikube start' test case starts ?.

tstromberg: Agree with @medyagh here, even if it's a bit weird. `TestFunctional/parallel` is designed for tests that can be run in parallel and share a single minikube cluster for both performance and resource consumption reasons. We should make the comments clearer here.

I second medya's suggestion. Move the file creation to `validateStartWithProxy`, where it will then then be later available in `validateFileSync`.

test/integration/functional_test.go
```diff
@@ -568,6 +572,40 @@ func validateMySQL(ctx context.Context, t *testing.T, profile string) {
...
+
+	// 2. start VM (to trigger the file sync process)
+	c := exec.CommandContext(ctx, Target(), startArgs...)
+	rr, err := Run(t, c)
+	if err != nil {
+		t.Errorf("%s failed: %v", rr.Args, err)
+	}
+
+	// 3. check file existence
+	statusCmd := "(ls /etc/filesync.txt  >> /dev/null 2>&1 && echo yes) || echo no"
```
tstromberg: No need to make the check so complicated, just use `cat /etc/filesync.txt`and check for the error code.

If you want to make the check more comprehensive, assert that the output matches `testdata/filesync.txt`.

test/integration/functional_test.go
nanikjava: After re-reading the comments I think I understand what need to be done. Will fix it up. Thanks @medyagh and @tstromberg for the comments

nanikjava: Fix the PR

test/integration/functional_test.go
```diff
@@ -59,6 +62,7 @@ func TestFunctional(t *testing.T) {
 			name      string
 			validator validateFunc
 		}{
+			{"CopySyncFile", validateCopySynFile},      // Set file for the file sync test case
```
nanikjava: Naming wise not 100% confident it is correct but just to follow the rest of the test case. Would prefer to name it setupCopySyncFile(..) to make it more readable.

tstromberg: I like your idea as well - let's do that. You could even go shorter with the name:`setupFileSync`

test/integration/functional_test.go
```diff
@@ -573,6 +578,33 @@ func validateMySQL(ctx context.Context, t *testing.T, profile string) {
 	}
 }
 
+// Setup extra file for file sync test (validateFileSync), this
+// need to be done prior to starting up minikube.
+func validateCopySynFile(ctx context.Context, t *testing.T, profile string) {
+	// 1. copy random file to MINIKUBE_HOME/files/etc
+	tempfile := filepath.Join(localpath.MiniPath(), "/files/etc/filesync.txt")
```
tstromberg: This is a personal issue, but I tend to think of variables named `file` as representing `os.File` rather than a path. This variable's scope only lives on for a single line, so a shorter variable name should be used:

https://github.com/golang/go/wiki/CodeReviewComments#variable-names

`p`, `f`, `path` or `dest`, or omitting the variable declaration entirely would all be OK with me.

test/integration/functional_test.go
```diff
@@ -573,6 +578,33 @@ func validateMySQL(ctx context.Context, t *testing.T, profile string) {
...
+	err := copy.Copy("./testdata/filesync.txt", tempfile)
+	if err != nil {
+		t.Fatalf("copy: %v", err)
+	}
+}
+
+// validateFileSync start up VM and test sync file between host and VM.
```
tstromberg: This comment is no longer accurate. Just keep the comment short and describe what is tested, but not how, to prevent the comment from going stale.

test/integration/functional_test.go
```diff
@@ -573,6 +578,33 @@ func validateMySQL(ctx context.Context, t *testing.T, profile string) {
...
+// On startup Minikube should copy the filesync.txt to /etc folder
+func validateFileSync(ctx context.Context, t *testing.T, profile string) {
+	// check file existence
```
tstromberg: I don't find this comment particularly useful, so I recommend removing it. Comments are more useful when they describe "why" rather than "what".

test/integration/functional_test.go
```diff
@@ -573,6 +578,33 @@ func validateMySQL(ctx context.Context, t *testing.T, profile string) {
...
+	statusCmd := "cat /etc/filesync.txt"
```
tstromberg: You can remove some clutter here by removing `statusCmd` and moving the string directly to the `Run()` arguments.

test/integration/functional_test.go
```diff
@@ -573,6 +578,33 @@ func validateMySQL(ctx context.Context, t *testing.T, profile string) {
...
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", statusCmd))
+	if err != nil {
+		t.Errorf("%s failed: %v", rr.Args, err)
+	}
+
+	// 'no' result means sync did not happen
```
tstromberg: Remove this comment: the test failure message should always provide enough context.

test/integration/functional_test.go
```diff
@@ -573,6 +578,33 @@ func validateMySQL(ctx context.Context, t *testing.T, profile string) {
...
+	if strings.Contains(rr.Stdout.String(), "no") {
```
tstromberg: Since the previous command succeeded, we know that the file exists. If we're going to check output, let's check it against the original file instead:

```
expected, err := ioutil.ReadFile("./testdata/filesync.txt")

if diff := cmp.Diff(want, rr.Bytes()); diff != "" {
    t.Errorf("/etc/filesync.txt content mismatch (-want +got):\n%s", diff)
}
```

test/integration/functional_test.go
tstromberg: Seeing a .txt file is weird in /etc ... so, perhaps instead of `filesync.txt`, maybe a clearer test-filename like `sync.test`, so that folks seeing a future test error are less likely to think it's a normally expected file?

tstromberg:
```
test/integration/functional_test.go:601:13: undefined: cmp
test/integration/functional_test.go:601:22: undefined: want
test/integration/functional_test.go:601:30: rr.Bytes undefined (type *RunResult has no field or method Bytes)
```

Please consider running `TestFunctional` locally: 

`env TEST_ARGS="-test.run TestFunctional" make integration`

See https://minikube.sigs.k8s.io/docs/contributing/testing/

nanikjava:
> ```
> test/integration/functional_test.go:601:13: undefined: cmp
> test/integration/functional_test.go:601:22: undefined: want
> test/integration/functional_test.go:601:30: rr.Bytes undefined (type *RunResult has no field or method Bytes)
> ```
> 
> Please consider running `TestFunctional` locally:
> 
> `env TEST_ARGS="-test.run TestFunctional" make integration`
> 
> See https://minikube.sigs.k8s.io/docs/contributing/testing/

Weird!!! I did ran it. Will investigate. Thanks

nanikjava:
> env TEST_ARGS="-test.run TestFunctional" make integration

Found the issue was running a different branch test case in different terminal :)

medyagh:
@nanikjava 
```
        --- FAIL: TestFunctional/parallel/FileSync (0.03s)
            functional_test.go:596: (dbg) Run:  out/minikube-linux-amd64 -p minikube ssh "cat /etc/sync.test"
            functional_test.go:596: (dbg) Non-zero exit: out/minikube-linux-amd64 -p minikube ssh "cat /etc/sync.test": exit status 64 (30.306027ms)
                 
                ** stderr ** 
                * 'none' driver does not support 'minikube ssh' command
                 
                ** /stderr **
            functional_test.go:598: [out/minikube-linux-amd64 -p minikube ssh cat /etc/sync.test] failed: exit status 64
            functional_test.go:607: /etc/sync.test content mismatch (-want +got):
                Â Â string(
                -Â     "Test file for checking file sync process",
                +Â     "",
                Â Â )
```

make sure to skip this test for None driver, (t.Skip("not supported in none,....")

nanikjava:
> make sure to skip the tests for none driver

PR updated with the checks.

PR #5868, opened 2019-11-08
Getting rid of GetMachineName()

sharifelgamal:
This refactor is needed for multinode.

### DIFF
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -88,7 +88,7 @@ func CacheISO(config cfg.MachineConfig) error {
 
 // StartHost starts a host VM.
 func StartHost(api libmachine.API, config cfg.MachineConfig) (*host.Host, error) {
-	exists, err := api.Exists(cfg.GetMachineName())
+	exists, err := api.Exists(config.Name)
 	if err != nil {
 		return nil, errors.Wrapf(err, "exists: %s", cfg.GetMachineName())
 	}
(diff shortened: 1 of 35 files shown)


pkg/minikube/cluster/cluster.go
```diff
@@ -88,9 +88,9 @@ func CacheISO(config cfg.MachineConfig) error {
 
 // StartHost starts a host VM.
 func StartHost(api libmachine.API, config cfg.MachineConfig) (*host.Host, error) {
```
tstromberg: I find our inconsistent naming of `cfg` vs `config` absolutely confusing, mostly because it violates the following Go tenet: "Avoid renaming imports except to avoid a name collision; good package names should not require renaming. In the event of collision, prefer to rename the most local or project-specific import.": https://github.com/golang/go/wiki/CodeReviewComments#imports

For a follow-up PR, we should more consistently name our config references. I suggest:

-  `k8s.io/minikube/pkg/minikube/config` should be referred to as `config`
- Types defined in `pkg/minikube/config` should not stutter: perhaps `config.Profile`, `config.Machine`, `config.Kubernetes`?
- `cfg` currently can refer to `config.Config`, `config.MachineConfig`, or `config.KubernetesConfig`. We should find more consistent identifiers based on whatever the non-stuttering names are.

I won't make you do this in this PR however. Perhaps before you get to the point that a profile (or `config.Config`) object needs to refer to multiple machines though :)

tstromberg:
Feel free to merge once tests pass. It might take some work to get it sorted out :)

medyagh:
This is a great refacot ! thank you for doing it ! I love it

pkg/minikube/cluster/cluster.go
tstromberg: I followed up with @sharifelgamal and there will be a follow-up PR that addresses this and some concerns that @medyagh had:

- An object (named something like config.Cluster or config.Profile) will contain multiple config.Machine objects.
- config.Machine will contain a single config.Kubernetes object associated to that machine

This will allow us to handle multi-node and drivers that need additional Kubernetes hints like the version, as well as give us room to configure kubelet differently on the master versus worker nodes.

pkg/minikube/config/types.go
```diff
@@ -37,6 +37,7 @@ type Config struct {
 
 // MachineConfig contains the parameters used to start a cluster.
 type MachineConfig struct {
+	Name                string
```
medyagh: add comment what is Name, Cluster name ? Profile name ? or this specific machine's unique name ?

sharifelgamal: it's the machine's name, I just didn't want MachineConfig.MachineName to be a thing

sharifelgamal:
ok to test

sharifelgamal:
@minikube-bot OK to test

pkg/minikube/config/types.go
medyagh: Since it's a public field. For anyone reading it.

sharifelgamal:
test this please

Issue #5832, opened 2019-11-05
Feature: new release link on front page of website


nanikjava:
Website is the first place that normally user goes to find something. What if we update the front page of the website when new releases happen ?. Maybe some sort of a link to the latest version.

Screenshot shows the idea.

Comment and suggestions welcome !.

![ReleaseNotification](https://user-images.githubusercontent.com/883406/68184416-1078d180-fff3-11e9-92bd-a64c237ed7a6.png)

nanikjava:
Happy to take on the job of updating the website everytime there is a new release :+1:

medyagh:
I would like an easy access to download on first page, however at the same time the minikube first page is getting a bit crowded.
I would like to see a more simple minikube website front page. 


@tstromberg

medyagh:
@nanikjava currently we have Two links to github on the front page, I recommend removing one of them and replacing them with the Link to the Latest release download page.

nanikjava:
![Screenshot from 2019-11-08 07-08-56](https://user-images.githubusercontent.com/883406/68424541-581c7a80-01f8-11ea-9a84-6ff35d5d74b8.png)

@medyagh is the diagram correctly highlight the part that need to be changed to point to download the latest release ?

nanikjava:
PR submitted https://github.com/kubernetes/minikube/pull/5879

afbjorklund:
I think the idea was that the "Getting Started" would have a link to the download...

Especially since there are **so** many steps involved, to choosing-the-right-binary.

1)
 Linux
 macOS
 Windows

2)
 Direct
 Debian/Ubuntu (deb)
 Fedora/Red Hat (rpm)

3)
 KVM
 VirtualBox
 None (bare-metal)

etc) etc) etc)

These are not very easy to figure out, given just a list of names of cryptic binaries ?

Normally the user only wants **one** of these, probably same as their browser OS...

- docker-machine-driver-hyperkit
- docker-machine-driver-hyperkit.sha256
- docker-machine-driver-kvm2
- docker-machine-driver-kvm2.sha256
- minikube-1.5.2.rpm
- minikube-darwin-amd64
- minikube-darwin-amd64.sha256
- minikube-installer.exe
- minikube-linux-amd64
- minikube-linux-amd64.sha256
- minikube-windows-amd64.exe
- minikube-windows-amd64.exe.sha256
- minikube_1.5.2.deb

nanikjava:
[...]

PR #5880, opened 2019-11-10
Change the front page to link to the /release page

nanikjava:
#5832

To make it easier for user the front page link will now point to the
/release page


tstromberg:
I'm going to close this until we discuss and rethink our front-page strategy. Maybe a good office hours discussion?

I would argue that pointing users to the downloads page does not benefit them as much as pointing them to the installation instructions. Feel free to leave a comment that disagrees, however.

PR #5888, opened 2019-11-11
CI: Save test output for each commit, add status placeholders

tstromberg:
The question often a rises when reviewing a PR: is this test output stale? Is the test still running?

This attempts to address this question by:

- Giving each commit it's own output file
- Putting status placeholders in until the final results are uploaded

NOTE: This still requires updating the URL that Jenkins uses to update the final statuts result, so don't be surprised if GitHub doesn't show full logs at this URL.


tstromberg:
Closing for now as I wasn't able to get this to work.

PR #5889, opened 2019-11-11
Refactor config.Config to prepare for multinode

sharifelgamal:
This PR allows a profile to have multiple MachineConfigs, to account for multiple nodes in a cluster. It also inlines KubernetesConfig under MachineConfig, so that each node in a multinode cluster can have its own k8s config.


sharifelgamal:
Note: we'll eventually have to deal with supporting configs in the old style, but for now this will do.

PR #5912, opened 2019-11-14
Make lock names uid and path specific to avoid conflicts

### LINKED ISSUE #5911: Lock files should be per-user!
I was running into this issue while trying to let users create their own test clusters at a shared box.

Is there a similar issue for the following files, which need to be separated as well?
```
/tmp/juju-kubeconfigUpdate
/tmp/juju-kube-config
```

In addition, is it safe using a shared user for now (e.g. via `sudo -u <user> minikube ...`) to manage different "profiles"?
As far as I understood from the code (from `juju/mutex_flock`), these files only get used for locking temporarily, correct?
May I also ask why these temporary files never actually seem to get deleted?

_Originally posted by @enote-kane in https://github.com/kubernetes/minikube/issues/5353#issuecomment-544565261_

tstromberg:
This does now mean that if multiple users try to modify a locked file, that it will proceed rather than complain about permissions to a common lock file.

Fixes #5911

### DIFF
--- a/cmd/minikube/cmd/config/util_test.go
+++ b/cmd/minikube/cmd/config/util_test.go
@@ -122,17 +122,16 @@ func TestValidateProfile(t *testing.T) {
 			profileName: "82374328742_2974224498",
 		},
 		{
-			profileName: "minikube",
+			profileName: "validate_test",
 		},
 	}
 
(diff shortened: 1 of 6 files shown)


medyagh:
I like this new lock approach much better

PR #5916, opened 2019-11-14
Make it possible to recover from a previously aborted StartCluster (Ctrl-C)

tstromberg:
I ran into an issue earlier today where minikube was failing:

```
W1114 14:23:00.947646   38011 exit.go:101] Error restarting cluster: apiserver healthz: apiserver healthz never reported healthy

💣  Error restarting cluster: apiserver healthz: apiserver healthz never reported healthy
```

This was because the kubelet was crash-looping due to a missing file, probably because I hit Ctrl-C on a previous run:

```
Nov 14 22:19:36 minikube kubelet[2939]: F1114 22:19:36.393394    2939 server.go:196] failed to load Kubelet config file /var/lib/kubelet/config.yaml, error failed to read kubelet config file "/var/lib/kubelet/config.yaml", error: open /var/lib/kubelet/config.yaml: no such file or directory
```

```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1210,22 +1210,11 @@ func configureRuntimes(runner cruntime.CommandRunner, drvName string, k8s cfg.Ku
 
 // bootstrapCluster starts Kubernetes using the chosen bootstrapper
 func bootstrapCluster(bs bootstrapper.Bootstrapper, r cruntime.Manager, runner command.Runner, kc cfg.KubernetesConfig, preexisting bool, isUpgrade bool) {
-	// hum. bootstrapper.Bootstrapper should probably have a Name function.
-	bsName := viper.GetString(cmdcfg.Bootstrapper)
-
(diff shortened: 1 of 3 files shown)


tstromberg:
Flakes are troubling, but unrelated:

* none failed at `FAIL: TestAddons/parallel/Registry`. #5926
* KVM failed at `TestStartStop/group/cni` with `machine didn't return an IP after 120 seconds`. #5927
* HyperKit failed with `IP address never found in dhcp leases file`
* Hyper-V failed with `pod "integration-test=busybox" failed to start within 4m0s: timed out waiting for the condition`, which reports:

` Ready            False   Fri, 15 Nov 2019 11:31:37 -0800   Fri, 15 Nov 2019 11:28:10 -0800   KubeletNotReady              [container runtime is down, PLEG is not healthy: pleg was last seen active 3m59.182531534s ago; threshold is 3m0s]`
* VirtualBox failed due to running out of time:

Issue #5858, opened 2019-11-07
RBAC: minikube start error


BYTZJB:
**The exact command to reproduce the issue**:
minikube start --extra-config=apiserver.Authorization.Mode=RBAC

**The full output of the command that failed**:<details>
😄  minikube v1.5.2 on Darwin 10.15.1
🔥  Creating virtualbox VM (CPUs=2, Memory=2000MB, Disk=20000MB) ...
🌐  Found network options:
    ▪ HTTP_PROXY=http://yangweideMacBook-Pro.local:1087
    ▪ HTTPS_PROXY=http://yangweideMacBook-Pro.local:1087
    ▪ NO_PROXY=localhost,127.0.0.1,10.96.0.0/12,192.168.99.0/24,192.168.39.0/24
    ▪ http_proxy=http://yangweideMacBook-Pro.local:1087
    ▪ https_proxy=http://yangweideMacBook-Pro.local:1087
🐳  Preparing Kubernetes v1.16.2 on Docker '18.09.9' ...
    ▪ env HTTP_PROXY=http://yangweideMacBook-Pro.local:1087
    ▪ env HTTPS_PROXY=http://yangweideMacBook-Pro.local:1087
    ▪ env NO_PROXY=localhost,127.0.0.1,10.96.0.0/12,192.168.99.0/24,192.168.39.0/24
    ▪ env HTTP_PROXY=http://yangweideMacBook-Pro.local:1087
    ▪ env HTTPS_PROXY=http://yangweideMacBook-Pro.local:1087
    ▪ apiserver.Authorization.Mode=RBAC
🚜  Pulling images ...
🚀  Launching Kubernetes ...

💣  Error starting cluster: init failed. cmd: "/bin/bash -c \"sudo env PATH=/var/lib/minikube/binaries/v1.16.2:$PATH kubeadm init --config /var/tmp/minikube/kubeadm.yaml  --ignore-preflight-errors=DirAvailable--etc-kubernetes-manifests,DirAvailable--var-lib-minikube,FileAvailable--etc-kubernetes-manifests-kube-scheduler.yaml,FileAvailable--etc-kubernetes-manifests-kube-apiserver.yaml,FileAvailable--etc-kubernetes-manifests-kube-controller-manager.yaml,FileAvailable--etc-kubernetes-manifests-etcd.yaml,Port-10250,Swap\"": Process exited with status 1

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
[...]

PR #5917, opened 2019-11-15
SSHRunner: include stdout/stderr with error message

tstromberg:
With the recent runner refactor, SSHRunner was no longer returning stdout/stderr with the error, but ExecRunner was. This PR makes both behave identically. Example output, admittedly for a command with no output:

```
I1114 19:38:36.826497   58306 kubeadm.go:212] /data/minikube skipping compat symlinks: sudo test -d /data/minikube: Process exited with status 1
stdout: 
stderr: 
```

Issue seen in #5858

### DIFF
--- a/pkg/minikube/command/exec_runner.go
+++ b/pkg/minikube/command/exec_runner.go
@@ -75,7 +75,7 @@ func (*ExecRunner) RunCmd(cmd *exec.Cmd) (*RunResult, error) {
 		return rr, nil
 	}
 
-	return rr, fmt.Errorf("%s: %v\nstdout: %s\nstderr: %s", rr.Command(), err, rr.Stdout.String(), rr.Stderr.String())
+	return rr, fmt.Errorf("%s: %v\nstdout:\n%s\nstderr:\n%s", rr.Command(), err, rr.Stdout.String(), rr.Stderr.String())
 }
 
 // Copy copies a file and its permissions
--- a/pkg/minikube/command/exec_runner.go
+++ b/pkg/minikube/command/exec_runner.go
@@ -69,7 +69,7 @@ func (*ExecRunner) RunCmd(cmd *exec.Cmd) (*RunResult, error) {
 	}
 	// Decrease log spam
 	if elapsed > (1 * time.Second) {
-		glog.Infof("Exit %d for %v: (%s)", rr.Command(), elapsed)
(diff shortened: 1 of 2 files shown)


pkg/minikube/command/exec_runner.go
```diff
@@ -39,7 +40,7 @@ type ExecRunner struct{}
 // RunCmd implements the Command Runner interface to run a exec.Cmd object
 func (*ExecRunner) RunCmd(cmd *exec.Cmd) (*RunResult, error) {
 	rr := &RunResult{Args: cmd.Args}
-	glog.Infof("(ExecRunner) Run:  %v", rr.Command())
```
medyagh: I personally like that (ExecRunner) prrfix. In cases which a driver might use two runners it helps to know what is run in where... Less confusing.

pkg/minikube/command/exec_runner.go
```diff
@@ -62,19 +63,19 @@ func (*ExecRunner) RunCmd(cmd *exec.Cmd) (*RunResult, error) {
...
-	} else {
-		if exitError, ok := err.(*exec.ExitError); ok {
-			rr.ExitCode = exitError.ExitCode()
-		}
-		glog.Infof("(ExecRunner) Non-zero exit: %v: %v (%s)\n%s", rr.Command(), err, elapsed, rr.Output())
-		err = errors.Wrapf(err, "command failed: %s\nstdout: %s\nstderr: %s", rr.Command(), rr.Stdout.String(), rr.Stderr.String())
+		return rr, nil
 	}
-	return rr, err
+
```
medyagh: Consider using a way to indent each result.
Not simply \n
Mind that \t would only indent first line.

We need a helper function that scans a string line by line and idents them and returns them back.

medyagh: Clearifying.
For example

Command .....:
Error:
            ----------
Stdout:
             -------
             -------
Stderr:
              ------
              ------

medyagh: And since most of time stderr is more important. Maybe it can even be upper than stdout.

Usually stdout is very long....
If stdout is very long. Maybe the CMD should be repeated front of stderr
So we know this stderr is for this command

pkg/minikube/command/exec_runner.go
tstromberg: The log line already mentions exec_runner.go.

PR #5942, opened 2019-11-18
Add performance analysis packages to minikube ISO

priyawadhwa:
I was able to build the ISO locally and run `iostat` (included in sysstat) and `strace` in the minikube VM.


afbjorklund:
Do we _really_ need to add `python` ? Or can we run it from containers instead ?

priyawadhwa:
@afbjorklund @medyagh python was included as a dependency of iotop, but actually it seems that iotop is installed by another package and seems to work fine. I've removed `BR2_PACKAGE_PYTHON` and `BR2_PACKAGE_IOTOP`

Issue #5807, opened 2019-10-31
sudo dashboard crash: Error getting machine status: state: machine does not exist


jeger:
Trying to run 
sudo minikube dashboard

ends up with output: 
💣  Error getting machine status: state: machine does not exist

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:

but it works alright with just minikube dashboard

tstromberg:
This is because minikube clusters are per-user: when you run "sudo", you are running minikube as a different user (root -- not recommended), which does not know about the cluster.

Anyways, minikube should not "crash", but simply mention that it can't see the cluster. Leaving open as a bug because of that.

medyagh:
as tstromberg  mentioned, we should not crash on this ! we could replace the crash with a solution message that user doesn't need sudo for it !

priyawadhwa:
This seems to have been fixed by https://github.com/kubernetes/minikube/pull/7176

PR #5945, opened 2019-11-19
sudo dashboard crash: Error getting machine status: state: machine do…

RA489:
sudo dashboard crash: Error getting machine status: state: machine does not exist.
fixes https://github.com/kubernetes/minikube/issues/5807


tstromberg:
I don't believe this regexp is correct, nor the advice. I don't think there is a way from the current framework to detect that the dashboard is running as sudo, but the cluster isn't.

Issue #5677, opened 2019-10-21
crio.conf is hard-coded to a specific version


afbjorklund:
The contents of `/etc/crio/crio.conf` is _very_ context-sensitive to a particular version of cri-o:

https://github.com/kubernetes/minikube/commits/master/deploy/iso/minikube-iso/package/crio-bin/crio.conf.default

Yet we are overwriting the whole thing with a static template, even if only changing `pause_image`.

```toml
pause_image = "{{ .PodInfraContainerImage }}"
```

This breaks things, when there is a mismatch between `minikube` and `minikube.iso` (or other OS)

We should either make a smaller configuration change, or version the config file accordingly.

It also seem to be located in the _wrong_ module, it's in (the generic) CRI rather than in CRI-O ?

https://github.com/kubernetes/minikube/blob/master/pkg/minikube/cruntime/cri.go#L34_L328

I tried to use crio on a platform with crio 1.13, and it blew up.

The new configuration made it unable to find the "seccomp.json"

afbjorklund:
There is also other things missing, from the systemd crio.service: #4087

So the crio.conf, /etc/sysconfig/crio* and crio*.service all need overhaul

afbjorklund:
I will replace the overwrite, with some kind of more specific config update instead of the clobber:

```go
	c := exec.Command("/bin/bash", "-c", fmt.Sprintf("sudo mkdir -p %s && printf %%s \"%s\" | base64 -d | sudo tee %s", path.Dir(cPath), base64.StdEncoding.EncodeToString(b.Bytes()), cPath))
```

Would be nice if we could edit this TOML, with having to include all of the cri-o code to do it ?

```toml
# The image used to instantiate infra containers.
# This option supports live configuration reload.
pause_image = "k8s.gcr.io/pause:3.1"
```

afbjorklund:
I went with `sed` (instead of previous `tee`)

PR #5962, opened 2019-11-23
Add 'pause' command to freeze Kubernetes cluster

tstromberg:
By default, it only pauses namespaces populated by minikube, so that applications running on top remain running. This behavior can be overridden by passing `-A`, similar to `kubectl`, or by passing `--namespaces`.

WIP because it needs tests.


afbjorklund:
This change would look nicer if we fixed #5677 first.

blueelvis:
What if we pause the VM in its entirety and then resume it?

cmd/minikube/cmd/pause.go
```diff
@@ -0,0 +1,71 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
```
medyagh: 2019

cmd/minikube/cmd/pause.go
```diff
@@ -0,0 +1,71 @@
...
+			exit.WithError("Failed to get command runner", err)
+		}
+
+		config := cruntime.Config{Type: cc.ContainerRuntime, Runner: r}
+		cr, err := cruntime.New(config)
+		if err != nil {
+			exit.WithError("Failed runtime", err)
+		}
+
+		err = pause.Pause(cr, r)
```
medyagh: would it sound better if it was 
cruntime.Pause() ?

maybe that would sound more clear that this pauses the container runtimes and not the actual vm ?

cmd/minikube/cmd/unpause.go
```diff
@@ -0,0 +1,70 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
```
medyagh: 2019

pkg/minikube/pause/pause.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+	"k8s.io/minikube/pkg/minikube/command"
+	"k8s.io/minikube/pkg/minikube/cruntime"
+	"k8s.io/minikube/pkg/minikube/kubelet"
+)
+
+// Pause pauses a Kubernetes cluster
+func Pause(cr cruntime.Manager, r command.Runner) error {
```
medyagh: What do you think about moving it either to cruntime or make a new package for all actions
such as ops.Start, ops.Stop, ops.Pause,...

tstromberg:
> What if we pause the VM in its entirety and then resume it?

I decided against this approach, for a number of reasons:

- This allows us to pause Kubernetes without pausing the apps running on top of Kubernetes. The intent is for a follow-up PR to allow  for pausing only Kubernetes by default, but also allowing `-A` and `--namespace` arguments.
- The diifference between pausing the VM and pausing containers is negligible (<5% of a core)
- libmachine doesn't support suspend/resume

pkg/minikube/pause/pause.go
tstromberg: Good idea. I think this is the intent of the `cluster` class, so we should probably just use that.

medyagh:
@tstromberg are you still wokring on this ?

tstromberg:
@medyagh - Yes.

pkg/minikube/config/profile.go.orig
```diff
@@ -0,0 +1,201 @@
+/*
+Copyright 2019 The Kubernetes Authors All rights reserved.
+
```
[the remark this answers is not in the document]
tstromberg: Correct. Removed.

cmd/minikube/cmd/pause.go
tstromberg: Went with `cluster.Pause`, since it effectively pauses a cluster.

It gets a bit weird since it could also operate on particular namespaces, but that isn't the usual case, so l left it there.

cmd/minikube/cmd/status.go
```diff
@@ -40,16 +42,13 @@ import (
...
-	Misconfigured: `Misconfigured`,
-}
+const (
+	// Additional states used by kubeconfig
+	Configured    = "Configured"    // analagous to state.Saved
+	Misconfigured = "Misconfigured" // analagous to state.Error
+)
 
-// Status represents the status
+// Status holds string representations of libmachine.state.State
```
medyagh: libmachine status + plus kuberentes status

cmd/minikube/cmd/status.go
```diff
@@ -81,89 +80,98 @@ var statusCmd = &cobra.Command{
...
 		default:
 			exit.WithCodeT(exit.BadUsage, fmt.Sprintf("invalid output format: %s. Valid values: 'text', 'json'", output))
 		}
 
-		os.Exit(returnCode)
+		os.Exit(exitCode(st))
 	},
 }
 
+func exitCode(st *Status) int {
```
priyawadhwa: We could add a unit test for this function

pkg/minikube/cluster/pause.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+	"k8s.io/minikube/pkg/minikube/command"
+	"k8s.io/minikube/pkg/minikube/cruntime"
+	"k8s.io/minikube/pkg/minikube/kubelet"
+)
+
+// DefaultNamespaces are namespaces used by minikube, including addons
```
priyawadhwa: I think the istio addon is in `istio-operator` so might want to add that to the list

cmd/minikube/cmd/status.go
tstromberg: Thanks for the suggestion! I've added unit tests for this and the status functions.

Issue #5974, opened 2019-11-26
minikube cache add, doesn't add to other profiles unless restart


medyagh:
Here is how to reproduce:

### start two minikube profiles
```
minikube start
minikube start -p p1
```

### tag an image and add it to the cache
```
medya@~ $ docker tag alpine justforfun
medya@~ $ minikube cache add justforfun
```

### ssh into minikube and see the images

```
medya@~ $ minikube ssh -p p1 docker images | grep justforfun
medya@~ $ minikube ssh -p minikube docker images | grep justforfun
justforfun                                latest              965ea09ff2eb        5 weeks ago         5.55MB

```

as we see the cached image was loaded to minikube profile but not p1

however if we restart p1, it will pick up the cached image

medyagh:
What I suggest is this

- minikube start load all the images in global cache
- minikube cache add be be bound to a profile only ! and if you add an image using this command only be lodaded from that profile (as opposed to to all profiles on restart)

afbjorklund:
It's probably fine if it loads to the current profile only, but confusing if it does load to the others but isn't visible right away... That sounds like a bug, or at least something that was unintended ?

medyagh:
hm...for the PR I tried to keep the current behaviour.

tstromberg:
@medyagh - was this fixed?

sharifelgamal:
This is fixed.

PR #5975, opened 2019-11-26
wip: Improve minikube cache

medyagh:
before this PR
cache add was not consistent as described in this bug: https://github.com/kubernetes/minikube/issues/5974

This will be a breaking change !

- `cache add` command will be per profile ! 
- new field in machine config for user Cached Images 
- added integration tests for (cache add, cache delete and cache list, and check image loaded inside node)
- choose smaller images for in integration tests ( 6 mb instead of 655 mb)
- clean up the temporary files

### before this PR 

```
 $ cat ~/.minikube/config/config.json 
{
    "cache": {
        "gcr.io/hello-minikube-zero-install/hello-node": null,
        "k8s.gcr.io/echoserver:1.4": null,
        "med": null
    }
}
```

### after this PR 
- will be new map in profile config
```
{
    "Name": "minikube",
    "KeepContext": false,
    "EmbedCerts": false,
    "MinikubeISO": "https://storage.googleapis.com/minikube/iso/minikube-v1.5.1.iso",
    "Memory": 2000,
    "CPUs": 2,
    "DiskSize": 20000,
    "VMDriver": "hyperkit",
    "ContainerRuntime": "docker",
    "CachedImages": {
        "busybox": "loaded",
        "e4": "loaded",
        "k8s.gcr.io/echoserver:1.4": "loaded",
        "med": "loaded"
    },
...
...
...
```

- will warn if there is a older cache in global config
```
$ ./out/minikube cache list
⚠️  Warning:
        Images [gcr.io/hello-minikube-zero-install/hello-node k8s.gcr.io/echoserver:1.4 med] were cached using older version minikube.
        Please consider re-caching them again using "minikube cache add" commmand.
        Caches are now per-profile.
busybox
e4
k8s.gcr.io/echoserver:1.4
med
```

## TODO: 

1- load image only if it is not already loaded.
2- move the cache logics from machine package to its own package
3- store the image sha in the config map,
4- cache even if minikube is not runninng

### DIFF
--- a/cmd/minikube/cmd/cache.go
+++ b/cmd/minikube/cmd/cache.go
@@ -19,10 +19,7 @@ package cmd
 import (
 	"github.com/spf13/cobra"
 	cmdConfig "k8s.io/minikube/cmd/minikube/cmd/config"
-	"k8s.io/minikube/pkg/minikube/config"
-	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/minikube/exit"
-	"k8s.io/minikube/pkg/minikube/localpath"
 	"k8s.io/minikube/pkg/minikube/machine"
 )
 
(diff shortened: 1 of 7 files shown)


priyawadhwa:
Instead of warning users that old cached images need to be added again, could we automatically add them to all profiles? That way we can do the work for users.

medyagh:
> Instead of warning users that old cached images need to be added again, could we automatically add them to all profiles? That way we can do the work for users.

good idea

medyagh:
this PR is closed in favor of https://github.com/kubernetes/minikube/pull/5975

PR #5987, opened 2019-11-27
cache add: load images to all profiles & skip previously cached images

medyagh:
before this PR as described in this bug: #5974 

`minikube cache add` would only load images into the current profile (minikube) and other profiles had to be restarted to pick up the cache.

- skip re-copy existing images if image name with same sha exists on the node.
- added integration tests for (cache add, cache delete and cache list, and check image loaded inside node)
- choose smaller images for in integration tests ( 6 mb instead of 655 mb)
- clean up the temporary files
- refactor config.Load make profile Name mandatory for config.Load
- refactor cluster package to move machine  to machine package (to avoid cyclic 

Future PRs:

- do loading in parallel loading into multi profiles 
- decucple logic from cluster and machine packages


test/integration/functional_test.go
```diff
@@ -308,17 +308,52 @@ func validateDNS(ctx context.Context, t *testing.T, profile string) {
...
+				t.Errorf("failed to delete image busybox:1.28.4-glibc from cache: %v", err)
+			}
+		})
+
+		t.Run("list", func(t *testing.T) {
+			rr, err := Run(t, exec.CommandContext(ctx, Target(), "cache", "list"))
+			if err != nil {
+				t.Errorf("cache list failed: %v", err)
+			}
+			if !strings.Contains(rr.Output(), "k8s.gcr.io/pause") {
```
nanikjava: Just an idea, rather than checking for string content what if it split the output to array and check the array size with the image list listed in "add" ?

nanikjava:
The fix works as tested on local machine.

cmd/minikube/cmd/delete_test.go
```diff
@@ -117,7 +117,7 @@ func TestDeleteProfile(t *testing.T) {
 				t.Errorf("Profile folder of profile \"%s\" was not deleted", profile.Name)
 			}
 
-			pathToMachine := cluster.MachinePath(profile.Name, localpath.MiniPath())
+			pathToMachine := machine.Path(profile.Name, localpath.MiniPath())
```
tstromberg: I feel like this should be localpath.Machine. Thoughts?

pkg/minikube/config/config.go
```diff
@@ -124,8 +124,11 @@ func encode(w io.Writer, m MinikubeConfig) error {
 }
 
 // Load loads the kubernetes and machine config for the current machine
-func Load() (*MachineConfig, error) {
+func Load(profile ...string) (*MachineConfig, error) {
```
tstromberg: For consistency sake: let's just make the profile name a requirement.

pkg/minikube/machine/cache_images.go
```diff
@@ -131,20 +127,38 @@ func CacheAndLoadImages(images []string) error {
...
+	profiles, _, err := config.ListProfiles() // need to load image to all profiles
 	if err != nil {
-		return err
+		return errors.Wrap(err, "list profiles")
 	}
-
-	runner, err := CommandRunner(h)
-	if err != nil {
-		return err
+	for _, p := range profiles { // adding images to all the profiles
```
tstromberg: This feels weird: what happens if a new profile is added? It seems like the cache should be universal.

medyagh: can you clarify what you mean ?

pkg/minikube/config/config.go
medyagh: this will make the PR a bit bigger, but thats reasoanble aI will addd it

cmd/minikube/cmd/delete_test.go
medyagh: done

medyagh:
Shamefully after I added the ImageExist() ! this PR actually became 4 seconds slower when I do minikube cache add !

medyagh:
Update: I believe I fixed the performance issue of getting image sha.
and now a re-start minikube takes 18 seconds less !


```

-----------------------------------THIS PR------------------------------------------------------
medya@~/workspace/minikube (cache_fix2) $ time ./out/minikube start 
real    2m6.519s
user    0m7.788s
sys     0m8.187s
medya@~/workspace/minikube (cache_fix2) $ time ./out/minikube start 
real    0m7.614s
user    0m1.162s
sys     0m0.909s
medya@~/workspace/minikube (cache_fix2) $ time ./out/minikube stop
real    0m8.193s
user    0m0.087s
sys     0m0.094s
medya@~/workspace/minikube (cache_fix2) $ time ./out/minikube start
real    0m56.333s
user    0m1.084s
sys     0m0.887s

-----------------------------------HEAD------------------------------------------------------
 $ ./out/minikube delete --all --purge

$ time ./out/minikube start 
real    2m23.678s
user    0m9.359s
sys     0m10.000s


medya@~/workspace/minikube (master) $ time ./out/minikube start 
real    0m9.885s
user    0m0.670s
sys     0m0.446s

medya@~/workspace/minikube (master) $ time ./out/minikube stop
real    0m8.228s
user    0m0.087s
sys     0m0.109s

$ time ./out/minikube start
real    1m14.063s
user    0m0.555s
sys     0m0.467s
```

Issue #5867, opened 2019-11-07
driver's PreCreateCheck() is not used 


medyagh:
I added debugging statements, at appears no where in the code we are using them !

for hyperkit driver
```

// PreCreateCheck is called to enforce pre-creation steps
func (d *Driver) PreCreateCheck() error {
	fmt.Println("inside PreCreate hyperkit")
	return d.verifyRootPermissions()
}

```

for none driver
```
// PreCreateCheck checks for correct privileges and dependencies
func (d *Driver) PreCreateCheck() error {
	return d.runtime.Available()
}

```

andylibrian:
Is it a good first issue? I am interested to contribute.

afbjorklund:
Hmm, it looks to me that it _is_ being called ?

https://github.com/kubernetes/minikube/blob/master/pkg/minikube/machine/client.go#L183

Most likely it went back to default, when removed.

```go
// PreCreateCheck is called to enforce pre-creation steps
func (d *BaseDriver) PreCreateCheck() error {
        return nil
}
```

priyawadhwa:
As Anders pointed out, this code is being called now, so I'm going to close this issue

PR #5990, opened 2019-11-27
driver's PreCreateCheck() is not used

RA489:
driver's PreCreateCheck() is not used
fixes https://github.com/kubernetes/minikube/issues/5867


afbjorklund:
I would rather know why this driver function isn't being called, from pkg/minikube/machine/client.go

RA489:
@afbjorklund I think It is not being called from pkg/minikube/machine/client.go

medyagh:
I believe this function was used and wrongfully we thought not used.

Issue #5823, opened 2019-11-03
Consistent naming of packages


afbjorklund:
There aren't so many rules for naming buildroot packages, but some common ones are:

- Package names that end in `-bin` don't really build from source code, but download a binary

- Package names that end in `-master` don't build a released version, but from master branch.

The current minikube packages (for buildroot) are a historic mix of these, and should be aligned.

---

automount
cni-bin
cni-plugins-bin
**conmon-master** : _this is now releasing regularly, we currently use `v0.3.0`_
**containerd-bin** : _building this from source_
crictl-bin
**crio-bin** : _building this from source_
docker-bin
gluster
hyperv-daemons
podman
**runc-master** : _this is also non-releasing regularly, we use `v1.0.0-rc8`_
vbox-guest

afbjorklund:
We have now updated to newer versions:

conmon `v2.0.3`

runc `v1.0.0-rc9`

But the "-master" arguments above still hold.

afbjorklund:
Due to this way of building, our runc also lacks a proper version:

```
runc version commit: d736ef14f0288d6993a1845745d6756cfc9ddd5a
spec: 1.0.1-dev
```

While it is _supposed_ to look like this:

```
runc version 1.0.0-rc9
commit: d736ef14f0288d6993a1845745d6756cfc9ddd5a
spec: 1.0.1-dev
```

It is making docker confused (about the version number).

afbjorklund:
See also #5822, maybe we _shouldn't_ use binaries cni/cni-plugins and crictl - even if we do now ?

afbjorklund:
Well, apparently you _need_ to rename packages if you want to use a newer version than system...

**Buildroot 2019.02**
* [runc](https://github.com/buildroot/buildroot/tree/2019.02.x/package/runc) 1.0.0-rc8
* [docker-containerd](https://github.com/buildroot/buildroot/tree/2019.02.x/package/docker-containerd) v1.2.6
* [docker-engine](https://github.com/buildroot/buildroot/tree/2019.02.x/package/docker-engine) 18.09.9

afbjorklund:
I guess changing from the "go build" hack back to the Makefile, would also fix the version.

```Makefile
VERSION := ${shell cat ./VERSION}

runc: $(SOURCES)
	$(GO) build -buildmode=pie $(EXTRA_FLAGS) -ldflags "-X main.gitCommit=${COMMIT} -X main.version=${VERSION} $(EXT
[...]

PR #6002, opened 2019-12-01
WIP: Remove -master from package and install version

afbjorklund:
Since we are back to using released versions now, we don't need to pull
unreleased commits directly from the master branch repository any longer.

It is still exactly the same versions of conmon and runc, just renamed.
Currently we don't require any patches to the latest releases of each.

For #5823


afbjorklund:
Looks like Buildroot already has runc-1.0.0-rc8 in the default package list, so gets a naming conflict

https://github.com/buildroot/buildroot/tree/master/package/runc

Issue #5532, opened 2019-10-03
none without Docker installed: sudo systemctl start docker: exit status 5


laith-leo:
Unable to start minikube in RHEL 8 based, due to  the missing container runtime as a systemd service.

**[l@minikube ~]$ sudo minikube start --vm-driver=none
😄  minikube v1.4.0 on Centos 8.0.1905
💡  Tip: Use 'minikube start -p <name>' to create a new cluster, or 'minikube delete' to delete this one.
🔄  Starting existing none VM for "minikube" ...
⌛  Waiting for the host to be provisioned ...

💣  Failed to enable container runtime: running command: sudo systemctl start docker: exit status 5

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
👉  https://github.com/kubernetes/minikube/issues/new/choose
[l@minikube ~]$ sudo minikube start --vm-driver=none --container-runtime=cri-o
😄  minikube v1.4.0 on Centos 8.0.1905
💡  Tip: Use 'minikube start -p <name>' to create a new cluster, or 'minikube delete' to delete this one.
🔄  Starting existing none VM for "minikube" ...
⌛  Waiting for the host to be provisioned ...

💣  Failed to enable container runtime: running command: sudo systemctl restart crio: exit status 5

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
👉  https://github.com/kubernetes/minikube/issues/new/choose**:

**CentOS Linux release 8.0.1905 (Core)**:

afbjorklund:
Which runtime did you install/want, and what was the entire output of that system to start command ?

afbjorklund:
It seems to install OK, as long as docker has been started and selinux has been disabled.

```
sudo yum install --nobest docker-ce
sudo setenforce 0
```

There doesn't seem to be any Docker version for CentOS 8 yet, so it installs an older version:

* docker-ce-18.06.3.ce-3.el7.x86_64

```console
$ sudo minikube start --vm-driver=none
😄  minikube v1.4.0 on Centos 8.0.1905
🤹  Running on localhost (CPUs=2, Memory=3780MB, Disk=17394MB) ...
ℹ️   OS release is CentOS Linux 8 (Core)
🐳  Preparing Kubernetes v1.16.0 on Docker 18.06.3-ce ...
🚜  Pulling images ...
🚀  Launching Kubernetes ... 
🤹  Configuring local host environment ...

[...]

PR #6016, opened 2019-12-05
[dont merge]none without Docker installed: sudo systemctl start docker: exit stat…

RA489:
none without Docker installed: sudo systemctl start docker: exit status 5
fixes https://github.com/kubernetes/minikube/issues/5532


afbjorklund:
It’s not the only supported runtime, it is just a bug that requires you to have the docker client installed (unused).

medyagh:
I don't think this is the best way to handle this ! 
if something is required we should check before.

in this case, the none driver does not need docker.

PR #6031, opened 2019-12-09
Fix ingress-dns example manifest URL

anukul:
Minor change in the docs, using the raw YAML URL in the argument to `kubectl apply -f`


tstromberg:
Thank you!

PR #6043, opened 2019-12-10
Sync cached image list for Kubernetes v1.17

tstromberg:
- Updates the core addon versions required by Kubernetes v1.17
- Removes unused kube-dns images
- Removes pause version logic that was required for Kubernetes pre-v1.11
- Simplifies logic to make it easier to reason about what image versions are necessary
- Returns an error
- Adds tests!


pkg/minikube/bootstrapper/images/kubeadm.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	return path.Join(KubernetesRepo(mirror), "coredns"+ArchTag(false)+cv)
+}
+
+// etcd returns the image used for etcd
+func etcd(v semver.Version, mirror string) string {
+	// Should match `DefaultEtcdVersion` in:
+	// https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/constants/constants.go
+	ev := "3.4.3-0"
+	switch v.Minor {
```
medyagh: maybe future proof it and include the major version ?

tstromberg: I'd prefer to leave the extra complexity out until it's necessary: even if we added it, it would not future proof this function, as changes would still be required to select the correct versions for Kubernetes 2.0.

medyagh:
the PR description says , Removes pause logic.... is that still true?

pkg/minikube/bootstrapper/images/images.go
```diff
@@ -14,188 +14,63 @@ See the License for the specific language governing permissions and
...
-
-	images = append(images, []string{
-		imageRepository + "kube-addon-manager" + ArchTag(false) + "v9.0.2",
-		// This must match deploy/addons/dashboard/dashboard-dp.yaml
-		"kubernetesui/dashboard:v2.0.0-beta8",
-		minikubeRepository + "storage-provisioner" + ArchTag(false) + "v1.8.1",
-	}...)
-
-	return images
+// Pause returns the image name for thoß… pause image
```
medyagh: fix german

prasadkatti:
@tstromberg - Kubeadm gives you a list of images that are needed for a particular k8s version with `kubeadm config images list`. You can even pull the images directly using `kubeadm config images pull`. Is that something we can use to avoid having to update these versions to keep them in sync?

PR #6050, opened 2019-12-11
Don't check image versions with the none driver (no crictl)

tstromberg:
Fixes last-minute breakage with an integration test added in #6043


medyagh:
not related tot his but I see kvm duration is much more than vbox ! it used to be the other way

Issue #5734, opened 2019-10-25
beautify mininikube emojis !


medyagh:
I feel like there could be a brush up on the minikube emojis, 

## minikube delete
its kind of weird in one emoji we have breaking heart and the one after is fire...

```
🔥  Removing /root/.minikube/machines/minikube ...
💔  The "minikube" cluster has been deleted.
🔥  Successfully deleted profile "minikube"
```
this specially looks weird when deleting multiple profiles

## minikube stop
also for minikube stop (tested on none driver) I see two stop icons, which one of them is redundant

```
$ sudo ./out/minikube stop
✋  Stopping "minikube" in none ...
✋  Stopping "minikube" in none ...
🛑  "minikube" stopped.
```

I would be happy to review any PR that makes deleting minikube more beautiful

afbjorklund:
> also for minikube stop (tested on none driver) I see two stop icons

I think it only gets 5 seconds in order to `StopHost`.

This is OK for pulling the plug on a VM, but `none` doesn't support stop.
Instead it is doing a graceful stop of all the running containers instead...

And that probably takes longer, so it does a retry ?

afbjorklund:
> I would be happy to review any PR that makes deleting minikube more beautiful

So do you mean "swap for other random emojis", or do you envision a custom font or something ?

https://github.com/kubernetes/minikube/blob/v1.5.0/pkg/minikube/out/style.go#L51_L116

afbjorklund:
Think most of these were designed by @tstromberg

nicolereineke:
Has the team considered FontAwesome as a source of additional emojis ?  (https://fontawesome.com/icons?d=gallery&m=free) 
We could maintain the existing icons that are working, and as we need new ones, we can easily (and free) phase them in by adding Creative Commons Attribution.

In this instance, the trash or minus icon could easily meet the needs.
 https://fontawesome.com/icons/trash-alt?style=regular

If the concern is a need for maintaining the black outline and color scheme, I can contribute colorized versions in .svg or other format required.

afbjorklund:
@nicolereineke : how would you use custom fonts or icons in the terminal ?
Normally it just uses Unicode characters, which can render like anything...

Like so: (`'\U0001f5d1'`) 🗑️ https://emojipedia.org/wastebasket/
These can appear in black-and-white, or maybe not at all (ASCII)

nicolereineke:
It took setup, but we are able to add new fonts to terminal as described
here:
[...]

PR #6086, opened 2019-12-15
Add option to use icons instead of emojis

afbjorklund:
Make it possible to use icons from "Nerd Fonts",
instead of using the regular emojis from Unicode.

See https://www.nerdfonts.com/font-downloads for
the required fonts in order to display properly.

For #5734

### DIFF
--- a/pkg/minikube/out/out.go
+++ b/pkg/minikube/out/out.go
@@ -47,8 +47,12 @@ var (
 	errFile fdWriter
 	// useColor is whether or not color output should be used, updated by Set*Writer.
 	useColor = false
+	// useIcons is whether to use icons instead of emojis, updated by Set*Writer.
+	useIcons = false
 	// OverrideEnv is the environment variable used to override color/emoji usage
 	OverrideEnv = "MINIKUBE_IN_STYLE"
+	// NerdFontsEnv s the environment variable used to use icons instead of emojis
+	NerdFontsEnv = "MINIKUBE_NERD_FONTS"
 )
 
 // fdWriter is the subset of file.File that implements io.Writer and Fd()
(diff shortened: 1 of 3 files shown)


afbjorklund:
Choosing names instead of icons, made it harder to compare with emojis.
Could keep it as an alternative for later (?), used raw `rune` literals for now:

```
''
const human rune = '\uf7e5'
```

afbjorklund:
Here is a short summary of styles (emojis, icons, enums):  [nerdfonts.txt](https://github.com/kubernetes/minikube/files/3965152/nerdfonts.txt)

medyagh:
that is cool ! do you have a screenshot how it looks like ?

afbjorklund:
@medyagh : see https://github.com/kubernetes/minikube/issues/5734#issuecomment-564599519

tstromberg:
Are you sending this PR out because it's possible, or because you think it is the right thing to do?

afbjorklund:
@tstromberg : I thought I would send it in because I ended up writing it (just to see what it looked like)

I didn't know about the availability of this font, even though I had seen powerlines and Font Awesome individually. I'm not sure what the original plan was, maybe Medya only wanted to change one emoji ?

My finding was that changing just one emoji for one icon ended up looking _worse_ than keeping it. So it is better to change **all** of them in that case. The icons being monochrome is a pro and a con, I guess.

There is a lot of hidden complexity in this feature (i.e. telling users how to use it, maintaining two sets)

So I am OK with closing it, unless it grows a following. I'll keep the branch around, for a while.

It also got me thinking about using color for minikube, but I think I will leave that can of worms closed...

PR #625, opened 2016-09-23
Changed docker-machine and sshutil to support hostnames with colons i…

aaron-prindle:
…n them, for example ipv6 addresses

Issue #576, opened 2016-09-13
IPv6 problem with the SSH client in start.go


aoboturov:
**Is this a BUG REPORT** :

**Minikube version** (use `minikube version`):
minikube version: v0.9.0

**Environment**:
- **OS** (e.g. from /etc/os-release):
Windows 10 Pro
- **VM Driver** (e.g. `cat ~/.minikube/machines/minikubeVM/config.json | grep DriverName`):
"DriverName": "hyperv"
- **Docker version** (e.g. `docker -v`):
Docker version 1.12.0, build 8eab29e, experimental
- **Install tools**:
- **Others**:
Cygwin 64

**What happened**:
Looks like IPv6 does not work for minikube ssh client

```
(minikubeVM) DBG | [stderr =====>] :
(minikubeVM) Calling .GetSSHHostname
(minikubeVM) DBG | [executing ==>] : C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -NonInteractive ( Get-VM minikubeVM ).state
(minikubeVM) DBG | [stdout =====>] : Running
(minikubeVM) DBG |
(minikubeVM) DBG | [stderr =====>] :
(minikubeVM) DBG | [executing ==>] : C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -NonInteractive (( Get-VM minikubeVM ).networkadapters[0]).ipaddresses[0]
(minikubeVM) DBG | [stdout =====>] : fe80::215:5dff:fe4b:10f
(minikubeVM) DBG |
(minikubeVM) DBG | [stderr =====>] :
(minikubeVM) Calling .GetSSHPort
(minikubeVM) Calling .GetSSHKeyPath
(minikubeVM) Calling .GetSSHUsername
E0913 11:33:41.650710    2184 start.go:98] Error updating cluster:  dial tcp: too many colons in address fe80::215:5dff:fe4b:10f:22
```

**What you expected to happen**:

When I run:

```
minikube.exe start --alsologtostderr=true --show-libmachine-logs=true --vm-driver="hyperv"
```

it is expected the minikube to start services properly.

**How to reproduce it** (as minimally and precisely as possible):

Enable Hyper-V on Windows 10 Pro. Add user to the Hyper-V managers group. Download minikube. Run it with the command as specified above.

**Anything else do we need to know**:

aoboturov:
[...]

PR #6110, opened 2019-12-17
Fixed IPv6 format for SSH

digitalcoyote:
Not very familiar with go, so I have yet to be able to test this, but it appears to be similar to the issue in #576 . I made a similar fix to #625 and I'm working on setting up golang so I can test it.


digitalcoyote:
Finally got to test it. I don't know if I'll hit another issue, but It's gotten further than before. I double checked the project for another instance of this and did not see one (but as I said I am unfamiliar with Go so I may have overlooked one).
/assign @ra489

PR #6119, opened 2019-12-18
Address low-hanging integration timeout & image list issues

tstromberg:
- Wait longer for slow pods
- Offset longer
- Filter more images out


medyagh:
@tstromberg this seems to have meaniing fully improved the KVM integration tests but didn't help virtualbox and hyperkit

PR #6120, opened 2019-12-19
[WIP] Add checking for libvirt group

### LINKED ISSUE #5617: kvm2 driver: check if user in "libvirt" group and exit if they are not
Please, add that user starting minikube should be in "libvirt" group. Because without adding user in "libvirt" group will be an error "authentification unavailable"

nanikjava:
fixes #5617 

Currently there is no checking whether the current user is part of the libvirt group.  This check is to ensure the code will fail fast if the user is not in the group

### DIFF
--- a/pkg/minikube/registry/drvs/kvm2/kvm2.go
+++ b/pkg/minikube/registry/drvs/kvm2/kvm2.go
@@ -21,6 +21,7 @@ package kvm2
 import (
 	"fmt"
 	"os/exec"
+	"os/user"
 	"path/filepath"
 	"strings"
 
@@ -86,6 +87,24 @@ func configure(mc config.MachineConfig) interface{} {
 	}
 }
 
+// check whether current user is in either libvirt or libvirtd group
+func isUserInVirtGroup() (bool, string) {
+	v, _ := user.LookupGroup("libvirt")
+	vd, _ := user.LookupGroup("libvirtd")
+
+	u, _ := user.Current()
+	o, _ := u.GroupIds()
+	uf := false
+	for _, g := range o {
+		if ( ((v!=nil) && (g == v.Gid))  || ((vd!=nil) && (g == vd.Gid)) ) {
+			uf = true
+			break
+		}
+	}
+
+	return uf, u.Name
+}
+
 func status() registry.State {
 	path, err := exec.LookPath("virsh")
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/minikube/registry/drvs/kvm2/kvm2.go
```diff
@@ -87,6 +91,28 @@ func configure(mc config.MachineConfig) interface{} {
 }
 
 func status() registry.State {
+	// get libvirt id
+	virtid, err := group.LookupGid("libvirt")
```
tstromberg: Is there a case that the current check fails but this one succeeds?

I ask because checking specifically for `libvirt` will not be correct in all cases. Some distros use `libvirtd`. Other configurations may use something else entirely.

nanikjava: Have outlined the investigation in here https://github.com/kubernetes/minikube/issues/5617#issuecomment-554728642

pkg/minikube/registry/drvs/kvm2/kvm2.go
```diff
@@ -86,7 +90,32 @@ func configure(mc config.MachineConfig) interface{} {
 	}
 }
 
+// check whether current user is in either libvirt or libvirtd group
+func isUserInVirtGroup() (bool, string) {
+	vid, _ := group.LookupGid("libvirt")
```
josedonizetti: Let's use go stdlib only here. No need to add an extra dependency.

pkg/minikube/registry/drvs/kvm2/kvm2.go
```diff
@@ -86,7 +90,32 @@ func configure(mc config.MachineConfig) interface{} {
...
+		}
+	}
+
+	return uf, u.Name
+}
+
 func status() registry.State {
+	ig, n := isUserInVirtGroup()
+
+	if !ig {
```
josedonizetti: Let"s do the check after `virsh domcapabilities`.

pkg/minikube/registry/registry.go
```diff
@@ -68,11 +68,12 @@ type StatusChecker func() State
...
-	Healthy   bool
-	Error     error
-	Fix       string
-	Doc       string
+	Installed   bool
+	Healthy     bool
+	Error       error
+	Fix         string
+	Doc         string
+	FixRequired bool
```
josedonizetti: do we need the extra flag? Can't we use healthy, like we are using for the `domcapabilities`?

pkg/minikube/registry/drvs/kvm2/kvm2.go
nanikjava: Code refactored to use os.user package

pkg/minikube/registry/registry.go
nanikjava: On further investigation the code can use the Healthy flag. Changes made.

pkg/minikube/registry/drvs/kvm2/kvm2.go
nanikjava: Done

medyagh:
@nanikjava is there anything need help with on this pr?

tstromberg:
Closing this due to merge conflicts - feel free to reopen once addressed.

PR #6124, opened 2019-12-19
Enable or disable addons per profile

### LINKED ISSUE #4227: Addon enablement should be per-profile
1. Addons are global (maybe we should make them per instance?) - and gvisor enablement is only enabled if the container runtime backend is containerd. Unfortunately gvisor disablement is also subject to this check which is problematic as I can't turn off the gvisor addon, so all my minikube instances have the gvisor addon pod running and/or crashing by default. Should be a relatively easy fix. 

2. As I'm thinking about this - I wonder if it makes sense at all to check whether the "current minikube instance" has containerd when the addon enablement is instance idependent?

1. `minikube addons enable gvisor` on a containerd enabled minikube 
2.  minikube delete and recreate without containerd 
3.  `minikube addons disable gvisor`

```
💣  disable failed: [This addon can only be enabled with the containerd runtime backend.

To enable this backend, please first stop minikube with:

minikube stop

and then start minikube again with the following flags:

minikube start --container-runtime=containerd --docker-opt containerd=/var/run/containerd/containerd.sock]

```
**The operating system version**: Linux

priyawadhwa:
This PR makes a few changes to the addons code:

1. It moves a lot of the addons code for enabling/disabling addons from package `cmd/config/util` to a new package `minikube/pkg/addons`

1. We now enable/disable addons per profile, and so addons information is stored in machine configs per profile. Any addon related information in the global config is now disregarded.

1. Since addons can no longer be globally saved, on `minikube delete` the state of addons is wiped

This PR should fix #4227 

The new config has an `Addons` section:

```
{
	"Name": "minikube",
	"KeepContext": false,
	"EmbedCerts": false,
	"MinikubeISO": "https://storage.googleapis.com/minikube/iso/minikube-v1.6.0.iso",
	"Memory": 2000,
	"CPUs": 2,
	"DiskSize": 20000,
	"VMDriver": "hyperkit",
	"ContainerRuntime": "docker",
	"HyperkitVpnKitSock": "",
	"HyperkitVSockPorts": [],
	"DockerEnv": null,
	"InsecureRegistry": null,
	"RegistryMirror": null,
	"HostOnlyCIDR": "192.168.99.1/24",
	"HypervVirtualSwitch": "",
	"KVMNetwork": "default",
	"KVMQemuURI": "qemu:///system",
	"KVMGPU": false,
	"KVMHidden": false,
	"DockerOpt": null,
	"DisableDriverMounts": false,
	"NFSShare": [],
	"NFSSharesRoot": "/nfsshares",
	"UUID": "",
	"NoVTXCheck": false,
	"DNSProxy": false,
	"HostDNSResolver": true,
	"KubernetesConfig": {
              ...
	},
	"HostOnlyNicType": "virtio",
	"NatNicType": "virtio",
	"Addons": {
		"helm-tiller": true
	}
}
```


pkg/addons/addons.go
```diff
@@ -0,0 +1,233 @@
...
+	if err != nil {
+		out.ErrT(out.Conflict, "{{.error}}", out.V{"error": err})
+		return err
+	}
+	//if addon is already enabled or disabled, do nothing
+	if alreadySet {
+		return nil
+	}
+
+	// TODO(r2d4): config package should not reference API, pull this out
```
tstromberg: Maybe remove this TODO? :)

pkg/minikube/config/profile.go
```diff
@@ -73,6 +74,15 @@ func ProfileExists(name string, miniHome ...string) bool {
 	return err == nil
 }
 
+// CurrentProfile returns the name of the current profile
+func CurrentProfile() string {
+	profile := viper.GetString(MachineProfile)
+	if profile != "" {
+		return profile
+	}
+	return "minikube"
```
tstromberg: A lot of trouble has come from API's using a similar function, but needing to access multiple profiles. 

Instead, can we simply update our code to use viper.GetString(MachineProfile) in the `cmd` package, and make sure to pass the correct profile into other packages?

NOTE: I don't think viper.GetString(MachineProfile) should ever be "" unless the user specifies it to shoot themselves in the foot.

pkg/addons/addons.go
priyawadhwa: done!

pkg/minikube/config/profile.go
priyawadhwa: sounds good, I'll use viper.GetString(MachineProfile) instead.

medyagh:
nicely done, just tested on my machine it works great.

here is the profile config after this PR.


```
{
	"Name": "minikube",
	"KeepContext": false,
	"EmbedCerts": false,
	"MinikubeISO": "https://storage.googleapis.com/minikube/iso/minikube-v1.6.0.iso",
	"Memory": 2000,
	"CPUs": 2,
	"DiskSize": 20000,
	"VMDriver": "hyperkit",
	"ContainerRuntime": "docker",
	"HyperkitVpnKitSock": "",
	"HyperkitVSockPorts": [],
	"DockerEnv": null,
	"InsecureRegistry": null,
	"RegistryMirror": null,
	"HostOnlyCIDR": "192.168.99.1/24",
	"HypervVirtualSwitch": "",
	"KVMNetwork": "default",
	"KVMQemuURI": "qemu:///system",
	"KVMGPU": false,
	"KVMHidden": false,
	"DockerOpt": null,
	"DisableDriverMounts": false,
	"NFSShare": [],
	"NFSSharesRoot": "/nfsshares",
	"UUID": "",
	"NoVTXCheck": false,
	"DNSProxy": false,
	"HostDNSResolver": true,
	"KubernetesConfig": {
		"KubernetesVersion": "v1.17.0",
		"NodeIP": "192.168.64.190",
		"NodePort": 8443,
		"NodeName": "minikube",
		"APIServerName": "minikubeCA",
		"APIServerNames": null,
		"APIServerIPs": null,
		"DNSDomain": "cluster.local",
		"ContainerRuntime": "docker",
		"CRISocket": "",
		"NetworkPlugin": "",
		"FeatureGates": "",
		"ServiceCIDR": "10.96.0.0/12",
		"ImageRepository": "",
		"ExtraOptions": null,
		"ShouldLoadCachedImages": true,
		"EnableDefaultCNI": false
	},
	"HostOnlyNicType": "virtio",
	"NatNicType": "virtio",
	"Addons": {
		"helm-tiller": true
	}
}
```

medyagh:
one thing that could be for a PR is  output of  addons list 
which can be converted to a table that 
name, status, profile

```
 ./out/minikube addons list
- addon-manager: enabled
- dashboard: disabled
- default-storageclass: enabled
- efk: disabled
- freshpod: disabled
- gvisor: disabled
- helm-tiller: enabled
- ingress: disabled
- ingress-dns: disabled
- logviewer: disabled
- metrics-server: disabled
- nvidia-driver-installer: disabled
- nvidia-gpu-device-plugin: disabled
- registry: disabled
- registry-creds: disabled
- storage-provisioner: enabled
- storage-provisioner-gluster: disabled
```

priyawadhwa:
@medyagh the output of addons list should also take the `--profile` flag and output for that profile!

In addons_list.go we call [addonBundle.IsEnabled()](https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/config/addons_list.go#L103)

and in this PR I updated `IsEnabled()` to check for the specific profile [here](https://github.com/kubernetes/minikube/pull/6124/files#diff-07ad0c54f98b231e68537d908a214659R59)

tstromberg:
TestAddons failed:

```
  *  X addon enable failed: property name "ingress" not found
```

Check that the `minikube start --addons` feature still works.

medyagh:
weird to see "verify_cache_inside_node" test fail ! I wonder how could that be related to this PR

cmd/minikube/cmd/dashboard.go
```diff
@@ -33,7 +33,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	configcmd "k8s.io/minikube/cmd/minikube/cmd/config"
+	pkgaddons "k8s.io/minikube/pkg/addons"
```
tstromberg: Can this be made just "addons"?

pkg/addons/addons.go
```diff
@@ -0,0 +1,245 @@
...
+
+func Set(name, value, profile string) error {
+	a, valid := isAddonValid(name)
+	if !valid {
+		return errors.Errorf("%s is not a valid addon", name)
+	}
+
+	// Run any additional validations for this property
+	if err := run(name, value, profile, a.validations); err != nil {
+		return err
```
tstromberg: since there are 6 possible errors emitted by this function, do you mind adding error breadcrumbs to help us when debugging it in the future?

For example:

return errors.Wrap(err, "run")

pkg/addons/addons.go
```diff
@@ -0,0 +1,245 @@
...
+				}
+
+			} else {
+				addonFile = addon
+			}
+			if err := cmd.Copy(addonFile); err != nil {
+				return errors.Wrapf(err, "enabling addon %s", addon.AssetName)
+			}
+		}
+	} else {
```
tstromberg: Can this be refactored with less indentation for readability?

The easy win here would be to return if enabled, so that the disable state is dedented one level. 

Looking more closely, it seems like the loops are nearly identical, so you may be able to simply generate an array of addonFiles outside of the if statement, so that the only code that needs to be indented is `cmd.Copy` vs `cmd.Remove`

test/integration/addons_test.go
```diff
@@ -36,7 +36,6 @@ import (
 
 // TestAddons tests addons that require no special environment -- in parallel
 func TestAddons(t *testing.T) {
-	MaybeParallel(t)
```
tstromberg: Can we take this line out for now? It's too difficult to see if this PR makes the parallel integration testing behavior better or worse with it.

cmd/minikube/cmd/dashboard.go
priyawadhwa: so the compiler complains that there's a constant named [addons](https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/start.go#L106) in the same package :(

test/integration/addons_test.go
priyawadhwa: yup, it should be out!

pkg/addons/addons.go
priyawadhwa: done, there shouldn't be any duplication now

pkg/addons/addons.go
priyawadhwa: done!

tstromberg:
Thank you!

PR #5833, opened 2019-11-05
Add hyperkit version check whether user's hyperkit is newer

govargo:
### What type of PR is this?

/kind bug

### What this PR does / why we need it:

(Updated 2019/11/11)
I removed point1 of the PR due to duplication of #5840.

<del>#### 1. FastFail if driver error occurres.</del>

<del>So far, even if driver error occurres(like not installed) minikube keeps to starting VM when `minikube start`.</del>
<del>Of course this finish with failure.</del>
<del>This PR fix this behaviour.</del>
<del>If driver error occurres, minikube stop starting VM(minikube exits).</del>

#### 2. Add version check of hyperkit which inspect user's hyperkit is newer than specific version

If hyperkit's version is older than expected, `minikube start(hyperkit)` will fail due to version.
Related Issue #5594, #5780

This PR add suggestion about hyperkit upgrade information, if user's hyperkit version is older than expected.

### Which issue(s) this PR fixes:
Fixes #5478, #5821

### Does this PR introduce a user-facing change?
<del>Yes. This PR add hyperkit version check to `minikube start` command and FastFail if driver error occurres.</del>
Yes. This PR add hyperkit version check to `minikube start` command.

(Removed FastFail due to duplication of #5840)<details>
**Before this PR, minikube keeps to starting VM(but keep tp failing) **

```
$ minikube start --vm-driver=hyperkit
😄  minikube v1.5.1 on Darwin 10.14.5

⚠️  'hyperkit' driver reported a possible issue: exec: "hyperkit": executable file not found in $PATH
💡  Suggestion: Run 'brew install hyperkit'
📘  Documentation: https://minikube.sigs.k8s.io/docs/reference/drivers/hyperkit/

💾  Downloading driver docker-machine-driver-hyperkit:
    > docker-machine-driver-hyperkit.sha256: 65 B / 65 B [---] 100.00% ? p/s 0s
    > docker-machine-driver-hyperkit: 10.79 MiB / 10.79 MiB  100.00% 4.57 MiB p
🔑  The 'hyperkit' driver requires elevated permissions. The following commands will be executed:

    $ sudo chown root:wheel /Users/k-iso/.minikube/bin/docker-machine-driver-hyperkit
    $ sudo chmod u+s /Users/k-iso/.minikube/bin/docker-machine-driver-hyperkit


Password:
💿  Downloading VM boot image ...
    > minikube-v1.5.1.iso.sha256: 65 B / 65 B [--------------] 100.00% ? p/s 0s
[...]

PR #6127, opened 2019-12-19
Add build tags so that the test only runs on Darwin

tstromberg:
Fixes Linux / Travis build issue caused as a side-effect of me force-merging #5833:

```
# k8s.io/minikube/pkg/minikube/registry/drvs/hyperkit [k8s.io/minikube/pkg/minikube/registry/drvs/hyperkit.test]
pkg/minikube/registry/drvs/hyperkit/hyperkit_test.go:55:15: undefined: splitHyperKitVersion
pkg/minikube/registry/drvs/hyperkit/hyperkit_test.go:85:15: undefined: convertVersionToDate
pkg/minikube/registry/drvs/hyperkit/hyperkit_test.go:119:18: undefined: isNewerVersion
```

### DIFF
--- a/pkg/minikube/registry/drvs/hyperkit/hyperkit_test.go
+++ b/pkg/minikube/registry/drvs/hyperkit/hyperkit_test.go
@@ -1,3 +1,5 @@
+// +build darwin
+
 /*
 Copyright 2019 The Kubernetes Authors All rights reserved.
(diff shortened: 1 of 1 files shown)


govargo:
I understand why my unit test failed.
Thank you.

PR #6130, opened 2019-12-20
Bump up version to v1.6.2

medyagh:


medyagh:
The are failures on windows tests for test offline, but I can see the failures been there before this PR as well.

Issue #5144, opened 2019-08-20
Feature request: Allow customizing host path used for generated PersistentVolumes


Eelis:
Dynamic volume provisioning works great, but uses paths like

    /tmp/hostpath-provisioner/pvc-433c40c9-7cf1-40a8-91ab-f9210dceec50

which are not easy to remember and type. For development and testing, it would be much more convenient if the path could be customized to something along the lines of:

    /data/<claim>

This would make a PVC named `mysqldb` resolve to an automatically generated PV that stores the data in `/data/mysqldb`. Much easier to remember and type! :)

tstromberg:
This sounds like a good idea! PR's welcome! 

Since `/data` has some things in it already, what would you think about another path, such as `/pvc/<claim>`

Eelis:
Ah, `/pvc/<claim>` would be equally nice. :)

afbjorklund:
I thought the complaint was about the `<claim>` rather than the `/tmp/hostpath-provisioner` ?
We already know that the default paths for hostpath provisioner isn't perfect, as noted in #3318

But I'm not sure about giving human names to a machine-managed resources, ripe for conflicts ?
Seems better to have a UUID-style host directory, and then mostly worry about the mount point...

Eelis:
> I thought the complaint was about the `<claim>` rather than the `/tmp/hostpath-provisioner` ?
> We already know that the default paths for hostpath provisioner isn't perfect, as noted in #3318

This feature request (not complaint!) is about both, because I did not know about #3318. :)

> But I'm not sure about giving human names to a machine-managed resources, ripe for conflicts ?

For there to be a conflict, there would have to be two PVCs with the same name, which is already disallowed by kubernetes, right?

afbjorklund:
> For there to be a conflict, there would have to be two PVCs with the same name, which is already disallowed by kubernetes, right?

Possibly. The whole hostpath provisioner desperately needs a rewrite anyway, so maybe this is a good feature to have in the **new** one

11janci:
I could look into this, along with #3318. Did a quick look into the code, do you think it is enough to just change/parametrize the path (minikube-only change), or should we change how the PV name is created (change in sig-storage-lib-external-provisioner, affecting all implementations)?

tstromberg:
From my point-of-view,It's probably good enough to fix this for minikube. 

Would switching to https://github.com/rancher/local-path-provisioner help?

medyagh:
[...]

PR #5400, opened 2019-09-18
[WIP] Make dynamic persistent volumes path readable and configurable

11janci:
For #5144 & #3318

PR #6156, opened 2019-12-23
Customizing host path for dynamically provisioned PersistentVolumes

nanikjava:
#5144 

This fix contains few things:

* Used k8s.gcr.io/debian-base-amd64:v2.0.0 as base image to build storage-provisioner
* Modify RBAC permission used to cluster-admin


deploy/addons/storage-provisioner/storage-provisioner.yaml.tmpl
```diff
@@ -31,7 +31,7 @@ metadata:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: system:persistent-volume-provisioner
+  name: cluster-admin
```
nanikjava: This permission works fine but uncertain whether this is the correct one to use. Tried using the following

system:controller:expand-controller
system:controller:persistent-volume-binder

but it always complain permission issue either with persistentvolume/persistentvolumeclaims.

Tried defining customised RBAC inside storage-provisioner.yaml.tmp also does not work as it keep on thrown permission issue with persistentvolume(claims) and endpoints (using ClusterRole similar to defined here https://github.com/pragkent/aliyun-disk-provisioner/blob/38680a9607b0d33567be1bf0a7c57b26f0960549/deploy/rbac.yaml)

cmd/storage-provisioner/main.go
```diff
@@ -25,15 +25,18 @@ import (
 	"k8s.io/minikube/pkg/storage"
 )
 
+var pvDir = "/tmp/hostpath-provisioner"
+
 func main() {
 	// Glog requires that /tmp exists.
+	fmt.Println("This is changes is called")
```
Eelis: Is this intentional? :)

nanikjava: > Is this intentional? :)

Good catch. Will fix it. Thanks.

deploy/storage-provisioner/Dockerfile
```diff
@@ -12,6 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM scratch
-COPY out/storage-provisioner storage-provisioner
-CMD ["/storage-provisioner"]
+FROM k8s.gcr.io/debian-base-amd64:v2.0.0
```
tstromberg: Can this stay `scratch`?

cmd/storage-provisioner/main.go
```diff
@@ -25,6 +25,8 @@ import (
 	"k8s.io/minikube/pkg/storage"
 )
 
+var pvDir = "/tmp/hostpath-provisioner"
```
tstromberg: This location feels weird. Is this a placeholder?

afbjorklund: It's the paths that we are currently using, it gets bind-mounted/symlinked onto the persistent disk.

See https://minikube.sigs.k8s.io/docs/reference/persistent_volumes/  (it also uses `/tmp/hostpath_pv`)

deploy/storage-provisioner/Dockerfile
afbjorklund: I think we need to provide a build container for the copied file, right now it cheerfully builds it on the host ?

Use the normal MINIKUBE_BUILD_IN_DOCKER shenaningans, I don't see what it has do with path though...

nanikjava: > Can this stay `scratch`?

Can I resolve this conversation ? as it's clear now in the slack channel conversation that we cannot use scratch ?

afbjorklund: As long as you build static binaries (using CGO_ENABLED=0), then `scratch` should be fine.

What issues are you seeing ? We know that a _dynamic_ binary doesn't work with scratch...
But that was probably not the intention, but just a side-effect when redoing the make earlier.

I added the build container in #6257 if you want to have a look.

afbjorklund:
Is it possible to merge this PR with the previous one #5400 - with the same content ?

tstromberg:
Looks good, do you mind addressing the merge conflict?

GrahamDumpleton:
Can someone explain to me the justification for having the file system directory created use the PVC name instead of the PV name. In other words, the change at line 60 from:

```
path := path.Join(p.pvDir, options.PVName)
```

to:

```
path := path.Join(p.pvDir, options.PVC.Name)
```

This means that two different applications in different namespaces can't use the same persistent volume claim name because if they do, they end up using the same directory on the filesystem, and thus they corrupt each other if directories/file names they use overlap inside the volume. This can break so many applications. It basically means you can't deploy multiple instances of the one application in different namespaces.

This introduces a serious bug and is not a feature.

Eelis:
> This introduces a serious bug

Ah good point, I hadn't considered this.

> The obvious solution to avoid for this would have been to use path of form `root/namespace/pvc-name` if wanted more user friendly name.

Sounds good to me!

PR #6177, opened 2019-12-30
pkg/drivers/kic: add judgment of stop status before Stop

q384566678:


q384566678:
@medyagh @tstromberg PTAL

q384566678:
ping

medyagh:
not needed anymore

PR #6183, opened 2019-12-30
Add systemd patch for booting on AMD Ryzen

afbjorklund:
https://github.com/systemd/systemd/pull/12536

Backported to systemd version 240 for Ubuntu:
  [ Jeremy Soller ]
  * random-util: eat up bad RDRAND values seen on AMD CPUs.
    This fixes AMD Ryzen 3000 series failing to boot (LP: #1835809)

For #6168


afbjorklund:
As noted in the issue, this patch was confirmed to fix the booting issue on the AMD CPU

PR #6191, opened 2020-01-02
Fix a typo 'the the' -> 'the'

q384566678:

### DIFF
--- a/pkg/minikube/command/kic_runner.go
+++ b/pkg/minikube/command/kic_runner.go
@@ -67,8 +67,8 @@ func (k *kicRunner) RunCmd(cmd *exec.Cmd) (*RunResult, error) {
 	for _, env := range cmd.Env {
 		args = append(args, "-e", env)
 	}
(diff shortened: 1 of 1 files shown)


pkg/minikube/command/kic_runner.go
```diff
@@ -68,7 +68,7 @@ func (k *kicRunner) RunCmd(cmd *exec.Cmd) (*RunResult, error) {
 		args = append(args, "-e", env)
 	}
 	// specify the container and command, after this everything will be
-	// args the the command in the container rather than to docker
+	// args the command in the container rather than to docker
```
priyawadhwa: thanks for catching this! i'm not sure what the comment is trying to say though, would you mind rephrasing it to something like:

```
append container name to docker arguments. all subsequent args appended will be passed to the container instead of docker.
```

q384566678: updated, PTAL

PR #6193, opened 2020-01-02
kic/oci: parse all kind of states

q384566678:


medyagh:
Thank you @q384566678

PR #6194, opened 2020-01-02
kic: add status detection to Unpause

q384566678:
Avoid unpause a non-paused container.


q384566678:
@RA489 @afbjorklund PTAL

medyagh:
this is not needed anymore

Issue #4063, opened 2019-04-05
Setting node name via kubeadm extra config leads to a crash


simleo:
I tried setting the k8s node name with `minikube start --extra-config=kubeadm.node-name=foo`, but minikube crashed with a timeout. The kubelet logs contain `node "minikube" not found`, which is probably due to this line in `pkg/minikube/bootstrapper/kubeadm/versions.go`:

```
NewUnversionedOption(Kubelet, "hostname-override", constants.DefaultNodeName),
```

I was able to make it work with:

```
minikube start --extra-config=kubeadm.node-name=foo --extra-config=kubelet.hostname-override=foo
```

This took me by surprise, since I was expecting the `kubeadm` setting to affect the `kubelet` one in absence of any explicit settings for the latter. Maybe this just needs some extra docs, since the above workaround is available. Or perhaps a top-level `--node-name` option for `minikube start` that ensures a consistent handling of the flag in the various components.

tstromberg:
When possible, we should avoid flags that cause crashes, especially when it's easy to do the right thing. I do like your idea of a `--node-name` flag to minikube. Help wanted!

tstromberg:
It's unclear to me whether we should add a mechanism to set the node name outside of the `-p` (profile) feature. That said, I'm not sure if profiles currently sets a node name or not.

Regardless, we should do better than crashing.

PR #6200, opened 2020-01-03
Use profile name as cluster/node name

### LINKED ISSUE #2800: running minikube with --bootstrapper=kubeadm uses minikube as the hard-coded node name
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
FEATURE REQUEST

Please provide the following details:

**Environment**:

**Minikube version** (use `minikube version`): v0.25.2
- **OS** (e.g. from /etc/os-release): Red Hat Enterprise Linux Server 7.4 (Maipo)
- **VM Driver** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep DriverName`): "DriverName": "none"
- **ISO version** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep -i ISO` or `minikube ssh cat /etc/VERSION`): none
- **Install tools**:
- **Others**:
The above can be generated in one go with the following commands (can be copied and pasted directly into your terminal):
```
minikube version
echo "";
echo "OS:";
cat /etc/os-release
echo "";
echo "VM driver": 
grep DriverName ~/.minikube/machines/minikube/config.json
echo "";
echo "ISO version";
grep -i ISO ~/.minikube/machines/minikube/config.json
```

**What happened**:
I started minikube in the following way:
minikube start --vm-driver=none --bootstrapper=kubeadm
I want to use kubeadm bootstrapper. BUT I also want to change the name of node to be the same as the machine name (and not fixed to "minikube" ). when I did the above I got a node with the name "minkube" which i did not expect. and I did not manage to find a way to change the name of the node that will work. 

node name will be the same as machine name (the way it is if I use localkube). or at least have a way to change the name to what I want.

minikube start --vm-driver=none --bootstrapper=kubeadm
kubectl get nodes -> see that the node name is minikube even though the machine name is different.

**Output of `minikube logs` (if applicable)**:

laozc:
The node name is always `minikube` nowadays and this PR allows user to pick up one with `--node-name`.
If not specified, the node will be named `minikube` for the default profile name (`minikube`) or `[PROFILE]-minikube`.
For `none` driver the default node name will be set to the host name.

fixes #2800 #4063


cmd/minikube/cmd/start.go
```diff
@@ -95,6 +95,7 @@ const (
 	keepContext           = "keep-context"
 	createMount           = "mount"
 	featureGates          = "feature-gates"
+	nodeName              = "node-name"
```
medyagh: I don't think we need a new flag for node-name, in minikube's terminology profile name should be same as node name

cmd/minikube/cmd/start.go
```diff
@@ -886,6 +888,24 @@ func generateCfgFromFlags(cmd *cobra.Command, k8sVersion string, drvName string)
 		out.T(out.SuccessType, "Using image repository {{.name}}", out.V{"name": repository})
 	}
 
+	kubeNodeName := viper.GetString(nodeName)
```
medyagh: just use profile name for node name

cmd/minikube/cmd/start.go
```diff
@@ -907,6 +907,17 @@ func generateCfgFromFlags(cmd *cobra.Command, k8sVersion string, drvName string)
 		out.T(out.SuccessType, "Using image repository {{.name}}", out.V{"name": repository})
 	}
 
+	var kubeNodeName string
+	if drvName == driver.None {
```
medyagh: Is there any reason that the none driver should not have the same node name as Profile name ?
I would rather we treat all drivers same way as much as possible.

so lets have the profile name for none driver unless there is a good reason for it that I dont know ?

laozc: In the none driver the host is created by the user not minikube. If we're not setting the default host name, it will result a different node name with the host.
This issue would not exist for other drivers as those drivers create the VM and set the host name.
We may need to ask user to set the profile name to keep the same behavior as other drivers which may be confusing.

laozc:
> we also need to change the "clusterName: kubernetes" in the config template files
> for example this one
> https://github.com/kubernetes/minikube/blob/master/pkg/minikube/bootstrapper/bsutil/template/template.go#L80
> 
> make sure to change for all versions v1alpha, beta...

Let me work on it.

PR #6220, opened 2020-01-06
 Use absolute path, when calling the crictl command

### LINKED ISSUE #5547: sudo crictl: not found if installed to a directory outside of sudoers secure_path
We have some things breaking on CentOS, which does **not** have `/usr/local/bin` in the PATH.
That is: users do have it in _their_ PATH, but root does not. And `sudo` runs with the root PATH.

So if you are trying to run something that is not from a package, you will get a "command not found"
This goes both for "crio version" (when run as root) and for "sudo crictl", only work with `/usr/bin`.

One workaround is to move everything to `/usr/bin`, which is violating the [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) but whatever.

It would be nice if minikube was able to find `crio` and `crictl` - also in their _default_ location ?

This would have worked better if crio and cri-tools would be available in the default distribution.

But it doesn't (at the moment), and when you try to install it locally you run into these issues...

afbjorklund:
Fixes #5547

### DIFF
--- a/pkg/minikube/cruntime/cri.go
+++ b/pkg/minikube/cruntime/cri.go
@@ -331,8 +331,9 @@ plugin_dirs = [
 `
 )
 
-// getCommandPath returns the absolute path of a command
-func getCommandPath(cr CommandRunner, cmd string) string {
+// getCrictlPath returns the absolute path of crictl
+func getCrictlPath(cr CommandRunner) string {
+	cmd := "crictl"
 	rr, err := cr.RunCmd(exec.Command("which", cmd))
 	if err != nil {
 		return cmd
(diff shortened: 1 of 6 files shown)


afbjorklund:
Stupid unparam checker... Why would you _not_ make it a generic function, for later ?

```
pkg/minikube/cruntime/cri.go:335:39: `getCommandPath` - `cmd` always receives `"crictl"` (unparam)
```

pkg/minikube/cruntime/cri.go
```diff
@@ -331,16 +331,27 @@ plugin_dirs = [
 `
 )
 
+// getCrictlPath returns the absolute path of crictl
+func getCrictlPath(cr CommandRunner) string {
+	cmd := "crictl"
+	rr, err := cr.RunCmd(exec.Command("which", cmd))
```
priyawadhwa: you could also use [exec.LookPath](https://golang.org/pkg/os/exec/#example_LookPath) here, which should be easier than parsing stdout

afbjorklund: Not really (I do this mix-up myself, all the time), since it runs locally rather than remotely

priyawadhwa: oh of course!

afbjorklund: I actually think the docker runtime has that problem, but it is not involved in this particular PR

PR #6224, opened 2020-01-07
Add unit test for cmd/minikube/cmd/config/set.go

tiendc:
Signed-off-by: Dao Cong Tien <tiendc@vn.fujitsu.com>


medyagh:
looks good ! thank you

PR #6231, opened 2020-01-08
Update docker_registry.md

stoffen01:
Need to use docker deamon from Minikube for this method to work.

### DIFF
--- a/site/content/en/docs/Tasks/docker_registry.md
+++ b/site/content/en/docs/Tasks/docker_registry.md
@@ -23,6 +23,12 @@ Enable minikube registry addon:
 minikube addons enable registry
 ```
 
+Use the docker deamon inside Minikube:
+
+```shell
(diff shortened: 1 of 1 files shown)


stoffen01:
CLA added

afbjorklund:
The whole point of that document is to _avoid_ using the master docker daemon, but instead push the image from a local dockerd to the registry...

tstromberg:
LGTM, but needs CLA approval. Let me know if you need help getting it sorted out.

medyagh:
@tstromberg I think this PR is wrong ! the registery is an alternative to docker-env ! 

if u want to do docker-env u dont need to use registery.

priyawadhwa:
Hey @stoffen01 are you still working on this?

afbjorklund:
Thanks for contributing, open another PR if you find any issues with the minikube registry

PR #6235, opened 2020-01-08
Rebuild docker images for each release

### LINKED ISSUE #6146: Build new image for gvisor addon as part of each release
Recently https://github.com/kubernetes/minikube/pull/6000 was merged to allow a new gvisor version to be used with the addon.

This was released with v1.16.0, but the image that containts the addon was never rebuilt, as the latest date shown on gcr.io/k8s-minikube/gvisor-addon is still 26 Oct 2019.

What is needed to get a new image being pushed to the registry?

tstromberg:
Fixes #6146


afbjorklund:
The problem here seems to have been that it rebuilds the docker images for **minikube** versions ?

And not just when the versions (or contents) of the images change. So it might replace the existing...

GVISOR_IMAGE_VERSION ?= 2

STORAGE_PROVISIONER_TAG := v1.8.1

Issue #6007, opened 2019-12-03
Publish arm64 image for gcr.io/k8s-minikube/storage-provisioner


wangxiyuan:
**The exact command to reproduce the issue**:
`minikube start --vm-driver=none --kubernetes-version=v1.16.2` on an ARM machine

**The full output of the command that failed**:<details>

The `storage-provisioner` pod is falied to start.

</details>

**The output of the `minikube logs` command**:<details>

`Back-off pulling image "gcr.io/k8s-minikube/storage-provisioner-arm64:v1.8.1"`

</details>

**The operating system version**:
Ubuntu 18.04 for aarch64

sharifelgamal:
@wangxiyuan can you try this with minikube 1.6.1 and see if it's fixed?

wangxiyuan:
No, The detail:

```
root@ubuntu:~# ./minikube-linux-arm64 status
host: 
kubelet: 
apiserver: 
kubeconfig: 

root@ubuntu:~# ./minikube-linux-arm64 start --vm-driver=none --kubernetes-version=v1.16.2
* minikube v1.6.1 on Ubuntu 18.04 (/arm64)
* Selecting 'none' driver from user configuration (alternates: [])
* Running on localhost (CPUs=4, Memory=16038MB, Disk=95733MB) ...
* OS release is Ubuntu 18.04.3 LTS
* Preparing Kubernetes v1.16.2 on Docker '18.09.7' ...
E1217 01:34:18.397568    5330 cache_images.go:85] CacheImage k8s.gcr.io/coredns-arm64:1.6.2 -> /root/.minikube/cache/images/k8s.gcr.io/coredns-arm64_1.6.2 failed: MANIFEST_UNKNOWN: "Failed to fetch \"1.6.2\" from request \"/v2/coredns-arm64/manifests/1.6.2\"."
E1217 01:34:18.772220    5330 cache_images.go:85] CacheImage gcr.io/k8s-minikube/storage-provisioner-arm64:v1.8.1 -> /root/.minikube/cache/images/gcr.io/k8s-minikube/storage-provisioner-arm64_v1.8.1 failed: MANIFEST_UNKNOWN: "Failed to fetch \"v1.8.1\" from request \"/v2/k8s-minikube/storage-provisioner-arm64/manifests/v1.8.1\"."
[...]

PR #6243, opened 2020-01-08
Do not use an arch suffix for the coredns name

afbjorklund:
They are using multi-arch images instead...
So don't append something like -arm64 to it.

Just use e.g. "k8s.gcr.io/coredns:1.6.5",
and Docker will pull the right platform image.

See https://hub.docker.com/r/coredns/coredns

As seen in #6007 et al


medyagh:
@afbjorklund looks good to me also, after resolving the conflict

PR #6245, opened 2020-01-08
Remove unnecessary code to elevate system privileges

### LINKED ISSUE #6239: investigate do we need to Elevate Kube System Privileges?
while doing kic, I was able to spin up a kic cluster without doing ElevateKubeSystemPrivileges as seen 

and it worth noting, it takes a considerable amount of time to do ElevateKubeSystemPrivileges
https://github.com/kubernetes/minikube/blob/be79843f7557a1b4a652df97b04f8a946ee2808d/pkg/minikube/bootstrapper/kubeadm/kubeadm.go#L232

I would like us to investigate if we need to do this behiavour by default ? and what breaks if we dont do it ?

tstromberg:
Privilege elevation no longer appears to be necessary, at least with my testing of v1.11.10 and higher.

It only saves about ~27ms in the normal startup case, but I could see it taking longer in other cases.

Fixes #6239


medyagh:
I am curious to see if the isito addon still works without this.

sharifelgamal:
Is this still something we want to do?

tstromberg:
FWIW, istio works fine with this PR.

tstromberg:
NOTE: Based on the integration results, I wonder if this breaks the `metrics-server` addon?

tstromberg:
Closing this PR for now as it breaks the metrics-server. I am curious if this also applies to KIC.

/cc @medyagh

PR #6256, opened 2020-01-09
Add --dry-run option to start

### LINKED ISSUE #5181: Add --dry-run flag to surface configuration warnings
For embedding within an API, it would be handy to have a dry run mode to surface issues a user may have with the current configuration. For example:

- Warning if the user will need sudo permissions
- `minikube start --dry-run`
- Resource warnings (RAM, CPU)

Basically, anything that would normally cause `minikube start` to exit due to an invalid configuration should do so and output why - and do so quickly.

tstromberg:
Fixes #5181


cmd/minikube/cmd/start.go
```diff
@@ -158,6 +159,7 @@ func initMinikubeFlags() {
 
 	startCmd.Flags().Bool(force, false, "Force minikube to perform possibly dangerous operations")
 	startCmd.Flags().Bool(interactive, true, "Allow user prompts for more information")
+	startCmd.Flags().Bool(dryRun, false, "dry-run mode. Validates configuration, but does does not mutate system state")
```
medyagh: looks good, maybe say more explicitly what you mean by system state.
either here or in a doc on the site.

PR #6257, opened 2020-01-09
Allow building storage-provisioner for other arch

afbjorklund:
For #6007

Also allow building the static binary in docker

----

By setting GOARCH, you can cross-compile.
The default is to build for the local architecture.


afbjorklund:
I think that will work, would be good to add `arm` support - and drop ubuntu for the `s390x`.

Can't really verify whether that (scratch) works on the mainframe, but... it's a static binary ?

priyawadhwa:
My guess is it would work? But that's just a guess :sweat_smile:

afbjorklund:
I think you have to be _really_ interested, to set up a test machine for `s390x`:

https://astr0baby.wordpress.com/2018/06/03/installing-ubuntu-18-04-server-s390x-in-hercules-mainframe-simulator/

I would be surprised if it was more than a binary number of minikube users on it ?

And even if minikube _works_ ok on arm/arm64, we don't really support the desktop.

afbjorklund:
@priyawadhwa : syntax was close enough, Dockerfile uses `${}` while Makefile uses `$()`

PR #6258, opened 2020-01-09
Replace integration-test hacks with mutex locks around shared resources

tstromberg:
The goal here is to improve integration test reliability by moving the locks to protect the actual sensitive parts of the minikube binary. This also has the benefit of improving the life of users, rather than just our integration tests.


pkg/minikube/cluster/cluster.go
```diff
@@ -139,6 +149,20 @@ func StartHost(api libmachine.API, config cfg.MachineConfig) (*host.Host, error)
 	return h, nil
 }
 
+// acquireMachinesLock protects against code that is not parallel-safe (libmachine, cert setup)
+func acquireMachinesLock() (mutex.Releaser, error) {
+	// NOTE: Provisioning generally completes within 60 seconds
+	spec := lock.PathMutexSpec(filepath.Join(localpath.MiniPath(), "machines"))
```
medyagh: should it include the profile name ?

pkg/minikube/cluster/cluster.go
```diff
@@ -139,6 +149,20 @@ func StartHost(api libmachine.API, config cfg.MachineConfig) (*host.Host, error)
...
+	spec.Timeout = 5 * time.Minute
+
+	glog.Infof("acquiring machines lock: %+v", spec)
+	start := time.Now()
+	defer func() {
+		glog.Infof("acquired machines lock within %s", time.Since(start))
```
medyagh: it would be maybe usefull to include the profile name in the log, to find out which profile tried to access the lock.

tstromberg: Good suggestion. Done!

tstromberg:
> I am curious if we can add this to the SSH client as well ? so we have less flake in the uni test as well ?

I aim to solve the ssh unit test race a bit more gracefully: it seems like a completely in-process issue, and should not require a system-wide mutex lock.

medyagh:
wow the kvm test failures were reduces to 1 from 5 on master !

medyagh:
one of the integrtion tests failures  "old-docker" for virtualbox 

```
** stderr ** 
Error in configuration: context was not found for specified context: old-docker-20200109T160555.342534407-24524
** /stderr **
```

I wonder if we need to add this univeral lock to updating kubeconfig ?

tstromberg:
> @tstromberg
> one of the integrtion tests failures "old-docker" for virtualbox
> 
> ```
> ** stderr ** 
> Error in configuration: context was not found for specified context: old-docker-20200109T160555.342534407-24524
> ** /stderr **
> ```
> 
> I wonder if we need to add this univeral lock to updating kubeconfig ?

The error generated by the integration tests here is confusing: the context was never created. It died here:

`W0109 16:15:58.069338   30401 exit.go:101] Unable to start VM. Please investigate and run 'minikube delete' if possible: boot lock: timeout acquiring mutex`

virtualbox will take further tweaking, but I would prefer to do it in a follow-up PR to avoid conflating excessively slow integration tests with host locking.

medyagh:
> > @tstromberg
> > one of the integrtion tests failures "old-docker" for virtualbox
> > ```
> > ** stderr ** 
> > Error in configuration: context was not found for specified context: old-docker-20200109T160555.342534407-24524
> > ** /stderr **
> > ```
> > 
> > 
> > I wonder if we need to add this univeral lock to updating kubeconfig ?
> 
> The error generated by the integration tests here is confusing: the context was never created. It died here:
> 
> `W0109 16:15:58.069338 30401 exit.go:101] Unable to start VM. Please investigate and run 'minikube delete' if possible: boot lock: timeout acquiring mutex`
> 
> virtualbox will take further tweaking, but I would prefer to do it in a follow-up PR to avoid conflating excessively slow integration tests with host locking.

sounds good, we might need extra time out for virtualbox.
virtualbox is slowest one.

medyagh:
wow ! KVM tests all passed for first time in a long time !!!!

PR #6264, opened 2020-01-10
Create Hyper-V External Switch

laozc:
This PR tries to find an external switch or create one for Hyper-V driver if the user specifies -hyperv-use-external-switch.

If the adapter is not specified, the driver will first look up LAN adapters before other adapters (WiFi, ...).
Or the user may specify an adapter to attach to the external switch.

Since Windows 10 only allows one external switch for the same adapter, it will find the virtual switch before creating one. The external switch will be created and named as "minikube" if not the case.


medyagh:
@laozc  I am curious why the formatting changed so much ? we run goimports -w on all the code...it should be same ...

laozc:
> @laozc I am curious why the formatting changed so much ? we run goimports -w on all the code...it should be same ...

Do you mean the formatting in `cmd/minikube/cmd/start.go` and `pkg/minikube/config/types.go`?
I think golang would try to align the definition by column and the new field `hypervUseExternalSwitch` is one character more than existing fields.

pkg/minikube/registry/drvs/hyperv/vswitch.go
```diff
@@ -0,0 +1,212 @@
...
+func createVMSwitch(switchName string, adapter netAdapter) error {
+	err := cmd(fmt.Sprintf("Hyper-V\\New-VMSwitch -Name \"%s\" -NetAdapterInterfaceDescription \"%s\"", switchName, adapter.InterfaceDescription))
+	if err != nil {
+		return errors.Wrapf(err, "failed to create VM switch %s with adapter %s", switchName, adapter.InterfaceGUID)
+	}
+
+	return nil
+}
+
+func chooseSwitch(adapterName string) (string, error) {
```
tstromberg: I wasn't expecting a method called `chooseSwitch` to create a switch. Could we refactor it so that it passes the necessary information back in case a switch needs to be created?

Maybe like:

```
switch, err := selectSwitch(adapter)
if err != nil {
  switch, err := createSwitch(adapter)
```

pkg/minikube/registry/drvs/hyperv/vswitch.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	if len(foundSwitches) > 0 {
+		return foundSwitches[0].Name, nil
+	}
+
+	return "", nil
+}
+
+func getOrderedAdapters() ([]netAdapter, error) {
```
tstromberg: Can you add a docstring for each of these functions?

I'm unsure what the difference between `getOrderedAdapters` and `getNetAdapters` is.

pkg/minikube/registry/drvs/hyperv/vswitch.go
```diff
@@ -0,0 +1,212 @@
...
+		adapter = foundAdapters[0]
+		foundSwitch, err := findConnectedVMSwitch(adapter.InterfaceGUID)
+		if err != nil {
+			return "", err
+		}
+
+		if foundSwitch != "" {
+			return foundSwitch, nil
+		}
+
```
tstromberg: Can we return here to remove the requirement of having an else case?

pkg/minikube/registry/drvs/hyperv/vswitch.go
laozc: Done

pkg/minikube/registry/drvs/hyperv/vswitch.go
laozc: Done

pkg/minikube/registry/drvs/hyperv/vswitch.go
laozc: Done

medyagh:
@laozc thanks for resolving the code reviews, letss resolve the conflicts and I think we could merge this !

tstromberg:
Merge conflict, but otherwise seems to be ready!

medyagh:
@laozc thank you for your patience and thank you for resolving the conflict a few times, I promise this be the last time and your PR will get a high priority to be merged. 

thank you again

medyagh:
Thank you very much

PR #6265, opened 2020-01-10
drop = nil from declaration of var cniFile

q384566678:


medyagh:
this one is done amother PR I will close this one

PR #6273, opened 2020-01-10
Only grab post-mortem logs if VM is running

tstromberg:
This removes the confusing error messages that sometimes occur if minikube is unable to create a VM.


medyagh:
retest this please

PR #6276, opened 2020-01-10
Pass --kvm-qemu-uri value into kvm2 health check

### LINKED ISSUE #6275: kvm2 health check fails if LIBVIRT_DEFAULT_URI is unset
This was noticed on our upgraded GPU machine for VirtualBox testing: the health check that automatically picks which VM to use was failing from Jenkins, but not interactive use.

After some testing, I noticed the difference was that the interactive shell had LIBVIRT_DEFAULT_URI set.

It's worth noting that this only appears to have affected the health check. If you disregarded it, minikube ran perfectly with the kvm2 driver.

tstromberg:
Fixes #6275 as well as the VirtualBox Linux integration test.

### DIFF
--- a/pkg/minikube/registry/drvs/kvm2/kvm2.go
+++ b/pkg/minikube/registry/drvs/kvm2/kvm2.go
@@ -20,6 +20,7 @@ package kvm2
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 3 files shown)


pkg/minikube/registry/drvs/kvm2/kvm2.go
```diff
@@ -86,13 +87,25 @@ func configure(mc config.MachineConfig) interface{} {
 	}
 }
 
+// defaultURI returns the QEMU URI to connect to for health checks
+func defaultURI() string {
+	u := os.Getenv("LIBVIRT_DEFAULT_URI")
+	if u != "" {
+		return u
+	}
```
medyagh: also should check the start flag  --kvm-qemu-uri if set or not

PR #6283, opened 2020-01-12
Use the complete filenames for the packages

### LINKED ISSUE #6281: Add architecture (and revision) to the packages
Currently we truncate the filenames, to make it "easier" for users:
* [minikube_1.6.2.deb](https://github.com/kubernetes/minikube/releases/download/v1.6.2/minikube_1.6.2.deb)
* [minikube-1.6.2.rpm](https://github.com/kubernetes/minikube/releases/download/v1.6.2/minikube-1.6.2.rpm)

We should stop doing this, and use the **real** filenames instead.
* minikube_1.6.2-0_amd64.deb
* minikube-1.6.2-0.x86_64.rpm

This will make things less weird, when we support additional architectures...
* minikube_1.6.2-0_arm64.deb
* minikube-1.6.2-0.aarch64.rpm

For the raw minikube binary, we already include "-linux-amd64" in the filename.

----

The release is kept as 0, in case any distributions start packaging minikube...

* revision: https://www.debian.org/doc/debian-policy/ch-controlfields.html

* release: http://ftp.rpm.org/max-rpm/s1-rpm-build-creating-spec-file.html

And we still need to provide a real repository (apt, yum) for the vendor packages.

* apt: #3110

* yum: #4716

It's not like anyone _wants_ to look after versions and architectures, but instead do:

* `sudo apt install minikube`

* `sudo yum install minikube`

afbjorklund:
Follow standard file naming practices, for each package manager.

This opens up for allowing additional architectures, in future.

Closes #6281

### DIFF
--- a/hack/jenkins/release_build_and_upload.sh
+++ b/hack/jenkins/release_build_and_upload.sh
@@ -43,10 +43,10 @@ env GOPATH=$HOME/go BUILD_IN_DOCKER=y \
   make -j 16 \
   all \
   out/minikube-installer.exe \
-  "out/minikube_${DEB_VERSION}.deb" \
-  "out/minikube-${RPM_VERSION}.rpm" \
-  "out/docker-machine-driver-kvm2_${DEB_VERSION}.deb" \
-  "out/docker-machine-driver-kvm2-${RPM_VERSION}.rpm"
+  "out/minikube_${DEB_VERSION}-0_amd64.deb" \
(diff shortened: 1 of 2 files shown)


tstromberg:
Do you mind splitting out the docs change? 

Unfortunately our docs don't have a version switch yet, so they implicitly are for the most recent release. Merging this would result in the docs being wrong until v1.7 is merged.

Issue #4172, opened 2019-04-29
none: detect cgroupfs driver for kubelet configuration


itsallonetome:
`sudo minikube start --vm-driver=none --docker-opt bip='172.18.0.1/24'`

```
o   minikube v1.0.0 on linux (amd64)
$   Downloading Kubernetes v1.14.0 images in the background ...
2019/04/29 16:11:55 Unable to read "/home/k8s/.docker/config.json": open /home/k8s/.docker/config.json: no such file or directory
2019/04/29 16:11:55 No matching credentials were found, falling back on anonymous
>   Creating none VM (CPUs=2, Memory=2048MB, Disk=20000MB) ...
,,,
-   "minikube" IP address is 10.139.24.86
-   Configuring Docker as the container runtime ...
    - opt bip=172.18.0.1/24
-   Version of container runtime is 18.09.5
:   Waiting for image downloads to complete ...
-   Preparing Kubernetes environment ...
@   Downloading kubeadm v1.14.0
@   Downloading kubelet v1.14.0
-   Pulling images required by Kubernetes v1.14.0 ...
-   Launching Kubernetes v1.14.0 using kubeadm ...

!   Error starting cluster: kubeadm init:
sudo /usr/bin/kubeadm init --config /var/lib/kubeadm.yaml  --ignore-preflight-errors=DirAvailable--etc-kubernetes-manifests --ignore-preflight-errors=DirAvailable--data-minikube --ignore-preflight-errors=Port-10250 --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-kube-scheduler.yaml --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-kube-apiserver.yaml --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-kube-controller-manager.yaml --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-etcd.yaml --ignore-preflight-errors=Swap --ignore-preflight-errors=CRI


: running command:
[...]

PR #6287, opened 2020-01-12
Use CGroupDriver function from cruntime for kubelet

afbjorklund:
For #4172

### DIFF
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -157,12 +157,11 @@ func (k *Bootstrapper) StartCluster(k8s config.KubernetesConfig) error {
 		return errors.Wrap(err, "parsing kubernetes version")
 	}
 
+	extraFlags := bsutil.CreateFlagsFromExtraArgs(k8s.ExtraOptions)
 	r, err := cruntime.New(cruntime.Config{Type: k8s.ContainerRuntime})
 	if err != nil {
 		return err
 	}
-
-	extraFlags := bsutil.CreateFlagsFromExtraArgs(k8s.ExtraOptions)
 
 	ignore := []string{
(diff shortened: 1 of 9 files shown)


pkg/minikube/cruntime/containerd.go
```diff
@@ -240,6 +240,29 @@ func (r *Containerd) LoadImage(path string) error {
...
+		return "", errors.Wrapf(err, "missing config")
+	}
+	config, ok := info["config"].(map[string]interface{})
+	if !ok {
+		return "", errors.Wrapf(err, "config not map")
+	}
+	cgroupManager := "cgroupfs" // default
+	switch config["systemdCgroup"] {
+	case false:
+		cgroupManager = "cgroupfs"
```
medyagh: just a nit but could put it in a const ?

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -157,12 +157,13 @@ func (k *Bootstrapper) StartCluster(k8s config.KubernetesConfig) error {
 		return errors.Wrap(err, "parsing kubernetes version")
 	}
 
-	extraFlags := bsutil.CreateFlagsFromExtraArgs(k8s.ExtraOptions)
 	r, err := cruntime.New(cruntime.Config{Type: k8s.ContainerRuntime})
 	if err != nil {
 		return err
 	}
 
+	extraFlags := bsutil.CreateFlagsFromExtraArgs(k8s.ExtraOptions)
```
medyagh: this change needs to be added to the kicbs too
sorry for the inconvinence. the plan is to merge kicbs and kubeadm into one soon.

afbjorklund: I actually think this change was unnecessary, so I will remove it (for both)

medyagh:
the test restults show we are hitting github api limit
https://storage.googleapis.com/minikube-builds/logs/6287/KVM_Linux.txt

PR #6306, opened 2020-01-14
Add a clear example of minikube tunnel to the docs

medyagh:
I am not sure on the cmd description that we have in minikube but I left it untouched it says

```
tunnel creates a route to services deployed with type LoadBalancer and sets their Ingress to their ClusterIP.
```
I think it would be better to say it sets external ip for the service type 

but for this PR I added a clear example of using tunnel

### DIFF
--- a/site/content/en/docs/Tasks/loadbalancer.md
+++ b/site/content/en/docs/Tasks/loadbalancer.md
@@ -67,7 +67,7 @@ kubectl get svc
 <pre>
  $ kc get svc
 NAME             TYPE           CLUSTER-IP      EXTERNAL-IP     PORT(S)          AGE
-hello-minikube   LoadBalancer   10.96.184.178   10.96.184.178   8080:30791/TCP   40s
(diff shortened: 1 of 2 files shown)


site/content/en/docs/Tasks/loadbalancer.md
```diff
@@ -16,29 +16,72 @@ A LoadBalancer service is the standard way to expose a service to the internet.
...
-                loadbalancer emulator: no errors
+### Try in your browser
+open in your browser (make sure there is no proxy set)
+```
+http://REPLACE_WITH_EXTERNAL_IP:8080
 ```
 
 
-`minikube tunnel` runs as a separate daemon, creating a network route on the host to the service CIDR of the cluster using the cluster's IP address as a gateway.  The tunnel command exposes the external IP directly to any program running on the host operating system.
+Each service will get it is own external ip.
```
priyawadhwa: nit: "It is" -> "its"

PR #6318, opened 2020-01-15
Add the Node interface

### LINKED ISSUE #6298: Create the Node interface for multinode
ref #94

sharifelgamal:
Fixes #6298


pkg/minikube/assets/vm_assets.go
```diff
@@ -76,7 +77,7 @@ func (b *BaseAsset) GetModTime() (time.Time, error) {
 // FileAsset is an asset using a file
 type FileAsset struct {
 	BaseAsset
-	reader io.Reader
+	reader *io.SectionReader
```
tstromberg: Instead of specifying an exact type, can this be an io.ReadSeaker interface?

pkg/minikube/assets/vm_assets.go
```diff
@@ -157,7 +174,7 @@ func NewMemoryAsset(d []byte, targetDir, targetName, permissions string) *Memory
 // BinAsset is a bindata (binary data) asset
 type BinAsset struct {
 	BaseAsset
-	reader   io.Reader
+	reader   *bytes.Reader
```
tstromberg: io.ReadSeaker?

pkg/minikube/assets/vm_assets.go
```diff
@@ -117,17 +123,23 @@ func (f *FileAsset) GetModTime() (time.Time, error) {
...
+// Seek resets the reader to offset
+func (f *FileAsset) Seek(offset int64, whence int) (int64, error) {
+	return f.reader.Seek(offset, whence)
+}
+
 // MemoryAsset is a memory-based asset
 type MemoryAsset struct {
 	BaseAsset
-	reader io.Reader
+	reader *bytes.Reader
```
tstromberg: io.ReadSeeker?

pkg/minikube/config/node.go
```diff
@@ -0,0 +1,36 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+// AddNode adds a new node config to an existing cluster.
+func AddNode(cc *MachineConfig, name string, controlPlane bool, k8sVersion string, profileName string) error {
+	t := Worker
```
tstromberg: Without context, this feels somewhat confusing (What is worker? What is T going to be used for?). What do you feel about introducing the context first by supplying a default value?

```
n := Node{Name: name, Type: Worker}
if controlPlane {
  n.Type = Master
}
```

tstromberg: Master is a fuzzy term in Kubernetes, and isn't described in the official glossary any longer:  https://kubernetes.io/docs/reference/glossary/?fundamental=true

Since machines can be both workers and run the control plane, should the node object be able to describe all 3 possibilities?

For example:

```
Node.ControlPlane bool 
Node.Worker bool
```

That may be easier than working with an array of roles. It does however mean you could end up with a Node that isn't a worker, which would be a bit odd. Do whatever you feel is most readable.

pkg/minikube/config/profile.go
```diff
@@ -35,22 +36,31 @@ func (p *Profile) IsValid() bool {
...
-		return false
+	for _, n := range p.Config.Nodes {
+		if n.KubernetesVersion == "" {
+			return false
+		}
 	}
 	return true
 }
 
+// MasterNode gets the node specific config for the control plane
```
tstromberg: mention the behavior if there is more than one control-plane machine. I understand we do not support this configuration yet, however.

Instead of `master`, how about `Primary` for this case?

PR #6322, opened 2020-01-15
Configure etcd and kube-proxy metrics to listen on minikube node IP

chancez:
Configures etcd and kube-proxy to expose their metrics on the minikube node IP
instead of localhost so that their metrics can be scraped by Prometheus.

Metrics are already exposed for other control plane services such as kube
apiserver, controller-manager, and scheduler, this simply updates etcd and
kube-proxy to match.

This is desirable so that Prometheus monitoring works within minikube.


medyagh:
the ingress integraiton tests seems to be broken across all ... could it be realted?

chancez:
I see registry failing in hyperkit and ingress failing in kvm in the CI. It's possible this change is creating an invalid config or something. I'll run integration tests locally and see what I can find.

Issue #6308, opened 2020-01-14
hitting github api rate limit


medyagh:
one of our tools is mis-using the github api

as seen in https://storage.googleapis.com/minikube-builds/logs/6272/KVM_Linux.txt

```
HTTP code 403! Retrying in 128 ..
{
  "message": "API rate limit exceeded for user ID 20374350.",
  "documentation_url": "https://developer.github.com/v3/#rate-limiting"
}
HTTP code 403! Retrying in 256 ..
Build step 'Execute shell' marked build as failure

```

afbjorklund:
Possibly machine: https://github.com/docker/machine/blob/master/README.md#you-may-be-getting-rate-limited-by-github-error-message

medyagh:
update: I generated new API tokens and it seems to be working. I will close this issue for now till we see it agian.

medyagh:
nope that didnt help at all ! 

still getting

```
HTTP code 403! Retrying in 128 ..
{
  "message": "API rate limit exceeded for user ID 20374350.",
  "documentation_url": "https://developer.github.com/v3/#rate-limiting"
}
```

PR #6338, opened 2020-01-17
unify bootstrappers for both VM and contianer

medyagh:
This PR gets rid of kicbs.

for early experiemintal verison of kic I had made a separate bootstrapper for kic
this was useful for refactoring the code. now time to unify the bootstrapper for both vm and container.


medyagh:
the linux vbox test is available  at https://storage.googleapis.com/minikube-builds/logs/6338/VirtualBox_Linux.html


but github api rate limit didnt allow it to be updated  https://github.com/kubernetes/minikube/issues/6308

PR #6348, opened 2020-01-19
fix grammer and remove extra spaces

oke-py:
Only edited a document.


afbjorklund:
/approve
/lgtm

Is it not "grammar" ? :-)

Issue #6350, opened 2020-01-19
Add support for using the podman service within minikube


afbjorklund:
Currently we offer support for using the docker daemon remotely, through `docker-env`.

https://minikube.sigs.k8s.io/docs/tasks/docker_daemon/

It is possible to offer the same kind of support for podman-remote, using `podman-env`.

https://github.com/containers/libpod/blob/master/docs/tutorials/remote_client.md

----

This uses the [varlink](https://varlink.org) protocol over ssh, in order to start a podman socket on the VM.

Unlike Docker, there is no need to start any long-running daemon before connecting...

Currently the only supported podman commands are: `load`, `save`, `build`, `images`

There are currently some issues with running them, so use Kubernetes for that instead.

----

Any images built are immediately available for the CRI-O runtime, for rapid turnaround.

When issuing the `podman-remote build` command, the build context is sent to the VM.

afbjorklund:
Will need to add some minor documentation, on how to actually use this thing.

Maybe just some basic information, that would be rather similar to the above:

`eval $(minikube podman-env)`

`podman-remote ps`

----

Maybe one should should mention that podman and crio **don't** share containers:

```console
$ podman-remote ps
CONTAINER ID  IMAGE  COMMAND  CREATED  STATUS  PORTS  NAMES
```

Or that the remote podman does not implement image listing, that only works locally:

```console
$ podman-remote images
Error: unable to get images: org.varlink.service.MethodNotFound
```

But except for those little quirks, it works fine.

```
$ podman-remote version
Client:
Version:            1.7.0
RemoteAPI Version:  1
Go Version:         go1.13.6
Git Commit:         b7ce1157b00af09f4a09e39b377aa3abff46ee05
Built:              Sat Jan 18 15:26:34 2020
OS/Arch:            linux/amd64

Server:
Version:            1.6.4
RemoteAPI Version:  1
Go Version:         go1.11.13
Git Commit:         5cc92849f7fc9dd734ca2fd8f3ae8830b9a7eb26
Built:              Sat Jan  4 13:49:24 2020
OS/Arch:            linux/amd64
```

```console
$ podman-remote build -t hello-node minikube
STEP 1: FROM node:6.14.2
Getting image source signatures
Copying blob sha256:0c5cf711b890ca3e9df1635796c969d71320570f97bdd8d1a9e857f85ec6f33c
[...]

PR #6351, opened 2020-01-19
Add podman-env for connecting with podman-remote

afbjorklund:
For #6350


tstromberg:
Part of me wishes we had a vendor-neutral command we could point users to, but I think any terminology we use might be lost on them. docker-env & podman-env are likely more meaningful than any alternative we come up with.

tstromberg:
Merge conflict :(

afbjorklund:
I’m not sure `env --container-runtime=docker` would have really _improved_ the situation ? But that was the original machine command...

gbraad:
> docker-env & podman-env 

you could call it `runtime-env`... we have considered the same. it is in line with seting the runtime option. but just not sure if this is preferred. just wondering, does the code error when `docker-env` is used while the runtime is podman?

afbjorklund:
> just wondering, does the code error when `docker-env` is used while the runtime is podman? (and vice versa)

There will be an error if docker is not running or podman is not installed, but otherwise you're free to choose...

```
$ minikube docker-env
💣  The docker service is currently not active
```

```
$  ./out/minikube podman-env
💣  The podman service is currently not available
```

It is quite common to use docker to build images for `containerd`, or to use podman to build images for `cri-o`.

afbjorklund:
I will have to rewrite this

afbjorklund:
I moved the tests from the pkg (generic) to the cmd (specific), as per the new code at `master`.

There were a lot of symbol conflicts with "env", so had to prefix a lot of things (miss my `static`)

medyagh:
@afbjorklund any chance we could have integraiton test for this or documentation how to use it ?

afbjorklund:
@medyagh : normally I try to write the user-facing documentation in the issue (not the PR)

Included a doc page: https://deploy-preview-6351--kubernetes-sigs-minikube.netlify.com/docs/tasks/podman_service/

afbjorklund:
Stuck on lint

```
pkg/minikube/shell/shell.go:85:2: var `unset` is unused (unused)
cmd/minikube/cmd/podman-env.go:216:60: podmanEnvVars - result 1 (error) is always nil (unparam)
```

Honiix:
Hi @afbjorklund, Thanks for your work!
I landed here because I was reading this: https://minikube.sigs.k8s.io/docs/tasks/podman_service/
But I got "Error: unknown command "podman-env" for "minikube".
It seems to be because this PR is merged but not included in the [last release (1.7.2)](https://github.com/kubernetes/minikube/releases/tag/v1.7.2). Right?

If it's the case, it may be a good idea to release the doc at the same time as the code. Is it feasible?

afbjorklund:
It should be in 1.7.3 now, unfortunately (?) the docs are showing `master`

Issue #6364, opened 2020-01-22
TestFunctional/parallel/DryRun: exit code = -1, wanted = 78: signal: killed


medyagh:
I believe the dry run returns a different exist code on kvm for some reason !
as seen in https://storage.googleapis.com/minikube-builds/logs/6338/KVM_Linux.html#failtestwindow5

```
=== RUN   TestFunctional/parallel/DryRun
=== PAUSE TestFunctional/parallel/DryRun
=== CONT  TestFunctional/parallel/DryRun
--- FAIL: TestFunctional/parallel/DryRun (3.07s)
functional_test.go:313: (dbg) Run:  out/minikube-linux-amd64 start -p functional-20200121T174445.238510036-4853 --dry-run --memory 250MB --vm-driver=kvm2 
functional_test.go:313: (dbg) Non-zero exit: out/minikube-linux-amd64 start -p functional-20200121T174445.238510036-4853 --dry-run --memory 250MB --vm-driver=kvm2 : signal: killed (2.001806296s)
-- stdout --
* [functional-20200121T174445.238510036-4853] minikube v1.7.0-beta.0 on Debian 9.11
- KUBECONFIG=/home/jenkins/minikube-integration/linux-amd64-kvm2-6338-2439-5a3c13815ed5bb198a92b2183d9542af4462ddb3/kubeconfig
- MINIKUBE_BIN=out/minikube-linux-amd64
- MINIKUBE_HOME=/home/jenkins/minikube-integration/linux-amd64-kvm2-6338-2439-5a3c13815ed5bb198a92b2183d9542af4462ddb3/.minikube
- MINIKUBE_LOCATION=6338
-- /stdout --
functional_test.go:317: dry-run(250MB) exit code = -1, wanted = 78: signal: killed
functional_test.go:324: (dbg) Run:  out/minikube-linux-amd64 start -p functional-20200121T174445.238510036-4853 --dry-run --vm-driver=kvm2 
functional_test.go:324: (dbg) Done: out/minikube-linux-amd64 start -p functional-20200121T174445.238510036-4853 --dry-run --vm-driver=kvm2 : (1.071343365s)
```

but not seen in same test for vbox https://storage.googleapis.com/minikube-builds/logs/6338/VirtualBox_Linux.html

medyagh:
@tstromberg

tstromberg:
It sounds like it ran out of the strict time limit (2s). On my machine it takes 2s. I'll add some logging.

[...]

PR #6370, opened 2020-01-22
Double dry-run time limit, add logging

tstromberg:
To address #6364


medyagh:
the logs for kvm show it fixes it
 https://storage.googleapis.com/minikube-builds/logs/6370/KVM_Linux.html

PR #6377, opened 2020-01-22
Bump default Kubernetes version to v1.17.2, latest to v1.18.0-alpha.2

tstromberg:
Please carefully look at latest test result to make sure we don't break it.


medyagh:
I suspect we might need v2beta2 template kubeadm ( there is a working PR but not complete)

PR #6422, opened 2020-01-29
Add a 2-second timeout to driver detection commands

### LINKED ISSUE #6228: Driver auto-detection can hang minikube
Minikube versions in 1.5.x all work properly and proceed properly after running `minikube start --vm-driver=virtualbox` but for each version of 1.6.x that I've tried, everything hangs.  

The logs seem to indicate that a `config.json` is missing, but from advice I've seen, this should be created upon `minikube start` and should not be manually added.  I've tried switching back and forth between versions and each time the 1.5.2 works properly, but 1.6.0/1.6.2 always hangs (I've left it for 20+ minutes).

I am running with a ryzen CPU, but not sure if this is related to #6168 as the examples mentioned there seem to get further in the process after `minikube start` than mine does.

`minikube start --vm-driver=virtualbox`

```
$ minikube start --vm-driver=virtualbox
🎉  minikube 1.6.2 is available! Download it: https://github.com/kubernetes/minikube/releases/tag/v1.6.2
💡  To disable this notice, run: 'minikube config set WantUpdateNotification false'

🙄  minikube v1.6.0 on Ubuntu 19.10
```

</details>

```
$ minikube logs

💣  Error getting config: stat /home/myuser/.minikube/profiles/minikube/config.json: no such file or directory

```

</details>

Ubuntu 19.10

tstromberg:
Confirmed locally. Fixes #6228

Logs will show:

`I0129 10:59:07.582677    9816 driver.go:154] not recommending "docker" due to health: signal: killed`

### DIFF
--- a/pkg/minikube/registry/drvs/hyperkit/hyperkit.go
+++ b/pkg/minikube/registry/drvs/hyperkit/hyperkit.go
@@ -19,6 +19,7 @@ limitations under the License.
 package hyperkit
 
 import (
+	"context"
 	"fmt"
 	"os/exec"
 	"strings"
@@ -87,7 +88,11 @@ func status() registry.State {
 		return registry.State{Error: err, Fix: "Run 'brew install hyperkit'", Doc: docURL}
 	}
 
-	cmd := exec.Command(path, "-v")
+	// Allow no more than 2 seconds for querying state
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	cmd := exec.CommandContext(ctx, path, "-v")
 	out, err := cmd.CombinedOutput()
 	if err != nil {
 		return registry.State{Installed: true, Error: fmt.Errorf("%s failed:\n%s", strings.Join(cmd.Args, " "), out), Fix: "Run 'brew install hyperkit'", Doc: docURL}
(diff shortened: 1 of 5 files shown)


pkg/minikube/registry/drvs/hyperkit/hyperkit.go
```diff
@@ -87,7 +88,11 @@ func status() registry.State {
 		return registry.State{Error: err, Fix: "Run 'brew install hyperkit'", Doc: docURL}
 	}
 
-	cmd := exec.Command(path, "-v")
+	// Allow no more than 2 seconds for querying state
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
```
medyagh: some times docker takes more than 2 secs, how about 3 ? in honor of three "good algorithm, good syntax, good e2e test"

tstromberg: I was going to send the PR in as 1 second, because it seemed to be a good hard limit for auto-detection. Asking a user to wait 3 seconds for a bad hypervisor they are not intending to use is asking too much.

To make e2e stable, I doubled it. None of these commands should ever take >250ms to execute. If they do, we're running the wrong command.

`docker info` reliably takes 0.10s on Linux on macOS, so the timeout is already 20x what normal should be. I'd be happy to raise it if we see evidence of it causing flakiness, but I would aim for less time (1.5s?) rather than more time spent in autodetection.

Issue #6404, opened 2020-01-28
Make minikube work in Google Cloud Shell (using kic)


medyagh:
also related https://github.com/kubernetes/minikube/issues/5900

```
medya@cloudshell:~ (k8s-minikube)$ ./minikube start --vm-driver=docker
* minikube v1.7.0-beta.0 on Debian 9.11
* Selecting experimental 'docker' driver from user configuration (alternates: [])
* Creating Kubernetes in docker container with (CPUs=2), Memory=2000MB (1995MB available) ...
*
X Failed to enable container runtime: br_netfilter: command failed: [docker exec --privileged minikube sudo modprobe br_netfilter]: exit status 1
*
* minikube is exiting due to an error. If the above message is not useful, open an issue:
  - 

```

currently it fails at this line :
```
func enableIPForwarding(cr CommandRunner) error {
	c := exec.Command("sudo", "modprobe", "br_netfilter")
	if _, err := cr.RunCmd(c); err != nil {
		return errors.Wrap(err, "br_netfilter")
	}
}
```

running the command manually I get this output

```
$ docker exec -it minikube sudo modprobe br_netfilter
modprobe: ERROR: ../libkmod/libkmod.c:586 kmod_search_moddep() could not open moddep file '/lib/modules/4.19.79+/modules.dep.bin'
modprobe: FATAL: Module br_netfilter not found in directory /lib/modules/4.19.79+
```

medyagh:
I tried to not fail on modeprob net_filter

and get this error
```
medya@cloudshell:~/workspace/minikube (k8s-minikube)$ ./out/minikube start --vm-driver=docker
* minikube v1.7.0-beta.1 on Debian 9.11
* Selecting experimental 'docker' driver from user configuration (alternates: [])
* Creating Kubernetes in docker container with (CPUs=2), Memory=2000MB (1995MB available) ...
* Preparing Kubernetes v1.17.2 on containerd 1.3.0-27-g54658b88 ...
  - kubeadm.pod-network-cidr=10.244.0.0/16
* Downloading kubectl v1.17.2
* Downloading kubeadm v1.17.2
* Downloading kubelet v1.17.2
* Pulling images ...
* Launching Kubernetes ...
E0128 19:44:26.300018    4887 logs.go:174] Failed to list containers for "kube-apiserver": crictl list: sudo crictl ps -a --quiet --name=kube-api
server: exit status 1
stdout:

stderr:
[...]

Issue #5900, opened 2019-11-13
none: Failing to load a module due to module being built-in (br_netfilter)


battlesnake:
[

It looks like minikube is attempting to load a module that's built-in (in my case), and is then failing when the module isn't found.

**The exact command to reproduce the issue**:

```
$ sudo minikube start --network-plugin=cni  --container-runtime=containerd --enable-default-cni --vm-driver=none
```

**The full output of the command that failed**:<details>

```
$ sudo minikube start --network-plugin=cni  --container-runtime=containerd --enable-default-cni --vm-driver=none
😄  minikube v1.5.2 on Arch rolling (vbox/amd64)
💡  Tip: Use 'minikube start -p <name>' to create a new cluster, or 'minikube delete' to delete this one.
🏃  Using the running none "minikube" VM ...
⌛  Waiting for the host to be provisioned ...

💣  Failed to enable container runtime: br_netfilter: command failed: sudo modprobe br_netfilter
stdout:
stderr: modprobe: FATAL: Module br_netfilter not found in directory /lib/modules/5.3.8-arch1-1
: exit status 1

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
👉  https://github.com/kubernetes/minikube/issues/new/choose
```

```
$ lsmod | grep br_netfilter
br_netfilter           28672  0
bridge                212992  1 br_netfilter
```

</details>

**The output of the `minikube logs` command**:<details>

```
$ sudo minikube logs
E1113 13:57:05.724655 3932286 logs.go:175] Failed to list containers for "kube-apiserver": command failed: sudo crictl ps -a --name=kube-apiserver --state=Running --quiet
stdout:
stderr: sudo: crictl: command not found
: exit status 1
E1113 13:57:05.733915 3932286 logs.go:175] Failed to list containers for "coredns": command failed: sudo crictl ps -a --name=coredns --state=Running --quiet
stdout:
stderr: sudo: crictl: command not found
: exit status 1
E1113 13:57:05.743823 3932286 logs.go:175] Failed to list containers for "kube-scheduler": command failed: sudo crictl ps -a --name=kube-scheduler --state=Running --quiet
stdout:
stderr: sudo: crictl: command not found
: exit status 1
E1113 13:57:05.752668 3932286 logs.go:175] Failed to list containers for "kube-proxy": command failed: sudo crictl ps -a --name=kube-proxy --state=Running --quiet
stdout:
stderr: sudo: crictl: command not found
[...]

PR #6427, opened 2020-01-29
improve checking modprob netfilter

medyagh:
fixes this https://github.com/kubernetes/minikube/issues/5900 (none)
and helps with https://github.com/kubernetes/minikube/issues/6404  (docker on cloud shell)

based on suggestion by @tstromberg here https://github.com/kubernetes/minikube/issues/5900#issuecomment-554189392

### DIFF
--- a/pkg/minikube/cruntime/cruntime.go
+++ b/pkg/minikube/cruntime/cruntime.go
@@ -169,10 +169,10 @@ func disableOthers(me Manager, cr CommandRunner) error {
 func enableIPForwarding(cr CommandRunner) error {
 	c := exec.Command("sudo", "sysctl", "net.netfilter.nf_conntrack_count")
 	if rr, err := cr.RunCmd(c); err != nil {
-		glog.Infof("couldn't verify netfilter by %q which might be okay. output=%q error=%v", rr.Command(), rr.Output(), err)
+		glog.Infof("couldn't verify netfilter by %q which might be okay. error: %v", rr.Command(), err)
 		c = exec.Command("sudo", "modprobe", "br_netfilter")
-		if rr, err := cr.RunCmd(c); err != nil {
-			return errors.Wrapf(err, "br_netfilter. output: %s", rr.Output())
+		if _, err := cr.RunCmd(c); err != nil {
+			return errors.Wrapf(err, "br_netfilter")
 		}
 	}
 	c = exec.Command("sudo", "sh", "-c", "echo 1 > /proc/sys/net/ipv4/ip_forward")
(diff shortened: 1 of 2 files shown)


pkg/minikube/cruntime/cruntime.go
```diff
@@ -167,11 +167,14 @@ func disableOthers(me Manager, cr CommandRunner) error {
...
 func enableIPForwarding(cr CommandRunner) error {
-	c := exec.Command("sudo", "modprobe", "br_netfilter")
-	if _, err := cr.RunCmd(c); err != nil {
-		return errors.Wrap(err, "br_netfilter")
+	c := exec.Command("sudo", "sysctl", "net.netfilter.nf_conntrack_count")
+	if rr, err := cr.RunCmd(c); err != nil {
+		glog.Infof("couldn't verify netfilter by %q which might be okay. output=%q error=%v", rr.Command(), rr.Output(), err)
+		c = exec.Command("sudo", "modprobe", "br_netfilter")
+		if rr, err := cr.RunCmd(c); err != nil {
+			return errors.Wrapf(err, "br_netfilter. output: %s", rr.Output())
```
tstromberg: Here too.

pkg/minikube/cruntime/cruntime.go
```diff
@@ -167,11 +167,14 @@ func disableOthers(me Manager, cr CommandRunner) error {
 // enableIPForwarding configures IP forwarding, which is handled normally by Docker
 // Context: https://github.com/kubernetes/kubeadm/issues/1062
 func enableIPForwarding(cr CommandRunner) error {
-	c := exec.Command("sudo", "modprobe", "br_netfilter")
-	if _, err := cr.RunCmd(c); err != nil {
-		return errors.Wrap(err, "br_netfilter")
+	c := exec.Command("sudo", "sysctl", "net.netfilter.nf_conntrack_count")
+	if rr, err := cr.RunCmd(c); err != nil {
+		glog.Infof("couldn't verify netfilter by %q which might be okay. output=%q error=%v", rr.Command(), rr.Output(), err)
```
tstromberg: I believe rr.Output() here is redundant: RunCmd errors already include stdout/stderr, or at least non-kic drivers do. Can we fix kic's RunCmd to be consistent instead?

medyagh: sure thing ! I

medyagh: I just noticed we have same thing in kic as well.
```
fmt.Errorf("%s: %v\nstdout:\n%s\nstderr:\n%s", rr.Command(), err, rr.Stdout.String(), rr.Stderr.String())```

PR #6438, opened 2020-01-30
status: Explicitly state that the cluster does not exist

tstromberg:
While debugging a local integration test failure that said the cluster status was "", I dug in and noticed that "minikube status" on a nonexistent cluster emits no error:

```
host: 
kubelet: 
apiserver: 
kubeconfig: 
```

This PR changes the behavior to emit an error to stderr, and populate the fields with "Nonexistent":

```
E0130 14:43:04.992476    7478 status.go:98] The "minikube" cluster does not exist!
host: Nonexistent
kubelet: Nonexistent
apiserver: Nonexistent
kubeconfig: Nonexistent
```

With `-o json`, it emits:

```
E0130 14:43:07.520741    7534 status.go:98] The "minikube" cluster does not exist!
{"Host":"Nonexistent","Kubelet":"Nonexistent","APIServer":"Nonexistent","Kubeconfig":"Nonexistent"} 
```

### DIFF
--- a/cmd/minikube/cmd/status.go
+++ b/cmd/minikube/cmd/status.go
@@ -45,8 +45,10 @@ var output string
 
 const (
 	// Additional states used by kubeconfig
-	Configured    = "Configured"    // analogous to state.Saved
-	Misconfigured = "Misconfigured" // analogous to state.Error
+	Configured    = "Configured"    // ~state.Saved
+	Misconfigured = "Misconfigured" // ~state.Error
+	// Additional states used for clarity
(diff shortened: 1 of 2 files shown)


sharifelgamal:
is the line printed to stderr going to break stuff ingesting the json output?

tstromberg:
> is the line printed to stderr going to break stuff ingesting the json output?

It could: but that's fine with me: stderr is never designed to be parsed, as it may always contain a random unrelated error message.

Issue #5727, opened 2019-10-25
Changing service cidr range breaks DNS resolution


jgoeres:
Our company intranet uses large parts of the 10.0.0.1/8 subnet. Alas, the default service CIDR range used by minikube (resp. kubeadm?) is 10.96.0.*. To avoid problems, I wanted to change pod and service CIDR range with these two options passed to minikube start:

--service-cluster-ip-range='172.20.0.0/14' --extra-config=kubelet.pod-cidr=172.16.0.0/14 

The VM seemed to startup fine, but after deploying our application, I noticed that the DNS names of services could not be resolved by our pods, e.g.

    nslookup our-service
    nslookup: can't resolve '(null)': Name does not resolve

I checked /etc/resolv.conf inside the pod, and indeed, it still pointed to the default IP of the kubedns service, i.e., 10.96.0.10. I would expect that changing the service CIDR range would also take care of that.
Looking for a workaround, I tried using extra-config also for the service-cidr range, i.e.,

    --extra-config=kubelet.pod-cidr=172.16.0.0/14 --extra-config=kubelet.service-cidr=172.20.0.0/14

Now the minikube start would fail cause kubelet wouldn't start: 

    [...]
    Oct 25 11:40:15 minikube kubelet[24917]: F1025 11:40:15.308498   24917 server.go:156] unknown flag: --service-cidr
    Oct 25 11:40:15 minikube systemd[1]: kubelet.service: Main process exited, code=exited, status=255/n/a
    [...]
.

medyagh:
@jgoeres  thanks for creating this issue ! this does seem to be a bug. I would love minikube to be able to create service in CIDR ranges that works well for everyone including corp networks !

I would be happy to review any PR to fix this issue !

evankanderson:
I was able to repro this with in January with v1.6.2:

```shell
minikube start --keep-context --kubernetes-version=1.16.6 --memory=2536mb --service-cluster-ip-range=10.64.0.0/16 --vm-driver=vmware -p newerkube
```

```shell
kubectl run my-shell --rm -i --tty --image cosmintitei/bash-curl -- bash
```

```shell
bash-4.4# cat /etc/resolv.conf 
nameserver 10.96.0.10
search default.svc.cluster.local svc.cluster.local cluster.local
options ndots:5
```

dmvk:
Same isssue here on macOS with the latest version.

PR #6472, opened 2020-02-03
Use cluster-dns IP setup by kubeadm

vijaykatam:
Removed hardcoded `cluster-dns` IP 10.96.0.10, this allows kubelet to use `clusterDNS` configured by kubeadm in `/var/lib/kubelet/config.yaml`

Tested by starting minikube with `--service-cluster-ip-range=100.64.0.0/13` and verifying DNS resolution works.

/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
```
[Unit]
Wants=docker.socket

[Service]
ExecStart=
ExecStart=/var/lib/minikube/binaries/v1.17.0/kubelet --authorization-mode=Webhook --bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf --cgroup-driver='cgroupfs' --client-ca-file=/v
ar/lib/minikube/certs/ca.crt --cluster-domain=cluster.local --config=/var/lib/kubelet/config.yaml --container-runtime=docker --fail-swap-on=false --hostname-override=minikube --kubeconfig=/e
tc/kubernetes/kubelet.conf --node-ip=192.168.64.7 --pod-manifest-path=/etc/kubernetes/manifests

[Install]
```

DNS resolution test
```
kubectl run -i --tty --rm busybox --image=busybox --restart=Never -- sh
If you don't see a command prompt, try pressing enter.
/ # nslookup kubernetes.io
Server:		100.64.0.10
Address:	100.64.0.10:53

Non-authoritative answer:
Name:	kubernetes.io
Address: 147.75.40.148
```

Fixes: https://github.com/kubernetes/minikube/issues/5727

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -208,7 +208,7 @@ func enableOrDisableAddonInternal(addon *assets.Addon, cmd command.Runner, data 
 	if err != nil {
 		return err
 	}
-	glog.Infof("Running: %s", command)
+	glog.Infof("Running: %v", command)
 	rr, err := cmd.RunCmd(command)
 	if err != nil {
 		return errors.Wrapf(err, "addon apply")
--- a/pkg/minikube/bootstrapper/bsutil/kubelet_test.go
+++ b/pkg/minikube/bootstrapper/bsutil/kubelet_test.go
@@ -54,7 +54,7 @@ Wants=docker.socket
 
 [Service]
 ExecStart=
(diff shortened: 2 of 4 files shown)


medyagh:
@vijaykatam interesting PR ! I am curious if this will break anyone ! 
could we use the current behaviour if the user didn't pass a an extra settings?

btw the unit tests datas need to be updated for the PR.

vijaykatam:
> could we use the current behaviour if the user didn't pass a an extra settings?

@medyagh if user didn't pass `--service-cluster-ip-range` the default behavior of kubeadm is to use `10.96.0.10`, see https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/componentconfigs/kubelet.go#L108-L112. Since `/var/lib/kubelet/config.yaml` is already being passed in the kubelet drop-in, I guess always use `clusterDNS` specified by kubeadm?

vijaykatam:
I updated unit test setup data and they pass now. Integration test failures seem unrelated to this change.

medyagh:
Thanks @vijaykatam I will investigate

vijaykatam:
@medyagh I didn't see lint errors running `make lint`. Travis build failure is not due to lint error
```
$ travis_setup_go
I don't have any idea what to do with '1.13.4'.
  (using download type 'auto')
Failed to run gimme
The command "travis_setup_go" failed and exited with 86 during .
```
I rebased with latest from master and I see 2 lint errors unrelated to my change which I am fixing.
```
pkg/minikube/cluster/start.go:189: File is not `goimports`-ed (goimports)
  if driver.BareMetal(h.Driver.DriverName()) {
pkg/addons/addons.go:211:2: printf: Infof format %s has arg command of wrong type *os/exec.Cmd (govet)
	glog.Infof("Running: %s", command)
	^
```

tstromberg:
Thank you!

Issue #6485, opened 2020-02-04
make buildroot provision package re-usable for kic


medyagh:
while debugging kic driver that why crictl can not list images in docker I found out the /etc/crictl.yaml is not pointing to the docker.sock and docker sytemd config looks different !
on VM it is
```
   CGroup: /system.slice/docker.service
           ├─1908 /usr/bin/dockerd -H tcp://0.0.0.0:2376 -H unix:///var/run/docker.sock --default-ulimit=nofile=1048576:1048576 --tlsverify --tlscacert /etc/docker/ca.pem --tlscert /etc/docker/server.pem --tlskey /etc/docker/server-key.pem --label provider=hyperkit --insecure-registry 10.96.0.0/12
```

but it in container its naked and not configured.

the provision package configures our dockerd
https://github.com/kubernetes/minikube/blob/master/pkg/provision/buildroot.go

has useful functions that could be used for both VM and container
for example: https://github.com/kubernetes/minikube/blob/master/pkg/provision/buildroot.go#L90
this function generates systemd files for docker. which inncludes insecure regteries...
which is used by VM but we should make this package agonistic of VM or Container.

Issue #6463, opened 2020-02-02
kic with docker-env: endpoint read: connection reset by peer


balopat:
**The exact command to reproduce the issue**:

```bash
$ minikube start --vm-driver=docker --container-runtime=docker 
# ....
$ minikube docker-env                                                                                                                                                
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://127.0.0.1:32778"
export DOCKER_CERT_PATH="/Users/balintp/.minikube/certs"
# Run this command to configure your shell:
# eval $(minikube docker-env)
```
Now, if I eval...

```bash
$ eval $(minikube docker-env)
echo $DOCKER_HOST
tcp://127.0.0.1:32778
# great! 
$ docker images                                                                                                                                                               
error during connect: Get https://127.0.0.1:32778/v1.40/images/json: EOF
# not that great :( 
```

**The operating system version**:
MacOSX Mojave

afbjorklund:
I think the container runtime option (for docker driver) is quite new, it originally used `containerd`.

tstromberg:
I was able to replicate this on Linux.

balopat:
> I think the container runtime option (for docker driver) is quite new, it originally used `containerd`.
> 
> As a workaround you can avoid the docker-env and talk directly to same docker as minikube uses ?

The point would be to reuse the same daemon the containers are loaded from in the cluster. e.g. skaffold relies on this to avoid pushing.

medyagh:
@balopat I can re-produce this and I have a fix for it. the problem is dockerd is not lsitening on 0.0.0.0 insice the conrtianer

medyagh:
more details https://github.com/kubernetes/minikube/issues/6485

Issue #6445, opened 2020-01-31
kic test failure: verify_cache_inside_node : different output of crictil and docker images


medyagh:
thanks to this integration test I found out the cache add command doesnt work as expected on "docker driver" with docker run time.

```
=== RUN   TestFunctional/serial/CacheCmd/cache/verify_cache_inside_node
--- FAIL: TestFunctional/serial/CacheCmd/cache/verify_cache_inside_node (0.27s)
functional_test.go:365: (dbg) Run:  out/minikube-linux-amd64 -p functional-20200130T194619.520931734-14593 ssh sudo crictl images
functional_test.go:370: expected '1.28.4-glibc' to be in the output: -- stdout --
IMAGE                                TAG                 IMAGE ID            SIZE
k8s.gcr.io/coredns                   1.6.2               bf261d1579144       44.2MB
k8s.gcr.io/etcd                      3.3.15-0            b2756210eeabf       248MB
k8s.gcr.io/kube-apiserver            v1.16.2             9c26fe80bbb38       185MB
k8s.gcr.io/kube-controller-manager   v1.16.2             2f6b22cf92d72       128MB
k8s.gcr.io/kube-proxy                v1.16.2             41d59a0b79325       103MB
k8s.gcr.io/kube-scheduler            v1.16.2             fd6c6fdef2251       105MB
k8s.gcr.io/pause                     3.1                 da86e6ba6ca19       746kB
-- /stdout --
```

I did manual checking:
```
./out/minikube cache add busybox
```
in the integration tests we check the output of crictl images inside minikube and expect busybox to be there.

```
$ ./out/minikube ssh sudo crictl images
IMAGE                                TAG                 IMAGE ID            SIZE
k8s.gcr.io/coredns                   1.6.2               bf261d1579144       44.2MB
k8s.gcr.io/etcd                      3.3.15-0            b2756210eeabf       248MB
k8s.gcr.io/kube-apiserver            v1.16.2             9c26fe80bbb38       185MB
k8s.gcr.io/kube-controller-manager   v1.16.2             2f6b22cf92d72       128MB
k8s.gcr.io/kube-proxy                v1.16.2             41d59a0b79325       103MB
k8s.gcr.io/kube-scheduler            v1.16.2             fd6c6fdef2251       105MB
k8s.gcr.io/pause                     3.1                 da86e6ba6ca19       746kB
```

the intertesting is the output of  "docker images " has what we need !
```
$ ./out/minikube ssh sudo docker images
[...]

PR #6487, opened 2020-02-04
fix docker-env for kic drivers

medyagh:
- refactor provision package to be reusable for non-buildroot 
( created a new provisioner called ubuntuProvisioner and broke down the package two three files and created an interface)
- configure docker on vm and container same way
- fixes crictl setting that caused cache integration test failure https://github.com/kubernetes/minikube/issues/6445
- remove dependency on minikube-automount.service only for kic driver (till we implement mount for kic)
```
After=network.target  minikube-automount.service docker.socket
Requires= minikube-automount.service docker.socket 

```  

closes https://github.com/kubernetes/minikube/issues/6485
closes https://github.com/kubernetes/minikube/issues/6463
closes https://github.com/kubernetes/minikube/issues/6445


afbjorklund:
@medyagh: Weird that the standard Ubuntu provisioner was not working out properly ?

https://github.com/docker/machine/blob/master/libmachine/provision/ubuntu_systemd.go

This seems more like a hack, or suggests that `CompatibleWithHost` is broken:

```go
		var pv lib_provision.Provisioner
		if driver.IsKIC(h.Driver.DriverName()) {
			pv = provision.NewUbuntuProvisioner(h.Driver)
		} else {
			pv = provision.NewBuildrootProvisioner(h.Driver)
		}
```

Because libmachine is _supposed_ to be able to detect that, during provisioning:

https://github.com/docker/machine/blob/master/libmachine/provision/provisioner.go#L121_L129

PR #6492, opened 2020-02-04
improve "docker-env" command to point to correct docker

medyagh:
- adds a new minikube env var, to track if user is using minikube's docker daemon in the their shell or if they are using user's host docker
- use constants instead of hard code
- provide -p in the hint command , so none default profiles use the right command.
- reworded slightly for more politeness and added a new line for better readabilty

Before
```

```

After:
```
$ ./out/minikube -p k2 docker-env
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://127.0.0.1:32769"
export DOCKER_CERT_PATH="/Users/medya/.minikube/certs"
export MINIKUBE_ACTIVE_DOCKERD="k2"

# Please run command bellow to point your shell to minikube's docker-daemon :
# eval $(minikube -p k2 docker-env)
```


cmd/minikube/cmd/env.go
```diff
@@ -144,7 +128,32 @@ type NoProxyGetter interface {
 // EnvNoProxyGetter gets the no_proxy variable, using environment
 type EnvNoProxyGetter struct{}
 
-func generateUsageHint(userShell string) string {
+func generateUsageHint(profile string, userShell string) string {
+	const usgPlz = "Please run command bellow to point your shell to minikube's docker-daemon :"
+	var usgCmd = fmt.Sprintf("minikube -p %s docker-env", profile)
```
tstromberg: If they are using the default profile, this should omit the `-p profile`.  Similar to what we do in `start.go`.

pkg/drivers/kic/oci/oci.go
```diff
@@ -323,3 +358,22 @@ func listContainersByLabel(ociBinary string, label string) ([]string, error) {
...
+// to make sure it points to the docker daemon installed by user.
+func PointToHostDockerDaemon() error {
+	p := os.Getenv(constants.MinikubeActiveDockerdEnv)
+	if p != "" {
+		glog.Infof("shell is pointing to docker inside minikube. will unset to use host")
+	}
+
+	for i := range constants.DockerDaemonEnvs {
+		e := constants.DockerDaemonEnvs[i]
+		err := os.Setenv(e, "")
```
tstromberg: Does `minikube start` work with this PR?

Instead of manipulating a global variable, I think this would be cleaner if it returned a sanitized environment map.

The new environment can then be passed into `cmd.Env`

One caveat  I see however is that this may not be directly be applicable to go-containerregistry calls in start.go. Is this a workable idea?

pkg/minikube/constants/constants.go
```diff
@@ -39,8 +39,21 @@ const (
...
+	// DockerTLSVerifyEnv is used for docker daemon settings
+	DockerTLSVerifyEnv = "DOCKER_TLS_VERIFY"
+	// DockerHostEnv is used for docker daemon settings
+	DockerHostEnv = "DOCKER_HOST"
+	// DockerCertPathEnv is used for docker daemon settings
+	DockerCertPathEnv = "DOCKER_CERT_PATH"
+	// MinikubeActiveDockerdEnv holds the docker daemon which user's shell is pointing at
+	// value would be profile or empty if pointing to the user's host daemon.
+	// DockerDaemonEnvs has list of environment variables to control docker daemon shell is using
+
```
tstromberg: extra space?

pkg/minikube/constants/constants.go
```diff
@@ -39,8 +39,21 @@ const (
...
+	MinikubeActiveDockerdEnv = "MINIKUBE_ACTIVE_DOCKERD"
 )
 
+var DockerDaemonEnvs = [3]string{DockerHostEnv, DockerTLSVerifyEnv, DockerCertPathEnv}
```
tstromberg: any reason to make this an array rather than a slice?

medyagh: Was gonna emulate a const. Emphasizing the length be fixed.

pkg/drivers/kic/oci/oci.go
medyagh: Yeah the start and status and stop and delete works.

cmd/minikube/cmd/env.go
medyagh: I see some values that to emphasize specifying profile explicitly. Docker-env command could be confusing enough the more explicit we be the better.

Plus making defualt profile except for the hint text would require a refactoring for the unit tests that would be bigger than this pr.

I am open to do that in a follow up pr

pkg/drivers/kic/oci/oci.go
medyagh: I totally agree that it is much much better to pass it as cmd.Env and i will send a follow-up pr to improve it after release

medyagh: On a second thought though. If we don't set it globally how can we make go container register use host docker?

I think in no case minikube needs to talk to the docker-env inside minikube ( for that we run our command inside the node)

PR #6440, opened 2020-01-30
Add addon enablement to start

tstromberg:
This PR explicitly enables addons on startup, in lieu of there being an addon-manager to do so.

There is a new line emitted during startup:

```
🚀  Launching Kubernetes ... 
🌟  Enabling addons: storage-provisioner, default-storageclass
⌛  Waiting for cluster to come online ...
```

## Behavior changes

There is also a behavioral change: if an addon is enabled/disabled, and you attempt to enable it again, it will run the underlying kubectl command. It should still be idempotent. 

I expect this will add ~1 second to our start time, but it does restore the previous behavior.

## API changes

The addons package has a new function to handle startup logic: `addons.Start()`. This obsoletes what was in `start.go:enableAddons()` and `files.go:AddAddons()`

To improve testing, the following public functions now take a profile name:

* `addon.IsEnabled()`
* `cluster.IsHostRunning()`

Fixes #6363

PR #6494, opened 2020-02-04
Fix local assets sync broken by addons refactor

tstromberg:
We accidentally removed local file synchronization in this change:

https://github.com/kubernetes/minikube/pull/6440/commits/d5a83c7b244add0eb178edf7ce1628f5da7ec20d#diff-28d4bbd7728c5ae9a24a8f7e56083a1d

Three main changes here:

* File synchronization functionality has been moved from `addons` to `cluster` package
* Functions in the `util` package were removed and/or isolated to a smaller scope
* Time synchronization is now only run for existing VM's.


medyagh:
KVM 26/48 failures ??

PR #6526, opened 2020-02-06
Makefile: push docker images to latest instead of specific revisions

tstromberg:
This will allow us to decouple the image tagging and certification process from the build process. 

Re-using old specific versions like v1.8.1 turns out to not be good. 

The suggested new procedure is:

- An admin will tag new image releases to a specific version
- When they are ready to test, we will update the code within minikube to specify that tag


lakshmitadi:
When is this scheduled to be released?

PR #6531, opened 2020-02-06
Provide images with pre-loaded Kubernetes images & binaries

priyawadhwa:
Adds support for specifying kubernetes version and preloading docker
images into the kic base image.

Though this increases the kic base image by ~1.6G, it should improve
start latency in conjunction with removing 'kubeadm pull images' in
another PR.

The kic base image name now includes the k8s version in the tag eg `gcr.io/k8s-minikube/kicbase-v0.0.5-k8s-v1.17.2`


hack/preload-images/generate-preloaded-images-tar.sh
```diff
@@ -0,0 +1,32 @@
...
+KUBERNETES_VERSION=${KUBERNETES_VERSION:-""}
+TARBALL_FILENAME=preloaded-images-k8s-$KUBERNETES_VERSION.tar
+
+function delete_minikube {
+    out/minikube delete --profile=$PROFILE
+}
+
+trap "delete_minikube" ERR
+
+out/minikube start --memory=10000 --profile=$PROFILE --kubernetes-version=$KUBERNETES_VERSION
```
medyagh: is that 10gb ram ?

hack/preload-images/generate-preloaded-images-tar.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
medyagh: how hard would to convert this to a go script ?

Makefile
```diff
@@ -500,10 +502,13 @@ storage-provisioner-image: out/storage-provisioner-$(GOARCH) ## Build storage-pr
 	docker build -t $(STORAGE_PROVISIONER_IMAGE) -f deploy/storage-provisioner/Dockerfile  --build-arg arch=$(GOARCH) .
 
 .PHONY: kic-base-image
-kic-base-image: ## builds the base image used for kic.
-	docker rmi -f $(REGISTRY)/kicbase:v0.0.5-snapshot || true
-	docker build -f ./hack/images/kicbase.Dockerfile -t $(REGISTRY)/kicbase:v0.0.5-snapshot  --build-arg COMMIT_SHA=${VERSION}-$(COMMIT)  .
+kic-base-image: generate-preloaded-images-tar ## builds the base image used for kic.
```
medyagh: lets have a make target just for building the base and one just for building pre-loaded version, that way people who do not want a preloaded verison have an option to provide (to --iso-url or --kic_img)

tstromberg: Agreed, but mostly so that if someone specifies a version of Kubernetes that isn't supported, they can instead download the base image that shares most of the same layers, rather than downloading an image with the wrong version of Kubernetes baked in.

hack/images/kicbase.Dockerfile
```diff
@@ -34,12 +34,21 @@ USER root
...
+  RUN echo "kic! Build: ${COMMIT_SHA} Time :$(date)" > "/kic.txt"
+
+
+FROM busybox
+ARG KUBERNETES_VERSION
+COPY out/preloaded-images-k8s-$KUBERNETES_VERSION.tar /preloaded-images.tar
+RUN tar xvf /preloaded-images.tar -C /
+
+FROM base
+COPY --from=1 /var/lib/docker /var/lib/docker
```
tstromberg: This is a clever way to avoid coding specific versions of images to docker pull. Does `docker ps` actually show these images?

Also, we should copy `/var/lib/minikube/binaries`. Feel free to save that for a follow-up PR if you prefer.

medyagh: What happens to containerd images ?

medyagh: I wonder if there is a way we give the images to crictl and let both docker and containerd use it?

medyagh: maybe using crctl we could make it available for all runtimes ?

hack/preload-images/generate-preloaded-images-tar.sh
priyawadhwa: yup, it's needed since 1.6 G of files needs to be tarred up within the VM

Makefile
priyawadhwa: Sure, I'll add the preloaded build as a separate Makefile target.

Right now I append the preloaded images as a layer on top of the regular base image, so they should share all layers except an additional top one on the preloaded images. I'm envisioning this flow in the code:

1. user specifies kubernetes version 
1. if kic_base_image:v0.0.5-k8s-$(k8s version) exists remotely, use that as the base image
1. otherwise, use the regular base image

hack/images/kicbase.Dockerfile
priyawadhwa: I was going to handle crictl and containerd in a separate PR, they may or may not have the same storage structure for images.

Does `/var/lib/minikube/binaries` exist in the current kic base image? If so then it should be there in the preloaded image as well.

hack/preload-images/generate-preloaded-images-tar.sh
priyawadhwa: done

hack/images/kicbase.Dockerfile
priyawadhwa: Currently, `docker ps` shows:

`gcr.io/k8s-minikube/kicbase:v0.0.5-k8s-v1.17.2` if using the preloaded image or

`gcr.io/k8s-minikube/kicbase:v0.0.5` (current default) if using a k8s version that doesn't have a preloaded image

medyagh: thanks for converting it to a go script, 

heads up "/var/lib/minikube/binaries" is going to be changed to "/usr/local/minikube/..."
it is better to use the vmpath.GuestPersistentDir

or also if we run minikube start it should create all those stuff for us in it so we dont need to depend on that constant

hack/preload-images/preload_images.go
```diff
@@ -0,0 +1,116 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
```
medyagh: 2019

hack/images/kicbase.Dockerfile
medyagh: I am okay with shipping pre-load first for our docker run time then think about others later. as most ppl just go with the default one

Makefile
```diff
@@ -19,6 +19,8 @@ VERSION_BUILD ?= 1
 RAW_VERSION=$(VERSION_MAJOR).$(VERSION_MINOR).${VERSION_BUILD}
 VERSION ?= v$(RAW_VERSION)
 
+KUBERNETES_VERSION ?= $(shell cat pkg/minikube/constants/constants.go | grep "DefaultKubernetesVersion =" | cut -d ' ' -f4 | sed 's/"//g')
```
tstromberg: `egrep "^var DefaultKubernetesVersion" pkg/minikube/constants/constants.go | cut -d\" -f2 is more succinct and precise.

Makefile
```diff
@@ -502,8 +504,17 @@ storage-provisioner-image: out/storage-provisioner-$(GOARCH) ## Build storage-pr
 .PHONY: kic-base-image
 kic-base-image: ## builds the base image used for kic.
 	docker rmi -f $(REGISTRY)/kicbase:v0.0.5-snapshot || true
-	docker build -f ./hack/images/kicbase.Dockerfile -t $(REGISTRY)/kicbase:v0.0.5-snapshot  --build-arg COMMIT_SHA=${VERSION}-$(COMMIT)  .
```
tstromberg: v.0.0.5 seems to be repeated a lot: please make it a variable.

hack/preload-images/preload_images.go
```diff
@@ -0,0 +1,120 @@
...
+	minikubePath = "out/minikube"
+)
+
+var (
+	kubernetesVersion = ""
+	tarballFilename   = ""
+)
+
+func init() {
+	if kv := os.Getenv("KUBERNETES_VERSION"); kv != "" {
```
tstromberg: I suggest using os.Args or a flag over an environment variable if possible, as env variables are not very discoverable.

hack/preload-images/preload_images.go
```diff
@@ -0,0 +1,120 @@
...
+		return err
+	}
+	if err := createImageTarball(); err != nil {
+		return err
+	}
+	return copyTarballToHost()
+}
+
+func startMinikube() error {
+	cmd := exec.Command(minikubePath, "start", "-p", profile, "--memory", "10000", "--kubernetes-version", kubernetesVersion)
```
tstromberg: Can the 10GB memory requirement be removed?

hack/preload-images/preload_images.go
```diff
@@ -0,0 +1,120 @@
...
+		return errors.Wrap(err, "getting ssh-key")
+	}
+
+	ip, err := runCmdCaptureStdout([]string{minikubePath, "ip", "-p", profile})
+	if err != nil {
+		return errors.Wrap(err, "getting ip")
+	}
+
+	dest := filepath.Join("out/", tarballFilename)
+	args := fmt.Sprintf("scp -o StrictHostKeyChecking=no -i %s docker@%s:/home/docker/%s %s", sshKey, ip, tarballFilename, dest)
```
tstromberg: I'm glad you figured this hack out. It's good evidence that we should just implement a scp command.

tstromberg: For clarity, consider making args []string{} up front.

tstromberg: add /var/lib/minikube/binaries?

hack/preload-images/preload_images.go
```diff
@@ -0,0 +1,120 @@
...
+	return err
+}
+
+func deleteMinikube() error {
+	cmd := exec.Command(minikubePath, "delete", "-p", profile)
+	cmd.Stdout = os.Stdout
+	return cmd.Run()
+}
+
+func runCmdCaptureStdout(command []string) (string, error) {
```
tstromberg: In newer Go versions, you don't need to handle processing output any more. Just call cmd.Output() instead of cmd.Run().

Makefile
priyawadhwa: done

hack/preload-images/preload_images.go
priyawadhwa: Lowered it to 4GB -- at the typical 2GB we run out of space on the device because we're trying to tar up 1.6GB of files.

hack/preload-images/preload_images.go
priyawadhwa: added it to the tarball!

priyawadhwa:
@tstromberg thanks for your suggestions, PTAL when you have the chance.

Issue #6524, opened 2020-02-06
docker-env command does not work in powershell (v1.7)


suecideTech:
**Environment**
- minikube v1.7.1 on Microsoft Windows 10 Pro 10.0.18362 Build 18362

**Commnd Execute result**
```
PS C:\Windows\system32> minikube docker-env --shell powershell | Invoke-Expression
Invoke-Expression : 引数が空の文字列であるため、パラメーター 'Command' にバインドできません。
発生場所 行:1 文字:42
+ minikube docker-env --shell powershell | Invoke-Expression
+                                          ~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidData: (:PSObject) [Invoke-Expression]、ParameterBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationErrorEmptyStringNotAllowed,Microsoft.PowerShell.Commands.InvokeExpressionCommand
```

*Remarks*
Line break position is different between v1.7.2 and v1.6.2.
There is a line break after $Env.
<details>
  <summary>v1.6.2</summary>

```
PS C:\Windows\system32> docker-env --shell powershell
$Env:DOCKER_TLS_VERIFY = "1"
$Env:DOCKER_HOST = "tcp://192.168.163.55:2376"
$Env:DOCKER_CERT_PATH = "C:\Users\***\.minikube\certs"
# Run this command to configure your shell:
# & minikube docker-env | Invoke-Expression

PS C:\Windows\system32>
```
</details>

<details>
  <summary>v1.7.2</summary>

```
PS C:\Windows\system32> minikube docker-env --shell powershell
$Env:DOCKER_TLS_VERIFY = "1"
$Env:DOCKER_HOST = "tcp://192.168.163.55:2376"
$Env:DOCKER_CERT_PATH = "C:\Users\***\.minikube\certs"
$Env:MINIKUBE_ACTIVE_DOCKERD = "minikube"

# Please run command bellow to point your shell to minikube's docker-daemon :
# & minikube -p minikube docker-env | Invoke-Expression

PS C:\Windows\system32>
```
</details>

tstromberg:
Fixed in v1.7.2. Thank you for your patience!

PR #6540, opened 2020-02-07
Fix a variety of bugs in `docker-env` output

tstromberg:
# Bugfixes:

- Remove errant newline in powershell and cmd output #6524 
- Remove unnecessary semicolon in fish output
- Fix corrupt unset output if --unset and --no-proxy are used together
- Remove trailing tab from cmd, powershell, and emacs output
- Remove incorrect comment from unset output

# Other changes:

- Now unsets with a single command in compatible environments
- Minor rewording of comment


medyagh:
@tstromberg thank you for fixing up after my human errors

PR #6541, opened 2020-02-07
Fix to delete context when delete minikube

### LINKED ISSUE #6538: Need to delete minikube-cluster,user,context config from kubeconfig file.
1) `cat $HOME/.kube/config` and check the stdout.
(Maybe the path will be different, it depends on your KUBECONFIG enviroment variable. I guess you already know.) 

2) `minikube start` (with or without -p option to specify profile)

3) `cat $HOME/.kube/config` and recheck the stdout again. And compare with output of 1).
    - you can check the information about your new minikube instance's cluster, user, context added.

4) `minikube delete`

5) `cat $HOME/.kube/config` and recheck the stdout. And compare with output of 1), 3)
    - Added information in 3) remains.
    - I think the output of 5) needs to be same as 1). Not 3)
    - I think if `minikube start` set kubeconfig, `minikube delete` should unset kubeconfig.

- I think there is this logic in the source code (`func deleteContext` in `cmd/minikube/cmd/delete.go`), but it won't actually works.

NONE

</details>

NONE

</details>

Ubuntu 18.04.4 LTS
minikube Version : tested with current master branch (commit: 9b7573ee2064a24615b7d84ede66c047d36ab953) binary

anencore94:
- When `minikube delete`, minikube must delete its contexts which are set when `minikube start`.
- But, there was some mistakes in the function that does this logic.
- `DeleteContext(machineName string, configPath ...string)` in `pkg/minikube/kubeconfig/context.go` was not matching with usage in `deleteContext`.
    - it calls the function in the wrong order of parameter :
    - `kubeconfig.DeleteContext(constants.KubeconfigPath, machineName)`

Fixes #6538

### DIFF
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -238,7 +238,7 @@ func deleteProfile(profile *pkg_config.Profile) error {
 }
 
 func deleteContext(machineName string) error {
-	if err := kubeconfig.DeleteContext(constants.KubeconfigPath, machineName); err != nil {
+	if err := kubeconfig.DeleteContext(machineName, constants.KubeconfigPath); err != nil {
 		return DeletionError{Err: fmt.Errorf("update config: %v", err), Errtype: Fatal}
 	}
 
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -238,7 +238,7 @@ func deleteProfile(profile *pkg_config.Profile) error {
 }
 
 func deleteContext(machineName string) error {
-	if err := kubeconfig.DeleteContext(machineName, constants.KubeconfigPath); err != nil {
+	if err := kubeconfig.DeleteContext(machineName); err != nil {
 		return DeletionError{Err: fmt.Errorf("update config: %v", err), Errtype: Fatal}
 	}
 
--- a/test/integration/start_stop_delete_test.go
+++ b/test/integration/start_stop_delete_test.go
@@ -153,6 +153,14 @@ func TestStartStop(t *testing.T) {
 					if err != nil {
 						t.Errorf("%s failed: %v", rr.Args, err)
 					}
+
+					rr, err = Run(t, exec.CommandContext(ctx, "kubectl", "config", "get-contexts"))
+					if err != nil {
+						t.Fatalf("%s failed: %v", rr.Args, err)
+					}
+					if notDeleted := strings.Contains(rr.Output(), profile); notDeleted {
+						t.Errorf(" kubeconfig of %s is not deleted = %t; want = %t", profile, notDeleted, false)
+					}
 				}
 			})
 		}


medyagh:
the code doesnt build ! @anencore94

anencore94:
@medyagh Thanks for checking !!
The error message of travis ci build failed is as follows :
```
$ make
which go-bindata || GO111MODULE=off GOBIN="/home/travis/gopath/bin" go get github.com/jteeuwen/go-bindata/...
# cd .; git clone -- https://github.com/jteeuwen/go-bindata /home/travis/gopath/src/github.com/jteeuwen/go-bindata
Cloning into '/home/travis/gopath/src/github.com/jteeuwen/go-bindata'...
fatal: unable to access 'https://github.com/jteeuwen/go-bindata/': Failed to connect to github.com port 443: Connection timed out
```
I think it doesn't build since some connection error occured ?

medyagh:
I restarted the build ! this does look like a LEGIT bug ! thank you for findind this ! @anencore94

medyagh:
@anencore94 how do you feel about adding an integeration test for this here https://github.com/kubernetes/minikube/blob/master/test/integration/start_stop_delete_test.go#L153
so it never happens again ?
+10 points to you if you sign up for this challenge !

anencore94:
@medyagh
I'd love too :) I'll add commit on this pr later.

anencore94:
@medyagh 
I've been force-updated my commit. Just added simple if-else assertions.
Please re-check this pr and review on my test code!! Thanks :)

Also, the travis-ci failed on lint by this error:
```
./test.sh
= make lint =============================================================
golangci/golangci-lint info checking GitHub for tag 'v1.23.2'
golangci/golangci-lint crit unable to find 'v1.23.2' - use 'latest' or see https://github.com/golangci/golangci-lint/releases for details
Makefile:350: recipe for target 'out/linters/golangci-lint-v1.23.2' failed
make[1]: *** [out/linters/golangci-lint-v1.23.2] Error 1
```
I guess something happened in the travisci-running-node..?

anencore94:
By the way, before this commit merged, the master branch source code always goes into this way :
`cmd/minikube/cmd/delete.go` 240th line => 241th line 
=> `pkg/minikube/kubeconfig/context.go` 66th line => 76th line (since api.IsConfigEmpty(kcfg) is true) => 78th line => finally, return nil (not error)

Is it ok to just changing the wrong-ordered-parameter? Or does the condition should be changed somehow ?

test/integration/start_stop_delete_test.go
```diff
@@ -153,6 +153,14 @@ func TestStartStop(t *testing.T) {
 					if err != nil {
 						t.Errorf("%s failed: %v", rr.Args, err)
 					}
+
+					rr, err = Run(t, exec.CommandContext(ctx, "kubectl", "config", "get-contexts"))
+					if err != nil {
+						t.Fatalf("%s failed: %v", rr.Args, err)
+					}
+					if notDeleted := strings.Contains(rr.Output(), profile); notDeleted {
+						t.Errorf(" kubeconfig of %s is not deleted = %t; want = %t", profile, notDeleted, false)
```
tstromberg: Remove the leading space, and add `rr.Output()` to the error for debugging.

cmd/minikube/cmd/delete.go
```diff
@@ -238,7 +238,7 @@ func deleteProfile(profile *pkg_config.Profile) error {
 }
 
 func deleteContext(machineName string) error {
-	if err := kubeconfig.DeleteContext(constants.KubeconfigPath, machineName); err != nil {
+	if err := kubeconfig.DeleteContext(machineName, constants.KubeconfigPath); err != nil {
```
medyagh: this is using the "Default" kubeconfig file, we should be using whatever user has passed
please use
kubeconfig.PathFromEnv() to get the Kubeconfig

medyagh: I also noticed we have many other places in the code that we are using 
"constants.KubeconfigPath"
for example
	err := kubeconfig.SetCurrentContext(profile, constants.KubeconfigPath)

we should change all of them

test/integration/start_stop_delete_test.go
```diff
@@ -153,6 +153,14 @@ func TestStartStop(t *testing.T) {
 					if err != nil {
 						t.Errorf("%s failed: %v", rr.Args, err)
 					}
+
+					rr, err = Run(t, exec.CommandContext(ctx, "kubectl", "config", "get-contexts"))
```
tstromberg: Instead of pattern matching, can I suggest using:

`kubectl config get-contexts <profile>`

Then you can check for an error code. For example:

```
if rr.ExitCode != 1 {
  t.Errorf("wanted exit code 1, got %d. output: %s", rr.ExitCode, rr.Output())
}
```

tstromberg:
Thank you so much for fixing this. It's been broken for so long that honestly I had imagined it as a feature.

cmd/minikube/cmd/delete.go
medyagh: This is not your fault, this was in minikube code base and this is a good opportunity to refactor all of them.

please refactor in a way that delete context does not get a PATH at all and we make sure everyone in the code base is using 
"kubeconfig.PathFromEnv()" instead of constants.KubeconfigPath

I suggest moving constants.KubeconfigPath to a Private field in "kubeconfig" package so others can not access it

anencore94: Thanks for checking that issue @medyagh 
If you don't mind, i could try to refactor to use user Path rather than "constants.KubeconfigPath" :)
Have a nice weekend !

test/integration/start_stop_delete_test.go
anencore94: I couldn't agree with you more.
Yeah, for debugging when the error occured, rr.Output() will have much more infomation.
Thank you for reviewing @tstromberg , I will create another pr to refactoring this test after this weekend.

Have a nice weekend :)

test/integration/start_stop_delete_test.go
anencore94: DONE

test/integration/start_stop_delete_test.go
anencore94: DONE in #6567

cmd/minikube/cmd/delete.go
anencore94: > please refactor in a way that delete context does not get a PATH at all and we make sure everyone in the code base is using
> "kubeconfig.PathFromEnv()" instead of constants.KubeconfigPath
> 
> I suggest moving constants.KubeconfigPath to a Private field in "kubeconfig" package so others can not access it

@medyagh 
If we move constants.KubeconfigPath to a Private field in "kubeconfig" package, then the "kubeconfig.PathFromEnv()" function also cannot access it.
Then we have to put this stirng constant in the kubeconfig package. But i think it should be placed in the constant package..
Is there any other way to prevent access from others except PathFromEnv() function ?
I have related WIP pr (#6568 )

PR #6548, opened 2020-02-07
Add instructions for using the minikube registry with Docker for Windows

CrossBound:
Instructions for MacOS do not work as-is for Windows. Need to add an additional step.


CrossBound:
I have signed the CLA

PR #6552, opened 2020-02-07
Update script to use bump-formula-pr

### LINKED ISSUE #5779: release_update_brew: access_token: unbound variable
minikube migrated from cask to brew, but the automation has not been updated to use `bump-formula-pr.rb` to trigger updates: https://github.com/Linuxbrew/brew/blob/master/docs/How-To-Open-a-Homebrew-Pull-Request.md

Here is where the automation should go:

https://github.com/kubernetes/minikube/blob/9be404689b6860f189147bb53050b6f09a15681a/hack/jenkins/release_update_brew.sh#L43

tstromberg:
If this works, it fixes #5779

### DIFF
--- a/hack/jenkins/release_update_brew.sh
+++ b/hack/jenkins/release_update_brew.sh
@@ -19,66 +19,39 @@
 #   VERSION_MAJOR
 #   VERSION_MINOR
 #   VERSION_BUILD
-#   BOT_PASSWORD
+#   access_token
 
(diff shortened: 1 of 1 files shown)


medyagh:
how could we verify if it works?

tstromberg:
> how could we verify if it works?

I ran the script locally, but this PR really depends on how things are plumbed in Jenkins.

PR #6556, opened 2020-02-08
Create the Node subcommands

### LINKED ISSUE #6300: Create Node operations for multinode
```
minikube node add
minikube node start
minikube node stop
minikube node delete
```

Implement those 4 operations and all the config nonsense that goes along with it.

Make sure to fix MachineConfig so we can refactor all the node specific stuff.

ref #94

sharifelgamal:
Fixes #6300 

Also refactors `minikube start` and `minikube stop`


cmd/minikube/cmd/logs.go
```diff
@@ -66,7 +67,7 @@ var logsCmd = &cobra.Command{
 		if err != nil {
 			exit.WithError("command runner", err)
 		}
-		bs, err := getClusterBootstrapper(api, viper.GetString(cmdcfg.Bootstrapper))
+		bs, err := node.Bootstrapper(api, viper.GetString(cmdcfg.Bootstrapper))
```
tstromberg: This placement feels a bit weird: the bootstrapper is a cluster-property rather than a node property.

cmd/minikube/cmd/stop.go
```diff
@@ -53,15 +53,15 @@ func runStop(cmd *cobra.Command, args []string) {
 
 	nonexistent := false
 	stop := func() (err error) {
-		err = cluster.StopHost(api)
+		err = cluster.StopHost(api, profile)
```
tstromberg: This should probably be node.Stop() or cluster.Stop()

pkg/minikube/node/start.go
```diff
@@ -0,0 +1,123 @@
...
+	"k8s.io/minikube/pkg/minikube/driver"
+	"k8s.io/minikube/pkg/minikube/exit"
+	"k8s.io/minikube/pkg/minikube/kubeconfig"
+	"k8s.io/minikube/pkg/minikube/localpath"
+	"k8s.io/minikube/pkg/minikube/out"
+	"k8s.io/minikube/pkg/util"
+)
+
+// Start spins up a guest and starts the kubernetes node.
+func Start(mc *config.MachineConfig, n *config.Node, primary bool, existingAddons map[string]bool) (*kubeconfig.Settings, error) {
```
tstromberg: Unless this is altering the contents of mc & n, consider passing in values rather than pointers.

"Don't pass pointers as function arguments just to save a few bytes. If a function refers to its argument x only as *x throughout, then the argument shouldn't be a pointer."

pkg/minikube/node/start.go
```diff
@@ -0,0 +1,123 @@
...
+	// exits here in case of --download-only option.
+	handleDownloadOnly(&cacheGroup, k8sVersion)
+	mRunner, preExists, machineAPI, host := startMachine(mc, n)
+	defer machineAPI.Close()
+	// configure the runtime (docker, containerd, crio)
+	cr := configureRuntimes(mRunner, driverName, mc.KubernetesConfig)
+	showVersionInfo(k8sVersion, cr)
+	waitCacheRequiredImages(&cacheGroup)
+
+	// Must be written before bootstrap, otherwise health checks may flake due to stale IP
```
tstromberg: Feel free to do this later, but kubeconfig (and other bits of this function) is on a cluster level rather than a node level.

medyagh:
the integration tests look good !!!!

medyagh:
only test that looks a bit suspectious is TestVersionUpgrade that fails across all of them

PR #6560, opened 2020-02-08
Add packaging of the falco_probe kernel module

afbjorklund:
See also https://falco.org/docs/installation/ on how to install the rest of it (i.e. falco)

Minikube doesn't feature a compiler or kernel headers, so provide a pre-built module.

```console
$ minikube kubectl -- logs -l app=falco-example
Your kernel headers for kernel 4.19.88 cannot be found at
/lib/modules/4.19.88/build or /lib/modules/4.19.88/source.
* Running dkms build failed, couldn't find /var/lib/dkms/falco/0.19.0/build/make.log
* Trying to load a system falco-probe, if present
falco-probe found and loaded with modprobe
```

Installs `/lib/modules/${KERNEL_VERSION}/extra/falco-probe.ko` on the ISO

----

Sysdig _used_ to provide prebuilt modules for minikube, but have stopped doing so after 0.18...

> To address this, starting with falco 0.13.1 we pre-build kernel modules for the last 10 Minikube versions and make them available at https://s3.amazonaws.com/download.draios.com. This allows the download fallback step to succeed with a loadable kernel module.

> Going forward, we’ll continue to support 10 most recent versions of Minikube with each new Falco release. We currently retain previously-built kernel modules for download, so we will continue to provide limited historical support as well.

https://s3.amazonaws.com/download.draios.com/stable/sysdig-probe-binaries/index.html

Eventually it might become available from https://github.com/falcosecurity/kernel-module


afbjorklund:
Related issue in the upstream project: https://github.com/falcosecurity/falco/issues/1009

afbjorklund:
@medyagh : Thanks for merging, we should bump it to 0.20 before minikube v1.8.0:

https://github.com/falcosecurity/falco/issues/1048

https://github.com/falcosecurity/falco/milestone/5

But hopefully that should just be a minor version increase, and nothing more than that.

afbjorklund:
It was bumped in fc3a5656451ced57cc8dd68b685dbf12604545a1 (#6777)

PR #6564, opened 2020-02-09
improve github actions to generate one line test status same style as jenkins

### LINKED ISSUE #6447: consitstant commit status for github actions and jenkins
We need to make github actions to spit out number of failures into the commit status and the duration.

plus currently it always sets the commit status to success. it should also show fail if test has failure.
the jenkins ones look like this : 

```
Docker_Linux — Jenkins: completed with 21 / 48 failures in 26.80 minute(s).
KVM_Linux — Jenkins: completed with 3 / 48 failures in 26.55 minute(s).
VirtualBox_Linux — Jenkins: completed with 2 / 48 failures in 23.01 minute(s).
```

but the github actions dont have same format.
here is the file on jenkins that creates that format : 
https://github.com/kubernetes/minikube/blob/master/hack/jenkins/common.sh#L331

here is the file that needs to be edited.
https://github.com/kubernetes/minikube/blob/master/.github/workflows/main.yml

there should be examples on the web how to set commit status in github actions

KaenDagger:
Fixes #6447 

Image for mac
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,8 +2,11 @@ name: CI
 
 on: [pull_request]
 
+
 jobs:
   docker_ubuntu_16_04:
+    env:
+      ELAPSED: time
     runs-on: ubuntu-16.04
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


alonyb:
@KaenDagger 
There exist a way to print those details in GitHub UI(image)? as Jenkins.

image

KaenDagger:
Based on this comment. It seems currently there is no way for setting status on UI.

https://github.community/t5/GitHub-Actions/Github-Actions-as-Status-checks/m-p/36182/highlight/true#M2467
Still I have asked them explicitly , if there is any way

medyagh:
> have asked them explicitly , if there is any way

Thanks for the exploring the possibililty.  it seems like there is a way if we put this .js script in the same folder with github workflow and also add block runs

please try this : https://github.com/qawolf/actions-test/tree/master/.github/actions/create-status-check

KaenDagger:
> please try this : https://github.com/qawolf/actions-test/tree/master/.github/actions/create-status-check

@medyagh I saw this. But this looked kinda hack to me. So didn't say about this earlier. Ok I will try to add and see the result

medyagh:
> > please try this : https://github.com/qawolf/actions-test/tree/master/.github/actions/create-status-check
> 
> @medyagh I saw this. But this looked kinda hack to me. So didn't say about this earlier. Ok I will try to add and see the result

thank you for trying, if it is not possible thats okay! this is good progress that now we have produce same wording.

do you think you can make the Test to Fail if the ingeration test is failed ?
or do you think u can make the artifacts to be downloaded in one file instead of 4 different files

medyagh:
@KaenDagger not sure if this is helpful but seems like this one is doing it https://github.com/proxer/ProxerAndroid/blob/edd5b0b9767f79f50d26d07be67f524ff3432c3e/.github/workflows/ci.yml#L105

KaenDagger:
> @KaenDagger not sure if this is helpful but seems like this one is doing it https://github.com/proxer/ProxerAndroid/blob/edd5b0b9767f79f50d26d07be67f524ff3432c3e/.github/workflows/ci.yml#L105

Lemme try this one

KaenDagger:
@medyagh  I think js trick might work 
But currently it depends on `@action/core` and on running it throws 
`Error: Cannot find module '@actions/core'`

core module is actually responsible for setting status AFAIU

medyagh:
That okay to add npm to gitbub action workflow of if it fixes it .

I am curious did u try to make this work ?

actions/github-script@0.3.0
https://github.com/proxer/ProxerAndroid/blob/edd5b0b9767f79f50d26d07be67f524ff3432c3e/.github/workflows/ci.yml#L105

medyagh:
@KaenDagger one more request, could we use the sentence you generated in the file name for the artifact?
 that gets uploaded ?
for example 
docker_ubuntu_18_04_Docker_Ubuntu_16_04_Completed_with_22_48_failures_in_19_min_24_second

KaenDagger:
Html file  could be created the same way as in Jenkins. As far as it uses bash it should be possible to create it

medyagh:
> Html file could be created the same way as in Jenkins. As far as it uses bash it should be possible to create it

yeah we already create that and upload it to the job articats in github actions

I meant just rename the file name for the test report to match the test one line status

KaenDagger:
Sure. I will try for UI status check

medyagh:
@KaenDagger  that would be great !
and this was good work thank you very mcuh

PR #6566, opened 2020-02-09
TestCacheBinariesForBootstrapper in a temp dir

afbjorklund:
I was wondering why I had "invalid version" in my cache dir.

The downloads should probably be mocked, would be faster too ?


medyagh:
That is a great cache thank you

PR #6568, opened 2020-02-10
Fix to kubeConfigPath could be set from env

anencore94:
Some codes used kubeConfigPath just as $HOME/.kube/config,
but it may be different if KUBECONFIG env was set.
Thus, modify those codes to use kubeConfigPath by
using kubeconfig.PathFromEnv function.

- applying review from #6541


medyagh:
I was hoping someone would make this PR ! thank you for doing it

/ok-to-test

medyagh:
thank you @anencore94

PR #6572, opened 2020-02-10
Enable CONFIG_CGROUP_BPF

brb:
This PR adds `CONFIG_CGROUP_BPF=y` to the minikube kernel config, which enables running eBPF cgroup programs.

An example of a producer of such programs is Cilium CNI.


priyawadhwa:
Thank you for adding this!

Issue #6583, opened 2020-02-11
minikube kubect: permission denied


bartoszj:
After upgrading to minikube v.1.7.2 I'm not longer able to use `minikube kubectl`.

**The exact command to reproduce the issue**:
```
rm -rf ~/.minikube
minikube start
minikube kubectl
```

**The full output of the command that failed**:<details>
```
Error running /Users/<USER>/.minikube/cache/v1.17.2/kubectl: fork/exec /Users/<USER>/.minikube/cache/v1.17.2/kubectl: permission denied
```

</details>

**Contents of the `~/.minikube/cache/v1.17.2/kubectl`**:<details>
```
ls -l ~/.minikube/cache/v1.17.2
-rw-------  1 <USER>  staff   39342080 11 lut 10:27 kubeadm
-rw-------  1 <USER>  staff   43491328 11 lut 10:27 kubectl
-rw-------  1 <USER>  staff  111568408 11 lut 10:27 kubelet
```

```
file ~/.minikube/cache/v1.17.2/kubectl
/Users/<USER>/.minikube/cache/v1.17.2/kubectl: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=itoADCJwxOm7z3ymMN2Z/yoeLs_MYMz8Q5D0z_bWX/K9sqErNfMu6XJcqYVzr4/nHZFbmivkI7eipSyzj5i, stripped
```

</details>

**The operating system version**: macOS Catalina 10.15.3 (19D76)

It looks like minikube downloaded executables for Linux operating system, where it should download executable to macOS. Currently the workaround is to delete files from `~/.minikube/cache/v1.17.2` folder and execute `minikube kubectl` which will download correct executables for macOS.

afbjorklund:
Same as #6559 - but in reverse

afbjorklund:
Fixed in v1.7.3

PR #6586, opened 2020-02-11
Separate cached kubernetes binaries by OS on host machine

### LINKED ISSUE #6559: addon enable fails after update: kubectl: cannot execute binary file
**The exact command to reproduce the issue**: minikube addons enable ingress

💣  enable failed: running callbacks: [addon apply: sudo KUBECONFIG=/var/lib/minikube/kubeconfig /var/lib/minikube/binaries/v1.17.0/kubectl apply -f /etc/kubernetes/addons/ingress-configmap.yaml -f /etc/kubernetes/addons/ingress-rbac.yaml -f /etc/kubernetes/addons/ingress-dp.yaml: Process exited with status 126
stdout:

stderr:
/var/lib/minikube/binaries/v1.17.0/kubectl: /var/lib/minikube/binaries/v1.17.0/kubectl: cannot execute binary file
]

</details>

</details>

**The operating system version**: macOSv10.15.2

priyawadhwa:
This way, all kubernetes binaries stored in ~/.minikube/cache/linux will be copied over to the VM and used for `minikube kubectl` on linux machines. kubectl will be stored in `~/.minikube/cache/{windows or darwin}` on windows/darwin so that `minikube kubectl` still works.

Should fix #6559

### DIFF
--- a/pkg/minikube/machine/cache_binaries.go
+++ b/pkg/minikube/machine/cache_binaries.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
 	"runtime"
 
 	"github.com/golang/glog"
@@ -63,7 +64,7 @@ func KubernetesReleaseURLSHA1(binaryName, version, osName, archName string) stri
 
 // CacheBinary will cache a binary on the host
 func CacheBinary(binary, version, osName, archName string) (string, error) {
-	targetDir := localpath.MakeMiniPath("cache", version)
+	targetDir := localpath.MakeMiniPath(filepath.Join("cache", osName), version)
 	targetFilepath := path.Join(targetDir, binary)
 
 	url := KubernetesReleaseURL(binary, version, osName, archName)
--- a/pkg/minikube/machine/cache_binaries.go
+++ b/pkg/minikube/machine/cache_binaries.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"path/filepath"
 	"runtime"
 
 	"github.com/golang/glog"
@@ -64,7 +63,7 @@ func KubernetesReleaseURLSHA1(binaryName, version, osName, archName string) stri
 
 // CacheBinary will cache a binary on the host
 func CacheBinary(binary, version, osName, archName string) (string, error) {
-	targetDir := localpath.MakeMiniPath(filepath.Join("cache", osName), version)
+	targetDir := localpath.MakeMiniPath("cache", osName, version)
 	targetFilepath := path.Join(targetDir, binary)
 
 	url := KubernetesReleaseURL(binary, version, osName, archName)
--- a/test/integration/aaa_download_only_test.go
+++ b/test/integration/aaa_download_only_test.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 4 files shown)


afbjorklund:
This means that the current cache will be invalidated, right ?

Should we set up a backwards compatible link or something

pkg/minikube/machine/cache_binaries.go
```diff
@@ -63,7 +64,7 @@ func KubernetesReleaseURLSHA1(binaryName, version, osName, archName string) stri
 
 // CacheBinary will cache a binary on the host
 func CacheBinary(binary, version, osName, archName string) (string, error) {
-	targetDir := localpath.MakeMiniPath("cache", version)
+	targetDir := localpath.MakeMiniPath(filepath.Join("cache", osName), version)
```
afbjorklund: I think you can use `MakeMiniPath("cache", osName, version)` here

afbjorklund:
@priyawadhwa : do you think we need to add the architecture to the path as well ?

hypothetical case about a linux-arm64 host running towards a linux-amd64 cluster

then again we can probably worry about that later, when we support remote machines
the image caching infrastructure, so we will need to fix more than just this single place

at the moment, all non-amd64 platforms requires using the `none` driver anyway.
that means that the linux binaries will be for the local platform, including kubernetes

afbjorklund:
This will also fix #6583

priyawadhwa:
@afbjorklund Yah I agree, I don't think we need to factor in architectures right now. 

I also don't think it's super necessary to set up a backwards compatible link. On the first run it'll take a few seconds to cache binaries in the right spot, but after that it shouldn't matter again. Unless you feel strongly, in which case I can add it in.

pkg/minikube/machine/cache_binaries.go
priyawadhwa: thanks, updated!

tstromberg:
Looks good. Do you mind following this up with a PR to add a `kubectl` command integration test to assert that we don't break it again in the future?

afbjorklund:
Sounds good, can worry about arch later - better to fix the darwin/Mac use case now.

test/integration/aaa_download_only_test.go
```diff
@@ -74,7 +74,7 @@ func TestDownloadOnly(t *testing.T) {
 
 				for _, img := range imgs {
 					img = strings.Replace(img, ":", "_", 1) // for example kube-scheduler:v1.15.2 --> kube-scheduler_v1.15.2
-					fp := filepath.Join(localpath.MiniPath(), "cache", "images", img)
+					fp := filepath.Join(localpath.MiniPath(), "cache", runtime.GOOS, "images", img)
```
afbjorklund: hmm, did we add some code changes for the images ? couldn't see any

priyawadhwa: nope, i added that to the wrong spot. just pushed the fix.

tstromberg:
These tests don't look good:

https://storage.googleapis.com/minikube-builds/logs/6586/VirtualBox_macOS.html

test/integration/aaa_download_only_test.go
```diff
@@ -83,7 +84,7 @@ func TestDownloadOnly(t *testing.T) {
 
 				// checking binaries downloaded (kubelet,kubeadm)
 				for _, bin := range constants.KubernetesReleaseBinaries {
-					fp := filepath.Join(localpath.MiniPath(), "cache", v, bin)
+					fp := filepath.Join(localpath.MiniPath(), "cache", runtime.GOOS, v, bin)
```
medyagh: for kubeadm,kubelet we always want linux, since we run it in linux VM or linux container.

medyagh:
we shouldn't cache kubeadm for mac os ! we need kubeadm and kublet always be linux !

test/integration/aaa_download_only_test.go
medyagh: the test should check for linux for kubeadm,kubelet...always.
since those are run always inside linux

priyawadhwa: done

priyawadhwa: done

medyagh:
thank you

afbjorklund:
Did you make kubectl download mandatory ? Though it was “on demand”

priyawadhwa:
@afbjorklund I figured it should be downloaded if `--download-only` is specified so that users can still use `minikube kubectl`

afbjorklund:
@priyawadhwa : good idea! I think I missed the context, when I just looked at the diff...

PR #6589, opened 2020-02-11
Don't install recommended packages in kic base image

priyawadhwa:
This reduces base image size from 816MB to 673MB (17.5% reduction)


medyagh:
@priyawadhwa nice finding !

Issue #6593, opened 2020-02-12
Upload test logs per commit


medyagh:
I noticed my html log was not being updated even in another browser .

But when I changed https to http it finally showed me the real latest logs.

I believe there is some sort of cdn happening for static files 
We better upload the file to a 6 digit comit number. So we see the real latest logs and also be able to see what was integration tests results for a specific commit

http://storage.googleapis.com/minikube-builds/logs/6515/Podman_Linux.html#failsection

**The exact command to reproduce the issue**:

**The full output of the command that failed**:<details>

</details>

**The output of the `minikube logs` command**:<details>

</details>

**The operating system version**:

PR #6609, opened 2020-02-13
upload "per commit" logs for integration test 

medyagh:
This PR will let every commit have its own Log File attached to it, and fixes and closes https://github.com/kubernetes/minikube/issues/6593

Before this PR:
all the test of different commits of a PR will have the same URL

https://storage.googleapis.com/minikube-builds/logs/6599/Docker_Linux.html

After this PR
each commit will have its own URL and avoids the race condition and possible overwrite of a latest log by a slower older commit

https://storage.googleapis.com/minikube-builds/logs/6599/abcdefg/Docker_Linux.html

abcdefg (would be shortened commit sha)

also fixes the CDN caching issue too (CDN caches an older .txt or .html) and you would have to do some tricks to see real latest logs


tstromberg:
Thanks for doing this!

PR #6617, opened 2020-02-13
WIP: Upgrade podman to 1.8.0

afbjorklund:
Skipped 1.7.0 and 1.7.1, because they were somewhat buggy.

https://github.com/containers/libpod/releases/tag/v1.8.0


afbjorklund:
@medyagh : Possibly we should have gone with 1.6.5 instead, for a more careful approach ?

These new things always seem to have some nasty bugs hidden, so always scared of a `.0`

afbjorklund:
Closed in favor of #6623, will do podman 1.8.0 and conmon 2.0.10 later

PR #6621, opened 2020-02-13
Upgrade buildroot minor version

afbjorklund:
Buildroot 2019.02.9

http://lists.busybox.net/pipermail/buildroot/2020-January/271410.html


afbjorklund:
Stumbled on the quotes...

`https://www.kernel.org/pub/linux/kernel/v4.x/linux-%224.19.94%22.tar.xz` (404)

PR #6629, opened 2020-02-14
Stopped and running machines should count as existing

sharifelgamal:


pkg/minikube/machine/fix.go
```diff
@@ -199,6 +199,9 @@ func adjustGuestClock(h hostRunner, t time.Time) error {
 // machineExists checks if virtual machine does not exist
 // if the virtual machine exists, return true
 func machineExists(d string, s state.State, err error) (bool, error) {
```
tstromberg: Maybe this should be called 'exists'?

medyagh:
thank you

PR #6635, opened 2020-02-14
improve delete by labeling created docker volumes 

medyagh:
closes https://github.com/kubernetes/minikube/issues/6662
might also close https://github.com/kubernetes/minikube/issues/6646

### DIFF
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -32,6 +32,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	cmdcfg "k8s.io/minikube/cmd/minikube/cmd/config"
+	"k8s.io/minikube/pkg/drivers/kic/oci"
 	"k8s.io/minikube/pkg/minikube/cluster"
 	"k8s.io/minikube/pkg/minikube/config"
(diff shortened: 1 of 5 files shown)


cmd/minikube/cmd/delete.go
```diff
@@ -97,7 +98,7 @@ func runDelete(cmd *cobra.Command, args []string) {
 
 	validProfiles, invalidProfiles, err := pkg_config.ListProfiles()
 	profilesToDelete := append(validProfiles, invalidProfiles...)
-
+	glog.Infof("error listing profiless %v", err)
```
afbjorklund: typo, and should probably check for nil ?

cmd/minikube/cmd/delete.go
```diff
@@ -112,8 +113,9 @@ func runDelete(cmd *cobra.Command, args []string) {
 			exit.UsageT("usage: minikube delete --all")
 		}
 
-		if err != nil {
```
afbjorklund: still need to check for the old err, before checking for the new ?

cmd/minikube/cmd/delete.go
medyagh: thanks done

Issue #5822, opened 2019-11-03
Build CRI and CNI from source ?


afbjorklund:
Currently we are downloading binaries for crictl and for cni and cni-plugins:

https://github.com/kubernetes-sigs/cri-tools/releases/

https://github.com/containernetworking/cni/releases/
https://github.com/containernetworking/plugins/releases/

Should we build these packages locally, from the source code, instead ?

priyawadhwa:
@afbjorklund is there an advantage to building these packages locally as opposed to downloading them?

afbjorklund:
@priyawadhwa : the advantage is that we _know_ what goes into them, and are OK if upstream stops building them

afbjorklund:
Just to prove a point, podman now requires CNI spec 0.4.0 and CNI 0.7.0 does not have binaries...

PR #6669, opened 2020-02-17
Build cni and cni-plugins from source instead

afbjorklund:
And make sure to install cnitool, not noop (!)

For #5822 and for upgrading to a later version


afbjorklund:
Note that `cnitool` has been broken ever since 41fdd53e1add6dd0173223da838ed0bdcd799845

I guess nobody actually uses that particular program ?

PR #6671, opened 2020-02-17
Adds python3 to Dockerized ISO build environment

edge0701:
This fixes the Dockerized ISO build that was broken due to the `varlink` package explicitly requiring `python3`.

To build, the steps:
```console
$ make buildroot-image
$ make out/minikube.iso
```
from the documentation [here](https://minikube.sigs.k8s.io/docs/contributing/iso/) were performed.

Failure before fix applied:
```
>>> varlink 18 Building
PATH="/mnt/out/buildroot/output/host/bin:/mnt/out/buildroot/output/host/sbin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"  PYTHONNOUSERSITE=y /mnt/out/buildroot/output/host/bin/ninja  -j17  -C /mnt/out/buildroot/output/build/varlink-18//build
ninja: Entering directory `/mnt/out/buildroot/output/build/varlink-18//build'
[1/49] Generating org.varlink.service.varlink with a custom command.
FAILED: lib/org.varlink.service.varlink.c.inc 
/mnt/out/buildroot/output/build/varlink-18/./varlink-wrapper.py ../lib/org.varlink.service.varlink lib/org.varlink.service.varlink.c.inc
/bin/sh: 1: /mnt/out/buildroot/output/build/varlink-18/./varlink-wrapper.py: not found
[2/49] Compiling C object 'tool/7c9bbe5@@varlink@exe/command.c.o'.
[3/49] Compiling C object 'lib/76b5a35@@test-error@exe/test-error.c.o'.
[4/49] Compiling C object 'tool/7c9bbe5@@varlink@exe/cli-bridge.c.o'.
[5/49] Compiling C object 'tool/7c9bbe5@@varlink@exe/command-help.c.o'.
[6/49] Compiling C object 'tool/7c9bbe5@@varlink@exe/cli-activate.c.o'.
[7/49] Compiling C object 'lib/76b5a35@@test-server-client@exe/test-server-client.c.o'.
[8/49] Compiling C object 'lib/76b5a35@@test-array@exe/test-array.c.o'.
[9/49] Compiling C object 'tool/7c9bbe5@@varlink@exe/command-complete.c.o'.
[10/49] Compiling C object 'lib/76b5a35@@test-interface@exe/test-interface.c.o'.
[11/49] Compiling C object 'tool/7c9bbe5@@varlink@exe/command-format.c.o'.
[...]

PR #6675, opened 2020-02-18
Need to use python3 from the path for varlink

afbjorklund:
Replaces #6671 

Varlink hardcodes `/usr/bin/python3`, but _should_ use `/usr/bin/env python3` and the $PATH.

We have a python3 version in `out/buildroot/output/host/bin/python3`, so don't need it installed...


tstromberg:
Code looks fine, but do you mind clarifying the PR title and description? I don't quite get it.

- The PR title makes it sound like the build process will use whatever version of python3 is installed in the host PATH. 
- The PR description makes it sound like we should be using one we've built by other means.

afbjorklund:
> Code looks fine, but do you mind clarifying the PR title and description? I don't quite get it.

Hopefully somewhat better now ?

afbjorklund:
Opened an upstream bug report at last: https://github.com/varlink/libvarlink/issues/18

PR #6681, opened 2020-02-19
Document to improve docker performance if docker driver is too slow

RA489:
Document if docker driver is too slow and how to improve its performance.


RA489:
Thanks for your review. I will update that it is vallid for only Mac.

RA489:
Updated the PR for Mac os only.

priyawadhwa:
Thank you for your contribution!

PR #6695, opened 2020-02-19
improve docker volume clean up

medyagh:
first delete container
then delete volume
then prune abandoned volumes that minikube created

### DIFF
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -115,7 +115,7 @@ func runDelete(cmd *cobra.Command, args []string) {
 		}
 		errs := oci.DeleteAllContainersByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "true"))
 		if errs != nil { // it will error if there is no container to delete
-			glog.Infof("no left over minikube  container found.", errs)
+			glog.Infof("no left over minikube  container found. %v", errs)
 		}
 		errs = oci.DeleteAllVolumesByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "true"))
 		if errs != nil { // it will not error if there is nothing to delete
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -186,7 +186,7 @@ func deleteProfile(profile *pkg_config.Profile) error {
 
 	errs := oci.DeleteAllContainersByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.ProfileLabelKey, profile.Name))
 	if errs != nil { // it will error if there is no container to delete
-		glog.Infof("no left over kic container for %s found to delete.", profile.Name, errs)
+		glog.Infof("no left over kic container for %s found to delete. %+v", profile.Name, errs)
 	}
 	errs = oci.DeleteAllVolumesByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.ProfileLabelKey, profile.Name))
 	if errs != nil { // it will not error if there is nothing to delete
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -113,13 +113,14 @@ func runDelete(cmd *cobra.Command, args []string) {
 		if profileFlag != constants.DefaultMachineName {
 			exit.UsageT("usage: minikube delete --all")
 		}
-		errs := oci.DeleteAllContainersByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "true"))
-		if errs != nil { // it will error if there is no container to delete
-			glog.Infof("no left over minikube  container found. %v", errs)
+		delLabel := fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "true")
+		errs := oci.DeleteAllContainersByLabel(oci.Docker, delLabel)
+		if errs != nil && len(errs) > 0 { // it will error if there is no container to delete
+			glog.Infof("error delete containers by label %q (might be okay): %+v", delLabel, err)
 		}
-		errs = oci.DeleteAllVolumesByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "true"))
-		if errs != nil { // it will not error if there is nothing to delete
-			glog.Warningf("error deleting left docker volumes. To see the list of volumes run: 'docker volume ls' \n:%+v", errs)
+		errs = oci.DeleteAllVolumesByLabel(oci.Docker, delLabel)
+		if errs != nil && len(errs) > 0 { // it will not error if there is nothing to delete
+			glog.Warningf("error delete volumes by label %q (might be okay): %+v", delLabel, errs)
 		}
 
 		errs = DeleteProfiles(profilesToDelete)
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -113,13 +113,16 @@ func runDelete(cmd *cobra.Command, args []string) {
 		if profileFlag != constants.DefaultMachineName {
 			exit.UsageT("usage: minikube delete --all")
 		}
-
-		err := oci.DeleteAllVolumesByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "true"))
-		if err != nil { // if there is no volume there won't be any error
-			glog.Warningf("error deleting left docker volumes. \n%v:\nfor more information please refer to docker documentation: https://docs.docker.com/engine/reference/commandline/volume_prune/", err)
+		errs := oci.DeleteAllContainersByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "true"))
+		if errs != nil { // it will error if there is no container to delete
+			glog.Infof("no left over kic container for %s found.", errs)
+		}
+		errs = oci.DeleteAllVolumesByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "true"))
+		if errs != nil { // it will not error if there is nothing to delete
+			glog.Warningf("error deleting left docker volumes. To see the list of volumes run: 'docker volume ls' \n:%+v", errs)
 		}
 
-		errs := DeleteProfiles(profilesToDelete)
+		errs = DeleteProfiles(profilesToDelete)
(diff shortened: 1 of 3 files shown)


cmd/minikube/cmd/delete.go
```diff
@@ -113,13 +113,16 @@ func runDelete(cmd *cobra.Command, args []string) {
 		if profileFlag != constants.DefaultMachineName {
 			exit.UsageT("usage: minikube delete --all")
 		}
-
-		err := oci.DeleteAllVolumesByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "=true"))
-		if err != nil { // if there is no volume there won't be any error
-			glog.Warningf("error deleting left docker volumes. \n%v:\nfor more information please refer to docker documentation: https://docs.docker.com/engine/reference/commandline/volume_prune/", err)
+		errs := oci.DeleteAllContainersByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "true"))
+		if errs != nil { // it will error if there is no container to delete
+			glog.Infof("no left over kic container for %s found.", errs)
```
priyawadhwa: Is %s meant to be the label name?

pkg/drivers/kic/oci/volumes.go
```diff
@@ -26,18 +28,48 @@ import (
...
+	return deleteErrs
+}
+
+// allVolumesByLabel returns name of all docker volumes by a specific label
+// will not return error if there is no volume found.
+func allVolumesByLabel(ociBin string, label string) ([]string, error) {
+	cmd := exec.Command(ociBin, "volume", "ls", "--filter", "label="+label, "--format", "{{.Name}}")
+	stdout, err := cmd.Output()
+	outs := strings.Split(strings.Replace(string(stdout), "\r", "", -1), "\n")
+	var tirmOuts []string
```
priyawadhwa: was this supposed to be `timeOuts`?

cmd/minikube/cmd/delete.go
```diff
@@ -113,13 +113,16 @@ func runDelete(cmd *cobra.Command, args []string) {
 		if profileFlag != constants.DefaultMachineName {
 			exit.UsageT("usage: minikube delete --all")
 		}
-
-		err := oci.DeleteAllVolumesByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "=true"))
-		if err != nil { // if there is no volume there won't be any error
-			glog.Warningf("error deleting left docker volumes. \n%v:\nfor more information please refer to docker documentation: https://docs.docker.com/engine/reference/commandline/volume_prune/", err)
+		errs := oci.DeleteAllContainersByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "true"))
+		if errs != nil { // it will error if there is no container to delete
+			glog.Infof("no left over minikube  container found. %v", errs)
```
tstromberg: This seems to be making a large assumption about the type of error without making it part of the API contract. How about:

`glog.Infof("delete %q containers: %v (may be ok)"), oci.CreatedByLabelKey, err)`

Please do this elsewhere instead of hardcoding "no left over" in your other error messages, to avoid the possibility of some very confusing error messages.

In the future, please consider making the oci API idempotent. It is generally not helpful to return errors for non-error conditions like this. When it comes to automating systems, any API that is not idempotent will result in fragile failure handling.

pkg/drivers/kic/oci/oci.go
```diff
@@ -34,6 +34,30 @@ import (
...
+		}
+	}
+	cs, err := listContainersByLabel(ociBin, label)
+	if err != nil {
+		glog.Infof("error listing containers by label %q: %v", label, err)
+	}
+	for _, c := range cs {
+		cmd := exec.Command(ociBin, "rm", "-f", "-v", c)
+		if out, err := cmd.CombinedOutput(); err != nil {
+			glog.Infof("error deleting container %s: output: %s", c, string(out))
```
tstromberg: Either log the error, or return it. 

Anytime a function does both, it results in duplicate error messages being sent to the log.

pkg/drivers/kic/oci/oci.go
```diff
@@ -34,6 +34,30 @@ import (
 	"strings"
 )
 
+// DeleteAllContainersByLabel deletes all containers that have a specific label
+// if contatiners founds will not return an error
```
tstromberg: I don't understand this sentence.

cmd/minikube/cmd/delete.go
```diff
@@ -113,13 +113,16 @@ func runDelete(cmd *cobra.Command, args []string) {
 		if profileFlag != constants.DefaultMachineName {
 			exit.UsageT("usage: minikube delete --all")
 		}
-
-		err := oci.DeleteAllVolumesByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "=true"))
-		if err != nil { // if there is no volume there won't be any error
-			glog.Warningf("error deleting left docker volumes. \n%v:\nfor more information please refer to docker documentation: https://docs.docker.com/engine/reference/commandline/volume_prune/", err)
+		errs := oci.DeleteAllContainersByLabel(oci.Docker, fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "true"))
+		if errs != nil { // it will error if there is no container to delete
```
tstromberg: This should probably be:

`if errs != nil && len(errs) > 0 {`

here and elsewhere. An empty list of errors is just as good as a nil.

pkg/drivers/kic/oci/oci.go
```diff
@@ -34,6 +34,30 @@ import (
...
+	var deleteErrs []error
+	if ociBin == Docker {
+		if err := PointToHostDockerDaemon(); err != nil {
+			return []error{errors.Wrap(err, "point host docker-daemon")}
+		}
+	}
+	cs, err := listContainersByLabel(ociBin, label)
+	if err != nil {
+		glog.Infof("error listing containers by label %q: %v", label, err)
+	}
```
tstromberg: return?

pkg/drivers/kic/oci/oci.go
```diff
@@ -406,16 +430,15 @@ func listContainersByLabel(ociBinary string, label string) ([]string, error) {
...
-	var b bytes.Buffer
-	cmd.Stdout = &b
-	cmd.Stderr = &b
-	err := cmd.Run()
-	var lines []string
-	sc := bufio.NewScanner(&b)
-	for sc.Scan() {
-		lines = append(lines, sc.Text())
+	stdout, err := cmd.Output()
+	outs := strings.Split(strings.Replace(string(stdout), "\r", "", -1), "\n")
```
tstromberg: This particular line feels weird to me. What is it doing?

tstromberg: Don't forget to check err.

pkg/drivers/kic/oci/volumes.go
```diff
@@ -26,18 +28,49 @@ import (
...
 		if err := PointToHostDockerDaemon(); err != nil {
-			return errors.Wrap(err, "point host docker-daemon")
+			return []error{errors.Wrap(err, "point host docker-daemon")}
 		}
 	}
-	cmd := exec.Command(ociBin, "volume", "prune", "-f", "--filter", label)
+
+	vs, err := allVolumesByLabel(ociBin, label)
+	if err != nil {
+		glog.Infof("error listing volumes by label %q: %v", label, err)
```
tstromberg: return?

pkg/drivers/kic/oci/volumes.go
```diff
@@ -26,18 +28,49 @@ import (
...
+
+// allVolumesByLabel returns name of all docker volumes by a specific label
+// will not return error if there is no volume found.
+func allVolumesByLabel(ociBin string, label string) ([]string, error) {
+	cmd := exec.Command(ociBin, "volume", "ls", "--filter", "label="+label, "--format", "{{.Name}}")
+	stdout, err := cmd.Output()
+	outs := strings.Split(strings.Replace(string(stdout), "\r", "", -1), "\n")
+	var vols []string
+	for _, o := range outs {
+		if strings.TrimSpace(o) != "" {
```
tstromberg: it'd be cleaner if the code did not call TrimSpace twice on the same string - here and elsewhere.

I also can't help but suspect that this will be cleaner and faster:

```
   s := bufio.NewScanner(stdout)
   for s.Scan() {
        if s.Text() != "" {
          vols = append(vols, s.Text())
        }
    }
```

pkg/drivers/kic/oci/volumes.go
```diff
@@ -26,18 +28,49 @@ import (
...
-	return nil
+	return deleteErrs
+}
+
+// allVolumesByLabel returns name of all docker volumes by a specific label
+// will not return error if there is no volume found.
+func allVolumesByLabel(ociBin string, label string) ([]string, error) {
+	cmd := exec.Command(ociBin, "volume", "ls", "--filter", "label="+label, "--format", "{{.Name}}")
+	stdout, err := cmd.Output()
+	outs := strings.Split(strings.Replace(string(stdout), "\r", "", -1), "\n")
```
tstromberg: Is the strings.Replace call really necessary? It seems like on Windows we are already calling strings.TrimSpace anyways.

pkg/drivers/kic/oci/volumes.go
medyagh: then reason I am not returning is because I wanna do a prune in the end anyways.
prune would not work on the in-use volumes. so at first I try delete them without prune
and use prune as last effort.

pkg/drivers/kic/oci/volumes.go
medyagh: thanks will fix 👍

pkg/drivers/kic/oci/volumes.go
medyagh: docker output is very insconsistent made me paranoidly remove spaces anywhere I can.

cmd/minikube/cmd/delete.go
medyagh: done

cmd/minikube/cmd/delete.go
medyagh: done

pkg/drivers/kic/oci/oci.go
medyagh: done

pkg/drivers/kic/oci/volumes.go
medyagh: based on your review, I ended up breaking this function to two functions. so I can return on errors better

PR #6696, opened 2020-02-19
Fix bug in --install-addons flag

priyawadhwa:
Respect --install-addons flag, don't install anything if it is false


medyagh:
unit test failure is not related to your PR, travis is crazy today

Issue #6711, opened 2020-02-20
TestGetOK  unit test failure


medyagh:
this unit test test fails on travis but not on my linux

```
--- FAIL: TestGetOK (0.00s)
    get_test.go:32: Set returned error for property vm-driver
--- FAIL: TestSetOK (0.00s)
    set_test.go:55: Set returned error for valid property value
    set_test.go:51: failed to unset vm-driver
--- FAIL: TestUnsetConfig (0.00s)
    unset_test.go:30: Failed to set the property "cpus"
    unset_test.go:35: Failed to read config "specified key could not be found in config"
    unset_test.go:39: Expected cpus to be 1 but got 
    unset_test.go:44: Failed to unset property "create /home/travis/.minikube/config/config.json: open /home/travis/.minikube/config/config.json: no such file or directory"
```

sayboras:
@medyagh sorry, I didn't see your PR before, I fixed this in https://github.com/kubernetes/minikube/pull/6726/files

PR #6726, opened 2020-02-21
Fix unit test

sayboras:
Fixes https://github.com/kubernetes/minikube/issues/6711

Unit tests are passing in local dev environemnt as there is file in `$HOME/.minikube/config/config.json`, however, this file is not available in CI docker environment. 

**Approach:**

- Generate the temporary config file as part of test cases.
- Setting MinikubeHome environment variable pointint to temp config.

### DIFF
--- a/cmd/minikube/cmd/config/get_test.go
+++ b/cmd/minikube/cmd/config/get_test.go
@@ -16,24 +16,28 @@ limitations under the License.
 
 package config
 
-import "testing"
+import (
+	"testing"
+)
 
 func TestGetNotFound(t *testing.T) {
+	createTestConfig(t)
 	_, err := Get("nonexistent")
-	if err == nil {
+	if err == nil || err.Error() != "specified key could not be found in config" {
 		t.Fatalf("Get did not return error for unknown property")
 	}
 }
 
 func TestGetOK(t *testing.T) {
+	createTestConfig(t)
 	name := "vm-driver"
 	err := Set(name, "virtualbox")
 	if err != nil {
-		t.Fatalf("Set returned error for property %s: %v", name, err)
+		t.Fatalf("Set returned error for property %s, %+v", name, err)
 	}
 	val, err := Get(name)
 	if err != nil {
-		t.Fatalf("Get returned error for property %s: %v", name, err)
+		t.Fatalf("Get returned error for property %s, %+v", name, err)
 	}
 	if val != "virtualbox" {
 		t.Fatalf("Get returned %s, expected virtualbox", val)
--- a/cmd/minikube/cmd/config/set_test.go
+++ b/cmd/minikube/cmd/config/set_test.go
@@ -19,23 +19,21 @@ package config
 import (
 	"io/ioutil"
 	"os"
-	"path/filepath"
 	"testing"
 
-	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/localpath"
 )
 
 func TestNotFound(t *testing.T) {
-	createTestProfile(t)
+	createTestConfig(t)
 	err := Set("nonexistent", "10")
-	if err == nil {
+	if err == nil || err.Error() != "property name \"nonexistent\" not found" {
 		t.Fatalf("Set did not return error for unknown property")
 	}
 }
 
 func TestSetNotAllowed(t *testing.T) {
-	createTestProfile(t)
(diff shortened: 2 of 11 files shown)


Makefile
```diff
@@ -29,7 +29,7 @@ DEB_VERSION ?= $(subst -,~,$(RAW_VERSION))
 RPM_VERSION ?= $(DEB_VERSION)
 
 # used by hack/jenkins/release_build_and_upload.sh and KVM_BUILD_IMAGE, see also BUILD_IMAGE below
-GO_VERSION ?= 1.13.4
+GO_VERSION ?= 1.13.7
```
medyagh: why is it not 1.13.8 like all others?

hack/jenkins/common.sh
```diff
@@ -31,7 +31,7 @@ export KUBECONFIG="${TEST_HOME}/kubeconfig"
 export PATH=$PATH:"/usr/local/bin/:/usr/local/go/bin/:$GOPATH/bin"
 
 # installing golang so we could do go get for gopogh
-sudo ./installers/check_install_golang.sh "1.13.4" "/usr/local" || true
+sudo ./installers/check_install_golang.sh "1.13.7" "/usr/local" || true
```
medyagh: there is another one also, in 

hack/jenkins/minikube_cross_build_and_upload.sh

sayboras: Thanks :100:

sayboras: this file already retrieved goland version from Makefile in command `WANT_GOLANG_VERSION=$(grep '^GO_VERSION' Makefile | awk '{ print $3 }')`

cmd/minikube/cmd/config/get_test.go
```diff
@@ -16,24 +16,28 @@ limitations under the License.
...
-import "testing"
+import (
+	"testing"
+)
 
 func TestGetNotFound(t *testing.T) {
+	createTestConfig(t)
 	_, err := Get("nonexistent")
-	if err == nil {
+	if err == nil || err.Error() != "specified key could not be found in config" {
```
sayboras: Verify the error message, as err might be something else.

pkg/minikube/localpath/localpath.go
```diff
@@ -32,7 +32,9 @@ import (
 const MinikubeHome = "MINIKUBE_HOME"
 
 // ConfigFile is the path of the config file
-var ConfigFile = MakeMiniPath("config", "config.json")
+func ConfigFile() string {
```
sayboras: Change this package variable into func(). Otherwise, setting MinikubeHome env in test cases will not have any impact.

cmd/minikube/cmd/config/set_test.go
```diff
@@ -19,68 +19,62 @@ package config
...
-		t.Fatalf("Get returned error for valid property")
+		t.Fatalf("Get returned error for valid property: %+v", err)
 	}
 	if val != "virtualbox" {
 		t.Fatalf("Get returned %s, expected \"virtualbox\"", val)
 	}
 }
 
-func createTestProfile(t *testing.T) {
+func createTestConfig(t *testing.T) {
```
sayboras: config.Get/Set/Unset functions are reading the values from config/config.json only. So creating temp profile here doesn't make sense. 

I have renamed the function and change implementation to create team config.

hack/jenkins/common.sh
sayboras: I move golang upgrade to another PR

medyagh:
oopps sorry @sayboras I didn't see your PR, the PR looks good, just ned a rebase.

also how about specifiying a tmp MINIKUBE_HOME for unit tests so we don't use user's MINIKUBE_HOME ?


any "make test" should probably create its own temp dir (something that works on all platforms)

sayboras:
> oopps sorry @sayboras I didn't see your PR, the PR looks good, just ned a rebase.
> 
> also how about specifiying a tmp MINIKUBE_HOME for unit tests so we don't use user's MINIKUBE_HOME ?
> 
> any "make test" should probably create its own temp dir (something that works on all platforms)

No worry, I just re-based with master. 

Agreed that the unit test should create its own MINIKUBE_HOME, so I called method `createTestConfig(t)` before the test to do so. Let me know if you have any other comments.

medyagh:
@sayboras thank you for  this fix

PR #6737, opened 2020-02-21
WIP Add dependency for integration test job

sayboras:
Integration test jobs are depending on `build` step only. So it's unnecessary to run these job if `lint` or `unit test` failed. Additionally, we didn't save much time for head start as build, lint and unit test are running in parallel.

Making integration tests run only after lint and unit test jobs passed will help contributors to be aware of the issues earlier.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,7 +66,7 @@ jobs:
   # Run the following integration tests after the build_minikube
   # They will run in parallel and use the binaries in previous step
   docker_ubuntu_16_04:
-    needs: [build_minikube]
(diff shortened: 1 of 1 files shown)


medyagh:
@sayboras it is okay to run the integration tests if the lint or even the unit test fails. sometimes a lint or a unit test is flaky and we still like to see the integration tests for it

sayboras:
> @sayboras it is okay to run the integration tests if the lint or even the unit test fails. sometimes a lint or a unit test is flaky and we still like to see the integration tests for it

Oh I see :(

PR #6754, opened 2020-02-22
add sudo to systemctl commands in integration tests

medyagh:
For docker driver need to run systemctl with sudo 
to pass TestDockerFlags

```
docker@minikube:~$ systemctl show docker --property=Environment --no-pager
Failed to connect to bus: No such file or directory
but it passes with sudo

docker@minikube:~$ sudo systemctl show docker --property=Environment --no-pager
Environment=FOO=BAR BAZ=BAT

```

### DIFF
--- a/test/integration/docker_test.go
+++ b/test/integration/docker_test.go
@@ -42,7 +42,7 @@ func TestDockerFlags(t *testing.T) {
 		t.Errorf("%s failed: %v", rr.Args, err)
 	}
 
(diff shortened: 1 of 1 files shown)


afbjorklund:
I think the reason for the error output is the lack of a session dbus-daemon for the test runner

I'm not sure which is worst, adding one for the user or run sudo towards the system dbus ?

PR #6761, opened 2020-02-23
Fix TestDownloadOnly integration test

sayboras:
Addresses https://github.com/kubernetes/minikube/issues/6742

`StartArgs` was not passed to command.

TestDownloadOnly failed for all driver as per https://github.com/kubernetes/minikube/issues/6742.

Issue #6742, opened 2020-02-22
fix TestDownloadOnly on github actions


medyagh:
only on github actions the following tests fail

"
  "TestDownloadOnly/group/v1.11.10",
  "TestDownloadOnly/group/v1.17.3",
  "TestDownloadOnly/group/v1.17.3#01"
"

as seen here 
https://github.com/kubernetes/minikube/pull/6741/checks?check_run_id=461427007

the logs for them can be found here:

PR #6762, opened 2020-02-23
Fix TestDownloadOnly for --vm-driver=none

sayboras:
Do not verify the cache image (e.g. kube-proxy, kube-scheduler, kube-controller-manager, etc) for vm-driver=none as per existing logic in below files

```go
startCmd.Flags().Bool(cacheImages, true, "If true, cache docker images for the current bootstrapper and load them into the machine. Always false with --vm-driver=none.")
```

https://github.com/kubernetes/minikube/blob/master/pkg/minikube/driver/driver.go#L141
https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/start.go#L156

Related PR https://github.com/kubernetes/minikube/pull/6761
Related Issue https://github.com/kubernetes/minikube/issues/6742

### DIFF
--- a/test/integration/aaa_download_only_test.go
+++ b/test/integration/aaa_download_only_test.go
@@ -73,7 +73,7 @@ func TestDownloadOnly(t *testing.T) {
 				}
 
 				// skip verify for cache images if --vm-driver=none
-				if !strings.Contains(strings.Join(StartArgs(), " "), "--vm-driver=none") {
+				if !strings.Contains(strings.Join(rrr.Args, " "), "--vm-driver=none") {
 					for _, img := range imgs {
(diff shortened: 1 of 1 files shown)


medyagh:
Thank you very much for fixing up minikube bugs. You are doing a great job.

We got a couple more test flakes and good first issues I would love to see more contributions from you.

Any interest to add windows tests to GitHub actions ?

medyagh:
@sayboras
There are a few other tests that only fail on GitHub actions. On docker driver and not jenkins would be nice to find the root cause for example

  "TestFunctional/parallel/MySQL",
  "TestFunctional/parallel/DockerEnv"

sayboras:
Oh sure, let me take a look on above linux failure, hope that i can figure something out. :100:

Issue #6525, opened 2020-02-06
Request: Support ppc64le


edmondsw:
I want to run minikube on ppc64le, but I found that there are no minikube-linux-ppc64le artifacts built today. I'd specifically like to see v1.2.0 and newer work with a command like this:

`curl -Lo /usr/local/bin/minikube https://storage.googleapis.com/minikube/releases/v1.2.0/minikube-linux-ppc64le`

Having minikube for ppc64le can be very useful for multi-arch CI testing of a number of projects (e.g. see https://github.com/bsycorp/kind), for instance.

I found this existing issue which was closed due to lack of a specific architecture request, so I'm specifically asking for ppc64le here: https://github.com/kubernetes/minikube/issues/3890

But I also found that there's another issue essentially asking for arm64 support: https://github.com/kubernetes/minikube/issues/6280

And I'm aware of some folks that will probably be asking for support for s390x very shortly as well.

The primary issue may be how to test on non-amd64 architectures. Fortunately, Travis CI now has workers for all of these architectures. See https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z . So minikube could presumably use Travis to test on all of those architectures. It could even replace the current cross-compilation logic with native builds using Travis, but that wouldn't necessarily be required.

afbjorklund:
Cross-compiling for all Kubernetes architectures can probably be done.

We also need to update the Docker builds to build for all architectures

afbjorklund:
Platforms:
amd64
arm (32-bit)
arm64
ppc64le
s390x 

https://github.com/kubernetes/community/blob/master/contributors/design-proposals/multi-platform.md

afbjorklund:
We have KVM support for `amd64` and `arm64`*. Not sure which hypervisors are supported on the other architectures, so probably only native (none) support.

\* No ISO for ARM, though.

afbjorklund:
@edmondsw : can you describe the use case a little ? Like how Kubernetes would be used on the PPC etc. Think we already have S/390, but no official binaries... (but also interested in uses on _that_ mainframe)

edmondsw:
ppc64le supports both KVM and PowerVM hypervisors. There are several hypervisors supported on s390x as well, one of which is KVM.

The use case I'm currently looking at would be to use in https://github.com/bsycorp/kind so you could check that out to see how it's using it.

afbjorklund:
[...]

Issue #5667, opened 2019-10-19
Add support for running on arm64 (aarch64)


afbjorklund:
For instance on the Raspberry Pi, using the `none` driver (localhost).

Note that you will need more than 1G of memory, or use a swap file.

No virtualization on this platform, but works OK with `generic` driver.

That is, work on the laptop and use ssh to connect to the "bare metal".

afbjorklund:
```console
# minikube start --vm-driver=none
😄  minikube v1.4.1 on Fedora 30 (/arm64)
🤹  Running on localhost (CPUs=4, Memory=953MB, Disk=29198MB) ...
ℹ️   OS release is Fedora 30 (Server Edition)
🐳  Preparing Kubernetes v1.16.1 on Docker 18.09.9 ...
🚜  Pulling images ...
🚀  Launching Kubernetes ... 
🤹  Configuring local host environment ...

⚠️  The 'none' driver provides limited isolation and may reduce system security and reliability.
⚠️  For more information, see:
👉  https://minikube.sigs.k8s.io/docs/reference/drivers/none/

⚠️  kubectl and minikube configuration will be stored in /root
⚠️  To use kubectl or minikube commands as your own user, you may need to relocate them. For example, to overwrite your own settings, run:

    ▪ sudo mv /root/.kube /root/.minikube $HOME
    ▪ sudo chown -R $USER $HOME/.kube $HOME/.minikube

💡  This can also be done automatically by setting the env var CHANGE_MINIKUBE_NONE_USER=true
⌛  Waiting for: apiserver proxy etcd scheduler controller dns
🏄  Done! kubectl is now configured to use "minikube"
```

* https://www.raspberrypi.org/products/raspberry-pi-3-model-b/

* https://fedoraproject.org/wiki/Architectures/ARM/Raspberry_Pi

* https://download.docker.com/linux/static/stable/aarch64/

* https://docs.docker.com/config/daemon/systemd/

nanikjava:
This is going to be interesting. 

I've got ODROID XU4 (https://www.hardkernel.com/shop/odroid-xu4-special-price/)  and it has 2GB RAM and can run Kernel 4.14 LTS

Interested to help out to make it run on the ODROID board. 

Thanks

afbjorklund:
The dedicated boards are probably a better match, all the Raspberry Pi (except for the latest model 4) have either 512M or 1G of memory. And this is **not** enough for running `kubeadm` (only for `localkube`)

[...]

PR #6764, opened 2020-02-23
Build minikube cross for exotic architectures

afbjorklund:
Split out the non-amd64 from the linux arch,
and build those "exotic" arch separately.

The default is to build all operating systems,
so restore that definition of the "cross".

----

Build for all Kubernetes architectures:
```
out/minikube-linux-amd64:   ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/l, not stripped
out/minikube-linux-arm:     ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), statically linked, not stripped
out/minikube-linux-arm64:   ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, not stripped
out/minikube-linux-ppc64le: ELF 64-bit LSB executable, 64-bit PowerPC or cisco 7500, version 1 (SYSV), statically linked, not stripped
out/minikube-linux-s390x:   ELF 64-bit MSB executable, IBM S/390, version 1 (SYSV), statically linked, not stripped
```

For #5667 and #6525 and #6843

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,7 +243,7 @@ test-pkg/%: pkg/minikube/assets/assets.go pkg/minikube/translate/translations.go
 	go test -v -test.timeout=60m ./$* --tags="$(MINIKUBE_BUILD_TAGS)"
 
 .PHONY: all
-all: cross drivers e2e-cross out/gvisor-addon ## Build all different minikube components
+all: cross drivers e2e-cross exotic out/gvisor-addon ## Build all different minikube components
 
 .PHONY: drivers
 drivers: docker-machine-driver-hyperkit docker-machine-driver-kvm2 ## Build Hyperkit and KVM2 drivers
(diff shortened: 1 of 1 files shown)


afbjorklund:
This should probably wait, until we have images for all the addons (e.g. provisioner)

Makefile
```diff
@@ -302,7 +302,10 @@ endif
 	@sed -i -e 's/Json/JSON/' $@ && rm -f ./-e
 
 .PHONY: cross
-cross: minikube-linux-amd64 minikube-linux-arm64 minikube-darwin-amd64 minikube-windows-amd64.exe ## Build minikube for all platform
+cross: minikube-linux-amd64 minikube-darwin-amd64 minikube-windows-amd64.exe ## Build minikube for all platform
+
+.PHONY: exotic
```
medyagh: add ## help text what exotic is to the new contributor

afbjorklund: Sure thing, basically here it means non-amd64 (i.e. non-standard)

afbjorklund:
Main reason these architectures are secondary, is because we don't have ISO or KIC for them (yet?)

So that means that they have to run with `none` or `generic' driver - they _are_ supported by kubeadm

afbjorklund:
And probably the 80-20 split in this case is more like 99-1, I think there's very few clusters ?

But I do think that the `arm64` is turning out to be a worthy competitor to the `amd64`...
Especially when it comes to power consumption, but also a couple of new server models

The `arm` is more or less too small for regular k8s, and the mainframe targets are niche.

As documented elsewhere, it is going to be hard to test or support these extra targets.
But dropping binaries for all kubernetes platforms over the fence, that much we can do ?

afbjorklund:
Images for storage-provisioner are now build and pushed by @medyagh

Makefile
medyagh: once ## comment added for this we can merge this

afbjorklund: Thought I added it ?

Makefile
```diff
@@ -302,7 +302,10 @@ endif
...
+exotic: out/minikube-linux-arm out/minikube-linux-arm64 out/minikube-linux-ppc64le out/minikube-linux-s390x ## Build minikube for non-amd64 linux
```
afbjorklund: Added new help text

PR #6768, opened 2020-02-24
Add travis builds for s390x architecture 

prankkelkar:
As Travis CI [officially supports](https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z) s390x builds, adding support for the same.


medyagh:
I don't understand this pr could you please describe what you are trying to solve exactly?

prankkelkar:
This PR will enable travis builds on s390x architecture. We are looking forward to add support for s390x architecture and get minikube binaries released for the same. These CI builds will be necessary to verify any incoming changes/PR's in the future.

Explanation for changes in test files : 
* Replaced hard coded `amd64` values with a generic function `archTag` thereby fixing test case failures for architectures other than amd64 .

afbjorklund:
> This PR will enable travis builds on s390x architecture. We are looking forward to add support for s390x architecture and get minikube binaries released for the same. These CI builds will be necessary to verify any incoming changes/PR's in the future.

While we do want to enable multi-arch support in minikube, it would still nice to have an issue describing further the motives and the use cases behind adding the extra "exotic" architectures.

We are trying to get away from having the architecture in the image names, and instead have multi-arch manifest - but like you noticed this has not yet been done for the storage provisioner #6007

However, it should **not** be needed for standarding images like pause and kubernetes ?

Are you sure that these won't install just fine using the regular name, with the multi manifest ?

https://kubernetes.io/docs/concepts/containers/images/#building-multi-architecture-images-with-manifests

afbjorklund:
If you could rebase to get rid of the merges and the reverts, that would also be appreciated

afbjorklund:
I'm not sure running linting/building/unit testing on another architecture brings much value.
What we would need to do, is to take those binaries and run some real integration tests...

And I don't think we can use Travis for that, but would need a "real" testing infrastructure ?
Similar to the nodes that we have for real cross builds done today: linux/darwin/windows

----

Since most of the effort now goes into KIC, I think we will wait with multi-arch until next time.

We could still add _builds_ for exotic arch, but without provisioner images it's not really complete.

#6764 build minikube binaries for all 5 kubernetes linux architectures (amd64+4)

#6007 build storage-provisioner images for all 5 kubernetes linux architectures

We are also moving some of the Travis job to GitHub Actions, this would only confuse further.
(duplicated jobs between travis and github, and now also duplicated between amd64 and s390x?)

Personally I am working with `arm64`, since it is easier to set up than `s390x` and `ppc64le`
It is nice that IBM has donated these 40 containers to Travis, but sometimes you want your own?

> It’s LXD based and your build jobs are run within an LXD container. <https://linuxcontainers.org/lxd/>
_(interesting to notice that Travis is going back to containers again, having left OpenVZ earlier...)_

> The initial infrastructural build capacity has a total of 40 concurrent jobs on IBM Z (s390x) and 64 concurrent jobs on IBM Power (ppc64le) split between travis-ci.org and travis-ci.com.

prankkelkar:
Minikube is a great tool to quickly setup a single node kubernetes cluster. It would be nice to have minikube is offically supported on s390x architecture.
> However, it should not be needed for standarding images like pause and kubernetes ?
> Are you sure that these won't install just fine using the regular name, with the multi manifest ?


Yes you are right. Standard kubernetes images like pause does not need arch suffix in the image name. It works fine without arch names suffixed to it and this is expected since they are built with multiarch manifest. 

> And I don't think we can use Travis for that, but would need a "real" testing infrastructure ?
> Similar to the nodes that we have for real cross builds done today: linux/darwin/windows

As you said travis handles basic unit tests and linting. We could provide VM's if they are needed for setting up e2e integration tests.

If we can get multiarch storage provisioner image and binaries for minikube it would be sufficient to get basic kubernetes cluster up with none driver. If you need any help my my side please let me know.

prankkelkar:
> If you could rebase to get rid of the merges and the reverts, that would also be appreciated

Rebased the branch to clean up extra commits.

afbjorklund:
> /ok-to-test

@ RA489 : Why ? We **don't** want to merge this

----

@prankkelkar
Binaries for `s390x` have now been added (for minikube and storage-provisioner), for 1.8.0
We will need to see if we can set up some automatic testing for the mainframe architectures...

It would be great if you could test the new release, and feedback how it is working on IBM ?
Hopefully we can merge the provisioner image too into a multi-arch, for the next release...

prankkelkar:
Thanks @afbjorklund . Currently i am trying to cross-compile master and verify it on s390x platform. Once the binaries are out i can test them and see if it has any issues.

afbjorklund:
Now available here:
https://storage.googleapis.com/minikube/releases/latest/minikube-linux-s390x

prankkelkar:
@afbjorklund I have verified s390x binaries for kubernetes version 1.15 and 1.16. I am able to get basic cluster up with none driver.  There is an [issue ](https://github.com/kubernetes/kubernetes/issues/87197) with kubernetes v1.17.x on s390x platform. Currently i could verify minikube for 1.17.x using the workaround found [here](https://github.com/kubernetes/kubernetes/issues/87197#issuecomment-588897102). Issues what i faced were related to kubernetes and not minikube hence can conclude that **minikube binary is working on s390x.**

prankkelkar:
@afbjorklund I was just thinking whether changes in this PR other than travis file are still needed? 
Its always nice to use variables rather than using hard coded strings. What are your views on this?

afbjorklund:
I didn’t see anything else ? The arch strings are unwanted and unneeded, since the images are now multi-arch

prankkelkar:
Yes. you are right but in that case we should remove code appending arch strings from the source code as well.

afbjorklund:
Discussion moved to #6798

PR #6773, opened 2020-02-24
Fix none-test integration test

sayboras:
Fix TestChangeNoneUser for non driver integration test.

As per https://github.com/sayboras/minikube/pull/190/checks?check_run_id=494463943, this test was failing.

The test was expecting KUBECONFIG from default directory, but not from environment variable.

Due to recent changes in docker overlap, lz4 is required in host system, hence I added the steps to install this before running integration test. It might be not required for podman for now.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,6 +88,11 @@ jobs:
         docker info || true
         docker version || true
         docker ps || true
+    - name: install lz4
+      shell: bash
+      run: |
+        sudo apt-get update -qq
+        sudo apt-get -qq -y install liblz4-tool
     - name: Install gopogh
       shell: bash
       run: |
@@ -150,6 +155,11 @@ jobs:
       SHELL: "/bin/bash" # To prevent https://github.com/kubernetes/minikube/issues/6643
     needs: [build_minikube]
     steps:
+    - name: install lz4
+      shell: bash
+      run: |
+        sudo apt-get update -qq
+        sudo apt-get -qq -y install liblz4-tool
     - name: Docker Info
       shell: bash
       run: |
@@ -218,6 +228,11 @@ jobs:
       SHELL: "/bin/bash" # To prevent https://github.com/kubernetes/minikube/issues/6643
     runs-on: ubuntu-16.04
     steps:
+    - name: install lz4
+      shell: bash
+      run: |
+        sudo apt-get update -qq
+        sudo apt-get -qq -y install liblz4-tool
     - name: Install gopogh
       shell: bash
       run: |
@@ -280,6 +295,11 @@ jobs:
       SHELL: "/bin/bash" # To prevent https://github.com/kubernetes/minikube/issues/6643
     runs-on: ubuntu-18.04
     steps:
+    - name: install lz4
+      shell: bash
+      run: |
+        sudo apt-get update -qq
+        sudo apt-get -qq -y install liblz4-tool
     - name: Install gopogh
       shell: bash
       run: |
@@ -342,6 +362,11 @@ jobs:
         SHELL: "/bin/bash" # To prevent https://github.com/kubernetes/minikube/issues/6643
       runs-on: ubuntu-18.04
       steps:
+      - name: install lz4
+        shell: bash
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get -qq -y install liblz4-tool
       - name: install podman
         shell: bash
         run: |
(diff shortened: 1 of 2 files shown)


sayboras:
> /ok-tot-test

@medyagh can you put `/ok-to-test` one more time ? Thanks

medyagh:
> ok-to-test

the ok-to-test doesnt need renewal but because of github rate limit we check every 10 minutes for changes.
so most of the time in less than 10 mins it will trigger a new test

pkg/minikube/tunnel/kic/ssh_tunnel.go
```diff
@@ -118,7 +118,7 @@ func (t *SSHTunnel) stopMarkedConnections() {
 
 // sshConnName creates a uniq name for the tunnel, using its name/clusterIP/ports.
 // This allows a new process to be created if an existing service was changed,
-// the new process will support the IP/Ports change ocurred.
+// the new process will support the IP/Ports change occurred.
```
medyagh: thanks :)

medyagh:
@sayboras 
it is werid that on github actions this PR adds two new test failures...which I dont know why but fixes the ChangeNone...

I wonder if those two tests use a wrong kube config or something like that ?

pkg/minikube/tunnel/kic/ssh_tunnel.go
medyagh: could u also fix the other lint error on master?
"
consider running: /home/runner/work/minikube/minikube/hack/boilerplate/fix.sh"
Feel free to make a new PR both lints I will merge quickly.

sayboras:
Sure. Let me do it now.

In your previous comment, there was typo in ok-to-test, maybe its why jenkins didn’t kick off.

medyagh:
@sayboras no as long as there is a label called ok-to-test jenkins will pull every 10 minutes to see if there is any new commit to run.

sayboras:
@medyagh 

Oh thanks for your clarification. However, the previous comment you put is `/ok-tot-test` (there was extra t there). I can see the label still `needs-ok-to-test`

medyagh:
> @medyagh
> 
> Oh thanks for your clarification. However, the previous comment you put is `/ok-tot-test` (there was extra t there). I can see the label still `needs-ok-to-test`

you are right ! my bad :) 
/ok-to-test

medyagh:
it is interesting, these two tests constistanly fail, only on none driver on this PR:
```
  "TestStartStop/group/old-docker",
  "TestVersionUpgrade"
```

medyagh:
one difference I see in old-docker test is
it has --keep-context=false (do not keep the kubecontext)
```
			{"old-docker", constants.OldestKubernetesVersion, []string{
				// default is the network created by libvirt, if we change the name minikube won't boot
				// because the given network doesn't exist
				"--kvm-network=default",
				"--kvm-qemu-uri=qemu:///system",
				"--disable-driver-mounts",
				"--keep-context=false",
				"--container-runtime=docker",
			}}
```

sayboras:
Thanks for your pointer, let me investigate on these two failures :100:

medyagh:
@sayboras still interested to finish this PR ?

sayboras:
Yup, I am :), hope that it can be done this weekends

test/integration/start_stop_delete_test.go
```diff
@@ -50,7 +50,6 @@ func TestStartStop(t *testing.T) {
 				"--kvm-network=default",
 				"--kvm-qemu-uri=qemu:///system",
 				"--disable-driver-mounts",
-				"--keep-context=false",
```
medyagh: We do not want to not test this option ! we still want to test this option, but make sure the test is looking at the right thing.

medyagh:
hi @sayboras any update on this ?

sayboras:
> hi @sayboras any update on this ?

@medyagh sorry for late reply. Tried to setup my local development to to run test inside docker to have faster feedback. Still unable to complete this due to multiple priveledge container issue.

However, TestChangeNoneUser seems to be fixed now, and it didn't introduced any more failure compared to master branch. Here is the CI for master branch for your reference https://github.com/sayboras/minikube/pull/190/checks?check_run_id=494463943.

Now, integration test are failing for below cases, which can be addressed in seperate PR.

```
[
  "TestOffline/group/docker",
  "TestStartStop/group/old-docker",
  "TestVersionUpgrade"
]
```

Please let me know if anything else is required.

tstromberg:
Thank you!

PR #6781, opened 2020-02-24
Fix typo detected by linter

sayboras:
Fixes typo caused travis and gh action failed.

@medyagh I feel like we should enable the `required` check  for at least (build and lint gh action) to avoid such occurrence later. There might be the case that golangci-lint misses some rule, however, it still helps.

### DIFF
--- a/pkg/minikube/tunnel/kic/ssh_tunnel.go
+++ b/pkg/minikube/tunnel/kic/ssh_tunnel.go
@@ -118,7 +118,7 @@ func (t *SSHTunnel) stopMarkedConnections() {
 
 // sshConnName creates a uniq name for the tunnel, using its name/clusterIP/ports.
 // This allows a new process to be created if an existing service was changed,
-// the new process will support the IP/Ports change ocurred.
+// the new process will support the IP/Ports change occurred.
 func sshConnUniqName(service v1.Service) string {
 	n := []string{
 		service.Name,
--- a/pkg/minikube/tunnel/kic/ssh_tunnel.go
+++ b/pkg/minikube/tunnel/kic/ssh_tunnel.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2020 The Kubernetes Authors All rights reserved.
+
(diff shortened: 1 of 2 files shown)


pkg/minikube/tunnel/kic/ssh_tunnel.go
```diff
@@ -118,7 +118,7 @@ func (t *SSHTunnel) stopMarkedConnections() {
 
 // sshConnName creates a uniq name for the tunnel, using its name/clusterIP/ports.
 // This allows a new process to be created if an existing service was changed,
-// the new process will support the IP/Ports change ocurred.
+// the new process will support the IP/Ports change occurred.
```
medyagh: also fix the boiler plate

medyagh:
> Fixes typo caused travis and gh action failed.
> 
> @medyagh I feel like we should enable the `required` check for at least (build and lint gh action) to avoid such occurrence later. There might be the case that golangci-lint misses some rule, however, it still helps.

I agree we should enable required for lint and build ! you are right.
lets create an issue for it and make that happen.

pkg/minikube/tunnel/kic/ssh_tunnel.go
medyagh: run this script: 
/hack/boilerplate/fix.sh

it will add right boilerplate

afbjorklund:
> @medyagh I feel like we should enable the `required` check for at least (build and lint gh action) to avoid such occurrence later. There might be the case that golangci-lint misses some rule, however, it still helps.

> I agree we should enable required for lint and build ! you are right.
lets create an issue for it and make that happen.

`go lint` is quite clear about that it should **not** be blocking the build, but other parts of golangci-lint (such as `go fmt` and `go vet`) probably should block. Not so sure about blocking on typos, though...

The `misspell` test did catch it, though:

```
misspell -error cmd pkg test
pkg/minikube/tunnel/kic/ssh_tunnel.go:110:52: "ocurred" is a misspelling of "occurred"
```

Issue #6664, opened 2020-02-17
Upgrade CNI to support version 0.4.0


afbjorklund:
`ERRO[0000] Error adding network: incompatible CNI versions; config is "0.4.0", plugin supports ["0.1.0" "0.2.0" "0.3.0" "0.3.1"] `

We are running 0.6.0 (Aug 2017), should upgrade to 0.7.0 (Apr 2019) - or 0.7.1

And we probably have to build it from source, in order to do so (no more binaries)

Required for `podman run`

afbjorklund:
Related to #5822

afbjorklund:
* https://github.com/containernetworking/cni/releases/tag/v0.7.1

* https://github.com/containernetworking/plugins/releases/tag/v0.8.5

afbjorklund:
If you want to run `podman-remote run` _before_ this upgrade has landed, you can use `--network=none` or `--network=host` to avoid failing on the missing CNI default network (i.e.`/etc/cni/net.d/87-podman-bridge.conflist`, the one that was failing with the error above)
Normally we only use `podman-remote build` and `podman load`, but it should be available...

afbjorklund:
Not worth holding the release for, unless something else (than podman) requires newer CRI ?

PR #6784, opened 2020-02-24
Upgrade cni and cni-plugins to spec 0.4.0

afbjorklund:
Needs testing...

For #6664


afbjorklund:
@medyagh : should we make a beta ISO, including this CNI upgrade ?

Issue #6801, opened 2020-02-25
docker driver should display cpu and ram availabe to docker-machine not system


medyagh:
docker system info --format '{{json .NCPU}}'
docker system info --format '{{json .MemTotal}}'

afbjorklund:
Makes sense, we added a similar feature (but backwards) to show the localhost information for `none` instead of the VM configuration...

PR #6803, opened 2020-02-26
Display more meaningful system info for kic Drivers

medyagh:
## After this PR: 
it will show the actual memory size from the docker-machine (as opposed to the user system) and how much is available to them.
in this case, my docker-machine has less than 2GB but it used to show 16 GB

```
😄  minikube v1.7.3 on Darwin 10.13.6
✨  Using the docker (experimental) driver based on user configuration
🔥  Creating Kubernetes in docker container with (CPUs=2) (4 available), Memory=2000MB (1989MB available) ...
```

## before this PR 
the info before this PR was useless because it was showing ram on the host machine not the docker-machine.that user can decide to allocate limtited resource to it. so their machine might have 16 GB ram but their docker-machine only have 2 GB

```
medmac@~/minikube (d_cpu_ram) $ minikube start --vm-driver=docker
😄  minikube v1.7.3 on Darwin 10.13.6
✨  Using the docker (experimental) driver based on user configuration
🔥  Creating Kubernetes in docker container with (CPUs=2), Memory=2000MB (16384MB available) ...
```

Closes https://github.com/kubernetes/minikube/issues/6801
closes https://github.com/kubernetes/minikube/issues/6801

### DIFF
--- a/pkg/drivers/kic/oci/info.go
+++ b/pkg/drivers/kic/oci/info.go
@@ -209,7 +209,7 @@ func DaemonInfo(ociBin string) (SysInfo, error) {
 func dockerSystemInfo() (SysInfo, error) {
 	var ds dockerSysInfo
 	var s SysInfo
-	cmd := exec.Command(Docker, "system", "info", "--format", "'{{json .}}'")
+	cmd := exec.Command(Docker, "system", "info", "--format", "{{json .}}")
 	out, err := cmd.CombinedOutput()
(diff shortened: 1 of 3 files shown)


medyagh:
> Shouldn’t this be in cruntime?
> 
> Or more likely a similar library
> (instead of in the driver wrapper)

it is a tricky one, since Podman is not a run time but this implements for both docker and podman

PR #6806, opened 2020-02-26
Disable default hypervisors mounts + fix 9P path on Darwin 

### LINKED ISSUE #6788: dangerous default behaviour (mounting home directory inside minikube VM)
Hello,
When starting a minikube instance 
`minikube start` 
 by default, the `--mount-string=` is set to home directory   `~` of the host machine, this argument is passed to the minikube mount command on start, the home directory is mounted under `/minikube-host` (mode -493 =>).
executing `rm -rf /` inside the minikube VM (or with ssh) would have a catastrophic effect (~ directory of the host machine deleted).
Since minikube is made for testing purposes, executing deletion command is highly probable.
I think this default behavior is very dangerous and should be changed.
Regards 
UPDATE:
This behavior is related  to VirtualBox's driver, not to minikube's mount, disabling the fs mounts provided by the hypervisors should fix this behavior (by setting disable-driver-mounts to true)
thanks to @afbjorklund for his remarks

starbassma:
Hello,

this PR : 

- Fix default 9P mount path on Darwin ($HOME instead of /Users, same path on all OS) 
- Add the default 9P mount path for Xhyve Hypervisor (doc) 
- Disable default mounts provided by the hypervisors (lax behavior, can lead to data loss on the host machine)

check 
fixes #6788

Regards,

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -190,7 +190,7 @@ func initKubernetesFlags() {
 // initDriverFlags inits the commandline flags for vm drivers
 func initDriverFlags() {
 	startCmd.Flags().String("vm-driver", "", fmt.Sprintf("Driver is one of: %v (defaults to auto-detect)", driver.DisplaySupportedDrivers()))
-	startCmd.Flags().Bool(disableDriverMounts, false, "Disables the filesystem mounts provided by the hypervisors")
+	startCmd.Flags().Bool(disableDriverMounts, true, "Disables the filesystem mounts provided by the hypervisors")
 
 	// kvm2
 	startCmd.Flags().String(kvmNetwork, "default", "The KVM network name. (kvm2 driver only)")
--- a/pkg/minikube/constants/constants_darwin.go
+++ b/pkg/minikube/constants/constants_darwin.go
@@ -22,5 +22,5 @@ import (
 	"k8s.io/client-go/util/homedir"
 )
 
-// DefaultMountDir is the default mount directory for Darwin
+// DefaultMountDir is the default mount directory
 var DefaultMountDir = homedir.HomeDir()
--- a/pkg/minikube/constants/constants_darwin.go
+++ b/pkg/minikube/constants/constants_darwin.go
@@ -18,5 +18,9 @@ limitations under the License.
 
 package constants
 
-// DefaultMountDir is the default mounting directory for Darwin
-var DefaultMountDir = "/Users"
+import (
+	"k8s.io/client-go/util/homedir"
+)
+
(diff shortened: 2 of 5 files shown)


starbassma:
> mkcmp

 seems unrelated to the modifications

priyawadhwa:
@starbassma yah you can ignore mkcmp errors here

cmd/minikube/cmd/start.go
```diff
@@ -190,7 +190,7 @@ func initKubernetesFlags() {
 // initDriverFlags inits the commandline flags for vm drivers
 func initDriverFlags() {
 	startCmd.Flags().String("vm-driver", "", fmt.Sprintf("Driver is one of: %v (defaults to auto-detect)", driver.DisplaySupportedDrivers()))
-	startCmd.Flags().Bool(disableDriverMounts, false, "Disables the filesystem mounts provided by the hypervisors")
+	startCmd.Flags().Bool(disableDriverMounts, true, "Disables the filesystem mounts provided by the hypervisors")
```
medyagh: double negative sin the flags are tricky,

pkg/minikube/constants/constants_darwin.go
```diff
@@ -18,5 +18,9 @@ limitations under the License.
 
 package constants
 
-// DefaultMountDir is the default mounting directory for Darwin
-var DefaultMountDir = "/Users"
+import (
+	"k8s.io/client-go/util/homedir"
+)
+
+// DefaultMountDir is the default mount directory for Darwin
```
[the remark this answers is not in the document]
starbassma: done

cmd/minikube/cmd/start.go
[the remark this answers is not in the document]
starbassma: No, it won't, certificates are copied through the api

starbassma: I agree, should we consider renaming this flag ?

site/content/en/docs/Tasks/mount.md
```diff
@@ -71,10 +71,17 @@ Some hypervisors, have built-in host folder sharing. Driver mounts are reliable
 | VirtualBox | macOS | /Users | /Users |
 | VirtualBox | Windows | C://Users | /c/Users |
 | VMware Fusion | macOS | /Users | /Users |
+| Xhyve | macOS | /Users | /Users |
```
afbjorklund: The xhyve driver is not supported anymore

cmd/minikube/cmd/start.go
medyagh: yes I would say lets rename this flag !

tstromberg:
I agree with this PR - but will close until it can be rebased.

PR #6807, opened 2020-02-26
warn about "stuck docker"  in minikube delete

medyagh:
### Before PR:
delete would hang forever...

because docker container was stuck (however docker itself is healthy and no signs of docker problem but in reality docker needs restart, only way to identify this issue is to do docker inspect on the container and it would hang forever ... (no error) and other docker commands like info or system info were just acting normal)

### after this PR:
Fail Fast. Warn User to restart docker.
nothing should hang forever.
```
E0226 12:34:52.650785   31788 oci.go:61] docker daemon seems to be stuck. Please try restarting your docker.
💀  Removed all traces of the "minikube" cluster.```

### DIFF
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -59,13 +59,13 @@ func DeleteContainersByLabel(ociBin string, label string) []error {
 		if err != nil {
 			deleteErrs = append(deleteErrs, errors.Wrapf(err, "delete container %s: %s daemon is stuck. please try again!", c, ociBin))
 			glog.Errorf("%s daemon seems to be stuck. Please try restarting your %s.", ociBin, ociBin)
-		} else {
-			cmd := exec.Command(ociBin, "rm", "-f", "-v", c)
-			if out, err := cmd.CombinedOutput(); err != nil {
-				deleteErrs = append(deleteErrs, errors.Wrapf(err, "delete container %s: output %s", c, out))
-			}
-
+			continue
 		}
+		cmd := exec.Command(ociBin, "rm", "-f", "-v", c)
+		if out, err := cmd.CombinedOutput(); err != nil {
+			deleteErrs = append(deleteErrs, errors.Wrapf(err, "delete container %s: output %s", c, out))
+		}
+
 	}
 	return deleteErrs
 }
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -17,9 +17,11 @@ limitations under the License.
 package oci
 
 import (
+	"context"
 	"os"
 	"path/filepath"
 	"strconv"
+	"time"
 
 	"bufio"
 	"bytes"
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: pkg/drivers/kic/oci/oci.go [hunk 1]
```diff
@@ -51,9 +53,18 @@ func DeleteAllContainersByLabel(ociBin string, label string) []error {
 		return nil
 	}
 	for _, c := range cs {
-		cmd := exec.Command(ociBin, "rm", "-f", "-v", c)
```
tstromberg: Unrelated to this PR, but I think this function should be named `DeleteContainersByLabel`. I don't understand what the importance of `All` in the name is.

pkg/drivers/kic/oci/oci.go
```diff
@@ -51,9 +53,18 @@ func DeleteAllContainersByLabel(ociBin string, label string) []error {
...
-		if out, err := cmd.CombinedOutput(); err != nil {
-			deleteErrs = append(deleteErrs, errors.Wrapf(err, "delete container %s: output %s", c, out))
+		_, err := containerStatus(ociBin, c)
+		// only try to delete if docker/podman inspect returns
+		// if it doesn't it means docker-daemon is stuck and needs restart
+		if err != nil {
+			deleteErrs = append(deleteErrs, errors.Wrapf(err, "delete container %s: -daemon is stuck.", c, ociBin))
+			glog.Errorf("%s-daemon seems to be stuck. Please try restarting your %s.", ociBin, ociBin)
```
tstromberg: Won't this result in duplicate errors being sent to the console? In general, you either log an error, or return it, but not both.

pkg/drivers/kic/oci/oci.go
```diff
@@ -51,9 +53,18 @@ func DeleteAllContainersByLabel(ociBin string, label string) []error {
 		return nil
 	}
 	for _, c := range cs {
-		cmd := exec.Command(ociBin, "rm", "-f", "-v", c)
-		if out, err := cmd.CombinedOutput(); err != nil {
-			deleteErrs = append(deleteErrs, errors.Wrapf(err, "delete container %s: output %s", c, out))
+		_, err := containerStatus(ociBin, c)
```
tstromberg: Is there a command available to detect this that does not require querying the status of an individual container?

It seems like it would be better to detect a stuck daemon in a general fashion (like `docker status` or some other command), rather than one that could be impacted by a container not existing.

pkg/drivers/kic/oci/oci.go
tstromberg: `docker-daemon` isn't a thing. Perhaps this should say something like:

`Your %s daemon appears to be stuck. Please try restarting %s.`

pkg/drivers/kic/oci/oci.go
```diff
@@ -51,9 +53,18 @@ func DeleteAllContainersByLabel(ociBin string, label string) []error {
...
+		// only try to delete if docker/podman inspect returns
+		// if it doesn't it means docker-daemon is stuck and needs restart
+		if err != nil {
+			deleteErrs = append(deleteErrs, errors.Wrapf(err, "delete container %s: -daemon is stuck.", c, ociBin))
+			glog.Errorf("%s-daemon seems to be stuck. Please try restarting your %s.", ociBin, ociBin)
+		} else {
```
tstromberg: call `continue` instead of else, so that the happy path is the left-most.

pkg/drivers/kic/oci/oci.go
medyagh: no because we dont log in parent

pkg/drivers/kic/oci/oci.go
medyagh: unfortunately that is not the case ! docker status works great ! and it even lets your create new containers

and everything is normal ! only this container is stuck and u can not do anything to it

that was only way I could detect it.

pkg/drivers/kic/oci/oci.go
medyagh: I think this is complicated be because after this we have purge ... I rather not touch this in this PR

medyagh: but I agree I dont love this nested if from previous PRs we merged

medyagh: ah I just undresttood what u meant. I will push a commit

Issue #6778, opened 2020-02-24
implement service command for docker on mac os


medyagh:
I think we could use the same SSH connection that was added by @josedonizetti for making tunnel work for service, with one difference that this could be multiple service at random ports and not needed to make the ports map to same port.

PR #6811, opened 2020-02-26
kic on mac: add service cmd support

josedonizetti:
Fix for https://github.com/kubernetes/minikube/issues/6778

### DIFF
--- a/cmd/minikube/cmd/service.go
+++ b/cmd/minikube/cmd/service.go
@@ -175,6 +175,8 @@ func openURLs(svc string, urls []string) {
 		if err := browser.OpenURL(u); err != nil {
 			exit.WithError(fmt.Sprintf("open url failed: %s", u), err)
 		}
+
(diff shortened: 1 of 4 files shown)


medyagh:
I tried it locally on mac, and it works

```
$ ./out/minikube service hello-minikube
starting tunnel for hello-minikube
|-----------|----------------|-------------|-----------------------|
| NAMESPACE |      NAME      | TARGET PORT |          URL          |
|-----------|----------------|-------------|-----------------------|
| default   | hello-minikube |             | http://127.0.0.1:8080 |
|-----------|----------------|-------------|-----------------------|
🎉  Opening service default/hello-minikube in default browser...

```

one suggestion 👍 
where it says
🎉  Opening service default/hello-minikube in default browser...
we could add another line

```
(INFOicon) Because you are using docker driver on Mac, the terminal needs to be open
to run it in background use the following command ...


```

PR #6816, opened 2020-02-27
give hint if 'minikube delete' cannot delete the machine directory

halmai:
In some cases the command `minikube delete` can produce error message like this on Windows:

```
! Unable to get the status of the minikube cluster.
* Removing C:\Users\...\.minikube\machines\minikube ...
*
X Unable to remove machine directory: %v: remove C:\Users\...\.minikube\machines\minikube\disk.vhd: The process cannot access the file because it is being used by another process.
*
  - https://github.com/kubernetes/minikube/issues/new/choose
```

A possible explanation for this is that the Hyper-V Manager keeps some files locked which prevents the directory from being deleted. 

This Pull Request aims to give a more reasonable explanation if such a situation happens.

### DIFF
--- a/pkg/minikube/problem/err_map.go
+++ b/pkg/minikube/problem/err_map.go
@@ -111,7 +111,7 @@ var vmProblems = map[string]match{
 	},
 	"HYPERV_FILE_DELETE_FAILURE": {
 		Regexp: re(`Unable to remove machine directory`),
-		Advice: "Stop the  Hyper-V Manager and run 'minikube delete'.",
+		Advice: "You may need to stop the  Hyper-V Manager and run `minikube delete` again.",
 		Issues: []int{6804},
 		GOOS:   []string{"windows"},
 	},
--- a/pkg/minikube/problem/err_map.go
+++ b/pkg/minikube/problem/err_map.go
@@ -109,6 +109,12 @@ var vmProblems = map[string]match{
 		Issues: []int{4511},
 		GOOS:   []string{"windows"},
 	},
+	"HYPERV_FILE_DELETE_FAILURE": {
+		Regexp: re(`Unable to remove machine directory`),
+		Advice: "Stop the  Hyper-V Manager and and run 'minikube delete'.",
+		Issues: []int{6804},
+		GOOS:   []string{"windows"},
+	},
 
 	// KVM
 	"KVM2_NOT_FOUND": {
--- a/pkg/minikube/problem/err_map.go
+++ b/pkg/minikube/problem/err_map.go
@@ -111,7 +111,7 @@ var vmProblems = map[string]match{
 	},
 	"HYPERV_FILE_DELETE_FAILURE": {
 		Regexp: re(`Unable to remove machine directory`),
-		Advice: "You may need to stop the  Hyper-V Manager and run `minikube delete` again.",
+		Advice: "You may need to stop the Hyper-V Manager and run `minikube delete` again.",
 		Issues: []int{6804},
 		GOOS:   []string{"windows"},
 	},
--- a/pkg/minikube/problem/err_map.go
+++ b/pkg/minikube/problem/err_map.go
@@ -111,7 +111,7 @@ var vmProblems = map[string]match{
 	},
 	"HYPERV_FILE_DELETE_FAILURE": {
 		Regexp: re(`Unable to remove machine directory`),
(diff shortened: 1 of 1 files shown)


halmai:
I signed it

halmai:
I signed it

pkg/minikube/problem/err_map.go
```diff
@@ -109,6 +109,12 @@ var vmProblems = map[string]match{
 		Issues: []int{4511},
 		GOOS:   []string{"windows"},
 	},
+	"HYPERV_FILE_DELETE_FAILURE": {
+		Regexp: re(`Unable to remove machine directory`),
+		Advice: "Stop the  Hyper-V Manager and and run 'minikube delete'.",
```
josedonizetti: `and and` can you fix the typo pls?

pkg/minikube/problem/err_map.go
```diff
@@ -109,6 +109,12 @@ var vmProblems = map[string]match{
 		Issues: []int{4511},
 		GOOS:   []string{"windows"},
 	},
+	"HYPERV_FILE_DELETE_FAILURE": {
+		Regexp: re(`Unable to remove machine directory`),
```
josedonizetti: wonder if we can make the regex more specific? wdyt?

josedonizetti:
@halmai Thank you for the PR. Left a couple comments.

pkg/minikube/problem/err_map.go
halmai: Oops, sorry. I fixed it.

pkg/minikube/problem/err_map.go
halmai: The complete error string looks like this:

```
Unable to remove machine directory: %v: remove C:\Users\...\.minikube\machines\minikube\disk.vhd: The process cannot access the file because it is being used by another process.
```

The path part is changing a lot, depending on the user's installation. We could check for an arbitrary path which starts with a capital letter and a colon, but after that, almost anything can come, depending on the minikube installation path. I am not saying many users would install their minikube onto the path `Y:\HardcoreUsers\i\ am\ a\ hacker\minikube\` or `Z:\Segmentation fault\minikube`, I just think we cannot check it in a bullet-proof way.

The last part of the error message (`The process cannot access the file because it is being used by another process.`) is also varying a lot, depending on the user's environment. It is sent by the operating system. In case of an English Win10, we can assume this string but it can be a Portugal, Hungarian, Korean, whatever Win10. The exact string on these languages are very hard to check. I wouldn't like to go on this path.

I think it would be better to re-phrase the message in a more flexible, allowing form, like `Maybe you need to stop the  Hyper-V Manager and run 'minikube delete'.` It is less assertive than the current `Stop the  Hyper-V Manager and run 'minikube delete'.`. 

What do you think?

josedonizetti: Thanks for explaining. I agree that re-phrasing would be better.

halmai: I pushed the newly phrased advice.

medyagh: @halmai 

while it is good to have a general fall back to tell user something about this failure, 
how about we handle it more gracefully in the code rather than wait till exit ?

is there any way we could handle this error before it comes to exist ?

or at least could we make the regex a bit more explicit ?

halmai: @medyagh When the user invokes the `minikube delete` command, they obviously want to delete it. But it won't be possible until the program that locks the files, releases them, or exits. This won't happen if we are just waiting. That program needs to be stopped manually by the user. Our code cannot do ot. I cannot see what other way of graceful handling could be done here. Do you have any concrete idea?

Related to the more explicit regexp, please read my comments above. It starts with 
> The complete error string looks like this:
Based on the things I wrote there, I don't know how we could be more specific. However, I am happy to consider any proposal. :)

pkg/minikube/problem/err_map.go
```diff
@@ -109,6 +109,12 @@ var vmProblems = map[string]match{
...
+		Advice: "You may need to stop the  Hyper-V Manager and run `minikube delete` again.",
```
tstromberg: Remove the extra space before Hyper-V

halmai: Thanks, I did it.

tstromberg:
Thank you!

Issue #6832, opened 2020-02-28
kic: fix tunnel can't bind to ports for docker on OSX 


josedonizetti:
The existing integration test for tunnel are not passing for the kic/docker on mac because the test create a service on port 80, which is unprivileged, we need sudo to be able to create a ssh tunnel for it.

Considering that tunnel asks for `sudo` for all other drivers, we can do the same for kic/docker on osx.

PR #6833, opened 2020-02-28
kic: fix unprivileged port bind tunnel docker for mac

josedonizetti:
Fix for https://github.com/kubernetes/minikube/issues/6832

If any service exposes privileged ports, we print a message informing the user which service/port and ask sudo permission.

```
minikube tunnel
The service nginx-svc requires priviledged ports to be exposed: [80]
sudo permission will be asked for it.
```

This PR also fixes the functional test for tunnel, it now passes for docker on osx too. https://github.com/kubernetes/minikube/blob/master/test/integration/fn_tunnel_cmd.go


medyagh:
looks good thank you for this !

PR #6835, opened 2020-02-28
doc: minikube tunnel command runs as a process in the same terminal

kschaab:
The documentation suggests that minikube tunnel runs in a separate terminal window and as a daemon, but actually it is just a process that requires root and runs until terminated.

### DIFF
--- a/site/content/en/docs/Tasks/loadbalancer.md
+++ b/site/content/en/docs/Tasks/loadbalancer.md
@@ -14,7 +14,7 @@ A LoadBalancer service is the standard way to expose a service to the internet. 
 
 ## Using `minikube tunnel`
 
-Services of type `LoadBalancer` can be exposed via the `minikube tunnel` command. It will run in a separate terminal until Ctrl-C is hit.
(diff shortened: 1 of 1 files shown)


site/content/en/docs/Tasks/loadbalancer.md
```diff
@@ -14,7 +14,7 @@ A LoadBalancer service is the standard way to expose a service to the internet.
 
 ## Using `minikube tunnel`
 
-Services of type `LoadBalancer` can be exposed via the `minikube tunnel` command. It will run in a separate terminal until Ctrl-C is hit.
+Services of type `LoadBalancer` can be exposed via the `minikube tunnel` command. It will run in the terminal until Ctrl-C is hit.
```
josedonizetti: I agree the wording before wasn't clear. But I'm not sure we are properly explaining now as well, the idea here is that the user must leave the tunnel running in a different terminal to have the LoadBalancer services exposed. WDYT?

medyagh: I think word separate was meant to make sure user know they need two terminals for this.

kschaab: Better?

josedonizetti:
Thanks!

PR #6787, opened 2020-02-24
Implement multi-node cluster capabilities

sharifelgamal:
Fixes #6505 
Begins to address #94 

You can spin up a multinode cluster in 2 ways:
```
minikube start --nodes=2
```

```
minikube start
minikube node add --name=node2
```

You can spin up and down individual nodes by name:
```
minikube node stop node2
minikube node start node2
```

Multi-node clusters are still experimental and won't be fully featured.

NOTE: This PR does not account for CNI, that will come in a subsequent PR. Currently, you will need to install CNI manually for multi-node to work properly.

PR #6836, opened 2020-02-28
Decouple machine name from profile/cluster name

sharifelgamal:
Currently we assume that the name of the VM host is the same as the name of our minikube cluster, but this needs to change with multi-node so this refactor attempts to do that with a new `driver.MachineName` function.


cmd/minikube/cmd/dashboard.go
```diff
@@ -80,7 +82,13 @@ var dashboardCmd = &cobra.Command{
 			exit.WithError("Error getting client", err)
 		}
 
-		if _, err = api.Load(cc.Name); err != nil {
+		cp, err := config.PrimaryControlPlane(*cc)
+		if err != nil {
+			exit.WithError("Error getting primary control plane", err)
+		}
+
+		machineName := driver.MachineName(cc.Name, cp.Name)
```
tstromberg: I feel like there is some room here for an API which is something like `cluster.Exists(profileName)`, `cluster.IsRunning(profileName)`, `cluster.IsHealthy(profileName)`. 

The dashboard command shouldn't need to know or care about individual machines. This abstraction is not going to scale if/when we allow multiple control plane VM's.

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -75,7 +75,7 @@ func NewBootstrapper(api libmachine.API) (*Bootstrapper, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "command runner")
 	}
-	return &Bootstrapper{c: runner, contextName: name, k8sClient: nil}, nil
+	return &Bootstrapper{c: runner, contextName: viper.GetString(config.MachineProfile), k8sClient: nil}, nil
```
tstromberg: contextName feels like it shouldn't map to a machine profile - whatever that is.

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -65,8 +65,8 @@ type Bootstrapper struct {
 }
 
 // NewBootstrapper creates a new kubeadm.Bootstrapper
-func NewBootstrapper(api libmachine.API) (*Bootstrapper, error) {
-	name := viper.GetString(config.MachineProfile)
+func NewBootstrapper(api libmachine.API, nodeName string) (*Bootstrapper, error) {
```
tstromberg: I'm a bit confused: what and why is nodeName here? 

It feels like the bootstrapper shouldn't need to know about individual nodes names. It's unclear to me which one I would pass here.

pkg/minikube/constants/constants.go
```diff
@@ -34,7 +34,9 @@ const (
 	// OldestKubernetesVersion is the oldest Kubernetes version to test against
 	OldestKubernetesVersion = "v1.11.10"
 	// DefaultMachineName is the default name for the VM
-	DefaultMachineName = "minikube"
+	DefaultMachineName = "minikube-minikube"
```
tstromberg: I don't think DefaultMachineName should exist, but if it must - can we pick a name that isn't so repetitive? 

I feel like may just cause problems, and that machine names should be dynamically generated with something like  `minikube-<profile>-<index>`. I realize that this is effectively repetitive given our current default, but I also think we should consider renaming our default profile to `default` to reduce repetition at some point as well.

pkg/minikube/cluster/cluster.go
```diff
@@ -42,12 +42,12 @@ func init() {
 }
 
 // Bootstrapper returns a new bootstrapper for the cluster
-func Bootstrapper(api libmachine.API, bootstrapperName string) (bootstrapper.Bootstrapper, error) {
+func Bootstrapper(api libmachine.API, bootstrapperName string, nodeName string) (bootstrapper.Bootstrapper, error) {
```
tstromberg: This feels like an abstraction violation: `cluster.Bootstrapper()` should not require a node be passed to it. It doesn't make sense which node one would select.

tstromberg:
Given the size of this PR, there wasn't much to comment on. Nice work unravelling the spaghetti!

medyagh:
plz make sure all the integration tests pass: 
```
[
  "TestDownloadOnly/group/v1.11.10",
  "TestDownloadOnly/group/v1.17.3",
  "TestDownloadOnly/group/v1.17.3#01",
  "TestDownloadOnly/group/DeleteAll",
  "TestDownloadOnly/group/DeleteAlwaysSucceeds",
  "TestOffline/group/docker",
  "TestAddons",
  "TestFunctional/serial/StartWithProxy",
  "TestFunctional/serial/KubeContext",
  "TestFunctional/serial/KubectlGetPods",
  "TestFunctional/serial/MinikubeKubectlCmd",
  "TestFunctional/parallel/ComponentHealth",
  "TestFunctional/parallel/ConfigCmd",
  "TestFunctional/parallel/DashboardCmd",
  "TestFunctional/parallel/DNS",
  "TestFunctional/parallel/DryRun",
  "TestFunctional/parallel/StatusCmd",
  "TestFunctional/parallel/LogsCmd",
  "TestFunctional/parallel/ProfileCmd/profile_not_create",
  "TestFunctional/parallel/ProfileCmd/profile_list",
  "TestFunctional/parallel/ProfileCmd/profile_json_output",
  "TestFunctional/parallel/ServiceCmd",
  "TestFunctional/parallel/AddonsCmd",
```

pkg/minikube/cluster/cluster.go
sharifelgamal: You're absolutely right, I'm just using Bootstrapper to get at the command runner, which is per machine instead of per cluster. I think the solution would be to extract command runner out of bootstrapper, which would likely be a large enough change for its own PR.

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
sharifelgamal: our kubectl context name has always mapped one to one with the name of the cluster. constants.MachineProfile, bad as the name is, is just the name of profile passed into whatever minikube command is run. that's the reason I'm using that directly.

pkg/minikube/cluster/cluster.go
tstromberg: Hmm, Could you use `machine.Runner` instead?

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
tstromberg: OK. Seems reasonable - kubeadm bootstrapper uses contextName to find the apiserver to talk to.

pkg/minikube/constants/constants.go
sharifelgamal: I suspect we can just get rid of it. The MachineName func should be able to be invoked anywhere we currently use this.

sharifelgamal: and yes, we are planning the machine name to be exactly of that format, I just haven't made the change on this PR yet,

tstromberg: Just to follow-up: I think DefaultMachineName and DefaultNodeName should be removed. I think you are leaning toward this already, but my gut instinct is that:

* Seen from the VM, nodes should be named something simple like `minikube01`, `mini01` or `m01`. 
* Seen from the  hypervisor, nodes should be named `<cluster>-minikube01`, `<cluster>-mini01`, or `<cluster>-m01`, or `<cluster>-01`. 

Probably the least confusing is a scenario where the node sees themself as `m01`, and the hypervisor sees nodes as `<cluster>-m01`. 



Nodes should just be named <cluster>-<index>

sharifelgamal: This definitely makes the most sense. I had picked the VM name and node name being the same for simplicity, but this would be much clearer.

pkg/minikube/cluster/cluster.go
sharifelgamal: I suspect fixing this will involve moving command runner out of the bootstrapper entirely and changing how the bootstrapper interface works entirely.

cmd/minikube/cmd/dashboard.go
sharifelgamal: This is correct, and I'm going to fix it in #6787

medyagh:
test results seem much better now  ! thank you for this progress, do you feel like you are ready to invoke ok to test on it for jenkin tests?

pkg/drivers/kic/oci/types.go
```diff
@@ -35,7 +37,8 @@ const (
 type CreateParams struct {
 	Name          string            // used for container name and hostname
 	Image         string            // container image to use to create the node.
-	ClusterLabel  string            // label the containers we create using minikube so we can clean up
+	ClusterLabel  string            // label the clusters we create using minikube so we can clean up
+	NodeLabel     string            //l abel the nodes so we can clean up by node name
```
tstromberg: Fix comment spacing.

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -65,8 +64,8 @@ type Bootstrapper struct {
 }
 
 // NewBootstrapper creates a new kubeadm.Bootstrapper
-func NewBootstrapper(api libmachine.API) (*Bootstrapper, error) {
-	name := viper.GetString(config.MachineProfile)
+func NewBootstrapper(api libmachine.API, cc config.ClusterConfig, n config.Node) (*Bootstrapper, error) {
```
tstromberg: Please add a TODO(#xxx) to remove node as an argument, and/or clarify in the docstring what the node is doing here.

pkg/minikube/cluster/cluster.go
```diff
@@ -42,12 +43,12 @@ func init() {
 }
 
 // Bootstrapper returns a new bootstrapper for the cluster
-func Bootstrapper(api libmachine.API, bootstrapperName string) (bootstrapper.Bootstrapper, error) {
+func Bootstrapper(api libmachine.API, bootstrapperName string, cc config.ClusterConfig, n config.Node) (bootstrapper.Bootstrapper, error) {
```
tstromberg: Please add a TODO(#xxx) to remove node as an argument, and/or clarify in the docstring what the node is doing here.

pkg/minikube/constants/constants.go
```diff
@@ -33,11 +33,8 @@ const (
 	NewestKubernetesVersion = "v1.17.3"
 	// OldestKubernetesVersion is the oldest Kubernetes version to test against
 	OldestKubernetesVersion = "v1.11.10"
-	// DefaultMachineName is the default name for the VM
-	DefaultMachineName = "minikube"
-	// DefaultNodeName is the default name for the kubeadm node within the VM
```
tstromberg: Yay!

pkg/minikube/machine/cluster_test.go
```diff
@@ -60,11 +60,17 @@ func RegisterMockDriver(t *testing.T) {
 	}
 }
 
+var defaultNodeConfig = config.Node{
+	Name: viper.GetString("profile"),
```
tstromberg: Do you mind inlining this? I don't find the extra level of indirection to be useful for test readability.  I prefer test inputs be absolutely explicit, and this particular  config is small. 

`config.Node{Name: "minikube"}`

I would prefer it if the tests did not reach into global variables like viper flags unless absolutely necessary.

pkg/minikube/driver/driver.go
```diff
@@ -212,3 +213,12 @@ func SetLibvirtURI(v string) {
 	os.Setenv("LIBVIRT_DEFAULT_URI", v)
 
 }
+
+//MachineName returns the name of the machine given the cluster and node names
```
tstromberg: Add a space after "//".

Also, it may be worth clarifying here that driver.MachineName returns the machine name here is "as seen by the hypervisor". 

I would think other users who don't care about the hypervisor view would use Node.Name directly.

PR #6841, opened 2020-02-29
Negotiate Docker API Version before image query

afbjorklund:
Avoids error working with older Docker daemons
    Error response from daemon:
        client version 1.41 is too new.
        Maximum supported API version is 1.39

### DIFF
--- a/pkg/minikube/image/image.go
+++ b/pkg/minikube/image/image.go
@@ -44,6 +44,7 @@ var defaultPlatform = v1.Platform{
 func DigestByDockerLib(imgClient *client.Client, imgName string) string {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
 	defer cancel()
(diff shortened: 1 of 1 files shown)


afbjorklund:
The docker version was bumped (for some reason) in b911e8f038ea6c7f45f6fdb83ab0b88c1c430b67

medyagh:
This is interesting I wonder if will help with images with old manifest

afbjorklund:
This is not so much about the image, it is just that the library is using docker as a lookup mechanism to find the image id. It claims that this is faster than using go-containerregistry, but I wonder how this code is working when not using docker ? Seems to be querying docker for if it exists in e.g. cri-o ?

afbjorklund:
The errors were seen when running on Raspbian GNU/Linux 10 (buster)

The default version is 18.09.1: https://packages.debian.org/buster/docker.io

PR #6847, opened 2020-03-01
kic: improve error message for service not found

josedonizetti:
currently: 

```
minikube service test

💣  Error opening service: getting service: services "test" not found

```

with the PR:

```
minikube service test

💣  Error opening service: Service test was not found in "default" namespace. You may select another namespace by using 'minikube service test -n <namespace>: services "test" not found

```


medyagh:
How about we don't make ppl to create issue ?
If service doesn't exist let's not tell them minikube crashed ....



minikube is exiting due to an error. If the above message is not useful, open an issue:

josedonizetti:
@medyagh  can we tackle the message change in a separate PR? Because we also need to change the existing service, not only for kic. WDYT?

medyagh:
> @medyagh can we tackle the message change in a separate PR? Because we also need to change the service for other drivers, not only for kic. WDYT?

sounds good !

PR #6858, opened 2020-03-02
Fix native-ssh flag for the ssh command

### LINKED ISSUE #6852: Not possible to use external ssh client
For some reason, `minikube ssh` is not respecting the `--native-ssh` parameter.

```
Options:
      --native-ssh=true: Use native Golang SSH client (default true). Set to 'false' to use the command line 'ssh'
command when accessing the docker machine. Useful for the machine drivers when they will not start with 'Waiting for
SSH'.
```

But when running `minikube ssh --native-ssh=false`, it still uses the native client.

libmachine: Using SSH client type: native

Seems like something is wrong with the parsing of parameters, with viper/cobra ?

```go
                if viper.GetBool(nativeSSH) {
                        ssh.SetDefaultClient(ssh.Native)
                } else {
                        ssh.SetDefaultClient(ssh.External)
                }

```

afbjorklund:
Closes #6852

### DIFF
--- a/cmd/minikube/cmd/ssh.go
+++ b/cmd/minikube/cmd/ssh.go
@@ -30,6 +30,10 @@ import (
 	"k8s.io/minikube/pkg/minikube/out"
 )
 
+var (
+	nativeSSHClient bool
+)
+
 // sshCmd represents the docker-ssh command
 var sshCmd = &cobra.Command{
 	Use:   "ssh",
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/ssh.go
```diff
@@ -69,5 +73,5 @@ var sshCmd = &cobra.Command{
 }
 
 func init() {
-	sshCmd.Flags().Bool(nativeSSH, true, "Use native Golang SSH client (default true). Set to 'false' to use the command line 'ssh' command when accessing the docker machine. Useful for the machine drivers when they will not start with 'Waiting for SSH'.")
+	sshCmd.Flags().BoolVar(&nativeSSHClient, nativeSSH, true, "Use native Golang SSH client (default true). Set to 'false' to use the command line 'ssh' command when accessing the docker machine. Useful for the machine drivers when they will not start with 'Waiting for SSH'.")
```
medyagh: I am curious if we have other "bool" flags like this that could be fixed the same way ? (if not I am okay with merging this PR)

starbassma: No, this the only one.
Regards

PR #6870, opened 2020-03-03
Add k8s binaries to preloaded tarball

### LINKED ISSUE #6868: Copy in k8s binaries to preloaded tarball so we can skip transfer

priyawadhwa:
Include k8s binaries in preloaded tarball so that we can skip tranferring this over.

This increases the size of the tarball from around 430 MB to 500MB, but saves a couple seconds on start time.

Fixes #6868


medyagh:
Looks good to me, please feel free to merge.

PR #6874, opened 2020-03-03
Only cache k8s images if necessary

### LINKED ISSUE #6871: Don't download cached images if preloaded tarball exists

priyawadhwa:
We can skip caching k8s images if we are using the kic driver and a preloaded tarball for that k8s version exists.

Fixes #6871

### DIFF
--- a/pkg/minikube/preload/preload.go
+++ b/pkg/minikube/preload/preload.go
@@ -74,7 +74,7 @@ func TarballExists(k8sVersion, containerRuntime string) bool {
 		return false
 	}
 	url := remoteTarballURL(k8sVersion)
-	_, err := http.Get(url)
+	_, err := http.Head(url)
 	return err == nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/minikube/preload/preload.go
```diff
@@ -67,6 +67,17 @@ func remoteTarballURL(k8sVersion string) string {
...
 }
 
+// TarballExists returns true if there is a preloaded tarball
+// that can be used
+func TarballExists(k8sVersion, containerRuntime string) bool {
+	if containerRuntime != "docker" {
+		return false
+	}
+	url := remoteTarballURL(k8sVersion)
+	_, err := http.Get(url)
```
josedonizetti: Probably better to use `_, err := http.Head(url)` to avoid a the full `GET` return from the server. WDYT?

medyagh: I agree ! with @josedonizetti less network footbprint the better.

PR #6878, opened 2020-03-04
skip phases in cmd whitelist for kubeadm

ulm0:
Implements #6877


medyagh:
@ulm0 looks good thank you for this contribution and updating the docs as well !

Issue #6860, opened 2020-03-02
service list: info target port not being populated


medyagh:
```
|-----------|----------------|-------------|--------------------------|
| NAMESPACE |      NAME      | TARGET PORT |           URL            |
|-----------|----------------|-------------|--------------------------|
| default   | hello-minikube |             | http://192.168.9.2:30130 |
|-----------|----------------|-------------|--------------------------|

```

in this case the target port is 8080 from https://minikube.sigs.k8s.io/docs/examples/

starbassma:
Actually it show the name of the port (.spec.port.name) not the port

![Screenshot_20200303_005855](https://user-images.githubusercontent.com/6652482/75729433-16ee3000-5ceb-11ea-8ee1-e46a1659d59b.png)

the descriptor of the deployment :

![Screenshot_20200303_010730](https://user-images.githubusercontent.com/6652482/75729990-cdeaab80-5ceb-11ea-882c-4a9af76c42a4.png)

PR #6879, opened 2020-03-04
service list: populate target port

prasadkatti:
Populate the target port in `service list`

fixes: #6860 

Use port name for target port if provided. If not, use 'port'.
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#serviceport-v1-core

This is my first contribution to minikube. I am also fairly new to kubernetes. So please go easy on me. :)

### Before
image

### After
image

### DIFF
--- a/cmd/minikube/cmd/service_list.go
+++ b/cmd/minikube/cmd/service_list.go
@@ -66,6 +66,7 @@ var serviceListCmd = &cobra.Command{
 			if len(serviceURL.URLs) == 0 {
 				data = append(data, []string{serviceURL.Namespace, serviceURL.Name, "No node port"})
 			} else {
+				servicePortNames := strings.Join(serviceURL.PortNames, "\n")
 				serviceURLs := strings.Join(serviceURL.URLs, "\n")
 
 				// if we are running Docker on OSX we empty the internal service URLs
@@ -73,7 +74,7 @@ var serviceListCmd = &cobra.Command{
 					serviceURLs = ""
 				}
 
-				data = append(data, []string{serviceURL.Namespace, serviceURL.Name, "", serviceURLs})
+				data = append(data, []string{serviceURL.Namespace, serviceURL.Name, servicePortNames, serviceURLs})
 			}
 		}
 
--- a/pkg/minikube/service/service.go
+++ b/pkg/minikube/service/service.go
@@ -199,7 +199,7 @@ func printURLsForService(c typed_core.CoreV1Interface, ip, service, namespace st
 	for _, port := range svc.Spec.Ports {
 
 		if port.Name != "" {
-			m[port.TargetPort.IntVal] = port.Name
+			m[port.TargetPort.IntVal] = fmt.Sprintf("%s/%d", port.Name, port.Port)
 		} else {
 			m[port.TargetPort.IntVal] = strconv.Itoa(int(port.Port))
 		}
--- a/pkg/minikube/service/service.go
+++ b/pkg/minikube/service/service.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"net/url"
 	"os"
+	"strconv"
 	"strings"
 	"text/template"
 	"time"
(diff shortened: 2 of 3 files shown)


medyagh:
@prasadkatti can you please put before and after output in the PR description ?

pkg/minikube/service/service.go
```diff
@@ -196,6 +197,13 @@ func printURLsForService(c typed_core.CoreV1Interface, ip, service, namespace st
 	urls := []string{}
 	portNames := []string{}
 	for _, port := range svc.Spec.Ports {
+
+		if port.Name != "" {
+			m[port.TargetPort.IntVal] = port.Name
```
josedonizetti: I don't think the port name is useful by itself.  I'd rather change to print something like `name/port` if name is available, if not only `port`.

cmd/minikube/cmd/service_list.go
```diff
@@ -66,14 +66,15 @@ var serviceListCmd = &cobra.Command{
...
+				servicePortNames := strings.Join(serviceURL.PortNames, "\n")
 				serviceURLs := strings.Join(serviceURL.URLs, "\n")
 
 				// if we are running Docker on OSX we empty the internal service URLs
 				if runtime.GOOS == "darwin" && cfg.Driver == oci.Docker {
 					serviceURLs = ""
 				}
 
-				data = append(data, []string{serviceURL.Namespace, serviceURL.Name, "", serviceURLs})
+				data = append(data, []string{serviceURL.Namespace, serviceURL.Name, servicePortNames, serviceURLs})
```
josedonizetti: What we print for `minikube service list` must be consistent with what we print for `minikube service <service-name>`, so if we are printing `name || port` there, the same should be done here, but consider my recommendation to do it like `name/port`

pkg/minikube/service/service.go
medyagh: I agree ! if needed feel free to change the profile column name,

medyagh:
@prasadkatti btw welcome to contributing to minikbue :) happy to see your contribution !

prasadkatti:
@medyagh - Thanks. I am hoping to learn more about minikube/kubernetes through this process. I have added the before/after images that you requested to the issue description. The column header does need to change in my opinion, but I am not sure what I should change it to. Please provide suggestions.

pkg/minikube/service/service.go
prasadkatti: Thanks for the review. I have pushed a change to address your feedback. I have added before/after images to issue description. The column name does need to be changed in my opinion. But I am not sure what would be the right name for the column.

cmd/minikube/cmd/service_list.go
prasadkatti: Both `minikube service list` and `minikube service <service-name>` eventually call `printURLsForService` which is where I am making these changes. So the results should be consistent. Maybe the variable name needs to change?

prasadkatti:
Also, some tests are failing. I haven't quite figured that out yet. Are they failing because of these changes? I will need some help with that.

medyagh:
@prasadkatti unit tests are failing please make sure to run make test locally to see the error or checkout the https://github.com/kubernetes/minikube/pull/6879/checks?check_run_id=486144522

prasadkatti:
@medyagh - I have fixed the unit tests. I had to update the tests like you suspected.

prasadkatti:
Should I remove the `wip:` tag on this now?

medyagh:
Thanks @prasadkatti I will review this tommorow

tstromberg:
@medyagh - reminder to take a look at this PR when you get a chance

prasadkatti:
Any update on this? Do I need to fix anything else here?

medyagh:
@prasadkatti  thank you for your patience, I will take another look after the integration tests restuls come back.

medyagh:
Thank you very much for this PR, this will improve minikube service command !

Issue #6824, opened 2020-02-27
Minikube (re)start on docker driver breaks docker-env


balopat:
```
 minikube start -p docker3 --container-runtime docker --vm-driver=docker 
```
Restarting caused some issue with the IP (minikube docker-env command) that was working before - this is a skaffold log output:  

```
'docker load' command failed with error: error during connect: Post https://127.0.0.1:32772/v1.40/images/load?quiet=1: EOF
```

```
minikube version                                                                                                                     6.7s  Thu Feb 27 10:59:00 2020
minikube version: v1.7.3
commit: 436667c819c324e35d7e839f8116b968a2d0a3ff
```

_Originally posted by @balopat in https://github.com/kubernetes/minikube/issues/6820#issuecomment-592036441_

josedonizetti:
The bugs happens because kic is using random ports when exposing internal ports to localhost. When a container is restarted it might get the same port or not (likely to not).

https://github.com/kubernetes/minikube/blob/master/pkg/drivers/kic/oci/oci.go#L407

To fix this we need to control which port is exposed, but passing it to the run command. But this can have side effects, imagine we start kic with passing the port 32011, and then the container is stopped, and some other service is started on the user computer that is not related to minikube and this service is using the port 32011, if the user now tries to restart the container we have a conflict, which to be avoid will required another port, creating the same bug again.

One can also run `docker-env` again every time the container is restarted.

@medyagh thoughts?

medyagh:
Regarding that issue I was thinking. Minikube status should have a new line that tells you if you are using the docker env inside minikube or not . And if it needs updating it should also tell you.

medyagh:
Another almost related is we could add something to the bash prompt (if user selects to)
That tells you your terminal is using docker inside minikube

medyagh:
Additionally on restart, we should detect user is using docker env and re-new docker env for them or warn them to do so

medyagh:
I will work on this.

medyagh:
I also noticed on a terminal with docker-env
we warn this incorrectly

```
⚠️  'docker' driver reported an issue: exit status 1
💡  Suggestion: Docker is not running or is responding too slow. Try: restarting docker desktop.
```

PR #6885, opened 2020-03-04
warn to update docker-env on restart

medyagh:
### Before this PR:
in case of an minikube docker-env in terminal:

docker driver restart was sending wrong warnning because was pointing to wrong docker
```
⚠️  'docker' driver reported an issue: exit status 1
💡  Suggestion: Docker is not running or is responding too slow. Try: restarting docker desktop.
```

### Before this PR Also:

docker-env would break after a restart without any warnning to user

### After this PR when run in a terminal with docker env set:

- No False warrning
- Warn user to Re-eval the docker-env after restarting docker env.
```
😄  minikube v1.7.3 on Darwin 10.14.6
    ▪ MINIKUBE_ACTIVE_DOCKERD=minikube
✨  Using the docker (experimental) driver based on existing profile
📌  Noticed that you are using minikube docker-env:
⚠️  After minikube restart the dockerd ports might have changed. To ensure docker-env works properly.
Please re-eval the docker-env command:

        'minikube -p minikube docker-env'


🏃  Using the running docker "minikube" VM ...
```

### after this PR when run in a terminal without docker env set
```
$ ./out/minikube start
😄  minikube v1.7.3 on Darwin 10.14.6
✨  Using the docker (experimental) driver based on existing profile
🏃  Using the running docker "minikube" VM ...
```

closes https://github.com/kubernetes/minikube/issues/6824 
but still the user has to do manual step !

### DIFF
--- a/pkg/minikube/machine/fix.go
+++ b/pkg/minikube/machine/fix.go
@@ -159,7 +159,6 @@ func warnToReEvalEnv(drver string, name string) {
 		out.T(out.WarningType, `Please run the following command: 
 		'minikube -p {{.profile_name}} docker-env'`, out.V{"profile_name": name})
 	}
-	return
 }
 
 // ensureGuestClockSync ensures that the guest system clock is relatively in-sync
--- a/pkg/minikube/machine/fix.go
+++ b/pkg/minikube/machine/fix.go
@@ -156,7 +156,8 @@ func warnToReEvalEnv(drver string, name string) {
 	p := os.Getenv(constants.MinikubeActiveDockerdEnv)
 	if p != "" {
 		out.T(out.WarningType, "dockerd port changed since restart. minikube's docker-env need to be updated.")
-		out.T(out.WarningType, "Please run the following command: 'minikube -p {{.profile_name}} docker-env'", out.V{"profile_name": name})
+		out.T(out.WarningType, `Please run the following command: 
+		'minikube -p {{.profile_name}} docker-env'`, out.V{"profile_name": name})
 	}
 	return
 }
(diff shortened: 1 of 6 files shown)


pkg/minikube/machine/fix.go
```diff
@@ -145,6 +150,26 @@ func fixHost(api libmachine.API, mc config.ClusterConfig) (*host.Host, error) {
...
+// because docker changes the allocated bind ports after restart https://github.com/kubernetes/minikube/issues/6824
+func maybeWarnAboutEvalEnv(drver string, name string) {
+	if !driver.IsKIC(drver) {
+		return
+	}
+	p := os.Getenv(constants.MinikubeActiveDockerdEnv)
+	if p == "" {
+		return
+	}
+	out.T(out.Notice, "Noticed that you are using minikube docker-env:")
```
priyawadhwa: nit: maybe "We noticed"

pkg/minikube/machine/fix.go
```diff
@@ -145,6 +150,26 @@ func fixHost(api libmachine.API, mc config.ClusterConfig) (*host.Host, error) {
...
+	out.T(out.WarningType, `After minikube restart the dockerd ports might have changed. To ensure docker-env works properly.
+Please re-eval the docker-env command:
```
priyawadhwa: nit: "Please rerun the docker-env command:"

medyagh: I was thinking re run wouldn't actually help. They have to actually do the eval 
But since it is different for each bash I left that part to them.

Ideally we want to give them the right eval command based on their shell

PR #6888, opened 2020-03-04
Switch --vm-driver paramter to --driver

### LINKED ISSUE #6620: Rename --vm-driver to --driver
This makes kic less confusing.

We'll still need a compatibility flag.

sharifelgamal:
Fixes #6620 

Anywhere where `--vm-driver` was accepted (`minikube start` and `minikube config` most notably), now `--driver` will be accepted. `--vm-driver` is still supported but deprecated and will be removed sometime in the future.


Makefile
```diff
@@ -623,13 +623,13 @@ kvm_in_docker:
 	rm -f out/docker-machine-driver-kvm2
 	$(call DOCKER,$(KVM_BUILD_IMAGE),/usr/bin/make out/docker-machine-driver-kvm2 COMMIT=$(COMMIT))
 
-.PHONY: install-kvm-driver
-install-kvm-driver: out/docker-machine-driver-kvm2  ## Install KVM Driver
+.PHONY: install-kvmdriver
```
medyagh: I think this was a mistake replace

sharifelgamal: ah yeah good catch

PR #6898, opened 2020-03-05
Use preloaded tarball for VMs as well

priyawadhwa:
Since preloaded tarballs are container runtime specific, I added a Preload() func
to the cruntime interface. To use  it with VMs, the tarball is first
copied into the VM and then extracted. The leftover tarball is deleted,
and the container runtime (currently only docker is supported) is
restarted.


pkg/minikube/cruntime/containerd.go
```diff
@@ -308,3 +308,8 @@ func (r *Containerd) ContainerLogCmd(id string, len int, follow bool) string {
 func (r *Containerd) SystemLogCmd(len int) string {
 	return fmt.Sprintf("sudo journalctl -u containerd -n %d", len)
 }
+
+// Preload preloads the container runtime with k8s images
+func (r *Containerd) Preload(k8sVersion string) error {
+	return nil
```
medyagh: maybe return error "not implemented for containerd yet"

pkg/minikube/cruntime/crio.go
```diff
@@ -225,3 +225,8 @@ func (r *CRIO) ContainerLogCmd(id string, len int, follow bool) string {
 func (r *CRIO) SystemLogCmd(len int) string {
 	return fmt.Sprintf("sudo journalctl -u crio -n %d", len)
 }
+
+// Preload preloads the container runtime with k8s images
+func (r *CRIO) Preload(k8sVersion string) error {
+	return nil
```
medyagh: same as here

pkg/minikube/cruntime/containerd.go
priyawadhwa: done

pkg/minikube/cruntime/crio.go
priyawadhwa: done

pkg/minikube/cruntime/docker.go
```diff
@@ -252,3 +265,34 @@ func (r *Docker) ContainerLogCmd(id string, len int, follow bool) string {
...
 	return fmt.Sprintf("sudo journalctl -u docker -n %d", len)
 }
+
+// Preload preloads docker with k8s images:
+// 1. Copy over the preloaded tarball into the VM
+// 2. Extract the preloaded tarball to the correct directory
+// 3. Remove the tarball within the VM
+func (r *Docker) Preload(k8sVersion string) error {
+	tarballPath := download.TarballPath(k8sVersion)
+	dest := "/tmp/preloaded.tar"
```
tstromberg: .tar might be confusing - how about /tmp/preloaded.tar.lz4 ?

pkg/minikube/cruntime/docker.go
```diff
@@ -252,3 +265,34 @@ func (r *Docker) ContainerLogCmd(id string, len int, follow bool) string {
...
+	}
+	glog.Infof("Took %f seconds to copy over tarball", time.Since(t).Seconds())
+
+	// extract the tarball to /var in the VM
+	if rr, err := r.Runner.RunCmd(exec.Command("sudo", "tar", "-I", "lz4", "-C", "/var", "-xvf", dest)); err != nil {
+		return errors.Wrapf(err, "extracting tarball: %s", rr.Output())
+	}
+
+	//  remove the tarball in the VM
+	if rr, err := r.Runner.RunCmd(exec.Command("sudo", "rm", dest)); err != nil {
```
tstromberg: sudo shouldn't be required - but I believe you can just call `r.Runner.Remove(fa)` as well.

pkg/minikube/cruntime/docker.go
priyawadhwa: done

PR #6900, opened 2020-03-05
Automatically scale the default memory allocation

### LINKED ISSUE #5021: Improve the default --memory setting (dynamic sizing?)
While working with @priyawadhwa on a demo, I noticed that minikube was locking up. Increasing the memory size fixed the problem. Also, the first thing our documentation states after installation is to increase the memory size.

What if we were able to improve the first-start experience for most users by default, by dynamically selecting a more appropriate option for `--memory`?

My recommendation is to change the default 2GB setting to 37.5% of available memory by default: but never less than 2.25GB or more than 8GB. For instance:

- A host with 4GB of RAM dedicates 2.25GB to the VM
- A host with 6GB of RAM dedicates 2.25GB to the VM
- A host with 8GB of RAM dedicates 3GB to the VM
- A host with 16GB of RAM dedicates 6GB to the VM
- A host with 32GB of RAM dedicates 8GB to the VM
- A host with 64GB of RAM dedicates 8GB to the VM

The balance can be fine tuned, but the main point I was going for was to optimize that developers with a 16GB machine get a 6GB VM by default.

tstromberg:
Fixes #5021

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -38,9 +38,11 @@ import (
 	"github.com/pkg/errors"
 	"github.com/shirou/gopsutil/cpu"
 	gopshost "github.com/shirou/gopsutil/host"
+	"github.com/shirou/gopsutil/mem"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	cmdcfg "k8s.io/minikube/cmd/minikube/cmd/config"
+	"k8s.io/minikube/pkg/drivers/kic/oci"
 	"k8s.io/minikube/pkg/minikube/bootstrapper/bsutil"
 	"k8s.io/minikube/pkg/minikube/bootstrapper/images"
 	"k8s.io/minikube/pkg/minikube/cluster"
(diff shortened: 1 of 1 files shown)


medyagh:
tried it locally with my mac looks good

```
Password: 
    > minikube-v1.7.3.iso.sha256: 65 B / 65 B [--------------] 100.00% ? p/s 0s
    > minikube-v1.7.3.iso: 167.39 MiB / 167.39 MiB [] 100.00% 16.85 MiB p/s 10s
🔥  Creating hyperkit VM (CPUs=2, Memory=4096MB, Disk=20000MB) 
```

afbjorklund:
Did you feedback this to `kubeadm`, that it no longer works with the recommended 2048M ?

[https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#before-you-begin)
> * 2 GB or more of RAM per machine (any less will leave little room for your apps)
> * 2 CPUs or more

Or is the problem that it needs 2Gi free - just for Kubernetes, rather than 2Gi **total** ?

The wording of "per machine" is a little ambiguous. Elsewhere it says "per node"

Issue #6899, opened 2020-03-05
master: The container name "/minikube" is already in use by container 


tstromberg:
To replicate:

- Run `minikube start` while Docker is coming up. It errors out with `Unable to start VM. Please investigate and run 'minikube delete' if possible: creating host: create: creating: create kic node: creating volume for minikube container: output Error response from daemon: dial unix docker.raw.sock: connect: connection refused`

- Run `minikube startt` once Docker is up:

```
💣  Unable to start VM. Please investigate and run 'minikube delete' if possible: creating host: create: creating: create kic node: create a kic node: args: [run -d -t --privileged --security-opt seccomp=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --label mode.minikube.sigs.k8s.io=minikube --volume minikube:/var --cpus=2 --memory=6000mb --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 gcr.io/k8s-minikube/kicbase:v0.0.7@sha256:a6f288de0e5863cdeab711fa6bafa38ee7d8d285ca14216ecf84fcfb07c7d176]  output: [docker: Error response from daemon: Conflict. The container name "/minikube" is already in use by container "8f989d183f9a10ec0bc0f2061abdf32be34daeebe5f59049a22908b80c1bffde". You have to remove (or rename) that container to be able to reuse that name. See 'docker run --help'.] : exit status 125
```

If you run `minikube start` again, it gets stuck here instead:

```
Unable to start VM. Please investigate and run 'minikube delete' if possible: creating host: create: creating: create kic node: creating volume for minikube container: output Error response from daemon: dial unix docker.raw.sock: connect: connection refused
```

medyagh:
I was able to reproduce this bug this way:

```
medya@~/workspace/minikube (master) $ docker kill 8f47e4ec187d
8f47e4ec187d
medya@~/workspace/minikube (master) $ docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
medya@~/workspace/minikube (master) $ docker ps -a
CONTAINER ID        IMAGE                                COMMAND                  CREATED             STATUS                       PORTS               NAMES
[...]

Issue #6713, opened 2020-02-20
docker:  Conflict. The container name "/minikube" is already in use by container 


tstromberg:
I tried to delete a cluster, which complained:

```
🔥  Deleting "minikube" in docker ...
❌  Failed to delete cluster: host remove: removing container minikube, output Error response from daemon: removal of container minikube is already in progress: exit status 1
📌  You may need to manually remove the "minikube" VM from your hypervisor
🔥  Removing /Users/tstromberg/.minikube/machines/minikube ...
💀  Removed all traces of the "minikube" cluster.
```

when starting in up, it didn't work again either:

```
/out/minikube start --vm-driver=docker
😄  minikube v1.7.2 on Darwin 10.14.6
✨  Using the docker (experimental) driver based on user configuration
🔥  Creating Kubernetes in docker container with (CPUs=2), Memory=2000MB (32768MB available) ...

💣  Unable to start VM. Please investigate and run 'minikube delete' if possible: creating host: create: creating: create kic node: create a kic node: args: [run -d -t --privileged --security-opt seccomp=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --volume minikube:/var --cpus=2 --memory=2000mb --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 gcr.io/k8s-minikube/kicbase:v0.0.5@sha256:3ddd8461dfb5c3e452ccc44d87750b87a574ec23fc425da67dccc1f0c57d428a]  output: [docker: Error response from daemon: Conflict. The container name "/minikube" is already in use by container "76ba94a6bd7bc424cf0c06dffcf85a65c605b754cfb8dd6c83da388aa93d986e". You have to remove (or rename) that container to be able to reuse that name. See 'docker run --help'.] : exit status 125
```

`docker ps -a` shows:

```
CONTAINER ID        IMAGE                                COMMAND                  CREATED             STATUS              PORTS                                                                           NAMES
[...]

PR #6906, opened 2020-03-06
handle already in use container name

medyagh:
closes https://github.com/kubernetes/minikube/issues/6899
closes https://github.com/kubernetes/minikube/issues/6713

## before this PR:

if you delete machine folder and profile folder and kill the minikube container

```
medya@~/workspace/minikube (master) $ docker kill 8f47e4ec187d
medya@~/workspace/minikube (master) $ rm -rf ~/.minikube/profiles/minikube/
medya@~/workspace/minikube (master) $ ./out/minikube start --vm-driver=docker
```
and then start minikube  you would get this error
```
😄  minikube v1.7.3 on Darwin 10.14.6
🔥  Creating Kubernetes in docker container with (CPUs=2) (8 available), Memory=4096MB (7963MB available) ...

💣  Unable to start VM. Please investigate and run 'minikube delete' if possible: creating host: create: creating: create kic node: create a kic node: args: [run -d -t --privileged --security-opt seccomp=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --label mode.minikube.sigs.k8s.io=minikube --volume minikube:/var --cpus=2 --memory=4096mb --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 gcr.io/k8s-minikube/kicbase:v0.0.7@sha256:a6f288de0e5863cdeab711fa6bafa38ee7d8d285ca14216ecf84fcfb07c7d176]  output: [docker: Error response from daemon: Conflict. The container name "/minikube" is already in use by container "8f47e4ec187d722bdc748714c29cd7e8ec35a4a4dee5d39c77db050c020eda3f". You have to remove (or rename) that container to be able to reuse that name. See 'docker run --help'.] : exit status 125

```

### After this PR:

- if we find already existing container with same name, we will check if was created by minikube (by inspecting the labels) if yes we will delete it and recreate it.
```
😄  minikube v1.7.3 on Darwin 10.13.6
🔥  Creating Kubernetes in docker container with (CPUs=2) (2 available), Memory=4096MB (7964MB available) ...
🚀  Launching Kubernetes ... 
```

in the logs you would be seeing

```
I0305 22:27:02.148397   13158 kic.go:103] Found already existing abandoned minikube container, will try to delete.
```

- if container exists but was not created by minikube (a user's container) will exit with an error to tell user they need to delete the conflicting container name.

### DIFF
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -93,22 +93,24 @@ func (d *Driver) Create() error {
 		},
 	)
 
-	id, err := oci.ContainerID(d.OCIBinary, params.Name)
+	exists, err := oci.ContainerExists(d.OCIBinary, params.Name)
 	if err != nil {
-		return errors.Wrap(err, "check container exists")
+		glog.Infof("failed to check if container already exists: %v", err)
 	}
-	if id != "" {
-		// if container was created by minikube it is safe to delete it.
-		if oci.IsCreatedByMinikube(d.OCIBinary, id) {
-			if err := oci.DeleteContainersByID(d.OCIBinary, id); err != nil {
-				glog.Errorf("Failed to delete a conflicting minikube container %s. You might need to restart your %s daemon and delete it manually and try again.", id)
-				return errors.Wrapf(err, "deleting already in-use mini-created container %s", id)
+	if exists {
+		// if container was created by minikube it is safe to delete and recreate it.
+		if oci.IsCreatedByMinikube(d.OCIBinary, params.Name) {
+			glog.Info("Found already existing abandoned minikube container, will try to delete.")
+			if err := oci.DeleteContainersByID(d.OCIBinary, params.Name); err != nil {
+				glog.Errorf("Failed to delete a conflicting minikube container %s. You might need to restart your %s daemon and delete it manually and try again.", params.Name)
+				return errors.Wrapf(err, "deleting already in-use mini-created container %s", params.Name)
 			}
+		} else {
+			// if not the conflicting container name is not created by minikube
+			// that means it is a user has a container that conflicts with minikube profile name
+			glog.Errorf("You have a container named %s that conflicts with minikube profile name %s, please either remove manually or choose a different minikbue profile name", params.Name, params.Name)
+			return errors.Wrap(err, "conflicting name with user's container")
 		}
-		// if not the conflicting container name is not created by minikube
-		// that means it is a user has a container that conflicts with minikube profile name
-		glog.Errorf("You have a container named %s that conflicts with minikube profile name %s, please either remove manually or choose a different minikbue profile name", id)
-		return errors.Wrap(err, "conflicting name with user's container")
 	}
 
 	if err := oci.CreateContainerNode(params); err != nil {
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -106,8 +106,8 @@ func (d *Driver) Create() error {
 				return errors.Wrapf(err, "deleting already in-use mini-created container %s", params.Name)
 			}
 		} else {
-			// if not the conflicting container name is not created by minikube
-			// that means it is a user has a container that conflicts with minikube profile name
+			// The conflicting container name was not created by minikube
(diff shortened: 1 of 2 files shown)


pkg/drivers/kic/kic.go
```diff
@@ -92,6 +92,27 @@ func (d *Driver) Create() error {
...
+	if err != nil {
+		glog.Warningf("failed to check if container already exists: %v", err)
+	}
+	if exists {
+		// if container was created by minikube it is safe to delete and recreate it.
+		if oci.IsCreatedByMinikube(d.OCIBinary, params.Name) {
+			glog.Info("Found already existing abandoned minikube container, will try to delete.")
+			if err := oci.DeleteContainer(d.OCIBinary, params.Name); err != nil {
+				glog.Errorf("Failed to delete a conflicting minikube container %s. You might need to restart your %s daemon and delete it manually and try again.", params.Name, params.OCIBinary)
+				return errors.Wrapf(err, "deleting already in-use mini-created container %s", params.Name)
```
tstromberg: Does DeleteContainer fail if the container has been removed already? 

If so, we should just log the error and try CreateContainer anyways, rather than adding a possible race against another process which may have deleted the container in the mean time.

pkg/drivers/kic/kic.go
```diff
@@ -92,6 +92,27 @@ func (d *Driver) Create() error {
...
+			}
+		} else {
+			// The conflicting container name was not created by minikube
+			// user has a container that conflicts with minikube profile name, will not delete users container.
+			glog.Errorf("You have a container named %s that conflicts with minikube profile name %s, please either remove manually or choose a different minikbue profile name", params.Name, params.Name)
```
tstromberg: Log the error or return the error - not both. This results in duplicate errors being displayed.

Consider %q for the profile name for clarity. Also check the spelling of minikube.

pkg/drivers/kic/kic.go
tstromberg: Also, "mini-created" is not be meaningful to users. Is there another way to state this?

pkg/drivers/kic/oci/oci.go
```diff
@@ -69,6 +69,25 @@ func DeleteContainersByLabel(ociBin string, label string) []error {
...
+// DeleteContainer deletes a container by ID or Name
+func DeleteContainer(ociBin string, name string) error {
+	if err := PointToHostDockerDaemon(); err != nil {
+		return errors.Wrap(err, "point host docker daemon")
+	}
+	_, err := ContainerStatus(ociBin, name)
+	// only try to delete if docker/podman inspect returns
+	// if it doesn't it means docker daemon is stuck and needs restart
+	if err != nil {
+		glog.Errorf("%s daemon seems to be stuck. Please try restarting your %s. Will try to delete anyways", ociBin, ociBin)
```
tstromberg: Please include the error message seen, for debugging sake.

There may be other reasons that ContainerStatus returns an error other than being stucj.

pkg/drivers/kic/oci/oci.go
```diff
@@ -69,6 +69,25 @@ func DeleteContainersByLabel(ociBin string, label string) []error {
 	return deleteErrs
 }
 
+// DeleteContainer deletes a container by ID or Name
+func DeleteContainer(ociBin string, name string) error {
+	if err := PointToHostDockerDaemon(); err != nil {
+		return errors.Wrap(err, "point host docker daemon")
+	}
+	_, err := ContainerStatus(ociBin, name)
+	// only try to delete if docker/podman inspect returns
```
tstromberg: This comment is no longer accurate. Perhaps we should just remove the Status check?

pkg/drivers/kic/kic.go
medyagh: good point

pkg/drivers/kic/oci/oci.go
medyagh: good point

PR #6907, opened 2020-03-06
download.Binary: Mock downloads for testing

### LINKED ISSUE #6751: Running unit tests downloads kubernetes binaries
It would be nice if `TestCacheBinariesForBootstrapper` could mock the actual downloads...

 Downloading the `v1.16.0` can take a long time, and wastes space in the minikube cache.

`--- PASS: TestCacheBinariesForBootstrapper (183.97s)`

`418M	/home/anders/.minikube/cache/v1.16.0/`

nkprince007:
fixes #6751

### DIFF
--- a/pkg/minikube/bootstrapper/bsutil/binaries.go
+++ b/pkg/minikube/bootstrapper/bsutil/binaries.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	"github.com/golang/glog"
+	"github.com/hashicorp/go-getter"
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
 	"k8s.io/minikube/pkg/minikube/command"
@@ -59,7 +60,7 @@ func TransferBinaries(cfg config.KubernetesConfig, c command.Runner) error {
 	for _, name := range constants.KubernetesReleaseBinaries {
 		name := name
 		g.Go(func() error {
-			src, err := download.Binary(name, cfg.KubernetesVersion, "linux", runtime.GOARCH)
+			src, err := download.Binary(name, cfg.KubernetesVersion, "linux", runtime.GOARCH, &getter.FileGetter{})
 			if err != nil {
 				return errors.Wrapf(err, "downloading %s", name)
 			}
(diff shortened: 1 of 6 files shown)


tstromberg:
Something is weird about the checksum URL being generated in this PR:

`https://storage.googleapis.com/kubernetes-release/release/v1.11.10/bin/linux/amd64/kubeadm?checksum=file:https://stora
ge.googleapis.com/kubernetes-release/release/v1.11.10/bin/linux/amd64/kubeadm.sha1`

nkprince007:
> I appreciate the clarity of this solution, but don't love that it leaks the implementation detail of go-getter being used underneath.
> 
> What do you think about an alternative approach like the downloads package having a `downloads.Getter`:
> 
> ```
> var (
>   Getter = getter.FileGetter{}
>   FileGetter = getter.FileGetter{}
>   MockGetter = getter.MockGetter{}
> )
> ```
> 
> then tests can set: `download.Getter = download.MockGetter`
> 
> The advantages I see are:
> 
>     * This would now mock out all types of downloads (ISO, Binaries, Preload)
> 
>     * Callers won't have to know about getters
> 
> 
> The disadvantages I see are:
> 
>     * Global state can be confusing if tests are run in parallel (we currently don't do this)
> 
>     * Introduces a public package variable
> 
> 
> It's unlikely that unit tests will be toggling back between getters, however.
> 
> Thoughts?

Global state is a point of concern though.

tstromberg:
Seems reasonable. Go ahead and rebase, but remove the uri.Parse bit. It doesn't seem to be doing the right thing with the URL in this case.

nkprince007:
Alright, I'll update it :)

nkprince007:
> Seems reasonable. Go ahead and rebase, but remove the uri.Parse bit. It doesn't seem to be doing the right thing with the URL in this case.

@tstromberg, `Get` function requires a `*url.URL` parameter.

tstromberg:
PR is still broken:

```
# k8s.io/minikube/pkg/minikube/download
pkg/minikube/download/binary.go:76:23: cannot use url (type string) as type *url.URL in argument to gtr.GetFile
```

tstromberg:
Closing until this PR can be revisited. Please re-open it when it's ready for review again.

PR #6914, opened 2020-03-06
Add doc for running ebpf based tools in minikube

priyawadhwa:
This is super hacky, but it should work


site/content/en/docs/Tutorials/ebpf_tools_in_minikube.md
```diff
@@ -0,0 +1,64 @@
...
+
+```
+$ minikube start
+```
+
+Mount kernel headers that were built during the ISO build into the minikube VM:
+
+```
+$ cd minikube
+$ minikube mount $(pwd)/out/buildroot/output/build/linux-4.19.94:/lib/modules/4.19.94/build
```
priyawadhwa: Hey @afbjorklund -- these kernel headers are required for ebpf tools to run, and they exist on my host filesystem when I build the ISO. I can't figure out how to include the headers in the ISO so that this step can be skipped -- do you have any guidance around this?

afbjorklund: For Falco we shipped a prebuilt kernel module, rather than providing users with compilers and headers at runtime

site/content/en/docs/Tutorials/ebpf_tools_in_minikube.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+This command will need to stay running for as long as you want to run BPF tools.
+
+In another terminal window, you can now run [BCC tools](https://github.com/iovisor/bcc) as a Docker container in minikube:
```
afbjorklund: Would it be better or easier to provide a BCC package for minikube ? https://github.com/iovisor/bcc/blob/master/INSTALL.md

site/content/en/docs/Tutorials/ebpf_tools_in_minikube.md
tstromberg: One alternative approach: we could just fetch the appropriate kernel source tarball and extract the headers

priyawadhwa: That's much better, I created a tarball of the headers which an be downloaded from GCS and updated the doc.

@afbjorklund I don't really know anything about prebuilding kernel modules, would you know of any good references to get started with that?

site/content/en/docs/Tutorials/ebpf_tools_in_minikube.md
priyawadhwa: For now I think it's easier to run as a container because then we don't have to include extra dependencies in the ISO (python is the first one I can think of off the top of my head)

Issue #6595, opened 2020-02-12
eval (minikube -p minikube docker-env) doesn't work in fish-shell


agusti-t:
The command doesn't do what it's expected in fish-shell, may be related to: https://github.com/kubernetes/minikube/pull/6540
 
```
minikube version                                                                                                    Wed Feb 12 15:42:40 2020
minikube version: v1.7.2
commit: 50d543b5fcb0e1c0d7c27b1398a9a9790df09dfb
```

**The exact command to reproduce the issue**: `eval (minikube -p minikube docker-env)`

**The full output of the command that failed**:
```
 ~  minikube -p minikube docker-env                                                                                                                                              Wed Feb 12 16:30:00 2020
set -gx DOCKER_TLS_VERIFY "1"
set -gx DOCKER_HOST "tcp://192.168.64.33:2376"
set -gx DOCKER_CERT_PATH "/Users/agusti/.minikube/certs"
set -gx MINIKUBE_ACTIVE_DOCKERD "minikube"

# To point your shell to minikube's docker-daemon, run:
# eval (minikube -p minikube docker-env)
 ~  env | grep DOCKER                                                                                                                                                    328ms  Wed Feb 12 16:30:01 2020
 !  ~  eval (minikube -p minikube docker-env)                                                                                                                                   Wed Feb 12 16:30:15 2020
 ~  env | grep DOCKER                                                                                                                                                    192ms  Wed Feb 12 16:30:16 2020
DOCKER_TLS_VERIFY=1 set -gx DOCKER_HOST tcp://192.168.64.33:2376 set -gx DOCKER_CERT_PATH /Users/agusti/.minikube/certs set -gx MINIKUBE_ACTIVE_DOCKERD minikube
```

Quick workaround is to just copy and paste the set commands and run them manually:
```
set -gx DOCKER_TLS_VERIFY "1"
set -gx DOCKER_HOST "tcp://192.168.64.33:2376"
set -gx DOCKER_CERT_PATH "/Users/agusti/.minikube/certs"
set -gx MINIKUBE_ACTIVE_DOCKERD "minikube"
```

Something like:
```
 !  ~  set -e DOCKER_TLS_VERIFY                                                                                                                                                 Wed Feb 12 16:33:19 2020
 ~  env | grep DOCKER                                                                                                                                                            Wed Feb 12 16:33:36 2020
[...]

PR #6915, opened 2020-03-06
docker-env: Add semicolons back for fish, required for fish 2.x users

tstromberg:
While testing v1.8.0 today, i noticed that fish 2.7.1 was concatenating everything badly, creating the following environment variable:

`DOCKER_TLS_VERIFY=1set-gxDOCKER_HOSTtcp://127.0.0.1:32778set-gxDOCKER_CERT_PATH/Users/tstromberg/.minikube/certsset-gxMINIKUBE_ACTIVE_DOCKERDminikube`


afbjorklund:
Probably fixed #6595

PR #6919, opened 2020-03-06
none_test: Fix nil pointer dereference on stat error

tstromberg:
Seen while investigating failures in https://github.com/kubernetes/minikube/runs/491029693

### DIFF
--- a/test/integration/none_test.go
+++ b/test/integration/none_test.go
@@ -83,7 +83,7 @@ func TestChangeNoneUser(t *testing.T) {
 			t.Errorf("stat(%s): %v", p, err)
 			continue
 		}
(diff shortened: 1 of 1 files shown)


test/integration/none_test.go
```diff
@@ -81,6 +81,11 @@ func TestChangeNoneUser(t *testing.T) {
 		info, err := os.Stat(p)
 		if err != nil {
 			t.Errorf("stat(%s): %v", p, err)
+			continue
+		}
+		if info == nil && info.Sys() == nil {
```
sharifelgamal: should this be an OR?

tstromberg: yup. Thank you!

Issue #6923, opened 2020-03-06
docker: index .NetworkSettings.Ports index of untyped nil


tstromberg:
```
docker: Template parsing error: template: :1:4: executing "" at <index (index .NetworkSettings.Ports "8444/tcp") 0>: error calling index: index of untyped nil
```

https://storage.googleapis.com/minikube-builds/logs/6920/e1049e7/Docker_Linux.html

It seems like the `oci.HostPortBinding` may be called before the container has a port bound:

```
I0306 14:37:50.268074   26177 profile.go:100] Saving config to /home/jenkins/minikube-integration/linux-amd64-docker-6920-27507-e1049e78b46458054ad50ee6fca8baf5bbd2950a/.minikube/profiles/containerd-20200306T143724.746095642-29306/config.json ...

W0306 14:37:57.662509   26177 exit.go:101] Failed to get driver URL: api host port binding: get host-bind port 8444 for "containerd-20200306T143724.746095642-29306", output 

Template parsing error: template: :1:4: executing "" at <index (index .NetworkSettings.Ports "8444/tcp") 0>: error calling index: index of untyped nil
```

medyagh:
we still have this issue https://storage.googleapis.com/minikube-builds/logs/6924/a58a9f0/Docker_Linux.html#fail_TestStartStop%2fgroup%2fcontainerd

PR #6924, opened 2020-03-06
check if container is running after create

medyagh:
add up to 5 seconds of waiting to check and make sure the created container is "running"
and not "starting..."

fixes https://github.com/kubernetes/minikube/issues/6923

### DIFF
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -29,6 +29,7 @@ import (
 	"github.com/pkg/errors"
 	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/minikube/localpath"
+	"k8s.io/minikube/pkg/util/retry"
 
 	"fmt"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


pkg/drivers/kic/oci/oci.go
```diff
@@ -164,6 +165,23 @@ func CreateContainerNode(p CreateParams) error {
 	if err != nil {
 		return errors.Wrap(err, "create a kic node")
 	}
+
+	checkRunning := func() error {
+		s, err := ContainerStatus(p.OCIBinary, p.Name)
+		if err != nil {
+			return fmt.Errorf("temporary error checking status for %q : %v", p.Name, err)
+		}
+		if s != "running" {
```
priyawadhwa: nit: wdyt of having all possible states as constants? That way they can be reused later on.

medyagh: Thats a good idea ! eventually we should have a container status with libmachine's state strcut I will create an issue for this and do it in a following up PR

PR #6925, opened 2020-03-06
Cleanup remaining PointToHostDockerDaemon calls

tstromberg:


medyagh:
looks good, but will merge after 1.8 release since it is not critical

PR #6928, opened 2020-03-07
Block on preload download before extracting, fall back to caching images if it fails

priyawadhwa:


pkg/minikube/node/config.go
```diff
@@ -67,7 +68,10 @@ func configureRuntimes(runner cruntime.CommandRunner, drvName string, k8s config
 	}
 	if !driver.IsKIC(drvName) {
 		if err := cr.Preload(k8s.KubernetesVersion); err != nil {
-			glog.Infof("Failed to preload container runtime %s: %v", cr.Name(), err)
+			glog.Infof("Failed to preload container runtime %s: %v, falling back to caching images", cr.Name(), err)
```
tstromberg: How about glog.Errorf?

priyawadhwa: done

PR #6939, opened 2020-03-07
Upgrade Docker, from 19.03.6 to 19.03.7

afbjorklund:
containerd v1.2.13

https://docs.docker.com/engine/release-notes/#19037


deploy/iso/minikube-iso/package/containerd-bin/containerd-bin.mk
```diff
@@ -3,8 +3,8 @@
 # containerd
 #
 ################################################################################
-CONTAINERD_BIN_VERSION = v1.2.12
-CONTAINERD_BIN_COMMIT = 35bd7a5f69c13e1563af8a93431411cd9ecf5021
+CONTAINERD_BIN_VERSION = v1.2.13
```
tstromberg: Note this change in the PR description - or split into a second PR.

afbjorklund: Think it was there (from the commit message), accidentally replaced it while adding a link to the release notes - as per docker, we bundled runc/containerd/dockerd/docker together into one...

PR #6730, opened 2020-02-21
Upgrade golang to 1.13.6

sayboras:
Update the latest patch version for golang

PR #6949, opened 2020-03-08
Upgrade go version 1.13.6 -> 1.13.8

afbjorklund:
It is "soon" available for `kube-cross`

See: https://github.com/kubernetes/release/issues/1134

Similar to #6730

Need to wait on Kubernetes, for 1.14

See: https://github.com/kubernetes/kubernetes/pull/88638

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@
 os: linux
 language: go
 go:
-  - 1.13.6
+  - 1.13.8
 env:
   global:
     - GOPROXY=https://proxy.golang.org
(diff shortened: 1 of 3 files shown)


afbjorklund:
All the images are being moved, so currently it has a temporary home:

`us.gcr.io/k8s-artifacts-prod/build-image/kube-cross:v1.13.8-1`

afbjorklund:
Apparently I misunderstood, it is all going to change eventually in April (so might as well change):

https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/Vanity-Domain-Flip.md#the-solution

PR #6979, opened 2020-03-10
add integration test for upgrading from minikube v1.0.0 to latest

medyagh:
to make sure we don't break user configs from v1.0.0 and user be upgradable

### DIFF
--- a/test/integration/version_upgrade_test.go
+++ b/test/integration/version_upgrade_test.go
@@ -33,16 +33,20 @@ import (
 	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/util/retry"
 
+	"github.com/blang/semver"
 	"github.com/hashicorp/go-getter"
 	pkgutil "k8s.io/minikube/pkg/util"
 )
 
-// TestVersionUpgrade downloads the latest version of minikube and runs with
-// the oldest supported k8s version and then runs the current head minikube
-// and tries to upgrade from the oldest supported k8s to newest supported k8s
-func TestVersionUpgrade(t *testing.T) {
+// TestVersionUpgradeLatest is a comprehensive combination of tests
+// downloads the current latest release binary from https://github.com/kubernetes/minikube/releases/latest
+// and starts it with:  a custom ISO-URL and oldest supported kuberentes version
+// and then stops it
+// and then starts with head minikube binary with oldest supported minikube
+// and then starts with head minikube binary with oldest newest minikube
+func TestVersionUpgradeLatest(t *testing.T) {
 	MaybeParallel(t)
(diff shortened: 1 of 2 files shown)


test/integration/version_upgrade_test.go
```diff
@@ -136,3 +140,82 @@ func TestVersionUpgrade(t *testing.T) {
...
+	r := func() error {
+		rr, err = Run(t, exec.CommandContext(ctx, tf.Name(), args...))
+		return err
+	}
+
+	// Retry to allow flakiness for the previous release
+	if err := retry.Expo(r, 13*time.Second, Minutes(30), 2); err != nil {
+		t.Fatalf("v1.0.0 start failed: %v", err)
+	}
+	rr, err = Run(t, exec.CommandContext(ctx, tf.Name(), "stop", "-p", profile))
```
sharifelgamal: Should we check minikube status here for Running as well?

test/integration/version_upgrade_test.go
```diff
@@ -136,3 +140,82 @@ func TestVersionUpgrade(t *testing.T) {
...
+		} `json:"serverVersion"`
+	}{}
+	err = json.Unmarshal(s.Stdout.Bytes(), &cv)
+
+	if err != nil {
+		t.Fatalf("error traversing json output: %v", err)
+	}
+	if _, err = semver.Parse(cv.ServerVersion.GitVersion); err != nil {
+		t.Fatalf("err parsing the kubernets version %v", err)
+	}
```
sharifelgamal: Are we just checking to make sure it's parseable? Should we make sure the k8s version is correct?

sharifelgamal: actually log the k8s version, something like
t.Logf("k8s version is %s", k8sVersion)

test/integration/version_upgrade_test.go
```diff
@@ -136,3 +140,83 @@ func TestVersionUpgrade(t *testing.T) {
...
+		ServerVersion struct {
+			GitVersion string `json:"gitVersion"`
+		} `json:"serverVersion"`
+	}{}
+	err = json.Unmarshal(s.Stdout.Bytes(), &cv)
+	if err != nil {
+		t.Fatalf("error traversing json output: %v", err)
+	}
+	if _, err = semver.Parse(cv.ServerVersion.GitVersion); err != nil {
+		t.Fatalf("err parsing the kubernets version %v", err)
```
tstromberg: Your going to need to strip the leading v:

`version_upgrade_test.go:218: err parsing the kubernets version Invalid character(s) found in major number "v1"`

tstromberg:
Can you please update this test to be shippable? Otherwise we should just close  this.

PR #6998, opened 2020-03-11
Improve kubectl mismatch UI for clarity

tstromberg:
Old UI:

```
🚀  Launching Kubernetes ... 
❗  /usr/local/bin/kubectl is version 1.16.3, and is incompatible with Kubernetes 1.18.0-beta.2. You will need to update /usr/local/bin/kubectl or use 'minikube kubectl' to connect with this cluster
```

New UI:

```
🚀  Launching Kubernetes ... 

❗  /usr/local/bin/kubectl is v1.16.3, which may be incompatible with Kubernetes v1.18.0-beta.2.
💡  You can also use 'minikube kubectl -- get pods' to invoke a matching version
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -434,7 +434,7 @@ func showKubectlInfo(kcs *kubeconfig.Settings, k8sVersion string, machineName st
 		out.Ln("")
 		out.T(out.Warning, "{{.path}} is v{{.client_version}}, which may be incompatible with Kubernetes v{{.cluster_version}}.",
 			out.V{"path": path, "client_version": client, "cluster_version": cluster})
-		out.T(out.Tip, "You can also use 'minikube kubectl -- <arguments>' to invoke a matching version",
+		out.T(out.Tip, "You can also use 'minikube kubectl -- get pods' to invoke a matching version",
 			out.V{"path": path, "client_version": client})
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -431,8 +431,11 @@ func showKubectlInfo(kcs *kubeconfig.Settings, k8sVersion string, machineName st
 	glog.Infof("kubectl: %s, cluster: %s (minor skew: %d)", client, cluster, minorSkew)
 
 	if client.Major != cluster.Major || minorSkew > 1 {
-		out.WarningT("{{.path}} is version {{.client_version}}, and is incompatible with Kubernetes {{.cluster_version}}. You will need to update {{.path}} or use 'minikube kubectl' to connect with this cluster",
+		out.Ln("")
+		out.T(out.Warning, "{{.path}} is v{{.client_version}}, which may be incompatible with Kubernetes v{{.cluster_version}}.",
 			out.V{"path": path, "client_version": client, "cluster_version": cluster})
+		out.T(out.Tip, "You can also use 'minikube kubectl -- <arguments>' to invoke a matching version",
```
medyagh: just a nit: How about adding an example ? 
miniube kubectl -- get pods

medyagh: or alternatively can we just make it work without -- ?

tstromberg: I would love for it to work without --, but I do not believe it will be easy to do so, as flags are parsed during binary initialization :(

I'll go with your first idea.

Issue #6997, opened 2020-03-11
Flake TestStartStop: error looking up service account default/default: serviceaccount "default" not found


medyagh:
as seen in here https://storage.googleapis.com/minikube-builds/logs/6991/b5681d5/Docker_Linux.html#fail_TestStartStop%2fgroup%2fcontainerd

```
** stderr ** 
Error from server (Forbidden): error when creating "testdata/busybox.yaml": pods "busybox" is forbidden: error looking up service account default/default: serviceaccount "default" not found
```

priyawadhwa:
```
-------------------------------------------------
docker_Linux      | 2020-05-16 | 100% | 12 of 12
docker_Linux      | 2020-05-17 | 100% | 12 of 12
docker_Linux      | 2020-05-18 | 100% | 12 of 12
docker_Linux      | 2020-05-19 | 100% | 12 of 12
docker_Linux      | 2020-05-20 | 100% | 5 of 5
-------------------------------------------------
HyperKit_macOS    | 2020-05-16 | 100% | 7 of 7
HyperKit_macOS    | 2020-05-17 | 100% | 11 of 11
HyperKit_macOS    | 2020-05-18 | 100% | 10 of 10
HyperKit_macOS    | 2020-05-19 | 100% | 7 of 7
HyperKit_macOS    | 2020-05-20 | 100% | 6 of 6
-------------------------------------------------
KVM_Linux         | 2020-05-16 |  93% | 13 of 14
KVM_Linux         | 2020-05-17 | 100% | 10 of 10
KVM_Linux         | 2020-05-18 | 100% | 12 of 12
KVM_Linux         | 2020-05-19 | 100% | 10 of 10
KVM_Linux         | 2020-05-20 | 100% | 6 of 6
-------------------------------------------------
Illegal division by zero at -e line 22, <> line 1.
-------------------------------------------------
podman_Linux      | 2020-05-16 |  50% | 1 of 2
-------------------------------------------------
VirtualBox_Linux  | 2020-05-16 | 100% | 12 of 12
VirtualBox_Linux  | 2020-05-17 | 100% | 12 of 12
VirtualBox_Linux  | 2020-05-18 | 100% | 12 of 12
VirtualBox_Linux  | 2020-05-19 | 100% | 12 of 12
VirtualBox_Linux  | 2020-05-20 | 100% | 5 of 5
-------------------------------------------------
VirtualBox_macOS  | 2020-05-18 | 100% | 2 of 2
[...]

PR #6999, opened 2020-03-11
wip: Reduce flake for integration test TestStartStop

medyagh:
in first verison of this PR, I tried adding checking for service account being ready to minikube wait, but that adds 13seconds of waiting, for almost very small amount of use cases 
Fixes https://github.com/kubernetes/minikube/issues/6997

### DIFF
--- a/test/integration/fn_mount_cmd.go
+++ b/test/integration/fn_mount_cmd.go
@@ -106,7 +106,7 @@ func validateMountCmd(ctx context.Context, t *testing.T, profile string) {
 	}
 
 	start := time.Now()
-	if err := retry.Expo(checkMount, time.Second, 15*time.Second); err != nil {
+	if err := retry.Expo(checkMount, time.Second, Seconds(15)); err != nil {
 		// For local testing, allow macOS users to click prompt. If they don't, skip the test.
 		if runtime.GOOS == "darwin" {
 			t.Skip("skipping: mount did not appear, likely because macOS requires prompt to allow non-codesigned binaries to listen on non-localhost port")
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -595,6 +595,20 @@ func validateProfileCmd(ctx context.Context, t *testing.T, profile string) {
 
 // validateServiceCmd asserts basic "service" command functionality
 func validateServiceCmd(ctx context.Context, t *testing.T, profile string) {
+	// to avoid https://github.com/kubernetes/minikube/issues/6997
+	// adding this logic to minikube start is not necessary but useful for rare test flakes
+	saReady := func() error { // check if default service account is created.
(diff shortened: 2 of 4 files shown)


test/integration/fn_mount_cmd.go
```diff
@@ -106,7 +106,7 @@ func validateMountCmd(ctx context.Context, t *testing.T, profile string) {
 	}
 
 	start := time.Now()
-	if err := retry.Expo(checkMount, time.Second, 15*time.Second); err != nil {
+	if err := retry.Expo(checkMount, time.Second, Seconds(15)); err != nil {
```
tstromberg: Seconds is not a good name for this function. 

How about "ApproximateSeconds" or "SecondsWithMultiplier"?

test/integration/functional_test.go
```diff
@@ -595,6 +595,20 @@ func validateProfileCmd(ctx context.Context, t *testing.T, profile string) {
 
 // validateServiceCmd asserts basic "service" command functionality
 func validateServiceCmd(ctx context.Context, t *testing.T, profile string) {
+	// to avoid https://github.com/kubernetes/minikube/issues/6997
+	// adding this logic to minikube start is not necessary but useful for rare test flakes
+	saReady := func() error { // check if default service account is created.
+		if _, err := Run(t, exec.CommandContext(ctx, "kubectl", "--context", profile, "get", "serviceaccount", "default")); err != nil {
```
tstromberg: This seems to be covering up an actual flaw that users can face. Shouldn't the start command wait for this?

medyagh: I thought about this, this rarely happens but it eats up 13 seconds of waiting on my machine. I have a feeling that would not be helpful

tstromberg: As I understand it, this integration test is surfacing a bug: minikube start + kubectl is currently racy. As it is successfully capturing a rare race condition (yay!), we should either fix it, or leave it present, not paper over it.

I believe the right way to go about this is to add a `--wait-for-serviceaccount` flag to minikube start, and set it to false initially, but override it in these tests that are affected. That way other users who prefer reliability over latency can benefit: integration tests, IDE's, etc. This way we can better measure how to optimize for serviceaccount start latency.

I would argue that defaulting this flag to 'true' would be more in the spirit of minikube (reliability > latency) , but I do understand the trade-off here, and am comfortable leaving it as false for the time being.

Thoughts?

medyagh: @tstromberg 
I agree with you, lets not merge this PR so it bugs us in integeration test,
and we fix minikube, here are my thoughts, let keep the discussion in this issue so others can participate, https://github.com/kubernetes/minikube/issues/7011#issue-579354163

medyagh:
closing in favor of https://github.com/kubernetes/minikube/pull/7209

Issue #7000, opened 2020-03-11
preload: minikube continues before preload download is over


medyagh:
this is from head on docker driver:

to replicate : first purge
```
medmac@~/workspace/minikube (master) $ ./out/minikube delete --all --purge
🔥  Successfully deleted all profiles
💀  Successfully purged minikube directory located at - [/Users/medmac/.minikube]

```
then start with docker driver

```
medmac@~/workspace/minikube (master) $ ./out/minikube start --vm-driver=docker 
😄  minikube v1.8.1 on Darwin 10.13.6
✨  Using the docker driver based on user configuration
💾  Downloading preloaded images tarball for k8s v1.17.3 ...
🔥  Creating Kubernetes in docker container with (CPUs=2) (2 available), Memory=4000MB (7964MB available) ...
    > preloaded-images-k8s-v1-v1.17.3-docker-overlay2.tar.lz4: 499.26 MiB / 499
🐳  Preparing Kubernetes v1.17.3 on Docker 19.03.2 ...
    ▪ kubeadm.pod-network-cidr=10.244.0.0/16
❌  Unable to load cached images: loading cached images: stat /Users/medmac/.minikube/cache/images/gcr.io/k8s-minikube/storage-provisioner_v1.8.1: no such file or directory
    > kubelet.sha256: 65 B / 65 B [--------------------------] 100.00% ? p/s 0s
    > kubeadm.sha256: 65 B / 65 B [--------------------------] 100.00% ? p/s 0s
    > kubectl.sha256: 65 B / 65 B [--------------------------] 100.00% ? p/s 0s
    > kubeadm: 37.52 MiB / 37.52 MiB [----------------] 100.00% 8.85 MiB p/s 4s
    > kubectl: 41.48 MiB / 41.48 MiB [----------------] 100.00% 8.15 MiB p/s 6s
    > kubelet: 106.42 MiB / 106.42 MiB [-------------] 100.00% 14.14 MiB p/s 8s
🚀  Launching Kubernetes ... 

```

I believe this happnes to only docker driver because docker driver creates much faster than others. while for VMs while the VMs are booting up the preload finish.

That makes me think may be for VM worlds we could have a warm up VM (while the big download is happnenining we could warm up the VM) (for another issue)

medyagh:
CC: @priyawadhwa

PR #7003, opened 2020-03-11
block on preload download

medyagh:
### before this PR:

"Creating Kubernetes in docker container"...starts while the preload is in progress
while this might be okay in VM world, because VM takes time to warm up but in docker it causes failures.
```
💾  Downloading preloaded images tarball for k8s v1.17.3 ...
    > preloaded-images-k8s-v1-v1.17.3-docker-overlay2.tar.lz4: 499.26 MiB / 499
...
...

❌  Unable to load cached images: loading cached images: stat /Users/medmac/.minikube/cache/images/gcr.io/k8s-minikube/storage-provisioner_v1.8.1: no such file or directory
    > kubelet.sha256: 65 B / 65 B [--------------------------] 100.00% ? p/s 0s
    > kubeadm.sha256: 65 B / 65 B [--------------------------] 100.00% ? p/s 0s
...
...
...
```

### After this PR: the docker driver will not fail if you purge:

```
medmac@~/workspace/minikube (preload_block) $ ./out/minikube delete --all --purge && ./out/minikube start --vm-driver=docker
💀  Successfully purged minikube directory located at - [/Users/medmac/.minikube]
😄  minikube v1.8.1 on Darwin 10.13.6
💾  Downloading preloaded images tarball for k8s v1.17.3 ...
    > preloaded-images-k8s-v1-v1.17.3-docker-overlay2.tar.lz4: 499.26 MiB / 499
🚀  Launching Kubernetes ... 
⌛  Waiting for cluster to come online ...
```

Fixes https://github.com/kubernetes/minikube/issues/7000

### DIFF
--- a/pkg/minikube/node/cache.go
+++ b/pkg/minikube/node/cache.go
@@ -42,6 +42,7 @@ func beginCacheKubernetesImages(g *errgroup.Group, imageRepository string, k8sVe
 			glog.Info("Caching tarball of preloaded images")
 			return download.Preload(k8sVersion, cRuntime)
 		})
+		g.Wait()
 		return
 	}
 
--- a/pkg/minikube/node/cache.go
+++ b/pkg/minikube/node/cache.go
@@ -42,8 +42,12 @@ func beginCacheKubernetesImages(g *errgroup.Group, imageRepository string, k8sVe
 			glog.Info("Caching tarball of preloaded images")
 			return download.Preload(k8sVersion, cRuntime)
 		})
-		g.Wait()
-		return
+		err := g.Wait()
+		if err == nil {
+			glog.Infof("Finished downloading the preloaded tar for %s on %s", k8sVersion, cRuntime)
+			return // don't cache individual images if preload is successful.
+		}
+		glog.Errorf("Error downloading preloaded artifacts will continue without preload: %v", err)
 	}
 
 	if !viper.GetBool("cache-images") {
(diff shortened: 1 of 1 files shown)


pkg/minikube/node/cache.go
```diff
@@ -42,7 +42,12 @@ func beginCacheKubernetesImages(g *errgroup.Group, imageRepository string, k8sVe
 			glog.Info("Caching tarball of preloaded images")
 			return download.Preload(k8sVersion, cRuntime)
 		})
-		return
+		err := g.Wait()
```
tstromberg: I think we can remove the previous call to g.Go(), and just turn it into an err check.

pkg/minikube/node/cache.go
```diff
@@ -42,7 +42,12 @@ func beginCacheKubernetesImages(g *errgroup.Group, imageRepository string, k8sVe
...
+		if err == nil {
+			glog.Infof("Finished downloading the preloaded tar for %s on %s", k8sVersion, cRuntime)
+			return // don't cache individual images if preload is successful.
+		}
+		glog.Errorf("Error downloading preloaded artifacts will continue without preload: %v", err)
```
tstromberg: Will this result in an error being displayed to the console on older ISO's?

medyagh: no this seems to be just downloading it, and it wont untar it yet or anything special

PR #7014, opened 2020-03-11
Download to a temporary files and finalize using os.Rename

tstromberg:
Make other downloaders follow the ISO download behavior, where the download is atomic through use of os.Rename as the final step.

This allows us to avoid issues around partial downloads, and skip checksum remote validation if the finalized filename exists.


pkg/minikube/download/preload.go
```diff
@@ -164,9 +163,10 @@ func saveChecksumFile(k8sVersion string) error {
 
 // verifyChecksum returns true if the checksum of the local binary matches
 // the checksum of the remote binary
-func verifyChecksum(k8sVersion string) error {
+func verifyChecksum(k8sVersion string, path string) error {
+	glog.Infof("verifying checksumm of %s ...", path)
 	// get md5 checksum of tarball path
```
medyagh: nit, update comment

PR #7015, opened 2020-03-11
none: Skip driver preload and image caching

tstromberg:
This fixes this error I noticed with the none driver in our integration tests:

```
ℹ️   OS release is Debian GNU/Linux rodete
💡  Existing disk is missing new features (lz4). To upgrade, run 'minikube delete'
🐳  Preparing Kubernetes v1.17.3 on Docker 19.03.5 ...
🚀  Launching Kubernetes ... 
```

Historically, we avoided adding a second level of image caching for the none driver, as Docker does a fine job already. Also, none driver hosts are unlikely to have lz4, and this message does not make any sense.

### DIFF
--- a/pkg/minikube/node/config.go
+++ b/pkg/minikube/node/config.go
@@ -66,7 +66,9 @@ func configureRuntimes(runner cruntime.CommandRunner, drvName string, k8s config
 	if driver.BareMetal(drvName) {
 		disableOthers = false
 	}
-	if !driver.IsKIC(drvName) {
+
+	// Preload is overly invasive for VM's, and caching is not meaningful.
+	if driver.IsVM(drvName) {
 		if err := cr.Preload(k8s.KubernetesVersion); err != nil {
 			switch err.(type) {
 			case *cruntime.ErrISOFeature:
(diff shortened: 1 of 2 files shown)


pkg/minikube/node/config.go
```diff
@@ -66,7 +66,9 @@ func configureRuntimes(runner cruntime.CommandRunner, drvName string, k8s config
 	if driver.BareMetal(drvName) {
 		disableOthers = false
 	}
-	if !driver.IsKIC(drvName) {
+
+	// Preload is overly invasive for bare metal, and caching is not meaningful.
+	if driver.IsVM(drvName) {
```
medyagh: how about Kic?

medyagh: maybe use, if !driver.BareMetal(drvName)  instead

tstromberg: The previous if statement excluded KIC, IsVM preserves that. Was Preload previously broken in KIC, or am I missing something?

medyagh: ah my bad, for kic we extract it differently than VM.
it would worth adding a comment with reference where we do it for kic. (in kic driver create)

PR #7020, opened 2020-03-12
using sudo for all systemctl calls

medyagh:


afbjorklund:
Pretty vague when it comes to descriptions and reasons, for them running as root.

I guess the main reason was the lack of a proper dbus session, in a KIC container ?

Issue #7017, opened 2020-03-12
issue with docker-env on restart


medyagh:
with latest release 1.8.1

start minikube and eval docker env
build an image with docker

```
docker build -t med .
```

minikube stop
start minikube again (either stop and start or just start while it is running )

minikube correctly tells you that the docker-env port is changed and u need to re-eval
```
😄  minikube v1.8.1 on Darwin 10.13.6
    ▪ MINIKUBE_ACTIVE_DOCKERD=minikube
✨  Using the docker driver based on existing profile
⌛  Reconfiguring existing host ...
📌  Noticed that you are using minikube docker-env:
❗  After minikube restart the dockerd ports might have changed. To ensure docker-env works properly.
Please re-eval the docker-env command:

	'minikube -p minikube docker-env'


🔄  Starting existing docker container for "minikube" ..
```
try re-eval:
```
minikube -p minikube docker-env
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://127.0.0.1:32862"
export DOCKER_CERT_PATH="/Users/medmac/.minikube/certs"
export MINIKUBE_ACTIVE_DOCKERD="minikube"

# To point your shell to minikube's docker-daemon, run:
# eval $(minikube -p minikube docker-env)
```

 $ docker ps
error during connect: Get https://127.0.0.1:32862/v1.40/containers/json: EOF

 $ docker images
error during connect: Get https://127.0.0.1:32862/v1.40/images/json: EOF

inside minikube docker is working
```
medmac@~/workspace/minikube (priyawadhwa-no-overwrite-tar) $ minikube ssh

docker@minikube:~$ docker images
REPOSITORY                                TAG                 IMAGE ID            CREATED             SIZE
med                                       latest              acc98787071f        4 minutes ago       5.59MB
k8s.gcr.io/kube-proxy                     v1.17.3             ae853e93800d        4 weeks ago         116MB
k8s.gcr.io/kube-controller-manager        v1.17.3             b0f1517c1f4b        4 weeks ago         161MB
k8s.gcr.io/kube-apiserver                 v1.17.3             90d27391b780        4 weeks ago         171MB
k8s.gcr.io/kube-scheduler                 v1.17.3             d109c0821a2b        4 weeks ago         94.4MB
alpine                                    latest              e7d92cdc71fe        7 weeks ago         5.59MB
k8s.gcr.io/coredns                        1.6.5               70f311871ae1        4 months ago        41.6MB
[...]

Issue #3878, opened 2019-03-14
--apiserver-name has no effect on ~/.kube/config


aparamon:
To reproduce:
````console
powtest@chlorine:~$ minikube version
minikube version: v0.35.0
````console
powtest@chlorine:~$ sudo -E minikube start --vm-driver=none --apiserver-name=chlorine.acdlabs.ru
````
File `~/.kube/config` is populated with
````yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: <skip>
    server: https://82.97.201.217:8443
  name: minikube
````
Cluster is created successfully:
````console
powtest@chlorine:~$ kubectl cluster-info
Kubernetes master is running at https://82.97.201.217:8443
KubeDNS is running at https://82.97.201.217:8443/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy
````
However it's expected that `server` is https://chlorine.acdlabs.ru:8443. It works if modifying generated `~/.kube/config` manually:
````console
powtest@chlorine:~$ kubectl cluster-info
Kubernetes master is running at https://chlorine.acdlabs.ru:8443
KubeDNS is running at https://chlorine.acdlabs.ru:8443/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy
````

tstromberg:
Good find. This sounds like a fairly simple template mistake. Help wanted!

PR #7021, opened 2020-03-12
Fix dockerd internal port changing on restart

medyagh:
closes https://github.com/kubernetes/minikube/issues/7017
### Gotcha 0 : driver.GetURL() is a big Lie !
GetURL in our driver is a very dirty hack, and has nothing to do with our Machine's URL it is just something to satify libmachine that wants to find out the insalled docker inside the node. 
so I had to make kic driver follow hyperkit driver and make GetURL to return a BIG fat lie !
and later replace it in setupKubeConfig function in start ...very hacky and dirty
but till we come up with a solution for machine driver. that is the solution.

related: https://github.com/docker/machine/blob/master/libmachine/provision/utils.go#L159_L175

### Gotcha 1 : Daemon reload

That was the ps aux params for docker in kic driver:
```
root@minikube:/home/docker# ps aux | grep dockerd
root     42776  2.3  1.2 2118696 98856 ?       Ssl  20:48   0:21 /usr/bin/dockerd -H fd:// --containerd=/run/containerd/containerd.sock
```

and this was for VM
```
root      2097  1.1  1.6 632732 65972 ?        Ssl  20:35   0:20 /usr/bin/dockerd -H tcp://0.0.0.0:2376 -H unix:///var/run/docker.sock --default-ulimit=nofile=1048576:1048576 --tlsverify --tlscacert /etc/docker/ca.pem --tlscert /etc/docker/server.pem --tlskey /etc/docker/server-key.pem --label provider=hyperkit --insecure-registry 10.96.0.0/12
```

after doing daemon-reload the KIC one also picked up the right settings.

### Gotcha 2: Cert pining was missing 127.0.01
when dockerd for kic has the right parameters (same as vm) it compains about cert not having the same IP
but the cert is generated by libmachine. so I had to turn of the tls verify for docker inside minikube
```
Error during connect: Get https://127.0.0.1:32799/v1.40/containers/json: x509: certificate is valid for 172.17.0.2, not 127.0.0.1
```

for that I had to Fork libmachine and add 127.0.0.1 in addition to "localhost" to the certs 
I created a PR in libmachine, till that PR gets merged, we are gonna use my fork because we need to release https://github.com/machine-drivers/machine/pull/28/files

@afbjorklund might be able to help

### Gotcha 3: Libmachuine expects Docker Port to be 2376
it is poiintless to have dockerport in a variable, because of libmachine it always has to be "2376" 
so I hard coded it for less confusing debugging for next person. to save some of their hair on their head.

### DIFF
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -200,14 +200,14 @@ func (d *Driver) GetSSHKeyPath() string {
 	return d.SSHKeyPath
 }
 
-// GetURL returns ip of the container running kic control-panel
+// GetURL returns a Docker URL inside this host
+// e.g. tcp://1.2.3.4:2376
+// more info https://github.com/docker/machine/blob/b170508bf44c3405e079e26d5fdffe35a64c6972/libmachine/provision/utils.go#L159_L175
 func (d *Driver) GetURL() (string, error) {
 	ip, err := d.GetIP()
 	if err != nil {
 		return "", err
 	}
-	// because of libmachine problem with detecting docker port https://github.com/kubernetes/minikube/issues/7017
-	// we have to do this hack https://github.com/docker/machine/blob/master/libmachine/provision/utils.go#L159_L175
 	url := fmt.Sprintf("tcp://%s", net.JoinHostPort(ip, "2376"))
 	return url, nil
 }
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -202,9 +202,13 @@ func (d *Driver) GetSSHKeyPath() string {
 
 // GetURL returns ip of the container running kic control-panel
 func (d *Driver) GetURL() (string, error) {
+	ip, err := d.GetIP()
+	if err != nil {
+		return "", err
+	}
 	// because of libmachine problem with detecting docker port https://github.com/kubernetes/minikube/issues/7017
 	// we have to do this hack https://github.com/docker/machine/blob/master/libmachine/provision/utils.go#L159_L175
-	url := fmt.Sprintf("https://%s", net.JoinHostPort("127.0.0.1", "2376"))
+	url := fmt.Sprintf("tcp://%s", net.JoinHostPort(ip, "2376"))
 	return url, nil
 }
 
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -202,11 +202,9 @@ func (d *Driver) GetSSHKeyPath() string {
 
 // GetURL returns ip of the container running kic control-panel
 func (d *Driver) GetURL() (string, error) {
-	p, err := oci.HostPortBinding(d.NodeConfig.OCIBinary, d.MachineName, d.NodeConfig.APIServerPort)
-	url := fmt.Sprintf("https://%s", net.JoinHostPort("127.0.0.1", fmt.Sprint(p)))
-	if err != nil {
-		return url, errors.Wrap(err, "api host port binding")
-	}
+	// because of libmachine problem with detecting docker port https://github.com/kubernetes/minikube/issues/7017
+	// we have to do this hack https://github.com/docker/machine/blob/master/libmachine/provision/utils.go#L159_L175
+	url := fmt.Sprintf("https://%s", net.JoinHostPort("127.0.0.1", "2376"))
 	return url, nil
 }
 
--- a/pkg/minikube/node/config.go
+++ b/pkg/minikube/node/config.go
@@ -18,6 +18,7 @@ package node
 
 import (
 	"fmt"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -128,23 +129,26 @@ func setupKubeAdm(mAPI libmachine.API, cfg config.ClusterConfig, node config.Nod
 }
 
 func setupKubeconfig(h *host.Host, c *config.ClusterConfig, n *config.Node, clusterName string) (*kubeconfig.Settings, error) {
-	addr, err := h.Driver.GetURL()
-	if err != nil {
-		exit.WithError("Failed to get driver URL", err)
-	}
-	p := n.Port
+	addr := ""
+	var err
 	if driver.IsKIC(h.DriverName) {
-		p, err = oci.HostPortBinding(h.DriverName, h.Name, n.Port)
+		p, err := oci.HostPortBinding(h.DriverName, h.Name, n.Port)
 		if err != nil {
 			exit.WithError("Failed to get host binding port for api port", err)
 		}
-	}
+		addr = fmt.Sprintf("https://" + net.JoinHostPort("127.0.0.1", strconv.Itoa(p)))
+	} else {
+		addr, err = h.Driver.GetURL()
+		if err != nil {
+			exit.WithError("Failed to get driver URL", err)
+		}
 
-	addr = strings.Replace(addr, "tcp://", "https://", -1)
-	// because of libmachine problem with detecting docker port https://github.com/kubernetes/minikube/issues/7017
-	// we have to do this hack https://github.com/docker/machine/blob/master/libmachine/provision/utils.go#L159_L175
-	// @
-	addr = strings.Replace(addr, ":2376", ":"+strconv.Itoa(p), -1)
+		addr = strings.Replace(addr, "tcp://", "https://", -1)
+		// because of libmachine problem with detecting docker port https://github.com/kubernetes/minikube/issues/7017
+		// we have to do this hack https://github.com/docker/machine/blob/master/libmachine/provision/utils.go#L159_L175
+		// contact @afbjorklund for more details
+		addr = strings.Replace(addr, ":2376", ":"+strconv.Itoa(n.Port), -1)
+	}
 
 	if c.KubernetesConfig.APIServerName != constants.APIServerName {
 		addr = strings.Replace(addr, n.IP, c.KubernetesConfig.APIServerName, -1)
--- a/pkg/minikube/node/config.go
+++ b/pkg/minikube/node/config.go
@@ -132,7 +132,7 @@ func setupKubeAdm(mAPI libmachine.API, cfg config.ClusterConfig, node config.Nod
 func setupKubeconfig(h *host.Host, cc *config.ClusterConfig, n *config.Node, clusterName string) (*kubeconfig.Settings, error) {
 	addr, err := apiServerURL(*h, *n)
 	if err != nil {
-		exit.WithError("Failed to api server URL", err)
+		exit.WithError("Failed to get api server URL", err)
 	}
 	if cc.KubernetesConfig.APIServerName != constants.APIServerName {
 		addr = strings.Replace(addr, n.IP, cc.KubernetesConfig.APIServerName, -1)
--- a/pkg/minikube/node/config.go
+++ b/pkg/minikube/node/config.go
@@ -138,10 +138,12 @@ func setupKubeconfig(h *host.Host, c *config.ClusterConfig, n *config.Node, clus
 		if err != nil {
 			exit.WithError("Failed to get host binding port for api port", err)
 		}
-
 	}
 
 	addr = strings.Replace(addr, "tcp://", "https://", -1)
+	// because of libmachine problem with detecting docker port https://github.com/kubernetes/minikube/issues/7017
+	// we have to do this hack https://github.com/docker/machine/blob/master/libmachine/provision/utils.go#L159_L175
+	// @
 	addr = strings.Replace(addr, ":2376", ":"+strconv.Itoa(p), -1)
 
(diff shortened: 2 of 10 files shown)


pkg/provision/ubuntu.go
```diff
@@ -155,6 +155,10 @@ WantedBy=multi-user.target
 		return nil, err
 	}
 
+	if err := p.Service("", serviceaction.DaemonReload); err != nil {
```
afbjorklund: You can remove this, it is done automatically by serviceaction.Start

pkg/provision/buildroot.go
```diff
@@ -100,7 +100,7 @@ Environment=DOCKER_RAMDISK=yes
...
 # container runtimes. If left unlimited, it may result in OOM issues with MySQL.
 ExecStart=
-ExecStart=/usr/bin/dockerd -H tcp://0.0.0.0:{{.DockerPort}} -H unix:///var/run/docker.sock --default-ulimit=nofile=1048576:1048576 --tlsverify --tlscacert {{.AuthOptions.CaCertRemotePath}} --tlscert {{.AuthOptions.ServerCertRemotePath}} --tlskey {{.AuthOptions.ServerKeyRemotePath}} {{ range .EngineOptions.Labels }}--label {{.}} {{ end }}{{ range .EngineOptions.InsecureRegistry }}--insecure-registry {{.}} {{ end }}{{ range .EngineOptions.RegistryMirror }}--registry-mirror {{.}} {{ end }}{{ range .EngineOptions.ArbitraryFlags }}--{{.}} {{ end }}
+ExecStart=/usr/bin/dockerd -H tcp://0.0.0.0:2376 -H unix:///var/run/docker.sock --default-ulimit=nofile=1048576:1048576 --tlsverify --tlscacert {{.AuthOptions.CaCertRemotePath}} --tlscert {{.AuthOptions.ServerCertRemotePath}} --tlskey {{.AuthOptions.ServerKeyRemotePath}} {{ range .EngineOptions.Labels }}--label {{.}} {{ end }}{{ range .EngineOptions.InsecureRegistry }}--insecure-registry {{.}} {{ end }}{{ range .EngineOptions.RegistryMirror }}--registry-mirror {{.}} {{ end }}{{ range .EngineOptions.ArbitraryFlags }}--{{.}} {{ end }}
```
afbjorklund: most likely, you don't have to modify this file

pkg/drivers/kic/kic.go
```diff
@@ -202,11 +202,13 @@ func (d *Driver) GetSSHKeyPath() string {
...
 func (d *Driver) GetURL() (string, error) {
-	p, err := oci.HostPortBinding(d.NodeConfig.OCIBinary, d.MachineName, d.NodeConfig.APIServerPort)
-	url := fmt.Sprintf("https://%s", net.JoinHostPort("127.0.0.1", fmt.Sprint(p)))
+	ip, err := d.GetIP()
 	if err != nil {
-		return url, errors.Wrap(err, "api host port binding")
+		return "", err
 	}
+	// because of libmachine problem with detecting docker port https://github.com/kubernetes/minikube/issues/7017
+	// we have to do this hack https://github.com/docker/machine/blob/master/libmachine/provision/utils.go#L159_L175
```
tstromberg: It's unclear what the hack actually is - it looks fairly normal to me. A more useful comment would be:

`// oci.HostPortBinding can be wrong, so hardcode the port because of ...`

I like the links to the issue, but don't understand the link to the code. Also, the code link will go stale at some point, since it doesn't pin to a revision.

pkg/drivers/kic/kic.go
```diff
@@ -202,11 +202,13 @@ func (d *Driver) GetSSHKeyPath() string {
 
 // GetURL returns ip of the container running kic control-panel
```
tstromberg: Update the comment here, as it isn't returning an IP. Maybe specifically say it returns a URL to the Docker endpoint to Docker inside of the container? It is inside, isn't it?

pkg/minikube/node/config.go
```diff
@@ -127,12 +129,24 @@ func setupKubeAdm(mAPI libmachine.API, cfg config.ClusterConfig, node config.Nod
...
 func setupKubeconfig(h *host.Host, c *config.ClusterConfig, n *config.Node, clusterName string) (*kubeconfig.Settings, error) {
-	addr, err := h.Driver.GetURL()
-	if err != nil {
-		exit.WithError("Failed to get driver URL", err)
-	}
-	if !driver.IsKIC(h.DriverName) {
+	addr := ""
+	var err error
+	if driver.IsKIC(h.DriverName) {
+		p, err := oci.HostPortBinding(h.DriverName, h.Name, n.Port)
```
tstromberg: For another PR perhaps, but `HostPortBinding` may be a poor name for this function, as it sounds like it may be binding to a port rather than fetching information. Perhaps:

`oci.ForwardedPort` or `oci.BoundPort`, `oci.HostPort`?

pkg/minikube/node/config.go
```diff
@@ -127,12 +129,24 @@ func setupKubeAdm(mAPI libmachine.API, cfg config.ClusterConfig, node config.Nod
...
+		addr, err = h.Driver.GetURL()
+		if err != nil {
+			exit.WithError("Failed to get driver URL", err)
+		}
+
 		addr = strings.Replace(addr, "tcp://", "https://", -1)
+		// because of libmachine problem with detecting docker port https://github.com/kubernetes/minikube/issues/7017
+		// we have to do this hack https://github.com/docker/machine/blob/master/libmachine/provision/utils.go#L159_L175
+		// contact @afbjorklund for more details
 		addr = strings.Replace(addr, ":2376", ":"+strconv.Itoa(n.Port), -1)
```
tstromberg: This code is getting hard to follow, and I especially think it would be cleaner to call GetIP instead of search and replacing hardcoded numbers. What do you think about extracting this logic into a function like:

```
func apiServerURL() {
  if KIC() {
     p = oci.HostPortBinding(h.DriverName, h.Name, n.Port)
     return fmt.Sprintf("https://" + net.JoinHostPort(oci.DefaultBindIPV4, strconv.Itoa(p))) 
  }
  ip := h.Driver.GetIP()
  return fmt.Sprintf("https://" + net.JoinHostPort(ip, strconv.Itoa(p)))
}
```

pkg/provision/buildroot.go
```diff
@@ -151,6 +151,10 @@ WantedBy=multi-user.target
 		return nil, err
 	}
 
+	if err := p.Service("", serviceaction.DaemonReload); err != nil {
```
tstromberg: Now that the docker port is hardcoded, I think this can go away. If not, please add a comment describing why we need to reload systemd files.

pkg/minikube/node/config.go
medyagh: created an issue for it

pkg/provision/buildroot.go
medyagh: the reason is in kic we install it and it already has a service file

priyawadhwa: This step can take a couple seconds to run on my laptop -- if it's not necessary it would be better to remove it from a performance standpoint

pkg/minikube/node/config.go
medyagh: good suggestion. done

pkg/minikube/node/config.go
```diff
@@ -126,18 +129,13 @@ func setupKubeAdm(mAPI libmachine.API, cfg config.ClusterConfig, node config.Nod
...
 	}
-	if !driver.IsKIC(h.DriverName) {
-		addr = strings.Replace(addr, "tcp://", "https://", -1)
-		addr = strings.Replace(addr, ":2376", ":"+strconv.Itoa(n.Port), -1)
-	}
-
-	if c.KubernetesConfig.APIServerName != constants.APIServerName {
-		addr = strings.Replace(addr, n.IP, c.KubernetesConfig.APIServerName, -1)
+	if cc.KubernetesConfig.APIServerName != constants.APIServerName {
+		addr = strings.Replace(addr, n.IP, cc.KubernetesConfig.APIServerName, -1)
```
tstromberg: Now that the worst was refactored out, this search and replace is clearly broken, especially with KIC:

- The APIServerName is not guaranteed to be a hostname at all. It can be any arbitrary string.
- We're searching and replacing for an IP which may not exist in the string

I see this feature was added here: https://github.com/kubernetes/minikube/issues/3878

I don't totally love this feature, but my suggestion would be to move this logic to apiServerURL so that it can be done without search and replace. It probably isn't necessary to support with KIC, but the search and replace is currently broken in this use case, as n.IP == 172.17.0.x, and this URL uses 127.0.0.1.

pkg/provision/buildroot.go
```diff
@@ -151,6 +151,11 @@ WantedBy=multi-user.target
 		return nil, err
 	}
 
+	// because currently we install docker differently in kic, it is harmless to do this step for VM too so the code by the same
```
tstromberg: I wouldn't go so far as claiming harmless - it does add startup latency and room for failure in a place where it isn't necessary.  I'm not convinced we should call this in the VM, but could easily be swayed by a good comment.

// We have possibly changed the Docker file, inform systemctl of this fact.

pkg/provision/ubuntu.go
```diff
@@ -155,6 +155,11 @@ WantedBy=multi-user.target
 		return nil, err
 	}
 
+	// because in kic base image we pre-install docker it already has a service file, it is harmless to double check with this step
```
tstromberg: I don't want harmless commands to be executed. Docker is a hermetic environment: either we need to run this command or not. Be clear that this has to be run to pick up the reconfigured docker file.

pkg/minikube/node/config.go
medyagh: done

pkg/minikube/node/config.go
```diff
@@ -156,6 +152,32 @@ func setupKubeconfig(h *host.Host, c *config.ClusterConfig, n *config.Node, clus
...
+	if driver.IsKIC(h.DriverName) {
+		p, err := oci.HostPortBinding(h.DriverName, h.Name, n.Port)
+		if err != nil {
+			return "", errors.Wrap(err, "host port binding")
+		}
+		// for kic drivers we use 127.0.0.1 instead of node IP,
+		// because of Docker on MacOs limitations for reaching to container's IP.
+		addr := fmt.Sprintf("https://" + net.JoinHostPort(oci.DefaultBindIPV4, strconv.Itoa(p)))
+		if cc.KubernetesConfig.APIServerName != constants.APIServerName {
+			addr = strings.Replace(addr, oci.DefaultBindIPV4, cc.KubernetesConfig.APIServerName, -1)
```
tstromberg: Please do not use string replacement. It makes the code more difficult to follow and adds room for bugs, as we've already seen. What do you think of something like:

```
hostname := h.Driver.GetIP()
port := n.Port

if driver.IsKIC() {
  hostname := oci.DefaultBindIPV4
  port := ...
}

// Warning: this may be an arbitrary string
if  cc.KubernetesConfig.APIServerName != constants.APIServerName {
  hostname =  cc.KubernetesConfig.APIServerName
}

return fmt.Sprintf("https://" + net.JoinHostPort(hostname, strconv.Itoa(port)))
```

pkg/minikube/node/config.go
```diff
@@ -156,6 +151,35 @@ func setupKubeconfig(h *host.Host, c *config.ClusterConfig, n *config.Node, clus
...
+	var err error
+	if driver.IsKIC(h.DriverName) {
+		// for kic drivers we use 127.0.0.1 instead of node IP,
+		// because of Docker on MacOs limitations for reaching to container's IP.
+		hostname = oci.DefaultBindIPV4
+		port, err = oci.HostPortBinding(h.DriverName, h.Name, port)
+		if err != nil {
+			return "", errors.Wrap(err, "host port binding")
+		}
+		if cc.KubernetesConfig.APIServerName != constants.APIServerName {
```
tstromberg: Can we drop this duplicate check? I just know it's going to get refactored out some day.

```
if driver.IsKIC() {
  hostname =
  port = 
} else {
  hostname = h.Driver.GetIP()
}

}
return fmt.Sprintf("https://" + net.JoinHostPort(hostname, strconv.Itoa(n.Port))), nil
```

Then you can eliminate the duplicate APIServerName check as well as the duplicate URL generation.

pkg/provision/ubuntu.go
```diff
@@ -155,6 +155,11 @@ WantedBy=multi-user.target
 		return nil, err
 	}
 
+	// because in kic base image we pre-install docker it already has a service file. we need to daemon-reload for the new systemd file
+	if err := p.Service("", serviceaction.DaemonReload); err != nil {
```
tstromberg: Can we also do this in buildroot.go?

Otherwise we'll run into a similar problem when we update Docker configs to be different than a previously run.

medyagh: ok :)

PR #7022, opened 2020-03-12
wip: Remove configureAuth on restarts

priyawadhwa:
It's unclear if this is actually needed, so checking if tests pass without it. This function is also somehow responsible for rewriting the docker port on restarts, so without it, that bug would be fixed.

fixes https://github.com/kubernetes/minikube/issues/7017

### DIFF
--- a/pkg/minikube/machine/fix.go
+++ b/pkg/minikube/machine/fix.go
@@ -36,7 +36,6 @@ import (
 	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/minikube/driver"
 	"k8s.io/minikube/pkg/minikube/out"
-	"k8s.io/minikube/pkg/util/retry"
 )
 
 // hostRunner is a minimal host.Host based interface for running commands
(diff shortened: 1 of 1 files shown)


medyagh:
I tried this on my Mac it works BEAUTIFULLY!

medyagh:
The TestStartStop/old-docker
seems that could possibly be related to this PR : 

https://storage.googleapis.com/minikube-builds/logs/7022/f08a172/KVM_Linux.html#fail_TestStartStop%2fgroup%2fold-docker

```
The W0312 17:34:55.766084   16154 exit.go:101] Failed to enable container runtime: enable docker.: sudo systemctl start docker: Process exited with status 5
stdout:
stderr:
Failed to start docker.service: Unit docker.service not found.
* 
X Failed to enable container runtime: enable docker.: sudo systemctl start docker: Process exited with status 5
stdout:
stderr:
Failed to start docker.service: Unit docker.service not found.
```

medyagh:
after this release, we should re-evaluate, the usage of configureAuth and see if we can avoid it !
the libmachine is slowing down our restarts.

PR #7041, opened 2020-03-13
status: error properly if cluster does not exist

tstromberg:
Old UI, with exit code 70:

```
$ minikube status
💣  getting config: cluster "minikube" does not exist

😿  minikube is exiting due to an error. If the above message 
```

New UI, with exit code 7 as per documentation:

```
💣  The "minikube" cluster does not exist!
```


medyagh:
I would like to see more Error Types of DoesntExists (for Service and other things) so we can handle them nicely

PR #7042, opened 2020-03-13
Refactor preload_images.go to function as an automated release script for preloaded tarballs

### LINKED ISSUE #6935: Create process for generating preload tar files
Requirements 
- we should not ever overwrite a tar bar that was tested/used with a release.
example, minikube v1.8.0 was tested with a specific tar file.
if we change something in the tar file in minikube v1.9.0, the people who are using minikube v1.8.0 should still be using the tar file that we tested with 1.8.0

- Automated script to geneate tar files for a list of kuberentes versions 
- Push only if something changed. 
for example if we generated for kuberntes 1.17.0 and we already have that tar file we should not re-generate and re-upload unless something in the tar file changed. (SHA probably wont work here, as time stamps in the tar might be different while actual files be same)

priyawadhwa:
The preload_images script will now get the last ~30 releases of k8s from github and make sure a preloaded tarball exists. If not, it will generate and upload it to the GCS bucket.

Users can still locally build a specific kubernetes version by running

```
go run hack/preload-images/*.go --kubernetes-version=vx.y.z
```

which I'll add to the docs.

The preload script will be run in Jenkins as a cron once a day.

fixes #6935


hack/preload-images/preload_images.go
```diff
@@ -42,123 +33,47 @@ const (
...
-func copyTarballToHost() error {
-	dest := filepath.Join("out/", tarballFilename)
-	cmd := exec.Command("docker", "cp", fmt.Sprintf("%s:/%s", profile, tarballFilename), dest)
-	cmd.Stdout = os.Stdout
-	if err := cmd.Run(); err != nil {
-		return errors.Wrapf(err, "cp cmd: %s", cmd.Args)
-	}
-	return nil
+func tarballFilename(kubernetesVersion string, containerRuntime string) string {
+	return fmt.Sprintf("preloaded-images-k8s-%s-%s-%s-%s.tar.lz4", preloadedTarballVersion, kubernetesVersion, containerRuntime, dockerStorageDriver)
```
medyagh: I suggest we add the architect for the images to the filename.
So that we can generate images for other archs like arm arm64,.... Or if users wanna use this script to generate their own.
( Additionally we will have to support allow user specificy their preload url )

medyagh: the arch needs to be added here.
return fmt.Sprintf("preloaded-images-k8s-%s-%s-%s-%s-%s.tar.lz4", preloadedTarballVersion, kubernetesVersion, containerRuntime, dockerStorageDriver, arch)

priyawadhwa: done, PTAL

PR #7044, opened 2020-03-13
Skip preload if not running on amd64

### LINKED ISSUE #7039: preload should not load wrong arch images
per comment in this issue: https://github.com/kubernetes/minikube/issues/5667#issuecomment-598524603

and someone tried to start on arm64 arch and minikube tried to load preloaded images for amd64 arch.

we need to make our tar files Per architecture, and if if there is no preload for arm64 we should not load wrong architecture

we should also provide a flag, --preload=false

priyawadhwa:
Also make sure that preload is available before attempting to run preload.

Should fix #7039


medyagh:
Looks good after the rebase. Thank you for this PR.

medyagh:
> Average minikube: **68.844270**
> Average Minikube (PR 7044): **105.284784**


@priyawadhwa  perforamnce bot numbers look a bit scary

priyawadhwa:
@medyagh good catch -- I was using `runtime.GOOS` when I should have been using `runtime.GOARCH`. This should be fixed now!

tstromberg:
rebase?

PR #7043, opened 2020-03-13
Delete: output underlying status failure, better machine function names

tstromberg:
## Old UI if trying to delete a cluster while Docker is down:

```
❗  Unable to get the status of the minikube cluster.
🔥  Removing /Users/tstromberg/.minikube/machines/minikube ...
💀  Removed all traces of the "minikube" cluster.
```

It shows a warning, but no information as to why the status failed!

## New UI if trying to delete a cluster while Docker is down:

```
❗  Unable to get host status for "minikube": state: docker inspect -f {{.State.Status}} minikube: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?: exit status 1
🔥  Removing /Users/tstromberg/.minikube/machines/minikube ...
💀  Removed all traces of the "minikube" cluster.
```

This also removes some technical debt discovered while trying to improve this message:

* machine.DeleteHost now displays underlying status error in warning
* kic.GetState error now includes command-line and output
* Remove unused machine.List function
* Renamed machine.CheckIfHostExistsAndLoad to machine.LoadHost
* Renamed machine.Load to machine.LoadMachine (repetitive, but clearer vs LoadHost)
* Renamed machine.GetHostStatus to machine.Status
* Renamed machine.IsHostRunning to machine.IsRunning
* Moved CreateSSHShell from machine.go to ssh.go

PR #7049, opened 2020-03-14
Remove now unused machineDirs function

afbjorklund:
Hopefully fixing the broken golangci-lint (since PR #7043 and 196a8d7b8201e0931fce10c38b515cdeec957223)


tstromberg:
Thank you for catching this!

Issue #7067, opened 2020-03-16
Add --vm flag for users who want to autoselect only VM's


tstromberg:
Some users may need to use kernel modules.

tstromberg:
Just in case someone else has time to address this, here is where I would start:

* Add "vm" boolean flag here: https://github.com/kubernetes/minikube/blob/82900a7811e479755e03ae020ebb44152dfd93d8/cmd/minikube/cmd/start.go#L196

* Add option to `driver.Choices` function to filter for drivers which `IsVM()` is true:  https://github.com/kubernetes/minikube/blob/82900a7811e479755e03ae020ebb44152dfd93d8/pkg/minikube/driver/driver.go#L167

prasadkatti:
Hi. I would like to work on this issue.

prasadkatti:
Looks like there is a PR out already. I will unassign myself.

PR #7068, opened 2020-03-17
Add --vm flag for users who want to autoselect only VM's

vikkyomkar:
#7067

### DIFF
--- a/pkg/minikube/driver/driver.go
+++ b/pkg/minikube/driver/driver.go
@@ -165,26 +165,22 @@ func FlagDefaults(name string) FlagHints {
 
 // Choices returns a list of drivers which are possible on this system
 func Choices(vm bool) []registry.DriverState {
+	var drivers []registry.DriverState
 	options := registry.Available()
 	if vm {
-		var vmOptions []registry.DriverState 
 		for _, ds := range options {
 			if IsVM(ds.Name) {
-				vmOptions = append(vmOptions,ds)
+				drivers = append(drivers, ds)
 			}
 		}
-		// Descending priority for predictability and appearance
-                sort.Slice(vmOptions, func(i, j int) bool {
-                        return vmOptions[i].Priority > vmOptions[j].Priority
-		})
-		return vmOptions
-	}else {
-		// Descending priority for predictability and appearance
-		sort.Slice(options, func(i, j int) bool {
-			return options[i].Priority > options[j].Priority
-		})
-		return options
+	} else {
+		drivers = options
 	}
+	// Descending priority for predictability and appearance
+	sort.Slice(options, func(i, j int) bool {
+		return options[i].Priority > options[j].Priority
+	})
+	return drivers
 }
 
 // Suggest returns a suggested driver from a set of options
--- a/pkg/minikube/driver/driver.go
+++ b/pkg/minikube/driver/driver.go
@@ -164,14 +164,27 @@ func FlagDefaults(name string) FlagHints {
 }
 
 // Choices returns a list of drivers which are possible on this system
-func Choices() []registry.DriverState {
+func Choices(vm bool) []registry.DriverState {
 	options := registry.Available()
-
-	// Descending priority for predictability and appearance
-	sort.Slice(options, func(i, j int) bool {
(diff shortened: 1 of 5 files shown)


pkg/minikube/driver/driver.go
```diff
@@ -164,14 +164,23 @@ func FlagDefaults(name string) FlagHints {
...
 
 // Choices returns a list of drivers which are possible on this system
-func Choices() []registry.DriverState {
+func Choices(vm bool) []registry.DriverState {
+	var drivers []registry.DriverState
 	options := registry.Available()
-
+	if vm {
+		for _, ds := range options {
+			if IsVM(ds.Name) {
```
tstromberg: Suggestion: Now that I see how this looks, rather than applying the filter in the Choices function, could you pass the value to `registry.Available` and add the filtering in that function instead?

https://github.com/kubernetes/minikube/blob/07e8420c2ab57de73da61407e5dff777552b957b/pkg/minikube/registry/global.go#L62

This would be a slight performance improvement, as `registry.Available()` would not have to call`driver.Status()` on non-VM drivers if the result is going to be ignored anyways.

vikkyomkar: sure I will make the suggested changes soon

tstromberg:
Need any help with this PR?

pkg/minikube/driver/driver.go
vikkyomkar: > Need any help with this PR?

@tstromberg I have made the changes. I had to write **IsVM** function separately in registry folder to avoid "import cycle error" b/w driver and registry.

vikkyomkar:
> Need any help with this PR?

@tstromberg 
I have made the changes. I had to write **IsVM** function separately in registry folder to avoid "import cycle error" b/w driver and registry.

pkg/minikube/registry/global.go
```diff
@@ -76,7 +110,13 @@ func Available() []DriverState {
 			priority = Unhealthy
 		}
 
-		sts = append(sts, DriverState{Name: d.Name, Priority: priority, State: s})
+		if vm {
```
tstromberg: With go, we try to avoid if/else statements when possible, particularly in a loop. Kind of along the lines of https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

This would be clearer:

```
if vm && !IsVM(d.Name) {
  continue
}
```

pkg/minikube/registry/global.go
```diff
@@ -59,7 +93,7 @@ func Driver(name string) DriverDef {
 }
 
 // Available returns a list of available drivers in the global registry
-func Available() []DriverState {
+func Available(vm bool) []DriverState {
```
tstromberg: this might be clearer if it was named `vmOnly`

pkg/minikube/registry/global.go
```diff
@@ -24,6 +24,40 @@ import (
...
+}
+
+// IsMock checks if the driver is a mock
+func IsMock(name string) bool {
+	return name == Mock
+}
+
+// IsVM checks if the driver is a VM
+func IsVM(name string) bool {
+	if IsKIC(name) || IsMock(name) || BareMetal(name) {
```
tstromberg: I know you just copy and pasted the previous function, but because the other functions aren't used, just inline it:

```
if name == Docker || name == Mock || name == None || name == Podman {
  return true
}

Then you can remove IsKic and IsMock.
```

pkg/minikube/registry/global.go
```diff
@@ -24,6 +24,40 @@ import (
...
+	return name == Docker || name == Podman
+}
+
+// IsMock checks if the driver is a mock
+func IsMock(name string) bool {
+	return name == Mock
+}
+
+// IsVM checks if the driver is a VM
+func IsVM(name string) bool {
```
tstromberg: rename this to `isVM` -- it does not need to be made public.

tstromberg:
Heads up: I plan to merge this PR in it's current state today, so that v1.9.0 users can enjoy it. 

Thank you again for contributing this.

PR #7076, opened 2020-03-17
Use Kubernetes v1.18.0-rc.1 by default

tstromberg:


tstromberg:
The docker `TestVersionUpgrade` test results look concerning. Re-running locally.

PR #7081, opened 2020-03-18
WIP: Add SysV init support

tstromberg:
My slightly crazy attempt to support SysV init systems like Upstart and OpenRC.

I only did this because I wanted to try minikube on Cloud Run.


afbjorklund:
Are those systems also missing the standard `service` wrapper ?

tstromberg:
Obsolete.

PR #7090, opened 2020-03-18
Upgrade Docker driver to preferred (Linux), default on other platforms

tstromberg:
Docker driver seems pretty stable and usable now.

Still have some quirks to work out around disk space management and the service command.


sharifelgamal:
nice

PR #6872, opened 2020-03-03
wip: Fix DownloadOnlyDocker test on github actions

priyawadhwa:
Fixes #6869

PR #7094, opened 2020-03-18
Change download test to use unsupported preload version

### LINKED ISSUE #7085: TestDownloadOnly/group/v1.17.3  failing: expected image file exist at...
I think this is related to preload. Do you mind taking a look, @priyawadhwa ?

This test may be need to be updated to check for preloadable versions differently than others.

priyawadhwa:
So that we can still make sure that images are being pulled properly.

Also, remove the check for the kic base image. Right now, it isn't passing, and is increasing test time to >70 minutes. I'm working on a fix in another PR. (#6872 )

Should fix #7085

### DIFF
--- a/test/integration/aaa_download_only_test.go
+++ b/test/integration/aaa_download_only_test.go
@@ -51,7 +51,8 @@ func TestDownloadOnly(t *testing.T) {
 	t.Run("group", func(t *testing.T) {
 		versions := []string{
 			constants.OldestKubernetesVersion,
-			"v1.12.0",
+			constants.DefaultKubernetesVersion,
+			constants.NewestKubernetesVersion,
 		}
 		for _, v := range versions {
 			t.Run(v, func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


test/integration/aaa_download_only_test.go
```diff
@@ -54,8 +51,7 @@ func TestDownloadOnly(t *testing.T) {
 	t.Run("group", func(t *testing.T) {
 		versions := []string{
 			constants.OldestKubernetesVersion,
-			constants.DefaultKubernetesVersion,
-			constants.NewestKubernetesVersion,
+			"v1.12.0",
```
tstromberg: We need to download these versions to assert offline access in a later check.

How about just adding v1.12.0, and changing the test to check if there is a preload available, and altering it so that it checks for a preload file instead of stat'ing for docker image files?

priyawadhwa: That makes sense, I ended up removing v1.12.0 because there is not preloaded tarball for oldest version right now.

PR #7110, opened 2020-03-19
Only run deleteHosts if running a VM

### LINKED ISSUE #7097: delete: docker inspect -f {{.State.Status}} minikube: Error: No such object: minikube: exit status 1
If you start a Docker VM:

```
% ./out/minikube start --driver=docker                                                                                                                   😄  minikube v1.9.0-beta.1 on Darwin 10.15.3
🔥  Creating Kubernetes in docker container with (CPUs=2) (6 available), Memory=3889MB (3937MB available) ...
🐳  Preparing Kubernetes v1.18.0-rc.1 on Docker 19.03.2 ...
🚀  Launching Kubernetes ... 
⌛  Waiting for cluster to come online ...

❗  /usr/local/bin/kubectl is v1.16.3, which may be incompatible with Kubernetes v1.18.0-rc.1.
💡  You can also use 'minikube kubectl -- get pods' to invoke a matching version
```

Deletion shows a scary error message:

```
%./out/minikube delete --all                                                                                                                                   
❗  Unable to get host status for "minikube": state: docker inspect -f {{.State.Status}} minikube: Error: No such object: minikube: exit status 1
🔥  Removing /Users/tstromberg/.minikube/machines/minikube ...
```

priyawadhwa:
If minikube is started with the docker driver, all containers and volumes will be deleted in the call to `deleteProfileContainersAndVolumes`. If we then try to call `deleteHosts` on it, we get this warning:

```
❗  Unable to get host status for "minikube": state: docker inspect -f {{.State.Status}} minikube: Error: No such object: minikube: exit status 1
```

which will always happen because we've already deleted the container.

cc @medyagh does this seem reasonable to you?

Should fix #7097

### DIFF
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -239,9 +239,7 @@ func deleteProfile(profile *config.Profile) error {
 		out.T(out.FailureType, "Failed to kill mount process: {{.error}}", out.V{"error": err})
 	}
 
-	if driver.IsVM(cc.Driver) {
-		deleteHosts(api, cc)
-	}
+	deleteHosts(api, cc)
 
 	// In case DeleteHost didn't complete the job.
 	deleteProfileDirectory(profile.Name)
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -239,7 +239,9 @@ func deleteProfile(profile *config.Profile) error {
 		out.T(out.FailureType, "Failed to kill mount process: {{.error}}", out.V{"error": err})
 	}
 
-	deleteHosts(api, cc)
+	if driver.IsVM(cc.Driver) {
+		deleteHosts(api, cc)
+	}
 
 	// In case DeleteHost didn't complete the job.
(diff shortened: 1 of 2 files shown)


cmd/minikube/cmd/delete.go
```diff
@@ -239,7 +239,9 @@ func deleteProfile(profile *config.Profile) error {
 		out.T(out.FailureType, "Failed to kill mount process: {{.error}}", out.V{"error": err})
 	}
 
-	deleteHosts(api, cc)
+	if driver.IsVM(cc.Driver) {
```
tstromberg: This doesn't cover the `none` case, so perhaps this should be `if !driver.IsKIC`.

It feels weird to special case KIC here, so I'm OK with calling `delete` twice, but so long as the error isn't displayed. I feel like it should be able to interpret the lack of container as "nothing to do" rather than outputting a warning message.

priyawadhwa: sounds good -- I changed it so that if we get an error when getting host status, we assume that the host no longer exists and return

PR #7112, opened 2020-03-19
Improve error when docker-env is used with non-docker runtime

### LINKED ISSUE #7111: docker-env with crio: Error getting service status: Process exited with status 3

Make sure you're running Docker on Mac and run in a terminal:

```bash
$ minikube start --cpus=4 --memory=4g --container-runtime=cri-o --vm-driver=virtualbox
$ minikube docker-env
```

I guess it is due to the use of VirtualBox? With hyperkit _docker-env_ works, but hyperkit has some issues with the network.

```
$ minikube docker-env

💣  Error getting service status: Process exited with status 3

```
</details>

There is none regarding this issue.

macOS 10.14.6
Docker 19.03.8
minikube 1.8.2
VirtualBox 6.0.18

tstromberg:
Fixes #7111

## Old UI

```
💣  Docker runtime check failed: sudo systemctl is-active docker failed: Process exited with status 3
output: "inactive"

```

## New UI

`💣  The docker-env command is only compatible with the "docker" runtime, but this cluster was configured to use the "cri-o" runtime.`

### DIFF
--- a/cmd/minikube/cmd/docker-env.go
+++ b/cmd/minikube/cmd/docker-env.go
@@ -122,13 +122,18 @@ func isDockerActive(d drivers.Driver) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	output, err := client.Output("sudo systemctl is-active docker")
-	if err != nil {
-		return false, err
-	}
-	// systemd returns error code on inactive
+	cmd := "sudo systemctl is-active docker"
+
+	output, err := client.Output(cmd)
 	s := strings.TrimSpace(output)
-	return err == nil && s == "active", nil
+
+	if err != nil {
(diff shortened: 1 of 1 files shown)


afbjorklund:
Looks like this got broken with the `cruntime` introduction ?

Old check:

```go
func GetDockerActive(host *host.Host) (bool, error) {
       statusCmd := `sudo systemctl is-active docker`
       status, err := host.RunSSHCommand(statusCmd)
       // systemd returns error code on inactive
       s := strings.TrimSpace(status)
       return err == nil && s == "active", err
}

```
New check:

```go
// Active returns if docker is active on the host
func (r *Docker) Active() bool {
       err := r.Runner.Run("systemctl is-active --quiet service docker")
       return err == nil
}
```

The systemd error codes are somewhat flawed.

PR #7115, opened 2020-03-20
provisioner: only reload docker if necessary, don't install curl

### LINKED ISSUE #7103: docker: excessive systemctl reload/restarts
I noticed this while running `minikube start` on a running instance:

```
I0318 13:50:11.255632   48225 main.go:110] libmachine: About to run SSH command:
sudo systemctl -f daemon-reload 
I0318 13:50:11.533986   48225 main.go:110] libmachine: About to run SSH command:
sudo systemctl -f enable docker
I0318 13:50:11.794742   48225 main.go:110] libmachine: About to run SSH command:
sudo systemctl daemon-reload
I0318 13:50:12.073229   48225 main.go:110] libmachine: About to run SSH command:
sudo systemctl -f restart docker
```

This should be sufficient, though possibly not needed either:

`sudo systemctl -f daemon-reload && sudo systemctl -f restart docker`

tstromberg:
This modernizes the provisioner in three important ways:

* Only calls systemctl reload/restart if docker file has been updated - fixes #7103
* Adds it's own `NewSystemdProvisioner` to avoid unnecessary `apt-get` commands to install curl.
* Uses `glog` for consistency with rest of minikube codebase

This PR improves "minikube start" performance on an already running cluster by a second or so.

### DIFF
--- a/pkg/minikube/machine/client.go
+++ b/pkg/minikube/machine/client.go
@@ -172,6 +172,7 @@ func (api *LocalClient) Create(h *host.Host) error {
 		return fmt.Errorf("driver %q does not exist", h.DriverName)
 	}
 	if def.Init == nil {
+		// NOTE: This will call provision.DetectProvisioner
 		return api.legacyClient.Create(h)
 	}
 
--- a/pkg/provision/buildroot.go
+++ b/pkg/provision/buildroot.go
@@ -19,18 +19,16 @@ package provision
 import (
 	"bytes"
 	"fmt"
-	"path"
 	"text/template"
 	"time"
 
 	"github.com/docker/machine/libmachine/auth"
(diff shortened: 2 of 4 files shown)


afbjorklund:
The "curl" package is being installed, because the default install command is `curl -sSL`

```go
func installDockerGeneric(p Provisioner, baseURL string) error {
	// install docker - until cloudinit we use ubuntu everywhere so we
	// just install it using the docker repos
	if output, err := p.SSHCommand(fmt.Sprintf("if ! type docker; then curl -sSL %s | sh -; fi", baseURL)); err != nil {
		return fmt.Errorf("error installing docker: %s", output)
	}

	return nil
}
```

However, the `curl` package is a part of the regular Ubuntu cloud images nowadays.

PR #7116, opened 2020-03-20
Simplify provisioning logic, add fast provisioner detection

tstromberg:
Speeds up host restart by a second or so.

* Remove provision.DetectProvisioner cal, it isn't necessary.
* Only call provision.Provision if docker options have changed
* Remove call to ConfigureAuth, which just calls provision.Provision


tstromberg:
Looks like I have some unit tests to fix up.

tstromberg:
The `TestVersionUpgrade` result is suspicious - investigating.

PR #7117, opened 2020-03-20
exit earlier if the driver is invalid

sharifelgamal:
Before:
```
sudo out/minikube start --driver=none
😄  minikube v1.9.0-beta.1 on Darwin 10.14.6
✨  Using the  driver based on user configuration
💣  The driver <no value> '' is not supported on darwin
```

After:
```
sudo out/minikube start --driver=none
😄  minikube v1.9.0-beta.1 on Darwin 10.14.6
💣  The driver 'none' is not supported on darwin
```

### DIFF
--- a/pkg/minikube/driver/driver.go
+++ b/pkg/minikube/driver/driver.go
@@ -19,12 +19,15 @@ package driver
 import (
 	"fmt"
 	"os"
+	"runtime"
 	"sort"
 	"strings"
 
 	"github.com/golang/glog"
 	"k8s.io/minikube/pkg/drivers/kic"
 	"k8s.io/minikube/pkg/minikube/config"
+	"k8s.io/minikube/pkg/minikube/exit"
+	"k8s.io/minikube/pkg/minikube/out"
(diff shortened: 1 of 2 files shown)


pkg/minikube/driver/driver.go
```diff
@@ -214,6 +217,9 @@ func Suggest(options []registry.DriverState) (registry.DriverState, []registry.D
 // Status returns the status of a driver
 func Status(name string) registry.DriverState {
 	d := registry.Driver(name)
+	if d.Empty() {
+		exit.WithCodeT(exit.Unavailable, "The driver '{{.driver}}' is not supported on {{.os}}", out.V{"driver": name, "os": runtime.GOOS})
```
tstromberg: I don't love having libraries panic - is it possible to move this logic closer to the `main` package?

Alternatively, could this return a driver state that represents the issue and exits sooner in main?

sharifelgamal: moving it to main was easy enough, done

PR #7132, opened 2020-03-21
Update Nvidia GPU plugin

laozc:
The current plugin used by minikube is too old without properly setting up NVIDIA_VISIBLE_DEVICES for GPU isolation.
Update Nvidia device plugin to latest release to resolve this issue.


tstromberg:
Thank you!

PR #7141, opened 2020-03-22
Fix node name for "none" driver

laozc:
Recent changes break the host name for "none" driver and the node always falls back to `minikube`.
This makes sure that the host name is picked for node name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,41 @@
 # Release Notes
 
+## Version 1.9.0-beta.2 - 2020-03-21
+
+New features & improvements
+
+* 🎉 Experimental multi-node support 🎊 [#6787](https://github.com/kubernetes/minikube/pull/6787)
+* Add kubectl desc nodes to minikube logs [#7105](https://github.com/kubernetes/minikube/pull/7105)
+* bumpup helm-tiller v2.16.1 → v2.16.3 [#7130](https://github.com/kubernetes/minikube/pull/7130)
+* Update Nvidia GPU plugin [#7132](https://github.com/kubernetes/minikube/pull/7132)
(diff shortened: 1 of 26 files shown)


sharifelgamal:
`node.Name()` is already taken, but I agree that it should probably live in the node package. `node.InternalName()` perhaps? 

I'm ideally doing away with the need for triple hyphens in the machine name soon, but double hyphens is plenty enough for the time being.

laozc:
I believe `minikube---m01` could be a valid host name so it seems like using `---` as a separator is not a good idea.
Considering if the node is naming as `minikube---m01` it is possible with the `none` driver (as it is controlled by the user who creates the machine) but the logic to only splitting by `---` would break it.
It would be better to use `:` or other illegal character for hostname as separator for cluster name/node name.

sharifelgamal:
currently hostname and nodename are coupled, a good solution would be to decouple them then use a weird character as a separator. the best solution is clearly to not rely on hostname to re-extract the cluster name but that's outside the scope of this PR.

laozc:
Let me just move the `KubeNodeName` out of driver package for now.

medyagh:
I created an issue for this PR https://github.com/kubernetes/minikube/issues/7169

medyagh:
> I believe `minikube---m01` could be a valid host name so it seems like using `---` as a separator is not a good idea.
> Considering if the node is naming as `minikube---m01` it is possible with the `none` driver (as it is controlled by the user who creates the machine) but the logic to only splitting by `---` would break it.
> It would be better to use `:` or other illegal character for hostname as separator for cluster name/node name.

I find three dashes odd, I would vote for not having 3 dashes for names

sharifelgamal:
yeah let's use two dashes for now, I will work on no longer relying on embedding the cluster name in the machine name

PR #7149, opened 2020-03-22
Do not run GPU plugin under priviledge mode

laozc:
A follow-up PR with #7132 to correct securityContext for the updated GPU device plugin and host path mount on `/dev` (which is not needed any longer as it is done by `nvidia-container-runtime`).


tstromberg:
Thank you!

PR #7153, opened 2020-03-22
Use --driver instead of --vm-driver in GH workflow

prasadkatti:


medyagh:
I think this was intentional, for one of the github actions to use the old format, so in case that one breaks, we can see it in the tests

prasadkatti:
oh ok. cool. I will close the PR then.

PR #7160, opened 2020-03-23
hyperv Delete: call StopHost before removing VM

### LINKED ISSUE #6804: delete: The process cannot access the file because it is being used by another process
```
minikube delete
```
```
! Unable to get the status of the minikube cluster.
* Removing C:\Users\Csongor Halmai\.minikube\machines\minikube ...
*
X Unable to remove machine directory: %v: remove C:\Users\Csongor Halmai\.minikube\machines\minikube\disk.vhd: The process cannot access the file because it is being used by another process.
*
  - https://github.com/kubernetes/minikube/issues/new/choose
```
</details>

```
*
X api load: filestore "minikube": open C:\Users\Csongor Halmai\.minikube\machines\minikube\config.json: The system cannot find the file specified.
*
  - https://github.com/kubernetes/minikube/issues/new/choose
```
</details>

Microsoft Windows [Version 10.0.18362.592]

I hoped that creating a config.json file can solve the problem. I created one with 0 bytes which was simply deleted by  `minikube delete`. Then I created one with the content `{}` but it caused `minikube delete` to die with `panic: runtime error: invalid memory address or nil pointer dereference`. Then I gave up.

tstromberg:
Fixes #6804 - a bug I ran into this morning while running "minikube start/minikube delete" in an infinite loop.

Just in case, it also tries driver deletion twice, since it gets messy if we later delete the machine profile but the VM deletion never happened.

### DIFF
--- a/pkg/minikube/machine/delete.go
+++ b/pkg/minikube/machine/delete.go
@@ -75,18 +75,26 @@ func DeleteHost(api libmachine.API, machineName string) error {
 		return mcnerror.ErrHostDoesNotExist{Name: machineName}
 	}
 
-	// This is slow if SSH is not responding, but HyperV hangs otherwise, See issue #2914
+	// Hyper-V requires special care to avoid ACPI and file locking issues
 	if host.Driver.DriverName() == driver.HyperV {
-		if err := trySSHPowerOff(host); err != nil {
-			glog.Infof("Unable to power off minikube because the host was not found.")
+		if err := StopHost(api, machineName); err != nil {
+			glog.Warningf("stop host: %v", err)
 		}
-		out.T(out.DeletingHost, "Successfully powered off Hyper-V. minikube driver -- {{.driver}}", out.V{"driver": host.Driver.DriverName()})
+		// Hack: give the Hyper-V VM more time to stop before deletion
+		time.Sleep(1 * time.Second)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/minikube/machine/delete.go
```diff
@@ -75,18 +75,26 @@ func DeleteHost(api libmachine.API, machineName string) error {
...
 		}
-		out.T(out.DeletingHost, "Successfully powered off Hyper-V. minikube driver -- {{.driver}}", out.V{"driver": host.Driver.DriverName()})
+		// Hack: give the Hyper-V VM more time to stop before deletion
+		time.Sleep(1 * time.Second)
 	}
 
 	out.T(out.DeletingHost, `Deleting "{{.profile_name}}" in {{.driver_name}} ...`, out.V{"profile_name": machineName, "driver_name": host.DriverName})
 	if err := host.Driver.Remove(); err != nil {
-		return errors.Wrap(err, "host remove")
+		glog.Warningf("remove failed, will retry: %v", err)
```
medyagh: nit: how about using retry.Expo ( and set the max number of retry to 2 or 3)

tstromberg: I considered it, but feel like it may provide an awkward experience to retry more than once. Will reconsider based on user reports.

PR #7173, opened 2020-03-23
stop kubernetes containers on stopping docker/podman driver

medyagh:
hopefully closes 
```
TestVersionUpgrade: bind: address already in use
```
https://github.com/kubernetes/minikube/issues/7171

also uses -f to stop 

closes https://github.com/kubernetes/minikube/issues/7187


medyagh:
none: TestVersionUpgrade 'exit 0' : driver does not support ssh command
is not related to this PR
https://github.com/kubernetes/minikube/issues/7190

medyagh:
TestVersionUpgrade - for docker driver is expected to fail, because the latest stable minikube does not stop the containers correctly.

Issue #7168, opened 2020-03-23
none driver: service fail exec: "xdg-open": executable file not found in $PATH


medyagh:
on a brand new ubuntu 18.04
```
$ minikube service hello-minikube
|-----------|----------------|-------------|--------------------------|
| NAMESPACE |      NAME      | TARGET PORT |           URL            |
|-----------|----------------|-------------|--------------------------|
| default   | hello-minikube |             | http://10.128.0.21:31284 |
|-----------|----------------|-------------|--------------------------|
🎉  Opening service default/hello-minikube in default browser...
💣  open url failed: http://10.128.0.21:31284: exec: "xdg-open": executable file not found in $PATH
😿  minikube is exiting due to an error. If the above message is not useful, open an issue:
👉  
```

afbjorklund:
Supposed to be provided by `xdg-utils`, in any normal desktop environment...

https://github.com/pkg/browser/blob/master/browser_linux.go#L6

afbjorklund:
I suppose you _could_ detect a broken desktop, and default to printing the URL ?

https://www.freedesktop.org/wiki/Software/xdg-utils/

PR #7174, opened 2020-03-23
Centralize the handling of browser.OpenURL

afbjorklund:
So we can catch when xdg-open not installed

Might help with #7168

### DIFF
--- a/pkg/minikube/browser/browser.go
+++ b/pkg/minikube/browser/browser.go
@@ -1,0 +1,35 @@
+/*
+Copyright 2020 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


afbjorklund:
Right, have to come up with a nice way of handling this.
Maybe just a text saying "please open a browser at ..."

afbjorklund:
Went with emoji instead

pkg/minikube/browser/browser.go
```diff
@@ -0,0 +1,37 @@
...
+	"k8s.io/minikube/pkg/minikube/out"
+)
+
+// OpenURL opens a new browser window pointing to URL.
+func OpenURL(url string) error {
+	if runtime.GOOS == "linux" {
+		_, err := exec.LookPath("xdg-open")
+		if err != nil {
+			out.T(out.URL, url)
+			return nil
```
medyagh: how about some additional wording to the URL ?
how does this look like in the output ?

medyagh:
> Went with emoji instead

do u mind pasting the ouput in the PR description so we see how it looks like?

afbjorklund:
I don't have the environment set up to reproduce it, but I _guess_ it would look something like:

```
🎉  Opening service default/hello-minikube in default browser...
👉  http://10.128.0.21:31284
```

Issue #7128, opened 2020-03-21
Update commands to detect non-existent configuration


zhangfeifeifei:
**minikube stop**：

*** 
X Error getting cluster config: cluster "minikube" does not exist
* 
* minikube is exiting due to an error. If the above message is not useful, open an issue:
  - https://github.com/kubernetes/minikube/issues/new/choose**：<details>

</details>

**`minikube logs`命令的输出**： <details>

</details>

**使用的操作系统版本**：

tstromberg:
Thank you for the bug report! The issue is because there is no cluster to be found. I'll use this issue to track updating our commands to give better output.

Issue #7066, opened 2020-03-16
none start: cluster "minikube" does not exist


mw-poly:
**The exact command to reproduce the issue**:

**The full output of the command that failed**:<details>

</details>

**The output of the `minikube logs` command**:<details>

</details>

**The operating system version**:

tstromberg:
I don’t yet have a clear way to replicate this issue. Do you mind adding some additional details? Here is additional information that would be helpful:

* The exact command which showed this error message
*  The exact `minikube start` command line used
* The full output of both commands

Thank you for sharing your experience!

YannikSc:
Having the same issue. The `minikube logs` command just outputs more or less the same error message.
- Running on Archlinux
- Installed minikube via pacman, version 1.8.2-1
- Run `sudo minikube --driver=none start`

<details>
<summary>Issuing `minikube status` outputs</summary>

    💣  getting config: cluster "minikube" does not exist
    
    😿  minikube is exiting due to an error. If the above message is not useful, open an issue:
    👉  https://github.com/kubernetes/minikube/issues/new/choose
</details>

<details>
<summary>`minikube logs`</summary>

    💣  Error getting config: cluster "minikube" does not exist
    
    😿  minikube is exiting due to an error. If the above message is not useful, open an issue:
    👉  https://github.com/kubernetes/minikube/issues/new/choose
</details>

medyagh:
@YannikSc do you mind trying
sudo minikube delete
sudo minikube --driver=none --alsologtostderr -v8

and paste the output of the above command ?


alternatively i am curious, have you tried our newst docker driver ? that has better perforamnce and stalibty on linux ?

minikube start --driver=docker ?

YannikSc:
Well .. surprisingly the delete command worked ..
the second minikube command failed with the error "unknown flag --driver" so I assumed, you mean `sudo minikube start  [...]` which results in an error:

<details>
<summary>`sudo minikube start --driver=none --alsologtostderr -v8`</summary>

    I0418 01:16:24.885531  223439 notify.go:125] Checking for updates...
[...]

Issue #6814, opened 2020-02-27
Minikube status shows me error


lmurali89:
I tried to check my minikube status and i got the error as shown below.

C:\WINDOWS\system32>minikube status
E0227 12:31:23.986692   11016 status.go:95] status error: host: load: filestore "minikube": open C:\Users\Hp\.minikube\machines\minikube\config.json: The system cannot find the file specified.
E0227 12:31:23.991691   11016 status.go:98] The "minikube" cluster does not exist!
host: Nonexistent
kubelet: Nonexistent
apiserver: Nonexistent
kubeconfig: Nonexistent

priyawadhwa:
Hey @lmurali89 -- 

could you try running:

```shell
$ minikube delete
$ minikube start
$ minikube status
```

and see if that works?

tstromberg:
Fixed in v1.9.2

PR #7176, opened 2020-03-23
Update commands to consistently detect down & paused clusters

### LINKED ISSUE #7150: minikube service hello-minikube

</details>

</details>

tstromberg:
This brings all of the commands together to consistently load clusters in a way that we can deliver a friendly message. 

Also fixes minor bits of technical debt:

* consistent naming of imports
* consistent naming of cluster name
* adds `ClusterFlagValue()` function

Fixes #7150 #7128 #7127 #7066 #6814 #5807 

# Examples

## stopping a nonexistent cluster

Was:

```

💣  Error getting cluster config: cluster "minikube" does not exist

```

Now:

```
🤷  There is no local cluster named "minikube"
👉  To fix this, run: minikube start
```

## pausing a stopped cluster

Was:

```

💣  Pause: kubelet disable: disable: sudo systemctl disable kubelet: exit status 1
stdout:

stderr:
Error response from daemon: Container 9946e6e75d6826978a1323785252010ebb8e3cead58b4fec52e813de3398444a is not running


```

Now:

```
👉  To fix this, run: minikube -p xyzzy start
```

## service on a paused cluster

Was:

```
💣  error starting tunnel: Service hello-node was not found in "default" namespace. You may select another namespace by using 'minikube service hello-node -n <namespace>: Get https://127.0.0.1:32774/api/v1/namespaces/default/services/hello-node?timeout=1s: context deadline exceeded

```

Now:

```
🤷  The control plane for "xyzzy" is paused!
👉  To fix this, run: minikube -p xyzzy unpause
```

## Testing

Tested with:

```
status
./out/minikube stop
dashboard
pause
./out/minikube unpause
docker-env
podman-env
cache add moo
addons
config
profile
update-context
service xxx
tunnel
mount /tmp:/tmp
ssh
kubectl
ssh-key
ip
logs
update-check
version
options
completion
delete
```

### DIFF
--- a/cmd/minikube/cmd/flags.go
+++ b/cmd/minikube/cmd/flags.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2020 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package cmd
 
 import (
--- a/cmd/minikube/cmd/flags.go
+++ b/cmd/minikube/cmd/flags.go
@@ -1,0 +1,23 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/viper"
+	"k8s.io/minikube/pkg/minikube/config"
+	"k8s.io/minikube/pkg/minikube/constants"
+)
+
+// Return a minikube command containing the current profile name
+func minikubeCmd() string {
+	cname := ClusterFlagValue()
+	if cname != constants.DefaultClusterName {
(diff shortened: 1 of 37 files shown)


pkg/minikube/mustload/mustload.go
```diff
@@ -0,0 +1,169 @@
...
+		out.T(out.Warning, `This is unusual - you may want to investigate using "{{.command}} logs"`, out.V{"command": minikubeCmd(name)})
+		exitTip("start", name, exit.Unavailable)
+	}
+	return co
+}
+
+// Return a minikube command containing the current profile name
+func minikubeCmd(cname string) string {
+	if cname != constants.DefaultClusterName {
+		return fmt.Sprintf("minikube -p %s", cname)
```
medyagh: not sure if -p is better before or after start
```
minikube -p p1 start
```
or 
```
minikube start -p p1
```
but we need to be consistant in all minikube, in docker-env we use -p after. 
so either make this like others, or make others like this.

cmd/minikube/cmd/flags.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"fmt"
+
+	"github.com/spf13/viper"
+	"k8s.io/minikube/pkg/minikube/config"
+	"k8s.io/minikube/pkg/minikube/constants"
+)
+
+// Return a minikube command containing the current profile name
+func minikubeCmd() string {
```
medyagh: this seems to be duplicate code from mustlod. package'

PR #7177, opened 2020-03-23
Fix output of ssh command test

### LINKED ISSUE #7170: TestFunctional/parallel/SSHCmd: "hello\n", want = "hello\r\n"
as seen in 
https://storage.googleapis.com/minikube-builds/logs/7165/ba4e134/Docker_Linux.html#fail_TestFunctional%2fparallel%2fSSHCmd

also in https://storage.googleapis.com/minikube-builds/logs/7165/ba4e134/KVM_Linux.html#fail_TestFunctional%2fparallel%2fSSHCmd
```
=== RUN   TestFunctional/parallel/SSHCmd
=== PAUSE TestFunctional/parallel/SSHCmd
=== CONT  TestFunctional/parallel/SSHCmd
--- FAIL: TestFunctional/parallel/SSHCmd (0.39s)
functional_test.go:702: (dbg) Run:  out/minikube-linux-amd64 -p functional-20200323T124138.717452471-31589 ssh "echo hello"
functional_test.go:707: [out/minikube-linux-amd64 -p functional-20200323T124138.717452471-31589 ssh echo hello] = "hello\n", want = "hello\r\n"
```

priyawadhwa:
TestFunctional/parallel/SSHCmd was failing with the error:

```
    --- FAIL: TestFunctional/parallel (0.00s)
        --- FAIL: TestFunctional/parallel/SSHCmd (0.40s)
            functional_test.go:702: (dbg) Run:  out/minikube -p minikube ssh "echo hello"
            functional_test.go:707: [../../out/minikube -p minikube ssh echo hello] = "hello\n", want = "hello\r\n"
```

this fixes it locally on my mac

should fix #7170

### DIFF
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -698,7 +698,7 @@ func validateSSHCmd(ctx context.Context, t *testing.T, profile string) {
 	if NoneDriver() {
 		t.Skipf("skipping: ssh unsupported by none")
 	}
-	want := "hello\r\n"
+	want := "hello\n"
 	rr, err := Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", fmt.Sprintf("echo hello")))
 	if err != nil {
 		t.Errorf("%s failed: %v", rr.Args, err)


test/integration/functional_test.go
```diff
@@ -698,7 +698,7 @@ func validateSSHCmd(ctx context.Context, t *testing.T, profile string) {
 	if NoneDriver() {
 		t.Skipf("skipping: ssh unsupported by none")
 	}
-	want := "hello\r\n"
```
medyagh: rather than fixing the test, I rather find out why this test started to fail, are we using the wrong SSH client ?
could we either add a test to double check we are using the correct SSH client
(we have two options, native client and libmachine client...)

tstromberg:
I didn't like this PR at first because I felt like it was papering over a regression... but I think somehow a real bug was fixed, and this test confirms it.

Here's the hex dump of the data one would expect:

```
$ ssh <freebsd host> "echo hello | xxd
00000000: 6865 6c6c 6f0a                           hello.
```

Here is the hex dump at head:

```
./out/minikube ssh "echo hello" | xxd
00000000: 6865 6c6c 6f0a                           hello.
```

Here's a hex dump from v1.8.2, which differs:

```
00000000: 6865 6c6c 6f0d 0a                        hello..
```

v1.8.2 was even internally inconsistent, where `--native-ssh=false` showed the correct result:

```
minikube ssh --native-ssh=false "echo hello" | xxd
00000000: 6865 6c6c 6f0a                           hello.
```

test/integration/functional_test.go
tstromberg: FWIW, both options should yield the same output. 

I'd like to get a root cause, but I believe this PR is now doing the right thing (see my last comment).

priyawadhwa:
FWIW, running `git bisect` resulted in 577dfa339339d194f55a83b27bedee1ed0c46130 as the commit where SSH Command breaks. i don't see anything in that commit that seems obviously wrong though, so I may have done the bisect wrong.

PR #7183, opened 2020-03-24
add friendly error message for missing conntrack

sharifelgamal:
```
😄  minikube v1.9.0-beta.2 on Ubuntu 18.04
💣  The none driver requires conntrack to be installed for kubernetes version 1.18.0
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -835,12 +835,12 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 			out.WarningT("Using the '{{.runtime}}' runtime with the 'none' driver is an untested configuration!", out.V{"runtime": runtime})
 		}
 
-		conntrackK8sVersion, _ := semver.Make("1.18.0-rc.0")
-		currentK8sVersion, _ := semver.Make(strings.TrimPrefix(getKubernetesVersion(nil), version.VersionPrefix))
-		if currentK8sVersion.GE(conntrackK8sVersion) {
+		connVer, _ := semver.Make("1.18.0-rc.0")
+		k8sVersion, _ := semver.Make(strings.TrimPrefix(getKubernetesVersion(nil), version.VersionPrefix))
+		if k8sVersion.GE(connVer) {
 			err := exec.Command("conntrack").Run()
 			if err != nil {
-				exit.WithCodeT(exit.Config, "The none driver requires conntrack to be installed for kubernetes version {{.k8sVersion}}", out.V{"k8sVersion": currentK8sVersion.String()})
+				exit.WithCodeT(exit.Config, "The none driver requires conntrack to be installed for kubernetes version {{.k8sVersion}}", out.V{"k8sVersion": k8sVersion.String()})
 
 			}
 		}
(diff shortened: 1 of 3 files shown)


pkg/minikube/problem/err_map.go
```diff
@@ -225,6 +226,12 @@ var vmProblems = map[string]match{
 		Issues: []int{6083, 5636},
 		GOOS:   []string{"linux"},
 	},
+	"NONE_CONNTRACK": {
+		Regexp: re(`[ERROR FileExisting-conntrack]: conntrack not found in system path`),
+		Advice: "Kubernetes 1.18 and higher requires conntrack to be installed, to install it on your machine for the none driver simply run `apt-get install conntrack` as root.",
```
medyagh: don't mention apt-get install...since it could be different for centos and others. maybe add a link somewhere

cmd/minikube/cmd/start.go
```diff
@@ -834,6 +834,16 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 		if runtime != "docker" {
 			out.WarningT("Using the '{{.runtime}}' runtime with the 'none' driver is an untested configuration!", out.V{"runtime": runtime})
 		}
+
+		conntrackK8sVersion, _ := semver.Make("1.18.0-rc.0")
```
medyagh: consider shorter var, conntrackK8sVer
add a comment explain why.

medyagh: move this to one of our helper funcs, that checks against driver or k8s version.
validateDriver or ValidateFlag or one of those that has the right params

sharifelgamal: this is already inside validateFlags. validateDriver doesn't have any more context than this function had, and this seemed like a natural fit since we're already checking for none-driver specific stuff.

cmd/minikube/cmd/start.go
```diff
@@ -834,6 +834,17 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 		if runtime != "docker" {
 			out.WarningT("Using the '{{.runtime}}' runtime with the 'none' driver is an untested configuration!", out.V{"runtime": runtime})
 		}
+
+		// conntrack is required starting with kubernetes 1.18, include the release candidates for completion
+		connVer, _ := semver.Make("1.18.0-rc.0")
+		k8sVersion, _ := semver.Make(strings.TrimPrefix(getKubernetesVersion(nil), version.VersionPrefix))
```
tstromberg: optional: there is also `util.ParseKubernetesVersion`.

cmd/minikube/cmd/start.go
```diff
@@ -834,6 +834,17 @@ func validateFlags(cmd *cobra.Command, drvName string) {
...
+		if k8sVersion.GE(connVer) {
```
tstromberg: `if version.GTE(semver.MustParse("1.18.0-beta.1")) {`

cmd/minikube/cmd/start.go

cmd/minikube/cmd/start.go

PR #7189, opened 2020-03-24
add unit test for pkg/drivers/hyperkit/iso.go

Hellcatlk:
Signed-off-by: Zouyu <zouy.fnst@cn.fujitsu.com>


tstromberg:
Thank you!

PR #7206, opened 2020-03-24
Parse --disk-size and --memory sizes with binary suffixes

linkvt:
Hi!

when enabling the istio addon I got the message that I would need more memory in my VM - to my surprise it parsed the `--memory 8g` parameter as `8000MB` and not `8192MB`.
After looking at the rest of the code which uses 1024 in various locations for calculating sizes this seems not intended - I'm of course not sure if this is really the case :).

### Detailed Issue Description
```
 $ minikube addons enable istio
❗  Istio needs 8192MB of memory -- your configuration only allocates 4000MB
...
```

As it said I need to give my VM 8192MB I thought that I can just run it with `--memory 8g` which resulted in 8000MB instead of the expected 8GiB=8192MB:
```
 $ minikube start --memory 8g --cpus 4
😄  minikube v1.8.2 on Darwin 10.15.3
🏃  Using the running hyperkit "minikube" VM ...
🐳  Preparing Kubernetes v1.17.3 on Docker 19.03.6 ...
🚀  Launching Kubernetes ...
🌟  Enabling addons: dashboard, default-storageclass, efk, ingress, logviewer, registry, storage-provisioner
 $ minikube addons enable istio
❗  Istio needs 8192MB of memory -- your configuration only allocates 8000MB
...
```

Thanks!
Vincent

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -38,12 +38,13 @@ func CalculateSizeInMB(humanReadableSize string) (int, error) {
 	if err == nil {
 		humanReadableSize += "mb"
 	}
-	size, err := units.FromHumanSize(humanReadableSize)
+	// parse the size suffix binary instead of decimal so that 1G -> 1024MB instead of 1000MB
+	size, err := units.RAMInBytes(humanReadableSize)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


tstromberg:
Thank you! This seems fully reasonable to me.

/ok-to-test

PR #7209, opened 2020-03-24
wip: wait for Default Service Account

medyagh:
This Pr will break
minikube start --wait

but the following will still work
minikube start --wait=true (waits for 2 as the current default behavior(
minikube start --wait=false 
minikube start --wait=all  (waits for 3 components)
minikube start --wait=apiserver,system_pods,default_sa (waits for these 3)

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -44,6 +44,7 @@ import (
 	cmdcfg "k8s.io/minikube/cmd/minikube/cmd/config"
 	"k8s.io/minikube/pkg/drivers/kic/oci"
 	"k8s.io/minikube/pkg/minikube/bootstrapper/bsutil"
+	"k8s.io/minikube/pkg/minikube/bootstrapper/bsutil/kverify"
 	"k8s.io/minikube/pkg/minikube/bootstrapper/images"
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/constants"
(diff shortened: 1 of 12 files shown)


afbjorklund:
I suppose you could add a separate flag `--wait-components`, if you worry about backwards compat ?

medyagh:
> I suppose you could add a separate flag `--wait-components`, if you worry about backwards compat ?

I thought about that, but the thing is, I dont want the default behiavor to include waiting for default service account (that is not needed for 90% of ppl) and adds ups to 30s to start time.

if I had made a components flag, then if I keep the old flag, the True would need to do all of it ?

tstromberg:
It's unclear to me what this breaks if anything?

What do you think of a tiered approach so that folks don't have to worry about new components that are added?

`minikube start --wait=false`: wait for nothing, same as today
`minikube start --wait=responses` -- default: no behavioral change from today
`minikube start --wait=all` -- wait for everything

`minikube start --wait=true` -- alias for "all" for backwards compatibility

medyagh:
> It's unclear to me what this breaks if anything?
> 
> What do you think of a tiered approach so that folks don't have to worry about new components that are added?
> 
> `minikube start --wait=false`: wait for nothing, same as today
> `minikube start --wait=responses` -- default: no behavioral change from today
> `minikube start --wait=all` -- wait for everything
> 
> `minikube start --wait=true` -- alias for "all" for backwards compatibility

sounds like a good idea !

Issue #7184, opened 2020-03-24
KVM: TestStartStopWithPreload 


medyagh:
as seen here:

https://storage.googleapis.com/minikube-builds/logs/7165/ba4e134/KVM_Linux.html#fail_TestStartStopWithPreload

medyagh:
CC: @priyawadhwa

medyagh:
as seen here too https://storage.googleapis.com/minikube-builds/logs/7202/0478f2d/KVM_Linux.html#fail_TestStartStopWithPreload

priyawadhwa:
This seems to be the preload error:

```
I0324 13:55:49.458798    5115 docker.go:368] k8s.gcr.io/kube-proxy:v1.17.3 wasn't preloaded
I0324 13:55:49.458879    5115 ssh_runner.go:101] Run: sudo cat /var/lib/docker/image/overlay2/repositories.json
I0324 13:55:49.477302    5115 ssh_runner.go:101] Run: which lz4
W0324 13:55:49.483958    5115 start.go:205] Docker preload failed: getting file asset: Error opening file asset: /home/jenkins/minikube-integration/linux-amd64-kvm2-7202-30594-0478f2d2048e7b82627351c9afe530156f0843b9/.minikube/cache/preloaded-tarball/preloaded-images-k8s-v2-v1.17.3-docker-overlay2-amd64.tar.lz4: open /home/jenkins/minikube-integration/linux-amd64-kvm2-7202-30594-0478f2d2048e7b82627351c9afe530156f0843b9/.minikube/cache/preloaded-tarball/preloaded-images-k8s-v2-v1.17.3-docker-overlay2-amd64.tar.lz4: no such file or directory, falling back to caching images
```

so something is broken on restart

and minikube actually fails because of a coredns error:

```
error execution phase addon/coredns: unable to create a new DNS service: Post https://192.168.39.120:8443/api/v1/namespaces/kube-system/services?timeout=10s: dial tcp 192.168.39.120:8443: connect: connection refused
To see the stack trace of this error execute with --v=5 or higher
* 
[...]

PR #7210, opened 2020-03-24
Fix TestStartStopWithPreload bug

priyawadhwa:
This PR adds a --preload flag to optionally preload images, and makes sure preload still happens on hot restarts in case the kubernetes version has been upgraded. I added it in this PR because in 
TestStartStopWithPreload, we want to make sure the first start happens without preload. Once we generate a tarball for v1.17.0, that will break.

Also fixes a bug in which if a preloaded tarball didn't exist, we would still call `createVolume`, which would create a directory in the cache and break minikube (because we would try to extract the directory). I think this was causing #7184 , so checking if a preloaded tarball exists before creating the volume should fix it

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -123,6 +123,7 @@ const (
 	hostOnlyNicType         = "host-only-nic-type"
 	natNicType              = "nat-nic-type"
 	nodes                   = "nodes"
+	preload                 = "preload"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -175,6 +176,7 @@ func initMinikubeFlags() {
 	startCmd.Flags().Bool(autoUpdate, true, "If set, automatically updates drivers to the latest version. Defaults to true.")
 	startCmd.Flags().Bool(installAddons, true, "If set, install addons. Defaults to true.")
 	startCmd.Flags().IntP(nodes, "n", 1, "The number of nodes to spin up. Defaults to 1.")
+	startCmd.Flags().Bool(preload, true, "If true, download tarball of preloaded images if available to improve start time.")
```
medyagh: Add wording, Defaults to true, to match other flags wording.

pkg/minikube/download/preload.go
```diff
@@ -76,6 +77,10 @@ func remoteTarballURL(k8sVersion string) string {
 
 // PreloadExists returns true if there is a preloaded tarball that can be used
 func PreloadExists(k8sVersion, containerRuntime string) bool {
+	if !viper.GetBool("preload") {
+		return false
+	}
+
 	if containerRuntime != "docker" {
```
medyagh: nit: maybe add the issue link for when we will add containerd and crio runtimes.

cmd/minikube/cmd/start.go
priyawadhwa: done

PR #6142, opened 2019-12-20
Added support for "stable" and "latest"aliases for --kubernetes-version  

ths-laurent:
Added the check for the aliases.
In the issue's comment it was mentioned that the constant NewestKubernetesVersion should be 1.18.0-alpha.1 (now it is 1.17.0). I don't know if it was part of this issue so I left it unchanged.
Related to #6126

PR #7212, opened 2020-03-24
Add 'stable' and 'latest' as valid kubernetes-version values

### LINKED ISSUE #6126: Add support for --kubernetes-version=(latest|stable) aliases
As proposed by @robrich in #6122

As I mentioned there, minikube does not yet support `latest` or stable` version aliases, but it would be trivial to add.  If you or anyone else is interested in making this work, add a check for `paramVersion == "stable"` here:

https://github.com/kubernetes/minikube/blob/56c72e207965e6ff6fb760ed47a5039d0f3fb121/cmd/minikube/cmd/start.go#L1127

We currently store the stable and latest versions as constants that you can refer to.

linkvt:
Fixes #6126 

Hi,

mini PR, this was done with quite a lot of other changes in https://github.com/kubernetes/minikube/pull/6142 already but the user deleted his account.

I added the tags to the CLI help output, used a case insensitive comparison and really basic tests which are longer then the actual code but I think they still make sense e.g. to catch errors during refactorings or such.

CLI Output:
```
 $ out/minikube start --help | grep kubernetes-version
      --kubernetes-version='': The kubernetes version that the minikube VM will use (ex: v1.2.3 or latest/stable for supported versions)
```

**Edit**: I could adjust the CLI help output to contain the mapping from latest/stable to the appropriate version, it could look like:
```
 $ out/minikube start --help | grep kubernetes-version
      --kubernetes-version='': The kubernetes version that the minikube VM will use (ex: v1.2.3, 'stable' for v1.18.0-rc.1, 'latest' for v1.18.0-rc.1)
```

BR, Vincent

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -179,7 +179,7 @@ func initMinikubeFlags() {
 
 // initKubernetesFlags inits the commandline flags for kubernetes related options
 func initKubernetesFlags() {
-	startCmd.Flags().String(kubernetesVersion, "", "The kubernetes version that the minikube VM will use (ex: v1.2.3)")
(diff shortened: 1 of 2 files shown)


linkvt:
@tstromberg I added the CLI help changes here and flattened the ifs for parsing the version a bit.

Issue #7204, opened 2020-03-24
X Unable to start VM: host remove: Failed to find the VM status


garimagupta779:
**The exact command to reproduce the issue**:

**The full output of the command that failed**:<details>

</details>

**The output of the `minikube logs` command**:<details>

</details>

**The operating system version**:

tstromberg:
I don’t yet have a clear way to replicate this issue. Do you mind adding some additional details? Here is additional information that would be helpful:

*  The exact `minikube start` command line used
*  The full output of the `minikube start` command, preferably with `--alsologtostderr -v=4` for extra logging.

Thank you for sharing your experience!

tstromberg:
Fixed at head.

PR #7230, opened 2020-03-25
Make recreateIfNeeded more robust, improve UX

tstromberg:
Makes recreateIfNeeded more resilient to unexpected machine state failures, such as #7204

Also improves the UX. Here's an example where I've removed the VM of an active minikube cluster from VirtualBox:

## Before

```
🙄  machine 'minikube' does not exist. Proceeding ahead with recreating VM.
🔥  Creating virtualbox VM (CPUs=2, Memory=6000MB, Disk=20000MB) ...
🤦  StartHost failed, but will try again: machine does not exist
🏃  Using the running virtualbox "minikube" VM ...
```

## After

```
🤷  virtualbox "minikube" VM is missing, will recreate.
🔥  Creating virtualbox VM (CPUs=2, Memory=6000MB, Disk=20000MB) ...
```

### DIFF
--- a/pkg/minikube/machine/fix.go
+++ b/pkg/minikube/machine/fix.go
@@ -54,8 +54,6 @@ var (
 
 // fixHost fixes up a previously configured VM so that it is ready to run Kubernetes
 func fixHost(api libmachine.API, cc config.ClusterConfig, n config.Node) (*host.Host, error) {
-	out.T(out.Waiting, "Reconfiguring existing host ...")
-
 	start := time.Now()
 	glog.Infof("fixHost starting: %s", n.Name)
 	defer func() {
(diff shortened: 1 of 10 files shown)


pkg/minikube/machine/fix.go
```diff
@@ -101,51 +99,50 @@ func fixHost(api libmachine.API, cc config.ClusterConfig, n config.Node) (*host.
...
-				}
-				// return ErrMachineNotExist err to initialize preExists flag
-				return h, ErrorMachineNotExist
+		me, err := machineExists(h.Driver.DriverName(), s, serr)
+		glog.Infof("exists: %v err=%v", me, err)
+
+		if !me || err == ErrorMachineNotExist {
+			out.T(out.Shrug, `{{.driver_name}} "{{.cluster}}" {{.machine_type}} is missing, will recreate.`, out.V{"driver_name": cc.Driver, "cluster": cc.Name, "machine_type": machineType})
+			demolish(api, cc, n, h)
+			time.Sleep(1 * time.Second)
```
priyawadhwa: could you include a comment as to why this sleep is needed?

PR #7232, opened 2020-03-25
none:  Skip checkHelmTiller if socat is not installed

tstromberg:
Fixes #7197

PR #7236, opened 2020-03-25
install socat

medyagh:
for none tests we need kubectl portforward to test, as a follow up PR for https://github.com/kubernetes/minikube/pull/7232
to install socat on none machines, so we get higher test coverage.


priyawadhwa:
LGTM, only suggestion is this could use a more descriptive PR title

PR #7242, opened 2020-03-25
Add digest to image in daemon after pulling it

### LINKED ISSUE #7226: download only waits 18 seconds even if everything downloaded
to reproduce.

```
./out/minikube delete --all --purge
./out/minikube start --download-only --driver=docker 
```

(now it it should have nothing to download anymore lets run it again)

```
./out/minikube start --download-only --driver=docker anymore but it takes 18 seconds of waiting without any logs)  
./out/minikube start --download-only --driver=docker --alsologtostderr
```

```
medmac@~/workspace/minikube (master) $ time ./out/minikube start --download-only --driver=docker --alsologtostderr
W0325 00:39:58.079830   41672 root.go:248] Error reading config file at /Users/medmac/.minikube/config/config.json: open /Users/medmac/.minikube/config/config.json: no such file or directory
I0325 00:39:58.080424   41672 notify.go:125] Checking for updates...
I0325 00:39:58.377046   41672 start.go:258] hostinfo: {"hostname":"Medyas-iMac-10.local","uptime":115739,"bootTime":1585006259,"procs":359,"os":"darwin","platform":"darwin","platformFamily":"","platformVersion":"10.13.6","kernelVersion":"17.7.0","virtualizationSystem":"","virtualizationRole":"","hostid":"d8347098-9b31-3c8e-a1c9-ca857f05bb97"}
W0325 00:39:58.377216   41672 start.go:266] gopshost.Virtualization returned error: not implemented yet
😄  minikube v1.9.0-beta.2 on Darwin 10.13.6
W0325 00:39:58.400460   41672 start.go:546] selectDriver api.Load: filestore "minikube": Docker machine "minikube" does not exist. Use "docker-machine ls" to list machines. Use "docker-machine create" to add a new one.
I0325 00:39:58.400515   41672 driver.go:226] Setting default libvirt URI to qemu:///system
W0325 00:39:58.400591   41672 start.go:546] selectDriver api.Load: filestore "minikube": Docker machine "minikube" does not exist. Use "docker-machine ls" to list machines. Use "docker-machine create" to add a new one.
I0325 00:39:58.536883   41672 start.go:306] selected driver: docker
I0325 00:39:58.536904   41672 start.go:595] validating driver "docker" against &{Name:minikube KeepContext:false EmbedCerts:false MinikubeISO: Memory:4000 CPUs:2 DiskSize:20000 Driver:docker HyperkitVpnKitSock: HyperkitVSockPorts:[] DockerEnv:[] InsecureRegistry:[] RegistryMirror:[] HostOnlyCIDR:192.168.99.1/24 HypervVirtualSwitch: HypervUseExternalSwitch:false HypervExternalAdapter: KVMNetwork:default KVMQemuURI:qemu:///system KVMGPU:false KVMHidden:false DockerOpt:[] DisableDriverMounts:false NFSShare:[] NFSSharesRoot:/nfsshares UUID: NoVTXCheck:false DNSProxy:false HostDNSResolver:true HostOnlyNicType:virtio NatNicType:virtio KubernetesConfig:{KubernetesVersion:v1.18.0-rc.1 ClusterName:minikube APIServerName:minikubeCA APIServerNames:[] APIServerIPs:[] DNSDomain:cluster.local ContainerRuntime:docker CRISocket: NetworkPlugin: FeatureGates: ServiceCIDR:10.96.0.0/12 ImageRepository: ExtraOptions:[{Component:kubeadm Key:pod-network-cidr Value:10.244.0.0/16}] ShouldLoadCachedImages:true EnableDefaultCNI:false NodeIP: NodePort:0 NodeName:} Nodes:[{Name:m01 IP: Port:8443 KubernetesVersion:v1.18.0-rc.1 ControlPlane:true Worker:true}] Addons:map[]}
I0325 00:39:58.537042   41672 start.go:601] status for docker: {Installed:true Healthy:true Error:<nil> Fix: Doc:}
I0325 00:39:58.537071   41672 start.go:1040] auto setting extra-config to "kubeadm.pod-network-cidr=10.244.0.0/16".
I0325 00:39:58.647174   41672 start.go:945] Using suggested 4000MB memory alloc based on sys=16384MB, container=7964MB
🚜  Pulling base image ...
I0325 00:39:58.665784   41672 cache.go:104] Beginning downloading kic artifacts
I0325 00:39:58.665887   41672 preload.go:94] Found local preload: /Users/medmac/.minikube/cache/preloaded-tarball/preloaded-images-k8s-v2-v1.18.0-rc.1-docker-overlay2-amd64.tar.lz4
I0325 00:39:58.665902   41672 cache.go:46] Caching tarball of preloaded images
I0325 00:39:58.665917   41672 cache.go:106] Downloading gcr.io/k8s-minikube/kicbase:v0.0.8@sha256:2f3380ebf1bb0c75b0b47160fd4e61b7b8fef0f1f32f9def108d3eada50a7a81 to local daemon
I0325 00:39:58.665946   41672 image.go:84] Writing gcr.io/k8s-minikube/kicbase:v0.0.8@sha256:2f3380ebf1bb0c75b0b47160fd4e61b7b8fef0f1f32f9def108d3eada50a7a81 to local daemon
I0325 00:39:58.665924   41672 preload.go:123] Found /Users/medmac/.minikube/cache/preloaded-tarball/preloaded-images-k8s-v2-v1.18.0-rc.1-docker-overlay2-amd64.tar.lz4 in cache, skipping download
I0325 00:39:58.665964   41672 cache.go:49] Finished downloading the preloaded tar for v1.18.0-rc.1 on docker
I0325 00:39:58.666102   41672 profile.go:138] Saving config to /Users/medmac/.minikube/profiles/minikube/config.json ...
I0325 00:39:58.671298   41672 binary.go:57] Not caching binary, using https://storage.googleapis.com/kubernetes-release/release/v1.18.0-rc.1/bin/linux/amd64/kubectl?checksum=file:https://storage.googleapis.com/kubernetes-release/release/v1.18.0-rc.1/bin/linux/amd64/kubectl.sha256
I0325 00:39:58.671367   41672 binary.go:57] Not caching binary, using https://storage.googleapis.com/kubernetes-release/release/v1.18.0-rc.1/bin/linux/amd64/kubelet?checksum=file:https://storage.googleapis.com/kubernetes-release/release/v1.18.0-rc.1/bin/linux/amd64/kubelet.sha256
I0325 00:39:58.671379   41672 binary.go:57] Not caching binary, using https://storage.googleapis.com/kubernetes-release/release/v1.18.0-rc.1/bin/linux/amd64/kubeadm?checksum=file:https://storage.googleapis.com/kubernetes-release/release/v1.18.0-rc.1/bin/linux/amd64/kubeadm.sha256
I0325 00:39:58.671425   41672 binary.go:57] Not caching binary, using https://storage.googleapis.com/kubernetes-release/release/v1.18.0-rc.1/bin/darwin/amd64/kubectl?checksum=file:https://storage.googleapis.com/kubernetes-release/release/v1.18.0-rc.1/bin/darwin/amd64/kubectl.sha256
I0325 00:40:16.408337   41672 cache.go:117] Successfully downloaded all kic artifacts

------ when it comes to THIS PART to goes silent for 15 seconds without any logs------


✅  Download complete!

real	0m18.408s
user	0m5.363s
sys	0m3.155s
```

afbjorklund:
The image was added with digest "none", which caused it to be
written to daemon again next time since reference didn't match.

Currently this is missing from the go-containerregistry API,
so use "docker pull" as a workaround (it should be rather fast)

Fixes #7226

### DIFF
--- a/pkg/minikube/image/image.go
+++ b/pkg/minikube/image/image.go
@@ -113,6 +113,8 @@ func WriteImageToDaemon(img string) error {
 
 	//TODO: Make pkg/v1/daemon accept Ref too
 	//      Only added it to pkg/v1/tarball
+	//
+	// https://github.com/google/go-containerregistry/pull/702
 
 	glog.V(3).Infof("Pulling image %v", ref)
 
--- a/pkg/minikube/image/image.go
+++ b/pkg/minikube/image/image.go
@@ -96,6 +96,7 @@ func WriteImageToDaemon(img string) error {
 	if err != nil {
 		return errors.Wrap(err, "parsing reference")
 	}
+	glog.V(3).Infof("Getting image %v", ref)
 	i, err := remote.Image(ref)
 	if err != nil {
 		return errors.Wrap(err, "getting remote image")
@@ -104,8 +105,24 @@ func WriteImageToDaemon(img string) error {
 	if err != nil {
 		return errors.Wrap(err, "getting tag")
 	}
+	glog.V(3).Infof("Writing image %v", tag)
 	_, err = daemon.Write(tag, i)
-	return err
+	if err != nil {
+		return errors.Wrap(err, "writing image")
+	}
+
+	//TODO: Make pkg/v1/daemon accept Ref too
+	//      Only added it to pkg/v1/tarball
+
+	glog.V(3).Infof("Pulling image %v", ref)
+
+	// Pull digest
+	cmd = exec.Command("docker", "pull", "--quiet", img)
+	if _, err := cmd.Output(); err != nil {
+		return errors.Wrap(err, "pulling remote image")
+	}
+
+	return nil
 }
 
 func retrieveImage(ref name.Reference) (v1.Image, error) {


pkg/minikube/image/image.go
```diff
@@ -104,8 +105,24 @@ func WriteImageToDaemon(img string) error {
...
 		return errors.Wrap(err, "getting tag")
 	}
+	glog.V(3).Infof("Writing image %v", tag)
 	_, err = daemon.Write(tag, i)
-	return err
+	if err != nil {
+		return errors.Wrap(err, "writing image")
+	}
+
+	//TODO: Make pkg/v1/daemon accept Ref too
```
medyagh: please make an Issue for any TODO and link to the TODO comment

pkg/minikube/image/image.go
```diff
@@ -104,8 +105,24 @@ func WriteImageToDaemon(img string) error {
...
+	//      Only added it to pkg/v1/tarball
+
+	glog.V(3).Infof("Pulling image %v", ref)
+
+	// Pull digest
+	cmd = exec.Command("docker", "pull", "--quiet", img)
```
medyagh: use podman, if the user's driver is podman

medyagh: same as code in line 87, cmd := exec.Command("docker", "images", "--format", "{{.Repository}}:{{.Tag}}@{{.Digest}}")

they both should use podman I believe, if user has podman as the driver.right?

pkg/minikube/image/image.go
afbjorklund: Haven't even filed a bug report about it yet

medyagh:
i suggest also create an issue in  go-containerregistry and link to this PR, so they could fix it and we dont have to use docker and podman binaries directly (minikube does not require docker to be installed for non-docker drivers)

pkg/minikube/image/image.go
afbjorklund: I didn't change that part, it was hardcoded to "docker" already before

afbjorklund:
> i suggest also create an issue in go-containerregistry and link to this PR, so they could fix it and we dont have to use docker and podman binaries directly (minikube does not require docker to be installed for non-docker drivers)

Sounds like a good idea, this is similar to the previous issue https://github.com/google/go-containerregistry/commit/019cdfc6adf96a4905a1b93a7aeaea1e50c0b6cf

pkg/minikube/image/image.go
medyagh: yeah it was wrong before this PR.

afbjorklund:
@medyagh : There is a problem with the `go-containerregistry` Reference:

```go
// Reference defines the interface that consumers use when they can
// take either a tag or a digest.
```

```go
// Tag stores a docker tag name in a structured form.
type Tag struct {
	Repository
	tag      string
	original string
}
```

```go
// Digest stores a digest name in a structured form.
type Digest struct {
	Repository
	digest   string
	original string
}

```

Notice the "either/or" part. So when the digest is added, the tag is dropped...

For our use case, this actually makes it **worse**. Now the tag is `none` instead!

----

So it needs to have a feature added, to allow NewTagDigest with both. At once.

```
gcr.io/k8s-minikube/kicbase:v0.0.8@sha256:2f3380ebf1bb0c75b0b47160fd4e61b7b8fef0f1f32f9def108d3eada50a7a81
```

```go
type TagDigest struct {
	Repository
	tag      string
	digest   string
	original string
}
```

Including a ton of code that looks like "if tag else digest", all over the place. <sigh>

afbjorklund:
Actually Digest already handled this case, it just didn't save the Tag...

Added PR https://github.com/google/go-containerregistry/issues/703 which fixes the issue.

medyagh:
@afbjorklund  this is an important PR that will save seconds off minikube start,.

if we can't get it merged on time on 
https://github.com/google/go-containerregistry/pull/703 

how about we use your fork and put some miles on it, till that PR gets merged?

tstromberg:
I don't think we need to merge this. We know the folks who work on go-containerregistry.

afbjorklund:
You can use the `docker pull` workaround, or avoid using the digest for the `kicbase` image.

pkg/minikube/image/image.go
afbjorklund: Reported as https://github.com/google/go-containerregistry/pull/702

pkg/minikube/image/image.go
afbjorklund: This will need to implement WriteImageToDaemon, so that it works with podman.

afbjorklund:
The discussion about adding "pkg/v1/kind" in addition to "pkg/v1/daemon" moved here:

~https://github.com/google/go-containerregistry/issues/706~
https://github.com/google/ko/issues/149

> I think we want something like:
> https://github.com/google/go-containerregistry/tree/master/pkg/v1/daemon
> 
> ... but to write it directly to KinD.

I suppose we could have asked for a pkg/v1/minikube, but it looked the _wrong_ decision ?

----

Also, it seems like the OCI binary is missing a `sudo` ?

`docker=sudo podman`

afbjorklund:
@tstromberg : Apparently the go-containerregistry folks only know about Docker, and not about CRI
`"We don't have much bandwidth to review changes and support code related to new technologies we're not familiar with"`

So it looks like they will just hardcode something for KIND (and containerd), rather than add something more generic like minikube's cruntime (and the `LoadImage` function). So I guess we won't have any podman/crio support, in go-containerregistry.

afbjorklund:
This PR _should_ solve the original issue (for Docker).

We can leave the "wouldn't it be nice if" until later ?

Issue #7190, opened 2020-03-24
none: TestVersionUpgrade 'exit 0' : driver does not support ssh command


medyagh:
as seen in Two PRs:

https://storage.googleapis.com/minikube-builds/logs/7182/3c37556/none_Linux.html#fail_TestVersionUpgrade

and 

https://storage.googleapis.com/minikube-builds/logs/7173/aec6fdc/none_Linux.html#fail_TestVersionUpgrade

medyagh:
similar to errors in https://github.com/kubernetes/minikube/issues/7194

tstromberg:
Root issue: #4132

tstromberg:
Fixed by #7244

PR #7244, opened 2020-03-26
Avoid provision.Detector when possible

### LINKED ISSUE #4132: none: reusing node: detecting provisioner: Too many retries waiting for SSH to be available
Environment:

minikube version: `v1.0.0`
OS: `Ubuntu 16.04 LTS (Xenial Xerus)`
VM Driver: `none`

What happened: ```
Created a VM with none driver, stopped it, then started it again. The VM failed to start and minikube reported that it crashed.
```

What I expected to happen: 
the VM created by the first minikube start command is started.

Output from the second minikube start command:

😄  minikube v1.0.0 on linux (amd64)
🤹  Downloading Kubernetes v1.14.0 images in the background ...
💡  Tip: Use 'minikube start -p <name>' to create a new cluster, or 'minikube delete' to delete this one.
🔄  Restarting existing none VM for "minikube" ...
⌛  Waiting for SSH access ...

💣  Unable to start VM: detecting provisioner: Too many retries waiting for SSH to be available.  Last error: Maximum number of  
retries (60) exceeded

Output from 'sudo minikube start --alsologtostderr -v=8 --vm-driver=none':

⌛  Waiting for SSH access ...
Waiting for SSH to be available...
Getting to WaitForSSH function...
Error getting ssh command 'exit 0' : driver does not support ssh commands
```

To reproduce:
    sudo minikube start --vm-driver=none
    sudo minikube stop
    sudo minikube start --vm-driver=none

Starting a stopped VM was working in minikube v0.28.

tstromberg:
* fixHost no longer attempts to re-provision BareMetal hosts, as doing so requires SSH. This fixes a regression: ` 'exit 0' : driver does not support ssh command`
* logRemoteOsRelease now uses the command-runner interface, so it is not necessary to call provision.DetectProvisioner

Fixes #4132 and #7190


pkg/minikube/machine/fix.go
```diff
@@ -67,21 +67,24 @@ func fixHost(api libmachine.API, cc config.ClusterConfig, n config.Node) (*host.
...
 	}
 
-	// Technically, we should only have to call provision if Docker has changed,
-	// but who can predict what shape the existing VM is in.
-	e := engineOptions(cc)
-	h.HostOptions.EngineOptions.Env = e.Env
-	err = provisionDockerMachine(h)
-	if err != nil {
-		return h, errors.Wrap(err, "provision")
+	if !driver.BareMetal(h.Driver.DriverName()) {
```
medyagh: was this a wrong negatition ?

pkg/minikube/machine/fix.go
```diff
@@ -67,21 +67,24 @@ func fixHost(api libmachine.API, cc config.ClusterConfig, n config.Node) (*host.
...
+		glog.Infof("%s is local, skipping re-provision as it requires SSH", driverName)
```
medyagh: this doesnt match the if logic

pkg/minikube/machine/fix.go
tstromberg: negation was right, log message wasn't. fixed.

pkg/minikube/machine/fix.go
tstromberg: Sorry, I attempted to turn a comment into a log message and made it only more confusing.

PR #7258, opened 2020-03-26
Fix typo on macos.md

viniciusbds:
This PR only fix a typo on macos.md page


tstromberg:
We prefer to be called `minikube`.

PR #7267, opened 2020-03-26
kic: Use bind IP for minikube IP on linux

### LINKED ISSUE #7191: TestAddons/parallel/Registry consistent docker failure: GET http://127.0.0.1:5000 giving up after 5 attempts
as seen here
https://storage.googleapis.com/minikube-builds/logs/7173/67bd258/Docker_Linux.html#fail_TestAddons%2fparallel%2fRegistry

```
make integration -e TEST_ARGS="-test.run TestAddons/parallel/Registry --profile=minikube --cleanup=false --minikube-start-args="--driver=docker""
```

priyawadhwa:
We want to return the external ip when running $(minikube ip). This was causing a bug in `TestAddons/parallel/Registry`, which expects $(minikube ip):5000 to connect to an nginx service when the registry addon is enabled.

```
make integration -e TEST_ARGS="-test.run TestAddons/parallel/Registry --profile=minikube --cleanup=false --minikube-start-args="--driver=docker""
```

fixes #7191


cmd/minikube/cmd/docker-env.go
```diff
@@ -151,6 +151,13 @@ var dockerEnvCmd = &cobra.Command{
 		var err error
 		port := constants.DockerDaemonPort
 		if driver.IsKIC(driverName) {
+			// Update ip
+			ip := net.ParseIP(oci.DefaultBindIPV4)
+			if ip == nil {
+				exit.WithCodeT(exit.Software, fmt.Sprintf("Unable to parse driver IP: %q", ip))
+			}
+			co.DriverIP = ip
```
tstromberg: instead of updating `co.DriverIP`, define a  new variable named `dockerIP` to be more accurate.

pkg/minikube/mustload/mustload.go
```diff
@@ -111,10 +110,6 @@ func Running(name string) ClusterController {
 		exit.WithError("Unable to get driver IP", err)
 	}
 
-	if driver.IsKIC(host.DriverName) {
```
medyagh: If I am not wrong, I believe we still need this for macOS.
we can not hit the IP of the container but the DefaultBind 127.0.0.1

priyawadhwa:
So on Mac I'm getting an i/o timeout:

```
=== RUN   TestAddons
=== RUN   TestAddons/parallel
=== RUN   TestAddons/parallel/Registry
=== PAUSE TestAddons/parallel/Registry
=== CONT  TestAddons/parallel/Registry
2020/03/26 16:32:07 [DEBUG] GET http://172.17.0.2:5000
2020/03/26 16:32:37 [ERR] GET http://172.17.0.2:5000 request failed: Get http://172.17.0.2:5000: dial tcp 172.17.0.2:5000: i/o timeout
2020/03/26 16:32:37 [DEBUG] GET http://172.17.0.2:5000: retrying in 1s (4 left)
2020/03/26 16:33:08 [ERR] GET http://172.17.0.2:5000 request failed: Get http://172.17.0.2:5000: dial tcp 172.17.0.2:5000: i/o timeout
```

tstromberg:
If I understand, this test can't really work on macOS, because the port isn't exposed.

@medyagh - What's the mechanism that regular users should be using to expose this port properly? The test should be using that, at least in the `IsKIC()` case.

afbjorklund:
Not sure there is anything proper about 5000, probably should use a tunnel to expose it ?

Where `docker0` is broken

priyawadhwa:
cc @medyagh PTAL as a temporary fix on linux

Issue #6956, opened 2020-03-08
Test Flake: calculate hash for cacert: Expecting: TRUSTED CERTIFICATE


medyagh:
As seen on a PR on the head:
https://storage.googleapis.com/minikube-builds/logs/6953/9e72a2a/Docker_Linux.html#fail_TestStartStop%2fgroup%2fold-docker

for TestStartStop:
```
I0308 11:48:21.318387    5087 crypto.go:165] Writing key to /home/jenkins/minikube-integration/linux-amd64-docker-6953-3395-9e72a2ad9cfe3e9e63411de65ecae4a2f2f707eb/.minikube/proxy-client.key ...
I0308 11:48:21.318415    5087 lock.go:35] WriteFile acquiring /home/jenkins/minikube-integration/linux-amd64-docker-6953-3395-9e72a2ad9cfe3e9e63411de65ecae4a2f2f707eb/.minikube/proxy-client.key: {Name:mke8369eb326ee552b61eb2006c7133b4cc6de85 Clock:{} Delay:500ms Timeout:1m0s Cancel:<nil>}
W0308 11:48:29.371066    5087 exit.go:101] Failed to setup certs: certificate symlinks: calculate hash for cacert /usr/share/ca-certificates/5904.pem: openssl x509 -hash -noout -in /usr/share/ca-certificates/5904.pem: openssl x509 -hash -noout -in /usr/share/ca-certificates/5904.pem: exit status 1
stdout:
stderr:
unable to load certificate
140712931894464:error:0909006C:PEM routines:get_name:no start line:../crypto/pem/pem_lib.c:745:Expecting: TRUSTED CERTIFICATE
* 
X Failed to setup certs: certificate symlinks: calculate hash for cacert /usr/share/ca-certificates/5904.pem: openssl x509 -hash -noout -in /usr/share/ca-certificates/5904.pem: openssl x509 -hash -noout -in /usr/share/ca-certificates/5904.pem: exit status 1
stdout:
stderr:
unable to load certificate
140712931894464:error:0909006C:PEM routines:get_name:no start line:../crypto/pem/pem_lib.c:745:Expecting: TRUSTED CERTIFICATE
* 
* minikube is exiting due to an error. If the above message is not useful, open an issue:
- https://github.com/kubernetes/minikube/issues/new/choose
** /stderr **

```

CC: @tstromberg

tstromberg:
apparently introduced by #6889 - March 4th

tstromberg:
[...]

PR #7270, opened 2020-03-26
Add more debugging for certificate issues

tstromberg:
To help catch a  race condition: #6956

### DIFF
--- a/pkg/minikube/bootstrapper/certs.go
+++ b/pkg/minikube/bootstrapper/certs.go
@@ -117,6 +117,7 @@ func SetupCerts(cmd command.Runner, k8s config.KubernetesConfig, n config.Node) 
 	}
 
 	for _, f := range copyableFiles {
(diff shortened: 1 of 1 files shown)


pkg/minikube/bootstrapper/certs.go
```diff
@@ -360,9 +362,16 @@ func collectCACerts() (map[string]string, error) {
 
 // getSubjectHash calculates Certificate Subject Hash for creating certificate symlinks
 func getSubjectHash(cr command.Runner, filePath string) (string, error) {
+	lrr, err := cr.RunCmd(exec.Command("ls", "-la", filePath))
+	if err != nil {
+		return "", err
```
medyagh: Wrap the error with something helpful.

tstromberg: I don't think there is anything extra we can add - the error already includes the command-line:

getSubjectHash: ls -la <path>: returned exit code 1

PR #7280, opened 2020-03-27
translations facelift

sharifelgamal:
* Fixes spurious errors printed during `make extract`
* Removes need to add braces to empty translation json files
* Updates French translations for `minikube start` and `minikube delete`
* More detailed documentation on how to add/edit translations


afbjorklund:
Yikes, that's a lot of different stuff in a single commit.

PR #7292, opened 2020-03-27
Show all global flags in options command

### LINKED ISSUE #5117: --help doesn't output global flags anymore
**The exact command to reproduce the issue**: `minikube --help`

```
PS D:\dev\minikube\out> minikube --help                                                                                                                                                                    Minikube is a CLI tool that provisions and manages single-node Kubernetes clusters optimized for development workflows.

Basic Commands:
  start          Starts a local kubernetes cluster
  status         Gets the status of a local kubernetes cluster
  stop           Stops a running local kubernetes cluster
  delete         Deletes a local kubernetes cluster
  dashboard      Access the kubernetes dashboard running within the minikube cluster

Images Commands:
  docker-env     Sets up docker env variables; similar to '$(docker-machine env)'
  cache          Add or delete an image from the local cache.

Configuration and Management Commands:
  addons         Modify minikube's kubernetes addons
  config         Modify minikube config
  profile        Profile gets or sets the current minikube profile
  update-context Verify the IP address of the running cluster in kubeconfig.

Networking and Connectivity Commands:
  service        Gets the kubernetes URL(s) for the specified service in your local cluster
  tunnel         tunnel makes services of type LoadBalancer accessible on localhost

Advanced Commands:
  mount          Mounts the specified directory into minikube
  ssh            Log into or run a command on a machine with SSH; similar to 'docker-machine ssh'
  kubectl        Run kubectl

Troubleshooting Commands:
  ssh-key        Retrieve the ssh identity key path of the specified cluster
  ip             Retrieves the IP address of the running cluster
  logs           Gets the logs of the running instance, used for debugging minikube, not user code.
  update-check   Print current and latest version number
  version        Print the version of minikube

Other Commands:
  completion     Outputs minikube shell completion for the given shell (bash or zsh)

Use "minikube <command> --help" for more information about a given command.
PS D:\dev\minikube\out>  
```

</details>

**The operating system version**: Windows 10 Enterprise.

Not sure if this is intentional but I would like to show the available global flags along with their descriptions as well.

linkvt:
Fixes #5117
Before this fix the flags shown in `minikube options` were taken from a list of options (`viperWhiteList`) which seems a bit arbitrary to me with no background.
I still don't know for sure why this is done but taking all PersistentFlags (persistent=apply to all sub commands) from the root command seems like the correct way.

This doesn't fix the problem as described in the issue but keeps it consistent with e.g. kubectl, which also shows global flags only with the `options` command.

### Before
```
 $ minikube version && minikube options
minikube version: v1.9.0
commit: 48fefd43444d2f8852f527c78f0141b377b1e42a
The following options can be passed to any command:

      --alsologtostderr=false: log to standard error as well as files
      --log_dir='': If non-empty, write log files in this directory
  -v, --v=0: log level for V logs
```
### After
```
 $ make && out/minikube version && out/minikube options
minikube version: v1.9.0
commit: b47e8d502fb62ec0a968c47e75414aeadf125b4d
The following options can be passed to any command:

      --alsologtostderr=false: log to standard error as well as files
  -b, --bootstrapper='kubeadm': The name of the cluster bootstrapper that will set up the kubernetes cluster.
      --log_backtrace_at=:0: when logging hits line file:N, emit a stack trace
      --log_dir='': If non-empty, write log files in this directory
      --logtostderr=false: log to standard error instead of files
  -p, --profile='minikube': The name of the minikube VM being used. This can be set to allow having multiple instances of minikube independently.
      --stderrthreshold=2: logs at or above this threshold go to stderr
  -v, --v=0: log level for V logs
      --vmodule=: comma-separated list of pattern=N settings for file-filtered logging
```

### DIFF
--- a/cmd/minikube/cmd/options.go
+++ b/cmd/minikube/cmd/options.go
@@ -37,10 +37,9 @@ var optionsCmd = &cobra.Command{
 // runOptions handles the executes the flow of "minikube options"
 func runOptions(cmd *cobra.Command, args []string) {
 	out.String("The following options can be passed to any command:\n\n")
-	for _, flagName := range viperWhiteList {
-		f := pflag.Lookup(flagName)
-		out.String(flagUsage(f))
-	}
+	cmd.Root().PersistentFlags().VisitAll(func(flag *pflag.Flag) {
+		out.String(flagUsage(flag))
+	})
 }
 
 func flagUsage(flag *pflag.Flag) string {


medyagh:
@linkvt thank you for fixing this !

PR #7294, opened 2020-03-28
Improve localpath test coverage

n0npax:
Improve test coverage for `pkg/minikube/localpath`


pkg/minikube/localpath/localpath_test.go
```diff
@@ -61,3 +66,95 @@ func TestHasWindowsDriveLetter(t *testing.T) {
...
+func TestMiniPath(t *testing.T) {
+	var testCases = []struct {
+		env, basePath string
+	}{
+		{"/tmp/.minikube", "/tmp/"},
+		{"/tmp/", "/tmp"},
+		{"", homedir.HomeDir()},
+	}
+	for _, tc := range testCases {
+		originalEnv := os.Getenv(MinikubeHome)
```
medyagh: this originalEnv needs to be before the for loop, otherwise on the second iteration it will have the Wrong original value.

n0npax: good catch. Thanks! Fixing

pkg/minikube/localpath/localpath_test.go
```diff
@@ -61,3 +66,95 @@ func TestHasWindowsDriveLetter(t *testing.T) {
...
+	var testCases = []struct {
+		env, basePath string
+	}{
+		{"/tmp/.minikube", "/tmp/"},
+		{"/tmp/", "/tmp"},
+		{"", homedir.HomeDir()},
+	}
+	originalEnv := os.Getenv(MinikubeHome)
+	for _, tc := range testCases {
+		defer func() { // revert to pre-test env var
```
medyagh: The defer func needs to be before the for loop

n0npax: fixed

PR #7299, opened 2020-03-28
Run dashboard with internal kubectl if not in path

### LINKED ISSUE #7290: dashboard: fetch and run the built-in kubectl version
This would improves our getting started experience.

afbjorklund:
If "kubectl" is not in the PATH, then use the same cached binary
as with the "minikube kubectl" command (version matching cluster).

Closes #7290

### DIFF
--- a/cmd/minikube/cmd/dashboard.go
+++ b/cmd/minikube/cmd/dashboard.go
@@ -63,10 +63,8 @@ var dashboardCmd = &cobra.Command{
 			}
 		}
 
-		kubectl, err := exec.LookPath("kubectl")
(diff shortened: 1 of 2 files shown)


afbjorklund:
I think the name "mustload" is quite ugly. Maybe should accept a kubernetes "version" instead ?

```go
func KubectlCommand(co mustload.ClusterController, args ...string) (*exec.Cmd, error) {
	version := co.Config.KubernetesConfig.KubernetesVersion
```

tstromberg:
Tested this locally and it ran great. Thank you!

PR #7308, opened 2020-03-29
Add Metal LB addon

laozc:
This PR configures Metal LB for an ARP based Load Balancer in local network.
This addon was verified with both Hyper-V and KVM driver.

Before using the addons, you may need to use `minikube ip` to get the details for the subnet to find out which IP addresses can be used for the LB to avoid IP address collision.
```bash
$ minikube ip
192.168.39.225
```

You may then use the following command to configure and enable the addon.
```bash
$ minikube addons configure metallb
-- Enter Load Balancer Start IP: 192.168.39.100
-- Enter Load Balancer End IP: 192.168.39.200
✅  metallb was successfully configured
$ ./minikube addons enable metallb
✅  The 'metallb' addon is enabled
```

Then you may create a service with type `LoadBalancer`.
```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: nginx
  namespace: default
  labels:
    app: nginx
spec:
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: nginx
  type: LoadBalancer
```

Eventually the service will obtain an IP.
```bash
$ kubectl get service
NAME         TYPE           CLUSTER-IP      EXTERNAL-IP      PORT(S)        AGE
kubernetes   ClusterIP      10.96.0.1       <none>           443/TCP        117m
nginx        LoadBalancer   10.103.15.133   192.168.39.100   80:30063/TCP   113m
```

You may access the External IP on your host.
```bash
$ curl 192.168.39.100:80
<!DOCTYPE html>
<html>
<head>
<...
</html>
```


laozc:
It would be difficult to detect the right values for the load balancer IP since it may vary in the user's own network environment.
Removing those flags from `start` should be fine as long as the user uses `configure` to set the start/end IP.
Let me

medyagh:
I have been thinking about metallb for ingress or load balancer ( alternative to tunnel)

@laozc do you mind cleariying which driver have you tested this add-on with ?
And does it work on docker driver on linux and also docker driver on macOs

laozc:
I'm testing it with Hyper-V and KVM drivers and both work fine.
You can access the assigned service IP without tunneling since the service IP is bound to the MAC address by Metal LB.
I haven't done any test on macOS nor the docker driver yet and not sure if it would work on these scenarios.

medyagh:
> sting it with Hyper-V and KVM drivers and both work fine.
> You can access the assigned service IP without tunneling since the service IP is bound to the MAC address by Metal LB.
> I haven't done any test on macOS nor the docker driver yet a

do u mind also putting an example of the metal lb in the PR description ?

and how would this be different than our current minikube tunnel command ?

laozc:
The major change is that you don't need to run `minikube tunnel`.
Metal LB sets the ARP table for the exposed service to ensure the service IP is accessible from your machine.

You may check on the updated PR description.

pkg/minikube/config/types.go
```diff
@@ -69,19 +69,21 @@ type ClusterConfig struct {
...
+	APIServerNames      []string
+	APIServerIPs        []net.IP
+	DNSDomain           string
+	ContainerRuntime    string
+	CRISocket           string
+	NetworkPlugin       string
+	FeatureGates        string // https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
+	ServiceCIDR         string // the subnet which kubernetes services will be deployed to
+	ImageRepository     string
+	LoadBalancerStartIP string
```
tstromberg, line 84: Please add a comment that the LoadBalancer IP's are currently only used by MetalLB

PR #7317, opened 2020-03-30
Fix copy & paste errors in Podman driver setup docs

mestaritonttu:
Docs referred to docker while they should have referred to podman


mestaritonttu:
Closing, don't have time for CLA

PR #7324, opened 2020-03-30
docs: Update hugo version used by netlify

tstromberg:
We haven't updated in 9 months. 

NOTE: `make site` uses the latest version of hugo, which could lead to confusion.


tstromberg:
It appears that this causes some breakage, will address before proceeding:

https://deploy-preview-7324--kubernetes-sigs-minikube.netlify.com/

PR #7339, opened 2020-03-31
Correct typo "Retarting" into Restarting"

jpsalvesen:
Correct a user-exposed typo


tstromberg:
This was fixed at head already - these were merely the old translation traces of the typo, that have now also been removed at head.

PR #7345, opened 2020-03-31
add delete-on-failure flag

### LINKED ISSUE #7233: start: Add flag to signify that it's OK to delete broken clusters
Exact flag name TBD, but for embedded use cases, it isn't as important to preserve user data as it is to guarantee that a cluster is created somehow. 

We tried this as a default in v1.5, but users did not like the behavior.

sharifelgamal:
Fixes #7233

Example:
```
out/minikube start
😄  minikube v1.9.0 on Darwin 10.14.6

💣  startup failed: OH NO RANDOM FAILURE

out/minikube start --delete-on-failure
😄  minikube v1.9.0 on Darwin 10.14.6
❗  Node m01 failed to start, deleting and trying again.
🐳  Preparing Kubernetes v1.18.0 on Docker 19.03.8 ...
```


medyagh:
lint

cmd/minikube/cmd/start.go
```diff
@@ -177,6 +178,7 @@ func initMinikubeFlags() {
 	startCmd.Flags().Bool(installAddons, true, "If set, install addons. Defaults to true.")
 	startCmd.Flags().IntP(nodes, "n", 1, "The number of nodes to spin up. Defaults to 1.")
 	startCmd.Flags().Bool(preload, true, "If set, download tarball of preloaded images if available to improve start time. Defaults to true.")
+	startCmd.Flags().Bool(deleteOnFailure, false, "If set, delete the current cluster if start fails and try again.")
```
medyagh: add "Defaults to false." like other flags comments

cmd/minikube/cmd/start.go
```diff
@@ -353,7 +355,27 @@ func runStart(cmd *cobra.Command, args []string) {
...
+		// Start failed, delete the cluster and try again
+		profile, err := config.LoadProfile(cc.Name)
+		if err != nil {
+			out.ErrT(out.Meh, `"{{.name}}" profile does not exist, trying anyways.`, out.V{"name": cc.Name})
+		}
+
+		err = deleteProfile(profile)
+		if err != nil {
+			// We failed to delete? That's not good. Error out.
+			exit.WithError("deleting cluster failed", err)
```
tstromberg, line 370: I feel like there is nothing to lose by trying `node.Start` again at this point. Can this be an `out.Warning` instead of an exit?

cmd/minikube/cmd/start.go
```diff
@@ -353,7 +355,27 @@ func runStart(cmd *cobra.Command, args []string) {
 		}
 	}
 
-	kubeconfig := node.Start(cc, n, existingAddons, true)
+	kubeconfig, err := node.Start(cc, n, existingAddons, true)
+	if err != nil && viper.GetBool(deleteOnFailure) {
+		out.T(out.Warning, "Cluster {{.name}} failed to start, deleting and trying again.", out.V{"name": cc.Name})
```
tstromberg, line 360: Technically, it's a node start that failed. Should this output n.Name instead?

cmd/minikube/cmd/start.go

cmd/minikube/cmd/start.go
sharifelgamal: makes sense, done

cmd/minikube/cmd/start.go
sharifelgamal: yep, done

cmd/minikube/cmd/node_start.go
```diff
@@ -50,7 +50,10 @@ var nodeStartCmd = &cobra.Command{
 		}
 
 		// Start it up baby
```
medyagh, line 52: let keep the comments standard

sharifelgamal: that was already there

cmd/minikube/cmd/node_start.go
```diff
@@ -50,7 +50,10 @@ var nodeStartCmd = &cobra.Command{
...
-		node.Start(*cc, *n, nil, false)
+		_, err = node.Start(*cc, *n, nil, false)
```
medyagh, line 53: would if this node.Start would it exit and not r recreate on failure ?

sharifelgamal: I've fixed it to retry anytime node.Start fails anywhere

cmd/minikube/cmd/node_add.go
```diff
@@ -56,21 +54,18 @@ var nodeAddCmd = &cobra.Command{
...
-			exit.WithError("Error adding node to cluster", err)
+			maybeDeleteAndRetry(*cc, n, nil, err)
 		}
 
 		out.T(out.Ready, "Successfully added {{.name}} to {{.cluster}}!", out.V{"name": name, "cluster": cc.Name})
 	},
 }
 
 func init() {
+	//We should figure out which minikube start flags to actually import
```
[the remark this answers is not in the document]
sharifelgamal: done

PR #7353, opened 2020-04-01
[Trival] beta.kubernetes.io/os is already deprecated since v1.14, are  targeted for removal in v1.18

wawa0210:
/kind cleanup

Ref: kubernetes/kubernetes#89477

/assign @tstromberg

### DIFF
--- a/deploy/addons/dashboard/dashboard-dp.yaml
+++ b/deploy/addons/dashboard/dashboard-dp.yaml
@@ -57,7 +57,7 @@ spec:
             runAsGroup: 2001
       serviceAccountName: kubernetes-dashboard
       nodeSelector:
-        "beta.kubernetes.io/os": linux
+        "kubernetes.io/os": linux
       # Comment the following tolerations if Dashboard must not be deployed on master
       tolerations:
         - key: node-role.kubernetes.io/master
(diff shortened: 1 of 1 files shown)


tstromberg:
I'm curious if this breaks dashboard for --kubernetes-version=v1.11.x users.

tstromberg:
I did some local testing, and this PR does break dashboard deployments to Kubernetes versions older than v1.14 - versions that minikube currently supports.

Let's revisit this when Kubernetes v1.19 ships without this node selector. We can then consider whether to restrict the `dashboard` command to Kubernetes v1.14 releases and newer, or if we should search and replace the node selector based on the Kubernetes release.

wawa0210:
> I did some local testing, and this PR does break dashboard deployments to Kubernetes versions older than v1.14 - versions that minikube currently supports.
> 
> Let's revisit this when Kubernetes v1.19 ships without this node selector. We can then consider whether to restrict the `dashboard` command to Kubernetes v1.14 releases and newer, or if we should search and replace the node selector based on the Kubernetes release.

Sorry, I thought minikube was always the latest dashboard, because I just changed this in the dashboard project https://github.com/kubernetes/dashboard/pull/5045

If minikube needs to adapt to versions below 1.14, I think this pr can also be closed first, and then observed later

PR #7356, opened 2020-04-01
Use kubectl version --short if --output=json fails

### LINKED ISSUE #7350: CentOS 7 kubectl is too old for the --output argument
CentOS 8 with `kubernetes-client-1.5.2-0.7.git269f928.el7.x86_64.rpm`

installs a kubectl that does NOT include an "--output" argument. So,

```
$ kubectl version --client --output=json
Error: unknown flag: --output

Options:
  -c, --client=false: Client version only (no server required).
      --short=false: Print just the version number.

Usage:
  kubectl version [options]

Use "kubectl options" for a list of global command-line options (applies to all commands).

Error: unknown flag: --output
```

and `minikube start` ends with

```
E0331 21:01:18.230181   16751 start.go:383] kubectl info: exec: exit status 1
```

Since start.go is viewable, I go look at `start.go` line 383. That points me to the function `showKubectlInfo` at line 423 , and the `exec.Command` at line 436. From that, I'm concluding that the error is a false positive due to attempting a deprecated command argument.

afbjorklund:
Really old kubernetes clients did not have the --output parameter

Show a friendly reminder to upgrade, instead of an error message.

Closes #7350

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -433,22 +433,12 @@ func showKubectlInfo(kcs *kubeconfig.Settings, k8sVersion string, machineName st
 		return nil
 	}
 
-	j, err := exec.Command(path, "version", "--client", "--output=json").Output()
+	gitVersion, err := kubectlVersion(path)
 	if err != nil {
-		return errors.Wrap(err, "exec")
+		return err
 	}
 
-	cv := struct {
-		ClientVersion struct {
-			GitVersion string `json:"gitVersion"`
-		} `json:"clientVersion"`
-	}{}
-	err = json.Unmarshal(j, &cv)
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -467,6 +457,31 @@ func showKubectlInfo(kcs *kubeconfig.Settings, k8sVersion string, machineName st
 	return nil
 }
 
+func kubectlVersion(path string) (string, error) {
+	j, err := exec.Command(path, "version", "--client", "--output=json").Output()
+	if err != nil {
+		// really old kubernetes clients did not have the --output parameter
```
medyagh, line 463: just a nit, wouldn't hurt to info log the error.

PR #7362, opened 2020-04-01
Make live cache pushes non-fatal in case of stop/delete race

tstromberg:
Addresses `TestGvisorAddon` failures due to other cluster being deleted simultaneously to the test:

https://storage.googleapis.com/minikube-builds/logs/7279/f7f9498/KVM_Linux.html#fail_TestGvisorAddon

### DIFF
--- a/pkg/minikube/machine/cache_images.go
+++ b/pkg/minikube/machine/cache_images.go
@@ -219,8 +219,8 @@ func CacheAndLoadImages(images []string) error {
 		}
 	}
 
-	glog.Infof("succeeded pushing to: ", strings.Join(succeeded, " "))
-	glog.Infof("failed pushing to: ", strings.Join(failed, " "))
+	glog.Infof("succeeded pushing to: %s", strings.Join(succeeded, " "))
+	glog.Infof("failed pushing to: %s", strings.Join(failed, " "))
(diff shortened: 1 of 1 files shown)


pkg/minikube/machine/cache_images.go
```diff
@@ -158,49 +159,70 @@ func needsTransfer(imgClient *client.Client, imgName string, cr cruntime.Manager
...
+				continue
 			}
+
 			if status == state.Running.String() { // the not running hosts will load on next start
 				h, err := api.Load(m)
 				if err != nil {
-					return err
+					glog.Errorf("Failed to load machine %q: %v", m, err)
+					failed = append(failed, pName)
+					continue
```
medyagh, line 206: I had seen this api.Load failing other tests as well. is there a chance to see what other test is failing on this and do the same for them ?

tstromberg: This only impacts code which loads non-primary profiles. 

Looking through our code-base, this is the only example our integration tests hit -- the other is `delete --all`. If you see it elsewhere, please open an issue.

medyagh:
wow KVM tests are down to 0 !!! nice job

PR #7364, opened 2020-04-01
add missing kubelet eviction for v1beta2

medyagh:
- the PR that added v1beta2 missed to add  the kubelet config and proxy config  to v1beta2
- could solve the issue mentioned here https://kubernetes.slack.com/archives/C1F5CT6Q1/p1585667156186900

### This PR: 
- makes GH none failures Zero
- reduces GH docker driver failures from 20 to 9


pkg/minikube/bootstrapper/bsutil/ktmpl/v1beta1.go
```diff
@@ -73,11 +73,17 @@ networking:
 ---
 apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
-imageGCHighThresholdPercent: 100
+imageGCHighThresholdPercent: 70
```
tstromberg, line 76: This seems really low - significantly lower than the default of 85%.  Can you add a comment here or in the PR description as to how you arrived at this number?

I'll also note that we intentionally disabled this because `none` driver users complained. I wouldn't care about them, but this comment has me concerned:

https://github.com/kubernetes-sigs/kind/blob/4c7854c1347860ef04a7a8caef625ac4491afa4b/pkg/cluster/internal/patch/kubeyaml_test.go#L149

Can you run some tests to ensure that this setting does not result in data loss within the host docker?

medyagh: I could set it as 85%
by host docker do you mean the user's docker ? why would this affect the user's docker ?

tstromberg: @medyagh - I'm OK with custom handling for the none driver, but please verify that this does not generate data loss for non-none drivers. That comment in kind scares me.

Issue #7371, opened 2020-04-02
auto selet does not select none on a linux without docker


medyagh:
on a linux box that doesnt have docker
but has podman

```
medya@podman-exp-temp:~$ minikube start 
😄  minikube v1.9.0 on Debian 10.3
💣  Unable to determine a default driver to use. Try specifying --driver, or see https://minikube.sigs.k8s.io/docs/
start/
medya@podman-exp-temp:~$ docker version
-bash: docker: command not found
medya@podman-exp-temp:~$ un
unalias              unattended-upgrades  uniq                 unmkinitramfs        until                
uname                uncompress           unlink               unset                unxz                 
unattended-upgrade   unexpand             unlzma               unshare              
medya@podman-exp-temp:~$ uname -a
Linux podman-exp-temp 4.19.0-6-cloud-amd64 #1 SMP Debian 4.19.67-2+deb10u2 (2019-11-11) x86_64 GNU/Linux
```

we should start either none or podman (experimental) as oppsoed to no running anything

tstromberg:
`none` was an intentional decision, and also requires `docker`.

podman seems reasonable.

tstromberg:
I looked at the code, and think it should have fallen-back anyways. Do you mind running `--alsologtostderr`?

PR #7379, opened 2020-04-02
Add rejection reason to 'unable to find driver' error

tstromberg:
Our current error message is not as helpful as it could be:

`💣  Unable to determine a default driver to use. Try specifying --driver, or see https://minikube.sigs.k8s.io/docs/start/`

Here is the new error message this PR proposes:

```
    ▪ parallels: Not installed: exec: "docker-machine-driver-parallels": executable file not found in $PATH
    ▪ vmware: Not installed: exec: "docker-machine-driver-vmware": executable file not found in $PATH
👉  Try specifying a --driver, or see https://minikube.sigs.k8s.io/docs/start/
```

Solves mysteries such as #7371

### DIFF
--- a/pkg/minikube/driver/driver.go
+++ b/pkg/minikube/driver/driver.go
@@ -202,14 +202,14 @@ func Suggest(options []registry.DriverState) (registry.DriverState, []registry.D
 	for _, ds := range options {
 		if ds != pick {
 			glog.Errorf("%s: %s", ds.Name, ds.Rejection)
-			if !ds.State.Healthy {
-				ds.Rejection = fmt.Sprintf("Not healthy: %v", ds.State.Error)
+			if !ds.State.Installed {
+				ds.Rejection = fmt.Sprintf("Not installed: %v", ds.State.Error)
 				rejects = append(rejects, ds)
 				continue
 			}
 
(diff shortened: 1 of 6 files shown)


pkg/minikube/driver/driver.go
```diff
@@ -198,17 +198,30 @@ func Suggest(options []registry.DriverState) (registry.DriverState, []registry.D
 	}
 
 	alternates := []registry.DriverState{}
+	rejects := []registry.DriverState{}
 	for _, ds := range options {
 		if ds != pick {
-			if !ds.State.Healthy || !ds.State.Installed {
+			glog.Errorf("%s: %s", ds.Name, ds.Rejection)
```
sharifelgamal, line 204: I feel like all this logic could be done in one loop instead of two, though I'm not sure if there's a real advantage of either strategy.

Otherwise this PR is great.

tstromberg: I agree, but I wasn't smart enough to figure out how without making it overly complicated.

We aren't sure who to pick until the end of the first loop - so no matter what, we need to iterate twice to part out the alternates. I'l leave this for someone better at reasoning about algorithms.

PR #7385, opened 2020-04-02
Automatically generate commands docs and add unit test

priyawadhwa:
The unit test makes sure that these docs are always up to date.

I made two significant changes in this PR:

1. We need to pass in -ldflags to `go test`, since we need to know the current iso version to correctly generate the description for --iso-url

2. We need to rewrite the descriptions of the --driver flag and --mount-string flag on `minikube start` because they are operating system specific, which doesn't make sense for our documentation


site/content/en/docs/Reference/Commands/generate-docs.md
```diff
@@ -0,0 +1,50 @@
+---
```
tstromberg, line 1: This doc should be hidden because the `generate-docs` command is hidden.

priyawadhwa: done, thanks

Makefile
```diff
@@ -268,11 +268,11 @@ integration-versioned: out/minikube ## Trigger minikube integration testing
...
 .PHONY: test
 test: pkg/minikube/assets/assets.go pkg/minikube/translate/translations.go ## Trigger minikube test
-	./test.sh
+	MINIKUBE_LDFLAGS="${MINIKUBE_LDFLAGS}" ./test.sh
 
-.PHONY: gotest
-gotest: $(SOURCE_GENERATED) ## Trigger minikube test
-	go test -tags "$(MINIKUBE_BUILD_TAGS)" $(MINIKUBE_TEST_FILES)
+.PHONY: generate-docs
+generate-docs: out/minikube
```
medyagh, line 274: add ## comment here

pkg/generate/docs_templates.go
```diff
@@ -0,0 +1,29 @@
...
+limitations under the License.
+*/
+
+package generate
+
+var title = `---
+title: "{{.Command}}"
+linkTitle: "{{.Command}}"
+weight: 1
+date: {{.Date}}
```
tstromberg, line 23: Remove date: it isn't used and adds unnecessary change churn.

pkg/generate/docs_templates.go
```diff
@@ -0,0 +1,29 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package generate
+
+var title = `---
+title: "{{.Command}}"
+linkTitle: "{{.Command}}"
+weight: 1
```
tstromberg, line 22: Remove weight tag

pkg/generate/docs_templates.go
```diff
@@ -0,0 +1,29 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package generate
+
+var title = `---
+title: "{{.Command}}"
+linkTitle: "{{.Command}}"
```
tstromberg, line 21: Remove linkTitle tag and allow it to default to title

pkg/generate/rewrite.go
```diff
@@ -0,0 +1,71 @@
...
+func rewriteFlag(contents, flag, description string) string {
+	lines := strings.Split(contents, "\n")
+	for i, l := range lines {
+		if strings.Contains(l, flag) {
+			// docs start with a prefix of 6 spaces
+			replacement := "      "
+			replacement += flag
+			// there are 36 spaces between the start of the flag name
+			// and the description
+			spacesBetween := 36 - len(flag)
```
tstromberg, line 62: (This seems super sketchy)

pkg/generate/docs_templates.go
priyawadhwa: done

pkg/generate/rewrite.go
priyawadhwa: Agreed, was able to remove it by changing descriptions of flags in the Flag object before using cobra's markdown generator instead of after

priyawadhwa:
Hey @tstromberg, PTAL. I addressed review comments and opened #7455 to track future PRs

PR #7389, opened 2020-04-02
fallback to alternate drivers on failure

### LINKED ISSUE #7216: Fall-through to next driver if host creation fails
On some systems, the preferable driver may be broken in a way that we are unable to detect until host creation. 

If cluster creation fails, due to permissions issues for instance, we should delete the cluster and fall through to the next driver.

sharifelgamal:
Fixes #7216 

If the user doesn't explicitly supply a choice of driver to `minikube start` via the `driver` flag or config and there's a failure, we will now go down the list of alternate driver choices and retry starting the cluster with that.

```
out/minikube start
😄  minikube v1.9.2 on Darwin 10.14.6
❗  Startup with hyperkit driver failed, trying with docker: OH NO RANDOM FAILURE
🚜  Pulling base image ...
🔥  Creating Kubernetes in docker container with (CPUs=2) (4 available), Memory=1989MB (1989MB available) ...
```

Specifying a driver will still fail as expected
```
out/minikube start --driver=hyperkit
😄  minikube v1.9.2 on Darwin 10.14.6

💣  error provisioning host: OH NO RANDOM FAILURE

```

Starting Kubernetes fails as expected as well, we only fallback on provisioning failures: 
```
😄  minikube v1.9.2 on Darwin 10.14.6

💣  failed to start node: startup failed: OH NO RANDOM FAILURE

```


cmd/minikube/cmd/start.go
```diff
@@ -304,11 +304,43 @@ func runStart(cmd *cobra.Command, args []string) {
 	}
 
 	validateSpecifiedDriver(existing)
-	ds := selectDriver(existing)
+	ds, alts, specified := selectDriver(existing)
+	err = startWithDriver(cmd, ds, existing)
+	if err != nil && !specified {
+		// Walk down the rest of the options
+		for _, alt := range alts {
+			out.WarningT("Startup with {{.old_driver}} driver failed, trying with {{.new_driver}}.", out.V{"old_driver": ds.Name, "new_driver": alt.Name})
```
tstromberg, line 312: Let's start by outputting the error message. It's unlikely to be terribly long, and it should be very informative.

cmd/minikube/cmd/start.go
```diff
@@ -304,11 +304,43 @@ func runStart(cmd *cobra.Command, args []string) {
...
+			ds = alt
+			// Delete the existing cluster and try again with the next driver on the list
+			profile, err := config.LoadProfile(ClusterFlagValue())
+			if err != nil {
+				out.ErrT(out.Meh, `"{{.name}}" profile does not exist, trying anyways.`, out.V{"name": ClusterFlagValue()})
```
tstromberg, line 317: I think this can be hidden in a log message.

cmd/minikube/cmd/start.go
```diff
@@ -388,6 +423,8 @@ func runStart(cmd *cobra.Command, args []string) {
 	if err := showKubectlInfo(kubeconfig, k8sVersion, cc.Name); err != nil {
```
tstromberg, line 423: I think this can be moved out of `startWithDriver` and into the main loop. It raises no error, but it also has nothing to do with drivers.

medyagh:
This is a super neat feature! Thank you. 
I agree with Thomas on only falling back to the other driver if it's a provisioning problem.

sharifelgamal:
Makes sense, it's just going to require a more substantial refactor of node.Start. I'll work on it.

sharifelgamal:
This should be ready for review again.

tstromberg:
Looks great!

Thought for future development: I wonder if we should remember the default driver that finally succeeds. I guess we sort of do, because we only drop the driver if the profile is deleted.

tstromberg:
I tested this PR locally by:

* Creating /tmp/hyperkit as:

```
#!/bin/sh
echo "hyperkit: v0.20190802-4-g97f091"
```
* chmod 755 /tmp/hyperkit
* `./out/minikube delete`
* `./out/minikube start`

Unfortunately, it didn't seem to fallback in this case: https://gist.github.com/tstromberg/bdd8f5c3a54ab2f4ec5ef5570fcc1d46

Any chance you could add a fallback integration test to make sure this doesn't get accidentally broken in the future? It's somewhat tricky, but so is `start.go`. We may want to set the whitelist up to only work in environments where we're sure of a fallback -- so perhaps hyperkit -> Docker for Desktop, or something on Linux?

sharifelgamal:
Good catch, it's because part of provision are still exiting instead of returning the error. I'll fix it. I can try to add a test as well, my worry is that it will be brittle.

As for making sure for fallback, I figured the list of alternate drivers that we return was a pretty safe list to iterate down, any reason to not think so?

tstromberg:
> Good catch, it's because part of provision are still exiting instead of returning the error. I'll fix it. I can try to add a test as well, my worry is that it will be brittle.

Let's only enable it for the 1 or 2 scenarios where we can guarantee it isn't brittle then. I'd rather a stable test that only runs on a single platform than a flaky one. See what we did with gvisor, for instance.
 
> As for making sure for fallback, I figured the list of alternate drivers that we return was a pretty safe list to iterate down, any reason to not think so?

Seems reasonable. I recommend re-using the auto-detection to figure out if there are viable alternatives to fallback to.

sharifelgamal:
Ok, provisioning should never bomb now. I'm still working on a test.

tstromberg:
Please merge when ready.

Issue #7347, opened 2020-04-01
hyperv: The requested operation requires elevation.


soumilshah1995:
**Steps to reproduce the issue:** 
![image](https://user-images.githubusercontent.com/39345855/78090785-1bb10d00-7399-11ea-90a4-22f8893fecd0.png)

1. i have mini kube installed and configure when i run the commands minikube start 

$ minikube start

* minikube v1.9.0 on Microsoft Windows 10 Pro 10.0.18363 Build 18363
* Using the hyperv driver based on existing profile

! 'hyperv' driver reported an issue: C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe Get-WindowsOptionalFeature -FeatureName Microsoft-Hyper-V-All -Online failed:
Get-WindowsOptionalFeature : The requested operation requires elevation.
At line:1 char:1
+ Get-WindowsOptionalFeature -FeatureName Microsoft-Hyper-V-All -Online
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Get-WindowsOptionalFeature], COMException
    + FullyQualifiedErrorId : Microsoft.Dism.Commands.GetWindowsOptionalFeatureCommand

* Suggestion: Start PowerShell as Administrator, and run: 'Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Hyper-V -All'
* Documentation: https://minikube.sigs.k8s.io/docs/reference/drivers/hyperv/

X hyperv does not appear to be installed, but is specified by an existing profile. Please run 'minikube delete' or install hyperv

tstromberg:
Does this happen when you start Powershell as an Administrator, and run minikube from there?

jpsalvesen:
Experiencing the same issue.

```
PS C:\Users\username> minikube.exe start --driver=hyperv
* minikube v1.9.2 on Microsoft Windows 10 Enterprise 10.0.18363 Build 18363
* Using the hyperv driver based on user configuration

! 'hyperv' driver reported an issue: C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe Get-WindowsOptionalFeatur
e -FeatureName Microsoft-Hyper-V-All -Online failed:
Get-WindowsOptionalFeature : The requested operation requires elevation.
At line:1 char:1
[...]

Issue #7300, opened 2020-03-28
hyperv: Unable to remove machine directory: ....vmcx: The process cannot access the file because it is being used by another process.


tstromberg:
I have the same problem and tried this:

```
> minikube stop
*
X Unable to stop VM: open C:\Users\Csongor Halmai\.minikube\machines\minikube\config.json: The system cannot find the file specified.
*
* minikube is exiting due to an error. If the above message is not useful, open an issue:
  - https://github.com/kubernetes/minikube/issues/new/choose
```
and then
```
> minikube delete
* Removing C:\Users\Csongor Halmai\.minikube\machines\minikube ...
*
X Unable to remove machine directory: remove C:\Users\Csongor Halmai\.minikube\machines\minikube\minikube\Snapshots\0113C205-19E6-4E44-9903-D0AA9C750A39.vmcx: The process cannot access the file because it is being used by another process.
*
* minikube is exiting due to an error. If the above message is not useful, open an issue:
  - https://github.com/kubernetes/minikube/issues/new/choose
```
This seems to be show-stopper for me. :(
```
> minikube version
minikube version: v1.9.0
commit: 48fefd43444d2f8852f527c78f0141b377b1e42a
```

_Originally posted by @halmai in https://github.com/kubernetes/minikube/issues/4286#issuecomment-605488971_

tstromberg:
Hey @halmai  - 

I'm very curious about what happened here, because it's not supposed to be possible, but clearly it is:

```
minikube stop
X Unable to stop VM: open C:\Users\Csongor Halmai\.minikube\machines\minikube\config.json: The system cannot find the file specified.
```

minikube was asked to stop a cluster that does not have any local state associated to it, so it failed. Is it possible that you accidentally deleted or moved `C:\Users\Csongor Halmai\.minikube\machines`, or do we have a bug somewhere?

Alternatively, is it possible that you called `stop` with a pre-v1.9.0 version of minikube first? We didn't handle Hyper-V deletion so well before.

```
X Unable to remove machine directory: remove C:\Users\Csongor Halmai\.minikube\machines\minikube\minikube\Snapshots\0113C205-19E6-4E44-9903-D0AA9C750A39.vmcx: The process cannot access the file because it is being used by another process.
*
```

This is stating that Hyper-V does in fact have a virtual machine running from this directory, even if the minikube state is gone. This is recoverable by launching the Hyper-V console and shutting the minikube VM down. Then `minikube delete` should function as expected.

[...]

Issue #7074, opened 2020-03-17
virtualbox: AMD-V is disabled in the BIOS (or by the host OS)


SamarGooda:
**The exact command to reproduce the issue**:
minikube start --driver=virtualbox

**The full output of the command that failed**:<details>
😄  minikube v1.8.2 on Fedora 31
✨  Using the virtualbox driver based on user configuration
🔥  Creating virtualbox VM (CPUs=2, Memory=2200MB, Disk=20000MB) ...

💣  Unable to start VM. Please investigate and run 'minikube delete' if possible: creating host: create: creating: Unable to start the VM: /usr/bin/VBoxManage startvm minikube --type headless failed:
VBoxManage: error: AMD-V is disabled in the BIOS (or by the host OS) (VERR_SVM_DISABLED)
VBoxManage: error: Details: code NS_ERROR_FAILURE (0x80004005), component ConsoleWrap, interface IConsole

Details: 00:00:00.666308 Power up failed (vrc=VERR_SVM_DISABLED, rc=NS_ERROR_FAILURE (0X80004005))

</details>

**The output of the `minikube logs` command**:<details>

💣  command runner: getting ssh client for bootstrapper: Error dialing tcp via ssh client: dial tcp 127.0.0.1:22: connect: connection refused

</details>

**The operating system version**:
fedora 31

runlevel5:
@SamarGooda would you run into the same error after AMD-V is enabled in the BIOS?

priyawadhwa:
Hey @SamarGooda -- any update here?

Issue #7071, opened 2020-03-17
podman on ppc64le: Error: incompatible image architecture "amd64" on "ppc64le" host


runlevel5:
```
[root@orion tle]# minikube delete
🙄  "minikube" profile does not exist, trying anyways.
💀  Removed all traces of the "minikube" cluster.
[root@orion tle]# minikube delete
🙄  "minikube" profile does not exist, trying anyways.
💀  Removed all traces of the "minikube" cluster.
[root@orion tle]# rm -rf /tmp/minikube*
[root@orion tle]# rm -rf /tmp/juju*
[root@orion tle]# /usr/local/bin/minikube --alsologtostderr start --driver=podman
I0317 22:40:42.091465   10546 notify.go:125] Checking for updates...
I0317 22:40:42.369365   10546 start.go:252] hostinfo: {"hostname":"orion.dev","uptime":39818,"bootTime":1584405424,"procs":443,"os":"linux","platform":"fedora","platformFamily":"fedora","platformVersion":"32","kernelVersion":"5.6.0-0.rc5.git0.2.fc32.ppc64le","virtualizationSystem":"","virtualizationRole":"","hostid":"5632f077-29a6-48c4-9d05-933910ac9490"}
I0317 22:40:42.381682   10546 start.go:262] virtualization:
😄  minikube v1.8.2 on Fedora 32 (ppc64le)
I0317 22:40:42.381757   10546 driver.go:226] Setting default libvirt URI to qemu:///system
✨  Using the podman (experimental) driver based on user configuration
I0317 22:40:42.570332   10546 start.go:299] selected driver: podman
I0317 22:40:42.570356   10546 start.go:488] validating driver "podman" against <nil>
I0317 22:40:42.570382   10546 start.go:494] status for podman: {Installed:true Healthy:true Error:<nil> Fix: Doc:}
W0317 22:40:42.700205   10546 start.go:852] Unable to query memory limits: get podman system info: exit status 125
I0317 22:40:42.700471   10546 start.go:863] Using suggested 2200MB memory alloc based on sys=-1MB, container=-1MB
I0317 22:40:42.700663   10546 cache.go:97] Beginning downloading kic artifacts
I0317 22:40:42.700707   10546 preload.go:84] Found local preload: /root/.minikube/cache/preloaded-tarball/preloaded-images-k8s-v1-v1.17.3-docker-overlay2.tar.lz4
[...]

PR #7390, opened 2020-04-02
Update err map for v1.9.x

tstromberg:
Adds solution messages for:

#3898
#6070
#4617
#5392
#6366
#6391
#6411
#6596
#6825
#7070
#7071
#7072
#7074
#7256
#7286
#7300
#7347

Updates other solution messages to point users toward --driver=docker instead of --driver=none.

### DIFF
--- a/pkg/minikube/problem/err_map.go
+++ b/pkg/minikube/problem/err_map.go
@@ -55,7 +55,42 @@ var vmProblems = map[string]match{
 		Issues:        []int{6098},
 		ShowIssueLink: true,
 	},
-
+	"FILE_IN_USE": {
+		Regexp: re(`The process cannot access the file because it is being used by another process`),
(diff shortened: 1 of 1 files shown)


pkg/minikube/problem/err_map.go
```diff
@@ -55,7 +55,42 @@ var vmProblems = map[string]match{
...
-
+	"FILE_IN_USE": {
+		Regexp: re(`The process cannot access the file because it is being used by another process`),
+		Advice: "Another program is using a file required by minikube. If you are using Hyper-V, try stopping the minikube VM from within the Hyper-V manager",
+		URL:    "https://docs.docker.com/machine/drivers/hyper-v/",
+		GOOS:   []string{"windows"},
+		Issues: []int{7300},
+	},
+	"CREATE_TIMEOUT": {
+		Regexp: re(`create host timed out in 120`),
```
medyagh, line 66: is the 120 needed ? can we make the regex to search for a digit instead of 120 ?

PR #7393, opened 2020-04-03
allow ingress addon for docker/podman drivers only on linux for now

medyagh:
does not close https://github.com/kubernetes/minikube/issues/7332 but it will at least provide a better message.

## after this PR:
```
💡  Due to docker networking limitations on darwin, ingress addon is not supported for this driver.
Alternatively to use this addon you can use a vm-based driver:

        'minikube start --vm=true'

To track the update on this work in progress feature please check:
https://github.com/kubernetes/minikube/issues/7332```
```

also add docs to the website

meanwhile we will work on implementing this feature for mac an windows.

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -126,7 +126,9 @@ func enableOrDisableAddon(cc *config.ClusterConfig, name string, val string) err
 	// to match both ingress and ingress-dns adons
 	if strings.HasPrefix(name, "ingress") && enable && driver.IsKIC(cc.Driver) && runtime.GOOS != "linux" {
 		exit.UsageT(`Due to {{.driver_name}} networking limitations on {{.os_name}}, {{.addon_name}} addon is not supported for this driver.
-you can use a different driver such as hyperkit or virtualbox or hypev.
+Alternatively to use this addon you can use a vm-based driver:
+
+		'minikube start --vm=true'
 
 To track the update on this work in progress feature please check:
 https://github.com/kubernetes/minikube/issues/7332`, out.V{"driver_name": cc.Driver, "os_name": runtime.GOOS, "addon_name": name})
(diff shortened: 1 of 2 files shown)


pkg/addons/addons.go
```diff
@@ -122,6 +123,11 @@ func enableOrDisableAddon(cc *config.ClusterConfig, name string, val string) err
 		}
 	}
 
+	// to match both ingress and ingress-dns adons
+	if strings.HasPrefix(name, "ingress") && enable && driver.IsKIC(cc.Driver) && runtime.GOOS != "linux" {
+		exit.UsageT("Due to {{.driver_name}} networking limitations on {{.os_name}}, {{.addon_name}} addon is not supported for this driver. meanwhile a different driver such as 'hyperkit' or 'virtualbox' or 'hypev'. To track the update on this feature being worked on please check https://github.com/kubernetes/minikube/issues/7332", out.V{"driver_name": cc.Driver, "os_name": runtime.GOOS, "addon_name": name})
```
sharifelgamal, line 128: Simplify this sentence by saying "a VM-based driver" instead of enumerating specific driver. Something like

"Due to {{.driver_name}} networking limitations on {{.os_name}}, the {{.addon_name}} addon is not currently supported. Try using a VM-based driver instead. Follow updates on this feature at https://github.com/kubernetes/minikube/issues/7332."

medyagh: thanks

reneli5:
I had the same issue using minikube and docker driver on my Mac. I deleted the docker profile and used the virtual box driver. All worked.
Here is how:
1. minikube delete
2. minikube start --vm-driver virtualbox
3. minikube addons enable ingress

fzyzcjy:
Will ingress + docker be supported in the future? Thanks!

PR #7394, opened 2020-04-03
parallelize updating cluster and setting up certs

priyawadhwa:
Reduces time in setupKubeadm function from 4.6s to 3.2s

slowjam before change: http://104.197.255.167:8000

slowjam after change: http://34.67.230.103:8000

### DIFF
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -23,6 +23,7 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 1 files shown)


medyagh:
The following Docker Driver Failures 


### TestStartStop/group/containerd - 251.5s
* known flake 


--- needs to verified that is not related to this PR:
### TestStartStop/group/crio - 759.66s
### TestStartStop/group/embed-certs - 144.29s
### TestFunctional/parallel/ComponentHealth

tstromberg:
How much time does this save?

tstromberg:
This seems good. Looking at SlowJam, `SetupCerts` still blocks for an additional second:

Screen Shot 2020-04-03 at 9 17 42 AM

If we parallelized the `cmd.Copy(f)` call in SetupCerts, we might get to where SetupCerts is no longer the blocker.

Here's a bigger idea, that should result in shaving >5s off of the Docker runtime:

- Update kic_runner to make it possible to re-use a persistent ssh connection rather than `docker cp` / `docker run`.  See `ssh_runner` as an example for a persistent client, where we add sessions for each command.

If you compare the slowjam profiles of hyperkit versus docker, you can see the benefit:

SetupCerts:
- 2.7s on Docker
- 0.6s on hyperkit

Rough napkin math: Each command or copy on Docker costs at least 0.1 seconds. On ssh it doesn't even show up in the SlowJam graph. I reckon we execute and copy ~60 commands. Thus, I think we should be able to expect 5s in time savings.

tstromberg:
For comparison, SetupCerts on hyperkit. Same code, different runner:

Screen Shot 2020-04-03 at 9 57 08 AM

PR #7398, opened 2020-04-03
adjust mysql test on timeout

medyagh:
a small clean up PR:

I hope it closes this https://github.com/kubernetes/minikube/issues/7401 for gh action tests

-  adjusting the time for waiting for mysql test
- adjusting timeout for podman on gh action so it fails faster
- running goimports and gofmt on the whole project to find the missing ones


afbjorklund:
When did we upgrade to go 1.14, the Makefile (and k8s, I believe) still uses 1.13 ?

medyagh:
> When did we upgrade to go 1.14, the Makefile (and k8s, I believe) still uses 1.13 ?

it was just my local machine golang, but I get the same results for gofmt for 1.13

medyagh:
> When did we upgrade to go 1.14, the Makefile (and k8s, I believe) still uses 1.13 ?

AH ! what a shame it seems like goftm is in conflict with goimport !!! if you do gofmt after goimports it will ruin.

Issue #6122, opened 2019-12-19
Document upgrading a cluster


robrich:
Now that https://github.com/kubernetes/minikube/pull/5798 is merged, and `minikube start` doesn't auto-upgrade, how do I upgrade my cluster?  Will `minikube start --kubernetes-version ...` upgrade it or dump the VM and start over?  (and consume all the bandwidth as I download base images again.) Is there a `minikube start --kubernetes-version stable` or `minikube start --kubernetes-version latest` that will auto-upgrade the cluster?

Perhaps we need a docs page to answer this.

tstromberg:
`minikube start --kubernetes-version=v1.17.0` will upgrade you, without re-downloading the VM image. That said, the base VM image is only ~120MB compared to the ~600MB it takes to download a new Kubernetes release.

We don't yet support 'stable' or 'latest' designators, but it would certainly be a great feature to add. If you or anyone else is interested in making this work, add a check for `paramVersion == "stable"` here:

https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/start.go#L1127

As far as docs, I agree. I'd be happy to review any PR that adds it. Here is where I would probably put it: https://github.com/kubernetes/minikube/tree/master/site/content/en/docs/Tasks

tstromberg:
Also worth noting, this is the message users should see if there is an upgrade available:

`out.T(out.ThumbsUp, "Kubernetes {{.new}} is now available. If you would like to upgrade, specify: --kubernetes-version={{.new}}`

That doesn't invalidate the issues you've raised, just showing the guidance we are currently providing, as the documentation should match it.

sharifelgamal:
We've added the documentation for this, I'm going to go ahead and close this. Feel free to reopen if necessary.

life5ign:
Where is this in the docs?  I can't find it.

PR #7399, opened 2020-04-03
Add example showing how to upgrade k8s version

prasadkatti:
Stop-gap solution for #6122 until we have a separate page which discusses how to upgrade a minikube cluster.


prasadkatti:
@tstromberg - I can work on #6122 as well, I think. But I will need help.

Issue #7423, opened 2020-04-05
multi node: service broken


medyagh:
to replicate the issue:

```
kubectl create deployment hello-minikube --image=k8s.gcr.io/echoserver:1.4
kubectl expose deployment hello-minikube --type=NodePort --port=8080
./out/minikube service hello-minikube
```

then in the browser, u would see the router had returned a 404

also in the kubectl get pods -A I can see on the second node the kindnet is not created, and the error is :
```
Events:
  Type     Reason     Age                    From                   Message
  ----     ------     ----                   ----                   -------
  Normal   Scheduled  5m6s                   default-scheduler      Successfully assigned kube-system/kindnet-txxbm to minikube-m02
  Normal   Pulled     2m56s (x12 over 5m4s)  kubelet, minikube-m02  Container image "kindest/kindnetd:0.5.3" already present on machine
  Warning  Failed     2m44s (x13 over 5m4s)  kubelet, minikube-m02  Error: host IP unknown; known addresses: []

```

full commands and their outputs

```
medmac@~/workspace/minikube (paralel_kubeadm) $ ./out/minikube start --vm-driver=docker -n=2
😄  minikube v1.9.2 on Darwin 10.13.6
✨  Using the docker driver based on user configuration
👍  Starting control plane node m01 in cluster minikube
🚜  Pulling base image ...
🔥  Creating Kubernetes in docker container with (CPUs=2) (2 available), Memory=4000MB (7964MB available) ...
🐳  Preparing Kubernetes v1.18.0 on Docker 19.03.2 ...
    ▪ kubeadm.pod-network-cidr=10.244.0.0/16
🌟  Enabling addons: default-storageclass, storage-provisioner

👍  Starting node m02 in cluster minikube
🚜  Pulling base image ...
🔥  Creating Kubernetes in docker container with (CPUs=2) (2 available), Memory=4000MB (7964MB available) ...
🌐  Found network options:
    ▪ NO_PROXY=172.17.0.2
🐳  Preparing Kubernetes v1.18.0 on Docker 19.03.2 ...
🏄  Done! kubectl is now configured to use "minikube"
medmac@~/workspace/minikube (paralel_kubeadm) $ kc get nodes
NAME           STATUS     ROLES    AGE   VERSION
minikube       Ready      master   52s   v1.18.0
minikube-m02   NotReady   <none>   6s    v1.18.0
medmac@~/workspace/minikube (paralel_kubeadm) $ kc get pods
No resources found in default namespace.
[...]

PR #7425, opened 2020-04-05
remove kic overlay network for docker-runtime

medyagh:
After the refactor that we fixed the control-endpoint in the config, we no longer need to use kic overlay network.

## After this PR:
- save 3 seconds by not waiting for applying kic overlay network on the default runtime (docker)
- save disk space by not storing kindnet image 
- make preload images 78mb smaller for docker ( for the next version of preload images)
 
## does this PR affect multi node ?
no.

the mutli node start works fine as is now
```
medmac@~/workspace/minikube (remove_overlay) $ ./out/minikube start --driver=docker -n=2
🚜  Pulling base image ...

👍  Starting node m02 in cluster minikube
🚜  Pulling base image ...
🌐  Found network options:
    ▪ NO_PROXY=172.17.0.2
medmac@~/workspace/minikube (remove_overlay) $ kc get nodes
NAME           STATUS   ROLES    AGE    VERSION
minikube       Ready    master   103s   v1.18.0
minikube-m02   Ready    <none>   56s    v1.18.0
```

other operations of multi node is broken even before this PR see https://github.com/kubernetes/minikube/issues/7423
I belive to make this service work in multi node we currently need flannel (which means we do not use kic overlay network anyways_

# FUTURE PRs:
this PR does not remove kic overaly for containerd and crio as they are still needed till we address https://github.com/kubernetes/minikube/issues/7428


medyagh:
> LGTM -- could you please bump up the preload version [here](https://github.com/kubernetes/minikube/blob/0db42a6f1acd044627d4689123aa8ba7cc757795/pkg/minikube/download/preload.go#L44)? Then this should be good to merge

good idea, I will do that in a separate PR so we can have the images generted first for the tests

PR #7430, opened 2020-04-05
Update translations files by running extract

afbjorklund:
This should probably be part of the release procedure ?

`make extract`

----

I'm also missing some gettext maintenance commands...
* show summary per language
* creating new language (`msginit`)
* remove old unused strings (`msgmerge`)

For reference see https://www.gnu.org/software/gettext/manual/html_node/index.html

Maybe some utility functions could be provided for the Minikube Translations System ?

### DIFF
--- a/translations/de.json
+++ b/translations/de.json
@@ -43,7 +43,6 @@
 	"Basic Commands:": "",
 	"Because you are using docker driver on Mac, the terminal needs to be open to run it.": "",
 	"Bind Address: {{.Address}}": "",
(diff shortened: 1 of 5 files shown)


afbjorklund:
The main goal here was that `git diff` should be empty after running `make extract`.

But there are now some unused translations strings, that "should" be cleaned up.
Some of these needs to be re-translated, because we added a parameter or something.

This is what is referred to as "fuzzy" translations, in the traditional gettext system.

Issue #7448, opened 2020-04-06
soft "start" without args - will overwrite the container-runtime to docker.


medyagh:
in this PR I added an integeration test for Soft Start. in TestStartStop.
and it was revealed that our soft start is broken for not-docker runtimes

## verified locally

verified locally this happens only when on the Second Start you do NOT specify the run time.

So: 
```
minikube start --driver=hyperkit --container-runtime=conrtainerd
minikube start --driver=hyperkit --container-runtime=conrtainerd
```
(this one is fine)
but if you do
```
minikube start 
```
and in the config, you will see, the run time has been changed to docker
```
 $ cat ~/.minikube/profiles/minikube/config.json | grep "ContainerRuntime"
                "ContainerRuntime": "docker",
```

it is not fine !

---
# Test Logs: 
---
it appears that on a restart it generates a WRONG kubeadm config.

	-  criSocket: /run/containerd/containerd.sock
	+  criSocket: /var/run/dockershim.sock

https://storage.googleapis.com/minikube-builds/logs/7435/0f498bc/KVM_Linux.html#fail_TestStartStop%2fgroup%2fcontainerd

```
ing, string) (net.Conn, error))(nil)}
	E0406 03:53:37.937401   13793 kubeadm.go:331] Overriding stale ClientConfig host https://192.168.39.96:8444 with https://192.168.39.96:8443
	I0406 03:53:38.063752   13793 ssh_runner.go:101] Run: sudo diff -u /var/tmp/minikube/kubeadm.yaml /var/tmp/minikube/kubeadm.yaml.new
	I0406 03:53:38.118080   13793 kubeadm.go:404] needs reset: configs differ:
	-- stdout --
	--- /var/tmp/minikube/kubeadm.yaml
	+++ /var/tmp/minikube/kubeadm.yaml.new
	@@ -2,7 +2,7 @@
	 kind: InitConfiguration
	 localAPIEndpoint:
	   advertiseAddress: 192.168.39.96
	-  bindPort: 8444
	+  bindPort: 8443
	 bootstrapTokens:
	   - groups:
	       - system:bootstrappers:kubeadm:default-node-token
	@@ -11,7 +11,7 @@
	       - signing
	       - authentication
	 nodeRegistration:
	-  criSocket: /run/containerd/containerd.sock
	+  criSocket: /var/run/dockershim.sock
	   name: "containerd-20200406T034015.056085983-5464"
	   kubeletExtraArgs:
[...]

PR #7435, opened 2020-04-05
wip: docker/podman: make soft start 20 seconds faster

medyagh:
## Before this PR:
Soft Start:
```
$ time minikube start --driver=docker
😄  minikube v1.9.2 on Darwin 10.13.6
✨  Using the docker driver based on existing profile
👍  Starting control plane node m01 in cluster minikube
🚜  Pulling base image ...
🏃  Updating the running docker "minikube" container ...
🐳  Preparing Kubernetes v1.18.0 on Docker 19.03.2 ...
    ▪ kubeadm.pod-network-cidr=10.244.0.0/16
🌟  Enabling addons: default-storageclass, storage-provisioner
🏄  Done! kubectl is now configured to use "minikube"

real	0m28.200s
user	0m4.165s
sys	0m1.921s
```

## After this PR:

Soft Start:

```
$ make && time ./out/minikube start --driver=docker
make: `out/minikube' is up to date.
😄  minikube v1.9.2 on Darwin 10.13.6
✨  Using the docker driver based on existing profile
👍  Starting control plane node m01 in cluster minikube
🚜  Pulling base image ...
🏃  Updating the running docker "minikube" container ...
🐳  Preparing Kubernetes v1.18.0 on Docker 19.03.2 ...
    ▪ kubeadm.pod-network-cidr=10.244.0.0/16
🌟  Enabling addons: default-storageclass, storage-provisioner
🏄  Done! kubectl is now configured to use "minikube"

real	0m8.949s
user	0m2.953s
sys	0m1.418s
```

## this PR:
- do NOT disable other run-times when it is "docker-runtime on kic drivers"
    because in ubuntu docker service needs contianerd service. https://github.com/kubernetes/minikube/issues/7433#issuecomment-609395145 
(unlike our VM installation that uses runc)
- improve logging, add logging when it doesnt need reset. (to use in test)
- add integration tests for soft start to ensure, we do not Hard Start ever again !
- renaming the preload test and putting in in a different file

closes https://github.com/kubernetes/minikube/issues/7432

PR #7449, opened 2020-04-06
Behavior change: start with no arguments uses existing cluster config

medyagh:
## Change behavior of minikube start
- if minkube already started with args 
like `minikube start --container-runtime=containerd`
a `minikube start`  (without args) on an already started minikube should NOT change the run-time back to the default config ('docker').

- an attempt to fix https://github.com/kubernetes/minikube/issues/7448
unfourntaely, it is not clear since when this is broken but currently the minikube does not respect the profile config.

- also fixes the Port : 0 in our configs .

## before this PR:

minikube start --driver=hyperkit --container-runtime=conrtainerd
Fine.

minikube start --driver=hyperkit --container-runtime=conrtainerd
Fine Also

```
 $ cat ~/.minikube/profiles/minikube/config.json | grep Container
        "ContainerRuntime": "containerd",
```
### but !!!! if you do a soft start without args:
minikube start
 
and in the config, you will see, the run time has been changed to docker
```
 $ cat ~/.minikube/profiles/minikube/config.json | grep "ContainerRuntime"
                "ContainerRuntime": "docker",
```
and minikube disrepects the user and restarts the cluster with a hard start to a different container-runtime.

## Integeration test:
added a SoftStart integeration tests to the functional tests
also in 
https://github.com/kubernetes/minikube/pull/7435 there is additional tests for soft start that validates it doesn not do a reset.

I tested locally on mac hyperkit seems to be working.

I did this PR with a tired mind,  I do feel like I might have missed some flags in the 'func updateExistingConfigFromFlags' . please anyone who reviews this, review with 80% chance of human error on my side on This PR.


medyagh:
@tstromberg added integeration test to the functional tests for it.

test/integration/functional_test.go
```diff
@@ -206,6 +210,38 @@ func validateStartWithProxy(ctx context.Context, t *testing.T, profile string) {
 	}
 }
 
+// validateSoftStart validates that after minikube already started, a "minikube start" should not change the configs.
+func validateSoftStart(ctx context.Context, t *testing.T, profile string) {
+	apiPortBefore := 8441
```
tstromberg, line 215: For clarity, can you make this variable shared between the two tests, preferably with a verbose name?

tstromberg:
Also, since this is an explicit behavioral change, give the PR a really clear title.

sharifelgamal:
Merging start.go and flags.go into start_flags.go makes this change hard to review, and I don't think it's especially necessary. Is there a reason for the change?

test/integration/functional_test.go
medyagh: the handler func signature didnt allow a paramter. but I added it as an global var.

medyagh:
> Merging start.go and flags.go into start_flags.go makes this change hard to review, and I don't think it's especially necessary. Is there a reason for the change?

thank you thats a good point, I will do that on my next PRs.
the reason that I did this in this PR was I had to go back an forth many times in the code to create the new func updateExisting.... and start.go was extremely large, made the work extremely hard to keep looking for start flag variable names. it took a lot of time even with this breaking down

Issue #7431, opened 2020-04-05
Add CoreDNS to wait components


medyagh:
currently the wait components are
'apiserver,system_pods,default_sa'

while investigating https://github.com/kubernetes/minikube/issues/7428
I noticed  we can have another option for --wait

core_dns that we should wait for core_dns pod to be healthy

as seen in a test failure for TestStartStop the core dns was not running.

```

1765879-21439 get po -A --show-labels
helpers.go:223: (dbg) kubectl --context containerd-20200406T121108.591765879-21439 get po -A --show-labels:
NAMESPACE     NAME                                                                 READY   STATUS    RESTARTS   AGE     LABELS
default       busybox                                                              0/1     Pending   0          8m7s    integration-test=busybox
kube-system   coredns-66bff467f8-bds4s                                             0/1     Pending   0          8m9s    k8s-app=kube-dns,pod-template-hash=66bff467f8
kube-system   coredns-66bff467f8-ndb2x                                             0/1     Pending   0          8m9s    k8s-app=kube-dns,pod-template-hash=66bff467f8
kube-system   etcd-containerd-20200406t121108.591765879-21439                      1/1     Running   0          8m11s   component=etcd,tier=control-plane
kube-system   kube-apiserver-containerd-20200406t121108.591765879-21439            1/1     Running   0          8m11s   component=kube-apiserver,tier=control-plane
kube-system   kube-controller-manager-containerd-20200406t121108.591765879-21439   1/1     Running   0          8m11s   component=kube-controller-manager,tier=control-plane
kube-system   kube-proxy-ntq5b                                                     1/1     Running   0          8m8s    controller-revision-hash=c8bb659c5,k8s-app=kube-proxy,pod-template-generation=1
kube-system   kube-scheduler-containerd-20200406t121108.591765879-21439            1/1     Running   0          8m11s   component=kube-scheduler,tier=control-plane
kube-system   storage-provisioner                                                  0/1     Pending   0          8m13s 
```

PR #7460, opened 2020-04-06
add new wait component apps_running

medyagh:
closes https://github.com/kubernetes/minikube/issues/7431

wait=true or wait=all  will wait for a new component called "apps_running"
this should reduce test start stop fails, by waiting for more components 

the new wait components is called apps_running

```
$ ./out/minikube start --help | grep wait
      --wait=[apiserver,system_pods]: comma separated list of kubernetes components to verify and wait for after starting a cluster. defaults to "apiserver,system_pods", available options: "apiserver,system_pods,default_sa,apps_running" . other acceptable values are 'all' or 'none', 'true' and 'false'
```


medyagh:
This PR reveals that our NewestCNI is broken. created an issue for it https://github.com/kubernetes/minikube/issues/7466

PR #7478, opened 2020-04-07
Make KIC start message shorter

afbjorklund:
Don't show the host info any more (available),
since we don't show it for the VM host either.

### DIFF
--- a/pkg/minikube/machine/start.go
+++ b/pkg/minikube/machine/start.go
@@ -32,7 +32,6 @@ import (
 	"github.com/juju/mutex"
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
-	"k8s.io/minikube/pkg/drivers/kic/oci"
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/minikube/driver"
@@ -239,13 +238,7 @@ func showHostInfo(cfg config.ClusterConfig) {
 		return
 	}
 	if driver.IsKIC(cfg.Driver) { // TODO:medyagh add free disk space on docker machine
-		s, err := oci.DaemonInfo(cfg.Driver)
-		if err == nil {
-			var info hostInfo
-			info.CPUs = s.CPUs
-			info.Memory = megs(uint64(s.TotalMemory))
-			out.T(out.StartingVM, "Creating Kubernetes in {{.driver_name}} {{.machine_type}} with (CPUs={{.number_of_cpus}}) ({{.number_of_host_cpus}} available), Memory={{.memory_size}}MB ({{.host_memory_size}}MB available) ...", out.V{"driver_name": cfg.Driver, "number_of_cpus": cfg.CPUs, "number_of_host_cpus": info.CPUs, "memory_size": cfg.Memory, "host_memory_size": info.Memory, "machine_type": machineType})
-		}
+		out.T(out.StartingVM, "Creating {{.driver_name}} {{.machine_type}} (CPUs={{.number_of_cpus}}, Memory={{.memory_size}}MB) ...", out.V{"driver_name": cfg.Driver, "number_of_cpus": cfg.CPUs, "memory_size": cfg.Memory, "machine_type": machineType})
 		return
 	}
 	out.T(out.StartingVM, "Creating {{.driver_name}} {{.machine_type}} (CPUs={{.number_of_cpus}}, Memory={{.memory_size}}MB, Disk={{.disk_size}}MB) ...", out.V{"driver_name": cfg.Driver, "number_of_cpus": cfg.CPUs, "memory_size": cfg.Memory, "disk_size": cfg.DiskSize, "machine_type": machineType})


medyagh:
Can you please post before and after?

afbjorklund:
Before:

```
🚜  Pulling base image ...
🔥  Creating Kubernetes in docker container with (CPUs=2) (4 available), Memory=2
200MB (7800MB available) ...
```

After:

```
🚜  Pulling base image ...
🔥  Creating docker container (CPUs=2, Memory=2200MB, Disk=138379MB) ...
```

----

Compare:
```
🔥  Creating virtualbox VM (CPUs=2, Memory=2200MB, Disk=20000MB) ...
```

```
👍  Starting control plane node  in cluster minikube
ℹ️  OS release is Ubuntu 16.04.6 LTS
```

tstromberg:
I tried this on macOS and got:

`pkg/drivers/kic/oci/info.go:42:26: invalid operation: int64(fs.Blocks) * fs.Bsize (mismatched types int64 and uint32)`

Using `int64(fs.Bsize)` made it work properly.

afbjorklund:
I’m not even sure the disk size is accurate on the VM-based Docker ?

afbjorklund:
https://docs.docker.com/docker-for-mac/space/

`~/Library/Containers/com.docker.docker/Data/vms/0/data/Docker.raw`

https://docs.docker.com/docker-for-windows/install/

`C:\Users\Public\Documents\Hyper-V\Virtual hard disks\DockerDesktop.vhdx`

So the values above are **only** valid for Linux. :-(

----

The problem with `docker system df -v` is that it takes _forever_ to run... (like 15s or so)

And it only shows how much is used, not how much the total is or how much is available.

afbjorklund:
> Using `int64(fs.Bsize)` made it work properly.

golangci-lint had a different opinion, though...

```
pkg/drivers/kic/oci/info.go:42:33: unnecessary conversion (unconvert)
	return int64(fs.Blocks) * int64(fs.Bsize) / 1024 / 1024
	                               ^
```

tstromberg:
OK. Switch it back so the tests work. We can sort out any platform specific quirks afterwards.

```
pkg/drivers/kic/oci/info.go:42:33: unnecessary conversion (unconvert)
	return int64(fs.Blocks) * int64(fs.Bsize) / 1024 / 1024
```

medyagh:
I was thinking how about to keep the available somehow ?
Sometimes its nice to see it user had to give 2gb memory because available was 3 gb...


And also for disk size I was thinking for kic it makes more sense to use docker machine disk size not the system disk size. What do you think ?

tstromberg:
> I was thinking how about to keep the available somehow ?
> Sometimes its nice to see it user had to give 2gb memory because available was 3 gb...

So far, I personally haven't found avail to be useful.

> And also for disk size I was thinking for kic it makes more sense to use docker machine disk size not the system disk size. What do you think ?

I believe this PR does that.

afbjorklund:
The easiest short-term solution is to remove the disk size. Or show the config ?
(sort of implies that there is a check to see that you actually can use the 20G)

medyagh:
> t-term solution is to remove the disk size. Or show the config ?

what do u think about this

become like the kubernetes optiopns

so it will be:

🔥  Creating Kubernetes in docker container 
         ▪  CPUs=2 (2 available)
         ▪  Memory=4000MB (7964MB available) 
         ▪  Disk=20000MB (10000MB available)

afbjorklund:
I don't fully understand why we need to show the word "Kubernetes" twice.

The end goal was to make the start output more similar for different drivers:

`🔥  Creating virtualbox VM (CPUs=2, Memory=2200MB, Disk=20000MB) ...`

`🔥  Creating docker container (CPUs=2, Memory=2200MB, Disk=20000MB) ...`

medyagh:
> I don't fully understand why we need to show the word "Kubernetes" twice.
> 
> The end goal was to make the start output more similar for different drivers:
> 
> `🔥 Creating virtualbox VM (CPUs=2, Memory=2200MB, Disk=20000MB) ...`
> 
> `🔥 Creating docker container (CPUs=2, Memory=2200MB, Disk=20000MB) ...`

I agree that looks much better

medyagh:
> I don't fully understand why we need to show the word "Kubernetes" twice.
> 
> The end goal was to make the start output more similar for different drivers:
> 
> `🔥 Creating virtualbox VM (CPUs=2, Memory=2200MB, Disk=20000MB) ...`
> 
> `🔥 Creating docker container (CPUs=2, Memory=2200MB, Disk=20000MB) ...`

what do u think about using multi lines
🔥  Creating docker container
    ▪  CPUs=2 (2 available)
    ▪  Memory=4000MB (7964MB available) 
    ▪  Disk=20000MB (10000MB available)

afbjorklund:
> what do u think about using multi lines so we can still keep the availables

Sure, but then you should show host info for all drivers (not just docker)

afbjorklund:
Note that I wanted to show the disk info for the docker driver as well.

Just fell a bit short, when it comes to implementing it (cross-platform)

Maybe something similar. but using `ssh df` ?

But very hard to test this, on the VM platforms.

medyagh:
I would like to see that info for all of them. If not possible for a platform for now. We can just skip

afbjorklund:
> I would like to see that info for all of them. If not possible for a platform for now. We can just skip

Should open a new issue, for getting docker disk information on those platforms...
I'm not sure if we have `ssh` access, or if we know how those disk images work ?

For the Minikube VM and SSH, it is easy enough to connect to the virtual machine.
But for the Docker VM, it depends on what is implemented and what is accessible ?

medyagh:
> > I would like to see that info for all of them. If not possible for a platform for now. We can just skip
> 
> Should open a new issue, for getting docker disk information on those platforms...
> I'm not sure if we have `ssh` access, or if we know how those disk images work ?
> 
> For the Minikube VM and SSH, it is easy enough to connect to the virtual machine.
> But for the Docker VM, it depends on what is implemented and what is accessible ?

sure we can do it in a separate PR

PR #7486, opened 2020-04-07
CI: change wait for  cni

medyagh:
minikube with CNI has always had this issue  https://github.com/kubernetes/minikube/issues/7354
and we still needs to fix it.
but the new --wait feature can not be used with this test

this should make the tests green again.


sharifelgamal:
The unit test failure seems to be the standard SSH flake, we should look into that again.

PR #7489, opened 2020-04-07
fix readme link to minikube examples

prasadkatti:

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ See the [Getting Started Guide](https://minikube.sigs.k8s.io/docs/start/)
 
 See https://minikube.sigs.k8s.io/docs/
 
-## Minikube in action
+## More Examples
 
 See minikube in action [here](https://minikube.sigs.k8s.io/docs/handbook/controls/)
 
--- a/README.md
+++ b/README.md
@@ -49,9 +49,9 @@ See the [Getting Started Guide](https://minikube.sigs.k8s.io/docs/start/)
 
 See https://minikube.sigs.k8s.io/docs/
 
-## More Examples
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -49,9 +49,9 @@ See the [Getting Started Guide](https://minikube.sigs.k8s.io/docs/start/)
 
 See https://minikube.sigs.k8s.io/docs/
 
-## More Examples
+## Minikube in action
```
medyagh, line 52: minikube should always be lowercase ! we are mini :)

medyagh: lets keep it as More Examples , I like that one better

README.md
```diff
@@ -49,9 +49,9 @@ See the [Getting Started Guide](https://minikube.sigs.k8s.io/docs/start/)
...
 
-See our [examples page](https://minikube.sigs.k8s.io/docs/examples/)
+See minikube in action [here](https://minikube.sigs.k8s.io/docs/handbook/controls/)
```
medyagh, line 54: lets add an alias 
so anyone linked to this would be reffered to old location

medyagh: here is example of alias https://github.com/kubernetes/minikube/pull/7485

prasadkatti: do you mean people with the old link would be referred to the new page?

prasadkatti: @medyagh, I have added the alias and confirmed that it works using the netlify link.

prasadkatti:
@medyagh - This link (https://minikube.sigs.k8s.io/docs/reference/networking/) is broken but I can't find that page on the new site.

PR #7490, opened 2020-04-07
Extract preloaded tarball as soon as minikube volume is created (2.5s speedup)

priyawadhwa:
Extracting preload currently takes 6 seconds. This PR begins preload extraction as soon as the minikube volume is created, and runs it in parallel with creating the container node and setting up SSH.

This shaves 2-3 seconds off of `node.StartHost`


pkg/drivers/kic/kic.go
```diff
@@ -112,6 +113,28 @@ func (d *Driver) Create() error {
...
+		if !download.PreloadExists(d.NodeConfig.KubernetesVersion, d.NodeConfig.ContainerRuntime) {
+			return
+		}
+		t := time.Now()
+		glog.Infof("Starting extracting preloaded images to volume")
+		// Extract preloaded images to container
+		if err := oci.ExtractTarballToVolume(download.TarballPath(d.NodeConfig.KubernetesVersion, d.NodeConfig.ContainerRuntime), params.Name, BaseImage); err != nil {
+			glog.Infof("Unable to extract preloaded tarball to volume: %v", err)
+		} else {
+			glog.Infof("Took %f seconds to extract preloaded images to volume", time.Since(t).Seconds())
```
medyagh, line 134: anywhere else in minikube we have this format
"duration metric: took % to..." please use same format, for better search in logs expereince.

pkg/drivers/kic/oci/oci.go
```diff
@@ -84,6 +84,18 @@ func DeleteContainer(ociBin string, name string) error {
 	return nil
 }
 
+// SetupContainerNode sets up the container node
+func SetupContainerNode(p CreateParams) error {
```
medyagh, line 88: this doesnt seem to be creating a container, did you mean crate Volume?

pkg/drivers/kic/oci/oci.go
```diff
@@ -84,6 +84,18 @@ func DeleteContainer(ociBin string, name string) error {
...
+	if p.OCIBinary != Docker {
```
medyagh, line 89: I believe podman preload works as well. ( podman also has docker runtime) so no need to exclude podman.

pkg/drivers/kic/oci/oci.go
priyawadhwa: so creating this volume currently happens in `CreateContainerNode`, but I wanted it to happen ASAP so I moved it into this function `SetupContainerNode` to make it clear that it needs to run before `CreateContainerNode`

pkg/drivers/kic/oci/oci.go
priyawadhwa: This volume is currently only created for docker:

https://github.com/kubernetes/minikube/blob/d84b9425c9a57f1e8fea7f951229da4c5d0303b4/pkg/drivers/kic/oci/oci.go#L124-L125

is this a bug?

pkg/drivers/kic/kic.go
priyawadhwa: done

pkg/drivers/kic/oci/oci.go
medyagh: I think so... there is no reason we dont create volume for podman.

medyagh: so I was wrong, and you were right. sorry the original code was:

```
	if p.OCIBinary == Podman { // enable execing in /var
		// volume path in minikube home folder to mount to /var
		hostVarVolPath := filepath.Join(localpath.MiniPath(), "machines", p.Name, "var")
		if err := os.MkdirAll(hostVarVolPath, 0711); err != nil {
			return errors.Wrapf(err, "create var dir %s", hostVarVolPath)
		}
		// podman mounts var/lib with no-exec by default  https://github.com/containers/libpod/issues/5103
		runArgs = append(runArgs, "--volume", fmt.Sprintf("%s:/var:exec", hostVarVolPath))
	}
	if p.OCIBinary == Docker {
		if err := createDockerVolume(p.Name, p.Name); err != nil {
			return errors.Wrapf(err, "creating volume for %s container", p.Name)
		}
		glog.Infof("Successfully created a docker volume %s", p.Name)
		runArgs = append(runArgs, "--volume", fmt.Sprintf("%s:/var", p.Name))
		// setting resource limit in privileged mode is only supported by docker
		// podman error: "Error: invalid configuration, cannot set resources with rootless containers not using cgroups v2 unified mode"
		runArgs = append(runArgs, fmt.Sprintf("--cpus=%s", p.CPUs), fmt.Sprintf("--memory=%s", p.Memory))
	}
```

pkg/drivers/kic/oci/oci.go
medyagh: how about adding a comment to explain what this does, so future contributors know what and why.

and also maybe a different name, how about
prepareContainer
prepareNodeVolume

priyawadhwa: done

medyagh:
## TestOffline/group/crio - 566.34s
## TestStartStop/group/old-docker - 667.91s
related to https://github.com/kubernetes/minikube/issues/6956
## TestStartStop/group/crio - 614.77s
## TestStartStop/group/embed-certs - 299.

pkg/drivers/kic/oci/oci.go
```diff
@@ -84,6 +84,19 @@ func DeleteContainer(ociBin string, name string) error {
 	return nil
 }
 
+// PrepareContainerNode sets up the container node befpre CreateContainerNode is caleld
```
tstromberg, line 87: Check spellings

PR #7494, opened 2020-04-07
Use SSH runner once SSH is set up in kic

priyawadhwa:
This change shaves off 1.1 seconds on setupKubeadm

Slowjam at HEAD: http://104.197.255.167:8000 
Slowjam with this PR: http://35.193.233.143:8000


medyagh:
Thanks for this ! I have been meaning to do this PR. ( if we can find the issue for this) we can close that issue with this PR.

Issue #7428, opened 2020-04-05
unify Default CNI for VM and containers


medyagh:
currently we have two CNIs

https://github.com/kubernetes/minikube/blob/a9558856c54cc0b3b6c46a7ef96960f521a72fb6/pkg/minikube/bootstrapper/kubeadm/default_cni.go

- defaultCNIConfig
- kicCNIConfig

kicCNIConfig is only used by containerd and crio only on docker driver.
I have a feeling by tweaking the defaultCNIConfig we could get rid of the kicCNIConfig.

the defaultCNIConfig has   "subnet": "10.1.0.0/16",
```
// The config is being written to /etc/cni/net.d/k8s.conf.
const defaultCNIConfig = `
{
  "cniVersion": "0.3.0",
  "name": "rkt.kubernetes.io",
  "type": "bridge",
  "bridge": "mybridge",
  "mtu": 1460,
  "addIf": "true",
  "isGateway": true,
  "ipMasq": true,
  "ipam": {
    "type": "host-local",
    "subnet": "10.1.0.0/16",
    "gateway": "10.1.0.1",
    "routes": [
      {
        "dst": "0.0.0.0/0"
      }
    ]
  }
}
`

```

but for kic drivers we add 10.244.0.0/16 
```
🐳  Preparing Kubernetes v1.18.0 on Docker 19.03.2 ...
    ▪ kubeadm.pod-network-cidr=10.244.0.0/16
```

## The Error if we delete kic overlay for cotainerd on docker driver.

```
medmac@~/workspace/minikube (remove_overlay) $ kc get pods -A
NAMESPACE     NAME                               READY   STATUS              RESTARTS   AGE
kube-system   coredns-66bff467f8-l7qkl           0/1     ContainerCreating   0          96s
kube-system   coredns-66bff467f8-vf6bm           0/1     ContainerCreating   0          96s
kube-system   etcd-minikube                      1/1     Running             0          112s
kube-system   kube-apiserver-minikube            1/1     Running             0          112s
kube-system   kube-controller-manager-minikube   1/1     Running             0          112s
kube-system   kube-proxy-t424s                   1/1     Running             0          96s
kube-system   kube-scheduler-minikube            1/1     Running             0          112s
kube-system   storage-provisioner                1/1     Running             0          110s
```
and here is the descripbe  which says:
'failed to set bridge addr: could not add IP address to "cni0": permission denied'

```
[...]

Issue #7505, opened 2020-04-08
Docker Driver Stop : - Error: failed to create listener: failed to listen on 0.0.0.0:8444


medyagh:
as seen in https://storage.googleapis.com/minikube-builds/logs/7504/e0611b6/Docker_Linux.html#fail_TestStartStop%2fgroup%2fcontainerd

```

- Error: failed to create listener: failed to listen on 0.0.0.0:8444: listen tcp 0.0.0.0:8444: bind: address already in use

```

medyagh:
I belive this was fixed by my shut down gracefully PR

PR #7509, opened 2020-04-08
Respect driver.FlagDefaults even if --extra-config is set

medyagh:
this one line change is result of a full day of deep investigation and yes it fixes it !
here is what was happpenning .

## Before this PR:

```
minikube start -p p2 --memory=2200 --alsologtostderr -v=3 --wait=true --container-runtime=containerd --driver=docker  --kubernetes-version=v1.15.7 --extra-config=kubeadm.ignore-preflight-errors=SystemVerification
```

was failiing on waiting because coreDNS container was stuck in Creating:

```
kube-system   coredns-5d4dd4b4db-qhpdp   0/1     ContainerCreating
```
because

```
Failed create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "bf4e32817a8917e27d39b549b18f426b7db09cbd2204d0429aded681db77d2c4": failed to set bridge addr: could not add IP address to "cni0": permission denied
```

since  we added a new --wait component "aps_running" the 'TestStartStop' for 'Crio' and 'Containerd' revealed the hidden failure that we had not seen but always existed (on our previous stable releases it exists)

infact users had reported this issue  https://github.com/kubernetes/minikube/issues/7354 but it was hidden to our eyes before the new --wait flag.

### What was happening ?

for crio and contanerd on docker driver we need a CNI to make it work.
and we automatically set the Extra Options for the Pod CIDR to be routed to use the kic CNI (overlay).

however ---> if someone adds an Extra Option through the flag , we that messed up the  auto setting the CNI Extra option!

(thank you so much anyone who wrote that Integration Tests to start with extra options, you did a great job)

This one line change was found thanks to the new --wait=true that waits for more stuff and revealed all the secret bugs we had covering under a lot of dust.

(also tweaking the logs and the retry logic to be more efficient and produce less noise)

Cheers !
 
## After this PR : 

Works like a Charm !

### this PR fixes this issue for Docker and Podman Drivers.

I still hope one day we unify the CNI solution across VM and container and Multinode I hope we do this in the future : https://github.com/kubernetes/minikube/issues/7428

## not fixed in this PR
this PR still does not fix the CRIO problem with CNI on docker driver. that would be subject for another investigation. see more here https://github.com/kubernetes/minikube/issues/7522


pkg/minikube/bootstrapper/bsutil/kverify/system_pods.go
```diff
@@ -113,9 +110,20 @@ func ExpectAppsRunning(cs *kubernetes.Clientset, expected []string) error {
...
 	start := time.Now()
-	checkRunning := func() error { return ExpectAppsRunning(cs, expected) }
-	if err := retry.Expo(checkRunning, 500*time.Millisecond, timeout); err != nil {
-		return errors.Wrap(err, "waitings for k8s app running")
+
+	checkRunning := func() (bool, error) {
+		if err := ExpectAppsRunning(cs, expected); err != nil {
+			if time.Since(start) > minLogCheckTime {
+				glog.Infof("error waiting for apps to be running: %v", err)
+				time.Sleep(kconst.APICallRetryInterval * 5)
```
tstromberg, line 118: Why does this multiply the standard retry interval by 5?

tstromberg:
Can you give this PR a title that reflects the change? 

It doesn't seem to be specific to containerd.

pkg/minikube/bootstrapper/bsutil/kverify/system_pods.go
medyagh: that was something I stole from the simmilar func system pods

tstromberg: They do it to avoid pegging the system because they are running:

`announceProblems(r, bs, cfg, cr)`

You may want to call the same function here.

medyagh: we dont have the parameters that announceProblems needs in this function. (r cruntime.Manager, bs bootstrapper.Bootstrapper,)

medyagh:
The TestContainerd is related to BindAddress  https://github.com/kubernetes/minikube/issues/7505
https://github.com/kubernetes/minikube/issues/7521

medyagh:
TestStartStop/group/old-docker is related to Cert Flake https://github.com/kubernetes/minikube/issues/6956

PR #7525, opened 2020-04-08
start: Parallelize mount, cache push, and addon enablement

tstromberg:
Pushes 3 steps to run in parallel:

- Mount setup
- Pushing cached images from config
- Addon enablement (each addon gets run in parallel)

Shaves ~1.3s off of `minikube start --driver=docker` on macOS (average across 5 runs)

To make this race-safe, it required a small refactor of the addons package, as `addons.Set()` is not thread-safe. On the plus side, we no longer load & store the profile repeatedly.

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -290,9 +290,7 @@ func enableOrDisableStorageClasses(cc *config.ClusterConfig, name string, val st
 // Start enables the default addons for a profile, plus any additional
 func Start(wg sync.WaitGroup, profile string, toEnable map[string]bool, additional []string) {
 	wg.Add(1)
-	defer func() {
-		wg.Done()
-	}()
+	defer wg.Done()
 
 	start := time.Now()
 	glog.Infof("enableAddons start: toEnable=%v, additional=%s", toEnable, additional)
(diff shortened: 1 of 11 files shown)


pkg/addons/addons.go
```diff
@@ -287,7 +288,12 @@ func enableOrDisableStorageClasses(cc *config.ClusterConfig, name string, val st
 }
 
 // Start enables the default addons for a profile, plus any additional
-func Start(profile string, toEnable map[string]bool, additional []string) {
+func Start(wg sync.WaitGroup, profile string, toEnable map[string]bool, additional []string) {
+	wg.Add(1)
+	defer func() {
```
priyawadhwa, line 293: nit:  `defer wg.Done()` is cleaner

pkg/addons/addons.go
```diff
@@ -323,10 +329,14 @@ func Start(profile string, toEnable map[string]bool, additional []string) {
 
 	out.T(out.AddonEnable, "Enabling addons: {{.addons}}", out.V{"addons": strings.Join(toEnableList, ", ")})
 	for _, a := range toEnableList {
-		err := Set(a, "true", profile)
-		if err != nil {
-			// Intentionally non-fatal
-			out.WarningT("Enabling '{{.name}}' returned an error: {{.error}}", out.V{"name": a, "error": err})
-		}
+		go func() {
```
priyawadhwa, line 332: I think go routines in for loops usually require something like:

```go
go func(a string) {
  /// code
}(a)
```
since the value of a will change

[ref](https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables)

pkg/minikube/node/config.go
```diff
@@ -46,7 +47,12 @@ func showVersionInfo(k8sVersion string, cr cruntime.Manager) {
 }
 
 // configureMounts configures any requested filesystem mounts
-func configureMounts() {
+func configureMounts(wg sync.WaitGroup) {
+	wg.Add(1)
+	defer func() {
```
priyawadhwa, line 52: same nit as above

PR #7498, opened 2020-04-08
improve driver failure message with link to driver docs

medyagh:
Closes https://github.com/kubernetes/minikube/issues/7457

When Docker is not running:

## Before this PR:
- take long 24s!
- Crash ugly !

```
😄  minikube v1.9.2 on Darwin 10.13.6
✨  Using the docker driver based on existing profile

❗  'docker' driver reported an issue: exit status 1
💡  Suggestion: Docker is not running or is responding too slow. Try: restarting docker desktop.

👍  Starting control plane node m01 in cluster minikube
🚜  Pulling base image ...
E0407 19:21:47.210473   41415 cache.go:114] Error downloading kic artifacts:  error loading image: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
🤷  docker "minikube" container is missing, will recreate.
🔥  Creating Kubernetes in docker container with (CPUs=2) (0 available), Memory=4000MB (0MB available) ...
🤦  StartHost failed, but will try again: recreate: creating host: create: creating: create kic node: creating volume for minikube container: output Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
: exit status 1
🤷  docker "minikube" container is missing, will recreate.
🔥  Creating Kubernetes in docker container with (CPUs=2) (0 available), Memory=4000MB (0MB available) ...

💣  Failed to start docker container. "minikube start" may fix it.: recreate: creating host: create: creating: create kic node: creating volume for minikube container: output Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
: exit status 1

😿  minikube is exiting due to an error. If the above message is not useful, open an issue:
👉  https://github.com/kubernetes/minikube/issues/new/choose

real	0m24.654s
user	0m1.602s
sys	0m0.787s

```

## After This PR:
- Fail Pretty an in 0.18 Seconds
```
😄  minikube v1.9.2 on Darwin 10.15.3
✨  Using the docker driver based on user configuration

❗  'docker' driver reported an issue: exit status 1
💡  Suggestion: Start the Docker Service.
📘  Documentation: https://minikube.sigs.k8s.io/docs/drivers/docker

💣  Failed to validate 'docker' driver

real    0m0.260s
user    0m0.136s
sys     0m0.129s
```

PR #7531, opened 2020-04-08
Double docker driver timeout, use version instead of info

tstromberg:
On Google CloudShell, Docker auto-detection would sporadically fail because the timeout was too short, so I doubled it. I also changed the the error detection to the `version` command, which returns faster. 

Since this error should be fairly common, I included the arguments in the error message. Example output:

```
😄  minikube v1.9.2 on Darwin 10.15.4

❗  'docker' driver reported an issue: "docker version --format {{.Server.Version}}" exit status 1: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
💡  Suggestion: Start the Docker service
📘  Documentation: https://minikube.sigs.k8s.io/docs/drivers/docker/
```


medyagh:
do u mind pulling the changes from this PR ? https://github.com/kubernetes/minikube/pull/7498
or merge that first ?

Issue #7541, opened 2020-04-09
TestStartStop: "Failed to setup certs" flake


medyagh:
as seen here 
https://storage.googleapis.com/minikube-builds/logs/7435/70adc95/Docker_Linux.html#fail_TestStartStop%2fgroup%2fold-docker

```
	W0408 20:09:24.196997    4909 exit.go:101] Failed to setup certs: certificate symlinks: create symlink for /usr/share/ca-certificates/4970.pem: sudo /bin/bash -c "test -s /usr/share/ca-certificates/4970.pem && ln -fs /usr/share/ca-certificates/4970.pem /etc/ssl/certs/4970.pem": exit status 1
	stdout:
	
	stderr:
	* 
	X Failed to setup certs: certificate symlinks: create symlink for /usr/share/ca-certificates/4970.pem: sudo /bin/bash -c "test -s /usr/share/ca-certificates/4970.pem && ln -fs /usr/share/ca-certificates/4970.pem /etc/ssl/certs/4970.pem": exit status 1
	stdout:
	
	stderr:
	
	* 
	* minikube is exiting due to an error. If the above message is not useful, open an issue:
```

medyagh:
CC: @tstromberg

tstromberg:
```
	I0408 20:09:15.512190    4909 certs.go:342] found cert: /home/jenkins/minikube-integration/linux-amd64-docker-7435-2451-70adc95be27abfd1f53736cd23f13206bf501118/.minikube/certs/home/jenkins/minikube-integration/linux-amd64-docker-7435-2451-70adc95be27abfd1f53736cd23f13206bf501118/.minikube/certs/4970.pem (1338 bytes)
	W0408 20:09:15.512248    4909 certs.go:338] ignoring /home/jenkins/minikube-integration/linux-amd64-docker-7435-2451-70adc95be27abfd1f53736cd23f13206bf501118/.minikube/certs/home/jenkins/minikube-integration/linux-amd64-docker-7435-2451-70adc95be27abfd1f53736cd23f13206bf501118/.minikube/certs/4970_empty.pem, impossibly tiny 0 bytes
...
	I0408 20:09:22.527884    4909 certs.go:120] copying: /usr/share/ca-certificates/4970.pem
	I0408 20:09:23.815705    4909 kic_runner.go:91] Run: openssl version
[...]

PR #7559, opened 2020-04-09
make CommandRunner.Copy/Remove consistent across runners

tstromberg:
While investigating #7541 I found some inconsistency between how the 3 runners copied files, so I did some cleanup to help me find the underlying issue, if not fix it entirely:

- `Copy` now consistently logs and warn about 0-byte assets across all runners
- `kiRunner.Copy` now checks for pre-existing files (only over a certain size)
- `kicRunner.Copy` now buffer copies instead of reading the entire file into memory at once
- `Remove` now logs across all runners
-  Moved `teePrefix` from `ssh_runner.go` to `command_runner.go`
- For clarity, `asset.GetAssetName()` was renamed to `asset.GetSourcePath()`

I did some performance testing on macOS (cold and warm start, docker and hyperkit), and found the performance improvements negligible. The only scenario that showed a performance benefit  was docker warm start (9% faster).

### DIFF
--- a/pkg/minikube/assets/vm_assets.go
+++ b/pkg/minikube/assets/vm_assets.go
@@ -29,11 +29,15 @@ import (
 	"github.com/pkg/errors"
 )
 
+// MemorySource is the source name used for in-memory copies
+const MemorySource = "memory"
+
 // CopyableFile is something that can be copied
 type CopyableFile interface {
 	io.Reader
 	GetLength() int
-	GetAssetName() string
+	GetSourcePath() string
+
 	GetTargetDir() string
 	GetTargetName() string
 	GetPermissions() string
@@ -43,15 +47,16 @@ type CopyableFile interface {
 
 // BaseAsset is the base asset class
 type BaseAsset struct {
-	AssetName   string
+	SourcePath  string
 	TargetDir   string
 	TargetName  string
 	Permissions string
+	Source      string
 }
 
-// GetAssetName returns asset name
-func (b *BaseAsset) GetAssetName() string {
-	return b.AssetName
+// GetSourcePath returns asset name
+func (b *BaseAsset) GetSourcePath() string {
+	return b.SourcePath
 }
 
 // GetTargetDir returns target dir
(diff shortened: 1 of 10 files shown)


pkg/minikube/assets/vm_assets.go
```diff
@@ -234,9 +240,9 @@ func (m *BinAsset) loadData(isTemplate bool) error {
 
 	m.length = len(contents)
 	m.reader = bytes.NewReader(contents)
-	glog.V(1).Infof("Created asset %s with %d bytes", m.AssetName, m.length)
+	glog.V(1).Infof("Created asset %s with %d bytes", m.SourcePath, m.length)
```
medyagh, line 243: nit: %q

pkg/minikube/bootstrapper/certs.go
```diff
@@ -117,9 +117,8 @@ func SetupCerts(cmd command.Runner, k8s config.KubernetesConfig, n config.Node)
 	}
 
 	for _, f := range copyableFiles {
-		glog.Infof("copying: %s/%s", f.GetTargetDir(), f.GetTargetName())
 		if err := cmd.Copy(f); err != nil {
-			return nil, errors.Wrapf(err, "Copy %s", f.GetAssetName())
+			return nil, errors.Wrapf(err, "Copy %s", f.GetSourcePath())
```
medyagh, line 121: not related to this PR but something to think about, what if the number of files to copy is more than 5, then make a tar out of them them copy one tar, and then untar them. (crazy idea for future) ?

pkg/minikube/command/command_runner.go
```diff
@@ -84,3 +85,101 @@ func (rr RunResult) Output() string {
...
+	var line bytes.Buffer
+
+	for scanner.Scan() {
+		b := scanner.Bytes()
+		if _, err := w.Write(b); err != nil {
+			return err
+		}
+		if bytes.IndexAny(b, "\r\n") == 0 {
+			if line.Len() > 0 {
+				logger("%s%s", prefix, line.String())
```
medyagh, line 102: does this mean, for binary files or big files this will flood the logs?

pkg/minikube/assets/vm_assets.go
tstromberg: It's a filepath -- normally it isn't necessary to quote.

pkg/minikube/bootstrapper/certs.go
tstromberg: I've thought the same, but honestly, for the ssh runner, it wouldn't have much impact.

For kic_runner it would though.

pkg/minikube/command/command_runner.go
tstromberg: This function is only used for streaming SSH commands - and remains unchanged by this PR. The only thing I did was move which file it was located in.

Issue #7561, opened 2020-04-09
kvm2: minikube start fails, due to timeout before disk image is fully created


raphendyr:
**Steps to reproduce the issue:** 

1. Have a Linux host with HDD or other slow storage
2. Download a version of minikube in range v1.7.3 to v1.9.2
3. Create a new minikube `minikube-1.7.3 start -p minikube173 --auto-update-drivers=false -v=8 --alsologtostderr  --vm-driver=kvm2`
4. Minikube start fails after around 2 minutes, due to a hard coded timeout.

**Full output of failed command:** 

<details>
<summary>Output for v1.9.2 (fails)</summary>

```
I0410 00:52:01.014304 1546583 start.go:262] hostinfo: {"hostname":"HOSTNAME","uptime":651379,"bootTime":1585817741,"procs":266,"os":"linux","platform":"debian","platformFamily":"debian","platformVersion":"bullseye/sid","kernelVersion":"5.4.0-2-amd64","virtualizationSystem":"kvm","virtualizationRole":"host","hostid":"HOSTID"}
I0410 00:52:01.019119 1546583 start.go:272] virtualization: kvm host
😄  [minikube192] minikube v1.9.2 on Debian bullseye/sid
I0410 00:52:01.102648 1546583 driver.go:245] Setting default libvirt URI to qemu:///system
❗  Both driver=kvm2 and vm-driver=kvm2 have been set.

    Since vm-driver is deprecated, minikube will default to driver=kvm2.

    If vm-driver is set in the global config, please run "minikube config unset vm-driver" to resolve this warning.

✨  Using the kvm2 driver based on user configuration
I0410 00:52:01.270772 1546583 start.go:310] selected driver: kvm2
I0410 00:52:01.283545 1546583 start.go:656] validating driver "kvm2" against <nil>
I0410 00:52:01.283657 1546583 start.go:662] status for kvm2: {Installed:true Healthy:true Error:<nil> Fix: Doc:}
I0410 00:52:01.307835 1546583 install.go:49] acquiring lock: {Name:mk900956b073697a4aa6c80a27c6bb0742a99a53 Clock:{} Delay:500ms Timeout:10m0s Cancel:<nil>}
I0410 00:52:01.319666 1546583 install.go:115] Validating docker-machine-driver-kvm2, PATH=/home/USER/.minikube/bin:/home/USER/.local/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games
[...]

PR #7565, opened 2020-04-10
Increase host creation timeout

tstromberg:
To address #7561

### DIFF
--- a/pkg/minikube/machine/start.go
+++ b/pkg/minikube/machine/start.go
@@ -142,7 +142,7 @@ func createHost(api libmachine.API, cfg config.ClusterConfig, n config.Node) (*h
 	cstart := time.Now()
 	glog.Infof("libmachine.API.Create for %q (driver=%q)", cfg.Name, cfg.Driver)
 
-	if err := timedCreateHost(h, api, 5*time.Minute); err != nil {
+	if err := timedCreateHost(h, api, 4*time.Minute); err != nil {
 		return nil, errors.Wrap(err, "creating host")
 	}
 	glog.Infof("libmachine.API.Create for %q took %s", cfg.Name, time.Since(cstart))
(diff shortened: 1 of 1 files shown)


tstromberg:
Do you mind checking if this binary fixes your issue:

https://storage.googleapis.com/minikube-builds/7565/minikube-linux-amd64

pkg/minikube/machine/start.go
```diff
@@ -141,8 +141,8 @@ func createHost(api libmachine.API, cfg config.ClusterConfig, n config.Node) (*h
 
 	cstart := time.Now()
 	glog.Infof("libmachine.API.Create for %q (driver=%q)", cfg.Name, cfg.Driver)
-	// Allow two minutes to create host before failing fast
-	if err := timedCreateHost(h, api, 2*time.Minute); err != nil {
+
+	if err := timedCreateHost(h, api, 4*time.Minute); err != nil {
```
medyagh, line 145: one of the reasons that this was made lower was to catch 
connecting to SSH...

that meant the SSH keys were not copied to the machine, and no matter how long u try it would not be helpful

how about make this timedCreateHost smarter and fail fast on the things that we can do NOTHING about it
(like SSH problems or network or something like that ?)

medyagh: here is the exact error message if you wanna catch that one to fail fast:
```
I0212 20:18:31.490329   17811 main.go:110] libmachine: Error dialing TCP: dial tcp 192.168.64.44:22: connect: connection refused
I0212 20:18:35.497395   17811 main.go:110] libmachine: Error dialing TCP: dial tcp 192.168.64.44:22: connect: connection refused
I0212 20:18:39.502636   17811 main.go:110] libmachine: Error dialing TCP: dial tcp 192.168.64.44:22: connect: connection refused
I0212 20:18:43.512406   17811 main.go:110] libmachine: Error dialing TCP: dial tcp 192.168.64.44:22: connect: connection refused
```

(maybe this might need a PR on libmachine (not sure))

tstromberg: I don't think it's feasible to detect other conditions. Is 3 minutes an acceptable compromise?

Issue #7521, opened 2020-04-08
Docker_Linux: TestStartStop/group/containerd: listen tcp 0.0.0.0:8444: bind: address already in use


tstromberg:
https://storage.googleapis.com/minikube-builds/logs/7516/5110da9/Docker_Linux.html#fail_TestStartStop%2fgroup%2fcontainerd

```
helpers.go:279: TestStartStop/group/containerd: WARNING: pod list for "kubernetes-dashboard" "k8s-app=kubernetes-dashboard" returned: Get https://172.17.0.2:8444/api/v1/namespaces/kubernetes-dashboard/pods?labelSelector=k8s-app%3Dkubernetes-dashboard: dial tcp 172.17.0.2:8444: connect: connection refused

start_stop_delete_test.go:141: ***** TestStartStop/group/containerd: pod "k8s-app=kubernetes-dashboard" failed to start within 4m0s: timed out waiting for the condition ****

start_stop_delete_test.go:141: TestStartStop/group/containerd: showing logs for failed pods as of 2020-04-08 09:55:54.348541364 -0700 PDT m=+1755.844986383

start_stop_delete_test.go:142: failed waiting for 'addon dashboard' pod post-stop-start: k8s-app=kubernetes-dashboard within 4m0s: timed out waiting for the condition
```

stdout shows some initial trouble with restarting:

```
	  - Apr 08 16:52:55 containerd-20200408T094544.677419249-5291 kubelet[949]: E0408 16:52:55.918044     949 pod_workers.go:191] Error syncing pod 39a425d2906ed7aa1f9aa5ea712491eb ("kube-apiserver-containerd-20200408t094544.677419249-5291_kube-system(39a425d2906ed7aa1f9aa5ea712491eb)"), skipping: failed to "StartContainer" for "kube-apiserver" with CrashLoopBackOff: "back-off 40s restarting failed container=kube-apiserver pod=kube-apiserver-containerd-20200408t094544.677419249-5291_kube-system(39a425d2906ed7aa1f9aa5ea712491eb)"

[...]

PR #7570, opened 2020-04-10
improve stop for docker/podman drivers

medyagh:
Redoing this PR: https://github.com/kubernetes/minikube/pull/7545

hopefully closes https://github.com/kubernetes/minikube/issues/7521
which after miniube was stopped for containerd, on the start you would sometimes see this in the logs:

```
	  - Error: failed to create listener: failed to listen on 0.0.0.0:8444: listen tcp 0.0.0.0:8444: bind: address already in use
```

the problem was in kic we were trying to stop containers before kic stop to avoid
bind address problem

but the command stop container was wrong in the lib, it was rm
and even if the lib writer meant remove, u can not remove a not stopped container. so I changed to stop, we _might_ need to do both stop and rm but I dont think rm is needed.

this doesnt work in crictl:

```
docker@minikube:~$ sudo crictl rm 5c3039d13ecadafdd46c1cabe936f4cc48114f1596c60c86e2ca7d33b7358be5
ERRO[0000] container "5c3039d13ecadafdd46c1cabe936f4cc48114f1596c60c86e2ca7d33b7358be5" is running, please stop it first 
FATA[0000] unable to remove container(s)

```

but this works:

```

docker@minikube:~$ sudo crictl stop 5c3039d13ecadafdd46c1cabe936f4cc48114f1596c60c86e2ca7d33b7358be5
5c3039d13ecadafdd46c1cabe936f4cc48114f1596c60c86e2ca7d33b7358be5
```

### DIFF
--- a/pkg/drivers/common.go
+++ b/pkg/drivers/common.go
@@ -20,9 +20,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"os/exec"
 	"path/filepath"
-	"strconv"
 	"syscall"
 
 	"github.com/docker/machine/libmachine/drivers"
@@ -31,7 +29,6 @@ import (
 	"github.com/docker/machine/libmachine/ssh"
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
-	"k8s.io/minikube/pkg/minikube/command"
 )
 
 // This file is for common code shared among internal machine drivers
@@ -120,21 +117,6 @@ func MakeDiskImage(d *drivers.BaseDriver, boot2dockerURL string, diskSize int) e
 		machPath := d.ResolveStorePath(".")
 		if err := fixMachinePermissions(machPath); err != nil {
 			return errors.Wrapf(err, "fixing permissions on %s", machPath)
-		}
-	}
-	return nil
-}
-
-// KillAPIServerProc will kill an api server proc if it exists
-func KillAPIServerProc(runner command.Runner) error {
-	// first check if it exists
-	rr, err := runner.RunCmd(exec.Command("pgrep", "kube-apiserver"))
-	if err == nil { // this means we might have a running kube-apiserver
-		pid, err := strconv.Atoi(rr.Stdout.String())
-		if err == nil { // this means we have a valid pid
-			if _, err = runner.RunCmd(exec.Command("kill", "-9", string(pid))); err != nil {
-				return errors.Wrap(err, "kill")
-			}
 		}
 	}
 	return nil
--- a/pkg/drivers/common.go
+++ b/pkg/drivers/common.go
@@ -20,7 +20,9 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"strconv"
 	"syscall"
 
 	"github.com/docker/machine/libmachine/drivers"
@@ -29,6 +31,7 @@ import (
 	"github.com/docker/machine/libmachine/ssh"
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
+	"k8s.io/minikube/pkg/minikube/command"
 )
 
 // This file is for common code shared among internal machine drivers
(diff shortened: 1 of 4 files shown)


pkg/drivers/common.go
```diff
@@ -122,6 +125,21 @@ func MakeDiskImage(d *drivers.BaseDriver, boot2dockerURL string, diskSize int) e
...
 }
 
+// KillAPIServerProc will kill an api server proc if it exists
+func KillAPIServerProc(runner command.Runner) error {
+	// first check if it exists
+	rr, err := runner.RunCmd(exec.Command("pgrep", "kube-apiserver"))
+	if err == nil { // this means we might have a running kube-apiserver
+		pid, err := strconv.Atoi(rr.Stdout.String())
+		if err == nil { // this means we have a valid pid
+			if _, err = runner.RunCmd(exec.Command("kill", "-9", string(pid))); err != nil {
```
tstromberg, line 135: Can you clarify how it is possible for an apiserver pid to exist if the containers are stopped, or at least the scenario this function exists for?

I would rather not take this shot-gun approach, as it may kill unintended non-minikube processes.

(also, you could do pkill -9, but that's beside the point)

pkg/drivers/common.go
```diff
@@ -122,6 +125,21 @@ func MakeDiskImage(d *drivers.BaseDriver, boot2dockerURL string, diskSize int) e
 	return nil
 }
 
+// KillAPIServerProc will kill an api server proc if it exists
+func KillAPIServerProc(runner command.Runner) error {
+	// first check if it exists
+	rr, err := runner.RunCmd(exec.Command("pgrep", "kube-apiserver"))
+	if err == nil { // this means we might have a running kube-apiserver
+		pid, err := strconv.Atoi(rr.Stdout.String())
```
tstromberg, line 133: IMHO, it should be considered exceptional and show an error if there is an apiserver running after the containers have been removed.

pkg/drivers/kic/kic.go
```diff
@@ -347,13 +347,20 @@ func (d *Driver) Stop() error {
 		}
 		if len(containers) > 0 {
 			if err := runtime.StopContainers(containers); err != nil {
-				glog.Errorf("unable to stop containers : %v", err)
+				glog.Infof("unable to stop containers : %v", err)
```
tstromberg, line 350: Maybe a Warningf?

pkg/drivers/common.go
medyagh: on containerd I killed the container and I saw there was something still listening the api port with lsof command. this command only runs if the proc exists, If you agree I could take it out for none and leave it only for kic.
(for none as there might be their own kube-apiserver that we dont wanna kill) 

what do u think ?

pkg/drivers/kic/kic.go
medyagh: I leave the warrning for the second one (KillContainers) so we dont scare the user too early.
if we can not kill the containers then will warn them.

medyagh:
very interesting error here : 
```
-- stdout --
	  - Error: failed to create listener: failed to listen on 0.0.0.0:8443: listen tcp 0.0.0.0:8443: bind: address already in use
-- /stdout --
** stderr ** 
	* Problems detected in kube-apiserver [cb0e9fcd0614]:
** /stderr **
helpers.go:215: ======> post-mortem[TestStartStop/group/embed-certs]: disk usage <======
helpers.go:216: (dbg) Run:  out/minikube-linux-amd64 -p embed-certs-20200409T222010.769942656-5674 ssh "df -h /var/lib/docker/overlay2 /var /; du -hs /var/lib/docker/overlay2"
helpers.go:216: (dbg) Non-zero exit: out/minikube-linux-amd64 -p embed-certs-20200409T222010.769942656-5674 ssh "df -h /var/lib/docker/overlay2 /var /; du -hs /var/lib/docker/overlay2": exit status 1 (339.258477ms)
-- stdout --
	Filesystem      Size  Used Avail Use% Mounted on
	/dev/sda1       493G   45G  428G  10% /var
	/dev/sda1       493G   45G  428G  10% /var
	overlay         493G   45G  428G  10% /
	12K	/var/lib/docker/overlay2
-- /stdout --
** stderr ** 
	du: cannot read directory '/var/lib/docker/overlay2': Permission denied
	ssh: exit status 1
** /stderr **
```

PR #7578, opened 2020-04-10
Try Make Loading Profile Backward Compatible

### LINKED ISSUE #7062: Support backward compatibility for old profiles
I have installed the latest minikube version v1.8.1 (I can't remember the previous version), but then all my clusters wouldn't work because their profiles became invalid.

I checked the code that reads the configuration from the profile and obviously the structure of the profile has changed. 

I suggest to keep the profiles backward compatible as far back as possible, and display warnings about attributes that will be changed in the profile rather than prompting the users to delete the profile. 

If you feel that's the right way I can start with putting back some of the renamed attributes?

`minikube profile list`

⚠️  Found 4 invalid profile(s) !
	 keda-external-scaler
	 minikube
	 twitch-keda
	 twitter-scaler
	 $ minikube delete -p keda-external-scaler
	 $ minikube delete -p minikube
	 $ minikube delete -p twitch-keda
	 $ minikube delete -p twitter-scaler

</details>

MacOS Catalina

eashi:
Fixes #7062 

Note: _This might be a little too big PR, but it doesn't have to be; we can trim it down. I am happy to move it to "enhancements" if you like._

Without this PR, when developers introduce changes to profile configurations it may break old profiles. Minikube would try to load the profile and fails, then it prompts the user to delete those profiles than suggest to fix them.
I wrote about this in my blog [here](https://www.emadashi.com/2020/03/how-to-fix-minikube-invalid-profile)

With this PR, the code will try to load the profile, if it fails then it will try to load the profiles using the previous versions of the profile configuration using the old code, and then tries to translate it to the new configuration.

E.g. Imagine that the user has old profiles on version v1.5.2, and then she installed v1.9.2. The code will try to load the profiles using the v1.9.2 code, if it fails it will try using the code v1.8.2 if it fails it will try using code v1.7.2. If it succeeds it will load the profile and then tries to translate it to v1.8.2, if it succeeds then it will translate it to v1.9.2 and bubbles up to the current version. 

This is done recursively through the method `tryTranslate` in the `translators.go`. The method will recurse on the array `versionConfigTranslators` which will hold translators for older configurations.

I wanted to load the old profiles as close as they used to be loaded, so we will create a folder for each version in which the profile config has changed, and put the old code (if possible) in there with the least amount of change. This, however, is taking the toll of this PR, I hope there is an easier way, or that we can be very selective in what we take from the old code.

The PR is far from being ready, but the main skeleton is there. Any help or suggestions is highly welcomed.


medyagh:
this is a very promising PR ! thank you for doing it.
cc @sharifelgamal  could provide better insight on this PR review

pkg/minikube/config/translators.go
```diff
@@ -0,0 +1,172 @@
...
+	IsValid                isValid
+}
+type translateToNextVersion func(interface{}) (interface{}, error)
+type tryLoadFromFile func() (interface{}, error)
+type isValid func(interface{}) bool
+
+func translateFrom163ToCurrent(oldConfig v162.MachineConfig) (*ClusterConfig, error) {
+	hypervUseExternalSwitch := (oldConfig.HypervVirtualSwitch != "")
+
+	return &ClusterConfig{
```
priyawadhwa, line 49: Since the majority of the fields here are the same, instead of specifying each one individually you can:

1. Unmarshal the value of `oldConfig` to []byte
1. Marshal those contents into `newConfig`
1. Only manually set the fields that have changed in `newConfig` (like hypervUseExternalSwitch)

You could do that in any place where you're converting from an old config to a new config.

medyagh:
@eashi are you sitll working on this?

eashi:
Hey @medyagh sorry I was side tracked by a presentation I have to do on this Thursday. I will jump back to this right after the presentation starting Friday.

If this is a priority I am happy to collaborate on this with anyone who wants to take over :). But hopefully I will be able to jump to this again this Friday.

eashi:
I signed it

eashi:
@medyagh @priyawadhwa This PR is very close to graduated to a non-draft PR. The only remaining thing I can see is a unit test for version 1.5.2 configuration. 

Please let me know if you have any comments, and let me know how many versions back you want this backward-compatibility work.

eashi:
Ok, I have added the unit tests for version 1.5.2. Things look ok from my side.

tstromberg:
We ended up fixing this issue in v1.12.x by fixing the individual changes. We now test upgrades from v1.0 and onward as part of our integration tests. This was a valiant PR, but I'm going to close this until we are able to identify the need for the additional code.

My apologies for letting this one sit for so long :(

eashi:
@tstromberg thanks for taking the time to close this one properly. Glad things are going the right direction regardless of solution, was fun trying a stab on it anyway.

Cheers :)

Issue #7380, opened 2020-04-02
docker driver + crio: failed to set bridge addr: could not add IP address to "cni0": permission denied


tstromberg:
TestStartStop/group/crio has been broken with the Docker driver since Mar 23 19:11  - particularly with this commit:

GitHub pull request #7182 of commit 39b08c98765027721484fc8eb70abc880b604a76, no merge conflicts.

Here is an example failure:

TestStartStop/group/crio - 782.71s

`Warning  FailedCreatePodSandBox  7m18s                 kubelet, crio-20200401t175453.296999333-32003  Failed create pod sandbox: rpc error: code = Unknown desc = failed to create pod network sandbox k8s_busybox_default_1aa75488-2899-4d7d-98cb-50dc9ff62636_0(b2c34f2a8500f6f48f8d267786cbf3a04834cc756f94052f391ef074f5aac7a4): failed to set bridge addr: could not add IP address to "cni0": permission denied`

@afbjorklund - any clue here what might be going on? Can you think of a way where a podman/libglib update could impact this?

afbjorklund:
@stromberg:
I don't think that glib has anything to do with it, but conmon has _lots_ of releases...

I'm always a little vary of these projects that release too often (or not at all, like runc)
They tend to have a lot of bugs, which causes a lot of brown paper bag regressions.

For the minikube ISO, we tried with a more conservative v2.0.6 instead of v2.0.14

----

But now that podman had a major version bump, I might have to look into it again...

https://github.com/kubernetes/minikube/pull/7182/commits/3c37556627b8932f76e3db46b6ebdae8dc0ed156 podman-1.8.2

afbjorklund:
It seems that we have configured conmon to log to --syslog, but "forgot" to install rsyslogd...

afbjorklund:
Added some debug output to crio as well, but after restarting it - everything seems to work ?

afbjorklund:
Seems that it will *not* recover by itself:

``` 
Events:
  Type     Reason                  Age                     From               Message
  ----     ------                  ----                    ----               -------
[...]

PR #7580, opened 2020-04-10
update kicbase image : bump cri-o version

medyagh:
- bump crio version to match VM version
- bump kind base image to latest default that has some networking updates in the entrypoint https://github.com/kubernetes-sigs/kind/commit/9222508298c50ce8c5ba1f364f37307e81ba915e#diff-424162915c2ab4ab7f1b3f6dff73946e
- bump kindnet

- helps with investigating the root cause of https://github.com/kubernetes/minikube/issues/7380
by making sure VM and container use the same version of CRI-O


medyagh:
# Two Faiures for docker
- crio 
-and embed-cert
https://storage.googleapis.com/minikube-builds/logs/7580/10f924d/Docker_Linux.html#fail_TestStartStop%2fgroup%2fembed-certs

Failed to setup certs`

PR #7587, opened 2020-04-10
WIP: Restrict podman/docker GitHub CI tests to 2 threads/functional tests

tstromberg:
The idea here is:

* Keep GitHub CI doing what it does best: fast tests with no approvals
* Maximize CI signal by restricting tests to a subset that should run without issue
* Restrict parallelism to keep load down
* Get test results back quickly

It's worth noting that the `none` driver tests are running just fine in GitHub CI.


medyagh:
infact on github actions the only test that fails is TestFunctional so I am not sure if this will make it green

PR #7595, opened 2020-04-10
Skip containerd shutdown if Docker is bound to it

tstromberg:
A simple approach to something Medya was discussing for #7435

In the KIC configuration, Docker is bound to containerd.

### DIFF
--- a/pkg/minikube/cruntime/cruntime.go
+++ b/pkg/minikube/cruntime/cruntime.go
@@ -166,6 +166,7 @@ func ContainerStatusCommand() string {
 
 // disableOthers disables all other runtimes except for me.
 func disableOthers(me Manager, cr CommandRunner) error {
+
 	// valid values returned by manager.Name()
 	runtimes := []string{"containerd", "crio", "docker"}
 	for _, name := range runtimes {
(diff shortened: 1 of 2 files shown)


pkg/minikube/cruntime/cruntime.go
```diff
@@ -178,13 +179,22 @@ func disableOthers(me Manager, cr CommandRunner) error {
 		if r.Name() == me.Name() {
 			continue
 		}
+
+		// Don't disable containerd if we are bound to it
+		if me.Name() == "Docker" && r.Name() == "containerd" && dockerBoundToContainerd(cr) {
```
medyagh, line 184: is "Docker" capitalized ?

tstromberg: Yes. It's the string we output to console :(

medyagh:
the soft start on this PR is 13 seconds ( on my PR it was 8 seconds)
https://github.com/kubernetes/minikube/pull/7435

could be I am using different laptop ( but maybe somethign is wrong?)
```
medya@~/workspace/minikube (tstromberg-containerd-bound) $ time ./out/minikube start --driver=docker
😄  minikube v1.9.2 on Darwin 10.15.3
🚜  Pulling base image ...

real	0m13.866s
user	0m2.883s
sys	0m1.686s
```

tstromberg:
@medyagh - This is not expected to have a major impact on soft start speed. You can however confirm using `--alsologtostderr` that it does not shutdown containerd.

Issue #7601, opened 2020-04-10
TestErrorSpam: Restarting the docker service may improve performance


medyagh:
as seen in 
https://storage.googleapis.com/minikube-builds/logs/7586/d512c84/Docker_Linux.html#fail_TestErrorSpam

```
=== RUN   TestErrorSpam
=== PAUSE TestErrorSpam
=== CONT  TestErrorSpam
--- FAIL: TestErrorSpam (168.88s)
error_spam_test.go:42: (dbg) Run:  out/minikube-linux-amd64 start -p nospam-20200410T153613.393267403-15978 -n=1 --memory=2250 --wait=false --driver=docker 
error_spam_test.go:42: (dbg) Done: out/minikube-linux-amd64 start -p nospam-20200410T153613.393267403-15978 -n=1 --memory=2250 --wait=false --driver=docker : (2m15.474942738s)
error_spam_test.go:57: unexpected stderr line: "* Restarting the docker service may improve performance."
error_spam_test.go:57: unexpected stderr line: "stdout:"
error_spam_test.go:57: unexpected stderr line: "serviceaccount/storage-provisioner unchanged"
error_spam_test.go:57: unexpected stderr line: "clusterrolebinding.rbac.authorization.k8s.io/storage-provisioner unchanged"
error_spam_test.go:57: unexpected stderr line: "stderr:"
error_spam_test.go:49: unexpected error log in stderr: "Error from server (ServerTimeout): error when creating \"/etc/kubernetes/addons/storage-provisioner.yaml\": No API token found for service account \"storage-provisioner\", retry after the token is automatically created and added to the service account"
error_spam_test.go:57: unexpected stderr line: "]"
error_spam_test.go:69: *** TestErrorSpam FAILED at 2020-04-10 15:38:28.868492686 -0700 PDT m=+725.487531520
helpers.go:198: (dbg) Run:  out/minikube-linux-amd64 status --format={{.Host}} -p nospam-20200410T153613.393267403-15978
helpers.go:198: (dbg) Done: out/minikube-linux-amd64 status --format={{.Host}} -p nospam-20200410T153613.393267403-15978: (1.579189399s)
```

tstromberg:
This is two separate errors:

`* Restarting the docker service may improve performance`

and:

[...]

PR #7602, opened 2020-04-11
change ErrT from warn if slow  

medyagh:
to fix https://github.com/kubernetes/minikube/issues/7601


sharifelgamal:
can you link the issue this fixes?

medyagh:
> can you link the issue this fixes?

fixed the link

medyagh:
> I don't think this PR will help. `TestErrorSpam` does have a whitelist however that this can be put into.
> 
> That run also shows an unrelated error:
> 
> `Error from server (ServerTimeout): error when creating \"/etc/kubernetes/addons/storage-provisioner.yaml\": No API token found for service account \"storage-provisioner\", retry after the token is automatically created and added to the service account`
> 
> I think we retry this, so we should probably just hide it if so.

correct this PR will  fix the Error line for docker slow but not for other errors.
I think the Tip to restart docker should not be an error, as it is just a friendly tip.

PR #7608, opened 2020-04-11
wip: Improve kic start/stop and handle apparmor sec opt

medyagh:
hopefully this will solve most of the github action test failures
https://github.com/kubernetes/minikube/issues/7624

## notable changes
1-  add ignore for apparmor  and other local kubernetes experineces in docker also ignore it. this will help running minikube docker driver in github actions.
2- ensure start and ensure stopped by waiting for container status. 
3- make github actions run only functional tests for docker
4- make UniqueProfileName Container Friendly (remove Dot in the name and make it shorter)
5- kill APIServerID and ETCD on stop
6- Ensure SSHD service is up before returning SSHrunner for Kic drivers

## observations,
while debugging, I found out github actions does run out disk space and gets the DiskPressure when we run the full tests in parallel.
once we have Auto- Image Prune on start we could retry to run all tests in github actions again.

### DIFF
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -222,12 +222,12 @@ func (d *Driver) GetSSHPort() (int, error) {
 	findForward := func() error {
 		p, perr = oci.ForwardedPort(d.OCIBinary, d.MachineName, constants.SSHPort)
 		if perr != nil {
-			glog.Errorf("error getting foraded port: %v", perr)
+			glog.Errorf("(medya dbg) error getting forwaded port: %v", perr)
 			// for debugging
 			cmd := exec.Command(d.NodeConfig.OCIBinary, "ps", "-a")
 			b, err := cmd.CombinedOutput()
+			glog.Errorf("(medya dbg) docker ps -a debug: err: %v output : %q", err, string(b))
 			if err != nil {
-				glog.Errorf("error running debugging command ps -a: %v output %q", err, string(b))
 				return err
 			}
 			return perr
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -212,7 +212,7 @@ func (d *Driver) GetSSHPort() (int, error) {
 		return err
 	}
 
-	if err := retry.Expo(waitRunning, 500*time.Microsecond, time.Second); err != nil {
+	if err := retry.Expo(waitRunning, 500*time.Microsecond, time.Minute); err != nil {
 		// let it try anyways one last time. maybe by now it got there.
 		glog.Errorf("container is not running, might not be able to get SSH port: %v", err)
 	}
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -34,6 +34,7 @@ import (
 	pkgdrivers "k8s.io/minikube/pkg/drivers"
 	"k8s.io/minikube/pkg/drivers/kic/oci"
 	"k8s.io/minikube/pkg/minikube/assets"
+	"k8s.io/minikube/pkg/minikube/bootstrapper/bsutil/kverify"
 	"k8s.io/minikube/pkg/minikube/command"
 	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/minikube/cruntime"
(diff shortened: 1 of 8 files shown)


medyagh:
@tstromberg Ready for review

pkg/drivers/kic/kic.go
```diff
@@ -196,9 +197,26 @@ func (d *Driver) GetSSHHostname() (string, error) {
...
+	var perr error
+	var p int
+	findForward := func() error {
+		p, perr = oci.ForwardedPort(d.OCIBinary, d.MachineName, constants.SSHPort)
+		if perr != nil {
+			glog.Errorf("(medya dbg) error getting forwaded port: %v", perr)
+			// for debugging
+			cmd := exec.Command(d.NodeConfig.OCIBinary, "ps", "-a")
+			b, err := cmd.CombinedOutput()
+			glog.Errorf("(medya dbg) docker ps -a debug: err: %v output : %q", err, string(b))
```
tstromberg: Remove debug statement?

pkg/drivers/kic/kic.go
```diff
@@ -196,9 +197,26 @@ func (d *Driver) GetSSHHostname() (string, error) {
 
 // GetSSHPort returns port for use with ssh
 func (d *Driver) GetSSHPort() (int, error) {
-	p, err := oci.ForwardedPort(d.OCIBinary, d.MachineName, constants.SSHPort)
-	if err != nil {
-		return p, errors.Wrap(err, "get ssh host-port")
+	var perr error
+	var p int
+	findForward := func() error {
+		p, perr = oci.ForwardedPort(d.OCIBinary, d.MachineName, constants.SSHPort)
```
tstromberg, line 203: Should this retry go into oci.ForwardedPort so that it can be useful to other use cases?

pkg/drivers/kic/kic.go
```diff
@@ -358,13 +364,34 @@ func (d *Driver) Stop() error {
 	}
 
 	if err := killAPIServerProc(d.exec); err != nil {
-		glog.Warningf("couldn't stop kube-apiserver proc: %v", err)
+		glog.Infof("kill kube-apiserver proc: %v", err)
+	}
+
+	if err := killETCDProc(d.exec); err != nil {
```
tstromberg, line 363: this feels out of place to me The kic driver should not have to know about etcd.

pkg/drivers/kic/oci/oci.go
```diff
@@ -108,7 +109,9 @@ func CreateContainerNode(p CreateParams) error {
 		// including some ones docker would otherwise do by default.
 		// for now this is what we want. in the future we may revisit this.
 		"--privileged",
-		"--security-opt", "seccomp=unconfined", // also ignore seccomp
+		"--security-opt", "seccomp=unconfined", //  ignore seccomp
+		// for github actions: https://github.com/kubernetes/minikube/issues/7624
+		"--security-opt", "apparmor=unconfined", // ignore apparmor
```
tstromberg, line 114: Any idea if other local Kubernetes solutions set these flags? They make me a bit concerned.

pkg/minikube/machine/client.go
```diff
@@ -152,8 +154,25 @@ func CommandRunner(h *host.Host) (command.Runner, error) {
 	if driver.BareMetal(h.Driver.DriverName()) {
 		return command.NewExecRunner(), nil
 	}
+	sshRunner := command.NewSSHRunner(h.Driver)
```
tstromberg, line 157: This needs to be moved to a separate PR. I'm not entirely sure why this is here at the moment, and would like to  hear more about it's impact to start latency.

pkg/drivers/kic/oci/oci.go
medyagh: yes they do

pkg/minikube/machine/client.go
medyagh: I dont return the kic runner, I only use it once for verifying if the SSH service is UP

pkg/drivers/kic/kic.go
medyagh: I think this retry was actually useless, I will delete.

pkg/drivers/kic/kic.go
medyagh: this would only be used by kic drivers, since our base image init script meses with the SIG STOP, other drivers won't need to use this. I will add a comment why we need this.

pkg/drivers/kic/kic.go
medyagh: removed the retry

medyagh:
I will break this PR to 5 PRs as suggested.

medyagh:
> This PR seems to impact a couple of different areas. Can you please split it up logically for:
> 
> * GitHub workflow changes
> * AppArmor changes
> * SSH changes
> * ForwardedPort changes
> * etcd changes
> * Unrelated integration test changes
> 
> Thank you!

done.

PR #7609, opened 2020-04-11
Add a note that minikube does not support IPv6 yet

### LINKED ISSUE #7491: Document that minikube does not support IPv6
It would be great if there were instructions for running minikube and kubernetes in IPv6 only mode. 
Now it could be that that's not even possible from my tests as with the virtualbox driver at least (which is what I tested), the master node has IPv6 disabled in the sysctls (net.ipv6.conf.all.disable_ipv6 is set to 1 by default) - which I wouldn't expect to be default kubernetes behaviour.
Also, minikube only has one form of the host-only-cidr config item, which is IPv4 only (always in vbox driver context), as it tries to pass the configured cidr to the "VBoxManage hostonlyif ipconfig" command with the "--ip" flag, which is IPv4 only. For Ipv6 Vbox requires the "--ipv6" flag:

#VBoxManage hostonlyif
Usage:

VBoxManage hostonlyif       ipconfig <name>
                            [--dhcp |
                            --ip<ipv4> [--netmask<ipv4> (def: 255.255.255.0)] |
                            --ipv6<ipv6> [--netmasklengthv6<length> (def: 64)]]
                            create |
                            remove <name>

I've stopped trying stuff out here, and maybe I'm doing it wrong, but IF IPv6 is not supported in minikube or minikube and Vbox as driver (or other drivers) that should be clearly stated in the handbook.

prasadkatti:
fixes #7491

### DIFF
--- a/site/content/en/docs/faq/_index.md
+++ b/site/content/en/docs/faq/_index.md
@@ -6,11 +6,11 @@ description: >
   Questions that come up regularly
 ---
 
-## Operating-systems
+## Does minikube support IPv6?
 
-## Linux
+minikube currently doesn't support IPv6. However, it is on the [roadmap]({{< ref "/docs/contrib/roadmap.en.md" >}}).
 
-### Preventing password prompts
+## How can I prevent password prompts on Linux?
 
 The easiest approach is to use the `docker` driver, as the backend service always runs as `root`.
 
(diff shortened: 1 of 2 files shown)


medyagh:
@prasadkatti a better place for this is FAQ page

and if you could also please re-do the FAQ page and convert them to Questions and Answers

prasadkatti:
@medyagh - done!

PR #7619, opened 2020-04-11
Add proxy cidr check in order not to access nil address

### LINKED ISSUE #7618: proxy.go isInBlock: invalid memory address or nil pointer dereference in case of multi node
```
$ out/minikube version
minikube version: v1.9.2
commit: 36f879947d28bed2354e7b625c46f940eed71854
```

When minikube start with multi node(n=2), 2nd node is set NO_PROXY.
```
👍  Starting node m02 in cluster multinode-demo
🌐  Found network options:
    ▪ NO_PROXY=192.168.64.213
```

When PROXY env(e.g. HTTP_PROXY, NO_PROXY) is set, minikube check if it's ip exists in NO_PROXY CIDR at `proxy.IsIPExcluded`.

https://github.com/kubernetes/minikube/blob/93af9c1e43cab9618e301bc9fa720c63d5efa393/pkg/minikube/node/start.go#L370

In `proxy.IsIPExcluded`, this func call `checkEnv` func.

https://github.com/kubernetes/minikube/blob/93af9c1e43cab9618e301bc9fa720c63d5efa393/pkg/minikube/proxy/proxy.go#L69-L71

And `checkEnv` func call `isInBlock` func in `proxy.go`.

https://github.com/kubernetes/minikube/blob/93af9c1e43cab9618e301bc9fa720c63d5efa393/pkg/minikube/proxy/proxy.go#L102-L107

`isInBlock` func checks if the target ip is included in target CIDR.
In `isInBlock`, `block` var is expected as IP CIDR(e.g. 192.168.64.213/32). 
However, the invalid CIDR(192.168.64.213)  is set to `block` var when multi node.
This is missing `/XX`.
 
```
    ▪ NO_PROXY=192.168.64.213
```

https://github.com/kubernetes/minikube/blob/93af9c1e43cab9618e301bc9fa720c63d5efa393/pkg/minikube/proxy/proxy.go#L47-L50

So, above `b` var comes to be nil. And now we don't have its check or validation.

For further the returned error is igonred.

```
 	if yes, _ := isInBlock(ip, b); yes { 
```

We should check if the block is CIDR format before calling `net.ParseCIDR(block)`.

govargo:
### What type of PR is this?
/kind bug
/kind cleanup

Now the proxy CIDR is not checked, when invalid CIDR is set(e.g. `10.0.0.0` missing `/XX`).
This PR add the check if the proxy CIDR is valid before `net.ParseCIDR`.
This prevents the minikube from accessing nil address(SIGSEV).
I described the detail of this in #7618.

### Which issue(s) this PR fixes:
Fixes #7618 

Yes. the broken url for proxy is fixed.

The CIDR check in proxy.go is performed inside the process. Users doesn't face.

### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:
```
NONE
```

### DIFF
--- a/pkg/minikube/proxy/proxy.go
+++ b/pkg/minikube/proxy/proxy.go
@@ -112,7 +112,7 @@ func checkEnv(ip string, env string) bool {
 	for _, b := range noProxyBlocks {
 		yes, err := isInBlock(ip, b)
 		if err != nil {
-			glog.Errorf("fail to check proxy env: %v", err)
+			glog.Warningf("fail to check proxy env: %v", err)
 		}
 		if yes {
 			return true
--- a/pkg/minikube/proxy/proxy.go
+++ b/pkg/minikube/proxy/proxy.go
@@ -44,6 +44,9 @@ func isInBlock(ip string, block string) (bool, error) {
 	if i == nil {
 		return false, fmt.Errorf("parsed IP is nil")
 	}
+	if !strings.Contains(block, "/") {
+		return false, nil
+	}
 	_, b, err := net.ParseCIDR(block)
 	if err != nil {
 		return false, errors.Wrapf(err, "Error Parsing block %s", b)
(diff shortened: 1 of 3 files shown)


pkg/minikube/proxy/proxy.go
```diff
@@ -44,6 +44,9 @@ func isInBlock(ip string, block string) (bool, error) {
 	if i == nil {
 		return false, fmt.Errorf("parsed IP is nil")
 	}
+	if !strings.Contains(block, "/") {
```
tstromberg, line 47: Should this return `true` if `ip == block`, in this case?

govargo: I see.
NO_PROXY can be set ip(XX.XX.XX.XX) and cidr(XX.XX.XX.XX/XX).
It's case should be added.

I added the case(`ip == block`). And I changed to execute ParseCIDR only when the block is cidr.
Thanks!

Issue #7626, opened 2020-04-12
autoset --preload=false when storage driver is not overlay


afbjorklund:
So if your docker daemon runs anything else, it is broken after the `minikube start`.

```console
$ docker info | grep Driver
 Storage Driver: aufs
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
WARNING: the aufs storage-driver is deprecated, and will be removed in a future release.
```

```
$ minikube start --driver none

	[WARNING SystemVerification]: error verifying Docker info: "Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?"
```

```
I0412 09:38:01.674741   16553 exec_runner.go:78] Completed: sudo tar -I lz4 -C /var -xvf /preloaded.tar.lz4: (2.874924393s)
I0412 09:38:01.674767   16553 exec_runner.go:114] rm: /preloaded.tar.lz4
I0412 09:38:01.742410   16553 exec_runner.go:49] Run: sudo cat /var/lib/docker/image/overlay2/repositories.json
I0412 09:38:01.752632   16553 exec_runner.go:91] found /var/lib/docker/image/overlay2/repositories.json, removing ...
I0412 09:38:01.752681   16553 exec_runner.go:98] cp: memory --> /var/lib/docker/image/overlay2/repositories.json (3417 bytes)
I0412 09:38:01.752764   16553 exec_runner.go:49] Run: sudo systemctl daemon-reload
I0412 09:38:01.867623   16553 exec_runner.go:49] Run: sudo systemctl restart docker
I0412 09:38:02.357525   16553 kubeadm.go:634] prelaoding failed, will try to load cached images: sudo systemctl restart docker: exit status 1
stdout:

stderr:
Job for docker.service failed because the control process exited with error code. See "systemctl status docker.service" and "journalctl -xe" for details.
```

The systemd status is a silly `docker.service: Failed with result 'exit-code'.`
But the journal has the real error: `failed to start daemon: error initializing graphdriver: /var/lib/docker contains several valid graphdrivers: aufs, overlay2; Please cleanup or explicitly choose storage driver (-s <DRIVER>)
`

We should check the actual driver used, in case something else replaces "overlay2" in the future.

----

To restore the functionality, one needs to first `sudo rm -r /var/lib/docker/overlay2` (and restart)

Note that we do use the storage driver when _writing_ the tarball. Just not when actually reading it.

And of course, it's not flexible there either:

```go
[...]

PR #7628, opened 2020-04-12
WIP: Check if storage driver is overlay2 before preload

afbjorklund:
For #7626

### DIFF
--- a/pkg/minikube/download/preload.go
+++ b/pkg/minikube/download/preload.go
@@ -17,14 +17,17 @@ limitations under the License.
 package download
 
 import (
+	"bytes"
 	"context"
 	"crypto/md5"
 	"fmt"
 	"io/ioutil"
 	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	"cloud.google.com/go/storage"
 	"google.golang.org/api/option"
@@ -88,6 +91,22 @@ func PreloadExists(k8sVersion, containerRuntime string) bool {
 	// to track status of adding containerd & crio
 	if containerRuntime != "docker" {
 		glog.Info("Container runtime isn't docker, skipping preload")
+		return false
+	}
+
+	cmd := exec.Command("docker", "info", "-f", "{{.Info.Driver}}")
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+	output, err := cmd.Output()
+	if err != nil {
+		glog.Errorf("%v: %v:\n%s", cmd.Args, err, stderr.String())
+		return false
+	}
+	storageDriver := strings.Trim(string(output), " \n")
+
+	// See https://github.com/kubernetes/minikube/issues/7626
+	if storageDriver != "overlay2" {
+		glog.Info("Storage driver isn't overlay2, skipping preload")
 		return false
 	}
 


pkg/minikube/download/preload.go
```diff
@@ -91,6 +94,22 @@ func PreloadExists(k8sVersion, containerRuntime string) bool {
 		return false
 	}
 
+	cmd := exec.Command("docker", "info", "-f", "{{.Info.Driver}}")
```
medyagh, line 97: lets use the oci.DaemonInfo for this (add the storage driver to it)
(we can later make that a cached call so we dont have to keep calling it)

also lets try to download the image for that driver 
if that driver doesnt exist then dont preload
(we have the storage driver in the filename) so that shoudl make it easy.

so in the future if we mmake preload images for that storage driver we can suppor it
https://github.com/kubernetes/minikube/issues/7626#issuecomment-612580455

medyagh: here is the link for DaemonInfo

I believe u could update the struct with the new format, so the Storage Driver be unmarshalized into it

afbjorklund: I just copied the code from the generator

https://github.com/kubernetes/minikube/blob/master/hack/preload-images/preload_images.go#L82

afbjorklund:
The current code is just to stop minikube from killing docker (the current behavior)

Adding a working cache for non-overlayfs2 requires more work, and might not be doable
(for instance with some drivers - like [dm](https://docs.docker.com/storage/storagedriver/device-mapper-driver/), the actual content was kept on another partition)

So it was not part of this PR. Could use the so called "oci" DaemonInfo, though.

medyagh:
> The current code is just to stop minikube from killing docker (the current behavior)
> 
> Adding a working cache for non-overlayfs2 requires more work, and might not be doable
> (for instance with some drivers - like [dm](https://docs.docker.com/storage/storagedriver/device-mapper-driver/), the actual content was kept on another partition)
> 
> So it was not part of this PR. Could use the so called "oci" DaemonInfo, though.

I agree adding new storage should not be part of this PR,
what I meant was
we should check for the existance of the URL (which includes the storage driver in the filename) if the URL (with correct storage driver) doesnt exist then we will skip preload.

afbjorklund:
Need to make sure it is using the _correct_ docker daemon, when determining storage driver...

In this case, the `none` driver was using `aufs` but the `docker` driver was using `overlay2`

medyagh:
> Need to make sure it is using the _correct_ docker daemon, when determining storage driver...
> 
> In this case, the `none` driver was using `aufs` but the `docker` driver was using `overlay2`

do you mean pointing to the Docker on the user's system and not inside minikube ? if thats what you mean, we have a func "PointToHostDockerDaemon()" that we run during init

otherwise what is the __correct__ docker ?

afbjorklund:
> do you mean the Docker inside Docker should use the same storage driver as the Host ? (if yes why ? is it there evidence it doesnt work ? )

It is fine if we hardcode the internal storage driver, but the external one could be anything...

tstromberg:
What do we need to move this PR forward?

afbjorklund:
> What do we need to move this PR forward?

I got confused half-way through the implementation...

I thought that we could continue with hardcoding the internal storage driver (to `overlay2`) and let the user choose whatever external storage driver they want... But it turns out that if you pick an odd outer storage, that could limit your internal options (to something basic like `vfs`). So we probably need to re-visit this, and have it silently fallback to not use _any_ preloads in such cases ?

tstromberg:
Closing until we're sure what to do with this. Please feel free to `/reopen` once clarity can be achieved.

Issue #7480, opened 2020-04-07
The podman driver should not require sudo or root


afbjorklund:
Even though podman is run with sudo (docker uses a group instead), the driver should **not**.

```
✨  Using the podman (experimental) driver based on user configuration
💣  The "podman" driver requires root privileges. Please run minikube using 'sudo minikube --driver=podman'.
```

This will lead to the same kind of ownership and path issues that is plaguing the none driver...

Instead only the `podman` commands should be wrapped in `sudo` (to not try to run as rootless)

afbjorklund:
i.e. for the docker driver we require that the user is part of the root-equivalent `docker` group:

https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user

```console
$ docker ps
```

but for the podman driver we instead require that the user has passwordless access to `podman`

```console
$ sudo podman ps
```

This basically allows the same level of root access as docker, that is you still need to be an admin

----

Neither driver should try to wrap `minikube` command in `sudo`, but run it as a regular user.

Running minikube (or kubernetes) entirely as a non-root user is currently **not** supported.

See e.g. https://github.com/rootless-containers/usernetes

Running minikube requires a user with enough privileges.

afbjorklund:
ping @medyagh

tstromberg:
Related thread: https://twitter.com/rawkode/status/1239885013241470979

We can detect the necessity to run root via:

`podman info | grep rootless`

afbjorklund:
@tstromberg:

> We can detect the necessity to run root via:
> 
> `podman info | grep rootless`

We **don't** support running the docker-in-podman with rootless podman, only with the regular root one.

Eventually we might default to crio-in-podman, but for now that requires using `--container-runtime`

afbjorklund:
Now works OK, when running the `docker` container runtime.

```console
$ ./out/minikube kubectl get nodes
E0413 17:36:02.552882   31952 api_server.go:169] unable to get freezer state: cat: /sys/fs/cgroup/freezer/libpod_parent/libpod-a2ebc4baa2e317a42afb6ff01d07fab6ec608a5ab2893425e6e58b8203748c3d/kubepods/burstable/pod6c3eb4fb6e1f774f1248d50ac61483b5/076cdc3a21654ba0d67ccf4a21af54e7f2a31a3edaf0d719947635c36586f55a/freezer.state: No such file or directory
NAME       STATUS   ROLES    AGE   VERSION
[...]

PR #7631, opened 2020-04-12
Add "sudo" to podman calls

afbjorklund:
~~TODO is to replace three hundred places of exec.Command OCIBinary with appropriate "sudo".~~

To make it easier one could use a dummy wrapper like "env", for the docker commands as well.

Then it could be like: exec.Command(OCIPrefix, OCIBinary, ...)

`env docker info`

`sudo podman info`

Instead of having to manipulate arrays during the interpolation ?

For #7480


afbjorklund:
> as part of PR please change this in Jenkins so it does not run with sudo in intergraiton test either. https://github.com/kubernetes/minikube/blob/master/hack/jenkins/linux_integration_tests_podman.sh#L34

Done.

afbjorklund:
Also need to document that user needs to setup password-less sudo for podman.

Otherwise this is going to throw `Password:` prompts in the middle of minikube...

afbjorklund:
Now the API is genuinely horrible, with two strings instead of a class or something.

Also found some places where it is called with the **Driver** in the place of "ociBin"...

afbjorklund:
There are some nice little status differences as well:

`the created container "minikube" has a running status.`

`Error: cannot exec into container that is not running: container state improper`

So apparently "running" is not enough to be running. :-(

afbjorklund:
Here is the current startup with `podman`:

```
INFO: ensuring we can execute /bin/mount even with userns-remap
INFO: remounting /sys read-only
INFO: making mounts shared
INFO: fix cgroup mounts for all subsystems
INFO: clearing and regenerating /etc/machine-id
INFO: faking /sys/class/dmi/id/product_name to be "kind"
INFO: faking /sys/class/dmi/id/product_uuid to be random
INFO: faking /sys/devices/virtual/dmi/id/product_uuid as well
INFO: setting iptables to detected mode: legacy
update-alternatives: error: no alternatives for iptables

```

Here is the matching startup with `docker`:

```
INFO: ensuring we can execute /bin/mount even with userns-remap
INFO: detected aufs, calling sync
INFO: remounting /sys read-only
INFO: making mounts shared
INFO: fix cgroup mounts for all subsystems
INFO: clearing and regenerating /etc/machine-id
INFO: faking /sys/class/dmi/id/product_name to be "kind"
INFO: faking /sys/class/dmi/id/product_uuid to be random
INFO: faking /sys/devices/virtual/dmi/id/product_uuid as well
INFO: setting iptables to detected mode: legacy
systemd 242 running in system mode. (+PAM +AUDIT +SELINUX +IMA +APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=hybrid)
Detected virtualization docker.
Detected architecture x86-64.

Welcome to Ubuntu 19.10!
```

That is: the podman start fails (exit 2)

afbjorklund:
When doing it manually, I get:

```
$ sudo podman run -it ubuntu:19.10

# apt update
# apt install -y iptables
```
```
Setting up iptables (1.8.3-2ubuntu5) ...
update-alternatives: using /usr/sbin/iptables-legacy to provide /usr/sbin/iptables (iptables) in auto mode
update-alternatives: using /usr/sbin/ip6tables-legacy to provide /usr/sbin/ip6tables (ip6tables) in auto mode
update-alternatives: using /usr/sbin/arptables-nft to provide /usr/sbin/arptables (arptables) in auto mode
update-alternatives: using /usr/sbin/ebtables-nft to provide /usr/sbin/ebtables (ebtables) in auto mode
```

afbjorklund:
When tracing the entrypoint (\<sigh\>), it seems it was deleted (in podman):

```
openat(AT_FDCWD, "/var/lib/dpkg/alternatives/iptables", O_RDONLY) = -1 ENOENT (No such file or directory)
write(2, "update-alternatives: error: ", 28update-alternatives: error: ) = 28
write(2, "no alternatives for iptables", 28no alternatives for iptables
write(2, "\n", 1
)                       = 1
```

This is something that kind base has, wonder why it works in docker ?

```
openat(AT_FDCWD, "/var/lib/dpkg/alternatives/iptables", O_RDONLY) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=287, ...}) = 0
fstat(3, {st_mode=S_IFREG|0644, st_size=287, ...}) = 0
read(3, "auto\n/usr/sbin/iptables\niptables"..., 4096) = 287
stat("/usr/sbin/iptables-legacy", {st_mode=S_IFREG|0755, st_size=99272, ...}) = 0
stat("/usr/sbin/iptables-nft", {st_mode=S_IFREG|0755, st_size=232880, ...}) = 0
close(3)                                = 0

```

afbjorklund:
@medyagh:
It is because minikube replaces `/var` with an empty directory, killing dpkg (and iptables)!

https://github.com/kubernetes/minikube/blob/v1.9.2/pkg/drivers/kic/oci/oci.go#L115_L123

It needs to restore the old contents, after remounting var. Even better, **only** do /var/lib/minikube ?

Once /var is back, it boots properly in podman as well. Now, there was some `sudo` missing...

afbjorklund:
One side-effect of mounting /var properly, is that you cannot delete a machine anymore.

This is because some of the files under ~/.minikube/machines/*/var are now root-owned...

afbjorklund:
Need to volume mount the container runtime directories too, as well as `/var/lib/minikube`.

Now with everything setup, it boots properly. On my machine it took 38 seconds, not too bad.

afbjorklund:
I went back to /var - it seems it was the path mount that was causing it

With a regular named mount, the image boots up just fine - no missing files

afbjorklund:
Added better status check:

```
❗  'podman' driver reported an issue: "sudo -n podman info" exit status 1: sudo: a password is required
💡  Suggestion: Add your user to the 'sudoers' file: 'anders ALL=(ALL) NOPASSWD: /usr/local/bin/podman'
📘  Documentation: https://podman.io

💣  Failed to validate 'podman' driver
```

pkg/minikube/node/cache.go
```diff
@@ -99,14 +99,20 @@ func doCacheBinaries(k8sVersion string) error {
...
+		if !image.ExistsImageInDaemon(kic.BaseImage) {
+			out.T(out.Pulling, "Pulling base image ...")
+			g.Go(func() error {
+				glog.Infof("Downloading %s to local daemon", kic.BaseImage)
+				return image.WriteImageToDaemon(kic.BaseImage)
+			})
+		}
+	} else {
+		// TODO: driver == "podman"
+		glog.Info("Container runtime isn't docker, skipping download")
```
medyagh, line 115: was this supposed to say driver or container run time ?
because the if checks for driver, but else says runtime.

tstromberg:
FWIW, podman, at least as installed on Fedora 32, doesn't appear to need sudo. I haven't actually tried this with minikube yet, because when running v1.10.0-beta.1 as root, I only get as far as the issue @afbjorklund cited:

```
INFO: setting iptables to detected mode: legacy
update-alternatives: error: no alternatives for iptables
```

Any clue on how to get passed this?

afbjorklund:
> FWIW, podman, at least as installed on Fedora 32, doesn't appear to need sudo.

Just like docker*, podman can be run with (as root) or without (rootless) privileges

You can do _some_ things with rootless podman, but we **don't** support it for minikube.
So therefore the user needs to have (password-less) `sudo` access to `podman`.

\* I think `dockerd --rootless` is normally only used with [RootlessKit](https://github.com/rootless-containers/rootlesskit), not general use.

> I haven't actually tried this with minikube yet, because when running v1.10.0-beta.1 as root, I only get as far as the issue @afbjorklund cited:
> 
> ```
> INFO: setting iptables to detected mode: legacy
> update-alternatives: error: no alternatives for iptables
> ```
> 
> Any clue on how to get passed this?

When changing the mount option for `/var`, this podman problem goes away.
It also makes it more similar to docker, as well as much easier to clean up.

```diff
        }
 
        if p.OCIBinary == Podman { // enable execing in /var
-               // volume path in minikube home folder to mount to /var
-               hostVarVolPath := filepath.Join(localpath.MiniPath(), "machines", p.Name, "var")
-               if err := os.MkdirAll(hostVarVolPath, 0711); err != nil {
-                       return errors.Wrapf(err, "create var dir %s", hostVarVolPath)
-               }
                // podman mounts var/lib with no-exec by default  https://github.com/containers/libpod/issues/5103
-               runArgs = append(runArgs, "--volume", fmt.Sprintf("%s:/var:exec", hostVarVolPath))
+               runArgs = append(runArgs, "--volume", fmt.Sprintf("%s:/var:exec", p.Name))
        }
        if p.OCIBinary == Docker {
                runArgs = append(runArgs, "--volume", fmt.Sprintf("%s:/var", p.Name))

```

Before, it created files owned by root in the machine directory of the user.
Thus, the files could not just be removed but needed wrapping in `sudo`...

By letting the container runtime manage the volumes, cleanup is easy.

I'm not sure what the real problem was, nor am I very keen to find out ?

pkg/minikube/node/cache.go
afbjorklund: We can change this, but it is only for the internal logging anyway.

Much better would be to fix it, by dropping go-containerregistry ?

afbjorklund:
Tested OK on Ubuntu 18.04 LTS, with both docker and podman drivers - and with all 3 runtimes!

Both docker-env and podman-env work OK, altough we need to add "varlink" package to the ISO

cmd/minikube/cmd/delete.go
```diff
@@ -90,17 +90,17 @@ func init() {
 
 func deleteContainersAndVolumes() {
 	delLabel := fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "true")
-	errs := oci.DeleteContainersByLabel(oci.Docker, delLabel)
+	errs := oci.DeleteContainersByLabel("env", oci.Docker, delLabel)
```
medyagh, line 93: what do u think about moving "env" to the oci package, and name it something like "oci.DefaultSudoPrefix" or "oci.Prefix"

hack/jenkins/linux_integration_tests_podman.sh
```diff
@@ -31,7 +31,6 @@ JOB_NAME="Experimental_Podman_Linux"
 
 mkdir -p cron && gsutil -qm rsync "gs://minikube-builds/${MINIKUBE_LOCATION}/cron" cron || echo "FAILED TO GET CRON FILES"
 sudo install cron/cleanup_and_reboot_Linux.sh /etc/cron.hourly/cleanup_and_reboot || echo "FAILED TO INSTALL CLEANUP"
-SUDO_PREFIX="sudo -E "
```
medyagh, line 34: what would happen when the test invokes minikube status , and will invoke sudo podman inspect....wouldn't we need this sudo ?

afbjorklund: We're trying to avoid running `sudo minikube` (like we do for "none"), since it puts the files in /root or makes them owned by root and such.

You are perfectly right that we still need `sudo podman`, so it is likely to fail unless the CI servers have a rule to run it (without PASSWD)

cmd/minikube/cmd/delete.go
afbjorklund: I'm not really sure what these const strings bring to the table, maybe I can read more somewhere ? Like definining Foo as "foo" and Bar as "bar"

But if we do, they are likely to be called oci.Env and oci.Sudo instead of something odd.

afbjorklund:
This test result looks a bit odd:

```
start.go:585] status for podman: {Installed:true Healthy:false Error:exit status 125 Fix:Cant verify mininim required version for podman . See podman website for installation guide. Doc:https://podman.io/getting-started/installation.html}
	
	! 'podman' driver reported an issue: exit status 125
	* Suggestion: Cant verify mininim required version for podman . See podman website for installation guide.
	* Documentation: https://podman.io/getting-started/installation.html
	
	X Failed to validate 'podman' driver
```

Probably need some better output logging for it:

```go
        // Allow no more than 2 seconds for version command
        ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
        defer cancel()

        cmd := exec.CommandContext(ctx, oci.Podman, "version", "-f", "{{.Version}}")
        o, err := cmd.CombinedOutput()
        output := string(o)
        if err != nil {
                return registry.State{Error: err, Installed: true, Healthy: false, Fix: "Cant verify mininim required version for podman . See podman website for installation guide.", Doc: "https://podman.io/getting-started/installation.html"}
        }

```

afbjorklund:
One reason could be that it has `podman-remote' installed, but not installed the socket.

See https://www.projectatomic.io/blog/2018/05/podman-varlink/ for some background

That would lead to that exit code, so will check that by adding some output logging...

```console
$ podman version || echo $?
Version:            1.8.2
RemoteAPI Version:  1
Go Version:         go1.13.8
Git Commit:         028e3317eb1494b9b2acba4a0a295df80fae66cc
Built:              Mon Apr 13 15:02:24 2020
OS/Arch:            linux/amd64
$ podman-remote version || echo $?
Error: could not get runtime: dial unix /run/podman/io.podman: connect: permission denied
125
```

I don't really _want_ to support remote podman, but that's another story...

It seems like remote docker is used (on Mac/Win), even it works so-so.

afbjorklund:
And perhaps obvious but, the current approach will **not** work for `podman-remote`...

So it needs to be either of `sudo podman` (locally) or `podman-remote` (remotely)

afbjorklund:
@medyagh : this branch has now accumulated some bug fixes, in addition to the original feature:

* Don't download kicbase image to docker daemon when using podman (skip the download)

* Mounting /var as a named volume instead of as a path, to make it boot again ("iptables" bug)

We still need to fix this if we want the instructions to work but I guess we could look at cherry-picking

afbjorklund:
> That would lead to that exit code, so will check that by adding some output logging...

It was a configuration issue, we set the cgroup config for docker but not for podman:

```
0426 14:09:29.471143    6943 podman.go:85] podman version returned time="2020-04-26T14:09:29-07:00" level=warning msg="The cgroupv2 manager is set to systemd but there is no systemd user session available"

	time="2020-04-26T14:09:29-07:00" level=warning msg="For using systemd, you may need to login using an user session"

	time="2020-04-26T14:09:29-07:00" level=warning msg="Alternatively, you can enable lingering with: `loginctl enable-linger 1004` (possibly as root)"

	time="2020-04-26T14:09:29-07:00" level=warning msg="Falling back to --cgroup-manager=cgroupfs"
```

The real problem (below it) is however that we don't have any user namespaces:
```
	cannot clone: Operation not permitted

	user namespaces are not enabled in /proc/sys/kernel/unprivileged_userns_clone

	Error: could not get runtime: cannot re-exec process

```

So podman (in KIC) needs to be modified, to use `cgroup_manager = "cgroupfs"`

```console
$ docker info | grep -i driver
 Storage Driver: overlay2
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
WARNING: No swap limit support
```

And we need to use `sudo podman version`, was missed when adding the prefix.

afbjorklund:
@TravisBuddy : Hey "buddy", my code seems OK but we have a flaky test

https://travis-ci.org/github/kubernetes/minikube/jobs/680154938#L684

```
E0427 15:22:12.840394   11590 out.go:102] [unset errFile]: * 

E0427 15:22:12.840622   11590 out.go:102] [unset errFile]: X Error getting primary control plane: remove /tmp/profile870321857/.minikube/profiles/config.json: no such file or directory

E0427 15:22:12.840639   11590 out.go:102] [unset errFile]: * 

E0427 15:22:12.840649   11590 out.go:102] [unset errFile]: * minikube is exiting due to an error. If the above message is not useful, open an issue:

E0427 15:22:12.840663   11590 out.go:102] [unset errFile]:   - https://github.com/kubernetes/minikube/issues/new/choose

FAIL	k8s.io/minikube/pkg/addons	0.029s
```

pkg/drivers/kic/oci/cli_runner.go
```diff
@@ -63,8 +64,24 @@ func (rr RunResult) Output() string {
 	return sb.String()
 }
 
+// PrefixCmd adds any needed prefix (such as sudo) to the command
+func PrefixCmd(cmd *exec.Cmd) *exec.Cmd {
+	if cmd.Args[0] == Podman && runtime.GOOS == "linux" { // want sudo when not running podman-remote
```
tstromberg, line 69: This will make a good insertion point for folks who want to add support for handling `rootless`. Thank you!

PR #7636, opened 2020-04-13
bump lint

medyagh:
### bumped lint version :)

- bump golang-ci version
- bump go version to latest available
- putting a comment to specify where we can find latest available golang for kuberentes project
- remove un-needed replaces 

### didn't bump kubernetes version :(
tried to bump kubernetes version but I faced this issue https://github.com/kubernetes/minikube/issues/7638 till that is solved will not bump kubernets version which has nice nice new features such as passing context to client-go calls

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,8 @@ DEB_VERSION ?= $(subst -,~,$(RAW_VERSION))
 RPM_VERSION ?= $(DEB_VERSION)
 
 # used by hack/jenkins/release_build_and_upload.sh and KVM_BUILD_IMAGE, see also BUILD_IMAGE below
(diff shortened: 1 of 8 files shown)


afbjorklund:
Would it be possible to do this in two _separate_ PRs instead ? Hard to review/test with all-in-one.

medyagh:
> Would it be possible to do this in two _separate_ PRs instead ? Hard to review/test with all-in-one.

the golang bump didnt invoke any code change, this PR is basicly bumping lint + golang (golang could be easily deleted as one line change)

medyagh:
if it helps to review, most of them are un-needed fmt.Sprtinf("simple string")

PR #7640, opened 2020-04-13
Podman bugs

afbjorklund:
Here are some of the various podman driver bugs that I found, when trying to fix #7631

It seems like we have docker hardcoded in a _lot_ of places, and that podman is untested.

### DIFF
--- a/pkg/minikube/node/cache.go
+++ b/pkg/minikube/node/cache.go
@@ -99,13 +99,17 @@ func doCacheBinaries(k8sVersion string) error {
 }
 
 // BeginDownloadKicArtifacts downloads the kic image + preload tarball, returns true if preload is available
-func beginDownloadKicArtifacts(g *errgroup.Group) {
+func beginDownloadKicArtifacts(g *errgroup.Group, driver string, cRuntime string) {
 	out.T(out.Pulling, "Pulling base image ...")
-	glog.Info("Beginning downloading kic artifacts")
-	g.Go(func() error {
-		glog.Infof("Downloading %s to local daemon", kic.BaseImage)
-		return image.WriteImageToDaemon(kic.BaseImage)
-	})
+	glog.Infof("Beginning downloading kic artifacts for %s with %s", driver, cRuntime)
+	if driver == "docker" {
+		g.Go(func() error {
+			glog.Infof("Downloading %s to local daemon", kic.BaseImage)
+			return image.WriteImageToDaemon(kic.BaseImage)
+		})
+	}
+	// TODO: driver == "podman"
+	glog.Info("Container runtime isn't docker, skipping download")
 }
 
 // WaitDownloadKicArtifacts blocks until the required artifacts for KIC are downloaded.
--- a/pkg/drivers/kic/oci/info.go
+++ b/pkg/drivers/kic/oci/info.go
@@ -234,7 +234,7 @@ func dockerSystemInfo() (dockerSysInfo, error) {
 // podmanSysInfo returns podman system info --format '{{json .}}'
 func podmanSystemInfo() (podmanSysInfo, error) {
 	var ps podmanSysInfo
-	cmd := exec.Command(Podman, "system", "info", "--format", "'{{json .}}'")
+	cmd := exec.Command(Podman, "system", "info", "--format", "{{json .}}")
 	out, err := cmd.CombinedOutput()
 	if err != nil {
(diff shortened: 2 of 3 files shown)


afbjorklund:
Hmm, wonder why the "unable to find" didn't work

pkg/minikube/node/cache.go
```diff
@@ -99,13 +99,17 @@ func doCacheBinaries(k8sVersion string) error {
...
-func beginDownloadKicArtifacts(g *errgroup.Group) {
+func beginDownloadKicArtifacts(g *errgroup.Group, driver string, cRuntime string) {
 	out.T(out.Pulling, "Pulling base image ...")
-	glog.Info("Beginning downloading kic artifacts")
-	g.Go(func() error {
-		glog.Infof("Downloading %s to local daemon", kic.BaseImage)
-		return image.WriteImageToDaemon(kic.BaseImage)
-	})
+	glog.Infof("Beginning downloading kic artifacts for %s with %s", driver, cRuntime)
+	if driver == "docker" {
```
medyagh, line 105: shouldnt this be driver.IsKIC( ?

afbjorklund: It is driver.IsKIC (in the caller), but this function calls the docker-specific WriteImageToDaemon

afbjorklund:
So we just skip the cache, and let the `podman run` pull the image directly instead.

afbjorklund:
Note that there is no daemon for local podman, only for remote podman (varlink socket).

So this does not make so much sense either: `Try: restarting podman.`

Maybe that should be `Try: reinstalling podman` or something like that ?

Also still wonder why this is called "OCI",  but I guess it has to be called _something_.

i.e. this is mostly about the docker cli, and not so much about opencontainers.org

the standard is mostly about `runc`, and most people still use `docker://` transport

But I'm not about to change any of that, I just want it to be running (similar to docker)

afbjorklund:
Also need to document that user needs to setup password-less sudo for podman.

Otherwise this is going to throw `Password:` prompts in the middle of minikube...

pkg/minikube/node/cache.go
medyagh: Let's use oci.Docker not string

pkg/minikube/node/cache.go
```diff
@@ -99,13 +99,17 @@ func doCacheBinaries(k8sVersion string) error {
...
+		g.Go(func() error {
+			glog.Infof("Downloading %s to local daemon", kic.BaseImage)
+			return image.WriteImageToDaemon(kic.BaseImage)
+		})
+	}
+	// TODO: driver == "podman"
+	glog.Info("Container runtime isn't docker, skipping download")
```
medyagh, line 112: Did you mean the driver is not docker ?

pkg/minikube/node/cache.go
```diff
@@ -99,13 +99,17 @@ func doCacheBinaries(k8sVersion string) error {
 }
 
 // BeginDownloadKicArtifacts downloads the kic image + preload tarball, returns true if preload is available
-func beginDownloadKicArtifacts(g *errgroup.Group) {
```
medyagh, line 102: cRuntime is not used here

medyagh:
On the oci naming it was hard to find any name that wasn't wrong. I looked up Wikipedia on both docker and podman and I saw what they had in common is they are both and oci implementation.

By oci I meant anytime we have to talk to docker cli or podman cli outside the minikube node ( on th user's host machine and not using ssh runner or kic runner)

pkg/minikube/node/cache.go
afbjorklund: Well, the driver _also_ has a container runtime. Which is of course confusing.

We download `kicbase` for the outer runtime, and kubernetes for the inner...
But the commands and messages are the same for both of the runtimes.

pkg/minikube/node/cache.go
afbjorklund: oci.Docker _is_ a string, but ok. Would be nice to have something generic, in `image` ? Preferrably with a progressbar.

pkg/minikube/node/cache.go
afbjorklund: I thought it would be, but we don't use the preload hack for the kicbase image

afbjorklund:
I think most of this branch/PR will end up getting merged into #7631 instead.

PR #7648, opened 2020-04-13
WIP: generate join token on control plane

### LINKED ISSUE #7620: multinode: generating join token: open /root/.kube/config: no such file or directory
```
$ out/minikube version
minikube version: v1.9.2
commit: 2ef15e6a3cdc48db23910ba66912557ed3a776a9
```

1. start cluster with multi node. This fails

```
$ LC_ALL=en out/minikube start -n 2 --driver docker
😄  minikube v1.9.2 on Darwin 10.14.5
🚜  Pulling base image ...

👍  Starting node minikube-m02 in cluster minikube
🚜  Pulling base image ...
🌐  Found network options:
    ▪ NO_PROXY=172.17.0.3

💣  failed to start node: adding node: generating join token: generating bootstrap token: /bin/bash -c "sudo env PATH=/var/lib/minikube/binaries/v1.18.0:$PATH kubeadm token create --print-join-command --ttl=0": Process exited with status 1
stdout:

stderr:
failed to load admin kubeconfig: open /root/.kube/config: no such file or directory


```

I tried docker, virtualbox, hyperkit drivers. These driver doesn't work.

The Error was caused here.

https://github.com/kubernetes/minikube/blob/2ef15e6a3cdc48db23910ba66912557ed3a776a9/pkg/minikube/node/start.go#L164-L167

https://github.com/kubernetes/minikube/blob/2ef15e6a3cdc48db23910ba66912557ed3a776a9/pkg/minikube/bootstrapper/kubeadm/kubeadm.go#L561-L565

These code aren't changed. I guess this is due to #7591 (changing SSH runner).
But sorry, I have no idea about how to fix this.

sharifelgamal:
A refactor resulted in trying to generate the join token on the worker node, rather than the control plane, this restores the old behavior and should fix multi-node.

Fixes #7620

### DIFF
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -156,11 +156,24 @@ func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
 			return nil, errors.Wrap(err, "Updating node")
 		}
 
-		cpBs, err := cluster.Bootstrapper(starter.MachineAPI, viper.GetString(cmdcfg.Bootstrapper), *starter.Cfg, starter.Runner)
+		// Load the control plane host
+		cp, err := config.PrimaryControlPlane(starter.Cfg)
 		if err != nil {
-			return nil, errors.Wrap(err, "Getting bootstrapper")
+			return nil, errors.Wrap(err, "Getting primary control plane")
+		}
+		cpHost, err := machine.LoadHost(starter.MachineAPI, driver.MachineName(*starter.Cfg, cp))
+		if err != nil {
+			return nil, errors.Wrap(err, "Loading control plane host")
+		}
+		r, err := machine.CommandRunner(cpHost)
+		if err != nil {
+			return nil, errors.Wrap(err, "Getting cp command runner")
 		}
 
+		cpBs, err := cluster.Bootstrapper(starter.MachineAPI, viper.GetString(cmdcfg.Bootstrapper), *starter.Cfg, r)
+		if err != nil {
+			return nil, errors.Wrap(err, "Failed to get control plane bootstrapper")
+		}
 		joinCmd, err := cpBs.GenerateToken(*starter.Cfg)
 		if err != nil {
 			return nil, errors.Wrap(err, "generating join token")
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -157,20 +157,9 @@ func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
 		}
 
 		// Load the control plane host
-		cp, err := config.PrimaryControlPlane(starter.Cfg)
-		if err != nil {
-			return nil, errors.Wrap(err, "Getting primary control plane")
-		}
-		cpHost, err := machine.LoadHost(starter.MachineAPI, driver.MachineName(*starter.Cfg, cp))
-		if err != nil {
-			return nil, errors.Wrap(err, "Loading control plane host")
-		}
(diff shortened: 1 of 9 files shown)


pkg/minikube/node/start.go
```diff
@@ -156,11 +156,24 @@ func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
...
 		}
 
-		cpBs, err := cluster.Bootstrapper(starter.MachineAPI, viper.GetString(cmdcfg.Bootstrapper), *starter.Cfg, starter.Runner)
+		// Load the control plane host
+		cp, err := config.PrimaryControlPlane(starter.Cfg)
 		if err != nil {
-			return nil, errors.Wrap(err, "Getting bootstrapper")
+			return nil, errors.Wrap(err, "Getting primary control plane")
+		}
+		cpHost, err := machine.LoadHost(starter.MachineAPI, driver.MachineName(*starter.Cfg, cp))
```
medyagh, line 164: could mustload Package be used for this ?

sharifelgamal: yep, good call. mustload.Running does all this.

pkg/minikube/node/start.go
```diff
@@ -156,11 +156,13 @@ func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
...
 		}
 
-		cpBs, err := cluster.Bootstrapper(starter.MachineAPI, viper.GetString(cmdcfg.Bootstrapper), *starter.Cfg, starter.Runner)
+		// Load the control plane host
+		c := mustload.Running(starter.Cfg.Name)
+
+		cpBs, err := cluster.Bootstrapper(starter.MachineAPI, viper.GetString(cmdcfg.Bootstrapper), *starter.Cfg, c.CP.Runner)
 		if err != nil {
-			return nil, errors.Wrap(err, "Getting bootstrapper")
+			return nil, errors.Wrap(err, "Failed to get control plane bootstrapper")
```
medyagh, line 164: to match other error.Wraps, it should be, errors.Wrapf(err, "getting bootstrapper for %s",starter.Cfg.Name)

sharifelgamal: add the name of the cluster doesn't really help since there's no confusion about what cluster you're trying to start, I did change the format of the error message to conform with the rest of the file

sharifelgamal:
ok, so something is going wrong with multi-node and docker containers and I'm having trouble pinpointing exactly what's happening. essentially, the second docker container is exiting after awhile and everything is failing afterwards. The entirely of the docker logs for that container is as follows:

```
INFO: ensuring we can execute /bin/mount even with userns-remap
INFO: remounting /sys read-only
INFO: making mounts shared
INFO: fix cgroup mounts for all subsystems
INFO: clearing and regenerating /etc/machine-id
INFO: faking /sys/class/dmi/id/product_name to be "kind"
INFO: faking /sys/class/dmi/id/product_uuid to be random
INFO: faking /sys/devices/virtual/dmi/id/product_uuid as well
INFO: setting iptables to detected mode: legacy
update-alternatives: error: no alternatives for iptables
```

afbjorklund:
@sharifelgamal : probably the same as https://github.com/kubernetes/minikube/pull/7631#issuecomment-612870400 and https://github.com/kubernetes/minikube/pull/7631#issuecomment-612911774 ?

i.e. something is deleting the required `/var/lib/dpkg/alternatives/iptables` files

sharifelgamal:
closing in favor of #7700

PR #7653, opened 2020-04-13
wip: improve kic start restart and stop

medyagh:
as suggested by @tstromberg :
I am splitting this PR https://github.com/kubernetes/minikube/pull/7608 into smaller PRs :

### improve kic start and stop and restart by:
- allow restart proceed even if can't get state.
- container status to return state.State type.
- check container is running status after start.

### DIFF
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -39,6 +39,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/cruntime"
 	"k8s.io/minikube/pkg/minikube/download"
 	"k8s.io/minikube/pkg/minikube/sysinit"
+	"k8s.io/minikube/pkg/util/retry"
 )
 
 // Driver represents a kic driver https://minikube.sigs.k8s.io/docs/reference/drivers/docker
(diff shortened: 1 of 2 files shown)


pkg/drivers/kic/kic.go
```diff
@@ -357,8 +346,14 @@ func (d *Driver) Stop() error {
...
 	}
 
+	// this is needed only by kic drivers because of the entry-point of the base image overrides the STOP and KILL SIGs
+	// other drivers dont need this
 	if err := killAPIServerProc(d.exec); err != nil {
-		glog.Warningf("couldn't stop kube-apiserver proc: %v", err)
+		glog.Warningf("kill kube-apiserver proc: %v", err)
+	}
+
+	if err := killETCDProc(d.exec); err != nil {
```
tstromberg, line 355: The driver should absolutely not need to know about etcd by name.  Please remove this so that  it does not paper over a more fundamental driver flaw.

medyagh: done removed boht. i belive in the following up PR to try shut down before stop, we will not have the need to kill either apiserver or ectc so I deleted both of them but this PR should be merged after that other PR that fixes stop

PR #7654, opened 2020-04-13
wip: wait for sshd service to be up before returning ssh runner

medyagh:
as suggested by @tstromberg :
I am splitting this PR https://github.com/kubernetes/minikube/pull/7608 into smaller PRs :

### Wait for SSHD service to be up before returning SSH runner
- if too many docker daemon is under pressure, a container is created and it shows running
but the SSHD service might need couple more seconds to be up !

this PR ensures that before we return the SSH runner we use the Kic runner to ensure the SSH is up and then we return ssh runner. this might slow minikube down 1 second but it worth adding the reliablity to minikube that if tens of minikube clusters run under pressure they proceed reliably

### DIFF
--- a/pkg/minikube/bootstrapper/bsutil/kverify/system_pods.go
+++ b/pkg/minikube/bootstrapper/bsutil/kverify/system_pods.go
@@ -164,3 +164,14 @@ func KubeletStatus(cr command.Runner) state.State {
 	}
 	return state.Stopped
 }
+
+// SSHDStatus checks the sshd service status
+func SSHDStatus(cr command.Runner) state.State {
+	active := sysinit.New(cr).Active("sshd")
+	if active {
+		glog.Infof("sshd service is active.")
+		return state.Running
+	}
+	glog.Infof("sshd service is not active.")
+	return state.Stopped
+}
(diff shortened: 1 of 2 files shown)


tstromberg:
> as suggested by @tstromberg :
> I am splitting this PR #7608 into smaller PRs :
> 
> ### Wait for SSHD service to be up before returning SSH runner
> * if too many docker daemon is under pressure, a container is created and it shows running
>   but the SSHD service might need couple more seconds to be up !
> 
> this PR ensures that before we return the SSH runner we use the Kic runner to ensure the SSH is up and then we return ssh runner. this might slow minikube down 1 second but it worth adding the reliablity to minikube that if tens of minikube clusters run under pressure they proceed reliably

Can you provide proof of this PR helping?

medyagh:
> > as suggested by @tstromberg :
> > I am splitting this PR #7608 into smaller PRs :
> > ### Wait for SSHD service to be up before returning SSH runner
> > 
> > * if too many docker daemon is under pressure, a container is created and it shows running
> >   but the SSHD service might need couple more seconds to be up !
> > 
> > this PR ensures that before we return the SSH runner we use the Kic runner to ensure the SSH is up and then we return ssh runner. this might slow minikube down 1 second but it worth adding the reliablity to minikube that if tens of minikube clusters run under pressure they proceed reliably
> 
> Can you provide proof of this PR helping?

I dont really have evidence, I am gonna hold on to this PR till I am proven this is a good idea

PR #7672, opened 2020-04-15
docker/podman: graceful shutdown to avoid stuck containers

medyagh:
redoing this PR from sctratch after getting lost in the woods !
 https://github.com/kubernetes/minikube/pull/7660

### After this PR no more container get stuck:

```
medmac@~/workspace/minikube (kic_stuck_redo1) $ ./out/minikube start --driver=docker
🚜  Pulling base image ...
medmac@~/workspace/minikube (kic_stuck_redo1) $ ./out/minikube pause
⏯️  Paused kubelet and 8 containers in: kube-system, kubernetes-dashboard, storage-gluster, istio-operator
medmac@~/workspace/minikube (kic_stuck_redo1) $ ./out/minikube delete --all
🔥  Removing /Users/medmac/.minikube/machines/minikube ...
medmac@~/workspace/minikube (kic_stuck_redo1) $ docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
```

Also added integration tests, so this might never happen again !

### DIFF
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -282,6 +282,10 @@ func (d *Driver) Remove() error {
 		}
 
 	}
+
+	if id, err := oci.ContainerID(d.OCIBinary, d.MachineName); err == nil && id != "" {
+		return fmt.Errorf("expected no container ID be found for %q after delete. but got %q", d.MachineName, id)
+	}
 	return nil
 }
 
@@ -368,6 +372,11 @@ func (d *Driver) Stop() error {
 		return errors.Wrapf(err, "stopping %s", d.MachineName)
 	}
 	return nil
+}
+
+// RunSSHCommandFromDriver implements direct ssh control to the driver
+func (d *Driver) RunSSHCommandFromDriver() error {
+	return fmt.Errorf("driver does not support RunSSHCommandFromDriver commands")
 }
 
 // killAPIServerProc will kill an api server proc if it exists
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -268,7 +268,7 @@ func (d *Driver) Kill() error {
 // Remove will delete the Kic Node Container
 func (d *Driver) Remove() error {
 	if _, err := oci.ContainerID(d.OCIBinary, d.MachineName); err != nil {
-		glog.Info("could not find the container %s to remove it. will try anyways", d.MachineName)
+		glog.Infof("could not find the container %s to remove it. will try anyways", d.MachineName)
 	}
 
 	if err := oci.DeleteContainer(d.NodeConfig.OCIBinary, d.MachineName); err != nil {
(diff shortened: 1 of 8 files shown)


test/integration/pause_test.go
```diff
@@ -0,0 +1,88 @@
...
+		for _, tc := range tests {
+			tc := tc
+			t.Run(tc.name, func(t *testing.T) {
+				tc.validator(ctx, t, profile)
+			})
+		}
+	})
+}
+
+// validateFreshStart
```
tstromberg, line 56: remove meaningless comment.

test/integration/pause_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+func TestPause(t *testing.T) {
+	MaybeParallel(t)
+
+	type validateFunc func(context.Context, *testing.T, string)
+	profile := UniqueProfileName("pause")
+	ctx, cancel := context.WithTimeout(context.Background(), Minutes(30))
+	defer CleanupWithLogs(t, profile, cancel)
+
+	// Serial tests
```
tstromberg, line 35: optional - but IMHO, this test would be more readable if it was simplified into a single test-case.

medyagh: I think when something fails it's much nicer to see what failed and also when someone reads the report they can see what functionalities failed and what passed.

pkg/minikube/machine/stop.go
```diff
@@ -78,8 +79,13 @@ func trySSHPowerOff(h *host.Host) error {
 	}
 
 	out.T(out.Shutdown, `Powering off "{{.profile_name}}" via SSH ...`, out.V{"profile_name": h.Name})
-	out, err := h.RunSSHCommand("sudo poweroff")
-	// poweroff always results in an error, since the host disconnects.
-	glog.Infof("poweroff result: out=%s, err=%v", out, err)
+	if driver.IsKIC(h.DriverName) {
```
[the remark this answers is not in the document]
medyagh: Because we don't have access to an SSH runner in this part of the code and using host.RunSSHD command didn't work

pkg/drivers/kic/kic.go
```diff
@@ -269,16 +268,23 @@ func (d *Driver) Kill() error {
...
-	cmd := exec.Command(d.NodeConfig.OCIBinary, "rm", "-f", "-v", d.MachineName)
-	o, err := cmd.CombinedOutput()
-	out := strings.TrimSpace(string(o))
-	if err != nil {
-		if strings.Contains(out, "is already in progress") {
-			log.Warnf("Docker engine is stuck. please restart docker daemon on your computer.", d.MachineName)
+
+	if err := oci.DeleteContainer(d.NodeConfig.OCIBinary, d.MachineName); err != nil {
+		if strings.Contains(err.Error(), "is already in progress") {
+			glog.Warningf("Docker engine is stuck. please restart docker daemon on your computer.")
```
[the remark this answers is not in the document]
medyagh: Sounds good

test/integration/pause_test.go
medyagh: For example, if someone see we test a functionality named
"TestDeletePaused" they know why we pause and the. Delete in one or and when we have a test report we can link to only what failed

PR #7678, opened 2020-04-15
Exits gracefully if container runtime is mispelled

### LINKED ISSUE #7426: do not crash, when missplelled container runtime
```

```

the valid options are ...

zerefwayne:
fixes #7426 

image


zerefwayne:
@afbjorklund 

I forked my branch `fix-7426` from `master`. However the file `cmd/minikube/cmd/start.go` on GitHub is different from what I have on my local machine. Hence there is a merge conflict in this Pull Request.

PR #7679, opened 2020-04-15
Exits gracefully if container runtime is mispelled

### LINKED ISSUE #7426: do not crash, when missplelled container runtime

zerefwayne:
fixes #7426 

image

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -46,6 +46,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/bootstrapper/images"
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/constants"
+	"k8s.io/minikube/pkg/minikube/cruntime"
 	"k8s.io/minikube/pkg/minikube/download"
 	"k8s.io/minikube/pkg/minikube/driver"
 	"k8s.io/minikube/pkg/minikube/exit"
@@ -222,7 +223,7 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 	k8sVersion := getKubernetesVersion(existing)
 	cc, n, err := generateClusterConfig(cmd, existing, k8sVersion, driverName)
 	if err != nil {
-		exit.UsageT("failed to generate config. error: {{.error}}", out.V{"error": err.Error()})
+		return node.Starter{}, errors.Wrap(err, "Failed to generate config")
 	}
 
 	// This is about as far as we can go without overwriting config files
@@ -808,6 +809,17 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 		validateMemorySize()
 		if !driver.HasResourceLimits(drvName) {
 			out.WarningT("The '{{.name}}' driver does not respect the --memory flag", out.V{"name": drvName})
+		}
+	}
+
+	if cmd.Flags().Changed(containerRuntime) {
+
+		runtime := viper.GetString(containerRuntime)
+
+		_, isValidRuntime := util.ValidateString(runtime, cruntime.ValidOptions())
+
+		if !isValidRuntime {
+			exit.UsageT("Invalid Container Runtime: \"{{.runtime}}\". Valid runtimes are: {{.validOptions}}", out.V{"runtime": runtime, "validOptions": strings.Join(cruntime.ValidOptions(), ", ")})
 		}
 	}
 
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -222,7 +222,7 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 	k8sVersion := getKubernetesVersion(existing)
 	cc, n, err := generateClusterConfig(cmd, existing, k8sVersion, driverName)
 	if err != nil {
-		return node.Starter{}, errors.Wrap(err, "Failed to generate config")
+		exit.UsageT("failed to generate config. error: {{.error}}", out.V{"error": err.Error()})
 	}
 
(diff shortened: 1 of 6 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -222,7 +222,7 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 	k8sVersion := getKubernetesVersion(existing)
 	cc, n, err := generateClusterConfig(cmd, existing, k8sVersion, driverName)
 	if err != nil {
-		return node.Starter{}, errors.Wrap(err, "Failed to generate config")
```
medyagh, line 225: Instead of here we can catch during, validateFlags .

pkg/minikube/cruntime/cruntime.go
```diff
@@ -47,6 +48,11 @@ func (cs ContainerState) String() string {
 	return [...]string{"all", "running", "paused"}[cs]
 }
 
+// supportedContainerRuntimes lists the supported container runtimes
+func supportedContainerRuntimes() []string {
```
medyagh, line 52: Make this public and in Cobra's flag for --contaired runtime reference it.
Change the help text for --contaired-runtime to reference this. For example: 
Valid options are cruntime.ValidOptions()

medyagh: here is an example of cobra help
https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/start_flags.go#L130

cmd/minikube/cmd/start.go
zerefwayne: https://github.com/kubernetes/minikube/blob/13895fa1daa7282d067da618f11c5dc2fcda2808/cmd/minikube/cmd/start.go#L819-L822

@medyagh If I implement the check here, I'll have to check the flag in a list of valid flags which can be cumbersome in my opinion. So, I think like the other attributes we should let this one fall onto the switch statement as well.

pkg/minikube/cruntime/cruntime.go
zerefwayne: Implemented this in the latest commit.

pkg/minikube/cruntime/cruntime.go
```diff
@@ -154,8 +158,9 @@ func New(c Config) (Manager, error) {
 			Init:              sm,
 		}, nil
 	default:
-		return nil, fmt.Errorf("unknown runtime type: %q", c.Type)
+		return nil, fmt.Errorf("invalid runtime: %q", c.Type)
```
medyagh, line 161: no need to change this. revert irrelevent changes plz

pkg/minikube/cruntime/cruntime.go
```diff
@@ -47,6 +46,11 @@ func (cs ContainerState) String() string {
 	return [...]string{"all", "running", "paused"}[cs]
 }
 
+// ValidOptions lists the supported container runtimes
+func ValidOptions() []string {
```
medyagh, line 50: plz use this in the start_flags.go so the help text uses this list to show to people what are the valid options.

cmd/minikube/cmd/start.go
medyagh: no this is not the right way to do it, to let minikube go and fail then act on failure,
we have validateFlags and we can detect there before sending minikube to the field.

please handle the validation over there.

pkg/util/utils.go
```diff
@@ -32,6 +32,16 @@ const (
 	downloadURL = "https://storage.googleapis.com/minikube/releases/%s/minikube-%s-amd64%s"
 )
 
+// ValidateString returns index, true if a string exists in []string
+func ValidateString(input string, validOptions []string) (int, bool) {
```
zerefwayne, line 36: @medyagh I've created this util so that strings can be validated in the entire project if need arises.

pkg/util/utils_test.go
```diff
@@ -22,6 +22,28 @@ import (
 	"github.com/blang/semver"
 )
 
+func TestValidateString(t *testing.T) {
```
zerefwayne, line 25: I've added the tests for the ValidateString util as well.

pkg/util/utils.go
medyagh: plz avoid using util package, we are trying to get rid of util package, define this func in the onlly place that was used. (close to where it was used)

pkg/util/utils_test.go
```diff
@@ -22,6 +22,28 @@ import (
...
+	testData := []struct {
```
medyagh, line 26: I dont undrestand why the test has to count the length of the valid options,
the test should be called, no need for doing this.

pkg/util/utils.go
medyagh: plz use a simple look up,
first make the input lowercase and then check if it is in this slice cruntime.ValidOptions()
and if it is in the slice, then return True, otherwise return False

cmd/minikube/cmd/start.go
```diff
@@ -811,6 +812,17 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 		}
 	}
 
+	if cmd.Flags().Changed(containerRuntime) {
+
+		runtime := viper.GetString(containerRuntime)
+
+		_, isValidRuntime := util.ValidateString(runtime, cruntime.ValidOptions())
```
medyagh, line 819: change this to a simple check, no need for util.
make sure to change the user input to lower case,
so if user enters Docker it should be converted and saved as "docker" or Containerd to be saved as containerd

medyagh:
@zerefwayne I haven't heard from you I will have to close this PR if don't hear from you

tstromberg:
Please re-open when you have a chance to address feedback.

Issue #7012, opened 2020-03-11
show progress bar for downloading kic base image


medyagh:
we need to treat the kic base image same as VM ISO in the UI

1- download kic base image before creating container (in the same step that we download iso for VM)
2- show progress bar for docker pull (if proress bar not possible, we should show some sort of animation so the user dont be bored and they also know they are waiting on the base image be pulled.

afbjorklund:
This would be nice to have for _all_ containers (e.g. cache), even though the kindest is the biggest...

There's some discussion upstream, in https://github.com/google/go-containerregistry/issues/670 - it adds a lot of complexity

afbjorklund:
@medyagh : are you planning to load kic from a tarball too, similar to the preload ?

Or maybe just run `docker pull` in a separate step, before doing the 'docker run` ?

medyagh:
> @medyagh : are you planning to load kic from a tarball too, similar to the preload ?
> 
> Or maybe just run `docker pull` in a separate step, before doing the 'docker run` ?

yeah I meant that wherever we download the ISO, for kic drivers we just do a docker or podman pull
and show some sort of animation
so when docker run goes, it already has the image.

afbjorklund:
As workaround, we _could_ load the kicbase image from a tarball that we download ?

`313M	gcr.io/k8s-minikube/kicbase_v0.0.9`

`docker load -i kicbase_v0.0.9.tgz`

That way, cache would also work with the non-daemon drivers (such as podman).

afbjorklund:
```
docker pull  gcr.io/k8s-minikube/kicbase:v0.0.9
docker save gcr.io/k8s-minikube/kicbase:v0.0.9 | gzip -9 -c -n > kicbase_v0.0.9.tgz
```

afbjorklund:
Unfortunately progress is only implemented for `tarball`, but not for `daemon`.

But that API only supports running Docker anyway, so it is obsolete-by-design.

```go
        if cc.Driver != "docker" {
                // TODO: driver == "podman"
                klog.Info("Driver isn't docker, skipping base image download")
                return
        }
        if image.ExistsImageInDaemon(cc.KicBaseImage) {
                klog.Infof("%s exists in daemon, skipping pull", cc.KicBaseImage)
                return
        }

```

afbjorklund:
Turns out that go-containerregistry doesn't do pull, but uses load from archive internally...

praneshvr00:
[...]

PR #7683, opened 2020-04-15
Only show pulling base image when actually pulling

afbjorklund:
Only show the weird tractor, when actually doing the download


medyagh:
> Only show the weird tractor, when actually doing the download

could u please paste the output after this PR ?

afbjorklund:
BEFORE

```
🚜  Pulling base image ...
```

AFTER

```
```


It doesn't actually pull the base image, since it already has it.
(and even when it **does** pull it, there is no output.. as in #7012)

Issue #7704, opened 2020-04-16
TestStartStop/group/containerd: busybox does not appear after restart


medyagh:
I have a feeling that this is happening because we apply the kic overlay before default service account is up

https://storage.googleapis.com/minikube-builds/logs/7611/d3db795/Docker_Linux.html#failsection

```
-- stdout --
	  - Apr 16 03:02:00 containerd-20200415T195638-32245 kubelet[1799]: E0416 03:02:00.567033    1799 reflector.go:178] object-"kube-system"/"kindnet-token-gdctk": Failed to list *v1.Secret: secrets "kindnet-token-gdctk" is forbidden: User "system:node:containerd-20200415t195638-32245" cannot list resource "secrets" in API group "" in the namespace "kube-system": no relationship found between node "containerd-20200415t195638-32245" and this object
	  - Apr 16 03:02:00 containerd-20200415T195638-32245 kubelet[1799]: E0416 03:02:00.568857    1799 reflector.go:178] object-"kube-system"/"kindnet-token-gdctk": Failed to list *v1.Secret: secrets "kindnet-token-gdctk" is forbidden: User "system:node:containerd-20200415t195638-32245" cannot list resource "secrets" in API group "" in the namespace "kube-system": no relationship found between node "containerd-20200415t195638-32245" and this object
	  - Apr 16 03:02:00 containerd-20200415T195638-32245 kubelet[1799]: E0416 03:02:00.575966    1799 reflector.go:178] object-"kube-system"/"kube-proxy": Failed to list *v1.ConfigMap: configmaps "kube-proxy" is forbidden: User "system:node:containerd-20200415t195638-32245" cannot list resource "configmaps" in API group "" in the namespace "kube-system": no relationship found between node "containerd-20200415t195638-32245" and this object
	  - Apr 16 03:02:00 containerd-20200415T195638-32245 kubelet[1799]: E0416 03:02:00.592505    1799 reflector.go:178] object-"kube-system"/"kube-proxy-token-sg9t9": Failed to list *v1.Secret: secrets "kube-proxy-token-sg9t9" is forbidden: User "system:node:containerd-20200415t195638-32245" cannot list resource "secrets" in API group "" in the namespace "kube-system": no relationship found between node "containerd-20200415t195638-32245" and this object
-- /stdout --
```

To run locally:

```
make integration -e TEST_ARGS="-test.run TestStartStop/group/containerd --profile=minikube --cleanup=false --minikube-start-args="--driver=docker""
```

[...]

Issue #7613, opened 2020-04-11
kube-proxy" is forbidden: User "system:node:minikube" cannot list resource "configmaps" in API group "" in the namespace "kube-system": no relationship found between node "minikube" and this object


veeveksen:
**Steps to reproduce the issue:** 

1.Install minikube and virtualbox.
2. execute command minikube start --force .

:~ vivek$ minikube start --force
😄  minikube v1.9.2 on Darwin 10.14.6
✨  Using the virtualbox driver based on user configuration
👍  Starting control plane node m01 in cluster minikube
🔥  Creating virtualbox VM (CPUs=2, Memory=4000MB, Disk=20000MB) ...
🐳  Preparing Kubernetes v1.18.0 on Docker 19.03.8 ...
🌟  Enabling addons: default-storageclass, storage-provisioner
❗  Enabling 'default-storageclass' returned an error: running callbacks: [Error making standard the default storage class: Error listing StorageClasses: Get "https://192.168.99.131:8443/apis/storage.k8s.io/v1/storageclasses": dial tcp 192.168.99.131:8443: i/o timeout]
❌  Problems detected in kubelet:
    Apr 11 06:13:24 minikube kubelet[3983]: E0411 06:13:24.137584    3983 reflector.go:178] object-"kube-system"/"kube-proxy-token-2j4z7": Failed to list *v1.Secret: secrets "kube-proxy-token-2j4z7" is forbidden: User "system:node:minikube" cannot list resource "secrets" in API group "" in the namespace "kube-system": no relationship found between node "minikube" and this object
    Apr 11 06:13:24 minikube kubelet[3983]: E0411 06:13:24.138044    3983 reflector.go:178] object-"kube-system"/"kube-proxy": Failed to list *v1.ConfigMap: configmaps "kube-proxy" is forbidden: User "system:node:minikube" cannot list resource "configmaps" in API group "" in the namespace "kube-system": no relationship found between node "minikube" and this object
❌  Problems detected in kubelet:
    Apr 11 06:13:24 minikube kubelet[3983]: E0411 06:13:24.137584    3983 reflector.go:178] object-"kube-system"/"kube-proxy-token-2j4z7": Failed to list *v1.Secret: secrets "kube-proxy-token-2j4z7" is forbidden: User "system:node:minikube" cannot list resource "secrets" in API group "" in the namespace "kube-system": no relationship found between node "minikube" and this object
[...]

PR #7705, opened 2020-04-16
Improve reliability of cluster role binding creation for containerd/cri-o

medyagh:
closes https://github.com/kubernetes/minikube/issues/7704

### Integration Test Flake 
To fix flakes like this:
 https://storage.googleapis.com/minikube-builds/logs/7611/d3db795/Docker_Linux.html#fail_TestStartStop%2fgroup%2fcontainerd
```
 cannot list resource "secrets" in API group "" in the namespace "kube-system": no relationship found between node "containerd-20200415t195638-32245" and this object
```

### what was happening ?
we tried to apply kic overlay network and enable addons and Elevate System Permissions in Parallel.  we need default service account before creating things on minikube.

### which drivers are affected by this PR ?
- Expected slower --wait=true for on  only Containerd and CRIO runtimes on docker/podman drivers. 

### Future PRS ?
users reported same error for when they enabled addons on virtualbox 
https://github.com/kubernetes/minikube/issues/7613 
we need to refactor enable addons, in a way that it always ensures the Default SA is created. it could be part of Bootstrapper Interface.

### DIFF
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -207,6 +207,15 @@ func (k *Bootstrapper) init(cfg config.ClusterConfig) error {
 	go func() {
 		// the overlay is required for containerd and cri-o runtime: see #7428
 		if driver.IsKIC(cfg.Driver) && cfg.KubernetesConfig.ContainerRuntime != "docker" {
+			// this is a sepcial wait only for containerd,cri-o on docker
+			// because for containerd and cri-o we need to
+			// to wait for default SA to be up to avoid #7704
+			tmpCFG := cfg // making a temp config to use for this specific wait
+			tmpCFG.VerifyComponents = map[string]bool{kverify.DefaultSAWaitKey: true}
+			glog.Infof("waiting for default sevice account before we apply kic overlay")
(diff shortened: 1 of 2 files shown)


medyagh:
Docker Test:

TestStartStop/group/old-docker
a flake cert errr:  https://github.com/kubernetes/minikube/issues/7720

TestStartStop/group/containerd
known issue, waiting for busy box : https://github.com/kubernetes/minikube/issues/7704

TestStartStop/group/crio 
```
	W0416 08:08:07.917337    9367 kubeadm.go:299] delete failed: /bin/bash -c "sudo env PATH=/var/lib/minikube/binaries/v1.15.7:$PATH kubeadm reset --force": Process exited with status 1
	stdout:
	[reset] Reading configuration from the cluster...
	[reset] FYI: You can look at this config file with 'kubectl -n kube-system get cm kubeadm-config -oyaml'
	stderr:
	W0416 15:08:07.879498   27209 reset.go:98] [reset] Unable to fetch the kubeadm-config ConfigMap from cluster: failed to get config map: Get https://172.17.0.3:8443/api/v1/namespaces/kube-system/configmaps/kubeadm-config: dial tcp 172.17.0.3:8443: connect: connection refused
	Found multiple CRI sockets, please use --cri-socket to select one: /var/run/dockershim.sock, /var/run/crio/crio.sock
```

Issue #7695, opened 2020-04-15
Host preload image on a site accessible from China


tstromberg:
See issues like #7612

medyagh:
I heard that in china they are using alibaba for the preloaded images but I dont know how they are doing that, since we have have hard coded.

I would like to get alibaba accounts for minikube so we can publish the artifats directly to alibaba ourseleves.

RA489:
We don't have any credits with Alibaba as of now, because nobody has requested them yet. But we can make it work by filing a formal request, estimate our expenses here : https://github.com/kubernetes/funding

RA489:
I created an issue https://github.com/kubernetes/funding/issues/14 for Alibaba funds.

RA489:
@medyagh It will be great if you could mention our estimates on this issue: https://github.com/kubernetes/funding/issues/14

PR #7707, opened 2020-04-16
Skip preload download if --image-repository is set

### LINKED ISSUE #7612: Skip preload download if `--image-repository` is set
**重现问题所需的命令**：minikube start --cpus=2 --memory='2000mb' --vm-driver hyperkit --image-mirror-country='cn' --image-repository='registry.cn-hangzhou.aliyuncs.com/google_containers'

**失败的命令的完整输出**：<details>
😄  minikube v1.9.2 on Darwin 10.15.4
✅  Using image repository registry.cn-hangzhou.aliyuncs.com/google_containers
💾  Downloading Kubernetes v1.18.0 preload ...
    > preloaded-images-k8s-v2-v1.18.0-docker-overlay2-amd64.tar.lz4: 131.59 MiB
</details>

**`minikube logs`命令的输出**： <details>

</details>

**使用的操作系统版本**：macOS 10.15.4

prasadkatti:
fixes #7612

### DIFF
--- a/pkg/minikube/node/cache.go
+++ b/pkg/minikube/node/cache.go
@@ -43,7 +43,7 @@ const (
 
 // BeginCacheKubernetesImages caches images required for kubernetes version in the background
 func beginCacheKubernetesImages(g *errgroup.Group, imageRepository string, k8sVersion string, cRuntime string) {
-	if download.PreloadExists(k8sVersion, cRuntime) {
+	if imageRepository == "" && download.PreloadExists(k8sVersion, cRuntime) {
 		glog.Info("Caching tarball of preloaded images")
 		err := download.Preload(k8sVersion, cRuntime)
 		if err == nil {
--- a/pkg/minikube/node/cache.go
+++ b/pkg/minikube/node/cache.go
@@ -43,6 +43,7 @@ const (
 
 // BeginCacheKubernetesImages caches images required for kubernetes version in the background
 func beginCacheKubernetesImages(g *errgroup.Group, imageRepository string, k8sVersion string, cRuntime string) {
+	// TODO: remove imageRepository check once #7695 is fixed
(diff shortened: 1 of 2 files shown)


medyagh:
I dont agree with this PR ! I think if someone is in china we should try to download the preload from github.
we should instead try to upload our preload images to other than GCS bucket but also github

pkg/minikube/node/cache.go
```diff
@@ -43,7 +43,7 @@ const (
 
 // BeginCacheKubernetesImages caches images required for kubernetes version in the background
 func beginCacheKubernetesImages(g *errgroup.Group, imageRepository string, k8sVersion string, cRuntime string) {
-	if download.PreloadExists(k8sVersion, cRuntime) {
+	if imageRepository == "" && download.PreloadExists(k8sVersion, cRuntime) {
```
tstromberg, line 46: Add:

TODO: remove imageRepository check once https://github.com/kubernetes/minikube/issues/7695 is fixed.

tstromberg:
> I dont agree with this PR ! I think if someone is in china we should try to download the preload from github.
> we should instead try to upload our preload images to other than GCS bucket but also github

Yes, and this PR is to address the short-term bleeding for minikube v1.10 - as it's unlikely that we will have images elsewhere before then. Preload leaves man users in China in a broken state, causing an influx of incoming support issues. 

Additional thought: If a user sets --image-repository, they want the images to be loaded from a particular location, preloaded or not.  Even if we added GitHub, we should still allow a way to respect user preferences.

pkg/minikube/node/cache.go
prasadkatti: Thanks for the review. Added a TODO note.

prasadkatti:
@tstromberg - Added a TODO note as a comment based on your review feedback.

Issue #7728, opened 2020-04-16
error starting a stopped kic


medyagh:
```

	main.main()
		/app/cmd/minikube/main.go:66 +0xea
	W0416 15:41:02.092730    3564 out.go:201] error provisioning host: Failed to start host: driver start: starting a stopped kic node newest-cni-20200416T153818-5303: exit status 1
	* 
	X error provisioning host: Failed to start host: driver start: starting a stopped kic node newest-cni-20200416T153818-5303: exit status 1
	* 
	* minikube is exiting due to an error. If the above message is not useful, open an issue:
```

https://storage.googleapis.com/minikube-builds/logs/7726/1b85d48/Docker_Linux.html#fail_TestStartStop%2fgroup%2fnewest-cni

PR #7729, opened 2020-04-16
improve kic start and kill

medyagh:
redoing this PR: https://github.com/kubernetes/minikube/pull/7653/files

- improve kic start/restart
- improve kic kill (I actually never seen this func be called ever but we just implement the driver) so this should not be that much of an effective change

closes https://github.com/kubernetes/minikube/issues/7728

### DIFF
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -315,9 +315,9 @@ func (d *Driver) Restart() error {
 
 // Start an already created kic container
 func (d *Driver) Start() error {
-	cr := command.NewExecRunner() // using exec runner for interacting with
+	cr := command.NewExecRunner() // using exec runner for interacting with docker/podman daemon
 	if _, err := cr.RunCmd(exec.Command(d.NodeConfig.OCIBinary, "start", d.MachineName)); err != nil {
-		return err
+		return errors.Wrap(err, "start")
 	}
 	checkRunning := func() error {
 		s, err := oci.ContainerStatus(d.NodeConfig.OCIBinary, d.MachineName)
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -261,7 +261,7 @@ func (d *Driver) Kill() error {
 	}
 
 	if err := oci.ShutDown(d.OCIBinary, d.MachineName); err != nil {
-		glog.Warningf("couldn't shutdown the container, will contineu with kill anyways: %v", err)
+		glog.Warningf("couldn't shutdown the container, will continue with kill anyways: %v", err)
 	}
 
 	cr := command.NewExecRunner() // using exec runner for interacting with dameon.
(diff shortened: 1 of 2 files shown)


pkg/drivers/kic/kic.go
```diff
@@ -258,9 +259,14 @@ func (d *Driver) Kill() error {
 	if err := sysinit.New(d.exec).ForceStop("kubelet"); err != nil {
 		glog.Warningf("couldn't force stop kubelet. will continue with kill anyways: %v", err)
 	}
-	cmd := exec.Command(d.NodeConfig.OCIBinary, "kill", d.MachineName)
-	if err := cmd.Run(); err != nil {
-		return errors.Wrapf(err, "killing kic node %s", d.MachineName)
+
+	if err := oci.ShutDown(d.OCIBinary, d.MachineName); err != nil {
+		glog.Warningf("couldn't shutdown the container, will contineu with kill anyways: %v", err)
```
priyawadhwa, line 264: nit: spelling of continue

pkg/drivers/kic/kic.go
```diff
@@ -258,9 +259,14 @@ func (d *Driver) Kill() error {
...
+	}
+
+	cr := command.NewExecRunner() // using exec runner for interacting with dameon.
+	if _, err := cr.RunCmd(exec.Command(d.NodeConfig.OCIBinary, "kill", d.MachineName)); err != nil {
+		return errors.Wrapf(err, "killing %q", d.MachineName)
```
priyawadhwa, line 269: it would be useful to include stdout/stderr in the error here

pkg/drivers/kic/kic.go
```diff
@@ -292,40 +298,43 @@ func (d *Driver) Remove() error {
 func (d *Driver) Restart() error {
 	s, err := d.GetState()
 	if err != nil {
-		return errors.Wrap(err, "get kic state")
+		glog.Warningf("get state during restart: %v", err)
```
priyawadhwa, line 301: I'd make it clear this is an error, "error getting state during restart: %v"

pkg/drivers/kic/kic.go
```diff
@@ -292,40 +298,43 @@ func (d *Driver) Remove() error {
...
-// Start a _stopped_ kic container
-// not meant to be used for Create().
+// Start an already created kic container
 func (d *Driver) Start() error {
-	s, err := d.GetState()
-	if err != nil {
-		return errors.Wrap(err, "get kic state")
+	cr := command.NewExecRunner() // using exec runner for interacting with docker/podman daemon
+	if _, err := cr.RunCmd(exec.Command(d.NodeConfig.OCIBinary, "start", d.MachineName)); err != nil {
+		return errors.Wrap(err, "start")
```
priyawadhwa, line 320: same as above, it would be useful to include stdout/stderr here

pkg/drivers/kic/kic.go
medyagh: will do

pkg/drivers/kic/kic.go
medyagh: the exec runner already includes stdout and stderr int the output.

medyagh:
test failure are related to busybox 

https://storage.googleapis.com/minikube-builds/logs/7729/891e7a6/Docker_Linux.html#fail_TestStartStop%2fgroup%2fembed-certs

PR #7740, opened 2020-04-17
Bind mount /var/tmp to avoid exhausting memory with podman

### LINKED ISSUE #7718: Podman is OOM-killed when loading images

1. `minikube start --container-runtime=crio --disk-size=100GB --memory=3900m --driver=kvm2`
1. (on minikube VM) `watch -d 'df -h'`
1. `podman save <image> | minikube ssh "sudo podman load"`
1. *Repeat*, tmpfs mounted on / fills, eventually podman will be killed inside the minikube VM

Maybe I'm misunderstanding how the CRI-O driver works, but it seems like `podman` is using `/var/lib/containers/storage` as the GraphRoot instead of `/mnt/vda1/var/lib/containers/storage`.

```
😄  minikube v1.9.2 on Ubuntu 18.04
🔥  Creating kvm2 VM (CPUs=2, Memory=3900MB, Disk=102400MB) ...
🎁  Preparing Kubernetes v1.18.0 on CRI-O 1.17.1 ...
```

**dmesg output:**
```dmesg
[  340.592651] Out of memory: Kill process 10107 (podman) score 9 or sacrifice child
[  340.592696] Killed process 10107 (podman) total-vm:830324kB, anon-rss:7212kB, file-rss:4kB, shmem-rss:28888kB
[  340.594962] oom_reaper: reaped process 10107 (podman), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
```

vixus0:
Fixes #7718 

When the CRI-O container storage backend is used, loading images using
`podman load` inside the VM creates temporary files in `/var/tmp`. As `/var/tmp` is
mounted as tmpfs, these temporary files rapidly consume the available memory
and cause processes on the minikube VM to be OOM-killed.

Instead, bind mount `/var/tmp` like we do with` /var/log`, which will write
temporary files to disk instead of memory.

### DIFF
--- a/deploy/iso/minikube-iso/package/automount/minikube-automount
+++ b/deploy/iso/minikube-iso/package/automount/minikube-automount
@@ -109,6 +109,10 @@ if [ -n "$BOOT2DOCKER_DATA" ]; then
     mkdir /var/log
     mount --bind /mnt/$PARTNAME/var/log /var/log
 
+    mkdir -p /mnt/$PARTNAME/var/tmp
+    mkdir /var/tmp
+    mount --bind /mnt/$PARTNAME/var/tmp /var/tmp
+
     mkdir -p /mnt/$PARTNAME/var/lib/kubelet
     mkdir /var/lib/kubelet
     mount --bind /mnt/$PARTNAME/var/lib/kubelet /var/lib/kubelet


vixus0:
CLA signed.

PR #7749, opened 2020-04-17
Unify Kubernetes client creation around kapi.Client

### LINKED ISSUE #7745: Enabling 'default-storageclass' returned an error: no Auth Provider found for name "gcp"
-- Reported by @etanshaul 

I see the often in the Minikube startup logs (things still start up successfully):

`! Enabling 'default-storageclass' returned an error: running callbacks: [Error getting storagev1 interface Error creating new client from kubeConfig.ClientConfig(): no Auth Provider found for name "gcp" : Error creating new client from kubeConfig.ClientConfig(): no Auth Provider found for name "gcp"]`

tstromberg:
Fixes race condition where minikube may try to apply the default storage-class to the current Kubernetes context rather than to the minikube context. 

Fixes #7745

### DIFF
--- a/pkg/minikube/service/service.go
+++ b/pkg/minikube/service/service.go
@@ -52,7 +52,7 @@ const (
 
 // K8sClient represents a kubernetes client
 type K8sClient interface {
-	GetCoreClient() (typed_core.CoreV1Interface, error)
+	GetCoreClient(string) (typed_core.CoreV1Interface, error)
 }
 
 // K8sClientGetter can get a K8sClient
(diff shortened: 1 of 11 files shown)


pkg/minikube/service/service.go
```diff
@@ -156,7 +127,8 @@ func GetServiceURLs(api libmachine.API, namespace string, t *template.Template)
 
 // GetServiceURLsForService returns a SvcUrl object for a service in a namespace. Supports optional formatting.
 func GetServiceURLsForService(api libmachine.API, namespace, service string, t *template.Template) (SvcURL, error) {
```
medyagh, line 129: would it be out of scope to pass the profile name instead of viper ?

tstromberg: Done.

Issue #7709, opened 2020-04-16
Using podman run with container runtime crio


kameshsampath:
OS:  macOS Catalina 

podman version: 
```
Client:
Version:            1.6.3-dev
RemoteAPI Version:  1
Go Version:         go1.12.10
Git Commit:         6c6e78374f5be949d11a8608080c96e2d22ca872
Built:              Wed Oct 30 00:38:11 2019
OS/Arch:            darwin/amd64

Service:
Version:            1.8.2
RemoteAPI Version:  1
Go Version:         go1.11.13
Git Commit:         028e3317eb1494b9b2acba4a0a295df80fae66cc
Built:              Fri Mar 27 00:14:28 2020
OS/Arch:            linux/amd64
``
minikube version: 

```
miniikube version: v1.9.2
commit: 93af9c1e43cab9618e301bc9fa720c63d5efa393
```

**Steps to reproduce the issue:** 

1. 
```
./minikube start -p podman-node \
      --container-runtime=cri-o \
      --network-plugin=cni \
      --enable-default-cni
```
2.  `minikube -p podman-node podman-env `
3.  `podman run -it --rm busybox /bin/sh `

The command fails with error 

```
Trying to pull docker.io/library/busybox...
Getting image source signatures
Copying blob sha256:e2334dd9fee4b77e48a8f2d793904118a3acf26f1f2e72a3d79c6cae993e07f0
Copying config sha256:be5888e67be651f1fbb59006f0fd791b44ed3fceaa6323ab4e37d5928874345a
Writing manifest to image destination
Storing signatures
time="2020-04-16T06:10:09Z" level=error msg="error configuring network namespace for container 7fddc83c2a0dd031fab5c084235ed63b73322eb269706ee2ca9c66e726fe1840: Missing CNI default network"
```


**Full output of `minikube start` command used, if not already included:**

```
 ./minikube start -p podman-node \
      --container-runtime=cri-o \
      --network-plugin=cni \
      --enable-default-cni
😄  [podman-node] minikube v1.9.2 on Darwin 10.15.4
    ▪ MINIKUBE_HOME=/Users/kameshs/MyLabs/minikube/podman/
✨  Automatically selected the hyperkit driver
💾  Downloading driver docker-machine-driver-hyperkit:
    > docker-machine-driver-hyperkit.sha256: 65 B / 65 B [---] 100.00% ? p/s 0s
[...]

PR #7754, opened 2020-04-17
Add default CNI network for running wth podman

afbjorklund:
Forgot to add this after #6664

For #7709

### DIFF
--- a/deploy/iso/minikube-iso/package/podman/podman.mk
+++ b/deploy/iso/minikube-iso/package/podman/podman.mk
@@ -29,6 +29,8 @@ endef
 
 define PODMAN_INSTALL_TARGET_CMDS
 	$(INSTALL) -Dm755 $(@D)/bin/podman $(TARGET_DIR)/usr/bin/podman
+	$(INSTALL) -d -m 755 $(TARGET_DIR)/etc/cni/net.d/
+	$(INSTALL) -m 644 cni/87-podman-bridge.conflist $(TARGET_DIR)/etc/cni/net.d/87-podman-bridge.conflist
 endef
 
 $(eval $(generic-package))


deploy/iso/minikube-iso/package/podman/podman.mk
```diff
@@ -29,6 +29,8 @@ endef
 
 define PODMAN_INSTALL_TARGET_CMDS
 	$(INSTALL) -Dm755 $(@D)/bin/podman $(TARGET_DIR)/usr/bin/podman
+	$(INSTALL) -d -m 755 $(TARGET_DIR)/etc/cni/net.d/
+	$(INSTALL) -m 644 cni/87-podman-bridge.conflist $(TARGET_DIR)/etc/cni/net.d/87-podman-bridge.conflist
```
medyagh, line 33: is this something we could add to ubuntu image too ? kicbase?

afbjorklund: As far as I know, it is already included (provided by `podman` package):

```console
docker@minikube:~$ dpkg -S /etc/cni/net.d/87-podman-bridge.conflist 
podman: /etc/cni/net.d/87-podman-bridge.conflist
```

If you do a `sudo make install` in libpod, it will be included by default.

afbjorklund: Seems to work OK there, with the provided CNI:

```console
docker@minikube:~$ sudo podman run -it busybox
Trying to pull docker.io/library/busybox...
Getting image source signatures
Copying blob e2334dd9fee4 done  
Copying config be5888e67b done  
Writing manifest to image destination
Storing signatures
/ # 
```

So this was only a missing file on the minikube VM

PR #7758, opened 2020-04-18
WIP: fix multinode docker by downgrading kicbase

sharifelgamal:
This is half of an experiment to see which kicbase is the best solution for fixing multinode docker. We already know 0.0.9 won't, but 0.0.8 and 0.0.10 will.

The upgrade version is #7757


medyagh:
I added multi node test to the head, and they are passing , I dont think this PR is needed anymore

Issue #7770, opened 2020-04-18
hanndle cgroup freezer state error


medyagh:
great investigative work by @govargo 
we found out our service command was always failing in github actions, because we check if the STDERR is empty, turns out we had and errorf for from cgroup:

```
// example error from github action:
// cat: /sys/fs/cgroup/freezer/actions_job/e62ef4349cc5a70f4b49f8a150ace391da6ad6df27073c83ecc03dbf81fde1ce/kubepods/burstable/poda1de58db0ce81d19df7999f6808def1b/5df53230fe3483fd65f341923f18a477fda92ae9cd71061168130ef164fe479c/freezer.state: No such file or directory\n"
```

I found where this error retruns but I have no idea how we should fix except changing `t.Errorf`.

https://github.com/kubernetes/minikube/blob/2c4ba8ba9cf23d9d229e46f29fbb121070aedf14/pkg/minikube/bootstrapper/bsutil/kverify/api_server.go#L167-L171

_Originally posted by @govargo in https://github.com/kubernetes/minikube/issues/7765#issuecomment-615911155_

medyagh:
cc: @tstromberg

PR #7769, opened 2020-04-18
Change log for ServiceCmd to avoid github action(docker_ubuntu_16_04) fail

govargo:
### What type of PR is this?
/kind flake
/area testing

### What this PR does / why we need it:

In e2e test TestFunctional/parallel/ServiceCmd, `minikube service` test always fails in GitHub Actions.
This PR change testing log func which make to avoid cgroups error.

### Which issue(s) this PR fixes:
Fixes #7765 

### Does this PR introduce a user-facing change?

No.

### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:
```
NONE
```

PR #7771, opened 2020-04-18
change loglevel for api freezer if not found

medyagh:
a follow up PR for the great investigative work by @govargo https://github.com/kubernetes/minikube/pull/7769 

my bad I merged a PR too soon by mistake. before @govargo have time to see my review comment.

this will does not close https://github.com/kubernetes/minikube/issues/7770
but it will help to still test for stderr to be empty without failing on Not-critical.

so this will help us not flood the stderr for the user, if that does not affect the functionality of minikube. (and in this case we the test passes)

### DIFF
--- a/pkg/minikube/bootstrapper/bsutil/kverify/api_server.go
+++ b/pkg/minikube/bootstrapper/bsutil/kverify/api_server.go
@@ -166,7 +166,15 @@ func APIServerStatus(cr command.Runner, hostname string, port int) (state.State,
 
 	rr, err = cr.RunCmd(exec.Command("sudo", "cat", path.Join("/sys/fs/cgroup/freezer", fparts[2], "freezer.state")))
 	if err != nil {
-		glog.Errorf("unable to get freezer state: %s", rr.Stderr.String())
+		// example error from github action:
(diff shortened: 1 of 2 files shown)


pkg/minikube/bootstrapper/bsutil/kverify/api_server.go
```diff
@@ -166,7 +166,15 @@ func APIServerStatus(cr command.Runner, hostname string, port int) (state.State,
 
 	rr, err = cr.RunCmd(exec.Command("sudo", "cat", path.Join("/sys/fs/cgroup/freezer", fparts[2], "freezer.state")))
 	if err != nil {
-		glog.Errorf("unable to get freezer state: %s", rr.Stderr.String())
+		// example error from github action:
+		// cat: /sys/fs/cgroup/freezer/actions_job/e62ef4349cc5a70f4b49f8a150ace391da6ad6df27073c83ecc03dbf81fde1ce/kubepods/burstable/poda1de58db0ce81d19df7999f6808def1b/5df53230fe3483fd65f341923f18a477fda92ae9cd71061168130ef164fe479c/freezer.state: No such file or directory\n"*
+		// TODO: #7770 investigate how to handle this error better.
+		if strings.Contains(rr.Stderr.String(), "freezer.state: No such file or directory\n") {
+			glog.Infof("unable to get freezer state (might be okay and be related to #770): %s", rr.Stderr.String())
```
tstromberg, line 173: Getting a code review on something like this would be nice.

- Please do not parse error strings if possible. 
- The info statement is wrong
- I think just a general Warningf is fine here.

Please note that this error message is highly indicative of a system where pause/unpause may not work.

Issue #7368, opened 2020-04-02
satisify kubelet/kubeadm warnnings


medyagh:
currently kubeadm warns about a few things that we could do small things to make it happy so our logs look less scary

prasadkatti:
Are you talking about these logs? -
```
Apr 03 05:13:01 minikube kubelet[2294]: Flag --authorization-mode has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kube
let-config-file/ for more information.
Apr 03 05:13:01 minikube kubelet[2294]: Flag --cgroup-driver has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-c
onfig-file/ for more information.
Apr 03 05:13:01 minikube kubelet[2294]: Flag --client-ca-file has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-
config-file/ for more information.
Apr 03 05:13:01 minikube kubelet[2294]: Flag --cluster-domain has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-
config-file/ for more information.
Apr 03 05:13:01 minikube kubelet[2294]: Flag --fail-swap-on has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-co
nfig-file/ for more information.
Apr 03 05:13:01 minikube kubelet[2294]: Flag --pod-manifest-path has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubel
et-config-file/ for more information.
```

afbjorklund:
@medyagh : Do you have any examples ot scary kubeadm logs ?

I get this output:
```
[...]

PR #7786, opened 2020-04-19
Enable docker and kubelet units after adding them

afbjorklund:
This avoids warnings from `kubeadm init` later:

`! 	[WARNING Service-Docker]`: docker service is not enabled, please run 'systemctl enable docker.service'
`! 	[WARNING Service-Kubelet]`: kubelet service is not enabled, please run 'systemctl enable kubelet.service'

For #7368


afbjorklund:
I don't think the minikube VM supports a reboot anyway (have to use `minikube start` for boot)...

But it should cut down on the number of warnings from kubeadm, about them not being enabled.

medyagh:
I am curious how this PR will affect the start numbers.

Like to see performance bot results.


It also needs lint

Issue #7670, opened 2020-04-15
clean up after creating temp dirs in uni tests 


medyagh:
some unit tests are making temp files but they never delete them, I fixed a few of them but there are more of these

```
ioutil.TempDir(""
```

would be a good PR for anyone interested to make the code better

here is an example how to do it: https://github.com/kubernetes/minikube/pull/7669/files

the tricky for the left overs is, there are helper funcs in the unit tests that multple tests call them

the person who takes this PR should make sure there is it is not cleaned up too early before tests use them.

zerefwayne:
@medyagh I would like to take up this issue.

BLasan:
@medyagh Is this issue still opened?

medyagh:
@BLasan  I have not seen this issue recently, if there is evidence we leave left over files I would accept a PR for this.

PR #7787, opened 2020-04-19
clean up temp dirs in unittests

alonyb:
This PR is for #7670

Clean up after creating temporary directories in unit tests. With validation for tests that call a single directory multiple times.


cmd/minikube/cmd/config/get_test.go
```diff
@@ -21,15 +21,17 @@ import (
...
 
 func TestGetNotFound(t *testing.T) {
-	createTestConfig(t)
+	createTestConfig(t, "create")
 	_, err := Get("nonexistent")
 	if err == nil || err.Error() != "specified key could not be found in config" {
 		t.Fatalf("Get did not return error for unknown property")
 	}
+	// remove all temp dir
+	createTestConfig(t, "remove")
```
tstromberg, line 30: I like this PR - but have some suggestions to make this more idiomatic Go:

* have `createTestConfig` return either the path generated, or a cleanup function
* immediately after `createTestConfig` is called, call: `defer removeTestConfig(path)`

This way cleanup is guaranteed, and we omit the stuttering function signature of `create..*create`

alonyb: @tstromberg That's a pretty good idea, I implemented it.

pkg/minikube/tunnel/registry_test.go
```diff
@@ -251,6 +251,12 @@ func tmpFile(t *testing.T) string {
 	if err != nil {
 		t.Errorf("failed to create temp file %s", err)
 	}
+	defer func() { //clean up tempdir
```
medyagh, line 254: this is wrong, because the temp file is deleted when the tempFile func is finished, and the called wont have the chance to this this temp file

pkg/addons/addons_test.go
```diff
@@ -28,12 +29,20 @@ import (
...
 )
 
-func createTestProfile(t *testing.T) string {
+func createTestProfile(t *testing.T, status string) string {
 	t.Helper()
 	td, err := ioutil.TempDir("", "profile")
 	if err != nil {
 		t.Fatalf("tempdir: %v", err)
 	}
+	if strings.EqualFold(status, "remove") {
```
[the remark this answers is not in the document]
alonyb: yeah I made the change.
And no, all files are being removed.

pkg/minikube/tunnel/registry_test.go
alonyb: If you are correct, also each function already deletes those files, it is not necessary to use that `defer`

pkg/addons/addons_test.go
```diff
@@ -119,6 +128,9 @@ func TestSetAndSave(t *testing.T) {
 	if c.Addons["dashboard"] != false {
 		t.Errorf("expected dashboard to be enabled")
 	}
+
+	// remove all temp dir
+	createTestProfile(t, false)
```
sharifelgamal, line 133: should this be a remove call?

alonyb: @sharifelgamal yes, that is a remove call, to remove all tmp directories/files.
That's at the end, so they won't be removed before the rest of the unittests use them.

cmd/minikube/cmd/config/set_test.go
```diff
@@ -81,4 +86,13 @@ func createTestConfig(t *testing.T) {
...
 		t.Fatalf("error creating temporary profiles directory: %+v", err)
 	}
+
+	return td
+}
+
+func removeTempDir(t *testing.T, path string) {
+	err := os.RemoveAll(path)
+	if err != nil {
+		t.Errorf("failed to creal up temp folder %q", path)
```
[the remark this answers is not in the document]
alonyb: @priyawadhwa thanks for the observation, I didn't see that.

pkg/addons/addons_test.go
```diff
@@ -28,12 +28,21 @@ import (
 	"k8s.io/minikube/pkg/minikube/localpath"
 )
 
-func createTestProfile(t *testing.T) string {
+func createTestProfile(t *testing.T, status bool) string {
```
priyawadhwa, line 31: Adding the `defer os.RemoveAll(td)` to this function means that the temporary directory that we create at the start of the function will be removed as soon as `createTestProfile` finishes executing -- so right now   this function both creates and deletes the temp home directory

We probably want this function to return the home directory, and `defer os.RemoveAll(td)` in the tests that call this function instead

```go
func createTestProfile(t *testing.T) (string, string)
```

tstromberg:
Any update?

sunny-b:
I'm available to finish up this PR is @alonyb is not able to.

PR #7790, opened 2020-04-20
hyperv: Use 'Get-Wmiobject HyperVisorPresent' instead of 'Get-WindowsOptionalFeature'

### LINKED ISSUE #7789: Make Hyper-V usable by members of Hyper-V admins group
This would require moving from `Get-WindowsOptionalFeature` to `HypervisorPresent`, as mentioned here:

https://github.com/code-ready/crc/blob/0f91698f533fd4f4079f8cb5b61f951e91e1b95e/pkg/crc/preflight/preflight_checks_windows.go#L44

@gbraad  mentioned that perhaps we should just call all of the checks here before trying to use the Hyper-V driver.

regulskimichal:
May fix #7789

### DIFF
--- a/pkg/minikube/registry/drvs/hyperv/hyperv.go
+++ b/pkg/minikube/registry/drvs/hyperv/hyperv.go
@@ -89,9 +89,9 @@ func status() registry.State {
 	ctx, cancel := context.WithTimeout(context.Background(), 8*time.Second)
 	defer cancel()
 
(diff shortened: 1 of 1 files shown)


regulskimichal:
The CLA has been signed

tstromberg:
PR tests fine locally. 

There are still a few restrictions with the hyperv libmachine driver that require PowerShell to be called from an Administrative context, but this check keeps minikube from blocking, and appears to be much faster than the previous check.

Thank you!

Issue #7719, opened 2020-04-16
TestErrorSpam: retry after the token is automatically created and added to the service account


tstromberg:
This is two separate errors:

`* Restarting the docker service may improve performance`

This line of stderr should be whitelisted.

and:

`error_spam_test.go:49: unexpected error log in stderr: "Error from server (ServerTimeout): error when creating \"/etc/kubernetes/addons/storage-provisioner.yaml\": No API token found for service account \"storage-provisioner\", retry after the token is automatically created and added to the service account"`

This is indicative of a real problem.

_Originally posted by @tstromberg in https://github.com/kubernetes/minikube/issues/7601#issuecomment-614775410_

tstromberg:
This particular flake seems fairly rare.

```
~/bin/bughunt.sh /var/lib/jenkins/jobs/docker_Linux_integration "unexpected error log in stderr:.*Error from server"
Tests in last 90 days
---------------------
-rw-r--r-- 1 jenkins jenkins  752043 Apr 16 12:37 /var/lib/jenkins/jobs/docker_Linux_integration/builds/1683/Docker_Linux.txt
-rw-r--r-- 1 jenkins jenkins 1534674 Apr 15 19:01 /var/lib/jenkins/jobs/docker_Linux_integration/builds/1656/Docker_Linux.txt
-rw-r--r-- 1 jenkins jenkins  679678 Apr 10 15:50 /var/lib/jenkins/jobs/docker_Linux_integration/builds/1562/Docker_Linux.txt

Frequency per day
-----------------
      1 2020-04-10
      1 2020-04-15
      1 2020-04-16

Oldest 5 matches
---------------------

==> /var/lib/jenkins/jobs/docker_Linux_integration/builds/1683/Docker_Linux.txt <==
 GitHub pull request #7723 of commit 39e41297c888f2f02d52f107cd373a78b242d7f4, no merge conflicts.
>> Starting at Thu Apr 16 12:08:10 PDT 2020
minikube version: v1.9.2
commit: bd5d8d386913a5f25b31cb888d59352180b1e31b

==> /var/lib/jenkins/jobs/docker_Linux_integration/builds/1656/Docker_Linux.txt <==
 GitHub pull request #7611 of commit 4656a4e07e9cebf86828769118f061134d25f306, no merge conflicts.
>> Starting at Wed Apr 15 18:44:29 PDT 2020
minikube version: v1.9.2
commit: 10950d43b2aca43640426a17dce7a1fa568f579f

[...]

PR #7815, opened 2020-04-21
Add option to force docker to use systemd as cgroup manager

### LINKED ISSUE #6404: Make minikube work in Google Cloud Shell (using kic)

priyawadhwa:
Since minikube is running systemd, kubeadm expects kubeadm to be the cgroup manager. If docker is using a different cgroup manager like cgroupfs, this can cause unstable resource allocation. We were seeing this in Cloud Shell, and forcing docker to use systemd resolved the issue.

This k8s doc provides good context: https://kubernetes.io/docs/setup/production-environment/container-runtimes

I modified the kic base image to include the docker daemon config which forces systemd. I upgraded kic to v0.0.10. This file didn't exist before, so it isn't being overwritten.

I elected to go with `--force-systemd` instead of `--cgroup-manager[cgroupfs,systemd]` since the latter would introduce more complexity than needed. Crio defaults to systemd while docker/containerd default to cgroupfs, and we'd need to keep track of acceptable combos with the latter option.

Fixes #6404


pkg/minikube/cruntime/docker.go
```diff
@@ -365,6 +365,25 @@ func dockerImagesPreloaded(runner command.Runner, images []string) bool {
 	return true
 }
 
+// ForceSystemdCgroupManager forces docker to use systemd as cgroup manager, as recommended in k8s docs:
+// https://kubernetes.io/docs/setup/production-environment/container-runtimes/#docker
+func (r *Docker) ForceSystemdCgroupManager() error {
```
medyagh, line 370: how about we do this during the installation ? in kicbase ?
so we don't have to restart docker and also copy config files?

medyagh: https://github.com/kubernetes/minikube/blob/master/hack/images/kicbase.Dockerfile#L12

pkg/minikube/cruntime/docker.go
```diff
@@ -365,6 +365,25 @@ func dockerImagesPreloaded(runner command.Runner, images []string) bool {
...
+	daemonConfig := `{
+"exec-opts": ["native.cgroupdriver=systemd"],
```
afbjorklund, line 372: it would be nicer if this function _only_ changed the native.cgroupdriver setting ?

instead of deleting and overwriting the entire docker daemon.json configuration file

We had lots of issues like this with CRI-O before, when we replaced the crio.conf
The Docker config hasn't change in ages, but who knows. Maybe a new driver ?

afbjorklund:
Docker has some weird reason for **not** using this by default:

```
# the default is not to use systemd for cgroups because the delegate issues still
# exists and systemd currently does not support the cgroup feature set required
# for containers run by docker
ExecStart=/usr/bin/dockerd -H fd:// --containerd=/run/containerd/containerd.sock
```

But like you say, Kubernetes recommends using the same as host.

> Control groups are used to constrain resources that are allocated to processes. A single cgroup manager will simplify the view of what resources are being allocated and will by default have a more consistent view of the available and in-use resources.

pkg/minikube/cruntime/docker.go
priyawadhwa: So it looks like the daemon.json file doesn't exist at all currently. I switched to including it in the base image, instead of modifying it while minikube is starting.

Let's see how the integration tests respond, but I locally tested on my Macbook, my Linux workstation, and in Cloud Shell. It seems to be working normally on all three.

priyawadhwa:
@afbjorklund yah there's definitely a disparity. Since the goal here is to run k8s, I feel like we should stick to what they are recommending, especially if tests pass.

If we start to see issues around this, we could put it behind a flag so that only users who need it can use it.

pkg/minikube/cruntime/docker.go
priyawadhwa: cool, done

priyawadhwa:
It looks like the github actions environment doesn't like this -- 

```
functional_test_docker_ubuntu : completed with 20 / 33 failures in 27 min 44 seconds
----------------20 Failures----------------------------
[
  "TestFunctional/serial/StartWithProxy",
  "TestFunctional/serial/SoftStart",
  "TestFunctional/serial/KubeContext",
  "TestFunctional/serial/KubectlGetPods",
  "TestFunctional/serial/CacheCmd/cache/verify_cache_inside_node",
  "TestFunctional/serial/CacheCmd/cache/cache_reload",
  "TestFunctional/serial/MinikubeKubectlCmd",
  "TestFunctional/parallel/ComponentHealth",
  "TestFunctional/parallel/DashboardCmd",
  "TestFunctional/parallel/DNS",
  "TestFunctional/parallel/StatusCmd",
  "TestFunctional/parallel/LogsCmd",
  "TestFunctional/parallel/MountCmd",
  "TestFunctional/parallel/ServiceCmd",
  "TestFunctional/parallel/TunnelCmd",
  "TestFunctional/parallel/MySQL",
  "TestFunctional/parallel/UpdateContextCmd",
  "TestFunctional/parallel/DockerEnv",
  "TestFunctional/parallel/NodeLabels"
]
-------------------------------------------------------
*** 13 Passed ***
*** 20 Failed ***
```

I'm still not totally sure what's causing some environments to only work with systemd (Cloud Shell) and some to totally fail with systemd (Github Actions).

I'm thinking of putting this behind a flag for now, WDYT @afbjorklund @medyagh ?

afbjorklund:
We already noticed that GitHub Actions is running a totally different `docker`

i.e. they have their own fork of `moby`, rather than using the regular binary

afbjorklund:
Ouch, I just noticed there is a podman and crio upgrade in here as well...

hack/images/kicbase.Dockerfile
```diff
@@ -15,17 +15,26 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 # install podman
-RUN apt-get install -y --no-install-recommends podman=1.8.2~144
+ENV PODMAN_VERSION=1.9.0~2
+RUN apt-get install -y --no-install-recommends podman=${PODMAN_VERSION}
+
 # disable non-docker runtimes by default
 RUN systemctl disable containerd && systemctl disable crio && rm /etc/crictl.yaml
+
+# copy over docker daemon config which sets cgroup manager to systemd
+COPY hack/images/daemon.json /etc/docker/daemon.json
```
medyagh, line 34: just guessing in the dark but maybe put a system restart or daemon-reload here ?

priyawadhwa:
>Ouch, I just noticed there is a podman and crio upgrade in here as well...

@afbjorklund yah the Dockerfile build is failing with "version cannot be found" error for both crio and podman, so they need to be updated for this image to be built :/

hack/images/kicbase.Dockerfile
priyawadhwa: I think the image is OK. the issue seems to be with specific docker environments -- this seems to work fine on my local machines and passes most of the `Docker_Linux` tests. But, it fails completely in the github actions environment.

medyagh:
KVM tests look very scary https://storage.googleapis.com/minikube-builds/logs/7815/52b5737/KVM_Linux.html
 but it could be just a bad CI machine, I invoked a retest. lets double check the KVM results.

priyawadhwa:
@medyagh the KVM failures are normal now

medyagh:
@priyawadhwa 
the docker driver tests are abnormally high and seems to be affecting the multi node,https://storage.googleapis.com/minikube-builds/logs/7815/71ca99b/Docker_Linux.html

could it be because we need to make the docker on all the nodes use systemd ?

priyawadhwa:
@medyagh this seems to be a coincidence. It also doesn't seem like the tests reran, so I'll merge w/ master and we can rerun them

medyagh:
> @medyagh this seems to be a coincidence. It also doesn't seem like the tests reran, so I'll merge w/ master and we can rerun them

sounds good. please make sure there is a comment explaining every failure associated with an issue.

priyawadhwa:
[Docker_Linux Logs](https://storage.googleapis.com/minikube-builds/logs/7815/e3397d5/Docker_Linux.html#fail_TestStartStop%2fgroup%2fcrio)

**TestErrorSpam - 198.53s**
Same error as https://github.com/kubernetes/minikube/issues/7719

**TestPreload - 214.58s**

**TestStartStop/group/containerd - 273.53s**
Seen in https://github.com/kubernetes/minikube/issues/7704

**TestStartStop/group/crio - 1000.61s**
Same as https://storage.googleapis.com/minikube-builds/logs/7815/e3397d5/Docker_Linux.html#fail_TestStartStop%2fgroup%2fcrio

**TestStartStop/group/embed-certs - 844.77s**
```
start_stop_delete_test.go:141: (dbg) TestStartStop/group/embed-certs: waiting 9m0s for pods matching "k8s-app=kubernetes-dashboard" in namespace "kubernetes-dashboard" ...
start_stop_delete_test.go:141: ***** TestStartStop/group/embed-certs: pod "k8s-app=kubernetes-dashboard" failed to start within 9m0s: timed out waiting for the condition ****
```

**TestFunctional/parallel/StatusCmd - 5.23s**
```
functional_test.go:313: (dbg) Non-zero exit: out/minikube-linux-amd64 -p functional-20200427175241-5551 status: exit status 2 (2.43335847s)
-- stdout --
	functional-20200427175241-5551
	type: Control Plane
	host: Running
	kubelet: Running
	apiserver: Error
	kubeconfig: Configured
	
-- /stdout --
functional_test.go:315: failed to run minikube status. args "out/minikube-linux-amd64 -p functional-20200427175241-5551 status" : exit status 2
```

**TestMultiNode/serial/DeleteNode - 5.54s**
```
multinode_test.go:174: (dbg) Non-zero exit: out/minikube-linux-amd64 -p multinode-20200427175733-5551 node delete m03: exit status 69 (4.220365448s)
-- stdout --
	* This control plane is not running! (state=Error)
	  - To fix this, run: "minikube start -p multinode-20200427175733-5551"
```

PR #7856, opened 2020-04-22
Add --base-image flag for docker driver local dev

priyawadhwa:
Pass in the base image you want to use for the docker/podman drivers

Now,

```
minikube start --driver docker --docker-base-image kic
```

works if kic is an image in the local daemon

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -26,6 +26,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"k8s.io/minikube/pkg/drivers/kic"
 	"k8s.io/minikube/pkg/minikube/bootstrapper/bsutil"
 	"k8s.io/minikube/pkg/minikube/bootstrapper/bsutil/kverify"
 	"k8s.io/minikube/pkg/minikube/config"
@@ -98,6 +99,7 @@ const (
 	nodes                   = "nodes"
 	preload                 = "preload"
 	deleteOnFailure         = "delete-on-failure"
+	dockerBaseImage         = "docker-base-image"
 )
 
 // initMinikubeFlags includes commandline flags for minikube.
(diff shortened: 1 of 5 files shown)


cmd/minikube/cmd/start_flags.go
```diff
@@ -98,6 +99,7 @@ const (
 	nodes                   = "nodes"
 	preload                 = "preload"
 	deleteOnFailure         = "delete-on-failure"
+	dockerBaseImage         = "docker-base-image"
```
medyagh, line 102: change this to "base-image" since this is for both docker or podman.

site/content/en/docs/commands/start.md
```diff
@@ -36,6 +36,7 @@ minikube start [flags]
       --disk-size string                  Disk size allocated to the minikube VM (format: <number>[<unit>], where unit = b, k, m or g). (default "20000mb")
       --dns-domain string                 The cluster dns domain name used in the kubernetes cluster (default "cluster.local")
       --dns-proxy                         Enable proxy for NAT DNS requests (virtualbox driver only)
+      --docker-base-image string          The base image to use for docker/podman drivers. Intended for local development. (default "gcr.io/k8s-minikube/kicbase:v0.0.9@sha256:82a826cc03c3e59ead5969b8020ca138de98f366c1907293df91fc57205dbb53")
```
medyagh, line 39: does this accept values without sha ? because local images dont have repoTag Sha
(I think it will be fine but would be good if u confirm it works for passing without sha)

priyawadhwa: as stated in the PR description, this works without sha

cmd/minikube/cmd/start_flags.go
priyawadhwa: done

pkg/minikube/node/cache.go
```diff
@@ -102,11 +101,12 @@ func doCacheBinaries(k8sVersion string) error {
 // BeginDownloadKicArtifacts downloads the kic image + preload tarball, returns true if preload is available
 func beginDownloadKicArtifacts(g *errgroup.Group) {
 	glog.Info("Beginning downloading kic artifacts")
-	if !image.ExistsImageInDaemon(kic.BaseImage) {
+	baseImage := viper.GetString("base-image")
```
medyagh, line 104: just an optional nit: this string is in danger of if we change the string, it will be left unseen. but optional.

priyawadhwa: sure, i'll address that in my wip PR

afbjorklund:
Sounds good, before you had to hack the code in order to use the "snapshot" images.

PR #7857, opened 2020-04-22
wip: Update integration tests to use a locall built version of kic

priyawadhwa:
Depends on #7856 to merge first

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -18,6 +18,7 @@ jobs:
       run : |
         make minikube-linux-amd64
         make e2e-linux-amd64
+        make kic-base-image
         cp -r test/integration/testdata ./out
         whoami
         echo github ref $GITHUB_REF
@@ -120,7 +121,7 @@ jobs:
         sudo ln -s /etc/apparmor.d/usr.sbin.mysqld /etc/apparmor.d/disable/
         sudo apparmor_parser -R /etc/apparmor.d/usr.sbin.mysqld
         START_TIME=$(date -u +%s)
-        KUBECONFIG=$(pwd)/testhome/kubeconfig MINIKUBE_HOME=$(pwd)/testhome ./e2e-linux-amd64 -minikube-start-args=--vm-driver=docker  -test.run TestFunctional -test.timeout=30m -test.v -timeout-multiplier=1.5 -binary=./minikube-linux-amd64 2>&1 | tee ./report/testout.txt
+        KUBECONFIG=$(pwd)/testhome/kubeconfig MINIKUBE_HOME=$(pwd)/testhome ./e2e-linux-amd64 -minikube-start-args="--driver=docker --docker-base-image=kic"  -test.run TestFunctional -test.timeout=30m -test.v -timeout-multiplier=1.5 -binary=./minikube-linux-amd64 2>&1 | tee ./report/testout.txt
         END_TIME=$(date -u +%s)
         TIME_ELAPSED=$(($END_TIME-$START_TIME))
         min=$((${TIME_ELAPSED}/60))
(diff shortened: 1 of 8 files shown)


cmd/minikube/cmd/start_flags.go
```diff
@@ -98,6 +99,7 @@ const (
 	nodes                   = "nodes"
 	preload                 = "preload"
 	deleteOnFailure         = "delete-on-failure"
+	dockerBaseImage         = "docker-base-image"
```
medyagh, line 102: I would change this to base-image, since this is not docker specific, and could be podman too

hack/jenkins/common.sh
```diff
@@ -283,9 +283,12 @@ set -x
 if test -f "${TEST_OUT}"; then
   rm "${TEST_OUT}" || true # clean up previous runs of same build
 fi
+
+make kic-base-image 
+
 touch "${TEST_OUT}"
 ${SUDO_PREFIX}${E2E_BIN} \
-  -minikube-start-args="--driver=${VM_DRIVER} ${EXTRA_START_ARGS}" \
+  -minikube-start-args="--driver=${VM_DRIVER} --docker-base-image=kic ${EXTRA_START_ARGS}" \
```
medyagh, line 291: should be added to our github actions too
I recommend adding a new job in the 3 jobs groups that are running before integraiton tests
currently they are build_minikube, unit_test, lint
we should add another job called -build build-base image-
and make the follwing integraiton jobs adds
needs: [build-minikube , build-base-image-

Makefile
```diff
@@ -529,7 +529,9 @@ storage-provisioner-image: out/storage-provisioner-$(GOARCH) ## Build storage-pr
 .PHONY: kic-base-image
 kic-base-image: ## builds the base image used for kic.
 	docker rmi -f $(REGISTRY)/kicbase:$(KIC_VERSION)-snapshot || true
-	docker build -f ./hack/images/kicbase.Dockerfile -t $(REGISTRY)/kicbase:$(KIC_VERSION)-snapshot  --build-arg COMMIT_SHA=${VERSION}-$(COMMIT) --target base .
+	docker rmi -f kic || true
```
medyagh, line 532: where do we build an image called "kic" ?

priyawadhwa: in the next line of the makefile

.github/workflows/pr.yml
```diff
@@ -18,6 +18,8 @@ jobs:
       run : |
         make minikube-linux-amd64
         make e2e-linux-amd64
+        make kic-base-image
```
medyagh, line 21: we should do this step only if the file changed, because rebuilding the same image, might not work because kind-base image has not pinned down the ubuntu version and each time we get some packages missing.

.github/workflows/pr.yml
```diff
@@ -113,14 +115,15 @@ jobs:
...
+        docker load -i kic.tar
         mkdir -p report
         mkdir -p testhome
         chmod a+x e2e-*
         chmod a+x minikube-*
         sudo ln -s /etc/apparmor.d/usr.sbin.mysqld /etc/apparmor.d/disable/
         sudo apparmor_parser -R /etc/apparmor.d/usr.sbin.mysqld
         START_TIME=$(date -u +%s)
-        KUBECONFIG=$(pwd)/testhome/kubeconfig MINIKUBE_HOME=$(pwd)/testhome ./e2e-linux-amd64 -minikube-start-args=--vm-driver=docker  -test.run TestFunctional -test.timeout=30m -test.v -timeout-multiplier=1.5 -binary=./minikube-linux-amd64 2>&1 | tee ./report/testout.txt
+        KUBECONFIG=$(pwd)/testhome/kubeconfig MINIKUBE_HOME=$(pwd)/testhome ./e2e-linux-amd64 -minikube-start-args="--driver=docker --docker-base-image=kic"  -test.run TestFunctional -test.timeout=30m -test.v -timeout-multiplier=1.5 -binary=./minikube-linux-amd64 2>&1 | tee ./report/testout.txt
```
medyagh, line 126: same here, if docker image didn't change we should use the current base image, and avoid passing the build image here.

.github/workflows/pr.yml
priyawadhwa: I added the `--cache-from` flag to #7858 so if the Dockefile isn't changed, it'll reuse all the layers from the current image.

.github/workflows/pr.yml
priyawadhwa: resolved as mentioned above

.github/workflows/pr.yml
```diff
@@ -20,6 +20,8 @@ jobs:
         run: |
           make minikube-linux-amd64
           make e2e-linux-amd64
+          make kic-base-image
```
medyagh, line 23: we should only do this only if the Dockerfile been changed. if not it can cause all sort of issues. currently our base ubuntu image is not pinned and each time we get a different installed versions. not good idea for a reliable idea of testing what we ship to the users.

PR #7858, opened 2020-04-22
Upgrade docker driver base image to v0.0.10

priyawadhwa:
The dockerfile currently doesn't build because it can't find the specified versions of crio and podman to install.

### DIFF
--- a/hack/images/kicbase.Dockerfile
+++ b/hack/images/kicbase.Dockerfile
@@ -17,15 +17,17 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && rm /etc/crictl.yaml
 
 # install cri-o based on https://github.com/cri-o/cri-o/commit/96b0c34b31a9fc181e46d7d8e34fb8ee6c4dc4e1#diff-04c6e90faac2675aa89e2176d2eec7d8R128
-ARG CRIO_VERSION="1.17=1.17.3~2"
 RUN sh -c "echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_19.10/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list" && \    
     curl -LO https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/xUbuntu_19.10/Release.key && \
     apt-key add - < Release.key && apt-get update && \
-    apt-get install -y --no-install-recommends cri-o-${CRIO_VERSION}
+    apt-get install -y --no-install-recommends cri-o-1.17
 
 # install podman
-ARG PODMAN_VERSION=1.9.0~2
-RUN apt-get install -y --no-install-recommends podman=${PODMAN_VERSION}
+RUN apt-get install -y --no-install-recommends podman
+
+# install varlink
+RUN apt-get install -y --no-install-recommends varlink
+
 
 # disable non-docker runtimes by default
 RUN systemctl disable containerd && systemctl disable crio && rm /etc/crictl.yaml
--- a/hack/images/kicbase.Dockerfile
+++ b/hack/images/kicbase.Dockerfile
@@ -17,14 +17,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && rm /etc/crictl.yaml
 
 # install cri-o based on https://github.com/cri-o/cri-o/commit/96b0c34b31a9fc181e46d7d8e34fb8ee6c4dc4e1#diff-04c6e90faac2675aa89e2176d2eec7d8R128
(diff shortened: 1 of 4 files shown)


priyawadhwa:
**Failing Tests**

**TestStartStop/group/containerd - 533.86s**
```
	X failed to start node: startup failed: Wait failed: waiting for apps_running: checking k8s-apps to be running: timed out waiting for the condition
```

coredns pod:
```
failed to setup network for sandbox "ef187e8c7fe723cb6292b6b5c5ed43e49b1d073b37e758224516a813c4725368": failed to set bridge addr: could not add IP address to "cni0": permission denied
```

**TestStartStop/group/crio - 1175.54s**
```
start_stop_delete_test.go:138: ***** TestStartStop/group/crio: pod "integration-test=busybox" failed to start within 7m0s: timed out waiting for the condition ****
```
**TestStartStop/group/embed-certs - 724.03s**
coredns:
```
Warning  Unhealthy       7m54s              kubelet, embed-certs-20200422151018-5649  Readiness probe failed: HTTP probe failed with statuscode: 503
```

afbjorklund:
> The dockerfile currently doesn't build because it can't find the specified versions of crio and podman to install.

This will happen again, neither Ubuntu nor Kubic makes **any** guarantees to keep old packages...

If you want to keep "pinning" packages, then we need to create a cache of all the .deb used:

lz4=1.9.1-1
    gnupg=2.2.12-1ubuntu3
    sudo=1.8.27-1ubuntu4.1
    docker.io=19.03.2-0ubuntu1
    openssh-server=1:8.0p1-6build1
    dnsutils=1:9.11.5.P4+dfsg-5.1ubuntu2.1
    libglib2.0-0=2.62.1-1

cri-o-1.17=1.17.3~2

podman=1.9.0~2

Including all implicit dependencies, in case there are version specifics in those as well ?

Instead we should probably either use ranges, or _remove_ the version-release pinning.

hack/images/kicbase.Dockerfile
```diff
@@ -16,12 +16,16 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     libglib2.0-0=2.62.1-1 \
     && rm /etc/crictl.yaml
 # install cri-o based on https://github.com/cri-o/cri-o/commit/96b0c34b31a9fc181e46d7d8e34fb8ee6c4dc4e1#diff-04c6e90faac2675aa89e2176d2eec7d8R128
+ENV CRIO_VERSION="1.17=1.17.3~2"
```
afbjorklund, line 19: Why do we need these exported in the runtime environment ? Are they used somewhere else ?

hack/images/kicbase.Dockerfile
```diff
@@ -16,12 +16,16 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 RUN sh -c "echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_19.10/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list" && \    
     curl -LO https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/xUbuntu_19.10/Release.key && \
     apt-key add - < Release.key && apt-get update && \
-    apt-get install -y --no-install-recommends cri-o-1.17=1.17.2~1
+    apt-get install -y --no-install-recommends cri-o-${CRIO_VERSION}
+
 # install podman
-RUN apt-get install -y --no-install-recommends podman=1.8.2~144
+ENV PODMAN_VERSION=1.9.0~2
```
afbjorklund, line 26: Same question as $CRIO_VERSION, why is this exported ?

hack/images/kicbase.Dockerfile
priyawadhwa: I was inspired by the Skaffold dockerfile, which makes it really easy to upgrade versions by using variables:

https://github.com/GoogleContainerTools/skaffold/blob/master/deploy/skaffold/Dockerfile.deps#L17

But you're right that these don't need to be in the runtime env; it should be ARG instead of ENV

priyawadhwa:
@afbjorklund that's a really good point. I'm not sure what the best solution here is -- for now, I've added the `--cache-from` flag to the docker build command for kic. This way, layers from the current kic image will be reused as much as possible, so unless someone changes the versions of crio/podman manually we should have the same exact layers

This isn't a great solution because if the order of Dockerfile commands changes the cache will probably be invalidated, so I think we should definitely keep thinking about this

afbjorklund:
@priyawadhwa @medyagh : we need to add the "varlink" package to the kicbase, for podman-env

Whatever version is available from the Kubic repository is fine, so just add it after "podman" ?

priyawadhwa:
@afbjorklund thanks for the heads up, added.

PR #7862, opened 2020-04-23
clean up: refactor oci package exec calls

medyagh:
##### After this PR 
- all oci.Exec calls will go through runCmd(), similar to our command.Runner 
- we will have standard output errors across all minikube
- standard logging how long commands took across all minikube 
- delete unused code Copy func
- make more funcs private
- remove WarnSlow and embed functionality into the runCmd

##### Question: why not ommand.ExecRunner ?
because there would be a circular dependency, the only code that was copied from ExecRunner is the RunResult. oci package is meant to be a lib to others and implements its own minimalistic execrRunner. which will be in 'cli_runner' file.

### Follow Up PR
, to make it easy to review and to make this PR as small as possible I did not change any functionality but only made all exec calls to go through runner in follow up PR i will make the runner to return different Error Types based on the output of the cli .

such as
ErrDockerNotRunning
ErrDockerUserPermission
ErrDockerInternalError
ErrDockerStuck (means needs restart)
ErrDockerSlow
ErrDiskFull

...
...


pkg/drivers/kic/oci/cli_runner.go
```diff
@@ -0,0 +1,148 @@
...
+	"io"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/golang/glog"
+	"k8s.io/minikube/pkg/minikube/out"
+)
+
+var cli = newRunner()
```
tstromberg, line 32: Everything in this PR seems OK except for making this a global variable. It does not feel like idiomatic Go to do so.
cli. I think you could easily fix this by pulling `RunCmd` out of `cliRunner`.

medyagh: done

pkg/drivers/kic/oci/oci.go
```diff
@@ -453,12 +394,12 @@ func withPortMappings(portMappings []PortMapping) createOpt {
 }
 
 // ListContainersByLabel returns all the container names with a specified label
-func ListContainersByLabel(ociBinary string, label string) ([]string, error) {
-	stdout, err := WarnIfSlow(ociBinary, "ps", "-a", "--filter", fmt.Sprintf("label=%s", label), "--format", "{{.Names}}")
+func ListContainersByLabel(ociBin string, label string, warnSlow ...bool) ([]string, error) {
+	rr, err := runCmd(exec.Command(ociBin, "ps", "-a", "--filter", fmt.Sprintf("label=%s", label), "--format", "{{.Names}}"), true)
```
priyawadhwa, line 398: Instead of true, shouldn't warnSlow be passed in here?

pkg/drivers/kic/oci/oci.go
```diff
@@ -489,9 +430,10 @@ func PointToHostDockerDaemon() error {
 }
 
 // ContainerStatus returns status of a container running,exited,...
-func ContainerStatus(ociBin string, name string) (state.State, error) {
-	out, err := WarnIfSlow(ociBin, "inspect", name, "--format={{.State.Status}}")
-	o := strings.TrimSpace(string(out))
+func ContainerStatus(ociBin string, name string, warnSlow ...bool) (state.State, error) {
+	cmd := exec.Command(ociBin, "inspect", name, "--format={{.State.Status}}")
+	rr, err := runCmd(cmd, true)
```
priyawadhwa, line 435: same as above

pkg/drivers/kic/oci/volumes.go
```diff
@@ -40,7 +40,7 @@ func DeleteAllVolumesByLabel(ociBin string, label string) []error {
 	}
 
 	for _, v := range vs {
-		if _, err := WarnIfSlow(ociBin, "volume", "rm", "--force", v); err != nil {
+		if _, err := runCmd(exec.Command(ociBin, "volume", "rm", "--force", v), true); err != nil {
```
priyawadhwa, line 43: same as above

pkg/drivers/kic/oci/volumes.go
```diff
@@ -51,11 +51,11 @@ func DeleteAllVolumesByLabel(ociBin string, label string) []error {
...
 // if there is no volume to delete it will return nil
 // example: docker volume prune -f --filter label=name.minikube.sigs.k8s.io=minikube
-func PruneAllVolumesByLabel(ociBin string, label string) []error {
+func PruneAllVolumesByLabel(ociBin string, label string, warnSlow ...bool) []error {
 	var deleteErrs []error
 	glog.Infof("trying to prune all %s volumes with label %s", ociBin, label)
-
-	if _, err := WarnIfSlow(ociBin, "volume", "prune", "-f", "--filter", "label="+label); err != nil {
+	cmd := exec.Command(ociBin, "volume", "prune", "-f", "--filter", "label="+label)
+	if _, err := runCmd(cmd, true); err != nil {
```
priyawadhwa, line 58: same here

medyagh: done

pkg/drivers/kic/oci/volumes.go
medyagh: done

pkg/drivers/kic/oci/oci.go
medyagh: done

pkg/drivers/kic/oci/oci.go
medyagh: thanks for the good catch ! done

PR #7863, opened 2020-04-23
Drop duplicate import

tklauser:
The k8s.io/minikube/pkg/util package is imported twice, in one case
aliased to "pkgutil". Most users use this alias, so convert the single
user of the unaliased pacakge to use "pkgutil" as well and drop the
duplicate import.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -60,7 +60,6 @@ import (
 	"k8s.io/minikube/pkg/minikube/registry"
 	"k8s.io/minikube/pkg/minikube/translate"
 	"k8s.io/minikube/pkg/util"
-	pkgutil "k8s.io/minikube/pkg/util"
 	"k8s.io/minikube/pkg/version"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -822,7 +821,7 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 		}
 
 		// conntrack is required starting with kubernetes 1.18, include the release candidates for completion
-		version, _ := util.ParseKubernetesVersion(getKubernetesVersion(nil))
+		version, _ := pkgutil.ParseKubernetesVersion(getKubernetesVersion(nil))
```
medyagh, line 824: I would say keep the util and get rid of the pkgutil

tklauser:
> Thanks for noticing this!
> 
> Do you mind just searching and replacing `pkgutil` for `util` here? Our code base tries to follow the guidance from https://github.com/golang/go/wiki/CodeReviewComments when possible:
> 
> "Avoid renaming imports except to avoid a name collision"

Changed as suggested, thanks.

cmd/minikube/cmd/start.go
tklauser: Change as suggested, thanks.

PR #7864, opened 2020-04-23
addon registry-alias: change hosts update container image

### LINKED ISSUE #7715: registry aliases hosts update fails, `diff: command not found`
The hosts file update routine in registry aliases uses a recent Fedora image (`registry.fedoraproject.org/fedora`) that doesn't include a `diff` binary. This causes the hosts file update to fail and in turn causes the host environment to resolve registry names via DNS instead of finding aliases in the hosts file.

I've worked around this for now by modifying the [`registry-aliases-hosts-update`](https://github.com/kubernetes/minikube/blob/v1.9.2/deploy/addons/registry-aliases/node-etc-hosts-update.tmpl#L20) daemon set to use ubuntu instead of fedora.

1. `minikube start`
2. `minikube addons enable registry`
3. `minikube addons enable registry-aliases`
4. `kubectl -n kube-system logs registry-aliases-hosts-update-XXXXX -c update`

The relevant output is the log from the update container in the host update pod:

```
10.111.151.247	example.org
10.111.151.247	example.com
10.111.151.247	test.com
10.111.151.247	test.org
bash: line 7: diff: command not found
Done.
```

tmehlinger:
The most recent Fedora image doesn't include a `diff` binary. This changes the container image to use Ubuntu Bionic instead, which _does_ provide a diff binary.

Fixes #7715.

### DIFF
--- a/deploy/addons/registry-aliases/node-etc-hosts-update.tmpl
+++ b/deploy/addons/registry-aliases/node-etc-hosts-update.tmpl
@@ -17,7 +17,7 @@ spec:
     spec:
       initContainers:
         - name: update
-          image: ubuntu:bionic
+          image: alpine:3.11
           volumeMounts:
             - name: etchosts
               mountPath: /host-etc/hosts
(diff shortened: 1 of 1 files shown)


tmehlinger:
Bump for CLA check.

deploy/addons/registry-aliases/node-etc-hosts-update.tmpl
```diff
@@ -17,7 +17,7 @@ spec:
     spec:
       initContainers:
         - name: update
-          image: registry.fedoraproject.org/fedora
+          image: ubuntu:bionic
```
medyagh, line 20: why ubuntu ? could it be something smaller ?

tmehlinger: No reason in particular, that was just my workaround. I've changed it to Alpine 3.11 and fixed the diff command to account for the different between Busybox and GNU.

tstromberg:
Thank you!

PR #7865, opened 2020-04-23
Convert PVC storage path to /var/tmp

### LINKED ISSUE #7511: Driver: none host path should be `/var/tmp` for PVCs
Sorry in advance but the issue template really does not apply at all to my issue.

## Abstract
When using the none driver in a linux environment I get issues with PVCs after restarting. This is due to the `/tmp` directory being a `tmpfs` filesystem.

## Details
Many mainstream linux distributions adopted systemd as the init system quite a while ago, and under this system, the `/tmp` directory is managed as a mount service and is mounted as a `tmpfs` filesystem. This is ideal for performance since it uses memory for anything in that directory. My issue is that when I have a pod that uses a persistent volume claim and I restart my machine the directory backed by the host is cleared (since it was in memory). This causes the pod to not start back up when I start minikube. I can get around this by disabling the `tmp.mount` systemd service (and therefore let it default to the disk in my machine) and adding some rules to not clean up the `hostpath-provisioner` directory inside `/tmp` but this is highly undesirable since many other applications use the directory and benefit from `tmpfs`. I don't want to have to do that just so PVCs will work in minikube.

If you see [this article](https://systemd.io/TEMPORARY_DIRECTORIES/) describing the tmp system interaction with systemd they recommend using `/var/tmp` for temporary storage that is not in memory and hence can be configured to remain after reboots. I think the minikube binary should be modified to use this directory instead. This will solve my issue with PVCs and allow me to keep `tmp` as `tmpfs`.

I'm willing to provide a PR, granted the maintainers understand my problem and agree with my proposed solution. Otherwise, if there are other ideas on how to resolve this I'd love to know and contribute any way I can. Cheers!

jradtilbrook:
Fixes #7511

This changes the mount volume of the storage provisioner to `/var/tmp` on the host system.
On modern linux systems this is a normal mount point, as compared to `/tmp` which is generally a `tmpfs` which is removed upon reboots so persistant volume claims are not actually persistent.

### DIFF
--- a/deploy/addons/storage-provisioner/storage-provisioner.yaml.tmpl
+++ b/deploy/addons/storage-provisioner/storage-provisioner.yaml.tmpl
@@ -60,5 +60,5 @@ spec:
   volumes:
   - name: tmp
     hostPath:
-      path: /tmp
+      path: /var/tmp
       type: Directory
--- a/deploy/iso/minikube-iso/package/automount/minikube-automount
+++ b/deploy/iso/minikube-iso/package/automount/minikube-automount
@@ -122,12 +122,12 @@ if [ -n "$BOOT2DOCKER_DATA" ]; then
     mount --bind /mnt/$PARTNAME/data /data
 
     mkdir -p /mnt/$PARTNAME/hostpath_pv
-    mkdir /tmp/hostpath_pv
-    mount --bind /mnt/$PARTNAME/hostpath_pv /tmp/hostpath_pv
+    mkdir /var/tmp/hostpath_pv
+    mount --bind /mnt/$PARTNAME/hostpath_pv /var/tmp/hostpath_pv
 
     mkdir -p /mnt/$PARTNAME/hostpath-provisioner
-    mkdir /tmp/hostpath-provisioner
-    mount --bind /mnt/$PARTNAME/hostpath-provisioner /tmp/hostpath-provisioner
+    mkdir /var/tmp/hostpath-provisioner
+    mount --bind /mnt/$PARTNAME/hostpath-provisioner /var/tmp/hostpath-provisioner
 
     if [ -e "/userdata.tar" ]; then
         mv /userdata.tar /var/lib/boot2docker/


jradtilbrook:
The CLA has been signed

tstromberg:
need any help with the CLA?

jradtilbrook:
> need any help with the CLA?

Sorry @tstromberg it wasn't working and I forgot to come back to it. I believe I've done it now

jradtilbrook:
The logs are so verbose I'm not quite sure what the error is. If you can provide any more detail that would be much appreciated! :pray:

tstromberg:
KVM had the folowing failures that did not seem related:

```
TestErrorSpam - 369.91s
TestFunctional/parallel/ServiceCmd - 83.45s
TestStartStop/group/crio - 780.94s
```

Likewise docker had:

```
TestStartStop/group/old-k8s-version - 867.57s
TestStartStop/group/containerd - 356.23s
TestStartStop/group/crio - 930.88s
TestStartStop/group/embed-certs - 794.46s
TestFunctional/parallel/ServiceCmd - 59.56s
TestPause/serial/SecondStartNoReset - 47.51s
```

Our integration tests are pretty flaky at the moment. They all seem unrelated, but I'll re-run the tests to see if any of these are consistent.

jradtilbrook:
@tstromberg Is there anything I can do to assist?

tstromberg:
I'm going to have to give this some thought on how to roll this out:

If a user upgrades their minikube, but have an older cluster running - they will end up with a new storage provisioner running on an old ISO. How should we we handle that?

afbjorklund:
This also needs to be synchronized with the kicbase implementation.

As well as documented properly, for current users of the "none" drivers.

jradtilbrook:
@sharifelgamal I noticed you removed the stale label. I'm wondering if I should just close this as stale. I don't have time to work on it and have fixed my own issue with a bind mount on my host manually for now

medyagh:
> @sharifelgamal I noticed you removed the stale label. I'm wondering if I should just close this as stale. I don't have time to work on it and have fixed my own issue with a bind mount on my host manually for now

@jradtilbrook 
do u mind sharing the steps u did manually to fix ?

jradtilbrook:
> @jradtilbrook
> do u mind sharing the steps u did manually to fix ?

@medyagh I enabled a systemd mount unit that bind mounts the directory onto `/var/tmp`. You can see the unit file at https://github.com/jradtilbrook/dotfiles/blob/master/files/etc/systemd/system/tmp-hostpath%5Cx2dprovisioner.mount

tstromberg:
Closing by request until author has a chance to revisit this.

Issue #7651, opened 2020-04-13
Document that systemd is no longer a hard requirement for minikube


prasadkatti:
Based on @medyagh's comment [here](https://github.com/kubernetes/minikube/pull/7649#discussion_r407692805), there have been some feature additions to support init systems other than systemd. The docs specify in multiple places that a systemd based init system is required for minikube. This issue is to fix that in docs.

tstromberg:
Sent PR out, but it needs some work.

PR #7881, opened 2020-04-24
none docs: Mention OpenRC and Docker requirement

tstromberg:
Related: #7877 #7651

### DIFF
--- a/site/content/en/docs/drivers/includes/none_usage.inc
+++ b/site/content/en/docs/drivers/includes/none_usage.inc
@@ -1,8 +1,11 @@
 ## Requirements
 
-VM running a systemd-based Linux distribution ([see #2704](https://github.com/kubernetes/minikube/issues/2704))
+A Linux VM with the following:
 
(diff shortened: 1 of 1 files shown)


site/content/en/docs/drivers/includes/none_usage.inc
```diff
@@ -1,8 +1,11 @@
 ## Requirements
 
-VM running a systemd-based Linux distribution ([see #2704](https://github.com/kubernetes/minikube/issues/2704))
+A Linux VM with the following:
 
-kubeadm requirements listed here - https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/
+* Docker
+* systemd (OpenRC based systems are also supported in v1.10+)
```
medyagh, line 6: how about just say systemd or OpenRC. 
to clearify u can add example of distros with OpenRC that u have tesed it on

tstromberg: I don't want to just say OpenRC -- since support for it is not available in the current stable release.

tstromberg: ... also, I used Debian for OpenRC.

Issue #6795, opened 2020-02-25
Fedora 31 vm-driver=podman fail to start trying to start docker service


thobianchi:
**The exact command to reproduce the issue**:
minikube --vm-driver=podman start

**The full output of the command that failed**:<details>
[root@thomas-work]~# minikube --vm-driver=podman start
😄  minikube v1.7.3 on Fedora 31
✨  Using the podman (experimental) driver based on user configuration
🔥  Creating Kubernetes in podman container with (CPUs=2), Memory=2000MB (15719MB available) ...

💣  Unable to start VM. Please investigate and run 'minikube delete' if possible: creating host: create: provisioning: ssh command error:
command : sudo systemctl -f restart docker
err     : Process exited with status 1
output  : Job for docker.service failed because the control process exited with error code.
See "systemctl status docker.service" and "journalctl -xe" for details.

😿  minikube is exiting due to an error. If the above message is not useful, open an issue:
👉  https://github.com/kubernetes/minikube/issues/new/choose

</details>

**The operating system version**:

Fedora 31 : 5.5.5-200.fc31.x86_64
minikube version: v1.7.3
commit: 436667c819c324e35d7e839f8116b968a2d0a3ff
podman version 1.8.0
Selinux on permissive

It seems that even with podman driver, minikube is trying to restart docker service.

afbjorklund:
> It seems that even with podman driver, minikube is trying to restart docker service.

The driver uses podman to create a fake VM, in this node there is another container runtime...

medyagh:
even though the driver is podman, the run-time is docker,

we recently added support for cri-o runtime (will be in next release)

I think it would be a good default behavior to set cri-o as default run time for podman. since they are built together.

tstromberg:
Could you share the output of `minikube logs`?

thobianchi:
In the meanwhile I updated to 1.9.2

Output of: `sudo minikube  --vm-driver=podman start`
```
😄  minikube v1.9.2 on Fedora 31
✨  Using the podman (experimental) driver based on existing profile
👍  Starting control plane node m01 in cluster minikube
🚜  Pulling base image ...
E0416 09:44:49.897829    8747 cache.go:114] Error downloading kic artifacts:  error loading image: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
[...]

PR #7882, opened 2020-04-24
Fix sysctl fs.protected_regular=1 typo

tstromberg:
Mentioned in #6795


afbjorklund:
Yeah, it was recommended the bug reproducer rather than the fix...

Issue #7788, opened 2020-04-19
move off kindbase image


medyagh:
current issues

- the ubuntu version is not pinned down in kind base image, which causes re-building kic-base image impossible for example, the specific version of podman was not avialble anymore after re-building same kic base image.

- kind image entry point casues slow down on execs. 
- kind image comes with a preinstalled containerd, which is different version than our VM containerd, causing troubleshooting between drivers be more mystrious

Implementation:
- we need to change the base form kindest to ubuntu and pin down the exact ubuntu image build date
- create our own entrypoint to run systemctl inside an ubuntu image and only run the things that we really need.

afbjorklund:
It would be nice with some more details here...

Like: why do we want to pin the package versions ?

Why is the containerd version different (from kind)

They do settle on a **major** version, currently 19.10

----

It _would_ be very nice if we could share the ugly systemd hacks and such with KIND, if possible:

> This image needs to do a number of unusual things to support running systemd, 
> nested containers, and Kubernetes. All of what we do and why we do it 
> is documented inline in the Dockerfile.
> 
> If you make any changes to this image, please continue to document exactly 
> why we do what we do, citing upstream documentation where possible.

But otherwise I suppose we _could_ fork the Dockerfile + files, and patch / merge them as necessary.

afbjorklund:
It would be great if the scope of this one could be cut, to only do a OS version upgrade.

afbjorklund:
> It would be great if the scope of this one could be cut, to "only" do a OS version upgrade.

Added as #8250

afbjorklund:
Had some nice discussion with the `kind` folks, about co-operating on the podman driver

Don't see any reason why we should **not** also try to cooperate on the `base` image.
Maybe we need to convince them to move containerd to a intermediate layer, but still...

Currently kind has a "base" image based on ubuntu, that is then inherited by a "node" image.

* https://kind.sigs.k8s.io/docs/design/base-image

* https://kind.sigs.k8s.io/docs/design/node-image

> kind image entry point casues slow down on execs.

I guess this issue needs to be broken down further ?

https://github.com/kubernetes-sigs/kind/blob/master/images/base/files/usr/local/bin/entrypoint

afbjorklund:
Here are all the files needed for the kindbase:

https://github.com/kubernetes/minikube/compare/master...afbjorklund:kindbase

Enables you to rebuild it locally, as needed:

```
[...]

Issue #6942, opened 2020-03-07
Make custom image for KIC similar to minikube ISO


afbjorklund:
Currently we are using the KIND base image for the KIC driver:

https://github.com/kubernetes/minikube/blob/master/hack/images/kicbase.Dockerfile
```Dockerfile
# which is an ubuntu 19.10 with an entry-point that helps running systemd
# could be changed to any debian that can run systemd
FROM kindest/base:v20200122-2dfe64b2 as base
```

This image (docker.io/kindest/base) is in turn based on Ubuntu:

https://github.com/kubernetes-sigs/kind/blob/master/images/base/Dockerfile
```Dockerfile
# start from ubuntu 19.10, this image is reasonably small as a starting point
# for a kubernetes node image, it doesn't contain much we don't need
FROM ubuntu:19.10
```

As with the other base images, this one starts from a rootfs tarball:

https://hub.docker.com/_/ubuntu

```Dockerfile
FROM scratch
ADD ubuntu-eoan-core-cloudimg-amd64-root.tar.gz /
```

See https://docs.docker.com/develop/develop-images/baseimages/

----

We might want to investigate using the _same_ Linux that we are using for minikube.iso,
a custom distribution built using Buildroot which also includes systemd and the runtimes.

https://github.com/kubernetes/minikube/tree/master/deploy/iso/minikube-iso

The main difference would be that the regular ISO also includes a kernel (currently 4.19),
while the container image uses the host kernel (so it doesn't need to waste all that space)

Links:

* https://buildroot.org/downloads/manual/manual.html
* [blog.docker.com/2013/06/create-light-weight-docker-containers-buildroot](https://web.archive.org/web/20140804031955/blog.docker.com/2013/06/create-light-weight-docker-containers-buildroot)

There are lots of other small tricks needed, in order to make a running container image.
Including a lot of workaround and hacks, to be able to run `systemd` in a a container...

As per above, you can see several of these in the original Ubuntu image (vs Ubuntu ISO),
as well as from the KIND and KIC projects respectively. Some of these will _need_ to be added.

https://github.com/kubernetes-sigs/kind/blob/master/images/base/files/usr/local/bin/entrypoint

afbjorklund:
ping @medyagh

medyagh:
This sounds like a good idea to do. 

I support to have an option to run kic drivers with buildroot image in docker.
at least as an alternative, that users could choose. 
same way that users choose their base ISO_URL
[...]

PR #7884, opened 2020-04-24
wip: add experimental kicbase image  ubuntu20:04

alonyb:
## After this PR:

- image size is the same than older
- combine layers to 9 in 4 steps, older had 20 layers in 16 steps, you can see the detail [here](https://gist.github.com/alonyb/b9856b5264a9a33cf6c4e7c745081f65)
- use latest ubuntu 20.04 (minimal ubuntu focal image)
- more secure by closing 22 on the default expose


medyagh:
please move all your changes to a new folder called /hack/images/kic-base-experimental

then we will experiment with this new image, and meanwhile keep our current image, and over time we will switch to it, after we have done enough testing.

medyagh:
I like to see integration test results on this PR, preferably on your machine or your fork. without touching minikube test code.

afbjorklund:
We should split this in several steps, instead of doing all-at-once (even if that is how we test it...)

> use latest ubuntu 20.04 (minimal ubuntu focal image)

We can time this with the upgrade to Buildroot 2020.02

> more secure by closing 22 on the default expose

This one would be good to include in the next release

I can't see how `minikube ssh` would work without a port, but we **should** drop the root changes...

afbjorklund:
Aside from the ubuntu upgrade, it will be interesting to see how the layers affect the size...

Currently it is those "middle" layers that are building the size (ubuntu itself is only 73 MB)

`docker images gcr.io/k8s-minikube/kicbase:v0.0.10`

```
REPOSITORY                    TAG                 IMAGE ID            CREATED             SIZE
gcr.io/k8s-minikube/kicbase   v0.0.10             e6bc41c39dc4        2 days ago          974MB
```

* **974MB** (unpacked)

`docker history gcr.io/k8s-minikube/kicbase:v0.0.10`

```
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
<missing>           2 days ago          /bin/sh -c echo '%sudo ALL=(ALL) NOPASSWD:AL…   784B                
<missing>           2 days ago          /bin/sh -c adduser docker sudo                  2.22kB              
<missing>           2 days ago          /bin/sh -c adduser --ingroup docker --disabl…   10kB                
<missing>           2 days ago          /bin/sh -c sed -ri 's/UsePAM yes/#UsePAM yes…   3.24kB              
<missing>           2 days ago          /bin/sh -c sed -ri 's/^#?PermitRootLogin\s+.…   3.23kB              
<missing>           2 days ago          /bin/sh -c echo 'root:root' |chpasswd           744B                
<missing>           2 days ago          /bin/sh -c systemctl enable docker              34B                 
<missing>           2 days ago          /bin/sh -c apt-get install -y --no-install-r…   573kB               
<missing>           2 days ago          /bin/sh -c apt-get install -y --no-install-r…   158MB               
<missing>           2 days ago          /bin/sh -c sh -c "echo 'deb http://download.…   106MB               
<missing>           8 days ago          /bin/sh -c apt-get update && apt-get install…   425MB               
<missing>           6 weeks ago         RUN |3 CONTAINERD_VERSION=v1.3.3-14-g449e926…   210MB               buildkit.dockerfile.v0
<missing>           6 weeks ago         COPY files/ / # buildkit                        10.5kB              buildkit.dockerfile.v0
```

- 158MB (podman)
- 106MB (cri-o-1.17)
- 425MB (gnupg sudo docker.io openssh-server dnsutils)
- 210MB (KIND, including containerd, cni and crictl)

Cleaning up the apt cache for each step should help a bit...

tstromberg:
Do you mind helping me understand the tangible difference of forking these files? I worry because a lot of thought was put into these files, and they do see a fair bit of maintenance.

* I see SSH mentioned, but I don't see the change nor understand the attack vector
* I know folks on exotic platforms like ppcl64e have mentioned a wish to fork so long as kind isn't willing to accept build requests for their platform, but I'm not yet convinced it means we need to maintain our own files.
* I've heard mention about maintaining our own containerd/docker versions in the past, but this PR doesn't address this.
* I see layers mentioned, but I'm not sure it means much to an end user.

It's worth mentioning that kind has since also moved on to Ubuntu 20.04. 

Thanks!

afbjorklund:
This needs a squashing of commits, and a rebase towards upstream (kindbase)...

Should also sync Ubuntu 20.04 a bit with the upgraded files for Buildroot 2020.02 ?
For now you can see some of these changes under https://github.com/kubernetes/minikube/issues/6942#issuecomment-632352957 (no real PR yet)

I think the SSH comments was about the combination of "expose 22" and "chpasswd"

The buildroot-based image is currently 529MB, compared with 974MB for (old) kicbase
But I haven't looked at the new sizes, after squashing some of the layers as mentioned.

> I see layers mentioned, but I'm not sure it means much to an end user.

Docker layers have the same issue as git commits, if you add big files and then remove them it ends up keeping them around in the "history" which means a bigger download for the end user.

If you instead clean up and remove things in the _same_ commit layer, only the delta gets added. This leads to a smaller git repository or a smaller docker image, which means faster pull/load

afbjorklund:
Added PR #8251 with the just the ubuntu upgrade (that was not mentioned at all in #7788)

I don't think that pinning the package versions is a good idea, so I just removed them all.

afbjorklund:
> Do you mind helping me understand the tangible difference of forking these files? I worry because a lot of thought was put into these files, and they do see a fair bit of maintenance.

There doesn't seem to be any actual _changes_ done to the entrypoint, besides whitespace:

```diff
--- ../kind/images/base/files/usr/local/bin/entrypoint	2020-05-22 15:57:25.315972920 +0200
+++ hack/images/kic-base-experimental/files/entrypoint	2020-05-22 15:56:02.498304340 +0200
@@ -14,12 +14,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# using entrypoint created by kind https://github.com/kubernetes-sigs/kind/blob/master/images/base/files/usr/local/bin/entrypoint
+
 set -o errexit
 set -o nounset
 set -o pipefail
 
 fix_mount() {
-  echo 'INFO: ensuring we can execute /bin/mount even with userns-remap' 
+  echo 'INFO: ensuring we can execute /bin/mount even with userns-remap'
   # necessary only when userns-remap is enabled on the host, but harmless
   # The binary /bin/mount should be owned by root and have the setuid bit
   chown root:root /bin/mount
@@ -193,4 +195,4 @@
 enable_network_magic
 
 # we want the command (expected to be systemd) to be PID1, so exec to it
-exec "$@"
+exec "$@"
\ No newline at end of file
```

But since everything is now forked and pinned, we now miss all upstream improvements...

i.e. we **don't** get the new packages from ubuntu

and we **don't** get the changes from kind entrypoint

hack/images/kic-base-experimental/kicbase.Dockerfile
```diff
@@ -0,0 +1,146 @@
...
+    && chmod +x /usr/local/bin/entrypoint \
+ && echo "Installing Packages ..." \
+    && DEBIAN_FRONTEND=noninteractive apt-get update && \
+      apt-get install -y --no-install-recommends \
+      systemd=${SYSTEMD_VERSION} conntrack=${CONNTRACK_VERSION} iptables=${IPTABLES_VERSION} iproute2=${IPROUTE2_VERSION} ethtool=${ETHTOOL_VERSION} \
+      socat=${SOCAT_VERSION} util-linux=${UTIL_LINUX_VERSION} mount=${MOUNT_VERSION} ebtables=${EBTABLES_VERSION} udev=${UDEV_VERSION} kmod=${KMOD_VERSION} \
+      gnupg=${GNUPG_VERSION} libglib2.0-0=${LIBGLIB2_VERSION} libseccomp2=${LIBSECCOMP2_VERSION} ca-certificates=${CA_CERTIFICATES_VERSION} \
+      curl=${CURL_VERSION} rsync=${RSYNC_VERSION} \
+      && rm -rf /var/lib/apt/lists/* \
+    && sh -c "echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_19.10/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list" && \
```
afbjorklund, line 58: This still uses 19.10, rather than 20.04

tstromberg:
Based on this discussion, for now, I think we should just upgrade the underlying kicbase.

medyagh:
> Based on this discussion, for now, I think we should just upgrade the underlying kicbase.

based on coversations in office hours, minikube maintainers reached out to kind to ask for a base image that does not have nightly build containerd, and kind is meant for testing lastest source code and not meant for local kuberentes experience.

and also based on the cloud shell bug we ended up sed-ing the entrypoint by line number  in the dockerfile
https://github.com/kubernetes/minikube/pull/8465

this is an un-maintable approach, and we will need to start from scratch and mantain a fork of the entrypoint in the source code. therefore I re-open this PR

afbjorklund:
> and also based on the cloud shell bug we ended up sed-ing the entrypoint by line number in the dockerfile
> #8465
> 
> this is an un-maintable approach, and we will need to start from scratch and mantain a fork of the entrypoint in the source code. therefore I re-open this PR

Arguably that change should **not** go into the public image, for debugging a specific issue ?

And we don't have to re-open _this_ PR, if all we want to do is fork the kind base (#7788).
The main problem with it was that it was doing several unrelated things at the same time...

afbjorklund:
@medyagh @alonyb : I think we can close this old obsolete PR now, in favor of ~#8895~ and #9225

PR #7890, opened 2020-04-24
MEP: Add proposal for scheduled shutdown/pause

tstromberg:
Designed to support build tools which need to run commands within a Kubernetes cluster, but don't wish to implement a separate background service to free up the resources if they go unused for a while.

Prospective users are the Google Cloud Code IDE and the Google Cloud SDK, though this may prove to be a popular option for pipelines which prefer a <1s latency between executions with the occasional 30s latency if idle too long.

Example usage:

`minikube start && minikube stop --after 5m`

* If the cluster is up already, this updates the timer
* If the cluster is down, it starts minikube, and updates the timer

### DIFF
--- a/enhancements/proposed/20200424-scheduled-shutdown/20200424-scheduled-shutdown.md
+++ b/enhancements/proposed/20200424-scheduled-shutdown/20200424-scheduled-shutdown.md
@@ -26,12 +26,14 @@ This is useful for two sets of users:
 ## Goals
 
 * The ability to schedule a pause or shutdown event
-* The ability to defer or update the scheduled event
+* The ability to defer the scheduled pause or shutdown event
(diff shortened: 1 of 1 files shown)


tstromberg:
> I assume this would be solving a different problem than auto pause. The goal of this Mep is to design an idea for, someone wants to stop minikube but they don't want to wait for it. But they just want to make sure it will happen. Right ?

Correct, this is for someone who wants to free up all resources (CPU, memory) eventually but is not available or able to guarantee it for whatever reason. This is for the `Cloud Code` or `gcloud run` cases, where they would like to free everything up eventually. 

I've clarified the proposal to mention auto-pause as a complimentary idea.

> 
> Will auto pause goal would be making sure not to use CPU load when not needed.

PR #5543, opened 2019-10-04
Add addons flag to 'minikube start' in order to enable specified addons

govargo:
### What type of PR is this?
/kind feature

### What this PR does / why we need it:
So far we have to use the `minikube addons enable` command to enable the addons.
Now, this PR enables us to use addons with `minikube start --addons` command flag without `minikube addons enable`.

### Which issue(s) this PR fixes:
Fixes #5503

### Does this PR introduce a user-facing change?
Yes. this PR add flag option to `minikube start` command .

**Before this PR, addons flow to enable  is following**
```
1. Search the available addons
minikube addons list

2. Enable addons
minikube addons enable <name>

3. Start minikube
minikube start
```

**After this PR, addons flow to enable  is following**
```
1. Search the available addons
minikube addons list

2. Start minikube with available addons
minikube addons --addons <name1> --addons <name2> --addons <nameN>
```

### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:
```
NONE
```

PR #7924, opened 2020-04-28
Example how to enable multiple addons on start command

### LINKED ISSUE #7920: Clarify --addons flag syntax
I would like to start Minikube with one of the following commands, in order to start Minikube with more than one addons enabled:
`minikube start --addons="ingress registry-creds"`
`minikube start --addons="ingress, registry-creds"`
`minikube start --addons="ingress; registry-creds"`

None of the above seems to work.

In the [docs](https://minikube.sigs.k8s.io/docs/commands/start/), I can see that the required type is _minikube addons list_, although the actual syntax is not clear to me.

Anyways, I can enable multiple addons using mutiple `minikube addons` subcommand issued when Minikube is already started.

radeksm:
Example how to enable multiple addons on start command

Multiple addons have been added here: https://github.com/kubernetes/minikube/pull/5543
there is a short spec which explains how does it work.

fixes #7920

### DIFF
--- a/site/content/en/docs/handbook/deploying.md
+++ b/site/content/en/docs/handbook/deploying.md
@@ -29,10 +29,10 @@ To enable an add-on, see:
 minikube addons enable <name>
 ```
 
-To enable an addon at start-up:
+To enable an addon at start-up, where *--addons* option can be specified multiple times:
 
 ```shell
-minikube start --addons <name>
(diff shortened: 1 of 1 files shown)


tstromberg:
Wow, I really don't like this syntax at all. I had assumed comma-delimited worked when I made the name :( 

Thank you for documenting it!

Issue #7472, opened 2020-04-07
Minikube start fails offline with docker driver since the kicbase image is not available in the cache


eoinreilly93:
I am trying to install minikube on a CentOS 7 VM that has no access to the internet using the `docker` driver. To do this, I downloaded and installed minikube on a windows machine using Virtualbox with access to the internet and then I used the cache directory to install offline on the VM. However, when I run the minikube start command, it timed out as it was trying to download an image that was not located in the cache directory. This image was `gcr.io/k8s-minikube/kicbase:v0.0.07`. 

To try and resolve this, I pulled this image locally and saved it to `.tar` file, and brought it across to my VM, and used `docker load` to add it to my locally registry, thinking it would solve the issue. This has not worked though, since the image repo digest is not saved as part of the save command. Minikube appears to be using the repo digest as a way to verify the image, and since it doesn't exist, it then tries to pull the image from the internet, which it cannot do. 

1. Why is it using the image repo digest and not the image id to verify if the image is located in the local docker registry?
2. Is there a way I can get around this problem? As I see currently see it, there is no way to install minikube without an internet connection while using the `docker` driver. This seems like a major issue.

My environment is:
CentOS 7.71908
Docker 19.03.3
Minikube 1.8.2

**Steps to reproduce the issue:** 

1.Login as as non-root user
2. Bring minikube installation binary onto VM
3. Run chmod +x on the binary
4. Run install minikube /usr/local/bin
5. Run minikube version to check installation is successful
6. Bring the cache files into the .minikube/cache directory
7. Load the docker images into the local registry
8. Bring the tarred kicbase image on to the VM and load it into the local registry
 9. Start minikube with minikube start --driver=docker --kubernetes-version=1.16.3

**Full output of failed command:** 

[servicereporting@SR-SVR-206 ~]$ minikube start --driver=docker --kubernetes-version=1.16.3
! minikube v1.8.2 on Centos 7.7.1908
* Using the docker driver based on user configuration
E0406 17:25:10.412815    7184 cache.go:106] Error downloading kic artifacts:  getting remote image: Get https://gcr.io/v2/: dial tcp: lookup gcr.io on 15.6.0.42:53: read udp 15.6.1.206:41122->15.6.0.42:53: i/o timeout
* Creating Kubernetes in docker container with (CPUs=2) (8 available), Memory=15600MB (62753MB available) ...
*
[...]

Issue #7942, opened 2020-04-30
detect fast which image repo is best for kic base image, instead of timeout


medyagh:
### modifying  /etc/hosts to emulate not having access to gcr.io
```
$ cat /etc/hosts | grep gcr.io
127.0.0.2       gcr.io
```

```
medya@~/workspace/minikube (kicbase_gh_package) $ ./out/minikube start --driver=docker
😄  minikube v1.10.0-beta.2 on Darwin 10.15.4
✨  Using the docker driver based on user configuration
👍  Starting control plane node minikube in cluster minikube
🚜  Pulling base image ...
E0429 22:41:16.019427   88285 cache.go:128] Error downloading kic artifacts:  getting remote image: Get https://gcr.io/v2/: dial tcp 127.0.0.2:443: i/o timeout
🔥  Creating docker container (CPUs=2, Memory=4000MB) ...
🤦  StartHost failed, but will try again: creating host: create: creating: create kic node: create container: docker run -d -t --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --label mode.minikube.sigs.k8s.io=minikube --volume minikube:/var --cpus=2 --memory=4000mb -e container=docker --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 --publish=127.0.0.1::5000 gcr.io/k8s-minikube/kicbase:v0.0.10@sha256:f58e0c4662bac8a9b5dda7984b185bad8502ade5d9fa364bf2755d636ab51438: exit status 125
stdout:

stderr:
Unable to find image 'gcr.io/k8s-minikube/kicbase:v0.0.10@sha256:f58e0c4662bac8a9b5dda7984b185bad8502ade5d9fa364bf2755d636ab51438' locally
docker: Error response from daemon: Get http://gcr.io/v2/: net/http: request canceled (Client.Timeout exceeded while awaiting headers).
See 'docker run --help'.

🤷  docker "minikube" container is missing, will recreate.
[...]

Issue #7915, opened 2020-04-26
kicbase does not respect --image-repository or --image-mirror-country flags


wenhui-prudencemed:
**Steps to reproduce the issue:** 

minikube start --image-mirror-country=cn --alsologtostderr

**Full output of failed command:** 

I0426 21:05:42.188776    5845 cache.go:66] save to tar file registry.cn-hangzhou.aliyuncs.com/google_containers/kube-apiserver:v1.18.0 -> /home/prudencemed/.minikube/cache/images/registry.cn-hangzhou.al
iyuncs.com/google_containers/kube-apiserver_v1.18.0 succeeded
I0426 21:05:42.188790    5845 cache.go:73] Successfully saved all images to host disk.
E0426 21:08:16.404783    5845 cache.go:117] Error downloading kic artifacts:  getting remote image: Get https://gcr.io/v2/: dial tcp 64.233.189.82:443: i/o timeout
I0426 21:08:16.405202    5845 start.go:221] acquiring machines lock for minikube: {Name:mk658abb5985c745ceef5c35481ecb92e05e0b87 Clock:{} Delay:500ms Timeout:15m0s Cancel:<nil>}
I0426 21:08:16.405286    5845 start.go:225] acquired machines lock for "minikube" in 61.735µs
I0426 21:08:16.405304    5845 start.go:85] Skipping create...Using existing machine configuration
I0426 21:08:16.405312    5845 fix.go:53] fixHost starting:
I0426 21:08:16.405572    5845 oci.go:268] executing with [docker inspect -f {{.State.Status}} minikube] timeout: 19s
I0426 21:08:16.437775    5845 fix.go:105] recreateIfNeeded on minikube: state=Error err="docker inspect -f {{.State.Status}} minikube" failed: exit status 1: Template parsing error: template: :1:8: exec
uting "" at <.State.Status>: map has no entry for key "State"
I0426 21:08:16.437806    5845 fix.go:110] machineExists: false. err=machine does not exist
🤷  docker "minikube" container is missing, will recreate.
I0426 21:08:16.452437    5845 delete.go:117] DEMOLISHING minikube ...

**Environment: minikube v1.10.0-beta.1 on Ubuntu 18.04**

seems it still try to access gcr.io after I set --image-mirror-country=cn

maybe the image mirror doesn't download kicbase:v0.0.9 to local?

afbjorklund:
This is still a work-in-progress, see #7447 #7844. On `master` you can now select another location.

[...]

PR #7943, opened 2020-04-30
docker/podman: add alternative image repo for base image to fallback

medyagh:
This PR adds an alternative image-repo in https://github.com/kubernetes/minikube/packages
 for kic base image. instead of gcr.io
fixes: https://github.com/kubernetes/minikube/issues/7472

## First: Emulating not having access to GCR:
```
$ cat /etc/hosts | grep gcr.io
127.0.0.1       gcr.io

$ docker rmi gcr.io/k8s-minikube/kicbase:v0.0.10
```
### Before this PR: time out with ugly error when image not avialble
```
 $ minikube start --driver=docker
😄  minikube v1.10.0-beta.2 on Darwin 10.15.4
🚜  Pulling base image ...
E0429 22:29:51.276201   86892 cache.go:122] Error downloading kic artifacts:  getting remote image: Get https://gcr.io/v2/: dial tcp 127.0.0.2:443: i/o timeout
🤦  StartHost failed, but will try again: creating host: create: creating: create kic node: create container: docker run -d -t --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --label mode.minikube.sigs.k8s.io=minikube --volume minikube:/var --cpus=2 --memory=4000mb -e container=docker --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 --publish=127.0.0.1::5000 gcr.io/k8s-minikube/kicbase:v0.0.10@sha256:f58e0c4662bac8a9b5dda7984b185bad8502ade5d9fa364bf2755d636ab51438: exit status 125
stdout:

stderr:
Unable to find image 'gcr.io/k8s-minikube/kicbase:v0.0.10@sha256:f58e0c4662bac8a9b5dda7984b185bad8502ade5d9fa364bf2755d636ab51438' locally
docker: Error response from daemon: Get http://gcr.io/v2/: net/http: request canceled (Client.Timeout exceeded while awaiting headers).
See 'docker run --help'.

🤷  docker "minikube" container is missing, will recreate.
E0429 22:30:30.176700   86892 oci.go:79] docker daemon seems to be stuck. Please try restarting your docker. Will try to delete anyways: unknown state "minikube": docker inspect minikube --format={{.State.Status}}: exit status 1
stdout:


stderr:
Template parsing error: template: :1:8: executing "" at <.State.Status>: map has no entry for key "State"
😿  Failed to start docker container. "minikube start" may fix it: recreate: creating host: create: creating: create kic node: create container: docker run -d -t --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --label mode.minikube.sigs.k8s.io=minikube --volume minikube:/var --cpus=2 --memory=4000mb -e container=docker --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 --publish=127.0.0.1::5000 gcr.io/k8s-minikube/kicbase:v0.0.10@sha256:f58e0c4662bac8a9b5dda7984b185bad8502ade5d9fa364bf2755d636ab51438: exit status 125
stdout:

stderr:
Unable to find image 'gcr.io/k8s-minikube/kicbase:v0.0.10@sha256:f58e0c4662bac8a9b5dda7984b185bad8502ade5d9fa364bf2755d636ab51438' locally
docker: Error response from daemon: Get http://gcr.io/v2/: net/http: request canceled (Client.Timeout exceeded while awaiting headers).
See 'docker run --help'.


💣  error provisioning host: Failed to start host: recreate: creating host: create: creating: create kic node: create container: docker run -d -t --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --label mode.minikube.sigs.k8s.io=minikube --volume minikube:/var --cpus=2 --memory=4000mb -e container=docker --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 --publish=127.0.0.1::5000 gcr.io/k8s-minikube/kicbase:v0.0.10@sha256:f58e0c4662bac8a9b5dda7984b185bad8502ade5d9fa364bf2755d636ab51438: exit status 125
stdout:

stderr:
Unable to find image 'gcr.io/k8s-minikube/kicbase:v0.0.10@sha256:f58e0c4662bac8a9b5dda7984b185bad8502ade5d9fa364bf2755d636ab51438' locally
docker: Error response from daemon: Get http://gcr.io/v2/: net/http: request canceled (Client.Timeout exceeded while awaiting headers).
See 'docker run --help'.


```

### After this PR
```
$ make && ./out/minikube start --driver=docker
😄  minikube v1.10.0-beta.2 on Darwin 10.15.4
🚜  Pulling base image ...
E0429 23:07:03.587714   94304 cache.go:130] Error downloading kic artifacts:  getting remote image: GET https://docker.pkg.github.com/v2/kubernetes/minikube/kicbase/manifests/v0.0.10: UNAUTHORIZED: GitHub Docker Registry needs login
🐳  Preparing Kubernetes v1.18.1 on Docker 19.03.2 ...
```
#### verifying image with docker ps
```
medya@~/workspace/minikube (kicbase_gh_package) $ docker ps
CONTAINER ID        IMAGE                                                       COMMAND                  CREATED             STATUS              PORTS                                                                                                      NAMES
0e0a63c61b83        docker.pkg.github.com/kubernetes/minikube/kicbase:v0.0.10   "/usr/local/bin/entr…"   46 seconds ago      Up 44 seconds       127.0.0.1:32787->22/tcp, 127.0.0.1:32786->2376/tcp, 127.0.0.1:32785->5000/tcp, 127.0.0.1:32784->8443/tcp   minikube
```

### footnotes and not in the scope of this PR

1- github docker packages does NOT support pulling with SHA, which we consider it Acceptable but not ideal for our fallback image. ( see the note in this doc https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-docker-for-use-with-github-packages)

2- our go-containerregstiry lib is producing some errors about github needs login, that seems to be harmless error. (a spam error) but wont affect the functionality of using fallback. however that will be another PR to refactor our image.WriteToDaemon to not produce that error or handle it carefuly. so I will not consider to do that in this PR.

3- created another issue that we should detect early which image repo is best instead of failing one after time out and trying another one. that is also NOT in the scope of this PR.
 https://github.com/kubernetes/minikube/issues/7942

### DIFF
--- a/pkg/minikube/node/cache.go
+++ b/pkg/minikube/node/cache.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/viper"
 	"golang.org/x/sync/errgroup"
 	cmdcfg "k8s.io/minikube/cmd/minikube/cmd/config"
+	"k8s.io/minikube/pkg/drivers/kic"
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/minikube/download"
@@ -106,7 +107,14 @@ func beginDownloadKicArtifacts(g *errgroup.Group, driver string, cRuntime string
 			out.T(out.Pulling, "Pulling base image ...")
 			g.Go(func() error {
 				glog.Infof("Downloading %s to local daemon", baseImage)
-				return image.WriteImageToDaemon(baseImage)
+				err := image.WriteImageToDaemon(baseImage)
+				if err != nil {
+					// registry/driver package uses viper.Set
+					viper.Set("base-image", kic.BaseImageFallBack)
+					glog.Infof("failed to download %s will try to download the fallback image from %s", baseImage, kic.BaseImageFallBack)
+					return image.WriteImageToDaemon(kic.BaseImageFallBack)
+				}
+				return nil
 			})
 		}
 	} else {
--- a/pkg/minikube/node/cache.go
+++ b/pkg/minikube/node/cache.go
@@ -109,9 +109,8 @@ func beginDownloadKicArtifacts(g *errgroup.Group, cc *config.ClusterConfig) {
 				glog.Infof("Downloading %s to local daemon", cc.KicBaseImage)
 				err := image.WriteImageToDaemon(cc.KicBaseImage)
 				if err != nil {
-					origImage := cc.KicBaseImage
+					glog.Infof("failed to download base-image %q will try to download the fallback base-image %q instead.", cc.KicBaseImage, kic.BaseImageFallBack)
 					cc.KicBaseImage = kic.BaseImageFallBack
-					glog.Infof("failed to download base image %q will try to download the fallback image %q instead.", origImage, kic.BaseImageFallBack)
 					return image.WriteImageToDaemon(kic.BaseImageFallBack)
 				}
 				return nil
--- a/pkg/minikube/node/cache.go
+++ b/pkg/minikube/node/cache.go
@@ -106,6 +106,7 @@ func beginDownloadKicArtifacts(g *errgroup.Group, cc *config.ClusterConfig) {
 		if !image.ExistsImageInDaemon(cc.KicBaseImage) {
 			out.T(out.Pulling, "Pulling base image ...")
 			g.Go(func() error {
+				// TODO #8004 : make base-image respect --image-repository
 				glog.Infof("Downloading %s to local daemon", cc.KicBaseImage)
 				err := image.WriteImageToDaemon(cc.KicBaseImage)
 				if err != nil {
--- a/pkg/minikube/node/cache.go
+++ b/pkg/minikube/node/cache.go
@@ -109,7 +109,8 @@ func beginDownloadKicArtifacts(g *errgroup.Group, driver string, cRuntime string
 				glog.Infof("Downloading %s to local daemon", baseImage)
 				err := image.WriteImageToDaemon(baseImage)
 				if err != nil {
-					// registry/driver package uses viper.Set
+					// TODO : remove this global set when this issue is closed
+					// https://github.com/kubernetes/minikube/issues/7944
 					viper.Set("base-image", kic.BaseImageFallBack)
 					glog.Infof("failed to download %s will try to download the fallback image from %s", baseImage, kic.BaseImageFallBack)
 					return image.WriteImageToDaemon(kic.BaseImageFallBack)
--- a/pkg/minikube/node/cache.go
+++ b/pkg/minikube/node/cache.go
@@ -100,19 +100,18 @@ func doCacheBinaries(k8sVersion string) error {
 }
 
 // BeginDownloadKicArtifacts downloads the kic image + preload tarball, returns true if preload is available
-func beginDownloadKicArtifacts(g *errgroup.Group, driver string, cRuntime string, baseImage string) {
-	glog.Infof("Beginning downloading kic artifacts for %s with %s", driver, cRuntime)
-	if driver == "docker" {
-		if !image.ExistsImageInDaemon(baseImage) {
+func beginDownloadKicArtifacts(g *errgroup.Group, cc *config.ClusterConfig) {
+	glog.Infof("Beginning downloading kic artifacts for %s with %s", cc.Driver, cc.KubernetesConfig.ContainerRuntime)
+	if cc.Driver == "docker" {
+		if !image.ExistsImageInDaemon(cc.KicBaseImage) {
 			out.T(out.Pulling, "Pulling base image ...")
 			g.Go(func() error {
-				glog.Infof("Downloading %s to local daemon", baseImage)
-				err := image.WriteImageToDaemon(baseImage)
+				glog.Infof("Downloading %s to local daemon", cc.KicBaseImage)
+				err := image.WriteImageToDaemon(cc.KicBaseImage)
 				if err != nil {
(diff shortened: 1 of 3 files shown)


afbjorklund:
> github docker packages does NOT support pulling with SHA

What was the problem with using the regular tag (instead of the digest) ?

medyagh:
> > github docker packages does NOT support pulling with SHA
> 
> What was the problem with using the regular tag (instead of the digest) ?

that didnt work either, acording to their docs we must not use the sha

pkg/minikube/node/cache.go
```diff
@@ -107,7 +108,15 @@ func beginDownloadKicArtifacts(g *errgroup.Group, driver string, cRuntime string
 			out.T(out.Pulling, "Pulling base image ...")
 			g.Go(func() error {
 				glog.Infof("Downloading %s to local daemon", baseImage)
```
tstromberg, line 110: Please add a TODO to respect `--image-repository`: https://github.com/kubernetes/minikube/issues/7915

pkg/minikube/node/cache.go
```diff
@@ -107,7 +108,15 @@ func beginDownloadKicArtifacts(g *errgroup.Group, driver string, cRuntime string
...
-				return image.WriteImageToDaemon(baseImage)
+				err := image.WriteImageToDaemon(baseImage)
+				if err != nil {
+					// TODO : remove this global set when this issue is closed
+					// https://github.com/kubernetes/minikube/issues/7944
+					viper.Set("base-image", kic.BaseImageFallBack)
```
tstromberg, line 115: This feels really bad - I would rather see the global fixed in a PR before merging this one.

medyagh: the issue that was set to fix that was evaluated as Backlog priority. so that would be a longer wait to get this PR merged.

afbjorklund:
> that didnt work either, acording to their docs we must not use the sha.

My question was why we are not using the tag (`v0.0.10`) instead ?

That way it (kicbase) would work with all runtimes and all registries.

Do we plan to have multiple images per tag, or why do we need digest ?

pkg/minikube/node/cache.go
tstromberg: @medyagh - that's OK. I would rather see this done correctly than quickly. Feel free to update priorities as you see fit.

medyagh:
> > that didnt work either, acording to their docs we must not use the sha.
> 
> My question was why we are not using the tag (`v0.0.10`) instead ?
> 
> That way it (kicbase) would work with all runtimes and all registries.
> 
> Do we plan to have multiple images per tag, or why do we need digest ?

using SHA is very valuable for security and integrity, I prefer we use it whenever we can. at least for our official repo.
in this PR for all alternative repos we skip the SHA.

pkg/minikube/node/cache.go
medyagh: sounds good, here is a PR for that https://github.com/kubernetes/minikube/pull/7985 whenever you feel ready to review.

medyagh: done

afbjorklund:
> using SHA is very valuable for security and integrity, I prefer we use it whenever we can. at least for our official repo.

Problem is that the SHA is on something that we cannot verify (without the Docker registry)

The image itself doesn't know which tag/digest it will end up in, and the ID is not consistent...

So we end up having to patch every tool and library, for something that Docker doesn't support ?

pkg/minikube/node/cache.go
```diff
@@ -99,14 +100,20 @@ func doCacheBinaries(k8sVersion string) error {
...
-		if !image.ExistsImageInDaemon(baseImage) {
+func beginDownloadKicArtifacts(g *errgroup.Group, cc *config.ClusterConfig) {
+	glog.Infof("Beginning downloading kic artifacts for %s with %s", cc.Driver, cc.KubernetesConfig.ContainerRuntime)
+	if cc.Driver == "docker" {
+		if !image.ExistsImageInDaemon(cc.KicBaseImage) {
 			out.T(out.Pulling, "Pulling base image ...")
 			g.Go(func() error {
-				glog.Infof("Downloading %s to local daemon", baseImage)
-				return image.WriteImageToDaemon(baseImage)
+				glog.Infof("Downloading %s to local daemon", cc.KicBaseImage)
```
tstromberg, line 109: Please add TODO to respect the `--image-repository` flag.

medyagh: done

PR #7961, opened 2020-05-01
Use the correct binary for unpacking the preload

afbjorklund:
Previously hardcoded to *always* use a Docker volume,
even if using a different container driver (podman).

### DIFF
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -90,15 +90,12 @@ func DeleteContainer(ociBin string, name string) error {
 }
 
 // PrepareContainerNode sets up the container node before CreateContainerNode is called.
-// For the docker runtime, it creates a docker volume which will be mounted into kic
+// For the container runtime, it creates a volume which will be mounted into kic
 func PrepareContainerNode(p CreateParams) error {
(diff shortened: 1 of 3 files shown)


pkg/drivers/kic/oci/oci.go
```diff
@@ -90,15 +90,12 @@ func DeleteContainer(ociBin string, name string) error {
 }
 
 // PrepareContainerNode sets up the container node before CreateContainerNode is called.
-// For the docker runtime, it creates a docker volume which will be mounted into kic
+// For the docker runtime, it creates a volume which will be mounted into kic
```
medyagh, line 93: Update the comment, we have preload for both docker and containerd

afbjorklund:
@medyagh This failed when SELinux was enable on the host.
The mounted file was not properly accessible to the container:

`tar (child): /preloaded.tar: Cannot open: Permission denied`
`-??????????   ? ?    ?       ?            ? preloaded.tar`

So it also needs some kind of context added, to the mount.
That is, there needs to be a `:Z` appended, when getenforce...

https://github.com/containers/libpod/blob/master/troubleshooting.md#2-cant-use-volume-mount-get-permission-denied

Still missing Fedora machines in CI

PR #7968, opened 2020-05-01
Don't show any podman hint about restarting sudo

### LINKED ISSUE #7966: podman humor: restarting the sudo service may improve performance
* The slow message isn't helpful for podman at all.
* We shouldn't include the sudo timing in this. 

```
❗  Executing "sudo podman ps -a --format {{.Names}}" took an unusually long time: 10.547099308s
💡  Restarting the sudo service may improve performance.
```

afbjorklund:
Basically it only applies to the docker daemon, but _could_ also apply to an external podman socket.

Closes #7966

### DIFF
--- a/pkg/drivers/kic/oci/cli_runner.go
+++ b/pkg/drivers/kic/oci/cli_runner.go
@@ -131,7 +131,10 @@ func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
 	if warn {
 		if elapsed > warnTime {
 			out.WarningT(`Executing "{{.command}}" took an unusually long time: {{.duration}}`, out.V{"command": rr.Command(), "duration": elapsed})
-			out.ErrT(out.Tip, `Restarting the {{.name}} service may improve performance.`, out.V{"name": cmd.Args[0]})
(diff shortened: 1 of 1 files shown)


pkg/drivers/kic/oci/cli_runner.go
```diff
@@ -131,7 +131,10 @@ func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
 	if warn {
 		if elapsed > warnTime {
 			out.WarningT(`Executing "{{.command}}" took an unusually long time: {{.duration}}`, out.V{"command": rr.Command(), "duration": elapsed})
```
tstromberg, line 133: I don't think this should show for sudo either, as it may include the amount of time it took for a user to type their password.

afbjorklund: I was **not** planning to ask the user for their password, for running podman (it was supposed to be set up as NOPASSWD, so it should be able to run interactively without any password)

Since we run multiple podman commands at the same time, it ends up a bit confusing. And the regular sudo setup times out after 15 minutes, so it will come back as a surprise later on.

Issue #7352, opened 2020-04-01
create kic node: create container: Windows does not support privileged mode. 


devika-28:
command;  minikube start --driver=docker 
**Steps to reproduce the issue:** 

step 1. choco install minikube
step 2.  minikube start

the "--alsologtostderr" flag output:
<details>
W0401 14:28:03.949879     356 root.go:248] Error reading config file at C:\Users\Devika Bihani\.minikube\config\config.json: open C:\Users\Devika Bihani\.minikube\config\config.json: The system cannot find the file specified.
I0401 14:28:03.952879     356 notify.go:125] Checking for updates...
I0401 14:28:04.668073     356 start.go:259] hostinfo: {"hostname":"FX553VD","uptime":167305,"bootTime":1585564179,"procs":259,"os":"windows","platform":"Microsoft Windows 10 Enterprise","platformFamily":"Standalone Workstation","platformVersion":"10.0.18362 Build 18362","kernelVersion":"","virtualizationSystem":"","virtualizationRole":"","hostid":"4c0f7125-046e-4a58-8742-5227ba78ff5c"}
W0401 14:28:04.669072     356 start.go:267] gopshost.Virtualization returned error: not implemented yet
* minikube v1.9.0 on Microsoft Windows 10 Enterprise 10.0.18362 Build 18362
I0401 14:28:04.726059     356 driver.go:226] Setting default libvirt URI to qemu:///system
* Using the docker driver based on existing profile
I0401 14:28:05.069450     356 start.go:307] selected driver: docker
[...]

PR #7984, opened 2020-05-03
Detect windows container and exit with instructions

medyagh:
closes https://github.com/kubernetes/minikube/issues/7352
###  Before this PR
* confusing Error "Windows does not support privileged mode"
* minikube hopelessly retrying to re-create the container and waste user's time and make unpleasant logs.

which the real problem was, the new docker desktop by default has Windows  containers option "checked" during installation. so unless you change it, u will have windows conatiners.
Screen Shot 2020-05-02 at 4 19 12 PM

### After this PR: main scenario registry detects it
```
medmac@~/workspace/minikube (detect_windows_container) $ make && ./out/minikube start --driver=docker
😄  minikube v1.10.0-beta.2 on Darwin 10.13.6

❗  'docker' driver reported an issue: docker container type is windows
💡  Suggestion: Change container type to "linux" in Docker Desktop settings
📘  Documentation: https://minikube.sigs.k8s.io/docs/drivers/docker/#verify-docker-container-type-is-linux

💣  Failed to validate 'docker' driver
```

### After this PR scenario 2: Rare but possible
 ( it has passed the registry- the container type been switched after minikube registry had evaluated the it as healthy)

(please mind this output was produced on Mac by flipping the If statement to genearte the error)

```
😄  minikube v1.10.0-beta.2 on Darwin 10.13.6

💥  Your Docker Desktop container os type is Windows but Linux is required.
❗  Please change Docker settings to use Linux containers instead of Windows containers.
📘  https://docs.docker.com/docker-for-windows/#switch-between-windows-and-linux-containers
💡  You can verify your Docker container type by running:
        docker info --format '{{.OSType}}'
```


site/content/en/docs/drivers/docker.md
```diff
@@ -30,11 +30,20 @@ The Docker driver allows you to install Kubernetes into an existing Docker insta
 
 ## Troubleshooting
 
-- On macOS or Windows, you may need to restart Docker for Desktop if a command gets hung
```
medyagh, line 33: note that this advice was already stated above and not needed to repeat.

PR #7999, opened 2020-05-04
addons: fix initial retry delay, double maximum limit

### LINKED ISSUE #7719: TestErrorSpam: retry after the token is automatically created and added to the service account

tstromberg:
* The initial retry delay was accidentally set to microseconds!
* Double the maximum retry limit to 2 minutes to fight test flakes. This helps to ensure that addons are eventually deployed when requirements are met, without needing to understand what is required by each addon. 

Fixes #7719

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -258,7 +258,7 @@ func enableOrDisableAddonInternal(cc *config.ClusterConfig, addon *assets.Addon,
 		return err
 	}
 
-	return retry.Expo(apply, 100*time.Microsecond, time.Minute)
+	return retry.Expo(apply, 250*time.Millisecond, 2*time.Minute)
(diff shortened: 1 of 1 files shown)


tstromberg:
Updated PR description.

PR #8007, opened 2020-05-04
force systemd for start-stop test

sharifelgamal:
This is an attempt to get the embed-certs test to stabilize


medyagh:
Doesn't seem to make any difference

sharifelgamal:
Yeah I agree, I'll keep looking.

PR #8021, opened 2020-05-06
MEP: Add JSON output to minikube start

priyawadhwa:
Note: I haven't gotten access to the doc of requirements for this feature yet, so I may end up updating parts of this proposal in the next day or so.

### DIFF
--- a/enhancements/proposed/20200501-json-output/start-json-output.md
+++ b/enhancements/proposed/20200501-json-output/start-json-output.md
@@ -46,7 +46,7 @@ minikube start --output json
 
 ### Stderr
 Logs can be sent to stderr as usual.
-In addition, we will output error code from [err_map.go](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/problem/err_map.go) as a parsable JSON message.
+In addition, we will output error code from [err_map.go](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/problem/err_map.go) as a parsable JSON message to stderr.
 
 This will be done by adding the following function to the `out` package:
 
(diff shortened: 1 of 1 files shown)


enhancements/proposed/20200501-json-output/start-json-output.md
```diff
@@ -0,0 +1,177 @@
...
+
+
+We also need to know the total number of steps before minikube starts.
+For that to be possible, we need to know all of the logs we will print before starting.
+
+I propose creating a registry of logs, which is prefilled with all necessary logs before we start minikube.
+The registry would look like this:
+
+```go
+// Registry holds all user-facing logs
```
tstromberg, line 108: In general, I would prefer that implementation code & structs be kept out of a design proposal. Designs should focus on the novel interesting parts of the design.

For an API-focused change like this, it's fine to mention what the new API would look like, however. Comments  in particular should not be included.

enhancements/proposed/20200501-json-output/start-json-output.md
```diff
@@ -0,0 +1,177 @@
...
+#### Testing Plan
+Both unit tests and integration tests will be required to test this feature.
+
+Unit tests will cover:
+1. That the JSON output is correct and parsable
+
+Integration tests will cover:
+1. That all output to stdout is in JSON, ensuring that all user-facing logs have been registered
+   
+
```
tstromberg, line 174: Can you share how an example `minikube start` run would be rendered to stdout/stderr? It would be helpful to understand if the new outputs match what users are looking for.

PR #8035, opened 2020-05-07
Wait to add aliases to /etc/hosts before starting kubelet

### LINKED ISSUE #7921: TestStartStop/group/embed-certs on Docker: "k8s-app=kubernetes-dashboard" failed to start within 9m0s
```
start_stop_delete_test.go:141: (dbg) TestStartStop/group/embed-certs: waiting 9m0s for pods matching "k8s-app=kubernetes-dashboard" in namespace "kubernetes-dashboard" ...
start_stop_delete_test.go:141: ***** TestStartStop/group/embed-certs: pod "k8s-app=kubernetes-dashboard" failed to start within 9m0s: timed out waiting for the condition ****
```

As seen [here](https://storage.googleapis.com/minikube-builds/logs/7815/e3397d5/Docker_Linux.html#fail_TestStartStop%2fgroup%2fembed-certs)

priyawadhwa:
kubelet manages the value of `/etc/hosts` within running containers (e.g. kube-controller-manager). It takes this file from the host, but this causes a race condition between:

1. Us updating /etc/hosts with aliases on IP changes
1. kubelet starting k8s with current value of /etc/hosts

This change removes automatically starting kubelet when systemd comes up. Instead, we start it immediately after making our updates to `/etc/hosts`

This should fix the embed-certs integration test which has been flaky for a while.

Fixes #7921


pkg/drivers/kic/kic.go
```diff
@@ -294,6 +294,14 @@ func (d *Driver) Restart() error {
 	if err = d.Start(); err != nil {
 		return fmt.Errorf("start during restart %v", err)
 	}
+	// we need to restart kubelet to ensure that it picks up the correct node ip
+	// there seems to be a race condition between when the ip is updated
+	// on node restarts and kubelet starting up
+	glog.Infof("Restarting kubelet...")
+	cr := command.NewExecRunner() // using exec runner for interacting with docker/podman daemon
+	if _, err := cr.RunCmd(exec.Command("sudo", "systemctl", "restart", "kubelet")); err != nil {
```
medyagh, line 302: we should use the package we have, because it could be systemctl or init.d

medyagh: please use sysinit.New(k.c).Restart("kubelet")

pkg/minikube/node/start.go
```diff
@@ -180,12 +181,33 @@ func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
...
 }
 
+func restartKubeletIfNeeded(starter Starter) error {
+	if !starter.PreExists {
+		return nil
+	}
+	if !driver.IsKIC(starter.Host.DriverName) {
+		return nil
+	}
+	// on soft start and kic, we need to restart kubelet to ensure that it picks up the correct node ip
```
medyagh, line 201: I used to use "soft start" for starting a "running" minikube. I would use second-start or a different word that reflects starting a stopped minikube.

pkg/minikube/node/start.go
```diff
@@ -180,12 +181,40 @@ func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
...
 
 	// Write enabled addons to the config before completion
 	return kcs, config.Write(viper.GetString(config.ProfileName), starter.Cfg)
 }
 
+func restartKubeletIfNeeded(starter Starter) error {
+	if !starter.PreExists {
+		return nil
+	}
+	if !driver.IsKIC(starter.Host.DriverName) {
```
tstromberg, line 198: Can you give more detail on why we think this is only a KIC issue? It seems like it should be a problem for any case which changes IP or configuration.

pkg/minikube/node/start.go
```diff
@@ -180,12 +181,40 @@ func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
 		}
 	}
 
+	if err := restartKubeletIfNeeded(starter); err != nil {
+		return nil, errors.Wrap(err, "restarting kubelet")
```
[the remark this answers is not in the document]
priyawadhwa: Good suggestion on moving the restart, that's done!

I tried adding the kubelet-start phase, but it didn't resolve the race condition issue.

pkg/minikube/node/start.go
priyawadhwa: fair, new changes have removed the kic-specification

pkg/minikube/node/start.go
priyawadhwa: this code is now removed

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -542,6 +542,7 @@ func (k *Bootstrapper) restartCluster(cfg config.ClusterConfig) error {
 	cmds := []string{
 		fmt.Sprintf("%s phase certs all --config %s", baseCmd, conf),
 		fmt.Sprintf("%s phase kubeconfig all --config %s", baseCmd, conf),
+		fmt.Sprintf("sudo systemctl restart kubelet"),
```
tstromberg, line 545: I think this will break CloudShell.

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -801,7 +802,7 @@ func copyFiles(runner command.Runner, files []assets.CopyableFile) error {
 	return nil
 }
 
-func startKubeletIfRequired(runner command.Runner, sm sysinit.Manager) error {
+func startKubelet(runner command.Runner, sm sysinit.Manager) error {
```
tstromberg, line 805: I think this function can be removed and replaced entirely with calls to `sm.Enable("kubelet")` now.

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
priyawadhwa: We still run systemd within kic even in Cloud Shell -- I tried the following with this binary in a boosted Cloud Shell VM and it worked as expected:

```shell
minikube start
minikube stop
minikube addons enable dashboard
minikube start # Dashboard pods came up in about 20 seconds
```

Do you think we still need this change? I don't totally understand it's purpose right now.

tstromberg: You are totally right here. Sorry for the misunderstanding.

Anyways, restarting kubelet here is not entirely useful, as the new kubelet configuration hasn't been written to disk. Allegedly, the kubelet phase will write the updated config to disk, and restart it, but if you find that kubeadm does not actually restart kubelet for whatever reason, feel free to keep this statement.

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
priyawadhwa: I would prefer to keep it as a separate function, because it's called in `UpdateNode` which is a bit long & I think separating it into clear steps is more readable

tstromberg: `sm.Start("kubelet")` is as many lines as `startKubelet`

If you feel strongly, please just replace this function with a call to `sm.Start("kubelet")`. The rest of what it does is no longer meaningful.

priyawadhwa: Ah, does that mean we no longer need to copy over the kubelet systemd files over in this function? I thought that was the purpose of this function, but if they're copied over elsewhere then I'm all for removing this

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
priyawadhwa: That makes sense, I've removed it!

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
priyawadhwa: Found where they copy over & removed the function, thank you!

tstromberg: You'll need to fix `UpdateNode` so that it installs the files into `KubeletServiceFile` rather than `KubeletServiceFile+".new"`.

The reason for the complexity is that previously we were only restarting the kubelet if the configuration has changed on disk. Now we're completely controlling the kubelet lifecycle, so we will always start the kubelet when the node is coming up.

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
tstromberg: Are you going to add the kubeadm kubelet phase in it's place?

test/integration/start_stop_delete_test.go
```diff
@@ -71,6 +71,9 @@ func TestStartStop(t *testing.T) {
 				"--disable-driver-mounts",
 				"--extra-config=kubeadm.ignore-preflight-errors=SystemVerification",
 			}},
+			{"docker", constants.DefaultKubernetesVersion, []string{
```
medyagh, line 74: We won't need to add this additional test. We can remove this

priyawadhwa: done

PR #8039, opened 2020-05-07
Update golangci-lint to v1.26.0 (low priority)

tstromberg:
For no reason other than noticing we were not up to date.


Makefile
```diff
@@ -54,7 +54,7 @@ MINIKUBE_RELEASES_URL=https://github.com/kubernetes/minikube/releases/download
 
 KERNEL_VERSION ?= 4.19.107
 # latest from https://github.com/golangci/golangci-lint/releases
-GOLINT_VERSION ?= v1.23.6
+GOLINT_VERSION ?= v1.26.0
```
medyagh, line 57: do u mind solving the lints with this new version? I had tried this before and it discovered a lot of fmt.Sprintf("string")

Issue #8045, opened 2020-05-08
Ubuntu minikube can't start | Docker error 125


Phantomias3782:
Hey,
I'm a complete beginner to minikube. I'm trying to install it on an VirtualBox clear Ubuntu System and keep getting the 125 error by executing minikube start (or minikube start --memory 4096 --cpus 4) --> (Error) Code below.
I think the problem is, that docker only recognizes 1 cpu, but I don't know how to fix that.
docker system info returns:

Kernel Version: 5.4.0-29-generic
 Operating System: Ubuntu 20.04 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 1
 Total Memory: 7.776GiB

I have already uninstalled and reinstalled the complete Ubuntu system, and kepp getting the same Error.

~/Downloads$ minikube start --memory 4096 --cpus 4
😄  minikube v1.9.2 on Ubuntu 20.04
✨  Using the docker driver based on existing profile
👍  Starting control plane node m01 in cluster minikube
🚜  Pulling base image ...
🤷  docker "minikube" container is missing, will recreate.
🔥  Creating Kubernetes in docker container with (CPUs=4) (1 available), Memory=4096MB (7962MB available) ...
🤦  StartHost failed, but will try again: recreate: creating host: create: creating: create kic node: create container: failed args: [run -d -t --privileged --security-opt seccomp=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --label mode.minikube.sigs.k8s.io=minikube --volume minikube:/var --cpus=4 --memory=4096mb --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 gcr.io/k8s-minikube/kicbase:v0.0.8@sha256:2f3380ebf1bb0c75b0b47160fd4e61b7b8fef0f1f32f9def108d3eada50a7a81] output: docker: Error response from daemon: Range of CPUs is from 0.01 to 1.00, as there are only 1 CPUs available.
See 'docker run --help'.
: exit status 125

afbjorklund:
You will need more than 2 CPUs to run, so increase the resources given to your VM.

This is supposed to be verified and displayed better, already in the `minikube start`

afbjorklund:
[...]

Issue #7907, opened 2020-04-25
aws vm: Error : there are only 1 CPUs available.


r-suresh-kumar:
**Steps to reproduce the issue:** 

Pre-condition:
I am using Amazon EC2 linux instance with virtualization set to hvm.
But when I use command "grep -E --color 'vmx|svm' /proc/cpuinfo" to check if virtualization is enabled or disabled, it does not give any response.......I guess, virtualization is disabled. Please correct me if i am wrong

1. $ minikube start --vm-driver=none

OR

2. $ minikube start --driver=docker

OR

3. $ minikube start --no-vtx-check=false

  minikube v1.9.2 on Amazon 2018.03 (xen/amd64)
  Using the docker driver based on existing profile
  Starting control plane node m01 in cluster minikube
  Pulling base image ...
  docker "minikube" container is missing, will recreate.
  Creating Kubernetes in docker container with (CPUs=2) (1 available), Memory=985MB (985MB available) ...
  StartHost failed, but will try again: recreate: creating host: create: creating: create kic node: create container: failed args: [run -d -t --privileged --security-opt seccomp=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --label mode.minikube.sigs.k8s.io=minikube --volume minikube:/var --cpus=2 --memory=985mb --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 gcr.io/k8s-minikube/kicbase:v0.0.8@sha256:2f3380ebf1bb0c75b0b47160fd4e61b7b8fef0f1f32f9def108d3eada50a7a81] output: docker: Error response from daemon: Range of CPUs is from 0.01 to 1.00, as there are only 1 CPUs available.
See 'docker run --help'.
: exit status 125
  docker "minikube" container is missing, will recreate.
  Creating Kubernetes in docker container with (CPUs=2) (1 available), Memory=985MB (985MB available) ...

[...]

Issue #7850, opened 2020-04-22
Range of CPUs is from 0.01 to 1.00, as there are only 1 CPUs available


vinodkamisetti:
**Steps to reproduce the issue:** 

1. Failed to start docker container. "minikube start" may fix it.: recreate: creating host: create: creating: create kic node: create container: failed args: [run -d -t --privileged --security-opt seccomp=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --label mode.minikube.sigs.k8s.io=minikube --volume minikube:/var --cpus=2 --memory=983mb --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 gcr.io/k8s-minikube/kicbase:v0.0.8@sha256:2f3380ebf1bb0c75b0b47160fd4e61b7b8fef0f1f32f9def108d3eada50a7a81] output: docker: Error response from daemon: Range of CPUs is from 0.01 to 1.00, as there are only 1 CPUs available.
See 'docker run --help'.
: exit status 125
*
* minikube is exiting due to an error. If the above message is not useful, open an issue:
  - https://github.com/kubernetes/minikube/issues/new/choose

2. 
3.

**Full output of failed command:** 

**Full output of `minikube start` command used, if not already included:**

**Optional: Full output of `minikube logs` command:**
<details>

</details>

afbjorklund:
You need to have 2 CPU for Kubernetes, there is supposed to be a check for this though.

afbjorklund:
It also needs more memory (than 983mb), so maybe that check can be fixed too ?

Issue #7782, opened 2020-04-19
Unable to start the minikube on ubuntu 18.0  cloud 9 environment


Anumshr:
Steps to reproduce the issue: 
# installed kubectl
curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl
chmod +x ./kubectl
sudo mv ./kubectl /usr/local/bin/kubectl

#installed docker
sudo apt install -y docker.io

#install minikube
curl -Lo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64 \
  && chmod +x minikube
sudo mkdir -p /usr/local/bin/
sudo install minikube /usr/local/bin/

**Full output of failed command:** 
sudo minikube start –vm-driver=none –extra-config=kubeadm.ignore-preflight-errors=NumCPU –force –cpus 1
* minikube v1.9.2 on Ubuntu 18.04
* Using the none driver based on existing profile
* Starting control plane node  in cluster minikube
* Restarting existing none bare metal machine for "minikube" ...
* OS release is Ubuntu 18.04.4 LTS
* Preparing Kubernetes v1.18.0 on Docker 19.03.8 ...
  - kubelet.resolv-conf=/run/systemd/resolve/resolv.conf
! initialization failed, will try again: run: /bin/bash -c "sudo env PATH=/var/lib/minikube/binaries/v1.18.0:$PATH kubeadm init --config /var/tmp/minikube/kubeadm.yaml  --ignore-preflight-errors=DirAvailable--etc-kubernetes-manifests,DirAvailable--var-lib-minikube,DirAvailable--var-lib-minikube-etcd,FileAvailable--etc-kubernetes-manifests-kube-scheduler.yaml,FileAvailable--etc-kubernetes-manifests-kube-apiserver.yaml,FileAvailable--etc-kubernetes-manifests-kube-controller-manager.yaml,FileAvailable--etc-kubernetes-manifests-etcd.yaml,Port-10250,Swap,SystemVerification": exit status 1
stdout:
[init] Using Kubernetes version: v1.18.0
[preflight] Running pre-flight checks

stderr:
W0419 09:29:38.039336   20720 configset.go:202] WARNING: kubeadm cannot validate component configs for API groups [kubelet.config.k8s.io kubeproxy.config.k8s.io]
[...]

PR #8049, opened 2020-05-08
Document that minikube with kubeadm needs 2 CPUs

afbjorklund:
Eventually this also needs better validation, both for none and for docker driver.

This is supposed to be synced with the information https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#before-you-begin

> 2 GB or more of RAM per machine (any less will leave little room for your apps)
> 2 CPUs or more

The memory size is a little misleading (since it uses 2200M, not 2048M) but is not about cpus.
And we accept both 1GB (with a warning, and probably swap) and 2000 without any complaint.

For #6961 #7445 #7782 #7907 #7850 #8045

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -13,6 +13,7 @@ All you need is Docker (or similarly compatible) container or a Virtual Machine 
 
 ## What you’ll need
 
+* 2 CPUs or more
 * 2GB of free memory
 * 20GB of free disk space
 * Internet connection


afbjorklund:
It's a little confusing terminology, like "do you mean 2 cores ? or 2 sockets ? or 2 vCPU"
But better keep it same as upstream (2 CPUs). We just mean two threads, I think ?

Similarly, I'm not sure if it means 2000 MB or 2048 MB. Or should that be "mebibytes" ?
You're _supposed_ to spell it "2 GiB", but who does that. And the KVM translation is off...

But what we **do** want to avoid is an error, later on.

`Range of CPUs is from 0.01 to 1.00, as there are only 1 CPUs available.`

`[ERROR NumCPU]: the number of available CPUs 1 is less than the required 2` 

So let's use the term CPUs, it's what everyone is using.

afbjorklund:
And fortunately Docker allocates 2 vCPUs for their VM, by default. (see #7980)

So that means we only have problems with the memory, not with the CPUs.

site/content/en/docs/start/_index.md
```diff
@@ -13,6 +13,7 @@ All you need is Docker (or similarly compatible) container or a Virtual Machine
 
 ## What you’ll need
 
+* 2 CPUs or more
```
medyagh, line 16: just a nit: should we say 2 CPU cores ? I wonder if some ppl would think they acutally need two real CPUs.

afbjorklund: Commented above, think we'll use the same words as the official kubernetes installer ?

medyagh: sounds good

Issue #7818, opened 2020-04-21
Improve error handling on ARM architectures


LyleLee:
Can anyone instruct me a little bit about what is going on? I am working with:

+ ARM64 
+ CentOS 4.18.0-80.7.2.el7.aarch64

**Steps to reproduce the issue:** 

1. wget https://storage.googleapis.com/minikube/releases/v1.9.2/minikube-linux-arm64
2. ln -s minikube-linux-arm64 minikube
3. minikube start

**Full output of failed command:** 
```
[user1@arm64-server ~]$ minikube start
* minikube v1.9.2 on Centos 7.7.1908 (arm64)
* Using the docker driver based on existing profile
* Starting control plane node m01 in cluster minikube
* Pulling base image ...
* Restarting existing docker container for "minikube" ...
! StartHost failed, but will try again: provision: get ssh host-port: get host-bind port 22 for "minikube", output
Template parsing error: template: :1:4: executing "" at <index (index .NetworkSettings.Ports "22/tcp") 0>: error calling index: index of untyped nil
: exit status 1
* Restarting existing docker container for "minikube" ...
*
X Failed to start docker container. "minikube start" may fix it.: provision: get ssh host-port: get host-bind port 22 for "minikube", output
Template parsing error: template: :1:4: executing "" at <index (index .NetworkSettings.Ports "22/tcp") 0>: error calling index: index of untyped nil
: exit status 1
*
* minikube is exiting due to an error. If the above message is not useful, open an issue:
  - https://github.com/kubernetes/minikube/issues/new/choose
```

**Full output of `minikube start` command used, if not already included:**

As above.

**Optional: Full output of `minikube logs` command:**
```
[user1@kunpeng920 ~]$ minikube logs
* The control plane node must be running for this command
  - To fix this, run: "minikube start"
```

tstromberg:
Sorry about that. Do you mind sharing the output of: 

* `minikube start --alsologtostderr -v=1`
* `docker inspect minikube`

It would also be useful to know if using the latest beta fixes this issue for you: https://github.com/kubernetes/minikube/releases/tag/v1.10.0-beta.0

Thank you for the report!

tstromberg:
I'm marking this as a bug, because whatever the underlying root cause is here, our error handling is clearly not very good.

LyleLee:
Thanks a lot, here it is:

+ minikube start --alsologtostderr -v=1
```
[...]

PR #8053, opened 2020-05-09
docker/podman solution message for non-amd64 archs

medyagh:
adds solution message for https://github.com/kubernetes/minikube/issues/7818

### DIFF
--- a/pkg/minikube/registry/drvs/docker/docker.go
+++ b/pkg/minikube/registry/drvs/docker/docker.go
@@ -72,7 +72,7 @@ func configure(cc config.ClusterConfig, n config.Node) (interface{}, error) {
 func status() registry.State {
 	docURL := "https://minikube.sigs.k8s.io/docs/drivers/docker/"
 	if runtime.GOARCH != "amd64" {
-		return registry.State{Error: fmt.Errorf("doker driver is not supported on %q systems yet", runtime.GOARCH), Installed: false, Healthy: false, Fix: "Try other drivers", Doc: docURL}
+		return registry.State{Error: fmt.Errorf("docker driver is not supported on %q systems yet", runtime.GOARCH), Installed: false, Healthy: false, Fix: "Try other drivers", Doc: docURL}
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/minikube/registry/drvs/docker/docker.go
```diff
@@ -71,6 +71,10 @@ func configure(cc config.ClusterConfig, n config.Node) (interface{}, error) {
 
 func status() registry.State {
 	docURL := "https://minikube.sigs.k8s.io/docs/drivers/docker/"
+	if runtime.GOARCH != "amd64" {
+		return registry.State{Error: fmt.Errorf("doker driver is not supported on %q systems yet", runtime.GOARCH), Installed: false, Healthy: false, Fix: "Try other drivers", Doc: docURL}
```
afbjorklund, line 75: typo: docker

site/content/en/docs/drivers/podman.md
```diff
@@ -11,7 +11,14 @@ aliases:
...
 {{% /pageinfo %}}
 
-The podman driver is another kubernetes in container driver for minikube. similar to [docker](https://minikube.sigs.k8s.io/Drivers/docker/) driver. The podman driver is  experimental, and only supported on Linux and macOS (with a remote podman server)
+The podman driver is another kubernetes in container driver for minikube. similar to [docker](https://minikube.sigs.k8s.io/Drivers/docker/) driver. The podman driver is  experimental, and only supported on Linux and macOS (with a remote podman server).
+
+
+
+## Requirements
+- amd64 system 
+- Latest version of [Podman](https://podman.io/getting-started/installation)
```
afbjorklund, line 20: Shouldn't be required with latest version, then we will shut out platforms such as RHEL.

Probably should just say something like "Install Podman", similar to what Docker says ?

medyagh:
thanks @afbjorklund addressed the review comments.

afbjorklund:
Looks good to me! I think it will be easier to do kicbase for other arch, then to do the minikube.iso

PR #8057, opened 2020-05-09
Don't run the extraction tar container for podman

### LINKED ISSUE #8056: Podman driver crashes after timeout and volume recreate
It is _almost_ working, but there is an arbitrary 13 second timeout on start (for some reason).

Then it tries to delete the container and start over, but fails on the podman volume existing

```
😄  [podman] minikube v1.10.0-beta.2 on Ubuntu 18.04
✨  Using the podman (experimental) driver based on user configuration
👍  Starting control plane node podman in cluster podman
🔥  Creating podman container (CPUs=2, Memory=8000MB) ...
✋  Stopping "podman" in podman ...
🔥  Deleting "podman" in podman ...
🤦  StartHost failed, but will try again: creating host: create: creating: prepare kic ssh: apply authorized_keys file ownership, output 
** stderr ** 
Error: can only create exec sessions on running containers: container state improper

** /stderr **: chown docker:docker /home/docker/.ssh/authorized_keys: exit status 255
stdout:

stderr:
Error: can only create exec sessions on running containers: container state improper

🔥  Creating podman container (CPUs=2, Memory=8000MB) ...
😿  Failed to start podman container. "minikube start -p podman" may fix it: creating host: create: creating: setting up container node: creating volume for podman container: sudo -n podman volume create podman --label name.minikube.sigs.k8s.io=podman --label created_by.minikube.sigs.k8s.io=true: exit status 125
stdout:

stderr:
Error: volume with name podman already exists: volume already exists


💣  error provisioning host: Failed to start host: creating host: create: creating: setting up container node: creating volume for podman container: sudo -n podman volume create podman --label name.minikube.sigs.k8s.io=podman --label created_by.minikube.sigs.k8s.io=true: exit status 125
stdout:

stderr:
Error: volume with name podman already exists: volume already exists


```

afbjorklund:
The preloaded images will be extracted anyway, using ssh.

But if creating them on the volume before the container is
booted, means that /var will not be fully copied over to it.
And without /var/lib/dpkg and others, kicbase will not boot.
So skip the parallel extraction for podman, do it afterwards.

Probably shouldn't mount all of /var, but just a sub-set...

Closes #8056

### DIFF
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -122,22 +122,27 @@ func (d *Driver) Create() error {
 	}
 
 	var waitForPreload sync.WaitGroup
-	waitForPreload.Add(1)
-	go func() {
-		defer waitForPreload.Done()
-		// If preload doesn't exist, don't bother extracting tarball to volume
-		if !download.PreloadExists(d.NodeConfig.KubernetesVersion, d.NodeConfig.ContainerRuntime) {
-			return
-		}
-		t := time.Now()
-		glog.Infof("Starting extracting preloaded images to volume ...")
-		// Extract preloaded images to container
-		if err := oci.ExtractTarballToVolume(d.NodeConfig.OCIBinary, download.TarballPath(d.NodeConfig.KubernetesVersion, d.NodeConfig.ContainerRuntime), params.Name, BaseImage); err != nil {
-			glog.Infof("Unable to extract preloaded tarball to volume: %v", err)
-		} else {
-			glog.Infof("duration metric: took %f seconds to extract preloaded images to volume", time.Since(t).Seconds())
-		}
-	}()
+	if d.NodeConfig.OCIBinary == oci.Docker {
+		waitForPreload.Add(1)
+		go func() {
+			defer waitForPreload.Done()
+			// If preload doesn't exist, don't bother extracting tarball to volume
+			if !download.PreloadExists(d.NodeConfig.KubernetesVersion, d.NodeConfig.ContainerRuntime) {
+				return
+			}
+			t := time.Now()
+			glog.Infof("Starting extracting preloaded images to volume ...")
+			// Extract preloaded images to container
+			if err := oci.ExtractTarballToVolume(d.NodeConfig.OCIBinary, download.TarballPath(d.NodeConfig.KubernetesVersion, d.NodeConfig.ContainerRuntime), params.Name, BaseImage); err != nil {
+				glog.Infof("Unable to extract preloaded tarball to volume: %v", err)
+			} else {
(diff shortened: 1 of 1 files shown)


afbjorklund:
> Makes sense, I believe for podman the volume is just a folder on your system. If the host system had lz4 installed we could do it with lz4 in parallel. But probably that would be too much depending on user tools.

You could always go back to gzip, or even to plain .tar if that was the only problem. Or perhaps bundle lz4 with minikube, like we did for lzma when it was new and xz hadn't taken over yet.

afbjorklund:
> Makes sense, I believe for podman the volume is just a folder on your system.

It's a folder on Docker too, the main difference is in timing when it copies the files over.

But otherwise it is quite similar, there are only small differences in paths and timestamps:

```
 docker volume inspect docker
[
    {
        "CreatedAt": "2020-05-09T14:36:50+02:00",
        "Driver": "local",
        "Labels": {
            "created_by.minikube.sigs.k8s.io": "true",
            "name.minikube.sigs.k8s.io": "docker"
        },
        "Mountpoint": "/var/lib/docker/volumes/docker/_data",
        "Name": "docker",
        "Options": {},
        "Scope": "local"
    }
]
 sudo podman volume inspect podman
[
     {
          "Name": "podman",
          "Driver": "local",
          "Mountpoint": "/var/lib/containers/storage/volumes/podman/_data",
          "CreatedAt": "2020-05-09T14:35:14.431826651+02:00",
          "Labels": {
               "created_by.minikube.sigs.k8s.io": "true",
               "name.minikube.sigs.k8s.io": "podman"
          },
          "Scope": "local",
          "Options": {
               
          }
     }
]
```

So we have 5M of various junk under `/var` in the ubuntu image.
And that gets copied, from the image storage to the volume storage...

```
 docker run -it ubuntu:19.10 sh -c "du -hxcs /var/*"
4.0K	/var/backups
1.5M	/var/cache
3.7M	/var/lib
4.0K	/var/local
0	/var/lock
252K	/var/log
4.0K	/var/mail
4.0K	/var/opt
0	/var/run
4.0K	/var/spool
4.0K	/var/tmp
5.4M	total
 sudo ls /var/lib/docker/volumes/docker/_data
backups  cache	lib  local  lock  log  mail  opt  run  spool  tmp
 sudo ls /var/lib/containers/storage/volumes/podman/_data
backups  cache	lib  local  lock  log  mail  opt  run  spool  tmp
```

But we only actually _use_ a couple of those, like https://github.com/kubernetes/minikube/issues/8056#issuecomment-626199236

The rest just gets duplicated (from the image), for no particular reason.

afbjorklund:
My solution going forward would be to maybe keep the /var volume (for compatibiity when upgrading), but to only mount a _subset_ of the existing folders. Like we are doing with the ISO, to make them same.

But will open a new issue for that.

PR #8092, opened 2020-05-12
only copy apiserver certs to control plane

sharifelgamal:

### DIFF
--- a/pkg/minikube/bootstrapper/certs.go
+++ b/pkg/minikube/bootstrapper/certs.go
@@ -56,12 +56,9 @@ func SetupCerts(cmd command.Runner, k8s config.KubernetesConfig, n config.Node) 
 		return nil, errors.Wrap(err, "shared CA certs")
 	}
 
-	var xfer []string
-	if n.ControlPlane {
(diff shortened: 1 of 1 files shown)


pkg/minikube/bootstrapper/certs.go
```diff
@@ -56,9 +56,12 @@ func SetupCerts(cmd command.Runner, k8s config.KubernetesConfig, n config.Node)
 		return nil, errors.Wrap(err, "shared CA certs")
 	}
 
-	xfer, err := generateProfileCerts(k8s, n, ccs)
-	if err != nil {
-		return nil, errors.Wrap(err, "profile certs")
+	var xfer []string
```
medyagh, line 59: what is xfer short for ?

sharifelgamal: that's the name it already had, believe it's short of transfer?

sharifelgamal:
The test failures are all accounted for in existing issues. KVM is old k8s version, vbox linux is TestErrorSpam.

Issue #8014, opened 2020-05-05
TestFunctional/parallel/ComponentHealth flake on docker


priyawadhwa:
Original logs [here](https://storage.googleapis.com/minikube-builds/logs/8010/ec74a37/Docker_Linux.html#fail_TestFunctional%2fparallel%2fComponentHealth)

```
functional_test.go:288: (dbg) Run:  kubectl --context functional-20200505192422-29695 get cs -o=json
functional_test.go:307: unexpected status: False - item: {TypeMeta:{Kind:ComponentStatus APIVersion:v1} ObjectMeta:{Name:controller-manager GenerateName: Namespace: SelfLink:/api/v1/componentstatuses/controller-manager UID: ResourceVersion: Generation:0 CreationTimestamp:0001-01-01 00:00:00 +0000 UTC DeletionTimestamp:<nil> DeletionGracePeriodSeconds:<nil> Labels:map[] Annotations:map[]} Conditions:[{Type:Healthy Status:False Message:Get http://127.0.0.1:10252/healthz: dial tcp 127.0.0.1:10252: connect: connection refused Error:}]}
```

medyagh:
I still see this on HEAD today

medyagh:
seen again today, with more logging https://storage.googleapis.com/minikube-builds/logs/7959/3fa4b91/Docker_Linux.html

priyawadhwa:
According to Thomas's passrate script, this test has been a little flaky but passed on docker today:

```
$ /home/tstromberg/bin/passrate.sh TestFunctional/parallel/ComponentHealth
-------------------------------------------------
docker_Linux      | 2020-05-14 |  67% | 8 of 12
docker_Linux      | 2020-05-15 |  75% | 9 of 12
docker_Linux      | 2020-05-16 |  75% | 9 of 12
docker_Linux      | 2020-05-17 |  92% | 11 of 12
docker_Linux      | 2020-05-18 | 100% | 8 of 8
```

it also seemed to fail once today on KVM

```
KVM_Linux         | 2020-05-14 | 100% | 10 of 10
KVM_Linux         | 2020-05-15 |  92% | 11 of 12
KVM_Linux         | 2020-05-16 |  92% | 12 of 13
KVM_Linux         | 2020-05-17 | 100% | 11 of 11
KVM_Linux         | 2020-05-18 |  88% | 7 of 8
```

I'll do some digging.

To run locally:
```
[...]

PR #8095, opened 2020-05-12
recreate existing control plane node correctly

sharifelgamal:

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -881,8 +881,14 @@ func createNode(cc config.ClusterConfig, kubeNodeName string, existing *config.C
 	// Create the initial node, which will necessarily be a control plane
 	if existing != nil {
 		cp, err := config.PrimaryControlPlane(existing)
+		cp.KubernetesVersion = getKubernetesVersion(&cc)
 		if err != nil {
 			return cc, config.Node{}, err
+		}
+
+		for _, n := range existing.Nodes {
+			n.KubernetesVersion = getKubernetesVersion(&cc)
+			cc.Nodes = append(cc.Nodes, n)
 		}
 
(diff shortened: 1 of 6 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -877,16 +877,26 @@ func validateRegistryMirror() {
...
+	var err error
+	if existing == nil {
+		cp := config.Node{
+			Port:              cc.KubernetesConfig.NodePort,
+			KubernetesVersion: getKubernetesVersion(&cc),
+			Name:              kubeNodeName,
+			ControlPlane:      true,
+			Worker:            true,
+		}
+		cc.Nodes = []config.Node{cp}
```
medyagh, line 892: can we return here, and then get rid of else for less code complexity?

sharifelgamal: yep, done

test/integration/aaa_download_only_test.go
```diff
@@ -74,6 +74,7 @@ func TestDownloadOnly(t *testing.T) {
 
 					// skip for none, as none driver does not have preload feature.
 					if !NoneDriver() {
+						viper.Set("preload", true)
```
medyagh, line 77: this is not the right way to do this. setting global flag. please pass this as an arg to PreloadExists. maybe as an optional argument
...preload bool

medyagh (changes requested):

test/integration/aaa_download_only_test.go
sharifelgamal: is a hidden boolean trying to emulate a global flag really better than just using the flag?

sharifelgamal:
Docker test is a known flake: https://github.com/kubernetes/minikube/issues/8014

PR #8105, opened 2020-05-12
MEP: Add JSON output to `minikube start`

priyawadhwa:
This proposal may be updated in the next few days if additional requirements come in

Fixes https://github.com/kubernetes/minikube/issues/8133

### DIFF
--- a/enhancements/proposed/20200501-json-output/start-json-output.md
+++ b/enhancements/proposed/20200501-json-output/start-json-output.md
@@ -17,21 +17,20 @@ Please leave the above text in your proposal as instructions to the reader.
 ## Summary
 
 This proposal discusses adding JSON output to `minikube start`. 
-This feature will allow tools that rely on minikube, such as IDE extensions, to better parse errors and progress logs from `minikube start`.
+This feature will allow tools that rely on minikube, such as IDE extensions, to better parse errors and progress logs from STDOUT and STDERR on `minikube start`.
 This allows end users to see clear, and ideally actionable, error messages when minikube breaks.
 
-Minikube has the following to communicate information to users:
+Minikube currently communicates state to users via:
 1. Logs via glog (sent to stderr on `minikube start`)
-1. Outputs representing a step (sent to stdout on `minikube start`, e.g. "Preparing Kubernetes...")
-1. Outputs which don't represent a step. This can be an unexpected warning message or an `option`.
+1. Logs to stdout which represent a step on `minikube start` (e.g. "Preparing Kubernetes...")
+1. Logs to stdout which don't represent a step on `minikube start`. This can be an unexpected warning message or an `option`.
 
-This proposal focuses **only** on converting outputs representing clear steps in `minikube start` to JSON (option 2), and making sure error code typically sent to stderr is parsable.
+This proposal focuses **only** on converting outputs representing clear steps in `minikube start` to JSON (option 2), and making sure error code sent to stderr is parsable.
 
 ## Goals
 
 ### Stderr
 *   Error code from [err_map.go](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/problem/err_map.go) is parsable from stderr
-*   In case of a panic, default to sending all logs to stderr so that the user can see all logs
 
 ### Stdout
 *   Progress of each step of `minikube start` is communicated in JSON and includes:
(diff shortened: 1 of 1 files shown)


enhancements/proposed/20200501-json-output/start-json-output.md
```diff
@@ -57,6 +70,53 @@ If minikube fails, and an actionable error message exists, the following JSON wi
...
+
+`minikube start` output would look something like this:
+
+```
+$ minikube start
+{"Name":"Minikube Version","Message":"😄  minikube v1.10.0-beta.2 on Darwin 10.14.6\n","TotalSteps":9,"CurrentStep":1, "Type":"Log"}
+{"Name":"Selecting Driver","Message":"✨  Using the hyperkit driver based on user configuration\n","TotalSteps":9,"CurrentStep":2,"Type":"Log"}
+{"Name":"Starting Control Plane","Message":"👍  Starting node minikube in cluster minikube\n","TotalSteps":9,"CurrentStep":3,"Type":"Log"}
+{"Name":"Download Necessary Artifacts","Message":"💾  Downloading Kubernetes v1.18.1 preload ...\n","TotalSteps":9,"CurrentStep":4,"Type":"Log"}
+  {"Type":"Download", "Artifact":"preload.tar.gz", "Progress": "10%"}
```
jtzwu, line 102: If this message is an extension of Step 4, it would be nice to have some indication of that. It would remove the requirement that no other unrelated messages come in the middle.

enhancements/proposed/20200501-json-output/start-json-output.md
```diff
@@ -57,6 +70,53 @@ If minikube fails, and an actionable error message exists, the following JSON wi
...
+  {"Type":"Download", "Artifact":"preload.tar.gz", "Progress": "73%"}
+  {"Type":"Download", "Artifact":"preload.tar.gz", "Progress": "87%"}
+  {"Type":"Download", "Artifact":"preload.tar.gz", "Progress": "100%"}
+{"Name":"Creating Node","Message":"🔥  Creating hyperkit VM (CPUs=2, Memory=6000MB, Disk=20000MB) ...\n","TotalSteps":9,"CurrentStep":5,"Type":"Log"}
+{"Name":"Preparing Kubernetes","Message":"🐳  Preparing Kubernetes v1.18.1 on Docker 19.03.8 ...\n","TotalSteps":9,"CurrentStep":6,"Type":"Log"}
+{"Name":"Verifying Kubernetes","Message":"🔎  Verifying Kubernetes components...\n","TotalSteps":9,"CurrentStep":7,"Type":"Log"}
+{"Name":"Enabling Addons","Message":"🌟  Enabled addons: default-storageclass, storage-provisioner\n","TotalSteps":9,"CurrentStep":7,"Type":"Log"}
+{"Name":"Done","Message":"🏄  Done! kubectl is now configured to use \"minikube\"\n","TotalSteps":9,"CurrentStep":9,"Type":"Log"}
+```
+
```
jtzwu, line 113: Could you give an example of an error message?

enhancements/proposed/20200501-json-output/start-json-output.md
```diff
@@ -0,0 +1,231 @@
...
+}
+```
+
+### Stdout
+If `--output json` is specified, each step of `minikube start` will be output as JSON:
+
+```json
+{
+  "Name": "Selecting Driver",
+  "Message": "✨  Using the hyperkit driver based on user configuration\n",
```
balopat, line 77: I wonder if we really want to send emojis here... it feels odd for a structured eventing surface.

enhancements/proposed/20200501-json-output/start-json-output.md
```diff
@@ -0,0 +1,231 @@
...
+  1. The number of the current step
+  1. The expected total number of steps
+*   Progress of artifacts as they are being downloaded is communicated in JSON
+
+
+
+## Non-Goals
+
+*   Change the way we handle logs via glog
+*   Change the way we handle non-steps to stdout (Warnings, etc.)
```
drewpca, line 47: So are warnings/etc that would have gone to stdout completely suppressed in `--output=json` mode? 

Or are consumers expected to sort through lines-of-json and lines-of-not-json when they watch stdout? IMHO `--output=json` should guarantee that stdout is exclusively lines-of-json.

enhancements/proposed/20200501-json-output/start-json-output.md
drewpca: Consider something like this, and then other displays can decide if they want to be stylish like MK or not:
```
  Message: "brief message here with no decorations",
  Icon: "✨",
```

enhancements/proposed/20200501-json-output/start-json-output.md
```diff
@@ -0,0 +1,231 @@
...
+}
+```
+
+and each update on a downloaded artifact will be output as:
+
+```json
+{
+  "Type": "Download",
+  "Artifact": "preload.tar.gz",
+  "Progress": "10%"
```
drewpca, line 90: Consider `"CurrentStep": 2.1` if we're 10% of the way done with step 2. It conveys the same information, but readers might be able to do no processing and have their progress bars Just Work.

enhancements/proposed/20200501-json-output/start-json-output.md
```diff
@@ -0,0 +1,231 @@
...
+*	"Preparing Kubernetes"
+*	"Verifying Kubernetes"
+*	"Enabling Addons"
+*	"Done"
+
+When a log is called in the code, it will be associated with one of the above steps.
+
+This will allow us to determine which number step we are currently on, and include that information in the output.
+
+_Note: We may skip steps depending on the user's existing setup. For example, we may not need to "Verify Kubernetes" on a soft start. The current step/total step numbers will only be approximations_
```
drewpca, line 155: Can you bound those approximations? I'm concerned that if minikube only "sometimes" increases the number of steps, that's likely to be missed by consumers. Then, their minikube progress bar is glitchy (e.g. sometimes draws past 100%), and minikube ends up looking buggy.

Is it possible to include all possible steps in the count, so that TotalSteps never changes but CurrentStep often skips some numbers? That seems more robust against mistakes by event consumers. 

If not, i.e. if an unknown number of steps could suddenly get scheduled after the run starts, consider s/TotalSteps/EstimatedSteps/ or consider putting those new steps as fractional steps somewhere so progress bars continue to work.

enhancements/proposed/20200501-json-output/start-json-output.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+### Stdout - Download Steps
+
+To communicate progress on artifacts as they're being downloaded, we want JSON output that looks something like this during download:
+
+```json
+{
+  "Type": "Download",
+  "Artifact": "preload.tar.gz",
```
drewpca, line 195: For consumers who don't want to do anything special about download steps, consider keeping a Message field too:
```
"Message": "Downloading preload.tar.gz",
"Artifact": "preload.tar.gz",
```

enhancements/proposed/20200501-json-output/start-json-output.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+minikube uses the [go-getter](github.com/hashicorp/go-getter) library to download artifacts (preloaded tarballs, ISOs, etc).
+Currently, minikube passes in a [DefaultProgressBar](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/download/download.go#L48) to this library, which is used to communicate download progress to the user.
+
+Instead of passing in `DefaultProgressBar` we should be able to write our own object, something like `JSONOutput`, which will print the current progress of the download in the JSON format specified above instead of showing a progress bar in the terminal.
+
+
+#### Testing Plan
+Both unit tests and integration tests will be required to test these features feature.
```
drewpca, line 207: s/these features feature/this feature/

enhancements/proposed/20200501-json-output/start-json-output.md
```diff
@@ -0,0 +1,231 @@
...
+### Cloud Events
+
+I briefly looked into using [Cloud Events](https://github.com/cloudevents/spec) to send events to clients, specifically looking at the [Go SDK](https://github.com/cloudevents/sdk-go).
+
+Pros:
+* Standardized way of sending events
+* Supported by CNCF
+
+Cons:
+* Having minikube set up an HTTP server adds extra complexity to this proposal. If clients can instead parse JSON info from stdout/stderr then that would be the simplest solution.
```
drewpca, line 231: AIUI, json over whatever channel you want is also fine. https://github.com/cloudevents/spec/blob/master/spec.md#example even shows one. 

Diffs from your samples to a proper cloud event:
* add an 'id' to each event (so downstream processors can talk about them and deal with dups)
* add a 'source' to each event, like 'https://minikube.sigs.k8s.io/toolEvents' (as opposed to runtime events?)
* add `'specversion': '1.0'`
* change `'Type': 'Download'` to `'type': 'io.k8s.sigs.minikube.download'`, etc
* maybe add 'subject': 'step' (to separate minikube tool's progress steps from future reports that happen at tool runtime)
* optional: add `'time': '<rfc3339 time>'` (not sure if it helps anything here)
* lowercasesmash all your other field names, because that's the cloud events way :)


Example scenario where MK events get routed around:
1. I ask a cloud host to run minikube as part of some CI suite
2. Minikube outputs events on stdout like you're planning
3. The CI runner snarfs minikube's output (and other tools') and routes them back to me over an HTTP server-sent events connection
4. My UI renders those into a rich progress bar

enhancements/proposed/20200501-json-output/start-json-output.md
```diff
@@ -0,0 +1,231 @@
+# Add JSON Output to Minikube Start
```
drewpca, line 1: The same plan applies to `minikube stop` and maybe some other commands? Anything that could stall should be able to report what step it stalled on, I guess.

enhancements/proposed/20200501-json-output/start-json-output.md
priyawadhwa: I'm good with removing the emoji for now, and it should be easy to add in later if anybody wants it

enhancements/proposed/20200501-json-output/start-json-output.md
priyawadhwa: That's a good point -- I'll add warnings into this proposal.

enhancements/proposed/20200501-json-output/start-json-output.md
priyawadhwa: For sure, total steps should never change. Current steps may skip ahead, but it seems like that's ok. I'll rewrite this so it's clearer.

enhancements/proposed/20200501-json-output/start-json-output.md
priyawadhwa: thanks! fixed

enhancements/proposed/20200501-json-output/start-json-output.md
priyawadhwa: Just updated the proposal so there should be an error at the bottom now

enhancements/proposed/20200501-json-output/start-json-output.md
priyawadhwa: Added that in -- do you have any preference between this way and Drew's suggestion in this [comment](https://github.com/kubernetes/minikube/pull/8105#discussion_r424765969)?

enhancements/proposed/20200501-json-output/start-json-output.md
priyawadhwa: This should be extendable for other commands -- each command will have different associated "Steps" for the logs, so we're starting with figuring it out for `start`

enhancements/proposed/20200501-json-output/start-json-output.md
priyawadhwa: Thanks for doing that research @drewpca 

It seems like we could make the JSON struct themselves compatible by adding a few fields, which seems easy enough to do. We'll still print the JSON to stdout/stderr as planned.

cc @balopat WDYT?

enhancements/proposed/20200501-json-output/start-json-output.md
jtzwu: I would prefer the method shows in Drew's comment. Much easier to parse.

priyawadhwa:
closing for now, will reopen when i convert my google doc to markdown

Issue #7604, opened 2020-04-11
feature request: mount host volumes into docker driver via `docker -v`


ptramsey:
**Steps to reproduce the issue:** 

1. minikube config set vm-driver docker
2. minikube start

This is a feature request; there's no failed command here to document.  That said:

It would be really nice if the docker vm-driver allowed mapping paths from the local filesystem into the minikube container, using docker's volume binding functionality.  E.g.,

```
minikube  config set vm-driver docker
minikube start --docker-volume=/home/ptramsey:/home/ptramsey
```
... which would internally call something like,
```
docker run -v /home/ptramsey:/home/ptramsey gcr.io/k8s-minikube/kicbase:v0.0.8 /usr/local/bin/entrypoint
```

As it is, despite running without virtualization on linux (which has a noticeable performance advantage), we're stuck using nfs or 9p to mount host paths inside minikube, both of which are slower (the latter of which is much slower) than just mapping the path directly via docker.

priyawadhwa:
Hey @ptramsey thanks for opening this issue -- this seems like a reasonable feature request. If anybody would be interested in working on this, please comment here!

afbjorklund:
Currently we are mounting the `/var` directory as a docker volume, so that's the current workaround.

i.e. use this host directory, for getting things into the container ?

See e.g. `docker volume inspect minikube` for the details on it

Mounting the `/home` directory has security implications, some want it disabled by default for VMs...

See #6788

And #5012

ptramsey:
Having the ability to specify arbitrary directories that aren't /var, though, is useful, and I would be surprised if people aren't already depending on that functionality, since the hyperkit driver on macos has it.  Also, the virtualbox driver by default mounts /Users on macos and windows, and mounts /home  to a safe path on linux.

If people are concerned about mounting /home itself, to the extent that we think it should be completely disallowed (rather than just being a warning), it seems like this feature request could still be accomplished with that chnage.  After all, the feature request doesn't refer to mounting /home anyway --- the example it presents is of mounting a specific home directory (that isn't /home/minikube).

afbjorklund:
Agree, adding the possibility to add mounts in the docker/podman driver would be a nice feature.

AlbertMarashi:
bump

vincent-163:
Any progress on this? I stumbled upon this problem. It feels so hard to use Kubernetes local volumes without this feature.

Asarew:
[...]

PR #8136, opened 2020-05-13
Feature request #7604: use native docker bind mounts with driver-mounts option

g9rga:
Hello everyone! I added new feature for native driver mount **--driver-mounts** by request https://github.com/kubernetes/minikube/issues/7604. 
At this pr, only docker driver supported. Usage example:
`minikube start --driver=docker --driver-mounts=/tmp:/opt/tmp`

Here is test example output:
(⎈ |N/A:default)➜  minikube git:(fix-7604) mkdir /tmp/123
(⎈ |N/A:default)➜  minikube git:(fix-7604) echo "test" > /tmp/123/file.txt
(⎈ |N/A:default)➜  minikube git:(fix-7604) ./out/minikube-linux-amd64 start -p minikube-test --driver=docker --driver-mounts="/tmp/123:/tmp/123:rw"
😄  [minikube-test] minikube v1.12.1 on Arch 20.0.3
    ▪ KUBECONFIG=/home/eth/.kube/config
👍  Starting control plane node minikube-test in cluster minikube-test
🚜  Pulling base image ...
💾  Downloading Kubernetes v1.18.3 preload ...
    > preloaded-images-k8s-v4-v1.18.3-docker-overlay2-amd64.tar.lz4: 526.27 MiB
🏄  Done! kubectl is now configured to use "minikube-test"
(⎈ |minikube-test:default)➜  minikube git:(fix-7604) docker inspect -f '{{ .Mounts }}' minikube-test
[{bind  /lib/modules /lib/modules  ro false rprivate} {volume minikube-test /var/lib/docker/volumes/minikube-test/_data /var local z true } {bind  /tmp/123 /tmp/123   true rprivate}]
(⎈ |minikube-test:default)➜  minikube git:(fix-7604) docker exec minikube-test ls /tmp/123
file.txt
(⎈ |minikube-test:default)➜  minikube git:(fix-7604) docker exec minikube-test cat /tmp/123/file.txt                                     
test
(⎈ |minikube-test:default)➜  minikube git:(fix-7604) docker exec minikube-test sh -c "echo 123 > /tmp/123/file.txt"
(⎈ |minikube-test:default)➜  minikube git:(fix-7604) echo /tmp/123/file.txt 
/tmp/123/file.txt
(⎈ |minikube-test:default)➜  minikube git:(fix-7604) 

Screenshot_20200802_111301

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -105,6 +105,7 @@ const (
 	forceSystemd            = "force-systemd"
 	kicBaseImage            = "base-image"
 	startOutput             = "output"
+	driverMounts            = "driver-mounts"
 )
 
 // initMinikubeFlags includes commandline flags for minikube.
@@ -195,6 +196,11 @@ func initDriverFlags() {
 	startCmd.Flags().String(hypervVirtualSwitch, "", "The hyperv virtual switch name. Defaults to first found. (hyperv driver only)")
 	startCmd.Flags().Bool(hypervUseExternalSwitch, false, "Whether to use external switch over Default Switch if virtual switch not explicitly specified. (hyperv driver only)")
 	startCmd.Flags().String(hypervExternalAdapter, "", "External Adapter on which external switch will be created if no external switch is found. (hyperv driver only)")
+
+	// Driver specific mounts
+	startCmd.Flags().String(driverMounts, "",
+		`Use native driver mounts. Supported drivers:
+		Docker: -v flag(bind mounts). Has no effect in the existing docker container. Additional options rw and Z supported, for example --driver-mounts="/tmp:/tmp:rwZ"`)
 }
 
 // initNetworkingFlags inits the commandline flags for connectivity related flags for start
@@ -291,6 +297,7 @@ func generateClusterConfig(cmd *cobra.Command, existing *config.ClusterConfig, k
 			NFSSharesRoot:           viper.GetString(nfsSharesRoot),
 			DockerEnv:               config.DockerEnv,
 			DockerOpt:               config.DockerOpt,
+			DriverMounts:            viper.GetString(driverMounts),
 			InsecureRegistry:        insecureRegistry,
 			RegistryMirror:          registryMirror,
 			HostOnlyCIDR:            viper.GetString(hostOnlyCIDR),
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -609,3 +609,39 @@ func iptablesFileExists(ociBin string, nameOrID string) bool {
 	}
 	return true
 }
+
+// CreateMounts use for init mounts configuration, to use it in oci drivers like docker, podman
+func CreateMounts(mountString string) []Mount {
+	var mounts []Mount
+	isReadOnly := true
+	SelinuxRelabel := false
+
+	if mountString == "" {
+		return mounts
+	}
+
+	splittedMounts := strings.Split(mountString, ",")
+	for _, mountString := range splittedMounts {
+		mountPieces := strings.Split(mountString, ":")
+		if len(mountPieces) == 1 {
+			mountPieces = append(mountPieces, mountPieces[0])
+		}
+
+		if len(mountPieces) == 3 {
+			if strings.Contains(mountPieces[2], "rw") {
+				isReadOnly = false
+			}
+			if strings.Contains(mountPieces[2], "Z") {
+				SelinuxRelabel = true
+			}
+		}
+		mounts = append(mounts, Mount{
+			ContainerPath:  mountPieces[1],
+			HostPath:       mountPieces[0],
+			Readonly:       isReadOnly,
+			SelinuxRelabel: SelinuxRelabel,
+		})
+	}
+
+	return mounts
+}
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1045,6 +1045,13 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 		exit.UsageT("Sorry, please set the --output flag to one of the following valid options: [text,json]")
 	}
 
+	if cmd.Flags().Changed(driverMounts) && !driver.IsDocker(drvName) {
+		exit.UsageT(
+			"Sorry, provided flag --{{.flag_name}} doesn't supported by {{.name}} driver, use --mount-string instead",
+			out.V{"flag_name": driverMounts, "name": drvName},
+		)
+	}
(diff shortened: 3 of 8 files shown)


afbjorklund:
It would be nice if we could do something slightly more portable, instead of adding more docker-only flags ? Not _really_ looking forward to duplicating all of these for podman, like we did for docker-env

cmd/minikube/cmd/start_flags.go
```diff
@@ -101,6 +101,7 @@ const (
 	deleteOnFailure         = "delete-on-failure"
 	forceSystemd            = "force-systemd"
 	kicBaseImage            = "base-image"
+	dockerMounts            = "docker-mounts"
```
medyagh, line 104: any chance we use the same flag as the one we use for the VMs ?

g9rga: Do you mean --mount-string option or something else?

medyagh:
I agree with @afbjorklund , we should try to make this not docker secific. on these both PRs

I wonder if we should the existing "-mount-string" option on start? instead of making it docker-specific flag?

cmd/minikube/cmd/start_flags.go
medyagh: I am open to any solution that doesn't make it docker driver specifi.
--mount-string could be a solution 
or maybe --mount-option

g9rga:
@medyagh @afbjorklund thank you for your proposal. I refactored my PR to use existing option `mount-string`. On minikube start it will use native docker bind mount, instead of 9pfs(mount command). Is it ok?

cmd/minikube/cmd/start_flags.go
Asarew: I agree with @afbjorklund statements in the https://github.com/kubernetes/minikube/pull/8159#issuecomment-629603739.
The `--docker-mounts` has different attributes like the options for read-only, rslave, etc. i think adding this logic behind the `--mount-string` won't make the implementation more uniform as the options only applies for the container drivers.

ptramsey:
Hi!  Just wanted to say thank you for doing this work!  Is there anything I can do to help this along?  I'd love to see this landed!

g9rga:
@medyagh @afbjorklund what about **--driver-mounts** option? Format will depend on driver type.

medyagh:
@g9rga do you mind adding an example of using this command in the PR description so I could test it?
or is it possible to add an integration test for it?

sharifelgamal:
@g9rga Could you sign the CLA so we can proceed with testing?

g9rga:
> @g9rga Could you sign the CLA so we can proceed with testing?

I did, `cla/linuxfoundation — g9rga authorized`

ptramsey:
Just checking in here.  :) Looking forward to seeing this!

sharifelgamal:
@g9rga Can you fix the outstanding merge conflict? That way we can rerun the tests and make sure everything is ok.

g9rga:
> @g9rga Can you fix the outstanding merge conflict? That way we can rerun the tests and make sure everything is ok.
@sharifelgamal It fixed. Could you rerun tests?

medyagh:
I think the other Pr might be better, since it uses the already existing flag

https://github.com/kubernetes/minikube/pull/8159/files

pkg/drivers/kic/oci/oci.go
```diff
@@ -592,3 +592,38 @@ func iptablesFileExists(ociBin string, nameOrID string) bool {
 	}
 	return true
 }
+
+func CreateMounts(mountString string) []Mount {
```
medyagh, line 596: add comment for unexported funcs

cmd/minikube/cmd/start_flags.go
```diff
@@ -196,6 +197,11 @@ func initDriverFlags() {
 	startCmd.Flags().String(hypervVirtualSwitch, "", "The hyperv virtual switch name. Defaults to first found. (hyperv driver only)")
 	startCmd.Flags().Bool(hypervUseExternalSwitch, false, "Whether to use external switch over Default Switch if virtual switch not explicitly specified. (hyperv driver only)")
 	startCmd.Flags().String(hypervExternalAdapter, "", "External Adapter on which external switch will be created if no external switch is found. (hyperv driver only)")
```
medyagh, line 199: please add a check inside  validateFlags 
to ensure that if user provided this flag with a driver other than Docker it should tell the user this flag wont' work with their driver and they should use --mount-string instead for VM drivers.

medyagh:
@g9rga do you mind trying this PR and see if it works ? and put the output in the PR description ? I did a validation and it didn't seem to be working (if i put a file in the mounted folder I didn't see it in the host)

g9rga:
> @g9rga do you mind trying this PR and see if it works ? and put the output in the PR description ? I did a validation and it didn't seem to be working (if i put a file in the mounted folder I didn't see it in the host)

Did you try it with existing container? Due to docker architecture, mounts only works for new containers(.

cmd/minikube/cmd/start_flags.go
g9rga: Thank you! I did

pkg/drivers/kic/oci/oci.go
g9rga: Already added! :+1:

mszygenda:
Would be great if it supported "delegated" / "cached" / "consistent" mount flags which greatly improves performance on MacOS systems.

https://docs.docker.com/docker-for-mac/osxfs-caching/

wcjohnson:
Is there anything that I can do to help with this PR? This functionality is badly needed in the Docker driver. The `minikube mount` 9P file system is far too slow and unstable for anything more than trivial workloads.

ptramsey:
Is there anything that I could do to help this along? I hate to just be sitting on the sidelines cheering!  But I'd really love to see this landed.

wcjohnson:
@g9rga In order to make use of this PR, is it sufficient to just compile and replace the minikube binary?

g9rga:
> @g9rga In order to make use of this PR, is it sufficient to just compile and replace the minikube binary?

Yep, you can compile it with `make darwin/linux/windows` and run compiled binary, for example in my case:
`/home/minikube/out/minikube start --cpus=6 --kubernetes-version=v1.14.10 --driver=docker --memory=6G --driver-mounts="/home/g9rga/www:/home/g9rga/www:rw" --addons=ingress`

g9rga:
I added integration test for new option

tstromberg:
I believe this is obsoleted by #8159 - which does the same thing, but using the existing  --mount-string. My apologies for the confusion: I didn't realize there were two PR's out for the same feature.

Issue #8127, opened 2020-05-13
Add a command to list supported drivers


kschaab:
Need a way to call minikube to get a list of supported drivers on deployed system.  This information comes back from help but is extremely difficult to reliably parse via regex and looks to contain temporal content (like experimental):
`minikube start --help | grep -e "--driver='':"`
```      --driver='': Driver is one of: virtualbox, parallels, vmwarefusion, hyperkit, vmware, docker, podman (experimental) (defaults to auto-detect)```

This will aid us in fixing issue https://github.com/GoogleCloudPlatform/cloud-code-vscode/issues/234.

kschaab:
Would be really nice to also return the information from Suggest in pkg/minikube/driver/driver.go in the output from this command.

afbjorklund:
Someone was working on this (in #7679), but it seems to have stalled - we'll need to take over...

afbjorklund:
It's equally stupid in the code base now, but should evolve into something nice from registry.

afbjorklund:
@kschaab : the "options" command is like a side-effect of using the command framework. It's not really something that should be extended.

`Use "minikube options" for a list of global command-line options (applies to all commands).`

So this should probably use some kind of special JSON (or text?) output, similar to what we are doing for the progress output of `minikube start`

Like #8105

The easiest is probably doing it as an option to "start", even if that is a bit weird...

Another question is how much of the decision-making info we should dump
There's a side-project of building some kind of "smart" driver selection: #8012

So you could have reasons _why_ a driver is or is not being selected at the moment.

PR #8143, opened 2020-05-13
Add defaults subcommand to `minikube config`

priyawadhwa:
`minikube config defaults PROPERTY_NAME` will print out all acceptable default values for that property.

Currently only works for `minikube config defaults driver` which looks like this on Mac:

```
$ minikube config defaults driver
* virtualbox
* parallels
* vmwarefusion
* hyperkit
* vmware
* docker
* podman

$ minikube config defaults driver --output json
["virtualbox","parallels","vmwarefusion","hyperkit","vmware","docker","podman"]
```

Fixes https://github.com/kubernetes/minikube/issues/8127

cc @kschaab

### DIFF
--- a/cmd/minikube/cmd/config/defaults.go
+++ b/cmd/minikube/cmd/config/defaults.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+	"k8s.io/minikube/pkg/minikube/out"
 )
 
 var configDefaultsCommand = &cobra.Command{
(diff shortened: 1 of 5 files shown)


cmd/minikube/cmd/config/defaults.go
```diff
@@ -0,0 +1,92 @@
...
+	if output == "json" {
+		encoding, err := json.Marshal(defaults)
+		if err != nil {
+			return errors.Wrap(err, "encoding json")
+		}
+		out.Ln(string(encoding))
+		return nil
+	}
+	for _, d := range defaults {
+		out.Ln("* %s", d)
```
afbjorklund, line 74: It would be easier to parse the output, if you dropped the `* ` prefix.

priyawadhwa: I added a `--output json` flag in case anyone needs to parse the output; otherwise, I think the prefix makes for a nicer UI

PR #8158, opened 2020-05-15
experiment: add a single quote for docker/podman inspect formats

medyagh:
when trying to add windows to azure, I saw this be a source of problem where minikube created docker container but after when it was trying to inspect it couldnt work. manually I had to add a single quote. hopefully this doesn't break podman and others


medyagh:
nevermind.

PR #8159, opened 2020-05-15
Support for mounting host volumes on start with docker driver

Asarew:
To use this PR run the `minikube start` with the `--mount` option like so:

### Description:
```sh
$ minikube start --mount
```
this will automatically mount your home directory in the minikube host at path `/minikube-host`

Or specify a mount path:
```sh
$ minikube start --mount --mount-strinng "/dir/to/share:/dir/where/to/mount"
```
### Usage:
```sh
$ mkdir testdir
$ echo "hello" > testdir/world.txt
$ minikube start --mount --mount-string $(pwd)/testdir:/dir/testdir --driver=docker
😄  minikube v1.12.0-beta.0 on Linuxmint 19.2
🔥  Creating docker container (CPUs=8, Memory=10000MB) ...
🐳  Preparing Kubernetes v1.18.3 on Docker 19.03.2 ...
$ minikube ssh
docker@minikube:~$ cat /dir/testdir/world.txt
hello
docker@minikube:~$ echo "goodbye" > /dir/testdir/sky.txt
docker@minikube:~$ exit
$ cat testdir/sky.txt
goodbye
```
fixes: #7604


Asarew:
I signed it

RA489:
> I signed it

/check-cla

Asarew:
Don't know why but the cla-bot doesn't re-check my cla @RA489 any idea on what i should do next?

RA489:
> Don't know why but the cla-bot doesn't re-check my cla @RA489 any idea on what i should do next?

do you mind verifying that the e-mail address you have on GitHub matches the one you signed the CLA with?

Asarew:
Yeah, i double check that both commits have the correct author email address and that it's the same as my github email and the one i used to sign the CLA.

afbjorklund:
Same comment as with #8136, does this have to be docker-specific ?

Asarew:
Currently this is a docker specific things as and podman are the only drivers that would support this feature.
I do believe that the intended behaviour is similar like `--mount-string` but that would spawn the mount deamon, and i don't think we should overwrite that.

or are you talking about making the flag name not docker specific; something like `--volume-mount`.

afbjorklund:
> Currently this is a docker specific things as and podman are the only drivers that would support this feature.

We can copy it from docker-volume to podman-volume, just like with did with docker-env and podman-env. Just that it gets a bit boring after a while, with all the duplication and copy/paste.

Since 90% of the implementation is exactly the same, except some minor details like "sudo". And ultimately it would also be nice to align this with extra disk images for the hypervisors...

> or are you talking about making the flag name not docker specific; something like --volume-mount.

Yeah, but you can go ahead with this feature without it. We can copy/paste it, after this PR.

medyagh:
on these both PRs
@afbjorklund 
I wonder if we should the existing "-mount-string" option on start?

afbjorklund:
> we should try to make this not docker secific. on these both PRs

then again it already is, so this doesn't really make it _much_ worse

I mean since we already have "docker-env" and "docker-opt" :

```
      --docker-env=[]: Environment variables to pass to the Docker daemon. (format: key=value)
      --docker-opt=[]: Specify arbitrary flags to pass to the Docker daemon. (format: key=value)
```

It's easy enough to copy/paste these into podman-env/podman-opt.

Even with a new "docker-vol" (or whatever) flag added here as well.

Asarew:
@afbjorklund @RA489 I can add the `--podman-volume` flag. do you want this in this PR or should i create a new one?

medyagh:
@Asarew do you mind putting the PR description eample of using this?

tstromberg:
I don't love that this effectively duplicates `--mount-string` with a driver-specific flag, but I see there are some options as for the type of mount it supports.

tstromberg:
I like this PR. Can you make this use `--mount-string` instead of adding a new string? You could preserve the parsing & handling logic.

Asarew:
@medyagh i added an example in the PR description. can you give me some direction on why the CLA check is failing?

medyagh:
@Asarew  this PR needs the CLA signed, please check https://identity.linuxfoundation.org/projects/cncf

Asarew:
I signed the CLA and made sure that all commits have the correct email address as author.

Asarew:
@medyagh Figured out what i did wrong, CLA is signed!

medyagh:
> @medyagh Figured out what i did wrong, CLA is signed!

thank you :) I will re-review this PR tomorrow theoverall the PR looks good small minior things might need to change

pkg/minikube/config/types.go
```diff
@@ -43,6 +43,7 @@ type ClusterConfig struct {
 	HyperkitVpnKitSock      string   // Only used by the Hyperkit driver
 	HyperkitVSockPorts      []string // Only used by the Hyperkit driver
 	DockerEnv               []string // Each entry is formatted as KEY=VALUE.
+	VolumeMounts            []string
```
medyagh, line 46: this seems to be used only by Docker/Podman driver, change name to ContainerVolumeMounts and add comment that is used only by Docker/podman

pkg/minikube/registry/drvs/docker/docker.go
```diff
@@ -58,10 +58,20 @@ func init() {
 }
 
 func configure(cc config.ClusterConfig, n config.Node) (interface{}, error) {
+	mounts := make([]oci.Mount, len(cc.VolumeMounts))
```
medyagh, line 61: I believe we can add same exact code for podman driver. registry/drvs/podman/podman.go

medyagh:
@Asarew do you mind addressing the review comments and also update the PR descrtiption with the output of example of using it with vadlidaitons ? (mount a folder and then ls it inside minikube)

pkg/minikube/registry/drvs/docker/docker.go
Asarew: I agree, will do!

Asarew:
@medyagh I resolved all your comments, but the build is failing on some windows hyperv test. I don't see how this PR can impact those tests as there are clear checks that wouldn't allow any of these changes to run in the hyperv test.

sharifelgamal:
The Hyper-V test failures are not related to this PR at all. This is looking good.

/ok-to-test

Asarew:
@medyagh  i was just thinking that i should add a warning when you try to change the mount-string on start after the cluster has been created. Because docker doesn't allow changing of volumes after the container has been created.
Also i should probably disallow the `minikube mount` command with a message to use `--mount-string` on creation as it (currently) doesn't work anyway.

Asarew:
@medyagh @sharifelgamal @tstromberg is there anything i can do to get this merged asap?

tstromberg:
Done. Thank you for your patience! I like your idea about adding a warning if a mount is attempted after creation.

Asarew:
Thanks! I'll create a new PR in a couple of days for the warning(s). Do you know hen the next release is scheduled? So i can attempt to get it done before then.

tstromberg:
v1.13 will likely ship by tomorrow, with a v1.13.1 likely next Thursday to catch Kubernetes v1.19.1.

Asarew:
Warning added in: #9172 :+1:

ZaMaZaN4iK:
@Asarew Thanks for so nice addition. I am trying to test it a little bit and got the following error:
```
minikube start --driver=docker --mount-string "/media/storage/wallpaper_service/filtered_images:/data/wallpapers" --mount                                       
😄  minikube v1.13.1 on Fedora 32
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
k8s.io/minikube/cmd/minikube/cmd.runStart(0x2cd9c80, 0xc000613240, 0x0, 0x4)
        /app/cmd/minikube/cmd/start.go:216 +0x15d3
github.com/spf13/cobra.(*Command).execute(0x2cd9c80, 0xc000613200, 0x4, 0x4, 0x2cd9c80, 0xc000613200)
        /go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:846 +0x29d
github.com/spf13/cobra.(*Command).ExecuteC(0x2cdeb40, 0x0, 0x1, 0xc0005ede80)
        /go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x349
github.com/spf13/cobra.(*Command).Execute(...)
        /go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
k8s.io/minikube/cmd/minikube/cmd.Execute()
        /app/cmd/minikube/cmd/root.go:113 +0x929
main.main()
        /app/cmd/minikube/main.go:71 +0x11f
```
Any ideas?

Asarew:
@ZaMaZaN4iK you should run `$ minikube delete`

Alig1493:
`minikube start --mount --mount-string $(pwd)/data/db:/data/db --driver=docker`
Doesn't seem to work
minikube ssh has no contents of the mounted folder
image
whereas on the host, the folder has a file:
image

Asarew:
@Alig1493 

not 100 sure anymore on this as I'm not using minikube anymore but if i remember correctly you have also pass the `--mount` flag.

Alig1493:
> @Alig1493 not 100% sure anymore on this as I'm not using minikube anymore, but if i remember correctly you also have to pass the `--mount` flag.

Hey @Asarew thanks for responding regardless. As you can see in my comment I did pass the mount flag along with the mount-string args.

t1bb4r:
@Alig1493 see https://github.com/kubernetes/minikube/issues/10490

PR #8161, opened 2020-05-15
Add Ambassador Ingress controller as an addon

concaf:
This PR adds Ambassador (https://www.getambassador.io/) as an addon to Minikube. Also, a tutorial has been added on how to use the same.

The motivation behind adding Ambassador as an addon is mostly around improving the UX around installing and configuring Ambassador. Ambassador is installed via our operator which makes sure than the Ambassador installation is healthy and up to date at all times.

CC: @brucehorn @inercia

### DIFF
--- a/deploy/addons/ambassador/ambassador-operator-crds.yaml
+++ b/deploy/addons/ambassador/ambassador-operator-crds.yaml
@@ -1,0 +1,186 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: ambassadorinstallations.getambassador.io
+spec:
+  additionalPrinterColumns:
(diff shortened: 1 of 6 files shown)


medyagh:
@concaf this looks great ! thank you for adding this addon. do you mind sharing which driver and os did you try this on ? 

have you tried it on the docker driver ? 
I wonder if we need t open a port for it during container creation time for docker/podman drivers

concaf:
@medyagh @priyawadhwa sorry for the late reply!

> do you mind sharing which driver and os did you try this on ?
> have you tried it on the docker driver ?
I've tested this on the docker driver, on Fedora 30.

> I wonder if we need t open a port for it during container creation time for docker/podman drivers
Nope, no such requirement. This pretty much requires the same stuff as the ingress addons.

PR #8168, opened 2020-05-16
move node ip save farther down stack

sharifelgamal:
We need to capture the new IP address for a node as soon as it saved so that anything calling it will automatically get the new IP.


medyagh:
average on this PR is 2 seconds slower

Average time for minikube: 28.659565924000002
Average time for Minikube (PR 8168): 30.896566772

I wonder if that is related.

medyagh:
@sharifelgamal do you mind pulling upstream and let the test run one more time

sharifelgamal:
> average on this PR is 2 seconds slower
> 
> Average time for minikube: 28.659565924000002
> Average time for Minikube (PR 8168): 30.896566772
> 
> I wonder if that is related.

Unlikely, the times were totally sunk by the one run that was 33 seconds. We should do something slightly better with the timings, like run 5 and drop the highest and lowest number.

PR #8184, opened 2020-05-17
Fix missing cluster name in output of status command

radeksm:
When checking status of the cluster, in debug output there is a missing cluster name in one of the lines:

```
I0517 19:53:34.214936   13392 status.go:109] checking status of  ...
```

Fixed outout:
```
[radek@c8k20 ~]$ ./minikube-linux-amd64 --v=4 --alsologtostderr  status
I0517 19:36:18.739278    9787 mustload.go:64] Loading cluster: minikube
I0517 19:36:18.739584    9787 status.go:109] checking status of minikube ...
```


cmd/minikube/cmd/status.go
```diff
@@ -106,7 +106,7 @@ var statusCmd = &cobra.Command{
 
 		var statuses []*Status
 		for _, n := range cc.Nodes {
-			glog.Infof("checking status of %s ...", n.Name)
+			glog.Infof("checking status of %s ...", cname)
```
priyawadhwa, line 109: I think the purpose of this log is to print info about each node in the loop, instead of the entire cluster. Could you please move this log one line down and print out the value of `machineName` instead of `cname`?

radeksm: OK I see, now it makes more sense :)

Issue #7904, opened 2020-04-25
Upgrade Buildroot to 2020.02 LTS


afbjorklund:
The current release is nearing EOL. We should upgrade to the next LTS.
> Notice that the 2019.02.x series is now end of life unless somebody
> steps up to take over maintenance. Please migrate to the 2020.02 series
> instead which will be supported until April 2021.

"Buildroot 2020.02 released"
http://lists.busybox.net/pipermail/buildroot/2020-March/276231.html

"Buildroot 2019.02.11 released, 2019.02.x is EOL"
http://lists.busybox.net/pipermail/buildroot/2020-April/279523.html

This will feature upgrade Linux to 5.4 (~4.19~) and systemd to 244 (~240~)

Also upgrades Binutils to 2.32 (~2.30~) and GCC to 8.4.0 (~7.5.0~)
and the system Go compiler is bumped from 1.11.13 to 1.13.14

Also need to upgrade VirtualBox, to get guest additions support for Linux 5.4

afbjorklund:
The original CoreOS (Container Linux) is also reaching EOL, on May 26, 2020.

See http://coreos.com/os/eol/

So we might look at Fedora CoreOS for inspiration, or we might continue on our own.

The current ISO was changed from Boot2Docker-based to Buildroot-based, in 2016.

See #571 #692 (commit f0b7abf42e90e5a8b47237ae4524a130647fcbf8)

Now that we don't support rkt anymore, we could look at rearranging some things...

afbjorklund:
One thing we should do is to _downgrade_ the glibc toolchain headers, to be more compatible:

```diff
-BR2_TOOLCHAIN_HEADERS_AT_LEAST="5.4"
+BR2_TOOLCHAIN_HEADERS_AT_LEAST="4.0"
```

```
BR2_LINUX_KERNEL_VERSION="5.4.40"
```

This is so that we can start looking into running the ISO image also as a KIC image... #6942

`docker import - minikube < output/images/rootfs.tar`

` FATAL: kernel too old`

afbjorklund:
> One thing we should do is to _downgrade_ the glibc toolchain headers, to be more compatible:

> This is so that we can start looking into running the ISO image also as a KIC image... #6942

Out of scope for the 1.11 release, and not certain that we should use the _same_ defconfig

So work on making an image for KIC will be in that other issue, this will be only about the VM.

afbjorklund:
[...]

PR #8187, opened 2020-05-17
Buildroot 2020.02

afbjorklund:
The good news is that it actually compiles (at last)

`179M	out/minikube.iso`

The bad news is that it doesn't boot, but hangs...

minikube-buildroot

The upstream Buildroot 2020.02.x works just fine, so this is something in minikube config.
Possibly related to the initial file systems, like the initrd (rootfs) or the sysroot (tmpfs) ?

For #7904


afbjorklund:
Yup, seems like something went horrible wrong in the init.

It's where we copy the contents of the rootfs to the tmpfs:

```sh
#!/bin/sh
mkdir /sysroot
# the value 90% borrowed from tcl via boot2docker
mount -t tmpfs -o size=90% tmpfs /sysroot
# copy from rootfs, to be able to do switch_root(8)
tar -C / --exclude=sysroot -cf - . | tar -C /sysroot/ -xf -
```

```
�mount: /sysroot: wrong fs type, bad option, bad superblock on tmpfs, missing codepage or helper program, or other error.
```

And then, because of that, it dies in the final step:

```sh
exec /sbin/switch_root /sysroot /sbin/init "$@"
```

```
BusyBox v1.31.1 (2020-05-17 20:20:13 CEST) multi-call binary.

Usage: switch_root [-c CONSOLE_DEV] NEW_ROOT NEW_INIT [ARGS]

Free initramfs and switch to another root fs:
chroot to NEW_ROOT, delete all in /, move NEW_ROOT to /,
execute NEW_INIT. PID must be 1. NEW_ROOT must be a mountpoint.

	-c DEV	Reopen stdio to DEV after switch
[   24.079005] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000100
```

afbjorklund:
But the actual `mount` command is (pretty much) the same, from util-linux 2.35.1

So it must be something else that is different, most likely it's the kernel version ?

afbjorklund:
Duh, sometimes it is the simple stuff. Like checking if the plug is actually in the wall socket.

Or making sure that tmpfs support is actually in the kernel... (maybe a default changed ?)

At least it boots:

buildroot-booting

But ssh is sad.

afbjorklund:
`ssh-keygen` is segfaulting (when generating the host keys). That is probably bad...

The package was upgraded from 7.9p1 to 8.1p1, maybe something happened to it.

afbjorklund:
Well, ssh starts now. But unfortunately we don't have any network (besides lo)

`vboxguest: PCI device not found, probably running on physical hardware`

`lspci: /sys/bus/pci/devices: No such file or directory`

Apparently `CONFIG_VIRTIO_PCI` went missing along with `CONFIG_PCI`.

afbjorklund:
Good, now `ssh` is back on track as well. Maybe it will start now ?

afbjorklund:
Well, it's always the little things.. No DNS, and other parts missing.

`command_runner.go:102] ! sudo: pgrep: command not found`

afbjorklund:
The missing /etc/resolv.conf was for the same root cause as ssh.

Systemd didn't like LibreSSL, and I hadn't rebuilt everything...

`minikube systemd-resolved[23193]: /usr/lib/systemd/systemd-resolved: error while loading shared libraries: libcrypto.so.45: cannot enable executable stack as shared object requires: Operation not permitted`

It _requires_ the libraries to be built with  `-Wa,--noexecstack`

`MemoryDenyWriteExecute=yes`

----

And the missing access of k8s.gcr.io was for a different reason:

`command_runner.go:102] ! bash: curl: command not found`

afbjorklund:
@medyagh : was there a way to share these "snapshot" ISO, or to make the PR build it ?

afbjorklund:
Probably should try to get upstream to _upgrade_ sysdig, rather than patching the old version.

I don't really know what we are using it for in minikube, but it was enabled in the old config...

medyagh:
> @medyagh : was there a way to share these "snapshot" ISO, or to make the PR build it ?

hm... there should be a way to at leasst have the ISO URL for manual testing

PR #8199, opened 2020-05-18
Wait for kubernetes components on soft start

priyawadhwa:
I noticed that TestComponentHealth/parallel/ComponentHealth was failing with this error:

```
Error apiserver status: https://172.17.0.3:8441/healthz returned error 500:
[-]etcd failed: reason withheld
```

but by the time post mortem logs were printed the etcd container was up and running.

I think this test occasionally fails because apiserver healthz is not yet returning a 200 status when we run the test. We wait for healthz to return 200 on regular start, but not on soft start, which we run in `TestFunctional`.

With this PR, the --wait flag will be respected on soft start, so now we can be sure the apiserver is healthy before running this test

closes https://github.com/kubernetes/minikube/issues/8014


priyawadhwa:
@medyagh `--wait=all` does wait for apiserver, but not on soft starts (if the config hasn't changed, we just skip that part of the code). I'll add change that.

PR #8208, opened 2020-05-19
Avoid image builders to hang on registry port 443

nicolaferraro:
This adds a mapping for the 443 port on the registry, because image publishers, in case of insecure registries, first try to contact port 443, then fallback on port 80.

But if port 443 is not mapped to the container (even on a closed port of the container, as in this case), then the application remains stuck until a timeout occurs. This in the end leads to really slow times for publishing an image.

This is happening e.g. in https://github.com/GoogleContainerTools/kaniko/issues/1013


medyagh:
thank you for fixing this, do u mind sharing the output that this fixes it ?

Issue #8192, opened 2020-05-18
docker: invalid reference format


Manvi07:
**Steps to reproduce the issue:** 

1. curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube_latest_amd64.deb
2. sudo dpkg -i minikube_latest_amd64.deb
3. minikube start **or**
    minikube start --driver=docker **or** 
    minikube start --driver=docker --base-image= gcr.io/k8s-minikube/kicbase:v0.0.10

**Full output of failed command:** 

> 😄  minikube v1.10.1 on Ubuntu 18.04
> ✨  Using the docker driver based on existing profile
> 🆕  Kubernetes 1.18.2 is now available. If you would like to upgrade, specify: --kubernetes-version=1.18.2
> 🆕  Kubernetes 1.18.2 is now available. If you would like to upgrade, specify: --kubernetes-version=1.18.2
> 👍  Starting control plane node minikube in cluster minikube
> 🤷  docker "minikube" container is missing, will recreate.
> E0518 19:32:57.136815   27695 oci.go:79] docker daemon seems to be stuck. Please try restarting your docker. Will try to delete anyways: unknown state "minikube": docker inspect minikube --format={{.State.Status}}: exit status 1
> stdout:
> 
> 
> stderr:
> Template parsing error: template: :1:8: executing "" at <.State.Status>: map has no entry for key "State"
> 🔥  Creating docker container (CPUs=2, Memory=2200MB) ...
> 🤦  StartHost failed, but will try again: recreate: creating host: create: creating: create kic node: create container: docker run -d -t --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --label mode.minikube.sigs.k8s.io=minikube --volume minikube:/var --cpus=2 --memory=2200mb -e container=docker --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 --publish=127.0.0.1::5000 : exit status 125
> stdout:
> 
> stderr:
> docker: invalid reference format.
> See 'docker run --help'.
> 
> 🤷  docker "minikube" container is missing, will recreate.
[...]

PR #8229, opened 2020-05-21
kic: explicitly provide the type in inspect commands

afbjorklund:
Should avoid errors like: map has no entry for key "State"

When inspecting a volume, rather than a container or image

Should help with #8192

Reviewer note: also fixed a typo in an internal function name

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -572,7 +572,7 @@ out/docker-machine-driver-kvm2-aarch64: out/docker-machine-driver-kvm2-arm64
 
 out/docker-machine-driver-kvm2-%:
 ifeq ($(MINIKUBE_BUILD_IN_DOCKER),y)
-	docker inspect -f '{{.Id}} {{.RepoTags}}' $(KVM_BUILD_IMAGE) || $(MAKE) kvm-image
+	docker image inspect -f '{{.Id}} {{.RepoTags}}' $(KVM_BUILD_IMAGE) || $(MAKE) kvm-image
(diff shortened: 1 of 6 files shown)


afbjorklund:
Here is the issue we are trying to address:

```
 docker run -d --name foo -d busybox true
c4e9d99e7c959014b5e0b6c543bc4ca5ff54475b6e79dbc2c717a2fff304971d
 docker inspect foo --format={{.State.Status}}
exited
 docker container inspect foo --format={{.State.Status}}
exited
 docker volume create foo
 docker rm foo
 docker inspect foo --format={{.State.Status}}

Template parsing error: template: :1:8: executing "" at <.State.Status>: map has no entry for key "State"
 docker container inspect foo --format={{.State.Status}}

Error: No such container: foo
```

That is, when using "inspect" it will match the volume name too. So we add "container" explicitly.

This seems to be a common problem at minikube start, when it tries to re-create the container...

medyagh:
this is a great improvment ! thank you @afbjorklund do you know if this "type" inspect is supported in older docker versions? or is there a min version that we would require for this change?

afbjorklund:
> this is a great improvment ! thank you @afbjorklund do you know if this "type" inspect is supported in older docker versions? or is there a min version that we would require for this change?

It's available in 1.13 / 17.03, I don't think we need to support any older than that...

medyagh:
> > this is a great improvment ! thank you @afbjorklund do you know if this "type" inspect is supported in older docker versions? or is there a min version that we would require for this change?
> 
> It's available in 1.13 / 17.03, I don't think we need to support any older than that...

sounds good

Issue #8185, opened 2020-05-17
docker-env stops working after IP change


medyagh:
this is on docker-driver but it could be true on podman and kvm as well.
```
med@xmac:~/workspace/minikube (windows_gh_action)$ eval $(minikube docker-env); docker images
error during connect: Get https://172.17.0.2:2376/v1.40/images/json: x509: certificate is valid for 172.17.0.3, 127.0.0.1, not 172.17.0.2
```
### replicate
```
#!/bin/bash
for i in {1..5}
do
   echo "try $i"
   minikube start -p minikube
   minikube start -p p1
   minikube start -p p2
   eval $(minikube docker-env -p minikube)
   docker images
   eval $(minikube docker-env -p p1)
   docker images
   eval $(minikube docker-env -p p2)
   docker images
   minikube stop -p minikube & minikube stop -p p1 & minikube stop -p p2
   systemctl restart docker
done
```

### solution

restaring the docker inside minikube makes docker to pick the certs.

afbjorklund:
I think this got broken in 1.9.0, need to verify that with 1.8.2

Issue #8235, opened 2020-05-21
Automatic restart is hiding the real error


afbjorklund:
We get a lot of reports, that include a partial log from the restart - but nothing from the inital start.

It would be nice to improve this, to either fail on the first error or show how to report all logs ?

Typical report:

:sparkles: Using the docker driver based on existing profile
:shrug: docker "minikube" container is missing, will recreate.

* Using the virtualbox driver based on existing profile
* Restarting existing virtualbox VM for "minikube" ...

The details from the first start, that shows why it fails, is not included in the bug report or logs.

medyagh:
I was thinking we should by default write your logs for the current start to ~/.minikube/logs.txt or something, so whenever user does not run with --aloslogtostderr we should be able to ask them just pass ~/.minikube/logs.txt

tstromberg:
I believe this has been addressed. If someone sees this again, please re-open with an example.

PR #8239, opened 2020-05-21
docker-env: restart dockerd inside minikube on failure

medyagh:
Before this PR when running docker-env, if docker service is not running we exit with error, 
for example if minikube IP changes, the docker inside minikube needs to be restart to pick up the certs that have the new IP added to them.

closes https://github.com/kubernetes/minikube/issues/8185

This PR is meant for the last effort, when user reaches out to docker-env, the root cause can be fixed by a different PR. that will be for another release.

### Before this PR: docker env when IP changed after minikube start
```
$ eval $(minikube -p p1 docker-env)

$ docker ps
error during connect: Get https://172.17.0.2:2376/v1.40/containers/json: x509: certificate is valid for 172.17.0.3, 127.0.0.1, not 172.17.0.2

```
### After this PR: docker env when IP changed after minikube start

```
$ eval $(./out/minikube -p p1 docker-env)
$ docker ps
CONTAINER ID        IMAGE                  COMMAND                  CREATED             STATUS                  PORTS               NAMES
e81df52d2ad1        303ce5db0e90           "etcd --advertise-cl…"   1 second ago        Up Less than a second                       k8s_etcd_etcd-p1_kube-system_ce05cdf6ecce4bf90b96c7f248feb6a3_1
2f979b669a97        k8s.gcr.io/pause:3.2   "/pause"                 2 seconds ago       Up Less than a second                       k8s_POD_coredns-66bff467f8-2zfbs_kube-system_493e5a5f-fd51-424a-83b2-f515607de399_7
bec03d29f012        k8s.gcr.io/pause:3.2   "/pause"                 2 seconds ago       Up Less than a second                       k8s_POD_kube-scheduler-p1_kube-system_155707e0c19147c8dc5e997f089c0ad1_7
86a7cda645da        k8s.gcr.io/pause:3.2   "/pause"                 2 seconds ago       Up Less than a second                       k8s_POD_kube-proxy-nv2gh_kube-system_9dd0c993-ccc8-4f6e-9b69-673a53ed3572_7
e23ba5d4b230        k8s.gcr.io/pause:3.2   "/pause"                 2 seconds ago       Up Less than a second                       k8s_POD_kube-controller-manager-p1_kube-system_7f415a35d57cff5428871c5a51313bd5_7
cd8534425555        k8s.gcr.io/pause:3.2   "/pause"                 2 seconds ago       Up Less than a second                       k8s_POD_etcd-p1_kube-system_ce05cdf6ecce4bf90b96c7f248feb6a3_4
```

### DIFF
--- a/cmd/minikube/cmd/docker-env.go
+++ b/cmd/minikube/cmd/docker-env.go
@@ -151,6 +151,11 @@ var dockerEnvCmd = &cobra.Command{
 			Shell: shell.ForceShell,
 		}
 
+		if ok := isDockerActive(co.CP.Runner); !ok {
+			glog.Warningf("dockerd is not avtive will try to restart it...")
+			restartOrExitDaemon("docker", cname, co.CP.Runner)
+		}
+
 		var err error
 		port := constants.DockerDaemonPort
 		if driver.NeedsPortForward(driverName) {
@@ -177,12 +182,7 @@ var dockerEnvCmd = &cobra.Command{
 			}
 		}
 
-		if ok := isDockerActive(co.CP.Runner); !ok {
-			glog.Warningf("dockerd is not avtive will try to restart it...")
-			restartOrExitDaemon("docker", cname, co.CP.Runner)
-		}
-
-		out, err := tryConnectivity("docker", ec)
+		out, err := tryDockerConnectivity("docker", ec)
 		if err != nil { // docker might be up but been loaded with wrong certs/config
 			if strings.Contains(err.Error(), "x509: certificate is valid") {
 				glog.Infof("dockerd inside minkube is loaded with old certs with wrong IP. output: %s error: %v", string(out), err)
@@ -273,8 +273,8 @@ func dockerEnvVarsList(ec DockerEnvConfig) []string {
 	}
 }
 
-// tryConnectivity will try to connect to docker env from user's POV to detect the problem if it needs reset or not
-func tryConnectivity(bin string, ec DockerEnvConfig) ([]byte, error) {
+// tryDockerConnectivity will try to connect to docker env from user's POV to detect the problem if it needs reset or not
+func tryDockerConnectivity(bin string, ec DockerEnvConfig) ([]byte, error) {
 	c := exec.Command(bin, "version", "--format={{.Server}}")
 	c.Env = append(os.Environ(), dockerEnvVarsList(ec)...)
 	return c.CombinedOutput()
--- a/cmd/minikube/cmd/docker-env.go
+++ b/cmd/minikube/cmd/docker-env.go
@@ -24,8 +24,10 @@ import (
 	"io"
 	"net"
 	"os"
+	"os/exec"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/golang/glog"
 	"github.com/spf13/cobra"
@@ -120,6 +122,12 @@ func isDockerActive(r command.Runner) bool {
 	return sysinit.New(r).Active("docker")
 }
 
+func restartOrExitDaemon(bin string, name string, runner command.Runner) {
+	if err := sysinit.New(runner).Restart(bin); err != nil {
+		exit.WithCodeT(exit.Unavailable, `The {{.service_name}} service within '{{.name}}' is not active`, out.V{"name": name, "service_name": bin})
+	}
+}
+
 // dockerEnvCmd represents the docker-env command
 var dockerEnvCmd = &cobra.Command{
 	Use:   "docker-env",
@@ -137,16 +145,6 @@ var dockerEnvCmd = &cobra.Command{
 		if co.Config.KubernetesConfig.ContainerRuntime != "docker" {
 			exit.WithCodeT(exit.BadUsage, `The docker-env command is only compatible with the "docker" runtime, but this cluster was configured to use the "{{.runtime}}" runtime.`,
 				out.V{"runtime": co.Config.KubernetesConfig.ContainerRuntime})
-		}
-
-		// on minikube stop, or computer restart the IP might change.
-		// reloads the certs to prevent #8185
-		if err := sysinit.New(co.CP.Runner).Restart("docker"); err != nil {
-			glog.Warningf("failed to start the dockerd withhin %q minikube node: %v", cname, err)
-		}
-
-		if ok := isDockerActive(co.CP.Runner); !ok {
-			exit.WithCodeT(exit.Unavailable, `The docker service within '{{.name}}' is not active`, out.V{"name": cname})
 		}
 
 		sh := shell.EnvConfig{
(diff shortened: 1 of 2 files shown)


cmd/minikube/cmd/docker-env.go
```diff
@@ -138,6 +139,12 @@ var dockerEnvCmd = &cobra.Command{
 				out.V{"runtime": co.Config.KubernetesConfig.ContainerRuntime})
 		}
 
+		// on minikube stop, or computer restart the IP might change.
+		// reloads the certs to prevent #8185
+		if err := sysinit.New(co.CP.Runner).Restart("docker"); err != nil {
```
tstromberg, line 144: I don't think this is the right place to restart Docker, mostly because it is very surprising behavior. If you really must restart within `docker-env` calls, please update the help documentation accordingly.

My preference would be to see this restart done elsewhere, such as `fix.go` or `provision.go`:

* `provision` already restarts Docker if the systemd configuration has changed.
* folks may be trying to access the docker daemon outside of the `docker-env` command

One thought: if you include the IP address in the Docker config template, it will magically restart if the IP or Docker configuration changes. We could also do this more explicitly, of course.

cmd/minikube/cmd/docker-env.go
```diff
@@ -138,6 +139,12 @@ var dockerEnvCmd = &cobra.Command{
...
+			glog.Warningf("failed to start the dockerd withhin %q minikube node: %v", cname, err)
```
tstromberg, line 145: * `restart` instead of `start`
* `within`

cmd/minikube/cmd/docker-env.go
medyagh: I refactored this PR since this review

afbjorklund:
I agree that this behaviour is surprising. I was OK with starting dockerd if running `docker-env' on a node with only containerd, but maybe it should just fail in this scenario...

I don't think docker-machine has this issue, because the daemon is normally started with the VM and the certificates are regenerated by the provisioner when it starts up.


It only warns about the address (minikube ip) being changed:
```
Restarting "default"...
(default) Check network to re-create if needed...
(default) Waiting for an IP...
Waiting for SSH to be available...
Detecting the provisioner...
Restarted machines may have new IP addresses. You may need to re-run the `docker-machine env` command.
```

Then there is `reconfigure-certs` to run `configureAuth()`.

----

I think the main reason we need this in minikube is because of 5cd7660d1e49089155bb3c2b59ae25430e7eb7b5

That's were we hard-coded the Provisioner and removed ConfigureAuth:

```diff
-               glog.Infof("Detecting provisioner ...")
-               provisioner, err := provision.DetectProvisioner(h.Driver)
-               if err != nil {
-                       return h, errors.Wrap(err, "detecting provisioner")
```

```diff
-       glog.Infof("Configuring auth for driver %s ...", h.Driver.DriverName())
-       if err := h.ConfigureAuth(); err != nil {
-               return h, &retry.RetriableError{Err: errors.Wrap(err, "Error configuring auth on host")}
-       }
```

This is also causing problems with some of my other later experiments,
such as using Ubuntu for ISO or Buildroot for KIC - it doesn't work now.

cmd/minikube/cmd/docker-env.go
afbjorklund: As noted in #8235, it is getting very hard to debug all these auto-restart functions

medyagh:
> e we hard-coded the Provisioner and removed Confi

so the problem is, when you reconfigureAuth (that was refactored out of the code by thomas) it takes 5-6 times on each start (un-needed tax on minikube start)
even if the IP was not changed
not sure what is the best solution, my PR tries to connect to docker-env from User's point of view ( adding the docker env to a command and try to docker version on it) if the IP has changed or there any other errors , it will just restart the docker service to give it a try rat

tstromberg:
> so the problem is, when you reconfigureAuth (that was refactored out of the code by thomas) it takes 5-6 times on each start (un-needed tax on minikube start)
> even if the IP was not changed

Unless I am missing something, the IP only changes if `minikube start` has been run. If the IP changes, the kubelet needs to be restarted anyways. To minimize the number of issues, Docker should be restarted if it's in a broken state as close to the time that it comes up as possible.

cmd/minikube/cmd/docker-env.go
```diff
@@ -172,6 +182,21 @@ var dockerEnvCmd = &cobra.Command{
...
+			if strings.Contains(err.Error(), "x509: certificate is valid") {
+				glog.Infof("dockerd inside minkube is loaded with old certs with wrong IP. output: %s error: %v", string(out), err)
+			} else {
+				glog.Warningf("couldn't connect to docker inside minikube. output: %s error: %v", string(out), err)
+			}
+			// on minikube stop, or computer restart the IP might change.
+			// reloads the certs to prevent #8185
+			glog.Infof("will try to restart dockerd service...")
+			restartOrExitDaemon("docker", cname, co.CP.Runner)
+			// TODO: use kverify to wait for apisefver instead #8241
```
tstromberg, line 196: I don't understand this TODO. The apiserver should not be a requirement to push images to the Docker daemon. Otherwise docker push won't work on paused clusters.

cmd/minikube/cmd/podman-env.go
```diff
@@ -117,7 +118,8 @@ var podmanEnvCmd = &cobra.Command{
 		}
 
 		if ok := isPodmanAvailable(co.CP.Runner); !ok {
-			exit.WithCodeT(exit.Unavailable, `The podman service within '{{.cluster}}' is not active`, out.V{"cluster": cname})
+			glog.Warningf("podman service inside minikube is not active will try to restart it...")
+			restartOrExitDaemon("podman", cname, co.CP.Runner)
```
tstromberg, line 122: There is no podman daemon

cmd/minikube/cmd/docker-env.go
```diff
@@ -119,6 +122,12 @@ func isDockerActive(r command.Runner) bool {
 	return sysinit.New(r).Active("docker")
 }
 
+func restartOrExitDaemon(bin string, name string, runner command.Runner) {
```
tstromberg, line 125: nit: For idiomatic go, functions which may cause an exit should be named something like `mustRestartDocker`.

tstromberg:
PR description could use a mention of:

* What the problem is that you are trying to solve
* Why this is the best way to fix it

cmd/minikube/cmd/docker-env.go
medyagh: done

cmd/minikube/cmd/docker-env.go
```diff
@@ -172,6 +182,21 @@ var dockerEnvCmd = &cobra.Command{
 			}
 		}
 
+		out, err := tryDockerConnectivity("docker", ec)
+		if err != nil { // docker might be up but been loaded with wrong certs/config
+			if strings.Contains(err.Error(), "x509: certificate is valid") {
+				glog.Infof("dockerd inside minkube is loaded with old certs with wrong IP. output: %s error: %v", string(out), err)
```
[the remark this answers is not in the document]
medyagh: done

cmd/minikube/cmd/docker-env.go
medyagh: this was to pass the integraiton tests in functional test. because right after docker-env we do other tests we wanna give back a healthy cluster.

cmd/minikube/cmd/docker-env.go
```diff
@@ -172,6 +182,17 @@ var dockerEnvCmd = &cobra.Command{
...
 
+		out, err := tryDockerConnectivity("docker", ec)
+		if err != nil { // docker might be up but been loaded with wrong certs/config
+			glog.Warningf("couldn't connect to docker inside minikube. output: %s error: %v", string(out), err)
+			// to fix issues like this #8185
+			glog.Infof("will try to restart dockerd service...")
+			mustRestartDocker(cname, co.CP.Runner)
+			// TODO #8241: use kverify to wait for apisefver instead
+			// waiting for the basics like api-server to come up
+			time.Sleep(time.Second * 3)
```
tstromberg, line 193: Remove this sleep until it can be proved that this is helpful.

cmd/minikube/cmd/docker-env.go
```diff
@@ -172,6 +182,17 @@ var dockerEnvCmd = &cobra.Command{
 			}
 		}
 
+		out, err := tryDockerConnectivity("docker", ec)
+		if err != nil { // docker might be up but been loaded with wrong certs/config
+			glog.Warningf("couldn't connect to docker inside minikube. output: %s error: %v", string(out), err)
+			// to fix issues like this #8185
+			glog.Infof("will try to restart dockerd service...")
```
tstromberg, line 189: Can this be combined with the warning message? Users won't see the info message by default, leaving them with a scary warning and no explanation that we are trying to resolve it.

medyagh: done

cmd/minikube/cmd/docker-env.go
medyagh: done, but I am sure ppl will run into this, specially people who use this as part of another tool chain unvisible to the user

PR #8251, opened 2020-05-22
WIP: Upgrade kindbase to later version and ubuntu:20.04

afbjorklund:
Closes #8250 #8766 #8767

PR #8262, opened 2020-05-24
WIP: Test using the new ubuntu:focal based image

afbjorklund:
From https://github.com/kubernetes/minikube/pull/8251

Still not sure about the process, so doing an experiment

NOT TO BE MERGED

### DIFF
--- a/pkg/drivers/kic/types.go
+++ b/pkg/drivers/kic/types.go
@@ -24,9 +24,9 @@ import (
 
 const (
 	// Version is the current version of kic
-	Version = "v0.0.10"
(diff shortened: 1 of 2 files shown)


afbjorklund:
Wonder what this is ?

```
md5 signature computed for local file (kEBedcy3RfK6eWvNOzbvsw==) doesn't match cloud-supplied digest (3629n+VQYkGirESsY7w4cQ==). Cloud object (gs://minikube-builds/8262/minikube-linux-amd64) will be deleted.
ResumableUploadAbortException: 400 Bad Request
```

Looks like a timeout

PR #8275, opened 2020-05-26
Add KubeVirt addon

### LINKED ISSUE #2178: Add a KubeVirt add-on
KubeVirt provides adds a virtualization API to Kubernetes in order to launch regular virtual machines.

This request is about adding KubeVirt as an add-on to minikube, based on the work in https://github.com/kubevirt/kubevirt/pull/562.

Once this PR is merged, we can import manifests from the official releases in order to deploy KubeVirt in minikube.

ashleyschuett:
### Description 

This PR fixes #2178 

It adds the KubeVirt operator as an addon that can be enabled, as well as allowing users to separately create an instance of the KubeVirt custom resource to deploy the full environment.

### DIFF
--- a/deploy/addons/kubevirt/README.md
+++ b/deploy/addons/kubevirt/README.md
@@ -1,0 +1,4 @@
+## kubevirt Addon
+[kubevirt](https://kubevirt.io/) - KubeVirt technology addresses the needs of development teams that have adopted or want to adopt Kubernetes but possess existing Virtual Machine-based workloads that cannot be easily containerized.
+
+More info can be found by reading the tutorial ["How to use KubeVirt with minikube"](https://minikube.sigs.k8s.io/docs/tutorials/kubevirt/)
--- a/deploy/addons/kubevirt/pod.yaml.tmpl
+++ b/deploy/addons/kubevirt/pod.yaml.tmpl
@@ -1,0 +1,77 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: kube-system
+  name: kubevirt-scripts
+  labels:
+    kubernetes.io/minikube-addons: kubevirt
+    addonmanager.kubernetes.io/mode: Reconcile
+data:
+  uninstall.sh: |
+    #!/bin/bash
+
+    kubectl delete -f /manifests/kubevirt.yaml
+
+    kubectl delete -f /manifests/kubevirt-base.yaml
+
+  install.sh: |
+    #!/bin/bash
+
+    export KUBEVIRT_VERSION=$(curl -s https://api.github.com/repos/kubevirt/kubevirt/releases | grep tag_name | grep -v -- - | sort -V | tail -1 | awk -F':' '{print $2}' | sed 's/,//' | xargs)
+    echo $KUBEVIRT_VERSION
+
+    curl -Ls "https://github.com/kubevirt/kubevirt/releases/download/${KUBEVIRT_VERSION}/kubevirt-operator.yaml" -o /manifests/kubevirt-base.yaml
+
+    kubectl create -f /manifests/kubevirt-base.yaml
+
+    EMULATION=$(egrep 'svm|vmx' /proc/cpuinfo)
+    if [ -z "$EMULATION" ]; then
+      echo "use emulation"
+      kubectl create configmap kubevirt-config -n kubevirt --from-literal debug.useEmulation=true
+    fi;
+
+    curl -sL "https://github.com/kubevirt/kubevirt/releases/download/${KUBEVIRT_VERSION}/kubevirt-cr.yaml" -o /manifests/kubevirt.yaml
+
+    kubectl create -f /manifests/kubevirt.yaml
+
+    sleep infinity
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    kubernetes.io/minikube-addons: kubevirt
+    addonmanager.kubernetes.io/mode: Reconcile
+  name: kubevirt-install-manager
+  namespace: kube-system
+spec:
+  containers:
+  - command:
+    - /bin/bash
+    - -c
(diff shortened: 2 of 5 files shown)


deploy/addons/kubevirt/README.md
```diff
@@ -0,0 +1,45 @@
+## kubevirt Addon
```
medyagh, line 1: thank you for adding this readme, could u please this read me to our site section ?
maybe under tutorials ?

ashleyschuett: Hey, I added the readme under the tutorials section.  Would you like the readme in the deploy  folder to be removed, or is it okay for it to live in both places?

deploy/addons/kubevirt/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+In a minute or so kubevirt's default components will be installed into your cluster.
+You can run `kubectl get pods -n kubevirt` to see the progress of the kubevirt installation.
+
+
+### Disable kubevirt
+To disable this addon, simply run:
+```shell script
+minikube addons disable kubevirt
+minikube addons disable kubevirt-provisioner
```
fabiand, line 44: @ashleyschuett what about sqashing these two commands into one (just kubevirt).
If somebody wants to separate operator deployment from kubevirt deployment then the user can just do  it manually using the yamls.

site/content/en/docs/tutorials/kubevirt.md
```diff
@@ -0,0 +1,60 @@
...
+minikube addons enable kubevirt-provisioner
+```
+
+You will then need to check if your VM’s CPU supports virtualization extensions, you can check by executing the following command:
+```shell script
+minikube ssh -p kubevirt "egrep 'svm|vmx' /proc/cpuinfo"
+```
+If the command does not generate any output, create the following ConfigMap so that KubeVirt uses emulation mode, otherwise skip to the next section:
+```shell script
+kubectl create configmap kubevirt-config -n kubevirt --from-literal debug.useEmulation=true
```
fabiand, line 38: As discussed:

We can consider doing this with two options:

kubevirt -- regular
kubevirt -with-emmulation -- if emualation is needed to workaround missing kvm

deploy/addons/kubevirt/README.md
medyagh: thank you we can have a readme in the addon folder but with a link to the website, so anyone goes to that can find out where is the tutorial for this addon.

site/content/en/docs/tutorials/kubevirt.md
```diff
@@ -0,0 +1,60 @@
+---
+title: "KubeVirt support"
```
medyagh, line 2: I would change this to a 
How To... format, for example how to use KubeVirt with minikube

deploy/addons/kubevirt/README.md
ashleyschuett: After looking into this a bit and trying to get this to work I am unable to. 

The issue is if we enable kubevirt with a single command, it then also needs to be able to be disabled. When disabling all the manifests get removed at the same time. In this case, the operator deletes much quicker than the KubeVirt CR. When this happens the CR will still have a finalizer and will get stuck in a terminating state along with the namespace since the operator is gone and unable to remove it. 

I attempted to hook into the "preStop" lifecycle of the container but while the pod will then wait to delete it will still be in the "terminating" state meaning that the validating webhook will not respond and the kubevirt cr will not delete. 

Because of this, I think the options are to keep the deployments/commands as is. Or if we change them to be `kubevirt` and `kubevirt-with-emulation` then in the read me we have the user create the kubevirt CR, and let them know to remove it before disabling the addon. Does either of the options sound okay? Or is there another way to get around the previously mentioned issue?

fabiand: Nice catch. And sad.

Now, I dislike to say it, but what about writing a pod which does all this:
1. Deployment on creation and start
2. Undeploy on deletion of the pod

The reason that I think that this could work is, because we can set a terminationGrace period on the pod - which gives the pod some time to complete - and inside the pod we can just serialize the deleition of the CR and the followed deletion of the operator.

What do you think?

ashleyschuett: I think what you mean by this is that the enable/disable would only be deploying a pod, which would then have / run a script that would match the minikube quick start documentation we have?  Then on disable delete the resources in the proper order? 

The only problem I see what that is that namespace would still need to be created first in the manifests, and it would then also be deleted first which would mean that the operator deployment could be deleted before the pod runs the deletion. 

I have changed the minikube code to delete resources in the opposite that they are created in, but i think that would need buy in from more people in this project.

ashleyschuett: I have update the addon to use your suggestions, let me know what you think!

deploy/addons/kubevirt/pod.yaml.tmpl
```diff
@@ -0,0 +1,77 @@
...
+    #!/bin/bash
+
+    kubectl delete -f /manifests/kubevirt.yaml
+
+    kubectl delete -f /manifests/kubevirt-base.yaml
+
+  install.sh: |
+    #!/bin/bash
+
+    export KUBEVIRT_VERSION=$(curl -s https://api.github.com/repos/kubevirt/kubevirt/releases | grep tag_name | grep -v -- - | sort -V | tail -1 | awk -F':' '{print $2}' | sed 's/,//' | xargs)
```
fabiand, line 20: Ha. Nifty, nice to see that this is _not_ the naive approach of retrieving the highest version :)

site/content/en/docs/tutorials/kubevirt.md
ashleyschuett: this has been updated! and the PR is ready for a rereview!

Issue #8243, opened 2020-05-21
TestVersionUpgrade broken on master


medyagh:
```
med@xmac:~/workspace/minikube (test_ver_upgrade)$ make && ./out/minikube start --driver=docker --kubernetes-version=v1.12.0

med@xmac:~/workspace/minikube (test_ver_upgrade)$ kc version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-21T01:25:41Z", GoVersion:"go1.13.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.0", GitCommit:"0ed33881dc4355495f623c6f22e7dd0b7632b7c0", GitTreeState:"clean", BuildDate:"2018-09-27T16:55:41Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}

```

### stop 
```
med@xmac:~/workspace/minikube (test_ver_upgrade)$ ./out/minikube stop
✋  Stopping "minikube" in docker ...
🛑  Powering off "minikube" via SSH ...
🛑  Node "minikube" stopped.
```

### start with newest version 
```
med@xmac:~/workspace/minikube (test_ver_upgrade)$ ./out/minikube start --driver=docker --kubernetes-version=v1.18.3-beta.0
😄  minikube v1.10.1 on Ubuntu 20.04
✨  Using the docker driver based on existing profile
👍  Starting control plane node minikube in cluster minikube
🔄  Restarting existing docker container for "minikube" ...
🐳  Preparing Kubernetes v1.18.3-beta.0 on Docker 19.03.2 ...
    ▪ kubeadm.pod-network-cidr=10.244.0.0/16
🔎  Verifying Kubernetes components...
🌟  Enabled addons: default-storageclass, storage-provisioner
🏄  Done! kubectl is now configured to use "minikube"
med@xmac:~/workspace/minikube (test_ver_upgrade)$ kc version
[...]

PR #8279, opened 2020-05-26
fix parsing kubernetes version flag on an existing minikube

medyagh:
closes https://github.com/kubernetes/minikube/issues/8243


medyagh:
> perhaps the integration test that does upgrades should use this feature?

sounds good for another PR, maybe could be a good first issue for a new contributor

PR #8280, opened 2020-05-27
Merge TestOlmAddon into TestAddons

medyagh:
the test was spinnig up a new cluster and was failing the none but was passing docker and other drivers


sharifelgamal:
The olm test failed on none

medyagh:
> The olm test failed on none

thanks  skipped the none

PR #8289, opened 2020-05-27
Add more debug to TestFunctional/parallel/SSHCmd

tstromberg:


pkg/minikube/cluster/ip.go
```diff
@@ -86,6 +86,8 @@ func HostIP(host *host.Host) (net.IP, error) {
 		return net.IPv4(vmIP[0], vmIP[1], vmIP[2], byte(1)), nil
 	case driver.None:
 		return net.ParseIP("127.0.0.1"), nil
+	case driver.Parallels:
```
medyagh, line 89: is this an unrelated change?

PR #8311, opened 2020-05-28
github actions: add clean up to master too

medyagh:

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -311,7 +311,7 @@ jobs:
           $env:KUBECONFIG="${pwd}\testhome\kubeconfig"
           $env:MINIKUBE_HOME="${pwd}\testhome"
           $ErrorActionPreference = "SilentlyContinue"
-          ./e2e-windows-amd64.exe --minikube-start-args="--driver=hyperv" --test.timeout=13m --timeout-multiplier=1.5 --test.v --test.run="TestFunctional" --binary="./minikube-windows-amd64.exe" | Out-File -FilePath .\report\testout.txt -Encoding ASCII
+          ./e2e-windows-amd64.exe --minikube-start-args="--driver=hyperv" --test.timeout=13m --timeout-multiplier=2 --test.v --test.run="TestFunctional" --binary="./minikube-windows-amd64.exe" | Out-File -FilePath .\report\testout.txt -Encoding ASCII
           $END_TIME=(GET-DATE)
           echo $END_TIME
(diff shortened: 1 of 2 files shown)


.github/workflows/master.yml
```diff
@@ -176,6 +176,17 @@ jobs:
           echo $env:computerName
           ls 
           $ErrorActionPreference = "SilentlyContinue"
+          cd minikube_binaries
+          ls
+          $env:KUBECONFIG="${pwd}\testhome\kubeconfig"
+          $env:MINIKUBE_HOME="${pwd}\testhome"
+          .\minikube-windows-amd64.exe delete --all --purge
+          Get-VM | Where-Object {$_.Name -ne "DockerDesktopVM"} | Foreach {
```
tstromberg, line 184: This is very strange and needs a comment.

.github/workflows/master.yml
```diff
@@ -176,6 +176,17 @@ jobs:
...
+            minikube-windows-amd64.exe delete -p $_.Name
```
tstromberg, line 185: I think this line can be removed.

.github/workflows/master.yml
```diff
@@ -176,6 +176,17 @@ jobs:
...
+            Suspend-VM $_.Name
```
tstromberg, line 186: Suspend-VM is meaningless if Stop-VM is called subsequently.

medyagh: I copied from our jenkoins one

.github/workflows/master.yml
```diff
@@ -176,6 +176,17 @@ jobs:
...
+            Stop-VM $_.Name -Force
```
tstromberg, line 187: For clarity, I think this should specify `-Name`, as per https://docs.microsoft.com/en-us/powershell/module/hyper-v/stop-vm?view=win10-ps

I think `-TurnOff` may be a valuable flag as well.

.github/workflows/master.yml
```diff
@@ -176,6 +176,17 @@ jobs:
...
+            Remove-VM $_.Name -Force
+          }
           Remove-Item minikube_binaries -Force -Recurse
```
tstromberg, line 190: A previous step ran `cd minikube_binaries` -- does that mean there is a minikube_binaries within it, or do we need to remove ourselves from that directory first?

.github/workflows/master.yml
medyagh: sometimes it is better to do that if delete --all is not doing the right thing

PR #8320, opened 2020-05-29
docker driver: add fall back image in docker hub 

medyagh:

### DIFF
--- a/pkg/drivers/kic/types.go
+++ b/pkg/drivers/kic/types.go
@@ -35,10 +35,15 @@ const (
 var (
 	// BaseImage is the base image is used to spin up kic containers. it uses same base-image as kind.
 	BaseImage = fmt.Sprintf("gcr.io/k8s-minikube/kicbase:%s@sha256:%s", Version, baseImageSHA)
-	// BaseImageFallBack the fall back of BaseImage in case gcr.io is not available. stored in github packages https://github.com/kubernetes/minikube/packages/206071
+
+	// BaseImageFallBack1 the fall back of BaseImage in case gcr.io is not available. stored in docker hub
+	// same image is push to https://github.com/kicbase/stable
+	BaseImageFallBack1 = fmt.Sprintf("kicbase/stable:%s@sha256:%s", Version, baseImageSHA)
+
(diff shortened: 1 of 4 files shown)


pkg/drivers/kic/types.go
```diff
@@ -35,10 +35,15 @@ const (
...
+
+	// BaseImageFallBack1 the fall back of BaseImage in case gcr.io is not available. stored in docker hub
+	// same image is push to https://github.com/kicbase/stable
+	BaseImageFallBack1 = fmt.Sprintf("kicbase/stable:%s@sha256:%s", Version, baseImageSHA)
+
+	// BaseImageFallBack2 the fall back of BaseImage in case gcr.io is not available. stored in github packages https://github.com/kubernetes/minikube/packages/206071
 	// github packages docker does _NOT_ support pulling by sha as mentioned in the docs:
 	// https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-docker-for-use-with-github-packages
-	BaseImageFallBack = fmt.Sprintf("docker.pkg.github.com/kubernetes/minikube/kicbase:%s", Version)
+	BaseImageFallBack2 = fmt.Sprintf("docker.pkg.github.com/kubernetes/minikube/kicbase:%s", Version)
```
priyawadhwa, line 46: I think it would be cleaner to create an array of images, and then go through them in order until we find one that works

```go
BaseImages = []string{
 fmt.Sprintf("gcr.io/k8s-minikube/kicbase:%s@sha256:%s", Version, baseImageSHA),
 fmt.Sprintf("kicbase/stable:%s@sha256:%s", Version, baseImageSHA),
 fmt.Sprintf("docker.pkg.github.com/kubernetes/minikube/kicbase:%s", Version)
}

```

medyagh: sounds good the problem is when looping over the array, would need to detect if there is not any more fallback to return the error, seems like a good idea for a follow up refactor

tstromberg: I agree, this should definitely be a slice.

PR #8336, opened 2020-05-30
Remove kubelet options for k8s version < 1.12

prasadkatti:
This does some cleanup, now that the minimum k8s version is bumped up to 1.13

### DIFF
--- a/pkg/minikube/bootstrapper/bsutil/versions.go
+++ b/pkg/minikube/bootstrapper/bsutil/versions.go
@@ -50,14 +50,6 @@ var versionSpecificOpts = []config.VersionedExtraOption{
 	// Kubeconfig args
 	config.NewUnversionedOption(Kubelet, "kubeconfig", "/etc/kubernetes/kubelet.conf"),
 	config.NewUnversionedOption(Kubelet, "bootstrap-kubeconfig", "/etc/kubernetes/bootstrap-kubelet.conf"),
-	{
-		Option: config.ExtraOption{
-			Component: Kubelet,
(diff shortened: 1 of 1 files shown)


medyagh:
We should still be able to support v1.12 with --force

prasadkatti:
@medyagh - That should be fine. The options removed in this PR are for versions older than 1.12.

prasadkatti:
I think the test failures are just flakes. Can I get some reviews on this?

priyawadhwa:
Hey @prasadkatti I also think it might be best to leave this in, in case somebody is trying to use any k8s version <1.13 with the `--force` flag

prasadkatti:
@priyawadhwa What is the earliest k8s version one can ask for with minikube?

priyawadhwa:
so we officially only promise to support >=1.13, but right now people can use lower versions with `--force` if they want to. I figure we can keep this code in, in the case people are doing that (we just don't promise it'll work).

It would be nice if we knew if anyone was actually doing that, but I don't think we have those stats...

prasadkatti:
Also as new code gets added we are bound to be breaking things for the older versions. So it would be nice to know what is the oldest version of k8s that still works with minikube 1.11.

priyawadhwa:
@prasadkatti yah I see what you mean; unfortunately we don't really have data around what k8s versions people are using, so it's hard to say if anyone is using <1.12 with `--force`. I say we should play it safe and in case some people are using <1.12 with `--force` I think we should keep this code in.

I'm going to close this PR for now since we want to keep this code in, but thank you for bringing this up!

PR #8341, opened 2020-05-31
Add conntrack installation and fix some spacing

shubhindia:
On k8s v.1.18+ we get this error while trying to install with minikube.
root@pop-os:~# minikube start --driver=none --kubernetes-version=v1.18.0
😄  minikube v1.11.0 on Debian bullseye/sid
💣  Sorry, Kubernetes 1.18.0 requires conntrack to be installed in root's path

Can be easily solved by installing conntrack. So added it in requirements.
There is a space missing in "This VM must also meet thekubeadm requirements" which should be like "This VM must also meet the kubeadm requirements"
Before
before
After
after

### DIFF
--- a/site/content/en/docs/drivers/includes/none_usage.inc
+++ b/site/content/en/docs/drivers/includes/none_usage.inc
@@ -4,8 +4,13 @@ A Linux VM with the following:
 
 * Docker
 * systemd (OpenRC based systems are also supported in v1.10+)
+* conntrack (Kubernetes v1.18+ needs conntrack to be installed on the machine)
 
-This VM must also meet the[kubeadm requirements](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/)
+    ```shell
+    sudo apt-get install conntrack -y
+    ```
+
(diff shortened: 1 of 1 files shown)


medyagh:
Can you plz put in the description what you are trying to fix and also add the before after this PR output?

shubhindia:
@medyagh Added the before and after in the description.

site/content/en/docs/drivers/includes/none_usage.inc
```diff
@@ -4,8 +4,13 @@ A Linux VM with the following:
 
 * Docker
 * systemd (OpenRC based systems are also supported in v1.10+)
+* conntrack (Kubernetes v1.18+ needs conntrack to be installed on the machine)
 
-This VM must also meet the[kubeadm requirements](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/)
+    ```shell
```
medyagh, line 9: I suggest moving this installation bellow since this is only debian specific.
you can have a section

### how to install contntrack on debian :

Issue #7420, opened 2020-04-05
Support external DOCKER_HOST from inside WSL


nnashok:
**Steps to reproduce the issue:** 
1. I have Docker for Windows installed. I also have Ubuntu installed in WSL. I am able to connect to the Docker daemon from my Ubuntu shell:
```
ashoknn@ashoknn-yoga:~$ docker ps
CONTAINER ID        IMAGE                                COMMAND                  CREATED             STATUS              PORTS                                                                           NAMES
ashoknn@ashoknn-yoga:~$
```
2. If I try to create a cluster using minikube from within my Ubuntu shell, it fails to connect to the Docker daemon:
```ashoknn@ashoknn-yoga:~$ minikube start --driver=docker --alsologtostderr
W0404 19:52:48.033354   14510 root.go:248] Error reading config file at /home/ashoknn/.minikube/config/config.json: open /home/ashoknn/.minikube/config/config.json: no such file or directory
I0404 19:52:48.038441   14510 notify.go:125] Checking for updates...
I0404 19:52:48.445060   14510 start.go:262] hostinfo: {"hostname":"ashoknn-yoga","uptime":15431,"bootTime":1586039737,"procs":21,"os":"linux","platform":"ubuntu","platformFamily":"debian","platformVersion":"18.04","kernelVersion":"4.4.0-18362-Microsoft","virtualizationSystem":"","virtualizationRole":"","hostid":"20d448e0-882e-40b0-80bc-ed502a1eca7b"}
I0404 19:52:48.446325   14510 start.go:272] virtualization:
😄  minikube v1.9.2 on Ubuntu 18.04
I0404 19:52:48.450545   14510 driver.go:245] Setting default libvirt URI to qemu:///system
✨  Using the docker driver based on existing profile
I0404 19:52:48.586065   14510 start.go:310] selected driver: docker
[...]

Issue #5392, opened 2019-09-18
Support for Windows 10 WSL2


feaber:
Is there is a plan to support WSL2 on Windows 10?

Docker is already experimenting with it: 
https://docs.docker.com/docker-for-windows/wsl-tech-preview/

Do You think supporting WSL2 could save us from using virtual machines?

tstromberg:
Yes, there is a plan to support WSL2! I'm not sure if folks have experimented with `--vm-driver=none` and WSL2, but we definitely plan to support it using the Docker deployment model in #4772

Help wanted =)

afbjorklund:
Interestingly, using WSL will *not* save you from running a VM. Similar to Docker-for-Windows it will just let you run on _someone else_’s VM...

https://devblogs.microsoft.com/commandline/announcing-wsl-2/

> WSL 2 uses the latest and greatest in virtualization technology to run its Linux kernel inside of a lightweight utility virtual machine (VM).

LouisStAmour:
For folks looking for WSL 2 support via `vm-driver=none`, using a stable docker-ce build, it's not yet ready. I followed the stable instructions at https://docs.docker.com/install/linux/docker-ce/ubuntu/ after installing WSL 2 and Ubuntu, and had a few errors trying to get Docker to run.

In particular, it's worth noting the following:

https://github.com/WhitewaterFoundry/Pengwin/issues/485#issuecomment-518028465 for a Debian iptables incompatibility

It's possible the Debian or Pengwin distros are compatible with the suggested fix, but I was not having luck switching on legacy iptables in the store-provided Ubuntu distro.

Switching to virtualized minikube, it looks as if there are some Kernel flag incompatibilities in the stock WSL 2 kernel that ships with 1909:

https://github.com/microsoft/WSL/issues/4165#issuecomment-519245959 notes some kernel incompatibilities requiring a new kernel compile

The noted kernel flag is now turned on by default in future kernel builds: https://github.com/microsoft/WSL2-Linux-Kernel/blob/master/Microsoft/config-wsl#L1448 Running that config file through Docker's ./check-config, all but CONFIG_DEVPTS_MULTIPLE_INSTANCES under "Generally Necessary" are enabled in Microsoft's latest Linux kernel builds.

[...]

PR #8368, opened 2020-06-03
add support for microsoft wsl for docker driver

medyagh:
have not tested myself but acording to this comment it works 
https://github.com/kubernetes/minikube/issues/5392#issuecomment-638562724

closes https://github.com/kubernetes/minikube/issues/7420 
closes https://github.com/kubernetes/minikube/issues/5392

### DIFF
--- a/pkg/minikube/driver/driver.go
+++ b/pkg/minikube/driver/driver.go
@@ -136,7 +136,7 @@ func NeedsPortForward(name string) bool {
 
 // isMicrosoftWSL will return true if process is running in WSL in windows
 // checking for WSL env var based on this https://github.com/microsoft/WSL/issues/423#issuecomment-608237689
-// also based on https://github.com/microsoft/vscode/blob/master/resources/win32/bin/code.sh#L24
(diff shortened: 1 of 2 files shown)


pkg/minikube/driver/driver.go
```diff
@@ -131,7 +131,14 @@ func NeedsRoot(name string) bool {
...
 func NeedsPortForward(name string) bool {
 	// Docker for Desktop
-	return IsKIC(name) && (runtime.GOOS == "darwin" || runtime.GOOS == "windows")
+	return IsKIC(name) && (runtime.GOOS == "darwin" || runtime.GOOS == "windows" || isMicrosoftWSL())
+}
+
+// isMicrosoftWSL will return true if process is running in WSL in windows
+// checking for WSL env var based on this https://github.com/microsoft/WSL/issues/423#issuecomment-608237689
+// also based on https://github.com/microsoft/vscode/blob/90a39ba0d49d75e9a4d7e62a6121ad946ecebc58/resources/win32/bin/code.sh#L24
+func isMicrosoftWSL() bool {
```
priyawadhwa, line 140: Let's make this function public so that we don't need to repeat it in the integration package

medyagh: sounds good

PR #8392, opened 2020-06-05
Patch the registry-aliases addon missing entrypoint

### LINKED ISSUE #8383: Addon registry-aliases alpine image missing bash command (ash works)

1. minikube start
2. minikube addons enable registry
3. minikube addons enable registry-aliases

kubelet says:
`
Error: failed to start container "update": Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"bash\": executable file not found in $PATH": unknown
`

Easily fixed by changing the daemonset entrypoint (command[0]) from `bash` to `ash`

tsraza:
Fixes #8383

### DIFF
--- a/deploy/addons/registry-aliases/node-etc-hosts-update.tmpl
+++ b/deploy/addons/registry-aliases/node-etc-hosts-update.tmpl
@@ -29,7 +29,7 @@ spec:
                   name: registry-aliases
                   key: registryAliases
           command:
-            - bash
+            - ash
             - -ce
             - |
               NL=$'\n'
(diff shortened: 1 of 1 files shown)


medyagh:
@tsraza thanks this PR

deploy/addons/registry-aliases/node-etc-hosts-update.tmpl
```diff
@@ -29,7 +29,7 @@ spec:
                   name: registry-aliases
                   key: registryAliases
           command:
-            - bash
+            - ash
```
afbjorklund, line 32: Could consider using standard /bin/sh here

medyagh: @tsraza I agree with @afbjorklund

tsraza: changed to `sh` and tested locally in virtualbox -- it works

Issue #6887, opened 2020-03-04
refactor docker-env to a package


medyagh:
we should be able to generate eval commands per shell type

so for example in this PR, we could give them exact command to re-eval their docker-env or podman env
https://github.com/kubernetes/minikube/pull/6885

PR #8413, opened 2020-06-08
Refactor docker env and podmanenv

n0npax:
#6887 
Refactor podman and dockerenv to pkg
change `machine/fix.go` to show eval hints per shell


n0npax:
Sure :) PR is not ready yet.

n0npax:
@medyagh I'd like to ask for some advice

Given PR is not ready yet. It has some 2 major issues so far.

1. Cyclic dependency
To implement a nice suggestion for `docker-env`, I need to update https://github.com/kubernetes/minikube/pull/8413/files#diff-054b76023d58321f15e129dc4efd1f63R172 with all required params like: https://github.com/kubernetes/minikube/pull/8413/files#diff-9ea32766b084e8a52f556fbf585988f7R81 , to get those params I can exec https://github.com/kubernetes/minikube/blob/master/pkg/minikube/mustload/mustload.go#L83, but unfortunately, it will cause cyclic dependency as pkg `mustload` depends on `machine`. I can try to extract some logic from `mustload` to different pkg, but not sure if this is acceptable. Any suggestions?

2. Is there any elegant way to get the current profile name?

pkg/minikube/shell/shell.go
```diff
@@ -197,3 +192,17 @@ func UnsetScript(ec EnvConfig, w io.Writer, vars []string) error {
 	_, err := w.Write([]byte(sb.String()))
 	return err
 }
+
+// Detect detects user's current shell.
+func Detect() (string, error) {
+	return shell.Detect()
+}
+
+func GetShell(useShell string) (EnvConfig, error) {
```
n0npax, line 201: missing comment

n0npax:
Hey Maintainers,

I understand that size of the given PR makes you angry and I'd like to say sorry for this.

Unfortunately, I don't see any better option to solve #6887 than merging `mustload` package with `machine`.

To explain the pain and main issue here, please take a look at https://github.com/kubernetes/minikube/pull/8413/files#diff-054b76023d58321f15e129dc4efd1f63R170 `Running` which returns `ClusterControler`.  Function `Running` heavily depends on **`machine`** & `config` pkgs. To get required data I'd need to copy exactly the same code to  `machine` package and it is against DRY rule. Any attempt to extract common logic to other package will also fail as a cyclic dependency will stay (`machine` <-> `new package`).
I understand `mustload.Healthy` function was also moved to `machine`, but I believe it's a lesser evil. `Healthy` heavily depends on `Running`.
Solving cyclic dependency issue by coping code, would require me to copy around 70% of `mustload` code into `machine`.

once again sorry for PR size.

afbjorklund:
> I understand that size of the given PR makes you angry and I'd like to say sorry for this.

Not so much angry, but hard to review... Maybe should drop the original issue instead ?

i.e. maybe it is better to leave docker-env and podman-env as they are, if it is too hard.

I personally found the "shell" module to be a better abstraction.

n0npax:
To be honest, most of the modifications in the PR are caused by merging `machine` and `mustload` pkg, so reviewing this part should be fairly easy. (it's basically sed on files). Changes introduced by `sed` seems to be source of 32 out of 37 files in this PR:
```sh
git show -p 7401b454383dc69fb1076de31b209f90c30dd72f --name-only | grep '.go' | wc -l
32
```

Assuming https://github.com/kubernetes/minikube/blob/master/pkg/minikube/machine/fix.go#L162 can stay "as is" and extracting logic from `cmd` is still desired, I can revert merging `mustload` and `machine`.

n0npax:
> I recommend re-doing this PR only to address one issue and then other PRS for other issues.

Can you write it down as a more detailed list, please?

> provide a way to get Only the "Eval" instruction per shell, for example eval instruciton for fish,powershell, bash... so when we want to Warn the user to re-eval their docker-env, ...

`dockerSetScript` is using `shell.SetScript` which is providing `Eval` functionality.

> ... they should have the exact command for their shell (as opposed to telling them run minikube docker-env and then get their eval command

I believe it was implemented here.

To achieve this goal I had to pass those data to Shell package from machine pacakge (fix.go file)
```golang
daemonenv.DockerEnvConfig{
			EnvConfig: sh,
			Profile:   profileName,
			Driver:    co.CP.Host.DriverName,
			HostIP:    co.CP.IP.String(),
			Port:      constants.DockerDaemonPort,
			CertsDir:  localpath.MakeMiniPath("certs"),
		}
```
fetching those data inside `machine` pkg is causing an issue mentioned earlier.

If I'm missing or misunderstanding something here, can you please provide more detailed suggestions?

medyagh:
>  Can you write it down as a more detailed list, please?

sure thing, so the goal of this PR should be the following:

in this line of code:
https://github.com/medyagh/minikube/blob/e91e11eebaf32773659bc19fd0d22675a1fd7134/pkg/minikube/machine/fix.go#L166-L167
when we want to warn the user to re-eval their docker-env, instead of telling to run minikube docker-env and then get the command to eval we should get them the exact command to eval.

to make that happen you will need to make the template more reusable
https://github.com/medyagh/minikube/blob/e91e11eebaf32773659bc19fd0d22675a1fd7134/cmd/minikube/cmd/docker-env.go#L45-L46

so u can generate only the command part of it as opposed to the whole thing
for example we should we able to get
```
eval $(minikube -p minikube docker-env)
```
instead of whole thing
```
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://127.0.0.1:32838"
export DOCKER_CERT_PATH="/Users/medya/.minikube/certs"

```

mind that the
eval $(minikube -p minikube docker-env) is only valid for bash, but for powershell or zsh or fish we should get their eval command

so when we warn the user 
in this line https://github.com/medyagh/minikube/blob/e91e11eebaf32773659bc19fd0d22675a1fd7134/pkg/minikube/machine/fix.go#L166-L167 to reval their docker we could just give them exact eval command ( instead of asking them to do minikube docker-env to get the exact command)

pkg/minikube/machine/fix.go
```diff
@@ -158,25 +159,21 @@ func maybeWarnAboutEvalEnv(drver string, name string) {
...
+	}
 	if os.Getenv(constants.MinikubeActiveDockerdEnv) != "" {
 		out.T(out.Notice, "Noticed you have an activated docker-env on {{.driver_name}} driver in this terminal:", out.V{"driver_name": drver})
-		// TODO: refactor docker-env package to generate only eval command per shell. https://github.com/kubernetes/minikube/issues/6887
-		out.WarningT(`Please re-eval your docker-env, To ensure your environment variables have updated ports:
-
-	'minikube -p {{.profile_name}} docker-env'
-
-	`, out.V{"profile_name": name})
+		msg := shell.GenerateUsageHint(ec, "eval docker-env", "minikube -p {{.profile_name}} docker-env")
```
medyagh, line 169: there is no need to  eval docker-env, 
I suggest this function take two parameters
1- shell type (bash, powershell,fish,...)
2-profile name

and based on that it give you the exact command to eval.
that way the code will look cleaner.

n0npax:
Hey,

Before/After output

old fish for docker
```
W0624 08:24:20.113994 1374870 out.go:79] [unset outFile]: * Noticed you have an activated docker-env on docker driver in this terminal:
E0624 08:24:20.114040 1374870 out.go:102] [unset errFile]: ! Please re-eval your docker-env, To ensure your environment variables have updated ports:

        'minikube -p minikube docker-env'
```
new fish for podman
```
W0624 08:19:42.329065 1367350 out.go:79] [unset outFile]: * Noticed you have an activated podman-env on podman driver in W0624 08:19:42.329416 1367350 out.go:79] [unset outFile]: * Noticed you have an activated podman-env on podman driver in this terminal:
E0624 08:19:42.329449 1367350 out.go:102] [unset errFile]: ! 
# eval podman-env
# minikube -p minikube podman-env | source
```

pkg/minikube/machine/fix.go
n0npax: Hey,
I'm sorry but I'm missing sth here. `eval docker-env,` is just usage comment. I can replace it with something more descriptive like `please eval docker-env using given snippet` or similar. Variable `ec`  contains shell type.

Do you want me to add new function to https://github.com/kubernetes/minikube/blob/master/pkg/minikube/shell/shell.go#L86, so string `"minikube -p {{.profile_name}} docker-env"` will be extracted to shell pkg and invocation will look like: 
```golang
msg := shell.GenerateDockerEnvEval(ec, profile) 
out.WarningT(msg)
```
This will require adding 2 new functions to shell pkg (1 for podman 1 for docker).

Can I ask for additional clarification please?

n0npax: `2567e3af4a8ee655e47dac63831f54f5107457f0` provides some fixes here. Not sure if it's exactly what was requested.

n0npax:
/assign @medyagh 
@medyagh can you please review?

medyagh:
@n0npax do you mind please update the PR with Before and After output for the suggestion hint output ?

pkg/minikube/shell/shell.go
```diff
@@ -129,20 +129,25 @@ REM @FOR /f "tokens=*" %%i IN ('%s') DO @%%i
 	},
 }
 
-var defaultSh = "bash"
-var defaultShell shellData = shellConfigMap[defaultSh]
+// DefaultShellName is default shell which will be used in case of shell detection error
+const DefaultShellName = "bash"
```
medyagh, line 133: change the name, "shell.DefaultShellName" stutters.

how about shell.Default ?

n0npax:
Hey @medmedchiheb ,

It's in comment above

-----

Before/After output

old fish for docker
```
W0624 08:24:20.113994 1374870 out.go:79] [unset outFile]: * Noticed you have an activated docker-env on docker driver in this terminal:
E0624 08:24:20.114040 1374870 out.go:102] [unset errFile]: ! Please re-eval your docker-env, To ensure your environment variables have updated ports:

        'minikube -p minikube docker-env'
```
new fish for podman
```
W0624 08:19:42.329065 1367350 out.go:79] [unset outFile]: * Noticed you have an activated podman-env on podman driver in W0624 08:19:42.329416 1367350 out.go:79] [unset outFile]: * Noticed you have an activated podman-env on podman driver in this terminal:
E0624 08:19:42.329449 1367350 out.go:102] [unset errFile]: ! 
# eval podman-env
# minikube -p minikube podman-env | source
```

pkg/minikube/shell/shell.go
```diff
@@ -155,7 +160,7 @@ func (c EnvConfig) getShell() shellData {
 	return shell
 }
 
-func generateUsageHint(ec EnvConfig, usgPlz, usgCmd string) string {
+func GenerateUsageHint(ec EnvConfig, usgPlz, usgCmd string) string {
```
medyagh, line 163: change usgCmd to "daemon"

pkg/minikube/shell/shell.go
```diff
@@ -202,3 +207,15 @@ func UnsetScript(ec EnvConfig, w io.Writer, vars []string) error {
 	_, err := w.Write([]byte(sb.String()))
 	return err
 }
+
+func generateSthEnvEvalHint(ec EnvConfig, profile, command string) string {
```
medyagh, line 211: change command to "daemon"

pkg/minikube/shell/shell.go
medyagh, line 163: all exported functions need to have comment, is this function used anywhere outside the package ? if not make it private

medyagh:
I half reviewed this PR, I will make antoher review once the changes are done

cmd/minikube/cmd/docker-env.go
```diff
@@ -199,93 +109,28 @@ var dockerEnvCmd = &cobra.Command{
 		}
 
 		if dockerPath != "" {
-			out, err := tryDockerConnectivity("docker", ec)
+			out, err := daemonenv.TryDockerConnectivity("docker", ec)
```
medyagh, line 112: this feels not right that both the name has TryDockerConnectivity and we path "docker"
either the name should be TryConnectivity("docker, ec)
or it should be TryConnectivy("docker, ec)

n0npax: fixed

pkg/minikube/daemonenv/daemonenv.go
```diff
@@ -0,0 +1,43 @@
...
+	switch bin {
+	case "docker":
+		c := exec.Command(bin, "version", "--format={{.Server}}")
+		c.Env = append(os.Environ(), dockerEnvVarsList(ec)...)
+		glog.Infof("Testing Docker connectivity with: %v", c)
+		return c.CombinedOutput()
+	default:
+		msg := fmt.Sprintf("Tried to test connectivity of unsupported type: %s", bin)
+		glog.Infof(msg)
+		panic(msg)
```
medyagh, line 41: plz dont use panic in here,

pkg/minikube/daemonenv/daemonenv.go
```diff
@@ -0,0 +1,43 @@
...
+	"os"
+	"os/exec"
+
+	"github.com/golang/glog"
+)
+
+// TryConnectivity will try to connect to daemon env from user's POV to detect the problem if it needs reset or not
+func TryConnectivity(bin string, ec DockerEnvConfig) ([]byte, error) {
+	switch bin {
+	case "docker":
```
medyagh, line 33: it seems like this func only tries connectivity to Docker, why there is a case switch?

n0npax: This was done due to this comment https://github.com/kubernetes/minikube/pull/8413#discussion_r455264402

I followed `TryConnectivity("docker", ec)` and I thought it will be wrong to accept any other argument than "docker", so it's a reason for the switch and panic. I was afraid of passing any binary to exec here, as `--format` can have different output.
````golang
c := exec.Command(bin, "version", "--format={{.Server}}")
````

File: pkg/minikube/daemonenv/daemonenv.go (hunk 7 above)
Comment (n0npax) [reply pkg/minikube/daemonenv/daemonenv.go]: Sorry, still trying to get used to new principals.
Fixed.

File: pkg/minikube/daemonenv/daemonenv.go (hunk 8 above)
Comment (n0npax) [reply pkg/minikube/daemonenv/daemonenv.go:33]: happy to rename `TryConnectivity` to `TryDockerConnectivity` and remove switch if preferred.

File: pkg/minikube/daemonenv/podmanenv.go [hunk 9]
```diff
@@ -0,0 +1,130 @@
...
+func PodmanUnsetScript(ec PodmanEnvConfig, w io.Writer) error {
+	vars := []string{
+		constants.PodmanVarlinkBridgeEnv,
+		constants.MinikubeActivePodmanEnv,
+	}
+	return shell.UnsetScript(ec.EnvConfig, w, vars)
+}
+
+// podmanBridge returns the command to use in a var for accessing the podman varlink bridge over ssh
+func podmanBridge(client *ssh.ExternalClient) string {
```
n0npax, line 116: client can be consider as shittering

n0npax:
@medyagh can you please review once again?

pkg/minikube/daemonenv/daemonenv.go
n0npax: quite similar to https://github.com/kubernetes/minikube/blob/master/pkg/minikube/image/image.go#L106

tstromberg:
Oof. Sorry about the merge conflicts. Do you mind addressing them?

n0npax:
Hi, I'll close this and reopen as a new one once conflicts are solved. I'll try to solve them a little bit later. Sorry for the delay.

PR #8417, opened 2020-06-08
Enable support for offline kic

priyawadhwa:
This PR does a couple things:

1. For `--download-only=true`, if the docker daemon isn't running, store the kic base image as a tarball on the host machine
1. Before trying to pull the kic base image, check if it exists as a tarball in the cache and load it into the daemon if it does. If we can load the image into the daemon successfully, strip the base image if its digest, since docker doesn't include the digest when loading images into the daemon. 

This way, we can download all required artifacts to the host machine, and later run the docker driver offline (I tested this locally and it works).

This should fix one of the bugs around running minikube in Cloud Shell.

Fixes https://github.com/kubernetes/minikube/issues/8366


pkg/minikube/image/image.go
```diff
@@ -94,6 +96,26 @@ func ExistsImageInDaemon(img string) bool {
 	return false
 }
 
+// LoadFromTarball checks if the image exists as a tarball and tries to load it to the local daemon
+// TODO: Pass in if we are loading to docker or podman so this function can also be used for podman
```
medyagh, line 100: for simplicity, instead of this TODO. you can pass in the dameon-type and then for non-docker ones return "not implemented yet. see isssue #1234" error

priyawadhwa: SGTM, done!

priyawadhwa:
confirmed that this works locally

medyagh:
Docker Driver : TestFunctional/parallel/DockerEnv is related to this flake:
https://github.com/kubernetes/minikube/issues/8427

https://storage.googleapis.com/minikube-builds/logs/8417/1109616/Docker_Linux.html#fail_TestFunctional%2fparallel%2fDockerEnv

PR #8428, opened 2020-06-09
Replace whitelist/blacklist with allow/deny

tstromberg:
Allow and deny are easier to understand, particularly for non-native English speakers. This also omits unnecessary societal baggage.


cmd/minikube/cmd/start.go
```diff
@@ -865,10 +865,10 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 		}
 	}
 
-	// check that kubeadm extra args contain only whitelisted parameters
+	// check that kubeadm extra args contain only allowed parameters
 	for param := range config.ExtraOptions.AsMap().Get(bsutil.Kubeadm) {
-		if !config.ContainsParam(bsutil.KubeadmExtraArgsWhitelist[bsutil.KubeadmCmdParam], param) &&
-			!config.ContainsParam(bsutil.KubeadmExtraArgsWhitelist[bsutil.KubeadmConfigParam], param) {
+		if !config.ContainsParam(bsutil.KubeadmExtraArgsAllow[bsutil.KubeadmCmdParam], param) &&
```
medyagh, line 870: how about KubeadmExtraArgsAllowList

tstromberg: I started with AllowList (even if this is a dictionary, rather than a list), but removed it in preference to following the idea that one should omit words that are obvious given a variable's type declaration or are clear from the surrounding context.

pkg/minikube/bootstrapper/bsutil/extraconfig.go
```diff
@@ -56,12 +56,12 @@ var componentToKubeadmConfigKey = map[string]string{
 	Kubelet: "",
 }
 
-// KubeadmExtraArgsWhitelist is a whitelist of supported kubeadm params that can be supplied to kubeadm through
+// KubeadmExtraArgsAllow is a list of supported kubeadm params that can be supplied to kubeadm through
```
celestehorgan, line 59: Some suggestions if you're open to them:

- `KubeadmExtraArgAllowlist` (it's a bit more specific that this is replacing the term "whitelist", which could be helpful)
- `KubeadmExtraArgsAllowed` (grammatically/semantically correct. `ExtraArgsAllow` to me implies a boolean, where as `ExtraArgsAllowed` implies a list of some sort.)

💖

PR #8450, opened 2020-06-11
Upgrade falco-probe driver kernel module to 0.23

afbjorklund:


afbjorklund:
Maybe this helps with the build error, but both versions built fine here locally

PR #8457, opened 2020-06-11
Improve logging for cloud shell

priyawadhwa:
For cloud shell, on restart the home directory of the VM is preserved. This means that if minikube was running, the minikube profile config is leftover even though minikube has been deleted.

When running `minikube start` on a refreshed Cloud Shell VM, minikube thinks there's an existing cluster because the profile exists, and we get this ugly log to stdout:

```
$ minikube start
* minikube v1.10.1 on Debian 10.4
  - MINIKUBE_FORCE_SYSTEMD=true
* Pulling base image ...
* minikube 1.11.0 is available! Download it: https://github.com/kubernetes/minikube/releases/tag/v1.11.0
* To disable this notice, run: 'minikube config set WantUpdateNotification false'
* docker "minikube" container is missing, will recreate.
E0611 10:53:46.318887     504 oci.go:79] docker daemon seems to be stuck. Please try restarting your docker. Will try to delete anyways: unknown state "minikube": docker inspect minikube --format={{.State.Status}}: exit status 1
stdout:
stderr:
Error: No such object: minikube

E0611 10:53:46.318887     504 oci.go:79] docker daemon seems to be stuck. Please try restarting your docker. Will try to delete anyways: unknown state "minikube": docker inspect minikube --format={{.State.Status}}: exit status 1
stdout:
stderr:
Error: No such object: minikube

E0611 10:53:46.318887     504 oci.go:79] docker daemon seems to be stuck. Please try restarting your docker. Will try to delete anyways: unknown state "minikube": docker inspect minikube --format={{.State.Status}}: exit status 1
stdout:
stderr:
Error: No such object: minikube
```

I generally think all glog logs should be sent to stderr & if we want an explicit warning message here it should go through the `out` package.

In this case I also think the explicit warning to restart docker is confusing -- it takes a long time to get info on minikube in docker because it doesn't exist, and then we assume the daemon is stuck, which isn't correct.

### DIFF
--- a/pkg/drivers/kic/oci/cli_runner.go
+++ b/pkg/drivers/kic/oci/cli_runner.go
@@ -88,7 +88,7 @@ func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
 	}
 
 	killTime := 19 * time.Second // this will be applied only if warnSlow is true
-	warnTime := 10 * time.Second
+	warnTime := 2 * time.Second
 
 	if cmd.Args[1] == "volume" || cmd.Args[1] == "ps" { // volume and ps requires more time than inspect
 		killTime = 30 * time.Second
--- a/pkg/drivers/kic/oci/cli_runner.go
+++ b/pkg/drivers/kic/oci/cli_runner.go
@@ -88,7 +88,7 @@ func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
 	}
 
 	killTime := 19 * time.Second // this will be applied only if warnSlow is true
-	warnTime := 2 * time.Second
+	warnTime := 10 * time.Second
 	ctx, cancel := context.WithTimeout(context.Background(), killTime)
 	defer cancel()
 
@@ -138,7 +138,7 @@ func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
 		}
 
 		if ctx.Err() == context.DeadlineExceeded {
-			return rr, fmt.Errorf("%q timed out after %s", rr.Command(), killTime)
+			return rr, context.DeadlineExceeded
 		}
 	}
 
--- a/pkg/drivers/kic/oci/cli_runner.go
+++ b/pkg/drivers/kic/oci/cli_runner.go
@@ -89,13 +89,14 @@ func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
 
 	killTime := 19 * time.Second // this will be applied only if warnSlow is true
 	warnTime := 10 * time.Second
-	ctx, cancel := context.WithTimeout(context.Background(), killTime)
-	defer cancel()
 
 	if cmd.Args[1] == "volume" || cmd.Args[1] == "ps" { // volume and ps requires more time than inspect
 		killTime = 30 * time.Second
 		warnTime = 3 * time.Second
 	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), killTime)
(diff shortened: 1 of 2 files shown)


pkg/drivers/kic/oci/oci.go
```diff
@@ -76,7 +76,7 @@ func DeleteContainer(ociBin string, name string) error {
 
 	_, err := ContainerStatus(ociBin, name)
 	if err != nil {
-		glog.Errorf("%s daemon seems to be stuck. Please try restarting your %s. Will try to delete anyways: %v", ociBin, ociBin, err)
+		glog.Infof("%s daemon seems to be stuck. Will try to delete anyways: %v", ociBin, err)
```
medyagh, line 79: this is a good opportunity for a refactor,  this log is misleading because it could be NOT the case that the docker is stuck.
the only case when docker is stuck is when "docker inspect" cmomand times out ( for safty we could wait up to 13 second but if it is not returning 2 seconds it is totally stuck)
in that case we need to return an error type called DockerStuckNeedsRestart

and we should return that error to the Exit usage
in other cases we should log it as warnning
I recommend that we make ContainerStatus to return better error type

medyagh: aslo I recommend using warrningf instead of info. warrning would not show up to the user

priyawadhwa: sounds good, we now log it as a warning if docker inspect takes >10 seconds to respond

pkg/drivers/kic/oci/cli_runner.go
```diff
@@ -88,15 +88,16 @@ func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
 	}
 
 	killTime := 19 * time.Second // this will be applied only if warnSlow is true
-	warnTime := 2 * time.Second
-	ctx, cancel := context.WithTimeout(context.Background(), killTime)
-	defer cancel()
+	warnTime := 10 * time.Second
```
medyagh, line 91: we could have two warnTimes based on the command, I would leave the default warnTime for other commands as 2 seconds, as if some commands take more than 2 seconds user needs to know there is something wrong with their docker

pkg/drivers/kic/oci/oci.go
```diff
@@ -73,10 +75,11 @@ func DeleteContainersByLabel(ociBin string, label string) []error {
 
 // DeleteContainer deletes a container by ID or Name
 func DeleteContainer(ociBin string, name string) error {
-
 	_, err := ContainerStatus(ociBin, name)
-	if err != nil {
-		glog.Errorf("%s daemon seems to be stuck. Please try restarting your %s. Will try to delete anyways: %v", ociBin, ociBin, err)
+	if err == context.DeadlineExceeded {
+		out.WarningT("{{.ocibin}} daemon is taking an unsually long time to respond, consider restarting {{.ocibin}}", out.V{"ociBin": ociBin})
```
medyagh, line 80: {{.ocibin}} daemon --> {{.ocibin}}
podman doesn't have a daemon

pkg/drivers/kic/oci/cli_runner.go
priyawadhwa: I realized we don't actually need to change the warnTime -- the issue isn't that warnTime was too low, it was that we were printing out "please restart docker" on any error, instead of only on a DeadineExceeded error

I reverted this back to how it was

pkg/drivers/kic/oci/oci.go
priyawadhwa: thanks, done

PR #8342, opened 2020-06-01
Set `failSwapOn` and `clusterDomain` via kubelet config file

prasadkatti:
This will avoid these two warnings in kubelet logs -
```
minikube kubelet[3637]: Flag --cluster-domain has been deprecated ...
minikube kubelet[3637]: Flag --fail-swap-on has been deprecated ...
```
See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
These configurations can still be overridden using the `--extra-options`. When options for kubelet are provided using `--extra-options`, they end up on the kubelet command line which takes precedence over values in config file.

**Notes for the reviewers:** There are 76 changed files in this PR, but 72 of those are just changes to testdata. The files you want to review are -
- 3 kubeadm config files (`pkg/minikube/bootstrapper/bsutil/ktmpl/{v1alpha3.go,v1beta1.go,v1beta2.go}`
- `pkg/minikube/bootstrapper/bsutil/versions.go`

There are still some other warnings for the same reason -
```
minikube kubelet[3637]: Flag --authorization-mode has been deprecated ...
minikube kubelet[3637]: Flag --cgroup-driver has been deprecated ...
minikube kubelet[3637]: Flag --client-ca-file has been deprecated ...
minikube kubelet[3637]: Flag --pod-manifest-path has been deprecated ...
```

partially fixes #8380

PR #8468, opened 2020-06-13
Move kubelet params to config file

### LINKED ISSUE #8380: Move kubelet parameters to config file
Currently, the kubelet parameters are passed in as command line arguments. However, for most of the parameters this was [deprecated in v1.10](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/). As part of that, if those parameters continue to be passed on the command-line, that generates warnings in the kubelet logs, like these -

> minikube kubelet[3637]: Flag --authorization-mode has been deprecated ...
minikube kubelet[3637]: Flag --cgroup-driver has been deprecated ...
minikube kubelet[3637]: Flag --client-ca-file has been deprecated ...
minikube kubelet[3637]: Flag --pod-manifest-path has been deprecated ...
minikube kubelet[3637]: Flag --cluster-domain has been deprecated ...
minikube kubelet[3637]: Flag --fail-swap-on has been deprecated ...

The recommended solution is to pass in those parameters using kubelet's config file.

I would like to work on this issue. I already have a PR that partially resolves this - https://github.com/kubernetes/minikube/pull/8342

prasadkatti:
fixes #8380

This is a follow-up to #8342. It takes care of the following parameters which were being passed on the command-line (causing deprecation warnings in kubelet logs) -

1. `--authorization-mode` - removed as it already defaults to Webhook when --config flag is present
1. `--cgroup-driver` - set in the config file instead of cmdline
1. `--client-ca-file` - set in the config file for v1.16+. Before that a kubeadm bug did not allow overriding this via config file. (https://github.com/kubernetes/kubernetes/pull/81903)
1. `--pod-manifest-path` - set in config file instead of cmdline

This takes care of all such warnings in kubelet logs.
>minikube kubelet[3637]: Flag --authorization-mode has been deprecated ...
minikube kubelet[3637]: Flag --cgroup-driver has been deprecated ...
minikube kubelet[3637]: Flag --client-ca-file has been deprecated ...
minikube kubelet[3637]: Flag --pod-manifest-path has been deprecated ...

Tested by running `minikube start` (only for the hyperkit driver) for all minor versions between v1.13 - v1.18

### DIFF
--- a/pkg/minikube/bootstrapper/bsutil/testdata/v1.12/containerd-api-port.yaml
+++ b/pkg/minikube/bootstrapper/bsutil/testdata/v1.12/containerd-api-port.yaml
@@ -40,6 +40,7 @@ networking:
 ---
 apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
+clientCAFile: /var/lib/minikube/certs/ca.crt
 clusterDomain: "cluster.local"
 # disable disk resource management by default
 imageGCHighThresholdPercent: 100
@@ -48,3 +49,4 @@ evictionHard:
   nodefs.inodesFree: "0%"
   imagefs.available: "0%"
 failSwapOn: false
+staticPodPath: /etc/kubernetes/manifests
--- a/pkg/minikube/bootstrapper/bsutil/testdata/v1.12/containerd-api-port.yaml
+++ b/pkg/minikube/bootstrapper/bsutil/testdata/v1.12/containerd-api-port.yaml
@@ -40,7 +40,7 @@ networking:
 ---
 apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
-clientCAFile: /var/lib/minikube/certs/ca.crt
+cgroupDriver: systemd
 clusterDomain: "cluster.local"
 # disable disk resource management by default
 imageGCHighThresholdPercent: 100
--- a/pkg/minikube/bootstrapper/bsutil/ktmpl/v1alpha3.go
+++ b/pkg/minikube/bootstrapper/bsutil/ktmpl/v1alpha3.go
@@ -68,7 +68,7 @@ networking:
 ---
 apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
-clientCAFile: {{.ClientCAFile}}
+cgroupDriver: {{.CgroupDriver}}
 clusterDomain: "cluster.local"
 # disable disk resource management by default
 imageGCHighThresholdPercent: 100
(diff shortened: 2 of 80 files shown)


priyawadhwa:
Hey @prasadkatti looks like `TestStartStop/group/old-k8s-version` is failing across the board.

My guess is that the configuration for `v1alpha3` is different from the configuration for `v1beta1` or `v1beta2`

prasadkatti:
The clientCAFile needs to be nested in the config file -

```
authentication:
  x509:
    clientCAFile: ....
```

prasadkatti:
This was failing only on 1.13 because back then kubeadm was validating component config for kubelet. Starting with 1.17, kubeadm does not validate component config for kubelet. So kubeadm init didn't fail for the other tests.

prasadkatti:
I am passing this as part of the KubeletConfiguration that is provided to kubeadm init.
```yaml
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
authentication:
  anonymous:
    enabled: false
  webhook:
    cacheTTL: 2m0s
    enabled: true
  x509:
    clientCAFile: /var/lib/minikube/certs/ca.crt
```
But the actual kubelet config that I get is different for k8s v1.13 and k8s v1.17. In v1.13, the `clientCAFile` value that is provided does not take effect. Instead it seems to be defaulting to `etc/kubernetes/pki/ca.crt`. 
Trying to dig into this..

prasadkatti:
The code works functionally, but I need some help with the tests.

prasadkatti:
I am not sure if the test failure is a flake or related to the PR.

prasadkatti:
All podman tests are failing on this. I need some help with that. Don't have much experience with podman.

pkg/minikube/bootstrapper/bsutil/testdata/v1.12/containerd-api-port.yaml
```diff
@@ -40,6 +40,7 @@ networking:
 ---
 apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
+cgroupDriver: systemd
```
medyagh, line 43: this doesn't seem to be relevent change to this PR

prasadkatti: I have to set the `cgroupDriver` in the kubelet config file to get rid of this warning in the kubelet log - 

> minikube kubelet[3637]: Flag --cgroup-driver has been deprecated ...

With that change, the tests need to be updated as well.

prasadkatti:
@afbjorklund - Would you be willing to review the podman failures? I am having difficulties setting up a podman environment to test it. The kubelet service is failing to start and I don't have access to the logs to see why.

medyagh:
@prasadkatti do you mind pulling the master on this PR to test with the new changes?

prasadkatti:
Can someone please take a look at the failures (expect for the Podman ones)? I don't think they are related to the changes in the PR.

PR #8469, opened 2020-06-13
Add --cni flag and make multi-node respect CNI options

### LINKED ISSUE #8222: multi-node: unable to choose CNI (always applies kindnet)
Right now we enforce our own CNI for multinode clusters, ignoring any input from the user. We should respect the `--enable-default-cni` parameter and not apply CNI when that's false.

tstromberg:
Fixes #8222

Add support for the following `--cni` options:

* `auto`: default: behave as today. No CNI unless multi-node or Docker driver
* `custom`: Bring your own CNI (was: `--enable-default-cni`)
* `kindnet`: install kindnet
* `flannel`: install flannel
* `calico`: install calico

This handles the following corner cases:

* Adds CNI to a cluster once it becomes a multi-node cluster
* Allows CNI to be specified for Docker and/or multi-node clusters
* Adds pod-cidr to if `--network-plugin=kubenet` (required)


tstromberg:
NOTE: This still needs work for multi-node.

tstromberg:
Thanks! I'm going to mark as WIP until I can address 3 things:

- Add integration tests for each CNI option, specifically testing for hairpin mode.
- Rename the "custom" CNI to "bridge". That's the actual name of the built-in CNI plug-in it uses.
- Add JSON network config for kindnet

I'll un-WIP once the integration tests pass.

pkg/minikube/node/start.go
```diff
@@ -176,6 +177,15 @@ func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
 		if err = bs.JoinCluster(*starter.Cfg, *starter.Node, joinCmd); err != nil {
 			return nil, errors.Wrap(err, "joining cluster")
 		}
+
+		cnm, err := cni.New(*starter.Cfg)
+		if err != nil {
+			return nil, errors.Wrap(err, "cni")
+		}
+
+		if err := cnm.Apply(cpr, []cni.Runner{cpr, starter.Runner}); err != nil {
```
medyagh, line 186: just a comment,  could starter.Runner be used as cpr too ?

pkg/minikube/download/preload.go
```diff
@@ -40,7 +40,7 @@ const (
 	// PreloadVersion is the current version of the preloaded tarball
 	//
 	// NOTE: You may need to bump this version up when upgrading auxiliary docker images
-	PreloadVersion = "v3"
+	PreloadVersion = "v4"
```
medyagh, line 43: if you do this in a separate PR we chould make the preload images and then test this PR with the generated prelaoded imges.

pkg/minikube/node/start.go
tstromberg: `starter.Runner` refers to the newly created node. cpr is the control-plane, which we need to pass in because only the control-plane has kubectl.

pkg/minikube/download/preload.go
tstromberg: I've updated the v4 images - would you prefer that the addition of kindnet as a CNI be split out into a separate PR?

tstromberg:
PTAL. I've fixed kubenet support and added some fairly comprehensive integration tests. Results may need to be tuned for different environments.

medyagh:
The performance bot numbers are alamring !

tstromberg:
I'm going to re-open this with a couple of changes split out.

Issue #8497, opened 2020-06-16
docker: Error response from daemon: Range of CPUs is from 0.01 to 1.00, as there are only 1 CPUs available.


aamirashid:
docker: Error response from daemon: Range of CPUs is from 0.01 to 1.00, as there are only 1 CPUs available.

priyawadhwa:
Hey @aamirashid minikube requires a minimum of 2 CPUs to run, and it seems that only 1 is available in your docker daemon. Is there any chance you could increase the CPUs to 2?

afbjorklund:
This _could_ have a nicer message, like it does for the other drivers ? (same as #7850)

medyagh:
kuberentes needs at least 2 CPUS, @aamirashid 

do you mind sharing the full output of the start
minikube start --driver=docker --alsologtostderr -v=8

we could still provide a better solution message

PR #8499, opened 2020-06-16
docker/podman: add advice for cpu count error 

medyagh:
closes https://github.com/kubernetes/minikube/issues/8497

### Before this PR

```
medya@~/workspace/minikube (sol_msg_cpucount) $ minikube start --driver=docker --cpus 12
😄  minikube v1.11.0 on Darwin 10.15.5
🔥  Creating docker container (CPUs=12, Memory=4000MB) ...
🤦  StartHost failed, but will try again: creating host: create: creating: create kic node: create container: docker run -d -t --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --label mode.minikube.sigs.k8s.io=minikube --volume minikube:/var --cpus=12 --memory=4000mb -e container=docker --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 --publish=127.0.0.1::5000 gcr.io/k8s-minikube/kicbase:v0.0.10@sha256:f58e0c4662bac8a9b5dda7984b185bad8502ade5d9fa364bf2755d636ab51438: exit status 125
stdout:

stderr:
docker: Error response from daemon: Range of CPUs is from 0.01 to 8.00, as there are only 8 CPUs available.
See 'docker run --help'.

🤷  docker "minikube" container is missing, will recreate.
E0616 12:57:26.340125   11245 oci.go:79] docker daemon seems to be stuck. Please try restarting your docker. Will try to delete anyways: unknown state "minikube": docker container inspect minikube --format={{.State.Status}}: exit status 1
stdout:


stderr:
Error: No such container: minikube
🔥  Creating docker container (CPUs=12, Memory=4000MB) ...
😿  Failed to start docker container. "minikube start" may fix it: recreate: creating host: create: creating: create kic node: create container: docker run -d -t --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --label mode.minikube.sigs.k8s.io=minikube --volume minikube:/var --cpus=12 --memory=4000mb -e container=docker --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 --publish=127.0.0.1::5000 gcr.io/k8s-minikube/kicbase:v0.0.10@sha256:f58e0c4662bac8a9b5dda7984b185bad8502ade5d9fa364bf2755d636ab51438: exit status 125
stdout:

stderr:
docker: Error response from daemon: Range of CPUs is from 0.01 to 8.00, as there are only 8 CPUs available.
See 'docker run --help'.


💣  error provisioning host: Failed to start host: recreate: creating host: create: creating: create kic node: create container: docker run -d -t --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run -v /lib/modules:/lib/modules:ro --hostname minikube --name minikube --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --label role.minikube.sigs.k8s.io= --label mode.minikube.sigs.k8s.io=minikube --volume minikube:/var --cpus=12 --memory=4000mb -e container=docker --expose 8443 --publish=127.0.0.1::8443 --publish=127.0.0.1::22 --publish=127.0.0.1::2376 --publish=127.0.0.1::5000 gcr.io/k8s-minikube/kicbase:v0.0.10@sha256:f58e0c4662bac8a9b5dda7984b185bad8502ade5d9fa364bf2755d636ab51438: exit status 125
stdout:

stderr:
docker: Error response from daemon: Range of CPUs is from 0.01 to 8.00, as there are only 8 CPUs available.
See 'docker run --help'.


```

### After this PR

```
😄  minikube v1.11.0 on Darwin 10.15.5
🔥  Creating docker container (CPUs=11, Memory=4000MB) ...

💥  docker doesn't have enough CPUs. 
❗  Please consider changing your Docker desktop's resources.
📘  https://docs.docker.com/config/containers/resource_constraints/
💡  Ensure your docker system has enough CPUs. The minimum allowed is 2 CPUs.
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1051,7 +1051,7 @@ func maybeExitWithAdvice(err error) {
 		out.ErrLn("")
 		out.ErrT(out.Conflict, "{{.name}} doesn't have enough CPUs. ", out.V{"name": viper.GetString("driver")})
 		if runtime.GOOS != "linux" && viper.GetString("driver") == "docker" {
-			out.T(out.Warning, "Please consider changing your Docker desktop's resources.")
+			out.T(out.Warning, "Please consider changing your Docker Desktop's resources.")
 			out.T(out.Documentation, "https://docs.docker.com/config/containers/resource_constraints/")
 		} else {
 			cpuCount := viper.GetInt(cpus)
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1062,7 +1062,7 @@ func maybeExitWithAdvice(err error) {
 			}
 		}
 
-		exit.UsageT("Ensure your system has enough CPUs")
+		exit.UsageT("Ensure your {{.driver_name}} system has enough CPUs. The minimum allowed is 2 CPUs.", out.V{"driver_name": viper.GetString("driver")})
 	}
 
 }
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1039,7 +1039,7 @@ func getKubernetesVersion(old *config.ClusterConfig) string {
 func maybeExitWithAdvice(err error) {
 	if errors.Is(err, oci.ErrWindowsContainers) {
 		out.ErrLn("")
-		out.ErrT(out.Conflict, "Your Docker Desktop container os type is Windows but Linux is required.")
+		out.ErrT(out.Conflict, "Your Docker Desktop container OS type is Windows but Linux is required.")
 		out.T(out.Warning, "Please change Docker settings to use Linux containers instead of Windows containers.")
 		out.T(out.Documentation, "https://minikube.sigs.k8s.io/docs/drivers/docker/#verify-docker-container-type-is-linux")
 		exit.UsageT(`You can verify your Docker container type by running:
@@ -1049,15 +1049,20 @@ func maybeExitWithAdvice(err error) {
 
 	if errors.Is(err, oci.ErrCPUCountLimit) {
 		out.ErrLn("")
-		out.ErrT(out.Conflict, "Your {{.name}} doesn't have enough CPUs. ", out.V{"name": viper.GetString("driver")})
+		out.ErrT(out.Conflict, "{{.name}} doesn't have enough CPUs. ", out.V{"name": viper.GetString("driver")})
 		if runtime.GOOS != "linux" && viper.GetString("driver") == "docker" {
 			out.T(out.Warning, "Please consider changing your Docker desktop's resources.")
 			out.T(out.Documentation, "https://docs.docker.com/config/containers/resource_constraints/")
 		} else {
-			out.T(out.Warning, "Please ensure your system has at least {{.cpu_counts}} CPU cores", out.V{"cpu_counts": viper.GetInt(cpus)})
+			cpuCount := viper.GetInt(cpus)
+			if cpuCount == 2 {
+				out.T(out.Tip, "Please ensure your system has {{.cpu_counts}} CPU cores.", out.V{"cpu_counts": viper.GetInt(cpus)})
+			} else {
+				out.T(out.Tip, "Please ensure your {{.driver_name}} system has access to {{.cpu_counts}} CPU cores or reduce the number of the specified CPUs", out.V{"driver_name": viper.GetString("driver"), "cpu_counts": viper.GetInt(cpus)})
+			}
 		}
 
-		exit.UsageT("Esnure your system has enough CPUs")
+		exit.UsageT("Ensure your system has enough CPUs")
 	}
 
 }
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -155,15 +155,7 @@ func runStart(cmd *cobra.Command, args []string) {
 	ds, alts, specified := selectDriver(existing)
 	starter, err := provisionWithDriver(cmd, ds, existing)
 	if err != nil {
-		if errors.Is(err, oci.ErrWindowsContainers) {
-			out.ErrLn("")
-			out.ErrT(out.Conflict, "Your Docker Desktop container os type is Windows but Linux is required.")
-			out.T(out.Warning, "Please change Docker settings to use Linux containers instead of Windows containers.")
-			out.T(out.Documentation, "https://minikube.sigs.k8s.io/docs/drivers/docker/#verify-docker-container-type-is-linux")
-			exit.UsageT(`You can verify your Docker container type by running:
-	{{.command}}
-		`, out.V{"command": "docker info --format '{{.OSType}}'"})
-		}
+		maybeExitWithAdvice(err)
 		if specified {
 			// If the user specified a driver, don't fallback to anything else
 			exit.WithError("error provisioning host", err)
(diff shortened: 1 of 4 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -1042,3 +1034,30 @@ func getKubernetesVersion(old *config.ClusterConfig) string {
...
+		out.T(out.Warning, "Please change Docker settings to use Linux containers instead of Windows containers.")
+		out.T(out.Documentation, "https://minikube.sigs.k8s.io/docs/drivers/docker/#verify-docker-container-type-is-linux")
+		exit.UsageT(`You can verify your Docker container type by running:
+{{.command}}
+	`, out.V{"command": "docker info --format '{{.OSType}}'"})
+	}
+
+	if errors.Is(err, oci.ErrCPUCountLimit) {
+		out.ErrLn("")
+		out.ErrT(out.Conflict, "Your {{.name}} doesn't have enough CPUs. ", out.V{"name": viper.GetString("driver")})
```
priyawadhwa, line 1052: I would remove "Your" so that we see:

```
docker doesn't have enough CPUs
```

instead of

```
Your docker doensn't have enough CPUs
```

cmd/minikube/cmd/start.go
```diff
@@ -1042,3 +1034,30 @@ func getKubernetesVersion(old *config.ClusterConfig) string {
...
+		if runtime.GOOS != "linux" && viper.GetString("driver") == "docker" {
+			out.T(out.Warning, "Please consider changing your Docker desktop's resources.")
+			out.T(out.Documentation, "https://docs.docker.com/config/containers/resource_constraints/")
+		} else {
+			out.T(out.Warning, "Please ensure your system has at least {{.cpu_counts}} CPU cores", out.V{"cpu_counts": viper.GetInt(cpus)})
+		}
+
+		exit.UsageT("Esnure your system has enough CPUs")
```
priyawadhwa, line 1060: spelling: esnure/ensure

cmd/minikube/cmd/start.go
```diff
@@ -1042,3 +1034,30 @@ func getKubernetesVersion(old *config.ClusterConfig) string {
...
+	}
+
+	if errors.Is(err, oci.ErrCPUCountLimit) {
+		out.ErrLn("")
+		out.ErrT(out.Conflict, "Your {{.name}} doesn't have enough CPUs. ", out.V{"name": viper.GetString("driver")})
+		if runtime.GOOS != "linux" && viper.GetString("driver") == "docker" {
+			out.T(out.Warning, "Please consider changing your Docker desktop's resources.")
+			out.T(out.Documentation, "https://docs.docker.com/config/containers/resource_constraints/")
+		} else {
+			out.T(out.Warning, "Please ensure your system has at least {{.cpu_counts}} CPU cores", out.V{"cpu_counts": viper.GetInt(cpus)})
```
priyawadhwa, line 1057: If a user passes in `--cpus 1`, won't viper.GetInt() return 1?

We probably want to hardcode 2 in here, or have a `MinimumCPUs` variable

pkg/drivers/kic/oci/errors.go
```diff
@@ -20,3 +20,6 @@ import "errors"
 
 // ErrWindowsContainers is thrown when docker been configured to run windows containers instead of Linux
 var ErrWindowsContainers = errors.New("docker container type is windows")
+
+// ErrCPUCountLimit is thrown when docker daemon doesn't have enough CPUs for the requested container
+var ErrCPUCountLimit = errors.New("not enough CPUs is available for container")
```
priyawadhwa, line 25: You could create a `FailFastError` type to wrap around these errors, so that we don't need to check every type of error in `node/start.go`

```
type FailFastError error
var ErrWindowsContainers = FailFastError(errors.New("docker container type is windows"))
var ErrCPUCountLimit = FailFastError(errors.New("not enough CPUs is available for container"))
```

pkg/minikube/node/start.go
```diff
@@ -366,6 +366,12 @@ func startHost(api libmachine.API, cc *config.ClusterConfig, n *config.Node) (*h
 		return host, exists, err
 	}
 
+	// don't try to re-create if cpu count is not enough
```
priyawadhwa, line 369: Instead of lines 364-374, if we include the `FailFastError` from above we could replace these lines with:

```go
	if _, ff := err.(FailFastError); ff {
		glog.Infof("will skip retrying to create machine because error is not retriable: %v", err)
		return host, exists, err
	}

```

cmd/minikube/cmd/start.go
priyawadhwa: I'd also include that we require 2 here

```
docker doesn't have access to 2 CPUs, which is the minimum required for minikube
```

medyagh: I was thinking if user specifies more than 2 like 8 the solution message would no be relevent but I could improve by adding another if it is 2

pkg/drivers/kic/oci/errors.go
medyagh: Nice tip!

cmd/minikube/cmd/start.go
medyagh: for cpu 1 we already check for it in the validation,  so they user can not set it to 1 in first place
```
$ ./out/minikube start --driver=docker --cpus 1
😄  minikube v1.11.0 on Darwin 10.15.5
💡  Requested cpu count 1 is less than the minimum allowed of 2
```

pkg/drivers/kic/oci/errors.go
medyagh: done

cmd/minikube/cmd/start.go
medyagh: done

cmd/minikube/cmd/start.go
```diff
@@ -1042,3 +1034,35 @@ func getKubernetesVersion(old *config.ClusterConfig) string {
...
+		exit.UsageT(`You can verify your Docker container type by running:
+{{.command}}
+	`, out.V{"command": "docker info --format '{{.OSType}}'"})
+	}
+
+	if errors.Is(err, oci.ErrCPUCountLimit) {
+		out.ErrLn("")
+		out.ErrT(out.Conflict, "{{.name}} doesn't have enough CPUs. ", out.V{"name": viper.GetString("driver")})
+		if runtime.GOOS != "linux" && viper.GetString("driver") == "docker" {
+			out.T(out.Warning, "Please consider changing your Docker desktop's resources.")
```
priyawadhwa, line 1054: nit: desktop should be Desktop

Issue #8179, opened 2020-05-16
docker driver: don't mount to /var as it causes race condition


polzka90:
**Steps to reproduce the issue:** 

Im following the official guide to use minikube and i am getting an error inside the container image

INFO: ensuring we can execute /bin/mount even with userns-remap

INFO: remounting /sys read-only

INFO: making mounts shared

INFO: fix cgroup mounts for all subsystems

INFO: clearing and regenerating /etc/machine-id

Initializing machine ID from random generator.

INFO: faking /sys/class/dmi/id/product_name to be "kind"

INFO: faking /sys/class/dmi/id/product_uuid to be random

INFO: faking /sys/devices/virtual/dmi/id/product_uuid as well

INFO: setting iptables to detected mode: legacy

update-alternatives: error: no alternatives for iptables

i use windows 10 and my docker desktop version is  2.3.0.2

**Full output of failed command:** 
I0516 17:24:50.867547   12524 start.go:99] hostinfo: {"hostname":"DESKTOP-CFDRBNO","uptime":28728,"bootTime":1589631962,"procs":243,"os":"windows","platform":"Microsoft Windows 10 Enterprise","platformFamily":"Standalone Workstation","platformVersion":"10.0.18362 Build 18362","kernelVersion":"","virtualizationSystem":"","virtualizationRole":"","hostid":"d16cd74e-4922-43f3-8935-a8a21e85fb43"}
W0516 17:24:50.867547   12524 start.go:107] gopshost.Virtualization returned error: not implemented yet
* minikube v1.10.1 on Microsoft Windows 10 Enterprise 10.0.18362 Build 18362
I0516 17:24:50.881545   12524 notify.go:125] Checking for updates...
I0516 17:24:50.884546   12524 driver.go:253] Setting default libvirt URI to qemu:///system
I0516 17:24:51.279998   12524 docker.go:95] docker version: linux-19.03.8
* Using the docker driver based on existing profile
I0516 17:24:51.285999   12524 start.go:215] selected driver: docker
[...]

Issue #8163, opened 2020-05-15
docker inspect -f "'{{(index (index .NetworkSettings.Ports "22/tcp") 0).HostPort}}'" minikube: exit status 1


jenkins-kiwi:
**Steps to reproduce the issue:** 

1.
2. 
3.

**Full output of failed command:** 

**Full output of `minikube start` command used, if not already included:**

**Optional: Full output of `minikube logs` command:**
<details>

</details>

medyagh:
Thank you for sharing your experience! If you don't mind, could you please provide:

- The exact command-lines used, so that we may replicate the issue
- The version of minikube.
- The driver you are using.
- The Operation system you are using.
- The output of the "minikube logs" 

This will help us isolate the problem further. Thank you!

AhireSwati:
I too am getting the same issue.
`stderr:
Template parsing error: template: :1:4: executing "" at <index (index .NetworkSettings.Ports "22/tcp") 0>: error calling index: index of untyped nil

* Restarting existing docker container for "minikube" ...
* Failed to start docker container. "minikube start" may fix it: provision: get ssh host-port: get port 22 for "minikube": docker inspect -f "'{{(index (index .NetworkSettings.Ports "22/tcp") 0).HostPort}}'" minikube: exit status 1
stdout:
`

tstromberg:
If someone runs into this, can they please share the output of:

`docker inspect minikube`

Thank you.

j75:
* The exact command-lines used, so that we may replicate the issue => **minikube start --driver='docker'**
  * The version of minikube : **minikube version: v1.10.1
commit: 63ab801ac27e5742ae442ce36dff7877dcccb278**
  * The driver you are using: **docker**
  * The Operation system you are using: **Debian Buster (10.04)**
  * The output of the "minikube logs" 
``` INFO: ensuring we can execute /bin/mount even with userns-remap
 INFO: remounting /sys read-only
 INFO: making mounts shared
 INFO: fix cgroup mounts for all subsystems
 INFO: clearing and regenerating /etc/machine-id
 Initializing machine ID from random generator.
 INFO: faking /sys/class/dmi/id/product_name to be "kind"
 INFO: setting iptables to detected mode: legacy
 update-alternatives: error: no alternatives for iptables
 INFO: ensuring we can execute /bin/mount even with userns-remap
 INFO: remounting /sys read-only
 INFO: making mounts shared
 INFO: fix cgroup mounts for all subsystems
[...]

Issue #8151, opened 2020-05-14
Make the persistent storage actually survive reboots


afbjorklund:
For the KIC (and for "none"), we need to make sure that the persistant directories are preserved:

https://minikube.sigs.k8s.io/docs/handbook/persistent_volumes/

These are persisted today:
```
/var/lib/minikube
/var/lib/docker
/var/lib/containers
```
(the third one is for crio, and for podman)

These also need to be kept:
```
/data
/tmp/hostpath_pv
/tmp/hostpath-provisioner
```
(I don't think the third one is used anymore)

We can put them on the docker/podman volume, and at least _mention_ them in the "none" docs...

They are normally bind mounts, to some place that is persistent (like disk image or  docker volume)

```
TARGET                    SOURCE                           FSTYPE OPTIONS
/tmp/hostpath_pv          /dev/sda1[/hostpath_pv]          ext4   rw,relatime
/tmp/hostpath-provisioner /dev/sda1[/hostpath-provisioner] ext4   rw,relatime
/mnt/sda1                 /dev/sda1                        ext4   rw,relatime
/var/lib/docker           /dev/sda1[/var/lib/docker]       ext4   rw,relatime
/var/lib/containers       /dev/sda1[/var/lib/containers]   ext4   rw,relatime
/data                     /dev/sda1[/data]                 ext4   rw,relatime
/var/lib/minikube         /dev/sda1[/var/lib/minikube]     ext4   rw,relatime
```

A nice change to go with this one, would be to stop volume mounting all of `/var` (at the top level).

Because it conflicts with existing paths, in the ubuntu image. Like `/var/lib/dpkg/alternatives`.

See #8056 and #8100

The kicbase doesn't boot without it.

afbjorklund:
Suggestion is to add a compatibility symlink, for the existing docker/podman volumes:

`lib -> var/lib`

And then move the content down a level, before doing so. Same with "tmp" and "logs".

`tmp -> var/tmp`
`logs -> var/logs`

afbjorklund:
The other stuff in the current volume, we don't have to care about:

`backups  cache	lib  local  lock  log  mail  opt  run  spool  tmp`

Most of it is empty anyway.

```
backups/

0 directories, 0 files
```

```
cache/
├── apt
│   └── archives
│       ├── lock
│       └── partial
├── debconf
│   ├── config.dat
│   ├── passwords.dat
│   └── templates.dat
├── ldconfig
│   └── aux-cache
└── private

[...]

PR #8509, opened 2020-06-17
wip: Move docker root dir outside of /var and mount preload to that instead

priyawadhwa:
I think this could potentially fix https://github.com/kubernetes/minikube/issues/8179 and https://github.com/kubernetes/minikube/issues/8163

Still looking into it


afbjorklund:
Hi @priyawadhwa 

I'm not sure I like using `/tmp/var` any more than using `/var/tmp` (as in the storage-provisioner PR)...
And there are **lots** of places that will need changing, if we move those locations away from their defaults.

My preferred approach was to handle the container image in the same was as we already do with the VM.

That is, to use one mount for the disk image. And then do bind mounts to this shared location.
That way we avoid overloading all of `/var`, but still persist the files to disk (the docker volume).

We would need some kind of "minikube-automount" service for this though, so that it happens automatically.

Talked about it in #8151 and #7511

PR #8518, opened 2020-06-20
Make sure to use internal containerd for docker

### LINKED ISSUE #8203: Race condition between dockerd and containerd
Currently the VM startup procedure relies on Docker starting an internal version of containerd.

But depending on timing, dockerd _can_ choose to start using the systemd version instead...

See https://github.com/moby/moby/blob/v19.03.8/cmd/dockerd/daemon_unix.go#L155

This leads to the docker daemon not responding, when minikube shuts down system containerd.

----

We should probably do like the ubuntu packaging, and have docker.service use containerd.service

`BindsTo=containerd.service` and `dockerd --containerd=/run/containerd/containerd.sock`

afbjorklund:
When not using the containerd.service, we should make
sure that dockerd does not use the stopped containerd.

Closes #8203

### DIFF
--- a/pkg/minikube/cruntime/cruntime_test.go
+++ b/pkg/minikube/cruntime/cruntime_test.go
@@ -572,7 +572,7 @@ func TestEnable(t *testing.T) {
 		want    map[string]serviceState
 	}{
 		{"docker", map[string]serviceState{
-			"docker":        SvcRestarted,
+			"docker":        SvcRunning,
 			"containerd":    SvcExited,
 			"crio":          SvcExited,
 			"crio-shutdown": SvcExited,
--- a/pkg/minikube/cruntime/cruntime_test.go
+++ b/pkg/minikube/cruntime/cruntime_test.go
@@ -536,6 +536,14 @@ var defaultServices = map[string]serviceState{
 	"containerd":    SvcExited,
 }
 
+// allServices reflects the state of all actual services running at once
+var allServices = map[string]serviceState{
+	"docker":        SvcRunning,
+	"crio":          SvcRunning,
+	"crio-shutdown": SvcExited,
+	"containerd":    SvcRunning,
+}
+
 func TestDisable(t *testing.T) {
 	var tests = []struct {
 		runtime string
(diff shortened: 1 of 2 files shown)


afbjorklund:
Here was the sequence being used:

```
Run: sudo systemctl cat docker.service
Run: sudo systemctl is-active --quiet service containerd
Run: sudo systemctl stop -f containerd
Run: sudo systemctl is-active --quiet service containerd
Run: sudo systemctl is-active --quiet service crio
Run: sudo systemctl stop -f crio
Run: sudo systemctl is-active --quiet service crio
Run: sudo systemctl daemon-reload
Run: sudo systemctl start docker
```

When starting up dockerd with containerd running, it must be restarted (not started) after stopping containerd.

Otherwise it will still be using /run/containerd/containerd.sock instead of /run/docker/containerd/containerd.sock

`Error while dialing dial unix:///run/containerd/containerd.sock`

Only do this when not configured to use the containerd.socket.

The new sequence will instead be:

```
Run: sudo systemctl cat docker.service
Run: sudo systemctl is-active --quiet service containerd
Run: sudo systemctl stop -f containerd
Run: sudo systemctl is-active --quiet service containerd
Run: sudo systemctl is-active --quiet service crio
Run: sudo systemctl stop -f crio
Run: sudo systemctl is-active --quiet service crio
Run: sudo systemctl cat docker.service
Run: sudo systemctl daemon-reload
Run: sudo systemctl restart docker
```

afbjorklund:
It would be possible to "optimize" this, by looking to see if the containerd was Active before it was Disabled.
If it (containerd) was stopped _before_ dockerd was started, then the internal daemon would have been used.

But it is not worth the special case, and there some edge-cases (really need to be looking at what it uses)
And it's kinda stupid anyway, so the way to really optimize is to only start the services that are needed/wanted.

At least this way it doesn't just hang at boot, even if it might (?) take 10 seconds or so longer.

When docker.service is changed to used containerd.service instead, it will not have to restart.

medyagh:
Both KVM and Virtualbox fail on **TestPause/serial/SecondStartNoReconfiguration**

KVM:
https://storage.googleapis.com/minikube-builds/logs/8518/22e1a41/KVM_Linux.html#fail_TestPause%2fserial%2fSecondStartNoReconfiguration

Virtualbox:
https://storage.googleapis.com/minikube-builds/logs/8518/22e1a41/VirtualBox_Linux.html#fail_TestPause%2fserial%2fSecondStartNoReconfiguration

afbjorklund:
It was stopping due to the docker restart, now there should only be one restart (the first time) so should be fixed. Added a unit test.

Issue #7642, opened 2020-04-13
direct user to using minikube's inner kubectl when not installed


medyagh:
on a system without kubectl:

```
med@xmac:~$ minikube start 
😄  minikube v1.9.2 on Ubuntu 20.04
✨  Automatically selected the docker driver
👍  Starting control plane node m01 in cluster minikube
🚜  Pulling base image ...
💾  Downloading Kubernetes v1.18.0 preload ...
    > preloaded-images-k8s-v2-v1.18.0-docker-overlay2-amd64.tar.lz4: 542.91 MiB
🔥  Creating Kubernetes in docker container with (CPUs=2) (8 available), Memory=3900MB (15990MB available) ...
🐳  Preparing Kubernetes v1.18.0 on Docker 19.03.2 ...
    ▪ kubeadm.pod-network-cidr=10.244.0.0/16
🌟  Enabling addons: default-storageclass, storage-provisioner
🏄  Done! kubectl is now configured to use "minikube"
💡  For best results, install kubectl: https://kubernetes.io/docs/tasks/tools/install-kubectl/

```

instead of saying ```💡  For best results, install kubectl: https://kubernetes.io/docs/tasks/tools/install-kubectl/``` 
we could provide them a link to docs (and add nice docs in website) how to use kubectl inside minikube

minikube kubectl -- get pods ...

and then then in the end say , alternatively install kubectl

tstromberg:
I don't know - should `minikube kubectl`  really be considered the ideal outcome?

I do agree that we should mention it as an alternative, similar to what we do with a version mismatch.

9kranti:
I'm also getting same output and before this I was getting control plane node as `minikube` and now I'm getting it as `m01`
Why is this node changed? Then where is `minikube node`?

afbjorklund:
> I'm also getting same output and before this I was getting control plane node as `minikube` and now I'm getting it as `m01`
> Why is this node changed? Then where is `minikube node`?

It is not possible to understand what you mean, or how this is related to `kubectl`

Can you open a new issue instead ? It seems to be somewhat related to multi-node

Harkishen-Singh:
I will like to take up this.

Issue #7289, opened 2020-03-27
kubectl: remove -- leading flag requirement


tstromberg:
It's not fun from a usability point of view.

NOTE: This may be challenging, and require intercepting "kubectl" as a subcommand before Cobra gets it's hands on it.

afbjorklund:
One suggested variant was adding a command to install it, but we didn't want to do that (~#4697~).

Another approach would be add a shell alias, but that has different syntax on different shells...

```console
$ alias mk='minikube kubectl --'
$ mk version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:50:46Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```

Maybe it could be a feature of the `shell` module ? Still has to be called with e.g. `eval` though

Also need to decide on a name, like `minikube.kubectl` - since `k` or `mk` might conflict with user.

PR #8519, opened 2020-06-20
Adds better information for using kubectl inside minikube

Harkishen-Singh:
Signed-off-by: Harkishen-Singh <harkishensingh@hotmail.com>

Fixes: #7642

### DIFF
--- a/site/content/en/docs/handbook/kubectl.md
+++ b/site/content/en/docs/handbook/kubectl.md
@@ -1,0 +1,34 @@
+---
+title: "Kubectl"
+weight: 2
+description: >
+  Use kubectl inside minikube
+aliases:
+  - /docs/kubectl/
+---
+
+By default, [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) gets configured to access the kubernetes cluster control plane
+inside minikube when the `minikube start` command is executed. 
+
+However if `kubectl` is not installed locally, kubectl can be used inside the minikube
+as well.
+
+`minikube kubectl -- <kubectl commands>`
+
+You can also `alias kubectl=minikube kubectl --` for easier usage.
+
+Get pods
+
+`minikube kubectl -- get pods`
+
+Creating a deployment inside kubernetes cluster
+
+`minikube kubectl -- create deployment hello-minikube --image=k8s.gcr.io/echoserver:1.4`
+
+Exposing the deployment with a NodePort service
+
+`minikube kubectl -- expose deployment hello-minikube --type=NodePort --port=8080`
+
+For more help
+
+`minikube kubectl -- --help`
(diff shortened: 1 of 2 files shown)


afbjorklund:
Not sure I agree with "workaround", but I guess it is the best we can do here since we don't agree on the recommendation.

site/content/en/docs/handbook/kubectl.md
```diff
@@ -0,0 +1,32 @@
...
+weight: 2
+description: >
+  Use kubectl inside minikube
+aliases:
+  - /docs/kubectl/
+---
+
+By default, kubectl gets configured to access the kubernetes cluster control plane
+inside minikube when the `minikube start` command is executed. 
+
```
afbjorklund, line 12: Maybe add a link to the kubectl documentation here as well ?

https://kubernetes.io/docs/tasks/tools/install-kubectl/

Harkishen-Singh: I don't think that might be the best place for the link. Would be better to have that in the last of the document. Is that agreeable?

Harkishen-Singh: Like, `For more help related to kubectl, visit the [docs](https://kubernetes.io/docs/tasks/tools/install-kubectl/).`

Harkishen-Singh: Got it. Done!

afbjorklund:
If you want to add additional tips to the documentation, you could add the "alias" as in #7289

It would need the regular warnings and caveats, for people using inferior shells and such...

But avoids having to type out the full `minikube kubectl --` every time:

```bash
alias kubectl='minikube kubectl --'
```

Harkishen-Singh:
> If you want to add additional tips to the documentation, you could add the "alias" as in #7289

That would be great!

Harkishen-Singh:
> If you want to add additional tips to the documentation, you could add the "alias" as in #7289

That would be great! However, according to that issue, do we aim to remove the `--` support? If yes, then at that point, we need to change this doc as well.

afbjorklund:
I'm not sure it is possible to change the flag parsing of subcommands, without changing a lot of minikube internals (cobra/viper)

However, if the user don't have to type it anyway - does it really matter ? We _could_ have a command to print out the path as well.
If you prefer running `~/.minikube/cache/linux/v1.18.3/kubectl` over the wrapper (which does handle os and version for you)
Adding that directory to the PATH would be pretty much the same as "installing" it. Well, without the `sudo` requirement that is.

afbjorklund:
Typically, the logging of the path to the `kubectl` binary was broken: #8521

Harkishen-Singh:
@afbjorklund for now, I have added the `alias` as a tip for easier usage.

PR #8531, opened 2020-06-23
Do not output usage hint when shell=none.

jtzwu:
Since there is no shell, there is no clear way to expect how a usage hint will be interpretted.


tstromberg:
Thank you!

afbjorklund:
This is for the special hack of `SHELL=none` (whatever that means), rather than `SHELL=""` (or no $SHELL)

See a02541d18bf6fb9110e6b8efab6f76fc27fa6514

Basically a text-based output, key=value. However, I don't seem to get any usage hint with v1.11.0 either ?

afbjorklund:
For normal shells, the $SHELL variable will actually be the _full path_ to the shell. Like `/bin/bash` or so.

PR #8533, opened 2020-06-23
add unit test for mustload pkg

n0npax:
This was part of early-stage changes in https://github.com/kubernetes/minikube/pull/8413 but it's not connected anymore. Raising separate PR, as may be useful

### DIFF
--- a/pkg/minikube/config/config.go
+++ b/pkg/minikube/config/config.go
@@ -150,8 +150,8 @@ func encode(w io.Writer, m MinikubeConfig) error {
 }
 
 // Load loads the Kubernetes and machine config for the current machine
-func Load(profile string) (*ClusterConfig, error) {
-	return DefaultLoader.LoadConfigFromFile(profile)
+func Load(profile string, miniHome ...string) (*ClusterConfig, error) {
+	return DefaultLoader.LoadConfigFromFile(profile, miniHome...)
 }
 
 // Write writes the Kubernetes and machine config for the current machine
--- a/pkg/minikube/mustload/mustload_test.go
+++ b/pkg/minikube/mustload/mustload_test.go
@@ -17,11 +17,12 @@ limitations under the License.
 package mustload
 
 import (
+	"path/filepath"
 	"testing"
 )
 
 func TestPartial(t *testing.T) {
-	path := "../config/testdata/profile/.minikube"
+	path := filepath.Join("..", "config", "testdata", "profile", ".minikube")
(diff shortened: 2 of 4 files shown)


pkg/minikube/mustload/mustload_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+package mustload
+
+import (
+	"os"
+	"testing"
+)
+
+func TestPartial(t *testing.T) {
+	path := "../config/testdata/profile/.minikube"
```
n0npax, line 25: not sure if this is acceptable.
If not, do you want me to keep same/simillar profile in `pkh/minikube/mustload/testdata` ?

pkg/minikube/mustload/mustload_test.go
```diff
@@ -0,0 +1,35 @@
...
+	os.Setenv("MINIKUBE_HOME", path)
```
medyagh, line 26: I suggest passing the MINIKUBE_HOME as an optional parameter to the Partial, simmilar to this https://github.com/medyagh/minikube/blob/3d09d8e6ae4dcefcf0d630edaee632d522092734/pkg/minikube/config/profile.go#L108-L109


setting environment variable in testings, will cause race condition since we run many tests in parallel.

pkg/minikube/mustload/mustload_test.go
```diff
@@ -0,0 +1,34 @@
...
+*/
+
+package mustload
+
+import (
+	"testing"
+)
+
+func TestPartial(t *testing.T) {
+	path := "../config/testdata/profile/.minikube"
```
medyagh, line 24: make this windows-campatiple , use a lib like filepath

pkg/minikube/config/config.go
```diff
@@ -150,8 +150,8 @@ func encode(w io.Writer, m MinikubeConfig) error {
 }
 
 // Load loads the Kubernetes and machine config for the current machine
-func Load(profile string) (*ClusterConfig, error) {
-	return DefaultLoader.LoadConfigFromFile(profile)
+func Load(profile string, miniHome ...string) (*ClusterConfig, error) {
```
priyawadhwa, line 153: Was this added so that you could pass in `miniHome` for the unit test?

n0npax: yes, as this comment https://github.com/kubernetes/minikube/pull/8533#discussion_r444372072

PR #8537, opened 2020-06-23
Automatically limit integration test parallelism

tstromberg:
This addresses CPU starvation issues I consistently ran into while running CNI tests.

* Go normally runs 1 test per CPU core.
* Each minikube cluster can consume up to 2 CPU cores by default
* minikube binary and integration tests have additional modest overhead
* Current default can cause extreme CPU contention for highly parallel subtests, resulting in timeouts
* This PR effective limits test parallelism to ~1.75 cores

NOTE: This renames `main.go` to `main_test.go`, so that `TestMain` is actually executed. To make the linter happy with symbol resolution, this meant renaming other similar test helper files as well.


test/integration/main_test.go
```diff
@@ -51,12 +53,44 @@ const (
...
+func setMaxParallelism() {
+	procs := runtime.GOMAXPROCS(0)
+	pval := flag.Lookup("test.parallel").Value.String()
+	pv, err := strconv.Atoi(pval)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "unable to parse --test.parallel value: %q\n", pval)
+		return
+	}
+
+	if procs != pv {
```
medyagh, line 74: was this meant to be if proc < pv {  } ?

tstromberg: This is intentional: If someone tests manually --test-parallel, we preserve the value rather than ignoring them..

test/integration/testdata/weavenet.yaml
```diff
@@ -0,0 +1,255 @@
+apiVersion: v1
```
medyagh, line 1: are these yaml files unrelated changes ?

tstromberg: Yes. They have now been removed.

medyagh:
looks good but lets wait for VM based tests

tstromberg:
The sample size is small, but this appears to increase the total integration run times slightly:

* Docker: was ~23 minutes, now ~26 minutes
* None: was ~14 minutes, now ~15 minutes

On the other hand, it does make it possible for me to run the CNI tests locally.

medyagh:
> The sample size is small, but this appears to increase the total integration run times slightly:
> 
> * Docker: was ~23 minutes, now ~26 minutes
> * None: was ~14 minutes, now ~15 minutes
> 
> On the other hand, it does make it possible for me to run the CNI tests locally.

I am okay with that given that I think we will see less CPU related flakes

PR #8538, opened 2020-06-23
integration: add more MacOS test to github actions

medyagh:
- adding more macos tests
- shuffle tests and combine them in a way they finish together. since multi node test takes longer in mac, now multinode test goes by itself

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -1,8 +1,6 @@
-name: Master
+name: PR
 on:
-  push:
-    branches:
-      - master
+  pull_request:
     paths:
       - "**.go"
       - "**.yml"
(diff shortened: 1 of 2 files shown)


.github/workflows/master.yml
```diff
@@ -719,7 +797,80 @@ jobs:
           echo "*** $numPass Passed ***"
           if [ "$numFail" -gt 0 ];then echo "*** $numFail Failed ***";exit 2;fi
           if [ "$numPass" -eq 0 ];then echo "*** 0 Passed! ***";exit 2;fi
-          if [ "$numPass" -lt 12 ];then echo "*** Failed to pass at least 6 ! ***";exit 2;fi
+          if [ "$numPass" -lt 12 ];then echo "*** Failed to pass at least 12 ! ***";exit 2;fi
+          if [ "$numPass" -eq 0 ];then echo "*** Passed! ***";exit 0;fi
```
sharifelgamal, line 801: Is this line just in case we skip everything? I don't think we'll ever get here because of the previous line.

medyagh: this happens when test times out without any failure, like 2 test would be passing and it timed out (witout any failure) so it would give u a false sense of success

Issue #6167, opened 2019-12-27
unsupported graph driver: btrfs


thorstenhirsch:
**The exact command to reproduce the issue**:
```$ sudo minikube start --vm-driver=none```

**The full output of the command that failed**:<details>
```
😄  minikube v1.6.2 on Arch 
✨  Selecting 'none' driver from user configuration (alternates: [virtualbox])
🤹  Running on localhost (CPUs=12, Memory=16025MB, Disk=474830MB) ...
ℹ️   OS release is Arch Linux
⚠️  VM may be unable to resolve external DNS records
🐳  Preparing Kubernetes v1.17.0 on Docker '19.03.5-ce' ...
    ▪ kubelet.resolv-conf=/run/systemd/resolve/resolv.conf
🚜  Pulling images ...
🚀  Launching Kubernetes ... 

💣  Error starting cluster: init failed. cmd: "/bin/bash -c \"sudo env PATH=/var/lib/minikube/binaries/v1.17.0:$PATH kubeadm init --config /var/tmp/minikube/kubeadm.yaml  --ignore-preflight-errors=DirAvailable--etc-kubernetes-manifests,DirAvailable--var-lib-minikube,DirAvailable--var-lib-minikube-etcd,FileAvailable--etc-kubernetes-manifests-kube-scheduler.yaml,FileAvailable--etc-kubernetes-manifests-kube-apiserver.yaml,FileAvailable--etc-kubernetes-manifests-kube-controller-manager.yaml,FileAvailable--etc-kubernetes-manifests-etcd.yaml,Port-10250,Swap\"": /bin/bash -c "sudo env PATH=/var/lib/minikube/binaries/v1.17.0:$PATH kubeadm init --config /var/tmp/minikube/kubeadm.yaml  --ignore-preflight-errors=DirAvailable--etc-kubernetes-manifests,DirAvailable--var-lib-minikube,DirAvailable--var-lib-minikube-etcd,FileAvailable--etc-kubernetes-manifests-kube-scheduler.yaml,FileAvailable--etc-kubernetes-manifests-kube-apiserver.yaml,FileAvailable--etc-kubernetes-manifests-kube-controller-manager.yaml,FileAvailable--etc-kubernetes-manifests-etcd.yaml,Port-10250,Swap": exit status 1
stdout:
[init] Using Kubernetes version: v1.17.0
[preflight] Running pre-flight checks
[preflight] The system verification failed. Printing the output from the verification:
[...]

PR #8541, opened 2020-06-23
docker driver on linux: Give warning if overlay module is not enabled

medyagh:
Currently we install the docker inside minikube with overlayfs storage driver.
and it is not currently on our roadmap to support others, and and if you are on linux and your linux kernel does not have overlayfs then it will have significant  slow performances 
### after this PR:
(ignore that it is running on darwin I fliped the If statement to make error in darwin)
this should only add warnning for linux, since docker-machine comes with overlayfs on mac and windows
```
😄  minikube v1.12.0-beta.0 on Darwin 10.15.5

❗  'docker' driver reported a issue that could affect the performance.
💡  Suggestion: enable overlayfs kernel module on your Linux

🐳  Preparing Kubernetes v1.18.3 on Docker 19.03.2 ...
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -616,6 +616,13 @@ func validateDriver(ds registry.DriverState, existing *config.ClusterConfig) {
 	st := ds.State
 	glog.Infof("status for %s: %+v", name, st)
 
+	if st.NeedsImprovement { // warn but don't exit
+		out.ErrLn("")
+		out.WarningT("'{{.driver}}' driver reported a issue that could affect the performance.", out.V{"driver": name})
+		out.ErrT(out.Tip, "Suggestion: {{.fix}}", out.V{"fix": translate.T(st.Fix)})
+		out.ErrLn("")
+	}
+
 	if st.Error != nil {
 		out.ErrLn("")
 
--- a/pkg/minikube/registry/drvs/docker/docker.go
+++ b/pkg/minikube/registry/drvs/docker/docker.go
@@ -94,7 +94,7 @@ func status() registry.State {
 	}
 	if err == nil {
 		glog.Infof("docker version: %s", output)
-		return registry.State{Installed: true, Healthy: true}
+		return checkOverlayMod()
 	}
 
 	glog.Warningf("docker returned error: %v", err)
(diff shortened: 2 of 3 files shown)


afbjorklund:
What are we trying to guard against here ? Users that are using `vfs` driver for "special" reasons (NTFS) ? #7933

Or to avoid destroying the docker installation of older docker users ? #7626 Or users that are using btrfs ? #6167

pkg/minikube/registry/drvs/docker/docker.go
```diff
@@ -114,7 +114,49 @@ func status() registry.State {
...
+		cmd = exec.CommandContext(ctx, "cat", path)
+		out, err = cmd.Output()
+		if err != nil {
+			glog.Warningf("overlay module was not found in %q", path)
+			return registry.State{NeedsImprovement: true, Installed: true, Healthy: true, Fix: "enable overlayfs kernel module on your Linux"}
+		}
+		if strings.Contains(string(out), "overlay") { // success
+			return registry.State{NeedsImprovement: false, Installed: true, Healthy: true}
+		}
+		glog.Warningf("overlay module was not found")
```
afbjorklund, line 153: This is only checking the builtin modules, not the actual kernel modules ?

e.g. /lib/modules/5.4.40/kernel/fs/overlayfs/overlay.ko

Normally we use `modprobe` when dealing with modules.

`modprobe -n overlay`

medyagh: > What are we trying to guard against here ? Users that are using `vfs` driver for "special" reasons (NTFS) ? #7933
> 
> Or to avoid destroying the docker installation of older docker users ? #7626 Or users that are using btrfs ? #6167

on our current roadmap and capacitites we dont have room for supporting non-overlay in preload
so technically I am warnning the users that till we solve this issue, they will have a bad time with minikibe if they dont have overlay enabled.

I have personally seen people who were using minikube with non-overlay and it was extremely slow and it kept crashing... sometimies it works but....very poor performance, so for that reason I decided to maybe add a message and let the users know...

I could reword the message and  till this issue is solved, they need to use it minikube with overlay otherwsie they will see performance problems
https://github.com/kubernetes/minikube/issues/7626



what do u think @afbjorklund

medyagh: > iltin modules, not the actual kernel modules ?
> 
> e.g. /lib/modules/5.4.40/kernel/fs/overlayfs/overlay.ko

this is used as the second chance trying, if the first method didn't work... I remember on chromebook or maybe another platofrom the first way didnt' work but the second way worked.

what do u think ? @afbjorklund

afbjorklund: Ah, didn't read enough of the context - it is fine as a fallback solution, doing it "manually"

cmd/minikube/cmd/start.go
```diff
@@ -616,6 +616,13 @@ func validateDriver(ds registry.DriverState, existing *config.ClusterConfig) {
 	st := ds.State
 	glog.Infof("status for %s: %+v", name, st)
 
+	if st.NeedsImprovement { // warn but don't exit
+		out.ErrLn("")
+		out.WarningT("'{{.driver}}' driver reported a issue that could affect the performance.", out.V{"driver": name})
```
sharifelgamal, line 621: {{.driver}} driver reported an issue that could affect performance

Issue #8354, opened 2020-06-01
evaluate if can supply CNI by default for multinode and others


medyagh:
and do performance anaylssis

medyagh:
dupe of https://github.com/kubernetes/minikube/issues/1568

PR #8545, opened 2020-06-24
Add --cni flag (replaces --enable-default-cni), fix --network-plugin handling

### LINKED ISSUE #8055: CoreDNS fails on minions on multi-node clusters. Can't resolve external DNS from non-master pods.
So, I already fixed this and lost some of the logs. But it's pretty straight-forward.
1. Make a cluster
```bash
minikube start --vm-driver=kvm2 --cpus=2 --nodes 3 --network-plugin=cni \
--addons registry --enable-default-cni=false \
--insecure-registry "10.0.0.0/24" --insecure-registry "192.168.39.0/24" \
--extra-config=kubeadm.pod-network-cidr=10.244.0.0/16 \
--extra-config=kubelet.network-plugin=cni
kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml
```
n.b. I built from head a couple days ago
```bash
minikube version: v1.10.0-beta.2
commit: 80c3324b6f526911d46033721df844174fe7f597
```

2. make a pod on master and a pod on a node
3. from node pod: `curl google.com`
4. from master pod: `curl google.com`

CoreDNS was crashing per https://github.com/kubernetes/kubernetes/issues/75414

Fixed with
```bash
kubectl patch deployment coredns -n kube-system --patch '{"spec":{"template":{"spec":{"volumes":[{"name":"emptydir-tmp","emptyDir":{}}],"containers":[{"name":"coredns","volumeMounts":[{"name":"emptydir-tmp","mountPath":"/tmp"}]}]}}}}' 
```

Edit: had wrong flannel yaml listed.

### LINKED ISSUE #8222: multi-node: unable to choose CNI (always applies kindnet)

tstromberg:
Allows users to specify different CNI plug-ins to use. As a side-effect, this also fixes `--network-plugin=kubenet`

* Fixes #8222 
* Does not fix #8055 (kubelet configuration for CNI child nodes still needs work)

* Unblocks #1568 and #8354


cmd/minikube/cmd/start_flags.go
```diff
@@ -130,8 +132,9 @@ func initMinikubeFlags() {
 	startCmd.Flags().String(mountString, constants.DefaultMountDir+":/minikube-host", "The argument to pass the minikube mount command on start.")
 	startCmd.Flags().StringArrayVar(&config.AddonList, "addons", nil, "Enable addons. see `minikube addons list` for a list of valid addon names.")
 	startCmd.Flags().String(criSocket, "", "The cri socket path to be used.")
-	startCmd.Flags().String(networkPlugin, "", "The name of the network plugin.")
-	startCmd.Flags().Bool(enableDefaultCNI, false, "Enable the default CNI plugin (/etc/cni/net.d/k8s.conf). Used in conjunction with \"--network-plugin=cni\".")
+	startCmd.Flags().String(networkPlugin, "", "Kubelet network plug-in to use (default: auto)")
+	startCmd.Flags().Bool(enableDefaultCNI, false, "DEPRECATED: Replaced by --cni=custom")
```
tstromberg, line 136: This should be --cni=bridge

cmd/minikube/cmd/start_flags.go
```diff
@@ -130,8 +132,9 @@ func initMinikubeFlags() {
...
+	startCmd.Flags().String(cniFlag, "", "CNI plug-in to use. Valid options: auto, calico, custom, flannel, kindnet (default: auto)")
```
tstromberg, line 137: Remove calico.

tstromberg: custom is now named bridge.

pkg/minikube/bootstrapper/kubeadm/kubeadm.go.rej
```diff
@@ -0,0 +1,16 @@
+***************
```
tstromberg, line 1: Remove .rej files

pkg/minikube/cni/cni.go
```diff
@@ -0,0 +1,160 @@
...
+
+// Runner is the subset of command.Runner this package consumes
+type Runner interface {
+	RunCmd(cmd *exec.Cmd) (*command.RunResult, error)
+	Copy(assets.CopyableFile) error
+}
+
+// Manager is a common interface for CNI
+type Manager interface {
+	// Enable enables the CNI
```
tstromberg, line 49: Add comment on why there are two sets of runners

pkg/minikube/cni/cni.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// manifestAsset returns a copyable asset for the CNI manifest
+func manifestAsset(b []byte) assets.CopyableFile {
+	return assets.NewMemoryAssetTarget(b, manifestPath(), "0644")
+}
+
+// kubectlPath returns the path to the kubelet
+func kubectlPath(cc config.ClusterConfig) string {
+	return path.Join(vmpath.GuestPersistentDir, "binaries", cc.KubernetesConfig.KubernetesVersion, "kubectl")
```
tstromberg, line 128: Change to use new kapi.KubectlPath function

tstromberg:
> looks good to me, one nit: I suggest removing one of the CNI plugins from this PR and add it as a follow up PR.
> so we can make an example PR of it. how to add a CNI plugin to minikube, people would just follow that PR example

I'll follow this PR up with one which adds Calico support.

medyagh:
same line needs to be fixed in master github action too
https://github.com/medyagh/minikube/blob/b8229c3014e1a9a535f7ff8342d6b9b7317d7b76/.github/workflows/master.yml#L544

pkg/minikube/cni/bridge.go
```diff
@@ -0,0 +1,91 @@
...
+  "ipMasq": true,
+  "hairpinMode": true,
+  "ipam": {
+      "type": "host-local",
+      "subnet": "{{.PodCIDR}}"
+  }
+}
+`))
+
+// Bridge is a CNI manager than does nothing
```
priyawadhwa, line 50: nit : "than does nothing"

tstromberg: fixed. Thanks!

tstromberg:
I'm going to merge this and follow-it up with an improvement for CRIO.

Issue #5682, opened 2019-10-21
Reduce VM CPU overhead by 20%


tstromberg:
Related to #3207 and #5398 

#3207 can probably be closed - I don't believe we are at 50% usage any longer.

priyawadhwa:
Running `pidstat 1`, I noticed that every 5 seconds the following command results in a high uptick in CPU usage (around 80%, from ~20%, for one core) on my linux machine:

```
/usr/local/bin/kubectl apply -f /etc/kubernetes/addons -l kubernetes.io/cluster-service!=true,addonmanager.kubernetes.io/mode=Reconcile --prune=true --prune-whitelist core/v1/ConfigMap --prune-whitelist core/v1/Endpoints --prune-whitelist core/v1/Namespace --prune-whitelist core/v1/PersistentVolumeClaim --prune-whitelist core/v1/PersistentVolume --prune-whitelist core/v1/Pod --prune-whitelist core/v1/ReplicationController --prune-whitelist core/v1/Secret --prune-whitelist core/v1/Service --prune-whitelist batch/v1/Job --prune-whitelist batch/v1beta1/CronJob --prune-whitelist apps/v1/DaemonSet --prune-whitelist apps/v1/Deployment --prune-whitelist apps/v1/ReplicaSet --prune-whitelist apps/v1/StatefulSet --prune-whitelist extensions/v1beta1/Ingress --recursive 
```

the addon manager currently polls every [5 seconds](https://github.com/kubernetes/minikube/blob/master/deploy/addons/addon-manager.yaml.tmpl#L32-L33).

@tstromberg, WDYT of increasing polling time to 20 or 30 seconds? I have a graph showing CPU usage for the qemu process at [5 seconds](https://github.com/priyawadhwa/minikube/blob/addon-manager-cpu-uptick/graphs/five.png) and [20 seconds](https://github.com/priyawadhwa/minikube/blob/addon-manager-cpu-uptick/graphs/twenty.png)
 for addon manager poll time.

sharifelgamal:
This is being targeted for early next year.

PR #8552, opened 2020-06-25
Reduce coredns replicas from 2 to 1

priyawadhwa:
This is an easy way to improve overhead -- we don't need 2 coredns pods running by default. If a user wants more, they can easily scale the deployment up with `kubectl`

it won't make a huge difference alone, but in combination with some other PRs I'll be opening it should get us to our goal of 20% reduction

#5682

### DIFF
--- a/pkg/addons/kubectl.go
+++ b/pkg/addons/kubectl.go
@@ -32,7 +32,7 @@ func kubectlCommand(cc *config.ClusterConfig, files []string, enable bool) *exec
 		v = cc.KubernetesConfig.KubernetesVersion
 	}
 
-	kubectlBinary := kubectlBinaryPath(v)
+	kubectlBinary := KubectlBinaryPath(v)
 
 	kubectlAction := "apply"
 	if !enable {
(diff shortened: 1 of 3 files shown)


pkg/addons/kubectl.go
```diff
@@ -47,6 +47,7 @@ func kubectlCommand(cc *config.ClusterConfig, files []string, enable bool) *exec
 	return exec.Command("sudo", args...)
 }
 
-func kubectlBinaryPath(version string) string {
+// KubectlBinaryPath returns the path to kubectl on the node
+func KubectlBinaryPath(version string) string {
```
medyagh, line 51: nit: consider moving this to another package for future proofing cyclic dependcy. maybe kapi package?

priyawadhwa: good idea, done!

pkg/minikube/node/start.go
```diff
@@ -507,3 +514,15 @@ func prepareNone() {
...
 	}
 }
+
+// rescaleCoreDNS attempts to reduce coredns replicas from 2 to 1 to improve CPU overhead
+// no worries if this doesn't work
+func rescaleCoreDNS(cc *config.ClusterConfig, runner command.Runner) {
+	kubectl := kapi.KubectlBinaryPath(cc.KubernetesConfig.KubernetesVersion)
+	cmd := exec.Command("sudo", "KUBECONFIG=/var/lib/minikube/kubeconfig", kubectl, "scale", "deployment", "--replicas=1", "coredns", "-n=kube-system")
+	if _, err := runner.RunCmd(cmd); err != nil {
+		glog.Infof("unable to scale coredns replicas to 1: %v", err)
```
tstromberg, line 524: This should be warningf

Issue #6954, opened 2020-03-08
none driver: support non systemd ways to restart docker


medyagh:
currently if u run none driver on a system that does have docker but it is not systemd
it wont work !

because we expect docker restart to use systemctl (and init.d scripts wont work)

```
// Restart restarts Docker on a host
func (r *Docker) Restart() error {
	c := exec.Command("sudo", "systemctl", "restart", "docker")
	if _, err := r.Runner.RunCmd(c); err != nil {
		return errors.Wrap(err, "restarting docker.")
	}
	return nil
}
```

https://github.com/kubernetes/minikube/blob/master/pkg/minikube/cruntime/docker.go#L101

we could try do restart docker service using "service restart docker" if there is no systemctl service file for docker.

afbjorklund:
> we could try do restart docker service using "service restart docker" if there is no systemctl service file for docker.

sound good, if not running systemd then service should be available (or easy enough to add).

This issue applies to all the runtimes, not only docker. It is also used for the kubelet, for example.

So the functionality (start/stop/restart/is-active) should probably move into some library functions ?

afbjorklund:
Apparently last one needs LSB `status_of_proc` https://wiki.debian.org/LSBInitScripts/StatusSupport

afbjorklund:
@medyagh : do we have any supported environments remaining, that **don't** use systemd ?

it used to be mostly for the older versions of ubuntu, used when travis used containers not vms

sayboras:
I tried to run integration test with ubuntu docker (as I am developing in Mac). Then came across this issue as https://hub.docker.com/_/ubuntu 18.04 is running with `service` only.

Understand that running docker in docker is not ideal, but having VM is just too much for my 13' mac.

afbjorklund:
@sayboras : we have the "docker" driver for that, running the VM tests in docker won't be enough

afbjorklund:
> currently if u run none driver on a system that does have docker but it is not systemd

Do you have any examples of such an environment ? Both ubuntu and centos now have it

Just wondering it about the "soon" priority, as we already had a backlog item #2704 for this...

afbjorklund:
Being implemented in #7081

paddy-hack:
> Do you have any examples of such an environment ? Both ubuntu and centos now have [systemd]

[...]

Issue #2704, opened 2018-04-11
none: Add support for non-systemd Linux distributions


andyxning:
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
BUG REPORT
  - With latest v0.26.0, `minikube start --vm-driver=none --kubernetes-version=v1.8.0 --logtostderr` failed to start on travis ci. For more info: https://travis-ci.org/kubernetes/kube-state-metrics/jobs/365005727

asymmetric:
The [same thing](https://circleci.com/gh/kinvolk/habitat-operator/48) is happeninng on CircleCI.

It seems to be due to minikube expecting to find `systemctl`, which is not there on the CircleCi Ubuntu Trusty box at least (and I suspect on the Travis boxes either).

dlorenc:
I think you'll need to pass --bootstrapper=localkube to work with the none driver in environments that don't have system for now.

andyxning:
@dlorenc Is this the recommended way to work around systemd or this is a regression bug?

dlorenc:
It's a bit of a regression caused by the switch from localkube to kubeadm by default. Kubeadm with the none driver requires systemd.

The other problem here is that travis only supports up to Ubuntu 14.04, and systemd was added in 16.04.

andyxning:
@dlorenc seems we force use `--bootstrapper=localkube` is fine and it is ok even travis later bump Ubuntu from 14.04 to 16.04 if we have no strong dependency on kubeadm features.

fabiand:
We are also experiencing this.

It sounds like travis will need some time to provide an image with systemd.
Thus I wonder if minikube could at least do a check if systemd is present, to output a warning if the kubeadm bootstrapper is used.

This would at least highlight the problem.

I am not sure what the right fix is tho (wait for travis update or fix it in minikube).

andyxning:
@dlorenc @r2d4 @aaron-prindle

andyxning:
Ping @dlorenc @r2d4 @aaron-prindle Any suggestions? Pinning on an old version of minikube is not so good from a long run perspective.

andyxning:
Friendly ping.~

gbraad:
You do know they are in a different timezone?

andyxning:
Ping @dlorenc @r2d4 @aaron-prindle

dlorenc:
[...]

PR #8572, opened 2020-06-26
WIP: Add support for older sysv init as well

afbjorklund:
Not everything is Systemd or OpenRC yet

Previous code was assuming that everything non-systemd is openrc, which failed on some systems.

So look for OpenRC more explicitly, and add some generic fallback for the older SysV-like systems.

* https://en.wikipedia.org/wiki/OpenRC

* https://en.wikipedia.org/wiki/Init#SYSV

All supported minikube distributions _are_ using Systemd, so this is mostly for the "none" driver...

Should test this on some alternative distributions, such as older Ubuntu/CentOS or maybe Devuan.

* Ubuntu 14.04

* CentOS 6

For #2704 and #6954

### DIFF
--- a/pkg/minikube/sysinit/sysinit.go
+++ b/pkg/minikube/sysinit/sysinit.go
@@ -24,6 +24,7 @@ import (
 )
 
 var cachedSystemdCheck *bool
+var cachedOpenRCCheck *bool
 
 // Runner is the subset of command.Runner this package consumes
 type Runner interface {
@@ -68,6 +69,7 @@ func New(r Runner) Manager {
 	}
 
 	var systemd bool
+	var openrc bool
 
 	// Caching the result is important, as this manager may be created in many places,
 	// and ssh calls are expensive on some drivers, such as Docker.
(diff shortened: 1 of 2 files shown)


afbjorklund:
> since we dont have integration test for this, is it possible to test this PR in systemd and init.d distro ot make sure we dont break the current cases? and put the output in the description ?

the idea was that there should be **no** changes for users of systemd and openrc, i.e. all the currently covered sysinit

what init system does cloud shell use ? I suppose we could do some even more basic fallback, if both redhat and debian fail

afbjorklund:
Also note that they are pretty identical (opensrc and sysvinit), we mostly need to replace `start-stop-daemon`

We're assuming here that the standard `service` wrapper is available, so we can do basic start|stop|restart|status

```
# /usr/bin/service
#
# A convenient wrapper for the /etc/init.d init scripts.
#
# This script is a modified version of the /sbin/service utility found on
# Red Hat/Fedora systems (licensed GPLv2+).
```

Not to be confused with the `rc-service` program in openrc...

As used in for instance Alpine, and some other distributions:

```console
/ # file /sbin/service 
/sbin/service: symbolic link to /sbin/rc-service
/ # file /sbin/rc-service 
/sbin/rc-service: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-x86_64.so.1, stripped
```

Issue #8556, opened 2020-06-25
Kernel with CONFIG_IKHEADERS for BPF tools on Kubernetes


alban:
I would like the Minikube kernel to be compiled with [CONFIG_IKHEADERS](https://github.com/torvalds/linux/blob/435faf5c218a47fd6258187f62d9bb1009717896/init/Kconfig#L610-L617).

This is missing from [linux_defconfig](https://github.com/kubernetes/minikube/blob/master/deploy/iso/minikube-iso/board/coreos/minikube/linux_defconfig).

[BCC tools](https://github.com/iovisor/bcc) need to have access to kernel headers. It can be done either by installing `linux-headers` packages or by having a kernel compiled with `CONFIG_IKHEADERS` so that enough information can be retrieved via `/sys/kernel/kheaders.tar.xz`.

BCC tools are used in [Inspektor Gadget](https://github.com/kinvolk/inspektor-gadget/), a collection of tools for developers of Kubernetes applications. I would like Inspektor Gadget to support Minikube.

/cc @mauriciovasquezbernal

tstromberg:
@alban - building the headers in is likely the easier of the two options. I'd be happy to approve a PR that does this: https://minikube.sigs.k8s.io/docs/contrib/building/iso/#adding-kernel-modules

priyawadhwa:
@alban we have a doc on running BCC tools as a container in minikube, maybe this would work for you until the headers get added to the kernel

https://minikube.sigs.k8s.io/docs/tutorials/ebpf_tools_in_minikube/

alban:
This issue should be reopened:

Minikube `v1.12.0-beta.1` had Linux `5.4.40` (with `CONFIG_IKHEADERS`) but Minikube `v1.12.0` has Linux `4.19.107` (without `CONFIG_IKHEADERS` since this build option only appeared in Linux `5.2`).

See the revert commit: https://github.com/kubernetes/minikube/pull/8649

alban:
Additionally, I would like to have CONFIG_DEBUG_INFO_BTF (available from Linux 5.2) to benefit from BPF programs compiled with [CO-RE](https://facebookmicrosites.github.io/bpf/blog/2020/02/19/bpf-portability-and-co-re.html).

sharifelgamal:
So this won't be able to be supported until we update the kernel version of the ISO to v5, which we currently cannot for several reasons.

pnasrat:
[...]

PR #8582, opened 2020-06-27
Kernel with CONFIG_IKHEADERS for BPF tools on Kubernetes

alban:
BCC tools need to have access to kernel headers. It can be done either
by installing linux-headers packages or by having a kernel compiled with
CONFIG_IKHEADERS so that enough information can be retrieved via
/sys/kernel/kheaders.tar.xz.

BCC tools are used in Inspektor Gadget, a collection of tools for
developers of Kubernetes applications. I would like Inspektor Gadget to
support Minikube.

Fixes https://github.com/kubernetes/minikube/issues/8556


alban:
It seems to work:
```
$ ./out/minikube start --iso-url=file://$(pwd)/out/minikube.iso
$ ./out/minikube ssh
                         _             _            
            _         _ ( )           ( )           

$ uname -a
Linux minikube 5.4.40 #1 SMP Sat Jun 27 15:36:59 UTC 2020 x86_64 GNU/Linux
$ ls -l /sys/kernel/kheaders.tar.xz
-r--r--r-- 1 root root 3438056 Jun 27 16:33 /sys/kernel/kheaders.tar.xz
$ 

```

afbjorklund:
@alban : we don't have any packages on the minikube iso, so this could be a good alternative

3 MB doesn't seem to much to pay (size addition to the iso), for having better BPF compatibility

alban:
Thanks for the quick review @afbjorklund! I understand we need a `/lgtm` for a lgtm label (in addition to the github review). Could you lgtm or approve the PR?

afbjorklund:
Actually we need a merge and then a new ISO, and we are currently having some issues with it...

afbjorklund:
But I think the configuration change looks good, maybe we need to update the docs as well ?

PR #8586, opened 2020-06-28
Support more error messages for container inspect

afbjorklund:
The messages are different between "inspect" and "container inspect",
and also changed a bit in format between Podman v1 and Podman v2.

Support all of them, for compatibility with different versions of
Docker and Podman. Was causing restarts and failures, with KIC.

### DIFF
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -294,7 +294,12 @@ func StartContainer(ociBin string, container string) error {
 func ContainerID(ociBin string, nameOrID string) (string, error) {
 	rr, err := runCmd(exec.Command(ociBin, "container", "inspect", "-f", "{{.Id}}", nameOrID))
 	if err != nil { // don't return error if not found, only return empty string
-		if strings.Contains(rr.Stdout.String(), "Error: No such object:") || strings.Contains(rr.Stdout.String(), "unable to find") {
+		if strings.Contains(rr.Stdout.String(), "Error: No such object:") ||
+			strings.Contains(rr.Stdout.String(), "Error: No such container:") ||
+			strings.Contains(rr.Stdout.String(), "unable to find") ||
+			strings.Contains(rr.Stdout.String(), "Error: error inspecting object") ||
+			strings.Contains(rr.Stdout.String(), "Error: error looking up container") ||
+			strings.Contains(rr.Stdout.String(), "no such container") {
 			err = nil
 		}
 		return "", err


afbjorklund:
Has been a problem since 14937e46773d78d65d6c250dc7a286008a6a9f1f

afbjorklund:
There might still be some mismatch between stdout/stderr here, don't think we use CombinedOutput ?

pkg/drivers/kic/oci/oci.go
```diff
@@ -294,7 +294,12 @@ func StartContainer(ociBin string, container string) error {
 func ContainerID(ociBin string, nameOrID string) (string, error) {
 	rr, err := runCmd(exec.Command(ociBin, "container", "inspect", "-f", "{{.Id}}", nameOrID))
 	if err != nil { // don't return error if not found, only return empty string
-		if strings.Contains(rr.Stdout.String(), "Error: No such object:") || strings.Contains(rr.Stdout.String(), "unable to find") {
+		if strings.Contains(rr.Stdout.String(), "Error: No such object:") ||
+			strings.Contains(rr.Stdout.String(), "Error: No such container:") ||
+			strings.Contains(rr.Stdout.String(), "unable to find") ||
+			strings.Contains(rr.Stdout.String(), "Error: error inspecting object") ||
+			strings.Contains(rr.Stdout.String(), "Error: error looking up container") ||
```
medyagh, line 301: Can we make sure this one won't be a legit error ? Could this error possibly have am error we care about ? Do you mind commenting what cli tool has that error format ? Podman or docker ? Maybe paste the full output as a comment?

pkg/drivers/kic/oci/oci.go
```diff
@@ -294,7 +294,12 @@ func StartContainer(ociBin string, container string) error {
 func ContainerID(ociBin string, nameOrID string) (string, error) {
 	rr, err := runCmd(exec.Command(ociBin, "container", "inspect", "-f", "{{.Id}}", nameOrID))
 	if err != nil { // don't return error if not found, only return empty string
-		if strings.Contains(rr.Stdout.String(), "Error: No such object:") || strings.Contains(rr.Stdout.String(), "unable to find") {
+		if strings.Contains(rr.Stdout.String(), "Error: No such object:") ||
+			strings.Contains(rr.Stdout.String(), "Error: No such container:") ||
+			strings.Contains(rr.Stdout.String(), "unable to find") ||
+			strings.Contains(rr.Stdout.String(), "Error: error inspecting object") ||
```
medyagh, line 300: could we also group them by ociBin if something is only used for podman, lets not check it against docker...

PR #8589, opened 2020-06-28
add new method to clean

alonyb:
This PR is for #7670

An improvement of #7787

According to what was discussed in the triage, and suggestion of @tstromberg. Cleanup was implemented, which was integrated into Go 1.14 [details](https://github.com/golang/go/commit/1b2ff1013678bca2181d046a91857e78f3981f9c)

Clean up after creating temporary directories in unit tests. With validation for tests that call a single directory multiple times.

### DIFF
--- a/cmd/minikube/cmd/config/set_test.go
+++ b/cmd/minikube/cmd/config/set_test.go
@@ -81,4 +81,8 @@ func createTestConfig(t *testing.T) {
 	if err = os.MkdirAll(localpath.MakeMiniPath("profiles"), 0777); err != nil {
 		t.Fatalf("error creating temporary profiles directory: %+v", err)
 	}
+
+	t.Cleanup(func() {
+		os.RemoveAll(td)
+	})
 }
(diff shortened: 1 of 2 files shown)


afbjorklund:
> Cleanup was implemented, which was integrated into Go 1.14

We are still using 1.13.8 though, same as kubernetes upstream.

It will be 1.13.9 soon, but not sure when they/we will go to 1.14...

afbjorklund:
Need to make GitHub Actions use the same (major) version as everyone else...

https://github.com/actions/virtual-environments/blob/ubuntu18/20200621.1/images/linux/Ubuntu1804-README.md

> go:
>
>     go 1.11.13
>     go 1.12.17
>     go 1.13.12
>     go 1.14.4

https://github.com/actions/setup-go

alonyb:
@afbjorklund This will be on hold until #8590 is resolved. Thanks for the update.

@medyagh with this, `defer` could be removed from the rest of unittests, and use `Cleanup`

PR #8433, opened 2020-06-09
wip: Fixes duplicate messages to upgrade Kubernetes

sakshamkhanna:
Introduce a parameter displayKubernetesUpgradeMessage in func getKubernetesVersion that disables multiple prints of the upgrade message

Fixes #8418 

Before:
```
😄  minikube v1.11.0 on Darwin 10.14.6
✨  Using the hyperkit driver based on existing profile
🆕  Kubernetes 1.18.3 is now available. If you would like to upgrade, specify: --kubernetes-version=v1.18.3
🆕  Kubernetes 1.18.3 is now available. If you would like to upgrade, specify: --kubernetes-version=v1.18.3
🆕  Kubernetes 1.18.3 is now available. If you would like to upgrade, specify: --kubernetes-version=v1.18.3
👍  Starting control plane node minikube in cluster minikube
🏃  Updating the running hyperkit "minikube" VM ...
❗  This VM is having trouble accessing https://k8s.gcr.io
💡  To pull new external images, you may need to configure a proxy: https://minikube.sigs.k8s.io/docs/reference/networking/proxy/
🐳  Preparing Kubernetes v1.18.3 on Docker 19.03.8 ...
🔎  Verifying Kubernetes components...
🌟  Enabled addons: default-storageclass, storage-provisioner
🏄  Done! kubectl is now configured to use "minikube"

❗  /usr/local/bin/kubectl is version 1.15.5, which may be incompatible with Kubernetes 1.18.3.
💡  You can also use 'minikube kubectl -- get pods' to invoke a matching version
```

After:
```
😄  minikube v1.11.0 on Darwin 10.14.6
✨  Using the hyperkit driver based on existing profile
🆕  Kubernetes 1.18.3 is now available. If you would like to upgrade, specify: --kubernetes-version=v1.18.3
👍  Starting control plane node minikube in cluster minikube
🏃  Updating the running hyperkit "minikube" VM ...
❗  This VM is having trouble accessing https://k8s.gcr.io
💡  To pull new external images, you may need to configure a proxy: https://minikube.sigs.k8s.io/docs/reference/networking/proxy/
🐳  Preparing Kubernetes v1.18.3 on Docker 19.03.8 ...
🔎  Verifying Kubernetes components...
🌟  Enabled addons: default-storageclass, storage-provisioner
🏄  Done! kubectl is now configured to use "minikube"

❗  /usr/local/bin/kubectl is version 1.15.5, which may be incompatible with Kubernetes 1.18.3.
[...]

PR #8591, opened 2020-06-28
Prevent start cmd from logging multiple k8s upgrade messages

### LINKED ISSUE #8418: UI :  upgrade k8s version message gets repeated 4 times
as seen here in this issue https://github.com/kubernetes/minikube/issues/8386#issue-631602296

```
Starting minikube
😄  minikube v1.11.0 on Ubuntu 18.04
🐳  Preparing Kubernetes v1.17.5 on Docker 19.03.2 ...

💣  failed to start node: startup failed: Wait failed: waiting for apps_running: checking k8s-apps to be running: timed out waiting for the condition

```

sunny-b:
This fixes #8418

Currently, the `getKubernetesVersion` method is called upwards of 5 times when the `start` command is run. This leads to to the `Kubernetes 1.18.3 is now available. If you would like to upgrade, specify: --kubernetes-version=v1.18.3` message to get logged multiple times to the console.

One possible way to prevent this from happening is to remove the part of of `getKubernetesVersion` that logs the upgrade message and only have it log once in `runStart`. This would cause the fewest amount of code change.

However, another solution, which this PR uses, is to only call `getKubernetesVersion` once at the beginning of `runStart` and pass the resulting value to the functions that need it as an additional argument. This ensures that the message is only printed once and has the added benefit of slightly optimizing the speed of the `runStart` script.

Though I'm certainly open to suggestions for alternate/better implementations.

I realize there's another PR that is open for this issue (#8433), though that PR hasn't been updated in over 2 weeks. Feel free to close this PR if you don't want multiple PRs open for the same issue. 

Before: 
```bash
$ go run main.go start --kubernetes-version 1.17.4

😄  minikube v0.0.0-unset on Darwin 10.14.6
🔥  Creating docker container (CPUs=2, Memory=1991MB) ...
🐳  Preparing Kubernetes v1.17.4 on Docker 19.03.2 ...
```

After:
```bash
$ go run main.go start --kubernetes-version 1.17.4

😄  minikube v0.0.0-unset on Darwin 10.14.6
🔥  Creating docker container (CPUs=2, Memory=1991MB) ...
🐳  Preparing Kubernetes v1.17.4 on Docker 19.03.2 ...
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -153,9 +153,9 @@ func runStart(cmd *cobra.Command, args []string) {
 	}
 
 	validateSpecifiedDriver(existing)
+	validateKubernetesVersion(existing)
 	ds, alts, specified := selectDriver(existing)
-	k8sVersion := getKubernetesVersion(existing)
-	starter, err := provisionWithDriver(cmd, ds, existing, k8sVersion)
+	starter, err := provisionWithDriver(cmd, ds, existing)
 	if err != nil {
 		node.MaybeExitWithAdvice(err)
 		machine.MaybeDisplayAdvice(err, viper.GetString("driver"))
@@ -178,7 +178,7 @@ func runStart(cmd *cobra.Command, args []string) {
 				if err != nil {
 					out.WarningT("Failed to delete cluster {{.name}}, proceeding with retry anyway.", out.V{"name": ClusterFlagValue()})
 				}
-				starter, err = provisionWithDriver(cmd, ds, existing, k8sVersion)
+				starter, err = provisionWithDriver(cmd, ds, existing)
 				if err != nil {
 					continue
 				} else {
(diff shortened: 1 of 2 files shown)


sunny-b:
On second thought, using the first approach that I mention in my description (remove the logging side-effect of `getKuberneteVersion` and have it log at the beginning of `runStart`) would be not only simpler but also more extensible as it wouldn't require changing the function signatures (and violating Open-Close).

Though getting a second opinion would be appreciated.

sunny-b:
@medyagh Do you have an opinion on my current approach? It optimizes the `start` command a bit but also changes the function signature for a number of functions. Not sure if that is an issue

sunny-b:
Looking at the `getKuberneteVersion` method, it seems to be doing a lot of validation that could probably be moved to its own `validateKubernetesVersion`. That seems like a better place for it.

PR #8593, opened 2020-06-28
Gracefully exit if container runtime is misspelled

### LINKED ISSUE #7426: do not crash, when missplelled container runtime

sunny-b:
This fixes #7426 by creating a simple lookup for valid container runtimes `(docker crio containerd)` and adding it in the `validateFlags` check.

Before:
```bash
$ ./out/minikube-darwin-amd64 start --driver=docker --container-runtime=conatinerd                                                                                                                 
😄  minikube v1.12.0-beta.0 on Darwin 10.14.6

💣  error provisioning host: Failed to generate config: new runtime manager: unknown runtime type: "conatinerd"

```

After:
```bash
$ ./out/minikube-darwin-amd64 start --driver=docker --container-runtime=conatinerd                                                         
😄  minikube v1.12.0-beta.0 on Darwin 10.14.6
💡  Invalid Container Runtime: "conatinerd". Valid runtimes are: docker, crio, containerd
```

### DIFF
--- a/pkg/minikube/cruntime/cruntime.go
+++ b/pkg/minikube/cruntime/cruntime.go
@@ -49,7 +49,7 @@ func (cs ContainerState) String() string {
 
 // ValidRuntimes lists the supported container runtimes
 func ValidRuntimes() []string {
-	return []string{"docker", "crio", "containerd"}
+	return []string{"docker", "cri-o", "containerd"}
 }
 
 // CommandRunner is the subset of command.Runner this package consumes
--- a/pkg/minikube/cruntime/cruntime.go
+++ b/pkg/minikube/cruntime/cruntime.go
@@ -47,6 +47,11 @@ func (cs ContainerState) String() string {
 	return [...]string{"all", "running", "paused"}[cs]
 }
 
+// ValidRuntimes lists the supported container runtimes
+func ValidRuntimes() []string {
+	return []string{"docker", "crio", "containerd"}
+}
+
 // CommandRunner is the subset of command.Runner this package consumes
 type CommandRunner interface {
 	RunCmd(cmd *exec.Cmd) (*command.RunResult, error)
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -857,8 +857,12 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 	if cmd.Flags().Changed(containerRuntime) {
 		runtime := strings.ToLower(viper.GetString(containerRuntime))
 
+		validOptions := cruntime.ValidRuntimes()
+		// `crio` is accepted as an alternative spelling to `cri-o`
+		validOptions = append(validOptions, "crio")
+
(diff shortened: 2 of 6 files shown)


pkg/minikube/cruntime/cruntime.go
```diff
@@ -47,6 +47,11 @@ func (cs ContainerState) String() string {
 	return [...]string{"all", "running", "paused"}[cs]
 }
 
+// ValidRuntimes lists the supported container runtimes
+func ValidRuntimes() []string {
+	return []string{"docker", "crio", "containerd"}
```
afbjorklund, line 52: The default spelling is supposed to be "cri-o", like in https://cri-o.io/

But we accept **both** as parameters (and normalize "crio" to "cri-o")

sunny-b: Gotcha. Would you prefer that to be list in the docs as well? Currently, it has it spelt as `crio`:
`--container-runtime='docker': The container runtime to be used (docker, crio, containerd).`

Should that stay the same continue to silently accept `cri-o` or should `cri-o` be included in the docs?

medyagh: I think the doc should be one way but we should accept alternative spellings . Thank you for this contribution

sunny-b: I'm still slightly confused. Just to clarify, the docs should stay the same (spelling `crio` as `cri-o`) but I should ensure that both spellings are accepted. Do I have that correct?

medyagh: sorry for the lack of clarity, 
the docs should say cri-o
and our help text should also say cri-o
but we should also accept it if the user provides crio as an option.

sunny-b: Ah gotcha. Will update that!

afbjorklund: Upstream is confused as well (the project is called cri-o, the binary is called crio)

afbjorklund: I'm actually not sure if "Cry... Oh!" (CRI-O) is any better than "Ociieed!" (OCID)

afbjorklund: https://www.redhat.com/en/blog/introducing-cri-o-10

sunny-b:
Is it normal for the `TestPreload` integration test to fail? Is it flaky? I ran it locally and it passed fine. The logs in the failed test don't really show why it failed.

afbjorklund:
The Docker Desktop VM on the Mac looks sad: `failed to "StartContainer" for "etcd" with CreateContainerConfigError: "open /var/lib/kubelet/pods/6817a9c8a9aa865ba50269ccfa66b9f2/etc-hosts: read-only file system"`

sunny-b:
Okay I've updated it to use `crio` for backwards compatibility and convert `cri-o` to `crio` if passed in.

PR #8598, opened 2020-06-29
Don't verify nf_conntrack for br_netfilter

afbjorklund:
We need to make sure that the br_netfilter kernel module
has been loaded, when using non-Docker container runtimes.

But the code was verifying the nf_conntrack kernel module
instead, so it could still cause the "not loaded" error.

`[ERROR FileContent--proc-sys-net-bridge-bridge-nf-call-iptables]`


afbjorklund:
It was added by mistake in https://github.com/kubernetes/minikube/pull/6427 when trying to fix broken builtins (#5900) for a custom kernel with some files deleted.

As long as `/lib/modules/$(uname -r)/modules.builtin` is available, then "modprobe" will work fine on builtin modules.

medyagh:
None test time out is related to https://github.com/kubernetes/minikube/issues/8601

PR #8608, opened 2020-06-30
Docs: Clarify steps to get podman running as a driver

mbbroberg:
Hey all - I'm offering up some changes based on my experiences (read: struggle) to get podman up-and-running on Fedora 32. I also standardized some of the section names and tips from other drivers.

### DIFF
--- a/site/content/en/docs/drivers/includes/podman_usage.inc
+++ b/site/content/en/docs/drivers/includes/podman_usage.inc
@@ -1,20 +1,22 @@
-## experimental
+## Experimental
 
-This is an experimental driver. please use it only for experimental reasons.
-for a better kubernetes in container experience, use docker [driver]({{< ref "/docs/drivers/docker/" >}})
-
-## Install Podman
-
-- [Podman](https://podman.io/getting-started/installation.html)
+This is an experimental driver. Please use it only for experimental reasons until it has reached maturity. For a more reliable minikube experience, use a non-experimental driver, like [Docker]({{< ref "/docs/drivers/docker.md" >}}).
 
 ## Usage
 
-Start a cluster using the podman driver:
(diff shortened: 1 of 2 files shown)


mbbroberg:
I pushed a fresh commit after realizing there was repetitive information across the .md file and .inc files. Let me know if this is helpful :boat:

site/content/en/docs/drivers/includes/podman_usage.inc
```diff
@@ -1,20 +1,22 @@
...
+```shell
+minikube start --driver=podman --container-runtime=cri-o
+```
 
-Start a cluster using the podman driver:
+Alternatively, start minikube with the podman driver only: 
 
 ```shell
-minikube start --driver=podman
+minikube start --driver=podman --container-runtime=cri-o
```
clcollins, line 16: @mbbroberg is this a typo?  It looks the same as line 10, but this section suggests you're only changing the Podman driver.

afbjorklund: Normally people run Docker with the Docker driver and CRI-O with the Podman driver.

Probably because they are from the same vendor, either combination _should_ work.

afbjorklund:
Unfortunately the upstream instructions doesn't work at the moment, after Podman 2.0 happened

We will need to redirect people to https://github.com/containers/libpod/releases/tag/v1.9.3 for now

site/content/en/docs/drivers/includes/podman_usage.inc
mbbroberg: Good catch @clcollins - thanks for pointing out the repetitive call to cri-o.

mbbroberg:
@afbjorklund thanks for the note. In this case, I'm running 1.9.3 with minikube version v1.11.0.

PR #8621, opened 2020-07-01
Disable containerd from starting up at boot

afbjorklund:
Saves us having to restart docker later on (see f2ea21ab23aa49c5e924c5c340a1a6cf5108dd14)

Tested with docker and with containerd runtimes

With the currently generated `minikube.service`, the default is to run with `/run/docker/containerd/containerd.sock`

It is only when containerd is explicitly selected (and **not** docker), that we want to use `/run/containerd/containerd.sock`

Note that it will just disable the containerd.service

There will still be a containerd process running...

### DIFF
--- a/deploy/iso/minikube-iso/package/containerd-bin/50-minikube.preset
+++ b/deploy/iso/minikube-iso/package/containerd-bin/50-minikube.preset
@@ -1,0 +1,1 @@
+disable containerd.service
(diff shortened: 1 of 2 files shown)


afbjorklund:
Disabling services is supposed to be done by the "distro"...

https://www.freedesktop.org/software/systemd/man/systemd.preset.html
https://www.freedesktop.org/wiki/Software/systemd/Preset/

Also fixed a warning, from the previous service installation.

```
Configuration file /home/anders/minikube/out/buildroot/output/target/usr/lib/systemd/system/containerd.service is marked executable. Please remove executable permission bits. Proceeding anyway.
Created symlink /home/anders/minikube/out/buildroot/output/target/etc/systemd/system/multi-user.target.wants/containerd.service → /usr/lib/systemd/system/containerd.service.
```

afbjorklund:
@medyagh : this should avoid having to restart docker a second time

It will still be restarted explicitly after the preload, on purpose I believe ?

PR #8635, opened 2020-07-03
Stop creating .minikube directory when the base of MINIKUBE_HOME is minikube

itchyny:
## Problem
When the user configures as `export MINIKUBE_HOME=~/.local/share/minikube`, minikube creates `~/.local/share/minikube/.minikube` directory. This is not what the user would expect, minikube uses the directory they specify by `MINIKUBE_HOME`

## Solution
When the base of `MINIKUBE_HOME` is equal to `minikube`, use it for the root directory.

## Comparison to other tools
`GEM_HOME`, `CARGO_HOME`, `GRADLE_USER_HOME` are considered as the root directory for storing their data, defaults to `$HOME/.gem`, `$HOME/.cargo`, `$HOME/.gradle`. So I think `MINIKUBE_HOME` should be considered as the value which defaults to `~/.minikube`, not the env for configuring `$HOME`. But considering the compatibility, this pr suggests with minimal impact.

Ref: #1104

### DIFF
--- a/pkg/minikube/localpath/localpath.go
+++ b/pkg/minikube/localpath/localpath.go
@@ -42,7 +42,8 @@ func MiniPath() string {
 	if minikubeHomeEnv == "" {
 		return filepath.Join(homedir.HomeDir(), ".minikube")
 	}
-	if filepath.Base(minikubeHomeEnv) == ".minikube" {
+	switch filepath.Base(minikubeHomeEnv) {
+	case ".minikube", "minikube":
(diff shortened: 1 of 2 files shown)


priyawadhwa:
Hey @itchyny are you still working on this?

itchyny:
> Hey @itchyny are you still working on this?

No, I think it's done.

itchyny:
Okay, I understand. Thank you for your review.

PR #8649, opened 2020-07-05
WIP: Revert Buildroot version from 2020.02 to 2019.02

afbjorklund:
Since there were test failures when running the new ISO

Upgrade the minor Buildroot version instead, to 2019.02.11

There might be some configuration changes to revert as well

Trying to keep the new versions of the added packages though


alban:
@afbjorklund Do you have a link to the test failures?
I am interested in the new kernel for https://github.com/kubernetes/minikube/issues/8556#issuecomment-657205132 and I wonder which failures should I expect if I run Minikube without this revert.

afbjorklund:
Look for the PRs with “new ISO”

It worked OK for me, but there were some regression tests that timed out

PR #8653, opened 2020-07-06
Revert "The config flag for go packages were both changed"

afbjorklund:
Need to use the old config flag again for the old Buildroot

This reverts commit 661feaca3fb47ea9e257bac2428633cd3bf0e06d.


afbjorklund:
CRI-O 1.18 now requires Go 1.13 to build, unfortunately

Issue #8656, opened 2020-07-06
go mod: upgrade lib k8s.io/kubernetes v1.17.3 


medyagh:
in our go mod we are using k8s.io/kubernetes v1.17.3 https://github.com/kubernetes/minikube/blob/5e9bb011b7cea625de5b3493c70fe42968edac3d/go.mod#L86

I would accept a PR that bumps this to a more recent version

vinu2003:
Hey @medyagh, Isn't this task involves updating matching version of all three k8s.io/... components in go.mod file ?
```
k8s.io/api kubernetes-1.18.5
k8s.io/apimachinery kubernetes-1.18.5
k8s.io/client-go kubernetes-1.18.5
```

or only k8s.io/kubernetes v1.17.3  => k8s.io/kubernetes v1.18.5 ? Correct me if I am wrong please.

PR #4719, opened 2019-07-09
Bump k8s.io/kubernetes to 1.15.0

josedonizetti:
fix for https://github.com/kubernetes/minikube/issues/4656

still trying to figure out a couple of things here.

PR #8671, opened 2020-07-08
upgrade: bump kube library version to latest 1.18.5

vinu2003:
##what/why this PR##:
This PR addresses changes required in go.mod/go.sum files to update kube library version from v1.17.3 to v1.18.5.

##Which issue this PR fixes##:
fixes: https://github.com/kubernetes/minikube/issues/8656


vinu2003:
I am noticing from the log, the issue is with Netlify build.command
Screen Shot 2020-07-08 at 5 44 57 pm

The problem seem to be in podman.md file : 
@@line14: ({{< ref "/docs/drivers/docker.md >}}).

vinu2003:
Hey @afbjorklund Kindly let me know if i can go ahead and fix the Netlify build failure in this PR itself or as different one.

vinu2003:
I have created a PR to fix the build issues : https://github.com/kubernetes/minikube/pull/8678

pkg/minikube/bootstrapper/bsutil/featuregates_test.go
```diff
@@ -30,18 +30,18 @@ func TestParseFeatureArgs(t *testing.T) {
 		expectedComponentFeatureArgs string
 	}{
 		{
-			description:  "CoreDNS enabled",
-			featureGates: "CoreDNS=true",
```
medyagh, line 34: could you specify why this change is realted to this PR ? is this needed ?

vinu2003: v1.18.0 => has updated the k8s.io/kubernetes/cmd/kubeadm/app/features as below : 
var coreDNSMessage = "featureGates:CoreDNS has been removed in v1.13\n" +
	"\tUse kubeadm-config to select which DNS addon to install."

```
const (
	// IPv6DualStack is expected to be alpha in v1.16
	IPv6DualStack = "IPv6DualStack"
	// PublicKeysECDSA is expected to be alpha in v1.19
	PublicKeysECDSA = "PublicKeysECDSA"
)

// InitFeatureGates are the default feature gates for the init command
var InitFeatureGates = FeatureList{
	IPv6DualStack:   {FeatureSpec: featuregate.FeatureSpec{Default: false, PreRelease: featuregate.Alpha}},
	PublicKeysECDSA: {FeatureSpec: featuregate.FeatureSpec{Default: false, PreRelease: featuregate.Alpha}},
}
```

We initialize and validate in ```featuregates.go```. So this update.

medyagh: do you mind pasting what was the exact unit test error you got without this change?

vinu2003: sure.
```
--- FAIL: TestParseFeatureArgs (0.00s)
    --- FAIL: TestParseFeatureArgs/CoreDNS_enabled (0.00s)
        featuregates_test.go:59: Kubeadm Actual: map[], Expected: map[CoreDNS:true]
        featuregates_test.go:63: Component Actual: CoreDNS=true, Expected: 
    --- FAIL: TestParseFeatureArgs/CoreDNS_disabled (0.00s)
        featuregates_test.go:59: Kubeadm Actual: map[], Expected: map[CoreDNS:false]
        featuregates_test.go:63: Component Actual: CoreDNS=false, Expected: 
--- FAIL: TestSupport (0.00s)
    featuregates_test.go:90: expected supportedFG(CoreDNS) to be true ! 
```

vinu2003:
I do notice some specific failures in many pull requests. Trying to see if i can help to fix any of these.

pkg/minikube/bootstrapper/bsutil/featuregates_test.go
vinu2003: Hey @medyagh - would you like to point out anything specific regarding these changes ?

pkg/minikube/bootstrapper/bsutil/featuregates_test.go
```diff
@@ -30,18 +30,18 @@ func TestParseFeatureArgs(t *testing.T) {
...
+			description:  "IPv6DualStack enabled",
+			featureGates: "IPv6DualStack=true",
 			expectedKubeadmFeatureArgs: map[string]bool{
-				"CoreDNS": true,
+				"IPv6DualStack": true,
```
medyagh, line 36: I am not convinced these values are correct can you link in comment where the official site says the coredns is renamed to ipv6dualstack ?

vinu2003: Hey @medyagh I came across this issue while running the tests. And was referring to release notes v1.18
https://kubernetes.io/docs/setup/release/notes/

```kubeadm: remove the deprecated CoreDNS feature-gate. It was set to "true" since v1.11 when the feature went GA. In v1.13 it was marked as deprecated and hidden from the CLI. (#87400, @neolit123) [SIG Cluster Lifecycle]```

Checked this PR - https://github.com/kubernetes/kubernetes/pull/87400 - https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecation.

I will see if I am missing to understand anything else from my end as well.

vinu2003: @medyagh - CoreDNS is not renamed to IPv6DualStack.
In Kubeadm, CoreDNS feature gate is deprecated and removed. This particular test in minikube-bsutil is related to feature gates. So it makes sense to replace with IPv6DualStack.
https://kubernetes.io/docs/setup/release/notes/

medyagh: so that would make a problem if a user wants to deploy an older kubernetes version to minikube ?

vinu2003: Hi @medyagh Apologies for delayed responses -  i was away for a while.

It is indeed not backward compatible if,  the user starts cluster with feature gates option ```--feature-gates CoreDNS=true``` with this dependency upgrade as the kubeadm refers to v1.18.5

But I was able to start the cluster successfully with older version of kubernetes with this change.
./out/minikube start --kubernetes-version v1.17.0 ==> this is ok.

Haven't tried something like supporting multiple versions in go mod.Not sure if that is an optimal solution to this as I still exploring the architecture and learning.

Also, I  noticed that i should be bumping to 1.18.3  and not 1.18.5.
DefaultKubernetesVersion = "v1.18.3". - Thsi si seperate.

vinu2003: Hey @medyagh as far as i have analysed into this, adding k8s as module means adding the subcomponents under require directive and this specific feature gates is part of kubeadm and there is way to support different versions of it. 
i have also looking into this PR earlier when its bumped up to 1.15
https://github.com/kubernetes/minikube/pull/4719 - which eventually replaced the older feature gates and added CoreDNS part.
Provided this CoreDNS is depreciated, can we go for another Pr with v18.3 which will have the same impact of backward compatibility when the driver is started that eventually calls kubeadm init feature-gates=CoreDNS.
But the user can start the old driver without this option.

Not sure how to move forward with this.

PR #8676, opened 2020-07-08
Add Infof function to out package

priyawadhwa:
This is a refactor for the work around minikube start JSON output #8655 

One of the types of logs we will output in JSON is an `Info` type, which covers things like options, env variables, and other informational logs during `minikube start`

Separating these into its own function will make it easier to identify the informational type and print in JSON in future PRs

### DIFF
--- a/pkg/minikube/out/out.go
+++ b/pkg/minikube/out/out.go
@@ -66,7 +66,16 @@ type V map[string]interface{}
 
 // T writes a stylized and templated message to stdout
 func T(style StyleEnum, format string, a ...V) {
+	if style == Option {
+		Infof(format, a...)
+	}
 	outStyled := ApplyTemplateFormatting(style, useColor, format, a...)
+	String(outStyled)
+}
+
+// Infof is used for informational logs (options, env variables, etc)
+func Infof(format string, a ...V) {
+	outStyled := ApplyTemplateFormatting(Option, useColor, format, a...)
 	String(outStyled)
 }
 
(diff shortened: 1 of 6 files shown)


pkg/minikube/out/out_test.go
```diff
@@ -114,3 +114,7 @@ func TestErr(t *testing.T) {
 		t.Errorf("Err() = %q, want %q", got, want)
 	}
 }
+
+func TestInfof(t *testing.T) {
```
medyagh, line 118: empty test

priyawadhwa: thanks, removed. this is tested in `TestOutT` test

pkg/minikube/out/out.go
```diff
@@ -66,10 +66,20 @@ type V map[string]interface{}
...
+	if style == Option {
+		Infof(format, a...)
+		return
+	}
 	outStyled := ApplyTemplateFormatting(style, useColor, format, a...)
 	String(outStyled)
 }
 
+// Infof is used for informational logs (options, env variables, etc)
+func Infof(format string, a ...V) {
```
medyagh, line 78: I find Info confusing, that when I am supposed to use Info vs T.
Either we should update out.T to be something else or how about naming it `out.TabbedInfo`
or `out.DisplayOption` or `out.SubInfo`

priyawadhwa: Since the JSON struct is of type `Info` I think it may be more confusing if the function name and the struct name don't match.

Technically it's ok if someone uses T instead of Info, since T will just defer to Info in the correct case (when style=Option). This is more so just so that the code is cleaner for JSON output

medyagh: ok thats convincing enough.

PR #8679, opened 2020-07-08
logDir.Changed always true

usernameisnull2:

### DIFF
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -62,7 +62,7 @@ var RootCmd = &cobra.Command{
 		}
 
 		logDir := pflag.Lookup("log_dir")
-		if !logDir.Changed {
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/root.go
```diff
@@ -62,7 +62,7 @@ var RootCmd = &cobra.Command{
 		}
 
 		logDir := pflag.Lookup("log_dir")
-		if !logDir.Changed {
+		if logDir.Value.String() == "" {
```
priyawadhwa, line 65: It seems in other parts of the code we check this via:

```go
if cmd.Flags().Changed(containerRuntime) {}
```

could we do the same here to follow convention?

usernameisnull0: follow the convention,the same problem will still occur, unless you modify cmd.Flags().Changed(), but this impact may be significant

PR #8684, opened 2020-07-08
none CNI: error if portmap plug-in is required but unavailable

tstromberg:
`flannel` and `kindnet` both require that the `portmap` CNI plug-in to be installed.

Our VM and Docker image includes it, but it's presence cannot be assumed for the `none` driver, resulting in a very confusing error message. This makes PR makes the omission very apparent. 

It also fixes an integration test assumption that did not work for `none`, but does not endeavor to enable CNI tests for `none` yet.

### DIFF
--- a/pkg/minikube/cni/kindnet.go
+++ b/pkg/minikube/cni/kindnet.go
@@ -18,6 +18,7 @@ package cni
 
 import (
 	"bytes"
+	"os/exec"
 	"text/template"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


test/integration/net_test.go
```diff
@@ -90,7 +90,16 @@ func TestNetworkPlugins(t *testing.T) {
...
+
+						// none does not support 'minikube ssh'
+						if NoneDriver() {
+							rr, err = Run(t, exec.CommandContext(ctx, "pgrep", "-a", "kubelet"))
+						} else {
+							rr, err = Run(t, exec.CommandContext(ctx, Target(), "ssh", "-p", profile, "pgrep -a kubelet"))
+						}
+
 						if err != nil {
 							t.Fatalf("ssh failed: %v", err)
```
medyagh, line 104: update this message. since it is not SSH in none

pkg/minikube/cni/kindnet.go
```diff
@@ -168,6 +169,12 @@ func (c KindNet) manifest() (assets.CopyableFile, error) {
 
 // Apply enables the CNI
 func (c KindNet) Apply(r Runner) error {
+	// This is mostly applicable to the 'none' driver
+	_, err := r.RunCmd(exec.Command("stat", "/opt/cni/bin/portmap"))
```
medyagh, line 173: nit: could we possiblty detect the Runner (if it is not a SSHRunner or kicRunner) then don't do this ? to save micro seconds ?

PR #8687, opened 2020-07-09
restrict apparmor security opt to docker

sharifelgamal:


tstromberg:
`Experimental_Podman_Linux — Jenkins: completed with 96 / 136 failures in 9.01 minute(s).` seems high, but I don't know what the current state is. Spirit of the PR seems right though?

PR #8697, opened 2020-07-10
improve applying default service account

medyagh:
after this PR:
return error instead of ignoring it, return wrapped error if it times out


pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -858,7 +858,7 @@ func (k *Bootstrapper) elevateKubeSystemPrivileges(cfg config.ClusterConfig) err
 	}()
 
 	// Allow no more than 5 seconds for creating cluster role bindings
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), applyTimeoutSeconds*time.Second)
```
sharifelgamal, line 861: is applyTimeoutSeconds a constant somewhere? you should also update the comment accordingly.

PR #8715, opened 2020-07-13
improve advice when driver was auto-selected

medyagh:
a user sent me this ...and I noticed the driver field is empty.
Screen Shot 2020-07-13 at 4 13 09 PM

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -161,7 +161,7 @@ func runStart(cmd *cobra.Command, args []string) {
 	starter, err := provisionWithDriver(cmd, ds, existing)
 	if err != nil {
 		node.MaybeExitWithAdvice(err)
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -161,7 +161,7 @@ func runStart(cmd *cobra.Command, args []string) {
 	starter, err := provisionWithDriver(cmd, ds, existing)
 	if err != nil {
 		node.MaybeExitWithAdvice(err)
-		machine.MaybeDisplayAdvice(err, viper.GetString("driver"))
+		machine.MaybeDisplayAdvice(err, existing.Driver)
```
sharifelgamal, line 164: this should be ds.Name, existing can be nil

medyagh: good call, thanks

PR #8737, opened 2020-07-17
Use RWMutex instead of Mutex

### LINKED ISSUE #8731: Should they use RWMutex?
Should they use RWMutex instead of Mutex to avoid the contention between readings?

https://github.com/kubernetes/minikube/blob/6a21774a1e2687a669310b78759b244273f83a43/pkg/minikube/registry/registry.go#L135-L153

https://github.com/kubernetes/minikube/blob/edd095bf1f8e1a88e58bc5b383c9f6d049eac3a4/third_party/go9p/srv_file.go#L239-L250

If it's necessary,I am willing to contribute.

dddddai:
Fix #8731

### DIFF
--- a/pkg/minikube/registry/registry.go
+++ b/pkg/minikube/registry/registry.go
@@ -109,7 +109,7 @@ func (d DriverDef) String() string {
 
 type driverRegistry struct {
 	drivers map[string]DriverDef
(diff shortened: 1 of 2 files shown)


dddddai:
@priyawadhwa please take a look : )

PR #8741, opened 2020-07-17
Fix bugs which prevented upgrades from v1.0+ to v1.12

tstromberg:
Fixes a couple of upgrade quirks I noticed during testing:

* Code assumes newer config.Driver is non-empty 
* Code assumed newer cc.Name was non-empty
* Code assumed new paths of key/cert paths
* Code assumed cc.KicBaseImage was non-empty (also fixed by #8738)

This also adds integration tests to assert that we can always upgrade from a stopped v1.0 cluster, or a newer running cluster. 

There are likely other edge-cases provided by random command-line options, but this PR addresses the edge cases that affect most if not all users.


test/integration/version_upgrade_test.go
```diff
@@ -122,15 +207,64 @@ func TestVersionUpgrade(t *testing.T) {
...
 	t.Logf("Attempting restart after unsuccessful downgrade")
 	args = append([]string{"start", "-p", profile, "--memory=2200", fmt.Sprintf("--kubernetes-version=%s", constants.NewestKubernetesVersion), "--alsologtostderr", "-v=1"}, StartArgs()...)
 	rr, err = Run(t, exec.CommandContext(ctx, Target(), args...))
 	if err != nil {
-		t.Errorf("start after failed upgrade: %v", err)
+		t.Errorf("start after failed upgrade: %s: %v", rr.Command(), err)
+	}
+}
+
+// TestMissingContainerUpgrade tests a Docker upgrade where the underlying container is missing
```
sharifelgamal, line 222: This test already exists at HEAD, you can delete this copy.

PR #8751, opened 2020-07-17
Add 'Start Docker' guidance for /pipe/docker_engine errors

### LINKED ISSUE #8748: Improve error message when docker is stopped and attempting to start minikube with docker driver

1. Stop docker
2. minikube start --driver=docker

image

<details>
As you can see the resulting error from minikube does correctly state that docker might be stopped, but it has other text which might be confusing to the user.  You can query status of the docker service without admin, it would be good to have a message similar to that on Mac/Linux which is a bit more clear.
</details>

tstromberg:
Fixes #8748


etanshaul:
thanks for the quick fix

Issue #7756, opened 2020-04-17
docker driver : static ip, make container IPs survive restart.


medyagh:
Feature request: assign static IP for minikube docker/podman containers

currently if we stop minikube, we might actually, have a different IP that will cause a kubeadm Reset (expensive)

```
I0417 15:04:19.699937   20147 kubeadm.go:434] needs reset: configs differ:
	-- stdout --
	--- /var/tmp/minikube/kubeadm.yaml	2020-04-17 22:01:08.390045502 +0000
	+++ /var/tmp/minikube/kubeadm.yaml.new	0001-01-01 00:00:00.000000000 +0000
	@@ -1,7 +1,7 @@
	 apiVersion: kubeadm.k8s.io/v1beta1
	 kind: InitConfiguration
	 localAPIEndpoint:
	-  advertiseAddress: 172.17.0.4
	+  advertiseAddress: 172.17.0.2
	   bindPort: 8443
	 bootstrapTokens:
	   - groups:
	@@ -14,25 +14,25 @@
	   criSocket: /var/run/crio/crio.sock
	   name: "crio-20200417T145706-10869"
	   kubeletExtraArgs:
	-    node-ip: 172.17.0.4
	+    node-ip: 172.17.0.2
	   taints: []
	 ---
	 apiVersion: kubeadm.k8s.io/v1beta1
	 kind: ClusterConfiguration
	 apiServer:
	-  certSANs: ["127.0.0.1", "localhost", "172.17.0.4"]
	+  certSANs: ["127.0.0.1", "localhost", "172.17.0.2"]
```

as seen here:
https://storage.googleapis.com/minikube-builds/logs/7752/775a91c/Docker_Linux.html#fail_TestPause%2fserial%2fSecondStartNoReset

josedonizetti:
@medyagh If we create our own custom network, we can set a fixed ip when running the container `--ip`. The problem of creating our own network is possible conflicts of subnets with existing networks, so we would have to try a few subnets options until a non used one is found. 

Although we currently don't worry about this on the kvm2 driver, there is a fixed subnet that could conflict, and if it does we are not trying other options: https://github.com/kubernetes/minikube/blob/master/pkg/drivers/kvm/network.go#L41

PR #8764, opened 2020-07-18
Add static ip to kic/docker on linux

josedonizetti:
fixes https://github.com/kubernetes/minikube/issues/7756


pkg/drivers/kic/oci/oci.go
```diff
@@ -142,6 +142,13 @@ func CreateContainerNode(p CreateParams) error {
 		"--label", p.NodeLabel,
 	}
 
+	// network
+	if p.OCIBinary == Docker && runtime.GOOS == "linux" { // for now only docker on linux
```
medyagh, line 146: why only on linux?

priyawadhwa:
Tested locally and this works on linux -- still haven't figured out why it isn't working on my mac.

I ran this script across ~15 iterations and the IP stayed constant on linux:

<details>

```bash
#!/bin/bash


while true; do

minikube stop
minikube start --driver docker
STR=$(minikube ssh sudo systemctl status kubelet)
SUB='node-ip=192.168.39.2'
if [[ "$STR" == *"$SUB"* ]]; then
  echo "IP is what we expect."
else
   echo "Node IP is different"
   exit 1
fi
done
```

</details>

pkg/drivers/kic/oci/oci.go
priyawadhwa: I spoke to Jose, and he said:

Docker containers on OSx don’t support ip! It is always localhost
There is why we have the custom tunnel for example

test/integration/fn_mount_cmd_test.go
```diff
@@ -116,7 +116,7 @@ func validateMountCmd(ctx context.Context, t *testing.T, profile string) { // no
 	}
 
 	start := time.Now()
-	if err := retry.Expo(checkMount, time.Millisecond*500, Seconds(15)); err != nil {
+	if err := retry.Expo(checkMount, 5*time.Second, Seconds(15)); err != nil {
```
medyagh, line 119: I think the 5*Second is the amount of time that waits first time before retrying, again, and 15 is total time. is that what you wanted?

priyawadhwa: ah thanks! accidentally left that in from debugging

priyawadhwa:
@medyagh Linux tests are passing now; the failures on KVM are flakes. i reran the tests and they passed the second time around [logs](https://a646c87040050000000000000000001.proxy.googleprod.com/job/KVM_Linux_integration/11599/consoleFull)

PR #8787, opened 2020-07-21
fix error usage for the WaitGroup in the pkg addons

saltbo:
The method `Add` should out in the goroutine, because the `Wait` maybe exec before the goroutine.

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -340,7 +340,6 @@ func verifyAddonStatus(cc *config.ClusterConfig, name string, val string) error 
 
 // Start enables the default addons for a profile, plus any additional
 func Start(wg *sync.WaitGroup, cc *config.ClusterConfig, toEnable map[string]bool, additional []string) {
-	wg.Add(1)
 	defer wg.Done()
 
 	start := time.Now()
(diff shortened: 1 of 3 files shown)


saltbo:
I signed it

medyagh:
@saltbo can you please describe what issue are you trying to fix?

saltbo:
The method Add should out in the goroutine, because the Wait maybe exec before the goroutine.

saltbo:
I signed it

medyagh:
> Add should out in the goroutine, because the Wait maybe exec before the goroutine.

do you have an example minikube issue that this will solve so I can test this PR before and after this PR ?

saltbo:
https://github.com/kubernetes/minikube/blob/ad083ba7acafe4cd6018fb2fc92364e9eb148b33/pkg/addons/addons_test.go#L133

This function`Start` in Test should be called by the goroutine.  Because it is called by Goroutine in actual use.

https://github.com/kubernetes/minikube/blob/ad083ba7acafe4cd6018fb2fc92364e9eb148b33/pkg/minikube/node/start.go#L149

Then, you can found test failed. Because of the `wg.Add(1)` should be executed out of the goroutine.

PR #8796, opened 2020-07-21
Add support for Error type to JSON output

priyawadhwa:
adds support, unit tests, and new integration tests for Error type for JSON output

this is the last PR in adding support for JSON output

sample output with error:
```
$ minikube start --driver fail --output json
{"data":{"currentstep":"0","message":"minikube v1.12.1 on Darwin 10.14.6\n","name":"Initial Minikube Setup","totalsteps":"9"},"datacontenttype":"application/json","id":"4b69ea87-1b11-42cb-8190-dc414545cfd8","source":"https://minikube.sigs.k8s.io/","specversion":"1.0","type":"io.k8s.sigs.minikube.step"}
{"data":{"exitcode":"69","message":"The driver 'fail' is not supported on darwin\n"},"datacontenttype":"application/json","id":"d7440fee-ddf6-4961-a5e6-2c20d502e2a4","source":"https://minikube.sigs.k8s.io/","specversion":"1.0","type":"io.k8s.sigs.minikube.error"}
```

#8655


pkg/minikube/out/register/log.go
```diff
@@ -120,6 +122,26 @@ func NewInfo(message string) *Info {
...
+
+func NewError(err string) *Error {
+	return &Error{
+		map[string]string{
+			"message": err,
+		},
+	}
+}
+
+func NewErrorExitCode(err string, exitcode int, additionalData ...map[string]string) *Error {
```
medyagh, line 136: comment for exported func

priyawadhwa: thanks !

PR #8800, opened 2020-07-22
Improve json integration test

priyawadhwa:
adds a validation to make sure that each "step" log has a distinct step number

this will ensure that if at any point a user changes the logs, they remember to update the logs register

closes https://github.com/kubernetes/minikube/issues/8655


cmd/minikube/cmd/start.go
```diff
@@ -822,21 +821,21 @@ func validateMemorySize(req int, drvName string) {
...
 		if containerLimit < 1991 {
-			out.T(out.Tip, `Increase Docker for Desktop memory to at least 2.5GB or more:
+			out.WarningT(`Increase Docker for Desktop memory to at least 2.5GB or more:
 			
 	Docker for Desktop > Settings > Resources > Memory
 
 `)
 		} else if containerLimit < 2997 && sysLimit > 8000 { // for users with more than 8 GB advice 3 GB
-			out.T(out.Tip, `Your system has {{.system_limit}}MB memory but Docker has only {{.container_limit}}MB. For a better performance increase to at least 3GB.
+			out.WarningT(`Your system has {{.system_limit}}MB memory but Docker has only {{.container_limit}}MB. For a better performance increase to at least 3GB.
```
medyagh, line 838: why would this be a warning? we are only suggesting the suer with a lot of ram to maybe increase their default docker desktop to 3 GB (not required but suggestion)

priyawadhwa: So`out.T` is meant to represent a specific step in the minikube process (e.g. Starting VM, Preparing kubernetes), and that's how JSON interprets calls to `out.T`.

This log is informational or warning in nature. I could change it to `out.Infof` if you think that's more appropriate

priyawadhwa: In general, it's probably worth it to have a doc around what the different functions in the `out` package should be used for

PR #8816, opened 2020-07-23
Add stress test script

tstromberg:
Usage:

`./test/stress/stress.sh "--driver=docker --base-image gcr.io/k8s-minikube/kic:ubuntu-upgrade"`


test/stress/stress.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/sh
```
medyagh, line 1: probably wanna add the boiler plate to this !

Issue #8646, opened 2020-07-05
Failure of "minikube ip" returns "<nil>" with exit code 0 


pre:
**Steps to reproduce the issue:** 

1. Have cluster on not-ready state
2. Run "minikube ip" 

Expected: exit code != 0 with empty output

Actual: exit code == 0 
Output: `<nil>`

```bash
% minikube ip
<nil>
% echo $?
0
```

Minikube v1.11.0

Why is this a problem? 
When `minikube ip` is used in a script, it'd be straightforward to read the exit code for whether the command succeeded. However, now the exit code is always "success" so the script needs to check for failure against a flaky string `<nil>` which may change at a later Minikube release.

afbjorklund:
@pre : thanks for reporting this issue, it should not return `nil` like that. Which OS and which driver did this happen with ?

afbjorklund:
The problem is in the `ControlPlaneEndpoint`, which doesn't check the hostname or the return value of `net.ParseIP`

pre:
In case it matters, I have Minikube v1.11.0 on MacOS 10.15.5 with Hyperkit.

Dean-Coakley:
I'm taking a look at fixing this. But it's my first minikube contribution so if someone wants to take it first/take it over, let me know.

medyagh:
@Dean-Coakley are you still interested in making a PR for this ?

medyagh:
cc: @sharifelgamal

Dean-Coakley:
@medyagh I opened a PR that seems to resolve the issue.

PR #8820, opened 2020-07-24
Fix handling of parseIP error

Dean-Coakley:
Fixes: https://github.com/kubernetes/minikube/issues/8646

### Before changes:
```
$ minikube ip delete
$ minikube start
...
```
* In a separate terminal session
```
$ minikube ip
<nil>
$ echo $?
0
```

### Fix
* (Same repro steps)
```
$ minikube ip

💣  Unable to get forwarded endpoint: failed to parse ip

➜  minikube git:(fix-minikube-ip-errhandle) echo $?
70
```

### DIFF
--- a/pkg/minikube/driver/endpoint.go
+++ b/pkg/minikube/driver/endpoint.go
@@ -17,6 +17,7 @@ limitations under the License.
 package driver
 
 import (
+	"fmt"
 	"net"
 
 	"k8s.io/minikube/pkg/drivers/kic/oci"
@@ -30,6 +31,9 @@ func ControlPlaneEndpoint(cc *config.ClusterConfig, cp *config.Node, driverName 
 		port, err := oci.ForwardedPort(cc.Driver, cc.Name, cp.Port)
 		hostname := oci.DefaultBindIPV4
 		ip := net.ParseIP(hostname)
+		if ip == nil {
+			return hostname, ip, port, fmt.Errorf("failed to parse ip for %q", hostname)
+		}
 
 		// https://github.com/kubernetes/minikube/issues/3878
 		if cc.KubernetesConfig.APIServerName != constants.APIServerName {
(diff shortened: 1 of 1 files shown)


pkg/minikube/driver/endpoint.go
```diff
@@ -30,6 +32,10 @@ func ControlPlaneEndpoint(cc *config.ClusterConfig, cp *config.Node, driverName
 		port, err := oci.ForwardedPort(cc.Driver, cc.Name, cp.Port)
 		hostname := oci.DefaultBindIPV4
 		ip := net.ParseIP(hostname)
+		if ip == nil {
+			errMsg := fmt.Sprintf("failed to parse ip")
+			err = errors.Wrap(err, errMsg)
```
Dean-Coakley, line 37: * Is Wrap correct here?

* I'd like to add additional context, but from my testing the `ParseIP` fails due to an empty `hostname`. So I'm not sure what I can add that is helpful.

pkg/minikube/driver/endpoint.go
```diff
@@ -30,6 +32,10 @@ func ControlPlaneEndpoint(cc *config.ClusterConfig, cp *config.Node, driverName
 		port, err := oci.ForwardedPort(cc.Driver, cc.Name, cp.Port)
 		hostname := oci.DefaultBindIPV4
 		ip := net.ParseIP(hostname)
+		if ip == nil {
+			errMsg := fmt.Sprintf("failed to parse ip")
```
medyagh, line 36: since the var errMsg is not used anywhere else, consider just putting it in the Wrap

pkg/minikube/driver/endpoint.go
medyagh: instead of error.wraap
you can use fmt.Errorf("failed to parse ip for %s", hostname)

pkg/minikube/driver/endpoint.go
```diff
@@ -43,5 +47,9 @@ func ControlPlaneEndpoint(cc *config.ClusterConfig, cp *config.Node, driverName
 	if cc.KubernetesConfig.APIServerName != constants.APIServerName {
 		hostname = cc.KubernetesConfig.APIServerName
 	}
-	return hostname, net.ParseIP(cp.IP), cp.Port, nil
+	ip := net.ParseIP(cp.IP)
+	if ip == nil {
+		return hostname, ip, cp.Port, fmt.Errorf("failed to parse ip for %s", hostname)
```
medyagh, line 52: here it should be failed to parse ip for %q", cp.IP)

Dean-Coakley: Sorry!

Dean-Coakley:
(Kept same commit structure, needed to rebase to fix lint test)

sharifelgamal:
@Dean-Coakley can you rebase against master so we can retest? This change looks good and I'd like to merge it soon.

medyagh:
thank you @Dean-Coakley

PR #8845, opened 2020-07-27
Fixed LoadBalancer deployments example in "Get Started" section.

snappyjig:
The Load balancer deployment section mentions --port=8000 but that's not the correct port for the container. 
Using curl to <EXTERNAL IP>:8000/ gives an "Empty reply from server" error.
The parameter should be 8080 (--port=8080). Also, the reference to EXTERNAL-IP:8000 should be changed to EXTERNAL-IP:8080


medyagh:
@snappyjig please sign the CLA so we could accept this contribution. more info here https://identity.linuxfoundation.org/projects/cncf

RA489:
@snappyjig any issue you are facing while signing the CLA?

tstromberg:
This has been fixed in a separate PR.

PR #8870, opened 2020-07-28
Add warning for ingress addon enabled with driver of none

Dean-Coakley:
Fixes: https://github.com/kubernetes/minikube/issues/8841

* Minikube now prints a warning when using the addon `ingress`, with the driver of `none`, which is unsupported.

### Before
```
sudo ./out/minikube-linux-amd64 start --driver=none --addons=ingress --kubernetes-version=v1.16.0
😄  minikube v1.12.1 on Amazon 2 (xen/amd64)
🏃  Updating the running none "minikube" bare metal machine ...
ℹ️  OS release is Amazon Linux 2
🐳  Preparing Kubernetes v1.16.0 on Docker 19.03.6-ce ...

🔎  Verifying ingress addon...
🌟  Enabled addons: default-storageclass, ingress, storage-provisioner
💗  Kubectl not found in your path
👉  You can use kubectl inside minikube. For more information, visit https://minikube.sigs.k8s.io/docs/handbook/kubectl/
💡  For best results, install kubectl: https://kubernetes.io/docs/tasks/tools/install-kubectl/
```

### After
```
sudo ./out/minikube-linux-amd64 start --addons=ingress --kubernetes-version=v1.16.0
😄  minikube v1.12.1 on Amazon 2 (xen/amd64)
🏃  Updating the running none "minikube" bare metal machine ...
ℹ️  OS release is Amazon Linux 2
🐳  Preparing Kubernetes v1.16.0 on Docker 19.03.6-ce ...

💡  Due to networking limitations of driver none, ingress addon is not supported. Try using a different driver.
```

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -141,15 +141,17 @@ func enableOrDisableAddon(cc *config.ClusterConfig, name string, val string) err
 		}
 	}
 
-	// to match both ingress and ingress-dns adons
-	if strings.HasPrefix(name, "ingress") && enable && driver.IsKIC(cc.Driver) && runtime.GOOS != "linux" {
-		exit.UsageT(`Due to {{.driver_name}} networking limitations on {{.os_name}}, {{.addon_name}} addon is not supported for this driver.
+	// to match both ingress and ingress-dns addons
+	if strings.HasPrefix(name, "ingress") && enable {
+		if driver.IsKIC(cc.Driver) && runtime.GOOS != "linux" || driver.BareMetal(cc.Driver) {
+			exit.UsageT(`Due to {{.driver_name}} networking limitations on {{.os_name}}, {{.addon_name}} addon is not supported for this driver.
 Alternatively to use this addon you can use a vm-based driver:
 
 	'minikube start --vm=true'
 
 To track the update on this work in progress feature please check:
 https://github.com/kubernetes/minikube/issues/7332`, out.V{"driver_name": cc.Driver, "os_name": runtime.GOOS, "addon_name": name})
+		}
 	}
 
 	if strings.HasPrefix(name, "istio") && enable {
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -143,14 +143,17 @@ func enableOrDisableAddon(cc *config.ClusterConfig, name string, val string) err
 
 	// to match both ingress and ingress-dns addons
 	if strings.HasPrefix(name, "ingress") && enable {
-		if driver.IsKIC(cc.Driver) && runtime.GOOS != "linux" || driver.BareMetal(cc.Driver) {
-			exit.UsageT(`Due to {{.driver_name}} networking limitations on {{.os_name}}, {{.addon_name}} addon is not supported for this driver.
+		if driver.IsKIC(cc.Driver) && runtime.GOOS != "linux" {
+			exit.UsageT(`Due to networking limitations of driver {{.driver_name}} on {{.os_name}}, {{.addon_name}} addon is not supported.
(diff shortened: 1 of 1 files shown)


pkg/addons/addons.go
```diff
@@ -141,15 +141,20 @@ func enableOrDisableAddon(cc *config.ClusterConfig, name string, val string) err
 		}
 	}
 
-	// to match both ingress and ingress-dns adons
-	if strings.HasPrefix(name, "ingress") && enable && driver.IsKIC(cc.Driver) && runtime.GOOS != "linux" {
-		exit.UsageT(`Due to {{.driver_name}} networking limitations on {{.os_name}}, {{.addon_name}} addon is not supported for this driver.
+	// to match both ingress and ingress-dns addons
+	if strings.HasPrefix(name, "ingress") && enable {
+		if driver.IsKIC(cc.Driver) && runtime.GOOS != "linux" {
+			exit.UsageT(`Due to networking limitations of driver {{.driver_name}} on {{.os_name}}, {{.addon_name}} addon is not supported.
```
Dean-Coakley, line 147: I changed this text because I thought the text of
"Due to none networking limitations on linux..."

Could be confusing to some minikube users. Let me know if this should be reverted/chaged.

Alternatively, the driver name could be quoted to prevent confusion:
```suggestion
			exit.UsageT(`Due to "{{.driver_name}}" networking limitations on {{.os_name}}
```

sharifelgamal: This wording seems fine to me.

pkg/addons/addons.go
```diff
@@ -141,15 +141,20 @@ func enableOrDisableAddon(cc *config.ClusterConfig, name string, val string) err
...
 Alternatively to use this addon you can use a vm-based driver:
 
 	'minikube start --vm=true'
 
 To track the update on this work in progress feature please check:
 https://github.com/kubernetes/minikube/issues/7332`, out.V{"driver_name": cc.Driver, "os_name": runtime.GOOS, "addon_name": name})
+		} else if driver.BareMetal(cc.Driver) {
+			exit.UsageT(`Due to networking limitations of driver {{.driver_name}}, {{.addon_name}} addon is not supported.
+Try using a supported driver: "--driver=docker", or "--driver=podman", or "--driver=hyperkit"`, out.V{"driver_name": cc.Driver, "addon_name": name})
```
sharifelgamal, line 156: I would mention kvm2 instead of hyperkit as an alternative since the none driver only works on linux and hyperkit is a macos vm driver. We could also forgo giving alternate driver names altogether and just say to use a different driver.

Dean-Coakley: I'm going to remove the list for now.

I think different makes more sense. Any other driver available on linux should work.

Dean-Coakley:
@sharifelgamal Excuse my unfamiliarity,

Is this PR just pending final review or is there outstanding work I need to do? 
Do I need to babysit the PR and /retest until everything is green?

sharifelgamal:
Hey @Dean-Coakley, sorry the delay here. Can you fix the addons test (addons_test.go) to not try to enable ingress on none? Right now that test is failing because we're not expecting your new warning message.

Dean-Coakley:
Ah looks fixed in: https://github.com/kubernetes/minikube/commit/0afb311c5d571daa86a8cdefb188c726f49c2ce4

Thanks!

gattytto:
@Dean-Coakley I am using minikube inside LXC with the driver=none and I am having to get back to a version of minikube without this to be able to deploy the nginx ingress controller (which works perfectly in my usecase). Please consider adding a bypass arg like "--skip-none-ingress-check"

Issue #8857, opened 2020-07-27
run internal kubectl when os.Args[0] is 'kubectl'


drewpca:
It would be nice if this worked:

```
% sudo install minikube-linux-amd64 /usr/local/bin/minikube
% sudo ln -s /usr/local/bin/minikube /usr/local/bin/kubectl
% kubectl version
```
No harm to existing workflows (as long as we don't force the creation of that symlink-- see https://github.com/rancher/k3s/issues/521), and it would save on aliases and '--' argument separators.

Related: https://github.com/kubernetes/minikube/issues/4684
Related: https://github.com/kubernetes/minikube/issues/4697

afbjorklund:
I like this suggestion, we talked about our solutions like providing an "--install" flag or echoing the program path.

kadern0:
@sharifelgamal this has been fixed with #8872

sharifelgamal:
Excellent, thank you!

PR #8872, opened 2020-07-29
Running internal kubectl command when minikube is called as 'kubectl'

kadern0:
Fixes issue #8857

```
./out/minikube version
minikube version: v1.12.1
commit: f8406b6f02d65c1cf8d42219c30d24acedc9e8c3-dirty
```
```
 ./out/minikube kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:14:22Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:07:13Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
```
```
 ./kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:14:22Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:07:13Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
```

### DIFF
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -20,6 +20,7 @@ import (
 	goflag "flag"
 	"fmt"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strings"
 
@@ -73,6 +74,10 @@ var RootCmd = &cobra.Command{
 // Execute adds all child commands to the root command sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
+	_, callingCmd := filepath.Split(os.Args[0])
+	if callingCmd == "kubectl" {
+		os.Args = append([]string{"minikube", callingCmd}, os.Args[1:]...)
+	}
 	for _, c := range RootCmd.Commands() {
 		c.Short = translate.T(c.Short)
 		c.Long = translate.T(c.Long)
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -75,8 +75,9 @@ var RootCmd = &cobra.Command{
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
 	_, callingCmd := filepath.Split(os.Args[0])
+
 	if callingCmd == "kubectl" {
-		os.Args = append([]string{"minikube", callingCmd}, os.Args[1:]...)
+		os.Args = append([]string{RootCmd.Use, callingCmd}, os.Args[1:]...)
 	}
 	for _, c := range RootCmd.Commands() {
 		c.Short = translate.T(c.Short)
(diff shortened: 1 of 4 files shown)


kadern0:
@medyagh I'd say adding an entry to the f.a.q pages and also documenting it here?

https://github.com/kubernetes/minikube/blob/master/site/content/en/docs/handbook/kubectl.md

What do you think?

cmd/minikube/cmd/root.go
```diff
@@ -73,6 +74,10 @@ var RootCmd = &cobra.Command{
 // Execute adds all child commands to the root command sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
+	_, callingCmd := filepath.Split(os.Args[0])
+	if callingCmd == "kubectl" {
+		os.Args = append([]string{"minikube", callingCmd}, os.Args[1:]...)
```
medyagh, line 79: would this work in cases in integration test if the binary is called

"minikube-darwin-amd64" ? or only works if binary is called "minikube"?  do simmilar tools, support alternative binary names such as "minkube-linux-amd64" ? or "minikube.exe"

kadern0: Interesting question. I think the binary name doesn't matter once we reach such instruction. This is what comes next:
```
for _, c := range RootCmd.Commands()
```
As long as minikube can correctly understand the content of "RootCmd.Commands()" the binary name shouldn't matter. And I don't see any logic for different platforms. Am I right?

kadern0:
> do you mind adding an integration test for this? we already have a kubectl integraiton test
> we can add something like this to it (it could be a Linux Only Test if it is hard to implement for windows)
> 
> ```
> dir, err := ioutil.WhateverMakeTempDirectory()
> dest := , filepath.Join(dir, "kubectl")
> copy.Copy("minikube", dest)
> exec("PATH=%s", dir, "kubectl get po -A")
> ```

Instead of copying the binary, I've created a hard link.

cmd/minikube/cmd/root.go
kadern0: Instead of hardcoding "minikube" I've opted for using the value within the `RootCmd.Use` variable. I assume this will handle all the different cases.

Issue #8708, opened 2020-07-13
Advice user when they allocate 80% of their ram to minikube


saikatHi6:
I want to set up a minikube dev environment for my application which has 20+ microservices. My Host machine has 16 CPU and 64 GB RAM. I tried to run the minikube cluster using Virtualbox. Cluster is working when I assign 8 CPU but its failing after deploying few services. So I wanted to increase CPU and RAM. Can anyone help with the same?
1. Installed virtualbox
2. minikube start --memory=51238 --cpus=12 --vm-driver=virtualbox 
3. 

**Full output of failed command:** 
* minikube v1.10.1 on Microsoft Windows 10 Enterprise 10.0.18362 Build 18362
* Using the virtualbox driver based on existing profile
* Starting control plane node minikube in cluster minikube
* Restarting existing virtualbox VM for "minikube" ...
* minikube 1.12.0 is available! Download it: https://github.com/kubernetes/minikube/releases/tag/v1.12.0
* To disable this notice, run: 'minikube config set WantUpdateNotification false'

* Preparing Kubernetes v1.18.2 on Docker 19.03.8 ...
! Unable to restart cluster, will reset it: apiserver healthz: apiserver process never appeared
! initialization failed, will try again: run: /bin/bash -c "sudo env PATH=/var/lib/minikube/binaries/v1.18.2:$PATH kubeadm init --config /var/tmp/minikube/kubeadm.yaml  --ignore-preflight-errors=DirAvailable--etc-kubernetes-manifests,DirAvailable--var-lib-minikube,DirAvailable--var-lib-minikube-etcd,FileAvailable--etc-kubernetes-manifests-kube-scheduler.yaml,FileAvailable--etc-kubernetes-manifests-kube-apiserver.yaml,FileAvailable--etc-kubernetes-manifests-kube-controller-manager.yaml,FileAvailable--etc-kubernetes-manifests-etcd.yaml,Port-10250,Swap": Process exited with status 1
stdout:
[init] Using Kubernetes version: v1.18.2
[preflight] Running pre-flight checks
[preflight] Pulling images required for setting up a Kubernetes cluster
[preflight] This might take a minute or two, depending on the speed of your internet connection
[preflight] You can also perform this action in beforehand using 'kubeadm config images pull'
[kubelet-start] Writing kubelet environment file with flags to file "/var/lib/kubelet/kubeadm-flags.env"
[...]

PR #8877, opened 2020-07-29
advice user when allocate more than 80% of their memory

medyagh:
closes this https://github.com/kubernetes/minikube/issues/8708
also fixed a bug that the advice was showing twice.
and also fixed the bug that we were not checking for allowing max
also moves all the calls to getting CPU/memory info into a cached func for less overhead

## before this PR

allowing more than max memory was allowed.

#### Don't allow 17GB
```
medya@~/workspace/minikube (max_memory) $ make && ./out/minikube start --driver=hyperkit --memory=17gb

💡  To suppress memory validations you can use --force flag.
💣  Requested memory allocation 17408MB is more than your system limit 16384MB. Try specifying a lower memory:

        miniube start --memory=8192mb
```

#### Advice more than 80% allocation.
```

medya@~/workspace/minikube (max_memory) $ make && ./out/minikube start --driver=hyperkit --memory=15gb
❗  You are allocating 15360MB to memory and your system only has 16384MB. You might face issues. try specifying a lower memory:

                miniube start --memory=8192mb


💡  To suppress and ignore this warning you can use --force flag.
🔥  Creating hyperkit VM (CPUs=2, Memory=15360MB, Disk=20000MB) ...
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -37,7 +37,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/shirou/gopsutil/cpu"
 	gopshost "github.com/shirou/gopsutil/host"
-	"github.com/shirou/gopsutil/mem"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	cmdcfg "k8s.io/minikube/cmd/minikube/cmd/config"
@@ -764,13 +763,13 @@ func validateUser(drvName string) {
 	}
 }
 
-// memoryLimits returns the amount of memory allocated to the system and hypervisor
+// memoryLimits returns the amount of memory allocated to the system and hypervisor , the return value is in MB
 func memoryLimits(drvName string) (int, int, error) {
-	v, err := mem.VirtualMemory()
+	info, err := machine.CachedHostInfo()
 	if err != nil {
 		return -1, -1, err
 	}
-	sysLimit := int(v.Total / 1024 / 1024)
+	sysLimit := int(info.Memory)
 	containerLimit := 0
 
 	if driver.IsKIC(drvName) {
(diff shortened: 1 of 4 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -908,11 +934,6 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 	validateCPUCount(drvName)
 
 	if cmd.Flags().Changed(memory) {
-		req, err := util.CalculateSizeInMB(viper.GetString(memory))
```
sharifelgamal, line 911: Why get rid of this check?

medyagh: because it was doing it twice, and was priniting the advice two times.

medyagh: it was move to line 397 to be done all the time, instead of when check change memory

medyagh: also was moved to 397 because at that time we have the suggested auto calucated memory. we will need to verify that one is allowed too

pkg/minikube/machine/info.go
```diff
@@ -97,3 +99,45 @@ func logRemoteOsRelease(r command.Runner) {
...
+var cachedSystemMemoryErr *error
+
+//  cachedSysMemLimit will return a chaced limit for the system's virtual memory.
+func cachedSysMemLimit() (*mem.VirtualMemoryStat, error) {
+	if cachedSystemMemoryLimit == nil {
+		v, err := mem.VirtualMemory()
+		cachedSystemMemoryLimit = v
+		cachedSystemMemoryErr = &err
+	}
+	return cachedSystemMemoryLimit, *cachedSystemMemoryErr
```
sharifelgamal, line 113: Do we really want to cache the error? What if the error isn't persistent?

pkg/minikube/machine/info.go
```diff
@@ -97,3 +99,45 @@ func logRemoteOsRelease(r command.Runner) {
 
 	glog.Infof("Remote host: %s", osReleaseInfo.PrettyName)
 }
+
+var cachedSystemMemoryLimit *mem.VirtualMemoryStat
+var cachedSystemMemoryErr *error
+
+//  cachedSysMemLimit will return a chaced limit for the system's virtual memory.
```
sharifelgamal, line 106: fix typo

pkg/minikube/machine/info.go
```diff
@@ -97,3 +99,45 @@ func logRemoteOsRelease(r command.Runner) {
...
+		cachedSystemMemoryErr = &err
+	}
+	return cachedSystemMemoryLimit, *cachedSystemMemoryErr
+}
+
+var cachedDiskInfo *disk.UsageStat
+var cachedDiskInfoeErr *error
+
+//  cachedSysMemLimit will return a cached disk usage info
+func cachedDisInfo() (disk.UsageStat, error) {
```
sharifelgamal, line 120: fix function name

cmd/minikube/cmd/start.go
```diff
@@ -822,13 +821,41 @@ func suggestMemoryAllocation(sysLimit int, containerLimit int, nodes int) int {
 	return suggested
 }
 
+// validateMemoryHardLimit checks if the user system has enough memory at all !
+func validateMemoryHardLimit(drvName string) {
+	s, c, err := memoryLimits(drvName)
+	if err != nil {
+		glog.Warningf("Unable to query memory limits: %v", err)
+		out.T(out.Conflict, "Failed to verify system memory limits.")
```
sharifelgamal, line 829: this should probably be a warning

pkg/minikube/machine/info.go
medyagh: yeah, each time it is called a new error is cached. this is good so when we return a cached item, we can handle it with the same cached error

PR #8878, opened 2020-07-30
Update Japanese translation

inductor:
Add some translation especially since Docker driver has been enabled by default


inductor:
@josedonizetti
@medyagh 

Can you please have a look?

PR #8895, opened 2020-07-30
update crio to 1.18.3 and kicbase to ubuntu 20.04

sharifelgamal:
Currently versions of crio above 1.17.3 (which magically vanished into the ether) fails on a node restart, so I've had an admittedly horrible hack to fix it while we discover the root cause.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -207,6 +207,9 @@ func runStart(cmd *cobra.Command, args []string) {
 		out.WarningT("CRI-O currently has an issue and your cluster needs to be restarted.")
 		stopProfile(existing.Name)
 		starter, err = provisionWithDriver(cmd, ds, existing)
+		if err != nil {
+			exit.WithError("error provisioning host", err)
+		}
 	}
 
 	kubeconfig, err := startWithDriver(cmd, starter, existing)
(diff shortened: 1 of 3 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -202,6 +202,16 @@ func runStart(cmd *cobra.Command, args []string) {
 		}
 	}
 
+	if existing != nil && existing.KubernetesConfig.ContainerRuntime == "crio" && driver.IsKIC(existing.Driver) {
```
tstromberg, line 205: I'm not following this -- doesn't this run only if the previous provisionWithDriver error == nil?

sharifelgamal: correct, provision isn't what fails for crio, it's starting k8s that fails. what I found works is if we just stop and start again here, everything works as expected.

I'm basically emulating restarting and failing, then calling minikube stop then calling minikube start again, except I'm skipping the 6 minutes we need to wait for apiserver to not come back in the startWithDriver portion of the code.

again, I have no idea why, but empirically it works and it makes the tests pass.

cmd/minikube/cmd/start.go
```diff
@@ -202,6 +202,17 @@ func runStart(cmd *cobra.Command, args []string) {
...
+		// Stop and start again if it's crio because crio is bad and dumb
```
medyagh, line 206: Please change this unhelpful language.

PR #8901, opened 2020-08-02
Use mebibytes instead of megabytes

bluestealth:
It's a minor nitpick but I believe that the hard drive and memory assigned while using KVM and platforms using raw disks was meant to be MiB based on [this](https://github.com/kubernetes/minikube/blob/294a5c3e7b363a999bdb5f73ef6e1b2b8b275193/pkg/util/utils.go#L35-L48). This patch simply calculates disk size in MiB and assigns the correct unit in the domain configruation for libvirt.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -29,6 +29,7 @@ import (
 	"strings"
 
 	"github.com/blang/semver"
+	"github.com/docker/go-units"
 	"github.com/docker/machine/libmachine/ssh"
 	"github.com/golang/glog"
 	"github.com/google/go-containerregistry/pkg/authn"
@@ -793,7 +794,7 @@ func memoryLimits(drvName string) (int, int, error) {
 		if err != nil {
 			return -1, -1, err
 		}
-		containerLimit = int(s.TotalMemory / 1024 / 1024)
+		containerLimit = int(s.TotalMemory / units.MiB)
 	}
 
 	return sysLimit, containerLimit, nil
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -29,7 +29,6 @@ import (
 	"strings"
 
 	"github.com/blang/semver"
-	"github.com/docker/go-units"
 	"github.com/docker/machine/libmachine/ssh"
 	"github.com/golang/glog"
 	"github.com/google/go-containerregistry/pkg/authn"
@@ -794,7 +793,7 @@ func memoryLimits(drvName string) (int, int, error) {
 		if err != nil {
 			return -1, -1, err
 		}
-		containerLimit = int(s.TotalMemory / units.MiB)
+		containerLimit = util.ConvertBytesToMB(s.TotalMemory)
 	}
 
 	return sysLimit, containerLimit, nil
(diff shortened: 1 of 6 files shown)


afbjorklund:
> It's a minor nitpick but I believe that the hard drive and memory assigned while using KVM and platforms using raw disks was meant to be MiB based on [this](https://github.com/kubernetes/minikube/blob/294a5c3e7b363a999bdb5f73ef6e1b2b8b275193/pkg/util/utils.go#L35-L48). This patch simply calculates disk size in MiB and assigns the correct unit in the domain configruation for libvirt.

This PR does a whole lot  more than just change the output from MB to MiB, though.

Including changing the memory allocation from 2000 and disk allocation from 20000

It would be easier to review, if you opened a issue first describing what you want to "fix"

I think the libvirt unit mismatch was something we inherited from upstream (kvm) actually:

https://github.com/dhiltgen/docker-machine-kvm/issues/54

cmd/minikube/cmd/start_flags.go
```diff
@@ -65,7 +65,7 @@ const (
 	kvmHidden               = "kvm-hidden"
 	minikubeEnvPrefix       = "MINIKUBE"
 	installAddons           = "install-addons"
-	defaultDiskSize         = "20000mb"
+	defaultDiskSize         = "20gb"
```
afbjorklund, line 68: This is not about mebibytes

cmd/minikube/cmd/start_flags.go
```diff
@@ -93,9 +93,9 @@ const (
 	waitTimeout             = "wait-timeout"
 	nativeSSH               = "native-ssh"
 	minUsableMem            = 1024 // Kubernetes will not start with less than 1GB
-	minRecommendedMem       = 2000 // Warn at no lower than existing configurations
+	minRecommendedMem       = 2048 // Warn at no lower than existing configurations
```
afbjorklund, line 96: This is not about mebibytes

cmd/minikube/cmd/start_flags.go
```diff
@@ -93,9 +93,9 @@ const (
...
 	minimumCPUS             = 2
-	minimumDiskSize         = 2000
+	minimumDiskSize         = 2048
```
afbjorklund, line 98: This is not about mebibytes

bluestealth: > > This patch simply calculates disk size in MiB and assigns the correct unit in the domain configruation for libvirt.
> 
> This PR does a whole lot more than just change the output from MB to MiB, though.
> 
> Including changing the memory allocation from 2000 and disk allocation from 20000
> 
> It would be easier to review, if you opened a issue first describing what you want to "fix"
> 
> I think the libvirt unit mismatch was something we inherited from upstream (kvm) actually:
> 
> [dhiltgen/docker-machine-kvm#54](https://github.com/dhiltgen/docker-machine-kvm/issues/54)

That is true, and my earlier change didn't include any changes to start_flags... I could easily drop those.. they are unnecessary. That was mainly to align allocations to GiB... that should probably be in another PR after opening up an issue.

medyagh:
@bluestealth  please pull master on this

tstromberg (approved):

tstromberg:
Thank you!

Issue #8876, opened 2020-07-29
upgrade storage provisioner image and test


priyawadhwa:
related to https://github.com/kubernetes/minikube/issues/7828

#8564

MOZGIII:
Could this help with https://github.com/kubernetes/minikube/issues/8799?

PR #8909, opened 2020-08-03
Upgrade storage provisioner image 

priyawadhwa:
This PR:

1. Upgrades the storage provisioner image to v2. It also makes it easier to upgrade in the future (all we need to do now is upgrade the version in the makefile & push the image)
1. Fixes a bug in the storage provisioner, where we were not applying the ClusterRoleBindings to the correct service account that the storage provisioner pod was running under
1. Beefs up the integration test to make sure files are actually persisted in PVs across pod deletion
1. Upgrades preload to v5 since the storage provisioner image is one of the images in the preload tarball

Fixes https://github.com/kubernetes/minikube/issues/8876


sharifelgamal:
As long as the new tests pass this LGTM

priyawadhwa:
Merging, as the `TestDownloadOnly` and `TestStartStop` failures are happening because `v5` of preload hasn't been generated yet. I'll start a manual rebuild of preload tarballs after merge.

PR #8912, opened 2020-08-04
wip: Add driver alias for kvm -> kvm2

### LINKED ISSUE #8822: Add driver alias for kvm -> kvm2
Since we've dropped support for the original kvm like, 3 years ago, it would make sense to support users who drop the 2.

Example: https://twitter.com/csfinch/status/1286501747729604612/photo/1

Yogendra0Sharma:
fixes #8822 

 if the user supplies kvm as a driver name, it should just treat it as if they supplied kvm2 for everything.
Check on start and convert a kvm passed into kvm2.


cmd/minikube/cmd/start.go
```diff
@@ -520,6 +523,9 @@ func selectDriver(existing *config.ClusterConfig) (registry.DriverState, []regis
 	if d := viper.GetString("vm-driver"); d != "" {
 		ds := driver.Status(viper.GetString("vm-driver"))
 		if ds.Name == "" {
+			if d == "kvm" {
+				d = "kvm2"
```
medyagh, line 527: I think it is safer to also use viper.Set... (to set it globally, other parts of the code I think they might get the driver from the Viper.

Yogendra0Sharma: @medyagh  please verify the below code. is it fine?
```

if d == "kvm" {
        viper.Set("driver","kvm2")
	d = "kvm2"
}
```

medyagh: yes

Issue #8894, opened 2020-07-30
add instructions how to prune images in solution message


medyagh:
```
* [cloud-code-minikube] minikube v1.12.1 on Darwin 10.15.5
  - MINIKUBE_WANTUPDATENOTIFICATION=false
* Using the docker driver based on user configuration
* Starting control plane node cloud-code-minikube in cluster cloud-code-minikube
* Downloading Kubernetes v1.18.3 preload ...
    > preloaded-images-k8s-v4-v1.18.3-docker-overlay2-amd64.tar.lz4: 2.42 MiB ...
* Stopping "cloud-code-minikube" in docker ...
* Deleting "cloud-code-minikube" in docker ...
! StartHost failed, but will try again: creating host: create: creating: create kic node: container name "cloud-code-minikube": container exited unexpectedly
* Creating docker container (CPUs=2, Memory=7918MB) ...
* Failed to start docker container. "minikube start -p cloud-code-minikube" may fix it: creating host: create: creating: create kic node: container name "cloud-code-minikube": container exited unexpectedly

! The minikube docker container exited unexpectedly.
* If you are still interested to make docker driver work. The following suggestions might help you get passed this issue:
* 
	- Prune unused docker images, volumes and abandoned containers.
* 
	- Restart your docker service
* 
	- Ensure your docker daemon has access to enough CPU/memory resources. 
* 
	- Docs https://docs.docker.com/docker-for-mac/#resources
* 
	- Delete and recreate minikube cluster
		minikube delete
		minikube start --driver=docker
* 
X error provisioning host: Failed to start host: creating host: create: creating: create kic node: container name "cloud-code-minikube": container exited unexpectedly
* 
* minikube is exiting due to an error. If the above message is not useful, open an issue:
```

Dean-Coakley:
/assign

No need to add instructions for the other suggestions as they are a bit more trivial?
```
	- Restart your docker service
* 
	- Ensure your docker daemon has access to enough CPU/memory resources. 
* 
```

Dean-Coakley:
@tstromberg Was this closed in error? Is https://github.com/kubernetes/minikube/pull/8918 still required?

afbjorklund:
Maybe it was the "support" label that got confusing, looked more like "documentation"

PR #8918, opened 2020-08-05
Add prune docker images instructions

Dean-Coakley:
Fixes: https://github.com/kubernetes/minikube/issues/8894

Add instructions on the docker command required to remove the following:
  - all stopped containers
  - all networks not used by at least one container
  - all volumes not used by at least one container
  - all dangling images
  - all dangling build cache

#### Output:
```
💡  If you are still interested to make docker driver work. The following suggestions might help you get passed this issue:

	- Prune unused docker images, volumes, networks and abandoned containers.
		
		docker system prune --volumes

	- Restart your docker service

	- Delete and recreate minikube cluster
		minikube delete

```

### DIFF
--- a/pkg/minikube/machine/advice.go
+++ b/pkg/minikube/machine/advice.go
@@ -41,6 +41,7 @@ func MaybeDisplayAdvice(err error, driver string) {
 		out.T(style.Tip, "If you are still interested to make {{.driver_name}} driver work. The following suggestions might help you get passed this issue:", out.V{"driver_name": driver})
 		out.T(style.Empty, `
 	- Prune unused {{.driver_name}} images, volumes, networks and abandoned containers.
+
 		docker system prune --volumes`, out.V{"driver_name": driver})
 		out.T(style.Empty, `
 	- Restart your {{.driver_name}} service`, out.V{"driver_name": driver})
--- a/pkg/minikube/machine/advice.go
+++ b/pkg/minikube/machine/advice.go
@@ -40,7 +40,8 @@ func MaybeDisplayAdvice(err error, driver string) {
 	if errors.Is(err, oci.ErrExitedUnexpectedly) || errors.Is(err, oci.ErrDaemonInfo) {
 		out.T(style.Tip, "If you are still interested to make {{.driver_name}} driver work. The following suggestions might help you get passed this issue:", out.V{"driver_name": driver})
 		out.T(style.Empty, `
-	- Prune unused {{.driver_name}} images, volumes and abandoned containers.`, out.V{"driver_name": driver})
+	- Prune unused {{.driver_name}} images, volumes, networks and abandoned containers.
+		docker system prune --volumes`, out.V{"driver_name": driver})
 		out.T(style.Empty, `
 	- Restart your {{.driver_name}} service`, out.V{"driver_name": driver})
 		if runtime.GOOS != "linux" {
--- a/pkg/minikube/machine/advice.go
+++ b/pkg/minikube/machine/advice.go
@@ -39,10 +39,12 @@ func MaybeDisplayAdvice(err error, driver string) {
 
 	if errors.Is(err, oci.ErrExitedUnexpectedly) || errors.Is(err, oci.ErrDaemonInfo) {
 		out.T(style.Tip, "If you are still interested to make {{.driver_name}} driver work. The following suggestions might help you get passed this issue:", out.V{"driver_name": driver})
-		out.T(style.Empty, `
+		if driver == oci.Docker || driver == oci.Podman {
+			out.T(style.Empty, `
 	- Prune unused {{.driver_name}} images, volumes, networks and abandoned containers.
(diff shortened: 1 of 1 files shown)


pkg/minikube/machine/advice.go
```diff
@@ -39,7 +39,8 @@ func MaybeDisplayAdvice(err error, driver string) {
 	if errors.Is(err, oci.ErrExitedUnexpectedly) || errors.Is(err, oci.ErrDaemonInfo) {
 		out.T(out.Tip, "If you are still interested to make {{.driver_name}} driver work. The following suggestions might help you get passed this issue:", out.V{"driver_name": driver})
 		out.T(out.Empty, `
-	- Prune unused {{.driver_name}} images, volumes and abandoned containers.`, out.V{"driver_name": driver})
+	- Prune unused {{.driver_name}} images, volumes, networks and abandoned containers.
+		docker system prune --volumes`, out.V{"driver_name": driver})
```
medyagh, line 43: this only prunes volumes, how about prune everythign ?
and do you mind posting the Output after this PR ?
to make sure the indentation looks good ?

priyawadhwa:
Hey @Dean-Coakley are you still working on this?

Dean-Coakley:
@priyawadhwa I forgot about this PR.

I'll resume working on this.

pkg/minikube/machine/advice.go
Dean-Coakley: > this only prunes volumes, how about prune everythign ?

This does prune everything, right?
```
$ docker system prune 
WARNING! This will remove:
  - all stopped containers
  - all networks not used by at least one container
  - all dangling images
  - all dangling build cache

Are you sure you want to continue? [y/N] n

$ docker system prune --volumes
WARNING! This will remove:
  - all stopped containers
  - all networks not used by at least one container
  - all volumes not used by at least one container
  - all dangling images
  - all dangling build cache

Are you sure you want to continue? [y/N] 
```
If you think the command is confusing, I agree, surely it should be something like: `docker system prune --include-volumes` ? :smile: 

> and do you mind posting the Output after this PR ?
to make sure the indentation looks good ?

done

> and is this command good on windows too ?

Pretty sure, but not certain. I'll update you if/when I confirm.

> also consider adding one line before and after the command to make it stand out for better copy pasting.

Done, should I add the same to https://github.com/kubernetes/minikube/blob/42792d34c22aee48611f31ee7c605a61f7efffcd/pkg/minikube/machine/advice.go#L60-L61 , for consistency and better copy and paste there too?

Dean-Coakley: > also consider adding one line before and after the command to make it stand out for better copy pasting.

I lost these changes during rebase off master. Need to retest that output is ok.

pkg/minikube/machine/advice.go
```diff
@@ -40,7 +40,9 @@ func MaybeDisplayAdvice(err error, driver string) {
 	if errors.Is(err, oci.ErrExitedUnexpectedly) || errors.Is(err, oci.ErrDaemonInfo) {
 		out.T(style.Tip, "If you are still interested to make {{.driver_name}} driver work. The following suggestions might help you get passed this issue:", out.V{"driver_name": driver})
 		out.T(style.Empty, `
-	- Prune unused {{.driver_name}} images, volumes and abandoned containers.`, out.V{"driver_name": driver})
+	- Prune unused {{.driver_name}} images, volumes, networks and abandoned containers.
+
+		docker system prune --volumes`, out.V{"driver_name": driver})
```
priyawadhwa, line 45: We use {{.driver_name}} because it could be either docker or podman -- I think `podman` has a similar `system prune` command, so this could read:

```
{{,driverName}} system prune --volumes
```

priyawadhwa (commented):
Thanks for including this! Left a comment.

pkg/minikube/machine/advice.go
Dean-Coakley: Good catch!

I validated by the way:
```bash
$ podman system prune --help
Remove unused data

Description:

	podman system prune

        Remove unused data


Usage:
  podman system prune [flags]

Examples:
  podman system prune

Flags:
  -a, --all       Remove all unused data
  -f, --force     Do not prompt for confirmation.  The default is false
      --volumes   Prune volumes
```

Dean-Coakley: @priyawadhwa Do we know for certainty that driver is either `docker|podman` and not possibly anything else?

Dean-Coakley: Seems like it could. Made it conditional: https://github.com/kubernetes/minikube/pull/8918/commits/dcd00b2150f3555c3a52bd6250af880790185a90#diff-dbf20f1c6e11289e51f6072d04612f9dR42

medyagh (approved):

medyagh:
thanks @Dean-Coakley

PR #8948, opened 2020-08-10
Added startHostTimeout flag

### LINKED ISSUE #8897: StartHost timeout should be configurable
https://github.com/kubernetes/minikube/issues/8512 mentions everything.

My actual use case is on Hyper-V, when creating a disk image for 128 GBs takes more than 240 seconds (the hardcoded timeout), and thus the whole `minikube start` fails.

I guess, this might be two things

* One is to allow configuring the timeout.
* Second is figuring out why does the creation take so long, anyway. During the creation of the VM, it seems like the disk first gets created, then it gets converted. Not sure why would either of these operations take long time, since I am on NVMe PCIe3 SSD, and even filling the file with zeroes should be a matter of several seconds.

I think if we have configurabl timeout, then I can live with this. BUt if there's a way to make the creationg faster (Well withing the hard coded timeout), that would work too.

kadern0:

@sharifelgamal can you have a look at this?

would close this https://github.com/kubernetes/minikube/issues/8961

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -91,7 +91,6 @@ const (
 	dryRun                  = "dry-run"
 	interactive             = "interactive"
 	waitTimeout             = "wait-timeout"
-	startHostTimeout        = "starthost-timeout"
 	nativeSSH               = "native-ssh"
 	minUsableMem            = 1024 // Kubernetes will not start with less than 1GB
 	minRecommendedMem       = 2000 // Warn at no lower than existing configurations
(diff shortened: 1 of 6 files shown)


pkg/minikube/machine/start.go
```diff
@@ -157,7 +157,7 @@ func createHost(api libmachine.API, cfg *config.ClusterConfig, n *config.Node) (
 	cstart := time.Now()
 	glog.Infof("libmachine.API.Create for %q (driver=%q)", cfg.Name, cfg.Driver)
 
-	if err := timedCreateHost(h, api, 4*time.Minute); err != nil {
+	if err := timedCreateHost(h, api, cfg.StartHostTimeout); err != nil {
```
sharifelgamal, line 160: We want to be backwards compatible here, so if StartHostTimeout isn't set in cfg for some reason, we still want to default back to 4 minutes.

cmd/minikube/cmd/start_flags.go
```diff
@@ -91,6 +91,7 @@ const (
 	dryRun                  = "dry-run"
 	interactive             = "interactive"
 	waitTimeout             = "wait-timeout"
+	startHostTimeout        = "starthost-timeout"
```
medyagh, line 94: we already have a waitTimeout flag "wait-timeout"
it might have been unued. make sure it is used 

and do you mind sharing the details of a case that you needed this flag ? I am curious, if this would make sense to increase it for everyone by default?

pkg/minikube/machine/start.go
medyagh: I am ok with default being changed, I think we could and should re-use the existing flag we have
but update the description

```
startCmd.Flags().Duration(waitTimeout, 6*time.Minute, "max time to wait for kubernetes or host to be healthy")
```

cmd/minikube/cmd/start_flags.go
kadern0: The use case is described in the issue #8897. Quoted (not mine):
```
My actual use case is on Hyper-V, when creating a disk image for 128 GBs takes more than 240 seconds (the hardcoded timeout), and thus the whole minikube start fails.
```

(Basically what you said in your last comment.)

pkg/minikube/machine/start.go
kadern0: I agree with the reuse of the ``waitTimeout`` flag as @medyagh suggests. Do we default to 6 mins instead of the current 4 if the user doesn't specify any timeout  (to maintain consistency)?

medyagh: Yes

PR #8969, opened 2020-08-11
skip validations if --force is supplied

priyawadhwa:
This will fix a cloud shell bug

output now:

```
$ minikube start --driver docker --download-only --force
😄  minikube v1.12.2 on Darwin 10.14.6
❗  minikube skips various validations when --force is supplied; this may lead to unexpected behavior
🚜  Pulling base image ...
✅  Download complete!
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -673,7 +673,7 @@ func validateDriver(ds registry.DriverState, existing *config.ClusterConfig) {
 			}
 			exit.WithCodeT(exit.Unavailable, "{{.driver}} does not appear to be installed", out.V{"driver": name})
 		}
-		skipErrorIfForceEnabled(exit.Unavailable, "Failed to validate '{{.driver}}' driver", out.V{"driver": name})
+		exitIfNotForced(exit.Unavailable, "Failed to validate '{{.driver}}' driver", out.V{"driver": name})
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -1210,3 +1202,10 @@ func getKubernetesVersion(old *config.ClusterConfig) string {
 
 	return version.VersionPrefix + nvs.String()
 }
+
+func skipErrorIfForceEnabled(code int, message string, v out.V) {
```
medyagh, line 1206: rename to something more readable, maybe

"exitIfPossible" 
"exitIfNotForced"
"maybeExit"

cmd/minikube/cmd/start.go
```diff
@@ -1210,3 +1202,10 @@ func getKubernetesVersion(old *config.ClusterConfig) string {
...
+	if !viper.GetBool(force) {
+		exit.WithCodeT(code, message, v)
```
medyagh, line 1208: hint to the user if they really want this, they can use --force flag
similar to what we had before
```
To use this version anyway, use the `--force` flag
```

cmd/minikube/cmd/start.go
```diff
@@ -135,6 +135,9 @@ func runStart(cmd *cobra.Command, args []string) {
 	}
 
 	displayEnviron(os.Environ())
+	if viper.GetBool(force) {
+		out.WarningT("minikube skips various validations when --force is supplied; this may lead to unexpected behavior")
```
medyagh, line 139: add dot at the end.

cmd/minikube/cmd/start.go
priyawadhwa: I don't think we should be encouraging people to just use the `--force` flag if something isn't meeting minimum requirements -- it makes sense for k8s versions because users may be using <1.13 which is a reasonable use case, but not for anything else

cmd/minikube/cmd/start.go
priyawadhwa: I think that would look odd as none of the other logs have periods at the end (sample output is in the PR description)

cmd/minikube/cmd/start.go
priyawadhwa: done

PR #8973, opened 2020-08-11
add unit test for daemon info

medyagh:
Added unit test for docker system info and podman system info
to make sure we never break this functionality across differen docker/podman versions and OSes

### DIFF
--- a/pkg/drivers/kic/oci/info.go
+++ b/pkg/drivers/kic/oci/info.go
@@ -232,11 +232,11 @@ var dockerInfoGetter = func() (string, error) {
 // dockerSystemInfo returns docker system info --format '{{json .}}'
 func dockerSystemInfo() (dockerSysInfo, error) {
 	var ds dockerSysInfo
-	rawJson, err := dockerInfoGetter()
+	rawJSON, err := dockerInfoGetter()
 	if err != nil {
 		return ds, errors.Wrap(err, "docker system info")
 	}
-	if err := json.Unmarshal([]byte(strings.TrimSpace(rawJson)), &ds); err != nil {
+	if err := json.Unmarshal([]byte(strings.TrimSpace(rawJSON)), &ds); err != nil {
 		return ds, errors.Wrapf(err, "unmarshal docker system info")
 	}
 
@@ -251,12 +251,12 @@ var podmanInfoGetter = func() (string, error) {
 // podmanSysInfo returns podman system info --format '{{json .}}'
 func podmanSystemInfo() (podmanSysInfo, error) {
 	var ps podmanSysInfo
-	rawJson, err := podmanInfoGetter()
+	rawJSON, err := podmanInfoGetter()
 	if err != nil {
 		return ps, errors.Wrap(err, "podman system info")
 	}
 
-	if err := json.Unmarshal([]byte(strings.TrimSpace(rawJson)), &ps); err != nil {
+	if err := json.Unmarshal([]byte(strings.TrimSpace(rawJSON)), &ps); err != nil {
 		return ps, errors.Wrapf(err, "unmarshal podman system info")
 	}
 	return ps, nil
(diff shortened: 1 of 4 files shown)


pkg/drivers/kic/oci/info.go
```diff
@@ -225,29 +225,41 @@ type podmanSysInfo struct {
...
+	if err := json.Unmarshal([]byte(strings.TrimSpace(rawJson)), &ds); err != nil {
 		return ds, errors.Wrapf(err, "unmarshal docker system info")
 	}
 
 	return ds, nil
 }
 
 // podmanSysInfo returns podman system info --format '{{json .}}'
-func podmanSystemInfo() (podmanSysInfo, error) {
+func podmanSystemInfo(mockTest ...string) (podmanSysInfo, error) {
```
priyawadhwa, line 249: Instead of adding an argument to many functions, I would create a function that can be overwritten in the unit test itself and return whatever data you want:


```golang
var (
  GetDockerInfoOutput = dockerSystemInfo
)

```

and in the unit test:

```go
GetDockerInfoOutput = func() (dockerSysInfo, error) {
  return rawJson, nil
}
```

pkg/drivers/kic/oci/info_test.go
```diff
@@ -0,0 +1,129 @@
...
...ableArtifactsHostnames":[],"InsecureRegistryCIDRs":["127.0.0.0/8"],"IndexConfigs":{"docker.io":{"Name":"docker.io","Mirrors":[],"Secure":true,"Official":true}},"Mirrors":[]},"NCPU":16,"MemTotal":63336071168,"GenericResources":null,"DockerRootDir":"/var/lib/docker","HttpProxy":"","HttpsProxy":"","NoProxy":"","Name":"image-builder-cloud-shell-v20200811-102837","Labels":[],"ExperimentalBuild":false,"ServerVersion":"18.09.0","ClusterStore":"","ClusterAdvertise":"","Runtimes":{"runc":{"path":"runc"}},"DefaultRuntime":"runc","Swarm":{"NodeID":"","NodeAddr":"","LocalNodeState":"inactive","ControlAvailable":false,"Error":"","RemoteManagers":null},"LiveRestoreEnabled":false,"Isolation":"","InitBinary":"docker-init","ContainerdCommit":{"ID":"7ad184331fa3e55e52b890ea95e65ba581ae3429","Expected":"7ad184331fa3e55e52b890ea95e65ba581ae3429"},"RuncCommit":{"ID":"dc9208a3303feef5b3839f4323d9beb36df0a9dd","Expected":"dc9208a3303feef5b3839f4323d9beb36df0a9dd"},"InitCommit":{"ID":"fec3683","Expected":"fec3683"},"SecurityOptions":["name=seccomp,profile=default"],"ProductLicense":"Community Engine","Warnings":["WARNING: No swap limit support"],"ClientInfo":{"Debug":false,"Plugins":[],"Warnings":null}}`,
```
priyawadhwa, line 18: Consider putting the raw data in a `testdata` directory or as a constant in a different file so that this test is easier to read

pkg/drivers/kic/oci/info.go
medyagh: I like this idea, seems like an interesting mocking way, but I also think this one might be easier to maintain with less complexity. but I could refactor to do that if you think that would add more value. wdyt?

priyawadhwa: I definitely feel this way is cleaner than having to pass in arguments to multiple functions -- that could get complex quickly if you end up having to write more functions

pkg/drivers/kic/oci/info_test.go
medyagh: these ones are meant to be ready machine and I don't think there would be any value in writing it to a file and then reading it back to string (and somethings might go wrong in this process) I don't envision anyone go read this meant-for-machine jsons

we just need to add more and more like these for different versions of docker to make sure they all pass

pkg/drivers/kic/oci/info.go
medyagh: done

PR #8991, opened 2020-08-13
wip: Fix to not create unused file from broken test

### LINKED ISSUE #8614: `make test` leaves behind unused config file: ~/.minikube/profiles/config.json
we need to find out why we generate this usless file, I beleive maybe when there is no profile given we generate a empty config without a folder
~/.minikube/profiles/config.json

while normal profiles are in 
~/.minikube/profiles/profile_name/config.json

anencore94:
- Fix broken test to not create unused config file anymore

Signed-off-by: anencore94 <anencore94@kaist.ac.kr>

fixes #8614

### DIFF
--- a/pkg/minikube/config/profile_test.go
+++ b/pkg/minikube/config/profile_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package config
 
 import (
+	"github.com/spf13/viper"
+
 	"path/filepath"
 	"testing"
 )
@@ -136,7 +138,7 @@ func TestProfileNameInReservedKeywords(t *testing.T) {
 func TestProfileExists(t *testing.T) {
 	miniDir, err := filepath.Abs("./testdata/.minikube2")
 	if err != nil {
-		t.Errorf("error getting dir path for ./testdata/.minikube : %v", err)
+		t.Errorf("error getting dir path for ./testdata/.minikube2 : %v", err)
 	}
 
 	var testCases = []struct {
@@ -163,7 +165,7 @@ func TestProfileExists(t *testing.T) {
 func TestCreateEmptyProfile(t *testing.T) {
 	miniDir, err := filepath.Abs("./testdata/.minikube2")
 	if err != nil {
-		t.Errorf("error getting dir path for ./testdata/.minikube : %v", err)
+		t.Errorf("error getting dir path for ./testdata/.minikube2 : %v", err)
 	}
 
 	var testCases = []struct {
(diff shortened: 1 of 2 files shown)


anencore94:
Thanks for review @medyagh :)

I changed my commit to make it sense than before.
However, I think without changing the actual code somehow, the function `SaveProfile()` will save the converted version config.json to some weird directory..

pkg/minikube/config/profile_test.go
```diff
@@ -277,7 +279,11 @@ func TestGetPrimaryControlPlane(t *testing.T) {
 			t.Fatalf("Failed to load config for %s", tc.description)
 		}
 
-		n, err := PrimaryControlPlane(cc)
+		if tc.description == "old style" {
```
anencore94, line 282: Is it ok to make this line rather than separating the test with each testcase itself?

medyagh: can you explain/comment what does old style mean in this context?

anencore94: `TestGetPrimaryControlPlane` have two tests:
`{"old style", "p1", "192.168.64.75", 8443, "minikube"}, {"new style", "p2_newformat", "192.168.99.136", 8443, "m01"}`

such case load config from `minikube/pkg/minikube/config/testdata/.minikube2/profiles/p1/config.json` and `minikube/pkg/minikube/config/testdata/.minikube2/profiles/p2_newformat/config.json`

But for `p1`, `PrimaryControlPlane` function create "the new version config of p1" save it with same `ProfileName`.
Therefore, the testdata `minikube/pkg/minikube/config/testdata/.minikube2/profiles/p1/config.json` updated with "the new version config of p1".
(I guess this happens since the `PrimaryControlPlane` wants to be compatible with old version(1.6 or earlier) config file where `node` was not implemented.)

So for the "old version" test case, I think we have to create "the new version config of p1" save it with different `ProfileName`.
So I changed the `ProfileName` for "old version". So the test make "the new version config of p1" at `minikube/pkg/minikube/config/testdata/.minikube2/profiles/p1_converted/config.json` and delete the file when test finished.

pkg/minikube/config/profile_test.go
```diff
@@ -277,7 +279,11 @@ func TestGetPrimaryControlPlane(t *testing.T) {
...
+			viper.Set(ProfileName, tc.profile+"_converted")
```
tstromberg, line 283: I don't think this is correct. The profile name should be `tc.profile`. If we need things to be different for test data, we need to move the test data.

pkg/minikube/config/profile_test.go
tstromberg: I don't think the old style should require any special handling: I think the intent of this test is to assert that.

medyagh:
@antoineco are you planning on addressing the feedback ? I think the PR is doing the right thing except in the unit test part

antoineco:
@medyagh  You meant "anencore94" :)

anencore94:
@medyagh @tstromberg Thanks for reminding me :)
But I couldn't figure out how to fix unit test is right thing until now..
So if anyone wants to take on this issue, please do so with #8614 !

PR #9000, opened 2020-08-14
WIP: Add new external-dns addon for exposing ingress & LoadBalancer external IPs to the host

### LINKED ISSUE #8980: Add addon for external DNS
Add an addon for installing external DNS. See https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/coredns.md which uses CoreDNS as the DNS server.

Unlike ingress-dns, which is minikube specific:
1. This is sometimes used in production, so will give similar behavior in development to production.
2. Will support LoadBalancer services.

Some notes about configuring the host that also apply for ingress-dns:
1. For Linux distros using `systemd-resolved`, you can use [`systemd.network` units](https://www.freedesktop.org/software/systemd/man/systemd.network.html) for configuring DNS domain specific DNS servers. Using the `Domains` & `DNS` keys in a new network unit matching the required interface. I used it before, but don't remember the details.
2. For Windows, there is NRPT which should allow setting this.

segevfiner:
An initial version of an addon for
[external-dns](https://github.com/kubernetes-sigs/external-dns). Based on [ExternalDNS for CoreDNS with minikube](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/coredns.md). Cobbled together from that tutorial, upstream helm charts, and the way minikube/kubeadm deploy etcd & CoreDNS. Those manifests probably
need some reviewing and tweaking (See TODO below)

Add an addon for installing external DNS. See https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/coredns.md which uses CoreDNS as the DNS server.

Unlike ingress-dns, which is minikube specific:
1. This is sometimes used in production, so will give similar behavior in development to production.
2. Will support LoadBalancer services.

Some notes about configuring the host that also apply for ingress-dns:
1. For Linux distros using `systemd-resolved`, you can use [`systemd.network` units](https://www.freedesktop.org/software/systemd/man/systemd.network.html) for configuring DNS domain specific DNS servers. Using the `Domains` & `DNS` keys in a new network unit matching the required interface. I used it before, but don't remember the details.
2. For Windows, there is NRPT which should allow setting this.

At this point it is still WIP/incomplete, I will need some help from someone to finish this according to the TODO list below and any other feedback.

### TODO
- [ ] Which images to use for etcd and CoreDNS? The k8s.io ones minikube uses? or upstream ones (From docs or helm charts)? Using minikube's images means it can use the preloaded ones as long as it is kept up to date, at least for the latest Kubernetes version, though updating those is likely to be forgotten...
- [ ] `readinessProbe`/`livenessProbe`, what times, thresholds should be used, are the scripts, endpoints correct, etc. (I kinda copied them from somewhere and they are not necessarily optimal)
  - [ ] The `readinessProbe` for etcd breaks it because the headless service doesn't register it until it is ready but it queries for it to become ready, At first I tried just disabling it, but that also didn't always work, not sure why, so I added `publishNotReadyAddresses` to the service, which might be correct anyhow, though not sure if I should then add a separate service that does not publish unready addresses, though it is a single node deployment so might not matter much either way.
- [ ] Do any pods here need a `priorityClass`?
- [ ] Do any pods here need `tolerations`?
- [ ] Set `resources` correctly.
- [ ] Use `hostPath` instead of PVC for etcd? or maybe make it ephemeral and not use a volume at all?
- [ ] labels for the objects, etc. Are there any labels add-ons normally use? Any label schema this should use instead of just `app`?
- [ ] I used the headless service to access etcd, not sure that's the best/correct way, some etcd deployments seem to also add `ClusterIP` service and use that instead, might not matter much for a single node.
- [ ] The CoreDNS server currently forwards DNS to the in cluster DNS which will also answer for `cluster.local`, which is not desired on the host, had to explicitly `except` it. Not sure if that's the best config. I can't just `dnsPolicy: Default` like the in-cluster CoreDNS because the CoreDNS server needs to resolve the etcd server address.
- [ ] The CoreDNS server is not currently exposed to the host. I think we should use `hostPort` to expose it. This also requires `hostIP` due to having another DNS server listening on localhost on the `minikube` VM, this needs to be set automatically but I'm not sure how to add such a template parameter, as in: where to extract it from inside minikube. Just using `hostNetwork` will be problematic with accessing etcd.
- [ ] Add `README.md` explaining how to configure host, etc. Like `ingress-dns` has. Maybe also an example, though upstream examples using the appropriate domain should work too.
- [ ] The domain is currently hard-coded to `test`, we might want to make that configurable.
- [ ] Will likely be effected by whatever issue plagues `ingress` & `ingress-dns` on Docker driver, so might need the same provisions and documentation as them.

fixes #8980

### DIFF
--- a/deploy/addons/external-dns/extdns-coredns-configmap.yaml
+++ b/deploy/addons/external-dns/extdns-coredns-configmap.yaml
@@ -1,0 +1,29 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: extdns-coredns
+  namespace: kube-system
+  labels:
+    app: extdns-coredns
+data:
+  Corefile: |-
+    .:53 {
+        errors
+        health {
+            lameduck 5s
+        }
+        ready
+        etcd test {
+            stubzones
+            path /skydns
+            endpoint http://extdns-etcd:2379
+        }
+        prometheus 0.0.0.0:9153
+        forward . /etc/resolv.conf {
+          except cluster.local
+        }
+        cache 30
+        loop
+        reload
+        loadbalance
+    }
--- a/deploy/addons/external-dns/extdns-coredns-deploy.yaml
+++ b/deploy/addons/external-dns/extdns-coredns-deploy.yaml
@@ -1,0 +1,73 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: extdns-coredns
+  namespace: kube-system
+  labels:
+    app: extdns-coredns
+spec:
+  selector:
+    matchLabels:
+      app: extdns-coredns
+  template:
+    metadata:
+      labels:
+        app: extdns-coredns
+    spec:
+      containers:
+      - name: coredns
+        image: k8s.gcr.io/coredns:1.6.7
+        args:
+        - -conf
+        - /etc/coredns/Corefile
+        volumeMounts:
+        - mountPath: /etc/coredns
+          name: config-volume
+        ports:
+        - containerPort: 53
+          name: dns
+          protocol: UDP
+          # hostIP: {{.NodeIP}} ???
+          # hostPort: 53
+        - containerPort: 53
+          name: dns-tcp
+          protocol: TCP
+          # hostIP: {{.NodeIP}} ???
+          # hostPort: 53
+        - containerPort: 9153
+          name: metrics
+          protocol: TCP
+        resources:
+          limits:
+            cpu: 100m
+            memory: 128Mi
+          requests:
+            cpu: 100m
+            memory: 128Mi
+        livenessProbe:
+          httpGet:
+            path: /health
+            port: 8080
+            scheme: HTTP
+          initialDelaySeconds: 60
+          periodSeconds: 10
+          failureThreshold: 5
+          successThreshold: 1
+          timeoutSeconds: 5
+        readinessProbe:
+          httpGet:
+            path: /ready
+            port: 8181
+            scheme: HTTP
+          periodSeconds: 10
+          failureThreshold: 3
+          successThreshold: 1
+          timeoutSeconds: 1
+      volumes:
+      - configMap:
+          defaultMode: 420
+          items:
+          - key: Corefile
+            path: Corefile
+          name: extdns-coredns
+        name: config-volume
--- a/deploy/addons/external-dns/extdns-etcd-statefulset.yaml
+++ b/deploy/addons/external-dns/extdns-etcd-statefulset.yaml
@@ -1,0 +1,93 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: extdns-etcd
+  namespace: kube-system
+  labels:
+    app: extdns-etcd
+spec:
+  selector:
+    app: extdns-etcd
+  clusterIP: None
+  ports:
+  - name: etcd-client
+    port: 2379
+  - name: etcd-server
+    port: 2380
+  publishNotReadyAddresses: true
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: extdns-etcd
+  namespace: kube-system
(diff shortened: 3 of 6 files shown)


medyagh:
hi @segevfiner  this seems to be an interesting addon, do you mind pulling upstream and putting an example output in the description how to use this addon?

medyagh (changes requested):
if you please put steps in the description how to use this addon with example output this would be a good addon, if you can also document if it works with all drivers or only VM drivers?

segevfiner:
To make this operational, we still need to somehow set the `hostIP` for exposing the DNS server to the host.

> The CoreDNS server is not currently exposed to the host. I think we should use hostPort to expose it. This also requires hostIP due to having another DNS server listening on localhost on the minikube VM, this needs to be set automatically but I'm not sure how to add such a template parameter, as in: where to extract it from inside minikube. Just using hostNetwork will be problematic with accessing etcd.

But I wasn't sure how to get the addon infrastructure of minikube render that into a templated deployment spec.

The rest of the stuff in the TODO list I left in the PR description are about polishing this as far as I remember.

For trying it out:
```
kubectl create deployment nginx --image nginx
kubectl expose deployment nginx --port 80 --type LoadBalancer
kubectl annotate service nginx "external-dns.alpha.kubernetes.io/hostname=nginx.test"
```

And you should be able to query the created DNS server for it, but we need to fix the expose to host thing so you can do so from the host.

See for my initial POC of this https://github.com/kubernetes/minikube/issues/8980#issuecomment-673708670 for more information.

I only tested this with VM drivers, it likely suffers from whatever issue that prevent the ingress addon from working, that I don't fully understand, on the docker driver, as this too tries to expose a port via the host machine. Though it tries to do so differently using `hostPort`/`hostIP` so it can still communicate with the deployed etcdz rather than `hostNetwork` , once we get the `hostPort`/`hostIP` thing sorted, we can try to see if that works or not.

medyagh:
hi @segevfiner thanks for this PR, do you mind adding a tutorial of using this new addon with examples
the documentation for addons can be added to https://minikube.sigs.k8s.io/docs/handbook/addons/

segevfiner:
I will once we can get it fully working, I need help with the last issue I described here: https://github.com/kubernetes/minikube/pull/9000#issuecomment-711287826

The rest of the things mentioned in the PR description are about polishing it, stuff that I wasn't sure about and only someone more familiar with this project might be able to help with.

bl-ue:
@segevfiner Keep up the great work! 👍 This is a fantastic idea. It's also PR 9,000 :)

tstromberg:
is this ready to merge?

segevfiner:
See the PR description for the list of remaining tasks. The major ones being figuring out how to handle `hostIP` so the dns server is exposed, minikube needs to set the `hostIP` in the dns server pod template, and I couldn't figure out where to get it from, and documentation (Which I'm not sure where to write, it is similar to the existing ingress-dns in terms of host setup but I can add some more details on to of what's already there)

spowelljr:
@segevfiner Is this still something you're working on?

spowelljr:
@segevfiner I'm going to close this PR due to long duration without activity, but feel free to reopen if you have time to work on  this again.

bl-ue:
(off-topic: @spowelljr you must have had a reminder or something, didn't you?  Because you close this PR almost exactly 1 week after you last commented, off by only +3 minutes!)

spowelljr:
@bl-ue That's a really good observation! But actually there's no reminder, every Wednesday 11-12 PST minikube maintainers do through PRs and issues to comment/label/close. The 3 minute difference is somewhat of a coincidence, but would have been within an hour if that makes sense.

bl-ue:
Yeah that does make sense.  Thank you!

PR #9043, opened 2020-08-20
Lock debug

medyagh:


pkg/minikube/machine/start.go
```diff
@@ -265,10 +265,14 @@ func postStartSetup(h *host.Host, mc config.ClusterConfig) error {
 }
 
 // acquireMachinesLock protects against code that is not parallel-safe (libmachine, cert setup)
-func acquireMachinesLock(name string) (mutex.Releaser, error) {
+func acquireMachinesLock(name string, drver string) (mutex.Releaser, error) {
 	spec := lock.PathMutexSpec(filepath.Join(localpath.MiniPath(), "machines"))
 	// NOTE: Provisioning generally completes within 60 seconds
 	spec.Timeout = 15 * time.Minute
+	if driver.IsKIC(drver) {
+		spec.Timeout = 20 * time.Second
```
tstromberg, line 273: I expect that this will cause a great deal of integration testing instability.

My suggestion would be to change this lock for KIC so that the `spec` includes the machine name at the end. As best as I can tell, Docker does not have any problems with creating multiple containers simultaneously.

With that change, 120 seconds is probably fine for Docker. Without it, it will mean unnecessary crashing in integration tests, as `acquireMachinesLock` shares locks across machines.

pkg/minikube/machine/start.go
```diff
@@ -70,15 +70,15 @@ func StartHost(api libmachine.API, cfg *config.ClusterConfig, n *config.Node) (*
...
-		return nil, false, errors.Wrap(err, "boot lock")
-	}
+	releaser, err := acquireMachinesLock(machineName, cfg.Driver)
 	start := time.Now()
 	defer func() {
 		glog.Infof("releasing machines lock for %q, held for %s", machineName, time.Since(start))
 		releaser.Release()
 	}()
+	if err != nil {
+		return nil, false, errors.Wrap(err, "boot lock")
```
tstromberg, line 80: Can you confirm this behaves as intended? It's unusual to release a lock on error, because you don't know if the lock was successfully acquired. It may cause a panic.

tstromberg:
PR needs a better title.

pkg/minikube/machine/start.go
medyagh: I added a recover() to take care of that

pkg/minikube/driver/install.go
```diff
@@ -48,10 +48,11 @@ func InstallOrUpdate(name string, directory string, v semver.Version, interactiv
 	spec.Timeout = 10 * time.Minute
 	glog.Infof("acquiring lock: %+v", spec)
 	releaser, err := mutex.Acquire(spec)
+	start := time.Now()
+	defer lock.Release(releaser, "driverInstall", start)
```
tstromberg, line 52: I don't think this is correct. One should never attempt to release locks if they fail to be set.

pkg/minikube/machine/start.go
```diff
@@ -265,10 +262,15 @@ func postStartSetup(h *host.Host, mc config.ClusterConfig) error {
 }
 
 // acquireMachinesLock protects against code that is not parallel-safe (libmachine, cert setup)
-func acquireMachinesLock(name string) (mutex.Releaser, error) {
+func acquireMachinesLock(name string, drver string) (mutex.Releaser, error) {
 	spec := lock.PathMutexSpec(filepath.Join(localpath.MiniPath(), "machines"))
 	// NOTE: Provisioning generally completes within 60 seconds
-	spec.Timeout = 15 * time.Minute
+	// however in para;lel integration testing it might take longer
```
tstromberg, line 268: typo

pkg/minikube/machine/start.go
```diff
@@ -265,10 +262,15 @@ func postStartSetup(h *host.Host, mc config.ClusterConfig) error {
 }
 
 // acquireMachinesLock protects against code that is not parallel-safe (libmachine, cert setup)
-func acquireMachinesLock(name string) (mutex.Releaser, error) {
+func acquireMachinesLock(name string, drver string) (mutex.Releaser, error) {
```
tstromberg, line 265: perhaps just `drv` as a name?

pkg/minikube/machine/start.go
```diff
@@ -279,6 +281,16 @@ func acquireMachinesLock(name string) (mutex.Releaser, error) {
 	return r, err
 }
 
+func releaseMachineLock(releaser mutex.Releaser, name string, start time.Time) {
+	defer func() {
+		if r := recover(); r != nil {
```
tstromberg, line 286: Please do not call recover() outside of the `main` package. I don't think calling recover() here is the behavior we are looking for. If so, I would rather see the system redesigned.

PR #9044, opened 2020-08-20
add an new flag to stop command --keep-context-active

medyagh:


tstromberg:
Can you describe the use case for this more? I want to make sure this isn't just technical debt.

If we really do need this, I would want an integration test for it.

PR #9064, opened 2020-08-22
Update TestSkaffold to use unique profile names, more debuggable

tstromberg:
While investigating a flake - I made some adjustments to TestSkaffold:

Here was the flake:

https://storage.googleapis.com/minikube-builds/logs/9050/f5c3a82/Docker_Linux.html#fail_TestSkaffold

Three things I noticed while debugging:

* The `registry-server`OOM, but this test doesn't enable the registry
* Docker cert mismatches
* The test failed locally because it copied `minikube` to `minikube`

The first two issues ppear to be related to tests reusing the same profile name.

 I've updated this test to use the secret `--minikube-profile` flag to Skaffold to workaround the hard-coded "minikube" cluster requirement. In the next Skaffold release, we should be able to drop this flag.

### DIFF
--- a/test/integration/skaffold_test.go
+++ b/test/integration/skaffold_test.go
@@ -68,15 +68,6 @@ func TestSkaffold(t *testing.T) {
 		t.Fatalf("unable to determine abs path: %v", err)
 	}
 
-	oldPath := os.Getenv("PATH")
-	os.Setenv("PATH", fmt.Sprintf("%s:%s", filepath.Dir(abs), os.Getenv("PATH")))
-	defer func() {
-		os.Setenv("PATH", oldPath)
-		t.Logf("PATH is now %s", os.Getenv("PATH"))
-	}()
-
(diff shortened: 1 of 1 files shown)


test/integration/skaffold_test.go
```diff
@@ -51,32 +50,51 @@ func TestSkaffold(t *testing.T) {
...
+	oldPath := os.Getenv("PATH")
 	os.Setenv("PATH", fmt.Sprintf("%s:%s", filepath.Dir(abs), os.Getenv("PATH")))
+	defer func() {
+		os.Setenv("PATH", oldPath)
+		t.Logf("PATH is now %s", os.Getenv("PATH"))
+	}()
+
+	t.Logf("PATH is now %s", os.Getenv("PATH"))
+
+	if filepath.Base(Target()) != "minikube" {
```
sharifelgamal, line 80: Should we set the PATH env var after this part?

tstromberg: Done.

PR #9066, opened 2020-08-23
Add install minikube with homebrew on Linux

inductor:
Homebrew has supported Linux since a while ago and I've been using it on my Linux environment too :)

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -49,8 +49,6 @@ sudo rpm -ivh minikube-latest.x86_64.rpm
 
 ### Homebrew package
 
-If [Homebrew](https://brew.sh/) installed:
-
 ```shell
 brew install minikube
 ```
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -26,21 +26,6 @@ All you need is Docker (or similarly compatible) container or a Virtual Machine 
 
 For Linux users, we provide 4 easy download options:
 
-### Homebrew package
-
-If [Homebrew](https://brew.sh/) installed:
-
-```shell
-brew install minikube
-```
-
-If `which minikube` fails after installation via brew, you may have to remove the minikube cask and link the binary:
-
-```
-brew cask remove minikube
-brew link minikube
-```
-
 ### Binary download
 
 ```shell
@@ -60,6 +45,21 @@ sudo dpkg -i minikube_latest_amd64.deb
 ```shell
 curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-latest.x86_64.rpm
 sudo rpm -ivh minikube-latest.x86_64.rpm
+```
+
+### Homebrew package
+
+If [Homebrew](https://brew.sh/) installed:
+
+```shell
+brew install minikube
+```
+
+If `which minikube` fails after installation via brew, you may have to remove the minikube cask and link the binary:
+
+```
+brew cask remove minikube
+brew link minikube
 ```
 
 {{% /tab %}}
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -6,7 +6,7 @@ aliases:
   - /docs/start
 ---
 
-minikube is local Kubernetes, focusing on making it easy to learn and develop for Kubernetes. 
+minikube is local Kubernetes, focusing on making it easy to learn and develop for Kubernetes.
 
 All you need is Docker (or similarly compatible) container or a Virtual Machine environment, and Kubernetes is a single command away: `minikube start`
 
@@ -24,7 +24,22 @@ All you need is Docker (or similarly compatible) container or a Virtual Machine 
 {{% tabs %}}
 {{% tab "Linux" %}}
 
-For Linux users, we provide 3 easy download options:
+For Linux users, we provide 4 easy download options:
+
+### Homebrew package
+
+If [Homebrew](https://brew.sh/) installed:
+
+```shell
+brew install minikube
+```
+
+If `which minikube` fails after installation via brew, you may have to remove the minikube cask and link the binary:
+
+```
+brew cask remove minikube
+brew link minikube
+```
 
 ### Binary download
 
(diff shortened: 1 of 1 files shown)


inductor:
@medyagh

>if you think adding brew would add a lot of value, I recommend at least adding it in the end and with same format as others

Sure, but I do not think the format is h2(with ##) and `Homebrew install`. I followed other formats as follows

```markdown
### Debian package
### RPM package
```

site/content/en/docs/start/_index.md
```diff
@@ -47,10 +47,23 @@ curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-latest
 sudo rpm -ivh minikube-latest.x86_64.rpm
 ```
 
+### Homebrew package
+
+```shell
+brew install minikube
+```
+
+If `which minikube` fails after installation via brew, you may have to remove the minikube cask and link the binary:
```
medyagh, line 56: I rather we keep this first page short, about creating a full page for all the possible installation for linux, and then from here

that says
for hombrew and full notes see ...(link)

afbjorklund:
Worked fine here, defintely better than the other mess... (the binary, and deb and rpm "hacks")
With their `sudo` and multiple commands and whatever. I'll throw in [0install](https://0install.net/) option too.

Binary download

```
 curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
 sudo install minikube-linux-amd64 /usr/local/bin/minikube
```

Debian package

```
curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube_latest_amd64.deb
sudo dpkg -i minikube_latest_amd64.deb
```

RPM package

```
curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-latest.x86_64.rpm
sudo rpm -ivh minikube-latest.x86_64.rpm
```

Linux Homebrew

```
brew install minikube
```

Zero Install

```
alias minikube='0install run https://apps.0install.net/kubernetes/minikube.xml'
```

inductor:
>Worked fine here, defintely better than the other mess... (the binary, and deb and rpm "hacks")

Yeah, that's actually the biggest motivation for this Pull Request!

afbjorklund:
However, the sheer amount of text / different environments is also a problem with the install page.
The first three all deliver the very same "official" binary, which is why they are prominently featured.

It is nowhere as bad as the podman/cri-o pages, with dozens and dozens of distributions, but still...
Another approach would be to use the tabs for the package managers, like is being done here:

https://kubernetes.io/docs/tasks/tools/install-kubectl/

You know, the alternative to `brew install kubectl`. 😃

afbjorklund:
@inductor : actually we **do** recommend using `brew`:

https://kubernetes.io/docs/tasks/tools/install-minikube/

inductor:
@afbjorklund @medyagh 

I am actually from SIG-Docs 😅 
So regarding all these, what should the doc be?

afbjorklund:
>  So regarding all these, what should the doc describe?

I think the original idea (add brew) was OK, just formatting...

* Documentation
  * Get Started!
    * What you’ll need
        1) Installation
           * Linux
              * Binary download
              * RPM package
              * DEB package
              * **Brew Package Manager**
           * macOS
              * Brew Package Manager
              * Binary download
           * Windows
              * Chocolatey Package Manager
              * Binary download
         2) Start your cluster

So it is mostly about fiddling the order around, and improving wording.

It is also somewhat complicated, compared to a big Download button...

afbjorklund:
The comment from @medyagh was that adding brew would make the Linux list too long.

Maybe nested tabs would fix, but I'm not sure about that approach either - if it even works.

inductor:
>Maybe nested tabs would fix,

I think it's already nested

afbjorklund:
> I think it's already nested

I can only see one layer, but nested tabs is probably evil (and I'm not sure if Hugo allows it)

minikube-installation

Theoretically one _could_ put DEB and RPM tabs, like is being done for kubeadm now.

kubeadm-installation

But then there would be a distro tab inside an operating system tab, and it would be ugly.

And of course trying to get apt and yum working is like a whole **class** of fail on its own...
It is working for kubectl and kubeadm, but minikube is not allowed in the same repository.

We could of course "untab" the OS, but then the page would be longer rather than shorter ?

afbjorklund:
@priyawadhwa : we don't usually have priorities on PRs, so if we don't want to merge it as-is we need a backlog issue (for brew)

@medyagh : maybe we can discuss it on the next meeting, if you think that the Linux install page is indeed becoming too long ?

minikube-linux-install-brew

At the very least, we should probably work with @inductor and others to make the two install pages a bit more similar than now:

* https://kubernetes.io/docs/tasks/tools/install-minikube/

* https://minikube.sigs.k8s.io/docs/start/

medyagh (changes requested):
I am okay with adding home brew as an option but i would like to know first, who Maintains this minikube homebrew for linux ?

is it an automatic process to ensure we have brew for linux ? does that mean whenever we update the mac homebrew they automatically add linux version too ? till I dont know the details I am not comfortable merging this
and also we dont need the extra lines that says if it didnt work do this or that. first page should be as short as possible. u can have a Link to another page that says for more info see... (homebrew links)

afbjorklund:
> I am okay with adding home brew as an option but i would like to know first, who Maintains this minikube homebrew for linux ?

It is the same package:

https://github.com/Homebrew/homebrew-core/commits/master/Formula/minikube.rb

https://github.com/Homebrew/linuxbrew-core/commits/master/Formula/minikube.rb

Basically only header changes:

```ruby
class Minikube < Formula
  desc "Run a Kubernetes cluster locally"
  homepage "https://minikube.sigs.k8s.io/"
  url "https://github.com/kubernetes/minikube.git",
      tag:      "v1.14.0",
      revision: "b09ee50ec047410326a85435f4d99026f9c4f5c4"
  license "Apache-2.0"
  head "https://github.com/kubernetes/minikube.git"

...
```

And then the build system builds the binaries ("bottles") from the source ("formulas")

So homebrew is not using the same binaries as available on the github download page.

They call that "[cask](https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap#casks)", when speaking in :beer:

afbjorklund:
We close this for now, Brew and Arch are still available - even if undocumented "upstream".

https://formulae.brew.sh/formula/minikube

https://www.archlinux.org/packages/community/x86_64/minikube/

Thanks for your contribution! Maybe a future version of Getting Started will feature them...

PR #9029, opened 2020-08-18
Dockerfile: prioritize /etc/hosts over dns

tstromberg:
Some DNS servers issue fraudulent A/AAA responses for hosts which return NXDOMAIN:

https://www.verizon.com/support/residential/internet/home-network/settings/opt-out-of-dns-assist

This unfortunately breaks the ability for minikube nodes to discover the location of the apiserver. This change modifies the system to prioritize /etc/hosts first, and is propagated out to each pod.

Fixes #9051

PR #9071, opened 2020-08-24
bump kic base image

medyagh:
Pushed new kic image to include https://github.com/kubernetes/minikube/pull/9029

This PR should be converted to an Auto PR by minikube-bot

### DIFF
--- a/pkg/drivers/kic/types.go
+++ b/pkg/drivers/kic/types.go
@@ -24,9 +24,9 @@ import (
 
 const (
 	// Version is the current version of kic
-	Version = "v0.0.12-snapshot"
+	Version = "v0.0.12-snapshot2"
 	// SHA of the kic base image
-	baseImageSHA = "7be40a42fdfec56fbf7bc9de07ea2ed4a931cbb70dccb8612b2ba13763bf4568"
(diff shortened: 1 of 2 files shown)


pkg/drivers/kic/types.go
```diff
@@ -24,9 +24,9 @@ import (
 
 const (
 	// Version is the current version of kic
-	Version = "v0.0.12-snapshot"
+	Version = "v0.0.12-snapshot2"
```
tstromberg, line 27: What's with the snapshot name? Why not versions?

medyagh: This is before we release.

medyagh: so basically snapshot means current HEAD changes that has not been released to people yet
we can potentially use a different repo too but this pattern seems to straighforward

medyagh: the idea is to one day automate this process using this pattern https://github.com/kubernetes/minikube/issues/9072

afbjorklund: Probably need some additional identifier too, to distinguish between different images.
Could either use commit (from "git describe") or a date (like KIND is doing) perhaps ?

v1.12.3-153-g5dc6c83
v20200817-222e7545

Issue #9087, opened 2020-08-25
docker: minikube delete should delete the network


medyagh:
I notice minikube delete doesn't remoev the network after itself. created this issue.

```
medya@medya:~/workspace/minikube (master)$ ./out/minikube delete --all

medya@medya:~/workspace/minikube (master)$ docker network ls
NETWORK ID          NAME                DRIVER              SCOPE
fdc833b38ce5        bridge              bridge              local
43cf4c7ad471        host                host                local
0e6c505503f2        minikube            bridge              local
2d7623149fc1        none                null                local
```

Issue #9069, opened 2020-08-24
docker network create : Pool overlaps with other one on this address space


medyagh:
as seen here https://storage.googleapis.com/minikube-builds/logs/9065/8bc8325/Docker_Linux.html#fail_TestOffline%2fgroup%2fcontainerd

```
	I0822 18:04:37.920257  332897 cli_runner.go:109] Run: docker network create --driver=bridge --subnet=192.168.39.0/24 --gateway 192.168.39.1 offline-containerd-20200822180426-331183
	W0822 18:04:37.975776  332897 kic.go:85] unable to create docker network; node ip may not be stable: error creating network: docker network create --driver=bridge --subnet=192.168.39.0/24 --gateway 192.168.39.1 offline-containerd-20200822180426-331183: exit status 1
	stdout:
	
	stderr:
	Error response from daemon: Pool overlaps with other one on this address space
```

medyagh:
cc: @josedonizetti

PR #9094, opened 2020-08-26
wip: docker: fix network static

medyagh:
### After this PR
- Create one network per Cluster
- Assign Static IP based on the machine order (offset from gateway)
- if can't create network becaue subnet is taken, bump the subnet 3rd digit by 10, and try again.
- fix left over docker networks not being deleted with minikube delete --all
- add more error types
- add failfast error type for not trying again with same IP, when error is IP is in use.
- add created_by_minikube=true label to the networks we create for better clean up. and avoid deleting user's networks.

closes https://github.com/kubernetes/minikube/issues/9087 
closes https://github.com/kubernetes/minikube/issues/9069
closes https://github.com/kubernetes/minikube/issues/8936

### TODO: fix upgrading existing cluster
```
medya@medya:~/workspace/minikube$ ./out/minikube start 
😄  minikube v1.12.3 on Debian rodete
E0826 11:57:18.506400 2936672 start.go:97] Unable to get host IP: network inspect: kic network not found

💣  failed to start node: startup failed: Failed to setup kubeconfig: network inspect: kic network not found

```

### DIFF
--- a/cmd/minikube/cmd/mount.go
+++ b/cmd/minikube/cmd/mount.go
@@ -117,7 +117,6 @@ var mountCmd = &cobra.Command{
 			}
 		}
 		port, err := getPort()
-		fmt.Println("port is ", ip)
 		if err != nil {
 			exit.WithError("Error finding port for mount", err)
 		}
@@ -148,7 +147,6 @@ var mountCmd = &cobra.Command{
 		}
 
 		bindIP := ip.String() // the ip to listen on the user's host machine
-		fmt.Println("bind ip is", bindIP)
 		if driver.IsKIC(co.CP.Host.Driver.DriverName()) && runtime.GOOS != "linux" {
 			bindIP = "127.0.0.1"
 		}
--- a/cmd/minikube/cmd/mount.go
+++ b/cmd/minikube/cmd/mount.go
@@ -106,7 +106,7 @@ var mountCmd = &cobra.Command{
 		var ip net.IP
 		var err error
 		if mountIP == "" {
-			ip, err = cluster.HostIP(co.CP.Host)
+			ip, err = cluster.HostIP(co.CP.Host, co.Config.Name)
 			if err != nil {
 				exit.WithError("Error getting the host IP address to use from within the VM", err)
 			}
--- a/cmd/minikube/cmd/mount.go
+++ b/cmd/minikube/cmd/mount.go
@@ -117,6 +117,7 @@ var mountCmd = &cobra.Command{
 			}
 		}
 		port, err := getPort()
+		fmt.Println("port is ", ip)
 		if err != nil {
 			exit.WithError("Error finding port for mount", err)
 		}
@@ -147,6 +148,7 @@ var mountCmd = &cobra.Command{
 		}
 
 		bindIP := ip.String() // the ip to listen on the user's host machine
+		fmt.Println("bind ip is", bindIP)
 		if driver.IsKIC(co.CP.Host.Driver.DriverName()) && runtime.GOOS != "linux" {
 			bindIP = "127.0.0.1"
 		}
--- a/pkg/drivers/kic/oci/network.go
+++ b/pkg/drivers/kic/oci/network.go
@@ -32,10 +32,10 @@ import (
 
 // RoutableHostIPFromInside returns the ip/dns of the host that container lives on
 // is routable from inside the container
-func RoutableHostIPFromInside(ociBin string, containerName string) (net.IP, error) {
+func RoutableHostIPFromInside(ociBin string, clusterName string, containerName string) (net.IP, error) {
 	if ociBin == Docker {
 		if runtime.GOOS == "linux" {
-			_, gateway, err := dockerNetworkInspect(containerName)
+			_, gateway, err := dockerNetworkInspect(clusterName)
 			if err != nil {
 				return gateway, errors.Wrap(err, "network inspect")
 			}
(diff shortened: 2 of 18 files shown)


cmd/minikube/cmd/mount.go
```diff
@@ -117,6 +117,7 @@ var mountCmd = &cobra.Command{
 			}
 		}
 		port, err := getPort()
+		fmt.Println("port is ", ip)
```
tstromberg, line 120: glog.Infof (here and elsewhere)

pkg/drivers/kic/oci/network.go
```diff
@@ -33,14 +35,18 @@ import (
 func RoutableHostIPFromInside(ociBin string, containerName string) (net.IP, error) {
 	if ociBin == Docker {
 		if runtime.GOOS == "linux" {
-			return dockerGatewayIP(containerName)
+			_, gateway, err := dockerNetworkInspect(containerName)
+			if err != nil {
+				return gateway, errors.Wrap(err, "routable gateway")
```
tstromberg, line 40: This error annotation should be something like "network inspect". The caller to `RoutableHostIPFromInside` will probably add something like `routable host`.

pkg/drivers/kic/oci/network.go
```diff
@@ -168,42 +149,152 @@ func dockerContainerIP(name string) (string, string, error) {
...
+	_, subnet, err = net.ParseCIDR(defaultSubnet)
 	if err != nil {
-		return errors.Wrapf(err, "error creating network")
+		return nil, errors.Wrapf(err, "parse default subnet %s", defaultSubnet)
+	}
+
+	err = attemptCreateNework(subnet, name)
+	if err != nil {
+		if err == ErrNetworkSubnetTaken {
+			// try up tp 13 times
```
tstromberg, line 170: Please comment the "why", not the "what". Why retry, and why this many times?

pkg/drivers/kic/oci/network.go
```diff
@@ -168,42 +149,152 @@ func dockerContainerIP(name string) (string, string, error) {
...
+	attempt := 0
+	_, subnet, err = net.ParseCIDR(defaultSubnet)
 	if err != nil {
-		return errors.Wrapf(err, "error creating network")
+		return nil, errors.Wrapf(err, "parse default subnet %s", defaultSubnet)
+	}
+
+	err = attemptCreateNework(subnet, name)
+	if err != nil {
+		if err == ErrNetworkSubnetTaken {
```
tstromberg, line 169: exit early if err != ErrNetworkSubnetTaken, so that the longer case has the minimum indentation.

pkg/drivers/kic/oci/network.go
tstromberg, line 170: spelling

pkg/drivers/kic/oci/network.go
```diff
@@ -168,42 +149,152 @@ func dockerContainerIP(name string) (string, string, error) {
...
+			// try up tp 13 times
+			for attempt < 13 {
+				attempt++
+				glog.Infof("Couldn't create network %q at %q subnet will try again a new subnet ...", name, subnet)
```
tstromberg, line 173: "with a new subnet"

pkg/drivers/kic/oci/network.go
```diff
@@ -168,42 +149,152 @@ func dockerContainerIP(name string) (string, string, error) {
...
+				// increase 2nd digit by 10 each time
```
tstromberg, line 174: Document why the 2nd digit is special. 

Also, this appears to increment the 3rd octet, rather than the 2nd digit.

pkg/drivers/kic/oci/network.go
```diff
@@ -168,42 +149,152 @@ func dockerContainerIP(name string) (string, string, error) {
...
+	return subnet, nil
+}
 
+func attemptCreateNework(subnet *net.IPNet, name string) error {
+	gateway := subnet.IP.To4()
+	gateway[3]++ // first ip for gateway
+	glog.Infof("attempt to create network %q with subnet: %s and gateway %s...", subnet, name, gateway)
+	rr, err := runCmd(exec.Command(Docker, "network", "create", "--driver=bridge", fmt.Sprintf("--subnet=%s", subnet), fmt.Sprintf("--gateway=%s", gateway), "-o", "com.docker.network.bridge.enable_ip_masquerade=true", fmt.Sprintf("--label=%s=%s", CreatedByLabelKey, "true"), name))
+	if err != nil {
+		if strings.Contains(rr.Output(), "Pool overlaps with other one on this address space") {
```
tstromberg, line 197: Ouch. Is it possible that Docker has exit codes we can use instead? I guarantee this will break some day when someone fixes the bad grammar usage in this error message.

pkg/minikube/machine/delete.go
```diff
@@ -40,23 +40,30 @@ func deleteOrphanedKIC(ociBin string, name string) {
 		return
 	}
 
+	defer func() { // networks should be deleted after containers
```
tstromberg, line 43: why the defer? We have no idea at this point if we were successfully able to delete the container.

medyagh: beause in the code bellow somwhere we return if container doesnt exist. so I use defer.
and also you can not delete network if container is attached to it

pkg/drivers/kic/oci/network.go
medyagh: this is first iteration if this POC proves that this helps, we can invest in doing something fancy.
Rob Pike:
brute force very dumb first. then other things...

medyagh: 13 because it wont pass 255 max number. and it is not too small and not too large.
just a starting point, so later we can tune

pkg/drivers/kic/oci/network.go
medyagh: the exit code is 1 and it doesnt seem to be helpful
```
med@xmac:~/workspace/minikube (nework_delete)$ docker network create --driver=bridge --subnet=192.168.49.0/24 -o com.docker.network.bridge.enable_ip_masquerade=true --label=created_by.minikube.sigs.k8s.io=true p3
Error response from daemon: Pool overlaps with other one on this address space
med@xmac:~/workspace/minikube (nework_delete)$ echo $?
1
```

pkg/minikube/machine/delete.go
medyagh: and this is a clean up code, so we just try and if it doesnt get deleted, we at least have tried.

pkg/drivers/kic/oci/network.go
tstromberg: Sorry for the miscommunication. All I am asking for is that the comment be updated to describe the "why", rather than restating the "what" that is perfectly described in the next line of code.

pkg/minikube/machine/delete.go
tstromberg: I guess that's the part I don't understand: if this fails unless the container is deleted (the final step), just simplify this and move it to occur explicitly afterwards. 

If this defer is really fixing a specific behavior, please document it in the comment and/or propose it in another PR.

PR #9095, opened 2020-08-26
WIP - Add exit.WithProblem, increase exit code dynamic range

### LINKED ISSUE #9080: Add problem ID and exit code for "Ensure your system has enough CPUs. The minimum allowed is 2 CPUs"

1. Run `minikube start` on a default Google Cloud Compute VM with 1 cpu with `--output=json`
2. Get the error `
{"data":{"exitcode":"64","message":"Ensure your  system has enough CPUs. The minimum allowed is 2 CPUs.\n"},"datacontenttype":"application/json","id":"507528d8-0e85-47bb-a2d1-08b41d868683","source":"https://minikube.sigs.k8s.io/","specversion":"1.0","type":"io.k8s.sigs.minikube.error"}
`

`gcloud` needs a machine readable error code that matches 1-to-1 with the message for this so it can understand what the error was without parsing the message.

tstromberg:
This PR allows packages to exit with an explicit problem, solution, and exit code. 

The effective dynamic range of exit codes has expanded from 11 to 78, and allows problems to define a related exit code to use. Some gymnastics were required for backwards compatibility, such as retaining the meaning of exit code 69 as `ProviderNotAvailable`

To demonstrate the new capabilities, I've fixed two Docker insufficient resource issues to exit with an explicit problem, solution, and error code. For example:

## Out of CPU's (PR)

Fixes #9080


tstromberg:
I'm going to re-open this once I have things better sorted.

Issue #9098, opened 2020-08-26
minikube start --download-only downloads duplicate kubeadm/kubelet binaries


tstromberg:
kubeadm & kubelet are already included in the preload tarball.

```
😄  minikube v1.12.3 on Darwin 10.15.5
✨  Automatically selected the docker driver. Other choices: hyperkit, vmware, parallels, virtualbox, vmwarefusion
👍  Starting control plane node minikube in cluster minikube
💾  Downloading Kubernetes v1.18.3 preload ...
    > preloaded-images-k8s-v5-v1.18.3-docker-overlay2-amd64.tar.lz4: 510.91 MiB
    > kubelet.sha256: 65 B / 65 B [--------------------------] 100.00% ? p/s 0s
    > kubeadm.sha256: 65 B / 65 B [--------------------------] 100.00% ? p/s 0s
    > kubectl.sha256: 65 B / 65 B [--------------------------] 100.00% ? p/s 0s
    > kubectl: 41.99 MiB / 41.99 MiB [---------------] 100.00% 3.03 MiB p/s 14s
    > kubeadm: 37.97 MiB / 37.97 MiB [---------------] 100.00% 2.23 MiB p/s 17s
    > kubelet: 108.04 MiB / 108.04 MiB [-------------] 100.00% 3.21 MiB p/s 33s
    > kubectl.sha256: 65 B / 65 B [--------------------------] 100.00% ? p/s 0s
    > kubectl: 47.84 MiB / 47.84 MiB [---------------] 100.00% 4.10 MiB p/s 12s
✅  Download complete!
```

n0npax:
I'll try to fix this.

/assign n0npax

afbjorklund:
It also includes `kubectl` for linux (for uploading to the node, at least)

If you want to download kubectl for the host, it might need a download.

n0npax:
I'll unassign, as seems it requires a little bit better understanding of the project.

afbjorklund:
> I'll unassign, as seems it requires a little bit better understanding of the project.

I think it was almost done, just that we missed that kubectl was also in the preload.

i.e. there is one kubectl for linux, that is in the preload and that we install to the node
then there is one optional kubectl client for host OS, that we download on demand
in this case, we want to skip downloading the kubectl linux that is already in the preload
should also make the code easier, since it doesn't have to special-case like before ?

priyawadhwa:
Help wanted on this.

sharifelgamal:
[...]

PR #9105, opened 2020-08-27
if --download-only skip pulling kubeadm and kubelet

n0npax:
Hi,

I'm not totally sure if this is a right approach, but should solve: https://github.com/kubernetes/minikube/issues/9098

### DIFF
--- a/pkg/minikube/machine/cache_binaries.go
+++ b/pkg/minikube/machine/cache_binaries.go
@@ -34,13 +34,13 @@ import (
 func CacheBinariesForBootstrapper(version string, clusterBootstrapper string) error {
 	binaries := bootstrapper.GetCachedBinaryList(clusterBootstrapper)
 
-	// If --download-only, skip kubeadm and kubelet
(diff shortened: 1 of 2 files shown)


pkg/minikube/machine/cache_binaries.go
```diff
@@ -21,17 +21,36 @@ import (
...
 	"k8s.io/minikube/pkg/minikube/download"
 )
 
 // CacheBinariesForBootstrapper will cache binaries for a bootstrapper
 func CacheBinariesForBootstrapper(version string, clusterBootstrapper string) error {
 	binaries := bootstrapper.GetCachedBinaryList(clusterBootstrapper)
 
+	// If --download-only, skip KubernetesReleaseBinaries
+	if viper.GetBool("download-only") {
+		var cBinaries []string
```
n0npax, line 39: or maybe just `return nil` instead of this logic

n0npax:
Thanks, I'll close this in that case :+1: I'm missing to much to make a progress here. Thanks!

Issue #9106, opened 2020-08-27
Kubernetes v1.19.0 starts up 3X slower than v1.18 on macOS


tstromberg:
Using `minikube start --kubernetes-version=v1.19.0` on macOS, I see a huge startup latency increase versus v1.18.x. This is using the Docker driver:

![Kubernetes Startup Latency (time to serving, in seconds)](https://user-images.githubusercontent.com/101424/91474903-b54cef80-e84f-11ea-9efe-1d66b1d16cf1.png)

tstromberg:
Oddly enough, this does not appear to impact usage on Linux.

PR #9111, opened 2020-08-28
Don't enable kubelet until after kubeadm generates config

priyawadhwa:
As written in the k8s 1.19 [CHANGELOG](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.19.md#downloads-for-v1190):

```
Kubeadm: Move the "kubeadm init" phase "kubelet-start" later in the init workflow, after the "kubeconfig" phase. This makes kubeadm start the kubelet only after the KubeletConfiguration component config file (/var/lib/kubelet/config.yaml) is generated and solves a problem where init systems like OpenRC cannot crashloop the kubelet service. (#90892, @xphoniex) [SIG Cluster Lifecycle]
```

Starting kubelet before the config exists caused up to 3x longer start time with 1.19. This PR removes the call to start kubelet before `kubeadm init`, and also puts the kubeadm phases in the correct order.

fixes https://github.com/kubernetes/minikube/issues/9106

### DIFF
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -572,9 +572,9 @@ func (k *Bootstrapper) restartControlPlane(cfg config.ClusterConfig) error {
 
 	baseCmd := fmt.Sprintf("%s %s", bsutil.InvokeKubeadm(cfg.KubernetesConfig.KubernetesVersion), phase)
 	cmds := []string{
-		fmt.Sprintf("%s phase kubelet-start --config %s", baseCmd, conf),
 		fmt.Sprintf("%s phase certs all --config %s", baseCmd, conf),
 		fmt.Sprintf("%s phase kubeconfig all --config %s", baseCmd, conf),
+		fmt.Sprintf("%s phase kubelet-start --config %s", baseCmd, conf),
 		fmt.Sprintf("%s phase %s all --config %s", baseCmd, controlPlane, conf),
 		fmt.Sprintf("%s phase etcd local --config %s", baseCmd, conf),
 	}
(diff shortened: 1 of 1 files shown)


pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -572,9 +572,9 @@ func (k *Bootstrapper) restartControlPlane(cfg config.ClusterConfig) error {
 
 	baseCmd := fmt.Sprintf("%s %s", bsutil.InvokeKubeadm(cfg.KubernetesConfig.KubernetesVersion), phase)
 	cmds := []string{
-		fmt.Sprintf("%s phase kubelet-start --config %s", baseCmd, conf),
 		fmt.Sprintf("%s phase certs all --config %s", baseCmd, conf),
 		fmt.Sprintf("%s phase kubeconfig all --config %s", baseCmd, conf),
+		fmt.Sprintf("%s phase kubelet-start --config %s", baseCmd, conf),
```
priyawadhwa, line 577: Note: I'm not sure if this will break k8s versions <1.19. It worked with the default when I tested locally, but if this breaks integration tests I'll put in some version-specific logic here

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -831,7 +831,7 @@ func (k *Bootstrapper) UpdateNode(cfg config.ClusterConfig, n config.Node, r cru
 		return errors.Wrap(err, "host alias")
 	}
 
-	return sm.Start("kubelet")
```
tstromberg, line 834: Can you confirm that Kubernetes v1.13.x starts up with this change?

priyawadhwa: confirmed, takes 35s on my machine

Issue #9134, opened 2020-08-31
--extra-config controller-manager options are ignored (masked by leader-elect)


tstromberg:
We have a regression bug in our kubeadm YAML generation, seemingly caused by #8431 where we set leader-elect=false.

We end up writing two `controllerManager` sections to the YAML file, and only the last one wins. It's possible this broken behavior exists for other components. I was able to verify it using `./out/minikube start --extra-config=kubelet.max-pods=10  --extra-config controller-manager.kube-api-qps=5`

`

```
apiVersion: kubeadm.k8s.io/v1beta2
kind: ClusterConfiguration
apiServer:
  certSANs: ["127.0.0.1", "localhost", "172.17.0.3"]
  extraArgs:
    enable-admission-plugins: "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,DefaultTolerationSeconds,NodeRestriction,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota"
controllerManager:
  extraArgs:
    kube-api-qps: "5"
certificatesDir: /var/lib/minikube/certs
clusterName: mk
controlPlaneEndpoint: control-plane.minikube.internal:8443
dns:
  type: CoreDNS
etcd:
  local:
    dataDir: /var/lib/minikube/etcd
    extraArgs:
      proxy-refresh-interval: "70000"
controllerManager:
  extraArgs:
    "leader-elect": "false"
```

We currently set `leader-elect=false` by default. Unfortunately, we do so in a way that

PR #9136, opened 2020-08-31
Fix --extra-config for scheduler/controllerManager by removing hardcoded values

priyawadhwa:
Now, users can specify additional flags to scheduler/controllerManager without overriding the place in the config where we set leader-elect=false

Verified that this works locally

Closes https://github.com/kubernetes/minikube/issues/9134


tstromberg:
Thank you!

PR #9144, opened 2020-09-01
Remove conflicting & inaccurate Docker Desktop memory warning

### LINKED ISSUE #8892: Warn if Docker is likely to encounter swap issues (performance impact)
we correcelty and nicely warn the user that their docker desktop is low on memory but still when we are Erroring, we spit out ugly messages... I suspect we could add a regex to this and make sure if this ever happens, hint to the user what is going on.

```
medya@~/workspace/minikube (crio_17_3) $ ./out/minikube start --driver=docker --base-image=local/kicbase:v0.0.10-snapshot
❗  Requested memory allocation (1739MB) is less than the recommended minimum 2000MB. Kubernetes may crash unexpectedly.
❗  Increase Docker for Desktop memory to at least 2.5GB or more:

        Docker for Desktop > Settings > Resources > Memory


🔥  Creating docker container (CPUs=2, Memory=1739MB) ...
✋  Stopping "minikube" in docker ...
🤦  StartHost failed, but will try again: creating host: create: creating: create kic node: container name "minikube": log: 2020-07-30T17:16:31.009963700Z  546 535 0:39 /docker/af9f0def1ca7ccd9b72c4da5e42395e3f6ab4ecb136893e606b5e1088a02bbba /sys/fs/cgroup/hugetlb rw,nosuid,nodev,noexec,relatime shared:148 master:28 - cgroup hugetlb rw,hugetlb
2020-07-30T17:16:31.010339600Z  547 535 0:40 /docker/af9f0def1ca7ccd9b72c4da5e42395e3f6ab4ecb136893e606b5e1088a02bbba /sys/fs/cgroup/pids rw,nosuid,nodev,noexec,relatime shared:149 master:29 - cgroup pids rw,pids
2020-07-30T17:16:31.010378100Z  549 535 0:42 /docker/af9f0def1ca7ccd9b72c4da5e42395e3f6ab4ecb136893e606b5e1088a02bbba /sys/fs/cgroup/systemd rw,nosuid,nodev,noexec,relatime shared:151 master:31 - cgroup cgroup rw,name=systemd'
2020-07-30T17:16:31.012576500Z  + docker_cgroup=/docker/af9f0def1ca7ccd9b72c4da5e42395e3f6ab4ecb136893e606b5e1088a02bbba: container exited unexpectedly
```

tstromberg:
I propose we remove this warning for a couple of reasons:

* The 3GB recommendation here conflicts with the 2.25 GB recommendation we make 18 lines above it.

* This puts us in an awkward position of providing conflicting advice: we first ask the user to increase to 2.25GB, then we warn that they haven't increased it to 3GB

* I have observed no performance degradation with 2GB of memory. Intuitively, with the exception of OOM pressure, the amount of memory has little measurable effect on Kubernetes performance. I would be willing to accept measurements that disprove this assertion, however.

* If there was a warning to be made here, there is nothing that should make it a Docker Desktop specific warning

* For people who intentionally allocate only 2GB of memory, the warning is noisy and there is no way to turn it off.

Fixes #8892

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -868,15 +868,6 @@ func validateRequestedMemorySize(req int, drvName string) {
 		out.WarnReason(reason.RsrcInsufficientReqMemory, "Requested memory allocation ({{.requested}}MB) is less than the recommended minimum {{.recommend}}MB. Deployments may fail.", out.V{"requested": req, "recommend": minRecommendedMem})
 	}
 
-	if driver.IsDockerDesktop(drvName) && containerLimit < 2997 && sysLimit > 8000 { // for users with more than 8 GB advice 3 GB
-		r := reason.RsrcInsufficientDarwinDockerMemory
-		if runtime.GOOS == "Windows" {
-			r = reason.RsrcInsufficientWindowsDockerMemory
-		}
-		r.Style = style.Improvement
-		out.WarnReason(r, "Docker Desktop has access to only {{.size}}MiB of the {{.sys}}MiB in available system memory. Consider increasing this for improved performance.", out.V{"size": containerLimit, "sys": sysLimit, "recommend": "3 GB"})
-	}
-
 	advised := suggestMemoryAllocation(sysLimit, containerLimit, viper.GetInt(nodes))
 	if req > sysLimit {
 		exitIfNotForced(reason.Kind{ID: "RSRC_OVER_ALLOC_MEM", Advice: "Start minikube with less memory allocated: 'minikube start --memory={{.advised}}mb'"},


tstromberg:
I did think of one scenario where this could possibly be an accurate warning: an unexpected swap behavior?

It's possible to trigger a swapping scenario with 1GB of memory usage:

```
docker@minikube:~$ free
              total        used        free      shared  buff/cache   available
Mem:        1008588      665576       60520        1984      282492      200532
Swap:        524284      160768      363516
```

With 2GB, it's more modest. It seems like we should just turn off swap altogether for Docker.

```
              total        used        free      shared  buff/cache   available
Mem:        2039264      802480       96580        8476     1140204     1077124
Swap:        524284       18004      506280
```

tstromberg:
Sent out a separate PR to fix the underlying swap performance issue that caused this warning to be emitted in the first place.

sharifelgamal (approved):

PR #9152, opened 2020-09-02
fix make extract

zhijianli88:
$ make extract
go run cmd/extract/extract.go
panic: err_map.go doesn't exist

goroutine 1 [running]:
main.main()
	/home/lizj/workspace/k8s/minikube/cmd/extract/extract.go:47 +0x1b4
exit status 2
Makefile:304: recipe for target 'extract' failed
make: *** [extract] Error 1

Fixes #9126
Signed-off-by: Li Zhijian <lizhijian@cn.fujitsu.com>


tstromberg:
err_map.go is not gone, it was just renamed in a recent refactor:

https://github.com/kubernetes/minikube/blob/master/pkg/minikube/reason/known_issues.go

Do you mind changing the code to use the new path to the file rather than removing the feature?

zhijianli88:
> err_map.go is not gone, it was just renamed in a recent refactor:
> 
> https://github.com/kubernetes/minikube/blob/master/pkg/minikube/reason/known_issues.go
> 
> Do you mind changing the code to use the new path to the file rather than removing the feature?

Yeap! thank you

medyagh (approved):

medyagh:
thank you @zhijianli88

PR #9159, opened 2020-09-02
Increase preload version to v6

priyawadhwa:
We upgraded the dashboard image in #9129, this will rebuild tarballs to include the latest version.

This should also fix a small performance regression, where start time was increased by 25% on kvm2, probably because preload failed and we were waiting for the new dashboard image to download. [PR bot comment](https://github.com/kubernetes/minikube/pull/9129#issuecomment-684035146)


afbjorklund:
I think that once one image is missing from the preload, it gives up on all the other ones too.

tstromberg:
@priyawadhwa - can you approve https://github.com/kubernetes/minikube/pull/9151 first?

PR #9163, opened 2020-09-02
docker: Improve overlay module check (behavior & UX)

tstromberg:
Some improvements to the driver performance UX, based on testing in WSL2:

* Fix & simplify  built-in overlay module detection (we had forgotten to strip the newline from the kernel version)
* Display suggested performance improvement in a single-line of output again
* Docker for Desktop is now known as Docker Desktop: https://www.docker.com/products/docker-desktop

## Old UX

```

💨  The 'docker' driver reported a performance issue
💡  Suggestion: enable overlayfs kernel module on your Linux

```

## New UX

`💨  For improved Docker performance, enable the overlay Linux kernel module using 'modprobe overlay'`

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -647,14 +647,6 @@ func validateDriver(ds registry.DriverState, existing *config.ClusterConfig) {
 	st := ds.State
 	glog.Infof("status for %s: %+v", name, st)
 
-	if st.NeedsImprovement {
-		out.WarnReason(reason.Kind{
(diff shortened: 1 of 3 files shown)


medyagh (changes requested):
this PR is removing the Second way of verifying if the first way is failing, which is bad
because on some distros (that I tried) the first way fails but the second way works


path := fmt.Sprintf("/lib/modules/%s/modules.builtin", string(out))

tstromberg:
It is not removing the second way - it's replacing both with something that
will work on all kernels with sysfs.

sharifelgamal (approved):

PR #9172, opened 2020-09-03
added warning for changing mount configuration with KiC driver on exsting cluster

Asarew:
This PR prints a warning when changing the mount configuration on an existing cluster when using either Docker or podman.

As discussed in https://github.com/kubernetes/minikube/pull/8159#issuecomment-666996559

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -210,7 +210,7 @@ func runStart(cmd *cobra.Command, args []string) {
 		if viper.GetBool(createMount) {
 			mount := viper.GetString(mountString)
 			if len(existing.ContainerVolumeMounts) != 1 || existing.ContainerVolumeMounts[0] != mount {
-				out.WarningT("Due to the limitations of {{.driver}}, it's not possible to change mount configuration of an existing cluster.", out.V{"driver": existing.Driver})
+				out.WarningT("Due to the limitations of {{.driver}}, it's not possible to the change mount configuration of an existing cluster.", out.V{"driver": existing.Driver})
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -206,14 +206,24 @@ func runStart(cmd *cobra.Command, args []string) {
...
-		stopProfile(existing.Name)
-		starter, err = provisionWithDriver(cmd, ds, existing)
-		if err != nil {
-			exit.Error(reason.GuestProvision, "error provisioning host", err)
+	if existing != nil && driver.IsKIC(existing.Driver) {
+		if viper.GetBool(createMount) {
+			mount := viper.GetString(mountString)
+			if len(existing.ContainerVolumeMounts) != 1 || existing.ContainerVolumeMounts[0] != mount {
+				out.WarningT("Due to the limitations of {{.driver}}, it's not possible to the change mount configuration of an existing cluster.", out.V{"driver": existing.Driver})
+				out.WarningT("If necessary delete and recreate the cluster, proceeding with old mount configuration")
```
tstromberg, line 214: This should exit rather than ignoring the user. Maybe something like:

`exit.Error(reason.GuestMountConflict, "Sorry, {{.driver}} does not allow mounts to be changed after container creation (previous mount: {{.previous}}, new mount: {{.new}})")`

(You would need to add a GuestMountConflict reason, but it shouldn't be too difficult to do so)

tstromberg (approved):

tstromberg:
Merging as we should have this warning for this release, though we should continue to improve upon it. Tested locally, works great!

PR #9178, opened 2020-09-03
exit on existing cluster start with KiC driver mount configuration change

Asarew:
when changing the mount configuration on a kic cluster exit instead showing a warning. as discussed in: https://github.com/kubernetes/minikube/pull/9172


pkg/minikube/reason/reason.go
```diff
@@ -214,6 +214,7 @@ var (
 	GuestDeletion         = Kind{ID: "GUEST_DELETION", ExitCode: ExGuestError}
 	GuestLoadHost         = Kind{ID: "GUEST_LOAD_HOST", ExitCode: ExGuestError}
 	GuestMount            = Kind{ID: "GUEST_MOUNT", ExitCode: ExGuestError}
+	GuestMountConflict    = Kind{ID: "GUEST_MOUNT_CONFLICT", ExitCode: ExGuestError}
```
tstromberg, line 217: The exit code should be `ExGuestConflict`

PR #9193, opened 2020-09-06
Remove misleading and ugly kubectl parameters

afbjorklund:
Chances are users want something more than "get pods"

`💡  kubectl not found. If you need it, try: 'minikube kubectl -- get pods -A'`

Running the bare command will show some usage info

```
kubectl controls the Kubernetes cluster manager.

 Find more information at: https://kubernetes.io/docs/reference/kubectl/overview/
...

```
It looks a bit ugly in the new screen shot, in my opinion

Also helps a bit on the way to making an alias: #7289

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -413,7 +413,7 @@ func showKubectlInfo(kcs *kubeconfig.Settings, k8sVersion string, machineName st
 
 	path, err := exec.LookPath("kubectl")
 	if err != nil {
-		out.T(style.Tip, "kubectl not found. If you need it, try: 'minikube kubectl -- get pods -A'")
+		out.T(style.Tip, "kubectl not found. If you need it, try: 'minikube kubectl --'")
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


priyawadhwa (approved):

cmd/minikube/cmd/start.go
```diff
@@ -413,7 +413,7 @@ func showKubectlInfo(kcs *kubeconfig.Settings, k8sVersion string, machineName st
 
 	path, err := exec.LookPath("kubectl")
 	if err != nil {
-		out.T(style.Tip, "kubectl not found. If you need it, try: 'minikube kubectl -- get pods -A'")
+		out.T(style.Tip, "kubectl not found. If you need it, try: 'minikube kubectl --'")
```
tstromberg, line 416: Personally, given the strange syntax, I don't feel like this is obvious enough. I wanted to give an example that a user is likely to understand without consulting additional documentation.

tstromberg: Maybe 'minikube kubectl -- help'?

afbjorklund: Beauty, as they say, is in the eye of the beholder. Maybe I should have used a different word (than ugly).

Did you know that `kubectl`, `kubectl --help` and `kubectl help` all have the _same_ output, by the way ?

I like "help" as an improvement (over "get"), but it wasn't really what I meant.

Maybe I will just close this PR, it was just the new screenshot that caused it.

screenshot

It looked a bit like an error, while it is perfectly normal. But maybe we want to avoid "other solutions" like:

```

Command 'kubectl' not found, but can be installed with:

sudo snap install kubectl

```

It just seemed weird to give an example command, rather than show how to "install" it (e.g. as an alias).

afbjorklund:
Here is how it looked in bash, for reference.

minikube-start-ubuntu-virtualbox

PR #9237, opened 2020-09-13
Fix ForwardedPort for podman version 2.0.1 and up

### LINKED ISSUE #9120: Podman v2 breaks the podman driver
The podman driver is still broken, with v2.0.4 and 2.0.5:

```
new client: new client: Error creating new ssh host from driver: Error getting ssh port for driver: get ssh host-port: convert host-port '\x00' to number: strconv.Atoi: parsing "": invalid syntax
```

This is a regression/follow-up from previous #8587 (2.0.1)

The workaround is still to use the v1.9.3 release instead.

The goal is to remove the need for current warning #8783

`❗  Using podman 2 is not supported yet. your version is "2.0.5". minikube might not work. use at your own risk.`

----

Note that podman 1 is no longer supported upstream:
https://podman.io/blogs/2020/08/01/deprecate-and-remove-varlink-notice.html
> We also plan to drop support from the RHEL8.4 release, spring 2021

And that RHEL 8 does **not** support docker, only podman.
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/building_running_and_managing_containers/index#running-containers-without-docker_building-running-and-managing-containers
> For RHEL 8, Docker is not included and not supported by Red Hat 

This basically means that you can not run KIC on Red Hat Enterprise Linux.

Fedora still includes Docker (Moby), but doesn't work because of cgroups v2.

afbjorklund:
Increasing Docker compatibility broke old Podman v1 code

Change to use the Docker parsing code for Podman >= 2.0.1

Closes #9120

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -252,7 +252,9 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 
 	validateFlags(cmd, driverName)
 	validateUser(driverName)
-	validateDockerStorageDriver(driverName)
+	if driverName == oci.Docker {
+		validateDockerStorageDriver(driverName)
+	}
 
 	// Download & update the driver, even in --download-only mode
 	if !viper.GetBool(dryRun) {
--- a/pkg/drivers/kic/oci/network.go
+++ b/pkg/drivers/kic/oci/network.go
@@ -24,6 +24,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/blang/semver"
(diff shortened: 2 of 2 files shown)


afbjorklund:
Also removed stupid warning from Docker.

:exclamation:  podman is currently using the overlay storage driver, consider switching to overlay2 for better performance

When using Podman, they are the same...

https://github.com/containers/storage/blob/master/drivers/overlay/overlay.go#L121_L122

tstromberg (approved):

medyagh (approved):

PR #9239, opened 2020-09-13
Add podman storage warning to match docker

### LINKED ISSUE #9238: Using the Podman driver shows Docker Desktop warnings
Since https://github.com/kubernetes/minikube/issues/9024

```

🧯  Docker is nearly out of disk space, which may cause deployments to fail! (95% of capacity)
💡  Suggestion: 

    Try at least one of the following to free up space on the device:
    
    1. Run "docker system prune" to remove unused docker data
    2. Increase the amount of memory allocated to Docker for Desktop via
    Docker icon > Preferences > Resources > Disk Image Size
    3. Run "minikube ssh -- docker system prune" if using the docker container runtime
🍿  Related issue: https://github.com/kubernetes/minikube/issues/9024
```

There is a `sudo podman system prune` (1), but there is no Podman Desktop (2).

The suggestion to clean up the internal docker (3) could still be valid, though.

afbjorklund:
```

🧯  Podman is nearly out of disk space, which may cause deployments to fail! (95% of capacity)
💡  Suggestion: 

    Try at least one of the following to free up space on the device:
     
    1. Run "sudo podman system prune" to remove unused podman data
    2. Run "minikube ssh -- docker system prune" if using the docker container runtime
🍿  Related issue: https://github.com/kubernetes/minikube/issues/9024

```

Closes #9238

### DIFF
--- a/pkg/minikube/machine/start.go
+++ b/pkg/minikube/machine/start.go
@@ -36,6 +36,7 @@ import (
 	"github.com/juju/mutex"
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
+	"k8s.io/minikube/pkg/drivers/kic/oci"
 	"k8s.io/minikube/pkg/minikube/command"
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/constants"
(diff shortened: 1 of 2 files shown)


pkg/minikube/machine/start.go
```diff
@@ -218,17 +219,32 @@ func postStartValidations(h *host.Host, drvName string) {
 		glog.Warningf("error getting command runner: %v", err)
 	}
 
+	var kind reason.Kind
+	var name string
+	if drvName == oci.Docker {
+		kind = reason.RsrcInsufficientDockerStorage
+		name = "Docker"
```
medyagh, line 226: better to use this helper instead
```
driver.FullName(drvName)
```

medyagh (changes requested):

pkg/minikube/machine/start.go
afbjorklund: But that one has some weird "Docker Desktop" special case ?

I agree it is not ideal to compare driver name and program name

PR #9256, opened 2020-09-15
Avoid setting time for memory assets

### LINKED ISSUE #9252: Minikube memory assets have a strange modification date
The MemoryAsset in minikube doesn't have a date set (uses `time.Time{}`).

This causes errors on certain platforms, for instance on the Raspberry Pi:

```console
pi@raspberrypi:~ $ touch -d '0001-01-01 00:00:00 +0000' /tmp/test
touch: invalid date format '0001-01-01 00:00:00 +0000'
```

It "works" on the regular platform, but still looks a bit odd and is probably wrong.

afbjorklund:
These do not have any file modification time

Previously used '0001-01-01 00:00:00 +0000'

Closes #9252

### DIFF
--- a/pkg/minikube/command/ssh_runner.go
+++ b/pkg/minikube/command/ssh_runner.go
@@ -263,7 +263,7 @@ func (s *SSHRunner) Copy(f assets.CopyableFile) error {
 	mtime, err := f.GetModTime()
 	if err != nil {
 		glog.Infof("error getting modtime for %s: %v", dst, err)
-	} else {
+	} else if mtime != (time.Time{}) {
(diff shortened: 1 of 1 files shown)


medyagh (approved):

pkg/minikube/command/ssh_runner.go
```diff
@@ -263,7 +263,7 @@ func (s *SSHRunner) Copy(f assets.CopyableFile) error {
 	mtime, err := f.GetModTime()
 	if err != nil {
 		glog.Infof("error getting modtime for %s: %v", dst, err)
-	} else {
+	} else if mtime != (time.Time{}) {
```
tstromberg, line 266: In the future, this would probably better as:

`if !mtime.IsZero() { do something }`

PR #9260, opened 2020-09-16
Implement scheduled stop

priyawadhwa:
This commit implements scheduled stop by using a daemonize golang package. The flow is as follows:

1. User runs `minikube stop --schedule 1m` to schedule a stop 1m from now
1. Check if a previous scheduled stop is running, kill the process if it is
1. We daemonize the current command, store the new PID of that command in minikube home dir, and update the cluster config
1. Sleep for 1m, then execute stop as usual

Coming in next PRs:
1. Use the info in the cluster config to make sure `minikube status` shows that a stop has been scheduled
1. Add some easy way to cancel any scheduled stops

Note: I've tested that this works on mac and linux, but need to check windows still

### DIFF
--- a/cmd/minikube/cmd/stop.go
+++ b/cmd/minikube/cmd/stop.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cmd
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/docker/machine/libmachine"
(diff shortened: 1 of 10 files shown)


priyawadhwa:
closing as this daemon library doesn't support windows

PR #9267, opened 2020-09-17
Reduce cyclotomic complexity of CreateContainerNode to fix lint

sharifelgamal:
Lint was broken at HEAD, this should fix it.


priyawadhwa (approved):

afbjorklund:
Probably my fault, that

PR #9287, opened 2020-09-21
"cache add" command: improve error message for not existing image

Evalle:
PR Summary:
- Improves error message for not existing image

### DIFF
--- a/pkg/minikube/image/cache.go
+++ b/pkg/minikube/image/cache.go
@@ -60,7 +60,7 @@ func SaveToDir(images []string, cacheDir string) error {
 			dst := filepath.Join(cacheDir, image)
 			dst = localpath.SanitizeCacheDir(dst)
 			if err := saveToTarFile(image, dst); err != nil {
-				glog.Errorf("save image to file %q -> %q failed: %v", image, dst, err)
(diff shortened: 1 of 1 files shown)


pkg/minikube/image/cache.go
```diff
@@ -60,7 +60,7 @@ func SaveToDir(images []string, cacheDir string) error {
 			dst := filepath.Join(cacheDir, image)
 			dst = localpath.SanitizeCacheDir(dst)
 			if err := saveToTarFile(image, dst); err != nil {
-				glog.Errorf("save image to file %q -> %q failed: %v", image, dst, err)
+				glog.Errorf("save image to file %q -> %q failed: the image doesn't exist", image, dst)
```
medyagh, line 63: thank you for taking this task,  
this will changes the error log, this doesn't seem to solve the example in the issue.

I would rather this instead of erroring logs give a solution message to the user.

please make sure to run the example in the issue and paste the Before/After output in the PR description.

medyagh (changes requested):
#

tstromberg:
This PR is now stale. Please re-open once PR comments are resolved. Thank you for your contribution!

Issue #9290, opened 2020-09-21
Improve handling of DRV_CP_ENDPOINT error


priyawadhwa:
This error was seen on Cloud Shell:

![image](https://user-images.githubusercontent.com/10676490/93817330-cdcde100-fc26-11ea-9784-58d46a5e77f5.png)

There are no additional logs. Two observations made by the reporter (@russellwolf):

1. They didn't realize there had been an error at all, because the red X emoji blended in with the rest of the emojis
1. They were still able to deploy to the cluster even though it failed

It seems like kubernetes was able to fix itself, and that's why the reporter was able to deploy to the cluster. To improve the experience, we could:

1. Offer advice to run "minikube delete" and "minikube start" again when this error is seen
1. Change the error emoji or reformat the output to make it clear that something has gone wrong

PR #9291, opened 2020-09-21
Add advice for the DRV_CP_ENDPOINT error

priyawadhwa:
Output now looks like:

```
$ minikube start -p test
😄  [test] minikube v1.13.1 on Darwin 10.15.6
👍  Starting control plane node test in cluster test
🏃  Updating the running hyperkit "test" VM ...
🐳  Preparing Kubernetes v1.19.2 on Docker 19.03.8 ...

❌  Exiting due to DRV_CP_ENDPOINT: Failed to get API Server URL: <nil>
💡  Suggestion: 

    Recreate the cluster by running:
    minikube delete --profile=test
    minikube start --profile=test
```

Fixes https://github.com/kubernetes/minikube/issues/9290

### DIFF
--- a/pkg/minikube/reason/reason.go
+++ b/pkg/minikube/reason/reason.go
@@ -207,7 +207,13 @@ var (
 	ProviderNotFound    = Kind{ID: "PROVIDER_NOT_FOUND", ExitCode: ExProviderNotFound}
 	ProviderUnavailable = Kind{ID: "PROVIDER_UNAVAILABLE", ExitCode: ExProviderNotFound, Style: style.Shrug}
 
-	DrvCPEndpoint         = Kind{ID: "DRV_CP_ENDPOINT", ExitCode: ExDriverError}
+	DrvCPEndpoint = Kind{ID: "DRV_CP_ENDPOINT",
+		Advice: `Recreate the cluster by running:
+		minikube delete{{.profile}}
+		minikube start{{.profile}}`,
+		ExitCode: ExDriverError,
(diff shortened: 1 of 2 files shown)


tstromberg (approved):

pkg/minikube/reason/reason.go
```diff
@@ -207,7 +207,13 @@ var (
 	ProviderNotFound    = Kind{ID: "PROVIDER_NOT_FOUND", ExitCode: ExProviderNotFound}
 	ProviderUnavailable = Kind{ID: "PROVIDER_UNAVAILABLE", ExitCode: ExProviderNotFound, Style: style.Shrug}
 
-	DrvCPEndpoint         = Kind{ID: "DRV_CP_ENDPOINT", ExitCode: ExDriverError}
+	DrvCPEndpoint = Kind{ID: "DRV_CP_ENDPOINT",
+		Advice: `Recreate the cluster by running:
+		minikube delete{{.profile}}
```
medyagh, line 212: how would the output look like if the profile is not minikube?
should it include minikube delete -p {{.profile}}  ?

medyagh (changes requested):
please paste output of a profile other than default for confirmation

pkg/minikube/reason/reason.go
priyawadhwa: good catch! fixed it, and updated output in the PR description

medyagh (approved):

PR #9298, opened 2020-09-22
update_kubernetes_version: automate the update of DefaultKubernetesVersion

### LINKED ISSUE #4392: Write a script to automatically propose PR's due to a new Kubernetes release
I would love a script that we can run from a cronjob or Jenkins that monitors https://github.com/kubernetes/kubernetes/releases for new releases.

The script should propose a PR for any release newer than NewestKubernetesVersion:
 https://github.com/kubernetes/minikube/blob/f8fa0748c5a474bc2a62816ce318a6740fc00fea/pkg/minikube/constants/constants.go#L168 

If the release is a non-beta release, it should also change DefaultKubernetesVersion in the same PR:

https://github.com/kubernetes/minikube/blob/f8fa0748c5a474bc2a62816ce318a6740fc00fea/pkg/minikube/constants/constants.go#L165

This script should either build upon or replace  https://github.com/kubernetes/minikube/blob/master/hack/kubernetes_version/update_kubernetes_version.go - which makes the changes necessary in our code base, but is designed for interactive use and does not know how to look for new Kubernetes releases.

We can provide a github token via an environment variable.

prezha:
fixes #4392 (iteration 0)

### description:

existing `hack/kubernetes_version/update_kubernetes_version.go` script was modified to automate the update of `DefaultKubernetesVersion` and `NewestKubernetesVersion` - it now polls GitHub Releases (using `github.com/google/go-github`) for the latest Kubernetes [pre-]releases and updates the `pkg/minikube/constants/constants.go` accordingly, whereas:

- `DefaultKubernetesVersion` is updated to the latest release version 
- `NewestKubernetesVersion` is updated to the latest `-rc` or `-beta` pre-release version
- if such `NewestKubernetesVersion` is less than `DefaultKubernetesVersion`, then `DefaultKubernetesVersion` would be used for both
- `make test` passes (~~**note**: `make generate-docs` must be run before `make test`, or `TestGenerateDocs` test will fail~~ - _this dependency/requirement was removed in [commit 1444ce3](https://github.com/prezha/minikube/commit/1444ce37d7aadb3796c9b94345c21b6edfc9d49b)_)

### DIFF
--- a/hack/kubernetes_version/update_kubernetes_version.go
+++ b/hack/kubernetes_version/update_kubernetes_version.go
@@ -18,12 +18,13 @@ package main
 
 import (
 	"context"
-	"fmt"
+	"flag"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/golang/glog"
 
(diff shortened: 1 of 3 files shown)


hack/kubernetes_version/update_kubernetes_version.go
```diff
@@ -77,7 +128,7 @@ func main() {
 			return err
 		}
 		re = regexp.MustCompile(`kubernetesVersion: .*`)
-		cf = []byte(re.ReplaceAllString(string(cf), "kubernetesVersion: "+v))
+		cf = []byte(re.ReplaceAllString(string(cf), "kubernetesVersion: "+vDefaultMM+".0")) // TODO: let <PATCH> version to be the latest one instead of "0"
```
tstromberg, line 131: Can you clarify this TODO? I think it may be removable.

It's intentional that the tests are using `.0` to avoid unnecessary testdata updates for every minor release.

hack/kubernetes_version/update_kubernetes_version.go
```diff
@@ -25,17 +26,65 @@ import (
...
+			break
+		}
+		opt.Page = resp.NextPage
 	}
 
-	v := os.Args[1]
-	if !strings.HasPrefix(v, "v") {
-		v = "v" + v
+	if vDefault == "" || vNewest == "" {
+		glog.Errorf("Cannot determine DefaultKubernetesVersion or NewestKubernetesVersion")
```
tstromberg, line 86: Use glog.Fatalf

hack/kubernetes_version/update_kubernetes_version.go
```diff
@@ -53,17 +102,19 @@ func main() {
 	mode := info.Mode()
```
tstromberg, line 103: Do you mind adding a logging line describing the value of `vDefault` and `vNewest` that was discovered? It's difficult to tell if this is actually doing anything with the current repository state.

tstromberg (changes requested):
PR looks good and about ready to merge.

hack/kubernetes_version/update_kubernetes_version.go
prezha: i thought it could be needed sometimes later, but if we would stick always to `.0` minor release, then yes, that TODO is not needed
then also, do we need the testData update section here at all if it is static (ie, miniro release is fixed to `0.`) or shall we remove it altogether?

hack/kubernetes_version/update_kubernetes_version.go
prezha: i've changed that, thank you

hack/kubernetes_version/update_kubernetes_version.go
prezha: yes, it makes a lot of sense - i'll add a logging line describing the value of `vDefault` and `vNewest` that was discovered

hack/kubernetes_version/update_kubernetes_version.go
prezha: @tstromberg i've submitted a new pr where i removed the 'todo' and left `testData` update section as-is until we clarify it

hack/kubernetes_version/update_kubernetes_version.go
prezha: changed in new pr

hack/kubernetes_version/update_kubernetes_version.go
prezha: add in new pr

medyagh (approved):
thanks

PR #9299, opened 2020-09-22
Don't allow --base-image for VM drivers

### LINKED ISSUE #8891: don't allow --base-image flag to be set for VM drivers
currently if I specify --base-image for an existing hyperkit VM, minikube doesnt even notice it.
we should exit quick, and tell users, did you mean to use docker or podman driver insead ? and explain them base image is for docker/podman driver and for VM it is called ISO_URL

```
medya@~/workspace/minikube (crio_17_3) $ ./out/minikube start --base-image=local/kicbase:v0.0.10-snapshot
🏃  Updating the running hyperkit "minikube" VM ...
🐳  Preparing Kubernetes v1.18.3 on Docker 19.03.12 ...
```

dominikbraun:
Fixes #8891. 

**Before:**

```shell script
$ minikube start --base-image=local/kicbase:v0.0.10-snapshot

(minikube doesn't notice the bad combination and proceeds)
```

**After:**

```shell script
$ minikube start --base-image=local/kicbase:v0.0.10-snapshot
* minikube v1.13.1 auf Microsoft Windows 10 Pro 10.0.18363 Build 18363
* Kubernetes 1.19.2 is now available. If you would like to upgrade, specify: --kubernetes-version=v1.19.2

X Exiting due to MK_USAGE: Sorry, --base-image can only be used with the Docker or Podman driver. For VM drivers, use --iso-url instead
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1026,6 +1026,10 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 		}
 	}
 
+	if driver.IsVM(drvName) && cmd.Flags().Changed(kicBaseImage) {
+		exit.Message(reason.Usage, "Sorry, --base-image can only be used with the Docker or Podman driver. For VM drivers, use --iso-url instead")
+	}
+
 	// validate kubeadm extra args
 	if invalidOpts := bsutil.FindInvalidExtraConfigFlags(config.ExtraOptions); len(invalidOpts) > 0 {
 		out.WarningT(


dominikbraun:
I signed it

cmd/minikube/cmd/start.go
```diff
@@ -1026,6 +1026,10 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 		}
 	}
 
+	if driver.IsVM(drvName) && cmd.Flags().Changed(kicBaseImage) {
+		exit.Message(reason.Usage, "Sorry, --base-image can only be used with the Docker or Podman driver. For VM drivers, use --iso-url instead")
```
priyawadhwa, line 1030: I don't think we need to exit here, instead we can just warn the user:

```go
out.WarningT("The --base-image flag is only applicable to the docker and podman drivers and will be ignored")
```

priyawadhwa (commented):
Thanks for fixing this! Left a comment --

I also think we could add "for use with docker and podman drivers" to the help text for the `--base-image` flag

medyagh:
sorry @dominikbraun this issue was already assigned, but I hope I see another PR from you

PR #9302, opened 2020-09-22
Return ExHostPermission for insufficient config file permissions

### LINKED ISSUE #9292: use docker's "--restart" feature
when creating kic containers, we can apply this to auto-restart the container on failure upto once

```
--restart=on-failure:1
```

### LINKED ISSUE #9297: root-owned config exits with ExHostConfig istead of ExHostPermissions
To reproduce:

`minikube stop`
`sudo chown -R root $HOME/.minikube`
`minikube start`

```
./out/minikube start --driver=docker; echo $status                                                                                                  
E0921 18:18:31.603532   43694 cloud_events.go:57] unable to write to /Users/tstromberg/.minikube/profiles/minikube/events.json: open /Users/tstromberg/.minikube/profiles/minikube/events.json: permission denied
😄  minikube v1.13.1 on Darwin 10.15.6

❌  Exiting due to HOST_CONFIG_LOAD: Unable to load config: read: open /Users/tstromberg/.minikube/profiles/minikube/config.json: permission denied
```

dominikbraun:
Fixes #9297.

Insufficient permissions for reading the config will now result in a `HOST_PERMISSION_DENIED` instead of `HOST_CONFIG_LOAD`, which uses `ExHostPermission` as exit code. I'm yet not sure if this is an appropriate `reason.Kind` instance - the alternative could be to just stick with `HOST_CONFIG_LOAD` and change its `ErrorCode` manually.

### DIFF
--- a/pkg/minikube/reason/reason.go
+++ b/pkg/minikube/reason/reason.go
@@ -192,6 +192,7 @@ var (
 	HostHomeChown        = Kind{ID: "HOST_HOME_CHOWN", ExitCode: ExHostPermission}
 	HostBrowser          = Kind{ID: "HOST_BROWSER", ExitCode: ExHostError}
 	HostConfigLoad       = Kind{ID: "HOST_CONFIG_LOAD", ExitCode: ExHostConfig}
+	HostPermissionDenied = Kind{ID: "HOST_PERMISSION_DENIED", ExitCode: ExHostPermission}
 	HostCurrentUser      = Kind{ID: "HOST_CURRENT_USER", ExitCode: ExHostConfig}
 	HostDelCache         = Kind{ID: "HOST_DEL_CACHE", ExitCode: ExHostError}
 	HostKillMountProc    = Kind{ID: "HOST_KILL_MOUNT_PROC", ExitCode: ExHostError}
--- a/pkg/minikube/reason/reason.go
+++ b/pkg/minikube/reason/reason.go
@@ -188,11 +188,17 @@ var (
 
 	RsrcInsufficientStorage = Kind{ID: "RSRC_INSUFFICIENT_STORAGE", ExitCode: ExInsufficientStorage, Style: style.UnmetRequirement}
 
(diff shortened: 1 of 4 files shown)


pkg/minikube/reason/reason.go
```diff
@@ -192,6 +192,7 @@ var (
 	HostHomeChown        = Kind{ID: "HOST_HOME_CHOWN", ExitCode: ExHostPermission}
 	HostBrowser          = Kind{ID: "HOST_BROWSER", ExitCode: ExHostError}
 	HostConfigLoad       = Kind{ID: "HOST_CONFIG_LOAD", ExitCode: ExHostConfig}
+	HostPermissionDenied = Kind{ID: "HOST_PERMISSION_DENIED", ExitCode: ExHostPermission}
```
tstromberg, line 195: This looks basically good, but can get better if we remove the duplicate data within `known_issues.go`.  

Please rename this to `HostHomePermission` and copy the "HOST_HOME_PERMISSION" `Kind{}` object from: 

https://github.com/kubernetes/minikube/blob/3f9c6d10d9cd4a56abe436f6963d5edb5a00c67e/pkg/minikube/reason/known_issues.go#L205

This will then provide the user appropriate advice if they run into this error. Your exit code, ExHostPermission is the correct one to use, however.

Then, replace the `HOST_HOME_PERMISSION Kind{}` within `known_issues.go` with a reference to `HostHomePermission`. This way we may also provide the correct feedback for any other errors which match this regexp in the future.

tstromberg (commented):
This looks wonderful, just one comment to reduce code duplication. Thank you!

dominikbraun:
Thanks for your feedback! I've copied the `HOST_HOME_PERMISSION` instance from `known_issues.go`, changed the exit code and renamed the variable. I'm also reffering to the correct issue number in the `Fixes` clause of this PR now :sweat_smile:

tstromberg:
Looks good - please update `known_issues.go` within this PR to use your new `HostHomePermission` variable instead of duplicating the contents.

medyagh (approved):
thank you for this PR, this will improve minikube's usability

PR #9308, opened 2020-09-22
Add PR bot code to repo

priyawadhwa:
This PR adds the PR bot code to the repo. The bot:

1. Gets a list of all open PRs in the repo with the ok-to-test label that have new commits
1. Runs mkcmp on those PRs against minikube @ HEAD
1. Comments STDOUT from the mkcmp run on the PR

### DIFF
--- a/pkg/perf/monitor/constants.go
+++ b/pkg/perf/monitor/constants.go
@@ -1,0 +1,24 @@
+/*
+Copyright 2020 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


pkg/perf/monitor/constants.go
```diff
@@ -0,0 +1,25 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package monitor
+
+const (
+	GithubAccessTokenEnvVar = "GITHUB_ACCESS_TOKEN"
```
medyagh, line 20: this should be getting from ENV vars

medyagh (changes requested):

pkg/perf/monitor/constants.go
priyawadhwa: GITHUB_ACCESS_TOKEN is the name of the ENV var, which is set in the VM

tstromberg (approved):

PR #9314, opened 2020-09-24
Mention Docker for Linux in registries documentation

### LINKED ISSUE #8925: Documentation: change registries title from "docker on macOS" to "docker on Linux/macOS"
Isn't it worth mentioning something about Linux?

I managed to make the private registry work on my Ubuntu 20.04 machine by following the steps listed in "docker on macOS". Maybe it would be interesting to change that title to "Docker on Linux/macOS".

That said, I haven't followed the whole handbook. I just read the "Registries" page.

dominikbraun:
Fixes #8925. Anything else we could improve in this documentation section?

### DIFF
--- a/site/content/en/docs/handbook/registry.md
+++ b/site/content/en/docs/handbook/registry.md
@@ -44,9 +44,9 @@ One nifty hack is to allow the kubelet running in minikube to talk to registries
 with TLS certificates. Because the default service cluster IP is known to be available at 10.0.0.1, users can pull images from registries
 deployed inside the cluster by creating the cluster with `minikube start --insecure-registry "10.0.0.0/24"`.
 
-### docker on macOS
+### docker on Linux/macOS
 
-Quick guide for configuring minikube and docker on macOS, enabling docker to push images to minikube's registry.
+Quick guide for configuring minikube and docker on Linux or macOS, enabling docker to push images to minikube's registry.
 
 The first step is to enable the registry addon:
 


medyagh:
technically there is no difference between docker on winows and docker on macos.

but I think what this doc was meant to do was making sure to make users aware of the Random Port on Docker on MacOS and docker on Windows


on linux the port is same 5000
but on mac os/windows it will give you random port.

afbjorklund (changes requested):
It doesn't make sense to talk about the "docker virtual machine" on Linux, since there is no Docker Desktop product available (only Docker Engine which doesn't have a VM and doesn't need the `socat`)

afbjorklund:
It probably needs a **separate** entry on Linux. Normally you would talk to $(minikube ip):5000, as a "insecure registry".

Tunneling localhost:5000 to the Minikube VM is _one_ way of getting around the SSL requirement (localhost is "insecure").

site/content/en/docs/handbook/registry.md
```diff
@@ -44,9 +44,9 @@ One nifty hack is to allow the kubelet running in minikube to talk to registries
 with TLS certificates. Because the default service cluster IP is known to be available at 10.0.0.1, users can pull images from registries
 deployed inside the cluster by creating the cluster with `minikube start --insecure-registry "10.0.0.0/24"`.
 
-### docker on macOS
+### docker on Linux/macOS
```
medyagh, line 47: this doesn't seem right, 

docker on Linux is one way
docker on mac os and windows is the other way.

this is due to docker limitaitons on non-linux

and the limitation is, on non-linux we can not use the port 5000. but use a different random port.

medyagh: this doc needs to be reviewed from scratch to make sure the language is correct all the way.

medyagh (changes requested):

dominikbraun:
I get the point, just adding "Linux" to the heading doesn't seem to make sense here. Since correctly documenting this for all operating systems isn't quite easy, I'm going to close this PR for the time being.

Issue #9313, opened 2020-09-23
minikube preload extract: avoid hundreds page of error when disk is full


medyagh:
This error was too long to even be copy pasted as I see hundreds of pages of same message

to replicate:
Create 4 minikubes on a Vm with small disk ( I tried on ec2 ubuntu)
```
ubuntu@ip-172-31-6-55:~$ df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/root       7.7G  7.7G     0 100% /
devtmpfs        3.9G     0  3.9G   0% /dev
tmpfs           3.9G     0  3.9G   0% /dev/shm
tmpfs           796M   30M  767M   4% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           3.9G     0  3.9G   0% /sys/fs/cgroup
/dev/loop0       97M   97M     0 100% /snap/core/9804
/dev/loop1       56M   56M     0 100% /snap/core18/1885
/dev/loop2       71M   71M     0 100% /snap/lxd/16922
/dev/loop3       29M   29M     0 100% /snap/amazon-ssm-agent/2012
tmpfs           796M     0  796M   0% /run/user/1000
````

```
/usr/bin/tar: ./lib/docker/image/overlay2/distribution/v2metadata-by-diffid/sha256/e9933a1f21f5410592997c2b9549416206aa0fee7a8a765a2eb89e017a6d5eeb: Cannot open: No such file or directory

```

minikube should detect this and exit gracefully with soluttion message that there is no space left on your computer. 

```
ubuntu@ip-172-31-6-55:~$ ./minikube-pr-9294 start -p hello --alsologtostderr
W0923 20:40:49.968011   16683 root.go:252] Error reading config file at /home/ubuntu/.minikube/config/config.json: open /home/ubuntu/.minikube/config/config.json: no such file or directory
I0923 20:40:49.969355   16683 out.go:191] Setting JSON to false
I0923 20:40:49.972810   16683 start.go:102] hostinfo: {"hostname":"ip-172-31-6-55","uptime":1186,"bootTime":1600892463,"procs":286,"os":"linux","platform":"ubuntu","platformFamily":"debian","platformVersion":"20.04","kernelVersion":"5.4.0-1024-aws","virtualizationSystem":"xen","virtualizationRole":"guest","hostid":"7d421595-9427-4bda-8b92-85c19b2969d7"}
I0923 20:40:49.973219   16683 start.go:112] virtualization: xen guest
[...]

PR #9316, opened 2020-09-24
preload: print solution message when there is no space left on device

priyawadhwa:
I was initially debating between warning and exiting here, but decided to exit because the "cache and load" step would probably fail with the same error if preload extraction fails.

fixes https://github.com/kubernetes/minikube/issues/9313


medyagh (commented):
I suspect this will still print tens of pages of error that will fill up the terminal, if the disk space check is not done before the "extract" part.

do u mind pasting out an example to see how the full output looks like ?
(to reproduce you can use a default ec2 VM that only has 8 GB disk and creating 4 mnikube clusters will fill up the disk

priyawadhwa:
Hey @medyagh the output looks like this:

```
$ minikube start --driver docker
😄  minikube v1.13.1 on Darwin 10.15.6
🔥  Creating docker container (CPUs=2, Memory=4893MB) ...

❌  Exiting due to RSRC_DOCKER_STORAGE: preload extraction failed: "No space left on device"
💡  Suggestion: 

    Try at least one of the following to free up space on the device:
    
    1. Run "docker system prune" to remove unused docker data
    2. Increase the amount of memory allocated to Docker for Desktop via
    Docker icon > Preferences > Resources > Disk Image Size
    3. Run "minikube ssh -- docker system prune" if using the docker container runtime
🍿  Related issue: https://github.com/kubernetes/minikube/issues/9024
```

the error itself is never printed, which is why you don't see the pages of output. instead, we just print the solution.

pkg/drivers/kic/kic.go
```diff
@@ -134,6 +136,9 @@ func (d *Driver) Create() error {
 		glog.Infof("Starting extracting preloaded images to volume ...")
 		// Extract preloaded images to container
 		if err := oci.ExtractTarballToVolume(d.NodeConfig.OCIBinary, download.TarballPath(d.NodeConfig.KubernetesVersion, d.NodeConfig.ContainerRuntime), params.Name, d.NodeConfig.ImageDigest); err != nil {
+			if strings.Contains(err.Error(), "No space left on device") {
+				exit.Message(reason.RsrcInsufficientDockerStorage, "preload extraction failed: \"No space left on device\"")
```
medyagh, line 140: I am not a big fan of libraries to exit, how about retruning the error to the caller ? and ensure the caller fails fast and does not do retry?

medyagh (changes requested):

pkg/drivers/kic/kic.go
priyawadhwa: Sure, done.

medyagh (approved):
thanks

PR #9327, opened 2020-09-26
Fix SVG logo file (white/icon)

weibeld:
The file [`images/logo/logo_white.svg`](https://github.com/kubernetes/minikube/blob/master/images/logo/logo_white.svg) included invalid namespace URLs (`https` instead of `http`) and for this reason could not be displayed by any of the browsers that I tested (Chrome, Safari, Firefox, Opera). It resulted in the following error:

Screenshot 2020-09-26 at 14 55 09

 I fixed the URLs to be analogous to the black version of the logo ([`images/logo/logo.svg`](https://github.com/kubernetes/minikube/blob/master/images/logo/logo.svg)) and now the file can be correctly displayed.

Apparently, this bug has been introduced by #3061 (not sure what the reason for this was).


tstromberg:
Thank you!

PR #9225, opened 2020-09-12
Fork all files needed for kindbase and specify snapshot

afbjorklund:
This is for reference, to be able to combine them

Copied from https://github.com/kubernetes-sigs/kind/tree/2c0eee40/images/base

Currently the "entrypoint" is overridden with a more recent, and modified, version:

https://github.com/kubernetes/minikube/blob/v1.13.0/deploy/kicbase/entrypoint

Changed from latest greatest 20.04 to a snapshot

```diff
-FROM ubuntu:20.04
+FROM ubuntu:focal-20200423
```

Upstream KIND has moved away from Ubuntu LTS and moved on to 20.10 already

Fixes #7788

-----

There is **no** difference to the actual size being done here, that is a separate story.

```
REPOSITORY                     TAG                           IMAGE ID            CREATED             SIZE
local/kindbase                 v20200430-2c0eee40-snapshot   9e379cea8e97        6 minutes ago       284MB
gcr.io/k8s-minikube/kindbase   v20200430-2c0eee40            9e379cea8e97        6 minutes ago       284MB
gcr.io/k8s-minikube/kindbase   v20200430-2c0eee40-snapshot   9e379cea8e97        6 minutes ago       284MB
kindest/base                   v20200430-2c0eee40            feff3ae05a74        4 months ago        282MB
```

Here is the result that comes from rebuilding the image with later apt packages:

```

-----Apt-----

Packages found only in kindest/base:v20200430-2c0eee40: None

Packages found only in local/kindbase:v20200430-2c0eee40-snapshot: None

Version differences:
PACKAGE                   IMAGE1 (kindest/base:v20200430-2c0eee40)        IMAGE2 (local/kindbase:v20200430-2c0eee40-snapshot)
-bash                     5.0-6ubuntu1, 1.6M                              5.0-6ubuntu1.1, 1.6M
-ca-certificates          20190110ubuntu1, 384K                           20190110ubuntu1.1, 382K
-curl                     7.68.0-1ubuntu2, 401K                           7.68.0-1ubuntu2.2, 401K
-libapparmor1             2.13.3-7ubuntu5, 167K                           2.13.3-7ubuntu5.1, 167K
-libcurl4                 7.68.0-1ubuntu2, 688K                           7.68.0-1ubuntu2.2, 688K
-libjson-c4               0.13.1 dfsg-7, 92K                              0.13.1 dfsg-7ubuntu0.3, 92K
[...]

PR #9135, opened 2020-08-31
kic base image: install stable containerd and clean up packages

afbjorklund:
Remove local things from kind, replace with packages.

Use the "clean-install" script for a nicer Dockerfile.

Fixes #8767

----

Here is the diff, from `container-diff`:

```
-----File-----

These entries have been added to gcr.io/k8s-minikube/kicbase:v0.0.12-snapshot3:
FILE                                        SIZE
/usr/bin/crictl                             20.8M
/var/lib/dpkg/info/cri-tools.list           263B
/var/lib/dpkg/info/cri-tools.md5sums        325B

These entries have been deleted from gcr.io/k8s-minikube/kicbase:v0.0.12-snapshot3:
FILE                                          SIZE
/usr/local/bin/containerd                     52.8M
/usr/local/bin/containerd-shim                7.1M
/usr/local/bin/containerd-shim-runc-v2        8.7M
/usr/local/bin/crictl                         27.1M
/usr/local/bin/ctr                            26.3M
/usr/local/sbin/runc                          9.4M

These entries have been changed between gcr.io/k8s-minikube/kicbase:v0.0.12-snapshot3 and local/kicbase:v0.0.12-snapshot3-snapshot:
FILE                                                          SIZE1        SIZE2
/usr/lib/cri-o-runc/sbin/runc                                 9.4M         7.1M
/opt/cni/bin/dhcp                                             8.6M         8.9M
/opt/cni/bin/firewall                                         4.2M         4.2M
/opt/cni/bin/bridge                                           3.2M         3.2M
/opt/cni/bin/ptp                                              3.2M         3.2M
/opt/cni/bin/macvlan                                          3M           3.1M
/opt/cni/bin/ipvlan                                           3M           3M
/opt/cni/bin/vlan                                             3M           3M
/opt/cni/bin/bandwidth                                        2.9M         2.9M
/opt/cni/bin/host-device                                      2.9M         2.9M
/opt/cni/bin/portmap                                          2.7M         2.7M
/opt/cni/bin/host-local                                       2.5M         2.5M
/opt/cni/bin/sbr                                              2.3M         2.5M
/opt/cni/bin/tuning                                           2.3M         2.5M
/opt/cni/bin/loopback                                         2.2M         2.4M
/opt/cni/bin/flannel                                          2.1M         2.4M
/opt/cni/bin/static                                           2M           2.2M
/var/lib/dpkg/status                                          187K         187.3K
/var/lib/dpkg/status-old                                      187K         187.3K
[...]

PR #9330, opened 2020-09-27
Base kicbase directly on ubuntu, without kindbase

### LINKED ISSUE #7788: move off kindbase image

afbjorklund:
Here is the all-in-one version, that includes **all** the changes from PR #9135 and PR #9225 - but without the interim steps.

Closes #7788


deploy/kicbase/Dockerfile
```diff
@@ -1,44 +1,125 @@
...
+# limitations under the License.
+
+# kind node base image
+#
+# For systemd + docker configuration used below, see the following references:
+# https://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/
+
+# start from ubuntu 20.04, this image is reasonably small as a starting point
+# for a kubernetes node image, it doesn't contain much we don't need
+FROM ubuntu:focal-20200423
```
medyagh, line 22: either change to "FROM ubuntu:focal-20200423"
the 'maket build-docker-image' needs "as base"

or remove it from the make target

```
	docker build -f ./deploy/kicbase/Dockerfile -t local/kicbase:$(KIC_VERSION)-snapshot  --build-arg COMMIT_SHA=${VERSION}-$(COMMIT) --cache-from $(KIC_BASE_IMAGE_GCR) --target base ./deploy/kicbase

```

medyagh (changes requested):

medyagh (approved):

PR #9332, opened 2020-09-27
Don't trim whitespace while extracting strings

### LINKED ISSUE #7481: Extracting translations strips whitespace
This means that the original string now **won't** be translated, since the whitespace was removed.

So in order for `T` to work, we need to keep the original whitespace in the strings to be translated.

afbjorklund:
Remove some leading and trailing whitespace from non-string
constructs like description and automatic extraction filters.

But don't change all the translations or the translations here,
since that would make this code change impossible to review.

There are lots of translations missing at the moment:
```
Writing to de.json (51 translated, 683 untranslated)
Writing to es.json (51 translated, 683 untranslated)
Writing to fr.json (162 translated, 572 untranslated)
Writing to ja.json (283 translated, 451 untranslated)
Writing to ko.json (139 translated, 595 untranslated)
Writing to pl.json (82 translated, 652 untranslated)
Writing to zh-CN.json (214 translated, 520 untranslated)
```

Closes #7481


medyagh (approved):

afbjorklund:
I think we need a follow-up to this one, to "save" some of the white-space only translations that got "lost" now

afbjorklund:
But at first overview, it seemed that very few of the affected strings were actually translated in the first place...

PR #9340, opened 2020-09-28
Added target for Windows HTML & switched encoding for Windows Report

blueelvis:
1. Added the make target for Windows HTML Test Report.
2. Switched encoding to UTF8 instead of ASCII which was causing issues while parsing in Git Bash.

@medyagh - I wonder if we require the `open HTML report` command. Cannot we just remove this and just echo that the HTML report is located over there because `open` doesn't exist on Windows and Git Bash? This make target will work in Git Bash. If we need this makefile to work in PowerShell, I would suggest a separate makefile. We can discuss this in today's call.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -362,7 +362,7 @@ jobs:
           $env:KUBECONFIG="${pwd}\testhome\kubeconfig"
           $env:MINIKUBE_HOME="${pwd}\testhome"
           $ErrorActionPreference = "SilentlyContinue"
-          .\e2e-windows-amd64.exe --minikube-start-args="--driver=docker" --test.timeout=15m --timeout-multiplier=1 --test.v --test.run=TestFunctional --binary=./minikube-windows-amd64.exe | Out-File -Encoding utf8 -FilePath ".\report\testout.txt"
+          .\e2e-windows-amd64.exe --minikube-start-args="--driver=docker" --test.timeout=15m --timeout-multiplier=1 --test.v --test.run=TestFunctional --binary=./minikube-windows-amd64.exe | Tee-Object -FilePath ".\report\testout.txt"
           $END_TIME=(GET-DATE)
           echo $END_TIME
           $DURATION=(NEW-TIMESPAN -Start $START_TIME -End $END_TIME)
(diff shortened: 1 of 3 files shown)


.github/workflows/master.yml
```diff
@@ -362,7 +362,7 @@ jobs:
           $env:KUBECONFIG="${pwd}\testhome\kubeconfig"
           $env:MINIKUBE_HOME="${pwd}\testhome"
           $ErrorActionPreference = "SilentlyContinue"
-          .\e2e-windows-amd64.exe --minikube-start-args="--driver=docker" --test.timeout=15m --timeout-multiplier=1 --test.v --test.run=TestFunctional --binary=./minikube-windows-amd64.exe | Tee-Object -FilePath ".\report\testout.txt"
+          .\e2e-windows-amd64.exe --minikube-start-args="--driver=docker" --test.timeout=15m --timeout-multiplier=1 --test.v --test.run=TestFunctional --binary=./minikube-windows-amd64.exe | Out-File -Encoding utf8 -FilePath ".\report\testout.txt"
```
medyagh, line 365: revert unrelated changes t this PR.
I recall using out-file and I had a bug that would take long time to get the file or somehting would get courrputed. unless I am convinecved that Tee-Object is clearly wrong I would like this change to be reverted.

Makefile
```diff
@@ -309,6 +309,15 @@ html_report: ## Generate HTML  report out of the last ran integration test logs.
 	@echo "-----------------------------------------------------------------------------------"
 	@open $(CURDIR)/out/testout_$(COMMIT_SHORT).html || true
 
+.PHONY: html_report_windows
+html_report_windows: ## Generate the HTML Report out of the last ran integration test logs on Windows.
```
medyagh, line 313: can we make this automatic ?
like if we call make html_report and it is in windows, we automatically call this target instead ?

medyagh (changes requested):

Makefile
blueelvis: Done. Works now on both Linux and Windows and uses the `GOOS` environment variable.

medyagh (approved):

medyagh (approved):

Issue #9317, opened 2020-09-24
"unable to download gcr.io" but downloads gcr.io


medyagh:
* Starting control plane node minikube in cluster minikube
* Pulling base image ...
! minikube was unable to download gcr.io/k8s-minikube/kicbase:v0.0.12-snapshot3, but successfully downloaded gcr.io/k8s-minikube/kicbase:v0.0.12-snapshot3 as a fallback image
* Creating docker container (CPUs=2, Memory=4000MB)

afbjorklund:
Think it is just the output that is not showing the `cc.KubernetesConfig.ImageRepository` replacement ?

It should have been "baseImg".

afbjorklund:
It could also be that the cast to Tag is hiding any digest that could have been in the original Ref ?

```go
                        if finalImg != "" && finalImg != baseImg {
                                out.WarningT(fmt.Sprintf("minikube was unable to download %s, but successfully downloaded %s as a fallback image", image.Tag(cc.KicBaseImage), image.Tag(finalImg)))
                                cc.KicBaseImage = finalImg
                        }
```

Would be interesting to see some logs, to see what is going on.

PR #9348, opened 2020-09-28
Compare fully qualified image names when pulling kic base image

priyawadhwa:
Tested and passes on Cloud Shell.

Fixes a small bug where we were comparing one fully qualified image (tag+digest) to an image that had only tag specified. This resulted in an unnecessary warning, which only showed up when the base image was loaded from a tarball. This happens on Cloud Shell.

fixes https://github.com/kubernetes/minikube/issues/9317


medyagh (approved):
thanks

Issue #8775, opened 2020-07-20
avoid re-download re-install hyperkit/hyperv drivers no minikube new version


medyagh:
if drivers didn't change, we shouldn't download a new one with only commit sha of the version change and make user enter their password each time.

tstromberg:
A simple alternative would be to simply encode the minimum hyperkit driver version.

AFAIK, the only drivers we download currently is hyperkit and kvm2, but we should probably extend this to vmware and parallels in the future.

priyawadhwa:
Closing as a dupe of https://github.com/kubernetes/minikube/issues/7422

Issue #7422, opened 2020-04-05
hyperkit driver should be happy with current minimum verison


medyagh:
we have not updated the hyperkit driver in a year and we still each time we update minikube we redownload it and we ask user to give sudo persmiion for the user group.

currnetly with each new version of minikube, we re-download the hyperkit driver. even if the code for it was not touched.

the hyperkit driver udpate should not download the driver if the checksum is not changed.

simmilar to preload or other things in minikube that do not redownload.

medyagh:
CC @josedonizetti

tstromberg:
Here's another thought: this driver almost never requires updates anymore. Maybe once or twice a year.

How about encoding a minimum version and only upgrading it if necessary?

medyagh:
we still like to do this

medyagh:
@ilya-zuyev this is yours.

PR #9365, opened 2020-09-30
hyperkit driver should be happy with current minimum verison

ilya-zuyev:
This PR adds an ability to keep using the installed hyperkit driver after minikube upgrade,  if the drives is capable with the new version of minikube

Fixes https://github.com/kubernetes/minikube/issues/7422

The PR adds a new function `minDriverVersion(dirver, mkVersion)` to get the lowest version of  driver `driver` compatible with the current code. 

The minimum supported hyperkit driver version is set to `1.11.0`

## Before: 
```shell
izuyev@izuyev-macbookpro --- ~ » ~/.minikube/bin/docker-machine-driver-hyperkit version                                                                                                       
version: v1.11.0
commit: 57e2f55f47effe9ce396cea42a1e0eb4f611ebbd

izuyev@izuyev-macbookpro --- ~ » minikube version
minikube version: v1.13.1
commit: 111588c7f387d4244be053e71c930fc00af29323

izuyev@izuyev-macbookpro --- ~ » minikube start --driver=hyperkit
* minikube v1.13.1 on Darwin 10.15.6
* Using the hyperkit driver based on existing profile
* Downloading driver docker-machine-driver-hyperkit:
    > docker-machine-driver-hyperkit.sha256: 65 B / 65 B [---] 100.00% ? p/s 0s
    > docker-machine-driver-hyperkit: 10.90 MiB / 10.90 MiB  100.00% 7.94 MiB p
* The 'hyperkit' driver requires elevated permissions. The following commands will be executed:

    $ sudo chown root:wheel /Users/izuyev/.minikube/bin/docker-machine-driver-hyperkit 
    $ sudo chmod u+s /Users/izuyev/.minikube/bin/docker-machine-driver-hyperkit 


Password: 
* Updating the running hyperkit "minikube" VM ...
* Preparing Kubernetes v1.19.2 on Docker 19.03.12 ...
* Enabled addons: storage-provisioner, default-storageclass
* Done! kubectl is now configured to use "minikube" by default

izuyev@izuyev-macbookpro --- ~ » ~/.minikube/bin/docker-machine-driver-hyperkit version
version: v1.13.1
commit: 1fd1f67f338cbab4b3e5a6e4c71c551f522ca138
```

### 1.11.0 -> 1.13.1 The driver is NOT upgraded
```shell
izuyev@izuyev-macbookpro --- ~ » ~/.minikube/bin/docker-machine-driver-hyperkit version
version: v1.11.0
commit: 57e2f55f47effe9ce396cea42a1e0eb4f611ebbd

izuyev@izuyev-macbookpro --- ~ » minikube version 
minikube version: v1.13.1
commit: 7322fdd5377e66c0d37d6d8b9d6b7972c5c39414

izuyev@izuyev-macbookpro --- ~ » minikube start --driver=hyperkit
* minikube v1.13.1 on Darwin 10.15.6
* Using the hyperkit driver based on existing profile
* Restarting existing hyperkit VM for "minikube" ...
* Preparing Kubernetes v1.19.2 on Docker 19.03.12 ...
* Enabled addons: default-storageclass, storage-provisioner
* Done! kubectl is now configured to use "minikube" by default

izuyev@izuyev-macbookpro --- ~ » ~/.minikube/bin/docker-machine-driver-hyperkit version
version: v1.11.0
commit: 57e2f55f47effe9ce396cea42a1e0eb4f611ebbd
```
### 1.2.0 -> 1.13.1 The driver IS upgraded
```shell
izuyev@izuyev-macbookpro --- ~ » ~/.minikube/bin/docker-machine-driver-hyperkit version
version: v1.2.0
commit: ea999167a736ddb5b5a3161e614e30fb88e91dd4-dirty

izuyev@izuyev-macbookpro --- ~ » minikube version                                      
minikube version: v1.13.1
commit: 111588c7f387d4244be053e71c930fc00af29323

izuyev@izuyev-macbookpro --- ~ » minikube start --driver=hyperkit
* minikube v1.13.1 on Darwin 10.15.6
* Using the hyperkit driver based on existing profile
* Downloading driver docker-machine-driver-hyperkit:
    > docker-machine-driver-hyperkit.sha256: 65 B / 65 B [---] 100.00% ? p/s 0s
    > docker-machine-driver-hyperkit: 10.90 MiB / 10.90 MiB  100.00% 8.10 MiB p
* The 'hyperkit' driver requires elevated permissions. The following commands will be executed:

    $ sudo chown root:wheel /Users/izuyev/.minikube/bin/docker-machine-driver-hyperkit 
    $ sudo chmod u+s /Users/izuyev/.minikube/bin/docker-machine-driver-hyperkit 


Password: 
* Restarting existing hyperkit VM for "minikube" ...
* Preparing Kubernetes v1.19.2 on Docker 19.03.12 ...
* Enabled addons: storage-provisioner, default-storageclass
* Done! kubectl is now configured to use "minikube" by default

izuyev@izuyev-macbookpro --- ~ » ~/.minikube/bin/docker-machine-driver-hyperkit version
version: v1.13.1
commit: 1fd1f67f338cbab4b3e5a6e4c71c551f522ca138
```

closes

https://github.com/kubernetes/minikube/issues/8775


pkg/minikube/driver/version.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/blang/semver"
+	"github.com/golang/glog"
+)
+
+// minHyperkitVersion is the minimum version of the minikube hyperkit driver compatible with the current minikube code
+var minHyperkitVersion semver.Version
+
+const minHyperkitVersionStr = "1.10.0"
```
medyagh, line 11: maybe 1.11.0

pkg/minikube/driver/version.go
```diff
@@ -0,0 +1,31 @@
...
+
+func init() {
+	v, err := semver.New(minHyperkitVersionStr)
+	if err != nil {
+		glog.Fatalf("Failed to parse the hyperkit driver version: %v", err)
```
medyagh, line 16: lets not fatal and if severErr is not nil lets just download the latest version of driver as we do today.

pkg/minikube/driver/version_test.go
```diff
@@ -0,0 +1,36 @@
...
+		want   semver.Version
+	}{
+		{"Hyperkit", HyperKit, "1.1.1", minHyperkitVersion},
+		{"Invalid", "_invalid_", "1.1.1", v("1.1.1")},
+		{"KVM2", KVM2, "1.1.1", v("1.1.1")},
+	}
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			if got := minDriverVersion(tt.driver, v(tt.mkV)); !got.EQ(tt.want) {
+				t.Errorf("minDriverVersion() = %v, want %v", got, tt.want)
```
medyagh, line 24: wording to match other style of unit test

pkg/minikube/driver/version.go
ilya-zuyev: Done

pkg/minikube/driver/version.go
ilya-zuyev: Makes sense. Fixed.

pkg/minikube/driver/version_test.go
ilya-zuyev: Done

test/integration/driver_install_or_update_test.go
```diff
@@ -173,3 +174,85 @@ func TestHyperKitDriverInstallOrUpdate(t *testing.T) {
...
+
+			if err = os.Setenv("PATH", filepath.Dir(driverPath)); err != nil {
+				t.Fatalf("Unexpected error on SetEn, got: %v", err)
+			}
+			curVersion, err := version.GetSemverVersion()
+			if err != nil {
+				t.Fatalf("Expected new semver. test: %v, got: %v", tc.name, err)
+			}
+
+			if _, err = driver.AssureDriverBinary("hyperkit", dir, curVersion, true); err != nil {
```
medyagh, line 242: how about instead of using "driver.AssureDriverBinary" use
```
minikube start --download-only --driver=hyperkit
```
similar to 
https://github.com/medyagh/minikube/blob/32922f4184c31fb405fb8dab7a1d7a8ef120b47a/test/integration/aaa_download_only_test.go#L61

test/integration/driver_install_or_update_test.go
```diff
@@ -173,3 +174,85 @@ func TestHyperKitDriverInstallOrUpdate(t *testing.T) {
...
+			if _, err = os.Stat(driverPath); err != nil {
+				t.Fatalf("Expected driver to exist. test: %s, got: %v", tc.name, err)
+			}
+
+			// change permission to allow driver to be executable
+			if err = os.Chmod(driverPath, 0700); err != nil {
+				t.Fatalf("Expected not expected when changing driver permission. test: %s, got: %v", tc.name, err)
+			}
+
+			if err = os.Setenv("PATH", filepath.Dir(driverPath)); err != nil {
```
medyagh, line 234: instead of setting Env variable for the whole test (that could conflict the other parallel tests) 
consider using the env only for the Comand
for example
```
	cmd.Env = append(os.Environ(), "DOCKER_HOST=/does/not/exist")
	if _, err := Run(t, cmd); err != nil {
//...
	}

```
here is an example here https://github.com/medyagh/minikube/blob/32922f4184c31fb405fb8dab7a1d7a8ef120b47a/test/integration/aaa_download_only_test.go#L170

pkg/minikube/driver/version.go
```diff
@@ -0,0 +1,50 @@
...
+
+func init() {
+	v, err := semver.New(minHyperkitVersionStr)
+	if err != nil {
+		glog.Errorf("Failed to parse the hyperkit driver version: %v", err)
+	} else {
+		minHyperkitVersion = v
+	}
+}
+func minDriverVersion(driver string, mkVer semver.Version) semver.Version {
```
medyagh, line 37: add comment here, consider more descriptive name. maybe

minAcceptableVersion

medyagh (changes requested):

pkg/minikube/driver/version.go
ilya-zuyev: Done

test/integration/driver_install_or_update_test.go
ilya-zuyev: When `minikube start --download-only --driver=hyperkit` actually downloads the driver it tries to set driver permission as well, asking for sudo.  `make integration` is failing in this case. 
Cannot run as `sudo make integration` too, because of 
`X Exiting due to DRV_AS_ROOT: The "hyperkit" driver should not be used with root privileges.`

Another issue is that `minikube start --download-only --driver=hyperkit` downloads not only the driver, but also preloaded-images archive and minikube.iso. It's ~700Mb per test launch.

priyawadhwa: We have at least one integration test that runs `--download-only` (`TestDownloadOnly`) to make sure it's working properly or uses it as part of a larger test. I'm guessing all of these tests would fail on hyperkit if this doesn't work?

ilya-zuyev: Looks like my issue is specific to the external drivers only. 
`TestDownloadOnly` uses the default driver, which is usually docker. In this case minikube does not try to install an external driver binary and skips the sudo step. 

I tried to forcibly use hyperkit in `TestDownloadOnly`  passing `TEST_ARGS="-minikube-start-args --driver=hyperkit" `  and got the same error:
```
izuyev@izuyev-macbookpro --- z/minikube ‹gh_7422-reuse_hyperkit_driver* M› » make integration TEST_ARGS="-run TestDownloadOnly -minikube-start-args --driver=hyperkit"    
go test -ldflags="-X k8s.io/minikube/pkg/version.version=v1.13.1 -X k8s.io/minikube/pkg/version.isoVersion=v1.13.1 -X k8s.io/minikube/pkg/version.isoPath=minikube/iso -X k8s.io/minikube/pkg/version.gitCommitID="08da82702d38d646df313d793ab9ea0f1a4fa46d-dirty" -X k8s.io/minikube/pkg/version.storageProvisionerVersion=v3" -v -test.timeout=90m ./test/integration --tags="integration container_image_ostree_stub containers_image_openpgp go_getter_nos3 go_getter_nogcs" -run TestDownloadOnly -minikube-start-args --driver=hyperkit 2>&1 | tee "./out/testout_08da82702.txt"
ERROR: logging before flag.Parse: I1012 13:09:00.486559   47903 translate.go:64] Getting system locale failed: Could not detect Language
ERROR: logging before flag.Parse: I1012 13:09:00.486698   47903 translate.go:106] Setting Language to en-US ...
Found 12 cores, limiting parallelism with --test.parallel=6
=== RUN   TestDownloadOnly
=== RUN   TestDownloadOnly/crio
=== RUN   TestDownloadOnly/crio/v1.13.0
    aaa_download_only_test.go:65: (dbg) Run:  out/minikube start --download-only -p crio-20201012130900-47903 --force --alsologtostderr --kubernetes-version=v1.13.0 --container-runtime=crio --driver=hyperkit
Password: 

[test infinitely waits for password]
```

priyawadhwa: I'm not super familiar with this part of the code, but do you know if we try to set the permissions before asking for root access?

My guess is that we don't need a password to set the permissions in our testing environment. If it doesn't already work like this, we could try to set the driver permissions first and then ask for root access if that fails.

ilya-zuyev: Yep, I checked CI scripts, looks like out test environment just allows sudo without a password.

Updated the test to skip if sudo needs a password.

test/integration/driver_install_or_update_test.go
ilya-zuyev: Done

medyagh (changes requested):
lets rename  test/integration/driver_install_or_update_darwin_test.go
to something shorter
maybe driver_download
_test

test/integration/driver_install_or_update_darwin_test.go
```diff
@@ -0,0 +1,124 @@
...
+				t.Fatalf("Failed to prepare tempdir. test: %s, got: %v", tc.name, err)
+			}
+			defer func() {
+				if err := os.RemoveAll(mkDir); err != nil {
+					t.Errorf("Failed to remove mkDir %q: %v", mkDir, err)
+				}
+			}()
+
+			// start "minikube start --download-only --interactive=false --driver=hyperkit --preload=false"
+			cmd := exec.Command(Target(), "start", "--download-only", "--interactive=false", "--driver=hyperkit", "--preload=false")
```
medyagh, line 53: what is the reqson for setting preload=false here?

medyagh, line 53: what is reason for setting `--preload=false` here ? if not related remove that flag

test/integration/driver_install_or_update_test.go
```diff
@@ -173,3 +173,8 @@ func TestHyperKitDriverInstallOrUpdate(t *testing.T) {
 		}
 	}
 }
+
+func sudoNeedsPassword() bool {
+	err := exec.Command("sudo", "-n", "ls").Run()
```
medyagh, line 178: this wont work on windows, I recommend skipping this test for non-darwin

test/integration/driver_install_or_update_darwin_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	var resultVersion string
+	_, err = fmt.Sscanf(string(output), "version: %s\n", &resultVersion)
+	if err != nil {
+		return "", err
+	}
+	return resultVersion, nil
+}
+
+func tempMinikubeDir(name, driver string) (string, string, error) {
```
medyagh, line 91: this function needs comment  and it seems to be doing way more than its name, please change name to something describing of what it is doing.

test/integration/driver_install_or_update_darwin_test.go
```diff
@@ -0,0 +1,124 @@
...
+		return "", "", fmt.Errorf("expected driver to exist: %v", err)
+	}
+
+	// copy driver to temp bin
+	testDriverPath := filepath.Join(mkBinDir, "docker-machine-driver-hyperkit")
+	if err = ioext.CopyFile(testDataDriverPath, testDriverPath, false); err != nil {
+		return "", "", fmt.Errorf("failed to setup current hyperkit driver: %v", err)
+	}
+	// change permission to allow driver to be executable
+	if err = os.Chmod(testDriverPath, 0777); err != nil {
```
medyagh, line 120: is this same chmod number we have on our driver binary ? if not reduce from  755 to what we have in reality.

medyagh (changes requested):

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.13
 
 require (
 	cloud.google.com/go/storage v1.8.0
+	github.com/Azure/azure-sdk-for-go v42.3.0+incompatible
```
medyagh, line 7: is this change related?  to verify run "go mod tidy" before the PR

ilya-zuyev:
> lets rename test/integration/driver_install_or_update_darwin_test.go
> to something shorter
> maybe driver_download
> _test

`_darwin_` part is essential. It tells the go test tool to ignore the test on any platform other than Darwin. An alternative is an explicit check in test code:
```go
if runtime.GOOS != "darwin" {
		t.Skip("Skip if not darwin.")
}```
IMO, the filename based approach is cleaner and also reduces the overall test suite execution time.

File: go.mod (hunk 11 above)
Comment (ilya-zuyev) [reply go.mod:7]: Yes, I use `ioext.CopyFile()` from this module. We'd had an indirect dependency to this module, so I used it.

File: test/integration/driver_install_or_update_test.go (hunk 8 above)
Comment (ilya-zuyev) [reply test/integration/driver_install_or_update_test.go:178]: This function is used only in tests related to hyperkit driver. They don't run on windows.

File: test/integration/driver_install_or_update_darwin_test.go (hunk 7 above)
Comment (ilya-zuyev) [reply test/integration/driver_install_or_update_darwin_test.go:53]: Just an optimization. Downloading the preload tarball is pretty expensive in terms of time and traffic. This flag makes the test a bit faster.

File: test/integration/driver_install_or_update_darwin_test.go (hunk 10 above)
Comment (ilya-zuyev) [reply test/integration/driver_install_or_update_darwin_test.go:201]: Yep, good point! Done.

File: test/integration/driver_install_or_update_darwin_test.go (hunk 9 above)
Comment (ilya-zuyev) [reply test/integration/driver_install_or_update_darwin_test.go]: Done. Renamed to `prepareTempMinikubeDirWithHyperkitDriver` + added a comment

File: test/integration/driver_install_or_update_darwin_test.go (hunk 7 above)
Comment (medyagh) [reply test/integration/driver_install_or_update_darwin_test.go:139]: when setting preload=false it will still download those things except they are not preloaded (slightly bigger tar file) but TestDownloadOnly runs before all the tests and downloads the preload tar file so this test will not redownload it
but if you set preload=false the extra images will be downloaded here:

 ls ~/.minikube/cache/images/k8s.gcr.io/

File: test/integration/driver_install_or_update_darwin_test.go [hunk 12]
```diff
@@ -0,0 +1,221 @@
+/*
+Copyright 2020 The Kubernetes Authors All rights reserved.
```
medyagh, line 2: I still think we should use shorter filename and also youse T.Skip

this will give us visibiliity how many tests were skipped in the gopogh report
for example when running on windows, we will see a list of skipped tests so the tester would be more aware of what they are not testing on this platform.

how about driver_download_test.go ?

test/integration/driver_install_or_update_darwin_test.go
medyagh: dont see the updated code here

medyagh (changes requested):

test/integration/driver_install_or_update_test.go
```diff
@@ -173,3 +175,128 @@ func TestHyperKitDriverInstallOrUpdate(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("Failed to prepare tempdir. test: %s, got: %v", tc.name, err)
+			}
+			defer func() {
+				if err := os.RemoveAll(mkDir); err != nil {
+					t.Errorf("Failed to remove mkDir %q: %v", mkDir, err)
+				}
+			}()
+
+			// start "minikube start --download-only --interactive=false --driver=hyperkit --preload=false"
```
medyagh, line 219: get rid of these comment

medyagh (changes requested):

medyagh (approved):
looks good, whenever you feel comfortable with a last touch

test/integration/driver_install_or_update_test.go
ilya-zuyev: Done

test/integration/driver_install_or_update_darwin_test.go
ilya-zuyev: Returned T.Skip checks and moved my tests to driver_install_or_update_tst.go. It's an existent has and has other tests related to driver upgrades, so probably makes sense to keep all such tests in one place

test/integration/driver_install_or_update_darwin_test.go
ilya-zuyev: oops, my bad, changed the wrong line. Fixed

tstromberg (approved):

pkg/minikube/driver/install.go
```diff
@@ -133,6 +132,7 @@ func validateDriver(executable string, v semver.Version) (string, error) {
 	if err != nil {
 		return path, errors.Wrap(err, "can't parse driver version")
 	}
+
```
tstromberg, line 135: Now that we allow version mismatches, could you please add a log sttatement declaring what driver version we are using? This will help for future debugging;

`klog.Infof("%s version is %s", path, driverVersion)`

Thanks!

ilya-zuyev: Sure. Done.

medyagh:
thank you

Issue #3760, opened 2019-02-26
none: minikube should be able to be run by a non-root user (use sudo when necessary)


sourabh1031:
If this is a bug report, please include:

I tried starting minikube with the following  command 
minikube start --vm-driver=none

after downloading minikube v0.34.1 from this command 
curl -Lo minikube https://storage.googleapis.com/minikube/releases/v0.34.1/minikube-linux-amd64 && chmod +x minikube && sudo cp minikube /usr/local/bin/ && rm minikube

I got the output that minikube crashed

I am on ubuntu 18.04LTS machine

hpandeycodeit:
@sourabh1031 can you provide the full output/error that you got when it crashed?

sourabh1031:
minikube start --vm-driver=none
😄  minikube v0.34.1 on linux (amd64)
🤹  Configuring local host environment ...

⚠️  The 'none' driver provides limited isolation and may reduce system security and reliability.
⚠️  For more information, see:
👉  https://github.com/kubernetes/minikube/blob/master/docs/vmdriver-none.md

💡  Tip: Use 'minikube start -p <name>' to create a new cluster, or 'minikube delete' to delete this one.
🔄  Restarting existing none VM for "minikube" ...
⌛  Waiting for SSH access ...
📶  "minikube" IP address is 192.168.43.56
🐳  Configuring Docker as the container runtime ...
[sudo] password for sourabh: 
✨  Preparing Kubernetes environment ...
💣  Failed to update cluster: downloading binaries: copy: error removing file /usr/bin/kubelet: remove /usr/bin/kubelet: permission denied

😿  Sorry that minikube crashed. If this was unexpected, we would love to hear from you:
👉  https://github.com/kubernetes/minikube/issues/new

sourabh1031:
when i ran command using sudo i got this 
[sudo] password for sourabh: 
😄  minikube v0.34.1 on linux (amd64)
🤹  Configuring local host environment ...

⚠️  The 'none' driver provides limited isolation and may reduce system security and reliability.
⚠️  For more information, see:
👉  https://github.com/kubernetes/minikube/blob/master/docs/vmdriver-none.md

⚠️  kubectl and minikube configuration will be stored in /home/sourabh
⚠️  To use kubectl or minikube commands as your own user, you may
⚠️  need to relocate them. For example, to overwrite your own settings:

    ▪ sudo mv /home/sourabh/.kube /home/sourabh/.minikube $HOME
[...]

PR #9379, opened 2020-10-02
Make none driver work as regular user (use sudo on demand)

Git-Jiro:
Ref https://github.com/kubernetes/minikube/issues/3760

This almost works:
`minikube start --vm-driver=none --kubernetes-version=1.17.0`

>
> * minikube v1.13.1 on Ubuntu 18.04
> * Using the none driver based on user configuration
> * Starting control plane node minikube in cluster minikube
> * Running on localhost (CPUs=4, Memory=7449MB, Disk=38545MB) ...
> * OS release is Ubuntu 18.04.5 LTS
> * Preparing Kubernetes v1.17.0 on Docker 19.03.13 ...
>   - kubelet.resolv-conf=/run/systemd/resolve/resolv.conf
> ! initialization failed, will try again: run: /bin/bash -c "sudo env PATH=/var/lib/minikube/binaries/v1.17.0:$PATH kubeadminit > --config /var/tmp/minikube/kubeadm.yaml  --ignore-preflight-errors=DirAvailable--etc-kubernetes-manifests,DirAvailable--var-lib-minikube,DirAvailable--var-lib-minikube-etcd,FileAvailable--etc- kubernetes-manifests-kube-scheduler.yaml,FileAvailable--etc-kubernetes-manifests-kube-ap
iserver.yaml,FileAvailable--etc-kubernetes-manifests-kube-controller-manager.yaml,FileAvailable--etc-kubernetes-manifests-etcd.yaml,Port-10250,Swap": exit status 126
stdout:
> 
> stderr:
> env: ‘kubeadm’: Permission denied`

`sudo chmod 711 /var/lib/minikube/binaries/v1.17.0/*`
`minikube start --vm-driver=none --kubernetes-version=1.17.0`
`minikube kubectl -- get pods -A`

So the only open issue is how to properly set those permissions on /var/lib/minikube/binaries/v1.17.0/*

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -786,10 +786,6 @@ func validateUser(drvName string) {
 
 	useForce := viper.GetBool(force)
 
-	if driver.NeedsRoot(drvName) && u.Uid != "0" && !useForce {
-		exit.Message(reason.DrvNeedsRoot, `The "{{.driver_name}}" driver requires root privileges. Please run minikube using 'sudo -E minikube start --driver={{.driver_name}}'.`, out.V{"driver_name": drvName})
-	}
-
 	// None driver works with root and without root on Linux
 	if runtime.GOOS == "linux" && drvName == driver.None {
 		if !viper.GetBool(interactive) {
@@ -801,8 +797,8 @@ func validateUser(drvName string) {
 		return
 	}
 
-	// If root is required, or we are not root, exit early
-	if driver.NeedsRoot(drvName) || u.Uid != "0" {
+	// If we are not root, exit early
+	if u.Uid != "0" {
 		return
 	}
 
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -790,6 +790,17 @@ func validateUser(drvName string) {
 		exit.Message(reason.DrvNeedsRoot, `The "{{.driver_name}}" driver requires root privileges. Please run minikube using 'sudo -E minikube start --driver={{.driver_name}}'.`, out.V{"driver_name": drvName})
 	}
 
+	// None driver works with root and without root on Linux
+	if runtime.GOOS == "linux" && drvName == driver.None {
+		if !viper.GetBool(interactive) {
+			test := exec.Command("sudo", "-n", "echo", "-n")
+			if err := test.Run(); err != nil {
+				exit.Message(reason.DrvNeedsRoot, `sudo requires a password, and --interactive=false`)
+			}
+		}
+		return
+	}
+
 	// If root is required, or we are not root, exit early
 	if driver.NeedsRoot(drvName) || u.Uid != "0" {
 		return
--- a/pkg/minikube/command/exec_runner.go
+++ b/pkg/minikube/command/exec_runner.go
@@ -20,10 +20,12 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"time"
 
@@ -85,11 +87,11 @@ func (*execRunner) RunCmd(cmd *exec.Cmd) (*RunResult, error) {
 }
 
 // Copy copies a file and its permissions
-func (*execRunner) Copy(f assets.CopyableFile) error {
+func (e *execRunner) Copy(f assets.CopyableFile) error {
 	dst := path.Join(f.GetTargetDir(), f.GetTargetName())
 	if _, err := os.Stat(dst); err == nil {
 		klog.Infof("found %s, removing ...", dst)
-		if err := os.Remove(dst); err != nil {
+		if err := e.Remove(f); err != nil {
 			return errors.Wrapf(err, "error removing file %s", dst)
 		}
 	}
@@ -105,12 +107,47 @@ func (*execRunner) Copy(f assets.CopyableFile) error {
 		return errors.Wrapf(err, "error converting permissions %s to integer", f.GetPermissions())
 	}
 
-	return writeFile(dst, f, os.FileMode(perms))
+	switch runtime.GOOS {
+	case "linux":
+		// write to temp location ...
+		tmpfile, err := ioutil.TempFile("", "minikube")
+		if err != nil {
+			return errors.Wrapf(err, "error creating tempfile")
+		}
+		defer os.Remove(tmpfile.Name())
+		err = writeFile(tmpfile.Name(), f, os.FileMode(perms))
+		if err != nil {
+			return errors.Wrapf(err, "error writing to tempfile %s", tmpfile.Name())
+		}
+
+		// ... then use sudo to move to target ...
+		_, err = e.RunCmd(exec.Command("sudo", "cp", "-a", tmpfile.Name(), dst))
+		if err != nil {
+			return errors.Wrapf(err, "error copying tempfile %s to dst %s", tmpfile.Name(), dst)
+		}
+
+		// ... then fix file permission that should have been fine because of "cp -a"
+		err = os.Chmod(dst, os.FileMode(perms))
+		return err
+
+	default:
+		return writeFile(dst, f, os.FileMode(perms))
+	}
+
 }
 
 // Remove removes a file
-func (*execRunner) Remove(f assets.CopyableFile) error {
+func (e *execRunner) Remove(f assets.CopyableFile) error {
 	dst := filepath.Join(f.GetTargetDir(), f.GetTargetName())
 	klog.Infof("rm: %s", dst)
-	return os.Remove(dst)
+	if err := os.Remove(dst); err != nil {
+		if !os.IsPermission(err) {
+			return err
+		}
+		_, err = e.RunCmd(exec.Command("sudo", "rm", "-f", dst))
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
--- a/pkg/minikube/driver/driver.go
+++ b/pkg/minikube/driver/driver.go
@@ -143,7 +143,7 @@ func BareMetal(name string) bool {
 
 // NeedsRoot returns true if driver needs to run with root privileges
 func NeedsRoot(name string) bool {
-	return name == None
+	return false
 }
 
 // NeedsPortForward returns true if driver is unable provide direct IP connectivity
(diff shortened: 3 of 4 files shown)


Git-Jiro:
Hi @tstromberg!
Now I can reproduce the same error i have locally with the functional_baremetal_ubuntu18_04 test.
Any ideas why functional_virtualbox_macos is always failing at the same test case?

medyagh (commented):
I know this is WIP I am excited for this PR ! thanks for starting it

@Git-Jiro

Git-Jiro:
@medyagh can you give me an ok-to-test?
Seems like I was able to test this PR without an ok-to-test label until now.

cmd/minikube/cmd/start.go
```diff
@@ -790,6 +790,11 @@ func validateUser(drvName string) {
 		exit.Message(reason.DrvNeedsRoot, `The "{{.driver_name}}" driver requires root privileges. Please run minikube using 'sudo -E minikube start --driver={{.driver_name}}'.`, out.V{"driver_name": drvName})
 	}
 
+	// None driver works with root and without root on Linux
+	if runtime.GOOS == "linux" && drvName == driver.None {
```
medyagh, line 794: this PR should check if --interactive flag is set to false (if set to false, means user can not enter anything through stdin and in that case we should exit and advice user that they either have to run with sudo or allow interactive

pkg/minikube/command/exec_runner.go
```diff
@@ -105,12 +114,44 @@ func (*execRunner) Copy(f assets.CopyableFile) error {
 		return errors.Wrapf(err, "error converting permissions %s to integer", f.GetPermissions())
 	}
 
-	return writeFile(dst, f, os.FileMode(perms))
+	switch runtime.GOOS {
+	case "linux":
```
medyagh, line 118: please comment what this code is trying to solve

medyagh (changes requested):

cmd/minikube/cmd/start.go
tstromberg: If it helps, there is some existing code you can copy here, that tests (with sudo -n) whether or not sudo requires a prompt, and then exits if interactive mode is disabled:

https://github.com/kubernetes/minikube/blob/5ce9d99d68a57f833b7af035c384479122479968/pkg/minikube/driver/install.go#L96

pkg/minikube/command/exec_runner.go
Git-Jiro: This code is trying to solve a problem with bad permissions of /var/lib/minikube/binaries/v1.17.0/* when minikube is run as normal user with the none driver:
When minikube is started as a normal user with the none driver, those files are still created as owned by root and mode 700. And then minikube will fail with "permission denied" (to execute the files there).
The code fixes this issue.

(There might be a better way to fix the issue, but I haven't found one yet.)

cmd/minikube/cmd/start.go
```diff
@@ -790,6 +790,17 @@ func validateUser(drvName string) {
...
+		if !viper.GetBool(interactive) {
+			test := exec.Command("sudo", "-n", "echo", "-n")
+			if err := test.Run(); err != nil {
+				exit.Message(reason.Usage, `sudo requires a password, and --interactive=false`)
```
tstromberg, line 798: Please use `reason.DrvNeedsRoot` - we have a specific error code that integrators can use to detect this condition.

pkg/minikube/command/exec_runner.go
```diff
@@ -85,12 +87,19 @@ func (*execRunner) RunCmd(cmd *exec.Cmd) (*RunResult, error) {
...
 // Copy copies a file and its permissions
-func (*execRunner) Copy(f assets.CopyableFile) error {
+func (e *execRunner) Copy(f assets.CopyableFile) error {
 	dst := path.Join(f.GetTargetDir(), f.GetTargetName())
 	if _, err := os.Stat(dst); err == nil {
 		klog.Infof("found %s, removing ...", dst)
-		if err := os.Remove(dst); err != nil {
-			return errors.Wrapf(err, "error removing file %s", dst)
+		switch runtime.GOOS {
+		case "linux":
```
tstromberg, line 95: AFAIK, execRunner should only run on Linux. What would you feel about escalating privilege only in cases of permission errors? For instance:

```
if err := os.Remove(dst); err != nil {
  if os.IsPermission(err) { 
    _, err = e.RunCmd(exec.Command("sudo", "rm", "-f", dst))
  }
  return errors.Wrapf(err, "error removing file %s", dst)
}

Comment (tstromberg) [reply pkg/minikube/command/exec_runner.go:95]: Could this just call `.Remove()` to keep the logic in one place?

File: pkg/minikube/driver/driver.go [hunk 5]
```diff
@@ -143,7 +143,12 @@ func BareMetal(name string) bool {
 
 // NeedsRoot returns true if driver needs to run with root privileges
 func NeedsRoot(name string) bool {
-	return name == None
```
tstromberg, line 146: I think this can just be simplified to `return false` now.

none is only on Linux.

tstromberg (changes requested):

cmd/minikube/cmd/start.go
Git-Jiro: ok

pkg/minikube/command/exec_runner.go
Git-Jiro: Ok, i added the "isPermission" escalation and moved the code into the Remove() function

pkg/minikube/driver/driver.go
Git-Jiro: done

Git-Jiro:
Once everything is fine I will squash the commits.

pkg/minikube/registry/drvs/none/none.go
```diff
@@ -61,14 +61,10 @@ func status() registry.State {
 		return registry.State{Running: true, Error: err, Installed: false, Fix: "Install docker", Doc: "https://minikube.sigs.k8s.io/docs/reference/drivers/none/"}
 	}
 
-	u, err := user.Current()
+	_, err = user.Current()
 	if err != nil {
 		return registry.State{Running: true, Error: err, Healthy: false, Doc: "https://minikube.sigs.k8s.io/docs/reference/drivers/none/"}
 	}
```
tstromberg, line 68: This needs to check that either the user is uid==0 or has sudo access. 

Otherwise, the `none` driver may be auto-selected if the user doesn't meet the requirements. It also prevents minikube from breaking if the user doesn't have sudo access. For example, right now this happens:

```
 ./out/minikube start --driver=none          253ms  Wed 28 Oct 2020 11:38:48 AM PDT
😄  minikube v1.14.0 on Debian rodete
🤹  Running on localhost (CPUs=6, Memory=63978MB, Disk=411674MB) ...
🎉  minikube 1.14.2 is available! Download it: https://github.com/kubernetes/minikube/releases/tag/v1.14.2
💡  To disable this notice, run: 'minikube config set WantUpdateNotification false'

🔥  Deleting "minikube" in none ...
E1028 11:39:05.027805  957984 none.go:188] cleanup incomplete: sudo rm -rf /var/tmp/minikube /etc/kubernetes/manifests /var/lib/minikube: exit status 1
stdout:
stderr:
Sorry, user tstromberg is not allowed to execute '/bin/rm -rf /var/tmp/minikube /etc/kubernetes/manifests /var/lib/minikube' as root on beelzebufo1.sfo.corp.google.com.
🤦  StartHost failed, but will try again: post-start: sudo mkdir (none): sudo mkdir -p /etc/kubernetes/addons /etc/kubernetes/manifests /var/tmp/minikube /var/lib/minikube /var/lib/minikube/certs /var/lib/minikube/images /var/lib/minikube/binaries /tmp/gvisor /usr/share/ca-certificates /etc/ssl/certs: exit status 1
stdout:
stderr:
Sorry, user tstromberg is not allowed to execute '/bin/mkdir -p /etc/kubernetes/addons /etc/kubernetes/manifests /var/tmp/minikube /var/lib/minikube /var/lib/minikube/certs /var/lib/minikube/images /var/lib/minikube/binaries /tmp/gvisor /usr/share/ca-certificates /etc/ssl/certs' as root on beelzebufo1.sfo.corp.google.com.
🤹  Running on localhost (CPUs=6, Memory=63978MB, Disk=411674MB) ...
😿  Failed to start none bare metal machine. Running "minikube delete" may fix it: post-start: sudo mkdir (none): sudo mkdir -p /etc/kubernetes/addons /etc/kubernetes/manifests /var/tmp/minikube /var/lib/minikube /var/lib/minikube/certs /var/lib/minikube/images /var/lib/minikube/binaries /tmp/gvisor /usr/share/ca-certificates /etc/ssl/certs: exit status 1
stdout:
stderr:
Sorry, user tstromberg is not allowed to execute '/bin/mkdir -p /etc/kubernetes/addons /etc/kubernetes/manifests /var/tmp/minikube /var/lib/minikube /var/lib/minikube/certs /var/lib/minikube/images /var/lib/minikube/binaries /tmp/gvisor /usr/share/ca-certificates /etc/ssl/certs' as root on beelzebufo1.sfo.corp.google.com.
❌  Exiting due to GUEST_PROVISION: Failed to start host: post-start: sudo mkdir (none): sudo mkdir -p /etc/kubernetes/addons /etc/kubernetes/manifests /var/tmp/minikube /var/lib/minikube /var/lib/minikube/certs /var/lib/minikube/images /var/lib/minikube/binaries /tmp/gvisor /usr/share/ca-certificates
```

Git-Jiro: Okay, sounds good.

pkg/minikube/driver/driver.go
```diff
@@ -143,7 +143,7 @@ func BareMetal(name string) bool {
 
 // NeedsRoot returns true if driver needs to run with root privileges
 func NeedsRoot(name string) bool {
```
priyawadhwa, line 145: I think we can get rid of this function and all calls to it entirely -- pretty sure only the none driver needs roots so this isn't useful anymore with this PR :)

Git-Jiro: I removed that function with a new commit.
Once everything is fine I will squash everything before final merge.

tstromberg:
Thank you!

afbjorklund:
Seems to break normal usage, by requiring "sudo" for simple copy commands...

Issue #9352, opened 2020-09-29
Minikube status shows paused cluster as running


j-windsor:
**Steps to reproduce the issue:** 

1. Pause the cluster.
2. run the following commands in the following versions.
```
➜  Downloads ./minikube12-2 status --layout cluster --output json 
{"Name":"minikube","StatusCode":418,"StatusName":"Paused","BinaryVersion":"v1.12.2","Components":{"kubeconfig":{"Name":"kubeconfig","StatusCode":200,"StatusName":""}},"Nodes":[{"Name":"minikube","StatusCode":200,"StatusName":"OK","Components":{"apiserver":{"Name":"apiserver","StatusCode":418,"StatusName":"Paused"},"kubelet":{"Name":"kubelet","StatusCode":405,"StatusName":"Stopped"}}}]}     
➜  Downloads ./minikube13-1 status --layout cluster --output json 
{"Name":"minikube","StatusCode":200,"StatusName":"OK","BinaryVersion":"v1.13.1","Components":{"kubeconfig":{"Name":"kubeconfig","StatusCode":200,"StatusName":""}},"Nodes":[{"Name":"minikube","StatusCode":200,"StatusName":"OK","Components":{"apiserver":{"Name":"apiserver","StatusCode":418,"StatusName":"Paused"},"kubelet":{"Name":"kubelet","StatusCode":405,"StatusName":"Stopped"}}}]}
```

As you can see, the most recent version does not correctly output the status as paused.

j-windsor:
Might be related to this commit: https://github.com/kubernetes/minikube/commit/1783cc2939ec058c2742293693a39c45d4f626a4

@priyawadhwa any ideas?

tstromberg:
Still repeatable at HEAD.

PR #9383, opened 2020-10-02
Fix minikube status bug when cluster is paused

priyawadhwa:
fixes https://github.com/kubernetes/minikube/issues/9352 and adds an integration test to make sure the regression doesn't happen again

### DIFF
--- a/cmd/minikube/cmd/status.go
+++ b/cmd/minikube/cmd/status.go
@@ -66,6 +66,8 @@ const (
 	// Irrelevant is used for statuses that aren't meaningful for worker nodes
 	Irrelevant = "Irrelevant"
 
+	InsufficientStorage = reason.ExInsufficientStorage
+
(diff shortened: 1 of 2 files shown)


cmd/minikube/cmd/status.go
```diff
@@ -84,18 +86,18 @@ const (
 	Paused   = 418 // I'm a teapot!
 
 	// 5xx signifies a server-side error (that may be retryable)
-	Error               = 500
-	InsufficientStorage = 507
```
tstromberg, line 88: InsufficientStorage should stay 507 -- these are HTTP style error codes, not UNIX error codes.

tstromberg (changes requested):

cmd/minikube/cmd/status.go
priyawadhwa: Thanks for catching that, fixed

tstromberg (approved):

PR #9389, opened 2020-10-04
Do some simple cleanup of the golint complaints

afbjorklund:
The "magic strings" and enums still needs another approach

Simply echoing everything twice doesn't seem very useful ?


priyawadhwa (approved):
Nice!

PR #9329, opened 2020-09-27
Buildroot 2020.02.6

afbjorklund:
Closes #7904 

Still has the **old** LTS kernel, which should hopefully avoid the regressions from last time (PR #8187)

PR #9394, opened 2020-10-05
re-synchronize default linux kernel config with buildroot

dalehamel:
The current linux configuration appears to have been manually edited, and includes configuration options that do not exist on the current Linux kernel version used by minikube, or else configs were added that only applied to 5.4, which were never reverted when the 5.4 kernel itself was reverted (i think because https://github.com/kubernetes/minikube/pull/8187 was reverted?)

In particular, I think the PRs that caused this to diverge are:

https://github.com/kubernetes/minikube/pull/8582, which adds IKHEADERS, which only applies to newer (5.4) kernel version (cc @alban

https://github.com/kubernetes/minikube/pull/8187 which seems to be unecessary, since the config for CONFIG_PCI and CONFIG_TMPFS seem to be added by default (cc @afbjorklund), and it looks like CONFIG_VIRTIO_FS is only available in 5.4 or later

It looks like we tried upgrading to 5.4 and reverted it ,but never reverted the linux config?

Without a consistent, machine-generated kernel config, done via `make linux-menuconfig` and `make linux-update-defconfig`, it will be pretty hairy to make any kernel changes.

EDIT: I just realized, maybe this should be run as a CI check? Verify that "make linux-update-defconfig" results in 0 changes? Would prevent manual edits of the file, and enforce that it is machine generated as buildroot expects.

I also just noticed that #9329 keeps the old LTS kernel, so i think this patch is still relevant, and shouldn't conflict with the kernel config in that PR.


dalehamel:
Note that i haven't actually changed *anything* here, i've just run the scripts you are supposed to run with buildroot, to do a no-op change. The changes here are the result of buildroot's automation.

dalehamel:
This should wait for #9329 and then be regenerated, as the diff will make more sense.

afbjorklund:
> It looks like we tried upgrading to 5.4 and reverted it ,but never reverted the linux config?

That is pretty much the case. And we haven't made a final decision, if we should stick with 4.19 or not.
Mostly because we never got to the _bottom_ of what the problem with the new buildroot and kernel was ?

If time is running out, I'm all for cleaning up the previous kernel config and redoing it properly later on.
The minikube kernel config has grown quite organically, and doesn't always describe the changes made.

I made some efforts to compare with a more vanilla kernel config, in another project.

The boot2docker project had some nice division of kernel config into separate files.

dalehamel:
I don't have a problem with the current, minimal config, it's just missing a few options, and I think that it would be good to keep it consistent with what buildroot expects.

Since buildroot offers this feature to trim down the kernel config to "what you need", and that is what we're storing in here, it makes sense to assert this I think (eg, a CI check that the configured kernel has 0 changes as I suggested above).

We can re-sync it with 5.4, or try a separate 5.4 config (and build both 4.19 and 5.4 for a while, maybe?)

afbjorklund:
I think this needs to be rebased/regenerated, before we can merge it ?

PR #9401, opened 2020-10-06
Fixes existing Polish translations

programistka:


medyagh (approved):
Dizekojya @programistka

PR #9412, opened 2020-10-07
fix panic when updating mount-string with KiC driver after cluster creation without mount configuration

Asarew:
Hopefully the title says it all. https://github.com/kubernetes/minikube/pull/8159#issuecomment-704662275


medyagh (approved):
thank you for fixing that

medyagh (commented):
btw @Asarew it would be great if we add something to our documentation about mount feature in docker driver, would you be interested to add this doc https://minikube.sigs.k8s.io/docs/drivers/docker/ ?

Asarew:
I'll take a crack at it, going to open up a PR in a couple of days

Issue #9410, opened 2020-10-07
FR/crazy idea: delete context when minikube is stopped


tstromberg:
The context isn't useful if minikube is stopped, and can mislead folks into talking to a stopped cluster. What if, instead of unselecting it as a default, we just deleted it.

prezha:
not so crazy ;) `.kube/config` gets filled back in with the cluster data on subsequent minikube start - when it can actually be used again

tstromberg:
Thank you @prezha !

medyagh:
@tstromberg this could break the IDE users.  per this request, https://github.com/kubernetes/minikube/issues/9001

@prezha please make sure, this does NOT delete the context if the user has specified 
"--keep-context-active=false"

prezha:
@medyagh yes, the same logic still applies: if `keep-context-active` is set to true, it will not touch the context

jorgemoralespou:
This is not nice. Now, as the context is deleted and recreated, a user needs to select a specific namespace he's working on every time he starts/stops the cluster.

prezha:
@jorgemoralespou you can use `--keep-context-active=true` option to keep the context on `minikube stop`?

jorgemoralespou:
I was expecting that answer. Why would I want to keep the context active if the instance is stopped. When stopped what does make sense is unset, and only delete the context when minikube instance is deleted, hence, previous behavior.

prezha:
i don't think the issue is with deleting/recreating the context when minikube stops/starts, what we potentially could do is additionally saving/restoring also the default namespace

jorgemoralespou:
Yes, you could save/restore the namespace. But IMHO the fact that a namespace is deleted is more confusing than the fact that minikube is stopped and you have a context you can't connect to until you start minikube.

PR #9414, opened 2020-10-07
Delete context when stopped

prezha:
fixes: #9410 

as stated in the original issue: if `keep-context-active` is not set to `true`, the context for the stopped cluster should be deleted (instead of just unset as the current context) to avoid confusion

**example**:

```
❯ kubectl config get-contexts
CURRENT   NAME                          CLUSTER                    AUTHINFO                                                       NAMESPACE
          minikube                      minikube                   minikube
*         p2                            p2                         p2
          p3                            p3                         p3
```
```
❯ minikube stop -p p2
✋  Stopping node "p2"  ...
🛑  Powering off "p2" via SSH ...
🛑  1 nodes stopped.
```
**current behaviour**:
```
❯ kubectl config get-contexts
CURRENT   NAME                          CLUSTER                    AUTHINFO                                                       NAMESPACE
          minikube                      minikube                   minikube
          p2                            p2                         p2
          p3                            p3                         p3
```
**new behaviour**:
```
❯ kubectl config get-contexts
CURRENT   NAME                          CLUSTER                    AUTHINFO                                                       NAMESPACE
          minikube                      minikube                   minikube
          p3                            p3                         p3
```
```
❯ minikube start -p p2
😄  [p2] minikube v1.13.1 on "Opensuse-Tumbleweed"
❗  docker is currently using the btrfs storage driver, consider switching to overlay2 for better performance
👍  Starting control plane node p2 in cluster p2
🔄  Restarting existing docker container for "p2" ...
🐳  Preparing Kubernetes v1.19.2 on Docker 19.03.8 ...
🏄  Done! kubectl is now configured to use "p2" by default
```
```
❯ kubectl config get-contexts
CURRENT   NAME                          CLUSTER                    AUTHINFO                                                       NAMESPACE
          minikube                      minikube                   minikube
*         p2                            p2                         p2
          p3                            p3                         p3
```


cmd/minikube/cmd/stop.go
```diff
@@ -115,8 +115,8 @@ func stopProfile(profile string) int {
 	}
 
 	if !keepActive {
-		if err := kubeconfig.UnsetCurrentContext(profile, kubeconfig.PathFromEnv()); err != nil {
-			exit.Error(reason.HostKubeconfigUnset, "update config", err)
+		if err := kubeconfig.DeleteContext(profile, kubeconfig.PathFromEnv()); err != nil {
+			exit.Error(reason.HostKubeconfigUpdate, "update config", err)
```
medyagh, line 119: lets create a new Reason Exit here 
https://github.com/medyagh/minikube/blob/82a5189b934ea02d127be85dbdba50dfa4282984/pkg/minikube/reason/reason.go#L207

```
	HostKubeconfigDelete = Kind{ID: "HOST_KUBECONFIG_DELETE", ExitCode: ExHostConfig}
```

medyagh (changes requested):

cmd/minikube/cmd/stop.go
prezha: @medyagh you are most welcome!
that did cross my mind, but `HostKubeconfigDelete` sounded a bit wrong (we are not going to delete kubeconfig), while `HostKubeconfigUpdate` not so bad as we are effectively just updating the kubeconfig
on the other hand, maybe:
```
	HostKubeconfigContextDelete = Kind{ID: "HOST_KUBECONFIG_CONTEXT_DELETE", ExitCode: ExHostConfig}
```
would be more appropriate?
what do you think?

medyagh: that is actually good thinking ! I like that

how about something shorter but unique

"HostKubeconfigDeleteCtx"

the goal is if users collect these exit codes as metrics, they should find out what was the cause of the problem in a unique word

prezha: yes, that sounds just about right :)
the update is on its way...
thanks @medyagh !

medyagh (approved):
thanks

Issue #8359, opened 2020-06-02
minikube ssh: Unexpected handling of stdin input


joshmue:
**Steps to reproduce the issue:** 

1. Download Minikube v1.11.0 as `minikube_v1.11.0` (and v1.10.1 as `minikube_v1.10.1` for reference).
2. Do `./minikube_v1.11.0 start` (VirtualBox installed).
3. Then do:
```
prompt$ echo 'some test data' | ./minikube_v1.11.0 ssh -- cat -
some test data
some test data

```

The data is handled (twice?), but the process does not terminate.

For reference, the same command using `minikube_v1.10.1` is working as expected:

```
prompt$ echo 'some test data' | ./minikube_v1.10.1 ssh -- cat -
some test data
prompt$
```

I assume this is a regression? Please ask if you need any more information.

afbjorklund:
Seems to be related to 7224064b235a84220a0705cba4327a2630ced223, @sharifelgamal please take a look

```
 echo "hello world" | minikube ssh -- cat -
hello world
hello world
^C
```

When you revert that commit, it goes back to normal:

```
 echo "hello world" | ./out/minikube ssh -- cat -
hello world
```

However, the value of the `--native-ssh` doesn't seem to have any effect.

afbjorklund:
The difference that was made, was the order between the calls:

ssh.SetDefaultClient

host.CreateSSHClient

That would be from github.com/docker/machine/libmachine/ssh

afbjorklund:
> However, the value of the --native-ssh doesn't seem to have any effect.

Basically the old code _always_ used external, the new code always native.

--native-ssh=false
libmachine: Using SSH client type: external
--native-ssh=true
libmachine: Using SSH client type: external

--native-ssh=true
libmachine: Using SSH client type: native
--native-ssh=false
libmachine: Using SSH client type: native

And the native implementation seems to be broken (doesn't work).

afbjorklund:
Possibly related to this bug: https://github.com/docker/machine/issues/4806

```
hello world^M
hello world^M
```

sharifelgamal:
So you're exactly right, there are 2 separate issues here:
- We're not honoring the `native-ssh` parameter so it's native no matter what
- The native ssh implementation end lines with windows style carriage returns for reasons I don't quite understand.

The first one is easy enough to fix, it was just a regression I introduced. The second one should be able to be fixed, at least on our fork of libmachine.

afbjorklund:
[...]

PR #9417, opened 2020-10-08
Fix "--native-ssh" flag of "minikube ssh"

joshmue:
The "--native-ssh" flag was completely discarded; This PR fixes that and also makes sure that the SSH client is created after it is configured accordingly.

Refs #8359

Signing of CLA ~~will follow~~ is done.

### DIFF
--- a/cmd/minikube/cmd/ssh.go
+++ b/cmd/minikube/cmd/ssh.go
@@ -67,6 +67,6 @@ var sshCmd = &cobra.Command{
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


sharifelgamal:
This looks like it's doing the right thing. I'll mark it for testing once the CLA is signed.

joshmue:
> This looks like it's doing the right thing. I'll mark it for testing once the CLA is signed.

@sharifelgamal (Just in case you did not receive any notification): The CLA is signed now.

medyagh (approved):
thank you

Issue #9409, opened 2020-10-06
docker: Exiting due to DRV_CP_ENDPOINT: failed to parse ip for ""


medyagh:
in this embeded use case, minikube detected that DiskSpace is full but when it exited, it gave the exit code RV_CP_ENDPOINT which is not correct, we should improve the exit code to reflect the real reason

```
Failed to start minikube: Error while starting Minikube. Error: ! minikube was unable to download gcr.io/k8s-minikube/kicbase:v0.0.12-snapshot3, but successfully downloaded gcr.io/k8s-minikube/kicbase:v0.0.12-snapshot3 as a fallback image
,X Docker is nearly out of disk space, which may cause deployments to fail! (90% of capacity)
,* Suggestion: 

    Try at least one of the following to free up space on the device:
    
    1. Run "docker system prune" to remove unused docker data
    2. Increase the amount of memory allocated to Docker for Desktop via
    Docker icon > Preferences > Resources > Disk Image Size
    3. Run "minikube ssh -- docker system prune" if using the docker container runtime
* Related issue: https://github.com/kubernetes/minikube/issues/9024
,X Exiting due to DRV_CP_ENDPOINT: failed to parse ip for ""
,* 
,* If the above advice does not help, please let us know: 
,  - https://github.com/kubernetes/minikube/issues/new/choose
```

tstromberg:
I don't think this is correct. 90% of capacity should not cause a failure, or exit prematurely. Could you help me to see  connection between the warning and the inability to determine what IP to use?

That said, there is room for an additional check if the total available disk space is under say, 500MB, but I don't think that is likely what happened here.

priyawadhwa:
Yah @medyagh I think the disk space is just a warning at 90%. I think we can close this issue?

tstromberg:
As seen on Cloud Run:

```
I1006 22:33:51.377450   48191 translate.go:64] Getting system locale failed: Could not detect Language
I1006 22:33:51.378058   48191 translate.go:106] Setting Language to en-US ...
I1006 22:33:51.378421   48191 out.go:184] Setting OutFile to fd 1 ...
I1006 22:33:51.378433   48191 out.go:231] TERM=,COLORTERM=, which probably does not support color
I1006 22:33:51.378437   48191 out.go:197] Setting ErrFile to fd 2...
I1006 22:33:51.378441   48191 out.go:231] TERM=,COLORTERM=, which probably does not support color
[...]

PR #9426, opened 2020-10-08
Add further debugging for DRV_CP_ENDPOINT failure

priyawadhwa:
To make #9409 easier to debug.

### DIFF
--- a/pkg/drivers/kic/oci/network_create.go
+++ b/pkg/drivers/kic/oci/network_create.go
@@ -100,8 +100,10 @@ func tryCreateDockerNetwork(subnetAddr string, subnetMask int, name string) (net
 
 // returns subnet and gate if exists
 func dockerNetworkInspect(name string) (*net.IPNet, net.IP, error) {
-	rr, err := runCmd(exec.Command(Docker, "network", "inspect", name, "--format", "{{(index .IPAM.Config 0).Subnet}},{{(index .IPAM.Config 0).Gateway}}"))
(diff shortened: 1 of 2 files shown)


pkg/drivers/kic/oci/network_create.go
```diff
@@ -124,6 +126,15 @@ func dockerNetworkInspect(name string) (*net.IPNet, net.IP, error) {
 	return subnet, gateway, nil
 }
 
+func logDockerNetworkInspect(name string) {
+	cmd := exec.Command(Docker, "network", "inspect", name)
```
medyagh, line 130: to avoid flooding the logs and confusion, lets add a line that says, "running dubugging command to gather additional logs "

pkg/drivers/kic/oci/network_create.go
```diff
@@ -124,6 +126,15 @@ func dockerNetworkInspect(name string) (*net.IPNet, net.IP, error) {
...
+	rr, err := runCmd(cmd)
```
medyagh, line 131: consider also running this debug func LogContainerDebug
https://github.com/medyagh/minikube/blob/e1615c81545be00876e76504ca1075ac9bfaba48/pkg/drivers/kic/oci/errors.go#L67

medyagh (changes requested):

pkg/drivers/kic/oci/network_create.go
priyawadhwa: done

pkg/drivers/kic/oci/network_create.go
priyawadhwa: sounds good

medyagh (approved):
thank you

Issue #9399, opened 2020-10-05
--delete-on-failure does not delete on GUEST_DRIVER_MISMATCH


j-windsor:
**Steps to reproduce the issue:** 

1. `minikube start --driver=docker`
2. `minikube start --driver=hyperkit --delete-on-failure`
3. See message on GUEST_DRIVER_MISMATCH
4. `minikube profile list` to see that the profile is not deleted.

![Screen Shot 2020-10-05 at 3 48 09 PM](https://user-images.githubusercontent.com/554839/95140509-f0322500-0723-11eb-9999-683e8da9a835.png)

This is something that can be fixed by deleting, so I would expect deletion to occur.

**Related:**
- https://github.com/kubernetes/minikube/issues/9372
- https://github.com/kubernetes/minikube/issues/9397

vinu2003:
Hey @j-windsor can i work on this issue if you aren't keen to do this ? then i can fix #9397

tstromberg:
@vinu2003 - go for it!

vinu2003:
My understanding here: the --delete-on-failure when set, should delete the default minikube profile(in this case docker) if it doesn't match with the one provided in start command (in this case"hyperkit") - then it should delete the existing docker cluster and create default one with hyper kit as driver.

medyagh:
@vinu2003 I haven't heard from you, please let me know if you are still working on this

azhao155:
I think the issue got fixed. 

I repro the issue with same cmd and see following:

```
zyanshu@zyanshukvm:~/minikube$ minikube start --driver=docker
-bash: minikube: command not found
zyanshu@zyanshukvm:~/minikube$ ./out/minikube start --driver=docker
😄  minikube v1.16.0 on Debian rodete
    ▪ MINIKUBE_ACTIVE_DOCKERD=p1
✨  Using the docker driver based on user configuration
👍  Starting control plane node minikube in cluster minikube
🔥  Creating docker container (CPUs=2, Memory=26100MB) ...
🐳  Preparing Kubernetes v1.20.0 on Docker 20.10.0 ...
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔎  Verifying Kubernetes components...
🌟  Enabled addons: storage-provisioner, default-storageclass
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default
zyanshu@zyanshukvm:~/minikube$
[...]

PR #9433, opened 2020-10-10
WIP: start: fix to delete existing cluster when 'delete-on-failure' is set

vinu2003:
**What**
The PR address changes required to forcefully delete cluster when there exists another cluster with same profile making decision based on the user specified driver validations.

**Why**
fixes: #9399

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -19,6 +19,15 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
+	"math"
+	"net"
+	"net/url"
+	"os"
+	"os/exec"
+	"os/user"
+	"runtime"
+	"strings"
+
 	"github.com/blang/semver"
 	"github.com/docker/machine/libmachine/ssh"
 	"github.com/google/go-containerregistry/pkg/authn"
@@ -29,14 +38,6 @@ import (
 	gopshost "github.com/shirou/gopsutil/host"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"math"
-	"net"
-	"net/url"
-	"os"
-	"os/exec"
-	"os/user"
-	"runtime"
-	"strings"
 
 	"k8s.io/klog/v2"
 	cmdcfg "k8s.io/minikube/cmd/minikube/cmd/config"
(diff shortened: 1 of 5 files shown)


vinu2003:
Hey @tstromberg and @josedonizetti - the changes require some code refactoring as i would look more easier to understand. Also, runStart started having cyclomatic complexity issues > 30.
I might have to add code for redeploying deleted cluster if the provisioning fails which could be > 30 and so I might further break down when i am addressing code coverage issues.

cmd/minikube/cmd/start.go
```diff
@@ -191,11 +190,12 @@ func runStart(cmd *cobra.Command, args []string) {
 		}
 	}
 
-	starter, err := provisionWithDriver(cmd, ds, existing)
+	starter, clusterDeleted, err := provisionWithDriver(cmd, ds, existing, deleteExistingCluster)
```
priyawadhwa, line 193: We currently have something called a [FailFastError](https://github.com/kubernetes/minikube/blob/231bf12cc40c19d815e7b61f066cb259cfd7f6a0/pkg/minikube/bootstrapper/kubeadm/errors.go#L32). When this error is returned, we know that we should just exit out of minikube instead of trying to restart the cluster.

Instead of passing in `deleteExistingCluster` as a new field to the function, could we follow the above pattern and create a `DeleteClusterError`? We then check in the appropriate spot if the error we have is of the type, and if it is, then we delete the cluster.

example of fail fast error for reference:

https://github.com/kubernetes/minikube/blob/d7c517471dcfe47d4f481f8cbc3c3f0abbbe3116/pkg/minikube/node/start.go#L386

priyawadhwa (commented):
Thanks for taking this on! Left a suggestion.

cmd/minikube/cmd/start.go
vinu2003: Hey @priyawadhwa thanks for pitching in to help.

Agree. we must use the existing mechanism. 
Also, all I tried achieve here is user should have valid running cluster and would be happy if that existing cluster gets deleted and the creation of new one also fails.

I am deleting it in wrong spot and that's why I have added TODO section to recreate back the old cluster with old driver if the new cluster creation with user specified driver fails - this is not correct approach recreating old cluster back again.

I am reworking on the logic around that area.

Thank you.

medyagh: I agree with @priyawadhwa I think that there is no reason for select driver to return something which is already available globally viper.Get...

vinu2003: Hey @medyagh The select driver evaluates the combination of all cases of already existing profile with same driver or different driver and the user provided --driver option and/or --delete-on-failure options and sets this flag to indicate further that the --delete-on-failure has_to_be_considered or not. Thats why i refactored the code a bit more to consolidate the logic in one place.

What @priyawadhwa mentioned above is, to make use of FailFast error type which can be used to address many cases with a single action and delete the cluster appropriately.

I believe we are all syncing up now. :)

I have reworked around this logic as mentioned earlier and will send update.

Thank you.

vinu2003: however, @medyagh if you have any thought on the code refactor and select driver function kindly point out the same.

sharifelgamal:
@vinu2003 Could you add the output of minikube start before and after this change to show the difference of this PR working?

priyawadhwa:
Hey @vinu2003 -- this PR has been quiet for a few weeks now. I'm going to go ahead and close it for now, but please reopen whenever you have a chance to take another look at it!

I think the next couple things it needs are:

1. A rebase
1. To use the `FailFastError` concept I described above so that we aren't passing a bool between functions

Thank you!

PR #9459, opened 2020-10-13
new --wait component "kubelet"

medyagh:
while this does NOT fix the issue https://github.com/kubernetes/minikube/issues/9456 it might help to get to the bottom of it.

this adds a new wait component which is not enabled by default. 
if user passes --wait=all it will wait for everything including kubelet


pkg/minikube/bootstrapper/bsutil/kverify/kubelet.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/docker/machine/libmachine/state"
+	"github.com/golang/glog"
+	"k8s.io/minikube/pkg/minikube/command"
+	"k8s.io/minikube/pkg/minikube/sysinit"
+	"k8s.io/minikube/pkg/util/retry"
+)
+
+// KubeletStatus checks the kubelet status
+func KubeletStatus(cr command.Runner) state.State {
```
priyawadhwa, line 32: If we added `Status` as a function to systemd then it could be used for all services instead of special casing kubelet

Then, we should be able to use something like this instead

https://github.com/kubernetes/minikube/blob/5ce9d99d68a57f833b7af035c384479122479968/pkg/minikube/sysinit/systemd.go#L44

pkg/minikube/bootstrapper/bsutil/kverify/kubelet.go
```diff
@@ -0,0 +1,60 @@
...
+	glog.Infof("Checking kubelet status ...")
+	active := sysinit.New(cr).Active("kubelet")
+	if active {
+		return state.Running
+	}
+	return state.Stopped
+}
+
+// WaitForKubelet will wait for Kubelet service to be running ...
+func WaitForKubelet(cr command.Runner, timeout time.Duration) error {
```
priyawadhwa, line 42: Similarly, this could be generalized to wait for any service

pkg/minikube/bootstrapper/bsutil/kverify/kubelet.go
```diff
@@ -0,0 +1,60 @@
+/*
```
priyawadhwa, line 1: Since this file is about services, we should probably rename it from `kubelet.go`. I also think these functions would make more sense in the systemd package for that reason

pkg/minikube/node/start.go
```diff
@@ -196,7 +196,7 @@ func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
 		}
 	}
 
-	klog.Infof("Will wait %s for node ...", waitTimeout)
+	klog.Infof("Will wait %s for node up to ", viper.GetDuration(waitTimeout))
```
priyawadhwa, line 199: Not sure what "up to" is trying to convey here

medyagh: before this it was getting the "flag-name"
this means will wait up to this amount of time for the cluster to be up

pkg/minikube/bootstrapper/bsutil/kverify/kubelet.go
medyagh: thanks will move to a different file

medyagh: make sense, changed file name but moving that functionality to systemd package, it will be a follow up refactor PR. since it does not concern adding new component--wait flag in this PR

Issue #9427, opened 2020-10-08
combine TestPause into one to avoid timeout on Gh Action


medyagh:
on github actions "pause_preload_dockerflags_docker_ubuntu"
I see only 2 passes and the rest times out as seen here https://github.com/kubernetes/minikube/pull/9425/checks?check_run_id=1228465306

```
*******************2 Passes :) *******************
[
  "TestPreload",
  "TestPauseStatus"
]
*************************
```

we added a new test TestPauseStatus that spins up a new cluster, and that times out the github action test more than 15 minutes.

https://github.com/medyagh/minikube/blob/4216496bec8ab890577ba3ffa434fbc1c038feb1/test/integration/status_test.go#L102

if there is not a good reason, we should combine TestPauseStatus with the existing  TestPause that has the following subtests

```
			{"Start", validateFreshStart},
			{"SecondStartNoReconfiguration", validateStartNoReconfigure},
			{"Pause", validatePause},
			{"Unpause", validateUnpause},
			{"PauseAgain", validatePause},
			{"DeletePaused", validateDelete},
			{"VerifyDeletedResources", validateVerifyDeleted},

```

PR #9466, opened 2020-10-14
Combine pause status test into existing pause test

priyawadhwa:
fixes https://github.com/kubernetes/minikube/issues/9427#event-3868698847


medyagh (changes requested):
thanks for doing this, the PR / pause_preload_dockerflags_docker_ubuntu  test is timing out still, maybe we we need to increase the time out

or if increasing timeout doesnt help we should skip this test for Docker on Gh Action on Linux simmilar to this
```
	if GithubActionRunner() && runtime.GOOS == "darwin" {
		t.Skip("The test WaitService is broken on github actions in macos https://github.com/kubernetes/minikube/issues/8434")
	}
```

medyagh (approved):

medyagh:
even though this does not fix gh action, will still merge to have one clustter stareting for this  instead of multiple clusters

PR #9477, opened 2020-10-16
Upgrade Docker, from 19.03.12 to 19.03.13

afbjorklund:


tstromberg (approved):

tstromberg:
Integration test failures seem unrelated -- and in some cases, setup incorrectly for ISO tests (minikube v1.0.0 is not expected to work for this ISO)

PR #9478, opened 2020-10-16
Upgrade containerd to the same as docker-ce uses

afbjorklund:
Goes with PR #9477


afbjorklund:
Copy/paste error in first attempt.

tstromberg:
Thank you!

PR #9480, opened 2020-10-16
get greatest instead of just latest Kubernetes version

### LINKED ISSUE #4392: Write a script to automatically propose PR's due to a new Kubernetes release

prezha:
fixes #4392 (iteration 1 & 2)

fixing two mistakes i've previously made:
1. returning latest, but not necessarily also the greatest version (eg, Kubernetes released v1.18.10 after v1.19.3 but we need the latter)
2. comparing versions as strings, now using semver to check version validity, check if it's prerelease, and compare versions

and also fixed klog `no such flag -logtostderr` error

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -73,6 +73,7 @@ require (
 	golang.org/x/build v0.0.0-20190927031335-2835ba2e683f
 	golang.org/x/crypto v0.0.0-20200510223506-06a226fb4e37
 	golang.org/x/exp v0.0.0-20200224162631-6cc2880d07d6
+	golang.org/x/mod v0.3.0
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
 	golang.org/x/sync v0.0.0-20200317015054-43a5402ce75a
 	golang.org/x/sys v0.0.0-20200523222454-059865788121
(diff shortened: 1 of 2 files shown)


sharifelgamal:
Thanks for fixing this!

/ok-to-test

medyagh (approved):

PR #9487, opened 2020-10-18
add --watch flag for minikube status with optional interval duration

prezha:
fixes: #9468 

add `minikube status --watch[=duration]` (`-w`) flag that will produce continuous output waiting for a specified duration (defaulting to 1 sec) between each poll

### example output
```
❯ minikube status --output json -l=cluster --watch 2>/dev/null | jq .StatusName 
"Stopped"
"Stopped"
"Stopped"
...
"Starting"
"Starting"
"Starting"
...
"OK"
"OK"
"OK"
...
"Stopping"
"Stopping"
"Unknown"
"Stopped"
"Stopped"
"Stopped"
```

### DIFF
--- a/cmd/minikube/cmd/status.go
+++ b/cmd/minikube/cmd/status.go
@@ -217,7 +217,11 @@ var statusCmd = &cobra.Command{
 }
 
 // writeStatusesAtInterval writes statuses in a given output format - at intervals defined by duration
+<<<<<<< HEAD
 func writeStatusesAtInterval(duration time.Duration, api libmachine.API, cc *config.ClusterConfig) {
+=======
+func writeStatusesAtInterval(duration time.Duration, api libmachine.API, cc *config.ClusterConfig) int {
+>>>>>>> ab981fe40 (move write statuses logic to a separate func)
 	for {
 		var statuses []*Status
 
--- a/cmd/minikube/cmd/status.go
+++ b/cmd/minikube/cmd/status.go
@@ -54,6 +54,7 @@ var (
 	statusFormat string
 	output       string
 	layout       string
+	watch        time.Duration
 )
 
 const (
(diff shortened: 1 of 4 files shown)


cmd/minikube/cmd/status.go
```diff
@@ -203,59 +204,63 @@ var statusCmd = &cobra.Command{
...
 				}
+			default:
+				exit.Message(reason.Usage, fmt.Sprintf("invalid output format: %s. Valid values: 'text', 'json'", output))
 			}
-		default:
-			exit.Message(reason.Usage, fmt.Sprintf("invalid output format: %s. Valid values: 'text', 'json'", output))
+			if !watch {
+				os.Exit(exitCode(statuses))
+			}
+			time.Sleep(1 * time.Second)
```
medyagh, line 262: how about making this configureable ?
like --watch-interval with a default value of 1 second?

medyagh (changes requested):

cmd/minikube/cmd/status.go
prezha: like the idea!
and instead of adding a separate flag, would we want just to expand the `--watch` flag to also support an optional value (defaulting to 1 sec)? eg, `--watch 5` would sleep for 5 sec between consecutive polls

priyawadhwa (commented):
I think the functionality we want is:

1. If --watch isn't set, then only print status once
1. If --watch is set, then print status at the given duration

Right now, it looks to me like we default to printing every second.

cmd/minikube/cmd/status.go
```diff
@@ -203,59 +204,63 @@ var statusCmd = &cobra.Command{
 		cname := ClusterFlagValue()
 		api, cc := mustload.Partial(cname)
 
-		var statuses []*Status
+		for {
+			var statuses []*Status
```
priyawadhwa, line 208: Could you move all of the code in the for loop into a separate function? That would make this more readable.

prezha:
> I think the functionality we want is:
> 
> 1. If --watch isn't set, then only print status once
> 2. If --watch is set, then print status at the given duration
> 
> Right now, it looks to me like we default to printing every second.

thank you for your comment - would you say that the [proposal above](https://github.com/kubernetes/minikube/pull/9487#discussion_r512923251) is in line with that or you think it should be a separate flag?

priyawadhwa:
@prezha yes, I agree with your proposal. If something like, `--watch=2m` is set, then we print at intervals.

prezha:
> @prezha yes, I agree with your proposal. If something like, `--watch=2m` is set, then we print at intervals.

yes, that is the idea, thank you for confirming

cmd/minikube/cmd/status.go
prezha: @priyawadhwa i moved the code to a separate func `writeStatusesAtInterval`, please have a look

medyagh (commented):
looks good I have a suggestion that only for the --watch flag we set the log file size maximum to be 10 MB
that will prevent the user's machine tmp dir be filled up with a huge log dir

prezha:
thanks @medyagh for your review and `--watch` flag is potentially separate from `--log_file` and its `--log_file_max_size` complement?

medyagh (approved):

Issue #9493, opened 2020-10-19
Parallels driver is getting called as a built-in library instead of RPC Server


legal90:
The issue is well-described in this thread (starting from the linked comment): https://github.com/Parallels/docker-machine-parallels/issues/93#issuecomment-711048093

We realized that the output of parallels driver does not depend on the actual binary of `docker-machine-driver-parallels` available in PATH. It appeared that `minikube` was calling driver's functions as a usual library, without launching the binary as the RPC Server.

I found the root cause of this and will submit a fixing PR very soon

afbjorklund:
Interesting, this might actually be the same issue as with all other drivers in the registry: #5549

tstromberg:
This is definitely not by design -- the docs mention that the driver needs to be downloaded first.

tstromberg:
@legal90 - if you have a PR to address this, feel free to send the review my way.

legal90:
@tstromberg I sent a PR which "fixes" this issue by making _minikube <-> parallels driver_ communication trough RPC: https://github.com/kubernetes/minikube/pull/9494

However, based on https://github.com/kubernetes/minikube/pull/9494#issuecomment-713194751, it seems that you guys prefer to have drivers built-in. 
If so - I'll send a new PR soon, which will update the docs and remove the check of driver binary.

afbjorklund:
Some drivers are built-in (like virtualbox), some have an external binary (like docker-machine-driver-kvm2).

Our driver "registry" looks to be rather confused, though. So it might very well be that it tries to do both...

I think the main reason for not building _everything_ in, would be that it would then also link to extra libraries.

For instance, the kvm2 driver requires libvirt to be installed on the host. But virtualbox just uses VBoxManage

legal90:
@afbjorklund Thank you!
Here is the alternative fix which keeps the `parallels` driver built-in (as it is since a while) and removes the redundant check of `docker-machine-driver-parallels` binary.

PR #9494, opened 2020-10-19
Remove Init function from the parallels driver definition

legal90:
Fixes https://github.com/kubernetes/minikube/issues/9493

That's needed for `minikube` to stop treating `parallels` as a built-in driver and use RPC calls for calling its functionality.

### DIFF
--- a/pkg/minikube/registry/drvs/parallels/parallels.go
+++ b/pkg/minikube/registry/drvs/parallels/parallels.go
@@ -23,7 +23,6 @@ import (
 	"os/exec"
 
 	parallels "github.com/Parallels/docker-machine-parallels"
-	"github.com/docker/machine/libmachine/drivers"
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/download"
 	"k8s.io/minikube/pkg/minikube/driver"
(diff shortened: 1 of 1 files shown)


legal90:
This patch is based on this logic:
https://github.com/kubernetes/minikube/blob/97d96bf92528d7897a18895b2d8ee60c23b7e0ad/pkg/minikube/machine/client.go#L92-L94

But I'm a bit confused by the name `legacyClient` there. So please let me know if you want to keep drivers built-in instead of actually calling them via RPC. 
In this case I will submit another PR - with the dependency version bump + the removal of checking the `docker-machine-driver-parallels` path, as it won't be used anymore.

sharifelgamal:
> This patch is based on the following logic:
> https://github.com/kubernetes/minikube/blob/97d96bf92528d7897a18895b2d8ee60c23b7e0ad/pkg/minikube/machine/client.go#L92-L94
> 
> But I'm a bit confused by the name `legacyClient` there. So please let me know if you want to keep drivers built-in instead of actually calling them via RPC.
> In this case I will submit another PR - with the dependency version bump + the removal of checking the `docker-machine-driver-parallels` path, as it won't be used anymore.

I think this latter approach is going to be the better one. Feel free to open that PR and I'll review it.

medyagh (approved):

Issue #7344, opened 2020-03-31
`minikube ip` returns 127.0.0.1


isra17:
**Steps to reproduce the issue:** 

1. `minikube start`
2. `minikibe ip`

**Full output of failed command:** 

```
I0331 16:29:52.931261  527127 mustload.go:51] Loading cluster: minikube
I0331 16:29:52.969418  527127 host.go:65] Checking if "minikube" exists ...
127.0.0.1
```

**Full output of `minikube start` command used, if not already included:**

```
😄  minikube v1.9.0 on Arch rolling
✨  Using the docker driver based on existing profile
🚜  Pulling base image ...
🔄  Retarting existing docker container for "minikube" ...
🐳  Preparing Kubernetes v1.18.0 on Docker 19.03.2 ...
    ▪ kubeadm.pod-network-cidr=10.244.0.0/16
🌟  Enabling addons: default-storageclass, storage-provisioner
🏄  Done! kubectl is now configured to use "minikube"
```

I had a minikube instance before I upgraded to 1.9.0 and the bug started to appear. Creating a new cluster from scratch didn't solve the issue. `minikube service foo --url` work as expected. Same for the internal ip from `kubectl get node -o wide`.

eox-dev:
Having the same issue in WSL2.

andrebrait:
Same here in Ubuntu 18.04. Works fine with 1.8.2.

afbjorklund:
This is a design decision (for the `docker` driver), and not really a bug per se.

@medyagh : can you comment why it uses a tunnel rather than the bridge ?

andrebrait:
Ok. Is there another way to reach the cluster from the host then?

afbjorklund:
> Ok. Is there another way to reach the cluster from the host then?

I think you would have to use `minikube tunnel`, but unsure myself.

By default it will only tunnel 2376 (docker) and 8443 (kubernetes), in addition to the regular 22 (ssh):

```
CONTAINER ID        IMAGE                                COMMAND                  CREATED             STATUS              PORTS                                                                           NAMES
a4633bf7bde4        gcr.io/k8s-minikube/kicbase:v0.0.8   "/usr/local/bin/entr…"   2 hours ago         Up 2 hours          127.0.0.1:32770->22/tcp, 127.0.0.1:32769->2376/tcp, 127.0.0.1:32768->8443/tcp   minikube
```

When using something like VirtualBox, it will do two _different_ networks - one for ssh, one for "other"

[...]

PR #9496, opened 2020-10-20
doc: update minikube quick start in macos

AlphaWong:
LoadBalancer issue in macOS for MiniKube

# reference
https://github.com/kubernetes/minikube/issues/7344#issuecomment-688179776

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -90,6 +90,11 @@ From a terminal with administrator access (but not logged in as root), run:
 
 ```shell
 minikube start
+
+# if you want to use LoadBalancer in macOS
+# reference : https://github.com/kubernetes/minikube/issues/7344#issuecomment-688179776
+
+minikube start --vm-driver=hyperkit
 ```
 
 If minikube fails to start, see the [drivers page]({{<ref "/docs/drivers">}}) for help setting up a compatible container or virtual-machine manager.


tstromberg (changes requested):
This doesn't seem right to me - LoadBalancer should work on macOS with all drivers.

AlphaWong:
> This doesn't seem right to me - LoadBalancer should work on macOS with all drivers.

I get a wanting in default docker driver 

```
Because you are using a Docker driver on darwin, the terminal needs to be open to run it.
```

People inside https://github.com/kubernetes/minikube/issues/7344#issuecomment-688179776 also share the same experience. It seems not working in current version.

Which version u testing with?

medyagh (changes requested):
I think you meant to add this to docker driver docs, in MacOS.

https://minikube.sigs.k8s.io/docs/drivers/docker/#known-issues

AlphaWong:
> I think you meant to add this to docker driver docs, in MacOS.
> 
> https://minikube.sigs.k8s.io/docs/drivers/docker/#known-issues

I agree that we should put it at other page but end up the quick start from https://minikube.sigs.k8s.io/docs/start/ to https://minikube.sigs.k8s.io/docs/drivers/ is not quick enough also it prefers `docker` driver at MacOS. 

Seems more people like me as a new joiner we would like to get our local workplace up and run first. I believe some amendment or hints is required to help new user.

The real case is that I would like to setup a deployment with nginx ingress with my local Minikube for testing and I find the `localBalance` stuck at pending. The resolution needs to destroy the whole Minikube vm and reinstall is not that ideal in my concern.

Therefore, hints the user about the `loadBalance` issue or vm type or know issue can save a lot of time as it is not all the people know the history or know issue or read all the thread about Minikube. As Minikube is a good project and almost is a good learning playground for anyone I hold my opinion for adding any hints or remark on it.

Moreover, after google introduce https://github.com/GoogleContainerTools/skaffold for ensure the local and remote working environment alignment, more and more people install Minikube in their local. It is ready better to have some hints about it.

see the reference below and the screenshot.

# screenshot


# reference
https://github.com/kubernetes/minikube/issues/7344#issuecomment-727538130

afbjorklund (changes requested):
This should be more general, that you still need to use --vm=true if you want working network...

The docker driver is supposed to be able to limp along using `minikube tunnel` and such though.

priyawadhwa:
Hey @AlphaWong are you still working on this? I agree that `--vm=true` would be better as it is more generic.

priyawadhwa:
Hey @AlphaWong this PR has been quiet for a few months now, so I'm going to close it for now. Feel free to reopen whenever you have a chance to take a look!

PR #9501, opened 2020-10-20
revert unneeded hack for crio runtime

sharifelgamal:


sharifelgamal:
crio still looks broken, do we need to upgrade the version?

tstromberg:
> crio still looks broken, do we need to upgrade the version?

I assume we should probably close this PR then and follow up with the crio folks about the issue. Is there something else you want done in this code review?

sharifelgamal:
Yep, I'm closing this until we figure out what's going on.

Issue #9271, opened 2020-09-17
Enable scheduled stop 


priyawadhwa:
Top level issue tracking the following:

- [x] Implement schedule stop for unix https://github.com/kubernetes/minikube/pull/9503
- [x] implement scheduled stop on windows (https://github.com/kubernetes/minikube/pull/9689)
- [x] Add --cancel-scheduled-stop flag to cancel all existing scheduled stops https://github.com/kubernetes/minikube/issues/9730
- [x] `minikube status` should show existing scheduled stops and time left https://github.com/kubernetes/minikube/issues/9731

PR #9503, opened 2020-10-20
Implement schedule stop for unix

priyawadhwa:
Implement scheduled stop for mac/linux. Still need to figure out windows, but figured we could at least get the first piece of this in.

This solution uses the https://github.com/VividCortex/godaemon library, which daemonizes the current process and assigns it a new PID.

Users can schedule a stop via:

```
minikube stop --schedule 4m
```

once this is happens, minikube will:

1. Cancel any previous scheduled stops by reading their stored PIDs from a file in `.minikube`
1. Store the initial time of the command and the wait time in the profile config. This can later be used to tell users how much time is left if they call `minikube status`
1. Daemonize the current process and store the new PID of the process in `.minikube`
1. Wait the required amount of time before following through with stop

To get this to work with the VividCortex library, I did have to create my own fork to remove a CGO dependency that was preventing cross compilation. Currently, minikube points to my fork so that `make cross` succeeds. I have a PR open to remove the dependency here: https://github.com/VividCortex/godaemon/pull/32. Once that merges, we can remove the dependency on my fork.

https://github.com/kubernetes/minikube/issues/9271


cmd/minikube/cmd/stop.go
```diff
@@ -81,6 +84,18 @@ func runStop(cmd *cobra.Command, args []string) {
 		profilesToStop = append(profilesToStop, cname)
 	}
 
+	if scheduledStop != "" {
```
medyagh, line 87: we should check if windows, then exit with usge that is not implemented on windows yet

medyagh (changes requested):

cmd/minikube/cmd/stop.go
priyawadhwa: done

pkg/minikube/config/types.go
```diff
@@ -138,3 +139,9 @@ type VersionedExtraOption struct {
 	// flag is applied to
 	GreaterThanOrEqual semver.Version
 }
+
+// ScheduledStopConfig contains information around scheduled stop
```
medyagh, line 143: comment that code is not used yet.

test/integration/scheduled_stop_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/docker/machine/libmachine/state"
+	"k8s.io/minikube/pkg/minikube/localpath"
+	"k8s.io/minikube/pkg/util/retry"
+)
+
+func TestScheduledStop(t *testing.T) {
```
medyagh, line 37: take care of none diver

site/content/en/docs/commands/stop.md
```diff
@@ -23,6 +23,7 @@ minikube stop [flags]
       --all                   Set flag to stop all profiles (clusters)
       --keep-context-active   keep the kube-context active after cluster is stopped. Defaults to false.
   -o, --output string         Format to print stdout in. Options include: [text,json] (default "text")
+      --schedule duration     Set flag to stop cluster after a set amount of time (e.g. --schedule=5m)
```
medyagh, line 26: hiddden flag

priyawadhwa (commented):
1. Clean up processes on `minikube stop`
1. None driver
1. Make flag hidden

priyawadhwa:
Tested locally on mac (docker, hyperkit), linux (kvm2) and cloud shell

site/content/en/docs/commands/stop.md
priyawadhwa: fixed

test/integration/scheduled_stop_test.go
priyawadhwa: thanks, done

pkg/minikube/config/types.go
priyawadhwa: done

pkg/minikube/schedule/daemonize_unix.go
```diff
@@ -0,0 +1,80 @@
...
+	"strconv"
+	"time"
+
+	"github.com/VividCortex/godaemon"
+	"github.com/pkg/errors"
+	"k8s.io/klog/v2"
+	"k8s.io/minikube/pkg/minikube/localpath"
+)
+
+// KillExisting kills existing scheduled stops
```
tstromberg, line 34: Can you clarify this comment? It isn't clear what "killing" a "stop" is. 

I assume this is killing some sort of nanny process?

pkg/minikube/schedule/daemonize_unix.go
```diff
@@ -0,0 +1,80 @@
...
+func KillExisting(profiles []string) error {
+	for _, profile := range profiles {
+		file := localpath.PID(profile)
+		f, err := ioutil.ReadFile(file)
+		if os.IsNotExist(err) {
+			return nil
+		}
+		defer os.Remove(file)
```
tstromberg, line 42: Make sure to handle the error returned by os.Remove, using klog.Errorf if nothing else. Otherwise we'll have stale files laying around silently.

pkg/minikube/schedule/daemonize_unix.go
```diff
@@ -0,0 +1,80 @@
...
+		if err != nil {
+			return errors.Wrapf(err, "reading %s", file)
+		}
+		pid, err := strconv.Atoi(string(f))
+		if err != nil {
+			return errors.Wrapf(err, "converting %v to int", string(f))
```
tstromberg, line 48: to simply, I think you can use '%s' and f

pkg/minikube/schedule/daemonize_unix.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		p, err := os.FindProcess(pid)
+		if err != nil {
+			return errors.Wrap(err, "finding process")
+		}
+		klog.Infof("killing process %v as it is an old scheduled stop", pid)
+		if err := p.Kill(); err != nil {
+			return errors.Wrapf(err, "killing %v", pid)
```
tstromberg, line 56: This should keep processing subsequent pids instead of returning early. I think minikube is already importing some sort of library that gets the process name for a pid -- check the hyperkit driver.

Otherwise, if you reboot, this function will never work once it hits a stale pid.

pkg/minikube/schedule/daemonize_unix.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		defer os.Remove(file)
+		if err != nil {
+			return errors.Wrapf(err, "reading %s", file)
+		}
+		pid, err := strconv.Atoi(string(f))
+		if err != nil {
+			return errors.Wrapf(err, "converting %v to int", string(f))
+		}
+		p, err := os.FindProcess(pid)
```
tstromberg, line 50: Please check that the pid is minikube process before killing it. Otherwise this will kill random processes.

pkg/minikube/schedule/daemonize_unix.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	// now that this process has daemonized, it has a new PID
+	pid := os.Getpid()
+	return savePIDs(pid, profiles)
+}
+
+func savePIDs(pid int, profiles []string) error {
+	for _, p := range profiles {
+		file := localpath.PID(p)
+		if err := ioutil.WriteFile(file, []byte(fmt.Sprintf("%v", pid)), 0644); err != nil {
```
tstromberg, line 75: This should probably be 0600. I can't see why other users will need write permissions.

tstromberg (changes requested):

pkg/minikube/schedule/daemonize_unix.go
priyawadhwa: When `minikube stop` daemonizes, it saves its own PID via `os.Getpid()` to a file, and that is the PID that is being killed here.

I'm not sure how a random process could be killed from this code, could you explain if you still think that's the case?

pkg/minikube/schedule/daemonize_unix.go
priyawadhwa: That's a good point, updated to process all pids regardless of if some error out

pkg/minikube/schedule/daemonize_unix.go
priyawadhwa: thanks!

pkg/minikube/schedule/daemonize_unix.go
priyawadhwa: Yup, killing background processes that may still exist from previous attempts to schedule stops. Made that more clear.

medyagh (approved):

PR #9523, opened 2020-10-22
Upgrade buildroot and kernel minor version

afbjorklund:
http://lists.busybox.net/pipermail/buildroot/2020-October/294451.html

Our configuration still has a 4.19 kernel, rather than the default 5.4 kernel


afbjorklund:
Would be nice to get #9394 in before this one, to restore the kernel config along with bumping the minor version

I used the same kernel version as buildroot did (for 4.19), but I'm not sure how much it matters (150 vs 152 etc).

afbjorklund:
Did some basic smoke testing locally, seemed OK

🐳  Preparing Kubernetes v1.19.2 on Docker 19.03.13 ..

🎁  Preparing Kubernetes v1.19.2 on CRI-O 1.18.3 ...

`Linux minikube 4.19.150 #1 SMP Thu Oct 22 21:56:19 CEST 2020 x86_64 GNU/Linux`

Issue #9486, opened 2020-10-18
stopping a profile which doesn't exists creates profile folder instead of just throwing error


vinu2003:
##minikube stop##
vinodhinis-mbp:minikube vinodhinibalusamy$ ./out/minikube stop -p test2
🤷  There is no local cluster named "test2"
👉  To fix this, run: "minikube start -p test2"

this causes unwanted error message thrown in profile list.

vinodhinis-mbp:minikube vinodhinibalusamy$ ./out/minikube profile list
|----------|-----------|---------|--------------|------|---------|---------|
| Profile  | VM Driver | Runtime |      IP      | Port | Version | Status  |
|----------|-----------|---------|--------------|------|---------|---------|
| minikube | docker    | docker  | 192.168.49.2 | 8443 | v1.19.2 | Running |
| test1    | docker    | docker  | 192.168.59.2 | 8443 | v1.19.2 | Running |
|----------|-----------|---------|--------------|------|---------|---------|
❗  Found 2 invalid profile(s) ! 
	 test2
	 ttt
💡  You can delete them using the following command(s): 
	 $ minikube delete -p test2 
	 $ minikube delete -p ttt 
vinodhinis-mbp:minikube vinodhinibalusamy$ 

And i did notice /Users/vinodhinibalusamy/.minikube/profiles has test2 folder created which shouldn't be.

tstromberg:
Good catch! Help wanted to fix this -- there is no reason why `minikube stop` should cause a profile to be created.

vinu2003:
thank you.
/assign @vinu2003

PR #9534, opened 2020-10-23
stop: removed setEventLogPath for stop process

vinu2003:
##what##
This PR address code changes that eliminates the profile folder creation if that profile doesn't exist.
Basically it removes the setEventLogPath function which creates a folder for tracking the transient events. In case of stop function, if the profile is valid the file exists in the profile folder. So it's not required to create new one.

Fixes: #9486

### DIFF
--- a/cmd/minikube/cmd/stop.go
+++ b/cmd/minikube/cmd/stop.go
@@ -65,7 +65,6 @@ func init() {
 // runStop handles the executes the flow of "minikube stop"
 func runStop(cmd *cobra.Command, args []string) {
 	out.SetJSON(outputFormat == "json")
-	register.SetEventLogPath(localpath.EventLog(ClusterFlagValue()))
 	register.Reg.SetStep(register.Stopping)
 
 	// new code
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/stop.go
```diff
@@ -65,7 +65,6 @@ func init() {
 // runStop handles the executes the flow of "minikube stop"
 func runStop(cmd *cobra.Command, args []string) {
 	out.SetJSON(outputFormat == "json")
-	register.SetEventLogPath(localpath.EventLog(ClusterFlagValue()))
```
medyagh, line 68: lets check if if the path exists, then set it, so we still write the events to the event log

medyagh (changes requested):
thank you, this PR is doing the right this, with one minor request/

cmd/minikube/cmd/stop.go
vinu2003: Makes sense @medyagh. Thanks.

vinu2003: Question: Is there a case where a profile folder exists without event.json file in it. I believe it exists most of the time but still the SetEventtLogPath function creates one if it doesn't exists which means some corner cases.

tstromberg (approved):

Issue #8596, opened 2020-06-29
Podman v2 doesn't work with podman-env


afbjorklund:
**If you want to build images using `podman`, you will have to use release 1.9.3**

The earlier releases (1.8.2) hangs, and the later releases (2.0.0) gives an error.

It _should_ also be OK to use the 1.6.x releases, available in many distributions.

The "varlink" socket is not supported anymore, so need to change minikube for it.

----

https://podman.io/blogs/2020/08/01/deprecate-and-remove-varlink-notice.html

> ... Podman API v1.0 for Podman is considered to be deprecated. If there are issues with the Podman API v1.0 in versions of Podman prior to v2.0 and those versions are still under support on Red Hat Enterprise Linux (RHEL), the Podman team will make a best effort to address those issues. However, no new feature requests for the API v1.0 will be considered and any problems found with the API v1.0 in Podman v2.0 will not be addressed.

> This also serves as a notification that the Podman v1.0 (varlink) API will be removed from the main GitHub branch of Podman in the near future. With the release of Podman v2.0 the Podman developers deprecated the Podman API v1.0 in favor of the new Podman v2.0 RESTful API. The plan is to remove varlink completely from the Podman v3.0 development branch which will be created some time after September 2020.

Original post: https://podman.io/blogs/2020/06/29/podman-v2-announce.html

afbjorklund:
Binaries are available here: https://github.com/containers/libpod/releases/tag/v1.9.3

The packages in kubic repositories unfortunately only have the latest-and-greatest

afbjorklund:
I was trying again with podman 2.0.5, but now there is not even an error.

It just fails silently... Trying to build over ssh (as a workaround also fails)

afbjorklund:
The plan to replace sudo, is to make the socket owned by a local group:

https://github.com/containers/podman/issues/6809#issuecomment-651237669

And then make the default user part of this group (i..e same as for docker)

```bash
    mkdir -p /etc/systemd/system/podman.socket.d
    cat >/etc/systemd/system/podman.socket.d/override.conf <<EOF
[Socket]
SocketMode=0660
SocketUser=root
SocketGroup=podman
EOF
    systemctl daemon-reload
    echo "d /run/podman 0770 root podman" > /etc/tmpfiles.d/podman.conf
    systemd-tmpfiles --create
```

The original settings are in:

/usr/lib/systemd/system/podman.socket

```ini
[Unit]
Description=Podman API Socket
[...]

PR #9535, opened 2020-10-23
Add support for podman v2 to podman-env command

afbjorklund:
Podman v1 (like 1.9.3) uses the varlink bridge,
while podman v2 (like 2.1.x) uses a REST API.

Both are using ssh for their transport, but the
environment variables exported are different.

Not using `sudo` anymore, so user needs root access.

See #8596 for how to set that up for the "docker" user

Tested OK with 1.9.3 and 2.1.1 minikube profiles.

But don't recommend using podman 2.0.x (at all)

### DIFF
--- a/cmd/minikube/cmd/podman-env.go
+++ b/cmd/minikube/cmd/podman-env.go
@@ -208,34 +208,7 @@ func podmanSetScript(ec PodmanEnvConfig, w io.Writer) error {
 
 // podmanUnsetScript writes out a shell-compatible 'podman-env unset' script
 func podmanUnsetScript(ec PodmanEnvConfig, w io.Writer) error {
-	// podman v1
-	vars1 := []string{
-		constants.PodmanVarlinkBridgeEnv,
-	}
-	// podman v2
-	vars2 := []string{
-		constants.PodmanContainerHostEnv,
-		constants.PodmanContainerSSHKeyEnv,
-	}
-	// common
-	vars0 := []string{
-		constants.MinikubeActivePodmanEnv,
-	}
-
-	var vars []string
-	if ec.client != nil || ec.hostname != "" {
-		// getting ec.varlink needs a running machine
-		if ec.varlink {
-			vars = vars1
-		} else {
-			vars = vars2
-		}
-	} else {
-		// just unset *all* of the variables instead
-		vars = vars1
-		vars = append(vars, vars2...)
-	}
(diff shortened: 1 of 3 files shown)


afbjorklund:
Dear @TravisBuddy it is just a Travis flake

cmd/minikube/cmd/podman-env.go
```diff
@@ -39,12 +39,14 @@ import (
 	"k8s.io/minikube/pkg/minikube/shell"
 )
 
-var podmanEnvTmpl = fmt.Sprintf("{{ .Prefix }}%s{{ .Delimiter }}{{ .VarlinkBridge }}{{ .Suffix }}{{ .Prefix }}%s{{ .Delimiter }}{{ .MinikubePodmanProfile }}{{ .Suffix }}{{ .UsageHint }}", constants.PodmanVarlinkBridgeEnv, constants.MinikubeActivePodmanEnv)
+var podmanEnvTmpl = fmt.Sprintf("{{ if .VarlinkBridge }}{{ .Prefix }}%s{{ .Delimiter }}{{ .VarlinkBridge }}{{ .Suffix }}{{end}}{{ if .ContainerHost }}{{ .Prefix }}%s{{ .Delimiter }}{{ .ContainerHost }}{{ .Suffix }}{{end}}{{ if .ContainerSSHKey }}{{ .Prefix }}%s{{ .Delimiter }}{{ .ContainerSSHKey}}{{ .Suffix }}{{ end }}{{ .Prefix }}%s{{ .Delimiter }}{{ .MinikubePodmanProfile }}{{ .Suffix }}{{ .UsageHint }}", constants.PodmanVarlinkBridgeEnv, constants.PodmanContainerHostEnv, constants.PodmanContainerSSHKeyEnv, constants.MinikubeActivePodmanEnv)
```
priyawadhwa, line 42: It might be worth it to separate the templates into `podmanEnvTmplV1` and `podmanEnvmTmplV2` -- that way it'll be easier to read and maintain. And it will follow the current pattern we have with kubeadm configs

afbjorklund: Possibly ? Could do the same thing for docker, with `ssh:` vs `tcp:` ?

afbjorklund: Done, also refactored the two functions to look more similar

medyagh (approved):

PR #9530, opened 2020-10-22
docker: When creating networks, use MTU of built-in bridge network

medyagh:
closes https://github.com/kubernetes/minikube/issues/9528

### After this PR
```
medya@cloudshell:~$ ./minikube-linux-amd64 ssh
docker@minikube:~$ docker pull golang:1.15
1.15: Pulling from library/golang
e4c3d3e4f7b0: Pull complete
101c41d0463b: Pull complete
8275efcd805f: Pull complete
751620502a7a: Pull complete
aaabf962c4fc: Pull complete
7883babec904: Pull complete
1791d366c848: Pull complete
Digest: sha256:1ba0da74b20aad52b091877b0e0ece503c563f39e37aa6b0e46777c4d820a2ae
Status: Downloaded newer image for golang:1.15
docker.io/library/golang:1.15
docker@minikube:~$ 
docker@minikube:~$
docker@minikube:~$
docker@minikube:~$
docker@minikube:~$ exit
logout
```

```
gout
medya@cloudshell:~$ docker network inspect minikube
[
    {
        "Name": "minikube",
        "Id": "1d1e52f41e1b8a9ca89c3aba16a9d318251199f57a4b8d15769e1d0673685223",
        "Created": "2020-10-23T21:19:26.083577921Z",
        "Scope": "local",
        "Driver": "bridge",
        "EnableIPv6": false,
        "IPAM": {
            "Driver": "default",
            "Options": {},
            "Config": [
                {
                    "Subnet": "192.168.49.0/24",
                    "Gateway": "192.168.49.1"
                }
            ]
        },
        "Internal": false,
        "Attachable": false,
        "Ingress": false,
        "ConfigFrom": {
            "Network": ""
        },
        "ConfigOnly": false,
        "Containers": {
            "8d3585732ce0878bf5c85fabb3be2e0b72ea181baa7fc3f0299c0188c2ff9d1f": {
                "Name": "minikube",
                "EndpointID": "5810c510b63c76af8d8759e92d72624dc008f06b7eadb2f5aadae43b6bfea146",
                "MacAddress": "02:42:c0:a8:31:02",
                "IPv4Address": "192.168.49.2/24",
                "IPv6Address": ""
            }
        },
        "Options": {
            "--icc": "",
            "--ip-masq": "",
            "com.docker.network.driver.mtu": "1460"
        },
        "Labels": {
            "created_by.minikube.sigs.k8s.io": "true"
        }
    }
]
```

Issue #9528, opened 2020-10-22
docker: pulls may fail if MTU is <1500


medyagh:
on cloud shell if we dont set this paramter, docker pull on "some images" hangs... (for example on golang:1.15) and not on other images such as alpine ...

```
sudo sysctl -w net.ipv4.tcp_mtu_probing=1
```

example:

```
medya@cloudshell:~$ minikube ssh
docker@minikube:~$ docker pull golang:1.15
^C
docker@minikube:~$ sudo sysctl -w net.ipv4.tcp_mtu_probing=1
net.ipv4.tcp_mtu_probing = 1
docker@minikube:~$ docker pull golang:1.15
1.15: Pulling from library/golang
e4c3d3e4f7b0: Pull complete
101c41d0463b: Pull complete
8275efcd805f: Pull complete
751620502a7a: Pull complete
aaabf962c4fc: Pull complete
7883babec904: Pull complete
1791d366c848: Pull complete
Digest: sha256:1ba0da74b20aad52b091877b0e0ece503c563f39e37aa6b0e46777c4d820a2ae
Status: Downloaded newer image for golang:1.15
docker.io/library/golang:1.15
```

we need to investigate does it make sense to set this by default in our image/iso ?

medyagh:
this blog post suggests setting the TPU to 1400 would help

https://medium.com/@sylwit/how-we-spent-a-full-day-figuring-out-a-mtu-issue-with-docker-4d81fdfe2caf

for the records our TPUs are 1500
```
docker@minikube:~$ ip addr | grep mtu
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
2: docker0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default
4: veth0713d6b@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master docker0 state UP group default
26: eth0@if27: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default
```

tstromberg:
The MTU probing approach can definitely help. MTU's of less than 1500 are kind of exotic in 2020, but are definitely seen within GCP.

One alternative/addition is that the Docker driver could check the MTU of the bridge, and copy it to the custom network it creates.

[...]

PR #9537, opened 2020-10-23
Enable TCP Path MTU Discovery when an ICMP black hole is detected

tstromberg:
Partial fix for #9528 - the longer-term fix will come from #9530

This lets `docker pull` work when the network MTU is misconfigured. There are no apparent downsides to this approach, as best as I can tell. Here's the documentation from https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt

```
tcp_mtu_probing - INTEGER
	Controls TCP Packetization-Layer Path MTU Discovery.  Takes three
	values:
	  0 - Disabled (DEFAULT)
	  1 - Disabled by default, enabled when an ICMP black hole detected
	  2 - Always enabled, use initial MSS of tcp_base_mss.
```

NOTE That `tcp_base_mss` defaults to 1024. I believe that 1 is the best default going forward.

I can confirm that this allows `minikube ssh "docker pull golang:1.15"` to work properly on Cloud Shell. 

## Old behavior

```
minikube ssh "sysctl net.ipv4.tcp_mtu_probing"                                                                                                                              net.ipv4.tcp_mtu_probing = 0
```

## New behavior

```
minikube ssh "sysctl net.ipv4.tcp_mtu_probing"
net.ipv4.tcp_mtu_probing = 1
```

### DIFF
--- a/deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/etc/sysctl.d/11-tcp-mtu-probing.conf
+++ b/deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/etc/sysctl.d/11-tcp-mtu-probing.conf
@@ -1,0 +1,3 @@
+# Turn on MTU probing to avoid network hangs when the Docker MTU is larger than
+# the host or upstream network MTU.
+net.ipv4.tcp_mtu_probing=1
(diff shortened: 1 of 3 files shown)


priyawadhwa (approved):

medyagh (commented):
isn't it better to be added to both ISO and kicbase at same time?
this PR would need a new kic-base image, that I believe has some other changes, for a point release we might wanna avoid that since we have not tested the head kicbase

tstromberg:
> isn't it better to be added to both ISO and kicbase at same time?
> this PR would need a new kic-base image, that I believe has some other changes, for a point release we might wanna avoid that since we have not tested the head kicbase

Done.

PR #9538, opened 2020-10-23
Add --network flag to select docker network to run with docker driver

priyawadhwa:
Users can now select the network they want to run the docker driver with. If left empty, we will default to creating a new network so that static IP changes can be maintained. Users can go back to using default network by running:

```
minikube start --docker-network=bridge
```


tstromberg (approved):

cmd/minikube/cmd/start_flags.go
```diff
@@ -108,6 +108,7 @@ const (
 	kicBaseImage            = "base-image"
 	startOutput             = "output"
 	ports                   = "ports"
+	dockerNetwork           = "docker-network"
```
medyagh, line 111: this can be generalized for other drivers such as Podman
there is no need for it to be called docker-network

medyagh (changes requested):
there should be added a warnning if using Bridge network, minikube will not have static IP.

since bridge network is not controlable by non-docker itsef

cmd/minikube/cmd/start_flags.go
priyawadhwa: would you prefer `--container-network`? just `--network` seems too general to me

medyagh: I think --network is good enough, other drivers actually can implement this feature, we can later add same option for KVM and others.
but till then we should Exit with usage and say this feature is only currently suported in docker driver.

priyawadhwa:
@medyagh now that we have the MTU fix in, do you think it's worth it to add this flag? I'm worried it won't be particularly useful and we're just adding another flag option to `minikube start`

tstromberg:
> @medyagh now that we have the MTU fix in, do you think it's worth it to add this flag? I'm worried it won't be particularly useful and we're just adding another flag option to `minikube start`

I think this flag would still be useful, particularly for cases where Docker is running remotely. I had wanted to recommend this flag to a WSL2 user running into networking problems today :(

cmd/minikube/cmd/start_flags.go
priyawadhwa: done

cmd/minikube/cmd/start_flags.go
```diff
@@ -286,12 +288,17 @@ func generateClusterConfig(cmd *cobra.Command, existing *config.ClusterConfig, k
 			out.WarningT("With --network-plugin=cni, you will need to provide your own CNI. See --cni flag as a user-friendly alternative")
 		}
 
+		if !driver.IsKIC(drvName) && viper.GetString(containerNetwork) != "" {
+			exit.Advice(reason.Usage, "--network flag is only valid with the docker driver", "remove the --network flag from `minikube start`")
```
medyagh, line 292: I am thinking how about we only warn the user and not exit ? like if they have minikube on auto select driver,
but they want to use this driver if Minikube chooses docker....

I think we should change to out.Warrning

medyagh (changes requested):

cmd/minikube/cmd/start_flags.go
priyawadhwa: Personally I think it makes sense to exit since it forces correct usage of the flag, and minimizes potential confusion.

tstromberg: I agree -- we should not ignore the user's request.

tstromberg (approved):

cmd/minikube/cmd/start_flags.go
```diff
@@ -291,12 +293,17 @@ func generateClusterConfig(cmd *cobra.Command, existing *config.ClusterConfig, k
...
+			out.WarningT("--network flag is only valid with the docker driver, it will be ignored")
+		}
+
 		cc = config.ClusterConfig{
 			Name:                    ClusterFlagValue(),
 			KeepContext:             viper.GetBool(keepContext),
 			EmbedCerts:              viper.GetBool(embedCerts),
 			MinikubeISO:             viper.GetString(isoURL),
 			KicBaseImage:            viper.GetString(kicBaseImage),
+			ContainerNetwork:        viper.GetString(containerNetwork),
```
priyawadhwa, line 306: Rename to Network

test/integration/docker_network_test.go
```diff
@@ -0,0 +1,59 @@
...
+		t.Skip("only runs with docker driver")
+	}
+
+	tests := []struct {
+		description string
+		flag        string
+	}{
+		{
+			description: "create custom network",
+			flag:        "--docker-network=\"\"",
```
medyagh, line 38: it is not clear that what is this test case, 

do we epect if they specify "\" as network name to create a network with that name?
if thats the case we dont verify in the test that we do create that network.

pkg/drivers/kic/oci/network_create.go
```diff
@@ -48,9 +49,15 @@ func CreateNetwork(ociBin string, name string) (net.IP, error) {
 	return createDockerNetwork(name)
 }
 
-func createDockerNetwork(clusterName string) (net.IP, error) {
+func createDockerNetwork(networkName string) (net.IP, error) {
+	// check if the network is docker native, if this is the case return
+	if networkName == "bridge" {
```
medyagh, line 54: use the contantant that we have for this
```
const dockerDefaultBridge = "bridge"
```

medyagh (changes requested):

test/integration/docker_network_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+package integration
+
+import (
+	"context"
+	"os/exec"
+	"testing"
+)
+
+func TestDockerNetworkFlag(t *testing.T) {
```
medyagh, line 27: name could be 

"KicCustomNetwork"

(podman network PR is WIP, soon it will be able to test both)

medyagh (changes requested):

test/integration/docker_network_test.go
priyawadhwa: if we say "" then it defaults to creating its own network, but i can remove this if it's confusing

pkg/drivers/kic/oci/network_create.go
priyawadhwa: thanks

test/integration/docker_network_test.go
priyawadhwa: sure, done

test/integration/kic_custom_network_test.go
```diff
@@ -0,0 +1,58 @@
...
+	if !KicDriver() {
+		t.Skip("only runs with docker driver")
+	}
+
+	tests := []struct {
+		description string
+		flag        string
+	}{
+		{
+			description: "create custom network",
```
medyagh, line 37: this doesn't seem to add any value to have just a "standard start" since we have that covered in other tests

medyagh: consider adding a network name other than bridge

for example "custom-network-1234"

test/integration/kic_custom_network_test.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
+			profile := UniqueProfileName("docker-network")
+			ctx, cancel := context.WithTimeout(context.Background(), Minutes(5))
+			defer Cleanup(t, profile, cancel)
+
+			startArgs := []string{"start", "-p", profile, test.flag}
+			c := exec.CommandContext(ctx, Target(), startArgs...)
```
medyagh, line 51: after this we should verify that the network specified exists or created

consider checking using "docker inspect custom-network-1234" to verify its existance

medyagh (changes requested):

test/integration/docker_network_test.go
medyagh: yeah I think this test be removed,

test/integration/kic_custom_network_test.go
priyawadhwa, line 37: make sure we correctly create the network for this test

test/integration/kic_custom_network_test.go
```diff
@@ -0,0 +1,58 @@
...
+	tests := []struct {
+		description string
+		flag        string
+	}{
+		{
+			description: "create custom network",
+		}, {
+			description: "use default bridge network",
+			flag:        "--docker-network=bridge",
+		},
```
priyawadhwa, line 41: add test where we use a precreated network

medyagh (changes requested):
I suggest writing an intergation test that
creates a docker network 
docker network create example

and then make minikube use that Existing network

minikube start --network=existing network

and in another test, ask minikube to specify the network name but it should not exist and minikube should create it

we should verify the creation of the network using

docker network inspect <name>

medyagh (changes requested):
plz check failed test
use_default_bridge_network
" is supported on user defined networks only."

we can't specify ip on default netwokr

tstromberg:
@priyawadhwa - do you mind resolving the conflict and looking at the failing test @medyagh mentioned?

site/content/en/docs/commands/start.md
```diff
@@ -77,6 +77,7 @@ minikube start [flags]
       --namespace string                  The named space to activate after start (default "default")
       --nat-nic-type string               NIC Type used for nat network. One of Am79C970A, Am79C973, 82540EM, 82543GC, 82545EM, or virtio (virtualbox driver only) (default "virtio")
       --native-ssh                        Use native Golang SSH client (default true). Set to 'false' to use the command line 'ssh' command when accessing the docker machine. Useful for the machine drivers when they will not start with 'Waiting for SSH'. (default true)
+      --network string                    Docker network to run minikube with. Only available with the docker driver. If left empty, minikube will create a new network.
```
medyagh, line 80: please add Podman here, we added podman in last version of minikube.

medyagh (changes requested):

site/content/en/docs/commands/start.md
priyawadhwa: sure, done

cmd/minikube/cmd/start_flags.go
```diff
@@ -293,12 +295,17 @@ func generateClusterConfig(cmd *cobra.Command, existing *config.ClusterConfig, k
 			out.WarningT("With --network-plugin=cni, you will need to provide your own CNI. See --cni flag as a user-friendly alternative")
 		}
 
+		if !driver.IsKIC(drvName) && viper.GetString(network) != "" {
+			out.WarningT("--network flag is only valid with the docker driver, it will be ignored")
```
medyagh, line 299: + podman driver here

medyagh (changes requested):

medyagh:
thank you for this great PR @priyawadhwa

medyagh (approved):

Issue #7772, opened 2020-04-18
rename none driver to baremetal


medyagh:
none is a terrible name.
and doesnt look good at least in integration tests.I suggest we add an alias to none driver. 

we could do it in two phases:

change all the current integration tests names to BareMetal 
in second phase, if we add an alias, so both none or barmetal be accepted.

afbjorklund:
> none is a terrible name.

No, it is just out of context...

I think this has been described in multiple places, but this is because it was _changed_ from machine:

I'll copy it here (from the "generic" [discussion](https://github.com/kubernetes/minikube/pull/6873#issuecomment-615440542), another legacy driver name used by `docker-machine`):

### none

> And as we discussed earlier, "none" comes for Docker hosts that didn't have _any_ driver.
> 
> https://docs.docker.com/machine/get-started-cloud/#add-a-host-without-a-driver
> 
> > You can register an already existing docker host by passing the daemon url. With that, you can have the same workflow as on a host provisioned by docker-machine.

### generic

> Basically the name "generic" originates from being the driver for any existing Cloud Provider...
> 
> https://docs.docker.com/machine/get-started-cloud/#3rd-party-driver-plugins
> 
> * [docker.github.io/machine/AVAILABLE_DRIVER_PLUGINS.md](https://github.com/docker/docker.github.io/blob/master/machine/AVAILABLE_DRIVER_PLUGINS.md)
> 
> So if your cloud didn't have a built-in driver, you could provision your VM and give it to machine.
> 
> https://docs.docker.com/machine/get-started-cloud/#drivers-for-cloud-providers
> 
> * https://docs.docker.com/machine/drivers/generic/
> 
> > Create machines using an existing VM/Host with SSH.
> >
> > This is useful if you are using a provider that Machine does not support directly or if you would like to import an existing host to allow Docker Machine to manage.

### machine

> The terms from docker-machine can be confusing when it comes to minikube...
> 
> Since minikube assumes a local docker, and doesn't support any cloud providers.

Now that we have forked the "none" driver, we can rename it like we did with the "kvm2" driver.

But naming things is hard, even if we long considered renaming it as the "libvirt" driver instead.

This is also partly due to #4829

We lack technical documentation...

----

So, naming.

The problem is that "BareMetal" is **also** a horrible name. There are two things wrong with it:

The first one is philosophical, and that is that [bare metal](https://en.wikipedia.org/wiki/Bare_metal_(disambiguation)) usually came _before_ the OS itself.
[...]

PR #9545, opened 2020-10-24

afbjorklund:
Let's try another rebase.... (PR #6873, PR #4734)

Implements #4733


afbjorklund:
These comments still apply:

https://github.com/kubernetes/minikube/pull/6873#issuecomment-594176675

https://github.com/kubernetes/minikube/pull/6873#issuecomment-615449617

It would be nice to clean those up, perhaps by completing the fork of the "generic" driver from libmachine.

Maybe something that could be done at the same time as completing the fork of the "none" driver (#7772)

cmd/minikube/cmd/start_flags.go
```diff
@@ -212,6 +217,12 @@ func initNetworkingFlags() {
 	startCmd.Flags().String(serviceCIDR, constants.DefaultServiceCIDR, "The CIDR to be used for service cluster IPs.")
 	startCmd.Flags().StringArrayVar(&config.DockerEnv, "docker-env", nil, "Environment variables to pass to the Docker daemon. (format: key=value)")
 	startCmd.Flags().StringArrayVar(&config.DockerOpt, "docker-opt", nil, "Specify arbitrary flags to pass to the Docker daemon. (format: key=value)")
+
+	// generic
+	startCmd.Flags().String(genericIPAddress, "", "IP address (generic)")
```
priyawadhwa, line 222: We should validate that all of these flags are set and exit with a helpful error message if the user has forgotten anything

pkg/minikube/machine/fix.go
```diff
@@ -129,26 +135,34 @@ func recreateIfNeeded(api libmachine.API, cc *config.ClusterConfig, n *config.No
...
+			out.T(style.Restarting, `Restarting existing {{.driver_name}} {{.machine_type}} for "{{.cluster}}" ...`, out.V{"driver_name": cc.Driver, "cluster": machineName, "machine_type": machineType})
+		}
+		if err := h.Driver.Start(); err != nil {
+			MaybeDisplayAdvice(err, h.DriverName)
+			return h, errors.Wrap(err, "driver start")
+		}
+		if err := saveHost(api, h, cc, n); err != nil {
+			return h, err
+		}
+	} else {
```
priyawadhwa, line 160: Consider restructuring this to:

```
else if s == state.Running {
  // do something
}
else {
   return h, errors.Errorf("not running")
}
```
to minimize conditionals in conditionals

pkg/minikube/machine/stop.go
```diff
@@ -55,14 +55,16 @@ func stop(h *host.Host) error {
 		}
 	}
 
-	if err := h.Stop(); err != nil {
-		klog.Infof("stop err: %v", err)
-		st, ok := err.(mcnerror.ErrHostAlreadyInState)
-		if ok && st.State == state.Stopped {
-			klog.Infof("host is already stopped")
-			return nil
+	if h.DriverName != driver.Generic {
```
priyawadhwa, line 58: To make this more readable, consider:

```go
if h.DriverName == driver.Generic {
  return nil
}
```

priyawadhwa (commented):
This is looking really good! Is there any chance we could write an integration test? Maybe vagrant would work in one of our CI environments?

afbjorklund:
Thanks for the review, I will address some of the style issues shortly.

> Is there any chance we could write an integration test? Maybe vagrant would work in one of our CI environments?

As for testing, maybe we could "reuse" the same VM over and over ?

I also want to complete the fork of the driver, will get started on that...

cmd/minikube/cmd/start_flags.go
afbjorklund: I did that, but it was lost during the rebase. It should at least give an error if forgetting to provide an address ?

The other parameters are optional, by default it will connect as root on port 22 and assume keys are in ssh-agent

afbjorklund:
> I also want to complete the fork of the driver, will get started on that...

When I talk about the "fork", I mean the change that @medyagh did to the none driver...

Do the same thing here, change the API so that "stop" means stop kubelet + containers ?

Before:

```go
	return fmt.Errorf("hosts without a driver cannot be started")
}

	return fmt.Errorf("hosts without a driver cannot be stopped")
}

	return fmt.Errorf("hosts without a driver cannot be restarted")
}

func (d *Driver) Kill() error {
        return fmt.Errorf("hosts without a driver cannot be killed")
}
```

After:

```go
// Start a host
        var err error
        d.IPAddress, err = d.GetIP()
        if err != nil {
                return err
        }
        d.URL, err = d.GetURL()
        if err != nil {
                return err
        }
        return nil
}

// Stop a host gracefully, including any containers that we are managing.

	if err := sysinit.New(d.exec).Stop("kubelet"); err != nil {
		klog.Warningf("couldn't stop kubelet. will continue with stop anyways: %v", err)
		if err := sysinit.New(d.exec).ForceStop("kubelet"); err != nil {
			klog.Warningf("couldn't force stop kubelet. will continue with stop anyways: %v", err)
		}
	}
	containers, err := d.runtime.ListContainers(cruntime.ListOptions{})
	if err != nil {
		return errors.Wrap(err, "containers")
	}
	if len(containers) > 0 {
		if err := d.runtime.StopContainers(containers); err != nil {
			return errors.Wrap(err, "stop containers")
		}
	}
	klog.Infof("none driver is stopped!")
	return nil
}

// Restart a host
	return restartKubelet(d.exec)
}

// restartKubelet restarts the kubelet
func restartKubelet(cr command.Runner) error {
        klog.Infof("restarting kubelet.service ...")
        c := exec.Command("sudo", "systemctl", "restart", "kubelet.service")
        if _, err := cr.RunCmd(c); err != nil {
                return err
        }
        return nil
}

// Kill stops a host forcefully, including any containers that we are managing.
func (d *Driver) Kill() error {
        if err := sysinit.New(d.exec).ForceStop("kubelet"); err != nil {
                klog.Warningf("couldn't force stop kubelet. will continue with kill anyways: %v", err)
        }

        // First try to gracefully stop containers
        containers, err := d.runtime.ListContainers(cruntime.ListOptions{})
        if err != nil {
                return errors.Wrap(err, "containers")
        }
        if len(containers) == 0 {
                return nil
        }
        // Try to be graceful before sending SIGKILL everywhere.
        if err := d.runtime.StopContainers(containers); err != nil {
                return errors.Wrap(err, "stop")
        }

        containers, err = d.runtime.ListContainers(cruntime.ListOptions{})
        if err != nil {
                return errors.Wrap(err, "containers")
        }
        if len(containers) == 0 {
                return nil
        }
        if err := d.runtime.KillContainers(containers); err != nil {
                return errors.Wrap(err, "kill")
        }
        return nil
}
```

If we do the same API change here, we don't need those conditionals...

Basically we change it from managing the **host**, to managing **k8s**.
Which is a very weird thing to do for docker-machine, and thus: a fork

We could probably factor those common operations out to a separate file.

afbjorklund:
The before and after can be rather drastic:

Before:

```go
        _, err := drivers.RunSSHCommandFromDriver(d, "sudo shutdown -r now") // a.k.a. "reboot"
```

After:

```go
        _, err := drivers.RunSSHCommandFromDriver(d, "sudo systemctl restart kubelet.service")
```

But there is no `minikube restart` command...

medyagh (changes requested):
do you mind putting in the output an example usage of this driver?

afbjorklund:
I will give it some more testing locally, basically you will need a "raw" VM - I used Vagrant to set one up:

* https://github.com/kubernetes/minikube/issues/4733#issuecomment-510556159

But it could do with some updating, and it is rather spread out over several comments and issues too...

afbjorklund:
After the fork, we might also want to change the name of the driver...

afbjorklund:
> do you mind putting in the output an example usage of this driver?

Assuming that this works: `ssh root@$HOST`

(Normally that requires a `ssh-copy-id root@$HOST` or similar ssh key setup first)

Then you should be able to use this setup:

```
minikube start --driver generic --generic-ip-address $HOST
```


Where HOST would be the IP address of the VM that you are providing.

As usually, it needs to have 2 vCPU and 2 GB RAM (but not Docker etc)

You would also need to have ssh running, and the user needs sudo access.

Any VM that currently works with --driver none should also work remotely.

afbjorklund:
Forgot that the list of drivers is hard-coded now... (`supportedDrivers`)

```
✨  Using the generic driver based on user configuration

❌  Exiting due to DRV_UNSUPPORTED_OS: The driver 'generic' is not supported on linux

```

afbjorklund:
Added proper progress:

```
😄  minikube v1.14.1 on Ubuntu 20.04
✨  Using the generic driver based on existing profile
🔗  Running remotely (CPUs=2, Memory=1987MB, Disk=39643MB) ...
🐳  Preparing Kubernetes v1.19.2 on Docker 19.03.13 ...
🏄  Done! kubectl is now configured to use "minikube" by default
```

Compare with "none":

```
😄  minikube v1.14.1 on Ubuntu 20.04 (vbox/amd64)

🧯  The requested memory allocation of 1987MiB does not leave room for system overhead (total system memory: 1987MiB). You may face stability issues.
💡  Suggestion: Start minikube with less memory allocated: 'minikube start --memory=1987mb'

🤹  Running on localhost (CPUs=2, Memory=1987MB, Disk=39642MB) ...
ℹ️  OS release is Ubuntu 20.04 LTS
🐳  Preparing Kubernetes v1.19.2 on Docker 19.03.13 ...

🏄  Done! kubectl is now configured to use "minikube" by default
```

Main difference is that we need ssh to be up, so the progress and host info  comes later

afbjorklund:
Tested on:
* **Ubuntu 16.04**
* Ubuntu 18.04
* Ubuntu 20.04
* **CentOS 7**
* CentOS 8
* Fedora 32

Note that you _will_ need to install `conntrack`, and to turn off selinux and disable cgroups v2...

For CentOS 7, it also need `net-tools` and to revert the systemd disabling of netfilter for bridges.

```sh
    # setenforce 0
    sudo sed -i 's/^SELINUX=enforcing$/SELINUX=permissive/' /etc/selinux/config

    # cgroups v1: https://bugzilla.redhat.com/bugzilla/show_bug.cgi?id=1757078
    sudo grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"

    # revert /usr/lib/sysctl.d/00-system.conf
    cat <<EOF | sudo tee /etc/sysctl.d/k8s.conf
net.bridge.bridge-nf-call-ip6tables = 1
net.bridge.bridge-nf-call-iptables = 1
EOF
    sudo sysctl --system
```

afbjorklund:
I will do the driver fork in a separate PR, this one is complicated enough as it is.

pkg/util/utils.go
```diff
@@ -62,6 +63,63 @@ func ConvertUnsignedBytesToMB(byteSize uint64) int64 {
 	return int64(byteSize / units.MiB)
 }
 
+// ParseMemFree parses the output of the `free -m` command
```
tstromberg, line 66: Until we have multiple packages using these functions -- can you move it to `pkg/machine`? 

I'd rather this package not exist, to be honest.

afbjorklund: Will move that too, into a separate PR.

Trying to balance between minimal change so that it can ever get merged, and features so that it actually works...

afbjorklund:
I reverted the added feature to show proper progress and host info...

Needs rebasing before merging, but not needed for the review process

afbjorklund:
Should give it some weird helper function, so that these become "symmetrical":

```
 	if driver.IsVM(mc.Driver) || driver.IsKIC(mc.Driver) || mc.Driver == driver.Generic {
```

```
     if err == nil && (driver.BareMetal(cc.Driver) || cc.Driver == driver.Generic) {
```

Can start with IsGeneric for now

afbjorklund:
Not hearing anything, but will rebase and redo the testing tomorrow.

afbjorklund:
Will need yet another rebase, after christmas... Maybe have time to do the fork/rename as well, then: #9583 #7772

brandonros:
> libmachine will provision the docker daemon

Can you use `ssh` minikube driver without Docker daemon? Just a containr runtime like runc/cri-o/containerd?

afbjorklund:
Currently the provisoner in minikube is still broken, so you will have to provide the container runtime yourself.

But you can choose cri-o or containerd instead of docker, if you want. Not "runc" though, that is more low-level...

k8s runtimes: docker, cri-o, containerd

OCI runtimes: runc, crun

Issue #9229, opened 2020-09-13
Deprecate the old tcp port for docker-env


afbjorklund:
Currently we are using a standalone `tcp://` docker daemon, listening on port 2376.

```
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.99.100:2376"
export DOCKER_CERT_PATH="/home/anders/.minikube/certs"
```

We can use `ssh://` and connect directly to the unix socket instead, simplifying things.

This could use _either_ the current ssh shell tunnel, or we could use the regular address...

```
export DOCKER_HOST="ssh://docker@192.168.99.100:22"
ssh-add -k /home/anders/.minikube/machines/minikube/id_rsa
```

But we wouldn't have to manage all the extra ssl certificates for https, when using ssh.

```
/home/anders/.minikube/certs
├── ca-key.pem
├── ca.pem
├── cert.pem
└── key.pem

0 directories, 4 files
```

And this allows for having the docker-daemon socket-activated (on-demand) in the future...

Requirements: Docker 18.09 or later

Note that we already support both methods of connecting, so it can be a gradual change.

Current config: `/usr/bin/dockerd -H tcp://0.0.0.0:2376 -H unix:///var/run/docker.sock`

https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-socket-option

afbjorklund:
There is now an implementation in PR #9548, to allow you to try this out...

There are some quirks with this yet, to be sorted out by Docker upstream.

```
eval $(./out/minikube docker-env --ssh-host)
```

You will have to add the key, otherwise it will ask for the password _every time._

![ssh-password-authentication](https://user-images.githubusercontent.com/10364051/97104449-355ace80-16b4-11eb-8e0d-59f475d07a32.png)

```
ssh-add $(./out/minikube ssh-key)
```

You will have to add the host key too, since there is no setting to "ignore hosts".

![ssh-host-key-verification](https://user-images.githubusercontent.com/10364051/97104452-3c81dc80-16b4-11eb-95ef-252aadc6cd85.png)

_(currently done by choosing answer "yes")_

Podman has variables for this, but Docker doesn't have those features available.

----

Will look at saving the ssh host key on boot, instead of just disabling that ssh feature:

`-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null`

But it is a separate feature, docker has no support but delegates everything to ssh...

[...]

PR #9548, opened 2020-10-25
Optionally use ssh for docker-env instead of tcp

afbjorklund:
Since Docker 18.09, it supports ssh: as well as tcp:
for connecting to the docker daemon (in $DOCKER_HOST).

This avoids setting up certificates, but leaves key
handling (host keys and identity keys) to ssh config.

See #9229

Once this is default, we _only_ have to set up ssh...

And it will use `/var/run/docker.sock`, over ssh.


medyagh:
this is a cool PR, I look forward to see if this will fix some docker-env cert issues

medyagh:
/ok-to-test

do. youmind adding an integeration test for this , maybe under funcationl test

medyagh (changes requested):
we need to add docs on how to use this flag (specially docs on how to add SSH key)

afbjorklund:
> we need to add docs on how to use this flag (specially docs on how to add SSH key)

For some reason we always need to clean up Docker's mess (undocumented DOCKER_HOST)

afbjorklund:
We could export the value in a variable, but it seems a bit misleading if Docker is not using it ?
Also it is not mandatory to use ssh-agent, so we probably don't want to run `ssh-add` from env.

Currently we are using a special `ssh-key` command, to get the value to use for ssh identity.
Was planning to add a similar `ssh-host` (or something) command, to get the ssh host key(s).

afbjorklund:
Probably they will use the ssh tunnel rather than the main ip, but same difference...

Most likely we want to **only** show the `rsa` host key, to match our `id_rsa` identity key.

That way there is only one host key to worry about, the comment only goes on stderr:

`# 127.0.0.1:35717 SSH-2.0-OpenSSH_8.1`

afbjorklund:
One weird thing is that it doesn't actually talk _directly_ to the docker socket, it starts a `docker` program to do it:

`error during connect: Get http://docker/v1.40/version: command [ssh -l docker -p 35717 -- 127.0.0.1 docker system dial-stdio] has exited with exit status 255, please make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=Host key verification failed.`

```

Usage:	docker system dial-stdio

Proxy the stdio stream to the daemon connection. Should not be invoked manually.
```

So the client will need a `ssh` binary and the server will need a `docker` binary, behind the scenes

afbjorklund:
@medyagh now there is a `-ssh-add` option, to add the key to ssh-agent

medyagh:
we will do this PR after the release and also after the non-defaut docker PR is merged to ensure they work well together

Issue #9572, opened 2020-10-27
The Automatic PR after release for update.json should include docs


medyagh:
in this automatic PR 
https://github.com/kubernetes/minikube/pull/9571

we should include the docs 
https://github.com/kubernetes/minikube/edit/master/site/content/en/docs/_index.md
so this PR is automatic https://github.com/kubernetes/minikube/pull/9573/files

medyagh:
already dne

PR #9573, opened 2020-10-27
site: Update latest version

medyagh:
this PR should be auomated as part of the update json PR

https://github.com/kubernetes/minikube/issues/9572


priyawadhwa (approved):
If this has to be done manually every time, it should be added as a step in our Release process.

PR #9604, opened 2020-11-05
Bump kicbase on head

medyagh:
building changes on HEAD

also fixed the image build process, that was adding an extra -snapshot to the image name


Makefile
```diff
@@ -612,7 +611,7 @@ push-storage-provisioner-manifest: $(shell echo $(ALL_ARCH) | sed -e "s~[^ ]*~st
 	docker manifest push $(STORAGE_PROVISIONER_MANIFEST)
 
 .PHONY: push-docker
-push-docker: # Push docker image base on to IMAGE variable
+push-docker: # Push docker image base on to IMAGE variable (used internaly by other targets)
```
tstromberg, line 614: "internally"

tstromberg (changes requested):
Please split the Makefile & code change, In the unlikely case that this PR needs to reverted. Thanks!

sharifelgamal (approved):

PR #9510, opened 2020-10-21
Support non-default docker endpoints

ilya-zuyev:
Fixes https://github.com/kubernetes/minikube/issues/9463
This PR adds support of non-default docker configurations, defined by environment valirables DOCKER_HOST, DOCKER_CERT_PATH, and DOCKER_TLS_VERIFY.

Right now docker driver always uses the default values, and does not support docker damon runnig on external/VM host, or local docker daemon running on unix socker other that /var/run/docker.sock

Tested on Linux docker machine: 

```
ilyaz@skeletron --- z/minikube ‹master› » docker-machine create --driver virtualbox --virtualbox-cpu-count "2" --virtualbox-memory 4048  minik                                                                                         130 ↵
Running pre-create checks...
Creating machine...
(minik) Copying /home/ilyaz/.docker/machine/cache/boot2docker.iso to /home/ilyaz/.docker/machine/machines/minik/boot2docker.iso...
(minik) Creating VirtualBox VM...
(minik) Creating SSH key...
(minik) Starting the VM...
(minik) Check network to re-create if needed...
(minik) Waiting for an IP...
Waiting for machine to be running, this may take a few minutes...
Detecting operating system of created instance...
Waiting for SSH to be available...
Detecting the provisioner...
Provisioning with boot2docker...
Copying certs to the local machine directory...
Copying certs to the remote machine...
Setting Docker configuration on the remote daemon...
Checking connection to Docker...
Docker is up and running!
To see how to connect your Docker Client to the Docker Engine running on this virtual machine, run: docker-machine env minik


ilyaz@skeletron --- z/minikube ‹master› » docker-machine env minik                                                                                                                                                                       1 ↵
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.99.114:2376"
export DOCKER_CERT_PATH="/home/ilyaz/.docker/machine/machines/minik"
export DOCKER_MACHINE_NAME="minik"
# Run this command to configure your shell: 
# eval $(docker-machine env minik)

ilyaz@skeletron --- z/minikube ‹master› » eval $(docker-machine env minik)

ilyaz@skeletron --- z/minikube ‹master› » env | grep DOCKER
DOCKER_TLS_VERIFY=1
DOCKER_HOST=tcp://192.168.99.114:2376
DOCKER_CERT_PATH=/home/ilyaz/.docker/machine/machines/minik
DOCKER_MACHINE_NAME=minik
```

Before:
```
[...]

PR #9616, opened 2020-11-06
Pass environment to docker commands

4c0n:
Ran into some issues trying to use minikube on systems that do not use the unix socket to connect to the docker daemon.
The issue at hand is that the docker CLI is configured only through the use of arguments or environmental variables. Even when using contexts there is no other way to select a context.

This change allows the environmental variables to be passed along with the docker commands, so that the client can run in the same manner that it would as when it would be executed by the user (especially when configured using environmental variables).

To make this happen, I had to modify the behavior of always clearing the docker daemon related env vars within the application context to only unsetting them, when the log line actually states that it will.

### DIFF
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -524,16 +524,15 @@ func PointToHostDockerDaemon() error {
 	p := os.Getenv(constants.MinikubeActiveDockerdEnv)
 	if p != "" {
 		klog.Infof("shell is pointing to dockerd inside minikube. will unset to use host")
-	}
-
-	for i := range constants.DockerDaemonEnvs {
-		e := constants.DockerDaemonEnvs[i]
-		err := os.Setenv(e, "")
-		if err != nil {
-			return errors.Wrapf(err, "resetting %s env", e)
+		for i := range constants.DockerDaemonEnvs {
+			e := constants.DockerDaemonEnvs[i]
+			err := os.Setenv(e, "")
+			if err != nil {
+				return errors.Wrapf(err, "resetting %s env", e)
+			}
 		}
-
 	}
+
 	return nil
 }
 
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -526,7 +526,7 @@ func PointToHostDockerDaemon() error {
 		klog.Infof("shell is pointing to dockerd inside minikube. will unset to use host")
 		for i := range constants.DockerDaemonEnvs {
 			e := constants.DockerDaemonEnvs[i]
-			err := os.Unsetenv(e, "")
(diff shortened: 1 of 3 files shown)


4c0n:
I signed it

medyagh:
@4c0n thanks for this PR, have you looked athis PR https://github.com/kubernetes/minikube/pull/9510

this Pr might need to be added after that PR

4c0n:
@medyagh Thanks for pointing that PR out. I haven't looked at it before. It seems aimed to achieve a similar goal. Although for me this simple solution works. It also seems partially included in the other PR though. Not in a conflicting kind of way.

4c0n:
I signed it

pkg/minikube/cruntime/docker.go
```diff
@@ -137,6 +139,7 @@ func (r *Docker) Disable() error {
 func (r *Docker) ImageExists(name string, sha string) bool {
 	// expected output looks like [SHA_ALGO:SHA]
 	c := exec.Command("docker", "image", "inspect", "--format", "{{.Id}}", name)
+	c.Env = os.Environ()
```
tstromberg, line 142: I don't think this is necessary -- os.Environ() is always passed into exec.

4c0n: https://golang.org/pkg/os/exec/#Cmd
Looks like you're right, I overlooked this bit in the docs.

pkg/drivers/kic/oci/oci.go
```diff
@@ -524,16 +524,15 @@ func PointToHostDockerDaemon() error {
...
-	}
-
-	for i := range constants.DockerDaemonEnvs {
-		e := constants.DockerDaemonEnvs[i]
-		err := os.Setenv(e, "")
-		if err != nil {
-			return errors.Wrapf(err, "resetting %s env", e)
+		for i := range constants.DockerDaemonEnvs {
+			e := constants.DockerDaemonEnvs[i]
+			err := os.Setenv(e, "")
```
sharifelgamal, line 529: There's an os.UnsetEnv command you can use here.

4c0n: Thanks, I just placed this for loop inside the if block and didn't look to improve the logic itself. Using Unsetenv is better, so I modified it.

medyagh:
@4c0n we recenetly added support for non-default docker environments, do u mind checking if latest minikube on head has fixed the problem you are describing ?

medyagh:
does this PR do what you wanted ? https://github.com/kubernetes/minikube/pull/9510

tstromberg:
Do you mind resolving the merge conflicts and addressing @medyagh 's comments?

4c0n:
@medyagh Yes it looks like the change in this PR is also included/addressed in that PR. Closing this one.

Issue #9620, opened 2020-11-06
TestJSONOutput failing on Docker Linux


priyawadhwa:
[TestJSONOutput/start/parallel/IncreasingCurrentSteps](https://storage.googleapis.com/minikube-builds/logs/9604/df16364/Docker_Linux.html#fail_TestJSONOutput%2fstart%2fparallel%2fIncreasingCurrentSteps)
and [TestJSONOutput/start/parallel/DistinctCurrentSteps](https://storage.googleapis.com/minikube-builds/logs/9604/df16364/Docker_Linux.html#fail_TestJSONOutput%2fstart%2fparallel%2fIncreasingCurrentSteps) are failing at HEAD with

```
current step is not in increasing order
```

This test is likely failing because we don't have a separate step for pulling base image.

PR #9621, opened 2020-11-06
Fix failing TestJSONOutput/start test

priyawadhwa:
Should fix https://github.com/kubernetes/minikube/issues/9620

This test is likely failing because we don't have a separate step for pulling base image.

I also changed `out.T` to `out.Infof` for the suggestion to upgrade kubectl, which was failing locally for me and makes more sense since this is an informational message rather than a distinct step in the minikube process.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -464,7 +464,7 @@ func showKubectlInfo(kcs *kubeconfig.Settings, k8sVersion string, machineName st
 		out.Ln("")
 		out.WarningT("{{.path}} is version {{.client_version}}, which may have incompatibilites with Kubernetes {{.cluster_version}}.",
 			out.V{"path": path, "client_version": client, "cluster_version": cluster})
-		out.T(style.Tip, "Want kubectl {{.version}}? Try 'minikube kubectl -- get pods -A'", out.V{"version": k8sVersion})
+		out.Infof("Want kubectl {{.version}}? Try 'minikube kubectl -- get pods -A'", out.V{"version": k8sVersion})
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


medyagh (approved):

medyagh (changes requested):
build is failing

medyagh (approved):

cmd/minikube/cmd/start.go
```diff
@@ -464,7 +464,7 @@ func showKubectlInfo(kcs *kubeconfig.Settings, k8sVersion string, machineName st
 		out.Ln("")
 		out.WarningT("{{.path}} is version {{.client_version}}, which may have incompatibilites with Kubernetes {{.cluster_version}}.",
 			out.V{"path": path, "client_version": client, "cluster_version": cluster})
-		out.T(style.Tip, "Want kubectl {{.version}}? Try 'minikube kubectl -- get pods -A'", out.V{"version": k8sVersion})
```
tstromberg, line 467: This is intentional -- it's something users should absolutely see.

priyawadhwa: I think `out.Infof` is always shown?

tstromberg: Sorry, I missed that. So what's the difference between out.T and out.Infof then? Based on the naming of the functions, it isn't obvious why one would break this test.

priyawadhwa: so `out.T` could probably be better named to `out.Step` -- when I implemented JSON output I basically distinguished anything sent to `out.T` as a distinct progress step for `minikube start`, and anything sent to `out.Info` as an informational message that wouldn't be a distinct step in the process.

The test basically ensures that all things sent to `out.T` are distinctly numbered steps, so if someone adds a new step via `out.T` it ensures they will add that step to the JSON output registry & JSON output will still be correct.

priyawadhwa: I could keep this as `out.T` and add a `CheckKubectlVersion` step to the registry as an alternate fix, but this struck me as more of an informational message.

tstromberg: I think your idea of renaming `out.T` to `out.Step` would go a long way to making this less confusing for current and future readers. If we don't address this confusion, I'm sure this kind of issue will pop up again.

priyawadhwa: Cool, opened an issue to track that: https://github.com/kubernetes/minikube/issues/9664

PR #9634, opened 2020-11-08
Fix --extra-config when starting an existing cluster

### LINKED ISSUE #8242: --extra-config does not take effect on an existing cluster
For example, if you specify `-extra-config kubelet.kube-api-qps=5 --extra-config controller-manager.kube-api-qps=5`, and run `sudo ps -afe | grep controller`, you'll see that the new arguments do not appear.

dekkers:
Fixes #8242

Creating a minikube cluster with `minikube start` no arguments results in null value for ExtraOptions in `.minikube/profiles/minikube/config.json`:

```
        "ExtraOptions": null,
```

If we then stop and start minikube a second time with `minikube start --extra-config=apiserver.enable-admission-plugins=NamespaceAutoProvision` the ExtraOptions in the config file stays the same and in the cluster the apiserver config also doesn't change:

```
$ ps auxwf | grep -v grep | grep -o -- '--enable-admission[^ ]*'
--enable-admission-plugins=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,DefaultTolerationSeconds,NodeRestriction,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota
```

Doing the same thing with my patch the ExtraOptions in the profile will change to:

```
                "ExtraOptions": [
                        {
                                "Component": "apiserver",
                                "Key": "enable-admission-plugins",
                                "Value": "NamespaceAutoProvision"
                        }
                ],
```

And if we look in the cluster we see the config has been applied this time:

```
$ ps auxwf | grep -v grep | grep -o -- '--enable-admission[^ ]*'
--enable-admission-plugins=NamespaceAutoProvision
```

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -613,6 +613,10 @@ func updateExistingConfigFromFlags(cmd *cobra.Command, existing *config.ClusterC
 		cc.KubernetesConfig.ImageRepository = viper.GetString(imageRepository)
 	}
 
+	if cmd.Flags().Changed("extra-config") {
+		cc.KubernetesConfig.ExtraOptions = config.ExtraOptions
+	}
+
 	if cmd.Flags().Changed(enableDefaultCNI) && !cmd.Flags().Changed(cniFlag) {
 		if viper.GetBool(enableDefaultCNI) {
 			klog.Errorf("Found deprecated --enable-default-cni flag, setting --cni=bridge")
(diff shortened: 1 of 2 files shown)


medyagh (commented):
do you mind putting the PR description before and after this PR and check if the extra-config actually goes through on an existing cluster?

cmd/minikube/cmd/start_flags.go
```diff
@@ -611,6 +611,10 @@ func updateExistingConfigFromFlags(cmd *cobra.Command, existing *config.ClusterC
 		cc.KubernetesConfig.ImageRepository = viper.GetString(imageRepository)
 	}
 
+	if cmd.Flags().Changed("extra-config") {
+		cc.KubernetesConfig.ExtraOptions = config.ExtraOptions;
```
tstromberg, line 615: Thank you! Please remove the trailing semi-colon so that this passes the lint check.

An integration test would be nice too to prevent future breakage, but I'd be willing to leave it to a follow-up PR.

dekkers: I've fixed the lint error. I've added an integration test and verified that the test fails without the change.

priyawadhwa (approved):
LGTM, let's run the tests and make sure they pass

sharifelgamal:
Thanks for your contribution!

PR #9657, opened 2020-11-10
Document minikube kubectl autocompletion

### LINKED ISSUE #9603: Minikube Kubectl - completion does not work on zsh using alias
In the documentation it says
>You can also alias kubectl="minikube kubectl --" for easier usage.

when I use the current kubectl application from yum auto completion works:

```
➜  ~ kubectl
annotate        certificate     cordon          drain           get             port-forward    run             uncordon
api-versions    cluster-info    cp              edit            label           proxy           scale           version
apply           completion      create          exec            logs            replace         set
attach          config          delete          explain         options         rolling-update  taint
autoscale       convert         describe        expose          patch           rollout         top
```

When I use the "minikube kubectl -- " command it no longer works
```
➜  ~ minikube kubectl -- __minikube_handle_flag:31: bad math expression: operand expected at `'
➜  ~ minikube kubectl --
Desktop/              Downloads/            minikube-linux-amd64  Pictures/             Templates/
Documents/            install.sh*           Music/                Public/               Videos/

```

And I also get the same result when I now switch to an alias - I get an error and zsh instead completes with the files and folders from my current directory
```
➜  ~ alias kubectl="minikube kubectl --"
➜  ~ kubectl __minikube_handle_flag:31: bad math expression: operand expected at `'
     kubectl
Desktop/              Downloads/            minikube-linux-amd64  Pictures/             Templates/
Documents/            install.sh*           Music/                Public/               Videos/
```

I guess this is because zsh sees that "kubectl" is only an alias to "minikube ..." so it selects the minikube completion to handle this. From my understanding minikube could fix this in their completion, or there needs to be another way to do the alias so that zsh knows to call the kubectl autocompletion instead.

StefanLobbenmeier:
fixes #9603

### DIFF
--- a/site/content/en/docs/handbook/kubectl.md
+++ b/site/content/en/docs/handbook/kubectl.md
@@ -10,8 +10,7 @@ aliases:
 By default, [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) gets configured to access the kubernetes cluster control plane
 inside minikube when the `minikube start` command is executed. 
 
-However if `kubectl` is not installed locally, kubectl can be used inside the minikube
-as well.
+However if `kubectl` is not installed locally, minikube already includes kubectl which can be used like this:
 
 `minikube kubectl -- <kubectl commands>`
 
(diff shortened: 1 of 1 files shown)


StefanLobbenmeier:
> after that, please reply here with a new comment and we'll verify

StefanLobbenmeier:

Just gonna try again :)

StefanLobbenmeier:
Ok just got the confirmation from Hello Sign, should be good now

I signed it

medyagh (approved):
thanks

PR #9629, opened 2020-11-07
Upgrade cri-o in kicbase to version 1.18.4

afbjorklund:
Use the latest 1.18, since 1.18.4 is broken again \<sigh\>

Opened https://github.com/cri-o/cri-o/issues/4357 for it

Use the recommended cri-o-runc package instead of runc

From 1.0.0-rc10 to 1.0.0-rc92

PR #9659, opened 2020-11-10
Change back to the version-pinned cri-o channel

afbjorklund:
Left-over from PR #9629

Fixed in https://github.com/cri-o/cri-o/issues/4357

Both should install the same 1.18.4 package, though (at least for now - until 1.18.5 is out, that is)


medyagh (approved):

priyawadhwa:
Looks like the image did build successfully, even though KIC_IMAGE is failing. Not sure why it failed?

afbjorklund:
> Not sure why it failed?

The "build" job also runs an integration test

PR #9662, opened 2020-11-10
fix verifying kuberentes images test

medyagh:


test/integration/start_stop_delete_test.go
```diff
@@ -294,7 +294,6 @@ func testPulledImages(ctx context.Context, t *testing.T, profile string, version
 	for _, img := range jv["images"] {
 		for _, i := range img.Tags {
 			// Remove container-specific prefixes for naming consistency
-			i = strings.TrimPrefix(i, "docker.io/")
 			i = strings.TrimPrefix(i, "localhost/")
```
afbjorklund, line 297: probably remove this as well ? (it is what you get when loading local files with e.g. podman)

afbjorklund: actually I was wrong, we need to keep **both** of them. And then apply the docker.io transformation also to "want"

medyagh: done

sharifelgamal (approved):

PR #9672, opened 2020-11-11
Add Support for driver name alias

### LINKED ISSUE #8822: Add driver alias for kvm -> kvm2

sadlil:
This PR introduces support for driver name aliases. Currently it introduces only one alias for driver kvm2 to kvm. The changes also includes support for alternating between aliases and name when starting minikube. 

Fixes #8822 
Ref #7772 

Example
```
$ minikube start --driver=kvm
$ minikube stop
$ minikube start --driver=kvm2
$ minikube stop
$ minikube start --driver=kvm
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -660,6 +660,12 @@ func validateSpecifiedDriver(existing *config.ClusterConfig) {
 		return
 	}
 
+	// hostDriver always returns original driver name even if an alias is used to start minikube.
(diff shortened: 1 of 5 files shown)


medyagh (approved):
thank you for this PR looks good

PR #9674, opened 2020-11-11
Use special credential file location for cloud shell

sharifelgamal:

### DIFF
--- a/pkg/addons/gcpauth/enable.go
+++ b/pkg/addons/gcpauth/enable.go
@@ -19,6 +19,7 @@ package gcpauth
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"os"
 	"os/exec"
 	"strconv"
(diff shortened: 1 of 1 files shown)


pkg/addons/gcpauth/enable.go
```diff
@@ -63,6 +65,21 @@ func enableAddon(cfg *config.ClusterConfig) error {
 		exit.Message(reason.InternalCredsNotFound, "Could not find any GCP credentials. Either run `gcloud auth application-default login` or set the GOOGLE_APPLICATION_CREDENTIALS environment variable to the path of your credentials file.")
 	}
 
+	if creds.JSON == nil {
+		// If we're in a special environment, the creds file goes in a special place.
+		if e := os.Getenv("CLOUDSDK_CONFIG"); e != "" {
```
medyagh, line 70: lets have a func called, inGoogleCloudShell
(aws and others also have cloud shell)

and we can check for existance of both

CLOUDSDK_CONFIG env and the other Cloud shell specific env I forgot exact name (but can look up in `env` command)

medyagh (changes requested):

medyagh (approved):

PR #9676, opened 2020-11-11
Don't check for docker.io prefix in docker preload

priyawadhwa:
Right now, preload isn't working because it thinks that the
kubernetes/dashboard images don't exist. This is because they now have a
docker.io prefix introduced by 4c54e7868179471b9531e21b4d29472600860cbc

Since `docker images` won't return image names with that prefix, we want
to strip the images of `docker.io` before checking if they exists in the
docker daemon.

This also fixes the TestPause/serial/SecondStartNoReconfiguration
failing integration test.

fixes https://github.com/kubernetes/minikube/issues/9668


priyawadhwa:
Preload should be fixed by this, we can see a 10 second speedup in the PR bot ^ 

https://github.com/kubernetes/minikube/pull/9676#issuecomment-725709783

medyagh (approved):
thank you very much

PR #9689, opened 2020-11-12
Implement scheduled stop on windows

priyawadhwa:
Implements scheduled stop on windows.

This PR creates a minikube-scheduled-stop systemd service which is responsible for shutting down minikube from within. When a user runs:

```
minikube stop --schedule 2m
```

then minikube will:

1. Set an env variable SLEEP=120 via an environment file only the systemd service has access to
1. Starts the minikube-scheduled-stop systemd service

The systemd service reads the value of SLEEP and sleeps that amount of time. It then powers off minikube from within.

Cancelling old scheduled stops is as simple as running `sudo systemctl stop minikube-scheduled-stop` within minikube.

Tested the integration test on our windows instance with hyperv and docker and it passes.

Part of https://github.com/kubernetes/minikube/issues/9271


priyawadhwa:
TODO: build ISO/kicbase as a snapshot version and test with those

pkg/minikube/schedule/daemonize_windows.go
```diff
@@ -20,16 +20,93 @@ package schedule
...
+	if err != nil {
+		return errors.Wrap(err, "Error loading existing host. Please try running [minikube delete], then run [minikube start] again.")
+	}
+	runner, err := machine.CommandRunner(h)
+	if err != nil {
+		return errors.Wrap(err, "getting command runner")
+	}
+	// restart scheduled stop service in container
+	sysManger := sysinit.New(runner)
+	if err := sysManger.Stop(constants.ScheduledStopSystemdService); err != nil {
```
medyagh, line 59: what happens when error happens ? do we delete from config?

medyagh (approved):
approve with required testing upgrading k8s version

priyawadhwa:
@medyagh tested with k8s version upgrade, this still works as expected

pkg/minikube/schedule/daemonize_windows.go
priyawadhwa: updated to only save config if daemonize succeeds

PR #9693, opened 2020-11-13
update kubernetes version

medyagh:
I used the script 
```
cd hack/update/kubernetes_version
go run update_kubernetes_version.go
```

except it didnt generate the test data
```
pkg/minikube/bootstrapper/bsutil/kubeadm_test.go
pkg/minikube/bootstrapper/bsutil/testdata/v1.20.0-beta.1
```


prezha:
@medyagh yes, we'll find a way to add new test data with new k8s version
do you think that the folder should be called
>pkg/minikube/bootstrapper/bsutil/testdata/v1.20

ie, <major>.<minor>, instead of
>pkg/minikube/bootstrapper/bsutil/testdata/v1.20.0-beta.1

like all the others there, so that we don't need to change it with every patch and/or `-beta`/`-rc` release?

sharifelgamal (approved):

Issue #9707, opened 2020-11-15
fix kubeadm unit tests (testdata)


prezha:
ref: #9694
ref: #9693

currently, bsutil (kubeadm) unit tests have hardcoded list of versions to test, referencing `testdata` subdirectories for relevant k8s version to test, and it has to be manually updated with each new version released, eg:

https://github.com/kubernetes/minikube/blob/9b964853a9600eb4661f73a6aef7423c8da96f85/pkg/minikube/bootstrapper/bsutil/kubeadm_test.go#L80

and

https://github.com/kubernetes/minikube/blob/9b964853a9600eb4661f73a6aef7423c8da96f85/pkg/minikube/bootstrapper/bsutil/kubeadm_test.go#L111

we also need to cover the situation where we still don't have the `<major>.<minor>.0` stable version released (eg, as with 'v1.20.0-beta.1'), in which case we should use that latest k8s version until stable version gets released in test data files

therefore, it would be good to:
- have a dynamic (auto-discover) list of k8s versions to test (#9708 will automatically create new testdata set for newer versions), while also
- have only one set (subdirectory) with test data for each `<major>.<minor>` k8s version under `pkg/minikube/bootstrapper/bsutil/testdata/`

medyagh:
this is a good idea

PR #9713, opened 2020-11-15
fix bsutil (kubeadm) unit tests (testdata)

prezha:
fixes: #9707

this pr replaces the hard-coded list of k8s versions to test and builds it dynamically from the existing testdata versions subdirectories, and also handles the case when `<major>.<minor>.0` stable version is not yet released (eg, as with 'v1.20.0-beta.1') by using the latest k8s pre-release version available until stable version gets out, so to have only one set (subdirectory) with test data for each `<major>.<minor>` k8s version under `pkg/minikube/bootstrapper/bsutil/testdata/`

### DIFF
--- a/pkg/minikube/bootstrapper/bsutil/files.go
+++ b/pkg/minikube/bootstrapper/bsutil/files.go
@@ -18,8 +18,11 @@ limitations under the License.
 package bsutil
 
 import (
+	"fmt"
+	"io/ioutil"
 	"os/exec"
 	"path"
+	"sort"
 
 	"github.com/pkg/errors"
 	"k8s.io/minikube/pkg/minikube/assets"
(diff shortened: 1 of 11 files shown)


pkg/minikube/bootstrapper/bsutil/files.go
```diff
@@ -59,3 +62,23 @@ func CopyFiles(runner command.Runner, files []assets.CopyableFile) error {
 	}
 	return nil
 }
+
+// ReverseDirList returns a list of subdirectories under the given path, sorted in reverse order, and any error.
+// If n > 0, ReverseDirList returns at most n subdirectories.
+// If n <= 0, ReverseDirList returns all the subdirectories from the directory.
+func ReverseDirList(path string, n int) (list []string, err error) {
```
medyagh, line 69: seems like this function is only used in the test ? then let it put in the test file
also if func not used outside the package, it must be private 
also comment why we need this function ? what it is used for ?

medyagh (changes requested):

pkg/minikube/bootstrapper/bsutil/files.go
prezha: @medyagh all good points, thank you: i've merged ReverseDirList() into recentReleases() and then reused it where needed

medyagh (approved):

Issue #9704, opened 2020-11-14
Translation error in every minikube command since 1.15.0


galishmann:
When I am trying to execute any command in minikube I have error message:
```sh
~ $ minikube version
I1114 13:59:47.843637   11252 translate.go:89] Failed to load translation file for en: Asset translations/en.json not found
minikube version: v1.15.0
commit: 3e098ff146b8502f597849dfda420a2fa4fa43f0
```

afbjorklund:
Works here, which platform are you on ?

afbjorklund:
Hmm, can reproduce with non-american locale:

```
$ LC_ALL=en_US.UTF-8 ./out/minikube version
minikube version: v1.15.0
commit: 3e098ff146b8502f597849dfda420a2fa4fa43f0
$ LC_ALL=en_UK.UTF-8 ./out/minikube version
I1114 14:50:45.479064  444106 translate.go:89] Failed to load translation file for en: Asset translations/en.json not found
minikube version: v1.15.0
commit: 3e098ff146b8502f597849dfda420a2fa4fa43f0
```
Or for other languages:
```
$ LC_ALL=sv_SE.UTF-8 ./out/minikube version
I1114 14:52:27.157482  444194 translate.go:89] Failed to load translation file for sv: Asset translations/sv.json not found
minikube version: v1.15.0
commit: 3e098ff146b8502f597849dfda420a2fa4fa43f
```

Worked OK in v1.14.2.

afbjorklund:
No changes to translate itself, though. So it looks like a `klog` change.

fnichol:
I can confirm this behavior with the 1.15.0 release (running macOS 10.15.7).

Explicitly setting an `en_US` locale:

```console
$ env LC_ALL=en_US.UTF-8 minikube version
minikube version: v1.15.0
commit: 3e098ff146b8502f597849dfda420a2fa4fa43f0
```

Setting a non-US locale--mine in this case:

```console
$ env LC_ALL=en_CA.UTF-8 minikube version
I1114 13:07:25.983227   66526 translate.go:89] Failed to load translation file for en: Asset translations/en.json not found
minikube version: v1.15.0
commit: 3e098ff146b8502f597849dfda420a2fa4fa43f0
```

No locale environment variable set, presumably falling back to a variant of `en_CA` in my case:

```console
$ env -u LC_ALL minikube version
[...]

PR #9715, opened 2020-11-16
less verbose language selector

avoidik:
An attempt to resolve https://github.com/kubernetes/minikube/issues/9704


avoidik:
I'd assume the only reason of the "exec format error" message is incorrect target architecture

sharifelgamal:
Thanks for fixing this!

PR #9716, opened 2020-11-16
Add private network implementation for podman

### LINKED ISSUE #9714: Add dedicate network and static IP for podman driver
in minikube 1.14.0 we had docker driver with dedicate network, we should do same thing for podman for consistancy

- Similar to this PR 
https://github.com/kubernetes/minikube/pull/9294

- This will make Podman driver to also have a Static IP similar to our Docker Driver after that PR
- You would need to install Podman on Linux to test this

afbjorklund:
Most of it the same as docker, except for the options.

i.e. libnetwork "bridge" plugin vs. cni "bridge" plugin

Does not remove the network when deleting

Closes #9714

### DIFF
--- a/pkg/drivers/kic/oci/network_create.go
+++ b/pkg/drivers/kic/oci/network_create.go
@@ -37,20 +37,17 @@ const firstSubnetAddr = "192.168.49.0"
 // big enough for a cluster of 254 nodes
 const defaultSubnetMask = 24
 
-// name of the default Docker bridge network, used to lookup the MTU (see #9528)
-const dockerDefaultBridge = "bridge"
+// name of the default bridge network, used to lookup the MTU (see #9528)
+const defaultBridgeName = "bridge"
 
 // CreateNetwork creates a network returns gateway and error, minikube creates one network per cluster
 func CreateNetwork(ociBin string, name string) (net.IP, error) {
-	if ociBin != Docker {
-		return nil, fmt.Errorf("%s network not implemented yet", ociBin)
-	}
-	return createDockerNetwork(name)
+	return createDockerNetwork(ociBin, name)
 }
 
-func createDockerNetwork(clusterName string) (net.IP, error) {
+func createDockerNetwork(ociBin string, clusterName string) (net.IP, error) {
 	// check if the network already exists
-	info, err := dockerNetworkInspect(clusterName)
+	info, err := containerNetworkInspect(ociBin, clusterName)
 	if err == nil {
 		klog.Infof("Found existing network %+v", info)
 		return info.gateway, nil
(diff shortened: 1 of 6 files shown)


pkg/drivers/kic/oci/network_create.go
```diff
@@ -100,20 +97,25 @@ func tryCreateDockerNetwork(subnetAddr string, subnetMask int, mtu int, name str
...
+	if ociBin == Docker {
+		// options documentation https://docs.docker.com/engine/reference/commandline/network_create/#bridge-driver-options
+		args = append(args, "-o")
+		args = append(args, "--ip-masq")
 		args = append(args, "-o")
-		args = append(args, fmt.Sprintf("com.docker.network.driver.mtu=%d", mtu))
+		args = append(args, "--icc")
+
+		// adding MTU option because #9528
+		if mtu > 0 {
```
medyagh, line 109: could podman network benefit from specifying the MTU ? (this fixed the issue of minikube in cloud shell not being able to pull large images like `golang:1.15` due to different cloud shell MTUs

pkg/drivers/kic/oci/network_create.go
```diff
@@ -136,12 +138,16 @@ type netInfo struct {
 }
 
 // if exists returns subnet, gateway and mtu
-func dockerNetworkInspect(name string) (netInfo, error) {
+func dockerNetworkInspect(ociBin string, name string) (netInfo, error) {
```
medyagh, line 141: does this need to be renamed to networkInspect ? if it is working for both podman and docker ?

medyagh (changes requested):

pkg/drivers/kic/oci/network_create.go
afbjorklund: There is "docker" all over the place, I didn't rename any functions.

pkg/drivers/kic/oci/network_create.go
afbjorklund: It might, but it will not benefit from "com.docker.network.driver.mtu" at any rate.

afbjorklund:
Currently it is not possible to configure MTU:

```go
// HostLocalBridge describes a configuration for a bridge plugin
// https://github.com/containernetworking/plugins/tree/master/plugins/main/bridge#network-configuration-reference
type HostLocalBridge struct {
        PluginType   string            `json:"type"`
        BrName       string            `json:"bridge,omitempty"`
        IsGW         bool              `json:"isGateway"`
        IsDefaultGW  bool              `json:"isDefaultGateway,omitempty"`
        ForceAddress bool              `json:"forceAddress,omitempty"`
        IPMasq       bool              `json:"ipMasq,omitempty"`
        MTU          int               `json:"mtu,omitempty"`
        HairpinMode  bool              `json:"hairpinMode,omitempty"`
        PromiscMode  bool              `json:"promiscMode,omitempty"`
        Vlan         int               `json:"vlan,omitempty"`
        IPAM         IPAMHostLocalConf `json:"ipam"`
}
```

```go
// NewHostLocalBridge creates a new LocalBridge for host-local
func NewHostLocalBridge(name string, isGateWay, isDefaultGW, ipMasq bool, ipamConf IPAMHostLocalConf) *HostLocalBridge {
        hostLocalBridge := HostLocalBridge{
                PluginType:  "bridge",
                BrName:      name,
                IPMasq:      ipMasq,
                HairpinMode: true,
                IPAM:        ipamConf,
        }
        if isGateWay {
                hostLocalBridge.IsGW = true
        }
        if isDefaultGW {
                hostLocalBridge.IsDefaultGW = true
        }
        return &hostLocalBridge
}
```

So the feature needs to be added to podman first.

pkg/drivers/kic/oci/network_create.go
medyagh: I mean if we are using it for both docker and podman now (unlike before) we should rename it to networkInspect

afbjorklund:
Another request for MTU: https://github.com/containers/podman/issues/8454

See https://github.com/kubernetes/minikube/pull/9716#issuecomment-730161331

medyagh (approved):

afbjorklund:
Support for MTU will be available in Podman 3.0, can revisit it again then...

https://github.com/containers/podman/pull/8457

pkg/drivers/kic/oci/network_create.go
afbjorklund: It will be available in the future (like podman 3.0 or so), with a similar option

medyagh: maybe add a comment, to remember TODO add mtu for podman 3 (with an issue liunk)

PR #9734, opened 2020-11-18
Update ISO and kicbase to include scheduled stop 

priyawadhwa:


medyagh (changes requested):
please check KVM tests, the ISO is not available

medyagh (approved):

PR #9746, opened 2020-11-19
Updating link to main_test

Helithumper:
The link was set to `main.go` which no longer exists, thus the link was dead. Updated the link to now point to `main_test.go` which is alive.


medyagh (approved):
thanks for fixing this

Issue #9744, opened 2020-11-19
none driver: TestOffline/group/docker "panic: reflect: indirection through nil pointer to embedded struct "


medyagh:
```
	W1119 20:04:46.520507    9260 out.go:146] X Exiting due to K8S_INSTALL_FAILED: updating control plane: generating kubeadm cfg: getting cgroup driver: docker info --format : exit status 2
	stdout:
	
	stderr:
	panic: reflect: indirection through nil pointer to embedded struct [recovered]
		panic: reflect: indirection through nil pointer to embedded struct
```

as seen here 
https://storage.googleapis.com/minikube-builds/logs/9742/c237d5d/none_Linux.html#fail_TestOffline%2fgroup%2fdocker

this ran on Debian Agent 4 (kvm kic-docker docker) https://storage.googleapis.com/minikube-builds/logs/9742/none_Linux.txt

medyagh:
I sshed to Debian Agent 4 and I ran the command myself

```
jenkins@debian-jenkins-agent-4:/home/medya$ docker info --format {{.CgroupDriver}}
cgroupfs

```

and it worked

ilya-zuyev:
```
stderr:
	panic: reflect: indirection through nil pointer to embedded struct [recovered]
		panic: reflect: indirection through nil pointer to embedded struct
	
	goroutine 1 [running]:
	text/template.errRecover(0xc000563320)
		/usr/local/go/src/text/template/exec.go:170 +0x1b5
	panic(0x55fc322c0de0, 0x55fc32617480)
		/usr/local/go/src/runtime/panic.go:679 +0x1b6
	reflect.Value.FieldByIndex(0x55fc324a5c40, 0xc00004e400, 0x99, 0xc000422050, 0x2, 0x2, 0x0, 0x55fc326d4220, 0x55fc322c0de0)
		/usr/local/go/src/reflect/value.go:883 +0x39e
	text/template.(*state).evalField(0xc0005632a0, 0x55fc324a5c40, 0xc00004e400, 0x99, 0x7ffd4bd1e514, 0xc, 0x55fc326a31e0, 0xc0005d0180, 0xc0005d4070, 0x1, ...)
		/usr/local/go/src/text/template/exec.go:605 +0x475
[...]

PR #9747, opened 2020-11-19
Fix invalid DOCKER_HOST handling in integration tests

ilya-zuyev:
Fixes #9774

### DIFF
--- a/test/integration/aab_offline_test.go
+++ b/test/integration/aab_offline_test.go
@@ -60,38 +60,3 @@ func TestOffline(t *testing.T) {
 		}
 	})
 }
-
-func TestOfflineDownload(t *testing.T) {
-	t.Run("group", func(t *testing.T) {
-		for _, runtime := range []string{"docker", "crio", "containerd"} {
-			runtime := runtime
-			t.Run(runtime, func(t *testing.T) {
-				MaybeParallel(t)
-
-				if runtime != "docker" && NoneDriver() {
-					t.Skipf("skipping %s - incompatible with none driver", t.Name())
-				}
-
-				profile := UniqueProfileName(fmt.Sprintf("offline-%s", runtime))
-				ctx, cancel := context.WithTimeout(context.Background(), Minutes(15))
-				defer CleanupWithLogs(t, profile, cancel)
-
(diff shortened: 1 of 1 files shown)


test/integration/aab_offline_test.go
```diff
@@ -49,6 +49,40 @@ func TestOffline(t *testing.T) {
...
+				if err != nil {
+					// Fatal so that we may collect logs before stop/delete steps
+					t.Fatalf("%s failed: %v", rr.Command(), err)
+				}
+			})
+		}
+	})
+}
+
+func TestOfflineDownload(t *testing.T) {
```
medyagh, line 64: we might not need to do this , since we run download only before all tests

test/integration/aab_offline_test.go
```diff
@@ -49,6 +49,40 @@ func TestOffline(t *testing.T) {
 				// RFC1918 address that unlikely to host working a proxy server
 				env = append(env, "HTTP_PROXY=172.16.1.1:1")
 				env = append(env, "HTTP_PROXYS=172.16.1.1:1")
+
```
medyagh, line 52: the goal of this test should be, we simulate an enviroment that minikube is disconnected from internet,
so `minikube ssh docker pull a_new_image` should not work but everything else should work.

could u please verify, that setting these env variables, will Prevent minikube from downloading anything new ?
```
HTTP_PROXY=172.16.1.1:1"
HTTP_PROXYS=172.16.1.1:1
```
if we set those envs and do 
"minikube ssh"
and then try to "docker pull a_new_image"
that should fail, (to ensure there is no internet)
but kubernetes itself should not fail, since minikube should load the kubernets images from cache

if we add that verification to this test we can be esnure we dont break ppl who use minikube on an aireplane.

medyagh (changes requested):

test/integration/aab_offline_test.go
ilya-zuyev: removed

medyagh:
I belive this fixes https://github.com/kubernetes/minikube/issues/9744

medyagh (approved):

PR #9748, opened 2020-11-19
Remove template dependency in addons

alonyb:
This PR is to:

1. Fix logviewer addon, same bug as #9733
2. An improvement in the solution, with this, there will no longer be a dependency on the boolean variable `isTemplate`, in this case all templates will always be rendered. With this there will no longer be problems that folks have or don't have to use the boolean variable.

WIP:

1. Change all yaml files into tmpl format.
2. Test that all addons work

logviewer output err:
image

Now all addons are in tmpl format, yaml files have been removed. 
The output of the addons that were changed now still works:

Screen Shot 2020-11-21 at 18 16 22


medyagh (approved):

medyagh (approved):
thank you @alonyb for fixing this bug

PR #9752, opened 2020-11-20
Abbreviate too long filenames to show progress

### LINKED ISSUE #9740: Show download progress for the Preload Tarball
image

As you can see in the above screenshot, we don't show the progress of the download. It is missing the bandwidth, size of remaining content, ETA etc. like we show for the ISO.

Since this preload tarball is big, the above information is essential.

afbjorklund:
Closes #9740

### DIFF
--- a/pkg/minikube/download/progressbar.go
+++ b/pkg/minikube/download/progressbar.go
@@ -47,7 +47,8 @@ func (cpb *progressBar) TrackProgress(src string, currentSize, totalSize int64, 
 		cpb.progress = pb.New64(totalSize)
 	}
 	p := pb.Full.Start64(totalSize)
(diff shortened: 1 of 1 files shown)


afbjorklund:
This might need a less stupid way to abbreviate the long filenames.

preloaded-images-k8s-v6-v1.19.4-docker-overlay2-amd64.tar.lz4

```
    > preloaded-images-k8s-v6-v1....: 22.89 MiB / 486.35 MiB  4.71% 15.20 MiB p^C
```

We lost the "1.19.4" and "docker" information, along with .tar.lz4

afbjorklund:
Showing it on a separate line before the progress bar, might be one way to handle it

medyagh (commented):
this is a good idea

PR #9755, opened 2020-11-20
integration test: fail service test early if can't deploy example svc

medyagh:
also adding memory limit to tests that dont have memory limit

### DIFF
--- a/test/integration/status_test.go
+++ b/test/integration/status_test.go
@@ -40,7 +40,7 @@ func TestInsufficientStorage(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), Minutes(5))
 	defer Cleanup(t, profile, cancel)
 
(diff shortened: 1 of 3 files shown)


test/integration/status_test.go
```diff
@@ -40,7 +40,7 @@ func TestInsufficientStorage(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), Minutes(5))
 	defer Cleanup(t, profile, cancel)
 
-	startArgs := []string{"start", "-p", profile, "--output=json", "--wait=true"}
+	startArgs := []string{"start", "-p", profile, "--memory=1900", "--output=json", "--wait=true"}
```
sharifelgamal, line 43: why the memory allocation here?

medyagh: so it doesn't auto select 6GB for a test that doesnt need it

tstromberg (approved):

Issue #9691, opened 2020-11-13
Flake Fail: TestFunctional/parallel/DockerEnv	


medyagh:
this is a flake that API Server goes stopped after it started succesfully

TestFunctional/parallel/DockerEnv
Example: https://storage.googleapis.com/minikube-builds/logs/9747/3b5c265/Docker_Linux.html#fail_TestFunctional%2fparallel%2fDockerEnv
```

=== RUN   TestFunctional/parallel/DockerEnv
=== PAUSE TestFunctional/parallel/DockerEnv
=== CONT  TestFunctional/parallel/DockerEnv
functional_test.go:175: (dbg) Run:  /bin/bash -c "eval $(out/minikube-linux-amd64 -p functional-20201113002616-6290 docker-env) && out/minikube-linux-amd64 status -p functional-20201113002616-6290"
=== CONT  TestFunctional/parallel/DockerEnv
functional_test.go:175: (dbg) Non-zero exit: /bin/bash -c "eval $(out/minikube-linux-amd64 -p functional-20201113002616-6290 docker-env) && out/minikube-linux-amd64 status -p functional-20201113002616-6290": exit status 2 (10.381188135s)
-- stdout --
	functional-20201113002616-6290
	type: Control Plane
	host: Running
	kubelet: Running
	apiserver: Stopped
	kubeconfig: Configured
	
-- /stdout --
functional_test.go:181: failed to do status after eval-ing docker-env. error: exit status 2
helpers_test.go:216: -----------------------post-mortem--------------------------------
helpers_test.go:224: ======>  post-mortem[TestFunctional/parallel/DockerEnv]: docker inspect <======
helpers_test.go:225: (dbg) Run:  docker inspect functional-20201113002616-6290
helpers_test.go:229: (dbg) docker inspect functional-20201113002616-6290:
-- stdout --
	[
	    {
	        "Id": "d4134ddbdd8c8ea995b08ffe372cf0b92fe2ac529b5f5796e4895717941bea11",
	        "Created": "2020-11-13T00:26:17.836545463Z",
	        "Path": "/usr/local/bin/entrypoint",
	        "Args": [
	            "/sbin/init"
	        ],
	        "State": {
	            "Status": "running",
	            "Running": true,
	            "Paused": false,
	            "Restarting": false,
	            "OOMKilled": false,
	            "Dead": false,
	            "Pid": 38252,
	            "ExitCode": 0,
	            "Error": "",
[...]

PR #9756, opened 2020-11-20
docker-env: "reload" dockerd service first before "restart" to avoid restarting apiserver container

medyagh:
will attempt to fix this 
https://github.com/kubernetes/minikube/issues/9691

### DIFF
--- a/pkg/minikube/sysinit/systemd.go
+++ b/pkg/minikube/sysinit/systemd.go
@@ -34,8 +34,8 @@ func (s *Systemd) Name() string {
 	return "systemd"
 }
 
-// reload reloads systemd configuration
-func (s *Systemd) reload() error {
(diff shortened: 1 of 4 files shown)


sharifelgamal (approved):

pkg/minikube/sysinit/systemd.go
```diff
@@ -63,7 +63,7 @@ func (s *Systemd) Enable(svc string) error {
 
 // Start starts a service
 func (s *Systemd) Start(svc string) error {
-	if err := s.reload(); err != nil {
+	if err := s.daemonReload(); err != nil {
```
nirs, line 66: @medyagh do you remember why this was added? daemon-reload is not needed for starting, reloading, or restarting a service.

PR #9761, opened 2020-11-21
Enable Ingress Addon for Docker Windows

blueelvis:
Related bug - #7332 but just for Docker on Windows.

Before merging this PR, ensure that this PR is merged - https://github.com/kubernetes/minikube/pull/9753

With this PR,
1. Enable the Ingress addon for minikube for Docker on Windows.
2. Add a notice in the output which tells to run `minikube tunnel` to access the ingress resources.

Output -
```
PS C:\utilities> .\minikube-windows-amd64.exe addons enable ingress                                                                                           I1121 00:59:39.443965    3000 translate.go:89] Failed to load translation file for en: Asset translations/en.json not found
* After the addon is enabled, please run "minikube tunnel" and your ingress resources would be available at "127.0.0.1"
* Verifying ingress addon...
* The 'ingress' addon is enabled
```

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -145,14 +145,18 @@ func enableOrDisableAddon(cc *config.ClusterConfig, name string, val string) err
 
 	// to match both ingress and ingress-dns addons
 	if strings.HasPrefix(name, "ingress") && enable {
-		if driver.IsKIC(cc.Driver) && runtime.GOOS != "linux" {
-			exit.Message(reason.Usage, `Due to networking limitations of driver {{.driver_name}} on {{.os_name}}, {{.addon_name}} addon is not supported.
+		if driver.IsKIC(cc.Driver) {
+			if runtime.GOOS == "windows" {
+				out.Step(style.Tip,`After the addon is enabled, please run "minikube tunnel" and your ingress resources would be available at "127.0.0.1"`)
+			} else if runtime.GOOS != "linux" {
+				exit.Message(reason.Usage, `Due to networking limitations of driver {{.driver_name}} on {{.os_name}}, {{.addon_name}} addon is not supported.
 Alternatively to use this addon you can use a vm-based driver:
 
 	'minikube start --vm=true'
 
 To track the update on this work in progress feature please check:
 https://github.com/kubernetes/minikube/issues/7332`, out.V{"driver_name": cc.Driver, "os_name": runtime.GOOS, "addon_name": name})
+			}
 		} else if driver.BareMetal(cc.Driver) {
 			exit.Message(reason.Usage, `Due to networking limitations of driver {{.driver_name}}, {{.addon_name}} addon is not supported. Try using a different driver.`,
 				out.V{"driver_name": cc.Driver, "addon_name": name})
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -147,7 +147,7 @@ func enableOrDisableAddon(cc *config.ClusterConfig, name string, val string) err
 	if strings.HasPrefix(name, "ingress") && enable {
 		if driver.IsKIC(cc.Driver) {
 			if runtime.GOOS == "windows" {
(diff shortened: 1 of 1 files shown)


medyagh (changes requested):
please check lint

bl-ue (approved):

pkg/addons/addons.go
```diff
@@ -145,14 +145,18 @@ func enableOrDisableAddon(cc *config.ClusterConfig, name string, val string) err
 
 	// to match both ingress and ingress-dns addons
 	if strings.HasPrefix(name, "ingress") && enable {
-		if driver.IsKIC(cc.Driver) && runtime.GOOS != "linux" {
-			exit.Message(reason.Usage, `Due to networking limitations of driver {{.driver_name}} on {{.os_name}}, {{.addon_name}} addon is not supported.
+		if driver.IsKIC(cc.Driver) {
+			if runtime.GOOS == "windows" {
```
tstromberg, line 149: Would the same approach would work for macOS as well? It seems like it should.

blueelvis: I don't have access to a macOS machine. Could you please try the following and see if the addon is working?
1. Pull the latest from the master branch and build for macos.
2. `kubectl apply -f kubectl apply -f https://projectcontour.io/quickstart/contour.yaml`
3. `kubectl apply -f https://projectcontour.io/examples/kuard.yaml`
4. `minikube tunnel`
5. Try accessing kuard. If you are able to access kuard using Contour Ingress then we can enable the ingress addon for macos, otherwise not.

bl-ue: @blueelvis I just tried it&mdash;this is the output (repeated every 5s):
```
Status:	
	machine: minikube
	pid: 15490
	route: 10.96.0.0/12 -> 172.16.63.33
	minikube: Running
	services: [envoy]
    errors: 
		minikube: no errors
		router: no errors
		loadbalancer emulator: no errors
```

and this is http://172.16.63.33:

Screen Shot 2020-12-16 at 8 05 45 AM

bl-ue: But that was from `master`.  Did you mean to checkout your branch `blueelvis:task/enable-ingress-docker-windows`?

blueelvis: @bl-ue - Did you test this on minikube with Docker driver on MacOS? Or did minikube create a dedicated VM for you? And yes, I meant taking the latest master branch.

bl-ue: @blueelvis Yes, it did create a vm because I started minikube with `--vm=true`.  Would you like me to test using Docker, i.e. without `--vm=true`?

bl-ue: Oops! I wasn't even using my freshly built `minikube` 😄   Let me retest with Docker and minikube `master` v1.16.0-beta.0.

bl-ue: So this time, `minikube tunnel` doesn't produce any output after I enter my password:
```
$ minikube tunnel
❗  The service envoy requires privileged ports to be exposed: [80 443]
🔑  sudo permission will be asked for it.
🏃  Starting tunnel for service envoy.
Password:
```
- http://127.0.0.1:80 presents me with the same screen as in the screenshot above (with different IDs of course).
- https://127.0.0.1:443 shows `This site can’t be reached ... The connection was reset.`

tstromberg: FWIW, it did not work with me on macOS with the Docker driver. `kubectl get services` never showed an IP, and localhost:443/80, while listened on, didn't go to the service in question (`curl: (56) Recv failure: Connection reset by peer`).

I think we're close, but I guess not quite there yet.

tstromberg (approved):

pkg/addons/addons.go
blueelvis: Thanks both of you for testing this!

bl-ue: You're so welcome @blueelvis!!!!

PR #9764, opened 2020-11-23
Remove duplicated if statement in `(p *Profile) IsValid()`.

daehyeok:
Below code repeated inside `(p *Profile) IsValid()` function.
 ```Go
if p.Config == nil {
    return false
}
```
P.S Second if statement written by #6318 (47d918f).


medyagh (approved):
nice catch, thanks @daehyeok

Issue #9730, opened 2020-11-17
Add --cancel-scheduled-stop flag to cancel all existing scheduled stops


priyawadhwa:
Related to https://github.com/kubernetes/minikube/issues/9271

PR #9774, opened 2020-11-24
Add --cancel-scheduled flag to cancel all existing scheduled stops

priyawadhwa:
output looks like:

```
$ minikube stop --cancel-scheduled
🛑  All existing scheduled stops cancelled
```

adds to integration test to make sure this flag works as expected.

fixes https://github.com/kubernetes/minikube/issues/9730

### DIFF
--- a/site/content/en/docs/commands/stop.md
+++ b/site/content/en/docs/commands/stop.md
@@ -20,10 +20,10 @@ minikube stop [flags]
 ### Options
 
 ```
-      --all                     Set flag to stop all profiles (clusters)
-      --cancel-scheduled-stop   cancel any existing scheduled stop requests
-      --keep-context-active     keep the kube-context active after cluster is stopped. Defaults to false.
-  -o, --output string           Format to print stdout in. Options include: [text,json] (default "text")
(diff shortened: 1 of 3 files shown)


site/content/en/docs/commands/stop.md
```diff
@@ -20,9 +20,10 @@ minikube stop [flags]
 ### Options
 
 ```
-      --all                   Set flag to stop all profiles (clusters)
-      --keep-context-active   keep the kube-context active after cluster is stopped. Defaults to false.
-  -o, --output string         Format to print stdout in. Options include: [text,json] (default "text")
+      --all                     Set flag to stop all profiles (clusters)
+      --cancel-scheduled-stop   cancel any existing scheduled stop requests
```
medyagh, line 24: this seems like a very long name, and since this is already a stop command maybe it is redundent to say stop two times?

medyagh (changes requested):

site/content/en/docs/commands/stop.md
priyawadhwa: shortened to cancel-scheduled

medyagh (approved):

PR #9783, opened 2020-11-25
Update conformance tests for multinode testing

sharifelgamal:

### DIFF
--- a/hack/conformance_tests.sh
+++ b/hack/conformance_tests.sh
@@ -27,19 +27,21 @@ set -ex -o pipefail
 readonly PROFILE_NAME="k8sconformance"
 readonly MINIKUBE=${1:-./out/minikube}
 shift || true
+readonly START_ARGS=${@:-}
 
 # Requires a fully running Kubernetes cluster.
 "${MINIKUBE}" delete -p "${PROFILE_NAME}" || true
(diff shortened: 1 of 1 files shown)


hack/conformance_tests.sh
```diff
@@ -27,19 +27,21 @@ set -ex -o pipefail
 readonly PROFILE_NAME="k8sconformance"
 readonly MINIKUBE=${1:-./out/minikube}
 shift || true
+readonly START_ARGS=${@:-}
 
 # Requires a fully running Kubernetes cluster.
 "${MINIKUBE}" delete -p "${PROFILE_NAME}" || true
-"${MINIKUBE}" start -p "${PROFILE_NAME}" --wait=all
+"${MINIKUBE}" start -p "${PROFILE_NAME}" ${START_ARGS} --wait=all --nodes=2
```
medyagh, line 34: should it have --container-runtime=containerd ?

medyagh (approved):

hack/conformance_tests.sh
sharifelgamal: I'm keeping the script as generic as possible, so we might eventually be able to test with the docker runtime

medyagh (approved):

PR #9786, opened 2020-11-26
bump kindnet cni to latest

medyagh:


priyawadhwa (commented):
This should come with a bump in preload version, correct?

PR #9788, opened 2020-11-26
WIP: CI: check temporary directory for unittests

zhijianli88:
Signed-off-by: Li Zhijian <lizhijian@cn.fujitsu.com>

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -76,7 +76,9 @@ jobs:
       - name: Unit Test
         env:
           TESTSUITE: unittest
-        run: make test
+        run: |
+          TMPDIR=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


sharifelgamal:
@zhijianli88 Could you update your PR description to describe what problem you're trying to solve and why this is the best way to solve it? Thanks.

zhijianli88:
> @zhijianli88 Could you update your PR description to describe what problem you're trying to solve and why this is the best way to solve it? Thanks.

My First thought was to make sure all unit tests should not left over any temporary directories/files. However, this checking will is blocked by https://github.com/juju/mutex/issues/8 where a 3rd packages imported by minikube will always lefts over a temporary files.

let's close it first.

PR #9793, opened 2020-11-27
feat(minikube) display scheduledstop status in minikube status

### LINKED ISSUE #9731: `minikube status` should show existing scheduled stops and time left
Related to https://github.com/kubernetes/minikube/issues/9271

tharun208:
Signed-off-by: Tharun <rajendrantharun@live.com>

Updated minikube status command to show `scheduledStop` time

Fixes #9731

### DIFF
--- a/cmd/minikube/cmd/status.go
+++ b/cmd/minikube/cmd/status.go
@@ -143,6 +143,7 @@ type ClusterState struct {
 	BaseState
 
 	BinaryVersion string
+	TimeToStop    string
 	Components    map[string]BaseState
 	Nodes         []NodeState
 }
@@ -485,10 +486,6 @@ func clusterState(sts []*Status) ClusterState {
 	}
 	sc := statusCode(statusName)
 
-	timeToStopStatusCode := Configured
-	if sts[0].TimeToStop == Nonexistent {
-		timeToStopStatusCode = Nonexistent
-	}
 	cs := ClusterState{
 		BinaryVersion: version.GetVersion(),
 
@@ -499,9 +496,10 @@ func clusterState(sts []*Status) ClusterState {
 			StatusDetail: codeDetails[sc],
 		},
 
+		TimeToStop: sts[0].TimeToStop,
+
 		Components: map[string]BaseState{
 			"kubeconfig": {Name: "kubeconfig", StatusCode: statusCode(sts[0].Kubeconfig)},
-			"timetostop": {Name: "timetostop", StatusCode: statusCode(timeToStopStatusCode)},
 		},
 	}
 
--- a/cmd/minikube/cmd/status.go
+++ b/cmd/minikube/cmd/status.go
@@ -135,6 +135,7 @@ type Status struct {
 	APIServer  string
 	Kubeconfig string
 	Worker     bool
+	TimeToStop string
 }
 
 // ClusterState holds a cluster state representation
@@ -170,12 +171,6 @@ type BaseState struct {
 	StepDetail string `json:",omitempty"`
 }
 
-// ScheduledStopStatus holds string representation of scheduledStopDuration
-type ScheduledStopStatus struct {
-	InitiatedTime string `json:",omitempty"`
-	ScheduledTime string `json:",omitempty"`
-}
-
 const (
 	minikubeNotRunningStatusFlag = 1 << 0
 	clusterNotRunningStatusFlag  = 1 << 1
@@ -186,18 +181,13 @@ host: {{.Host}}
 kubelet: {{.Kubelet}}
 apiserver: {{.APIServer}}
 kubeconfig: {{.Kubeconfig}}
+timeToStop: {{.TimeToStop}}
 
 `
 	workerStatusFormat = `{{.Name}}
 type: Worker
 host: {{.Host}}
 kubelet: {{.Kubelet}}
-
-`
-
-	scheduledStopStatusFormat = `type: ScheduledDuration
-initiatedTime: {{.InitiatedTime}}
-scheduledTime: {{.ScheduledTime}}
 
 `
 )
@@ -268,11 +258,6 @@ func writeStatusesAtInterval(duration time.Duration, api libmachine.API, cc *con
 					exit.Error(reason.InternalStatusText, "status text failure", err)
 				}
 			}
-			if cc.ScheduledStop != nil {
-				if err := scheduledStopStatusText(cc.ScheduledStop, os.Stdout); err != nil {
-					exit.Error(reason.InternalStatusText, "status text failure", err)
-				}
-			}
 		case "json":
 			// Layout is currently only supported for JSON mode
 			if layout == "cluster" {
@@ -282,11 +267,6 @@ func writeStatusesAtInterval(duration time.Duration, api libmachine.API, cc *con
 			} else {
 				if err := statusJSON(statuses, os.Stdout); err != nil {
 					exit.Error(reason.InternalStatusJSON, "status json failure", err)
-				}
-			}
-			if cc.ScheduledStop != nil {
-				if err := scheduledStopStatusJSON(cc.ScheduledStop, os.Stdout); err != nil {
-					exit.Error(reason.InternalStatusJSON, "scheduledstop status json failure", err)
 				}
 			}
 		default:
@@ -329,6 +309,7 @@ func nodeStatus(api libmachine.API, cc config.ClusterConfig, n config.Node) (*St
 		Kubelet:    Nonexistent,
 		Kubeconfig: Nonexistent,
 		Worker:     !controlPlane,
+		TimeToStop: Nonexistent,
 	}
 
 	hs, err := machine.Status(api, name)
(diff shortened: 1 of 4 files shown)


medyagh (changes requested):
do u mind adding before/after this PR status output ?

in both cases when it is scheduled and not scheduled

cmd/minikube/cmd/status.go
```diff
@@ -170,6 +170,12 @@ type BaseState struct {
 	StepDetail string `json:",omitempty"`
 }
 
+// ScheduledStopStatus holds string representation of scheduledStopDuration
+type ScheduledStopStatus struct {
+	InitiatedTime string `json:",omitempty"`
```
priyawadhwa, line 175: So I think it would be more useful to provide the user with the time left until shutdown, instead of initiation time and scheduled time. It will save the user some calculation.

We can call this:

```go
type ScheduledStopStatus struct {
   TimeToStop string
}
```

cmd/minikube/cmd/status.go
```diff
@@ -170,6 +170,12 @@ type BaseState struct {
 	StepDetail string `json:",omitempty"`
 }
 
+// ScheduledStopStatus holds string representation of scheduledStopDuration
+type ScheduledStopStatus struct {
```
priyawadhwa, line 174: This struct should be part of the top level `Status` struct (Line 131)


```go

type Status struct {
ScheduledStop ScheduledStopStatus
}
```

cmd/minikube/cmd/status.go
```diff
@@ -187,6 +193,12 @@ type: Worker
 host: {{.Host}}
 kubelet: {{.Kubelet}}
 
+`
+
+	scheduledStopStatusFormat = `type: ScheduledDuration
+initiatedTime: {{.InitiatedTime}}
```
priyawadhwa, line 199: Similarly, let's only include TimeToStop here

priyawadhwa (commented):
Thank you for taking this issue on! Left a few comments, please let me know if you have any questions.

cmd/minikube/cmd/status.go
tharun208: @priyawadhwa I was thinking since the `ScheduledStopStatus` struct has the only field, we can keep that in `status` as `TimeToStop string`? Do we need to also need to show `type: ScheduledDuration` in status ?

priyawadhwa: That's a good question -- let's keep it as a struct in case we ever want to add more information to it.

I think we can start with just `TimeToStop` and we can think about `ScheduledDuration` later if we decide we need it.

cmd/minikube/cmd/status.go
```diff
@@ -367,6 +370,13 @@ func nodeStatus(api libmachine.API, cc config.ClusterConfig, n config.Node) (*St
 	stk := kverify.ServiceStatus(cr, "kubelet")
 	st.Kubelet = stk.String()
 
+	if cc.ScheduledStop != nil {
+		timeToStop := time.Now().Add(cc.ScheduledStop.Duration)
```
priyawadhwa, line 374: When a user runs:
```
minikube stop --schedule 2m
```

then `cc.ScheduledStop.InitiationTime` and `cc.ScheduledStop.Duration` is stored in cc.ScheduledStop

We want to compare the duration to that initiation time, not `time.Now()`, because `minikube status` will have been run after. 

I would rewrite this as:

```go
if cc.ScheduledStop != nil {
  st.TimeToStop = time.Until(cc.ScheduledStop.Initiation.Add(cc.ScheduledStop.Duration))
} else {
  st.TimeToStop = Irrelevant
}
```

priyawadhwa (commented):
Thanks for the update, left a suggestion,

cmd/minikube/cmd/status.go
tharun208: fixed it

tharun208:
@priyawadhwa It is working as expected. I will now add a check-in `int` tests for this and also change the format of the string representation

cmd/minikube/cmd/status.go
```diff
@@ -366,7 +369,12 @@ func nodeStatus(api libmachine.API, cc config.ClusterConfig, n config.Node) (*St
 
 	stk := kverify.ServiceStatus(cr, "kubelet")
 	st.Kubelet = stk.String()
-
+	if cc.ScheduledStop != nil {
+		initiationTime := time.Unix(cc.ScheduledStop.InitiationTime, 0)
+		st.TimeToStop = time.Until(initiationTime.Add(cc.ScheduledStop.Duration)).String()
+	} else {
```
priyawadhwa, line 375: We can actually remove this else statement since the default for `TimeToStop` is set as `Nonexistent`

priyawadhwa (commented):
An integration test would be great!

Could you also provide the output of:

```
minikube start
minikube stop --schedule 2m
minikube status
minikube status --output json --layout cluster
```

priyawadhwa:
You can actually add a check for `minikube status` in this integration test:

https://github.com/kubernetes/minikube/blob/8fc54b430ba616dbadb129c62c20053d2e1d9390/test/integration/scheduled_stop_test.go#L73

which basically just makes sure that `TimeToStop` is set after this block of code runs:

https://github.com/kubernetes/minikube/blob/8fc54b430ba616dbadb129c62c20053d2e1d9390/test/integration/scheduled_stop_test.go#L86-L90

tharun208:
> minikube status --output json --layout cluster

Screen Shot 2020-12-04 at 1 01 57 PM

cmd/minikube/cmd/status.go
```diff
@@ -490,6 +501,7 @@ func clusterState(sts []*Status) ClusterState {
 
 		Components: map[string]BaseState{
 			"kubeconfig": {Name: "kubeconfig", StatusCode: statusCode(sts[0].Kubeconfig)},
+			"timetostop": {Name: "timetostop", StatusCode: statusCode(timeToStopStatusCode)},
```
priyawadhwa, line 504: Instead of a component, I think it would make more sense to have this higher level in `ClusterState`. That way we don't need to worry about status codes either.

Right now, it looks like this:

```json
{
  "Name": "minikube",
  "StatusCode": 200,
  "StatusName": "OK",
  "BinaryVersion": "v1.15.1",
  "Components": {
    "kubeconfig": {
      "Name": "kubeconfig",
      "StatusCode": 200,
      "StatusName": ""
    },
    "timetostop": {
      "Name": "timetostop",
      "StatusCode": 200,
      "StatusName": ""
    }
  },
  "Nodes": [
    {
      "Name": "minikube",
      "StatusCode": 200,
      "StatusName": "OK",
      "Components": {
        "apiserver": {
          "Name": "apiserver",
          "StatusCode": 200,
          "StatusName": "OK"
        },
        "kubelet": {
          "Name": "kubelet",
          "StatusCode": 200,
          "StatusName": "OK"
        }
      }
    }
  ]
}
```

and the following would be simpler:

```json
{
  "Name": "minikube",
  "StatusCode": 200,
  "StatusName": "OK",
  "BinaryVersion": "v1.15.1",
  "TimeToStop": "23s",
  "Components": {
    "kubeconfig": {
      "Name": "kubeconfig",
      "StatusCode": 200,
      "StatusName": ""
    }
  },
  "Nodes": [
    {
      "Name": "minikube",
      "StatusCode": 200,
      "StatusName": "OK",
      "Components": {
        "apiserver": {
          "Name": "apiserver",
          "StatusCode": 200,
          "StatusName": "OK"
        },
        "kubelet": {
          "Name": "kubelet",
          "StatusCode": 200,
          "StatusName": "OK"
        }
      }
    }
  ]
}
```

test/integration/scheduled_stop_test.go
```diff
@@ -154,6 +154,10 @@ func ensureMinikubeStatusStopped(ctx context.Context, t *testing.T, profile stri
 		if got != state.Stopped.String() {
 			return fmt.Errorf("expected post-stop host status to be -%q- but got *%q*", state.Stopped, got)
 		}
+		got = Status(ctx, t, Target(), profile, "TimeToStop", profile)
+		if got != "Nonexistent" {
```
priyawadhwa, line 158: Instead of making sure that `TimeToStop` is nonexistant after stop, let's make sure it's set to something after we schedule the stop but before it happens.

That could happen here:

https://github.com/kubernetes/minikube/blob/master/test/integration/scheduled_stop_test.go#L54

priyawadhwa (commented):
Thank you for your changes! I tested locally and it's really close, just a couple more suggestions and we should be ready to merge :)

cmd/minikube/cmd/status.go
tharun208: done

test/integration/scheduled_stop_test.go
```diff
@@ -84,6 +87,8 @@ func TestScheduledStopUnix(t *testing.T) {
 
 	// schedule a stop for 5 min from now and make sure PID is created
 	stopMinikube(ctx, t, profile, []string{"--schedule", "5m"})
+	// make sure timeToStop is present in status
+	ensureMinikubeStatus(ctx, t, profile, "TimeToStop", "3m")
```
priyawadhwa, line 91: I see that you had to change to using  `strings.Contains` in the `ensureMinikubeStatus` function, since we can't know what the exact value of `TimeToStop` will be. Instead of doing a contains, which could potentially fail, I would just check that TimeToStop is a valid duration < 5m.

Then we can also revert back to making sure `got == wantStatus` in the `ensureMinikubeStatus` function

priyawadhwa (commented):
Thank you! This is looking really good, just one more comment.

test/integration/scheduled_stop_test.go
tharun208: So, can we pull this check to a seperate function or modify the comparison based on the `key` value to the `ensureMinikubeStatus` itself ?.

priyawadhwa: Let's do a separate function

priyawadhwa (approved):
Looks good, thank you so much for contributing this feature!

PR #9801, opened 2020-11-29
Add support for restoring existing podman env

afbjorklund:
If user had a podman environment configured before podman-env,
restore it on unset (similar to existing code for docker-env)

Only do this for podman v2, not for the deprecated podman v1...
And yes, it is called CONTAINER_HOST and not PODMAN_HOST


medyagh (approved):

medyagh (changes requested):
PR needs lint please

medyagh (approved):

PR #9808, opened 2020-11-30
Add script to collect metrics for minikube start

priyawadhwa:
This script will run continuously to collect `minikube start` time across multiple operating systems to make sure we have no performance regressions.


hack/metrics/metrics.go
```diff
@@ -0,0 +1,140 @@
...
+
+func minikubeStartTime(ctx context.Context, projectID string) (float64, error) {
+	minikubePath, err := downloadMinikube()
+	if err != nil {
+		return 0, errors.Wrap(err, "downloading minikube")
+	}
+	defer os.Remove(minikubePath)
+	defer deleteMinikube(ctx, minikubePath)
+
+	cmd := exec.CommandContext(ctx, minikubePath, "start", "--driver=docker", "-p", profile, "--memory=2000", "--trace=gcp")
```
medyagh, line 120: is there a reason to do 2000mb ram ? why not more realistic one, on most systems we auto-select way more than 2GB

medyagh (changes requested):

hack/metrics/metrics.go
priyawadhwa: I want to measure the minimum required memory so that we can compare across platforms (linux/cloud shell/windows) with a constant variable.

I also don't want minikube to auto-select a high memory, which improves start time, and falsely lets us believe that minikube is fast when maybe it is slow with our minimum requirement of 2GB.

hack/metrics/minikube.go
```diff
@@ -0,0 +1,77 @@
...
+	"cloud.google.com/go/storage"
+	"github.com/pkg/errors"
+)
+
+const (
+	bucketName = "minikube/latest"
+)
+
+// download minikube latest to a tmp file
+func downloadMinikube() (string, error) {
```
priyawadhwa, line 37: Update to only download if the binary has changed

hack/metrics/metrics.go
medyagh: sounds good, I am ok with this, however if the number is too slow, we could try a more realistic number like 4GB (as we choose in our functional test)

medyagh (approved):

PR #9797, opened 2020-11-28
Ability to use a custom TLS certificate with the Ingress

msedzins:
Description of functionality can be found here: #9335

**How to use it**
$ minikube addons configure ingress
-- Enter custom cert(format is "namespace/secret"): kube-system/mkcert
✅  ingress was successfully configured

Enable addons plugin (or disable and then enable if already enabled):
$ minikube addons disable ingress
$ minikube addons enable ingress

**Expected result**
After running:
$ kubectl  -n kube-system get deployment ingress-nginx-controller -o yaml

There will be one extra parameter added to nginx (the last one):
spec:
      containers:
      - args:
        - /nginx-ingress-controller
        - --configmap=$(POD_NAMESPACE)/nginx-load-balancer-conf
        - --report-node-internal-ip-address
        - --tcp-services-configmap=$(POD_NAMESPACE)/tcp-services
        - --udp-services-configmap=$(POD_NAMESPACE)/udp-services
        - --validating-webhook=:8443
        - --validating-webhook-certificate=/usr/local/certificates/cert
        - --validating-webhook-key=/usr/local/certificates/key
        **- --default-ssl-certificate=kube-system/mkcert**

PR #9813, opened 2020-11-30
Ability to use a custom TLS certificate with the Ingress addon

msedzins:
Ability to use a custom TLS certificate with the Ingress: #9797


msedzins:
not needed

PR #9819, opened 2020-12-01
Debian installer: only recommend virtualbox when running on amd64

afbjorklund:
Remove the "recommendation" from the arm64 deb


Makefile
```diff
@@ -462,6 +462,11 @@ out/minikube_$(DEB_VERSION)-0_%.deb: out/minikube-linux-%
 	chmod 0755 out/minikube_$(DEB_VERSION)/DEBIAN
 	sed -E -i 's/--VERSION--/'$(DEB_VERSION)'/g' out/minikube_$(DEB_VERSION)/DEBIAN/control
 	sed -E -i 's/--ARCH--/'$*'/g' out/minikube_$(DEB_VERSION)/DEBIAN/control
+	if [ "$*" = "amd64" ]; then \
+	    sed -E -i 's/--RECOMMENDS--/virtualbox/' out/minikube_$(DEB_VERSION)/DEBIAN/control; \
```
medyagh, line 466: should we consider adding recommend docker ? since it is our most tested driver?

afbjorklund: Quite different from removing (not available) virtualbox, but I suppose it could be done

medyagh (approved):

PR #9821, opened 2020-12-01
CI: add test order to gopogh report

medyagh:
- Add sort by test order (and default to order by test order rather than duration)
-  Adds alternating background color to test name
Screen Shot 2020-12-01 at 11 44 48 AM


medyagh:
/hold 
github action that releases binaries are down, will need to wait till gopogh binaries are published

priyawadhwa (approved):

ilya-zuyev (approved):
lgtm

PR #9823, opened 2020-12-01
Add a button to copy the command to the browser on the docs site

### LINKED ISSUE #9712: Add a button to copy the command to the browser on the docs site
I think that it would be good user experience if we can add a button besides each code snippet which should be copied to the clipboard like several sites do it.

edtrist:
Fixes #9712.

Adds a copy button to code blocks using Prism through Docsy (https://www.docsy.dev/docs/adding-content/lookandfeel/#code-highlighting-with-prism). The default is that the button only appears when the user hovers over the code block.

This PR:
- Updates Docsy to the latest version to include the Prism feature ([diff](https://github.com/google/docsy/compare/dd303fd19cc13ffc01bcbe86ff54f21e423d04de...2536303cad19991c673037f4f16332075141364a))
- Enables Prism highlighting on code blocks
- Tweaks some styles to counteract changes in the newer version of Docsy and preserve existing look
- Brings in a custom version of the Cobra doc generation code that adds the `shell` language annotation to runnable commands, adding the copy button (this also removes the need for the `removeHelpText` function, since this text is now not generated at all)
- Regenerated and manually edited Markdown files to add missing language annotations

Before

After


medyagh (approved):
this is a great feature added to our website. thank you very much @edtrist

PR #9857, opened 2020-12-05
Create reusable SSH session in factory

spowelljr:
This change opens the gate to allow log streaming on SSH commands. It will hopefully also result in a speed increase as new sessions won't have to be created on every command.


medyagh:
the integration test failures seem abnoramly high, have u tried running functioanl test locally ?

spowelljr:
@medyagh I looked into the source code of ssh and it's not possible to call `Start` more than once in a single session, so a long running session that's reused multiple times is not possible. I'll close this PR and try a different strategy.

PR #9866, opened 2020-12-05
Fix podman network inspect format and error

afbjorklund:
Somehow this ended up with the docker network inspect format,
instead of the CNI based one used by podman network inspect ?

Also add a check for when trying to name the cluster "podman",
which runs into trouble (same as docker driver with name "bridge")

### DIFF
--- a/pkg/drivers/kic/oci/network_create.go
+++ b/pkg/drivers/kic/oci/network_create.go
@@ -38,14 +38,21 @@ const firstSubnetAddr = "192.168.49.0"
 const defaultSubnetMask = 24
 
 // name of the default bridge network, used to lookup the MTU (see #9528)
-const defaultBridgeName = "bridge"
+const dockerDefaultBridge = "bridge"
+
+// name of the default bridge network
+const podmanDefaultBridge = "podman"
 
 // CreateNetwork creates a network returns gateway and error, minikube creates one network per cluster
-func CreateNetwork(ociBin string, name string) (net.IP, error) {
-	return createDockerNetwork(ociBin, name)
-}
+func CreateNetwork(ociBin string, clusterName string) (net.IP, error) {
+	var defaultBridgeName string
+	if ociBin == Docker {
+		defaultBridgeName = dockerDefaultBridge
+	}
+	if ociBin == Podman {
+		defaultBridgeName = podmanDefaultBridge
+	}
 
-func createDockerNetwork(ociBin string, clusterName string) (net.IP, error) {
 	// check if the network already exists
 	info, err := containerNetworkInspect(ociBin, clusterName)
 	if err == nil {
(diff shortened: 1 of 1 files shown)


afbjorklund:
Left-overs from #9716, which actually caused the whole thing to come crashing down...

```
✨  Using the podman (experimental) driver based on user configuration
👍  Starting control plane node podman in cluster podman
panic: runtime error: index out of range [1] with length 1
```

pkg/drivers/kic/oci/network_create.go
```diff
@@ -38,14 +38,30 @@ const firstSubnetAddr = "192.168.49.0"
...
+
+// name of the default bridge network
+const podmanDefaultBridge = "podman"
 
 // CreateNetwork creates a network returns gateway and error, minikube creates one network per cluster
 func CreateNetwork(ociBin string, name string) (net.IP, error) {
 	return createDockerNetwork(ociBin, name)
 }
 
 func createDockerNetwork(ociBin string, clusterName string) (net.IP, error) {
```
medyagh, line 51: the name here seems odd, createDockerNetwork creates both docker and podman, maybe move the implementaiton to the public func ?

medyagh (changes requested):

pkg/drivers/kic/oci/network_create.go
```diff
@@ -38,14 +38,30 @@ const firstSubnetAddr = "192.168.49.0"
...
+	if ociBin == Docker {
+		defaultBridgeName = dockerDefaultBridge
+	}
+	if ociBin == Podman {
+		defaultBridgeName = podmanDefaultBridge
+	}
+
+	// validate the name of the new network
+	if clusterName == defaultBridgeName {
+		return nil, errors.Errorf("Can't use same name as %s's default network: %s", ociBin, clusterName)
```
medyagh, line 62: I wonder if we should not Error, but out.Warnning
that you specified the DefaultNetwork, we will Not create a new one and just use that.
I think this will come handy when a paranoid user like me wants to ensure the network is bridge and not create a minikube network. 

some users need to Use the default bridge maybe for weird MTU issues we seen in the cloud shell

medyagh (changes requested):

pkg/drivers/kic/oci/network_create.go
medyagh: I recommend checking before creating if the name is Bridge then just return the net.IP of that bridge.

medyagh (changes requested):

pkg/drivers/kic/oci/network_create.go
afbjorklund: Just wondering what person names their cluster/profile to "bridge"

pkg/drivers/kic/oci/network_create.go
afbjorklund: Yeah, no effort has been made to clean up the function names

tstromberg (approved):

Issue #9838, opened 2020-12-03
Problematic Multi Node Networking with docker driver and kindnetd CNI


sadlil:
Creating a multi node minikube cluster with docker driver and kindnetd CNI seems to be resulting in broken netwroking inside the pods running in the worker nodes. This networking problem is not presents with calico CNI plugin.

```bash
$ minikube version
minikube version: v1.15.1
commit: 23f40a012abb52eff365ff99a709501a61ac5876
```

```bash
$ system_profiler SPSoftwareDataType
Software:
      System Version: macOS 10.15.7 (19H15)
      Kernel Version: Darwin 19.6.0
```

```bash
$ docker version
Client: Docker Engine - Community
 Cloud integration: 1.0.2
 Version:           19.03.13
 API version:       1.40
 Go version:        go1.13.15
 Git commit:        4484c46d9d
 Built:             Wed Sep 16 16:58:31 2020
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.13
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       4484c46d9d
  Built:            Wed Sep 16 17:07:04 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.3.7
  GitCommit:        8fba4e9a7d01810a393d5d25a3621dc101981175
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

**Steps to reproduce the issue:** 

1. minikube start -p multi-node -n 3
2. Create nginx deployment and svc, and another network test deployment.
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-nginx
spec:
  selector:
    matchLabels:
      run: my-nginx
  replicas: 4
  template:
    metadata:
      labels:
        run: my-nginx
    spec:
      containers:
      - name: my-nginx
        image: nginx
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: my-nginx
  labels:
    run: my-nginx
spec:
  ports:
  - port: 80
    protocol: TCP
  selector:
    run: my-nginx
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: net-test
spec:
  selector:
    matchLabels:
[...]

Issue #7538, opened 2020-04-09
Multi-node GA Reqs


sharifelgamal:
This is the meta issue tracking everything that needs to happen to get our multi-node feature more stable and no longer experimental.

- [ ] Multi-node should survive restarts (#7496)
- [ ] Integration tests (#7462) 
- [ ] Automated CNI at run time (#7459)
- [ ] Multi control plane clusters (#7461) 
- [ ] `minikube node start/add` should respect relevant `minikube start` parameters (#7366)
- [ ] `minikube profile` commands need to not be broken (#6780) 
- [ ] docker-env (#6506) 

Once this issue is closed, we should be good to close #94 

Please add any other relevant issues that I might have missed.

priyawadhwa:
still in progress :)

danmutblix:
An absolutely feat! ❤

PR #9875, opened 2020-12-07
Fix multi node two pods getting same IP and nodespec not having PodCIDR

sadlil:
Fixes https://github.com/kubernetes/minikube/issues/9838
Ref https://github.com/kubernetes/minikube/issues/7538

With this fixes enabled In multi node mode nodes gets the podCIDR set
```
$ kubectl get nodes -o custom-columns=NAME:.metadata.name,SPEC:.spec
NAME           SPEC
minikube       map[podCIDR:10.244.0.0/24 podCIDRs:[10.244.0.0/24]]
minikube-m02   map[podCIDR:10.244.1.0/24 podCIDRs:[10.244.1.0/24]]
minikube-m03   map[podCIDR:10.244.2.0/24 podCIDRs:[10.244.2.0/24]]
```

Pods running in a node gets IP from the node CIDR as intended. And No reuse of same IP.
```
$ kubectl get pods --all-namespaces -o wide
NAMESPACE     NAME                               READY   STATUS    RESTARTS   AGE     IP             NODE           NOMINATED NODE   READINESS GATES
default       my-nginx-5b56ccd65f-dghzv          1/1     Running   0          43s     10.244.1.2     minikube-m02   <none>           <none>
default       my-nginx-5b56ccd65f-l5f2k          1/1     Running   0          43s     10.244.2.2     minikube-m03   <none>           <none>
default       my-nginx-5b56ccd65f-tlwx5          1/1     Running   0          43s     10.244.0.3     minikube       <none>           <none>
default       my-nginx-5b56ccd65f-xfhmp          1/1     Running   0          43s     10.244.1.4     minikube-m02   <none>           <none>
default       net-test-c4f9cfdd4-4w95m           1/1     Running   0          43s     10.244.1.3     minikube-m02   <none>           <none>
default       net-test-c4f9cfdd4-9bsks           1/1     Running   0          43s     10.244.2.3     minikube-m03   <none>           <none>
default       net-test-c4f9cfdd4-jxqt8           1/1     Running   0          43s     10.244.0.4     minikube       <none>           <none>
default       net-test-c4f9cfdd4-wkfrg           1/1     Running   0          43s     10.244.2.4     minikube-m03   <none>           <none>
kube-system   coredns-f9fd979d6-wcdm8            1/1     Running   0          3m1s    10.244.0.2     minikube       <none>           <none>
```

Pods running in worker nodes are able to resolve dns and connect.
```
kubectl exec -it net-test-c4f9cfdd4-9bsks -- /bin/bash
bash-5.0# curl google.com
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="http://www.google.com/">here</A>.
</BODY></HTML>
bash-5.0# curl my-nginx.default
<!DOCTYPE html>
<html>
<head>
<title>Welcome to nginx!</title>
<style>
    body {
        width: 35em;
        margin: 0 auto;
        font-family: Tahoma, Verdana, Arial, sans-serif;
    }
</style>
</head>
<body>
<h1>Welcome to nginx!</h1>
<p>If you see this page, the nginx web server is successfully installed and
working. Further configuration is required.</p>

<p>For online documentation and support please refer to
<a href="http://nginx.org/">nginx.org</a>.<br/>
Commercial support is available at
<a href="http://nginx.com/">nginx.com</a>.</p>

<p><em>Thank you for using nginx.</em></p>
</body>
</html>
```

Running with only single node also have the kindnet enabled as CNI and getting ips from the podCIDR. 
```
$ kubectl get pods --all-namespaces -o wide
NAMESPACE     NAME                               READY   STATUS    RESTARTS   AGE    IP             NODE       NOMINATED NODE   READINESS GATES
default       my-nginx-5b56ccd65f-56gnr          1/1     Running   0          38s    10.244.0.10    minikube   <none>           <none>
default       my-nginx-5b56ccd65f-8ntxh          1/1     Running   0          38s    10.244.0.7     minikube   <none>           <none>
default       my-nginx-5b56ccd65f-msxws          1/1     Running   0          38s    10.244.0.3     minikube   <none>           <none>
default       my-nginx-5b56ccd65f-sdm7v          1/1     Running   0          38s    10.244.0.8     minikube   <none>           <none>
default       net-test-c4f9cfdd4-8wbp8           1/1     Running   0          37s    10.244.0.9     minikube   <none>           <none>
default       net-test-c4f9cfdd4-d9hr9           1/1     Running   0          37s    10.244.0.6     minikube   <none>           <none>
default       net-test-c4f9cfdd4-klmdg           1/1     Running   0          38s    10.244.0.4     minikube   <none>           <none>
default       net-test-c4f9cfdd4-vrb2l           1/1     Running   0          37s    10.244.0.5     minikube   <none>           <none>
kube-system   coredns-f9fd979d6-25pwj            1/1     Running   0          106s   10.244.0.2     minikube   <none>           <none>
kube-system   etcd-minikube                      1/1     Running   0          119s   192.168.49.2   minikube   <none>           <none>
kube-system   kindnet-2klfc                      1/1     Running   0          106s   192.168.49.2   minikube   <none>           <none>
kube-system   kube-apiserver-minikube            1/1     Running   0          119s   192.168.49.2   minikube   <none>           <none>
kube-system   kube-controller-manager-minikube   1/1     Running   0          119s   192.168.49.2   minikube   <none>           <none>
kube-system   kube-proxy-txqpb                   1/1     Running   0          106s   192.168.49.2   minikube   <none>           <none>
kube-system   kube-scheduler-minikube            1/1     Running   0          119s   192.168.49.2   minikube   <none>           <none>
kube-system   storage-provisioner                1/1     Running   0          118s   192.168.49.2   minikube   <none>           <none>
```

### DIFF
--- a/pkg/minikube/bootstrapper/bsutil/ktmpl/v1alpha3.go
+++ b/pkg/minikube/bootstrapper/bsutil/ktmpl/v1alpha3.go
@@ -57,6 +57,7 @@ etcd:
   local:
     dataDir: {{.EtcdDataDir}}
 controllerManagerExtraArgs:
+  allocate-node-cidrs: "true"
   leader-elect: "false"
 schedulerExtraArgs:
   leader-elect: "false"
--- a/pkg/minikube/cni/cni.go
+++ b/pkg/minikube/cni/cni.go
@@ -93,6 +93,21 @@ func New(cc config.ClusterConfig) (Manager, error) {
 	}
 }
 
+func IsDisabled(cc config.ClusterConfig) bool {
+	if cc.KubernetesConfig.NetworkPlugin != "" && cc.KubernetesConfig.NetworkPlugin != "cni" {
+		return true
+	}
+
+	if cc.KubernetesConfig.CNI == "false" {
+		return true
+	}
+
+	if chooseDefault(cc).String() == "Disabled" {
+		return true
+	}
+	return false
+}
+
 func chooseDefault(cc config.ClusterConfig) Manager {
 	// For backwards compatibility with older profiles using --enable-default-cni
 	if cc.KubernetesConfig.EnableDefaultCNI {
@@ -109,23 +124,20 @@ func chooseDefault(cc config.ClusterConfig) Manager {
 		return Bridge{cc: cc}
 	}
 
-	// Recommending KindNet for single and multi node clusters as well. This comes at a very small
-	// cost of having one extra KindNet pod running in all control plane node, but provides the
-	// following benefits:
-	//   - With Disabled CNI, Single node cluster doesn't respect the podCIDR even users ask for it
-	// via --extra-config=kubeadm.pod-network-cidr=10.244.0.0/16. This results in pods having ips
-	// like 172.17.0.2.
-	//
-	//   - Enables KindNet CNI in master in multi node cluster, This solves the network problem
-	// inside pod for multi node clusters. See https://github.com/kubernetes/minikube/issues/9838.
-	//
-	//   - Reduces the complexities for reconfiguring the control plane node, in a scenario
-	// where minikube cluster is created as a single node cluster and more node was added later.
-	//
-	// If a --cni flag is explicitly specified we do not do a node length check, this behaviour should
-	// also be persevered for chooseDefault.
-	klog.Infof("Recommending kindnet as default CNI", len(cc.Nodes))
-	return KindNet{cc: cc}
+	if driver.BareMetal(cc.Driver) {
+		klog.Infof("Driver %s used, CNI unnecessary in this configuration, recommending no CNI", cc.Driver)
+		return Disabled{cc: cc}
+	}
+
+	if len(cc.Nodes) > 1 || cc.MultiNodeRequested {
+		// Enables KindNet CNI in master in multi node cluster, This solves the network problem
+		// inside pod for multi node clusters. See https://github.com/kubernetes/minikube/issues/9838.
+		klog.Infof("%d nodes found, recommending kindnet", len(cc.Nodes))
+		return KindNet{cc: cc}
+	}
+
+	klog.Infof("CNI unnecessary in this configuration, recommending no CNI")
+	return Disabled{cc: cc}
 }
 
 // manifestPath returns the path to the CNI manifest
--- a/pkg/minikube/cni/cni.go
+++ b/pkg/minikube/cni/cni.go
@@ -109,13 +109,23 @@ func chooseDefault(cc config.ClusterConfig) Manager {
 		return Bridge{cc: cc}
 	}
 
-	if len(cc.Nodes) > 1 {
-		klog.Infof("%d nodes found, recommending kindnet", len(cc.Nodes))
-		return KindNet{cc: cc}
-	}
-
-	klog.Infof("CNI unnecessary in this configuration, recommending no CNI")
-	return Disabled{}
+	// Recommending KindNet for single and multi node clusters as well. This comes at a very small
+	// cost of having one extra KindNet pod running in all control plane node, but provides the
+	// following benefits:
+	//   - With Disabled CNI, Single node cluster doesn't respect the podCIDR even users ask for it
+	// via --extra-config=kubeadm.pod-network-cidr=10.244.0.0/16. This results in pods having ips
+	// like 172.17.0.2.
+	//
+	//   - Enables KindNet CNI in master in multi node cluster, This solves the network problem
+	// inside pod for multi node clusters. See https://github.com/kubernetes/minikube/issues/9838.
+	//
+	//   - Reduces the complexities for reconfiguring the control plane node, in a scenario
+	// where minikube cluster is created as a single node cluster and more node was added later.
+	//
+	// If a --cni flag is explicitly specified we do not do a node length check, this behaviour should
+	// also be persevered for chooseDefault.
+	klog.Infof("Recommending kindnet as default CNI", len(cc.Nodes))
+	return KindNet{cc: cc}
 }
 
 // manifestPath returns the path to the CNI manifest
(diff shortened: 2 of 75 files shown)


pkg/minikube/cni/disabled.go
```diff
@@ -47,5 +47,6 @@ func (c Disabled) Apply(r Runner) error {
 
 // CIDR returns the default CIDR used by this CNI
 func (c Disabled) CIDR() string {
-	return ""
+	// Even without any CNI we want our nodes to have spec.PodCIDR set.
+	return DefaultPodCIDR
```
azhao155, line 51: why disabled need to return DefaultPodCIDR since right now we already change default cni as kindnet for both single node and multi nodes, and kindnet already has spec.PodCIDR.

func (c KindNet) CIDR() string {
	return DefaultPodCIDR
}

azhao155 (commented):
Question about change for disable cni

pkg/minikube/cni/disabled.go
sadlil: Actually it doesn't now as we set kindnet as default. But now that we are enabling `allocate-node-cidrs` the podCIDR value can't be empty. Empty podCIDR will result in kube-controller-manager errors. Keeping this as failsafe/default value if someone force uses disabled cni via https://github.com/kubernetes/minikube/blob/master/pkg/minikube/cni/cni.go#L79.

sadlil: yes.

pkg/minikube/bootstrapper/bsutil/ktmpl/v1alpha3.go
```diff
@@ -57,6 +57,7 @@ etcd:
   local:
     dataDir: {{.EtcdDataDir}}
 controllerManagerExtraArgs:
+  allocate-node-cidrs: "true"
```
medyagh, line 60: I am curious, other versions v1beta1 and 1vbeta2 need no addition to yam ?
I think this yaml file is only used for very old kuberentes versions...

pkg/minikube/cni/cni.go
```diff
@@ -109,13 +109,23 @@ func chooseDefault(cc config.ClusterConfig) Manager {
...
+	//   - Enables KindNet CNI in master in multi node cluster, This solves the network problem
+	// inside pod for multi node clusters. See https://github.com/kubernetes/minikube/issues/9838.
+	//
+	//   - Reduces the complexities for reconfiguring the control plane node, in a scenario
+	// where minikube cluster is created as a single node cluster and more node was added later.
+	//
+	// If a --cni flag is explicitly specified we do not do a node length check, this behaviour should
+	// also be persevered for chooseDefault.
+	klog.Infof("Recommending kindnet as default CNI", len(cc.Nodes))
+	return KindNet{cc: cc}
```
medyagh, line 128: for this PR could we keep the existing behaviour (single node have no CNI pod) ? 
a lot of our users will never need multinode

we could automaticalyy enable the CNi only if the user specifies the pod subnet through flags...otherwise for a lot of our IDE users I think having least amount of CPU usage and least amount of pods running be idea.

medyagh (changes requested):

pkg/minikube/bootstrapper/bsutil/ktmpl/v1alpha3.go
sadlil: For other versions template it comes from the `versionSpecificOpts` options in version.go.

pkg/minikube/cni/cni.go
sadlil: Is there any known way to measure cpu and memory usages for minikube? I would like to compare the performances for these with and without kindnet.
What I am worried about is - if a minikube cluster is created with only single node now and more node was added later? how will we fix the cni in already running control plane nodes?

medyagh: thats a good quesiton @priyawadhwa has done a lot of work in this area. @tstromberg also has script for it.
It would be great if we had this on website or as a shell script to run manually .

medyagh: https://github.com/tstromberg/time-to-k8s

tstromberg: https://github.com/tstromberg/time-to-k8s/blob/master/minikube-cni.yaml has a CNI benchmark test example you can use:

`go run . --config minikube-cni.yaml --iterations=4` should be sufficient to get the test results.

It outputs a CSV, which can be used for further analysis, such as: https://docs.google.com/spreadsheets/d/1oZq04X0QmHvucCLKNmLsvTJBdv75dOZnPP9dssOLteU/edit?usp=sharing

sadlil: Thanks for sharing those, let me take a look at them.

sharifelgamal:
I believe the none driver test failures are because of the change to kindnet by default. I think we need to disable CNI for baremetal no matter what.

pkg/minikube/cni/cni.go
tstromberg: https://docs.google.com/spreadsheets/d/1Rv7KCtVUYeMhFFzgY72J2T-ti3KDEdStggy3KyRfgt0/edit#gid=1614668143 has an initial graph, but it only has 2 iterations, so it shouldn't be considered authoritative.

sadlil: Thank you for the doc @tstromberg.

@medyagh while I am debugging why there is the extra delay setting up CNIs, how about an alternate solution to solve problem with multiple nodes. 

We set and store a cluster config something like `CreatedWithMultiNode bool`  true whenever user provides `--node > 1`. And we use this value to decide the CNI between KindNet and Disabled. And when a cluster is created with only control plane node and a node is added later, we show a warning that multi node networking might be broken (if it is not reconfigured already).

tstromberg (approved):

medyagh (approved):
this is a great PR and I can't thank you more for fixing such mystrious bug in minikube

PR #9897, opened 2020-12-09
update default kubernetes version to 1.20.0

medyagh:


sharifelgamal:
TestKubernetesUpgrade seems to failing across drivers

medyagh:
> TestKubernetesUpgrade seems to failing across drivers

good catch, I think the Newest version is not supported
v1.20.1-rc.0

I just have to stick with v1.20.0

sharifelgamal (approved):

PR #9918, opened 2020-12-10
Remove verbose options

### LINKED ISSUE #9916: Remove verbose log `-v`  in commands like tar
When I am developing multi control plane feature, I need to view kubeadm and other commands output so I made a change like this:
```go
func (s *SSHRunner) RunCmd(cmd *exec.Cmd) (*RunResult, error) {
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr
```
Then I found some commands like `tar` uses a `-v` option, produces a massive log which is useless:
```go
r.Runner.RunCmd(exec.Command("sudo", "tar", "-I", "lz4", "-C", "/var", "-xvf", dest));
```

Could we just remove these verbose options?

/kind cleanup

lingsamuel:
Fixes #9916
/kind cleanup

### DIFF
--- a/hack/preload-images/generate.go
+++ b/hack/preload-images/generate.go
@@ -181,7 +181,7 @@ func createImageTarball(tarballFilename, containerRuntime string) error {
 		dirs = append(dirs, "./lib/containers")
 	}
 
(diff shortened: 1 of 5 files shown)


medyagh (approved):
thank you for this PR this makes the log files readable.

PR #9922, opened 2020-12-10
Only use sudo with the none driver when on linux

### LINKED ISSUE #9919: Starting minikube requires sudo due to error with certs copy
Something changed in `copyHostCerts`, that runs locally instead of on the remote ?

`sudo cp -a /tmp/minikube123456789 ~/.minikube/ca.pem`

This fails, and is of course wrong to start with. It should not have needed `sudo`...

Using the virtualbox driver, other commands ran properly.

afbjorklund:
Closes #9919


tstromberg (approved):

tstromberg:
Tested locally on Darwin/Linux, doesn't seem to break things.

PR #9923, opened 2020-12-10
make tests more reliable

sharifelgamal:

### DIFF
--- a/hack/jenkins/common.sh
+++ b/hack/jenkins/common.sh
@@ -44,7 +44,7 @@ sudo ./installers/check_install_golang.sh "1.15.5" "/usr/local" || true
 sudo ./installers/check_install_docker.sh
 
 # let's just clean all docker artifacts up
-docker system prune --all --force --volumes || true
+docker system prune --force --volumes || true
 docker system df || true
 
 echo ">> Starting at $(date)"
--- a/hack/jenkins/common.sh
+++ b/hack/jenkins/common.sh
@@ -30,6 +30,8 @@ export GOPATH="$HOME/go"
 export KUBECONFIG="${TEST_HOME}/kubeconfig"
 export PATH=$PATH:"/usr/local/bin/:/usr/local/go/bin/:$GOPATH/bin"
 
+readonly TIMEOUT=${1:-70m}
+
 if [ "$(uname)" != "Darwin" ]; then
   # install lsof for finding none driver procs, psmisc to use pstree in cronjobs
   sudo apt-get -y install lsof psmisc
@@ -41,8 +43,8 @@ sudo ./installers/check_install_golang.sh "1.15.5" "/usr/local" || true
 # install docker and kubectl if not present
 sudo ./installers/check_install_docker.sh
 
-docker rm -f -v $(docker ps -aq) >/dev/null 2>&1 || true
-docker volume prune -f || true
+# let's just clean all docker artifacts up
+docker system prune --all --force --volumes || true
 docker system df || true
 
 echo ">> Starting at $(date)"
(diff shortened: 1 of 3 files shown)


hack/jenkins/common.sh
```diff
@@ -41,8 +43,8 @@ sudo ./installers/check_install_golang.sh "1.15.5" "/usr/local" || true
 # install docker and kubectl if not present
 sudo ./installers/check_install_docker.sh
 
-docker rm -f -v $(docker ps -aq) >/dev/null 2>&1 || true
-docker volume prune -f || true
+# let's just clean all docker artifacts up
+docker system prune --all --force --volumes || true
```
medyagh, line 47: does this remove all the cached docker  images too ? 
in my experinece this was adding a lot trouble with docker rate limit and had to redownload kicbase images again and again each single time it runs..


but if --volumes (makes it not delete images  I m okay with this)

hack/jenkins/linux_integration_tests_virtualbox.sh
```diff
@@ -28,9 +28,10 @@ set -e
...
 VM_DRIVER="virtualbox"
 JOB_NAME="VirtualBox_Linux"
+EXTRA_TEST_ARGS="-timeout-multiplier=2"
 EXPECTED_DEFAULT_DRIVER="kvm2"
 
 mkdir -p cron && gsutil -qm rsync "gs://minikube-builds/${MINIKUBE_LOCATION}/cron" cron || echo "FAILED TO GET CRON FILES"
 sudo install cron/cleanup_and_reboot_Linux.sh /etc/cron.hourly/cleanup_and_reboot || echo "FAILED TO INSTALL CLEANUP"
 
-source ./common.sh
+source ./common.sh 3h
```
medyagh, line 37: is this 3 hours really needed?

hack/jenkins/linux_integration_tests_virtualbox.sh
```diff
@@ -28,9 +28,10 @@ set -e
 OS_ARCH="linux-amd64"
 VM_DRIVER="virtualbox"
 JOB_NAME="VirtualBox_Linux"
+EXTRA_TEST_ARGS="-timeout-multiplier=2"
```
medyagh, line 31: the intent of the multiplier was for slow machines, I dont consider our GPU machines to be slow one. I don think this would help in fixing the failures

medyagh (changes requested):

hack/jenkins/linux_integration_tests_virtualbox.sh
sharifelgamal: this is mostly for testing, we're consistently getting really close to the 70 minute timeout even when we're succeeding

hack/jenkins/linux_integration_tests_virtualbox.sh
sharifelgamal: virtualbox is just very slow in general, it doesn't really have anything to do with the machine in this case

hack/jenkins/common.sh
sharifelgamal: this does delete images, but that was the point. there were a couple of debian agents that got messed up because there were SO many images in docker that it was causing failures for the none driver.

medyagh: there is also one option that deletes only ancient images that one is better

hack/jenkins/linux_integration_tests_virtualbox.sh
medyagh: if we prove virtualbox is slow, we should fix that in minikube code not hack our way in tests.
in my experience vbox is slow to create machines, because it allocates memory for real. not dymanmically but after that it is solid and fast.  so in minikube code we could adjust the waiting for host creation for vbox to be more than other drivers.

this was meant to be used only in slow machines.

hack/jenkins/common.sh
sharifelgamal: ok switched it to only delete dangling images

hack/jenkins/linux_integration_tests_virtualbox.sh
sharifelgamal: removed

hack/jenkins/linux_integration_tests_virtualbox.sh
sharifelgamal: lowered to 2 hours

PR #9963, opened 2020-12-14
Upgrade kubernetes dashboard to v2.1.0 for 1.20

afbjorklund:
Now with support for Kubernetes version 1.20 too

https://github.com/kubernetes/dashboard/releases/tag/v2.1.0


afbjorklund:
This will probably affect the preload

afbjorklund:
And the bump is done just before generating the release preload, right ?

priyawadhwa:
Yup, once the PR is merged then the release will kick off

afbjorklund:
@priyawadhwa : my question was actually it should be included in _this_ commit

Or if there will be a separate preload version bump, later on ? (in the release process)

priyawadhwa:
Ah, sorry! I think it's fine to include it in this PR.

PR #9974, opened 2020-12-16
Adding warning for docker windows tunnel

blueelvis:
Adds a warning for users using minikube with Docker on Windows to upgrade to latest version of SSH.

### DIFF
--- a/pkg/minikube/tunnel/kic/ssh_conn.go
+++ b/pkg/minikube/tunnel/kic/ssh_conn.go
@@ -79,6 +79,11 @@ func createSSHConn(name, sshPort, sshKey string, svc *v1.Service) *sshConn {
 		command = "sudo"
 		sshArgs = append([]string{"ssh"}, sshArgs...)
 	}
+
+	if askForSudo && runtime.GOOS == "windows" {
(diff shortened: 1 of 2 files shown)


pkg/minikube/tunnel/kic/ssh_conn.go
```diff
@@ -79,6 +79,11 @@ func createSSHConn(name, sshPort, sshKey string, svc *v1.Service) *sshConn {
 		command = "sudo"
 		sshArgs = append([]string{"ssh"}, sshArgs...)
 	}
+
+	if askForSudo && runtime.GOOS == "windows" {
+		out.Step(style.Warning, "Access to ports <1024 may not work with some SSH clients. Kindly refer to the docs over here - https://minikube.sigs.k8s.io/docs/handbook/accessing/#access-to-ports-1024-on-windows-requires-root-permission")
```
tstromberg, line 84: To fit within Kubernetes style guidance, let's see if we can remove some of the unnecessary politeness:

https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language

What do you think about:

`Access to ports below 1024 may fail on Windows with OpenSSH clients older than v8.1. For more information, see:  <url>`

tstromberg: I also feel that `out.WarningT()` is more appropriate here.

blueelvis: Really appreciate your review comments Thomas!

medyagh (approved):

PR #9978, opened 2020-12-17
Removed parallel running from test func to prevent ctx timeout

### LINKED ISSUE #9905: PersistentVolumeClaim time out causes all other functional test fail
I believe this time out causes all other functional tests fail
https://storage.googleapis.com/minikube-builds/logs/9897/34ff43c/KVM_Linux.html#fail_TestFunctional%2fparallel%2fPersistentVolumeClaim

I believe the Context with timeout that is being passed from the parent test is Getting Cancleled in some Defer (maybe postmortem) and that causes all other tests to faill to

for example 
```
TestFunctional/parallel/DryRun
```

fails in less than 232 Nano Seconds (means the context was canceled before it tired)
```
functional_test.go:511: (dbg) Non-zero exit: out/minikube-linux-amd64 start -p functional-20201209230615-12455 --dry-run --alsologtostderr -v=1 --driver=kvm2 : context deadline exceeded (232ns)
functional_test.go:513: dry-run exit code = 0, wanted = 0: context deadline exceeded
```

I suggest whoever picks this up to, replicat manally by making Tests Do nothing (just time sleep)

and then make the time sleep 10 seconds..and timeout 10 secconds..
and ensure if PersistentVolumeClaim Timesout it should Not Cancel or affect other tests which are running in parallel

spowelljr:
Should fix #9905 

Issue is that context timeout was started, but tests were marked to run in parallel with other tests, so if unlucky, tests would run at the end after ctx had already timed out, hence the tests intermittently failing.

### DIFF
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -102,9 +102,6 @@ func TestFunctional(t *testing.T) {
 		}
 	})
 
-	// Now that we are out of the woods, lets go.
-	MaybeParallel(t)
-
 	// Parallelized tests
 	t.Run("parallel", func(t *testing.T) {
 		tests := []struct {


tstromberg:
What's the run-time impact of this PR?

spowelljr:
> What's the run-time impact of this PR?

@tstromberg I was estimating that testing would take ~70 seconds longer. The other option is to increase the timeout.

PR #9983, opened 2020-12-17
RSRC_DOCKER_STORAGE advice: remove typo and improve text

tstromberg:
I noticed it mentioned memory instead of storage.


medyagh (approved):
thanks for fixing this

PR #9986, opened 2020-12-17
fix ip node retrieve for none driver

sharifelgamal:
We hardcode a node name for the none driver, which broke minikube ip when we added the `node.Retrieve` call to it.


tstromberg (approved):
It seems like this shouldn't be a special case -- that cc.Nodes should have the correct data in it.

That said, I'm approving this to unblock the release.

sharifelgamal:
> It seems like this shouldn't be a special case -- that cc.Nodes should have the correct data in it.
> 
> That said, I'm approving this to unblock the release.

cc.Nodes does have the correct info, it's the request default of empty that's bad, but only for the none driver.

PR #9988, opened 2020-12-17
Performance bot wait for binary before run

### LINKED ISSUE #9697: performance bot wait for binary before run
as seen in comment, performance bot is running and erroring

https://github.com/kubernetes/minikube/pull/9693#issuecomment-726993071
```

kvm2 Driver
error collecting results for kvm2 driver: timing run 0 with Minikube (PR 9693): timing cmd: [/home/performance-monitor/.minikube/minikube-binaries/9693/minikube start --driver=kvm2]: starting cmd: fork/exec /home/performance-monitor/.minikube/minikube-binaries/9693/minikube: exec format error
docker Driver
error collecting results for docker driver: timing run 0 with Minikube (PR 9693): timing cmd: [/home/performance-monitor/.minikube/minikube-binaries/9693/minikube start --driver=docker]: starting cmd: fork/exec /home/performance-monitor/.minikube/minikube-binaries/9693/minikube: exec format error
```

need to add logic to retry the test or wait till binary is available

daehyeok:
fixes #9697

### DIFF
--- a/pkg/minikube/perf/binary.go
+++ b/pkg/minikube/perf/binary.go
@@ -25,11 +25,13 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"time"
 
 	"cloud.google.com/go/storage"
 	"github.com/pkg/errors"
 	"google.golang.org/api/option"
 	"k8s.io/minikube/pkg/minikube/constants"
+	"k8s.io/minikube/pkg/util/retry"
 )
 
 // Binary holds a minikube binary
@@ -109,9 +111,11 @@ func newBinaryFromPR(pr string) (*Binary, error) {
 		path: localMinikubePath(i),
 		pr:   i,
 	}
-	if err := b.download(); err != nil {
+
+	if err := retry.Expo(b.download, 1*time.Minute, 10*time.Minute); err != nil {
 		return nil, errors.Wrapf(err, "downloading binary")
 	}
+
 	return b, nil
 }
 


pkg/minikube/perf/binary.go
```diff
@@ -70,10 +114,9 @@ func (b *Binary) download() error {
 	}
 	defer client.Close()
 
-	// first make sure object exists
-	obj := client.Bucket(bucket).Object(fmt.Sprintf("%d/minikube-%s-amd64", b.pr, runtime.GOOS))
-	if _, err := obj.Attrs(ctx); err != nil {
-		return fmt.Errorf("minikube binary for pr %v does not exist in bucket", b.pr)
+	obj, err := newObjectFromStorage(ctx, client, b.pr)
```
priyawadhwa, line 117: Why was this code moved into a new function? It doesn't seem like there is any difference between `newObjectFromStorage` and the code that was here before

pkg/minikube/perf/binary.go
```diff
@@ -98,11 +141,9 @@ func (b *Binary) download() error {
 
 // newBinaryFromPR downloads the minikube binary built for the pr by Jenkins from GCS
 func newBinaryFromPR(pr string) (*Binary, error) {
-	pr = strings.TrimPrefix(pr, prPrefix)
-	// try to convert to int
-	i, err := strconv.Atoi(pr)
+	i, err := extractPRIDFromPath(pr)
```
priyawadhwa, line 144: We shouldn't need to extract the PR ID from path, since it is passed into the function as a string

priyawadhwa (commented):
Hey @daehyeok thank you for taking this issue on. I've left a few comments, but overall I think this can be simplified considerably. Instead of writing new functions, we can just set a timer on `b.Download()` in `mkcmp`, and continuously retry for up to 10 minutes. WDYT?

daehyeok (commented):
mkcmp run as external command. I'm not sure how can i handling error from mkcmp.

priyawadhwa (commented):
Hey @daehyeok sorry I wasn't clear.

We try to download the binary here:

https://github.com/kubernetes/minikube/blob/92511decd0ece326701c385f1c642d11af092d16/pkg/minikube/perf/binary.go#L112

and we fail if it doesn't work. We could add a retry here, to continuously attempt to download the binary once every minute for 10 minutes. I think 10 minutes is a safe bet because jenkins take around 4 minutes to build and upload the binaries.

We shouldn't need any additional code.

daehyeok:
@priyawadhwa  oh sorry i misunderstood. I thought  you want to mkcmp return error but still retying in `monitor/execute.go`.
I will fix it.

tstromberg (approved):

PR #9990, opened 2020-12-17
Update Makefile and release notes for 1.16.0 release

sharifelgamal:


tstromberg:
Some registry-server OOM's we should sort out, but I've seen this before and they seem unrelated to this release. Just going to check locally.

PR #9996, opened 2020-12-18
Fix profile list when there are multi node clusters 

### LINKED ISSUE #9772: profile list for multi node shows error wrongfully
as noted by @sharifelgamal the missing part of this PR is left https://github.com/kubernetes/minikube/pull/9702#issuecomment-731445695
```

|----------|-----------|---------|--------------|------|---------|---------|-------|
| Profile  | VM Driver | Runtime |      IP      | Port | Version | Status  | Nodes |
|----------|-----------|---------|--------------|------|---------|---------|-------|
| minikube | docker    | docker  | 192.168.49.2 | 8443 | v1.19.4 | Running |     1 |
| p2       | docker    | docker  | 192.168.59.2 | 8443 | v1.19.4 | Running |     2 |
|----------|-----------|---------|--------------|------|---------|---------|-------|
❗  Found 1 invalid profile(s) ! 
	 p2-m02
	 $ minikube delete -p p2-m02 
```

daehyeok:
Fix problem which ListProfiles return sub-nodes in Mult-Node clusters as
a part of inValidPs.

fixes #9772
 
before
```
❯ minikube profile list
|----------|-----------|---------|--------------|------|---------|---------|-------|
| Profile  | VM Driver | Runtime |      IP      | Port | Version | Status  | Nodes |
|----------|-----------|---------|--------------|------|---------|---------|-------|
| minikube | docker    | docker  | 192.168.49.2 | 8443 | v1.20.0 | Running |     2 |
|----------|-----------|---------|--------------|------|---------|---------|-------|
❗  Found 1 invalid profile(s) !
	 minikube-m02
	 $ minikube delete -p minikube-m02
```
after
```
❯ minikube profile list
|----------|-----------|---------|--------------|------|---------|---------|-------|
| Profile  | VM Driver | Runtime |      IP      | Port | Version | Status  | Nodes |
|----------|-----------|---------|--------------|------|---------|---------|-------|
| minikube | docker    | docker  | 192.168.49.2 | 8443 | v1.20.0 | Running |     2 |
|----------|-----------|---------|--------------|------|---------|---------|-------|
❗  Found 0 invalid profile(s) !
```

### DIFF
--- a/pkg/minikube/config/profile.go
+++ b/pkg/minikube/config/profile.go
@@ -207,8 +207,9 @@ func ListProfiles(miniHome ...string) (validPs []*Profile, inValidPs []*Profile,
 	if err == nil {
 		pDirs = append(pDirs, cs...)
 	}
-	pDirs = removeDupes(pDirs)
-	for _, n := range pDirs {
+
+	nodeNames := map[string]bool{}
+	for _, n := range removeDupes(pDirs) {
 		p, err := LoadProfile(n, miniHome...)
 		if err != nil {
 			inValidPs = append(inValidPs, p)
@@ -219,7 +220,13 @@ func ListProfiles(miniHome ...string) (validPs []*Profile, inValidPs []*Profile,
 			continue
 		}
 		validPs = append(validPs, p)
+
+		for _, child := range p.Config.Nodes {
+			nodeNames[MachineName(*p.Config, child)] = true
+		}
 	}
+
+	inValidPs = removeChildNodes(inValidPs, nodeNames)
 	return validPs, inValidPs, nil
 }
 
@@ -241,6 +248,18 @@ func removeDupes(profiles []string) []string {
 	}
 	// Return the new slice.
 	return result
+}
+
+// removeChildNodes remove invalid profiles which have a same name with any sub-node's machine name
+// it will return nil if invalid profiles are not exists.
+func removeChildNodes(inValidPs []*Profile, nodeNames map[string]bool) (ps []*Profile) {
+	for _, p := range inValidPs {
+		if _, ok := nodeNames[p.Name]; !ok {
+			ps = append(ps, p)
+		}
+	}
+
+	return ps
 }
 
 // LoadProfile loads type Profile based on its name
--- a/test/integration/multinode_test.go
+++ b/test/integration/multinode_test.go
@@ -20,10 +20,13 @@ package integration
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os/exec"
 	"strings"
 	"testing"
+
+	"k8s.io/minikube/pkg/minikube/config"
 )
 
 func TestMultiNode(t *testing.T) {
@@ -43,6 +46,7 @@ func TestMultiNode(t *testing.T) {
 		}{
 			{"FreshStart2Nodes", validateMultiNodeStart},
 			{"AddNode", validateAddNodeToMultiNode},
+			{"ProfileList", validateProfileListWithMultiNode},
 			{"StopNode", validateStopRunningNode},
 			{"StartAfterStop", validateStartNodeAfterStop},
 			{"DeleteNode", validateDeleteNodeFromMultiNode},
@@ -107,6 +111,44 @@ func validateAddNodeToMultiNode(ctx context.Context, t *testing.T, profile strin
 	if strings.Count(rr.Stdout.String(), "kubelet: Running") != 3 {
 		t.Errorf("status says all kubelets are not running: args %q: %v", rr.Command(), rr.Stdout.String())
 	}
+}
+
+func validateProfileListWithMultiNode(ctx context.Context, t *testing.T, profile string) {
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), "profile", "list", "--output", "json"))
+	if err != nil {
+		t.Errorf("failed to list profiles with json format. args %q: %v", rr.Command(), err)
+	}
+
+	var jsonObject map[string][]config.Profile
+	err = json.Unmarshal(rr.Stdout.Bytes(), &jsonObject)
+	if err != nil {
+		t.Errorf("failed to decode json from profile list: args %q: %v", rr.Command(), err)
+	}
+
+	validProfiles := jsonObject["valid"]
+	var profileObject *config.Profile
+	for _, obj := range validProfiles {
+		if obj.Name == profile {
+			profileObject = &obj
+			break
+		}
+	}
+
+	if profileObject == nil {
+		t.Errorf("expected the json of 'profile list' to include %q but got *%q*. args: %q", profile, rr.Stdout.String(), rr.Command())
+	} else if expected, numNodes := 3, len(profileObject.Config.Nodes); expected != numNodes {
+		t.Errorf("expected profile %q in json of 'profile list' include %d nodes but have %d nodes. got *%q*. args: %q", profile, expected, numNodes, rr.Stdout.String(), rr.Command())
+	}
+
+	if invalidPs, ok := jsonObject["invalid"]; ok {
+		for _, ps := range invalidPs {
+			if strings.Contains(ps.Name, profile) {
+				t.Errorf("expected the json of 'profile list' to not include profile or node in invalid profile but got *%q*. args: %q", rr.Stdout.String(), rr.Command())
+			}
+		}
+
+	}
+
 }
 
 func validateStopRunningNode(ctx context.Context, t *testing.T, profile string) {


pkg/minikube/config/profile.go
```diff
@@ -242,6 +250,27 @@ func removeDupes(profiles []string) []string {
 	return result
 }
 
+// TODO: MachineName is copied from driver to avoid dependency cycle. Remove it or move to other module
+// MachineName returns the name of the machine, as seen by the hypervisor given the cluster and node names
+func MachineName(cc *ClusterConfig, n *Node) string {
```
medyagh, line 255: lets move this a package that everyone can use, and we could refactor all other existances of this function

pkg/minikube/config/profile.go
```diff
@@ -206,8 +207,9 @@ func ListProfiles(miniHome ...string) (validPs []*Profile, inValidPs []*Profile,
 	if err == nil {
 		pDirs = append(pDirs, cs...)
```
medyagh, line 208: the source of the problem is this ListOwnedContainers, that returns "all continuers" including multi node ones
for example it returns

```
[ "minikube",  "minikube-m02"]
```

we could add a check here, to append only if it is not a node conaterin

```
	// try to get profiles list based on all containers created by docker driver
	cs, err := oci.ListOwnedContainers(oci.Docker)
	if err == nil {
                   // add logic here to check if it is primary container or worker node container
		pDirs = append(pDirs, cs...)
	}
```

medyagh (changes requested):

pkg/minikube/config/profile.go
medyagh: additionaly we can make the oci.ListOwnedContainers(oci.Docker) to have an option to only return the Master Containers

medyagh (changes requested):

pkg/minikube/config/profile.go
daehyeok: Which package is good for new home of this func?

pkg/minikube/config/profile.go
daehyeok: I agree fix `ListOwnedContainers` is the best way.
Could you let me know how can i filter nodes?

medyagh: unfornatunately in multi node implementaiotn we didnt use the labels https://githu
b.com/medyagh/minikube/blob/9e5beb4e37298d296694cd7339740e336c62a653/pkg/drivers/kic/oci/types.go#L39

so we can't filter the nodes by master there ... ( we could do that for future releases though)
but one simple thing we can do is filter out any -m01 -m02 (digits) from the returned list
so it only returns minikube and not minikbe-m02

medyagh (changes requested):

pkg/minikube/config/profile.go
daehyeok: hmm I'm worrying about user make profile with -m02 postfix..
```
minikube start -p test-profile-m02
```

pkg/minikube/config/profile.go
medyagh: does it allow to put it in 'drivers"  package ?

medyagh: or can we put it in the `machine` package?
so we could call it something like this

```
machine.NameFromConfig(cc *ClusterConfig, n *Node)
```

medyagh (changes requested):

tstromberg:
Any update here? I wonder if @medyagh 's suggestion will also result in a circular dependency.

pkg/minikube/config/profile.go
medyagh: how about we refactor the MachineName func to not depend on clusterConfig or Node ?

so we pass everything we want from it ?

machine.Name(name string,  int nodes, controlPlane bool, ...)

medyagh (changes requested):

pkg/minikube/config/profile.go
daehyeok: Function required following arguments,
Cluster name, Node name, number of node in cluster, is controlPlane or not
`machine.Name(clusterName, nodeName string, int nodes, controlPlane bool)`. 
Is it okay?

medyagh: sure that sounds like a good comprimise

pkg/minikube/config/profile.go
daehyeok: Filter out any -m0x node will have a potential false-negative result with profile.
Do we have any other way??

medyagh (changes requested):
thanks for this PR 
it looks good just one thing I request to add integation test for it.

we have a TestMultiNode https://github.com/medyagh/minikube/blob/89a8cdc794098f9ba0550fbb5eeb19bc3dd8774d/test/integration/multinode_test.go#L29
lets add a Profile List to that test
and ensure the Profile List Exits Normally
only check for exit code 0
```
		}{
			{"FreshStart2Nodes", validateMultiNodeStart},
			{"AddNode", validateAddNodeToMultiNode},
			{"StopNode", validateStopRunningNode},
			{"StartAfterStop", validateStartNodeAfterStop},
			{"DeleteNode", validateDeleteNodeFromMultiNode},
			{"StopMultiNode", validateStopMultiNodeCluster},
			{"RestartMultiNode", validateRestartMultiNodeCluster},
```

lets add profile list after "FreshStart2Nodes",

daehyeok:
Test added case to integration test.

sharifelgamal:
Can we be sure that we don't show the "Found 0 invalid profiles" when there are no invalid profiles? When everything is ok, it should just show the table.

daehyeok:
> Can we be sure that we don't show the "Found 0 invalid profiles" when there are no invalid profiles? When everything is ok, it should just show the table.

How about JSON output? Should we remove empty list too ( "invalid":[] )?
```
❯ minikube profile list -o json
{"invalid":[],"valid":[{"Name":"minikube","Status":"Stopped","Config":{"Name":"minikube","KeepContext":false,"EmbedCerts":false,"MinikubeISO":"","KicBaseImage":"gcr.io/k8s-minikube/kicbase:v0.0.16-snapshot1@sha256:dff16232547bb3ac3f2a9e09a42246a96ecf8f40d9a1c5bcf5a37953690954b6","Memory":3890,"CPUs":2,"DiskSize":20000,"VMDriver":"","Driver":"docker","HyperkitVpnKitSock":"","HyperkitVSockPorts":[],"DockerEnv":null,"ContainerVolumeMounts":null,"InsecureRegistry":null,"RegistryMirror":null,"HostOnlyCIDR":"192.168.99.1/24","HypervVirtualSwitch":"","HypervUseExternalSwitch":false,"HypervExternalAdapter":"","KVMNetwork":"default","KVMQemuURI":"qemu:///system","KVMGPU":false,"KVMHidden":false,"DockerOpt":null,"DisableDriverMounts":false,"NFSShare":[],"NFSSharesRoot":"/nfsshares","UUID":"","NoVTXCheck":false,"DNSProxy":false,"HostDNSResolver":true,"HostOnlyNicType":"virtio","NatNicType":"virtio","SSHIPAddress":"","SSHUser":"","SSHKey":"","SSHPort":0,"KubernetesConfig":{"KubernetesVersion":"v1.20.0","ClusterName":"minikube","Namespace":"default","APIServerName":"minikubeCA","APIServerNames":null,"APIServerIPs":null,"DNSDomain":"cluster.local","ContainerRuntime":"docker","CRISocket":"","NetworkPlugin":"","FeatureGates":"","ServiceCIDR":"10.96.0.0/12","ImageRepository":"","LoadBalancerStartIP":"","LoadBalancerEndIP":"","CustomIngressCert":"","ExtraOptions":null,"ShouldLoadCachedImages":true,"EnableDefaultCNI":false,"CNI":"","NodeIP":"","NodePort":8443,"NodeName":""},"Nodes":[{"Name":"","IP":"192.168.49.2","Port":8443,"KubernetesVersion":"v1.20.0","ControlPlane":true,"Worker":true}],"Addons":{"default-storageclass":true,"storage-provisioner":true},"VerifyComponents":{"apiserver":true,"system_pods":true},"StartHostTimeout":360000000000,"ScheduledStop":null,"ExposedPorts":[],"Network":"","MultiNodeRequested":false}}]}%                                     
```

daehyeok:
> Can we be sure that we don't show the "Found 0 invalid profiles" when there are no invalid profiles? When everything is ok, it should just show the table.

Oops my bad.  I confused master brach also print warn. Please ignore my previous comment.

fixed it.
```
❯ minikube profile list
|----------|-----------|---------|--------------|------|---------|---------|-------|
| Profile  | VM Driver | Runtime |      IP      | Port | Version | Status  | Nodes |
|----------|-----------|---------|--------------|------|---------|---------|-------|
| minikube | docker    | docker  | 192.168.49.2 | 8443 | v1.20.0 | Stopped |     1 |
|----------|-----------|---------|--------------|------|---------|---------|-------|
```

medyagh (approved):

PR #9997, opened 2020-12-18
Improve documentation for minikube installation via brew using cask

hyunjun529:
continue to #8196

when I install minikube using brew, i found deprecated command `brew cask`.

```bash
Warning: Calling brew cask uninstall is deprecated! Use brew uninstall [--cask] instead.
```

- brew cask depreacted at 2.5.0 : https://brew.sh/2020/09/08/homebrew-2.5.0/

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -84,7 +84,7 @@ brew install minikube
 If `which minikube` fails after installation via brew, you may have to remove the minikube cask and link the binary:
 
 ```shell
-brew cask remove minikube
+brew uninstall --cask minikube
 brew link minikube
 ```
 


hyunjun529:

CLA signed

afbjorklund:
We could also consider removing those instructions completely.

The cask was removed over a year ago, with minikube v1.5.0

* https://github.com/Homebrew/homebrew-cask/pull/71160

* https://github.com/Homebrew/homebrew-core/pull/45603

priyawadhwa:
I agree with @afbjorklund that it probably makes sense to just remove the instructions around cask entirely.

spowelljr:
@hyunjun529 This PR needs a rebase, also there were a couple comments left in regards to removing the instructions entirely.

hyunjun529:
outdated

Issue #9879, opened 2020-12-08
ISO: Enable CONFIG_XFS_QUOTA


floek:
Please enable CONFIG_XFS_QUOTA in linux kernel options for the VirtualBox ISO image. This leeds to problems with the kadalu storage provider. See: https://github.com/kadalu/kadalu/issues/351

**Steps to reproduce the issue:** 

1. Boot minikube
2. try to mount a xfs filesystem with quota support.

**Full output of failed command:** 
mount: /bricks/storage-pool-1/data: wrong fs type, bad option, bad superblock on /dev/loop0, missing codepage or helper program, or other error.

**Full output of `minikube start` command used, if not already included:**
😄  minikube v1.15.1 auf Darwin 11.0.1
✨  Using the virtualbox driver based on existing profile
👍  Starting control plane node minikube in cluster minikube
🔄  Restarting existing virtualbox VM for "minikube" ...
🐳  Vorbereiten von Kubernetes v1.19.4 auf Docker 19.03.13...
🔎  Verifying Kubernetes components...
🌟  Enabled addons: storage-provisioner, default-storageclass
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default

tstromberg:
If you are willing to send a PR, we'd be happy to merge it. Here's the doc around adding kernel modules:

https://minikube.sigs.k8s.io/docs/contrib/building/iso/#adding-kernel-modules

Here's an example PR: https://github.com/kubernetes/minikube/pull/8670

PR #9999, opened 2020-12-18
iso: add XFS_QUOTA support to guest vm

amarts:
Having the Quota support for xfs would allow some of the CSI projects
to utilize the feature in their automated testing, as PV Claim requests
has volume-size as an option, which is implemented with help of quota
feature in many cases.

fixes: #9879
Signed-off-by: Amar Tumballi <amar@kadalu.io>


amarts:
@tstromberg adding you here, as this PR is done with the references you pointed in the issue. Thanks.

medyagh (approved):
thanks for fixing this @amarts it will be included in the next release. meanwhile you can build the ISO and use the --iso-url flag to use local iso

PR #10018, opened 2020-12-21
The kubeadm bootstrapper now requires 1700M mem

### LINKED ISSUE #10016: Memory requirements do not match kubeadm any longer (since 1.20)
Minikube still has:

```go
        minUsableMem            = 953  // 1GB In MiB: Kubernetes will not start with less
        minRecommendedMem       = 1907 // 2GB In MiB: Warn at no lower than existing configurations
```

Kubeadm now has:

```go
       // Below that amount of RAM running a stable control plane would be difficult.
       ControlPlaneMem = 1700
```

Since https://github.com/kubernetes/kubernetes/commit/e5bb66f89909861fc96717b0e9ca55197ec224b8
(and https://github.com/kubernetes/kubernetes/commit/c6975a77508fef3b03dcaf9f46e5118941a86724, from 2048M to 1700M)

`v1.20.0-alpha.0-252-ge5bb66f8990`
`v1.20.0-alpha.0-253-gc6975a77508`

----

Problem seen in #10014
```
 [ERROR Mem]: the system RAM (953 MB) is less than the minimum 1700 MB
```

So we now need to require 2G more strict, and maybe suggest 4G ? Like suggestd in #8050

Alternatively _try_ to run with `--ignore-preflight-errors=Mem`, like we currently do for Swap ?

afbjorklund:
The node requirement is 2 vCPU and 2.0 GiB RAM

You can use --force to bypass the requirement

Closes #10016

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -94,8 +94,8 @@ const (
 	interactive             = "interactive"
 	waitTimeout             = "wait-timeout"
 	nativeSSH               = "native-ssh"
-	minUsableMem            = 953  // 1GB In MiB: Kubernetes will not start with less
-	minRecommendedMem       = 1907 // 2GB In MiB: Warn at no lower than existing configurations
+	minUsableMem            = 1700 // Kubernetes (kubeadm) will not start with less
+	minRecommendedMem       = 1900 // Warn at no lower than existing configurations
 	minimumCPUS             = 2
 	minimumDiskSize         = 2000
 	autoUpdate              = "auto-update-drivers"
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -194,6 +194,11 @@ func (k *Bootstrapper) init(cfg config.ClusterConfig) error {
 		"Port-10250", // For "none" users who already have a kubelet online
 		"Swap",       // For "none" users who have swap configured
 	}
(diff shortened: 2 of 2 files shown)


cmd/minikube/cmd/start_flags.go
```diff
@@ -94,7 +94,7 @@ const (
 	interactive             = "interactive"
 	waitTimeout             = "wait-timeout"
 	nativeSSH               = "native-ssh"
-	minUsableMem            = 953  // 1GB In MiB: Kubernetes will not start with less
+	minUsableMem            = 1700 // 2GB In MiB: Kubernetes (kubeadm) will not start with less
```
spowelljr, line 97: The comment is slightly misleading since 2GB is 1907MiB, maybe change 2GB to 1.78GB?

afbjorklund: It's all misleading, the real value is supposed to be 1700 MiB (and the old minimum was supposed to be 1024 MiB).

Historical minikube versions and even latest kubernetes are not politically correct, but say "MB" for 1024*1024 bytes.

afbjorklund: If you have a hardware with 2048, it is unlikely to show up as >=2000 - so might as well use something like 1900.
The main thing that we want to warn about now is trying to start on a hardware with 1024, our own default is 2200.

afbjorklund: If you start a VM with the fallback allocation of 2200, the system will report that as 2135 (after the kernel "tax")

medyagh (approved):

PR #10037, opened 2020-12-23
The kubernetes memory requirement is for the total

### LINKED ISSUE #10016: Memory requirements do not match kubeadm any longer (since 1.20)

afbjorklund:
The kernel will reserve some memory for itself, so if you want
1700 MB free you need to allocate 1800 MB for the VM hardware.

The recommendation is still 2048 (or 2000), with an even 2.0 GiB.
So let's increase the old values with +100, for ~60 overhead.

See ~~#10018~~
Closes #10016

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -94,8 +94,8 @@ const (
 	interactive             = "interactive"
 	waitTimeout             = "wait-timeout"
 	nativeSSH               = "native-ssh"
-	minUsableMem            = 1700 // Kubernetes (kubeadm) will not start with less
-	minRecommendedMem       = 1900 // Warn at no lower than existing configurations
+	minUsableMem            = 1800 // Kubernetes (kubeadm) will not start with less
+	minRecommendedMem       = 2000 // Warn at no lower than existing configurations
 	minimumCPUS             = 2
 	minimumDiskSize         = 2000
 	autoUpdate              = "auto-update-drivers"


afbjorklund:
```console
$ minikube start --memory 1800
$ minikube ssh -- free -m
              total        used        free      shared  buff/cache   available
Mem:           1740         214         442         604        1083         779
Swap:             0           0           0
```

medyagh (approved):

PR #10042, opened 2020-12-24
Fix docker driver cert copy on ubuntu's snap package manager

### LINKED ISSUE #10020: fix minikube installation using gcloud snap
https://cloud.google.com/sdk/docs/downloads-snap

gcloud components install minikube

spowelljr:
Fixes #10020

### DIFF
--- a/pkg/minikube/command/kic_runner.go
+++ b/pkg/minikube/command/kic_runner.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/minikube/command/kic_runner.go
```diff
@@ -185,7 +186,11 @@ func (k *kicRunner) Copy(f assets.CopyableFile) error {
 		}
 	}
 	klog.Infof("%s (temp): %s --> %s (%d bytes)", k.ociBin, src, dst, f.GetLength())
-	tf, err := ioutil.TempFile("", "tmpf-memory-asset")
+	tmpFolder := ""
+	if isSnapBinary() {
+		tmpFolder = os.Getenv("HOME")
```
medyagh, line 191: lets use a more safe way of getting the home 
how about using  
```
home, err := os.UserHomeDir()

```

pkg/minikube/command/kic_runner.go
```diff
@@ -185,7 +186,11 @@ func (k *kicRunner) Copy(f assets.CopyableFile) error {
 		}
 	}
 	klog.Infof("%s (temp): %s --> %s (%d bytes)", k.ociBin, src, dst, f.GetLength())
-	tf, err := ioutil.TempFile("", "tmpf-memory-asset")
+	tmpFolder := ""
+	if isSnapBinary() {
```
medyagh, line 190: add comment with the link to the issue, that why we are doing this.

medyagh (changes requested):

pkg/minikube/command/kic_runner.go
spowelljr: Good idea, done

pkg/minikube/command/kic_runner.go
spowelljr: Done

medyagh (approved):

PR #10046, opened 2020-12-24
Kic: make sure NO_PROXY contains api endpoint

lingsamuel:
Signed-off-by: Ling Samuel <lingsamuelgrace@gmail.com>

We should make sure api endpoint exists in NO_PROXY.

With noProxy in .docker/config.json:
image

without noProxy in .docker/config.json:
image


medyagh (commented):
@lingsamuel thank you very much for this PR,

do u mind pasting the out of minikube after this PR. and verify that the proxy env was added ?

to do that u can do 

make kic-base-image
minikube start --base-image="...the name of the image that was produced in last step...which starts with local/kicbase..."
then u can do

minikube ssh
and cat the systemfile to show that it is there

lingsamuel:
added

medyagh (approved):
thank you for fixing this

PR #10049, opened 2020-12-26
Use appropriate ClusterDomain value in kubeadm KubeletConfiguration

### LINKED ISSUE #9881: Calico + CIDR changes result in non functioning DNS without FQDN

minikube start \
--driver='virtualbox' \
--profile=cluster0 \
--cpus=4 \
--memory=4096 \
--cni='calico' \
--extra-config=kubeadm.pod-network-cidr=10.0.1.0/24 \
--service-cluster-ip-range='10.0.2.0/24' \
--dns-domain=cluster0.local

Quick easy test:

```
curl kube-dns:53
curl: (6) Could not resolve host: kube-dns
```

If a FQDN is used, I see what I would expect:
```
curl kube-dns.kube-system.svc.cluster0.local:53
curl: (52) Empty reply from server
```

kubedns looks OK here, all system pods are running too:
```
kubectl --context cluster0 get svc -n kube-system
NAME       TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)                  AGE
kube-dns   ClusterIP   10.0.2.10    <none>        53/UDP,53/TCP,9153/TCP   55m
```

But on the node (minikube ssh), resolv.conf for some reason has a .3 address:
```
cat /etc/resolv.conf | grep nameserver
nameserver 10.0.2.3
```

I tried to override with --extra-config=kubelet.cluster-dns but did not work.

I have also tried this with v1.15

😄  [cluster0] minikube v1.14.2 on Darwin 10.14.6
👍  Starting control plane node cluster0 in cluster cluster0
🔥  Creating virtualbox VM (CPUs=4, Memory=4096MB, Disk=20000MB) ...
❗  This VM is having trouble accessing https://k8s.gcr.io
🐳  Preparing Kubernetes v1.19.2 on Docker 19.03.12 ...
    ▪ kubeadm.pod-network-cidr=10.0.1.0/24
🔗  Configuring Calico (Container Networking Interface) ...

❗  /usr/local/bin/kubectl is version 1.16.2, which may have incompatibilites with Kubernetes 1.19.2.
💡  Want kubectl v1.19.2? Try 'minikube kubectl -- get pods -A'
🏄  Done! kubectl is now configured to use "cluster0" by default

sadlil:
Fixes #9881 

User provided flag `--dns-domain` is only used in kubeadm network configurations. Kubelet clusterDomian is always set to cluster.local and that results in broken dns resolving for service names like `kube-dns.kube-system`. This PR uses the value of the flag for clusterDomain if provided or else defaults to cluster.local.
According to https://github.com/kubernetes/kubernetes/blob/cea1d4e20b4a7886d8ff65f34c6d4f95efcb4742/pkg/kubelet/apis/config/types.go#L164 ClusterDomain should be same as DNSDomain for the cluster.

### DIFF
--- a/pkg/minikube/bootstrapper/bsutil/ktmpl/v1alpha3.go
+++ b/pkg/minikube/bootstrapper/bsutil/ktmpl/v1alpha3.go
@@ -70,7 +70,7 @@ networking:
 apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
 cgroupDriver: {{.CgroupDriver}}
-clusterDomain: "cluster.local"
+clusterDomain: "{{if .DNSDomain}}{{.DNSDomain}}{{else}}cluster.local{{end}}"
 # disable disk resource management by default
 imageGCHighThresholdPercent: 100
 evictionHard:
--- a/pkg/minikube/bootstrapper/bsutil/ktmpl/v1beta1.go
+++ b/pkg/minikube/bootstrapper/bsutil/ktmpl/v1beta1.go
@@ -77,7 +77,7 @@ authentication:
   x509:
     clientCAFile: {{.ClientCAFile}}
 cgroupDriver: {{.CgroupDriver}}
-clusterDomain: "cluster.local"
+clusterDomain: "{{if .DNSDomain}}{{.DNSDomain}}{{else}}cluster.local{{end}}"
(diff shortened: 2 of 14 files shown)


medyagh:
thank you for this PR @sadlil

medyagh (approved):

Issue #10033, opened 2020-12-23
fix docker network inspect to handle networks without MTU


medyagh:
while this is rate but it might happen (if user uses a bad docker verison once and then later upgrade docker but still have a stuck old minikube network)

to replicate, create a network without mtu

```
$ docker network create m1
c681cfeab80f7cce0e16948aa6c31b2c451d03974e2ec3675ddbba54c7f3c8c9
```

## Currrent behaviour "
using this code
https://github.com/medyagh/minikube/blob/9e5beb4e37298d296694cd7339740e336c62a653/pkg/drivers/kic/oci/network_create.go#L175-L176
it will run this:
```
$ docker network inspect m1 --format '{"Name": "{{.Name}}","Driver": "{{.Driver}}","Subnet": "{{range .IPAM.Config}}{{.Subnet}}{{end}}","Gateway": "{{range .IPAM.Config}}{{.Gateway}}{{end}}","MTU": {{(index .Options "com.docker.network.driver.mtu")}},{{$first := true}} "ContainerIPs": [{{range $k,$v := .Containers }}{{if $first}}{{$first = false}}{{else}}, {{end}}"{{$v.IPv4Address}}"{{end}}]}'
```
which outputs this  (and it is missing MTU)
```
{"Name": "m1","Driver": "bridge","Subnet": "172.19.0.0/16","Gateway": "172.19.0.1","MTU": , "ContainerIPs": []}
```

we want it to fix the go template to instead return 0 for the MTU so it does NOT have a parsing error.

so it should return
## desired behaviour

```
{"Name": "m1","Driver": "bridge","Subnet": "172.19.0.0/16","Gateway": "172.19.0.1","MTU": 0 , "ContainerIPs": []}
```

_Originally posted by @baiwfg2 in https://github.com/kubernetes/minikube/issues/9811#issuecomment-748493691_

azhao155:
Has pr: https://github.com/kubernetes/minikube/pull/10053

Issue #9811, opened 2020-11-30
Template parsing error when getting docker network


fsladkey:
## Template parsing error when getting docker network

We're running some minikube deployments on CI and recently (around the 21st) we started seeing failures with the following error message when starting minikube. This is currently happening only on CI, so I don't have full detailed logs, but I'm working on reproducing it locally so I can get more.

This is from Minikube latest builds, the last published version continues to work with no issues.

Based on the timing and error message I suspect it may be related to [this PR](https://github.com/kubernetes/minikube/pull/9710) but I'm not sure!

**Steps to reproduce the issue:** 

1. Run `minikube start` with minikube from master

**Full output of `minikube start` command used, if not already included:**
```
minikube v1.15.1 on Ubuntu 16.04
      - MINIKUBE_WANTUPDATENOTIFICATION=false
    * Using the docker driver based on user configuration
    * Starting control plane node minikube in cluster minikube
    * Pulling base image ...
    * Downloading Kubernetes v1.19.4 preload ...
        > preloaded-images-k8s-v6-v1....: 42.10 MiB / 486.35 MiB [>_] 8.66% ? p/s ?    > preloaded-images-k8s-v6-v1....: 90.44 MiB / 486.35 MiB [>] 18.60% ? p/s ?    > preloaded-images-k8s-v6-v1....: 167.30 MiB / 486.35 MiB [] 34.40% ? p/s ?    > preloaded-images-k8s-v6-v1....: 230.84 MiB / 486.35 MiB  47.46% 314.62 Mi    > preloaded-images-k8s-v6-v1....: 304.00 MiB / 486.35 MiB  62.51% 314.62 Mi    > preloaded-images-k8s-v6-v1....: 382.67 MiB / 486.35 MiB  78.68% 314.62 Mi    > preloaded-images-k8s-v6-v1....: 448.80 MiB / 486.35 MiB  92.28% 317.76 Mi    > preloaded-images-k8s-v6-v1....: 486.35 MiB / 486.35 MiB  100.00% 373.80 M* Creating docker container (CPUs=2, Memory=3700MB) ...
    * Preparing Kubernetes v1.19.4 on Docker 19.03.13 ...
[...]

PR #10053, opened 2020-12-27
Fix docker inspect network go template for network which doesn't have MTU

azhao155:
First create network  without MTU

```
docker@p1:~$ docker network create m1
c3d778d8147416d2dd0d466e77c845191b26915614c3de3e4b7aacad08fd63f9
docker@p1:~$ docker network inspect m1 --format '{"Name": "{{.Name}}","Driver": "{{.Driver}}","Subnet": "{{range .IPAM.Config}}{{.Subnet}}{{end}}","Gateway": "{{range .IPAM.Config}}{{.Gateway}}{{e
nd}}","MTU": {{if (index .Options "com.docker.network.driver.mtu")}}{{(index .Options "com.docker.network.driver.mtu")}}{{else}}0{{end}},{{$first := true}} "ContainerIPs": [{{range $k,$v := .Conta
iners }}{{if $first}}{{$first = false}}{{else}}, {{end}}"{{$v.IPv4Address}}"{{end}}]}'
{"Name": "m1","Driver": "bridge","Subnet": "172.18.0.0/16","Gateway": "172.18.0.1","MTU": 0, "ContainerIPs": []}
```
Create network with MTU

```
docker@p1:~$ docker network create m2 --opt com.docker.network.driver.mtu=9216
1b07fe461e1682b8010951bd17617f1c620144a8d34e22cd4e844860db2c3ee4
docker@p1:~$ docker network inspect m2 --format '{"Name": "{{.Name}}","Driver": "{{.Driver}}","Subnet": "{{range .IPAM.Config}}{{.Subnet}}{{end}}","Gateway": "{{range .IPAM.Config}}{{.Gateway}}{{e
nd}}","MTU": {{if (index .Options "com.docker.network.driver.mtu")}}{{(index .Options "com.docker.network.driver.mtu")}}{{else}}0{{end}},{{$first := true}} "ContainerIPs": [{{range $k,$v := .Conta
iners }}{{if $first}}{{$first = false}}{{else}}, {{end}}"{{$v.IPv4Address}}"{{end}}]}'
{"Name": "m2","Driver": "bridge","Subnet": "172.19.0.0/16","Gateway": "172.19.0.1","MTU": 9216, "ContainerIPs": []}
```

So this go template will work in both scenarios.

### DIFF
--- a/pkg/drivers/kic/oci/network_create_test.go
+++ b/pkg/drivers/kic/oci/network_create_test.go
@@ -31,7 +31,7 @@ var dockerInspectGetterMock = func(name string) (*RunResult, error) {
 	return response, nil
 }
 
-func TestDockerInspectWithMTU(t *testing.T) {
+func TestDockerInspect(t *testing.T) {
 	var tests = []struct {
 		name                  string
 		dockerInspectResponse string
--- a/pkg/drivers/kic/oci/network_create_test.go
+++ b/pkg/drivers/kic/oci/network_create_test.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2019 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package oci
+
+import (
+	"bytes"
+	"net"
+	"testing"
+)
+
+var dockerResponse string
+var dockerInspectGetterMock = func(name string) (*RunResult, error) {
+	var responseInBytes bytes.Buffer
+	responseInBytes.WriteString(dockerResponse)
+	response := &RunResult{Stdout: responseInBytes}
+
+	return response, nil
+}
+
+func TestDockerInspectWithMTU(t *testing.T) {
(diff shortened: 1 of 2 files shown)


azhao155:
@medyagh

azhao155:
fix https://github.com/kubernetes/minikube/issues/10033

medyagh (commented):
how about adding a unit test for this func ? 

maybe we can have a a "mockDocker" driver

with an optinal mock parameter passed, if it is mock

then we can have a unit test for a series of examples of response...

medyagh:
we can also have a Specific Error Type to return if the user's docker has issue and user needs to upgrade docker

https://github.com/kubernetes/minikube/issues/9811#issuecomment-748465051

medyagh (commented):
example of mocking unit test https://github.com/medyagh/minikube/blob/199ea3425b8343e533d8d1982ad750a8dab7d59e/pkg/drivers/kic/oci/info_test.go#L23

pkg/drivers/kic/oci/network_create_test.go
```diff
@@ -0,0 +1,84 @@
...
+var dockerResponse string
+var dockerInspectGetterMock = func(name string) (*RunResult, error) {
+	var responseInBytes bytes.Buffer
+	responseInBytes.WriteString(dockerResponse)
+	response := &RunResult{Stdout: responseInBytes}
+
+	return response, nil
+}
+
+func TestDockerInspectWithMTU(t *testing.T) {
```
medyagh, line 34: how about putting this in a for loop that goes over the test cases ? feel free to say no if u think this is better, but here is an example
https://github.com/medyagh/minikube/blob/199ea3425b8343e533d8d1982ad750a8dab7d59e/cmd/minikube/cmd/status_test.go#L68

so maybe we could put both of these tests in one test and have a for loop of test cases?

medyagh, line 34: rename the func

medyagh (changes requested):

medyagh (approved):

medyagh (approved):

medyagh:
thank you for fixing this and adding unit tests for it too

Issue #9992, opened 2020-12-17
MEP: Change from custom Buildroot ISO to standard Kubernetes OS


afbjorklund:
Currently we are using a custom image, which is built using the [Buildroot](https://buildroot.org/) tool...

It is a replacement for the original Boot2Docker image, based on Tiny Core Linux.
(The board config also draws some inspiration from the original CoreOS images)

For the KIC base image, we have instead changed to using a standard `ubuntu:20.04`

We _could_ do the same for the ISO image, base it on a standard Linux distribution.
(Preferrably one of the official ones, that are supported **and tested** by `kubeadm`)

 https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#before-you-begin

>    Ubuntu 16.04+
>    Debian 9+
>    CentOS 7
>    Red Hat Enterprise Linux (RHEL) 7
>    Fedora 25+
>    HypriotOS v1.0.1+
>    Flatcar Container Linux (tested with 2512.3.0)

Something like:
*    Ubuntu 16.04
*    Ubuntu 18.04
*    **Ubuntu 20.04**
*    CentOS 7
*    CentOS 8
*    Fedora 32

Comparison table:
| Year      | Debian    | Ubuntu    | CentOS    | Fedora    |
| ----      | ------    | ------    | ------    | ------    |
| 2010      | 6         | 10.04     | 6         | 12        |
| 2012      | 7         | 12.04     |           | 15        |
| 2014      | 8         | 14.04     | 7         | 18        |
| 2016      | 9         | 16.04     |           | 24        |
| 2018      | 10        | 18.04     | 8         | 28        |
| 2020      |           | 20.04     |           | 32        |

It's easier to use Ubuntu and CentOS (than Debian and RHEL), due to licensing issues etc.

Suggestion to use the _same_ OS version as the KIC base image, which is Ubuntu 20.04 LTS.

----

It only needs the kernel capability to run containers, the container runtime can be provisioned.

Using a package manager makes it easier to add/remove software, compared to rebuilding.

When using one of cloud drivers for `docker-machine`, it uses an Ubuntu 16.04 image.
You can see how it works, by checking out the minikube "generic" driver (with any VM).

https://docs.docker.com/machine/get-started-cloud/

https://github.com/kubernetes/minikube/pull/9545#issuecomment-718855132

The package list could be the _same_ as the docker image, to start with (minus kind-specific)
[...]

Issue #9989, opened 2020-12-17
Separate container runtimes from the image


afbjorklund:
Currently we include all **three** CRI runtimes, in the base image:
* Docker (docker)
* CRI-O (crio + podman)
* Containerd (containerd + buildkit)

This makes the image somewhat big, compared to having just one.

### ISO
![graph-size](https://user-images.githubusercontent.com/10364051/102545530-a916c780-40b6-11eb-984e-423427c8db3d.png)

### KIC
![installed-size](https://user-images.githubusercontent.com/10364051/103348940-c4f97080-4a9b-11eb-96ff-a034adbd55d6.png)

We only _start_ one of them, but the others still takes up space on the disk.

(actually we _always_ start docker, as long as using the "docker machine" API...
but we stop it again quickly afterwards, if using the containerd or cri-o runtimes)

----

One alternative to doing three "base" images, would be to use the "provisioner".
This is a docker-machine feature, that is used to install docker on the machine.

https://github.com/docker/machine/tree/master/libmachine/provision

So the base image wouldn't have _any_ of them (maybe `runc`) from the start.
Instead they would be downloaded and installed from separate tarballs/packages.

This is how we currently handle different kubernetes versions, for instance.

* linux/v1.20.0/kubeadm
* linux/v1.20.0/kubectl
* linux/v1.20.0/kubelet

It's also how we handle the preload, instead of making three "node" images:

* preloaded-images-k8s-v8-v1.20.0-containerd-overlay2-amd64.tar.lz4
* preloaded-images-k8s-v8-v1.20.0-cri-o-overlay-amd64.tar.lz4
* preloaded-images-k8s-v8-v1.20.0-docker-overlay2-amd64.tar.lz4

As compared with kind:

* `docker.io/kindest/node:v1.20.0` (always uses containerd-overlay2)

afbjorklund:
We would need an installation script for CRI-O, similar to [get.docker.com](https://get.docker.com) for Docker...

`if ! type docker; then curl -sSL https://get.docker.com | sh -; fi`

See https://github.com/cri-o/cri-o/issues/4343

We also need one for containerd (and buildkit), but maybe we can do it ourselves ?

afbjorklund:
For containerd, it will probably use the binaries from the github releases:

https://github.com/containerd/containerd/releases/download/v1.4.3/containerd-1.4.3-linux-amd64.tar.gz

[...]

PR #10067, opened 2020-12-30
MEP: Standard Linux Distribution ("distro")

afbjorklund:
See #9992

### DIFF
--- a/enhancements/proposed/20201230-standard-distro/standard-distro.md
+++ b/enhancements/proposed/20201230-standard-distro/standard-distro.md
@@ -1,0 +1,41 @@
+# Standard Linux Distribution
+
+* First proposed: 2020-12-17
+* Authors: Anders F Björklund (@afbjorklund)
+
+## Reviewer Priorities
+
+Please review this proposal with the following priorities:
+
+*   Does this fit with minikube's [principles](https://minikube.sigs.k8s.io/docs/concepts/principles/)?
+*   Are there other approaches to consider?
+*   Could the implementation be made simpler?
+*   Are there usability, reliability, or technical debt concerns?
+
+Please leave the above text in your proposal as instructions to the reader.
+
+## Summary
+
+Change the distribution (OS) for the minikube ISO, from Buildroot to Ubuntu.
+
+## Goals
+
+* Use one of the supported Kubernetes OS, like Ubuntu 20.04
+* Use the same operating system for KIC base and ISO image
+
+## Non-Goals
+
+* Making major changes to the new standard operating system
+* Support production deployments, still intended for learning
+
+## Design Details
+
+Use external system image and external packages, same as for KIC image.
+
+Keep both images available (one being default), during transition period.
+
+## Alternatives Considered
+
+Continue to support custom distro, instead of using a standard distro.
+
+Make current Buildroot OS into standard supported Kubernetes distribution.


medyagh (commented):
if you think this the right way to do, we could approve this and work on it for Q1 2021.
should we have a transition period where people who really love our Buildroot ISO keep using it ?

enhancements/proposed/20201230-standard-distro/standard-distro.md
```diff
@@ -0,0 +1,37 @@
...
+* Making major changes to the new standard operating system
+* Support production deployments, still intended for learning
+
+## Design Details
+
+Use external system image and external packages, same as for KIC image.
+
+## Alternatives Considered
+
+Continue to support custom distro, instead of using a standard distro.
```
medyagh, line 37: should add consider making our buildroot ISO a separate project on its own?

medyagh (changes requested):

afbjorklund:
>  should we have a transition period where people who really love our Buildroot ISO keep using it ?

Indeed, the plan was to have multiple ISO files (one being the default) - and use URL to select the other

enhancements/proposed/20201230-standard-distro/standard-distro.md
afbjorklund: Yes, good suggestion! (add that option, explicitly)

tstromberg (approved):

priyawadhwa (approved):

medyagh (changes requested):
one i think  I liket his MEP to conver is the upgrade plan,

it seems to me the previous versions of minikube will not be able to upgrade to newer verison of minikube if we change the ISO's OS.
and they would need to do selete and start a new one.

if that this the case, we need to discuss it and possibliy roll out minikube 2 which will not support upgrading old versions.

this would mean minikube would have to add a mechanisim to detect old iso and and warn user to delete and start again.

afbjorklund:
We have changed iso before, when we went from boot2docker to buildroot...
As long as the files on the /var disk image are compatible, it should be doable.

It's also not a blocker, even if it _should_ require users to delete their cluster ?
I think we need a prototype, before we can tell if an upgrade is posssible or not

tstromberg:
Agreed. In the past it hasn't been necessary to ask users to delete their
old VM's. Given how few entry points minikube actually has into the
built-in userland, I don't foresee compatibility issues.

I can see upgrades potentially being something to work out, but outside of
the scope of this MEP.

minikube already handles Ubuntu (docker) and buildroot (ISO) with minimal
behavioral differences.

afbjorklund:
Before we look into making it a "general" distribution, we need something for the upgrade path.
Not saying that this is the only way, but it was _one_ of the options (keep the distro, but external)

I was going to look into A/B booting, like this: https://source.android.com/devices/tech/ota/ab
Basically you have **two** installations, one active and one passive (where you do the upgrade)

Alternative to using a package manager.

----

This would be similar to minikube rebooting with a new ISO, but slightly more hard disk-centric.
Both require that you also have a third image, where you keep the data that you want to keep.

I also had a fourth* image, that was used for pulling the images from (like the preload tarball)
This allows upgrading the container storage, if needed. But mostly used for the _first_ installation.

And a fifth image for `/data`, but anyway.

afbjorklund:
> Agreed. In the past it hasn't been necessary to ask users to delete their
> old VM's. Given how few entry points minikube actually has into the
> built-in userland, I don't foresee compatibility issues.

I also wanted to make some drastic _changes_ to the userland, like excluding the runtimes.

https://github.com/kubernetes/minikube/issues/9989

PR #10072, opened 2020-12-31
Fix test code not to create unsued config file

### LINKED ISSUE #8614: `make test` leaves behind unused config file: ~/.minikube/profiles/config.json

anencore94:
- move testdata in 'testdata/.minikube2' to 'testdata/minikube'
  - I guess there is no reason to split testdata as .minikube and .minikube2
  - since [localpath.MiniPath()](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/localpath/localpath.go#L42) checks whether the path ends with ".minikube" or not, when env(MinikubeHome) ends with ".minikube2", the output will become such as ".../.minikube2/.minikube"
    - which leads the test code complex
- fix not to create unsed config file at local 'env(MinikubeHome)' path
  - from now on, test code save temporal config files at testdata folder and
    delete those when test case finished

fixes #8614 and it was extended from my past pr #8991


tstromberg (approved):

tstromberg:
Thank you!

PR #10078, opened 2021-01-01
Removing installation details using cask . 

Aut0R3V:
continue to #8196
Since cask was removed over a year ago with minikube v1.5.0, the documentation might not need this specific portion.

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -81,13 +81,6 @@ If the [Brew Package Manager](https://brew.sh/) installed:
 brew install minikube
 ```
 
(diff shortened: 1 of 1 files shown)


Aut0R3V:
@tstromberg @priyawadhwa

priyawadhwa:
Hey @Aut0R3V I don't remember cask being removed, but I see that in our docs for releasing we seem to be releasing to Brew cask:

https://minikube.sigs.k8s.io/docs/contrib/releasing/binaries/#package-managers-which-include-minikube

Is there a PR where that was removed?

Aut0R3V:
This must be a misunderstanding. Apologies

Issue #9583, opened 2020-10-29
Fork the machine driver "generic" to minikube, prepare for renaming it


afbjorklund:
Previously we have forked the docker-machine (libmachine) driver called "none":

https://github.com/docker/machine/blob/master/drivers/none/driver.go

https://github.com/kubernetes/minikube/blob/master/pkg/drivers/none/none.go

We want to do the same thing for the driver called "generic", as used in PR #9545

https://github.com/docker/machine/blob/master/drivers/generic/generic.go

https://github.com/kubernetes/minikube/blob/master/pkg/drivers/generic/generic.go

Once it is in our "drivers" code base, we can rename it as see fit...

Possibly to something like "remote", to be decided (see #7772)

medyagh:
sounds good !

afbjorklund:
We might end up moving the other drivers as well, it's mostly cloud drivers so the only ones we use are "fake" and "err"...

amazonec2
azure
digitalocean
driverutil
**errdriver**
exoscale
**fakedriver**
**generic**
google
**hyperv**
**none**
openstack
rackspace
softlayer
**virtualbox**
vmwarefusion
vmwarevcloudair
vmwarevsphere

And of course the "virtualbox" driver. The "hyperv" is already modified, and "hyperkit" and "kvm" are external to libmachine.

```
 drivers/hyperv/hyperv.go     | 71 ++++++++++++++++++++++++++++++++++++++++++++---------------------------
 drivers/hyperv/powershell.go |  4 ++--
 2 files changed, 46 insertions(+), 29 deletions(-)
```

"driverutil" only has one function and "generic_test" only has one test.

```go
package driverutil

func SplitPortProto(raw string) (port string, protocol string)
```

```
=== RUN   TestSetConfigFromFlags
--- PASS: TestSetConfigFromFlags (0.00s)
```

Not needed.

afbjorklund:
We probably want to keep the external API unless we have a **good** reason to change it.

This in order to keep it compatible with e.g. "vmware" and "parallels" external drivers still

afbjorklund:
We should remove the "vmwarefusion" driver, it is deprecated in favor of "vmware".

We have already removed the "xhyve" (for hyperkit) and "kvm" (for kvm2) drivers.

afbjorklund:
The best would be to move _all_ the extra drivers out of the main minikube code base:

* hyperkit
* kvm(2)

And to some structure that we control, separate from machine and machine-drivers:

[...]

PR #10099, opened 2021-01-06
Add new driver "SSH" to bootstrap generic minkube clusters over ssh

### LINKED ISSUE #4733: Support the "generic" driver, for user provided VMs

afbjorklund:
New year, new attempt of the generic driver...

Rebased version of ~#9545~ ~#6873~ ~#4734~

Closes #4733 #9583


afbjorklund:
@medyagh : I forked the "generic" driver and changed the start/stop implementation, so now it looks more like the "none" driver.

This also means that both drivers are now free to be renamed or aliased or whatever, since they are no longer using libmachine.

pkg/drivers/generic/generic.go
```diff
@@ -0,0 +1,230 @@
...
+// Stop a host gracefully, including any containers that we are managing.
+func (d *Driver) Stop() error {
+	exec := command.NewSSHRunner(d)
+	if err := sysinit.New(exec).Stop("kubelet"); err != nil {
+		klog.Warningf("couldn't stop kubelet. will continue with stop anyways: %v", err)
+		if err := sysinit.New(exec).ForceStop("kubelet"); err != nil {
+			klog.Warningf("couldn't force stop kubelet. will continue with stop anyways: %v", err)
+		}
+	}
+	/* TODO
```
medyagh, line 149: either remove TODO or add a github issue link here

pkg/drivers/generic/generic.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// Kill stops a host forcefully, including any containers that we are managing.
+func (d *Driver) Kill() error {
+	exec := command.NewSSHRunner(d)
+	if err := sysinit.New(exec).ForceStop("kubelet"); err != nil {
+		klog.Warningf("couldn't force stop kubelet. will continue with kill anyways: %v", err)
+	}
+
+	/* TODO
```
medyagh, line 177: same here

cmd/minikube/cmd/start_flags.go
```diff
@@ -221,6 +227,12 @@ func initNetworkingFlags() {
 	startCmd.Flags().String(serviceCIDR, constants.DefaultServiceCIDR, "The CIDR to be used for service cluster IPs.")
 	startCmd.Flags().StringArrayVar(&config.DockerEnv, "docker-env", nil, "Environment variables to pass to the Docker daemon. (format: key=value)")
 	startCmd.Flags().StringArrayVar(&config.DockerOpt, "docker-opt", nil, "Specify arbitrary flags to pass to the Docker daemon. (format: key=value)")
+
+	// generic
+	startCmd.Flags().String(genericIPAddress, "", "IP address (generic)")
```
medyagh, line 232: i think given this is a new driver, it is better to provide more explicit help text

"IP address (generic) only used by generic driver"

cmd/minikube/cmd/start_flags.go
```diff
@@ -221,6 +227,12 @@ func initNetworkingFlags() {
...
+	startCmd.Flags().String(genericSSHUser, defaultSSHUser, "SSH user (generic)")
```
medyagh, line 233: lets add a Validation for these flags, so if users passes them to other drivers there should be exit.Usage
"generic-ssh-user: can only be used by generic driver...."

medyagh (changes requested):
mostly looks good just a couple knits.
btw do u mind pasting an example in the output I am curious how it looks for minikube start

cmd/minikube/cmd/start_flags.go
medyagh: also a doc page in the website would be nice.

afbjorklund: The other ones say "xxx driver only", so will go with that

afbjorklund:
Needs support for killing all containers. Probably better off as a function, instead of copy/paste (like in "none")

Also the restartKubelet could use a shared "init" function, instead of hard-coding systemd like it is doing now.

cmd/minikube/cmd/start_flags.go
afbjorklund: The other ones say "xxx driver only", so will go with that

afbjorklund:
@medyagh : the provisioner is currently broken in minikube, so doesn't really work after forking it:

 ```diff
-func (provisioner *UbuntuSystemdProvisioner) Provision(swarmOptions swarm.Options, authOptions auth.Options, engineOptions engine.Options) error {
-	provisioner.SwarmOptions = swarmOptions
-	provisioner.AuthOptions = authOptions
-	provisioner.EngineOptions = engineOptions
-	swarmOptions.Env = engineOptions.Env
+// Provision does the provisioning
+func (p *UbuntuProvisioner) Provision(swarmOptions swarm.Options, authOptions auth.Options, engineOptions engine.Options) error {
+	p.SwarmOptions = swarmOptions
+	p.AuthOptions = authOptions
+	p.EngineOptions = engineOptions
 
-	storageDriver, err := decideStorageDriver(provisioner, "overlay2", engineOptions.StorageDriver)
-	if err != nil {
+	klog.Infof("provisioning hostname %q", p.Driver.GetMachineName())
+	if err := p.SetHostname(p.Driver.GetMachineName()); err != nil {
 		return err
 	}
-	provisioner.EngineOptions.StorageDriver = storageDriver
 
-	log.Debug("setting hostname")
-	if err := provisioner.SetHostname(provisioner.Driver.GetMachineName()); err != nil {
-		return err
-	}
+	p.AuthOptions = setRemoteAuthOptions(p)
+	klog.Infof("set auth options %+v", p.AuthOptions)
 
-	log.Debug("installing base packages")
-	for _, pkg := range provisioner.Packages {
-		if err := provisioner.Package(pkg, pkgaction.Install); err != nil {
-			return err
+	klog.Infof("setting up certificates")
+	configAuth := func() error {
+		if err := configureAuth(p); err != nil {
+			klog.Warningf("configureAuth failed: %v", err)
+			return &retry.RetriableError{Err: err}
 		}
+		return nil
 	}
 
-	log.Info("Installing Docker...")
-	if err := installDockerGeneric(provisioner, engineOptions.InstallURL); err != nil {
-		return err
-	}
-
-	log.Debug("waiting for docker daemon")
-	if err := mcnutils.WaitFor(provisioner.dockerDaemonResponding); err != nil {
-		return err
-	}
+	err := retry.Expo(configAuth, 100*time.Microsecond, 2*time.Minute)
 
-	provisioner.AuthOptions = setRemoteAuthOptions(provisioner)
-
-	log.Debug("configuring auth")
-	if err := ConfigureAuth(provisioner); err != nil {
+	if err != nil {
+		klog.Infof("Error configuring auth during provisioning %v", err)
 		return err
 	}

```

The actual docker installation has been removed, so I'm not sure what it is actually provisioning now ?

```
🔥  Creating generic bare metal machine (CPUs=2, Memory=6000MB, Disk=20000MB) ...
🔥  Deleting "minikube" in generic ...
🤦  StartHost failed, but will try again: usermod: ssh command error:
command : sudo usermod -aG docker vagrant
err     : Process exited with status 6
output  : usermod: group 'docker' does not exist

🔥  Creating generic bare metal machine (CPUs=2, Memory=6000MB, Disk=20000MB) ...| 😿  Failed to start generic bare metal machine. Running "minikube delete" may fix it: usermod: ssh command error:
command : sudo usermod -aG docker vagrant
err     : Process exited with status 6
output  : usermod: group 'docker' does not exist



❌  Exiting due to GUEST_PROVISION: Failed to start host: usermod: ssh command error:
command : sudo usermod -aG docker vagrant
err     : Process exited with status 6
output  : usermod: group 'docker' does not exist


```

**Comment by afbjorklund** [comment]
Also the system information is totally bogus, but I had to revert that from the previous PR.

```
🔥  Creating generic bare metal machine (CPUs=2, Memory=6000MB, Disk=20000MB```
```

It is supposed to read the existing values (like with none driver), instead of showing VM defaults.

See 41979352b927b85e12574cb3bbbf4917ef1a06a0

**Comment by afbjorklund** [comment]
Cherry-picked and updated the old code, so now it shows proper info:

```
🔗  Running remotely (CPUs=2, Memory=1992MB, Disk=9862MB) ...
```

**Comment by tstromberg** [comment]
What would you think about an alternate name for this, such as "ssh" or "remote"? 

`generic` feels difficult to reason about.

**Comment by afbjorklund** [comment]
> What would you think about an alternate name for this, such as "ssh" or "remote"?
> 
> `generic` feels difficult to reason about.

The fork (from libmachine) is now completed, so the drivers can be renamed.

We had originally planned to do it as a separate feature, though: #7772

The boring names were local and remote, but there were also some more creative suggestions

Like "bare metal" or "cloud" and also "native". But not stuck with "none" and "generic" anymore...

**Comment by afbjorklund** [comment]
Think we will leave the provisioner as-is, and leave that to the other feature: #9989

But right now it works _differently_ for ubuntu and for any other distributions.

So we might want to add some specific KIC-check, to the "ubuntu" provisioner ?

```go
func (p *UbuntuProvisioner) String() string {
        return "ubuntu"
}

// CompatibleWithHost checks if provisioner is compatible with host
func (p *UbuntuProvisioner) CompatibleWithHost() bool {
        return p.OsReleaseInfo.ID == "ubuntu"
}
```

medyagh:
> What would you think about an alternate name for this, such as "ssh" or "remote"?
> 
> `generic` feels difficult to reason about.

I would vote for SSH name since everything in the flags points to SSH. 
wdyt @afbjorklund to change the name to SSH driver?

afbjorklund:
> > What would you think about an alternate name for this, such as "ssh" or "remote"?
> > `generic` feels difficult to reason about.
> 
> I would vote for SSH name since everything in the flags points to SSH.
> wdyt @afbjorklund to change the name to SSH driver?

I thought _all_ drivers were using ssh to talk to the machine, even docker ?
Well, maybe with the exception of the none driver that uses regular exec.

But we can go with ssh...

medyagh:
@afbjorklund there seem to be a build error need to be fixed

afbjorklund:
@medyagh : yeah, happened in 099c6b72a64ec892ed4cdd72744100a2a86393f2

cmd/minikube/cmd/start.go
```diff
@@ -851,7 +851,7 @@ func validateUser(drvName string) {
 
 // memoryLimits returns the amount of memory allocated to the system and hypervisor, the return value is in MiB
 func memoryLimits(drvName string) (int, int, error) {
-	info, cpuErr, memErr, diskErr := machine.CachedHostInfo()
+	info, cpuErr, memErr, diskErr := machine.LocalHostInfo()
```
medyagh, line 854: any reason this was changed for all drivers ? previously it used a cache, this might add a small overhead on "docker driver" which needs invoking docker (that could cost 2-3 seconds on slow dockers)

medyagh (changes requested):

cmd/minikube/cmd/start.go
afbjorklund: It _still_ uses the cache, I just renamed the function. If needed, we could keep the "CachedHostInfo" and give it a parameter ?

```go
// LocalHostInfo returns system information such as memory,CPU, DiskSize
func LocalHostInfo() (*HostInfo, error, error, error)
```

```go
// RemoteHostInfo returns system information such as memory,CPU, DiskSize
func RemoteHostInfo(r command.Runner) (*HostInfo, error, error, error)
```

This is how it is currently being used, but there is a timing difference (it needs SSH to be up first, to get the info)

cmd/minikube/cmd/start_flags.go
```diff
@@ -110,6 +110,12 @@ const (
 	network                 = "network"
 	startNamespace          = "namespace"
 	trace                   = "trace"
+	sshIPAddress            = "ip-address"
```
medyagh, line 113: how do u feel about `ssh-ip-address`  or `ssh-remote-ip`  ? so all the SSH driver flags start with SSH ? 
I think, it would get lost in the tons of flags we have, if a user is looking only for ssh driver flags...

pkg/drivers/ssh/ssh.go
```diff
@@ -0,0 +1,243 @@
...
+	return d.SSHKeyPath
+}
+
+func (d *Driver) PreCreateCheck() error {
+	if d.SSHKey != "" {
+		if _, err := os.Stat(d.SSHKey); os.IsNotExist(err) {
+			return fmt.Errorf("SSH key does not exist: %q", d.SSHKey)
+		}
+
+		// TODO: validate the key is a valid key
```
medyagh, line 106: Either add issue number or remove TODO

pkg/minikube/config/types.go
```diff
@@ -66,6 +66,10 @@ type ClusterConfig struct {
 	HostDNSResolver         bool   // Only used by virtualbox
 	HostOnlyNicType         string // Only used by virtualbox
 	NatNicType              string // Only used by virtualbox
+	IPAddress               string // Only used by ssh
```
medyagh, line 69: I think this should be called SSHIPAddress or SSHRemoteIP
 for clarity, so it doesn't get mixed with node IP or master IP...

medyagh (changes requested):

pkg/drivers/ssh/ssh.go
afbjorklund: That was just copied from docker-machine

cmd/minikube/cmd/start_flags.go
afbjorklund: Sure, it got confusing when removed the driver name (generic)

afbjorklund:
Things got a bit hard to read, when the machine driver was renamed.

i.e sshIPAddress, sshSSHUser, sshSSHKey, sshSSHPort

Easy to confuse the name of the driver (ssh) with the protocol (ssh)

Normally the IP of the VM is the same for _all_ uses, though

PR #10100, opened 2021-01-06
Add support for systemd cgroup to containerd runtime

priyawadhwa:
The `--force-systemd` flag will now work with the containerd runtime.

Tested on Cloud Shell, started in ~45 seconds, comparable to other runtimes.


medyagh (approved):
thanks for fixing this

Issue #9939, opened 2020-12-12
make perfomrance pr-bot to use json output


medyagh:
for more information ask @priyawadhwa

Aut0R3V:
@priyawadhwa can you please explain about this issue

priyawadhwa:
Hey @Aut0R3V -- we have a binary called `mkcmp` in our repo -- code can be found here:

1. https://github.com/kubernetes/minikube/blob/master/cmd/performance/mkcmp/cmd/mkcmp.go
1. https://github.com/kubernetes/minikube/tree/master/pkg/minikube/perf

mkcmp is responsible for comparting start time between two minikube binaries. It runs each binary three times and outputs information around how long it took. This output is pasted on our PRs by our PR bot ([example](https://github.com/kubernetes/minikube/pull/9538#issuecomment-715573428))

You can build the binary locally by running:

```
make out/mkcmp
```

Right now, we split up how long `minikube start` takes by log, the code is in this file: rnetes/minikube/blob/master/pkg/minikube/perf/logs.go

Instead of timing by logs to STDOUT, it would be better if we used the JSON output in minikube to more accurately parse logs and match up how long different steps take.

This would require running `minikube start` in `mkcmp` with the `--output=json` flag and parsing the output correctly.

Aut0R3V:
Thanks a lot for the detailed explanation!

Aut0R3V:
\assign

priyawadhwa:
@Aut0R3V thanks for taking this issue on!

Aut0R3V:
@priyawadhwa I had some questions :

1] What changes are to be added at the already existing mkcmp workflow ? 
[https://github.com/kubernetes/minikube/blob/master/cmd/performance/mkcmp/cmd/mkcmp.go]

2] Is this the correct defaultTimeFormat for the output :
type Status struct {
	defaultStatusFormat          = `host: {{.Host}}
kubelet: {{.Kubelet}}
apiserver: {{.APIServer}}
kubeconfig: {{.Kubeconfig}}
`
)
If not then
What would be the defaultTimeFormat for the output ?

priyawadhwa:
Hey @Aut0R3V --

1. The code that needs to be changed is in this function:
https://github.com/kubernetes/minikube/blob/fdab252f4f1d3aaeea028f90a2153cb1da8e54cb/pkg/minikube/perf/logs.go#L33

[...]

PR #10105, opened 2021-01-07
JSON output for pr-bot

Aut0R3V:
#9939

### DIFF
--- a/pkg/minikube/perf/logs.go
+++ b/pkg/minikube/perf/logs.go
@@ -22,7 +22,8 @@ import (
 	"os/exec"
 	"strings"
 	"time"
-
+	"encoding/json"
+	"fmt"
 	"github.com/pkg/errors"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


Aut0R3V (commented):
Made some initial changes

priyawadhwa (commented):
Hey @Aut0R3V thanks for looking into this. All you need to do to ensure JSON output is append `--output=json` to `minikube start`:

https://github.com/kubernetes/minikube/blob/55f3a77d6f85b79ced3a5fdca99314a46d165d74/pkg/minikube/perf/start.go#L97

In `logs.go`, we need to make sure we are parsing the JSON output correctly. Right now, we set `log=<entire line of text output> here:

https://github.com/kubernetes/minikube/blob/55f3a77d6f85b79ced3a5fdca99314a46d165d74/pkg/minikube/perf/logs.go#L54

and, instead, we want to parse the JSON output and set
`log=<name of JSON step>`

pkg/minikube/perf/logs.go
```diff
@@ -34,6 +35,11 @@ func timeCommandLogs(cmd *exec.Cmd) (*result, error) {
 	// matches each log with the amount of time spent on that log
 	r := newResult()
 
+	output = strings.ToLower(output)
```
priyawadhwa, line 38: We have already set `--output=json` in `start.go`, so we can remove all of this logic. Instead, we now want to parse the JSON output.

In `logs.go`, we need to make sure we are parsing the JSON output correctly. The only thing that needs to happen in this file is to this `log` variable:

https://github.com/kubernetes/minikube/blob/55f3a77d6f85b79ced3a5fdca99314a46d165d74/pkg/minikube/perf/logs.go#L54
 

and parse out the `Name` of the step, since `log` will be in JSON format.

priyawadhwa (changes requested):

priyawadhwa:
Hey @Aut0R3V this PR has been quiet for a few weeks -- are you still working on it?

Issue #10041, opened 2020-12-24
implement operation auditing for minkube (--user) flag


medyagh:
there should be a way to know what user did what operition on a minikube cluster

(to be used by multiple tools that use minikube at the same time
skaffold, cloud code, cloud run, cloud shel ....)

```
----------------------------------------
command, option, user , start_time , end_time
----------------------------------------
start, --driver=docker , driver, cloud-code, 223232,2 23232
start, --driver=hyperkit , driver, code-shell, 223232,2 23232
stop, steven, 223232,2 23232

```

if no user was provided, it should default to the OS user name

priyawadhwa:
Commands we will track: start, stop, pause, unpause, delete
Command we won't track because they may clutter the log: status, version

Options for deleting the audit file:
1. consistently append to a single file and assume that it will be deleted whenever a user runs `--purge`
2. put a TTL on the audit file (if it hasn't been modified in a week, delete it and start a new one)


Whenever we append to the file, we can also log what we are appending to STDERR which will satisfy https://github.com/kubernetes/minikube/issues/9520

PR #10106, opened 2021-01-07
Add new flag --user and to log executed commands

spowelljr:
Implements #10041 

Example log:
```
% cat .minikube/logs/audit.json
{"data":{"args":"--user user1","command":"start","endTime":"Thu, 14 Jan 2021 16:21:04 MST","profile":"minikube","startTime":"Thu, 14 Jan 2021 16:20:18 MST","user":"user1"},"datacontenttype":"application/json","id":"042ad16a-79fe-461d-aec7-375c413c52f8","source":"https://minikube.sigs.k8s.io/","specversion":"1.0","type":"io.k8s.sigs.minikube.audit"}
{"data":{"args":"--user user2","command":"delete","endTime":"Thu, 14 Jan 2021 16:21:27 MST","profile":"minikube,"startTime":"Thu, 14 Jan 2021 16:21:24 MST","user":"user2"},"datacontenttype":"application/json","id":"5b88b158-b9c4-4a25-b3db-fde1eb81563c","source":"https://minikube.sigs.k8s.io/","specversion":"1.0","type":"io.k8s.sigs.minikube.audit"}
```
 Pretty-print of entry
```
{
  "data": {
    "args": "--user test123 --alsologtostderr",
    "command": "start",
    "endTime": "Thu, 14 Jan 2021 16:21:04 MST",
    "profile": "minikube",
    "startTime": "Thu, 14 Jan 2021 16:20:18 MST",
    "user": "testUser"
  },
  "datacontenttype": "application/json",
  "id": "042ad16a-79fe-461d-aec7-375c413c52f8",
  "source": "https://minikube.sigs.k8s.io/",
  "specversion": "1.0",
  "type": "io.k8s.sigs.minikube.audit"
}
```


cmd/minikube/cmd/start_flags.go
```diff
@@ -110,6 +110,7 @@ const (
 	network                 = "network"
 	startNamespace          = "namespace"
 	trace                   = "trace"
+	userFlag                = "user"
```
medyagh, line 113: if this is a global flag, it doesnt need to be in start.go maybe in root.go ?

cmd/minikube/cmd/audit.go
```diff
@@ -0,0 +1,55 @@
...
+	"strings"
+	"time"
+
+	"github.com/spf13/viper"
+	"k8s.io/minikube/pkg/minikube/localpath"
+	"k8s.io/minikube/pkg/minikube/out/register"
+)
+
+// getUser pulls the user flag, if empty gets the os user
+func getUser() string {
```
medyagh, line 31: avoid using get in naming in golang, https://golang.org/doc/effective_go.html#Getters

maybe could be auditUser() or antyhing else

cmd/minikube/cmd/audit.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	osUser, err := user.Current()
+	if err != nil {
+		return "unable to get user"
+	}
+	return osUser.Username
+}
+
+// getArgs concats the args into space delimited string
+func getArgs() string {
```
medyagh, line 44: same here... https://golang.org/doc/effective_go.html#Getters

cmd/minikube/cmd/audit.go
```diff
@@ -0,0 +1,55 @@
...
+
+// getUser pulls the user flag, if empty gets the os user
+func getUser() string {
+	u := viper.GetString(userFlag)
+	if u != "" {
+		return u
+	}
+	osUser, err := user.Current()
+	if err != nil {
+		return "unable to get user"
```
medyagh, line 38: if there is error to get the username we should return a default username such as. "UKNOWN"

medyagh (changes requested):

site/content/en/docs/commands/start.md
```diff
@@ -89,6 +89,7 @@ minikube start [flags]
       --registry-mirror strings           Registry mirrors to pass to the Docker daemon
       --service-cluster-ip-range string   The CIDR to be used for service cluster IPs. (default "10.96.0.0/12")
       --trace string                      Send trace events. Options include: [gcp]
+      --user string                       Sets who will be logged as executing the command, will use os user if none provided
```
medyagh, line 92: lets make a better description.
how about, 

"specify the user doing an operation. useful for auditing operations done by 3rd party tools.  defaults to the operating system username"

medyagh (changes requested):

cmd/minikube/cmd/start_flags.go
spowelljr: Moved

cmd/minikube/cmd/audit.go
spowelljr: Updated

cmd/minikube/cmd/audit.go
spowelljr: Updated

cmd/minikube/cmd/audit.go
spowelljr: Updated

site/content/en/docs/commands/start.md
spowelljr: Updated

pkg/minikube/out/register/json.go
```diff
@@ -69,3 +71,9 @@ func PrintWarning(warning string) {
 	w := NewWarning(warning)
 	printAndRecordCloudEvent(w, w.data)
 }
+
+// RecordAudit records an Audit type in JSON format
+func RecordAudit(command string, args string, user string, startTime time.Time, endTime time.Time) {
+	a := NewAudit(command, args, user, startTime, endTime)
+	recordCloudEvent(a, a.data)
```
medyagh, line 78: how about making RecordCloudEvent public and then moving RecordAudit to audit package.
so that way all the audit related thigns will be in the audit package
and this package is only scoped down to doing one thing, "registering cloud event"

pkg/minikube/out/register/log.go
```diff
@@ -154,3 +155,26 @@ func NewErrorExitCode(err string, exitcode int, additionalData ...map[string]str
 func (s *Error) Type() string {
 	return "io.k8s.sigs.minikube.error"
 }
+
+// Audit represents the execution of a command
+type Audit struct {
```
medyagh, line 160: we can move these to audit package and it be called something like

Entry

so it would ready as
audit.Entry if it is referenced outside ( most porbaby would not need to be public though)

cmd/minikube/cmd/root.go
```diff
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"time"
```
medyagh, line 26: lets add integration test for this, we could add a subtest to "TestJSONOutput" that will check if the audit step is logged.

https://github.com/medyagh/minikube/blob/dce38fb331b6172c0c618ad486ac1b60661af687/test/integration/json_output_test.go#L55

medyagh (changes requested):

cmd/minikube/cmd/root.go
spowelljr: `audit.Log` doesn't output the event to stdout. We talked about how that might possibly interfere with 3rd party tools that are using the output to keep track of steps, so I excluded it from the json output. Would it make more sense to output it here or leave it as is?

medyagh: good point! we can do integration test by checking the existence of audit file for a specific word (such as the os username)

pkg/minikube/audit/audit.go
```diff
@@ -0,0 +1,57 @@
...
+	"time"
+
+	"github.com/spf13/viper"
+	"k8s.io/minikube/pkg/minikube/config"
+	"k8s.io/minikube/pkg/minikube/localpath"
+	"k8s.io/minikube/pkg/minikube/out/register"
+)
+
+// auditUser pulls the user flag, if empty gets the os user
+func auditUser() string {
```
medyagh, line 32: since we are in auditPackage, this name would stutter. aduit.aduitUser... we could just call it, user()

pkg/minikube/audit/audit.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	osUser, err := user.Current()
+	if err != nil {
+		return "UNKNOWN"
+	}
+	return osUser.Username
+}
+
+// auditArgs concats the args into space delimited string
+func auditArgs() string {
```
medyagh, line 45: same naming for here. args()

+ I am curious how this would work on "minikube kubectl command" ?

do u mind sharing the audit log for this "minikube kubectl --get pods -A"

medyagh (changes requested):

medyagh:
additionally, I am curious does this design allow us to ban one of the commands from auditing ? 
for example  in some 3rd party tool they call "minikube status" every second, this could result in recording a large audit footprint just for status.

maybe we could either disable it for not intrusive operations such as "status" or "version"

pkg/minikube/audit/audit.go
spowelljr:
```
{
  "data": {
    "args": "-- get pods -A",
    "command": "kubectl",
    "endTime": "2021-01-08 11:14:40.336806 -0700 MST m=+3.304018657",
    "startTime": "2021-01-08 11:14:37.069239 -0700 MST m=+0.036484991",
    "user": "spowelljr"
  },
  "datacontenttype": "application/json",
  "id": "1d695a0e-b7b3-4cb1-bf63-53741a2fd113",
  "source": "https://minikube.sigs.k8s.io/",
  "specversion": "1.0",
  "type": "io.k8s.sigs.minikube.audit"
}
```

spowelljr:
> additionally, I am curious does this design allow us to ban one of the commands from auditing ?
> for example in some 3rd party tool they call "minikube status" every second, this could result in recording a large audit footprint just for status.
> 
> maybe we could either disable it for not intrusive operations such as "status" or "version"

Added functionality to not run on specified commands.

pkg/minikube/audit/audit.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	return strings.Join(os.Args[2:], " ")
+}
+
+// Log details about the executed command.
+func Log(startTime time.Time) {
+	if !shouldLog() {
+		return
+	}
+	e := newEntry(os.Args[1], args(), username(), startTime, time.Now())
```
medyagh, line 56: consider adding a time format. maybe we could use unix time format ? or UTC ...

medyagh: or we could use "time.RFC3339" as I did for the Kubeconfig extension

spowelljr: Time formatting added to `entry.go`

pkg/minikube/audit/audit.go
```diff
@@ -0,0 +1,73 @@
...
+	osUser, err := user.Current()
+	if err != nil {
+		return "UNKNOWN"
+	}
+	return osUser.Username
+}
+
+// args concats the args into space delimited string.
+func args() string {
+	if len(os.Args) < 3 {
```
tstromberg, line 45: Why 3? This seems non-obvious enough to warrant a "why?" comment.

pkg/minikube/audit/logFile.go
```diff
@@ -0,0 +1,72 @@
...
+	logPath string
+)
+
+// setLogFile sets the logPath and creates the log file if it doesn't exist.
+func setLogFile() error {
+	logPath = localpath.AuditLog()
+	if _, err := os.Stat(logPath); os.IsNotExist(err) {
+		return createLogFile()
+	}
+	f, err := os.OpenFile(logPath, os.O_APPEND|os.O_WRONLY, os.ModeAppend)
```
tstromberg, line 38: Is it possible to add  os.O_CREAT so that `createLogFile()` can be removed?

pkg/minikube/audit/logFile.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"fmt"
+	"os"
+
+	"k8s.io/minikube/pkg/minikube/localpath"
+	"k8s.io/minikube/pkg/minikube/out/register"
+)
+
+var (
+	file    *os.File
```
tstromberg, line 28: Since these are globals with infinite scope, can you name them with more verbosity, and add comments describing what their usage is? For example:

`currentLogFile`

pkg/minikube/audit/logFile.go
```diff
@@ -0,0 +1,72 @@
...
+	logPath string
+)
+
+// setLogFile sets the logPath and creates the log file if it doesn't exist.
+func setLogFile() error {
+	logPath = localpath.AuditLog()
```
tstromberg, line 34: I don't think this needs to exist as a global. Instead, it could be passed to `createLogFile`

pkg/minikube/config/config.go
```diff
@@ -48,6 +49,10 @@ const (
 	ShowDriverDeprecationNotification = "ShowDriverDeprecationNotification"
 	// ShowBootstrapperDeprecationNotification is the key for ShowBootstrapperDeprecationNotification
 	ShowBootstrapperDeprecationNotification = "ShowBootstrapperDeprecationNotification"
+	// TimeFormat is the format that should be used when outputing time
+	TimeFormat = time.RFC1123
```
tstromberg, line 53: I don't understand what the time output has to do with config files?

I suspect this constant is in the wrong place.

pkg/minikube/config/config.go
```diff
@@ -48,6 +49,10 @@ const (
...
+	// User represents the key for the global user parameter
```
tstromberg, line 54: key in what? Also, mention what this parameter is used for (it's not obvious)

pkg/minikube/localpath/localpath.go
```diff
@@ -67,6 +67,11 @@ func EventLog(name string) string {
 	return filepath.Join(Profile(name), "events.json")
 }
 
+// AuditLog returns the path to the audit log.
+func AuditLog() string {
+	return filepath.Join(MiniPath(), "logs", "audit.json")
```
tstromberg, line 72: This should probably be saved in the same directory as `events.json`.

pkg/minikube/localpath/localpath.go
```diff
@@ -67,6 +67,11 @@ func EventLog(name string) string {
 	return filepath.Join(Profile(name), "events.json")
 }
 
+// AuditLog returns the path to the audit log.
```
tstromberg, line 70: Can you update the docstrings of this and EventLog to clarify the difference between the two? It isn't obvious to a reader what gets saved in what.

tstromberg (changes requested):

pkg/minikube/audit/logFile.go
medyagh: agree, that would make the funcs more unit-testable

pkg/minikube/localpath/localpath.go
medyagh: I would be okay with it be stored be in "logs" since the nature of audit.json is different (presistant) and event.json is per run and temporary.  but I am okay with different path too ! wdyt @tstromberg ?

test/integration/functional_test.go
```diff
@@ -239,6 +240,25 @@ func validateStartWithProxy(ctx context.Context, t *testing.T, profile string) {
 	if !strings.Contains(rr.Stderr.String(), want) {
 		t.Errorf("start stderr=%s, want: *%s*", rr.Stderr.String(), want)
 	}
+
+	l, err := auditLastLine()
```
medyagh, line 244: here try to add subtests, so when this fails, in gopogh it will show as
/TestFunction/Audit....

medyagh (changes requested):

test/integration/util_test.go
```diff
@@ -59,3 +61,41 @@ func UniqueProfileName(prefix string) string {
 	// example: prefix-20200413162239-3215
 	return fmt.Sprintf("%s-%s-%d", prefix, time.Now().Format("20060102150405"), os.Getpid())
 }
+
+// auditLastLine returns the last line of the audit.go log.
+// To make the runtime constant, even with a large file, it starts at the end and goes backwards
+func auditLastLine() (string, error) {
```
medyagh, line 67: not sure if this is needed (but I could be convinced otherwise)

how about just a simple checking the audit log for existence of a "command"

medyagh (changes requested):

test/integration/util_test.go
spowelljr: I am currently only using this function in serial tests, so the parallelization is not an issue, but if this was used in a parallel test there would definitely be a race condition.

The issue I see with checking for the existence of a command is that `audit.json` is an ongoing log file, so assuming at least a single command is run there will always be the existence of a command. For example, the log file already has several commands in it already, someone modifies something so new commands are not logged, the test will still pass because there is a command present in the log file.

With that said, if I'm assuming correctly that machines that run this in CI start fresh without the file, then the failure should be detected then.

Let me know what you think @medyagh

medyagh: refer reponse to the gvc comments, better to be "parallel safe" and using the distnitive information to get logs
for example "profile" name.

I always try to push back on any integration test that starts to get complex, the integration test code should be dumb simple

pkg/minikube/config/config.go
```diff
@@ -48,6 +48,8 @@ const (
 	ShowDriverDeprecationNotification = "ShowDriverDeprecationNotification"
 	// ShowBootstrapperDeprecationNotification is the key for ShowBootstrapperDeprecationNotification
 	ShowBootstrapperDeprecationNotification = "ShowBootstrapperDeprecationNotification"
+	// User is the key for the global user flag (ex. --user=user1)
+	User = "user"
```
medyagh, line 52: lets call it "UserFlag"

medyagh (changes requested):

test/integration/json_output_test.go
```diff
@@ -54,7 +54,7 @@ func TestJSONOutput(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.command, func(t *testing.T) {
-			args := []string{test.command, "-p", profile, "--output=json"}
+			args := []string{test.command, "-p", profile, "--output=json", "--user=test123"}
```
medyagh, line 57: change to "test-user123"

medyagh (changes requested):

medyagh (changes requested):
please add Basic Validaiton for user flag
for example dont allow more than ...X.. Length username.

medyagh (changes requested):

pkg/minikube/audit/audit.go
```diff
@@ -0,0 +1,78 @@
...
+	"strings"
+	"time"
+
+	"github.com/spf13/viper"
+	"k8s.io/klog"
+	"k8s.io/minikube/pkg/minikube/config"
+)
+
+// UserName pulls the user flag, if empty gets the os username.
+func UserName() string {
```
medyagh, line 31: private

medyagh (changes requested):

medyagh (approved):

pkg/minikube/out/register/json.go
spowelljr: Good idea, done!

pkg/minikube/out/register/log.go
spowelljr: Good idea, done!

pkg/minikube/audit/audit.go
spowelljr: Good idea, done!

pkg/minikube/audit/audit.go
spowelljr: Good point, commented

pkg/minikube/audit/logFile.go
spowelljr: Updated to pass it in

pkg/minikube/config/config.go
spowelljr: Moved to consts file

pkg/minikube/config/config.go
spowelljr: Updated comment to make it more descriptive

pkg/minikube/localpath/localpath.go
spowelljr: Updated comment to clarify difference between funcs

test/integration/functional_test.go
spowelljr: Added subtest

pkg/minikube/config/config.go
spowelljr: Updated name

test/integration/json_output_test.go
spowelljr: Updated name

pkg/minikube/config/user.go
```diff
@@ -0,0 +1,22 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+// UserNameValid checks if the user name is valid.
+func UserNameValid(name string) bool {
```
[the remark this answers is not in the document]
spowelljr: Moved validation to cmd package

pkg/minikube/audit/audit.go
spowelljr: Made the function private

pkg/minikube/audit/logFile.go
spowelljr: Good idea, done!

pkg/minikube/audit/logFile.go
spowelljr: Improved file name and added comment

pkg/minikube/localpath/localpath.go
spowelljr: I agree with Medya, I think it makes sense to leave this in the current directory.

test/integration/util_test.go
spowelljr: Good point, pivoted to this method instead

Issue #10061, opened 2020-12-29
Potential profile naming conflict with multi-node cluster.


daehyeok:
Minikube doesn't have a restriction on profile name. It have a potential conflict with multi-node cluster.
E.g) create test-profile with multiple node will fail if test-profile-m02 profile already exists.

Naming rule for profile or check routine should be added before create new profile(node).

```
❯ minikube start -p test-profile-m02
😄  [test-profile-m02] minikube v1.16.0 on Darwin 10.15.7
✨  Automatically selected the docker driver
👍  Starting control plane node test-profile-m02 in cluster test-profile-m02
🔥  Creating docker container (CPUs=2, Memory=1988MB) ...
🐳  Preparing Kubernetes v1.20.0 on Docker 20.10.0 ...
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔎  Verifying Kubernetes components...
🌟  Enabled addons: storage-provisioner, default-storageclass
🏄  Done! kubectl is now configured to use "test-profile-m02" cluster and "default" namespace by default
❯ minikube start -p test-profile -n 2
😄  [test-profile] minikube v1.16.0 on Darwin 10.15.7
✨  Automatically selected the docker driver
👍  Starting control plane node test-profile in cluster test-profile
🔥  Creating docker container (CPUs=2, Memory=1988MB) ...
🐳  Preparing Kubernetes v1.20.0 on Docker 20.10.0 ...
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔗  Configuring CNI (Container Networking Interface) ...
🔎  Verifying Kubernetes components...
🌟  Enabled addons: storage-provisioner, default-storageclass

👍  Starting node test-profile-m02 in cluster test-profile
🏃  Updating the running docker "test-profile-m02" container ...
🌐  Found network options:
    ▪ NO_PROXY=192.168.59.2
🐳  Preparing Kubernetes v1.20.0 on Docker 20.10.0 ...
    ▪ env NO_PROXY=192.168.59.2

[...]

PR #10114, opened 2021-01-08
refactor: migrate MachineName function to config package

daehyeok:
Migrate MachineName function from driver package to config package.

Related PR&Issues #9996, #10061.
`config` package need to call `driver.MachinName` func to fix several bugs.. but It can't import due to the circular dependency problem.  
 To `MachineName` function is more portable, migrate that to config package.


medyagh (approved):
thank you for this refactor

PR #10119, opened 2021-01-09
Don't allow profile names that conflict with a multi-node name

### LINKED ISSUE #10061: Potential profile naming conflict with multi-node cluster.

daehyeok:
Check profile and node name before add to prevent conflict with
machine name on multinode cluster.

Fixes #10061

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -176,6 +176,8 @@ func runStart(cmd *cobra.Command, args []string) {
 
 	if existing != nil {
 		upgradeExistingConfig(existing)
+	} else {
+		validateProfileName()
 	}
 
 	validateSpecifiedDriver(existing)
@@ -636,6 +638,25 @@ func hostDriver(existing *config.ClusterConfig) string {
 	}
 
 	return h.Driver.DriverName()
+}
+
+// validateProfileName makes sure that new profile name not duplicated with any of machine names in existing multi-node clusters.
+func validateProfileName() {
+	profiles, err := config.ListValidProfiles()
+	if err != nil {
+		exit.Message(reason.InternalListConfig, "Unable to list profiles: {{.error}}", out.V{"error": err})
+	}
+	for _, p := range profiles {
+		for _, n := range p.Config.Nodes {
+			machineName := config.MachineName(*p.Config, n)
+			if ClusterFlagValue() == machineName {
+				out.WarningT("Profile name '{{.name}}' is duplicated with machine name '{{.machine}}' in profile '{{.profile}}'", out.V{"name": ClusterFlagValue(),
+					"machine": machineName,
+					"profile": p.Name})
+				exit.Message(reason.Usage, "Profile name should be unique")
+			}
+		}
+	}
 }
 
 // validateSpecifiedDriver makes sure that if a user has passed in a driver
--- a/pkg/minikube/config/profile.go
+++ b/pkg/minikube/config/profile.go
@@ -223,7 +223,25 @@ func ListProfiles(miniHome ...string) (validPs []*Profile, inValidPs []*Profile,
 	return validPs, inValidPs, nil
 }
 
-// removeDupes removes duplicates
+// ListValidProfiles returns profiles in minikube home dir
+// Unlike `ListProfiles` this function doens't try to get profile from container
+func ListValidProfiles(miniHome ...string) (ps []*Profile, err error) {
+	// try to get profiles list based on left over evidences such as directory
+	pDirs, err := profileDirs(miniHome...)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, n := range pDirs {
+		p, err := LoadProfile(n, miniHome...)
+		if err == nil && p.IsValid() {
+			ps = append(ps, p)
+		}
+	}
+	return ps, nil
+}
+
+// removeDupes removes duplipcates
 func removeDupes(profiles []string) []string {
 	// Use map to record duplicates as we find them.
 	seen := map[string]bool{}
@@ -291,7 +309,7 @@ func ProfileFolderPath(profile string, miniHome ...string) string {
 // MachineName returns the name of the machine, as seen by the hypervisor given the cluster and node names
 func MachineName(cc ClusterConfig, n Node) string {
 	// For single node cluster, default to back to old naming
-	if len(cc.Nodes) == 1 || n.ControlPlane {
+	if (len(cc.Nodes) == 1 && cc.Nodes[0].Name == n.Name) || n.ControlPlane {
 		return cc.Name
 	}
 	return fmt.Sprintf("%s-%s", cc.Name, n.Name)
--- a/test/integration/multinode_test.go
+++ b/test/integration/multinode_test.go
@@ -48,6 +48,7 @@ func TestMultiNode(t *testing.T) {
 			{"DeleteNode", validateDeleteNodeFromMultiNode},
 			{"StopMultiNode", validateStopMultiNodeCluster},
 			{"RestartMultiNode", validateRestartMultiNodeCluster},
+			{"ValidateNameConflict", validatNameConflict},
 		}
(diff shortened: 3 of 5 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -638,6 +639,29 @@ func hostDriver(existing *config.ClusterConfig) string {
 	return h.Driver.DriverName()
 }
 
+// validateProfileName makes sure that new profile name not duplicated with any of machine names in existing multi-node clusters.
+func validateProfileName(existing *config.ClusterConfig) {
+	if existing != nil {
+		return
+	}
+
+	validPs, _, err := config.ListProfiles()
```
medyagh, line 648: I think this is an expensive call, since it checks the status of all Profiles
two options
1- Create a Light Weight ListProfile
config.ListProfileLight() 
which only returns list of profile names and doesnt check if they are running or valid

2- do not allow anyone to create any profile named "name-m0X"
so it never creates that problem.

what do u think ?

medyagh (changes requested):

cmd/minikube/cmd/start.go
daehyeok: I think first option is better idea. May be it can be used in other place and don't bother user if they already created -m0x profile.

medyagh (approved):
looks good, I would like to see example of it in the PR description

medyagh (commented):
can we add integration test for this here ? https://github.com/medyagh/minikube/blob/89a8cdc794098f9ba0550fbb5eeb19bc3dd8774d/test/integration/multinode_test.go#L46

medyagh:
TestMultiNode/serial/RestartMultiNode fails on both Docker and KVM i wonder if it is related?

daehyeok:
> TestMultiNode/serial/RestartMultiNode fails on both Docker and KVM i wonder if it is related?

Fixed it.
	https://github.com/kubernetes/minikube/blob/fb0658c1984d8b5a6a4422a719fcc181104d3748/pkg/minikube/node/node.go#L48-L50

medyagh:
@daehyeok do u mind pulling upstream master on this PR ?

daehyeok:
> @daehyeok do u mind pulling upstream master on this PR ?

rebased with upstream master branch

cmd/minikube/cmd/start.go
```diff
@@ -638,6 +639,29 @@ func hostDriver(existing *config.ClusterConfig) string {
 	return h.Driver.DriverName()
 }
 
+// validateProfileName makes sure that new profile name not duplicated with any of machine names in existing multi-node clusters.
+func validateProfileName(existing *config.ClusterConfig) {
+	if existing != nil {
```
priyawadhwa, line 644: Shouldn't this be

```go
if existing == nil {
  return
}
```

Either way, it doesn't look like `existing` is used at all in this function. If that's intentional, I would probably remove this if statement and rewrite the function header to:


```go
func validateProfileName(_ *config.ClusterConfig) {
```

pkg/minikube/config/profile.go
```diff
@@ -223,7 +223,25 @@ func ListProfiles(miniHome ...string) (validPs []*Profile, inValidPs []*Profile,
 	return validPs, inValidPs, nil
 }
 
-// removeDupes removes duplicates
+// ListProfilesLight returns profiles in minikube home dir
+// Unlike `ListProfiles` this function doens't try to get profile from container
+func ListProfilesLight(miniHome ...string) (ps []*Profile, err error) {
```
priyawadhwa, line 228: I would rename this function to `ListValidProfiles` to more accurately describe what is happening

test/integration/multinode_test.go
```diff
@@ -308,3 +309,39 @@ func validateDeleteNodeFromMultiNode(ctx context.Context, t *testing.T, profile
...
+	rr, err = Run(t, exec.CommandContext(ctx, Target(), startArgs...))
+	if err != nil {
+		t.Errorf("failed to start profile. args %q : %v", rr.Command(), err)
+	}
+
+	// Add a node to the current cluster. It's expected failture
+	addArgs := []string{"node", "add", "-p", profile}
+	rr, err = Run(t, exec.CommandContext(ctx, Target(), addArgs...))
+	if err == nil {
+		t.Errorf("expected add node command to fail fail. args %q : %v", rr.Command(), err)
```
priyawadhwa, line 340: nit: "fail fail" should just be "fail"

daehyeok: Oops, fixed

cmd/minikube/cmd/start.go
daehyeok: Remove argument, and changed to call function only existing is not nil.

```
if existing != nil {
		upgradeExistingConfig(existing)
		validateProfileName()
	}
```

pkg/minikube/config/profile.go
daehyeok: Fixed.

sharifelgamal (approved):

medyagh (approved):

Issue #9970, opened 2020-12-15
Add minikube annotation to kubeconfig context and cluster extensions


briandealwis:
Running `minikube profile list -ojson` is surprisingly slow, taking up to 2.5s on my 2016 MBP.  In part this may be due to my having a second profile from `gcloud code dev`.

This is relevant as Skaffold currently invokes `minikube profile list` to determine if it's building against Minikube.  (We actually use a cheaper heuristic by first checking if the context name is `minikube`, but that will be no longer possible now that people are experimenting with running Minikube with containerd as its container runtime https://github.com/GoogleContainerTools/skaffold/issues/5143).

Our use of `minikube profile list` causes a noticeable delay when deploying to a remote cluster.

**Steps to reproduce the issue:** 

1. `minikube profile list -ojson`

**Full output of failed command:** 

```
$  time minikube profile list -ojson --alsologtostderr
I1215 15:08:27.294810   91926 out.go:185] Setting OutFile to fd 1 ...
I1215 15:08:27.295371   91926 out.go:237] isatty.IsTerminal(1) = true
I1215 15:08:27.295381   91926 out.go:198] Setting ErrFile to fd 2...
I1215 15:08:27.295387   91926 out.go:237] isatty.IsTerminal(2) = true
I1215 15:08:27.295470   91926 root.go:279] Updating PATH: /Users/bdealwis/.minikube/bin
I1215 15:08:27.296589   91926 cli_runner.go:110] Run: docker ps -a --filter label=name.minikube.sigs.k8s.io --format {{.Names}}
I1215 15:08:27.456061   91926 host.go:66] Checking if "gcloud-local-dev" exists ...
I1215 15:08:27.457050   91926 cli_runner.go:110] Run: docker container inspect gcloud-local-dev --format={{.State.Status}}
W1215 15:08:27.617335   91926 cli_runner.go:148] docker container inspect gcloud-local-dev --format={{.State.Status}} returned with exit code 1
W1215 15:08:27.618631   91926 profile_list.go:104] error getting host state: unknown state "gcloud-local-dev": docker container inspect gcloud-local-dev --format={{.State.Status}}: exit status 1
stdout:


stderr:
Error: No such container: gcloud-local-dev
I1215 15:08:27.618664   91926 host.go:66] Checking if "minikube" exists ...
I1215 15:08:27.619629   91926 cli_runner.go:110] Run: docker container inspect minikube --format={{.State.Status}}
[...]

PR #10126, opened 2021-01-11
add  annotation to kubeconfig to identify contexts/clusters created by minikube

medyagh:
After this PR the following extension blocks to the context and cluster section of  ~/.kube/config
```
apiVersion: v1
clusters:
- cluster:
    certificate-authority: /Users/medya/.minikube/ca.crt
    extensions:
    - extension:
        last-update: Thu, 14 Jan 2021 15:41:04 PST
        provider: minikube.sigs.k8s.io
        version: v1.16.0
      name: cluster_info
    server: https://127.0.0.1:55000
  name: minikube
contexts:
- context:
    cluster: minikube
    extensions:
    - extension:
        last-update: Thu, 14 Jan 2021 15:41:04 PST
        provider: minikube.sigs.k8s.io
        version: v1.16.0
      name: context_info
    namespace: default
    user: minikube
  name: minikube
current-context: minikube
kind: Config
preferences: {}
users:
- name: minikube
  user:
    client-certificate: /Users/medya/.minikube/profiles/minikube/client.crt
    client-key: /Users/medya/.minikube/profiles/minikube/client.key


```

closes https://github.com/kubernetes/minikube/issues/9970


sharifelgamal:
do we want to use hyphens instead of underscores (`created-by` and `last-update` instead of `created_ by` and `last_update`) to follow the convention set by kubeconfig?

pkg/minikube/kubeconfig/settings.go
```diff
@@ -83,6 +85,14 @@ func PopulateFromSettings(cfg *Settings, apiCfg *api.Config) error {
 	} else {
 		cluster.CertificateAuthority = cfg.CertificateAuthority
 	}
+
+	lastUpdate := time.Now().String()
```
briandealwis, line 89: Can this use `.Format(time.RFC3339)`?

sharifelgamal: also it doesn't seem like assigning a variable here serves any real purpose. i don't think lastUpdate gets reused anywhere, right?

briandealwis: It may help humans who edit their `~/.kube/config` to determine how fresh this section might be?

briandealwis:
I get your dislike of using `cluster-type` since they're Kubernetes clusters.  But my concern about `created-by` is that it can refer to a tool like terraform.  I wonder if there's a different term that we could settle on like `provider`.  Basically within Skaffold I want to determine which helper command Skaffold should use to introspect on the cluster to find out more about its setup.

pkg/minikube/kubeconfig/settings.go
```diff
@@ -149,3 +160,36 @@ func Update(kcs *Settings) error {
 	}
 	return nil
 }
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
priyawadhwa, line 164: I think it would be useful to move these functions into internal_extension.go for organization

medyagh:
> I get your dislike of using `cluster-type` since they're Kubernetes clusters. But my concern about `created-by` is that it can refer to a tool like terraform. I wonder if there's a different term that we could settle on like `provider`. Basically within Skaffold I want to determine which helper command Skaffold should use to introspect on the cluster to find out more about its setup.

I dont think it is a good idea to call it cluster-type when minikube creates the cluster, minikube clusters are not a "type" they are same clusters created by standard "kubedam" it would be misleading or confusing for normal users to see in their kubeconfig it call it "cluster_type"= minikube

pkg/minikube/kubeconfig/settings.go
medyagh: good idea

pkg/minikube/kubeconfig/settings.go
medyagh: good catch the var was initially used two places got lost in refactor.

medyagh: good call on formatting.

pkg/minikube/kubeconfig/settings.go
medyagh: good idea

medyagh:
> I get your dislike of using `cluster-type` since they're Kubernetes clusters. But my concern about `created-by` is that it can refer to a tool like terraform. I wonder if there's a different term that we could settle on like `provider`. Basically within Skaffold I want to determine which helper command Skaffold should use to introspect on the cluster to find out more about its setup.

"provider" sounds like a good suggestion

medyagh:
> I get your dislike of using `cluster-type` since they're Kubernetes clusters. But my concern about `created-by` is that it can refer to a tool like terraform. I wonder if there's a different term that we could settle on like `provider`. Basically within Skaffold I want to determine which helper command Skaffold should use to introspect on the cluster to find out more about its setup.

"provider" sounds like a good suggestion

priyawadhwa (approved):

medyagh:
> ad of underscores (`created-by` and `last-update` instead of `created_ by` and `last_update`) to follow the convention set by kubeconfig?

good idea

PR #10132, opened 2021-01-12
CI: add TestScheduledStop on windows in github actions

priyawadhwa:


.github/workflows/pr.yml
```diff
@@ -277,11 +277,11 @@ jobs:
           if [ "$numPass" -eq 0 ];then echo "*** 0 Passed! ***";exit 2;fi
           if [ "$numPass" -lt 33 ];then echo "*** Failed to pass at least 33 ! ***";exit 2;fi
           if [ "$numPass" -eq 0 ];then echo "*** Passed! ***";exit 0;fi
-  functional_docker_windows:
+  functional_scheduled_stop_docker_windows:
```
medyagh, line 280: lets add this as its own job, since test functional itself is flaky

also needs to be added to both PR and master.yml

medyagh (changes requested):

.github/workflows/pr.yml
priyawadhwa: thanks

priyawadhwa:
Hey @medyagh looking at the logs for the failing github actions tests, it looks like the test is passing on both. It seems like gopogh isn't picking up the test success though, is there anything else I need to add?

medyagh (changes requested):
the test are failing and not producing logs...lets ensure they pass before merging

.github/workflows/pr.yml
```diff
@@ -555,6 +555,200 @@ jobs:
...
+        continue-on-error: true
+        shell: powershell
+        run: |
+          (New-Object Net.WebClient).DownloadFile("https://github.com/medyagh/gopogh/releases/download/v0.3.0/gopogh.exe", "C:\ProgramData\chocolatey\bin\gopogh.exe")
+          choco install -y kubernetes-cli
+          choco install -y jq
+          choco install -y caffeine
+          if (Test-Path 'C:\Program Files\Docker\Docker\resources\bin\kubectl.exe') { Remove-Item 'C:\Program Files\Docker\Docker\resources\bin\kubectl.exe' };
+      - name: Run Integration Test in powershell
+        continue-on-error: true
```
medyagh, line 635: I think this will always show as passed regardless if it fails or not

medyagh (changes requested):

medyagh (approved):

PR #10178, opened 2021-01-19
Add jenkins test for containerd on docker driver

priyawadhwa:
This also adds a global variable for CONTAINER_RUNTIME to jenkins


medyagh (changes requested):
need to update the set pending one

https://github.com/medyagh/minikube/blob/89a8cdc794098f9ba0550fbb5eeb19bc3dd8774d/hack/jenkins/minikube_set_pending.sh#L34

never mind my bad.

priyawadhwa:
@medyagh I did that -- https://github.com/kubernetes/minikube/pull/10178/files#diff-15b571bc5735e44e090f03f557359a393925a350c44c875037d6490904251c1aR44

medyagh:
The scripts need to be added to the Jenkins job manually (to be copied)
```
TimestampsView as plain text
--
System clock timeUse browser timezoneElapsed timeNone

```

priyawadhwa:
@medyagh yup, I added it to the main job as well so that it would kick off the new test.

medyagh:
> @medyagh yup, I added it to the main job as well so that it would kick off the new test.

I see, it is very confusing to click on every job configure, we should consider moving our jenkins configs to github https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/migrate.md

medyagh (approved):

PR #10187, opened 2021-01-20
Avoid warning about the suggested memory size

afbjorklund:
The recommendation is to use (at least) 2048 MB *for the VM*,
which means that you will end up something like 1992 MB total.
We should not issue warnings when using the recommended size,
so lower the limit to 1900 MB (kubeadm will validate 1700 MB).

The reason it is broken is because it uses different values
for different drivers, if running a VM or on Bare Metal...
There is also no difference anymore between the "usable"
and the "recommended", so this will need revisiting later.


afbjorklund:
Currently you get annoying warnings, if you create a 2 GiB VM and use the `none` driver on it.

tstromberg (approved):

medyagh (approved):

PR #10198, opened 2021-01-21
Add minikube cp command as new feature

### LINKED ISSUE #9930: minikube scp
It would be nice to be able to have `minikube scp` like `minikube ssh`, to copy files in the the minikube node.  An use case would be: copying Dockerfiles to the node to build custom docker images for Kubernetes.

In the meantime, for all of you that want this feature, here's a workaround:
```
scp -i $(minikube ssh-key) <file> docker@$(minikube ip):
```

**edit:** you can also use `minikube docker-env` to set up the `docker` client to point to minikube's docker daemon, and build the image locally.  However, this still doesn't work for *copying* files...

anencore94:
Add New Feature which represents scp command

Until now, it is possible to copy specific local file to specific path in minikube basically.
I mean 'basically' since it is possible to only file not directory.
And there are many more remaining tasks until now. (remaining problems are written in source code as `//TODO`)

Fix #9930

### DIFF
--- a/cmd/minikube/cmd/cp.go
+++ b/cmd/minikube/cmd/cp.go
@@ -1,0 +1,67 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"os"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"k8s.io/minikube/pkg/minikube/assets"
+	"k8s.io/minikube/pkg/minikube/exit"
+	"k8s.io/minikube/pkg/minikube/mustload"
+	"k8s.io/minikube/pkg/minikube/out"
+	"k8s.io/minikube/pkg/minikube/reason"
+)
+
+// placeholders for flag values
+var (
+	srcPath string
+	dstPath string
+)
+
+// cpCmd represents the cp command, similar to docker cp
+var cpCmd = &cobra.Command{
+	Use:   "cp <source file path> <target file path followed by '/home/docker/'>",
+	Short: "Copy the specified file into minikube",
+	Long: "Copy the specified file into minikube, it will be saved at path \"/home/docker/<target file path>\" in your minikube.\n" +
+		"Example Command : \"minikube cp a.txt b.txt\"\n",
+	Run: func(cmd *cobra.Command, args []string) {
+		// validate args
+		if len(args) != 2 {
+			exit.Message(reason.Usage, `Please specify the path to copy: 
+	minikube cp <source file path> <target file path>   (example: "minikube cp a/b/c/d.txt a.txt")`)
+		}
+		srcPath = args[0]
+		dstPath = args[1]
+
+		co := mustload.Running(ClusterFlagValue())
+		fa, err := assets.NewFileAsset(srcPath, "/home/docker/", dstPath, "0644")
+		if err != nil {
+			out.ErrLn("%v", errors.Wrap(err, "getting file asset"))
+			os.Exit(1)
+		}
+		if err = co.CP.Runner.Copy(fa); err != nil {
+			out.ErrLn("%v", errors.Wrap(err, "copying file"))
+			os.Exit(1)
+		}
+	},
+}
+
+func init() {
+}
--- a/cmd/minikube/cmd/cp.go
+++ b/cmd/minikube/cmd/cp.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2016 The Kubernetes Authors All rights reserved.
+Copyright 2021 The Kubernetes Authors All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -17,10 +17,12 @@ limitations under the License.
 package cmd
 
 import (
-	"os"
-
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+
+	"os"
+	pt "path"
+	"path/filepath"
 
 	"k8s.io/minikube/pkg/minikube/assets"
 	"k8s.io/minikube/pkg/minikube/exit"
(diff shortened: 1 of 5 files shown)


anencore94:
Before coding some other feature such as 'enable -r option to copy directory', 'enable progress bar option' and the others, I want to get some reviews from you guys to check whether I am going the right way.

I also wonder how to implement the test code for this. Is it ok to unit test only inner function? or needs some integration tests?

Thanks a lot in advance!

medyagh (changes requested):
please remove TODO comments and add integration test then it will be good to do !

medyagh:
@anencore94 sorry for the late reply on this PR, this sounds like a good PR, lets remove the TODO notes and yes lets add an integration test to our Funcitonal Test

it could be a subtest simmilar to other tests
```

	// Parallelized tests
	t.Run("parallel", func(t *testing.T) {
		tests := []struct {
			name      string
			validator validateFunc
		}{
			{"ConfigCmd", validateConfigCmd},
			{"DashboardCmd", validateDashboardCmd},
			{"DryRun", validateDryRun},
// ...add your test here
			{"ScpCmd", validateScp},
..
```

and u can make a dummy hello world file and try to copy it using scp and validate the file was copied by
```
minikube ssh cat /tmp/hello_scp.txt
```

anencore94:
Thanks for your review @medyagh !

I added the functional test for scp cmd as your comment. Could you re-check my PR?

* the functional test for "PR / functional_docker_windows (pull_request)" and "PR / functional_hyperv_windows (pull_request)" failed since the scp command does not support on windows i guess.
So I made the functional test to skip when the os is windows.

priyawadhwa (changes requested):
Instead of shelling out to the `scp` flag directly, let's instead copy files into minikube the way we do in other parts of the code (for things like preload, or kubernets binaries). Instead of scp:

1. Create a FileAsset with the [NewFileAsset](https://github.com/kubernetes/minikube/blob/f5478a411781c414fdbd06b6f8f0bf415f0cb364/pkg/minikube/assets/vm_assets.go#L95) function
2. Using a runner, run the built in Copy() function to copy the file asset into minikube

This way anyone can copy files in, regardless of if they have `scp` installed. This should also allow this command to run across all platforms, not just Linux. So, you shouldn't have to skip the functional test anymore.

Example of this in the code: 

https://github.com/kubernetes/minikube/blob/6afefe6b8300251b2fd54974ee4a642c8fbee519/pkg/minikube/cruntime/containerd.go#L350-L357 

If you need any more examples, searching for any call to `asset.NewFileAsset` should get you there.

medyagh:
> Instead of shelling out to the `scp` flag directly, let's instead copy files into minikube the way we do in other parts of the code (for things like preload, or kubernets binaries). Instead of scp:
> 
> 1. Create a FileAsset with the [NewFileAsset](https://github.com/kubernetes/minikube/blob/f5478a411781c414fdbd06b6f8f0bf415f0cb364/pkg/minikube/assets/vm_assets.go#L95) function
> 2. Using a runner, run the built in Copy() function to copy the file asset into minikube
> 
> This way anyone can copy files in, regardless of if they have `scp` installed. This should also allow this command to run across all platforms, not just Linux. So, you shouldn't have to skip the functional test anymore.
> 
> Example of this in the code:
> 
> https://github.com/kubernetes/minikube/blob/6afefe6b8300251b2fd54974ee4a642c8fbee519/pkg/minikube/cruntime/containerd.go#L350-L357
> 
> If you need any more examples, searching for any call to `asset.NewFileAsset` should get you there.

@anencore94 I agree with @priyawadhwa lets do that

priyawadhwa:
Since we're no longer using `scp` I would probably also rename this command to `minikube cp` (matches `docker` so is more intuitive)

anencore94:
@priyawadhwa I changed the implementation of scp by copy with your review comments, Thanks!
By the way, it seems the function Copy of Runner doesn't support directory copy on its own. So If we want to copy directory, I guess we have to make directory and copy every files in for loop or some other way with Copy function.
Thus, for now, I just implement "minikube cp" for only file.

I want to implement the directory copying feature with another PR. (or someone else could do this)

cmd/minikube/cmd/cp.go
```diff
@@ -0,0 +1,67 @@
...
+
+// placeholders for flag values
+var (
+	srcPath string
+	dstPath string
+)
+
+// cpCmd represents the cp command, similar to docker cp
+var cpCmd = &cobra.Command{
+	Use:   "cp <source file path> <target file path followed by '/home/docker/'>",
```
priyawadhwa, line 40: I'm a bit confused by this description, I thought the command would work something like: 

```
minikube cp a.txt. /b.txt
```

Where does the /home/docker come in?

Additionally, we should probably insist that the second argument is an absolute path.

cmd/minikube/cmd/cp.go
```diff
@@ -0,0 +1,67 @@
...
+	Long: "Copy the specified file into minikube, it will be saved at path \"/home/docker/<target file path>\" in your minikube.\n" +
+		"Example Command : \"minikube cp a.txt b.txt\"\n",
+	Run: func(cmd *cobra.Command, args []string) {
+		// validate args
+		if len(args) != 2 {
+			exit.Message(reason.Usage, `Please specify the path to copy: 
+	minikube cp <source file path> <target file path>   (example: "minikube cp a/b/c/d.txt a.txt")`)
+		}
+		srcPath = args[0]
+		dstPath = args[1]
```
priyawadhwa, line 51: Let's error out if dstPath isn't an absolute path

priyawadhwa (changes requested):
A couple more comments, this is almost there!

cmd/minikube/cmd/cp.go
anencore94: I agree with you the prefix_path '/home/docker/' is confusing :)
By the way, when we do 'minikube ssh', the starting path is '/home/docker' since the default username is docker.
Is it ok to copy on '/somewhere' where the permission is only for 'root user' ?

I think there maybe 3 options: (or you have any idea, please let me know :) )
1) minikube cp a.txt /home/docker/b.txt
  - make minikube user to write '/home/docker' to see exactly where the path is.
  - But, they should write the /home/docker path for every copying
2) minikube cp a.txt /b.txt
  - enable copy at root permissioned directory
3) minikube cp a.txt b.txt
  - minikube user doesn't care about the user name, but always will be at /home/docker/b.txt as what i've done

anencore94: What Option would be the best ? @priyawadhwa 
In my opinion for now, option 1 or option 3 will be safe.

priyawadhwa: I think we should ask the user for an absolute path and try to copy to wherever they want the file copied. If it fails because the command requires `sudo`, then we should provide a `--force` or `-f` flag which will copy over the file with root permissions.

cmd/minikube/cmd/cp.go
anencore94: DONE

cmd/minikube/cmd/cp.go
anencore94: Thanks for your comment.
I've changed to minikube copy can copy anywhere in minikube as your opinion.
Could you re-check? @priyawadhwa

medyagh (commented):
@anencore94 can u plz rebase the PR (pull upstream) ?

anencore94:
> @anencore94 can u plz rebase the PR (pull upstream) ?

Thanks, I've done @medyagh

medyagh (approved):

medyagh:
thank you @anencore94

Issue #10175, opened 2021-01-19
why crictl warns on Kic and not on VM "As the default settings are now deprecated, you should set the endpoint instead."


medyagh:
```
medya@~/workspace/minikube (extension_context) $ minikube ssh
docker@minikube:~$ sudo crictl images -o json
{
  "images": [
    {
      "id": "sha256:85069258b98ac4e9f9fbd51dfba3b4212d8cd1d79df7d2ecff44b1319ed641cb",
      "repoTags": [
        "gcr.io/k8s-minikube/storage-provisioner:v4"
      ],
      "repoDigests": [
        "gcr.io/k8s-minikube/storage-provisioner@sha256:06f83c679a723d938b8776510d979c69549ad7df516279981e23554b3e68572f"
      ],
      "size": "29683712",
      "uid": null,
      "username": ""
    },
    {
      "id": "sha256:bfe3a36ebd2528b454be6aebece806db5b40407b833e2af9617bf39afaff8c16",
      "repoTags": [
        "k8s.gcr.io/coredns:1.7.0"
      ],
      "repoDigests": [
        "k8s.gcr.io/coredns@sha256:73ca82b4ce829766d4f1f10947c3a338888f876fbed0540dc849c89ff256e90c"
      ],
      "size": "45227747",
      "uid": null,
      "username": ""
    },
    {
      "id": "sha256:0369cf4303ffdb467dc219990960a9baa8512a54b0ad9283eaf55bd6c0adb934",
      "repoTags": [
        "k8s.gcr.io/etcd:3.4.13-0"
      ],
      "repoDigests": [
        "k8s.gcr.io/etcd@sha256:4ad90a11b55313b182afc186b9876c8e891531b8db4c9bf1541953021618d0e2"
      ],
      "size": "253392289",
      "uid": null,
      "username": ""
    },
    {
      "id": "sha256:ca9843d3b545457f24b012d6d579ba85f132f2406aa171ad84d53caa55e5de99",
      "repoTags": [
        "k8s.gcr.io/kube-apiserver:v1.20.0"
      ],
      "repoDigests": [
        "k8s.gcr.io/kube-apiserver@sha256:8b8125d7a6e4225b08f04f65ca947b27d0cc86380bf09fab890cc80408230114"
      ],
      "size": "121665018",
      "uid": null,
      "username": ""
    },
    {
[...]

PR #10206, opened 2021-01-21
Fix failing TestStartStop/serial/VerifyKubernetesImages test

priyawadhwa:
This test is failing because we use crictl to get a list of k8s images running. crictl prints Warnings to STDOUT instead of STDERR, so when we tried to marshal output into a JSON struct we failed every time.

I fixed this by parsing out the warning, and only marshaling the JSON part of STDOUT.

This test fails 100% of the time according to our failure dashboard, so it should pass on this PR before we merge it.

It would be best if we could address the warning and fix this altogether, which is being tracked here: https://github.com/kubernetes/minikube/issues/10175


sharifelgamal (approved):

medyagh (commented):
while this is okay ! the better solution is fixing the warning 
on Kic there is no wranning only happens on VM and it is becasause we are not populating the CRI https://github.com/kubernetes/minikube/issues/10175#issuecomment-763379680

I accept this PR but I would like to see a better solution.  as described in the link

PR #10219, opened 2021-01-22
refactor download only test for better visibility

sharifelgamal:
TestDownloadOnly had one subtest ("check_json_event') but still had other tests running in the parent, which makes it difficult for gopogh to parse failures in the parent. This PR changes the tests so that we only have subtests, so we know exactly what fails.

### DIFF
--- a/test/integration/aaa_download_only_test.go
+++ b/test/integration/aaa_download_only_test.go
@@ -55,26 +55,31 @@ func TestDownloadOnly(t *testing.T) {
 				constants.NewestKubernetesVersion,
 			}
 
+			// Small optimization, don't run the exact same set of tests twice
+			if constants.DefaultKubernetesVersion == constants.NewestKubernetesVersion {
+				versions = versions[:len(versions)-1]
+			}
+
 			for _, v := range versions {
 				t.Run(v, func(t *testing.T) {
 					defer PostMortemLogs(t, profile)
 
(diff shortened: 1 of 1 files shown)


sharifelgamal:
The tests will get fixed once #10196 is merged

test/integration/aaa_download_only_test.go
```diff
@@ -54,26 +53,31 @@ func TestDownloadOnly(t *testing.T) {
...
-
-			rt, err := Run(t, exec.CommandContext(ctx, Target(), args...))
-			if rrr == nil {
-				// Preserve the initial run-result for debugging
-				rrr = rt
-			}
-			if err != nil {
-				t.Errorf("failed to download only. args: %q %v", args, err)
-			}
 			t.Run("check json events", func(t *testing.T) {
```
medyagh, line 65: how about choosing shorter names ? so it shows better in gopogh url ?
"json-events" instead "check json events"

medyagh: and same for other names

medyagh (changes requested):

medyagh (approved):

test/integration/aaa_download_only_test.go
sharifelgamal: done

PR #10223, opened 2021-01-22
Fix TestDownloadOnlyKic so it takes container runtime into account

priyawadhwa:
This test was previously failing Docker_Linux_containerd, now it should pass.


medyagh (approved):
thanks for fixing this

PR #10225, opened 2021-01-22
Fix TestForceSystemdFlag tests on containerd

priyawadhwa:
Currently, the tests assume the docker runtime. Fix them to work for containerd as well.

We could add support for crio to this test as well. However, I think this is of lower importance because crio defaults to systemd as cgroup manager anyways.


medyagh (approved):
pre-approved for convenience,  thanks for fixing it

PR #10227, opened 2021-01-22
Only run integration tests that depend on docker-env with docker container runtime

priyawadhwa:

### DIFF
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -159,6 +159,7 @@ func validateNodeLabels(ctx context.Context, t *testing.T, profile string) {
 }
 
 // check functionality of minikube after evaling docker-env
+// TODO: Add validatePodmanEnv for crio runtime: #10231
 func validateDockerEnv(ctx context.Context, t *testing.T, profile string) {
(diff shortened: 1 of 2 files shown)


test/integration/functional_test.go
```diff
@@ -160,6 +160,9 @@ func validateNodeLabels(ctx context.Context, t *testing.T, profile string) {
 
 // check functionality of minikube after evaling docker-env
 func validateDockerEnv(ctx context.Context, t *testing.T, profile string) {
+	if cr := ContainerRuntime(); cr != "docker" {
```
medyagh, line 163: looks good ! but lets add a TODO issued to add "vadliatePodmanEnv" if the container-runtime is cri-o

medyagh (changes requested):

test/integration/functional_test.go
priyawadhwa: Done, added a TODO and created an issue https://github.com/kubernetes/minikube/issues/10231

medyagh (approved):

medyagh (approved):

PR #10230, opened 2021-01-22
Fix TestFunctional subtests to run for all three runtimes

priyawadhwa:
Currently,  TestFunctional/serial/CacheCmd/cache/cache_reload and TestFunctional/parallel/LogsCmd are failing on containerd because they assume the docker container runtime. [failure gopogh](https://storage.googleapis.com/minikube-builds/logs/10225/35ba15b/Docker_Linux_containerd.html)

This change should fix subtests for all three runtimes.

### DIFF
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -699,9 +699,19 @@ func validateLogsCmd(ctx context.Context, t *testing.T, profile string) {
 	if err != nil {
 		t.Errorf("%s failed: %v", rr.Command(), err)
 	}
-	for _, word := range []string{"Docker", "apiserver", "Linux", "kubelet"} {
+	expectedWords := []string{"apiserver", "Linux", "kubelet"}
+	switch ContainerRuntime() {
+	case "docker":
+		expectedWords = append(expectedWords, "Docker")
+	case "containerd":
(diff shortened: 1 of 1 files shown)


test/integration/functional_test.go
```diff
@@ -617,7 +617,16 @@ func validateCacheCmd(ctx context.Context, t *testing.T, profile string) {
...
 			img := "k8s.gcr.io/pause:latest"
 			// deleting image inside minikube node manually
-			rr, err := Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "sudo", "docker", "rmi", img))
+
+			var binary string
+			switch ContainerRuntime() {
+			case "docker":
+				binary = "docker"
+			case "containerd", "crio":
+				binary = "crictl"
```
medyagh, line 626: Can we use crictl for docker runtime too ?
( I think it's possible) if yes just .make everything use crictl

priyawadhwa: no, deleting the image with crictl will not delete it from the docker daemon, so it can still be run with the docker runtime. we need to use the docker binary to correctly run this test.

medyagh (approved):

PR #10235, opened 2021-01-23
Add resource requests & limits for mysql test

### LINKED ISSUE #10185: Add Cpu/memory limits on minikube resources
for example on mysql test we can specify max cpu and memory to not allow mysql to over eat the resources
https://github.com/medyagh/minikube/blob/89a8cdc794098f9ba0550fbb5eeb19bc3dd8774d/test/integration/testdata/mysql.yaml#L29

we could possibly add limiations to api sever or other things (etcd) in the next iteration

example https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/

BLasan:
fixes #10185 

Deployments under test/integration do not contain resource limits & requests. By adding limits & requests, it will prevent over consuming the node resources

### DIFF
--- a/test/integration/testdata/mysql.yaml
+++ b/test/integration/testdata/mysql.yaml
@@ -24,8 +24,15 @@ spec:
         app: mysql
     spec:
       containers:
-      - image: mysql:5.6
+      - image: mysql:5.7
         name: mysql
+        resources:
+          requests:
+            memory: "512Mi"
+            cpu: "600m"           
+          limits:
+            memory: "700Mi"
+            cpu: "700m"
         env:
           # Use secret in real usage
         - name: MYSQL_ROOT_PASSWORD


test/integration/testdata/mysql.yaml
```diff
@@ -26,6 +26,13 @@ spec:
       containers:
       - image: mysql:5.6
         name: mysql
+        resources:
+          requests:
+            memory: "512Mi"
+            cpu: "500m"
+          limits:
+            memory: "1Gi"
+            cpu: "1000m"
```
medyagh, line 35: do u mind clarifying how you got these numbers? is there a chance we could go lower  ? would you please try running the functional test with "Lowest" possible CPU and memory limits before mysql breaks ?
to run funcitonal you can run:

"make functional"

medyagh (changes requested):

test/integration/testdata/mysql.yaml
BLasan: For the following configurations(minimum resources) the tests will get passed.
```
        resources:
          limits:
            memory: "512Mi"
            cpu: "1000m"
```

BLasan:
@medyagh updated the memory & cpu limits in mysql deployment by running the functional tests. Please review sir

test/integration/testdata/mysql.yaml
medyagh: why the request limits were deleted ?
1000m cpu seems a lot, have u tried lower and it failed?

test/integration/testdata/mysql.yaml
```diff
@@ -26,6 +26,10 @@ spec:
       containers:
       - image: mysql:5.6
```
medyagh, line 27: consider updating the mysql to a newer version maybe better performance ?

BLasan: yuh, I was thinking about that. We could use mysql:5.7 instead of mysql:5.6

test/integration/testdata/mysql.yaml
BLasan: Yuh, I've tried with 500m. But it fails the test sir. I'll check some values between 500-1000m. May be we could be able to reduce the CPU limit

BLasan:
@medyagh Sir, I 've updated mysql from 5.6 -> 5.7 and minimized the cpu limit to 600m as 500m fails the tests.

BLasan:
@medyagh MYSQL deployment needs only "512Mi" memory and "600m" cpu. I have added them as the resources requests and configured limits the resources as mentioned in the file so it'll not go beyond those values. Is it ok sir?

test/integration/testdata/mysql.yaml
medyagh: yes please lets try with less than 1000m cpu and see what is min that works, and add +100 to that absolute min

medyagh (approved):
thank you @BLasan

PR #10247, opened 2021-01-24
Replacing a command with the command line's suggested one.

didof:
While following the instructions for the correct use of minikube, I have come to the point where it is recommended to invoke the command `eval $(minikube docker-env)` in order to use the docker daemon inside the VM. However, when executed, the command does not appear to do what it should, so subsequent docker build commands are interpreted by the "local" docker daemon rather than the one in minikube.

By executing `minikube docker-env` I got the following output:
`
To point your shell to minikube's docker-daemon, run:
eval $(minikube -p minikube docker-env)
`

And actually, by running this last one, subsequent docker builds took place in the VM, so it was possible to the latter to find the images locally.

### DIFF
--- a/site/content/en/docs/handbook/pushing.md
+++ b/site/content/en/docs/handbook/pushing.md
@@ -38,7 +38,7 @@ this means you don't have to build on your host machine and push the image into 
 To point your terminal to use the docker daemon inside minikube run this:
 
 ```shell
-eval $(minikube docker-env)
+eval $(minikube -p minikube docker-env)
 ```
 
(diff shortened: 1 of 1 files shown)


didof:

I confirm that I have successfully completed the registration indicated.

leodido:
Hey Frà (@didof) comment here with the text "I signed it" to let the system recheck :)

didof:
I signed it

medyagh (changes requested):
this is not required, the command line suggested is to be used for other tools. for normal user this is not mandatory if they use one profile

didof:
Thank you for the comment. I am going to investigate more into it.

Issue #10130, opened 2021-01-12
Flaky TestFunctional/parallel/ComponentHealth: pending


lingsamuel:
/kind failing-test
/kind flake

```log

2021-01-12T08:54:30.1684190Z === CONT  TestFunctional/parallel/ComponentHealth
2021-01-12T08:54:30.1685461Z     functional_test.go:390: etcd phase: Running
2021-01-12T08:54:30.1687029Z     functional_test.go:390: control-plane phase: Running
2021-01-12T08:54:30.1687925Z     functional_test.go:390: kube-apiserver phase: Pending
2021-01-12T08:54:30.1689403Z     functional_test.go:395: kube-apiserver is not Running: {Phase:Pending Conditions:[] Message: Reason: HostIP: PodIP: StartTime:<nil> ContainerStatuses:[]}
2021-01-12T08:54:30.1690567Z     functional_test.go:390: control-plane phase: Pending
2021-01-12T08:54:30.1691459Z     functional_test.go:390: kube-controller-manager phase: Running
```

```log

2021-01-12T08:54:32.4341001Z         	* ==> kubelet <==
2021-01-12T08:54:32.4341664Z         	* -- Logs begin at Tue 2021-01-12 08:53:06 UTC, end at Tue 2021-01-12 08:54:32 UTC. --
2021-01-12T08:54:32.4344563Z         	* Jan 12 08:54:23 functional-20210112085224-2630 kubelet[5095]: I0112 08:54:23.848127    5095 reconciler.go:224] operationExecutor.VerifyControllerAttachedVolume started for volume "k8s-certs" (UniqueName: "kubernetes.io/host-path/dbabc6fcaa35495e1aa10964d680d780-k8s-certs") pod "kube-apiserver-functional-20210112085224-2630" (UID: "dbabc6fcaa35495e1aa10964d680d780")
2021-01-12T08:54:32.4356393Z         	* Jan 12 08:54:23 functional-20210112085224-2630 kubelet[5095]: I0112 08:54:23.848141    5095 reconciler.go:224] operationExecutor.VerifyControllerAttachedVolume started for volume "etc-ca-certificates" (UniqueName: "kubernetes.io/host-path/a3e7be694ef7cf952503c5d331abc0ac-etc-ca-certificates") pod "kube-controller-manager-functional-20210112085224-2630" (UID: "a3e7be694ef7cf952503c5d331abc0ac")
[...]

PR #10250, opened 2021-01-25
add resource limit for my sql pod in case it cause oom on node

azhao155:
We do see flakiness of test due to OOM issue, so add limit for mysql pod which consume the most memory in the integration test.

The flakiness test issue:https://github.com/kubernetes/minikube/issues/10130


azhao155:
@medyagh

PR #10251, opened 2021-01-25
Update addons.md

andrzejsydor:
fix typos - should be an upper case like in other commands.


priyawadhwa (approved):
Thanks for fixing this!

PR #10252, opened 2021-01-25
Update _index.md

anastaciocintra:
the links of latest downloads are mixed and/or does not exist

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -25,7 +25,24 @@ All you need is Docker (or similarly compatible) container or a Virtual Machine 
 
 For Linux users, we provide 3 easy download options (for each architecture):
 
-### x86
+## x86
+
+
+```shell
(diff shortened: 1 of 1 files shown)


anastaciocintra:
ok, CLA was signed

anastaciocintra:
I signed it

afbjorklund (changes requested):
Note: amd64 and x86_64 is the **same** architecture, so there is nothing to be changed here

```
e312901e12c347d0e4eec74d94b8d75512943eb62479b441bb1332f05cde0d09  minikube-linux-amd64
e312901e12c347d0e4eec74d94b8d75512943eb62479b441bb1332f05cde0d09  minikube-linux-x86_64
```

Thanks for reminding us that we don't need to upload the extra binary ("x86_64), though.

afbjorklund:
The headlines have been updated, and the extra copies removed. So there is nothing more to do here.

* amd64 / x86_64

* arm64 / aarch64

Issue #8284, opened 2020-05-27
losetup in buildroot VM does not support -j: MapBlockVolume failed: exit status 32


anencore94:
**Steps to reproduce the issue:** 

1. Create pv, pvc, pod with `spec.VolumeMode:Block`
  - the same yaml file succeed with another k8s clusters installed with kubespray and kubeadm.
  - also, with the same yaml only different with `spec.VolumeMode:Filesystem` was succeed

2. However, creating pod failed in minikube:
pod stuck in containerCreating status, but the error message was different depends on minikube version (error message in `describe pod` and `journalctl -u kubelet` and `minikube log` was similar, so added only describing pod events log)

- in minikube v1.10.1 (k8s v1.18.2)
```
Events:
  Type     Reason           Age    From               Message
  ----     ------           ----   ----               -------
  Normal   Scheduled        5m25s  default-scheduler  Successfully assigned default/busybox-vol to minikube
  Warning  FailedMapVolume  5m25s  kubelet, minikube  MapVolume.MapBlockVolume failed for volume "example-pv-volume" : blkUtil.MapDevice failed. devicePath: /home/docker/aaa, globalMapPath:/var/lib/kubelet/plugins/kubernetes.io~local-volume/volumeDevices/example-pv-volume, podUID: dd860442-0b3f-4dd3-bf13-7994f0d93b80, bindMount: true: failed to bind mount devicePath: /home/docker/aaa to linkPath /var/lib/kubelet/plugins/kubernetes.io~local-volume/volumeDevices/example-pv-volume/dd860442-0b3f-4dd3-bf13-7994f0d93b80: mount failed: exit status 32
Mounting command: systemd-run
Mounting arguments: --description=Kubernetes transient mount for /var/lib/kubelet/plugins/kubernetes.io~local-volume/volumeDevices/example-pv-volume/dd860442-0b3f-4dd3-bf13-7994f0d93b80 --scope -- mount  -o bind /home/docker/aaa /var/lib/kubelet/plugins/kubernetes.io~local-volume/volumeDevices/example-pv-volume/dd860442-0b3f-4dd3-bf13-7994f0d93b80
Output: Running scope as unit: run-r3be8748ee1704f7d9b575e971e6acc8b.scope
mount: /var/lib/kubelet/plugins/kubernetes.io~local-volume/volumeDevices/example-pv-volume/dd860442-0b3f-4dd3-bf13-7994f0d93b80: mount point is not a directory.
[...]

PR #10255, opened 2021-01-25
install losetup from util-linux in the ISO

nixpanic:
The losetup tool is used by Kubernetes to setup BlockMode volumes.
BusyBox provides a version, but it does not have the `-j` option.

Fixes: #8284
See-also: tested with ceph/ceph-csi#1840


nixpanic:
This is ready for review, including it in a next release would be much appreciated!

/assign @tstromberg

afbjorklund (approved):

Issue #10204, opened 2021-01-21
Host access Documentation Problem


bradurani:
Hi, the procedure outlined for connecting to the host in the documentation at: https://minikube.sigs.k8s.io/docs/handbook/host-access/
does not work because neither `ping` nor `telnet` are installed on the minikube docker image
```
docker@minikube:~$ ping host.minikube.internal
-bash: ping: command not found
docker@minikube:~$ telnet host.minikube.internal
-bash: telnet: command not found
```

priyawadhwa:
Hey @bradurani thanks for opening this issue. We can fix this by installing `iputils-ping` in our docker base image here:

https://github.com/kubernetes/minikube/blob/master/deploy/kicbase/Dockerfile#L96

If you, or anyone else is interested in adding this and submitting a PR, please feel free to comment /assign on this issue

bradurani:
PR opened: https://github.com/kubernetes/minikube/pull/10262

spowelljr:
@bradurani You seem to have solved this issue with your PR #10262 so I'm going to close this issue, thanks for your contribution!

PR #10262, opened 2021-01-25
Install ping and telnet in the Docker base image.

bradurani:
This allows performing the host access procedure outlined in the docs at
https://minikube.sigs.k8s.io/docs/handbook/host-access/

fixes: #10204


medyagh:
sorry this needs CLA to be signed @bradurani

PR #10277, opened 2021-01-27
Use gol.16 in gh actions pr workflow

ilya-zuyev:

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
-          go-version: '1.15.5'
+          go-version: '1.16beta1'
           stable: true
       - name: Download Dependencies
         run: go mod download


.github/workflows/pr.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
-          go-version: '1.15.5'
+          go-version: '1.16beta1'
```
medyagh, line 22: please use the update go version in the hack folder
to update it everywhere in master.yml and also the jenkins scripts

medyagh (changes requested):

.github/workflows/pr.yml
```diff
@@ -19,7 +19,7 @@ jobs:
...
           stable: true
```
medyagh, line 23: the stable needs to be changed to false for beta
example in documentation https://github.com/actions/setup-go

medyagh (changes requested):

PR #10279, opened 2021-01-27
wip:test: make componentHealth test not parallel

azhao155:
We do see flakiness of test due to node restart issue, so add wait=true for the second start so all components are up during parallel tests.
The flakiness test issue:#10130


azhao155:
@medyagh

medyagh:
it is still failing on gh aciton
try increasing memory for funcitoanl test, maybe that would be helpful,

medyagh (approved):

PR #10286, opened 2021-01-27
Revert "install losetup from util-linux in the ISO"

medyagh:
Reverts kubernetes/minikube#10255

this causes the ISO build to fail


spowelljr (approved):

nixpanic:
This is really unfortunate. Even with this PR, **build_test_iso** failed. Manual builds with #10255 succeed for me :confused: 

@medyagh or @spowelljr , what is your recommendation to get `losetup` from `util-linux` included in the ISO?

sharifelgamal:
@nixpanic, could you open up another PR adding it back in? we're working on a new ISO testing framework that should be able to handle this now.

Issue #10289, opened 2021-01-28
Minikube start with `ssh` as driver ignores `--ssh-key` parameter


songford:
**Steps to reproduce the issue:** 

1. `minikube start --driver=ssh --ssh-ip-address=192.168.233.5 --ssh-key='~/.ssh/id_rsa' --ssh-user=songford --alsologtostderr -v=4`

**Full output of failed command:** 
```
😄  minikube v1.17.0 on Darwin 11.1
✨  Using the ssh driver based on existing profile
👍  Starting control plane node minikube in cluster minikube
🏃  Updating the running ssh "minikube" bare metal machine ...
🤦  StartHost failed, but will try again: provision: fast detect: OS type not recognized
🔄  Restarting existing ssh bare metal machine for "minikube" ...
```

With `--alsologtostderr -v=4`:
```
I0128 11:29:34.855887   80187 out.go:229] Setting OutFile to fd 1 ...
I0128 11:29:34.856501   80187 out.go:281] isatty.IsTerminal(1) = true
I0128 11:29:34.856510   80187 out.go:242] Setting ErrFile to fd 2...
I0128 11:29:34.856516   80187 out.go:281] isatty.IsTerminal(2) = true
I0128 11:29:34.856621   80187 root.go:280] Updating PATH: /Users/I342328/.minikube/bin
I0128 11:29:34.857016   80187 out.go:236] Setting JSON to false
I0128 11:29:34.903806   80187 start.go:106] hostinfo: {"hostname":"C02CP0ZYMD6R","uptime":487923,"bootTime":1611316651,"procs":641,"os":"darwin","platform":"darwin","platformFamily":"Standalone Workstation","platformVersion":"11.1","kernelVersion":"20.2.0","kernelArch":"x86_64","virtualizationSystem":"","virtualizationRole":"","hostId":"82e2050c-5936-3d24-ad3b-ec4ec5c09e11"}
W0128 11:29:34.903942   80187 start.go:114] gopshost.Virtualization returned error: not implemented yet
I0128 11:29:34.977337   80187 out.go:119] 😄  minikube v1.17.0 on Darwin 11.1
😄  minikube v1.17.0 on Darwin 11.1
I0128 11:29:34.977609   80187 notify.go:126] Checking for updates...
I0128 11:29:34.979161   80187 ssh_runner.go:149] Run: systemctl --version
[...]

PR #10293, opened 2021-01-28
ssh: validate the ssh-key parameter if given

afbjorklund:
Fixes the old (2016) TODO* from docker-machine, and might help to troubleshoot #10289

\* https://github.com/docker/machine/blob/bd45ab13d88c32a3dd701485983354514abc41fa/drivers/generic/generic.go#L114

BEFORE:

🤦  StartHost failed, but will try again: creating host: create: creating: usermod: NewSession: new client: new client: Error creating new native config from ssh using: vagrant, &{[] [/home/anders/.minikube/machines/minikube/bogus]}: ssh: no key found

AFTER:

`precreate: SSH key does not parse: "/tmp/bogus": ssh: no key found`

### DIFF
--- a/pkg/drivers/ssh/ssh.go
+++ b/pkg/drivers/ssh/ssh.go
@@ -18,12 +18,15 @@ package ssh
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net"
 	"os"
 	"os/exec"
 	"path"
 	"strconv"
 	"time"
+
+	"golang.org/x/crypto/ssh"
 
 	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/engine"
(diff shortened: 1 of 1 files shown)


pkg/drivers/ssh/ssh.go
```diff
@@ -102,6 +105,16 @@ func (d *Driver) PreCreateCheck() error {
 		if _, err := os.Stat(d.SSHKey); os.IsNotExist(err) {
 			return fmt.Errorf("SSH key does not exist: %q", d.SSHKey)
 		}
+
+		key, err := ioutil.ReadFile(d.SSHKey)
+		if err != nil {
+			return err
```
spowelljr, line 111: Would it makes sense to add a message to the error?

afbjorklund: Not really expecting any I/O errors here, if the file exists.

afbjorklund:
Here are the current errors, if something unexpected happens while reading the key file:

```
🤦  StartHost failed, but will try again: creating host: create: precreate: open /tmp/inaccessible: permission denied
😿  Failed to start ssh bare metal machine. Running "minikube delete" may fix it: creating host: create: precreate: open /tmp/inaccessible: permission denied

❌  Exiting due to GUEST_PROVISION: Failed to start host: creating host: create: precreate: open /tmp/inaccessible: permission denied
```

medyagh (approved):

PR #10321, opened 2021-01-30
Move DeletePossibleKicLeftOver function to delete pkg

### LINKED ISSUE #8040: Move logic out of `cmd/delete.go` so that it can be called from other packages
`cmd/delete.go` has a lot of interesting logic for cleaning up that is not accessible to `machine/delete.go`, and thus can't be used for cleanup on failure. 

Specifically, it allows for cleanup of orphaned KIC drivers - but we should also handle cleanup for orphaned non-KIC drivers.

BLasan:
fixes #8040 

Moved DeletePossibleKicLeftOver function from cmd/minikube/delete.go to a new package "delete" in order to make the function accessible from other packages.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -96,6 +96,7 @@ require (
 	k8s.io/api v0.18.8
 	k8s.io/apimachinery v0.18.8
 	k8s.io/client-go v0.18.8
+	k8s.io/klog v1.0.0
 	k8s.io/klog/v2 v2.4.0
 	k8s.io/kubectl v0.0.0
 	k8s.io/kubernetes v1.18.5
--- a/go.mod
+++ b/go.mod
@@ -96,7 +96,6 @@ require (
 	k8s.io/api v0.18.8
 	k8s.io/apimachinery v0.18.8
 	k8s.io/client-go v0.18.8
-	k8s.io/klog v1.0.0
 	k8s.io/klog/v2 v2.4.0
 	k8s.io/kubectl v0.0.0
 	k8s.io/kubernetes v1.18.5
--- a/pkg/minikube/delete/delete.go
+++ b/pkg/minikube/delete/delete.go
@@ -21,7 +21,7 @@ import (
 	"fmt"
 	"os/exec"
 
-	"k8s.io/klog"
+	"k8s.io/klog/v2"
 	"k8s.io/minikube/pkg/drivers/kic/oci"
 	"k8s.io/minikube/pkg/minikube/driver"
 	"k8s.io/minikube/pkg/minikube/out"
--- a/pkg/minikube/delete/delete.go
+++ b/pkg/minikube/delete/delete.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package profile
+package delete
 
 import (
 	"context"
@@ -28,8 +28,8 @@ import (
 	"k8s.io/minikube/pkg/minikube/style"
 )
 
-//DeletePossibleLeftOvers ...
-func DeletePossibleLeftOvers(ctx context.Context, cname string, driverName string) {
+//PossibleLeftOvers deletes KIC & non-KIC drivers left
+func PossibleLeftOvers(ctx context.Context, cname string, driverName string) {
 	bin := ""
 	switch driverName {
 	case driver.Docker:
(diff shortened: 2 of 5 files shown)


BLasan:
@medyagh @tstromberg  Please review

medyagh (changes requested):
I dont agree with the naming, the package naming should not stutter.

delete.DeletePossibe... doesn't sounds right

BLasan:
> I dont agree with the naming, the package naming should not stutter.
> 
> delete.DeletePossibe... doesn't sounds right

Yuh, is it ok to change the function name as well? I'll change the function name in to a meaningful one :)

BLasan:
> I dont agree with the naming, the package naming should not stutter.
> 
> delete.DeletePossibe... doesn't sounds right

Renamed. Please review

BLasan:
@medyagh Please review

BLasan:
@medyagh Sir, could you please verify this? If there are changes need to be done I would like to add them as well :)

BLasan:
@afbjorklund Sir, could you please verify these changes?

medyagh (changes requested):
@BLasan what do u think about making a new package that is a noun

for example "profile"

so the could would read as

profile.DeletePossibleLeftovers()

BLasan:
> @BLasan what do u think about making a new package that is a noun
> 
> for example "profile"
> 
> so the could would read as
> 
> profile.DeletePossibleLeftovers()

That sounds ideal sir! Before that I think we need to merge new changes done in the issue #10332. According to your suggestion above, isn't it good if we could define all the profile related functions(which are defined in "cmd/delete.go" file) under "profile" package?

BLasan:
@medyagh @priyawadhwa Resolved the merge conflicts. Kindly take a look

pkg/minikube/profile/profile.go
```diff
@@ -0,0 +1,81 @@
...
+	"os/exec"
+
+	"k8s.io/klog"
+	"k8s.io/minikube/pkg/drivers/kic/oci"
+	"k8s.io/minikube/pkg/minikube/driver"
+	"k8s.io/minikube/pkg/minikube/out"
+	"k8s.io/minikube/pkg/minikube/style"
+)
+
+//DeletePossibleLeftOvers ...
```
priyawadhwa, line 31: can you fill in this comment?

pkg/minikube/profile/profile.go
```diff
@@ -0,0 +1,81 @@
+/*
```
priyawadhwa, line 1: Let's name this file `delete.go` instead of `profile.go`

priyawadhwa (commented):
Left a couple comments, this is pretty close to finished!

pkg/minikube/profile/profile.go
BLasan: Sure

BLasan:
@priyawadhwa Changes added. Please review ma'am

BLasan:
@priyawadhwa Any remaining things to be added ma'am?

go.mod
```diff
@@ -96,6 +96,7 @@ require (
 	k8s.io/api v0.18.8
 	k8s.io/apimachinery v0.18.8
 	k8s.io/client-go v0.18.8
+	k8s.io/klog v1.0.0
```
priyawadhwa, line 99: We want to use k8s.io/klog/v2 through the codebase, so to fix this I would:

1. Replace any imports of `k8s.io/klog` with `k8s.io/klog/v2`
2. Run `go mod tidy` to remove this dependency from go.mod

pkg/minikube/delete/delete.go
```diff
@@ -0,0 +1,81 @@
...
+*/
+
+package delete
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+
+	"k8s.io/klog"
```
priyawadhwa, line 24: this should be k8s.io/klog/v2

priyawadhwa (commented):
One last comment

BLasan:
@priyawadhwa Thanks for the review ma'am. Made the required changes

priyawadhwa (approved):
LGTM, thank you!

BLasan:
@medyagh @priyawadhwa Any remaining works?

Issue #10331, opened 2021-02-01
Add timeout to `minikube delete`


priyawadhwa:
If we haven't successfully deleted our cluster in 3 minutes, move on to the next cluster

Hanging on `minikube delete` is a poor experience for the user and has been forcing our integration tests to timeout.

PR #10332, opened 2021-02-01
 Add 5 minute timeout to deleting a profile

priyawadhwa:
Right now if the docker daemon is stuck then `minikube delete` will hang forever, which is a really poor user experience and also causes our integration tests to timeout in a pretty bad way.

I fixed this by passing in a `context.Context` to our deletion `exec.Cmd` calls, with a 5 minute timeout per profile. Since we didn't previously pass in context anywhere, I did have to add it into a few different files, so this change was bigger than initially expected.

Fixes https://github.com/kubernetes/minikube/issues/10331

### DIFF
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -166,6 +166,8 @@ func runDelete(cmd *cobra.Command, args []string) {
 		if len(args) > 0 {
 			exit.Message(reason.Usage, "usage: minikube delete")
 		}
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+		defer cancel()
 
 		cname := ClusterFlagValue()
 		profile, err := config.LoadProfile(cname)
@@ -185,8 +187,8 @@ func runDelete(cmd *cobra.Command, args []string) {
 
 		if orphan {
 			// TODO: generalize for non-KIC drivers: #8040
-			deletePossibleKicLeftOver(cname, driver.Docker)
-			deletePossibleKicLeftOver(cname, driver.Podman)
+			deletePossibleKicLeftOver(ctx, cname, driver.Docker)
+			deletePossibleKicLeftOver(ctx, cname, driver.Podman)
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


medyagh (changes requested):
I suspect the time out was happening in the "delete --all"
where we call to delete containers by a label (orpahned kic containers)

additional
https://github.com/medyagh/minikube/blob/72443005764661ff12dafedee6c0afed4829b379/cmd/minikube/cmd/delete.go#L151

```
if deleteAll {
		deleteContainersAndVolumes(oci.Docker)
		deleteContainersAndVolumes(oci.Podman)

		errs := DeleteProfiles(profilesToDelete)

```

Additionaly we should ensure the Timeout Allows minikube to go ahead and delete next time.
for example if it doing this order:

- deleting an abadoned stuck container with kic label
- deleting a profile that has config (not abandoned)
- deleting another profile (not abandoned)


if the  abandoned one times out it should sitll go delete next ones

medyagh (approved):

BLasan:
@medyagh Moved DeletePossibleKicLeftOver function #10321

medyagh (changes requested):
lets ensure to add time out for 			deletePossibleKicLeftOver(cname, driver.Docker)
			deletePossibleKicLeftOver(cname, driver.Podman)

priyawadhwa:
@medyagh all functions should be covered now.

cmd/minikube/cmd/delete.go
```diff
@@ -108,7 +110,8 @@ func deleteContainersAndVolumes(ociBin string) {
 		klog.Infof("error delete containers by label %q (might be okay): %+v", delLabel, errs)
 	}
 
-	errs = oci.DeleteAllVolumesByLabel(ociBin, delLabel)
+	ctx := context.Background()
```
medyagh, line 113: nit: how about a name that is more clear this is context for delete ? maybe `delCtx` ?

medyagh: does this one need a timeout too ?

medyagh (changes requested):

cmd/minikube/cmd/delete.go
priyawadhwa: thanks, fixed

medyagh (approved):

medyagh (approved):
looks good after Linux KVM test is normal, feel free to merge.
thanks @priyawadhwa

PR #10341, opened 2021-02-02
(docs) update install instructions for MacOS  

vladdoster:
- cask is now denoted as a flag `--cask`

Removing this now doesn't make much sense. I think it should be an attempt to unlink minikube:

```shell
brew unlink minikube
```

 and linking it again:

```shell
brew link minikube
```

image

# Before

image

# After

image

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -81,10 +81,10 @@ If the [Brew Package Manager](https://brew.sh/) installed:
 brew install minikube
 ```
 
-If `which minikube` fails after installation via brew, you may have to remove the minikube cask and link the binary:
+If `which minikube` fails after installation via brew, you may have to remove the old minikube links and link the newly installed binary:
 
 ```shell
-brew cask remove minikube
+brew unlink minikube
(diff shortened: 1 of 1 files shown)


site/content/en/docs/start/_index.md
```diff
@@ -81,10 +81,10 @@ If the [Brew Package Manager](https://brew.sh/) installed:
 brew install minikube
 ```
 
-If `which minikube` fails after installation via brew, you may have to remove the minikube cask and link the binary:
+If which minikube fails after installation via brew, you may have to remove the old minikube links and link the newly installed binary:
```
priyawadhwa, line 84: Can you keep the backticks around `which minikube` to indicate that it is a command to run

priyawadhwa (changes requested):
Thanks for improving this! Just left one comment.

vladdoster:
> Thanks for improving this! Just left one comment.

Thank you for catching that. I have added back `` around `which minikube`.

priyawadhwa (approved):
LGTM, thanks for fixing this!

Issue #10254, opened 2021-01-25
Don't upload multiple copies of the same file to release


afbjorklund:
Someone was confused about having **both** `minikube-linux-amd64` and `minikube-linux-x86_64` files ?

Turns out it has been duplicated since v1.10.0, after merging 056af16aabdfc35f895566cb5870433e98506627 which started uploaded _everything_

* https://github.com/kubernetes/minikube/releases/tag/v1.9.0
* https://github.com/kubernetes/minikube/releases/tag/v1.10.0

Duplicates: (without `.sha256` checksum file)
* minikube-linux-x86_64
* docker-machine-driver-kvm2-x86_64
* minikube-linux-aarch64
* minikube-windows-amd64

afbjorklund:
It was noticed during PR #10252 (which is otherwise wrong)

afbjorklund:
The "latest" packages serve a special purpose (not having to update documentation), so are not affected by this issue.

Issue #6159, opened 2019-12-25
Add installation documentation for ARM64


chendave:
The guide here: https://minikube.sigs.k8s.io/docs/start/linux/ seems for X86 only, the deb package I get is not supported on ARM64,

```
# dpkg -i minikube_1.6.2.deb
dpkg: error processing archive minikube_1.6.2.deb (--install):
 package architecture (amd64) does not match system (arm64)
Errors were encountered while processing:
 minikube_1.6.2.deb
```

Is there any guide for installation for ARM64 and the deb is ready?

Thanks a lot!

afbjorklund:
There is no guide just yet, since the support is still rather experimental (#5667) - still lacking a couple of images, for instance. So the support for building the deb was added, but we haven't started using it just yet. However, I think the binary is posted to GCS (even if not added to GitHub): [minikube-linux-arm64](https://storage.googleapis.com/minikube/releases/v1.6.2/minikube-linux-arm64)

Usage should be very similar to normal, but some container images _only_ have support for amd64. Originally Docker didn't have support for multiple architectures, but it has been added to newer images. You will need a powerful CPU like the Raspberry Pi 4 to run with `kvm2`, otherwise use `none`.

The main problem is the 2G memory requirement, which might mean having to add a swapfile...
There was also heat issues (overhearing) on the earlier modules, due to processing requirements.

But with the addition of more powerful ARM processors, it should work fine to run regular kubernetes.
So it will be regular deployment in a future version of minikube, with a `minikube_1.6.2-0_arm64.deb`

afbjorklund:
Added a reminder to add the architecture to the package filenames in #6281

The main "problem" at the moment is that we **only** support the server use case... The main desktop OS solution (Raspbian) is using `arm` rather than `arm64`, and even though KVM is supported we don't have an ISO for that architecture. Same with the new `docker` driver (KIC), no ARM images for that. The kubeadm requirements (2 CPU, 4G RAM) also puts it at the high end of the available hardware.

So for the end user, the problems are similar to when running on a virtual desktop. And the total lack of isolation in the `none` driver still makes it non-recommended to use on a regular machine, and the use of `sudo` even makes it impractical  for the client (since all the files end up being owned by root). The biggest long-term benefit of this form factor ([RPi](https://www.raspberrypi.org/)) also comes when running _multiple_ such nodes locally.

But for trying it out, things "should work" now - on both the RPi3b and RPi4b.

[...]

PR #10347, opened 2021-02-02
Changed x86 to amd64 in installation guide to reflect actual architecture name

mwht:
Minikube installation guide has multiple architectures to choose from - including "x86".
Actual links however lead to amd64 versions of minicube. This might mislead potential x86 users into thinking they are downloading package for 32-bit version of Linux.

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -25,7 +25,7 @@ All you need is Docker (or similarly compatible) container or a Virtual Machine 
 
 For Linux users, we provide 3 easy download options (for each architecture):
 
-### amd64
+### amd64 / x86_64
 
 #### Binary download
 
@@ -49,7 +49,7 @@ curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-latest
 sudo rpm -ivh minikube-latest.x86_64.rpm
 ```
 
-### ARM64
+### arm64 / aarch64
 
 #### Binary download
 
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -25,7 +25,7 @@ All you need is Docker (or similarly compatible) container or a Virtual Machine 
 
 For Linux users, we provide 3 easy download options (for each architecture):
 
-### x86
+### amd64
 
 #### Binary download
 
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -49,7 +49,7 @@ curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-latest
 sudo rpm -ivh minikube-latest.x86_64.rpm
 ```
 
-### ARM
+### ARM64
 
 #### Binary download
 


mwht:
Registered on CLA

mwht:
I signed it

site/content/en/docs/start/_index.md
```diff
@@ -25,7 +25,7 @@ All you need is Docker (or similarly compatible) container or a Virtual Machine
 
 For Linux users, we provide 3 easy download options (for each architecture):
 
-### x86
+### amd64
```
afbjorklund, line 28: Might as well change both, if you are at it:

x86 --> amd64 (x86_64)

ARM -> arm64 (aarch64)

Do people still run 32-bit Intel ? Wow.

afbjorklund (changes requested):

site/content/en/docs/start/_index.md
mwht: Good point. New changes pushed.

mwht: And yes, it happens sometimes. Edge case I suppose :)

afbjorklund:
@ilya-zuyev
We should do the same for the Mac binaries too, that is for the operating system known as "Darwin"

PR #10337

site/content/en/docs/start/_index.md
afbjorklund: People do run 32-bit ARM (arm), but we don't advertise it. The install page would look awful, with all 5 architectures...

site/content/en/docs/start/_index.md
```diff
@@ -49,7 +49,7 @@ curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-latest
 sudo rpm -ivh minikube-latest.x86_64.rpm
 ```
 
-### ARM
+### ARM64
```
afbjorklund, line 52: Should make it consistent, i.e. "arm64". Probably throw in the other names as well, they are used by the .rpm

afbjorklund (changes requested):

site/content/en/docs/start/_index.md
mwht: How about going with "arm64 / aarch64" and "amd64 / x86_64"? Actual artifact is named "minikube-latest.aarch64.rpm" on storage - same applies to amd64 - RPM artifact is named "x86_64", whereas both refer to the same architecture (arm64 and aarch64 to ARM64, x86_64 and arm64 to 64-bit version of x86 with AMD instruction set)

afbjorklund: Sounds good

afbjorklund (approved):

medyagh (commented):
@mwht I know not related to this PR but how do u feel about adding a Tab or Drop Down to choose your CPU ... I feel like this is getting so long and not as easy to follow.

this is mean to be quick start

afbjorklund:
> I know not related to this PR but how do u feel about adding a Tab or Drop Down to choose your CPU ... I feel like this is getting so long and not as easy to follow.

Currently our headlines say:
* **x86**
* **ARM**

After adding https://github.com/kubernetes/minikube/commit/318bdcd7881814c940f1c203f98ae23dc4dfdc0f

That _could_ be interpreted as 32-bit.

Want to make it clear, what the arch is...

* **amd64 / x86_64**
* **arm64 / aarch64**

----

Nested tabs can be rather ugly and confusing, but one alternative is putting each architecture on a separate tab:

* Linux (amd64/x86_64)
* Linux (arm64/aarch64)
* macOS (Intel)
* macOS (M1)
* Windows

Another option is to make three separate headings, like on https://kubernetes.io/docs/tasks/tools/install-kubectl/ :

1a) Installation on Linux
1b) Installation on macOS
1c) Installation on Windows
2) Start your cluster
3) Interact with your cluster

We have been talking about this since #6159, but so far there are **no** design suggestions (like dropdowns or such) ?

So far we have only included 2 out of 5 architectures, and didn't bother including checking the sha256 checksums etc.

* **amd64**
* arm
* **arm64**
* ppc64le
* s390x

I opened an issue to remove the duplicate binaries currently present under releases and in storage, as #10254

Ultimately we would want it to be as nice as when using the package managers to do it, like so: "install minikube".

afbjorklund:
Opened #10353 for improving the appearance (and not just the content)

PR #10354, opened 2021-02-03
Provide advice when running windows binary (.exe) inside WSL

### LINKED ISSUE #10311: running minikube.exe in WSL looks for home in C:\...
when minikube.exe runs in Powershell it will create a MINIKUBE_HOME in C:\Program ...

but when the same binary is called in WSL (windows sub linux) it looks for minikube home in linux Path ~/.minikube 
this causes confusion ! 

for that reason we need to either, not allow running .exe binary in WSL and make suer make a choice of downloading Linux binary on WSL or .exe binary needs to do the right thing if it is run in WSL

I suggest as a start we disallow running the .exe binary WSL (unless --force is applied) 

as seen in https://github.com/GoogleCloudPlatform/cloud-code-vscode/issues/338

hetong07:
Fixes #10311 

Throws out MK_WRONG_BINARY_WSL error if user tries to run windows binary (.exe) inside WSL and advises them either to run with '--force' or download correct binary.

**Before PR**
```
#  Inside WSL on windows 10
$  ./minikube-windows-amd64.exe start --dry-run
😄  minikube v1.17.1 on Microsoft Windows 10 Education 10.0.19041 Build 19041
✨  Automatically selected the docker driver. Other choices: virtualbox, ssh
🌵  dry-run validation complete!
```

**After PR**
```
# Inside WSL on windows 10

$ ./minikube-windows-amd64.exe start --dry-run --force
😄  minikube v1.17.1 on Microsoft Windows 10 Education 10.0.19041 Build 19041
❗  minikube skips various validations when --force is supplied; this may lead to unexpected behavior
✨  Automatically selected the docker driver. Other choices: virtualbox, ssh
🌵  dry-run validation complete!

$ ./minikube-windows-amd64.exe start --dry-run

❌  Exiting due to MK_WRONG_BINARY_WSL: You are trying to run windows .exe binary inside WSL, for better integration please use Linux binary instead (Download at https://minikube.sigs.k8s.io/docs/start/.). Otherwise if you still want to do this, you can do it using --force
```

### DIFF
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/audit"
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/constants"
+	"k8s.io/minikube/pkg/minikube/driver"
 	"k8s.io/minikube/pkg/minikube/exit"
 	"k8s.io/minikube/pkg/minikube/localpath"
 	"k8s.io/minikube/pkg/minikube/out"
@@ -97,6 +98,17 @@ func Execute() {
 			os.Args = append([]string{RootCmd.Use, callingCmd, profile, "--"}, os.Args[1:]...)
 		} else {
 			os.Args = append([]string{RootCmd.Use, callingCmd, "--"}, os.Args[1:]...)
+		}
+	} else if filepath.Ext(callingCmd) == ".exe" && driver.IsMicrosoftWSL() {
+		var found = false
+		for _, a := range os.Args {
+			if a == "--force" {
+				found = true
+				break
+			}
+		}
+		if !found {
+			exit.Message(reason.WrongBinary, "Cannot run Windows binary inside WSL, please download linux binary from https://minikube.sigs.k8s.io/docs/start/. Or you can use '--force' to force execution which would be at your own risk.")
 		}
 	}
 	for _, c := range RootCmd.Commands() {
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -108,7 +108,7 @@ func Execute() {
 			}
 		}
 		if !found {
-			exit.Message(reason.WrongBinaryWSL, "You are trying to run windows .exe binary inside WSL, for better integration please use Linux binary instead (Download at https://minikube.sigs.k8s.io/docs/start/.). Otherwise if you still want to do this, you can do it using --force"
+			exit.Message(reason.WrongBinaryWSL, "You are trying to run windows .exe binary inside WSL, for better integration please use Linux binary instead (Download at https://minikube.sigs.k8s.io/docs/start/.). Otherwise if you still want to do this, you can do it using --force")
 		}
 	}
 	for _, c := range RootCmd.Commands() {
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -79,6 +79,20 @@ var RootCmd = &cobra.Command{
 func Execute() {
 	defer audit.Log(time.Now())
 
+	// Check whether this is a windows binary (.exe) running inisde WSL.
+	if runtime.GOOS == "windows" && driver.IsMicrosoftWSL() {
+		var found = false
+		for _, a := range os.Args {
+			if a == "--force" {
+				found = true
+				break
+			}
+		}
+		if !found {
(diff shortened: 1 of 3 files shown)


hetong07:
/retitle Disallow and give advice when running windows binary inside WSL

cmd/minikube/cmd/root.go
```diff
@@ -98,6 +99,17 @@ func Execute() {
 		} else {
 			os.Args = append([]string{RootCmd.Use, callingCmd, "--"}, os.Args[1:]...)
 		}
+	} else if filepath.Ext(callingCmd) == ".exe" && driver.IsMicrosoftWSL() {
+		var found = false
+		for _, a := range os.Args {
```
medyagh, line 104: I am curious why wouldn't we just use viper.GetBool(force) 
for example 	if !viper.GetBool(force) { 
}
https://github.com/medyagh/minikube/blob/446e0f0946f820fe612dfdda5d7b1a34d2b7b36f/cmd/minikube/cmd/start.go#L1258

cmd/minikube/cmd/root.go
```diff
@@ -98,6 +99,17 @@ func Execute() {
...
+			if a == "--force" {
+				found = true
+				break
+			}
+		}
+		if !found {
+			exit.Message(reason.WrongBinary, "Cannot run Windows binary inside WSL, please download linux binary from https://minikube.sigs.k8s.io/docs/start/. Or you can use '--force' to force execution which would be at your own risk.")
```
medyagh, line 111: You are trying to run windows .exe binary inside WSL, for better integration please use Linux binary instead. otherwise if you still want to do this, you can do it using --force"

pkg/minikube/reason/reason.go
```diff
@@ -190,6 +190,7 @@ var (
 
 	RsrcInsufficientStorage = Kind{ID: "RSRC_INSUFFICIENT_STORAGE", ExitCode: ExInsufficientStorage, Style: style.UnmetRequirement}
 
+	WrongBinary        = Kind{ID: "WRONG_BINARY", ExitCode: ExHostError}
```
medyagh, line 193: how about 
WRONG_BINARY_WSL

cmd/minikube/cmd/root.go
```diff
@@ -98,6 +99,17 @@ func Execute() {
 		} else {
 			os.Args = append([]string{RootCmd.Use, callingCmd, "--"}, os.Args[1:]...)
 		}
+	} else if filepath.Ext(callingCmd) == ".exe" && driver.IsMicrosoftWSL() {
```
medyagh, line 102: instead of checking for the extension of the file (.exe) how about we use runtime.GOOS

medyagh (changes requested):

cmd/minikube/cmd/root.go
hetong07: That was my first attempt, and it fails. It seems in here, the viper has not been initialized, so using `viper.GetBool(force)` cannot retrieve the flag value.

cmd/minikube/cmd/root.go
hetong07: done

cmd/minikube/cmd/root.go
hetong07: Yes, the runtime.GOOS shows it is windows even though it is inside WSL. thank you.

hetong07:
I feel like that we do not need to create a new family for error code since we are not expecting such error to happen frequently. So I choose to use the ExProgramUnsupported as the error code. Using windows binary on linux/WSL is indeed not a supported feature.

hetong07:
I signed it

cmd/minikube/cmd/root.go
```diff
@@ -98,6 +99,17 @@ func Execute() {
 		} else {
 			os.Args = append([]string{RootCmd.Use, callingCmd, "--"}, os.Args[1:]...)
 		}
+	} else if runtime.GOOS == "windows" && driver.IsMicrosoftWSL() {
```
medyagh, line 102: @hetong07 why does this have to be in an else clause? could it be just a root level if statement ?

medyagh (changes requested):

cmd/minikube/cmd/root.go
medyagh: any chance I could see example of this output in real WSL ?

hetong07: @medyagh Yes you are right. I just prefer to put all special case handler in one if-else tree. Since they are irrelevant, we should put it in a separate if-else.

The running demo is:
image

medyagh (approved):
Thank you for this PR

PR #10361, opened 2021-02-03
upgrade gcp-auth-webhook image to v0.0.4 for multiarch

### LINKED ISSUE #10325: ARM64: GCP Auth addon
add support for gcp-auth addon arm64

sharifelgamal:
Fixes #10325

### DIFF
--- a/deploy/addons/gcp-auth/gcp-auth-webhook.yaml.tmpl.tmpl
+++ b/deploy/addons/gcp-auth/gcp-auth-webhook.yaml.tmpl.tmpl
@@ -41,7 +41,7 @@ rules:
       - update
 
 ---
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: minikube-gcp-auth-certs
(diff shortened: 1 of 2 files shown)


deploy/addons/gcp-auth/gcp-auth-webhook.yaml.tmpl.tmpl
```diff
@@ -41,7 +41,7 @@ rules:
       - update
 
 ---
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
```
medyagh, line 44: which version of kubernetes has `apiVersion: rbac.authorization.k8s.io/v1` ?
would this mean older kubernetes versions wont be able to use this addon ?

medyagh (changes requested):

deploy/addons/gcp-auth/gcp-auth-webhook.yaml.tmpl.tmpl
sharifelgamal: v1beta1 has been deprecated since 1.17 and has started causing failures in newer versions of k8s, this is really the only answer here.

Issue #10365, opened 2021-02-05
Add `minikube image load` command


priyawadhwa:
This will load an image into the correct runtime but won't store it in the cache.

afbjorklund:
What is with the double command ? We don't call it `minikube image cache`

priyawadhwa:
This could be extended into `minikube image build` or `minikube image delete`. I agree that right now just `minikube load` would be easier but for future it could be nice to organize all image related content under one subcommand.

PR #10366, opened 2021-02-05
Implement minikube image load command

priyawadhwa:
And add an integration test for it.

closes https://github.com/kubernetes/minikube/issues/10365


pkg/minikube/machine/cache_images.go
```diff
@@ -61,6 +62,57 @@ func CacheImagesForBootstrapper(imageRepository string, version string, clusterB
 	return nil
 }
 
+// LoadImage loads the local image into the container runtime
+func LoadImage(profile, img string) error {
```
medyagh, line 66: how about we make this a re-usable function that Cache Add also uses same, so we don't have to maintain both Image Load and Cache Add ?

I suggest adding a --all label to "cache add" command so if that is passed it will load image to all clusters if not, it will add it to the active cluster.

test/integration/load_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+func TestImageLoad(t *testing.T) {
```
medyagh, line 29: lets not create new clusters just for a functional test, this could be a sub test of functional test
similar to Cache Add Test in functional test

medyagh (changes requested):

cmd/minikube/cmd/image.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"k8s.io/minikube/pkg/minikube/config"
+	"k8s.io/minikube/pkg/minikube/exit"
+	"k8s.io/minikube/pkg/minikube/machine"
+	"k8s.io/minikube/pkg/minikube/reason"
+)
+
+// imageCmd represents the image command
+var imageCmd = &cobra.Command{
```
afbjorklund, line 29: It would be simpler if "load" was a top-level command, like all other minikube commands

priyawadhwa: Just commented this on the issue as well:

>This could be extended into minikube image build or minikube image delete. I agree that right now just minikube load would be easier but for future it could be nice to organize all image related content under one subcommand.

medyagh: @afbjorklund the reason for adding "image load" is we need other commands 
```
image delete
image list
image build
```

without a sub command it woudl be hard to add other operations for image
we could add a "hidden" alias "load"

medyagh (approved):

PR #10381, opened 2021-02-06
Refactor tmp dir detection to improve testability

spowelljr:
Testing to prevent: #10407

In its current state, testing the temp directory detection logic is quite difficult to do due to the size of the function the logic is nested within. Due to the difficulty of the testing, tests have not been implemented and resulted in a bug (#10372) being merged that should easily be caught with tests. This PR refactors the temp directory setting logic into its own function, allowing unit tests to easily be written for it. Unit tests are included in this PR that will prevent the bug from being merged again.

### DIFF
--- a/pkg/minikube/command/kic_runner.go
+++ b/pkg/minikube/command/kic_runner.go
@@ -186,17 +186,13 @@ func (k *kicRunner) Copy(f assets.CopyableFile) error {
 		}
 	}
 	klog.Infof("%s (temp): %s --> %s (%d bytes)", k.ociBin, src, dst, f.GetLength())
-	tmpFolder := ""
 
-	// Snap only allows an application to see its own files in /tmp, making Docker unable to copy memory assets
-	// https://github.com/kubernetes/minikube/issues/10020
-	if isSnapBinary() {
-		home, err := os.UserHomeDir()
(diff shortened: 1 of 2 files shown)


pkg/minikube/command/kic_runner.go
```diff
@@ -209,6 +205,24 @@ func (k *kicRunner) Copy(f assets.CopyableFile) error {
 	return k.copy(tf.Name(), dst)
 }
 
+// tempDirectory returns the directory to use as the temp directory
+// or an empty string if it should use the os default temp directory.
+func tempDirectory(isSnap bool) (string, error) {
+	if !isSnap {
```
medyagh, line 211: how about instead of passing the the bool, we just call snap detection inside the func ?
also lets choose a better name
how about ?
```
func tmpDirPath () (string, error)
 ```

medyagh (changes requested):

pkg/minikube/command/kic_runner.go
spowelljr: The problem with that is it makes it very hard to test the snap path, which is really the whole point of this PR.

medyagh (approved):

PR #10384, opened 2021-02-06
remove podman cni config when custom cni is used

michaelhenkel:
When a custom CNI is defined via --cni flag it might take a while to populate the CNIs configuration into /etc/cni/net.d.
Minikube comes up and in case of crio or containerd runtime it has a CNI configuration in /etc/cni/net,d already.
Now if pods using pod network (such as coredns) come up BEFORE the custom CNIs config file is populated, this pod ip address is served by podman. Subsequent pods will be served by the custom CNI but won't be able to communicate with the pods using podman CNI.
This PR removes all CNI configs from /etc/cni/net.d just before the custom CNI manifest is applied.

### DIFF
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -351,7 +351,8 @@ func (k *Bootstrapper) applyCNI(cfg config.ClusterConfig, registerStep ...bool) 
 		return errors.Wrap(err, "cni apply")
 	}
 
-	if cfg.KubernetesConfig.ContainerRuntime == constants.CRIO {
+	_, custom := cnm.(cni.Custom)
+	if cfg.KubernetesConfig.ContainerRuntime == constants.CRIO && !custom {
 		if err := cruntime.UpdateCRIONet(k.c, cnm.CIDR()); err != nil {
 			return errors.Wrap(err, "update crio")
 		}
--- a/pkg/minikube/cni/custom.go
+++ b/pkg/minikube/cni/custom.go
@@ -21,8 +21,10 @@ import (
 	"path"
 
 	"github.com/pkg/errors"
+	"k8s.io/klog/v2"
 	"k8s.io/minikube/pkg/minikube/assets"
 	"k8s.io/minikube/pkg/minikube/config"
+	"k8s.io/minikube/pkg/minikube/sysinit"
 )
 
 // Custom is a CNI manager than applies a user-specified manifest
@@ -51,6 +53,9 @@ func NewCustom(cc config.ClusterConfig, manifest string) (Custom, error) {
 
 // Apply enables the CNI
 func (c Custom) Apply(r Runner) error {
+	if err := c.addFakeConfig(r); err != nil {
+		return errors.Wrap(err, "create fake cni config")
+	}
 	m, err := assets.NewFileAsset(c.manifest, path.Dir(manifestPath()), path.Base(manifestPath()), "0644")
 	if err != nil {
 		return errors.Wrap(err, "manifest")
@@ -62,4 +67,24 @@ func (c Custom) Apply(r Runner) error {
 // CIDR returns the default CIDR used by this CNI
 func (c Custom) CIDR() string {
 	return DefaultPodCIDR
+}
+
+// addFakeConfig adds a fake CNI configuration file to the node to prevent
+// podman CNI to serve ip addresses to pods in case 3rd party CNI is enabled
+func (c Custom) addFakeConfig(r Runner) error {
+	klog.Infof("creating fake CNI config")
+	fakeCNI := `{
+  "cniVersion": "0.4.0",
+  "name": "fake",
+  "type": "host-local"
+}`
+
+	b := []byte(fakeCNI)
+	fakeCNIAsset := assets.NewMemoryAssetTarget(b, "/etc/cni/net.d/86-fake.conf", "0644")
+	if err := r.Copy(fakeCNIAsset); err != nil {
+		return errors.Wrapf(err, "copy")
+	}
+
+	// crio service must be restarted otherwise the podman cni is still used
+	return sysinit.New(r).Restart(c.cc.KubernetesConfig.ContainerRuntime)
 }


pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -335,7 +335,8 @@ func (k *Bootstrapper) applyCNI(cfg config.ClusterConfig) error {
 		return errors.Wrap(err, "cni apply")
 	}
 
-	if cfg.KubernetesConfig.ContainerRuntime == constants.CRIO {
+	_, custom := cnm.(cni.Custom)
+	if cfg.KubernetesConfig.ContainerRuntime == constants.CRIO && !custom {
```
medyagh, line 339: I am curious why only crio, how about containerd, wouldn't we need to remove the the kindnet cni if user specifies the custom cni ?

medyagh (changes requested):

krezovic:
There is a chicken and egg problem with calico and containerd when this PR is merged. containerd requires valid CNI config to start up, and calico will use init container to install its cni config file and cni executable.

The only way I could workaround this is start with calico cni, wait until it's fully up by using "kubectl wait", delete the non-calico config file from /etc/cni/net.d and restart kube-proxy and coredns.

Not sure if other components need restart, as they use host IPs and are static pods anyway.

michaelhenkel:
hm, in my case I also use an init container to copy the CNI binary and
config to the host and containerd seems to be happy with it. I will test
with Calico next week (PTO this week).

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
sadlil: If user uses a custom cni, the `kindnet.conflist` file shouldn't be created. But even if kindnet is used the delay in config file creation is there. For CRIO - the kind kindnet file is created almost 17 seconds after the crio-bridge and podman config file. See - 

```
root@minikube:/etc/cni/net.d# stat 10-kindnet.conflist
  File: 10-kindnet.conflist
  Size: 391       	Blocks: 8          IO Block: 4096   regular file
Device: 74h/116d	Inode: 3288136     Links: 1
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2021-02-26 22:36:37.270550000 +0000
Modify: 2021-02-26 22:36:37.241550000 +0000
Change: 2021-02-26 22:36:37.243550000 +0000
 Birth: -


root@minikube:/etc/cni/net.d# stat 100-crio-bridge.conf
  File: 100-crio-bridge.conf
  Size: 438       	Blocks: 8          IO Block: 4096   regular file
Device: 74h/116d	Inode: 3288094     Links: 1
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2021-02-26 22:36:20.385550000 +0000
Modify: 2021-02-26 22:36:20.356550000 +0000
Change: 2021-02-26 22:36:20.356550000 +0000
 Birth: -


root@minikube:/etc/cni/net.d# stat 87-podman-bridge.conflist
  File: 87-podman-bridge.conflist
  Size: 641       	Blocks: 8          IO Block: 4096   regular file
Device: 74h/116d	Inode: 3288095     Links: 1
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2021-02-26 22:36:20.365550000 +0000
Modify: 2021-02-26 22:36:20.356550000 +0000
Change: 2021-02-26 22:36:20.356550000 +0000
 Birth: -
```

But probably we do not need to delete other config files - instead a kubelet should start after the file is created. https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#cni.

pkg/minikube/cni/custom.go
```diff
@@ -63,3 +69,12 @@ func (c Custom) Apply(r Runner) error {
 func (c Custom) CIDR() string {
 	return DefaultPodCIDR
 }
+
+func (c Custom) removeCNIConfig(r Runner) error {
+	klog.Infof("Removing existing cni configuration (/etc/cni/net.d/*)")
+	if _, err := r.RunCmd(exec.Command("sudo", "/bin/bash", "-c", "rm -rf /etc/cni/net.d/*")); err != nil {
```
afbjorklund, line 75: This removes the podman network, so that `podman run` doesn't work anymore.

We had similar issues with `kubeadm reset` before, that also destroyed podman.

afbjorklund (changes requested):
Why is it trying to use the wrong CNI, and starting with podman instead of the desired one ?

It would be better if it would just fail for a while, until the correct network is available.

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
afbjorklund: The podman and cri-o CNI config is installed with the package, so they are there at boot.

The podman configuration is also totally unrelated to Kubernetes, even though it uses CNI

michaelhenkel: I don't think that it is unrelated. When --cni is specified /etc/cni/net.d/1-k8s.conf is NOT created which only leaves 87-podman-bridge.conflist in /etc/cni/net.d. In absence of any other cni config, the kubelet uses the podman cni to assign IPs to Pods which come up BEFORE the config of the 3rd party is copied. 
Hence the coredns is served by podman cni. Manually deleting the coredns Pod solves the problem but that's not really ideal.
Removing/not creating  87-podman-bridge.conflist solves that problem: coredns Pod is only created when the kubelet finds the 3rd party cni config and everything is fine.
And yes, podman run now has a problem.

pkg/minikube/cni/custom.go
michaelhenkel: I changed the PR in a way that the podman CNI is not deleted. Instead a fake cni config is created IF the cni plugin is custom. This makes Pods using podNetwork waiting for the 3rd party CNI to come up.
Btw. Calico is suffering from the same problem. Following the operator manual (https://docs.projectcalico.org/getting-started/kubernetes/minikube) create coredns with a different IP than Calico should provide:
```
NAMESPACE         NAME                                       READY   STATUS    RESTARTS   AGE     IP              NODE       NOMINATED NODE   READINESS GATES
calico-system     calico-kube-controllers-6797d7cc89-pg2sg   1/1     Running   0          74s     10.244.0.3      minikube   <none>           <none>
calico-system     calico-node-q2qzd                          1/1     Running   0          74s     192.168.64.30   minikube   <none>           <none>
calico-system     calico-typha-54df7c5c64-qnqm9              1/1     Running   0          75s     192.168.64.30   minikube   <none>           <none>
kube-system       coredns-74ff55c5b-88zbx                    1/1     Running   0          2m47s   10.244.0.2      minikube   <none>           <none>
kube-system       etcd-minikube                              1/1     Running   0          2m57s   192.168.64.30   minikube   <none>           <none>
kube-system       kube-apiserver-minikube                    1/1     Running   0          2m57s   192.168.64.30   minikube   <none>           <none>
kube-system       kube-controller-manager-minikube           1/1     Running   0          2m57s   192.168.64.30   minikube   <none>           <none>
kube-system       kube-proxy-lppr4                           1/1     Running   0          2m48s   192.168.64.30   minikube   <none>           <none>
kube-system       kube-scheduler-minikube                    1/1     Running   0          2m57s   192.168.64.30   minikube   <none>           <none>
kube-system       storage-provisioner                        1/1     Running   0          3m2s    192.168.64.30   minikube   <none>           <none>
tigera-operator   tigera-operator-549bf46b5c-2jm8j           1/1     Running   0          114s    192.168.64.30   minikube   <none>           <none>
```

michaelhenkel: my PR actually fixes Calico as well:
```bash
minikube start  --cni=/tmp/tigera-operator.yaml --extra-config=kubeadm.pod-network-cidr=192.168.0.0/16 --driver hyperkit  --container-runtime cri-o
kubectl create -f https://docs.projectcalico.org/manifests/custom-resources.yaml
```

```bash
NAMESPACE         NAME                                       READY   STATUS    RESTARTS   AGE     IP               NODE       NOMINATED NODE   READINESS GATES
calico-system     calico-kube-controllers-6797d7cc89-fshvj   1/1     Running   0          70s     192.168.120.65   minikube   <none>           <none>
calico-system     calico-node-dpvcq                          1/1     Running   0          70s     192.168.64.31    minikube   <none>           <none>
calico-system     calico-typha-66fd8b686-l8c6f               1/1     Running   0          70s     192.168.64.31    minikube   <none>           <none>
kube-system       coredns-74ff55c5b-65mq8                    1/1     Running   0          2m19s   192.168.120.66   minikube   <none>           <none>
kube-system       etcd-minikube                              1/1     Running   0          2m29s   192.168.64.31    minikube   <none>           <none>
kube-system       kube-apiserver-minikube                    1/1     Running   0          2m29s   192.168.64.31    minikube   <none>           <none>
kube-system       kube-controller-manager-minikube           1/1     Running   0          2m28s   192.168.64.31    minikube   <none>           <none>
kube-system       kube-proxy-pshfd                           1/1     Running   0          2m19s   192.168.64.31    minikube   <none>           <none>
kube-system       kube-scheduler-minikube                    1/1     Running   0          2m28s   192.168.64.31    minikube   <none>           <none>
kube-system       storage-provisioner                        1/1     Running   0          2m31s   192.168.64.31    minikube   <none>           <none>
tigera-operator   tigera-operator-549bf46b5c-p7mjr           1/1     Running   0          2m19s   192.168.64.31    minikube   <none>           <none>
```

prezha:
as stated [here (step no. 8.)](https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/#init-workflow): 
> Please note that although the DNS server is deployed, it will not be scheduled until CNI is installed.

also, [here](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/#initializing-your-control-plane-node) are the steps used to initialise the control plane - maybe the issue could be in this part specifically:
```
2. Choose a Pod network add-on, and verify whether it requires any arguments to be passed to kubeadm init.
**Depending on which third-party provider you choose,
  you might need to set the --pod-network-cidr to a provider-specific value.**
See Installing a Pod network add-on.
```

michaelhenkel:
> as stated [here (step no. 8.)](https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/#init-workflow):
> 
> > Please note that although the DNS server is deployed, it will not be scheduled until CNI is installed.
> 
> also, [here](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/#initializing-your-control-plane-node) are the steps used to initialise the control plane - maybe the issue could be in this part specifically:
> 
> ```
> 2. Choose a Pod network add-on, and verify whether it requires any arguments to be passed to kubeadm init.
> **Depending on which third-party provider you choose,
>   you might need to set the --pod-network-cidr to a provider-specific value.**
> See Installing a Pod network add-on.
> ```

form Kubelet point of view the CNI is "installed" as soon as there is a cni config in /etc/cni/net.d. Now if you specify --cni parameter in minikube, initially the only cni config file in that directory is the podman cni config. The third party cni file will eventually land there but in between the coredns pod comes up and kubelet calls the podman CNI and the coredns pod gets an ip address from podman and not the third party cni.
As I showed this behavior applies to all CNIs which are installed via the --cni parameter and cannot get their cni config copied to /etc/cni/net.d BEFORE the coredns Pod is instantiated.

michaelhenkel:
any update?

michaelhenkel:
any update?

pkg/minikube/cni/custom.go
```diff
@@ -51,6 +53,10 @@ func NewCustom(cc config.ClusterConfig, manifest string) (Custom, error) {
 
 // Apply enables the CNI
 func (c Custom) Apply(r Runner) error {
+	//if err := c.removeCNIConfig(r); err != nil {
+	if err := c.createFakeCNIConfig(r); err != nil {
+		return errors.Wrap(err, "remove cni config")
```
tstromberg, line 58: change message to "fake cni"

pkg/minikube/cni/custom.go
```diff
@@ -63,3 +69,20 @@ func (c Custom) Apply(r Runner) error {
 func (c Custom) CIDR() string {
 	return DefaultPodCIDR
 }
+
+func (c Custom) createFakeCNIConfig(r Runner) error {
```
tstromberg, line 73: Can we rename this with a bit less stutter? This is in the `cni` package, after all. Perhaps:

`createFakeConfig` or `addFakeConfig`

Also, Since the idea of a fake cni config is a little weird, do you mind adding a single-line comment describing why we are doing this?

`// addFakeConfig adds a fake CNI configuration file to the node to (keep the kubelet from crashing?)`

pkg/minikube/cni/custom.go
```diff
@@ -63,3 +69,20 @@ func (c Custom) Apply(r Runner) error {
...
+  "type": "host-local"
+}`
+
+	b := []byte(fakeCNI)
+	fakeCNIAsset := assets.NewMemoryAssetTarget(b, "/etc/cni/net.d/86-fake.conf", "0644")
+	if err := r.Copy(fakeCNIAsset); err != nil {
+		return errors.Wrapf(err, "copy")
+	}
+
+	return sysinit.New(r).Restart(c.cc.KubernetesConfig.ContainerRuntime)
```
tstromberg, line 87: Is this necessary? If so, document why.

tstromberg (changes requested):
Thank you for doing this! The actual change to CNI looks good. I've added some minor quibbles, but think that overall that this can be merged quickly.

One weird thing has happened to this PR though: it seems to include all sorts of unrelated commits from master:

Screen Shot 2021-03-24 at 6 34 39 AM

Typically these show as a single merge-to-head commit, but in this PR, they are showing off as individual commits. I'm not sure if it's easier for you to remove the unrelated commits or start a new PR, but here is the workflow I use for merging against head after I've sent a PR in:

```
  git remote add upstream https://github.com/kubernetes/minikube
  git fetch upstream; \
    and git checkout master; \
    and git merge upstream/master; \
    and git checkout $branch; \
    and git merge master
```

I do this often enough that I've made it into a shell function. 

Thank you again for contributing this improvement!

pkg/minikube/cni/custom.go
michaelhenkel: done

pkg/minikube/cni/custom.go
michaelhenkel: done

michaelhenkel:
@tstromberg thanks for the thorough review! I think I addressed all comments. Can you please take a look?

medyagh:
this might not be needed anymore after @prezha PRs

PR #10388, opened 2021-02-06
Add condition to check --cpus count with available cpu count

### LINKED ISSUE #10386: Confusing error message when starting minikube with insufficient cpus
System:
```
$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="20.04.2 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.2 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

Minikube start:

```
$ minikube start --memory=6000 --cpus=6 --kubernetes-version=v1.14.1
* minikube v1.17.1 on Ubuntu 20.04 (amd64)
* Automatically selected the docker driver. Other choices: ssh, none
* Creating docker container (CPUs=6, Memory=6000MB) ...

X Exiting due to RSRC_INSUFFICIENT_CORES: Docker has less than 2 CPUs available, but Kubernetes requires at least 2 to be available
```

This is running on a VM with 4 vCPUs - clearly requesting 6 vCPUs was not sensible, but the error message could be made a little more abstract/generic as, in this case, Docker had more than 2 (v)CPUs available. Looks like a change to this line is required:

https://github.com/kubernetes/minikube/blob/ff051f9a3357947386f35b85e1a6c85d2620f517/pkg/minikube/node/advice.go#L53

(I originally posted [here](https://github.com/kubernetes/minikube/issues/9792#issuecomment-774445255) but was advised this is a separate issue).

BLasan:
fixes #10386 

Added a condition to check the argument value(--cpus) is less than the available cpus in the system

**Before PR**
When passing a value for --cpus argument, even if the value is larger than the available cpu count, it was going to print an error message which was not appropriate. see the logs mentioned in the issue #10386 

**After PR**
It'll print an appropriate error message according to the newly added condition. See changes in this PR
```
minikube start -p functional-20210207001733-2929731 --memory=4000 --apiserver-port=8441 --wait=true --cpus=10
😄  [functional-20210207001733-2929731] minikube v1.17.1 on Ubuntu 20.04

⛔  Exiting due to RSRC_INSUFFICIENT_CORES: Requested cpu count 10 is greater than the available cpus of 4
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1030,6 +1030,18 @@ func validateCPUCount(drvName string) {
 	}
 
 	if si.CPUs < cpuCount {
+
+		if driver.IsDockerDesktop(drvName) {
+			out.Step(style.Empty, `- Ensure your {{.driver_name}} daemon has access to enough CPU/memory resources.`, out.V{"driver_name": drvName})
+			if runtime.GOOS == "darwin" {
+				out.Step(style.Empty, `- Docs https://docs.docker.com/docker-for-mac/#resources`, out.V{"driver_name": drvName})
+			}
+			if runtime.GOOS == "windows" {
+				out.String("\n\t")
+				out.Step(style.Empty, `- Docs https://docs.docker.com/docker-for-windows/#resources`, out.V{"driver_name": drvName})
+			}
+		}
+
 		exitIfNotForced(reason.RsrcInsufficientCores, "Requested cpu count {{.requested_cpus}} is greater than the available cpus of {{.avail_cpus}}", out.V{"requested_cpus": cpuCount, "avail_cpus": si.CPUs})
 	}
 
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -995,7 +995,7 @@ func validateRequestedMemorySize(req int, drvName string) {
 	}
 }
 
-// validateCPUCount validates the cpu count matches the minimum recommended
+// validateCPUCount validates the cpu count matches the minimum recommended & not exceeding the available cpu count
 func validateCPUCount(drvName string) {
 	var cpuCount int
 	if driver.BareMetal(drvName) {
@@ -1027,6 +1027,10 @@ func validateCPUCount(drvName string) {
 			exit.Message(reason.Usage, "Ensure your {{.driver_name}} is running and is healthy.", out.V{"driver_name": driver.FullName(drvName)})
 		}
 
+	}
+
+	if si.CPUs < cpuCount {
+		exitIfNotForced(reason.RsrcInsufficientCores, "Requested cpu count {{.requested_cpus}} is greater than the available cpus of {{.avail_cpus}}", out.V{"requested_cpus": cpuCount, "avail_cpus": si.CPUs})
 	}
 
 	// looks good


BLasan:
@afbjorklund Please review

medyagh (commented):
@BLasan do u mind putting the out Before and After this PR in the PR Description ?

BLasan:
> @BLasan do u mind putting the out Before and After this PR in the PR Description ?

Added. Please take a look

cmd/minikube/cmd/start.go
```diff
@@ -1029,6 +1029,10 @@ func validateCPUCount(drvName string) {
 
 	}
 
+	if si.CPUs < cpuCount {
+		exitIfNotForced(reason.RsrcInsufficientCores, "Requested cpu count {{.requested_cpus}} is greater than the available cpus of {{.avail_cpus}}", out.V{"requested_cpus": cpuCount, "avail_cpus": si.CPUs})
```
medyagh, line 1033: before exit lets add an advice message to them if they are using Docker Desktop to increase their cpu in the docker desktop
```
if driver.IsDockerDesktop(drvName) {
			out.Step(style.Empty, `- Ensure your {{.driver_name}} daemon has access to enough CPU/memory resources.`, out.V{"driver_name": driver})
			if runtime.GOOS == "darwin" {
				out.Step(style.Empty, `- Docs https://docs.docker.com/docker-for-mac/#resources`, out.V{"driver_name": driver})
			}
			if runtime.GOOS == "windows" {
				out.String("\n\t")
				out.Step(style.Empty, `- Docs https://docs.docker.com/docker-for-windows/#resources`, out.V{"driver_name": driver})
			}
		}
}
```

also since this logic
and then do the exitIfNotFroced anyways

medyagh (changes requested):
@BLasan 
thank you for fixing this, lets improve the Exit message with advice for Docker Desktop on Mac and windows so they can allocate more CPUs

cmd/minikube/cmd/start.go
BLasan: I'll add the changes and re-push sir :)

afbjorklund: We already have that (Docker Desktop) warning in other places, but they will also have the same problem for this one.

It's not possible to allocate more CPU to the Docker VM, than what is available on the machine. So nothing to do here ?

afbjorklund:
> thank you for fixing this, lets improve the Exit message with advice for Docker Desktop on Mac and windows so they can allocate more CPUs

But only if there are more CPUs available on the host, than what is actually allocated to the Docker VM (reported by it)

Typical scenario would be that the host has 4 CPU (hyperthreads), but that the Docker VM only has 2 vCPU (by default).

So if user is trying to limit docker to 4 CPU, there would be this instruction. You can increase the VM, to use 4 vCPU.

But if they are trying to limit it to 8 CPU, the error from this change would apply... There is no more hardware available.

cmd/minikube/cmd/start.go
BLasan: > We already have that (Docker Desktop) warning in other places, but they will also have the same problem for this one.
> 
> It's not possible to allocate more CPU to the Docker VM, than what is available on the machine. So nothing to do here ?

Agree with your point. I just noticed that :)

BLasan: I think there's no issues with @medyagh 's suggestion. As `CachedDaemonInfo` function will return the docker/podman info. Not the host machine's details. It would be worth displaying an advice message like this before exiting. Please correct me if I'm wrong

cmd/minikube/cmd/start.go
```diff
@@ -1029,6 +1029,18 @@ func validateCPUCount(drvName string) {
 
 	}
 
+	if si.CPUs < cpuCount {
+		out.Step(style.Empty, `- Ensure your {{.driver_name}} daemon has access to enough CPU/memory resources.`, out.V{"driver_name": drvName})
```
medyagh, line 1033: we want to show this advice only for Docker Desktop Drivers (docker on mac and windows) and not other drivers (kvm or docker on linux...)

so lets add this check
```
if driver.IsDockerDesktop(drvName) {
...
```

medyagh (changes requested):

cmd/minikube/cmd/start.go
BLasan: Added. Please review sir

BLasan:
@medyagh please review

medyagh (approved):
thank you @BLasan

BLasan:
@medyagh Any remaining work to be done sir?

medyagh (commented):
thank you

PR #10393, opened 2021-02-07
Also prints out alternative drivers info when no driver is picked.

### LINKED ISSUE #10009: Rejected drivers not shown in priority order

lxzhang000:
fixes #10009


lxzhang000:
Need to verify I am using the correct email address, close this pull request and recreate another one later.

Issue #10237, opened 2021-01-23
Try to make the podman driver less "experimental", for podman version 3.0


afbjorklund:
There is a new version of podman being prepared: `v3.0.0`

* https://github.com/containers/podman/tree/v3.0

* https://github.com/containers/podman/releases/tag/v3.0.0-rc1

This is the release track that will go into next RHEL (previously based on first 1.6.4 and then 2.0.5)

* https://github.com/containers/podman/tree/v1.6.4-rhel

* https://github.com/containers/podman/tree/v2.0.5-rhel

Hopefully this means that the podman driver is _finally_ stable enough to be made non-experimental.

----

The main features are support for the docker socket and for docker compose, so nothing breaking ?

Support for podman has already been merged, and both KIC and ISO are currently running `v2.2.1`

From the ["Podman v3.0.0-rc1 is now available for testing"](https://lists.podman.io/archives/list/podman@lists.podman.io/thread/DDWBOGJDOV5QWDC4MSUJ7DUVMHU2H43H/) post:

> We are anticipating a final release in several weeks. We will publish
> a second release candidate next Monday, and if no major bugs are
> identified, the 3.0.0 final release will be the subsequent Monday,
> February 1.

afbjorklund:
Preferrably this should be tested on Fedora, and using the container runtime CRI-O.

afbjorklund:
Note: You _can_ also run Fedora with cgroups v1, ext4 and "moby-engine" (19.03).

_"But where is the fun in that"_

afbjorklund:
Minikube seemed to be working more or less OK with podman 3.0.0-RC2.

But needless to say there _were_ bugs, and we now have a RC3 instead...

https://github.com/containers/podman/releases/tag/v3.0.0-rc3

More testing with minikube is needed, especially on Fedora (or on RHEL)

afbjorklund:
Podman 3.0 has been released: https://github.com/containers/podman/releases/tag/v3.0.0

afbjorklund:
@medya said:
> how about we add Funcitonal Test on Github Action for Podman before we bump up the priority?

efussi:
FWIW, I upgraded to podman version 3.0 and tried to start minikube, but it failed:
```
$ minikube start --driver podman
😄  minikube v1.17.1 on Ubuntu 20.04 (vbox/amd64)
✨  Using the podman (experimental) driver based on user configuration
👍  Starting control plane node minikube in cluster minikube
[...]

PR #10394, opened 2021-02-07
Sorts the outputs and also prints out alternative driver info when no driver is picked.

### LINKED ISSUE #10009: Rejected drivers not shown in priority order

lxzhang000:
fixes #10009

Looks like "Unhealthy" prioriy level is now only used for driver selection, and I think it is not necessary since we can get the unhealthy info to filter the driver out from `State.Healthy`.

Another problem associated with this one is that the current alternatives is not printed out when failed.

Before:
1) All rejects are in "Unhealthy" priority therefore can't not be displayed by preference.
2) No info of alternative drivers that are not picked.

After:
1) All rejects keeps their orginal priority.
2) The alternative drivers lower than "Discouraged" priority will also be printed out (take "ssh" as an example).
```
* Unable to pick a default driver. Here is what was considered, in preference order:
  ...
  - ssh: Rejected due to low priority
  ...

```


medyagh:
@lxzhang000 please check unit test

run `make test` locally

medyagh (changes requested):
I can not verify if this PR is brekaing a logic that @tstromberg intended or not (classifiying an unhealthy driver as a priority of its own)

I refer judgment to @tstromberg

pkg/minikube/driver/driver.go
```diff
@@ -279,7 +279,11 @@ func Suggest(options []registry.DriverState) (registry.DriverState, []registry.D
 				continue
 			}
 
-			ds.Rejection = fmt.Sprintf("%s is preferred", pick.Name)
+			if pick.Name == "" {
+				ds.Rejection = "Rejected due to low priority"
```
tstromberg, line 283: Let's make this more specific and actionable. What do you think about this message:

`Not selected: "experimental" drivers must be specified using --driver`

tstromberg (changes requested):
Overall, things seem right to me. Tested using:

`touch /tmp/podman && chmod 755 /tmp/podman && env PATH=/tmp ./out/minikube start`

Output:

```
😄  minikube v1.17.1 on Debian rodete
    ▪ kvm2: Not installed: exec: "virsh": executable file not found in $PATH
    ▪ vmware: Not installed: exec: "docker-machine-driver-vmware": executable file not found in $PATH
    ▪ ssh: Rejected due to low priority
    ▪ none: Not installed: exec: "iptables": executable file not found in $PATH
    ▪ podman: Not healthy: exec: "sudo": executable file not found in $PATH

```

The only request is to improve the "Rejected due to low priority" text.

pkg/minikube/driver/driver.go
afbjorklund: Needs some other word than "experimental" though, the `none` and `ssh` are more like opt-in.

`Not selected: "%s" driver must be specified explicitly using --driver` or something to that effect ?

afbjorklund:
For podman it is true that it is still experimental, but hope to change that for next release. It should be on par with Docker.

See #10237

pkg/minikube/driver/driver.go
lxzhang000: Updated. Thanks for your advice.

Used "Not selected: %s must be specified explicitly using --driver" in accordance with the the previous string
"%s is preferred"

lxzhang000:
Hi, @medyagh I have updated the failed unit tests. FYI, I could pass the uni test locally even before the modification and the reason is that if two drivers have the same priority, their order is not determined.

I suppose the root cause is the traversal order of go map is not determinded:
https://github.com/kubernetes/minikube/blob/47da1ffe3f4b45a0ffdedea1bbea2912278fda26/pkg/minikube/registry/registry.go#L158

Looks like for display it doesn't care the order if drivers have the same priority, I use different priorites for unit test.

PR #10398, opened 2021-02-07
Allow dependabot to check go dependencies

jlosito:
This change allows dependabot to check any go dependency which this project uses on a weekly basis and submit pull requests with version bumps in order to keep packages up-to-date.

https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"


jlosito:
I signed it

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
```
afbjorklund, line 5: Might only want to check the regular directories: cmd pkg test

jlosito: I'm not seeing any `go.mod` files under the cmd directory.
```
% find cmd -name 'go.mod'
```

Are you saying that dependabot should only check the depedencies which the cmd path uses? I believe that would be achieved by using the allow block.

https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#allow

medyagh: @jlosito thats a good idea that @afbjorklund has, I agree lets limit to cmd pkg test

medyagh (changes requested):

.github/dependabot.yml
jlosito: @medyagh I don't think my original response was ever addressed. I don't believe the change being suggested is possible. Since there isn't any `go.mod` file under either of the cmd, pkg, or test paths, one would not be able to change the directory field to those paths according to the dependabot docuemtation. In other words, a `go.mod` has to exist underneath that path in order to utilize the directory field in the dependabot config. As stated in the documentation, "You must define the location of the package manifests for each package manager (for example, the package.json or Gemfile)."

https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#directory

I'm wondering if we're agreeing that all of the dependencies which are used underneath the cmd, pkg, and test directories should be collected and explicitly defined in an allow block. In other words, the change would look something like the following.
```
...
directory: "/"
allow:
  - dependency-name: "cmd-depenency-a"
  - dependency-name: "another-cmd-dependency"
  - dependency-name: "i_am_a_pkg_dependency"
```

The snippet above would limit dependabot to only change the 3 dependencies listed in the allow block.

afbjorklund: It also make sense to check all the dependencies in the go.mod/go.sum just because they are there...

PR #10412, opened 2021-02-09
remove crio restart hack

sharifelgamal:

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -252,17 +252,6 @@ func runStart(cmd *cobra.Command, args []string) {
 				})
 			}
 		}
-
-		if existing.KubernetesConfig.ContainerRuntime == "crio" {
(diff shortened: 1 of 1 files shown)


medyagh (changes requested):
I belive we no longer test crio at all 
we need to add Jenkins job for CRIO simmilar to containerd
as part of this PR I think we should create a job simmilar to https://a646c87040050000000000000000001.proxy.googleprod.com/job/Docker_Linux_containerd_integration/
 (at least for Docker Driver and one of the VM driver ones KVM)

afbjorklund:
Did something actually change upstream ?

sharifelgamal:
> Did something actually change upstream ?

It's been a few releases since we tried to remove this, so this is more of a test than anything.

medyagh (changes requested):
CRIO test is timing out (doesnt run all tests)

medyagh (approved):

PR #10423, opened 2021-02-09
Fix TestPreload on containerd/crio

priyawadhwa:
Currently this test immediately fails because we assume docker daemon is running/docker runtime.


medyagh (approved):
thank you for fixing this

PR #10426, opened 2021-02-09
kic: Make sure to add crun for running on cgroups v2

afbjorklund:
Podman will default to crun, for v2 (runc for v1)

So these dependencies are needed for podman.service

Note that runc is now provided by "containerd.io"

As seen in https://github.com/kubernetes/minikube/issues/10182#issuecomment-764552645


medyagh:
ok-to-build-image

PR #10428, opened 2021-02-09
Fix TestNetworkPlugins/group/auto/KubeletFlags on containerd/crio

priyawadhwa:
For docker runtime, if we don't pass in `--network-plugin` to `minikube start`, then it isn't passed in to the kubelet.

For containerd/crio, if we pass in nothing then `--network-plugin=cni` is passed in  to the kubelet. This modifies the test to take the runtime into account.

fixes https://github.com/kubernetes/minikube/issues/10429


medyagh (approved):
good catch ! @priyawadhwa

PR #10430, opened 2021-02-09
wip: Split up kic base image into 3 container-runtime based images

priyawadhwa:


afbjorklund:
Do you prefer using multiple base images, rather than letting the provisioner do the runtime installation ?

priyawadhwa:
Hey @afbjorklund that's a good question. My main concern with installation during runtime is that it would slow down `minikube start`.

afbjorklund:
> My main concern with installation during runtime is that it would slow down `minikube start`.

Yes, that has always been the question - if it should optimize on startup time or on size.

Both the three images and the preload makes it bigger, in return for some faster initial time.

But mostly I didn't want the hassle of maintaining three different images (+ the VM ISOs)

Then again one have to implement the provisioner instead, with the packaging and everything...

Issue #10379, opened 2021-02-05
Minikube start after previous delete fails with "network ... is already being used by a cni configuration"


adamish:
**Steps to reproduce the issue:** 

1. minikube start --driver=podman
2.  minikube delete
3. minikube  start --driver=podman

**Full output of `minikube start` command used**
```
😄  minikube v1.17.1 on Redhat 8.2
✨  Using the podman (experimental) driver based on user configuration
👍  Starting control plane node minikube in cluster minikube
🔥  Creating podman container (CPUs=2, Memory=3800MB) ...| E0205 22:30:38.549998  304006 network_create.go:85] error while trying to create network create network minikube 192.168.49.0/24: sudo -n podman network create --driver=bridge --subnet=192.168.49.0/24 --gateway=192.168.49.1 minikube: exit status 125
stdout:

stderr:
Error: network 192.168.49.0/24 is already being used by a cni configuration
```

The cause is that minikube delete doesn't clean up the network it created with podman, and the subsequent start command then sees the network and complains about it. 

This can be worked around with the following, however this isn't obvious
```
sudo podman network rm minikube
```

BLasan:
@adamish try running `minikube delete --all` instead of `minikube delete`

alekonko:
@BLasan i have same problem. Cluster seems to work correctly.


**Steps to reproduce the issue:** 

1. minikube delete --all
2. minikube config set driver podman
3. minikube start --driver=podman --container-runtime=cri-o -p minikube --cpus=4 --memory=12000MB


**Full output of failed command:** 



**Full output of `minikube start` command used, if not already included:**
😄  minikube v1.17.1 on Fedora 33
✨  Using the podman (experimental) driver based on user configuration
👍  Starting control plane node minikube in cluster minikube
🔥  Creating podman container (CPUs=4, Memory=12000MB) ...- E0207 12:41:09.265224  831751 network_create.go:85] error while trying to create network create network minikube 192.168.49.0/24: sudo -n podman network create --driver=bridge --subnet=192.168.49.0/24 --gateway=192.168.49.1 minikube: exit status 125
stdout:

stderr:
Error: network 192.168.49.0/24 is already being used by a cni configuration

❗  Unable to create dedicated network, this might result in cluster IP change after restart: un-retryable: create network minikube 192.168.49.0/24: sudo -n podman network create --driver=bridge --subnet=192.168.49.0/24 --gateway=192.168.49.1 minikube: exit status 125
stdout:

[...]

PR #10435, opened 2021-02-10
Explicitly remove podman volume and network

### LINKED ISSUE #9705: Podman driver doesn't remove volume or network on delete
This makes the _next_ create attempt fail:

```
🤦  StartHost failed, but will try again: creating host: create: creating: setting up container node: creating volume for minikube container: sudo -n podman volume create minikube --label name.minikube.sigs.k8s.io=minikube --label created_by.minikube.sigs.k8s.io=true: exit status 125
stdout:

stderr:
Error: volume with name minikube already exists: volume already exists

🔄  Restarting existing podman container for "minikube" ...
😿  Failed to start podman container. Running "minikube delete" may fix it: podman inspect ip minikube: sudo -n podman container inspect -f {{.NetworkSettings.IPAddress}} minikube: exit status 125
stdout:

stderr:
Error: error inspecting object: no such container minikube
```

afbjorklund:
The --filter and --label functionality was broken,
in earlier versions of podman (before version 3.0)

Closes #9705

### DIFF
--- a/pkg/drivers/kic/oci/volumes.go
+++ b/pkg/drivers/kic/oci/volumes.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"os/exec"
 	"runtime"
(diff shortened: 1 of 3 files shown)


pkg/drivers/kic/oci/volumes.go
```diff
@@ -29,6 +30,47 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// RemoveVolume removes a volume
+func RemoveVolume(ociBin string, name string) error {
+	if !volumeExists(ociBin, name) {
```
medyagh, line 35: nit: what is the performance implication of running this ? would it be better if we just try remove it regardless if it exists or not ? and if it errors out we could check if error is Volume doesn't exist?

medyagh (commented):
look good @afbjorklund , but needs rebase

pkg/drivers/kic/oci/volumes.go
afbjorklund: Most of the code is just copy/pasted from the network, there are no specific performance considerations. The volume is normally removed with the container (the `-v`) so it is just fallback really.

pkg/drivers/kic/oci/volumes.go
```diff
@@ -30,6 +31,47 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// RemoveVolume removes a volume
+func RemoveVolume(ociBin string, name string) error {
```
medyagh, line 35: this func is not used anywhere ?

medyagh (changes requested):

pkg/drivers/kic/oci/volumes.go
```diff
@@ -30,6 +31,47 @@ import (
...
+	if !volumeExists(ociBin, name) {
+		return nil
+	}
+	rr, err := runCmd(exec.Command(ociBin, "volume", "rm", name))
```
medyagh, line 39: should we use -f here (force) ?

pkg/drivers/kic/oci/volumes.go
afbjorklund: For some reason, the "delete" command (above) is talking directly to oci module. And the usual Delete() is never called ?

oci.RemoveVolume
oci.DeleteAllVolumesByLabel

oci.RemoveNetwork
oci.DeleteKICNetworks

afbjorklund: The naming of those functions is not great, and it seems like the network deletion could do with a specific label ?

pkg/drivers/kic/oci/volumes.go
afbjorklund: Hmm, "Remove a volume by force, even if it is being used by a container"

I don't think the volume should be in use, we just deleted the container ?

medyagh (approved):

BLasan:
@afbjorklund Are we able to close the issue #10379 from this commit as well?

PR #10454, opened 2021-02-12
Change prepareVolume func to add container name for preload side car

### LINKED ISSUE #10397: handle -killed- preload side-car holding on to the "minikube' volume  preventing volume from being deleted
first noticed when I apt-get thought "vim" was installed on a Fresh minikube !!! (however not it was installed in path)
 and the reason was it was using the same volume as previous deleted minikube

```
 $ minikube ssh
Last login: Sun Feb  7 21:58:01 2021 from 192.168.49.1
docker@minikube:~$ vim  
-bash: vim: command not found
docker@minikube:~$ sudo apt-get update && sudo apt-get install -y vim haproxy
...
...
...
vim is already the newest version (2:8.1.2269-1ubuntu5).
haproxy is already the newest version (2.0.13-2ubuntu0.1).
```

```
 $ minikube delete --all

 $ docker volume ls 
DRIVER    VOLUME NAME
local     minikube


$ docker volume rm --force minikube
Error response from daemon: remove minikube: volume is in use - [b4697cfa9aa1c955140909fb0a9cbd6b02291d308a462233fc2d2f9034e6f39b]
```

further debugging showed that I have a volume attached to a random name container (that seems to be the side-car for preload)

```
 $ docker ps -a

CONTAINER ID   IMAGE                                 COMMAND                  CREATED      STATUS    PORTS     NAMES
b4697cfa9aa1   gcr.io/k8s-minikube/kicbase:v0.0.17   "/usr/bin/test -d /v…"   2 days ago   Created             wonderful_galileo
```

I purpose we give a name to the side-car for example "profilename-preload-sidecar" so when we try to delete the cluster using minikube delete or minikube delete --all we can clean up possible left overs of side car.

CC: @priyawadhwa

govargo:
### What type of PR is this?
/kind bug
/co preload

This PR fixes the potential bug of the volume of preload-side-car container.
This PR fixes this.

### Which issue(s) this PR fixes:
Fix #10397

No. This PR fixes internal behaivour.

**Before this PR**

`func prepareVolume`

**After this PR**

`func prepareVolumeSideCar`

### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:
```
NONE
```

### DIFF
--- a/pkg/drivers/kic/oci/volumes.go
+++ b/pkg/drivers/kic/oci/volumes.go
@@ -151,9 +151,9 @@ func createVolume(ociBin string, profile string, nodeName string) error {
 	return nil
 }
 
-// prepareVolume will copy the initial content of the mount point by starting a container to check the expected content
-func prepareVolume(ociBin string, imageName string, nodeName string) error {
-	cmdArgs := []string{"run", "--rm", "--entrypoint", "/usr/bin/test"}
+// prepareVolumeSideCar will copy the initial content of the mount point by starting a temporary container to check the expected content
+func prepareVolumeSideCar(ociBin string, imageName string, nodeName string) error {
+	cmdArgs := []string{"run", "--rm", "--name", fmt.Sprintf("%s-preload-sidecar", nodeName), "--entrypoint", "/usr/bin/test"}
 	cmdArgs = append(cmdArgs, "-v", fmt.Sprintf("%s:/var", nodeName), imageName, "-d", "/var/lib")
 	cmd := exec.Command(ociBin, cmdArgs...)
 	if _, err := runCmd(cmd); err != nil {
--- a/pkg/drivers/kic/oci/volumes.go
+++ b/pkg/drivers/kic/oci/volumes.go
@@ -153,7 +153,7 @@ func createVolume(ociBin string, profile string, nodeName string) error {
 
 // prepareVolumeSideCar will copy the initial content of the mount point by starting a temporary container to check the expected content
 func prepareVolumeSideCar(ociBin string, imageName string, nodeName string) error {
-	cmdArgs := []string{"run", "--rm", "--name", fmt.Sprintf("%s-preload-sidecar", nodeName), "--entrypoint", "/usr/bin/test"}
+	cmdArgs := []string{"run", "--rm", "--name", fmt.Sprintf("%s-preload-sidecar", nodeName), "--entrypoint", "--label", fmt.Sprintf("%s=%s", CreatedByLabelKey, "true"), "/usr/bin/test"}
 	cmdArgs = append(cmdArgs, "-v", fmt.Sprintf("%s:/var", nodeName), imageName, "-d", "/var/lib")
 	cmd := exec.Command(ociBin, cmdArgs...)
 	if _, err := runCmd(cmd); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/drivers/kic/oci/volumes.go
```diff
@@ -151,9 +151,9 @@ func createVolume(ociBin string, profile string, nodeName string) error {
 	return nil
 }
 
-// prepareVolume will copy the initial content of the mount point by starting a container to check the expected content
-func prepareVolume(ociBin string, imageName string, nodeName string) error {
-	cmdArgs := []string{"run", "--rm", "--entrypoint", "/usr/bin/test"}
+// prepareVolumeSideCar will copy the initial content of the mount point by starting a temporary container to check the expected content
+func prepareVolumeSideCar(ociBin string, imageName string, nodeName string) error {
```
medyagh, line 155: as part of this we should ensure the side car gets deleted as part of the clean up process
minikube delete

medyagh: https://github.com/medyagh/minikube/blob/d1758a355786707670e17f274553ee619fbb0b31/cmd/minikube/cmd/delete.go#L99


https://github.com/medyagh/minikube/blob/d1758a355786707670e17f274553ee619fbb0b31/cmd/minikube/cmd/delete.go#L155

and also 	deletePossibleKicLeftOver
https://github.com/medyagh/minikube/blob/d1758a355786707670e17f274553ee619fbb0b31/cmd/minikube/cmd/delete.go#L189

pkg/drivers/kic/oci/volumes.go
```diff
@@ -151,9 +151,9 @@ func createVolume(ociBin string, profile string, nodeName string) error {
...
+	cmdArgs := []string{"run", "--rm", "--name", fmt.Sprintf("%s-preload-sidecar", nodeName), "--entrypoint", "/usr/bin/test"}
```
medyagh, line 156: consider adding the same created_by_minikube"Labels" to this container as well
add this label to the container
	CreatedByLabelKey = "created_by.minikube.sigs.k8s.io"

medyagh (changes requested):

pkg/drivers/kic/oci/volumes.go
govargo: Thank you for your point out!
This is important for the guarantee of deletion.

pkg/drivers/kic/oci/volumes.go
govargo: I added the CreatedByLabelKey for preload-side car.

pkg/drivers/kic/oci/volumes.go
govargo: I added the CreatedByLabelKey for preload-side car.

medyagh: @govargo can u verify if a side container is creatred and not deleted (remove --rm for testing)
the minikube delete would clean it up?

medyagh (changes requested):

pkg/drivers/kic/oci/volumes.go
govargo: I tested in my local with removing --rm. Then preload-side-car container kept living after `minikube delete`.

I investigated why the container kept living. And I noticed that `minikube delete` requieres ProfileLabelKey.
https://github.com/kubernetes/minikube/blob/40ac52ec4f35a3e8bc8017fb366c38899d78a974/cmd/minikube/cmd/delete.go#L249-L261

govargo: I updated to include `ProfileLabelKey`.
I tested `minikube delete` again with removing --rm. And I confirmed that `minikube delete` deleted all containers including preload-side-car container.

medyagh (approved):

Issue #8003, opened 2020-05-04
Podman Desktop driver on Mac or Win (and Linux)


afbjorklund:
We should test (and document) running the `podman` driver remotely, on non-Linux platforms.

Basically these platforms install `podman-remote` as `podman`, and run commands over varlink.

There is no documentation on how to set up the remote VM for Mac or Win, unfortunately...

The upstream install docs all assume that such a Linux host is already available over SSH:

https://podman.io/getting-started/installation

> **MacOS**
> Podman is a tool for running Linux containers. You can do this from a MacOS desktop as long as you have access to a linux box either running inside of a VM on the host, or available via the network. You need to install the remote client and then setup ssh connection information in the podman-remote.conf file.
> 
> **Windows**
> Podman is a tool for running Linux containers. You can do this from a Windows desktop as long as you have access to a linux box either running inside of a VM on the host, or available via the network. You need to install the remote client and then setup ssh connection information in the podman-remote.conf file.
> 

https://github.com/containers/libpod/blob/master/docs/source/markdown/podman-remote.conf.5.md

----

Theoretically using `podman-machine` should work, as long as you use the systemd one:

https://boot2podman.github.io/2018/12/07/fedora-iso-variant.html

It is also possible to do it yourself, using the hypervisor of your choice... (and a Fedora ISO)

There is a walkthrough here: https://itnext.io/podman-and-skopeo-on-macos-1b3b9cf21e60

We need to document some minikube-specific configuration, like switching over to `cgroupfs`

afbjorklund:
There's some minor differences, that we need to handle in code

`Error: unrecognized command `podman cp``

afbjorklund:
There is also some other limitations, that we need to address:

```go
        // DefaultBindIPV4 is The default IP the container will listen on.
        DefaultBindIPV4 = "127.0.0.1"
```

```go
// GetExternalIP returns an IP which is accissble from outside
func (d *Driver) GetExternalIP() (string, error) {
        return oci.DefaultBindIPV4, nil
}

// GetSSHHostname returns hostname for use with ssh
func (d *Driver) GetSSHHostname() (string, error) {
        return oci.DefaultBindIPV4, nil
}
```

This assumes that we always set up tunnels for all ports on the VM.
Otherwise there is nothing transporting from the the host to the VM.

Like in: https://docs.docker.com/docker-for-mac/networking/#use-cases-and-workarounds

afbjorklund:
[...]

PR #10458, opened 2021-02-12
Change podman priority to default on Linux

### LINKED ISSUE #10237: Try to make the podman driver less "experimental", for podman version 3.0

afbjorklund:
Requires Podman 2.1 or greater (such as 3.0)

Podman is *not* yet supported for Mac or Win.
See #8003 regarding running with podman-remote

Closes #10237

### DIFF
--- a/pkg/minikube/registry/drvs/podman/podman.go
+++ b/pkg/minikube/registry/drvs/podman/podman.go
@@ -38,13 +38,19 @@ import (
 	"k8s.io/minikube/pkg/minikube/registry"
 )
 
+var docURL = "https://minikube.sigs.k8s.io/docs/drivers/podman/"
+
 // minReqPodmanVer is required the minimum version of podman to be installed for podman driver.
-var minReqPodmanVer = semver.Version{Major: 1, Minor: 7, Patch: 0}
+var minReqPodmanVer = semver.Version{Major: 2, Minor: 1, Patch: 0}
 
 func init() {
-	priority := registry.Experimental
+	priority := registry.Default
+	if runtime.GOOS != "linux" {
+		// requires external VM set up
+		priority = registry.Experimental
+	}
 	// Staged rollout for default:
-	// - Linux
+	// - Linux (sudo podman)
 	// - macOS (podman-remote)
 	// - Windows (podman-remote)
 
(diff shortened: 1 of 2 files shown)


afbjorklund:
/hold

Seems like our old friend the `/var` race is back again:

`update-alternatives: error: no alternatives for iptables`

pkg/minikube/registry/drvs/podman/podman.go
```diff
@@ -38,13 +38,19 @@ import (
...
 
+var docURL = "https://minikube.sigs.k8s.io/docs/drivers/podman/"
+
 // minReqPodmanVer is required the minimum version of podman to be installed for podman driver.
-var minReqPodmanVer = semver.Version{Major: 1, Minor: 7, Patch: 0}
+var minReqPodmanVer = semver.Version{Major: 2, Minor: 1, Patch: 0}
 
 func init() {
-	priority := registry.Experimental
+	priority := registry.Default
```
medyagh, line 47: how about we add Funcitonal Test on Github Action for Podman before we bump up the priority?

medyagh (changes requested):

pkg/minikube/registry/drvs/podman/podman.go
afbjorklund: Sounds like a good suggestion, please add it to the issue

afbjorklund: Still works fine with Podman 2.2.1, though. So nothing to change in this PR, just because 3.0.0 was buggy

site/content/en/docs/drivers/podman.md
```diff
@@ -7,23 +7,17 @@ aliases:
...
 
-{{% pageinfo %}}
-This driver is experimental and in active development. Help wanted!
-{{% /pageinfo %}}
-
 The podman driver is an alternative container runtime to the [Docker]({{< ref "/docs/drivers/docker.md" >}}) driver.
 
 ## Requirements
 
-- Linux or macOS operating systems on amd64 architecture
```
afbjorklund, line 18: This "or macOS" was left since before, but the (wrong) code was removed in d0260cca998d59b293a287027230f845c5957a9c

afbjorklund:
Works OK with podman 3.0.1

/hold cancel

pkg/minikube/registry/drvs/podman/podman.go
afbjorklund: Now 3.0.1 is out as well, help would be needed for adding a functional test

priyawadhwa (approved):

PR #10473, opened 2021-02-13
Improve UI advice in cluster deletion

### LINKED ISSUE #10460: improve UI advice when user needs to delete the cluster
here we confuse the user with two things

❗  You cannot change the memory size for an exiting minikube cluster. Please first delete the cluster.
and 

😿  Failed to start ssh bare metal machine. Running "minikube delete -p foo" may fix it: config: please provide an IP address

we should had sticked with first advice

```
drewpca(pts/0):~% minikube start -p foo --memory 6GB
😄  [foo] minikube v1.17.1 on Debian rodete
✨  Using the ssh driver based on existing profile
❗  You cannot change the memory size for an exiting minikube cluster. Please first delete the cluster.
👍  Starting control plane node foo in cluster foo
🤦  StartHost failed, but will try again: config: please provide an IP address
😿  Failed to start ssh bare metal machine. Running "minikube delete -p foo" may fix it: config: please provide an IP address
 
❌  Exiting due to GUEST_PROVISION: Failed to start host: config: please provide an IP address
 
```

BLasan:
fixes #10460 

Before PR
See #10460 

After PR
Cluster Deletion error message will be prompted under three conditions. When `memInMB != cc.Memory` or `viper.GetInt(cpus) != cc.CPUs`or `memInMB != existing.DiskSize` . According to the newly added condition, the confusing error message will not be printed if one of the conditions get satisfied.

### DIFF
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -456,7 +456,14 @@ func startHost(api libmachine.API, cc *config.ClusterConfig, n *config.Node, del
 
 	// Don't use host.Driver to avoid nil pointer deref
 	drv := cc.Driver
-	out.ErrT(style.Sad, `Failed to start {{.driver}} {{.driver_type}}. Running "{{.cmd}}" may fix it: {{.error}}`, out.V{"driver": drv, "driver_type": driver.MachineType(drv), "cmd": mustload.ExampleCmd(cc.Name, "delete"), "error": err})
+	memInMB, _ := util.CalculateSizeInMB(viper.GetString("memory"))
+	diskMemInMB, _ := util.CalculateSizeInMB(viper.GetString("disk-size"))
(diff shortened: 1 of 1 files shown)


BLasan:
@priyawadhwa @medyagh Please review

BLasan:
@medyagh @priyawadhwa Please review

BLasan:
@priyawadhwa @medyagh Please review

medyagh (changes requested):
I am not sure if this PR is fixing the issue, and also it is ignoring the errors _ which should not be done.

BLasan:
> I am not sure if this PR is fixing the issue, and also it is ignoring the errors _ which should not be done.

We need to avoid printing that particular error log if it satisfies one of the above conditions right? There's a mistake in logic I guess. If you have an idea I would like to hear from you sir

PR #10480, opened 2021-02-14
Make sure to use CRI for docker if configured

### LINKED ISSUE #9868: Consider using CRI for docker runtime
Just like the "dockershim" in Kubernetes, the cruntime uses direct commands for Docker.

https://github.com/kubernetes/minikube/blob/master/pkg/minikube/cruntime/docker.go

It would be nice if it would use the CRI (where possible), like the _other_ runtimes are doing.

https://github.com/kubernetes/minikube/blob/master/pkg/minikube/cruntime/cri.go

We do have the `crictl` all configured, so just a matter of "flipping a switch" basically.

The initial approach might be to introduce a feature toggle, to make it opt-in to start with...

afbjorklund:
The default is still the built-in dockershim,
but make sure to honor any external CRI socket.

Also forward the other configuration such as
the Kubernetes version and Image repository.

Closes #9868

----

This can be tested with https://github.com/dims/cri-dockerd

Installed on the node, and started with something like:

`sudo cri-dockerd --container-runtime-endpoint unix:///var/run/dockershim2.sock`

minikube start  start --container-runtime=docker --cri-socket=/var/run/dockershim2.sock

### DIFF
--- a/pkg/minikube/cruntime/cruntime.go
+++ b/pkg/minikube/cruntime/cruntime.go
@@ -151,9 +151,12 @@ func New(c Config) (Manager, error) {
 	switch c.Type {
 	case "", "docker":
 		return &Docker{
-			Socket: c.Socket,
-			Runner: c.Runner,
-			Init:   sm,
+			Socket:            c.Socket,
+			Runner:            c.Runner,
+			ImageRepository:   c.ImageRepository,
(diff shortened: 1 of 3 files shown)


medyagh (approved):

medyagh (approved):
nice work, and thanks to @dims

PR #10506, opened 2021-02-18
Verify that the build directory for buildroot is case sensitive

### LINKED ISSUE #3752: documentation: use a case-sensitive disk image when building the ISO on mac
**Environment**
macOS

When running `make out/minikube.iso` the build fails I'm having an issue with buildroot. The failure seems to be caused by `/bin/sh: 1: cannot create /mnt/out/buildroot/output/build/glibc-glibc-2.27-57-g6c99e37f6fb640a50a3113b2dbee5d5389843c1e/build/gnulib/stamp.os: File exists`.

```
rm -f /mnt/out/buildroot/output/build/glibc-glibc-2.27-57-g6c99e37f6fb640a50a3113b2dbee5d5389843c1e/build/gnulib/stamp.os; > /mnt/out/buildroot/output/build/glibc-glibc-2.27-57-g6c99e37f6fb640a50a3113b2dbee5d5389843c1e/build/gnulib/stamp.os
rm -f /mnt/out/buildroot/output/build/glibc-glibc-2.27-57-g6c99e37f6fb640a50a3113b2dbee5d5389843c1e/build/gnulib/stamp.o; > /mnt/out/buildroot/output/build/glibc-glibc-2.27-57-g6c99e37f6fb640a50a3113b2dbee5d5389843c1e/build/gnulib/stamp.o
/bin/sh: 1: cannot create /mnt/out/buildroot/output/build/glibc-glibc-2.27-57-g6c99e37f6fb640a50a3113b2dbee5d5389843c1e/build/gnulib/stamp.os: File exists
make[5]: *** [/mnt/out/buildroot/output/build/glibc-glibc-2.27-57-g6c99e37f6fb640a50a3113b2dbee5d5389843c1e/build/gnulib/stamp.os] Error 2
make[5]: *** Waiting for unfinished jobs....
../Makerules:920: recipe for target '/mnt/out/buildroot/output/build/glibc-glibc-2.27-57-g6c99e37f6fb640a50a3113b2dbee5d5389843c1e/build/gnulib/stamp.os' failed
make[5]: Leaving directory '/mnt/out/buildroot/output/build/glibc-glibc-2.27-57-g6c99e37f6fb640a50a3113b2dbee5d5389843c1e/gnulib'
Makefile:215: recipe for target 'gnulib/subdir_lib' failed
make[4]: *** [gnulib/subdir_lib] Error 2
make[4]: Leaving directory '/mnt/out/buildroot/output/build/glibc-glibc-2.27-57-g6c99e37f6fb640a50a3113b2dbee5d5389843c1e'
Makefile:9: recipe for target 'all' failed
make[3]: Leaving directory '/mnt/out/buildroot/output/build/glibc-glibc-2.27-57-g6c99e37f6fb640a50a3113b2dbee5d5389843c1e/build'
make[3]: *** [all] Error 2
package/pkg-generic.mk:223: recipe for target '/mnt/out/buildroot/output/build/glibc-glibc-2.27-57-g6c99e37f6fb640a50a3113b2dbee5d5389843c1e/.stamp_built' failed
make[2]: *** [/mnt/out/buildroot/output/build/glibc-glibc-2.27-57-g6c99e37f6fb640a50a3113b2dbee5d5389843c1e/.stamp_built] Error 2
make[2]: Leaving directory '/mnt/out/buildroot'
make[1]: *** [minikube_iso] Error 2
Makefile:127: recipe for target 'minikube_iso' failed
make[1]: Leaving directory '/mnt'
make: *** [out/minikube.iso] Error 2
Makefile:151: recipe for target 'out/minikube.iso' failed
make: *** [out/minikube.iso] Error 2
```

jandubois:
buildroot fails when the filesystem is just case-preserving, which is the default for macOS filesystems. So it is better to fail
immediately than after spending 90min building things.

This commit also includes a `make case-sensitive-build-dir` target (implemented on macOS only) that creates a case-sensitive spareimage and mounts it at the build dir location. Any files from the previous build dir are moved to this new sparseimage.

I do not understand why the `hdiutil detach` in the `clean` target needs the -force option when running via `make`. Otherwise it will claim "resource busy". This does not happen when running directly
from the shell.

Fixes #3752 (it has been closed because building the ISO now happens in a docker container, but that doesn't actually address the issue that the build directory is not case-sensitive; it is mounted as-is into the container).

### Demo

```
$ make minikube_iso
echo yes > ./out/.case-sensitive
echo no > ./out/.CASE-sensitive
if [ "$(cat ./out/.case-sensitive)" != "yes" ]; then \
		echo "Build directory ./out is not a case-sensitive filesystem."; \
		echo "On darwin 'make case-sensitive-build-dir' will move the build dir to a case-sensitive sparseimage."; \
		exit 1; \
	fi;
Build directory ./out is not a case-sensitive filesystem.
On darwin 'make case-sensitive-build-dir' will move the build dir to a case-sensitive sparseimage.
make: *** [minikube_iso] Error 1

$ make case-sensitive-build-dir
test ! -f ./out.sparseimage
mv ./out ./out.orig
hdiutil create -type SPARSE -fs 'Case-sensitive Journaled HFS+' -size 100G -volname out ./out.sparseimage
created: /Users/jan/tmp/minikube/out.sparseimage
hdiutil attach -mountpoint ./out ./out.sparseimage
/dev/disk18         	GUID_partition_scheme
/dev/disk18s1       	EFI
/dev/disk18s2       	Apple_HFS                      	/Users/jan/tmp/minikube/out
touch ./out.orig/ignore-me
mv ./out.orig/* ./out
rm -rf ./out.orig

$ make minikube_iso
echo yes > ./out/.case-sensitive
echo no > ./out/.CASE-sensitive
if [ "$(cat ./out/.case-sensitive)" != "yes" ]; then \
		echo "Build directory ./out is not a case-sensitive filesystem."; \
		echo "On darwin 'make case-sensitive-build-dir' will move the build dir to a case-sensitive sparseimage."; \
		exit 1; \
	fi;
echo v1.17.0 > deploy/iso/minikube-iso/board/coreos/minikube/rootfs-overlay/etc/VERSION
if [ ! -d ./out/buildroot ]; then \
		git clone --depth=1 --branch=2020.02.8 https://github.com/buildroot/buildroot ./out/buildroot; \
	fi
Cloning into './out/buildroot'...
remote: Enumerating objects: 14703, done.
remote: Counting objects: 100% (14703/14703), done.
^Cfetch-pack: unexpected disconnect while reading sideband packet
make: *** [minikube_iso] Interrupt: 2

$ make clean
if [ -d ./out -a -f ./out.sparseimage ]; then \
		hdiutil detach -force ./out; \
		rm -f ./out.sparseimage; \
	fi
"disk18" ejected.
rm -rf ./out
rm -f pkg/minikube/assets/assets.go
rm -f pkg/minikube/translate/translations.go
rm -rf ./vendor
rm -rf /tmp/tmp.*.minikube_*
```


Makefile
```diff
@@ -88,7 +88,11 @@ GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
 GOPATH ?= $(shell go env GOPATH)
 BUILD_DIR ?= ./out
-$(shell mkdir -p $(BUILD_DIR))
+# Re-attach case-sensitive sparseimage in case it got detached by e.g. a reboot
+ifeq ($(GOOS),darwin)
+	IGNORED := $(shell test ! -e $(BUILD_DIR) -a -f $(BUILD_DIR).sparseimage && hdiutil attach -mountpoint $(BUILD_DIR) $(BUILD_DIR).sparseimage)
```
afbjorklund, line 93: You probably want to leave "out" as-is, and only do "out/buildroot" as a sparse image ?

afbjorklund (changes requested):
I think the other targets are fine (with out), so this should be more buildroot-specific ?

Makefile
jandubois: > I think the other targets are fine (with out), so this should be more buildroot-specific ?

Yes, they seem to be, which is why the check for the case-sensitive filesystem is only in the `minikube_iso` target. So you don't need to run `make case-sensitive-build-dir` unless you want to build that target.

But when you switch to a sparseimage, I don't know why it should be a problem switching all of `./out`. It should be completely transparent. Doing it just for the buildroot dir makes the patch a bit more complex for no benefit afaict. Let me know if I'm missing something, and please confirm that this change really is required!

BTW, I noticed that support for changing `$(BUILD_DIR)` is pretty incomplete: there are plenty of verbatim `out/` strings in targets, dependencies and commands, so changing it from the default value doesn't seem to work in general. Is this something you want to see changed?

jandubois: > Doing it just for the buildroot dir makes the patch a bit more complex for no benefit afaict.

I realized that it doesn't actually make it significantly more complex, so I have changed the PR accordingly.

Please let me know if you want me to squash the commits in the PR? I've already rebased on the latest origin/master to hopefully get rid of the functional test failures.

priyawadhwa (commented):
This seems reasonable to me, but I think @afbjorklund should have final say as our buildroot expert :)

medyagh:
I recommend testing this PR with @sharifelgamal's iso automation

medyagh:
@jandubois sorry for the delay in the review, do u mind pulling master and redoing this PR ? We re-designed the ISO in the makefile, once pull upstream I will build an image using this PR to test it

jandubois:
I've rebased the PR on latest `master` (there were no conflicts) and also squashed the commits together, as the rebase would out-date previous review comments anyways.

jandubois:
Looking at the PR again I noticed that the Makefile just clones the buildroot repo and uses whatever the latest commit is on the default branch.

I wonder if it shouldn't select a specific tag to have better reproducible results. Also I believe re-using build artifacts from previous builds after updating buildroot to a later version is not guaranteed to rebuild all updated dependencies correctly.

But I guess as long as published ISOs are always build from a clean checkout (and maybe use ccache instead of re-using the build dir), then this may not matter that much.

medyagh:
ok-to-build-iso

jandubois:
I've rebased the PR one more time, but it is kind of ridiculous that the `minikube-bot` is adding  commits to the PR which then introduce merge conflicts. I've just removed the bot commit again: https://github.com/kubernetes/minikube/commit/46ccbdc

sharifelgamal:
The intent of the bot is not to introduce merge conflicts, we're building the iso based on the code changes in this PR then point the latest minikube code to that ISO. That way everyone gets that same ISO. The conflicts come from when we do this to several PRs at once, which is a failing on our part, sorry about that.

I'm going to rebuild that ISO, which will cause minikube-bot to commit to this PR again. There shouldn't be any merge conflicts this time. Once the tests pass with the newly built ISO, we should be able to merge this PR as is, with no more changes.

sharifelgamal:
ok-to-build-iso

jandubois:
> @jandubois: PR needs rebase.

I can continue to rebase, but I guess there is not much point in doing so until @afbjorklund has confirmed that the amendments I made on Feb 18 satisfy his code review request for changes.

Or is there anything else I should be doing to unblock this PR? Given the nature of this change I'm quite certain that any of the failed checks are not actually due to this PR itself.

medyagh:
@afbjorklund PTAL

@jandubois thank you for your patience

sharifelgamal:
ok-to-build-iso

jandubois:
I have no idea why this was never merged; I guess there is no real interest in being able to do the buildroot build on macOS.

I don't really care anymore either; but please let me know if I should delete the PR branch, or if I should leave it lingering around, in case somebody else runs into the same issue!

afbjorklund:
Totally forgot about it and don't use a Mac anymore so couldn't verify either

PR #10513, opened 2021-02-19
Enable handing over the kubeadm --v flag

sbueringer:
Hi,

I was debugging minikube today and my problem was that it was stuck somewhere in kubeadm, waiting for the control plane to come up. Maybe I missed some flags but it was very hard to debug until I build a custom binary which allows specifying `--v`
for the `kubeadm init` call.

It would be very nice, to have this enabled per default.

### DIFF
--- a/pkg/minikube/bootstrapper/bsutil/extraconfig.go
+++ b/pkg/minikube/bootstrapper/bsutil/extraconfig.go
@@ -74,6 +74,7 @@ var KubeadmExtraArgsAllowed = map[int][]string{
 		"certificate-key",
 		"rootfs",
 		"skip-phases",
+		"v",
 	},
 	KubeadmConfigParam: {
 		"pod-network-cidr",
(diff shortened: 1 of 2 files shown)


sbueringer:
Not sure about the ` addons_certs_docker_ubuntu` test

medyagh (approved):
@sbangari thank you for this PR, it seems good to me to add verbosity,
do u mind sharing before and after this PR (the logs) for reference ?

sbueringer:
> @sbangari thank you for this PR, it seems good to me to add verbosity,
> do u mind sharing before and after this PR (the logs) for reference ?

So per default the verbosity does not change. It's just when you set the verbosity you a specific level you can see the corresponding logs of this level of kubeadm. So the verbosity is basically like the one from kubeadm. I can share the logs on Monday though.

medyagh:
@sbueringer is this for a binary that only works for u or any other user can use this ?

sbueringer:
@medyagh  looks like the loglevel is already propagated somehow to kubeadm. Looks like we missed to `--logtostderr` flag at some point during the debugging :/

/close

PR #10540, opened 2021-02-20
Introduce alias native for none driver

sadlil:
First step to rename none to native, introducing the alias `native` for none driver.

Running the following will be equivalent
```
$ minikube start --driver=none
$ minikube start --driver=native
```

Ref: https://github.com/kubernetes/minikube/issues/7772

### DIFF
--- a/pkg/minikube/driver/driver.go
+++ b/pkg/minikube/driver/driver.go
@@ -59,6 +59,8 @@ const (
 	AliasKVM = "kvm"
 	// AliasSSH is driver name alias for ssh
 	AliasSSH = "generic"
+	// AliasNative is driver name alias for None driver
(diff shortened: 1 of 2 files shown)


medyagh (approved):

afbjorklund (approved):
Works for me, I suppose the next step is to rename everything and flip the Name and Alias ?

Name: driver.Native,
Alias:    []string{driver.AliasNone},

medyagh (approved):

medyagh:
@sadlil I agree with @afbjorklund  I think in next verison we should start flipping them, and making the native the default name.

sadlil:
Yes. I am going to do that. Thanks.

Issue #10541, opened 2021-02-20
crio: ctr does NOT show loaded image it but podman images shows it ...


medyagh:
looking at failure in this test that upgrades crio version https://storage.googleapis.com/minikube-builds/logs/10477/d120656/KVM_Linux_crio.html#fail_TestFunctional%2fparallel%2fLoadImage

when we load an image to crio run time, 
we are checking if ctr has it
and it doesnt
```
docker@minikube:~$ sudo ctr -n=k8s.io image ls
REF TYPE DIGEST SIZE PLATFORMS LABELS 

```

but when I check with podman I see the image 
```
localhost/busybox                        test-example
```

```


docker@minikube:~$ sudo podman image inspect --format {{.Id}} busybox:test-example^C
docker@minikube:~$ sudo podman image list
REPOSITORY                               TAG           IMAGE ID      CREATED        SIZE
localhost/busybox                        test-example  491198851f0c  46 hours ago   1.45 MB
k8s.gcr.io/kube-proxy                    v1.20.4       c29e6c583067  2 days ago     120 MB
k8s.gcr.io/kube-apiserver                v1.20.4       ae5eb22e4a9d  2 days ago     123 MB
k8s.gcr.io/kube-controller-manager       v1.20.4       0a41a1414c53  2 days ago     118 MB
k8s.gcr.io/kube-scheduler                v1.20.4       5f8cb769bd73  2 days ago     48.5 MB
docker.io/kubernetesui/dashboard         v2.1.0        9a07b5b4bfac  2 months ago   229 MB
gcr.io/k8s-minikube/storage-provisioner  v4            85069258b98a  2 months ago   29.7 MB
k8s.gcr.io/etcd                          3.4.13-0      0369cf4303ff  5 months ago   255 MB
k8s.gcr.io/coredns                       1.7.0         bfe3a36ebd25  8 months ago   45.4 MB
docker.io/kubernetesui/metrics-scraper   v1.0.4        86262685d9ab  11 months ago  36.9 MB
k8s.gcr.io/pause                         3.2           80d28bedfe5d  12 months ago  686 kB
```

https://github.com/medyagh/minikube/blob/d120656161437e5f3f36abe2ceb2e9ff61b6d9fc/test/integration/functional_test.go#L194

medyagh:
I think we should have used critctl instead of ctr simmilar to our cache add tests
```
docker@minikube:~$ sudo crictl images
[...]

PR #10542, opened 2021-02-20
test: use crictl inspecti for verify images in test

medyagh:
closes https://github.com/kubernetes/minikube/issues/10541

### DIFF
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -194,7 +194,11 @@ func validateLoadImage(ctx context.Context, t *testing.T, profile string) {
 	var cmd *exec.Cmd
 	if ContainerRuntime() == "docker" {
 		cmd = exec.CommandContext(ctx, Target(), "ssh", "-p", profile, "--", "docker", "image", "inspect", newImage)
+	} else if ContainerRuntime() == "containerd" {
+		// crictl inspecti busybox:test-example
+		cmd = exec.CommandContext(ctx, Target(), "ssh", "-p", profile, "--", "sudo", "crictl", "inspecti", newImage)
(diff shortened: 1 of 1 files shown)


test/integration/functional_test.go
```diff
@@ -193,9 +193,10 @@ func validateLoadImage(ctx context.Context, t *testing.T, profile string) {
 	// make sure the image was correctly loaded
 	var cmd *exec.Cmd
 	if ContainerRuntime() == "docker" {
-		cmd = exec.CommandContext(ctx, Target(), "ssh", "-p", profile, "--", "docker", "images", "--format", "{{.Repository}}:{{.Tag}}")
+		cmd = exec.CommandContext(ctx, Target(), "ssh", "-p", profile, "--", "docker", "image", "inspect", newImage)
 	} else {
-		cmd = exec.CommandContext(ctx, Target(), "ssh", "-p", profile, "--", "sudo", "ctr", "-n=k8s.io", "image", "ls")
+		// crictl inspecti localhost/busybox:test-example
+		cmd = exec.CommandContext(ctx, Target(), "ssh", "-p", profile, "--", "sudo", "crictl", "inspecti", "localhost/"+newImage)
```
afbjorklund, line 199: Not sure if containerd adds the localhost/ prefix, I think that is a podman thing. Might have to strip it out ?

Alternatively make sure to use qualified images everywhere, such as "docker.io/busybox". And re-add it later.

afbjorklund (approved):

test/integration/functional_test.go
medyagh: good catch

PR #10554, opened 2021-02-21
Don't select Discouraged or Obsolete by default

### LINKED ISSUE #10516: When docker driver fails, we try to use an (unspecified) external VM instead
As seen in https://github.com/kubernetes/minikube/issues/10459#issue-807564720

```
❗ Startup with docker driver failed, trying with alternate driver ssh
🤦 StartHost failed, but will try again: config: please provide an IP address
```

It's **not** possible to use a remote VM ("ssh" driver), without providing the address to it...

We should also not silently fallback to running locally ("none" driver), because of no isolation.

`✨  Automatically selected the docker driver. Other choices: kvm2, virtualbox, none, ssh, podman (experimental)`

Currently the priority is "Discouraged", which doesn't really the "Unconfigured" state properly.

The goal should be to make them both opt-in, and not have them on the regular Alternatives list.

```
pkg/minikube/registry/drvs/none/none.go:                Priority: registry.Discouraged, // requires root
pkg/minikube/registry/drvs/ssh/ssh.go:          Priority: registry.Discouraged, // requires external VM
```

afbjorklund:
These drivers require either an active choice (opt-in),
or a previous configuratio (for the warning message).

Closes #10516

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -213,6 +213,10 @@ func runStart(cmd *cobra.Command, args []string) {
 			success := false
 			// Walk down the rest of the options
 			for _, alt := range alts {
+				// Skip non-default drivers
+				if !ds.Default {
+					continue
+				}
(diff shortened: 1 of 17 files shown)


priyawadhwa (approved):
lgtm

PR #10557, opened 2021-02-21
chore(deps): bump go to v1.16

chenrui333:
go1.16 is a big release for go modules build, update the repo to use the latest and greatest golang. :)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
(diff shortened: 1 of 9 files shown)


afbjorklund:
> go1.16 is a big release for go modules build, update the repo to use the latest and greatest golang. :)

I would rather wait until Kubernetes says it is OK, so we get all the cross build images and everything

medyagh (commented):
I think it is better for us to match the kubernetes go version
https://github.com/kubernetes/kubernetes/blob/master/build/build-image/cross/VERSION

we will wait till kubernetes upgrade the go version to 16

PR #10558, opened 2021-02-21
chore(deps): update dependabot for github-action and docker

chenrui333:
update dependabot for `github-action` and `docker`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,15 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
+  # Maintain dependencies for GitHub Actions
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,3 +5,15 @@ updates:
...
       interval: "weekly"
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+
+  # Maintain dependencies for Docker images
+  - package-ecosystem: "docker"
```
medyagh, line 16: I don't think this is a good idea, this might break our addons or older kuberentes versions

PR #10619, opened 2021-02-25
wip: Add loading addons list in to cluster config on start

### LINKED ISSUE #10564: refactor: ClusterConfig does not store the addons map
while doing this PR https://github.com/kubernetes/minikube/pull/10427 
I noticed the addons is not saved in config when I need it

```
// ControlPlaneEndpoint returns the location where callers can reach this cluster
func ControlPlaneEndpoint(cc *config.ClusterConfig, cp *config.Node, driverName string) (string, net.IP, int, error) {
	if cc.Addons["auto-pause"] { // <---HERE IT IS it is empty map
		fmt.Println("auto-ause is on so cp port is", constants.AutoPauseProxyPort)
		cp.Port = constants.AutoPauseProxyPort
	}

```
```
 &{Name:minikube KeepContext:false EmbedCerts:false MinikubeISO: KicBaseImage:gcr.io/k8s-minikube/kicbase-builds:v0.0.17-1613861585-10427@sha256:c0827e1b58f31cbfeb871af9c7656e5822bd97dd0011b9edd45ddd1ee9ef9dfd Memory:4000 CPUs:2 DiskSize:20000 VMDriver: Driver:docker HyperkitVpnKitSock: HyperkitVSockPorts:[] DockerEnv:[] ContainerVolumeMounts:[] InsecureRegistry:[] RegistryMirror:[] HostOnlyCIDR:192.168.99.1/24 HypervVirtualSwitch: HypervUseExternalSwitch:false HypervExternalAdapter: KVMNetwork:default KVMQemuURI:qemu:///system KVMGPU:false KVMHidden:false DockerOpt:[] DisableDriverMounts:false NFSShare:[] NFSSharesRoot:/nfsshares UUID: NoVTXCheck:false DNSProxy:false HostDNSResolver:true HostOnlyNicType:virtio NatNicType:virtio SSHIPAddress: SSHUser:root SSHKey: SSHPort:22 KubernetesConfig:{KubernetesVersion:v1.20.2 ClusterName:minikube Namespace:default APIServerName:minikubeCA APIServerNames:[] APIServerIPs:[] DNSDomain:cluster.local ContainerRuntime:docker CRISocket: NetworkPlugin: FeatureGates: ServiceCIDR:10.96.0.0/12 ImageRepository: LoadBalancerStartIP: LoadBalancerEndIP: CustomIngressCert: ExtraOptions:[] ShouldLoadCachedImages:true EnableDefaultCNI:false CNI: NodeIP: NodePort:8443 NodeName:} Nodes:[{Name: IP:192.168.49.2 Port:8443 KubernetesVersion:v1.20.2 ControlPlane:true Worker:true}] Addons:map[] VerifyComponents:map[apiserver:true system_pods:true] StartHostTimeout:6m0s ScheduledStop:<nil> ExposedPorts:[] Network: MultiNodeRequested:false}todomedya: 127.0.0.1 55102 /Users/medya/.kube/config
```

BLasan:
fixes #10564 

**Before PR**
Addons map in cluster-config was not properly loaded on minikube start.

**After PR**
Addons map in cluster-config will be loading properly on minikube start

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -443,6 +443,21 @@ func Start(wg *sync.WaitGroup, cc *config.ClusterConfig, toEnable map[string]boo
 			klog.Errorf("store failed: %v", err)
 		}
 	}
+
+	loadAddonsListToClusterConfig(cc)
+
+}
+
+// load addons list with status in to cluster configs on start
+func loadAddonsListToClusterConfig(cc *config.ClusterConfig) {
+	addonsList := make(map[string]bool)
+	for addonName := range assets.Addons {
+		addonBundle := assets.Addons[addonName]
+		enabled := addonBundle.IsEnabled(cc)
+		addonsList[addonName] = enabled
+	}
+
+	cc.Addons = addonsList
 }
 
 // enableOrDisableAutoPause enables the service after the config was copied by generic enble


pkg/addons/addons.go
```diff
@@ -443,6 +443,21 @@ func Start(wg *sync.WaitGroup, cc *config.ClusterConfig, toEnable map[string]boo
 			klog.Errorf("store failed: %v", err)
 		}
 	}
+
+	loadAddonsListToClusterConfig(cc)
+
+}
+
+// load addons list with status in to cluster configs on start
+func loadAddonsListToClusterConfig(cc *config.ClusterConfig) {
```
medyagh, line 452: this doesnt seem to be the right place to fix it, by the the time that addons are enabled, the config already has the addons.

medyagh (changes requested):

pkg/addons/addons.go
BLasan: Yuh, that only happens when we enable a particular addon. Will check again and fix this

BLasan: So basically Addons filed in cluster config gets updated when we enable addons. Not in the start. But you can check that addons status by using `assets.Addons["auto_pause"].IsEnabled`. Would you mind explaining the missing point here?

BLasan: I checked by enabling the "auto-pause" addon and it's getting successfully added in tot the cluster configs. But if you need to load all the addons in to cluster-config (which is not necessary as we can achieve this through the above method I've mentioned), we'll have to do it on `minikube start` right?

medyagh: @BLasan the right place to fix this is in node package or start.go the idea is the ClusterConfig in the start.go was not filled correclty with addons, please feel free to close this PR if this seems to be a big refactor.

BLasan: This `addons.Start()` function gets called from `Start()` function inside the `pkg/minikube/node/start.go`. That `node.Start()` function gets called from `startWithDriver()` function inside the `cmd/minikube/cmd/start,go`. That function will be called by `runStart()` function which handles the execution flow of `minikube start`

PR #10638, opened 2021-02-26
test spam: add cgroup warnning to whitelist

medyagh:
currently this is the only failing test on Docker Linux Docker Runtime
https://storage.googleapis.com/minikube-builds/logs/10630/85eb356/Docker_Linux.html#fail_TestErrorSpam

and it is due to jenkins machine cgroups not allowing --memory


ilya-zuyev (approved):
lgtm

Issue #10402, opened 2021-02-08
Support containerd container runtime on darwin/linux arm with preload=false


ilya-zuyev:
Currently minikube is failed to start with `--container-runtime=containerd --preload=false`. It works fine if `--preload=true` thouth

```
minikube start --driver=docker --preload=false --container-runtime=containerd -v=8 --alsologtostderr |& tee start.log
[...]

W0208 19:32:27.810932 3580642 logs.go:135] Found kubelet problem: Feb 08 19:31:56 minikube kubelet[5559]: E0208 19:31:56.765691    5559 pod_workers.go:191] Error syncing pod c31fe6a5afdd142cf3450ac972274b36 ("etcd-minikube_kub
e-system(c31fe6a5afdd142cf3450ac972274b36)"), skipping: failed to "StartContainer" for "etcd" with CreateContainerError: "failed to create containerd container: error unpacking image: failed to resolve rootfs: content digest s
ha256:05b738aa1bc6355db8a2ee8639f3631b908286e43f584a3d2ee0c472de033c28: not found"
W0208 19:32:27.821905 3580642 logs.go:135] Found kubelet problem: Feb 08 19:32:05 minikube kubelet[5559]: E0208 19:32:05.925319    5559 pod_workers.go:191] Error syncing pod c767dbeb9ddd2d01964c2fc02c621c4e ("kube-apiserver-mi
nikube_kube-system(c767dbeb9ddd2d01964c2fc02c621c4e)"), skipping: failed to "StartContainer" for "kube-apiserver" with CrashLoopBackOff: "back-off 20s restarting failed container=kube-apiserver pod=kube-apiserver-minikube_kube
-system(c767dbeb9ddd2d01964c2fc02c621c4e)"
W0208 19:32:27.827154 3580642 logs.go:135] Found kubelet problem: Feb 08 19:32:07 minikube kubelet[5559]: E0208 19:32:07.774430    5559 pod_workers.go:191] Error syncing pod c31fe6a5afdd142cf3450ac972274b36 ("etcd-minikube_kub
e-system(c31fe6a5afdd142cf3450ac972274b36)"), skipping: failed to "StartContainer" for "etcd" with CreateContainerError: "failed to create containerd container: error unpacking image: failed to resolve rootfs: content digest s
ha256:05b738aa1bc6355db8a2ee8639f3631b908286e43f584a3d2ee0c472de033c28: not found"
[...]

PR #10642, opened 2021-02-26
Verify and fix image platform

ilya-zuyev:
Fixes https://github.com/kubernetes/minikube/issues/10402

This PR add a workaround to fix incorrect manifests in images retrieved from remote registries.

### DIFF
--- a/pkg/minikube/image/image.go
+++ b/pkg/minikube/image/image.go
@@ -19,6 +19,8 @@ package image
 import (
 	"context"
 	"fmt"
+	"github.com/docker/docker/pkg/platform"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


pkg/minikube/image/image.go
```diff
@@ -172,19 +173,45 @@ func retrieveImage(ref name.Reference) (v1.Image, error) {
...
 		return img, nil
 	}
 
 	klog.Warningf("authn lookup for %+v (trying anon): %+v", ref, err)
-	img, err = remote.Image(ref)
-	return img, err
+	return remote.Image(ref, remote.WithPlatform(p))
+}
+
+func fixPlatform(ref name.Reference, img v1.Image, p v1.Platform) (v1.Image, error) {
```
medyagh, line 195: does this hack still need to be there even with latest https://github.com/google/go-containerregistry ?
if yes I recommend creating an issue with example logs to describe it, so this be fixed in the upstream
and add that issue as a comment in this func so we know the exact reason for this hack and also help other projects to fix the bug for everyone

medyagh (changes requested):

pkg/minikube/image/image.go
medyagh: @ilya-zuyev please add the links to the issue for this hack in the comment

medyagh (changes requested):

pkg/minikube/image/image.go
ilya-zuyev: Done

pkg/minikube/image/image.go
```diff
@@ -19,6 +19,7 @@ package image
 import (
 	"context"
 	"fmt"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
```
spowelljr, line 22: Should probably group this import with the rest of the GitHub imports below instead

ilya-zuyev: good point. fixed

medyagh (approved):

PR #10656, opened 2021-03-01
CoreDNS early scale down to 1 replica

### LINKED ISSUE #10655: CoreDNS scale down to 1 replica - earlier
we use goroutine to scale down CoreDNS (at some point) during node start, while that could negatively affect other components that are initialising

currently, there are no ways we can set the number of CoreDNS replicas that get deployed initially (there were some conversations and possible plans to support that in a future kubeadm version - eg, kubernetes/kubeadm#1954, but actually kubernetes/kubernetes#85837 defaulted to 2 replicas; expected in v1beta3+?)

thus we should invoke CoreDNS rescaling as early as possible

prezha:
fixes #10655

this pr proposes to scale down the number of CoreDNS replica to 1 immediately after the master node is initialised, thus stabilising the whole system faster

also, metrics got a bit better (-5sec; both pr's `out` and released `minikube-rls` are based on the latest v1.18.0-beta.0):

```
out/minikube start  2.48s user 1.39s system 6% cpu 55.720 total
minikube-rls start  2.38s user 1.67s system 6% cpu 59.158 total
```
```
out/minikube start --vm-driver=docker --nodes=3  4.84s user 2.54s system 8% cpu 1:27.61 total
minikube-rls start --vm-driver=docker --nodes=3  4.40s user 2.70s system 7% cpu 1:32.79 total
```
```
out/minikube start --vm-driver=docker --nodes=3 --wait=all  6.52s user 2.68s system 4% cpu 3:31.89 total
minikube-rls start --vm-driver=docker --nodes=3 --wait=all  6.59s user 2.99s system 4% cpu 3:37.16 total
```

### DIFF
--- a/pkg/kapi/kapi.go
+++ b/pkg/kapi/kapi.go
@@ -211,3 +211,26 @@ func IsRetryableAPIError(err error) bool {
 func KubectlBinaryPath(version string) string {
 	return path.Join(vmpath.GuestPersistentDir, "binaries", version, "kubectl")
 }
+
+// ScaleDeployment sets the number of replicas of deployment in namespace and context
+func ScaleDeployment(context, namespace, deploymentName string, replicas int) error {
+	client, err := Client(context)
+	if err != nil {
+		return fmt.Errorf("client: %v", err)
+	}
+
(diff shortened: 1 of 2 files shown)


medyagh (approved):
thank you

PR #10694, opened 2021-03-02
Added WaitGroups to prevent stderr/stdout from being empty in error logs

### LINKED ISSUE #10570: kubeadm stderr is occasionally truncated
As noted in #10508 - I suspect there may be a recent regression in our ability to preserve kubeadm output, but it may also be a much older issue.

It might be worth checking against minikube v1.15 for instance, to see if the behavior changed, with an identical --kubernetes-version selected.

spowelljr:
Closes #10570

Issue was sometimes error logs were being written out, but stderr was empty. The cause was that there were no WaitGroups to ensure that the logs had finished piping before outputting the error, sometimes resulting with no or partial stderr logs. This fix implements WaitGroups to ensure that the logs have finished piping before outputting the error.


pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -254,11 +256,12 @@ func (k *Bootstrapper) init(cfg config.ClusterConfig) error {
 		}
 		return errors.Wrap(err, "wait")
 	}
+	kw.Close()
```
medyagh, line 259: I think this close should be after Wait

medyagh (changes requested):

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
spowelljr: When we get to this line, all the logs are in the pipeline but may not have finished being processed, so we close the pipeline. Closing the pipeline doesn't kill it, instead it writes `EOF` to the pipe and then prevents new writes. Once the pipeline is emptied it hits the `EOF` and calls `wg.Done()` (line 322) and then the wg.Wait() on the next line is unblocked.

tl;dr: If they were in the other order it would block forever because the pipeline processing would be waiting for a `EOF` which is only added on `kw.Close()`.

medyagh (approved):
lets pull upstream and verify the go mod tidy with go16

Issue #10595, opened 2021-02-24
auto-pause: make time to pause configurable


medyagh:

daehyeok:
we need to add extra flag into `start` command  because, we don't have any flag not exists for addon's option. so  is it right?
How about use `--addon-config` with `{addon_name}.{key}={value}` 
e.g.) 
```minikube start ..... --addon-config=auto-pause.time=10s```

r4j4h:
FWIW there is also `configure` inside of `minikube addons` but for configuring at time of enabling I think you are right and a new flag would be needed like `--images`. --images is used like `minikube addons enable <x> --images=image:a,image:b,image:c`

I don't know if this would belong as an argument to minikube start though, as the add-on isn't installed by default AFAIK

sharifelgamal:
@daehyeok are you still interested in this? I would be in favor of adding a new parameter to minikube start, perhaps as illustrated by @r4j4h

klaases:
Hi @vishjain, is this still something that you're working on?

PR #10427, opened 2021-02-09
auto-pause addon: automatically pause Kubernetes when not in use

medyagh:
closes https://github.com/kubernetes/minikube/issues/10326

Quick Demo
### start minikube and Enable auto-pause addon
```
medya@~/workspace/minikube (auto-pause) $ mk addons enable auto-pause
    ▪ Using image gcr.io/haproxy:2.3.5
🌟  The 'auto-pause' addon is enabled
```

### checkout the pods 

```
medya@~/workspace/minikube (auto-pause) $ kc get pods -A
NAMESPACE     NAME                               READY   STATUS    RESTARTS   AGE
auto-pause    auto-pause-proxy-dc6979c79-c2qz2   1/1     Running   0          7s
kube-system   coredns-74ff55c5b-7ks55            1/1     Running   0          6m10s
kube-system   etcd-minikube                      1/1     Running   0          6m25s
kube-system   kube-apiserver-minikube            1/1     Running   0          6m25s
kube-system   kube-controller-manager-minikube   1/1     Running   0          6m25s
kube-system   kube-proxy-578jp                   1/1     Running   0          6m10s
kube-system   kube-scheduler-minikube            1/1     Running   0          6m25s
kube-system   storage-provisioner                1/1     Running   1          6m24s
```

### check minukube status

```
medya@~/workspace/minikube (auto-pause) $ mk status
minikube
type: Control Plane
host: Running
kubelet: Running
apiserver: Running
kubeconfig: Configured
timeToStop: Nonexistent

```

### wait X minutes 
```
medya@~/workspace/minikube (auto-pause) $ date
Sun Feb 21 22:32:11 PST 2021
medya@~/workspace/minikube (auto-pause) $ mk status
minikube
type: Control Plane
host: Running
kubelet: Stopped
apiserver: Paused
kubeconfig: Configured
timeToStop: Nonexistent

```

### verify the containers paused
```
medya@~/workspace/minikube (auto-pause) $ mk ssh docker ps
CONTAINER ID   IMAGE                  COMMAND                  CREATED              STATUS                  PORTS                     NAMES
f8315dcb1bff   307c153b6ca1           "docker-entrypoint.s…"   About a minute ago   Up About a minute                                 k8s_auto-pause_auto-pause-proxy-dc6979c79-c2qz2_auto-pause_0169c9b9-a40c-475c-9b1a-dd14a64bd2ee_0
[...]

PR #10700, opened 2021-03-03
auto-pause: Using time.Ticker instead of time.Timer to fire idle time-out event.

### LINKED ISSUE #10596: auto-pause: make it memory-leak proof
each call of `time.After` creates a new timer instance which is not released until the timer fired. If we have high traffic in `incomeCh` we'll end up with a lot of pending timers

https://groups.google.com/g/golang-nuts/c/cCdm0Ixwi9A

_Originally posted by @ilya-zuyev in https://github.com/kubernetes/minikube/pull/10427#discussion_r580669849_

https://www.fatalerrors.org/a/use-with-caution-time.after-can-cause-memory-leak-golang.html

hetong07:
Fix #10596 

Using time.Ticker to replace time.Timer to solve the timer memory leak issue. Refer to [this blog](https://www.fatalerrors.org/a/use-with-caution-time.after-can-cause-memory-leak-golang.html) for more details.

### DIFF
--- a/cmd/auto-pause/auto-pause.go
+++ b/cmd/auto-pause/auto-pause.go
@@ -46,13 +46,19 @@ var runtime = "docker"
 func main() {
 	// TODO: #10595 make this configurable
 	const interval = time.Minute * 1
+	// interval must > 0 or the ticker will panic
+	if interval <= 0 {
+		exit.Message(reason.Usage, "Sorry, the auto-pause interval must be greater than 0, got {{.interval}}", out.V{"interval": interval.String()})
+	}
+	idleTimeout := time.NewTicker(interval)
+	defer idleTimeout.Stop()
+
 	// channel for incoming messages
 	go func() {
 		for {
 			// On each iteration new timer is created
 			select {
-			// TODO: #10596 make it memory-leak proof
-			case <-time.After(interval):
+			case <-idleTimeout.C:
 				runPause()
 			case <-unpauseRequests:
 				fmt.Printf("Got request\n")
--- a/cmd/auto-pause/auto-pause.go
+++ b/cmd/auto-pause/auto-pause.go
@@ -64,6 +64,15 @@ func main() {
 				fmt.Printf("Got request\n")
 				if runtimePaused {
 					runUnpause()
+					idleTimeout.Reset(interval)
+					// Drain the timeout channel in case of racing condition.
+					for {
+						select {
+						case <-idleTimeout.C:
+						default:
+							break
+						}
+					}
 				}
 
 				done <- struct{}{}


hetong07:
@medyagh Do we need to wrap the idleTimer and its parameter check into a helper function?

medyagh:
> @medyagh Do we need to wrap the idleTimer and its parameter check into a helper function? Feel like we can let PR for #10595 to make this decision?
> 
> Another question is I tried the auto-pause feature, but I notice after 1 minutes only the kubelet stopped, the apiserver is still running, shown below.
> 
> ```
> $ ./out/minikube status                                                                                                   
> minikube
> type: Control Plane
> host: Running
> kubelet: Stopped
> apiserver: Running
> kubeconfig: Configured
> timeToStop: Nonexistent
> ```
> 
> This behavior is different than the description in your original PR #10427 , is this behavior intended?

@hetong07 on my machine it works (mac docker driver)

``
$ minikube addons enable auto-pause
    ▪ Using image gcr.io/haproxy:2.3.5
    ▪ https://github.com/kubernetes/minikube/labels/co%2Fauto-pause
🌟  The 'auto-pause' addon is enabled

$ docker ps
CONTAINER ID   IMAGE                                 COMMAND                  CREATED         STATUS         PORTS                                                                                                                                  NAMES
c94347742976   gcr.io/k8s-minikube/kicbase:v0.0.18   "/usr/local/bin/entr…"   4 minutes ago   Up 4 minutes   127.0.0.1:55034->22/tcp, 127.0.0.1:55033->2376/tcp, 127.0.0.1:55032->5000/tcp, 127.0.0.1:55031->8443/tcp, 127.0.0.1:55030->32443/tcp   minikube
10:32:38 medya/workspace/minikube
pod_preset •3 ⌅1
$ code ~/.kube/config
10:33:27 medya/workspace/minikube
pod_preset •3 ⌅1
$ minikube status
minikube
type: Control Plane
host: Running
kubelet: Stopped
apiserver: Paused
kubeconfig: Configured
timeToStop: Nonexistent

```

File: cmd/auto-pause/auto-pause.go [hunk 1]
```diff
@@ -46,13 +46,19 @@ var runtime = "docker"
...
+	defer idleTimeout.Stop()
+
 	// channel for incoming messages
 	go func() {
 		for {
 			// On each iteration new timer is created
 			select {
-			// TODO: #10596 make it memory-leak proof
-			case <-time.After(interval):
+			case <-idleTimeout.C:
```
ilya-zuyev, line 61: We need to reset the ticker after we get an 'unpauseRequest'. Otherwise, it's possible that, say, we have 1 minute ticker interval, and 59 seconds elapsed since the last tick. We get an event in `unpauseRequest`, unpause minikube, but the very next second receive an event from ticker and pause it again

ilya-zuyev (changes requested):
Please reset the ticker after each 'unpause' event

hetong07:
@medyagh strange, I repeadly see the API service running, maybe configuration issue?
```
$ ./out/minikube status                                                                                                    
minikube
type: Control Plane
host: Running
kubelet: Stopped
apiserver: Running
kubeconfig: Configured
timeToStop: Nonexistent

```

cmd/auto-pause/auto-pause.go
```diff
@@ -46,18 +46,25 @@ var runtime = "docker"
...
 				runPause()
 			case <-unpauseRequests:
 				fmt.Printf("Got request\n")
 				if runtimePaused {
 					runUnpause()
+					idleTimeout.Reset(interval)
```
ilya-zuyev, line 67: There is a race condition here still. We can have a tick between `runUnpause` and `idleTimeout.Reset(interval)`. In this case on next `for` iteration we'll have an event in `idleTimeout.C` and do pause.

ilya-zuyev (changes requested):

cmd/auto-pause/auto-pause.go
hetong07: @ilya-zuyev Yes, I also think of that possibility, but I am not sure whether reset() could clear existing & pending tick? I do some investigation and could not find the answer. if it is indeed the case, we should recreate a ticker....

hetong07: Also, an more easier way is to reset the ticker before `runUnpause()` and then reset it again after
```
idleTimeout.Reset(/*very long inverval*/)
runUnpause()
idleTimeout.Reset(interval)
```
But this may introduce a potential problem of what is the proper value for the long interval.
We can discuss offline to be more efficient.

hetong07: If we recreate a timer, we cannot use defer to stop the timer? Since the defer target is determined at the time defer is executed? Adding stop() on each exist route may cause the code ugly either.

Another way is:
```
idleTimeout.Reset(/* longer interval*/)
runUnpause()
idleTimeout.Reset(interval)
```
But it introduces a protential problem: what is the long internal number?
We can discuss on slack offline.

hetong07: Or maybe we can try drain the ticker channel once after reset?
```
runUnpause()
idleTimeout.Reset(interval)
for {
    select {
    case <-idleTimeout.C:
    default:
       break
    }
}
```

ilya-zuyev: @hetong07 Sorry for late response. Yeah, this solution with draining `idleTimeout.C` looks good to me

azhao155:
@medyagh @ilya-zuyev why do we need to reset ticker? according to the blog. it should be reset by itself:

"It is good to use the ticker timer directly, because the ticker is executed once every certain period of time. Generally, it can be executed multiple times, which is equivalent to calling timer timer time.Reset ."

hetong07:
> @medyagh @ilya-zuyev why do we need to reset ticker? according to the blog. it should be reset by itself:
> 
> "It is good to use the ticker timer directly, because the ticker is executed once every certain period of time. Generally, it can be executed multiple times, which is equivalent to calling timer timer time.Reset ."

I think we have to reset it due to the racing conditions, as @ilya-zuyev stated in previous comments.

ilya-zuyev (approved):
lgtm

andriyDev:
@hetong07 Unsure if you're still around, we'd love to merge this PR. We just need it to be rebased! If we don't hear from you in a few days, I may just take ownership of this PR (creating a new PR with the same content).

PR #10711, opened 2021-03-04
Remove WSLENV empty check from IsMicrosoftWSL

### LINKED ISSUE #10689: windows  MK_WRONG_BINARY_WSL even when not in powershell
The offending function below is incorrectly assuming that if `WSLENV` has a non-empty value, that the shell is an WSL shell. According to [this](https://devblogs.microsoft.com/commandline/share-environment-vars-between-wsl-and-windows/) post, this environment variable is used for sharing variables into new WSL sessions, and is set within Windows too.

https://github.com/kubernetes/minikube/blob/51ec098b056e379d42ed8a9aae3d530309fe508d/pkg/minikube/driver/driver.go#L174-L176

1. Ensure the `WSLENV` environment variable has a non-empty value. Mine is `WT_SESSION::WT_PROFILE_ID`
2. Run `minikube` from a regular Windows `cmd.exe` shell
3. Observe the incorrect output
```
> minikube docker-env

❌  Exiting due to MK_WRONG_BINARY_WSL: You are trying to run windows .exe binary inside WSL, for better integration please use Linux binary instead (Download at https://minikube.sigs.k8s.io/docs/start/.). Otherwise if you still want to do this, you can do it using --force
```

spowelljr:
Closes #10689

Removed WSLENV empty check from IsMicrosoftWSL due to Windows Terminal setting the variable outside of WSL.

https://github.com/kubernetes/minikube/pull/10711#issuecomment-790207680
> We have code that tries to detect if we're trying to run a Windows executable within a WSL shell, which should be running a Linux binary. The most reliable way to do that is to check environment variables that WSL sets. Unfortunately, one of the 3 variables we picked (WSLENV) is set by Windows Terminal, outside of WSL, as a way of carrying data across connections. This is causing some Windows users to not be able to use minikube in a valid config.

### DIFF
--- a/pkg/minikube/driver/driver.go
+++ b/pkg/minikube/driver/driver.go
@@ -174,7 +174,7 @@ func NeedsPortForward(name string) bool {
 // checking for WSL env var based on this https://github.com/microsoft/WSL/issues/423#issuecomment-608237689
 // also based on https://github.com/microsoft/vscode/blob/90a39ba0d49d75e9a4d7e62a6121ad946ecebc58/resources/win32/bin/code.sh#L24
 func IsMicrosoftWSL() bool {
-	return os.Getenv("WSL_DISTRO_NAME") != "" || os.Getenv("WSLPATH") != "" || os.Getenv("WSLENV") != ""
+	return os.Getenv("WSL_DISTRO_NAME") != "" || os.Getenv("WSLPATH") != ""
 }
 
 // HasResourceLimits returns true if driver can set resource limits such as memory size or CPU count.


tstromberg:
> Closes #10689
> 
> Removed WSLENV empty check from IsMicrosoftWSL due to Windows Terminal setting the variable outside of WSL.

Could you add more detail to the PR description? 

I don't quite understand the problem that is being solved, and why this is the best approach to do so.

sharifelgamal:
We have code that tries to detect if we're trying to run a Windows executable within a WSL shell, which should be running a Linux binary. The most reliable way to do that is to check environment variables that WSL sets. Unfortunately, one of the 3 variables we picked (WSLENV) can be set in Powershell, outside of WSL, as a way of carrying data across connections. This is causing some Windows users to not be able to use minikube in a valid config.

NatsumiHB (approved):
Just built this for Windows and can confirm this fixes the issue.

PR #10714, opened 2021-03-04
Correctly detect powershell when minikube is called on windows through SSH

### LINKED ISSUE #10626: Failing test: TestForceSystemdFlag on Windows Docker
https://storage.googleapis.com/minikube-builds/logs/10603/fb9f08b/Docker_Windows.html#fail_TestForceSystemdFlag

hetong07:
Fix #10626 

Make the minikube correctly detect it is running in powershell.

**Before PR**:
```
PS > \minikube-windows-amd64.exe docker-env
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.99.101:2376"
export DOCKER_CERT_PATH="C:\Users\imwise\.minikube\certs"

```
**After PR**:
```
PS > .\minikube-windows-amd64.exe docker-env
$Env:DOCKER_TLS_VERIFY = "1"
$Env:DOCKER_HOST = "tcp://192.168.99.101:2376"
$Env:DOCKER_CERT_PATH = "C:\Users\imwise\.minikube\certs"
$Env:MINIKUBE_ACTIVE_DOCKERD = "minikube"
# & minikube -p minikube docker-env | Invoke-Expression

```

### DIFF
--- a/cmd/minikube/cmd/docker-env.go
+++ b/cmd/minikube/cmd/docker-env.go
@@ -26,6 +26,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -250,7 +251,7 @@ var dockerEnvCmd = &cobra.Command{
 			}
 		}
 		sh := shell.EnvConfig{
-			Shell: shl,
+			Shell: strings.TrimSuffix(shl, filepath.Ext(shl)),
 		}
 
 		if dockerUnset {
--- a/go.mod
+++ b/go.mod
@@ -1,10 +1,10 @@
 module k8s.io/minikube
 
-go 1.15
+go 1.16
 
 require (
 	cloud.google.com/go/storage v1.13.0
-	contrib.go.opencensus.io/exporter/stackdriver v0.13.5
+	contrib.go.opencensus.io/exporter/stackdriver v0.12.1
 	github.com/Azure/azure-sdk-for-go v42.3.0+incompatible
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace v0.16.0
 	github.com/Microsoft/go-winio v0.4.15-0.20190919025122-fc70bd9a86b5 // indirect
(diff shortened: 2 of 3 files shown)


cmd/minikube/cmd/docker-env.go
```diff
@@ -245,6 +245,7 @@ var dockerEnvCmd = &cobra.Command{
 		shl := shell.ForceShell
 		if shl == "" {
 			shl, err = shell.Detect()
+			shl = strings.Trim("powershell.exe", ".exe")
```
ilya-zuyev: The result of `strings.Trim("powershell.exe", ".exe")` is always `powershell`. Here we effectively just set shl = "powershell" and ignore the result of `shell.Detect()`. Not sure it's that we really need here.

medyagh: @hetong07 this seems like a mistake to me, I wonder if u meant to do something else ? simmilar to ilya I suggest we fix the shell.Detect() logic so it does the right thing

cmd/minikube/cmd/docker-env.go
```diff
@@ -245,6 +246,7 @@ var dockerEnvCmd = &cobra.Command{
 		shl := shell.ForceShell
 		if shl == "" {
 			shl, err = shell.Detect()
+			shl = strings.TrimSuffix(shl, filepath.Ext(shl))
```
ilya-zuyev, line 249: Would be a little bit better to first check the error and then if it's nil trim the '.exe'

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.15
 
 require (
 	cloud.google.com/go/storage v1.13.0
-	contrib.go.opencensus.io/exporter/stackdriver v0.12.1
+	contrib.go.opencensus.io/exporter/stackdriver v0.13.5
```
ilya-zuyev, line 7: If this PR can work with old versions of dependencies, it's better to keep it and update go.[mod|sum] in a separate PR

hetong07: Directly copied and replace the go.mod and go.sum file from minikube master branch will resolve this issue.

cmd/minikube/cmd/docker-env.go
hetong07: @medyagh and @ilya-zuyev Sorry I mistakenly upload the experimental changes here, and have fixed it in subsequent CL, do I need to make a --force push to overwrite this commit?

hetong07: @medyagh and @ilya-zuyev do a check and find my branch has divert from the minikube HEAD, so I will close issue and create another PR to solve this problem.

Issue #9809, opened 2020-11-30
Add static IP for KVM driver


medyagh:
similar to our docker and podman driver implementation

RA489:
if we create custom network for KVM , we can set a fixed ip when running the container --ip and hopefully we don't have conflict with subnets with existing networks as KVM has fixed subnet and if it doesn't we can try other options https://github.com/kubernetes/minikube/blob/master/pkg/drivers/kvm/network.go#L41

medyagh:
this task is assigned to @prezha

PR #10716, opened 2021-03-04
[do-not-merge] Add static IP for KVM driver

RA489:
Add static IP for KVM driver
fixes https://github.com/kubernetes/minikube/issues/9809

### DIFF
--- a/pkg/drivers/kvm/network_create.go
+++ b/pkg/drivers/kvm/network_create.go
@@ -1,0 +1,245 @@
+/*
+Copyright 2021 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


pkg/drivers/kvm/network_create.go
```diff
@@ -0,0 +1,245 @@
...
+func networkExists(ociBin string, name string) bool {
+	_, err := ContainerNetworkInspect(ociBin, name)
+	if err != nil && !errors.Is(err, ErrNetworkNotFound) { // log unexpected error
+		klog.Warningf("Error inspecting docker network %s: %v", name, err)
+	}
+	return err == nil
+}
+
+// networkNameByLabel returns all network names created by a label
+func networkNameByLabel(ociBin string, label string) ([]string, error) {
```
medyagh, line 213: this is not right this is not docker driver, there is no OCI in KVM

medyagh (changes requested):

medyagh (changes requested):
this is a wrong PR, this is not doing the right thing. this PR is trying to create docker networks for KVM drivers.

PR #10718, opened 2021-03-04
Release 1.18.1

ilya-zuyev:


medyagh (approved):

Makefile
```diff
@@ -15,7 +15,7 @@
 # Bump these on release - and please check ISO_VERSION for correctness.
 VERSION_MAJOR ?= 1
 VERSION_MINOR ?= 18
-VERSION_BUILD ?= 0
+VERSION_BUILD ?= 1
```
spowelljr, line 18: Bump the ISO version as well I believe

spowelljr (changes requested):

Makefile
spowelljr: Actually, with this comment I'm not sure...
```
# Default to .0 for higher cache hit rates, as build increments typically don't require new ISO versions
```

ilya-zuyev: did we change anything in iso?

spowelljr: I don't think so, I'm just not sure if we duplicate the v1.18.0 ISO and rename it to v1.18.1 or not

spowelljr: Based on the comment I don't think we do

medyagh: we didnt change ISO so no need to do that

spowelljr (approved):

PR #10732, opened 2021-03-06
The cluster doesn't have to be healthy for kubectl

### LINKED ISSUE #10717: website/handbook: don't advise users to create a symbolic link to minikube’s binary named ‘kubectl’
this confuses users and doesn't work (eg, minikube calling kubectl would loopback to itself)
ref: https://minikube.sigs.k8s.io/docs/handbook/kubectl/ (https://github.com/kubernetes/minikube/blob/master/site/content/en/docs/handbook/kubectl.md)

```
Alternatively, you can create a symbolic link to minikube’s binary named ‘kubectl’.

ln -s $(which minikube) /usr/local/bin/kubectl
```

example: https://github.com/kubernetes/minikube/issues/9976

previously, there were discussions in issue #8857 and pr #8872

alternatively, we can make it work by 'recognising' that minikube was called as kubectl and assign the flags as flags for minikube's own downloaded kubectl binary...

afbjorklund:
We only want to match the kubectl version with it

Closes #10717

BEFORE

```console
$ kubectl version --client --output=json
```

AFTER

```console
$ kubectl version --client --output=json
{
  "clientVersion": {
    "major": "1",
    "minor": "20",
    "gitVersion": "v1.20.2",
    "gitCommit": "faecb196815e248d3ecfb03c680a4507229c2a56",
    "gitTreeState": "clean",
    "buildDate": "2021-01-13T13:28:09Z",
    "goVersion": "go1.15.5",
    "compiler": "gc",
    "platform": "linux/amd64"
  }
}
```

### DIFF
--- a/cmd/minikube/cmd/kubectl.go
+++ b/cmd/minikube/cmd/kubectl.go
@@ -24,8 +24,8 @@ import (
 
 	"github.com/spf13/cobra"
 	"k8s.io/klog/v2"
+	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/constants"
-	"k8s.io/minikube/pkg/minikube/mustload"
 	"k8s.io/minikube/pkg/minikube/node"
 	"k8s.io/minikube/pkg/minikube/out"
 )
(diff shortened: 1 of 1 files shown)


afbjorklund:
Still not fully fixed, since mustload will abort if there is no config:

```
$ kubectl --client version
* Profile "minikube" not found. Run "minikube profile list" to view all profiles.
  - To start a cluster, run: "minikube start
```

Maybe it should just have used constants.DefaultKubernetesVersion

medyagh (approved):

medyagh (changes requested):
how about adding an integration test for this ? so on a NOT running minikube profile (maybe start with --dry-run) and then try kubectl on it ? similar to the kubectl subtest in functional test?

PR #10743, opened 2021-03-07
Add instructions for building images with buildkit

### LINKED ISSUE #10685: Document how to use buildkit remotely, for building images
Currently we only have instructions on how to use it locally.

We should also add instructions on how to to use it remotely.

Unfortunately it is currently still too complicated for a `buildkit-env`

But there are manual workarounds meanwhile, like a ssh tunnel

afbjorklund:
Documentation still refers to building as "pushing"

Closes #10685


afbjorklund:
For some reason, buildkit looks slightly broken with the docker driver (but works with the virtualbox driver)

`error: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to read dockerfile: failed to mount sq23m4z9cl0p619chgfvd5sso: snapshot sq23m4z9cl0p619chgfvd5sso does not exist: not found
`

medyagh (approved):

PR #10767, opened 2021-03-10
print out error on failed iso and kicbase builds

sharifelgamal:

### DIFF
--- a/hack/jenkins/build_iso.sh
+++ b/hack/jenkins/build_iso.sh
@@ -40,12 +40,16 @@ else
 	export ISO_BUCKET
 fi
 
-make release-iso
+make release-iso | tee iso-logs.txt
 # Abort with error message if above command failed
 ec=$?
 if [ $ec -gt 0 ]; then
 	if [ "$release" = false ]; then
-		gh pr comment ${ghprbPullId} --body "Hi ${ghprbPullAuthorLoginMention}, building a new ISO failed, please try again."
+		err=$(tail -50 iso-logs.txt)
+		gh pr comment ${ghprbPullId} --body "Hi ${ghprbPullAuthorLoginMention}, building a new ISO failed, with the error below:
+		
+		${err}
+		"
 	fi
 	exit $ec
 fi
--- a/hack/jenkins/kicbase_auto_build.sh
+++ b/hack/jenkins/kicbase_auto_build.sh
@@ -61,13 +61,17 @@ export KICBASE_IMAGE_REGISTRIES="${GCR_IMG} ${DH_IMG}"
 
 
 # Build a new kicbase image
-yes|make push-kic-base-image
+yes | make push-kic-base-image | tee kic-logs.txt
 
 # Abort with error message if above command failed
 ec=$?
 if [ $ec -gt 0 ]; then
 	if [ "$release" = false ]; then
-		gh pr comment ${ghprbPullId} --body "Hi ${ghprbPullAuthorLoginMention}, building a new kicbase image failed, please try again."
+		err=$(tail -50 kic-logs.txt)
+		gh pr comment ${ghprbPullId} --body "Hi ${ghprbPullAuthorLoginMention}, building a new kicbase image failed, with the error below:
+		
+		${err}
+		"
 	fi
 	exit $ec
 fi


hack/jenkins/build_iso.sh
```diff
@@ -40,12 +40,16 @@ else
...
+make release-iso | tee iso-logs.txt
 # Abort with error message if above command failed
 ec=$?
 if [ $ec -gt 0 ]; then
 	if [ "$release" = false ]; then
-		gh pr comment ${ghprbPullId} --body "Hi ${ghprbPullAuthorLoginMention}, building a new ISO failed, please try again."
+		err=$(tail -50 iso-logs.txt)
+		gh pr comment ${ghprbPullId} --body "Hi ${ghprbPullAuthorLoginMention}, building a new ISO failed, with the error below:
+		
+		${err}
```
medyagh, line 51: how about we make this
<details>
<pre>
${err}

</pre>
</details>

hack/jenkins/kicbase_auto_build.sh
```diff
@@ -61,13 +61,17 @@ export KICBASE_IMAGE_REGISTRIES="${GCR_IMG} ${DH_IMG}"
...
-yes|make push-kic-base-image
+yes | make push-kic-base-image | tee kic-logs.txt
 
 # Abort with error message if above command failed
 ec=$?
 if [ $ec -gt 0 ]; then
 	if [ "$release" = false ]; then
-		gh pr comment ${ghprbPullId} --body "Hi ${ghprbPullAuthorLoginMention}, building a new kicbase image failed, please try again."
+		err=$(tail -50 kic-logs.txt)
+		gh pr comment ${ghprbPullId} --body "Hi ${ghprbPullAuthorLoginMention}, building a new kicbase image failed, with the error below:
```
medyagh, line 71: same here

how about we make this
<details>
<pre>
${err}

</pre>
</details>

hack/jenkins/kicbase_auto_build.sh
```diff
@@ -61,13 +61,17 @@ export KICBASE_IMAGE_REGISTRIES="${GCR_IMG} ${DH_IMG}"
...
 # Build a new kicbase image
-yes|make push-kic-base-image
+yes | make push-kic-base-image | tee kic-logs.txt
 
 # Abort with error message if above command failed
 ec=$?
 if [ $ec -gt 0 ]; then
 	if [ "$release" = false ]; then
-		gh pr comment ${ghprbPullId} --body "Hi ${ghprbPullAuthorLoginMention}, building a new kicbase image failed, please try again."
+		err=$(tail -50 kic-logs.txt)
```
medyagh, line 70: why not 1000 lines ? (kic logs are small) and we can fit them in a foldable <details> <pre> block

hack/jenkins/build_iso.sh
```diff
@@ -40,12 +40,16 @@ else
...
 fi
 
-make release-iso
+make release-iso | tee iso-logs.txt
 # Abort with error message if above command failed
 ec=$?
 if [ $ec -gt 0 ]; then
 	if [ "$release" = false ]; then
-		gh pr comment ${ghprbPullId} --body "Hi ${ghprbPullAuthorLoginMention}, building a new ISO failed, please try again."
+		err=$(tail -50 iso-logs.txt)
```
medyagh, line 48: 50 lines would probaby not give much info lets do 1000
would it be possible to upload the log and provide link to the file ?

medyagh (changes requested):

hack/jenkins/build_iso.sh
sharifelgamal: my thought here is usually the error is printed right at the end of the file. i wanted to display the error without too much other cruft that can overwhelms us with unnecessary info

hack/jenkins/build_iso.sh
sharifelgamal: the logs are already uploaded, the tee is only specifically for this purpose.

hack/jenkins/kicbase_auto_build.sh
sharifelgamal: I can add the foldable details, my worry is supplying too much useless information. 90% of the time when debugging an issue with these scripts, I just scroll right to the end of the script, and I wanted to avoid having to do this but just showing the end of the script.

medyagh (approved):

Issue #10773, opened 2021-03-10
docker: Update prune suggestion to exclude minikube containers


tejal29:
I love`minikube pause` a lot. It saves me CPU, makes my laptop faster. 

However, after running a bunch of docker builds, my laptop ran out of memory.  I decided to run `docker system prune`
which ended up deleting minikube stopped containers.

```
➜  skaffold git:(fix_release) ✗ docker system prune 
WARNING! This will remove:
  - all stopped containers
  - all networks not used by at least one container
  - all dangling images
  - all dangling build cache

Are you sure you want to continue? [y/N] y
Deleted Containers:
8a6ffa08e4c50f7759071a7e047db6e43f22da65f8c865e791011ce50c1b8958
416a5a5eefbb25683ce97867110287183a44e5d12b9a314524ae17dfa7bac4d0
d872cdd0d0fbd57110c9350770f0b1537278398c8e8fb117529a7135f7607960
ec6ec9a968379ed54362127f2546e2ec3d779fae72445b8c846257202f2f81bf
21d3ef81b35b96f96a91c24a049a59b2cd4663d796f93945c75a154bfb332ac1

Deleted Networks:
minikube

Deleted Images:
deleted: sha256:fb344716f42ca1d99b1aa3352aeb26bcd3365b2bf8f93c3fdd290b7afc4a6dd8
deleted: sha256:09591f14029b906d83cc2305a5f33d7afb4fc7afa38a4066acd539405316362a
deleted: sha256:ed7e58353be3e7a7e3d8b0585da0cadec53ea4372bb47069533a9c67ea896c7a
....
deleted: sha256:d7547014673a01801c9cb33f3c69819930a24c7db96a4826de5e6ee1ea0ed27a

Total reclaimed space: 6.657GB
```

Issues
1) `docker system prune` delete stopped containers and i had totally forgotten this will delete my minikube.
  - would have appreciated some notice when i ran `minikube pause` e.g. `Warning: docker system prune will delete paused minikube`
2) Now, i have to create minikube cluster again :(

**Expected**
Is there a way we would not delete minikube containers?

```
➜  skaffold git:(fix_release) ✗ minikube status
E0310 09:47:33.266232    8299 status.go:246] status error: host: state: unknown state "minikube": docker container inspect minikube --format={{.State.Status}}: exit status 1
stdout:


stderr:
Error: No such container: minikube
E0310 09:47:33.266625    8299 status.go:249] The "minikube" host does not exist!
minikube
[...]

PR #10801, opened 2021-03-12
update prune instructions to not prune minikube itself

medyagh:
closes https://github.com/kubernetes/minikube/issues/10773

### Before this PR
```
$ mk start
😄  minikube v1.18.1 on Darwin 11.2.3
❗  Executing "docker container inspect minikube --format={{.State.Status}}" took an unusually long time: 140.562388ms
💡  Restarting the docker service may improve performance.
🐳  Preparing Kubernetes v1.20.2 on Docker 20.10.3 ...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v4
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by defaul
```

### After this PR
fix the suggestion not to prune minikube itself (if it is stopped)

```
😄  minikube v1.18.1 on Darwin 11.2.3
❗  Executing "docker container inspect minikube --format={{.State.Status}}" took an unusually long time: 196.475601ms
💡  Restarting or pruning the docker service may improve performance.
    ▪ To prune run: $ docker system prune --filter="label!=created_by.minikube.sigs.k8s.io=true"
    ▪ To restart: Click Docker 🐳 in the system tray and then click "Restart Docker"
🐳  Preparing Kubernetes v1.20.2 on Docker 20.10.3 ...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v4
```

### DIFF
--- a/pkg/drivers/kic/oci/cli_runner.go
+++ b/pkg/drivers/kic/oci/cli_runner.go
@@ -145,7 +145,7 @@ func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
 			warnLock.Unlock()
 
 			if !ok {
-				out.WarningT(`Executing this command "{{.command}}" took an unusually long time: {{.duration}}`, out.V{"command": rr.Command(), "duration": elapsed})
+				out.WarningT(`Executing "{{.command}}" took an unusually long time: {{.duration}}`, out.V{"command": rr.Command(), "duration": elapsed})
 				// Don't show any restarting hint, when running podman locally (on linux, with sudo). Only when having a service.
 				if cmd.Args[0] != "sudo" {
 					out.ErrT(style.Tip, `Restarting or pruning the {{.name}} service may improve performance.`, out.V{"name": cmd.Args[0]})
--- a/pkg/drivers/kic/oci/cli_runner.go
+++ b/pkg/drivers/kic/oci/cli_runner.go
@@ -145,11 +145,17 @@ func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
 			warnLock.Unlock()
 
 			if !ok {
-				out.WarningT(`Executing "{{.command}}" took an unusually long time: {{.duration}}`, out.V{"command": rr.Command(), "duration": elapsed})
+				out.WarningT(`Executing this command "{{.command}}" took an unusually long time: {{.duration}}`, out.V{"command": rr.Command(), "duration": elapsed})
 				// Don't show any restarting hint, when running podman locally (on linux, with sudo). Only when having a service.
 				if cmd.Args[0] != "sudo" {
-					out.ErrT(style.Tip, `Restarting the {{.name}} service may improve performance.`, out.V{"name": cmd.Args[0]})
+					out.ErrT(style.Tip, `Restarting or pruning the {{.name}} service may improve performance.`, out.V{"name": cmd.Args[0]})
+					out.Infof(`To prune run: $ docker system prune --filter="label!=created_by.minikube.sigs.k8s.io=true"`)
+					if runtime.GOOS != "linux" {
+						out.Infof(`To restart: Click Docker 🐳 in the system tray and then click "Restart Docker"`, out.V{"icon": style.Docker})
+					}
+
 				}
+
 			}
 		}
 
--- a/pkg/minikube/reason/reason.go
+++ b/pkg/minikube/reason/reason.go
@@ -173,8 +173,9 @@ var (
 		ID:       "RSRC_DOCKER_STORAGE",
 		ExitCode: ExInsufficientStorage,
 		Advice: `Try one or more of the following to free up space on the device:
-	
-			1. Run "docker system prune" to remove unused Docker data (optionally with "-a")
+			1. Remove unused Docker data (optionally with "-a") 
+			Run:
+				$ docker system prune --filter="label!=created_by.minikube.sigs.k8s.io=true"
 			2. Increase the storage allocated to Docker for Desktop by clicking on:
 				Docker icon > Preferences > Resources > Disk Image Size
 			3. Run "minikube ssh -- docker system prune" if using the Docker container runtime`,
--- a/site/content/en/docs/drivers/docker.md
+++ b/site/content/en/docs/drivers/docker.md
@@ -15,7 +15,7 @@ The Docker driver allows you to install Kubernetes into an existing Docker insta
 
 - Cross platform (linux, macOS, Windows)
 - No hypervisor required when run on Linux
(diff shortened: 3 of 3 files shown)


tstromberg:
Could you please provide more context? I don't yet agree with this PR.

* Pruning containers should not improve performance. The containers are after all, stopped. Please provide evidence otherwise.
* Pruning does not affect usage of 'minikube pause'. It does not stop containers.

medyagh:
> Could you please provide more context? I don't yet agree with this PR.
> 
> * Pruning containers should not improve performance. The containers are after all, stopped. Please provide evidence otherwise.
> * Pruning does not affect usage of 'minikube pause'. It does not stop containers.

@tstromberg I agree it is probably not an advice that will help everybody, but in some cases I have seen when users have hundreds of dangling images, docker ps and docker imgages is very slow (due to docker calucating hashesh)
so while it might not help eveyrone, but it might help someone and it is the best advice we got anyways (if they reach out for us to help)

not strong opinion though, I am okay with removing the prune, wdyt ?

medyagh:
> * Pruning does not affect usage of 'minikube pause'. It does not stop containers.

@tstromberg this PR doesnt affec the Paused clusters, only adds more specific instructions for pruninng, so the user doesn't remove minikube itself in the process of prunning

pkg/drivers/kic/oci/cli_runner.go
```diff
@@ -148,8 +148,14 @@ func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
 				out.WarningT(`Executing "{{.command}}" took an unusually long time: {{.duration}}`, out.V{"command": rr.Command(), "duration": elapsed})
 				// Don't show any restarting hint, when running podman locally (on linux, with sudo). Only when having a service.
 				if cmd.Args[0] != "sudo" {
```
tstromberg, line 150: The logic to determine if this is Docker should be improved so as not to require the clarifying comment. It's pretty obscure.

pkg/drivers/kic/oci/cli_runner.go
```diff
@@ -148,8 +148,14 @@ func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
...
-					out.ErrT(style.Tip, `Restarting the {{.name}} service may improve performance.`, out.V{"name": cmd.Args[0]})
+					out.ErrT(style.Tip, `Restarting or pruning the {{.name}} service may improve performance.`, out.V{"name": cmd.Args[0]})
+					out.Infof(`To prune run: $ docker system prune --filter="label!=created_by.minikube.sigs.k8s.io=true"`)
```
tstromberg, line 152: This still intuitively doesn't make much sense to me. Can you provide some additional context, such as a documentation from Docker or a benchmark? I want to make sure we aren't selling snake-oil here.

site/content/en/docs/drivers/docker.md
```diff
@@ -35,6 +35,14 @@ The Docker driver allows you to install Kubernetes into an existing Docker insta
 
 [comment]: <> (this title is used in the docs links, don't change)
 
+### Prune docker to avoid performance downgrade in minikube
+ the following command will prune all on unused/stopped containers, volumes on your docker.
```
tstromberg, line 39: Provide contextual documentation on why this improves performance.

Also, remove the leading whitespace and capitalize "the".

pkg/minikube/reason/reason.go
```diff
@@ -173,8 +173,9 @@ var (
 		ID:       "RSRC_DOCKER_STORAGE",
 		ExitCode: ExInsufficientStorage,
 		Advice: `Try one or more of the following to free up space on the device:
-	
-			1. Run "docker system prune" to remove unused Docker data (optionally with "-a")
+			1. Remove unused Docker data (optionally with "-a")
```
tstromberg, line 176: Should this should say "Remove Docker daemon generated outside of minikube"?

tstromberg (changes requested):

pkg/drivers/kic/oci/cli_runner.go
medyagh: This is a compromise that @afbjorklund suggested to avoid code complexity and since we do not allow running docker with “Sudo” , it is also most efficient way

pkg/drivers/kic/oci/cli_runner.go
medyagh: One systems which user has hundreds of dangling images without owner, docker commands do take longer.

It is not really selling snake oil, It is the only suggestions that we got, and might help someone, specially if they wanna do prune on their own (such as tejal, we better at least give them the Right prune command)

I would be happy to change the wording to say, This MIGH not help u at all, might 10% chance this might be the problem. *but that would make the dislclaimer text long*


We already tell the suer “May improve”

And to be clear, on some slow dockers that I noticed,  Restarting didn’t do anything either.

And if the user comes creates issue in github, I would tell them same suggestion, restart, prune,... there isn’t much else to do tbh

pkg/minikube/reason/reason.go
medyagh: I am not sure if talking about outside and inside minikbue would make sense for the average end user. That might confuse them even more.
Since we give them the shell command with $ ... to copy paste.

If we wanted inside minikube we would give them this command
$ minikbue ssh docker system prune ...

pkg/drivers/kic/oci/cli_runner.go
tstromberg: If the runner can't be made aware of which environment it is in, perhaps it's a good sign that it should be passing the signal upwards rather than providing advice for an environment. For instance, returning duration upwards in RunResult.

If that isn't feasible, lets simplify this check, and make it more accurate at the same time. Is there a way to check if the OCI is Docker (based on either the object, cmd.Args, or some other magic)? Ideally, this should say something like:

```
if isDocker() {
  .. show docker advice ...
}
```

For instance, this function could simply check if cmd arg 0 or 1 is `docker`.

pkg/drivers/kic/oci/cli_runner.go
tstromberg: Here's my recommendation then to take a data driven approach to this:

* Starting 10,000 abandoned containers.
* Run minikube and note the run durations mentioned in the log file (you mention `ps`). minikube should complain about slowdowns.
* prune
* Run minikube and note the run durations that changed since prune
* Adjust this logic to only suggest prune for the commands which would in fact be improved, such as `ps` as you have mentioned.

pkg/minikube/reason/reason.go
tstromberg: By outside of minikube, I mean images that were not created by minikube, as this command explicitly filters those out. I'm not sure of the exact best verbiage here, but the comment is no longer accurate to the command it describes.

tstromberg (changes requested):

PR #10806, opened 2021-03-13
make sure path to logs directory exists

### LINKED ISSUE #10804: TestLogFile/SetLogFile fails: open ~/.minikube/logs/audit.json: no such file or directory
if `~/.minikube` folder structure does not exist (incl. after `minikube delete --all --purge`), the test will fail:

```
--- FAIL: TestLogFile (0.00s)
    --- FAIL: TestLogFile/SetLogFile (0.00s)
        logFile_test.go:30: unable to open /home/prezha/.minikube/logs/audit.json: open /home/prezha/.minikube/logs/audit.json: no such file or directory
    --- FAIL: TestLogFile/AppendToLog (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xbec6fd]

goroutine 33 [running]:
testing.tRunner.func1.2(0xcb2d20, 0x13ec100)
        /usr/local/go/src/testing/testing.go:1144 +0x332
testing.tRunner.func1(0xc000583380)
        /usr/local/go/src/testing/testing.go:1147 +0x4b6
panic(0xcb2d20, 0x13ec100)
        /usr/local/go/src/runtime/panic.go:965 +0x1b9
k8s.io/minikube/pkg/minikube/audit.TestLogFile.func2(0xc000583380)
        /home/prezha/dev/k8s/minikube/github.com/prezha/minikube/pkg/minikube/audit/logFile_test.go:41 +0x15d
testing.tRunner(0xc000583380, 0xe2db30)
        /usr/local/go/src/testing/testing.go:1194 +0xef
created by testing.(*T).Run
        /usr/local/go/src/testing/testing.go:1239 +0x2b3
FAIL    k8s.io/minikube/pkg/minikube/audit      0.051s
```

prezha:
fixes #10804

the error description is in the issue, here is a small fix to make sure the path to the directory exists before calling os.OpenFile

### DIFF
--- a/pkg/minikube/audit/logFile.go
+++ b/pkg/minikube/audit/logFile.go
@@ -19,7 +19,6 @@ package audit
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 
 	"k8s.io/minikube/pkg/minikube/localpath"
 	"k8s.io/minikube/pkg/minikube/out/register"
@@ -31,9 +30,6 @@ var currentLogFile *os.File
 // setLogFile sets the logPath and creates the log file if it doesn't exist.
 func setLogFile() error {
 	lp := localpath.AuditLog()
-	if err := os.MkdirAll(filepath.Dir(lp), 0755); err != nil {
-		return fmt.Errorf("unable to create directory: %w", err)
-	}
 	f, err := os.OpenFile(lp, os.O_APPEND|os.O_CREATE|os.O_RDWR, 0644)
 	if err != nil {
 		return fmt.Errorf("unable to open %s: %v", lp, err)
--- a/pkg/minikube/audit/logFile.go
+++ b/pkg/minikube/audit/logFile.go
@@ -19,6 +19,7 @@ package audit
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"k8s.io/minikube/pkg/minikube/localpath"
 	"k8s.io/minikube/pkg/minikube/out/register"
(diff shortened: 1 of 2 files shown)


pkg/minikube/audit/logFile.go
```diff
@@ -30,6 +31,9 @@ var currentLogFile *os.File
 // setLogFile sets the logPath and creates the log file if it doesn't exist.
 func setLogFile() error {
 	lp := localpath.AuditLog()
+	if err := os.MkdirAll(filepath.Dir(lp), 0755); err != nil {
```
medyagh, line 34: would this make a Log Dir for "minikube delete --all --purge" ? (in that case we dont wanna create dirs if users wants to purge everything)

medyagh (changes requested):

pkg/minikube/audit/logFile.go
prezha: no, no, it will just create dir for setLogFile when it's called

"minikube delete --all --purge" was mention as to how it can be replicated, eg, `make test` w/o ~/.minikube dir structure will fail when calling setLogFile

medyagh (changes requested):
I am not sure if this PR would help users or would potnetially cause more problems (either adding tiny performance issues on slow disks) since it is gets called on each log run.

or maybe permission issues if 755 is not allowed by the admin. i think this only affects the integration tests. that could be fixed there?

prezha:
> I am not sure if this PR would help users or would potnetially cause more problems (either adding tiny performance issues on slow disks) since it is gets called on each log run.
> 
> or maybe permission issues if 755 is not allowed by the admin. i think this only affects the integration tests. that could be fixed there?

thank you for your review and comments @medyagh 
i think that you are right - atm seems that only the tests are affected by this issue, so i moved the correction there

medyagh (approved):

PR #10807, opened 2021-03-13
Image load: Allow loading local images from tar or cache

### LINKED ISSUE #10744: The new load command still uses the cache directory
Instead of loading to the container runtime directly, it still copies to the cache...

`machine.CacheAndLoadImages`

This is **not** needed, it should be able to transfer the file to the machine directly.

With the current API we have to make a remote disk copy, but that is more minor*...

----

\* Eventually it should be possible to stream the tarball directly into the runtime.
But not every runtime supports this at the moment, so load from a temporary file.

afbjorklund:
Now can load both from daemon/remote and from tarball:

```
minikube image load image
minikube image load image.tar
```

To be on the safe side, one _should_ use qualified names:

```
minikube image load docker.io/image
minikube image load ./image.tar
```

Also allows for loading more than one image at a time...

Closes #10744


medyagh:
@afbjorklund please rebase, so we could include in this release for end of march

pkg/minikube/image/image.go
```diff
@@ -49,6 +49,12 @@ var defaultPlatform = v1.Platform{
 	OS:           "linux",
 }
 
+// UseDaemon is if we should look in local daemon for image ref
+var UseDaemon = true
```
medyagh, line 53: Why public ? Private if not used outside

medyagh:
@afbjorklund this pr does not change the behavior of Image Load to not load to all profiles, right ?

Please clarify,
Currently we have a —all flag for cache add  Is cache being addded to all profiles only when —all is set ?

(It would be great if our integration test would have covered it(

pkg/minikube/image/image.go
afbjorklund: Added public method

afbjorklund:
> this pr does not change the behavior of Image Load to not load to all profiles, right ?

There are no changes to the "cache" commands, only renamed some internal functions.

Added a "Cache" to the functions that wrap the cache dir, and Load does the load directly.

medyagh (approved):

medyagh:
the cache related tests  are failing
https://storage.googleapis.com/minikube-builds/logs/10807/e35057a/Docker_Linux.html#fail_TestFunctional%2fserial%2fCacheCmd%2fcache%2fadd_remote

```

TestFunctional/serial/CacheCmd/cache/add_remote | 1.4
-- | --
52 | TestFunctional/serial/CacheCmd/cache/add_local | 1.85
53 | TestFunctional/serial/CacheCmd/cache/delete_k8s.gcr.io/pause:3.3 | 0.09
55 | TestFunctional/serial/CacheCmd/cache/verify_cache_inside_node | 0.37
56 | TestFunctional/serial/CacheCmd/cache/cache_reload | 1.44
57 | TestFunctional/serial/CacheCmd/cache/delete


```

medyagh (changes requested):
plz check the cache tests failures

https://storage.googleapis.com/minikube-builds/logs/10807/e35057a/Docker_Linux.html#fail_TestFunctional%2fserial%2fCacheCmd%2fcache%2fadd_remote

afbjorklund:
> plz check the cache tests failures

it didn't fall back to remote correctly

afbjorklund:
@medyagh : There is also some bogus message about "cache" being replaced by "image load" ?

`! "minikube cache" will be deprecated in upcoming versions, please switch to "minikube image load"`

Even though image doesn't have any "memory".

So you have to remember to load them again

medyagh:
> @medyagh : There is also some bogus message about "cache" being replaced by "image load" ?
> 
> `! "minikube cache" will be deprecated in upcoming versions, please switch to "minikube image load"`
> 
> Even though image doesn't have any "memory".
> 
> So you have to remember to load them again

yeah saw that, thats not very good, we gotta remove that

medyagh (approved):

PR #10814, opened 2021-03-14
fix potential nil pointer dereference

### LINKED ISSUE #10813: potential nil pointer dereference error
in cmd.validateDriver(), there could be a nil pointer dereference error:

https://github.com/kubernetes/minikube/blob/ff0e25ada888cf1bd7b262857ce96531c1231e69/cmd/minikube/cmd/start.go#L749-L750

/assign

prezha:
fixes #10813

check first if the returned value of reason.MatchKnownIssue is nil before checking its properties

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -746,7 +746,8 @@ func validateDriver(ds registry.DriverState, existing *config.ClusterConfig) {
 		return
 	}
 
-	if r := reason.MatchKnownIssue(reason.Kind{}, st.Error, runtime.GOOS); r != nil && r.ID != "" {
+	r := reason.MatchKnownIssue(reason.Kind{}, st.Error, runtime.GOOS)
+	if r != nil && r.ID != "" {
 		exitIfNotForced(*r, st.Error.Error())
 	}
 
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -746,8 +746,7 @@ func validateDriver(ds registry.DriverState, existing *config.ClusterConfig) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


medyagh (approved):

cmd/minikube/cmd/start.go
```diff
@@ -747,7 +747,7 @@ func validateDriver(ds registry.DriverState, existing *config.ClusterConfig) {
 	}
 
 	r := reason.MatchKnownIssue(reason.Kind{}, st.Error, runtime.GOOS)
-	if r.ID != "" {
+	if r != nil && r.ID != "" {
 		exitIfNotForced(*r, st.Error.Error())
```
medyagh, line 751: how about move the logic inside the exitIfNotForced func, that way it should still Exit but without the Match ?

medyagh (changes requested):

cmd/minikube/cmd/start.go
medyagh: that way minikube will still exit if it is nil

medyagh (changes requested):

cmd/minikube/cmd/start.go
prezha: here the idea was that it should use the 'reason' if it is already set and call exitIfNotForced with it, otherwise 'compile' the reason and then call the exitIfNotForced:

https://github.com/kubernetes/minikube/blob/a05f887651bd65102c6559f3c30439af3e792427/cmd/minikube/cmd/start.go#L749-L782

this pr just makes sure that before checking if reason ID is set, first check if r is not nil

medyagh:
this change been aded already by another PR that we got merged,https://github.com/medyagh/minikube/blob/c48a70f38b6610f3198d196ddffdf252aa45519e/cmd/minikube/cmd/start.go#L750 
 sorry will close this one/

PR #10818, opened 2021-03-14
Add a red box around docker desktop registry port

### LINKED ISSUE #10781: UI: make it hard to miss Important notices
I noticed many ppl miss the notice to use different Port for Mac Os for registry addon

```
$ minikube addons enable registry
💡  Registry addon on with docker uses 55012 please use that instead of default 5000
📘  For more information see: https://minikube.sigs.k8s.io/docs/drivers/docker
    ▪ Using image registry:2.7.1
    ▪ Using image gcr.io/google_containers/kube-registry-proxy:0.4
🔎  Verifying registry addon...
🌟  The 'registry' addon is enabled

```

maybe we can have a new style that creates a Index Pointer Emoji in a Box with one line empty above and bellow so it is hard to miss

```
$ minikube addons enable registry
👉  ------------------------------------------------------------------------------------------
👉   Registry addon on with docker uses 55012 please use that instead of default 5000.   
👉  ------------------------------------------------------------------------------------------
📘  For more information see: https://minikube.sigs.k8s.io/docs/drivers/docker
    ▪ Using image registry:2.7.1
    ▪ Using image gcr.io/google_containers/kube-registry-proxy:0.4
🔎  Verifying registry addon...
🌟  The 'registry' addon is enabled
```

afbjorklund:
To make it harder to miss the notice, with the rest of the emojis

Also remove some stray "on" text, but do add "driver" and "port"

Uses https://github.com/Delta456/box-cli-maker

Closes #10781

----

BEFORE

```
💡  Registry addon on with docker uses 55012 please use that instead of default 5000
```

AFTER

```
    ╭──────────────────────────────────────────────────────────────────────────────────────────╮
💡  │Registry addon on with docker uses port 49170 please use that instead of default port 5000│
    │                                                                                          │
    ╰──────────────────────────────────────────────────────────────────────────────────────────╯
```

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -29,6 +29,8 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 
+	"github.com/Delta456/box-cli-maker/v2"
+
 	"k8s.io/klog/v2"
 	"k8s.io/minikube/pkg/drivers/kic/oci"
 	"k8s.io/minikube/pkg/kapi"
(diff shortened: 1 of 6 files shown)


medyagh (commented):
lets add the same box for the Exit With Advice

the Suggestion is usually not paid attention to,  example https://github.com/kubernetes/minikube/issues/10781#issuecomment-799720054

pkg/addons/addons.go
```diff
@@ -199,7 +201,13 @@ https://github.com/kubernetes/minikube/issues/7332`, out.V{"driver_name": cc.Dri
 				return errors.Wrap(err, "registry port")
 			}
 			if enable {
-				out.Styled(style.Tip, `Registry addon on with {{.driver}} uses {{.port}} please use that instead of default 5000`, out.V{"driver": cc.Driver, "port": port})
+				str := out.Sprintf(style.None, `Registry addon with {{.driver}} driver uses port {{.port}} please use that instead of default port 5000`, out.V{"driver": cc.Driver, "port": port})
```
medyagh, line 204: what do you think about a helper that makes the code look better ?

for example
```
out.BoxNotice(Style, "message to be printed in the box") 
```
so we dont have to repeat this for all other thigns ?

we could use this BoxNotice for important notices that most users usually ignore and they keep asking in the issues (even though we had sent suggestion)

medyagh (changes requested):

pkg/addons/addons.go
afbjorklund: Sure. I'm thinking that zero padding came out a bit wrong, anyway

afbjorklund:
New looks:

box-true
box-false

Obviously I faked the actual contents, to make it show something on Linux...

medyagh (approved):

medyagh:
looks great ! thank you

PR #10742, opened 2021-03-07
Add initial re-implementation of the build command

afbjorklund:
~~Currently only handles tarballs, not directories~~

Where "testbuild" is my build context, with Dockerfile:

./out/minikube --alsologtostderr build testbuild

./out/minikube --alsologtostderr build testbuild.tgz
```
I0307 20:58:38.927939 2574193 build_images.go:105] Building image from path: testbuild.tgz
I0307 20:58:38.928189 2574193 ssh_runner.go:149] Run: sudo mkdir -p /var/lib/minikube/build
I0307 20:58:38.946466 2574193 ssh_runner.go:316] scp testbuild.tgz --> /var/lib/minikube/build/testbuild.tgz (155 bytes)
I0307 20:58:38.964251 2574193 ssh_runner.go:149] Run: sudo mkdir -p /var/lib/minikube/build/testbuild
I0307 20:58:38.971081 2574193 ssh_runner.go:149] Run: sudo tar -C /var/lib/minikube/build/testbuild -xf /var/lib/minikube/build/testbuild.tgz
I0307 20:58:38.978176 2574193 containerd.go:254] Building image: /var/lib/minikube/build/testbuild
I0307 20:58:38.978271 2574193 ssh_runner.go:149] Run: sudo buildctl build --frontend dockerfile.v0 --local context=/var/lib/minikube/build/testbuild --local dockerfile=/var/lib/minikube/build/testbuild --output type=image,name=test:latest
I0307 20:58:53.169144 2574193 ssh_runner.go:189] Completed: sudo buildctl build --frontend dockerfile.v0 --local context=/var/lib/minikube/build/testbuild --local dockerfile=/var/lib/minikube/build/testbuild --output type=image,name=test:latest: (14.190848434s)
I0307 20:58:53.169212 2574193 ssh_runner.go:149] Run: sudo rm -rf /var/lib/minikube/build/testbuild
I0307 20:58:53.177007 2574193 ssh_runner.go:149] Run: sudo rm -f /var/lib/minikube/build/testbuild.tgz
I0307 20:58:53.182428 2574193 build_images.go:152] Built test:latest from testbuild.tgz
I0307 20:58:53.182451 2574193 build_images.go:94] succeeded building to: minikube
I0307 20:58:53.182461 2574193 build_images.go:95] failed building to: 
```

Still no systemd or packages for buildkit, so needs to be started manually.

`sudo -b buildkitd --oci-worker=false --containerd-worker=true --containerd-worker-namespace=k8s.io`

PR #10819, opened 2021-03-14
Add basic documentation for minikube image command

afbjorklund:
Using the image command is an alternative to using the
different environment commands for the different clients.

/hold

Requires these PR first:

* #10807 minikube image load

* #10742 minikube image build

The pushing/building page in the handbook is a bit messy...


medyagh (approved):

afbjorklund:
@medyagh : seems a bit weird to have documentation for commands that don't exist yet...

But I guess we can build upon this placeholder, and provide some _actual_ documentation

PR #10839, opened 2021-03-15
Show last start and audit logs on `minikube logs` if minikube not running

### LINKED ISSUE #10729: Make minikube logs command output audit and last start logs when minikube not running
Running `minikube logs` when minkube is not running

Current:
```
% minikube logs        
🤷  Profile "minikube" not found. Run "minikube profile list" to view all profiles.
```

Expected:
```
==> Audit <==
...

==> Last Start <==
...
```

spowelljr:
Closes #10729

The first thing the `minkube logs` command does is tries loading the running cluster, and if the cluster config does not exist (not running), exits within the load function with the message:
```
🤷  Profile "minikube" not found. Run "minikube profile list" to view all profiles.
```

But if a user is having a problem starting minikube and wants to create a GitHub issue, this prevents them from getting the logs they need to provide in the issue.

I moved the last start and audit logs output before the cluster load to ensure they're output regardless of the status of the cluster.

Running `minikube logs` on non-running cluster:
Before:
```
🤷  Profile "minikube" not found. Run "minikube profile list" to view all profiles.
```

Now:
```
==> Audit <==
...

==> Last Start <==
...

🤷  Profile "minikube" not found. Run "minikube profile list" to view all profiles.
```

### DIFF
--- a/cmd/minikube/cmd/logs.go
+++ b/cmd/minikube/cmd/logs.go
@@ -23,6 +23,7 @@ import (
 	"github.com/spf13/viper"
 	cmdcfg "k8s.io/minikube/cmd/minikube/cmd/config"
 	"k8s.io/minikube/pkg/minikube/cluster"
+	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/cruntime"
 	"k8s.io/minikube/pkg/minikube/exit"
 	"k8s.io/minikube/pkg/minikube/logs"
@@ -52,6 +53,11 @@ var logsCmd = &cobra.Command{
 	Short: "Returns logs to debug a local Kubernetes cluster",
 	Long:  `Gets the logs of the running instance, used for debugging minikube, not user code.`,
 	Run: func(cmd *cobra.Command, args []string) {
+		if !isClusterRunning() {
+			logs.OutputMinikubeLogs(numberOfLines)
+			return
+		}
+
 		co := mustload.Running(ClusterFlagValue())
 
 		bs, err := cluster.Bootstrapper(co.API, viper.GetString(cmdcfg.Bootstrapper), *co.Config, co.CP.Runner)
@@ -83,6 +89,16 @@ var logsCmd = &cobra.Command{
 			os.Exit(reason.ExSvcError)
 		}
 	},
+}
+
+func isClusterRunning() bool {
+	if _, err := config.Load(ClusterFlagValue()); err != nil {
+		if config.IsNotExist(err) {
+			return false
+		}
+		exit.Error(reason.HostConfigLoad, "Error getting cluster config", err)
+	}
+	return true
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


pkg/minikube/logs/logs.go
```diff
@@ -243,6 +243,16 @@ func outputLastStart() error {
 	return nil
 }
 
+// OutputMinikubeLogs outputs the audit and last start logs only
+func OutputMinikubeLogs(lines int) {
```
medyagh, line 247: `logs.OutputMinikubeLogs` stutters and doesnt read well

how about, Offline or DisplayOffline() ?
```
// Offline outputs logs that doesn't need a running cluster.
func logs.Offline()


```

cmd/minikube/cmd/logs.go
```diff
@@ -52,6 +53,11 @@ var logsCmd = &cobra.Command{
 	Short: "Returns logs to debug a local Kubernetes cluster",
 	Long:  `Gets the logs of the running instance, used for debugging minikube, not user code.`,
 	Run: func(cmd *cobra.Command, args []string) {
+		if !isClusterRunning() {
```
medyagh, line 56: why bother checking if it is NOT running at all ?
how about we just output the audit and last start if it exists at all the time ? (without checking if there is a minikube config or not)

medyagh (changes requested):

cmd/minikube/cmd/logs.go
spowelljr: We could do this, the only thing is the profile not found message will still be displayed, so it will look like the following:
```
==> Audit <==
...

==> Last Start <==
...
🤷  Profile "minikube" not found. Run "minikube profile list" to view all profiles.
```
If that's something we're ok with then we could definitely output the logs without the check, the check only exists to make sure that message is not output.

pkg/minikube/logs/logs.go
spowelljr: That makes sense, I think having `Output` in the func name is important so there's no question whether the func returns or outputs the logs, and it also follows in convention with other funcs. I'll update with `logs.OutputOffline()`.

cmd/minikube/cmd/logs.go
medyagh: yeah I think that is acceptable

how about we add a Block header for it ?
==> Audit <==
...

==> Last Start <==
...
==> Internal Logs <==

spowelljr: That makes sense if the cluster if not running, the only thing is if the cluster is running, there will be a double block header for the first internal log.

Example:
```
==> Audit <==
...

==> Last Start <==
...
==> Internal Logs <==
==> Docker <==
```

What's your thoughts on that?

pkg/minikube/logs/logs.go
medyagh: `OutputOffline` sounds good

cmd/minikube/cmd/logs.go
medyagh: good question ! I think u are right, lets not have the "internal logs" header at all.

medyagh (approved):

Issue #9907, opened 2020-12-10
Provide more descriptive error codes for GUEST_PROVISIONG


go-nelango:
GUEST_PROVISION comprises 8.5% of minikube start errors in our environment.

These could be because minikube failed to obtain driver options/ generate cluster config/ cache minikube ISO. 

Can minikube provide more descriptive error codes for their scenarios? It would also help to make the error messages actionable wherever possible.

tstromberg:
As a catchall error code, In order to break this code down further, we'll need to collect example error messages for known issues. We can then add support for them here:

https://github.com/kubernetes/minikube/blob/master/pkg/minikube/reason/known_issues.go

tstromberg:
Here's the best advice I have for collecting this info:

https://github.com/kubernetes/minikube/issues?q=is%3Aissue+GUEST_PROVISION

Looks like we have 21 examples currently that can get broken down to more specific exit codes.

tstromberg:
I noticed at least one bug here in #10151 - it should have raised PR_DOCKER_NO_SSH instead of GUEST_PROVISION. This makes me wonder if there is a bug that is causing known_issues to not be parsed.

medyagh:
@azhao155 has too many tasks this milestone, I unassign it for now if anyone wants to pick it up

spowelljr:
In-progress

Issue #9190, opened 2020-09-05
minikube start with kvm-qemu-uri still fails due to non-existent "default" network


ge79so:
**Steps to reproduce the issue:** 

_On Debian stable 10.5_
_NB. KVM is installed and working with virt-manager_

`$ minikube start --kvm-qemu-uri qemu:///session`

**Full output of failed command:** 

🔄  Restarting existing kvm2 VM for "minikube" ...
😿  Failed to start kvm2 VM. Running "minikube delete" may fix it: driver start: creating network: network default doesn't exist: virError(Code=43, Domain=19, Message='Network not found: no network with matching name 'default'')

❌  Exiting due to GUEST_PROVISION: Failed to start host: driver start: creating network: network default doesn't exist: virError(Code=43, Domain=19, Message='Network not found: no network with matching name 'default'')

😿  If the above advice does not help, please let us know: 
👉  https://github.com/kubernetes/minikube/issues/new/choose

`$ user@pc:/usr/local/bin$ minikube delete`

🔥  Removing /home/user/.minikube/machines/minikube ...
💀  Removed all traces of the "minikube" cluster.

user@pc:/usr/local/bin$ minikube start --kvm-qemu-uri qemu:///session

😄  minikube v1.13.0 on Debian 10.5
✨  Automatically selected the kvm2 driver
👍  Starting control plane node minikube in cluster minikube
🔥  Creating kvm2 VM (CPUs=2, Memory=2200MB, Disk=20000MB) ...
🤦  StartHost failed, but will try again: creating host: create: Error creating machine: Error in driver during machine creation: creating network: network default doesn't exist: virError(Code=43, Domain=19, Message='Network not found: no network with matching name 'default'')
🔄  Restarting existing kvm2 VM for "minikube" ...
😿  Failed to start kvm2 VM. Running "minikube delete" may fix it: driver start: creating network: network default doesn't exist: virError(Code=43, Domain=19, Message='Network not found: no network with matching name 'default'')

❌  Exiting due to GUEST_PROVISION: Failed to start host: driver start: creating network: network default doesn't exist: virError(Code=43, Domain=19, Message='Network not found: no network with matching name 'default'')

😿  If the above advice does not help, please let us know: 
👉  https://github.com/kubernetes/minikube/issues/new/choose

**Full output of `minikube start` command used, if not already included:**

**Optional: Full output of `minikube logs` command:**
<details>
$ minikube logs

[...]

PR #10841, opened 2021-03-15
unique error codes for KVM network and docker ip conflict

medyagh:
helps with https://github.com/kubernetes/minikube/pull/10806
and https://github.com/kubernetes/minikube/issues/9907 and  https://github.com/kubernetes/minikube/issues/9190


ilya-zuyev (approved):
lgtm

PR #10844, opened 2021-03-16
Allow running kubectl over the ssh connection

### LINKED ISSUE #10843: Add support for running kubectl remotely as an alternative
Some clients don't have the possibility to run the kubernetes client locally, due to binaries not being available...

But we already install `kubectl` on the control plane in the cluster, so we can run that using the `ssh` connection.

```
$ minikube kubectl --ssh=false -- get nodes
NAME       STATUS   ROLES                  AGE   VERSION
minikube   Ready    control-plane,master   12h   v1.20.2
```
```
$ minikube kubectl --ssh=true -- get nodes
NAME       STATUS   ROLES                  AGE   VERSION
minikube   Ready    control-plane,master   12h   v1.20.2
```

This makes minikube accessible on more platforms.

Especially when using the `ssh` driver for the cluster.

afbjorklund:
Closes #10843

### DIFF
--- a/cmd/minikube/cmd/kubectl.go
+++ b/cmd/minikube/cmd/kubectl.go
@@ -66,7 +66,7 @@ minikube kubectl -- get pods --namespace kube-system`,
 				waitStatus := exitError.Sys().(syscall.WaitStatus)
 				rc = waitStatus.ExitStatus()
 			} else {
-				fmt.Fprintf(os.Stderr, "Error running %s: %v\n", path, err)
+				fmt.Fprintf(os.Stderr, "Error running %s: %v\n", c.Path, err)
 				rc = 1
 			}
 			os.Exit(rc)
--- a/cmd/minikube/cmd/kubectl.go
+++ b/cmd/minikube/cmd/kubectl.go
@@ -44,9 +44,13 @@ var kubectlCmd = &cobra.Command{
 	Short: "Run a kubectl binary matching the cluster version",
 	Long: `Run the Kubernetes client, download it if necessary. Remember -- after kubectl!
 
-Examples:
-minikube kubectl -- --help
-minikube kubectl -- get pods --namespace kube-system`,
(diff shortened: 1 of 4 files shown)


cmd/minikube/cmd/kubectl.go
```diff
@@ -87,3 +126,7 @@ func KubectlCommand(version string, args ...string) (*exec.Cmd, error) {
 
 	return exec.Command(path, args...), nil
 }
+
+func init() {
+	kubectlCmd.Flags().BoolVar(&useSSH, "ssh", false, "Use SSH for running kubernetes client")
```
medyagh, line 131: does this mean, it will be using the kubectl inside minikube ? (is there a reason that we dont do that all the time ?)

medyagh: lets clarify the help text


how about add a section either in FAQ https://minikube.sigs.k8s.io/docs/faq/ or the SSH driver page https://minikube.sigs.k8s.io/docs/drivers/ssh/
how to use this.

maybe with a way that when users google it, they find it, "How to use kubectl with SSH driver on minikube ? or how to talk to minikube cluster without kubectl ? "

medyagh (changes requested):

cmd/minikube/cmd/kubectl.go
afbjorklund: There are lots of reasons why we don't have this as the default, one reason is working with files...
```
  # Apply the configuration in pod.json to a pod.
  kubectl apply -f ./pod.json
```

You cannot do "kubectl apply" unless you use a path that are on both. You need to use `-` and `<`, etc.

 ```
 # Apply the JSON passed into stdin to a pod.
  cat pod.json | kubectl apply -f -
```

afbjorklund: Added more help text, also noticed that "minikube kubectl -- --help" had also been broken by --cluster

medyagh (approved):

PR #10865, opened 2021-03-18
some words are not correct then change it

lilongfeng0902:
find some wrong words, then correct it.

### DIFF
--- a/enhancements/proposed/20201030-tracing/README.md
+++ b/enhancements/proposed/20201030-tracing/README.md
@@ -37,7 +37,7 @@ There are two pieces to the design: collecting the data and exporting the data.
 
 ### Collecting Data
 Luckily, we already have a lot of this infrastructure set up for JSON output.
-We know when a new substep of `minikube start` has started, because we call it explictly via `register.SetStep`.
+We know when a new substep of `minikube start` has started, because we call it explicitly via `register.SetStep`.
 We also know that substep has ended when a new substep begins.
 
 We can start new spans whenever `register.SetStep` is called, and thus collect tracing data.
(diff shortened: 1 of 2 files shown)


camba1:
The build is failing (undefined: retryablehttp) because "github.com/hashicorp/go-retryablehttp" was removed from [functional_test.go](https://github.com/kubernetes/minikube/blob/master/test/integration/functional_test.go)  but it is used in function validateDashboardCmd. 

I suggest the dependency is re-added and the PR re-submitted as the rest of the change is good

lilongfeng0902:
> The build is failing (undefined: retryablehttp) because "github.com/hashicorp/go-retryablehttp" was removed from [functional_test.go](https://github.com/kubernetes/minikube/blob/master/test/integration/functional_test.go) but it is used in function validateDashboardCmd.
> 
> I suggest the dependency is re-added and the PR re-submitted as the rest of the change is good

Thanks for your help，this is my first PR, some situations may be difficult for me . By your notion, i have change the file functional_test.go .

camba1:
Thanks for the change @lilongfeng0902 
Unfortunately, it looks like the new commit only changed the comment on line 87 but did not re-add the missing dependency ("github.com/hashicorp/go-retryablehttp"). As such, the build failed again on the same spot and the same error. Please try to re-add the dependency. See line 49 of the [existing code](https://github.com/kubernetes/minikube/blob/master/test/integration/functional_test.go) .
Thanks again!

lilongfeng0902:
> Thanks for the change @lilongfeng0902
> Unfortunately, it looks like the new commit only changed the comment on line 87 but did not re-add the missing dependency ("github.com/hashicorp/go-retryablehttp"). As such, the build failed again on the same spot and the same error. Please try to re-add the dependency. See line 49 of the [existing code](https://github.com/kubernetes/minikube/blob/master/test/integration/functional_test.go) .
> Thanks again!

Yeah，I find my misti> Thanks for the change @lilongfeng0902
> Unfortunately, it looks like the new commit only changed the comment on line 87 but did not re-add the missing dependency ("github.com/hashicorp/go-retryablehttp"). As such, the build failed again on the same spot and the same error. Please try to re-add the dependency. See line 49 of the [existing code](https://github.com/kubernetes/minikube/blob/master/test/integration/functional_test.go) .
> Thanks again!

Yeah，I find my mistake and I re-add  the dependency. Please check my code again, thanks.

Issue #10852, opened 2021-03-17
--containerd-runtime=containerd: ctr: content digest sha256:x: not found (reproducible)


tstromberg:
On macOS (arm64), running `minikube start --driver=docker --container-runtime=containerd` with v1.18.1 results in load issues across multiple runs. In this latest run, it is with the storage-provisioner, in the other it was with coredns.

Altogether, it feels like a race condition of some sort.

```
📦  Preparing Kubernetes v1.20.2 on containerd 1.4.3 ...
❌  Unable to load cached images: loading cached images: containerd load /var/lib/minikube/images/storage-provisioner_v4: ctr images import: sudo ctr -n=k8s.io images import /var/lib/minikube/images/storage-provisioner_v4: Process exited with status 1
stdout:
unpacking gcr.io/k8s-minikube/storage-provisioner:v4 (sha256:aac8e78f2bcc7fc7a1f5f9d8a84c2b8e6915da43e3a03a7bc5d25036b9452ebb)...
stderr:
ctr: content digest sha256:1c7d4f85af5d300a03d4b29ff618046447e136412853db455955545953c7ac78: not found
```

[containerd-1-startlog.txt](https://github.com/kubernetes/minikube/files/6153110/containerd-1-startlog.txt)

tstromberg:
Second log:

[containerd-2-startlog.txt](https://github.com/kubernetes/minikube/files/6153283/containerd-2-startlog.txt)

tstromberg:
At first I thought this might be a race condition relating to preload & running purge before-hand, but I'm also seeing similar issues simply running `minikube start --driver=docker --container-runtime=containerd                                                                                                                                                                                `:

```
😄  minikube v1.18.1 on Darwin 11.2.2 (arm64)
✨  Using the docker driver based on user configuration
👍  Starting control plane node minikube in cluster minikube
🚜  Pulling base image ...
🔥  Creating docker container (CPUs=2, Memory=1981MB) ...
📦  Preparing Kubernetes v1.20.2 on containerd 1.4.3 ...
❌  Unable to load cached images: loading cached images: containerd load /var/lib/minikube/images/kube-proxy_v1.20.2: ctr images import: sudo ctr -n=k8s.io images import /var/lib/minikube/images/kube-proxy_v1.20.2: Process exited with status 1
stdout:
[...]

PR #10868, opened 2021-03-18
bridge cni: Make sure to create the directory for cni config

afbjorklund:
Otherwise it will be created with the mktemp permissions
and become unreadable (0700), making chmod fail later on.

Fixes error seen in https://github.com/kubernetes/minikube/issues/10852#issuecomment-801849771

`💣  Error starting cluster: apply cni: cni apply: copy: chmod /etc/cni/net.d/1-k8s.conf: permission denied`

### DIFF
--- a/pkg/minikube/cni/bridge.go
+++ b/pkg/minikube/cni/bridge.go
@@ -19,6 +19,7 @@ package cni
 import (
 	"bytes"
 	"fmt"
+	"os/exec"
 	"text/template"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


pkg/minikube/cni/bridge.go
```diff
@@ -84,6 +85,10 @@ func (c Bridge) Apply(r Runner) error {
 		return fmt.Errorf("bridge CNI is incompatible with multi-node clusters")
 	}
 
+	if _, err := r.RunCmd(exec.Command("sudo", "mkdir", "-p", "/etc/cni/net.d")); err != nil {
+		return errors.Wrap(err, "netconf")
```
ilya-zuyev, line 89: some error message about `sudo mkdir`, not `netconf` here, may be

afbjorklund: Just returning the error instead of wrapping would probably work too

medyagh (approved):

medyagh:
Weirdly both docker Linux and kvm tests timed out...I am gonna retest to see what is going on.

PR #10872, opened 2021-03-19
add docker-env and podman-env to minikube status

### LINKED ISSUE #6535: feature: minikube status- indicate if using "docker-env"
one of the common things I do after eval $(minikube docker-env) is checking if I am using my own docker or minikube's docker

it would be nice if "minikube status" command would tell me 

```
host: ...
kubelet: ..
apiserver: ...
kubeconfig: ...
docker-env: in-use
```
and if we are not using docker-env no need to add that line and keep it hidden.
but we could includ that in the json output.

tharun208:
Signed-off-by: Tharun <rajendrantharun@live.com>

Before: `minikube status`
Screen Shot 2021-03-23 at 2 36 18 PM

After: `minikube docker-env` and `minikube-status` and `minikube docker-env --unset`
Screen Shot 2021-03-23 at 2 37 17 PM

`minikube status --output=json` result:
Screen Shot 2021-03-23 at 3 36 54 PM

Fixes #6535

### DIFF
--- a/cmd/minikube/cmd/docker-env.go
+++ b/cmd/minikube/cmd/docker-env.go
@@ -252,8 +252,7 @@ var dockerEnvCmd = &cobra.Command{
 		sh := shell.EnvConfig{
 			Shell: shl,
 		}
-		cname := ClusterFlagValue()
-		co := mustload.Running(cname)
+
 		if dockerUnset {
 			if err := dockerUnsetScript(DockerEnvConfig{EnvConfig: sh}, os.Stdout); err != nil {
 				exit.Error(reason.InternalEnvScript, "Error generating unset output", err)
@@ -265,6 +264,9 @@ var dockerEnvCmd = &cobra.Command{
 			out.SetSilent(true)
 			exit.SetShell(true)
 		}
+
+		cname := ClusterFlagValue()
+		co := mustload.Running(cname)
 
 		driverName := co.CP.Host.DriverName
 
--- a/cmd/minikube/cmd/docker-env.go
+++ b/cmd/minikube/cmd/docker-env.go
@@ -252,7 +252,8 @@ var dockerEnvCmd = &cobra.Command{
 		sh := shell.EnvConfig{
 			Shell: shl,
 		}
-
+		cname := ClusterFlagValue()
+		co := mustload.Running(cname)
 		if dockerUnset {
 			if err := dockerUnsetScript(DockerEnvConfig{EnvConfig: sh}, os.Stdout); err != nil {
 				exit.Error(reason.InternalEnvScript, "Error generating unset output", err)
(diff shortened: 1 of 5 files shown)


medyagh (changes requested):
lets add same thing for podman-env

medyagh:
Do you mind pasting the out of minikube status --output=json
And can we ensure the json output has it too?

pkg/minikube/config/types.go
```diff
@@ -81,6 +81,7 @@ type ClusterConfig struct {
 	ListenAddress           string   // Only used by the docker and podman driver
 	Network                 string   // only used by docker driver
 	MultiNodeRequested      bool
+	DockerEnvInUse          bool
```
medyagh, line 84: This should not be part of config. Because in one terminal it might be in use and other terminal not but config is global. So let's remove it

medyagh (changes requested):

pkg/minikube/config/types.go
tharun208: @medyagh, without the config change. can you help me with how we can implement this for podman?

for docker, we can make use of this `MINIKUBE_ACTIVE_DOCKERD` env variable at runtime to check if it is the same as the cluster name.

medyagh: For podman u could see the podman-env.go file

But also lets make sure the minikbue Config does not have that DockerEnvInUse ...

tharun208:
@medyagh, I updated the code without making any changes in local. But, I am not able to run the `functional test` suite successfully locally.

cmd/minikube/cmd/docker-env.go
```diff
@@ -252,7 +252,8 @@ var dockerEnvCmd = &cobra.Command{
 		sh := shell.EnvConfig{
 			Shell: shl,
 		}
```
medyagh, line 254: Any reason this was moved up ? If not revert the unrelated changes

cmd/minikube/cmd/status.go
```diff
@@ -136,6 +137,8 @@ type Status struct {
 	Kubeconfig string
 	Worker     bool
 	TimeToStop string
```
medyagh, line 139: Would u plz paste the output of minikube status —outpus=json after this PR ?
Does this field reflect in the json output ?

medyagh (changes requested):

cmd/minikube/cmd/status.go
tharun208: @medyagh, I added the output for `minikube status —outpus=json` in the pr description

medyagh (approved):

medyagh:
thank you for this contribution @tharun208

PR #10881, opened 2021-03-19
Add documentation for the ctr containerd client

afbjorklund:
As a "replacement" for Docker, this (containerd/buildkitd) is quite complicated.

For now we don't bother with "nerdctl", but use "ctr" and "buildctl" directly.

The main complication is that the clients **don't** work remotely, with `ssh:`.

And the usual root authentication issues, similar to docker and podman.


medyagh (approved):

medyagh:
@afbjorklund rebase plz then good to merge

medyagh:
Thank youy

PR #10886, opened 2021-03-21
Added error check for extra quotation for extra-config.

### LINKED ISSUE #8694: --extra-config with double quotes breaks kubeadm init
we should validate options like this
```
./out/minikube start --driver=docker --extra-config=etcd.client-cert-auth=false”
```

currently running that will make minikube kubeadm hang.

vishjain:
fixes #8694 

Here is the output with the PR changes: 
vishals-MacBook-Pro:minikube vishal$ ./out/minikube start --driver=docker --extra-config=etcd.client-cert-auth=false”
Error: invalid argument "etcd.client-cert-auth=false”" for "--extra-config" flag: invalid value: cannot contain end quotation: "etcd.client-cert-auth=false”"
See 'minikube start --help' for usage.

Before, the ./out/minikube start command would just hang. This PR handles the case where end of --extra-config value has an open/close quotation because that's where the hanging behavior occurs.

### DIFF
--- a/pkg/minikube/config/extra_options.go
+++ b/pkg/minikube/config/extra_options.go
@@ -66,14 +66,11 @@ func (es *ExtraOptionSlice) Exists(value string) bool {
 // Set parses the string value into a slice
 func (es *ExtraOptionSlice) Set(value string) error {
 
-	// Check doesn't start or end with an opening quotation.
-	if strings.HasPrefix(value, "“") || strings.HasSuffix(value, "“") {
-		return fmt.Errorf("invalid value: canot contain quotation: %q", value)
-	}
-
-	// Check doesn't start or end with a closing quotation.
-	if strings.HasPrefix(value, "”") || strings.HasSuffix(value, "”") {
-		return fmt.Errorf("invalid value: canot contain quotation: %q", value)
+	// Check we don't end with suffix quotation.
+	prefixExists := strings.HasPrefix(value, "”") || strings.HasPrefix(value, "”")
+	suffixExists := strings.HasSuffix(value, "”") || strings.HasSuffix(value, "”")
+	if !prefixExists && suffixExists {
+		return fmt.Errorf("invalid value: cannot contain end quotation: %q", value)
 	}
 
 	// The component is the value before the first dot.
--- a/pkg/minikube/config/extra_options.go
+++ b/pkg/minikube/config/extra_options.go
@@ -65,7 +65,13 @@ func (es *ExtraOptionSlice) Exists(value string) bool {
 
 // Set parses the string value into a slice
 func (es *ExtraOptionSlice) Set(value string) error {
-	// Check doesn't start or end with a quotation.
+
+	// Check doesn't start or end with an opening quotation.
+	if strings.HasPrefix(value, "“") || strings.HasSuffix(value, "“") {
+		return fmt.Errorf("invalid value: canot contain quotation: %q", value)
+	}
+
+	// Check doesn't start or end with a closing quotation.
 	if strings.HasPrefix(value, "”") || strings.HasSuffix(value, "”") {
 		return fmt.Errorf("invalid value: canot contain quotation: %q", value)
 	}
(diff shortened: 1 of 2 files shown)


pkg/minikube/config/extra_options.go
```diff
@@ -65,6 +65,17 @@ func (es *ExtraOptionSlice) Exists(value string) bool {
 
 // Set parses the string value into a slice
 func (es *ExtraOptionSlice) Set(value string) error {
+
+	// Check doesn't start or end with an opening quotation.
+	if strings.HasPrefix(value, "“") || strings.HasSuffix(value, "“") {
+		return fmt.Errorf("invalid value: canot contain quotation: %q", value)
```
medyagh, line 71: Can you verify if this is actually true ? Could it be it hangs only if the there is One Qutation without closing ?


Do u mind posting the output of minikube fore and after this PR?

medyagh (changes requested):

pkg/minikube/config/extra_options.go
vishjain: My mistake: There is only hanging currently if at the end there is a quotation mark: 
minikube vishal$ ./out/minikube start --driver=docker --extra-config=etcd.client-cert-auth=false“
😄  minikube v1.18.1 on Darwin 10.14.6
🐳  Preparing Kubernetes v1.20.2 on Docker 20.10.3 ...
    ▪ etcd.client-cert-auth=false“

vishjain: There is already an error that shows up when there is a quote at the beginning too: 
./out/minikube start --driver=docker --extra-config=“etcd.client-cert-auth=false”
😄  minikube v1.18.1 on Darwin 10.14.6
❗  These --extra-config parameters are invalid: [“etcd]

❌  Exiting due to MK_USAGE: Valid components are: [apiserver controller-manager scheduler etcd kubeadm kubelet kube-proxy]

So we need to just check the end of the string for a quote.

vishjain: ./out/minikube start --driver=docker --extra-config=“etcd.client-cert-auth=false
😄  minikube v1.18.1 on Darwin 10.14.6
❗  These --extra-config parameters are invalid: [“etcd]

❌  Exiting due to MK_USAGE: Valid components are: [apiserver controller-manager scheduler etcd kubeadm kubelet kube-proxy]

vishjain: I have preserved the initial errors and also added the error when there's a suffix quote at the end: 
 ./out/minikube start --driver=docker --extra-config=etcd.client-cert-auth=false”
Error: invalid argument "etcd.client-cert-auth=false”" for "--extra-config" flag: invalid value: cannot contain end quotation: "etcd.client-cert-auth=false”"

pkg/minikube/config/extra_options.go
```diff
@@ -65,6 +65,14 @@ func (es *ExtraOptionSlice) Exists(value string) bool {
 
 // Set parses the string value into a slice
 func (es *ExtraOptionSlice) Set(value string) error {
+
+	// Check we don't end with suffix quotation.
+	prefixExists := strings.HasPrefix(value, "”") || strings.HasPrefix(value, "“")
+	suffixExists := strings.HasSuffix(value, "”") || strings.HasSuffix(value, "“")
+	if !prefixExists && suffixExists {
+		return fmt.Errorf("invalid value: cannot contain end quotation: %q", value)
```
medyagh, line 73: what happens when error happens ? could u paste in the PR description

Before and After this PR ? 
how about adding Unit test for `Set` Function ? with example of things that should Error and example of things that should NOT error

medyagh: also improve the error wording, ```-extra-config```

medyagh (changes requested):

pkg/minikube/config/extra_options.go
vishjain: I've added a test and in the PR description above I've included what happens when we do the error checking. We are stopping he hanging occurring when a extra-config value ends with a open or close quote. It's important to note if the extra-config value has a prefix with an open or close quote then there's a separate error check in start.go file (cli) - I have left that alone.

medyagh (approved):

medyagh (changes requested):
@vishjain please fix the lint, run "make test"

vishjain:
@medyagh I ran make test and make lint. Hope it's good now.

medyagh (approved):

PR #10888, opened 2021-03-21
site: windows instructions, reload powershell after install

StingyJack:
if you dont the `minikube start` will fail because the command isnt in the PATH for the current shell

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -125,6 +125,9 @@ choco install minikube
 ### Stand-alone Windows Installer
 Otherwise, download and run the [Windows installer](https://storage.googleapis.com/minikube/releases/latest/minikube-installer.exe)
 
+
+_If you used a CLI to perform the installation, you will need to close that CLI and open a new one before proceeding._ 
+
 {{% /windowstab %}}
 {{% /tabs %}}
 


StingyJack:
cla signed

site/content/en/docs/start/_index.md
```diff
@@ -125,6 +125,9 @@ choco install minikube
 ### Stand-alone Windows Installer
 Otherwise, download and run the [Windows installer](https://storage.googleapis.com/minikube/releases/latest/minikube-installer.exe)
 
+
+_If you used a CLI to perform the installation, you will need to close that CLI and open a new one before proceeding._
```
medyagh, line 129: how about something shorter,

you might need to reload the powershell after installation through "..x... "

medyagh (changes requested):

site/content/en/docs/start/_index.md
StingyJack: How do you "reload a powershell"?

tstromberg: I believe medyagh was referring to the path. It is possible to refresh the path in Powershell, but they may be a bit advanced: https://stackoverflow.com/questions/17794507/reload-the-path-in-powershell

The problem isn't installing from inside the CLI, but that the CLI does not instantly reflect changes to the global Machine/User path. If you used the GUI installer, you would still have the same issue.

Can I suggest something like:

* In the Chocalatey instructions, ask users to run `refreshenv`
* In the winget instructions, ask users to add it to path using `$env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path","User") `
* In the standalone installer instructions, ask users to start a brand new Administrative console window for the new command to appear in path. 

`If this command returns "Not found", please close and re-open Powershell so that it is reflected in your path".

StingyJack: @tstromberg - I couldnt tell if that was a comment by someone who knew how to do that hard thing (harder than closing and opening a new shell anyway) in powershell, or someone that was (reasonably) assuming that powershell had something like bash `source`. Its always better to ask for clarification 

 I think what I said originally covers this and in only a few words that dont require wizardry or tricks to accomplish.

StingyJack: Also windows has two CLI's and the same would need to be done in each of them. My additional sentence covers both.   Is there still an objection to the sentence being too long?

sharifelgamal (approved):
This looks good to me. I'm going to go ahead and merge this.

PR #10895, opened 2021-03-22
go download url has changed

MaXinjian:
`$ curl https://golang.org/dl/go1.16.linux-amd64.tar.gz`
  `<a href="https://dl.google.com/go/go1.16.linux-amd64.tar.gz">Found</a>.`

Signed-off-by: Ma Xinjian <maxj.fnst@fujitsu.com>


afbjorklund (changes requested):
No, the URL is the same as always: https://golang.org/dl/

Currently hosted by google.com, but that could change.
The `-L` flag will make sure to follow the https redirects.

Issue #10578, opened 2021-02-23
ssh driver --ssh-ip-address seems to be ignored


horihel:
**Steps to reproduce the issue:** 

1. prepare VM listening on 10.0.0.10, allowing root login with ssh key
2. `minikube start --driver=ssh --ssh-ip-address=10.0.0.10 --ssh-key='id_rsa' --alsologtostderr`

**Full output of failed command:** 
```
I0223 09:18:23.815988   11444 out.go:229] Setting OutFile to fd 76 ...
I0223 09:18:23.816966   11444 out.go:281] isatty.IsTerminal(76) = true
I0223 09:18:23.818920   11444 out.go:242] Setting ErrFile to fd 76...
I0223 09:18:23.824778   11444 out.go:281] isatty.IsTerminal(76) = true
I0223 09:18:23.843751   11444 out.go:236] Setting JSON to false
I0223 09:18:23.856498   11444 start.go:106] hostinfo: {"hostname":"hellap-18","uptime":6142,"bootTime":1614062161,"procs":313,"os":"windows","platform":"Microsoft Windows 10 Pro","platformFamily":"Standalone Workstation","platformVersion":"10.0.19042 Build 19042","kernelVersion":"10.0.19042 Build 19042","kernelArch":"x86_64","virtualizationSystem":"","virtualizationRole":"","hostId":"76c9fbb9-3105-4327-84cc-0aa28cf6cf2b"}
W0223 09:18:23.857939   11444 start.go:114] gopshost.Virtualization returned error: not implemented yet
I0223 09:18:23.863446   11444 out.go:119] 😄  minikube v1.17.1 auf Microsoft Windows 10 Pro 10.0.19042 Build 19042
😄  minikube v1.17.1 auf Microsoft Windows 10 Pro 10.0.19042 Build 19042
I0223 09:18:23.869254   11444 notify.go:126] Checking for updates...
W0223 09:18:23.869811   11444 start.go:631] api.Load failed for minikube: filestore "minikube": Docker machine "minikube" does not exist. Use "docker-machine ls" to list machines. Use "docker-machine create" to add a new one.
I0223 09:18:23.871142   11444 driver.go:315] Setting default libvirt URI to qemu:///system
W0223 09:18:23.871673   11444 start.go:631] api.Load failed for minikube: filestore "minikube": Docker machine "minikube" does not exist. Use "docker-machine ls" to list machines. Use "docker-machine create" to add a new one.
[...]

Issue #10759, opened 2021-03-09
Add overwrite flag into start command to change values in config files.


daehyeok:
Currently it looks we can overwrite flags  in config file only when first start fail.

https://github.com/kubernetes/minikube/blob/1fe7615e9e9489e5abb595da89e156b5d0a662bc/cmd/minikube/cmd/start_flags.go#L433-L435

How about add `--overwrite` flag to intentionally update value?
Then we can solve some problem occurred by machine or nodes internal change(#10578)

sharifelgamal:
Seems like a reasonable idea. Help wanted!

daehyeok:
Actually, it should be updated always.

https://github.com/kubernetes/minikube/blob/d2cb052fca726f023c8c57b0377bc716061973cc/cmd/minikube/cmd/start_flags.go#L247-L252

but `updateExistingConfigFromFlags` func have multiple missing flags and check routine. Let's see we can fix&refactoring it.

PR #10899, opened 2021-03-22
Refactoring generate/update config file code.

daehyeok:
Clean up `generateClusterConfig` and `updateExistingConfigFromFlags` functions.
FIxs #10759
Github diff looks awful :(.

Here is some summary for this PR.

-  Move upgrade filed code to `upgradeExistingConfig` from `updateExistingConfigFromFlags`
   - Currently, Empty `cc.KubernetesConfig.NodePort`  is validated in `updateExistingConfigFromFlags` function. I think it should be checked in `upgradeExistingConfig` function.
https://github.com/kubernetes/minikube/blob/185b4a70502aa2cb42bc2d208ec2981e0d4804d0/cmd/minikube/cmd/start_flags.go#L628-L634 
-  Refactoring `generateClusterConfig`
   - Split validate/set code with helper function to reuse.
- Refactoring  `updateExistingConfigFromFlags`
  - Add missing flags  to overwrite,(E.g, SSHAddress)
  - Fix wrong mapped cmd.Flags().Changed https://github.com/kubernetes/minikube/blob/185b4a70502aa2cb42bc2d208ec2981e0d4804d0/cmd/minikube/cmd/start_flags.go#L624-L626
  - Reuse validate/setting helper function
Some Validation routine slightly different from `generateClusterConfig`
E.g.) memory validate code in  `generateClusterConfig` https://github.com/kubernetes/minikube/blob/185b4a70502aa2cb42bc2d208ec2981e0d4804d0/cmd/minikube/cmd/start_flags.go#L260-L272
  but `updateExistingConfigFromFlags` func don 't check containerLimit 
https://github.com/kubernetes/minikube/blob/185b4a70502aa2cb42bc2d208ec2981e0d4804d0/cmd/minikube/cmd/start_flags.go#L490-L498

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -177,7 +177,7 @@ func runStart(cmd *cobra.Command, args []string) {
 	}
 
 	if existing != nil {
-		upgradeExistingConfig(existing)
+		upgradeExistingConfig(cmd, existing)
 	} else {
 		validateProfileName()
 	}
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -290,10 +290,7 @@ func generateClusterConfig(cmd *cobra.Command, existing *config.ClusterConfig, k
 	return createNode(cc, kubeNodeName, existing)
 }
 
-// generateNewConfigFromFlags generate a config.ClusterConfig based on flags
-func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, drvName string) config.ClusterConfig {
-	var cc config.ClusterConfig
-
+func getMemorySize(cmd *cobra.Command, drvName string) int {
 	sysLimit, containerLimit, err := memoryLimits(drvName)
 	if err != nil {
 		klog.Warningf("Unable to query memory limits: %+v", err)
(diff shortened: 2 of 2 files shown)


medyagh:
This seems like an interesting refactor, I suggest we merge this after the release so we have more chance to test the possible side effects. But it seems like it might make the code a bit more readable or easier to maintain

daehyeok:
> This seems like an interesting refactor, I suggest we merge this after the release so we have more chance to test the possible side effects. But it seems like it might make the code a bit more readable or easier to maintain

Yes i agree, if i didn't make mistake, this PR will not change program behavior except overwrite more flags. but i'm not sure i added flag which shouldn't overwritten.
 Originally, i tried to fix #10578 , but it was annoying to add, validate flags in the `updateExistingConfigFromFlags` function.  :(

daehyeok:
Few more notes

 - Comment is outdated. 
https://github.com/kubernetes/minikube/blob/f29366a59670d616d71400b8f4a1bad67bf1d62e/cmd/minikube/cmd/start_flags.go#L457-L459
it always called on `start` command by `generateClusterConfig`(so not only used on second start). and `extra-config` value updated if flags are changed.

- Re-pushed brach with split commits for better diff result.

medyagh:
@daehyeok would you please pull upstream one more time ?

daehyeok:
@medyagh  rebased again

PR #10902, opened 2021-03-23
Regenerate documentation to fix failing unit test

afbjorklund:
Currently broken on master


afbjorklund:
Tests didn't kick in ? Not sure why we have a unit test for docs, that looks wrong to me.

Issue #10905, opened 2021-03-24
minikube status: show timeToStop only if it exists


medyagh:
$ minikube status
minikube
type: Control Plane
host: Running
kubelet: Running
apiserver: Running
kubeconfig: Configured
timeToStop: Nonexistent

medyagh:
so the idea is, if minikube is NOT scheduled to stop, it shouldn't bother to mention it in the status

instead of saying timeToStop: Nonexistent

PR #10906, opened 2021-03-24
status: Omit `timeToStop` if nonexistent

tharun208:
Previously, in `minikube status` we show `TimeToStop` as `NonExistent` this pr introduces changes as instead of showing `NonExistent` it does not mention it in the status.

Before:
Screen Shot 2021-03-23 at 2 36 18 PM

After:
Screen Shot 2021-03-24 at 11 43 54 AM

Fixes: #10905

### DIFF
--- a/test/integration/scheduled_stop_test.go
+++ b/test/integration/scheduled_stop_test.go
@@ -67,6 +67,8 @@ func TestScheduledStopWindows(t *testing.T) {
 
 	// sleep for 5 seconds
 	time.Sleep(5 * time.Second)
+	// make sure minikube timetoStop is not present
+	ensureTimeToStopNotPresent(ctx, t, profile)
 	// make sure minikube status is "Stopped"
 	ensureMinikubeStatus(ctx, t, profile, "Host", state.Stopped.String())
 }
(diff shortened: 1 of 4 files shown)


test/integration/scheduled_stop_test.go
```diff
@@ -69,8 +69,6 @@ func TestScheduledStopWindows(t *testing.T) {
 	time.Sleep(5 * time.Second)
 	// make sure minikube status is "Stopped"
 	ensureMinikubeStatus(ctx, t, profile, "Host", state.Stopped.String())
-	// make sure minikube timtostop is "Nonexistent"
-	ensureMinikubeStatus(ctx, t, profile, "TimeToStop", "Nonexistent")
```
tharun208, line 73: @medyagh, what I am thinking is instead of removing this check from the test. Can we check for `TimeToStop` status value to be `""`?

tstromberg: This should check that the `timeToStop` line is omitted entirely.

tstromberg (changes requested):
This PR looks great! Please make sure the integration test is updated so that it would fail if this PR was reverted accidentally in the future.

medyagh:
@tharun208 could this test failure be related ? https://storage.googleapis.com/minikube-builds/logs/10906/c03b282/Docker_Linux.html#fail_TestErrorSpam%2fstatus

medyagh (approved):

PR #10910, opened 2021-03-24
wip: fix golint errors in cmd/minikube/cmd/status

huazhihao:

### DIFF
--- a/cmd/minikube/cmd/status.go
+++ b/cmd/minikube/cmd/status.go
@@ -73,11 +73,18 @@ const (
 
 	// 1xx signifies a transitional state. If retried, it will soon return a 2xx, 4xx, or 5xx
 
-	Starting  = 100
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/status.go
```diff
@@ -73,11 +73,18 @@ const (
 
 	// 1xx signifies a transitional state. If retried, it will soon return a 2xx, 4xx, or 5xx
 
-	Starting  = 100
-	Pausing   = 101
+	// Transient Codes:
+
+	// Starting means Starting
```
medyagh, line 78: how about more meaningful comment ?

medyagh (changes requested):

afbjorklund (changes requested):
These comments don't really add much, it is enough to comment the block.

See: #11155

PR #10917, opened 2021-03-24
Added cloud shell functional tests

### LINKED ISSUE #10643: add integration test on cloud shell for minikube-auth-test
as suggested by Rob
```
 could  use gcr.io/cloudshell-images/minikube-auth-test container.
  It uses the python ADC library to print an access token every 5 seconds 
(it was also previously listing a GCS bucket, but I don't think it's needed to verify the authentication flow).  
I wrapped that in a simple deployment to verify minikube auth when running in Cloud Shell.
```

spowelljr:
Closes #10643

Added a cloud shell functional test script to run tests within cloud shell.

### DIFF
--- a/hack/jenkins/cloud_shell_functional_tests.sh
+++ b/hack/jenkins/cloud_shell_functional_tests.sh
@@ -31,6 +31,15 @@ gcloud cloud-shell ssh --authorize-session << EOF
  CONTAINER_RUNTIME="docker"
  EXTRA_TEST_ARGS="-test.run (TestFunctional|TestAddons)"
 
+ # Need to set these in cloud-shell or will not be present in common.sh
+ MINIKUBE_LOCATION=$MINIKUBE_LOCATION
+ COMMIT=$COMMIT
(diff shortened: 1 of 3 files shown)


medyagh (approved):

hack/jenkins/cloud_shell_functional_tests.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
medyagh, line 1: you would need to add this to also to 
https://github.com/medyagh/minikube/blob/9136127de65b2732a08d132954d09cb9aa90eba9/hack/jenkins/minikube_set_pending.sh#L34

and ensure the List of Jobs that get Commit Status is updated (since I dont see a commit status for cloud shell)

medyagh (changes requested):

hack/jenkins/cloud_shell_functional_tests.sh
spowelljr: Added the Jenkins job to job list

medyagh:
Cloud shell gopogh link is not clickable and not txt either

Issue #7766, opened 2020-04-18
Create archive for the kicbase image


afbjorklund:
(breakout from issue #7012, as a requirement)

In order to be able to show some progress while downloading the big "kicbase" image,
and for accessing it from environments with limited access to Google Container Registry -
we want to create a tarball with the image. This also allows loading it for **all** container runtimes,
the current `github.com/google/go-containerregistry` library only supports the docker daemon.

We want to treat it similar to the `minikube.iso`

----

As workaround, we could load the kicbase image from a tarball that we download ?

```
313M gcr.io/k8s-minikube/kicbase_v0.0.9
docker load -i kicbase_v0.0.9.tar # and similar functions, from the "cruntime" facade
```

That way, cache would also work with the non-daemon drivers (such as podman).

```
docker pull gcr.io/k8s-minikube/kicbase:v0.0.9
docker save gcr.io/k8s-minikube/kicbase:v0.0.9 > kicbase_v0.0.9.tar
```

This would also make it available in china #7447 and without dockerhub #7316

There is a related problem, that the "digest" is **not** loaded from these alternative formats.

The suggestion is that we verify the tarball with a separate sha256 checksum file instead.

medyagh:
any metrics of how much additional tax we will have to pay in seconds for load and save?

afbjorklund:
> ny metrics of how much additional tax we will have to pay in seconds for load and save?

No, but that could be a good investigation for someone. It's a one time thing, though...
For most users, downloading a couple of hundred megabytes take _longer_ than unpacking.

As opposed to the preload, I don't think there will be any _major_ difference in the time.
The goal here is to enable showing a progress bar, and allow using more container runtimes.

It _could_ have been fixed in the library instead, but development seems to have stalled ?

i.e. if [github.com/google/go-containerregistry](https://github.com/google/go-containerregistry/issues) supported progress bars and non-daemon

the first feature (progress) has been open for _months_

and the second feature (daemon) was even scoped out

afbjorklund:
@medyagh : you need to be willing to drop the "digest" demand, though. only works with `pull`.

So the "tag" would have to be enough, otherwise we still need to invoke `docker pull` and friends.

### docker inspect
```json
[
    {
        "Id": "sha256:819f427ce27572658c87e41384c2e67bae040a5942dd692233c19cad218e36b3",
        "RepoTags": [
            "gcr.io/k8s-minikube/kicbase:v0.0.9"
        ],
[...]

PR #10918, opened 2021-03-24
Add code for downloading kicbase image to cache

### LINKED ISSUE #10912: --download-only should store/load kicbase on local filesystem
To ensure minikube can startup in an air-gapped environment.

Related issue: #10149

afbjorklund:
Download the image to the cache, instead of to the daemon.

This way it also works for Podman, and not only for Docker

Left to decide:

1) Should we store the KIC images in a separate directory
Before they were mixed with the container runtime images

2) Delete the old Docker-only code, only use the cache ?
This means **not** using the digest, always load from files

Currently the image is 354M on the disk (gz-compressed)

Closes #10912 #7766


afbjorklund:
Note that in order to make this work for Podman, we need to fix `LoadFromTarball` to be compatible

It would also be nice to avoid loading again, if the image is already present in the container engine...

afbjorklund:
But the current daemon.Write is quite stupid, so not too hard to re-implement with podman:

https://github.com/google/go-containerregistry/blob/main/pkg/v1/daemon/write.go#L57

Instead of using github.com/docker/docker/client, one could just fork a `sudo podman` ?

If we want to implement the lookup, we would also need something instead of daemon.Image

This is the other end, so instead of `| sudo podman load` it would need `sudo podman save |`

----

// Write saves the image into the daemon as the given tag.
func Write(tag name.Tag, img v1.Image) (string, error)

// Tag adds a tag to an already existent image.
func Tag(src, dest name.Tag) error

// Image provides access to an image reference from the Docker daemon,
// applying functional options to the underlying imageOpener before
// resolving the reference into a v1.Image.
func Image(ref name.Reference, options ...ImageOption) (v1.Image, error)

medyagh (approved):

medyagh:
Looks good to me , Thank you for this PR

PR #10921, opened 2021-03-25
Wip: initialization of variable update

Akiros001:


medyagh (changes requested):
@Akiros001 do u mind clarifying what are you trying to solve ?

medyagh:
@Akiros001 would u plz sign the CLA so we could accept this contribution ?

Issue #10925, opened 2021-03-25
Flake TestMultiNode/serial/RestartMultiNode test times out due to drain node process got stuck


prezha:
in TestMultiNode/serial/RestartMultiNode test, sometimes the node drain process gets stuck and the test then times out, as seen here:

https://storage.googleapis.com/minikube-builds/logs/10792/5bac07a/KVM_Linux.html#fail_TestMultiNode%2fserial%2fRestartMultiNode

```
...
	I0325 03:00:49.746371   14119 ssh_runner.go:149] Run: sudo KUBECONFIG=/var/lib/minikube/kubeconfig /var/lib/minikube/binaries/v1.20.2/kubectl drain multinode-20210325025138-7476-m02 --force --grace-period=1 --disable-eviction --ignore-daemonsets --delete-emptydir-data
	I0325 03:00:49.746396   14119 main.go:126] libmachine: (multinode-20210325025138-7476) Calling .GetSSHHostname
	I0325 03:00:49.750904   14119 main.go:126] libmachine: Found binary path at /home/jenkins/workspace/KVM_Linux_integration/out/docker-machine-driver-kvm2
	I0325 03:00:49.750960   14119 main.go:126] libmachine: Launching plugin server for driver kvm2
	I0325 03:00:49.755590   14119 main.go:126] libmachine: (multinode-20210325025138-7476) Calling .GetSSHPort
	I0325 03:00:49.755801   14119 main.go:126] libmachine: (multinode-20210325025138-7476) Calling .GetSSHKeyPath
	I0325 03:00:49.755996   14119 main.go:126] libmachine: (multinode-20210325025138-7476) Calling .GetSSHUsername
	I0325 03:00:49.756157   14119 sshutil.go:53] new ssh client: &{IP:192.168.39.156 Port:22 SSHKeyPath:/home/jenkins/minikube-integration/linux-amd64-kvm2--10792-3069-5bac07ae99bb4f4627dd9f883c8d230b880fa80e/.minikube/machines/multinode-20210325025138-7476/id_rsa Username:docker}
	I0325 03:00:49.764364   14119 main.go:126] libmachine: Plugin server listening at address 127.0.0.1:37305
	I0325 03:00:49.764984   14119 main.go:126] libmachine: () Calling .GetVersion
	I0325 03:00:49.765719   14119 main.go:126] libmachine: Using API Version  1
	I0325 03:00:49.765752   14119 main.go:126] libmachine: () Calling .SetConfigRaw
[...]

PR #10926, opened 2021-03-25
add additional options to avoid node drain or delete getting stuck

prezha:
fixes: #10925 

an example error condition is given in the issue's description

add additional options to avoid kubectl node drain or api node delete processes getting stuck


medyagh (commented):
@prezha do u mind link to the documentation of these flags and evaluate what they exaclty do and what are the possible side effects of these ?

prezha:
> @prezha do u mind link to the documentation of these flags and evaluate what they exaclty do and what are the possible side effects of these ?

@medyagh in the comments just above those changed/added lines there are references to the docs and an explanation of what are they're for: ie, the old line 108 and new line 124:
https://github.com/kubernetes/minikube/blob/b1353c5a6ac2eb8df33943c7ef80d650a95e92df/pkg/minikube/node/node.go#L107-L108
https://github.com/kubernetes/minikube/blob/b1353c5a6ac2eb8df33943c7ef80d650a95e92df/pkg/minikube/node/node.go#L124
do you think more details should be added?

medyagh (approved):

Issue #10790, opened 2021-03-11
Add real addon registry integration test


spowelljr:
Currently, the addon registry integration test tests against the Kubernetes client.
https://github.com/kubernetes/minikube/blob/master/test/integration/addons_test.go#L140

We should add a new test where we actually build and push the image to the registry.

anencore94:
Hi @spowelljr !, If  you don't mind, would I take this issue ?

And I guess you mean the Line #L224 not #L140, right?

https://github.com/kubernetes/minikube/blob/ff0e25ada888cf1bd7b262857ce96531c1231e69/test/integration/addons_test.go#L224

Also, I think testing `docker push` and `docker pull` the image to the registry is fine, building the new image on test time is not what we want to test actually. I think we could use pre-built docker image with .tar or just pull from docker hub and then push it to our addon-registry

By the way, If I make the test, I wonder do we have to push image **in** some pod **or not**(just in the host). I think it is just fine to test push in the host machine(=minikube), if it has docker cli binary or some other requirements I missed.

medyagh:
@anencore94 go for it !

spowelljr:
@anencore94 Yes, I did mean line 224 and not 140, that was my mistake.

PR #10932, opened 2021-03-26
[WIP] Enhance addon registry test

anencore94:
- check if enable to docker push to registry

[Work In Progress]

Here are TO-DO List that I've not decided yet. I need some feedback. Thanks in advance

1. how can we check these when driver is not docker, so docker push is impossible

2. do we need to setup insecure-registry setting ? or any other detour way to docker push to private registry (registry-pod) ?
    - such as docker login or edit /etc/docker/daemon.json)
3. for setup to-be-pushed image, what is the better idea? using docker build with Dockerfile or docker pull ?
    - For now, I think 'docker pull' is simpler than docker build

resolve : #10790

### DIFF
--- a/test/integration/addons_test.go
+++ b/test/integration/addons_test.go
@@ -54,7 +54,6 @@ func TestAddons(t *testing.T) {
 		if err != nil {
 			t.Fatalf("Failed setting GOOGLE_APPLICATION_CREDENTIALS env var: %v", err)
 		}
-
 		err = os.Setenv("GOOGLE_CLOUD_PROJECT", "this_is_fake")
 		defer os.Unsetenv("GOOGLE_CLOUD_PROJECT")
 		if err != nil {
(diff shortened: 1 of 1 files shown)


anencore94:
/assign @spowelljr 

Please let me know if any idea. Thanks :)

afbjorklund:
@anencore94 
>Here are TO-DO List that I've not decided yet. I need some feedback. Thanks in advance

>    how can we check these when driver is not docker, so docker push is impossible

The driver only runs the machine (node), it is the container runtime inside the cluster

Unfortunately `crictl` only implements `push` and not `pull`.

We have done it as a part of the build command (--push flag), but not as-standalone.

That could be an improvement though, allow for pushing ?

>    do we need to setup insecure-registry setting ? or any other detour way to docker push to private registry (registry-pod) ?
        such as docker login or edit /etc/docker/daemon.json)

Currently we use the localhost exception, and setup a proxy on port 5000

>    for setup to-be-pushed image, what is the better idea? using docker build with Dockerfile or docker pull ?
>        For now, I think 'docker pull' is simpler than docker build

You can also just load the prepared image directly from a local tarball

spowelljr:
@anencore94 Since there hasn't been any progress since the suggestions on March 27 I'm going to close this PR. If you have time to work on this in the future and address the comments feel free to reopen, thanks!

PR #10947, opened 2021-03-29
Update olm addon to v0.17.0

### LINKED ISSUE #10768: update the Outdated catalog in OLM addon
I was wondering why my minikube instance is installing old operator versions.

Then I found out that the `CatalogSource` image in the OLM addon points to the SHA of the image that is several months old.  https://github.com/kubernetes/minikube/blob/81eeb7dd979e29a712ac5153fc4b6b40bd519f9f/pkg/minikube/assets/addons.go#L294

It might make sense for this one case to be an exception from https://github.com/kubernetes/minikube/issues/9884 and to use  `latest` tag. Otherwise, every user will have to update the image refernce themself to pick up the latest changes in Operator Catalog.

kadel:
- update olm addon to use OLM v0.17.0

fixes #10768


deploy/addons/olm/olm.yaml.tmpl
```diff
@@ -78,9 +78,9 @@ spec:
           command:
           - /bin/olm
           args:
-          - -namespace
+          - --namespace
```
medyagh, line 81: what is the reason for the additional -- ?

medyagh (changes requested):
the crds.yaml.tmpl diff says 
```
11,302 additions, 10,357 deletions not shown because the diff is too large. Please use a local Git client to view these changes.
```
this doesnt sound right, do u mind clarifying what is inside that file?

kadel:
> the crds.yaml.tmpl diff says
> 
> ```
> 11,302 additions, 10,357 deletions not shown because the diff is too large. Please use a local Git client to view these changes.
> ```
> 
> this doesnt sound right, do u mind clarifying what is inside that file?

those are CustomeResourceDefinitions used by olm taken from v0.17.0 release. You can check it at https://github.com/operator-framework/operator-lifecycle-manager/releases/tag/v0.17.0

deploy/addons/olm/olm.yaml.tmpl
kadel: old OLM versions were using `flags` at some point they switched to `github.com/spf13/pflag` library, hence move from `-` to `--`

sharifelgamal:
Hey @kadel, we have a test (in test/integration/addons_test.go) that currently skip testing the OLM addon. Could you remove the skip so that we can properly test this PR?

Thanks!

kadel:
> Hey @kadel, we have a test (in test/integration/addons_test.go) that currently skips testing the OLM addon. Could you remove the skip so that we can properly test this PR?
> 
> Thanks!

oh, I haven't noticed the integration test. 
Done.

deploy/addons/olm/olm.yaml.tmpl
```diff
@@ -326,8 +328,16 @@ spec:
                   requests:
                     cpu: 10m
                     memory: 50Mi
+                securityContext:
+                  runAsUser: 1000
```
medyagh, line 332: is this required ? this seems to give Root Access to the addon,
do u mind clarifying why we need to run with more special access ?

deploy/addons/olm/olm.yaml.tmpl
```diff
@@ -308,7 +310,7 @@ spec:
                 - --global-namespace
                 - olm
                 image: {{.CustomRegistries.OLM  | default .ImageRepository | default .Registries.OLM }}{{.Images.OLM}}
-                imagePullPolicy: IfNotPresent
+                imagePullPolicy: Always
```
medyagh, line 313: is this change related to bumping the version ?

medyagh (changes requested):

deploy/addons/olm/olm.yaml.tmpl
kadel: I don't work on OLM directly I'm just a user and I did not write this definition. This comes from the official OLM release artifacts. 

But If I'm not mistaken root has uid `0`. This is not giving any special access to the container, it is doing quite the opposite. It ensures that the container can't run anything as root. 
`1000` is the first non-root user.  Even if the image is built in a way that it runs as root inside the container, this will force it to be non-root.

deploy/addons/olm/olm.yaml.tmpl
kadel: yes

deploy/addons/olm/olm.yaml.tmpl
medyagh: that makes sense ! thanks for the clarifying

medyagh:
Looks good to me @kadel  , Thank you for this PR and improving olm addon

look forward for more PRs for you

PR #10967, opened 2021-03-31
correct the error serial number

zhangshj-inspur:
correct the error serial number


sharifelgamal:
Markdown accepts either increasing or non-increasing numbers as a correct list, therefore this PR isn't necessary.

PR #10976, opened 2021-04-02
define a variable instead of duplicating "kube-system"

wangchenglong01:
## Which problem is this PR solving?
- define a variable instead of duplicating "kube-system"

## Short description of the changes
- define a variable instead of duplicating "kube-system"


ilya-zuyev (approved):
lgtm

PR #11003, opened 2021-04-06
Use golang.org/x/term

tklauser:
The `golang.org/x/crypto/ssh/terminal` package is deprecated and merely a
wrapper around `golang.org/x/term`. Thus, use the latter directly.

### DIFF
--- a/cmd/minikube/cmd/config/prompt.go
+++ b/cmd/minikube/cmd/config/prompt.go
@@ -23,7 +23,7 @@ import (
 	"os"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


medyagh (commented):
do u mind clarifying why the site submodule been updated with this PR ? I would rather separate PR for separate changes for auditablity

tklauser:
> do u mind clarifying why the site submodule been updated with this PR ? I would rather separate PR for separate changes for auditablity

Sorry, that change was unintentional. I'll revert that part of the PR.

spowelljr (approved):

PR #11015, opened 2021-04-07
Fix wrong error message on memory validation

afbjorklund:
Before:

❌  Exiting due to MK_CONFIG_SET: run validations for "memory" with value of "foo": [invalid disk size: invalid size: 'foo']

After:

❌  Exiting due to MK_CONFIG_SET: run validations for "memory" with value of "foo": [invalid memory size: invalid size: 'foo']


medyagh:
please check the unit test

PR #11027, opened 2021-04-08
Detect if go.sum changed after running go mod tidy

spowelljr:
Currently, we run `go mod tidy` in `make test` and check if the go.mod changed and fail if it did. However, we don't check if the go.sum changed.

This becomes a problem when we go to release, `go mod tidy` is ran which will correct the go.sum, but makes the commit dirty and we have to make a PR to update the go.sum and run the release again.

This will detect if the go.sum should be updated during the PR unit test check in GitHub and will notify us, preventing the above issue.

Also updated workflow so PR tests are run if the go.mod file is updated. This will prevent some issues where packages are updated, new docs need to be generated, but the linting job isn't run, so we're left unaware until later.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -2,6 +2,7 @@ name: PR
 on:
   pull_request:
     paths:
+      - "go.mod"
       - "**.go"
       - "**.yml"
       - "**.yaml"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - master
     paths:
+      - "go.mod"
       - "**.go"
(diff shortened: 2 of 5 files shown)


.github/workflows/pr.yml
```diff
@@ -2,6 +2,7 @@ name: PR
 on:
   pull_request:
     paths:
+      - "go.mod"
```
medyagh, line 5: add same thing to the master.yml

medyagh (changes requested):

.github/workflows/pr.yml
spowelljr: Done

medyagh (approved):

PR #11039, opened 2021-04-09
bump default and newest k8s version to 1.21

medyagh:


ilya-zuyev:
Looks like v1.21.0 still has a performance regression affecting minikube start time significantly

PR #10975, opened 2021-04-01
change crio-bin checksum since it changed upstream

sharifelgamal:

PR #11046, opened 2021-04-09
Revert "change crio-bin checksum since it changed upstream"

ilya-zuyev:
Reverts kubernetes/minikube#10975


afbjorklund:
They changed it again ? These git archives are unreliable for sure

Issue #10845, opened 2021-03-16
Running minikube on outdated and unsupported platforms


afbjorklund:
Running minikube requires `amd64` for x86, and at least an `armv7` (but preferrably `arm64`) for ARM...

It also requires 2 GB of RAM and 20 GB of hard disk, which might not be available on the local machine.

But we could _still offer_ `minikube start` and the kubernetes experience on these machines, for learning ?

Assuming that there is _another_ computer available to run the actual cluster (but **not** the desktop interface)

~~#4733~~

----

With the possibility of loading and building images in the cluster, and also running kubernetes client there*,
the lack of `docker` and `kubectl` is not really blocking. The client would only be used for editing files etc.

\* #10843

This would allow running minikube on ~~`i586`~~ `i686` and ~~`armv5`~~ `armv6`, towards an `amd64` or `arm64` VM.
Ultimately, we also want to be able to use a different architecture for the host and the cluster (different story*)

\* #9593 

`GOARCH=386`

`GOARCH=arm GOARM=5`

medyagh:
I dont see why not !

tstromberg:
What changes are you suggesting to minikube exactly?

afbjorklund:
> What changes are you suggesting to minikube exactly?

My initial idea was to build a static minikube binary for these, but that also assumes it has access to another machine for the cluster.

So it might need some improvements to the ssh driver, and it would also need splitting of architecture between host and machine.

Basically I want to stretch the definition of "local Kubernetes", without changing the main target or introduce any security concerns.

The end goal is to bring it to more platforms, you might not have access to _one_ big computer (or VM) - but maybe two smaller ?

medyagh:
I believe the pre-requisite for this is, to have our cached fixed

medyagh:
I am going to close this issue for now till we have more interest from the users

Issue #9593, opened 2020-11-01
Add architecture to the binaries in the cache


afbjorklund:
When supporting remote clusters of a different architecture, we need to add the arch in addition to the OS:

```
~/.minikube/cache/linux/v1.19.3
├── kubeadm
├── kubectl
└── kubelet
```

Simplest would be to add it after the OS, we could support **both** in order to preserve existing user caches ?

```
~/.minikube/cache/linux
├── amd64
│   └── v1.19.3
│       ├── kubeadm
│       ├── kubectl
│       └── kubelet
└── arm64
    └── v1.19.3
        ├── kubeadm
        ├── kubectl
        └── kubelet
```

For instance when running an `amd64` host, towards a `arm64` machine. Either local hardware or cloud instance.

Same thing for the images, if not using the preload (the preloaded-tarball files already include the arch, though)

```
~/.minikube/cache/images/amd64
~/.minikube/cache/images/arm64
```

afbjorklund:
Also, this code needs to be parameterized:

```go
                        src, err := download.Binary(name, cfg.KubernetesVersion, "linux", runtime.GOARCH)
                        if err != nil {
                                return errors.Wrapf(err, "downloading %s", name)
                        }

```

Using runtime.GOARCH is a good default, but it needs to be possible to override.

sharifelgamal:
Ok, this has become more of an issue recently, since if someone accidentally runs an amd64 minikube binary on an arm machine, it corrupts their cache. Let's fix this one and for all.

PR #11060, opened 2021-04-10
Allow building minikube for outdated architectures

afbjorklund:
This allows building the minikube client for some outdated 32-bit architectures:

`minikube-linux-i686`

`minikube-linux-armv6`

They can run remotely, but not locally (due to Kubernetes not being available...)

❌  Exiting due to DRV_UNSUPPORTED_OS: The driver 'docker' is not supported on linux/386

❌  Exiting due to DRV_UNSUPPORTED_OS: The driver 'docker' is not supported on linux/armv6

----

However, they can (eventually) be used with the --driver=ssh and --ssh=true

#10845

That would require minikube support for running cross-architecture, though...

#9593

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -206,6 +206,10 @@ out/minikube-windows-amd64.exe: out/minikube-windows-amd64
 	$(if $(quiet),@echo "  CP       $@")
 	$(Q)cp $< $@
 
+out/minikube-linux-i686: out/minikube-linux-386
+	$(if $(quiet),@echo "  CP       $@")
+	$(Q)cp $< $@
+
 out/minikube-linux-x86_64: out/minikube-linux-amd64
 	$(if $(quiet),@echo "  CP       $@")
 	$(Q)cp $< $@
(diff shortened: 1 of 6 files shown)


afbjorklund:
Note that GOARCH="arm" images have GOARM=7 by default, which means that they will **not** work on a Raspberry Pi 1-2...

There are "linux/arm/v7", "linux/arm/v6" sub-platforms, but those are **not** being used by Kubernetes (which uses "linux/arm")

Kubernetes 1.6 dropped support for armv6 (`armel`) and requires armv7 (`armhf`), for the 32-bit ARM platform support.

> Motivation:
>*    The only GOARM=6 board Go will support in go1.8 is the Raspberry Pi 1 which is just too slow to run newer Kubernetes versions.
>*    Small performance improvements when using GOARM=7
>*    The armel (http://hub.docker.com/u/armel) images are not updated as often as the armhf (http://hub.docker.com/u/armhf) images are.

Kubernetes requires a Raspberry Pi 3-4, but then you might as well also consider upgrading OS and going from `arm` to `arm64`.

afbjorklund:
Both `i586` and `armv5` are obsolete, so no use considering those. Go for `i686` and `armv6`, if anything.

All regular users would be using `amd64` and `arm64`, the only stragglers are people on Raspberry Pi OS.

PR #11063, opened 2021-04-10
Allow using podman as well as docker daemon

### LINKED ISSUE #8577: Calling Docker daemon even for other container runtimes
From https://github.com/kubernetes/minikube/issues/8554#issuecomment-650387112

Still have a lot of unconditional calls to Docker, around the code base:

`DigestByDockerLib`

`ExistsImageInDaemon`

`LoadFromTarball`

`WriteImageToDaemon`

Usually these fail since we stop dockerd, but _should_ be conditionalized.

afbjorklund:
Rewrite functions:
* DigestByDockerLib
* ExistsImageInDaemon
* LoadFromTarball
* WriteImageToDaemon

Closes #8577

See https://github.com/kubernetes/minikube/pull/10918#issuecomment-815168383

### DIFF
--- a/pkg/minikube/image/image.go
+++ b/pkg/minikube/image/image.go
@@ -19,7 +19,6 @@ package image
 import (
 	"context"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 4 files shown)


afbjorklund:
Some good news, and some bad news. Now things are implemented both for "daemon" (docker) and for "podman".

But, unfortunately the current logic is totally broken. It uses the Id of the local engine and the Id of the remote engine...

```console
$ sudo podman images busybox
REPOSITORY                 TAG     IMAGE ID      CREATED       SIZE
docker.io/library/busybox  latest  22667f53682a  2 months ago  1.45 MB
```

```console
$ minikube ssh docker images busybox
REPOSITORY   TAG       IMAGE ID       CREATED      SIZE
busybox      latest    388056c9a683   2 days ago   1.23MB
```

So it will _always_ say that image needs a transfer. And it will still skip doing even that comparison, for preload/cache...

`Run: docker image inspect --format {{.Id}} busybox`
"busybox" needs transfer: "busybox" does not exist at hash "22667f53682a2920948d19c7133ab1c9c3f745805c14125859d20cede07f11f9" in container runtime

`Run: docker images --format {{.Repository}}:{{.Tag}}`
Images are preloaded, skipping loading

medyagh:
@afbjorklund the cloud shell test failure is not a usuall flake I have seen. https://storage.googleapis.com/minikube-builds/logs/11063/fa5a811/Docker_Cloud_Shell.html#fail_TestFunctional%2fparallel%2fLoadImage

medyagh:
@afbjorklund one more rebase please

afbjorklund:
> @afbjorklund one more rebase please

Hopefully this also fixes the scenario with the new "kic" cache, where the image is already loaded into daemon/podman.

medyagh (approved):

PR #11087, opened 2021-04-12
don't run error spam test in parallel

sharifelgamal:
This is an attempt to diagnose #10628, which we currently can't since it's not getting run. My assumption is that it's getting pre-empted by the functional tests then shut down for some reason.


medyagh (approved):
good idea

sharifelgamal:
Well, the test is failing on Windows still but at least it's still getting run now. I'll work on fixing the issue itself in a future PR.

NicolasSequeira (commented):
IGNORE

PR #11122, opened 2021-04-17
Better to use "0777" than "0o777",instead of only in Go

lilongfeng0902:
fileMode type is uint32, I think the value is 0o777,which is not correct, please check my proposal whether is reasonable.

### DIFF
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -62,7 +62,7 @@ var RootCmd = &cobra.Command{
 	Long:  `minikube provisions and manages local Kubernetes clusters optimized for development workflows.`,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		for _, path := range dirs {
-			if err := os.MkdirAll(path, 0777); err != nil {
+			if err := os.MkdirAll(path, 0o777); err != nil {
 				exit.Error(reason.HostHomeMkdir, "Error creating minikube directory", err)
 			}
 		}
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -62,7 +62,7 @@ var RootCmd = &cobra.Command{
 	Long:  `minikube provisions and manages local Kubernetes clusters optimized for development workflows.`,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		for _, path := range dirs {
-			if err := os.MkdirAll(path, 0o777); err != nil {
+			if err := os.MkdirAll(path, 0777); err != nil {
 				exit.Error(reason.HostHomeMkdir, "Error creating minikube directory", err)
 			}
 		}


afbjorklund (changes requested):
They are both the same value, `0777` in octal is the same as `511` in decimal.

https://golang.org/ref/spec#Integer_literals

lilongfeng0902:
> They are both the same value, `0777` in octal is the same as `511` in decimal.
> 
> https://golang.org/ref/spec#Integer_literals

Now I kown that 0777、0o777、0O777 are both the same value，but I‘m confusing that why change 0777 to 0o777. Other way, there has a  compile error in Goland tool，so I suggest this PR. Can you explain my confusing? Thanks.

afbjorklund:
> Other way, there has a compile error in Goland tool，so I suggest this PR. Can you explain my confusing? Thanks.

Can you show the compile error ? I think you will find octal contents all over the place (in historical `0111` notation)

But we should probably pick one style everywhere. Note that `0o755` doesn't work in shell commands, for instance.

`chmod: invalid mode: ‘0o755’`

lilongfeng0902:
> > Other way, there has a compile error in Goland tool，so I suggest this PR. Can you explain my confusing? Thanks.
> 
> Can you show the compile error ? I think you will find octal constants all over the place (in historical `0111` notation)
> 
> But we should probably pick one style everywhere. Note that `0o755` doesn't work in shell commands, for instance.
> 
> `chmod: invalid mode: ‘0o755’`

Now I understand why that changes “0777” to “0o777”. In my Goland tool, "o777" is treated as variable, the compile error tell me`Unresloved reference 'o777'`  What should I do?

afbjorklund:
> In my Goland tool, "o777" is treated as variable, the compile error tell me`Unresloved reference 'o777'`

OK, now I see the issue. It is either `0777` or `0o777` - but not `o777` which is invalid syntax.

You might want to get a "programmer font" installed, that makes it easier to tell `0` from `O`...

lilongfeng0902:
> > In my Goland tool, "o777" is treated as variable, the compile error tell me`Unresloved reference 'o777'`
> 
> OK, now I see the issue. It is either `0777` or `0o777` - but not `o777` which is invalid syntax.
> 
> You might want to get a "programmer font" installed, that makes it easier to tell `0` from `O`...
> 
> For instance, this one:
> 
> https://blog.jetbrains.com/blog/2020/01/15/jetbrains-mono-a-new-font-made-for-developers/
> 
> jetbrains-mono-font-symbols

image
I think that you haven't seen my issue, now I upload the picture about the error. This maybe can help you to unnderstand me. However I have used “JetBrains Mono” font, it looks very good to me. Thank you.

afbjorklund:
If your IDE doesn't understand octal literals, that sounds like an IDE bug. It _should_ accept `0o777` since it is in the Go standard.

lilongfeng0902:
> If your IDE doesn't understand octal literals, that sounds like an IDE bug. It _should_ accept `0o777` since it is in the Go standard.

Yeah，I agree with you. Maybe I need to track and understand IDE bug, but I prefer to take time to  study the code of minikube. Thank you again. ^o^

afbjorklund:
Actually the "o" octal literals did break some other tools as well, such as old `gocyclo`. So we should try to avoid them.

Better to use `0777`, since it works in all the tools and also in other languages such as bash (instead of only in Go)

lilongfeng0902:
> Actually the "o" octal literals did break some other tools as well, such as old `gocyclo`. So we should try to avoid them.
> 
> Better to use `0777`, since it works in all the tools and also in other languages such as bash (instead of only in Go)

Whether can I open this PR again, it means my PR was reasonable. I'm so happy to make contribute to the minikube community.

lilongfeng0902:
Reasonable than open. ^0^

medyagh (approved):
Looks good to me @lilongfeng0902 , Thank you for this PR and improving minikube, this will prevent another user building minikube with IDEs get fraustrated, thank you.

btw I am curious how did you end up touching that part of the code, were u trying to add a feature or a bug ? 

look forward for more PRs for you

lilongfeng0902:
> Looks good to me @lilongfeng0902 , Thank you for this PR and improving minikube, this will prevent another user building minikube with IDEs get fraustrated, thank you.
> 
> btw I am curious how did you end up touching that part of the code, were u trying to add a feature or a bug ?
> look forward for more PRs for you

As mentioned above, I am just learning about minikube, like how to organize the code, because "root.go" is an entry file, I look it first, but is‘s not compiled exactly in my IDE, I realized that this is an improvement, then I proposed a PR. Maybe it's lucky for me. In short, I am very happy to make a meaningful contribution to the community. ^0^

PR #11127, opened 2021-04-18
new command "image pull: allow to load remote images directly without cache

afbjorklund:
When looking at the benchmarks, there seems to be a need to pull images without too much intermediate layers...

This bypasses the local daemon and the file cache, and asks the container runtime to pull directly from the registry.

Basically the same as:

`minikube ssh -- sudo crictl pull image`

### DIFF
--- a/pkg/minikube/machine/cache_images.go
+++ b/pkg/minikube/machine/cache_images.go
@@ -294,6 +294,86 @@ func transferAndLoadImage(cr command.Runner, k8s config.KubernetesConfig, src st
 	return nil
 }
 
+// pullImages pulls images to the container run time
+func pullImages(cruntime cruntime.Manager, images []string) error {
+	klog.Infof("PullImages start: %s", images)
+	start := time.Now()
+
+	defer func() {
+		klog.Infof("PullImages completed in %s", time.Since(start))
+	}()
+
+	var g errgroup.Group
+
+	for _, image := range images {
+		image := image
+		g.Go(func() error {
+			return cruntime.PullImage(image)
+		})
+	}
+	if err := g.Wait(); err != nil {
+		return errors.Wrap(err, "error pulling images")
+	}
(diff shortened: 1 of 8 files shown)


afbjorklund:
@spowelljr : Would avoid having to save + scp + load the image help anything, with your benchmarks ?

```
crane pull image image.tar
minikube cp image.tar /tmp/image.tar
minikube ssh -- docker load -i /tmp/image.tar
```

vs

```
minikube ssh -- docker pull image
```

medyagh:
> uld avoid having to save + scp + load the image help anything, with your benchmarks ?
> 
> minikube image load image

@afbjorklund would this be only useful for publicly available images ?

afbjorklund:
> would this be only useful for publicly available images ?

You would have to run a registry _somewhere_, but it doesn't have to use docker hub or any other publicly available one.

The main difference is if you plan to do it more than once (e.g. for another cluster, later on). Then you might want to cache.

medyagh (approved):

pkg/minikube/machine/cache_images.go
```diff
@@ -294,6 +294,86 @@ func transferAndLoadImage(cr command.Runner, k8s config.KubernetesConfig, src st
...
+		})
+	}
+	if err := g.Wait(); err != nil {
+		return errors.Wrap(err, "error pulling images")
+	}
+	klog.Infoln("Successfully pulled images")
+	return nil
+}
+
+func DoPullImages(images []string, profile *config.Profile) error {
```
medyagh, line 321: exported funcs, add comment plz

medyagh (changes requested):

pkg/minikube/machine/cache_images.go
afbjorklund: will do, the others didn't have any either

medyagh (approved):

PR #11128, opened 2021-04-18
Add API for saving an image to a local path

afbjorklund:
This function SaveImage is the opposite of LoadImage,
useful for saving built images to the cache directory.

This is an alternative to using the push functionality,
which requires a registry (similar to the pull function).

For #11130

### DIFF
--- a/pkg/minikube/cruntime/cruntime.go
+++ b/pkg/minikube/cruntime/cruntime.go
@@ -99,6 +99,8 @@ type Manager interface {
 	PullImage(string) error
 	// Build an image idempotently into the runtime on a host
 	BuildImage(string, string, string, bool, []string, []string) error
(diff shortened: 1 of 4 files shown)


pkg/minikube/cruntime/cruntime.go
```diff
@@ -99,6 +99,8 @@ type Manager interface {
 	PullImage(string) error
 	// Build an image idempotently into the runtime on a host
 	BuildImage(string, string, string, bool, []string, []string) error
+	// Load an image idempotently into the runtime on a host
```
medyagh, line 102: Fix comment

medyagh (changes requested):

medyagh:
Looks good to me

medyagh (approved):

PR #11132, opened 2021-04-18
Add checksum to the preload tarball download

### LINKED ISSUE #11061: VM starts but the download of `preloaded-images-k8s` was not completed
*(This is not a regular issue, but a support request)*

```
minikube start
minikube delete
minikube start
```
>😄  minikube v1.19.0 on Darwin 10.14.6
>    ▪ MINIKUBE_ACTIVE_DOCKERD=minikube
>✨  Using the virtualbox driver based on user configuration
>👍  Starting control plane node minikube in cluster minikube
>💾  Downloading Kubernetes v1.20.2 preload ...
>    \> preloaded-images-k8s-v10-v1...: 308.70 MiB / 491.71 MiB  62.78% 84.05 KiB
>🔥  Creating virtualbox VM (CPUs=3, Memory=5120MB, Disk=20000MB) ...
> ...

### Questions:

* Does Minikube has any way to *avoid re-downloading the `preloaded-images-k8s`, every time* that a VM is created?
* Where does Minikube stores the `preloaded-images-k8s`?\
So I can manually *copy it* before `delete` and *restore it* before `start`.

I am having connectivity problems and I would like to **reuse the images that were already downloaded on the first execution of `minikube start`**, when re-creating the Virtual Machine.

Is there **any** way to achieve this?

afbjorklund:
Check the checksum before putting the download into the cache,
leave the previous checksum file even if it is now redundant.

Closes #11061

### DIFF
--- a/pkg/minikube/download/preload.go
+++ b/pkg/minikube/download/preload.go
@@ -143,8 +143,15 @@ func Preload(k8sVersion, containerRuntime string) error {
 	url := remoteTarballURL(k8sVersion, containerRuntime)
 
 	checksum, err := getChecksum(k8sVersion, containerRuntime)
+	var realPath string
 	if err != nil {
 		klog.Warningf("No checksum for preloaded tarball for k8s version %s", k8sVersion)
+		realPath = targetPath
+		tmp, err := ioutil.TempFile(targetDir(), TarballName(k8sVersion, containerRuntime)+".*")
+		if err != nil {
+			return errors.Wrap(err, "tempfile")
+		}
+		targetPath = tmp.Name()
 	} else if checksum != "" {
 		url += "?checksum=" + checksum
 	}
@@ -159,6 +166,14 @@ func Preload(k8sVersion, containerRuntime string) error {
 
 	if err := verifyChecksum(k8sVersion, containerRuntime, targetPath); err != nil {
 		return errors.Wrap(err, "verify")
+	}
+
+	if realPath != "" {
+		klog.Infof("renaming tempfile to %s ...", TarballName(k8sVersion, containerRuntime))
+		err := os.Rename(targetPath, realPath)
+		if err != nil {
+			return errors.Wrap(err, "rename")
+		}
 	}
 
 	return nil
--- a/pkg/minikube/download/preload.go
+++ b/pkg/minikube/download/preload.go
@@ -161,16 +161,24 @@ func Preload(k8sVersion, containerRuntime string) error {
 	return nil
 }
 
-func getChecksum(k8sVersion, containerRuntime string) (string, error) {
-	klog.Infof("getting checksum for %s ...", TarballName(k8sVersion, containerRuntime))
+func getStorageAttrs(name string) (*storage.ObjectAttrs, error) {
 	ctx := context.Background()
 	client, err := storage.NewClient(ctx, option.WithoutAuthentication())
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/minikube/download/preload.go
```diff
@@ -141,6 +142,10 @@ func Preload(k8sVersion, containerRuntime string) error {
 	out.Step(style.FileDownload, "Downloading Kubernetes {{.version}} preload ...", out.V{"version": k8sVersion})
 	url := remoteTarballURL(k8sVersion, containerRuntime)
 
+	if checksum, err := getChecksum(k8sVersion, containerRuntime); err == nil {
+		url += "?checksum=" + checksum
+	}
```
medyagh, line 147: how about adding a warning log when there is err is not nil and just log it for reference ?

medyagh (changes requested):

pkg/minikube/download/preload.go
afbjorklund: we already break when failing to save the checksum, so might as well do the same here

pkg/minikube/download/preload.go
```diff
@@ -141,6 +142,12 @@ func Preload(k8sVersion, containerRuntime string) error {
 	out.Step(style.FileDownload, "Downloading Kubernetes {{.version}} preload ...", out.V{"version": k8sVersion})
 	url := remoteTarballURL(k8sVersion, containerRuntime)
 
+	checksum, err := getChecksum(k8sVersion, containerRuntime)
+	if err != nil {
+		return errors.Wrap(err, "getting checksum")
```
medyagh, line 147: should we return error here or should just at least try to download it regardless of being able to get checksum ?

afbjorklund: It will fail a couple of lines down anyway, when it tries to save the same checksum to a file.

afbjorklund: I don't know how unrealiable Google Cloud Storage, if it usually fails when trying to get metadata ?

afbjorklund: Implemented the version without a checksum up front, doing it all manually instead (tempfile + verify)

afbjorklund:
The code would look much better, if we just let go-getter do the tempfile and the validation.

pkg/minikube/download/preload.go
```diff
@@ -141,6 +142,20 @@ func Preload(k8sVersion, containerRuntime string) error {
 	out.Step(style.FileDownload, "Downloading Kubernetes {{.version}} preload ...", out.V{"version": k8sVersion})
 	url := remoteTarballURL(k8sVersion, containerRuntime)
 
+	checksum, err := getChecksum(k8sVersion, containerRuntime)
+	var realPath string
+	if err != nil {
+		klog.Warningf("No checksum for preloaded tarball for k8s version %s", k8sVersion)
```
medyagh, line 148: please log the error there, in addition to the k8sVersion

medyagh (changes requested):

pkg/minikube/download/preload.go
afbjorklund: Done

medyagh (approved):

medyagh:
Looks good to me, Thank you for this PR

Pictor13:
Thank you for this fix! :)

PR #11151, opened 2021-04-19
Pre-cleanup for GitHub actions self-hosted machines

ilya-zuyev:
This PR:
- adds a cleanup step to remove all minikube/docker leftovers on arm64 external runner machines
- adds `docker system prune --volumes` for windows self-hosted jobs (`docker system prune` is already there, but it doesn't remove volumes by default)

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -332,7 +332,8 @@ jobs:
           $docker_running = $?
           }
           Write-Output "Docker is running"
-          docker system prune -f 
+          docker system prune -f
+          docker system prune --volumes -f
       - name: Info
         shell: powershell
         run: |
@@ -471,6 +472,7 @@ jobs:
           }
           Write-Output "Docker is running"
           docker system prune -f
+          docker system prune --volumes -f
       - name: Info
         continue-on-error: true
         shell: powershell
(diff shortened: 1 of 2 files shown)


.github/workflows/master.yml
```diff
@@ -711,6 +711,13 @@ jobs:
         with:
           name: minikube_binaries
 
+      - name: Pre-cleanup
```
medyagh, line 714: this build step runs on GH action VMs and it will not run on the Self-hosted VMS
this pre clean up needs to be added to Two Windows, ARM Jobs,

medyagh (changes requested):

.github/workflows/master.yml
ilya-zuyev: it's in `functional_docker_ubuntu_arm64` job and runs only on self-hosted arm64 runner as it's stated in PR name. AFAIK, we only have the issue with leftovers there. 
Do we need to fix windows machines as well?

ilya-zuyev: https://github.com/kubernetes/minikube/pull/11151/checks?check_run_id=2385428597 runs on our arm64 aws instance  as expected

medyagh: sorry I was confused, you are right, yes let please let add to windows machines as well, since they are also self hosted

ilya-zuyev: done!

PR #11157, opened 2021-04-20
test: Add target for the test report as html

afbjorklund:
Just something I did during lunch, easier to read than the JSON

`go get -u github.com/vakenbolt/go-test-report`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ _cgo_export.*
 
 _testmain.go
 
+test-report.html
 coverage.html
 
 *.exe
--- a/.gitignore
+++ b/.gitignore
@@ -22,9 +22,6 @@ _cgo_export.*
 
 _testmain.go
 
-test-report.html
-coverage.html
-
 *.exe
 *.test
 *.prof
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -374,6 +374,11 @@ out/test-report.json: $(SOURCE_FILES) $(GOTEST_FILES)
 	-coverprofile=out/coverage.out -json > out/test-report.json
 out/coverage.out: out/test-report.json
 
+# Generate go test report (from gotest) as a a HTML page
+test-report.html: out/test-report.json
```
medyagh, line 378: could we call it unit-test-report.html ?

.gitignore
```diff
@@ -22,6 +22,7 @@ _cgo_export.*
 
 _testmain.go
 
+test-report.html
```
medyagh, line 25: how about we generate this in the /out/unit-test-report.html ?

Makefile
medyagh: I wanna ensure this wont confuse the users of 
```
$ make help | grep html
html_report                    Generate HTML  report out of the last ran integration test logs.
```

medyagh: or you could make the "html_report" convert all tests to html (including last integraiton and unit test)
that way we dont have to remember which one to call...

jsut ask makefile to give us html for all tests in out/unit-test.html

medyagh (changes requested):

Makefile
afbjorklund: Renamed and moved to out

medyagh (approved):
Looks good to me, Thank you for adding more useful unit test report

PR #11158, opened 2021-04-20
Add log file to GitHub issue output

### LINKED ISSUE #9398: Fatal errors should include the path to the post-mortem INFO log
Currently, we provide this information, which no one sees: https://minikube.sigs.k8s.io/docs/handbook/troubleshooting/#post-mortem-minikube-debug-logs

I was thinking that when we have a fatal exit, we should be able to say something like:

```
Please open a GitHub issue. If you are able to drag and drop the following log-file into the issue, we'll be able to make faster progress: /path/to/minikube.INFO.log
```

Related: #9323

spowelljr:
Closes #9398

Adds a line to the output when we tell the user to create a GitHub issue. The line contains the path of the latest log file to be attached to the GitHub issue to aid in debugging the issue.

Tested on Mac and Linux so far, will test on Windows as well.

Before
```
❌  Exiting due to GUEST_START: wait 6m0s for node: waiting for apps_running: expected k8s-apps: missing components: kube-dns

```

After
```
❌  Exiting due to GUEST_START: wait 6m0s for node: waiting for apps_running: expected k8s-apps: missing components: kube-dns

│    - /Users/powellsteven/.minikube/logs/lastStart.txt              │
```

### DIFF
--- a/pkg/minikube/out/out.go
+++ b/pkg/minikube/out/out.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -384,8 +385,9 @@ func getLatestLogFilePath() (string, error) {
 		lastModName = file.Name()
 		lastModTime = file.ModTime()
 	}
+	fullPath := filepath.Join(tmpdir, lastModName)
 
-	return tmpdir + lastModName, nil
+	return fullPath, nil
 }
 
 func displayLogLocationMessage() error {
--- a/pkg/minikube/out/out.go
+++ b/pkg/minikube/out/out.go
@@ -334,7 +334,7 @@ func wantsColor(w fdWriter) bool {
 
 // LogEntries outputs an error along with any important log entries.
 func LogEntries(msg string, err error, entries map[string][]string) {
-	DisplayError(msg, err)
+	displayError(msg, err)
 
 	for name, lines := range entries {
 		Step(style.Failure, "Problems detected in {{.entry}}:", V{"entry": name})
@@ -347,8 +347,8 @@ func LogEntries(msg string, err error, entries map[string][]string) {
 	}
 }
 
-// DisplayError prints the error and displays the standard minikube error messaging
-func DisplayError(msg string, err error) {
+// displayError prints the error and displays the standard minikube error messaging
+func displayError(msg string, err error) {
 	klog.Warningf(fmt.Sprintf("%s: %v", msg, err))
 	if JSON {
 		FatalT("{{.msg}}: {{.err}}", V{"msg": translate.T(msg), "err": err})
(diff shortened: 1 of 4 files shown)


medyagh (commented):
please add before after this PR to the description

pkg/minikube/out/out.go
```diff
@@ -359,6 +360,29 @@ func DisplayError(msg string, err error) {
...
+	}
+	ErrT(style.Tip, "If you are able to drag and drop the following log-file into the issue, we'll be able to make faster progress: {{.logFileName}}", V{"logFileName": logFileName})
+}
+
+func getLatestLogFileName() (string, error) {
+	dir := os.Getenv("TMPDIR")
+	if dir == "" {
+		dir = "/tmp/"
+	}
+	args := fmt.Sprintf("cd %s && echo %s$(ls -t *minikube_*_*_*log | head -1)", dir, dir)
```
medyagh, line 376: this wont work on windows.

how about we point them to the log file we store in the minikube home folder ? 

ideally if there is a trick to Open the Issue with that file attached would best

medyagh (changes requested):

pkg/minikube/out/out.go
spowelljr: We could just always point them to the start file, but what if `minikube addon enable ingress` fails? Should we still point them to the start logs?

spowelljr: Also, I don't think there is anyway to create a URL that will attach the file. Sounds like that would be a large security threat if it was a thing.

medyagh: good point, how about we detect what command is failing, if it is not start, we will look for latest temp files.

regardless,  this command wont work windows, can you make it a Go way of finding latest temp files that has word minikube in their file name?

medyagh (changes requested):

pkg/minikube/out/out.go
spowelljr: > how about we detect what command is failing, if it is not start, we will look for latest temp files
Can do that for sure

> can you make it a Go way of finding latest temp files
Will look into how to do that

medyagh (changes requested):
also change UI to be more easy to see 
maybe use the Red Box ui helper we have

```
❌  Exiting due to GUEST_START: wait 6m0s for node: waiting for apps_running: expected k8s-apps: missing components: kube-dns

😿  If the above advice does not help, please let us know by filing 
----- Red Box thing ----- ----- ----- ----- -----  --

   Drag and rop the following files into a github issue...
    -  /Users/powellsteven/.minikube/logs/lastStart.txt
----- ----- ----- ----- ----- ----- ----- 
```

pkg/minikube/out/out.go
spowelljr: Done

medyagh:
please update the UI in the PR description whenever new UI ready

medyagh (changes requested):
lets add back the crying cat emoji
`😿 `

medyagh (approved):

PR #11168, opened 2021-04-22
Use canonical image name if one exists remotely

### LINKED ISSUE #11126: Use canonical image names when adding to the cache
Currently, we save the images as archives using only a short name such as "busybox".

```json
[
  {
    "Config": "sha256:388056c9a6838deea3792e8f00705b35b439cf57b3c9c2634fb4e95cfc896de6",
    "RepoTags": [
      "busybox"
    ],
    "Layers": [
      "f531cdc67389c92deac44e019e7a1b6fba90d1aaa58ae3e8192f0e0eed747152.tar.gz"
    ]
  }
]
```

We should make sure to use the canonical name, e.g. "docker.io/library/busybox:latest"

```json
[
  {
    "Config": "sha256:388056c9a6838deea3792e8f00705b35b439cf57b3c9c2634fb4e95cfc896de6",
    "RepoTags": [
      "docker.io/library/busybox:latest"
    ],
    "Layers": [
      "f531cdc67389c92deac44e019e7a1b6fba90d1aaa58ae3e8192f0e0eed747152.tar.gz"
    ]
  }
]
```

The go-container-registry library can do this (`Name`), only have to work around index.docker.io

```go
var (
	legacyDefaultDomain = "index.docker.io"
	defaultDomain       = "docker.io"
	officialRepoName    = "library"
	defaultTag          = "latest"
)
```

```go
       // Rewrite "docker.io" to "index.docker.io".
        // See: https://github.com/google/go-containerregistry/issues/68
        if name == defaultRegistryAlias {
                name = DefaultRegistry
        }

```

Having the canonical name means that we can load them into container runtimes such as cri-o.

The library calls this "strict validation", as opposed to "weak validation" (without registry and tag)

----

Eventually it also opens up for having a _different_ default registry.

```
/ WithDefaultRegistry sets the default registry that will be used if one is not
// provided.
func WithDefaultRegistry(r string) Option
```

But in the short term, it fixes some shortcomings with the cache: #8554

See also https://www.redhat.com/sysadmin/container-image-short-names

We can use the same short name as argument and as a cache filename.

afbjorklund:
Testing with one image from remote, and one image that only exists in daemon:

```console
$ docker pull hello-world
latest: Pulling from library/hello-world
b8dfde127a29: Pull complete 
Digest: sha256:f2266cbfc127c960fd30e76b7c792dc23b588c0db76233517e1891a4e357d519
Status: Downloaded newer image for hello-world:latest
docker.io/library/hello-world:latest
$ docker tag hello-world hw
```

```console
$ docker images
REPOSITORY    TAG       IMAGE ID       CREATED       SIZE
hello-world   latest    d1165f221234   6 weeks ago   13.3kB
hw            latest    d1165f221234   6 weeks ago   13.3kB
```

----

When adding an image to the cache, check if it exists remote.
If it does, then use the canonical name when saving to cache.

`docker.io/library/hello-world:latest`

If **not** checking with the remote registry, or it doesn't exist,
then use the short name when saving to cache - like before...

`hw:latest`

This means that cri-o will add the "localhost" prefix, when loading:

```
IMAGE                                     TAG                  IMAGE ID            SIZE
docker.io/library/hello-world             latest               d1165f2212346       17.7kB
localhost/hw                              latest               d1165f2212346       17.7kB
```

But the regular images will work with any runtime, including cri-o.

Closes #11126 (and #8554)

### DIFF
--- a/pkg/minikube/image/cache.go
+++ b/pkg/minikube/image/cache.go
@@ -20,7 +20,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/name"
@@ -34,11 +33,6 @@ import (
 	"k8s.io/minikube/pkg/minikube/localpath"
 	"k8s.io/minikube/pkg/minikube/out"
 	"k8s.io/minikube/pkg/util/lock"
-)
-
-const (
-	legacyDefaultDomain = "index.docker.io"
-	defaultDomain       = "docker.io"
 )
 
 type cacheError struct {
@@ -142,9 +136,6 @@ func saveToTarFile(iname, rawDest string) error {
 	}
 	if img == nil {
 		return errors.Wrapf(err, "nil image for %s", iname)
-	}
-	if strings.HasPrefix(cname, legacyDefaultDomain) {
-		cname = strings.Replace(cname, legacyDefaultDomain, defaultDomain, 1)
 	}
 
 	// use new canonical name
--- a/pkg/minikube/image/cache.go
+++ b/pkg/minikube/image/cache.go
@@ -138,13 +138,15 @@ func saveToTarFile(iname, rawDest string) error {
 		return errors.Wrapf(err, "nil image for %s", iname)
 	}
 
-	// use new canonical name
-	ref, err = name.ParseReference(cname, name.WeakValidation)
-	if err != nil {
-		return errors.Wrapf(err, "parsing image ref name for %s", cname)
-	}
-	if ref == nil {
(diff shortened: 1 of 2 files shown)


medyagh:
@afbjorklund does it help with https://github.com/kubernetes/minikube/issues/11102#issuecomment-825083124 ?

afbjorklund:
> @afbjorklund does it help with [#11102 (comment)](https://github.com/kubernetes/minikube/issues/11102#issuecomment-825083124) ?

No, this is only for the name - it doesn't change anything regarding digest or arch.

medyagh:
Looks good to me @afbjorklund  , Thank you for this PR and improving minikube image pulling

PR #11169, opened 2021-04-22
Add opt-in notifications for beta releases

### LINKED ISSUE #10328: allow  Minikube users to config receive notification for BETA and RC

spowelljr:
Closes #10328

Added a new config option that the user can enable to get notified of beta releases.

The user can opt-in with the following command:
```
minikube config set WantBetaUpdateNotification true
```

Release script need to be updated to update the `releases-beta.json` file with the release version.


spowelljr:
Just realized that with the current logic, a older beta will take priority over a newer full release. Will resolve this conflict and add a test for it.

medyagh (changes requested):
lets change the get Names to proper go idiomatic and then looks good to me

medyagh (approved):

PR #11180, opened 2021-04-23
fix kubeadm join error: "Found multiple CRI sockets"

### LINKED ISSUE #11179: kubeadm join error: "Found multiple CRI sockets, please use --cri-socket to select one: /var/run/dockershim.sock, /var/run/crio/crio.sock"
kubeadm join will fail with this error if multiple CRI sockets exist (eg, both docker and crio are started at the same time)

/assign

prezha:
fixes #11179

### DIFF
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -758,6 +758,15 @@ func (k *Bootstrapper) JoinCluster(cc config.ClusterConfig, n config.Node, joinC
 	// Join the master by specifying its token
 	joinCmd = fmt.Sprintf("%s --node-name=%s", joinCmd, config.MachineName(cc, n))
 
+	// avoid "Found multiple CRI sockets, please use --cri-socket to select one: /var/run/dockershim.sock, /var/run/crio/crio.sock" error
+	cr, err := cruntime.New(cruntime.Config{Type: cc.KubernetesConfig.ContainerRuntime, Runner: k.c, Socket: cc.KubernetesConfig.CRISocket})
+	if err != nil {
+		return errors.Wrap(err, "runtime")
+	}
+	if sp := cr.SocketPath(); sp != "" {
+		joinCmd = fmt.Sprintf("%s --cri-socket=%s", joinCmd, sp)
+	}
+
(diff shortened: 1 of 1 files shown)


afbjorklund (changes requested):
There is already a check for this when generating the join command, but it doesn't add any configuration when using the (previously default) dockershim.

```go
        if cc.KubernetesConfig.CRISocket != "" {
                joinCmd = fmt.Sprintf("%s --cri-socket %s", joinCmd, cc.KubernetesConfig.CRISocket)
        }
```

We have yet another hack in DeleteCluster, which has the right code:

```go
        sp := cr.SocketPath()
        if sp == "" {
                sp = kconst.DefaultDockerCRISocket
        }

```

prezha:
> There is already a check for this when generating the join command, but it doesn't add any configuration when using the (previously default) dockershim.
> 
> ```go
>         if cc.KubernetesConfig.CRISocket != "" {
>                 joinCmd = fmt.Sprintf("%s --cri-socket %s", joinCmd, cc.KubernetesConfig.CRISocket)
>         }
> ```
> 
> We have yet another hack in DeleteCluster, which has the right code:
> 
> ```go
>         sp := cr.SocketPath()
>         if sp == "" {
>                 sp = kconst.DefaultDockerCRISocket
>         }
> ```

instead of those (now 3) hacks, wouldn't it be then maybe better to set it at the beginning - eg, in cmd.generateClusterConfig right after https://github.com/kubernetes/minikube/blob/d1b09893e7eaa93383870bfcb3c546a310a99ab8/cmd/minikube/cmd/start_flags.go#L272-L275 ?

prezha:
> There is already a check for this when generating the join command, but it doesn't add any configuration when using the (previously default) dockershim.
> 
> ```go
>         if cc.KubernetesConfig.CRISocket != "" {
>                 joinCmd = fmt.Sprintf("%s --cri-socket %s", joinCmd, cc.KubernetesConfig.CRISocket)
>         }
> ```
> 
> We have yet another hack in DeleteCluster, which has the right code:
> 
> ```go
>         sp := cr.SocketPath()
>         if sp == "" {
>                 sp = kconst.DefaultDockerCRISocket
>         }
> ```

@afbjorklund agree, i've used the 'right code' to fix the existing join cmd in #11209 

i'll close this one in favour of #11209
/close

PR #11191, opened 2021-04-25
completion: add specific shell completion

zhijianli88:
completion: add specific shell completion

Previously, `minikube completion` is unable to auto-complete the
specific shell as other go-lang applications. For example, when user
type below command:
```
minikube$ minikube completion <tab> <tab>
```
In minikube, there is no candidate shells coming up, which is different
from kubectl:
```
$ kubectl completion <tab> <tab>
bash  zsh
```

After this patch, as following, the candidata shell appears:
```
minikube$ minikube completion <tab> <tab>
bash  fish  zsh
```

### DIFF
--- a/cmd/minikube/cmd/completion.go
+++ b/cmd/minikube/cmd/completion.go
@@ -77,23 +77,50 @@ var completionCmd = &cobra.Command{
 		}
 		if args[0] != "bash" && args[0] != "zsh" && args[0] != "fish" {
 			exit.Message(reason.Usage, "Sorry, completion support is not yet implemented for {{.name}}", out.V{"name": args[0]})
-		} else if args[0] == "bash" {
-			err := GenerateBashCompletion(os.Stdout, cmd.Parent())
-			if err != nil {
-				exit.Error(reason.InternalCompletion, "bash completion failed", err)
-			}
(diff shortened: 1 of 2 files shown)


medyagh (changes requested):
@zhijianli88 thank you very much for this PR, does this break the old users who did NOT specify the type of the bash ?

Could we make it backward compatible ?

zhijianli88:
> @zhijianli88 thank you very much for this PR, does this break the old users who did NOT specify the type of the bash ?
> 
> Could we make it backward compatible ?

Yes, i have just restored it

```bash
$ minikube completion

❌  Exiting due to MK_USAGE: Usage: minikube completion SHELL
```

ilya-zuyev:
Hey @zhijianli88. Thanks for your PR! Could you extend the PR description, adding more info why this PR important and how it improves the current behavior of `minikube completion`? Thanks!

zhijianli88:
> Hey @zhijianli88. Thanks for your PR! Could you extend the PR description, adding more info why this PR important and how it improves the current behavior of `minikube completion`? Thanks!

PR description updated, PTAL.

medyagh (approved):

medyagh (approved):

PR #11196, opened 2021-04-26
Exec Unit Testing, not find asset method, then add it.

lilongfeng0902:
E.g. execute Unit Testing on my windows, `minikube/cmd/minikube/cmd/config/config_test.go`,  it prompt like below:
image
image


lilongfeng0902:
image
image

it prompt ``Asset` redeclared in this block`, what should I do on my local system(windows).

lilongfeng0902:
Find from the error message,  `go-bindata -nomemcopy` may be the mechanism for described above.

PR #11197, opened 2021-04-26
slice declaration: nil better than empty

lilongfeng0902:
slice declaration：
`var fields []string` is better than `fields := []string{}`

### DIFF
--- a/cmd/minikube/cmd/config/config.go
+++ b/cmd/minikube/cmd/config/config.go
@@ -191,7 +191,7 @@ Configurable fields: ` + "\n\n" + configurableFields(),
 }
 
 func configurableFields() string {
-	fields := []string{}
+	var fields []string
 	for _, s := range settings {
 		fields = append(fields, " * "+s.name)
 	}
--- a/pkg/minikube/assets/vm_assets.go
+++ b/pkg/minikube/assets/vm_assets.go
@@ -21,8 +21,11 @@ import (
 	"fmt"
 	"html/template"
 	"io"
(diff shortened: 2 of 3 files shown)


medyagh:
@lilongfeng0902 do u mind explaining why this is better?

lilongfeng0902:
> @lilongfeng0902 do u mind explaining why this is better? and what problem are you trying to solve?

@medyagh, thamks for your review. This PR is just coding review, the point that is modified may be better. You may find losts of commit histories, the commits are related with [https://github.com/kubernetes/minikube/pull/11196](url). Now let me explain why this modification is better：
First, I compare the slice declaration with others, most programmers use the style which I proposal.
Second, the style of `fields := []string{}` is prompted `empty slice declaration via literal`, not mistake, just suggestion.
Third, the style of `fields := []string{}` may be allocated space, the style of `var fields []string` is just declared.
In summary, my explanation may not be reasonable, then I will make other efforts for  learning minikube. Thank you again.

lilongfeng0902:
@medyagh btw, you may find that thare are five commit histories, but there is only the last one related with PR. How could I deal with this situation?

ilya-zuyev:
Go compiler generates the same code for both options. IMO, it's only matter of style

lilongfeng0902:
> Go compiler generates the same code for both options. IMO, it's only matter of style

Thanks for your attention and review,  from your point and medyagh's doubt, this PR is unreasonable and no value, I will avoid this in the future. Thank you again.

PR #11217, opened 2021-04-27
warn about performance for certain versions of kubernetes

### LINKED ISSUE #10327: Add solution message for slowdown in k8s 1.20.3 and above.

sharifelgamal:
Fixes #10327 

Before:
```
$ minikube start --kubernetes-version=1.18.16
😄  minikube v1.19.0 on Darwin 11.2.3
✨  Automatically selected the docker driver. Other choices: hyperkit, ssh
🚜  Pulling base image ...
🐳  Preparing Kubernetes v1.18.16 on Docker 20.10.6 ...
    ▪ Configuring RBAC rules ...
```

After:
```
$ minikube start --kubernetes-version=1.18.16
😄  minikube v1.19.0 on Darwin 11.2.3
❗  Kubernetes 1.18.16 has a known performance issue on cluster startup. It might take 2 to 3 minutes for a cluster to start.
❗  For more information, see: https://github.com/kubernetes/kubeadm/issues/2395
✨  Automatically selected the docker driver. Other choices: hyperkit, ssh
🚜  Pulling base image ...
🐳  Preparing Kubernetes v1.18.16 on Docker 20.10.6 ...
    ▪ Configuring RBAC rules ...
```


cmd/minikube/cmd/start.go
```diff
@@ -77,6 +77,15 @@ var (
...
 	apiServerIPs     []net.IP
 	hostRe           = regexp.MustCompile(`^[^-][\w\.-]+$`)
+	slowK8sVersions  = map[string]struct{}{
+		"1.18.16": {},
+		"1.18.17": {},
+		"1.19.8":  {},
+		"1.19.9":  {},
+		"1.20.3":  {},
+		"1.20.4":  {},
+		"1.20.5":  {},
```
afbjorklund, line 87: 1.21.0

afbjorklund (changes requested):

cmd/minikube/cmd/start.go
sharifelgamal: added

afbjorklund (approved):

cmd/minikube/cmd/start.go
```diff
@@ -77,6 +77,16 @@ var (
 	apiServerNames   []string
 	apiServerIPs     []net.IP
 	hostRe           = regexp.MustCompile(`^[^-][\w\.-]+$`)
+	slowK8sVersions  = map[string]struct{}{
```
medyagh, line 80: how about moving this to a package like `reason` 
Similar to 
https://github.com/medyagh/minikube/blob/b45c782df392d90613d6e8bcc94e029cd13128bb/pkg/minikube/reason/match.go#L54
```
// MatchKnownIssue returns a known issue from an error on an OS
func MatchKnownIssue(r Kind, err error, goos string) *Kind {
```


if there is no issue it will return Nil

if there is an issue it will return a K8sIssueKind
```
type K8sIssueKind struct {
	IsssueSummar string // For example Causes signigicant slowdown in creating cluster

	// URL is a reference URL for more information
	URL string
	// Do not attempt to match this reason to a specific known issue
	NoMatch bool
}

```

cmd/minikube/cmd/start.go
```diff
@@ -1370,6 +1380,11 @@ func validateKubernetesVersion(old *config.ClusterConfig) {
 		exitIfNotForced(reason.KubernetesTooOld, "Kubernetes {{.version}} is not supported by this release of minikube", out.V{"version": nvs})
 	}
 
+	if _, ok := slowK8sVersions[nvs.String()]; ok {
```
medyagh, line 1383: if i:= reason.MatchK8sIssues(); i != nil {
out.Box ( there is a known issue with X k8s verison : Issue :%i.IssueSummary, for more information see i.IssueURL)"
}

cmd/minikube/cmd/start.go
```diff
@@ -77,6 +77,16 @@ var (
 	apiServerNames   []string
 	apiServerIPs     []net.IP
 	hostRe           = regexp.MustCompile(`^[^-][\w\.-]+$`)
+	slowK8sVersions  = map[string]struct{}{
+		"1.18.16": {},
```
medyagh, line 81: how did we get this list of versions,? is there a comment or a Url we could add for comments ?

medyagh (changes requested):

cmd/minikube/cmd/start.go
sharifelgamal: This feels like overengineering to me, unless we intend on abstract all possible issues that can stem from Kubernetes itself.

cmd/minikube/cmd/start.go
sharifelgamal: I started with 1.20.3, which is the version we know started this issue. 1.18.16 and 1.19.8 are its backports, then I just moved forward in time until present day, testing each release version manually to verify the issue was still present. There's not URL (other than the Kubernetes release page) to add here.

cmd/minikube/cmd/start.go
medyagh: for sake of maintainability of minikube thats the right thing to do, since we have a package for matching issues,
and will make our start.go not longer than currently is, it would make it easier for future contributors to see what is going on

sharifelgamal: makes sense, done

cmd/minikube/cmd/start.go
sharifelgamal: done.

pkg/minikube/reason/k8s.go
```diff
@@ -0,0 +1,51 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reason
+
+import "github.com/blang/semver"
+
+type K8sIssue struct {
+	VersionsAffected map[string]struct{}
```
medyagh, line 22: how about it just be a flag list ? 

```
VersionsAffected []string
```

pkg/minikube/reason/k8s.go
```diff
@@ -0,0 +1,51 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reason
+
+import "github.com/blang/semver"
+
+type K8sIssue struct {
```
medyagh, line 21: add comment for all public structs  and their feilds

pkg/minikube/reason/k8s.go
```diff
@@ -0,0 +1,51 @@
...
+			"1.18.16": {},
+			"1.18.17": {},
+			"1.19.8":  {},
+			"1.19.9":  {},
+			"1.20.3":  {},
+			"1.20.4":  {},
+			"1.20.5":  {},
+			"1.21.0":  {},
+		},
+		Suggestion: "Kubernetes {{.version}} has a known performance issue on cluster startup. It might take 2 to 3 minutes for a cluster to start.",
```
medyagh, line 39: replace {{.verison}} with something else since it cant be translated..

how about
"Due to upstream bug cluster start is significantly slower "

pkg/minikube/reason/k8s.go
```diff
@@ -0,0 +1,51 @@
...
+limitations under the License.
+*/
+
+package reason
+
+import "github.com/blang/semver"
+
+type K8sIssue struct {
+	VersionsAffected map[string]struct{}
+	Suggestion       string
```
medyagh, line 23: how about suggestion, we have "Description"

medyagh (changes requested):

pkg/minikube/reason/k8s.go
sharifelgamal: There's no reason this can't be translated, the upstream call from this calls `out.WarningT` which calls translate.

pkg/minikube/reason/k8s.go
sharifelgamal: done

pkg/minikube/reason/k8s.go
sharifelgamal: done

pkg/minikube/reason/k8s.go
sharifelgamal: yeah I was thinking the same thing, I wanted to be slick with the key lookup but looping is just fine and not that worse runtime wise.

cmd/minikube/cmd/start.go
```diff
@@ -1370,6 +1370,14 @@ func validateKubernetesVersion(old *config.ClusterConfig) {
 		exitIfNotForced(reason.KubernetesTooOld, "Kubernetes {{.version}} is not supported by this release of minikube", out.V{"version": nvs})
 	}
 
+	// If the version of Kubernetes has a known issue, print a warning out to the screen
+	if issue := reason.ProblematicK8sVersion(nvs); issue.Description != "" {
```
medyagh, line 1374: This should return a pointer. And check if not nil

pkg/minikube/reason/k8s.go
```diff
@@ -0,0 +1,57 @@
...
+			"1.20.4",
+			"1.20.5",
+			"1.21.0",
+		},
+		Description: "Kubernetes {{.version}} has a known performance issue on cluster startup. It might take 2 to 3 minutes for a cluster to start.",
+		URL:         "https://github.com/kubernetes/kubeadm/issues/2395",
+	},
+}
+
+func ProblematicK8sVersion(v semver.Version) K8sIssue {
```
medyagh, line 48: Add comment

medyagh: For public func

medyagh, line 48: Return pointer so if it's nil later be checked if nil

medyagh (changes requested):

cmd/minikube/cmd/start.go
sharifelgamal: done

pkg/minikube/reason/k8s.go
sharifelgamal: done

sharifelgamal: done

medyagh (approved):

medyagh:
Looks good to me @sharifelgamal  , Thank you for this PR

PR #115, opened 2016-05-25
Make sure we stop the localkube binary before starting another instance.

dlorenc:
This makes "minikube start" re-entrant. There are a few cases where calling start multiple times in a row (without a stop) causes trouble.

PR #11223, opened 2021-04-28
Add locks to download functions.

### LINKED ISSUE #11166: minikube start should wailt till "download-only" is finishing
we could use our lock package https://github.com/kubernetes/minikube/tree/master/pkg/util/lock

to put a lock when ```minikube start --download-only``` happens

and then when `minikube start` runs when this lock exists will Wait for it to finish up to a reasonable time

andriyDev:
fixes #11166.

This prevents multiple `minikube start --download-only` from repeatedly downloading the same files (and more importantly preventing each other from corrupting downloaded files).

Running `minikube start --download-only` two times at the same time:
- Before: both processes download and write files.
1. Process 1:
```
$ minikube start --download-only
😄  minikube v1.19.0 on Debian rodete
✨  Automatically selected the docker driver. Other choices: kvm2, ssh
🚜  Pulling base image ...
    > preloaded-images-k8s-v10-v1...: 491.71 MiB / 491.71 MiB  100.00% 21.66 Mi
    > kubeadm.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubectl.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubelet.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubeadm: 37.40 MiB / 37.40 MiB [-------------] 100.00% 12.38 MiB p/s 3.2s
    > kubectl: 38.37 MiB / 38.37 MiB [-------------] 100.00% 12.14 MiB p/s 3.4s
    > gcr.io/k8s-minikube/kicbase...: 360.59 MiB / 360.59 MiB  100.00% 11.70 Mi
    > kubelet: 108.73 MiB / 108.73 MiB [-----------] 100.00% 16.23 MiB p/s 6.9s
```
2. Process 2:
```
$ minikube start --download-only
😄  minikube v1.19.0 on Debian rodete
✨  Automatically selected the docker driver. Other choices: kvm2, ssh
🚜  Pulling base image ...
    > preloaded-images-k8s-v10-v1...: 491.71 MiB / 491.71 MiB  100.00% 22.65 Mi
    > kubelet.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubeadm.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubectl.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubeadm: 37.40 MiB / 37.40 MiB [--------------] 100.00% 9.14 MiB p/s 4.3s
    > kubectl: 38.37 MiB / 38.37 MiB [--------------] 100.00% 9.13 MiB p/s 4.4s
    > kubelet: 108.73 MiB / 108.73 MiB [-----------] 100.00% 13.40 MiB p/s 8.3s
✅  Download complete!
```
- After: one process downloads and writes files, other process blocks (writing a message to output) until other process finishes downloading and immediately skips downloading the file.
1. Process 1:
```
$ minikube start --download-only
😄  minikube v1.19.0 on Debian rodete
✨  Automatically selected the docker driver. Other choices: kvm2, ssh
🚜  Pulling base image ...
    > preloaded-images-k8s-v10-v1...: 491.71 MiB / 491.71 MiB  100.00% 21.68 Mi
    > gcr.io/k8s-minikube/kicbase...: 357.46 MiB / 357.47 MiB  100.00% 14.30 Mi
    > kubectl.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubelet.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubeadm.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubeadm: 37.40 MiB / 37.40 MiB [-------------] 100.00% 19.44 MiB p/s 2.1s
    > kubectl: 38.37 MiB / 38.37 MiB [-------------] 100.00% 17.28 MiB p/s 2.4s
    > kubelet: 108.73 MiB / 108.73 MiB [-----------] 100.00% 21.18 MiB p/s 5.3s
✅  Download complete!
```
2. Process 2 (Note process 2 did download some files, but there is no duplication).
```
$ minikube start --download-only
😄  minikube v1.19.0 on Debian rodete
✨  Automatically selected the docker driver. Other choices: kvm2, ssh
🚜  Pulling base image ...
⌛  Another minikube instance is downloading dependencies...
⌛  Another minikube instance is downloading dependencies...
⌛  Another minikube instance is downloading dependencies...
⌛  Another minikube instance is downloading dependencies...
⌛  Another minikube instance is downloading dependencies...
✅  Download complete!
```


medyagh (commented):
thank you for this PR, do u mind adding the full output of multiple minikube start --download-only before'after this PR to compare?

andriyDev:
Fixed, I originally thought there wasn't a good way to show it since one of them just blocks, but I realized you can just tell from the lack of duplication.

Also note there is no message for when it blocks. The lock does not provide the ability to check if the lock is held, only attempt to acquire it. A potential workaround was to print a message before hand and then erase it afterwards, but this results in some nonsense where multiple things try to download at once causing multiple lock messages to show up, and then it is unable to erase them. Or a progress bar will be rendered and the lock message permanently shows up to the right of the progress bar. Regardless, ugly UI stuff is not ideal.

pkg/minikube/node/cache.go
```diff
@@ -151,14 +146,8 @@ func beginDownloadKicBaseImage(g *errgroup.Group, cc *config.ClusterConfig, down
...
 
 			if driver.IsDocker(cc.Driver) {
-				if image.ExistsImageInDaemon(img) {
-					klog.Infof("%s exists in daemon, skipping pull", img)
-					finalImg = img
-					return nil
-				}
-
 				klog.Infof("Downloading %s to local daemon", img)
-				err = image.WriteImageToDaemon(img)
```
medyagh, line 161: how about only move the exact "Download" part of the logic into the download image and still keep the `WriteImageToDaemon` in the `image` package ? 

feel free to break down the WriteImageToDaemon into smaller part if the "download" part of it is not modular.

pkg/minikube/download/image.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+	}
+	// Else, pull it
+	return false
+}
+
+var checkImageExistsInCache = ImageExistsInCache
+
+// ImageExistsInDaemon if img exist in local docker daemon
+func ImageExistsInDaemon(img string) bool {
```
medyagh, line 68: any func not used outside a package should be private (start with lower case)

medyagh (changes requested):

medyagh:
is there a way to have a Spinner in the UI that tells users waiting on the other terminal to finish ? 
the spainner could simply say
``
🚜  Another minikube instance is downloading depdenencies ...(spinner library)
``

pkg/minikube/node/cache.go
andriyDev: That would be ideal, but the trouble is the lock. In order to properly lock the download, it needs to download and write while locked. Separating the download and the write would mean we'd need a function anyway to wrap both those using the lock (and moving the tests while we're at it) which just moves the problem. The download would also be returning an unresolved image. This moves the downloading behavior out of the download function (as in, the function wouldn't be blocked while downloading), which doesn't match the semantics of the rest of the package.

@medyagh We could move just the image writing stuff back to the image package and have the download functions directly call those writing functions. Thoughts? Or is that just unnecessary?

andriyDev:
> is there a way to have a Spinner in the UI that tells users waiting on the other terminal to finish ?
> the spainner could simply say
> `🚜 Another minikube instance is downloading depdenencies ...(spinner library)`

Oh I just realized how to do this! The mutex API has no way to check whether the lock is present, but we can run the lock in a goroutine and then select between it and a timer and once the timer expires print the wait message and continue waiting. I need to practice thinking with Go haha.

pkg/minikube/download/image.go
andriyDev: Done.

pkg/minikube/download/download_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+import (
+	"fmt"
+	"io/fs"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/go-logr/logr"
```
medyagh, line 27: comment here, why we needed a new dependency here

pkg/minikube/download/download_test.go
```diff
@@ -0,0 +1,190 @@
...
+type mockLogger struct {
+	downloads int
+	t         *testing.T
+}
+
+func (ml *mockLogger) Enabled() bool {
+	return true
+}
+func (ml *mockLogger) Info(msg string, keysAndValues ...interface{}) {
+	fmt.Println("hi - ", msg)
```
medyagh, line 41: mistake ?

pkg/minikube/download/download_test.go
```diff
@@ -0,0 +1,190 @@
...
+func (ml *mockLogger) WithName(name string) logr.Logger {
+	return ml
+}
+
+func TestBinaryDownloadPreventsMultipleDownload(t *testing.T) {
+	EnableMock(true)
+	defer EnableMock(false)
+	tlog := &mockLogger{downloads: 0, t: t}
+
+	klog.SetLogger(tlog)
```
medyagh, line 69: this seems to be setting it globally. and unit tests happen in parallel.

medyagh: either set it for an instance of klog or make this test not parallel. preferably the first case

pkg/minikube/download/download_test.go
andriyDev: Done.

pkg/minikube/download/download_test.go
andriyDev: No longer relevant.

pkg/minikube/download/download_test.go
andriyDev: Fixed in a different way! First I removed the dependence on klog by allowing to insert a "mocked" download implementation. Much cleaner and gives us much more flexibility in the future. I also reorganized download_test to force its tests to run in serial. Since each test file is compiled as its own binary, it is safe to modify the global state of download.go. Note other tests all set the download mock to the default create file mock, so these are all safe as well.

andriyDev:
Rebased onto master.

pkg/minikube/download/download_test.go
```diff
@@ -0,0 +1,167 @@
...
+	t.Run("BinaryDownloadPreventsMultipleDownload", testBinaryDownloadPreventsMultipleDownload)
+	t.Run("PreloadDownloadPreventsMultipleDownload", testPreloadDownloadPreventsMultipleDownload)
+	t.Run("ImageToCache", testImageToCache)
+	t.Run("ImageToDaemon", testImageToDaemon)
+}
+
+func testBinaryDownloadPreventsMultipleDownload(t *testing.T) {
+	downloads := 0
+	SetDownloadMock(func(src, dst string) error {
+		// Sleep for a second to assure locking must have occurred.
```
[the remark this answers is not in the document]
andriyDev: Done. Thanks for the suggesion, me!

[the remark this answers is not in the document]
andriyDev: Good work, me!

andriyDev:
Rebased onto master x2.

pkg/minikube/download/binary.go
```diff
@@ -46,13 +46,20 @@ func binaryWithChecksumURL(binaryName, version, osName, archName string) (string
...
 	if err != nil {
 		return "", err
 	}
 
-	if _, err := os.Stat(targetFilepath); err == nil {
+	releaser, err := lockDownload(targetLock)
+	if err != nil {
+		return "", err
+	}
+	defer releaser.Release()
```
medyagh, line 60: defer before returning

pkg/minikube/download/download.go
```diff
@@ -22,20 +22,30 @@ import (
...
 var (
-	mockMode = false
+	mockDownload func(src, dst string) error = nil
+	checkCache                               = os.Stat
 )
 
-// EnableMock allows tests to selectively enable if downloads are mocked
-func EnableMock(b bool) {
-	mockMode = b
+func CreateDstDownloadMock(src, dst string) error {
```
medyagh, line 41: add comment for all public funcs

pkg/minikube/download/download_test.go
```diff
@@ -0,0 +1,157 @@
...
+	return func(src, dst string) error {
+		// Sleep for 200ms to assure locking must have occurred.
+		time.Sleep(time.Millisecond * 200)
+		*downloadsCounter++
+		return CreateDstDownloadMock(src, dst)
+	}
+}
+
+func testBinaryDownloadPreventsMultipleDownload(t *testing.T) {
+	downloads := 0
```
medyagh, line 48: downloadNums ? or add a small comment

pkg/minikube/download/download_test.go
```diff
@@ -0,0 +1,157 @@
...
+		group.Done()
+	}
+
+	go dlCall()
+	go dlCall()
+
+	group.Wait()
+
+	if downloads != 1 {
+		t.Errorf("Wrong number of downloads occurred. Actual: %v, Expected: 1", downloads)
```
medyagh, line 73: Expected to download only 1 in parallel minikube runs but got ! ...

pkg/minikube/download/download_test.go
```diff
@@ -0,0 +1,157 @@
...
+	SetDownloadMock(mockSleepDownload(&downloads))
+
+	checkCache = func(file string) (fs.FileInfo, error) {
+		if downloads == 0 {
+			return nil, fmt.Errorf("some error")
+		}
+		return nil, nil
+	}
+	checkPreloadExists = func(k8sVersion, containerRuntime string, forcePreload ...bool) bool { return true }
+	compareChecksum = func(k8sVersion, containerRuntime, path string) error { return nil }
```
medyagh, line 88: consider adding a bonous unit test to check if SHA is not same it should re-download

pkg/minikube/download/download.go
```diff
@@ -22,20 +22,30 @@ import (
...
 	"github.com/pkg/errors"
 	"k8s.io/klog/v2"
 	"k8s.io/minikube/pkg/minikube/out"
+	"k8s.io/minikube/pkg/minikube/style"
+	"k8s.io/minikube/pkg/util/lock"
 )
 
 var (
-	mockMode = false
+	mockDownload func(src, dst string) error = nil
```
medyagh, line 37: make this public instead of setting

pkg/minikube/download/binary.go
andriyDev: Done.

pkg/minikube/download/download.go
andriyDev: Done.

pkg/minikube/download/download_test.go
andriyDev: Done.

pkg/minikube/download/download.go
andriyDev: Done.

pkg/minikube/download/download_test.go
andriyDev: This PR is already massive, I think that should be in its own PR (since it has little relevance to this change).

andriyDev:
@medyagh Tested on Windows - works exactly as expected!

medyagh:
sorry I had to merged this PR first https://github.com/kubernetes/minikube/pull/11063/files
but it affects your PR you would need to redo the download for podman as well

andriyDev:
Rebased onto master x3.

pkg/minikube/download/binary.go
```diff
@@ -46,13 +46,20 @@ func binaryWithChecksumURL(binaryName, version, osName, archName string) (string
...
+	targetLock := targetFilepath + ".lock"
 
 	url, err := binaryWithChecksumURL(binary, version, osName, archName)
 	if err != nil {
 		return "", err
 	}
 
-	if _, err := os.Stat(targetFilepath); err == nil {
+	releaser, err := lockDownload(targetLock)
+	defer releaser.Release()
```
ilya-zuyev, line 57: releaser can be nil. (Line #115 of `lockDownload`)

pkg/minikube/download/download.go
```diff
@@ -91,3 +99,32 @@ func withinUnitTest() bool {
...
+
+// lockDownload locks `file` if possible and returns a releaser that must be called to release the lock.
+func lockDownload(file string) (mutex.Releaser, error) {
+	type retPair struct {
+		mutex.Releaser
+		error
+	}
+	lockChannel := make(chan retPair)
+
+	go func() {
```
ilya-zuyev, line 111: Are we starting a separate goroutine to acquire the mutex only to be able to display `"Another minikube instance is downloading dependencies... "` message if the mutex is already locked? We still block forever on line 128.  May be it makes sense to simplify the code and get rid of extra goroutine, channel and select?

pkg/minikube/download/image.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+	}
+	// Else, pull it
+	return false
+}
+
+var checkImageExistsInDaemon = imageExistsInDaemon
+
+// ImageToCache writes img to the local cache directory
+func ImageToCache(img string) error {
```
ilya-zuyev, line 83: Why `ImageToCache` is a better name than `WriteImageToCache`? :)

pkg/minikube/download/binary.go
andriyDev: Fixed.

pkg/minikube/download/download.go
ilya-zuyev: got it. a comment about this internal mutex timeout would be great here, actually.  Regarding the message, may be just a couple of logs before/immediately after getting the mutex? something like `acquiring the fs lock..`/`acquired`?

pkg/minikube/download/image.go
ilya-zuyev: It looks weird if `download` part is changed during import -  for example, with dot imports (` import . ".../download" ` bad idea) or if the package is given a new name - `import q "../download"`. Also we have `download.Tag(string)` for example, which doesn't actually download any tags :) So I'd rename `download.Binary` instead :)
Not a big deal though.

pkg/minikube/download/download.go
medyagh: lets keep the message for the user so they don't wait without any explaintion, our users like to be updated when something is waiting longer than a few sceonds (exmaple the complain about kubeadm init)

medyagh: for Func Names not related to this PR I would suggest making follow up PRs to keep this PR as small as possible to review

pkg/minikube/download/image.go
```diff
@@ -0,0 +1,244 @@
...
```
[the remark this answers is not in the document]
andriyDev: Done.

pkg/minikube/download/image.go
andriyDev: I would say this is relatively minor issue. There are other places that already use the package name as a verb (ex. `delete.PossibleLeftOvers`), also renaming `download.Binary` to something meaningful is hard, since the best name for it `download.DownloadBinary` would trigger a stuttering lint. I would say the solution here if necessary would be to move/rename all non-conforming functions, but that's a separate PR.

medyagh:
@andriyDev lets do one more rebase and pull upstream to fix the build

andriyDev:
Rebased onto master x5. @medyagh

PR #11225, opened 2021-04-29
migrate log of pkg/minikube/service/service.go 

ariesliuwei:
migrate  logs of pkg/minikube/service/service.go  to structured logging


medyagh:

PR #11227, opened 2021-04-29
migrate some logs of to structured logging

ariesliuwei:
migrate some logs in pkg/provision/buildroot.go


medyagh:

PR #11230, opened 2021-04-29
Output GitHub issue message to stderr

spowelljr:
Created a `BoxedErr` func which does the same as `Boxed` but outputs to stderr instead.

### DIFF
--- a/pkg/minikube/out/out.go
+++ b/pkg/minikube/out/out.go
@@ -113,15 +113,30 @@ func Styled(st style.Enum, format string, a ...V) {
 	}
 }
 
-// Boxed writes a stylized and templated message in a box to stdout
-func Boxed(format string, a ...V) {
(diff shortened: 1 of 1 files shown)


pkg/minikube/out/out.go
```diff
@@ -113,17 +113,32 @@ func Styled(st style.Enum, format string, a ...V) {
...
 	if useColor {
 		box.Config.Color = "Red"
 	}
+	return str, box
+}
+
+// Boxed writes a stylized and templated message in a box to stdout
+func Boxed(format string, a ...V) {
+	str, box := boxedCommon(format, a...)
 	box.Println("", str)
```
medyagh, line 129: how come the normal box does not need range lines ?

spowelljr: Fixed

afbjorklund (approved):

PR #11247, opened 2021-05-01
Goroutine leak fix

hex0punk:
A goroutine leak could occur as a result of timeouts when the function `readByteWithTimeout` is called after a user calls `minikube dashboard`. This is not likely to result in in an issue, but general best practice should be to use buffered channels in the same function as the capacity is known. 

The function in question is the following file
https://github.com/kubernetes/minikube/blob/fe0a1774a6be272a255f6de1930fb81a82430e28/cmd/minikube/cmd/dashboard.go#L174-L195

The same functionality is replicated here: https://play.golang.org/p/L5PNk76iJga. Notice that one hanging goroutine results from the forced timeout

### DIFF
--- a/cmd/minikube/cmd/dashboard.go
+++ b/cmd/minikube/cmd/dashboard.go
@@ -172,8 +172,8 @@ func kubectlProxy(kubectlVersion string, contextName string) (*exec.Cmd, string,
 
 // readByteWithTimeout returns a byte from a reader or an indicator that a timeout has occurred.
 func readByteWithTimeout(r io.ByteReader, timeout time.Duration) (byte, bool, error) {
-	bc := make(chan byte)
-	ec := make(chan error)
+	bc := make(chan byte, 1)
+	ec := make(chan error, 1)
 	go func() {
 		b, err := r.ReadByte()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


hex0punk:
Could someone approve the next workflow steps?

ilya-zuyev (commented):
Thanks for the PR. Looks good to me, let's wait for tests.

Could you add a bit more details about the leak scenario? It's concurrency and nothing is obvious :) Is it only that we're blocked in reading goroutine forever if timeout happened or something else?

hex0punk:
> Thanks for the PR. Looks good to me, let's wait for tests.
> 
> Could you add a bit more details about the leak scenario? It's concurrency and nothing is obvious :) Is it only that we're blocked in reading goroutine forever if timeout happened or something else?

Yep, exactly. This is unlikely to cause a noticeable issue given the use case, so in this case, it is more of best practice. To add more detail:

Either send operation in this anonymous goroutine will block:

https://github.com/kubernetes/minikube/blob/2474931ab3b22cb399b8f392b6da74f039eb22c8/cmd/minikube/cmd/dashboard.go#L177-L186

To let's say that this send occurs:

https://github.com/kubernetes/minikube/blob/2474931ab3b22cb399b8f392b6da74f039eb22c8/cmd/minikube/cmd/dashboard.go#L182

Then, the only way that goroutine would be unblocked would be if the corresponding receive occurs here 

https://github.com/kubernetes/minikube/blob/2474931ab3b22cb399b8f392b6da74f039eb22c8/cmd/minikube/cmd/dashboard.go#L188

But if a timeout occurs  first this case would execute:

https://github.com/kubernetes/minikube/blob/2474931ab3b22cb399b8f392b6da74f039eb22c8/cmd/minikube/cmd/dashboard.go#L192-L193

and the function would return, leaving the goroutine hanging. I replicated the issue here: https://play.golang.org/p/L5PNk76iJga where I forced the timeout to occur first, and then I count the number of hanging goroutines.

ilya-zuyev (approved):

Issue #11249, opened 2021-05-01
Use gotestsum to generate minikube integration test JSON


afbjorklund:
The current logs all claim to have been run the same second, with duration in microseconds:

```
{"Time":"2021-05-01T17:13:07.828742425+02:00","Action":"run","Test":"TestFunctional/parallel/BuildImage"}
{"Time":"2021-05-01T17:13:07.8287592+02:00","Action":"pause","Test":"TestFunctional/parallel/BuildImage"}
{"Time":"2021-05-01T17:13:07.828763322+02:00","Action":"cont","Test":"TestFunctional/parallel/BuildImage"}
{"Time":"2021-05-01T17:13:07.828858375+02:00","Action":"pass","Test":"TestFunctional/parallel/BuildImage","Elapsed":0.7}
```

In reality, the tests had rather different timings which can be seen when reading the live output:

```
{"Time":"2021-05-01T17:13:53.877285629+02:00","Action":"run","Test":"TestFunctional/parallel/BuildImage"}
{"Time":"2021-05-01T17:13:53.877327628+02:00","Action":"pause","Test":"TestFunctional/parallel/BuildImage"}
{"Time":"2021-05-01T17:13:53.877331911+02:00","Action":"cont","Test":"TestFunctional/parallel/BuildImage"}
{"Time":"2021-05-01T17:13:54.571565772+02:00","Action":"pass","Test":"TestFunctional/parallel/BuildImage","Elapsed":0.69}
```

The root cause is misusing the `test2json` tool.

Inserting timestamps assumes a _live_ test stream.

afbjorklund:
ping @medyagh

afbjorklund:
This tool looks interesting: https://github.com/gotestyourself/gotestsum

afbjorklund:
Besides fixing JSON, it also looks nicer...

Before: https://asciinema.org/a/411101
After: https://asciinema.org/a/411102

medyagh:
> Besides fixing JSON, it also looks nicer...
> (when just running interactively, that is)
> 
> Before: https://asciinema.org/a/411101
> After: https://asciinema.org/a/411102
> 
> The old formats are still available as well.
> 
> * `standard-quiet` - the standard go test format.
> * `standard-verbose` - the standard go test -v format.
> 
> The JSON also goes into the `--jsonfile`

Could it be used for both unit and integration test?

medyagh:
[...]

PR #11250, opened 2021-05-02
Run test2json during the test and not afterwards

afbjorklund:
Running it after the test is completed, means that the timestamps
of the output will be wrong (not reflecting the actual test run).

This should use "go test -json" instead of the "test2json" pipe,
but need something like "gotestsum" to show both out and json...

Issue #11249

----

This is only a beginning of the cleanup, haven't touched GitHub Actions.
This half-way approach will make _all_ output show only after test is done.

The plan is to use another tool like "gotestsum", to show both at once...
When using "go test", it can show text or it can show json. But not both.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -339,6 +339,7 @@ integration-functional-only: out/minikube$(IS_EXE) ## Trigger only functioanl te
 
 .PHONY: html_report
 html_report: ## Generate HTML  report out of the last ran integration test logs.
+	@#TODO the json needs to be recorded when the go test is run, for timestamps to work properly
 	@go tool test2json -t < "./out/testout_$(COMMIT_SHORT).txt" > "./out/testout_$(COMMIT_SHORT).json"
 	@gopogh -in "./out/testout_$(COMMIT_SHORT).json" -out ./out/testout_$(COMMIT_SHORT).html -name "$(shell git rev-parse --abbrev-ref HEAD)" -pr "" -repo github.com/kubernetes/minikube/  -details "${COMMIT_SHORT}"
 	@echo "-------------------------- Open HTML Report in Browser: ---------------------------"
--- a/hack/jenkins/common.sh
+++ b/hack/jenkins/common.sh
@@ -316,6 +316,10 @@ if test -f "${TEST_OUT}"; then
   rm "${TEST_OUT}" || true # clean up previous runs of same build
 fi
 touch "${TEST_OUT}"
+if test -f "${JSON_OUT}"; then
+  rm "${JSON_OUT}" || true # clean up previous runs of same build
+fi
+touch "${JSON_OUT}"
 
 if [ ! -z "${CONTAINER_RUNTIME}" ]
 then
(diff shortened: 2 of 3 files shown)


hack/jenkins/common.sh
```diff
@@ -354,17 +355,6 @@ echo ">> Copying ${TEST_OUT} to gs://${JOB_GCS_BUCKET}out.txt"
 gsutil -qm cp "${TEST_OUT}" "gs://${JOB_GCS_BUCKET}out.txt"
 
 
-echo ">> Attmpting to convert test logs to json"
-if test -f "${JSON_OUT}"; then
-  rm "${JSON_OUT}" || true # clean up previous runs of same build
-fi
-
-touch "${JSON_OUT}"
```
medyagh, line 362: How about move this touch to above. I dont know why I had to do this, But I remember it was needed for some odd situations...

medyagh (changes requested):

hack/jenkins/common.sh
afbjorklund: Now that it is using a file redirect, the file is _always_ created. But note also that "" is not legal JSON (but file is a json stream)

Makefile
```diff
@@ -339,7 +339,8 @@ integration-functional-only: out/minikube$(IS_EXE) ## Trigger only functioanl te
 
 .PHONY: html_report
 html_report: ## Generate HTML  report out of the last ran integration test logs.
-	@go tool test2json -t < "./out/testout_$(COMMIT_SHORT).txt" > "./out/testout_$(COMMIT_SHORT).json"
+	@#TODO the json needs to be recorded when the go test is run, for timestamps to work properly
+	@go tool test2json < "./out/testout_$(COMMIT_SHORT).txt" > "./out/testout_$(COMMIT_SHORT).json"
```
afbjorklund, line 343: There was a side-effect of this, removing the timestamp also removes "Elapsed" (being parsed from the output)

```
{"Action":"run","Test":"TestFunctional/parallel/BuildImage"}
{"Action":"pause","Test":"TestFunctional/parallel/BuildImage"}
{"Action":"cont","Test":"TestFunctional/parallel/BuildImage"}
{"Action":"pass","Test":"TestFunctional/parallel/BuildImage"}
```

```
{"Time":"2021-05-03T08:00:09.027479789+02:00","Action":"run","Test":"TestFunctional/parallel/BuildImage"}
{"Time":"2021-05-03T08:00:09.027542553+02:00","Action":"pause","Test":"TestFunctional/parallel/BuildImage"}
{"Time":"2021-05-03T08:00:09.027553681+02:00","Action":"cont","Test":"TestFunctional/parallel/BuildImage"}
{"Time":"2021-05-03T08:00:09.027858066+02:00","Action":"pass","Test":"TestFunctional/parallel/BuildImage","Elapsed":0.69}

```

hack/jenkins/common.sh
afbjorklund: Actually it was always doing a redirect before, so the file should always have been created.

```
> "${JSON_OUT}" 
```

afbjorklund: Can leave it in, but I don't think it has any effect.

Makefile
```diff
@@ -339,6 +339,7 @@ integration-functional-only: out/minikube$(IS_EXE) ## Trigger only functioanl te
 
 .PHONY: html_report
 html_report: ## Generate HTML  report out of the last ran integration test logs.
+	@#TODO the json needs to be recorded when the go test is run, for timestamps to work properly
```
medyagh, line 342: @afbjorklund is this todo still a todo ? (After this pr)

afbjorklund: Indeed, you can't run test2json after the test without broken timestamps. But at least the "Elapsed" is back again

afbjorklund: The complication was that the -json needs to go into all the _other_ targets, the ones that generate the .txt

PR #11273, opened 2021-05-04
Revert "Run test2json during the test and not afterwards"

sharifelgamal:
Reverts kubernetes/minikube#11250

Losing streaming logs in jenkins isn't worth fixing timestamps, we should probably use something like [gotestsum](https://github.com/gotestyourself/gotestsum) to do this properly


medyagh (approved):

afbjorklund:
Agreed, I mostly did some code for the discussion - should have marked it "WIP"

Issue #10673, opened 2021-03-01
add podman functional tests to gh action 


medyagh:

ilya-zuyev:
in progress

Issue #11293, opened 2021-05-05
TestFunctional/parallel/MountCmd fails in GitHub actions with podman driver


ilya-zuyev:
Mount comand fails:

```
        I0505 04:52:14.911379   21117 out.go:170] * Mounting host path /tmp/mounttest079057694 into VM as /mount-9p ...
        I0505 04:52:14.912059   21117 out.go:170]   - Mount type:   
        I0505 04:52:14.912443   21117 out.go:170]   - User ID:      docker
        I0505 04:52:14.912896   21117 out.go:170]   - Group ID:     docker
        I0505 04:52:14.913952   21117 out.go:170]   - Version:      9p2000.L
        I0505 04:52:14.914529   21117 out.go:170]   - Message Size: 262144
        I0505 04:52:14.915176   21117 out.go:170]   - Permissions:  755 (-rwxr-xr-x)
        I0505 04:52:14.915610   21117 out.go:170]   - Options:      map[]
        I0505 04:52:14.916120   21117 out.go:170]   - **Bind Address: <nil>:43651**
...
       I0505 04:52:15.388921   21117 ssh_runner.go:149] Run: /bin/bash -c "sudo mount -t 9p -o dfltgid=$(grep ^docker: /etc/group | cut -d: -f3),dfltuid=$(id -u docker),msize=262144,port=43651,trans=tcp,version=9p2000.L <nil> /mount-9p"
        I0505 04:52:15.407716   21117 out.go:170] 
        W0505 04:52:15.407873   21117 out.go:235] X Exiting due to GUEST_MOUNT: mount with cmd /bin/bash -c "sudo mount -t 9p -o dfltgid=$(grep ^docker: /etc/group | cut -d: -f3),dfltuid=$(id -u docker),msize=262144,port=43651,trans=tcp,version=9p2000.L <nil> /mount-9p" : /bin/bash -c "sudo mount -t 9p -o dfltgid=$(grep ^docker: /etc/group | cut -d: -f3),dfltuid=$(id -u docker),msize=262144,port=43651,trans=tcp,version=9p2000.L <nil> /mount-9p": Process exited with status 1
```

Please not, the `Bind Address` has "nil" host part. 
Comes from the result of `oci.containerGatewayIP` which runs `podman container inspect --format {{.NetworkSettings.Gateway}}`, and `inspect` returns 
```
{
 ...
 NetworkSettings": {
      "EndpointID": "",
      "Gateway": "",
      "IPAddress": "",
[...]

PR #11295, opened 2021-05-05
Add a functional test for podman

ilya-zuyev:
Fixes https://github.com/kubernetes/minikube/issues/10673

TestMountCmd is failing and is skipped for now for podman/GitHub env, see https://github.com/kubernetes/minikube/issues/11293

### DIFF
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -46,7 +46,7 @@ import (
 	"k8s.io/minikube/pkg/util/retry"
 
 	"github.com/elazarl/goproxy"
-	retryablehttp "github.com/hashicorp/go-retryablehttp"
+	"github.com/hashicorp/go-retryablehttp"
 	"github.com/otiai10/copy"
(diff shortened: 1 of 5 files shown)


test/integration/functional_test.go
```diff
@@ -399,8 +400,6 @@ func validateListImages(ctx context.Context, t *testing.T, profile string) {
 	}
 }
 
-// check functionality of minikube after evaling docker-env
-// TODO: Add validatePodmanEnv for crio runtime: #10231
 func validateDockerEnv(ctx context.Context, t *testing.T, profile string) {
```
medyagh, line 403: bring comment back for valdiateDockerEnv (only remove TODO)

medyagh (changes requested):

test/integration/functional_test.go
ilya-zuyev: done

PR #11300, opened 2021-05-05
Fix getting of gateway IP for podman driver

ilya-zuyev:
Fixes https://github.com/kubernetes/minikube/issues/11293

### DIFF
--- a/pkg/drivers/kic/oci/network.go
+++ b/pkg/drivers/kic/oci/network.go
@@ -92,9 +92,8 @@ func gatewayIP(ociBin, containerName string) (string, error) {
 		if gatewayIP := strings.TrimSpace(rr.Stdout.String()); gatewayIP != "" {
 			return gatewayIP, nil
 		}
-		klog.Infof("Couldn't find gateway for container %s", containerName)
 	}
-
+	klog.Infof("Couldn't find gateway for container %s", containerName)
 	return "", nil
 }
 
--- a/pkg/drivers/kic/oci/network.go
+++ b/pkg/drivers/kic/oci/network.go
@@ -87,7 +87,7 @@ func gatewayIP(ociBin, containerName, defNetwork string) (string, error) {
 
 	// https://github.com/kubernetes/minikube/issues/11293
 	// need to check nested network
-	format := fmt.Sprintf("{{(index .NetworkSettings.Networks %s).Gateway}}", containerName)
+	format := fmt.Sprintf("{{(index .NetworkSettings.Networks %q).Gateway}}", containerName)
 	rr, err = runCmd(exec.Command(ociBin, "container", "inspect", "--format", format, containerName))
 	if err != nil {
 		return "", errors.Wrapf(err, "inspect gateway")
@@ -96,7 +96,7 @@ func gatewayIP(ociBin, containerName, defNetwork string) (string, error) {
 		return gatewayIP, nil
 	}
 
-	format = fmt.Sprintf("{{(index .NetworkSettings.Networks %s).Gateway}}", defNetwork)
+	format = fmt.Sprintf("{{(index .NetworkSettings.Networks %q).Gateway}}", defNetwork)
 	rr, err = runCmd(exec.Command(ociBin, "container", "inspect", "--format", format, containerName))
 	if err != nil {
 		return "", errors.Wrapf(err, "inspect gateway")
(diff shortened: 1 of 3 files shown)


pkg/drivers/kic/oci/network.go
```diff
@@ -142,7 +177,7 @@ func podmanContainerIP(ociBin string, name string) (string, string, error) {
 		return "", "", errors.Wrapf(err, "podman inspect ip %s", name)
 	}
 	output := strings.TrimSpace(rr.Stdout.String())
-	if err == nil && output == "" { // podman returns empty for 127.0.0.1
```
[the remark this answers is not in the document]
ilya-zuyev: because it's always nil, we checked it on line 176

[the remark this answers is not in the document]
medyagh: makes sense thanks

pkg/drivers/kic/oci/network.go
```diff
@@ -69,14 +69,49 @@ func digDNS(ociBin, containerName, dns string) (net.IP, error) {
...
+	}
+	if gatewayIP := strings.TrimSpace(rr.Stdout.String()); gatewayIP != "" {
+		return gatewayIP, nil
+	}
+
+	// https://github.com/kubernetes/minikube/issues/11293
+	// need to check nested network
+	// check .NetworkSettings.Networks["cluster-name"].Gateway and then
+	// .NetworkSettings.Networks["bridge"|"podman"].Gateway
+	for _, network := range []string{containerName, defaultBridgeName(ociBin)} {
```
medyagh, line 86: in this for loop,  we have two things to iterate 
containerName and defaultBridgeName(ociBin)
will the second iteration ever happen ? (since it is returning error if there is nothing for the first one?

pkg/drivers/kic/oci/network.go
```diff
@@ -69,14 +69,49 @@ func digDNS(ociBin, containerName, dns string) (net.IP, error) {
...
+		gatewayIP, err := networkGateway(ociBin, containerName, network)
+		if err != nil {
+			return "", err
+		}
+		if gatewayIP != "" {
+			return gatewayIP, nil
+		}
+	}
+	klog.Infof("Couldn't find gateway for container %s", containerName)
+	return "", nil
```
medyagh, line 96: shouldn't this be returning error ? instead of nil ?
how about we define an nil error before the `for loop`
```
        var gateErr error
	for _, network := range []string{containerName, defaultBridgeName(ociBin)} {....
	       if err != nil {
	      	       gateErr = err 
                }

	
	
	...}

return "", gateErr
```

medyagh (changes requested):

pkg/drivers/kic/oci/network.go
ilya-zuyev: It returns an error, if `err != nil`. if `err` is not `nil' `gatewayIP` still can be empty

medyagh: so in what case the Second iteration will run ?
```
for _, network := range []string{containerName, defaultBridgeName(ociBin)}
```
first iteration  - containerName
second iteration - defaultBridgeName

as far I see as soon on the first call we Return error or Nil so it will not get to the second iteration, will it ?

medyagh (changes requested):

pkg/drivers/kic/oci/network.go
ilya-zuyev: non-nil err on line 74 means `docker container inspect` is broken. if we just don't  have the gateway ip for the specified network, err is nil. Both `If`s on lines 75 and 78 are false and we go to the second interation

medyagh: ok so lets add a comment here something like so next person reading this code dont have to decrypt it

// first trying to get .... then if there is no err and also the gateway IP is "" we will try the default brdige name....

medyagh (changes requested):

medyagh:
Please check the failed test
https://storage.googleapis.com/minikube-builds/logs/11300/a1ae930/Docker_Linux.html#fail_TestKicExistingNetwork

pkg/drivers/kic/oci/network.go
ilya-zuyev: done!

medyagh (approved):
Thank you @ilya-zuyev

Issue #11296, opened 2021-05-05
`minikube cp` should support copying from minkube to user's host


medyagh:
currently minikube cp does NOT support doing this.

```
minikube cp minikube:/var/lib/minikube/certs/etcd/peer.crt /tmp
```

current workaround by @tstromberg 
```
minikube ssh "cat /var/lib/minikube/certs/etcd/peer.crt" > /tmp/peer.cert
```

afbjorklund:
This needs support in the Asset API, which currently only has `Copy`.

vishjain:
@daehyeok Are you working on this?

daehyeok:
@vishjain yes. i will make PR after little clean up code

TairanLee:
Hi, now that minikube has `minikube cp`, but it seems it's underlying mechanism is still `cat >`?  
I'm asking this because I found a strange `cat` process, and the `minikube cp` running at low efficiency.  
If it can be done through scp or rsync, I think it would greatly improve the speed of copying.  
Thanks!

PR #11304, opened 2021-05-06
Implement target node option for cp command

### LINKED ISSUE #11146: minikube cp missing -n
minikube cp doesn't work when you have multiple nodes. There is no way to specify which node to copy to.

daehyeok:
fixes #11146

before
```
Copy the specified file into minikube, it will be saved at path <target file absolute path> in your minikube.
Example Command : "minikube cp a.txt /home/docker/b.txt"

Usage:
  minikube cp <source file path> <target file absolute path> [flags] [options]

```

after
```
Copy the specified file into minikube, it will be saved at path <target file absolute path> in your minikube.
Example Command : "minikube cp a.txt /home/docker/b.txt"
                  "minikube cp a.txt minikube-m02:/home/docker/b.txt"

Usage:
  minikube cp <source file path> <target node name>:<target file absolute path> [flags] [options]

```

### DIFF
--- a/cmd/minikube/cmd/cp.go
+++ b/cmd/minikube/cmd/cp.go
@@ -25,8 +25,11 @@ import (
 	"strings"
 
 	"k8s.io/minikube/pkg/minikube/assets"
+	"k8s.io/minikube/pkg/minikube/command"
 	"k8s.io/minikube/pkg/minikube/exit"
+	"k8s.io/minikube/pkg/minikube/machine"
 	"k8s.io/minikube/pkg/minikube/mustload"
+	"k8s.io/minikube/pkg/minikube/node"
 	"k8s.io/minikube/pkg/minikube/out"
 	"k8s.io/minikube/pkg/minikube/reason"
 )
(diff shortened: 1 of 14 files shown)


ilya-zuyev (changes requested):
':' can be a part of filepath
```
➜  minikube git:(master) ✗ out/minikube ssh -- sudo mkdir "/tmp/bar:"
➜  minikube git:(master) ✗ out/minikube cp foo /tmp/bar:/foo         
➜  minikube git:(master) ✗ out/minikube ssh -- ls '/tmp/bar:/'      
foo
```

```
➜  minikube git:(copy_node) ✗ out/minikube ssh -- sudo mkdir "/tmp/bar:"
➜  minikube git:(copy_node) ✗ out/minikube cp foo /tmp/bar:/foo    

X Exiting due to GUEST_NODE_RETRIEVE: Node /tmp/bar does not exist.
```
May be it's better to add a new cli key, `--target-node` or something like this

daehyeok:
@ilya-zuyev  oops. changed to extract node name only destPath is not a absolute path(not start with '/'). 
We only allowed absolute path for remote destination, so file with ':' character will be not a problem.
and if we use another flag for node name, we will need another flag for #11296 in future.

ilya-zuyev (approved):

PR #11309, opened 2021-05-06
docker driver: relax version constraint

AkihiroSuda:
When Docker (Moby) was installed from the source code, the version string is typically set to ["dev"](https://github.com/moby/moby/blob/v20.10.6/hack/make.sh#L42), or ["library-import"](https://github.com/moby/moby/blob/v20.10.6/dockerversion/version_lib.go#L10).

Previously, minikube could not start with these versions.

--- 
Before:
```console
$ minikube start --driver=docker
* minikube v1.20.0 on Ubuntu 21.04
* Using the docker driver based on user configuration

X Exiting due to PROVIDER_DOCKER_NOT_RUNNING: expected version format is "<year>.<month>.{patch}". but got dev
* Documentation: https://minikube.sigs.k8s.io/docs/drivers/docker/
```

After:
```console
$ minikube start --driver=docker
 ...
* minikube v1.20.0 on Ubuntu 21.04
! For improved Docker performance, Install the official release of Docker (Minimum recommended version is 18.09.0, current version is dev)
...
 ```

### DIFF
--- a/pkg/minikube/registry/drvs/docker/docker.go
+++ b/pkg/minikube/registry/drvs/docker/docker.go
@@ -87,7 +87,7 @@ func configure(cc config.ClusterConfig, n config.Node) (interface{}, error) {
 	}), nil
 }
 
-func status() registry.State {
+func status() (retState registry.State) {
 	_, err := exec.LookPath(oci.Docker)
 	if err != nil {
 		return registry.State{Error: err, Installed: false, Healthy: false, Fix: "Install Docker", Doc: docURL}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by medyagh** [comment]
@AkihiroSuda do u mind sharing the instructions to build docker from source ? 
and how about instead of relaxing the requirement, we detect the docker from source and add an info line to users that you are using Docker from Source and this is not Tested and use it as your own risk ?

**Comment by afbjorklund** [comment]
Thank you for this fix, we had a similar bug for podman (should have looked twice)

**Review by ilya-zuyev — COMMENTED** [review]
Thanks for the PR! Could you update the description with examples of output before and after  the change?

**Review by ilya-zuyev — CHANGES_REQUESTED** [review]
Please add a unit test for `checkDockerVersion`

**Comment by AkihiroSuda** [comment]
> I admitably I had never seen anyone build docker from source.

I'm a maintainer of Moby (aka dockerd), so I'm always using github.com/moby/moby@master

> do u mind sharing the instructions to build docker from source ?

```bash
git clone https://github.com/docker/docker.git
cd docker
make binary
make install
```

The version string is set to `dev`.

Alternatively, you can also install dockerd with `go build github.com/docker/docker/cmd/dockerd`. In this case, the version is set to `library-import`. Direct execution of `go build` is unrecommended but still valid.


> and how about instead of relaxing the requirement, we detect the docker from source and add an info line to users that you are using Docker from Source and this is not Tested and use it as your own risk ?

Returned to the caller as `{NeedsImprovement: true, Fix: "Install the official release of Docker (Minimum recommended version is 18.09.0, current version is dev)"}`


>  Could you update the description with examples of output before and after the change?

Done

> Please add a unit test for checkDockerVersion

Done

medyagh (approved):
Thank you @AkihiroSuda for improving the docker version check on minikube

medyagh:
look forward to see more PRs from you @AkihiroSuda

AkihiroSuda:
Thanks!

PR #11329, opened 2021-05-07
Simplify zsh completion code

baryluk:
Zsh is supported by Cobra for few years now.

No need for hacks (converting bash script to zsh).

The cobra's zsh completion relays on dynamically running minikube with `__complete` to learn completions. In my testing it is plenty fast (can't observe any delay, and manually running `./out/minikube __complete ...` for various combinations, it always finishes in less than 50ms).

I also manually verified that it works in zsh, and essentially all tested sub-commands and `--options` behave same way as in `bash` completions. There might be minor semantic differences (I couldn't find any tho), but it should be inconsequential, not guaranteed to be stable.

Also update the help text a little to be both more universal and "correct".


medyagh:
@baryluk thank you for this PR, do u mind pasting in the PR description the output Before and After this PR , I dont use zshell and there is no way for me to verify this, I would love to see ouput before and after this PR to help me review this PR 

agian thank you

baryluk:
> @baryluk thank you for this PR, do u mind pasting in the PR description the output Before and After this PR , ...

What output?

afbjorklund:
@baryluk : I get an error when testing this:

```console
 % ./out/minikube completion zsh | tee "${fpath[1]}/_minikube"
tee: /usr/local/share/zsh/site-functions/_minikube: Permission denied
```

You probably want to add a `sudo`, or use something like ~/.local ?

baryluk:
> @baryluk : I get an error when testing this:
> 
> ```
>  % ./out/minikube completion zsh | tee "${fpath[1]}/_minikube"
> tee: /usr/local/share/zsh/site-functions/_minikube: Permission denied
> ```
> 
> You probably want to add a `sudo`, or use something like ~/.local ?

Yes. Probably (I used sudo). I don't see any standard user-accessible location in zsh that is enabled by default. :/

afbjorklund:
> I don't see any standard user-accessible location in zsh that is enabled by default. :/

You can install it anywhere, but apparently you need to add it to `~/.zshrc` manually:

```
  fpath=(path/to/zsh-completions/src $fpath)
```

Here is what Docker is doing:

https://docs.docker.com/compose/completion/#zsh

afbjorklund:
In general we don't have _any_ instructions for doing a user-level (non-admin) installation of minikube.

`sudo install minikube-linux-amd64 /usr/local/bin/minikube`

So it is OK to use root also for installation of the completion, like `/etc/bash_completion.d/`

It would be better if it installed to `/etc` instead of `/usr/local/share`, but that's another story.

baryluk:
I will go back to rebasing this code soon. Thanks for the initial checks.

xaocon:
Could we get a static copy of a working zsh completion file until this is working?

medyagh:
@baryluk sorry for the delay in PR review, would u plz re-base your PR and also paste the output Before/After this PR ?

PR #11341, opened 2021-05-07
docs: fix sidebar current page highlight issue

dinever:
Fixes the issue where the sidebar doesn't support highlight on the selected page.

| Before                                                                                                          | After                                                                                                          |
|-----------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------|
| Before | After |

Thanks!

### DIFF
--- a/site/layouts/partials/sidebar-tree.html
+++ b/site/layouts/partials/sidebar-tree.html
@@ -85,7 +85,7 @@
           <!-- always show pages in the current section -->
           {{ if $activeSection }}
               {{ $showPage = true }}
-              {{ $activePage := eq . $p }}
+              {{ $activePage = eq . $p }}
           {{ end }}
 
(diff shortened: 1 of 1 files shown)


site/layouts/partials/sidebar-tree.html
```diff
@@ -85,7 +85,7 @@
           <!-- always show pages in the current section -->
           {{ if $activeSection }}
               {{ $showPage = true }}
-              {{ $activePage := eq . $p }}
+              {{ $activePage = eq . $p }}
```
dinever, line 88: Using `:=` shadows the outer scope `activePage` variable. So the outer scope variable will always be `false` before this fix.

medyagh (approved):

medyagh:
thank you @dinever

PR #11346, opened 2021-05-08
Load kicbase image from right cache and add log

### LINKED ISSUE #11321: The kicbase image downloads twice
There is a regression in minikube 1.20.0, that the kicbase image will download twice: once to cache, once to daemon

The workaround is to do minikube start --download-only (or maybe `docker pull`), or to just let it download

The fix was in PR ~~#11063~~

But it didn't make it to release

afbjorklund:
It was silently failing to load from the (wrong) cache directory,
causing the image to be download twice from the network instead.

```
🚜  Pulling base image ...
    > gcr.io/k8s-minikube/kicbase...: 358.10 MiB / 358.10 MiB  100.00% 15.00 Mi
    > gcr.io/k8s-minikube/kicbase...: 358.10 MiB / 358.10 MiB  100.00% 9.23 MiB
🔥  Creating docker container (CPUs=2, Memory=7900MB) ...
```

Also the name of the function was misleading.
It takes an image, and not actually a filename.

Closes #11321


medyagh:
> **kvm2 driver with docker runtime**
> 
> ```
> +----------------+----------+---------------------+
> |    COMMAND     | MINIKUBE | MINIKUBE (PR 11346) |
> +----------------+----------+---------------------+
> | minikube start | 52.7s    | 50.9s               |
> | enable ingress | 35.1s    | 38.8s               |
> +----------------+----------+---------------------+
> ```
> 
> **docker driver with docker runtime**
> 
> ```
> +----------------+----------+---------------------+
> |    COMMAND     | MINIKUBE | MINIKUBE (PR 11346) |
> +----------------+----------+---------------------+
> | minikube start | 21.5s    | 24.1s               |
> | enable ingress | 34.2s    | 32.8s               |
> +----------------+----------+---------------------+
> ```
> 
> **docker driver with containerd runtime**
> 
> ```
> +----------------+----------+---------------------+
> |    COMMAND     | MINIKUBE | MINIKUBE (PR 11346) |
> +----------------+----------+---------------------+
> | minikube start | 41.5s    | 46.1s               |
> | enable ingress |          |                     |
> +----------------+----------+---------------------+
> ```

@afbjorklund could this PR legit made minikube start slower ?

medyagh:
lets pull upstream after this PR is merged https://github.com/kubernetes/minikube/pull/11223

medyagh:
@afbjorklund other PR merged. Ready for pulling upstream

afbjorklund:
I made the existance check public again, because it looked assymetric in the mini-API.

For some reason we do download and load separate for the cache, but not for daemon ?
i.e. it is hidden away in the go-containerregistry abstraction, which just calls both of them

https://github.com/google/go-containerregistry/blob/main/pkg/v1/daemon/write.go#L38

andriyDev:
> @andriyDev
> I made the existance check public again, because it looked assymetric in the mini-API.
> 
> ```
> func ImageExistsInCache(img string) bool
> func ImageExistsInDaemon(img string) bool
> func ImageToCache(img string) error
> func ImageToDaemon(img string) error
> ```
> 
> For some reason we do download and load separate for the cache, but not for daemon ?
> i.e. it is hidden away in the go-containerregistry abstraction, which just calls both of them
> 
> https://github.com/google/go-containerregistry/blob/main/pkg/v1/daemon/write.go#L38
> 
> Possibly we should move the remaining function into "download" as well, but name is odd.
> 
> download.ImagePathInCache
> image.LoadFromTarball
> 
> Something like:
> 
> download.ImageFromCache (?)
> download.CacheToDaemon (?)
> 
> And we still have Docker hardcoded (as Daemon), so there is no support for Podman yet.
> 
> #8577 (~PR #11063~ ~PR #11356~)
> 
> daemon.Write
> 
> ```
> func ImageExistsInPodman(img string) bool
> func ImageToPodman(img string) error
> ```

How about instead of making ImageExistsInCache public, we make ImageExistsInDaemon private? Our only use case is in cache.go. We should move that call into download.CacheToDaemon.

I am a little concerned however on the semantics of "download.CacheToDaemon" - there's no download involved there, which is a big reason the ImageToCache/Daemon functions were moved from the image package.

I wonder if this solution is wrong. Why do we have ImageToDaemon? Perhaps we should force the only route to be to download to the cache, then load the tarball. Else we may end up having people download multiple images to their docker through restarts (maybe I'm misunderstanding that though).

afbjorklund:
> I wonder if our assumption is wrong. Why do we have ImageToDaemon? Perhaps we should force the only route to be to download to the cache, then load the tarball. Else we may end up having people download images to their docker multiple times through restarts (maybe I'm misunderstanding that though).

I opened a story for it: #11411

afbjorklund:
> How about instead of making ImageExistsInCache public, we make ImageExistsInDaemon private? Our only use case is in cache.go. We should move that call into download.CacheToDaemon.

I think it currently skips using that function, and instead always tries to load the (possibly non-existent) file and fails silently.

At some point it should give up on the image and the registry if they fail to download, and move on to the fallback images.

afbjorklund:
Having "daemon" in the API is not great either, but that was inherited from go-containerregistry

However, it is not portable to other engines. It even has a docker client, importing all of _that_ code.

github.com/docker/docker/client
```go
type imageOpener struct {
	ref      name.Reference
	buffered bool
	client   Client
}
```

And I guess saving the image to memory or tempfile is going to fail at some point, for any big images.

pkg/minikube/download/image.go
```diff
@@ -43,10 +43,16 @@ var (
...
-func imageExistsInCache(img string) bool {
+// imagePathInCache returns path in local cache directory
+func imagePathInCache(img string) string {
 	f := filepath.Join(constants.KICCacheDir, path.Base(img)+".tar")
 	f = localpath.SanitizeCacheDir(f)
+	return f
+}
+
+// ImageExistsInCache if img exist in local cache directory
+func ImageExistsInCache(img string) bool {
```
medyagh, line 54: I dont see this func be used outside `download` paockage any reason it is public ?
and also how about we move it to `image` package `image.ExistsInCache(img)` reads much better  than `download.ImageExistsInCache`

wdyt @afbjorklund

afbjorklund: https://github.com/kubernetes/minikube/pull/11346#issuecomment-841095916

medyagh:
@afbjorklund would u please (I tried to push to the branch but it didnt let me)

medyagh:
the containerd test failures seem to be flake

PR #11354, opened 2021-05-09
Use upgrade instead of install for the RPM installation

alenzen:
-U installs or upgrades the package currently installed to a newer version, so it might be preferable to -i.

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -46,7 +46,7 @@ sudo dpkg -i minikube_latest_amd64.deb
 
 ```shell
 curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-latest.x86_64.rpm
-sudo rpm -ivh minikube-latest.x86_64.rpm
+sudo rpm -Uvh minikube-latest.x86_64.rpm
 ```
 
 ### arm64 / aarch64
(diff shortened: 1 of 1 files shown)


afbjorklund:
Change sounds alright, but needs a better commit message and the CLA signed

alenzen:
I have successfully signed the Contributor License Agreement.

afbjorklund (approved):

afbjorklund:
For some reason we use "dpkg -i" but "rpm -Uvh" (with verbose and hashes and stuff). But I guess that is normal.

Sad that we cannot use "apt" and "yum" to manage the packages, but this change will make .rpm a little better...

BEFORE:

```console
# rpm -ivh minikube-1.20.0-0.x86_64.rpm 
Verifying...                          ################################# [100%]
Preparing...                          ################################# [100%]
	file /usr/bin/minikube from install of minikube-1.20.0-0.x86_64 conflicts with file from package minikube-1.19.0-0.x86_64

```

AFTER:

```console
# rpm -Uvh minikube-1.20.0-0.x86_64.rpm 
Verifying...                          ################################# [100%]
Preparing...                          ################################# [100%]
Updating / installing...
   1:minikube-1.20.0-0                ################################# [ 50%]
Cleaning up / removing...
   2:minikube-1.19.0-0                ################################# [100%]

```

PR #11355, opened 2021-05-09
Sort rejected drivers by their preference order

### LINKED ISSUE #10009: Rejected drivers not shown in priority order

afbjorklund:
Don't show drivers not being selected by default

Closes #10009

BEFORE

* Unable to pick a default driver. Here is what was considered, in preference order:
  - vmware: Not installed: exec: "docker-machine-driver-vmware": executable file not found in $PATH
  - docker: Not installed: exec: "docker": executable file not found in $PATH
  - kvm2: Not installed: exec: "virsh": executable file not found in $PATH
  - none: Not installed: exec: "iptables": executable file not found in $PATH
  - podman: Not installed: exec: "podman": executable file not found in $PATH
  - virtualbox: Not installed: unable to find VBoxManage in $PATH

AFTER

* Unable to pick a default driver. Here is what was considered, in preference order:
  - docker: Not installed: exec: "docker": executable file not found in $PATH
  - kvm2: Not installed: exec: "virsh": executable file not found in $PATH
  - vmware: Not installed: exec: "docker-machine-driver-vmware": executable file not found in $PATH
  - podman: Not installed: exec: "podman": executable file not found in $PATH
  - virtualbox: Not installed: unable to find VBoxManage in $PATH

The priority was being clobbered for unhealthy drivers, so save original priority as "preference"

### DIFF
--- a/pkg/minikube/registry/global.go
+++ b/pkg/minikube/registry/global.go
@@ -66,11 +66,16 @@ var (
 
 // DriverState is metadata relating to a driver and status
 type DriverState struct {
-	Name       string
-	Default    bool
+	// Name is the name of the driver used internally
+	Name string
+	// Default drivers are selected automatically
+	Default bool
+	// Preference is the original priority from driver
 	Preference Priority
-	Priority   Priority
-	State      State
+	// Priority is the effective priority with health
+	Priority Priority
+	// State is the state of driver and dependencies
+	State State
 	// Rejection is why we chose not to use this driver
 	Rejection string
 	// Suggestion is how the user could improve health
--- a/pkg/minikube/registry/global.go
+++ b/pkg/minikube/registry/global.go
@@ -66,10 +66,11 @@ var (
 
 // DriverState is metadata relating to a driver and status
 type DriverState struct {
-	Name     string
-	Default  bool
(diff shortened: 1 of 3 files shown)


medyagh (approved):

pkg/minikube/registry/global.go
```diff
@@ -66,10 +66,11 @@ var (
 
 // DriverState is metadata relating to a driver and status
 type DriverState struct {
-	Name     string
-	Default  bool
-	Priority Priority
-	State    State
+	Name       string
+	Default    bool
+	Preference Priority
```
medyagh, line 71: how about add a comment for what is Preference in this Struct
and also add a comment for Priority
and also a comment for Default 

so the new contributors have an easier time learning this part of the code ?

medyagh (changes requested):

pkg/minikube/registry/global.go
afbjorklund: Added comments

afbjorklund:
Note that this PR only changes the output to the user. The actual bug fix was in PR #11371

spowelljr (approved):

PR #11356, opened 2021-05-09
Revert "base-image: Allow using podman as well as docker daemon"

medyagh:
Reverts kubernetes/minikube#11063

LoadImage test is consistently failing all on GH action tests
  "TestFunctional/parallel/LoadImage" 
Could it be from this PR ?

https://github.com/kubernetes/minikube/pull/11063


afbjorklund (changes requested):
I don't think it is the kicbase changes that is causing the failures, but something from the other PRs. For instance it could be an issue with the canonical names, or a test framework issue ?

medyagh:
> I don't think it is the kicbase changes that is causing the failures, but something from the other PRs. For instance it could be an issue with the canonical names, or a test framework issue ?

Revering the pr seems to make all functional green (except containerd)

afbjorklund (approved):

afbjorklund:
There was a problem with the daemonBinary not being set in all code paths

PR #11366, opened 2021-05-10
`image load`: overwrite same image unless `--overwrite=false`

### LINKED ISSUE #11276: `minikube image load` not updating image if loading image with existing name:tag
```
% minikube start

% docker build -t benchmark-image -f Dockerfile.large .

% docker images --all
REPOSITORY                                TAG        IMAGE ID       CREATED          SIZE
benchmark-image                           latest     e066761bca06   4 minutes ago    937MB

% minikube image load benchmark-image:latest

% eval $(./out/minikube docker-env) && docker image ls
REPOSITORY                                TAG        IMAGE ID       CREATED          SIZE
benchmark-image                           latest     e066761bca06   4 minutes ago    937MB

% docker build -t benchmark-image -f Dockerfile.small .

% docker images --all
REPOSITORY                                TAG        IMAGE ID       CREATED          SIZE
benchmark-image                           latest     60c0c9285193   2 minutes ago    721MB

% minikube image load benchmark-image:latest

% eval $(./out/minikube docker-env) && docker image ls
REPOSITORY                                TAG        IMAGE ID       CREATED          SIZE
benchmark-image                           latest     e066761bca06   4 minutes ago    937MB
```

As you can see above, pushing an image with the same name:tag does not update within minikube.

spowelljr:
Closes #11276

Before:
If you had an image already loaded into minikube, and then created a new image with the same name:tag as one already loaded into minikube and ran `minikube image load` it would skip loading the image since it detected a matching name:tag, even though the image itself is different.

After:
`minikube image load` will load the image, even if an image with the existing name:tag is already loaded into minikube.

### DIFF
--- a/cmd/minikube/cmd/image.go
+++ b/cmd/minikube/cmd/image.go
@@ -137,7 +137,7 @@ var loadImageCmd = &cobra.Command{
 		} else if local {
 			// Load images from local files, without doing any caching or checks in container runtime
 			// This is similar to tarball.Image but it is done by the container runtime in the cluster.
-			if err := machine.DoLoadImages(args, []*config.Profile{profile}, "", false); err != nil {
+			if err := machine.DoLoadImages(args, []*config.Profile{profile}, "", overwrite); err != nil {
 				exit.Error(reason.GuestImageLoad, "Failed to load image", err)
 			}
 		}
--- a/cmd/minikube/cmd/image.go
+++ b/cmd/minikube/cmd/image.go
@@ -45,6 +45,7 @@ var (
 	pull       bool
 	imgDaemon  bool
 	imgRemote  bool
+	overWrite  bool
 	tag        string
 	push       bool
 	dockerFile string
@@ -130,7 +131,7 @@ var loadImageCmd = &cobra.Command{
 		if imgDaemon || imgRemote {
 			image.UseDaemon(imgDaemon)
 			image.UseRemote(imgRemote)
-			if err := machine.CacheAndLoadImages(args, []*config.Profile{profile}, true); err != nil {
+			if err := machine.CacheAndLoadImages(args, []*config.Profile{profile}, overWrite); err != nil {
 				exit.Error(reason.GuestImageLoad, "Failed to load image", err)
 			}
 		} else if local {
@@ -248,6 +249,7 @@ func init() {
 	loadImageCmd.Flags().BoolVarP(&pull, "pull", "", false, "Pull the remote image (no caching)")
 	loadImageCmd.Flags().BoolVar(&imgDaemon, "daemon", false, "Cache image from docker daemon")
 	loadImageCmd.Flags().BoolVar(&imgRemote, "remote", false, "Cache image from remote registry")
+	loadImageCmd.Flags().BoolVar(&overWrite, "over-write", true, "Over write the existing image if the name:tag of the images are the same")
 	imageCmd.AddCommand(loadImageCmd)
 	imageCmd.AddCommand(removeImageCmd)
 	buildImageCmd.Flags().StringVarP(&tag, "tag", "t", "", "Tag to apply to the new image (optional)")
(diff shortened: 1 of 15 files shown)


cmd/minikube/cmd/image.go
```diff
@@ -248,6 +249,7 @@ func init() {
 	loadImageCmd.Flags().BoolVarP(&pull, "pull", "", false, "Pull the remote image (no caching)")
 	loadImageCmd.Flags().BoolVar(&imgDaemon, "daemon", false, "Cache image from docker daemon")
 	loadImageCmd.Flags().BoolVar(&imgRemote, "remote", false, "Cache image from remote registry")
+	loadImageCmd.Flags().BoolVar(&overWrite, "over-write", true, "Over write the existing image if the name:tag of the images are the same")
```
medyagh, line 252: should this be overwrite ? hm...

medyagh (changes requested):

cmd/minikube/cmd/image.go
```diff
@@ -248,6 +249,7 @@ func init() {
 	loadImageCmd.Flags().BoolVarP(&pull, "pull", "", false, "Pull the remote image (no caching)")
 	loadImageCmd.Flags().BoolVar(&imgDaemon, "daemon", false, "Cache image from docker daemon")
 	loadImageCmd.Flags().BoolVar(&imgRemote, "remote", false, "Cache image from remote registry")
+	loadImageCmd.Flags().BoolVar(&overwrite, "overwrite", true, "Overwrite the existing image if the name:tag of the images are the same")
```
medyagh, line 252: overwrite image even same image:tag name exists.

medyagh (changes requested):

pkg/minikube/machine/cache_images.go
```diff
@@ -259,20 +264,26 @@ func DoLoadImages(images []string, profiles []*config.Profile, cacheDir string)
...
 // transferAndLoadImage transfers and loads a single image
-func transferAndLoadImage(cr command.Runner, k8s config.KubernetesConfig, src string) error {
+func transferAndLoadImage(cr command.Runner, k8s config.KubernetesConfig, src string, imgName string) error {
 	r, err := cruntime.New(cruntime.Config{Type: k8s.ContainerRuntime, Runner: cr})
 	if err != nil {
 		return errors.Wrap(err, "runtime")
 	}
+
+	if err := r.RemoveImage(imgName); err != nil {
+		klog.Warningf("remove image: %v", err)
```
medyagh, line 278: if image doesnt exist it will always warn ? in case the error is doesnt exist we could skip warn ?

pkg/minikube/image/cache.go
```diff
@@ -90,7 +90,7 @@ func SaveToDir(images []string, cacheDir string) error {
 }
 
 // saveToTarFile caches an image
-func saveToTarFile(iname, rawDest string) error {
+func saveToTarFile(iname, rawDest string, force ...bool) error {
```
medyagh, line 93: how about call it same, "overwrite" and then no need to make it optional, just pass it as required flag

medyagh (changes requested):

pkg/minikube/machine/cache_images.go
```diff
@@ -269,8 +269,15 @@ func transferAndLoadImage(cr command.Runner, k8s config.KubernetesConfig, src st
 		return errors.Wrap(err, "runtime")
 	}
 
-	if err := r.RemoveImage(imgName); err != nil {
-		klog.Warningf("remove image: %v", err)
+	exists, err := imageExists(r, imgName)
```
medyagh, line 272: I worry this add performance downside to it, how about instead of Image List, we do something cheaper ? try to remove it, and if the error message says it doesnt exist then only klog.Info if error message is not that klog.Warnning

or alternatively we could do

klog.Warniningf("failed to remove image (which might be okay if it doesn't exists")

medyagh (changes requested):

medyagh (approved):

PR #11369, opened 2021-05-11
Updated URL and remove zsh note

guettli:
The URL was outdated.
Removed the zsh note, too. I think it makes sense that it gets explained on the target page, not here.


afbjorklund (approved):

medyagh (approved):
Looks good to me @guettli  , Thank you for improving minikube's docs ,

PR #11371, opened 2021-05-11
Fix error picking non-default alternate drivers

afbjorklund:
We should not select drivers like "ssh" or "none",
unless explicitly requested by the user. Wrong var...

Fix small  bug from PR #10554
When fixing issue from PR #7389


ilya-zuyev (approved):
lgtm

medyagh (approved):

medyagh:
@afbjorklund needs a rebase plz

medyagh (approved):

PR #11387, opened 2021-05-12
Use the first non flag argument for the log file name instead of the first argument

### LINKED ISSUE #11382: Putting flag as first argument causes unintended log file names
Current:
```
$ minikube -p m1 status

Log file created is: /tmp/minikube_-p_<hash>.log
```

Expected:
```
$ minikube -p m1 status

Log file created is: /tmp/minikube_status_<hash>.log
```

vigothehacker:
fixes #11382


vigothehacker:
I signed it

vigothehacker:
I signed it

vigothehacker:
fixes #11382

cmd/minikube/main.go
```diff
@@ -205,7 +205,7 @@ func setFlags(parse bool) {
 		if pflag.CommandLine.Changed("log_dir") && pflag.Lookup("log_dir").Value.String() != "" {
 			dir = pflag.Lookup("log_dir").Value.String()
 		}
-		l := logFileName(dir, 0)
```
spowelljr, line 208: `pflag` is a global variable, so you can avoid having to pass it as an argument to `logFileName`.

spowelljr (changes requested):
Looks good, thanks for your contribution! Once you address the comment this looks good to merge.

cmd/minikube/main.go
```diff
@@ -205,7 +205,7 @@ func setFlags(parse bool) {
...
+		l := logFileName(dir, 0, pflag.Args())
```
spowelljr, line 208: `pflag` is a global variable, so you can avoid having to pass it as an argument to `logFileName`.

vigothehacker: Thanks for the review, I have made the changes

`nonFlagArgs := pflag.Args()`
`l := logFileName(dir, 0, nonFlagArgs)`
in the new commit.

Regards

cmd/minikube/main.go
```diff
@@ -141,7 +141,7 @@ func checkLogFileMaxSize(file string, maxSizeKB int64) bool {
 }
 
 // logFileName generates a default logfile name in the form minikube_<argv[1]>_<hash>_<count>.log from args
-func logFileName(dir string, logIdx int64) string {
+func logFileName(dir string, logIdx int64, nonFlagArgs []string) string {
```
medyagh, line 144: Thank you for fixing this @vigothehacker how about adding a unit test with examples and test cases?

medyagh (changes requested):

cmd/minikube/main.go
spowelljr: That's not exactly what I meant, this still work perfectly fine though.

What I'm trying to communicate is that because `pflag` is a global variable, it doesn't need to be passed into `logFileName`, be can accessed within because it's global.

So you can revert the signature back to `l := logFileName(dir, 0)`

And then before `for _, s := range nonFlagArgs {` you can do `nonFlagArgs := pflag.Args()`

Let me know if that makes sense.

vigothehacker: Hi. I have fixed this in the latest commits

test/integration/generated_logs_test.go
```diff
@@ -0,0 +1,173 @@
...
+		t.Run(test.command, func(t *testing.T) {
+
+			// before starting the test, ensure no other logs from the current command are written
+			logFiles, err := filepath.Glob(filepath.Join(logDir, fmt.Sprintf("minikube_%s*", test.command)))
+			if err != nil {
+				t.Errorf("failed to get old log files for command %s : %v", test.command, err)
+			}
+			cleanupLogFiles(t, logFiles)
+
+			args := []string{"-p", profile, "--log_dir", logDir, test.command}
```
vigothehacker, line 108: Hi , I have created a test file test/integration/generated_logs_test.go to verify the generated logs when running the minikube subcommands. Most of the functionality here was part of the file  test/integration/error_spam_test.go . Please let me know if this would be the correct approach for this. Or else should I make these changes part of the previous file

The changes in test are 

1) Remove all files that match `minikube_*` after the first start ( line 61)
2) Test a single run to see the proper filename getting generated before we run multiple runs
3) Test for #11382 by changing the args variable to include flags before the subcommand

cmd/minikube/main.go
vigothehacker: Implemented TestGeneratedLogs in tests. Please see my review as I had some doubts

test/integration/generated_logs_test.go
```diff
@@ -0,0 +1,173 @@
...
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+)
+
+// Verifies logs generated in the /tmp directory
+func TestGeneratedLogs(t *testing.T) {
```
medyagh, line 31: I am not keen on making a new cluster just for this test, unless there was a "blocking" reason not to test this with TestErrorSpam ?

medyagh (changes requested):

test/integration/generated_logs_test.go
vigothehacker: Moved the changes to the previous file

vigothehacker:
Hi, the PR is ready for test and merge

test/integration/error_spam_test.go
```diff
@@ -118,6 +118,12 @@ func TestErrorSpam(t *testing.T) {
 		}
 	}
 
+	logsToBeRemoved, err := filepath.Glob(filepath.Join(logDir, "minikube_*"))
+	if err != nil {
+		t.Error("failed to clean old log files", err)
+	}
+	cleanupLogFiles(t, logsToBeRemoved)
```
spowelljr, line 125: I don't see the need for removing all minikube logs when we delete the logs for each specific command below (line 170).

test/integration/error_spam_test.go
```diff
@@ -161,6 +169,18 @@ func TestErrorSpam(t *testing.T) {
 			}
 			cleanupLogFiles(t, logFiles)
 
+			singleRun, err := Run(t, exec.CommandContext(ctx, Target(), args...))
```
spowelljr, line 172: Curious what the reasoning is for doing a single run and then doing the rest after. If there any issues with just doing the full looped run below?

spowelljr (changes requested):

test/integration/error_spam_test.go
vigothehacker: Not a blocker but it will return the error faster in case logs are not getting generated for a single case instead of the command running 175 times and then giving the error. Removed the changes from the branch. Can take it up as a enhancement if needed

spowelljr (approved):

Issue #11335, opened 2021-05-07
fix TestNetworkPlugins/group/false/Start for docker/containerd


ilya-zuyev:
the test TestNetworkPlugins/group/false/Start times out for docker/containerd

https://storage.googleapis.com/minikube-builds/logs/master/c31bd57/Docker_Linux_containerd.html#fail_TestNetworkPlugins%2fgroup%2ffalse%2fStart

https://storage.googleapis.com/minikube-builds/logs/master/c31bd57/Docker_Linux_containerdout.txt

PR #11394, opened 2021-05-12
Skip network tests with disabled CNI for containerd/crio runtimes

ilya-zuyev:
Fixes https://github.com/kubernetes/minikube/issues/11335


test/integration/net_test.go
```diff
@@ -95,10 +99,13 @@ func TestNetworkPlugins(t *testing.T) {
...
 				if !t.Failed() {
 					t.Run("KubeletFlags", func(t *testing.T) {
-						// none does not support 'minikube ssh'
-						rr, err := Run(t, exec.CommandContext(ctx, Target(), "ssh", "-p", profile, "pgrep -a kubelet"))
+						var rr *RunResult
+						var err error
 						if NoneDriver() {
 							rr, err = Run(t, exec.CommandContext(ctx, "pgrep", "-a", "kubelet"))
+						} else {
+							// none does not support 'minikube ssh'
```
medyagh, line 107: This comment seems to be at wrong place

medyagh (changes requested):

test/integration/net_test.go
ilya-zuyev: moved

test/integration/net_test.go
```diff
@@ -70,6 +70,10 @@ func TestNetworkPlugins(t *testing.T) {
 					t.Skipf("flannel is not yet compatible with Docker driver: iptables v1.8.3 (legacy): Couldn't load target `CNI-x': No such file or directory")
 				}
 
+				if ContainerRuntime() != "docker" && tc.name == "false" {
```
medyagh, line 73: @ilya-zuyev 
in case of containerd , should even allow users to specify the cni to be false?

I tried on my machine the coreDNS will not come up 
```
NAMESPACE     NAME                               READY   STATUS              RESTARTS   AGE
kube-system   coredns-74ff55c5b-qjj8r            0/1     ContainerCreating   0          58s
kube-system   etcd-minikube                      1/1     Running             0          70s
kube-system   kube-apiserver-minikube            1/1     Running             0          70s
kube-system   kube-controller-manager-minikube   1/1     Running             0          70s
kube-system   kube-proxy-6spkq                   1/1     Running             0          58s
kube-system   kube-scheduler-minikube            1/1     Running             0          70s
kube-system   storage-provisioner                1/1     Running             1          71s
```

I think the the right thing is, instead of skipping the test inside minikube test itself we should check if user passed `--cni=false` with non-docker runtime, we should Exit with usage that if you really want this you need to pass --force otherwise coredns needs this and wont come up

medyagh (changes requested):

test/integration/net_test.go
medyagh: and then for this test instead of skipping for non-docker we should check if it Exited with Usage Exit code or not

cmd/minikube/cmd/start.go
```diff
@@ -1182,7 +1184,20 @@ func validateFlags(cmd *cobra.Command, drvName string) {
...
+}
 
+// if container runtime is not docker, check that cni is not disabled
+func validateCNI(cmd *cobra.Command, runtime string) {
+	if runtime == "docker" {
+		return
+	}
+	if cmd.Flags().Changed(cniFlag) && strings.ToLower(viper.GetString(cniFlag)) == "false" {
+		if viper.GetBool(force) {
+			out.WarnReason(reason.Usage, "The '{{.name}} container runtime requires CNI", out.V{"name": runtime})
```
medyagh, line 1196: We can add this message to be more clear “You have chosen to disable the CNI but ... (continue with your message)

test/integration/net_test.go
```diff
@@ -70,6 +70,10 @@ func TestNetworkPlugins(t *testing.T) {
...
+					t.Skipf("skipping the test as CNI is required for container runtime %s", ContainerRuntime())
```
medyagh, line 74: Here lets check in the test, for Exit Usage if chi=false and container runtime non-docker

medyagh (changes requested):

test/integration/net_test.go
```diff
@@ -32,6 +32,60 @@ import (
...
+				t.Errorf("Expected %q line not found in ouput %s", expectedMsg, rr.Output())
+			}
+		})
+	}
+}
+
+// TestCNIFalseForce checks that minikube returns not error
+// if container runtime is "containerd" or "crio"
+// and --cni=false, but --force=true
+func TestCNIFalseForce(t *testing.T) {
```
medyagh, line 65: instead of creating a new test, lets add this as part of the same SubTest that we have `--cni=false`
```
			{"false", []string{"--cni=false"}, "", "", false},

```



in which we only expect exit Usage code if --cni=false and container runtime is not Docker and in case of docker it continue same test as before

```
if strings.Contains(tc.name, "false") { 
..... if Non Docker ...
          Expect Exit Usage
If Docker (continue the test as it is...now)


}



```

medyagh (changes requested):

test/integration/net_test.go
ilya-zuyev: Not sure, actually. `TestNetworkPlugins` is focused on how network is configured.  In`TestCNIFalseForce` we mostly check how the command line arguments are validated. These tests check different parts of minikube code. Also,`TestNetworkPlugins` is already too big, overloaded with checks and hard to understand. Adding extra 20 lines wouldn't make it better.

medyagh: --force is an option we discourage users form using, and it means you are at your own risk and not supported, we do not have tests for --force and I am certainly okay for skipping testing if --force gives a warning message to users if it comes at the cost of starting a new minikube. our integration test machines are limitted specially on expensive options like hyperkit and mac and we can not add a new minikube start for this option.

lets move it under Cni tests as a subtest

ilya-zuyev:
ok, all affected integration tests are passing now for containerd and cri-o

medyagh (approved):
thank you very much @ilya-zuyev we are only 2 steps to green containerd !

PR #11399, opened 2021-05-13
Implement a quiz-style installation instructions

### LINKED ISSUE #10353: Make the getting started / installation page look nice again
After adding architectures to the "Installation" section, it's an even bigger mess than it was before...

Now it separates both on OS and on CPU, and it can be confusing to know what to actually download:

**(1) Installation**
* Linux
  * x86
    * Binary download
    * Debian package
    * RPM package 
  * ARM
    * Binary download
    * Debian package
    * RPM package 
* macOS
  * Brew package manager
  * x86
    * Binary download
  * ARM
    * Binary download
* Windows
  * Windows Package Manager
  * Chocolatey
  * Stand-alone Windows Installer

Hopefully there is some nice web design that could be done, especially for the Linux tab ?

minikube-installation

I described three different workarounds (in text), but am looking for a better looking solution.

dinever:
Hi,

This PR fixes #10353, see https://github.com/kubernetes/minikube/issues/10353#issuecomment-835701332

The idea is to add a "quiz" style installation page. Inspired by CUDA and PyTorch:

| CUDA                                                                                                                                          	| PyTorch                                                                                                                                       	|
|-----------------------------------------------------------------------------------------------------------------------------------------------	|-----------------------------------------------------------------------------------------------------------------------------------------------	|
| Screen Shot 2021-05-13 at 2 23 14 AM 	| Screen Shot 2021-05-13 at 2 23 23 AM 	|

Showcase:

quiz

Some features:

- [x] Since this style is more scalable, I've added `ARMv7`, `ppc64` and `S390x` architecture under Linux as well
- [x] If a row has only one option, it's automatically selected. e.g. Windows -> x86_64
    auto-select
- [x] The first row (OS) is automatically selected based on the user's OS

Would love to hear some feedback! Thanks!

### DIFF
--- a/hack/jenkins/release_build_and_upload.sh
+++ b/hack/jenkins/release_build_and_upload.sh
@@ -90,8 +90,14 @@ make checksum
 # unversioned names to avoid updating upstream Kubernetes documentation each release
 cp "out/minikube_${DEB_VERSION}-0_amd64.deb" out/minikube_latest_amd64.deb
 cp "out/minikube_${DEB_VERSION}-0_arm64.deb" out/minikube_latest_arm64.deb
+cp "out/minikube-${DEB_VERSION}-0_armhf.deb" out/minikube_latest_armhf.deb
+cp "out/minikube-${DEB_VERSION}-0_ppc64el.deb" out/minikube_latest_ppc64el.deb
+cp "out/minikube-${DEB_VERSION}-0_s390x.deb" out/minikube_latest_s390x.deb
 cp "out/minikube-${RPM_VERSION}-0.x86_64.rpm" out/minikube-latest.x86_64.rpm
 cp "out/minikube-${RPM_VERSION}-0.aarch64.rpm" out/minikube-latest.aarch64.rpm
+cp "out/minikube-${RPM_VERSION}-0.armv7hl.rpm" out/minikube-latest.armv7hl.rpm
+cp "out/minikube-${RPM_VERSION}-0.ppc64le.rpm" out/minikube-latest.ppc64le.rpm
+cp "out/minikube-${RPM_VERSION}-0.s390x.rpm" out/minikube-latest.s390x.rpm
 
 gsutil -m cp out/* "gs://$BUCKET/releases/$TAGNAME/"
 
--- a/hack/jenkins/release_build_and_upload.sh
+++ b/hack/jenkins/release_build_and_upload.sh
@@ -90,14 +90,8 @@ make checksum
 # unversioned names to avoid updating upstream Kubernetes documentation each release
 cp "out/minikube_${DEB_VERSION}-0_amd64.deb" out/minikube_latest_amd64.deb
 cp "out/minikube_${DEB_VERSION}-0_arm64.deb" out/minikube_latest_arm64.deb
-cp "out/minikube-${DEB_VERSION}-0_armhf.deb" out/minikube_latest_armhf.deb
-cp "out/minikube-${DEB_VERSION}-0_ppc64el.deb" out/minikube_latest_ppc64el.deb
-cp "out/minikube-${DEB_VERSION}-0_s390x.deb" out/minikube_latest_s390x.deb
 cp "out/minikube-${RPM_VERSION}-0.x86_64.rpm" out/minikube-latest.x86_64.rpm
 cp "out/minikube-${RPM_VERSION}-0.aarch64.rpm" out/minikube-latest.aarch64.rpm
-cp "out/minikube-${RPM_VERSION}-0.armv7hl.rpm" out/minikube-latest.armv7hl.rpm
-cp "out/minikube-${RPM_VERSION}-0.ppc64le.rpm" out/minikube-latest.ppc64le.rpm
-cp "out/minikube-${RPM_VERSION}-0.s390x.rpm" out/minikube-latest.s390x.rpm
 
 gsutil -m cp out/* "gs://$BUCKET/releases/$TAGNAME/"
 
--- a/netlify.toml
+++ b/netlify.toml
@@ -4,7 +4,7 @@ publish = "site/public/"
 command = "pwd && cd themes/docsy && git submodule update -f --init && cd ../.. && hugo"
 
 [build.environment]
-HUGO_VERSION = "0.68.3"
+HUGO_VERSION = "0.74.3"
 
 [context.production.environment]
 HUGO_ENV = "production"
--- a/netlify.toml
+++ b/netlify.toml
@@ -4,7 +4,7 @@ publish = "site/public/"
 command = "pwd && cd themes/docsy && git submodule update -f --init && cd ../.. && hugo"
 
 [build.environment]
-HUGO_VERSION = "0.74.3"
+HUGO_VERSION = "0.83.1"
 
 [context.production.environment]
 HUGO_ENV = "production"
--- a/site/assets/scss/_variables_project.scss
+++ b/site/assets/scss/_variables_project.scss
@@ -23,7 +23,7 @@ $gray-800: #333 !default;
 $gray-900: #222 !default;
 $black: #000 !default;
 
-$primary: #f2771a !default;
+$primary: $mk-dark !default;
 $primary-light: $mk-light;
 $secondary: #403F4C;
 $success: #3772FF !default;
@@ -227,4 +227,17 @@ div.td-content {
 
 div.code-toolbar > .toolbar {
     top: -.3em !important;
+}
+
+.option-row {
+    margin-bottom: 0.5rem;
+}
+
+.option-button {
+    border-radius: 0.2rem !important;
+    margin-right: 0.2rem;
+}
+
+.hide {
+    display: none !important;
 }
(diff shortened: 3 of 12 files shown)


hack/jenkins/release_build_and_upload.sh
```diff
@@ -90,8 +90,14 @@ make checksum
...
 cp "out/minikube_${DEB_VERSION}-0_amd64.deb" out/minikube_latest_amd64.deb
 cp "out/minikube_${DEB_VERSION}-0_arm64.deb" out/minikube_latest_arm64.deb
+cp "out/minikube-${DEB_VERSION}-0_armhf.deb" out/minikube_latest_armhf.deb
+cp "out/minikube-${DEB_VERSION}-0_ppc64el.deb" out/minikube_latest_ppc64el.deb
+cp "out/minikube-${DEB_VERSION}-0_s390x.deb" out/minikube_latest_s390x.deb
 cp "out/minikube-${RPM_VERSION}-0.x86_64.rpm" out/minikube-latest.x86_64.rpm
 cp "out/minikube-${RPM_VERSION}-0.aarch64.rpm" out/minikube-latest.aarch64.rpm
+cp "out/minikube-${RPM_VERSION}-0.armv7hl.rpm" out/minikube-latest.armv7hl.rpm
+cp "out/minikube-${RPM_VERSION}-0.ppc64le.rpm" out/minikube-latest.ppc64le.rpm
+cp "out/minikube-${RPM_VERSION}-0.s390x.rpm" out/minikube-latest.s390x.rpm
```
dinever, line 100: Note that we don't currently cp the deb & rpm packages for the other arches to the`latest` copy.

Meaning links like https://storage.googleapis.com/minikube/releases/latest/minikube_latest_armhf.deb will not work.

So I added these in the release script. The docs will only provide binary download instructions for now. Once they are available I can add those installation instructions in docs.

site/assets/scss/_variables_project.scss
```diff
@@ -23,7 +23,7 @@ $gray-800: #333 !default;
 $gray-900: #222 !default;
 $black: #000 !default;
 
-$primary: #f2771a !default;
+$primary: $mk-dark !default;
```
dinever, line 26: I don't think this `primary` color is actually used anywhere, and it's orange. So I changed it to our theme color which is the k8s blue.

medyagh:
@dinever thank you very much for taking this task ! I am so happy to see this page getting simplified, 
btw the netlify site build is failing, https://app.netlify.com/sites/kubernetes-sigs-minikube/deploys/609ccc2bb8a7a400089227d0

dinever:
> @dinever thank you very much for taking this task ! I am so happy to see this page getting simplified,
> btw the netlify site build is failing, https://app.netlify.com/sites/kubernetes-sigs-minikube/deploys/609ccc2bb8a7a400089227d0

Thanks! I've fixed the build failure.

I have to upgrade our Hugo version from `0.68.3` to 0.74.3` since I need to use the `strings.Count` template function.

The site is now available for preview at: https://deploy-preview-11399--kubernetes-sigs-minikube.netlify.app/docs/start/

Thanks!

medyagh:
@dinever thank you very much this looks much much better ! Is it possible that we have a default case on first view ?
For example in case of Linux show x86 binary

And also we used to have an auto detection for Mac and Linux and windows.... Does this pr preserve that ?

dinever:
> @dinever thank you very much this looks much much better ! Is it possible that we have a default case on first view ?
> For example in case of Linux show x86 binary
> 
> And also we used to have an auto detection for Mac and Linux and windows.... Does this pr preserve that ?

Thanks!

It auto-detects the OS and chooses it for the user. I've tested on Linux and macOS and the auto-detection worked as expected. Will test on Windows, too.

> Is it possible that we have a default case on first view ?

Sounds good. I can make it go one step further:

1. If user OS is Linux: default to Linux -> x86 -> binary
2. If user OS is macOS: default to macOS -> x86 -> binary
3. If user OS is Windows: default to Windows -> x86 -> .exe download
4. If user OS can't be detected: default to Linux -> x86 -> binary

dinever:
Hi @medyagh,

I've pushed the latest commit cc0d0652c571722a18b7adf70c420d35275fc3ef to make sure the first option of each row is auto-selected (OS->x86->binary download). In addition, I've confirmed that OS auto-detection works for all 3 platforms. This way we can minimize the number of clicks users need to perform.

Please see the following GIF for a quick demo, and feel to try it at https://deploy-preview-11399--kubernetes-sigs-minikube.netlify.app/docs/start/. Thanks a lot for the feedback!

installation

medyagh:
thank you very much @dinever it looks great, 



only thing is, possible to make the whole Install Section have a different background color simmilar to old way ? a light green/blue-ish ?


Screen Shot 2021-05-14 at 11 37 49 AM


btw in another note, not related to this PR but I wish our Boxes were not too long
for example the gray box for minikube start becomes very very long

hack/jenkins/release_build_and_upload.sh
medyagh: how about we do this Part in a separate PR ?

medyagh (changes requested):

netlify.toml
```diff
@@ -4,7 +4,7 @@ publish = "site/public/"
 command = "pwd && cd themes/docsy && git submodule update -f --init && cd ../.. && hugo"
 
 [build.environment]
-HUGO_VERSION = "0.68.3"
+HUGO_VERSION = "0.74.3"
```
medyagh, line 7: any reason we dont do latest version ? how did we pick 0.74.3 ? 
https://github.com/gohugoio/hugo/releases

medyagh (changes requested):
do u mind also pasting screenshot for a user who has disabled JavaScript, will it still have an acceptable look in case of no javascript ?

site/static/js/quiz.js
```diff
@@ -0,0 +1,67 @@
...
+        selectQuizOption(btn.attr('data-quiz-id'));
+    } catch(e) {
+        const elements = document.getElementsByClassName("quiz-instruction");
+        for (let element of elements) {
+            element.classList.remove("hide");
+        }
+    }
+}
+
+const getQuizUserOS = () => {
```
ilya-zuyev, line 54: Why different style for `getQuizUserOS` definition? Can we use `function getQuizUserOS()`to be consistent?

hack/jenkins/release_build_and_upload.sh
afbjorklund: It's perfectly fine to **only** document packages for amd64 and arm64

You could redirect the others to the "latest" github page or something
https://github.com/kubernetes/minikube/releases/latest

"For other architectures, see **here** for additional minikube binaries"

afbjorklund: It's quite possible that we will only do binaries for those architectures.

medyagh:
@dinever here is the screenshot without Javascript, could we make it in a way that a user without javascript can see the default page (even if it is ugly) ?
Screen Shot 2021-05-17 at 12 50 34 PM

hack/jenkins/release_build_and_upload.sh
dinever: Thanks! I reverted the changes in 120925eb7ba1080eb1320b8524219e3e96892767, and added a link to the release page:

Screen Shot 2021-05-18 at 1 52 18 AM

netlify.toml
dinever: I picked 0.74.3 because that's the lowest version with the util function I needed. But sure I've upgraded it to the latest `0.83.1` in ed7827913eadf498de154e5527907afe3a21d182 and didn't see any breaking changes. Thanks!

site/static/js/quiz.js
dinever: Thanks for the feedback! I originally want `quiz.js` to be self-contained. But yeah I just dropped it in a7d8a06f728eb44d77bceb3bf0c7ea42b4d288de to reuse the existing one. Thanks!

dinever:
Hi all,

Thanks for the feedback! I've pushed the latest commits:
1. ed7827913eadf498de154e5527907afe3a21d182: Update Hugo version to the latest 0.83.1 since I needed some template functions.
1. a7d8a06f728eb44d77bceb3bf0c7ea42b4d288de: Reuse the existing `getUserOS` from `tabs.js`
1. d50e86138d766ca28f727853677de1c2d69b2ece: Wrap the installation section within a blue card:
Screen Shot 2021-05-18 at 2 02 07 AM
1. 120925eb7ba1080eb1320b8524219e3e96892767: Revert the changes in `release_build_and_upload.sh`
1. d29d0a8fce1342cab5f6750f4e7f4ee4f7151239: Make sure the page renders acceptable result when JS is disabled:

| With JavaScript                                                                                                  	| Without JavaScript or with browsers that don't support jQuery (IE6)                                                         	|
|------------------------------------------------------------------------------------------------------------------	|-----------------------------------------------------------------------------------------------------------------------------	|
| install 	| install-without-js 	|

Feel free to try the preview site at https://deploy-preview-11399--kubernetes-sigs-minikube.netlify.app/docs/start/.

Thanks a lot!

medyagh (approved):

medyagh:
thank you very much @dinever

medyagh:
looks amazing

Issue #11333, opened 2021-05-07
fix TestNetworkPlugins/group/kubenet/Start for docker/containerd


ilya-zuyev:
TestNetworkPlugins/group/kubenet/Start times out for docker/containerd

https://storage.googleapis.com/minikube-builds/logs/master/c31bd57/Docker_Linux_containerd.html#fail_TestNetworkPlugins%2fgroup%2fkubenet%2fStart

Full log: https://storage.googleapis.com/minikube-builds/logs/master/c31bd57/Docker_Linux_containerdout.txt

PR #11412, opened 2021-05-14
Skip network test with kubenet if container runtime is not docker

ilya-zuyev:
Fixes https://github.com/kubernetes/minikube/issues/11333

### DIFF
--- a/test/integration/net_test.go
+++ b/test/integration/net_test.go
@@ -69,6 +69,10 @@ func TestNetworkPlugins(t *testing.T) {
 					t.Skipf("flannel is not yet compatible with Docker driver: iptables v1.8.3 (legacy): Couldn't load target `CNI-x': No such file or directory")
 				}
 
+				if !DockerDriver() && tc.name == "kindnet" {
+					t.Skipf("Skipping the test as %s container runtimes requires CNI", ContainerRuntime())
+				}
+
 				start := time.Now()
 				MaybeParallel(t)
 				profile := UniqueProfileName(tc.name)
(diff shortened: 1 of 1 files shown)


test/integration/net_test.go
```diff
@@ -69,6 +69,10 @@ func TestNetworkPlugins(t *testing.T) {
 					t.Skipf("flannel is not yet compatible with Docker driver: iptables v1.8.3 (legacy): Couldn't load target `CNI-x': No such file or directory")
 				}
 
+				if !DockerDriver() && tc.name == "kindnet" {
```
medyagh, line 72: mind commenting why Kindnet should Not work with CNI ?

ilya-zuyev: done!

test/integration/net_test.go
```diff
@@ -69,6 +69,13 @@ func TestNetworkPlugins(t *testing.T) {
 					t.Skipf("flannel is not yet compatible with Docker driver: iptables v1.8.3 (legacy): Couldn't load target `CNI-x': No such file or directory")
 				}
 
+				if !DockerDriver() && tc.name == "kubenet" {
+					// CNI is disabled when --network-plugin=kubenet option is passed. See cni.New(..) function
+					// But for containerd/crio CNI has to be configured
+					t.Skipf("Skipping the test as %s container runtimes requires CNI", ContainerRuntime())
+					return
```
spowelljr, line 76: The `return` is redundant since `t.Skipf` calls `t.SkipNow`

`SkipNow marks the test as having been skipped and stops its execution by calling runtime.Goexit.`

spowelljr (approved):

PR #11413, opened 2021-05-14
Add VMware as a provider for Linux & Windows

evankanderson:
There are Linux & Windows tabs at https://minikube.sigs.k8s.io/docs/drivers/vmware/, connecting them here.


medyagh:
@evankanderson I haven't tried vmware on Windows and Linux myself, so I can't verify, have you personally tried them on windows and linux ?

evankanderson:
I can verify on Windows; if the directions don't work, we should probably remove them, no?

evankanderson:
I just noticed that the linux directions are empty... but Windows works (quite smoothly, if you have VMware already installed).

I've removed the linux pointer.

PR #11420, opened 2021-05-15
Use fully qualified image name for kicbase image from docker hub

ntkme:
This PR fixes the following error under `podman` driver, where using docker hub fallback image  would not work due to using image short name.

```
! minikube was unable to download gcr.io/k8s-minikube/kicbase:v0.0.22, but successfully downloaded kicbase/stable:v0.0.22 as a fallback image
E0515 21:21:55.185211    8544 cache.go:189] Error downloading kic artifacts:  failed to download kic base image or any fallback image
* Creating podman container (CPUs=2, Memory=2200MB) ...
! StartHost failed, but will try again: creating host: create: creating: setting up container node: preparing volume for minikube container: sudo -n podman run --rm --name minikube-preload-sidecar --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --entrypoint /usr/bin/test -v minikube:/var kicbase/stable:v0.0.22 -d /var/lib: exit status 125
stdout:

stderr:
Error: error getting default registries to try: short-name resolution enforced but cannot prompt without a TTY
```

Reference: https://www.redhat.com/en/blog/be-careful-when-pulling-images-short-name

### DIFF
--- a/hack/jenkins/kicbase_auto_build.sh
+++ b/hack/jenkins/kicbase_auto_build.sh
@@ -46,13 +46,13 @@ if [[ -z $KIC_VERSION ]]; then
 	now=$(date +%s)
 	KV=$(egrep "Version =" pkg/drivers/kic/types.go | cut -d \" -f 2 | cut -d "-" -f 1)
 	GCR_REPO=gcr.io/k8s-minikube/kicbase-builds
-	DH_REPO=kicbase/build
+	DH_REPO=docker.io/kicbase/build
 	export KIC_VERSION=$KV-$now-$ghprbPullId
 else
 	# Actual kicbase release here
 	release=true
 	GCR_REPO=${GCR_REPO:-gcr.io/k8s-minikube/kicbase}
-	DH_REPO=${DH_REPO:-kicbase/stable}
+	DH_REPO=${DH_REPO:-docker.io/kicbase/stable}
 	export KIC_VERSION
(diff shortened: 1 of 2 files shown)


ntkme:
I signed it

afbjorklund:
It also works if you have your podman in compatible mode, instead of in breaking mode (and gcr.io is the default registry)

```
short-name-mode="permissive"
```

https://src.fedoraproject.org/rpms/containers-common/c/69f29483f1c7fed4e4fcdd5f9dace73b058de347?branch=rawhide

Either way, the fix is OK.

ntkme:
podman's `short-name-mode="permissive"` has security implication on the possibility of dependency hijacking due to podman would search all registries in the configuration for short name, which is different from docker's behavior that it would only search docker.io for short name.

sharifelgamal:
ok-to-build-kicbase

sharifelgamal:
Pulling master will fix the failing unit test here.

ntkme:
@sharifelgamal Rebased with latest master. Looks like it may need re-approval to run the test?

sharifelgamal:
ok-to-build-image

ntkme:
The CI seems to be flaky that it is always failing for unrelated things. Any way we can make progress?

spowelljr (approved):

PR #11431, opened 2021-05-17
Add "resume" as an alias for "unpause"

briandealwis:
This PR adds an alias for the "unpause" command for "resume".  I always make a mistaken running `minikube resume` and then realize that it's `unpause`.

An alternative would be to use a `SuggestFor: ["resume"]`.

And yes, I should be using auto-pause.

### DIFF
--- a/cmd/minikube/cmd/unpause.go
+++ b/cmd/minikube/cmd/unpause.go
@@ -39,9 +39,9 @@ import (
 
 // unpauseCmd represents the docker-pause command
 var unpauseCmd = &cobra.Command{
-	Use:   "unpause",
+	Use:     "unpause",
 	Aliases: []string{"resume"},
-	Short: "unpause Kubernetes",
+	Short:   "unpause Kubernetes",
 	Run: func(cmd *cobra.Command, args []string) {
 		cname := ClusterFlagValue()
 		register.SetEventLogPath(localpath.EventLog(cname))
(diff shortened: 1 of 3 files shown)


afbjorklund:
We sorta left suspend/resume for the VM, but then again it was never implemented either.

And I think it is called "pause" and "resume" in VBoxManage, so there's confusion anyway.

https://www.virtualbox.org/manual/ch08.html#vboxmanage-controlvm

> **VBoxManage controlvm <vm> pause**: Temporarily puts a virtual machine on hold, without permanently changing its state. The VM window is gray, to indicate that the VM is currently paused. This is equivalent to selecting the Pause item in the Machine menu of the GUI.
>
> **VBoxManage controlvm <vm> resume**: Undoes a previous pause command. This is equivalent to selecting the Resume item in the Machine menu of the GUI.

Not sure if the Docker VM has a similar feature ? How to best separate processes/machines.

briandealwis:
Then maybe the `SuggestFor` is better so as to keep `resume` available.

afbjorklund:
> Then maybe the `SuggestFor` is better so as to keep `resume` available.

Nah, pause/resume is fine. We can do something else, if we do VM sleep

afbjorklund (commented):
You need to run `make generate-docs`, then it should be good to go ?

medyagh (approved):

medyagh (changes requested):
thank you for this PR @briandealwis would u please `make generate-docs`

briandealwis:
I was just running `generate-docs` and saw that you'd beaten me to the punch @medyagh :-)

medyagh (approved):
thank you @briandealwis

PR #11443, opened 2021-05-18
add polkit check for kvm2 driver

maxamillion:
When using Fedora Silverblue 34 with libvirt installed in the ostree
overlay, the group memberships aren't setup as they would be in a
traditional GNU/Linux system. However, if a polkit policy is in place
that allows the user access to the "org.libvirt.unix.manage" action then
the user is able to create virtual machines via libvirt regardless of
group membership.

Signed-off-by: Adam Miller <maxamillion@gmail.com>

### DIFF
--- a/pkg/minikube/registry/drvs/kvm2/kvm2.go
+++ b/pkg/minikube/registry/drvs/kvm2/kvm2.go
@@ -28,6 +28,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/amenzhinsky/go-polkit"
 	"github.com/docker/machine/libmachine/drivers"
 
 	"k8s.io/minikube/pkg/minikube/config"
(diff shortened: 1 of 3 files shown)


maxamillion:
As an note, the CLA Check is pointing to the wrong GitHub account. I'm already working with GitHub support to fix that.

maxamillion:
The go-polkit library I used is currently unlicensed, I've requested it be licensed here: https://github.com/amenzhinsky/go-polkit/issues/1

maxamillion:
I got my GitHub account fixed so the CLA check should pass now.

pkg/minikube/registry/drvs/kvm2/kvm2.go
```diff
@@ -206,3 +221,26 @@ func isCurrentUserLibvirtGroupMember() (bool, error) {
 	}
 	return false, nil
 }
+
+// doesCurrentUserHaveLibvirtPolkitPermissions returns if the current user is allowed to perform the org.libvirt.unix.manage polkit action
+func doesCurrentUserHaveLibvirtPolkitPermissions() (bool, error) {
```
sharifelgamal, line 226: can we rename this to something a little shorter? hasPolkitPermissions() is a much shorter but equally clear name

medyagh:
@maxamillion we just merged a PR that did something simmilar,
https://github.com/kubernetes/minikube/pull/11842

do u mind studying that PR and check if this PR still needs to be done ?

maxamillion:
@medyagh LGTM! Thank you, I'll go ahead and close this. :+1:

PR #11448, opened 2021-05-18
ci: Add script to make sure docker is running for macos integration tests

sharifelgamal:

### DIFF
--- a/hack/jenkins/setup_docker_desktop_macos.sh
+++ b/hack/jenkins/setup_docker_desktop_macos.sh
@@ -16,6 +16,11 @@
 
 set -x
 
+if docker system info > /dev/null 2>&1; then
+  echo "Docker is already running, exiting"
+  exit 0
+fi
+
 # kill docker first
 osascript -e 'quit app "Docker"'
 
--- a/hack/jenkins/setup_docker_desktop_macos.sh
+++ b/hack/jenkins/setup_docker_desktop_macos.sh
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -x
+
 # kill docker first
 osascript -e 'quit app "Docker"'
 
@@ -34,4 +36,3 @@ do
 done
 
 echo "Docker Desktop started!"
-exit 0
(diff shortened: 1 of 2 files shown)


medyagh:
Docker_macOS — Jenkins: completed with 32 / 189 failures in 120.00 minute(s) seems like a timeout

sharifelgamal:
> Docker_macOS — Jenkins: completed with 32 / 189 failures in 120.00 minute(s) seems like a timeout

Perhaps, but that has nothing to do with this PR. We can investigate in a followup PR.

hack/jenkins/setup_docker_desktop_macos.sh
```diff
@@ -0,0 +1,38 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -x
+
+# kill docker first
+osascript -e 'quit app "Docker"'
```
medyagh, line 20: why kill docker if it is already running ? this could cause containers who have a volumes attached to them get stuck and not be remove-able.

how about restart it only if it is not already running successfully ?

medyagh (changes requested):

hack/jenkins/setup_docker_desktop_macos.sh
sharifelgamal: I was just following what we normally do, which is to always start fresh with a completely new docker.

sharifelgamal: it would be easy to add a check to see if docker is already running, but our tests shouldn't be leaving containers laying around either way

sharifelgamal: done

medyagh (approved):

medyagh:
great to see another all green test ! Hyperkit MacOS. thank you @sharif https://storage.googleapis.com/minikube-builds/logs/11448/40fd08e/Hyperkit_macOS.html thank you

Issue #2846, opened 2018-05-26
Failed to install longhorn, iscsiadm/open-iscsi is missing


ston3o:
**Minikube version**: v0.26.1
- **OS**: Arch Linux
- **VM Driver**: virtualbox
- **ISO version**: minikube-v0.26.0.iso

**What happened**:

When I try to install [longhorn](https://github.com/rancher/longhorn). I have this error : 

```
ERRO[0000] Failed environment check, please make sure you have iscsiadm/open-iscsi installed on the host
FATA[0000] Error starting manager: Environment check failed: Failed to execute: nsenter [--mount=/host/proc/1/ns/mnt --net=/host/proc/1/ns/net iscsiadm --version], output nsenter: failed to execute iscsiadm: No such file or directory           
, error exit status 1
```

**What you expected to happen**:

open-iscsi installed on minikube.

**How to reproduce it**:

```bash
kubectl create -f https://raw.githubusercontent.com/rancher/longhorn/master/deploy/longhorn.yaml
kubectl -n longhorn-system logs -f longhorn-manager-tbps9
```

**Anything else do we need to know**:

This Issue is related to another [closed issue](https://github.com/kubernetes/minikube/issues/1174).

lshmouse:
@johackim 

The reason is that minikube run in docker and there is no open-iscsi package that longhorn required (https://longhorn.io/docs/0.8.0/install/requirements/) in the minikube container. So just installing the open-iscsi in the minikube docker container will solve this issue, eg:
```
minikube ssh "sudo apt-get update;sudo apt-get install -y open-iscsi"
```

PR #1176, opened 2017-02-24
Added iscsi package to buildroot iso

aaron-prindle:
Ref: #1174

Issue #1174, opened 2017-02-24
iSCSI volumes can not be used because iscsiadm (and eventually kmods) are missing


fabiand:
**Minikube version** (use `minikube version`):
$ minikube version
minikube version: v0.16.0

**Environment**:
- **OS** (e.g. from /etc/os-release): linux/fedora
- **VM Driver** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep DriverName`): libvirt
- **ISO version** (e.g. `cat ~/.minikube/machines/minikube/config.json | grep ISO` or `minikube ssh cat /etc/VERSION`):
- **Install tools**:
- **Others**:

**What happened**:
I tried to use an iSCSI PVC on a pod

**What you expected to happen**:
The kubelet is able to attach the PVC to the pod

**How to reproduce it** (as minimally and precisely as possible):
Always

**Anything else do we need to know**:
The error is a little cryptic:
```
2017-02-24 16:51:08 +0100 CET   2017-02-24 16:43:00 +0100 CET   12        mypod     Pod                 Warning   FailedMount   {kubelet minikube}   MountVolume.SetUp failed for volume "kubernetes.io/iscsi/ebd01c3a-faa7-11e6-8b79-5254009a9694-myvol" (spec.Name: "myvol") pod "ebd01c3a-faa7-11e6-8b79-5254009a9694" (UID: "ebd01c3a-faa7-11e6-8b79-5254009a9694") with: executable file not found in $PATH
```
but I guess that this is because `iscsiadm` is missing in the kubelet env and thus the iscsi lun can not get attached.

aaron-prindle:
Right now the PR that I have does not properly build the open-iscsi binaries (iscsid and iscsiadm) and they are then not available in the ISO.  I used the buildfile here: http://lists.busybox.net/pipermail/buildroot/2007-February/013810.html as an example but it seems this source repository: http://www.open-iscsi.org/bits  Is no longer correct.  I am looking at using: https://github.com/open-iscsi/open-iscsi/releases but am having issues building it.  Any help/thoughts would be appreciated, thanks.

fabiand:
@aaron-prindle do you have a hint of how to test the builds? The I could take a look

fredleger:
Please re-open because this is still the case i think:

```shell
kubectl apply -f https://raw.githubusercontent.com/rancher/longhorn/v0.2/deploy/longhorn.yaml
[...]

PR #11458, opened 2021-05-19
Add support for iscsid in minikube

rajasaur:
There is no support currently for using  longhorn (or any storage implementation that requires open-iscsi) using minikube on VirtualBox. There are solutions when we use the docker driver (#2846, #1174) which also require to install it after minikube is started, but no out of the box solution available. 

There was a PR (#1176) started 3 years ago but it was incomplete and closed. I took that as a start to include open-iscsi and its related open-isns to bake this into a buildroot image so its avaiable as part of minikube.

This patch builds by adding 2 packages (open-iscsi , which depends on open-isns) and adding the required mounts that are required to persist the volumes that Longhorn creates in /var/lib/longhorn. The related kernel modules have been added as well.

Testing Procedure:
1. Build the ISO with this patch
2. Create a minikube install using this iso
    `minikube start -p test --cpus=2 --memory=4096 --driver=virtualbox --iso-url=file:///tmp/mk2.iso  --kubernetes-version=1.19.4`
3. Install LongHorn using the instructions at https://longhorn.io/docs/1.1.1/deploy/install/install-with-kubectl/
4. Create a PVC and a deployment that uses this. Scale up/down to make sure that the PVC is being used properly
5. Shut down minikube
6. Start it back up and make sure that the data is still available at the pod using the PVC.


rajasaur:
@afbjorklund Can you let me know if this should be made an addons instead (to start iscsi services for e.g.) ? I have only added support for open-iscsi as part of the main ISO so anything like longhorn or other plugins that need iscsi can be added independently after starting minikube.

geek (approved):

sharifelgamal:
Hey @rajasaur, this should probably be an addon instead of being baked into the ISO. I'd be happy to provide guidance there or review anything.

rajasaur:
Thanks @sharifelgamal , thought about that, but since this only adds the iscsi related utilities, and not really longhorn or any of the other providers that use iscsi, I dint go that route. 

If we make it an addon, do you think its okay to just make it install iscsi utils or include longhorn as well in the addons ?

sharifelgamal:
Addons are meant for ease-of-use for users, so if making it a longhorn addon makes life easier for you, then I say go for it. I have no real preference either way.

afbjorklund:
Haven't looked at this

spowelljr:
There haven't been any commits pushed to this PR in a while so I'm going to close it, feel free to reopen it if you come back to it, thanks!

PR #11461, opened 2021-05-19
Prevent downloading duplicate binaries already present in preload

### LINKED ISSUE #9098: minikube start --download-only downloads duplicate kubeadm/kubelet binaries

andriyDev:
fixes #9098.

Before
```
$ minikube start --download-only
😄  minikube v1.20.0 on Debian rodete
✨  Automatically selected the docker driver. Other choices: kvm2, ssh, none
🚜  Pulling base image ...
    > preloaded-images-k8s-v10-v1...: 491.71 MiB / 491.71 MiB  100.00% 22.69 Mi
    > kubelet.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubeadm.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubectl.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > gcr.io/k8s-minikube/kicbase...: 358.10 MiB / 358.10 MiB  100.00% 13.37 Mi
    > kubeadm: 37.40 MiB / 37.40 MiB [-------------] 100.00% 19.08 MiB p/s 2.2s
    > kubectl: 38.37 MiB / 38.37 MiB [-------------] 100.00% 17.89 MiB p/s 2.3s
    > kubelet: 108.73 MiB / 108.73 MiB [-----------] 100.00% 20.55 MiB p/s 5.5s
✅  Download complete!
```
After
```
$ minikube start --download-only
😄  minikube v1.20.0 on Debian rodete
✨  Automatically selected the docker driver. Other choices: kvm2, none, ssh
🚜  Pulling base image ...
    > preloaded-images-k8s-v10-v1...: 491.71 MiB / 491.71 MiB  100.00% 21.03 Mi
    > gcr.io/k8s-minikube/kicbase...: 358.09 MiB / 358.10 MiB  100.00% 14.94 Mi
    > kubectl.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubectl: 38.37 MiB / 38.37 MiB [-------------] 100.00% 19.40 MiB p/s 2.2s
✅  Download complete!
```
Note kubectl is still downloaded since the host requires it to be present. This does mean that downloading a Linux preload on a Linux host will still duplicate downloads for kubectl.

### DIFF
--- a/pkg/minikube/download/preload.go
+++ b/pkg/minikube/download/preload.go
@@ -45,15 +45,17 @@ const (
 	PreloadVersion = "v11"
 	// PreloadBucket is the name of the GCS bucket where preloaded volume tarballs exist
 	PreloadBucket = "minikube-preloaded-volume-tarballs"
+)
 
-	// Enumeration for preload existence cache.
-	preloadExistsUNKNOWN = 0 // Value when preload status has not been checked.
-	preloadExistsMISSING = 1 // Value when preload has been checked and is missing.
-	preloadExistsEXISTS  = 2 // Value when preload has been checked and is present.
+// Enumeration for preload existence cache.
+const (
+	preloadUnknown = iota // Value when preload status has not been checked.
+	preloadMissing = iota // Value when preload has been checked and is missing.
+	preloadPresent = iota // Value when preload has been checked and is present.
 )
 
 var (
-	preloadExistsState int = preloadExistsUNKNOWN
+	preloadState int = preloadUnknown
 )
 
 // TarballName returns name of the tarball
@@ -110,15 +112,15 @@ func PreloadExists(k8sVersion, containerRuntime string, forcePreload ...bool) bo
 	}
 
 	// If the preload existence is cached, just return that value.
-	if preloadExistsState != preloadExistsUNKNOWN {
-		return preloadExistsState == preloadExistsEXISTS
+	if preloadState != preloadUnknown {
+		return preloadState == preloadPresent
 	}
 
 	// Omit remote check if tarball exists locally
 	targetPath := TarballPath(k8sVersion, containerRuntime)
 	if _, err := checkCache(targetPath); err == nil {
 		klog.Infof("Found local preload: %s", targetPath)
-		preloadExistsState = preloadExistsEXISTS
+		preloadState = preloadPresent
 		return true
 	}
 
@@ -126,19 +128,19 @@ func PreloadExists(k8sVersion, containerRuntime string, forcePreload ...bool) bo
 	resp, err := http.Head(url)
 	if err != nil {
 		klog.Warningf("%s fetch error: %v", url, err)
-		preloadExistsState = preloadExistsMISSING
+		preloadState = preloadMissing
 		return false
 	}
 
 	// note: err won't be set if it's a 404
 	if resp.StatusCode != 200 {
 		klog.Warningf("%s status code: %d", url, resp.StatusCode)
-		preloadExistsState = preloadExistsMISSING
+		preloadState = preloadMissing
 		return false
 	}
 
 	klog.Infof("Found remote preload: %s", url)
-	preloadExistsState = preloadExistsEXISTS
+	preloadState = preloadPresent
 	return true
 }
 
@@ -202,7 +204,7 @@ func Preload(k8sVersion, containerRuntime string) error {
 	}
 
 	// If the download was successful, mark off that the preload exists in the cache.
-	preloadExistsState = preloadExistsEXISTS
+	preloadState = preloadPresent
 	return nil
 }
 
--- a/pkg/minikube/download/preload.go
+++ b/pkg/minikube/download/preload.go
@@ -47,9 +47,9 @@ const (
 	PreloadBucket = "minikube-preloaded-volume-tarballs"
 
 	// Enumeration for preload existence cache.
-	preloadExistsUNKNOWN = 0
-	preloadExistsMISSING = 1
-	preloadExistsEXISTS  = 2
+	preloadExistsUNKNOWN = 0 // Value when preload status has not been checked.
+	preloadExistsMISSING = 1 // Value when preload has been checked and is missing.
+	preloadExistsEXISTS  = 2 // Value when preload has been checked and is present.
 )
 
 var (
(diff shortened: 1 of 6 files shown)


pkg/minikube/download/preload.go
```diff
@@ -190,6 +193,38 @@ func Preload(k8sVersion, containerRuntime string) error {
 	return nil
 }
 
+var listTarContents = func(k8sVersion, containerRuntime string) (*bytes.Buffer, error) {
```
medyagh, line 196: It is safe to assume if there is a preload for a specific kubernetes version, then there is kubectl and other binaries inside of it, this operation is expensive and also wouldn’t work on non-Linux.

How about a simple check if the preload doesn’t exist for that k8s version or if the user explicitly specified —preload=false then we download those binaries ... kubeadm,kubectl,...

medyagh (changes requested):

medyagh:
marking as WIP please remove WIP once ready for another review

pkg/minikube/download/preload.go
andriyDev: Done!

medyagh (changes requested):

pkg/minikube/download/preload.go
```diff
@@ -45,6 +45,15 @@ const (
 	PreloadVersion = "v11"
 	// PreloadBucket is the name of the GCS bucket where preloaded volume tarballs exist
 	PreloadBucket = "minikube-preloaded-volume-tarballs"
+
+	// Enumeration for preload existence cache.
+	preloadExistsUNKNOWN = 0
+	preloadExistsMISSING = 1
```
medyagh, line 51: add comment here for all fields

pkg/minikube/download/preload.go
```diff
@@ -45,6 +45,15 @@ const (
 	PreloadVersion = "v11"
 	// PreloadBucket is the name of the GCS bucket where preloaded volume tarballs exist
 	PreloadBucket = "minikube-preloaded-volume-tarballs"
+
+	// Enumeration for preload existence cache.
```
medyagh, line 49: why not have boolean for all of these so it easier to read ?
either 
var exists *bool
or maybe a struct
```
type preloadStatus struct{
    exists *bool
}
```

then we could check 
```
if exists != nil && exists  {
...
```

medyagh (changes requested):

pkg/minikube/download/preload.go
andriyDev: I'm not a big fan of the constants-as-enum strategy, but I also think pointer to bool does not work well, since in order to assign exists you have to do:
```
var t = <value>
exists = &t
```
(unless I'm missing something).
I also considered just several bools, one for unknown and one for exists. So `unknown = true` means the preload must be checked, `unknown = false; exists = false` means definitely missing, and `unknown = false; exists = true` means definitely present. The trouble here is when `unknown = true` exists can "float" which leaves room for bugs. Thoughts?

medyagh: thats fair we could do this way too

pkg/minikube/download/preload.go
andriyDev: Done.

pkg/minikube/download/preload.go
```diff
@@ -45,6 +45,15 @@ const (
...
+	preloadExistsUNKNOWN = 0 // Value when preload status has not been checked.
```
sharifelgamal, line 50: `iota` seems like the better option here instead of 0, 1, and 2 directly. Makes it easier to maintain later on.
also I think we can agree these names are less than ideal. what about preload* instead of preloadExists* for them? it will still be pretty clear what's going and will be less awkward to read.

andriyDev: Ahhh ok I remember reading something about iota but I don't think those examples were good. I now understand and yes that is way better. I agree with preload* and have done accordingly. @medyagh also recommended removing the all caps, since that's not really the style in our codebase. Finally I had to rename preloadExists to preloadPresent because having preloadExists and PreloadExists does not sound fun :P

Done.

pkg/minikube/download/preload.go
```diff
@@ -47,6 +47,17 @@ const (
 	PreloadBucket = "minikube-preloaded-volume-tarballs"
 )
 
+// Enumeration for preload existence cache.
+const (
+	preloadUnknown = iota // Value when preload status has not been checked.
+	preloadMissing = iota // Value when preload has been checked and is missing.
```
sharifelgamal, line 53: all iota's below the first one are optional, which we normally omit in this codebase, otherwise LGTM

andriyDev: Done.

sharifelgamal (approved):

test/integration/aaa_download_only_test.go
```diff
@@ -131,17 +131,6 @@ func TestDownloadOnly(t *testing.T) {
 				}
 			})
 
-			t.Run("binaries", func(t *testing.T) {
```
sharifelgamal, line 134: how about instead of just deleting this test, we check for the preloadExists boolean and skip if it's true?

andriyDev: Done. Reverted the removal commit, and added the skip.

sharifelgamal (approved):

PR #11463, opened 2021-05-19
Fixed download path for linux rpm

### LINKED ISSUE #11460: https://minikube.sigs.k8s.io/docs/start linux x86_64 rpm download link fails
Download  link for 'Linux x86-64 rpm in https://minikube.sigs.k8s.io/docs/start/ is given as:
curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-latest.x86-64.rpm

But this doesn't work.

Real link appears to be at curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-latest.x86_64.rpm
(underscore instead of hyphen)

spowelljr:
Closes #11460


ilya-zuyev (approved):
lgtm

PR #11464, opened 2021-05-19
[wip] Fix GitHub Actions workflows to not run podman_functional on arm64

ilya-zuyev:


.github/workflows/pr.yml
```diff
@@ -287,7 +287,7 @@ jobs:
       JOB_NAME: functional_podman_ubuntu
       GOPOGH_RESULT: ""
       SHELL: "/bin/bash" # To prevent https://github.com/kubernetes/minikube/issues/6643
-    runs-on: ubuntu-20.04
+    runs-on: [ubuntu-20.04, !arm64]
```
medyagh, line 290: we could do is 
`!self-hosted` so it never runs on self-hosted ones

medyagh (changes requested):

.github/workflows/pr.yml
ilya-zuyev: It's still draft, trying to test different options :)

ilya-zuyev:
ok, GitHub Actions runner doesn't support `!` syntax, so we cannot exclude our runner using `!arm64`.  Instead, I'm going to close this PR and remove `ubuntu-20.04` from our aws arm64 runners, our arm jobs don't need it.

PR #11471, opened 2021-05-20
Update registry.md with instructions to avoid common pitfalls

EmandM:
I recently attempted to follow this guide and ran into two issues.
1. I already had a cluster running when I ran `minikube start --insecure-registry` and as such the insecure registry flag wasn't respected. I needed to run `minikube delete` and then could run the start command. As far as I'm aware the hard reset is only necessary on MacOS.
2. I was confused by the output from `minikube addons enable registry`. There is a big message of `Registry addon with docker driver uses port <generated-port> please use that instead of default port 5000` and so I switched a few things over to the generated port and nothing connected. Retrying with only port 5000 worked successfully.

### DIFF
--- a/site/content/en/docs/handbook/registry.md
+++ b/site/content/en/docs/handbook/registry.md
@@ -42,7 +42,8 @@ You can use the `--insecure-registry` flag on the
 
 One nifty hack is to allow the kubelet running in minikube to talk to registries deployed inside a pod in the cluster without backing them
 with TLS certificates. Because the default service cluster IP is known to be available at 10.0.0.1, users can pull images from registries
-deployed inside the cluster by creating the cluster with `minikube start --insecure-registry "10.0.0.0/24"`.
+deployed inside the cluster by creating the cluster with `minikube start --insecure-registry "10.0.0.0/24"`. If using macOS, ensure the cluster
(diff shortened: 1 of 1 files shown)


site/content/en/docs/handbook/registry.md
```diff
@@ -42,7 +42,8 @@ You can use the `--insecure-registry` flag on the
 
 One nifty hack is to allow the kubelet running in minikube to talk to registries deployed inside a pod in the cluster without backing them
 with TLS certificates. Because the default service cluster IP is known to be available at 10.0.0.1, users can pull images from registries
-deployed inside the cluster by creating the cluster with `minikube start --insecure-registry "10.0.0.0/24"`.
+deployed inside the cluster by creating the cluster with `minikube start --insecure-registry "10.0.0.0/24"`. If using macOS, ensure the cluster
```
medyagh, line 45: lets remove "if using macOS" this would be same for all platforms

medyagh (changes requested):

site/content/en/docs/handbook/registry.md
EmandM: Thanks, I wasn't sure if it was true on all operating systems

PR #11483, opened 2021-05-21
Add ability to create extra disks on hyperkit vms

### LINKED ISSUE #3883: Ability to add disks to the minikube VM on Hyperkit?
While adding extra disks with the kvm2 driver is easy, I couldn't find anything to add disks on `hyperkit`. Anyone already doing this perhaps?

Thanks!

BlaineEXE:
Add the ability to create and attach extra disks to hyperkit vms.

For example, if I wish to add 3 extra disk to my minikube VM, each 10GB in size, the command with the 2 new options would look like below. 
```sh
> minikube start --driver=hyperkit --extra-disks=3 --extra-disk-size=10gb
```
In the minikube VM, I would expect to see the extra disks (in addition to `vda` and `vda1` normally present) available in `/dev` as shown below.
```sh
> minikube ssh
$  ls /dev/vd*
vda
vda1
vdb
vdc
vdd
```

Signed-off-by: Blaine Gardner <blaine.gardner@redhat.com>

Fixes #3883 
This will be useful for me for developing on Rook, and I think it maybe useful for others doing other types of storage development on Kubernetes with minikube.

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -120,6 +120,7 @@ const (
 	defaultSSHUser          = "root"
 	defaultSSHPort          = 22
 	listenAddress           = "listen-address"
+	extraDisks              = "extra-disks"
 )
 
 var (
@@ -166,6 +167,7 @@ func initMinikubeFlags() {
 	startCmd.Flags().StringP(network, "", "", "network to run minikube with. Now it is used by docker/podman and KVM drivers. If left empty, minikube will create a new network.")
 	startCmd.Flags().StringVarP(&outputFormat, "output", "o", "text", "Format to print stdout in. Options include: [text,json]")
 	startCmd.Flags().StringP(trace, "", "", "Send trace events. Options include: [gcp]")
+	startCmd.Flags().Int(extraDisks, 0, "Number of extra disks created and attached to the minikube VM (currently only implemented for hyperkit driver)")
 }
 
 // initKubernetesFlags inits the commandline flags for Kubernetes related options
@@ -407,6 +409,8 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, drvName s
 
 	checkNumaCount(k8sVersion)
 
+	checkExtraDiskOptions(cmd, drvName)
+
 	cc = config.ClusterConfig{
 		Name:                    ClusterFlagValue(),
 		KeepContext:             viper.GetBool(keepContext),
@@ -449,6 +453,7 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, drvName s
 		SSHUser:                 viper.GetString(sshSSHUser),
 		SSHKey:                  viper.GetString(sshSSHKey),
 		SSHPort:                 viper.GetInt(sshSSHPort),
+		ExtraDisks:              viper.GetInt(extraDisks),
 		KubernetesConfig: config.KubernetesConfig{
 			KubernetesVersion:      k8sVersion,
 			ClusterName:            ClusterFlagValue(),
@@ -558,6 +563,11 @@ func updateExistingConfigFromFlags(cmd *cobra.Command, existing *config.ClusterC
 
 	if cmd.Flags().Changed(humanReadableDiskSize) && getDiskSize() != existing.DiskSize {
 		out.WarningT("You cannot change the disk size for an existing minikube cluster. Please first delete the cluster.")
+	}
+
+	checkExtraDiskOptions(cmd, cc.Driver)
+	if cmd.Flags().Changed(extraDisks) && viper.GetInt(extraDisks) != existing.ExtraDisks {
+		out.WarningT("You cannot add or remove extra disks for an existing minikube cluster. Please first delete the cluster.")
 	}
 
 	updateStringFromFlag(cmd, &cc.MinikubeISO, isoURL)
@@ -705,4 +715,21 @@ func interpretWaitFlag(cmd cobra.Command) map[string]bool {
 	}
 	klog.Infof("Waiting for components: %+v", waitComponents)
 	return waitComponents
+}
+
+func checkExtraDiskOptions(cmd *cobra.Command, driverName string) {
+	supportedDrivers := []string{driver.HyperKit}
+
+	if cmd.Flags().Changed(extraDisks) {
+		supported := false
+		for _, driver := range supportedDrivers {
+			if driverName == driver {
+				supported = true
+				break
+			}
+		}
+		if !supported {
+			out.WarningT("Specifying extra disks is currently only supported for the following drivers: {{.supported_drivers}}. If you can contribute to add this feature, please create a PR.", out.V{"supported_drivers": supportedDrivers})
+		}
+	}
 }
--- a/pkg/drivers/common.go
+++ b/pkg/drivers/common.go
@@ -17,6 +17,7 @@ limitations under the License.
 package drivers
 
 import (
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -39,6 +40,12 @@ import (
 // GetDiskPath returns the path of the machine disk image
 func GetDiskPath(d *drivers.BaseDriver) string {
 	return filepath.Join(d.ResolveStorePath("."), d.GetMachineName()+".rawdisk")
+}
+
+// ExtraDiskPath returns the path of an additional disk suffixed with an ID.
+func ExtraDiskPath(d *drivers.BaseDriver, diskID int) string {
+	file := fmt.Sprintf("%s-%d.rawdisk", d.GetMachineName(), diskID)
+	return filepath.Join(d.ResolveStorePath("."), file)
 }
 
 // CommonDriver is the common driver base class
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ KUBERNETES_VERSION ?= $(shell egrep "DefaultKubernetesVersion =" pkg/minikube/co
 KIC_VERSION ?= $(shell egrep "Version =" pkg/drivers/kic/types.go | cut -d \" -f2)
 
 # Default to .0 for higher cache hit rates, as build increments typically don't require new ISO versions
-ISO_VERSION ?= v1.22.0
+ISO_VERSION ?= v1.22.0-1627488369-11483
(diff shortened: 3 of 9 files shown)


medyagh (commented):
@BlaineEXE thank you for contributing this interesting PR. do u mind adding in the PR description the After this PR in action ?

BlaineEXE:
> @BlaineEXE thank you for contributing this interesting PR. do u mind adding in the PR description the After this PR in action ?

I'm happy to add more details or move them as is most helpful. This is my first PR to the minikube repo, and I'm not quite sure what you mean exactly, though. I don't see anything like that in the contributor guide to help me get there.

medyagh:
> > @BlaineEXE thank you for contributing this interesting PR. do u mind adding in the PR description the After this PR in action ?
> 
> I'm happy to add more details or move them as is most helpful. This is my first PR to the minikube repo, and I'm not quite sure what you mean exactly, though. I don't see anything like that in the contributor guide to help me get there.

thank you for reading the contributor guide and welcome to contributing to minikube. I meant like if you run the PR with this new feature locally and paste the output of how u would use this new feature in the PR description, so it be easier for me to review it

BlaineEXE:
> thank you for reading the contributor guide and welcome to contributing to minikube. I meant like if you run the PR with this new feature locally and paste the output of how u would use this new feature in the PR description, so it be easier for me to review it

Done. Let me know if more details would be helpful.

BlaineEXE:
@medyagh are you able to approve running workflows? I think to fix the issue I mentioned I need to make a change to the `deploy/iso/minikube-iso/package/automount/minikube-automount` script to support multiple disks, and I'd like to be able to use the ISO created by the CI if possible since it seems to take a very long time to build on my machine and failed recently for me.

cmd/minikube/cmd/start_flags.go
```diff
@@ -164,6 +166,8 @@ func initMinikubeFlags() {
 	startCmd.Flags().StringP(network, "", "", "network to run minikube with. Now it is used by docker/podman and KVM drivers. If left empty, minikube will create a new network.")
 	startCmd.Flags().StringVarP(&outputFormat, "output", "o", "text", "Format to print stdout in. Options include: [text,json]")
 	startCmd.Flags().StringP(trace, "", "", "Send trace events. Options include: [gcp]")
+	startCmd.Flags().Int(extraDisks, 0, "Number of extra disks created and attached to the minikube VM.")
```
medyagh, line 169: add `(only used by HyperKit driver)`

cmd/minikube/cmd/start_flags.go
```diff
@@ -164,6 +166,8 @@ func initMinikubeFlags() {
...
+	startCmd.Flags().String(extraDiskSize, defaultDiskSize, "Disk size allocated for extra disks attached to the minikube VM (format: <number>[<unit>], where unit = b, k, m or g).")
```
medyagh, line 170: add ` only used by hyperkit driver`

pkg/drivers/common.go
```diff
@@ -41,6 +42,12 @@ func GetDiskPath(d *drivers.BaseDriver) string {
 	return filepath.Join(d.ResolveStorePath("."), d.GetMachineName()+".rawdisk")
 }
 
+// GetExtraDiskPath returns the path of an additional disk suffixed with an ID.
+func GetExtraDiskPath(d *drivers.BaseDriver, diskID int) string {
```
medyagh, line 46: no need for Get here

lets rename to `ExtraDiskPath`

cmd/minikube/cmd/start_flags.go
```diff
@@ -525,9 +531,15 @@ func updateExistingConfigFromFlags(cmd *cobra.Command, existing *config.ClusterC
 	// validate the memory size in case user changed their system memory limits (example change docker desktop or upgraded memory.)
 	validateRequestedMemorySize(cc.Memory, cc.Driver)
```
medyagh, line 533: add a warning if user added this flag to non-hyperkit drivers, that
```
out.WarningT("Specifying extra disk is currently only supported for hyperkit driver, if you can contribute to add this feature please create a PR....")
```

medyagh (changes requested):

cmd/minikube/cmd/start_flags.go
BlaineEXE: I added this check for flags on both create and update. If I were a user, I think I would expect using these flags for unsupported drivers would cause a failure since the option isn't supported. Would it be better for me to return an error instead of just logging a warning?

cmd/minikube/cmd/start_flags.go
BlaineEXE: I opted to message this as `currently only implemented for hyperkit driver` since I think that makes it more clear that it can be implemented for other drivers but isn't currently. Let me know if your wording is still preferred.

cmd/minikube/cmd/start_flags.go
medyagh: since we are not passing the option to the driver if it doesnt suppor it, it wont do anything, I would be okay with ExitWith Usage or just warning. your call !

cmd/minikube/cmd/start_flags.go
```diff
@@ -118,6 +118,8 @@ const (
 	defaultSSHUser          = "root"
 	defaultSSHPort          = 22
 	listenAddress           = "listen-address"
+	extraDisks              = "extra-disks"
+	extraDiskSize           = "extra-disk-size"
```
medyagh, line 122: for the sake of not having too many flags, what do u think/feel about re-suing the existing the flag we have `disk-size`
```
$ minikube start --help | grep disk
      --disk-size='20000mb': Disk size allocated to the minikube VM (format: <number>[<unit>], where unit = b, k, m or g).
```

so if user passes this flag we will use disk-size value as the extra disk size and if they pass --disk-size it will be used for both main disk and extra disk, is that something acceptable ?

medyagh (changes requested):
also consider adding to the Hyperkit website documentation a section called 

"driver specific features"
- add extra disk...
 example ...

https://minikube.sigs.k8s.io/docs/drivers/hyperkit/

cmd/minikube/cmd/start_flags.go
BlaineEXE: I'll change this to exit with usage soon.

cmd/minikube/cmd/start_flags.go
BlaineEXE: I like that idea, and I thought about doing that. It is more flexible to have separate options, but it does add complexity for the user. I keep waffling on this. For all of the use-cases I have in mind, I don't think there is a very strong reason to have the extra parameter for development just to save a few 10s of gigabytes on a hard disk. I think I will take your suggestion to simply re-use the `disk-size` parameter for extra disks as well, and if that ends up being insufficient, it can always be changed in the future.

BlaineEXE:
@medyagh the ISO build keeps failing on my machine, so I'm unable to test my latest changes to the ISO. It's my understanding from [here](https://minikube.sigs.k8s.io/docs/contrib/building/iso/) that the CI will build the ISO which I can test locally.

medyagh:
@BlaineEXE no worries we can build the ISO on the PR

medyagh:
ok-to-build-iso

leseb (commented):
I don't see code to clean up the disks, is this handled automatically when removing the VM? Just want to make sure we don't have disk files lingering around when the env is destroyed.

BlaineEXE:
> I don't see code to clean up the disks, is this handled automatically when removing the VM? Just want to make sure we don't have disk files lingering around when the env is destroyed.

I don't think it's necessary. When creating a VM, minikube creates a directory for each machine in the config directory (e.g., `$HOME/.minikube/machines/<machine-name>`), and when the machine is deleted the dir (with the raw disk files) is deleted also.

BlaineEXE:
> /ok-to-build-iso

@medyagh is this the correct command? I don't see it in [this list](https://prow.k8s.io/command-help?repo=kubernetes%2Fminikube), and it doesn't seem like building the ISO has been authorized. Should this be `/ok-to-test`?

medyagh:
ok-to-build-iso

medyagh:
@BlaineEXE the ISO builds on CI too, so I think there is something not right

sharifelgamal:
> @BlaineEXE the ISO builds on CI too, so I think there is something not right

It turns out there was an issue on our CI machines. It's fixed and the job is running again.

sharifelgamal:
ok-to-build-iso

BlaineEXE:
It seems like there are some CI failures, and I'm not sure if they could have been a different issue unrelated to this PR. I will try rebasing on the latest master soon.

Now that the new ISO is built, I was able to test locally, and it works! 🥳  I tested the following scenarios:
- single node with no extra disks
- single node with 3 extra disks
- 3 nodes each with 1 extra disk

BlaineEXE:
@medyagh am I able to request the CI build the ISO for me again since the master ISO updated to 1.21?

sharifelgamal:
ok-to-build-iso

BlaineEXE:
No problem @medyagh. Thanks for looking. I just updated on the upstream master.

BlaineEXE:
@medyagh It's a little unclear to me if the `Docker_Linux` and `Docker_Linux_containerd` tests have failed or passed. It seems like there were 267 and 256 successes respectively with some flakiness. Is that right?

Are there more things I should be doing from here to get this merged?

sharifelgamal:
ok-to-build-iso

BlaineEXE:
Thanks @sharifelgamal. Would you be able to help me find logs from the script `deploy/iso/minikube-iso/package/automount/minikube-automount` as it runs on test VMs in the failing tests? I'm having a little trouble figuring out if those are collected and where I can find them if they are.

sharifelgamal:
I don't personally know where those are collected, if I had to guess, I'd say there are not. Either way, if the logs aren't exported, they're eventually lost.

I'm not sure I see where the problem is, all the hyperkit tests passed. I'm comfortable merging this if you are.

BlaineEXE:
@sharifelgamal I think I feel good if this is merged. The flakes here don't seem to be far off what I see in other PRs, and I have tested this pretty thoroughly on my side with starting, adding, stopping, and re-starting the cluster with the extra disks.

sharifelgamal (approved):
Sounds good. Thanks for your contribution and your patience!

PR #11517, opened 2021-05-26
vmware: Add new flag to specify network type through `--vmware-network-type`

zarenner:
Following https://github.com/machine-drivers/docker-machine-driver-vmware/pull/34, adds support for specifying the VMWare network connection type (e.g. `nat` vs `bridged`)

usage text:
```
      --vm-driver='': DEPRECATED, use `driver` instead.
      --vmware-network-type='nat': Network connection type to use. One of 'nat', 'bridged', 'hostonly', 'custom' (vmware
driver only)
      --wait=[apiserver,system_pods]: comma separated list of Kubernetes components to verify and wait for after
```

Running e.g. `minikube start --driver=vmware --vmware-network-type=hostonly` results in a `.vmx` with `ethernet0.connectionType = "hostonly"`
image


zarenner:
Hmm I have signed the CLA... investigating.

medyagh:
@zarenner do u mind posting an example of this new flag output in the PR description ?
consider adding a section in the webste for this flag and why and how to use it
https://minikube.sigs.k8s.io/docs/drivers/vmware/

zarenner:
@medyagh Added a screenshot to the description for the example there, and added a docs change. Thanks for the quick feedback!

medyagh (approved):

medyagh:
@zarenner please take a look at the unit test

```
make test
```

cmd/minikube/cmd/start_flags.go
```diff
@@ -118,6 +118,7 @@ const (
 	defaultSSHUser          = "root"
 	defaultSSHPort          = 22
 	listenAddress           = "listen-address"
+	vmwareNetworkType       = "vmware-network-type"
```
medyagh, line 121: @zarenner sorry for the changing in my opinion, but how about we called this flag "network-type"
and then then specify this is only currently used for for vmwware and then if user specificed a value for this flag for non-vmware drivers, we should Exit with Exit With Usage

```
				exit.Message(reason.Usage, "network-type flag is currently only supported only by the vm-ware driver "
```

medyagh (changes requested):

PR #11522, opened 2021-05-27
make generate-docs

spowelljr:
`make generate-docs` to update error codes


ilya-zuyev (commented):
Could you add a short PR description? 
Just something like "make generate-docs to fix linter errors" or whatever this PR is for

PR #11525, opened 2021-05-27
bump default k8s version and newest

medyagh:


medyagh:
merge only if the tests are all normal and performance metric looks normal

spowelljr (approved):

PR #11528, opened 2021-05-27
Change PreloadExists to report no preload for BareMetal drivers

### LINKED ISSUE #11526: TestDownloadOnly failing

andriyDev:
fixes #11526.

Previously, PreloadExists would report true for none driver. Now it doesn't (as expected)!


medyagh (changes requested):
I belive SSH driver would need the binaries too (maybe)
since that is also without virtualization
I dont think preload works on SSH driver

how about adding a new func here
```

// BareMetal returns if this driver is unisolated
func BareMetal(name string) bool {
	return name == None || name == Mock
}

// IsSSH checks if the driver is ssh
func IsSSH(name string) bool {
	return name == SSH
}
```

that is like

```
driver.NeedsDownloadK8SBinaries....

and then returns true if Baremetal or SSH
```

pkg/minikube/download/preload.go
```diff
@@ -97,8 +98,12 @@ func remoteTarballURL(k8sVersion, containerRuntime string) string {
 	return fmt.Sprintf("https://storage.googleapis.com/%s/%s", PreloadBucket, TarballName(k8sVersion, containerRuntime))
 }
 
+func driverAllowsPreload(driverName string) bool {
```
medyagh, line 101: driver.AllowsPreload(

medyagh (changes requested):

medyagh:
@andriyDev your PR looks good

medyagh (approved):

PR #11535, opened 2021-05-29
Fix TestFunctional/parallel/LogsCmd flakiness

### LINKED ISSUE #11442: Test Flake: "TestFunctional/parallel/LogsCmd"
as seen here

https://github.com/kubernetes/minikube/pull/11431/checks?check_run_id=2606299017

dinever:
Hi,

This PR fixes the flakiness of `TestFunctional/parallel/LogsCmd`, closes #11442 

The idea is to avoid logging the containers at status `CREATED` since there are no logs yet.

To reproduce the issue:
1. Start a `containerd` runtime profile
2. SSH into the image
3. Manually create a container:
    1. `sudo crictl pull busybox`
    1. Create `container-config.json`:
        ```
        {
          "metadata": {
            "name": "storage-provisioner"
          },
          "image":{
            "image": "docker.io/busybox"
          },
          "command": [
            "top"
          ],
          "log_path":"busybox.0.log",
          "linux": {
          }
        }
        ```
    1. Create `podsandbox-config.json`:
        ```
        {
          "metadata": {
            "name": "busybox-sandbox",
            "namespace": "default",
            "attempt": 1,
            "uid": "hdishd83djaidwnduwk28bcsb"
          },
          "log_directory": "/tmp",
          "linux": {
          }
        }
        ```
    1. Create the container: ``sudo crictl create `sudo crictl pods --quiet --name=storage-provisioner` container-config.json podsandbox-config.json``
    1. `sudo crictl ps -a` now shows a container at status `CREATED`
4. `minikube logs` fails now because `CREATED` containers have no logs

Thanks!


pkg/minikube/cruntime/cri.go
```diff
@@ -40,8 +40,20 @@ type container struct {
 func crictlList(cr CommandRunner, root string, o ListContainersOptions) (*command.RunResult, error) {
 	klog.Infof("listing CRI containers in root %s: %+v", root, o)
 
+	var filter string
+	switch o.State {
+	case All:
+		filter = "-a"
+	case Running, Paused:
+		// crictl does not understand paused containers
+		// we have to list all the running ones and later filter out paused ones by runc
```
medyagh, line 49: mind clarifying where we do the filtering out paused ones? is this required for this PR or is it a good to have ?

pkg/minikube/logs/logs.go
```diff
@@ -258,11 +258,38 @@ func OutputOffline(lines int, logOutput *os.File) {
...
+	}
+	return loggableIDs, nil
+}
+
 // logCommands returns a list of commands that would be run to receive the anticipated logs
 func logCommands(r cruntime.Manager, bs bootstrapper.Bootstrapper, cfg config.ClusterConfig, length int, follow bool) map[string]string {
 	cmds := bs.LogCommands(cfg, bootstrapper.LogOptions{Lines: length, Follow: follow})
 	for _, pod := range importantPods {
-		ids, err := r.ListContainers(cruntime.ListContainersOptions{Name: pod})
+		ids, err := listLoggableContainers(r, pod)
```
medyagh, line 292: @dinever 
it seems like, after this PR, for each PR , we do a Fresh new "critctl ls" and filter out the running ones from paused ones, I suspect this would cause perfromance problems, (maybe if minikube is already under resource problem and user wants to see the `minikube logs` 

is there a better way to do this ? 
what are alternatives to this solution that we could evaluate their pros and cons?

medyagh (changes requested):

pkg/minikube/logs/logs.go
dinever: Thanks for the feedback!

Actually the original logic gets all containers, too. But this PR will make an extra call to `r.ListContainers`.

Some other options:

### only print out a warning if we can't get logs for a container

This is an easier way to resolve the issue. Something like this for [logs/logs.go#L202-L204](https://github.com/kubernetes/minikube/blob/90137069195bd3aafd0a4b8d77eaa0f7d00d677d/pkg/minikube/logs/logs.go#L202-L204):

```diff
        if len(failed) > 0 {
-               return fmt.Errorf("unable to fetch logs for: %s", strings.Join(failed, ", "))
+               out.WarningT("unable to fetch logs for: %s", strings.Join(failed, ", "))
        }
        return nil
```

The question is how many failures should we tolerate? What if we can't get logs from all containers? Maybe we should return an error if all containers failed:

```diff
        if len(failed) > 0 {
-               return fmt.Errorf("unable to fetch logs for: %s", strings.Join(failed, ", "))
+               out.WarningT("unable to fetch logs for: %s", strings.Join(failed, ", "))
+       }
+       // if failed to retrieve logs from all containers
+       if len(failed) == len(names) {
+               return fmt.Errorf("unable to fetch logs for all containers: %s", strings.Join(failed, ", "))
        }
```

### if a container log fails, could we check if it is "created" and in that case not fail ?

This will work too, just need to add a new method for the container runtime interface:

```diff
@@ -110,6 +110,9 @@ type Manager interface {
        // RemoveImage remove image based on name
        RemoveImage(string) error

+       // GetContainerState gets the state of a container based on container ID
+       GetContainerState(string) (ContainerState, error)
+
```

Also, there could be a race condition issue. For example:
1. Container is created
2. `logs` failed to get the container just created
3. . Container starts running
4. `logs` checks the container, gets a `Running` state
5. `logs` fails

Another issue is if we are adding all possible container states in `ContainerState`:

```diff
@@ -41,6 +41,8 @@ const (
        Running
        // Paused is only paused
        Paused
+       Created
+       Exited
 )
```

Should we support filtering containers by these too? Since `ContainerState` is also used to filter containers:

```go
// ListContainersOptions are the options to use for listing containers
type ListContainersOptions struct {
	// State is the container state to filter by (All, Running, Paused)
	State ContainerState
	// Name is a name filter
	Name string
	// Namespaces is the namespaces to look into
	Namespaces []string
}
```

Let me know what you think. Thanks!

pkg/minikube/cruntime/cri.go
dinever: Thanks! It's done with `runc` later:

https://github.com/kubernetes/minikube/blob/e6ea64792f1521f5444897171d4a852ecfaea910/pkg/minikube/cruntime/cri.go#L134-L137

This is just a nice to have to avoid getting all containers when we are only looking for `paused` onces.

The original logic:
1. Get all containers from `crictl`
2. Filter paused containers by `runc`

New logic:
1. Get running containers from `crictl` (Paused containers will have `Running` state in crictl)
2. Filter paused containers by `runc`

pkg/minikube/logs/logs.go
medyagh: @dinever 
```this PR will make an extra call to r.ListContainers.
```

I am curious how much does this extra call to list containers on for eveyr Pod costs, do u mind sharing the `time ` for minikube logs before and after this PR for containerd

dinever: Hi @medyagh, sorry for the late update. I ran some benchmarks with `minikube logs` on this profile:

<details>

```
CONTAINER           IMAGE               CREATED             STATE               NAME                        ATTEMPT             POD ID
95e21923395bf       9a07b5b4bfac0       9 minutes ago       Running             kubernetes-dashboard        0                   ab574f1fa4208
d36b5f624e7ae       86262685d9abb       9 minutes ago       Running             dashboard-metrics-scraper   0                   8814222dc5c62
4329d264cc76c       60dc18151daf8       10 minutes ago      Running             registry-proxy              0                   a361628ce8458
aa2e7d38abc91       678dfa38fcfa3       10 minutes ago      Running             registry                    0                   7b01e66fdf4d1
c944a1cf57af1       435df390f3673       12 minutes ago      Running             controller                  0                   25bfd0f32767e
69a4c15b5470a       a013daf8730db       12 minutes ago      Exited              patch                       0                   3262b4af19d9c
cc3ddf29e46fc       a013daf8730db       12 minutes ago      Exited              create                      0                   19e3b34cb0b30
5a098e06202df       bfe3a36ebd252       14 minutes ago      Running             coredns                     0                   7278fb89d20cc
45e4ccb3d3358       6e38f40d628db       14 minutes ago      Running             storage-provisioner         0                   0c68c7f5ad351
9b6facef62f8d       6de166512aa22       14 minutes ago      Running             kindnet-cni                 0                   12890d7dc41ff
f16e46a29d112       43154ddb57a83       14 minutes ago      Running             kube-proxy                  0                   34f2fca84cb17
924e3b896b7b7       a27166429d98e       14 minutes ago      Running             kube-controller-manager     0                   ca972111f6f83
71521574abc7e       a8c2fdb8bf76e       14 minutes ago      Running             kube-apiserver              0                   0cf7defef9d89
f69bb27686da9       ed2c44fbdd78b       14 minutes ago      Running             kube-scheduler              0                   790675330c4ae
b7236acae53f1       0369cf4303ffd       14 minutes ago      Running             etcd                        0                   9c6ad8fb20834
```
</details>

Herer are the results:

| `master`     | This branch  |
|--------------|--------------|
| 524.035838ms | 611.59094ms  |
| 523.230265ms | 611.087835ms |
| 534.198272ms | 624.215021ms |
| 535.836799ms | 624.9862ms   |
| 518.748063ms | 643.71546ms  |
| 533.488633ms | 619.309344ms |
| 533.141091ms | 614.404019ms |
| 525.430241ms | 627.025792ms |
| 525.421662ms | 612.77719ms  |
| 525.409901ms | 620.369326ms |

So looks like an extra `crictl ps --state=created` call did add about 100ms overhead to the run time. Maybe a safer option is to just print out a warning if we can't get logs for a container? (see my above comment)

Another option is that we refactor the `ListContainers` method to return a list of Container structs instead of just string IDs. Then we can later filter out the containers we need in code. This will only take one call to `crictl ps -a --output=json`.

Something like:

```diff
--- a/pkg/minikube/cruntime/cruntime.go
+++ b/pkg/minikube/cruntime/cruntime.go
@@ -31,6 +31,11 @@ import (
        "k8s.io/minikube/pkg/minikube/sysinit"
 )
 
+type Container struct {
+       ID    string
+       State ContainerState
+}
+
 // ContainerState is the run state of a container
 type ContainerState int
 
@@ -113,7 +118,7 @@ type Manager interface {
        RemoveImage(string) error
 
        // ListContainers returns a list of containers managed by this container runtime
-       ListContainers(ListContainersOptions) ([]string, error)
+       ListContainers(ListContainersOptions) ([]Container, error)
        // KillContainers removes containers based on ID
        KillContainers([]string) error
        // StopContainers stops containers based on ID
```

Let me know what you think. Thanks!

pkg/minikube/logs/logs.go
```diff
@@ -261,13 +261,26 @@ func OutputOffline(lines int, logOutput *os.File) {
 // logCommands returns a list of commands that would be run to receive the anticipated logs
 func logCommands(r cruntime.Manager, bs bootstrapper.Bootstrapper, cfg config.ClusterConfig, length int, follow bool) map[string]string {
 	cmds := bs.LogCommands(cfg, bootstrapper.LogOptions{Lines: length, Follow: follow})
+	containers, err := r.ListContainers(cruntime.ListContainersOptions{})
+	if err != nil {
+		klog.Errorf("Failed to list containers: %v", err)
+		return nil
+	}
```
dinever, line 268: Added a `Name` field under `Container` struct.

This way we only need to do **one** `ListContainers` call during `minikube logs`, and later filter container names in code. This makes `minikube logs` about 15% - 20% faster e2e.

Some benchmark results (done on a minikube profile I was working on with 14 containers):

| `master`     | This branch  |
|--------------|--------------|
| 824.937374ms | 680.089196ms |
| 845.681224ms | 670.916441ms |
| 827.478504ms | 677.877367ms |
| 831.836328ms | 688.230854ms |
| 825.594147ms | 678.318603ms |
| 809.029915ms | 669.383535ms |

dinever (commented):
Hi @medyagh,

I've done refactoring `ListContainers` to return a slice of struct `[]Container` instead of just container IDs.

```go
// Container holds the information of a container
type Container struct {
	// ID is the ID of the container
	ID string
	// Name is the name of the container, e.g. `kube-apiserver`
	Name string
	// State is the container state
	State ContainerState
}
```

Please take a look when available. It would be great if you can allow this PR to be tested too. Thanks!

medyagh (approved):

medyagh:
sorry @dinever for delay in this PR review, would u plz rebase this PR one more time

dinever:
> sorry @dinever for delay in this PR review, would u plz rebase this PR one more time

Hi @medyagh, thanks for checking in! Just rebased it. Feel free to take a look when you're available. Thanks!

medyagh:
@dinever sorry for the long wait on this PR review do u mind rebasing this PR

PR #11576, opened 2021-06-03
Fix embed-certs global config

### LINKED ISSUE #11253: `minikube config set embed-certs true` does have no effect

1. Run `minikube config set embed-certs true`.
2. Verify that it has been set: `minikube config get embed-certs`
3. Restart minikube.
4. View `$HOME/.kube/config`. There is no field `user.client-certificate-data` in the `minikube` profile.

**Expected behavior**

The certificate should be embedded directly in the `minikube` profile inside `$HOME/.kube/config`.

However, starting the cluster using `minikube start --embed-certs` will work.

JacekDuszenko:
fixes #11253
Fix embed-certs global config by adding `embed-certs` json tag to `EmbedCerts` field in `ClusterConfig` struct.

Unmarshaler in https://github.com/kubernetes/minikube/blob/master/pkg/minikube/config/config.go#L199 wasn't able to deserialize  `embed-certs` json  field into `EmbedCerts` string member variable in `ClusterConfig`. 

That's why existing config in https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/start.go#L173 was created without given flag. 

Now global config option successfully initialises cert embedding on cluster creation

### DIFF
--- a/cmd/minikube/cmd/config/config.go
+++ b/cmd/minikube/cmd/config/config.go
@@ -172,7 +172,7 @@ var settings = []Setting{
 		setMap: SetMap,
 	},
 	{
-		name: "embed-certs",
+		name: config.EmbedCerts,
 		set:  SetBool,
 	},
 	{
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -301,6 +301,7 @@ func setupViper() {
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 	viper.AutomaticEnv()
 
+	viper.RegisterAlias(config.EmbedCerts, embedCerts)
 	viper.SetDefault(config.WantUpdateNotification, true)
 	viper.SetDefault(config.ReminderWaitPeriodInHours, 24)
 	viper.SetDefault(config.WantReportError, false)
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -169,7 +169,6 @@ func runStart(cmd *cobra.Command, args []string) {
 		out.WarningT("Profile name '{{.name}}' is not valid", out.V{"name": ClusterFlagValue()})
 		exit.Message(reason.Usage, "Only alphanumeric and dashes '-' are permitted. Minimum 2 characters, starting with alphanumeric.")
 	}
-
 	existing, err := config.Load(ClusterFlagValue())
 	if err != nil && !config.IsNotExist(err) {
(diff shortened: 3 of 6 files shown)


JacekDuszenko:
@medyagh could you mark this PR as "ok-to-test" please?

ilya-zuyev (changes requested):
When set with `minikube config set embed-certs true` the cluster config is:

``` 
ilyaz@skeletron --- g/minikube ‹fix-embed-certs-from-config* ?› » head  ~/.minikube/profiles/minikube/config.json -n 20
{
        "Name": "minikube",
        "KeepContext": false,
        "embed-certs": true,
        "MinikubeISO": "",
        "KicBaseImage": "gcr.io/k8s-minikube/kicbase-builds:v0.0.22-1620785771-11384@sha256:f5844fe35994179bbad8dda27d4912304a2fedccdf0bf93ce8b2ec2b3b83af1c",
        "Memory": 8000,
        "CPUs": 2,
        "DiskSize": 20000,
        "VMDriver": "",
        "Driver": "docker",
        "HyperkitVpnKitSock": "",
        "HyperkitVSockPorts": [],
        "DockerEnv": null,
        "ContainerVolumeMounts": null,
        "InsecureRegistry": null,
        "RegistryMirror": null,
        "HostOnlyCIDR": "192.168.99.1/24",
        "HypervVirtualSwitch": "",
        "HypervUseExternalSwitch": false,
```
Using different naming scheme for "embed-certs" is a bit ugly,  we should use the same style for all fields. Also, it's probably best to avoid exceptions in serialization of individual fields of `ClusterConfig`.
Let's fix the `config` command instead, and use `"EmbedCerts"` name to store the config value

JacekDuszenko:
Sure, will submit a PR shortly

JacekDuszenko:
@ilya-zuyev @medyagh I fixed the PR. Now config is saved as `EmbedCerts`. The viper key had to stay as `embed-certs` to preserve backwards compatibility, because the startup flag `--embed-certs` would be changed if we changed the viper key (because of how the code is structured). I went with the approach of creating `EmbedCerts` alias to `embed-certs` viper prop key and it works, allowing us to use `EmbedCerts` in global config and `--embed-certs` as a startup flag.

JacekDuszenko:
Hey, @ilya-zuyev, would you mind taking a look at this?

ilya-zuyev (approved):
looks good now. Thanks!

JacekDuszenko:
Great, thank you! Are we able to merge it anytime soon @ilya-zuyev?

PR #11585, opened 2021-06-05
minikube start --image-repository will now accept URLs with port

### LINKED ISSUE #11579: minikube start --image-repository=xxx does not work with ports in URL
Hi there, 
I have identified an issue whereby I consider this to be a regression, because it was certainly working in some of the past minikube releases (a number of releases back, cannot find this now easily). 

With https://github.com/kubernetes/minikube/blob/v1.19.0/cmd/minikube/cmd/start_test.go minikube 1.19 there is now a `TestValidateImageRepository`. This checks for different formatting of the registry URL. 

Our nexus3 docker registry has this format: `nexus.devops.firm.com:6666`. We use this type of URL everywhere else successfully and no known issues have been observed with this format. However, we started to transition to using nexus registry "everywhere", i.e. also for minikube and when installing minikube from scratch with following:

1. `minikube start --image-repository=nexus.devops.firm.com:6666 --driver=none --delete-on-failure=true`

We get this error:

```
 minikube v1.20.0 on Centos 7.9.2009
* Using the none driver based on user configuration
  - The --image-repository flag your provided contains Scheme: nexus.devops.firm.com, it will be as a domian, removed automatically
* Using image repository
```

I.e. 6666 is stripped, hence the internal URL is ignored and hence it falls back to the assumption that VM in question has access to the internet (which in our case does not and hence we found out about this). 

Cannot post because a lot of private details (username, etc.). Sorry. But I am willing to test previews/rc/beta versions and maybe even contribute :)

dmpe:
fixes #11579 

fixes:

- [x] misspellings -> rename variables and functions

- [x] --image-repository shall now accept URLs with ports

### Notes:
- my first attempt at `go`
- my first (potential) contribution to kubernetes/minikube project (k8s for that matter either) :palm_tree:


dmpe:
@afbjorklund can you please ok'it to test - so that test can be run. 

thanks.

dmpe:
@RA489 maybe you can also take a look? Thanks

dmpe:
@medyagh Hi, can you please take a look on this. I would like to get it merged :)

medyagh (approved):

dmpe:
Hi @medyagh, @RA489 @afbjorklund 

Thanks for running tests on this. Besides them I have also just run my own testing with sonatype/nexus3 container. 
I have started empty nexus3, configured 2 repos: `k8s.gcr.io` and `gcr.io` in order to proxy images locally (as we do at work) and then run:
```
sudo ./minikube-linux-amd64 start --driver=none --image-repository http://localhost:6668 (proxy to k8s.gcr.io)
```
Result:
```
😄  minikube v1.22.0-beta.0 on Ubuntu 21.04
    ▪ The --image-repository flag your provided contains Scheme: http, which will be removed automatically
✅  Using image repository localhost:6668
🤹  Running on localhost (CPUs=4, Memory=4178MB, Disk=79614MB) ...
ℹ️  OS release is Ubuntu 21.04
❗  This bare metal machine is having trouble accessing https://localhost:6668
🐳  Preparing Kubernetes v1.20.8 on Docker 20.10.7 ...
    ▪ Configuring RBAC rules ...

    ▪ Using image localhost:6668/k8s-minikube/storage-provisioner:v5 (global image repository)
```
I.e. it is now working and using nexus3 docker images, instead of going outside. This is also validated by the fact, that nexus3 contains all those docker images. 


2 Points:

> `The --image-repository flag your provided contains Scheme: http, which will be removed automatically`

In order to use minikube with nexus3 ports, you will need to specify scheme, even though I am not sure it is necessary. Certainly `docker pull localhost:6669/k8s-minikube/storage-provisioner:v5` is working, while `docker pull http://localhost:6669/k8s-minikube/storage-provisioner:v5` is correctly not working. So the message applies only for certain contexts -  and with ports it shall not apply at all.


> `This bare metal machine is having trouble accessing https://localhost:6668`

This message is also confusing, because port 6668 is not `https`, but rather (as already sayd above) `http`. It seems to be hardcoded https://github.com/kubernetes/minikube/blob/master/pkg/minikube/node/start.go#L653 here. Probably a new issue and new PR, later. 

wdyt? There are some tests which fail, but seem to be unrelated to my PR. 

The executable, used for testing, was downloaded using 
```
gsutil ls gs://minikube-builds/11585/minikube-linux-amd64
```

cmd/minikube/cmd/start.go
```diff
@@ -1274,26 +1274,41 @@ func validateRegistryMirror() {
...
+	// tips when imageRepo started with scheme such as http(s).
 	if URL.Scheme != "" {
-		out.Infof("The --image-repository flag your provided contains Scheme: {{.scheme}}, it will be as a domian, removed automatically", out.V{"scheme": URL.Scheme})
+		out.Infof("The --image-repository flag your provided contains Scheme: {{.scheme}}, which will be removed automatically", out.V{"scheme": URL.Scheme})
+	}
+
+	if hasPorts {
+		validImageRepo = URL.Hostname() + imageRepoPort + strings.TrimSuffix(URL.Path, "/")
+	} else {
+		validImageRepo = URL.Hostname() + strings.TrimSuffix(URL.Path, "/")
```
spowelljr, line 1308: This works perfectly fine, but could be streamlined a bit.

The `hasPorts` variable could be removed altogether, if the `if` on 1292 is false, `imageRepoPort` will be empty, so you can always do `validImageRepo = URL.Hostname() + imageRepoPort + strings.TrimSuffix(URL.Path, "/")` and will get the same result as now.

spowelljr:
@dmpe The tests look fine, we have many flaky tests. I left a comment on this PR, it's not a requirement to address the comment, so I'll leave it up to you to decide. If you don't want to implement the comment lmk and I'll merge this PR now.

dmpe:
Hi @spowelljr - thank you very much for looking into this and giving me a feedback. 

I have rebased on latest master, run `make tests` and did again above testing (works) and it looks good to me. There is an issue with coredns image (`coredns:v1.8.0` - neither from gcr.io nor from k8s.gcr.io - does not exist, but retagging an image from dockerhub `coredns/coredns:1.8.0` did the trick) but otherwise it is working good. Feel free to merge :)

Thanks again

PR #11590, opened 2021-06-07
Add detailed info generation for integration tests

### LINKED ISSUE #11571: add detailed info about what each integration test is doing to the site
currently our `make generate-docs` creates this page

https://minikube.sigs.k8s.io/docs/contrib/tests/

which generates the site based on the Comments on the Test and validate the func,

I would like these descriptions to exaclty say what it is doing instead of a [general text ](https://minikube.sigs.k8s.io/docs/contrib/tests/#validatebuildimage)

```
validateBuildImage 
makes sures that minikube image build works as expected


```

it should say
```
// validateBuildImage makes sures that `minikube image build` works as expected
// builds an image using `minikube image` and the Dockerfile located in test/integration/testdata/build/Dockerfile 
// which adds a relative file into a busybox container
// and then ensures the image was build using image inspect inside minikbue 
// special case :
// in case of contained it starts the buildkit inside minikube
// currently skipped on None Driver and Github Actions on MacOs,

```

dinever:
Closes #11571

Hi, this one is an initial draft for the detailed testing docs generation.

It currently only adds the docs for `validateBuildImage`. Will add other docs and tests once we reach an agreement on the general design.

Thanks!


test/integration/functional_test.go
```diff
@@ -324,21 +324,24 @@ func listImages(ctx context.Context, t *testing.T, profile string) (*RunResult,
 
 // validateBuildImage makes sures that `minikube image build` works as expected
 func validateBuildImage(ctx context.Context, t *testing.T, profile string) {
+	// docs(skip): Skips on non driver since load image is not available on none driver
 	if NoneDriver() {
 		t.Skip("load image not available on none driver")
```
dinever, line 329: It would be cool if we can just parse out the skip info from the parameter of `t.Skip`. However, it's not doable for `t.Skipf` calls because we can't get the evaluated string without actually running the code, e.g.:

```go
t.Skipf("only validate docker env with docker container runtime, currently testing %s", cr)
```

So I added a `docs(skip): ` tag to document the skip cases.

An alternative approach is that we make sure we are only using `t.Skip`:

```diff
- t.Skipf("only validate docker env with docker container runtime, currently testing %s", cr)
+ t.Skip("only validate docker env with docker container runtime")
```

Then it's possible to parse out the string from the function parameter.

medyagh (commented):
thank you for doing this PR @dinever plz undraft once ready for review and I would love to see this detailed comment for most other tests to tell the contributor what this test is actually doing,
so we dont need to read the full code to see what it is doing

dinever:
> thank you for doing this PR @dinever plz undraft once ready for review and I would love to see this detailed comment for most other tests to tell the contributor what this test is actually doing,
> so we dont need to read the full code to see what it is doing

Yep, thanks! Will add docs for other tests as well.

pkg/generate/testdocs.go
```diff
@@ -107,6 +81,113 @@ func TestDocs(docPath string, pathToCheck string) error {
...
+	b := &strings.Builder{}
+	if d.isSubTest {
+		b.WriteString("#### " + d.name + "\n")
+	} else {
+		b.WriteString("## " + d.name + "\n")
+	}
+
+	srcLinkText := fmt.Sprintf("%s#L%d-L%d", d.path, d.lineStart, d.lineEnd)
+	srcLinkURL := fmt.Sprintf("https://github.com/kubernetes/minikube/blob/master/%s#L%d-L%d", d.path, d.lineStart, d.lineEnd)
+	b.WriteString(fmt.Sprintf("Source code: [%s](%s)\n\n", srcLinkText, srcLinkURL))
```
dinever, line 117: Screenshot from 2021-06-16 00-24-52

Added a `source code` link for each test since that might be useful? Note that the URL links to the master branch git tree. When building the site from this branch the line numbers may not match the current master branch line numbers since this branch modifies the test file. However, when building the site from master branch the line numbers should match.

dinever: Actually, I removed this `source code` for now since it requires developers to rerun `make generate-docs` every time they touch the testing files (because the line numbers will change, most likely) otherwise unit testing will fail. This can be a little bit annoying...

dinever:
Hi @medyagh this PR is ready for an initial review. I've finished the documentation for all the functional tests. There are still many other integration tests to document. I may have to do those in another PR. Thanks!

medyagh:
@dinever  rebase plz

dinever:
> @dinever rebase plz

Rebased. Thanks

sharifelgamal:
Hey @dinever, this change looks really good. Could you rebase one more time before we merge it?

dinever:
> Hey @dinever, this change looks really good. Could you rebase one more time before we merge it?

Hi @sharifelgamal, sorry for the delayed response. I got the conflict resolved. Would be great if you can take a look again. Thanks!

dinever:
> Thanks so much for doing this! We'll be adding more detailed descriptions for tests over time.

Thanks for reviewing!

PR #11598, opened 2021-06-07
Support copy file from node to local host or between nodes.

### LINKED ISSUE #11296: `minikube cp` should support copying from minkube to user's host

### LINKED ISSUE #11336: Support `cp` between minikube nodes
Related to https://github.com/kubernetes/minikube/issues/11296

It would be nice to have an ability to copy files not only between the host and a node, but also between minikube nodes

`minikube cp node1:/tmp/foo node2:/tmp/bar`

daehyeok:
Fixes #11296 
Fixes #11336 

Support copy file from node to local or node to node.
For the implementation, add new `type ReadableFile interface` to vm_assets and `func (s *SSHRunner) ReadableFile` to SSHRunner. 
 current only SSHRunner used after start, so leave `ReadableFile` as a empty in  ExecRunner and KICRunner. 
 Let me know my implementation not looks good :) 
  thanks


pkg/minikube/command/command_runner.go
```diff
@@ -77,6 +77,8 @@ type Runner interface {
 
 	// Remove is a convenience method that runs a command to remove a file
 	Remove(assets.CopyableFile) error
+
+	ReadableFile(sourcePath string) (assets.ReadableFile, error)
```
medyagh, line 81: comment on this interface, what is this method exactly and is this the best name for it ?

medyagh (changes requested):

pkg/minikube/command/command_runner.go
daehyeok: I split `assets.CopyableFile` to source(`assets.ReadableFile` interface) and target part in `pkg/minikube/assets/vm_assets.go`.
And this function defined to get source file info/io.Readable interface from remote node. hmm any other suggestion?

pkg/minikube/command/ssh_runner.go
```diff
@@ -49,6 +52,32 @@ type SSHRunner struct {
 	s *ssh.Session
 }
 
+type SSHReadableFile struct {
```
medyagh, line 55: add comment to all public structs, 
does this need to be public ?

pkg/minikube/command/command_runner.go
medyagh: thats fine but add a comment, all public funcs needs to have comment

cmd/minikube/cmd/cp_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"testing"
+)
+
+func TestParsePath(t *testing.T) {
+	var passedCases = []struct {
+		path     string
+		expected remotePath
+	}{
+
+		{"", remotePath{"", ""}},
```
medyagh, line 29: instead of calling remotePath here, lets add the concrete string here, so the test be simple

medyagh (changes requested):

medyagh:
@daehyeok please pull upstream on this PR

daehyeok:
rebased

daehyeok:
sorry to couldn't care this isuue for while. 

Copy function looks already implemented by other commit.
https://github.com/kubernetes/minikube/commit/e59c41e6f2c9a78be4aff3d0a615ce9f93a49d2f

i will change PR to use new api soon.

medyagh:
@daehyeok sorry for the changes, could u please rebase, we added some API

daehyeok:
Hi i modified my commit and rebased.

on the previous change(e59c41e) `commandRunner` interface have a new api `CopyFrom` and `assets.Copyablefile` include `io.Writer`

but  CopyFrom looks only working for between host and node(not node to node) and `io.Writer` also only used in `ssh_runner.Copyfrom`. I added `BaseCopyableFile` interface and used empty write function to support node to node copy through `ssh_runner`. 
 Maybe we should more clean up those implementation later.

medyagh:
@daehyeok are you still working on this ?

PR #11601, opened 2021-06-07
Prevent misleading "save to tar file succeeded" message from being logged

spowelljr:
If we don't return early, the message `save to tar file %s -> %s succeeded` gets logged, which is incorrect because the save to tar failed since the cached image doesn't exist.

This early return will prevent the success message from being logged and possibly causing confusion.

**Before:**
Output to user:
```
❗  The image you are trying to add ubuntu doesn't exist!
```
Log output:
```
save to tar file ubuntu -> /usr/home/.minikube succeeded
```

**After:**
Output to user:
```
❗  The image 'ubuntu' was not found; unable to add it to cache.
```
Log output:
_Empty_

### DIFF
--- a/pkg/minikube/image/cache.go
+++ b/pkg/minikube/image/cache.go
@@ -73,7 +73,7 @@ func SaveToDir(images []string, cacheDir string, overwrite bool) error {
 			dst = localpath.SanitizeCacheDir(dst)
 			if err := saveToTarFile(image, dst, overwrite); err != nil {
 				if err == errCacheImageDoesntExist {
-					out.WarningT("The image you are trying to add {{.imageName}} doesn't exist!", out.V{"imageName": image})
+					out.WarningT("The image '{{.imageName}}' was not found; unable to add it to cache.", out.V{"imageName": image})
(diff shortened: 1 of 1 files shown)


pkg/minikube/image/cache.go
```diff
@@ -74,9 +74,9 @@ func SaveToDir(images []string, cacheDir string, overwrite bool) error {
 			if err := saveToTarFile(image, dst, overwrite); err != nil {
 				if err == errCacheImageDoesntExist {
 					out.WarningT("The image you are trying to add {{.imageName}} doesn't exist!", out.V{"imageName": image})
```
sharifelgamal, line 76: I know this isn't related to this PR, but this error message is so grammatically poor. Could you change it something a little better? I would be forever in your debt.

spowelljr: I reworded it; no guarantee it's any better.

sharifelgamal: this is much better, thanks

sharifelgamal (approved):

PR #11616, opened 2021-06-09
Delete `minikube-integration` folder in case last test failed to delete it

spowelljr:
**Problem:**
If the previous test was unexpectedly ended, the teardown script that deletes the `minikube-integration` folder is not run. This leaves the folder up containing the kube config and minikube home folder which can be in a bad state and affect the next test.

**Solution:**
Delete the folder on the setup script as well to ensure the folder is empty.

### DIFF
--- a/hack/jenkins/windows_integration_setup.ps1
+++ b/hack/jenkins/windows_integration_setup.ps1
@@ -17,5 +17,5 @@ $env:KUBECONFIG="$test_home\kubeconfig"
 $env:MINIKUBE_HOME="$test_home\.minikube"
 
 # delete in case previous test was unexpectedly ended and teardown wasn't run
-rm -r $test_home
+rm -r -Force $test_home
 mkdir -p $test_home
--- a/hack/jenkins/windows_integration_setup.ps1
+++ b/hack/jenkins/windows_integration_setup.ps1
@@ -12,9 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-$test_root="$env:HOMEDRIVE$env:HOMEPATH\minikube-integration"
-$test_home="$test_root\$env:COMMIT"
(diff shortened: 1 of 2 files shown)


hack/jenkins/windows_integration_setup.ps1
```diff
@@ -12,9 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-$test_root="$env:HOMEDRIVE$env:HOMEPATH\minikube-integration"
-$test_home="$test_root\$env:COMMIT"
+$test_home="$env:HOMEDRIVE$env:HOMEPATH\minikube-integration"
```
sharifelgamal, line 15: this is effectively changing what $test_home is actually set to. is that safe?

spowelljr: I originally set it to a commit  specific folder because I thought that would prevent the described problem. However, I forgot to take into that these tests are run a schedule and the same commit on head could be run multiple times.

Also, if I used a commit specific folder and the teardown script doesn't run the folder will stay there forever. Using the same folder for every test will ensure it's always properly wiped.

sharifelgamal: gotcha, makes sense

hack/jenkins/windows_integration_setup.ps1
```diff
@@ -12,9 +12,10 @@
...
 $env:KUBECONFIG="$test_home\kubeconfig"
 $env:MINIKUBE_HOME="$test_home\.minikube"
 
+# delete in case previous test was unexpectedly ended and teardown wasn't run
+rm -r $test_home
```
medyagh, line 20: should this be -rf ? so it does't fail ?

spowelljr: Added

PR #11626, opened 2021-06-10
add test for proxies with custom certs

### LINKED ISSUE #11267: create mitm proxy test for Custom Cert
https://github.com/kubernetes/minikube/issues/8834
https://github.com/kubernetes/minikube/issues/11068

sharifelgamal:
This uses the [mitmdump binary](https://docs.mitmproxy.org/stable/#mitmdump) to create a certificate that then gets installed. We then run minikube start with the HTTPS_PROXY env var.

Closes #11267


test/integration/functional_test.go
```diff
@@ -1782,6 +1783,80 @@ users:
...
+	}
+
+	certDir, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatalf("creating temp dir failed: %v", err)
+	}
+
+	// Start an interactive session (since mitmproxy requires it) asyncronously
+	// This will create the necessary .mitmproxy directory with its certs
+	mitmCmd := exec.CommandContext(ctx, "docker", "run", "-it", "--rm", "--name", "mitmproxy", "-v", certDir, ":/home/mitmproxy/.mitmproxy", "-p", "8080:8080", "mitmproxy/mitmproxy")
```
medyagh, line 1807: this might cause docker rate limit. I recommend using non docker registry or install the binary on the github action maybe?

sharifelgamal: yeah we can create our own docker image with the mitmproxy binary if necessary

sharifelgamal: we're using a binary directly now, so this is no longer an issue

site/content/en/docs/contrib/tests.en.md
```diff
@@ -96,6 +96,10 @@ check functionality of minikube after evaluating podman-env
 #### validateStartWithProxy
 makes sure minikube start respects the HTTP_PROXY environment variable
 
+#### validateStartWithCorpProxy
```
medyagh, line 99: lets rename to Proxy with custom cert

sharifelgamal: done

site/content/en/docs/contrib/tests.en.md
```diff
@@ -96,6 +96,10 @@ check functionality of minikube after evaluating podman-env
 #### validateStartWithProxy
 makes sure minikube start respects the HTTP_PROXY environment variable
 
+#### validateStartWithCustomCerts
+makes sure minikube start respects the HTTPS_PROXY environment variable
```
medyagh, line 100: and "custom certs" 
since this is a new test, consider adding details of how it runs

like 

a mitmproxy runs with a customer proxy alongside...this test runs only on ... then checks for ...

this could be example for other tests docs

medyagh (changes requested):

site/content/en/docs/contrib/tests.en.md
sharifelgamal: done. i've removed the entry in the docs so that the gen-docs github action can run as a test

medyagh (approved):

PR #11631, opened 2021-06-11
Add too high k8s version validation

### LINKED ISSUE #11627: improve exit message when trying to pull Non-Existing Kubernetes Version
I made a mistake and put in the wrong k8s version, we should have a more beautiful looking error, and ideally do not even try to create container when a non-existing k8s version is specified
```
mk start --kubernetes-version=20.10.7
😄  minikube v1.21.0-beta.0 on Darwin 11.4
✨  Automatically selected the docker driver. Other choices: hyperkit, parallels, virtualbox, ssh
🚜  Pulling base image ...
🔥  Creating docker container (CPUs=2, Memory=4000MB) ...\ E0610 16:43:42.794488   82266 cache.go:211] Error caching images:  cached images list: version too new: 20.10.7

🐳  Preparing Kubernetes v20.10.7 on Docker 20.10.7 ...

❌  Exiting due to K8S_INSTALL_FAILED: kubeadm images: version too new: 20.10.7

│    - /Users/medya/.minikube/logs/lastStart.txt                     │


```

JacekDuszenko:
fixes #11627
This PR introduces another constraint to `validateKubernetesVersion` function, checking if version is less or equal than the newest supported version.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -32,6 +32,8 @@ import (
 	"strconv"
 	"strings"
 
+	"k8s.io/minikube/pkg/minikube/github"
+
 	"github.com/blang/semver"
 	"github.com/docker/machine/libmachine/ssh"
 	"github.com/google/go-containerregistry/pkg/authn"
@@ -1424,7 +1426,20 @@ func validateKubernetesVersion(old *config.ClusterConfig) {
 	}
 
 	if nvs.GT(newestVersion) {
-		exit.Message(reason.KubernetesTooNew, "Specified Kubernetes version {{.specified}} is higher than newest supported version: {{.newest}}", out.V{"specified": nvs, "newest": constants.NewestKubernetesVersion})
+		k8sReleases, err := github.RecentK8sVersions()
+		if err != nil {
+			exit.Message(reason.KubernetesReleaseFetchFailed, "Failed fetching available Kubernetes releases from GitHub repository. Check your internet connection")
+		}
+		versionFound := false
+		nvsStr := nvs.String()
+		for _, release := range k8sReleases {
+			if strings.TrimPrefix(release, version.VersionPrefix) == nvsStr {
+				versionFound = true
+			}
+		}
+		if !versionFound {
+			exit.Message(reason.KubernetesVersionNotFound, "Specified Kubernetes version {{.specified}} was not found in Kubernetes releases", out.V{"specified": nvs})
+		}
 	}
 
 	// If the version of Kubernetes has a known issue, print a warning out to the screen
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1406,6 +1406,10 @@ func validateKubernetesVersion(old *config.ClusterConfig) {
 	if err != nil {
 		exit.Message(reason.InternalSemverParse, "Unable to parse oldest Kubernetes version from constants: {{.error}}", out.V{"error": err})
 	}
+	newestVersion, err := semver.Make(strings.TrimPrefix(constants.NewestKubernetesVersion, version.VersionPrefix))
+	if err != nil {
+		exit.Message(reason.InternalSemverParse, "Unable to parse newest Kubernetes version from constants: {{.error}}", out.V{"error": err})
+	}
 	defaultVersion, err := semver.Make(strings.TrimPrefix(constants.DefaultKubernetesVersion, version.VersionPrefix))
 	if err != nil {
 		exit.Message(reason.InternalSemverParse, "Unable to parse default Kubernetes version from constants: {{.error}}", out.V{"error": err})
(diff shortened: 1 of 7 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -1419,6 +1423,10 @@ func validateKubernetesVersion(old *config.ClusterConfig) {
 		exitIfNotForced(reason.KubernetesTooOld, "Kubernetes {{.version}} is not supported by this release of minikube", out.V{"version": nvs})
 	}
 
+	if nvs.GT(newestVersion) {
```
medyagh, line 1426: thank you @JacekDuszenko for this PR ! while we don't support k8s versions newer than newest we still allow users to set a a Newer kubernetes version.
in this case I suggest in if the k8s version is Newer than our Support Newest version, we check the Kubernetes releases and see if that is an actual released version if not exit
if yes allow the user to do it at their own risk

medyagh (changes requested):

cmd/minikube/cmd/start.go
JacekDuszenko: @medyagh Can you please elaborate on how I should check the k8s releases? Is there a bit of code somewhere in minikube that already does that? If not, what's the best place to find releases? Should I make a REST call to some kind of a web service or rather try to parse kubernetes docs page where all releases all listed?

afbjorklund: Parsing the available versions sounds like a totally different feature. I just think you need to make version higher (like 1.26)

JacekDuszenko: @medyagh I found the code that fetches versions myself and implemented fetching and checking the version according to your proposal. This is updated and ready to be reviewed now

JacekDuszenko:
Updated the PR with versions fetching. Feel free to review and comment. Thanks!

cmd/minikube/cmd/start.go
JacekDuszenko: @afbjorklund thanks! So now we could either:
1. Withdraw parsing available versions feature and make version higher.
2. Leave parsing versions as they are now and make version higher.
3. Do not make the version higher as I think with fetching available releases and parsing them the issue is resolved? 

What should we do then?

afbjorklund: Parsing the available version is a good feature, just thought it was a _different_ one

afbjorklund: This needs to have some kind of offline override, when running in an air-gapped installation.

afbjorklund (changes requested):
It should be possible run minikube, without contacting outside resources.

cmd/minikube/cmd/start.go
```diff
@@ -1419,6 +1425,23 @@ func validateKubernetesVersion(old *config.ClusterConfig) {
...
+		k8sReleases, err := github.RecentK8sVersions()
+		if err != nil {
+			exit.Message(reason.KubernetesReleaseFetchFailed, "Failed fetching available Kubernetes releases from GitHub repository. Check your internet connection")
```
afbjorklund, line 1431: This needs to have some kind of offline override, when running in an air-gapped installation.

JacekDuszenko: How would you design this? We're not able to get the versions offline without hardcoding them in the code. But we can make this feature dependent on a configuration option (basically make a feature flag). The question is, is there already a config flag that specifies that minikube is running on an air-gapped environment? If yes, we'd just have to implement feature flag here. If not, we should probably add a config option? What do you reckon?

afbjorklund: Possibly re-use the force flag, which is the generic bail-out.

afbjorklund: `❗  minikube skips various validations when --force is supplied; this may lead to unexpected behavior`

JacekDuszenko: So if force flag is enabled, we skip just fetching releases, not the whole validation, right?

cmd/minikube/cmd/start.go
medyagh: @JacekDuszenko here is one place
https://github.com/medyagh/minikube/blob/5331cbeb8f1664ba69335e08e688d635c1767e37/hack/update/kubernetes_version/update_kubernetes_version.go#L148

JacekDuszenko: Thanks @medyagh. I've already found code that does that. It would be really useful for me to have you look at this chain of comments: https://github.com/kubernetes/minikube/pull/11631#discussion_r650674494 and let us know if we definitely want to go with the feature flag approach here. Thanks once again!

cmd/minikube/cmd/start.go
afbjorklund: There needs to be a way to skip all validation, in case GitHub is down or if you want to run a custom kubernetes

JacekDuszenko: So essentially you want to wrap this line: https://github.com/kubernetes/minikube/blob/d76259e86f4845bd01fe3e81e65214b9e231ecdb/cmd/minikube/cmd/start.go#L189
with a feature flag based on `--force` ?

Why have somebody not done that yet? 
I don't know if we should go that far in this PR. Validation is good even if we specify the `force` flag. Error will happen anyway but in the other place and with an obscure message if we turn off the validation.

I'd personally wrap just the release fetching inside the validation, because turning off that one particular validation really does matter in particular setting - while we have no internet connection.

JacekDuszenko: Also, tagging @medyagh, I think we might want to hear your opinion on that

medyagh: I agree with @afbjorklund we better make it work that if user already downloaded minikube and k8s preload there shouldnt be any need to make outgoing calls
so I suggest doing this
if Major Version is not "1"
then we would do a Github Release check, 
and if that failed we shoudl add a out.BoxedWarnning that the kuberentes verison u asked for doesnt seem to be right

but do not exit and let it continue

and also make sure this check does NOT happen for versions before our current NewestKubernetesVersion

medyagh:
Closing this or please recreate one ready

Issue #11512, opened 2021-05-26
new containerd config causes KVM Containerd test to time out and Docker Containerd test addons to fail


medyagh:
After New Containerd Config in this PR https://github.com/kubernetes/minikube/pull/11325/files

 the KVM test times out like this in 90 minutes (used to be Green in 63 minutes)

here is full KVM logs that timed out

[kvm_log.txt](https://github.com/kubernetes/minikube/files/6548314/kvm_log.txt)

Also the Docker Driver Test Addons Fail

https://storage.googleapis.com/minikube-builds/logs/11504/4567aaa/Docker_Linux_containerd.html#fail_TestAddons%2fparallel%2fRegistry

The integraiton tests run on Debian 9
```
jenkins@debian-jenkins-agent-9:~$ uname -r
4.9.0-15-amd64
```

### the New Containerd Config improves the perofmrance on cloud shell

interestingly after that PR the performance chart on cloud shell shows, containerd has same performance as Docker after that PR
<img width="1581" alt="Screen Shot 2021-05-26 at 9 46 59 AM" src="https://user-images.githubusercontent.com/4564227/119699747-54b30900-be07-11eb-9bda-9bd2a9884a06.png">

the cloud shell linux uses kernel 5
```
medya@cloudshell:~$ uname -r
5.4.104+
```

AkihiroSuda:
If the CI env has DOCKER_RAMDISK, it might not work as expected, though I assume it is no longer set

AkihiroSuda:
Removing NoPivotRoot = true may work?

medyagh:
> Removing NoPivotRoot = true may work?

do u mean for every case or only kernel 4 (debian 9)?

medyagh:
> If the CI env has DOCKER_RAMDISK, it might not work as expected, though I assume it is no longer set
doesn't seem like we have that ENV set
```
jenkins@debian-jenkins-agent-9:~$ echo $DOCKER_RAMDISK
jenkins@debian-jenkins-agent-9:~$ env | grep DOCKER
```

AkihiroSuda:
Basically, NoPivotRoot is insecure and should never be used on any version of kernel, unless it is really needed (for some initrd stuff).

I added NoPivotRoot in the v2 config PR just because v1 config had equivalent of that.

I'm not sure this is really related to the CI failure, but I think it is worth trying.

AkihiroSuda:
> echo $DOCKER_RAMDISK

[...]

PR #11632, opened 2021-06-11
Restore "containerd: upgrade io.containerd.runtime.v1.linux to io.containerd.runc.v2 (suppot cgroup v2)" #2

ilya-zuyev:
Another attempt to update containerd config


ilya-zuyev:
ok-to-build-iso

sharifelgamal:
Might fix https://github.com/kubernetes/minikube/issues/11512

ilya-zuyev:
integration tests `Test*BinaryUpgrade` are failing: 

` Jun 15 00:37:47 test-foo containerd[28515]: time="2021-06-15T00:37:47.951529929Z" level=error msg="RunPodSandbox for &PodSandboxMetadata{Name:kube-apiserver-test
-foo,Uid:022743265a20fd457df0a34731943068,Namespace:kube-system,Attempt:0,} failed, error" error="failed to create containerd task: runtime "io.containerd.runc.v
2" binary not installed "containerd-shim-runc-v2": file does not exist: unknown"`

test/integration/version_upgrade_test.go
```diff
@@ -125,7 +125,7 @@ func TestRunningBinaryUpgrade(t *testing.T) {
 		t.Fatalf("legacy %s start failed: %v", legacyVersion, err)
 	}
 
-	args = append([]string{"start", "-p", profile, "--memory=2200", "--alsologtostderr", "-v=1"}, StartArgs()...)
+	args = append([]string{"start", "-p", profile, "--memory=2200", "--delete-on-failure=true", "--alsologtostderr", "-v=1"}, StartArgs()...)
```
medyagh, line 128: lets not do "delet-on-failure"

medyagh (changes requested):

medyagh:
that is a very weird error: 
https://storage.googleapis.com/minikube-builds/logs/11632/fbc5a82/Docker_Linux.html#fail_TestRunningBinaryUpgrade
```
	* Selecting 'docker' driver from user configuration (alternates: [none])
-- /stdout --
** stderr ** 
	X The driver 'docker' is not supported on linux
** /stderr **
```

test/integration/version_upgrade_test.go
```diff
@@ -82,11 +82,13 @@ func TestRunningBinaryUpgrade(t *testing.T) {
...
 			// arm64 KIC driver is supported starting from v1.17.0
 			legacyVersion = "v1.17.0"
-		} else {
-			// v1.8.0 would be selected, but: https://github.com/kubernetes/minikube/issues/8740
-			legacyVersion = "v1.9.0"
 		}
 	}
+	// the version containerd in ISO was upgraded to 1.4.2
+	// we need it to use runc.v2 plugin
+	if ContainerRuntime() == "containerd" {
```
medyagh, line 89: this seems to be setting it globally for both KIC and VM drivers
previous logic has a different legacy per driver


since this is getting a complex, consider making a data structure to simplify it, for example how about  map. the map would look something like this
```
legacayMap = map[string][string] semver

legacayMap["KIC"] ["containerd"] = v1.X.X
legacayMap["VM"] ["Docker"] = v1.X.X
```
and that could be a constant in our constants package and be easily manintained

and then you could use a helper func to see if a driver is KIC or NOT

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -568,13 +568,20 @@ func (k *Bootstrapper) needsReconfigure(conf string, hostname string, port int,
 		klog.Infof("needs reconfigure: configs differ:\n%s", rr.Output())
 		return true
 	}
-
-	st, err := kverify.APIServerStatus(k.c, hostname, port)
+	// cruntime.Enable() may restart kube-apiserver but does not wait for it to return back
```
medyagh, line 571: could we add this additional check into cruntime.Enable() instead ?

pkg/minikube/cruntime/cruntime.go
```diff
@@ -163,6 +163,27 @@ type ListImagesOptions struct {
 // ErrContainerRuntimeNotRunning is thrown when container runtime is not running
 var ErrContainerRuntimeNotRunning = errors.New("container runtime is not running")
 
+// ErrRuntimeVersion is the error returned when disk image has incompatible version of service
+type ErrRuntimeVersion struct {
+	Service   string
```
medyagh, line 168: does this mean container runtime name ? if yes how about call it "Name"

pkg/minikube/cruntime/cruntime.go
```diff
@@ -163,6 +163,27 @@ type ListImagesOptions struct {
 // ErrContainerRuntimeNotRunning is thrown when container runtime is not running
 var ErrContainerRuntimeNotRunning = errors.New("container runtime is not running")
 
+// ErrRuntimeVersion is the error returned when disk image has incompatible version of service
```
medyagh, line 166: improve this comment, doesnt seem descriptive to me,

```
// ErrOutdatedVersion happens when base image or ISO have the unsupported container runtime See issue #1234 
type ErrOutdatedVersion
```

pkg/minikube/node/start.go
```diff
@@ -223,6 +229,14 @@ func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
 	return kcs, config.Write(viper.GetString(config.ProfileName), starter.Cfg)
 }
 
+func validateRuntimeVersion(cr cruntime.Manager) error {
```
medyagh, line 232: lets move this helper to cruntime package

so it will be
cruntime.ValidateVersion(cr) and then call cr.Version inside there
there shouldn't be crutnime logic code in node package

pkg/minikube/node/start.go
```diff
@@ -97,6 +97,12 @@ func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
 
 	// configure the runtime (docker, containerd, crio)
 	cr := configureRuntimes(starter.Runner, *starter.Cfg, sv)
+
+	// check if installed runtime is compatible with current minikube code
+	if err = validateRuntimeVersion(cr); err != nil {
```
medyagh, line 102: change this to 
```
if err = cruntime.ValidateVersion(cr); err != nil {
```

medyagh (changes requested):

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
medyagh: alternatively consider adding a new helper in kverify

kverify.APIServerStatusWithRetry(... 

so the logic here be simple and short

medyagh (changes requested):

pkg/minikube/node/start.go
ilya-zuyev: we already have such function - `cruntime.CompatibleWithVersion()` all cruntime specific logic is there. `validateRuntimeVersion` exists just to make the calling `Start(..)` method a bit more compact

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -568,13 +568,13 @@ func (k *Bootstrapper) needsReconfigure(conf string, hostname string, port int,
...
+	apiStatusTimeout := 1500 * time.Millisecond
```
medyagh, line 572: since this var is not used anywhere else why not pass it directly ?

pkg/minikube/node/start.go
medyagh: the problem is this validate runtime should be done by cruntime package, seems like we are passing func to func back to cruntime....let's do all of that in crunitime package, so we dont have do pass variables and also maintain code in "Node" package not related

medyagh (changes requested):

pkg/minikube/cruntime/cruntime.go
ilya-zuyev: It doesn't. It's any installed service with incompatible version

pkg/minikube/node/start.go
ilya-zuyev: What exactly is the problem here? What do you mean by `we are passing func to func`?

medyagh: the logic for runtimes needs to be in cruntime package and the the helper func you wrote has nothing that is related to "Node" Package.

so you are passing `cr cruntime.Manager` to node package
and then again from there you again call crutime package

instead of this, it is better the Whole logic be done in cruntime package, 


this code is Easier to write but it is harder to maintain, as someone who needs to maintain this code, please follow the  minikube packages and the reason they were meant to be

pkg/minikube/cruntime/cruntime.go
medyagh: fair enough

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
ilya-zuyev: First, it makes the next line shorter and easier to read. Another reason why I like this style is that a variable name gives a bit more context about the value, working like a comment. In this case variable's purpose is more or less obvious, but in general such approach helps to understand the meaning of parameters without looking into the target function.

medyagh:
@ilya-zuyev  the following tests fail on Docker on Docker 

i suspect it is related to this PR
30 | TestAddons/parallel/Ingress | 243.59
193 | TestRunningBinaryUpgrade | 5.79
194 | TestStoppedBinaryUpgrade | 66.15

pkg/minikube/assets/assets.go
```diff
@@ -0,0 +1,2621 @@
+// Code generated by go-bindata. (@generated) DO NOT EDIT.
```
medyagh, line 1: please remove these files these not meant to be in this PR (this is old asset files that been deleted)

pkg/minikube/translate/translations.go-e
```diff
@@ -0,0 +1,408 @@
+// Code generated by go-bindata. (@generated) DO NOT EDIT.
```
medyagh: please remove all of these go-bin data files from your PR those are from previous minikube, we removed the go-bindata

medyagh (changes requested):

pkg/minikube/cluster/pause.go
```diff
@@ -47,12 +47,10 @@ func pause(cr cruntime.Manager, r command.Runner, namespaces []string) ([]string
...
 	// Disable the kubelet so it does not attempt to restart paused pods
 	sm := sysinit.New(r)
-	if err := sm.Disable("kubelet"); err != nil {
-		return ids, errors.Wrap(err, "kubelet disable")
-	}
+	klog.Info("kubelet running: ", sm.Active("kubelet"))
 
-	if err := sm.Stop("kubelet"); err != nil {
-		return ids, errors.Wrap(err, "kubelet stop")
+	if err := sm.DisableNow("kubelet"); err != nil {
```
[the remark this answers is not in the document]
ilya-zuyev: https://www.freedesktop.org/software/systemd/man/systemctl.html
Please take a look.

`DisableNow` is an equivalent of `disable` + `stop`. This change replaces two calls (Disable + Stop) with one - 
DisableNow

medyagh:
@ilya-zuyev api health failure is saying 500 error I dont think increasing time limit would fix

consider rebasing the  40 commits for a cleaner history, only keep the relevant commits that actually made it it to the PR and remove debug ones

pkg/minikube/cluster/pause.go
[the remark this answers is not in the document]
medyagh: I undrestand DisableNow does but I am asking, is Before this PR
it was only Stop and Not Disable
after this PR is both ! which means on the Next Restart of the container the "kubelet" will not Auto-Start and will stay disabled.

why this PR needs to do that ?  and before that we didnt have to do that?

medyagh (changes requested):

pkg/minikube/cluster/pause.go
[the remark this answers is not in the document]
ilya-zuyev: We did disable, actually. See removed line 50 (red one)

test/integration/version_upgrade_test.go
ilya-zuyev: Checked this test for [driver=docker, runtime=containerd] config. It also fails if legacy version is less than 1.16.0. It looks like for conainerd we have to use the same legacy for both kvm and kicbase

pkg/minikube/cluster/pause.go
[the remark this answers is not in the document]
medyagh: sorry my bad you are right

medyagh (approved):
Thank you for the persistence that went through making this work !

good work !

PR #11653, opened 2021-06-14
Fix a download link to use arm64 instead of amd64

dongjoon-hyun:
Currently, the error message is correct, but the download link is still pointing `amd64`. We need to download `arm64` binary.
```
$ minikube version --short

❌  Exiting due to MK_WRONG_BINARY_M1: You are trying to run amd64 binary on M1 system.
Please use darwin/arm64 binary instead
(Download at https://github.com/kubernetes/minikube/releases/download/v1.21.0/minikube-darwin-amd64.)
```


ilya-zuyev (approved):

dongjoon-hyun:
Thank you, @ilya-zuyev !

PR #11663, opened 2021-06-15
add local-kicbase make target

### LINKED ISSUE #10670: make target for local kicbase
For local testing, there should be a make target `make local-kic-base` that builds  updates the types.go in kic package
with unixtime

local/kicbase-unixtmie

sharifelgamal:
this also adds the local-kicbase-debug target, that will automatically change minikube to point to the newly built local kicbase image

Closes #10670

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ KVM_GO_VERSION ?= $(GO_VERSION:.0=)
 
 INSTALL_SIZE ?= $(shell du out/minikube-windows-amd64.exe | cut -f1)
 BUILDROOT_BRANCH ?= 2020.02.12
-REGISTRY?=gcr.io/k8s-minikube
+REGISTRY ?=gcr.io/k8s-minikube
 
 # Get git commit id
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -40,7 +40,7 @@ KVM_GO_VERSION ?= $(GO_VERSION:.0=)
 
 INSTALL_SIZE ?= $(shell du out/minikube-windows-amd64.exe | cut -f1)
 BUILDROOT_BRANCH ?= 2020.02.12
-REGISTRY?=gcr.io/k8s-minikube
+REGISTRY ?=gcr.io/k8s-minikube
```
spowelljr, line 43: Don't think it matters at all, but since you added the space before, this is the only one that doesn't have a space after `?=` as well.

sharifelgamal: yeah it's purely aesthetic change, good call

sharifelgamal: fixed

spowelljr (approved):

Issue #11622, opened 2021-06-10
Fix "Failed to stop docker.service: Unit docker.service not loaded" log error on start with kvm2 driver and containerd/crio runtimes


ilya-zuyev:
when minikube is starting we try to disable non-active runtimes , but for kvm2/containerd combination it leads to a log error
`E0610 10:36:19.140582   12060 docker.go:159] "Failed to stop" err="sudo systemctl stop -f docker.service: Process exited with status 5\nstdout:\n\nstderr:\nFailed to stop docker.service: Unit docker.service not loaded.\n" service="docker.service"
`
```
ilyaz@skeletron --- g/minikube ‹master* ?› » out/minikube start --driver=kvm2 --container-runtime=containerd                                                                                                                                         7 ↵
* minikube v1.21.0-beta.0 on Ubuntu 21.04
* Using the kvm2 driver based on existing profile
* Starting control plane node minikube in cluster minikube
* Restarting existing kvm2 VM for "minikube" ...
E0610 10:36:19.140582   12060 docker.go:159] "Failed to stop" err="sudo systemctl stop -f docker.service: Process exited with status 5\nstdout:\n\nstderr:\nFailed to stop docker.service: Unit docker.service not loaded.\n" service="docker.service"
* Preparing Kubernetes v1.20.7 on containerd 1.4.4 ..
* ```

afbjorklund:
Docker was normally _always_ started, due to using Docker Machine.

PR #11667, opened 2021-06-15
Do not return an error from Systemd.ForceStop(svc) if svc is already stopped

ilya-zuyev:
Fixes https://github.com/kubernetes/minikube/issues/11622


pkg/minikube/sysinit/systemd.go
```diff
@@ -123,7 +125,14 @@ func (s *Systemd) Stop(svc string) error {
 
 // ForceStop terminates a service with prejudice
 func (s *Systemd) ForceStop(svc string) error {
-	_, err := s.r.RunCmd(exec.Command("sudo", "systemctl", "stop", "-f", svc))
+	rr, err := s.r.RunCmd(exec.Command("sudo", "systemctl", "stop", "-f", svc))
+	if err == nil {
```
sharifelgamal, line 129: this if statement isn't really necessary, `return err` will return nil if err is nil

medyagh: agree with @sharifelgamal  this "if" is not needed !

pkg/minikube/sysinit/systemd.go
```diff
@@ -123,7 +125,14 @@ func (s *Systemd) Stop(svc string) error {
 
 // ForceStop terminates a service with prejudice
 func (s *Systemd) ForceStop(svc string) error {
```
medyagh, line 127: consider adding this to openRC too
https://github.com/medyagh/minikube/blob/1bac30bbe7caa5b8c64b35cfb97c197ffe08840a/pkg/minikube/sysinit/openrc.go#L166

medyagh (changes requested):

pkg/minikube/sysinit/systemd.go
ilya-zuyev: This `if` lets us to avoid matching substrings in `strings.Contains`. Also the whole logic is clearer with this shortcut - if `systemctl` returned ok - return immediately as no extra processing required

pkg/minikube/sysinit/systemd.go
ilya-zuyev: Not sure what to do there. This PR is targeting a very specific systemd issue

pkg/minikube/sysinit/systemd.go
medyagh: this "if statement" adds to the code complexity in terms of number of If statments. and makes the code harder to read.

this is not inline with go style used in minikube, I suggest getting rid of it

medyagh (changes requested):

pkg/minikube/sysinit/systemd.go
ilya-zuyev: just curious, how it makes the code harder to read?

medyagh: because following the code we dont expect to check for err nil and return while it already returns,
a typical go reader with following the styles is used to look at the return and it is obvisous if it is nil it returns nil too,

plus it adds to the go "gocyclo" score https://goreportcard.com/report/github.com/kubernetes/minikube
so the less If statments, the better

ilya-zuyev: `because following the code we dont expect to check for err nil and return while it already returns,
a typical go reader with following the styles is used to look at the return and it is obvisous if it is nil it returns nil too.`  I don't understand this sentence

Cyclomatic complexity metric evaluates the number of branches inside the function, it's goal is to prevent functions overgrow. This method is 7 lines long, talking about cyclomatic complexity makes no sense here.

medyagh: this is not inline with minikube code style, it doesn't make sense to check for "nil" and return while we are returning in next line, this an example of code seems easier to write but harder to maintain 

I suggest reading more minikube code or  https://www.kubernetes.dev/docs/guide/coding-convention/

medyagh (approved):

medyagh:
plz pull upstream and then it is good to go

pkg/minikube/sysinit/systemd.go
sharifelgamal: I don't personally think this is more complex, it's just an awkward line of code for a small optimization.

medyagh:
@ilya-zuyev 
this PR doesnt seem to fix the failed test
https://storage.googleapis.com/minikube-builds/logs/11667/5a3ad1e/KVM_Linux_containerd.html#fail_TestErrorSpam%2fsetup
```
error_spam_test.go:108: minikube stderr:
E0622 01:56:15.187503    9483 docker.go:159] "Failed to stop" err="sudo systemctl stop -f docker.service: Process exited with status 5\nstdout:\n\nstderr:\nFailed to stop docker.service: Unit docker.service not loaded.\n" service="docker.service"
```

PR #11671, opened 2021-06-16
Use goembed instead of go-bindata for embedding addon config files

### LINKED ISSUE #11664: Use goembed instead of go-bindata for embedding addon files.
goembed is being used now in gopogh due to https://github.com/medyagh/gopogh/pull/30. There it was replacing rice, but goembed would likely also improve our current setup, since assets.go is kind of a mess.

andriyDev:
fixes #11664.

### DIFF
--- a/deploy/addons/assets.go
+++ b/deploy/addons/assets.go
@@ -1,0 +1,135 @@
+/*
+Copyright 2021 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package addons
+
+import "embed"
+
+var (
+	// AutoPauseAssets assets for auto-pause addon
+	//go:embed auto-pause/*.tmpl
+	//go:embed auto-pause/auto-pause.service
+	//go:embed auto-pause/unpause.lua
+	AutoPauseAssets embed.FS
+
+	// DashboardAssets assets for dashboard addon
+	//go:embed dashboard/*.yaml dashboard/*.tmpl
+	DashboardAssets embed.FS
+
+	// DefaultStorageClassAssets assets for default-storageclass addon
+	//go:embed storageclass/storageclass.yaml.tmpl
+	DefaultStorageClassAssets embed.FS
+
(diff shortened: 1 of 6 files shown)


andriyDev:
Measuring the binary size before and after on my machine:
| |Bytes|
|---|---|
|Before|78048|
|After|63392|

This is at least partly due to the fact that the new `assets.go` file does not include every file in the deploy/assets folder - only those actually used by the addons.

This also seems to fix a previous circular dependency in the old `assets.go` since the auto-pause-hook and it depended on each other. With no generated `assets.go`, this dependency is resolved.

deploy/addons/assets.go
```diff
@@ -0,0 +1,135 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package addons
+
+import "embed"
+
+var (
```
andriyDev, line 21: We could instead have one BIG embed.FS for all the files. I feel however this could be confusing as to what addons need what files, or if all deploy/addons files are included, we could include unnecessary files (like the old system).

andriyDev:
Unsure whether we should remove `make clean` entry for `assets.go`.We are no longer generating this file, but also make clean should really get rid of everything. In addition, the size improvements aren't realized with the old `assets.go` present.

Similarly, should we remove the entry from .gitignore to stop ignoring assets.go? If so, people might start adding it by accident everywhere which would be quite annoying. I originally decided to go with not cleaning it, but still ignoring it - probably the wrong decision actually.

medyagh:
> we remove the entry from .gitignore to stop ignoring assets.go? If so, people might start adding it by accident everywhere which would be quite annoying. I originally decided to go with not clean

I think once they pull the new .gitignore it should delete those files for you ? probably better to have it removed everywhere including gitignore

and yes make clean should nuke everything

medyagh:
@andriyDev any data on minikube binary size ?
can we use this for the translation files too ?

andriyDev:
@medyagh I wrote a comment up above about it! It does improve the binary size, at least with the better control of what to include.

I haven't looked into the translation files yet, but I was going to after.

deploy/addons/assets.go
```diff
@@ -0,0 +1,135 @@
...
+	// AutoPauseAssets assets for auto-pause addon
+	//go:embed auto-pause/*.tmpl
+	//go:embed auto-pause/auto-pause.service
+	//go:embed auto-pause/unpause.lua
+	AutoPauseAssets embed.FS
```
medyagh, line 26: why this vars are not used anywhere?

andriyDev: They are used in `pkg/minikube/assets/addons.go`.

medyagh (approved):
thank you for this PR ! look forward to see everything else be migrated to go embed

sharifelgamal (approved):

PR #11673, opened 2021-06-16
Restore previous support for progressbar pool

### LINKED ISSUE #11069: UI progress bars stepping on each other
I noticed that `preloaded-images-k8s` and `index.docker.io/kicbase` progress bars step on each other when downloading is in progress and it is confusing for the user and might be interpreted as a glitch. When the download is finished, they are displayed in separate lines:

```
😄  minikube v1.19.0 on Darwin 10.15.7
🚜  Pulling base image ...
    > gcr.io/k8s-minikube/kicbase...: 357.67 MiB / 357.67 MiB  100.00% 396.74 K
    > preloaded-images-k8s-v10-v1...: 491.71 MiB / 491.71 MiB  100.00% 411.47 K
    > index.docker.io/kicbase/sta...: 357.67 MiB / 357.67 MiB  100.00% 417.22 K
```

afbjorklund:
Now shows output on two _different_ lines instead of just one:

```
🚜  Pulling base image ...
💾  Downloading Kubernetes v1.20.7 preload ...
    > preloaded-images-k8s-v11-v1...:  492.20 MiB / 492.20 MiB  100.00% 22.08 MiB/s 22s01s
    > gcr.io/k8s-minikube/kicbase...:  359.09 MiB / 359.09 MiB  7.50 MiB/s 47s 
```

Had to downgrade github.com/cheggaaa/pb from V3 to V1 for it.

Pool was removed earlier in 7fa4e054f3f6b312572598f5dde0c681dfd6844b

Closes #11069


afbjorklund:
Apparently this might break things on Windows, so should disable it there.

https://github.com/kubernetes/minikube/issues/4996#issuecomment-519094421

medyagh:
> #4996 (comment)

hm...how about we use V3 in windows ? ( I think any progress bar is better than no progress bar ? ) we had a lot of users complanining they hate to wait looking at nothing when somehting was being downloaded

afbjorklund:
I was just wondering how bad V1 was on Windows, without the pool...

But the progressbar should probably be abstracted into a separate class.
Then there could be different implementations for Unix and for Windows.

medyagh:
> I was just wondering how bad V1 was on Windows, without the pool...
> 
> But the progressbar should probably be abstracted into a separate class.
> Then there could be different implementations for Unix and for Windows.

good idea

afbjorklund:
So a new PR for that.

PR #11679, opened 2021-06-17
ISO: Add packaging for crun

afbjorklund:
https://github.com/containers/crun


ilya-zuyev:
Why do we need crun?

afbjorklund:
CRI-O defaults to crun under certain conditions, mostly Cgroups V2 I think

medyagh:
ok-to-build-iso

afbjorklund:
ok-to-build-iso

afbjorklund:
ok-to-build-iso

afbjorklund:
ok-to-build-iso

spowelljr:
Is this something we're still actively pursuing as well? `crun` has made a stable release it seems with `1.1`, but I'm happy to review this if this PR is still relevant.

afbjorklund:
KIC is currently including crun 1.2 from kubic, hopefully both are compatible with podman ?

```
      crun |  100:1.2-2 | http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04  Packages
```

afbjorklund:
ok-to-build-iso

spowelljr:
ok-to-build-iso

spowelljr (approved):

jesperpedersen:
Any work planned for crun 1.2+, like crun 1.4.4 ?

afbjorklund:
Currently there are no Ubuntu 20.04 binaries after Podman 3.4.2 and crun 1.2, so left it the same in the ISO... Might have to use binaries also in KIC, if we want to bump to 3.4.5 and 1.4.3

Is there anything important?

jesperpedersen:
@afbjorklund Ok, good to know -- just interested in when the fixed CVEs were going to be included

afbjorklund:
It seems like podman gave up on packaging mid-air, so I think we will have to use some PPA or do our own packages

https://build.opensuse.org/project/show/devel:kubic:libcontainers:stable

jesperpedersen:
I think it is left to each OS vendor, because

https://koji.fedoraproject.org/koji/packageinfo?packageID=9688

afbjorklund:
minikube uses podman for building images and for loading images, so normally it is cri-o `runc` that is the runtime.

Using minikube to run containers with podman is probably unsupported. Can use podman machine for that instead.

afbjorklund:
> I think it is left to each OS vendor, because
> 
> https://koji.fedoraproject.org/koji/packageinfo?packageID=9688

Yeah well, it seems that it would require changing distribution

https://packages.ubuntu.com/focal/crun

jesperpedersen:
Note, https://podman.io/blogs/2022/03/06/why_no_podman4_f35.html

Issue #11650, opened 2021-06-14
timeToK8s: Give minikube the same amount of memory as kind and k3d


sharifelgamal:
By default, kind and k3d do not have a set limit on the amount of memory they use. minikube by default decides how much memory to allocate, this difference in available memory puts minikube at a disadvantage. Give minikube unlimited or a large amount of memory during the time-to-k8s benchmark to make the results more comparable.

spowelljr:
In-progress

PR #11692, opened 2021-06-17
Add ability to pass 'max' value to memory and cpus flags

spowelljr:
Part of #11650

**Problem:**
There's currently no way to automatically maximize memory and cpu allocation.

**Solution:**
Added the ability to pass `mamx` to the `memory` and `cpus` flags on `minikube start`. Passing that value will allocate maximum resources to minikube.

Example:
```
minikube start --memory=max cpus=max
...
🔥  Creating docker container (CPUs=6, Memory=5945MB) ...
```


cmd/minikube/cmd/start_flags.go
```diff
@@ -135,8 +135,8 @@ func initMinikubeFlags() {
 	startCmd.Flags().Bool(interactive, true, "Allow user prompts for more information")
 	startCmd.Flags().Bool(dryRun, false, "dry-run mode. Validates configuration, but does not mutate system state")
 
-	startCmd.Flags().Int(cpus, 2, "Number of CPUs allocated to Kubernetes.")
-	startCmd.Flags().String(memory, "", "Amount of RAM to allocate to Kubernetes (format: <number>[<unit>], where unit = b, k, m or g).")
+	startCmd.Flags().String(cpus, "2", "Number of CPUs allocated to Kubernetes. Use 'nolimit' to use the maximum number of CPUs.")
```
medyagh, line 138: use the constant in the string, so if we change it it changes too

pkg/minikube/constants/constants.go
```diff
@@ -114,6 +114,8 @@ const (
 
 	// TimeFormat is the format that should be used when outputting time
 	TimeFormat = time.RFC1123
+	// NoLimit is the value that can be passed into the memory and cpus flags to specify to use maximum resources
+	NoLimit = "nolimit"
```
medyagh, line 118: how about `no-limit` or maybe even `max`

medyagh (changes requested):

pkg/minikube/constants/constants.go
spowelljr: Changed to max, much more instinctive name for it

cmd/minikube/cmd/start_flags.go
spowelljr: Done

medyagh (commented):
if this is a VM driver user sets --CPUs=Max what happens?

medyagh:
as part of this PR I think I was expecting to see the oci package be touched in a way that this line should NOt be appended (and leave it without --memory

```
		if memcg {
			runArgs = append(runArgs, fmt.Sprintf("--memory=%s", p.Memory))
		}


```
https://github.com/medyagh/minikube/blob/00cc926649f84edf2dc5b4a202ceaafb7a79e727/pkg/drivers/kic/oci/oci.go#L212

medyagh:
could we see metrics of time2k8s after this PR ? does it make any diff?

spowelljr:
> could we see metrics of time2k8s after this PR ? does it make any diff?

The `minikube start` command completes 2.7 seconds faster with `--memory=max --cpus=max` flags used.

medyagh (approved):

Issue #11207, opened 2021-04-26
Periodically tell user about minikube features/tips and tricks


medyagh:
we should have a Random Tips that on each time they start minikube it will tell them about a feature of minikube...(the Tips of the day should be a way that u could disable it)

```
$ minikube start
😄  minikube v1.19.0 on Darwin 11.2.3
✨  Using the docker driver based on existing profile
👍  Starting control plane node minikube in cluster minikube
🏃  Updating the running docker "minikube" container ...
🐳  Preparing Kubernetes v1.20.2 on Docker 20.10.5 ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: storage-provisioner, default-storageclass
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default

----- Tip of the day ---------------------------------------------------------------------------
|  Some Fancy fox UI: 
|  Did you know you can specify any kubernetes version you want ? for example minikube start --kubernetes-version=v1.19.0
|
| to disable this Tips Run .... minikube config set BLAHBLAH
-----------------------------------------------------------------------------------------------
```

or 

----- Tip of the day ---------------------------------------------------------------------------
Did you know you can enroll getting notifcation for beta releases by minkube config set .... ?

| to disable tips of the day run .... minikube config set BLAHBLAH
-----------------------------------------------------------------------------------------------

afbjorklund:
We can add more styles to the out.Boxed perhaps, I sort of hardcoded the red "warning" one for now:

Type: "Round"
Color = "Red"

See https://github.com/Delta456/box-cli-maker#box-types

medyagh:
this task is available to pick up

dinever:
Working on this next. Will share a proposal once I have it. Thanks!

/assign

sharifelgamal:
@dinever still working on this?

dinever:
> @dinever still working on this?

Hi @sharifelgamal, yes I'm still working on it. Had some proof of concept working at local, will open an initial PR for review this week. Thanks!

sharifelgamal:
Hey @dinever, looks like we approved your proposal and initial refactor PR. Any plans on continuing to work on this?

PR #11696, opened 2021-06-18
Proposal for minikube tips

dinever:
Hi,

This is a proposal for #11207. Would like to collect some feedback before submitting any PRs.

Thanks!

### DIFF
--- a/enhancements/proposed/20210618-tips/tips.md
+++ b/enhancements/proposed/20210618-tips/tips.md
@@ -1,0 +1,215 @@
+# Periodically tell user about minikube features/tips and tricks
+
+* First proposed: 2021-06-18
+* Authors: Peixuan Ding (@dinever)
+
+## Reviewer Priorities
+
+Please review this proposal with the following priorities:
+
+*   Does this fit with minikube's [principles](https://minikube.sigs.k8s.io/docs/concepts/principles/)?
+*   Are there other approaches to consider?
+*   Could the implementation be made simpler?
+*   Are there usability, reliability, or technical debt concerns?
+
+Please leave the above text in your proposal as instructions to the reader.
+
+## Summary
+
+minikube has lots of great features. We want to proactively remind users that
+those features are available.
+
+To achieve this, we can have a tips feature that randomly shows a tip
+from a curated list whenever the user starts a new minikube profile.
+
+For example:
+
+![Screenshot from 2021-06-18 00-58-02](https://user-images.githubusercontent.com/1311594/122508665-53bd6380-cfd0-11eb-9e99-a6c5935514d5.png)
+
+## Goals
+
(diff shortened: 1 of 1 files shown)


enhancements/proposed/20210618-tips/tips.md
```diff
@@ -0,0 +1,215 @@
...
+1. `out.Boxed` with custom style
+2. random `tips` display with ability to disable through config, with an initial set of about 10 tips
+3. `make generate-docs` to sync tips to docs
+4. Add more tips
+
+## Alternatives Considered
+
+1. Is there a more preferred file format to YAML?
+
+2. Maybe we just want to sync the tips to the `FAQ` page list instead of creating a new page?
```
andriyDev, line 181: I think the original plan was to sync with FAQ but I think new page is the way to go, FAQ should be specifically for answering questions to common problems, tips should be interesting features people might want to try out.

enhancements/proposed/20210618-tips/tips.md
```diff
@@ -0,0 +1,215 @@
...
+      - question: How do I access the Kubernetes Dashboard UI?
+        answer: |   
+          minikube has the built-in Kubernetes Dashboard UI. To access it:
+
+          ```
+          minikube dashboard
+          ```
+    ```
+
+   On the docs side we should both questions and answers. On the CLI side
```
andriyDev, line 210: Interesting idea! I think wording that correctly can be very tricky though, so it makes sense as a tip and as an answer. I'm a much bigger fan of just a set of tips.

enhancements/proposed/20210618-tips/tips.md
```diff
@@ -0,0 +1,215 @@
...
+I plan to open at least 4 PRs:
+
+1. `out.Boxed` with custom style
+2. random `tips` display with ability to disable through config, with an initial set of about 10 tips
+3. `make generate-docs` to sync tips to docs
+4. Add more tips
+
+## Alternatives Considered
+
+1. Is there a more preferred file format to YAML?
```
andriyDev, line 179: The only other file format we might use is JSON, and while I do like the JSON structure, it has a big flaw which is that new lines must be replaced with \n which can get messy, and make it hard to see how the tip actually looks. I'm pretty happy with YAML for the file format because it makes it very clear how it looks.

andriyDev (approved):
I like this proposal - simple, clean, good. Very well done!

enhancements/proposed/20210618-tips/tips.md
dinever: Thanks for the feedback @andriyDev! Yep it's kind of hard to fit these tips with the current FAQ items. Will bring this up in office hour and see what everyone thinks.

PR #11703, opened 2021-06-19
Add Autopause Time parameter to addons

### LINKED ISSUE #10595: auto-pause: make time to pause configurable

vishjain:
fixes #10595

### DIFF
--- a/cmd/auto-pause/auto-pause.go
+++ b/cmd/auto-pause/auto-pause.go
@@ -23,6 +23,7 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/minikube/pkg/addons"
 	"k8s.io/minikube/pkg/minikube/cluster"
 	"k8s.io/minikube/pkg/minikube/command"
 	"k8s.io/minikube/pkg/minikube/cruntime"
@@ -43,9 +44,6 @@ var version = "0.0.1"
 var runtime = "docker"
 
 func main() {
-	// TODO: #10595 make this configurable
-	const interval = time.Minute * 1
-
 	// Check current state
 	alreadyPaused()
 
@@ -55,7 +53,7 @@ func main() {
 			// On each iteration new timer is created
 			select {
 			// TODO: #10596 make it memory-leak proof
-			case <-time.After(interval):
+			case <-time.After(addons.AutoPauseTime):
 				runPause()
 			case <-unpauseRequests:
 				fmt.Printf("Got request\n")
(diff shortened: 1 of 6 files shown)


medyagh (commented):
I am not sure if this is the right place to add the config, would u please add the example usage of this after this PR with the output ?

cmd/minikube/cmd/config/configure.go
```diff
@@ -231,7 +233,14 @@ var addonsConfigureCmd = &cobra.Command{
 			if err := config.SaveProfile(profile, cfg); err != nil {
 				out.ErrT(style.Fatal, "Failed to save config {{.profile}}", out.V{"profile": profile})
 			}
+		case "auto-pause":
+			pauseTimeStr := AskForStaticValue("-- Enter interval time to pause (in minutes): ")
+			pauseTime, err := strconv.Atoi(pauseTimeStr)
```
ilya-zuyev, line 238: Please validate that `pauseTime` is non-negative

pkg/addons/addons.go
```diff
@@ -54,6 +54,10 @@ var Force bool = false
 // Currently only used for gcp-auth
 var Refresh bool = false
 
+// AutoPauseTime is used to specify how often minikube needs to check to autpause
+// cluster.
+var AutoPauseTime time.Duration = time.Minute * 1
```
andriyDev, line 59: AutoPauseTime is not persisted anywhere. This means whenever the minikube binary starts, this will be reset to 1 minute. You need to store this in the Kubernetes config.

cmd/auto-pause/auto-pause.go
```diff
@@ -55,7 +53,7 @@ func main() {
 			// On each iteration new timer is created
 			select {
 			// TODO: #10596 make it memory-leak proof
-			case <-time.After(interval):
+			case <-time.After(addons.AutoPauseTime):
```
andriyDev, line 56: auto-pause runs on the guest, so the settings on the host will NOT propagate here. We need to pass this in through the command line arguments for auto-pause.

andriyDev (changes requested):

cmd/auto-pause/auto-pause.go
vishjain: Made it so that it should be configurable via the configurate command. @andriyDev

medyagh (commented):
@vishjain do u mind adding in the PR Description how u use this with the output?

vishjain:
Closing this PR, and putting a new updated PR.

vishjain:
> @vishjain do u mind adding in the PR Description how u use this with the output?

Addressed in other pr. thanks!

PR #11708, opened 2021-06-20
Make sure to log errors from DaemonInfo

afbjorklund:
Currently hiding errors, from running or parsing output from `system info`

### DIFF
--- a/pkg/drivers/kic/oci/info.go
+++ b/pkg/drivers/kic/oci/info.go
@@ -252,9 +252,11 @@ func dockerSystemInfo() (dockerSysInfo, error) {
 	var ds dockerSysInfo
 	rawJSON, err := dockerInfoGetter()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/drivers/kic/oci/info.go
```diff
@@ -252,9 +252,11 @@ func dockerSystemInfo() (dockerSysInfo, error) {
 	var ds dockerSysInfo
 	rawJSON, err := dockerInfoGetter()
 	if err != nil {
+		klog.Errorf("docker info: %v", err)
```
medyagh, line 255: change to warningf to avoid poluting the UI

medyagh (changes requested):

pkg/drivers/kic/oci/info.go
afbjorklund: ok

mahalrs (approved):
/approve

PR #11709, opened 2021-06-21
Add method for customized box output

dinever:
Part of #11207

This gives developers the ability to print out customized box output.

For example:

```go
boxCfg := box.Config{
	Py:       1,
	Px:       2,
	TitlePos: "Inside",
	Type:     "Double",
	Color:    "Yellow",
}

out.BoxedWithConfig(boxCfg, style.Warning, "Warning", "This is a warning")
```
Screenshot from 2021-06-20 23-37-26

### DIFF
--- a/pkg/minikube/out/out.go
+++ b/pkg/minikube/out/out.go
@@ -68,6 +68,8 @@ var (
 	JSON = false
 	// spin is spinner showed at starting minikube
 	spin = spinner.New(spinner.CharSets[style.SpinnerCharacter], 100*time.Millisecond)
+	// defaultBoxCfg is the default style config for cli box output
+	defaultBoxCfg = box.Config{Py: 1, Px: 4, Type: "Round", Color: "Red"}
 )
 
 // MaxLogEntries controls the number of log entries to show for each source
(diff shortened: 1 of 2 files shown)


pkg/minikube/out/out.go
```diff
@@ -113,28 +115,38 @@ func Styled(st style.Enum, format string, a ...V) {
...
 
-// BoxedErr writes a stylized and templated message in a box to stderr
+// BoxedErr writes a stylized and templated message in a box to stderr using the default style config
 func BoxedErr(format string, a ...V) {
-	boxedCommon(Err, format, a...)
+	boxedCommon(Err, defaultBoxCfg, "", format, a...)
+}
+
+// BoxedWithConfig writes a templated message in a box with customized style config to stdout
+func BoxedWithConfig(cfg box.Config, st style.Enum, title string, format string, a ...V) {
```
medyagh, line 143: this is currently not being used anywhere in the code, and the rule of thumb is not to keep code never used anywhere...how about find a good place/example for this func to be used ?

dinever: Yes this part of the `minikube tips` feature. Please see my proposal: https://github.com/kubernetes/minikube/pull/11696

I was trying to create smaller PRs for easier review. But I can have all the changes in one PR as well if it makes more sense.

medyagh (approved):
Thank you @dinever for this pr look forward to see more

PR #11721, opened 2021-06-21
let windows users use the LC_ALL env var to set locale

sharifelgamal:

### DIFF
--- a/pkg/minikube/translate/translate.go
+++ b/pkg/minikube/translate/translate.go
@@ -19,6 +19,8 @@ package translate
 import (
 	"encoding/json"
 	"fmt"
+	"os"
+	"runtime"
 	"strings"
 
 	"github.com/cloudfoundry-attic/jibber_jabber"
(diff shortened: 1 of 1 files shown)


pkg/minikube/translate/translate.go
```diff
@@ -60,10 +62,20 @@ func T(s string) string {
...
 // DetermineLocale finds the system locale and sets the preferred language for output appropriately.
 func DetermineLocale() {
-	locale, err := jibber_jabber.DetectIETF()
-	if err != nil {
-		klog.V(1).Infof("Getting system locale failed: %v", err)
-		locale = ""
+	var locale string
+	// Allow windows users to overload the same env vars as unix users
+	if runtime.GOOS == "windows" {
+		if os.Getenv("LC_ALL") != "" {
```
spowelljr, line 68: A nit, but isn't the `if os.Getenv("LC_ALL") != ""` redundant since if it is empty and we set it to `locale` it will go into the following `if` anyway?

sharifelgamal: that's a good point! I'll get rid of the if.

spowelljr (approved):

PR #11727, opened 2021-06-22
ISO: Upgrade podman to 3.1.2

### LINKED ISSUE #11706: Stop using podman2 for loading and building images for cri-o
We need to upgrade from **podman2** (2.0/2.1/2.2) to one of the **podman3** (3.0/3.1/3.2)

Currently using 2.2.1, and upgrading to 3.1.2. The minor versions are not compatible.

Eventually it should be possible to upgrade to 3.2.1, but it is still in "testing" (not "stable")

* https://build.opensuse.org/package/show/devel:kubic:libcontainers:stable/podman

* https://build.opensuse.org/package/show/devel:kubic:libcontainers:testing/podman

See https://podman.io/getting-started/installation#linux-distributions for details

RHEL versions (of Podman):
* v1.6.4
* v2.0.5
* v2.1.1
* v2.2.1
* v3.0.1

It is possible to use a newer version of `podman` for running with the podman driver.
But the version of `podman-remote` needs to match what is running on the machine.

Development is now happening on 3.3.0-dev, which might even become Podman 4.0.
(Podman 3.0 was the _final_ release for older distributions*, like CentOS 7, Ubuntu 18.04)

\* see ~~#10937~~

afbjorklund:
Note: podman3 requires go to be updated first (new OS)

`build flag -mod=vendor only valid when using modules`

Closes #11706


medyagh:
ok-to-build-iso

medyagh:
ok-to-build-iso

afbjorklund:
As stated above, this requires the OS to be updated first

afbjorklund:
ok-to-build-iso

spowelljr:
@afbjorklund Based on https://github.com/kubernetes/minikube/issues/11706#issuecomment-902911545, is this PR something we still want to pursue? And if so, does anything else still have to be done? https://github.com/kubernetes/minikube/pull/11727#issuecomment-889970284?

afbjorklund:
Well, there is a 3.4.0 now. They say it will be stable. Except of course that `podman machine` still has a broken ISO, etc

afbjorklund:
Going with this 3.1.2 would work, perhaps drop back to 3.0.1 if there are any issues with it ?

Bumping to 3.3.1 should also be possible, but there is no _urgency_ for minikube to do it

spowelljr:
ok-to-build-iso

marchmallow:
Waiting for this too.. I am trying to migrate local development stack from docker desktop + minikube with docker env to podman + minukube with podman env but blocked by having podman 3.x version installed locally vs. minikube podman env on older 2.x version

marchmallow:
Happy to help if anything needed to get this ready

afbjorklund:
The podman installation in minikube is mostly for loading images into cri-o, it works poorly for running containers with.

afbjorklund:
> Many people and organizations are using Minikube as a Docker Desktop replacement and are opting for Podman/CRI-O instead of Docker daemon as backend, I think it would be good to keep Podman versions up-to-date, although it is not the primary use case of Minikube. That said, jumping to any version of Podman 3 ASAP would be a good start :)

Red Hat have their own VM solution for running Podman on macOS, so you might want to use that instead.

albertdb:
> The podman installation in minikube is mostly for loading images into cri-o, it works poorly for running containers with.

Not really. The only issue is Podman 2.x due to lack of docker compose support, otherwise runs nice :)


> Red Hat have their own VM solution for running Podman on macOS, so you might want to use that instead.

Not a macOS user.

afbjorklund:
> Not a macOS user.

Ok, I think they recommend WSL for Windows and running directly on the host for Linux.

But anyway, I will try to remove the bot crap from the PR and see if it still builds OK...

afbjorklund:
We could also try a newer version, unfortunately they are not fully compatible with eachother and not really stable.

v3.0.0
v3.0.1
v3.1.0
v3.1.1
**v3.1.2**
v3.2.0
v3.2.1
v3.2.2
v3.2.3
v3.3.0
v3.3.1
v3.4.0
v3.4.1
v3.4.2

albertdb:
Thanks for you effort, I really appreciate it. I would personally opt for the latest (and hopefully greatest) release. They fixed a ton of bugs, yes, but that for me is a god sign, as they actively fix them.

afbjorklund:
It will soon break API again, with the introduction of Podman v4. Should be OK with `minikube ssh sudo podman`, though...

But supplying a compatible podman-remote binary and testing minikube with rootless containers is beyond the scope, I think.

afbjorklund:
If you really like `podman`, you could also try with Fedora and Lima: https://github.com/afbjorklund/fedora-lima

But I think the rest of us will run Ubuntu and `containerd`... There is no "docker-env" or "podman-env", though.

albertdb:
> It will soon break API again, with the introduction of Podman v4. Should be OK with `minikube ssh sudo podman`, though...
> 
> But supplying a compatible podman-remote binary and testing minikube with rootless containers is beyond the scope, I think.

Is there any impediment to use official podman remote (for the same version)? Minikube's podman-env just sets up the environment variables for it to work. Simple but works great.

Do API breaking changes affect anything that is not abstracted by the CLI?

> If you really like `podman`, you could also try with Fedora and Lima: https://github.com/afbjorklund/fedora-lima
> 
> But I think the rest of us will run Ubuntu and `containerd`... There is no "docker-env" or "podman-env", though.

Lima is a macOS thing, if I'm not wrong.

afbjorklund:
> Is there any impediment to use official podman remote (for the same version)?

Nope, just that they delete the old packages so users have to use github binaries.

> Lima is a macOS thing, if I'm not wrong.

macOS and Linux, actually. It uses QEMU.

albertdb:
> > Is there any impediment to use official podman remote (for the same version)?
> 
> Nope, just that they delete the old packages so users have to use github binaries.
> 
> > Do API breaking changes affect anything that is not abstracted by the Podman CLI?
> 
> As long as the client and server versions match perfectly, you should be more or less OK.

I think that is perfectly OK for Windows users =)

afbjorklund:
ok-to-build-iso

albertdb:
So is v3.1.2 the final choice or is it a reference to then try and compare with v3.4.2?

afbjorklund:
> So is v3.1.2 the final choice or is it a reference to then try and compare with v3.4.2?

We can bump it further, I don't think the version was ever the problem with the ISO build.

The go version requirement was a bit of a blocker before, but it has 1.15 now at least.
After upgrading the OS version, that version would increase from 1.15.15 to 1.16.10

Usually the versions that RHEL and Debian choose, are better choices for long term

albertdb:
> > So is v3.1.2 the final choice or is it a reference to then try and compare with v3.4.2?
> 
> We can bump it further, I don't think the version was ever the problem with the ISO build.
> 
> ~The go version requirement was a bit of a blocker before, but it has 1.15 now at least. After upgrading the OS version, that version would increase from 1.15.15 to 1.16.10~
> 
> Usually the versions that RHEL and Debian choose, are better choices for long term
> 
> * RHEL 8.5: v3.3.1
> * Debian 11: v3.0.1
> 
> EDIT: Forgot that go was patched to 1.17 already
> 
> Can try upgrading it from v3.1.2 to v3.4.2 later on

Did you have success? v3,4,4 is out and v3.4.3 fixes a CVE.

afbjorklund:
Nope. Not sure how much the podman machine CVE affects minikube, it mostly affects the host Podman ?

afbjorklund:
Anyway, ISO build failures were unrelated to podman

albertdb:
> Anyway, ISO build failures were unrelated to podman

Nice to hear. Any chance to rebase on top of latest stable release? I don't know what is your Git workflow, but if something is polluting the tests, it would be better to separate concerns.

afbjorklund:
It builds OK here

albertdb:
> It builds OK here

Quite strange then, if the only change is the podman version and the reason for not building is unrelated to podman  :S

albertdb:
> The monolithic build is a stability issue for CI, but also the packages can have their own issues (currently KIC us failing on checksums)

So then there is no reason to choose one instead of another, isn't it?

I mean, being up-to-date today means not being too out-of-date soon.

afbjorklund:
Getting some help now, so hopefully will be back on track soon. First we want to upgrade os version and build for arm64.

But upgrading kernel and container runtimes is also on the wishlist, and long overdue due to the other upgrade issues...

afbjorklund:
Let's do another PR, when this is possible.

gallomas:
@afbjorklund I would be willing to help on this. Looking at comments and trying to understand what was the problem blocking the PR / tests. Is there a issue opened to track?

afbjorklund:
Not really, it was a simple version upgrade...

Should probably upgrade to something like 3.4.2, before the big 4.0.0 comes and deletes all the packages.

For minikube purposes, the actual 3.x probably doesn't matter at all. And 3.0.1 is still "supported", for afaik.

* https://packages.debian.org/stable/podman

  * 3.0.1+dfsg1-3+b2

* https://podman.io/blogs/2021/03/02/podman-support-for-older-distros.html

  * _Podman 3.0 will be the last major build on CentOS 7, Debian 10 and Ubuntu 18.04._

Most requests are still for dockerd, some for containerd.

albertdb:
There are legitimate use cases for more recent versions of Podman, such as https://github.com/devonfw-forge/podman-minikube

afbjorklund:
> There are legitimate use cases for more recent versions of Podman, such as https://github.com/devonfw-forge/podman-minikube (i.e. ditching Docker Desktop and even docker)

Podman has its own built-in solution with CoreOS, but sure it could be useful for `minikube start --no-kubernetes`

albertdb:
> > There are legitimate use cases for more recent versions of Podman, such as https://github.com/devonfw-forge/podman-minikube (i.e. ditching Docker Desktop and even docker)
> 
> Podman has its own built-in solution with CoreOS, but sure it could be useful for `minikube start --no-kubernetes`

The idea is to "silently" introduce Kubernetes and let developers use it just because they already have it.

afbjorklund:
The plan is still to make both podman and nerdctrl versions available, presumably using modern versions of each.

* https://github.com/kubernetes/minikube/issues/11706

albertdb:
> > There are legitimate use cases for more recent versions of Podman, such as https://github.com/devonfw-forge/podman-minikube (i.e. ditching Docker Desktop and even docker)
> 
> Podman has its own built-in solution with CoreOS, but sure it could be useful for `minikube start --no-kubernetes`

I guess you were talking about Podman machine for MacOS, I was talking about Windows, where WSL2 is not feasible.

PR #11746, opened 2021-06-22
Add GitHub Action to check if translations are valid

spowelljr:
Added GitHub Action workflow that will run when anything in the translation folder is updated to ensure all the translation files are valid.

First checks if the JSON itself is valid, then checks if there are matching variables in the translated string.

Fixed translations that failed the test.


.github/workflows/translations.yml
```diff
@@ -0,0 +1,28 @@
+name: Translations Check
```
medyagh, line 1: how about "Translations Validation"

.github/workflows/translations.yml
```diff
@@ -0,0 +1,28 @@
...
+on:
+  pull_request:
+    paths:
+      - "translations/**"
+env:
+  GOPROXY: https://proxy.golang.org
+  GO_VERSION: 1.16.4
```
medyagh, line 8: update "goland updater" to include this yaml file too

https://github.com/medyagh/minikube/blob/b82c56cbcf03910087a1a153faa327d9326ecb73/hack/update/golang_version/update_golang_version.go#L73

.github/workflows/translations.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  unit_test:
+    runs-on: ubuntu-18.04
```
medyagh, line 11: lets go newer ubuntu...this is kind of ancient...why not ?

medyagh (changes requested):

pkg/minikube/translate/translate_test.go
```diff
@@ -97,3 +102,59 @@ func TestT(t *testing.T) {
 		})
 	}
 }
+
+func TestTranslationFilesValid(t *testing.T) {
+	languages := []string{"de", "es", "fr", "ja", "ko", "pl", "zh-CN"}
```
ilya-zuyev, line 107: May be instead of hardcoding the list of languages it's better to check all translations/*.json files?
In this case we don't need to update this test if we add a new translation.

pkg/minikube/translate/translate_test.go
```diff
@@ -97,3 +102,59 @@ func TestT(t *testing.T) {
...
+				if v == "" {
+					continue
+				}
+
+				// get all variables (ex. {{.name}})
+				keyVariables := re.FindAllString(k, -1)
+				valueVariables := re.FindAllString(v, -1)
+
+				// check if number of original string and translated variables match
+				if len(keyVariables) != len(valueVariables) {
```
ilya-zuyev, line 140: `FindAllString` result can have duplicates.  Technically, it's possible to have something like "{{.node}} is broken" : "{{.node}} is broken,  please restart {{.node}}". In this case len(keyVariables)=1 and len(valueVariables)=2. We probably need to compare the number of unique matches in both key and value

pkg/minikube/translate/translate_test.go
```diff
@@ -97,3 +102,59 @@ func TestT(t *testing.T) {
...
+
+				// sort so comparing variables is easier
+				sort.Strings(keyVariables)
+				sort.Strings(valueVariables)
+
+				// for each variable in the original string
+				for i, keyVar := range keyVariables {
+					// check if translated string has same variable
+					if keyVar != valueVariables[i] {
+						t.Errorf("line %q has mismatching variables; original string variables: %s do not match translated variables: %s", k, keyVariables, valueVariables)
```
ilya-zuyev, line 153: Let's output here `v` as well. Would add a bit more context to fix the translation

pkg/minikube/translate/translate_test.go
medyagh: good idea ! lets do that

translations/pl.json
```diff
@@ -549,7 +549,7 @@
 	"Show only log entries which point to known problems": "Pokaż logi które wskazują na znane problemy",
 	"Show only the most recent journal entries, and continuously print new entries as they are appended to the journal.": "",
 	"Simulate numa node count in minikube, supported numa node count range is 1-8 (kvm2 driver only)": "",
-	"Skipped switching kubectl context for {{.profile_name}} because --keep-context was set.": "Zignorowano zmianę kontekstu kubectl ponieważ --keep-context zostało przekazane",
+	"Skipped switching kubectl context for {{.profile_name}} because --keep-context was set.": "Zignorowano zmianę kontekstu kubectl {{.profile_name}} ponieważ --keep-context zostało przekazane",
```
ilya-zuyev, line 552: zmianę kontekstu kubectl _dla_ {{.profile_name}} ponieważ

pkg/minikube/translate/translate_test.go
```diff
@@ -97,3 +102,81 @@ func TestT(t *testing.T) {
...
+	variables := re.FindAllString(line, -1)
+	distinctMap := make(map[string]bool)
+
+	// add them to a map to get distinct list of variables
+	for _, variable := range variables {
+		distinctMap[variable] = true
+	}
+	distinct := []string{}
+
+	// convert map into slice
```
ilya-zuyev, line 173: just a nitpick: in this case it's obvious what's going on in the next three lines, this comment is a bit of redundant

ilya-zuyev (approved):
lgtm

medyagh (approved):

PR #11749, opened 2021-06-23
wip: bump default and newest kubernetes versions

medyagh:


medyagh:
dont merge seems to be making minikube start slower

Issue #11452, opened 2021-05-19
docker: failed to register layer: ApplyLayer exit status 1 stdout:  stderr: unlinkat /usr/share/doc/adduser/examples/adduser.local.conf.examples: directory not empty.


Ashwin-Raj18:
**Steps to reproduce the issue:** 

1.On linux CentOs 7
2.cmd : minikube start

**Full output of `minikube logs` command:**
<details>
$ minikube start
* minikube v1.20.0 on Centos 7.2.1511
* Using the docker driver based on existing profile
* Starting control plane node minikube in cluster minikube
* Pulling base image ...
    > gcr.io/k8s-minikube/kicbase...: 358.10 MiB / 358.10 MiB  100.00% 1.07 MiB
* docker "minikube" container is missing, will recreate.
* Creating docker container (CPUs=2, Memory=2200MB) ...
! StartHost failed, but will try again: recreate: creating host: create: creating: setting up container node: preparing volume for minikube container: docker run --rm --name minikube-preload-sidecar --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --entrypoint /usr/bin/test -v minikube:/var gcr.io/k8s-minikube/kicbase:v0.0.22@sha256:7cc3a3cb6e51c628d8ede157ad9e1f797e8d22a1b3cedc12d3f1999cb52f962e -d /var/lib: exit status 125
stdout:

stderr:
Unable to find image 'gcr.io/k8s-minikube/kicbase:v0.0.22@sha256:7cc3a3cb6e51c628d8ede157ad9e1f797e8d22a1b3cedc12d3f1999cb52f962e' locally
gcr.io/k8s-minikube/kicbase@sha256:7cc3a3cb6e51c628d8ede157ad9e1f797e8d22a1b3cedc12d3f1999cb52f962e: Pulling from k8s-minikube/kicbase
a70d879fa598: Pulling fs layer
...
74754ab5b883: Waiting
...
c4394a92d1f8: Verifying Checksum
c4394a92d1f8: Download complete
10e6159c56c0: Verifying Checksum
10e6159c56c0: Download complete
...
3493142a6c0f: Pull complete
b123634b22fa: Verifying Checksum
b123634b22fa: Download complete
...
3507e244768c: Verifying Checksum
3507e244768c: Download complete

[...]

PR #11754, opened 2021-06-23
wip: regex solution message for docker space

medyagh:
fixes this https://github.com/kubernetes/minikube/issues/11452

### DIFF
--- a/pkg/minikube/reason/known_issues.go
+++ b/pkg/minikube/reason/known_issues.go
@@ -876,6 +876,15 @@ var internetIssues = []match{
 var guestIssues = []match{
 	{
 		Kind: Kind{
+			ID:       "GUEST_PROVISION_NOSPACE_DOCKER",
+			ExitCode: ExInsufficientStorage,
+			Advice:   "Ensure you have at least 20GB of free disk space accessible to Docker Desktop",
+		},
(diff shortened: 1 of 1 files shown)


pkg/minikube/reason/known_issues.go
```diff
@@ -874,6 +874,15 @@ var internetIssues = []match{
...
 
 var guestIssues = []match{
+	{
+		Kind: Kind{
+			ID:       "GUEST_PROVISION_NOSPACE_DOCKER",
+			ExitCode: ExInsufficientStorage,
+			Advice:   "Ensure you have at least 20GB of free disk space accessible to Docker Desktop",
+		},
+		// https://github.com/kubernetes/minikube/issues/11452
+		Regexp: re(`docker: failed to register layer: ApplyLayer exit status 1`),
```
ilya-zuyev, line 884: This message from docker can indicate a lot of possible issues, not only insufficient disk space
See for example https://bugzilla.redhat.com/show_bug.cgi?id=1311758

medyagh: that a good point !

how about 

```
	Kind: Kind{
			ID:       "GUEST_PROVISION_DOCKER_INTERNAL",
			ExitCode: ExInternal,
			Advice:   "Ensure your Docker installation is healthy",
```

PR #11755, opened 2021-06-24
Improve finding of local network containing the ip address 

ilya-zuyev:
we have code that looks for local network interface containing the given ip address
ATM it looks like this (pseudo-code):

```
// find network for 'ip'
for _, iface := range ifaces {
		for _, ifAddr := range  iface.Addrs() {
			ifip, lan := net.ParseCIDR(ifAddr.String())
			if lan.Contains(ip) {
				n.IfaceName = iface.Name
				...
				network = lan
				break
			}
		}
}
```
`break` ends the most nested loop, and the algorithm continues to check other interfaces, although we already have the answer

This PR changes this code to break the outer, interface loop instead

### DIFF
--- a/pkg/network/network.go
+++ b/pkg/network/network.go
@@ -97,6 +97,8 @@ func inspect(addr string) (*Parameters, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed listing network interfaces: %w", err)
 	}
+
+ifLoop:
 	for _, iface := range ifaces {
 		ifAddrs, err := iface.Addrs()
 		if err != nil {
@@ -114,7 +116,7 @@ func inspect(addr string) (*Parameters, error) {
 				n.IfaceMAC = iface.HardwareAddr.String()
 				n.Gateway = n.IfaceIPv4
 				network = lan
-				break
+				break ifLoop
 			}
 		}
 	}
--- a/pkg/network/network.go
+++ b/pkg/network/network.go
@@ -77,11 +77,52 @@ type Interface struct {
 	IfaceMAC  string
 }
 
+// lookupInInterfaces iterates over all local network interfaces
+// and tries to match "ip" with associated networks
+// returns (network parameters, ip network, nil) if found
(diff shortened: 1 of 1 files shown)


ilya-zuyev:
@prezha could you please take a look when you have time?

pkg/network/network.go
```diff
@@ -97,6 +97,8 @@ func inspect(addr string) (*Parameters, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed listing network interfaces: %w", err)
 	}
+
+ifLoop:
```
medyagh, line 101: thank you for this PR I think you point out correctly the logic seems wrong but I am curious, did you face any issue that made you investiage this ? if yes please link the issue for it.

pkg/network/network.go
```diff
@@ -97,6 +97,8 @@ func inspect(addr string) (*Parameters, error) {
...
 	for _, iface := range ifaces {
```
medyagh, line 102: for better code readability, how about we move this to a func

and instead of break, we just return

something like 
```
parseSomething(ifs []Interface ) *Parameters{}
```

pkg/network/network.go
ilya-zuyev: No, there is no issue here. This PR is just a small optimization

prezha:
> @prezha could you please take a look when you have time?

@ilya-zuyev good catch, thank you for correcting this - yes, you are right: no need to continue checking other interfaces if we've already found the one we were looking for!

ilya-zuyev:
> > @prezha could you please take a look when you have time?
> 
> @ilya-zuyev good catch, thank you for correcting this - yes, you are right: no need to continue checking other interfaces if we've already found the one we were looking for!

cool. thank you!

pkg/network/network.go
ilya-zuyev: We need to change `network` as well if we find the interface (and don't touch it if not)

medyagh (approved):
thank you

PR #11766, opened 2021-06-25
Create page with all tests for an environment in flake rate charts

### LINKED ISSUE #11758: Make flake rate page to show all tests for a single environment
We would like to see the flake rates of all tests for a single environment in the flake charts.

andriyDev:
fixes #11758.

In order to view it, just use the env query parameter, instead of both env and test.

The main chart only shows the charts of the top 10 flak-iest tests, since all the tests can make it very hard to read.

Main chart:
image

Below (contains all flake rates):
image

### DIFF
--- a/hack/jenkins/test-flake-chart/flake_chart.html
+++ b/hack/jenkins/test-flake-chart/flake_chart.html
@@ -1,6 +1,18 @@
 <html>
   <head>
     <script type="text/javascript" src="https://www.gstatic.com/charts/loader.js"></script>
+    <style>
+      table {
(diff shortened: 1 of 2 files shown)


medyagh (approved):

medyagh (commented):
in a separate PR or this PR could we also add two more charts
1- total duration of the environment taking the test over time
2- total number of tests passing/failing/skipping over time
so it will be 3 charts
and one big table

andriyDev:
@medyagh I think those should be a separate PR, I'll create those issues as soon as this PR is merged.

Issue #11220, opened 2021-04-28
Using the none driver sets the hostname to 'control-plane'


afbjorklund:
This is not desired, the hostname should stay the way it is.

For multi-node VMs that we create ourselves, then sure.

But existing machines get to keep their current hostnames.

afbjorklund:
This happens when systemd co-conspires with NetworkManager.

```
Apr 28 17:09:22 control-plane NetworkManager[873]: <info>  [1619622562.5389] policy: set-hostname: set hostname to 'control-plane.minikube.internal' (from address lookup)
Apr 28 17:09:22 control-plane systemd-hostnamed[885]: Changed host name to 'control-plane.minikube.internal'
```

Adding the address entry to the `/etc/hosts` file caused it, I think.

afbjorklund:
Here is the way to reproduce this behaviour:

1) Start a VM, let's call it "localhost"
2) Add an entry to `/etc/hosts`, for the `-i` interface
3) Reboot the VM, see the hostname change as well

afbjorklund:
Also we add another entry for localhost, as "host.minikube.internal".

```
127.0.0.1    host.minikube.internal
10.0.2.15    control-plane.minikube.internal
```

Not sure if that makes any sense either, maybe just remove both ?

sharifelgamal:
This should be as simple as skipping setting the hostname for the none or ssh drivers. Help wanted!

afbjorklund:
This is actually a systemd "feature", and not something that is caused by minikube:

https://www.freedesktop.org/wiki/Software/systemd/hostnamed/

If you have a system without a "proper" hostname, and it gets a network interface host.
Then systemd assigns it a "transient" hostname, while keeping the "static" hostname:

```console
$ hostnamectl 
   Static hostname: localhost.localdomain
Transient hostname: control-plane.minikube.internal
         Icon name: computer-vm
           Chassis: vm
        Machine ID: b164b6646e3f4d75aa151a3848a9ac20
           Boot ID: f690c03a68f04e408f057e7f86ce9e70
    Virtualization: oracle
  Operating System: CentOS Linux 8
       CPE OS Name: cpe:/o:centos:centos:8
            Kernel: Linux 4.18.0-240.1.1.el8_3.x86_64
      Architecture: x86-64
```



Ubuntu sets up a network for the host, so it is **not** affected:

```
127.0.0.1	localhost
127.0.1.1	ubuntu-focal	ubuntu-focal
```

[...]

PR #11775, opened 2021-06-27
Avoid adding an extra hosts entry for localhost

afbjorklund:
When running with the "none" driver, the HostIP returns localhost.
But we don't need multiple entries for localhost in /etc/hosts.

When running with the "ssh" driver, the HostIP returns external IP.
It also doesn't need multiple entries in the /etc/hosts file.

For #11220

Trying to contact localhost through `host.minikube.internal` will be broken...

This is similar to trying to contact localhost through `minikube ssh`, I guess ?

I don't think it made much sense adding localhost as "host" in the pods anyway.

So we just remove the `/etc/hosts` entry, for the "none" and the "ssh" drivers.


medyagh:
@afbjorklund functional tests seems being broken after this

afbjorklund:
Will have to resumed later, try some other way

Issue #11609, opened 2021-06-08
The check for memory limits is currently broken for cgroups v1


afbjorklund:
Currently it checks "" (empty string), if cgroup2 is false:

```go
// HasMemoryCgroup checks whether it is possible to set memory limit for cgroup.
func HasMemoryCgroup() bool {
        memcg := true
        if runtime.GOOS == "linux" {
                var memory string
                if cgroup2, err := IsCgroup2UnifiedMode(); err == nil && cgroup2 {
                        memory = "/sys/fs/cgroup/memory/memsw.limit_in_bytes"
                }
                if _, err := os.Stat(memory); os.IsNotExist(err) {
                        klog.Warning("Your kernel does not support memory limit capabilities or the cgroup is not mounted.")
                        memcg = false
                }
        }
        return memcg
}
```

Since 7b0bf57f4c9b87a7b0c9362bdfbb5c0add2094d7

Most likely the code is also wrong for cgroups v2 as well ?

The end result is no memory limits, for the docker driver.

Seems to be the same issue as in #10935

sharifelgamal:
Yeah this seems like a simple fix, let's do it ASAP.

sharifelgamal:
@daehyeok could you look at this?

afbjorklund:
Will let someone look at this, wasn't so simple it seems.

PR #11777, opened 2021-06-27
WIP: Fix the broken cgroups checks

afbjorklund:
See #11609

Basically a bunch of duplicated checks from moby, which in turn gets them from libcontainer. Needs to be updated for cgroups v2.

### DIFF
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -123,7 +123,8 @@ func HasMemoryCgroup() bool {
 	return memcg
 }
 
-func hasMemorySwapCgroup() bool {
+// HasMemorySwapCgroup checks whether it is possible to set memory swap limit for cgroup.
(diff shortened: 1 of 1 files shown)


pkg/drivers/kic/oci/oci.go
```diff
@@ -139,6 +140,36 @@ func hasMemorySwapCgroup() bool {
 	return memcgSwap
 }
 
+// HasCPUCFSPeriod checks whether it is possible to set cpu cfs period for cgroup.
+func HasCPUCFSPeriod() bool {
+	cpuCfsPeriod := true
+	if runtime.GOOS == "linux" {
```
ilya-zuyev, line 146: for non-Linux `HasCPUCFSPeriod()` always returns `true`. It it's by design, let's change then name. 
"CFS" make sense only in Linux context

afbjorklund:
Not really in progress

PR #11804, opened 2021-06-29
Lower parallelism when testing on Windows

spowelljr:
Windows tests are failing with `[kubelet-check] Initial timeout of 40s passed.`

This is due to too much parallelism, lowering the parallelism number on Windows to prevent the errors.

### DIFF
--- a/test/integration/main_test.go
+++ b/test/integration/main_test.go
@@ -99,7 +99,7 @@ func setMaxParallelism() {
 
 	// Windows tests were failing from timeouts due to too much parallelism
 	if runtime.GOOS == "windows" {
-		limit = int(limit / 2)
+		limit /= 2
 	}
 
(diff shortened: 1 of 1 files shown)


test/integration/main_test.go
```diff
@@ -97,6 +97,11 @@ func setMaxParallelism() {
 	// Each "minikube start" consumes up to 2 cores, though the average usage is somewhat lower
 	limit := int(math.Floor(float64(maxp) / 1.75))
 
+	// Windows tests were failing from timeouts due to too much parallelism
+	if runtime.GOOS == "windows" {
+		limit = int(limit / 2)
```
sharifelgamal, line 102: `limit /= 2` should be enough here and should fix lint

spowelljr:
Windows integration tests completed with `12 / 258 failures in 120.57 minute(s)`

test/integration/main_test.go
spowelljr: Done, thanks

sharifelgamal (approved):

PR #11816, opened 2021-06-29
Report flakes only after synchronizing all integration tests

### LINKED ISSUE #11738: Make flake rates only be reported once and cover all environments
Currently, flake rates are reported once per environment (we specify which environments). We should be reporting flake rate of all environments in one comment instead of mutliple.

### LINKED ISSUE #11770: Docker Windows flakes not updating
Flake rates for all other integration tests seem to be updating test data with every master run, but Docker Windows does not update test data

andriyDev:
fixes #11738.
fixes #11770.

## How this works
1. Minikube cross build job builds, then triggers each integration test with ROOT_JOB_ID=$BUILD_NUMBER. This is a unique ID to reference each set of integration jobs.
2. Each Jenkins integration job (before running the integration tests) creates an entry in "gs://<log_directory>/started_environments_${ROOT_JOB_ID}.txt" with the name of the job. As each job registers itself, it will ensure that only active jobs are waited on.
3. Each Jenkins integration job runs integration tests
4. Each Jenkins integration job creates an entry in "gs://<log_directory>/finished_environments_${ROOT_JOB_ID}.txt" with the name of the job. As a post-build action, each job will trigger the "Flake Rate Upload" job (jobs which use upload_integration_reports will have this done for them since upload_integration_reports triggers the job iteslf).
5. The "Flake Rate Upload" job checks if ...finished_environments... contains all the environments in ...started_environments.... If so, the job takes all gopogh summaries, and either reports flakes or uploads the tests to the flake data (just as before). Otherwise, the job stops without error - another job should occur later after another integration test completes to retry.

The ROOT_JOB_ID prevents different test runs from clobbering each other, since each test run will have a different build number for the minikube cross build job, and each integration test writes to a unique text file.

Since "Flake Rate Upload" actually performs its work once, flakes are now only reported once.

### DIFF
--- a/hack/jenkins/common.sh
+++ b/hack/jenkins/common.sh
@@ -142,6 +142,17 @@ fi
 # Add the out/ directory to the PATH, for using new drivers.
 export PATH="$(pwd)/out/":$PATH
 
+STARTED_ENVIRONMENTS="gs://minikube-builds/logs/${MINIKUBE_LOCATION}/${COMMIT:0:7}/started_environments_${ROOT_JOB_ID}.txt"
+# Ensure STARTED_ENVIRONMENTS exists (but don't clobber)
+< /dev/null gsutil cp -n - "${STARTED_ENVIRONMENTS}"
+# Copy the job name to APPEND_TMP
+APPEND_TMP="gs://minikube-builds/logs/${MINIKUBE_LOCATION}/${COMMIT:0:7}/$(basename $(mktemp))"
+echo "${JOB_NAME}"\
+  | gsutil cp - "${APPEND_TMP}"
+# Append
+gsutil compose "${STARTED_ENVIRONMENTS}" "${APPEND_TMP}" "${STARTED_ENVIRONMENTS}"
+gsutil rm "${APPEND_TMP}"
+
 echo
 echo ">> Downloading test inputs from ${MINIKUBE_LOCATION} ..."
 gsutil -qm cp \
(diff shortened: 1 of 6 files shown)


andriyDev:
Currently the "Flake Rate Upload" job is disabled, so there will be no flake rate message here. Once this PR is merged, the Flake Rate Upload job should be enabled.

/ok-to-test

hack/jenkins/common.sh
```diff
@@ -142,6 +142,17 @@ fi
 # Add the out/ directory to the PATH, for using new drivers.
 export PATH="$(pwd)/out/":$PATH
 
+STARTED_ENVIRONMENTS="gs://minikube-builds/logs/${MINIKUBE_LOCATION}/${COMMIT:0:7}/started_environments_${ROOT_JOB_ID}.txt"
+# Ensure STARTED_ENVIRONMENTS exists (but don't clobber)
```
medyagh, line 146: elaborate better, what on what u mean by don't clobber

hack/jenkins/test-flake-chart/sync_tests.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
```
medyagh, line 2: add a comment block what does this script do ?

hack/jenkins/test-flake-chart/sync_tests.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
medyagh, line 1: copy the license boilerplate here

hack/jenkins/common.sh
andriyDev: Done.

hack/jenkins/test-flake-chart/sync_tests.sh
andriyDev: Done.

hack/jenkins/test-flake-chart/sync_tests.sh
andriyDev: Done.

medyagh (approved):

PR #11819, opened 2021-06-29
windows: support renaming binary to `kubectl.exe` and  running as kubectl

### LINKED ISSUE #11838: Fix TestFunctional/serial/MinikubeKubectlCmdDirectly on Windows
This is a failing test on Windows that should be fixed.

spowelljr:
Closes #11838

We support renaming the minikube binary to `kubectl` and the binary will run as kubectl. However, we don't support renaming the binary to `kubectl.exe` on Windows and running as kubectl.

This PR adds support for this and also makes the the Windows test pass.

### DIFF
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -100,6 +100,7 @@ func Execute() {
 	}
 
 	_, callingCmd := filepath.Split(os.Args[0])
+	callingCmd = strings.TrimSuffix(callingCmd, ".exe")
 
 	if callingCmd == "kubectl" {
 		// If the user is using the minikube binary as kubectl, allow them to specify the kubectl context without also specifying minikube profile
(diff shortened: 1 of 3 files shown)


cmd/minikube/cmd/root.go
```diff
@@ -101,7 +101,8 @@ func Execute() {
 
 	_, callingCmd := filepath.Split(os.Args[0])
 
-	if callingCmd == "kubectl" {
+	isExe := callingCmd == "kubectl.exe" && runtime.GOOS == "windows"
```
medyagh, line 104: could it also be called as kubectl on windows ? (without user typing .exe) 
so maybe we should call it isExe in case of `kubectl` && runtime.GOOS == "windows ?

spowelljr: The line below is `if callingCmd == "kubectl" || isExe`
So if they're calling `kubectl` on Windows it will be caught by `callingCmd == "kubectl"`

sharifelgamal (approved):

PR #11822, opened 2021-06-29
Fixing previously renamed func from GetAllVersionsFromURL to AllVersionsFromURL

spowelljr:
I forgot to update one of the func callers after I renamed the func.


ilya-zuyev (approved):
lgtm

PR #11826, opened 2021-06-29
Add a sanity test for beta releases

ilya-zuyev:
This PR extends `make check-release` target to validate beta releases along with GA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -637,7 +637,7 @@ release-hyperkit-driver: install-hyperkit-driver checksum ## Copy hyperkit using
 
 .PHONY: check-release
 check-release: ## Execute go test
-	go test -v ./deploy/minikube/release_sanity_test.go -tags=release
+	go test -timeout 42m -v ./deploy/minikube/release_sanity_test.go
 
 buildroot-image: $(ISO_BUILD_IMAGE) # convenient alias to build the docker container
 $(ISO_BUILD_IMAGE): deploy/iso/minikube-iso/Dockerfile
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -637,7 +637,7 @@ release-hyperkit-driver: install-hyperkit-driver checksum ## Copy hyperkit using
 
 .PHONY: check-release
 check-release: ## Execute go test
-	go test -v ./deploy/minikube/release_sanity_test.go -tags=release
+	go test -timeout 42m -v ./deploy/minikube/release_sanity_test.go
```
medyagh, line 640: why is the timeout changed to 42minutes?

deploy/minikube/releases-beta.json
```diff
@@ -18,9 +18,9 @@
   {
       "name": "v1.20.0-beta.0",
       "checksums": {
-          "darwin": "",
-          "linux": "",
-          "windows": ""
+          "darwin": "686f8d7c06c93f28543f982ec56a68544ab2ad6c7f70b39ede5174d7bac29651",
```
medyagh, line 21: is this change related to this PR?

ilya-zuyev: yes, with empty hashes the test for beta versions would fail

Makefile
ilya-zuyev: The test downloads all the releases, it takes some time. Now we're going to check beta releases too and it's going to be event longer. The default timeout was 10 minutes and it wasn't enough.

PR #11851, opened 2021-07-01
Replace integration test "hooks" with standalone scripts

### LINKED ISSUE #11849: Cloud shell integration tests not reporting completion for flake rate system
/assign

andriyDev:
fixes #11849.

Previously, each integration test flow had to register itself in start and finished lists. Now, we will have the minikube Jenkins job register all the tests at once. After each test completes, they trigger the "Flake Rate Upload" job which not only tries to upload, but also reports that the current test has completed.

This consolidates the flake rate system mechanisms into fewer locations. Instead of needing to modify multiple integration tests, now we only modify the minikube_set_pending script and the sync_tests script.


andriyDev:
Properly rebased now!

medyagh (approved):
seems reasonable and a better solution !

PR #11860, opened 2021-07-01
reformat PR description for gendocs

sharifelgamal:


.github/workflows/docs.yml
```diff
@@ -33,4 +33,9 @@ jobs:
           base: master
           delete-branch: true
           title: 'Update auto-generated docs and translations'
-          body: 'Committing changes resulting from `make generate-docs`\n\nAutogenerated by the generate-docs github action workflow.\n\n${{ steps.gendocs.outputs.changes }}'
+          body: |
+            Committing changes resulting from `make generate-docs`
+            Autogenerated by the generate-docs github action workflow.
```
medyagh, line 38: how about this wording
```
Updated minikube docs using `make generate-docs`

This PR is auto-generated by -docs(link)- CI workflow
```
consider adding a link to the workflow file path
https://github.com/kubernetes/minikube/blob/master/.github/workflows/docs.yml

sharifelgamal: fixed

PR #11862, opened 2021-07-01
Fix upload_tests only using local files - not GCS files

andriyDev:
This also means that a missing file will definitely cause a known error.


andriyDev:
This depends on #11851. Woops!

andriyDev:
This depends on merging in #11851.

PR #11867, opened 2021-07-01
add ok-to-test label automatically to update k8s workflow

sharifelgamal:


.github/workflows/update-k8s-versions.yml
```diff
@@ -1,8 +1,8 @@
 name: "update-kubernetes-versions"
 on:
   schedule:
-    # every week on Thursday at 20:30 UTC
-    - cron: "45 20 * * 4"
+    # every day at around 1 am pacific/8 am UTC
+    - cron: "0 8 * * *"
```
medyagh, line 5: everyday is not good, incase we want to skip a release, it will keep bugging us every day...(so it is better it bugs us every week)

medyagh (changes requested):

.github/workflows/update-k8s-versions.yml
sharifelgamal: fixed

PR #11911, opened 2021-07-06
Add integration test in TestPause to ensure `namespaces` flag works

### LINKED ISSUE #11890: Add integration test in TestPause to ensure `namespace` flag works
There was a typo in the `pause` (#11873) and `unpause` (#11879) command where `--` was included in the `namespaces` flag name, so to call it would require `----namespaces` which resulted in `bad flag syntax: ----namespaces`.

@medyagh suggested we add an integration test for this https://github.com/kubernetes/minikube/pull/11875#issuecomment-872625683.

mahalrs:
fixes #11890

### DIFF
--- a/test/integration/pause_test.go
+++ b/test/integration/pause_test.go
@@ -277,10 +277,10 @@ func validateStatus(ctx context.Context, t *testing.T, profile string) {
 	}
 	// verify the status looks as we expect
 	if cs.StatusCode != cmd.Paused {
-		t.Fatalf("incorrect status code: %v", cs.StatusCode)
+		t.Errorf("incorrect status code: expected %v, got %v", cmd.Paused, cs.StatusCode)
 	}
-	if cs.StatusName != "Paused" {
-		t.Fatalf("incorrect status name: %v", cs.StatusName)
+	if name := "Paused"; cs.StatusName != name {
+		t.Errorf("incorrect status name: expected %v, got %v", name, cs.StatusName)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


test/integration/pause_test.go
```diff
@@ -193,3 +283,24 @@ func validateStatus(ctx context.Context, t *testing.T, profile string) {
...
+	defer PostMortemLogs(t, profile)
+
+	statusOutput := runStatusCmd(ctx, t, profile, false)
+	var cs cmd.ClusterState
+	if err := json.Unmarshal(statusOutput, &cs); err != nil {
+		t.Fatalf("unmarshalling: %v", err)
+	}
+	// verify the status looks as we expect
+	if cs.StatusCode != cmd.OK {
+		t.Fatalf("incorrect status code: %v", cs.StatusCode)
```
spowelljr, line 298: I wouldn't do `t.Fatalf` for this and the following two checks. This is because if the first one fails, we can't see what the values for the other fields are, which might be useful for debugging the test if it fails.

Also, please output what the expected value in the message as well, this helps while debugging a failure.

spowelljr (changes requested):

test/integration/pause_test.go
mahalrs: Done

spowelljr:
@mahalrs Some of the tests are failing https://storage.googleapis.com/minikube-builds/logs/11911/37a59bf/Docker_Linux_containerd.html

mahalrs:
@spowelljr If I use the command `minikube pause -n mynamespace` then what is the right way to check if the command successfully paused everything in `mynamespace`.

PR #11917, opened 2021-07-07
Create Github Action automation of update-leaderboard

andriyDev:
Now update-leaderboard is fully automated!

This is just a hack together of `.github/workflows/twitter-bot.yml` and `.github/workflows/docs.yml`


.github/workflows/leaderboard.yml
```diff
@@ -0,0 +1,40 @@
+name: "update-leaderboard"
+on:
+  push:
+    tags-ignore:
+      - 'v*-beta.*'
+  release:
+    types: [published]
+jobs:
+  generate-docs:
```
medyagh, line 9: rename this to something else

.github/workflows/leaderboard.yml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ubuntu-18.04
```
medyagh, line 10: use more recent

.github/workflows/leaderboard.yml
andriyDev: Done.

.github/workflows/leaderboard.yml
andriyDev: Done.

PR #11941, opened 2021-07-08
minor fixes in common.sh

ilya-zuyev:
This PR fixes `check_install_golang.sh` invocation and logging in `common.sh`

### DIFF
--- a/hack/jenkins/common.sh
+++ b/hack/jenkins/common.sh
@@ -91,7 +91,7 @@ else
 fi
 
 # installing golang so we could do go get for gopogh
-ARCH="$ARCH" ./installers/check_install_golang.sh "1.16.4" "/usr/local" || true
+./installers/check_install_golang.sh "1.16.4" "/usr/local" || true
 
 # install docker and kubectl if not present
 sudo ARCH="$ARCH" ./installers/check_install_docker.sh || true
--- a/hack/jenkins/common.sh
+++ b/hack/jenkins/common.sh
@@ -91,7 +91,7 @@ else
 fi
 
 # installing golang so we could do go get for gopogh
-sudo ARCH="$ARCH"./installers/check_install_golang.sh "1.16.4" "/usr/local" || true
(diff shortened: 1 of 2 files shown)


hack/jenkins/common.sh
```diff
@@ -91,7 +91,7 @@ else
 fi
 
 # installing golang so we could do go get for gopogh
-sudo ARCH="$ARCH"./installers/check_install_golang.sh "1.16.4" "/usr/local" || true
+ARCH="$ARCH" ./installers/check_install_golang.sh "1.16.4" "/usr/local" || true
```
andriyDev, line 94: ARCH="$ARCH" is no longer needed I'm pretty sure since, it's not under a sudo anymore.

hack/jenkins/installers/check_install_golang.sh
```diff
@@ -67,20 +67,28 @@ function check_and_install_golang() {
...
+  local -r GO_DIR="$2/go"
+  echo "Installing golang version: $GO_VER in $GO_DIR"
 
   INSTALLOS=linux
   if [[ "$OSTYPE" == "darwin"* ]]; then
     INSTALLOS=darwin
   fi
+
+  local -r GO_TGZ="go${GO_VER}.${INSTALLOS}-${ARCH}.tar.gz"
+  pushd /tmp
```
andriyDev, line 80: Maybe we should preserve the >/dev/null? Since otherwise it will print out the current stack which seems unnecessary.

andriyDev (changes requested):

hack/jenkins/common.sh
ilya-zuyev: yep, good catch. fixed

hack/jenkins/installers/check_install_golang.sh
ilya-zuyev: Let's keep it. It's just one line in logs, but it gives an extra bit of information about where we are.

sharifelgamal (approved):

PR #11943, opened 2021-07-08
time-to-k8s: Leave committing to GitHub Action

spowelljr:
**Problem:**
I was committing changes in the `time-to-k8s.sh` script, the GitHub Action now handles this and states `If there are no changes (i.e. no diff exists with the checked-out base branch), no pull request will be created and the action exits silently.` This was causing no PR to be created.

**Solution:**
Remove committing from the `time-to-k8s.sh` script


sharifelgamal:
did we make an associated change to the time-to-k8s yaml file?

spowelljr:
> did we make an associated change to the time-to-k8s yaml file?

Yup https://github.com/kubernetes/minikube/pull/11929

sharifelgamal (approved):

PR #11948, opened 2021-07-09
Retain full container name when passing --image-repository

### LINKED ISSUE #11934: --image-repository strips off container prefix
I'm trying to get minikube to work in an offline environment, or at least without direct access to the internet. I have Artifactory available and configured with a virtual Docker repository that ties together multiple upstream repositories (docker.io, k8s.gcr.io, gcr.io, etc.) together and presents them as a single registry.

The problem I'm having is that when I use the `--image-repository` option with minikube, it is removing any prefix from the image names, so for example `kubernetesui/dashboard` or `gcr.io/k8s-minikube/storage-provisioner` are being pulled as `docker.artifactory.example.com/dashboard` or `docker.artifactory,example.com/storage-provisioner` which doesn't work, however if you manually pull `docker.artifactory.example.com/k8s-minikube/storage-provisioner` it will succeed.

1. `minikube start --download-only --image-repository=docker.artifactory.example.com`
2. All images will be pulled apart from `kubernetesui/dashboard`, `kubernetesui/metrics-scraper` & `gcr.io/k8s-minikube/storage-provisioner` as they're the only ones with a prefix, all of the `k8s.gcr.io/*` images are fetched successfully.
3. It also affects the pulling of the kicbase image as the original name is `gcr.io/k8s-minikube/kicbase` so I suspect the `/k8s-minikube/` is being stripped there as well.

**Full output of `minikube start` command:**
<details>
minikube start --download-only --image-repository=docker.artifactory.example.com --kubernetes-version=v1.19.0
😄  minikube v1.22.0 on Redhat 7.9
🆕  Kubernetes 1.21.2 is now available. If you would like to upgrade, specify: --kubernetes-version=v1.21.2
❗  docker is currently using the devicemapper storage driver, consider switching to overlay2 for better performance

🚜  Pulling base image ...
❗  The image 'docker.artifactory.example.com/metrics-scraper:v1.0.4' was not found; unable to add it to cache.
❗  The image 'docker.artifactory.example.com/storage-provisioner:v5' was not found; unable to add it to cache.
❗  The image 'docker.artifactory.example.com/dashboard:v2.1.0' was not found; unable to add it to cache.
E0708 11:28:24.112160  101078 cache.go:200] Error downloading kic artifacts:  getting remote image: GET https://docker.artifactory.example.com/v2/kicbase/manifests/sha256:6f936e3443b95cd918d77623bf7b595653bb382766e280290a02b4a349e88b79: MANIFEST_UNKNOWN: The named manifest is not known to the registry.; map[manifest:kicbase]
✅  Download complete!
</details>

So either the `--image-repository` option needs to not strip any prefix from the image names or an additional option to tune whether to do that or not if it's going to break behaviour with other registries. So to summarise, I'm looking to be able to map the image names as follows:

kubernetesui/dashboard -> docker.artifactory.example.com/kubernetesui/dashboard
k8s.gcr.io/kube-apiserver -> docker.artifactory.example.com/kube-apiserver
gcr.io/k8s-minikube/storage-provisioner -> docker.artifactory.example.com/k8s-minikube/storage-provisioner
etc.

bodgit:
This PR updates how the container image names are changed when the `--image-repository` flag is passed.

Previously, a container image like `gcr.io/k8s-minikube/storage-provisioner` would be changed to `registry.example.com/storage-provisioner`. With this PR, it is now changed to `registry.example.com/k8s-minikube/storage-provisioner`, i.e. the registry hostname is the only part that is replaced.

Fixes #11934

### DIFF
--- a/pkg/minikube/bootstrapper/images/images.go
+++ b/pkg/minikube/bootstrapper/images/images.go
@@ -138,19 +138,19 @@ func storageProvisioner(mirror string) string {
 // dashboardFrontend returns the image used for the dashboard frontend
 func dashboardFrontend(repo string) string {
 	if repo == "" {
-		repo = "docker.io/kubernetesui"
+		repo = "docker.io"
 	}
 	// See 'kubernetes-dashboard' in deploy/addons/dashboard/dashboard-dp.yaml
-	return path.Join(repo, "dashboard:v2.1.0")
+	return path.Join(repo, "kubernetesui", "dashboard:v2.1.0")
 }
 
 // dashboardMetrics returns the image used for the dashboard metrics scraper
 func dashboardMetrics(repo string) string {
 	if repo == "" {
-		repo = "docker.io/kubernetesui"
+		repo = "docker.io"
 	}
 	// See 'dashboard-metrics-scraper' in deploy/addons/dashboard/dashboard-dp.yaml
-	return path.Join(repo, "metrics-scraper:v1.0.4")
+	return path.Join(repo, "kubernetesui", "metrics-scraper:v1.0.4")
 }
 
 // KindNet returns the image used for kindnet
--- a/pkg/minikube/bootstrapper/images/images_test.go
+++ b/pkg/minikube/bootstrapper/images/images_test.go
@@ -96,9 +96,9 @@ func TestAuxiliary(t *testing.T) {
 
 func TestAuxiliaryMirror(t *testing.T) {
 	want := []string{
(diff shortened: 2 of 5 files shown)


medyagh (commented):
Thank you very much for this pr. I believe we have unit tests for this. How about adding unit tests for this case ?

(When mirror is set ...)

bodgit:
There were some existing tests that were checking the container image names when a registry mirror was set. I've fixed those up now so all tests are passing again.

sharifelgamal (approved):

PR #11951, opened 2021-07-10
time-to-k8s Public Charts

### LINKED ISSUE #11944: Create public charts for time-to-k8s
1. Create a GitHub Action workflow that runs time-to-k8s benchmark daily against HEAD
2. Store the results of the job as JSON and upload it to bucket
3. Use the data in the bucket to generate a chart and upload it to the website

spowelljr:
Closes #11944

This PR creates a GitHub Action that runs everyday. It runs the time-to-k8s benchmark against HEAD of minikube for both the Docker and containerd runtimes, then appends the result to the existing chart. The chart below is an example of what it looks like.

image

### DIFF
--- a/hack/benchmark/time-to-k8s/public-chart/generate-chart.go
+++ b/hack/benchmark/time-to-k8s/public-chart/generate-chart.go
@@ -1,0 +1,219 @@
+/*
+Copyright 2021 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


hack/benchmark/time-to-k8s/public-chart/generate-chart.go
```diff
@@ -0,0 +1,211 @@
...
+	flag.Parse()
+
+	qq := readInCSV(*csvPath)
+	d := readData(*pastRunsPath)
+	d.Data = append(d.Data, qq)
+	updateJSON(d, *pastRunsPath)
+	createChart(d, *chartPath)
+}
+
+func readInCSV(csvPath string) data {
```
medyagh, line 48: Change name for both func and the data structures and add comment

I would like to be able to read the code without looking at each data structures and implemenation of func
5to see what it doing.

Currently i have to read the func to see what it does
And then look at every data structure to see what it actually holds

medyagh (changes requested):
add containerd too

spowelljr:
> add containerd too

Added

PR #11968, opened 2021-07-13
Added restart command after setting crio options

### LINKED ISSUE #11729: CRI-O systemd unit not picking up /etc/sysconfig/crio.minikube and thus not setting --insecure-registry

Started thus:
```
minikube start \
	--vm-driver=kvm2 \
	--container-runtime=cri-o \
	--cni=bridge \
	--addons=registry \
	--cpus=4 \
	--service-cluster-ip-range=10.96.0.0/12
```
Now, do `minikube ssh` and:
```
sudo systemctl status crio
```
You will see that the command used is `/usr/bin/crio --log-level=debug`. However, this is incomplete. Looking at `/etc/systemd/system/multi-user.target.wants/crio.service` you will see that it refers to `/etc/sysconfig/crio.minikube`, which importantly has extra arguments for `--insecure-registry`. However, it seems that `/etc/sysconfig/crio.minikube` was not used.

This breaks the registry add-on out of the box.

As a workaround, do a quick:
```
sudo systemctl restart crio
sudo systemctl status crio
```
And you will see that `--insecure-registry` has now been set properly.

My guess is that somehow the "crio.minikube" file was create too late during the setup process.

kadern0:
Fixes #11729 

Before:

Screenshot from 2021-07-13 20-37-28

After:

Screenshot from 2021-07-13 20-37-19

### DIFF
--- a/pkg/provision/provision.go
+++ b/pkg/provision/provision.go
@@ -254,7 +254,7 @@ CRIO_MINIKUBE_OPTIONS='{{ range .EngineOptions.InsecureRegistry }}--insecure-reg
 		return err
 	}
 
-	if _, err = p.SSHCommand(fmt.Sprintf("sudo mkdir -p %s && printf %%s \"%s\" | sudo tee %s", path.Dir(crioOptsPath), crioOptsBuf.String(), crioOptsPath)); err != nil {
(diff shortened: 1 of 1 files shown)


sharifelgamal:
@afbjorklund what do you think?

afbjorklund:
I thought we already did a crio restart below, maybe it was removed since ?

PR #11979, opened 2021-07-14
Add configurable port for minikube mount

### LINKED ISSUE #11978: Add configurable port for minikube mount
The minikube mount command currently just uses a random free port to run the server on. If the user is using a firewall (definitely not me), this can block any requests from coming in, so mounting fails. Using a configurable port users can ensure the port is allowed through their firewall (again, not mine of course).

andriyDev:
fixes #11978.

Now users can specify what port to use for mounting (in addition to allowing random port assignment). Also, adds advice for timing out mounting to check firewall.

### DIFF
--- a/cmd/minikube/cmd/mount.go
+++ b/cmd/minikube/cmd/mount.go
@@ -190,7 +190,13 @@ var mountCmd = &cobra.Command{
 			}
 		}()
 
-		cluster.Mount(co.CP.Runner, ip.String(), vmPath, cfg)
+		err = cluster.Mount(co.CP.Runner, ip.String(), vmPath, cfg)
+		if err != nil {
+			if rtErr, ok := err.(*cluster.MountError); ok && rtErr.ErrorType == cluster.MountErrorConnect {
+				exit.Error(reason.GuestMountCouldNotConnect, "mount could not connect", rtErr)
+			}
+			exit.Error(reason.GuestMount, "mount failed", err)
+		}
 		out.Step(style.Success, "Successfully mounted {{.sourcePath}} to {{.destinationPath}}", out.V{"sourcePath": hostPath, "destinationPath": vmPath})
 		out.Ln("")
 		out.Styled(style.Notice, "NOTE: This process must stay alive for the mount to be accessible ...")
--- a/cmd/minikube/cmd/mount.go
+++ b/cmd/minikube/cmd/mount.go
@@ -50,6 +50,7 @@ const (
 // placeholders for flag values
 var (
 	mountIP      string
+	mountPort    uint16
 	mountVersion string
 	mountType    string
 	isKill       bool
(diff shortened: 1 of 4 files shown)


andriyDev:
This PR is great to read each commit separately I'd say, hint hint

pkg/minikube/cluster/mount.go
```diff
@@ -55,22 +57,25 @@ type mountRunner interface {
 }
 
 // Mount runs the mount command from the 9p client on the VM to the 9p server on the host
-func Mount(r mountRunner, source string, target string, c *MountConfig) error {
+func Mount(r mountRunner, source string, target string, c *MountConfig) {
 	if err := Unmount(r, target); err != nil {
-		return errors.Wrap(err, "umount")
+		exit.Error(reason.GuestMount, "mount failed", errors.Wrap(err, "umount"))
```
medyagh, line 62: lets not exit in here and return an error type like this
https://github.com/medyagh/minikube/blob/0e665930e86e118976b5c6b3912aaed51ccfe20f/pkg/drivers/kic/oci/errors.go?plain=1#L37

medyagh (changes requested):

pkg/minikube/cluster/mount.go
andriyDev: Done.

cmd/minikube/cmd/mount.go
```diff
@@ -191,6 +193,7 @@ var mountCmd = &cobra.Command{
 
 		err = cluster.Mount(co.CP.Runner, ip.String(), vmPath, cfg)
 		if err != nil {
+			node.ExitIfFatal(err)
```
medyagh, line 196: I dont think we should call the node package for this. node is meant for starting the node.

how about just checking the error type here and if that error type is do the right thing ?

medyagh (changes requested):

pkg/minikube/cluster/mount.go
```diff
@@ -54,19 +54,38 @@ type mountRunner interface {
...
 }
 
+const (
+	// MountErrorUnknown failed with unknown error
+	MountErrorUnknown = iota
+	// MountErrorConnect
+	MountErrorConnect
+)
+
+type MountError struct {
```
medyagh, line 64: add comment for all exported types. (or dont make it exported if it hasn't been called outside this package

medyagh (changes requested):

cmd/minikube/cmd/mount.go
andriyDev: Done.

pkg/minikube/cluster/mount.go
andriyDev: Done. It needs to be exported, else we can't check the ErrorType.

medyagh (approved):

PR #11988, opened 2021-07-15
Set specific GO_VERSION env for leaderboard workflow

andriyDev:
https://github.com/kubernetes/minikube/runs/3080009605?check_suite_focus=true This broke.


andriyDev:
Once this PR is merged, the merger should run the update-leaderboard workflow.

spowelljr (approved):

Issue #9947, opened 2020-12-13
Start the buildkit daemon on demand (socket-activated)


afbjorklund:
Buildkit uses a server (daemon), so trying to use the buildkit client fails if it is not started:

`error: failed to get status: rpc error: code = Unavailable desc = connection error: desc = "transport: error while dialing: dial unix /run/buildkit/buildkitd.sock: connect: no such file or directory"`

See #9640 (and https://github.com/moby/buildkit#quick-start)

Currently the `buildkitd` has to be started manually, as there is no systemd unit for it.

```console
$ sudo -b buildkitd
$ INFO[0000] auto snapshotter: using overlayfs            
WARN[0000] using host network as the default            
INFO[0000] found worker "p6bbje3hwqqbczwx909fjnaed", labels=map[org.mobyproject.buildkit.worker.executor:oci org.mobyproject.buildkit.worker.hostname:minikube org.mobyproject.buildkit.worker.snapshotter:overlayfs], platforms=[linux/amd64 linux/386] 
WARN[0000] using host network as the default            
INFO[0000] found worker "kgulhjepnr8a15fqgv0q4uag7", labels=map[org.mobyproject.buildkit.worker.executor:containerd org.mobyproject.buildkit.worker.hostname:minikube org.mobyproject.buildkit.worker.snapshotter:overlayfs], platforms=[linux/amd64 linux/386] 
INFO[0000] found 2 workers, default="p6bbje3hwqqbczwx909fjnaed" 
WARN[0000] currently, only the default worker can be used. 
INFO[0000] running server on /run/buildkit/buildkitd.sock
```

We should add a "buildkit.service" file for it, so that it can started/stopped/logged with systemd.

Similar to the current "containerd.service"

----

It would be nice if it was started automatically, using a systemd socket-activated service.

`sudo systemctl enable buildkit.socket`
`sudo systemctl start buildkit.socket`

This also allows setting group permissions, so that the socket can be accessed over a tunnel.

```
[Socket]
SocketMode=0660
SocketUser=root
SocketGroup=buildkit
```

Locally one can use `sudo`, but that doesn't work when using ssh tunnels to the unix socket.

`unix:///run/buildkit/buildkitd.sock`

`error: failed to get status: rpc error: code = Unavailable desc = connection error: desc = "transport: error while dialing: dial unix /run/buildkit/buildkitd.sock: connect: permission denied"`

Similar to the other runtimes:

* docker.sock #9229 #9232 
[...]

PR #12032, opened 2021-07-21
Upgrade buildkit from 0.8.2 to 0.9.0

spowelljr:
Related #9947

Upgrading buildkit to the newest version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ KUBERNETES_VERSION ?= $(shell egrep "DefaultKubernetesVersion =" pkg/minikube/co
 KIC_VERSION ?= $(shell egrep "Version =" pkg/drivers/kic/types.go | cut -d \" -f2)
 
 # Default to .0 for higher cache hit rates, as build increments typically don't require new ISO versions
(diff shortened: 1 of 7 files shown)


afbjorklund:
Are you will to try the systemd unit ?

examples/systemd/
├── buildkit.service
└── buildkit.socket

Might need to change the path, though.

spowelljr:
> Are you willing to try the systemd unit ?
examples/systemd/
├── buildkit.service
└── buildkit.socket
Might need to change the path, though.

@afbjorklund Since we both tried this and it seems to work, would you be interested in reviewing this?

afbjorklund:
We can do the systemd unit in the other PR, and just do the version upgrade here ?

spowelljr:
@afbjorklund Yeah, that was my plan, or we could do them together, doesn't really matter to me.

medyagh:
> /ok-to-build-image

that one is for KIC . :)

spowelljr:
ok-to-build-image

spowelljr:
ok-to-build-image

spowelljr:
ok-to-build-iso

medyagh (approved):

PR #12060, opened 2021-07-26
Replace duration chart with total duration of tests -line

### LINKED ISSUE #12015: Remove duration chart and add total duration to total test count chart
Context: https://storage.googleapis.com/minikube-flake-rate/flake_chart.html?env=Docker_Linux

andriyDev:
fixes #12015.

Since no gopogh reports have been uploaded containing the root job id, all job ids are set to 0, so we currently have no averaged data and only a single instance per day. As new master integration tests complete, this chart should become more accurate.

image

### DIFF
--- a/hack/jenkins/common.ps1
+++ b/hack/jenkins/common.ps1
@@ -50,7 +50,7 @@ $elapsed=$ended-$started
 $elapsed=$elapsed/60
 $elapsed=[math]::Round($elapsed, 2)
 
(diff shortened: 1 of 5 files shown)


andriyDev:
After merging, we need to run hack/jenkins/test-flake-chart/collect_data_manual.sh

medyagh:
> After merging, we need to run hack/jenkins/test-flake-chart/collect_data_manual.sh

What will this do for old data ?

medyagh:
Will it generate chart for old tests ?

medyagh (approved):

Issue #11658, opened 2021-06-15
minikube start fails with podman or kvm2 drivers on s390x


vmorris:
minikube claims to run on s390x, but I have not had success yet. I am on Fedora 33 here, so perhaps another disto would would better, but I've seen Ubuntu fail similarly using the kvm2 driver.

**Steps to reproduce the issue:** 

1. minikube config set driver podman && minikube delete
2. minikube start

*or*

1. minikube config set driver kvm2 && minikube delete
2. minikube start

**Full output of `minikube logs` command (for `podman`)**
<details>
==> Last Start <==
Log file created at: 2021/06/15 14:42:08
Running on machine: minikube1
Binary: Built with gc go1.16.4 for linux/s390x
Log line format: [IWEF]mmdd hh:mm:ss.uuuuuu threadid file:line] msg
I0615 14:42:08.728959   30552 out.go:291] Setting OutFile to fd 1 ...
I0615 14:42:08.729059   30552 out.go:343] isatty.IsTerminal(1) = true
I0615 14:42:08.729067   30552 out.go:304] Setting ErrFile to fd 2...
I0615 14:42:08.729071   30552 out.go:343] isatty.IsTerminal(2) = true
I0615 14:42:08.729176   30552 root.go:316] Updating PATH: /home/fedora/.minikube/bin
I0615 14:42:08.729372   30552 out.go:298] Setting JSON to false
I0615 14:42:08.730027   30552 start.go:111] hostinfo: {"hostname":"minikube1.zdalisv.dfw.ibm.com","uptime":3644,"bootTime":1623764485,"procs":351,"os":"linux","platform":"fedora","platformFamily":"fedora","platformVersion":"33","kernelVersion":"5.12.10-200.fc33.s390x","kernelArch":"s390x","virtualizationSystem":"kvm","virtualizationRole":"host","hostId":"457c81ed-72d0-436d-a743-6a916ac685bb"}
I0615 14:42:08.730074   30552 start.go:121] virtualization: kvm host
I0615 14:42:08.732896   30552 out.go:170] 😄  minikube v1.21.0 on Fedora 33 (s390x)
I0615 14:42:08.733072   30552 notify.go:169] Checking for updates...
I0615 14:42:08.733074   30552 driver.go:335] Setting default libvirt URI to qemu:///system
I0615 14:42:08.893533   30552 podman.go:121] podman version: 3.1.2
[...]

Issue #9762, opened 2020-11-21
minikube on raspberry pi (ARM desktop)


afbjorklund:
This is a follow-up to to the previous issues about supporting arm architectures in general, and raspberry pi in particular.

Since before, we have support for running the "none" driver but soon there is also support for running the "docker" driver.

Normally the Raspberry Pi is used to _build_ clusters. And then you connect to this from a separate machine, like a laptop.

But with the Raspberry Pi 4 (and 400), it is now possible to run both the desktop **and** the cluster on the same machine...

* https://www.raspberrypi.org/products/raspberry-pi-4-desktop-kit/

* https://www.raspberrypi.org/products/raspberry-pi-400/ ($100 US)

Distros
* Raspberry Pi OS, `arm` 32-bit OS
  <https://www.raspberrypi.org/software/operating-systems/>
* Ubuntu, `arm64`64-bit OS
  <https://ubuntu.com/download/raspberry-pi/>

Drivers
* None (~#6843~, ~#5667~, #6280)
* Docker (soon, see #9227)

**4 GB of memory is recommended**. If your Raspberry Pi 4 has 2 GB, you need to add 2 GB of swap (`dphys-swapfile`)

It is not recommended to run minikube directly on the Raspberry Pi version 2-3, due to them having 1 GB of memory.
(you can still use these old machines remotely, for instance by using the "generic" driver, but that is a separate story...)

If running a virtual desktop in the cloud, the recommendation is also to use a separate VM instead of nested virtualization.

i.e. get two of them

----

Remaining issues:
* kicbase image is not available
* etcd still adds the arch tag
* storage-provisioner v3 missing
* warnings due to resources

We also need some better documentation, for non-amd64 architectures #6159
It should show how to pick "minikube-linux-arm" or "minikube-linux-arm64"

```bash
 curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-linux-arm
 sudo install minikube-linux-arm /usr/local/bin/minikube
```

```bash
 curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-linux-arm64
 sudo install minikube-linux-arm64 /usr/local/bin/minikube

```
In order to run containers, support for memory cgroups needs to be added to kernel cmdline and then it needs a reboot.

When running with Raspbian or Xfce, you also need to install "gnome-terminal" and "fonts-noto-color-emoji" packages
[...]

PR #12079, opened 2021-07-30
Build kicbase image for all architectures

afbjorklund:
Synchronize ALL_ARCH and KICBASE_ARCH

Build release image for all 5 kubernetes arch:

`ALL_ARCH = amd64 arm arm64 ppc64le s390x`

#9762

#11658

Otherwise the minikube binary won't run (with docker)


sharifelgamal:
ok-to-build-image

medyagh:
ok-to-build-image

Issue #10068, opened 2020-12-30
No vendor packages available for buildkit installation


afbjorklund:
Unfortunately there are no packages available for "buildkit" yet, only for containerd.

This makes a lot more cumbersome to install and maintain it, using only the tarballs.

Compare:

```sh
# install docker
RUN sh -c "echo 'deb https://download.docker.com/linux/ubuntu focal stable' > /etc/apt/sources.list.d/docker.list" && \
    curl -L https://download.docker.com/linux/ubuntu/gpg -o docker.key && \
    apt-key add - < docker.key && \
    clean-install docker-ce docker-ce-cli containerd.io
```

With:

```sh
ARG BUILDKIT_VERSION="v0.8.1"

# install buildkit
RUN export ARCH=$(dpkg --print-architecture | sed 's/ppc64el/ppc64le/' | sed 's/armhf/arm-v7/') \
 && echo "Installing buildkit ..." \
    && export BUILDKIT_BASE_URL="https://github.com/moby/buildkit/releases/download/${BUILDKIT_VERSION}" \
    && curl -sSL --retry 5 --output /tmp/buildkit.tgz "${BUILDKIT_BASE_URL}/buildkit-${BUILDKIT_VERSION}.linux-${ARCH}.tar.gz" \
    && tar -C /usr/local -xzvf /tmp/buildkit.tgz \
    && rm -rf /tmp/buildkit.tgz \
    && chmod 755 /usr/local/bin/buildctl \
    && chmod 755 /usr/local/bin/buildkit-runc \
    && chmod 755 /usr/local/bin/buildkitd
```

See https://github.com/moby/buildkit/issues/1794

----

It would be even better if upstream provided _individual_ deb packages, but beggars can't be choosers...

docker-ce
* /usr/bin/dockerd

docker-ce-cli
* /usr/bin/docker

containerd.io
* /usr/bin/containerd
* /usr/bin/ctr
* /usr/bin/runc

buildkit
* /usr/bin/buildkitd
* /usr/bin/buildctl
* /usr/bin/buildkit-runc

Another _general_ problem with the vendor packages (unlike the system packages) is the lack of source code.

[docker.io](https://packages.ubuntu.com/focal/docker.io) 
[containerd](https://packages.ubuntu.com/focal/containerd) 
[runc](https://packages.ubuntu.com/focal/runc)
[buildkit](https://bintray.com/afbjorklund/buildkit/)

But this goes for the other vendor packages, too.

```console
[...]

PR #12081, opened 2021-07-30
buildkit socket: Add systemd unit

### LINKED ISSUE #9947: Start the buildkit daemon on demand (socket-activated)

### LINKED ISSUE #12264: Add configuration for buildkit, in order to not have to use flags
We can pass the setup as configuration, rather than flags.

That makes it easier to start from systemd units, and so on.

We want to use the containerd worker, by default.

`-oci-worker false --containerd-worker true `

And store the result in the "k8s.io" namespace.

`--containerd-worker-namespace k8s.io`

Helps with #9947

spowelljr:
Closes #9947
Closes #12264

Creating a systemd unit that we can use to start the buildkit daemon.


deploy/kicbase/Dockerfile
```diff
@@ -191,7 +191,7 @@ EXPOSE 22
 # create docker user for minikube ssh. to match VM using "docker" as username
 RUN adduser --ingroup docker --disabled-password --gecos '' docker 
 RUN adduser docker sudo
-RUN adduser docker podman
+RUN adduser docker podman buildkit
```
afbjorklund, line 194: I think needs to be given twice ?

```
adduser docker podman
adduser docker buildkit
```

Also make sure to create the group.

```
addgroup --system buildkit
```

At the `echo "Installing buildkit ...` block (line 135)

deploy/iso/minikube-iso/package/buildkit-bin/buildkit.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=BuildKit
+Requires=buildkit.socket
+After=buildkit.socket
+Documentation=https://github.com/moby/buildkit
+
+[Service]
+ExecStart=/usr/local/bin/buildkitd --addr fd:// --oci-worker=false --containerd-worker=true --containerd-worker-namespace=k8s.io
```
afbjorklund, line 8: This should probably be `/usr/sbin/buildkitd`, same as in the install command.

afbjorklund (changes requested):

afbjorklund (approved):
Looking good! Need to try it now, see if we missed something else

deploy/kicbase/Dockerfile
```diff
@@ -124,6 +124,7 @@ RUN sh -c "echo 'deb https://download.docker.com/linux/ubuntu focal stable' > /e
 # install buildkit
 RUN export ARCH=$(dpkg --print-architecture | sed 's/ppc64el/ppc64le/' | sed 's/armhf/arm-v7/') \
  && echo "Installing buildkit ..." \
+    && addgroup --system buildkit \
     && export BUILDKIT_BASE_URL="https://github.com/moby/buildkit/releases/download/${BUILDKIT_VERSION}" \
     && curl -sSL --retry 5 --output /tmp/buildkit.tgz "${BUILDKIT_BASE_URL}/buildkit-${BUILDKIT_VERSION}.linux-${ARCH}.tar.gz" \
```
afbjorklund, line 129: Almost forgot, you also have to download and install the systemd unit files: (to /usr/local/lib/systemd/system/)
https://raw.githubusercontent.com/moby/buildkit/v0.9.0/examples/systemd/buildkit.service
https://raw.githubusercontent.com/moby/buildkit/v0.9.0/examples/systemd/buildkit.socket
Too bad there is no decent packaging (like a .deb) for buildkit... (See #10068)

afbjorklund (changes requested):

deploy/iso/minikube-iso/package/buildkit-bin/buildkit.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=BuildKit
+Requires=buildkit.socket
+After=buildkit.socket
+Documentation=https://github.com/moby/buildkit
+
+[Service]
+ExecStart=/usr/sbin/buildkitd --addr fd:// --oci-worker=false --containerd-worker=true --containerd-worker-namespace=k8s.io
```
afbjorklund, line 8: Eventually these flags can go in /etc/buildkit instead

deploy/kicbase/Dockerfile
afbjorklund: We want to to the minikube changes as overrides, so that we can leave the upstream files intact (see podman).

https://wiki.archlinux.org/title/Systemd#Drop-in_files

afbjorklund: Eventually the /etc/buildkit will have to be populated here as well

deploy/iso/minikube-iso/package/buildkit-bin/buildkit.service
afbjorklund: `ExecStart=/usr/sbin/buildkitd --addr fd://`

afbjorklund: We _could_ use /usr/bin (like docker), even if it is "wrong"

deploy/kicbase/Dockerfile
spowelljr: Thanks for the guidance @afbjorklund! I implemented downloading and installing the systemd unit files and added the `/etc/buildkit` for kicbase as well, I tested the changes and successfully built the image using `image build`. I created the `/etc/buildkit` file by echoing to a file, I'm not sure if there was somewhere specific you want to put the file and copy it over so please take a look at my changes and let me know.

But the gist is `image build` via kicbase was successful, onto iso.

spowelljr:
Tested and working, waiting for https://github.com/kubernetes/minikube/pull/12425 to be merged before building ISO and kicbase to prevent merge conflicts

spowelljr:
ok-to-build-image

spowelljr:
ok-to-build-iso

spowelljr:
ok-to-build-image

spowelljr:
ok-to-build-image

spowelljr:
ok-to-build-image

spowelljr:
ok-to-build-image

spowelljr:
Blocked by https://github.com/cri-o/cri-o/issues/5343

spowelljr:
ok-to-build-image

sharifelgamal (approved):

spowelljr:
ok-to-build-iso

PR #12105, opened 2021-08-03
Add message at bottom of page saying when the data was collected

### LINKED ISSUE #12056: Add timestamp to flake rate charts for when the data was updated
This is just a check to do with caching and stuff.

andriyDev:
fixes #12056.

Bottom right:
image


andriyDev:
Once this is merged I'll just force start the Flake Rate Computation job to update the html+js.

PR #12106, opened 2021-08-03
configure gcp-auth addon pull secret to work with all GCR and AR mirrors

### LINKED ISSUE #12003: gcp-auth registries is too restrictive
The gcp-auth addon configures `kubernetes.io/dockercfg` secrets but only for `gcr.io` and `us-docker.pkg.dev`.
https://github.com/kubernetes/minikube/blob/791e3b59c193e0f2b3f1b675f52e3f71c6b1ab57/pkg/addons/addons_gcpauth.go#L134-L136

We should either make this configurable or add support for a broader set of endpoints.

One approach would be to query GCR/AR for the list of registries visible to the account.

cc: @sharifelgamal

sharifelgamal:
Fixes #12003


medyagh:
KVM tests 134 seems like a bad test run

sharifelgamal:
> KVM tests 134 seems like a bad test run

Yeah, one of our machines was misconfigured. It currently out of the rotation so this shouldn't happen again.

sharifelgamal:
It looks like I somehow broke the helm-tiller addon test, but I can't replicate it and can't figure out why the gcp-auth addon would somehow prevent minikube from pulling the helm-tiller image. So I'm going to rerun the tests to see if it's an actual issue.

medyagh (approved):

PR #12109, opened 2021-08-03
Simplify bodyByLinesIterator to only use splitting instead of regex

andriyDev:
Regex can be a little slower and also it isn't helpful in this context. It only serves to make the code less readable. Splitting is much more clear!


andriyDev:
NOPE IM MAKING IT FASTER

PR #12129, opened 2021-08-05
ISO: Upgrade crictl (from cri-tools) to v1.21.0

afbjorklund:


medyagh:
ok-to-build-iso

afbjorklund:
ok-to-build-iso

PR #12132, opened 2021-08-05
Move legacy version selection to its own function so tests are synced

andriyDev:
TestStoppedBinaryUpgrade seemed to be out of sync with TestRunningBinaryUpgrade. On my machine, this broke the TestStoppedBinaryUpgrade test (my machine doesn't have virtualbox, so the v1.0.0 minikube didn't have support for any of my drivers). These two tests should be testing with the same version anyway.

This may also fix TestStoppedBinaryUpgrade for integration tests.

### DIFF
--- a/test/integration/version_upgrade_test.go
+++ b/test/integration/version_upgrade_test.go
@@ -58,6 +58,26 @@ func installRelease(version string) (f *os.File, err error) {
 	return tf, nil
 }
 
+func legacyVersion() string {
+	// Should be a version from the last 6 months
+	version := "v1.6.2"
+	if KicDriver() {
+		if arm64Platform() {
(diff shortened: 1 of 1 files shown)


medyagh (approved):

medyagh (changes requested):
check build error

andriyDev:
@medyagh 

Build error:
```
Fail to upload '/home/runner/work/minikube/minikube/out/e2e-windows-amd64.exe' due to 'Operation is not valid due to the current state of the object.'.
GitHub.Services.WebApi.VssServiceResponseException: Operation is not valid due to the current state of the object.
```

As far as I can tell, this is just a Github Actions problem, seems to have been an isolated incident, all other binaries were uploaded successfully. This is safe to merge imo.

Issue #12151, opened 2021-08-06
Frequent test failures of `another_fake_test3`


andriyDev:
This test has high flake rates for the following environments:

|Environment|Flake Rate (%)|
|---|---|
|B|95|
|B|30|
|B|15|

Issue #12148, opened 2021-08-06
Frequent test failures of `another_fake_test2`


andriyDev:
This test has high flake rates for the following environments:

|Environment|Flake Rate (%)|
|---|---|
|B|50|

Issue #12147, opened 2021-08-06
Frequent test failures of `another_fake_test`


andriyDev:
This test has high flake rates for the following environments:

|Environment|Flake Rate (%)|
|---|---|
|C|90|
|D|75|

Issue #12138, opened 2021-08-05
Frequent test failures of `FAKE.TEST/PLEASE/IGNORE123`


andriyDev:
This test has high flake rates for the following environments:

|Environment|Flake Rate (%)|
|---|---|
|B|90|

PR #12154, opened 2021-08-06
Create issues automatically for flaky tests

### LINKED ISSUE #12153: Automagically create issues for flaky tests

andriyDev:
fixes #12153.

This change first moves all Jenkins code for flake rate computation into a single script. Then it uses the results of the flake rate computation to filter out tests with high flake rates, and ensures only currently flaky test issues are open.

Example issue created by this PR: #12151
Other test issues: #12148, #12138, #12147

Note since this must be called by Flake Rate Computation, this cannot be tested through ok-to-test

### DIFF
--- a/hack/jenkins/test-flake-chart/compute_flake_rate.sh
+++ b/hack/jenkins/test-flake-chart/compute_flake_rate.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# Copyright 2021 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eux -o pipefail
+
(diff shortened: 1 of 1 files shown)


andriyDev:
Once this is merged, the Jenkins config must be updated for Flake Rate Computation to just run the compute_flake_rate.sh script (and also have the minikube-bot access token).

medyagh:
Sweet. Can we add an identifier or something that ensures only creates issue if not already created by minikube-bot ?

andriyDev:
@medyagh Yep that's already set up! The title of the issue has to be "Frequent test failures of `<test name>`" with label kind/failing-test

andriyDev:
@medyagh I wonder if maybe automatically closing issues is the wrong strategy? Perhaps instead we should just add a comment saying "Flake rates are low now?" Or just update the body to say that? If we do the comment thing, some tests might get many comments if the flake rate straddles the threshold. If we update the body, we might not notice that the flake rate is low, so issues might not get closed by us... Thoughts?

andriyDev:
Talking in person, we have come to the conclusion that we should have both an open threshold and a closing threshold. If a test has a flake rate higher than the open threshold, issues are opened. A test with a flake rate lower than the close threshold will be closed. This way, if a test straddles the open threshold, it will not repeatedly open/close the issue, since we can set the close threshold to be much lower than the open threshold.

andriyDev:
Done, issues will now only be opened if there are tests with a flake rate higher than the OPEN_ISSUE_THRESHOLD, and issues will only be closed if below the CLOSE_ISSUE_THRESHOLD. Any middling flake rates will update the body of the corresponding issue (regardless of if open or closed since that was easier to implement).

medyagh (approved):

PR #12185, opened 2021-08-10
WIP: Add vagrant version of the kicbase docker image

afbjorklund:
Only for VirtualBox and Docker, for now (PoC)

For #9992

This is probably **not** the final solution (running the VM with Vagrant), but allows for testing the OS - rather than a new ISO

https://github.com/kubernetes/community/blob/master/contributors/design-proposals/cluster-lifecycle/local-cluster-ux.md#vagrant

> Since we are targeting a single command install/teardown experience, vagrant needs to be an implementation detail and not be exposed to our users.

The software installation was copied over to the Vagrantfile provisioner from the Dockerfile RUN, so should be the same...

### DIFF
--- a/deploy/kicbase/Vagrantfile
+++ b/deploy/kicbase/Vagrantfile
@@ -1,0 +1,77 @@
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


afbjorklund:
Running with a different provider requires a different box (image), and some different configuration for provider:

```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "ubuntu/focal64"
  config.vm.provider "virtualbox" do |vb|
    vb.cpus = 2
    vb.memory = "2048"
  end
end
```

https://app.vagrantup.com/ubuntu/boxes/focal64

* virtualbox

```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "generic/ubuntu2004"
  config.vm.provider :libvirt do |libvirt|
    libvirt.cpus = 2
    libvirt.memory = "2048"
  end
end
```

https://app.vagrantup.com/generic/boxes/ubuntu2004

* vmware_desktop
* virtualbox
* parallels
* libvirt
* hyperv

The software installation is more a question of copying/pasting from the docker installation, than anything else.

PR #12228, opened 2021-08-10
Print profile driver, container runtime, and Kubernetes version on load

### LINKED ISSUE #11977: All commands should print cluster settings (driver, container runtime) to logs
Currently, in issues or when debugging, it can be hard to figure out what driver and container runtime is being used for any given command. Having these in the logs for all commands (not just minikube start) would be very helpful.

andriyDev:
fixes #11977.

Whenever a profile is loaded, some of its settings are printed out. Whenever profiles are listed, this also logs.

Example: `minikube -p mk2 addons enable dashboard --alsologtostderr`
```
... Stuff ...
I0810 13:39:50.012993  195672 config.go:177] Profile "mk2": Driver=docker, ContainerRuntime=docker, KubernetesVersion=v1.21.3
... Stuff ...
```

### DIFF
--- a/pkg/minikube/config/config.go
+++ b/pkg/minikube/config/config.go
@@ -26,6 +26,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 
+	"k8s.io/klog/v2"
 	"k8s.io/minikube/pkg/minikube/localpath"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/minikube/config/config.go
```diff
@@ -171,7 +172,15 @@ func encode(w io.Writer, m MinikubeConfig) error {
 
 // Load loads the Kubernetes and machine config for the current machine
 func Load(profile string, miniHome ...string) (*ClusterConfig, error) {
-	return DefaultLoader.LoadConfigFromFile(profile, miniHome...)
+	cc, err := DefaultLoader.LoadConfigFromFile(profile, miniHome...)
+	if err == nil {
+		klog.Infof("Profile \"%s\": Driver=%s, ContainerRuntime=%s, KubernetesVersion=%s",
```
medyagh, line 177: lets add more context here,

Loaded Profile config

medyagh (changes requested):

pkg/minikube/config/config.go
andriyDev: Done.

medyagh (approved):

PR #12230, opened 2021-08-10
bump calico to v3.20 and move away from v1beta apis

medyagh:
getting rid of all v1beta apis in our code since it is removed in next kubernetes version
https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/

I hade to bump the calico version to support the New kubernetes API as part of that I also had to move the string yaml file into go:embed

same refactor needs to be done for other CNIs (to use goembed) but that will be done in separate PRs. this one needed to be done to pass the tests


pkg/minikube/bootstrapper/images/images.go
```diff
@@ -163,18 +163,37 @@ func KindNet(repo string) string {
...
+	if repo == "" {
+		repo = "docker.io/calico"
+	}
+	return path.Join(repo, "pod2daemon-flexvol:"+calicoVersion)
+}
+
+// CalicoBin returns image used for calico binary image
+func CalicoBin(repo string) string {
+	if repo == "" {
+		repo = "docker.io/calico"
```
spowelljr, line 196: I think it makes sense to move `"docker.io/calico"` to a const since we're repeating it 4 times.

pkg/minikube/bootstrapper/images/images.go
```diff
@@ -163,18 +163,37 @@ func KindNet(repo string) string {
...
 	}
-	return path.Join(repo, "kube-controllers:v3.14.1")
+	return path.Join(repo, "cni:"+calicoVersion)
```
spowelljr, line 198: Also, the four above funcs are all the same except the pod type/name. Maybe it makes sense to make a common func the funcs call?
```
func calicoCommon(repo string, name string) string {
    if repo == "" {
        repo = "docker.io/calico"
    }
    return path.Join(repo, fmt.Sprintf("%s:%s", name, calicoVersion)
}
```

Could remove the need for my above comment.

medyagh: That's a good idea

sharifelgamal (approved):

PR #12233, opened 2021-08-11
Added port validation

### LINKED ISSUE #10495: Recreate minikube cluster on WSL2 using “minikube delete” and “minikube start --driver=docker --ports...” fails on "connection refused" errors
**Summary**

I can’t recreate minikube cluster on WSL2 when using the `--ports` option to expose ports.
The `minikube start` command fails on connecting to the various ports that is sets up on the Docker container.

The problem does not occur when I don’t use the `--ports` option.

Workarounds:
1. Map the ports used by the previous minikube cluster using an extra `--ports` option, then a new minikube cluster is created.
2. Reboot Windows

**Environment**

* WSL 2 in Windows 10 1909, 18363.1379
* Ubuntu 20.04
* Docker for Windows 3.1.0
* Minikube 1.17.1

1. Reboot Windows without any minikube cluster created

2. Create minikube cluster

    ```
    minikube start --driver=docker --ports=80:80 --ports=443:443
    ```

3. Delete the minikube cluster

    ```
    minikube delete
    ```

4. Recreate the minikube cluster and enable logging to see the error

    ```
    minikube start --driver=docker --ports=80:80 --ports=443:443 -v=1 --alsologtostderr
    ```
    The command now reports errors like:

    ```
    I0217 11:05:40.014928    3353 main.go:119] libmachine: Error dialing TCP: dial tcp 127.0.0.1:49160: connect: connection refused
    ```

5. **WORKAROUND #1**: Recreate the minikube cluster mapping the ports used by the previous minikube cluster

    1. Lookup ports used by minikube

        ```
        docker ps
        ```

        Sample response:
        ```
        f73deb28d024   gcr.io/k8s-minikube/kicbase:v0.0.17   "/usr/local/bin/entr…"   About a minute ago   Up About a minute   0.0.0.0:80->80/tcp, 0.0.0.0:443->443/tcp, 127.0.0.1:49160->22/tcp, 127.0.0.1:49159->2376/tcp, 127.0.0.1:49158->5000/tcp, 127.0.0.1:49157->8443/tcp   minikube
        ```

    2.  Recreate the minikube cluster and map the ports used by the previous minikube cluster

        ```
        minikube delete
        minikube start --driver=docker --ports=80:80 --ports=443:443 --ports=49157-49160:49157-49160
        ```

    Sometimes I have to repeat this procedure a couple of times to get rid of the connection refused error

6. **WORKAROUND #2**: Delete the minikube cluster and reboot Windows

    Trying to avoid this workaround, but sometimes it seems to be the only way out of the problem...

**NOTE**: Recreating the minikube cluster without the `--ports` option works without any problem:

These commands can be run multiple times without any errors:

```
minikube delete
```

kadern0:
Fixes #10495 

Tests:
```
minikube start --driver=docker --ports=asdf:80 --ports=443:443
😄  minikube v1.22.0 on Ubuntu 20.04

❌  Exiting due to MK_USAGE: Sorry, one of the ports provided with --ports flag is not valid [asdf:80 443:443]
```
```
minikube start --driver=docker --ports=8080:80 --ports=443:443
😄  minikube v1.22.0 on Ubuntu 20.04

❌  Exiting due to MK_USAGE: Sorry, you cannot use privileged ports on the host (below 1024) [8080:80 443:443]
```
```
minikube start --driver=docker --ports=8080:80 --ports=6443:443
😄  minikube v1.22.0 on Ubuntu 20.04
🚜  Pulling base image ...
🐳  Preparing Kubernetes v1.21.3 on Docker 20.10.7 ...
    ▪ Using image k8s.gcr.io/ingress-nginx/controller:v0.44.0
    ▪ Using image docker.io/jettech/kube-webhook-certgen:v1.5.1
    ▪ Using image docker.io/jettech/kube-webhook-certgen:v1.5.1
🔎  Verifying ingress addon...
🌟  Enabled addons: storage-provisioner, default-storageclass, ingress

```


cmd/minikube/cmd/start.go
```diff
@@ -1314,6 +1318,25 @@ func validateListenAddress(listenAddr string) {
 	}
 }
 
+// This function validates that the --ports are not below 1024 for the host and not outside range
+func validatePorts(ports []string) {
+	for _, portDuplet := range ports {
+		for i, port := range strings.Split(portDuplet, ":") {
+			p, err := strconv.Atoi(port)
+			if err == nil {
```
spowelljr, line 1326: Small nit, but if you did `if err != nil` instead there would be less nesting and would look a bit cleaner.

```
if err != nil {
    exit.Message(...)
}
if p > 65535 { ... }
if p < 1024 { ... }
```

kadern0: And that's exactly what I should have done :)

spowelljr:
Thanks for the validation @kadern0, it looks good! If I could ask you to do one more thing, if you wouldn't mind writing a test for this validation in `start_test.go`.

If you call https://github.com/kubernetes/minikube/blob/a70bdb5e4b949334c25a3cdd224edcc0368bc11f/cmd/minikube/cmd/start_flags.go#L252 in the test it will run flag validation, there are other tests in `start_test.go` that do this exact same thing if you're looking for an example.

Thanks!

kadern0:
@spowelljr thanks for the feedback! 
Adding tests for this function might not be as straightforward (for me, at least) as it seems. Unlike the other tests on the file, this function doesn't return anything (same as the other validation functions) and unlike the other two tests that call 'generateClusterConfig', we can't test the function from there since this function is called from 'runStart'. So unless I'm missing anything, options are: no test (as with most of validating functions), refactor and add a return statement and test the function directly (although this isn't being done for the other validation functions), making a call to the function during the configuration generation so we can test it from there. 
What do you think?

cmd/minikube/cmd/start.go
```diff
@@ -1314,6 +1318,24 @@ func validateListenAddress(listenAddr string) {
 	}
 }
 
+// This function validates that the --ports are not below 1024 for the host and not outside range
+func validatePorts(ports []string) {
+	for _, portDuplet := range ports {
+		for i, port := range strings.Split(portDuplet, ":") {
+			p, err := strconv.Atoi(port)
+			if err != nil {
+				exit.Message(reason.Usage, "Sorry, one of the ports provided with --ports flag is not valid {{.ports}}", out.V{"ports": ports})
```
medyagh, line 1327: how about making this return fmt.Errorf("error string...")
and then Exit in the caller with the error message

and then we could add a unit test for

validatePort in the start_test.go

medyagh (changes requested):

medyagh:
> @spowelljr thanks for the feedback!
> Adding tests for this function might not be as straightforward (for me, at least) as it seems. Unlike the other tests on the file, this function doesn't return anything (same as the other validation functions) and unlike the other two tests that call 'generateClusterConfig', we can't test the function from there since this function is called from 'runStart'. So unless I'm missing anything, options are: no test (as with most of validating functions), refactor and add a return statement and test the function directly (although this isn't being done for the other validation functions), making a call to the function during the configuration generation so we can test it from there.
> What do you think?

how about adding a unit test in start_test.go 
just for this func
```
validatePorts(ports []string)
```

kadern0:
@medyagh so the plan is: refactoring the function so it returns error and then create a unit test for it, right?

cmd/minikube/cmd/start.go
kadern0: I just followed the "style" of the rest of validation functions but this makes total sense

medyagh:
> @medyagh so the plan is: refactoring the function so it returns error and then create a unit test for it, right?

Yes please . Thank you for your helpful prs

kadern0:
@medyagh I've made the changes as suggested but now it seems the cyclomatic complexity of the function `validateFlags` is > 30. It seems it might require additional changes.

kadern0:
Maybe these could be made functions in order to reduce the complexity of `validateFlags`:

https://github.com/kubernetes/minikube/blob/c3c4d0455dfed89650fdf54f9f70d551912b4969/cmd/minikube/cmd/start.go#L1090:L1099
https://github.com/kubernetes/minikube/blob/c3c4d0455dfed89650fdf54f9f70d551912b4969/cmd/minikube/cmd/start.go#L1121:L1145
https://github.com/kubernetes/minikube/blob/c3c4d0455dfed89650fdf54f9f70d551912b4969/cmd/minikube/cmd/start.go#L1147:L1164

What do you think?

kadern0:
@medyagh , @spowelljr I think this PR is ready for another review :)

spowelljr:
@kadern0 The code looks good! One last thing, could you run `make generate-docs` and push the changes up and then we'll be good to go, thanks for your patience.

spowelljr (changes requested):

cmd/minikube/cmd/start.go
```diff
@@ -1088,13 +1088,9 @@ func validateCPUCount(drvName string) {
...
 	if cmd.Flags().Changed(humanReadableDiskSize) {
-		diskSizeMB, err := util.CalculateSizeInMB(viper.GetString(humanReadableDiskSize))
+		err := validateDiskSize(viper.GetString(humanReadableDiskSize))
 		if err != nil {
-			exitIfNotForced(reason.Usage, "Validation unable to parse disk size '{{.diskSize}}': {{.error}}", out.V{"diskSize": viper.GetString(humanReadableDiskSize), "error": err})
-		}
-
-		if diskSizeMB < minimumDiskSize {
-			exitIfNotForced(reason.RsrcInsufficientStorage, "Requested disk size {{.requested_size}} is less than minimum of {{.minimum_size}}", out.V{"requested_size": diskSizeMB, "minimum_size": minimumDiskSize})
+			exitIfNotForced(reason.RsrcInsufficientStorage, "Requested disk size {{.requested_size}} is less than minimum of {{.minimum_size}}", out.V{"requested_size": humanReadableDiskSize, "minimum_size": minimumDiskSize})
```
spowelljr, line 1093: Could you do the `{{.err}}` format that you did for the other ones below here. The `validateDiskSize` function can return two different types of errors, insufficient storage or invalid disk size, but we're hard coding the error here to insufficient storage and not leveraging the error that gets returned instead.

spowelljr (changes requested):

spowelljr (approved):

PR #12243, opened 2021-08-12
Fix test failures occuring in partially-"branching" tests

andriyDev:
Should drop flake rates of #12206, #12188, #12223.

These are only the partially-branching tests I've been able to spot. A branching test is any test/subtest without any subtests. A partially-branching test is a branching test where a failure could occur outside any of its subtests. This causes incredibly high flake rates for the partially-branching tests even if those tests may pass more of the time. This is because once test2json sees a subtest, it no longer reports the branching test's success. So only failures are reported in partially-branching tests, whereas successes are suppressed, causing 100% flake rates (even if failures are rare).

All tests should avoid containing partially-branching tests. The way I've fixed this here, is to just take any un-wrapped code and put it in a subtest like Setup or whatever better fits.

Note this DOES NOT fix any of the underlying issues for these tests, this will just fix how the errors are reported, giving us a better indication of how often they are failing, and how much of a problem these issues are.

### DIFF
--- a/test/integration/addons_test.go
+++ b/test/integration/addons_test.go
@@ -46,7 +46,7 @@ func TestAddons(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), Minutes(40))
 	defer Cleanup(t, profile, cancel)
 
-	setupFailed := t.Run("Setup", func(t *testing.T) {
+	setupSucceeded := t.Run("Setup", func(t *testing.T) {
 		// We don't need a dummy file is we're on GCE
 		if !detect.IsOnGCE() || detect.IsCloudShell() {
 			// Set an env var to point to our dummy credentials file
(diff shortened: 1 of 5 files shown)


medyagh (commented):
great idea ! thank you for picking up this task, 

lets ensure that the tests which are meant to be Serial do NOT run in parallel after this refactor

for example in your PR the "Upgrade" should always happen after previous things

andriyDev:
@medyagh I'm pretty sure since the Upgrade subtest doesn't call MaybeParallel, it will be run in Serial, so MinikubeLogs should happen after Upgrade

andriyDev:
I went through all other tests as best I could, and these last two commits are the only places that could have had a similar problem. Not sure if they did, but now the tests are safer.

andriyDev:
Added a check to prevent addon tests from running if setup fails. Also added proper cleanup since defer won't work when inside the subtest.

PR #12262, opened 2021-08-13
original pr 12162 with fix

prezha:
original pr #12162 + fix for:
`pkg/minikube/image/image.go:237:27: cannot use ref (type name.Reference) as type name.Tag in argument to daemon.Write: need type assertion`


prezha:
closed in favour of https://github.com/kubernetes/minikube/pull/12265

Issue #12232, opened 2021-08-10
Fix  `minikube start && stop && start` in Cloud Shell: ubepods.slice already exists." 


spowelljr:
Error Message 
```
Aug 16 21:40:19 minikube kubelet[8019]: E0816 21:40:19.352054 8019 kubelet.go:1384] "Failed to start ContainerManager" err="Unit kubepods.slice already exists." 
```

Running:
```
minikube start
minikube stop
minikube start
```
In Cloud Shell, results in the following failure on the second start.

[lastStart.txt](https://github.com/kubernetes/minikube/files/6964986/lastStart.3.txt)

medyagh:
we should add this as the solution message whenever minikube fails this way

medyagh:
for record here is a full log including exit

<details>
<code>
$ minikube start
* minikube v1.22.0 on Debian 10.10 (amd64)
  - MINIKUBE_FORCE_SYSTEMD=true
  - MINIKUBE_HOME=/google/minikube
  - MINIKUBE_WANTUPDATENOTIFICATION=false
* Using the docker driver based on existing profile
* Starting control plane node minikube in cluster minikube
* Pulling base image ...
* Restarting existing docker container for "minikube" ...
* Preparing Kubernetes v1.21.2 on Docker 20.10.7 ...
X Problems detected in kubelet:
  Aug 16 21:40:15 minikube kubelet[7575]: E0816 21:40:15.634392    7575 kubelet.go:1384] "Failed to start ContainerManager" err="Unit kubepods.slice already exists."
  Aug 16 21:40:17 minikube kubelet[7799]: E0816 21:40:17.545005    7799 kubelet.go:1384] "Failed to start ContainerManager" err="Unit kubepods.slice already exists."
  Aug 16 21:40:19 minikube kubelet[8019]: E0816 21:40:19.352054    8019 kubelet.go:1384] "Failed to start ContainerManager" err="Unit kubepods.slice already exists."
  Aug 16 21:40:21 minikube kubelet[8239]: E0816 21:40:21.090448    8239 kubelet.go:1384] "Failed to start ContainerManager" err="Unit kubepods.slice already exists."
  Aug 16 21:40:22 minikube kubelet[8460]: E0816 21:40:22.914457    8460 kubelet.go:1384] "Failed to start ContainerManager" err="Unit kubepods.slice already exists."
X Problems detected in kubelet:
[...]

PR #12279, opened 2021-08-16
Add solution message for cgroup related issues on nested cloud-shell

medyagh:
Helps with https://github.com/kubernetes/minikube/issues/12232 but doesn't close it yet

### DIFF
--- a/pkg/minikube/reason/known_issues.go
+++ b/pkg/minikube/reason/known_issues.go
@@ -177,6 +177,37 @@ var hostIssues = []match{
 	},
 	{
 		Kind: Kind{
+			ID:       "HOST_CGROUP_NOT_SUPPORTED",
+			ExitCode: ExHostUnsupported,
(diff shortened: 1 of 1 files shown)


pkg/minikube/reason/known_issues.go
```diff
@@ -175,6 +175,37 @@ var hostIssues = []match{
 		},
 		Regexp: re(`Container.*is not running.*chown docker:docker`),
 	},
+	{
+		Kind: Kind{
+			ID:       "HOST_CGROUP_NOT_SUPPORTED",
+			ExitCode: ExHostUnsupported,
+			Advice: `CGroup allocation is not available in your environment, You are possibly running minikube in a nested container, try run with these options:
```
sharifelgamal, line 182:
```suggestion
			Advice: `CGroup allocation is not available in your environment. You might be running minikube in a nested container. Try running:
```

sharifelgamal: for both

PR #12298, opened 2021-08-17
wip: Replace --profile with --cluster and make --profile an alias for --cluster

### LINKED ISSUE #12275: Allow using `minikube --cluster whatever` in place of --profile flag
`minikube --profile` can be confusing for some users, and we have had several requests to add multi-cluster support (even though we've had it for a while). To mitigate this, we should add a `minikube --cluster` flag as an alias for profile.

andriyDev:
fixes #12275

### DIFF
--- a/cmd/minikube/cmd/config/profile.go
+++ b/cmd/minikube/cmd/config/profile.go
@@ -97,3 +97,7 @@ var ProfileCmd = &cobra.Command{
 		}
 	},
 }
+
+func init() {
+	ProfileCmd.Aliases = append(ProfileCmd.Aliases, "cluster")
+}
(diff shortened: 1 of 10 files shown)


medyagh:
@andriyDev  please check the unit test

medyagh:
I like making it more clear, but I am not sure what is the best thing to do here..since there is no short-hand for --cluster
most users might prefer a shorter flag (like -p)

in this case we are introducing a new confusion that -p is short for --cluster 

is it possible to do something else ?

maybe -cn short for --cluster-name

sharifelgamal:
as far as I know, viper requires the shorthand to be a single letter, and we know we can't use `-c` for backwards compat reasons, which is why we decided to not touch it and allow people to continue using `-p` if they so desired.

sharifelgamal:
the unit test failure is fixed at HEAD, so that should just need a rebase.

andriyDev:
@medyagh My argument for that is that if people are trying to use shorthands, they're likely to look them up in the help. And for all commands except for one (I think `image save` or something), -c has no meaning, so an error is likely to get them to look it up there too. Shorthands should be for more advanced usage anyway imo? So the fact that -p doesn't align with --cluster isn't too big a deal. That's my 2 cents.

sharifelgamal:
Let's close this and give it another shot a different time.

Issue #11275, opened 2021-05-04
Pod Security Policies addon blocks ingress addon


awreece:
I'd like to use both Pod Security Policies and ingress. Unfortunately, this doesn't appear to work.

```
% minikube start --vm=true --addons=ingress --addons=pod-security-policy --extra-config=apiserver.enable-admission-plugins=PodSecurityPolicy
😄  minikube v1.19.0 on Darwin 11.2.3
✨  Automatically selected the hyperkit driver
👍  Starting control plane node minikube in cluster minikube
🔥  Creating hyperkit VM (CPUs=2, Memory=6000MB, Disk=20000MB) ...
🐳  Preparing Kubernetes v1.20.2 on Docker 20.10.4 ...
    ▪ apiserver.enable-admission-plugins=PodSecurityPolicy
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
    ▪ Using image k8s.gcr.io/ingress-nginx/controller:v0.44.0
    ▪ Using image docker.io/jettech/kube-webhook-certgen:v1.5.1
    ▪ Using image docker.io/jettech/kube-webhook-certgen:v1.5.1
🔎  Verifying ingress addon...
🌟  Enabled addons: storage-provisioner, pod-security-policy, default-storageclass, ingress
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default
```

Unfortunately, this is a bit of a lie -- ingress is _not_ actually working:

```
% kubectl get deployments -A
NAMESPACE       NAME                       READY   UP-TO-DATE   AVAILABLE   AGE
ingress-nginx   ingress-nginx-controller   0/1     0            0           3m29s
kube-system     coredns                    1/1     1            1           3m31s
```

Digging in, we can see that the replicaset is getting denied by the PodSecurityPolicy:

```
$ kubectl -n ingress-nginx get events  --sort-by='.lastTimestamp' 
...
[...]

PR #12313, opened 2021-08-19
Use privileged pod security policy for nginx ingress addon

jroes:
This is a quick draft that is intended to be a fix for #11275. Mostly creating this PR to get CI running on it so I don't need to jump through all the hoops locally yet :).

Some open items:

- [ ] This assumes the cluster is using pod security policies, which appears to be a separate addon. Is this acceptable when that addon is not installed in the cluster?
- [ ] Is the privileged policy too permissive? I don't think we have another option unless we venture outside of the new PodSecurity model.
- [ ] Untested outside of my workaround in #11275.

### DIFF
--- a/deploy/addons/ingress/ingress-rbac.yaml.tmpl
+++ b/deploy/addons/ingress/ingress-rbac.yaml.tmpl
@@ -79,7 +79,7 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: ingress-nginx
+  name: ingress-nginx-privileged
   labels:
     app.kubernetes.io/name: ingress-nginx
     app.kubernetes.io/instance: ingress-nginx
@@ -87,11 +87,11 @@ metadata:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: ingress-nginx
+  name: psp:privileged
 subjects:
-- kind: ServiceAccount
-  name: ingress-nginx
-  namespace: ingress-nginx
+- kind: Group
+  name: system:serviceaccounts:ingress-nginx
+  apiGroup: rbac.authorization.k8s.io
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role


medyagh:
@jroes thank you for contributing to minikube, 
I believe PSPs are depreciated ...I think this could cause breaking ingress for future kuberentes versions
https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/

jroes:
@medyagh Thanks for the feedback!

The way I understood the blog post was that the replacement, PodSecurity, is not available quite yet, and the recommendation is to use the new [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/).

From the [blog post](https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/) you shared:

> You can approximate the functionality of PSP Replacement Policy today by using the Pod Security Standards’ PSPs. If you set the cluster default by binding a Baseline or Restricted policy to the system:serviceaccounts group, and then make a more-permissive policy available as needed in certain Namespaces using ServiceAccount bindings, you will avoid many of the PSP pitfalls and have an easy migration to PSP Replacement Policy. If your needs are much more complex than this, your effort is probably better spent adopting one of the more fully-featured external admission controllers mentioned above.

And the [recommendation](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#recommended-practice) from the latest PodSecurityPolicy docs:

> Follow these guidelines to simplify migration from PodSecurityPolicy to the new admission controller:
> 1. Limit your PodSecurityPolicies to the policies defined by the Pod Security Standards:
> 2. Only bind PSPs to entire namespaces, by using the system:serviceaccounts:<namespace> group (where <namespace> is the target namespace). [...]

Let me know if I've got this wrong, I'm just getting up to speed myself! :)

medyagh:
I would need more context on what would be possible side effects of this change, since this is a popular addon, I wanna ensure we dont break other use cases...

could this be possiblily a Separate addon ? or something we could configure through
something like
```
minikube addons configure ingress PSP...... ?
```

jroes:
I'm not familiar with `minikube addons configure` but that sounds reasonable.

Creating another addon seems rather straightforward based on a quick glance, looks like adding a new directory and corresponding file [here](https://github.com/kubernetes/minikube/tree/master/deploy/addons).

Happy to help move it in whatever direction you feel like is best for the project, let me know how I can help. :)

medyagh:
> 
@jroes sorry for the long wait on the PR review
I would be okay with both ways, either a new addon or a configure field for ingress addon. maybe a new addon would be easier

spowelljr:
This PR hasn't been pushed to for a while since the last comment so I'm going to close this issue. If you come back to it feel free to reopen this issue, thanks!

PR #12318, opened 2021-08-20
pause each container separately

sharifelgamal:
Our containerd pause tests are failing with the following error:
```
X Exiting due to GUEST_PAUSE: runc: sudo runc --root /run/containerd/runc/k8s.io pause 44e5c29356861d3eb4eea140081d1869eb05d894f0a088323bd8c4db9fc8ed35 508221fbe40ab163913f6c34d29dd4548e075546b675d1ec9f56fffbb8bc48d7: Process exited with status 1
	stdout:
	Incorrect Usage.
	
	NAME:
	   runc pause - pause suspends all processes inside the container
	
	USAGE:
	   runc pause <container-id>
```

The issue is that we're appending each container id as we pause each one, which is not correct. This change fixes that.

### DIFF
--- a/pkg/minikube/cruntime/cri.go
+++ b/pkg/minikube/cruntime/cri.go
@@ -134,9 +134,8 @@ func pauseCRIContainers(cr CommandRunner, root string, ids []string) error {
 		args = append(args, "--root", root)
 	}
 	args = append(args, "pause")
-	cargs := args
 	for _, id := range ids {
-		cargs = append(cargs, id)
+		cargs := append(args, id)
 		if _, err := cr.RunCmd(exec.Command("sudo", cargs...)); err != nil {
 			return errors.Wrap(err, "runc")
 		}
--- a/pkg/minikube/cruntime/cri.go
+++ b/pkg/minikube/cruntime/cri.go
@@ -135,8 +135,8 @@ func pauseCRIContainers(cr CommandRunner, root string, ids []string) error {
 	}
 	args = append(args, "pause")
 	for _, id := range ids {
-		cargs := append(args, id)
-		if _, err := cr.RunCmd(exec.Command("sudo", cargs...)); err != nil {
+		args := append(args, id)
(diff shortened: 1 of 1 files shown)


pkg/minikube/cruntime/cri.go
```diff
@@ -134,10 +134,9 @@ func pauseCRIContainers(cr CommandRunner, root string, ids []string) error {
 		args = append(args, "--root", root)
 	}
 	args = append(args, "pause")
-	cargs := args
 	for _, id := range ids {
-		cargs = append(cargs, id)
-		if _, err := cr.RunCmd(exec.Command("sudo", cargs...)); err != nil {
+		args := append(args, id)
```
afbjorklund, line 138: you still need a local variable, no ? maybe call it something different than the global one, less confusion ?

sharifelgamal: if I do something like `cargs := append(args, id)` then golint complains

sharifelgamal:
```
pkg/minikube/cruntime/cri.go:138:12: appendAssign: append result not assigned to the same slice (gocritic)
		cargs := append(args, id)
```

sharifelgamal:
The Docker Linux containerd pause tests failed, but for reasons unrelated to the fix I'm making here.

pkg/minikube/cruntime/cri.go
medyagh: How about keeping the cargs, I am not confident what would be side effects of getting rid of that local cargs? Would at some point the inner args append stuff to the outer args ?

sharifelgamal: the local cargs is what's causing the bug. i'm redefining args inside of the loop, so tainting the outer args isn't going to happen.

medyagh: gotcha

medyagh (approved):

Issue #12374, opened 2021-08-30
Update bump golang version logic after Kube-cross image change


medyagh:
This PR broke our atuomation by adding a "Prefix" to the golang image
v1.16.7-1 ---> v1.23.0-go1.17-buster.0

https://github.com/kubernetes/kubernetes/pull/103692

medyagh:
https://github.com/kubernetes/minikube/pull/12373

PR #12375, opened 2021-08-30
Fix go bump automation to adopt Kubernetes new kube-cross image name

medyagh:
closes https://github.com/kubernetes/minikube/issues/12374


sharifelgamal:
do we have to care about older versions of golang that doesn't have the k8sversion prefix?

medyagh:
> do we have to care about older versions of golang that doesn't have the k8sversion prefix?

no since we are not automating older versions

sharifelgamal (approved):

PR #12398, opened 2021-09-02
remove ppc64le/armv7 from kicbase archs

sharifelgamal:
there are some packages missing for ppc64le that we need to kicbase, so remove it for now until we can fix it.


sharifelgamal:
ok-to-build-image

medyagh (approved):
merge if it builds

sharifelgamal:
ok-to-build-image

PR #12432, opened 2021-09-08
Fix GCE VMs using service account

amontalban:
In #12231 the logic for GCE instances using a service account was broken, the error message was:

```
W0908 14:11:49.063401   16613 out.go:242] * Exiting due to MK_CREDENTIALS_NOT_FOUND: Could not find any GCP credentials. Either run `gcloud auth application-default login` or set the GOOGLE_APPLICATION_CREDENTIALS environment variable to the path of your credentials file.
* Exiting due to MK_CREDENTIALS_NOT_FOUND: Could not find any GCP credentials. Either run `gcloud auth application-default login` or set the GOOGLE_APPLICATION_CREDENTIALS environment variable to the path of your credentials file.
I0908 14:11:49.065011   16613 out.go:177] 
```

With this patch the output is:

```
ubuntu@lb-minikube-dev-test:~$ minikube addons enable gcp-auth
! It seems that you are running in GCE, which means authentication should work without the GCP Auth addon. If you would still like to authenticate using a credentials file, use the --force flag.
* The 'gcp-auth' addon is enabled
```

And I can now download private images from GCR.

cc @sharifelgamal

### DIFF
--- a/pkg/addons/addons_gcpauth.go
+++ b/pkg/addons/addons_gcpauth.go
@@ -69,19 +69,32 @@ func enableAddonGCPAuth(cfg *config.ClusterConfig) error {
 	cc := mustload.Running(cfg.Name)
 	r := cc.CP.Runner
 
-	// Grab credentials from where GCP would normally look
 	ctx := context.Background()
+
+	// We need to detect early if we use GOOGLE_APPLICATION_CREDENTIALS and have the --force flag set because
+	// google.FindDefaultCredentials() reads this setting
+	if credentialsFile := os.Getenv("GOOGLE_APPLICATION_CREDENTIALS"); credentialsFile != "" && !Force {
+		exit.Message(reason.Usage, "It seems that you are running in GCE and have GOOGLE_APPLICATION_CREDENTIALS environment variable set, if you would still like to authenticate using a credentials file use the --force flag.")
+	}
+
+	// Grab credentials from where GCP would normally look
 	creds, err := google.FindDefaultCredentials(ctx)
-	if err != nil || creds.JSON == nil {
+	if err != nil {
 		if detect.IsCloudShell() {
 			if c := os.Getenv("CLOUDSDK_CONFIG"); c != "" {
 				f, err := ioutil.ReadFile(path.Join(c, "application_default_credentials.json"))
 				if err == nil {
 					creds, _ = google.CredentialsFromJSON(ctx, f)
+				} else {
+					exit.Message(reason.InternalCredsNotFound, "Could not find any GCP credentials. Either run `gcloud auth application-default login` or set the GOOGLE_APPLICATION_CREDENTIALS environment variable to the path of your credentials file.")
 				}
 			}
 		} else {
 			exit.Message(reason.InternalCredsNotFound, "Could not find any GCP credentials. Either run `gcloud auth application-default login` or set the GOOGLE_APPLICATION_CREDENTIALS environment variable to the path of your credentials file.")
+		}
+	} else {
+		if _, credentialsPresent := os.LookupEnv("GOOGLE_APPLICATION_CREDENTIALS"); !credentialsPresent && detect.IsOnGCE() {
+			out.WarningT("It seems that you are running in GCE, we will use the attached service account.")
 		}
 	}
 
@@ -90,12 +103,6 @@ func enableAddonGCPAuth(cfg *config.ClusterConfig) error {
 	err = createPullSecret(cfg, creds)
 	if err != nil {
 		return errors.Wrap(err, "pull secret")
-	}
-
-	// If the env var is explicitly set, even in GCE, then defer to the user and continue
-	if !Force && detect.IsOnGCE() && os.Getenv("GOOGLE_APPLICATION_CREDENTIALS") == "" {
-		out.WarningT("It seems that you are running in GCE, which means authentication should work without the GCP Auth addon. If you would still like to authenticate using a credentials file, use the --force flag.")
-		return nil
 	}
 
 	// Actually copy the creds over


amontalban:
I have just signed CLA

medyagh:
CC: @sharifelgamal

sharifelgamal:
Ah, good catch. Thanks for the fix.

pkg/addons/addons_gcpauth.go
```diff
@@ -80,9 +80,9 @@ func enableAddonGCPAuth(cfg *config.ClusterConfig) error {
 					creds, _ = google.CredentialsFromJSON(ctx, f)
 				}
 			}
-		} else {
-			exit.Message(reason.InternalCredsNotFound, "Could not find any GCP credentials. Either run `gcloud auth application-default login` or set the GOOGLE_APPLICATION_CREDENTIALS environment variable to the path of your credentials file.")
 		}
+	} else {
+		exit.Message(reason.InternalCredsNotFound, "Could not find any GCP credentials. Either run `gcloud auth application-default login` or set the GOOGLE_APPLICATION_CREDENTIALS environment variable to the path of your credentials file.")
```
sharifelgamal, line 85: On second look, this doesn't seem to be doing the right thing. This `else` will get called if err is nil and the creds json is not nil, which is exactly the opposite of what we want. 

We should be getting by this if statement if your service account is authenticated correctly and the JSON is in a place we're expecting.

amontalban: You're right, I rushed a little bit to unblock us.

Let me push a new commit.

amontalban: @sharifelgamal let me know what do you think now, basically if we have an error when retrieving credentials or they are empty, we first check if it is running in CloudShell or GCE, if its not the case we fail with `InternalCredsNotFound`.

pkg/addons/addons_gcpauth.go
```diff
@@ -80,6 +80,10 @@ func enableAddonGCPAuth(cfg *config.ClusterConfig) error {
 					creds, _ = google.CredentialsFromJSON(ctx, f)
 				}
 			}
+		} else if !Force && detect.IsOnGCE() && os.Getenv("GOOGLE_APPLICATION_CREDENTIALS") == "" {
```
amontalban, line 83: @sharifelgamal sorry, after testing this new change I see that the `return nil` is wrong. If the secret `gcp-auth` is not created then pulling images does not work.

Unless I'm missing some context, I would like to give it another try.

sharifelgamal: Yeah, I think your use-case is that creds.JSON is nil (but getting the credentials doesn't return an error). We should handle adding the pull secret in that case (as long as creds.TokenSource is not nil). I think the best way to do that would be to change where the createPullSecret call gets made.

Instead of `if err != nil || creds.JSON == ""`, we should change that statement to `if err != nil`

then we can create the pull secret

then we can check if creds.JSON is empty after the call to createPullSecret. if it is empty, then return the same error as in the err != nil if statement.

does that make sense?

amontalban: I took a slightly different route and basically checked if `GOOGLE_APPLICATION_CREDENTIALS` is set and `--force` was present, if that is the case then `google.FindDefaultCredentials` will use it.

If that fails, then we do the `detect.IsCloudShell` part and finally if everything else failed and `GOOGLE_APPLICATION_CREDENTIALS` is not set and `detect.IsOnGCE` is true then we just show a warning that the attached service account will be used and continue.

This are my tests cases:

- **With service account**

```
ubuntu@lb-minikube-dev-test:~/minikube$ minikube addons enable gcp-auth
! It seems that you are running in GCE, we will use the attached service account.
* The 'gcp-auth' addon is enabled
```

- **With an empty credentials file (Just to test potential failure) without force**

```
ubuntu@lb-minikube-dev-test:~/minikube$ env GOOGLE_APPLICATION_CREDENTIALS="/home/ubuntu/gcp.json" minikube addons enable gcp-auth

X Exiting due to MK_USAGE: It seems that you are running in GCE and have GOOGLE_APPLICATION_CREDENTIALS environment variable set, if you would still like to authenticate using a credentials file use the --force flag.
```

- **With an empty credentials file (Just to test potential failure) without force**

```
ubuntu@lb-minikube-dev-test:~/minikube$ env GOOGLE_APPLICATION_CREDENTIALS="/home/ubuntu/gcp.json" minikube addons enable gcp-auth --force

* Exiting due to MK_CREDENTIALS_NOT_FOUND: Could not find any GCP credentials. Either run `gcloud auth application-default login` or set the GOOGLE_APPLICATION_CREDENTIALS environment variable to the path of your credentials file.
```

- **With a valid credentials file**

```
ubuntu@lb-minikube-dev-test:~/minikube$ env GOOGLE_APPLICATION_CREDENTIALS="/home/ubuntu/gcp.json" minikube addons enable gcp-auth --force
  - Using image k8s.gcr.io/ingress-nginx/kube-webhook-certgen:v1.0
  - Using image gcr.io/k8s-minikube/gcp-auth-webhook:v0.0.7
* Verifying gcp-auth addon...
* Your GCP credentials will now be mounted into every pod created in the minikube cluster.
* If you don't want your credentials mounted into a specific pod, add a label with the `gcp-auth-skip-secret` key to your pod configuration.
* If you want existing pods to be mounted with credentials, either recreate them or rerun addons enable with --refresh.
```

pkg/addons/addons_gcpauth.go
```diff
@@ -69,20 +69,33 @@ func enableAddonGCPAuth(cfg *config.ClusterConfig) error {
...
 					creds, _ = google.CredentialsFromJSON(ctx, f)
+				} else {
+					exit.Message(reason.InternalCredsNotFound, "Could not find any GCP credentials. Either run `gcloud auth application-default login` or set the GOOGLE_APPLICATION_CREDENTIALS environment variable to the path of your credentials file.")
 				}
 			}
 		} else {
 			exit.Message(reason.InternalCredsNotFound, "Could not find any GCP credentials. Either run `gcloud auth application-default login` or set the GOOGLE_APPLICATION_CREDENTIALS environment variable to the path of your credentials file.")
 		}
+	} else {
+		if _, credentialsPresent := os.LookupEnv("GOOGLE_APPLICATION_CREDENTIALS"); !credentialsPresent && detect.IsOnGCE() {
```
sharifelgamal, line 96: I don't think this is necessary, we can just exit after we add the pull secret with the same message as before. The reason I return nil is because we don't want to actually attempt to start the gcp-auth pods and copy the credentials in if `cred.JSON` is nil. We just need to let GCE's metadata server do its thing. So once the pull secret is added, we should return nil.

sharifelgamal (changes requested):

amontalban:
@sharifelgamal sorry for the delay in my response, I was out on vacation but I saw that you fixed this in the latest release.

Thanks!

PR #12439, opened 2021-09-09
filter out arm64, macOS, and crio from flake reports

sharifelgamal:
We want to filter these out of PR comments until they're more stable and less noisy

### DIFF
--- a/hack/jenkins/test-flake-chart/report_flakes.sh
+++ b/hack/jenkins/test-flake-chart/report_flakes.sh
@@ -87,7 +87,7 @@ awk -F, 'NR>1 {
 
 # Filter out arm64, macOS and windows tests until they're more stable
 TMP_FAILED_RATES_FILTERED=$(mktemp)
-grep -v "arm64\|macOS\|Windows" "$TMP_FAILED_RATES" > "$TMP_FAILED_RATES_FILTERED"
+grep -v "arm64\|macOS\|crio" "$TMP_FAILED_RATES" > "$TMP_FAILED_RATES_FILTERED"
 
 FAILED_RATES_LINES=$(wc -l < "$TMP_FAILED_RATES_FILTERED")
 if [[ "$FAILED_RATES_LINES" -eq 0 ]]; then
--- a/hack/jenkins/test-flake-chart/report_flakes.sh
+++ b/hack/jenkins/test-flake-chart/report_flakes.sh
@@ -85,9 +85,9 @@ awk -F, 'NR>1 {
   | sort -g -t, -k2,2 \
   >> "$TMP_FAILED_RATES"
 
-# Filter out arm64 and macOS tests until they're more stable
+# Filter out arm64, macOS and windows tests until they're more stable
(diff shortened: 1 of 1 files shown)


hack/jenkins/test-flake-chart/report_flakes.sh
```diff
@@ -85,6 +85,10 @@ awk -F, 'NR>1 {
   | sort -g -t, -k2,2 \
   >> "$TMP_FAILED_RATES"
 
+# Filter out arm64 and macOS tests until they're more stable
+grep -v arm64 "$TMP_FAILED_RATES" > "$TMP_FAILED_RATES"
+grep -v macOS ""$TMP_FAILED_RATES" > ""$TMP_FAILED_RATES"
```
spowelljr, line 90: I'm assuming the extra double quotes are a mistake?

sharifelgamal: yeah i fixed it

hack/jenkins/test-flake-chart/report_flakes.sh
```diff
@@ -85,6 +85,10 @@ awk -F, 'NR>1 {
   | sort -g -t, -k2,2 \
   >> "$TMP_FAILED_RATES"
 
+# Filter out arm64 and macOS tests until they're more stable
+grep -v arm64 "$TMP_FAILED_RATES" > "$TMP_FAILED_RATES"
+grep -v macOS "$TMP_FAILED_RATES" > "$TMP_FAILED_RATES"
```
spowelljr, line 90: Could do grep -v "arm64\|macOS" to reduce some duplication but not required

spowelljr (approved):

hack/jenkins/test-flake-chart/report_flakes.sh
sharifelgamal: yeah, that's definitely true, I just wanted to make it super clear and allow for easy deletion.

hack/jenkins/test-flake-chart/report_flakes.sh
```diff
@@ -85,7 +85,11 @@ awk -F, 'NR>1 {
   | sort -g -t, -k2,2 \
   >> "$TMP_FAILED_RATES"
 
-FAILED_RATES_LINES=$(wc -l < "$TMP_FAILED_RATES")
+# Filter out arm64, macOS and windows tests until they're more stable
+TMP_FAILED_RATES_FILTERED=$(mktemp)
+grep -v "arm64\|macOS\|Windows" "$TMP_FAILED_RATES" > "$TMP_FAILED_RATES_FILTERED"
```
medyagh, line 90: I think instead of filtering out Windows

lets filterout Crio and Arm64

medyagh (changes requested):

hack/jenkins/test-flake-chart/report_flakes.sh
sharifelgamal: done

PR #12443, opened 2021-09-09
Fix Ingress Addon

sharifelgamal:

### DIFF
--- a/cmd/minikube/cmd/config/enable.go
+++ b/cmd/minikube/cmd/config/enable.go
@@ -64,7 +64,7 @@ var addonsEnableCmd = &cobra.Command{
 `, out.V{"profileArg": tipProfileArg})
 
 		}
-		if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/minikube/cmd/config/enable.go
```diff
@@ -64,7 +64,7 @@ var addonsEnableCmd = &cobra.Command{
 `, out.V{"profileArg": tipProfileArg})
 
 		}
-		if err != nil {
+		if err == nil {
```
medyagh, line 67: are we sure this was wrong before ? this code been arround long time ...maybe it was a odd logic but correct ?

sharifelgamal: yeah, this bug was introduced in https://github.com/kubernetes/minikube/pull/12231

medyagh (approved):

Issue #12402, opened 2021-09-03
Minikube dashboard fails with error 


rvishwas:
I have setup minikube as per the given URL https://minikube.sigs.k8s.io/docs/start/ using docker as default driver.
I am able to run minikube kubectl successfully, but not able to start dashboard.

**Steps to reproduce the issue:** 

1.  minikube start   
2.  minikube dashboard   

**Run `minikube logs --file=logs.txt` and drag and drop the log file into this issue**

**Full output of failed command if not `minikube start`:**
<details>
minikube dashboard command fails with an error 
![image](https://user-images.githubusercontent.com/1425360/131973381-dd3bec0d-9dfc-4df7-90f8-b3245fc5eaa6.png)

(https://github.com/kubernetes/minikube/files/7104307/minikube_dashboard_3decaa05cd245e29205d0e6649b53eb77b34f0cb_0.log)
[](url)

</details>

shadowshot-x:
I also encountered a similar issue while using minikube on Ubuntu 16.04. There might be an issue in configuration. Are you using Kubernetes tools other than minikube on your system separately? [Try to uninstall them and install only minikube.](https://github.com/kubernetes/minikube/issues/6278)

I resolved my issues with the following steps

1.` minikube delete`
2. `docker system prune`
3. `minikube start --driver=docker`

PR #12475, opened 2021-09-14
Use newer dashboard, to support ingress extensions

leifcr:
This should solve #12402 and issues where ingresses will not show properly in the dashboard.
This error shows on kubernetes 1.22

It bumps the dashboard and metrics to newest versions


medyagh (changes requested):
@leifcr please fix the unit test and also the Preload Version needs to be bumped after this PR

medyagh (approved):

leifcr:
@sharifelgamal and @medyagh Thanks for fixing unit tests and getting this merged. I'll dig further into the code, so I can do that the next time.

PR #12487, opened 2021-09-16
doc: be more specific as to what the gcp-auth does

briandealwis:
The gcp-auth docs should describe the actual effects, and should also mention that the registry pull secrets are configured too.


sharifelgamal (approved):
Thanks for adding this documentation!

PR #12495, opened 2021-09-16
Replace BusyBox image with gcr.io image in functional tests

### LINKED ISSUE #12494: Replace busybox image in tests with gcr.io image
Sometimes we hit rate limiting against Docker Hub, we can greatly reduce our hit count be replacing that image we use for image load tests with a gcr.io image. The actual image we use doesn't actually matter because we're just loading the image into minikube, but obviously the smaller the better.

spowelljr:
Closes #12494

**Problem:**
We were getting rate limited against Docker Hub.

**Solution:**
Pull an image from gcr.io instead of using BusyBox from Docker Hub.

### DIFF
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -221,6 +221,7 @@ func validateNodeLabels(ctx context.Context, t *testing.T, profile string) {
 	}
 }
 
+// validateImageCommands runs tests on all the `minikube image` commands, ex. `minikube image load`, `minikube image list`, etc.
 func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
 	if NoneDriver() {
 		t.Skip("image commands are not available on the none driver")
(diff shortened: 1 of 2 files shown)


medyagh (approved):

test/integration/functional_test.go
```diff
@@ -227,72 +227,82 @@ func validateNodeLabels(ctx context.Context, t *testing.T, profile string) {
...
 
 // validateLoadImageFromFile makes sure that `minikube image load` works from a local file
 func validateLoadImageFromFile(ctx context.Context, t *testing.T, profile string) {
-	if NoneDriver() {
-		t.Skip("load image not available on none driver")
-	}
-	if GithubActionRunner() && runtime.GOOS == "darwin" {
-		t.Skip("skipping on github actions and darwin, as this test requires a running docker daemon")
-	}
+	checkSkip(t)
```
medyagh, line 295: better name here, checkSkip can be any skip

medyagh (changes requested):

PR #12497, opened 2021-09-16
fix refreshing gcp-auth pull secret

sharifelgamal:
This should fix the issues raised in https://github.com/kubernetes/minikube/issues/12481

### DIFF
--- a/deploy/addons/gcp-auth/gcp-auth-webhook.yaml.tmpl
+++ b/deploy/addons/gcp-auth/gcp-auth-webhook.yaml.tmpl
@@ -61,6 +61,7 @@ metadata:
   name: gcp-auth-certs-create
   namespace: gcp-auth
 spec:
+  ttlSecondsAfterFinished: 120
   template:
     metadata:
       name: gcp-auth-certs-create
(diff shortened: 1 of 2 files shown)


deploy/addons/gcp-auth/gcp-auth-webhook.yaml.tmpl
```diff
@@ -61,6 +61,7 @@ metadata:
   name: gcp-auth-certs-create
   namespace: gcp-auth
 spec:
+  ttlSecondsAfterFinished: 30
```
medyagh, line 64: long term it would be better to have the replace the job instead of hoping it gets deleted on time before refresh but I understand this is a less invasive fix

medyagh (approved):

deploy/addons/gcp-auth/gcp-auth-webhook.yaml.tmpl
sharifelgamal: this is probably a good idea regardless of a longer term fix, but yeah, there should be a more stable solution to avoid colliding jobs

PR #12503, opened 2021-09-17
Add github-release install script

spowelljr:
Add script to install latest version of github-release and run it on release.

### DIFF
--- a/hack/jenkins/installers/check_install_github_release.sh
+++ b/hack/jenkins/installers/check_install_github_release.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# Copyright 2021 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


hack/jenkins/installers/check_install_github_release.sh
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eux -o pipefail
+
+echo "Installing latest version of github-release"
+go get github.com/github-release/github-release
```
sharifelgamal: we should use `go install` instead of `go get` here

sharifelgamal (approved):

hack/jenkins/installers/check_install_github_release.sh
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eux -o pipefail
+
+echo "Installing latest version of github-release"
+go install github.com/github-release/github-release@latest
```
medyagh, line 20: will this crouppot the go.mod if ran inside minikube folder? (I believe this will add the grelease to our go.mod ? )

maybe we should run it in a /tmp folder

spowelljr: Will update

spowelljr: I tried and it works fine, it's because it's `go install` vs `go get`

PR #12533, opened 2021-09-21
change default crio cgroup to systemd

spowelljr:


medyagh (commented):
please lets paste the two config files one from minikube 1.23.0 and one from 1.23.1  and explain why cri-o needs systemd ...what is the default for containerd?

PR #12587, opened 2021-09-27
Fix Ipv4 cidr in cilium

### LINKED ISSUE #12077: Cilium CNI does not use Minikube Default Pod CIDR
When using Cilium CNI, pods are not assigned to the [Minikube Default Pod CIDR 10.244.0.0/16](https://github.com/kubernetes/minikube/blob/v1.21.0/pkg/minikube/cni/cni.go#L40) but instead assigned to [Cilium Default Pod CIDR 10.0.0.0/8](https://github.com/kubernetes/minikube/blob/v1.21.0/pkg/minikube/cni/cilium.go#L169) .

Consequently, there is also a mismatch between the [Pod CIDR defined on Kubeadm](https://github.com/kubernetes/minikube/blob/v1.21.0/pkg/minikube/bootstrapper/bsutil/kubeadm.go#L82) and the actual used [Cilium Pod CIDR ](https://github.com/kubernetes/minikube/blob/v1.21.0/pkg/minikube/cni/cilium.go#L169) since the [`CIDR()` function always returns the Minikube Default Pod CIDR](https://github.com/kubernetes/minikube/blob/v1.21.0/pkg/minikube/cni/cilium.go#L693-L695).

1. `minikube start --driver=virtualbox --cni=cilium` (Note: MacOS with VirtualBox)
2. `kubectl run example --image=busybox --restart=Never -- sleep 1d`
3. `kubectl get pod -o wide`
     Example output: 
     ```
     NAME      READY   STATUS    RESTARTS   AGE   IP          NODE       NOMINATED NODE   READINESS GATES
     example   1/1     Running   0          31s   10.88.0.3   minikube   <none>           <none>
     ```

It seems that Cilium CNI is not using the Kubeadm configuration.

**First quick-fix:**

1. Extract [Cilium manifest](https://github.com/kubernetes/minikube/blob/v1.21.0/pkg/minikube/cni/cilium.go#L27) to a local file.
2. Change `cluster-pool-ipv4-cidr: "10.0.0.0/8"` to `cluster-pool-ipv4-cidr: "10.244.0.0/16"` 
3. `minikube start --driver=virtualbox --cni=<path-to-modified-cilium-manifest>`
4. `kubectl run example --image=busybox --restart=Never -- sleep 1d`
5. `kubectl get pod -o wide`
     Example output: 
     ```
     NAME      READY   STATUS    RESTARTS   AGE   IP             NODE       NOMINATED NODE   READINESS GATES
     example   1/1     Running   0          8s    10.244.0.230   minikube   <none>           <none>
     ```

duttaANI:
Description
fixes #12077

issue link : [issue](https://github.com/kubernetes/minikube/issues/12077)


pkg/minikube/cni/cilium.go
```diff
@@ -175,7 +175,7 @@ data:
   hubble-tls-key-file: /var/lib/cilium/tls/hubble/server.key
   hubble-tls-client-ca-files: /var/lib/cilium/tls/hubble/client-ca.crt
   ipam: "cluster-pool"
-  cluster-pool-ipv4-cidr: "10.0.0.0/8"
+  cluster-pool-ipv4-cidr: "10.244.0.0/16"
```
medyagh, line 178: thank you for fixing this
we should make "tempaltized" for example
```
"{{ .PodSubnet }}"
```
so if user passes extra args we should use that simmilar to here
https://github.com/kubernetes/minikube/blob/db4f0dc414087c1b646b95699280bbab9b61ed0c/pkg/minikube/bootstrapper/bsutil/ktmpl/v1beta2.go#L74

medyagh (changes requested):

pkg/minikube/cni/cilium.go
duttaANI: Added the template for extra args .

sharifelgamal:
Thank you for your contribution!

PR #12588, opened 2021-09-27
clarify that Windows install commands are powershell-based

carehart:
Someone reading this line could presume it should work at the regular Windows command line (and the curl would), but of course the 2nd and 3rd lines will not.

And if someone who may approve this might want to also revise the line after step 2 here (or move to here its point about running as admin), that would of course make sense. I just didn't want to presume to do that in this simpler commit.

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -425,7 +425,7 @@ choco install minikube
 {{% quiz_instruction id="/Windows/x86-64/Stable/.exe download" %}}
 1. Download the [latest release](https://storage.googleapis.com/minikube/releases/latest/minikube-installer.exe).  
 <br>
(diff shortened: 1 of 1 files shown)


carehart:
I believe I did just successfully sign the CLA. First-time contributor to this project.

spowelljr (changes requested):
Hi @carehart, thanks for creating this PR! This looks good, the only thing is line 450 has the same string as well, could you change it there as well please, thanks!

medyagh (changes requested):

medyagh:
closing in favor of https://github.com/kubernetes/minikube/pull/12397

spowelljr:
@carehart #12397 is merged and is now live on the site

PR #12590, opened 2021-09-27
remove all windows tests from github actions

sharifelgamal:


.github/workflows/functional_verified.yml
```diff
@@ -0,0 +1,218 @@
...
+env:
+  GOPROXY: https://proxy.golang.org
+  GO_VERSION: '1.17.1'
+
+jobs:
+  # Runs before all other jobs
+  # builds the minikube binaries
+  build_minikube:
+    if: contains(github.event.pull_request.labels.*.name, 'ok-to-test')
+    runs-on: ubuntu-18.04
```
spowelljr, line 31: Not sure if we need `18.04` specifically, but if not we should use `ubuntu-latest` which will auto update to LTS without needing to update manually.

sharifelgamal: i'm happy to try that, I was just copying what already existed

spowelljr (approved):

PR #12592, opened 2021-09-27
GitHubActions: Update `ubuntu-18.04` to `ubuntu-20.04`

spowelljr:
Updating GitHub Actions from using `ubuntu-18.04` to `ubuntu-20.04`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ env:
   GO_VERSION: '1.17.1'
 jobs:
   build_minikube:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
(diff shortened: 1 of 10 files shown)


.github/workflows/build.yml
```diff
@@ -15,7 +15,7 @@ env:
   GO_VERSION: '1.17.1'
 jobs:
   build_minikube:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
medyagh, line 18: I am generally in favor of using a pinned version of everything and once we upgrade to something it is a conscious desciions, so we Decide to test minikube on new ubuntu, and ubuntu is not something that gets updated often.


so I suggest instead of update all of them to a pinned version

spowelljr: Do you want to update to 20.04 or stay at 18.04?

medyagh: 20.04  everywhere seems right

spowelljr: We have a functional test called `functional_baremetal_ubuntu18_04` do you want me to leave that as 18.04 or update that to 20.04 as well? I'm not sure if that test is for testing backwards compatibility or something like that.

medyagh (approved):

.github/workflows/build.yml
medyagh: yeah we can keep that one as 18.04

PR #12615, opened 2021-09-28
Move from deprecated ioutil to os and io packages

kallydev:
Since Minikube is currently using Go 1.17, so it is safe to remove deprecated `ioutil` package, ref https://golang.org/doc/go1.16#ioutil.


spowelljr (changes requested):
Looks good, just a small comment to address

spowelljr:
Hi @kallydev, your code is perfectly fine, but we have an issue with our CI with your changes. In one of our jobs we're using an image that is still using Go 1.15, and since a lot of the functions you added were released with Go 1.16 the job is failing. You can track the issue here (https://github.com/kubernetes/minikube/issues/12664), but we'll have to resolve this before merging your change.

spowelljr:
Hi @kallydev, I believe I resolved the CI issue, can you please rebase master on your branch and force push your branch back up after, thanks!

PR #12623, opened 2021-09-29
Add CPU time to time-to-k8s charts

### LINKED ISSUE #12238: Run CPU benchmark daily via GitHub Actions
We have the ability to run CPU benchmarks via `make cpu-benchmark-idle` & `make cpu-benchmark-autopause` which creates a chart such as https://minikube.sigs.k8s.io/docs/benchmarks/cpuusage/.

Two problems with this:
1. This command has to be run manually, and then manually commit and create a PR
2. This command only shows the latest run, it would be good to see the CPU usage over time

So we should create a GCP bucket to store the run results, and then make a script that takes in past results and the results from the day and put them all together to make a chart like this https://minikube.sigs.k8s.io/docs/benchmarks/timetok8s/daily_benchmark/ and then push this chart and data to the bucket.

And then run this script should automatically run daily via a GitHub Action so this can be as handsfree as possible.

yayaha:
fixes #12238.

Unfortunately, I can't pull or push to the GCS buckets, so can someone with the right permission help me test this? Thanks!

### DIFF
--- a/hack/benchmark/time-to-k8s/public-chart/generate-chart.go
+++ b/hack/benchmark/time-to-k8s/public-chart/generate-chart.go
@@ -114,6 +114,10 @@ func readInLatestBenchmark(latestBenchmarkPath string) benchmark {
 	var total float64
 	for _, step := range steps {
 		*step /= float64(count)
+		// Don't add CPU time to the total time.
+		if step == &cpu {
+			continue
+		}
 		total += *step
 	}
 
--- a/hack/benchmark/time-to-k8s/public-chart/generate-chart.go
+++ b/hack/benchmark/time-to-k8s/public-chart/generate-chart.go
@@ -43,6 +43,7 @@ type benchmark struct {
 	App    float64   `json:"app"`
 	DNSAns float64   `json:"dnsAns"`
 	Total  float64   `json:"total"`
+	CPU    float64   `json:"cpu"`
 }
 
 // benchmarks contains a list of benchmarks, used for storing benchmark results to JSON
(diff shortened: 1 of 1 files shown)


medyagh:
thank you @yayaha , you could actually test it locally by running ./hack/benchmark/time-to-k8s/public-chart/public-chart.sh docker
and then u can change that Script locally to not upload the png file to GCS

yayaha:
Tested locally w/o previous run results:

chart

medyagh:
@spowelljr would u please review this, will this not break our current chart?

yayaha:
It'll be nice if someone with access to the runs.json files on GCS can run this locally to test w/ previous run results, w/o uploading the latest result back to GCS.

spowelljr:
I'll run it

hack/benchmark/time-to-k8s/public-chart/generate-chart.go
```diff
@@ -70,8 +71,8 @@ func readInLatestBenchmark(latestBenchmarkPath string) benchmark {
 		log.Fatal(err)
 	}
 
-	var cmd, api, k8s, dnsSvc, app, dnsAns float64
-	steps := []*float64{&cmd, &api, &k8s, &dnsSvc, &app, &dnsAns}
+	var cmd, api, k8s, dnsSvc, app, dnsAns, cpu float64
+	steps := []*float64{&cmd, &api, &k8s, &dnsSvc, &app, &dnsAns, &cpu}
```
spowelljr, line 75: The issue with adding `cpu` here is that on line 117 is adds all the steps to the total, so `cpu` is going to be added to the total.

If you remove `cpu` from the steps slice, the only thing you have to do is add an assignment statement after the for loop on line 111 to append the cpu value from `values` to `cpu` instead.

spowelljr (changes requested):

spowelljr:
chart
Here's the chart I generated, it looks fine to me, once the CPU issue I commented above is addressed it should be fine. The reason Command Exec is higher is because it running on my Mac vs the more powerful CI machine.

hack/benchmark/time-to-k8s/public-chart/generate-chart.go
yayaha: Good catch! Thanks! Fixed.

spowelljr (approved):

PR #12657, opened 2021-10-05
Update _index.md

singsong2016:
1.i just think minikube start --driver=docker (default driver=kvm2)  is better

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -487,7 +487,7 @@ choco install minikube
 From a terminal with administrator access (but not logged in as root), run:
 
 ```shell
-minikube start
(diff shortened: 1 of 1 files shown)


spowelljr:
I don't understand what this change is doing:
1. `default driver kvm2` Docker is the default driver for minikube if it is installed and running, so this statement is incorrect.
2. `if docker should` I'm not sure what this means.
3. People want to be able to copy and paste commends, so having a comment in the middle of the command is not a good practice.
4. Why are we telling people to use the Docker driver specifically? minikube is designed to check what drivers are present and use the most stable one, the `--driver` flag is for if people want to override minikube's default choice.

This change doesn't make sense so I'm going to close this PR.

PR #12702, opened 2021-10-12
update ingress controller to v1/1.0.4 and v1beta1/0.49.3

prezha:
fix: #12710 

update ingress controller to current v1/1.0.4 (for k8s >= v1.19) and v1beta1/0.49.3 (for k8s < v1.19)

should resolve the issue reported in the slack channel: https://kubernetes.slack.com/archives/C1F5CT6Q1/p1634063547313500

ref: https://kubernetes.slack.com/archives/C1F5CT6Q1/p1634069399318200

### DIFF
--- a/deploy/addons/ingress/ingress-deploy.yaml.tmpl
+++ b/deploy/addons/ingress/ingress-deploy.yaml.tmpl
@@ -299,6 +299,11 @@ metadata:
   namespace: ingress-nginx
 spec:
   type: NodePort
+  {{- if eq .IngressAPIVersion "v1"}}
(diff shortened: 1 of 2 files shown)


prezha:
> We are just releasing v1.0.4 if you want to wait :D

@rikatz sounds great, thanks! yes, we can include that as well

i guess it should also cover the:
```
Known Issues

    Ingress controller now (starting from v1.0.0) mandates cluster scoped access to IngressClass. This leads to problems when updating old Ingress controller to newest version, as described here. We plan to fix it in v1.0.4 (promise!!), see this.
```
;)

PR #9577, opened 2020-10-28
Turn on ingress addon on linux baremetal

sharifelgamal:
Fixes #9322

PR #12735, opened 2021-10-19
Remove useless check of baremetal driver for ingress addon

zigarn:
Remove wrongly re-introduced (ee74d42eba9efb74d4b78de3f03503fdbeb94d6f, wrongly nested, so useless) check initially removed in #9577


sharifelgamal:
So this is a little more complex than it seems. The check itself is definitely in the wrong spot (a driver can't both be container-based and baremetal, clearly), but ingress is currently not working on baremetal at all, thus the reintroduction of the check,

PR #12800, opened 2021-10-27
WIP: PoC of manually downloading and installing deb packages

spowelljr:


spowelljr:
Will re-open if change is required

PR #12804, opened 2021-10-27
New flag "--binary-mirror" to override mirror URL downloading (kubectl, kubelet, & kubeadm)

### LINKED ISSUE #11935: Ability to set a mirror for downloading kubectl, kubelet, & kubeadm
This is partly related to #11934 in that I'm trying to get minikube to work in an offline environment, or at least without direct internet connectivity. I have Artifactory configured as a Docker registry so assuming that issue is fixed, I don't need direct internet access to download the container images.

However minikube still needs internet access to download the three `kube*` binaries. If I configure proxy access then there's nothing really to stop the container images being pulled through the proxy as well, which I'd like to avoid as it circumvents our internal policies and checks. It would be handy if I could supply an alternative URL that could be used to download the three `kube*` binaries. I can then set up an internal mirror of the https://storage.googleapis.com/kubernetes-release/release/* content.

1. `minikube start --download-only --image-repository=docker.artifactory.example.com`
2. Containers are (mostly, see linked issue) downloaded, but it times out downloading `kubelet`, etc.

There is obviously the `--image-repository` option already, as well as the `--iso-url` option that could be used to point to an alternative location for the minikube ISO (I don't need/use that particular option), so it seems odd that there's no way to specify an alternative location for the remaining `kube*` binaries.

presztak:
This PR adds flag `kube-binaries-mirror` to `minikube start` command.

Usage:
`minikube start --kube-binaries-mirror http://localhost:8000 --download-only`

Output:
```
😄  minikube v1.24.0 on Ubuntu 20.04 (vbox/amd64)
🆕  Kubernetes 1.22.4 is now available. If you would like to upgrade, specify: --kubernetes-version=v1.22.4
🚜  Pulling base image ...
    > kubectl: 44.73 MiB / 44.73 MiB [--------------------] 100.00% ? p/s 100ms
✅  Download complete!
```
Fixes  #11935

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -132,6 +132,7 @@ const (
 	listenAddress           = "listen-address"
 	extraDisks              = "extra-disks"
 	certExpiration          = "cert-expiration"
+	binaryMirror            = "binary-mirror"
 )
 
 var (
@@ -189,6 +190,7 @@ func initMinikubeFlags() {
 	startCmd.Flags().StringP(trace, "", "", "Send trace events. Options include: [gcp]")
 	startCmd.Flags().Int(extraDisks, 0, "Number of extra disks created and attached to the minikube VM (currently only implemented for hyperkit and kvm2 drivers)")
 	startCmd.Flags().Duration(certExpiration, constants.DefaultCertExpiration, "Duration until minikube certificate expiration, defaults to three years (26280h).")
+	startCmd.Flags().String(binaryMirror, "", "Location to fetch kubectl, kubelet, & kubeadm binaries from.")
 }
 
 // initKubernetesFlags inits the commandline flags for Kubernetes related options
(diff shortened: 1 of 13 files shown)


medyagh:
@presztak would you plz add in PR description the example using it with output ?

cmd/minikube/cmd/config/config.go
```diff
@@ -163,6 +163,10 @@ var settings = []Setting{
 		name: "native-ssh",
 		set:  SetBool,
 	},
+	{
+		name: "kube-binaries-mirror",
```
medyagh, line 167: I suggest putting this in cluster config struct instead of global config
So we can have multiple clusters with different mirrors and we could test in integration tests

medyagh (changes requested):

cmd/minikube/cmd/config/config.go
presztak: Fixed.

presztak:
@medyagh description was changed.

cmd/minikube/cmd/start_flags.go
```diff
@@ -124,6 +124,7 @@ const (
 	listenAddress           = "listen-address"
 	extraDisks              = "extra-disks"
 	certExpiration          = "cert-expiration"
+	kubeBinariesMirror      = "kube-binaries-mirror"
```
medyagh, line 127: I think this would be easier for users to find and use if it is called "binary-mirror" 

what do u think ?

presztak: Sounds good, shorter name will be better

presztak:
@medyagh I added integration test as you requested on slack. I tested this only on amd64 architecture.

medyagh (approved):

Issue #12761, opened 2021-10-22
Don't throw `Docker is nearly out of disk space` errors if more than X GB of disk space available


HarHarLinks:
I have just cleared out some space so minikube would even start at "100%" used disk - which in this case is several terabytes large and still had several 100GB free.
It is now warning me that "Docker is nearly out of disk space" at 98% which equals 843GB currently available space which is quite frankly ridiculous.

F1ko:
going to remove the frozen lifecycle since there are new commits being pushed to the PR

unless there is another reason to freeze this issue? (@spowelljr )

/remove-lifecycle frozen

spowelljr:
Frozen prevents it from going stale

mayur-tolexo:
Has this issue been worked on? I will like to take it up.

F1ko:
Yes, it actually is done already, I am just waiting for a final review.

PR #12814, opened 2021-10-28
Add sanity check to disk space check

F1ko:
Fixes https://github.com/kubernetes/minikube/issues/12761


spowelljr (changes requested):
This looks good, could you update/add tests to test this new check please

pkg/minikube/machine/start.go
```diff
@@ -233,17 +234,36 @@ func postStartValidations(h *host.Host, drvName string) {
...
 		klog.Warningf("error getting percentage of /var that is free: %v", err)
 	}
-	if percentageFull >= 99 {
-		exit.Message(kind, `{{.n}} is out of disk space! (/var is at {{.p}}% of capacity)`, out.V{"n": name, "p": percentageFull})
+
+	availableGiB, err := DiskAvailable(r, "/var")
+	if err != nil {
+		klog.Warningf("error getting GiB of /var that is available: %v", err)
+	}
+	thresholdGiB := 20
```
klaases, line 251: Can this be a const?

pkg/minikube/machine/start.go
```diff
@@ -233,17 +234,36 @@ func postStartValidations(h *host.Host, drvName string) {
 		return
 	}
 
+	if viper.GetBool("force") {
+		return
+	}
+
 	// make sure /var isn't full,  as pod deployments will fail if it is
```
klaases, line 241: Trim extra space.

pkg/minikube/machine/start.go
```diff
@@ -233,17 +234,36 @@ func postStartValidations(h *host.Host, drvName string) {
...
 	percentageFull, err := DiskUsed(r, "/var")
 	if err != nil {
 		klog.Warningf("error getting percentage of /var that is free: %v", err)
```
klaases, line 244: Safe to continue, or return on error?

pkg/minikube/machine/start.go
```diff
@@ -233,17 +234,36 @@ func postStartValidations(h *host.Host, drvName string) {
...
 	}
-	if percentageFull >= 99 {
-		exit.Message(kind, `{{.n}} is out of disk space! (/var is at {{.p}}% of capacity)`, out.V{"n": name, "p": percentageFull})
+
+	availableGiB, err := DiskAvailable(r, "/var")
+	if err != nil {
+		klog.Warningf("error getting GiB of /var that is available: %v", err)
```
klaases, line 249: Safe to continue, or return on error?

pkg/minikube/machine/start.go
```diff
@@ -262,6 +282,21 @@ func DiskUsed(cr command.Runner, dir string) (int, error) {
...
+	if err != nil {
+		klog.Warningf("error running df -BG /var: %v\n%v", err, output.Output())
+		return 0, err
+	}
+	gib := strings.TrimSpace(output.Stdout.String())
+	gib = strings.Trim(gib, "G")
+	return strconv.Atoi(gib)
+}
+
 // postStart are functions shared between startHost and fixHost
```
klaases, line 300: Function comment should begin like `// postStartSetup ...`

F1ko: good catch!

general question: that comment for example has nothing to do with the issue this PR refers to; is it advised to fix minor stuff like this whenever possible or should there be a separate cleanup issue (even though it doesn't change any logic)?

pkg/minikube/machine/start.go
F1ko: good point, i just left it as a warning since it was like this already before my changes, however, it does not really make sense to continue since we cannot ensure that everything works smoothly

therefore changed the behavior

pkg/minikube/machine/start.go
F1ko: sure, changed

pkg/minikube/machine/start.go
F1ko: yh, safe to continue since that flag skips the checks

pkg/minikube/machine/start.go
F1ko: which extra space are you referring to?
the empty line on 240?
I think it makes sense to add that empty line because of the comment that is about to follow, imo easier to catch up on it that way

F1ko:
Any idea why this PR acts crazy all of a sudden?

Comparing the changes shows me the "normal" expected result: 2 files changed with only the code that I touched.

Screenshot_20220217_214900

F1ko:
Will close and reopen the same PR again.

Issue #10708, opened 2021-03-03
remove kicbase image from local docker daemon on delete --purge


ilya-zuyev:
Currently, if 'docker' driver is used, `minikube delete --all --purge` command does not remove the pulled kicbase image 

```
➜  minikube git:(master) ✗ out/minikube start --driver=docker
* minikube v1.18.0 on Darwin 11.2
* Using the docker driver based on user configuration
* Starting control plane node minikube in cluster minikube
* Pulling base image ...
* Creating docker container (CPUs=2, Memory=3890MB) ...
* Preparing Kubernetes v1.20.2 on Docker 20.10.3 ...
  - Generating certificates and keys ...
  - Booting up control plane ...
  - Configuring RBAC rules ...
* Verifying Kubernetes components...
  - Using image gcr.io/k8s-minikube/storage-provisioner:v4
* Enabled addons: default-storageclass, storage-provisioner
* Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default

➜  minikube git:(master) ✗ docker images | grep kicbase
gcr.io/k8s-minikube/kicbase                           v0.0.18                           a776c544501a   5 days ago      1.08GB


➜  minikube git:(master) ✗ out/minikube delete --all --purge
* Deleting "minikube" in docker ...
* Removing /Users/izuyev/.minikube/machines/minikube ...
* Removed all traces of the "minikube" cluster.
* Successfully deleted all profiles
* Successfully purged minikube directory located at - [/Users/izuyev/.minikube]

➜  minikube git:(master) ✗ docker images | grep kicbase     
gcr.io/k8s-minikube/kicbase                           v0.0.18                           a776c544501a   5 days ago      1.08GB
```

This can be unexpected for user. For example, if he/she is using the locally built kicbase image and wants to reset it to the official one. `--purge` could be interpreted as `remove everything related`. 

May be we should delete the image or, if not, print an explicit warning/info message that we keep it when running `minikube delete`

medyagh:
I think a lot of people use --purge just to purge configs but they dont really wanna delete the image
I think the right thing to do is, to add a "Info" and letting them know we did NOT delete the kicbase
and here is the command to run to delete the kicbase .

but I can be convinced other way

prezha:
@sharifelgamal agreed with @ilya-zuyev's 2nd option and also with what @medyagh proposed - ie, warn the user and provide a cleanup command to use if they wish so

tstromberg:
[...]

PR #12835, opened 2021-11-01
Print a message about preserving kicbase image when `minkube delete --purge --all` used

presztak:
This PR introduce printing additional info about preserving kicbase image and generates
image deletion command for user when running `minikube delete`.
Message is printed only when `--purge` option is used.

E.g.
```
$ minikube delete --all --purge
🔥  Removing /home/piotr/.minikube/machines/minikube ...
💀  Successfully purged minikube directory located at - [/home/piotr/.minikube]
📌  Kicbase images have not been deleted. To delete images run:
    ▪ docker rmi gcr.io/k8s-minikube/kicbase:v0.0.28
```

Fixes: #10708

### DIFF
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -23,6 +23,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/docker/machine/libmachine/mcnerror"
@@ -34,6 +35,7 @@ import (
 	"github.com/spf13/viper"
 	"k8s.io/klog/v2"
 	cmdcfg "k8s.io/minikube/cmd/minikube/cmd/config"
+	"k8s.io/minikube/pkg/drivers/kic"
 	"k8s.io/minikube/pkg/drivers/kic/oci"
 	"k8s.io/minikube/pkg/minikube/cluster"
 	"k8s.io/minikube/pkg/minikube/config"
@@ -145,6 +147,60 @@ func deleteContainersAndVolumes(ctx context.Context, ociBin string) {
 	}
 }
 
+// kicbaseImages returns kicbase images
+func kicbaseImages(ctx context.Context, ociBin string) ([]string, error) {
+	if _, err := exec.LookPath(ociBin); err != nil {
+		return nil, nil
+	}
+
+	// create list of possible kicbase images
+	kicImages := []string{kic.BaseImage}
+	kicImages = append(kicImages, kic.FallbackImages...)
+
+	kicImagesRepo := []string{}
+	for _, img := range kicImages {
+		kicImagesRepo = append(kicImagesRepo, strings.Split(img, ":")[0])
+	}
+
+	allImages, err := oci.ListImagesRepository(ctx, ociBin)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []string
+	for _, img := range allImages {
+		for _, kicImg := range kicImagesRepo {
+			if kicImg == strings.Split(img, ":")[0] {
+				result = append(result, img)
+				break
+			}
+		}
+	}
+	return result, nil
+}
+
+// printDeleteImagesCommand prints command which remove images
+func printDeleteImagesCommand(ociBin string, imageNames []string) {
+	if _, err := exec.LookPath(ociBin); err != nil {
+		return
+	}
+
+	if len(imageNames) > 0 {
+		out.Styled(style.Command, `{{.ociBin}} rmi {{.images}}`, out.V{"ociBin": ociBin, "images": strings.Join(imageNames, " ")})
+	}
+}
+
+// printDeleteImageInfo prints info about removing kicbase images
+func printDeleteImageInfo(dockerImageNames, podmanImageNames []string) {
+	if len(dockerImageNames) == 0 && len(podmanImageNames) == 0 {
+		return
+	}
+
+	out.Styled(style.Notice, `Kicbase images have not been deleted. To delete images run:`)
+	printDeleteImagesCommand(oci.Docker, dockerImageNames)
+	printDeleteImagesCommand(oci.Podman, podmanImageNames)
+}
+
 // runDelete handles the executes the flow of "minikube delete"
 func runDelete(cmd *cobra.Command, args []string) {
 	if len(args) > 0 {
@@ -213,6 +269,16 @@ func runDelete(cmd *cobra.Command, args []string) {
 	// If the purge flag is set, go ahead and delete the .minikube directory.
 	if purge {
 		purgeMinikubeDirectory()
+
+		dockerImageNames, err := kicbaseImages(delCtx, oci.Docker)
+		if err != nil {
+			klog.Warningf("error fetching docker images: %v", err)
+		}
+		podmanImageNames, err := kicbaseImages(delCtx, oci.Podman)
+		if err != nil {
+			klog.Warningf("error fetching podman images: %v", err)
+		}
+		printDeleteImageInfo(dockerImageNames, podmanImageNames)
 	}
 }
 
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -554,6 +554,30 @@ func ListContainersByLabel(ctx context.Context, ociBin string, label string, war
 	return names, err
 }
 
+// ListImagesRepository returns all the images names
+func ListImagesRepository(ctx context.Context, ociBin string) ([]string, error) {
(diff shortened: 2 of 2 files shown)


cmd/minikube/cmd/delete.go
```diff
@@ -171,6 +198,16 @@ func runDelete(cmd *cobra.Command, args []string) {
 	delCtx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
 	defer cancel()
 
+	// get image names before containers deletion
+	dockerImageNames, err := kicbaseImages(delCtx, oci.Docker)
+	if err != nil {
+		klog.Warningf("error fetching docker images: %v", err)
+	}
+	podmanImageNames, _ := kicbaseImages(delCtx, oci.Podman)
```
spowelljr, line 206: Skipping the error check here

cmd/minikube/cmd/delete.go
```diff
@@ -145,6 +146,32 @@ func deleteContainersAndVolumes(ctx context.Context, ociBin string) {
...
+// printDeleteImagesCommand prints command which remove images
+func printDeleteImagesCommand(ociBin string, imageNames []string) {
+	if _, err := exec.LookPath(ociBin); err != nil {
+		return
+	}
+
+	if len(imageNames) > 0 {
+		out.Styled(style.Command, `{{.ociBin}} rmi {{.images}}`, out.V{"ociBin": ociBin, "images": strings.Join(imageNames, "")})
+	} else {
+		out.Styled(style.Command, `{{.ociBin}} rmi <KICBASE IMAGE>`, out.V{"ociBin": ociBin})
```
spowelljr, line 171: I'm not sure if having this line makes the most sense. Assuming the user is only using Docker or Podman, the output is going to look something like:
```
docker rmi kicbase:latest
podman rmi <KICBASE IMAGE>
```
We could omit the line, but would look odd if there was no KIC image for either and the output was empty. Maybe it would make sense to replace with something like `No docker kicbase images to remove`?

medyagh (changes requested):
the PR title doesnt seem to be doing what the code is doing, is this adding Deleting KicBase on minikube delete?

cmd/minikube/cmd/delete.go
presztak: @spowelljr thanks for the review. Fixed.

cmd/minikube/cmd/delete.go
presztak: I improved that and now message is only printed when images are found.
But this is a little bit tricky as well when user will run `minikube delete --purge -all` twice.
Message will be printed only for the first time. During the second run we won't find any images because containers no more exist. Is this sufficient solution?

presztak:
@medyagh thans for the review.

This change doesn't remove kicbase images but only print info about preserving them and generate image deletion command for user. I've changed title and description.

pkg/drivers/kic/oci/oci.go
```diff
@@ -554,6 +554,23 @@ func ListContainersByLabel(ctx context.Context, ociBin string, label string, war
...
+		return nil, err
+	}
+	s := bufio.NewScanner(bytes.NewReader(rr.Stdout.Bytes()))
+	var names []string
+	for s.Scan() {
+		n := strings.TrimSpace(s.Text())
+		if n != "" {
+			names = append(names, n)
+		}
+	}
```
spowelljr, line 570: After the scan loop should check for errors with `s.Err()` to ensure there possibly wasn't more images but an error occurred.

spowelljr (changes requested):

cmd/minikube/cmd/delete.go
```diff
@@ -145,6 +146,39 @@ func deleteContainersAndVolumes(ctx context.Context, ociBin string) {
...
+	}
+
+	if len(imageNames) > 0 {
+		out.Styled(style.Command, `{{.ociBin}} rmi {{.images}}`, out.V{"ociBin": ociBin, "images": strings.Join(imageNames, "")})
+	}
+}
+
+// printDeleteImageInfo prints info about removing kicbase images
+func printDeleteImageInfo(dockerImageNames, podmanImageNames []string) {
+	if len(dockerImageNames) > 0 || len(podmanImageNames) > 0 {
```
spowelljr, line 175: I think it makes sense to exit early here instead if both lists are empty, otherwise the `printDeleteImagesCommand` are being run unnecessarily if we already know both image lists are empty.
```
if empty
return
// rest
```

spowelljr (changes requested):

cmd/minikube/cmd/delete.go
spowelljr: Yeah, I looked at your changes and this is fine

cmd/minikube/cmd/delete.go
```diff
@@ -145,6 +146,41 @@ func deleteContainersAndVolumes(ctx context.Context, ociBin string) {
 	}
 }
 
+// kicbaseImage fetches images based on container label
+func kicbaseImages(ctx context.Context, ociBin string) ([]string, error) {
+	if _, err := exec.LookPath(ociBin); err != nil {
+		return nil, nil
+	}
+	delLabel := fmt.Sprintf("%s=%s", oci.CreatedByLabelKey, "true")
+	names, err := oci.ListImagesByContainerLabel(ctx, ociBin, delLabel)
```
spowelljr, line 155: One last thing, we might as well do an inline return here since that's all we're effectively doing anyway.

cmd/minikube/cmd/delete.go
```diff
@@ -145,6 +146,37 @@ func deleteContainersAndVolumes(ctx context.Context, ociBin string) {
...
+}
+
+// printDeleteImagesCommand prints command which remove images
+func printDeleteImagesCommand(ociBin string, imageNames []string) {
+	if _, err := exec.LookPath(ociBin); err != nil {
+		return
+	}
+
+	if len(imageNames) > 0 {
+		out.Styled(style.Command, `{{.ociBin}} rmi {{.images}}`, out.V{"ociBin": ociBin, "images": strings.Join(imageNames, "")})
```
spowelljr, line 165: Missed this, but I'm assuming we should have a space between the images?

presztak: That's correct.

cmd/minikube/cmd/delete.go
```diff
@@ -171,6 +203,16 @@ func runDelete(cmd *cobra.Command, args []string) {
 	delCtx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
 	defer cancel()
 
+	// get image names before containers deletion
+	dockerImageNames, err := kicbaseImages(delCtx, oci.Docker)
```
medyagh, line 207: we should add these only when the user does "--purge" so we dont make delete slower for normal users

medyagh (changes requested):
thank you for this PR, it is almost good, just need to move the list logic to only purge

pkg/drivers/kic/oci/oci.go
```diff
@@ -554,6 +554,26 @@ func ListContainersByLabel(ctx context.Context, ociBin string, label string, war
 	return names, err
 }
 
+// ListImagesByContainerLabel returns all the images names with a specified container label
+func ListImagesByContainerLabel(ctx context.Context, ociBin string, label string) ([]string, error) {
+	rr, err := runCmd(exec.CommandContext(ctx, ociBin, "ps", "-a", "--filter", fmt.Sprintf("label=%s", label), "--format", "{{.Image}}"))
```
medyagh, line 559: this doesnt seem to catch any kic image if u delete minikube
```
14:07:31 medya/workspace/minikube
master_srikrishnabh ⌅1
$ minikube delete --all
🔥  Removing /Users/medya/.minikube/machines/minikube ...
14:07:40 medya/workspace/minikube
master_srikrishnabh ⌅1
$ docker ps -a
```

but in reality

```
$ docker images | grep kicbase
gcr.io/k8s-minikube/kicbase          v0.0.28                                                 c3db8807ea16   7 weeks ago     981MB
gcr.io/k8s-minikube/kicbase          v0.0.27                                                 f94e08eb0335   2 months ago    980MB
```

medyagh (changes requested):

pkg/drivers/kic/oci/oci.go
medyagh: how about instead of docker ps, we do docker images and then we filter for all kicbase repos

both gcr and docker hub

presztak: Done.

cmd/minikube/cmd/delete.go
presztak: After changing the way how we retrieve images (with `images` cmd) I moved this part of code below to condition with `purge`

spowelljr (approved):
This looks good to me, once the integration tests pass and if they look good I'm happy. to merge this.

PR #12837, opened 2021-11-01
`minikube addons list` shows addons if cluster does not exist

spowelljr:
Closes https://github.com/kubernetes/minikube/issues/12736

**Before:**
```
$ minikube addons list
🤷  Profile "minikube" not found. Run "minikube profile list" to view all profiles.
👉  To start a cluster, run: "minikube start"
```

**Now:**
```
$ minikube addons list
|-----------------------------|-----------------------|
|         ADDON NAME          |      MAINTAINER       |
|-----------------------------|-----------------------|
| ambassador                  | unknown (third-party) |
| auto-pause                  | google                |
| csi-hostpath-driver         | kubernetes            |
| dashboard                   | kubernetes            |
| default-storageclass        | kubernetes            |
| efk                         | unknown (third-party) |
| freshpod                    | google                |
| gcp-auth                    | google                |
| gvisor                      | google                |
| helm-tiller                 | unknown (third-party) |
| ingress                     | unknown (third-party) |
| ingress-dns                 | unknown (third-party) |
| istio                       | unknown (third-party) |
| istio-provisioner           | unknown (third-party) |
| kubevirt                    | unknown (third-party) |
| logviewer                   | google                |
| metallb                     | unknown (third-party) |
| metrics-server              | kubernetes            |
| nvidia-driver-installer     | google                |
| nvidia-gpu-device-plugin    | unknown (third-party) |
| olm                         | unknown (third-party) |
| pod-security-policy         | unknown (third-party) |
| portainer                   | portainer.io          |
| registry                    | google                |
| registry-aliases            | unknown (third-party) |
| registry-creds              | unknown (third-party) |
| storage-provisioner         | kubernetes            |
| storage-provisioner-gluster | unknown (third-party) |
| volumesnapshots             | kubernetes            |
|-----------------------------|-----------------------|
```

PR #12839, opened 2021-11-02
`minikube addons list` tests

spowelljr:
Follow up to https://github.com/kubernetes/minikube/pull/12837

Adding tests and fix JSON output.

### DIFF
--- a/cmd/minikube/cmd/config/addons_list.go
+++ b/cmd/minikube/cmd/config/addons_list.go
@@ -137,10 +137,6 @@ var printAddonsList = func(cc *config.ClusterConfig) {
 }
 
 var printAddonsJSON = func(cc *config.ClusterConfig) {
-	if cc == nil {
-		out.String("{}")
-		return
-	}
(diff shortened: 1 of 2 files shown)


cmd/minikube/cmd/config/addons_list.go
```diff
@@ -137,6 +137,10 @@ var printAddonsList = func(cc *config.ClusterConfig) {
 }
 
 var printAddonsJSON = func(cc *config.ClusterConfig) {
+	if cc == nil {
```
sharifelgamal, line 140: so for JSON we're just returning nothing if there's no cluster?

sharifelgamal: could we not do the same thing we do for table mode? like output the list of addons without a status?

spowelljr: Yup, done

sharifelgamal (approved):

PR #12877, opened 2021-11-07
Added tests around loaded images.

### LINKED ISSUE #10451: Add more complex integration tests for `minikube image load`
In #10366  I added a simple test for the `minikube image load` command:
1. Pull busybox
2. Try to load it into minikube
3. Make sure busybox exists in minikube

It would be nice to beef up the tests a little bit and also verify the following cases:
1. Try to load an image that already is loaded in, and make sure nothing breaks
2. Load in an image with a new tag and make sure the new tag exists in the daemon

If anyone would be interested in picking this up, please feel free to assign this issue to yourself by commenting /assign.

vishjain:
fixes #10451

### DIFF
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -223,6 +223,27 @@ func validateNodeLabels(ctx context.Context, t *testing.T, profile string) {
 	}
 }
 
+// tagAndLoadImage is a helper function to pull, tag, load image (decreases cyclomatic complexity for linter).
+func tagAndLoadImage(ctx context.Context, t *testing.T, profile, taggedImage string) {
+	newPulledImage := fmt.Sprintf("%s:%s", addonResizer, "1.8.9")
+	rr, err := Run(t, exec.CommandContext(ctx, "docker", "pull", newPulledImage))
+	if err != nil {
+		t.Fatalf("failed to setup test (pull image): %v\n%s", err, rr.Output())
+	}
+
+	rr, err = Run(t, exec.CommandContext(ctx, "docker", "tag", newPulledImage, taggedImage))
+	if err != nil {
+		t.Fatalf("failed to setup test (tag image) : %v\n%s", err, rr.Output())
+	}
+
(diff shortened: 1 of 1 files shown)


test/integration/functional_test.go
```diff
@@ -316,6 +337,26 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
 		checkImageExists(ctx, t, profile, taggedImage)
 	})
 
+	// docs: Try to load image already loaded and make sure `minikube image load --daemon` works
+	t.Run("ImageReloadDaemon", func(t *testing.T) {
+		rr, err := Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "image", "load", "--daemon", taggedImage))
```
medyagh, line 342: did you mean this to be true ? lets explicitly set the true or false here

--daemon="TrueOrFalse"

medyagh (changes requested):

test/integration/functional_test.go
spowelljr: Putting a `bool` flag without a value defaults to `true`, for example `--force`, you don't have to put `--force=true`, it will default to true.

test/integration/functional_test.go
```diff
@@ -316,6 +337,26 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
 		checkImageExists(ctx, t, profile, taggedImage)
 	})
 
+	// docs: Try to load image already loaded and make sure `minikube image load --daemon` works
+	t.Run("ImageReloadDaemon", func(t *testing.T) {
```
spowelljr, line 341: The test before this does an identical daemon image load, so this test only needs to do a single image load for it to be a reload.

test/integration/functional_test.go
vishjain: Yeah meant to be true, wanted to follow the style of all the other tests written in this file.

medyagh (approved):

PR #12893, opened 2021-11-08
only copy binaries to minikube/latest on non-PR builds

sharifelgamal:

### DIFF
--- a/hack/jenkins/minikube_cross_build_and_upload.sh
+++ b/hack/jenkins/minikube_cross_build_and_upload.sh
@@ -83,7 +83,7 @@ gsutil -m rsync -dJR out "gs://${bucket}/${ghprbPullId}"
 
 readonly bucket_mirror="minikube/latest"
 readonly HEAD="master"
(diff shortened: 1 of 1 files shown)


hack/jenkins/minikube_cross_build_and_upload.sh
```diff
@@ -83,5 +83,7 @@ rm -rf out/buildroot
 # -R: recursive. strangely, this is not the default for sync.
 gsutil -m rsync -dJR out "gs://${bucket}/${ghprbPullId}"
 
-# Copy artifacts to known mirror location
-gsutil cp -R "gs://${bucket}/${ghprbPullId}" "gs://${bucket_mirror}"
+if [[ "${ghprbPullId}" -eq "master" ]] then
```
spowelljr, line 86: Does this need an `;` between the conditional then `then`? Only asking because the rest of them have it.

```suggestion
if [[ "${ghprbPullId}" -eq "master" ]]; then
```

sharifelgamal: good call, just a mistake on my part

PR #12921, opened 2021-11-10
kubevirt addon: use KubeVirt CR instead of deprecated ConfigMap

zcahana:
Modify the kubevirt addon (#8275) to configure KubeVirt via its operator's custom resource (`KubeVirt`), instead of via the deprecated (and ~soon~ sometime to be removed) `kubevirt-config` ConfigMap.

Also slight changes to the saved manifest file names (avoid confusing renames).

### DIFF
--- a/deploy/addons/kubevirt/pod.yaml.tmpl
+++ b/deploy/addons/kubevirt/pod.yaml.tmpl
@@ -10,9 +10,9 @@ data:
   uninstall.sh: |
     #!/bin/bash
 
-    kubectl delete -f /manifests/kubevirt.yaml
+    kubectl delete -f /manifests/kubevirt-cr.yaml
 
-    kubectl delete -f /manifests/kubevirt-base.yaml
+    kubectl delete -f /manifests/kubevirt-operator.yaml
 
   install.sh: |
     #!/bin/bash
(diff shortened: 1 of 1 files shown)


medyagh:
@zcahana do u mind sharing Before and After this PR with an example and output ?

zcahana:
Well, the usage and the output of the addon enablement command hasn't changed:
```sh
zvic@ZVICAHANA-D2LK:~/Workspace/minikube (kubevirt_configmap)$ ./out/minikube addons enable kubevirt
    ▪ Using image bitnami/kubectl:1.17
🌟  The 'kubevirt' addon is enabled
```

Semantically, both before and after, we have a functional kubevirt installation.

Internally, there are some differences (which also affect users who wish to tinker with KubeVirt configuration):
- Before, there's been both a KubeVirt CR as well as a `kubevirt-config` ConfigMap, whereas the ConfigMap takes precedence over anything specified in the `KubeVirt.spec.configuration` field.
- After, there's no `kubevirt-config` ConfigMap created at all, and the `KubeVirt.spec.configuration.developerConfiguration.useEmulation` field is set to `true`. All other configuration should go into `KubeVirt.spec.configuration` as well.

deploy/addons/kubevirt/pod.yaml.tmpl
```diff
@@ -10,29 +10,27 @@ data:
...
     EMULATION=$(egrep 'svm|vmx' /proc/cpuinfo)
     if [ -z "$EMULATION" ]; then
       echo "use emulation"
-      kubectl create configmap kubevirt-config -n kubevirt --from-literal debug.useEmulation=true
-    fi;
-
-    curl -sL "https://github.com/kubevirt/kubevirt/releases/download/${KUBEVIRT_VERSION}/kubevirt-cr.yaml" -o /manifests/kubevirt.yaml
-
-    kubectl create -f /manifests/kubevirt.yaml
+      kubectl patch kubevirt kubevirt -n kubevirt --type merge --patch '{"spec": {"configuration": {"developerConfiguration": {"useEmulation": true}}}}'
```
rmohr, line 32: This definitely works, but in general our kubevirt-cr.yaml does not set any non-default option and for declarative kubevirt installs with kubevirt you can just `apply` such a `cr`:

```yaml
apiVersion: kubevirt.io/v1
kind: KubeVirt
metadata:
  name: kubevirt
  namespace: kubevirt
spec: {}
```

Or in your case:

```yaml
apiVersion: kubevirt.io/v1
kind: KubeVirt
metadata:
  name: kubevirt
  namespace: kubevirt
spec:
  developerConfiguration:
    useEmulation: true
```

So in stead of 

```
curl -sL "https://github.com/kubevirt/kubevirt/releases/download/${KUBEVIRT_VERSION}/kubevirt-cr.yaml" -o /manifests/kubevirt-cr.yaml
```

just directly write the above content.

zcahana: Updated accordingly.

Note that I too prefer the `cat << EOF | kubectl apply -f -` syntax, but this doesn't work in this context since the above `pod.yaml.tmpl` template is rendered using the `html/template` pkg, which doesn't like the `<<`.

rmohr (approved):
Looks great, at least for me.

medyagh (approved):
Thank you

Issue #12802, opened 2021-10-27
Add `minikube mount` flags to `minikube start`


spowelljr:
You can tell minikube to mount on start using `minikube start --mount` but the many flags included with the `minikube mount` command are not available with `minikube start`, so users are required to use `minikube mount`.

All the `minikube mount` flags should be available on `minikube start` as well.

PR #12930, opened 2021-11-11
Add all mount flags to start command

spowelljr:
Closes https://github.com/kubernetes/minikube/issues/12802

Implemented all the mount flags from `minikube mount` into `minikube start`.

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -83,6 +83,15 @@ const (
 	imageRepository         = "image-repository"
 	imageMirrorCountry      = "image-mirror-country"
 	mountString             = "mount-string"
+	mount9PVersion          = "mount-9p-version"
+	mountGID                = "mount-gid"
+	mountIPFlag             = "mount-ip"
+	mountMode               = "mount-mode"
+	mountMSize              = "mount-msize"
+	mountOptions            = "mount-options"
+	mountPortFlag           = "mount-port"
+	mountTypeFlag           = "mount-type"
+	mountUID                = "mount-uid"
 	disableDriverMounts     = "disable-driver-mounts"
 	cacheImages             = "cache-images"
 	uuid                    = "uuid"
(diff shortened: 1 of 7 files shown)


henrik242 (approved):

spowelljr:
Waiting for https://github.com/kubernetes/minikube/pull/13018 to be merged to fix mode failure in tests

pkg/minikube/node/config.go
```diff
@@ -79,3 +77,34 @@ func configureMounts(wg *sync.WaitGroup, mount bool, mountString string) {
...
+		{constants.Mount9PVersionFlag, cc.Mount9PVersion},
+		{constants.MountGIDFlag, cc.MountGID},
+		{constants.MountIPFlag, cc.MountIP},
+		{constants.MountMSizeFlag, fmt.Sprintf("%d", cc.MountMSize)},
+		{constants.MountModeFlag, fmt.Sprintf("%d", cc.MountMode)},
+		{constants.MountPortFlag, fmt.Sprintf("%d", cc.MountPort)},
+		{constants.MountTypeFlag, cc.MountType},
+		{constants.MountUIDFlag, cc.MountUID},
+	}
+	for _, flag := range flags {
```
sharifelgamal, line 103: are there any weird side effects if only some of the mount options are set and not others? what happens if one of these is an empty string?

spowelljr: Good question, but the command runner seems to handle empty strings fine, `MountIP` is empty by default and the mounting works fine.

cmd/minikube/cmd/start_flags.go
```diff
@@ -83,6 +83,15 @@ const (
 	imageRepository         = "image-repository"
 	imageMirrorCountry      = "image-mirror-country"
 	mountString             = "mount-string"
+	mount9PVersion          = "mount-9p-version"
+	mountGID                = "mount-gid"
+	mountIPFlag             = "mount-ip"
```
klaases, line 88: Some have "Flag" at the end and some do not. Are all of these flags? Could we drop "Flag" for consistency between these?

spowelljr: `mountIP`, `mountPort` & `mountType` are already defined in `mount.go` so I had to differentiate them

pkg/minikube/node/config.go
```diff
@@ -79,3 +77,34 @@ func configureMounts(wg *sync.WaitGroup, mount bool, mountString string) {
...
+		args = append(args, fmt.Sprintf("--%s", flag.name), flag.value)
+	}
+	for _, option := range cc.MountOptions {
+		args = append(args, fmt.Sprintf("--%s", constants.MountOptionsFlag), option)
+	}
+	return args
```
klaases, line 109: Are there any cases where flags could possess inconsistent values?

Should we have logic, to say, for example, that mountSize value is out of range (maybe we want to do this in a separate PR).

test/integration/mount_start_test.go
```diff
@@ -84,11 +94,60 @@ func validateStartWithMount(ctx context.Context, t *testing.T, profile string) {
...
+	}
+
+	args = sshArgs
+	args = append(args, "stat", "--format", "'%a'", "/minikube-host")
+	rr, err = Run(t, exec.CommandContext(ctx, Target(), args...))
+	if err != nil {
+		t.Fatalf("failed to get directory mode: %v", err)
+	}
+
+	want := "777"
```
klaases, line 118: Prefer const.

klaases (commented):
Added a few comments, mainly curious about if validating flag values is needed.

pkg/minikube/node/config.go
spowelljr: Yeah, I can add that is a follow up PR, I created an issue

https://github.com/kubernetes/minikube/issues/13142

test/integration/mount_start_test.go
spowelljr: Done!

pkg/minikube/node/config.go
klaases: Makes sense!

cmd/minikube/cmd/start_flags.go
klaases: Alright, works for me.

klaases (approved):

sharifelgamal (approved):
Thanks for fixing this

PR #12941, opened 2021-11-13
fix:  do not let users run --no-kubernetes on the none driver

### LINKED ISSUE #12934: do not let users run `--no-kubernetes` on the none driver
The none driver is just running kubeadm directly on the user's machine, so no kubernetes makes no sense. We should just tell the user and error out.

kadern0:
Fixes #12934 

```
$ sudo ./out/minikube start --no-kubernetes --driver=none
[sudo] password for: 
😄  minikube v1.24.0 on Ubuntu 20.04

❌  Exiting due to MK_USAGE: Cannot use the option --no-kuberenes on the 'none driver
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1126,6 +1126,10 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 
 	validateCPUCount(drvName)
 
+	if drvName == driver.None && viper.GetBool(noKubernetes) {
+		exit.Message(reason.Usage, "Cannot use the option --no-kubernetes on the {{.name}} driver", out.V{"name": drvName})
+	}
+
 	if cmd.Flags().Changed(memory) {
 		validateChangedMemoryFlags(drvName)
 	}


cmd/minikube/cmd/start.go
```diff
@@ -1126,6 +1126,10 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 
 	validateCPUCount(drvName)
 
+	if drvName == "none" && viper.GetBool(noKubernetes) {
```
spowelljr, line 1129:
```suggestion
	if drvName == driver.None && viper.GetBool(noKubernetes) {
```

cmd/minikube/cmd/start.go
```diff
@@ -1126,6 +1126,10 @@ func validateFlags(cmd *cobra.Command, drvName string) {
...
+		exit.Message(reason.Usage, "Cannot use the option --no-kuberenes on the {{.name}} driver", out.V{"name": drvName})
```
spowelljr, line 1130:
```suggestion
		exit.Message(reason.Usage, "Cannot use the option --no-kubernetes on the {{.name}} driver", out.V{"name": drvName})
```

spowelljr (changes requested):

cmd/minikube/cmd/start.go
kadern0: Good catch!

spowelljr (approved):

Issue #12919, opened 2021-11-10
Port validation breaks options that used to work on Linux


mans0954:
Prior to minikube 1.24.0, it used to be possible to create the minikube container so that ports were exposed on `127.0.0.1:443` and `127.0.0.1:80` e.g.

```
# Install minikube 1.23.2
sudo dpkg -i minikube_1.23.2-0_amd64.deb

# Start minikube, exposing port `443` on `127.0.0.1:443` and port `80` `127.0.0.1:80`
minikube start  --ports=127.0.0.1:443:443 --ports=127.0.0.1:80:80 --profile minikube-1-23-2

# Inspect the container
docker inspect minikube-1-23-2 | jq .[0].NetworkSettings.Ports
{
  "22/tcp": [
    {
      "HostIp": "127.0.0.1",
      "HostPort": "49172"
    }
  ],
  "2376/tcp": [
    {
      "HostIp": "127.0.0.1",
      "HostPort": "49171"
    }
  ],
  "32443/tcp": [
    {
      "HostIp": "127.0.0.1",
      "HostPort": "49168"
    }
  ],
  "443/tcp": [
    {
      "HostIp": "127.0.0.1",
      "HostPort": "443"
    }
  ],
  "5000/tcp": [
    {
      "HostIp": "127.0.0.1",
      "HostPort": "49170"
    }
  ],
  "80/tcp": [
    {
      "HostIp": "127.0.0.1",
      "HostPort": "80"
    }
  ],
  "8443/tcp": [
    {
      "HostIp": "127.0.0.1",
      "HostPort": "49169"
    }
  ]
}

# Enable the ingress addon
minikube addons enable ingress --profile minikube-1-23-2

# Check that nginx ingress responds to HTTP requests on `127.0.0.1:80`
curl --head http://127.0.0.1
HTTP/1.1 404 Not Found
Date: Wed, 10 Nov 2021 13:47:07 GMT
Content-Type: text/html
Content-Length: 146
Connection: keep-alive

# Check that  nginx ingress responds to HTTPS requests on `127.0.0.1:443`
curl --head -k https://127.0.0.1
HTTP/2 404 
date: Wed, 10 Nov 2021 13:47:20 GMT
content-type: text/html
content-length: 146

# Check that ports are exposed only on 127.0.0.1
curl http://127.0.0.2
curl: (7) Failed to connect to 127.0.0.2 port 80: Connection refused

curl -k https://127.0.0.2
[...]

PR #12952, opened 2021-11-15
Only check privileged ports for WSL

spowelljr:
Related: https://github.com/kubernetes/minikube/issues/12919

**Problem:**
We are checking for privileged ports always, but we should only be checking for it on WSL.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1226,7 +1226,7 @@ func validatePorts(ports []string) error {
 			if p > 65535 || p < 1 {
 				return errors.Errorf("Sorry, one of the ports provided with --ports flag is outside range %s", ports)
 			}
-			if p < 1024 && i == 0 {
+			if detect.IsMicrosoftWSL() && p < 1024 && i == 0 {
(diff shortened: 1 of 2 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -1226,7 +1226,7 @@ func validatePorts(ports []string) error {
 			if p > 65535 || p < 1 {
 				return errors.Errorf("Sorry, one of the ports provided with --ports flag is outside range %s", ports)
 			}
-			if p < 1024 && i == 0 {
+			if detect.IsMicrosoftWSL() && p < 1024 && i == 0 {
```
medyagh, line 1229: this means after trhis PR u can use privileged  ports on Non WSL ? is this intentional?

spowelljr: Correct, the initial issue was to only add the privilege port validation to WSL but it got added to everything instead. Users are bringing up the issue as now validation is not allowing them to use ports that were previously working.

medyagh: are those users Docker on Linux ? or docker on windows too?

spowelljr: For sure Docker Linux, no reports from Docker Windows

PR #12961, opened 2021-11-16
fix string extraction for translations

sharifelgamal:
This solves two issues:
1. Our gendocs github action was broken because of an added Advice entry that uses fmt.Sprintf, which we should just support.
2. Function calls with multiple string arguments (such as BoxedWithConfig) were only getting their first string argument extracts (in BoxedWithConfig's case, its title).


pkg/minikube/extract/extract.go
```diff
@@ -542,7 +540,25 @@ func extractAdvice(f ast.Node, e *state) error {
...
+			advice, ok := kvp.Value.(*ast.BasicLit)
+			if !ok {
+				// Maybe the advice is a function call, like fmt.Sprintf
+				ad, ok := kvp.Value.(*ast.CallExpr)
+				if !ok {
+					// Ok, we tried. Abort.
+					return true
+				}
+				// Let's support fmt.Sprintf with a string argument only
+				adArg, ok := ad.Args[0].(*ast.BasicLit)
```
spowelljr, line 552: I don't know if it's ever possible that there are no `Args` and that `Args[0]` can cause an out of bounds panic, just mentioning it incase it is possible.

sharifelgamal: good call, i'll just iterate through the args instead

spowelljr (approved):

Issue #12115, opened 2021-08-04
Minikube tunnel hangs on Ubuntu 20 in Windows 10


johnnybigert:
**Steps to reproduce the issue:** 

1. `minikube start`
2. `kubectl apply -f manifests/storage-deployment.yaml`
3. `kubectl apply -f manifests/storage-service.yaml`
2. `minikube tunnel`

**Full output of `minikube logs` command:**
[out.txt](https://github.com/kubernetes/minikube/files/6929825/out.txt)

**Full output of failed command:**
Observed behavior: The `minikube tunnel` command provides no output and appears to be frozen. (However, using `htop` I can see that the tunnel command is using cpu and memory. Also, if I do `minikube stop` in another terminal, the tunnel command starts outputting error messages.)
Expected behavior: Output of service IP addresses etc. like the output in [this example](https://minikube.sigs.k8s.io/docs/handbook/accessing/).

Note that the Ubuntu 20.04 is running in Windows 10 (WSL2).

spowelljr:
Hi @johnnybigert, thanks for reporting your issue with minikube!

There's a large discussion about running minikube on WSL2 on this issue #7879.

There are multiple comments discussing the tunnel command, I'd recommend reading through them and seeing if any of them are applicable to you and maybe solve your issue, thanks!

johnnybigert:
@spowelljr Thanks for responding. I read #7879, nothing helpful there what I could see, just @ashishsecdev having the same issue as me.

spowelljr:
@johnnybigert Thanks for checking out that issue, I'll see if I have time to check it out today. If you could though, try running `minikube tunnel --alsologtostderr` that should give you a more verbose output and possibly see where the problem lies.

johnnybigert:
Nice @spowelljr, got some more info:

`$ minikube tunnel --alsologtostderr
I0810 16:20:47.832318   42964 out.go:286] Setting OutFile to fd 1 ...
I0810 16:20:47.833475   42964 out.go:338] isatty.IsTerminal(1) = true
I0810 16:20:47.833541   42964 out.go:299] Setting ErrFile to fd 2...
I0810 16:20:47.834013   42964 out.go:338] isatty.IsTerminal(2) = true
I0810 16:20:47.834427   42964 root.go:312] Updating PATH: /home/johnny/.minikube/bin
I0810 16:20:47.836683   42964 mustload.go:65] Loading cluster: minikube
[...]

PR #12976, opened 2021-11-17
Output tunnel has started message so users aren't confused if tunnel is running

spowelljr:
Related https://github.com/kubernetes/minikube/issues/12115

When starting the tunnel on a KIC driver, there's no indication to the user that the tunnel is running so users may thing the tunnel command is frozen.

**Before:**
```
$ minikube tunnel
```

**After:**
```
$ minikube tunnel
✅  Tunnel successfully started

📌  NOTE: This process must stay alive for the tunnel to be accessible ...
```

### DIFF
--- a/cmd/minikube/cmd/tunnel.go
+++ b/cmd/minikube/cmd/tunnel.go
@@ -109,7 +109,7 @@ var tunnelCmd = &cobra.Command{
 func outputTunnelStarted() {
 	out.Styled(style.Success, "Tunnel successfully started")
 	out.Ln("")
-	out.Styled(style.Notice, "NOTE: This process must stay alive for the tunnel to be accessible ...")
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/tunnel.go
```diff
@@ -103,6 +106,13 @@ var tunnelCmd = &cobra.Command{
 	},
 }
 
+func outputTunnelStarted() {
+	out.Styled(style.Success, "Tunnel successfully started")
+	out.Ln("")
+	out.Styled(style.Notice, "NOTE: This process must stay alive for the tunnel to be accessible ...")
```
medyagh, line 112: How about adding 

"Please do not close this terminal, as this process has to stay alive for ....(rest)"

spowelljr: Updated!

medyagh (approved):

PR #12985, opened 2021-11-18
site: tests docs change name

medyagh:
we might need to delete the existing page after this PR so we dont have duplicate page (if it changes the URL of the new page)


sharifelgamal (approved):
Changing the title won't change the path of the page, no need to delete anything. The gendocs github action will take care of the rest.

Issue #11932, opened 2021-07-08
Exiting due to K8S_KUBELET_NOT_RUNNING


rafacouto:
**Steps to reproduce the issue:** 

1. minikube start

**Full output of `minikube logs` command:**
<details>

* ==> Docker <==
* -- Logs begin at Thu 2021-07-08 09:23:44 UTC, end at Thu 2021-07-08 09:33:41 UTC. --
Jul 08 09:23:44 minikube systemd[1]: Starting Docker Application Container Engine...
Jul 08 09:23:44 minikube dockerd[148]: time="2021-07-08T09:23:44.998760636Z" level=info msg="Starting up"
Jul 08 09:23:44 minikube dockerd[148]: time="2021-07-08T09:23:44.999619832Z" level=info msg="parsed scheme: \"unix\"" module=grpc
Jul 08 09:23:44 minikube dockerd[148]: time="2021-07-08T09:23:44.999639457Z" level=info msg="scheme \"unix\" not registered, fallback to default scheme" module=grpc
Jul 08 09:23:44 minikube dockerd[148]: time="2021-07-08T09:23:44.999659502Z" level=info msg="ccResolverWrapper: sending update to cc: {[{unix:///run/containerd/containerd.sock  <nil> 0 <nil>}] <nil> <nil>}" module=grpc
Jul 08 09:23:44 minikube dockerd[148]: time="2021-07-08T09:23:44.999668581Z" level=info msg="ClientConn switching balancer to \"pick_first\"" module=grpc
Jul 08 09:23:45 minikube dockerd[148]: time="2021-07-08T09:23:45.000894168Z" level=info msg="parsed scheme: \"unix\"" module=grpc
Jul 08 09:23:45 minikube dockerd[148]: time="2021-07-08T09:23:45.000921896Z" level=info msg="scheme \"unix\" not registered, fallback to default scheme" module=grpc
Jul 08 09:23:45 minikube dockerd[148]: time="2021-07-08T09:23:45.000941940Z" level=info msg="ccResolverWrapper: sending update to cc: {[{unix:///run/containerd/containerd.sock  <nil> 0 <nil>}] <nil> <nil>}" module=grpc
Jul 08 09:23:45 minikube dockerd[148]: time="2021-07-08T09:23:45.000951439Z" level=info msg="ClientConn switching balancer to \"pick_first\"" module=grpc
[...]

PR #12990, opened 2021-11-19
Add advice if user using btrfs storage driver

spowelljr:
Related https://github.com/kubernetes/minikube/issues/7923
Fixes https://github.com/kubernetes/minikube/issues/11932

Add advice if user is using btrfs storage driver to pass the `--feature-gates="LocalStorageCapacityIsolation=false"` flag to their start command.


pkg/minikube/reason/known_issues.go
```diff
@@ -1053,6 +1053,15 @@ var guestIssues = []match{
 		},
 		Regexp: re(`'/var/lib/dpkg': No such file or directory`),
 	},
+	{
```
medyagh, line 1056: how about in addition to this ( as a fall back ) we set the feature gate automatically for the user?

medyagh (changes requested):

pkg/minikube/reason/known_issues.go
medyagh: we have example of it in cni.go

```
		// for containerd and docker: auto-set custom CNI via kubelet's 'cni-conf-dir' param, if not user-specified
		eo := fmt.Sprintf("kubelet.cni-conf-dir=%s", CustomConfDir)
		if !cc.KubernetesConfig.ExtraOptions.Exists(eo) {
			klog.Infof("auto-setting extra-config to %q", eo)
			if err := cc.KubernetesConfig.ExtraOptions.Set(eo); err != nil {
				return fmt.Errorf("failed auto-setting extra-config %q: %v", eo, err)
			}
```

spowelljr: Added!

pkg/minikube/driver/driver.go
```diff
@@ -144,6 +145,24 @@ func IsDockerDesktop(name string) bool {
 	return false
 }
 
+// IsDockerBTRFS checks if Docker is using btrfs storage driver
+func IsDockerBTRFS(name string) bool {
+	if !IsDocker(name) {
+		return false
+	}
+	cmd := exec.Command("docker", "info", "-f", "{{.Driver}}")
```
medyagh, line 153: docker info is an expensive call, for that reason we have a cached version of it.


lets change this to oci.CachedDaemonInfo("docker")
"oci.CachedDaemonInfo"


I suspect this is same for podman, since it uses same base image?

medyagh (changes requested):

pkg/minikube/driver/driver.go
spowelljr: Done!

spowelljr:
User tested it and confirmed it works
https://github.com/kubernetes/minikube/issues/11932#issuecomment-985717791

medyagh (approved):

PR #13013, opened 2021-11-23
mount: fix `Bad file descriptor`

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/12301

Removed the `if fid.opened` then `req.RespondError(Ebaduse)`, the comment for the statement is `/* we can't walk open files */`, but when the check is skipped we can walk the supposedly open files fine, so I feel confident we can skip this check.


medyagh (approved):

medyagh:
nice ! thank you for fixing this annoying bug

PR #13024, opened 2021-11-25
Fix IstioOperator CustomResourceDefinition for istio-provisioner addon

pedrothome1:
This PR fixes a bug that prevents enabling the istio-provisioner addon.
The following error was occurring:
`error: error validating "/etc/kubernetes/addons/istio-operator.yaml": error validating data: ValidationError(CustomResourceDefinition.spec): unknown field "subresources" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec; if you choose to ignore these errors, turn validation off with --validate=false`


medyagh:
@pedrothome1 do u mind sharing the output of minikube addon enable  before and after this PR ?

PR #13040, opened 2021-11-29
Add some clarification on mounts needed for none

### LINKED ISSUE #13038: Persistent Volumes Clarification
_I am using minikube to run my cluster and ran across a problem. I am not sure wheter it belongs to the docs or source code._

The documentation gives an example for `PVs` and how to define that [here](https://minikube.sigs.k8s.io/docs/handbook/persistent_volumes/#a-note-on-mounts-persistence-and-minikube-hosts). Using the persistant volume claims, you do not have to specify the `host path` and a generic one gets generated. **BUT**, why do you choose the `tmp` folder per default?! Please please put a warning in the docs that using `PVCs` and the default minikube provisioner does, in fact, only store your data for a limited time.

Alternatively change the default location to somewhere really persistant!

afbjorklund:
Normally these are provided by "minikube-automount",
but with the none driver you are on your own (as usual)

Closes #13038

### DIFF
--- a/site/content/en/docs/drivers/none.md
+++ b/site/content/en/docs/drivers/none.md
@@ -38,6 +38,16 @@ This document is written for system integrators who wish to run minikube within 
 
 * minikube and the Kubernetes services it starts may interfere with other running software on the system. For instance, minikube will start and stop container runtimes via systemd, such as docker, containerd, cri-o.
 
+### Persistent storage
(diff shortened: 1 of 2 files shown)


medyagh (approved):

medyagh:
thanks for making it more clear

PR #13054, opened 2021-11-30
bump k8s lib

medyagh:
to address https://deps.dev/advisory/GHSA/GHSA-74j8-88mm-7496?from=%2Fgo%2Fk8s.io%252Fminikube


go.mod
```diff
@@ -73,9 +72,9 @@ require (
 	github.com/xeipuuv/gojsonschema v0.0.0-20180618132009-1d523034197f
 	github.com/zchee/go-vmnet v0.0.0-20161021174912-97ebf9174097
 	go.opencensus.io v0.23.0
-	go.opentelemetry.io/otel v0.17.0
-	go.opentelemetry.io/otel/sdk v0.16.0
-	go.opentelemetry.io/otel/trace v0.17.0
+	go.opentelemetry.io/otel v1.0.0-RC1
```
spowelljr, line 75: Just curious, is there a reason we can't go to `v1.2.0`?

go.mod
```diff
@@ -217,28 +216,34 @@ replace (
...
+	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.22.4
+	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.22.4
+	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.22.4
+	k8s.io/kube-proxy => k8s.io/kube-proxy v0.22.4
+	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.22.4
+	k8s.io/kubectl => k8s.io/kubectl v0.22.4
+	k8s.io/kubelet => k8s.io/kubelet v0.22.4
+	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.22.4
+	k8s.io/metrics => k8s.io/metrics v0.22.4
+	k8s.io/mount-utils => k8s.io/mount-utils v0.22.5-rc.0
```
spowelljr, line 241: And do we want to be updating to RC versions?

go.mod
medyagh: that was done automatically

medyagh: but thats a good idea I changed it to 1.2.0

go.mod
medyagh: hm.... I think you are right, for simplicity we could just bump everything to 22.4

medyagh: done

PR #13100, opened 2021-12-06
log the virtualbox version in minikube start logs

### LINKED ISSUE #13060: log the virtualbox version in our minikube start logs
this info can come handy

and we can also warn ppl with very old vbox to update them

balalnx16:
fixes #13060

### DIFF
--- a/pkg/minikube/registry/drvs/virtualbox/virtualbox.go
+++ b/pkg/minikube/registry/drvs/virtualbox/virtualbox.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
+	"strconv"
 	"strings"
 	"time"
 
@@ -31,11 +32,13 @@ import (
 	"k8s.io/minikube/pkg/minikube/download"
 	"k8s.io/minikube/pkg/minikube/driver"
 	"k8s.io/minikube/pkg/minikube/localpath"
+	"k8s.io/minikube/pkg/minikube/out"
 	"k8s.io/minikube/pkg/minikube/registry"
 )
 
 const (
-	docURL = "https://minikube.sigs.k8s.io/docs/reference/drivers/virtualbox/"
+	docURL                    = "https://minikube.sigs.k8s.io/docs/reference/drivers/virtualbox/"
+	vboxSupportedMajorVersion = 5
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


spowelljr:
Hi @balalnx16, thanks for creating a PR, we need you to follow the steps in this comment before you can continue, thanks!

https://github.com/kubernetes/minikube/pull/13100#issuecomment-987059326

balalnx16:
Hi @spowelljr, Hope I resolved all the open issues. thanks

pkg/minikube/registry/drvs/virtualbox/virtualbox.go
```diff
@@ -111,5 +111,15 @@ func status() registry.State {
 		}
 	}
 
-	return registry.State{Installed: true, Healthy: true}
+	version := ""
+	cmd = exec.CommandContext(ctx, path, "--version")
```
medyagh, line 115: how about we combine this command with the one above?
instead of runnig
```
	cmd := exec.CommandContext(ctx, path, "list", "hostinfo")
```
we could run the version
that we we save time by running one command instead of two and save the version to be used too

pkg/minikube/registry/drvs/virtualbox/virtualbox.go
```diff
@@ -111,5 +111,15 @@ func status() registry.State {
...
+	stdout, verr := cmd.Output()
+	if verr != nil {
+		klog.Warningf("unable to get virtualbox version, returned: %v", err)
+	} else {
+		version = string(stdout)
```
medyagh, line 120: how about we check the version and it is bellow certain version (older than 1 year) to tell the user about their outdated vbox using 
```
out.WarningT
```

medyagh (changes requested):

pkg/minikube/registry/drvs/virtualbox/virtualbox.go
balalnx16: Hi @medyagh, I made the requested changes, Please review

pkg/minikube/registry/drvs/virtualbox/virtualbox.go
balalnx16: Hi @medyagh, Made the requested changes, please review

balalnx16:
Hi @medyagh @spowelljr 5 workflows are failing, are there any test issues or is there a problem with my change ? But those workflows seem to be unrelated to my change ? Should all these workflows succeed for the PR to be approved ? Please let me know if there is anything pending from side to move forward

spowelljr (approved):

PR #13110, opened 2021-12-07
gomod: remove depending on k8s.io/kubernetes as a lib

medyagh:
This should reduce a few gigabytes of dependency downloading for go mod and get rid of complicated go.mod replaces

```
$ du -sh ~/workspace/kubernetes/
1.1G	/Users/medya/workspace/kubernetes/
```

copying only the part of kubernetes/kubernetes repo that we rely on (constants and features packages)

```
$ git clone https://github.com/kubernetes/kubernetes
# pick latest stable version
$ git reset --hard v1.22.4

# cd into minikube

$ mkdir -p ./third_party/kubeadm/app/
$ cp -r ../kubernetes/cmd/kubeadm/app/features ./third_party/kubeadm/app/
$ cp -r ../kubernetes/cmd/kubeadm/app/constants ./third_party/kubeadm/app/
```


cmd/minikube/cmd/docker-env.go
```diff
@@ -50,6 +49,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/reason"
 	"k8s.io/minikube/pkg/minikube/shell"
 	"k8s.io/minikube/pkg/minikube/sysinit"
+	kconst "k8s.io/minikube/third_party/kubeadm/app/constants"
```
dims, line 52: cc @neolit123

afbjorklund:
Sounds like a good idea (we even talked about YAML), will have to look at the PR later.

third_party/kubeadm/README.md
```diff
@@ -0,0 +1,13 @@
+maintaining a fork of kubernetes constants to avoid depending on k8s.io/kubernetes/ as a lib
+how to update this fork
+
+# clone kuberentes
+
+# pick latest stable version
+$ git reset --hard v1.22.4
```
neolit123, line 7: probably the fastest way to get only a specific tag instead of cloning the whole k8s is:
```
git clone --depth 1 --branch v1.22.4 git@github.com:kubernetes/kubernetes.git
```

this would download ~30MB of objects vs ~1GB of objects.

cmd/minikube/cmd/docker-env.go
neolit123: the constants and features in use in minikube can technically be manually maintained in local constants and slices of features etc instead of being forked, but overall i don't see a big issue with keeping a small fork with a couple of files until kubeadm moves out.

the features package will likely be exposed publicly once that happens, but implementation details such as APICallRetryInterval are debatable, since they are more of an implementation detail - i.e. these could be constants/internal instead.

third_party/kubeadm/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+# cd into minikube
+
+$ mkdir -p ./third_party/kubeadm/app/
+$ cp -r ../kubernetes/cmd/kubeadm/app/features ./third_party/kubeadm/app/
+$ cp -r ../kubernetes/cmd/kubeadm/app/constants ./third_party/kubeadm/app/
```
neolit123, line 13: i think it's OK to not copy the _test.go files as long as the forked .go files are 1-to-1.

cmd/minikube/cmd/docker-env.go
medyagh: I agree and thank you for the review :)

sharifelgamal (approved):

klaases (approved):

PR #13112, opened 2021-12-07
Feature/add disk-usage command

### LINKED ISSUE #12967: Implement a `minikube disk-usage` command so users can see how much space minikube is using
There have been some comments by users that they find minikube uses a lot of disk space and there's no easy way to see how much it's using.

We should implement a command so users can easily view the disk space usage. This will include things like preloads and image caching and output the usage categorically.

mpuckett159:
fixes #12967

This PR implements a subcommand to minikube called disk-usage that will, by default, output a list of the child directories of the user's minikube directory in a human readable format. It has additional flags for printing the full, unrounded bytes of the directory sizes, and for only displaying the size of the root directory. The flags can be used serially. See below for what the new output looks like.

Example output:
```bash
❯ ./out/minikube disk-usage -h
Shows current disk usage of minikube

Options:
  -l, --long=false: Whether to return a human readable formatted number, e.g. 1GB
  -r, --root-only=false: Only show total of root minikube file instead of breaking out all sub directories

Usage:
  minikube disk-usage [flags] [options]

```

```bash
❯ ./out/minikube disk-usage              
    ▪ .minikube - 23.65GB
    ▪ addons - 0B
    ▪ bin - 11.43MB
    ▪ cache - 3.276GB
    ▪ certs - 8.272kB
    ▪ config - 2B
    ▪ files - 3.416kB
    ▪ logs - 175.9kB
    ▪ machines - 20.37GB
    ▪ profiles - 72.45kB
```

```bash
❯ ./out/minikube disk-usage -l
    ▪ .minikube - 23654741962
    ▪ addons - 0
    ▪ bin - 11428724
    ▪ cache - 3276493928
    ▪ certs - 8272
    ▪ config - 2
    ▪ files - 3416
    ▪ logs - 176318
    ▪ machines - 20366530406
    ▪ profiles - 72453
```

```bash
❯ ./out/minikube disk-usage -r
    ▪ .minikube - 23.65GB
```

```bash
❯ ./out/minikube disk-usage -r -l
    ▪ .minikube - 23654742710
```


medyagh:
I find the flags a bit confusing and also it is showing the disk usage outside minikube which what I first thought, ( I thought would show disk usage inside minikube node)

afbjorklund:
So basically, `du -hxcs ~/.minikube/*` ?

cmd/minikube/cmd/disk-usage.go
```diff
@@ -0,0 +1,128 @@
...
+		var decimalUnitPrefixMapKey int
+		if !longOutFlag {
+			totalFileSizeOut, decimalUnitPrefixMapKey = int64ToFloat32FormattedString(totalFileSize)
+		}
+
+		out.Infof("{{.directory}} - {{.diskSize}}{{.decimalUnitPrefix}}B", out.V{"directory": dirName, "diskSize": totalFileSizeOut, "decimalUnitPrefix": decimalUnitPrefixMap[decimalUnitPrefixMapKey]})
+	}
+}
+
+func int64ToFloat32FormattedString(in_int int64) (string, int) {
```
afbjorklund, line 89: See also github.com/docker/go-units

mpuckett159:
> I find the flags a bit confusing

Yeah I'm not super happy with the flags either. I wanted to mimic du more exactly but it wouldn't let me (or I couldn't figure out how to at least) override the `-h` flag for help to be the "human readable" flag.

cmd/minikube/cmd/disk-usage.go
mpuckett159: image

afbjorklund: The biggest fight is between the mebibytes and the megabytes

afbjorklund:
Note that "du" uses `M` to mean 1048576 bytes, which makes some people angry. (minikube does the same)

mpuckett159:
Updated the original PR comment to reflect the usage of the `github.com/docker/go-units` function `HumanSize` for formatting the size output for human readable formatting.

mpuckett159:
"also it is showing the disk usage outside minikube which what I first thought, ( I thought would show disk usage inside minikube node)"

@medyagh Could you explain this a little more for me? I seem to have misunderstood the ask some but I'm not exactly sure what part of it I missed. You can DM me in the K8S Slack as well if you want, username there is `@Marcus Puckett`

medyagh:
@mpuckett159 I think minikube has many commands and I would love to have a feature that shows the disk usage but I would like that to be part of a bigger picture that could also be inline to what most users are used to in simmilar tools

like docker

I would perfer a MEP on this before a PR sine this adds a new command, I would like more disscusssion on this

medyagh:
here I recommend adding a MEP for. this so we could discuss what is best way to go forward
https://github.com/kubernetes/minikube/tree/master/enhancements

mpuckett159:
Hah, MEP are my initials.

Ok I'll fill that out. Sorry got confused because the Issue linked said it was good for a PR but I guess given the obvious confusion here this route is best.

PR #13125, opened 2021-12-08
Refactored to remove config.ExtraOptions

spowelljr:
Refactored the `--extra-config` flag to a string flag and removed the `config.ExtraConfig` object to prevent modifying a global variable.

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -124,6 +124,7 @@ const (
 	listenAddress           = "listen-address"
 	extraDisks              = "extra-disks"
 	certExpiration          = "cert-expiration"
+	extraOptions            = "extra-config"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


cmd/minikube/cmd/start_flags.go
```diff
@@ -402,6 +403,18 @@ func getCNIConfig(cmd *cobra.Command) string {
 	return chosenCNI
 }
 
+func getExtraOptions() config.ExtraOptionSlice {
```
medyagh, line 406: add comment, say u mean kubernetes extra options

cmd/minikube/cmd/start_flags.go
```diff
@@ -402,6 +403,18 @@ func getCNIConfig(cmd *cobra.Command) string {
...
+	extraOptionVals := config.ExtraOptionSlice{}
+	val := viper.GetString(extraOptions)
+	if val == "" {
+		return extraOptionVals
+	}
+	if err := extraOptionVals.Set(val); err != nil {
```
klaases, line 412: Will this be able to handle multiple flags/options?

Or, if an error occurs, will all the options be ignored?

klaases: Update: please include the .Set() error message in the klog.Errorf().

cmd/minikube/cmd/start_flags.go
spowelljr: Done!

cmd/minikube/cmd/start_flags.go
spowelljr: Added the error message!

klaases (approved):

Issue #13003, opened 2021-11-22
Supporting file mounting with`--no-kubernetes` flag


spowelljr:
Currently, if the `--no-kubernetes` flag is passed, mounting is skipped, but users still want to be able to mount files with `--no-kubernetes` so we should not skip it.

spowelljr:
Change is done, waiting for https://github.com/kubernetes/minikube/pull/12930 to be merged first since it changes mount func signature.

PR #13144, opened 2021-12-10
Support mounting with the --no-kubernetes flag

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/13003

**Problem:**
If the `--no-kubernetes` flag is used, file mounting is skipped.

**Solution:**
Make sure mounting occurs whether `--no-kubernetes` is passed or not.

### DIFF
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -148,9 +148,7 @@ func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
 		}
 	}
 
-	if !driver.IsKIC(starter.Cfg.Driver) {
-		go configureMounts(&wg, *starter.Cfg)
-	}
+	go configureMounts(&wg, *starter.Cfg)
 
 	wg.Add(1)
 	go func() {
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -90,11 +90,13 @@ type Starter struct {
 
 // Start spins up a guest and starts the Kubernetes node.
 func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
+	var wg sync.WaitGroup
 	stop8ks, err := handleNoKubernetes(starter)
 	if err != nil {
 		return nil, err
 	}
 	if stop8ks {
+		configureMounts(&wg, *starter.Cfg)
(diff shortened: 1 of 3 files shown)


pkg/minikube/node/start.go
```diff
@@ -90,11 +90,13 @@ type Starter struct {
 
 // Start spins up a guest and starts the Kubernetes node.
 func Start(starter Starter, apiServer bool) (*kubeconfig.Settings, error) {
+	var wg sync.WaitGroup
 	stop8ks, err := handleNoKubernetes(starter)
 	if err != nil {
 		return nil, err
 	}
 	if stop8ks {
+		configureMounts(&wg, *starter.Cfg)
```
sharifelgamal, line 99: when kubernetes actually starts, this only gets run if the driver is VM-based. do we need to check for that here too?

spowelljr: Good call, that slipped my mind, I moved the check within `configureMounts` to simplify

test/integration/mount_start_test.go
```diff
@@ -81,7 +81,7 @@ func TestMountStart(t *testing.T) {
 func validateStartWithMount(ctx context.Context, t *testing.T, profile string) {
 	defer PostMortemLogs(t, profile)
 
-	args := []string{"start", "-p", profile, "--memory=2048", "--mount", "--mount-gid", mountGID, "--mount-msize", mountMSize, "--mount-port", mountPort, "--mount-uid", mountUID}
+	args := []string{"start", "-p", profile, "--memory=2048", "--mount", "--mount-gid", mountGID, "--mount-msize", mountMSize, "--mount-port", mountPort, "--mount-uid", mountUID, "--no-kubernetes"}
```
medyagh, line 84: are we changing this test to only test with no-kubernetes ? if yes why

spowelljr: I was going to change it to only test with `--no-kubernetes` since we're just checking that the `start --mount` flags are being respected so having k8s running is not important, this will also speed up the test since we're starting two minikube instances and they'll start much faster with `--no-kubernetes`.

spowelljr:
Seems some mount tests are failing, going to take a look

spowelljr:
The failing mount tests should be fixed with https://github.com/kubernetes/minikube/pull/13183

medyagh (approved):

PR #13148, opened 2021-12-12
docs: use ingress-dns addon inside cluster

oprudkyi:
docs improvement - use ingress-dns addon to resolve local DNS names inside cluster


medyagh (approved):

medyagh:
Thank you @oprudkyi  that is a nice tip ! this would be actually a good idea to add an full wallkthrough tutorial on dns to minikube website. please feel free to add one if interested

https://minikube.sigs.k8s.io/docs/tutorials/

PR #13156, opened 2021-12-13
update build image to bullseye

medyagh:
should fix https://github.com/kubernetes/minikube/pull/13155


sharifelgamal:
I think we need to do this in tandem with the golang upgrade?

medyagh:
> I think we need to do this in tandem with the golang upgrade?

they seem to have added the bulleye since 1.17.3

spowelljr:
Can confirm that they added bullseye for 1.17.3

/ok-to-test

sharifelgamal (approved):

Issue #13136, opened 2021-12-09
k8s v1.24.0-alpha.1: The image 'k8s.gcr.io/coredns/coredns:1.8.4' was not found; unable to add it to cache.


tstromberg:
### What Happened?

Using minikube from HEAD: 90300a4c5263a47a5457e9868bdb732f0a7f07d2

`./out/minikube start --kubernetes-version=v1.24.0-alpha.1`:

```
😄  minikube v1.24.0 on Darwin 11.5.2
✨  Automatically selected the hyperkit driver
...
👍  Starting control plane node minikube in cluster minikube
🔥  Creating hyperkit VM (CPUs=2, Memory=4000MB, Disk=20000MB) ...
❗  The image 'k8s.gcr.io/coredns/coredns:1.8.4' was not found; unable to add it to cache.
🐳  Preparing Kubernetes v1.24.0-alpha.1 on Docker 20.10.8 ...
❌  Unable to load cached images: loading cached images: stat /Users/tstromberg/.minikube/cache/images/k8s.gcr.io/coredns/coredns_1.8.4: no such file or directory
```

It does successfully progress

### Attach the log file

Filtered for coredns:

```
I1209 15:26:17.455637   91527 image.go:134] retrieving image: k8s.gcr.io/coredns/coredns:1.8.4
I1209 15:26:17.496764   91527 image.go:180] daemon lookup for k8s.gcr.io/coredns/coredns:1.8.4: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
W1209 15:26:17.499631   91527 image.go:190] authn lookup for k8s.gcr.io/coredns/coredns:1.8.4 (trying anon): error getting credentials - err: exec: "docker-credential-desktop": executable file not found in $PATH, out: ``
I1209 15:26:17.909258   91527 image.go:194] remote lookup for k8s.gcr.io/coredns/coredns:1.8.4: GET https://k8s.gcr.io/v2/coredns/coredns/manifests/1.8.4: MANIFEST_UNKNOWN: Failed to fetch "1.8.4" from request "/v2/coredns/coredns/manifests/1.8.4".
I1209 15:26:17.909368   91527 cache.go:96] cache image "k8s.gcr.io/coredns/coredns:1.8.4" -> "/Users/tstromberg/.minikube/cache/images/k8s.gcr.io/coredns/coredns_1.8.4" took 454.875267ms
[...]

PR #13165, opened 2021-12-14
include alpha releases in kubeadm const update

medyagh:
will help with 
https://github.com/kubernetes/minikube/issues/13136

our constants did not have alpha versions https://github.com/kubernetes/minikube/blob/master/pkg/minikube/constants/constants_kubeadm_images.go#L21


hack/update/github.go
```diff
@@ -235,7 +235,12 @@ func GHReleases(ctx context.Context, owner, repo string) (stable, latest string,
 				if semver.Compare(ver, latest) == 1 {
 					latest = ver
 				}
+			} else if strings.Contains(prerls, "-alpha") {
```
sharifelgamal, line 238: what about beta or rc? do we care about those?

medyagh: those are covered by latest above

sharifelgamal (approved):

PR #13168, opened 2021-12-14
Added env to suppress Docker performance messages

spowelljr:
We're having issues where Docker performance messages are being output to stderr and is making tests where we check to see if stederr is empty fail, for example `TestFunctional/parallel/ImageCommands/ImageLoadFromFile`.

Then passed the new env to tests so they messages will stop being displayed.

### DIFF
--- a/pkg/drivers/kic/oci/cli_runner.go
+++ b/pkg/drivers/kic/oci/cli_runner.go
@@ -21,8 +21,10 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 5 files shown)


pkg/drivers/kic/oci/cli_runner.go
```diff
@@ -84,6 +86,17 @@ func PrefixCmd(cmd *exec.Cmd) *exec.Cmd {
 	return cmd
 }
 
+func suppressDockerMessage() bool {
+	envKey := "MINIKUBE_SUPPRESS_DOCKER_PERFORMANCE"
+	env := os.Getenv(envKey)
+	suppress, err := strconv.ParseBool(env)
+	if err != nil {
+		klog.Errorf("failed to parse bool from the %s env, defaulting to 'false'; received: %s: %v", envKey, env, err)
```
sharifelgamal, line 94: this should be a warning at worst, maybe even an info

site/content/en/docs/handbook/config.md
```diff
@@ -114,14 +114,12 @@ Some features can only be accessed by minikube specific environment variables, h
 
 * **MINIKUBE_IN_STYLE** - (bool) manually sets whether or not emoji and colors should appear in minikube. Set to false or 0 to disable this feature, true or 1 to force it to be turned on.
 
-* **MINIKUBE_WANTUPDATENOTIFICATION** - (bool) sets whether the user wants an update notification for new minikube versions
```
sharifelgamal, line 117: are these being delete automatically? or do we not use them anymore?

spowelljr: I manually deleted them, I searched for the string and it wasn't included anywhere in the project. They exist in the config now, so I think they got moved there but this reference wasn't removed.

pkg/drivers/kic/oci/cli_runner.go
spowelljr: That's fair, my main concern was letting the user know that they messed by the config, but I can add an `out` message to satisfy that.

spowelljr: Updated

sharifelgamal (approved):

PR #13201, opened 2021-12-20
Fix french translation

jeffmaury:
Signed-off-by: Jeff MAURY <jmaury@redhat.com>


spowelljr (approved):
Thanks for keeping up with the translations as always!

PR #13221, opened 2021-12-21
Use minikube's own busybox images

spowelljr:
**Problem:**
We're using busybox images from Docker Hub, which get rate limited and cause our tests to fail.

**Solution:**
Push busybox images to minikube's own image repository and use that image to prevent the rate limiting.

### DIFF
--- a/test/integration/testdata/busybox-mount-test.yaml
+++ b/test/integration/testdata/busybox-mount-test.yaml
@@ -9,7 +9,7 @@ spec:
   restartPolicy: Never
   containers:
     - name: mount-munger
-      image: busybox:1.28.4-glibc
+      image: gcr.io/k8s-minikube/busybox:1.28.4-glibc
       command: ["/bin/sh", "-c", "--"]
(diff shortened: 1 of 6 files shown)


test/integration/testdata/busybox-mount-test.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   restartPolicy: Never
   containers:
     - name: mount-munger
-      image: busybox:1.28.4-glibc
+      image: gcr.io/k8s-minikube/busybox:1.28.4-glibc
```
sharifelgamal, line 12: did we push this tag to GCR as well?

sharifelgamal: nevermind, checked myself and it's there

spowelljr: Yup, I pushed `latest`, `1.28` & `1.28.4-glibc`

PR #13223, opened 2021-12-21
Fix TestFunctional/serial/LogsFileCmd test failure

spowelljr:
**Problem:**
We were only outputting stdout to the file, which was resulting in stderr still being output to the user, which should be logged to the file instead.

**Solution:**
Log both stdout and stderr to the file.


medyagh (approved):

medyagh:
functional verified started failing again

medyagh (approved):

PR #13251, opened 2021-12-28
Make the default container runtime dynamic

### LINKED ISSUE #13250: Make default container runtime dynamic
Since Kubernetes 1.24, there is no longer a "default" container runtime (since the removal of dockershim).

* https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker/

* https://kubernetes.io/blog/2021/11/12/are-you-ready-for-dockershim-removal/

Instead, the user will **have** to supply a container runtime. But the only _tested_ container runtime is containerd.

----

The current constant should be converted, into a default placeholder - which translates into a dynamic value

```go
        // Containerd is the default name and spelling for the containerd container runtime
        Containerd = "containerd"
        // CRIO is the default name and spelling for the cri-o container runtime
        CRIO = "crio"
        // DefaultContainerRuntime is our default container runtime
        DefaultContainerRuntime = "docker"
```

A new constant `Docker` should be introduced instead, to replace all the hardcoded `"docker"` strings...

User can still choose it as the container runtime (no matter the default), using `--container-runtime=docker`.

**Minikube should still provide a default container runtime for each k8s version, even if Kubernetes does not.**

`      --container-runtime='': The container runtime to be used. Valid options: docker, cri-o, containerd (default: auto)`

https://kubernetes.io/docs/setup/production-environment/container-runtimes/

> Items on this page refer to third party products or projects that provide functionality required by Kubernetes.
> The Kubernetes project authors aren't responsible for those third-party products or projects.

afbjorklund:
Since the dockershim removal, there is no longer a constant
default container runtime provided by upstream. Only CRI.

Closes #13250


medyagh (commented):
does this PR mean, if minikube default kubernete is 1.24+ 

`minikube start` without any args will start with containerd ?

afbjorklund:
> `minikube start` without any args will start with containerd ?

You don't know. It is dynamic and random, depending on k8s version.

It might be docker, it might be containerd. Something following the CRI.
Probably **not** rkt, but who knows. Maybe it will make a great comeback ?

If you want a special one, you have to use `--container-runtime` explicitly.

medyagh:
> > `minikube start` without any args will start with containerd ?
> 
> You don't know. It is dynamic and random, depending on k8s version.

so this might not be what some of our embedded users expect like "skaffold" who would expect docker runtime by default, this would make them have to explicitly add the flag....

I assume this means for 1.24+ k8s versions the dynamic will choose containerd ?

afbjorklund:
> I assume this means for 1.24+ k8s versions the dynamic will choose containerd ?

I gave up on that concept, since skaffold is dragging their legs and since docker is still the most popular

anyway, which container runtime should be selected by default was moved into a different issue

* #13255

sharifelgamal:
the DownloadOnly tests are failing because we're not always passing in container-runtime into our tests suites (and therefore into our minikube start calls). this is usually fine since we default back to docker in a normal codepath, but these tests call download.PreloadExists directly, and then passes and an empty string for containerRuntime (since ContainerRuntime() in main_test.go defaults to constants.DefaultContainerRuntime. 

the solution i think is to change one line (main_test.go at L165) that to constants.Docker.

sharifelgamal (approved):

PR #13259, opened 2021-12-29
Add custom addon loading

jcdickinson:
Let me know if this PR is something you're interested in merging, I'll maintain a fork if not. I am not massively experienced with Go, so apologies if there are any egregious mistakes.

# Load Addon Config From YAML

Addresses this TODO: https://github.com/kubernetes/minikube/blob/b1a39269f3279e76ae62a76ad3cca23a31a9f36c/pkg/minikube/assets/addons.go#L84

There are several changes as part of this:

1. Create all the yaml files for the existing addons. This was done by hand, and triple-checked, but more eyes can't hurt.
2. Create a `addon-registry.yaml` that lists all the addons.
3. Use URIs for loading addons. The embedded addons use the `embedfs` scheme, with the host `addon.<EmbedName>`.
4. I shuffled some types around in anticipation of the custom plugins, so there are a bunch of tiny changes across many files.
5. Given the URI approach, I wanted to avoid doing too much string manipulation. So I replaced `TargetName` and `TargetDir` with `TargetPath`: there are various fixups across the repo for this.
6. Dev docs have been updated.

# Allow Custom Addons

1. I cleaned up the existing addon type, creating `AddonPackage`. 
2. Allow templating opt-out again (addons and templates live under separate properties in the addon yaml).
3. I decided not to re-use `FileAsset`, as I didn't any write-backs hitting the original addon repo.
4. Addon repositories are stored under `.minikube/addons.json`.
5. `Addons` (now `AddonOverrides`) now only lists exceptional cases, as defaults are provided by the custom addon `isValidAddon`.

# Why Custom Addons?

Like, probably, most engineering teams we have a bunch of microservices. Developers aren't usually cross-cutting the micros, so there's no reason they can't use containers created by CI. We currently replicate a portion of minikube with bash scripts and docker compose, so I'd like us to just use minikube and a repo of minikube addons.

Happy new year!

*Re-opening for CLA*


medyagh:
sorry for the late comment, this does look an interesting PR ! does this mean users can have an addon without waiting for a new minikube to be released?
I would love us to invest if this is a good to invest in for minikube 1.27

medyagh:
adding more eyes on this PR @sharifelgamal and @afbjorklund for comments

jcdickinson:
> sorry for the late comment, this does look an interesting PR !

No problem, my fault for not pinging anyone on slack :)

> does this mean users can have an addon without waiting for a new minikube to be released?

Yup. I guess they would clone minikube and then add it to their addons config. You could consider moving addons into their own repo entirely (and just have minikube download it under `~/.cache` or something).

Another use-case is sharing addons inside a team, which is why I opened this PR.

medyagh:
@jcdickinson I am curious can we make this PR in a way that could run custom code as well ? (from the idea point of view) other than just deploying the yaml file

jcdickinson:
It would certainly make sense to allow plugins to run code, but what language would that code be in? Pulling in v8 feels like using a bazooka to kill a fly, maybe Lua?

Something, I am going through my changes and I no longer have any idea why some of the changes had to be made. Reviving this PR is going to be extremely difficult.

medyagh:
> It would certainly make sense to allow plugins to run code, but what language would that code be in? Pulling in v8 feels like using a bazooka to kill a fly, maybe Lua?
> 
> Something, I am going through my changes and I no longer have any idea why some of the changes had to be made. Reviving this PR is going to be extremely difficult.

I would be open to ideas ! lua seems like something that could work. 
would the user need to install anything extra on their end for lua?

PR #13263, opened 2021-12-30
flake charts: allow toggling between all-time and last 90 day flake charts

spowelljr:
- Added script that downloads `data.csv` removes all entries older than 90 days and uploads results as `data-last-90.csv`
- Updated html and js to allow toggling between all-time data and last 90 days

Top left of page when viewing all-time data:
Screen Shot 2021-12-29 at 5 10 23 PM

Top left of page when viewing last 90 days of data:
Screen Shot 2021-12-29 at 5 10 42 PM

Showing proof of different csv's being downloaded by different pages:
Screen Shot 2021-12-29 at 5 11 30 PM

Proof that chart is only showing last 90 days of data
Screen Shot 2021-12-29 at 5 12 38 PM


hack/jenkins/test-flake-chart/flake_chart.js
```diff
@@ -664,30 +665,39 @@ function displayEnvironmentChart(testData, environmentName) {
...
   if (desiredTest === undefined) {
     displayEnvironmentChart(testData, desiredEnvironment);
   } else {
     displayTestAndEnvironmentChart(testData, desiredTest, desiredEnvironment);
   }
   document.querySelector('#data_date_container').style.display = 'block';
   document.querySelector('#data_date').innerText = responseDate.toLocaleString();
+  let periodDisplay, newURL;
+  if (desiredPeriod === 'last90') {
+    newURL = window.location.href.replace(/&?period=last90/gi, '');
```
sharifelgamal, line 694: should these conditions be flipped?

spowelljr: This is creating the link to the opposite page, `desiredPeriod` is the current page, so I'm stripping the current link for the all data link that's displayed on the top of the page if that makes sense.

sharifelgamal: ok I get it. a little confusing so maybe a comment is in order.

otherwise lgtm

sharifelgamal (approved):

Issue #13252, opened 2021-12-28
Add nerdctl command for users outside kubernetes


afbjorklund:
The default `ctr` and `buildctl` are a bit rough around the edges (i.e. user-hostile)

There is another program called `nerdctl`, which looks more like `docker` and `podman`.

It doesn't have a remote protocol, so users will have to use `minikube ssh` to use it...

(i.e. there is no equivalent to `docker-env` and `podman-env`, except for ssh tunneling)

<https://github.com/containerd/nerdctl#command-reference>

https://minikube.sigs.k8s.io/docs/handbook/pushing/#6-pushing-directly-to-in-cluster-containerd-buildkitd

----

With Docker and Podman, there are both local clients as well as remote clients:

```
minikube ssh -- docker --help

eval $(minikube docker-env)
docker --help
```

```
minikube ssh -- sudo podman --help

eval $(minikube podman-env)
podman --remote --help
```

With nerdctl/containerd/buildkitd, there are **no** local clients but only remote client:

```
minikube ssh -- sudo nerdctl --help
```

```console
$ eval $(./out/minikube docker-env)
❌  Exiting due to MK_USAGE: The docker-env command is only compatible with the "docker" runtime, but this cluster was configured to use the "containerd" runtime.
$ eval $(./out/minikube podman-env)
❌  Exiting due to MK_USAGE: The podman-env command is only compatible with the "crio" runtime, but this cluster was configured to use the "containerd" runtime.
```

"local" client would be: running on the host, i.e. remotely (to server)
"remote" client would be: running on the node, i.e. locally (to server)

afbjorklund:
For regular use, there is `kubectl` and `crictl` as always. Or the "minikube image" wrapper.

afbjorklund:
Pre-built binaries: https://github.com/containerd/nerdctl/releases/tag/v0.15.0

```console
docker@minikube:~$ sudo nerdctl version
Client:
 Version:	v0.15.0
 Git commit:	b72b5ca14550b2e23a42787664b6182524c5053f

Server:
 containerd:
  Version:	1.4.9
  GitCommit:	e25210fe30a0a703442421b0f60afac609f950a3

docker@minikube:~$ sudo nerdctl -n k8s.io ps
docker@minikube:~$ sudo nerdctl -n k8s.io images
```

afbjorklund:
Rootless nerdctl/containerd is not supported, and will give an error message:

```console
docker@minikube:~$ nerdctl 
[...]

PR #13269, opened 2021-12-31
ISO: Add nerdctl from vendor binary

afbjorklund:
Adds `nerdctl`, for interacting with `containerd` and `buildkitd` (when using "containerd" container runtime)

See #13252


klaases (approved):

afbjorklund:
ok-to-build-iso

afbjorklund:
ok-to-build-iso

staticdev:
@sharifelgamal LGTM?

afbjorklund:
It just adds bloat to the image, it shouldn't really _break_ anything

catarse-natural:
@blueelvis
 @prasadkatti could you check if we can get nerdctl in minikube? Looks a great addition since containerd is the default k8s runtime.

deploy/iso/minikube-iso/package/nerdctl-bin/nerdctl-bin.mk
```diff
@@ -0,0 +1,19 @@
+################################################################################
+#
+# nerdctl-bin
+#
+################################################################################
+
+NERDCTL_BIN_VERSION = v0.17.1
```
AkihiroSuda, line 7: Now we have v0.18.0.

deploy/iso/minikube-iso/package/nerdctl-bin/nerdctl-bin.hash
```diff
@@ -0,0 +1,5 @@
+sha256 1371da3f6bd461f331946654f6dd3ef2ef4b9da0dd7bc5f78ed1166f32ad5adc  nerdctl-0.15.0-linux-amd64.tar.gz
+sha256 00abb395a6c7c19f2e0612a65e413534b6ec42e995bfbcabd59886b99ca5e43e  nerdctl-0.16.0-linux-amd64.tar.gz
+sha256 543d58c057a5a59c67ca523016c1dac57f9a09d1d9e9c98bd092b9df82ec5246  nerdctl-0.16.1-linux-amd64.tar.gz
+sha256 4c08a6ce657ff851dd7a7b1d21c64f1c1950e35de03fa7f1853eab47fa2b2d53  nerdctl-0.17.0-linux-amd64.tar.gz
+sha256 6d18d3eaa74cd2f8e36e51f551282355d0bb1dd962ecd2d1dfecee161d0bd39a  nerdctl-0.17.1-linux-amd64.tar.gz
```
AkihiroSuda, line 5: Probably you don't need to retain old versions, and you can just squash your commits.

deploy/iso/minikube-iso/package/nerdctl-bin/nerdctl-bin.mk
staticdev: @afbjorklund version 0.18 supports `nerdctl build` for local images, may help replacing `docker build`.

afbjorklund: All of the versions use `buildctl build` behind the scenes, but it can't replace `docker build` without a remote API.

This was why we had to invent the "minikube image" client, so that we could transport the build context to the VM.

The thing that is new in 0.18 is that it allows using containerd also for rootless builds, which we don't support (yet).

So I don't think it makes any difference for the minikube use case, and we need more docs for using nerdctl with k8s.

afbjorklund:
I will re-open this for the next iteration (1.27), once the other more important features are done.

* https://github.com/kubernetes/minikube/issues/13252

afbjorklund:
> Looks a great addition since containerd is the default k8s runtime.

minikube is still using docker as the default container runtime, for 1.24

PR #13295, opened 2022-01-04
Add hint to translation check to documentation

jklippel:
Added hint that one can run the translate check. This checks  whether the translations are at least syntactically correct.


sharifelgamal (approved):
Thank you for help improving our documentation!

PR #13297, opened 2022-01-05
Increase test timeout for Docker Windows

spowelljr:
`TestFunctional/parallel/InternationalLanguage` & `TestFunctional/parallel/DryRun` are failing on Docker Windows at 100% because the test timeout is too short.


spowelljr:
Both tests passed

sharifelgamal (approved):

Issue #10158, opened 2021-01-16
minikube incorrectly assumes the DOCKER_HOST is always a numeric IP address


alanblairverint:
I am unable to get minikube to start on Windows 10 using the following versions:

**Minikube**
minikube version: v1.16.0
commit: 9f1e482427589ff8451c4723b6ba53bb9742fbb1

**Kubectl:**
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.0", GitCommit:"af46c47ce925f4c4ad5cc8d1fca46c7b77d13b38", GitTreeState:"clean", BuildDate:"2020-12-08T17:59:43Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"windows/amd64"}

**Docker:**
Client: Docker Engine - Community
 Cloud integration: 1.0.7
 Version:           20.10.2
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        2291f61
 Built:             Mon Dec 28 16:14:16 2020
 OS/Arch:           windows/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.2
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       8891c58
  Built:            Mon Dec 28 16:15:28 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

minikube start --driver=docker --alsologtostderr

Steps to reproduce the issue:

1. Fresh install of minikbe & docker desktop
2. Run minikube start --driver=docker --alsologtostderr

Logs
c:\>minikube start --driver=docker --alsologtostderr
I0116 18:32:49.280441    3812 out.go:221] Setting OutFile to fd 92 ...
I0116 18:32:49.284427    3812 out.go:268] TERM=,COLORTERM=, which probably does not support color
I0116 18:32:49.284427    3812 out.go:234] Setting ErrFile to fd 96...
I0116 18:32:49.284427    3812 out.go:268] TERM=,COLORTERM=, which probably does not support color
[...]

PR #13300, opened 2022-01-05
Support DOCKER_HOST not being numeric IP

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/10158

We always assumed that the IP was going to be numeric, but people are using hostnames and `localhost`, so adding an IP lookup to get the true IP.

### DIFF
--- a/pkg/minikube/driver/endpoint.go
+++ b/pkg/minikube/driver/endpoint.go
@@ -35,16 +35,16 @@ func ControlPlaneEndpoint(cc *config.ClusterConfig, cp *config.Node, driverName 
 		}
 		hostname := oci.DaemonHost(driverName)
 
-		ip := net.ParseIP(hostname)
-		if ip == nil {
-			return hostname, ip, port, fmt.Errorf("failed to parse ip for %q", hostname)
+		ips, err := net.LookupIP(hostname)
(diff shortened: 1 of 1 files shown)


pkg/minikube/driver/endpoint.go
```diff
@@ -52,10 +52,12 @@ func ControlPlaneEndpoint(cc *config.ClusterConfig, cp *config.Node, driverName
 	if cc.KubernetesConfig.APIServerName != constants.APIServerName {
 		hostname = cc.KubernetesConfig.APIServerName
 	}
-	ip := net.ParseIP(cp.IP)
-	if ip == nil {
-		return hostname, ip, cp.Port, fmt.Errorf("failed to parse ip for %q", cp.IP)
+	ips, err := net.LookupIP(cp.IP)
+	if err != nil || len(ips) == 0 {
+		return hostname, net.IP{}, cp.Port, fmt.Errorf("failed to lookup ip for %q", cp.IP)
```
klaases: Would we be able to return `nil` here instead of of `net.IP{}`?

klaases (approved):

sharifelgamal (approved):

PR #13301, opened 2022-01-05
WIP: Build default configuration for ARM64 to test.

klaases:
I would like to build the default config for ARM64 in order to see if there are any missing dependencies or error when executing on my build machines, both local and remote.


klaases:
ok-to-build-iso

klaases:
Note that cmake is required for building the ISO.
https://cmake.org/install/

klaases:
Need to install cmake on remote machine(s).

`CMake Error at CMakeLists.txt:15 (project):
  No CMAKE_CXX_COMPILER could be found.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.`

klaases:
There seems to be an issue with GCC, going to purge and re-install:

`sudo apt purge gcc-10 gcc c-compiler && sudo apt autoremove && sudo apt install gcc g++ build-essential`

klaases:
Our Dockerfile contains `build-essential`, which should include the compilers necessary to satisfy CMAKE, however, the following error continues to occur.

Is there something else that must be added to the Dockerfile in order to build?

**Dockerfile**
deploy/iso/minikube-iso/Dockerfile

**Error:**
```
No CMAKE_CXX_COMPILER could be found.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.
```

afbjorklund:
@klaases 

> Is there something else that must be added to the Dockerfile in order to build?

Compare:

https://github.com/buildroot/buildroot/blob/2021.02.x/support/docker/Dockerfile

klaases:
Closing as this will be included in a separate PR.

Issue #11141, opened 2021-04-19
add support of 802.1Q kernel module


sahid:
The current kernel build does not provide support for 802.1Q module.

This module is adding ability to configure VLAN sub-interfaces which are for networking use-cases more than required.

> Select this and you will be able to create 802.1Q VLAN interfaces on your Ethernet interfaces. 802.1Q VLAN supports almost everything a regular Ethernet interface does, including firewalling, bridging, and of course IP traffic. You will need the 'ip' utility in order to effectively use VLANs. See the VLAN web page for more information: http://www.candelatech.com/~greear/vlan.html

https://cateee.net/lkddb/web-lkddb/VLAN_8021Q.html

```
diff --git a/deploy/iso/minikube-iso/board/coreos/minikube/linux_defconfig b/deploy/iso/minikube-iso/board/coreos/minikube/linux_defconfig
index 1871f86d8..5c8ce5465 100644
--- a/deploy/iso/minikube-iso/board/coreos/minikube/linux_defconfig
+++ b/deploy/iso/minikube-iso/board/coreos/minikube/linux_defconfig
@@ -273,6 +273,7 @@ CONFIG_BRIDGE_EBT_SNAT=m
 CONFIG_BRIDGE_EBT_LOG=m
 CONFIG_BRIDGE_EBT_NFLOG=m
 CONFIG_BRIDGE=m
+CONFIG_VLAN_8021Q=m
 CONFIG_NET_SCHED=y
 CONFIG_NET_SCH_HTB=y
 CONFIG_NET_SCH_PRIO=y
```

afbjorklund:
What are these extra VLAN used for ?

sahid:
Without this module we can't create VLAN interfaces, example using:

```
ip link add link eth1 name eth1.100 type vlan id 100
```

In some workflows users may want to have their Pods attached to a VLAN tagged network to communicate.

medyagh:
@sahid does our KIC base have this module ?

it sounds reasonable, we could add this module, I would accept and review a PR that does that

sharifelgamal:
This is still something we'd be happy to accept PRs for.

PR #13304, opened 2022-01-06
add support of 802.1Q kernel module

sahid:
This module is adding ability to configure VLAN sub-interfaces.

Signed-off-by: Sahid Orentino Ferdjaoui <sahid.ferdjaoui@industrialdiscipline.com>
Fixes: #11141


sahid:
I think I'm missing something to make EasyCLA happy, any idea? I have signed it.

PR #13307, opened 2022-01-06
Write commands to audit log regardless of whether it terminates cleanly

### LINKED ISSUE #13186: Audit log defer not always being called
When any minikube command is run we have a [defer statement](https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/root.go#L81) that logs the command to the audit log. However, on `minikube start` sometimes we terminate the execution in a way that the defer is not run and then the start command is not added to the audit log.

I have a proposed way to remedy this, the only reason we defer it is so we can add then end time to the audit log, but to prevent the defer issue we can log everything but the end time to the audit log, and then if the command completes we can add the end time to the audit entry.

The one case we have to watch out for is if the user starts two `minikube start` commands at the same time, we simply can't modify the last entry in the log, so we'll need to make sure to track the index of the entry in the audit log.

Example:
```
# 1st minikube start
$ minikube start -p p1

# Audit log
minikube start -p p1 started at 1:00

# 2nd minikube start
$ minikube start -p p2

# Audit log
minikube start -p p1 started at 1:00
minikube start -p p2 started at 1:01

# 1st minikube start finishs

# Need to make sure we update the second last entry
minikube start -p p1 started at 1:00 ended at 1:03
minikube start -p p2 started at 1:01
```

NikhilSharmaWe:
Fixes #13186

**Before (note start entry is not present):**
```
$ minikube delete
$ minikube delete
$ minikube start --driver docker
😄  minikube v1.26.0-beta.1 on Darwin 12.4

💣  Exiting due to PROVIDER_DOCKER_NOT_RUNNING: "docker version --format -" exit status 1: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
💡  Suggestion: Start the Docker service
📘  Documentation: https://minikube.sigs.k8s.io/docs/drivers/docker/

$ minikube logs --audit

==> Audit <==
|---------|------|----------|--------------|----------------|---------------------|---------------------|
| Command | Args | Profile  |     User     |    Version     |     Start Time      |      End Time       |
|---------|------|----------|--------------|----------------|---------------------|---------------------|
| delete  |      | minikube | powellsteven | v1.26.0-beta.1 | 14 Jun 22 15:44 PDT | 14 Jun 22 15:44 PDT |
| delete  |      | minikube | powellsteven | v1.26.0-beta.1 | 14 Jun 22 15:44 PDT | 14 Jun 22 15:44 PDT |
|---------|------|----------|--------------|----------------|---------------------|---------------------|
```

**After (note start entry is present):**
```
$ minikube delete
$ minikube delete
$ minikube start --driver docker
😄  minikube v1.26.0-beta.1 on Darwin 12.4

💣  Exiting due to PROVIDER_DOCKER_NOT_RUNNING: "docker version --format -" exit status 1: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
💡  Suggestion: Start the Docker service
📘  Documentation: https://minikube.sigs.k8s.io/docs/drivers/docker/

$ minikube logs --audit

==> Audit <==
|---------|-----------------|----------|--------------|----------------|---------------------|---------------------|
| Command |      Args       | Profile  |     User     |    Version     |     Start Time      |      End Time       |
|---------|-----------------|----------|--------------|----------------|---------------------|---------------------|
| delete  |                 | minikube | powellsteven | v1.26.0-beta.1 | 14 Jun 22 15:42 PDT | 14 Jun 22 15:42 PDT |
| delete  |                 | minikube | powellsteven | v1.26.0-beta.1 | 14 Jun 22 15:42 PDT | 14 Jun 22 15:42 PDT |
| start   | --driver docker | minikube | powellsteven | v1.26.0-beta.1 | 14 Jun 22 15:42 PDT |                     |
|---------|-----------------|----------|--------------|----------------|---------------------|---------------------|
```

### DIFF
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -23,7 +23,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
-	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -84,8 +83,16 @@ var RootCmd = &cobra.Command{
 // Execute adds all child commands to the root command sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
-	defer audit.Log(time.Now())
-
+	auditID, err := audit.LogCommandStart()
+	if err != nil {
+		klog.Errorf("failed to log command start to audit: %v", err)
+	}
+	defer func() {
+		err := audit.LogCommandEnd(auditID)
+		if err != nil {
+			klog.Errorf("failed to log command end to audit: %v", err)
+		}
+	}()
 	// Check whether this is a windows binary (.exe) running inisde WSL.
 	if runtime.GOOS == "windows" && detect.IsMicrosoftWSL() {
 		var found = false
--- a/pkg/minikube/audit/audit.go
+++ b/pkg/minikube/audit/audit.go
@@ -84,16 +84,15 @@ func LogCommandEnd(id string) error {
 			return fmt.Errorf("failed to set the log file: %v", err)
 		}
 	}
-	_, err := currentLogFile.Seek(0, 0)
-	if err != nil {
-		return fmt.Errorf("failed to offset for the next Read or Write on currentLogFile: %v", err)
+	if err := seekToBeginning(); err != nil {
+		return err
 	}
 	var logs []string
 	s := bufio.NewScanner(currentLogFile)
 	for s.Scan() {
 		logs = append(logs, s.Text())
 	}
-	if err = s.Err(); err != nil {
+	if err := s.Err(); err != nil {
 		return fmt.Errorf("failed to read from audit file: %v", err)
 	}
 	rowSlice, err := logsToRows(logs)
@@ -117,9 +116,8 @@ func LogCommandEnd(id string) error {
 	if entriesNeedsToUpdate == 0 {
 		return fmt.Errorf("failed to find a log row with id equals to %v", id)
 	}
-	_, err = currentLogFile.Seek(0, 0)
-	if err != nil {
-		return fmt.Errorf("failed to offset for the next Read or Write on currentLogFile: %v", err)
+	if err := currentLogFile.Truncate(0); err != nil {
+		return err
 	}
 	_, err = currentLogFile.Write([]byte(auditContents))
 	if err != nil {
@@ -140,7 +138,7 @@ func shouldLog() bool {
 	}
 
 	// commands that should not be logged.
-	no := []string{"status", "version"}
+	no := []string{"status", "version", "logs", "generate-docs"}
 	a := pflag.Arg(0)
 	for _, c := range no {
 		if a == c {
--- a/pkg/minikube/audit/audit.go
+++ b/pkg/minikube/audit/audit.go
@@ -31,6 +31,7 @@ import (
 	"github.com/spf13/viper"
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/constants"
+	"k8s.io/minikube/pkg/minikube/localpath"
 	"k8s.io/minikube/pkg/version"
 )
 
@@ -116,7 +117,14 @@ func LogCommandEnd(id string) error {
 	if entriesNeedsToUpdate == 0 {
 		return fmt.Errorf("failed to find a log row with id equals to %v", id)
 	}
-	if err := currentLogFile.Truncate(0); err != nil {
+	// have to close the audit file, truncate it, then reopen it as Windows can't truncate an open file
+	if err := currentLogFile.Close(); err != nil {
+		return fmt.Errorf("failed to close audit file: %v", err)
+	}
+	if err := os.Truncate(localpath.AuditLog(), 0); err != nil {
+		return fmt.Errorf("failed to truncate audit file: %v", err)
+	}
+	if err := setLogFile(); err != nil {
 		return err
 	}
 	_, err = currentLogFile.Write([]byte(auditContents))
--- a/pkg/minikube/audit/audit.go
+++ b/pkg/minikube/audit/audit.go
@@ -17,17 +17,24 @@ limitations under the License.
 package audit
 
 import (
+	"bufio"
+	"encoding/json"
+	"fmt"
 	"os"
 	"os/user"
 	"strings"
+	"sync"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
-	"k8s.io/klog/v2"
 	"k8s.io/minikube/pkg/minikube/config"
+	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/version"
 )
+
+var mutex sync.Mutex
 
 // userName pulls the user flag, if empty gets the os username.
 func userName() string {
@@ -52,14 +59,73 @@ func args() string {
 }
 
 // Log details about the executed command.
-func Log(startTime time.Time) {
-	if !shouldLog() {
-		return
+func LogCommandStart() (string, error) {
+	mutex.Lock()
+	defer mutex.Unlock()
+	if len(os.Args) < 2 || !shouldLog() {
+		return "", nil
 	}
-	r := newRow(pflag.Arg(0), args(), userName(), version.GetVersion(), startTime, time.Now())
+	id := uuid.New().String()
+	r := newRow(pflag.Arg(0), args(), userName(), version.GetVersion(), time.Now(), id)
 	if err := appendToLog(r); err != nil {
-		klog.Warning(err)
+		return "", err
 	}
+	return r.id, nil
+}
+
+func LogCommandEnd(id string) error {
+	mutex.Lock()
+	defer mutex.Unlock()
+	if id == "" {
+		return nil
+	}
+	if currentLogFile == nil {
+		if err := setLogFile(); err != nil {
+			return fmt.Errorf("failed to set the log file: %v", err)
+		}
+	}
+	_, err := currentLogFile.Seek(0, 0)
+	if err != nil {
+		return fmt.Errorf("failed to offset for the next Read or Write on currentLogFile: %v", err)
+	}
+	var logs []string
+	s := bufio.NewScanner(currentLogFile)
+	for s.Scan() {
+		logs = append(logs, s.Text())
+	}
+	if err = s.Err(); err != nil {
+		return fmt.Errorf("failed to read from audit file: %v", err)
+	}
+	rowSlice, err := logsToRows(logs)
+	if err != nil {
+		return fmt.Errorf("failed to convert logs to rows: %v", err)
+	}
+	auditContents := ""
+	var entriesNeedsToUpdate int
+	for _, v := range rowSlice {
+		if v.id == id {
+			v.endTime = time.Now().Format(constants.TimeFormat)
+			v.Data = v.toMap()
+			entriesNeedsToUpdate++
+		}
+		auditLog, err := json.Marshal(v)
+		if err != nil {
+			return err
+		}
+		auditContents += string(auditLog) + "\n"
+	}
+	if entriesNeedsToUpdate == 0 {
+		return fmt.Errorf("failed to find a log row with id equals to %v", id)
+	}
+	_, err = currentLogFile.Seek(0, 0)
+	if err != nil {
+		return fmt.Errorf("failed to offset for the next Read or Write on currentLogFile: %v", err)
+	}
+	_, err = currentLogFile.Write([]byte(auditContents))
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
 // shouldLog returns if the command should be logged.
--- a/pkg/minikube/audit/audit.go
+++ b/pkg/minikube/audit/audit.go
@@ -76,6 +76,7 @@ func LogCommandEnd(id string) error {
 	if err := openAuditLog(); err != nil {
 		return err
 	}
+	defer closeAuditLog()
 	var logs []string
 	s := bufio.NewScanner(currentLogFile)
 	for s.Scan() {
@@ -84,9 +85,7 @@ func LogCommandEnd(id string) error {
 	if err := s.Err(); err != nil {
 		return fmt.Errorf("failed to read from audit file: %v", err)
 	}
-	if err := closeAuditLog(); err != nil {
-		return err
-	}
+	closeAuditLog()
 	rowSlice, err := logsToRows(logs)
 	if err != nil {
 		return fmt.Errorf("failed to convert logs to rows: %v", err)
@@ -118,7 +117,7 @@ func LogCommandEnd(id string) error {
 	if _, err = currentLogFile.Write([]byte(auditContents)); err != nil {
 		return fmt.Errorf("failed to write to audit log: %v", err)
 	}
-	return closeAuditLog()
+	return nil
 }
 
 // shouldLog returns if the command should be logged.
--- a/pkg/minikube/audit/audit.go
+++ b/pkg/minikube/audit/audit.go
@@ -23,7 +23,6 @@ import (
 	"os"
 	"os/user"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/google/uuid"
@@ -34,8 +33,6 @@ import (
 	"k8s.io/minikube/pkg/minikube/localpath"
 	"k8s.io/minikube/pkg/version"
 )
-
-var mutex sync.Mutex
 
 // userName pulls the user flag, if empty gets the os username.
 func userName() string {
@@ -61,8 +58,6 @@ func args() string {
 
 // Log details about the executed command.
 func LogCommandStart() (string, error) {
-	mutex.Lock()
-	defer mutex.Unlock()
 	if len(os.Args) < 2 || !shouldLog() {
 		return "", nil
 	}
@@ -75,17 +70,10 @@ func LogCommandStart() (string, error) {
 }
 
 func LogCommandEnd(id string) error {
-	mutex.Lock()
-	defer mutex.Unlock()
 	if id == "" {
 		return nil
 	}
-	if currentLogFile == nil {
-		if err := setLogFile(); err != nil {
-			return fmt.Errorf("failed to set the log file: %v", err)
-		}
-	}
-	if err := seekToBeginning(); err != nil {
+	if err := openAuditLog(); err != nil {
 		return err
 	}
 	var logs []string
@@ -95,6 +83,9 @@ func LogCommandEnd(id string) error {
 	}
 	if err := s.Err(); err != nil {
 		return fmt.Errorf("failed to read from audit file: %v", err)
+	}
+	if err := closeAuditLog(); err != nil {
+		return err
 	}
 	rowSlice, err := logsToRows(logs)
 	if err != nil {
@@ -117,21 +108,17 @@ func LogCommandEnd(id string) error {
 	if entriesNeedsToUpdate == 0 {
 		return fmt.Errorf("failed to find a log row with id equals to %v", id)
 	}
-	// have to close the audit file, truncate it, then reopen it as Windows can't truncate an open file
-	if err := currentLogFile.Close(); err != nil {
-		return fmt.Errorf("failed to close audit file: %v", err)
-	}
+	// have to truncate the audit log while closed as Windows can't truncate an open file
 	if err := os.Truncate(localpath.AuditLog(), 0); err != nil {
-		return fmt.Errorf("failed to truncate audit file: %v", err)
+		return fmt.Errorf("failed to truncate audit log: %v", err)
 	}
-	if err := setLogFile(); err != nil {
+	if err := openAuditLog(); err != nil {
 		return err
 	}
-	_, err = currentLogFile.Write([]byte(auditContents))
-	if err != nil {
-		return err
+	if _, err = currentLogFile.Write([]byte(auditContents)); err != nil {
+		return fmt.Errorf("failed to write to audit log: %v", err)
 	}
-	return nil
+	return closeAuditLog()
 }
 
 // shouldLog returns if the command should be logged.
--- a/pkg/minikube/audit/audit_test.go
+++ b/pkg/minikube/audit/audit_test.go
@@ -20,7 +20,6 @@ import (
 	"os"
 	"os/user"
 	"testing"
-	"time"
 
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
@@ -177,7 +176,27 @@ func TestAudit(t *testing.T) {
 	})
 
 	// Check if logging with limited args causes a panic
-	t.Run("Log", func(t *testing.T) {
+	t.Run("LogCommandStart", func(t *testing.T) {
+		oldArgs := os.Args
+		defer func() { os.Args = oldArgs }()
+		os.Args = []string{"minikube", "start"}
+
+		oldCommandLine := pflag.CommandLine
+		defer func() {
+			pflag.CommandLine = oldCommandLine
+			pflag.Parse()
+		}()
+		mockArgs(t, os.Args)
+		auditID, err := LogCommandStart()
+		if auditID == "" {
+			t.Fatal("audit ID should not be empty")
+		}
+		if err != nil {
+			t.Fatal(err)
+		}
+	})
+
+	t.Run("LogCommandEnd", func(t *testing.T) {
 		oldArgs := os.Args
 		defer func() { os.Args = oldArgs }()
 		os.Args = []string{"minikube"}
@@ -188,8 +207,32 @@ func TestAudit(t *testing.T) {
 			pflag.Parse()
 		}()
 		mockArgs(t, os.Args)
+		auditID, err := LogCommandStart()
+		if err != nil {
+			t.Fatal("start failed")
+		}
+		err = LogCommandEnd(auditID)
 
-		Log(time.Now())
+		if err != nil {
+			t.Fatal(err)
+		}
+	})
+
+	t.Run("LogCommandEnd", func(t *testing.T) {
+		oldArgs := os.Args
+		defer func() { os.Args = oldArgs }()
+		os.Args = []string{"minikube"}
+
+		oldCommandLine := pflag.CommandLine
+		defer func() {
+			pflag.CommandLine = oldCommandLine
+			pflag.Parse()
+		}()
+		mockArgs(t, os.Args)
+		err := LogCommandEnd("non-existing-id")
+		if err == nil {
+			t.Fatal("function LogCommandEnd should return an error when a non-existing id is passed in it as an argument")
+		}
 	})
 }
 
--- a/pkg/minikube/audit/audit_test.go
+++ b/pkg/minikube/audit/audit_test.go
@@ -218,7 +218,7 @@ func TestAudit(t *testing.T) {
 		}
 	})
 
-	t.Run("LogCommandEnd", func(t *testing.T) {
+	t.Run("LogCommandEndNonExistingID", func(t *testing.T) {
 		oldArgs := os.Args
 		defer func() { os.Args = oldArgs }()
 		os.Args = []string{"minikube"}
--- a/pkg/minikube/audit/row_test.go
+++ b/pkg/minikube/audit/row_test.go
@@ -23,6 +23,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/uuid"
 	"k8s.io/minikube/pkg/minikube/constants"
 )
 
@@ -36,8 +37,10 @@ func TestRow(t *testing.T) {
 	stFormatted := st.Format(constants.TimeFormat)
 	et := time.Now()
 	etFormatted := et.Format(constants.TimeFormat)
+	id := uuid.New().String()
 
-	r := newRow(c, a, u, v, st, et, p)
+	r := newRow(c, a, u, v, st, id, p)
+	r.endTime = etFormatted
 
 	t.Run("NewRow", func(t *testing.T) {
 		tests := []struct {
@@ -51,7 +54,7 @@ func TestRow(t *testing.T) {
 			{"user", r.user, u},
 			{"version", r.version, v},
 			{"startTime", r.startTime, stFormatted},
-			{"endTime", r.endTime, etFormatted},
+			{"id", r.id, id},
 		}
 
 		for _, tt := range tests {
@@ -83,7 +86,7 @@ func TestRow(t *testing.T) {
 			{"user", u},
 			{"version", v},
 			{"startTime", stFormatted},
-			{"endTime", etFormatted},
+			{"id", id},
 		}
 
 		for _, tt := range tests {
@@ -97,7 +100,7 @@ func TestRow(t *testing.T) {
 	t.Run("toFields", func(t *testing.T) {
 		got := r.toFields()
 		gotString := strings.Join(got, ",")
-		want := []string{c, a, p, u, v, stFormatted, etFormatted}
+		want := []string{c, a, p, u, v, stFormatted, etFormatted, id}
 		wantString := strings.Join(want, ",")
 
 		if gotString != wantString {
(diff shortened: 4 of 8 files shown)


spowelljr:
@NikhilSharmaWe
```
Error: cmd/minikube/cmd/root.go:26:2: "time" imported but not used (typecheck)
	"time"
	^
Error: pkg/minikube/audit/audit.go:57:3: wrong number of return values (want 1, got 0) (typecheck)
		return
		^
Error: pkg/minikube/audit/audit.go:70:11: no result values expected (typecheck)
			return nil, fmt.Errorf("failed to set the log file: %v", err)
			       ^
Error: pkg/minikube/audit/audit.go:73:31: cannot use currentLogFile (variable of type *os.File) as []string value in argument to logsToRows (typecheck)
	for _, v := range logsToRows(currentLogFile) {
	                             ^
Error: pkg/minikube/audit/audit_test.go:178:3: undeclared name: `Log` (typecheck)
		Log(time.Now())
		^
Error: pkg/minikube/audit/logFile_test.go:51:62: cannot use time.Now() (value of type time.Time) as string value in argument to newRow (typecheck)
		r := newRow("start", "-v", "user1", "v0.17.1", time.Now(), time.Now())
		                                                           ^
Error: pkg/minikube/audit/row_test.go:40:30: cannot use et (variable of type time.Time) as string value in argument to newRow (typecheck)
	r := newRow(c, a, u, v, st, et, p)
```

You can check for these failures locally by running `make test`

NikhilSharmaWe:
@spowelljr Yes thanks, I am working on them.

pkg/minikube/audit/audit.go
```diff
@@ -51,14 +53,34 @@ func args() string {
...
+	return r.id, nil
+}
+
+func LogCommandEnd(id string) err{
+	if currentLogFile == nil {
+		if err := setLogFile(); err != nil {
+			return fmt.Errorf("failed to set the log file: %v", err)
+		}
+	}
+	auditLogs, err := logsToRows(currentLogFile)
```
NikhilSharmaWe, line 74: @spowelljr How can we convert `currentLogFile` from `*os.File` type to `[]string` type. We can use `ioutil.ReadFile()` function for converting it to []byte and then string but how can we convert it to []string type so that we can convert to []row through `logsToRows()` function.

spowelljr: You can use https://pkg.go.dev/encoding/json#Unmarshal to convert `[]byte` to a struct (row).

NikhilSharmaWe:
@spowelljr We are getting error where funcs `Log()` and `newRow()` are used other than `audit.go` and `root.go`. When previously I checked `Log()` function was not used any other place than `root.go` but now I think it is not the case. 
How can we deal with them ?

spowelljr:
> @spowelljr We are getting error where funcs `Log()` and `newRow()` were used other than `audit.go` and `root.go`. When previously I checked `Log()` function was not used any other place than `root.go` but now I think it is not the case.
> 
> * I think we should add functions `Log()` and `newRow()`(with older logic) along side with the new functions which we have added. I will change name of `newRow()` function which we have used in `LogCommandStart()` to `newRowStart()`.
> 
> Could you please give your thoughts on this.

Those are the audit test files that are giving you errors, they need to updated as the func changes.

NikhilSharmaWe:
@spowelljr 
Could you please changes I have made in test files.
Am I going in the right direction through these changes or not ?

cmd/minikube/cmd/root.go
```diff
@@ -78,8 +77,16 @@ var RootCmd = &cobra.Command{
 // Execute adds all child commands to the root command sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
-	defer audit.Log(time.Now())
-
+	auditID, err := audit.LogCommandStart()
+	if err != nil{
+		klog.Errorf("%v", err)
```
spowelljr, line 82:
```suggestion
		klog.Errorf("failed to log command start to audit: %v", err)
```
The error messages should be more descriptive for debugging purposes later.

cmd/minikube/cmd/root.go
```diff
@@ -78,8 +77,16 @@ var RootCmd = &cobra.Command{
...
+	}
+	defer func(){
+		err := audit.LogCommandEnd(auditID)
+		if err != nil{
+			klog.Errorf("%v", err)
```
spowelljr, line 87:
```suggestion
			klog.Errorf("failed to log command end to audit: %v", err)
```

pkg/minikube/audit/audit.go
```diff
@@ -51,14 +54,40 @@ func args() string {
 }
 
 // Log details about the executed command.
-func Log(startTime time.Time) {
+func LogCommandStart() (string, error) {
 	if len(os.Args) < 2 || !shouldLog() {
-		return
+		return "", fmt.Errorf("this command should not be logged and len(os.Args) : %v, should be less than 2", len(os.Args))
```
spowelljr, line 59:
```suggestion
		return "", nil
```
We shouldn't be logging an error here

pkg/minikube/audit/audit.go
```diff
@@ -51,14 +54,40 @@ func args() string {
...
 	}
-	r := newRow(os.Args[1], args(), userName(), version.GetVersion(), startTime, time.Now())
+	id := uuid.New().String()
+	r := newRow(os.Args[1], args(), userName(), version.GetVersion(), time.Now(), id)
 	if err := appendToLog(r); err != nil {
-		klog.Warning(err)
+		return "", fmt.Errorf("%v", err)
```
spowelljr, line 64:
```suggestion
		return "", err
```

pkg/minikube/audit/audit.go
```diff
@@ -51,14 +54,40 @@ func args() string {
...
+func LogCommandEnd(id string) error {
+	if currentLogFile == nil {
+		if err := setLogFile(); err != nil {
+			return fmt.Errorf("failed to set the log file: %v", err)
+		}
+	}
+	var auditLogs []byte
+	_, err := currentLogFile.Read(auditLogs)
+	if err != nil {
+		return fmt.Errorf("%v", err)
```
spowelljr, line 78: More descriptive error

pkg/minikube/audit/audit.go
```diff
@@ -51,14 +54,40 @@ func args() string {
...
+	}
+	var rowSlice []row
+	err = json.Unmarshal(auditLogs, &rowSlice)
+	if err != nil {
+		return fmt.Errorf("%v", err)
```
spowelljr, line 83: More descriptive error

pkg/minikube/audit/audit.go
```diff
@@ -51,14 +54,40 @@ func args() string {
...
+	}
+	for _, v := range rowSlice {
+		if v.id == id {
+			v.endTime = time.Now().Format(constants.TimeFormat)
+		}
+	}
```
spowelljr, line 89: We need to write the result back to the file

pkg/minikube/audit/audit.go
```diff
@@ -51,14 +54,40 @@ func args() string {
...
+	return nil
```
spowelljr, line 90: We should return an error if the entry is never found.

pkg/minikube/audit/audit_test.go
```diff
@@ -170,11 +170,25 @@ func TestAudit(t *testing.T) {
...
 		os.Args = []string{"minikube"}
 
-		Log(time.Now())
+		_, err := LogCommandStart()
+		if err != nil {
+			t.Fatal(err)
+		}
+	})
+
+	t.Run("LogCommandEnd", func(t *testing.T) {
```
spowelljr, line 184: We should have one test that's successful and one that fails to find an entry and make sure it returns an error.

spowelljr (changes requested):

NikhilSharmaWe:
@spowelljr 
- Made changes to add the results back to `currentLogFile`.
- But I am not able to understand your point [here](https://github.com/kubernetes/minikube/pull/13307#discussion_r780492155) about the changes in `audit_test.go`, could you please elaborate. I have not worked on test functions in the repo before, so I am not able to understand that what are we trying to achieve and what should we keep in mind while we write these tests.

pkg/minikube/audit/audit_test.go
spowelljr: We want to test all the possible paths so we can be sure that the code is working. So we have two cases in particular we can test.

1. We log a start command, we get back the auditID, and then we log the end command by passing in the auditID, and this case should be successful.

Example:
```
auditID := logStart()
err := logEnd(auditID)
// expect success
```

2. We try writing an end command with a non-existing auditID, this should return an error, so we want to test if it actually does.

Example:
```
err := logEnd(nonExistingID)
// expect error to be returned
```

spowelljr (changes requested):
I added a couple comments, hopefully that helps

NikhilSharmaWe:
@spowelljr I am not very sure about the changes made in `audit_test.go` file, could you please review the changes.

pkg/minikube/audit/audit_test.go
```diff
@@ -170,11 +169,39 @@ func TestAudit(t *testing.T) {
...
+		oldArgs := os.Args
+		defer func() { os.Args = oldArgs }()
+		os.Args = []string{"minikube"}
+
+		auditID, _ := LogCommandStart()
```
spowelljr, line 188: We should still check the error and fatal with a message saying the start failed.

pkg/minikube/audit/audit_test.go
```diff
@@ -170,11 +169,39 @@ func TestAudit(t *testing.T) {
...
 
 	// Check if logging with limited args causes a panic
-	t.Run("Log", func(t *testing.T) {
+	t.Run("LogCommandStart", func(t *testing.T) {
 		oldArgs := os.Args
 		defer func() { os.Args = oldArgs }()
 		os.Args = []string{"minikube"}
 
-		Log(time.Now())
+		_, err := LogCommandStart()
```
spowelljr, line 177: Should also check that the ID being returned is non-empty.

pkg/minikube/audit/audit.go
```diff
@@ -51,14 +54,50 @@ func args() string {
...
+	id := uuid.New().String()
+	r := newRow(os.Args[1], args(), userName(), version.GetVersion(), time.Now(), id)
 	if err := appendToLog(r); err != nil {
-		klog.Warning(err)
+		return "", err
 	}
+	return r.id, nil
+}
+
+func LogCommandEnd(id string) error {
```
spowelljr, line 69: If we're given an empty id we should return without an error.

spowelljr (changes requested):
Looking good, just a few more changes.

pkg/minikube/audit/audit.go
NikhilSharmaWe: @spowelljr should we just `return` or `return nil`. In just `return` it will show an error.

pkg/minikube/audit/row_test.go
```diff
@@ -126,7 +126,7 @@ func TestRow(t *testing.T) {
 			{"user", r.user, u},
 			{"version", r.version, v},
 			{"startTime", r.startTime, stFormatted},
-			{"endTime", r.endTime, etFormatted},
+			{"id", r.id, id},
```
NikhilSharmaWe, line 129: @spowelljr 
Made changes as suggested
What can be the reason for the test fails ? 
Could you please review the code
- I think the changes I made in this (`row_test.go`) file can be the reason for it. Here I replaced `endTime` with `id` because parameter `endTime` was changed in `newRow()` func by `id`.

spowelljr: `TestAudit/LogCommandStart` is failing because you're only passing one arg
```
os.Args = []string{"minikube"}
```
But the func has:
```
if len(os.Args) < 2 || !shouldLog() {
    return "", nil
}
```

Also, you haven't updated `assignFields()` and `toMap()` to use the new `id` field
```
No "id" field below

func (e *row) toMap() map[string]string {
	return map[string]string{
		"args":      e.args,
		"command":   e.command,
		"endTime":   e.endTime,
		"profile":   e.profile,
		"startTime": e.startTime,
		"user":      e.user,
		"version":   e.version,
	}
}
```

NikhilSharmaWe: @spowelljr 
Made changes as suggested
```
- Now `TestAudit/LogCommandStart` is failing and saying that `audit ID should not be empty`. 
- So should I make a condition in `LogCommandStart()` for empty id. I am not very sure about this because in few conditions in the function like

if len(os.Args) == 0 || !shouldLog() {
		return "", nil
}
 we are returning empty string, so how can we deal with it ? 
```

NikhilSharmaWe:
@spowelljr 
Made changes as suggested
```
- Now `TestAudit/LogCommandStart` is failing and saying that `audit ID should not be empty`. 
- So should I make a condition in `LogCommandStart()` for empty id. I am not very sure about this because in few conditions in the function like

if len(os.Args) == 0 || !shouldLog() {
		return "", nil
}
 we are returning empty string, so how can we deal with it ? 
```

pkg/minikube/audit/row_test.go
NikhilSharmaWe: @spowelljr 
Made changes as suggested
```
- Now `TestAudit/LogCommandStart` is failing and saying that `audit ID should not be empty`. 
- So should I make a condition in `LogCommandStart()` for empty id. I am not very sure about this because in few conditions in the function like

if len(os.Args) == 0 || !shouldLog() {
		return "", nil
}
 we are returning empty string, so how can we deal with it ? 
```

NikhilSharmaWe:
@spowelljr Could you please respond to this comment https://github.com/kubernetes/minikube/pull/13307#discussion_r787398455.

pkg/minikube/audit/row_test.go
spowelljr: You need to increase the number of args to 2.
Right now you have:
```
os.Args = []string{"minikube"}
```
Try
```
os.Args = []string{"minikube", "start"}
```

pkg/minikube/audit/row_test.go
```diff
@@ -97,7 +97,7 @@ func TestRow(t *testing.T) {
 	t.Run("toFields", func(t *testing.T) {
 		got := r.toFields()
 		gotString := strings.Join(got, ",")
-		want := []string{c, a, p, u, v, stFormatted, etFormatted}
+		want := []string{c, a, p, u, v, stFormatted, id}
```
NikhilSharmaWe, line 100: @spowelljr Made changes as suggested 
- This test was failing before because it was not expecting `endTime` to be returned but func `toFields()` was returning it, so I removed `endTime` in the function. Do we need the func `toFields()` to also return `endTime` or is it fine?

NikhilSharmaWe:
@spowelljr Could you please respond to this https://github.com/kubernetes/minikube/pull/13307#discussion_r791237744.

pkg/minikube/audit/row_test.go
spowelljr: We need `toFields()` to return `endTime` because that's what we show to the user.

NikhilSharmaWe:
@spowelljr Could you please explain that what can be the reason for failing tests ?

spowelljr:
> @spowelljr Could you please explain that what can be the reason for failing tests ?

`TestRow/toFields` is expecting an `endTime` but it doesn't have one in the test.

NikhilSharmaWe:
@spowelljr 
Fixed the issue with `endTime`, now unit_test is not failing but still multiple different tests are failing. What can be the reason about these failing ?

spowelljr:
> @spowelljr Fixed the issue with `endTime`, now unit_test is not failing but still multiple different tests are failing. What can be the reason about these failing ?

Your changes seemed to corrupt the JSON audit file:
```
=== CONT  TestFunctional/parallel/ConfigCmd
    functional_test.go:1206: expected config error for "./minikube-linux-amd64 -p functional-20220131185208-1818 config unset cpus" to be -""- but got *"E0131 18:58:41.405826    9453 root.go:87] failed to log command end to audit: failed to parse JSON-encoded logs : unexpected end of JSON input"*
    functional_test.go:1195: (dbg) Run:  ./minikube-linux-amd64 -p functional-20220131185208-1818 config get cpus
    functional_test.go:1195: (dbg) Non-zero exit: ./minikube-linux-amd64 -p functional-20220131185208-1818 config get cpus: exit status 14 (67.596874ms)
        
        ** stderr ** 
        	Error: specified key could not be found in config
        
        ** /stderr **
    functional_test.go:1195: (dbg) Run:  ./minikube-linux-amd64 -p functional-20220131185208-1818 config set cpus 2
    functional_test.go:1206: expected config error for "./minikube-linux-amd64 -p functional-20220131185208-1818 config set cpus 2" to be -"! These changes will take effect upon a minikube delete and then a minikube start"- but got *"! These changes will take effect upon a minikube delete and then a minikube start\nE0131 18:58:41.529787    9480 root.go:87] failed to log command end to audit: failed to parse JSON-encoded logs : unexpected end of JSON input"*
```

NikhilSharmaWe:
@spowelljr I am not able to figure out what to do next in this PR so that the JSON do not corrupt.

Need some direction here.

Apologies for not asking specific questions.

NikhilSharmaWe:
@spowelljr I am stuck at this point. I am not able to make out how to prevent JSON file to corrupt.

What could be the reason behind, could you please give me some idea and how should we fix it ?

NikhilSharmaWe:
@spowelljr I am stuck at this point. I am not able to make out how to prevent JSON file to corrupt.

What could be the reason behind, could you please give me some idea and how should we fix it ?

NikhilSharmaWe:
@spowelljr and @medyagh I am not able to make out how to prevent JSON file to corrupt.

I think it can be because of the changes made in test files where I replace endTime with id.
What could be the reason behind, could you please give me some idea and how should we fix it so we can move forward in fixing this issue ?

cmd/minikube/cmd/root.go
```diff
@@ -78,8 +77,16 @@ var RootCmd = &cobra.Command{
 // Execute adds all child commands to the root command sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
-	defer audit.Log(time.Now())
-
+	auditID, err := audit.LogCommandStart()
+	if err != nil {
+		klog.Errorf("failed to log command start to audit: %v", err)
```
klaases, line 82: Is it safe to continue?

Or should we return on error?

klaases, line 82: Is it safe to continue?

Or should we return on error?

cmd/minikube/cmd/root.go
```diff
@@ -78,8 +77,16 @@ var RootCmd = &cobra.Command{
...
+	}
+	defer func() {
+		err := audit.LogCommandEnd(auditID)
+		if err != nil {
+			klog.Errorf("failed to log command end to audit: %v", err)
```
klaases, line 87: Is it safe to continue?

Or should we return on error?

klaases, line 87: Is it safe to continue?

Or should we return on error?

spowelljr:
@NikhilSharmaWe 

I looked into it, so it's not actually that the file is being corrupted, but you're using a byte array of size 0 `var auditLogs []byte` to read the file contents, so it's not actually reading anything from the file because the size is 0. Then when it goes to do an unmarshal it fails due to the array being empty.

NikhilSharmaWe:
@spowelljr So what should be the appropriate value length and capacity for making `auditLogs []byte` ?

NikhilSharmaWe:
@spowelljr 
https://github.com/kubernetes/minikube/blob/6958854288c0aa1a53e603cdb9e58597b7efd0bb/pkg/minikube/audit/audit.go#L78-L82
Added code for making `auditLogs []byte` whose length is equal to the number of bytes `currentLogFile` is consist of, but still the same tests are failing as before.

Could you please review the changes and give your thoughts.

NikhilSharmaWe:
@spowelljr Could you please give your input on my https://github.com/kubernetes/minikube/pull/13307#issuecomment-1073160089 comment.

NikhilSharmaWe:
@spowelljr Could you please give your input on my [previous](https://github.com/kubernetes/minikube/pull/13307#issuecomment-1073160089) comment.

spowelljr:
There are two problems:
1. Because we're using the same file reference to read and write to the file, the pointer is set to the end of the file after `LogCommandStart` completes, so we will need to `Seek` back to the start of the file before we can read from the beginning of the file again. `_, err = currentLogFile.Seek(0, 0)`

2. You're trying to convert the entire audit file into a JSON object, but each line is an individual JSON object.

The second issue (and maybe the first as well) can be solved by iterating over the audit entries using the same format I use when outputting the audit logs. https://github.com/kubernetes/minikube/blob/538f5fc3c9fa73e25e0f4a90df53d3e332f436e6/pkg/minikube/audit/report.go#L41-L48

NikhilSharmaWe:
@spowelljr Could you please review the changes. Before I did not add currentLogFile.Seek(0,  0) and test failed. So I added this statement. 

Did not added https://github.com/kubernetes/minikube/blob/538f5fc3c9fa73e25e0f4a90df53d3e332f436e6/pkg/minikube/audit/report.go#L44-L46
condition because I am not able to properly understand it also don't have a variable like `lastNLines` is present.

Could you please make understand about what is wrong with my approach ?

spowelljr:
```
 if len(logs) == lastNLines { 
 	logs = logs[1:] 
 } 
```
That was specific to the function because the user could pass a flag to indicate how many lines of logs they wanted. That would just make sure the last n number of number of lines were included, you can ignore it for this case.

NikhilSharmaWe:
@spowelljr Could you please explain what is wrong in my approach in func `LogCommandEnd` regarding the making of rowSlice.

NikhilSharmaWe:
@spowelljr Could you please explain what is wrong in my approach in func `LogCommandEnd` regarding the making of `rowSlice`.

NikhilSharmaWe:
@spowelljr Could you please explain what is wrong in my approach in func `LogCommandEnd` regarding the making of `rowSlice`.

NikhilSharmaWe:
@spowelljr and @medyagh Could you please explain what is wrong with my approach in func `LogCommandEnd` regarding the making of `rowSlice`.

spowelljr:
@NikhilSharmaWe The problem is you're writing back the logs in a different format.

Assuming the `audit.json` file starts like this:
```
{"specversion":"1.0","id":"42267f69-a9aa-4a0c-9e54-4352168a9e94","source":"https://minikube.sigs.k8s.io/","type":"io.k8s.sigs.minikube.audit","datacontenttype":"application/json","data":{"args":"--help","command":"start","endTime":"Fri, 22 Apr 2022 13:06:06 PDT","profile":"minikube","startTime":"Fri, 22 Apr 2022 13:06:06 PDT","user":"powellsteven","version":"v1.25.2"}}
{"specversion":"1.0","id":"6aae027d-edb4-4641-981b-d27472613548","source":"https://minikube.sigs.k8s.io/","type":"io.k8s.sigs.minikube.audit","datacontenttype":"application/json","data":{"args":"--all","command":"delete","endTime":"Fri, 22 Apr 2022 14:17:25 PDT","profile":"minikube","startTime":"Fri, 22 Apr 2022 14:17:24 PDT","user":"powellsteven","version":"v1.25.2"}}
```

But you're writing back:
```
[
    {"data":{"args":"--help","command":"start","endTime":"Fri, 22 Apr 2022 13:06:06 PDT","profile":"minikube","startTime":"Fri, 22 Apr 2022 13:06:06 PDT","user":"powellsteven","version":"v1.25.2"}},
    {"data":{"args":"--all","command":"delete","endTime":"Fri, 22 Apr 2022 14:17:25 PDT","profile":"minikube","startTime":"Fri, 22 Apr 2022 14:17:24 PDT","user":"powellsteven","version":"v1.25.2"}}
]
```

The file is initially a single JSON object per line, but you're writing back an array of JSON objects. This format mismatch is causing the JSON marshaling to fail the next time it runs.

After taking a look at the code this should be fairly easy to resolve, but there are also two bugs causing issues as well.

Currently you have:
```
	for _, v := range rowSlice {
		if v.id == id {
			v.endTime = time.Now().Format(constants.TimeFormat)
			rowSlice = append(rowSlice, v)
			auditLogs, err := json.Marshal(rowSlice)
			if err != nil {
				return fmt.Errorf("failed to return JSON-encoding of slice of type row : %v", err)
			}
			_, err = currentLogFile.Write(auditLogs)
			if err != nil {
				return fmt.Errorf("failed to write to current log file : %v", err)
			}
			return nil
		}
	}
```

**Bug 1.**
You're not actually modifying an existing entry, you're pushing an additional modified entry, resulting in duplicate entries, shown here:
```
v.endTime = time.Now().Format(constants.TimeFormat)
rowSlice = append(rowSlice, v)
```

Example of what this looks like:
```
Before command:
Log 1
Log 2

After command:
Log 1
Log 2
Log 3 without date
Log 3 with date
```

**Bug 2.**
You're setting the end date to the struct, but it's not actually making its way to the JSON struct that's being output.

This is the struct you're adding end date to:
```
type row struct {
        args      string
        command   string
        endTime   string
        profile   string
        startTime string
        user      string
        version   string
        id        string
        Data      map[string]string `json:"data"`
}
```

You can see `endTime` doesn't actually have a `json` tag so it's not going to make its way to the JSON output. We already have a func `assignFields()`:
```
func (e *row) assignFields() {
        e.args = e.Data["args"]
        e.command = e.Data["command"]
        e.endTime = e.Data["endTime"]
        e.profile = e.Data["profile"]
        e.startTime = e.Data["startTime"]
        e.user = e.Data["user"]
        e.version = e.Data["version"]
        e.id = e.Data["id"]
}
```

So all that needs to be done to resolve this bug is to call this func before marshaling the struct to JSON, which will ensure the `endTime` value gets added to the `Data` map which is what's actually marshaled to JSON.

**Core issue:**
My idea for solving this issue is to maintain the for loop you have, and for every entry, marshal the entry to JSON, and append it to a string with a newline, and once you've iterated over all the entries overwrite the `audit.json` file with the string you're been appending to. When you're looping over the entries, if you find the one you're looking to append the `endTime` to, add the field then carry on with the same as the other steps.

Example:
```
auditContents := ""
for _, v := range rowSlice {
    if v.id == id {
        v.endTime = time.Now().Format(constants.TimeFormat)
        v.assignFields()
    }
    auditLog, err := json.Marshal(v)
    if err != nil {
        return fmt.Errorf(...)
    }
    auditContents += auditLog + "\n"
}
// overwrite audit.json with aduitContents
```

NikhilSharmaWe:
@spowelljr Made changes as suggested.
Could you please take a look.

pkg/minikube/audit/audit.go
```diff
@@ -52,14 +56,69 @@ func args() string {
...
+	rowSlice, err := logsToRows(logs)
+	if err != nil {
+		return fmt.Errorf("failed to convert logs to rows: %v", err)
+	}
+	auditContents := ""
+	var entriesNeedsToUpdate int
+	for _, v := range rowSlice {
+		if v.id == id {
+			v.endTime = time.Now().Format(constants.TimeFormat)
+			v.assignFields()
```
spowelljr, line 101: Sorry, I made a mistake when I was referencing `assignFields`. `assignFields` actually does the opposite, it assigns the `Data` values to the other fields. We instead want to do:
```
v.Data = v.toMap()
```

spowelljr (changes requested):

spowelljr:
@NikhilSharmaWe It seems you're not able to run the truncate command
```
failed to log command end to audit: truncate C:\\Users\\jenkins.minikube8\\minikube-integration\\.minikube\\logs\\audit.json: Access is denied."
```

spowelljr:
The audit logging is failing because you added os.O_TRUNC and so by default every write is truncating the file.
Example:
$ minikube start

audit file:
// start command has started but hasn't finished yet
{"command": "start", "id": "123", "startTime": "12:00", "endTime": "" ...}

// in another terminal window
$ minikube stop

audit file:
// overwrites existing contents
{"command": "stop", "id": "456", "startTime": "12:01", "endTime": "12:01" ...}

// start completes but can't find `id: 123` so it throws an error

Also, `failed to log command end to audit: truncate C:\\Users\\jenkins.minikube8\\minikube-integration\\.minikube\\logs\\audit.json: Access is denied.` is occurring specifically on Windows as Windows generally locks files when writing to them. So you should add mutex to the start and end logging files so only one write can occur at a time and these issues should be resolved.

cmd/minikube/cmd/root.go
```diff
@@ -84,8 +84,19 @@ var RootCmd = &cobra.Command{
 // Execute adds all child commands to the root command sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
-	defer audit.Log(time.Now())
-
+	var mutex sync.Mutex
```
spowelljr, line 87: I'd rather the `mutex` variable goes inside the `audit` package.

cmd/minikube/cmd/root.go
```diff
@@ -84,8 +84,19 @@ var RootCmd = &cobra.Command{
...
+	mutex.Lock()
+	auditID, err := audit.LogCommandStart()
+	mutex.Unlock()
+	if err != nil {
+		klog.Errorf("failed to log command start to audit: %v", err)
+	}
+	defer func() {
+		err := audit.LogCommandEnd(auditID)
```
spowelljr, line 95: `LogCommandEnd` needs to be mutexed as well

spowelljr (changes requested):

NikhilSharmaWe:
@spowelljr Made changes as suggested.
Could you please review.

cmd/minikube/cmd/root.go
```diff
@@ -84,8 +83,20 @@ var RootCmd = &cobra.Command{
 // Execute adds all child commands to the root command sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
-	defer audit.Log(time.Now())
-
+	audit.Mutex.Lock()
```
spowelljr, line 86: We should do this locking inside `LogCommandStart` and `LogCommandEnd`, that way if we call it somewhere else we don't have to repeat this locking.

spowelljr (changes requested):

spowelljr:
Windows is having `failed to log command end to audit: truncate C:\\Users\\jenkins.minikube8\\minikube-integration\\.minikube\\logs\\audit.json: Access is denied.` issues again since truncate was removed.

There are two possible solutions I thought of:
1. The file is opened and closed between every operation, and when we open the file to write the result in `LogCommandEnd` it's opened with `os.O_TRUNC`.

2. We seek back to the beginning of the file before writing the result in `LogCommandEnd`, I believe this should allow us to overwrite the existing contents on Windows but haven't tried it myself.

NikhilSharmaWe:
@spowelljr We seek back to the beginning of the file here in `LogCommandEnd` already. https://github.com/kubernetes/minikube/blob/d3235e09ee05b723c934c0ea4759925a8f8a66c9/pkg/minikube/audit/audit.go#L87

Is there something I am missing here?

pkg/minikube/audit/audit_test.go
```diff
@@ -188,8 +207,32 @@ func TestAudit(t *testing.T) {
...
+		}
+		err = LogCommandEnd(auditID)
 
-		Log(time.Now())
+		if err != nil {
+			t.Fatal(err)
+		}
+	})
+
+	t.Run("LogCommandEnd", func(t *testing.T) {
```
sharifelgamal, line 221: this test should have a different name than the one above it

spowelljr: Good catch, renamed

sharifelgamal (approved):

spowelljr:
The problem is we were leaving the audit log file for the entirety of the command run. The issue is Windows can only truncate a file when it's closed, so if you have more than one command running at the same time (always happening during testing) if one command completes it tried to truncate the log but the other command has it open so the truncation fails. Refactored the code so we only have the audit log open for the minimum required amount of time required before closing it again.

sharifelgamal (approved):

PR #13309, opened 2022-01-06
Make warning for running amd64 binary on M1 more visible

sharifelgamal:
Before:
```
$ ./minikube-darwin-amd64 start
    ▪ You are trying to run amd64 binary on M1 system. Please consider running darwin/arm64 binary instead (Download at https://github.com/kubernetes/minikube/releases/download/v1.24.0/minikube-darwin-arm64.)
😄  minikube v1.24.0 on Darwin 12.0.1
...
```

After:
```
$ ./minikube-darwin-amd64 start 
╭────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                            │
│    You are trying to run the amd64 binary on an M1 system.                                                 │
│    Please consider running the darwin/arm64 binary instead.                                                │
│    Download at https://github.com/kubernetes/minikube/releases/download/v1.24.0/minikube-darwin-arm64      │
│                                                                                                            │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
😄  minikube v1.24.0 on Darwin 12.0.1
...
```

### DIFF
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -95,7 +95,7 @@ func Execute() {
 	}
 
 	if runtime.GOOS == "darwin" && detect.IsAmd64M1Emulation() {
-		out.Boxed("You are trying to run amd64 binary on M1 system.\nPlease consider running the darwin/arm64 binary instead.\nDownload at {{.url}}",
+		out.Boxed("You are trying to run the amd64 binary on an M1 system.\nPlease consider running the darwin/arm64 binary instead.\nDownload at {{.url}}",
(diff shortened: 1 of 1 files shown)


spowelljr (approved):

afbjorklund:
Wonder why this was not needed for the Linux users ?

sharifelgamal:
Yeah, you make a good point, this should be a generic architecture incompatibility warning. This PR was purely cosmetic, so it didn't even occur to me to check the condition.

afbjorklund:
Once we have bettter control over the cluster arch, we can come back to this. We want to avoid accidents, but still allow for more advanced use (and the "M1" is not important)

PR #13313, opened 2022-01-07
Update time-to-k8s-repo

VRECO:


spowelljr:
Hi @VRECO, it seems our time-to-k8s submodule is already at the latest version.

If you look at https://github.com/kubernetes/minikube/tree/master/hack/benchmark/time-to-k8s you can see the module is already pointing to `6e61334` which is the latest commit.

And I just checked out the minikube repo fresh and checked out submodules and it pulled the latest commit as well.

PR #13328, opened 2022-01-11
Remove deprecated global-housekeeping-interval flag

spowelljr:
kubelet v1.23: `--global-housekeeping-interval`: `Interval between global housekeepings (default 1m0s) (DEPRECATED: This is a cadvisor flag that was mistakenly registered with the Kubelet. Due to legacy concerns, it will follow the standard CLI deprecation timeline before being removed.)`


klaases (approved):

medyagh:
ok-to-test

PR #13178, opened 2021-12-16
WIP: Configure minikube ISO for aarch64 (ARM64)

klaases:
_Work in progress_

How to test architecture on Linux: run `uname -m`.

For more information: `uname -a`.

I am still in the testing and debugging phase of creating the minikube ISO with aarch64 support for Linux.

PR #13331, opened 2022-01-12
ISO: Upgrade falco-module to version 0.28.0

afbjorklund:


deploy/iso/minikube-iso/package/falco-module/falco-module.mk
```diff
@@ -4,15 +4,15 @@
 #
 ########################################################################
 
-FALCO_MODULE_VERSION = 0.24.0
+FALCO_MODULE_VERSION = 0.28.0
```
medyagh, line 7: why not all the way 0.30.0 ?

afbjorklund: there was some internal changes in 0.30, that causes it to take longer time

https://falco.org/blog/contribution-drivers-kmod-ebpf-libraries/

medyagh: makes sense thanks for explaining

medyagh:
ok-to-build-iso

deploy/iso/minikube-iso/package/falco-module/falco-module.mk
klaases: Yes, we were working on this last evening. If I recall correctly, Anders mentioned that Falco made some more significant changes going to 0.30 than 0.28.

It looks like Falco broke out from sysdig.

Our plan was to see if 0.28 builds, and potentially update to 0.30 later. Also, the last time Falco was updated in Minikube was Aug 25, 2020.

klaases:
Looking forward - this may resolve the build errors seen recently with sysdig 0.23 when building an ARM64 ISO.

https://github.com/kubernetes/minikube/pull/13178

deploy/iso/minikube-iso/package/falco-module/falco-module.mk
afbjorklund: It might not be more than a rename of a variable, but more testing is needed

afbjorklund:
@klaases should we do something more with this ? maybe upgrade to 0.31.0 as well ?

https://github.com/falcosecurity/falco/releases

PR #13333, opened 2022-01-12
WIP: Add new external-dns addon for exposing Ingress, LoadBalancer & Istio external IPs to the host

### LINKED ISSUE #8980: Add addon for external DNS

denniseffing:
Hi!

I created a new and refined version iterated over the initial draft in #9000. Special thanks to @segevfiner for figuring out the biggest issues! Made my PR a whole lot easier.

This addon deploys a bind named DNS server and an [ExternalDNS](https://github.com/kubernetes-sigs/external-dns) instance. ExternalDNS observes `Ingress`, `Service`, Istio `Gateway` and `VirtualService` resources and configures the DNS server accordingly. The DNS zone can be configured using `minikube addons configure external-dns` and is set to `demo.` by default. This means that by default, you can only configure hostnames ending in `.demo`.

The DNS server is exposed to the host network and shall be used as DNS resolver for the host machine.
This way, ingresses and services of type `LoadBalancer` are resolvable by the host.

I also added a handbook page for the addon based on the handbook page of the ingress-dns addon. Since both addons utilize DNS servers, the instructions are quite the same and shamelessly reused, if still applicable. Please let me know if that is okay.

I included examples as well. Please let me know if you run into any issues. Thanks!

## How to test with provided example

```
$ ./out/minikube start

$ ./out/minikube addons enable ingress

$ ./out/minikube addons enable external-dns

$ kubectl apply -f deploy/addons/external-dns/example/example.yaml
```

Add the `minikube ip` as a DNS server as shown here: https://minikube.sigs.k8s.io/docs/handbook/addons/ingress-dns/

```
$ curl http://hello-john.demo

Hello, world!
Version: 1.0.0
Hostname: hello-world-app-86d5b6469f-wndph

$ curl http://hello-jane.demo

Hello, world!
Version: 1.0.0
Hostname: hello-world-app-86d5b6469f-wndph
```

## TODO

- [x] Ingress Hostnames currently resolve to localhost instead of the minikube IP because the ingress addon starts the nginx ingress-controller using the argument `--publish-status-address=localhost`. We should consider removing the argument from the ingress addon. I'm not quite sure why the argument is needed in the first place, everything works perfectly without the argument
- [ ] The internal CoreDNS server should forward to the bind DNS server for the configured DNS zone to allow resolving hostnames internally as well. Steps on how to achieve this manually is already documented on the ingress-dns handbook and external-dns handbook page but I think it would be the best experience, if we could automate this using the external-dns plugin. I don't really know how to patch existing deployments using minikube addons though, so some pointers on how to achieve this would be much appreciated!
- [x] The ExternalDNS pod goes into a `CrashLoopBackoff` when the Istio CRDs are not installed on the cluster. This seems to be an issue with ExternalDNS itself, see https://github.com/kubernetes-sigs/external-dns/issues/2490. I guess I will work around this issue by introducing a configurable setting for Istio support. This setting can be removed in the future when ExternalDNS fixed the issue.

Fixes #8980


denniseffing:
I am currently ironing out the details regarding the CLA internally. I expect that I can finally sign the CLA sometime next week.

denniseffing:
I rebased the branch, signed the CLA and fixed another bug with the custom DNS zone configuration.
The issue due to the `--publish-status-address=localhost` parameter of the nginx ingress controller configured by the ingress addon has been resolved with commit 2be6b99d79651f6353cdd1543684bf868bea904b. 

I still need some help regarding the CoreDNS configuration of the k8s cluster itself. Is modifying the CoreDNS ConfigMap and restarting the pod even possible with minikube addons?

A review and feedback by admins or someone else would be much appreciated!

PR #13353, opened 2022-01-15
Change file path for preload tar files to be organized by version

spowelljr:
Currently all the preloads are stored at the root level of the bucket, it would be easier to navigate with a folder structure.

Now the structure is `<preload-version>/<kubernetes-version>/<filename>`


medyagh (approved):

medyagh (changes requested):
we should ensure to bump the preload version and then also Run the preload job after this get merged

medyagh (approved):

PR #13354, opened 2022-01-15
Add exit message for too new Kubernetes version

### LINKED ISSUE #11627: improve exit message when trying to pull Non-Existing Kubernetes Version

nishipy:
Fix #11627 in the simplest way.

If specified Kubernetes version is more than the newest supported one, the output will be like:
```
# minikube  start --kubernetes-version=1.23.3
😄  minikube v1.24.0 on Redhat 8.5
❗  Specified Kubernetes version 1.23.3 is more than the newest supported version: v1.23.2-rc.0
❗  You can force an unsupported Kubernetes version via the --force flag if it exists

❌  Exiting due to K8S_OLD_UNSUPPORTED: Kubernetes 1.23.3 is not supported by this release of minikube
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1533,19 +1533,22 @@ func validateKubernetesVersion(old *config.ClusterConfig) {
 		klog.Infof("No Kuberentes version set for minikube, setting Kubernetes version to %s", constants.NoKubernetesVersion)
 		return
 	}
+	if nvs.Major > newestVersion.Major {
+		out.WarningT("Specified Major version of Kubernetes {{.specifiedMajor}} is newer than the newest supported Major version: {{.newestMajor}}", out.V{"specifiedMajor": nvs.Major, "newestMajor": newestVersion.Major})
+		if !viper.GetBool(force) {
+			out.WarningT("You can force an unsupported Kubernetes version via the --force flag")
+		}
+		exitIfNotForced(reason.KubernetesTooNew, "Kubernetes {{.version}} is not supported by this release of minikube", out.V{"version": nvs})
+	}
+	if nvs.GT(newestVersion) {
+		out.WarningT("Specified Kubernetes version {{.specified}} is newer than the newest supported version: {{.newest}}", out.V{"specified": nvs, "newest": constants.NewestKubernetesVersion})
+	}
 	if nvs.LT(oldestVersion) {
 		out.WarningT("Specified Kubernetes version {{.specified}} is less than the oldest supported version: {{.oldest}}", out.V{"specified": nvs, "oldest": constants.OldestKubernetesVersion})
 		if !viper.GetBool(force) {
 			out.WarningT("You can force an unsupported Kubernetes version via the --force flag")
 		}
 		exitIfNotForced(reason.KubernetesTooOld, "Kubernetes {{.version}} is not supported by this release of minikube", out.V{"version": nvs})
-	}
-	if nvs.GT(newestVersion) {
-		out.WarningT("Specified Kubernetes version {{.specified}} is newer than the newest supported version: {{.newest}}", out.V{"specified": nvs, "newest": constants.NewestKubernetesVersion})
-		if !viper.GetBool(force) {
-			out.WarningT("You can force an unsupported Kubernetes version via the --force flag if it exists")
-		}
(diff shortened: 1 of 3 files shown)


klaases:
Hi @sharifelgamal, do you think this PR is `/ok-to-test`?

cmd/minikube/cmd/start.go
```diff
@@ -1539,6 +1540,13 @@ func validateKubernetesVersion(old *config.ClusterConfig) {
 		}
 		exitIfNotForced(reason.KubernetesTooOld, "Kubernetes {{.version}} is not supported by this release of minikube", out.V{"version": nvs})
 	}
+	if nvs.GT(newestVersion) {
+		out.WarningT("Specified Kubernetes version {{.specified}} is more than the newest supported version: {{.newest}}", out.V{"specified": nvs, "newest": constants.NewestKubernetesVersion})
```
sharifelgamal, line 1544:
```suggestion
		out.WarningT("Specified Kubernetes version {{.specified}} is newer than the newest supported version: {{.newest}}", out.V{"specified": nvs, "newest": constants.NewestKubernetesVersion})
```

sharifelgamal:
Hey @nishipy, thanks for this PR. Could you add an integration test to make sure this codepath stays covered?

cmd/minikube/cmd/start.go
medyagh: @nishipy I think the goal of this was to prevent accidental Large versions such as 2.5.0
so we should still allow users to try patch versions which are newer,

so maybe we should have a constant in our Code 

Max major version allowed and that would be 1

so if user types --kubernetes-version=2.10.0 it will fail without trying

medyagh (changes requested):

cmd/minikube/cmd/start.go
nishipy: @medyagh Thanks for your comment. I understood our goal.

I think we have two ways to specify the Max major version allowed:
1. define a new constant like `MaxKubernetesMajorVersion = 1`
2. use existing constant (e.g. `NewestKubernetesVersion = "v1.23.2-rc.0"`) and trim the Major version from it. then, we use it as the Max major version allowed.

I think 1st one is better but we have to manage the new constant. On the other hand, 2nd one needs no new constant. What do you think?

nishipy:
> Hey @nishipy, thanks for this PR. Could you add an integration test to make sure this codepath stays covered?

Hi @sharifelgamal, thanks for your comment.
Do you mean we will add the test code to `test/integration/`?
If my understanding is correct, I think it will be similar to [`test/integration/no_kubernetes_test.go`](https://github.com/kubernetes/minikube/blob/master/test/integration/no_kubernetes_test.go).

cmd/minikube/cmd/start.go
medyagh: number 2 seems like a good idea

medyagh: I like the idea of getting the Newest Major version from our NewestKubernetesVersion

nishipy: Thank you @medyagh.
I implemented this flow as following commit: 
9da7049f8ac7b3d33ca6b699da2c99d4f6066653

klaases:
Hi @nishipy, were you able to make the requested changes?

medyagh (approved):

medyagh (approved):
looks good to me , if @sharifelgamal is good with it too, lets merge it

sharifelgamal (approved):

PR #13412, opened 2022-01-20
Update grub.cfg working directory for UEFI on Hyperkit

klaases:


klaases:
ok-to-build-iso

klaases:
ok-to-build-iso

klaases:
ok-to-build-iso

klaases:
ok-to-build-iso

klaases:
ok-to-build-iso

klaases:
Closing PR as this was based on `origin/arm1` branch instead of master.

Replaced with # 13416

PR #13123, opened 2021-12-07
Bootloader ISO EFI Support

klaases:
Fixes: #13122

This PR is an important prerequisite for building minikube ISO on ARM64.

Note: the `containerd` version has been lowered 1.4.9, as 1.5.8 has a rotating hash issue that prevents the ISO from building properly. For more information, see: https://github.com/kubernetes/minikube/issues/13193 

Note: make sure to merge #12892 before submission. [CONFIRMED]

---

Build the ISO, see https://minikube.sigs.k8s.io/docs/contrib/building/iso/

Use `mk ssh` to ssh into minikube ISO.

Run `[ -d /sys/firmware/efi ] && echo UEFI || echo BIOS`

**Before:**

`BIOS`

**After:**

`UEFI`

---

Have tested the new ISO with the following commands.

Download the latest ISO, and place in `minikube/out` folder.
`curl -LO 
https://storage.googleapis.com/minikube-builds/iso/13123/minikube-v1.24.0-1640194376-13123.iso`

Alternative URL:
`gs://minikube-builds/iso/13123/minikube-v1.24.0-1640194376-13123.iso`

Run `make` from minikube home folder.

Run `./out/minikube start --iso-url=file://$(pwd)/out/minikube-v1.24.0-1640194376-13123.iso` to start minikube with the new ISO.

Use `mk ssh` to ssh into minikube ISO.

Run `[ -d /sys/firmware/efi ] && echo UEFI || echo BIOS` to confirm that minikube is running with the UEFI bootloader.

PR #13414, opened 2022-01-20
Revert bootloader ISO EFI Support PR

spowelljr:
Reverting https://github.com/kubernetes/minikube/pull/13123 due to HyperKit driver issue.


spowelljr:
ok-to-build-iso

PR #13416, opened 2022-01-20
Update grub.cfg working directory for UEFI on Hyperkit

klaases:
Replaces: #13412 


klaases:
Replaced with #13417

Issue #13413, opened 2022-01-20
v1.25.0 Hyperkit missing `/grub/grub.cfg`


klaases:
### What Happened?

When running v1.25.0 on Mac using Hyperkit, the following file is missing `/grub/grub.cfg` from the Buildroot configuration.

### Attach the log file

```
😿  Failed to start hyperkit VM. Running "minikube delete" may fix it: creating host: create: Error creating machine: Error in driver during machine creation: extracting kernel: unable to find file /grub/grub.cfg

❌  Exiting due to GUEST_PROVISION: Failed to start host: creating host: create: Error creating machine: Error in driver during machine creation: extracting kernel: unable to find file /grub/grub.cfg
```

### Operating System

macOS (Default)

### Driver

HyperKit

th3fallen:
same here

afbjorklund:
The file looks unused, both isolinux.cfg (and grub.cfg)

klaases:
Hi @afbjorklund, yes, this line was removed. However, we will still need the grub.cfg file to exist somewhere in the ISO in order to boot into UEFI.

Closing this out as the issue was resolved in release v1.25.1.

PR #13417, opened 2022-01-20
WIP: Update grub.cfg working directory for UEFI on Hyperkit

klaases:
Fixes: #13413

Replaces previous PR: #13412


klaases:
ok-to-build-iso

klaases:
Closing as this will be included in a separate PR.

PR #13419, opened 2022-01-21
Fix bug in AUX Driver Download URL for ARM64

klaases:
A bug exists in the way that the downloads are made for ARM64.

Without the architecture, the default works fine.

**Before:**
`https://github.com/kubernetes/minikube/releases/download/v1.24.0-arm64/docker-machine-driver-kvm2`

**After:**
`https://github.com/kubernetes/minikube/releases/download/v1.24.0/docker-machine-driver-kvm2-arm64`

### DIFF
--- a/pkg/minikube/download/driver.go
+++ b/pkg/minikube/download/driver.go
@@ -33,7 +33,7 @@ func driverWithChecksumURL(name string, v semver.Version) string {
 	return fmt.Sprintf("%s?checksum=file:%s.sha256", base, base)
 }
 func driverWithArchAndChecksumURL(name string, v semver.Version) string {
-	base := fmt.Sprintf("https://github.com/kubernetes/minikube/releases/download/v%s-%s/%s", v, runtime.GOARCH, name)
+	base := fmt.Sprintf("https://github.com/kubernetes/minikube/releases/download/v%s/%s-%s", v, name, runtime.GOARCH)
 	return fmt.Sprintf("%s?checksum=file:%s.sha256", base, base)
 }
 


afbjorklund:
Seems that there is a stray duplicate driver, should be deleted.

`docker-machine-driver-kvm2-x86_64`

afbjorklund:
Looks like another file is missing instead:

`docker-machine-driver-kvm2-1.25.1-0.aarch64.rpm`

klaases:
Is the stray `docker-machine-driver-kvm2-x86_64` driver a part of the `makefile`?

```
out/docker-machine-driver-kvm2: out/docker-machine-driver-kvm2-$(GOARCH)
	$(if $(quiet),@echo "  CP       $@")
	$(Q)cp $< $@

out/docker-machine-driver-kvm2-x86_64: out/docker-machine-driver-kvm2-amd64
	$(if $(quiet),@echo "  CP       $@")
	$(Q)cp $< $@

out/docker-machine-driver-kvm2-aarch64: out/docker-machine-driver-kvm2-arm64
	$(if $(quiet),@echo "  CP       $@")
	$(Q)cp $< $@
```
---

For this file `docker-machine-driver-kvm2-1.25.1-0.aarch64.rpm`, do we want to download this in addition?

afbjorklund:
We "need" the binary for making the RPM files, but it's not needed as a separate download

klaases:
After giving this some thought, I think that the "stray" driver `docker-machine-driver-kvm2-x86_64` is actually a byproduct of the Driver.go file that falls back to a "common version" of the driver.

So long as the architecture is specified, we should not see the duplicate driver being downloaded.

Or, where else would changes be needed to make sure only one driver is downloaded?

```
	archURL := driverWithArchAndChecksumURL(name, v)
	if err := download(archURL, destination); err != nil {
		klog.Infof("failed to download arch specific driver: %v. trying to get the common version", err)
		if err := download(driverWithChecksumURL(name, v), destination); err != nil {
			return errors.Wrap(err, "download")
		}
	}
```

PR #13438, opened 2022-01-22
Add support for iscsid in minikube

589290:
This is a rebase of: 
https://github.com/kubernetes/minikube/pull/11458

Original description:
There is no support currently for using longhorn (or any storage implementation that requires open-iscsi) using minikube on VirtualBox. There are solutions when we use the docker driver (#2846, #1174) which also require to install it after minikube is started, but no out of the box solution available.

There was a PR (#1176) started 3 years ago but it was incomplete and closed. I took that as a start to include open-iscsi and its related open-isns to bake this into a buildroot image so its avaiable as part of minikube.

This patch builds by adding 2 packages (open-iscsi , which depends on open-isns) and adding the required mounts that are required to persist the volumes that Longhorn creates in /var/lib/longhorn. The related kernel modules have been added as well.

Testing Procedure:

Build the ISO with this patch
Create a minikube install using this iso
minikube start -p test --cpus=2 --memory=4096 --driver=virtualbox --iso-url=file:///tmp/mk2.iso --kubernetes-version=1.19.4
Install LongHorn using the instructions at https://longhorn.io/docs/1.1.1/deploy/install/install-with-kubectl/
Create a PVC and a deployment that uses this. Scale up/down to make sure that the PVC is being used properly
Shut down minikube
Start it back up and make sure that the data is still available at the pod using the PVC.


589290:
I signed it

PR #13302, opened 2022-01-05
KIC: Upgrade cri-dockerd

afbjorklund:
Upgrade to v0.2.0, but we still might need a fix for the socket path (maybe in 0.2.1?)

* https://github.com/Mirantis/cri-dockerd/issues/27

PR #13488, opened 2022-01-25
fix external cri-docker socket name

sharifelgamal:
This fixes minikube with kubernetes v1.24.0-alpha.1 and is verified to work with current stable, v1.23.2

### DIFF
--- a/pkg/minikube/cruntime/docker.go
+++ b/pkg/minikube/cruntime/docker.go
@@ -43,7 +43,7 @@ import (
 const KubernetesContainerPrefix = "k8s_"
 
 const InternalDockerCRISocket = "/var/run/dockershim.sock"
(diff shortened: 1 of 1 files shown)


afbjorklund (changes requested):
The location of the cri-dockerd socket was _changed_, in the cri-dockerd 0.2.0 release.

Now the systemd unit is called "cri-docker.socket", but the file is "cri-dockerd.sock".

afbjorklund:
Use this PR instead:

* https://github.com/kubernetes/minikube/pull/13302

sharifelgamal:
ok, I'll close this and we'll upgrade the version of cri-dockerd instead.

afbjorklund:
Hopefully they don't change their mind again, so this should be it

afbjorklund:
* https://github.com/Mirantis/cri-dockerd/pull/31

PR #13501, opened 2022-01-27
docs: Discourage pasting secrets into docs

alexec:


alexec:
I had a load of secrets, they were buried in the logs. They ended up in your issue. I had to get them all revoked. Annoying.

PR #13508, opened 2022-01-28
added condition for the case when user try to enable addons without kubernetes

### LINKED ISSUE #13487: Output message if user tries to enable addons with `--no-kubernetes`
It's not possible to enable addons with `--no-kubernetes` passed because the addons in their current state are deployed to Kubernetes. So if a user tries to enable an addon with `--no-kubernetes` output a message notifying them that it's not possible without Kubernetes and exit execution.

Related https://github.com/kubernetes/minikube/issues/13461

NikhilSharmaWe:
Fixes #13487

- When user try to enable addons with `--no-kubernetes=true` an output should be logged informing that it is not possible to enable addons with `--no-kubernetes` passed. This PR works on logging that output.

### DIFF
--- a/cmd/minikube/cmd/config/enable.go
+++ b/cmd/minikube/cmd/config/enable.go
@@ -40,6 +40,11 @@ var addonsEnableCmd = &cobra.Command{
 		if len(args) != 1 {
 			exit.Message(reason.Usage, "usage: minikube addons enable ADDON_NAME")
 		}
+		if ClusterConfig, err := config.Load(ClusterFlagValue()); err == nil {
+			if ClusterConfig.KubernetesConfig.KubernetesVersion == constants.NoKubernetesVersion {
+				exit.Message(reason.Usage, "Cannot enable addons with --no-kubernetes passed")
+			}
+		}
 		addon := args[0]
 		// replace heapster as metrics-server because heapster is deprecated
 		if addon == "heapster" {
@@ -62,7 +67,7 @@ var addonsEnableCmd = &cobra.Command{
 			}
 			out.Styled(style.Tip, `Some dashboard features require the metrics-server addon. To enable all features please run:
 
-	minikube{{.profileArg}} addons enable metrics-server	
+	minikube{{.profileArg}} addons enable metrics-server
 
 `, out.V{"profileArg": tipProfileArg})
 
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -286,6 +286,7 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 
 	validateFlags(cmd, driverName)
 	validateUser(driverName)
+	validateNoAddonsForNoKubernetes()
 	if driverName == oci.Docker {
 		validateDockerStorageDriver(driverName)
 	}
@@ -1651,4 +1652,12 @@ func exitGuestProvision(err error) {
 		exit.Message(reason.GuestProvisionContainerExited, "Docker container exited prematurely after it was created, consider investigating Docker's performance/health.")
 	}
 	exit.Error(reason.GuestProvision, "error provisioning guest", err)
+}
+
+func validateNoAddonsForNoKubernetes() {
+	if ClusterConfig, err := config.Load(ClusterFlagValue()); err == nil {
+		if ClusterConfig.KubernetesConfig.KubernetesVersion == constants.NoKubernetesVersion {
+			exit.Message(reason.Usage, "Cannot enable addons with --no-kubernetes passed")
+		}
+	}
 }


NikhilSharmaWe:
@spowelljr and @SilkeDH is my thinking correct here for adding the improvement.
If yes, 
- An error is `Error: cmd/minikube/cmd/start.go:1134:1: cyclomatic complexity 31 of func validateFlags is high (> 30) (gocyclo)` occuring. How can we increase the cyclomatic complexity or what is correct way the deal with it ?

cmd/minikube/cmd/start.go
```diff
@@ -1236,6 +1236,10 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 		exit.Message(reason.Usage, "Sorry, please set the --output flag to one of the following valid options: [text,json]")
 	}
 
+	if viper.GetStringSlice(config.AddonListFlag) != nil && viper.GetBool(noKubernetes) {
```
medyagh, line 1239: consider adding it to a helper func to fix the lint error "cyclomatic complexity 31"

helper could be caled
validateNoAddonsForNoKubernetes(config...)

medyagh (changes requested):
thank you for this PR just a lint comment
and plz paste the output the PR before and after this PR

cmd/minikube/cmd/start.go
NikhilSharmaWe: @medyagh 
- Is the function `validateNoAddonsForNoKubernetes()` correctly added ?
- I have called func `validateNoAddonsForNoKubernetes()` inside the func `provisionWithDriver()`, func `validateFlags()` was also called here, so is it a correct place to call it ?

NikhilSharmaWe:
@spowelljr I am not able to understand your thought when you say : 
- `Also, this PR is not catching the case if the user starts minikube and then run minikube addons enable <addon>.`

Could you explain me that what case actually we have tackled in this PR and how can we deal with the case you are mentioning.

spowelljr:
> @spowelljr I am not able to understand your point when you say :
> 
> * `Also, this PR is not catching the case if the user starts minikube and then run minikube addons enable <addon>.`
> 
> Could you explain me that what case actually we have tackled in this PR till now and how can we deal with the other case you have mentioned in this comment ?

Right now you're checking the case where the user runs:
```
minikube start --addons metallb
```

But also need to check for the case where user starts minikube and then enables the addon after:
```
minikube start
minikube addons enable metallb
```

NikhilSharmaWe:
@spowelljr 
- You suggested in [this](https://github.com/kubernetes/minikube/issues/13487#issuecomment-1023631485) comment of yours that use function [config.Load()](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/config/config.go#L173) to find out that `--no-kubernetes` has been set or not. Function `config.Load()` return value of type [ClusterConfig](https://github.com/kubernetes/minikube/blob/74452b5802af0850d47a113bbbc717fc8d9a0dc5/pkg/minikube/config/types.go#L34) and in this type I am not able to find out how to figure out whether `--no-kubernetes` has been set or not.

- I have done something different here, could you please review the method I have followed.

cmd/minikube/cmd/config/enable.go
```diff
@@ -40,6 +40,9 @@ var addonsEnableCmd = &cobra.Command{
 		if len(args) != 1 {
 			exit.Message(reason.Usage, "usage: minikube addons enable ADDON_NAME")
 		}
+		if ClusterFlagValue() == constants.NoKubernetesVersion {
```
spowelljr, line 43: `ClusterFlagValue()` returns the profile, so it doesn't make any sense to compare that against a Kubernetes version.

cmd/minikube/cmd/start.go
```diff
@@ -1652,3 +1653,9 @@ func exitGuestProvision(err error) {
 	}
 	exit.Error(reason.GuestProvision, "error provisioning guest", err)
 }
+
+func validateNoAddonsForNoKubernetes() {
+	if viper.GetStringSlice(config.AddonListFlag) != nil && viper.GetBool(noKubernetes) {
```
spowelljr, line 1658: `viper.GetStringSlice(config.AddonListFlag)` isn't nil when no `addons` flag is passed, try running `minikube start --no-kubernetes` yourself.

spowelljr (changes requested):

spowelljr:
> @spowelljr
> 
> * You suggested in [this](https://github.com/kubernetes/minikube/issues/13487#issuecomment-1023631485) comment of yours that use function [config.Load()](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/config/config.go#L173) to find out that `--no-kubernetes` has been set or not. Function `config.Load()` return value of type [ClusterConfig](https://github.com/kubernetes/minikube/blob/74452b5802af0850d47a113bbbc717fc8d9a0dc5/pkg/minikube/config/types.go#L34) and in this type I am not able to find out how to figure out whether `--no-kubernetes` has been set or not.
> * I have done something different here, could you please review the method I have followed.

`ClusterConfig.KubernetesConfig.KubernetesVersion` is set to `constants.NoKubernetesVersion` to indicate that `--no-kubernetes` is set.

NikhilSharmaWe:
@spowelljr 
- Made [changes](https://github.com/kubernetes/minikube/pull/13508/files#diff-db0de879aa374537c5a8dd95c611798a741123353f9f830d8398edfb58587066R43-R47) in the logic in `enable.go`.
- Also added the same logic inside the function [`validateNoAddonsForNoKubernetes`](https://github.com/kubernetes/minikube/pull/13508/files#diff-785e0c3520557a4a553e977cac218c88514f688e6f2e01e387236c7ca6dd492fR1657-R1663) in `start.go`.

Do these changes solves the issue we were having before ? Please inform.

cmd/minikube/cmd/config/enable.go
```diff
@@ -40,6 +40,11 @@ var addonsEnableCmd = &cobra.Command{
 		if len(args) != 1 {
 			exit.Message(reason.Usage, "usage: minikube addons enable ADDON_NAME")
 		}
+		if ClusterConfig, err := config.Load(ClusterFlagValue()); err == nil {
```
spowelljr, line 43:
```suggestion
		if clusterConfig, err := config.Load(ClusterFlagValue()); err == nil {
```
As per Go naming conventions

cmd/minikube/cmd/config/enable.go
```diff
@@ -40,6 +40,11 @@ var addonsEnableCmd = &cobra.Command{
...
+			if ClusterConfig.KubernetesConfig.KubernetesVersion == constants.NoKubernetesVersion {
```
spowelljr, line 44:
```suggestion
			if clusterConfig.KubernetesConfig.KubernetesVersion == constants.NoKubernetesVersion {
```

cmd/minikube/cmd/start.go
```diff
@@ -1652,3 +1653,11 @@ func exitGuestProvision(err error) {
 	}
 	exit.Error(reason.GuestProvision, "error provisioning guest", err)
 }
+
+func validateNoAddonsForNoKubernetes() {
+	if ClusterConfig, err := config.Load(ClusterFlagValue()); err == nil {
```
spowelljr, line 1658:
```suggestion
	if clusterConfig, err := config.Load(ClusterFlagValue()); err == nil {
```

spowelljr (changes requested):

Issue #12594, opened 2021-09-28
minikube tunnel hard coded docker


bhester01:
**Steps to reproduce the issue:** 

1. minikube start --embed-certs --kubernetes-version 1.20.10 --addons registry-creds --driver podman
2. minikube tunnel

<details>
E0927 19:02:29.077832  374498 out.go:451] unable to execute get port 22 for "minikube": docker container inspect -f "'{{(index (index .NetworkSettings.Ports "22/tcp") 0).HostPort}}'" minikube: exec: "docker": executable file not found in $PATH

</details>

presztak:
I've started working on this issue.
/assign

PR #13514, opened 2022-01-29
Fix hard coded docker driver in minikube service command

presztak:
This PR inroduces fix for bug similar to this one: #12594 but this time `minikube service` command is affected.
Docker driver is hard coded in ForwardPort function. This will lead to error when user is using podman instead of docker driver.


totollygeek:
Is there a way to workaround the issue, without installing docker, while this one is released?
That would be super helpful. Thanks a lot!

Issue #13407, opened 2022-01-20
Create new `releases.json` schema to support SHAs for all archs


unixlab:
### What Happened?

Hi!

I noticed that the current [releases.json](https://github.com/kubernetes/minikube/blob/master/deploy/minikube/releases.json) does not comply with the [schema](https://github.com/kubernetes/minikube/blob/master/deploy/minikube/schema.json).

I noticed because `minikube update-check` is broken since the new `releases.json` was pushed out.
(see `Attach the log file` section)

The `releases.json` generation [here](https://github.com/kubernetes/minikube/blob/master/hack/jenkins/release_update_releases_json.go#L33), does not match the corresponting counterpart [here](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/notify/notify.go#L134)

I tried to write a fix and currently all tests besides the schema check are green.
However, I'm struggling with the schema and would need help or someone else to take this over.
The current state I have is here: https://github.com/unixlab/minikube/commit/5f3f93b60ebc408e4c9c162964c99100557623a6

### Attach the log file

```
13:53:28 ~ $ minikube update-check

X Exiting due to INET_VERSION_UNAVAILABLE: Error getting json from minikube version url: json: cannot unmarshal object into Go struct field Release.Checksums of type string
* 
╭─────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                             │
│    * If the above advice does not help, please let us know:                                 │
│      https://github.com/kubernetes/minikube/issues/new/choose                               │
│                                                                                             │
│    * Please run `minikube logs --file=logs.txt` and attach logs.txt to the GitHub issue.    │
│    * Please also attach the following file to the GitHub issue:                             │
│    * - /tmp/minikube_update-check_7db9daf36ec61d5021304844f4320678ea49b1a1_0.log            │
│                                                                                             │
╰─────────────────────────────────────────────────────────────────────────────────────────────╯
```

rotty3000:
I observed exact issue just now.

Here is the log as error message requests.

[minikube_update-check_ae3ada15876f093074b98305c9e67f32241694da_0.log](https://github.com/kubernetes/minikube/files/7907648/minikube_update-check_ae3ada15876f093074b98305c9e67f32241694da_0.log)

paurana:
This is occurring because the first element of the json array is
```{
	"checksums": {
		"amd64": {
[...]

PR #13541, opened 2022-02-02
Support legacy releases.json file

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/13407

### DIFF
--- a/deploy/minikube/releases-beta-v2.json
+++ b/deploy/minikube/releases-beta-v2.json
@@ -1,0 +1,34 @@
+[
+{
+      "name": "v1.24.0-beta.0",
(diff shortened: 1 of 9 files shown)


medyagh (approved):
looks good to me, just one question, do older versions of minikube will still get notifciation that there is a new minikube version even though they only check v1 ?

spowelljr:
> looks good to me, just one question, do older versions of minikube will still get notifciation that there is a new minikube version even though they only check v1 ?

@medyagh Yes! The automation will continue to update the v1 & v2 releases files, so older versions will still get update notifications.

PR #13614, opened 2022-02-15
WIP: Automated addon security scanning

sharifelgamal:


cmd/minikube/cmd/config/scan.go
```diff
@@ -0,0 +1,130 @@
...
+				out, err := snyk.Output()
+				if err == nil {
+					klog.Infof("no vulnerabilities found for %s", image)
+					continue
+				}
+				outmap := make(map[string]interface{})
+				err = json.Unmarshal(out, &outmap)
+				if err != nil {
+					klog.Errorf("error unmarshalling json for %s: %v", image, err)
+				}
```
klaases, line 76: Should we expect to return here, on error?

Or, is it safe to continue?

cmd/minikube/cmd/config/scan.go
```diff
@@ -0,0 +1,130 @@
...
+								break
+							}
+						}
+						if !logCVE {
+							continue
+						}
+						klog.Infof("CVE for %s: %s, %s, %s, %v", image, vuln["title"], vuln["packageName"], vuln["severity"], vuln["nearestFixedInVersion"])
+						status.Enabled = false
+						updatedVersion := ""
+						if uv, ok := vuln["nearestFixedInVersion"].(string); ok {
```
klaases, line 95: Optional, we can declare without initialization here.

var updatedVersion string

cmd/minikube/cmd/config/toggle.go
```diff
@@ -0,0 +1,76 @@
...
+		// If enable is true, then we can reset manual to false since we're turning the addon back on
+		// If enable is false, then we're disabling the addon manually and manual should be true
+		addonStatus.Manual = !enable
+		addonStatus.ManualReason = toggleReason
+		status[addon] = addonStatus
+		writeStatusYAML(status)
+	},
+}
+
+var (
```
klaases, line 65: Consider inlining single entries.

pkg/addons/addons.go
```diff
@@ -60,6 +61,32 @@ var Refresh = false
 // ErrSkipThisAddon is a special error that tells us to not error out, but to also not mark the addon as enabled
 var ErrSkipThisAddon = errors.New("skipping this addon")
 
+type AddonSecurityError struct {
+	Status util.AddonStatus
+}
+
+// TODO: make this easier for the string extractor to parse
```
klaases, line 68: For Go doc, the comment should begin like:
// Error ...

pkg/addons/addons.go
```diff
@@ -117,6 +148,42 @@ func SetAndSave(profile string, name string, value string) error {
 	return config.Write(profile, cc)
 }
 
+func securityCheck(addonName string, value string) error {
```
klaases, line 151: Repeated param types can be omitted.

...(addonName, value string)...

pkg/addons/addons.go
```diff
@@ -117,6 +148,42 @@ func SetAndSave(profile string, name string, value string) error {
...
+		return nil
+	}
+
+	// --force always wins, please don't use unless absolutely necessary
+	if Force {
+		return nil
+	}
+
+	// Allow users to pass in custom images
+	// This check is totally broken, uncomment when fixed.
```
klaases, line 163: Is there an issue we can attach to this?

pkg/addons/addons.go
klaases, line 68: Is there an issue we can attach to this?

pkg/minikube/download/addons.go
```diff
@@ -0,0 +1,77 @@
...
+
+	addonsFile, err := os.ReadFile(fileLocation)
+	if err != nil {
+		return addonMap, err
+	}
+	err = yaml.Unmarshal(addonsFile, addonMap)
+	if err != nil {
+		return addonMap, err
+	}
+
```
klaases, line 75: Remove extra line break.

pkg/minikube/download/download.go
```diff
@@ -58,10 +58,18 @@ func CreateDstDownloadMock(src, dst string) error {
...
+	return download(src, dst, true)
+}
+
+func downloadNoProgressBar(src string, dst string) error {
+	return download(src, dst, false)
+}
+
 // download is a well-configured atomic download function
-func download(src string, dst string) error {
+func download(src string, dst string, progressBar bool) error {
```
klaases, line 70: Remove repeated param type.

klaases (commented):
Added a few comments for Go

cmd/minikube/cmd/config/scan.go
```diff
@@ -0,0 +1,130 @@
...
+					continue
+				}
+			}
+			for _, image := range images {
+				klog.Infof("scanning image %s for addon %s", image, a)
+				imageStatus := util.ImageStatus{Image: image.(string), CVEs: []util.CVE{}}
+				snyk := exec.Command("snyk", "container", "test", image.(string), "--json", "--severity-threshold=high")
+				out, err := snyk.Output()
+				if err == nil {
+					klog.Infof("no vulnerabilities found for %s", image)
```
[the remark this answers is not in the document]
sharifelgamal: klog.Infof automatically does the conversion for us

cmd/minikube/cmd/config/scan.go
sharifelgamal: i specifically want this to be an empty string in case it doesn't get set, so we can still use it in the CVE object below

cmd/minikube/cmd/config/scan.go
sharifelgamal: if err is not nil, then outmap will be empty and the following if will get skipped, making it safe to continue. we could otherwise add a continue statement after logging the error which is effectively the same thing

PR #13630, opened 2022-02-17
Add sanity check to disk space check

F1ko:
Fixes https://github.com/kubernetes/minikube/issues/12761

### DIFF
--- a/pkg/minikube/machine/start_test.go
+++ b/pkg/minikube/machine/start_test.go
@@ -21,6 +21,8 @@ import (
 	"net"
 	"os"
 	"testing"
+
+	"k8s.io/minikube/pkg/minikube/command"
 )
 
 const initialEtcHostsContent string = `127.0.0.1	localhost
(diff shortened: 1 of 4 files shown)


F1ko:
This PR is identical to PR https://github.com/kubernetes/minikube/pull/12814 which acted out for some reason.

@spowelljr labeld the original PR with `/ok-to-test`.

Would be nice to pick up where we left off.

F1ko:
Looking through the git history I observed that workflow failures seem to be the case quite often, even for already merged commits. Does anyone mind explaining to me how to interpret that correctly? Are these certain "know-issues" and expected to fail? In that case, how would I know whether this is due to my changes or because it indeed is a known issue?

sharifelgamal:
We do unfortunately have some known test failures, and other tests that are simply flaky, that fail for reasons unrelated to your PR. Once all the tests have run, we have a bot that will comment the flake rate of all the failed tests. If there are tons of 0% tests, then it's a safe bet that your PR broke something. Otherwise, maintainers just use their best judgement when merging PRs.

F1ko:
hi @spowelljr 

are the flake rates high enough now or should I look into other issues as well?
any advice on debugging hyper-v windows without a native windows machine?

pkg/minikube/machine/start_test.go
```diff
@@ -96,3 +98,102 @@ func writeContentToTempFile(content string) (string, error) {
...
+		},
+	}
+	for _, tt := range tests {
+		t.Logf("starting %v", tt.name)
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := DiskUsed(tt.args.cr, tt.args.dir)
+			t.Logf("err: %v\n", err)
+			t.Logf("got: %v\n", got)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("DiskUsed() error = %v, wantErr %v", err, tt.wantErr)
```
spowelljr, line 142: If you don't want execution to continue you can do `t.Fatalf` instead and then you don't have to add the return yourself.

pkg/minikube/machine/start_test.go
```diff
@@ -96,3 +98,102 @@ func writeContentToTempFile(content string) (string, error) {
...
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Logf("starting %v", tt.name)
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := DiskAvailable(tt.args.cr, tt.args.dir)
+			t.Logf("err: %v\n", err)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("DiskAvailable() error = %v, wantErr %v", err, tt.wantErr)
```
spowelljr, line 191: Same as above

spowelljr (changes requested):
Just a small thing and then it's good to merge

F1ko:
@spowelljr thanks, fixed

spowelljr (approved):

PR #13639, opened 2022-02-20
Add the qemu2 driver to the minikube registry

afbjorklund:
Since the machine drivers are hardcoded in minikube,
drivers need to be added to the registry to be tested.

Add a basic sanity check for the qemu-system binary,
and set up the basic configuration such as cpus/memory.

----

Most likely needs an updated version of the upstream driver...

https://github.com/machine-drivers/docker-machine-driver-qemu

Note: this driver uses [qemu](https://qemu.org) and runs in userspace but _can_ use kvm

There is another driver called kvm2, which uses the [libvirt](https://libvirt.org) library


mprimeaux:
@afbjorklund How can I help to progress this PR?

afbjorklund:
@mprimeaux : first of all, thanks for volunteering!

This particular PR is more about adding "qemu" to the hard-coded list of allowed machine drivers for minikube
Most likely we will end up having to fork a "qemu2" for minikube, like there was a "kvm2" done for libvirt earlier ?

There are two issues going on here, at the same time...

First we need an ISO that we can test with, and second we need an (Open Source) VM driver that can run it.
Since the current ones are behind in their development, we are trying first to update them for amd64 - then arm64.

----

Ideally, we would want a solution (for `aarch64`) that could work for `minikube` as well as for `docker-machine`.

But that is secondary, working for minikube is primary... Docker stopped all development on the project in 2019.

Our current buildroot is based on BIOS and isolinux, and needs to be changed to use UEFI and grub to boot up.

The current drivers are based on VirtualBox or on HyperKit, which are only available for the x86_64 architecture.

klaases:
Hi @mprimeaux, were you able to review @afbjorklund's comments?

mprimeaux:
@afbjorklund @klaases Thanks for the overall context and related details.  

Tactically, I am unsure where to begin as I am not as familiar with the Minikube code base (yet). Seems focusing on a buildroot that uses UEFI and grub is a logical first foray. Thoughts?

afbjorklund:
It is pretty much two separate projects. And it is possible to start with x86_64, and move on to aarch64.

You will find that this PR is mostly about the minikube driver "registry", and doesn't really have any code.

The main reason why it is so big, is that the config part has been forked from libmachine (for some reason)

https://github.com/machine-drivers/docker-machine-driver-qemu/blob/master/qemu.go#L247_L256

----

The docker-machine drivers use the driver GetCreateFlags and SetConfigFromFlags methods instead.

That means that the drivers can be decoupled (using gRPC), and doesn't have to be included/linked...

So most likely we will end up having to fork "qemu" into a "qemu2", and then adopt it to minikube ?

Note that the driver doesn't really _work_ at the moment, it barely compiles and only for amd64 at that.

afbjorklund:
The driver actually works now, if you upgrade docker-machine and boot2docker and compile it from source.

afbjorklund:
I have added flags to the driver, so that it will be possible to boot different architectures (beyond amd64).

There is a truckload of them, but they can be defaulted. Both machine and cpu are mandatory for arm64.

qemu-program
qemu-machine (new)
qemu-cpu (new)
qemu-bios  (new, bool)
qemu-uefi (new, bool)
qemu-firmware (new)

afbjorklund:
Now this has both the QEMU (docker-machine) and QEMU2 (new minikube) drivers.

Briefly tested on Linux, needs some more parameters before it works on Darwin too.

Basic support for ARM64 parameters.

sharifelgamal (approved):

PR #13654, opened 2022-02-22
[benchmark] add cpu utilization

chungjin:
Append cpu benchmark test result to the site page.

eg:
Before the change: 
https://minikube.sigs.k8s.io/docs/benchmarks/timetok8s/v1.25.1/

After the change:
append following markdown table and chart to the page:
image

### DIFF
--- a/hack/benchmark/time-to-k8s/chart.go
+++ b/hack/benchmark/time-to-k8s/chart.go
@@ -74,7 +74,7 @@ func readInCSV(csvPath string, apps map[string]runs) error {
 
 		values := []float64{}
 
-		// 8-13 contain the run results
+		// 8-16 contain the run results
 		for i := 8; i <= 16; i++ {
 			v, err := strconv.ParseFloat(d[i], 64)
 			if err != nil {
--- a/hack/benchmark/time-to-k8s/chart.go
+++ b/hack/benchmark/time-to-k8s/chart.go
@@ -19,10 +19,8 @@ package main
 import (
 	"bytes"
 	"encoding/csv"
-	"flag"
 	"fmt"
 	"io"
-	"log"
 	"os"
 	"strconv"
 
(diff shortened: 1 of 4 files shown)


klaases:
Hi @medyagh, are you the right person to verify this patch?

spowelljr (changes requested):
You can prevent the graph from going into the legend by tracking the maximum value and then appending a value to it to set the height. For example the highest value + 3 would probably give proper clearance of the legend in your case.

Here's how I'm doing this for reference: https://github.com/kubernetes/minikube/blob/master/hack/benchmark/time-to-k8s/public-chart/generate-chart.go#L280

chungjin:
> You can prevent the graph from going into the legend by tracking the maximum value and then appending a value to it to set the height. For example the highest value + 3 would probably give proper clearance of the legend in your case.
> 
> Here's how I'm doing this for reference: https://github.com/kubernetes/minikube/blob/master/hack/benchmark/time-to-k8s/public-chart/generate-chart.go#L280

Thank you @spowelljr for the suggestion! I have applied it and update the graph.

hack/benchmark/time-to-k8s/chart.go
```diff
@@ -92,14 +75,15 @@ func readInCSV(csvPath string, apps map[string]runs) error {
 		values := []float64{}
 
 		// 8-13 contain the run results
```
spowelljr, line 77: Just update this comment to match the range we're using.

hack/benchmark/time-to-k8s/cpu.go
```diff
@@ -0,0 +1,116 @@
...
+	p := plot.New()
+	p.Title.Text = "CPU utilization to go from 0 to successful Kubernetes deployment"
+	p.Y.Label.Text = "CPU utilization"
+	setYMax(p, values)
+
+	w := vg.Points(20)
+
+	barsA, err := plotter.NewBarChart(values[0], w)
+	if err != nil {
+		panic(err)
```
spowelljr, line 66: Any reason we're panicking instead of just returning an error?

spowelljr (changes requested):

hack/benchmark/time-to-k8s/cpu.go
medyagh: I agree lets return error instead of panic here

medyagh (changes requested):

hack/benchmark/time-to-k8s/chart.go
chungjin: addressed

hack/benchmark/time-to-k8s/cpu.go
chungjin: done

spowelljr (approved):

PR #13722, opened 2022-02-24
fix test for checking pods after minikube restart

sharifelgamal:
This test was checking for the wrong pod, for the dashboard pods instead of the user deployed busybox pod. This will catch any sort of regression like the one we had with 1.25.0, where restarting minikube erased user deployed pods on startup.


medyagh (commented):
didnt study the PR but I am curious is this something that be improved in minikube code itself ?or the test was actually not doing the right thing?

sharifelgamal:
the test was doing the wrong thing, it wasn't checking for the user deployed app (busybox), but for the pods from the dashboard addon, which is already checked separately in another test.

klaases (approved):

klaases:
LGTM and approved, ready for merge :)

spowelljr:
`UserAppExistsAfterStop` seems to be consistently failing

PR #13723, opened 2022-02-24
Added initial GUI

spowelljr:
v0.0.1 of the minikube GUI

Screenshots on Linux
Screen Shot 2022-03-03 at 9 37 44 AM
Screen Shot 2022-03-03 at 9 38 15 AM
Screen Shot 2022-03-03 at 9 38 30 AM

**GUI Binaries**
Mac: https://storage.googleapis.com/minikube-gui/v0.0.1/minikube-gui-mac.zip
Linux: https://storage.googleapis.com/minikube-gui/v0.0.1/minikube-gui-linux.zip
Windows: https://storage.googleapis.com/minikube-gui/v0.0.1/minikube-gui-windows.zip


gui/LICENSE
```diff
@@ -0,0 +1,27 @@
+Copyright (C) 2016 The Qt Company Ltd.
```
afbjorklund, line 1: I think this one should have a minikube Copyright at the top of it.

gui/README.md
```diff
@@ -0,0 +1,24 @@
...
+$ make
+...
+$ ./systray
+```
+
+----
+
+See <https://doc.qt.io/qt-5/qtwidgets-desktop-systray-example.html>
+
+minikube logo from the <https://github.com/kubernetes/minikube> project.
```
afbjorklund, line 24: this might be redundant, if the code is _in_ the minikube project ?

gui/instance.h
```diff
@@ -0,0 +1,67 @@
+#ifndef INSTANCE_H
+#define INSTANCE_H
+
+#include <QAbstractListModel>
+#include <QString>
+#include <QList>
+#include <QMap>
+
+//! [0]
+class Instance
```
afbjorklund, line 10: You could rename this class to "Profile", or maybe better yet "Cluster", to match minikube vocabulary ?

gui/window.cpp
```diff
@@ -0,0 +1,472 @@
...
+    QModelIndexList index = model->match(start, Qt::DisplayRole, instance);
+    if (index.size() == 0) {
+        return;
+    }
+    instanceListView->setCurrentIndex(index[0]);
+}
+
+void Window::createInstanceGroupBox()
+{
+    instanceGroupBox = new QGroupBox(tr("Instances"));
```
afbjorklund, line 295: Maybe "Clusters", or something ? Same as `minikube profile list`

afbjorklund:
Looking great! I think it should have some minor cleanups and overhaul, to "translate" some of the Lima things (like "instance")

You might want to consider adding `clang-format` or some other automatic .cpp cleanup tool, similar to the `gofmt` used with .go

afbjorklund:
I would love to see a "SSH" button and a "Dashboard" button, opening up a terminal or a web browser - respectively

afbjorklund (changes requested):
Please update the license, and possible change some class names.

afbjorklund (approved):
Looks good! Not sure if the license needs to be in every file, but thanks for updating it!

gui/window.cpp
```diff
@@ -0,0 +1,587 @@
...
+{
+    QDialog dialog;
+    dialog.setWindowTitle(tr("Set Cluster Values"));
+    dialog.setModal(true);
+
+    QFormLayout form(&dialog);
+    driverComboBox = new QComboBox;
+    driverComboBox->addItem("docker");
+    driverComboBox->addItem("hyperv");
+    driverComboBox->addItem("hyperkit");
```
afbjorklund, line 478: the "kvm2" driver is missing in this list, for Linux users.

Also "virtualbox" should probably be available, no ?

afbjorklund:
@medyagh : I would still prefer to have this as a separate git repository

gui/window.cpp
spowelljr: Added all the drivers and gated OS specific ones

medyagh:
> @medyagh : I would still prefer to have this as a separate git repository

sounds good once we have a stable release, we should be asking for a repo in kuberentes sigs

PR #13739, opened 2022-03-02
Add processing to remove duplicated escape characters

### LINKED ISSUE #13500: Some messages are not replaced with the translated text
This issue is regarding a text extraction process from the source code for multilingual support.

### What happened?
Some messages are output as the original text despite there are translated in `translations/*.json` files.
As an easy to reproduce example, the output message by `minikube cache add` command.

Output on Japanese language environment (not replaced with the translated text):
`! "minikube cache" will be deprecated in upcoming versions, please switch to "minikube image load"`

ja.json:
`"\\\"minikube cache\\\" will be deprecated in upcoming versions, please switch to \\\"minikube image load\\\"": "「minikube cache」は今後のバージョンで廃止予定になりますので、「minikube image load」に切り替えてください",`

Source code:
`out.WarningT("\"minikube cache\" will be deprecated in upcoming versions, please switch to \"minikube image load\"")`

### Cause
Escaped strings in source code are being escaped again in the extraction process.
Therefore, it does not match the source code when comparing.
As a result, the original text is output without using the translated text.

### Suggestion
Strings enclosed in double quotes in the source code should avoid duplicate escaping.
I suggest that add the processing to remove unneeded backslashes at checkString() in pkg/minikube/extract/extract.go

### What needs to be considered
There is a work required to do with this fix.
After applying this fix, running `make extract` will add new correctly escaped lines to * .json and strings.txt.
If the original text with the duplicate escape has already been translated in * .json, that line will be remained.
This is because extract.go does not remove translated lines, even if they are unused.
So, we need to move the translated text to the newly added line.

t-inu:
fixes #13500

### DIFF
--- a/pkg/minikube/extract/extract.go
+++ b/pkg/minikube/extract/extract.go
@@ -355,6 +355,12 @@ func checkString(s string) string {
 		}
 	}
 
+	// Remove unnecessary backslashes
+	if s[0] == '"' {
+		r := strings.NewReplacer(`\\`, "\\", `\a`, "\a", `\b`, "\b", `\f`, "\f", `\n`, "\n", `\r`, "\r", `\t`, "\t", `\v`, "\v", `\"`, "\"")
+		stringToTranslate = r.Replace(stringToTranslate)
+	}
+
 	// Hooray, we can translate the string!
 	return stringToTranslate
 }


sharifelgamal (approved):
Thanks for fixing this.

sharifelgamal:
I'll deal with removing dupes from the json files themselves.

PR #13762, opened 2022-03-09
Add support for building aarch64 ISO

### LINKED ISSUE #9228: Build ISO minikube image for ARM (aarch64)
If we want to run _any_ hypervisor driver on arm64, we need a new "minikube.iso" that works on the architecture.

Currently the image is full of amd64 binaries that we download from other places, so it does **not** build from source.

Buildroot does support the ARM architectures (armv7, arm64).

For instance the Raspberry Pi OS still uses 32-bit by default...

sharifelgamal:
This rather large PR adds the ability to build an arm64 buildroot ISO.
Specific changes include:
- adding CHANGELOG to the ISO rootfs overlay
- support for building ISO with UEFI bootloader (although the amd64 ISO remains BIOS for several reasons)
- full refactor of the `deploy/iso/minikube-iso` subdirectories
    - adding the `arch` directory for all arch specific packages
    - templatizing `Config.in` so each arch can define it at build time
    - move `board/coreos/minikube` to `board/minikube/<arch>`
- Make sure automation builds and recognizes both ISOs
- Make sure minikube auto-selects ISO based on arch automatically

Closes #9228

Thanks for your patience everyone.


afbjorklund:
@sharifelgamal : I added the sysdig patches from https://github.com/kubernetes/minikube/pull/13178 to make it compile, but there seems to be some issues still

also some discrepancies with "arm64" vs "aarch64" remaining, in the file names ?

sharifelgamal:
> @sharifelgamal : I added the sysdig patches from #13178 to make it compile, but there seems to be some issues still
> 
> also some discrepancies with "arm64" vs "aarch64" remaining, in the file names ?

yeah, I have done all the work with aarch64 yet, I'm still working on getting a good UEFI bootloader working on x86_64

Makefile
```diff
@@ -888,7 +888,7 @@ ifeq ($(MINIKUBE_BUILD_IN_DOCKER),y)
 	test "`strings $@ | grep '^LIBVIRT_[0-9]' | sort | tail -n 1`" = "LIBVIRT_1.2.9"
 else
 	$(if $(quiet),@echo "  GO       $@")
-	$(Q)GOARCH=$* \
+	$(Q)GOARCH=$* GOOS=linux \
```
afbjorklund, line 891: The kvm2 driver needs libvirt for linking, so it doesn't support cross-compiling anyway - thus the GOOS is not needed

sharifelgamal: gotcha, good catch. i was testing something and this change was leftover.

sharifelgamal:
ok-to-build-iso

sharifelgamal:
ok-to-build-iso

sharifelgamal:
ok-to-build-iso

sharifelgamal:
ok-to-build-iso

sharifelgamal:
ok-to-build-iso

sharifelgamal:
ok-to-build-iso

sharifelgamal:
ok-to-build-iso

sharifelgamal:
ok-to-build-iso

Makefile
```diff
@@ -305,11 +314,11 @@ iso-menuconfig: ## Configure buildroot configuration
 linux-menuconfig:  ## Configure Linux kernel configuration
```
sharifelgamal, line 314: change to linux-menuconfig-%

Makefile
```diff
@@ -322,7 +331,7 @@ iso_in_docker:
 		$(ISO_BUILD_IMAGE) /bin/bash
 
 test-iso:
```
sharifelgamal, line 333: get rid of this target

deploy/iso/minikube-iso/configs/minikube_aarch64_defconfig
```diff
@@ -1,11 +1,13 @@
 # Architecture
-BR2_x86_64=y
+BR2_aarch64=y
 
-# Toolchain
+# Toolchainmv
```
sharifelgamal, line 4: typo

deploy/iso/minikube-iso/configs/minikube_x86_64_defconfig
```diff
@@ -0,0 +1,115 @@
...
+BR2_TARGET_GENERIC_GETTY_PORT="tty1"
+BR2_ROOTFS_DEVICE_CREATION_DYNAMIC_EUDEV=y
+
+# Required tools to create bootable media
+BR2_PACKAGE_HOST_DOSFSTOOLS=y
+BR2_PACKAGE_HOST_GENIMAGE=y
+BR2_PACKAGE_HOST_MTOOLS=y
+
+# Bootloader
+#BR2_TARGET_GRUB2=y
```
sharifelgamal, line 18: explain BIOS vs EFI in comment

hack/jenkins/build_iso.sh
```diff
@@ -78,7 +79,7 @@ if [ "$release" = false ]; then
 	sed -i "s|isoBucket := .*|isoBucket := \"${ISO_BUCKET}\"|" pkg/minikube/download/iso.go
 	make generate-docs
 
-	git add Makefile pkg/minikube/download/iso.go site/content/en/docs/commands/start.md
+	git add Makefile pkg/minikube/download/iso.go site/content/en/docs/commands/start.md deploy/iso/minikube-iso/board/minikube/x86_64/rootfs-overlay/etc/CHANGELOG deploy/iso/minikube-iso/board/minikube/aarch64/rootfs-overlay/etc/CHANGELOG
```
sharifelgamal, line 82: revert this

pkg/addons/addons_gcpauth.go
```diff
@@ -101,7 +101,7 @@ func enableAddonGCPAuth(cfg *config.ClusterConfig) error {
 	}
 
 	if creds.JSON == nil {
-		out.WarningT("You have authenticated with a service account that does not have an associated JSON. The GCP Auth requires credentials with a JSON file in order to continue. The image pull secret has been imported.")
+		out.WarningT("You have authenticated with a service account that does not have an associated JSON. The GCP Auth addon requires credentials with a JSON file to in order to continue. The image pull secret has been imported.")
```
sharifelgamal, line 104: different PR

pkg/drivers/kvm/kvm.go
```diff
@@ -546,7 +546,7 @@ func (d *Driver) undefineDomain(conn *libvirt.Connect, dom *libvirt.Domain) erro
 		return nil
 	}
 
-	return dom.Undefine()
+	return dom.UndefineFlags(libvirt.DOMAIN_UNDEFINE_KEEP_NVRAM)
```
sharifelgamal, line 549: test this change with BIOS amd64 ISO + KVM

pkg/minikube/node/start.go
```diff
@@ -328,7 +328,7 @@ func Provision(cc *config.ClusterConfig, n *config.Node, apiServer bool, delOnFa
 	register.Reg.SetStep(register.StartingNode)
 	name := config.MachineName(*cc, *n)
 
-	// for sake of trasnlation process be easy we make the code a bit more verbose and the if statements may seem unnecessary
+	// Be explicit with each case for the sake of translations
```
sharifelgamal, line 331: pull this change out

medyagh (changes requested):

pkg/drivers/kvm/kvm.go
sharifelgamal: this seems to work just fine.

sharifelgamal:
*deep breath* here we go

/ok-to-test

medyagh:
omg !  Am I dreaming ? KVM and Hyperkit Tests are Green on this ISO PR ! this is one of the biggest works delivered in minikube in years ! Thank you @sharifelgamal and also @afbjorklund for the guidance

sharifelgamal:
ok-to-build-iso

sharifelgamal:
ok-to-build-iso

afbjorklund:
I am not a big fan of duplicating the package directory, but I would also be happy to refactor this _after_ it has been delivered.

sharifelgamal:
> I am not a big fan of duplicating the package directories, but I would also be happy to refactor this _after_ it has been delivered...
> 
> There should only be one package directory (for each "external" project), and there is no need to add a template to Config.in
> 
> Each package already has a dependency on the arch, where needed.
> 
> `depends on BR2_x86_64` `depends on BR2_AARCH64`

yeah, this was my understanding as well, but i was having issues with arch's building the wrong packages, so i built this system to make sure it didn't happen while testing

afbjorklund:
> i built this system to make sure it didn't happen while testing

sounds reasonable, it can be changed later without affecting the pacakges... i.e. the end make is the same, either way.

the main annoyance was when having to translate from the OS arch to the Go arch, but that's more tedious than "hard"

medyagh (approved):

Issue #12809, opened 2021-10-28
Website page for 'minikube cache' has wrong description


yosshy:
https://minikube.sigs.k8s.io/docs/commands/cache/ describes `minikube cache` command. It says:

> Add, delete, or push a local image into minikube

But the command has 3 sub-commands below:

- add
- delete
- reload

So, it should be:

> Add, delete, or reload an image into minikube

yosshy:
/assign @tstromberg
any comment?

medyagh:
@yosshy the source of confusion is that was a description not meant for listing sub commands, I would accept a PR that would fix that to be more simple and not confusing :)

t-inu:
According to https://github.com/kubernetes/minikube/pull/13764, document pages are automatically updated by modifying the source code.
So, I'll create a PR to fix the description within the source code.
/assign

t-inu:
I confirmed [`minikube cache` command page](https://minikube.sigs.k8s.io/docs/commands/cache/) had been updated automatically.

PR #13764, opened 2022-03-10
Improve 'minikube cache' page

RA489:
Improve 'minikube cache' page
fixes https://github.com/kubernetes/minikube/issues/12809


sharifelgamal:
These docs pages are automatically generated, using the info from the cmd package. If you want to change wording in command descriptions, change them there.

In addition, pushing and reloading an image are not the same thing and shouldn't replaced like this.

PR #13771, opened 2022-03-12
Pinned actions to SHA and included permissions for actions

naveensrinivasan:
Pinned dependencies for the GitHub actions - https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies
Set permissions for the actions - https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,12 +13,15 @@ on:
 env:
   GOPROXY: https://proxy.golang.org
   GO_VERSION: '1.17.7'
+permissions:
+  contents: read
+
 jobs:
   build_minikube:
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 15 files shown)


naveensrinivasan:
>Pin actions to a full length commit SHA

>Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload.

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

naveensrinivasan:
@sharifelgamal Could I please have some 👀 ?

spowelljr (approved):

PR #13777, opened 2022-03-13
use Kubernetes release versions' tags insted of just release tags

### LINKED ISSUE #13776: use Kubernetes release versions' tags insted of just release tags
### What Happened?

up to this minikube v1.25.2 we used Kubernetes release tags directly (that do not necessarily have to become actual _released versions_) to determine the newest Kubernetes stable, latest and edge versions and we should use respective Kubernetes release versions' tags instead

refs:
- https://github.com/kubernetes/minikube/blob/362d5fdc0a3dbee389b3d3f1034e8023e72bd3a7/pkg/minikube/constants/constants.go#L39
- https://github.com/kubernetes/kubernetes/releases
- https://github.com/kubernetes/kubernetes/releases/tag/v1.23.4-rc.0 (_not released_)

### Attach the log file

```
$ minikube-1.25.2 start --driver kvm2 --kubernetes-version latest
😄  minikube v1.25.2 on Opensuse-Tumbleweed
💾  Downloading Kubernetes v1.23.4-rc.0 preload ...
```

### Operating System

_No response_

### Driver

_No response_

prezha:
fixes #13776

as detailed in the issue's description, we should use github's repo ListReleases instead of ListTags to get stable, latest and edge Kubernetes versions

as a bonus, i've:
- corrected `NewestKubernetesVersion` from "v1.23.4-rc.0" to "v1.23.4" (didn't want to go to "v1.24" w/o testing)
- corrected `OldestKubernetesVersion` from "v1.16.0" to "v1.17.0", as per https://minikube.sigs.k8s.io/docs/
- bumped "github.com/google/go-github/v36/github" to "github.com/google/go-github/v43/github"


klaases (approved):

pkg/minikube/constants/constants.go
```diff
@@ -32,13 +32,12 @@ var (
 
 const (
 	// DefaultKubernetesVersion is the default Kubernetes version
-	// dont update till #10545 is solved
 	DefaultKubernetesVersion = "v1.23.5"
 	// NewestKubernetesVersion is the newest Kubernetes version to test against
 	// NOTE: You may need to update coreDNS & etcd versions in pkg/minikube/bootstrapper/images/images.go
-	NewestKubernetesVersion = "v1.23.6-rc.0"
```
sharifelgamal, line 39: any particular reason we're changing this here?

prezha: hi, @sharifelgamal, yes, the tag "v1.23.6-rc.0" was never released  and it should not be used (but was picked up by the old code logic now corrected in this pr); more details are in the issue's description: https://github.com/kubernetes/minikube/issues/13776

prezha: @sharifelgamal in the current master, we still use the unreleased "v1.23.6-rc.0" tag for the NewestKubernetesVersion, but i can remove this change from this pr if needed but then we should still correct that in a separate one?

sharifelgamal: we're upgrading newest to 1.24.0 in a separate PR soon, so this is kind of a moot point. but in general, we allow RCs and betas to be NewestKuberenetesVersion, mostly for testing.

prezha: yes, we do allow RCs and betas, that's fine, but at times we've also (mistakenly) picked up some _tags_ that were not _released_ tags - ie, that never made to the https://github.com/kubernetes/kubernetes/releases list

the [v1.23.6-rc.0](https://github.com/kubernetes/kubernetes/releases/tag/v1.23.6-rc.0) here is a good example of that, while eg [v1.24.0-rc.1](https://github.com/kubernetes/kubernetes/releases/tag/v1.24.0-rc.1) is an example of a valid, ie - _released_ RC tag

this pr should fix that issue so that we only pick up _released_ tags (including RCs, betas, etc.)

spowelljr (approved):
Thanks for the PR @prezha!

Issue #12284, opened 2021-08-16
TestPause/serial/Pause failing


medyagh:
https://storage.googleapis.com/minikube-builds/logs/master/20006/Docker_Linux_containerd.html#fail_TestPause%2fserial%2fPause

```
	I0814 09:37:19.058238  164716 ssh_runner.go:149] Run: sudo runc --root /run/containerd/runc/k8s.io pause 0fcd2105780a328964f9c30e4fc83c19689d1d0a6aac05dea8ef621aa6bb0216
	I0814 09:37:19.073009  164716 ssh_runner.go:149] Run: sudo runc --root /run/containerd/runc/k8s.io pause 0fcd2105780a328964f9c30e4fc83c19689d1d0a6aac05dea8ef621aa6bb0216 66b515b3e4a14fa94b7c66bf716bbb6b1a292a0066cd3bd9aa09cd86441b0a97
	I0814 09:37:19.085378  164716 retry.go:31] will retry after 276.165072ms: runc: sudo runc --root /run/containerd/runc/k8s.io pause 0fcd2105780a328964f9c30e4fc83c19689d1d0a6aac05dea8ef621aa6bb0216 66b515b3e4a14fa94b7c66bf716bbb6b1a292a0066cd3bd9aa09cd86441b0a97: Process exited with status 1
	stdout:
	Incorrect Usage.
	
	NAME:
	   runc pause - pause suspends all processes inside the container
	
	USAGE:
	   runc pause <container-id>
	
	Where "<container-id>" is the name for the instance of the container to be
	paused. 
	
	DESCRIPTION:
	   The pause command suspends all processes in the instance of the container.
	
	Use runc list to identify instances of containers and their current status.
	
	stderr:
	time="2021-08-14T09:37:19Z" level=error msg="runc: \"pause\" requires exactly 1 argument(s)"
	I0814 09:37:19.361816  164716 ssh_runner.go:149] Run: sudo systemctl is-active --quiet service kubelet
	I0814 09:37:19.371957  164716 pause.go:50] kubelet running: false
	I0814 09:37:19.371999  164716 ssh_runner.go:149] Run: sudo systemctl disable --now kubelet

```

medyagh:
could it be the "runc" changed the api to no longer accept Multiple containers to be Paused with one command ?

```
[...]

PR #13783, opened 2022-03-14
Fix trying to pause multiple containers with runc

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/12284

**Problem:**
`runc` only supports pausing one container at a time. The container IDs were being looped over properly, but was appending to the previous command causing multiple container IDs being provided.

**Before:**
```
// assuming container IDs 123, 456, 789
sudo runc --root /run/containerd/runc/k8s.io pause 123
sudo runc --root /run/containerd/runc/k8s.io pause 123 456
sudo runc --root /run/containerd/runc/k8s.io pause 123 456 789
```

**After:**
```
// assuming container IDs 123, 456, 789
sudo runc --root /run/containerd/runc/k8s.io pause 123
sudo runc --root /run/containerd/runc/k8s.io pause 456
sudo runc --root /run/containerd/runc/k8s.io pause 789
```

### DIFF
--- a/pkg/minikube/cruntime/cri.go
+++ b/pkg/minikube/cruntime/cri.go
@@ -140,13 +140,14 @@ func listCRIContainers(cr CommandRunner, root string, o ListContainersOptions) (
 
 // pauseContainers pauses a list of containers
 func pauseCRIContainers(cr CommandRunner, root string, ids []string) error {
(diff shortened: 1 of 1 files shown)


klaases (approved):
LGTM!

PR #13790, opened 2022-03-15
site: support nested tabs

spowelljr:
The `find`s are too general to support nested tabs, so replaced with `children` to not touch nested tabs.

**Before:**
Screen Shot 2022-03-15 at 12 08 03 PM

**After:**
Screen Shot 2022-03-15 at 12 07 03 PM


klaases (approved):
LGTM

Issue #13656, opened 2022-02-22
minikube upper than 1.24 fail to generate metrics container_cpu_system_seconds_total as a counter


gaellm:
### What Happened?

When starting minikube 1.25.0 or 1.25.1 (with docker driver, or virtualbox), my CPU metrics graphs fail. I've observed that some prometheus counters value stay constant:
<img width="1420" alt="Capture d’écran 2022-02-22 à 00 17 42" src="https://user-images.githubusercontent.com/7977296/155094922-bb3a8a62-14dc-448e-89ec-d415a94620a0.png">

 When I downgrade to minikube 1.24.0, all works fine:
<img width="1570" alt="Screenshot 2022-02-22 at 09 30 08" src="https://user-images.githubusercontent.com/7977296/155094515-8b9e5bca-b46c-461e-8d1e-11cfc2844329.png">

My Prometheus scrap url is: /api/v1/nodes/minikube/proxy/metrics/cadvisor

To reproduce: 
```bash
brew install --cask virtualbox
brew install minikube
minikube start
git clone https://github.com/gaellm/minikube-1-25-issue-demo.git
cd minikube-1-25-issue-demo
./deploy.sh
```
Then with a port-forward 9090 on prometheus pod you can observe the container_cpu_system_seconds_total metric.

(tested with 2 computers, an arm and an amd mac, with same conclusions)

### Attach the log file

[log.txt](https://github.com/kubernetes/minikube/files/8115123/log.txt)

### Operating System

macOS (Default)

### Driver

VirtualBox

spowelljr:
Hi @gaellm, thanks for reporting your issue with minikube!

If you do `minikube start --extra-config=kubelet.housekeeping-interval=10s` it should report metrics correctly, the usage reporting interval was increased to alleviate some CPU usage but you can set it back to the default.

gaellm:
Hi, thanks @spowelljr, indeed, this parameter solves the problem.

asaintsever:
Hi @spowelljr 

I noticed same kind of issue today as I was playing with the metrics server on minikube 1.25.2. I was puzzled first as I was sure I got metrics some month ago on previous minikube releases. So I searched here for known related issues and, luckily, stumbled on this one.
It happens that providing your extra param fixes the issue. I can see that the new default at cluster creation is 'kubelet.housekeeping-interval=5m'. Problem is that metrics server/metrics are totally broken now and we have no way to find the root cause as:

- `kubectl top node` is working
[...]

Issue #13620, opened 2022-02-15
CPU usage is not shown in minikube v1.25.1 when --kubernetes-version=v1.21.2 is specified


Noksa:
### What Happened?

CPU metrics are not working anymore. They always are zero for all containers.
It works the same for `none` and `docker` drivers. About other drivers I'm not sure.

```
➜  ~ minikube version
minikube version: v1.25.1
commit: 3e64b11ed75e56e4898ea85f96b2e4af0301f43d
```
```
➜  ~ minikube start --driver docker --kubernetes-version=v1.21.2
😄  minikube v1.25.1 on Linuxmint 20.1
    ▪ KUBECONFIG=/home/alexandr/.kube/config:/home/alexandr/.kube/k3s_config:/home/alexandr/Work/DEBUG/rke/kube_config_cluster.yml
✨  Using the docker driver based on user configuration
👍  Starting control plane node minikube in cluster minikube
🚜  Pulling base image ...
💾  Downloading Kubernetes v1.21.2 preload ...
    > preloaded-images-k8s-v16-v1...: 499.08 MiB / 499.08 MiB  100.00% 9.94 MiB
🔥  Creating docker container (CPUs=2, Memory=16000MB) ...
🐳  Preparing Kubernetes v1.21.2 on Docker 20.10.12 ...
    ▪ kubelet.housekeeping-interval=5m
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: storage-provisioner, default-storageclass

❗  /home/linuxbrew/.linuxbrew/bin/kubectl is version 1.23.3, which may have incompatibilites with Kubernetes 1.21.2.
    ▪ Want kubectl v1.21.2? Try 'minikube kubectl -- get pods -A'
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default
➜  ~ minikube addons enable metrics-server
    ▪ Using image k8s.gcr.io/metrics-server/metrics-server:v0.4.2
🌟  The 'metrics-server' addon is enabled
```

CPU usage is always 0 for all containers all time:
```
➜  ~ kubectl top pod -A
NAMESPACE     NAME                              CPU(cores)   MEMORY(bytes)
kube-system   coredns-558bd4d5db-jc4qb          0m           8Mi
[...]

PR #13802, opened 2022-03-16
 Add `--disable-metrics` flag

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/13656

Problem:
Implemented https://github.com/kubernetes/minikube/pull/13189 which increased the housekeeping interval to help alleviate CPU usage. However, this has seemed to completely disable metrics reporting and we have been getting numerous reports such as https://github.com/kubernetes/minikube/issues/13620.

Solution:
Removed increasing the housekeeping interval by default. Added a --disable-metrics (false by default) flag which will increase the housekeeping interval again and disable the metrics reporting for people that want to do so for CPU usage improvements.


klaases (approved):
LGTM!

sharifelgamal (approved):

PR #13803, opened 2022-03-16
Improve icacls error logging

spowelljr:
We have a user experiencing an issues with icacls https://github.com/kubernetes/minikube/issues/13404. The problem is right now we're only logging the error the command returns which is just `exit status 1`. The output of the command contains more useful information, that could help with debugging.

For example I did `icacls cat123`:
Error: `exit status 2`
Output: `cat123: The system cannot find the file specified.
Successfully processed 0 files; Failed processing 1 files`


klaases (approved):
LGTM!

sharifelgamal (approved):

Issue #11315, opened 2021-05-06
Disabling auto-pause seems to hang and then errors


briandealwis:
**Steps to reproduce the issue:** 

1. minikube start --driver=hyperkit
2. minikube addons enable autopause
3. minikube addons disable autopause

```console
$ out/minikube addons disable auto-pause
    ▪ Using image gcr.io/k8s-minikube/auto-pause-hook:v0.0.2
    ▪ auto-pause addon is an alpha feature and still in early development. Please file issues to help us make it better.
    ▪ https://github.com/kubernetes/minikube/labels/co/auto-pause

❌  Exiting due to MK_DISABLE: run callbacks: running callbacks: [sudo KUBECONFIG=/var/lib/minikube/kubeconfig /var/lib/minikube/binaries/v1.20.2/kubectl delete -f /etc/kubernetes/addons/auto-pause.yaml -f /etc/kubernetes/addons/auto-pause-hook.yaml: Process exited with status 1
stdout:

stderr:
the path "/etc/kubernetes/addons/auto-pause.yaml" does not exist
the path "/etc/kubernetes/addons/auto-pause-hook.yaml" does not exist
]

╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                         │
│    😿  If the above advice does not help, please let us know:                                                           │
│    👉  https://github.com/kubernetes/minikube/issues/new/choose                                                         │
│                                                                                                                         │
│    Please attach the following file to the GitHub issue:                                                                │
│    - /var/folders/jp/fyjblw1x6k71gyjnz3smzbjr00kl65/T/minikube_addons_ce22822d07a3ab9915e7892b74ddb1d9142b0896_0.log    │
│                                                                                                                         │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

[minikube_logs_dcdd5708ad01e033d64061e75b78269013f8617a_0.log](https://github.com/kubernetes/minikube/files/6437239/minikube_logs_dcdd5708ad01e033d64061e75b78269013f8617a_0.log)

medyagh:
thank you for this issue @briandealwis

klaases:
Note: according to `minikube addons list` the addon is now named "auto-pause".

PR #13805, opened 2022-03-16
Improve autopause error handling

klaases:
Fixes: #11315 

- Disabling auto-pause no longer returns an error.
- Disabling auto-pause for the first time is faster, but still takes about a minute so please be patient.
- Subsequent enable/disable only takes a few seconds.

Other improvements:

1. error handling - return an error if an error occurs during enabling auto-pause.
2. error handling - return an error if an error occurs when calculating ForwardedPort().
3. log formatting - fix a log formatting error where an int is passed as a string.
4. log parameter - if port is updated, log the calculated forwarding port rather than original port.

Minikube start.
```
./out/minikube start --driver=hyperkit
😄  minikube v1.25.2 on Darwin 11.6.4
🐳  Preparing Kubernetes v1.23.3 on Docker 20.10.12 ...
    ▪ kubelet.housekeeping-interval=5m
    ▪ Configuring RBAC rules ...
```

Enable auto-pause.
```
./out/minikube addons enable auto-pause
    ▪ Using image gcr.io/k8s-minikube/auto-pause-hook:v0.0.2
    ▪ https://github.com/kubernetes/minikube/labels/co/auto-pause
🌟  The 'auto-pause' addon is enabled
```

Disable auto-pause.
```
./out/minikube addons disable auto-pause
    ▪ https://github.com/kubernetes/minikube/labels/co/auto-pause
🌑  "The 'auto-pause' addon is disabled
```

### DIFF
--- a/cmd/auto-pause/auto-pause.go
+++ b/cmd/auto-pause/auto-pause.go
@@ -103,7 +103,7 @@ func runPause() {
 	mu.Lock()
 	defer mu.Unlock()
 	if runtimePaused {
-		out.Step(style.AddonEnable, "Auto-pause is already enabled.")
+		out.Styled(style.AddonEnable, "Auto-pause is already enabled.")
 		return
 	}
 
--- a/cmd/auto-pause/auto-pause.go
+++ b/cmd/auto-pause/auto-pause.go
@@ -103,6 +103,7 @@ func runPause() {
 	mu.Lock()
 	defer mu.Unlock()
 	if runtimePaused {
+		out.Step(style.AddonEnable, "Auto-pause is already enabled.")
 		return
 	}
 
--- a/pkg/addons/addons_autopause.go
+++ b/pkg/addons/addons_autopause.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors All rights reserved.
+Copyright 2021 The Kubernetes Authors All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


cmd/auto-pause/auto-pause.go
```diff
@@ -103,6 +103,7 @@ func runPause() {
 	mu.Lock()
 	defer mu.Unlock()
 	if runtimePaused {
+		out.Step(style.AddonEnable, "Auto-pause is already enabled.")
```
medyagh, line 106: this doesnt need to be a Step,
Step is specially used for Progress bar

pkg/addons/addons_autopause.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Kubernetes Authors All rights reserved.
+Copyright 2022 The Kubernetes Authors All rights reserved.
```
medyagh, line 2: copyright stays same since it was created, no need to change this line

cmd/auto-pause/auto-pause.go
klaases: Yes, we can use "out.Styled" instead of "out.Step".

https://pkg.go.dev/k8s.io/minikube/pkg/minikube/out#Styled

At the end of this function (line 124) there is a call to "out.Step". Should this be replaced with "out.Styled" for consistency?

pkg/addons/addons_autopause.go
klaases: Yes, will retain original copyright year.

cmd/auto-pause/auto-pause.go
spowelljr: I believe that one should stay, as line 124 is saying "The containers have paused" and that's essentially a step.

medyagh:
ok-to-test

PR #13813, opened 2022-03-20
Line 20 --no-vtx-check  ..  Adding  some prerequisite here many of us may have virtualization disabled in BIOS.

shubham-cmyk:
Enabling the virtualization in the BIOS first before running '--no-vtx-check', by default the virtualization is disabled in BIOS


afbjorklund:
This is supposed to be documented by VirtualBox

https://www.virtualbox.org/manual/

Issue #11543, opened 2021-05-31
Exit with error when Docker version is too old


dinever:
#10369 dropped the error in the Docker version validation to support old Docker versions.

With this change, the users get a confusing `docker info --format : exit status 2` error when they're using ancient Docker versions (<1.13.0) that don't even support `docker info --version`, see 
#11480
#11538

Should we exit with an error and tell users to upgrade their Docker version when it's way too old?

One option is to define two constants: `minDockerVer` (e.g. 1.13.0) and `recommendedDockerVer` (18.09.0):
1. If user Docker version < `minDockerVer`: Exit with error message
2. If user Docker version < `recommendedDockerVer`: Only show suggestions to upgrade Docker
3. If user Docker version >= `recommendedDockerVer`: Do nothing

dinever:
Related to #9077

`docker info --format` and `docker system` were both added in [v1.13.0](https://github.com/moby/moby/releases?after=v1.13.0)

`docker cp -a` was added in v18.06.0: https://github.com/docker/cli/commit/7a767a1a7e5413f37fdfbb2a1fed87e4718a79b5

dinever:
Another option is to remind the user to upgrade Docker when commands like `docker info --format` and `docker system` failed.

afbjorklund:
We should bump the required version for the Docker driver to 19.03, and leave the old versions _only_ for the container runtime...

We don't want to spend all resources "supporting" old deprecated versions of Docker and Podman, when even upstream doesn't.

afbjorklund:
@dinever : we already have `var minDockerVersion = []int{18, 9, 0}` since 4500be129ffe7a4e75761c58d09815c5fae368e9

It should exit with an error, at least require some kind of force flag to run. Recommend 20.10.

medyagh:
I am in favor of at least giving solution message to user if their old docker is cause of the problem like this https://github.com/kubernetes/minikube/issues/11480

(we should still allow users to run old dockers with --force if they wanna do it anyways)

@dinever  I would accept a PR that does that (or updates our current validation)

sharifelgamal:
Let's add this back in soon.

medyagh:
@dinever are u interested in this PR ?

PR #13842, opened 2022-03-23
Implemented minimum and recommended Docker versions

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/11543

If user is using a Docker version older than the minimum supported version (18.09) will exit and notify them to update to a newer version, unless `--force` is specified.

If used is using a Docker version older than the recommended version (20.10) will notify them to update to a newer version, unless `--force` is specified.


spowelljr:
KVM Linux machine is experiencing issues, running tests again

klaases (approved):

PR #13844, opened 2022-03-24
Fix German translation

stocc:
The output of `minikube stop` in German caused an error

## Before
```bash
$ minikube stop
✋  Stoppe Node "minikube" ...
E0324 11:12:13.006700   92294 out.go:449] unable to parse "🛑  {{.count}} Node{{if get .count 1}}s{{end}} angehalten.\n": template: 🛑  {{.count}} Node{{if get .count 1}}s{{end}} angehalten.
:1: function "get" not defined - returning raw string.
E0324 11:12:13.006847   92294 out.go:449] unable to parse "🛑  {{.count}} Node{{if get .count 1}}s{{end}} angehalten.\n": template: 🛑  {{.count}} Node{{if get .count 1}}s{{end}} angehalten.
:1: function "get" not defined - returning raw string.
🛑  {{.count}} Node{{if get .count 1}}s{{end}} angehalten.

```

## Now

```bash
$ ./out/minikube stop
✋  Stoppe Node "minikube" ...
🛑  1 Node angehalten.
```


spowelljr (approved):
Thanks for the fix!

PR #13856, opened 2022-03-25
Improve description for minikube cache command

### LINKED ISSUE #12809: Website page for 'minikube cache' has wrong description

t-inu:
fixes #12809

`minikube cache` command has sub commands that are add, delete, reload.
But current description is below.
`Add, delete, or push a local image into minikube`

It's better to use same words as the sub command names.
Also, "push" means "add" in the following document, so it makes confusion that both "add" and "push" are present in the description.
https://minikube.sigs.k8s.io/docs/handbook/pushing/#2-push-images-using-cache-command

### DIFF
--- a/cmd/minikube/cmd/cache.go
+++ b/cmd/minikube/cmd/cache.go
@@ -38,8 +38,8 @@ const allFlag = "all"
 // cacheCmd represents the cache command
 var cacheCmd = &cobra.Command{
 	Use:   "cache",
-	Short: "Add, delete, or push a local image into minikube",
-	Long:  "Add, delete, or push a local image into minikube",
+	Short: "Manage cache for images",
+	Long:  "Add an image into minikube as a local cache, or delete, reload the cached images",
 }
 
 // addCacheCmd represents the cache add command


cmd/minikube/cmd/cache.go
```diff
@@ -38,8 +38,8 @@ const allFlag = "all"
 // cacheCmd represents the cache command
 var cacheCmd = &cobra.Command{
 	Use:   "cache",
-	Short: "Add, delete, or push a local image into minikube",
-	Long:  "Add, delete, or push a local image into minikube",
+	Short: "Add an image into minikube as a local cache, or delete, reload the cached images",
```
medyagh, line 41: I suggest better wording and while we are at it, lets have different short and long descriptions

medyagh (changes requested):

cmd/minikube/cmd/cache.go
t-inu: I agree. How about the following as Short?
"Cache images into minikube, or delete, reload them"
or more simply
"Manage cache for images"

FYI, Long is output as the command usage. Both Short and Long are used in the document site.

medyagh:
[t-inu](https://github.com/t-inu) are u still working on this

t-inu:
@medyagh
I updated Short to a simple message.
If you have any opinions, please comment.

medyagh (approved):

PR #13862, opened 2022-03-26
NodePort from 8080 to 80

traviscancode604:
Making this change will allow the application to be viewable when the port forwarding is performed.

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -533,11 +533,11 @@ minikube dashboard
 
 <h2 class="step"><span class="fa-stack fa-1x"><i class="fa fa-circle fa-stack-2x"></i><strong class="fa-stack-1x text-primary">4</strong></span>Deploy applications</h2>
 
-Create a sample deployment and expose it on port 8080:
+Create a sample deployment and expose it on port 80:
 
 ```shell
 kubectl create deployment hello-minikube --image=k8s.gcr.io/echoserver:1.4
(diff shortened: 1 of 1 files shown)


afbjorklund (changes requested):
~~The 8080 is where it is exposed, not where it runs in the pod~~

My bad, actually the 8080 _was_ where it was running - it is exposed on a random port, which is probably how you want it.

```console
$ kubectl expose deployment hello-minikube --type=NodePort --port=8080
service/hello-minikube exposed
$ kubectl get svc hello-minikube
NAME             TYPE       CLUSTER-IP    EXTERNAL-IP   PORT(S)          AGE
hello-minikube   NodePort   10.98.73.79   <none>        8080:32671/TCP   3s
$ minikube ip
192.168.49.2
$ minikube service --url hello-minikube
http://192.168.49.2:32671
```

afbjorklund:
> Making this change will allow the application to be viewable when the port forwarding is performed.

It worked just fine before this change, since the echoserver listens on port 8080 (not on port 80)

```
/etc/nginx/nginx.conf:        listen 8080 default_server reuseport;
```

I'm not sure what you are trying to do here ?

traviscancode604:
It didn't work for my ARM64 M1 when I followed all the instructions. I needed to expose port 80 and not 8080 for it to work. But if this is an issue for only me and no one else then no sense in a PR.

afbjorklund:
Hmm, maybe it was a separate nginx-based image ?

The 1.4 above is ancient, and only for amd64

PR #13873, opened 2022-03-28
Silent fail when user trying to generate log file if cluster not running

spowelljr:
**Problem:**
Our GitHub issue template tells users to run `minikube logs --file=logs.txt` to generate log files and upload them to the issue.

The problem arises when if the host isn't running (which is common if you're reporting an issue) the following occurs:
```
$ minikube logs --file=logs.txt
```
This confuses many users as the message states `The control plane node must be running for this command` and they think that their log file wasn't generated and results in a log file not being attached to the issue.

**Solution:**
If the user specified the `--file` flag and the host is not running, I've suppressed outputting the above message to not confuse the user.

### DIFF
--- a/cmd/minikube/cmd/logs.go
+++ b/cmd/minikube/cmd/logs.go
@@ -19,14 +19,17 @@ package cmd
 import (
 	"os"
 
+	"github.com/docker/machine/libmachine/state"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


sharifelgamal:
Will this work if there's no cluster running?

spowelljr:
> Will this work if there's no cluster running?

If no profile even exists the user would see:
```
🤷  Profile "minikube" not found. Run "minikube profile list" to view all profiles.
```

And if the cluster is not running the user will not see any message if the `--file` flag is passed, that's the goal of this PR.

sharifelgamal (approved):

PR #13876, opened 2022-03-29
[Addon] add volcano addon

hwdef:
Add addon for [volcano](https://github.com/volcano-sh/volcano)


hwdef:
Ref: https://github.com/volcano-sh/volcano/issues/2128

spowelljr:
Hi @hwdef, thanks for your PR!

It would be important to add integration tests for this addon to ensure that it's working and we can detect if it breaks in the future, thanks!

hwdef:
@spowelljr 
I'm a little confused. I didn't add new functions, I don't know how to add integration tests.
And I checked other addon's PRs and also didn't find how they add tests

spowelljr:
You didn't add new functions, but you did add an entirely new addon to the project. For example if `apiextensions.k8s.io/v1` gets deprecated (like what recently happened to `apiextensions.k8s.io/v1beta1`), without tests we won't know the addon has stopped working until someone tells us.

For example, here is how we test the GCP addon: https://github.com/kubernetes/minikube/blob/03b5bb7bef42648bcf4d88f62f0f5337a1d55f8c/test/integration/addons_test.go#L600

test/integration/addons_test.go
```diff
@@ -70,7 +71,7 @@ func TestAddons(t *testing.T) {
 			args = append(args, "--addons=ingress", "--addons=ingress-dns")
 		}
 		if !arm64Platform() {
-			args = append(args, "--addons=helm-tiller")
+			args = append(args, "--addons=helm-tiller", "--addons=volcano")
```
spowelljr, line 74: Will the volcano addon not work on arm64? If not we should prevent the user from being able to enable it on if they're on arm64.

spowelljr (changes requested):

test/integration/addons_test.go
hwdef: Volcano's code supports arm64, but volcano is not currently a multi-architecture image, amd64 and arm64 are different tags.  I think that according to the current design of addon, the mirror image of arm64 cannot be pulled, so the use of arm64 is temporarily disabled.  Volcano is advancing the work of mixed mirroring, when volcano supports mixed mirroring, it will open the use of arm64.

hwdef: I will modify the code to disable the use of arm64, thanks for your reminder.

hwdef: I will modify the code to disable the use of arm64, thanks for your reminder.

hwdef: @spowelljr 
I checked the helm-tiller addon and didn't find the code to disable user enablement, am I missing something?

hwdef: And there is one thing I don't understand, We fixed the sha256 value of addon, how about pulling amd64 and arm64 images with a sha256 value?

sharifelgamal: so in order to do that properly, the addon itself will need to be smart enough to know which image to apply based on the user's architecture, which we don't currently have built in to the addon functionality. i'm sure there's a way to do it in the yaml template.

medyagh:
@hwdef the volcano test fails on KVM please check the test logs
https://storage.googleapis.com/minikube-builds/logs/13876/23308/KVM_Linux.html#fail_TestAddons%2fparallel%2fVolcano

hwdef:
> @hwdef the volcano test fails on KVM please check the test logs https://storage.googleapis.com/minikube-builds/logs/13876/23308/KVM_Linux.html#fail_TestAddons%2fparallel%2fVolcano

yes, I f found the error, but I haven't found a solution yet

spowelljr:
@hwdef That shouldn't effect other tests as that command should only run once the test already fails, also, those should be outputting as warnings not errors, so there shouldn't be any effect on other tests.

The volcano tests seem to be failing due to the pod not coming up:
`addons_test.go:796: failed waiting for app=volcano-controllers pod: app=volcano-controllers within 6m0s: timed out waiting for the condition`

hwdef:
> @hwdef That shouldn't effect other tests as that command should only run once the test already fails, also, those should be outputting as warnings not errors, so there shouldn't be any effect on other tests.
> 
> The volcano tests seem to be failing due to the pod not coming up: `addons_test.go:796: failed waiting for app=volcano-controllers pod: app=volcano-controllers within 6m0s: timed out waiting for the condition`

Thanks for your explanation. How can I see the logs that the pods are not running correctly?
The volcano plugin is no different from other plugins, or even simpler, just deploy three deployments and a job, and let them communicate with each other.
I have referred to other plugins, but I didn't find something missing in my code. Is there any way to troubleshoot?

spowelljr:
I don't think there is anything wrong with the addon itself, I think the params you're using to check if the pods are up might be slightly off.

hwdef:
> It's specifically `volcano-controllers`, you have the `PodWait` set to `app=volcano-controllers` but in your YAML you have it set to `app: volcano-controller` (no trailing `s`). So if you set either change the YAML value or the test value I think the tests will pass and should be good to go.

Thanks a lot. It's my mistake. I fixed it, but the CI still wrong.

spowelljr (changes requested):
@hwdef Now the deployment check is failing. The created deployment has the name `volcano-controllers` but the test is looking for `volcano-controller` (missing trailing `s`).

hwdef:
> @hwdef Now the deployment check is failing. The created deployment has the name `volcano-controllers` but the test is looking for `volcano-controller` (missing trailing `s`).

well, I think this is a mistake from Volcano repo, I will fix it in minikube and volcano. Thanks!

hwdef:
I find some error in Docker_Linux 

```
** stderr ** 
	Error from server (NotFound): pods "ingress-nginx-admission-create-78rv8" not found
	Error from server (NotFound): pods "ingress-nginx-admission-patch-8ztbc" not found
** /stderr **
```

I don't know if it's my fault

test/integration/addons_test.go
```diff
@@ -743,3 +745,91 @@ func validateHeadlampAddon(ctx context.Context, t *testing.T, profile string) {
...
+
+	// Wait for the volcano component installation to complete
+	start := time.Now()
+	var wg sync.WaitGroup
+	wg.Add(3)
+	go func() {
+		if err := kapi.WaitForDeploymentToStabilize(client, volcanoNamespace, "volcano-scheduler", Minutes(6)); err != nil {
+			t.Errorf("failed waiting for volcano-scheduler deployment to stabilize: %v", err)
+		}
+		t.Logf("volcano-scheduler stabilized in %s", time.Since(start))
```
spowelljr, line 768: Can you move this log to an else block, currently, even if the deployment never comes up it logs that it stabilized which is not true.

spowelljr (changes requested):

spowelljr:
The Volcano addon is causing issues with the other addons, do you know what would be causing this?

https://storage.googleapis.com/minikube-builds/logs/13876/26292/Docker_Linux.html#fail_TestAddons/serial/GCPAuth

test/integration/addons_test.go
hwdef: fixed

hwdef:
> The Volcano addon is causing issues with the other addons, do you know what would be causing this?
> 
> https://storage.googleapis.com/minikube-builds/logs/13876/26292/Docker_Linux.html#fail_TestAddons/serial/GCPAuth

the volcano-admission will filter all pods, when volcano-admission is not `Running`.It will intercept all pods.

spowelljr:
Could you move the Volcano test to the serial runner than and test it first, then when it's disabled everything should work as normal, correct?

medyagh:
Hi @hwdef are you still working on this PR?

hwdef:
> Hi @hwdef are you still working on this PR?

Yes, I'm still doing it, but I'm on a business trip right now, please give me some more time.

hwdef:
There are currently some problems, which will be solved in volcano1.17. When volcano1.17 is released, continue this pr

spowelljr:
I'm going to mark the PR as WIP for now, feel free to change it back when changes are made

hwdef:
@spowelljr
I made some changes and upgraded the version of the volcano, but CI is still reporting errors, I can’t find useful error messages, can you give me some help, tell me why CI is reporting errors

hwdef:
@spowelljr @sharifelgamal
Do we still need this PR?

pkg/minikube/assets/addons.go
```diff
@@ -593,6 +593,19 @@ var Addons = map[string]*Addon{
 		"GCPAuthWebhook":     "gcr.io",
 		"KubeWebhookCertgen": "registry.k8s.io",
 	}),
+	"volcano": NewAddon([]*BinAsset{
+		MustBinAsset(addons.VolcanoAssets,
+			"volcano/volcano-development.yaml.tmpl",
+			vmpath.GuestAddonsDir,
+			"volcano-deployment.yaml",
+			"0640"),
+	}, false, "volcano", "third-party (volcano)", "", "", map[string]string{
```
spowelljr, line 602:
```suggestion
	}, false, "volcano", "third-party (volcano)", "<github_username>", "", map[string]string{
```

Before adding new addons we would like to have a maintainer listed (GitHub username), this is someone we can reach out to if say the addon is going to break with a new version of Kubernetes or a critical CVE is found in the image as we have little knowledge about the addon itself. Could you please find someone from the project that's willing to take this responsibility and put their GitHub username in the above show spot. Us reaching out will likely be extremely rare so it wouldn't be much or a burden, thanks.

spowelljr (changes requested):

pkg/minikube/assets/addons.go
hwdef: I can be the maintainer of this, I am one of the approver of volcano

pkg/minikube/assets/addons.go
```diff
@@ -593,6 +593,19 @@ var Addons = map[string]*Addon{
...
+			"volcano/volcano-development.yaml.tmpl",
+			vmpath.GuestAddonsDir,
+			"volcano-deployment.yaml",
+			"0640"),
+	}, false, "volcano", "third-party (volcano)", "hwdef", "", map[string]string{
+		"vc_webhook_manager":    "volcanosh/vc-webhook-manager:v1.7.0@sha256:75adf0548e998e08c0d2e85a9b02a159a7cc21794e04f58abbae8c316a253615",
+		"vc_controller_manager": "volcanosh/vc-controller-manager:v1.7.0@sha256:4b18a03b04bcf47aa7230f472c904480757c42d906e468067c06eeaf7d22e9c5",
+		"vc_scheduler":          "volcanosh/vc-scheduler:v1.7.0@sha256:72a507f8c4d5f26103e061bfeb8bb40bb8093c263837f6ac55c8e6bf7403778b",
+	}, map[string]string{
+		"volcanosh": "docker.io",
```
spowelljr, line 607: `volcanosh` is not an image in the above map so it should be removed. It should be replaced with a matching entry from the above map.

ie.
```
map[string]string{
		"vc_webhook_manager": "docker.io",
		"vc_controller_manager": "docker.io",
		"vc_scheduler": "docker.io",
}
```

spowelljr (changes requested):

pkg/minikube/assets/addons.go
hwdef: thanks for your suggestions. I fixed this.

lowang-bh:
Is this still need? Can we go on it?

spowelljr:
@lowang-bh It needs a rebase, I tried but I don't have write permission to their repo

Monokaix:
@hwdef Are you still working on this?

hwdef:
> @hwdef Are you still working on this?

I didn't push any further, I couldn't pass all the test bots. But I've tested my code locally with no issues. You can use my code if you want to continue.

Monokaix:
retrigger ci @minikube-bot

PR #13878, opened 2022-03-29
Add registry.k8s.io as a containerd registry endpoint.

ameukam:
Related:
  - https://github.com/kubernetes/k8s.io/issues/3411

The Kubernetes project is moving away from k8s.gcr.io to registry.k8s.io

registy.k8s.io is currently redirecting to k8s.gcr.io. Each time
minikube is started using containerd as a runtime, the container images
from k8s.gcr.io will be pulled through registry.k8s.io.

Annonce: https://groups.google.com/g/kubernetes-sig-testing/c/U7b_im9vRrM/m/7qywJeUTBQAJ

Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

### DIFF
--- a/deploy/iso/minikube-iso/package/containerd-bin/config.toml
+++ b/deploy/iso/minikube-iso/package/containerd-bin/config.toml
@@ -103,6 +103,8 @@ oom_score = 0
       [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
         [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
           endpoint = ["https://registry-1.docker.io"]
+        [plugins."io.containerd.grpc.v1.cri".registry.mirrors."k8s.gcr.io"]
+          endpoint = ["https://registry.k8s.io","https://k8s.gcr.io"]
     [plugins."io.containerd.grpc.v1.cri".image_decryption]
       key_model = ""
     [plugins."io.containerd.grpc.v1.cri".x509_key_pair_streaming]


afbjorklund:
Surely this will be handled by containerd anyway, and fixed when kubeadm changes the URL ?

deploy/iso/minikube-iso/package/containerd-bin/config.toml.default
```diff
@@ -103,6 +103,8 @@ oom_score = 0
       [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
         [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
           endpoint = ["https://registry-1.docker.io"]
+        [plugins."io.containerd.grpc.v1.cri".registry.mirrors."k8s.gcr.io"]
```
afbjorklund, line 106: the default is supposed to be whatever the `containerd config default`spits out

afbjorklund:
If the idea is to deprecate "k8s.gcr.io" at some point, then we need solutions for _all_ container runtimes

Similar to how minikube provides mirrors for people that don't have direct access to GCR at all

afbjorklund:
Note that _most_ of the minikube traffic goes to storage.googleapis.com

There should be no need to hit the registry, for anything included in k8s

ameukam:
> 

Currently registry.k8s.io is a pure redirect to k8s.gcr.io with no modification on the http requests sent by the container runtimes. The same buckets will be hit by any minikube instance.

ameukam:
will be replaced by https://github.com/containerd/containerd/pull/7038. 
/close

PR #13884, opened 2022-03-30
OLM: stop testing OLM and warn users about the addon

spowelljr:
Related https://github.com/operator-framework/operator-lifecycle-manager/issues/2534

The OLM addon hash keeps changing and is breaking the addon. We are going to skip testing OLM and warn uses that the addon no longer works until the above issue is resolved.


klaases (approved):
LGTM - the inline link to the relevant Github issue works great.

PR #13885, opened 2022-03-30
Remove docker requirement for none driver

### LINKED ISSUE #10908: None driver should take care of --container-runtime parameter

```

💡  Suggestion: Install docker

💡  Suggestion: Install docker
```

<details>

</details>

staticdev:
* Removed docker validation on none driver. This validation does not make sense since it run for all other container-runtimes, even when not using docker. It is also redundant with the same validation on docker runtime side.

### DIFF
--- a/pkg/minikube/registry/drvs/none/none.go
+++ b/pkg/minikube/registry/drvs/none/none.go
@@ -60,10 +60,6 @@ func status() registry.State {
 		return registry.State{Running: true, Error: err, Fix: "iptables must be installed", Doc: "https://minikube.sigs.k8s.io/docs/reference/drivers/none/"}
 	}
 
-	if _, err := exec.LookPath("docker"); err != nil {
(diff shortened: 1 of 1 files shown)


afbjorklund:
This check _should_ be removed, but the underlying issue still needs to be fixed. It will still fail, just somewhat later.

So while I support removing it, it will **not** fix the issue where running the "none" driver requires a fake `docker` binary

afbjorklund (approved):

staticdev:
You are right @afbjorklund. I removed the fixed #10908 from this PR. Maybe I follow up with another PR as the discussion in the issue evolves.

staticdev:
@prasadkatti
@sharifelgamal
LGTM?

sharifelgamal:
/ok-to-test

makes sense to me, let's make sure the tests don't complain, not that they should.

sharifelgamal (approved):

PR #13888, opened 2022-03-30
pr-flake-rate: Default links to last 90 days

spowelljr:
Screen Shot 2022-03-30 at 2 49 35 PM

**Before:**
The `Environment` and `Flake Rate (%)` links are linking to all-time data and take a long time to load.

**After:**
The `Environment` and `Flake Rate (%)` links are linking to last 90 days data and load much faster.


klaases (approved):
LGTM

PR #13891, opened 2022-03-30
Remove old build tags

spowelljr:
Since updating to Go 1.18, removing old build tags

Go release notes:

Go 1.17 introduced //go:build lines as a more readable way to write build constraints, instead of // +build lines. As of Go 1.17, gofmt adds //go:build lines to match existing +build lines and keeps them in sync, while go vet diagnoses when they are out of sync.

Since the release of Go 1.18 marks the end of support for Go 1.16, all supported versions of Go now understand //go:build lines. In Go 1.18, go fix now removes the now-obsolete // +build lines in modules declaring go 1.17 or later in their go.mod files.

For more information, see https://go.dev/design/draft-gobuild.

### DIFF
--- a/cmd/drivers/hyperkit/main.go
+++ b/cmd/drivers/hyperkit/main.go
@@ -1,5 +1,4 @@
 //go:build darwin && !arm64
-// +build darwin,!arm64
 
 /*
 Copyright 2016 The Kubernetes Authors All rights reserved.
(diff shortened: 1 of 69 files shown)


klaases (approved):
LGTM - thanks for removing these obsolete tags.

sharifelgamal (approved):
as long as this is still compatible with go 1.17, which it sure seems like it is, i'm cool with it.

it is worth noting that minikube won't build with anything older than that, which seems totally fine.

Issue #13868, opened 2022-03-28
Falsely stating applying icacls failed due to localization


spowelljr:
When checking if applying icacls worked, we check for a specific string in the output, however we're checking for the string in English and if the user has their OS language set to a different language the string compare fails and we output to the user the icacls operation failed.

https://github.com/kubernetes/minikube/blob/7ab41df6e7af2a225e297c9b64839005a888342a/pkg/drivers/kic/kic.go#L266-L268

PR #13896, opened 2022-03-31
Remove unnecessary icacls check

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/13868

**Problem:**
We have a string comparison checking if the command successfully based on the output of the command. However, the string comparison is hard coded to English, so if the users OS is in a different language, the check will fail and we output to the user that the command failed, which is false.

**Solution:**
I did some checking and found the following:
```
> icacls cat
cat NT AUTHORITY\SYSTEM:(I)(OI)(CI)(F)
    BUILTIN\Administrators:(I)(OI)(CI)(F)
    mini-manual-2\jenkins:(I)(OI)(CI)(F)

> $?
True

> icacls dog
dog: The system cannot find the file specified.
Successfully processed 0 files; Failed processing 1 files

> $?
False
```
If the command fails, it will be detected as an error, which we already have a check for, so the string comparison check is redundant and can safely be removed.


klaases (approved):
LGTM

Issue #13894, opened 2022-03-31
DetermineLocale() doesn't check LC_MESSAGES to select the output language


pablombg:
### What Happened?

According to the [Single UNIX ® Specification](https://pubs.opengroup.org/onlinepubs/007908799/xbd/locale.html) and the [glibc manual](https://www.gnu.org/software/libc/manual/html_node/Locale-Categories.html), the `LC_MESSAGES` environment variable should be used to determine the language of the user interface. But minikube doesn't take this variable into account.

Currently, minikube uses the unmaintained [jibber_jabber library](https://github.com/cloudfoundry-attic/jibber_jabber), which checks only LC_ALL and LANG variables on Unix-like systems, leading to the use of the wrong language in some cases.

#### Locale:

```
$ locale
LANG=es_ES.UTF-8
LC_CTYPE="es_ES.UTF-8"
LC_NUMERIC="es_ES.UTF-8"
LC_TIME="es_ES.UTF-8"
LC_COLLATE="es_ES.UTF-8"
LC_MONETARY="es_ES.UTF-8"
LC_MESSAGES=en_US.UTF-8
LC_PAPER="es_ES.UTF-8"
LC_NAME="es_ES.UTF-8"
LC_ADDRESS="es_ES.UTF-8"
LC_TELEPHONE="es_ES.UTF-8"
LC_MEASUREMENT="es_ES.UTF-8"
LC_IDENTIFICATION="es_ES.UTF-8"
LC_ALL=
```

The output messages should be in English.

#### Current output:

```
$ minikube stop
✋  Stopping node "minikube"  ...
🛑  Apagando "minikube" mediante SSH...
🛑  1 node stopped.
```

The output uses the incomplete Spanish translation.

#### Expected output:

```
$ minikube stop
✋  Stopping node "minikube"  ...
🛑  Powering off "minikube" via SSH...
🛑  1 node stopped.
```

### Attach the log file

None

### Operating System

Other

spowelljr:
Hi @pablombg, thanks for bringing up this issue!

PR #13900, opened 2022-04-01
improve language detection

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/13894

We were using an old language detection package that was archived and was missing checks listed in UNIX documentation.

Updated to a different detection package that contains language detection for all OSes.

### DIFF
--- a/pkg/minikube/translate/translate.go
+++ b/pkg/minikube/translate/translate.go
@@ -19,11 +19,9 @@ package translate
 import (
 	"encoding/json"
 	"fmt"
-	"os"
-	"runtime"
 	"strings"
 
-	"github.com/cloudfoundry-attic/jibber_jabber"
+	"github.com/Xuanwo/go-locale"
 	"golang.org/x/text/language"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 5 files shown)


klaases (approved):
Nice! Much cleaner implementation.

pkg/minikube/translate/translate.go
```diff
@@ -62,20 +60,11 @@ func T(s string) string {
 
 // DetermineLocale finds the system locale and sets the preferred language for output appropriately.
 func DetermineLocale() {
-	var locale string
-	// Allow windows users to overload the same env vars as unix users
-	if runtime.GOOS == "windows" {
-		locale = os.Getenv("LC_ALL")
```
medyagh, line 68: can u verify after this PR LC_ALL env variable can still override the language ?

spowelljr: The comment in the library says
`// LC_ALL will overwrite all LC_* options.`

medyagh: lets paste the output of LC_ALL=fr minikube profile list and verify

spowelljr:
```
$ echo $LC_ALL
fr
$ ./out/minikube profile list

🤹  Fermeture en raison de MK_USAGE_NO_PROFILE : Aucun profil minikube n'a été trouvé.
💡  Suggestion : 

    Vous pouvez en créer un en utilisant 'minikube start'.
```

Issue #13452, opened 2022-01-24
kicbase image not being cached when using `--download-only`


spowelljr:
Reproduction:

1. `minikube start --driver=docker`
2. `minikube delete --all --purge // and don't remove kic image from Docker`
3. `minikube start --driver=docker --download-only`
4. `ls ~/.minikube/cache`
5. `kic` directory will be missing

`--download-only` is used for offline usage, but if you already have the kicbase image in Docker then it skips adding it to cache.

PR #13910, opened 2022-04-04
Only skip downloading kic if --download-only=false

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/13452

**Problem:**
If the kic image already exists in Docker we skip downloading the kic image to cache. However, this causes issues if `--download-only` is specified as you can have the kic image loaded in Docker but not in the cache, resulting in the `cache` directory not containing the kic and confusing users why their offline usage isn't working.

**Solution:**
Only skip downloading the kic image if `--download-only=false` to prevent the above from happening. Also, the download checks for the existence of the file before downloading, so it will skip the download if the file is already in the cache, so this PR does not result in more unnecessary downloading.


klaases (approved):
LGTM

PR #13920, opened 2022-04-06
GUI: output error details when minikube fails to start

spowelljr:
Screen Shot 2022-04-06 at 3 12 36 PM


medyagh (commented):
can we also link to verbose logs?

medyagh (commented):
can we make the error message be printed in a block format ?
maybe a different bg color ? like github code?

```
example
````

spowelljr:
> can we also link to verbose logs?

I added the code for this, but it's currently commented out due to https://github.com/kubernetes/minikube/issues/13925

> can we make the error message be printed in a block format ? maybe a different bg color ? like github code?
> 
> ```
> example
> ```

Screen Shot 2022-04-08 at 11 38 46 AM

How is this?

medyagh (approved):

PR #13953, opened 2022-04-12
disable gcloud update check when enabling gcp-auth addon

sharifelgamal:
When enabling gcp-auth, `gcloud auth` is run under the hood, which causes gcloud to run an update check, which can potentially cause latency. Setting this env var removes that possibility without affecting the user's environment.

### DIFF
--- a/pkg/addons/addons_gcpauth.go
+++ b/pkg/addons/addons_gcpauth.go
@@ -45,10 +45,11 @@ import (
 )
 
 const (
-	credentialsPath = "/var/lib/minikube/google_application_credentials.json"
-	projectPath     = "/var/lib/minikube/google_cloud_project"
-	secretName      = "gcp-auth"
-	namespaceName   = "gcp-auth"
+	credentialsPath    = "/var/lib/minikube/google_application_credentials.json"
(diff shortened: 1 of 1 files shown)


pkg/addons/addons_gcpauth.go
```diff
@@ -64,6 +65,20 @@ func enableOrDisableGCPAuth(cfg *config.ClusterConfig, name string, val string)
 }
 
 func enableAddonGCPAuth(cfg *config.ClusterConfig) error {
+	// Disable the gcloud update check if the env var isn't already set
+	if _, c := os.LookupEnv(disableCheckEnvVar); !c {
+		err := os.Setenv(disableCheckEnvVar, "1")
```
medyagh, line 70: this is global, lets set this pass this env as part of the Command Exec Runner so only passed to the commands that we run as opposed to globally

medyagh (changes requested):

sharifelgamal:
Since this env var can't actually passed into a command runner (it's `google.FindDefaultCredentials` that respects the env var), we either need to keep it this way or just abandon this PR. I'm gonna choose the latter for the time being.

PR #13960, opened 2022-04-14
Fix ingress-dns documentation for MacOS

### LINKED ISSUE #13067: ingress-dns test fails on nslookup (MacOs)
Following https://minikube.sigs.k8s.io/docs/start/#what-youll-need

and 

https://minikube.sigs.k8s.io/docs/handbook/addons/ingress-dns/

I am not able to ping or reach the example applications for testing ingress-dns addon. Perhaps I am missing something, but the documentation doesn't seem to be clear on what else I should have done.

```zsh
brew install minikube
minikube start --memory 8192 --cpus 4
minikube addons enable ingress
minikube addons enable ingress-dns
sudo vim /etc/resolver/minikube-test
```

```config
# /etc/resolver/minikube-test
domain test
nameserver 192.168.49.2
search_order 1
timeout 5
```

```zsh
kubectl apply -f https://raw.githubusercontent.com/kubernetes/minikube/master/deploy/addons/ingress-dns/example/example.yaml
# command that fails...
nslookup hello-john.test $(minikube ip)
>> ;; connection timed out; no servers could be reached
```

[logs.txt](https://github.com/kubernetes/minikube/files/7638828/logs.txt)

gerrywd:
As can be seen on the following issue you must use the hyperkit driver or lookups will fail.

fixes #13067  

We need to tell people this clearly on the setup page or they will end up in days of debugging.

### DIFF
--- a/site/content/en/docs/handbook/addons/ingress-dns.md
+++ b/site/content/en/docs/handbook/addons/ingress-dns.md
@@ -46,6 +46,7 @@ Address: 192.168.99.169
 <h3 class="step"><span class="fa-stack fa-1x"><i class="fa fa-circle fa-stack-2x"></i><strong class="fa-stack-1x text-primary">1</strong></span>Start minikube</h2>
 
 ```bash
+# Note: On MacOS you MUST use `minikube start --driver hyperkit` or ingress-dns will not work
 minikube start
 ```
 


gerrywd:
Its a trivial documentation update and I'm not prepared to sign a contract to add this. I release this change into the public domain, so hopefully somebody else can add it and avoid pain for any other MacOS users.

PR #13982, opened 2022-04-19
fix display order for error codes documentation

sharifelgamal:
The error strings are the more useful piece of documentation, so we should put to at the top of the error codes page.


klaases (approved):
LGTM!

spowelljr (approved):

Issue #13949, opened 2022-04-12
minikube image load doesn't show any progress by default, and it should


matteoolivi:
### What Happened?

I ran `minikube image load <my really really big image>`, the command took a while, and stdout showed nothing.

I know that logs can be enabled with appropriate flags, and can convey some progress, but in my opinion **it'd be useful if by default the command showed some progress** (if nothing to know if everything is ok or not), that's akin to what's done for other commands such as `start`, that show progress to stdout even without logging

### Attach the log file

N/A

### Operating System

macOS (Default)

### Driver

Docker

afbjorklund:
Currently the "minikube image" commands doesn't show any output at all, except for the in the log, which is a bad thing...

I'm not sure it will be possible to show progress bars like when using "docker load", but at least the _steps_ could be shown.

`minikube --alsologtostderr image load busybox`

1. pull image from registry
2. save image to disk
3. copy from cache to machine
4. load image from disk

Shivansh-yadav13:
Is anyone working on this issue?

matteoolivi:
@Shivansh-yadav13 apparently no. Do you have the time and want to fix that?

Shivansh-yadav13:
@matteoolivi yes, I want to try working on this.

medyagh:
This would be a neat feature ! If anyone interested in this issue I would be happy to review the PR

Shivansh-yadav13:
@medyagh I have created a pr but I'm not sure if that is what this issue requests, I feel there is more to be done than what I have added please suggest changes. @matteoolivi @afbjorklund

eggachecat:
Any updates?

peterbecich:
Can we re-open this issue please?

PR #13997, opened 2022-04-21
wip: minikube image load show progress output

Shivansh-yadav13:
- ref: #13949 
- please suggest changes

### DIFF
--- a/cmd/minikube/cmd/image.go
+++ b/cmd/minikube/cmd/image.go
@@ -30,7 +30,9 @@ import (
 	"k8s.io/minikube/pkg/minikube/exit"
 	"k8s.io/minikube/pkg/minikube/image"
 	"k8s.io/minikube/pkg/minikube/machine"
+	"k8s.io/minikube/pkg/minikube/out"
 	"k8s.io/minikube/pkg/minikube/reason"
+	"k8s.io/minikube/pkg/minikube/style"
 	docker "k8s.io/minikube/third_party/go-dockerclient"
 )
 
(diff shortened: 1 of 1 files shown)


medyagh (commented):
does this actually add a progress bar ? or only add a info ?

Shivansh-yadav13:
> does this actually add a progress bar ? or only add a info ?

Only info but we can work on that if we plan to add a progress bar?

cmd/minikube/cmd/image.go
```diff
@@ -135,12 +140,14 @@ var loadImageCmd = &cobra.Command{
 		if imgDaemon || imgRemote {
 			image.UseDaemon(imgDaemon)
 			image.UseRemote(imgRemote)
+			out.Step(style.Caching, `Caching and loading images to all profiles.`)
```
medyagh, line 143: if we are making out.Step we need to define a step, 
if you wanna use step you can add a Json output
I suggest you either add a json output to minikube image cache command or simply use 

out.Styled instead of out.Step

medyagh (changes requested):

cmd/minikube/cmd/image.go
Shivansh-yadav13: @medyagh thanx I'm  going to use out.Styled and I'm also planning to add progress bar 
https://github.com/cheggaaa/pb

medyagh: @Shivansh-yadav13 are you still working on this ?

medyagh (changes requested):

peterbecich:
@Shivansh-yadav13 any objection to my branching this PR and completing it? Thanks

PR #14016, opened 2022-04-22
Remove week day from time in Audit log 

### LINKED ISSUE #13977: Shorten the audit log dates
Currently:
```
==> Audit <==
|---------|---------------------|----------|--------------|---------|-------------------------------|-------------------------------|
| Command |        Args         | Profile  |     User     | Version |          Start Time           |           End Time            |
|---------|---------------------|----------|--------------|---------|-------------------------------|-------------------------------|
| delete  | --user minikube-gui | minikube | minikube-gui | v1.25.2 | Mon, 18 Apr 2022 15:52:01 PDT | Mon, 18 Apr 2022 15:52:01 PDT |
|---------|---------------------|----------|--------------|---------|-------------------------------|-------------------------------|
```

You can see the `Start Time` & `End Time` dates are quite long and include things like the day of the week which are unnecessary.

It would be nice to shorten these, maybe mm/dd/yyyy + time or something similar.

Nikhil2001:
before startTime =    Fri, 22 Apr 2022 05:40:38 CEST
with this fix , startTime =  04/22/2022 05:40:38 CEST;


pkg/minikube/constants/constants.go
```diff
@@ -123,7 +123,7 @@ const (
 	ExistingContainerHostEnv = MinikubeExistingPrefix + "CONTAINER_HOST"
 
 	// TimeFormat is the format that should be used when outputting time
-	TimeFormat = time.RFC1123
+	TimeFormat = "01/02/2006 15:04:05 MST"
```
medyagh, line 126: Thank you for this PR is there a standard time.RFC that we could use ?

medyagh: how about we use   time.RFC822 
which is "02 Jan 06 15:04 MST"

medyagh (changes requested):

pkg/minikube/constants/constants.go
Nikhil2001: @medyagh so seconds are not that important and RFC822 certainly looks better. 
for example if we start 2 clients at same minute different seconds will look as below
Mon, 18 Apr 2022 15:52:01 PDT  -> 18 Apr 2022 15:52 PDT
Mon, 18 Apr 2022 15:52:10 PDT ->18 Apr 2022 15:52 PDT
may be seconds does not have any significant value here. i will upload new patch

Issue #13193, opened 2021-12-18
containerd package has the same tarball checksum problems as others


afbjorklund:
Building the ISO fails, since the upstream git archive randomly changes checksum.

This is due to including `vendor/k8s.io/client-go/pkg/version/.gitattributes`

```
base.go export-subst
```

The substitution will corrupt the git archive, by filling the git info with wrong information.

```diff
58,59c58,59
< 	gitVersion   string = "v0.0.0-master+$Format:%h$"
< 	gitCommit    string = "$Format:%H$" // sha1 from git, output of $(git rev-parse HEAD)
---
> 	gitVersion   string = "v0.0.0-master+1e5ef943eb"
> 	gitCommit    string = "1e5ef943eb76627a6d3b6de8cd1ef6537f393a71" // sha1 from git, output of $(git rev-parse HEAD)
```

It was intended to record the kubernetes version, but here has a random short rev of containerd.

Problems started with containerd 1.5.x, that started including `k8s.io/client-go` (unlike 1.4.x)

afbjorklund:
Broke the ISO build in #13089

afbjorklund:
Reported as https://github.com/containerd/containerd/issues/6387

sharifelgamal:
Yeah, we've seen this happen on occasion for containerd and crio, not a whole lot we can do about it except fix the hash for the time being.

afbjorklund:
Fixed in containerd 1.6

PR #14047, opened 2022-04-25
package/containerd-bin: update hash

eiffel-fl:
Hi.

I had troubles building minikube ISO with `containerd-bin` because its hash is not good.
So, I updated it.
Can you please check if you have the problem on your side?

Best regards.


eiffel-fl:
cc @afbjorklund

afbjorklund:
The problem is upstream.

* https://github.com/kubernetes/minikube/issues/13193

eiffel-fl:
Hum sad...
Thank you for pointing it to me.

Nonetheless, what about using the same way of getting as [upstream](https://github.com/buildroot/buildroot/blob/2e49ff10202e28a8dc942bfe11d7a623aca0bf6c/package/containerd/containerd.mk#L8)?
If I understood correctly, they first clone the `containerd` repo at v1.6.2 and creates the archive locally.
Maybe by doing so they avoid the problem you are facing?

eiffel-fl:
I will close it to avoid noise as my contribution will, sadly, not solve the problem.

afbjorklund:
containerd 1.4 and 1.6 don't have the problem

But otherwise upstream use the same tarball hashes

afbjorklund:
Not sure what you mean, the only difference is that they have proper tarball naming.

The reason why they are not affected is that containerd 1.6 changed from the short hash to the long hash.

eiffel-fl:
> The reason why they are not affected is that containerd 1.6 changed from the short hash to the long hash.

What is this short vs long hash difference?
Can you please give me a link to have a better understanding?

afbjorklund:
* https://github.com/containerd/containerd/issues/6387

Issue #14010, opened 2022-04-21
GUI: Clicking tray icon should open GUI


spowelljr:
Clicking the tray icon should open the GUI.

PR #14049, opened 2022-04-25
GUI: Clicking the tray icon opens GUI

spowelljr:
Closes https://github.com/kubernetes/minikube/issues/14010

Clicking the tray icon opens the GUI, it will also bring the GUI to the front even in the GUI is already opened to help locate the window.


medyagh (approved):

gui/window.h
```diff
@@ -179,6 +180,7 @@ private slots:
     QProcessEnvironment setMacEnv();
     QStackedWidget *stackedWidget;
     bool isBasicView;
+    void delay();
```
medyagh, line 183: in addition to the the delay how about show a progress bar or "Loading .." message and wait till it actually it finishes it

medyagh (changes requested):

gui/window.h
spowelljr: Good idea! Added
Screen Shot 2022-04-28 at 11 58 33 AM

medyagh (approved):

PR #14053, opened 2022-04-26
Drop Go template extension from YAML files

klaases:
We do not actually need to wrap the .yaml file with Go templates .tmpl extension.

_[MustBinAsset](https://pkg.go.dev/k8s.io/minikube/pkg/minikube/assets#MustBinAsset)() accepts an embed file as .yaml directly._

This helps improve developer experience with corrected syntax highlighting.


klaases:
Will keep with the established convention for now.

Issue #14073, opened 2022-04-28
add * next to current active cluster in minikube profile list


medyagh:
### What Happened?

if you have 3 clusters and you make one of them active
all minikube commands will run against that cluster,
```
$ minikube profile p1
✅  minikube profile was successfully set to p1
15:16:37 medya
$ minikube status
p1
type: Control Plane
host: Running
kubelet: Running
apiserver: Running
kubeconfig: Configured
```

this should be reflected in the minikube profile list with a star in a new column called "active"

```
$ minikube profile list
|----------|-----------|---------|--------------|------|---------|---------|-------|
| Profile  | VM Driver | Runtime |      IP      | Port | Version | Status  | Nodes |
|----------|-----------|---------|--------------|------|---------|---------|-------|
| minikube | docker    | docker  | 192.168.67.2 | 8443 | v1.23.5 | Running |     1 |
| p1       | docker    | docker  | 192.168.49.2 | 8443 | v1.23.5 | Running |     1 |
| p2       | docker    | docker  | 192.168.58.2 | 8443 | v1.23.5 | Running |     1 |
|----------|-----------|---------|--------------|------|---------|---------|-------|

```

### Attach the log file

n/a

PR #14079, opened 2022-05-01
enhance: add active column for profile list cmd

vharsh:
fixes: #14073

Before:
```bash
➜  ~ minikube profile openebs
✅  minikube profile was successfully set to openebs
➜  ~ minikube profile list   
|----------|-----------|---------|--------------|------|---------|---------|-------|
| Profile  | VM Driver | Runtime |      IP      | Port | Version | Status  | Nodes |
|----------|-----------|---------|--------------|------|---------|---------|-------|
| minikube | docker    | docker  | 192.168.49.2 | 8443 | v1.20.2 | Stopped |     1 |
| openebs  | docker    | docker  | 192.168.58.2 | 8443 | v1.20.2 | Stopped |     1 |
|----------|-----------|---------|--------------|------|---------|---------|-------|
```

After:

```bash
➜  minikube git:(profileListFix) ✗ ./out/minikube profile list
|----------|-----------|---------|--------------|------|---------|---------|-------|--------|
| Profile  | VM Driver | Runtime |      IP      | Port | Version | Status  | Nodes | Active |
|----------|-----------|---------|--------------|------|---------|---------|-------|--------|
| minikube | docker    | docker  | 192.168.49.2 | 8443 | v1.20.2 | Stopped |     1 |        |
| openebs  | docker    | docker  | 192.168.58.2 | 8443 | v1.20.2 | Stopped |     1 | *      |
|----------|-----------|---------|--------------|------|---------|---------|-------|--------|
➜  minikube git:(profileListFix) ✗ minikube profile minikube
✅  minikube profile was successfully set to minikube
➜  minikube git:(profileListFix) ✗ ./out/minikube profile list
|----------|-----------|---------|--------------|------|---------|---------|-------|--------|
| Profile  | VM Driver | Runtime |      IP      | Port | Version | Status  | Nodes | Active |
|----------|-----------|---------|--------------|------|---------|---------|-------|--------|
| minikube | docker    | docker  | 192.168.49.2 | 8443 | v1.20.2 | Stopped |     1 | *      |
| openebs  | docker    | docker  | 192.168.58.2 | 8443 | v1.20.2 | Stopped |     1 |        |
|----------|-----------|---------|--------------|------|---------|---------|-------|--------|

➜  minikube git:(profileListFix) ✗ ./out/minikube profile list -ojson | jq | grep active
  "active": "minikube",

```

Signed-off-by: Harsh Vardhan <harsh.vardhan@mayadata.io>

### DIFF
--- a/cmd/minikube/cmd/config/profile_list.go
+++ b/cmd/minikube/cmd/config/profile_list.go
@@ -149,7 +149,7 @@ func profileStatus(p *config.Profile, api libmachine.API) string {
 
 func renderProfilesTable(ps [][]string) {
 	table := tablewriter.NewWriter(os.Stdout)
-	table.SetHeader([]string{"Profile", "VM Driver", "Runtime", "IP", "Port", "Version", "Status", "Nodes"})
+	table.SetHeader([]string{"Profile", "VM Driver", "Runtime", "IP", "Port", "Version", "Status", "Nodes", "Active"})
 	table.SetAutoFormatHeaders(false)
 	table.SetBorders(tablewriter.Border{Left: true, Top: true, Right: true, Bottom: true})
 	table.SetCenterSeparator("|")
@@ -159,6 +159,7 @@ func renderProfilesTable(ps [][]string) {
 
 func profilesToTableData(profiles []*config.Profile) [][]string {
 	var data [][]string
+	currentProfile := ClusterFlagValue()
 	for _, p := range profiles {
 		cp, err := config.PrimaryControlPlane(p.Config)
 		if err != nil {
@@ -169,7 +170,11 @@ func profilesToTableData(profiles []*config.Profile) [][]string {
 		if k8sVersion == constants.NoKubernetesVersion { // for --no-kubernetes flag
 			k8sVersion = "N/A"
 		}
-		data = append(data, []string{p.Name, p.Config.Driver, p.Config.KubernetesConfig.ContainerRuntime, cp.IP, strconv.Itoa(cp.Port), k8sVersion, p.Status, strconv.Itoa(len(p.Config.Nodes))})
+		var c string
+		if p.Name == currentProfile {
+			c = "*"
+		}
+		data = append(data, []string{p.Name, p.Config.Driver, p.Config.KubernetesConfig.ContainerRuntime, cp.IP, strconv.Itoa(cp.Port), k8sVersion, p.Status, strconv.Itoa(len(p.Config.Nodes)), c})
 	}
 	return data
 }
--- a/cmd/minikube/cmd/config/profile_list.go
+++ b/cmd/minikube/cmd/config/profile_list.go
@@ -197,14 +197,12 @@ func warnInvalidProfiles(invalidProfiles []*config.Profile) {
 
 func printProfilesJSON() {
 	validProfiles, invalidProfiles, err := listProfiles()
-	activeProfile := ClusterFlagValue()
 	updateProfilesStatus(validProfiles)
 
 	var body = map[string]interface{}{}
 	if err == nil || config.IsNotExist(err) {
 		body["valid"] = profilesOrDefault(validProfiles)
 		body["invalid"] = profilesOrDefault(invalidProfiles)
-		body["active"] = activeProfile
 		jsonString, _ := json.Marshal(body)
 		out.String(string(jsonString))
 	} else {
--- a/cmd/minikube/cmd/config/profile_list.go
+++ b/cmd/minikube/cmd/config/profile_list.go
@@ -197,13 +197,14 @@ func warnInvalidProfiles(invalidProfiles []*config.Profile) {
 
 func printProfilesJSON() {
 	validProfiles, invalidProfiles, err := listProfiles()
-
+	activeProfile := ClusterFlagValue()
 	updateProfilesStatus(validProfiles)
 
 	var body = map[string]interface{}{}
 	if err == nil || config.IsNotExist(err) {
 		body["valid"] = profilesOrDefault(validProfiles)
 		body["invalid"] = profilesOrDefault(invalidProfiles)
+		body["active"] = activeProfile
 		jsonString, _ := json.Marshal(body)
 		out.String(string(jsonString))
 	} else {
--- a/pkg/minikube/config/profile.go
+++ b/pkg/minikube/config/profile.go
@@ -194,7 +194,7 @@ var DockerContainers = func() ([]string, error) {
 // invalidPs are the profiles that have a directory or config file but not usable
 // invalidPs would be suggested to be deleted
 func ListProfiles(miniHome ...string) (validPs []*Profile, inValidPs []*Profile, err error) {
-
+	activeP := viper.GetString(ProfileName)
 	// try to get profiles list based on left over evidences such as directory
 	pDirs, err := profileDirs(miniHome...)
 	if err != nil {
@@ -218,7 +218,9 @@ func ListProfiles(miniHome ...string) (validPs []*Profile, inValidPs []*Profile,
 			continue
 		}
 		validPs = append(validPs, p)
-
+		if p.Name == activeP {
+			p.Active = true
+		}
 		for _, child := range p.Config.Nodes {
 			nodeNames[MachineName(*p.Config, child)] = true
 		}
@@ -236,10 +238,13 @@ func ListValidProfiles(miniHome ...string) (ps []*Profile, err error) {
 	if err != nil {
 		return nil, err
 	}
-
+	activeP := viper.GetString(ProfileName)
 	for _, n := range pDirs {
 		p, err := LoadProfile(n, miniHome...)
 		if err == nil && p.IsValid() {
+			if p.Name == activeP {
+				p.Active = true
+			}
 			ps = append(ps, p)
 		}
 	}
--- a/test/integration/minikube_profile_test.go
+++ b/test/integration/minikube_profile_test.go
@@ -41,23 +41,23 @@ func TestMinikubeProfile(t *testing.T) {
 	for _, p := range profiles {
 		rr, err := Run(t, exec.CommandContext(ctx, Target(), "start", "-p", p))
 		if err != nil {
-			t.Fatalf("test pre-condition failed. args %q: %v", rr.Command(), err.Error())
+			t.Fatalf("test pre-condition failed. args %q: %v", rr.Command(), err)
 		}
 	}
 	// 2. Change minikube profile
 	for _, p := range profiles {
 		rr, err := Run(t, exec.CommandContext(ctx, Target(), "profile", p))
 		if err != nil {
-			t.Fatalf("cmd: %s failed with error: %v\n", rr.Command(), err.Error())
+			t.Fatalf("cmd: %s failed with error: %v\n", rr.Command(), err)
 		}
 		r, err := Run(t, exec.CommandContext(ctx, Target(), "profile", "list", "-ojson"))
 		if err != nil {
-			t.Fatalf("cmd: %s failed with error: %v\n", r.Command(), err.Error())
+			t.Fatalf("cmd: %s failed with error: %v\n", r.Command(), err)
 		}
 		var profile profileJSON
 		err = json.NewDecoder(r.Stdout).Decode(&profile)
 		if err != nil {
-			t.Fatalf("unmarshalling %s cmd output failed with error: %v\n", r.Command(), err.Error())
+			t.Fatalf("unmarshalling %s cmd output failed with error: %v\n", r.Command(), err)
 		}
 		// 3. Assert minikube profile is set to the correct profile in JSON
 		for _, s := range profile.Valid {
--- a/test/integration/minikube_profile_test.go
+++ b/test/integration/minikube_profile_test.go
@@ -1,0 +1,52 @@
+package integration
+
+import (
+	"context"
+	"encoding/json"
+	"os/exec"
+	"testing"
+
+	"k8s.io/minikube/pkg/minikube/config"
+)
+
+// profileJSON is the output of `minikube profile list -ojson`
+type profileJSON struct {
+	Valid   []config.Profile `json:"valid"`
+	Invalid []config.Profile `json:"invalid"`
+}
+
+func TestMinikubeProfile(t *testing.T) {
+	// 1. Setup two minikube cluster profiles
+	profiles := []string{UniqueProfileName("first"), UniqueProfileName("second")}
+	ctx, cancel := context.WithTimeout(context.Background(), Minutes(40))
+	for _, p := range profiles {
+		defer CleanupWithLogs(t, p, cancel)
+		_, err := Run(t, exec.CommandContext(ctx, Target(), "start", "-p", p))
+		if err != nil {
+			t.Errorf("test pre-condition failed: minikube start -p " + p + " failed: " + err.Error())
+			return
+		}
+	}
+	// 2. Change minikube profile
+	for _, p := range profiles {
+		_, err := Run(t, exec.CommandContext(ctx, Target(), "profile", p))
+		if err != nil {
+			t.Errorf("minikube profile %s failed with error: %v\n", p, err.Error())
+			return
+		}
+		r, err := Run(t, exec.CommandContext(ctx, Target(), "profile", "list", "-ojson"))
+		var profile profileJSON
+		err = json.NewDecoder(r.Stdout).Decode(&profile)
+		if err != nil {
+			t.Fatalf("minikube profile list -ojson failed with error: %v\n", err.Error())
+			return
+		}
+		// 3. Assert minikube profile is set to the correct profile in JSON
+		for _, s := range profile.Valid {
+			if s.Name == p && !s.Active {
+				t.Fatalf("minikube profile %s is not active\n", p)
+				return
+			}
+		}
+	}
+}
--- a/test/integration/minikube_profile_test.go
+++ b/test/integration/minikube_profile_test.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2022 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package integration
 
 import (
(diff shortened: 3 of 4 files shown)


vharsh:
@medyagh PTAL, thanks.

cmd/minikube/cmd/config/profile_list.go
```diff
@@ -149,7 +149,7 @@ func profileStatus(p *config.Profile, api libmachine.API) string {
 
 func renderProfilesTable(ps [][]string) {
 	table := tablewriter.NewWriter(os.Stdout)
-	table.SetHeader([]string{"Profile", "VM Driver", "Runtime", "IP", "Port", "Version", "Status", "Nodes"})
+	table.SetHeader([]string{"Profile", "VM Driver", "Runtime", "IP", "Port", "Version", "Status", "Nodes", "Active"})
```
medyagh, line 152: thank you could you please post the output of "minikube profile list -o=json"
and ensure the json output also is doing the right thing ?

I think the json output should say

"active:true"
or maybe
"default:true"

medyagh (changes requested):

cmd/minikube/cmd/config/profile_list.go
medyagh: @spowelljr do u think it should be called active or default?

spowelljr: I think `active` is better, `default` sounds a little ambiguous to me. I think `active` makes it inherently known that another can be changed to `active`, while `default` could make the user think that's always the `default` and there's not a way to change it.

vharsh: @medyagh Should I add the field into the `config.ClusterConfig` structure or should I just modify the `body map[string]interface{}`  in [profile_list.go](https://github.com/kubernetes/minikube/blob/4dd50b40aec3398dcf57758b083fb05b1b3cda15/cmd/minikube/cmd/config/profile_list.go#L198) ?

spowelljr: Hi @vharsh, the active config is already present globally in `~/.minikube/config/config.json` so it doesn't need to be added to `config.ClusterConfig`.

vharsh: @spowelljr So I should set the structure to be like, such that `config.Profile` & `config.Config.ClusterConfig` structure doesn't need changes, i.e. I just add it to the `map[string]interface{}`, correct as the other way will make the key duplicate across all configs(and due to poor structure design it'd be harder to figure out which is the default config).

```json
{
    "valid": [ {...} ],
    "invalid": [ {...} ],
    "active": "<string-name-of-the-current-minikube-profile>"
}
```

spowelljr: Hi @vharsh, that's a good question.

Let me preface my response with saying that I'm imagining this output is being consumed by another application and it's being formatting it into a nice user interface where the user can check the status of the clusters.

With that said, in my opinion having `active` in the `config.Profile` makes the most sense to me.

I understand the reasoning to have `active` at the top level as it makes it easy to know which profile is active. However, my rebuttal to `it'd be harder to figure out which is the default config` is that if someone wants to know the active profile, they can simply run `minikube profile` and get that answer.

The issue I'm seeing with this solution is that we're leaving it up to the receiver to have to inject `"active": true` into the active profile if we're considering my preface. And if we look at the output (`"valid": [ {...} ]`) we can see we're returning an array of object, so this would require them to loop through the profiles until they find the profile that matches `active` and inject it. This would be an inconvenience to the user that I would like the avoid.

If we ignore `it'd be harder to figure out which is the default config` (which can be resolved with `minikube profile`) I don't see any downsides to the method of putting `active` into `config.Profile`. However, I think implementing both methods would also be an acceptable solution as it would satisfy both cases without any downside.

Let me know what you think @vharsh

spowelljr: My above comment contradicts https://github.com/kubernetes/minikube/pull/14079#discussion_r864259308, as I was asked your previous question verbally by someone else and the intent of your question got lost in the process and I responded without properly reading it, sorry for any confusion.

vharsh: @spowelljr okay, I'm just re-iterating to make sure that I've got it right, I've got to

1. add an `Active` bool field in the `config.Profile` so that both the table & the JSON format have the field coming from a struct member
2. fill it with the correct active profile everywhere i.e. when an active one gets deleted, figure out which would be the next active one and fill it in; this would likely warrant a _separate PR_ as `minikube profile` can still output a non-existent default one(i.e. `minikube`)

Let me know if I got this right.

I've also observed something regarding profiles which isn't being handled right now, is this something we should be bothering about?

```bash
➜  minikube git:(profileListFix) ✗ minikube profile list
|----------|-----------|---------|--------------|------|---------|---------|-------|
| Profile  | VM Driver | Runtime |      IP      | Port | Version | Status  | Nodes |
|----------|-----------|---------|--------------|------|---------|---------|-------|
| minikube | docker    | docker  | 192.168.58.2 | 8443 | v1.20.2 | Running |     1 |
| openebs  | docker    | docker  | 192.168.49.2 | 8443 | v1.20.2 | Running |     1 |
| third    | docker    | docker  | 192.168.67.2 | 8443 | v1.20.2 | Running |     1 |
|----------|-----------|---------|--------------|------|---------|---------|-------|

➜  minikube git:(profileListFix) ✗ minikube delete -p minikube
🔥  Deleting container "minikube" ...
🔥  Removing /home/harsh/.minikube/machines/minikube ...
➜  minikube git:(profileListFix) ✗ minikube profile
minikube
# the above profile doesn't exist ^^
➜  minikube git:(profileListFix) ✗ ./out/minikube profile list
|---------|-----------|---------|--------------|------|---------|---------|-------|--------|
| Profile | VM Driver | Runtime |      IP      | Port | Version | Status  | Nodes | Active |
|---------|-----------|---------|--------------|------|---------|---------|-------|--------|
| openebs | docker    | docker  | 192.168.49.2 | 8443 | v1.20.2 | Running |     1 |        |
| third   | docker    | docker  | 192.168.67.2 | 8443 | v1.20.2 | Running |     1 |        |
|---------|-----------|---------|--------------|------|---------|---------|-------|--------|
```

vharsh:
@medyagh @spowelljr Can you approve the workflows now? Do the current changes look okay?

cmd/minikube/cmd/config/profile_list.go
spowelljr: @vharsh Yes, the code looks good for this PR other than the two small comments I left.

I will note that there is an "edge" case, but I don't think we should be supporting it. But if the user runs `minikube profile list -p minikube2` `minikube2` would be listed as active while the config itself may have `minikube` set as the active profile. Like I said, it doesn't make sense for the user to do that, but just thought I'd point it out, `minikube profile -p minikube2` experiences this same issue.

In regard to the active profile, we probably need to have a larger discussion about it. As you're likely aware, if you try to set a non-existing profile using `minikube profile` it will throw the following error:
```
$ minikube profile minikube2
❌  profile "minikube1" not found
💡  if you want to create a profile you can by this command: minikube start -p minikube1
```
And as you correctly pointed out, if you set `minikube2` on an existing profile and then delete that cluster it defaults back to `minikube`. If there is a `minikube` cluster it should go that, but if not I'd be okay with it picking an existing cluster at random, it should output a message after the delete to notify the user of this change such as `Switching default profile to "minikube3"` or something like that. The one thing is there is a second way to set the default profile, such as `minikube config set profile minikube2` which doesn't check if a matching cluster exists before setting it. So with this command it would still be possible to get in the situation where none of the clusters are the actually the active one. The two choices is we decide that it's okay to not have any cluster be the active cluster, or we modify the `minikube config set profile` command to do an active check like `minikube profile` does. Either way, this should be in a separate PR, but it's a good discussion to have.

pkg/minikube/config/profile.go
```diff
@@ -194,7 +194,7 @@ var DockerContainers = func() ([]string, error) {
 // invalidPs are the profiles that have a directory or config file but not usable
 // invalidPs would be suggested to be deleted
 func ListProfiles(miniHome ...string) (validPs []*Profile, inValidPs []*Profile, err error) {
-
+	activeP := viper.GetString(ProfileName)
```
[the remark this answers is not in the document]
vharsh: @spowelljr This will create a circular import, should the `ClusterFlagValue()` be moved to some `pkg/util`  as it's more of a utility function.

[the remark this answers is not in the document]
vharsh:
```bash
go: downloading k8s.io/kube-openapi v0.0.0-20220328201542-3ee0da9b0b42
package k8s.io/minikube/cmd/minikube
	imports k8s.io/minikube/cmd/minikube/cmd
	imports k8s.io/minikube/cmd/minikube/cmd/config
	imports k8s.io/minikube/pkg/addons
	imports k8s.io/minikube/pkg/kapi
	imports k8s.io/minikube/pkg/minikube/proxy
	imports k8s.io/minikube/pkg/minikube/config
	imports k8s.io/minikube/cmd/minikube/cmd/config: import cycle not allowed
make: *** [Makefile:215: out/minikube] Error 1
```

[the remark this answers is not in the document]
spowelljr: Sorry, I missed that this file was outside the `cmd` package. That could be something that's done in the future, that flag is used extensively, so that's something that should be in a separate PR as to not bloat this one. I'm happy with this as is.

spowelljr (changes requested):
I think it would be good to implement a small integration test to make sure that the JSON output is acting correctly. ie. Start a cluster, check that the correct cluster has `Active: true`, then change the active `minikube profile <profile>` and then check the JSON output again and make sure the `Active: true` changed.

vharsh:
Okay, I think I've understood what I need to do.
I've got to create another file and create two minikube clusters with different profiles and check them out, I'll go ahead with this approach for now.

test/integration/minikube_profile_test.go
```diff
@@ -0,0 +1,52 @@
...
+// profileJSON is the output of `minikube profile list -ojson`
+type profileJSON struct {
+	Valid   []config.Profile `json:"valid"`
+	Invalid []config.Profile `json:"invalid"`
+}
+
+func TestMinikubeProfile(t *testing.T) {
+	// 1. Setup two minikube cluster profiles
+	profiles := []string{UniqueProfileName("first"), UniqueProfileName("second")}
+	ctx, cancel := context.WithTimeout(context.Background(), Minutes(40))
```
spowelljr, line 21: 40 minutes is too high of a timeout for this sort of test, I think 10 minutes would be more appropriate

test/integration/minikube_profile_test.go
```diff
@@ -0,0 +1,52 @@
...
+	for _, p := range profiles {
+		defer CleanupWithLogs(t, p, cancel)
+		_, err := Run(t, exec.CommandContext(ctx, Target(), "start", "-p", p))
+		if err != nil {
+			t.Errorf("test pre-condition failed: minikube start -p " + p + " failed: " + err.Error())
```
spowelljr, line 26:
```suggestion
			t.Fatalf("test pre-condition failed: minikube start -p " + p + " failed: " + err.Error())
```
If the start fails we should kill the test, as it's impossible for the test to pass at this point

test/integration/minikube_profile_test.go
```diff
@@ -0,0 +1,52 @@
...
+		err = json.NewDecoder(r.Stdout).Decode(&profile)
+		if err != nil {
+			t.Fatalf("minikube profile list -ojson failed with error: %v\n", err.Error())
+			return
+		}
+		// 3. Assert minikube profile is set to the correct profile in JSON
+		for _, s := range profile.Valid {
+			if s.Name == p && !s.Active {
+				t.Fatalf("minikube profile %s is not active\n", p)
+				return
```
spowelljr, line 48: Don't need return's after `Fatal`

test/integration/minikube_profile_test.go
```diff
@@ -0,0 +1,52 @@
...
+		if err != nil {
+			t.Errorf("minikube profile %s failed with error: %v\n", p, err.Error())
+			return
+		}
+		r, err := Run(t, exec.CommandContext(ctx, Target(), "profile", "list", "-ojson"))
+		var profile profileJSON
+		err = json.NewDecoder(r.Stdout).Decode(&profile)
+		if err != nil {
+			t.Fatalf("minikube profile list -ojson failed with error: %v\n", err.Error())
+			return
```
spowelljr, line 42: Don't need return's after Fatal

test/integration/minikube_profile_test.go
```diff
@@ -0,0 +1,52 @@
...
+		if err != nil {
+			t.Errorf("test pre-condition failed: minikube start -p " + p + " failed: " + err.Error())
+			return
+		}
+	}
+	// 2. Change minikube profile
+	for _, p := range profiles {
+		_, err := Run(t, exec.CommandContext(ctx, Target(), "profile", p))
+		if err != nil {
+			t.Errorf("minikube profile %s failed with error: %v\n", p, err.Error())
```
spowelljr, line 34: If this fails we should probably fatal as well

spowelljr (changes requested):

test/integration/minikube_profile_test.go
vharsh: Right, because it'd the fatalf would internally call `panic`

test/integration/minikube_profile_test.go
```diff
@@ -0,0 +1,53 @@
...
+	// 1. Setup two minikube cluster profiles
+	profiles := []string{UniqueProfileName("first"), UniqueProfileName("second")}
+	ctx, cancel := context.WithTimeout(context.Background(), Minutes(10))
+	// TODO(@vharsh): Figure out why go vet complains when this is moved into a loop
+	defer CleanupWithLogs(t, profiles[0], cancel)
+	defer CleanupWithLogs(t, profiles[1], cancel)
+	for _, p := range profiles {
+		_, err := Run(t, exec.CommandContext(ctx, Target(), "start", "-p", p))
+		if err != nil {
+			t.Fatalf("test pre-condition failed: minikube start -p " + p + " failed: " + err.Error())
```
spowelljr, line 28:
```suggestion
			t.Fatalf("test pre-condition failed. args %q: %v", rr.Command(), err)
```
This is a good format for output test failures that we use in some other places. Would be nice to keep the consistency. Use this format for the ones below as well if you could, thanks.

test/integration/minikube_profile_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+func TestMinikubeProfile(t *testing.T) {
+	// 1. Setup two minikube cluster profiles
+	profiles := []string{UniqueProfileName("first"), UniqueProfileName("second")}
+	ctx, cancel := context.WithTimeout(context.Background(), Minutes(10))
+	// TODO(@vharsh): Figure out why go vet complains when this is moved into a loop
+	defer CleanupWithLogs(t, profiles[0], cancel)
+	defer CleanupWithLogs(t, profiles[1], cancel)
+	for _, p := range profiles {
+		_, err := Run(t, exec.CommandContext(ctx, Target(), "start", "-p", p))
```
spowelljr, line 26:
```suggestion
		rr, err := Run(t, exec.CommandContext(ctx, Target(), "start", "-p", p))
```

test/integration/minikube_profile_test.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+		var profile profileJSON
+		err = json.NewDecoder(r.Stdout).Decode(&profile)
+		if err != nil {
+			t.Fatalf("minikube profile list -ojson failed with error: %v\n", err.Error())
+		}
+		// 3. Assert minikube profile is set to the correct profile in JSON
+		for _, s := range profile.Valid {
+			if s.Name == p && !s.Active {
+				t.Fatalf("minikube profile %s is not active\n", p)
```
spowelljr, line 49:
```suggestion
				t.Errorf("minikube profile %s is not active\n", p)
```
This one is an outcome vs a setup, so it's fine to make this an `Error` instead of a `Fatal`. And makes more sense with the above comment, that way we can loop through all the profiles and see all of the failures instead of just quitting on the first one.

test/integration/minikube_profile_test.go
```diff
@@ -0,0 +1,53 @@
...
+			t.Fatalf("minikube profile list -ojson failed with error: %v\n", err.Error())
+		}
+		var profile profileJSON
+		err = json.NewDecoder(r.Stdout).Decode(&profile)
+		if err != nil {
+			t.Fatalf("minikube profile list -ojson failed with error: %v\n", err.Error())
+		}
+		// 3. Assert minikube profile is set to the correct profile in JSON
+		for _, s := range profile.Valid {
+			if s.Name == p && !s.Active {
```
spowelljr, line 48: We should also check for the opposite case, we should make sure profiles that shouldn't be active aren't set as active.

spowelljr (changes requested):

test/integration/minikube_profile_test.go
```diff
@@ -0,0 +1,53 @@
+package integration
```
spowelljr, line 1: This file is missing copyright boilerplate, you can copy it from another .go file and just change the date to 2022.

test/integration/minikube_profile_test.go
```diff
@@ -0,0 +1,53 @@
...
+	Invalid []config.Profile `json:"invalid"`
+}
+
+func TestMinikubeProfile(t *testing.T) {
+	// 1. Setup two minikube cluster profiles
+	profiles := []string{UniqueProfileName("first"), UniqueProfileName("second")}
+	ctx, cancel := context.WithTimeout(context.Background(), Minutes(10))
+	// TODO(@vharsh): Figure out why go vet complains when this is moved into a loop
+	defer CleanupWithLogs(t, profiles[0], cancel)
+	defer CleanupWithLogs(t, profiles[1], cancel)
```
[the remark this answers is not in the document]
spowelljr: My guess is that it's assuming that `profiles` could be empty (even though that's not possible) and that cancel would never be called. Could maybe try doing `defer cancel.Cancel()` right after `ctx, cancel := context.WithTimeout` and then putting `CleanupWithLogs` back in the `for` loop.

test/integration/minikube_profile_test.go
```diff
@@ -0,0 +1,71 @@
...
+			t.Fatalf("cmd: %s failed with error: %v\n", rr.Command(), err.Error())
+		}
+		r, err := Run(t, exec.CommandContext(ctx, Target(), "profile", "list", "-ojson"))
+		if err != nil {
+			t.Fatalf("cmd: %s failed with error: %v\n", r.Command(), err.Error())
+		}
+		var profile profileJSON
+		err = json.NewDecoder(r.Stdout).Decode(&profile)
+		if err != nil {
+			t.Fatalf("unmarshalling %s cmd output failed with error: %v\n", r.Command(), err.Error())
```
spowelljr, line 60:
```suggestion
			t.Fatalf("unmarshalling %s cmd output failed with error: %v\n", r.Command(), err)
```
nit: with `%v` it will call `.Error()` under the hood for us

test/integration/minikube_profile_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+func TestMinikubeProfile(t *testing.T) {
+	// 1. Setup two minikube cluster profiles
+	profiles := [2]string{UniqueProfileName("first"), UniqueProfileName("second")}
+	ctx, cancel := context.WithTimeout(context.Background(), Minutes(10))
+	// TODO(@vharsh): Figure out why go vet complains when this is moved into a loop
+	defer CleanupWithLogs(t, profiles[0], cancel)
+	defer CleanupWithLogs(t, profiles[1], cancel)
+	for _, p := range profiles {
+		rr, err := Run(t, exec.CommandContext(ctx, Target(), "start", "-p", p))
```
spowelljr, line 42:
```suggestion
		rr, err := Run(t, exec.CommandContext(ctx, Target(), "start", "-p", p), StartArgs()...)
```

vharsh: Okay, I believe `StartArgs()` are the things which are making the E2E run with different minikube-drivers, I'll just add `StartArgs()` call to all `minikube start` operations in my test.

spowelljr (approved):

Issue #14091, opened 2022-05-03
GUI: Support starting minikube with HyperKit


spowelljr:
When downloading a new `docker-machine-driver-hyperkit` permissions how to be set on the file that require `sudo`. However, the GUI doesn't have `sudo` permission, so a flow to support setting the permissions with the GUI will need to be designed and implmented.

PR #14092, opened 2022-05-03
GUI: Support setting permission on HyperKit machine driver

spowelljr:
Closes https://github.com/kubernetes/minikube/issues/14091

**How It Works:**
If a start fails on Mac, it checks the error message to see if the cause of the error was due to the `docker-machine-driver-hyperkit` not having the proper permissions set. If this is the case we notify the user of the issue via the message box shown below and once they click OK a terminal instance is automatically opened with the command that will apply the required permissions to the file, the user just has to enter their password to execute the command. After they enter the command the terminal closes and the start retries again.

Screen Shot 2022-05-04 at 3 10 06 PM
Screen Shot 2022-05-03 at 11 23 15 AM

### DIFF
--- a/gui/window.cpp
+++ b/gui/window.cpp
@@ -323,11 +323,7 @@ void Window::startMinikube(QStringList moreArgs)
     args << moreArgs;
     bool success = sendMinikubeCommand(args, text);
 #if __APPLE__
-    if (text.contains("docker-machine-driver-hyperkit needs to run with elevated permissions")) {
-        showHyperKitMessage();
-        hyperkitPermission();
-        success = sendMinikubeCommand(args, text);
-    }
+    hyperkitPermissionFix(args, text);
 #endif
     updateClustersTable();
     if (success) {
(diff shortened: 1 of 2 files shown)


gui/window.cpp
```diff
@@ -931,6 +936,39 @@ void Window::sshConsole()
...
+    process->waitForFinished(-1);
+}
+
+void Window::showHyperKitMessage()
+{
+    QMessageBox dialog;
+    dialog.setWindowTitle(tr("HyperKit Permissions Required"));
+    dialog.setWindowIcon(*trayIconIcon);
+    dialog.setModal(true);
+    dialog.setText(tr(
```
medyagh, line 963: "The HyperKit driver requires a one-time sudo permission, If you like to proceed, Press OK and then enter your password in the terminal the start will resume after."

medyagh: there should be a cancel button to not continue.

medyagh: there should be a cancel button to not continue.

medyagh (changes requested):

gui/window.cpp
spowelljr: Message updated and cancel button added.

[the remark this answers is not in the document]
spowelljr: This was included in the previous commit

PR #14098, opened 2022-05-04
Add cleanup and reboot to preload generation

klaases:
We should install a cron to automatically clean up and reboot the preload generation server.

---
Background:
We recently encountered a service failure in the preload generation server due to lack of disk space. The /tmp directory contained over 350 GB of data that was no longer needed.

By cleaning up, we aim to ensure that our services perform with additional relaibility.


hack/jenkins/preload_generation.sh
```diff
@@ -20,6 +20,9 @@
 
 set -eux -o pipefail
 
+mkdir -p cron && gsutil -qm rsync "gs://minikube-builds/${MINIKUBE_LOCATION}/cron" cron || echo "FAILED TO GET CRON FILES"
```
spowelljr, line 23:
```suggestion
mkdir -p cron && gsutil -qm rsync "gs://minikube-builds/master/cron" cron || echo "FAILED TO GET CRON FILES"
```

Unless you updated the job to pass in `MINIKUBE_LOCATION`.

spowelljr (changes requested):

spowelljr (approved):

PR #14103, opened 2022-05-05
readability improvements for addons.go

klaases:
Found some coding errors while studying the addons.go file while working on the GCP-AUTH task.

1. Directly return on error.

2. Fix inefficient map merge algorithm.

3. Fix: Variable name vs. type
_The compiler always knows the type of a variable, and in most cases it is also clear to the reader what type a variable is by how it is used. It is only necessary to clarify the type of a variable if its value appears twice in the same scope._

### DIFF
--- a/pkg/minikube/assets/addons.go
+++ b/pkg/minikube/assets/addons.go
@@ -781,7 +781,7 @@ func filterKeySpace(keySpace map[string]string, target map[string]string) map[st
 	return result
 }
 
-// overrideDefaults creates a copy of `def` where `overrideMap` replaces any of its values that `overrideMap` contains.
+// overrideDefaults creates a copy of `def` where `override` replaces any of its values that `override` contains.
 func overrideDefaults(def, override map[string]string) map[string]string {
 	return mergeMaps(def, filterKeySpace(def, override))
 }
--- a/pkg/minikube/assets/addons.go
+++ b/pkg/minikube/assets/addons.go
@@ -787,7 +787,7 @@ func overrideDefaults(def, override map[string]string) map[string]string {
 }
 
 // SelectAndPersistImages selects which images to use based on addon default images, previously persisted images, and newly requested images - which are then persisted for future enables.
-func SelectAndPersistImages(addon *Addon, cc *config.ClusterConfig) (images, customRegistries map[string]string, err error) {
+func SelectAndPersistImages(addon *Addon, cc *config.ClusterConfig) (images, customRegistries map[string]string, _ error) {
 	addonDefaultImages := addon.Images
 	if addonDefaultImages == nil {
 		addonDefaultImages = make(map[string]string)
--- a/pkg/minikube/assets/addons.go
+++ b/pkg/minikube/assets/addons.go
@@ -836,6 +836,7 @@ func SelectAndPersistImages(addon *Addon, cc *config.ClusterConfig) (images, cus
 	// If images or registries were specified, save the config afterward.
 	if viper.IsSet(config.AddonImages) || viper.IsSet(config.AddonRegistries) {
 		// Since these values are only set when a user enables an addon, it is safe to refer to the profile name.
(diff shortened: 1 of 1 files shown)


pkg/minikube/assets/addons.go
```diff
@@ -707,32 +707,28 @@ func parseMapString(str string) map[string]string {
...
 			result[name] = value
 		}
 	}
 	return result
 }
 
-// overrideDefaults creates a copy of `defaultMap` where `overrideMap` replaces any of its values that `overrideMap` contains.
-func overrideDefaults(defaultMap, overrideMap map[string]string) map[string]string {
-	return mergeMaps(defaultMap, filterKeySpace(defaultMap, overrideMap))
+// overrideDefaults creates a copy of `def` where `overrideMap` replaces any of its values that `overrideMap` contains.
```
spowelljr, line 729:
```suggestion
// overrideDefaults creates a copy of `def` where `override` replaces any of its values that `override` contains.
```

pkg/minikube/assets/addons.go
```diff
@@ -782,22 +778,21 @@ func SelectAndPersistImages(addon *Addon, cc *config.ClusterConfig) (images, cus
...
-	err = nil
 	// If images or registries were specified, save the config afterward.
 	if viper.IsSet(config.AddonImages) || viper.IsSet(config.AddonRegistries) {
 		// Since these values are only set when a user enables an addon, it is safe to refer to the profile name.
-		err = config.Write(viper.GetString(config.ProfileName), cc)
 		// Whether err is nil or not we still return here.
+		return images, customRegistries, config.Write(viper.GetString(config.ProfileName), cc)
 	}
-	return images, customRegistries, err
+	return images, customRegistries, nil
```
spowelljr, line 787: Since we removed the `err` variable we can also remove the variable name from the return signature on L735.

spowelljr (changes requested):

pkg/minikube/assets/addons.go
klaases: Yes, set variable name for error to _, rather than remove altogether since we want to keep images, customRegistries.

pkg/minikube/assets/addons.go
```diff
@@ -762,36 +762,32 @@ func parseMapString(str string) map[string]string {
...
-	result := make(map[string]string)
-	for name, value := range sourceMap {
-		result[name] = value
-	}
-	for name, value := range overrideMap {
-		result[name] = value
+// mergeMaps returns a map with the union of `source` and `override` where collisions take the value of `override`.
+func mergeMaps(source, override map[string]string) map[string]string {
+	for k, v := range override {
+		source[k] = v
```
spowelljr, line 768: This is changing the functionality of the function.

Before, both of the incoming maps were left untouched and a new map was created, but this is modifying `source`.

Before:
```
source := map[string]string{"123": "abc", "345": "def"}
override := map[string]string{"123": "qwe", "678": "ghj"}
result := mergeMaps(source, override)
fmt.Printf("%+v\n", source)
// map[123:abc 345:def] (untouched)
```

After:
```
source := map[string]string{"123": "abc", "345": "def"}
override := map[string]string{"123": "qwe", "678": "ghj"}
result := mergeMaps(source, override)
fmt.Printf("%+v\n", source)
// map[123:qwe 345:def 678:ghj] (modified)
```

spowelljr (changes requested):

pkg/minikube/assets/addons.go
klaases: Thanks for the helpful feedback. I see that Go is actually passing the pointer to `soruce` through to `mergeMaps()` which is fine, but this leads to the unintended consequence of potentially modifying`source`, I reverted the change and cleaned up the function.

Created a Go Playground to confirm: https://go.dev/play/p/wT5ltyK1KHH

spowelljr (approved):

PR #14111, opened 2022-05-06
Fix typo on kong-ingress.md

Ileriayo:


site/content/en/docs/handbook/addons/kong-ingress.md
```diff
@@ -80,7 +80,7 @@ $ curl -v localhost
 ## Creating Ingress object
 
 Let's create a service.
-As an example, we use `tyoe-ExternalName` to point to https://httpbin.org
+As an example, we use `type-ExternalName` to point to https://httpbin.org
```
spowelljr, line 83: Is it also supposed to be an `=` between instead of a `-`?

I don't know for sure, but I saw this in the Kong docs

https://docs.konghq.com/kubernetes-ingress-controller/2.0.x/guides/using-external-service/#:~:text=a%20Kubernetes%20Service-,type%3DExternalName,-using%20the%20hostname

Ileriayo: This makes sense. It should be `=` rather than `-`.

Ileriayo: I've made the change.

sharifelgamal (approved):
Thank you for fixing our documentation!

Issue #12842, opened 2021-11-02
minikube 1.23: Your cgroup does not allow setting memory


cmanzur:
**Steps to reproduce the issue:** 

1. `minikube start --memory=8192 --cpus 2 --kubernetes-version=v1.22.1`
```
😄  minikube v1.23.2 on Ubuntu 20.04
✨  Automatically selected the docker driver. Other choices: virtualbox, ssh
❗  Your cgroup does not allow setting memory.
    ▪ More information: https://docs.docker.com/engine/install/linux-postinstall/#your-kernel-does-not-support-cgroup-swap-limit-capabilities
👍  Starting control plane node minikube in cluster minikube
🚜  Pulling base image ...
🔥  Creating docker container (CPUs=2, Memory=8192MB) ...
🐳  Preparing Kubernetes v1.22.1 on Docker 20.10.8 ...
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: storage-provisioner, default-storageclass
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default
```

- minikube 1.23.2
- docker 20.10.5
- ubuntu 20.04.2

```
docker stats

CONTAINER ID   NAME       CPU %     MEM USAGE / LIMIT     MEM %     NET I/O         BLOCK I/O    PIDS
4ec4a910c4e5   minikube   14.05%    573.4MiB / 62.74GiB   0.89%     191kB / 294kB   0B / 229kB   323
```

As you can see, there's NO memory limit.

```
grep GRUB_CMDLINE_LINUX /etc/default/grub

GRUB_CMDLINE_LINUX_DEFAULT="quiet splash"
GRUB_CMDLINE_LINUX="cgroup_enable=memory swapaccount=1"
```
Of course, I run `sudo update-grub2` without errors and reboot the PC.

medyagh:
@cmanzur this is a limitation of your ubuntu's cgroups as minikube detected and informed you 

```
✨  Automatically selected the docker driver. Other choices: virtualbox, ssh
❗  Your cgroup does not allow setting memory.
    ▪ More information: https://docs.docker.com/engine/install/linux-postinstall/#your-kernel-does-not-support-cgroup-swap-limit-capabilities

```

[...]

PR #14115, opened 2022-05-08
Fix: "Your cgroup does not allow setting memory"

kadern0:
Fixes issue #12842 

This fix seems to be working, at least for me and a few others (please see the discussion within the issue URL). So far Ubuntu, Arch and Pop!_OS are working but additional testing would be good.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -122,7 +122,9 @@ require (
 	github.com/containerd/cgroups v1.0.1 // indirect
 	github.com/containerd/containerd v1.5.2 // indirect
 	github.com/containerd/stargz-snapshotter/estargz v0.7.0 // indirect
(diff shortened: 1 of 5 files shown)


kadern0:
I cannot find anywhere this function is being used `func IsCgroup2UnifiedMode()`. Maybe it could be removed.

kadern0:
@medyagh , thanks for merging. I think #12842 can be closed.  What do you think about removing the function `IsCgroup2UnifiedMode()`?

spowelljr:
Hi @kadern0, if that function is not used anywhere then feel free to create a PR removing it. I can close #12842, just as a heads up for next time, if you put `Fixes #12842` instead of `Fixes issue #12842` it will auto close that issue on PR merge.

PR #14152, opened 2022-05-12
Upgrade Docker, from 20.10.14 to 20.10.15

afbjorklund:
* runc v1.1.1

* containerd v1.6.4

https://docs.docker.com/engine/release-notes/#201015


afbjorklund:
Apparently there was a 20.10.16 in stealth mode, as well

sharifelgamal (approved):

PR #14158, opened 2022-05-12
fix: removed unused code

kadern0:

After merging #14115 this code can be removed. It should probably have been done in the previous PR itself, my apologies for the extra work.


spowelljr (approved):
Thanks for removing this!

PR #14194, opened 2022-05-18
GUI: fix appearing as non-responsive while commands running

spowelljr:


sharifelgamal (approved):
massive refactor, nice work

PR #14230, opened 2022-05-26
kicbase: don't build cri-dockerd from source

sharifelgamal:
kicbase was taking over 4 hours to build, because we were building both auto-pause and cri-dockerd from source for 5 different architectures, 4 of which use qemu for cross-compilation. 

This PR grabs pre-compiled cri-dockerd binaries currently stored in GCS and copies them into the images. This takes the kicbase build back under 2 hours, which is where it was before we were forced to add cri-dockerd.

### DIFF
--- a/deploy/kicbase/Dockerfile
+++ b/deploy/kicbase/Dockerfile
@@ -18,22 +18,30 @@
 # https://systemd.io/CONTAINER_INTERFACE/
 
 
-# multi-tage docker build so we can build auto-pause for arm64
-FROM golang:1.17
+# multi-stage docker build so we can build auto-pause for arm64
+FROM golang:1.17 as auto-pause
 WORKDIR /src
-# becaue auto-pause binary depends on minikube's code we need to pass the whole source code as the context
-ADD . .
+# auto-pause depends on core minikube code so we need to pass the whole source code as the context
+# copy in the minimal amount of source code possible
+COPY pkg/ ./pkg
+COPY cmd/ ./cmd
+COPY deploy/addons ./deploy/addons
+COPY translations/ ./translations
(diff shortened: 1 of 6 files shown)


deploy/kicbase/Dockerfile
```diff
@@ -162,9 +170,9 @@ RUN export ARCH=$(dpkg --print-architecture | sed 's/ppc64el/ppc64le/') && \
     curl -LO https://downloadcontent.opensuse.org/repositories/devel:kubic:libcontainers:stable/xUbuntu_20.04/Release.key && \
     apt-key add - < Release.key && \
     if [ "$ARCH" != "ppc64le" ]; then \
-    	clean-install containers-common catatonit conmon containernetworking-plugins cri-tools podman-plugins crun; \
+       clean-install containers-common conmon containernetworking-plugins cri-tools crun podman-plugins catatonit; \
```
afbjorklund, line 173: Don't really see how this change is related to cri-dockerd ?

afbjorklund:
Were is the cri-dockerd being built now ? i.e. what uploads to that GCS bucket

deploy/kicbase/Dockerfile
sharifelgamal: it's not, i was debugging other issues related to the openSUSE repo and remove then added back some packages, this can be reverted.

sharifelgamal:
> Were is the cri-dockerd being built now ? i.e. what uploads to that GCS bucket

Good question. I did it manually at first just for testing, but now I've added a script so we can add some automation to it as well.

sharifelgamal:
For reference, this kicbase build took 71 minutes.

spowelljr (approved):

PR #14242, opened 2022-05-29
correct GPU flag in documentation

peterbecich:
Using `minikube version: v1.25.2`, the correct flag is:
```
--kvm-gpu=false: Enable experimental NVIDIA GPU support in minikube
```


spowelljr (approved):
Thanks for the fix!

PR #14245, opened 2022-05-30
add bind address option for cmd tunnel

### LINKED ISSUE #14234: Add bind address with `minikube tunnel` command
The `minikube tunnel` command should have an option for `bind-address'.

`minikube tunnel` bind service and ingress with same port on localhost interface, in develop envrionment, such port sometimes has been bond by other program, especially for `80:http` and `443:https`.
I used to make interface alias so that I can seperate network for different work, for example:
```
ifconfig en0 alias 10.0.0.101 netmask 0xFFFFFFFF up
ifconfig en0 alias 10.0.0.102 netmask 0xFFFFFFFF up
```
I use `10.0.0.101` interface for all my works on product-A, and `10.0.0.102` for product-B. So if we have `bind-address` option, we can tunnel k8s and bind on a specify interface, that will be usefull.

After dig into codes I notice tunnel is generated by `ssh -L port:host:hostport`, I think that just replace it with `ssh -L bind_address:port:host:hostport` if we provide `bind-address` option, and all things are done.

I have implement one on my branch here, but not well tested:
31843939d8b8b7ce09e675c4754a17a662fa8807
hope that is useful.

Thanks,
Simon Ren

te-simonren:
the PR added an `bind-address` option for `minikube tunnel`, so we can choose to one interface to bind with.
that's useful as I described in #14234 

`minikube tunnel --help` will show as follow:
```
➜  ~ minikube tunnel --help
tunnel creates a route to services deployed with type LoadBalancer and sets their Ingress to their ClusterIP. for a
detailed example see https://minikube.sigs.k8s.io/docs/tasks/loadbalancer

Options:
    --bind-address='':
	set tunnel bind address, empty or `*' indicates that tunnel should be available for all interfaces

    -c, --cleanup=true:
	call with cleanup=true to remove old tunnels

Usage:
  minikube tunnel [flags] [options]

```

fixes #14234

### DIFF
--- a/cmd/minikube/cmd/tunnel.go
+++ b/cmd/minikube/cmd/tunnel.go
@@ -120,5 +120,5 @@ func outputTunnelStarted() {
 
 func init() {
 	tunnelCmd.Flags().BoolVarP(&cleanup, "cleanup", "c", true, "call with cleanup=true to remove old tunnels")
-	tunnelCmd.Flags().StringVar(&bindAddress, "bind-address", "", "set tunnel bind address, empty or `*' indicates that tunnel should be available for all interfaces")
+	tunnelCmd.Flags().StringVar(&bindAddress, "bind-address", "", "set tunnel bind address, empty or '*' indicates the tunnel should be available for all interfaces")
 }
--- a/cmd/minikube/cmd/tunnel.go
+++ b/cmd/minikube/cmd/tunnel.go
@@ -41,6 +41,7 @@ import (
 )
 
 var cleanup bool
+var bindAddress string
 
 // tunnelCmd represents the tunnel command
 var tunnelCmd = &cobra.Command{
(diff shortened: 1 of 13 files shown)


cmd/minikube/cmd/tunnel.go
```diff
@@ -119,4 +120,5 @@ func outputTunnelStarted() {
 
 func init() {
 	tunnelCmd.Flags().BoolVarP(&cleanup, "cleanup", "c", true, "call with cleanup=true to remove old tunnels")
+	tunnelCmd.Flags().StringVar(&bindAddress, "bind-address", "", "set tunnel bind address, empty or `*' indicates that tunnel should be available for all interfaces")
```
spowelljr, line 123:
```suggestion
	tunnelCmd.Flags().StringVar(&bindAddress, "bind-address", "", "set tunnel bind address, empty or '*' indicates the tunnel should be available for all interfaces")
```
Matching quotes

spowelljr (changes requested):
Could you run `make generate-docs` to update the documentation with your new flag and then push those changes up, thanks!

cmd/minikube/cmd/tunnel.go
te-simonren: Thank you for your suggestion, I have changed the quotes and generated docs via 2813810.

PR #14267, opened 2022-06-01
CI: Automate updating kubectl version

spowelljr:
- Dynamically pull the latest kubectl version
- Removed hardcoded Go version


spowelljr:
> for consistency we should add this to Jenkins scripts too (maybe it is in common)

Jenkins scripts already do this

sharifelgamal (approved):

PR #14275, opened 2022-06-04
Fix compilation of minikube gui on qt5

afbjorklund:
commandrunner.cpp:133:20: error: invalid use of incomplete type ‘class QDebug’

updater.cpp:52:62: error: ‘endl’ is not a member of ‘Qt’

### DIFF
--- a/gui/commandrunner.cpp
+++ b/gui/commandrunner.cpp
@@ -5,6 +5,7 @@
 #include <QJsonArray>
 #include <QJsonDocument>
 #include <QJsonParseError>
+#include <QDebug>
 
 CommandRunner::CommandRunner(QDialog *parent)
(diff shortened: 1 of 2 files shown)


spowelljr:
Just curious what version of Qt you're using, I've been using `5.15.2` and didn't see those errors.

spowelljr (approved):

afbjorklund:
> Just curious what version of Qt you're using, I've been using `5.15.2` and didn't see those errors.

It was just the vanilla version from Ubuntu 20.04, I have some Dockerfile if interested

`qt5-default/focal-updates,focal-security,now 5.12.8+dfsg-0ubuntu2.1 amd64 [installed]`

Issue #14146, opened 2022-05-11
Known QEMU2 Driver Issues


sharifelgamal:
Now that #13639 is (soon to be) merged, we need a place to aggregate all the known issues so we can eventually graduate it from experimental to recommended driver.

[ ] containerd runtime is broken (permissions on the socket file)
[ ] crio runtime is broken (conmon installation in the arm64 ISO is busted)
[ ] `minikube service` (unimplemented)
[ ] `minikube tunnel` (unimplemented)
[ ] Full test suite on macOS arm64

afbjorklund:
Networking currently for "user" (slirp) only. Proper networking on Mac is possible, by using the VDE and Socket drivers from Lima. But not implemented in driver yet - should be easy to do, though

afbjorklund:
The current tunneling of the docker port is kinda gross, but it was blocked on upstream not allowing any changes...

* https://github.com/docker/machine/issues/3361

i.e. the docker port was hardcoded to 2376, so currently it does some kind of bait-and-switch on the URL instead

Now when the driver is forked, should remove that workaround - and add the tunneling of the minikube 8443 port.

afbjorklund:
Naming: once the dust settles, the "qemu2" driver should be aliased as the "qemu" driver - similar to how the "kvm2" driver eventually overtook the original "kvm" driver. The forks are more an historical fact, and machine upstream is now dead.

afbjorklund:
The QEMU2 driver is supposed to work on all three platforms, with hardware acceleration, but it needs more testing.
On Linux it uses KVM, on macOS it uses HVF and on Windows it uses WHPX. The qemu parameter is `-accel`.

```console
anders@ubuntu:~$ qemu-system-x86_64 -accel help
Accelerators supported in QEMU binary:
tcg
kvm
anders@ubuntu:~$ qemu-system-aarch64 -accel help
Accelerators supported in QEMU binary:
tcg

```

afbjorklund:
@sharifelgamal : issue with workaround for lots of memory: #14273

Should be possible to remove the workaround, on recent macOS

mprimeaux:
Where can I find the `kube-registry-proxy` image that backs the registry add-on?

mprimeaux:
I'm just familiarizing myself with the Minikube code base and would like to focus on one of the items in this issue (ARM supported registry). 

[...]

PR #14291, opened 2022-06-07
qemu: only set highmem=off for darwin if qemu version is below 7.0 or memory is below 3GB

### LINKED ISSUE #14273: ERROR: qemu-system-aarch64: Addressing limited to 32 bits
### What Happened?

When starting minikube [v1.26.0-beta.1](https://github.com/kubernetes/minikube/releases/tag/v1.26.0-beta.1) with the following configuration, it fails to start.

```yaml
- cpus: 8
- driver: qemu2
- memory: 32768
```

Log file is attached though the error indicates a 32-bit memory address limitation.  My host configuration is...
* OS: macOS Monterey (12.4)
* CPU: Apple M1 Ultra
* Memory: 128 GB

**NOTE:** I wasn't able to select `qemu2` as the driver below on this form.

### Attach the log file

[logs.txt](https://github.com/kubernetes/minikube/files/8835280/logs.txt)

### Operating System

macOS (Default)

### Driver

_No response_

sharifelgamal:
qemu 7.0.0 fixes an issue that required highmem=off to be specified when using hardware acceleration in aarch64, so we should only specify it for older versions of qemu or if there isn't enough memory allocation to make highmem work.

Fixes #14273 

Related #14146


pkg/minikube/registry/drvs/qemu2/qemu2.go
```diff
@@ -122,9 +122,12 @@ func configure(cc config.ClusterConfig, n config.Node) (interface{}, error) {
 		qemuCPU = "cortex-a72"
 		// highmem=off needed, see https://patchwork.kernel.org/project/qemu-devel/patch/20201126215017.41156-9-agraf@csgraf.de/#23800615 for details
 		if runtime.GOOS == "darwin" {
-			qemuMachine = "virt,highmem=off"
```
mprimeaux, line 125: Should we use "+=" here also?

mprimeaux (commented):
Just a minor change but nothing to hold back the PR.

pkg/minikube/registry/drvs/qemu2/qemu2.go
sharifelgamal: yeah, new commit fixes this

mprimeaux (approved):
LGTM

spowelljr (approved):

PR #14294, opened 2022-06-07
fix minikube build

sharifelgamal:
We merged a new addon after having changed the format of the assets object. This fixes that. Also runs `go mod tidy`.


spowelljr (approved):
If the build succeeds looks good to me

PR #14323, opened 2022-06-11
Don't allow --control-plane=true for node add #14300

### LINKED ISSUE #14300: Don't allow `--control-plane=true` for `node add`
This isn't currently supported, the flag was put in for future proofing to make it easier if we do implement it.

If the user tries to use the flag we should output a message to the user saying the flag is not yet supported and the flag value should be ignored.

lakshkeswani:
Fixes #14300
output message for if the user tries to add a control panel to the node and avoid the crash
control panel feature for node add is not supported


cmd/minikube/cmd/node_add.go
```diff
@@ -51,6 +51,11 @@ var nodeAddCmd = &cobra.Command{
 		name := node.Name(len(cc.Nodes) + 1)
 
 		out.Step(style.Happy, "Adding node {{.name}} to cluster {{.cluster}}", out.V{"name": name, "cluster": cc.Name})
+		// for now control-plane feature is not supported
+		if !cp {
+			out.Step(style.Unsupported, "control-plane feature is not supported {{.name}} to {{.cluster}}!", out.V{"name": name, "cluster": cc.Name})
```
spowelljr, line 56:
```suggestion
			out.Step(style.Unsupported, "Adding a control-plane node is not yet supported, setting control-plane flag to false")
```

Could you also move this `if` statement above the `out.Step(style.Happy, "Adding node` line.

spowelljr (changes requested):

spowelljr:
Can you also edit your PR description to include `Fixes https://github.com/kubernetes/minikube/issues/14300`, this will auto close the issue when this PR is merged, thanks

cmd/minikube/cmd/node_add.go
```diff
@@ -50,8 +50,13 @@ var nodeAddCmd = &cobra.Command{
 
 		name := node.Name(len(cc.Nodes) + 1)
 
-		out.Step(style.Happy, "Adding node {{.name}} to cluster {{.cluster}}", out.V{"name": name, "cluster": cc.Name})
+		// for now control-plane feature is not supported
+		if !cp {
```
spowelljr, line 54: This should be if `cp` is true, right?

spowelljr (changes requested):

cmd/minikube/cmd/node_add.go
lakshkeswani: yes

knrt10:
@lakshkeswani Instead of creating a new commit [check for control panel](https://github.com/kubernetes/minikube/pull/14323/commits/b7c75c996735cd8375a4060fb91c23a224c0ed78). I would suggest you can edit your first commit with the change.

You can do the following steps.

- git rebase -i "commitID^" or git rebase -i HEAD~4
- On your editor change the commit you want to change i.e [Don't allow --control-plane=true for node add](https://github.com/kubernetes/minikube/pull/14323/commits/daf310b8fd5c2dbb346c40634491a4be8ef49308) from `pick` to `edit`. Save and exit editor
- Make your change to code.
- git add Filename
- git rebase --continue (this will generate the old commit)
- Save the commit
- git push origin lakshkeswani:master --force-with-lease

spowelljr (approved):
Thanks for the change, looks great!

spowelljr:
Actually one more thing, on line 97 could you add to the flag description that `This flag is currently unsupported.` And then run `make generate-docs` to update the site documentation, thanks!

lakshkeswani:
Generated the docs and squashed the commits, I hope this will get merged.
Thanks

PR #14337, opened 2022-06-14
build auto-pause outside kicbase dockerfile

sharifelgamal:
Building from source in a multiarch dockerfile is very slow, let's avoid that slowdown.


sharifelgamal:
23 minutes 😮

spowelljr (approved):

PR #14343, opened 2022-06-15
pr-bot: Fix first run taking longer

spowelljr:
**Problem:**
`minikube start --download-only` saves the kic image to `~/.minikube/cache`, however the first benchmark loads the cached kic image into Docker which takes 20+ seconds, which skews the start times for the PR-bot.

**Solution:**
Instead of doing `minikube start --download-only` before starting the benchmark, now doing a `minikube start` followed by a `minikube delete` which ensures the kic image is already in Docker for the first benchmark run.

Also doing `docker system prune -a --volumes -f` as we're not clearing out artifacts after and the machines are accumulating many images.


medyagh (approved):
good find ! I always wondered why it was like that

PR #14366, opened 2022-06-21
Modified a syntax error for a comment

lixin963:


lixin963:
Modified a syntax error for a comment

hack/benchmark/time-to-k8s/cpu.go
```diff
@@ -31,7 +31,7 @@ var fields = []string{"CPU Utilization(%)", "CPU Time(seconds)"}
 
 func cpuMarkdownTable(categories []plotter.Values, names []string) {
 
-	// categories row is the either cpu pct or time, col is process name
+	// categories row either is the  cpu pct or time, col is process name
```
spowelljr, line 34:
```suggestion
	// categories row is either cpu pct or time, col is process name
```

spowelljr (changes requested):

gui/tray.cpp
```diff
@@ -13,7 +13,7 @@ Tray::Tray(QIcon icon)
 
     connect(trayIcon, &QSystemTrayIcon::activated, this, &Tray::iconActivated);
 
-    minimizeAction = new QAction(tr("Mi&nimize"), this);
+    minimizeAction = new QAction(tr("&Minimize"), this);
```
spowelljr, line 16: The `&` should stay where it is, it means that `alt + I` will minimize it, this will move the command to `alt + M`

cmd/minikube/cmd/start.go
```diff
@@ -764,7 +764,7 @@ func validateSpecifiedDriver(existing *config.ClusterConfig) {
 	}
 
 	// hostDriver always returns original driver name even if an alias is used to start minikube.
-	// For all next start with alias needs to be check against the host driver aliases.
+	// For all next start with alias needs to be checked against the host driver aliases.
```
spowelljr, line 767:
```suggestion
	// All future starts with alias need to be checked against the host driver aliases.
```

spowelljr (changes requested):

gui/tray.cpp
lixin963: ok

PR #14377, opened 2022-06-22
Remove punctuation

jiasr:
Signed-off-by: jiasirui <jiasirui3@163.com>


spowelljr:
A serial comma is valid punctuation, I'm not sure why we should be removing this

PR #14390, opened 2022-06-22
Bump Kubernetes version default: v1.24.2 and latest: v1.24.2

spowelljr:


medyagh (approved):
lgtm when test are acceptable

PR #14437, opened 2022-06-27
Document .localhost usage

aude:
This is just a minor edit that adds links to some RFCs, providing an explanation for why `.localhost` doesn't resolve to `minikube ip` even if you try.


klaases (approved):
Hi @aude, thanks for taking the time to help update the documentation with additional information, approved.

spowelljr (approved):
Thanks for the PR!

PR #14461, opened 2022-06-28
Update build docs for different archs

### LINKED ISSUE #14459: Building ISO fails
### What Happened?

Trying to build the Minikube ISO [as per the docs](https://github.com/kubernetes/minikube/blob/474b52c02fc9f6b3e44dfd1c34fec8f065ab2d44/site/content/en/docs/contrib/building/iso.md?plain=1#L40) yields an error:

```
$ make out/minikube.iso
make: *** No rule to make target 'out/minikube.iso'.  Stop.
```

Looks like the rule was updated in https://github.com/kubernetes/minikube/commit/d92f42a6e459313fe22ca935bc5869cee05146e9 , so seems like the docs just need to be updated.

### Attach the log file

n/a

### Operating System

Redhat/Fedora

### Driver

N/A

psschwei:
Signed-off-by: Paul S. Schweigert <paulschw@us.ibm.com>

Fixes #14459

### DIFF
--- a/site/content/en/docs/contrib/building/iso.md
+++ b/site/content/en/docs/contrib/building/iso.md
@@ -35,19 +35,26 @@ cd minikube
 
 ### Building
 
+To build for x86
 ```shell
 $ make buildroot-image
-$ make out/minikube.iso
+$ make out/minikube-amd64.iso
+```
+
+To build for ARM
+```shell
+$ make buildroot-image
+$ make out/minikube-arm64.iso
 ```
 
 The build will occur inside a docker container. If you want to do this on
-baremetal, replace `make out/minikube.iso` with `IN_DOCKER=1 make out/minikube.iso`.
-The bootable ISO image will be available in `out/minikube.iso`.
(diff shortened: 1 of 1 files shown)


psschwei:
The [last line](https://github.com/kubernetes/minikube/pull/14461/files#diff-e20bcb0b65804fb7c7a6237fc16522f4d6b77514a62f2b6031a0fc2d59b501e8R109) may need an update also... I wasn't able to access the builds, so wasn't sure what the update actually should be. Happy to fix that if needed.

spowelljr:
> The [last line](https://github.com/kubernetes/minikube/pull/14461/files#diff-e20bcb0b65804fb7c7a6237fc16522f4d6b77514a62f2b6031a0fc2d59b501e8R109) may need an update also... I wasn't able to access the builds, so wasn't sure what the update actually should be. Happy to fix that if needed.

Thanks for the PR @psschwei, looking at the Makefile it looks like that line would need to be updated as well.
`$(BUILD_DIR)/minikube-arm64.iso;`
`$(BUILD_DIR)/minikube-amd64.iso;`

So your change looks good to me

PR #14473, opened 2022-06-30
Update location of published PR minikube isos

psschwei:
Signed-off-by: Paul S. Schweigert <paulschw@us.ibm.com>

Following up on a change missed in #14461

### DIFF
--- a/site/content/en/docs/contrib/building/iso.md
+++ b/site/content/en/docs/contrib/building/iso.md
@@ -99,4 +99,5 @@ We publish CI builds of minikube, built at every Pull Request. Builds are availa
 
 We also publish CI builds of minikube-iso, built at every Pull Request that touches deploy/iso/minikube-iso.  Builds are available at:
 
-- <https://storage.googleapis.com/minikube-builds/PR_NUMBER/minikube-testing.iso>
+- <https://storage.googleapis.com/minikube-builds/PR_NUMBER/minikube-amd64.iso>
+- <https://storage.googleapis.com/minikube-builds/PR_NUMBER/minikube-arm64.iso>


spowelljr:
Sorry for merging your previous PR before this change, I misunderstood what line you were referencing.

The path we store the ISOs in is actually different now, this seems to be old documentation that was missed.

Example:
`https://storage.googleapis.com/minikube-builds/iso/14420/minikube-v1.26.0-1656448385-14420-amd64.iso`

The problem is the filename is in the format `minikube-<VERSION>-<TIME_STAMP>-<PR_NUMBER>-<ARCH>.iso`

The time stamp is very arbitrary, so I'm not sure if there's a great way to represent it in our documentation.

psschwei:
I see... should I just remove that section then? 

Or is there a chance that a bot could add the links to the PR? If so, in that case I could update the docs to note that the link can be found in the PR...

spowelljr:
On a successful ISO build on a PR the minikube-bot commits changes to the PR that defaults the ISO version to the one just built. So if the user wants to use the change all they have to do is pull their branch to get the minikube-bot commit and then do `make` and then they can start the built binary (`./out/minikube`) with the ISO changes.

medyagh (approved):

psschwei:
Got it, so I'll just cut the "We also publish CI builds of minikube-iso..." part, leaving only the "We publish CI builds of minikube ..." piece

medyagh:
> https://storage.googleapis.com/minikube-builds/iso/14420/

maybe we could mention the ISO for the PR could be found in the PR comments by minikube-bot

psschwei:
> > https://storage.googleapis.com/minikube-builds/iso/14420/
> 
> maybe we could mention the ISO for the PR could be found in the PR comments by minikube-bot

Only issue with that is I get a `NoSuchKey` error when trying to access that link. I only was able to access via the full example path @spowelljr posted earlier.

psschwei:
That said, I'm happy to fix however you folks think best...

Issue #14410, opened 2022-06-24
unable to start minikube with docker without cri-dockerd


sudlo:
### What Happened?

I have installed minikube on aws ec2 instance followed all the steps.
Please find the logs attached for your reference 

### Attach the log file

root@ip-172-31-47-92:~# cat logs.txt
*
* ==> Audit <==
* |---------|-------------------------|----------|------|---------|---------------------|---------------------|
| Command |          Args           | Profile  | User | Version |     Start Time      |      End Time       |
|---------|-------------------------|----------|------|---------|---------------------|---------------------|
| start   | --vm-driver=none        | minikube | root | v1.26.0 | 24 Jun 22 03:57 UTC |                     |
| delete  |                         | minikube | root | v1.26.0 | 24 Jun 22 04:00 UTC | 24 Jun 22 04:00 UTC |
| start   | --driver=docker         | minikube | root | v1.26.0 | 24 Jun 22 04:00 UTC |                     |
| delete  |                         | minikube | root | v1.26.0 | 24 Jun 22 04:00 UTC | 24 Jun 22 04:00 UTC |
| start   |                         | minikube | root | v1.26.0 | 24 Jun 22 04:00 UTC |                     |
| start   | --force --driver=docker | minikube | root | v1.26.0 | 24 Jun 22 04:01 UTC |                     |
| start   | --driver=none           | minikube | root | v1.26.0 | 24 Jun 22 04:01 UTC |                     |
| start   |                         | minikube | root | v1.26.0 | 24 Jun 22 04:02 UTC |                     |
|---------|-------------------------|----------|------|---------|---------------------|---------------------|

*
* ==> Last Start <==
* Log file created at: 2022/06/24 04:02:49
Running on machine: ip-172-31-47-92
Binary: Built with gc go1.18.3 for linux/amd64
Log line format: [IWEF]mmdd hh:mm:ss.uuuuuu threadid file:line] msg
I0624 04:02:49.109566    6897 out.go:296] Setting OutFile to fd 1 ...
I0624 04:02:49.109700    6897 out.go:348] isatty.IsTerminal(1) = true
I0624 04:02:49.109705    6897 out.go:309] Setting ErrFile to fd 2...
I0624 04:02:49.109712    6897 out.go:348] isatty.IsTerminal(2) = true
I0624 04:02:49.109921    6897 root.go:329] Updating PATH: /root/.minikube/bin
[...]

PR #14483, opened 2022-06-30
Add solution message for when `cri-docker` is missing

spowelljr:
Related: https://github.com/kubernetes/minikube/issues/14410

Added solution messages for when cri-docker is missing.

One message is specifically for the none driver and the other is for the rest of the drivers.


medyagh (approved):

klaases (approved):
LGTM

PR #14498, opened 2022-07-03
add VerifiedMaintainer field to Addons and display a warning if empty - #14493

### LINKED ISSUE #14493: Add GitHub maintainer field to addons & warn for all addons missing a maintainer
We should add a GitHub maintainer field to the addons (below):
https://github.com/kubernetes/minikube/blob/044fe4c1fd2779b5ac3e4974556c3ff7b4d26792/pkg/minikube/assets/addons.go#L37-L47

An empty field should be added for each addon, and for each addon that has an empty field, when the user tries to enable the addon it should give a warning to the user that there's not a verified maintainer for the addon.

inifares23lab:
The Addon type now has another string Field: VerifiedMaintainer
The new field is added to the NewAddon function and to the Addons map.

When enabling an addon if the VerifiedMaintainer field is empty a warning message is displayed:
- ! The <ADDON_NAME> addon doesn't have a verified maintainer.

Since for now I just added an empty string for every addon, it will always display the warning when enabling

### DIFF
--- a/cmd/minikube/cmd/config/enable.go
+++ b/cmd/minikube/cmd/config/enable.go
@@ -23,6 +23,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"k8s.io/minikube/pkg/addons"
+	"k8s.io/minikube/pkg/minikube/assets"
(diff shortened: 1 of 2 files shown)


inifares23lab:
I signed it

Issue #14469, opened 2022-06-29
Deprecate `storage-provisioner-gluster` addon


akhal3d96:
### What Happened?

This image doesn't exist

https://github.com/kubernetes/minikube/blob/319dd493820e517410fb2faa5255f549942de757/pkg/minikube/assets/addons.go#L191

```
NAME                                       READY   STATUS             RESTARTS   AGE
glusterfile-provisioner-59c9647cb6-7tdgx   1/1     Running            0          12m
glusterfs-54gfm                            0/1     ImagePullBackOff   0          3m36s
heketi-77585cd657-4nkzx                    1/1     Running            0          12m

```

```
  Warning  Failed     3m27s (x4 over 5m57s)  kubelet            Failed to pull image "quay.io/nixpanic/glusterfs-server:pr_fake-disk@sha256:3c58ae9d4e2007758954879d3f4095533831eb757c64ca6a0e32d1fc53fb6034": rpc error: code = Unknown desc = Error response from daemon: unauthorized: access to the requested resource is not authorized

```

### Attach the log file

[log.txt](https://github.com/kubernetes/minikube/files/9012617/log.txt)

### Operating System

Ubuntu

### Driver

Docker

spowelljr:
It looks like the image repo that had that image is gone now. We're going to have to deprecate the addon unless we can get one of the maintainers of the addon to resurrect the image repo as I wasn't able to find an official replacement.

Madhu-1:
cc @nixpanic

nixpanic:
The location for the image was not intentionally removed, but it also isn't the right location. It seems I forgot to update it once [the PR was merged](https://github.com/gluster/gluster-containers/pull/95) in the [gluster-containers](https://github.com/gluster/gluster-containers) project.

The image `docker.io/gluster/gluster-centos:latest` would be more appropriate, although it seems that the Gluster Containers do not get updated (anymore?). I am happy to send a PR with the right image location. However, if you prefer to deprecate the addon, I understand that decision too.

spowelljr:
Hi @nixpanic, thanks for the update, I'm happy to accept a PR that updates the location of the image.

PR #14518, opened 2022-07-06
storage-provisioner-gluster: use the Gluster project's container-image

nixpanic:
Use docker.io/gluster/gluster-centos:latest instead of a temporary image
that had a change which needed to get merged. All required features are
included in the 'official' container-image.

Fixes: #14469

### DIFF
--- a/deploy/addons/aliyun_mirror.json
+++ b/deploy/addons/aliyun_mirror.json
@@ -24,7 +24,7 @@
   "k8s.gcr.io/sig-storage/csi-snapshotter": "registry.cn-hangzhou.aliyuncs.com/google_containers/csi-snapshotter",
   "k8s.gcr.io/sig-storage/csi-provisioner": "registry.cn-hangzhou.aliyuncs.com/google_containers/csi-provisioner",
   "registry": "registry.cn-hangzhou.aliyuncs.com/google_containers/registry",
-  "quay.io/nixpanic/glusterfs-server": "registry.cn-hangzhou.aliyuncs.com/google_containers/glusterfs-server",
+  "docker.io/gluster/gluster-centos": "registry.cn-hangzhou.aliyuncs.com/google_containers/glusterfs-server",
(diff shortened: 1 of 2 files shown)


deploy/addons/aliyun_mirror.json
```diff
@@ -24,7 +24,7 @@
   "k8s.gcr.io/sig-storage/csi-snapshotter": "registry.cn-hangzhou.aliyuncs.com/google_containers/csi-snapshotter",
   "k8s.gcr.io/sig-storage/csi-provisioner": "registry.cn-hangzhou.aliyuncs.com/google_containers/csi-provisioner",
   "registry": "registry.cn-hangzhou.aliyuncs.com/google_containers/registry",
-  "quay.io/nixpanic/glusterfs-server": "registry.cn-hangzhou.aliyuncs.com/google_containers/glusterfs-server",
+  "docker.io/gluster/gluster-centos": "registry.cn-hangzhou.aliyuncs.com/google_containers/glusterfs-server",
```
spowelljr, line 27:
```suggestion
  "gluster/gluster-centos": "registry.cn-hangzhou.aliyuncs.com/google_containers/glusterfs-server",
```
If it's from DockerHub we omit `docker.io`, it will work without, but for consistency sake could you mind removing it, thanks!

spowelljr (changes requested):

deploy/addons/aliyun_mirror.json
medyagh: it would be better to use docker.io (fully qualified name), but to keep this PR small and reverent, how about a follow up PR to change all of them to docker.io ?

spowelljr: Works for me

spowelljr: https://github.com/kubernetes/minikube/issues/14520

spowelljr (approved):

PR #14522, opened 2022-07-07
Fix Post installation tip for headlamp addon

yolossn:
For Kubernetes versions > 1.24 the tokens for service account are not created by default, this patch fixes the help command that gets printed after installation of the headlamp addon to generate a token.  

Signed-off-by: Santhosh Nagaraj S <santhosh@kinvolk.io>

### DIFF
--- a/cmd/minikube/cmd/config/enable.go
+++ b/cmd/minikube/cmd/config/enable.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/blang/semver/v4"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"k8s.io/minikube/pkg/addons"
@@ -29,6 +30,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/out"
 	"k8s.io/minikube/pkg/minikube/reason"
 	"k8s.io/minikube/pkg/minikube/style"
+	"k8s.io/minikube/pkg/util"
 )
 
 var addonsEnableCmd = &cobra.Command{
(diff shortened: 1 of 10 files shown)


yolossn:
@spowelljr Gentle Ping, Can you please take a look?

cmd/minikube/cmd/config/flags.go
```diff
@@ -25,3 +25,18 @@ import (
...
+	if err != nil {
+		return ""
+	}
+
+	for _, profile := range validProfiles {
+		if profile.Name == clusterProfile {
+			return profile.Config.KubernetesConfig.KubernetesVersion
+		}
+	}
+	return ""
```
spowelljr, line 41:
```suggestion
	_, cc := mustload.Partial(clusterProfile)
	return cc.KubernetesConfig.KubernetesVersion
```
This should do the same thing in an easier way.

spowelljr (changes requested):
Once you make the changes could you run `make generate-docs` to update the translations and such, thanks

cmd/minikube/cmd/config/enable.go
```diff
@@ -75,16 +76,28 @@ var addonsEnableCmd = &cobra.Command{
...
+			getSvcAccountToken := `export SECRET=$(kubectl get secrets --namespace headlamp -o custom-columns=":metadata.name" | grep "headlamp-token")
+kubectl get secret $SECRET --namespace headlamp --template=\{\{.data.token\}\} | base64 --decode`
+
+			clusterName := ClusterFlagValue()
+			clusterVersion := ClusterKubernetesVersion(clusterName)
+			parsedClusterVersion, err := util.ParseKubernetesVersion(clusterVersion)
+			if err != nil {
+				tokenGenerationTip = fmt.Sprintf("%s\nIf Kubernetes Version is <1.24:\n%s\n\nIf Kubernetes Version is >1.24:\n%s\n", tokenGenerationTip, createSvcAccountToken, getSvcAccountToken)
+			} else {
+				if parsedClusterVersion.Major >= 1 && parsedClusterVersion.Minor >= 24 {
```
spowelljr, line 90: I know this might be an extreme case, but this would fail on `2.0`

PR #14533, opened 2022-07-08
Add QEMU tests

spowelljr:
- Added tests for QEMU
- Removed unused env `EXPECTED_DEFAULT_DRIVER`
- Removed commented out set pending tests
- Added QEMU to filtering out on PR bot
- Added QEMU tests to flake chart page

### DIFF
--- a/hack/jenkins/linux_integration_tests_kvm.sh
+++ b/hack/jenkins/linux_integration_tests_kvm.sh
@@ -29,7 +29,6 @@ OS="linux"
 ARCH="amd64"
 DRIVER="kvm2"
 JOB_NAME="KVM_Linux"
-EXPECTED_DEFAULT_DRIVER="kvm2"
 
 # We pick kvm as our gvisor testbed because it is fast & reliable
 EXTRA_TEST_ARGS="-gvisor"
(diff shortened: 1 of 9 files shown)


hack/jenkins/linux_integration_tests_kvm.sh
```diff
@@ -29,7 +29,6 @@ OS="linux"
 ARCH="amd64"
 DRIVER="kvm2"
 JOB_NAME="KVM_Linux"
-EXPECTED_DEFAULT_DRIVER="kvm2"
```
medyagh, line 32: why do we delete this ?
on linux we still expect kvm to be default

spowelljr: I removed it because we're not using it anywhere, those are the only usage of the value, we don't use it anywhere

medyagh (commented):
looks good once comment is resolved

hack/jenkins/linux_integration_tests_kvm.sh
spowelljr: Also, these values are incorrect, the default driver would be Docker

medyagh: I see good call !

PR #14535, opened 2022-07-09
Update ubuntu image

spowelljr:
Update Ubuntu from 16.04 to 20.04

Before:
93 vulnerabilities

After:
22 vulnerabilities

### DIFF
--- a/installers/linux/kvm/Dockerfile.amd64
+++ b/installers/linux/kvm/Dockerfile.amd64
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


spowelljr:
`test "`strings out/docker-machine-driver-kvm2-amd64 | grep '^LIBVIRT_[0-9]' | sort | tail -n 1`" = "LIBVIRT_1.2.9"`
Looks like we're expecting a specific version of `LIBVIRT`, this is likely different in 20.04

klaases (approved):
Thanks @spowelljr - wow `ubuntu_16_0_4` is relatively dated!

Glad to see this update and glad that it can be managed through the Dockerfile.

Should definitely help improve security.

PR #14541, opened 2022-07-11
Update dashboard.md

thulsidk1:
added flag for passing the cluster name


thulsidk1:
additional flag for the command

PR #14554, opened 2022-07-11
CI: Add arm64 Homebrew dir to PATH env

spowelljr:
As per Homebrew install instructions, the install path for arm64 is different than amd64:
`/usr/local for macOS Intel, /opt/homebrew for Apple Silicon`

Adding `/opt/homebrew/bin` to `PATH` on macOS arm64 to execute `brew` commands.

Also adding building `e2e-darwin-arm64` to `e2e-cross` as we need that to run tests on M1.


medyagh (approved):

klaases (approved):
LGTM

PR #14582, opened 2022-07-14
Update profile.md

champagne:
Improve the part of minikube profile.


spowelljr (changes requested):
Hi @champagne, these file are auto generated. If you want to make these changes you should update https://github.com/kubernetes/minikube/blob/a433bded0932048a004117b05294e1048a777a4d/cmd/minikube/cmd/config/profile.go and then run `make generate-docs` and then push the result up, thanks

PR #14658, opened 2022-07-28
fixes url index out of range error in service

anoop142:
* if there is no node port, service panics accessing url
* since the length of array is 3 when there is no node port, accessing u[3] results in
	index out of range


spowelljr:
Hi @anoop142, could you run `gofmt -w -s .` and push the changes up, thanks

anoop142:
> Hi @anoop142, could you run `gofmt -w -s .` and push the changes up, thanks

done!

spowelljr (approved):
Thanks for adding this check!

PR #14689, opened 2022-08-01
Enable `CONFIG_PROC_CHILDREN` kernel configuration in Minikube ISO

### LINKED ISSUE #14351: Feature Request: enable `CONFIG_PROC_CHILDREN` kernel configuration option for the VirtualBox VM
### What Happened?

Hi there!

We have an existing tool that runs in containers and relies on the `/proc` pseudo-filesystem to retrieve children processes of any given parent process. It leverages the `/proc/[pid]/task/[tid]/children` file, the presence of which is governed by a `CONFIG_PROC_CHILDREN` kernel config option on the host running those containers [1]

While this works great on most environments, we tested recently with Minikube (using the VirtualBox driver), and noticed that this file was missing because the kernel of the Minikube VirtualBox VM surprisingly does not set this config:

```
$ minikube ssh
$ zgrep CONFIG_PROC_CHILDREN /proc/config.gz                             
# CONFIG_PROC_CHILDREN is not set
```

We worked around this by building a custom minikube ISO, but I was wondering if it wouldn't make more sense to have this set as a default config instead.

If you think this makes sense, I can happily try to contribute a PR :-)

Thanks.

[1] https://man7.org/linux/man-pages/man5/proc.5.html

### Attach the log file

[log.txt](https://github.com/kubernetes/minikube/files/8936506/log.txt)

### Operating System

Redhat/Fedora

### Driver

VirtualBox

rm3l:
As described in #14351, this PR enables the `CONFIG_PROC_CHILDREN` kernel configuration in the Minikube ISO. This way, applications running in containers and relying on the `/proc` pseudo-filesystem would be able to leverage the `/proc/[pid]/task/[tid]/children` file if needed.

I've tested this manually by building a custom ISO with the changes here, but please do let me know if things should be done differently or if I missed any other files.

fixes #14351


afbjorklund:
Seems like an odd kernel config, mostly intended for criu

>              This option is intended for use by the checkpoint-restore
>              (CRIU) system, and reliably provides a list of children
>              only if all of the child processes are stopped or frozen.


The benefit to minikube is not obvious from #14351 ("existing tool" ?)

But it seems to be enabled in Ubuntu, and shouldn't really hurt ?

afbjorklund:
That man page does say "most code should probably not use this interface"

rm3l:
> Seems like an odd kernel config, mostly intended for criu
> 
> > ```
> >          This option is intended for use by the checkpoint-restore
> >          (CRIU) system, and reliably provides a list of children
> >          only if all of the child processes are stopped or frozen.
> > ```
> 
> The benefit to minikube is not obvious from #14351 ("existing tool" ?)
> 
> But it seems to be enabled in Ubuntu, and shouldn't really hurt ?
> 
Yes, I think it shouldn't hurt. Since this is set by default in most distributions/kernels out there, I thought it would also be useful  to have it enabled in the Minikube VM kernel too, so other applications that want to use it could do so.

> That man page does say "most code should probably not use this interface"
>
Yes, I saw that notice. And we currently use it as one possible (fast) implementation for getting information about children of a running process.

afbjorklund (approved):
LGTM, maybe @klaases wants to have a look (and I think the defconfig is out-of-order)

klaases:
> LGTM, maybe @klaases wants to have a look (and I think the defconfig is out-of-order)

Yes, I think this makes sense as far as fixing https://github.com/kubernetes/minikube/issues/14351 goes and this makes sense to me as a viable solution.

From the [man page](https://man7.org/linux/man-pages/man5/proc.5.html):

```
/proc/[pid]/task/[tid]/children (since Linux 3.5)
              A space-separated list of child tasks of this task.  Each
              child task is represented by its TID.

              This option is intended for use by the checkpoint-restore
              (CRIU) system, and reliably provides a list of children
              only if all of the child processes are stopped or frozen.
              It does not work properly if children of the target task
              exit while the file is being read!  Exiting children may
              cause non-exiting children to be omitted from the list.
              This makes this interface even more unreliable than
              classic PID-based approaches if the inspected task and its
              children aren't frozen, and most code should probably not
              use this interface.

              Until Linux 4.2, the presence of this file was governed by
              the CONFIG_CHECKPOINT_RESTORE kernel configuration option.
              Since Linux 4.2, it is governed by the
              CONFIG_PROC_CHILDREN option.
```

@rm3l can you confirm that this will work for your use case: `reliably provides a list of children only if all of the child processes are stopped or frozen`?

I was under the impression that most processes would be ongoing, are they manually stopped/frozen for this to work?

Note: in the spirit of due diligence, I confirmed that `CONFIG_CHECKPOINT_RESTORE` is also not currently configured for minikube.

rm3l:
> > LGTM, maybe @klaases wants to have a look (and I think the defconfig is out-of-order)
> 
> Yes, I think this makes sense as far as fixing #14351 goes and this makes sense to me as a viable solution.
> 
> From the [man page](https://man7.org/linux/man-pages/man5/proc.5.html):
> 
> ```
> /proc/[pid]/task/[tid]/children (since Linux 3.5)
>               A space-separated list of child tasks of this task.  Each
>               child task is represented by its TID.
> 
>               This option is intended for use by the checkpoint-restore
>               (CRIU) system, and reliably provides a list of children
>               only if all of the child processes are stopped or frozen.
>               It does not work properly if children of the target task
>               exit while the file is being read!  Exiting children may
>               cause non-exiting children to be omitted from the list.
>               This makes this interface even more unreliable than
>               classic PID-based approaches if the inspected task and its
>               children aren't frozen, and most code should probably not
>               use this interface.
> 
>               Until Linux 4.2, the presence of this file was governed by
>               the CONFIG_CHECKPOINT_RESTORE kernel configuration option.
>               Since Linux 4.2, it is governed by the
>               CONFIG_PROC_CHILDREN option.
> ```
> 
> @rm3l can you confirm that this will work for your use case: `reliably provides a list of children only if all of the child processes are stopped or frozen`?
> 
> I was under the impression that most processes would be ongoing, are they manually stopped/frozen for this to work?
> 
Thanks for your input, @klaases !
What that means to me is that this `/proc/[pid]/task/[tid]/children` interface is mostly intended to be fast and cheap; as such, it doesn't guarantee to provide precise results, which means, if there a new child created (or a child exits) while we are reading the "children" file, it might or might not be counted. So one need to either stop or freeze processes if precise results are needed.

In our use case, this is fine, as we control both the parent and its "first-level" children processes, but we wanted to be able to get the children PIDs, given the parent PID (and any of its threads). At the point of reading this "children" file, we are sure that all the direct children were still running, and the parent process would not be creating additional children. So this does not require to stop/freeze the child processes, and works if the processes are running. In our use case, this was sufficient even if for some reason, any child exits while we are reading the "children" file.
But I agree that applications relying on the "children" file should be aware of that particular point (already mentioned in the man page).

> Note: in the spirit of due diligence, I confirmed that `CONFIG_CHECKPOINT_RESTORE` is also not currently configured for minikube.
>
Thanks for checking. Indeed, I remember I also checked this and the kernel version in the Minikube VM.

klaases (approved):
Yes, this makes sense from a logical flow perspective, especially in the sense that you're mainly reaching for a snapshot or essentially a status check at a particular point in time, without expressing heavy reliance one way or another - LGTM

rm3l:
Hi @afbjorklund @klaases !
Sorry to ask, but is there anything left to have this PR merged in?
Thanks.

spowelljr:
ok-to-build-iso

PR #14521, opened 2022-07-07
Update addon files to use fully qualified name

siddhant-khisty:
Updated the files in deploy/addons/aliyun_mirror.json 
As the issue mentions files have to be changed in pkg/minikube/assets/addons.go as well, but I don't see anything that needs to be changed. Please let me know if I'm missing something.

fixes #14520

PR #14690, opened 2022-08-01
Fix registry when custom images provided

spowelljr:
Related: https://github.com/kubernetes/minikube/pull/14521

**Before:**
```
minikube addons enable dashboard --images=MetricsScraper=k8s.gcr.io/echoserver:1.4
...
# Resulting (incorrect) image: docker.io/k8s.gcr.io/echoserver:1.4
```

**After:**
```
minikube addons enable dashboard --images=MetricsScraper=k8s.gcr.io/echoserver:1.4
...
# Resulting (correct) image: k8s.gcr.io/echoserver:1.4
```

Don't use default registry if custom image is defined.

### DIFF
--- a/pkg/minikube/assets/addons.go
+++ b/pkg/minikube/assets/addons.go
@@ -785,6 +785,8 @@ func SelectAndPersistImages(addon *Addon, cc *config.ClusterConfig) (images, cus
 		}
 		// Use newly configured custom images.
 		images = overrideDefaults(addonDefaultImages, newImages)
+		// Store custom addon images to be written.
+		cc.CustomAddonImages = mergeMaps(cc.CustomAddonImages, newImages)
 	}
 
 	// Use previously configured custom registries.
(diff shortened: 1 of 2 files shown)


pkg/minikube/assets/addons.go
```diff
@@ -902,6 +902,15 @@ func GenerateTemplateData(addon *Addon, cfg config.KubernetesConfig, netInfo Net
 			opts.Registries[name] = "" // Avoid nil access when rendering
 		}
 
+		// Without the line below, if you try to overwrite an image the default registry is still used in the templating
```
klaases, line 905: Maybe: "Check if custom image is provided, if so, strip out default registry." as a first line for ease of reference.

spowelljr: Added a concise first line

klaases (approved):
Great - thanks for clarifying the change

PR #14720, opened 2022-08-03
GUI: Add QEMU to driver list

spowelljr:


klaases (approved):
Thanks for adding Qemu to the GUI, yes just applies to Linux and Mac - LGTM

PR #14723, opened 2022-08-03
GUI: Change reload to restart & tooltip

spowelljr:
Screen Shot 2022-08-03 at 12 59 27 PM

### DIFF
--- a/gui/advancedview.cpp
+++ b/gui/advancedview.cpp
@@ -126,7 +126,7 @@ void AdvancedView::update(Cluster cluster)
     startButton->setText(getStartLabel(isRunning));
     QString startToolTip = "";
     if (isRunning) {
-        startToolTip = "Restart the Kubernetes API";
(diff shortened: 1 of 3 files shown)


gui/advancedview.cpp
```diff
@@ -124,6 +124,11 @@ void AdvancedView::update(Cluster cluster)
 #endif
     pauseButton->setText(getPauseLabel(isPaused));
     startButton->setText(getStartLabel(isRunning));
+    QString startToolTip = "";
+    if (isRunning) {
+        startToolTip = "Restart the Kubernetes API";
```
medyagh, line 129: Restarts the already started minikube that might refresh the config

medyagh (approved):

gui/advancedview.cpp
spowelljr: Reworded to `Restart an already running minikube instance to pickup config changes.`

PR #14725, opened 2022-08-03
CI: Install cri-docker from pre-built binaries

spowelljr:
Before, we were downloading the cri-dockerd GitHub project and then building the binaries during the CI and copying them to their locations.

However, we've already pre-built the binaries and stored them, so to speed up the CI and make it less complicated this changes it so we're simply downloading the pre-built files to their locations.

Also renamed `VERSION` to `CRICTL_VERSION` as original name was very vague.


klaases (approved):
Instead of pointing to the Github repo, we point directly to storage.googleapi.com.

1. Master - Github actions, run on master, hourly.
2. PR - run on every PR.
3. Bare metal - make sure all deps are installed.

Note: good to see more version names - LGTM

PR #14727, opened 2022-08-03
Fix id for yearly leaderboard job

spowelljr:


klaases (approved):
Thanks for helping to maintain the release process from yesterday - LGTM

PR #14729, opened 2022-08-03
Set output for yearly-leaderboard job

spowelljr:
Was not setting the output, so `if: ${{ steps.yearlyLeaderboard.outputs.changes != '' }}` was always failing


klaases (approved):
Gotcha - glad to have some output in order to avoid the empty output issue - LGTM

PR #14743, opened 2022-08-05
GUI: Fix first dashboard start crashing app

spowelljr:
Sometimes `dashboardProcess` isn't properly init yet and results in the check `if (dashboardProcess)` passing when it shouldn't. Resulting in `dashboardProcess->terminate()` running against `NULL` and crashing the app.

So setting `dashboardProcess = NULL` so it will fail the conditional check on the first pass and prevent the app from crashing.

### DIFF
--- a/gui/operator.cpp
+++ b/gui/operator.cpp
@@ -21,6 +21,7 @@ Operator::Operator(AdvancedView *advancedView, BasicView *basicView, CommandRunn
     m_stackedWidget = stackedWidget;
     m_parent = parent;
     m_isBasicView = true;
+    dashboardProcess = NULL;
 
     connect(m_basicView, &BasicView::start, this, &Operator::startMinikube);
     connect(m_basicView, &BasicView::stop, this, &Operator::stopMinikube);


klaases (commented):
For the `if (dashboardProcess)...` check, is it possible to improve the test itself to check for a `dashboardProcess` itself being unset, rather than explicitly for `NULL`?

I like the idea of declarative value for `dashboardProcess`, even if just `NULL`, however I am curious as to whether we can evaluate on the unset value itself.

spowelljr:
> For the `if (dashboardProcess)...` check, is it possible to improve the test itself to check for a `dashboardProcess` itself being unset, rather than explicitly for `NULL`?
> 
> I like the idea of declarative value for `dashboardProcess`, even if just `NULL`, however I am curious as to whether we can **properly** evaluate on the unset value itself.

The problem is sometimes on startup `dashboardProcess` is pointing to a (lingering?) init object instead of defaulting to `NULL`. When it's set to the init object, it passes the `if (dashboardProcess)` check so we assume `dashboardProcess` is already created, but it actually isn't and then any command run on `dashboardProcess` fails. I found that setting `dashboardProcess` to `NULL` resolves the issue as `dashboardProcess` is then in a state we expect.

klaases (approved):
The init process is working some of the time, but sometimes does not complete in time. This causes us to fail frequently, but based more on timing than anything else.

Not quite a race-condition, just something flaky in the framework. The state by process is not always stable.

We also want to terminate any previous instances, which will only occur after successful init. By setting to NULL, we will start with a clean slate.

By setting the value manually, we won't hit this pending state error - LGTM

PR #14758, opened 2022-08-08
Fix unit tests on macOS arm64

spowelljr:
These tests were failing on an M1 Mac. Changed all VirtualBox drivers to use SSH as M1 Mac doesn't currently support SSH.

### DIFF
--- a/pkg/util/utils_test.go
+++ b/pkg/util/utils_test.go
@@ -19,7 +19,6 @@ package util
 import (
 	"os"
 	"os/user"
-	"runtime"
 	"syscall"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


pkg/util/utils_test.go
```diff
@@ -38,7 +37,7 @@ func TestGetBinaryDownloadURL(t *testing.T) {
 	}
 
 	for _, tt := range testData {
-		url := GetBinaryDownloadURL(tt.version, tt.platform, runtime.GOARCH)
+		url := GetBinaryDownloadURL(tt.version, tt.platform, "amd64")
```
klaases, line 40: Are we sure that this code will always execute on an `amd64` device?

spowelljr: The test has the URLs hardcoded to `amd64` (below), so for this test we should always be generating `amd64` URLs.
```
{"v0.0.1", "linux", "https://storage.googleapis.com/minikube/releases/v0.0.1/minikube-linux-amd64"},
{"v0.0.1", "darwin", "https://storage.googleapis.com/minikube/releases/v0.0.1/minikube-darwin-amd64"},
{"v0.0.1", "windows", "https://storage.googleapis.com/minikube/releases/v0.0.1/minikube-windows-amd64.exe"},
```

klaases (approved):
Great to see, makes sense that we should NOT be checking the local user's arch when running tests.

PR #14759, opened 2022-08-08
Fix arm64 preload upload

spowelljr:
The `-` in `cri-o` was causing issues when generating the Kubernetes version.

**Before:**
```
preloaded-images-k8s-v18-v1.24.3-cri-o-overlay-arm64.tar.lz4
k8s version: v1.24.3-cri

preloaded-images-k8s-v18-v1.25.0-beta.0-cri-o-overlay-arm64.tar.lz4
k8s veresion: v1.25.0
```

**After:**
```
preloaded-images-k8s-v18-v1.24.3-cri-o-overlay-arm64.tar.lz4
k8s version: v1.24.3

preloaded-images-k8s-v18-v1.25.0-beta.0-cri-o-overlay-arm64.tar.lz4
k8s veresion: v1.25.0-beta.0
```

### DIFF
--- a/hack/preload-images/upload.go
+++ b/hack/preload-images/upload.go
@@ -67,7 +67,7 @@ func getVersionsFromFilename(filename string) (string, string) {
 	preloadVersion := parts[3]
 	k8sVersion := parts[4]
 	// this check is for "-rc" and "-beta" versions that would otherwise be stripped off
-	if len(parts) == 9 {
+	if len(parts) >= 9 && parts[5] != "cri" {
 		k8sVersion += fmt.Sprintf("-%s", parts[5])
 	}
 	return preloadVersion, k8sVersion


hack/preload-images/upload.go
```diff
@@ -67,7 +67,7 @@ func getVersionsFromFilename(filename string) (string, string) {
 	preloadVersion := parts[3]
 	k8sVersion := parts[4]
 	// this check is for "-rc" and "-beta" versions that would otherwise be stripped off
-	if len(parts) == 9 {
+	if len(parts) >= 9 && parts[5] != "cri" {
```
klaases, line 70: Is it possible that `parts[5]` could be `cri-o` instead of just `cri`?

spowelljr: No, the string is being split on `-`, so `cri` would be the value of `parts[5]`

klaases (approved):
I see, in this case we don't need `cri-o` to be appended as a part of the k8s version at all - LGTM

PR #14762, opened 2022-08-08
Document qemu start and default commands

klaases:
Add some basics to using qemu on our driver documentation page.

### DIFF
--- a/site/content/en/docs/drivers/qemu.md
+++ b/site/content/en/docs/drivers/qemu.md
@@ -13,6 +13,20 @@ The `qemu` driver users QEMU (system) for VM creation.
 
 <https://www.qemu.org/>
 
+## Usage
+
+To start minikube with the qemu driver:
+
+```shell
+minikube start --driver=qemu
+```
+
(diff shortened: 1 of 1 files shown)


spowelljr (approved):

site/content/en/docs/drivers/qemu.md
```diff
@@ -13,6 +13,20 @@ The `qemu` driver users QEMU (system) for VM creation.
...
 
+## Usage
+
+To start minikube with the qemu driver:
+
+```shell
+minikube start --driver=qemu
+```
+
+To make qemu the default driver:
```
medyagh, line 24: lets not advertise making it default for all clusters (since this is an expermintal)

medyagh (changes requested):

site/content/en/docs/drivers/qemu.md
klaases: Sure, we can remove this section on making qemu default.

spowelljr (approved):

PR #14763, opened 2022-08-08
Remove duplicated k8s versions from preload generation

spowelljr:
`constants.DefaultKubernetesVersion`, `constants.NewestKubernetesVersion`, and `constants.OldestKubernetesVersion` are pushed onto an array with all Kubernetes versions.

When `constants.DefaultKubernetesVersion` & `constants.NewestKubernetesVersion` were both `v1.24.3` the generation for `v1.24.3` occurred three times.

Added a loop that removes duplicates so duplicate preload generation doesn't occur.

### DIFF
--- a/hack/preload-images/preload_images.go
+++ b/hack/preload-images/preload_images.go
@@ -32,6 +32,7 @@ import (
 	"github.com/spf13/viper"
 	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/minikube/download"
+	"k8s.io/minikube/pkg/util"
 )
 
 const (
@@ -134,21 +135,7 @@ func collectK8sVers() ([]string, error) {
 		constants.NewestKubernetesVersion,
 		constants.OldestKubernetesVersion,
 	}, k8sVersions...)
-	return removeDuplicates(versions), nil
-}
-
-func removeDuplicates(versions []string) []string {
-	prevVersions := make(map[string]bool)
-	for i := 0; i < len(versions); i++ {
-		v := versions[i]
-		if ok := prevVersions[v]; !ok {
-			prevVersions[v] = true
-			continue
-		}
-		versions = append(versions[:i], versions[i+1:]...)
-		i--
-	}
-	return versions
+	return util.RemoveDuplicateStrings(versions), nil
 }
 
 func makePreload(cfg preloadCfg) error {
--- a/hack/preload-images/preload_images.go
+++ b/hack/preload-images/preload_images.go
@@ -129,11 +129,26 @@ func collectK8sVers() ([]string, error) {
 		}
 		k8sVersions = recent
 	}
-	return append([]string{
+	versions := append([]string{
 		constants.DefaultKubernetesVersion,
 		constants.NewestKubernetesVersion,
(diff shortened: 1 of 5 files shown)


hack/preload-images/preload_images.go
```diff
@@ -129,11 +129,26 @@ func collectK8sVers() ([]string, error) {
...
+	versions := append([]string{
 		constants.DefaultKubernetesVersion,
 		constants.NewestKubernetesVersion,
 		constants.OldestKubernetesVersion,
-	}, k8sVersions...), nil
+	}, k8sVersions...)
+	return removeDuplicates(versions), nil
+}
+
+func removeDuplicates(versions []string) []string {
```
klaases, line 140: This seems like a reusable function that we might want to keep in a shared library (although I have not yet checked elsewhere in the code for whether we de duplicate string slices).

Also, would it be possible to add a test?

I think the `range` function is little easier to read, here is what I came up with:

```
// dedupeStrings deduplicates a string slice, removes empty strings.
func dedupeStrings(dupes []string) []string {
	var uniques []string
	m := make(map[string]struct{}, len(dupes))
	for _, v := range dupes {
		if _, ok := m[v]; !ok {
			m[v] = struct{}{}
			uniques = append(uniques, v)
		}
	}
	return uniques
}
```

Test

```
func TestDedupeStrings(t *testing.T) {
	testCases := []struct {
		desc  string
		slice []string
		want  []string
	}{
		{
			desc:  "no duplicates",
			slice: []string{"alpha", "bravo", "charlie"},
			want:  []string{"alpha", "bravo", "charlie"},
		},
		{
			desc:  "adjacent duplicates",
			slice: []string{"alpha", "bravo", "bravo", "charlie"},
			want:  []string{"alpha", "bravo", "charlie"},
		},
		{
			desc:  "non-adjacent duplicates",
			slice: []string{"alpha", "bravo", "alpha", "charlie"},
			want:  []string{"alpha", "bravo", "charlie"},
		},
		{
			desc:  "multiple duplicates",
			slice: []string{"alpha", "bravo", "alpha", "alpha", "charlie", "charlie", "alpha", "bravo"},
			want:  []string{"alpha", "bravo", "charlie"},
		},
		{
			desc:  "unsorted duplicates",
			slice: []string{"charlie", "bravo", "alpha", "bravo"},
			want:  []string{"charlie", "bravo", "alpha"},
		},
	}
		t.Run(tc.desc, func(t *testing.T) {
			got := dedupeStrings(tc.slice)
			if diff := cmp.Diff(got, tc.want); diff != "" {
				t.Errorf("dedupeStrings(%v) = %v, want: %v", tc.slice, got, tc.want)
			}
		})
	}
}
```

spowelljr: That makes sense, I was trying to go the path of removing duplicated elements from the original slice, but it's cleaner to go the path of creating a new slice and add unique elements.

I can also add the test, we don't currently run tests in the `hack` dir, but I can update our `make test` command to check for tests in there as well.

klaases (approved):
Great to see this as utility class.

Glad to see at least two instances of RemoveDuplicateStrings() can be invoked.

Glad to see added tests - LGTM

PR #14765, opened 2022-08-08
Site: add documentation for flag: 'qemu-firmware-path'

### LINKED ISSUE #14622: qemu from macports is not recognized
### What Happened?

I installed minikube and qemu from macports. It does not work. It exits with the following error: "Exiting due to PROVIDER_QEMU2_NOT_FOUND: The 'qemu2' provider was not found: lookup qemu: open /opt/homebrew/Cellar/qemu: no such file or directory". Is it possible to configure it to work with macports qemu?

```
% minikube version
minikube version: v1.26.0
% qemu-system-aarch64 --version
QEMU emulator version 7.0.0
% which qemu-system-aarch64 
/opt/local/bin/qemu-system-aarch64
% minikube start --driver=qemu2 --alsologtostderr -v=4
I0722 18:44:57.381344    2008 out.go:296] Setting OutFile to fd 1 ...
I0722 18:44:57.381511    2008 out.go:348] isatty.IsTerminal(1) = true
I0722 18:44:57.381517    2008 out.go:309] Setting ErrFile to fd 2...
I0722 18:44:57.381521    2008 out.go:348] isatty.IsTerminal(2) = true
I0722 18:44:57.381795    2008 root.go:329] Updating PATH: /Users/vbezhenar/.minikube/bin
W0722 18:44:57.381880    2008 root.go:307] Error reading config file at /Users/vbezhenar/.minikube/config/config.json: open /Users/vbezhenar/.minikube/config/config.json: no such file or directory
I0722 18:44:57.381976    2008 out.go:303] Setting JSON to false
I0722 18:44:57.402903    2008 start.go:115] hostinfo: {"hostname":"Vladimirs-MacBook-Pro.local","uptime":22151,"bootTime":1658471746,"procs":488,"os":"darwin","platform":"darwin","platformFamily":"Standalone Workstation","platformVersion":"12.4","kernelVersion":"21.5.0","kernelArch":"arm64","virtualizationSystem":"","virtualizationRole":"","hostId":"6e488582-1dc2-5e6d-a8ea-0e579bef8fc1"}
W0722 18:44:57.403030    2008 start.go:123] gopshost.Virtualization returned error: not implemented yet
I0722 18:44:57.425357    2008 out.go:177] 😄  minikube v1.26.0 on Darwin 12.4 (arm64)
😄  minikube v1.26.0 on Darwin 12.4 (arm64)
W0722 18:44:57.463505    2008 preload.go:295] Failed to list preload files: open /Users/vbezhenar/.minikube/cache/preloaded-tarball: no such file or directory
I0722 18:44:57.463516    2008 notify.go:193] Checking for updates...
I0722 18:44:57.463760    2008 driver.go:360] Setting default libvirt URI to qemu:///system
I0722 18:44:57.501651    2008 out.go:177] ✨  Using the qemu2 (experimental) driver based on user configuration
✨  Using the qemu2 (experimental) driver based on user configuration
I0722 18:44:57.520906    2008 start.go:284] selected driver: qemu2
I0722 18:44:57.520920    2008 start.go:805] validating driver "qemu2" against <nil>
I0722 18:44:57.520947    2008 start.go:816] status for qemu2: {Installed:false Healthy:false Running:false NeedsImprovement:false Error:lookup qemu: open /opt/homebrew/Cellar/qemu: no such file or directory Reason: Fix: Doc:https://minikube.sigs.k8s.io/docs/reference/drivers/qemu2/ Version:}
I0722 18:44:57.541117    2008 out.go:177] 

W0722 18:44:57.561329    2008 out.go:239] 🤷  Exiting due to PROVIDER_QEMU2_NOT_FOUND: The 'qemu2' provider was not found: lookup qemu: open /opt/homebrew/Cellar/qemu: no such file or directory
🤷  Exiting due to PROVIDER_QEMU2_NOT_FOUND: The 'qemu2' provider was not found: lookup qemu: open /opt/homebrew/Cellar/qemu: no such file or directory
W0722 18:44:57.561471    2008 out.go:239] 📘  Documentation: https://minikube.sigs.k8s.io/docs/reference/drivers/qemu2/
📘  Documentation: https://minikube.sigs.k8s.io/docs/reference/drivers/qemu2/
I0722 18:44:57.600400    2008 out.go:177] 
```

### Attach the log file

[log.txt](https://github.com/kubernetes/minikube/files/9167844/log.txt)

### Operating System

macOS (Default)

### Driver

N/A

klaases:
Fixes #14622

In case the firmware path needs to be specified, use the `--qemu-firmware-path` flag.


vbezhenar:
Hey, I think that it would be useful to provide an example for this parameter. `--qemu-firmware-path=/opt/local/share/qemu/edk2-aarch64-code.fd` is what I used with macports. It might not be immediately obvious which file should be supplied with this parameter.

medyagh (approved):

klaases:
> Hey, I think that it would be useful to provide an example for this parameter. `--qemu-firmware-path=/opt/local/share/qemu/edk2-aarch64-code.fd` is what I used with macports. It might not be immediately obvious which file should be supplied with this parameter.

Hi @vbezhenar, sure I think it makes sense to add this parameter here, for those who are also using M1. Applied suggestion.

klaases:
I think it makes more sense to fix this within minikube.

We can make it so that minikube checks for the Brew path, and if that is missing, checks for MacPort's path instead.

site/content/en/docs/drivers/qemu.md
```diff
@@ -13,6 +13,14 @@ The `qemu` driver users QEMU (system) for VM creation.
 
 <https://www.qemu.org/>
 
+## Special features
+
+minikube start supports some qemu specific flags:
+
+* **`--qemu-firmware-path`**: The path to the firmware image to be used.
+  * Note: if this flag does not take effect, try removing the file `~/.minikube` which may have a reference to this setting.
```
medyagh, line 21: this hopefully should not happen as the flag should override config
but still we should never ask users to delete the config manually, this will make any existing clusters stall and abandonded
instead "minikube delete --all --purge"

medyagh (changes requested):

klaases:
> I think it makes more sense to fix this within minikube.
> 
> We can make it so that minikube checks for the Brew path, and if that is missing, checks for MacPort's path instead.

We should update the documentation now to help others, and I will fix the path detection in a separate PR.

site/content/en/docs/drivers/qemu.md
klaases: Very much appreciated, have updated the documentation as suggested.

Glad to learn that simply removing the `~/.minikube` home directory will cause the existing clusters to stall and become abandoned. In that case I think that those processes would need to be stopped manually or could be stopped by a machine restart.

site/content/en/docs/drivers/qemu.md
```diff
@@ -21,6 +21,14 @@ To start minikube with the qemu driver:
 minikube start --driver=qemu
 ```
 
+## Special features
+
+minikube start supports some qemu specific flags:
+
+* **`--qemu-firmware-path`**: The path to the firmware image to be used.
+  * Note: while the flag should override the config, if the flag does not take effect try running `minikube delete --all --purge`.
+  * Macports: if you are installing [minikube](https://ports.macports.org/port/minikube/) and [qemu](https://ports.macports.org/port/qemu/) via Macports on a Mac with M1, use the following flag: `--qemu-firmware-path=/opt/local/share/qemu/edk2-aarch64-code.fd`
```
spowelljr, line 30:
```suggestion
  * MacPorts: if you are installing [minikube](https://ports.macports.org/port/minikube/) and [qemu](https://ports.macports.org/port/qemu/) via MacPorts on a Mac with M1, use the following flag: `--qemu-firmware-path=/opt/local/share/qemu/edk2-aarch64-code.fd`
```

site/content/en/docs/drivers/qemu.md
```diff
@@ -21,6 +21,14 @@ To start minikube with the qemu driver:
 minikube start --driver=qemu
 ```
 
+## Special features
+
+minikube start supports some qemu specific flags:
+
+* **`--qemu-firmware-path`**: The path to the firmware image to be used.
+  * Note: while the flag should override the config, if the flag does not take effect try running `minikube delete --all --purge`.
```
spowelljr, line 29: Could you explain why we need to delete the entire `~/.minikube` directory, this seems overkill. This flag takes effect at a profile level, so worst case only the individual profile should be removed.

`while the flag should override the config, if the flag does not take effect` Is there a reason the flag won't override the config? Is there something we can change in the code to make it consistently override so we don't have to have this note?

spowelljr (changes requested):

site/content/en/docs/drivers/qemu.md
klaases: The website https://www.macports.org/ uses MacPorts, does it have to be Macports?

site/content/en/docs/drivers/qemu.md
klaases: In his [comment](https://github.com/kubernetes/minikube/issues/14622#issuecomment-1192602670), @vbezhenar found that only by removing `~/.minikube` was he able to make progress.

Instead of removing `~/.minikube`, @medyagh suggested `minikube delete --all --purge`.

The note is a suggestion to help others save time when debugging, but I agree that the code should work without the override. I have not been able to reproduce this.

In a later PR, I plan to specifically check for the qemu firmware path, if not available, I will also check if Qemu was installed via MacPorts instead.

site/content/en/docs/drivers/qemu.md
spowelljr: I'm recommending it to be `MacPorts`, it's currently `Macports` in the PR

site/content/en/docs/drivers/qemu.md
spowelljr: I just read the comment and I agree that deleting `~/.minikube` will resolve this, but that seems like more of a nuclear option. In the issue thread he never talks about `minikube delete`, likely because he wasn't aware of it, so he manually deleted all of `~/.minikube` instead. I'm sure a simple `minikube delete` would have sufficed since that would have deleted the entire config for the profile including `qemu-firmware-path` which is all he truly wanted removed but went about it a different way.

klaases: Sure, that makes sense. In this case, is your preference:

1. Update the note to say `minikube delete`.
2. Remove the note altogether, as most users may already know `minikube delete`.

site/content/en/docs/drivers/qemu.md
klaases: Nice catch, `MacPorts` is my preference as well.

site/content/en/docs/drivers/qemu.md
spowelljr: 1. makes more sense to me, and hopefully we can remove the note altogether in the future if we can guarantee that  the flag will always take effect.

klaases: Yes, have updated the PR per 1. and I agree with hopefully removing the note altogether in the future.

spowelljr (approved):

site/content/en/docs/drivers/qemu.md
vbezhenar: > I just read the comment and I agree that deleting `~/.minikube` will resolve this, but that seems like more of a nuclear option. In the issue thread he never talks about `minikube delete`, likely because he wasn't aware of it, so he manually deleted all of `~/.minikube` instead. I'm sure a simple `minikube delete` would have sufficed since that would have deleted the entire config for the profile including `qemu-firmware-path` which is all he truly wanted removed but went about it a different way.

You're right, I just did the first thing that came in my head as I just started using minikube and there wasn't anything important for me.

PR #14768, opened 2022-08-09
The DefaultKubernetesRepo changed for 1.25.0

afbjorklund:
See: https://github.com/kubernetes/kubernetes/commit/50bea1dad89930ad565526910aadc314b9e9f38b

Changing the old releases would mean having to redo the old preload as well.

But it could be a good idea to make sure that 1.25.0 comes out the right way...


afbjorklund:
@medyagh : why is the old minikube storage-provisioner included, in the "kubeadm" images ?

Shouldn't it be a separate add-on, since it does not originate from the `kubeadm images list` ?

spowelljr (approved):

Issue #14724, opened 2022-08-03
cri-dockerd v0.2.4 does not work out of the box with minikube


mgabeler-lee-6rs:
### What Happened?

https://github.com/Mirantis/cri-dockerd/releases/tag/v0.2.4 now defaults to using the `kubenet` network plugin (https://github.com/Mirantis/cri-dockerd/pull/99). This does not work, at least with the none driver.

The cri-dockerd README suggests the `cni` driver as perhaps an alternative, but this doesn't work either.

Only by checking the old code from cri-dockerd v0.2.3 can one find that it used to run with "no" network plugin (`--network-plugin=` on the cri-dockerd command line) as a valid arg, and that is what is needed to make it work.

### Attach the log file

With the `kubenet` driver, kubelet reports:
```
Aug 03 16:04:29 MYHOSTNAME kubelet[3512192]: E0803 16:04:29.502654 3512192 kubelet.go:2349] "Container runtime network not ready" networkReady="NetworkReady=false reason:NetworkPluginNotReady message:docker: network plugin is not ready: could not locate kubenet required CNI plugins [bridge host-local loopback] at [\"/opt/cni/bin\" \"/opt/cni/bin\"]"
```
and the node never becomes ready and so no pod is ever scheduled.

With the `cni` driver:
```
Aug 03 16:05:15 MYHOSTNAME kubelet[3517922]: E0803 16:05:15.295210 3517922 kubelet.go:2349] "Container runtime network not ready" networkReady="NetworkReady=false reason:NetworkPluginNotReady message:docker: network plugin is not ready: cni config uninitialized"
```

### Operating System

Ubuntu

### Driver

None (Baremetal)

afbjorklund:
This seems like a breaking change from upstream

afbjorklund:
It is fixed by this PR, which _always_ sets the network plugin: (avoids the previous shortcut)

* https://github.com/kubernetes/minikube/pull/14703

afbjorklund:
@mgabeler-lee-6rs the errors (with "cni") seems weird, it is like CNI plugins were not properly installed ?

`network plugin is not ready: could not locate kubenet required CNI plugins [bridge host-local loopback] at [\"/opt/cni/bin\" \"/opt/cni/bin\"]`

`network plugin is not ready: cni config uninitialized`

If not using the "no-op" (previous default network plugin), then both `cri-tools` and `cni-plugins` are required.

* https://github.com/kubernetes-sigs/cri-tools

[...]

PR #14703, opened 2022-08-02
Handle CRI config of NetworkPlugin and PauseImage

afbjorklund:
These have been removed from the kubelet, and are supposed to be
handled by the CRI config. Remove the earlier Docker-only hacks.

This fixes the issue where kubeadm said 3.7, and docker ran 3.6...

**k8s.gcr.io/pause:3.7**
**k8s.gcr.io/pause:3.6**
k8s.gcr.io/kube-scheduler:v1.24.1
k8s.gcr.io/kube-proxy:v1.24.1
k8s.gcr.io/kube-controller-manager:v1.24.1
k8s.gcr.io/kube-apiserver:v1.24.1
k8s.gcr.io/etcd:3.5.3-0
k8s.gcr.io/coredns/coredns:v1.8.6
gcr.io/k8s-minikube/storage-provisioner:v5

Also removes the old flag, which avoids a warning from kubeadm

`Flag --pod-infra-container-image has been deprecated, will be removed in 1.27. Image garbage collector will get sandbox image information from CRI.`

PR #14780, opened 2022-08-11
Always use cni unless running with dockershim

### LINKED ISSUE #14760: Default CNI to enabled, even for Docker runtime
With the release of cri-dockerd 0.2.5, it now defaults to CNI... So it should be recommended, for _all_ runtimes and single-node.

Previously it was only enabled for runtimes using CRI, such as containerd and cri-o, or multi-node. But now, also for docker.

----

This also applies to "none" driver, which has some weird special case (breaking containerd!)

From before, the default CNI plugin is "kindnet" for KIC and multi-node and "bridge" otherwise.

To be conservative with ancient versions, it can use the dockershim change point (1.24.0-alpha.2)

It was suggested before: #6094 (to use CNI), but then it was regarded as overkill (similar to CRI)

----

* ~~https://github.com/kubernetes/minikube/issues/2207~~
* ~~https://github.com/kubernetes/minikube/issues/3567~~

Basically, rules and interfaces only applied to "others"...

Now `crictl` and `/opt/cni/bin` are needed **everywhere**.

afbjorklund:
Since cri-dockerd 0.2.5, it now defaults to cni.

So when using cri (not dockershim), use cni too.

----

Closes #14760

### DIFF
--- a/pkg/minikube/cni/cni.go
+++ b/pkg/minikube/cni/cni.go
@@ -87,7 +87,7 @@ func New(cc *config.ClusterConfig) (Manager, error) {
 	var err error
 	switch cc.KubernetesConfig.CNI {
 	case "", "auto":
-		cnm = chooseDefault(*cc)
+		cnm, err = chooseDefault(*cc)
 	case "false":
 		cnm = Disabled{cc: *cc}
 	case "kindnet", "true":
@@ -117,33 +117,40 @@ func IsDisabled(cc config.ClusterConfig) bool {
 		return true
 	}
 
-	if chooseDefault(cc).String() == "Disabled" {
+	def, err := chooseDefault(cc)
+	if err == nil && def.String() == "Disabled" {
 		return true
 	}
 	return false
 }
 
-func chooseDefault(cc config.ClusterConfig) Manager {
+func chooseDefault(cc config.ClusterConfig) (Manager, error) {
 	// For backwards compatibility with older profiles using --enable-default-cni
 	if cc.KubernetesConfig.EnableDefaultCNI {
 		klog.Infof("EnableDefaultCNI is true, recommending bridge")
-		return Bridge{}
+		return Bridge{}, nil
 	}
 
 	if len(cc.Nodes) > 1 || cc.MultiNodeRequested {
 		// Enables KindNet CNI in master in multi node cluster, This solves the network problem
 		// inside pod for multi node clusters. See https://github.com/kubernetes/minikube/issues/9838.
 		klog.Infof("%d nodes found, recommending kindnet", len(cc.Nodes))
-		return KindNet{cc: cc}
+		return KindNet{cc: cc}, nil
 	}
 
-	if cc.KubernetesConfig.ContainerRuntime != constants.Docker {
+	version, err := util.ParseKubernetesVersion(cc.KubernetesConfig.KubernetesVersion)
+	if err != nil {
+		return nil, err
+	}
+
+	if cc.KubernetesConfig.ContainerRuntime != constants.Docker || version.GTE(semver.MustParse("1.24.0-alpha.2")) {
+		// Always use CNI when running with CRI (without dockershim)
 		if driver.IsKIC(cc.Driver) {
 			klog.Infof("%q driver + %q runtime found, recommending kindnet", cc.Driver, cc.KubernetesConfig.ContainerRuntime)
-			return KindNet{cc: cc}
+			return KindNet{cc: cc}, nil
 		}
 		klog.Infof("%q driver + %q runtime found, recommending bridge", cc.Driver, cc.KubernetesConfig.ContainerRuntime)
-		return Bridge{cc: cc}
+		return Bridge{cc: cc}, nil
 	}
 
 	// for docker container runtime and k8s v1.24+ where dockershim and kubenet were removed, we fallback to bridge cni for cri-docker(d)
@@ -155,11 +162,11 @@ func chooseDefault(cc config.ClusterConfig) Manager {
 	kv, err := util.ParseKubernetesVersion(cc.KubernetesConfig.KubernetesVersion)
 	if err == nil && kv.GTE(semver.MustParse("1.24.0-alpha.2")) {
 		klog.Infof("%q driver + %q container runtime found on kubernetes v1.24+, recommending bridge", cc.Driver, cc.KubernetesConfig.ContainerRuntime)
-		return Bridge{cc: cc}
+		return Bridge{cc: cc}, nil
 	}
 
 	klog.Infof("CNI unnecessary in this configuration, recommending no CNI")
-	return Disabled{cc: cc}
+	return Disabled{cc: cc}, nil
 }
 
 // manifestPath returns the path to the CNI manifest
--- a/site/content/en/docs/drivers/includes/none_usage.inc
+++ b/site/content/en/docs/drivers/includes/none_usage.inc
@@ -14,6 +14,7 @@ This VM must also meet the [kubeadm requirements](https://kubernetes.io/docs/set
 * iptables (in legacy mode)
 * conntrack
 * crictl
+* cni-plugins
 * SELinux permissive
 * cgroups v1 (v2 is not yet supported by Kubernetes)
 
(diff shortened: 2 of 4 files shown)


medyagh:
Thank you for the PR
@afbjorklund mind taking a look at this failure to see if it is related to this PR or not ?
https://storage.googleapis.com/minikube-builds/logs/14780/25250/Docker_Linux.html#fail_TestNetworkPlugins%2fgroup%2fauto%2fHairPin

pkg/minikube/cni/cni.go
```diff
@@ -132,42 +132,44 @@ func IsDisabled(cc config.ClusterConfig) bool {
...
 
-func chooseDefault(cc config.ClusterConfig) Manager {
+func chooseDefault(cc config.ClusterConfig) (Manager, error) {
 	// For backwards compatibility with older profiles using --enable-default-cni
 	if cc.KubernetesConfig.EnableDefaultCNI {
 		klog.Infof("EnableDefaultCNI is true, recommending bridge")
-		return Bridge{}
-	}
-
-	if driver.BareMetal(cc.Driver) {
```
medyagh, line 148: is this change related to this PR ?

afbjorklund:
> mind taking a look at this failure to see if it is related to this PR or not ?

it is related, it was hardcoding "not Docker" in the expectations.

`                        {"auto", []string{}, "", "", ContainerRuntime() != "docker"},`

pkg/minikube/cni/cni.go
afbjorklund: yeah, it removed some weird special case for "none" driver - which needs CNI, just like all other drivers

spowelljr:
@afbjorklund prow is failing to start

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/minikube/14780/pull-minikube-build/1565039962942869504

mgabeler-lee-6rs:
With the latest release of `crictl`, minikube with the `none` driver no longer works without a newer `cri-dockerd`, which doesn't work without this PR, and I believe #14703.

Any outlook on when those two (and perhaps #14827 which is related) might be ready to merge?

afbjorklund:
There is a new cri-dockerd 0.3.0 out, with support for Kubernetes 1.26.0. Hopefully it can include the other fixes... ?

mgabeler-lee-6rs:
Hmm, at one point I needed some combination of this & the other two PRs I listed to address #14724 -- that specifically lists #14703 more than this one for that issue, but I'd been testing with a local build merging all 3. Perhaps some further changes in cri-dockerd have made that not critical now, as it _seems_ to work out of the box again if I grab the latest releases of minikube & cri-dockerd & crictl.

maxb:
How will the changes here impact users like me, who just want to run with the `none` driver and no network plugin at all, for a simple single-node cluster on a development laptop? I currently do this by adding the `--network-plugin=` option to the `cri-dockerd` command line.

afbjorklund:
> How will the changes here impact users like me, who just want to run with the `none` driver and no network plugin at all, for a simple single-node cluster on a development laptop? I currently do this by adding the `--network-plugin=` option to the `cri-dockerd` command line.

The no-op plugin (--network-plugin=) and kubenet has been removed from upstream (retro-actively), so since Kubernetes 1.24 it is now _mandatory_ to use CRI and CNI... But you can still use a simple single-node network if you want to, by using `--cni=bridge` (it is rather similar).

maxb:
Currently I have a functional single node cluster with no CNI at all, using Kubernetes 1.26.0, minikube 1.28.0, cri-dockerd 0.3.0, passing `--network-plugin=` to cri-dockerd.

afbjorklund:
If you really want it, it would be possible to make a special case for docker.

afbjorklund:
Previously the code used `""` to mean Docker container runtime, or `""` to mean No-op network plugin.

But now they mean "use the default" (whatever that happens to be), so need explicit strings for the old values.

maxb:
My main interest is that the learning curve towards operating a single node driver=none setup, should avoid extra complexity where practical. It has already grown a fair bit recently, (cri-dockerd, crictl, needing to research the requirement to add `--network-plugin=`) but I think there's still value in allowing people to leave out CNI where not necessary, and defer tangling with that until a later stage of their Kubernetes educational journey.

I recognise driver=none is never going to be the preferred option for "just make it work, as quickly as possible" use-cases, but it really is great for people willing to put in a minor additional setup cost, in exchange for easier educational tinkering.

afbjorklund:
> My main interest is that the learning curve towards operating a single node driver=none setup, should avoid extra complexity where practical. It has already grown a fair bit recently, (cri-dockerd, crictl, needing to research the requirement to add `--network-plugin=`) but I think there's still value in allowing people to leave out CNI where not necessary, and defer tangling with that until a later stage of their Kubernetes educational journey.

The simple days of Docker are _gone_, upstream has decided to make CRI mandatory (in 1.24) and to make CNI the default (in 1.25, retroactively also for 1.24). And on the latest release day, Docker was not even available (for 1.26).

So unfortunately people now **have** to learn both CRI and CNI, before the cluster will even start. It is possible that this will also include CSI in the future, but for now storage remains "optional" for a single-node cluster (and NFS still works...)

I fully assume that we will have to change to containerd as the default for 1.27, simple because dockerd isn't tested.
That will make life even worse, since you have to learn about containerd and buildkitd and other low-level stuff.

The `nerdctl` tools makes them slightly more comfortable to use, but it is not fully compatible with the old tools ?
You can also go the `podman` route (with CRI-O), if you prefer to use something "different". But still not Docker.

> I recognise driver=none is never going to be the preferred option for "just make it work, as quickly as possible" use-cases, but it really is great for people willing to put in a minor additional setup cost, in exchange for easier educational tinkering.

The "none" driver is used in both learning and testing environments, and it remains an important target.

I detailed some of the **worst** current bugs with it here:

* https://github.com/kubernetes/minikube/issues/15097

The lack of packaging and the lack of documentation makes it more cumbersome, than what it needed to be.

With Docker no longer being Open Source (with Desktop, and Machine no longer being developed), it is "gone".
I personally think that Lima is the best successor, after losing faith in Podman Desktop. Your mileage may vary.

So it seems that Kubernetes is now more Linux, in that you can tinker with it yourself or you can use a distribution ?
Hopefully, `minikube start` will continue to be such a Kubernetes distribution - with all the batteries included.

Anyway, _this_ feature (i.e. network-plugin) will remain - while available.

For the future of minikube, there is always the Slack channel: #minikube

medyagh (approved):

medyagh:
@afbjorklund the only test failing is 


TestNetworkPlugins/group/false
--


<br class="Apple-interchange-newline">[TestNetworkPlugins/group/false](https://storage.googleapis.com/minikube-builds/logs/14780/28076/KVM_Linux.html#fail_TestNetworkPlugins%2fgroup%2ffalse)

could u take a look if we fix that we could merge

spowelljr (approved):

tuanaac:
@afbjorklund could you please advise me how to change the subnet? 

> > How will the changes here impact users like me, who just want to run with the `none` driver and no network plugin at all, for a simple single-node cluster on a development laptop? I currently do this by adding the `--network-plugin=` option to the `cri-dockerd` command line.
> 
> The no-op plugin (--network-plugin=) and kubenet has been removed from upstream (retro-actively), so since Kubernetes 1.24 it is now _mandatory_ to use CRI and CNI... But you can still use a simple single-node network if you want to, by using `--cni=bridge` (it is rather similar).
> 
> ```yaml
@afbjorklund could you please advise me how to update the 
> {
>   "cniVersion": "0.3.1",
>   "name": "bridge",
>   "plugins": [
>     {
>       "type": "bridge",
>       "bridge": "bridge",
>       "addIf": "true",
>       "isDefaultGateway": true,
>       "forceAddress": false,
>       "ipMasq": true,
>       "hairpinMode": true,
>       "ipam": {
>           "type": "host-local",
>           "subnet": "10.244.0.0/16"
>       }
>     },
>     {
>       "type": "portmap",
>       "capabilities": {
>           "portMappings": true
>       }
>     }
>   ]
> }
> ```

Issue #14769, opened 2022-08-09
Stop using "k8s.gcr.io" registry, use "registry.k8s.io" for new


afbjorklund:
### Background

https://github.com/kubernetes/k8s.io/wiki/New-Registry-url-for-Kubernetes-(registry.k8s.io)

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.25.md#changelog-since-v1240

> There is a new OCI image registry (registry.k8s.io) that can be used to pull kubernetes images. The old registry (k8s.gcr.io) will continue to be supported for the foreseeable future, but the new name should perform better because it frontends equivalent mirrors in other clouds. Please point your clusters to the new registry going forward.

----

All the images from `kubeadm`, needs to be moved: (they have _already_ changed, in `v1.25.0-alpha.1`)

* https://github.com/kubernetes/minikube/pull/14768

Probably would be a good idea to put the 5 year old (!) `k8s.gcr.io/echoserver:1.4` out of its misery, too ?

* https://github.com/kubernetes/minikube/issues/11107

----

Unfortunately, "k8s.gcr.io" seems to be hardcoded in a _lot_ of add-ons. Not sure if they are affected ?

```
pkg/addons/addons.go:				"IngressController": "k8s.gcr.io",
pkg/minikube/assets/addons.go:		"Elasticsearch":        "k8s.gcr.io",
pkg/minikube/assets/addons.go:		"FluentdElasticsearch": "k8s.gcr.io",
pkg/minikube/assets/addons.go:		"KubeWebhookCertgenCreate": "k8s.gcr.io/ingress-nginx/kube-webhook-certgen:v1.1.1@sha256:64d8c73dca984af206adf9d6d7e46aa550362b1d7a01f3a0a91b20cc67868660",
pkg/minikube/assets/addons.go:		"KubeWebhookCertgenPatch": "k8s.gcr.io/ingress-nginx/kube-webhook-certgen:v1.1.1@sha256:64d8c73dca984af206adf9d6d7e46aa550362b1d7a01f3a0a91b20cc67868660",
pkg/minikube/assets/addons.go:		"IngressController": "k8s.gcr.io",
pkg/minikube/assets/addons.go:		"MetricsServer": "k8s.gcr.io",
pkg/minikube/assets/addons.go:		"NvidiaDriverInstaller": "k8s.gcr.io",
pkg/minikube/assets/addons.go:		"Pause":                 "k8s.gcr.io",
[...]

PR #14791, opened 2022-08-15
Move from k8s.gcr.io to registry.k8s.io

RA489:
Move from k8s.gcr.io to registry.k8s.io
As kubeadm images are moved from k8s.gcr.io to registry.k8s.io
part of xref: https://github.com/kubernetes/minikube/issues/14769


afbjorklund (changes requested):
As I wrote in PR #14768, I don't think it's a good idea to change the historic releases - since it won't match the output of `kubeadm` and it won't match the contents of the preloads ?

spowelljr:
This change was merged in https://github.com/kubernetes/minikube/pull/14768

PR #14829, opened 2022-08-21
Upgrade cni-plugins from 0.8.5 to 0.8.7

afbjorklund:
This is the same version as the kubernetes packages are using,
even though there are newer versions available (0.9.1, 1.1.1).

Add back the source code, in case we go back to building the
cni-plugins properly again in the future (currently: binary).


afbjorklund:
ok-to-build-iso

klaases (approved):

PR #14831, opened 2022-08-21
Upgrade cri-tools from 1.21.0 to 1.24.2

afbjorklund:
As usual, only install "crictl" but not "critest"

Version matches what is being used by k8s packages


afbjorklund:
ok-to-build-iso

Issue #12946, opened 2021-11-14
Running minikube under multipass


afbjorklund:
Our friends at Canonical have made a "workflow" for running minikube with [multipass](https://multipass.run/)...

It will create an Ubuntu VM for you, install Docker Engine and then run `minikube` on it:

```
$ multipass launch minikube
Launched: minikube

$ multipass exec minikube -- minikube status
minikube
type: Control Plane
host: Running
kubelet: Running
apiserver: Running
kubeconfig: Configured
```

See https://github.com/canonical/multipass-workflows/blob/main/v1/minikube.yaml

(it is using the deb installation from https://docs.docker.com/engine/install/ubuntu/)

```shell
adduser ubuntu docker
...
sudo -u ubuntu minikube start --driver=docker
```

afbjorklund:
This script would probably have been easier to do, if minikube hadn't been removed from the Ubuntu app store:

* Issue https://github.com/kubernetes/minikube/issues/7977

Before:
```yaml
        snap:
          commands:
          - snap install minikube
```

After:
```yaml
        packages:
        - dpkg-dev

        runcmd:
        - wget https://storage.googleapis.com/minikube/releases/latest/minikube_latest_$( dpkg-architecture -q DEB_HOST_ARCH ).deb
        - apt-get install --yes --no-install-recommends -o Debug::pkgProblemResolver=yes ./minikube_latest_$( dpkg-architecture -q DEB_HOST_ARCH ).deb
```

afbjorklund:
Currently multipass uses the docker-in-docker cluster, but it would also be possible to use docker directly.

`minikube start --driver=docker`

`minikube start --driver=none`

It also installs `kubectl` from snap on the VM, instead of using the client binary from `minikube kubectl`.

petiepooo:
I think there may be a simpler solution.  Multipass (v1.9.0) includes hyperkit under /Library/Application Support/com.canonical.multipass/bin.  I tried creating a symlink for it in /usr/local/bin, and starting minikube, but it exits with the following failure:

> $ minikube start
> 😄  minikube v1.25.2 on Darwin 10.14.6
> ✨  Automatically selected the hyperkit driver
> 
> 💣  Exiting due to PROVIDER_HYPERKIT_ERROR: hyperkit version check failed:
parse date: parsing time "40cbd5" as "20060102": cannot parse "d5" as "2006"

[...]

PR #14859, opened 2022-08-25
remove hyperkit version checks

petiepooo:
Minikube was using an undocumented API. The hyperkit developers
never promised a stable format for their version string.  In fact,
the version has changed from date-based to commit-based version
strings.  Eg:
> hyperkit: 40cbd5

Since it has been several releases since the last issue caused by
older versions of hyperkit, and removal of obsolete code is often
something to be celebrated, this removes any checks for a minimum
compatible version of hyperkit.  If hyperkit is installed, it is
assumed to be a compatible version.

Related to #12946 

Signed-off-by: Pete Nelson <petiepooo@gmail.com>


medyagh (changes requested):
Thank you for this PR and letting us know of the new change int he hyperkit versioning
Instead of removing this code, I suggest have the code be smarter and if the version doesn’t not match the date format we could skip the version check

However for users who are still using an ancient hyperkit it is useful that we let them know
 The min version required.
I know many users who hardly `remember to update their hyperkiet

petiepooo:
You're welcome.  Glad to help out.  I've always found a PR gets a quicker response than an issue, and I'm glad to see that is true here as well.

Would you suggest leaving the version check in place for the life of the project?  If not, at what point in time do you feel it could be safely removed?

I ask because the hard-coded minimum version of hyperkit was released in 2019-02, which was a full 3.5 years ago.  In my opinion, that is a long enough time that nobody would be running something that old.  It was a new and fast-moving project at the time, and they were fixing tons of bugs and security issues, including many more after that time, IIRC.

If the consensus is to leave the version check in place and deal with each new version string format the hyperkit devs choose to use, that's fine.  I am not a go developer, though, so would prefer to withdraw/close this PR and leave the implementation to you all.

Rarely does one get to decrease executable size, increase its speed, and make it more maintainable.  That only happens when code has outlived its usefulness and is deleted.  I'll leave you with this quote:

Every line of code is written without reason, maintained out of weakness, and deleted by chance.
-- [_attributed to Jean-Paul Sartre_](https://programmingisterrible.com/post/139222674273/how-to-write-disposable-code-in-large-systems)

petiepooo:
A short history of hyperkit versions:
20150610 - hyperkit was initially forked from bhyve repository
20150702 - hyperkit added -v flag to display version
20170214 - first tag-based hyperkit version on github
20180403 - hyperkit version first released by brew for macOS
20190201 - minimum hyperkit version required by minikube
20200224 - hyperkit version used when Canonical Multipass was first released
20200909 - current stable hyperkit version used by brew for macOS
20210107 - last tag-based hyperkit version released (also released as version 40cbd5)
20210216 - Multipass releases v1.7.0 which updates hyperkit to version 40cbd5 (current version: 1.10.1)
20220825 - the date I first noticed that hyperkit has a versioning string minikube cannot parse

Less code means less code to maintain and update.  I still advise you to delete the version check completely.  People installing minikube aren't doing so on a system where hyperkit is a version from more than 3.5 years ago, and if they were to get a warning that prompted them to upgrade to a version based on hash instead of date, it would still not work.

PR #14864, opened 2022-08-26
Fix wrong use of loop variable in parallel test

renatolabs:
This fixes an occurrence of a loop variable being incorrectly captured
in a parallel test (`TestVersionIsBetween`). With the previous code,
only the last test case is actually exercised. To work around this
problem, we create a local coapy of the range variable before the
parallel test, as suggested in the documentation for the `testing`
package:

https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks

Some assertions had to be changed as they were incorrect.

Issue was found using the `loopvarcapture` linter.

### DIFF
--- a/pkg/minikube/bootstrapper/bsutil/versions_test.go
+++ b/pkg/minikube/bootstrapper/bsutil/versions_test.go
@@ -49,7 +49,7 @@ func TestVersionIsBetween(t *testing.T) {
 			ver:         semver.MustParse("2.8.0"),
 			gte:         semver.MustParse("1.7.0"),
 			lte:         semver.MustParse("1.9.0"),
-			expected:    true,
+			expected:    false,
 		},
 		{
 			description: "equal to max version",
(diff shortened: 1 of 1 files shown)


spowelljr:
Hi @renatolabs, can you please sign the CLA commented above (https://github.com/kubernetes/minikube/pull/14864#issuecomment-1228658675) so we can continue with the PR, thanks!

renatolabs:
> Hi @renatolabs, can you please sign the CLA commented above ([#14864 (comment)](https://github.com/kubernetes/minikube/pull/14864#issuecomment-1228658675)) so we can continue with the PR, thanks!

Done!

pkg/minikube/bootstrapper/bsutil/versions_test.go
```diff
@@ -70,7 +70,7 @@ func TestVersionIsBetween(t *testing.T) {
 			ver:         semver.MustParse("1.8.0-alpha.0"),
 			gte:         semver.MustParse("1.8.0"),
 			lte:         semver.MustParse("1.9.0"),
-			expected:    true,
+			expected:    false,
```
medyagh, line 73: why this was changed to false?

medyagh (changes requested):

pkg/minikube/bootstrapper/bsutil/versions_test.go
renatolabs: Should the alpha version be between those two versions? I'm not too sure what the actual expectation is here, but the reason I changed the expected values was because the assertion fails otherwise. In both cases I believe changing the expectation does make sense, but let me know if I'm missing something.

spowelljr: Hi @renatolabs, you are correct in that the expected value should be `false`. Would you mind updating the test description to `alpha before` as that's actually what the test case is.

Also, for the following test `beta greater than alpha`, it's impossible for any version to pass as the `gte` and `lte` are backwards. Would you mind swapping those and changing the expected to `true`.

One that change is made I'd be happy to merge this PR, thanks

renatolabs: Updated!

spowelljr (approved):
Tests are passing, thanks for fixing this!

Issue #12036, opened 2021-07-22
Cannot connect to service from localhost on M1 Mac


bradyben:
I'm trying to do the [getting started example](https://minikube.sigs.k8s.io/docs/start/) on an M1 Macbook Air but I'm unable to access the service from the browser. Minikube dashboard works however so that gives me hope that it is possible.
Note that the echoserver:1.4 image used in the docs crashed work and so I substituted it for an arm64 compatible fork (claudiubelu/echoserver:2.3).

**Steps to reproduce the issue:** 

```
~ % minikube start
😄  minikube v1.22.0 on Darwin 11.3.1 (arm64)
✨  Automatically selected the docker driver
👍  Starting control plane node minikube in cluster minikube
🚜  Pulling base image ...
💾  Downloading Kubernetes v1.21.2 preload ...
    > preloaded-images-k8s-v11-v1...: 522.45 MiB / 522.45 MiB  100.00% 7.27 MiB
    > gcr.io/k8s-minikube/kicbase...: 326.19 MiB / 326.19 MiB  100.00% 3.68 MiB
🔥  Creating docker container (CPUs=2, Memory=1988MB) ...
🐳  Preparing Kubernetes v1.21.2 on Docker 20.10.7 ...
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: storage-provisioner, default-storageclass
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default
...
~ % kubectl create deployment hello-minikube --image=claudiubelu/echoserver:2.3
~ % kubectl expose deployment hello-minikube --type=NodePort --port=8080
~ % kubectl get services hello-minikube
NAME             TYPE       CLUSTER-IP      EXTERNAL-IP   PORT(S)          AGE
hello-minikube   NodePort   10.109.188.73   <none>        8080:32048/TCP   9s
~ % service/hello-minikube exposed
~ % minikube service hello-minikube
|-----------|----------------|-------------|---------------------------|
| NAMESPACE |      NAME      | TARGET PORT |            URL            |
|-----------|----------------|-------------|---------------------------|
| default   | hello-minikube |        8080 | http://192.168.49.2:32048 |
|-----------|----------------|-------------|---------------------------|
[...]

PR #14581, opened 2022-07-14
Update - add tips about echoserver image for apple m1 users

NicoleRobin:
the echoserver image in get started is not work for apple m1 user, i waste a little time here, so add a tip here to notify people who use apple m1.

detail discussion about this problem is here https://github.com/kubernetes/minikube/issues/12036#issuecomment-890100715

PR #14867, opened 2022-08-28
Added the arm of echoserver in the deployment section fixes#12036

alinoaimi:
the command: 'kubectl create deployment hello-minikube --image=k8s.gcr.io/echoserver:1.4' doesn't work on arm64 (tested on macOS, m2 processor), 
more details: https://github.com/kubernetes/minikube/issues/12036#issuecomment-890100715
added the arm64 image, to save others time


alinoaimi:
I found out that there was another similar pull request which is still open.

previous pull req: https://github.com/kubernetes/minikube/pull/14581

I'm closing the pull request as it's duplicate.

PR #14868, opened 2022-08-28
Fix kubectl tab-completion and improve error messages

### LINKED ISSUE #12938: Tab completion breaks if using recommended alias kubectl='minikube kubectl --'
1. install minkube
2. add alias kubectl='minikube kubectl --' to .bashrc
3. add source <(kubectl completion bash) to .bashrc

result:
only the commands get tab-completed, nothing after that.

the underlying issue is this:
```
$ minikube kubectl __completeNoDesc get p
Error: flags cannot be placed before plugin name: --cluster
```

this works however:
```
~/.minikube/cache/linux/v1.22.3/kubectl __completeNoDesc get p
```

related to this: https://github.com/kubernetes/kubectl/issues/884

oldium:
Try to fix the following issues when `kubectl` is an alias for `minikube kubectl --` and tab-completion for `kubectl` is [enabled](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/#enable-shell-autocompletion):

 1. When `--cluster name` form of command-line option is used, it conflicts with argument parsing, `name` is interpreted as plugin name. Call to `kubectl --cluster test help` leads to error `Error: flags cannot be placed before plugin name: --cluster`. Fix this by using `--cluster=name` form, which is correctly handled also in call `kubectl --cluster=test help`.
 2. When `--cluster=name` is added as first argument (as seen in original code), it still leads to error mentioned above when the command is unknown like in the call `kubectl --cluster=test unknown`. Fix this by inserting `--cluster=name` after all (sub-)commands and before any flags/options.
 3. The original code passed `--cluster=name` to normal `kubectl` calls, but omitted it for tab-completion calls. This might lead to wrong tab suggestions about pods, deployments and the like. Fix this by always adding `--cluster=name` option to the command line.
 4. In case of tab-completion, put the `--cluster=name` immediately after `__complete` command to prevent any interference with incomplete commands (just in case). This works fine and does not generate any errors like mentioned in point 2.

Fixes #12938

### DIFF
--- a/cmd/minikube/cmd/kubectl.go
+++ b/cmd/minikube/cmd/kubectl.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"strings"
 	"syscall"
 
 	"github.com/spf13/cobra"
@@ -96,9 +97,24 @@ host. Please be aware that when using --ssh all paths will apply to the remote m
 			os.Exit(1)
 		}
 
-		if len(args) > 1 && args[0] != "--help" && args[0] != cobra.ShellCompRequestCmd {
-			cluster := []string{"--cluster", cname}
-			args = append(cluster, args...)
+		if len(args) > 0 {
+			insertIndex := 0
+			if args[0] == cobra.ShellCompRequestCmd || args[0] == cobra.ShellCompNoDescRequestCmd {
+				// Insert right after __complete to allow code completion from the correct cluster.
+				insertIndex = 1
+			} else {
+				// Add cluster argument before first flag, but after all commands.
+				// This improves error message of kubectl in case the command is wrong.
+				insertIndex = len(args)
+				for i, arg := range args {
+					if strings.HasPrefix(arg, "-") {
+						insertIndex = i
+						break
+					}
+				}
+			}
+			clusterArg := "--cluster=" + cname
+			args = append(append(append([]string{}, args[:insertIndex]...), clusterArg), args[insertIndex:]...)
 		}
 
 		c, err := KubectlCommand(version, binaryMirror, args...)


medyagh (commented):
@[oldium](https://github.com/oldium) thank you for this PR, have you checked if this PR will break other use cases? 
we do not have enough unit-test or integration test arround this

oldium:
I am not sure, really.

I figured out now that the error could be as well fixed when the `--cluster cname` parameter is replaced by `--cluster=cname`. In this form also `--help` works without issues. This should be a more safe fix then.

oldium:
The actual fix could be:

```go
                if len(args) > 1 && args[0] != "--help" {
                    cluster := []string{"--cluster=" + cname}
                    args = append(cluster, args...)
                }
```

or

```go
                if len(args) > 0 {
                    cluster := []string{"--cluster=" + cname}
                    args = append(cluster, args...)
                }
```

or even

```go
                cluster := []string{"--cluster=" + cname}
                args = append(cluster, args...)
```

Tested all versions (`minikube kubectl --help`, `minikube kubectl -- --help`, `minikube kubectl __complete app`, `minikube kubectl __complete logs ""`) and the output looks reasonable.

For the code completion the `--cluster` argument is actually necessary to give correct hints.

oldium:
Please advise which version should I put into the patch 😊

oldium:
Updated patch to be the simplest version (no `if` for `--help` check), tested by installing built `minikube-installer.exe`, tested the following in PowerShell:

```powershell
minikube.exe kubectl --help
minikube.exe kubectl -- --help
minikube.exe kubectl __complete app
minikube.exe kubectl __complete logs `"`"
minikube.exe kubectl get pods
```

And the output looks good, simple `--help` is help from `minikube`, call with `-- --help` is help from `kubectl`, completion works, the pods from `logs` correspond to the `get pods` call.

oldium:
Updated patch to insert the cluster argument before other flags, but after commands to improve error message from kubectl. In case of code completion (__complete arg), insert cluster argument at the beginning.

oldium:
Tested with

```powershell
minikube.exe kubectl
minikube.exe kubectl --help
minikube.exe kubectl --
minikube.exe kubectl -- --help
minikube.exe kubectl __complete app
minikube.exe kubectl __complete logs `"`"
minikube.exe kubectl get pods
```

oldium:
Small update (works exactly the same, just for consistency), rebased to latest master.

oldium:
It looks like the errors in the pipeline are unrelated to my patch, I checked on other Pull Requests and it fails with the same messages.

oldium:
CC @ben-krieger 

You made recent change to allow completion for kubectl command, but removed the cluster argument from completion, which affects the completion results. Please check my Pull Request, which keeps the cluster argument while maintains ability to do correct code completion.

oldium:
Please keep in mind that my PR is older than your work, so it does not touch your work in `Add shell completion for minikube kubectl subcommand` commit. Adding the cluster argument to `ValidArgsFunction` function is necessary too to keep consistent code completion results I think.

ben-krieger:
@oldium Thanks for testing my changes!

By cluster argument, I assume you mean the cluster flag. So this fixes tab complete after `kubectl` if you were to try to type something like `minikube kubectl --cluster cluster-1 -- get pods -A`?

This is definitely made more complex by the common CLI pattern of only requiring `--` when necessary.

oldium:
> @oldium Thanks for testing my changes!
> 
> By cluster argument, I assume you mean the cluster flag. So this fixes tab complete after `kubectl` if you were to try to type something like `minikube kubectl --cluster cluster-1 -- get pods -A`?

I meant the `--cluster` flag passed to `kubectl` command.

I am using PowerShell on Windows, so with some PowerShell magic I have tab-completion fully working. Here we are talking about command `minikube kubectl --profile=cluster-1 -- get pods -A`, which is then translated to `kubectl --cluster=cluster-1 get pods -A` command. This works as intended.

With my PR, tab-completion for `kubectl --cluster=cluster-1 get pods <TAB>` (with `kubectl` translated to `minikube kubectl --`)  offers pod names from the correct cluster/minikube profile, while `minikube kubectl --profile=cluster-1 -- get pods <TAB>` ignores the profile flag completely.

> This is definitely made more complex by the common CLI pattern of only requiring `--` when necessary.

I have re-tested kubectl and it looks like it could work even with the simplest change I originally [proposed](https://github.com/kubernetes/minikube/pull/14868#issuecomment-1230434700). The `--cluster=name` change is a must (instead of `--cluster name`). Just try output of `kubectl --cluster cluster-1 help` (error) and  `kubectl --cluster=cluster-1 help` (works). I do not know if there was some change to `kubectl` in this regard, but the `--cluster` flag can be placed before `help` and even `__complete` command-line arguments now. I remember having the error `Error: flags cannot be placed before...` during my experiments with placing the flag on different places. I need to re-check this.

> I'll take a look at this code later. Is it alright if I suggest a fix for the `ValidArgsFunction`?

Yep, feel free to suggest a fix 😁

ben-krieger:
@oldium 

> The --cluster=name change is a must (instead of --cluster name). Just try output of kubectl --cluster cluster-1 help (error) and kubectl --cluster=cluster-1 help (works). I do not know if there was some change to kubectl in this regard, but the --cluster flag can be placed before help and even __complete command-line arguments now. I remember having the error Error: flags cannot be placed before... during my experiments with placing the flag on different places. I need to re-check this.

What?! That's crazy. That seems like a bug in either kubectl or Cobra. I'm also going to test this out from scratch, which may take a little time. Hopefully we can get this working better before the next release.

oldium:
As soon as I sent it I was able to reproduce my issue... Just compare the output of

```
#> kubectl nope --cluster=cluster-1
error: unknown command "nope" for "kubectl"

Did you mean this?
        top
```

and

```
#> kubectl --cluster=cluster-1 nope
Error: flags cannot be placed before plugin name: --cluster=cluster-1
```

This is why my change looks complex - I am adding the `--cluster=name` behind all non-flag command-line arguments just to allow kubectl to provide better error text.

oldium:
Updated description and commit-message to contain the full reasoning behind the solution. Rebased to latest master.

If there is anything more to be solved in `kubectl` app or `Cobra` library (like mentioned by Ben), it can be done independently on this solution. This solution works without changes to other software components.

spowelljr:
@oldium Sorry for the delay, I just rebased the PR, is this good to go now?

oldium:
> @oldium Sorry for the delay, I just rebased the PR, is this good to go now?

Looks good, ready to go 👍

PR #14890, opened 2022-08-30
enable `socket_vmnet` network driver for qemu Mac M1 ARM64

klaases:
Start minikube with `socket_vmnet` network driver for Mac OS M1 / AARCH64.

- Respect network flag (previously hard-coded to "user").
- Create custom `socket_vmnet_client` and `socket_vmnet` paths.

Overview:
- With default `socket_vmnet_client` and `socket_vmnet` paths:
   - `./out/minikube start --driver=qemu --network=socket`
- With optional custom `socket_vmnet_client` and `socket_vmnet` custom paths:
   - `./out/minikube start --driver=qemu --network=socket --socket-vmnet-client-path=/opt/socket_vmnet/bin/socket_vmnet_client --socket-vmnet-path=/var/run/socket_vmnet`

We are still in the process of implementing finalized socket_vmnet network flags, and will have a follow-up PR shortly.

```
😄  minikube v1.26.1 on Darwin 12.5.1 (arm64)
✨  Using the qemu2 (experimental) driver based on user configuration
❗  Using qemu with --network=socket for 'socket_vmnet' is experimental

❌  Exiting due to MK_UNIMPLEMENTED: socket_vmnet network flags are not yet implemented with the qemu2 driver.
    See https://github.com/kubernetes/minikube/pull/14890 for details.
```


pkg/drivers/qemu/qemu.go
```diff
@@ -445,7 +445,9 @@ func (d *Driver) Start() error {
 			d.diskPath())
 	}
 
-	if stdout, stderr, err := cmdOutErr(d.Program, startCmd...); err != nil {
+	socketCmd := append([]string{"/var/run/socket_vmnet", d.Program}, startCmd...)
```
AkihiroSuda, line 448: You have to provide the socket path
https://github.com/lima-vm/socket_vmnet#usage

klaases: I think we are providing the path?

Client: `"/opt/socket_vmnet/bin/socket_vmnet_client"`
Path: `/var/run/socket_vmnet`

On line 450, we take the client and append the socket path along with other arguments.

AkihiroSuda: Ah, yes, overlooked “/var/run” sorry

klaases: Sure, I appreciate your review.

polyverse-arm64.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: polyverse
```
AkihiroSuda, line 4: Seems a irrelevant commit

pkg/drivers/qemu/qemu.go
```diff
@@ -445,7 +445,9 @@ func (d *Driver) Start() error {
...
+	fmt.Printf("socketCmd: %v\n", socketCmd)
+	if stdout, stderr, err := cmdOutErr("/opt/socket_vmnet/bin/socket_vmnet_client", socketCmd...); err != nil {
```
AkihiroSuda, line 450: The paths should be probably configurable.
Also needs documentation.

polyverse-arm64.yaml
klaases: Yes, this one is just for testing, I was hoping to be able to get a service and/or tunnel running with Minikube + Qemu + socket_vmnet.

I'll remove this before merge.

pkg/drivers/qemu/qemu.go
klaases: That's a good idea, I was hoping to get a minimum viable product working. I'll see if I can make a flag for allowing for a custom `socket_vmnet_client` and `socket_vmnet` path.

I appreciate your review.

polyverse-arm64.yaml
klaases: Removed.

pkg/drivers/qemu/qemu.go
klaases: Updated PR with customizable flag paths for socket_vmnet binaries, please let me know if any more changes are needed.

pkg/drivers/qemu/qemu.go
```diff
@@ -408,6 +409,10 @@ func (d *Driver) Start() error {
 		startCmd = append(startCmd,
 			"-nic", fmt.Sprintf("user,model=virtio,hostfwd=tcp::%d-:22,hostfwd=tcp::%d-:2376,hostname=%s", d.SSHPort, d.EnginePort, d.GetMachineName()),
 		)
+	case "socket":
+		startCmd = append(startCmd,
+			"-nic", fmt.Sprintf("socket,model=virtio,sock=%s", d.NetworkSocket),
```
klaases, line 414: Hi @AkihiroSuda, in your experience, do these parameters for the `-nic` flag look correct for socket_vmnet with minikube socket network?

AkihiroSuda: Looks good

spowelljr (changes requested):
`d.Network` is hardcoded so it's currently impossible to use `socket`.

When I do manually change `d.Network` to `socket` it fails because of line:
`startCmd = append(startCmd, "-nic", fmt.Sprintf("socket,model=virtio,sock=%s", d.NetworkSocket)`

`d.NetworkSocket` isn't set anywhere in minikube, so it fails with:
`ERROR: qemu-system-aarch64: -nic socket,model=virtio,sock=: Invalid parameter 'sock'`

pkg/drivers/qemu/qemu.go
klaases: > `d.Network` is hardcoded so it's currently impossible to use `socket`.
> 
> When I do manually change `d.Network` to `socket` it fails because of line: `startCmd = append(startCmd, "-nic", fmt.Sprintf("socket,model=virtio,sock=%s", d.NetworkSocket)`
> 
> `d.NetworkSocket` isn't set anywhere in minikube, so it fails with: `ERROR: qemu-system-aarch64: -nic socket,model=virtio,sock=: Invalid parameter 'sock'`

Appreciate the comment, I have updated the code to reflect the network flag that is being passed into minikube rather than being hardcoded.

For now, if the network flag is empty, will default to the "user" network to maintain current functionality.

Once we have working flags for the qemu socket_vmnet network, will update the default network to "socket" for qemu.

klaases:
> `d.Network` is hardcoded so it's currently impossible to use `socket`.
> 
> When I do manually change `d.Network` to `socket` it fails because of line: `startCmd = append(startCmd, "-nic", fmt.Sprintf("socket,model=virtio,sock=%s", d.NetworkSocket)`
> 
> `d.NetworkSocket` isn't set anywhere in minikube, so it fails with: `ERROR: qemu-system-aarch64: -nic socket,model=virtio,sock=: Invalid parameter 'sock'`

Yes, makes sense. Fixed the hard-coded issue.

For now will default to "user" network to maintain backwards compatibility.

Once we have a set of working qemu + socket_vmnet flags, we'll update the default network to "socket".

pkg/drivers/qemu/qemu.go
klaases: Hi @AkihiroSuda, we found some errors with these flags. I'll omit them for now but we have a follow-up PR that should more closely express working flags for minikube + qemu + socket_vmnet, appreciate your patience and review.

spowelljr (approved):

PR #14894, opened 2022-08-31
Fix VerifyKubernetesImages test for k8s 1.25+

spowelljr:
Test fix for https://github.com/kubernetes/minikube/pull/14768

Add the new registry to the expected image list.


klaases (approved):
I looked into whether strings.Contains() could handle an array but the functionality is not built-in, so this works fine for our use case.

PR #14933, opened 2022-09-09
Fix minikube tunnel repeated printout of status

oldium:
The check to suppress repeated status printout does not work, it compares pointers, while the state is always cloned when it is passed to the Report function. Do the DeepEqual check instead and change the test to be real.


oldium:
Rebased to latest master.

spowelljr (approved):
Confirmed this works, thanks for the fix

PR #14934, opened 2022-09-09
Removed unused config value

spowelljr:
`NoTaintMaster` is no longer used in the template file.


klaases (approved):
I did not see `NoTaintMaster` available in the latest code, makes sense to remove it altogether.

https://pkg.go.dev/github.com/miciah/kubernetes/cmd/kubeadm/app/apis/kubeadm

Double checked and found no other references to `NoTaintMaster` in the minikube code base.

PR #14947, opened 2022-09-12
CI: fix Golang install on M1 Macs

spowelljr:
Error logs:
```
++ ./installers/check_install_golang.sh /usr/local
+ (( 1 < 1 ))
+ VERSION_TO_INSTALL=1.19.1
+ INSTALL_PATH=/usr/local
++ current_arch
++ case $(arch) in
+++ arch
+++ arch
++ echo 'unexpected arch: arm64. use amd64'
unexpected arch: arm64. use amd64
++ echo amd64
+ ARCH=amd64
+ check_and_install_golang
+ go version
+ sudo chown -R ****:**** /usr/local/go
chown: ****: illegal group name
++ true
```

Is trying to install amd64 version of Go on arm64 machines

### DIFF
--- a/hack/jenkins/installers/check_install_golang.sh
+++ b/hack/jenkins/installers/check_install_golang.sh
@@ -30,7 +30,7 @@ function current_arch() {
   "x86_64" | "i386")
      echo "amd64"
   ;;
-  "aarch64")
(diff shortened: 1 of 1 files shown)


klaases (approved):
Nice catch!

PR #14950, opened 2022-09-13
Fix QEMU delete errors

spowelljr:
Sometimes QEMU was returning two JSON objects in the response and minikube was failing to unmarshal the response as it's only expecting one JSON object.

Response:
```
{"return": {}}
{"timestamp": {"seconds": 1663094607, "microseconds": 836663}, "event": "SHUTDOWN", "data": {"guest": false, "reason": "host-qmp-quit"}}
```

Split the response on `\n` and only unmarshal the first JSON object.

Also wrapped returns on all the errors for better logging.

### DIFF
--- a/pkg/drivers/qemu/qemu.go
+++ b/pkg/drivers/qemu/qemu.go
@@ -709,7 +709,7 @@ func (d *Driver) RunQMPCommand(command string) (map[string]interface{}, error) {
 	}
 	var response qmpResponse
 	if err := json.Unmarshal(buf[:nr], &response); err != nil {
-		return nil, errors.Wrap(err, "unmarshal qmp_capabilities reponse")
+		return nil, errors.Wrap(err, "unmarshal qmp_capabilities response")
 	}
 	// expecting empty response
 	if len(response.Return) != 0 {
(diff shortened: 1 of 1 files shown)


medyagh (approved):

klaases (approved):

pkg/drivers/qemu/qemu.go
```diff
@@ -696,22 +695,21 @@ func (d *Driver) RunQMPCommand(command string) (map[string]interface{}, error) {
...
+		return nil, errors.Wrap(err, "marshal qmp_capabilities")
 	}
 	if _, err := conn.Write(jsonCommand); err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "write qmp_capabilities")
 	}
 	nr, err = conn.Read(buf[:])
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "read qmp_capabilities response")
```
medyagh, line 705: knit, but maybe "resp" instead of "response" for shorter wrapping exit message

klaases:
Great to see error wrapping for better logging.

pkg/drivers/qemu/qemu.go
spowelljr: Shortened to "resp"

Issue #14712, opened 2022-08-02
`minikube tunnel` bind address flag doesn't work on Linux


hal-shin:
### What Happened?

The ability to specify a bind address was added in https://github.com/kubernetes/minikube/pull/14245 but it doesn't seem to work on Linux (Ubuntu) while it works on Windows.

Not sure if it's an SSH issue, but the SSH version is 8.2.

### Attach the log file

$ minikube tunnel --bind-address=127.0.0.1 --alsologtostderr
I0803 02:36:33.295925   16878 out.go:296] Setting OutFile to fd 1 ...
I0803 02:36:33.296078   16878 out.go:348] isatty.IsTerminal(1) = true
I0803 02:36:33.296086   16878 out.go:309] Setting ErrFile to fd 2...
I0803 02:36:33.296095   16878 out.go:348] isatty.IsTerminal(2) = true
I0803 02:36:33.296256   16878 root.go:329] Updating PATH: /home/akash/.minikube/bin
I0803 02:36:33.296398   16878 mustload.go:65] Loading cluster: granular
I0803 02:36:33.296677   16878 config.go:178] Loaded profile config "granular": Driver=docker, ContainerRuntime=docker, KubernetesVersion=v1.23.1
I0803 02:36:33.297004   16878 cli_runner.go:164] Run: docker container inspect granular --format={{.State.Status}}
I0803 02:36:33.335247   16878 host.go:66] Checking if "granular" exists ...
I0803 02:36:33.335619   16878 cli_runner.go:164] Run: docker system info --format "{{json .}}"
[...]

PR #14951, opened 2022-09-14
Use SSH tunnel if user specifies bindAddress

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/14712

Currently, if the user is on Linux and using the Docker driver, specifying the `bindAddress` doesn't take effect as `driver.NeedsPortForward` returns false and the SSH tunnel is not used.

Updated the logic gating so if the user is on Linux and using the Docker driver and they specify the `bindAddress` it will use the SSH tunnel.

### DIFF
--- a/cmd/minikube/cmd/tunnel.go
+++ b/cmd/minikube/cmd/tunnel.go
@@ -85,7 +85,7 @@ var tunnelCmd = &cobra.Command{
 			cancel()
 		}()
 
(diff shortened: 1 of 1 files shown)


klaases (approved):
Summary: if it is not KIC, false, else use SSH tunnel if needs port forward or has a bind address.

PR #14956, opened 2022-09-15
fix: trim docker.io to prevent always writing to daemon

g1eny0ung:
Signed-off-by: Yue Yang <g1enyy0ung@gmail.com>

Since the `docker images` will output `Repository` without the `docker.io` prefix. I originally decided to trim this prefix from the `img` param in `ImageExistsInDaemon`.

But then I discovered the `trimDockerIO` function, so I move it from the `pkg/minikube/cruntime` package into `pkg/minikube/download` to achieve code reuse.

Thanks for the review. 😃

### DIFF
--- a/pkg/minikube/cruntime/docker.go
+++ b/pkg/minikube/cruntime/docker.go
@@ -604,7 +604,7 @@ func dockerImagesPreloaded(runner command.Runner, images []string) bool {
 	}
 	preloadedImages := map[string]struct{}{}
 	for _, i := range strings.Split(rr.Stdout.String(), "\n") {
-		i = trimDockerIO(i)
+		i = TrimDockerIO(i)
 		preloadedImages[i] = struct{}{}
 	}
 
@@ -612,7 +612,7 @@ func dockerImagesPreloaded(runner command.Runner, images []string) bool {
 
 	// Make sure images == imgs
 	for _, i := range images {
-		i = trimDockerIO(i)
+		i = TrimDockerIO(i)
 		if _, ok := preloadedImages[i]; !ok {
 			klog.Infof("%s wasn't preloaded", i)
 			return false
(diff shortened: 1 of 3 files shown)


g1eny0ung:
@medyagh Have a look when you have time? 👀

medyagh:
> @medyagh Have a look when you have time? 👀

Thanks for the PR, I am curious, could we add docker.io to all docker hub images instead of trimming it ? 
in our preload images tarball we include images with FQDN

g1eny0ung:
> > @medyagh Have a look when you have time? 👀
> 
> Thanks for the PR, I am curious, could we add docker.io to all docker hub images instead of trimming it ? in our preload images tarball we include images with FQDN

Makes sense to me. This problem is due to using the `docker images` command. It will check whether an image is from the docker official registry internally, if it is, the default domain will be removed.

But for now, the `docker images` command can not set an option to output the repo name in FQDN format, so I think the easiest and most straightforward way is to keep trimming the docker.io prefix.

g1eny0ung:
@medyagh PTAL again? This problem currently affects me every time I run `minikube start`.

g1eny0ung:
@blueelvis Hi, mind taking a look at this please?

spowelljr (changes requested):
Hi @g1eny0ung, this looks good, but could you move the trim function to: https://github.com/kubernetes/minikube/blob/master/pkg/minikube/image/image.go

The functions in that file seem to fit more in line with what you're doing here.

g1eny0ung:
> Hi @g1eny0ung, this looks good, but could you move the trim function to: https://github.com/kubernetes/minikube/blob/master/pkg/minikube/image/image.go
> 
> The functions in that file seem to fit more in line with what you're doing here.

Sure. Thanks for pointing out, I'll update it now.

spowelljr (approved):

PR #14988, opened 2022-09-21
Enabling RISC-V support

advancedwebdeveloper:
This is required to launch minikube on a 64bit RISC-V board

Currently testing under Qemu.

### DIFF
--- a/pkg/minikube/constants/constants.go
+++ b/pkg/minikube/constants/constants.go
@@ -27,7 +27,7 @@ import (
 
 var (
 	// SupportedArchitectures is the list of supported architectures
-	SupportedArchitectures = [5]string{"amd64", "arm", "arm64", "ppc64le", "s390x"}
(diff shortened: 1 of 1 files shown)


advancedwebdeveloper:
CC @lazyparser

afbjorklund:
Building minikube for another arch is a start, but it also needs the k8s binaries and the k8s images...

advancedwebdeveloper:
> > This is required to launch minikube on a 64bit RISC-V board
> > Currently testing under Qemu.
> 
> hi I dont think removing this check will add the support, the ISO would need to be re-built for the the new arch

That is just a required step.

medyagh:
> > > This is required to launch minikube on a 64bit RISC-V board
> > > Currently testing under Qemu.
> > 
> > 
> > hi I dont think removing this check will add the support, the ISO would need to be re-built for the the new arch
> 
> That is just a required step.

Once there is a support for RISC-V this would be the last step.

PR #14997, opened 2022-09-21
Fix update check logic

spowelljr:
**Before:**
Update check is skipped if `--interactive=true && --dry-run=true`

**After:**
I'm assuming the above is a mistake, refactored update check to skip if `--interactive=false`


medyagh (approved):
check function names if they still make sense

PR #15003, opened 2022-09-22
Fix to prevent accepting addons flag when no kubernetes

### LINKED ISSUE #13487: Output message if user tries to enable addons with `--no-kubernetes`

kpiljoong:
Fixes #13487

Disallow accepting addons flag when no kubernetes. Also, a new output string is added.

Before:
```
$ minikube start --no-kubernetes
$ minikube addons enable ingress
💡  ingress is an addon maintained by Kubernetes. For any concerns contact minikube on GitHub.
You can view the list of minikube maintainers at: https://github.com/kubernetes/minikube/blob/master/OWNERS
💡  After the addon is enabled, please run "minikube tunnel" and your ingress resources would be available at "127.0.0.1"
    ▪ Using image docker.io/jettech/kube-webhook-certgen:v1.5.1
    ▪ Using image docker.io/jettech/kube-webhook-certgen:v1.5.1
    ▪ Using image k8s.gcr.io/ingress-nginx/controller:v0.49.3
🔎  Verifying ingress addon...

❌  Exiting due to MK_ADDON_ENABLE: run callbacks: running callbacks: [sudo KUBECONFIG=/var/lib/
minikube/kubeconfig /var/lib/minikube/binaries/v0.0.0/kubectl apply -f /etc/kubernetes/addons/ingress-deploy.yaml: Process exited with status 1
stdout:

stderr:
sudo: /var/lib/minikube/binaries/v0.0.0/kubectl: command not found
 get kube-client to validate ingress addon: client config: context "minikube" does not exist: client config: context "minikube" does not exist]
```

After
```
$ minikube start --no-kubernetes
$ minikube addons enable ingress      

❌  Exiting due to MK_USAGE: You cannot enable addons on a cluster without Kubernetes, to enable Kubernetes on your cluster, run: minikube start --kubernetes-version=stable
```

Before
```
$ minikube start --no-kubernetes --addons ingress
😄  minikube v1.26.1 on Microsoft Windows 10 Pro 10.0.19043 Build 19043
👍  Starting minikube without Kubernetes in cluster minikube
🚜  Pulling base image ...
🐳  Preparing Docker 20.10.17 ...
🏄  Done! minikube is ready without Kubernetes!
```

After
```
$ minikube start --no-kubernetes --addons ingress
😄  minikube v1.27.0 on Microsoft Windows 10 Pro 10.0.19043 Build 19043

❌  Exiting due to MK_USAGE: You cannot enable addons on a cluster without Kubernetes, to enable Kubernetes on your cluster, run: minikube start --kubernetes-version=stable
```


medyagh (commented):
@kpiljoong I like this PR thank you, this makes a lot of sense !

one thing i could think of is, in the future, if we have an addon that does NOT require kubernetes, they should be allowed. but I believe all of our current addons do require kubernetes which is fine for this PR

cmd/minikube/cmd/start.go
```diff
@@ -294,6 +294,16 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 		validateDockerStorageDriver(driverName)
 	}
 
+	k8sVersion := getKubernetesVersion(existing)
+
+	// Disallow accepting addons flag without Kubernetes
+	// It places here because cluster config is required to get the old version.
+	if cmd.Flags().Changed(config.AddonListFlag) {
+		if k8sVersion == constants.NoKubernetesVersion || viper.GetBool(noKubernetes) {
+			exit.Message(reason.Usage, "Cannot enable addons without Kubernetes")
```
medyagh, line 303: how about add a message here that tells user what to do.
"You can not enable this addon with a cluster without Kubernetes, to re-configure your cluster with Kubernetes run this command
minikube start --no-kubernetes=false
 "

kpiljoong: Much better. I appreciate it. I just re-request a review.

spowelljr: The problem is `--no-kubernetes=false` won't actually re-enable Kubernetes. I'd recommend we change the command to `--kubernetes-version=stable`

cmd/minikube/cmd/config/enable.go
```diff
@@ -48,7 +48,7 @@ var addonsEnableCmd = &cobra.Command{
 			out.ErrT(style.Sad, `Unable to load config: {{.error}}`, out.V{"error": err})
 		}
 		if cc.KubernetesConfig.KubernetesVersion == constants.NoKubernetesVersion {
```
Aut0R3V, line 50: Hey, any updates on why this is not resolved?

Issue #13657, opened 2022-02-22
The CLIs Suggestion when starting minkube on hyperv without proper hyperv-module is missing a space


ja-nko:
minikube version: 1.24.0
windows version: Windows 10 Pro 10.0.19041 (windows is set to german)

I started minikube with `minikube start --driver=hyperv` without the Hyper-V PowerShell Module installed. The cli suggests running `Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-V-Tools-All`; but `Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-V-Tools -All` would be correct. (A space before `-All` is missing).

The command is correct in your [documentation](https://minikube.sigs.k8s.io/docs/drivers/hyperv/#enabling-hyper-v).

minikube suggesting the wrong command (Seen after "Suggestion: Run" 5th line from the bottom):
<img width="610" alt="k8s2" src="https://user-images.githubusercontent.com/48922273/155098297-eb5addea-fa2e-425f-8e84-2d41003747e3.PNG">

running the wrong and command and the correct one from your documentation:
<img width="613" alt="k8s1" src="https://user-images.githubusercontent.com/48922273/155098313-6884abeb-9540-420e-b911-756a37e5a1d5.PNG">

spowelljr:
Thanks for reporting the bug @ja-nko!

I created a PR to fix this

profnandaa:
@ja-nko -- The issue raised is correct but I think you wrongly assumed that it `Microsoft-V-Tools-All` was a typo, it just needed adding the `-All` switch after, hence the right command is:
```
Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Hyper-V-Tools-All -All
```

The command that you ended up running actually is a different one but I guess it serves the same purpose since all the related dependencies are also enabled:
> Enable-WindowsOptionalFeature -Online -FeatureName **Microsoft-Hyper-V** -All

PR #15012, opened 2022-09-26
fix: typo in known issue powershell command tip

profnandaa:
The powershell tip suggests a Windows Feature that is not there: `Microsoft-Hyper-V-Tools` instead of `Microsoft-Hyper-V-Tools-All`. This was probably because of the -All switch.

Runnning this on Win 10 Pro, 21H2:
```
PS C:\WINDOWS\system32> Get-WindowsOptionalFeature -Online | where FeatureName -Like "*Hyper-V*"


FeatureName : Microsoft-Hyper-V-All
State       : Enabled

FeatureName : Microsoft-Hyper-V
State       : Enabled

FeatureName : Microsoft-Hyper-V-Tools-All
State       : Enabled

FeatureName : Microsoft-Hyper-V-Management-PowerShell
State       : Enabled

FeatureName : Microsoft-Hyper-V-Hypervisor
State       : Enabled

FeatureName : Microsoft-Hyper-V-Services
State       : Enabled

FeatureName : Microsoft-Hyper-V-Management-Clients
State       : Enabled
```

The correct command should therefore be:
```ps
Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Hyper-V-Tools-All -All
```

This is a follow up of #13657, #13658
The original issue was wrongly reported, they assumed it was a typo when `-All` switch needed to be added instead, but I understand the naming confusion here. So, the origial tip was correct, only lacking `-All` switch at the end.

/cc. @spowelljr


spowelljr (approved):
Thanks for the fix!

PR #15071, opened 2022-10-04
tests: Shorten qmp path

spowelljr:
Most the tests on QEMU are failing due to the qmp path being too long.
```
I1004 10:38:26.286981    8680 main.go:134] libmachine: STDERR: qemu-system-aarch64: -qmp unix:/Users/jenkins/minikube-integration/darwin-arm64-qemu2--15070-5426-113d61471aea8d1d728ca9022a145b9e97990c51/.minikube/machines/default-k8s-different-port-20221004103818-6060/monitor,server,nowait: UNIX socket path '/Users/jenkins/minikube-integration/darwin-arm64-qemu2--15070-5426-113d61471aea8d1d728ca9022a145b9e97990c51/.minikube/machines/default-k8s-different-port-20221004103818-6060/monitor' is too long
Path must be less than 104 bytes
```

Shortened unnecessary info from paths while still ensuring they're unique.

**Before:**
`/Users/jenkins/minikube-integration/darwin-arm64-qemu2--15070-5426-113d61471aea8d1d728ca9022a145b9e97990c51/.minikube/machines/default-k8s-different-port-20221004103818-6060/monitor`
181 characters

**After:**
`/Users/jenkins/minikube-integration/15070-5426/.minikube/machines/default-k8s-diff-port-103818/monitor`
102 characters

### DIFF
--- a/hack/jenkins/common.sh
+++ b/hack/jenkins/common.sh
@@ -28,7 +28,7 @@
 
 readonly OS_ARCH="${OS}-${ARCH}"
 readonly TEST_ROOT="${HOME}/minikube-integration"
-readonly TEST_HOME="${TEST_ROOT}/${OS_ARCH}-${DRIVER}-${CONTAINER_RUNTIME}-${MINIKUBE_LOCATION}-$$-${COMMIT}"
+readonly TEST_HOME="${TEST_ROOT}/$$-${COMMIT:0:7}"
 
 export GOPATH="$HOME/go"
 export KUBECONFIG="${TEST_HOME}/kubeconfig"
--- a/hack/jenkins/common.sh
+++ b/hack/jenkins/common.sh
@@ -28,7 +28,7 @@
 
 readonly OS_ARCH="${OS}-${ARCH}"
 readonly TEST_ROOT="${HOME}/minikube-integration"
-readonly TEST_HOME="${TEST_ROOT}/$$-${COMMIT:0:7}"
+readonly TEST_HOME="${TEST_ROOT}/${MINIKUBE_LOCATION}-$$"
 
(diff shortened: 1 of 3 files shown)


hack/jenkins/common.sh
```diff
@@ -28,7 +28,7 @@
 
 readonly OS_ARCH="${OS}-${ARCH}"
 readonly TEST_ROOT="${HOME}/minikube-integration"
-readonly TEST_HOME="${TEST_ROOT}/${OS_ARCH}-${DRIVER}-${CONTAINER_RUNTIME}-${MINIKUBE_LOCATION}-$$-${COMMIT}"
+readonly TEST_HOME="${TEST_ROOT}/$$-${COMMIT:0:7}"
```
medyagh, line 31: how about we keep the ${MINIKUBE_LOCATION}
for when debugging manually looking into the machine, if something be probalamatic, we can quickly find out what PR that is causing it

medyagh (changes requested):

hack/jenkins/common.sh
spowelljr: I can but I'll have to shorten some profile names unless there's something else you think we can remove

spowelljr: I replaced `COMMIT` with `MINIKUBE_LOCATION`

medyagh (approved):

hack/jenkins/common.sh
medyagh: minikube location alone could be problematic, if a PR get tested couple times and previous workspace as stuff in it.
can we make it like MINIKUBE_LOCATION-4Digit comm it ?

Issue #14980, opened 2022-09-19
Add Documentation on using Google Artifact Registry + service account + Permission)


medyagh:
### What Happened?

based on User survey
that asked us to ```
detailed and fully functional docs related with cloud provider registry integration (Google Artifact Registry + service account + Permission)
```


to do this task, one would need to create a Private google artifact registry
and push an image there
and then try to pull that image inside minikube

I suspect this would be do-able if the user enbables the GCP-Auth Addon in minikube and it should work
if it does there should be a new tutorial in our website that explains step by step of using Google Artifact Registry

medyagh:
More clarifications for this task:

- push ngnix image to private googel a.r
- deploy a pod (yaml) to minikube using that image
- ensure that pod is working (running)

check if Gcp-auth addon will let the image be pulled.

PR #15072, opened 2022-10-04
Site: Google Artifact Registry Integration

klaases:
Fixes: #14980

Documents the process involved with pushing a private image to Google's artifact registry, and then pulling that image into a minikbue pod while being authenticated with GCR.

### DIFF
--- a/site/content/en/docs/handbook/registry_integration.md
+++ b/site/content/en/docs/handbook/registry_integration.md
@@ -1,0 +1,95 @@
+---
+title: "Registry Integration"
+linkTitle: "Registry Integration"
+weight: 6
+description: >
+  Cloud provider registry integration (Google Artifact Registry + Service Account + Permission)
+aliases:
+ - /docs/tasks/registry_integration
+---
+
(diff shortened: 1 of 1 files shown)


klaases:
Does anyone know how to push a Nginx image to a private Google Artifact Registry using GCP-Auth?

medyagh (changes requested):
plz update the PR description Before and After

the goal is before enalbing gcp-auth addon, we should see the Pod not being started with a Private Image in A.R

and with gcp-auth addon enabled the appying new image should work (delete pod first then apply again)

rahil-p:
Had a couple of follow-up points:

1. I think this should be in "Tutorials" instead of "Handbook".  This page seems to fit the bill of a tutorial — it guides users through a specific use case with step-by-step examples.

2. If I'm not mistaken, the first section has some redundancy.  The `gcloud auth configure-docker` command (no longer in beta) authenticates Docker to Artifact Registry — the next two sections repeat this using a different method (standalone credential helper). [more info](https://cloud.google.com/artifact-registry/docs/docker/authentication#methods)

PR #15074, opened 2022-10-05
fix: handle DOS-style path for Dockerfile in minikube build

### LINKED ISSUE #15007: minikube image build path handling issue in Windows
### What Happened?

- v1.27.0 minikube-windows-amd64.exe
- Root of the project contains .dockerignore file with a line **/Dockerfile*
- minikube image build -t myproject:latest -f subfolder/Dockerfile .
--> C:\Users\user\AppData\Local\Temp\build.xxxxxxxxxx.tar is missing subfolder\Dockerfile and the build obviously fails within Minikube VM due to missing Dockerfile
- minikube image build -t myproject:latest -f subfolder\Dockerfile .
--> C:\Users\user\AppData\Local\Temp\build.xxxxxxxxxx.tar contains subfolder\Dockerfile, but the build obviously fails within Minikube VM because subfolder\Dockerfile is not a valid Linux path
- Workaround: remove .dockerignore and use path subfolder/Dockerfile

### Attach the log file

-

### Operating System

Windows

### Driver

VMware

ComradeProgrammer:
fixes #15007


cmd/minikube/cmd/image.go
```diff
@@ -310,6 +310,8 @@ var buildImageCmd = &cobra.Command{
 				// Otherwise, assume it's a tar
 			}
 		}
+		//if dockerFile is a DOS path, translate it into UNIX path
+		dockerFile = strings.ReplaceAll(dockerFile, "\\", "/")
```
medyagh, line 314: @ComradeProgrammer do you mind pasting an example Dockerfile and the error that you get?

this change seems to be dangrous, I suggest to limit this line to only run on Windows !
it is better that the user Fixes the Dockerfile themselves and minikube gives them an educational instruciton how to do that

medyagh (changes requested):

cmd/minikube/cmd/image.go
ComradeProgrammer: Well, I don't think it is a problem which can be solved by user's dockerfile. It is normal to include Dockerfile in the .dockerignore so that the Dockerfile itself won't be packed into the image.

However, when user is using minikube on windows, in order to prevent the Dockerfile from being ignored when we try to tar the workspace, user need to specify the path to the Dockerfile, which should be a DOS-style path. However, when this image is built, the image is built on the virtual machine we created, which requires a UNIX-style path.

Therefore if user pass the DOS-style path of Dockerfile, virtual machine cannot build this image because the path is not  a valid UNIX path; if user pass the UNIX-style path, the Dockerfile will be ignored and wont't be packed into the tar file which we send to the virtual machine.

Unless users waive the thought of preventing Dockerfile to be packed into image, and remove Dockerfile from .dockerignore, there is no way for user to prevent this bug from being triggered.

ComradeProgrammer: The example has been updated into the comment of this pr. Besides, changes have been made to limit this line to only run on Windows

ComradeProgrammer:
Here is an example to reproduce the problem. The folder is
```
.
├── .dockerignore
└── test
    └── Dockerfile
```

The content of .dockerignore
```
**/Dockerfile*
```

The content of dockerfile
```
FROM ubuntu:latest

CMD ["/bin/bash", "-c"]
```

Output:
When I execute ` .\minikube-windows-amd64.exe image build -t aaa:latest -f test\Dockerfile .`, using DOS-style path
```
PS D:\workspace\minikube> .\minikube-windows-amd64.exe image build -t aaa:latest -f test\Dockerfile .
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat var/lib/minikube/build/build.519194499/test\Dockerfile: no such file or directory
```
When I execute ` .\minikube-windows-amd64.exe image build -t aaa:latest -f test\Dockerfile .`, using UNIX-style path
```
PS D:\workspace\minikube> .\minikube-windows-amd64.exe image build -t aaa:latest -f test/Dockerfile .
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /var/lib/minikube/build/build.4139476469/test/Dockerfile: no such file or directory
```

cmd/minikube/cmd/image.go
```diff
@@ -310,6 +310,10 @@ var buildImageCmd = &cobra.Command{
 				// Otherwise, assume it's a tar
 			}
 		}
+		if runtime.GOOS == "windows" {
+			//if dockerFile is a DOS path, translate it into UNIX path
```
spowelljr, line 314:
```suggestion
			// if dockerFile is a DOS path, translate it into UNIX path
```

spowelljr (changes requested):

spowelljr:
Tested this and confirms it works

**Before:**
```
> minikube image build -t thing:latest -f thing2\Dockerfile .
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /var/lib/minikube/build/build.3965132197/thing2\Dockerfile: no such file or directory
```

**After:**
```
> minikube image build -t thing:latest -f thing2\Dockerfile .
Sending build context to Docker daemon   2.56kB
Step 1/2 : FROM ubuntu:latest
 ---> 216c552ea5ba
Step 2/2 : CMD ["/bin/bash", "-c"]
 ---> Running in 5b64e4107d63
Removing intermediate container 5b64e4107d63
 ---> a5b667236e51
Successfully built a5b667236e51
Successfully tagged thing:latest
```

medyagh (changes requested):
@ComradeProgrammer Lets make sure this replace is only suggested by minikube and not automatically done, if minikube gonna replace it, it has to get user permission to clearly say "Yes" it is better to leave this operation to the user

ComradeProgrammer:
> @ComradeProgrammer Lets make sure this replace is only suggested by minikube and not automatically done, if minikube gonna replace it, it has to get user permission to clearly say "Yes" it is better to leave this operation to the user

This function has been added, now the we need user's permission to replace the \

cmd/minikube/cmd/image.go
```diff
@@ -310,6 +312,22 @@ var buildImageCmd = &cobra.Command{
 				// Otherwise, assume it's a tar
 			}
 		}
+		if runtime.GOOS == "windows" && strings.Contains(dockerFile, "\\") {
+			// if dockerFile is a DOS path, translate it into UNIX path
+			// but we need to ask for permission from user first
+			out.String("minikube detects that you are using DOS-style path %s. Should minikube convert it to UNIX-style by replacing all \\ to /? [y/n]", dockerFile)
+			var answer string
+			for {
+				fmt.Scan(&answer)
```
spowelljr, line 321: Should check for error

spowelljr, line 321: Many people use minikube in their CI for testing. If someone is running this command in a script it will hang forever since there will be no input.

Either some sort of interaction flag will need to be added so users running via scripts can bypass this, or instead of automating this we can just suggest that the user makes this change themself.

cmd/minikube/cmd/image.go
```diff
@@ -310,6 +312,22 @@ var buildImageCmd = &cobra.Command{
...
+				answer := strings.ToLower(answer)
+				if answer == "y" || answer == "yes" || answer == "n" || answer == "no" {
+					break
+				}
+			}
+			if answer == "y" || answer == "yes" {
```
spowelljr, line 327: `answer` is always going to be empty here because you're creating `answer` with a new scope within the `for` loop on L322 because of the `:=`. If the `:` is removed it will update the value instead of creating a new one.

spowelljr (changes requested):

ComradeProgrammer:
@spowelljr Indeed you are right, so I changed the code. Instead of asking for user's input, I switched to command line arguments.
I added a param called "--no-dospath-conversion". By default the DOS-like path is automatically converted, but if user use this --no-dospath-conversion then there will be no conversion. Do you think it is better this time?

spowelljr (changes requested):
`make generate-docs` to update the new command on the website

ComradeProgrammer:
Anything I need to do next?
@medyagh  @spowelljr  @afbjorklund

andez2000:
Is this going to get done?  I have an issue in my build with specifying the files to the command line

spowelljr (changes requested):

test/integration/functional_test.go
```diff
@@ -299,9 +299,10 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
 	// docs: Make sure image building works by `minikube image build`
 	t.Run("ImageBuild", func(t *testing.T) {
 		MaybeParallel(t)
-
-		if _, err := Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "pgrep", "buildkitd")); err == nil {
-			t.Errorf("buildkitd process is running, should not be running until `minikube image build` is ran")
+		if runtime.GOOS != "windows" {
```
spowelljr, line 302: I think there might have been some confusion on my last comment. We should check do this check on Windows as well, so the if statement should be removed.

What I was trying to say is that on line 311, if not on Windows the path to the Dockerfile should continue being `filepath.Join(*testdataDir, "build")`, but on Windows we should use `"testdata\\build"` to test this change.

spowelljr (changes requested):

test/integration/functional_test.go
ComradeProgrammer: Actually I found that on Windows, the result of filepath.Join(*testdataDir, "build") is exactly "testdata\\build", so actually we don't need to modify this part, or write a seperate test.

test/integration/functional_test.go
```diff
@@ -299,7 +299,7 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
 	// docs: Make sure image building works by `minikube image build`
 	t.Run("ImageBuild", func(t *testing.T) {
 		MaybeParallel(t)
-
+
```
spowelljr, line 302: Could you remove this whitespace change?

spowelljr (changes requested):

test/integration/functional_test.go
spowelljr: Ah, gotcha, thanks for the info

ComradeProgrammer:
@spowelljr  It seems that the test under Docker_Windows has passed

cmd/minikube/cmd/image.go
```diff
@@ -310,6 +311,12 @@ var buildImageCmd = &cobra.Command{
 				// Otherwise, assume it's a tar
 			}
 		}
+		if runtime.GOOS == "windows" && strings.Contains(dockerFile, "\\") {
+			// if dockerFile is a DOS path, and user did not explicitly forbid this conversion, translate it into UNIX path
```
spowelljr, line 315:
```suggestion
			// if dockerFile is a DOS path, translate it into UNIX path
```
With the changes the user can no longer forbid this, so comment is incorrect

spowelljr (changes requested):

spowelljr (approved):
PR looks good, thanks for all the changes

PR #15110, opened 2022-10-11
Add 'qemu' to list of issue template drivers

klaases:
Fixes: #15111

Adds an additional network driver 'Qemu' to the valid list of Drivers used in creating a new issue.

Only EN and RU templates use the driver dropdown.

**Before:**

Screen Shot 2022-10-11 at 16 05 56

**After:**

Same as above but with `Qemu` as a drop down option.

### DIFF
--- a/.github/ISSUE_TEMPLATE/__en-US.yaml
+++ b/.github/ISSUE_TEMPLATE/__en-US.yaml
@@ -49,6 +49,6 @@ body:
         - SSH
         - VMware
         - Parallels
-        - Qemu
+        - QEMU
     validations:
       required: false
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/__en-US.yaml
```diff
@@ -49,5 +49,6 @@ body:
         - SSH
         - VMware
         - Parallels
+        - Qemu
```
spowelljr, line 52:
```suggestion
        - QEMU
```

.github/ISSUE_TEMPLATE/ru.yaml
```diff
@@ -47,5 +47,6 @@ body:
         - SSH
         - VMware
         - Parallels
+        - Qemu
```
spowelljr, line 50:
```suggestion
        - QEMU
```

spowelljr (changes requested):

spowelljr (approved):

Issue #14202, opened 2022-05-21
minikube on driver docker not running using docker desktop vm


prakharporwal:
### What Happened?

- I run `minikube start`
I get above logs.

![image](https://user-images.githubusercontent.com/49529937/169645450-a18d8edd-ce2d-4f16-a060-77df5372bedd.png)

I have tried multiple times 
I tried with driver virtualbox but failes because virtual box and docker desktop cannot run together ( because 2 VM cannot run together on Ubuntu)

`minikube status` gives this

![image](https://user-images.githubusercontent.com/49529937/169645376-cdb31689-739b-4544-8e13-fd9176b4f092.png)

 I have tried 
- reinstallng minikube  
- restarting the minikube container
- tried `minikube delete --all` and `minikube start`
Not working in all the above cases.

### Attach the log file

[logs.txt](https://github.com/kubernetes/minikube/files/8747325/logs.txt)

### Operating System

Ubuntu

### Driver

Docker

afbjorklund:
I don't think Docker Desktop (VM) has bern tested on Ubuntu, only Docker Engine...

prakharporwal:
Is it possible to give some fix for now from minikube end ?

loyaltytrooper:
Yes I am also getting the same error

![image](https://user-images.githubusercontent.com/89660252/169927608-64b43dd0-36bc-4af9-83e9-3ebeb5398abf.png)

afbjorklund:
Please use text instead of images.

mesketh:
Same issue:

---8<-----

* ==> Last Start <==
* Log file created at: 2022/09/12 12:08:55
Running on machine: markx1
Binary: Built with gc go1.18.3 for linux/amd64
Log line format: [IWEF]mmdd hh:mm:ss.uuuuuu threadid file:line] msg
I0912 12:08:55.207032 1552433 out.go:296] Setting OutFile to fd 1 ...
I0912 12:08:55.207148 1552433 out.go:348] isatty.IsTerminal(1) = true
I0912 12:08:55.207151 1552433 out.go:309] Setting ErrFile to fd 2...
I0912 12:08:55.207155 1552433 out.go:348] isatty.IsTerminal(2) = true
I0912 12:08:55.207240 1552433 root.go:333] Updating PATH: /home/mark/.minikube/bin
I0912 12:08:55.207553 1552433 out.go:303] Set.ting JSON to false
[...]

PR #15126, opened 2022-10-12
use port forwarding on Linux with Docker Desktop

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/14202

Detects if the user is using Docker Desktop on Linux and requires port forwarding if they are.

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -611,6 +611,9 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 			klog.Info("auto-setting LocalStorageCapacityIsolation to false because using btrfs storage driver")
 			cc.KubernetesConfig.FeatureGates = addFeatureGate(cc.KubernetesConfig.FeatureGates, "LocalStorageCapacityIsolation=false")
 		}
+		if runtime.GOOS == "linux" && si.DockerOS == "Docker Desktop" {
+			out.WarningT("For an improved experience it's recommended to use Docker Engine over Docker Desktop")
+		}
 	}
 
 	return cc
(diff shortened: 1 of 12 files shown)


pkg/drivers/kic/oci/info.go
```diff
@@ -35,6 +35,7 @@ type SysInfo struct {
 	Rootless      bool     // Weather or not the docker is running on rootless mode
 	StorageDriver string   // the storage driver for the daemon  (for example overlay2)
 	Errors        []string // any server issues
+	DockerOS      string   // used to detect if using Docker Desktop or Docker Engine on Linux
```
medyagh, line 38: what is OS short for here?

cmd/minikube/cmd/start_flags.go
```diff
@@ -611,6 +611,9 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 			klog.Info("auto-setting LocalStorageCapacityIsolation to false because using btrfs storage driver")
 			cc.KubernetesConfig.FeatureGates = addFeatureGate(cc.KubernetesConfig.FeatureGates, "LocalStorageCapacityIsolation=false")
 		}
+		if runtime.GOOS == "linux" && si.DockerOS == "Docker Desktop" {
+			out.WarningT("For an improved experience it's recommended to use Docker Engine over Docker Desktop")
```
medyagh, line 615: instead of ...

also consider adding a todo: follow up add page to site

medyagh (changes requested):

pkg/drivers/kic/oci/info.go
spowelljr: Operating System

cmd/minikube/cmd/start_flags.go
spowelljr: Update message to include a link to Docker Engine install

medyagh (approved):

PR #15137, opened 2022-10-14
Enable Parallels driver on darwin arm64

spowelljr:
Creating a PR so a user can test the Parallels driver on darwin arm64. If it works successfully we can merge the PR.


spowelljr:
Test was unsuccessful, closing

PR #15163, opened 2022-10-18
test: add integration test cases for minikube image build

ComradeProgrammer:
test: add 5 integration test cases for "minikube image build"


test/integration/image_test.go
```diff
@@ -0,0 +1,132 @@
...
+			{"BuildWithBuildArg", validateImageBuildWithBuildArg},
+			{"BuildWithDockerIgnore", validateImageBuildWithDockerIgnore},
+			{"BuildWithSpecifiedDockefile", validateNormalImageBuildWithSpecifiedDockefile},
+		}
+		for _, tc := range tests {
+			if ctx.Err() == context.DeadlineExceeded {
+				t.Fatalf("Unable to run more tests (deadline exceeded)")
+			}
+			if t.Failed() {
+				t.Fatalf("Previous test failed, not running dependent tests")
```
spowelljr, line 52: These tests don't seem to be dependent on each other so this breakout can be removed.

test/integration/image_test.go
```diff
@@ -0,0 +1,132 @@
...
+// normal test case for minikube image build, with -t and -f parameter
+func validateNormalImageBuildWithSpecifiedDockefile(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+	args := append([]string{"image", "build", "-t", "aaa:latest", "-f", "inner/Dockerfile", "./testdata/image-build/test-f", "-p", profile}, StartArgs()...)
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), args...))
+	data, _ := io.ReadAll(rr.Stdout)
+
+	output := string(data)
+	fmt.Println(output)
+	if err != nil {
```
spowelljr, line 84: Is there a reason to not have the err check right after the func? I think it would be best to stick with the standard and have it directly after.

ie.
```
rr, err := Run(t, exec.CommandContext(ctx, Target(), args...))
if err != nil {
        t.Fatalf("failed to build image with args: %q : %v", rr.Command(), err)
}
...
```

test/integration/image_test.go
```diff
@@ -0,0 +1,132 @@
...
+	defer PostMortemLogs(t, profile)
+	args := append([]string{"image", "build", "-t", "aaa:latest", "./testdata/image-build/test-normal", "-p", profile}, StartArgs()...)
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), args...))
+	if err != nil {
+		t.Fatalf("failed to start minikube with args: %q : %v", rr.Command(), err)
+	}
+}
+
+// normal test case for minikube image build, with -t and -f parameter
+func validateNormalImageBuildWithSpecifiedDockefile(ctx context.Context, t *testing.T, profile string) {
```
spowelljr, line 76:
```suggestion
func validateNormalImageBuildWithSpecifiedDockerfile(ctx context.Context, t *testing.T, profile string) {
```

test/integration/image_test.go
```diff
@@ -0,0 +1,132 @@
...
+	// Serial tests
+	t.Run("serial", func(t *testing.T) {
+		tests := []struct {
+			name      string
+			validator validateFunc
+		}{
+			{"NormalBuild", validateNormalImageBuild},
+			{"BuildWithBuildArg", validateImageBuildWithBuildArg},
+			{"BuildWithDockerIgnore", validateImageBuildWithDockerIgnore},
+			{"BuildWithSpecifiedDockefile", validateNormalImageBuildWithSpecifiedDockefile},
```
spowelljr, line 45:
```suggestion
			{"BuildWithSpecifiedDockerfile", validateNormalImageBuildWithSpecifiedDockerfile},
```

test/integration/image_test.go
```diff
@@ -0,0 +1,132 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+func TestImageBuild(t *testing.T) {
```
spowelljr, line 30: This func should have a comment above it in the format of `// TestImageBuild <explanation>` and then after you make that change run `make generate-docs` as that will update the test documentation on the website.

test/integration/image_test.go
```diff
@@ -0,0 +1,132 @@
...
+	})
+}
+
+// normal test case for minikube image build, with -t parameter
+func validateNormalImageBuild(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+	args := append([]string{"image", "build", "-t", "aaa:latest", "./testdata/image-build/test-normal", "-p", profile}, StartArgs()...)
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), args...))
+	if err != nil {
+		t.Fatalf("failed to start minikube with args: %q : %v", rr.Command(), err)
```
spowelljr, line 71:
```suggestion
		t.Fatalf("failed to build image with args: %q : %v", rr.Command(), err)
```

test/integration/image_test.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// normal test case for minikube image build, with -t and -f parameter
+func validateNormalImageBuildWithSpecifiedDockefile(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+	args := append([]string{"image", "build", "-t", "aaa:latest", "-f", "inner/Dockerfile", "./testdata/image-build/test-f", "-p", profile}, StartArgs()...)
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), args...))
+	data, _ := io.ReadAll(rr.Stdout)
+
+	output := string(data)
```
spowelljr, line 82: This value isn't being used anywhere, not sure if you forgot to do something with it, but if it's not needed it should be removed.

test/integration/image_test.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+// build with .dockerignore
+func validateImageBuildWithDockerIgnore(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+	args := append([]string{"image", "build", "-t", "aaa:latest", "./testdata/image-build/test-normal", "--build-opt=no-cache", "-p", profile}, StartArgs()...)
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), args...))
+	if err != nil {
+		t.Fatalf("failed to start minikube with args: %q : %v", rr.Command(), err)
```
spowelljr, line 130:
```suggestion
		t.Fatalf("failed to build image with args: %q : %v", rr.Command(), err)
```

test/integration/image_test.go
```diff
@@ -0,0 +1,132 @@
...
+func validateNormalImageBuildWithSpecifiedDockefile(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+	args := append([]string{"image", "build", "-t", "aaa:latest", "-f", "inner/Dockerfile", "./testdata/image-build/test-f", "-p", profile}, StartArgs()...)
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), args...))
+	data, _ := io.ReadAll(rr.Stdout)
+
+	output := string(data)
+	fmt.Println(output)
+	if err != nil {
+		t.Fatalf("failed to start minikube with args: %q : %v", rr.Command(), err)
```
spowelljr, line 85:
```suggestion
		t.Fatalf("failed to build image with args: %q : %v", rr.Command(), err)
```

test/integration/image_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+// test case for bug in https://github.com/kubernetes/minikube/issues/12384
+// build with --build-opt
+func validateImageBuildWithBuildArg(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+	args := append([]string{"image", "build", "-t", "aaa:latest", "--build-opt=build-arg=ENV_A=aaasdftggdf", "--build-opt=no-cache", "./testdata/image-build/test-arg", "-p", profile}, StartArgs()...)
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), args...))
+	data, _ := io.ReadAll(rr.Stdout)
+
+	output := string(data)
```
spowelljr, line 97: This can be simplified:
```suggestion
	output := rr.Stdout.String()
```

test/integration/testdata/image-build/test-arg/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM ubuntu:latest
```
spowelljr, line 1: `ubuntu:latest` is quite a large image, I'd recommend using a smaller image if you can, I'm sure there are a few options.

We use `gcr.io/k8s-minikube/busybox` for a few other tests, I'm not 100% sure if it has `echo` & `/bin/bash` though.

test/integration/image_test.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		t.Fatalf("failed to build image with args: %q : %v", rr.Command(), err)
+	}
+	if !strings.Contains(output, "aaasdftggdf") {
+		t.Fatalf("failed to pass build-args with args: %q : %v", rr.Command(), err)
+	}
+}
+
+// current this test cannot be passed because issue https://github.com/kubernetes/minikube/issues/12431 hasn't been fixed, so this test case is not enabled
```
spowelljr, line 106: It makes sense to skip the test, but the way it's being skipped makes it hard to notice it's being skipped without reading the comment. I'd recommend you add the test to test list above and add the following to the first line of the func below.
```
t.Skip("skipping due to https://github.com/kubernetes/minikube/issues/12431")
```

It will show up as being skipped in test output.

test/integration/image_test.go
```diff
@@ -0,0 +1,132 @@
...
+				}
+			})
+		}
+	})
+}
+
+// normal test case for minikube image build, with -t parameter
+func validateNormalImageBuild(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+	args := append([]string{"image", "build", "-t", "aaa:latest", "./testdata/image-build/test-normal", "-p", profile}, StartArgs()...)
```
spowelljr, line 68: Start args are only when starting minikube.
```suggestion
	args := []string{"image", "build", "-t", "aaa:latest", "./testdata/image-build/test-normal", "-p", profile}
```

test/integration/image_test.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		t.Fatalf("failed to start minikube with args: %q : %v", rr.Command(), err)
+	}
+}
+
+// test case for bug in https://github.com/kubernetes/minikube/issues/12384
+// build with --build-opt
+func validateImageBuildWithBuildArg(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+	args := append([]string{"image", "build", "-t", "aaa:latest", "--build-opt=build-arg=ENV_A=aaasdftggdf", "--build-opt=no-cache", "./testdata/image-build/test-arg", "-p", profile}, StartArgs()...)
```
spowelljr, line 93: Nothing wrong with `aaasdftggdf` but could we use a more readable value.

spowelljr (changes requested):

ComradeProgrammer:
Thank you for the suggestions given in the review. All suggestions have been taken, and revisions have been made accordingly.

spowelljr:
Hi @ComradeProgrammer, can you take a look at the lint errors. It looks like you're using the `append` func but you're only passing it one param, so you can remove the `append` part and just set the variable to the array.

test/integration/image_test.go
```diff
@@ -0,0 +1,127 @@
...
+				if t.Failed() && *postMortemLogs {
+					PostMortemLogs(t, profile)
+				}
+			})
+		}
+	})
+}
+
+// normal test case for minikube image build, with -t parameter
+func validateNormalImageBuild(ctx context.Context, t *testing.T, profile string) {
```
spowelljr, line 60: All the `validate` functions should also have a comment above it in the format `// validateNormalImageBuild <explanation>`

spowelljr (changes requested):

ComradeProgrammer:
> Hi @ComradeProgrammer, can you take a look at the lint errors. It looks like you're using the `append` func but you're only passing it one param, so you can remove the `append` part and just set the variable to the array.

Thanks a lot, lint errors and comments has been fixed

spowelljr (changes requested):
Looks good, could you just run `make generate-docs` and push up the documentation changes, thanks

ComradeProgrammer:
> Looks good, could you just run `make generate-docs` and push up the documentation changes, thanks
docs has been added

ComradeProgrammer:
> /ok-to-test

I saw there are so many errors in the test, are these failures caused by my pr?

ComradeProgrammer:
Is there anything that I should do next? @spowelljr

test/integration/image_test.go
```diff
@@ -0,0 +1,127 @@
...
+	// Serial tests
+	t.Run("serial", func(t *testing.T) {
+		tests := []struct {
+			name      string
+			validator validateFunc
+		}{
+			{"NormalBuild", validateNormalImageBuild},
+			{"BuildWithBuildArg", validateImageBuildWithBuildArg},
+			{"BuildWithDockerIgnore", validateImageBuildWithDockerIgnore},
+			{"BuildWithSpecifiedDockefile", validateNormalImageBuildWithSpecifiedDockerfile},
```
spowelljr, line 45:
```suggestion
			{"BuildWithSpecifiedDockerfile", validateNormalImageBuildWithSpecifiedDockerfile},
```

test/integration/image_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+// validateNormalImageBuildWithSpecifiedDockerfile is normal test case for minikube image build, with -t and -f parameter
+func validateNormalImageBuildWithSpecifiedDockerfile(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+	args := []string{"image", "build", "-t", "aaa:latest", "-f", "inner/Dockerfile", "./testdata/image-build/test-f", "-p", profile}
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), args...))
+	if err != nil {
+		t.Fatalf("failed to build image with args: %q : %v", rr.Command(), err)
+	}
+	_, err = io.ReadAll(rr.Stdout)
```
spowelljr, line 77: I don't really see what this is checking, we're throwing away that value we're reading. Plus we could simply get this value with `rr.Stdout.String()`. I think this should be removed.

test/integration/testdata/image-build/test-normal/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM alpine:latest
+
+CMD ["/bin/sh", "-c"]
```
spowelljr, line 3: There are three files with the exact same contents, I'd rather we just have one and modify the tests to point to the same file.

spowelljr (changes requested):

test/integration/testdata/image-build/test-normal/Dockerfile
ComradeProgrammer: Well I think we cannot modify the tests to point to the same file, because the path to the file is exactly the variable in different tests. 
The first file is in a multiple-layer folder , which means '\' will appear in the path
The second file is in a single-layer folder, which means '\' will not appear in the path
The third file is in a special folder with a .dockerignore file. 
We cannot mix them up.

ComradeProgrammer: Other comments are all accepted.

spowelljr: Gotcha, thanks for the explanation

ComradeProgrammer:
please trigger the test again. Just limited the test on docker engines(not containered). Thanks. @spowelljr

ComradeProgrammer:
Well I think it seems that everything is fine now, is there any other suggestions?

medyagh (approved):

ComradeProgrammer:
Anything I need to do next?
@medyagh  @spowelljr  @afbjorklund

spowelljr (approved):

PR #15172, opened 2022-10-19
Portainer addon version bump to 2.15.1 & HTTPS access enabled

jevontane:


spowelljr:
Hi @jevontane, I can confirm that this works. Would you be interested in adding an integration test for the Portainer addon?

Here's the addons test file for reference: https://github.com/kubernetes/minikube/blob/master/test/integration/addons_test.go

We just want to make sure in the future the addon continues to work. Without integration tests for the addon we can't guarantee it will continue working with upgrades such as future version of Kubernetes as we won't be able to tell when it stops working.

spowelljr (approved):

PR #15244, opened 2022-10-29
Get gateway ip from lima host DNS record on podman/LimaVM

rastaman:
# Context

When installing minikube on MacOS on a VM built with [lima-vm/lima: Linux virtual machines, typically on macOS, for running containerd](https://github.com/lima-vm/lima), minikube fails to find the gateway ip and complain about it:

```sh
./out/minikube -p miniforge start --rootless=false --container-runtime=containerd --cpus=4 --cni=cilium --disk-size=20000mb  --dns-domain=cluster.local --driver=podman --memory=7916m --nodes=1 
...
📦  Préparation de Kubernetes v1.25.3 sur containerd 1.6.9...
E1029 11:35:16.036962   77006 start.go:130] Unable to get host IP: RoutableHostIPFromInside is currently only implemented for linux
...
🏄  Terminé ! kubectl est maintenant configuré pour utiliser "miniforge" cluster et espace de noms "default" par défaut.
```

This PR add a lookup on the `host.lima.internal` record name to find the gateway for Darwin OSs, in a similar way it is done for `docker` hosts a few lines above in the code.

# Implementation

In `pkg/drivers/kic/oci/network.go`, if running under MacOS, get the gateway from the `host.lima.internal` record which is added for hosts provisioned by Lima VM (see: [`The host.lima.internal name is predefined to specify the gateway address to the host`](https://github.com/lima-vm/lima/blob/082576f99cb4fbb3844b69df21c2e13c65bd58be/examples/default.yaml#L344)).

The check on the name occur only for Darwin OS with podman driver as Lima is mostly use with them to my knowledge.

# Tests

This has been tested with a VM build with Lima v0.12.0, with the podman driver and minikube master.

After applying this patch, the startup text doesn't complain anymore about not finding the gateway ip.

At one point i added a debug statement which was shown in the logs:
```sh
$ minikube logs | grep ociBin
I1029 12:57:11.667710   82495 network.go:91] Will use the following gateway IP: 192.168.5.2 (ociBin: podman, GOOS: darwin)
```


pkg/drivers/kic/oci/network.go
```diff
@@ -82,6 +82,17 @@ func RoutableHostIPFromInside(ociBin string, clusterName string, containerName s
 		return containerGatewayIP(ociBin, containerName)
 	}
 
+	// on mac os and podman, try lima host record
+	if ociBin == "podman" && runtime.GOOS == "darwin" {
```
medyagh, line 86: Thank you for this PR, a couple questions

1- why this is only for pod man? Could this approach be used it be Docker on Lima  on macO too?

2- it seems to me this gonna use host.ima.internal for all Podmans on MacOS 
(Including if the user is gonna use Podman-Machine https://github.com/afbjorklund/podman-machine

Could we instead write a detection logic, to check if we are running inside Lima VM ? We have a package called “Detect” that has all kind of logic to detect where were are running (maybe we need a new detect for running inside Lima

https://github.com/kubernetes/minikube/blob/master/pkg/minikube/detect/detect.go


CC: Adding to comment here.
@afbjorklund

medyagh (changes requested):

pkg/drivers/kic/oci/network.go
rastaman: Hi @medyagh thanks for your time reviewing this,

1- it would be for any lima created VM i think yes. I am not sure on how to detect well at the moment, i will look in the lima code what would be the most sensible way to detect we are running in a lima vm.
2- i don't think  that podman use lima as a vm builder on windows. In fact podman doesn't use lima on macos too, it is only if you build you lima vm for podman and configure podman that lima will be used (else podman-machine use 
straight qemu with gsock wivor tap for network - i prefer to use lima because it support well socket_vmnet which allows to create brdiged network interfaces in the VM)
3 - limavm is not used by default by podman. i agree that i should add some detection logic in the file you mention so whatever the case if are using a limavm we can use the domain name to get the gateway ip, i will look for this.

thanks a lot for your comments i look forward to improve the code in the way described here, have a good day!

medyagh: yes please if this is not something that default users use, we need to only apply it for specific environment that is being used

rastaman:
I updated the code to:
- use a function in `detect`
- rely on an environment variable to define if we are using minikube inside a LimaVM managed VM, as there is no proper heuristic to know it

rastaman:
That is ok, i agree it is a specific use case, if someone need it they can still apply the attached patch.

PR #15288, opened 2022-11-04
Pin cloud-spanner-emulator/emulator image

spowelljr:
The image `cloud-spanner-emulator/emulator` for the `cloud-spanner` addon was not pinned. This is a compatibility and security issue.

They could introduce a v2 with breaking changes and users would automatically be upgraded, potentially causing the addon to break.

If a bad actor got authenticated to the repo they could push a compromised image.

Pinning the version and the hash prevents both of these issues.


medyagh (approved):
good call

PR #15293, opened 2022-11-04
CI: Fix setting multiline output

spowelljr:
The `git status --porcelain` commands sometimes output multiple lines if multiple lines were modified. However the new way of setting output doesn't work with multilines and fails with:
```
Error: Unable to process file command 'output' successfully.
Error: Invalid format ' M .github/workflows/docs.yml'
```

Updated the way to push data to support multiline as per https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings

### DIFF
--- a/.github/workflows/update-golint-version.yml
+++ b/.github/workflows/update-golint-version.yml
@@ -22,7 +22,9 @@ jobs:
         id: bumpGolint
         run: |
           make update-golint-version
-          echo "changes=$(git status --porcelain)" >> $GITHUB_OUTPUT
+          echo "changes<<EOF" >> $GITHUB_OUTPUT
+          echo "$(git status --porcelain)" >> $GITHUB_OUTPUT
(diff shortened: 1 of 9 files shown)


.github/workflows/update-golint-version.yml
```diff
@@ -22,7 +22,9 @@ jobs:
         id: bumpGolint
         run: |
           make update-golint-version
-          echo "changes=$(git status --porcelain)" >> $GITHUB_OUTPUT
+          echo "changes<<EOF" >> $GITHUB_OUTPUT
+          echo "$(git status --porcelain)" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
```
medyagh, line 27: I wonder if we can add a comment here for readabilty, why we are doing this ? (you could link to an issue or something

medyagh:
other than that, it looks good

PR #15356, opened 2022-11-14
Handle latest hyperkit version string

petiepooo:
Several years ago, hyperkit switched from using a data-based version to using a shortened build hash.  The version currently used by Canonical Multipass, for example, displays as such:

hyperkit: 40cbd5

Homepage: https://github.com/docker/hyperkit
License: BSD

This change adds a test that attempts to parse the new version.

### DIFF
--- a/pkg/minikube/registry/drvs/hyperkit/hyperkit.go
+++ b/pkg/minikube/registry/drvs/hyperkit/hyperkit.go
@@ -121,8 +121,9 @@ func isNewerVersion(currentVersion string, specificVersion string) (bool, error)
 	// Convert hyperkit version to time.Date to compare whether hyperkit version is not old.
 	layout := "20060102"
 	currentVersionDate, err := time.Parse(layout, currentVersion)
+	// newer versions of hyperkit no longer have a data embedded in their version
 	if err != nil {
-		return false, errors.Wrap(err, "parse date")
+		return true, nul
 	}
 	specificVersionDate, err := time.Parse(layout, specificVersion)
 	if err != nil {
--- a/pkg/minikube/registry/drvs/hyperkit/hyperkit_test.go
+++ b/pkg/minikube/registry/drvs/hyperkit/hyperkit_test.go
@@ -25,6 +25,7 @@ import (
 var (
 	versionOutputType1 = "hyperkit: v0.20190201-11-gc0dd46\n\nHomepage: https://github.com/docker/hyperkit\nLicense: BSD\n\n"
 	versionOutputType2 = "hyperkit: 0.20190201\n\nHomepage: https://github.com/docker/hyperkit\nLicense: BSD\n\n"
+	versionOutputType3 = "hyperkit: 40cbd5\n\nHomepage: https://github.com/docker/hyperkit\nLicense: BSD"
 )
 
 func TestSplitHyperKitVersion(t *testing.T) {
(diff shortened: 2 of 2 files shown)


medyagh (commented):
this seems to be only changing the unit tests, is this expected? but I always happy to see increase our test cases and cover edge cases.

petiepooo:
You are right, we will need more than just test cases.

Looking at: https://github.com/kubernetes/minikube/blob/84d6bc8ad24b59466d44b50af8640e4c048f1117/pkg/minikube/registry/drvs/hyperkit/hyperkit.go#L120 isNewerVersion() assumes a valid date is part of hyperkit's version string.  If a date is not detected, it is assumed to be the original semver format used when hyperkit was first forked from bhyve, and assigned a date of 0 (!!) which is less than the minimum version of 20190201, causing minikube to abort.

That assumption is now incorrect, as version strings for hyperkit starting with 20210107 use the first few characters from the build instead of a date, which for that release is 40cbd5.

Canonical's [Multipass](https://github.com/canonical/multipass) (for running Ubuntu VMs on MacOS via hyperkit) currently uses version 40cbd5.  They have never released a version with hyperkit older than 20200224.

Brew, the popular build system for MacOS, has now updated their formula for hyperkit from 20200908 to [20210107](https://github.com/Homebrew/homebrew-core/pull/113251), aka version 40cbd5, which minikube is not able to use.  They have had a hyperkit formula that works with minikube since [20190618](https://github.com/Homebrew/homebrew-core/pull/41057)

IMHO, the whole version check method here smells.  There was never any contract or promise that hyperkit would keep using date-based version strings, and indeed they no longer do.

The minimum version required by minikube was released on 20190201, almost 4 years ago.  I highly doubt you would find a single person who has a pre-2019 version of hyperkit already on their system, is just now downloading minikube, and is trying to get minikube running using their ancient hyperkit executable.

I believe most MacOS users running minikube are using a driver other than hyperkit.  Users that are using hyperkit have either not updated to the latest version via brew yet, or like me, they hacked the newer hyperkit binary so it fakes a date-based version string just for minikube (which works, FWIW).

Here is your chance to simplify the code base a small amount, @medyagh: let us just remove the check for a minimum hyperkit version.  Instead of assuming that the inability to parse a date out of the version string means hyperkit is new enough (vs too old), we can just assume that if hyperkit is present, it is new enough.  The version check is not helping anymore, and is causing more trouble than it is preventing.

If test coverage is a concern, consider how many tests it takes to cover 100% of nothing.

Some pertinent quotes from a few of my tech heroes:

The cheapest, fastest, and most reliable components are those that aren’t there.
        — Gordon Bell

One of my most productive days was throwing away 1000 lines of code.
        — Ken Thompson

Deleted code is debugged code.
        — Jeff Sickel

Controlling complexity is the essence of computer programming.
        — Brian Kernighan

petiepooo:
FWIW, I do not like this patch.  I would much rather the hyperkit version checks be removed entirely, but PR #14859 went nowhere...

petiepooo:
/remove-lifecycle stale

all this needs is some eyeballs...

petiepooo:
/remove-lifecycle rotten

On reviewing the testcases, I don't believe there's any change required there.  The TestIsNewerVersion() function test cases only validate correct parsing and ordering of valid dates from isNewerVersion().  I think this PR can be accepted as is with no additional test cases.

Again, my preference would be to actually remove the version checks entirely, as the hyperkit versions that would not work are beyond ancient now.  The fact that people aren't up in arms about this being broken indicates to me that nobody uses hyperkit anymore, or at least have easy alternatives once they see it isn't working.

PR #15393, opened 2022-11-22
CI: Add workflow to update gh version

spowelljr:


medyagh (approved):
good one ! thanks for adding to our automations and increasing our developer velocity

PR #15394, opened 2022-11-23
Fix subnet checking

spowelljr:
**Problem:**
While checking for a free subnet, if an IPv6 network is found `n.IP` is set to `0.0.0.0`, which is then set to `startSubnet`. This will cause an error on the next loop of the check as it add 9 to `0.0.0.0` instead of the starting IP, resulting in the second IP to check for being `0.9.0.0`.

**Expected:**
```
Checking 192.168.49.0
Checking 192.177.49.0
Checking 192.186.49.0
...
```

**Actual:**
```
Checking 192.168.49.0
Checking 0.9.0.0
Checking 0.18.0.0
...
```

**Solution:**
Assign the value returned from `n.IP` to its own variable to not interrupt the checking loop.

### DIFF
--- a/pkg/network/network.go
+++ b/pkg/network/network.go
@@ -230,14 +230,14 @@ func FreeSubnet(startSubnet string, step, tries int) (*Parameters, error) {
 		if err != nil {
 			return nil, err
 		}
-		startSubnet = n.IP
-		if isSubnetPrivate(startSubnet) {
-			taken, err := isSubnetTaken(startSubnet)
+		subnet := n.IP
+		if isSubnetPrivate(subnet) {
+			taken, err := isSubnetTaken(subnet)
 			if err != nil {
 				return nil, err
 			}
 			if !taken {
-				if ok := reserveSubnet(startSubnet, defaultReservationPeriod); ok {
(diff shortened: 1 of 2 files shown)


pkg/network/network.go
```diff
@@ -230,14 +230,14 @@ func FreeSubnet(startSubnet string, step, tries int) (*Parameters, error) {
 		if err != nil {
 			return nil, err
 		}
-		startSubnet = n.IP
-		if isSubnetPrivate(startSubnet) {
-			taken, err := isSubnetTaken(startSubnet)
+		subnet := n.IP
+		if isSubnetPrivate(subnet) {
```
medyagh, line 234: could we add unit test for this ?

medyagh (changes requested):

medyagh:
adding @josedonizetti as reviwer

pkg/network/network.go
spowelljr: Unit tests added

medyagh (approved):

medyagh (approved):
lgtm

PR #15403, opened 2022-11-25
gcp-auth addon now adds the image registry pull secret to newly created namespaces

### LINKED ISSUE #13351: gcp-auth addon does not create image pull secret for new namespace
### What Happened?

I enabled gcp-auth addon, and I realized that this addon does not detect namespace creation.
I usually use skaffold with helm, and it automatically creates a namespace for me.
However, with this addon, I have to run  `minikube addons enable gcp-auth --refresh` every time helm create new namespace, and It is not good workaround for my situation.

Here is the step to reproduce.

1. run `minikube addons enable gcp-auth` to install addon
2. run `kubectl create ns new-namespace`
3. run `kubectl get secret -n new-namespace`

no gcp-auth secret found on that namespace.

This problem may solved by adding webhook for namespace creation.
When new namespace created, the webhook creates a imagePullSecret for the namespace.
https://github.com/GoogleContainerTools/gcp-auth-webhook

However the code of creating imagePullSecret is in this repo.
https://github.com/kubernetes/minikube/blob/master/pkg/addons/addons_gcpauth.go#L140

I'm a bit confused where to report the issue, so let me know if it is a wrong place.

Thanks.

### Attach the log file

no logs for minikube.

### Operating System

macOS (Default)

### Driver

Docker

iastewar:
fixes #13351

This should be merged after the [webhook PR](https://github.com/GoogleContainerTools/gcp-auth-webhook/pull/32) has been merged. Please see its description for more details.


spowelljr (changes requested):
Could you run `make generate-docs` to update the translations, thanks

pkg/addons/addons_gcpauth.go
```diff
@@ -88,9 +85,9 @@ func enableAddonGCPAuth(cfg *config.ClusterConfig) error {
 		}
 	}
 
-	// Create a registry secret in every namespace we can find
-	// Always create the pull secret, no matter where we are
-	if err := createPullSecret(cfg, creds); err != nil {
+	// Patch service accounts for all namespaces to include the image pull secret.
+	// The image registry pull secret is added to the namespaces in the webhook.
+	if err := patchServiceAccounts(cfg); err != nil {
 		return errors.Wrap(err, "pull secret")
```
spowelljr, line 91: Probably want to update the error since function changed

spowelljr (changes requested):

spowelljr:
Would you mind adding a small test to our GCP-Auth tests that creates a new namespace and ensures the secret is there?

iastewar:
> Would you mind adding a small test to our GCP-Auth tests that creates a new namespace and ensures the secret is there?

Done

spowelljr:
The test you added is failing, I think you might need to add a small timeout to allow the webhook a chance run and inject the secret.

test/integration/addons_test.go
```diff
@@ -598,10 +598,25 @@ func validateCSIDriverAndSnapshots(ctx context.Context, t *testing.T, profile st
...
+	if err != nil {
+		t.Errorf("%s failed: %v", rr.Command(), err)
+	}
+}
+
 // validateGCPAuthAddon tests the GCP Auth addon with either phony or real credentials and makes sure the files are mounted into pods correctly
 func validateGCPAuthAddon(ctx context.Context, t *testing.T, profile string) {
 	defer PostMortemLogs(t, profile)
 
+	validateGCPAuthNamespaces(ctx, t, profile)
```
spowelljr, line 618: This func should be wrapped in:
```
t.Run("AuthNamespace", func(t *testing.T) {
   validateGCPAuthNamespaces(ctx, t, profile)
})
```
As currently the failure will be reported in the parent test so the `t.Fatalf` will kill the entire GCP-Auth test instead of just the namespace test itself.

spowelljr (changes requested):

iastewar:
> The test you added is failing, I think you might need to add a small timeout to allow the webhook a chance run and inject the secret.

Weird it worked on my machine. I'll try adding the timeout

test/integration/addons_test.go
iastewar: hmm I just figured I'd pull this code into a separate function since the gcpAuth test is already quite long. I could create a subtest as you suggest or I could just inline it to be consistent. What do you think?

iastewar:
Seems like adding a timeout did not help unfortunately. Any other ideas why the test would be failing? Do you know if Jenkins is doing any image caching?

iastewar:
> Seems like adding a timeout did not help unfortunately. Any other ideas why the test would be failing? Do you know if Jenkins is doing any image caching?

Looking at the webhook container logs, it seems that the token cannot be fetched due to an authorization problem so I'll look into this further.

iastewar:
> > Seems like adding a timeout did not help unfortunately. Any other ideas why the test would be failing? Do you know if Jenkins is doing any image caching?
> 
> Looking at the webhook container logs, it seems that the token cannot be fetched due to an authorization problem so I'll look into this further.

So the problem was actually an obvious one: the credentials are invalid. They are pulled in from `test/integration/testdata/gcp-creds.json`. Previously, the code didn't do anything if the token could not be retrieved as can be seen [here](https://github.com/kubernetes/minikube/blob/master/pkg/addons/addons_gcpauth.go#L149). I changed this to return the error instead. Either way though this new test will fail since it is actually checking if the secret exists.

A few ideas:
1. Modify the webhook code to use a dummy token if the credentials are equal to the test credentials.
2. Use an env variable (or similar) to start the webhook in TEST mode. in this mode, token will be mocked.
3. Use real credentials.

Number 1 is the easiest and I am guessing we can't go with number 3. Thoughts?

spowelljr:
1 & 2 both seem fine, 3 likely not

test/integration/addons_test.go
```diff
@@ -66,6 +67,15 @@ func TestAddons(t *testing.T) {
 			t.Fatalf("Failed setting GOOGLE_CLOUD_PROJECT env var: %v", err)
 		}
 
+		// MOCK_GOOGLE_TOKEN forces the gcp-auth webhook to use a mock token instead of trying to get a valid one from the credentials.
+		err = os.Setenv("MOCK_GOOGLE_TOKEN", "true")
```
spowelljr, line 71: Can we use `t.Setenv` here instead? Then we don't have to do the cleanup ourselves.

spowelljr (changes requested):

test/integration/addons_test.go
iastewar: Done

spowelljr (approved):
Looks good, thanks for the PR

PR #15404, opened 2022-11-25
update the links for go documenation

zdxgs:
The links has changed, update to the new links
 https://godoc.org/——> https://pkg.go.dev


spowelljr (changes requested):
If you want to update this you need to make this change in `cmd/minikube/cmd/cache_list.go` and then run `make generate-docs`

medyagh (approved):

Issue #13029, opened 2021-11-26
MSI installer puts shortcuts on the per-user Start Menu, not the system-wide Start Menu


xenophonf:
**Steps to reproduce the issue:** 

1. Run the minikube installer on Windows.
2. Continue through the UAC elevation prompt.
3. Take all of the defaults to complete the installation.

![minikube-installer-wrong-start-menu](https://user-images.githubusercontent.com/7139195/143595243-3e87f8b6-eb7e-4186-b110-4003ff070623.png)

Note that the installation path is `C:\Program Files\Kubernetes\Minikube` (indicating that elevation worked correctly).

According to the NSIS documentation, `$SMPROGRAMS` defaults to the current user's start menu.

https://nsis-dev.github.io/Documentation/Variables/SMPROGRAMS.html

Both the `Install` and `Uninstall` sections of the installation program should start with a `SetShellVarContext all` command.

https://nsis-dev.github.io/Documentation/Reference/SetShellVarContext.html

E.g., as of commit 23c85d3 (pushed 2021-11-24), that would be here:

https://github.com/kubernetes/minikube/blob/23c85d3efc0ec01b7530a86ec0ca565b74358cf4/installers/windows/minikube.nsi#L124

And here:

https://github.com/kubernetes/minikube/blob/23c85d3efc0ec01b7530a86ec0ca565b74358cf4/installers/windows/minikube.nsi#L169

If your intent is to support per-user installs, the script that updates the executable search path will have to be modified accordingly.

**Full output of failed command if not `minikube start`:**
<details>
The minikube installer does not have `LogSet on`, so the attached screen capture linked above is all I have.
</details>

spowelljr:
Thanks for creating this issue @xenophonf. So just to double check I understand, we're installing minikube globally but we're only adding the shortcut for the current user and we should be adding the shortcut globally as well?

xenophonf:
@spowelljr You're understanding this correctly.  If a program is installed for all users on Windows, its shortcut should be created in the All Users start menu.

correialabbate:
hey guys, can I work on this issue?

correialabbate:
Created a PR for it: https://github.com/kubernetes/minikube/pull/15405

PR #15405, opened 2022-11-25
Update Windows installer script to create system-wide shortcut

correialabbate:
Related issue: https://github.com/kubernetes/minikube/issues/13029
Problem: when using the Windows installer, it only creates a shortcut in the Start Menu for the user installing minikube. Since the installation process is a system-wide process, it should also create a Start Menu shortcut for other users on the machine.

Implemented the solution suggested by [https://github.com/xenophonf](xenophonf) - who opened the issue.
Tested both install and uninstall locally using a Windows 10 - Version 21H2 machine, and got the new expected behaviour.

### DIFF
--- a/installers/windows/minikube.nsi
+++ b/installers/windows/minikube.nsi
@@ -121,6 +121,8 @@ Function .onInit
 FunctionEnd
 
 Section "Install"
+  SetShellVarContext all
+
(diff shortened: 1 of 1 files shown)


spowelljr (approved):
Tested it and confirm that it worked, thanks for the PR

PR #15453, opened 2022-12-02
checking if supports putting v before hugo version

spowelljr:


spowelljr:
Confirm that it works

PR #15454, opened 2022-12-02
Remove gui from repo

spowelljr:
The GUI can now be found in its own repo: https://github.com/kubernetes-sigs/minikube-gui


afbjorklund:
Seems to have lost all the git history, though. Something like "filter-branch" could have helped.

spowelljr:
Ah, will use that and push that changes up

PR #15457, opened 2022-12-02
add cache to all golang

medyagh:


spowelljr:
You're missing one in `.github/workflows/update-hugo-version.yml`, you might be on an older branch and don't have that file pulled in yet

medyagh:
> 

thanks, added it

spowelljr (approved):

Issue #14965, opened 2022-09-16
could not bootstrap kubernetes cluster with CNI flannel


STRRL:
### What Happened?

minikube version

```
minikube version
minikube version: v1.26.1
commit: 62e108c3dfdec8029a890ad6d8ef96b6461426dc
```

step to reproduce:

- execute `minikube start --cni flannel`
- every workload (including coredns) would stuck in `ContainerCreating` 
- execute `k describe pod -n kube-system coredns-x-x`

There would be some error message in the event

```
  Warning  FailedCreatePodSandBox  3m                kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = [failed to set up sandbox container "b17871cf37fec75d431480096c668f785d555f2b07d94ce1e699a9590d4f54e0" network for pod "coredns-6d4b75cb6d-78ktc": networkPlugin cni failed to set up pod "coredns-6d4b75cb6d-78ktc_kube-system" network: missing network name:, failed to clean up sandbox container "b17871cf37fec75d431480096c668f785d555f2b07d94ce1e699a9590d4f54e0" network for pod "coredns-6d4b75cb6d-78ktc": networkPlugin cni failed to teardown pod "coredns-6d4b75cb6d-78ktc_kube-system" network: missing network name]
  Normal   SandboxChanged          8s (x14 over 3m)  kubelet            Pod sandbox changed, it will be killed and re-created.
```

Manually bypass:

- after I append `"name": "loopback"` into the `/etc/cni/net.d/200-loopback.conf`, the pod could finally be created. 

### Attach the log file

[log.txt](https://github.com/kubernetes/minikube/files/9580674/log.txt)

### Operating System

macOS (Default)

### Driver

Docker

STRRL:
maybe related with https://github.com/containernetworking/plugins/issues/167

STRRL:
I found that here is a mismatch between cri-o loopback config file and CNI spec, I have no idea why it works well before.

I create a PR for upstream: https://github.com/cri-o/cri-o/pull/6273

PR #15463, opened 2022-12-05
improve how CRs and k8s work with CNI plugins and cgroup drivers

prezha:
\<update>

## initial goal
reduce flakiness of Docker_Linux tests

## initial analysis & conclusion
tests are actually mostly ok, we have some real issues

## redefined goal
find & fix the issues

## outcome
  - Docker_Linux tests: all pass/green
  - KVM_Linux tests: all pass/green
  - Docker_Cloud_Shell tests: all pass/green
  - Hyperkit_macOS tests: all pass/green
  - Docker_Linux_containerd tests: all but one pass (TestPreload)
  - KVM_Linux_containerd tests: all but two pass (TestPreload and TestStoppedBinaryUpgrade/Upgrade - fixed but not yet committed)

## a bit more details & context
i've spent some time trying to figure out why all the tests are consistently passing on my machine but then just failing when run on ci/jenkins

_spoiler_: as it turns out, it all mostly boils down to inconsistent cgroups across the "stack" and also how CNIs and CRs play (or not!) together

part of the conclusion from that "investigation" was completely unexpected and surprising (to me, at least) discovery that we have ci/jenkins agents with different configurations! here the difference is not just in ServerVersion but also in the CgroupDriver used inside the os (ie, how the agent machine was booted)  - examples:

Docker_Linux-26999:
```
I1215 02:07:39.609753   11500 info.go:266] docker info: {ID:TS6T:UINC:MIYS:RZPA:KS6T:4JQK:7JHN:D6RA:LDP2:MHAE:G32M:C5NQ Containers:0 ContainersRunning:0 ContainersPaused:0 ContainersStopped:0 Images:4 Driver:overlay2 DriverStatus:[[Backing Filesystem extfs] [Supports d_type true] [Native Overlay Diff true] [userxattr false]] SystemStatus:<nil> Plugins:{Volume:[local] Network:[bridge host ipvlan macvlan null overlay] Authorization:<nil> Log:[awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog]} MemoryLimit:true SwapLimit:true KernelMemory:true KernelMemoryTCP:true CPUCfsPeriod:true CPUCfsQuota:true CPUShares:true CPUSet:true PidsLimit:true IPv4Forwarding:true BridgeNfIptables:true BridgeNfIP6Tables:true Debug:false NFd:24 OomKillDisable:true NGoroutines:34 SystemTime:2022-12-15 02:07:39.535647944 +0000 UTC LoggingDriver:json-file CgroupDriver:cgroupfs NEventsListener:0 KernelVersion:5.15.0-1025-gcp OperatingSystem:Ubuntu 20.04.5 LTS OSType:linux Architecture:x86_64 IndexServerAddress:https://index.docker.io/v1/ RegistryConfig:{AllowNondistributableArtifactsCIDRs:[] AllowNondistributableArtifactsHostnames:[] InsecureRegistryCIDRs:[127.0.0.0/8] IndexConfigs:{DockerIo:{Name:docker.io Mirrors:[] Secure:true Official:true}} Mirrors:[]} NCPU:8 MemTotal:33660665856 GenericResources:<nil> DockerRootDir:/var/lib/docker HTTPProxy: HTTPSProxy: NoProxy: Name:ubuntu-20-agent-3 Labels:[] ExperimentalBuild:false ServerVersion:20.10.21 ClusterStore: ClusterAdvertise: Runtimes:{Runc:{Path:runc}} DefaultRuntime:runc Swarm:{NodeID: NodeAddr: LocalNodeState:inactive ControlAvailable:false Error: RemoteManagers:<nil>} LiveRestoreEnabled:false Isolation: InitBinary:docker-init ContainerdCommit:{ID:a05d175400b1145e5e6a735a6710579d181e7fb0 Expected:a05d175400b1145e5e6a735a6710579d181e7fb0} RuncCommit:{ID:v1.1.4-0-g5fd4c4d Expected:v1.1.4-0-g5fd4c4d} InitCommit:{ID:de40ad0 Expected:de40ad0} SecurityOptions:[name=apparmor name=seccomp,profile=default] ProductLicense: Warnings:<nil> ServerErrors:[] ClientInfo:{Debug:false Plugins:[map[Experimental:true Name:app Path:/usr/libexec/docker/cli-plugins/docker-app SchemaVersion:0.1.0 ShortDescription:Docker App Vendor:Docker Inc. Version:v0.9.1-beta3] map[Name:buildx Path:/usr/libexec/docker/cli-plugins/docker-buildx SchemaVersion:0.1.0 ShortDescription:Docker Buildx Vendor:Docker Inc. Version:v0.9.1-docker] map[Name:compose Path:/usr/libexec/docker/cli-plugins/docker-compose SchemaVersion:0.1.0 ShortDescription:Docker Compose Vendor:Docker Inc. Version:v2.12.2] map[Name:scan Path:/usr/libexec/docker/cli-plugins/docker-scan SchemaVersion:0.1.0 ShortDescription:Docker Scan Vendor:Docker Inc. Version:v0.21.0]] Warnings:<nil>}}
```

Docker_Linux-27065:
```
I1219 20:54:45.962607   10285 info.go:266] docker info: {ID:TS6T:UINC:MIYS:RZPA:KS6T:4JQK:7JHN:D6RA:LDP2:MHAE:G32M:C5NQ Containers:0 ContainersRunning:0 ContainersPaused:0 ContainersStopped:0 Images:4 Driver:overlay2 DriverStatus:[[Backing Filesystem extfs] [Supports d_type true] [Native Overlay Diff true] [userxattr false]] SystemStatus:<nil> Plugins:{Volume:[local] Network:[bridge host ipvlan macvlan null overlay] Authorization:<nil> Log:[awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog]} MemoryLimit:true SwapLimit:true KernelMemory:true KernelMemoryTCP:true CPUCfsPeriod:true CPUCfsQuota:true CPUShares:true CPUSet:true PidsLimit:true IPv4Forwarding:true BridgeNfIptables:true BridgeNfIP6Tables:true Debug:false NFd:24 OomKillDisable:true NGoroutines:34 SystemTime:2022-12-19 20:54:45.130424738 +0000 UTC LoggingDriver:json-file CgroupDriver:systemd NEventsListener:0 KernelVersion:5.15.0-1025-gcp OperatingSystem:Ubuntu 20.04.5 LTS OSType:linux Architecture:x86_64 IndexServerAddress:https://index.docker.io/v1/ RegistryConfig:{AllowNondistributableArtifactsCIDRs:[] AllowNondistributableArtifactsHostnames:[] InsecureRegistryCIDRs:[127.0.0.0/8] IndexConfigs:{DockerIo:{Name:docker.io Mirrors:[] Secure:true Official:true}} Mirrors:[]} NCPU:8 MemTotal:33660669952 GenericResources:<nil> DockerRootDir:/var/lib/docker HTTPProxy: HTTPSProxy: NoProxy: Name:ubuntu-20-agent-14 Labels:[] ExperimentalBuild:false ServerVersion:20.10.22 ClusterStore: ClusterAdvertise: Runtimes:{Runc:{Path:runc}} DefaultRuntime:runc Swarm:{NodeID: NodeAddr: LocalNodeState:inactive ControlAvailable:false Error: RemoteManagers:<nil>} LiveRestoreEnabled:false Isolation: InitBinary:docker-init ContainerdCommit:{ID:78f51771157abb6c9ed224c22013cdf09962315d Expected:78f51771157abb6c9ed224c22013cdf09962315d} RuncCommit:{ID:v1.1.4-0-g5fd4c4d Expected:v1.1.4-0-g5fd4c4d} InitCommit:{ID:de40ad0 Expected:de40ad0} SecurityOptions:[name=apparmor name=seccomp,profile=default] ProductLicense: Warnings:<nil> ServerErrors:[] ClientInfo:{Debug:false Plugins:[map[Experimental:true Name:app Path:/usr/libexec/docker/cli-plugins/docker-app SchemaVersion:0.1.0 ShortDescription:Docker App Vendor:Docker Inc. Version:v0.9.1-beta3] map[Name:buildx Path:/usr/libexec/docker/cli-plugins/docker-buildx SchemaVersion:0.1.0 ShortDescription:Docker Buildx Vendor:Docker Inc. Version:v0.9.1-docker] map[Name:compose Path:/usr/libexec/docker/cli-plugins/docker-compose SchemaVersion:0.1.0 ShortDescription:Docker Compose Vendor:Docker Inc. Version:v2.14.1] map[Name:scan Path:/usr/libexec/docker/cli-plugins/docker-scan SchemaVersion:0.1.0 ShortDescription:Docker Scan Vendor:Docker Inc. Version:v0.23.0]] Warnings:<nil>}}
```

while i couldn't change/affect the ServerVersion, i went on and made minikube auto-adaptable to the underlying cgroup(s) driver, and that helped eliminate some of the "flakiness" but also hopefully made minikube more "robust" in terms of the "flavours" of os/settings our users have
sounds like good timing as well, since k8s now [recommends](https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cgroup-drivers) using systemd as the default cgroup driver, and with this, we shouldn't have much to change additionally going forward

## code fixes and improvements (candidates for individual PRs - next steps)
  - cgroup(s) driver: auto-detect on linux (defaulting to cgroupfs/v1 for non-linux) and auto-configure all components accordingly
  - CNIs and CRs: refactored (again!) how those work together and also removed the need to use/have custom `/etc/cni/net.mk` dir for CRs and CNIs (note: that dir should be removed from our "distro", as it confuses containerd thinking it's default one to look into; until then, there's a "hack" that removes it)
  - container runtimes: ensured containerd is also properly configured when docker is selected as container runtime (and bound to it)
  - network - cni/localhost: automatically fix the missing name and update to the now-expected version (v1.0.0)
  - network - cni/bridge: automatically mask all default bridges when cni is used and auto-fix subnet to use DefaultPodCIDR and appropriate gateway (instead of default 10.8x.0.0) otherwise - when default bridge should be used
  - network: improved detection and reservation of free subnet, so it works across processes (and not just locally) - reduced possibility of collisions and thus the time needed to pick actually free subnet (also avoid giving up after 5 failed attempts)
  - network - cni/flannel: updated to the latest v1.1.2/v0.20.2 and moved manifest out from "code" to _embedding_ separate yaml manifest file
  - network - cni/calico: updated to the latest v3.24.5
  - general: prevent race condition over multiple minikube instances rewriting shared "ca-certs" using locks, that was causing some auth issues before
  - network: made hair-pin work (now configurable in kubelet config) - also for tests
  - general: made kubelet respect custom **_runtimeRequestTimeout_**, so eg, it doesn't CrashLoopBackOff on pulling larger images after "hardcoded" default of 5mins (now configurable in kubelet config; note: "runtime-request-timeout" it's not supported as kubelet param anymore, so i moved it to config)

## hacks (that should be replaced with proper "distribution" updates!)
used to eliminate some of the "known issues" with the upstreams:
  - cri-docker: use v0.2.6
  - containerd: use v1.6.14
  - contained: change systemd service to use KillMode=process (in line with all other container runtimes) so it doesn't restart containers when the service itself is restarted
  - remove '/etc/cni/net.mk' dir

## a number of other minor tweaks, additions and fixes
eg, timeouts, juju packages updates, comments, docs, spellings, etc.

## tests fixes and improvements
  - TestStartStop: this one was a pain to discover - we used 192.168.111.111/16 as kubeadm.pod-network-cidr that overlapped with all other "standard" network subnets we use; that is fixed now
  - fixed TestJSONOutput when unexpected but still valid cases of out-of-order events occur (due to goroutines finishing at different times)
  - updated "netcat" deployment to use the latest dnsutils/agnhost image v2.40
  - GCPAuth addon test skipped until 'Permission \"artifactregistry.repositories.downloadArtifacts\" denied on resource \"projects/k8s-minikube/locations/us/repositories/test-artifacts\" (or it may not exist)' issue is resolved;
    => here, the issue actually might be in os.Setenv("MOCK_GOOGLE_TOKEN", "true") and we just need to address that - https://github.com/GoogleContainerTools/gcp-auth-webhook#gcp-auth-webhook:
 > Setting the environment variable MOCK_GOOGLE_TOKEN to true will prevent using the google application credentials to fetch the token used for the image pull secret. Instead the token will be mocked.
  - ensure tests have unique but still short "random" names (previously, it was based on the second they've started, which could lead to potential conflicts)
  - added a number of additional debug logs collected for TestNetworkPlugins (can be removed/suspended now)

## general notes
  - optimising the code flow resulted in an overall reduction of tests' duration but
  - in the current pr, in addition to the hacks above (that take additional time to download and replace binaries), there are also some additional "waits" i've implemented to force staging of phases and make debugging easier (so, eg, i've implemented additional "Ready" method for CNIs to wait for and only continue once they are ready, and also WaitForPods is waiting for them to become Ready not just Running) - that's the reason why we see "longer" times for 'minikube start' and 'enable ingress'; those extra "waits" should not be necessary in "production" after other fixes i've made, while we can improve how we check/wait for different resources health (but that's another topic)

\</update>

---

the goal of this pr is to see if we can reduce the rate of errors and flakes with the TestNetworkPlugins test group with the linux+docker combo

locally, these tests all pass ([TestNetworkPlugins-linux+docker.log](https://github.com/kubernetes/minikube/files/10150317/TestNetworkPlugins-linux%2Bdocker.log)), and if ci_cd tests show similar, we might want to breakdown these into separate PRs

key points:
- starting from scratch (ie, no images cached lcoally) - avoid `race condition` overwriting minikube's certs (should fix relevant flake with other tests as well)
  - `unauthorised` (as a consequence of the previous point) is a non-retryable error - fail fast w/o retrying
example:
```
...
I1204 12:52:01.424956 2504034 certs.go:54] Setting up /home/prezha/.minikube/profiles/custom-flannel-124934 for IP: 172.17.0.2
I1204 12:52:01.424999 2504034 certs.go:187] generating minikubeCA CA: /home/prezha/.minikube/ca.key
I1204 12:52:01.543493 2504034 crypto.go:156] Writing cert to /home/prezha/.minikube/ca.crt ...
...
I1204 12:52:01.435619 2504033 certs.go:54] Setting up /home/prezha/.minikube/profiles/bridge-124934 for IP: 172.17.0.5
I1204 12:52:01.435645 2504033 certs.go:187] generating minikubeCA CA: /home/prezha/.minikube/ca.key
I1204 12:52:01.647799 2504033 crypto.go:156] Writing cert to /home/prezha/.minikube/ca.crt ...
...
I1204 12:52:01.443825 2504037 certs.go:54] Setting up /home/prezha/.minikube/profiles/flannel-124934 for IP: 192.168.76.2
I1204 12:52:01.443859 2504037 certs.go:187] generating minikubeCA CA: /home/prezha/.minikube/ca.key
I1204 12:52:01.781017 2504037 crypto.go:156] Writing cert to /home/prezha/.minikube/ca.crt ...
...
I1204 12:52:01.444867 2504029 certs.go:54] Setting up /home/prezha/.minikube/profiles/auto-124934 for IP: 192.168.67.2
I1204 12:52:01.444891 2504029 certs.go:187] generating minikubeCA CA: /home/prezha/.minikube/ca.key
I1204 12:52:01.521760 2504029 crypto.go:156] Writing cert to /home/prezha/.minikube/ca.crt ...
...
```
just to fail afterwards:
```
...
W1204 12:54:30.228254 2504033 out.go:239] X Exiting due to GUEST_START: wait 10m0s for node: waiting for node to be ready: waitNodeCondition: error getting node "bridge-124934": Unauthorized
...
W1204 12:54:33.471702 2504037 out.go:239] X Exiting due to GUEST_START: wait 10m0s for node: waiting for node to be ready: waitNodeCondition: error getting node "flannel-124934": Unauthorized
...
W1204 12:54:33.496623 2504029 out.go:239] X Exiting due to GUEST_START: wait 10m0s for node: waiting for node to be ready: waitNodeCondition: error getting node "auto-124934": Unauthorized
...
W1204 12:54:34.868234 2504030 out.go:239] X Exiting due to GUEST_START: wait 10m0s for node: waiting for node to be ready: waitNodeCondition: error getting node "kindnet-124934": Unauthorized
...
```
- "mask" `cri-o bridge` for all CNIs as it interferes with others
- give `loopback` a name (as reported by @STRRL in https://github.com/kubernetes/minikube/issues/14965 with the [pr](https://github.com/cri-o/cri-o/pull/6273) still pending in cri-o) - example of errors resolved:
```
Failed to create pod sandbox: rpc error: code = Unknown desc = [failed to set up sandbox container "..." network for pod "...": networkPlugin cni failed to set up pod "..." network: missing network name:,
```
and
```
failed to clean up sandbox container "..." network for pod "...": networkPlugin cni failed to teardown pod "..." network: missing network name]
```
- `kubenet` also needs the cni (eg, bridge) to support hairpin mode
- updated `calico` to the latest version
- updated `flannel` to the latest version and extracted k8s manifests from code to a separate file that's then embedded
  - solved: "iptables v1.8.3 (legacy): Couldn't load target `CNI-x': No such file or directory" => enabled TestNetworkPlugins for docker + flannel
- increased download `timeout for k8s binaries`
- increased tests' `memory` and `wait timeout` to avoid weird issues with constraint resources
- updated `juju` packages to the latest version - ours are ~3 years old


medyagh:
> /ok-to-test

I still see the failures... on Docker_linux

prezha:
> > /ok-to-test
> 
> I still see the failures... on Docker_linux

yep, i've created a draft pr to test initial fixes, some other issues surfaced - should be further improved with the commit i just made, we'll see

medyagh:
@prezha this PR touches the ISO but we haven't build the ISO on this PR, do you still wanna do this as part of this PR ?

if this test is green without building the ISO, does that mean we do NOT need the ISO/kicbase changes? can we put them in a different PR ? or do you suggest this be included in this PR and why ?
if yes lets invoke the commands to build iso here

deploy/iso/minikube-iso/arch/x86_64/package/containerd-bin/containerd.service
```diff
@@ -16,7 +16,7 @@ ExecStart=/usr/bin/containerd \
       --root ${PERSISTENT_DIR}/var/lib/containerd
 TasksMax=8192
 Delegate=yes
-KillMode=mixed
+KillMode=process
```
medyagh, line 19: why this ?

prezha:
> @prezha this PR touches the ISO but we haven't build the ISO on this PR, do you still wanna do this as part of this PR ?
> 
> if this test is green without building the ISO, does that mean we do NOT need the ISO/kicbase changes? can we put them in a different PR ? or do you suggest this be included in this PR and why ? if yes lets invoke the commands to build iso here

@medyagh i think we'll need to update iso/kicbase, and although there are dependencies between those and other changes, it would probably be better to do it in a separate pr - changes would include, eg:
- cri-dockerd v0.3.0 (so we support k8s v1.26.0+ using api v1)
- contained v1.6.14 (so that k8s/kubelet timeout is actually respected and works)
- removal of automated creation of '/etc/cni/net.mk' (not needed anymore after we merge these, and it confuses containerd)
- some other containerd default configs (mostly for forward compatibility and removal of deprecated/removed features & params)
- avoidance of "hardcoded" mixing of different cgroup managers (eg, we boot with cgroupfs but we then set systemd as "default", whereas only some config params are "dynamically" changeable during runtime)
etc.

prezha:
> @prezha Once you've got this in a state you're happy with we should hop on a call and you can go over the changes with me/us, the PR is very large and having you walk us through it would help us with the review process.

hey @spowelljr, sure, that sounds good to me - how about Monday around this time (ie, 11 am PST), would that work for you?

medyagh:
> > @prezha this PR touches the ISO but we haven't build the ISO on this PR, do you still wanna do this as part of this PR ?
> > if this test is green without building the ISO, does that mean we do NOT need the ISO/kicbase changes? can we put them in a different PR ? or do you suggest this be included in this PR and why ? if yes lets invoke the commands to build iso here
> 
> @medyagh i think we'll need to update iso/kicbase, and although there are dependencies between those and other changes, it would probably be better to do it in a separate pr - changes would include, eg:
> 
> * cri-dockerd v0.3.0 (so we support k8s v1.26.0+ using api v1)
> * contained v1.6.14 (so that k8s/kubelet timeout is actually respected and works)
> * removal of automated creation of '/etc/cni/net.mk' (not needed anymore after we merge these, and it confuses containerd)
> * some other containerd default configs (mostly for forward compatibility and removal of deprecated/removed features & params)
> * avoidance of "hardcoded" mixing of different cgroup managers (eg, we boot with cgroupfs but we then set systemd as "default", whereas only some config params are "dynamically" changeable during runtime)
>   etc.

for sake of a smaller PR lets do those changes in a follow UP pr

prezha:
> > > @prezha this PR touches the ISO but we haven't build the ISO on this PR, do you still wanna do this as part of this PR ?
> > > if this test is green without building the ISO, does that mean we do NOT need the ISO/kicbase changes? can we put them in a different PR ? or do you suggest this be included in this PR and why ? if yes lets invoke the commands to build iso here
> > 
> > 
> > @medyagh i think we'll need to update iso/kicbase, and although there are dependencies between those and other changes, it would probably be better to do it in a separate pr - changes would include, eg:
> > 
> > * cri-dockerd v0.3.0 (so we support k8s v1.26.0+ using api v1)
> > * contained v1.6.14 (so that k8s/kubelet timeout is actually respected and works)
> > * removal of automated creation of '/etc/cni/net.mk' (not needed anymore after we merge these, and it confuses containerd)
> > * some other containerd default configs (mostly for forward compatibility and removal of deprecated/removed features & params)
> > * avoidance of "hardcoded" mixing of different cgroup managers (eg, we boot with cgroupfs but we then set systemd as "default", whereas only some config params are "dynamically" changeable during runtime)
> >   etc.
> 
> for sake of a smaller PR lets do those changes in a follow UP pr

@medyagh yes, i agree, will extract those into a separate pr

pkg/minikube/cni/cni.go
```diff
@@ -200,41 +205,105 @@ func applyManifest(cc config.ClusterConfig, r Runner, f assets.CopyableFile) err
...
+	if len(configs) == 0 {
+		klog.Infof("no bridge cni configs found in %q - nothing to disable", path)
+		return nil
+	}
+	klog.Infof("disabled [%s] bridge cni config(s)", configs)
+
+	return nil
+}
+
+func configureAllBridgeCNIs(r Runner, cidr string) error {
```
medyagh, line 263: please add a comment for this func

prezha: a detailed comment is with the calling func ConfigureDefaultBridgeCNIs() just above this one, but sure - i can move those comments down to be with this func

test/integration/net_test.go
```diff
@@ -268,3 +273,361 @@ func verifyKubeletFlagsOutput(t *testing.T, k8sVersion, kubeletPlugin, out strin
 		t.Errorf("expected --network-plugin=%s, got %s", kubeletPlugin, out)
 	}
 }
+
+// debug logs for dns and other network issues
+func debugLogs(t *testing.T, profile string) {
```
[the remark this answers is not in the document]
prezha: sure, will add that

test/integration/start_stop_delete_test.go
```diff
@@ -61,7 +61,7 @@ func TestStartStop(t *testing.T) {
 				"--feature-gates",
 				"ServerSideApply=true",
 				"--network-plugin=cni",
-				"--extra-config=kubeadm.pod-network-cidr=192.168.111.111/16",
+				"--extra-config=kubeadm.pod-network-cidr=10.42.0.0/16",
```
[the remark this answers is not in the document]
prezha: yes, it's inside minikube, and this is to avoid overlapping network segments:
```
$ minikube ssh -- ip a s
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
2: docker0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default
    link/ether 02:42:1c:74:0b:31 brd ff:ff:ff:ff:ff:ff
    inet 172.17.0.1/16 brd 172.17.255.255 scope global docker0
       valid_lft forever preferred_lft forever
10: eth0@if11: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default
    link/ether 02:42:c0:a8:31:02 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 192.168.49.2/24 brd 192.168.49.255 scope global eth0
       valid_lft forever preferred_lft forever
```
in the example above we would have `192.168.111.111/16` and `192.168.49.2/24`

test/integration/net_test.go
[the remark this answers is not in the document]
prezha: i've added the start/end markers around the debug logs

pkg/minikube/cni/cni.go
prezha: i've moved and added some more comments around the three related functions

deploy/iso/minikube-iso/arch/aarch64/package/containerd-bin-aarch64/config.toml
```diff
@@ -40,8 +40,7 @@ oom_score = 0
     [plugins."io.containerd.grpc.v1.cri".containerd]
       discard_unpacked_layers = true
       snapshotter = "overlayfs"
-      [plugins."io.containerd.grpc.v1.cri".containerd.default_runtime]
-        runtime_type = "io.containerd.runc.v2"
+      default_runtime_name = "runc"
```
spowelljr, line 43: Should this also be removed?

https://github.com/kubernetes/minikube/blob/4f3d0da802c7514e57551f42ba6fb9f0abc3d4b0/deploy/iso/minikube-iso/arch/aarch64/package/containerd-bin-aarch64/config.toml.default#L77

deploy/iso/minikube-iso/arch/x86_64/package/containerd-bin/config.toml
```diff
@@ -40,8 +40,7 @@ oom_score = 0
     [plugins."io.containerd.grpc.v1.cri".containerd]
       discard_unpacked_layers = true
       snapshotter = "overlayfs"
-      [plugins."io.containerd.grpc.v1.cri".containerd.default_runtime]
-        runtime_type = "io.containerd.runc.v2"
+      default_runtime_name = "runc"
```
spowelljr, line 43: Should this also be removed?

https://github.com/kubernetes/minikube/blob/4f3d0da802c7514e57551f42ba6fb9f0abc3d4b0/deploy/iso/minikube-iso/arch/x86_64/package/containerd-bin/config.toml.default#L77

pkg/kapi/kapi.go
```diff
@@ -223,25 +222,33 @@ func ScaleDeployment(kcontext, namespace, deploymentName string, replicas int) e
 	err = wait.PollImmediate(kconst.APICallRetryInterval, ReasonableMutateTime, func() (bool, error) {
 		scale, err := client.AppsV1().Deployments(namespace).GetScale(context.Background(), deploymentName, meta.GetOptions{})
 		if err != nil {
-			klog.Warningf("failed getting deployment scale, will retry: %v", err)
-			return false, nil
+			if IsRetryableAPIError(err) {
+				klog.Warningf("failed getting %q deployment scale, will retry: %v", deploymentName, err)
+				return false, nil
+			}
+			klog.Info("non-retryable failure while getting %q deployment scale: %v", deploymentName, err)
```
spowelljr, line 229:
```suggestion
			klog.Infof("non-retryable failure while getting %q deployment scale: %v", deploymentName, err)
```
Should this be a `Warningf` instead of an `Infof`?

pkg/kapi/kapi.go
```diff
@@ -223,25 +222,33 @@ func ScaleDeployment(kcontext, namespace, deploymentName string, replicas int) e
...
 		}
 		if scale.Spec.Replicas != int32(replicas) {
 			scale.Spec.Replicas = int32(replicas)
 			_, err = client.AppsV1().Deployments(namespace).UpdateScale(context.Background(), deploymentName, scale, meta.UpdateOptions{})
 			if err != nil {
-				klog.Warningf("failed rescaling deployment, will retry: %v", err)
+				if IsRetryableAPIError(err) {
+					klog.Warningf("failed rescaling %s deployment, will retry: %v", deploymentName, err)
+				} else {
+					klog.Info("non-retryable failure while rescaling %s deployment: %v", deploymentName, err)
```
spowelljr, line 239:
```suggestion
					klog.Infof("non-retryable failure while rescaling %s deployment: %v", deploymentName, err)
```
Should this be a `Warningf` instead of an `Infof`?

pkg/minikube/node/start.go
```diff
@@ -401,14 +414,30 @@ func configureRuntimes(runner cruntime.CommandRunner, cc config.ClusterConfig, k
...
+			klog.Warningf("cannot ensure containerd is configured properly and reloaded for docker - cluster might be unstable: %v", err)
 		}
 	}
 
-	inUserNamespace := strings.Contains(cc.KubernetesConfig.FeatureGates, "KubeletInUserNamespace=true")
-	err = cr.Enable(disableOthers, forceSystemd(), inUserNamespace)
+	disableOthers := true
+	if driver.BareMetal(cc.Driver) {
+		disableOthers = false
+	}
```
spowelljr, line 439:
```suggestion
	disableOthers := !driver.BareMetal(cc.Driver)
```
nit: I find it cleaner and easier to read "disableOthers if not bare metal"

pkg/network/network.go
```diff
@@ -271,39 +267,13 @@ func ParseAddr(addr string) (net.IP, *net.IPNet, error) {
...
-			// unexpired reservation: restore original createdAt value
-			reservedSubnets.Store(subnet, reservation{createdAt: createdAt})
-			klog.Infof("skipping subnet %s that has unexpired reservation: %+v", subnet, &reservedSubnets)
-			return false
-		}
-		// expired reservation: renew setting createdAt to now
-		reservedSubnets.Store(subnet, reservation{createdAt: time.Now()})
-		klog.Infof("reusing subnet %s that has expired reservation: %+v", subnet, &reservedSubnets)
-		return true
+// reserveSubnet returns releaser if subnet was successfully reserved for given period, creating lock for subnet to avoid race condition between multiple minikube instances (especially white testing in parallel).
```
spowelljr, line 270:
```suggestion
// reserveSubnet returns releaser if subnet was successfully reserved for given period, creating lock for subnet to avoid race condition between multiple minikube instances (especially while testing in parallel).
```

test/integration/main_test.go
```diff
@@ -154,6 +154,14 @@ func KicDriver() bool {
 	return DockerDriver() || PodmanDriver()
 }
 
+// VMDriver checks if the driver is a VM
+func VMDriver() bool {
+	if KicDriver() || NoneDriver() {
+		return false
+	}
+	return true
```
spowelljr, line 162:
```suggestion
	return !KicDriver() && !NoneDriver()
```
nit: can shorten

spowelljr (changes requested):

test/integration/main_test.go
prezha: nice!

pkg/network/network.go
prezha: eagle-eye :)

pkg/minikube/node/start.go
prezha: agree

pkg/kapi/kapi.go
prezha: good catch, thanks!

i think we could actually return the error here and effectively stop, otherwise, we'd continue retrying - something like:

> return false, fmt.Errorf("non-retryable failure while rescaling %s deployment: %v", deploymentName, err)

pkg/kapi/kapi.go
prezha: like below -

i think we could actually return the error here and effectively stop, otherwise, we'd continue retrying - something like:

> return false, fmt.Errorf("non-retryable failure while getting %q deployment scale: %v", deploymentName, err)

deploy/iso/minikube-iso/arch/x86_64/package/containerd-bin/config.toml
prezha: as agreed, we'll remove these _config.toml.default_ files (for both x86_64 and aarch64) as they're not used

btw, they can be generated with `containerd config default`

deploy/iso/minikube-iso/arch/aarch64/package/containerd-bin-aarch64/config.toml
prezha: similar to the comment below: we'll remove it as agreed

pkg/minikube/cruntime/containerd.go
```diff
@@ -194,13 +223,26 @@ func (r *Containerd) Enable(disOthers, forceSystemd, inUserNamespace bool) error
...
 	}
 
+	// TODO (@prezha): remove this hack after proper version update in minikube release
+	// ref: https://github.com/containerd/containerd/blob/main/RELEASES.md#kubernetes-support
+	targetVersion := "1.6.15"
+	currentVersion, err := r.Version()
+	if err == nil && semver.MustParse(targetVersion).GT(semver.MustParse(currentVersion)) {
+		klog.Infof("replacing original containerd with v%s-%s-%s", targetVersion, runtime.GOOS, runtime.GOARCH)
+		_ = r.Init.ForceStop("containerd")
+		if err := updateContainerdBinary(r.Runner, targetVersion, runtime.GOOS, runtime.GOARCH); err != nil {
```
prezha, line 241: instead of 'runtime.GOOS', it should actually always be "linux" os (still with an appropriate arch) as that's what we use, but we're going to replace this "hack" with the proper binary in iso/kicbase anyway, so it's not critical

prezha: fixed

pkg/minikube/bootstrapper/bsutil/kubeadm.go
```diff
@@ -68,6 +68,12 @@ func GenerateKubeadmYAML(cc config.ClusterConfig, n config.Node, r cruntime.Mana
 		}
 		return nil, errors.Wrap(err, "getting cgroup driver")
 	}
+	// TODO: investigate why containerd (v1.6.15) does not work with k8s (v1.25.3) when both are set to use systemd cgroup driver
```
spowelljr, line 71: Could you create an issue for this if not already done and link the issue here

spowelljr (changes requested):

pkg/minikube/bootstrapper/bsutil/kubeadm.go
prezha: sure - i've created https://github.com/kubernetes/minikube/issues/15633 with more details and logs there, and referenced it in the comment

prezha:
hmmm, we again had the same [issue](https://github.com/kubernetes/minikube/actions/runs/3907867809/jobs/6677508611) as seen earlier today - network issues?
```
...
E: Failed to fetch http://ports.ubuntu.com/ubuntu-ports/pool/main/i/icu/libicu66_66.1-2ubuntu2.1_arm64.deb  Connection failed [IP: 185.125.190.36 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Fetched 73.1 MB in 9min 51s (124 kB/s)
...
```

pkg/minikube/node/start.go
```diff
@@ -379,9 +381,27 @@ func configureRuntimes(runner cruntime.CommandRunner, cc config.ClusterConfig, k
...
 	}
 
-	disableOthers := true
-	if driver.BareMetal(cc.Driver) {
-		disableOthers = false
+	// 87-podman.conflist cni conf potentially conflicts with others and is created by podman on its first invocation,
+	// so we "provoke" it here to ensure it's generated and that we can disable it
+	// note: using 'help' or '--help' would be cheaper, but does not trigger that; 'version' seems to be next best option
+	if co.Type == constants.CRIO {
+		_, _ = runner.RunCmd(exec.Command("sudo", "sh", "-c", `podman version >/dev/null`))
```
spowelljr, line 388: Maybe not necessary, but could check for an error if somehow `podman` is not present?

pkg/drivers/kic/oci/network_create.go
```diff
@@ -145,6 +145,7 @@ func tryCreateDockerNetwork(ociBin string, subnet *network.Parameters, mtu int,
 
 	rr, err := runCmd(exec.Command(ociBin, args...))
 	if err != nil {
+		klog.Errorf("failed to create %s network %s %s with gateway %s and mtu of %d: %v", ociBin, name, subnet.CIDR, subnet.Gateway, mtu, err)
```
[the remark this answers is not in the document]
prezha: that's a good point, and on the other hand, i also think that with these changes, it should be very unlikely to happen (especially for the end users that usually do not start a number of instances concurrently - more than ~5), and if that's not the case, perhaps we would want it to surface and be reported so that we can address it?

pkg/minikube/cruntime/containerd.go
```diff
@@ -194,13 +230,26 @@ func (r *Containerd) Enable(disOthers, forceSystemd, inUserNamespace bool) error
...
 		return err
 	}
 
+	// TODO (@prezha): remove this hack after proper version update in minikube release
+	// ref: https://github.com/containerd/containerd/blob/main/RELEASES.md#kubernetes-support
+	targetVersion := "1.6.15"
+	currentVersion, err := r.Version()
+	if err == nil && semver.MustParse(targetVersion).GT(semver.MustParse(currentVersion)) {
+		klog.Infof("replacing original containerd with v%s-linux-%s", targetVersion, runtime.GOARCH)
+		_ = r.Init.ForceStop("containerd")
```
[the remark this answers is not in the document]
prezha: agree - even if it fails, we would still want to continue in this case, but it would be useful to know about an error - if we are to keep this temporary segment for some time, then i can surely add that

spowelljr (approved):
Thank you very much for your large undertaking to improving minikube as a whole, it is very much appreciated!

PR #15477, opened 2022-12-07
docs: add missing OutFile parameter

al-cheb:
https://minikube.sigs.k8s.io/docs/tutorials/setup_minikube_gui/ - the OutFile parameter is missing in the windows steps tab. It should be like in the installation section https://minikube.sigs.k8s.io/docs/start/:
```
-OutFile
Specifies the output file for which this cmdlet saves the response body. Enter a path and filename. If you omit the path, the default is the current location.
```

Before:
image

After:
image


al-cheb:
@spowelljr, review pls

Issue #15456, opened 2022-12-02
minikube should not warn you to NOT use virtualbox if it auto-selected it


medyagh:
### What Happened?

as seen in this github action run on Mac Os minikube choce vbox because it was only option

but it also warns the user not to use vbox

https://github.com/medyagh/local-dev-example-with-minikube/actions/runs/3605679013/jobs/6076306844

### Attach the log file

```
minikube v1.28.0 on Darwin 12.6.1
* Automatically selected the virtualbox driver. Other choices: ssh, qemu2 (experimental)
* Downloading VM boot image ...
* Starting control plane node minikube in cluster minikube
* Downloading Kubernetes v1.25.3 preload ...
* Creating virtualbox VM (CPUs=2, Memory=3500MB, Disk=[20](https://github.com/medyagh/local-dev-example-with-minikube/actions/runs/3605679013/jobs/6076306844#step:4:21)000MB) ...
* Preparing Kubernetes v1.25.3 on Docker 20.10.20 ...
  - Generating certificates and keys ...
  - Booting up control plane ...
  - Configuring RBAC rules ...
╭───────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                   │
│    You have selected "virtualbox" driver, but there are better options !                          │
│    For better performance and support consider using a different driver:                          │
│            - qemu2                                                                                │
│                                                                                                   │
│    To turn off this warning run:                                                                  │
│                                                                                                   │
│            $ minikube config set WantVirtualBoxDriverWarning false                                │
│                                                                                                   │
│                                                                                                   │
│    To learn more about on minikube drivers checkout https://minikube.sigs.k8s.io/docs/drivers/    │
│    To see benchmarks checkout https://minikube.sigs.k8s.io/docs/benchmarks/cpuusage/              │
│                                                                                                   │
╰───────────────────────────────────────────────────────────────────────────────────────────────────╯
* Verifying Kubernetes components...
  - Using image gcr.io/k8s-minikube/storage-provisioner:v5
* Enabled addons: storage-provisioner
* ```

### Operating System

None

### Driver

None

axrav:
@medyagh what would be the best way to test if the bug is fixed?, the one you are using is using your actions afaik, how can I test it properly?

Tzurrr:
is this issue still active?

PaulOchoa952:
still open?

nirs:
We have several problems:
[...]

PR #15485, opened 2022-12-08
Remove VirtualBox selection warning if it is autoselected

DiptoChakrabarty:
fixes issue #15456

Would require prow test as having issues starting with driver Vbox in local system


spowelljr:
@medyagh This code is going to need to have a bit of a bigger rework.

Looking at the `warnVirtualBox` func https://github.com/kubernetes/minikube/blob/691cc40705ffd187ed50b0518ed184c441667d88/pkg/minikube/node/start.go#L832

It's written based on the user selecting the driver directly, for example the wording of the message is `You have selected "virtualbox"` (easy fix).

But the logic to get the list of drivers checks if the driver is installed and healthy. VirtualBox already has a very low priority, so if the VirtualBox driver gets auto selected there's likely no other driver on the users machine to use, so the warning will likely be skipped for all cases.

Essentially, if this change is implemented, the warning will likely never be shown.

Should the intention be that we just list other drivers the user can install and use?

PR #15513, opened 2022-12-14
Add groups check to SSH driver

### LINKED ISSUE #15512: `sudo usermod` fails due to lack of interactive terminal
### What Happened?

The SSH driver fails to create a cluster using the Docker runtime. It attempts to add the SSH user to the docker group using sudo, but this fails immediately since the SSH session is not interactive, and no sudo password is passed using `-S`.

### Attach the log file

[log.txt](https://github.com/kubernetes/minikube/files/10230497/log.txt)

### Operating System

Other - Manjaro using kernel 5.15.81

### Driver

SSH

kaylendog:
Adds a groups check to avoid adding SSH users to the `docker` group if they're already in the `docker` group.

fixes #15512

### DIFF
--- a/pkg/drivers/ssh/ssh.go
+++ b/pkg/drivers/ssh/ssh.go
@@ -146,7 +146,7 @@ func (d *Driver) Create() error {
 		if err != nil {
 			return errors.Wrap(err, "groups")
 		}
-		if !strings.Contains(groups.Stderr.String(), "docker") {
+		if !strings.Contains(groups.Stdout.String(), "docker") {
 			if _, err := d.exec.RunCmd(exec.Command("sudo", "usermod", "-aG", "docker", d.GetSSHUsername())); err != nil {
 				return errors.Wrap(err, "usermod")
 			}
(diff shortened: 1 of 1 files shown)


afbjorklund (approved):
Sounds good to me, but it could still require (passwordless) sudo for _other_ operations

kaylendog:
> Sounds good to me, but it could still require (passwordless) sudo for _other_ operations

Yes, I'm intending this to be more of a quick patch to enable minikube to get pass the usermod command. There's definitely some more stuff to do here!

afbjorklund:
Sounds reasonable to avoid the redundant commands, I think it was a "quick fix" for bugs in the provisioner.

a0d2c1a0c5b9d1f99fc607183d3043b0c5409b03

For legacy reasons it assumed that docker is in the root group, but that podman is run by using `sudo`
(there is no reason why it couldn't run docker with `sudo`, or have podman use the "podman" root group)

Once the rootless drivers entered the scene, it was supposed to be done in a more orderly fashion

But since this is the bare metal (well, virtual machine) drivers - that applies to the cluster container runtime.
And that one by default run as root, it will probably be "a while" before all of Kubernetes can run as rootless.

PR #15530, opened 2022-12-17
Added `--subnet` validation

### LINKED ISSUE #15517: Add `--subnet` validation
If the IP/CIDR passed in is not a private IP it should fail during validation as the way the incrementing and retries work it's impossible to start with a public IP and for it to increment into a private IP. Also, we should require that if a CIDR is passed in that the mask is no greater than `/30`. A mask of `/30` gives a range of 4 IPs, anything larger reduces that number and is guaranteed to fail as minikube requires at least 4 IPs.

shubhbapna:
fixes #15517 

Added `validateSubnet` that makes sure the given subnet has a valid private IP address and if it is cidr then the mask shouldn't be greater than 30

To reduce code duplication, I refactored `inspect` to use `ParseAddr`

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -47,6 +47,7 @@ import (
 	"golang.org/x/text/cases"
 	"golang.org/x/text/language"
 	"k8s.io/minikube/pkg/minikube/command"
+	netutil "k8s.io/minikube/pkg/network"
 
 	"k8s.io/klog/v2"
 	cmdcfg "k8s.io/minikube/cmd/minikube/cmd/config"
@@ -1212,6 +1213,13 @@ func validateFlags(cmd *cobra.Command, drvName string) {
 
 	}
 
+	if cmd.Flags().Changed(subnet) {
+		err := validateSubnet(viper.GetString(subnet))
+		if err != nil {
+			exit.Message(reason.Usage, "{{.err}}", out.V{"err": err})
+		}
+	}
+
 	if cmd.Flags().Changed(containerRuntime) {
 		err := validateRuntime(viper.GetString(containerRuntime))
 		if err != nil {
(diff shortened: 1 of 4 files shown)


shubhbapna:
@spowelljr  I have marked it as draft since `make lint-ci` was failing for me with the following error
```
cmd/minikube/cmd/start.go:1177:1: cyclomatic complexity 32 of func `validateFlags` is high (> 30) (gocyclo)
func validateFlags(cmd *cobra.Command, drvName string) {
^
```
But I don't see any way around reducing cyclomatic complexity since I did need to add `validateSubnet` to this function which adds another path to it. Any suggestions?

cmd/minikube/cmd/start.go
```diff
@@ -1763,6 +1752,48 @@ func validateDockerStorageDriver(drvName string) {
...
+		return err
+	}
+	if !ip.IsPrivate() {
+		return errors.Errorf("Sorry, %s is not a private IP", ip)
+	}
+
+	if cidr != nil {
+		mask, _ := cidr.Mask.Size()
+		if mask > 30 {
+			return errors.Errorf("Sorry, mask must be less than /30")
```
spowelljr, line 1767: Less than or equal to

cmd/minikube/cmd/start.go
```diff
@@ -1763,6 +1752,48 @@ func validateDockerStorageDriver(drvName string) {
 	viper.Set(preload, false)
 }
 
+func validateSubnet(subnet string) error {
+	ip, cidr, err := netutil.ParseAddr(subnet)
+	if err != nil {
+		return err
```
spowelljr, line 1758: Should wrap this will some helpful text like the others

cmd/minikube/cmd/start.go
```diff
@@ -1763,6 +1752,48 @@ func validateDockerStorageDriver(drvName string) {
...
+	}
+	if !ip.IsPrivate() {
+		return errors.Errorf("Sorry, %s is not a private IP", ip)
```
spowelljr, line 1761: The message should include something about this being a subnet, because there are many flags in minikube that are IP related `Sorry, the subnet provided is not...`

spowelljr (changes requested):

cmd/minikube/cmd/start.go
spowelljr: I think you might have forgot to wrap this error

spowelljr (changes requested):

medyagh:
[shubhbapna](https://github.com/shubhbapna) are you still working on this ?

shubhbapna:
Yep just came back from the holidays! Will start working on it again today

spowelljr (approved):
Looks good, thank you for the PR

PR #15541, opened 2022-12-21
ISO: Update docker, containerd, and cri-dockerd

spowelljr:
https://docs.docker.com/engine/release-notes/#201022

https://github.com/containerd/containerd/releases/tag/v1.6.15

https://github.com/Mirantis/cri-dockerd/releases/tag/v0.3.0


spowelljr:
ok-to-build-iso

afbjorklund:
Historically we used the same versions as Docker, to not have to test/verify it... But it _was_ separated, since then.

```
	# As of 2019-05, we use upstream containerd so that we may update it independently of docker.

	# As of 2019-01, we use upstream runc so that we may update it independently of docker.
```

afbjorklund:
It would be easier to provision different versions, if the runtimes were separated from eachother.

afbjorklund (approved):

afbjorklund:
Sounds like a good reason to upgrade

mqasimsarfraz:
> Sounds like a good reason to upgrade (I forgot if the Buildroot ISO has swap memory controller or not)
>
I believe the ISO do have swap memory controller. 
> But somehow I think that pesky setting is debian-specific ?
Yes (maybe Ubuntu 20.04 as well, need to confirm) so better upgrade I reckon.

mqasimsarfraz:
> Anyway, it is enabled by default in recent kernels.

Ah okay, got it! Just for understanding, would upgrading the version here also updates it for docker driver? Like kickbase image?

mqasimsarfraz (approved):

deploy/iso/minikube-iso/arch/x86_64/package/containerd-bin/containerd-bin.hash
```diff
@@ -18,3 +18,4 @@ sha256 f422e21e35705d1e741c1f3280813e43f811eaff4dcc5cdafac8b8952b15f468 v1.6.4.t
 sha256 27afb673c20d53aa5c31aec07b38eb7e4dc911e7e1f0c76fac9513bbf070bd24 v1.6.6.tar.gz
 sha256 f5f938513c28377f64f85e84f2750d39f26b01262f3a062b7e8ce35b560ca407 v1.6.8.tar.gz
 sha256 a034b2273533207d5d96bef8bd3fce1efff85139815efb756d90c705ae1a05ce v1.6.9.tar.gz
+sha256 d6d022385c4dadd81edf4fb5ad4b084dcfc3d6d921ed433bb3aa0d2e0bfbb19d v1.6.13.tar.gz
```
AkihiroSuda, line 21: 1.6.13 had a serious regression. Please use v1.6.14
https://github.com/containerd/containerd/releases/tag/v1.6.14

afbjorklund: Could be a good reason to start decoupling containerd and docker (20.10.22 contains 1.6.13 still)

spowelljr:
ok-to-build-iso

spowelljr:
ok-to-build-iso

spowelljr:
ok-to-build-image

medyagh (approved):

deploy/iso/minikube-iso/arch/x86_64/package/containerd-bin/containerd-bin.hash
spowelljr: Updated to v1.6.15

PR #15546, opened 2022-12-23
Changing the kubeadm registry was backported

afbjorklund:
To kubernetes versions: 1.22.18, 1.23.15, 1.24.9

Versions -1.21 use old, versions 1.26+ use new.

Follow up to:
* https://github.com/kubernetes/minikube/pull/14768

Reference:


afbjorklund:
If we override the default value and retag the preload after loading, we can change even the existing releases.

`--image-repository=registry.k8s.io`

`docker tag k8s.gcr.io/pause:3.2 registry.k8s.io/pause:3.2`

But it will go in a separate PR, since it is a separate feature. Changing the current releases is a good start...

afbjorklund:
The binaries were changed here, to use `dl.k8s.io`:

* #15840

spowelljr:
Due to https://github.com/kubernetes/minikube/issues/16017 we're always going to be using the new registry so this PR is no longer needed. Superseded by https://github.com/kubernetes/minikube/pull/16029

afbjorklund:
It was mostly if you wanted a more fine-grained report on when `kubeadm` default was bumped ?

But it should be safe to supply the kubernetes registry as a parameter, even when it _is_ the default...

PR #15552, opened 2022-12-27
Update Korean translation

mardi2020:


spowelljr (approved):
Thanks for the translations

PR #15571, opened 2023-01-02
refactor: avoiding repetitive code

axrav:
avoiding repetitive code where checking if the instance is vm or not which wont' make a difference as the core functionality is still the same.

### DIFF
--- a/pkg/minikube/registry/global.go
+++ b/pkg/minikube/registry/global.go
@@ -123,14 +123,10 @@ func Available(vm bool) []DriverState {
 		if !s.Healthy {
 			priority = Unhealthy
 		}
-
-		if vm {
-			if IsVM(d.Name) {
-				sts = append(sts, DriverState{Name: d.Name, Default: d.Default, Preference: preference, Priority: priority, State: s})
-			}
-		} else {
-			sts = append(sts, DriverState{Name: d.Name, Default: d.Default, Preference: preference, Priority: priority, State: s})
+		if vm && !IsVM(d.Name) {
+			continue
 		}
+		sts = append(sts, DriverState{Name: d.Name, Default: d.Default, Preference: preference, Priority: priority, State: s})
 	}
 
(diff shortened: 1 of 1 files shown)


sharifelgamal:
The VM boolean is choice is necessary, so we can filter out non-VM choices (none, docker) when  `--vm=true` is passed into `minikube start`

axrav:
@sharifelgamal but the bool doesnt actually makes any difference, the same code is being used for both cases even if its true/false which can be avoided.

spowelljr:
> @sharifelgamal but the bool doesnt actually makes any difference, the same code is being used for both cases even if its true/false which can be avoided.

```
		if vm {
			if IsVM(d.Name) {
				sts = append(sts, DriverState{Name: d.Name, Default: d.Default, Preference: preference, Priority: priority, State: s})
			}
		}
```
It loops through the list of drivers and if the user specifies `vm=true` it only adds VM drivers to the list. If you remove the `if` statements if will add all drivers to the list regardless if the user only wants VM drivers or not.

spowelljr:
You could reduce a bit of code if you did:
```
if vm && !IsVM(d.Name) {
        continue
}
```
But the if statements are still necessary

axrav:
@spowelljr yeah, this makes sense, I'll just fix that and commit for the same in a while.

axrav:
@spowelljr @sharifelgamal I have made changes accordingly, you can have a look.

spowelljr:
I think that `if` statement should be at the top of the `for` loop. Currently it's getting the status of all drivers, even non-VM drivers, but if the user wants to use a VM driver getting the status of Docker is irrelevant as the result is thrown away anyway.

axrav:
yeah thats actually correct!, I'll make those changes

axrav:
the changes have been made you can have a look.!

spowelljr (approved):
Looks good, thanks for the PR

PR #15578, opened 2023-01-03
Add workaround for broken DNS DNAT on Fedora 37

### LINKED ISSUE #15573: Minikube to host communication not working on Fedora 37
### What Happened?

There seems to be a difference in minikube iptables rules when comparing a fully updated Fedora 36 and Fedora 37 system.
On Fedora 36:
```
$ sudo iptables -t nat -S|grep -e '--to-destination 127.0.0.11'
-A DOCKER_OUTPUT -d 192.168.49.1/32 -p tcp -m tcp --dport 53 -j DNAT --to-destination 127.0.0.11:39397
-A DOCKER_OUTPUT -d 192.168.49.1/32 -p udp -m udp --dport 53 -j DNAT --to-destination 127.0.0.11:34196
```
On Fedora 37:
```
$ sudo iptables -t nat -S|grep -e '--to-destination 127.0.0.11'
-A DOCKER_OUTPUT -d 192.168.49.1/32 -p tcp -j DNAT --to-destination 127.0.0.11:46739
-A DOCKER_OUTPUT -d 192.168.49.1/32 -p udp -j DNAT --to-destination 127.0.0.11:37392
```
The missing `--dport 53` condition on the destination NAT breaks all non-DNS communication between host and minikube. 
What might be causing this difference? 

### Attach the log file

[log.txt](https://github.com/kubernetes/minikube/files/10332315/log.txt)

### Operating System

Redhat/Fedora

### Driver

Docker

mnk:
fixes #15573

### DIFF
--- a/deploy/kicbase/entrypoint
+++ b/deploy/kicbase/entrypoint
@@ -447,6 +447,8 @@ enable_network_magic(){
       -e 's/-A OUTPUT \(.*\) -j DOCKER_OUTPUT/\0\n-A PREROUTING \1 -j DOCKER_OUTPUT/' \
       `# switch docker DNS SNAT rules rules to our chosen IP` \
       -e "s/--to-source :53/--to-source ${docker_host_ip}:53/g"\
+      `# add mssing port condition in DNS DNAT rules` \
+      -e "s/p -j DNAT --to-destination ${docker_embedded_dns_ip}/p --dport 53 -j DNAT --to-destination ${docker_embedded_dns_ip}/g"\
     | iptables-restore
 
   # now we can ensure that DNS is configured to use our IP


mnk:
I think the root-cause should be found and fixed instead, but this could be a temporary workaround.

medyagh:
thank you for this contribution, we inherit this file from kind project, I recommend also making the same contribution to the upstream
https://github.com/kubernetes-sigs/kind/blob/main/images/base/files/usr/local/bin/entrypoint


CC: @BenTheElder

deploy/kicbase/entrypoint
```diff
@@ -447,6 +447,8 @@ enable_network_magic(){
       -e 's/-A OUTPUT \(.*\) -j DOCKER_OUTPUT/\0\n-A PREROUTING \1 -j DOCKER_OUTPUT/' \
       `# switch docker DNS SNAT rules rules to our chosen IP` \
       -e "s/--to-source :53/--to-source ${docker_host_ip}:53/g"\
+      `# add mssing port condition in DNS DNAT rules` \
+      -e "s/p -j DNAT --to-destination ${docker_embedded_dns_ip}/p --dport 53 -j DNAT --to-destination ${docker_embedded_dns_ip}/g"\
```
BenTheElder, line 451: This seems extra fragile?

The iptables rules here come from docker and aren't documented, we're just swapping the IP address in the otherwise existing rules to move from loopback to an address we can route to from (nested) containers.

Can you share the _full_ iptables-save between the two systems? https://github.com/kubernetes/minikube/issues/15573

mnk: I agree that the workaround is very fragile, and should probably not be used as-is.
I have attached full iptables-save to https://github.com/kubernetes/minikube/issues/15573

BenTheElder: I see no `--port` on my ~debian system as well, I wonder why docker changed this. I guess even if we got the upstream rule changed back we'd need to work around systems on version inbetween.

mnk: Fedora 36 and 37 shows same version of moby-engine (20.10.21-1) and containerd (1.6.14-2), but different versions of iptables-nft (1.8.7-15.fc36 vs 1.8.8-3.fc37). Could that be the source of the change?

BenTheElder: Hmm maybe, also noting the kind image has 1.8.7 currently.

I wouldn't expect something like this to disappear but kubernetes has had issues in the past with mismatched versions. 

The initial iptables rules will be done with the host binary (not the one in the image), but the altered rules will be written with the binary in the image.

Are the configs identical for dockerd?

prezha: @mnk thanks for reporting this issue

there are a couple of improvements i'm currently working on in draft pr #15463 

could you please try with https://storage.googleapis.com/minikube-builds/15463/minikube-linux-amd64 and let us know if it works for you

if not, could you pull that pr and run:

`make integration -e TEST_ARGS="-minikube-start-args='--driver=docker --container-runtime=docker --alsologtostderr -v=7' -test.run TestNetworkPlugins --cleanup=true"`

then share the _whole_ output you get

i've tried to replicate your setup (ie, fresh fedora 37 install [in kvm] + docker) and all the tests above passed for me, so i'm curious to know if it would work for you

prezha: from the [iptables-save-f37.txt](https://github.com/kubernetes/minikube/files/10347871/iptables-save-f37.txt) logs you've provided, coredns got ip 172.17.0.2 whereas it should have been attached to the bridge "cni" (so getting something from default pod cidr - eg, 10.244.0.2)

i'm _almost_ certain that pr will help with your problem :)

here are the logs i got with fedora37:
[minikube_start - fedora37.log](https://github.com/kubernetes/minikube/files/10355742/minikube_start.-.fedora37.log)
[iptables-save - fedora37.log](https://github.com/kubernetes/minikube/files/10355744/iptables-save.-.fedora37.log)

BenTheElder: I agree with @mnk's assessment. I don't think anything currently in #15463 is related.

mnk:
I think this is the right solution: https://github.com/kubernetes-sigs/kind/pull/3059

PR #5955, opened 2019-11-20
Fix profile list json command output

cheeseandcereal:
Fixes #5898

Behavior of `minikube profile list -o json` was dependent whether or not the `$MINIKUBE_HOME/profiles` folder exists or not.

This small update simply checks if the error is due to the folder not existing, and if so, simply returns the normal empty array results with a normal exit code

Before:
```
$ ls ~/.minikube/profiles
ls: cannot access '/home/adam/.minikube/profiles': No such file or directory
$ minikube profile list -o json
{"error":{"Op":"open","Path":"/root/.minikube/profiles","Err":2}}
$ echo $?
1
$ mkdir ~/.minikube/profiles
$ minikube profile list -o json
{"invalid":[],"valid":[]}
$ echo $?
0
```

After:
```
$ ls ~/.minikube/profiles
ls: cannot access '/home/adam/.minikube/profiles': No such file or directory
$ minikube profile list -o json
{"invalid":[],"valid":[]}
$ echo $?
0
$ mkdir ~/.minikube/profiles
$ minikube profile list -o json
{"invalid":[],"valid":[]}
$ echo $?
0
```

Issue #5898, opened 2019-11-13
Inconsistent behavior listing profiles in json when none exist


cheeseandcereal:
Using minikube version 1.5.2 tested on Linux and Windows 10 with the same behavior:

The command `minikube profile list -o json` provides inconsistent behavior whether or not the `$MINIKUBE_HOME/profiles` folder exists or not.

When `$MINIKUBE_HOME/profiles` does not exist, the command outputs: `{"error":{"Op":"open","Path":"/root/.minikube/profiles","Err":2}}` and minikube exits with status 1

When that folder does exist, the command outputs: `{"invalid":[],"valid":[]}` with exit status 0 (as expected).

This seems like undesirable behavior, especially because `minikube profile list` (without json output) is consistent in both scenarios, outputting the generic ```* No minikube profile was found. You can create one using `minikube start`.``` error message and exiting with status code 64 regardless of whether the `$MINIKUBE_HOME/profiles` folder exists or not

Example of full replication output in bash:
```
root@d3a5e2df82d8:/# minikube profile list
* No minikube profile was found. You can create one using `minikube start`.
root@d3a5e2df82d8:/# echo $?
64
root@d3a5e2df82d8:/# minikube profile list -o json
{"error":{"Op":"open","Path":"/root/.minikube/profiles","Err":2}}
root@d3a5e2df82d8:/# echo $?
1
root@d3a5e2df82d8:/# mkdir ~/.minikube/profiles
root@d3a5e2df82d8:/# minikube profile list
* No minikube profile was found. You can create one using `minikube start`.
root@d3a5e2df82d8:/# echo $?
64
root@d3a5e2df82d8:/# minikube profile list -o json
{"invalid":[],"valid":[]}
root@d3a5e2df82d8:/# echo $?
0
```

Minikube should probably ensure that the `$MINIKUBE_HOME/profiles` folder exists (and create it if missing) before the profile list operation in order to provide consistent, expected behavior.

priyawadhwa:
Hey @cheeseandcereal thanks for opening this issue. Definitely agree the behavior is a bit confusing and could be improved.

WDYT of outputting:

`* No minikube profile was found. You can create one using 'minikube start'.` to STDOUT
`{"invalid":[],"valid":[]}` for `-o json`
and exit code 0

for both cases (when `$MINIKUBE_HOME/profiles` exists and when it doesn't`)?

[...]

PR #15594, opened 2023-01-04
Fix profile list when profiles directory is missing

### LINKED ISSUE #15593: minikube profile list --output json fails with ENONT after fresh install
### What Happened?

After minikube on a fresh machine, running:

    minikube profile list --output json

Fails, writing error to stdout:

    {"error":{"Op":"open","Path":"/home/runner/.minikube/profiles","Err":2}}

The expected output:

    {"invalid":[],"valid":[]}

Issues:
- Missing ~/.minikube/profiles not handled gracefully
- Reporting errors to stdout
- Reporting error number (2) instead of description (No such file or directory).

Workaround:
- Create .minikube/profiles manually before running the command

### Attach the log file

Full reproducer:

Add this github workflow:
```
name: CI
on:
  - push
  - pull_request
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - name: Reproduce minikube bug
        run: |
          curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
          sudo install minikube-linux-amd64 /usr/local/bin/minikube
          minikube --alsologtostderr --v=7 profile list --output json
```

Example run:

```
I0104 21:22:57.498168    1632 out.go:296] Setting OutFile to fd 1 ...
I0104 21:22:57.498301    1632 out.go:343] TERM=,COLORTERM=, which probably does not support color
I0104 21:22:57.498306    1632 out.go:309] Setting ErrFile to fd 2...
I0104 21:22:57.498311    1632 out.go:343] TERM=,COLORTERM=, which probably does not support color
I0104 21:22:57.503171    1632 root.go:334] Updating PATH: /home/runner/.minikube/bin
W0104 21:22:57.503308    1632 root.go:311] Error reading config file at /home/runner/.minikube/config/config.json: open /home/runner/.minikube/config/config.json: no such file or directory
I0104 21:22:57.503638    1632 out.go:303] Setting JSON to true
I0104 21:22:57.503749    1632 notify.go:220] Checking for updates...
{"error":{"Op":"open","Path":"/home/runner/.minikube/profiles","Err":2}}
```

Can also be reproduced by removing (an empty) `~/.minikube/profiles` directory.

### Operating System

Ubuntu

### Driver

N/A

nirs:
This was reported in [1] and fixed in [2] in 2019, but later it was
broken by [3] in commit c3aafaeeb4541b0cd5303afe211fe281dbf5e4f3.
    
`listProfiles()` can return `os.ErrNotExist` from `os.ReadDir()` so it
must be tested with `os.IsNotExist()`.
    
[1] https://github.com/kubernetes/minikube/issues/5898
[2] https://github.com/kubernetes/minikube/pull/5955
[3] https://github.com/kubernetes/minikube/pull/6440

Fixes #15593

### DIFF
--- a/cmd/minikube/cmd/config/profile_list.go
+++ b/cmd/minikube/cmd/config/profile_list.go
@@ -205,7 +205,7 @@ func printProfilesJSON() {
 	updateProfilesStatus(validProfiles)
 
 	var body = map[string]interface{}{}
-	if err == nil || config.IsNotExist(err) {
+	if err == nil || os.IsNotExist(err) {
 		body["valid"] = profilesOrDefault(validProfiles)
 		body["invalid"] = profilesOrDefault(invalidProfiles)
 		jsonString, _ := json.Marshal(body)


nirs:
Tested on Fedora 37 by removing the profiles directory:

```
$ printenv MINIKUBE_HOME
/data/minikube

$ rm -r /data/minikube/.minikube/profiles
```

Before the fix:
```
$ minikube profile list -o json
{"error":{"Op":"open","Path":"/data/minikube/.minikube/profiles","Err":2}}
```

After the fix:
```
$ out/minikube profile list -o json
{"invalid":[],"valid":[]}
```

nirs:
@tstromberg can you take a look?

cmd/minikube/cmd/config/profile_list.go
```diff
@@ -205,7 +205,7 @@ func printProfilesJSON() {
 	updateProfilesStatus(validProfiles)
 
 	var body = map[string]interface{}{}
-	if err == nil || config.IsNotExist(err) {
+	if err == nil || os.IsNotExist(err) {
```
medyagh, line 208: This PR is  bypassing the config. package, the better approach would be moving this logic down into the config package pkg/minikube/config/config.go
the relevant code is here:
```
func (c *simpleConfigLoader) LoadConfigFromFile(profileName string, miniHome ...string) (*ClusterConfig, error) {
	var cc ClusterConfig
	// Move to profile package
	path := profileFilePath(profileName, miniHome...)

	if _, err := os.Stat(path); err != nil {
		if os.IsNotExist(err) {
			return nil, &ErrNotExist{fmt.Sprintf("cluster %q does not exist", profileName)}
		}
		return nil, errors.Wrap(err, "stat")
	}


```

so I am curious can we check (by adding temporarily logs) if our code does NOT detect an error here ? 	`if _, err := os.Stat(path); err != nil {`
https://github.com/medyagh/minikube/blob/4d631bab7e32635a90afbbe6a8677e675be7d69c/pkg/minikube/config/config.go#L210

nirs: I agree this is not the best fix, but it is the simplest way to fix this regression
safely - basically restoring the original fix from 2019.

In general passing an expected condition (no profiles dir) as an error from `profileDirs()` is the real issue.

```
// profileDirs gets all the folders in the user's profiles folder regardless of valid or invalid config
func profileDirs(miniHome ...string) (dirs []string, err error) {                                                                   
    ...   
    pRootDir := filepath.Join(miniPath, "profiles")
    items, err := os.ReadDir(pRootDir)
    for _, f := range items {
        if f.IsDir() {
            dirs = append(dirs, f.Name())
        }   
    }   
    return dirs, err 
}
```

We can handle it early as expected condition:

```
diff --git a/cmd/minikube/cmd/config/profile_list.go b/cmd/minikube/cmd/config/profile_list.go
index 2cd2f1432..61eff8e25 100644
--- a/cmd/minikube/cmd/config/profile_list.go
+++ b/cmd/minikube/cmd/config/profile_list.go
@@ -203,11 +203,11 @@ func warnInvalidProfiles(invalidProfiles []*config.Profile) {
 func printProfilesJSON() {
        validProfiles, invalidProfiles, err := listProfiles()
        updateProfilesStatus(validProfiles)
 
        var body = map[string]interface{}{}
-       if err == nil || config.IsNotExist(err) {
+       if err == nil {
                body["valid"] = profilesOrDefault(validProfiles)
                body["invalid"] = profilesOrDefault(invalidProfiles)
                jsonString, _ := json.Marshal(body)
                os.Stdout.Write(jsonString)
        } else {
diff --git a/pkg/minikube/config/profile.go b/pkg/minikube/config/profile.go
index d14a8f732..040ac8d53 100644
--- a/pkg/minikube/config/profile.go
+++ b/pkg/minikube/config/profile.go
@@ -299,10 +299,16 @@ func profileDirs(miniHome ...string) (dirs []string, err error) {
        if len(miniHome) > 0 {
                miniPath = miniHome[0]
        }
        pRootDir := filepath.Join(miniPath, "profiles")
        items, err := os.ReadDir(pRootDir)
+
+       // Missing profiles dir is expected.
+       if err != nil && os.IsNotExist(err) {
+               return dirs, nil
+       }
+
        for _, f := range items {
                if f.IsDir() {
                        dirs = append(dirs, f.Name())
                }
        }

```

and then no check is needed when printing the json later.

What do you think?

nirs: @medyagh what do think about https://github.com/kubernetes/minikube/pull/15594#discussion_r1062956489?

medyagh: I think it is better to not rely on os here and use our package that handles it, and if our packag edo snot handle it, we could improve it to do it

nirs: I'm not sure how do yo suggest to solve this.

The code for listing profiles directories is documented as:
> // profileDirs gets all the folders in the user's profiles folder regardless of valid or invalid config

So using os.ReadDir() look like the right thing to do. The issue is propagating
os.Error to the caller instead of config.Error.

In https://github.com/kubernetes/minikube/pull/15594#discussion_r1062956489
we handle the expected condition of missing profiles directory so we don't
propagate os.Error to the rest of the code. This avoids the confusing when
check if the error is non existing directory.

medyagh: I think what we can do is create a new error type and the first function that returns the err can create this new error type and the upstream funcs can check against this err type

nirs: This is possible, but I don't think this condition should be reported as an error.
This is an expected condition when starting minikube in the first time, or after 
a user deletes the .minikube directory.

PR #15622, opened 2023-01-10
Use `IsPrivate` function provided by `net` pkg

### LINKED ISSUE #15544: `network` doesn't use the default `IsPrivate` function provided by `net` pkg
### What Happened?

From go 1.17, `IsPrivate` was added to the net pkg. So we don't need `isSubnetPrivate` function (See discussion in #15517)

While trying to migrate to `IsPrivate`, it was discovered that [TestFreeSubnet](https://github.com/kubernetes/minikube/blob/master/pkg/network/network_test.go#L51) does not restore the mocks after each test case which has an effect on the subsequent tests giving incorrect test results and coverage.

For eg: [NonPrivateSubnet](https://github.com/kubernetes/minikube/blob/master/pkg/network/network_test.go#L51) test ends up using the mocked `inspect` function from the [FirstSubnetIPV6NetworkFound](https://github.com/kubernetes/minikube/blob/master/pkg/network/network_test.go#L86) test

### Attach the log file

NA

### Operating System

None

### Driver

N/A

shubhbapna:
Fixes #15544

Removed `isSubnetPrivate` function and started using the default `IsPrivate` function provided by the `net` pkg. Updated test cases and added restoration of mocked functions to avoid side effects on subsequent tests.

### DIFF
--- a/pkg/network/network.go
+++ b/pkg/network/network.go
@@ -28,28 +28,7 @@ import (
 
 const defaultReservationPeriod = 1 * time.Minute
 
-var (
-	reservedSubnets = sync.Map{}
-
-	// valid private network subnets (RFC1918)
-	privateSubnets = []net.IPNet{
-		// 10.0.0.0/8
-		{
-			IP:   []byte{10, 0, 0, 0},
-			Mask: []byte{255, 0, 0, 0},
-		},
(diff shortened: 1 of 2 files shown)


spowelljr (approved):
Thanks for the follow up PR

Issue #15274, opened 2022-11-02
virtualbox: unable to start minikube on macOS Ventura


andriymahats:
### What Happened?

After upgrading macOS to Ventura from Monterey the minikube start started to fail with the response:

```sh
😄  minikube v1.27.1 on Darwin 13.0
✨  Using the virtualbox driver based on user configuration
👍  Starting control plane node minikube in cluster minikube
🔥  Creating virtualbox VM (CPUs=2, Memory=8192MB, Disk=20000MB) ...
🔥  Deleting "minikube" in virtualbox ...
🤦  StartHost failed, but will try again: creating host: create: creating: Error setting up host only network on machine start: The host-only adapter we just created is not visible. This is a well known VirtualBox bug. You might want to uninstall it and reinstall at least version 5.0.12 that is is supposed to fix this issue
🔥  Creating virtualbox VM (CPUs=2, Memory=8192MB, Disk=20000MB) ...
😿  Failed to start virtualbox VM. Running "minikube delete" may fix it: creating host: create: creating: Error setting up host only network on machine start: The host-only adapter we just created is not visible. This is a well known VirtualBox bug. You might want to uninstall it and reinstall at least version 5.0.12 that is is supposed to fix this issue

❌  Exiting due to IF_VBOX_NOT_VISIBLE: Failed to start host: creating host: create: creating: Error setting up host only network on machine start: The host-only adapter we just created is not visible. This is a well known VirtualBox bug. You might want to uninstall it and reinstall at least version 5.0.12 that is is supposed to fix this issue
💡  Suggestion: Reboot to complete VirtualBox installation, verify that VirtualBox is not blocked by your system, and/or use another hypervisor
📘  Documentation: https://stackoverflow.com/questions/52277019/how-to-fix-vm-issue-with-minikube-start
🍿  Related issues:
    ▪ https://github.com/kubernetes/minikube/issues/3614
    ▪ https://github.com/kubernetes/minikube/issues/4222
    ▪ https://github.com/kubernetes/minikube/issues/5817
```

### Attach the log file

[log.txt](https://github.com/kubernetes/minikube/files/9922957/log.txt)

### Operating System

macOS (Default)

### Driver

VirtualBox

crengifo:
same issue, i've tried minikube delete and start and update virtualbox and nothing...

hakjoonglee:
Same issue. I also tried minikube delete and start and updated VirtualBox version 7.0.2

TestRep1:
[...]

PR #15624, opened 2023-01-10
Warn users if using VirtualBox on macOS 13+

spowelljr:
Related https://github.com/kubernetes/minikube/issues/15274

```
$ minikube start --driver virtualbox
❗  Due to changes in macOS 13+ minikube doesn't currently support VirtualBox. You can use alternative drivers such as docker or qemu.

    For more details on the issue see: https://github.com/kubernetes/minikube/issues/15274

...
```

If the user tries using the VirtualBox driver on macOS 13+ the warning above is output.

If the machine is amd64 it will recommend the hyperkit driver, if arm64 it will recommend the qemu driver.


cmd/minikube/cmd/start.go
```diff
@@ -371,6 +372,20 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
...
 
+func virtualBoxMacOS13PlusWarning(driverName string) {
+	if driverName != "virtualbox" || !detect.MacOS13Plus() {
+		return
+	}
+	driver := "hyperkit"
+	if runtime.GOARCH == "arm64" {
+		driver = "qemu"
+	}
+	out.WarningT(`Due to changes in macOS 13+ minikube doesn't currently support VirtualBox. We recommend using the {{.driver}} driver: https://minikube.sigs.k8s.io/docs/drivers/{{.driver}}/
```
medyagh, line 383: instead of we...

let say something like
"You could use an alternative driver such as Qemu or Docker"

medyagh (changes requested):

cmd/minikube/cmd/start.go
spowelljr: Updated message

medyagh (approved):

PR #15628, opened 2023-01-11
Update kvm docker machine base image

spowelljr:
- Updated `without_lxc` to `libvirt_without_lxc` as per https://gitlab.com/libvirt/libvirt-go-module/-/merge_requests/34
- Updated base image of `docker-machine-driver-kvm2-amd64` builder from Ubuntu 16.04 to Ubuntu 20.04 to fix libvirt build error
- Added `docker system prune -a --volumes -f` to cross build to clear out previous run artifacts

This makes the build on https://github.com/kubernetes/minikube/pull/15537 complete

### DIFF
--- a/.github/workflows/functional_verified.yml
+++ b/.github/workflows/functional_verified.yml
@@ -46,7 +46,6 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y libvirt-dev
-          ls -la ./out
           MINIKUBE_BUILD_IN_DOCKER=y make cross e2e-cross debs
           strings out/docker-machine-driver-kvm2-amd64 | grep ^LIBVIRT | sort
           shasum -a 256 out/docker-machine-driver-kvm2-amd64
--- a/.github/workflows/functional_verified.yml
+++ b/.github/workflows/functional_verified.yml
@@ -46,10 +46,16 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y libvirt-dev
+          ls -la ./out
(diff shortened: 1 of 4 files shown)


spowelljr:
There seems there might be something weird going on in CI.

If I build the `docker-machine-driver-kvm2-amd64` binary locally (but through docker) the libvirt version is 6.0.0, which is expected.

But if I download https://storage.googleapis.com/minikube-builds/15628/docker-machine-driver-kvm2-amd64 the libvirt version is still old (1.2.19). However the LXC versions are correctly removed, so the build is successful but it's like it's using an old version of the image (Ubuntu 16) to build the with.

Investigating what might be going on

spowelljr:
With debugging I was able to confirm that the file contains the expected libvirt version in CI, but the uploaded file has a different libvirt version, will continue investigating

spowelljr:
Confirmed that the binary shasum doesn't match the file uploaded
`276b2b9060f961071b5531c81749064ca8783fc1b577e1e2b8db9a323d51daff  out/docker-machine-driver-kvm2-amd64`
`be3d176c1eb1ad0ea8b738466339d2469679119667bbd2925cad4efd80e06a6d  ./docker-machine-driver-kvm2-amd64`

spowelljr:
Discovered the root of the problem, we aren't purging Docker on build in Jenkins so when is checks for the existing image it exists and doesn't rebuild it, resulting in my Dockerfile changes not being built and used. Will remove all the debugging changes and update CI to purge Docker on start.

medyagh (approved):

PR #15642, opened 2023-01-13
Use absolute path when calling crictl version

afbjorklund:
Some legacy systems don't have /usr/local/bin in secure_path

So they are not able to find any locally installed tools...

Regression:

* https://github.com/kubernetes/minikube/pull/6220

### DIFF
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -471,8 +471,15 @@ func waitForCRIVersion(runner cruntime.CommandRunner, socket string, wait int, i
 
 	klog.Infof("Will wait %ds for crictl version", wait)
 
+	cmd := exec.Command("which", "crictl")
+	rr, err := runner.RunCmd(cmd)
+	if err != nil {
+		return err
+	}
+	crictl := strings.TrimSuffix(rr.Stdout.String(), "\n")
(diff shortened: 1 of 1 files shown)


pkg/minikube/node/start.go
```diff
@@ -471,8 +471,15 @@ func waitForCRIVersion(runner cruntime.CommandRunner, socket string, wait int, i
 
 	klog.Infof("Will wait %ds for crictl version", wait)
 
+	cmd := exec.Command("which", "crictl")
+	rr, err := runner.RunCmd(cmd)
+	if err != nil {
+		return err
+	}
+	crictl := strings.TrimSuffix(rr.Stdout.String(), "\n")
```
medyagh, line 479: maybe use  "TrimSpace" ? for removing all sort of unexpected spaces ?

medyagh (approved):
Looks good but would look better with TrimSpace

pkg/minikube/node/start.go
afbjorklund: There is no whitespace expected, except for the newline from the `which` command

```console
$ which crictl | cat -vte
/usr/local/bin/crictl$
```

PR #15646, opened 2023-01-13
Add check for Docker Desktop 4.16.0

spowelljr:
```
$ docker version --format {{.Server.Platform.Name}}
Docker Desktop 4.16.0 (95345)

$ ./out/minikube start --driver docker

💣  Exiting due to PROVIDER_DOCKER_DESKTOP_VERSION_BAD: Docker Desktop 4.16.0 has a regression that prevents minikube from starting
💡  Suggestion: Update Docker Desktop to 4.16.1 or greater

```

### DIFF
--- a/pkg/minikube/registry/drvs/docker/docker.go
+++ b/pkg/minikube/registry/drvs/docker/docker.go
@@ -137,7 +137,10 @@ func status() (retState registry.State) {
 
 	klog.Infof("docker version: %s", o)
 	if !viper.GetBool("force") {
-		s := checkDockerVersion(strings.TrimSpace(string(o))) // remove '\n' from o at the end
+		if s := checkDockerDesktopVersion(); s != nil {
+			return *s
+		}
+		s := checkDockerEngineVersion(strings.TrimSpace(string(o))) // remove '\n' from o at the end
 		if s.Error != nil {
 			return s
 		}
@@ -159,7 +162,7 @@ func status() (retState registry.State) {
 	return checkNeedsImprovement()
 }
 
-func checkDockerVersion(o string) registry.State {
+func checkDockerEngineVersion(o string) registry.State {
 	parts := strings.SplitN(o, "-", 2)
 	if len(parts) != 2 {
 		return registry.State{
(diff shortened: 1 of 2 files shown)


pkg/minikube/registry/drvs/docker/docker.go
```diff
@@ -240,6 +243,35 @@ func checkDockerVersion(o string) registry.State {
 		Doc:              docURL + "#requirements"}
 }
 
+func checkDockerDesktopVersion() *registry.State {
```
medyagh, line 246: can we ensure not to call the docker version command two times ? (since it takes 2-3 seconds)

we had this CachedVersion mechanisim in the past, 

I recommend getting the version in a struct once that has everything, and we can get the fields from the struct.

something like this
https://github.com/medyagh/minikube/blob/4d631bab7e32635a90afbbe6a8677e675be7d69c/pkg/drivers/kic/oci/info.go#L47

medyagh (changes requested):

pkg/minikube/registry/drvs/docker/docker.go
spowelljr: The command only takes 133ms on my machine
```
$ time docker version --format {{.Server.Platform.Name}}
Docker Desktop 4.16.0 (95345)

real	0m0.133s
user	0m0.054s
sys	0m0.036s
```

But can definitely implement using a cached version

spowelljr: Done

pkg/minikube/registry/drvs/docker/docker.go
```diff
@@ -102,7 +102,7 @@ func status() (retState registry.State) {
 	ctx, cancel := context.WithTimeout(context.Background(), 6*time.Second)
 	defer cancel()
 
-	cmd := exec.CommandContext(ctx, oci.Docker, "version", "--format", "{{.Server.Os}}-{{.Server.Version}}")
+	cmd := exec.CommandContext(ctx, oci.Docker, "version", "--format", "{{.Server.Os}}-{{.Server.Version}}:{{.Server.Platform.Name}}")
```
medyagh, line 105: will this fail on Docker Linux without DD ?

medyagh (approved):

pkg/minikube/registry/drvs/docker/docker_test.go
```diff
@@ -135,8 +135,8 @@ func TestCheckDockerVersion(t *testing.T) {
 	}...)
 
 	for _, c := range tc {
-		t.Run("checkDockerVersion test", func(t *testing.T) {
-			s := checkDockerVersion(c.version)
+		t.Run("checkDockerEngineVersion test", func(t *testing.T) {
```
medyagh, line 138: will this test be skipped if docker is not in the path?

spowelljr: No, but this function doesn't interact with Docker, a version is passed to it and it returns the driver status based on the version.

For example, if the version passed in is older than the minimum required version, return needs update in the driver status.

pkg/minikube/registry/drvs/docker/docker.go
spowelljr: No, it returns `Docker Engine - Community` for `{{.Server.Platform.Name}}`

pkg/minikube/registry/drvs/docker/docker_test.go
medyagh: looks good to me

PR #15668, opened 2023-01-19
Updates knative minimum resource requirements

psschwei:
Signed-off-by: Paul S. Schweigert <paul@paulschweigert.com>

We currently recommend a minimum of 3 GB RAM / 3 CPUs when running Knative on minikube, so this PR updates the minikube FAQ to use those recommendations. 

(my guess is the 6 CPU / 8G ram is from when istio was a hard requirement, but that hasn't been the case for a while now)

xref: https://github.com/knative-sandbox/kn-plugin-quickstart/issues/382


spowelljr (approved):
Thanks for the doc update

Issue #15677, opened 2023-01-21
validate KicBase cache by contentDigest


x7upLime:
Currently the main difficulty we found in moving to a less docker-centric approach, 
is the the fact that the current cache-invalidation mechanism 
is based upon image digest(a.k.a. ditributionDigest, a.k.a. registry reference) verification,
and that this information is not maintained in the container image itself(kicBase cache works using tarballs)
and it seems like any container engine has its own idea on how that information should be stored/retrieved.

Since this mechanism for cache-invalidation already caused quite some headache,
and the idea of dropping this mechanism for cache-invalidation has been expressed a couple of times...
https://github.com/kubernetes/minikube/pull/15491#issuecomment-1356245842
https://github.com/google/go-containerregistry/issues/895#issuecomment-753467607

I'd propose to change the cache invalidation mechanism first, to something based on 
contentDigest(a.k.a. imageId), before trying to change the cache-access mechanism from the
kicDriver perspective. I'm sure then it'll be easier to solve a number of issues.

[+] refs:
https://windsock.io/explaining-docker-image-ids/ (contentDigest vs. distributionDigest)
https://stackoverflow.com/questions/45533005/why-digests-are-different-depend-on-registry (same as above..)
https://github.com/distribution/distribution/issues/1662#issuecomment-213079540 -- (contentDigest is content-addressable)
https://github.com/google/go-containerregistry/issues/895#issuecomment-753521526 (distributionDigest is not reliable)

afbjorklund:
The kicbase cache is just an internal thing, similar to the iso cache. And minikube can do whatever with it.

`gcr.io/k8s-minikube/kicbase:v0.0.36`

`minikube-v1.28.0-amd64.iso`

It can store the digest in some side channel, it can do sha256 on the kicbase tarball. And then verify it, itself.

```
.minikube/cache/kic/amd64/kicbase_v0.0.36@sha256_8debc1b6a335075c5f99bfbf131b4f5566f68c6500dc5991817832e55fcc9456.tar
```

---

For `kind`, the digest is needed since they include kubernetes inside the image as well as the node image.
So if you update the node image or other internals, but keep the same k8s version - you bump the digest.

kindest/node:v1.25.3@sha256:f52781bc0d7a19fb6c405c2af83abfeb311f130707a0e219175677e366cc45d1

kindest/node:v1.25.3@sha256:f1de3b0670462f43280114eccceab8bf1b9576d2afe0582f8f74529da6fd0365

[...]

PR #15491, opened 2022-12-10
KicDriver -- less docker-centric driver approach

x7upLime:
Some code refactoring to make the driver(podman/docker.. now KicDriver) interactions less docker centric.
Tries to simplify the ..DownloadkicBaseImage code in the startup process.
Adds distinction from minikube cache to kicDriver cache in comments/signatures
Adds output steps for transparency on the kicbase download process:
  -- before
> 😄  minikube v1.28.0 on Ubuntu 22.10
>    ▪ MINIKUBE_ROOTLESS=true
> ✨  Using the podman driver based on existing profile
> 👍  Starting control plane node minikube in cluster minikube
> 🚜  Pulling base image ...
> E1210 20:45:42.709990 2392963 cache.go:203] Error downloading kic artifacts:  not yet implemented, see issue #8426
> 🔥  Creating podman container (CPUs=2, Memory=8000MB) .../ ^C

   -- after
> 😄  minikube v1.28.0 on Ubuntu 22.10
> ✨  Using the docker driver based on user configuration
> 📌  Using Docker driver with root privileges
> 👍  Starting control plane node minikube in cluster minikube
> 🚜  Pulling base image to minikube cache ...
> ⌛  Loading cached image to KicDriver...
> f4462d5b2da2: Loading layer [==================================================>]  28.58MB/28.58MB
> f8c9b9e134e5: Loading layer [==================================================>]     289B/289B
> f0cc48ff0767: Loading layer [==================================================>]     287B/287B
> a0d1a19e49c4: Loading layer [==================================================>]     255B/255B
> a46401d0288f: Loading layer [==================================================>]     823B/823B
> ...
> Loaded image ID: sha256:fc27a248bc74ebfeee6e23949796c0207c7892e924b780fcd7204ed3e09ea2a3
> 💾  Downloading digest-specific layer to KicDriver...
> 🔥  Creating docker container (CPUs=2, Memory=8000MB) ...\ ^C

(fixes = "it-tries-to-fix")
fixes #13509
fixes #8426 // even if already closed..
fixes #14664

PR #15678, opened 2023-01-21
cache-invalidation based on image's contentDigest(a.k.a. imageId)

x7upLime:
In order to accomplish contentDigest-based cache-invalidation,
this proposes a very simple mechanism based on cache's filesystem:

KicDriver's stored images are checked against the cache based on contentDigest,
the contentDigest from the cached image is retrieved reading the manifest.json file inside the tarball,
the tarball is selected based on sanitized image name.

related to #15677 
solves #currently_nothing.. but would facilitate the work in accomplishing #15491 
and solving related issues..

### DIFF
--- a/pkg/drivers/kic/types.go
+++ b/pkg/drivers/kic/types.go
@@ -26,8 +26,8 @@ const (
 	// Version is the current version of kic
 	Version = "v0.0.36-1674600772-15703"
 
-	// SHA of the kic base image
-	baseImageSHA = "2d98ad015e4d4e62e5732f2fee3fee64134bacf791a9271872bbd57d47937d5f"
+	// SHA of compressed image, used as a registry reference
+	distributionDigest = "2d98ad015e4d4e62e5732f2fee3fee64134bacf791a9271872bbd57d47937d5f"
 	// The name of the GCR kicbase repository
 	gcrRepo = "gcr.io/k8s-minikube/kicbase-builds"
 	// The name of the Dockerhub kicbase repository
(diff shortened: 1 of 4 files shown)


pkg/drivers/kic/types.go
```diff
@@ -23,11 +23,13 @@ import (
 )
 
 const (
-	// Version is the current version of kic
+	// Version is the current version of kic -- also used as image tag
 	Version = "v0.0.36-1673540226-15630"
 
-	// SHA of the kic base image
-	baseImageSHA = "03c9592728381094cbd0ff9603f75ae6b485dd7a390c3e35f02ae5ec10f2f3ad"
+	// SHA of compressed image, used as a registry reference -- more reliable than a tag
```
afbjorklund, line 29: For the minikube images, there is nothing that makes the digests more reliable than the tags.

pkg/drivers/kic/types.go
```diff
@@ -23,11 +23,13 @@ import (
...
+	distributionDigest = "03c9592728381094cbd0ff9603f75ae6b485dd7a390c3e35f02ae5ec10f2f3ad"
+	// SHA of SHAS of individual image layers -- content addressable
+	ContentDigest = "243d3449b30fd2029b685cafa1191f13fbce109441e8c74001ff370d444b1927"
```
afbjorklund, line 32: You would need one such Id per architecture, the one shown here is for a single architecture.

afbjorklund (changes requested):

pkg/drivers/kic/types.go
x7upLime: I meant it in a general way: a tag can theoretically point to another (newer?)image, possibly breaking workflows.
minikube could not have this kind of problem.

pkg/drivers/kic/types.go
x7upLime: removed ContentDigest on the last commit..
as commit message points out, it is an unnecessary information here.

PR #15699, opened 2023-01-24
Adds check for volume existence in oci driver's createVolume()

### LINKED ISSUE #15697: oci.PrepareContainerNode() fails in subsequent calls with ociBin == Podman
### What Happened?

The `minikube start --driver=podman --container-runtime=crio` run,
fails reporting errors in the "🤦  StartHost failed" message, but the errors reported 
at the end of the output are deceiving:
```
❌  Exiting due to GUEST_PROVISION: Failed to start host: creating host: create: creating: setting up container node: creating volume for minikube container: podman volume create minikube --label name.minikube.sigs.k8s.io=minikube --label created_by.minikube.sigs.k8s.io=true: exit status 125
stdout:

stderr:
Error: volume with name minikube already exists: volume already exists
```
The creation of the podman volume was not the cause of the issue..   
What happens is that:
+ h.Driver.Create is called inside pkg/minikube/machine.Create
+ First call for oci.PrepareContainerNode() -- success -- inside pkg/drivers/kic.Create()
+ oci.CreateContainerNode() gets called
+ createContainer() is called -- returning no err and effectively creating container
+ fail on retry.Expo(checkRunning(p), 15*time.Millisecond, 25*time.Second)
++++++ Another attempt is made inside pkg/minikube/node.startHostInternal()
+ This time we block inside Second call of oci.PrepareContainerNode() -- returns err

```
😄  minikube v1.28.0 on Ubuntu 22.10
    ▪ MINIKUBE_ROOTLESS=true
✨  Using the podman driver based on user configuration
📌  Using rootless Podman driver
🔥  Creating podman container (CPUs=2, Memory=8000MB) ...- [*] Volume creation cmd: podman volume create minikube --label name.minikube.sigs.k8s.io=minikube --label created_by.minikube.sigs.k8s.io=true

✋  Stopping node "minikube"  ...
🔥  Deleting "minikube" in podman ...
🤦  StartHost failed, but will try again: creating host: create: creating: create kic node: container name "minikube": log: 2023-01-24T17:14:37.586373000+02:00 + grep -qw cpu /sys/fs/cgroup/cgroup.controllers
2023-01-24T17:14:37.587304000+02:00 + echo 'ERROR: UserNS: cpu controller needs to be delegated'
2023-01-24T17:14:37.587405000+02:00 ERROR: UserNS: cpu controller needs to be delegated
2023-01-24T17:14:37.587478000+02:00 + exit 1: container exited unexpectedly
🔥  Creating podman container (CPUs=2, Memory=8000MB) ...- [*] Volume creation cmd: podman volume create minikube --label name.minikube.sigs.k8s.io=minikube --label created_by.minikube.sigs.k8s.io=true

😿  Failed to start podman container. Running "minikube delete" may fix it: creating host: create: creating: setting up container node: creating volume for minikube container: podman volume create minikube --label name.minikube.sigs.k8s.io=minikube --label created_by.minikube.sigs.k8s.io=true: exit status 125
stdout:

stderr:
Error: volume with name minikube already exists: volume already exists


❌  Exiting due to GUEST_PROVISION: Failed to start host: creating host: create: creating: setting up container node: creating volume for minikube container: podman volume create minikube --label name.minikube.sigs.k8s.io=minikube --label created_by.minikube.sigs.k8s.io=true: exit status 125
stdout:

stderr:
Error: volume with name minikube already exists: volume already exists
```
##### Docker
`$ docker volume create minikube`
`minikube`
`$ docker volume create minikube`
`minikube`

##### Podman
`$ podman volume create minikube`
`minikube`
`$ podman volume create minikube`
`Error: volume with name minikube already exists: volume already exists`

### Attach the log file

.

### Operating System

Ubuntu

### Driver

Podman

x7upLime:
As the function's description states:
It should not return err or change labels if volume already exists..

fixes #15697

### DIFF
--- a/pkg/drivers/kic/oci/volumes.go
+++ b/pkg/drivers/kic/oci/volumes.go
@@ -144,15 +144,16 @@ func ExtractTarballToVolume(ociBin string, tarballPath, volumeName, imageName st
 // createVolume creates a volume to be attached to the container with correct labels and prefixes based on profile name
 // Caution ! if volume already exists does NOT return an error and will not apply the minikube labels on it.
 func createVolume(ociBin string, profile string, nodeName string) error {
-	if rr, err := runCmd(exec.Command(ociBin, "volume", "create", nodeName, "--label", fmt.Sprintf("%s=%s", ProfileLabelKey, profile), "--label", fmt.Sprintf("%s=%s", CreatedByLabelKey, "true"))); err != nil {
+	rr, err := runCmd(exec.Command(ociBin, "volume", "create", nodeName, "--label", fmt.Sprintf("%s=%s", ProfileLabelKey, profile), "--label", fmt.Sprintf("%s=%s", CreatedByLabelKey, "true")))
+	if err != nil {
 		if strings.Contains(rr.Stderr.String(), "volume already exists") && ociBin == Podman {
 			// podman doesn't fail as gracefully as docker when a volume already exists.
 			// we're reusing the same NODENAME volume:
 			return nil
 		}
-		return err
 	}
-	return nil
+
+	return err
 }
 
 // prepareVolumeSideCar will copy the initial content of the mount point by starting a temporary container to check the expected content
--- a/pkg/drivers/kic/oci/volumes.go
+++ b/pkg/drivers/kic/oci/volumes.go
@@ -143,12 +143,17 @@ func ExtractTarballToVolume(ociBin string, tarballPath, volumeName, imageName st
 
 // createVolume creates a volume to be attached to the container with correct labels and prefixes based on profile name
 // Caution ! if volume already exists does NOT return an error and will not apply the minikube labels on it.
-// TODO: this should be fixed as a part of https://github.com/kubernetes/minikube/issues/6530
 func createVolume(ociBin string, profile string, nodeName string) error {
-	if _, err := runCmd(exec.Command(ociBin, "volume", "create", nodeName, "--label", fmt.Sprintf("%s=%s", ProfileLabelKey, profile), "--label", fmt.Sprintf("%s=%s", CreatedByLabelKey, "true"))); err != nil {
-		return err
+	rr, err := runCmd(exec.Command(ociBin, "volume", "ls"))
+	if err == nil {
+		if strings.Contains(rr.Output(), nodeName) {
+			klog.Infof("Trying to create %s volume using %s: Volume already exists !", nodeName, ociBin)
+			return nil
+		}
+
+		_, err = runCmd(exec.Command(ociBin, "volume", "create", nodeName, "--label", fmt.Sprintf("%s=%s", ProfileLabelKey, profile), "--label", fmt.Sprintf("%s=%s", CreatedByLabelKey, "true")))
 	}
-	return nil
+	return err
 }
 
 // prepareVolumeSideCar will copy the initial content of the mount point by starting a temporary container to check the expected content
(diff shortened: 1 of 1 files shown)


pkg/drivers/kic/oci/volumes.go
```diff
@@ -143,12 +143,17 @@ func ExtractTarballToVolume(ociBin string, tarballPath, volumeName, imageName st
 
 // createVolume creates a volume to be attached to the container with correct labels and prefixes based on profile name
 // Caution ! if volume already exists does NOT return an error and will not apply the minikube labels on it.
-// TODO: this should be fixed as a part of https://github.com/kubernetes/minikube/issues/6530
 func createVolume(ociBin string, profile string, nodeName string) error {
-	if _, err := runCmd(exec.Command(ociBin, "volume", "create", nodeName, "--label", fmt.Sprintf("%s=%s", ProfileLabelKey, profile), "--label", fmt.Sprintf("%s=%s", CreatedByLabelKey, "true"))); err != nil {
-		return err
+	rr, err := runCmd(exec.Command(ociBin, "volume", "ls"))
```
medyagh, line 147: I dont think minikube should try to use a volume that is not created by minikube itself.
since the minikube-created volume is prefilled preload images and minikube delete also deletes that volume.(and that might have users data that was not meant to be deleted)

medyagh: this seems to be affecting the podman driver, since docker doesnt fail on recreation of a volume,

so how about we add a check for Podman Only (to avold additional calls) 

what we can do is, check for the "error" when we create and it contains "volume already exists"
and  if it is podman then we can ignore that error.

medyagh (changes requested):

pkg/drivers/kic/oci/volumes.go
x7upLime: > I dont think minikube should try to use a volume that is not created by minikube itself

Then we currently have no way to tell if the volume is minikube's or not:
The volume is not recreated during subsequent calls of docker volume create
```
$ docker volume inspect minikube 
[]
Error: No such volume: minikube

$ docker volume create minikube 
minikube

$ docker volume inspect minikube 
[
    {
        "CreatedAt": "2023-01-24T20:26:29+02:00",
        "Driver": "local",
        "Labels": {},
        "Mountpoint": "/var/lib/docker/volumes/minikube/_data",
        "Name": "minikube",
        "Options": {},
        "Scope": "local"
    }
]

$ docker volume create minikube --label name.minikube.sigs.k8s.io=minikube --label created_by.minikube.sigs.k8s.io=true
minikube

$ docker volume inspect minikube 
[
    {
        "CreatedAt": "2023-01-24T20:26:29+02:00",
        "Driver": "local",
        "Labels": {},
        "Mountpoint": "/var/lib/docker/volumes/minikube/_data",
        "Name": "minikube",
        "Options": {},
        "Scope": "local"
    }
]

## also this same examples, mounting the volume inside containers
## show that subsequent docker calls of volume create,
## doesn't change content inside the volume
```

In order to assure that the volume is minikube's, we should add an ociBin call,
and parse the  output of `ociBin volume inspect nodeName` to check for minikube labels.

x7upLime: if podman ignores the error,
it will effectively use a volume which we're not sure if it was created by minikube or not.

x7upLime: ..or I can just assure myself that a user wouldn't unintentionally use a volume named minikube, 
if not intended it to be used by minikube

PR #15701, opened 2023-01-24
Check brew install paths for socket_vmnet

spowelljr:
**Before:**
If no socket_vmnet flags were provided minikube would check the install path if socket_vmnet was installed from source.

**After:**
If no socket_vmnet flags are provided minikube checks the source, brew amd64, and brew arm64 install paths for socket_vmnet.

**Info:**
socket_vmnet_client relative path: `/opt/socket_vmnet/bin/socket_vmnet_client`
socket_vmnet relative path: `/var/run/socket_vmnet`

source install path prefix: `<none>`
arm64 brew install path prefix: `/opt/homebrew`
amd64 brew install path prefix: `/usr/local`

**Before:**
```
# socket_vmnet installed via brew

minikube start --driver qemu --network socket_vmnet

🙈  Exiting due to NOT_FOUND_SOCKET_VMNET: 


💡  Suggestion: 

    socket_vmnet was not found on the system, resolve by:
    
    Option 1) Installing socket_vmnet:
    
    https://minikube.sigs.k8s.io/docs/drivers/qemu/#networking
    
    Option 2) Using the user network:
    
    minikube start --driver qemu --network user
```

**After:**
```
# socket_vmnet installed via brew

$ minikube start --driver qemu --network socket_vmnet
🐳  Preparing Kubernetes v1.26.1 on Docker 20.10.22 ...
    ▪ Configuring RBAC rules ...
```


cmd/minikube/cmd/start_flags.go
```diff
@@ -281,8 +281,8 @@ func initNetworkingFlags() {
 	startCmd.Flags().Int(sshSSHPort, defaultSSHPort, "SSH port (ssh driver only)")
 
 	// socket vmnet
-	startCmd.Flags().String(socketVMnetClientPath, "/opt/socket_vmnet/bin/socket_vmnet_client", "Path to the socket vmnet client binary")
-	startCmd.Flags().String(socketVMnetPath, "/var/run/socket_vmnet", "Path to socket vmnet binary")
+	startCmd.Flags().String(socketVMnetClientPath, "", "Path to the socket vmnet client binary")
```
medyagh, line 284: (lets add QEMU only to the beggining of this two flags, so ppl dont read it when not relevant to them)


```Qemu driver only:....
```

pkg/minikube/detect/detect.go
```diff
@@ -140,12 +140,46 @@ func SocketVMNetInstalled() bool {
 	if runtime.GOOS != "darwin" {
 		return false
 	}
-	_, err := os.Stat(viper.GetString("socket-vmnet-path"))
+	return SocketVMNetPath() != "" && SocketVMNetClientPath() != ""
+}
+
+// SocketVMNetPath returns the path of socket_vmnet
```
medyagh, line 146: Add Qemu driver only (for code readability)

pkg/minikube/detect/detect.go
```diff
@@ -140,12 +140,46 @@ func SocketVMNetInstalled() bool {
...
+func SocketVMNetPath() string {
+	return checkSocketVMnetInstallLocations("socket-vmnet-path", "/var/run/socket_vmnet")
```
medyagh, line 148: is this correct ("socket-vmnet-path") ? or meant to be a var ?

pkg/minikube/detect/detect.go
```diff
@@ -140,12 +140,46 @@ func SocketVMNetInstalled() bool {
...
+}
+
+// SocketVMNetClientPath returns the path of socket_vmnet_client
+func SocketVMNetClientPath() string {
+	return checkSocketVMnetInstallLocations("socket-vmnet-client-path", "/opt/socket_vmnet/bin/socket_vmnet_client")
```
medyagh, line 153: same here is that string supposed to be a var?

pkg/minikube/detect/detect.go
```diff
@@ -140,12 +140,46 @@ func SocketVMNetInstalled() bool {
...
+func checkSocketVMnetInstallLocations(flagName, path string) string {
+	p := viper.GetString(flagName)
+	if p != "" {
+		return p
+	}
+	// source install, arm64 brew install, amd64 brew install
+	prefixes := []string{"", "/opt/homebrew", "/usr/local"}
+	for _, prefix := range prefixes {
+		fullPath := prefix + path
+		if fileExists(fullPath) {
```
medyagh, line 169: I swear I had seen a builtin golang tool to check if a file exists in the path

medyagh (changes requested):

pkg/minikube/detect/detect.go
spowelljr: This is correct, it's a flag key, we check if the user set their own value, and if they do just use what they provided.

pkg/minikube/detect/detect.go
spowelljr: Everything I see online just points to `os.Stat` and `os.ErrNotExist`

cmd/minikube/cmd/start_flags.go
spowelljr: We have `(Docker driver only)` at the end of a lot of our flags, think I should follow that convention to make it consistent

medyagh: sure that works, and later we can decide to change the style

pkg/minikube/detect/detect.go
medyagh: how about "exec.LookPath("...")"

spowelljr: Unfortunately, socket_vmnet doesn't get installed into the normal bin folders, so by default neither of the files we use are available on the path so `exec.LookPath` won't work.

pkg/minikube/detect/detect.go
medyagh: sounds good

medyagh (changes requested):
after adding the comments it looks good to be merged

PR #15720, opened 2023-01-27
Infer HyperKit HostIP as Gateway rather than hardcode to 192.168.64.1 

p2c2e:
fixes #11510
resolves #12729 (was closed without fix) 
Rather than assume 192.168.64.1 - which might not be true if dhcpd lease is already active for some other VM etc. 
We try to use the gateway IP is usually the hostip with the last octet as "1"

```

This PR specifically applies to MacOS users who have chosen Hyperkit (specified or since it is default). The code path does NOT affect any other combination of users.
What does the code do?
90% of the users on MacOS using Hyperkit driver will have issues with mounting a filesystem (or anything that requires access to host from within VM). Unless the user is using Minikube or similar tools first time, there is a strong chance that the IP assigned to the VM will NOT be in 192.168.64.X range. On the host, we can check all existing dhcp leases on Mac in /var/db/dhcpd_leases. For regular users, the dhcp subnet assigned by Hyperkit will be a different range. In following example, I was assigned dhcpd lease with host IP of "192.168.205.5" and gateway/nameserver will then be "192.168.205.1"
Before PR* : attempting to mount a filesystem with “minikube mount” will fail via timeout since it will be trying on some non-existent IP.
$ minikube mount /tmp:/myhosttmp --v=7
📁  Mounting host path /tmp into VM as /myhosttmp ...
    ▪ Mount type:
    ▪ User ID:      docker
    ▪ Group ID:     docker
    ▪ Version:      9p2000.L
    ▪ Message Size: 262144
    ▪ Options:      map[]
    ▪ Bind Address: 192.168.64.1:53740            <<<< PROBLEM: This IP is hardcoded and not right…..
🛑  Userspace file server is shutdown

❌  Exiting due to GUEST_MOUNT_COULD_NOT_CONNECT: /bin/bash -c "sudo mount -t 9p -o dfltgid=$(grep ^docker: /etc/group | cut -d: -f3),dfltuid=$(id -u docker),msize=262144,port=53740,trans=tcp,version=9p2000.L 192.168.64.1 /myhosttmp": Process exited with status 32
stdout:

stderr:
mount: /myhosttmp: mount(2) system call failed: Connection timed out.

After PR : attempting to mount a filesystem should just work… See below verbose log..
$ minikube mount /tmp:/myhosttmp --v=7

📁  Mounting host path /tmp into VM as /myhosttmp ...
    ▪ Mount type:   
    ▪ User ID:      docker
    ▪ Group ID:     docker
    ▪ Version:      9p2000.L
    ▪ Message Size: 262144
    ▪ Options:      map[]
    ▪ Bind Address: 192.168.205.1:53935              <<<< SOLVED : Uses correct ‘inferred’ IP. Works …
✅  Successfully mounted /tmp to /myhosttmp

We cannot 'increment' IPs etc., in this case because it HAS to be a "1" for routing/gateway to the host. Any other solution will be more complicated.

Hope this clarifies. Seems like lot of folks have this issue and this PR will help them :)
```

### DIFF
--- a/pkg/minikube/cluster/ip.go
+++ b/pkg/minikube/cluster/ip.go
@@ -128,7 +128,9 @@ func HostIP(host *host.Host, clusterName string) (net.IP, error) {
 
 		return net.ParseIP(ip), nil
 	case driver.HyperKit:
-		return net.ParseIP("192.168.64.1"), nil
+		vmIPString, _ := host.Driver.GetIP()
+		gatewayIPString := vmIPString[:strings.LastIndex(vmIPString, ".")+1] + "1"
+		return net.ParseIP(gatewayIPString), nil
 	case driver.VMware:
 		vmIPString, err := host.Driver.GetIP()
 		if err != nil {
--- a/cmd/minikube/cmd/mount.go
+++ b/cmd/minikube/cmd/mount.go
@@ -193,7 +193,7 @@ var mountCmd = &cobra.Command{
 			bindIP = "127.0.0.1"
 		}
 		out.Step(style.Mounting, "Mounting host path {{.sourcePath}} into VM as {{.destinationPath}} ...", out.V{"sourcePath": hostPath, "destinationPath": vmPath})
-		out.Infof("Mount type:   {{.name}}", out.V{"type": cfg.Type})
+		out.Infof("Mount type:   {{.name}}", out.V{"name": cfg.Type})
 		out.Infof("User ID:      {{.userID}}", out.V{"userID": cfg.UID})
 		out.Infof("Group ID:     {{.groupID}}", out.V{"groupID": cfg.GID})
 		out.Infof("Version:      {{.version}}", out.V{"version": cfg.Version})


pkg/minikube/cluster/ip.go
```diff
@@ -128,7 +128,9 @@ func HostIP(host *host.Host, clusterName string) (net.IP, error) {
 
 		return net.ParseIP(ip), nil
 	case driver.HyperKit:
-		return net.ParseIP("192.168.64.1"), nil
+		vmIPString, _ := host.Driver.GetIP()
+		gatewayIPString := vmIPString[:strings.LastIndex(vmIPString, ".")+1] + "1"
```
medyagh, line 132: Thanks for contributing to minikube, can you please put the output of minikube before/after this PR ?

and does this break normal users who currently dont have an issue ?
please comment what this code does? and how about using a library that handles incrementing the IPS instead of using string.

medyagh (changes requested):

pkg/minikube/cluster/ip.go
p2c2e: 1. This PR specifically applies to MacOS users who have chosen Hyperkit (specified or since it is default). The code path does NOT affect any other combination of users. 
2. What does the code do?
	90% of the users on MacOS using Hyperkit driver will have issues with mounting a filesystem (or anything that requires access to host from within VM). Unless the user is using Minikube or similar tools first time, there is a strong chance that the IP assigned to the VM will NOT be in 192.168.64.X range. On the host, we can check all existing dhcp leases on Mac in /var/db/dhcpd_leases. For regular users, the dhcp subnet assigned by Hyperkit will be a different range. In following example, I was assigned dhcpd lease with host IP of "192.168.205.5" and gateway/nameserver will then be "192.168.205.1"
3. **Before PR*** : attempting to mount a filesystem with “minikube mount” will fail via timeout since it will be trying on some non-existent IP.

$ minikube mount /tmp:/myhosttmp --v=7
📁  Mounting host path /tmp into VM as /myhosttmp ...
    ▪ Mount type:
    ▪ User ID:      docker
    ▪ Group ID:     docker
    ▪ Version:      9p2000.L
    ▪ Message Size: 262144
    ▪ Options:      map[]
    ▪ Bind Address: 192.168.64.1:53740            <<<< PROBLEM: This IP is hardcoded and not right…..
🛑  Userspace file server is shutdown

❌  Exiting due to GUEST_MOUNT_COULD_NOT_CONNECT: /bin/bash -c "sudo mount -t 9p -o dfltgid=$(grep ^docker: /etc/group | cut -d: -f3),dfltuid=$(id -u docker),msize=262144,port=53740,trans=tcp,version=9p2000.L 192.168.64.1 /myhosttmp": Process exited with status 32
stdout:

stderr:
mount: /myhosttmp: mount(2) system call failed: Connection timed out.


4. **After PR** : attempting to mount a filesystem should just work… See below verbose log..

$ minikube mount /tmp:/myhosttmp --v=7

📁  Mounting host path /tmp into VM as /myhosttmp ...
    ▪ Mount type:   
    ▪ User ID:      docker
    ▪ Group ID:     docker
    ▪ Version:      9p2000.L
    ▪ Message Size: 262144
    ▪ Options:      map[]
    ▪ Bind Address: 192.168.205.1:53935              <<<< SOLVED : Uses correct ‘inferred’ IP. Works …
✅  Successfully mounted /tmp to /myhosttmp

5. We cannot 'increment' IPs etc., in this case because it HAS to be a "1" for routing/gateway to the host. Any other solution will be more complicated.

Hope this clarifies. Seems like lot of folks have this issue and this PR will help them :)

p2c2e:
Hi @medyagh - Gentle nudge on this :) This will change will help lot of folks + help me avoid keeping a custom version of minikube! Thank You

p2c2e:
/cc @sharifelgamal 
/cc @prezha 

Not sure if this is the right thing to do - Trying to bump this issues - since the reviewers were removed when requesting 're-review' and this is stuck in 'Changes requested'. Thanks!

medyagh:
@p2c2e thank you for your patience on this, we have a long queue of PRs to review :) but the test look good

p2c2e:
> @p2c2e thank you for your patience on this, we have a long queue of PRs to review :) but the test look good

Thanks - I completely understand. What is next though? As per the bot messages - I see need for "lgtm" before I can assign the PR to @sharifelgamal ; ( I sorely need a 'state diagram" to visualize the flow)

Issue #15644, opened 2023-01-13
investigate if minikube service "is waiting"


medyagh:
### What Happened?

as far as I know "minikube service" command has a logic to wait for the service and pods to be running before displaying the URL to the user, 
it is possible that minikube service is not doing the right "waiting" for the service to be up before printing out the URL

resulting in  ```curl $(minikube service example --url)``` to fail

https://github.com/kubernetes/minikube/issues/13595
https://github.com/medyagh/setup-minikube/issues/29

### Attach the log file

n/a

### Operating System

None

### Driver

None

shogohida:
Hi @medyagh, can I work on this? I'm new to minikube so I might need your help while dealing with this issue.

shogohida:
Should I check the code around here?

https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/service.go#L123

elizabeth-dev:
Hi, I've been investigating on this issue, and it is true that `minikube service` is not waiting for the pods to be available. I tried looking for a solution, but since we check the k8s API for services, and there's no status info for the pods in there (not even a health check we could use as a workaround), there's no right away solution.

We could get the service `selector` and query the pods serving it, checking if there's at least one running, but that seems a bit too coupled to me, so I'd like a maintainer's opinion on that.

On a side note, while checking this issue, I noticed that we are not respecting the `--wait` argument, even though the logic is in there, so I opened a PR for that: https://github.com/kubernetes/minikube/pull/15735

Sajiyah-Salat:
I think this issue is fixed so It should be closed.

Ritikaa96:
@kubernetes/minikube-maintainers

Ritikaa96:
The PR had resolved the above issue as per me, we can take a second opinion from maintainers before it is closed though.

prezha:
@konstantinoskolios, thanks for the bump, i'm going to close ths really old issue, as we've already [confirmed](https://github.com/kubernetes/minikube/pull/15735#pullrequestreview-1293856892) it working

PR #15735, opened 2023-01-29
fix: service CLI command not honoring --wait arg

elizabeth-dev:
The minikube service <service> CLI command has a --wait argument documented that enabled the user to wait the given seconds for a service to be available in the cluster, but this was being ignored.

The logic for waiting was already implemented, but was being overriden by an unneeded early return. I removed that, and added a test case that verifies this feature is working as intended, since it was untested.

### DIFF
--- a/pkg/minikube/service/service.go
+++ b/pkg/minikube/service/service.go
@@ -261,11 +261,6 @@ func WaitForService(api libmachine.API, cname string, namespace string, service 
 		interval = 1
 	}
 
-	err := CheckService(cname, namespace, service)
-	if err != nil {
-		return nil, &SVCNotFoundError{err}
-	}
-
 	chkSVC := func() error { return CheckService(cname, namespace, service) }
 
 	if err := retry.Expo(chkSVC, time.Duration(interval)*time.Second, time.Duration(wait)*time.Second); err != nil {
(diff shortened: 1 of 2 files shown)


elizabeth-dev:
I found this while investigating issue #15644 , but it doesn't address the specific issue raised in there, so I didn't reference it in the PR. I read in the contributing docs that I should raise an issue before opening a PR, but since I already had a solution in mind, I decided to just go on with the PR. If an associated issue is really needed, I'll be happy to open it (and apologize for the caused inconveniences).

spowelljr:
Hi @elizabeth-dev, an issue is not necessary, there is a lint failure if you could fix that

medyagh:
thank you for this contribution @elizabeth-dev 
the lint says 
```
Error: pkg/minikube/service/service_test.go:926:53: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (staticcheck)
				_, _ = serviceNamespaces[test.namespace].Create(nil, &core.Service{
				          
```

elizabeth-dev:
@medyagh @spowelljr Yes, sorry, I meant to fix it, but I've been a bit busy these days and I haven't been able to open my personal laptop. I'll make the changes today.

elizabeth-dev:
@medyagh @spowelljr Fixed it, sorry for the delay

medyagh (approved):
thank you for fixing this issue, look forward to see more contributions from you

elizabeth-dev:
> thank you for fixing this issue, look forward to see more contributions from you

Thanks for the encouragement! I intend to do so in the future

Issue #8434, opened 2020-06-10
virtualbox macos: Tunnel Test doesn't work in mac os in github actions


medyagh:
example:
as seen in https://github.com/kubernetes/minikube/runs/756148873?check_suite_focus=true

the raw logs:

https://pipelines.actions.githubusercontent.com/ZGpgH9zUuTeRfFbWuBd3XFWGXuWbsKLh0sZK7k8Ux0yprYkJ4U/_apis/pipelines/1/runs/2947/signedlogcontent/39?urlExpires=2020-06-10T19%3A15%3A20.0658053Z&urlSigningMethod=HMACV1&urlSignature=qTyDyTqxuiq9GDwQUj71ei5gnVRB92Fic6ceL4upHrM%3D

sharifelgamal:
The links in the issue description has expired. I'm going to go ahead and close this for now. I'll open new issues with failing tests soon anyway.

PR #15736, opened 2023-01-29
WIP: none driver refactoring

### LINKED ISSUE #5549: Starting minikube without docker requires... docker
There is something wrong in the runtime detection of the precreate:

```console
$ sudo minikube start --vm-driver=none --container-runtime=cri-o
😄  minikube v1.4.0 on Ubuntu 16.04
🔄  Retriable failure: create: precreate: exec: "docker": executable file not found in $PATH
```

For some reason it is calling the wrong `Available` function ?

```go
func (r *Docker) Available() error {
	_, err := exec.LookPath("docker")
	return err
}
```

```go
func (r *CRIO) Available() error {
	return r.Runner.Run("command -v crio")
}
```
And the docker runtime seems to be checking _locally_ ?
(I guess us using docker machine makes it always there)

It also forgot to look for `crictl`, but that is another story.

### LINKED ISSUE #10908: None driver should take care of --container-runtime parameter

x7upLime:
For the reasons discussed in #5549, this issue has to do mainly with the (in my opinion) (*LocalClient).NewHost implementation of the [docker-machine]libmachine api interface, that wouldn't allow to pass any initialized thing, from the minikube/registry configure() methods.
Managing multiple driver-things that support multiple runtime-things through docker-machine api, feels like trying to squeeze something that has grown big, into something that has remained thin.

Even tho (*LocalClient) refactoring could definitely be a better idea than the proposed modifications, I feel like that is something that would require a thorough discussion, that I easily see taking place in a long term frame.

fixes #5549 
fixes #10908

### DIFF
--- a/pkg/drivers/none/none.go
+++ b/pkg/drivers/none/none.go
@@ -47,9 +47,10 @@ var cleanupPaths = []string{
 type Driver struct {
 	*drivers.BaseDriver
 	*pkgdrivers.CommonDriver
-	URL     string
-	runtime cruntime.Manager
-	exec    command.Runner
+	URL        string
+	runtime    cruntime.Manager
+	exec       command.Runner
+	NodeConfig Config
 }
 
 // Config is configuration for the None driver
@@ -61,24 +62,26 @@ type Config struct {
 
 // NewDriver returns a fully configured None driver
 func NewDriver(c Config) *Driver {
-	runner := command.NewExecRunner(true)
-	runtime, err := cruntime.New(cruntime.Config{Type: c.ContainerRuntime, Runner: runner})
-	// Libraries shouldn't panic, but there is no way for drivers to return error :(
-	if err != nil {
-		klog.Fatalf("unable to create container runtime: %v", err)
-	}
 	return &Driver{
 		BaseDriver: &drivers.BaseDriver{
 			MachineName: c.MachineName,
 			StorePath:   c.StorePath,
 		},
-		runtime: runtime,
-		exec:    runner,
+		NodeConfig: c,
 	}
 }
 
 // PreCreateCheck checks for correct privileges and dependencies
 func (d *Driver) PreCreateCheck() error {
+	runner := command.NewExecRunner(true)
+	runtime, err := cruntime.New(cruntime.Config{Type: d.NodeConfig.ContainerRuntime, Runner: runner})
+	if err != nil {
+		klog.Fatalf("unable to create container runtime: %v", err)
+	}
+
+	d.runtime = runtime
+	d.exec = runner
+
 	return d.runtime.Available()
 }
 


x7upLime:
Something as simple as modifying the none.(*Driver).PrecreateCheck() in the following way:

```
./pkg/drivers/none/none.go b/pkg/drivers/none/none.go
 type Driver struct {
 	*drivers.BaseDriver
 	*pkgdrivers.CommonDriver
 	URL        string
	runtime    cruntime.Manager
	exec       command.Runner
+	NodeConfig Config
 }
 
func NewDriver(c Config) *Driver {
 	if err != nil {
 		klog.Fatalf("unable to create container runtime: %v", err)
 	}
+
+	// we're gonna use this to check runtime existence
+	// in the PreCreateCheck phase
+	if c.ContainerRuntime == "cri-o" {
+		c.ContainerRuntime = "crio"
+	}
+
 	return &Driver{
 		BaseDriver: &drivers.BaseDriver{
 			MachineName: c.MachineName,
 			StorePath:   c.StorePath,
 		},
		runtime:    runtime,
		exec:       runner,
+		NodeConfig: c,
 	}
 }
 
 // PreCreateCheck checks for correct privileges and dependencies
 func (d *Driver) PreCreateCheck() error {
-	return d.runtime.Available()
+	if _, err := exec.LookPath(d.NodeConfig.ContainerRuntime); err != nil {
+		return err
+	}
+	return nil
 }
 ```

would probably work for the crio/containerd runtime..
It wouldn't do it for docker runtime:
```
// Available returns an error if it is not possible to use this runtime on a host
func (r *Docker) Available() error {
	// If Kubernetes version >= 1.24, require both cri-dockerd and dockerd.
	if r.KubernetesVersion.GTE(semver.Version{Major: 1, Minor: 24}) {
		if _, err := exec.LookPath("cri-dockerd"); err != nil {
			return err
		}
		if _, err := exec.LookPath("dockerd"); err != nil {
			return err
		}
	}
	_, err := exec.LookPath("docker")
	return err
}
```

It could be generally a better idea to leave the implementation of this Availabe() method,
to the cruntime we've initialized in the driver.

---

...or 
It would work if the code for none.Driver would include a more complex configuration,
and the none.NewDriver() would take care of that... it would be cleaner perhaps...

I can see that cruntime.New() is called in a bunch of other places; 
other than pkg/drivers/... NewDriver() functions.

**Comment by medyagh** [comment]
[x7upLime](https://github.com/x7upLime) this is Awesome you are working on this, I suggest you add a workflow in github action just to test minikube with none driver on Containerd (functional test only)

u can create a new block here
https://github.com/kubernetes/minikube/blob/6815f9b06e3078b6e64ba19cd53502f4b851f48a/.github/workflows/pr.yml#L616
and add baremetal_containerd...

**Comment by x7upLime** [comment]
@medyagh I setup actions on my fork to investigate the weird behavior of those functional tests with the none driver.
it seems there is something wrong with this test:
```
=== RUN   TestFunctional/parallel/TunnelCmd/serial/AccessDirect
panic: test timed out after 10m0s
```

I added a log to that test and reproduced on a test machine, to have an idea about what was happening:
```
// validateAccessDirect validates if the test service can be accessed with LoadBalancer IP from host
func validateAccessDirect(ctx context.Context, t *testing.T, profile string) {
	if runtime.GOOS == "windows" {
		t.Skip("skipping: access direct test is broken on windows: https://github.com/kubernetes/minikube/issues/8304")
	}
	if detect.GithubActionRunner() && runtime.GOOS == "darwin" {
		t.Skip("skipping: access direct test is broken on github actions on macos https://github.com/kubernetes/minikube/issues/8434")
	}

	checkRoutePassword(t)

	got := []byte{}
	url := fmt.Sprintf("http://%s", hostname)

	fetch := func() error {
		t.Logf("[!!] Trying to contact %s\n", url)
		h := &http.Client{Timeout: time.Second * 10}
```

##### out:
```
=== RUN   TestFunctional/parallel/TunnelCmd/serial/AccessDirect
    functional_test_tunnel_test.go:205: [!!] Trying to contact http://
    functional_test_tunnel_test.go:205: [!!] Trying to contact http://
    functional_test_tunnel_test.go:205: [!!] Trying to contact http://
    functional_test_tunnel_test.go:205: [!!] Trying to contact http://
    functional_test_tunnel_test.go:205: [!!] Trying to contact http://
    functional_test_tunnel_test.go:205: [!!] Trying to contact http://
    functional_test_tunnel_test.go:205: [!!] Trying to contact http://
    functional_test_tunnel_test.go:205: [!!] Trying to contact http://
    functional_test_tunnel_test.go:224: failed to hit nginx at "http://": Temporary Error: Get "http:": http: no Host in request URL
```

this could possibly need some further investigation..
should it be part of the same pr?
seems like a "fixing integration" pr.

PR #15774, opened 2023-02-02
Upgrade Headlamp version to 0.15.0

joaquimrocha:
Upgrade Headlamp version to 0.15.0
  cc: @joaquimrocha

### DIFF
--- a/pkg/minikube/assets/addons.go
+++ b/pkg/minikube/assets/addons.go
@@ -738,7 +738,7 @@ var Addons = map[string]*Addon{
 		MustBinAsset(addons.HeadlampAssets, "headlamp/headlamp-clusterrolebinding.yaml", vmpath.GuestAddonsDir, "headlamp-clusterrolebinding.yaml", "6040"),
 	}, false, "headlamp", "3rd party (kinvolk.io)", "", "https://minikube.sigs.k8s.io/docs/handbook/addons/headlamp/",
 		map[string]string{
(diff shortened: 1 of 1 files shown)


pkg/minikube/assets/addons.go
```diff
@@ -738,7 +738,7 @@ var Addons = map[string]*Addon{
 		MustBinAsset(addons.HeadlampAssets, "headlamp/headlamp-clusterrolebinding.yaml", vmpath.GuestAddonsDir, "headlamp-clusterrolebinding.yaml", "6040"),
 	}, false, "headlamp", "3rd party (kinvolk.io)", "", "https://minikube.sigs.k8s.io/docs/handbook/addons/headlamp/",
 		map[string]string{
-			"Headlamp": "kinvolk/headlamp:v0.14.1@sha256:162f6888461362125f4f8c60cd0a9c4031a81de15f1ed4c67bd3600b48b24029",
+			"Headlamp": "kinvolk/headlamp:v0.15.0@",
```
spowelljr, line 741: The image is missing the sha256

spowelljr (changes requested):

pkg/minikube/assets/addons.go
joaquimrocha: Thanks for spotting this. Must be a race condition in our github actions workflow. We'll look into it.

PR #15775, opened 2023-02-02
Add ingress pod to logs output

spowelljr:
- Add ingress pod to logs output
- Refactor logic to make it easy to add new addons to logs in the future
- Move existing addon pods to new logic to save checking for pod that may not exist

### DIFF
--- a/pkg/minikube/logs/logs.go
+++ b/pkg/minikube/logs/logs.go
@@ -289,6 +289,8 @@ func logCommands(r cruntime.Manager, bs bootstrapper.Bootstrapper, cfg config.Cl
 	return cmds
 }
 
+// enabledAddonPods returns the pod names for enabled addons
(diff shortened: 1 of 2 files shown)


pkg/minikube/logs/logs.go
```diff
@@ -291,3 +288,19 @@ func logCommands(r cruntime.Manager, bs bootstrapper.Bootstrapper, cfg config.Cl
 
 	return cmds
 }
+
+func enabledAddonPods(cfg config.ClusterConfig) []string {
```
medyagh, line 292: add comment, the name is not clear what does it do.

spowelljr: Added a comment

YasinAshorian (approved):

medyagh (approved):

medyagh:
merge after tests are passing

Issue #15650, opened 2023-01-15
Different echoserver images used in different places


afbjorklund:
The old `k8s.gcr.io/echoserver:1.4` was replaced with _different_ images:

1)
https://kubernetes.io/docs/tutorials/hello-minikube/

`kubectl create deployment hello-node --image=registry.k8s.io/e2e-test-images/agnhost:2.39 -- /agnhost netexec --http-port=8080`

* https://github.com/kubernetes/website/commit/25d48ec80a1e79c154e40d3ccb0255ecccf87015
* https://github.com/kubernetes/website/commit/a180cd685329268bc01f0319fe00cad517661d1c

* https://github.com/kubernetes/website/pull/37383

2)
https://minikube.sigs.k8s.io/docs/start/

`kubectl create deployment hello-minikube --image=kicbase/echo-server:1.0`

* https://github.com/kubernetes/minikube/commit/428a55fc5498fdb63cf9547d304bb2a3d1c2fc88
* https://github.com/kubernetes/minikube/commit/02061caecde19b60a8b9a597c2e5fbb736150425

* https://github.com/kubernetes/minikube/pull/15397

----

It would probably be better to have this image hosted elsewhere ?

i.e. not "registry.k8s.io/e2e-test-images", not "docker.io/kicbase"

The kubernetes.io image is slightly more enormous, than the minikube:

```
REPOSITORY                                TAG       IMAGE ID       CREATED        SIZE
registry.k8s.io/e2e-test-images/agnhost   2.39      a05bd3a9140b   7 months ago   127MB
alpine                                    3.12      24c8ece58a1a   9 months ago   5.58MB
kicbase/echo-server                       1.0       9056ab77afb8   6 months ago   4.94MB
```

The minikube image is based on https://github.com/jmalloc/echo-server

afbjorklund:
cc @spowelljr

NitishKumar06:
> It would probably be better to have this image hosted elsewhere ?
> 
> i.e. not "registry.k8s.io/e2e-test-images", not "docker.io/kicbase"
> 
> The kubernetes.io image is slightly more enormous, than the minikube:
> 
> ```
> REPOSITORY                                TAG       IMAGE ID       CREATED        SIZE
> registry.k8s.io/e2e-test-images/agnhost   2.39      a05bd3a9140b   7 months ago   127MB
> alpine                                    3.12      24c8ece58a1a   9 months ago   5.58MB
[...]

Issue #11107, opened 2021-04-15
Replace references to echoserver with supported image from all documentation


sschober:
Trying out minikube on my Mac mini apple silicon/m1. 
But step 4 from [minikube start](https://minikube.sigs.k8s.io/docs/start/) is not working for me.
 
I am using Docker Desktop 3.3.1 (63152). I have activated the experimental feature: 
"Use new virtualisation framework" 
> Enables the new Big Sur virtualization.framework instead of the hypervisor.framework. You must reset your Kubernetes cluster when enabling for the first time.

Please let me know if I can provide further information!

**Steps to reproduce the issue:** 

1. brew install minikube
2. minikube start
3. kubectl create deployment hello-minikube --image=k8s.gcr.io/echoserver:1.10

**Full output of failed command:** 
```
kubectl logs hello-minikube-5d9b964bfb-bscl4
Generating self-signed cert
Generating a 2048 bit RSA private key
.............+++
................................................................................................+++
writing new private key to '/certs/privateKey.key'
-----
Starting nginx
2021/04/15 19:56:10 [error] 14#14: failed to initialize Lua VM in /etc/nginx/nginx.conf:88
nginx: [error] failed to initialize Lua VM in /etc/nginx/nginx.conf:88
```

**Full output of `minikube start` command used, if not already included:**

😄  minikube v1.19.0 auf Darwin 11.2.3 (arm64)
✨  Automatically selected the docker driver
👍  Starting control plane node minikube in cluster minikube
🚜  Pulling base image ...
\    > gcr.io/k8s-minikube/kicbase...: 357.67 MiB / 357.67 MiB  100.00% 1.82 MiB
🔥  Creating docker container (CPUs=2, Memory=4000MB) ...
    > kubectl.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubelet.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubeadm.sha256: 64 B / 64 B [--------------------------] 100.00% ? p/s 0s
    > kubectl: 35.44 MiB / 35.44 MiB [------------] 100.00% 2.18 MiB p/s 16.44s
    > kubeadm: 34.50 MiB / 34.50 MiB [-----------] 100.00% 1.72 MiB p/s 20.229s
    > kubelet: 100.57 MiB / 100.57 MiB [---------] 100.00% 3.32 MiB p/s 30.452s

    ▪ Generating certificates and keys ...
[...]

PR #15777, opened 2023-02-02
Replace k8s.gcr.io with registry.k8s.io

mrbobbytables:
k8s.gcr.io is in the process of being deprecated with future images no longer being served from that endpoint.

This is a general "find & replace" I ignored the changelog, but updated everything else. 

For more information - see this blog post: https://kubernetes.io/blog/2022/11/28/registry-k8s-io-faster-cheaper-ga/


afbjorklund:
@mrbobbytables : the goal with the code changes, was to have the minikube cache/preload match `kubeadm`.

This is why there was a cut-off, so that we don't have to invalidate old caches and old preloads needlessly ?

* https://github.com/kubernetes/minikube/pull/14768

* https://github.com/kubernetes/minikube/pull/15546

So it is good that the examples and tests are updated, but not changing e.g. 20 was done on purpose...

Versions -1.21 use old, versions 1.25+ use new. The others have their cutoff points, 1.22.18, 1.23.15, 1.24.9

site/content/en/docs/handbook/offline.md
```diff
@@ -27,18 +27,18 @@ For offline use on other hosts, one can copy the contents of `~/.minikube/cache`
...
-cache/images/k8s.gcr.io/kube-scheduler_v1.14.0
-cache/images/k8s.gcr.io/coredns_1.3.1
-cache/images/k8s.gcr.io/kube-controller-manager_v1.14.0
-cache/images/k8s.gcr.io/kube-apiserver_v1.14.0
-cache/images/k8s.gcr.io/pause_3.1
-cache/images/k8s.gcr.io/etcd_3.3.10
-cache/images/k8s.gcr.io/kube-addon-manager_v9.0
-cache/images/k8s.gcr.io/k8s-dns-kube-dns-amd64_1.14.13
-cache/images/k8s.gcr.io/kube-proxy_v1.14.0
+cache/images/registry.k8s.io/k8s-dns-sidecar-amd64_1.14.13
```
afbjorklund, line 30: the kubernetes version needs to be updated, in this example (not changing anything for 1.14)

site/content/en/docs/contrib/tests.en.md
```diff
@@ -294,7 +294,7 @@ Steps:
 asserts basic "service" command functionality
 
 Steps:
-- Create a new `k8s.gcr.io/echoserver` deployment
+- Create a new `registry.k8s.io/echoserver` deployment
```
afbjorklund, line 297: The echoserver no longer lives in the main registry, it has been forked to multiple locations.

pkg/minikube/config/types.go
```diff
@@ -76,7 +76,7 @@ type ClusterConfig struct {
 	KubernetesConfig        KubernetesConfig
 	Nodes                   []Node
 	Addons                  map[string]bool
-	CustomAddonImages       map[string]string // Maps image names to the image to use for addons. e.g. Dashboard -> k8s.gcr.io/echoserver:1.4 makes dashboard addon use echoserver for its Dashboard deployment.
+	CustomAddonImages       map[string]string // Maps image names to the image to use for addons. e.g. Dashboard -> registry.k8s.io/echoserver:1.4 makes dashboard addon use echoserver for its Dashboard deployment.
```
afbjorklund, line 79: the echoserver image needs to be updated to the new one

pkg/minikube/bootstrapper/images/repo.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 // OldDefaultKubernetesRepo is the old default Kubernetes repository
-const OldDefaultKubernetesRepo = "k8s.gcr.io"
+const OldDefaultKubernetesRepo = "registry.k8s.io"
```
afbjorklund, line 24: Changing the old registry to be the same as the new, makes this rather meaningless.

pkg/minikube/bootstrapper/images/kubeadm_test.go
```diff
@@ -75,43 +75,43 @@ func TestKubeadmImages(t *testing.T) {
...
 		}},
 		{"v1.16.0", "", false, []string{
-			"k8s.gcr.io/kube-proxy:v1.16.0",
-			"k8s.gcr.io/kube-scheduler:v1.16.0",
-			"k8s.gcr.io/kube-controller-manager:v1.16.0",
-			"k8s.gcr.io/kube-apiserver:v1.16.0",
-			"k8s.gcr.io/coredns:1.6.2",
-			"k8s.gcr.io/etcd:3.3.15-0",
-			"k8s.gcr.io/pause:3.1",
+			"registry.k8s.io/kube-proxy:v1.16.0",
```
afbjorklund, line 108: not changing anything for 1.16 ?

afbjorklund (changes requested):

afbjorklund:
The old echoserver needs to be updated to the new location, 

* https://github.com/kubernetes/minikube/issues/11107

@spowelljr I think this will still use the "kicbase" echoserver ?

* https://github.com/kubernetes/minikube/issues/15650

site/content/en/docs/contrib/tests.en.md
afbjorklund: For minikube, I think `k8s.gcr.io/echoserver:1.4` was replaced with `kicbase/echo-server:1.0`

pkg/minikube/bootstrapper/images/repo.go
mrbobbytables: ah yeah - sorry, this was a more generic find & replace. Will update

pkg/minikube/bootstrapper/images/kubeadm_test.go
mrbobbytables: I must have accidentally just replaced that 1 line vs all in the file, will update

mrbobbytables:
@afbjorklund should I just close this? It looks like things are being tackled / tracked separately in https://github.com/kubernetes/minikube/issues/14769

FWIW - theres a policy discussion going on about agining out old images, we're sadly still on track to exceed our 3M in GCP credits (to the tune of an additional 1M 😬 ) and are looking for ways to reduce costs.

Outside of policy decisions on removing old images, one of the big things we can do is get as much user facing things to switch over to the registry.k8s.io which will spread the load across multiple providers. Updating k/website and this repo looked like an easy win because they are so frequently used / referenced by end users.

afbjorklund:
A little reminder is a good thing, it missed the 1.29.0 release for instance.

mrbobbytables:
I'll close this for now and move things over to separate PRs, and track in https://github.com/kubernetes/minikube/issues/14769

/close

afbjorklund:
@mrbobbytables if there is any other discussion on how to make the downloads smaller, please invite

I don't know if the preload bucket and release bucket are also in the same spreadsheet, somewhere...

afbjorklund:
Ok, for minikube it is mostly those two.

1) Downloading all the images into one file, instead of one-by-one with `crane` like now.
When not using a daemon, there is nobody to notice that the images are sharing layers.

```
111M	images/amd64/registry.k8s.io/etcd_3.5.6-0
17M	images/amd64/registry.k8s.io/kube-scheduler_v1.26.1
17M	images/amd64/registry.k8s.io/coredns/coredns_v1.9.3
34M	images/amd64/registry.k8s.io/kube-apiserver_v1.26.1
21M	images/amd64/registry.k8s.io/kube-proxy_v1.26.1
344K	images/amd64/registry.k8s.io/pause_3.9
31M	images/amd64/registry.k8s.io/kube-controller-manager_v1.26.1
230M	total
```

`145M	images/amd64/v1.26.1.tar.xz`

2) Downloading a compressed version of the binaries, instead of getting the executables.
Currently compression (xz) is only offered for the packages (deb, rpm) but not the exe.

```
116M	linux/amd64/v1.26.1/kubelet
46M	linux/amd64/v1.26.1/kubectl
45M	linux/amd64/v1.26.1/kubeadm
207M	total
```

`39M	linux/amd64/v1.26.1.tar.xz`

But the URLs should be changed, as well.

BenTheElder:
> Versions -1.21 use old, versions 1.25+ use new. The others have their cutoff points, 1.22.18, 1.23.15, 1.24.9

But it would only invalidate old caches? How long will minikube continue to support these old versions?
The reason it only goes back so far in Kubernetes is we can no longer release older versions.

> https://storage.googleapis.com/minikube-preloaded-volume-tarballs/

I have no idea where this is and who pays for it ... 😬 

> https://storage.googleapis.com/kubernetes-release/

This is billed to google internally / different bill that SIG K8s Infra. There's other work ongoing to migrate that out to fully community control but it's a lot so we're seeking out other funding before doing so ....

> @mrbobbytables if there is any other discussion on how to make the downloads smaller, please invite

We've had pretty active work targeting images like kube-proxy but nothing at this time. distroless kube-proxy shipped a pretty major reduction in 1.25 https://github.com/kubernetes/kubernetes/issues/109406

I think out of the top images, for current tags only ingress-nginx/controller jumps out as something that appears to be particularly large.

afbjorklund:
> > Versions -1.21 use old, versions 1.25+ use new. The others have their cutoff points, 1.22.18, 1.23.15, 1.24.9
> 
> But it would only invalidate old caches? How long will minikube continue to support these old versions? The reason it only goes back so far in Kubernetes is we can no longer release older versions.

This was just about the default in `kubeadm`, the plan is to use `registry.k8s.io` for everything

I don't see why the version support should be different in minikube, from the rest of the project...
But as long as the tutorials* are running 1.20, I guess some leeway can be "allowed" and tolerated ?
 
\* the old tutorials will soon be deleted anyway

> > https://storage.googleapis.com/minikube-preloaded-volume-tarballs/
> 
> I have no idea where this is and who pays for it ... grimacing

The preload is an optional pre-install, so I guess it is the same as the kind "node" image.

It would be possible to not provide this feature anymore, and use the regular container images.
Maybe generate the preload on the client, but then it would only speed up the **second** installation ?

> > @mrbobbytables if there is any other discussion on how to make the downloads smaller, please invite
> 
> We've had pretty active work targeting images like kube-proxy but nothing at this time. distroless kube-proxy shipped a pretty major reduction in 1.25 [kubernetes/kubernetes#109406](https://github.com/kubernetes/kubernetes/issues/109406)
> 
> I think out of the top images, for current tags only ingress-nginx/controller jumps out as something that appears to be particularly large.

The preload is all about making the install go _faster_, so it is actually bigger than the usual images. (it uses lz4)

But it would be possible to make a "batteries included" version and optimize for a small download... (using xz)

Otherwise I think we are hoping for improvements in the container runtimes and registries will "fix" the problem.
Like when allowing for peer-to-peer pulling, and so on ? Having local mirrors of the official registry also helps a lot.

PR #15782, opened 2023-02-03
Fixes mount cleaning mechanism

### LINKED ISSUE #15778: fix minikube mount --kill
### What Happened?

minikube mount --kill should kill all already mount processes spwan by minikube

to reproduce
try to mount 3 different folders
minikube mount -p minikube /tmp:/a1 
minikube mount -p minikube /tmp:/a2
minikube mount -p minikube /tmp:/a3

and then 
minikube mount --kill=true

and expect there be no process

ps aux | grep "minikube mount"

### Attach the log file

n/a

### Operating System

None

### Driver

None

x7upLime:
### MASTER:
```
$ minikube mount -p minikube /tmp:/a1 &
$ minikube mount -p minikube /tmp:/a2 &
$ minikube mount -p minikube /tmp:/a3 &

$ ps -ef | grep "minikube mount"
andrew   1357636 1241621  0 21:30 pts/5    00:00:00 minikube mount -p minikube /tmp:/a1
andrew   1357728 1241621  0 21:30 pts/5    00:00:00 minikube mount -p minikube /tmp:/a2
andrew   1357819 1241621  0 21:30 pts/5    00:00:00 minikube mount -p minikube /tmp:/a3

$ minikube mount --kill=true

$ ps -ef | grep "minikube mount"
andrew   1357636 1241621  0 21:30 pts/5    00:00:00 minikube mount -p minikube /tmp:/a1
andrew   1357728 1241621  0 21:30 pts/5    00:00:00 minikube mount -p minikube /tmp:/a2
andrew   1357819 1241621  0 21:30 pts/5    00:00:00 minikube mount -p minikube /tmp:/a3
```

### FIX:
```
$ minikube mount -p minikube /tmp:/a1 &
$ minikube mount -p minikube /tmp:/a2 &
$ minikube mount -p minikube /tmp:/a3 &

$ ps -ef | grep "minikube mount"
andrew   1360220 1241621  1 21:37 pts/5    00:00:00 minikube mount -p minikube /tmp:/a1
andrew   1360313 1241621  1 21:37 pts/5    00:00:00 minikube mount -p minikube /tmp:/a2
andrew   1360409 1241621  3 21:37 pts/5    00:00:00 minikube mount -p minikube /tmp:/a3

$ minikube mount --kill=true

$ ps -ef | grep "minikube mount"
```
fixes #15778


medyagh (commented):
Thank you for this pr, could you please add more comments on pid and how this pr adds new mechanisms to record pids

medyagh:
Let’s also add integration test

x7upLime:
> Let’s also add integration test

Done.
Added a test that ensures process removal for profile-specific mount processes

x7upLime:
> Thank you for this pr, could you please add more comments on pid and how this pr adds new mechanisms to record pids

Sure.
The killMountProcess() mechanism from master does the following
#### From master
```
// func killMountProcess() error {
  // checks the profile we're passing
	profile := viper.GetString("profile") 

  // goes to the minikube profile path
	paths := []string{  
		localpath.MiniPath(),
		localpath.Profile(profile),
	}

 // passes the profile path to killProcess
	for _, path := range paths {
		if err := killProcess(path); err != nil {
			return err

// func killProcess(path string) error {
     // looks at a ".mount-process" file
	pidPath := filepath.Join(path, constants.MountProcessFileName)

      /// reads a single pid out of it
     	out, err := os.ReadFile(pidPath)
	pid, err := strconv.Atoi(string(out))
 // The rest tries to find it and kill it, failing if something goes wrong
```

That .mount-process file gets also created by pkg/minikube/node.configureMounts:
```
// func configureMounts(wg *sync.WaitGroup, cc config.ClusterConfig) {

  // writes a single pid/file
	if err := lock.WriteFile(filepath.Join(localpath.Profile(profile), constants.MountProcessFileName), []byte(strconv.Itoa(mountCmd.Process.Pid)), 0o644); err != nil {
		exit.Error(reason.HostMountPid, "Error writing mount pid", err)
	}
```
subsequent calls of this would overwrite the previously written pid.

---

What I'm proposing is...
#### from fix-minikube_mount
```
// We write the .mount-process with this new function,
// so that we can be sure that subsequent files will not overwrite
func AppendToFile(filename string, data []byte, perm os.FileMode) error {

// Which we're using here -> pkg/minikube/cluster.Mount()
// func Mount(r mountRunner, source string, target string, c *MountConfig, pid int) error {
	if err := lock.AppendToFile(filepath.Join(localpath.Profile(profile), constants.MountProcessFileName), []byte(fmt.Sprintf("%s ", strconv.Itoa(pid))), 0o644); err != nil {
// to append a pid with trailing space [ ! ]
```
That brings us to the new mechanism:
```
// killMountProcess() stays the same

// func killProcess(path string) error {
// we read the .mount-process file
	cnt, err := os.ReadFile(pidPath)

// we 'strings.Fields()' it, so that even if there's a single pid without a trailing space,
// written by a call to lock.WriteToFile, we can still obtain a slice with one elem
	ppp := strings.Fields(string(cnt))

// then we're initializing a slice of errors
	var errs []error

// and instead of failing at each error during pid processing:
	for _, pp := range ppp {

// we still try to kill all processes, and add the reference to that err
			errs = append(errs, errors.Wrap(err, fmt.Sprintf("error parsing pid for %s", pp)))

// then at the end of pid processing we check the gathered errs references
// if only one.. we're returning it
	if len(errs) == 1 {
		return errs[0]
// if multiple errors we're showing them
// and returning a new multiple-errors error
	} else if len(errs) != 0 {
		out.Step(style.Failure, "Ecountered multiple errors:")
		for _, e := range errs {
			out.Err("%v\n", e)
		}
		return errors.New("multiple errors encountered while closing mount processes")
	}
```

cmd/minikube/cmd/delete.go
```diff
@@ -608,42 +608,64 @@ func killProcess(path string) error {
 	}
 
 	klog.Infof("Found %s ...", pidPath)
-	out, err := os.ReadFile(pidPath)
+	cnt, err := os.ReadFile(pidPath)
```
medyagh, line 611: this is a good thing that minikube will clean up after itself
lets put this logic in a helper function with a good name, like

``func killedMountedPorcesses ```

pkg/util/lock/lock.go
```diff
@@ -43,6 +43,26 @@ func WriteFile(filename string, data []byte, perm os.FileMode) error {
 	return os.WriteFile(filename, data, perm)
 }
 
+// AppendToFile
```
medyagh, line 46: comment is incomplete

test/integration/functional_test_mount_test.go
```diff
@@ -280,4 +285,90 @@ func validateMountCmd(ctx context.Context, t *testing.T, profile string) { // no
 			t.Fatalf("failed to find bind address with port 46464. Mount command out: \n%v", mountText)
 		}
 	})
+
+	t.Run("clean-mechanism", func(t *testing.T) {
```
medyagh, line 289: lets have a better name,
how about "verify-cleanup"

medyagh (changes requested):
overall looks like a good PR just minor changes requested, thank you for this PR @x7upLime

cmd/minikube/cmd/delete.go
medyagh: I suggest wither merge them into one func or make an addtional helper called , additionalMountCleanUp or anything reallly, the goal is to make the code readable and modular

x7upLime:
I clicked on something I definitely shouldn't have..

tools/pleasedontkillme/main.go
```diff
@@ -0,0 +1,25 @@
+package main
```
medyagh, line 17: Thanks for adding this interesting test, I think we should come up with a better name, maybe "not_ending_proc" and and lets choose a better  path for it than the root folder of minikube
we generally wanna keep the root folder clean and make it more oragnized.

, how about consider putting it under "test" folder or maybe "hack folder"

medyagh (changes requested):

tools/pleasedontkillme/main.go
x7upLime: There you go :)

I wasn't sure because there was no such thing in the repo..
so I tried [this](https://github.com/golang-standards/project-layout#tools) and hoped for the best

medyagh:
I think you forgot to commit the chnage to move 
tools/pleasedontkillme/main.go 

and also lets use a better name, i dont like violent verbs to be seen in our folder structure, 
How about something like
Hack/cleanuptest
Test/tools/cleanup_test

x7upLime:
I removed the /tools folder and put the source code for the tool inside the unit test itself :)
There are no more references to tools/pleasedontkillme/main.go in my fork.

I changed the name of the variable from pleasedontkillme to waitforsig

medyagh (changes requested):

cmd/minikube/cmd/delete.go
```diff
@@ -601,49 +602,99 @@ func killMountProcess() error {
...
+		}
+
+	}
+
+	if len(errs) == 1 {
+		// if we've encountered only one error, we're returning it:
+		return errs[0]
+	} else if len(errs) != 0 {
+		// if multiple errors were encountered, we're outputting them
+		// and returning a new multiple-errors error:
```
spowelljr, line 636:
```suggestion
		// if multiple errors were encountered, combine them into a single error
```

cmd/minikube/cmd/delete.go
```diff
@@ -601,49 +602,99 @@ func killMountProcess() error {
...
+		out.Step(style.Failure, "Ecountered multiple errors:")
```
spowelljr, line 637:
```suggestion
		out.Styled(style.Failure, "Multiple errors encountered:")
```
`out.Step` should only be used when a JSON step has been setup for the action.

cmd/minikube/cmd/delete.go
```diff
@@ -601,49 +602,99 @@ func killMountProcess() error {
...
 	if err != nil {
-		return errors.Wrap(err, "ReadFile")
+		return err
 	}
-	klog.Infof("pidfile contents: %s", out)
-	pid, err := strconv.Atoi(string(out))
-	if err != nil {
-		return errors.Wrap(err, "error parsing pid")
+
+	// we're giving a change to each pid, to be killed
```
spowelljr, line 621: I'm not sure what this comment means, could you explain it?

cmd/minikube/cmd/delete.go
```diff
@@ -601,49 +602,99 @@ func killMountProcess() error {
...
+		// and returning a new multiple-errors error:
+		out.Step(style.Failure, "Ecountered multiple errors:")
+		for _, e := range errs {
+			out.Err("%v\n", e)
+		}
+		return errors.New("multiple errors encountered while closing mount processes")
 	}
-	// os.FindProcess does not check if pid is running :(
+
+	// if no errors were encoutered, it is safe to clean out the pidFile
```
spowelljr, line 644:
```suggestion
	// if no errors were encoutered, it's safe to delete pidFile
```

cmd/minikube/cmd/delete.go
```diff
@@ -601,49 +602,99 @@ func killMountProcess() error {
...
+	if err := os.Remove(pidPath); err != nil {
+		return errors.Wrap(err, "While closing mount-pids file")
+	}
+
+	return err
+}
+
+// tryKillOne takes a PID as argument and tries to SIGKILL it
+func tryKillOne(pid int) error {
```
spowelljr, line 653: This name is a bit odd, I'd recommend something like `killProcess` maybe?

cmd/minikube/cmd/delete_test.go
```diff
@@ -220,3 +222,60 @@ func TestDeleteAllProfiles(t *testing.T) {
...
+}
+`)
+	td := t.TempDir()
+	tmpfile := filepath.Join(td, "waitForSig.go")
+
+	if err := os.WriteFile(tmpfile, waitForSig, 0o700); err != nil {
+		t.Fatalf("copying source to %s", tmpfile)
+	}
+
+	dontkillmeProc := exec.Command("go", "run", tmpfile)
```
spowelljr, line 265:
```suggestion
	processToKill := exec.Command("go", "run", tmpfile)
```
Variable name doesn't follow camel casing, also the variable name is confusing

test/integration/functional_test_mount_test.go
```diff
@@ -280,4 +284,95 @@ func validateMountCmd(ctx context.Context, t *testing.T, profile string) { // no
 			t.Fatalf("failed to find bind address with port 46464. Mount command out: \n%v", mountText)
 		}
 	})
+
+	t.Run("verify-cleanup", func(t *testing.T) {
+		tempDir := t.TempDir()
+
+		ctx, cancel := context.WithTimeout(ctx, Minutes(10))
+
+		inNodeMounts := []string{"/mount1", "/mount2", "/mount3"}
```
spowelljr, line 293:
```suggestion
		guestMountPaths := []string{"/mount1", "/mount2", "/mount3"}
```

test/integration/functional_test_mount_test.go
```diff
@@ -280,4 +284,95 @@ func validateMountCmd(ctx context.Context, t *testing.T, profile string) { // no
...
+
+		// are the mounts alive yet..?
+		checkMount := func() error {
+			for _, mnt := range inNodeMounts {
+				rr, err := Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "findmnt -T", mnt))
+				if err != nil {
+					// if something weird has happened from previous tests..
+					// this could at least spare us some waiting
+					if strings.Contains(rr.Stdout.String(), fmt.Sprintf("Profile \"%s\" not found.", profile)) {
+						t.Fatalf("somehow the profile got cancelled; failing....")
```
spowelljr, line 330:
```suggestion
						t.Fatalf("profile was deleted, cancelling the test")
```

test/integration/functional_test_mount_test.go
```diff
@@ -280,4 +284,95 @@ func validateMountCmd(ctx context.Context, t *testing.T, profile string) { // no
...
+				t.Skip("skipping: mount did not appear, likely because macOS requires prompt to allow non-codesigned binaries to listen on non-localhost port")
+			}
+			t.Fatalf("mount was not ready in time: %v", err)
+		}
+
+		checkProcsAlive := func(end chan bool) {
+			for _, mntp := range mntProcs {
+				// Trying to wait for process end
+				// if the wait fail with ExitError we know that the process
+				// doesnt' exist anymore..
```
spowelljr, line 349:
```suggestion
				// doesn't exist anymore..
```

cmd/minikube/cmd/delete.go
```diff
@@ -584,7 +584,8 @@ func deleteMachineDirectories(cc *config.ClusterConfig) {
 	}
 }
 
-// killMountProcess kills the mount process, if it is running
+// killMountProcess looks for the legacy path and for profile path for a pidfile,
+// it then tries to kill all the pids listed in the pidfile(one or more)
```
spowelljr, line 588:
```suggestion
// it then tries to kill all the pids listed in the pidfile (one or more)
```

cmd/minikube/cmd/delete.go
```diff
@@ -601,49 +602,99 @@ func killMountProcess() error {
 	return nil
 }
 
+// killProcess takes a path to look for a pidfile(space-separated),
```
spowelljr, line 605:
```suggestion
// killProcess takes a path to look for a pidfile (space-separated),
```

spowelljr (changes requested):

cmd/minikube/cmd/delete.go
x7upLime: We're not stopping on the first error found while trying to kill mount processes.
Instead we're trying to send the signal to all mount processes, capturing eventual errors; then doing error handling logic

cmd/minikube/cmd/delete.go
x7upLime: Unfortunately this is a helper function of the killProcess function :/
What about something like `trySignalProcess`?

cmd/minikube/cmd/delete.go
x7upLime: oh.. it was supposed to be `chance`, not `change`. sorry

I modified it.. now it should be more readable

medyagh:
@x7upLime there seems to be a build failure for the code, I recommend running "make clean;make" on your end to ensure the code is building

spowelljr:
> @x7upLime there seems to be a build failure for the code, I recommend running "make clean;make" on your end to ensure the code is building

It looks like it failed on `gsutil`, I'll try running it again

cmd/minikube/cmd/delete.go
```diff
@@ -601,49 +602,99 @@ func killMountProcess() error {
...
+	// if no errors were encoutered, it's safe to delete pidFile
+	if err := os.Remove(pidPath); err != nil {
+		return errors.Wrap(err, "While closing mount-pids file")
+	}
+
+	return err
+}
+
+// trySignalProcess takes a PID as argument and tries to SIGKILL it
+func trySignalProcess(pid int) error {
```
medyagh, line 653: rename to trySigKillProcess()

cmd/minikube/cmd/delete.go
```diff
@@ -601,49 +602,99 @@ func killMountProcess() error {
...
 	if entry == nil {
 		klog.Infof("Stale pid: %d", pid)
-		if err := os.Remove(pidPath); err != nil {
-			return errors.Wrap(err, "Removing stale pid")
-		}
-		return nil
+		return errors.Wrap(err, fmt.Sprintf("removing stale pid: %d", pid))
 	}
 
 	// We found a process, but it still may not be ours.
```
medyagh, line 663: here can we ensure the PID's binary contains the name minikube ?
in normal "ps aux" in terminal we can see the PID have minikube binary there is that campability, I bet go-ps lets us do that too.

we can move that logic to a func called

"doesPIDBelongToMinikube"
and it should ensure that PID 's binary path has word minikube in it

medyagh (changes requested):

medyagh:
hi @x7upLime  are you still working on this PR

x7upLime:
@medyagh sure. I'll be back on it as soon as possible(most probably tomorrow)

medyagh:
@x7upLime this test is failing https://storage.googleapis.com/minikube-builds/logs/15782/28211/Docker_Linux.html#fail_TestFunctional%2fparallel%2fMountCmd%2fverify-cleanup

x7upLime:
@medyagh I promised myself I'd investigate further the next day 😬
Now it seems it's working again...

spowelljr:
cleanup minikube-bot comments

medyagh:
here is the possible relavant failed test on KVM
https://storage.googleapis.com/minikube-builds/logs/15782/28319/KVM_Linux.html#fail_TestMountStart%2fserial%2fStop

x7upLime:
the tests were failing because the `minikube start -p someprof --mount` would trigger a .mount-process writer that would not insert a whitespace between pids, instead of the one in pkg/minikube/cluster/mount.go, which has already been taken care of (responsible for the `minikube mount ...`)

I also ran `make generate-docs` to solve the lint error.

x7upLime:
there was an annoying error in unit tests caused by the ownership check for the pid we're trying to send the signal to.
I saw the quickest road in adding the check as a dependency (see commit msg).

cmd/minikube/cmd/delete.go
```diff
@@ -601,49 +602,120 @@ func killMountProcess() error {
...
+		return errors.Wrap(err, "While closing mount-pids file")
+	}
+
+	return err
+}
+
+// trySigKillProcess takes a PID as argument and tries to SIGKILL it.
+// It performs the OCHECK ownership check of the pid,
+// before trying to send a sigkill signal to it
+func trySigKillProcess(pid int, ocheck func(int) (bool, error)) error {
```
spowelljr, line 655: I see you're trying to mock `doesPIDBelongToMinikube`.

What you can do instead is:
```
var doesPIDBelongToMinikube = func(pid int) (bool, error) {
```

Then during the unit test you can do
```
doesPIDBelongToMinikube = func(pid int) (bool, error) { return true, nil }
```

spowelljr (changes requested):

cmd/minikube/cmd/delete.go
x7upLime: got it, thanks :)

cmd/minikube/cmd/delete_test.go
```diff
@@ -220,3 +222,64 @@ func TestDeleteAllProfiles(t *testing.T) {
...
+		close(done)
+	}()
+
+	<-done
+}
+`)
+	td := t.TempDir()
+	tmpfile := filepath.Join(td, "waitForSig.go")
+
+	if err := os.WriteFile(tmpfile, waitForSig, 0o700); err != nil {
```
spowelljr, line 261:
```suggestion
	if err := os.WriteFile(tmpfile, waitForSig, 0o600); err != nil {
```
The go file shouldn't need to be executable

cmd/minikube/cmd/delete_test.go
```diff
@@ -220,3 +222,64 @@ func TestDeleteAllProfiles(t *testing.T) {
...
+	}
+	pid := processToKill.Process.Pid
+
+	doesPIDBelongToMinikube = func(int) (bool, error) {
+		return true, nil
+	}
+
+	err = trySigKillProcess(pid)
+	if err != nil {
+		t.Fatalf("while trying to kill child proc %d: %s\n", pid, err)
```
spowelljr, line 278:
```suggestion
		t.Fatalf("while trying to kill child proc %d: %v\n", pid, err)
```

cmd/minikube/cmd/delete_test.go
```diff
@@ -220,3 +222,64 @@ func TestDeleteAllProfiles(t *testing.T) {
...
+	}()
+
+	<-done
+}
+`)
+	td := t.TempDir()
+	tmpfile := filepath.Join(td, "waitForSig.go")
+
+	if err := os.WriteFile(tmpfile, waitForSig, 0o700); err != nil {
+		t.Fatalf("copying source to %s: %s\n", tmpfile, err)
```
spowelljr, line 262:
```suggestion
		t.Fatalf("copying source to %s: %v\n", tmpfile, err)
```

cmd/minikube/cmd/delete_test.go
```diff
@@ -220,3 +222,64 @@ func TestDeleteAllProfiles(t *testing.T) {
...
+	}
+
+	processToKill := exec.Command("go", "run", tmpfile)
+	err := processToKill.Start()
+	if err != nil {
+		t.Fatalf("while execing child process: %s\n", err)
```
spowelljr, line 268:
```suggestion
		t.Fatalf("while execing child process: %v\n", err)
```

cmd/minikube/cmd/delete_test.go
```diff
@@ -220,3 +222,64 @@ func TestDeleteAllProfiles(t *testing.T) {
...
+	}
+
+	err = trySigKillProcess(pid)
+	if err != nil {
+		t.Fatalf("while trying to kill child proc %d: %s\n", pid, err)
+	}
+
+	// waiting for process to exit
+	if err := processToKill.Wait(); !strings.Contains(err.Error(), "killed") {
+		t.Fatalf("unable to kill process: %s\n", err)
```
spowelljr, line 283:
```suggestion
		t.Fatalf("unable to kill process: %v\n", err)
```

test/integration/functional_test_mount_test.go
```diff
@@ -280,4 +284,95 @@ func validateMountCmd(ctx context.Context, t *testing.T, profile string) { // no
 			t.Fatalf("failed to find bind address with port 46464. Mount command out: \n%v", mountText)
 		}
 	})
+
+	t.Run("verify-cleanup", func(t *testing.T) {
```
spowelljr, line 288:
```suggestion
	t.Run("VerifyCleanup", func(t *testing.T) {
```
We're using pascal case for test names, we still have some legacy test names using the old way

spowelljr (changes requested):
Just some small things

spowelljr (changes requested):
I just tried this and one thing I noticed is that the PID is not removed from the pid file if the mount is terminated. Then when the user goes to delete the profile they get:
```
❌  Failed to kill mount process: stale pid: 47622
```

If a user were to have numerous mounts over the lifetime of the cluster we'll try killing PIDs we already know aren't going to exist, and the numerous error outputs could be confusing to the user.

We're able to detect that a mount has been killed:
```
✅  Successfully mounted /Users/powellsteven to /thing5

^C🔥  Unmounting /thing5 ...

❌  Exiting due to MK_INTERRUPTED: Received interrupt signal
```

So I propose that we add removing the PID from the file when we detect the mount process was killed.

x7upLime:
the last commit is just duplicated logic from the one used in the delete command...
I'm thinking we could move it inside the pkg/util and use it from there..

cmd/minikube/cmd/mount.go
```diff
@@ -266,3 +279,58 @@ func getPort() (int, error) {
...
+		// we skip the pid we're looking for
+		if intPid == pid {
+			continue
+		}
+
+		pids = append(pids, intPid)
+	}
+
+	// we convert the pids list back to string and write it back to file
+	newPids := fmt.Sprintf("%s ", strings.Trim(strings.Join(strings.Fields(fmt.Sprint(pids)), " "), "[]"))
```
spowelljr, line 334: This line is a bit complex, currently this func takes in an int and then convert everything to an int to compare, but I think it would make everything easier if you convert the incoming pid to a string. Then `strconv.Atoi(p)` could be removed, and this could change into:

```suggestion
	newPids := strings.Join(pids, " ")
```

I'll include a couple code suggestions to hopefully make more sense of it

cmd/minikube/cmd/mount.go
```diff
@@ -266,3 +279,58 @@ func getPort() (int, error) {
...
+	}
+
+	// we found the correct file
+	// we're reading the pids...
+	out, err := os.ReadFile(pidPath)
+	if err != nil {
+		return errors.Wrap(err, "ReadFile")
+	}
+
+	pids := []int{}
```
spowelljr, line 317:
```suggestion
	pids := []string{}
```

cmd/minikube/cmd/mount.go
```diff
@@ -266,3 +279,58 @@ func getPort() (int, error) {
...
+	strPids := strings.Fields(string(out))
+	for _, p := range strPids {
+		intPid, err := strconv.Atoi(p)
+		if err != nil {
+			return errors.Wrap(err, "while converting pids")
+		}
```
spowelljr, line 323:
```suggestion
```

cmd/minikube/cmd/mount.go
```diff
@@ -266,3 +279,58 @@ func getPort() (int, error) {
...
+
+		// we skip the pid we're looking for
+		if intPid == pid {
```
spowelljr, line 326:
```suggestion
		if p == pid {
```

cmd/minikube/cmd/mount.go
```diff
@@ -266,3 +279,58 @@ func getPort() (int, error) {
...
+			continue
+		}
+
+		pids = append(pids, intPid)
```
spowelljr, line 330:
```suggestion
		pids = append(pids, p)
```

spowelljr (changes requested):

cmd/minikube/cmd/mount.go
```diff
@@ -266,3 +279,56 @@ func getPort() (int, error) {
 	defer l.Close()
 	return l.Addr().(*net.TCPAddr).Port, nil
 }
+
+// removePidFromFile looks at the default locations for the mount-pids file,
+// for the profile in use. If a file is found and its content shows PID, PID gets removed.
+func removePidFromFile(pid int) error {
+	profile := viper.GetString("profile")
```
spowelljr, line 286:
```suggestion
	profile := ClusterFlagValue()
```

spowelljr (changes requested):

spowelljr (approved):

PR #15784, opened 2023-02-05
Look for cri-dockerd instead of hardcoding

### LINKED ISSUE #15265: cri-dockerd system unit override assumes `/usr/bin/cri-dockerd` but install instructions from mirantis point to `/usr/local/bin/cri-dockerd`
### What Happened?

The [template for cri-dockerd systemd override](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/cruntime/docker.go#L700) uses a hard coded path:

```golang
	var CRIDockerServiceConfTemplate = template.Must(template.New("criDockerServiceConfTemplate").Parse(`[Service]
ExecStart=
ExecStart=/usr/bin/cri-dockerd --container-runtime-endpoint fd:// --network-plugin={{.NetworkPlugin}}{{.ExtraArguments}}`))
```

Which differs from the install location suggested by the [cri-dockerd installation instructions](https://github.com/Mirantis/cri-dockerd#to-use-with-kubernetes), specifically, this part:

```bash
install -o root -g root -m 0755 bin/cri-dockerd /usr/local/bin/cri-dockerd
cp -a packaging/systemd/* /etc/systemd/system
sed -i -e 's,/usr/bin/cri-dockerd,/usr/local/bin/cri-dockerd,' /etc/systemd/system/cri-docker.service
```

### Attach the log file

```text
Nov 01 16:51:50 MM269697-PC systemd[4939]: cri-docker.service: Failed at step EXEC spawning /usr/bin/cri-dockerd: No such file or directory
```

### Operating System

Redhat/Fedora

### Driver

None (Baremetal)

afbjorklund:
Some people install the daemon in /usr/local/bin,
rather than under /usr/bin as the unit expects.

Closes #15265

### DIFF
--- a/pkg/minikube/cruntime/docker.go
+++ b/pkg/minikube/cruntime/docker.go
@@ -688,6 +688,14 @@ const (
 	CNICacheDir = "/var/lib/cni/cache"
 )
 
+func getCriDockerdPath(cr CommandRunner) string {
+	rr, err := cr.RunCmd(exec.Command("which", "cri-dockerd"))
+	if err != nil {
+		return "/usr/bin/cri-dockerd"
+	}
+	return strings.TrimSuffix(rr.Stdout.String(), "\n")
+}
+
 func dockerConfigureNetworkPlugin(cr CommandRunner, networkPlugin string) error {
 	// $ cri-dockerd --version
 	// cri-dockerd 0.2.6 (d8accf7)
@@ -702,9 +710,11 @@ func dockerConfigureNetworkPlugin(cr CommandRunner, networkPlugin string) error 
 		networkPlugin = "cni"
 	}
 	opts := struct {
+		ExecPath       string
 		NetworkPlugin  string
 		ExtraArguments string
 	}{
+		ExecPath:       getCriDockerdPath(cr),
 		NetworkPlugin:  networkPlugin,
 		ExtraArguments: args,
 	}
@@ -712,7 +722,7 @@ func dockerConfigureNetworkPlugin(cr CommandRunner, networkPlugin string) error 
 	const CRIDockerServiceConfFile = "/etc/systemd/system/cri-docker.service.d/10-cni.conf"
 	var CRIDockerServiceConfTemplate = template.Must(template.New("criDockerServiceConfTemplate").Parse(`[Service]
 ExecStart=
-ExecStart=/usr/bin/cri-dockerd --container-runtime-endpoint fd:// --network-plugin={{.NetworkPlugin}}{{.ExtraArguments}}`))
+ExecStart={{.ExecPath}} --container-runtime-endpoint fd:// --network-plugin={{.NetworkPlugin}}{{.ExtraArguments}}`))
 
 	b := bytes.Buffer{}
 	if err := CRIDockerServiceConfTemplate.Execute(&b, opts); err != nil {


pkg/minikube/cruntime/docker.go
```diff
@@ -688,6 +688,14 @@ const (
 	CNICacheDir = "/var/lib/cni/cache"
 )
 
+func getCriDockerdPath(cr CommandRunner) string {
+        rr, err := cr.RunCmd(exec.Command("which", "cri-dockerd"))
```
neersighted, line 692: `which` is not a standard command, and parsing its output is dangerous at best. The stdlib [`LookPath`](https://pkg.go.dev/os/exec#LookPath) is designed for this use-case.

pkg/minikube/cruntime/docker.go
```diff
@@ -688,6 +688,14 @@ const (
...
+        if err != nil {
+                return "/usr/bin/cri-dockerd"
```
neersighted, line 694: It's probably better to return an error fail early, where we can describe what happened, rather than in the systemd unit...

neersighted (changes requested):

pkg/minikube/cruntime/docker.go
afbjorklund: How do you run LookPath over ssh ? We could run `command -v`, at least for the recently modern ones

pkg/minikube/cruntime/docker.go
afbjorklund: I think we already looked for `cri-dockerd` in the user-facing command code, so this is not supposed to fail (at all)

pkg/minikube/cruntime/docker.go
neersighted: Hmm, `command -v` is at least POSIX. I suppose the remote use-case makes this harder; but at the same time it feels very impure. The ability to `go run` over SSH might be nice in the end, but that's a lot of machinery to build out.

Does it make sense to abstract this a layer deeper and write a `cr.LookPath` that uses the external binary or `os.LookPath` based on the target?

pkg/minikube/cruntime/docker.go
neersighted: In that case, it seems better for everyone's collective sanity that an error is returned here, and that it does tear down the call stack/bubble up, as otherwise if the prelude to this code being invoked changes (and these assumptions are not obvious), severe bit-rot/surprising effects could ensue...

pkg/minikube/cruntime/docker.go
afbjorklund: Sure, but that is more of a separate refactor

medyagh (approved):

shu-mutou:
Is it possible that what you look for with `which` is different from what is written in the systemd file?

shu-mutou:
The current minikube code creates a `10-cni.conf` file as one of the settings for the installed systemd file, so it would be better to write the same as the existing systemd file, IMHO.

neersighted:
There is a possibility of such, especially as the user (and not minikube) controls who you execute as and thus the content of the remote `$PATH`. That being said, I think this is much more sane than trying to parse a systemd unit, and if the user has an incorrect cri-dockerd in the path, minikube using it is not necessarily surprising. The results will be the same as `sudo $(command -v cri-dockerd)`; if that gives unexpected results, then there is something more wrong with the system.

afbjorklund:
Note that `sudo` does **not** include `/usr/local/bin`, on older systems. It failed `crictl`, earlier.

pkg/minikube/cruntime/docker.go
afbjorklund: Well, it was not _worse_ (than before)

medyagh: @afbjorklund how about returning the current default value in case of "which" failing, 
currently it assumes  "/usr/bin/cri-dockerd" in case "which" fails we dont even try the old way.

medyagh (changes requested):
@afbjorklund sorry I changed my review

shu-mutou:
@neersighted Thanks for your reply.

I've only used minikube with the none driver, so I don't really understand how it works with other drivers, but my understanding is that if cri-docker.service doesn't work, minikube won't work either. right?
If the path written in the existing systemd file is wrong, cri-docker.service will not work in the first place, so I think that's the cri-dockerd installation guide or the user's responsibility.

neersighted:
The issue here is that cri-dockerd.service does work. Then minikube attempts to change the command line to set the CNI, and we end up here. This PR represents an attempt to prevent minikube from having strong opinions about what path cri-dockerd is available at.

shu-mutou:
@neersighted Thank you for your explanations!
I understood the intention of minikube team and respect it.

pkg/minikube/cruntime/docker.go
afbjorklund: That is the default.

medyagh (approved):

PR #15831, opened 2023-02-11
Add ability to get json for "minikube service list"

### LINKED ISSUE #15827: add ability to get json for "minikube sevice list"
we already have json output for many of minikube commands, for example 
"minikube profile list -o=json" 

we have forgotten to implement it for "service list"

"minikube service list -o=json"

should be easy to follow other examples of doing it

OmSaran:
fixes #15827 

Before:
```
$ minikube service list -o=json
Error: unknown shorthand flag: 'o' in -o=json
See 'minikube service list --help' for usage.
```

After:
```
$ ./out/minikube service list -o=json
[{"Namespace":"default","Name":"kubernetes","URLs":[],"PortNames":[]},{"Namespace":"kube-system","Name":"kube-dns","URLs":[],"PortNames":[]}]
```


cmd/minikube/cmd/service_list.go
```diff
@@ -48,28 +51,43 @@ var serviceListCmd = &cobra.Command{
...
-				data = append(data, []string{serviceURL.Namespace, serviceURL.Name, "No node port"})
-			} else {
-				servicePortNames := strings.Join(serviceURL.PortNames, "\n")
-				serviceURLs := strings.Join(serviceURL.URLs, "\n")
+		switch output {
+		case "table":
+			printServicesTable(serviceURLs, co)
+		case "json":
+			printServicesJSON(serviceURLs)
+		}
```
spowelljr, line 59: There should be a default case that outputs something like:
```
exit.Message(reason.Usage, fmt.Sprintf("invalid output format: %s. Valid values: 'table', 'json'", output))
```

spowelljr (changes requested):
Could you also add a simple test to
https://github.com/kubernetes/minikube/blob/89dbf519e84a5b216e5068ac89f84a7a56aeb921/test/integration/functional_test.go#L1396-L1397

Similar to
https://github.com/kubernetes/minikube/blob/89dbf519e84a5b216e5068ac89f84a7a56aeb921/test/integration/functional_test.go#L1337-L1338

medyagh:
@OmSaran do you have any questions on how to add integration test?

cmd/minikube/cmd/service_list.go
OmSaran: Thanks for catching this, fixed it!

OmSaran:
> Could you also add a simple test to
> 
> https://github.com/kubernetes/minikube/blob/89dbf519e84a5b216e5068ac89f84a7a56aeb921/test/integration/functional_test.go#L1396-L1397
> 
> Similar to
> 
> https://github.com/kubernetes/minikube/blob/89dbf519e84a5b216e5068ac89f84a7a56aeb921/test/integration/functional_test.go#L1337-L1338

Done! Please take a look

OmSaran:
> @OmSaran do you have any questions on how to add integration test?

I took a stab at it, let me know if this looks okay!

medyagh:
@OmSaran please check the failed test https://github.com/kubernetes/minikube/actions/runs/4188409014/jobs/7262233741

to be clear we dont really need to do a full Service test again there (such as deploying a new service) we could just ensure the service list -o=json is a valid json and does have the service we expect there in json

medyagh:
@OmSaran you can run the functional tests on your own machine using "make functional"

OmSaran:
> @OmSaran you can run the functional tests on your own machine using "make functional"

This time I tested it using a linux machine and it was successful. Thanks!
```
<truncated>
--- PASS: TestFunctional/parallel/ServiceCmd (30.40s)
            --- PASS: TestFunctional/parallel/ServiceCmd/service_json_output (2.33s)
<truncated>
```


Previously it was not working on my local machine (M1 Mac) as it was getting hung in `ServiceCmdConnect` test (so I had a few hacks to test it which included testing it with hardcoded profile and thus didn't catch the failure scenario in the build).

test/integration/functional_test.go
```diff
@@ -1520,6 +1523,42 @@ func validateServiceCmd(ctx context.Context, t *testing.T, profile string) {
 	}
 }
 
+func validateServiceCmdJSON(ctx context.Context, t *testing.T, profile string) {
+	// docs: Run `minikube service list -o JSON` and make sure the services are correctly listed as JSON output
+	t.Run("service_json_output", func(t *testing.T) {
```
spowelljr, line 1528:
```suggestion
	t.Run("ServiceJSONOutput", func(t *testing.T) {
```
Test names should follow pascal case

test/integration/functional_test.go
```diff
@@ -1520,6 +1523,42 @@ func validateServiceCmd(ctx context.Context, t *testing.T, profile string) {
...
+		targetSvcName := "hello-node"
+		// helper function to run command then, return target service object from json output.
+		extractServiceObjFunc := func(rr *RunResult) *service.SvcURL {
+			var jsonObjects service.URLs
+			err := json.Unmarshal(rr.Stdout.Bytes(), &jsonObjects)
+			if err != nil {
+				t.Errorf("failed to decode json from profile list: args %q: %v", rr.Command(), err)
+				return nil
```
spowelljr, line 1536:
```suggestion
				t.Fatalf("failed to decode json from profile list: args %q: %v", rr.Command(), err)
```
Using `Fatalf` will remove the need for `return` plus it's pointless to try completing the test if we can't unmarshal the output anyway so `Fatalf` makes sense

test/integration/functional_test.go
```diff
@@ -1520,6 +1523,42 @@ func validateServiceCmd(ctx context.Context, t *testing.T, profile string) {
...
+				}
+			}
+			return nil
+		}
+
+		start := time.Now()
+		rr, err := Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "service", "list", "-o", "json"))
+		elapsed := time.Since(start)
+		if err != nil {
+			t.Errorf("failed to list services with json format. args %q: %v", rr.Command(), err)
```
spowelljr, line 1551: This should be a `Fatalf` as well as we can't continue on if this command fails.

spowelljr (changes requested):
Can you also run `make generate-docs` to update the website documentation with the new flag, thanks

test/integration/functional_test.go
OmSaran: Fixed!

test/integration/functional_test.go
OmSaran: Fixed!

test/integration/functional_test.go
OmSaran: Fixed!

OmSaran:
> Can you also run `make generate-docs` to update the website documentation with the new flag, thanks

I did, but I see some other changes too which I did not intend but was a consequence of `make generate-docs` e.g., [here](https://github.com/kubernetes/minikube/pull/15831/files#diff-b52b100e981fc7f9ee47d189c006d31d4bc41d3dd1c7fc904a6014229728789dR991). Would be great if you could let me know if I should fix it and guide me on how to fix it.

spowelljr:
> > Can you also run `make generate-docs` to update the website documentation with the new flag, thanks
> 
> I did, but I see some other changes too which I did not intend but was a consequence of `make generate-docs` e.g., [here](https://github.com/kubernetes/minikube/pull/15831/files#diff-b52b100e981fc7f9ee47d189c006d31d4bc41d3dd1c7fc904a6014229728789dR991). Would be great if you could let me know if I should fix it and guide me on how to fix it.

That's just a change from a previous PR that didn't get the command run on it, it's fine to include it.

spowelljr (changes requested):
My only concern is that if using Docker on macOS we strip out the service URLs for table output but not for JSON which will cause a discrepancy between the two outputs.

To solve this we can loop through the services before the switch statement to remove the URLs so the outputs match.

OmSaran:
> My only concern is that if using Docker on macOS we strip out the service URLs for table output but not for JSON which will cause a discrepancy between the two outputs.
> 
> To solve this we can loop through the services before the switch statement to remove the URLs so the outputs match.

I fixed this in the new commit, please take a look.

PR #15834, opened 2023-02-12
Ensure only one instance of tunnel process runs

### LINKED ISSUE #15825: minikube tunnel should not be allowed to run twice
### What Happened?

right now we can run minikube tunnel in two different terminals two times ! which makes no sense

(follow example of Loadbalancer in site)

we should tell user another instance of minikube tunnel is running and exit safely

### Attach the log file

na

### Operating System

None

### Driver

None

OmSaran:
fixes #15825 

Both outputs are for the second invocation

Before: 
```
$ ./out/minikube tunnel
✅  Tunnel successfully started

📌  NOTE: Please do not close this terminal as this process must stay alive for the tunnel to be accessible ...

🏃  Starting tunnel for service balanced.
```

After:
```
$ ./out/minikube tunnel

💡  Exiting due to TUNNEL_ALREADY_RUNNING: Another tunnel process already running, terminate the older to start a new one

```

**If cluster is not running, this is the behavior:**
Cluster status:
```
$ ./out/minikube status
minikube
type: Control Plane
host: Stopped
kubelet: Stopped
apiserver: Stopped
kubeconfig: Stopped
```

Behavior:
```
$ ./out/minikube tunnel
```


cmd/minikube/cmd/tunnel.go
```diff
@@ -54,6 +57,7 @@ var tunnelCmd = &cobra.Command{
 		RootCmd.PersistentPreRun(cmd, args)
 	},
 	Run: func(cmd *cobra.Command, args []string) {
+		defer cleanupLock()
```
medyagh, line 60: Can you plz also put output of minikube when cluster is not running ? To ensure this wont crash

cmd/minikube/cmd/tunnel.go
```diff
@@ -116,6 +122,33 @@ var tunnelCmd = &cobra.Command{
...
+
+func checkNoOtherTunnelProcess() {
+	tunnelLockPath := filepath.Join(localpath.MiniPath(), ".tunnel_lock")
+
+	_, err := os.OpenFile(tunnelLockPath, os.O_RDWR|os.O_TRUNC, 0600)
+	if err != nil {
+		if os.IsNotExist(err) {
+			_, err = os.Create(tunnelLockPath)
+			if err != nil {
+				exit.Error(reason.SvcTunnelStart, fmt.Sprintf("error creating lock for tunnel file (%s)", tunnelLockPath), err)
```
medyagh, line 139: Lets create a new reason in reason package and we can call it
TUNNEL_ALREADY_RUNNING

also make sure the exit message does not have github box since this is not a crash or issue. This is a usage problem.

look at minikube profile list code, when there is no cluster running it exits nicely (no need to show github issue box and crying emoji)

medyagh (changes requested):

medyagh (changes requested):
Consider adding an integration test.

in functional test we already have a test that runs minikube tunnel, we could add a new test and run tunnel second time and expect it to fail on second time with exact error type

cmd/minikube/cmd/tunnel.go
OmSaran: @medyagh 
1. I've added the new reason as proposed.
2. Fixed the github box, however I've made this change only when we fail to _acquire_ filelock. If there are errors in creating the file itself or creating the lock handle itself, then I have left it as is (with github box). Let me know if this is not the correct thing to do. (Note: I referred to `service list` code instead of `profile list`. I couldn't seem to find the example in `profile list`)

OmSaran:
@medyagh I've added the integration test as well. Would be great if you could take a look!

cmd/minikube/cmd/tunnel.go
OmSaran: Updated PR description with the output. Let me know if that is not what you were expecting!

test/integration/functional_test_tunnel_test.go
```diff
@@ -131,6 +133,49 @@ func validateTunnelStart(ctx context.Context, t *testing.T, profile string) {
 	tunnelSession = *ss
 }
 
+// validateNoSecondTunnel ensures more than 1 tunnel can NOT run simultaneously
+func validateNoSecondTunnel(ctx context.Context, t *testing.T, profile string) {
```
medyagh, line 137: these tests all run in parallel, so ("StartTunnel", validateTunnelStart) test might be running before this and would cause niether of them be running,..


so I recommend adding this test Under Same test "validateTunnelStart" and right after Starting tunnel, we can run it one more time, just to ensure second time wont run.

cmd/minikube/cmd/tunnel.go
```diff
@@ -116,6 +122,36 @@ var tunnelCmd = &cobra.Command{
...
+func cleanupLock() {
+	if lockHandle != nil {
+		err := lockHandle.Unlock()
+		if err != nil {
+			out.Styled(style.Warning, fmt.Sprintf("failed to release lock during cleanup: %s", err.Error()))
+		}
+	}
+}
+
+func checkNoOtherTunnelProcess() {
```
medyagh, line 134: rename this to mustLockOrExit

medyagh (changes requested):
hi @OmSaran I requested a couple changes, but over all looks good

test/integration/functional_test_tunnel_test.go
OmSaran: @medyagh according to my understanding the tests within this file, based on code [here](https://github.com/kubernetes/minikube/blob/ba699acc9bf090d466453d19c6f8caaeb8992ff5/test/integration/functional_test_tunnel_test.go#L59) and [here](https://github.com/kubernetes/minikube/blob/ba699acc9bf090d466453d19c6f8caaeb8992ff5/test/integration/functional_test_tunnel_test.go#L72), will run serially, in order according to the list defined [here](https://github.com/kubernetes/minikube/blob/ba699acc9bf090d466453d19c6f8caaeb8992ff5/test/integration/functional_test_tunnel_test.go#L64). I've put this new test before `validateTunnelStart` to ensure that we don't face the scenario you mentioned. Please correct me if I'm wrong! If you still think I should put it inside `validateTunnelStart`, let me know, I'm happy to make the change.

cmd/minikube/cmd/tunnel.go
OmSaran: Done!

test/integration/functional_test_tunnel_test.go
```diff
@@ -131,6 +133,49 @@ func validateTunnelStart(ctx context.Context, t *testing.T, profile string) {
 	tunnelSession = *ss
 }
 
+// validateNoSecondTunnel ensures more than 1 tunnel can NOT run simultaneously
```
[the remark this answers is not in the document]
OmSaran: Done!

cmd/minikube/cmd/tunnel.go
```diff
@@ -116,6 +122,36 @@ var tunnelCmd = &cobra.Command{
...
+				exit.Error(reason.SvcTunnelStart, fmt.Sprintf("error creating lock for tunnel file (%s)", tunnelLockPath), err)
+			}
+		} else {
+			exit.Error(reason.SvcTunnelStart, fmt.Sprintf("error creating lock for tunnel file (%s)", tunnelLockPath), err)
+		}
+	}
+	lockHandle = fslock.New(tunnelLockPath)
+	err = lockHandle.TryLock()
+	if err == fslock.ErrLocked {
+		exit.Message(reason.SvcTunnelAlreadyRunning, "Another tunnel process already running, terminate the older to start a new one")
```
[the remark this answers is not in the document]
OmSaran: Done!

cmd/minikube/cmd/tunnel.go
```diff
@@ -116,6 +122,36 @@ var tunnelCmd = &cobra.Command{
...
+	_, err := os.OpenFile(tunnelLockPath, os.O_RDWR|os.O_TRUNC, 0600)
+	if err != nil {
+		if os.IsNotExist(err) {
+			_, err = os.Create(tunnelLockPath)
+			if err != nil {
+				exit.Error(reason.SvcTunnelStart, fmt.Sprintf("error creating lock for tunnel file (%s)", tunnelLockPath), err)
+			}
+		} else {
+			exit.Error(reason.SvcTunnelStart, fmt.Sprintf("error creating lock for tunnel file (%s)", tunnelLockPath), err)
+		}
```
spowelljr, line 146:
```suggestion
		if !os.IsNotExist(err) {
			exit.Error(reason.SvcTunnelStart, fmt.Sprintf("error creating lock for tunnel file (%s)", tunnelLockPath), err)
		}
		_, err = os.Create(tunnelLockPath)
		if err != nil {
			exit.Error(reason.SvcTunnelStart, fmt.Sprintf("error creating lock for tunnel file (%s)", tunnelLockPath), err)
		}
```
Can prevent a lot of this nesting by negating the first `if`

spowelljr (changes requested):

cmd/minikube/cmd/tunnel.go
OmSaran: Thanks, fixed!

cmd/minikube/cmd/tunnel.go
```diff
@@ -116,6 +122,35 @@ var tunnelCmd = &cobra.Command{
...
+		_, err = os.Create(tunnelLockPath)
+		if err != nil {
+			exit.Error(reason.SvcTunnelStart, fmt.Sprintf("error creating lock for tunnel file (%s)", tunnelLockPath), err)
+		}
+	}
+	lockHandle = fslock.New(tunnelLockPath)
+	err = lockHandle.TryLock()
+	if err == fslock.ErrLocked {
+		exit.Message(reason.SvcTunnelAlreadyRunning, "Another tunnel process is already running, terminate the existing instance to start a new one")
+	}
```
spowelljr, line 151: We're only checking for `fslock.ErrLocked`, but if the error is not nil but not `fslock.ErrLocked` we should at least `klog.Warningf` it so we can see the cause of the lock error in the logs for debugging pruposes.

spowelljr (changes requested):

cmd/minikube/cmd/tunnel.go
OmSaran: I updated the code to exit (we should only start the tunnel if we succeed to acquire the lock) if we fail to acquire the lock due to unexpected error, let me know if that is okay!

spowelljr (changes requested):
Just a couple small things and then it's good to merge, thanks

cmd/minikube/cmd/tunnel.go
```diff
@@ -116,6 +122,38 @@ var tunnelCmd = &cobra.Command{
...
+}
+
+func mustLockOrExit() {
+	tunnelLockPath := filepath.Join(localpath.MiniPath(), ".tunnel_lock")
+
+	_, err := os.OpenFile(tunnelLockPath, os.O_RDWR|os.O_TRUNC, 0600)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			exit.Error(reason.SvcTunnelStart, fmt.Sprintf("error opening lock file for tunnel file (%s)", tunnelLockPath), err)
```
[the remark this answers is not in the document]
OmSaran: Fixed!

cmd/minikube/cmd/tunnel.go
```diff
@@ -116,6 +122,38 @@ var tunnelCmd = &cobra.Command{
 	},
 }
 
+func cleanupLock() {
+	if lockHandle != nil {
+		err := lockHandle.Unlock()
+		if err != nil {
+			out.Styled(style.Warning, fmt.Sprintf("failed to release lock during cleanup: %s", err.Error()))
```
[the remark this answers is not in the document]
OmSaran: Fixed!

test/integration/functional_test_tunnel_test.go
```diff
@@ -131,6 +133,49 @@ func validateTunnelStart(ctx context.Context, t *testing.T, profile string) {
...
+		if err != nil {
+			t.Errorf("Failed to start tunnel for the first time")
+		}
+		sessions[idx] = session
+		defer cancel()
+
+		err = session.cmd.Wait()
+		if exErr, ok := err.(*exec.ExitError); ok {
+			exitCodeCh <- exErr.ExitCode()
+		} else {
```
spowelljr, line 157:
```suggestion
		} else if err != nil {
```
The `else` will also include `if err == nil`, for example, if the tunnel command runs successfully it won't return an `ExitError` so it will hit the `else` case and run `t.Errorf` even though there was no error.

test/integration/functional_test_tunnel_test.go
```diff
@@ -131,6 +133,49 @@ func validateTunnelStart(ctx context.Context, t *testing.T, profile string) {
...
+			exitCodeCh <- -1
+			t.Errorf("Second tunnel command failed due to a different reason: %s", err)
```
spowelljr, line 159:
```suggestion
			t.Errorf("tunnel command failed due to a different reason: %s", err)
```
This error can be misleading, if the first tunnel fails to run for some reason, the error message `Second tunnel command failed due to a different reason` can confuse the user as they're going to assume it's the second tunnel when it can be the first tunnel that failed.

spowelljr (changes requested):

test/integration/functional_test_tunnel_test.go
OmSaran: Nice catch, I fixed it, thanks!

test/integration/functional_test_tunnel_test.go
OmSaran: Fixed.

test/integration/functional_test_tunnel_test.go
```diff
@@ -131,6 +133,49 @@ func validateTunnelStart(ctx context.Context, t *testing.T, profile string) {
...
+		}
+		sessions[idx] = session
+		defer cancel()
+
+		err = session.cmd.Wait()
+		if exErr, ok := err.(*exec.ExitError); ok {
+			exitCodeCh <- exErr.ExitCode()
+		} else if err != nil {
+			exitCodeCh <- -1
+			t.Errorf("tunnel command failed due to a different reason: %s", err)
```
spowelljr, line 159:
```suggestion
			t.Errorf("tunnel command failed due to a different reason: %v", err)
```

spowelljr (changes requested):
Test failed on Docker Windows

https://storage.googleapis.com/minikube-builds/logs/15834/28009/Docker_Windows.html#fail_TestFunctional/parallel/TunnelCmd/serial/RunSecondTunnel

test/integration/functional_test_tunnel_test.go
spowelljr, line 159: For helping debug the error it would be useful to include `session.Stdout.String()` & `session.Stderr.String()` in the error string so we can see the actual error message as currently only the exit code is being passed.

`Failed to due to some other reason: 81`

spowelljr (changes requested):

OmSaran:
Hi, anyway I can help move this forward?

spowelljr:
> Hi, anyway I can help move this forward?

Looks like the new test you added fails on Windows

https://storage.googleapis.com/minikube-builds/logs/15834/28009/Docker_Windows.html#fail_TestFunctional/parallel/TunnelCmd/serial/RunSecondTunnel

https://storage.googleapis.com/minikube-builds/logs/15834/28303/Hyper-V_Windows.html#fail_TestFunctional/parallel/TunnelCmd/serial/RunSecondTunnel

OmSaran:
> > Hi, anyway I can help move this forward?
> 
> Looks like the new test you added fails on Windows
> 
> https://storage.googleapis.com/minikube-builds/logs/15834/28009/Docker_Windows.html#fail_TestFunctional/parallel/TunnelCmd/serial/RunSecondTunnel
> 
> https://storage.googleapis.com/minikube-builds/logs/15834/28303/Hyper-V_Windows.html#fail_TestFunctional/parallel/TunnelCmd/serial/RunSecondTunnel

I can think of 2 ways of debugging this:
1. Run this test manually on a windows machine.
2. Log the outputs in the code to take a look at the error in tests of CI/CD pipeline (I just pushed a commit to log `stdout` and `stderr`). 

What do you think I should do? Or is there a better way to debug this? (Note that for 1. I'll need access to a windows machine which I don't have)

cmd/minikube/cmd/tunnel.go
```diff
@@ -116,6 +122,38 @@ var tunnelCmd = &cobra.Command{
...
+	_, err := os.OpenFile(tunnelLockPath, os.O_RDWR|os.O_TRUNC, 0600)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			exit.Error(reason.SvcTunnelStart, fmt.Sprintf("error opening lock file for tunnel file (%s)", tunnelLockPath), err)
+		}
+		_, err = os.Create(tunnelLockPath)
+		if err != nil {
+			exit.Error(reason.SvcTunnelStart, fmt.Sprintf("error creating lock for tunnel file (%s)", tunnelLockPath), err)
+		}
+	}
```
spowelljr, line 146:
```suggestion
```

spowelljr (changes requested):
> What do you think I should do? Or is there a better way to debug this? (Note that for 1. I'll need access to a windows machine which I don't have)

@OmSaran I ran it on Windows and found the issue. Window is finicky about open files, so you're creating the lock file with `os.Create` but it's not closed, so on Windows you can't interact with a file if it's open in another process. So when you call `fslock.New(tunnelLockPath)` & `lockHandle.TryLock()` it throws an error as `os.Create` is the active one using it. I would have close the file after creating it, but I don't think all this open and creation code is needed as I believe `fslock.New` will handle creating the file for us. We have other cases of using `fslock.New` and we don't create the file ahead of time. I also tried removing all the file creation code and the file still existed for me after running the tunnel command.

cmd/minikube/cmd/tunnel.go
OmSaran: Thank you so much for debugging it. I have made changes as suggested. I also confirmed source code of the package to ensure they file is created upon `TryLock()` across platforms (e..g, for *nix [here](https://github.com/juju/fslock/blob/4d5c94c67b4b207e1ab4ebca6b4e47f174618b86/fslock_nix.go#L36) and [here](https://github.com/juju/fslock/blob/4d5c94c67b4b207e1ab4ebca6b4e47f174618b86/fslock_nix.go#L49)).

test/integration/functional_test_tunnel_test.go
```diff
@@ -131,6 +133,69 @@ func validateTunnelStart(ctx context.Context, t *testing.T, profile string) {
...
+		if err != nil {
+			if exErr, ok := err.(*exec.ExitError); !ok {
+				t.Logf("failed to coerce exit error: %v", err)
+				exitCode = -1
+			} else {
+				exitCode = exErr.ExitCode()
+			}
+		}
+
+		sessCh <- SessInfo{Stdout: string(stdout), Stderr: string(stderr), ExitCode: exitCode}
```
spowelljr, line 180:
```suggestion
		exitCode := 0
		err = session.cmd.Wait()
		if err != nil {
			if exErr, ok := err.(*exec.ExitError); !ok {
				t.Logf("failed to coerce exit error: %v", err)
				exitCode = -1
			} else {
				exitCode = exErr.ExitCode()
			}
		}

		sessCh <- SessInfo{Stdout: session.Stdout.String(), Stderr: session.Stderr.String(), ExitCode: exitCode}
```

To get the contents of `session.Stdout` you don't have to do `stdout, err := io.ReadAll(session.Stdout)` you can simply do `session.Stdout.String()`

spowelljr (changes requested):

test/integration/functional_test_tunnel_test.go
OmSaran: So I had previously tried out exactly what you said (to try to print it in build job), it doesn't work because we need to do the read before the `wait()`, otherwise content is lost since pipe is closed after wait. Reference: [pipe source code](https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/os/exec/exec.go;l=1060-1064) and [here](https://github.com/kubernetes/minikube/blob/2d8bc3e0565733559bc78c0f72e8ab12c046e7e8/test/integration/helpers_test.go#L135) (i.e., session.Stderr = cmd.StderrPipe()).

spowelljr: I see, my concern isn't necessarily where we read the contents, just that instead of:
```
		stderr, err := io.ReadAll(session.Stderr)
		if err != nil {
			t.Logf("Failed to read stderr: %v", err)
		}
		stdout, err := io.ReadAll(session.Stdout)
		if err != nil {
			t.Logf("Failed to read stdout: %v", err)
		}
```

It could be replaced with:
```
stderr := session.Stderr.String()
stdout := session.Stdout.String()
```

OmSaran: Ahh okay. But I don't see a method called `String()` for `*bufio.Reader` (which is the type of `Stderr`/`Stdout`). 
Got the following compile time error: `session.Stderr.String undefined (type *bufio.Reader has no field or method String)`

Let me know if I'm missing something!

spowelljr (changes requested):
Looks good, just one last thing, can you run `make generate-docs` to update the translations and test list, thanks!

OmSaran:
> Looks good, just one last thing, can you run `make generate-docs` to update the translations and test list, thanks!

Done, thank you for your patience and guidance!

spowelljr (approved):
Thanks for persisting through this PR and helping add this feature!

raylau1234:
What if I have multiple profiles with multi-node clusters?

spowelljr:
> What if I have multiple profiles with multi-node clusters?

There's no way to specify which node to use with `minikube tunnel` so I don't think the multi-node part is a concern, but you are correct that the lock file should be per profile and not a global one, good catch, I'll make a PR to fix this!

raylau1234:
> There's no way to specify which node to use with `minikube tunnel` so I don't think the multi-node part is a concern, but you are correct that the lock file should be per profile and not a global one, good catch, I'll make a PR to fix this!

Thanks!! 

Yes, that's what I was intending to say - multi profile... I said "multi node" bec that's the multiple profiles is introduced (AFAIK) in docs at https://minikube.sigs.k8s.io/docs/tutorials/multi_node/

spowelljr:
> Yes, that's what I was intending to say - multi profile... I said "multi node" bec that's the multiple profiles is introduced (AFAIK) in docs at https://minikube.sigs.k8s.io/docs/tutorials/multi_node/

Just to clarify, multiple profiles and multi-node are two different things.

You can start independent clusters with completely different configurations using the `--profile` flag. If you don't provide a `--profile` flag it defaults the flag to `minikube`.

Example:
```
$ minikube start -p profile1 --driver docker
😄  [profile1] minikube v1.30.1 on Darwin 13.4.1 (arm64)
...

$ minikube start -p profile2 --driver qemu --container-runtime=containerd
😄  [profile2] minikube v1.30.1 on Darwin 13.4.1 (arm64)
...

$ minikube profile list
|----------|-----------|------------|----------------|------|---------|---------|-------|--------|
| Profile  | VM Driver |  Runtime   |       IP       | Port | Version | Status  | Nodes | Active |
|----------|-----------|------------|----------------|------|---------|---------|-------|--------|
| profile1 | docker    | docker     | 192.168.49.2   | 8443 | v1.27.3 | Running |     1 |        |
| profile2 | qemu2     | containerd | 192.168.105.63 | 8443 | v1.27.3 | Running |     1 |        |
|----------|-----------|------------|----------------|------|---------|---------|-------|--------|
```

The link you posted is multi-node, which is adding worker nodes to a profile.

Example:
```
$ minikube start -n=3

# see the Nodes column below
$ minikube profile list
|----------|-----------|---------|--------------|------|---------|---------|-------|--------|
| Profile  | VM Driver | Runtime |      IP      | Port | Version | Status  | Nodes | Active |
|----------|-----------|---------|--------------|------|---------|---------|-------|--------|
| minikube | docker    | docker  | 192.168.49.2 | 8443 | v1.27.3 | Running |     3 | *      |
|----------|-----------|---------|--------------|------|---------|---------|-------|--------|

# below you can see there's an instance of kindnet and kube-proxy on each node, hence the three of each
$ kubectl get pods -A
kubectl get pod -A
NAMESPACE     NAME                               READY   STATUS    RESTARTS      AGE
kube-system   coredns-5d78c9869d-tw7xr           1/1     Running   1 (81s ago)   95s
kube-system   etcd-minikube                      1/1     Running   0             108s
kube-system   kindnet-22wzx                      1/1     Running   0             95s
kube-system   kindnet-n8bmm                      1/1     Running   0             90s
kube-system   kindnet-xrj72                      1/1     Running   0             76s
kube-system   kube-apiserver-minikube            1/1     Running   0             107s
kube-system   kube-controller-manager-minikube   1/1     Running   0             107s
kube-system   kube-proxy-bm529                   1/1     Running   0             95s
kube-system   kube-proxy-q7kr8                   1/1     Running   0             90s
kube-system   kube-proxy-wccfj                   1/1     Running   0             76s
kube-system   kube-scheduler-minikube            1/1     Running   0             109s
kube-system   storage-provisioner                1/1     Running   1 (64s ago)   106s
```

Hope that clears it up

raylau1234:
Understood... my usual practice is to have one profile per project.... but have multi nodes w/in the profile bec testing rack awareness is usually on the list....

So when I have 2 projects going (or sometimes even 3!), would love to be able to have 2 tunnels going.

spowelljr:
> So when I have 2 projects going (or sometimes even 3!), would love to be able to have 2 tunnels going.

For sure, that was an oversight by me when reviewing the PR, thanks for catching it before the release!

Here's the PR https://github.com/kubernetes/minikube/pull/16839

PR #15850, opened 2023-02-14
Fix bash completion for kubectl symlinked to minikube by not adding `--cluster` flag for the `kubectl __complete` subcommand

### LINKED ISSUE #14959: Minikube kubectl autocompletion not working
### What Happened?

As the documentation specified I tried using the `kubectl` provided in the project using:

```bash
sudo ln -s $(which minikube) /usr/local/bin/kubectl
```

I am on Ubuntu 22.04 and I tried both `bash` and `zsh`. In both cases the autocompletion doesn't work. I tried multiple things, sourcing the auto completion manually and putting them in the configuration file for each respective shell:

```bash
source <(kubectl completion zsh)
source <(kubectl completion bash)
```

I tried using the `oh-my-zsh` plugin. Nothing - it just doesn't work.

After installing a dedicated kubectl package via `snapd` everything works fine. There is just something wrong with the minikube setup.

### Attach the log file

Not applicable.

### Operating System

Ubuntu

### Driver

Docker

ben-krieger:
Fixes #14959

Before:

```bash
$ go run ./cmd/minikube kubectl __complete ge
Error: flags cannot be placed before plugin name: --cluster
exit status 1
```

After:

```bash
$ go run ./cmd/minikube kubectl __complete ge
get	Display one or many resources
:4
Completion ended with directive: ShellCompDirectiveNoFileComp
```

Signed-off-by: Ben Krieger <ben.krieger@intel.com>

### DIFF
--- a/cmd/minikube/cmd/kubectl.go
+++ b/cmd/minikube/cmd/kubectl.go
@@ -130,7 +130,7 @@ host. Please be aware that when using --ssh all paths will apply to the remote m
 			os.Exit(1)
 		}
 
-		if len(args) > 1 && args[0] != "--help" {
+		if len(args) > 1 && args[0] != "--help" && args[0] != cobra.ShellCompRequestCmd {
 			cluster := []string{"--cluster", cname}
 			args = append(cluster, args...)
 		}


cmd/minikube/cmd/kubectl.go
```diff
@@ -96,7 +96,7 @@ host. Please be aware that when using --ssh all paths will apply to the remote m
 			os.Exit(1)
 		}
 
-		if len(args) > 1 && args[0] != "--help" {
+		if len(args) > 1 && args[0] != "--help" && args[0] != "__complete" {
```
medyagh, line 99: Thanks for fixing this issue, @ben-krieger 
I am curious, 
can we guarantee the __complete would always be args[0] ? if not maybe we can loop over the args and flag it  as skip dash cluster... ?

ben-krieger: So put another way, "Are there any valid flags for `kubectl` when using the `__complete` subcommand?" That's a good question. I assumed not, but I'll check the docs/code. Then there's the question of whether we think that any such flags may get added in the future...

ben-krieger: The global flags are `namespace`, `context`, `cluster`, and `user` (https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/cmd.go#L499-L520). None of those make sense as flags to `__complete`.

But I guess more importantly, `__complete` is a plugin (I guess?) and for the plugin handler code https://github.com/kubernetes/kubectl/blob/31af32762da0400245ada0c342327da4b78eb182/pkg/cmd/cmd.go#L227-L239 it guarantees that there are no flags before the plugin name.

I'm not familiar enough with the plugin architecture, but this seems like a strong guarantee that `__complete` will always be `args[0]`.

medyagh: ok thats fair enough for me ! and thanks for this contribution ! I look forward to see more contributions from you

PR #15851, opened 2023-02-14
Prevent panic if docker version returns exit code 0 with unexpected output

spowelljr:
There exists a way that `docker version --format {{.Server.Os}}-{{.Server.Version}}:{{.Server.Platform.Name}}` returns exit code 0 but does not return the expected output, such that it doesn't contain a `:`. This causes an error as we split on `:` as assuming that we'd have a length of 2 if `docker version` returns successfully. But since this assumption is incorrect, we need to prevent a panic.

### DIFF
--- a/pkg/minikube/registry/drvs/docker/docker.go
+++ b/pkg/minikube/registry/drvs/docker/docker.go
@@ -94,32 +94,9 @@ func configure(cc config.ClusterConfig, n config.Node) (interface{}, error) {
 }
 
 func status() (retState registry.State) {
-	_, err := exec.LookPath(oci.Docker)
-	if err != nil {
-		return registry.State{Error: err, Installed: false, Healthy: false, Fix: "Install Docker", Doc: docURL}
-	}
-
-	ctx, cancel := context.WithTimeout(context.Background(), 6*time.Second)
-	defer cancel()
-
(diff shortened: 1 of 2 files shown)


pkg/minikube/registry/drvs/docker/docker.go
```diff
@@ -136,6 +136,9 @@ func status() (retState registry.State) {
 	}()
 
 	versions := strings.Split(string(o), ":")
+	if len(versions) < 2 {
```
medyagh, line 139: maybe we can add a unit test case in TestCheckDockerEngineVersion ?
https://github.com/medyagh/minikube/blob/99756bef0b9a90b9097668cb4e9c255f387cbadd/pkg/minikube/registry/drvs/docker/docker_test.go#L67

medyagh (approved):
look good, might look better if possible to add cases to unit test if not thats fine too

pkg/minikube/registry/drvs/docker/docker.go
spowelljr: Added units tests

medyagh (approved):

medyagh:
I really like the refactor and simpllfying and huge increase in code readability , thank you ! @spowelljr

PR #15856, opened 2023-02-14
Add shell completion for minikube kubectl subcommand

ben-krieger:
Before:

```bash
$ : Tab complete performs file completion
$ minikube kubectl [TAB]
fileA fileB
```

After:

```bash
$ : Tab complete uses kubectl __complete under the hood
$ minikube kubectl [TAB]
alpha          apply          certificate    cordon         delete         edit           expose         label          plugin         rollout        taint          wait           
annotate       attach         cluster-info   cp             describe       events         get            logs           port-forward   run            top            
api-resources  auth           completion     create         diff           exec           help           options        proxy          scale          uncordon       
api-versions   autoscale      config         debug          drain          explain        kustomize      patch          replace        set            version
```

### DIFF
--- a/cmd/minikube/cmd/kubectl.go
+++ b/cmd/minikube/cmd/kubectl.go
@@ -17,14 +17,19 @@ limitations under the License.
 package cmd
 
 import (
+	"bufio"
+	"bytes"
 	"fmt"
 	"os"
 	"os/exec"
 	"path"
+	"strconv"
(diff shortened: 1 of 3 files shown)


ben-krieger:
I'll try to add a test for this, but tab completions aren't the easiest to test given the Cobra API. Likely it'll have to be with os/exec.

PR #15868, opened 2023-02-16
Don't allow enabling addon on a paused cluster

### LINKED ISSUE #15857: Prevent users from modifying addons while cluster is paused
Currently you can try enabling/disabling an addon while the cluster is paused, but it will hang. We should prevent users from doing this and notify them that they can't perform this action while the cluster is paused and they should unpause the cluster if they wish to modify the addons.

Example:
```
$ minikube start
$ minikube pause
$ minikube addons enable gcp-auth
# hangs
```

norshtein:
fixes #15857

Tested in my local environment, when trying to enable `auto-pause` on a paused cluster, it gives following output and exit immediately:
```
$ minikube addons enable auto-pause
💡  auto-pause is an addon maintained by Google. For any concerns contact minikube on GitHub.
You can view the list of minikube maintainers at: https://github.com/kubernetes/minikube/blob/master/OWNERS
🤷  Can't enable addon on a paused cluster, please unpause the cluster firstly.
    ▪ https://github.com/kubernetes/minikube/labels/co/auto-pause

❌  Exiting due to MK_ADDON_ENABLE: run callbacks: running callbacks: [Can't enable addon on a paused cluster, please unpause the cluster firstly.]

```

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -292,8 +292,13 @@ func EnableOrDisableAddon(cc *config.ClusterConfig, name string, val string) err
 		return errors.Wrap(err, "check paused")
 	}
 	if runtimePaused {
-		out.Styled(style.Shrug, `Can't enable addon on a paused cluster, please unpause the cluster firstly.`)
-		return errors.New("Can't enable addon on a paused cluster, please unpause the cluster firstly.")
+		action := "disable"
+		if enable {
+			action = "enable"
+		}
+		msg := fmt.Sprintf("can't %s addon on a paused cluster, please unpause the cluster firstly.", action)
+		out.Styled(style.Shrug, msg)
+		return errors.New(msg)
 	}
 	bail, err := addonSpecificChecks(cc, name, enable, runner)
 	if err != nil {
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -632,7 +632,7 @@ func CheckPaused(profile string, enable bool) error {
 		if enable {
 			action = "enable"
 		}
-		msg := fmt.Sprintf("can't %s addon on a paused cluster, please unpause the cluster firstly.", action)
+		msg := fmt.Sprintf("Can't %s addon on a paused cluster, please unpause the cluster first.", action)
 		out.Styled(style.Shrug, msg)
 		return errors.New(msg)
 	}
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -36,9 +36,11 @@ import (
 	"k8s.io/minikube/pkg/drivers/kic/oci"
 	"k8s.io/minikube/pkg/kapi"
 	"k8s.io/minikube/pkg/minikube/assets"
+	"k8s.io/minikube/pkg/minikube/cluster"
 	"k8s.io/minikube/pkg/minikube/command"
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/constants"
+	"k8s.io/minikube/pkg/minikube/cruntime"
 	"k8s.io/minikube/pkg/minikube/driver"
 	"k8s.io/minikube/pkg/minikube/exit"
 	"k8s.io/minikube/pkg/minikube/machine"
@@ -280,6 +282,16 @@ func EnableOrDisableAddon(cc *config.ClusterConfig, name string, val string) err
 		return errors.Wrap(err, "command runner")
 	}
 
+	crName := cc.KubernetesConfig.ContainerRuntime
+	cr, err := cruntime.New(cruntime.Config{Type: crName, Runner: runner})
+	runtimePaused, err := cluster.CheckIfPaused(cr, []string{"kube-system"})
+	if err != nil {
+		return errors.Wrap(err, "check paused")
+	}
+	if runtimePaused {
+		out.Styled(style.Shrug, `Can't enable addon on a paused cluster, please unpause the cluster firstly.`)
+		return errors.New("Can't enable addon on a paused cluster, please unpause the cluster firstly.")
+	}
 	bail, err := addonSpecificChecks(cc, name, enable, runner)
 	if err != nil {
 		return err
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -29,6 +29,7 @@ import (
 	"time"
 
 	"github.com/blang/semver/v4"
+	"github.com/docker/machine/libmachine/state"
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 
@@ -611,6 +612,15 @@ func VerifyNotPaused(profile string, enable bool) error {
 	host, err := machine.LoadHost(api, config.MachineName(*cc, cp))
 	if err != nil {
 		return errors.Wrap(err, "get host")
+	}
+
+	s, err := host.Driver.GetState()
+	if err != nil {
+		return errors.Wrap(err, "get state")
+	}
+	if s != state.Running {
+		// can't check the status of pods on a non-running cluster
+		return nil
 	}
 
 	runner, err := machine.CommandRunner(host)
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -637,14 +637,14 @@ func VerifyNotPaused(profile string, enable bool) error {
 	if err != nil {
 		return errors.Wrap(err, "check paused")
 	}
-	if runtimePaused {
-		action := "disable"
-		if enable {
-			action = "enable"
-		}
-		msg := fmt.Sprintf("Can't %s addon on a paused cluster, please unpause the cluster first.", action)
-		out.Styled(style.Shrug, msg)
-		return errors.New(msg)
+	if !runtimePaused {
+		return nil
 	}
-	return nil
+	action := "disable"
+	if enable {
+		action = "enable"
+	}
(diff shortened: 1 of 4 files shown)


pkg/addons/addons.go
```diff
@@ -279,6 +281,19 @@ func EnableOrDisableAddon(cc *config.ClusterConfig, name string, val string) err
...
+	if err != nil {
+		return errors.Wrap(err, "container runtime")
+	}
+	runtimePaused, err := cluster.CheckIfPaused(cr, []string{"kube-system"})
+	if err != nil {
+		return errors.Wrap(err, "check paused")
+	}
+	if runtimePaused {
+		out.Styled(style.Shrug, `Can't enable addon on a paused cluster, please unpause the cluster firstly.`)
+		return errors.New("Can't enable addon on a paused cluster, please unpause the cluster firstly.")
```
spowelljr, line 295: Since the exact same string is being used twice it should be assigned to a variable and then pass the variable to each function. That way if we update the message in the future we don't have to update two strings.

spowelljr, line 295: Both strings say `Can't enable addon` but the user might be trying to `disable` the addon and the message will confuse them. The text should match the action they're trying to take.

spowelljr (changes requested):

medyagh (changes requested):
I rather we exit before showing the information about the addon, since it is useless for the users and makes it harder to read the error message

pkg/addons/addons.go
norshtein: Thanks for reviewing! Updated the code to use a shared string variable.

norshtein: Updated the code to correctly print enable/disable based on user's action.

norshtein:
> I rather we exit before showing the information about the addon, since it is useless for the users and makes it harder to read the error message

Thanks for reviewing! The information about the addon is shown at the very first when enabling addon https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/config/enable.go#L54, should we put the code for checking runtime status before that?

medyagh:
> > I rather we exit before showing the information about the addon, since it is useless for the users and makes it harder to read the error message
> 
> Thanks for reviewing! The information about the addon is shown at the very first when enabling addon https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/config/enable.go#L54, should we put the code for checking runtime status before that?

I rather it be like this

```

$ minikube addons enable auto-pause
🤷  Can't enable addon on a paused cluster, please unpause the cluster firstly.

❌  Exiting due to MK_ADDON_ENABLE_PAUSED:


```

norshtein:
Hi @medyagh , I've updated the code, now the output becomes:
```
$ minikube addons enable auto-pause
🤷  can't enable addon on a paused cluster, please unpause the cluster firstly.

❌  Exiting due to MK_ADDON_ENABLE_PAUSED: can't enable addon on a paused cluster, please unpause the cluster firstly.
```

pkg/addons/addons.go
```diff
@@ -570,3 +572,59 @@ func UpdateConfig(cc *config.ClusterConfig, enabled []string) {
 		}
 	}
 }
+
+// CheckPaused checks whether the cluster is paused before enable/disable an addon.
+func CheckPaused(profile string, val string) error {
```
spowelljr, line 577: Any reason `val` can't be a bool?

pkg/addons/addons.go
```diff
@@ -570,3 +572,59 @@ func UpdateConfig(cc *config.ClusterConfig, enabled []string) {
...
+	runtimePaused, err := cluster.CheckIfPaused(cr, []string{"kube-system"})
+	if err != nil {
+		return errors.Wrap(err, "check paused")
+	}
+	if runtimePaused {
+		action := "disable"
+		if enable {
+			action = "enable"
+		}
+		msg := fmt.Sprintf("can't %s addon on a paused cluster, please unpause the cluster firstly.", action)
```
spowelljr, line 625:
```suggestion
		msg := fmt.Sprintf("Can't %s addon on a paused cluster, please unpause the cluster first.", action)
```

pkg/addons/addons.go
```diff
@@ -570,3 +572,59 @@ func UpdateConfig(cc *config.ClusterConfig, enabled []string) {
...
+	defer api.Close()
+
+	cp, err := config.PrimaryControlPlane(cc)
+	if err != nil {
+		exit.Error(reason.GuestCpConfig, "Error getting primary control plane", err)
+	}
+
+	host, err := machine.LoadHost(api, config.MachineName(*cc, cp))
+	if err != nil {
+		exit.Error(reason.GuestLoadHost, "Error getting host", err)
```
spowelljr, line 603: It's a bit odd that it switches between `return errors.Wrap` and `exit.Error`. The calling function will handle `exit.Error` for us so I think it makes sense to keep to just keep consistent with `return errors.Wrap`.

spowelljr (changes requested):

pkg/addons/addons.go
norshtein: Updated to use `errors.Wrap`.

pkg/addons/addons.go
norshtein: I was thinking to keep consistent with `EnableOrDisableAddon`, which also makes `val` as a string. Now switch to use bool.

norshtein:
Hi @spowelljr , I've rebased on main branch and updated the code based on your code review, could you please take a look?

pkg/addons/addons.go
```diff
@@ -578,3 +580,54 @@ func UpdateConfigToDisable(cc *config.ClusterConfig) {
 		}
 	}
 }
+
+// CheckPaused checks whether the cluster is paused before enable/disable an addon.
+func CheckPaused(profile string, enable bool) error {
```
spowelljr, line 585: The name of the func `CheckPaused` and the logic of it is a bit counter-intuitive in my opinion. For example, if a func was named `CheckRunning` I would assume it to throw an error if it wasn't running, but `CheckPaused` throws an error if it is paused, so the name `CheckUnpaused` would make more sense to me.

ie. Reading through the func makes me think we want the cluster to be paused
```
		if len(args) != 1 {
			exit.Message(reason.Usage, "usage: minikube addons disable ADDON_NAME")
		}
		err := addons.CheckPaused(ClusterFlagValue(), false)
		if err != nil {
			exit.Error(reason.InternalAddonDisablePaused, "disable failed", err)
		}
```

spowelljr (changes requested):

pkg/addons/addons.go
medyagh: @norshtein I agree with spowelljr  I think the name should be something like"verifyNotPaused" or "CheckNotPaused" would be a better name

medyagh:
[norshtein](https://github.com/norshtein) this PR is close to be finihsed, please let me know when you update, we could merge it

norshtein:
@medyagh sorry I was busy with some affairs last week, will update this PR soon.

pkg/addons/addons.go
norshtein: Thanks for the suggestion, that makes great sense. I've renamed the function to `VerifyNotPaused`

spowelljr:
Looks like this is causing some failures on `TestAddons/StoppedEnableDisable`

spowelljr (changes requested):
The problem looks like it's not checking if the cluster is stopped before trying to check if the pods are paused. If the cluster is stopped have `VerifyNotPaused` return nil, otherwise then check if kube-system is paused or not. Trying to check the pods on a stopped cluster will result in errors.

medyagh:
> The problem looks like it's not checking if the cluster is stopped before trying to check if the pods are paused. If the cluster is stopped have `VerifyNotPaused` return nil, otherwise then check if kube-system is paused or not. Trying to check the pods on a stopped cluster will result in errors.

@norshtein are you still working on this PR?

spowelljr:
Fixed throwing an error on a stopped cluster

**Before:**
```
$ minikube addons enable dashboard

❌  Exiting due to MK_ADDON_ENABLE_PAUSED: enabled failed: check paused: list paused: docker: NewSession: new client: new client: Error creating new ssh host from driver: Error getting ssh port for driver: get ssh host-port: unable to inspect a not running container to get SSH port

╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                         │
│    😿  If the above advice does not help, please let us know:                                                           │
│    👉  https://github.com/kubernetes/minikube/issues/new/choose                                                         │
│                                                                                                                         │
│    Please run `minikube logs --file=logs.txt` and attach logs.txt to the GitHub issue.                                  │
│    Please also attach the following file to the GitHub issue:                                                           │
│    - /var/folders/9l/6wpxv6wd1b901m1146r579wc00rqw3/T/minikube_addons_72a2d59c3bf8ff14a7f994656d47b6b59ed7b266_0.log    │
│                                                                                                                         │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

**After:**
```
$ minikube addons enable dashboard

💡  dashboard is an addon maintained by Kubernetes. For any concerns contact minikube on GitHub.
You can view the list of minikube maintainers at: https://github.com/kubernetes/minikube/blob/master/OWNERS
💡  Some dashboard features require the metrics-server addon. To enable all features please run:

	minikube addons enable metrics-server	


🌟  The 'dashboard' addon is enabled
```

spowelljr (approved):

PR #15908, opened 2023-02-22
Update --container-runtime docs to "crio" instead of "cri-o"

### LINKED ISSUE #15246: minikube podman-env complains about cri-o runtime spelling
### What Happened?

1. Start a cluster with `--container-runtime cri-o`
2. Run `minikube podman-env`
3. See an error message: `X Exiting due to MK_USAGE: The podman-env command is only compatible with the "crio" runtime, but this cluster was configured to use the "cri-o" runtime.`

It seems the code disagrees on how to spell CRI-O: 

https://github.com/kubernetes/minikube/blob/074e681975c6c82deeebd5317bf328e0f0fab512/pkg/minikube/cruntime/cruntime.go#L51-L54

vs

https://github.com/kubernetes/minikube/blob/074e681975c6c82deeebd5317bf328e0f0fab512/pkg/minikube/constants/constants.go#L59-L60

(checked here)

https://github.com/kubernetes/minikube/blob/074e681975c6c82deeebd5317bf328e0f0fab512/cmd/minikube/cmd/podman-env.go#L173-L176 

ValidRuntimes is used in the help message too, so it is misleading.

### Attach the log file

... do I really need a log file for this?

### Operating System

Windows

### Driver

Hyper-V

lednubr:
Mapped by this [issue](https://github.com/kubernetes/minikube/issues/15246). If you provide "cri-o" as config runtime, minikube will validate podman related commands agains "crio" word.

fixes #15246

### DIFF
--- a/site/content/en/docs/handbook/config.md
+++ b/site/content/en/docs/handbook/config.md
@@ -98,7 +98,7 @@ minikube start --container-runtime=docker
 Options available are:
 
 * [containerd]({{<ref "/docs/runtimes/containerd">}})
-* [cri-o]({{<ref "/docs/runtimes/cri-o">}})
+* [crio]({{<ref "/docs/runtimes/cri-o">}})
 * [docker]({{<ref "/docs/runtimes/docker">}})
 
 See <https://kubernetes.io/docs/setup/production-environment/container-runtimes/>


medyagh (changes requested):
the official name is https://github.com/cri-o/cri-o
minikube config changes it to crio for internal usage reasons but the name that public should use is cri-o

lednubr:
@medyagh do you have any suggestion on how to fix https://github.com/kubernetes/minikube/issues/15246 ? since if we provide cri-o, it fails to run podman commands

Issue #15420, opened 2022-11-29
Update GCP-Auth documentation before release


spowelljr:
https://github.com/kubernetes/minikube/pull/15403 will resolve the namespace issue that we have documented on our [website](https://minikube.sigs.k8s.io/docs/handbook/addons/gcp-auth/#adding-new-namespaces).

Shortly before the release we should include something like: `As of minikube v1.XX this has been fixed and the pull secret is automatically added to new namespaces.`

PR #15933, opened 2023-02-27
site: Update GCP-Auth docs regarding new namespaces

spowelljr:
Closes https://github.com/kubernetes/minikube/issues/15420

### DIFF
--- a/site/content/en/docs/handbook/addons/gcp-auth.md
+++ b/site/content/en/docs/handbook/addons/gcp-auth.md
@@ -102,6 +102,10 @@ To resolve this problem, run:
 
 ## Adding new namespaces
 
+### minikube v1.29.0+
+Newly created namespaces automatically have the image pull secret configured, no action is required.
+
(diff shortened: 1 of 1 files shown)


site/content/en/docs/handbook/addons/gcp-auth.md
```diff
@@ -102,6 +102,10 @@ To resolve this problem, run:
 
 ## Adding new namespaces
 
+### minikube v1.29.0+
```
medyagh, line 105: can we make it tabbed? if it makes sense

medyagh (changes requested):

medyagh (approved):

site/content/en/docs/handbook/addons/gcp-auth.md
spowelljr: The text under is only a line or two, so I don't think tabbing makes sense here

PR #15944, opened 2023-03-01
CI: Windows Docker setup fixes

spowelljr:
This PR includes 3 fixes and 1 improvement

## Fix 1
The error `target_url` was hard-coded to `Hyper-V_Windows.txt` if Docker Desktop setup failed. This would cause clicking the details on Docker Windows to take you to the Hyper-V logs.

Screenshot 2023-03-01 at 10 46 21 AM
Clicking the above would incorrectly take you to https://storage.googleapis.com/minikube-builds/logs/15939/28106/Hyper-V_Windows.txt

**Before:**
`https://storage.googleapis.com/$gcs_bucket/Hyper-V_Windows.txt`

**After:**
`https://storage.googleapis.com/$gcs_bucket/$env:JOB_NAME.txt`

## Fix 2
All Windows tests were requiring Docker Desktop to be running even when it wasn't required. If Docker Desktop was broken on a machine it would prevent Hyper-V tests from running when it should of. This may also increase performance on other tests as Docker Desktop will no longer be running and can free up some resources.

## Fix 3
`docker system prune -a --volumes -f` was only being run if starting Docker Desktop failed, this is the opposite of what should be happening.

## Improvement
Since Fix 2 will now allow tests to run without Docker Desktop if not needed, added a check to ensure Docker CLI is still present since it is required for `docker-env` tests.

### DIFF
--- a/hack/jenkins/common.ps1
+++ b/hack/jenkins/common.ps1
@@ -41,17 +41,26 @@ gsutil.cmd -m cp -r gs://minikube-builds/$env:MINIKUBE_LOCATION/windows_cleanup_
 gsutil.cmd -m cp -r gs://minikube-builds/$env:MINIKUBE_LOCATION/windows_cleanup_cron.ps1 out/
 ./out/windows_cleanup_cron.ps1
 
-# Make sure Docker is up and running
-gsutil.cmd -m cp -r gs://minikube-builds/$env:MINIKUBE_LOCATION/setup_docker_desktop_windows.ps1 out/
-./out/setup_docker_desktop_windows.ps1
-If ($lastexitcode -gt 0) {
-	echo "Docker failed to start, exiting."
+# check for Docker CLI
+docker --help
+if ($lastexitcode -gt 0) {
+        echo "Docker CLI not found, exiting."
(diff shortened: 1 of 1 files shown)


hack/jenkins/common.ps1
```diff
@@ -41,17 +41,26 @@ gsutil.cmd -m cp -r gs://minikube-builds/$env:MINIKUBE_LOCATION/windows_cleanup_
...
 
-# Make sure Docker is up and running
-gsutil.cmd -m cp -r gs://minikube-builds/$env:MINIKUBE_LOCATION/setup_docker_desktop_windows.ps1 out/
-./out/setup_docker_desktop_windows.ps1
-If ($lastexitcode -gt 0) {
-	echo "Docker failed to start, exiting."
-
-	$json = "{`"state`": `"failure`", `"description`": `"Jenkins: docker failed to start`", `"target_url`": `"https://storage.googleapis.com/$gcs_bucket/Hyper-V_Windows.txt`", `"context`": `"$env:JOB_NAME`"}"
+# check for Docker CLI
+docker --help
```
medyagh, line 45: I like this ! i think there are placed in actual minikube code we use docker version to check if docker cli is there... (i think) and we could use this trick over there too

medyagh (approved):

medyagh (approved):

PR #15961, opened 2023-03-03
tests: Split up ServiceCmd test into sub-tests and fix on Windows

spowelljr:
## Fix 1
`TestFunctional/parallel/ServiceCmd` recently had a child test added to it (`ServiceJSONOutput`), making `ServiceCmd` a parent. This causes a problem in regard to `gopogh` as `gopogh` ignores test parents so if a child test fails it only marks the child as failed instead of all the parents as well (one test failure vs one + # of parent tests). However, `ServiceCmd` itself has lots of test logic in it itself but these test failures will always be hidden to how `gopogh` works. So to solve this I moved all the testing in `ServiceCmd` into sub-tests. Now `ServiceCmd` is just a caller for a bunch of child tests, now test failures will show as expected.

## Fix 2
`ServiceCmd` had a 100% failure rate on Windows. This was related to the networking constraints Docker Desktop has requiring creating an SSH tunnel to connect to a service. This command has to be left running to keep the SSH tunnel connected so the command keeps running until the user kills it. This causes an issue in testing as we can't keep a command running forever so we need a way to stop it. The existing way to solve this was an odd custom implementation of `context` in the form of the function `killContext`. After the provided timeout it would call `Process.Signal(os.Interrupt)` to kill the command, while this worked on macOS, on Windows it would produce the error `Failed to sent interrupt to proc not supported by windows` and would fail the test. To solve this I simply used `WithTimeout` to create a timeout for the command and let Go handle killing the process.

### DIFF
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -25,6 +25,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


test/integration/functional_test.go
```diff
@@ -1539,16 +1540,10 @@ func validateServiceCmdFormat(ctx context.Context, t *testing.T, profile string)
...
-		endpoint := strings.TrimSpace(rr.Stdout.String())
-		t.Logf("found endpoint for hello-node: %s", endpoint)
-
-		u, err := url.Parse(endpoint)
-		if err != nil {
-			t.Fatalf("failed to parse %q: %v", endpoint, err)
-		}
+		stringIP := strings.TrimSpace(rr.Stdout.String())
 
-		if strings.TrimSpace(rr.Stdout.String()) != u.Hostname() {
```
medyagh: would the hostname always be ip ? or would it be sometimes like svc-name.default.local.svc ?

spowelljr: It's always an IP, it's either `minikube ip` or the IP of the SSH tunnel (which is `127.0.0.1`)

medyagh (approved):
very nice refactor and organizing the test

PR #16001, opened 2023-03-08
wait for pods to get IPs

### LINKED ISSUE #15870: Multi-node tests failing after updating cri-dockerd
https://github.com/kubernetes/minikube/pull/15752 caused `TestMultiNode/serial/DeployApp2Nodes` & `TestMultiNode/serial/PingHostFrom2Pods` to fail on Linux and macOS Docker tests.

Linux test output: https://storage.googleapis.com/minikube-builds/logs/15752/27907/Docker_Linux.html
macOS test output: https://storage.googleapis.com/minikube-builds/logs/15752/27907/Docker_macOS.html

Investigation into the cause will need to take place.

prezha:
fixes #15870

i could replicate the original issue with --driver=docker and --container-runtime=docker on both linux and macos and observed that the busybox pod on the cp node gets ip after ~5sec while the other busybox pod takes significantly longer but eventually (after ~40sec) gets ip on the second node

with this pr we retry until both pods get IPs (max 120sec) so that the multinode tests don't fail because of this delay

the below screenshot from linux (similarly on macos) shows that there's a delay between two pods getting ip - marked with a red rectangle

on a side note, but i think still interesting, both pods were marked as having `Ready` condition while only one had ip, which probably should not happen (based on the [Pod Lifecycle](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions)), but that's potentially a separate topic - marked with amber rectangle on the screenshot, which is the reason why we cannot wait for the pods to become eg `Ready` but retry instead
linux

### DIFF
--- a/test/integration/multinode_test.go
+++ b/test/integration/multinode_test.go
@@ -28,9 +28,11 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
 	"k8s.io/minikube/cmd/minikube/cmd"
 	"k8s.io/minikube/pkg/minikube/config"
+	"k8s.io/minikube/pkg/util/retry"
 )
 
 // TestMultiNode tests all multi node cluster functionality
@@ -487,20 +489,31 @@ func validateDeployAppToMultiNode(ctx context.Context, t *testing.T, profile str
 	}
 
 	// resolve Pod IPs
-	rr, err := Run(t, exec.CommandContext(ctx, Target(), "kubectl", "-p", profile, "--", "get", "pods", "-o", "jsonpath='{.items[*].status.podIP}'"))
-	if err != nil {
-		t.Errorf("failed to retrieve Pod IPs")
+	resolvePodIPs := func() error {
+		rr, err := Run(t, exec.CommandContext(ctx, Target(), "kubectl", "-p", profile, "--", "get", "pods", "-o", "jsonpath='{.items[*].status.podIP}'"))
+		if err != nil {
+			err := fmt.Errorf("failed to retrieve Pod IPs (may be temporary): %v", err)
+			t.Logf(err.Error())
+			return err
+		}
+		podIPs := strings.Split(strings.Trim(rr.Stdout.String(), "'"), " ")
+		if len(podIPs) != 2 {
+			err := fmt.Errorf("expected 2 Pod IPs but got %d (may be temporary), output: %q", len(podIPs), rr.Output())
+			t.Logf(err.Error())
+			return err
+		} else if podIPs[0] == podIPs[1] {
+			err := fmt.Errorf("expected 2 different pod IPs but got %s and %s (may be temporary), output: %q", podIPs[0], podIPs[1], rr.Output())
+			t.Logf(err.Error())
+			return err
+		}
+		return nil
 	}
-	podIPs := strings.Split(strings.Trim(rr.Stdout.String(), "'"), " ")
-	if len(podIPs) != 2 {
-		t.Errorf("expected 2 Pod IPs but got %d, output: %q", len(podIPs), rr.Output())
-	} else if podIPs[0] == podIPs[1] {
-		t.Errorf("expected 2 different pod IPs but got %s and %s. output: %q", podIPs[0], podIPs[1], rr.Output())
-
+	if err := retry.Expo(resolvePodIPs, 1*time.Second, Seconds(120)); err != nil {
+		t.Errorf("failed to resolve pod IPs: %v", err)
 	}
 
 	// get Pod names
-	rr, err = Run(t, exec.CommandContext(ctx, Target(), "kubectl", "-p", profile, "--", "get", "pods", "-o", "jsonpath='{.items[*].metadata.name}'"))
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), "kubectl", "-p", profile, "--", "get", "pods", "-o", "jsonpath='{.items[*].metadata.name}'"))
 	if err != nil {
 		t.Errorf("failed get Pod names")
 	}


medyagh (commented):
thanks for this PR that fixes the failed test, is this something we can add to the minikube code it self as additional Verificaiton ? like if we do minkube start --wait=all would we need to add a new Wait componenet to make sure it waits for all ?

medyagh:
@prezha what is your guess that these tests started ti fail around same time as cri-dockerd change?

https://github.com/Mirantis/cri-dockerd/issues/163

medyagh:
> ok, so following the "side note" from the above, i think i've just confirmed that the issue is not with the cri-dockerd _upgrade_ but actually, probably an issue/bug introduced in kubernetes 1.26.1+, as with 1.26.0 we didn't and don't have that delay issue - ie, both pods get IPs within few seconds and the multinode tests complete w/o issues (and without these pr changes) - as shown in the screenshot below:
> Screenshot 2023-03-08 at 22 04 17
> [k8s-1.26.0-pass.log](https://github.com/kubernetes/minikube/files/10925787/k8s-1.26.0-pass.log)
> [k8s-1.26.1-fail.log](https://github.com/kubernetes/minikube/files/10925788/k8s-1.26.1-fail.log)
> [k8s-1.26.2-fail.log](https://github.com/kubernetes/minikube/files/10925789/k8s-1.26.2-fail.log)

Oh thats interesting. Should we create an issue on k8s repo?

prezha:
@medyagh there could be a couple of things that interfere in a bad way - not sure which one exactly atm, but yeah: the same minikube current _head_ (ie, containing the updated cri-dockerd v0.3.1) _does_ work with k8s v1.26.0 (set via `--kubernetes-version` flag) and _does not_ work with k8s v1.26.1+ (ie, having that delay of assigning ip addresses to pods deployed to non-cp nodes, with k8s v1.26.2 being the current default "not working" the same way) - that could be due to the changes introduced in k8s v1.26.1, a combination of cri-dockerd v0.3.1 and the changes introduced in k8s v1.26.1, and/or one/both of these two in combination with the cni (kindnet in this case)
i'll try to run some more tests to see if we can rule out at least some of these factors to narrow the problem a bit

medyagh:
> 

How about we try a diffrent cni and see if that gets passed the problem of pods getting ips late?

prezha:
@medyagh i've run a number of TestMultiNode tests with docker/docker on different OSs (opensuse tumbleweed, ubuntu 20.04/22.04, macos m1), with different images (busybox, agnhost) and different CNIs (flannel, calico and also older kindnet images) and they all had the similar issue (the difference is only how big the delay is before all pods eventually get an ip, but the test, as it is now, would fail for most of the time - at times it just gets lucky to pass)

tl;dr: _i think that this pr is valid and could help us mitigate this test's flakiness as it explicitly waits for all the test pods to get IPs_ 
(which will eventually happen after some delay)

details:

as mentioned earlier, test pods (ie, busybox) are both marked with "Ready" status even though the pod on the worker node did not get ip - example snapshot:
TestMultinode
which is the reason why our [awaiting](https://github.com/kubernetes/minikube/blob/c4cb107f1d438869c00b4b5f702d8308d4aef9a5/test/integration/multinode_test.go#L484) ends prematurely (ie, [kubectl rollout status](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#-em-status-em-) has _Watch the status of the rollout until it's done_ set to _true_ by default) and then the test fails when the pods' [IPs are checked](https://github.com/kubernetes/minikube/blob/c4cb107f1d438869c00b4b5f702d8308d4aef9a5/test/integration/multinode_test.go#L489-L500) right thereafter

now, based on the aforementioned [Pod Lifecycle/Pod conditions](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions):

> Ready: the Pod is able to `serve requests` and should be added to the load balancing pools of all matching Services.

which i interpret as: if a pod is _Ready_, then it would probably also mean that it has to have an ip to be able to communicate over the network and _serve requests_

if i'm not wrong, kubernetes gets pod/container(s) status (amongst other things) via the kubelet's (generic atm) PLEG module that talks with the cr (cri-docker, in this case)

now, looking at the code, i think that cri-docker separates pod's status from whether it has ip or not, eg:
[`PodSandboxStatus`](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/sandbox_status.go#L26) func and [toRuntimeAPISandboxState](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/convert.go#L250-L259) func (whereas _libdocker.StatusRunningPrefix_ relies purely on having the ["Up*"](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/libdocker/client.go#L36) value in response)

as seen on the snapshot above, _docker can report that container/pod is up/running even though it does not have an ip address_

it then also seems that the kubelet/pleg accepts the status response from the runtime "as-is" and does not check if the Ready pod also has an ip assigned - eg, the [updateCache](https://github.com/kubernetes/kubernetes/blob/fc04e732bb3e7198d2fa44efa5457c7c6f8c0f5b/pkg/kubelet/pleg/generic.go#L430) func calls runtime to [GetsPodStatus](https://github.com/kubernetes/kubernetes/blob/fc04e732bb3e7198d2fa44efa5457c7c6f8c0f5b/pkg/kubelet/pleg/generic.go#L441), then [gets pod's ip(s)](https://github.com/kubernetes/kubernetes/blob/fc04e732bb3e7198d2fa44efa5457c7c6f8c0f5b/pkg/kubelet/pleg/generic.go#L461) (if any, but does not check!) and then just [sets](https://github.com/kubernetes/kubernetes/blob/fc04e732bb3e7198d2fa44efa5457c7c6f8c0f5b/pkg/kubelet/pleg/generic.go#L478) the status as received in runtime's response - from increased-verbosity kubelet log re: snapshot above (level 7, attached "kubelet-m02.log" below in full) - line 2884:

> Mar 19 `20:53:16` multinode-877589-m02 kubelet[1337]: I0319 20:53:16.289327    1337 generic.go:453] "PLEG: Write status" pod="default/busybox-6b86dd6d48-ttnwd" podStatus=&{ID:6fe19c31-5d2c-444e-8562-69f4c31186f4 Name:busybox-6b86dd6d48-ttnwd Namespace:default `IPs:[]` ContainerStatuses:[] SandboxStatuses:[&`PodSandboxStatus`{Id:a926fc57856373944f563b0d8ebc1bd3befda7fe42e00932dc6ae0e8b97d91e7,Metadata:&PodSandboxMetadata{Name:busybox-6b86dd6d48-ttnwd,Uid:6fe19c31-5d2c-444e-8562-69f4c31186f4,Namespace:default,Attempt:0,},`State:SANDBOX_READY`,CreatedAt:1679259195437451087,`Network:&PodSandboxNetworkStatus{Ip:,AdditionalIps:[]*PodIP{}`,},Linux:&LinuxPodSandboxStatus{Namespaces:&Namespace{Options:&NamespaceOption{Network:POD,Pid:CONTAINER,Ipc:POD,TargetId:,UsernsOptions:nil,},},},Labels:map[string]string{app: busybox,io.kubernetes.pod.name: busybox-6b86dd6d48-ttnwd,io.kubernetes.pod.namespace: default,io.kubernetes.pod.uid: 6fe19c31-5d2c-444e-8562-69f4c31186f4,pod-template-hash: 6b86dd6d48,},Annotations:map[string]string{kubernetes.io/config.seen: 2023-03-19T20:53:15.093241411Z,kubernetes.io/config.source: api,},RuntimeHandler:,}] TimeStamp:0001-01-01 00:00:00 +0000 UTC}

now, i'm not sure if Ready should also mean _being able to communicate_ (i think it should), and if so, whose "responsibility" is to check whether the pod has an ip before stating it's finally Ready?

based on the [cri runtime api](https://github.com/kubernetes/cri-api/blob/c75ef5b473bbe2d0a4fc92f82235efd665ea8e9f/pkg/apis/runtime/v1/api.proto#L38-L40):
> RunPodSandbox creates and starts a pod-level sandbox. Runtimes must ensure the sandbox is in the ready state on success.

then, [cri-docker implements](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/sandbox_run.go#L28-L33):
> RunPodSandbox creates and starts a pod-level sandbox. Runtimes should ensure the sandbox is in ready state.
For docker, PodSandbox is implemented by a container holding the network namespace for the pod.

and also this [comment](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/sandbox_run.go#L135-L141):
> Step 5: Setup networking for the sandbox.
All pod networking is setup by a CNI plugin discovered at startup time.
This plugin assigns the pod ip, sets up routes inside the sandbox, creates interfaces etc. In theory, its jurisdiction ends with pod sandbox networking, but it might insert iptables rules or open ports on the host as well, to satisfy parts of the pod spec that aren't recognized by the CNI standard yet.

what we see in cri-docker logs (attached below), is the following `info` message:
> Mar 19 `20:53:15` multinode-877589-m02 cri-dockerd[1059]: time="2023-03-19T20:53:15Z" level=info msg="Failed to read pod IP from plugin/docker: networkPlugin cni failed on the status hook for pod \"busybox-6b86dd6d48-ttnwd_default\": unexpected command output Device \"eth0\" does not exist.\n with error: exit status 1"

that comes from the [getIPs](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/sandbox_helpers.go#L147) func, that's in turn, called by the [`PodSandboxStatus`](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/sandbox_status.go#L53) func highlighted above (notice the one-second difference!) - there's [comment](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/sandbox_helpers.go#L145-L148):

> 	// If all else fails, warn but don't return an error, as pod status
	// should generally not return anything except fatal errors
	logrus.Infof("Failed to read pod IP from plugin/docker: %v", err)

which, i think, then allows the `PodSandboxStatus` to return Ready status to kubelet/PLEG regardless if container's/pod's networking is correctly initialised, as detailed above

finally, from the [cri-docker upgrade changes from v0.3.0 to v0.3.1](https://github.com/Mirantis/cri-dockerd/compare/v0.3.0...v0.3.1), it's not obvious (amongst thousands of lines changed including a bunch of external dependencies) if something is directly changed in the logic that worked before, so it also might be something in one of many libraries that are updated in this upgrade

attached are the kubelet (-v=7)  and cri-docker logs that are related to the snapshot above:
- [kubelet-m02.log](https://github.com/kubernetes/minikube/files/11012631/kubelet-m02.log)
- [cri-docker-m02.log](https://github.com/kubernetes/minikube/files/11012633/cri-docker-m02.log)

bottom line: i might be wrong in the above analysis, but until this is clarified (and fixed somewhere upstream, if needed), this pr should help us reduce the TestMultiNode tests' flakiness (by waiting for test pods to get ip)

medyagh:
> @medyagh i've run a number of TestMultiNode tests with docker/docker on different OSs (opensuse tumbleweed, ubuntu 20.04/22.04, macos m1), with different images (busybox, agnhost) and different CNIs (flannel, calico and also older kindnet images) and they all had the similar issue (the difference is only how big the delay is before all pods eventually get an ip, but the test, as it is now, would fail for most of the time - at times it just gets lucky to pass)
> 
> tl;dr: _i think that this pr is valid and could help us mitigate this test's flakiness as it explicitly waits for all the test pods to get IPs_ (which will eventually happen after some delay)
> 
> details:
> 
> as mentioned earlier, test pods (ie, busybox) are both marked with "Ready" status even though the pod on the worker node did not get ip - example snapshot: TestMultinode which is the reason why our [awaiting](https://github.com/kubernetes/minikube/blob/c4cb107f1d438869c00b4b5f702d8308d4aef9a5/test/integration/multinode_test.go#L484) ends prematurely (ie, [kubectl rollout status](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#-em-status-em-) has _Watch the status of the rollout until it's done_ set to _true_ by default) and then the test fails when the pods' [IPs are checked](https://github.com/kubernetes/minikube/blob/c4cb107f1d438869c00b4b5f702d8308d4aef9a5/test/integration/multinode_test.go#L489-L500) right thereafter
> 
> now, based on the aforementioned [Pod Lifecycle/Pod conditions](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions):
> 
> > Ready: the Pod is able to `serve requests` and should be added to the load balancing pools of all matching Services.
> 
> which i interpret as: if a pod is _Ready_, then it would probably also mean that it has to have an ip to be able to communicate over the network and _serve requests_
> 
> if i'm not wrong, kubernetes gets pod/container(s) status (amongst other things) via the kubelet's (generic atm) PLEG module that talks with the cr (cri-docker, in this case)
> 
> now, looking at the code, i think that cri-docker separates pod's status from whether it has ip or not, eg: [`PodSandboxStatus`](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/sandbox_status.go#L26) func and [toRuntimeAPISandboxState](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/convert.go#L250-L259) func (whereas _libdocker.StatusRunningPrefix_ relies purely on having the ["Up*"](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/libdocker/client.go#L36) value in response)
> 
> as seen on the snapshot above, _docker can report that container/pod is up/running even though it does not have an ip address_
> 
> it then also seems that the kubelet/pleg accepts the status response from the runtime "as-is" and does not check if the Ready pod also has an ip assigned - eg, the [updateCache](https://github.com/kubernetes/kubernetes/blob/fc04e732bb3e7198d2fa44efa5457c7c6f8c0f5b/pkg/kubelet/pleg/generic.go#L430) func calls runtime to [GetsPodStatus](https://github.com/kubernetes/kubernetes/blob/fc04e732bb3e7198d2fa44efa5457c7c6f8c0f5b/pkg/kubelet/pleg/generic.go#L441), then [gets pod's ip(s)](https://github.com/kubernetes/kubernetes/blob/fc04e732bb3e7198d2fa44efa5457c7c6f8c0f5b/pkg/kubelet/pleg/generic.go#L461) (if any, but does not check!) and then just [sets](https://github.com/kubernetes/kubernetes/blob/fc04e732bb3e7198d2fa44efa5457c7c6f8c0f5b/pkg/kubelet/pleg/generic.go#L478) the status as received in runtime's response - from increased-verbosity kubelet log re: snapshot above (level 7, attached "kubelet-m02.log" below in full) - line 2884:
> 
> > Mar 19 `20:53:16` multinode-877589-m02 kubelet[1337]: I0319 20:53:16.289327    1337 generic.go:453] "PLEG: Write status" pod="default/busybox-6b86dd6d48-ttnwd" podStatus=&{ID:6fe19c31-5d2c-444e-8562-69f4c31186f4 Name:busybox-6b86dd6d48-ttnwd Namespace:default `IPs:[]` ContainerStatuses:[] SandboxStatuses:[&`PodSandboxStatus`{Id:a926fc57856373944f563b0d8ebc1bd3befda7fe42e00932dc6ae0e8b97d91e7,Metadata:&PodSandboxMetadata{Name:busybox-6b86dd6d48-ttnwd,Uid:6fe19c31-5d2c-444e-8562-69f4c31186f4,Namespace:default,Attempt:0,},`State:SANDBOX_READY`,CreatedAt:1679259195437451087,`Network:&PodSandboxNetworkStatus{Ip:,AdditionalIps:[]*PodIP{}`,},Linux:&LinuxPodSandboxStatus{Namespaces:&Namespace{Options:&NamespaceOption{Network:POD,Pid:CONTAINER,Ipc:POD,TargetId:,UsernsOptions:nil,},},},Labels:map[string]string{app: busybox,io.kubernetes.pod.name: busybox-6b86dd6d48-ttnwd,io.kubernetes.pod.namespace: default,io.kubernetes.pod.uid: 6fe19c31-5d2c-444e-8562-69f4c31186f4,pod-template-hash: 6b86dd6d48,},Annotations:map[string]string{kubernetes.io/config.seen: 2023-03-19T20:53:15.093241411Z,kubernetes.io/config.source: api,},RuntimeHandler:,}] TimeStamp:0001-01-01 00:00:00 +0000 UTC}
> 
> now, i'm not sure if Ready should also mean _being able to communicate_ (i think it should), and if so, whose "responsibility" is to check whether the pod has an ip before stating it's finally Ready?
> 
> based on the [cri runtime api](https://github.com/kubernetes/cri-api/blob/c75ef5b473bbe2d0a4fc92f82235efd665ea8e9f/pkg/apis/runtime/v1/api.proto#L38-L40):
> 
> > RunPodSandbox creates and starts a pod-level sandbox. Runtimes must ensure the sandbox is in the ready state on success.
> 
> then, [cri-docker implements](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/sandbox_run.go#L28-L33):
> 
> > RunPodSandbox creates and starts a pod-level sandbox. Runtimes should ensure the sandbox is in ready state.
> > For docker, PodSandbox is implemented by a container holding the network namespace for the pod.
> 
> and also this [comment](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/sandbox_run.go#L135-L141):
> 
> > Step 5: Setup networking for the sandbox.
> > All pod networking is setup by a CNI plugin discovered at startup time.
> > This plugin assigns the pod ip, sets up routes inside the sandbox, creates interfaces etc. In theory, its jurisdiction ends with pod sandbox networking, but it might insert iptables rules or open ports on the host as well, to satisfy parts of the pod spec that aren't recognized by the CNI standard yet.
> 
> what we see in cri-docker logs (attached below), is the following `info` message:
> 
> > Mar 19 `20:53:15` multinode-877589-m02 cri-dockerd[1059]: time="2023-03-19T20:53:15Z" level=info msg="Failed to read pod IP from plugin/docker: networkPlugin cni failed on the status hook for pod "busybox-6b86dd6d48-ttnwd_default": unexpected command output Device "eth0" does not exist.\n with error: exit status 1"
> 
> that comes from the [getIPs](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/sandbox_helpers.go#L147) func, that's in turn, called by the [`PodSandboxStatus`](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/sandbox_status.go#L53) func highlighted above (notice the one-second difference!) - there's [comment](https://github.com/Mirantis/cri-dockerd/blob/6daf9ac0f630da59ac1ab6f2af8293793609f22e/core/sandbox_helpers.go#L145-L148):
> 
> > // If all else fails, warn but don't return an error, as pod status
> > // should generally not return anything except fatal errors
> > logrus.Infof("Failed to read pod IP from plugin/docker: %v", err)
> 
> which, i think, then allows the `PodSandboxStatus` to return Ready status to kubelet/PLEG regardless if container's/pod's networking is correctly initialised, as detailed above
> 
> finally, from the [cri-docker upgrade changes from v0.3.0 to v0.3.1](https://github.com/Mirantis/cri-dockerd/compare/v0.3.0...v0.3.1), it's not obvious (amongst thousands of lines changed including a bunch of external dependencies) if something is directly changed in the logic that worked before, so it also might be something in one of many libraries that are updated in this upgrade
> 
> attached are the kubelet (-v=7) and cri-docker logs that are related to the snapshot above:
> 
> * [kubelet-m02.log](https://github.com/kubernetes/minikube/files/11012631/kubelet-m02.log)
> * [cri-docker-m02.log](https://github.com/kubernetes/minikube/files/11012633/cri-docker-m02.log)
> 
> bottom line: i might be wrong in the above analysis, but until this is clarified (and fixed somewhere upstream, if needed), this pr should help us reduce the TestMultiNode tests' flakiness (by waiting for test pods to get ip)

Thank you for the analysis and report on this, would you mind sharing this investigation in this issue for cri-dockerd in a format that they could possibily benefit from it ?

https://github.com/Mirantis/cri-dockerd/issues/163

prezha:
> Thank you for the analysis and report on this, would you mind sharing this investigation in this issue for cri-dockerd in a format that they could possibily benefit from it ?
> 
> [Mirantis/cri-dockerd#163](https://github.com/Mirantis/cri-dockerd/issues/163)

sure, @medyagh, i've tried to provide some details in the [comment](https://github.com/Mirantis/cri-dockerd/issues/163#issuecomment-1481947298) there

PR #16010, opened 2023-03-09
test: Fix having to change test byte size on new addon

spowelljr:
Every time someone would add a new test this test fails as we make the byte slice just big enough to fit the JSON, but a new addon increases this size of the JSON and the test fails.

Changed this to use a JSON decoder which has a cleaner implementation and prevents unnecessary changes and test failures.


medyagh (approved):

medyagh:
unit test failing

spowelljr:
> unit test failing

The test passed fine on my Mac but failed on Linux, I changed it to use a JSON decoder which resolves the issue and is a cleaner implementation.

PR #16038, opened 2023-03-13
feat: implement a prototype for sudominikube

### LINKED ISSUE #15847: Evaluate adding miniube to sudoers group only for Commands that requires sudo
### What Happened?

the idea is, by default minikube run as normal user (without sudo)

whenever minikube needs to run a feature that needs sudo (such as tunnel on VM) it would ask for users password once, and use that to add minikube binary to sudoers list , and after that every other minikube tunnel command will run without asking for passwor.d

sonething like this
https://github.com/lima-vm/socket_vmnet/blob/master/etc_sudoers.d/socket_vmnet

### Attach the log file

n/a

### Operating System

None

### Driver

None

ComradeProgrammer:
fix #15847

This draft pr contains the following things:
- it adds an executable called "sudominikube". Current it support "tunnel" command
- it adds an a command called "minikube sudominikube install", which install sudominikube to /opt/minikube/bin, add a user group called "minikube", add current user to group 'minikube', and add sudominikube to sudoers.d /etc/sudoers.d.
- it modifies the logic of "minikube tunnel" command. It will try to detect whether sudominikube exists, and if so, it will try to use `sudo sudominikube tunnel` to start a tunnel, so the user won't need to enter a password

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,6 +220,14 @@ else
 	$(Q)go build $(MINIKUBE_GOFLAGS) -tags "$(MINIKUBE_BUILD_TAGS)" -ldflags="$(MINIKUBE_LDFLAGS)" -o $@ k8s.io/minikube/cmd/minikube
 endif
 
+out/sudominikube$(IS_EXE): $(SOURCE_FILES) $(ASSET_FILES) go.mod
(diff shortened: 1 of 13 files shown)


ComradeProgrammer:
@spowelljr  @afbjorklund Could you please have a look at this PR?

PR #16074, opened 2023-03-17
CNI: Update flannel from v0.20.2 to v0.21.4 and change repository

zhengtianbao:
flannel has moved its images to: docker.io we should switch to docker.io/flannel and update to the latest version.
stay up to date with upstream:  https://github.com/flannel-io/flannel/blob/master/Documentation/kube-flannel.yml

### DIFF
--- a/pkg/minikube/cni/flannel.yaml
+++ b/pkg/minikube/cni/flannel.yaml
@@ -131,7 +131,7 @@ spec:
       initContainers:
       - name: install-cni-plugin
         image: docker.io/flannel/flannel-cni-plugin:v1.1.2
-        #image: docker.io/rancher/mirrored-flannelcni-flannel-cni-plugin:v1.1.2
+       #image: docker.io/rancher/mirrored-flannelcni-flannel-cni-plugin:v1.1.2
         command:
         - cp
         args:
@@ -142,8 +142,8 @@ spec:
         - name: cni-plugin
           mountPath: /opt/cni/bin
       - name: install-cni
-        image: docker.io/flannel/flannel:v0.21.3
-        #image: docker.io/rancher/mirrored-flannelcni-flannel:v0.21.3
+        image: docker.io/flannel/flannel:v0.21.4
+       #image: docker.io/rancher/mirrored-flannelcni-flannel:v0.21.4
         command:
         - cp
         args:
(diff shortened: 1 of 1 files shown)


pkg/minikube/cni/flannel.go
```diff
@@ -51,12 +50,6 @@ func (c Flannel) String() string {
 
 // Apply enables the CNI
 func (c Flannel) Apply(r Runner) error {
-	// Mostly applicable to the 'none' driver
-	_, err := r.RunCmd(exec.Command("stat", "/opt/cni/bin/portmap"))
-	if err != nil {
-		return errors.Wrap(err, "required 'portmap' CNI plug-in not found")
```
spowelljr, line 57: Is `portmap` no longer required for flannel?

afbjorklund: It is still required, as part of the CNI config

pkg/minikube/cni/flannel.yaml
```diff
@@ -168,9 +168,6 @@ spec:
           requests:
             cpu: "100m"
             memory: "50Mi"
-          limits:
```
afbjorklund, line 171: This change also seems unrelated.

afbjorklund (changes requested):

pkg/minikube/cni/flannel.yaml
afbjorklund: It is from upstream ([30b56a37](https://github.com/flannel-io/flannel/commit/30b56a3794737135bf69d983417dcaaf3f1798b8)), but should probably be mentioned separately.

pkg/minikube/cni/flannel.go
zhengtianbao: ok, the error may occur, keep these lines.

pkg/minikube/cni/flannel.yaml
zhengtianbao: reverted.

pkg/minikube/cni/flannel.yaml
```diff
@@ -130,8 +130,8 @@ spec:
       serviceAccountName: flannel
       initContainers:
       - name: install-cni-plugin
-        #image: flannelcni/flannel-cni-plugin:v1.1.2 #for ppc64le and mips64le (dockerhub limitations may apply)
-        image: docker.io/rancher/mirrored-flannelcni-flannel-cni-plugin:v1.1.2
+        image: docker.io/flannel/flannel-cni-plugin:v1.1.2
+        #image: docker.io/rancher/mirrored-flannelcni-flannel-cni-plugin:v1.1.2
```
medyagh, line 134: why keeping the old one in comment ? we could remove the un-used deprecated image

medyagh (changes requested):

pkg/minikube/cni/flannel.yaml
spowelljr: I believe it's commented in the upstream file, seems like a backup if the main one is down

spowelljr (approved):

PR #16127, opened 2023-03-22
Fix Typo in German translation

coffemakingtoaster:
Change spelling of the German word for consideration:

"Erwähgung" -> "Erwägung"

As verified by the biggest/official German dictionary https://www.duden.de/rechtschreibung/Erwaegung


spowelljr (approved):
Thanks for the PR!

Issue #16140, opened 2023-03-23
Storage provisioner does not clear data on PVC deleted


vanhtuan0409:
### What Happened?

Minikube storage provisioner does not delete data folder when PVC deleted. This only happened after minikube got restart

Step to reproduce:
- Start minikube
- Create a PVC
- Restart minikube (stop then start again)
- Remove PVC

Root cause is because storage provisioned generate a random UUID on start as identity. This will change after every restart
https://github.com/kubernetes/minikube/blob/master/pkg/storage/storage_provisioner.go#L52

### Attach the log file

```
I0323 09:48:55.972215       1 controller.go:1478] delete "pvc-b3aafdb8-1b59-49a1-a0a6-732bcff0a69f": volume deletion ignored: ignored because identity annotation on PV does not match ours
```

### Operating System

Ubuntu

### Driver

Docker

PR #16141, opened 2023-03-23
fix: storage provision identity got changed after every restart

vanhtuan0409:
fix https://github.com/kubernetes/minikube/issues/16140

Before:

Storage privisioner's identity got random generated at the start. Which will be changed after every restart

After:

Pass in provisioner's identity by environment variable. This will work in single node setup but will not work well for multi node

### DIFF
--- a/pkg/storage/storage_provisioner.go
+++ b/pkg/storage/storage_provisioner.go
@@ -26,8 +26,6 @@ import (
 
 	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
@@ -42,14 +40,18 @@ type hostPathProvisioner struct {
 
 	// Identity of this hostPathProvisioner, generated. Used to identify "this"
 	// provisioner's PVs.
-	identity types.UID
+	identity string
 }
 
 // NewHostPathProvisioner creates a new Provisioner using host paths
 func NewHostPathProvisioner(pvDir string) controller.Provisioner {
+	nodeName := os.Getenv("NODE_NAME")
+	if nodeName == "" {
+		klog.Fatal("env variable NODE_NAME must be set so that this provisioner can identify itself")
+	}
 	return &hostPathProvisioner{
 		pvDir:    pvDir,
-		identity: uuid.NewUUID(),
+		identity: nodeName,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


medyagh (commented):
hi @vanhtuan0409 would you please post before/after this PR in the PR description ?

vanhtuan0409:
Hi @medyagh , I added before and after in the description

This is only partial fix as it still not work well in multi node setup

medyagh:
> Hi @medyagh , I added before and after in the description
> 
> This is only partial fix as it still not work well in multi node setup

did it work in multinode before this PR ? how about instead of random we use the minikube profile name ? 


do they have to be unique name for each node ? if yes then you could use the machine name
currently minikube nodes are named proilfe-m0 m1... m2...

vanhtuan0409:
As you can see from this line. The provisioner checks for identity before perform deleting the data folder. If the identity got changed compared to when to PVC was created, the provisioner would skip deleting and return an error
https://github.com/kubernetes/minikube/blob/master/pkg/storage/storage_provisioner.go#L103

The current master branch does not work on single-node and multi-node setups. We don't need different uniq name for each node. But for multi-node setup to work, you need to locate which node the PVC was created in and delete the data folder in exactly that node. It will be a complete redesign I think

Using the machine name is also good, but I don't know how to pass the machine name to the provisioner yet

medyagh:
> As you can see from this line. The provisioner checks for identity before perform deleting the data folder. If the identity got changed compared to when to PVC was created, the provisioner would skip deleting and return an error https://github.com/kubernetes/minikube/blob/master/pkg/storage/storage_provisioner.go#L103
> 
> The current master branch does not work on single-node and multi-node setups. We don't need different uniq name for each node. But for multi-node setup to work, you need to locate which node the PVC was created in and delete the data folder in exactly that node. It will be a complete redesign I think
> 
> Using the machine name is also good, but I don't know how to pass the machine name to the provisioner yet

If you look at multi node code u can find out

maybe check the code that implements 
minikube node list

vanhtuan0409:
storage provisioner is deployed into minikube as a deployment. It is not a part of minikube cli. So I don't think it will be able to access to profile name

My current approach is to passed in environment. AFAIK, minikube node name is also based on profile name
https://github.com/kubernetes/minikube/pull/16141/files#diff-89dd13ca41e2948be16d9c0f163009b685a3536b234be0c9d482253eff5036a3R106

pkg/storage/storage_provisioner.go
```diff
@@ -42,14 +40,18 @@ type hostPathProvisioner struct {
 
 	// Identity of this hostPathProvisioner, generated. Used to identify "this"
 	// provisioner's PVs.
-	identity types.UID
+	identity string
 }
 
 // NewHostPathProvisioner creates a new Provisioner using host paths
 func NewHostPathProvisioner(pvDir string) controller.Provisioner {
+	nodeName := os.Getenv("NODE_NAME")
```
medyagh, line 48: who sets this environment variable ? and I dont see it being set anywhere ? 

also the name should be unique to minikube to not collide withe other tools, such as
MINIKUBE_STORAGE_PROVISIONER_NAME...or something like that

medyagh (changes requested):

medyagh (changes requested):
do you mind pasting the Before After this PR output with a full example describing the problem and example showing solving the probem?
[vanhtuan0409](https://github.com/vanhtuan0409)

medyagh:
@vanhtuan0409 thank you for your patience on this PR review time, do you mind adding an Example of Before and After?
by before after I mean, provide an example with minikube ouptput Before and After your PR

PR #16143, opened 2023-03-23
CNI/bridge: Ensure pod communications are allowed

maxbrunet:
Using `none` driver with `docker` runtime, the `bridge` CNI is selected since Minikube v1.29.0, but pods cannot communicate, this PR adds the `firewall` plugin to the CNI config to set up the correct `iptables` rules.

I found the solution here: https://github.com/containernetworking/plugins/issues/430#issuecomment-567917273

The config for reference:
* https://github.com/containernetworking/plugins/blob/v1.2.0/plugins/meta/firewall/firewall.go#L33-L68
* https://www.cni.dev/plugins/current/meta/firewall/

Tested and working for my use case, but I am not entirely sure about all the implications of this change.

Fixes medyagh/setup-minikube#110

### DIFF
--- a/pkg/minikube/cni/bridge.go
+++ b/pkg/minikube/cni/bridge.go
@@ -30,12 +30,13 @@ import (
 // bridge is what minikube defaulted to when `--enable-default-cni=true`
 // ref: https://www.cni.dev/plugins/current/main/bridge/
 // ref: https://www.cni.dev/plugins/current/meta/portmap/
+// ref: https://www.cni.dev/plugins/current/meta/firewall/
 
 // note: "cannot set hairpin mode and promiscuous mode at the same time"
 // ref: https://github.com/containernetworking/plugins/blob/7e9ada51e751740541969e1ea5a803cbf45adcf3/plugins/main/bridge/bridge.go#L424
 var bridgeConf = template.Must(template.New("bridge").Parse(`
 {
-  "cniVersion": "0.3.1",
+  "cniVersion": "0.4.0",
   "name": "bridge",
   "plugins": [
     {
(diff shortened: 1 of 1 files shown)


spowelljr:
That PR makes sense to me to get the cri-dockerd logs, if you don't see system logs at the end of the failed test it likely doesn't have the following at the start of the test func.

`defer PostMortemLogs(t, profile)`

This will produce the following: https://storage.googleapis.com/minikube-builds/logs/16143/28462/Docker_Linux.html#fail_TestPause%2fserial%2fSecondStartNoReconfiguration

If you look at the bottom of the logs you should see `==> Docker <==` and that's where the logs would be in your other PR

maxbrunet:
So, my first attempt was using the most restrictive ingress policy, `same-bridge`, but it appears to break to many tests. I have removed it to use the default `open` instead, it looks much better.

https://github.com/containernetworking/plugins/blob/v1.2.0/plugins/meta/firewall/firewall.go#L59-L67
https://www.cni.dev/plugins/current/meta/firewall/#ingress-policy

The config seems to load properly everywhere as I have not seen any error in the `cri-dockerd` logs.

@spowelljr what do you think about the solution? I cannot say for sure if the remaining failing tests are related or not, but in this state, `firewall` only adds a FORWARD chain and no isolation/DROP rule, it should be safe: https://www.cni.dev/plugins/current/meta/firewall/#forward

maxbrunet:
@medyagh @spowelljr friendly nudge :slightly_smiling_face:

spowelljr (approved):
Sorry for the delay, the tests look good, thank you for the PR!

PR #16194, opened 2023-03-29
Doc fix: use LANGUAGE instead of LC_ALL to change minikube output language

jmloyola:
### What type of PR is this?

/kind documentation

The [PR #7550](https://github.com/kubernetes/minikube/pull/7550) replaced the use of `LANG` in the translation documentation with `LC_ALL` because it has a higher priority.
This correctly changed the output language of the program, except when the `LANGUAGE` is already set.

Since the environment variable `LANGUAGE` has a higher priority than `LC_ALL`, we should use the former to change the minikube output language.
https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html

Only when `LC_ALL=C` the `LANGUAGE` environment variable is ignored.
https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html

Although this behavior is yet to be implemented in the corresponding Go package. https://github.com/Xuanwo/go-locale
https://github.com/Xuanwo/go-locale/issues/50

Yes.

Before this PR:
```bash
$ env | grep -P 'LC_.*|LANG.*'                                                                                    
LANG=es_AR.UTF-8                                                                                                                                                      
LANGUAGE=es_AR:es

$ LC_ALL=en ./out/minikube-linux-amd64 start
😄  minikube v1.29.0 en Ubuntu 18.04
✨  Controlador docker seleccionado automáticamente. Otras opciones: kvm2, ssh
```

After this PR:
```bash
$ env | grep -P 'LC_.*|LANG.*'                                                                                    
LANG=es_AR.UTF-8                                                                                                                                                      
LANGUAGE=es_AR:es

$ LANGUAGE=en ./out/minikube-linux-amd64 start
😄  minikube v1.29.0 on Ubuntu 18.04
✨  Automatically selected the docker driver. Other choices: kvm2, ssh
```

### DIFF
--- a/site/content/en/docs/contrib/translations.md
+++ b/site/content/en/docs/contrib/translations.md
@@ -92,15 +92,23 @@ All translations are stored in the top-level `translations` directory.
 
 * You now have a fresh minikube binary in the `out` directory. If your system locale is that of the language you added translations for, a simple `out/minikube start` will work as a test, assuming you translated phrases from `minikube start`. You can use whatever command you'd like in that way.
 
-* If you have a different system locale, you can override the printed language using the LC_ALL environment variable:
+* If you have a different system locale, you can override the printed language using the LANGUAGE environment variable:
 
  ```
- ~/minikube$ LC_ALL=fr out/minikube start
- 😄  minikube v1.9.2 sur Darwin 10.14.5
- ✨  Choix automatique du driver hyperkit. Autres choix: docker
+ ~/minikube$ LANGUAGE=fr out/minikube start
+ 😄  minikube v1.29.0 sur Ubuntu 18.04
+ ✨  Choix automatique du pilote docker. Autres choix: kvm2, ssh
+ 📌  Utilisation du pilote Docker avec le privilège root
  👍  Démarrage du noeud de plan de contrôle minikube dans le cluster minikube
- 🔥  Création de VM hyperkit (CPUs=2, Mémoire=4000MB, Disque=20000MB)...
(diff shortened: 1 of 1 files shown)


medyagh (commented):
@jmloyola thank you for this PR, do you mind sharing what OS did you try this on ? is it same on windows and mac and linux ?

jmloyola:
I'm on Ubuntu 18.04 and I haven't tested this on Windows or Mac because I don't have the hardware.
If anyone can test it, it'd be great.

In case this doesn't work for all operating systems, we could show a [note](https://kubernetes.io/docs/contribute/style/style-guide/#note) in the documentation to notify users that may be hit by this issue.

jmloyola:
I don't believe the `LANGUAGE` environment variable is set by default either. But I don't know which steps of the OS installation, applications or desktop environment set it.

Either way, it is similar to `LC_ALL` in the sense that both are not always set, but when they are, you need to set them to the language you want the output.

Here is more information about the `LANGUAGE` environment variable from the Ubuntu documentation:

> The "LANGUAGE" environment variable, which is set by the GUIs more often than not on Ubuntu desktops, controls language for message and menu display for GNU compatible applications. For such applications it overrides whatever locale names are set in "LANG" and "LC_MESSAGES".

[source](https://help.ubuntu.com/community/EnvironmentVariables#The_LANGUAGE_priority_list)

spowelljr (approved):
Thanks for the PR, looks good to me

PR #16124, opened 2023-03-21
Remove container-runtime flag from later versions

afbjorklund:
The `--container-runtime` flag was deprecated in k8s 1.24, and removed in k8s 1.27

Add the extra `unix://` prefix everywhere, and remove the redundant endpoint config.

Closes #16112

PR #16205, opened 2023-03-31
Refactor kubelet options

spowelljr:
Follow up to https://github.com/kubernetes/minikube/pull/16124

Refactored the duplicated `KubeletOptions` into a common func

### DIFF
--- a/pkg/minikube/cruntime/cruntime.go
+++ b/pkg/minikube/cruntime/cruntime.go
@@ -339,3 +339,14 @@ func CheckKernelCompatibility(cr CommandRunner, major, minor int) error {
 	}
 	return nil
 }
+
+// defaultKubeletOptions returns the default kubelet options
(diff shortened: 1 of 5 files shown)


pkg/minikube/cruntime/cruntime.go
```diff
@@ -339,3 +339,14 @@ func CheckKernelCompatibility(cr CommandRunner, major, minor int) error {
 	}
 	return nil
 }
+
+// defaultKubeletOptions returns the default kubelet options
+func defaultKubeletOptions(cr Manager, kubernetesVersion semver.Version) map[string]string {
```
medyagh, line 344: add in comments that retunrs the default container-runtime-endpoint for kublet

medyagh (changes requested):

pkg/minikube/cruntime/cruntime.go
spowelljr: Updated func name and comment

medyagh (approved):

PR #16232, opened 2023-04-04
Add PowerShell code completion

oldium:


go.mod
```diff
@@ -51,7 +51,7 @@ require (
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/samalba/dockerclient v0.0.0-20160414174713-91d7393ff859 // indirect
 	github.com/shirou/gopsutil/v3 v3.23.2
-	github.com/spf13/cobra v1.6.1
+	github.com/spf13/cobra v1.6.1-0.20230322224151-4dd4b25de384
```
spowelljr, line 54: They released v1.7.0 which includes the change you mentioned, please update to that instead

spowelljr (changes requested):

go.mod
oldium: Yes, I will do it, they made a release when I pulled my changes 😁

oldium:
The build error in GitHub CI/CD pipeline is caused by [checksum problem in gopogh v0.15.0 package](https://github.com/medyagh/gopogh/issues/46).

spowelljr:
> The build error in GitHub CI/CD pipeline is caused by [checksum problem in gopogh v0.15.0 package](https://github.com/medyagh/gopogh/issues/46).

The gopogh version was bumped and merged, could you rebase

https://github.com/kubernetes/minikube/pull/16234

oldium:
Rebased

oldium:
On another note, if you @spowelljr have some time, please - I am also trying to get my [other PR](https://github.com/kubernetes/minikube/pull/14868) merged to fix the `--profile` to `--cluster` argument translation in `minikube kubectl __complete` code-completion call. The last one looking at it (I also pointed him to my PR aftewards) was Ben Krieger with his [PR](https://github.com/kubernetes/minikube/pull/15850), which unfortunately completely removed the `--cluster` argument from the code completion call.

cmd/minikube/cmd/completion.go
```diff
@@ -47,6 +47,16 @@ const longDescription = `Outputs minikube shell completion for the given shell (
 
 	Additionally, you may want to output the completion to a file and source in your .bashrc
 
+	Windows:
+		## Save completion code to a script and execute in the profile
+		PS> minikube completion powershell > $HOME\.minikube-completion.ps1
+		PS> Add-Content $PROFILE ". ` + "`" + `$HOME\.minikube-completion.ps1"
```
spowelljr, line 53: I tried this command but it gives me a newline like the command is not complete then I enter the command
```
> Add-Content $PROFILE ". ` + "`" + `$HOME\.minikube-completion.ps1"
>>
```
I think the quote escaping might be slightly off

spowelljr (changes requested):

cmd/minikube/cmd/completion.go
oldium: You are inside Golang back-tick block. The actual output when run with `minikube completion --help` is the following:

```
        Windows:
                ## Save completion code to a script and execute in the profile
                PS> minikube completion powershell > $HOME\.minikube-completion.ps1
                PS> Add-Content $PROFILE ". `$HOME\.minikube-completion.ps1"

                ## Execute completion code in the profile
                PS> Add-Content $PROFILE "if (Get-Command minikube -ErrorAction SilentlyContinue) {
                        minikube completion powershell | Out-String | Invoke-Expression
                    }"
```

oldium: The actual line added to your $PROFILE should be dot-sourcing of the generated file (tested right now):

```
. $HOME\.minikube-completion.ps1
```

oldium: The same effect could be achieved by using non-interpreting single-quotes `Add-Content $PROFILE '. $HOME\.minikube-completion.ps1'`, which does not use the PowerShell back-tick escaping and is probably more readable. I will change it.

spowelljr: Ah, didn't realize, thanks for the clarification

oldium:
Updated to use single-quotes, which is more readable.

spowelljr (changes requested):
Can you run `make generate-docs` to update all the docs and push the changes up

oldium:
Done

spowelljr (approved):
Thanks for the PR! I'll don't know a ton about auto completion but I'll take a look at our other PR on Monday.

Issue #16201, opened 2023-03-30
Expired certs not being renewed


Kimi450:
### What Happened?

This was supposed fixed in [issue 10122](https://github.com/kubernetes/minikube/issues/10122) but I dont think this is fixed. Or at least not as I would expect it to be fixed. Ive linked the issue from my repo here. You can check my https://github.com/Kimi450/ubuntu_server/issues/4#issuecomment-1483833790on that issue which describes why I think its broken (and how to replicate it). Unless im doing something wrong, I think this issue still persists.

Basically, if you use --cert-expiration tag, that only affects

```
/var/lib/minikube/certs/proxy-client.crt
/var/lib/minikube/certs/apiserver.crt

# these are on my mounted ones, minikube start --mount --mount-string "/home/kimi450:/minikube-host" --cert-expiration="120s"
/minikube-host/.minikube/profiles/minikube/proxy-client.crt
/minikube-host/.minikube/profiles/minikube/client.crt
/minikube-host/.minikube/profiles/minikube/apiserver.crt
```
and the problematic (not renewed) ones are

```
/var/lib/minikube/certs/etcd/healthcheck-client.crt
/var/lib/minikube/certs/etcd/server.crt
/var/lib/minikube/certs/etcd/peer.crt
/var/lib/minikube/certs/apiserver-etcd-client.crt
/var/lib/minikube/certs/apiserver-kubelet-client.crt
/var/lib/minikube/certs/front-proxy-client.crt
```
Without these being renewed, the minikube instance cannot be accessed (unless you change the system time to a date in the past)

minikube version

```
minikube version: v1.29.0
commit: ddac20b4b34a9c8c857fc602203b6ba2679794d3
```

Reproduce by doing the following
- install minikube (no special setting required)
- change system time
    ```
    timedatectl
    timedatectl set-ntp no

    # some time in the future, more than a year 
    timedatectl set-time 2025-08-10
    ```

- restart minikube (optional)
    ```
    minikube stop
    minikube start
    ```

### Attach the log file

I do not have this at the moment, but I can provide later if absolutely necessary

### Operating System

Ubuntu

### Driver

Docker

spowelljr:
Thanks for the information @Kimi450, we occasionally got issues related to certs after we thought it was fixed, but this confirms why it's failing.

Kimi450:
> Thanks for the information @Kimi450, we occasionally got issues related to certs after we thought it was fixed, but this confirms why it's failing.

[...]

PR #16249, opened 2023-04-05
Renew expired kubeadm certs

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/16201

Explanation for below commands in before and after:
```
$ sudo timedatectl set-time 2024-03-03
$ minikube ssh -- sudo service kubelet restart
```
When testing my changes I ran into the kubelet certs being expired. I found the following comment https://github.com/kubernetes/kubeadm/issues/2054#issuecomment-595318438:
> normally this will not happen because the kubelet certificate manager has logic to monitor for certificate expiration and it will auto-rotate your client certificates once ~70% of the expiration period is reached. by forcing a date you are bypassing this mechanism and you end up with a certificate which was not rotated.

Manually forcing the date forward beyond kubelets certs expiration date was resulting in the certs not being rotated at all, so set the time less than a year, restart the kubelet service to force the certs rotation, then move time forward to make kubeadm certs be expired, but not far enough to make the kubelet certs expired. This would be handled for us if a cluster was running for a year.

```
# start minikube
# kubeadm expiration: 2024-04-07 kubelet expiration: 2024-04-07
# set time to 2024-03-03
# refresh kubelet certs
# kubeadm expiration: 2024-04-07 kubelet expiration: 2025-03-03
# set time to 2025-02-02
# kubeadm expiration: 2024-04-07 (EXPIRED) kubelet expiration: 2025-03-03 (VALID)
```

**Before:**
```
$ minikube start
$ sudo timedatectl set-ntp no
$ sudo timedatectl set-time 2024-03-03
$ minikube ssh -- sudo service kubelet restart
$ minikube stop
$ sudo timedatectl set-time 2025-02-02
$ minikube start
...
W0810 00:00:52.022507    4629 certs.go:521] WARNING: could not validate bounds for certificate apiserver-kubelet-client: the certificate has expired: NotBefore: 2023-04-04 17:14:07 +0000 UTC, NotAfter: 2024-04-04 17:14:09 +0000 UTC
error execution phase certs/apiserver-kubelet-client: [certs] certificate apiserver-kubelet-client not signed by CA certificate ca: x509: certificate has expired or is not yet valid: current time 2025-08-10T00:00:52Z is after 2024-04-04T17:14:09Z


```

**After:**
```
$ minikube start
$ sudo timedatectl set-ntp no
$ sudo timedatectl set-time 2024-03-03
$ minikube ssh -- sudo service kubelet restart
$ minikube stop
$ sudo timedatectl set-time 2025-02-02
$ minikube start
😄  minikube v1.30.1 on Debian 11.6 (amd64)
🚜  Pulling base image ...
❗  This container is having trouble accessing https://registry.k8s.io
🐳  Preparing Kubernetes v1.26.3 on Docker 23.0.2 ...
❗  kubeadm certificates have expired. Generating new ones...
```

```
❗  This container is having trouble accessing https://registry.k8s.io
```
Is due to SSL certs being expired with forced time change

### DIFF
--- a/pkg/minikube/bootstrapper/certs.go
+++ b/pkg/minikube/bootstrapper/certs.go
@@ -349,14 +349,12 @@ func generateProfileCerts(cfg config.ClusterConfig, n config.Node, ccs CACerts, 
 }
 
 func generateKubeadmCerts(cmd command.Runner, cc config.ClusterConfig) error {
-	gt, err := guestTime(cmd)
-	if err != nil {
-		return err
+	if _, err := cmd.RunCmd(exec.Command("ls", path.Join(vmpath.GuestPersistentDir, "certs", "etcd"))); err != nil {
+		klog.Infof("certs directory doesn't exist, likely first start: %v", err)
+		return nil
 	}
-	// add one day to the time so if the certs expire in less than a day we regenerate
-	gt = gt.Add(24 * time.Hour)
 
-	needsRefresh := false
+	expiredCerts := false
 	certs := []string{"apiserver-etcd-client", "apiserver-kubelet-client", "etcd-server", "etcd-healthcheck-client", "etcd-peer", "front-proxy-client"}
 	for _, cert := range certs {
 		certPath := []string{vmpath.GuestPersistentDir, "certs"}
@@ -366,11 +364,11 @@ func generateKubeadmCerts(cmd command.Runner, cc config.ClusterConfig) error {
 			certPath = append(certPath, "etcd")
 		}
 		certPath = append(certPath, strings.TrimPrefix(cert, "etcd-")+".crt")
-		if !isKubeadmCertValid(cmd, path.Join(certPath...), gt) {
-			needsRefresh = true
+		if !isKubeadmCertValid(cmd, path.Join(certPath...)) {
+			expiredCerts = true
 		}
 	}
-	if !needsRefresh {
+	if !expiredCerts {
 		return nil
 	}
 	out.WarningT("kubeadm certificates have expired. Generating new ones...")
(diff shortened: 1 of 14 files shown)


pkg/minikube/bootstrapper/certs.go
```diff
@@ -545,3 +575,32 @@ func isValid(certPath, keyPath string) bool {
...
+	}
+
+	cert, err := x509.ParseCertificate(certData.Bytes)
+	if err != nil {
+		klog.Infof("failed to parse cert file %s: %v\n", certPath, err)
+		return false
+	}
+
+	if cert.NotAfter.Before(time.Now()) {
+		out.WarningT("Certificate {{.certPath}} has expired. Generating a new one...", out.V{"certPath": filepath.Base(certPath)})
```
Kimi450, line 600: May I suggest moving "Generating a new one" to [the line](https://github.com/kubernetes/minikube/pull/16249/files#diff-c19d80ca90b0f40897c50414dcfd992cf6abeb549015075ea14a06b130c77d49R371) just before the renewal cmd is called. This this function ends up being reused somewhere itll be a bit misleading to say "generating" when its not actually done here (its just a check)

spowelljr: Good call, moved

pkg/minikube/bootstrapper/certs.go
```diff
@@ -545,3 +576,31 @@ func isValid(certPath, keyPath string) bool {
 
 	return true
 }
+
+func isKubeadmCertValid(cmd command.Runner, certPath string) bool {
```
medyagh, line 580: how about using linux to check if a cert is valid or not to avoid time zone difference between minikube and the host
```
Open a terminal window in Linux.

Type the following command, replacing <cert-file> with the path and name of the certificate file you want to check:

php
Copy code
openssl x509 -enddate -noout -in <cert-file>
Press Enter.

The output of the command will display the end date of the certificate in the format notAfter=YYYYMMDDHHMMSSZ.

For example:

makefile
Copy code
notAfter=20230410000000Z
Check the end date of the certificate. If the end date is in the past, then the certificate has expired.

For example, in the output above, the certificate will expire on April 10, 2023, at midnight UTC. If today's date is after April 10, 2023, then the certificate has expired.
```

medyagh (approved):

PR #16259, opened 2023-04-07
translate some fields

syxunion:
update zh-CN.json

I want to supplement the translation of some fields


spowelljr (approved):
Thanks for the translations!

Issue #16323, opened 2023-04-14
changes between 1.28 and 1.29+ lead to decrease in docker/daemon.json options (aka daemon.json is being always overwritten)


dmpe:
### What Happened?

Hi, 

We are using Centos 7 VM where we install minikube using something like this:

```
minikube start --cni="calico" \
                      --container-runtime="docker" \
                      --driver=none \
                      --delete-on-failure=true
```

This works perfectly fine, both using 1.28 as well as 1.29+ minikube releases. However, while in 1.28 `/etc/docker/daemon.json` was not replaced during the installation, in 1.29+ it is being replaced via this function 
https://github.com/kubernetes/minikube/blob/v1.29.0/pkg/minikube/cruntime/docker.go#L524. 
See for 1.28 here: https://github.com/kubernetes/minikube/blob/v1.28.0/pkg/minikube/cruntime/docker.go#L511

Well, of course, when we install minikube & docker and use `none` driver, we already setup docker in a way which allows us to use such configuration. :) :) 

In fact, we already set CGroupDriver to `systemd`, and not just that. We also customize many other settings in to make it work in a strictly regulated environment in our company, resulting into file be more like this:

```
{
  "data-root": "/srv/docker",
  "storage-driver": "overlay2",
  "bip": "10.xx.xx.xx/24",
  "exec-opts": ["native.cgroupdriver=systemd"],
  "log-driver": "json-file",
  "log-opts": {
    "max-size": "10m",
    "max-file": "3"
  }
}
```

Now, we are stuck on using minikube 1.28 because with 1.29 the minikube will replace daemon.json file which -- even though makes minikube start properly ! -- it is not usable for our purposes in the company. Reason being "docker data" (images, etc.) are stored in `/var/lib/docker` and `bip` with logging options cannot adjusted either, making it hard work in our environment.

What would be your suggestion here? Should we maybe add a new option which will disable such overwrite of a file (e.g. `minikube start --disable-daemon-overwrite=true`, with a default being `false` (current behavior). Or maybe fix it in a different way - e.g. check if driver is `systemd`, and if it is then do nothing, while it is not, attempt to change it. 

cc @prezha @afbjorklund

Thank you for any suggestions, or help. 

### Attach the log file

Minikube runs perfectly fine, it is just that daemon.json options for docker are no longer configurable.

### Operating System

Other

### Driver

None

afbjorklund:
[...]

PR #16363, opened 2023-04-22
fix(docker): add cgroup driver setting, but do not overwrite deamon.json file

dmpe:
attempt at fixing https://github.com/kubernetes/minikube/issues/16323

## Logic:

- `/etc/docker/daemon.json` is attempted to be read
- if it does not exist, then file is created, with `native.cgroupdriver` key
- if it exists, then it is parsed. 
  - if such JSON does not containt `exec-opts`, then it is added (with only this key/value and cgroup as passed to the function)
  - if such JSSON contains already `exec-opts`, then it's value is checked.
    - if the key is `native.cgroupdriver` then we skip any further processing, and file is not touched anymore
    - if the key is not `native.cgroupdriver`, then we add it with the value of the parameter that was passed. No other keys such as e.g. `bip` are adjusted/overwritten.  

## Tests

Besides CI, also run:

```
# example 1
sudo ./out/minikube stop && sudo ./out/minikube delete
sudo ./out/minikube start --driver none --cni calico --force-systemd=true
```

with `/etc/docker/daemon.json` being one of many combinations:

```
{"bip":"192.168.1.5/24","exec-opts":["native.cgroupdriver=systemd"],"log-driver":"json-file","log-opts":{"max-size":"100m"},"storage-driver":"overlay2"}
```
```
{"bip":"192.168.1.5/24","log-driver":"json-file","log-opts":{"max-size":"100m"},"storage-driver":"overlay2"}
```

Locally I tested this on Ubuntu 23.04 + docker 23.0.5

### DIFF
--- a/pkg/minikube/cruntime/docker.go
+++ b/pkg/minikube/cruntime/docker.go
@@ -49,6 +49,8 @@ const KubernetesContainerPrefix = "k8s_"
 const InternalDockerCRISocket = "/var/run/dockershim.sock"
 const ExternalDockerCRISocket = "/var/run/cri-dockerd.sock"
 
+const DockerDaemonJSON = "/etc/docker/daemon.json"
+
 // ErrISOFeature is the error returned when disk image is missing features
 type ErrISOFeature struct {
 	missing string
@@ -546,31 +548,79 @@ type dockerDaemonRuntimes struct {
 
 // configureDocker configures the docker daemon to use driver as cgroup manager
 // ref: https://docs.docker.com/engine/reference/commandline/dockerd/#options-for-the-runtime
+// If /etc/docker/daemon.json does not exist, file is created. Otherwise we just ensure that 
+// cgroup manager is set and do not adjust it beyond that. 
 func (r *Docker) configureDocker(driver string) error {
 	if driver == constants.UnknownCgroupDriver {
 		return fmt.Errorf("unable to configure docker to use unknown cgroup driver")
 	}
-
 	klog.Infof("configuring docker to use %q as cgroup driver...", driver)
+	
+	var daemonData map[string]any
+	natCgroupDriver := "native.cgroupdriver"
+	cgroupKeyValue := natCgroupDriver + "=" + driver
+
+	// defines default docker json 
 	daemonConfig := dockerDaemonConfig{
-		ExecOpts:  []string{"native.cgroupdriver=" + driver},
+		ExecOpts:  []string{cgroupKeyValue},
 		LogDriver: "json-file",
 		LogOpts: dockerDaemonLogOpts{
 			MaxSize: "100m",
 		},
 		StorageDriver: "overlay2",
 	}
+	
 	if r.Type == "nvidia-docker" {
 		daemonConfig.DefaultRuntime = "nvidia"
 		runtimes := &dockerDaemonRuntimes{}
 		runtimes.Nvidia.Path = "/usr/bin/nvidia-container-runtime"
 		daemonConfig.Runtimes = runtimes
 	}
-	daemonConfigBytes, err := json.Marshal(daemonConfig)
+
+	file, err := os.ReadFile(DockerDaemonJSON)
+	if err != nil {
+		klog.Warningf("Failed to load %s. Creating new json file.", DockerDaemonJSON)
+		dataBytes, err := json.Marshal(daemonConfig)
+		if err != nil {
+			return err
+		}
+		return r.writeToDockerdJSON(dataBytes)
+	}
+
+	// we don't know the exact structure that user has already been using
+	err = json.Unmarshal(file, &daemonData)
+	if err != nil {
+		klog.Warningf("Failed to parse %s", file)
+	}
+	execOpts := daemonData["exec-opts"]
+	if execOpts == nil {
+		klog.Infof("%s does not contain any 'exec-opts' key. Configuring docker to use %s as cgroup driver...", DockerDaemonJSON, driver)
+		daemonData["exec-opts"] = []string{cgroupKeyValue}
+		dataBytes, _ := json.MarshalIndent(daemonData, "", "  ")
+		return r.writeToDockerdJSON(dataBytes)
+	}
+	klog.Infof("%s contains exec-opts option. Checking for %s key presence.", DockerDaemonJSON, driver)
+	jsonExecOpts, ok := execOpts.([]string)
+	if ok {
+		for _, value := range jsonExecOpts {
+			dKey := strings.Split(value, "=")[0]
+			if dKey == natCgroupDriver {
+				return nil
+			}
+		}
+	}
+	klog.Infof("Insert %s driver with option %s into %s file.", natCgroupDriver, driver, DockerDaemonJSON)
+	daemonData["exec-opts"] = append(jsonExecOpts, cgroupKeyValue)
+
+	dataBytes, err := json.MarshalIndent(daemonData, "", "  ")
 	if err != nil {
 		return err
 	}
-	ma := assets.NewMemoryAsset(daemonConfigBytes, "/etc/docker", "daemon.json", "0644")
+	return r.writeToDockerdJSON(dataBytes)
+}
+
+func (r *Docker) writeToDockerdJSON(data []byte) error {
+	ma := assets.NewMemoryAsset(data, "/etc/docker", "daemon.json", "0644")
 	return r.Runner.Copy(ma)
 }
 


dmpe:
Hi reviewers, can you take a look please on this. If the logic/approach to it fine, could you pls trigger additional tests? Thanks

dmpe:
Do you mind @spowelljr taking a look on this? I havent rebased for some time, but would appreciate a feedback - maybe even running some additional CI tests?

medyagh:
@dmpe do you mind rebasing your PR, thanks for patience on the review

dmpe:
> @dmpe do you mind rebasing your PR, thanks for patience on the review

Hi @medyagh, thank you very much for the feedback. I believe i have rebased the PR, shortly before you got back to me.

pkg/minikube/cruntime/docker.go
```diff
@@ -531,19 +537,56 @@ func (r *Docker) setCGroup(driver string) error {
...
+			dKey := strings.Split(value.(string), "=")[0]
+			if dKey == natCgroupDriver {
+				containsCGroupName = true
+			}
+		}
+		if !containsCGroupName {
+			klog.Infof("Insert %s driver into %s file.", natCgroupDriver, DockerDaemonJSON)
+			daemonData["exec-opts"] = append(jsonExecOpts, cgroupKeyValue)
+			writeToDockerdJSON = true
+		}
```
spowelljr, line 582:
```suggestion
				return nil
			}
		}
		klog.Infof("Insert %s driver into %s file.", natCgroupDriver, DockerDaemonJSON)
		daemonData["exec-opts"] = append(jsonExecOpts, cgroupKeyValue)
		writeToDockerdJSON = true
```
If Cgroup is already expect can just return and save some nesting.

pkg/minikube/cruntime/docker.go
```diff
@@ -531,19 +537,56 @@ func (r *Docker) setCGroup(driver string) error {
...
+	}
+	if writeToDockerdJSON {
+		dataBytes, _ := json.MarshalIndent(daemonData, "", "  ")
+		ma := assets.NewMemoryAsset(dataBytes, "/etc/docker", "daemon.json", "0644")
+		return r.Runner.Copy(ma)
+	}
+	return nil
```
spowelljr, line 589:
```suggestion
		return writeToDockerdConfig(daemonStr)
	}

	// we don't know exact structure that user has already been using
	err = json.Unmarshal(file, &daemonData)
	if err != nil {
		klog.Warningf("Failed to parse %s", file)
	}
	execOpts := daemonData["exec-opts"]
	if execOpts == nil {
		klog.Infof("%s does not contain any 'exec-opts' key. Configuring docker to use %s as cgroup driver...", DockerDaemonJSON, driver)
		daemonData["exec-opts"] = []string{cgroupKeyValue}
		return writeToDockerdConfig(daemonData)
	}
	klog.Infof("%s contains exec-opts option. Checking for %s key presence.", DockerDaemonJSON, natCgroupDriver)
	jsonExecOpts := execOpts.([]interface{})
	for _, value := range jsonExecOpts {
		dKey := strings.Split(value.(string), "=")[0]
		if dKey == natCgroupDriver {
			return nil
		}
	}
	klog.Infof("Insert %s driver into %s file.", natCgroupDriver, DockerDaemonJSON)
	daemonData["exec-opts"] = append(jsonExecOpts, cgroupKeyValue)
	return writeToDockerdConfig(daemonData)
}

func writeToDockerdConfig(data any) error {
	dataBytes, _ := json.MarshalIndent(data, "", "  ")
	ma := assets.NewMemoryAsset(dataBytes, "/etc/docker", "daemon.json", "0644")
	return r.Runner.Copy(ma)
```
Having a `writeToDockerdConfig` func would remove most of the nesting and make it a lot easier to read

pkg/minikube/cruntime/docker.go
```diff
@@ -531,19 +537,56 @@ func (r *Docker) setCGroup(driver string) error {
 	if driver == constants.UnknownCgroupDriver {
 		return fmt.Errorf("unable to configure docker to use unknown cgroup driver")
 	}
+	var (
+		containsCGroupName bool
+		writeToDockerdJSON bool
+		daemonData         map[string]any
+	)
```
spowelljr, line 544:
```suggestion
	var daemonData map[string]any
```

spowelljr (changes requested):

pkg/minikube/cruntime/docker.go
dmpe: Thanks @spowelljr ! I have incorporated the fixes. The `writeToDockerdConfig` had to be slightly adjusted as `(r *Docker)` struct type was missing.

pkg/minikube/cruntime/docker.go
```diff
@@ -531,18 +537,47 @@ func (r *Docker) setCGroup(driver string) error {
...
+		klog.Warningf("Failed to parse %s", file)
+	}
+	execOpts := daemonData["exec-opts"]
+	if execOpts == nil {
+		klog.Infof("%s does not contain any 'exec-opts' key. Configuring docker to use %s as cgroup driver...", DockerDaemonJSON, driver)
+		daemonData["exec-opts"] = []string{cgroupKeyValue}
+		return r.writeToDockerdJSON(daemonData)
+	} 
+	klog.Infof("%s contains exec-opts option. Checking for %s key presence.", DockerDaemonJSON, natCgroupDriver)
+	jsonExecOpts := execOpts.([]interface{})
```
spowelljr, line 565:
```suggestion
	jsonExecOpts, ok := execOpts.([]interface{})
	if !ok {
```
Just incase the user malformed their config file (string instead of []string) we should use `ok` to prevent a panic.

spowelljr (changes requested):

pkg/minikube/cruntime/docker.go
```diff
@@ -531,18 +537,47 @@ func (r *Docker) setCGroup(driver string) error {
...
+	for _, value := range jsonExecOpts {
+		dKey := strings.Split(value.(string), "=")[0]
```
spowelljr, line 567: Same with `value.(string)` if an `int` snuck it would cause a panic.

spowelljr (changes requested):

pkg/minikube/cruntime/docker.go
spowelljr, line 565:
```suggestion
	jsonExecOpts, ok := execOpts.([]string)
```
I believe that would prevent needing the `value.(string)` below.

dmpe:
Thanks again.

pkg/minikube/cruntime/docker.go
```diff
@@ -531,18 +537,49 @@ func (r *Docker) setCGroup(driver string) error {
...
+	natCgroupDriver := "native.cgroupdriver"
+	cgroupKeyValue := natCgroupDriver + "=" + driver
+
+	file, err := os.ReadFile(DockerDaemonJSON)
+	if err != nil {
+		klog.Warningf("Failed to load %s. Creating new json file.", DockerDaemonJSON)
+		daemonStr := &DockerDaemon{
+			Exopt: []string{cgroupKeyValue},
+		}
+		return r.writeToDockerdJSON(daemonStr)
```
spowelljr, line 550: If there's no file could you continue to use the JSON string we were previously writing. For non-none driver usage we're basically using this command to write the entire daemon.json but now it's only getting populated with the exec-opts, so continue writing the whole file to prevent unexpected issues.

**Before:**
```
{
"exec-opts": ["native.cgroupdriver=cgroupfs"],
"log-driver": "json-file",
"log-opts": {
	"max-size": "100m"
},
"storage-driver": "overlay2"
}
```

**Now:**
```
{
"exec-opts": ["native.cgroupdriver=cgroupfs"]
}
```

spowelljr (changes requested):

dmpe:
The `pull-minikube-build` now fails with `! Image was not built for the current minikube version. To resolve this you can delete and recreate your minikube cluster using the latest images. Expected minikube version: v1.31.0 -> Actual minikube version: v1.30.1`. 
In other news, I think i have adjusted as proposed.

spowelljr:
> The `pull-minikube-build` now fails with `! Image was not built for the current minikube version. To resolve this you can delete and recreate your minikube cluster using the latest images. Expected minikube version: v1.31.0 -> Actual minikube version: v1.30.1`. In other news, I think i have adjusted as proposed.

That's fine, that's just because you rebased and we're about to go for a release to the kicbase is updated to expect the minikube version to be v1.31.0, you can ignore it

pkg/minikube/cruntime/docker.go
```diff
@@ -531,18 +533,55 @@ func (r *Docker) setCGroup(driver string) error {
...
-"log-driver": "json-file",
-"log-opts": {
-	"max-size": "100m"
-},
-"storage-driver": "overlay2"
+	file, err := os.ReadFile(DockerDaemonJSON)
+	if err != nil {
+		klog.Warningf("Failed to load %s. Creating new json file.", DockerDaemonJSON)
+		daemonConfig := fmt.Sprintf(`{
+			"exec-opts": ["native.cgroupdriver=%s"],
```
dmpe, line 544:
```suggestion
			"exec-opts": ["%s"],
```

pkg/minikube/cruntime/docker.go
```diff
@@ -531,18 +533,55 @@ func (r *Docker) setCGroup(driver string) error {
...
+			"log-driver": "json-file",
+			"log-opts": {
+				"max-size": "100m"
+			},
+			"storage-driver": "overlay2"
+		}`, driver)
```
dmpe, line 550:
```suggestion
		}`, cgroupKeyValue)
```

dmpe:
I am not sure why a lot of tests are failing, and I have a _feeling_ it is not due to being flake, rather some code issue. E.g. [this test](https://storage.googleapis.com/minikube-builds/logs/16363/30224/KVM_Linux.html#fail_TestMultiNode/serial/StartAfterStop) failed when using cgroupfs (minikube start was without force systemd option).

dmpe:
@spowelljr Hi, could you take a look on this again, sometime. Thanks

dmpe:
Hi, Can you take a look and if it is possible merge this already ? @spowelljr @medyagh. Thanks.

pkg/minikube/cruntime/docker.go
dmpe: Has been addressed.

dmpe:
> Hi, Can you take a look and if it is possible merge this already ? @spowelljr @medyagh. Thanks.

How I help to get this over the finishing line ? Am I missing something ?

The issue really prevents us from upgrading minikube to newer versions, because we require that some docker deamon options are set manually - not by minikube (e.g. `bip`). There are already other people impacted by it as well https://github.com/NVIDIA/k8s-device-plugin/issues/423 - I am however without the ability to switch to docker driver. 

Maybe @prezha you could take a look as well ? Many thanks.

prezha:
hey @dmpe, looks like the underlying code changed in the meantime:

```
This branch has conflicts that must be resolved
Use the [web editor](https://github.com/kubernetes/minikube/pull/16363/conflicts) or the to resolve conflicts.
Conflicting files
pkg/minikube/cruntime/docker.go
```
also, users reported that the nvidia gpu addon works for them (eg, [this](https://github.com/NVIDIA/k8s-device-plugin/issues/423#issuecomment-1701735916) message)

can you please re-check?

dmpe:
> hey @dmpe, looks like the underlying code changed in the meantime:
> 
> ```
> This branch has conflicts that must be resolved
> Use the [web editor](https://github.com/kubernetes/minikube/pull/16363/conflicts) or the to resolve conflicts.
> Conflicting files
> pkg/minikube/cruntime/docker.go
> ```
> 
> also, users reported that the nvidia gpu addon works for them (eg, [this](https://github.com/NVIDIA/k8s-device-plugin/issues/423#issuecomment-1701735916) message)
> 
> can you please re-check?

I have just rebased and hope that it is going to be fine, with all of you. Please, dont forget to approve the workflows else cannot be merged -:). cc @spowelljr 

Speaking of the nvidia issue I linked: The user has indeed "fixed" that by using a different driver - which however I am not able  to do (existing scripting, testing need for new approach, etc.). As I wrote previously, we require several additional options being set in daemon json file ... Hence overwriting this file by minikube installation is something that prohibits us from upgrading to the newer minikube versions for now. 

Many thanks.

pkg/minikube/cruntime/docker.go
```diff
@@ -546,31 +548,79 @@ type dockerDaemonRuntimes struct {
...
+		dataBytes, err := json.Marshal(daemonConfig)
+		if err != nil {
+			return err
+		}
+		return r.writeToDockerdJSON(dataBytes)
+	}
+
+	// we don't know the exact structure that user has already been using
+	err = json.Unmarshal(file, &daemonData)
+	if err != nil {
```
prezha, line 592: nit: with a single return error value, we can compress it to the `if err := ...; err != nil {...` form

pkg/minikube/cruntime/docker.go
```diff
@@ -546,31 +548,79 @@ type dockerDaemonRuntimes struct {
...
+		klog.Warningf("Failed to parse %s", file)
```
prezha, line 593: should it fail here, there's probably no point in continuing, as we might get some weird, unexpected results?

pkg/minikube/cruntime/docker.go
```diff
@@ -546,31 +548,79 @@ type dockerDaemonRuntimes struct {
...
+	}
+	execOpts := daemonData["exec-opts"]
+	if execOpts == nil {
+		klog.Infof("%s does not contain any 'exec-opts' key. Configuring docker to use %s as cgroup driver...", DockerDaemonJSON, driver)
+		daemonData["exec-opts"] = []string{cgroupKeyValue}
+		dataBytes, _ := json.MarshalIndent(daemonData, "", "  ")
```
prezha, line 599: we should not ignore eventual error returned

pkg/minikube/cruntime/docker.go
```diff
@@ -546,31 +548,79 @@ type dockerDaemonRuntimes struct {
...
+		return r.writeToDockerdJSON(dataBytes)
```
prezha, line 600: this would only add missing `exec-opts` in the existing daemon.json and ignore/break the `nvidia-docker` config, as that would be discarded

pkg/minikube/cruntime/docker.go
```diff
@@ -546,31 +548,79 @@ type dockerDaemonRuntimes struct {
...
+	}
+	klog.Infof("%s contains exec-opts option. Checking for %s key presence.", DockerDaemonJSON, driver)
+	jsonExecOpts, ok := execOpts.([]string)
+	if ok {
+		for _, value := range jsonExecOpts {
+			dKey := strings.Split(value, "=")[0]
+			if dKey == natCgroupDriver {
+				return nil
```
prezha, line 608: similarly, this would ignore/break the nvidia-docker config, as that would be discarded

pkg/minikube/cruntime/docker.go
```diff
@@ -546,31 +548,79 @@ type dockerDaemonRuntimes struct {
...
+	}
+	klog.Infof("Insert %s driver with option %s into %s file.", natCgroupDriver, driver, DockerDaemonJSON)
+	daemonData["exec-opts"] = append(jsonExecOpts, cgroupKeyValue)
+
+	dataBytes, err := json.MarshalIndent(daemonData, "", "  ")
 	if err != nil {
 		return err
 	}
-	ma := assets.NewMemoryAsset(daemonConfigBytes, "/etc/docker", "daemon.json", "0644")
+	return r.writeToDockerdJSON(dataBytes)
```
prezha, line 619: as above, this would ignore/break the nvidia-docker config, as that would be discarded

prezha (changes requested):
@dmpe i left a few comments, please have a look
most critically is to more specifically account for and retain the "nvidia-docker" config, if applicable
you can log the config that'd be applied, and then please share _before_ (original) and _after_ (final) config outputs
ideally, we'd have tests to cover these

one thing i'm thinking of:
we're fine with overwriting our own `kicbase`/`iso` isolated docker daemon.json
the problem is only with the `none` driver - ie, when we're on a "bare-metal" or directly on a user's vm
should we, in the `none` driver case, **_just not touch the docker daemon.json_** and let users configure it appropriately (and also customise it to their needs)?
this would then also leave it to the user to configure the docker daemon for "nvidia-docker" (which might not be bad, as it's not hard-coded and is under users' control, should the config need to be changed in the future)
apart from this, historically, we needed to rewrite docker daemon.json just to set the correct cgroups driver (in our kicbase/iso), which should be automatically and correctly set by docker for `none` machines anyway

@medyagh @spowelljr @afbjorklund wdyt?

dmpe:
Thank you again @prezha for both feedback and quick responses!

>  and let users configure it appropriately (and also customise it to their needs)

I _absolute_ AGREE - not just agree but also think it should have been done from day 1. And I mean, it has worked well before...

As a `none` user, you are already telling me at least twice that if I run minikube I need `cri-docker` these days and install `crictl` as well. Myself! Minikube does not do it, even though I certainly could ...

Your proposal would also fix the nvidia issue as well my additional configuration that I have to have. See the parent issue how my docker json file looks like.
Meaning that either I had to include additional docker options into the `struct` (which I considered a lot; but how many should minikube additionally expose or even include - all/some ?!) or use a different approach of merging default config with any additional one like here.

I would be **very happy** if this existing behaviour got changed, so that daemon json is fully in my own ownership. Minikube project could then tell me that I have to configure it properly myself and project can't make sure that it will work. And I am fine with that because you are already assuming that I am "expert user". 

I will wait for others to reply before going through you comments.

dmpe:
@prezha how would I go about implementing what we actually agree on, namely let users configure their docker daemon config by themselves. 

There are at least 2 ways AFAIK:
- remove `force-systemd` parameter from `minikube start`
  - more code adjustments required (not bad!) but I am not sure if `non-systemd` driver is still being used somewhere - as of now, with 1.27+ being minikube's default k8s version. 
    -  e.g. `cgroupDriver` function will be removed, etc.
- only adjust `configureDocker` function to do "nothing" when `--driver` is set to `none` ?
  - less invasive to minikube code base because we only add if/else check.  

[Docker Documentation states following](https://docs.docker.com/engine/reference/commandline/dockerd/):
> If you omit the native.cgroupdriver option, cgroupfs is used on cgroup v1 hosts, systemd is used on cgroup v2 hosts with systemd available.

Considering that we are doing a new major release, and `systemd` become standard I would be in favor of entirely removing `cgroup` option(s) from minikube and rely on OS default/what docker/user sets by themselves. WDYT?

PR #16367, opened 2023-04-23
chore: replace go-ps with gopsutil/process

Juneezee:
This commit replaces the usage of `github.com/mitchellh/go-ps` with `github.com/shirou/gopsutil/v3/process`. `gopsutil` offers the same set of functionalities, eliminating the need for two dependencies.

### DIFF
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -27,8 +27,8 @@ import (
 	"time"
 
 	"github.com/docker/machine/libmachine/mcnerror"
-	"github.com/mitchellh/go-ps"
 	"github.com/pkg/errors"
+	"github.com/shirou/gopsutil/v3/process"
 
 	"github.com/docker/machine/libmachine"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


medyagh (commented):
I’m curious how much does this affect minikube binary size ?

Do u mind sharing the minikube binary size before after this pr

Juneezee:
Hi @medyagh 

#### Before (~82.17 MiB)
```
$ ls -al
total 84152
drwxr-xr-x  2 jun jun     4096 Apr 24 14:39 .
drwxr-xr-x 17 jun jun     4096 Apr 24 14:38 ..
-rwxr-xr-x  1 jun jun 86158639 Apr 24 14:39 minikube-linux-amd64
```

#### After (~82.21 MiB)
```
$ ls -al
total 84152
drwxr-xr-x  2 jun jun     4096 Apr 24 14:37 .
drwxr-xr-x 17 jun jun     4096 Apr 24 14:38 ..
-rwxr-xr-x  1 jun jun 86203064 Apr 24 14:37 minikube-linux-amd64
```

spowelljr:
> I’m curious how much does this affect minikube binary size ?
> 
> Do u mind sharing the minikube binary size before after this pr

`github.com/mitchellh/go-ps` is a dependency of `github.com/moby/hyperkit` so it's going to be included in the binary either way.

spowelljr (commented):
Since `github.com/mitchellh/go-ps` is required anyway (above comment), changing `github.com/mitchellh/go-ps` to `github.com/shirou/gopsutil/v3/process` seems to make the code more complex as now a cast is required `process.NewProcess(int32(pid))` and an extra error check
```
processName, err := process.Name()
if err != nil {
	return errors.Wrap(err, "error getting process name")
}
```

So I don't really see a benefit to this change

PR #16373, opened 2023-04-24
Libmachine onboarding

x7upLime:


afbjorklund (changes requested):
There seems to be some paths still in there, such as:

	"github.com/docker/machine/libmachine/log"
	"github.com/docker/machine/libmachine/mcnutils"

afbjorklund:
Both projects (both machine and minikube) are bad at stating copyright, so we might need to address that sometime.

But basically the code is:

`Copyright 2014 Docker, Inc.`

`Copyright 2016 The Kubernetes Authors All rights reserved.`

x7upLime:
sorry @afbjorklund, just saw your comments.
I started over and moved this into #16459
The idea is still the same.. now the new version of the pr compiles/has better description...

Issue #16378, opened 2023-04-25
minikube fails to start w/ none driver and docker CRI w/o containernetworking-plugins


spowelljr:
If [containernetworking-plugins](https://github.com/containernetworking/plugins) are not installed (`/opt/cni/bin` is empty) starting minikube with the `none` driver and `docker` CRI fails.

If the `--wait=all` flag is not passed it will appear to succeed, but if you check `kubectl pods -A` you'll see the `coredns` and `storage-provisioner` will never come up.
```
NAMESPACE     NAME                                        READY   STATUS    RESTARTS   AGE
kube-system   coredns-787d4945fb-74xcd                    0/1     Pending   0          8s
kube-system   etcd-ubuntu-20-agent-2                      1/1     Running   2          22s
kube-system   kube-apiserver-ubuntu-20-agent-2            1/1     Running   2          23s
kube-system   kube-controller-manager-ubuntu-20-agent-2   1/1     Running   2          23s
kube-system   kube-proxy-7x2k9                            1/1     Running   0          9s
kube-system   kube-scheduler-ubuntu-20-agent-2            1/1     Running   2          23s
kube-system   storage-provisioner                         0/1     Pending   0          7s
```

If you pass `--wait=all` the start will fail after waiting 6 minutes for the above pods to start.

This stems from cri-docker failing due to the missing plugins:
```
cri-dockerd[11762]: time="2023-04-25T13:11:31Z" level=error msg="Error validating CNI config list (\n{\n  \"cniVersion\": \"0.3.1\",\n  \"name\": \"bridge\",\n  \"plugins\": [\n    {\n      \"type\": \"bridge\",\n      \"bridge\": \"bridge\",\n      \"addIf\": \"true\",\n      \"isDefaultGateway\": true,\n      \"forceAddress\": false,\n      \"ipMasq\": true,\n      \"hairpinMode\": true,\n      \"ipam\": {\n          \"type\": \"host-local\",\n          \"subnet\": \"10.244.0.0/16\"\n      }\n    },\n    {\n      \"type\": \"portmap\",\n      \"capabilities\": {\n          \"portMappings\": true\n      }\n    }\n  ]\n}\n): [failed to find plugin \"bridge\" in path [/opt/cni/bin] failed to find plugin \"portmap\" in path [/opt/cni/bin]]"
```

Resulting in kubelet failing:
```
[...]

PR #16380, opened 2023-04-25
CI: install cni-plugins for none driver tests

spowelljr:
Related https://github.com/kubernetes/minikube/issues/16378


medyagh (approved):
Is this something minikube should advice on users too? To install this?

spowelljr:
> Is this something minikube should advice on users too? To install this?

Yes, it's in the task list for for the issue https://github.com/kubernetes/minikube/issues/16378

PR #16395, opened 2023-04-28
Fix csi hostpath custom registry error

tzzcfrank:
Fix the problem that the image address in the csi hostpath template is fixed and cannot be customized.


spowelljr (approved):
Thanks for the fix!

Issue #15772, opened 2023-02-02
create automation to bump the ubuntu base image


medyagh:
### What Happened?

we need to create add an automation to do PRs like this https://github.com/kubernetes/minikube/pull/15768/files

we already have a script that does a lot of updating PRs that would need to be updated to include this type of update
here is the path to the script

https://github.com/kubernetes/minikube/tree/master/hack/update

salasberryfin:
Hi @medyagh,

I'd like to work on this but I may need some extra help.

What I understand is that this automation should get the latest version from the registry and modify the Dockerfile.

This is my first contribution to minikube and I'd appreciate some help navigating the code base. Should this be integrated into one of the existing automations? Is it specific for the kicbase Dockerfile or does it need to be extensible to other similar cases?

Thanks!

/assign

spowelljr:
@salasberryfin It should get the latest `focal` tag (https://hub.docker.com/_/ubuntu/tags?page=1&name=focal) and update the kicbase Dockerfile.

Here's an example of what the change should look like:

https://github.com/kubernetes/minikube/pull/15768/files#diff-7b41e58d929bdb16083790c3500e6f9aa19efb0976e6b65609b60396cd9ceeebL40

It can be specific to the kicbase Dockerfile.

If you look at `hack/update` you'll see a bunch of scripts that do updating related logic, and if you look in `.github/workflows` you can see some crons to run the update scripts.

The hardest part is going to find somewhere we can curl to get the latest Ubuntu focal version, after that everything else has already been done and can copy existing logic.

salasberryfin:
Thanks for the info @spowelljr. I will take a look at this!

salasberryfin:
Hi @spowelljr,

I was taking a look at the Docker Hub API, which is probably the easiest solution to get the latest `focal` tag and I was wondering, if this is supposed to automatically retrieve and update to the latest version, couldn't it point directly to `ubuntu:focal` instead of the specific tag name? Would this cause any unexpected issue?

spowelljr:
[...]

PR #16409, opened 2023-04-29
CI: Automate ubuntu version update #15772

salasberryfin:
This PR is related to the following issue #15772 

It adds a new automation that retrieves the latest `ubuntu:focal` version from Docker Hub and updates the Dockerfile in `deploy/kicbase/Dockerfile`. It is configured to run weekly as a GitHub Action and create a new PR, following the procedure found in existing workflows. The image version can also be updated using `make update-ubuntu-version`.

The details of the image are retrieved from the Docker Hub API.


spowelljr (changes requested):
Looks good, just a couple small things

spowelljr (changes requested):
Also, could you ask this so it will build the kicbase image

https://github.com/kubernetes/minikube/blob/master/.github/workflows/update-docker-version.yml#L51-L60

.github/workflows/update-ubuntu-version.yml
```diff
@@ -0,0 +1,58 @@
...
+        id: bumpUbuntu
+        run: |
+          make update-ubuntu-version
+          c=$(git status --porcelain)
+          c="${c//$'\n'/'%0A'}"
+          c="${c//$'\r'/'%0D'}"
+          echo "changes<<EOF" >> $GITHUB_OUTPUT
+          echo "$c" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+      - name: Create PR
```
spowelljr, line 32: Will need to add:
```
id: createPR
```

So `${{ steps.createPR.outputs.pull-request-number }}` can get the PR number

spowelljr (changes requested):

.github/workflows/update-ubuntu-version.yml
salasberryfin: Sorry about that. Took the code snippet from the `update-docker-version` and didn't realize this was missing. I think it is solved now.

salasberryfin:
I applied the proposed changes and I think all of them are resolved now. :)

spowelljr:
Hi @salasberryfin, this looks great! There's one other change that would be good to see, but it can be in a follow up PR if you'd prefer.

It's getting the from and to versions in the PR title, for example our buildkit PR:
```
Kicbase/ISO: Update buildkit from v0.11.4 to v0.11.6
```

If you look at:
https://github.com/kubernetes/minikube/blob/92f364887c20f87d09509d4762c69ccb1cef9d2c/.github/workflows/update-cri-o-version.yml#L25-L27
That's how we get the versions, and then do:
https://github.com/kubernetes/minikube/blob/92f364887c20f87d09509d4762c69ccb1cef9d2c/.github/workflows/update-cri-o-version.yml#L46

The logic for getting the version is in https://github.com/kubernetes/minikube/blob/master/hack/update/get_version/get_version.go

You'd add `ubuntu` to the map and then the regexp to get the version from the `UBUNTU_FOCAL_IMAGE` variable you created.

So the outcome would be your PR title would look like:
```
Kicbase: Update base image from focal-20230308 to focal-20230412
```

If you'd prefer to do this in a follow up PR just let me know and I can merge this one.

salasberryfin:
Hi @spowelljr. If you're okay with it, I'd prefer doing it in a separate PR since I won't be able to look at it before the weekend. I'll be picking it up then and creating a new PR.

spowelljr (approved):

PR #16459, opened 2023-05-06
Libmachine onboarding/plumbing

x7upLime:


medyagh (changes requested):
I would rather we do not add the libmachine code into minimube code base, 

@x7upLime can u add the code to a personal Repo called LibMachine and then only import that form minikube code without copying the code into minikube

that would be easier to review

medyagh:
hi @x7upLime in last office hours we decided to create a new repo to move all machine related codes to there (including driver) codes

would you please make a PR to this empty repo https://github.com/minikube-machine/machine

afbjorklund:
I imported the current libmachine (from machine-drivers) into github.com/minikube-machine/machine (as github.com/docker/machine)) and made some other PRs to clean up the minikube registry and drivers.

Once we have the external repositories binaries under control, we can rename the API (including the gRPC)
And once _that_ is completed, then we can start uplifting the API - since it would no longer use docker-machine

medyagh:
@x7upLime If u dont mind, we close this PR in favor of https://github.com/kubernetes/minikube/pull/16672
we could start working on the plan of moving to this new org for machine driver code

PR #16543, opened 2023-05-19
audit log: set default max size

spowelljr:
I was looking at some of our CI machines and the audit logging was failing:
```
E0519 17:42:33.138427  334137 logFile.go:53] failed to close the audit log: invalid argument
```

Then I looked at the size of the file:
```
-rw-r--r-- 1 perf perf 651M May 19 17:44 /home/perf/.minikube/logs/audit.json
```

It has over 1.6 million lines of audits:
```
1621891 /home/perf/.minikube/logs/audit.json
```

This PR sets the default number of audit lines to 1000 if not explicitly set by the user.

### DIFF
--- a/pkg/minikube/audit/audit.go
+++ b/pkg/minikube/audit/audit.go
@@ -121,7 +121,11 @@ func LogCommandEnd(id string) error {
 }
 
 func getStartIndex(entryCount int) int {
-	maxEntries := viper.GetInt(config.MaxAuditEntries)
+	// default to 1000 entries
+	maxEntries := 1000
+	if viper.IsSet(config.MaxAuditEntries) {
+		maxEntries = viper.GetInt(config.MaxAuditEntries)
+	}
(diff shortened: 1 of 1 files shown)


medyagh (approved):

pkg/minikube/audit/audit.go
```diff
@@ -121,7 +121,11 @@ func LogCommandEnd(id string) error {
 }
 
 func getStartIndex(entryCount int) int {
-	maxEntries := viper.GetInt(config.MaxAuditEntries)
+	// default to 1000 entries
```
medyagh, line 124: hm... I wonder what is the best thing to do to if the audilog was already more than that in the files...should delete it ?

spowelljr: This truncated the existing file, so if it's currently 2000 lines, it will truncate it to the last 1000

Issue #16558, opened 2023-05-22
`--kubernetes-version` without patch version should use latest patch


spowelljr:
**Current:**
```
$ minikube start --kubernetes-version=v1.26
😄  minikube v1.30.1 on Darwin 13.4 (arm64)

❌  Exiting due to MK_USAGE: Unable to parse "v1.26": No Major.Minor.Patch elements found
```

If the user omits passing the patch version we should use https://github.com/kubernetes/minikube/blob/master/pkg/minikube/constants/constants_kubeadm_images.go to find the latest patch version and use that.

**Expected:**
```
$ minikube start --kubernetes-version=v1.26
<emoji> Using Kubernetes v1.26.5 since patch version was unspecified
✨  Automatically selected the docker driver. Other choices: qemu2, ssh
📌  Using Docker Desktop driver with root privileges
👍  Starting control plane node minikube in cluster minikube
🚜  Pulling base image ...
🔥  Creating docker container (CPUs=2, Memory=4000MB) ...
🐳  Preparing Kubernetes v1.26.5 on Docker 24.0.0 ...
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔗  Configuring bridge CNI (Container Networking Interface) ...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🔎  Verifying Kubernetes components...
🌟  Enabled addons: storage-provisioner, default-storageclass
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default
```

PR #16563, opened 2023-05-23
Add list of Kubernetes versions

spowelljr:
This PR adds a hardcoded list of Kubernetes versions to be used in three future PRs.

PR 1. Update kubeadm constants automation to also update this list
PR 2. Replace `minikube config defaults kubernetes-version` to pull from this list instead of calling GitHub every time
PR 3. To be used in https://github.com/kubernetes/minikube/issues/16558 to get the latest minor version


pkg/minikube/constants/constants_kubernetes_versions.go
```diff
@@ -0,0 +1,270 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package constants
+
+var KubernetesVersions = []string{
```
medyagh, line 19: add a comment why we need this and how it is used.
maybe a better name ? how about ValidKubernetesVersions

medyagh (changes requested):

pkg/minikube/constants/constants_kubernetes_versions.go
medyagh: if it is only by automation, add a comment  that says do not touch only automation

medyagh (changes requested):

PR #16569, opened 2023-05-23
--kubernetes-version: assume latest patch version if not specified by user

JudahNour:
Closes https://github.com/kubernetes/minikube/issues/16558

## Before: 
```
$ mk start --kubernetes-version=1.26
😄  minikube v1.30.1 on Debian rodete (amd64)

❌  Exiting due to MK_USAGE: Unable to parse "1.26": No Major.Minor.Patch elements found
```

## After:
```
$ mk start --kubernetes-version=1.26
😄  minikube v1.30.1 on Debian rodete (amd64)
👉  Using Kubernetes 1.26.5 since patch version was unspecified
🚜  Pulling base image ...
🔥  Creating docker container (CPUs=2, Memory=8000MB) ...
🐳  Preparing Kubernetes v1.26.5 on Docker 24.0.0 ...
    ▪ Configuring RBAC rules ...
```

## Before: 
```
$ mk start --kubernetes-version=1.50
😄  minikube v1.30.1 on Debian rodete (amd64)

❌  Exiting due to MK_USAGE: Unable to parse "1.50": No Major.Minor.Patch elements found
```
## After:
```
$ mk start --kubernetes-version=1.50
😄  minikube v1.30.1 on Debian rodete (amd64)

❌  Exiting due to MK_PATCH_NOT_FOUND: Unable to detect the latest patch release for specified major.minor version v1.50
💡  Suggestion: Specify --kubernetes-version in v<major>.<minor.<build> form. example: 'v1.1.14'
```
## After:
```
$ mk start --kubernetes-version=hahahahha
😄  minikube v1.30.1 on Debian rodete (amd64)

❌  Exiting due to MK_USAGE: Unable to parse "hahahahha": No Major.Minor.Patch elements found

```


cmd/minikube/cmd/start.go
```diff
@@ -1757,8 +1757,17 @@ $ minikube config unset kubernetes-version`)
 	} else if strings.EqualFold(strings.ToLower(paramVersion), "latest") || strings.EqualFold(strings.ToLower(paramVersion), "newest") {
 		paramVersion = constants.NewestKubernetesVersion
 	}
-
-	nvs, err := semver.Make(strings.TrimPrefix(paramVersion, version.VersionPrefix))
+	kubernetesSemver := strings.TrimPrefix(paramVersion, version.VersionPrefix)
```
medyagh, line 1760: lets make sure we notify the user that we chose latest patch for them since they didn't provide a patch version

medyagh (changes requested):
lets make

cmd/minikube/cmd/start.go
medyagh: also a better PR title would be, "--kubernetes-version: assume latest patch ver if not specified by user"

JudahNour: Done. I abstracted out the majorMinor check and print it out in validateKubernetesVersion. The function uses the param version so this could be slightly redundant as we do that processing of the param in getKubernetesVersion already and thus redo some of it in validateKubernetesVersion so we can run the check there as well.

cmd/minikube/cmd/start.go
```diff
@@ -1659,6 +1659,13 @@ func validateKubernetesVersion(old *config.ClusterConfig) {
 	newestVersion := semver.MustParse(strings.TrimPrefix(constants.NewestKubernetesVersion, version.VersionPrefix))
 	zeroVersion := semver.MustParse(strings.TrimPrefix(constants.NoKubernetesVersion, version.VersionPrefix))
 
+	paramVersion := viper.GetString(kubernetesVersion)
+	paramVersion = strings.TrimPrefix(paramVersion, version.VersionPrefix)
+	isMajorMinorOnly, _ := getSemverFromMajorMinorOnly(paramVersion)
+
+	if isMajorMinorOnly {
+		out.Step(style.Workaround, `Using Kubernetes {{.version}} since patch version was unspecified`, out.V{"version": nvs})
```
medyagh, line 1667: we should out.Styled here

medyagh (changes requested):

cmd/minikube/cmd/start.go
```diff
@@ -1659,6 +1659,13 @@ func validateKubernetesVersion(old *config.ClusterConfig) {
 	newestVersion := semver.MustParse(strings.TrimPrefix(constants.NewestKubernetesVersion, version.VersionPrefix))
 	zeroVersion := semver.MustParse(strings.TrimPrefix(constants.NoKubernetesVersion, version.VersionPrefix))
 
+	paramVersion := viper.GetString(kubernetesVersion)
```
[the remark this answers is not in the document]
JudahNour: Capital V does not work. Do we want it so that the paramVersion is completely set to lowercase so that this is accounted for? Or do we want it to specifically set the first character to lowercase? I think the former would mean someone could input something like "StAble" as the flag and it would recognize it as "stable"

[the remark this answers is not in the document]
medyagh: yeah lets convert the user input to lowercase for all cases, that makes our life easier

cmd/minikube/cmd/start.go
```diff
@@ -1766,14 +1799,19 @@ $ minikube config unset kubernetes-version`)
...
-		kubernetesSemver = getLatestPatch(kubernetesSemver)
+		potentialPatch := getLatestPatch(kubernetesSemver)
+		if potentialPatch != `` {
+			kubernetesSemver = potentialPatch
+		} else {
+			return potentialPatch, ErrPatchNotFound
+		}
 	}
 	nvs, err := semver.Make(kubernetesSemver)
 	if err != nil {
```
JudahNour, line 1810: Not sure if here we want to send back an error here instead of just exiting. With the error return value it could  be better to now return an error and handle it outside getKubernetesVersion

cmd/minikube/cmd/start.go
```diff
@@ -88,6 +88,9 @@ type versionJSON struct {
 	Commit          string `json:"commit"`
 }
 
+// ErrPatchNotFound is when given volume was not found
+var ErrPatchNotFound = errors.New("Unable to detect the latest patch release for specified major.minor version")
```
medyagh, line 92: this error message doesn't mention it is talking about Kubernetes version, that could be ambiguous to the user.

how about something like this
ErrKubernetesPatchNotFound = errors.New("Unable to detect the latest patch release for specified Kubernetes version")

cmd/minikube/cmd/start.go
```diff
@@ -318,7 +321,10 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 		stopk8s = true
 	}
 
-	k8sVersion := getKubernetesVersion(existing)
+	k8sVersion, err := getKubernetesVersion(existing)
+	if err != nil {
+		klog.Warningf("get kubernetesVersion failed : %v", err)
```
medyagh, line 326: to be consistent  with other logs warning:
"failed to get kubernetes version"

cmd/minikube/cmd/start.go
```diff
@@ -1758,12 +1797,21 @@ $ minikube config unset kubernetes-version`)
...
 	}
 
-	nvs, err := semver.Make(strings.TrimPrefix(paramVersion, version.VersionPrefix))
+	kubernetesSemver := strings.TrimPrefix(strings.ToLower(paramVersion), version.VersionPrefix)
+	if isTwoDigitSemver(kubernetesSemver) {
+		potentialPatch := getLatestPatch(kubernetesSemver)
+		if potentialPatch != `` {
+			kubernetesSemver = potentialPatch
+		} else {
+			return potentialPatch, ErrPatchNotFound
```
medyagh, line 1806: nit picking:
```return "", ErrPatchNotFound```
would be more clear to read

File: cmd/minikube/cmd/start.go (hunk 4 above)
Comment (medyagh) [reply cmd/minikube/cmd/start.go:1810]: good point and good catch ! I think since it is already been implemented like that we can leave this one alone, and if we want we can refactor it in a follow up PR (for sake of keeping this PR small)

File: cmd/minikube/cmd/start.go [hunk 8]
```diff
@@ -88,6 +88,9 @@ type versionJSON struct {
 	Commit          string `json:"commit"`
 }
 
+// ErrPatchNotFound is when given volume was not found
```
spowelljr, line 91:
```suggestion
// ErrPatchNotFound is when given version was not found
```
I'm guessing that was meant to be version?

cmd/minikube/cmd/start.go
```diff
@@ -1878,3 +1930,20 @@ func exitGuestProvision(err error) {
...
+	for _, k := range constants.ValidKubernetesVersions {
+		if strings.HasPrefix(k, fmt.Sprintf("v%s", majorMinorVer)) {
+			return strings.TrimPrefix(k, version.VersionPrefix)
+		}
+
+	}
+	return ""
+}
+
+func isTwoDigitSemver(ver string) bool {
```
medyagh, line 1946: lets write a unit test for this in start_test.go

medyagh (changes requested):

medyagh:
@JudahNour thank for updating the PR, do you mind also re-doing the Before/After after comming the last changes to ensure it is still good

cmd/minikube/cmd/start.go
spowelljr, line 326:
```suggestion
		klog.Warningf("failed getting Kubernetes version: %v", err)
```

cmd/minikube/cmd/start.go
```diff
@@ -1585,7 +1591,13 @@ func createNode(cc config.ClusterConfig, existing *config.ClusterConfig) (config
 	// Create the initial node, which will necessarily be a control plane
 	if existing != nil {
 		cp, err := config.PrimaryControlPlane(existing)
-		cp.KubernetesVersion = getKubernetesVersion(&cc)
+
+		newVer, err2 := getKubernetesVersion(&cc)
+		cp.KubernetesVersion = newVer
+
+		if err2 != nil {
+			klog.Warningf("get kubernetesVersion failed : %v", err2)
```
spowelljr, line 1599:
```suggestion
			klog.Warningf("failed getting Kubernetes version: %v", err2)
```

cmd/minikube/cmd/start.go
```diff
@@ -1595,7 +1607,10 @@ func createNode(cc config.ClusterConfig, existing *config.ClusterConfig) (config
 		// KubernetesVersion is the only attribute that the user can override in the Node object
 		nodes := []config.Node{}
 		for _, n := range existing.Nodes {
-			n.KubernetesVersion = getKubernetesVersion(&cc)
+			n.KubernetesVersion, err = getKubernetesVersion(&cc)
+			if err != nil {
+				klog.Warningf("get kubernetesVersion failed : %v", err)
```
spowelljr, line 1612:
```suggestion
				klog.Warningf("failed getting Kubernetes version: %v", err)
```

cmd/minikube/cmd/start.go
```diff
@@ -1604,9 +1619,13 @@ func createNode(cc config.ClusterConfig, existing *config.ClusterConfig) (config
 		return cc, cp, nil
 	}
 
+	kubeVer, err := getKubernetesVersion(&cc)
+	if err != nil {
+		klog.Warningf("get kubernetesVersion failed : %v", err)
```
spowelljr, line 1624:
```suggestion
		klog.Warningf("failed getting Kubernetes version: %v", err)
```

cmd/minikube/cmd/start.go
```diff
@@ -1653,12 +1672,32 @@ func autoSetDriverOptions(cmd *cobra.Command, drvName string) (err error) {
...
 	oldestVersion := semver.MustParse(strings.TrimPrefix(constants.OldestKubernetesVersion, version.VersionPrefix))
 	defaultVersion := semver.MustParse(strings.TrimPrefix(constants.DefaultKubernetesVersion, version.VersionPrefix))
 	newestVersion := semver.MustParse(strings.TrimPrefix(constants.NewestKubernetesVersion, version.VersionPrefix))
 	zeroVersion := semver.MustParse(strings.TrimPrefix(constants.NoKubernetesVersion, version.VersionPrefix))
 
+	paramVersion := viper.GetString(kubernetesVersion)
+	paramVersion = strings.TrimPrefix(strings.ToLower(paramVersion), version.VersionPrefix)
+
+	if isTwoDigitSemver(paramVersion) {
+		if getLatestPatch(paramVersion) != `` {
```
spowelljr, line 1697:
```suggestion
		if getLatestPatch(paramVersion) != "" {
```

cmd/minikube/cmd/start.go
```diff
@@ -1758,12 +1797,21 @@ $ minikube config unset kubernetes-version`)
 		paramVersion = constants.NewestKubernetesVersion
 	}
 
-	nvs, err := semver.Make(strings.TrimPrefix(paramVersion, version.VersionPrefix))
+	kubernetesSemver := strings.TrimPrefix(strings.ToLower(paramVersion), version.VersionPrefix)
+	if isTwoDigitSemver(kubernetesSemver) {
+		potentialPatch := getLatestPatch(kubernetesSemver)
+		if potentialPatch != `` {
```
spowelljr, line 1803:
```suggestion
		if potentialPatch != "" {
```

cmd/minikube/cmd/start.go
```diff
@@ -1848,7 +1896,11 @@ func validateBareMetal(drvName string) {
 	}
 
 	// conntrack is required starting with Kubernetes 1.18, include the release candidates for completion
-	version, _ := util.ParseKubernetesVersion(getKubernetesVersion(nil))
+	kubeVer, err := getKubernetesVersion(nil)
+	if err != nil {
+		klog.Warningf("get kubernetesVersion failed : %v", err)
```
spowelljr, line 1901:
```suggestion
		klog.Warningf("failed getting Kubernetes version: %v", err)
```

spowelljr (changes requested):

cmd/minikube/cmd/start.go
```diff
@@ -1585,7 +1591,13 @@ func createNode(cc config.ClusterConfig, existing *config.ClusterConfig) (config
...
-		cp.KubernetesVersion = getKubernetesVersion(&cc)
+
+		newVer, err2 := getKubernetesVersion(&cc)
+		cp.KubernetesVersion = newVer
+
+		if err2 != nil {
+			klog.Warningf("get kubernetesVersion failed : %v", err2)
+		}
 		cp.ContainerRuntime = getContainerRuntime(&cc)
 		if err != nil {
```
spowelljr, line 1602: This was setup wrong from the start, but there shouldn't be any lines of code between `cp, err := config.PrimaryControlPlane(existing)` and `if err != nil {`

So this should be reordered to:
```
cp, err := config.PrimaryControlPlane(existing)
if err != nil {
        return cc, config.Node{}, err
}
newVer, err := getKubernetesVersion(&cc)
...
```
And then you can change `err2` back to `err`

spowelljr (changes requested):

cmd/minikube/cmd/start.go
```diff
@@ -1878,3 +1930,20 @@ func exitGuestProvision(err error) {
 	}
 	exit.Error(reason.GuestProvision, "error provisioning guest", err)
 }
+
+// Example input = 1.26 => output = "1.26.5"
+// Example input = 1.26.5 => output = ""
```
spowelljr, line 1935: This comment is false, if `1.26.5` it will return `1.26.5` as `v1.26.5` is a prefix of `v1.26.5`

spowelljr (changes requested):

cmd/minikube/cmd/start_flags.go
```diff
@@ -811,7 +811,11 @@ func updateExistingConfigFromFlags(cmd *cobra.Command, existing *config.ClusterC
 	updateStringFromFlag(cmd, &cc.SocketVMnetPath, socketVMnetPath)
 
 	if cmd.Flags().Changed(kubernetesVersion) {
-		cc.KubernetesConfig.KubernetesVersion = getKubernetesVersion(existing)
+		kubeVer, err := getKubernetesVersion(existing)
+		if err != nil {
+			klog.Warningf("get kubernetesVersion failed : %v", err)
```
spowelljr, line 816:
```suggestion
			klog.Warningf("failed getting Kubernetes version: %v", err)
```

cmd/minikube/cmd/start_test.go
```diff
@@ -92,7 +103,10 @@ func TestGetKubernetesVersion(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.description, func(t *testing.T) {
 			viper.SetDefault(kubernetesVersion, test.paramVersion)
-			version := getKubernetesVersion(test.cfg)
+			version, err := getKubernetesVersion(test.cfg)
+			if err != nil {
+				klog.Warningf("get kubernetesVersion failed : %v", err)
```
spowelljr, line 108:
```suggestion
				klog.Warningf("failed getting Kubernetes version: %v", err)
```

spowelljr (changes requested):

cmd/minikube/cmd/start.go
JudahNour: Meant to be patch was not found. Should it be version instead?

JudahNour: Well actually it would be when the patch isnt found when given the major.minor

cmd/minikube/cmd/start.go
JudahNour: Got it! This was leftover from when isTwoDIgitSemver was a part of this function.

cmd/minikube/cmd/start.go
```diff
@@ -88,6 +88,9 @@ type versionJSON struct {
 	Commit          string `json:"commit"`
 }
 
+// ErrKubernetesPatchNotFound is when a patch was not found for the given <major>.<minor> version
+var ErrKubernetesPatchNotFound = errors.New("Unable to detect the latest patch release for specified Kubernetes version")
+
 var (
```
spowelljr, line 94:
```suggestion
var (
	// ErrKubernetesPatchNotFound is when a patch was not found for the given <major>.<minor> version
	ErrKubernetesPatchNotFound = errors.New("Unable to detect the latest patch release for specified Kubernetes version")
```
nit: This could go in the vars below

cmd/minikube/cmd/start.go
```diff
@@ -1585,17 +1591,24 @@ func createNode(cc config.ClusterConfig, existing *config.ClusterConfig) (config
 	// Create the initial node, which will necessarily be a control plane
 	if existing != nil {
 		cp, err := config.PrimaryControlPlane(existing)
-		cp.KubernetesVersion = getKubernetesVersion(&cc)
-		cp.ContainerRuntime = getContainerRuntime(&cc)
 		if err != nil {
 			return cc, config.Node{}, err
 		}
+		newVer, err := getKubernetesVersion(&cc)
+		cp.KubernetesVersion = newVer
```
spowelljr, line 1598:
```suggestion
		cp.KubernetesVersion, err = getKubernetesVersion(&cc)
```
There shouldn't be anything between the declaration of an error and the error check, so `cp.KubernetesVersion = newVer` should be after the error check. But because the `newVer` is only being used to set `cp.KubernetesVersion` we can just set it directly

cmd/minikube/cmd/start.go
```diff
@@ -1653,12 +1670,32 @@ func autoSetDriverOptions(cmd *cobra.Command, drvName string) (err error) {
...
 	zeroVersion := semver.MustParse(strings.TrimPrefix(constants.NoKubernetesVersion, version.VersionPrefix))
 
+	paramVersion := viper.GetString(kubernetesVersion)
+	paramVersion = strings.TrimPrefix(strings.ToLower(paramVersion), version.VersionPrefix)
+
+	if isTwoDigitSemver(paramVersion) {
+		if getLatestPatch(paramVersion) != "" {
+			out.Styled(style.Workaround, `Using Kubernetes {{.version}} since patch version was unspecified`, out.V{"version": nvs})
+		}
+	}
```
spowelljr, line 1698:
```suggestion
	if isTwoDigitSemver(paramVersion) && getLatestPatch(paramVersion) != "" {
		out.Styled(style.Workaround, `Using Kubernetes {{.version}} since patch version was unspecified`, out.V{"version": nvs})
	}
```
Unnecessary double nesting

cmd/minikube/cmd/start.go
```diff
@@ -1878,3 +1928,21 @@ func exitGuestProvision(err error) {
 	}
 	exit.Error(reason.GuestProvision, "error provisioning guest", err)
 }
+
+// Example input = 1.26 => output = "1.26.5"
+// Example input = 1.26.5 => output = "1.26.5"
+// Example input = 1.26.6 => output = ""
```
spowelljr, line 1934:
```suggestion
// Example input = 1.26.999 => output = ""
```
Might want to make this a bit more explicit, `1.26.6` will be coming in the future and someone reading this comment might be confused why `1.26.5` is fine but `1.26.6` isn't

cmd/minikube/cmd/start.go
```diff
@@ -1758,12 +1795,21 @@ $ minikube config unset kubernetes-version`)
...
 
-	nvs, err := semver.Make(strings.TrimPrefix(paramVersion, version.VersionPrefix))
+	kubernetesSemver := strings.TrimPrefix(strings.ToLower(paramVersion), version.VersionPrefix)
+	if isTwoDigitSemver(kubernetesSemver) {
+		potentialPatch := getLatestPatch(kubernetesSemver)
+		if potentialPatch != "" {
+			kubernetesSemver = potentialPatch
+		} else {
+			return "", ErrKubernetesPatchNotFound
+		}
```
spowelljr, line 1805:
```suggestion
		if potentialPatch == "" {
			return "", ErrKubernetesPatchNotFound
		}
		kubernetesSemver = potentialPatch
```
nit: Due to the return, flipping the if can reduce nesting

cmd/minikube/cmd/start_test.go
```diff
@@ -451,6 +465,99 @@ func TestValidateRuntime(t *testing.T) {
...
+			version:  ".2",
+			expected: false,
+		},
+		{
+			desc:     "a valid two digit version with whitespace between components",
+			version:  "1. 1",
+			expected: false,
+		},
+		{
+			desc:     "a two digit version with a nondigit major component",
```
spowelljr, line 530:
```suggestion
			desc:     "a two digit version with a non-digit major component",
```

cmd/minikube/cmd/start_test.go
```diff
@@ -451,6 +465,99 @@ func TestValidateRuntime(t *testing.T) {
...
+			version:  "a.12",
+			expected: false,
+		},
+		{
+			desc:     "a two digit vesion with a nondigit minor component",
```
spowelljr, line 535:
```suggestion
			desc:     "a two digit version with a non-digit minor component",
```

cmd/minikube/cmd/start_test.go
```diff
@@ -451,6 +465,99 @@ func TestValidateRuntime(t *testing.T) {
...
+			version:  "1.a",
+			expected: false,
+		},
+		{
+			desc:     "a two digit vesion with extraneous nondigits in minor component",
```
spowelljr, line 540:
```suggestion
			desc:     "a two digit version with extraneous non-digits in minor component",
```

cmd/minikube/cmd/start_test.go
```diff
@@ -451,6 +465,99 @@ func TestValidateRuntime(t *testing.T) {
...
+			version:  "0.0",
+			expected: true,
+		},
+		{
+			desc:     "a two digit version with negative major version",
+			version:  "-1.0",
+			expected: false,
+		},
+		{
+			desc:     "a two digit vesion with negative minor version",
```
spowelljr, line 510:
```suggestion
			desc:     "a two digit version with negative minor version",
```

cmd/minikube/cmd/start_test.go
```diff
@@ -451,6 +465,99 @@ func TestValidateRuntime(t *testing.T) {
...
+			version:  "1.a",
+			expected: false,
+		},
+		{
+			desc:     "a two digit vesion with extraneous nondigits in minor component",
+			version:  "1.2a",
+			expected: false,
+		},
+		{
+			desc:     "a two digit vesion larger major/minor components",
```
spowelljr, line 545:
```suggestion
			desc:     "a two digit version larger major/minor components",
```

spowelljr (changes requested):
Run `make generate-docs` to update translations

cmd/minikube/cmd/start_test.go
```diff
@@ -451,6 +465,99 @@ func TestValidateRuntime(t *testing.T) {
...
+			version:  "123456789.987654321",
+			expected: true,
+		},
+	}
+	for _, tc := range tcs {
+		t.Run(tc.desc, func(t *testing.T) {
+			actual := isTwoDigitSemver(tc.version)
+			// check whether the function correctly verifies if it is a 2 digit semver
+			if actual != tc.expected {
+				t.Fatalf("test failed. Expected version v%s to return %v", tc.version, tc.expected)
```
spowelljr, line 555:
```suggestion
				t.Fatalf("test failed. Expected version v%s to return %t", tc.version, tc.expected)
```
Bool format

spowelljr (changes requested):

cmd/minikube/cmd/start_test.go
```diff
@@ -451,6 +465,99 @@ func TestValidateRuntime(t *testing.T) {
...
+		{
+			desc:     "a two digit version with extraneous non-digits in minor component",
+			version:  "1.2a",
+			expected: false,
+		},
+		{
+			desc:     "a two digit version larger major/minor components",
+			version:  "123456789.987654321",
+			expected: true,
+		},
```
spowelljr, line 548: Could you also add a test with a `v` prefix

cmd/minikube/cmd/start_test.go
```diff
@@ -451,6 +465,99 @@ func TestValidateRuntime(t *testing.T) {
...
+	}
+	for _, tc := range tcs {
+		t.Run(tc.desc, func(t *testing.T) {
+			actual := isTwoDigitSemver(tc.version)
+			// check whether the function correctly verifies if it is a 2 digit semver
+			if actual != tc.expected {
+				t.Fatalf("test failed. Expected version v%s to return %t", tc.version, tc.expected)
```
spowelljr, line 555:
```suggestion
				t.Fatalf("test failed. Expected version %s to return %t", tc.version, tc.expected)
```
The input doesn't have a `v` prefix, so for clarity it should be omitted, especially when combined with my above comment

spowelljr (changes requested):

cmd/minikube/cmd/start.go
```diff
@@ -1653,12 +1668,30 @@ func autoSetDriverOptions(cmd *cobra.Command, drvName string) (err error) {
...
+	}
+
+	nvs, _ := semver.Make(strings.TrimPrefix(kubernetesVer, version.VersionPrefix))
 	oldestVersion := semver.MustParse(strings.TrimPrefix(constants.OldestKubernetesVersion, version.VersionPrefix))
 	defaultVersion := semver.MustParse(strings.TrimPrefix(constants.DefaultKubernetesVersion, version.VersionPrefix))
 	newestVersion := semver.MustParse(strings.TrimPrefix(constants.NewestKubernetesVersion, version.VersionPrefix))
 	zeroVersion := semver.MustParse(strings.TrimPrefix(constants.NoKubernetesVersion, version.VersionPrefix))
 
+	paramVersion := viper.GetString(kubernetesVersion)
+	paramVersion = strings.TrimPrefix(strings.ToLower(paramVersion), version.VersionPrefix)
```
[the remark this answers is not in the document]
JudahNour: I made the first change, but I don't think we can just replace line 1683.  kubernetesVer accounts for stable and latest as those checks/modifications are done in getKubernetesVersion. I think removing the use of getKubernetesVersion (kubernetesVer) here wouldn't make sense as that is the actual kubernetes version we have while the paramVersion can be something like 1.26 which we want so that we can tell the user their major.minor version not having a patch (as in the second before/after).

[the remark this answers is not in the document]
spowelljr: Ah, I missed that it's using `kubernetesVer` instead of `paramVersion` so ignore the line 1683 suggestion, I think the rest of the comment still stands though

medyagh (approved):

medyagh:
looks good to me ! Congratulations  on your first contribution ! looking forward to see more !

Issue #13740, opened 2022-03-02
Can't get ingress-dns running on Ubuntu 20.04


jkempf42:
### What Happened?

I can't seem to get ingress-dns running on Ubuntu 20.04. I tried the installation and configuration suggestions in the installation instructions [here](https://minikube.sigs.k8s.io/docs/handbook/addons/ingress-dns/) and the instructions in the issue list for this repo  [here](https://github.com/kubernetes/minikube/issues/10517) but I could not get the basic `hello-john.test` and `hello-jane.test `example working. I tried both enabling `dnsmasq `and configuring `systemd-resolved`  as described in the two links above. 

The last thing I tried was to disable name resolution through` NetworkManager.service` by putting` dns=none` in `/etc/NetworkManager/NetworkManager.conf` and restarting `NetworkManager`, then focusing on systemd-resolved. I added a config file `01-minikube.config` to` /usr/lib/systemd/resolved.conf.d/ `as follows;
```
          [Resolve]
          DNS=192.168.0.1 192.168.49.2
```
which should have added the` minikube` ip as a name server along with my existing name server, restarted `systemd-resolved `through `systemctl` but that didn't work either. 

Any suggestions?

### Attach the log file

[log.txt](https://github.com/kubernetes/minikube/files/8166636/log.txt)

### Operating System

Ubuntu

### Driver

Docker

r0ushann:
hey! I'm interested in this issue.

tonyrizzotto:
@jak42 I think we're having the same issue. Just curious, are you able to access your app on a NodePort service? I believe this issue is related to the Minikube Ingress not installing on Linux. To confirm, I'd suggest deleting your minikube cluster with `minikube delete --purge --all` and run `minikube start --addons=ingress --alsologtostderr` and see in the ingress installation hangs. You'll notice something like: 
```
I0313 14:39:53.031414  177055 kapi.go:96] waiting for pod "app.kubernetes.io/name=ingress-nginx", current state: Pending: [<nil>]
I0313 14:39:53.502730  177055 kapi.go:96] waiting for pod "app.kubernetes.io/name=ingress-nginx", current state: Pending: [<nil>]
I0313 14:39:54.006302  177055 kapi.go:96] waiting for pod "app.kubernetes.io/name=ingress-nginx", current state: Pending: [<nil>]
```
At which point it'll time out and say the addon was installed properly. However when you do a `kubectl get pod -n kube-system`

jurca:
[...]

PR #16575, opened 2023-05-24
docs(dns): improve ubuntu instructions

kaotika:
According to issue https://github.com/kubernetes/minikube/issues/13740#issuecomment-1235711158 this change is needed

### DIFF
--- a/site/content/en/docs/handbook/addons/ingress-dns.md
+++ b/site/content/en/docs/handbook/addons/ingress-dns.md
@@ -109,6 +109,7 @@ echo "server=/test/$(minikube ip)" >/etc/NetworkManager/dnsmasq.d/minikube.conf
 
 Restart Network Manager
 ```
+sudo rm /etc/resolv.conf # will be needed, at least on ubuntu 22.04
 systemctl restart NetworkManager.service
 ```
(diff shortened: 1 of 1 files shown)


site/content/en/docs/handbook/addons/ingress-dns.md
```diff
@@ -109,6 +109,7 @@ echo "server=/test/$(minikube ip)" >/etc/NetworkManager/dnsmasq.d/minikube.conf
 
 Restart Network Manager
 ```
+sudo rm /etc/resolv.conf # will be needed, at least on ubuntu 22.04
```
medyagh, line 112: this seems to be too invasive and not probably gonna destroy a lot of peoples workflows. I suggest that you provide a command that backs up the current reseolv.conf and then if user wants to revert that they can restore it.

medyagh (changes requested):

medyagh:
@kaotika thanks for contributing to minikube please ensure you sign the CLA agreement https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/18706487/56353740/16575/#/?version=2

PR #16577, opened 2023-05-24
addons registry: Use multi-arch kube-registry-proxy image

spowelljr:
The existing kube-registry-proxy image (`gcr.io/google_containers/kube-registry-proxy`) was only built for amd64 and doesn't work for users using QEMU on arm64 machines.

The source code for the image was removed in https://github.com/kubernetes/kubernetes/pull/58564.

I took the source code, updated the nginx base image and added it to https://github.com/spowelljr/kube-registry-proxy and built the kube-registry-proxy image for our five supported architectures and pushed it to `gcr.io/k8s-minikube/kube-registry-proxy`.

It was confirmed to work by another user as well.
https://kubernetes.slack.com/archives/C1F5CT6Q1/p1684942501397369?thread_ts=1684931469.550259&cid=C1F5CT6Q1

I'm happy to move the source code into the project or into some minikube owned repo.

### DIFF
--- a/pkg/minikube/assets/addons.go
+++ b/pkg/minikube/assets/addons.go
@@ -389,7 +389,7 @@ var Addons = map[string]*Addon{
 			"0640"),
 	}, false, "registry", "Google", "", "", map[string]string{
 		"Registry":          "registry:2.8.1@sha256:83bb78d7b28f1ac99c68133af32c93e9a1c149bcd3cb6e683a3ee56e312f1c96",
-		"KubeRegistryProxy": "google_containers/kube-registry-proxy:0.4@sha256:1040f25a5273de0d72c54865a8efd47e3292de9fb8e5353e3fa76736b854f2da",
+		"KubeRegistryProxy": "k8s-minikube/kube-registry-proxy:0.0.1@sha256:498e7da3b171e24630615854c4a2924cfb8adbd43b0f252b1458c859ca9b746c",
 	}, map[string]string{
 		"KubeRegistryProxy": "gcr.io",
 		"Registry":          "docker.io",
--- a/pkg/minikube/assets/addons.go
+++ b/pkg/minikube/assets/addons.go
@@ -389,7 +389,7 @@ var Addons = map[string]*Addon{
 			"0640"),
 	}, false, "registry", "Google", "", "", map[string]string{
 		"Registry":          "registry:2.8.1@sha256:83bb78d7b28f1ac99c68133af32c93e9a1c149bcd3cb6e683a3ee56e312f1c96",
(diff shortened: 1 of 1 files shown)


pkg/minikube/assets/addons.go
```diff
@@ -389,7 +389,7 @@ var Addons = map[string]*Addon{
 			"0640"),
 	}, false, "registry", "Google", "", "", map[string]string{
 		"Registry":          "registry:2.8.1@sha256:83bb78d7b28f1ac99c68133af32c93e9a1c149bcd3cb6e683a3ee56e312f1c96",
-		"KubeRegistryProxy": "google_containers/kube-registry-proxy:0.4@sha256:1040f25a5273de0d72c54865a8efd47e3292de9fb8e5353e3fa76736b854f2da",
+		"KubeRegistryProxy": "k8s-minikube/kube-registry-proxy:0.0.1@sha256:498e7da3b171e24630615854c4a2924cfb8adbd43b0f252b1458c859ca9b746c",
```
medyagh, line 392: lets create an issue on Kubernetes and ask them where they plan to maintain the deleted code ?

medyagh (changes requested):

pkg/minikube/assets/addons.go
medyagh: @BenTheElder mentioned anything under cluster has been depricated, so I think in minikube we should create an issue upstream to see if there is no other sponsor, minikube could host it on our repo

medyagh (changes requested):

pkg/minikube/assets/addons.go
BenTheElder: Kubernetes/Kubernetes will not host this anymore since kube-up CI is not using it and excluding the etcd image this directory is only used for CI and not supported anymore (see cluster/README.md) but minikube could host these sources and iterate on it or SIG cluster lifecycle could sponsor bringing it back as a SIG project repo. I suspect the latter will not be supported though based on past conversations.

BenTheElder: SIG Testing + Infra is also trying to phase out the rest as well because we need to be running CI on multiple clouds instead of only the GCE bash. We're starting small with just one scale test on kops-aws but the hope is to use kOps AWS+GCE instead of kube-up GCE more generally, eventually. So there's not much future for anything in cluster/. It's barely maintained as-is out of necessity for CI.

mprimeaux:
@medyagh I am just catching up with this PR. I had commented [here](https://github.com/kubernetes/minikube/issues/14146#issuecomment-1563459990). Let me know if I can help.

pkg/minikube/assets/addons.go
medyagh: since there is no owner for this image, we could just publish one last patch of it in arm64 but in a follow up PR we could change the ownership of this addon from Google to minikube

medyagh (changes requested):

medyagh (approved):

PR #16600, opened 2023-05-30
Makefile: refactor buildx related targets

spowelljr:
Removed `X_BUILD_ENV ?= DOCKER_CLI_EXPERIMENTAL=enabled`

Why?

As per Docker documentation: `Starting with Docker 20.10, experimental CLI features are enabled by default, and require no configuration to enable them.`

https://docs.docker.com/engine/reference/commandline/cli/#experimental-features

-----

Removed:
```
env $(X_BUILD_ENV) docker buildx rm --builder $(X_DOCKER_BUILDER) || true
env $(X_BUILD_ENV) docker buildx create --name $(X_DOCKER_BUILDER) --buildkitd-flags '--debug' || true
--builder $(X_DOCKER_BUILDER)
```

Why?

We no longer need to create a new buildx instance, we just need to ensure QEMU binaries are installed.

-----

Replaced:
```
env $(X_BUILD_ENV) docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
```
With:
```
docker run --rm --privileged tonistiigi/binfmt:latest --install all
```

Why?

`docker run --rm --privileged tonistiigi/binfmt:latest --install all` is what is used inside https://github.com/docker/setup-qemu-action

-----

Removed `--load` from `docker buildx build` commands

Why?

`--load` is required when using a buildx instance with the `docker-container` driver to push the built image to `docker images`, otherwise the built image just sits in cache, but with the new method we're using the default build instance so we can omit it

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -698,14 +698,10 @@ storage-provisioner-image-%: out/storage-provisioner-%
 	docker build -t $(REGISTRY)/storage-provisioner-$*:$(STORAGE_PROVISIONER_TAG) -f deploy/storage-provisioner/Dockerfile  --build-arg arch=$* .
 
 
-X_DOCKER_BUILDER ?= minikube-builder
-X_BUILD_ENV ?= DOCKER_CLI_EXPERIMENTAL=enabled
-
-.PHONY: docker-multi-arch-builder
-docker-multi-arch-builder:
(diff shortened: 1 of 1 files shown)


spowelljr:
ok-to-build-image

PR #16610, opened 2023-05-31
CI: Automate updating Kubernetes versions list

spowelljr:
GitHub Actions cron job that will update the `ValidKubernetesVersions` list


cmd/minikube/cmd/config/kubernetes_version.go
```diff
@@ -34,6 +32,5 @@ func supportedKubernetesVersions() (releases []string) {
 		}
 		releases = append(releases, ver)
 	}
-	sort.Slice(releases, func(i, j int) bool { return semver.Compare(releases[i], releases[j]) == 1 })
```
spowelljr, line 37: Removed this because the automation is going to ensure the list is in reverse sorted order, so we don't have to do it in the command itself

PR #16640, opened 2023-06-06
CI: Automate updating calico

spowelljr:
GitHub Action that runs on a cron to update the calico CNI


medyagh:
@spowelljr needs rebase

PR #16655, opened 2023-06-08
VM drivers: Fix images getting removed on stop/start

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/12217

For full details see: https://github.com/kubernetes/minikube/issues/12217#issuecomment-1583303474

The short is that on a VM driver using the containerd or cri-o runtime, when starting after a stop, we would check for the existence of preload images prior to CRIs being configured and started, resulting in an empty image list being returned. This would then trigger to untar the preload and overwrite all the existing image on the machine. There's another preload check later on after the CRIs are configured and started, which is why the wasn't an issue with Docker.

### DIFF
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -929,7 +929,12 @@ func (k *Bootstrapper) UpdateCluster(cfg config.ClusterConfig) error {
 	}
 
 	if err := r.Preload(cfg); err != nil {
-		klog.Infof("preload failed, will try to load cached images: %v", err)
+		switch err.(type) {
+		case *cruntime.ErrISOFeature:
+			out.ErrT(style.Tip, "Existing disk is missing new features ({{.error}}). To upgrade, run 'minikube delete'", out.V{"error": err})
+		default:
+			klog.Infof("preload failed, will try to load cached images: %v", err)
+		}
 	}
 
 	if cfg.KubernetesConfig.ShouldLoadCachedImages {
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -413,23 +413,6 @@ func configureRuntimes(runner cruntime.CommandRunner, cc config.ClusterConfig, k
 		klog.Errorf("unable to disable preinstalled bridge CNI(s): %v", err)
 	}
 
-	// Preload is overly invasive for bare metal, and caching is not meaningful.
-	// KIC handles preload elsewhere.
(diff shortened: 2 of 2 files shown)


pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -929,7 +929,12 @@ func (k *Bootstrapper) UpdateCluster(cfg config.ClusterConfig) error {
 	}
 
 	if err := r.Preload(cfg); err != nil {
-		klog.Infof("preload failed, will try to load cached images: %v", err)
+		switch err.(type) {
+		case *cruntime.ErrISOFeature:
+			out.ErrT(style.Tip, "Existing disk is missing new features ({{.error}}). To upgrade, run 'minikube delete'", out.V{"error": err})
```
medyagh, line 934: If deleteOnFailure flag is specified minikube should delete and recreate

spowelljr: You want that added to the message?

medyagh: hm...I was thinking maybe if the user allows us to delete minikube we just do it
if --deleteOnFailure...then recreate the cluster instead of telling the user to do it..

medyagh (changes requested):

medyagh:
Really good job on this PR @spowelljr 

```
KVM_Linux — Jenkins: completed with success in 62.81 minutes.
Details
KVM_Linux_containerd — Jenkins: completed with success in 59.66 minutes.

```


and also makes KVM Containerd test much faster
before this  PR it was taking 70+ mins to finish KVM containerd now 59 mins
11 min faster KVM test

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
spowelljr: I tried using `--delete-on-failure`, but it seems that it only applies to creating the cluster, since the cluster is already created once we get here and this is K8s logic it doesn't retry. If we want this I'd recommend it be in a separate PR since there will have to be a lot of logic added for it.

```
minikube start --driver qemu --delete-on-failure=true
😄  minikube v1.30.1 on Darwin 13.4 (arm64)
🌐  Automatically selected the socket_vmnet network
🐳  Preparing Kubernetes v1.27.2 on Docker 24.0.2 ...

❌  Exiting due to K8S_INSTALL_FAILED: Failed to update cluster: this error is expected

```

medyagh (approved):

PR #16662, opened 2023-06-09
CI: Automate updating hide-minikube-bot-comments action

spowelljr:


hack/update/golang_version/update_golang_version.go
```diff
@@ -36,41 +36,42 @@ var (
...
+		".github/workflows/update-hugo-version.yml":                       workflowReplace,
+		".github/workflows/update-ingress-version.yml":                    workflowReplace,
+		".github/workflows/update-inspektor-gadget-version.yml":           workflowReplace,
+		".github/workflows/update-k8s-versions.yml":                       workflowReplace,
+		".github/workflows/update-kubeadm-constants.yml":                  workflowReplace,
+		".github/workflows/update-kubernetes-versions-list.yml":           workflowReplace,
+		".github/workflows/update-metrics-server-version.yml":             workflowReplace,
+		".github/workflows/update-runc-version.yml":                       workflowReplace,
+		".github/workflows/update-ubuntu-version.yml":                     workflowReplace,
+		".github/workflows/yearly-leaderboard.yml":                        workflowReplace,
```
medyagh, line 74: I wonder can we do a regex all the yaml and yml files in the .github/workflows instead of adding them one by one?

medyagh (changes requested):

hack/update/golang_version/update_golang_version.go
spowelljr: That's a good idea, can walk the directory and add them

PR #16711, opened 2023-06-14
fix legacy in test missing container

medyagh:
arm64 did not exist in 1.9.1 and we should have used legacyversion per environment


spowelljr (approved):
Awaiting test results, but looks good to me

PR #16756, opened 2023-06-22
gvisor addon: Only allow on clusters that support running amd64 images

spowelljr:
Added a validation function that checks if cluster can support running amd64 images and added it to the gvisor addons validations.

### DIFF
--- a/pkg/addons/validations.go
+++ b/pkg/addons/validations.go
@@ -18,12 +18,14 @@ package addons
 
 import (
 	"fmt"
+	"runtime"
 	"strconv"
 
 	"github.com/spf13/viper"
 	"k8s.io/minikube/pkg/minikube/assets"
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/cruntime"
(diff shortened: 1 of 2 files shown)


pkg/addons/validations.go
```diff
@@ -93,3 +95,15 @@ func contains(slice []string, val string) bool {
 	}
 	return false
 }
+
+// SupportsAmd64 ensures that the cluster supports running amd64 images
+func SupportsAmd64(cc *config.ClusterConfig, name, _ string) error {
```
medyagh, line 100: lets see if we can run it through translations

medyagh (changes requested):

pkg/addons/validations.go
medyagh: lets investigate what are the steps to do to run through translation

spowelljr: arm64 can be supported by installing nerdctl and running:
```
sudo nerdctl run --privileged --rm tonistiigi/binfmt --install all
```

medyagh (approved):

PR #15452, opened 2022-12-02
feat: direct image build on containerd via docker-env

ComradeProgrammer:
fixes #15389 
This pr enable the use of `minikube docker-env` command, so that the minikube can be used as docker, even when the runtime is containerd

**Before:**
```
$ minikube start  --container-runtime="containerd"
$ minikube ssh-host --append-known
$ minikube docker-env --ssh-host --ssh-add
❌  Exiting due to MK_USAGE: The docker-env command is only compatible with the "docker" runtime, but this cluster was configured to use the "containerd" runtime.
```

**After:**

$ minikube docker-env --ssh-host --ssh-add
stdout: 
```
export DOCKER_HOST="ssh://docker@127.0.0.1:49157"
export MINIKUBE_ACTIVE_DOCKERD="minikube"
export SSH_AUTH_SOCK="/tmp/ssh-XXXXXXqYJ2oD/agent.4680"
export SSH_AGENT_PID="4681"
```
stderr: 
```
❗  Using the docker-env command with the containerd runtime is a highly experimental feature, please provide feedback or contribute to make it better
❗  Please ensure you have executed 'ssh-agent bash' and 'minikube ssh-host --append-known' in this shell before using docker-env on containerd. Ignore this message if you have done it

# To point your shell to minikube's docker-daemon, run:
# eval $(minikube -p minikube docker-env --ssh-host)
Identity added: /home/tjm/.minikube/machines/minikube/id_rsa (/home/tjm/.minikube/machines/minikube/id_rsa)
```
$ DOCKER_HOST="ssh://docker@127.0.0.1:49157" docker version
```

Client: Docker Engine - Community
 Version:           20.10.21
 API version:       1.40
 Go version:        go1.18.7
 Git commit:        baeda1f
 Built:             Tue Oct 25 18:01:58 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server:
 nerdctl:
  Version:          1.0.0
 buildctl:
  Version:          0.10.3
  GitCommit:        c8d25d9a103b70dc300a4fd55e7e576472284e31
 containerd:
  Version:          1.6.10
  GitCommit:        770bd0108c32f3fb5c73ae1264f7e503fe7b2661
 runc:
  Version:          1.1.4
 Engine:
  Version:          1.0.0
  API version:      1.40 (minimum version 1.24)
  Go version:       go1.19.2
  Git commit:       c00780a1f5b905b09812722459c54936c9e070e6
  Built:            
[...]

PR #16761, opened 2023-06-23
Add ability to start & stop ssh-agent process

spowelljr:
This PR adds an `sshagent` package that allows starting and stoping an ssh-agent process. Once a process is started it's auth socket and agent pid are added to the cluster config. When running `docker-env`, if the auth socket and agent pid are populated in the cluster config they'll be included in the command output. On cluster delete the process is killed.

When creating this output I added `sshagent.Start()` to the beginning of `cmd/minikube/cmd/docker-env.go`

`sshagent.Start()` will be implemented in https://github.com/kubernetes/minikube/pull/15452

```
$ pgrep ssh-agent
1219

$ minikube docker-env
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://127.0.0.1:59146"
export DOCKER_CERT_PATH="/Users/powellsteven/.minikube/certs"
export SSH_AUTH_SOCK="/var/folders/9l/6wpxv6wd1b901m1146r579wc00rqw3/T//ssh-KCQt1sNqrCPI/agent.29227"
export SSH_AGENT_PID="29228"

$ pgrep ssh-agent
1219
29228

$ minikube delete
🔥  Deleting container "minikube" ...
🔥  Removing /Users/powellsteven/.minikube/machines/minikube ...

$ pgrep ssh-agent
1219
```

**Before:**
```
$ minikube docker-env
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://127.0.0.1:51030"
export DOCKER_CERT_PATH="/Users/powellsteven/.minikube/certs"

```

**After w/ agent stopped:**
```
$ minikube docker-env
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://127.0.0.1:51030"
export DOCKER_CERT_PATH="/Users/powellsteven/.minikube/certs"

```

**After w/ agent started:**
```
$ minikube docker-env
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://127.0.0.1:51030"
export DOCKER_CERT_PATH="/Users/powellsteven/.minikube/certs"
export SSH_AUTH_SOCK="/var/folders/9l/6wpxv6wd1b901m1146r579wc00rqw3/T//ssh-rrVWjcEBi4pO/agent.37343"
export SSH_AGENT_PID="37344"

```

### DIFF
--- a/cmd/minikube/cmd/docker-env.go
+++ b/cmd/minikube/cmd/docker-env.go
@@ -71,6 +71,12 @@ var dockerEnvTCPTmpl = fmt.Sprintf(
 		"{{ if .NoProxyVar }}"+
 		"{{ .Prefix }}{{ .NoProxyVar }}{{ .Delimiter }}{{ .NoProxyValue }}{{ .Suffix }}"+
 		"{{ end }}"+
+		"{{ if .SSHAuthSock }}"+
+		"{{ .Prefix }}%s{{ .Delimiter }}{{ .SSHAuthSock }}{{ .Suffix }}"+
+		"{{ end }}"+
+		"{{ if .SSHAgentPID }}"+
+		"{{ .Prefix }}%s{{ .Delimiter }}{{ .SSHAgentPID }}{{ .Suffix }}"+
+		"{{ end }}"+
 		"{{ .UsageHint }}",
 	constants.DockerTLSVerifyEnv,
 	constants.DockerHostEnv,
@@ -78,7 +84,9 @@ var dockerEnvTCPTmpl = fmt.Sprintf(
 	constants.ExistingDockerTLSVerifyEnv,
 	constants.ExistingDockerHostEnv,
 	constants.ExistingDockerCertPathEnv,
-	constants.MinikubeActiveDockerdEnv)
+	constants.MinikubeActiveDockerdEnv,
+	constants.SSHAuthSock,
+	constants.SSHAgentPID)
 var dockerEnvSSHTmpl = fmt.Sprintf(
 	"{{ .Prefix }}%s{{ .Delimiter }}{{ .DockerHost }}{{ .Suffix }}"+
 		"{{ .Prefix }}%s{{ .Delimiter }}{{ .MinikubeDockerdProfile }}{{ .Suffix }}"+
@@ -99,6 +107,9 @@ type DockerShellConfig struct {
 	ExistingDockerCertPath  string
 	ExistingDockerHost      string
 	ExistingDockerTLSVerify string
+
+	SSHAuthSock string
+	SSHAgentPID string
 }
 
 var (
@@ -141,6 +152,9 @@ func dockerShellCfgSet(ec DockerEnvConfig, envMap map[string]string) *DockerShel
 	s.ExistingDockerTLSVerify = envMap[constants.ExistingDockerTLSVerifyEnv]
 
 	s.MinikubeDockerdProfile = envMap[constants.MinikubeActiveDockerdEnv]
+
+	s.SSHAuthSock = envMap[constants.SSHAuthSock]
+	s.SSHAgentPID = envMap[constants.SSHAgentPID]
 
 	if ec.noProxy {
 		noProxyVar, noProxyValue := defaultNoProxyGetter.GetNoProxyVar()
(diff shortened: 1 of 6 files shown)


medyagh (commented):
two questions
1-
does this PR make an ssh-agent by default even if it is not used by docker-env ?

2-
can we make minikube use this SSH agent ? I bet this would make our command runner faster
 after we create a KIC container we switch from docker exec to SSH that made minikube faster.

here are some relevent code:

- we copy the SSH key to remote for SSH runner
https://github.com/medyagh/minikube/blob/ecd457fc2eceba5ea3c37d7045d77f652a652d49/pkg/provision/provision.go#L171

- here is some remote options for SSH runner
- https://github.com/medyagh/minikube/blob/ecd457fc2eceba5ea3c37d7045d77f652a652d49/pkg/provision/provision.go#L211

- here is where ssh runner is created https://github.com/medyagh/minikube/blob/ecd457fc2eceba5ea3c37d7045d77f652a652d49/pkg/minikube/machine/client.go#L163

medyagh:
> 2. I don't see why not, would need additional modifications though

would you be interested to plug it in to make our SSH_RUNNER to use this agent and benchmark the Command Runner metrics ( we could see how many miliseconds or seconds it takes to run pwd or ls commands using with or without agent..

because without agent I assume it would have to load the SSH key on each connection to memory

spowelljr:
> > 2. I don't see why not, would need additional modifications though
> 
> would you be interested to plug it in to make our SSH_RUNNER to use this agent and benchmark the Command Runner metrics ( we could see how many miliseconds or seconds it takes to run pwd or ls commands using with or without agent..
> 
> because without agent I assume it would have to load the SSH key on each connection to memory

Could do that, would do it in a follow up PR though

cmd/minikube/cmd/docker-env.go
```diff
@@ -71,14 +71,22 @@ var dockerEnvTCPTmpl = fmt.Sprintf(
...
 	constants.DockerHostEnv,
 	constants.DockerCertPathEnv,
 	constants.ExistingDockerTLSVerifyEnv,
 	constants.ExistingDockerHostEnv,
 	constants.ExistingDockerCertPathEnv,
-	constants.MinikubeActiveDockerdEnv)
+	constants.MinikubeActiveDockerdEnv,
+	constants.SSHAuthSock,
+	constants.SSHAgentPID)
 var dockerEnvSSHTmpl = fmt.Sprintf(
```
ComradeProgrammer, line 90: why dockerEnvSSHTmpl doesn't need to be modified? I think perhaps we need ssh-agent exactly when we are trying to connect to minikube's docker daemon with SSH.

spowelljr: Good call, added to the SSH template as well

medyagh:
@spowelljr could I plz see the output of docker-env before/after this PR (both incase the agent is started and incase the agent is not started)

spowelljr:
> @spowelljr could I plz see the output of docker-env before/after this PR (both incase the agent is started and incase the agent is not started)

Added before/after the PR description

medyagh (approved):

medyagh:
looks good to me @spowelljr my only question is, for docker-env on docker runtime, does the AFTER need to export the regular docker-env envs too ?

$ minikube docker-env
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://127.0.0.1:51030"
export DOCKER_CERT_PATH="/Users/powellsteven/.minikube/certs"
export SSH_AUTH_SOCK="/var/folders/9l/6wpxv6wd1b901m1146r579wc00rqw3/T//ssh-rrVWjcEBi4pO/agent.37343"
export SSH_AGENT_PID="37344"

or is there a way to make regular docker-env to use that SSH_AGENT (kind of adding DOCKER_CERT_PATH="/Users/powellsteven/.minikube/certs" to the ssh agent ? ) maybe not

spowelljr:
> looks good to me @spowelljr my only question is, for docker-env on docker runtime, does the AFTER need to export the regular docker-env envs too ?
> 
> $ minikube docker-env export DOCKER_TLS_VERIFY="1" export DOCKER_HOST="tcp://127.0.0.1:51030" export DOCKER_CERT_PATH="/Users/powellsteven/.minikube/certs" export MINIKUBE_ACTIVE_DOCKERD="minikube" export SSH_AUTH_SOCK="/var/folders/9l/6wpxv6wd1b901m1146r579wc00rqw3/T//ssh-rrVWjcEBi4pO/agent.37343" export SSH_AGENT_PID="37344"
> 
> or is there a way to make regular docker-env to use that SSH_AGENT (kind of adding DOCKER_CERT_PATH="/Users/powellsteven/.minikube/certs" to the ssh agent ? ) maybe not

Responded offline:
`The ssh-agent is only going to get exported if it's started, and in my current PR it's not started at all. I'm leaving that up to #15452 to start when the user calls docker-env and they're using containerd, so if they're using docker CRI then the agent won't be started and it won't be included in the export`

Issue #16776, opened 2023-06-26
unblock firewall for macos one more time if it it is unblocked and macos still doesnt unblock


medyagh:
### What Happened?

unblock firewall for macos one more time if it it is unblocked and macos still doesnt unblock`

### Attach the log file

n/a

### Operating System

None

### Driver

None

akosyakov:
> unblock firewall for macos one more time if it it is unblocked and macos still doesnt unblock`

@medyagh Hello 👋 sorry for asking out of the blue, but would it be possible to share how did you manage to reproduce such conditions?

spowelljr:
@akosyakov You can look at the PR that implemented it https://github.com/kubernetes/minikube/pull/16789

PR #16789, opened 2023-06-28
QEMU: Unblock bootpd if start fails to due to blocking

spowelljr:
Closes https://github.com/kubernetes/minikube/issues/16776

Even if you have bootpd unblocked in the builtin macOS firewall, you can run into the situation where it's still being blocked. To resolve it you have to run the unblock command again.

This PR checks the output of a start failure on QEMU, and if it's due to bootpd being blocked runs the unblock command and retries the start again.

**Before:**
```
$ minikube start --driver qemu
😄  minikube v1.30.1 on Darwin 13.4.1 (arm64)
🌐  Automatically selected the socket_vmnet network

❌  Exiting due to IF_BOOTPD_FIREWALL: ip not found: failed to get IP address: could not find an IP address for fe:70:96:84:47:94
💡  Suggestion: 

    Your firewall is likely blocking bootpd, to unblock it run:
    sudo /usr/libexec/ApplicationFirewall/socketfilterfw --add /usr/libexec/bootpd
    sudo /usr/libexec/ApplicationFirewall/socketfilterfw --unblock /usr/libexec/bootpd

```

**After:**
```
minikube start --driver qemu
😄  minikube v1.30.1 on Darwin 13.4.1 (arm64)
🌐  Automatically selected the socket_vmnet network
🔑  Your firewall is blocking bootpd which is required for socket_vmnet. The following commands will be executed to unblock bootpd:

    $ sudo /usr/libexec/ApplicationFirewall/socketfilterfw --add /usr/libexec/bootpd 
    $ sudo /usr/libexec/ApplicationFirewall/socketfilterfw --unblock /usr/libexec/bootpd 


Password: 
🔄  Sucessfully unblocked bootpd process from firewall, retrying
🔥  Deleting "minikube" in qemu2 ...
🤦  StartHost failed, but will try again: creating host: create: creating: ip not found: failed to get IP address: could not find an IP address for 52:cd:17:52:c9:4d
🐳  Preparing Kubernetes v1.27.3 on Docker 24.0.2 ...
    ▪ Configuring RBAC rules ...
```

### DIFF
--- a/pkg/drivers/qemu/qemu.go
+++ b/pkg/drivers/qemu/qemu.go
@@ -44,7 +44,9 @@ import (
 	pkgdrivers "k8s.io/minikube/pkg/drivers"
 	"k8s.io/minikube/pkg/minikube/exit"
 	"k8s.io/minikube/pkg/minikube/firewall"
+	"k8s.io/minikube/pkg/minikube/out"
 	"k8s.io/minikube/pkg/minikube/reason"
+	"k8s.io/minikube/pkg/minikube/style"
 	"k8s.io/minikube/pkg/network"
 )
 
@@ -528,7 +530,8 @@ func (d *Driver) Start() error {
 			klog.Errorf("failed unblocking bootpd from firewall: %v", unblockErr)
 			exit.Error(reason.IfBootpdFirewall, "ip not found", err)
 		}
-		return fmt.Errorf("bootpd process is unblocked, will retry")
+		out.Styled(style.Restarting, "Sucessfully unblocked bootpd process from firewall, retrying")
+		return fmt.Errorf("ip not found: %v", err)
 	}
 
 	log.Infof("Waiting for VM to start (ssh -p %d docker@%s)...", d.SSHPort, d.IPAddress)
--- a/pkg/drivers/qemu/qemu.go
+++ b/pkg/drivers/qemu/qemu.go
@@ -40,8 +40,10 @@ import (
 	"github.com/docker/machine/libmachine/state"
 	"github.com/pkg/errors"
 
+	"k8s.io/klog/v2"
 	pkgdrivers "k8s.io/minikube/pkg/drivers"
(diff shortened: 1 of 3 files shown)


pkg/drivers/qemu/qemu.go
```diff
@@ -515,13 +519,19 @@ func (d *Driver) Start() error {
...
+			log.Debugf("IP: %s", d.IPAddress)
+			break
+		}
+		if !isBootpdError(err) {
 			return errors.Wrap(err, "IP address never found in dhcp leases file")
 		}
-		log.Debugf("IP: %s", d.IPAddress)
+		if unblockErr := firewall.UnblockBootpd(); unblockErr != nil {
+			klog.Errorf("failed unblocking bootpd from firewall: %v", unblockErr)
+			exit.Error(reason.IfBootpdFirewall, "ip not found", err)
```
medyagh, line 531: I am curious if we could return an error and let the CMD package to exit ? I generally frown upon exit in non-cmd code, but if there is a good reason or makes way cleaner code I am okay with Exit here.. 

what u thnk?

spowelljr: The reason I did `exit.Error` is because it will retry if the first start fails. In the case where we know bootpd is being blocked and we failed to unblock the process it's almost certainly going to fail the retry so I was exiting. I could remove it,, the downside would be that it would have to go through the start loop one more time.

medyagh, line 531: add comment, and say if return errr, it will retry which makes no sense.

medyagh (changes requested):

pkg/drivers/qemu/qemu.go
medyagh: might be able to use re-triable errors pattern.

medyagh (approved):

PR #16793, opened 2023-06-29
ISO: Add docker-buildx

spowelljr:
If you run `docker build` in the ISO you get the following:
```
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/
```

So this installs the docker-buildx components to prevent deprecation


spowelljr:
I created this PR during the GitHub outage and it's not picking up changes to the branch so going to close it and recreate it.

PR #16233, opened 2023-04-04
Mark VMware driver as unsupported 

spowelljr:
Related https://github.com/kubernetes/minikube/issues/16221

- Remove VMware from driver auto select
- Output exit message if user tries to use VMware driver
- Small cleanup of using magic strings for drivers

**Before:**
```
$ minikube start --driver vmware
😄  minikube v1.30.0 on Darwin 13.2.1
✨  Using the vmware driver based on user configuration
👍  Starting control plane node minikube in cluster minikube
💾  Downloading Kubernetes v1.26.3 preload ...
    > preloaded-images-k8s-v18-v1...:  397.02 MiB / 397.02 MiB  100.00% 53.18 M
🔥  Creating vmware VM (CPUs=2, Memory=6000MB, Disk=20000MB) ...

# hangs...
```

**After:**
```
$ minikube start --driver vmware
😄  minikube v1.30.0 on Darwin 13.2.1
✨  Using the vmware driver based on user configuration

❌  Exiting due to DRV_UNSUPPORTED: Due to security improvements to minikube, the VMware driver is currently not supported. Available workarounds are to use a different driver or downgrade minikube to v1.29.0.

    We are accepting community contributions to fix this, for more details on the issue see: https://github.com/kubernetes/minikube/issues/16221

```

PR #16796, opened 2023-06-30
feat: add VMware driver support for new minikube ISO

### LINKED ISSUE #16221: Recently built minikube fails to start with vmware driver
### What Happened?

Build both minikube and the amd64 ISO with recent changes and it fails to start with the vmware driver.

```
$ ./minikube-darwin-amd64 start --driver vmware --iso-url=file://$(pwd)/minikube-amd64.iso
😄  minikube v1.30.0 on Darwin 13.2.1
✨  Using the vmware driver based on user configuration
💾  Downloading Kubernetes v1.26.3 preload ...
    > preloaded-images-k8s-v18-v1...:  397.02 MiB / 397.02 MiB  100.00% 53.18 M
🔥  Creating vmware VM (CPUs=2, Memory=6000MB, Disk=20000MB) .../ ^C
```
From the logs this appears to be due to the docker user account being removed. Looking at commit history, it was removed via [Disable plain text login for ISO](https://github.com/kubernetes/minikube/pull/15849) by @medyagh.

### Attach the log file

```
I0403 10:43:40.090076   20589 main.go:141] libmachine: (minikube) DBG | IP found in DHCP lease table: 192.168.111.164
I0403 10:43:40.090091   20589 main.go:141] libmachine: (minikube) DBG | Got an ip: 192.168.111.164
I0403 10:43:40.090737   20589 main.go:141] libmachine: (minikube) DBG | Creating Tar key bundle...
I0403 10:43:40.091232   20589 main.go:141] libmachine: (minikube) DBG | executing: /Users/mark/bin/vmrun -gu docker -gp tcuser directoryExistsInGuest /Users/mark/.minikube/machines/minikube/minikube.vmx /var/lib/boot2docker
- I0403 10:43:40.426605   20589 main.go:141] libmachine: (minikube) DBG | Error: Invalid user name or password for the guest OS
I0403 10:43:40.429937   20589 main.go:141] libmachine: (minikube) DBG | executing: /Users/mark/bin/vmrun -gu docker -gp tcuser CopyFileFromHostToGuest /Users/mark/.minikube/machines/minikube/minikube.vmx /Users/mark/.minikube/machines/minikube/userdata.tar /home/docker/userdata.tar
/ I0403 10:43:40.690751   20589 main.go:141] libmachine: (minikube) DBG | Error: Invalid user name or password for the guest OS
I0403 10:43:40.694008   20589 main.go:141] libmachine: (minikube) DBG | executing: /Users/mark/bin/vmrun -gu docker -gp tcuser runScriptInGuest /Users/mark/.minikube/machines/minikube/minikube.vmx /bin/sh sudo sh -c "tar xvf /home/docker/userdata.tar -C /home/docker > /var/log/userdata.log 2>&1 && chown -R docker:staff /home/docker"
\ I0403 10:43:40.954608   20589 main.go:141] libmachine: (minikube) DBG | Error: Invalid user name or password for the guest OS
I0403 10:43:40.957541   20589 main.go:141] libmachine: (minikube) DBG | executing: /Users/mark/bin/vmrun -gu docker -gp tcuser runScriptInGuest /Users/mark/.minikube/machines/minikube/minikube.vmx /bin/sh sudo /bin/mv /home/docker/userdata.tar /var/lib/boot2docker/userdata.tar
- I0403 10:43:41.219867   20589 main.go:141] libmachine: (minikube) DBG | Error: Invalid user name or password for the guest OS
I0403 10:43:41.223214   20589 main.go:141] libmachine: (minikube) DBG | executing: /Users/mark/bin/vmrun -gu docker -gp tcuser enableSharedFolders /Users/mark/.minikube/machines/minikube/minikube.vmx
/ I0403 10:43:41.497243   20589 main.go:141] libmachine: (minikube) DBG | executing: /Users/mark/bin/vmrun -gu docker -gp tcuser addSharedFolder /Users/mark/.minikube/machines/minikube/minikube.vmx Users /Users
| I0403 10:43:41.777151   20589 main.go:141] libmachine: (minikube) DBG | executing: /Users/mark/bin/vmrun -gu docker -gp tcuser runScriptInGuest /Users/mark/.minikube/machines/minikube/minikube.vmx /bin/sh [ ! -d /hosthome ]&& sudo mkdir /hosthome; sudo mount --bind /mnt/hgfs//hosthome /hosthome || [ -f /usr/local/bin/vmhgfs-fuse ]&& sudo /usr/local/bin/vmhgfs-fuse -o allow_other .host:/Users /hosthome || sudo mount -t vmhgfs -o uid=$(id -u),gid=$(id -g) .host:/Users /hosthome
- I0403 10:43:42.041025   20589 main.go:141] libmachine: (minikube) DBG | Error: Invalid user name or password for the guest OS
- ```

### Operating System

macOS (Default)

### Driver

VMware

### PR DESCRIPTION
**Pull request by lbogdan**
Resolves #16221

This PR adds VMware driver support for the new minikube ISO, which removed the password for the `docker` user in #15849.

This ISO change broke the VMware driver, because it uses `ssh` and `vmrun` commands , which assume that the [`docker` user password is `tcuser`](https://github.com/machine-drivers/docker-machine-driver-vmware/blob/b30963f81e6e25743b42d469a22a5a9f5db742fb/pkg/drivers/vmware/config/config.go#L29C1-L30), in order to initialize the VM.

It uses [the same approach as the QEMU driver](https://github.com/kubernetes/minikube/blob/08896fd1dc362c097c925146c4a0d0dac715ace0/pkg/drivers/qemu/qemu.go#L626-L685): it creates a `tar` stream having the `"boot2docker, please format-me"` magic string and the public key in `.ssh/authorized_keys` and writes it to the beginning of the VM disk, which will then be [detected and used by the automount script](https://github.com/kubernetes/minikube/blob/08896fd1dc362c097c925146c4a0d0dac715ace0/deploy/iso/minikube-iso/package/automount/minikube-automount#L32) during the first boot.

To be able to make changes to `docker-machine-driver-vmware`, this vendors its `master` branch into `pkg/drivers/vmware`, moving `config/config.go` to the parent folder and to the `vmware` package. To check the actual changes, see the [second commit](https://github.com/kubernetes/minikube/pull/16796/commits/48d28d96c8e368119b342a12e8814a3d73ced829).

### CODE REVIEW

**Comment by medyagh** [comment]
@lbogdan  thank you very much for this contribution, do you mind sharing output of "minikube start" Before / After this PR ?

**Comment by lbogdan** [comment]
Sure!

Without this PR, latest `minikube`; VMware driver was disabled in #16233:

```
PS C:\Users\Bogdan> minikube version
minikube version: v1.30.1
commit: 08896fd1dc362c097c925146c4a0d0dac715ace0
PS C:\Users\Bogdan> minikube start --driver vmware
😄  minikube v1.30.1 on Microsoft Windows 11 Pro 10.0.22621.1848 Build 22621.1848
✨  Using the vmware driver based on user configuration

❌  Exiting due to DRV_UNSUPPORTED: Due to security improvements to minikube the VMware driver is currently not supported. Available workarounds are to use a different driver or downgrade minikube to v1.29.0.

    We are accepting community contributions to fix this, for more details on the issue see: https://github.com/kubernetes/minikube/issues/16221
```

With this PR on top of `master`:

```
PS C:\Users\Bogdan> minikube start --driver vmware
😄  minikube v1.30.1 on Microsoft Windows 11 Pro 10.0.22621.1848 Build 22621.1848
✨  Using the vmware driver based on user configuration
🔥  Creating vmware VM (CPUs=2, Memory=6000MB, Disk=20000MB) ...
🐳  Preparing Kubernetes v1.27.3 on Docker 24.0.2 ...
    ▪ Configuring RBAC rules ...
PS C:\Users\Bogdan> kubectl get nodes -o wide
NAME       STATUS   ROLES           AGE   VERSION   INTERNAL-IP       EXTERNAL-IP   OS-IMAGE               KERNEL-VERSION   CONTAINER-RUNTIME
minikube   Ready    control-plane   36s   v1.27.3   192.168.126.128   <none>        Buildroot 2021.02.12   5.10.57          docker://24.0.2
```

**Comment by lbogdan** [comment]
There are quite some lint errors caused by the vendoring of `docker-machine-driver-vmware`, I'm not that proficient in Go, but I'll try to take a look tomorrow.

**Comment by afbjorklund** [comment]
> There are quite some lint errors caused by the vendoring of `docker-machine-driver-vmware`, I'm not that proficient in Go, but I'll try to take a look tomorrow.
> 
> But before I do, please confirm vendoring is the way to go here.

We are working towards lifting all the drivers out, so it is probably **not** the way to go (long term).

It is of course a perfectly reasonable way to _test_ it, so that one doesn't have to only use "docker-machine"

* https://github.com/kubernetes/minikube/pull/16688

But we made a related change, to stop using the external binary in anticipation of "minikube-machine"

* https://github.com/kubernetes/minikube/pull/16687

I would have expected that this password removal would happen on docker-machine-driver-vmware ?
Have not compared the new driver with the old driver, but thought it would be more of a patch / PR...

What we do for docker/machine and for the other drivers, is that we replace them with a patched fork.
Maybe backporting the fixes to machine-drivers/docker-machine-driver-vmware would be one way.

----

Eventually, we want to start using the new organization:

* https://github.com/kubernetes/minikube/pull/16672

Not sure where it would live, separately or in the main repo ?

* https://github.com/minikube-machine/machine/issues/1

The reason for forking libmachine and all the drivers, is both that it is dead and that we want to add new API.
Currently the drivers and provisioners are forked in the minikube mono-repo, and it is not easy to work with.

Related to this, we probably want to lift "minikube-iso" out again since it has become a very big component.
It is also in need of some refactoring, especially after the arch work, and wants to become more standalone.

@lbogdan : finally, thanks for fixing the vmware driver!

I think we want something similar for the parallels driver.

**Comment by lbogdan** [comment]
> Maybe backporting the fixes to `machine-drivers/docker-machine-driver-vmware` would be one way.

This would probably make the most sense, but considering there was no activity there for over an year, and the "fix" is unrelated to the actual `docker-machine` driver, it will probably take some time to get merged, if ever.

Another approach would be to fork it, but that would mean managing a new repository.

I would suggest to go with vendoring (is this the same as what you call "forked in the minikube mono-repo"?) short term, and then include it in the process of lifting out the drivers to `minikube-machine`.

Unrelated, I see most of the Jenkins integration tests fail (flake?), is this a known issue, and is anyone looking into it?

**Comment by afbjorklund** [comment]
> I would suggest to go with vendoring it (is this the same as what you call "forked in the minikube mono-repo"?) short term, and then include it in the process of lifting out the drivers to `minikube-machine`.

If by vendoring you mean copying it to the minikube repository under pkg/drivers, then it is not the same thing. If you include it in the `go.mod` it will end up in `vendor` after running "go mod vendor", that is more vendoring these days...

The main difference is that in docker-machine, the drivers were standalone grpc programs and did not share code. The "Config" structure was never exported, and all the settings were transferred using the docker-machine string Flags.

Now in minikube, the driver is referenced in the code base (from the "registry") and all the code is vendored in.

So the main difference was to build the driver in as well, except for those where it makes sense to link it separately:

* hyperkit (due to linking with some Mac libraries)

* kvm (due to requiring linking with the C libraries)

**Comment by lbogdan** [comment]
I've now also fixed the lint errors.

In the process, I added error checks to the `vmrun` commands used to initialize the VM, which were failing because of the removed password, so bubbling up the error started failing `minikube start`.

So I went ahead and removed all `ssh` and `vmrun` commands, and all related code, that assumed password access.

**Comment by afbjorklund** [comment]
The main outstanding question is if all drivers should be included with minikube-machine/machine, or if some should be in their own repositories still. Either way, they will all be included by minikube (in some way) - just like they were before.

i.e. do we create a directory:

minikube-machine/machine/pkg/drivers/vmware

or do we create a repository:

minikube-machine/minikube-machine-driver-vmware

**Comment by lbogdan** [comment]
Oh, OK, then I guess I had a more inclusive understanding of vendoring, meaning "copying a dependency's source files into the project". 🙂

**Comment by afbjorklund** [comment]
> Oh, OK, then I guess I had a more inclusive understanding of vendoring, meaning "copying a dependency's source files into the project". 🙂

I'm not sure the terms are properly defined anywhere, just as long as we understand what we are talking about. The roadmap items would be to move "minikube-machine" (including libmachine, drivers, and minikube-iso) out of the monorepo. With the increased focus on the KIC drivers, it could make more sense to have the VM and ISO separate.

There is some bigger picture discussion that should happen on the roadmap planning - or in the Slack channel perhaps

**Comment by lbogdan** [comment]
Well, I'm only interested in fixing VMware support as quick as possible, so let me know if and how can I help in getting this in the next release.

**Comment by afbjorklund** [comment]
> Well, I'm only interested in fixing VMware support as quick as possible, so let me know if and how I can help in getting this in the next release.

Patching github.com/machine-drivers/docker-machine-driver-vmware v0.1.5 would be a faster route.

**Review by afbjorklund — CHANGES_REQUESTED** [review]
Would prefer patching the docker-machine-driver-vmware repository

**Comment by lbogdan** [comment]
Sounds good!

Should I fork it under my GitHub `lbogdan` username, or do you want to fork it under the (`kubernetes`? `minikube-machine`?) organization and give me access?

**Comment by lbogdan** [comment]
Forked it under [`lbogdan/docker-machine-driver-vmware`](https://github.com/lbogdan/docker-machine-driver-vmware) for now, and tagged the changes with `v0.2.0`.

Also manually tested it on macOS with VMware Fusion in the meanwhile.

**Comment by lbogdan** [comment]
While testing on macOS, I've noticed a small issue: the `minikube` part of the VMware driver checks for `vmrun` to be in the path https://github.com/kubernetes/minikube/blob/f58582b7eb7b72a261c68a41d00be86bf69f952e/pkg/minikube/registry/drvs/vmware/vmware.go#L59-L65 but the driver has [special logic on macOS](https://github.com/lbogdan/docker-machine-driver-vmware/blob/a391c48b14d523058072d491f0fec2296dd461e4/pkg/drivers/vmware/vmware_darwin.go#L39-L53) (and [on Windows](https://github.com/lbogdan/docker-machine-driver-vmware/blob/a391c48b14d523058072d491f0fec2296dd461e4/pkg/drivers/vmware/vmware_windows.go#L40-L70), too) to look for it, even if it's not in path.

I was thinking that we can export `SetVmwareCmd()` and use it to check for `vmrun` in `status()`. Alternatively, we can export a new `CheckVmrun()` function. What do you think?

**Comment by afbjorklund** [comment]
You actually don't have to tag your fork, Go will "make something up" for a version if you just give it the commit

Normally we don't tag forks, since if upstream wakes up of their coma we would suddenly end up with two `v0.2.0`

**Comment by afbjorklund** [comment]
Fixing the duplicate check in the "registry" sounds like the most obvious, if it is currently broken (can't find `vmrun`)

It is not possible to export new functions in libmachine, which is one of the reasons that we want minikube-machine

**Comment by lbogdan** [comment]
Can't we just export (capitalize the first letter of) `SetVmwareCmd()` in the `vmware` package in `docker-machine-driver-vmware`, and then in `minikube/pkg/minikube/registry/drvs/vmware/vmware.go` just do

```go
	cmd := vmware.SetVmwareCmd("vmrun") 
	if cmd == "vmrun" { // if it can't find it, it just returns the argument
		return registry.State{Error: err, Fix: "Install VMware", Doc: "https://minikube.sigs.k8s.io/docs/reference/drivers/vmware/"} 
	} 
	return registry.State{Installed: true, Healthy: true} 
}
```
?

lbogdan:
> Fixing the duplicate check in the "registry" sounds like the most obvious, if it is currently broken (can't find vmrun)

If you meant removing the `vmrun` registry check entirely, the behavior will be a bit different:

With updated check:

```
PS C:\Users\Bogdan> .\Downloads\minikube.exe start --driver vmware
😄  minikube v1.30.1-761-g36238f402 on Microsoft Windows 11 Pro 10.0.22621.1928 Build 22621.1928
✨  Using the vmware driver based on user configuration

🤷  Exiting due to PROVIDER_VMWARE_NOT_FOUND: The 'vmware' provider was not found: vmrun not found
💡  Suggestion: Install VMware
📘  Documentation: https://minikube.sigs.k8s.io/docs/reference/drivers/vmware/
```

Without check:

```
PS C:\Users\Bogdan> .\Downloads\minikube.exe start --driver vmware
😄  minikube v1.30.1-761-g36238f402 on Microsoft Windows 11 Pro 10.0.22621.1928 Build 22621.1928
✨  Using the vmware driver based on user configuration
🔥  Creating vmware VM (CPUs=2, Memory=6000MB, Disk=20000MB) ...
🔥  Deleting "minikube" in vmware ...
🤦  StartHost failed, but will try again: creating host: create: creating: open C:\Users\Bogdan\.minikube\machines\minikube\minikube-tmp-flat.vmdk: The system cannot find the file specified.
🔄  Restarting existing vmware VM for "minikube" ...
😿  Failed to start vmware VM. Running "minikube delete" may fix it: driver start: exec: no command

❌  Exiting due to GUEST_PROVISION: error provisioning guest: Failed to start host: driver start: exec: no command

```

afbjorklund:
Actual I was thinking more like fixing the check, by adding the missing paths on the missing platforms.

missed in commit 3c489854776bc9e10625fbd7c987c7c7af5b3ab1

lbogdan:
So just go ahead and duplicate all the logic from `docker-machine-driver-vmware`, for Windows and macOS?

Windows - reads from the Windows registry: https://github.com/lbogdan/docker-machine-driver-vmware/blob/a391c48b14d523058072d491f0fec2296dd461e4/pkg/drivers/vmware/vmware_windows.go#L40-L70

macOS - looks into an additional path: https://github.com/lbogdan/docker-machine-driver-vmware/blob/a391c48b14d523058072d491f0fec2296dd461e4/pkg/drivers/vmware/vmware_darwin.go#L39-L53

afbjorklund:
Yeah, not great either... Having some feature for it in the new API would be good, I think?

Could use the same thing for VBoxManage or qemu-system-x86_64/qemu-system-aarch64 etc

afbjorklund:
It is just that at some path in the future we want to restore the feature of decoupling the "registry" and the drivers...

So we should try to not import any code from the drivers, outside of the libmachine API... Including the `Config`.

medyagh:
> Sure!
> 
> Without this PR, latest `minikube`; VMware driver was disabled in #16233:
> 
> ```
> PS C:\Users\Bogdan> minikube version
> minikube version: v1.30.1
> commit: 08896fd1dc362c097c925146c4a0d0dac715ace0
> PS C:\Users\Bogdan> minikube start --driver vmware
> 😄  minikube v1.30.1 on Microsoft Windows 11 Pro 10.0.22621.1848 Build 22621.1848
> ✨  Using the vmware driver based on user configuration
> 
> ❌  Exiting due to DRV_UNSUPPORTED: Due to security improvements to minikube the VMware driver is currently not supported. Available workarounds are to use a different driver or downgrade minikube to v1.29.0.
> 
>     We are accepting community contributions to fix this, for more details on the issue see: https://github.com/kubernetes/minikube/issues/16221
> ```
> 
> With this PR on top of `master`:
> 
> ```
> PS C:\Users\Bogdan> minikube start --driver vmware
> 😄  minikube v1.30.1 on Microsoft Windows 11 Pro 10.0.22621.1848 Build 22621.1848
> ✨  Using the vmware driver based on user configuration
> 👍  Starting control plane node minikube in cluster minikube
> 🔥  Creating vmware VM (CPUs=2, Memory=6000MB, Disk=20000MB) ...
> 🐳  Preparing Kubernetes v1.27.3 on Docker 24.0.2 ...
>     ▪ Generating certificates and keys ...
>     ▪ Booting up control plane ...
>     ▪ Configuring RBAC rules ...
> 🔗  Configuring bridge CNI (Container Networking Interface) ...
>     ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
> 🔎  Verifying Kubernetes components...
> 🌟  Enabled addons: storage-provisioner, default-storageclass
> 🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default
> PS C:\Users\Bogdan> kubectl get nodes -o wide
> NAME       STATUS   ROLES           AGE   VERSION   INTERNAL-IP       EXTERNAL-IP   OS-IMAGE               KERNEL-VERSION   CONTAINER-RUNTIME
> minikube   Ready    control-plane   36s   v1.27.3   192.168.126.128   <none>        Buildroot 2021.02.12   5.10.57          docker://24.0.2
> ```

thank you @lbogdan, I see @afbjorklund already reviewing the PR, once @afbjorklund approves the PR, you got mine too :)

go.mod
```diff
@@ -235,7 +235,7 @@ replace (
 	github.com/Parallels/docker-machine-parallels/v2 => github.com/minikube-machine/machine-driver-parallels/v2 v2.0.1
 	github.com/briandowns/spinner => github.com/alonyb/spinner v1.12.7
 	github.com/docker/machine => github.com/minikube-machine/machine v0.0.0-20230610170757-350a83297593
-	github.com/machine-drivers/docker-machine-driver-vmware => github.com/minikube-machine/machine-driver-vmware v0.1.5
+	github.com/machine-drivers/docker-machine-driver-vmware => github.com/lbogdan/docker-machine-driver-vmware v0.2.0
```
medyagh, line 238: @lbogdan do you mind making this PR to the new Org https://github.com/minikube-machine/machine-driver-vmware ?

medyagh (approved):

lbogdan:
Please don't merge this yet, I want to get back to it on Friday, weekend at the latest.

go.mod
lbogdan: Yes, I will get to it on Friday, weekend at the latest.

spowelljr (approved):

go.mod
spowelljr: For getting the release out in time we'll merge this and then change it back to our fork once the change gets merged upstream.

afbjorklund: Created https://github.com/machine-drivers/docker-machine-driver-vmware/issues/75

Issue #15841, opened 2023-02-13
Bump ISO Go version back up to 1.20 after buildroot fix


spowelljr:
Bumped Go version back down to 1.19.5 from 1.20 due in https://github.com/kubernetes/minikube/pull/15752/commits/3006b3ea0aa9bf5e40414417b40cb7062a4a4ffd due to https://lore.kernel.org/buildroot/CA+h8R2rtcynkCBsz=_9yANOEguyPCOcQDj8_ns+cv8RS8+8t9A@mail.gmail.com/T/

spowelljr:
I don't believe that buildroot has released a new version yet so I don't think any action is able to be taken yet.

https://buildroot.org/news.html

XDRAGON2002:
Sure, I'll keep this issue on a tracking thread then.

PR #16798, opened 2023-06-30
Add docker buildx iso

spowelljr:


afbjorklund:
Sooner or later we need to start cleaning up the technical debt, and unify the packages for all arch...

For binaries that means we need to use the proper name after downloading, and not just the URL base.

`curl -RLOJ https://github.com/docker/buildx/archive/refs/tags/v0.11.0.tar.gz`

`curl: Saved to filename 'buildx-0.11.0.tar.gz'`

For sources it means to use more of the existing make macros, like the github one (`$(call github,x,y,z,v)`).

It looks more like this one: https://github.com/buildroot/buildroot/tree/master/package/docker-compose

----

But that is a major refactoring, and something that can  happen well after docker-build has been merged.

afbjorklund (approved):

medyagh (approved):

spowelljr:
ok-to-build-iso

spowelljr:
ISO build failed as docker-buildx requires Go 1.20 but we're stuck on Go 1.19 for the ISO due to https://github.com/kubernetes/minikube/issues/15841, created a PR to update buildroot https://github.com/kubernetes/minikube/pull/16838

PR #16672, opened 2023-06-11
Replace machine-drivers with minikube-machine

afbjorklund:
The new repository is a "filtered" version of the old docker-machine, with only libmachine and local drivers.

```
drivers/
├── errdriver
├── fakedriver
├── generic
├── hyperv
├── none
└── virtualbox

```

It provides the basis for making a new incompatible API version of libmachine, to be called minikube-machine.

https://github.com/minikube-machine/machine

----

Note to reviewer:

This is a no-op, as far as "vendor" is concerned.

PR #16804, opened 2023-07-02
Use minikube-machine repository for drivers

afbjorklund:
We want to upgrade libmachine beyond docker/machine,
so need to start using our own version of the drivers.

For now they still import github.com/docker/machine@v0.16.2,
but soon they will import github.com/minikube-machine/machine

----

Follow up to previous PR, there is **no** code change here - only URL.
* #16672

Next would be to change all references to use minikube-machine,
and all binaries to start using `minikube-machine-driver-*` instead.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -232,8 +232,10 @@ require (
 
 replace (
 	git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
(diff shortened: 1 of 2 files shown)


afbjorklund:
@lbogdan this would be the new repository in the future, where we could merge your vmware patch

here we don't have to bother about backwards compatibility with docker-machine, but only with minikube

PR #16915, opened 2023-07-19
Release v1.31.1

spowelljr:


cmd/minikube/cmd/start_test.go
```diff
@@ -840,3 +840,23 @@ func TestValidateStaticIP(t *testing.T) {
 		}
 	}
 }
+
+func TestImageMatchesBinaryVersion(t *testing.T) {
+	tests := []struct {
+		imageVersion  string
+		binaryVersion string
+		want          bool
```
medyagh, line 848: better name would be, want what
like wantFail, mustPass..or something like that

medyagh (approved):

PR #16936, opened 2023-07-23
Solves docker cruntime notavail bug for --driver=none

### LINKED ISSUE #5549: Starting minikube without docker requires... docker

### LINKED ISSUE #10908: None driver should take care of --container-runtime parameter

### LINKED ISSUE #16722: Starting none driver w/ containerd w/o docker installed fails precheck
If you try doing as the title says you get the following:
```
$ minikube start --driver none --container-runtime=containerd
😄  minikube v1.30.1 on Debian 11.7 (kvm/amd64)
❗  Using the 'containerd' runtime with the 'none' driver is an untested configuration!
🤹  Running on localhost (CPUs=8, Memory=30095MB, Disk=9853MB) ...
🤦  StartHost failed, but will try again: creating host: create: precreate: exec: "docker": executable file not found in $PATH
🤹  Running on localhost (CPUs=8, Memory=30095MB, Disk=9853MB) ...
😿  Failed to start none bare metal machine. Running "minikube delete" may fix it: creating host: create: precreate: exec: "docker": executable file not found in $PATH

❌  Exiting due to GUEST_PROVISION: error provisioning guest: Failed to start host: creating host: create: precreate: exec: "docker": executable file not found in $PATH

```

The cause is in:
https://github.com/kubernetes/minikube/blob/ad4acb0186a20f83472c7810dd446f84b3f8765d/pkg/minikube/machine/start.go#L135-L151

`runtime` is a private func in the none driver struct
https://github.com/kubernetes/minikube/blob/ad4acb0186a20f83472c7810dd446f84b3f8765d/pkg/drivers/none/none.go#L47-L53

So when the `Marshal` occurs, the runtime data doesn't get passed
https://github.com/kubernetes/minikube/blob/ad4acb0186a20f83472c7810dd446f84b3f8765d/pkg/minikube/machine/start.go#L143

In `NewHost`
https://github.com/kubernetes/minikube/blob/ad4acb0186a20f83472c7810dd446f84b3f8765d/pkg/minikube/machine/client.go#L98

Which calls
https://github.com/kubernetes/minikube/blob/ad4acb0186a20f83472c7810dd446f84b3f8765d/pkg/minikube/registry/drvs/none/none.go#L39

Which calls
https://github.com/kubernetes/minikube/blob/ad4acb0186a20f83472c7810dd446f84b3f8765d/pkg/drivers/none/none.go#L65

And since `Type` is not set, defaults to Docker CRI
https://github.com/kubernetes/minikube/blob/ad4acb0186a20f83472c7810dd446f84b3f8765d/pkg/minikube/cruntime/cruntime.go#L209-L213

So then
https://github.com/kubernetes/minikube/blob/ad4acb0186a20f83472c7810dd446f84b3f8765d/pkg/drivers/none/none.go#L81-L83

And since the runtime is now Docker calls
https://github.com/kubernetes/minikube/blob/ad4acb0186a20f83472c7810dd446f84b3f8765d/pkg/minikube/cruntime/docker.go#L110-L125

It actually skips the `cri-dockerd` and `dockerd` checks because nothing is passed in when the CRI is created so it defaults to `v0.0.0`.

x7upLime:
Solves the docker cruntime not avail bug, that occurs whenever using none driver with any container runtime other than docker.

fixes #5549
fixes #10908
fixes #16722

#### minikube start --driver=none --container-runtime=containerd
```
ubuntu@minikube:~/minikube$ sudo -E ./out/minikube start --driver=none --container-runtime=containerd
😄  minikube v1.31.1 on Ubuntu 22.10 (kvm/amd64)
❗  Using the 'containerd' runtime with the 'none' driver is an untested configuration!
🤹  Running on localhost (CPUs=4, Memory=3905MB, Disk=4781MB) ...
ℹ️  OS release is Ubuntu 22.10
📦  Preparing Kubernetes v1.27.3 on containerd 1.6.12-0ubuntu1 ...
    > kubeadm.sha256:  64 B / 64 B [-------------------------] 100.00% ? p/s 0s
    > kubectl.sha256:  64 B / 64 B [-------------------------] 100.00% ? p/s 0s
    > kubelet.sha256:  64 B / 64 B [-------------------------] 100.00% ? p/s 0s
    > kubelet:  101.24 MiB / 101.24 MiB [----------] 100.00% 61.89 MiB p/s 1.8s
    > kubectl:  46.98 MiB / 46.98 MiB [------------] 100.00% 22.64 MiB p/s 2.3s
    > kubeadm:  45.93 MiB / 45.93 MiB [--------------] 100.00% 1.68 MiB p/s 28s
    ▪ Configuring RBAC rules ...

❗  kubectl and minikube configuration will be stored in /home/ubuntu

    ▪ sudo mv /home/ubuntu/.kube /home/ubuntu/.minikube $HOME

```

#### which docker
```
ubuntu@minikube:~/minikube$ which docker

```

#### which containerd
```
ubuntu@minikube:~/minikube$ which containerd
/usr/bin/containerd
```

#### ubuntu@minikube:~/minikube$ sudo -E ./out/minikube kubectl -- get po -A
Needs `sudo sysctl fs.protected_regular=0`.   
Check: HOST_JUJU_LOCK_PERMISSION [here](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/reason/known_issues.go)
```
ubuntu@minikube:~/minikube$ sudo -E ./out/minikube kubectl -- get po -A
NAMESPACE     NAME                               READY   STATUS    RESTARTS   AGE
kube-system   coredns-5d78c9869d-krprh           1/1     Running   0          5m59s
kube-system   etcd-minikube                      1/1     Running   0          6m11s
kube-system   kube-apiserver-minikube            1/1     Running   0          6m11s
kube-system   kube-controller-manager-minikube   1/1     Running   0          6m11s
kube-system   kube-proxy-mh7bh                   1/1     Running   0          5m58s
kube-system   kube-scheduler-minikube            1/1     Running   0          6m11s
kube-system   storage-provisioner                1/1     Running   0          6m10s

```

### DIFF
--- a/pkg/drivers/none/none.go
+++ b/pkg/drivers/none/none.go
@@ -47,9 +47,10 @@ var cleanupPaths = []string{
 type Driver struct {
 	*drivers.BaseDriver
 	*pkgdrivers.CommonDriver
-	URL     string
-	runtime cruntime.Manager
-	exec    command.Runner
+	URL        string
+	runtime    cruntime.Manager
+	exec       command.Runner
+	NodeConfig Config
 }
 
 // Config is configuration for the None driver
@@ -61,24 +62,29 @@ type Config struct {
 
 // NewDriver returns a fully configured None driver
 func NewDriver(c Config) *Driver {
-	runner := command.NewExecRunner(true)
-	runtime, err := cruntime.New(cruntime.Config{Type: c.ContainerRuntime, Runner: runner})
-	// Libraries shouldn't panic, but there is no way for drivers to return error :(
-	if err != nil {
-		klog.Fatalf("unable to create container runtime: %v", err)
-	}
 	return &Driver{
 		BaseDriver: &drivers.BaseDriver{
 			MachineName: c.MachineName,
 			StorePath:   c.StorePath,
 		},
-		runtime: runtime,
-		exec:    runner,
+		NodeConfig: c,
 	}
 }
 
 // PreCreateCheck checks for correct privileges and dependencies
 func (d *Driver) PreCreateCheck() error {
+	runner := command.NewExecRunner(true)
+	runtime, err := cruntime.New(cruntime.Config{Type: d.NodeConfig.ContainerRuntime, Runner: runner})
+	// Libraries shouldn't panic, but there is no way for drivers to return error :(
+	if err != nil {
+		klog.Fatalf("unable to create container runtime: %v", err)
+	}
+
+	d.runtime = runtime
+	d.exec = runner
+
+	fmt.Printf("We've put a command runner inside Driver in PrecreateCheck: %#v\n\n\n", d.exec)
+
 	return d.runtime.Available()
 }
 
@@ -125,6 +131,9 @@ func (d *Driver) GetURL() (string, error) {
 
 // GetState returns the state that the host is in (running, stopped, etc)
 func (d *Driver) GetState() (state.State, error) {
+	runner := command.NewExecRunner(true)
+	d.exec = runner
+
 	hostname, port, err := kubeconfig.Endpoint(d.BaseDriver.MachineName)
 	if err != nil {
 		klog.Warningf("unable to get port: %v", err)
--- a/pkg/drivers/none/none.go
+++ b/pkg/drivers/none/none.go
@@ -83,8 +83,6 @@ func (d *Driver) PreCreateCheck() error {
 	d.runtime = runtime
 	d.exec = runner
 
-	fmt.Printf("We've put a command runner inside Driver in PrecreateCheck: %#v\n\n\n", d.exec)
-
 	return d.runtime.Available()
 }
 


pkg/drivers/none/none.go
```diff
@@ -61,24 +62,29 @@ type Config struct {
 
 // NewDriver returns a fully configured None driver
 func NewDriver(c Config) *Driver {
-	runner := command.NewExecRunner(true)
-	runtime, err := cruntime.New(cruntime.Config{Type: c.ContainerRuntime, Runner: runner})
-	// Libraries shouldn't panic, but there is no way for drivers to return error :(
-	if err != nil {
-		klog.Fatalf("unable to create container runtime: %v", err)
-	}
```
x7upLime, line 69: There is just no point in initializing d.exec, d.runtime at this phase, since init is only called before the json marshal/unmarshal mechanism of libmachine, i.e. it would not survive that.

pkg/drivers/none/none.go
```diff
@@ -61,24 +62,29 @@ type Config struct {
...
+	// Libraries shouldn't panic, but there is no way for drivers to return error :(
+	if err != nil {
+		klog.Fatalf("unable to create container runtime: %v", err)
+	}
+
+	d.runtime = runtime
+	d.exec = runner
+
+	fmt.Printf("We've put a command runner inside Driver in PrecreateCheck: %#v\n\n\n", d.exec)
+
```
x7upLime, line 87: we initialize runner/runtime here, since it's the first method from the driver, that gets called when creating the machine

pkg/drivers/none/none.go
```diff
@@ -125,6 +131,9 @@ func (d *Driver) GetURL() (string, error) {
 
 // GetState returns the state that the host is in (running, stopped, etc)
 func (d *Driver) GetState() (state.State, error) {
+	runner := command.NewExecRunner(true)
+	d.exec = runner
+
```
x7upLime, line 136: we initialize a runner here, since it is expected after provisioning phase, for enabling addons

medyagh (changes requested):
plz paste the output of minikube start with the followinfg conditions

1- no docker in path
$ which docker
2- have containerd installed
3- minikube start --driver=none --container-runtime=containerd
4- kubecltl get pods -A

x7upLime:
Reminder for myself:
as suggested by @medyagh, I will try to put the runtime and exec fields of the none driver structs inside the BaseDriver or CommonDriver to see where that gets us for this and maybe other issues, in a separate pr.

x7upLime:
Unfortunately putting the runtime/exec fields inside the CommonDriver still won't get those past the marshal/unmarshal stage in the driver lifecycle:

```
$ sudo -E ./out/minikube start --driver=none --container-runtime=containerd

😄  minikube v1.31.1 on Ubuntu 22.10 (kvm/amd64)
❗  Using the 'containerd' runtime with the 'none' driver is an untested configuration!
🤹  Running on localhost (CPUs=4, Memory=3905MB, Disk=4781MB) ...
🤦  StartHost failed, but will try again: creating host: create: precreate: exec: "docker": executable file not found in $PATH
😿  Failed to start none bare metal machine. Running "minikube delete" may fix it: boot lock: unable to open /tmp/juju-mk827fcceb822ddd434756f789a161c4dd798db6: permission denied

❌  Exiting due to HOST_JUJU_LOCK_PERMISSION: Failed to start host: boot lock: unable to open /tmp/juju-mk827fcceb822ddd434756f789a161c4dd798db6: permission denied
💡  Suggestion: Run 'sudo sysctl fs.protected_regular=0', or try a driver which does not require root, such as '--driver=docker'
🍿  Related issue: https://github.com/kubernetes/minikube/issues/6391
```

The patch that I used:
```
diff --git a/pkg/drivers/common.go b/pkg/drivers/common.go
index d7e3d3c44..da55b9726 100644
--- a/pkg/drivers/common.go
+++ b/pkg/drivers/common.go
@@ -34,6 +34,8 @@ import (
 	"github.com/pkg/errors"
 
 	"k8s.io/klog/v2"
+	"k8s.io/minikube/pkg/minikube/cruntime"
 	"k8s.io/minikube/pkg/util"
 )
 
@@ -86,7 +88,10 @@ func CreateRawDisk(diskPath string, sizeMB int) error {
 }
 
 // CommonDriver is the common driver base class
-type CommonDriver struct{}
+type CommonDriver struct {
+	Runtime cruntime.Manager
+	Exec    command.Runner
+}
 
 // GetCreateFlags is not implemented yet
 func (d *CommonDriver) GetCreateFlags() []mcnflag.Flag {
diff --git a/pkg/drivers/none/none.go b/pkg/drivers/none/none.go
index 060150c26..55b0ae2ce 100644
--- a/pkg/drivers/none/none.go
+++ b/pkg/drivers/none/none.go
@@ -48,8 +48,6 @@ type Driver struct {
 	*drivers.BaseDriver
 	*pkgdrivers.CommonDriver
 	URL        string
-	runtime    cruntime.Manager
-	exec       command.Runner
 	NodeConfig Config
 }
 
@@ -62,28 +60,29 @@ type Config struct {
 
 // NewDriver returns a fully configured None driver
 func NewDriver(c Config) *Driver {
+	runner := command.NewExecRunner(true)
+	runtime, err := cruntime.New(cruntime.Config{Type: c.ContainerRuntime, Runner: runner})
+	// Libraries shouldn't panic, but there is no way for drivers to return error :(
+	if err != nil {
+		klog.Fatalf("unable to create container runtime: %v", err)
+	}
+
 	return &Driver{
 		BaseDriver: &drivers.BaseDriver{
 			MachineName: c.MachineName,
 			StorePath:   c.StorePath,
 		},
 		NodeConfig: c,
+		CommonDriver: &pkgdrivers.CommonDriver{
+			Runtime: runtime,
+			Exec:    runner,
+		},
 	}
 }
 
 // PreCreateCheck checks for correct privileges and dependencies
 func (d *Driver) PreCreateCheck() error {
-	runner := command.NewExecRunner(true)
-	runtime, err := cruntime.New(cruntime.Config{Type: d.NodeConfig.ContainerRuntime, Runner: runner})
-	// Libraries shouldn't panic, but there is no way for drivers to return error :(
-	if err != nil {
-		klog.Fatalf("unable to create container runtime: %v", err)
-	}
-
-	d.runtime = runtime
-	d.exec = runner
-
-	return d.runtime.Available()
+	return d.Runtime.Available()
 }
 
 // Create a host using the driver's config
@@ -129,8 +128,8 @@ func (d *Driver) GetURL() (string, error) {
 
 // GetState returns the state that the host is in (running, stopped, etc)
 func (d *Driver) GetState() (state.State, error) {
-	runner := command.NewExecRunner(true)
-	d.exec = runner
+	// runner := command.NewExecRunner(true)
+	// d.Exec = runner
 
 	hostname, port, err := kubeconfig.Endpoint(d.BaseDriver.MachineName)
 	if err != nil {
@@ -139,7 +138,7 @@ func (d *Driver) GetState() (state.State, error) {
 	}
 
 	// Confusing logic, as libmachine.Stop will loop until the state == Stopped
-	ast, err := kverify.APIServerStatus(d.exec, hostname, port)
+	ast, err := kverify.APIServerStatus(d.Exec, hostname, port)
 	if err != nil {
 		return ast, err
 	}
@@ -149,17 +148,17 @@ func (d *Driver) GetState() (state.State, error) {
 		return state.Running, nil
 	}
 
-	return kverify.ServiceStatus(d.exec, "kubelet"), nil
+	return kverify.ServiceStatus(d.Exec, "kubelet"), nil
 }
 
 // Kill stops a host forcefully, including any containers that we are managing.
 func (d *Driver) Kill() error {
-	if err := sysinit.New(d.exec).ForceStop("kubelet"); err != nil {
+	if err := sysinit.New(d.Exec).ForceStop("kubelet"); err != nil {
 		klog.Warningf("couldn't force stop kubelet. will continue with kill anyways: %v", err)
 	}
 
 	// First try to gracefully stop containers
-	containers, err := d.runtime.ListContainers(cruntime.ListContainersOptions{})
+	containers, err := d.Runtime.ListContainers(cruntime.ListContainersOptions{})
 	if err != nil {
 		return errors.Wrap(err, "containers")
 	}
@@ -167,18 +166,18 @@ func (d *Driver) Kill() error {
 		return nil
 	}
 	// Try to be graceful before sending SIGKILL everywhere.
-	if err := d.runtime.StopContainers(containers); err != nil {
+	if err := d.Runtime.StopContainers(containers); err != nil {
 		return errors.Wrap(err, "stop")
 	}
 
-	containers, err = d.runtime.ListContainers(cruntime.ListContainersOptions{})
+	containers, err = d.Runtime.ListContainers(cruntime.ListContainersOptions{})
 	if err != nil {
 		return errors.Wrap(err, "containers")
 	}
 	if len(containers) == 0 {
 		return nil
 	}
-	if err := d.runtime.KillContainers(containers); err != nil {
+	if err := d.Runtime.KillContainers(containers); err != nil {
 		return errors.Wrap(err, "kill")
 	}
 	return nil
@@ -191,7 +190,7 @@ func (d *Driver) Remove() error {
 	}
 	klog.Infof("Removing: %s", cleanupPaths)
 	args := append([]string{"rm", "-rf"}, cleanupPaths...)
-	if _, err := d.exec.RunCmd(exec.Command("sudo", args...)); err != nil {
+	if _, err := d.Exec.RunCmd(exec.Command("sudo", args...)); err != nil {
 		klog.Errorf("cleanup incomplete: %v", err)
 	}
 	return nil
@@ -199,7 +198,7 @@ func (d *Driver) Remove() error {
 
 // Restart a host
-	return sysinit.New(d.exec).Restart("kubelet")
+	return sysinit.New(d.Exec).Restart("kubelet")
 }
 
 // Start a host
@@ -218,18 +217,18 @@ func (d *Driver) Start() error {
 
 // Stop a host gracefully, including any containers that we are managing.
-	if err := sysinit.New(d.exec).Stop("kubelet"); err != nil {
+	if err := sysinit.New(d.Exec).Stop("kubelet"); err != nil {
 		klog.Warningf("couldn't stop kubelet. will continue with stop anyways: %v", err)
-		if err := sysinit.New(d.exec).ForceStop("kubelet"); err != nil {
+		if err := sysinit.New(d.Exec).ForceStop("kubelet"); err != nil {
 			klog.Warningf("couldn't force stop kubelet. will continue with stop anyways: %v", err)
 		}
 	}
-	containers, err := d.runtime.ListContainers(cruntime.ListContainersOptions{})
+	containers, err := d.Runtime.ListContainers(cruntime.ListContainersOptions{})
 	if err != nil {
 		return errors.Wrap(err, "containers")
 	}
 	if len(containers) > 0 {
-		if err := d.runtime.StopContainers(containers); err != nil {
+		if err := d.Runtime.StopContainers(containers); err != nil {
 			return errors.Wrap(err, "stop containers")
 		}
 	}
```

spowelljr:
@x7upLime It looks like because you removed setting exec and runtime in `NewDriver` they're nil and can run into nil panics

PR #16972, opened 2023-07-31
Verify containerd storage separately from docker

afbjorklund:
Previously docker and containerd were both using docker checks, for preload

### DIFF
--- a/hack/preload-images/preload_images.go
+++ b/hack/preload-images/preload_images.go
@@ -41,16 +41,17 @@ const (
 )
 
 var (
-	dockerStorageDriver = "overlay2"
-	podmanStorageDriver = "overlay"
-	containerRuntimes   = []string{"docker", "containerd", "cri-o"}
-	k8sVersions         []string
-	k8sVersion          = flag.String("kubernetes-version", "", "desired Kubernetes version, for example `v1.17.2`")
-	noUpload            = flag.Bool("no-upload", false, "Do not upload tarballs to GCS")
-	force               = flag.Bool("force", false, "Generate the preload tarball even if it's already exists")
-	limit               = flag.Int("limit", 0, "Limit the number of tarballs to generate")
-	armUpload           = flag.Bool("arm-upload", false, "Upload the arm64 preload tarballs to GCS")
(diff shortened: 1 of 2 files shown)


hack/preload-images/preload_images.go
```diff
@@ -41,16 +41,17 @@ const (
...
-	containerRuntimes   = []string{"docker", "containerd", "cri-o"}
-	k8sVersions         []string
-	k8sVersion          = flag.String("kubernetes-version", "", "desired Kubernetes version, for example `v1.17.2`")
-	noUpload            = flag.Bool("no-upload", false, "Do not upload tarballs to GCS")
-	force               = flag.Bool("force", false, "Generate the preload tarball even if it's already exists")
-	limit               = flag.Int("limit", 0, "Limit the number of tarballs to generate")
-	armUpload           = flag.Bool("arm-upload", false, "Upload the arm64 preload tarballs to GCS")
-	armPreloadsDir      = flag.String("arm-preloads-dir", "artifacts", "Directory containing the arm64 preload tarballs")
+	dockerStorageDriver   = "overlay2"
+	containerdSnapshotter = "overlayfs"
```
medyagh, line 45: @afbjorklund do u mind adding details, since when containerd snapshotter is different from docker storage driver?
and how is our current containerd NOT broken with this preload ?

do we need to Bump a new preload version for force all new ones be with this snapshotter?

afbjorklund: There are no changes, i.e. it is the same format. Just that we should not be checking docker, when looking for containerd config. In the future, one of them might change format - without affecting the other one.

Therefore, we should name the preload (**without** changing the contents):
preloaded-images-k8s-v18-v1.28.0-rc.0-containerd-overlayfs-amd64.tar.lz4

afbjorklund:
```go
        var storageDriver string
        if containerRuntime == "cri-o" {
                storageDriver = "overlay"
        } else {
                storageDriver = "overlay2"
        }
```

missing a code path for containerd

medyagh: I see makes sense

PR #17003, opened 2023-08-04
CI: Fix gopogh not installed & missing DB_BACKEND

spowelljr:
Adding missing `-db_backend` flag

Also fixed gopogh not being installed on upload integration report job

### DIFF
--- a/hack/jenkins/installers/check_install_gopogh.sh
+++ b/hack/jenkins/installers/check_install_gopogh.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+# Copyright 2023 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


hack/jenkins/installers/check_install_gopogh.sh
```diff
@@ -0,0 +1,25 @@
...
+
+set -eux -o pipefail
+
+# installing golang so we can go install gopogh
+./installers/check_install_golang.sh "/usr/local" || true
+
+# temporary: remove the old install of gopogh as it's taking priority over our current install, preventing updating
+sudo rm -f /usr/local/bin/gopogh
+
+go install github.com/medyagh/gopogh/cmd/gopogh@v0.19.0
```
medyagh, line 25: does this need to be added to the auto-updater?

medyagh (changes requested):

hack/jenkins/installers/check_install_gopogh.sh
spowelljr: Good catch, fixed

medyagh (approved):

Issue #16933, opened 2023-07-21
`docker-env` fails if `~/.ssh` doesn't exist


sethrylan:
### What Happened?

Starting with `v1.31.0`, `docker-env` will fail if the `~/.ssh` directory doesn't exist. This is the case on most linux hosts that have not generated ssh keys (e.g., short-lived VMs).

Introduced in https://github.com/kubernetes/minikube/pull/15452.

The workaround is to run `mkdir -p ~/.ssh`.

```
➜  minikube start     
😄  minikube v1.31.1 on Ubuntu 20.04 (docker/amd64)
✨  Automatically selected the docker driver. Other choices: none, ssh
📌  Using Docker driver with root privileges
👍  Starting control plane node minikube in cluster minikube
🚜  Pulling base image ...
💾  Downloading Kubernetes v1.27.3 preload ...
    > preloaded-images-k8s-v18-v1...:  393.19 MiB / 393.19 MiB  100.00% 79.18 M
    > gcr.io/k8s-minikube/kicbase...:  446.99 MiB / 447.62 MiB  99.86% 53.11 Mi
🔥  Creating docker container (CPUs=2, Memory=2200MB) ...
🐳  Preparing Kubernetes v1.27.3 on Docker 24.0.4 ...
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔗  Configuring bridge CNI (Container Networking Interface) ...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🔎  Verifying Kubernetes components...
🌟  Enabled addons: storage-provisioner, default-storageclass
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default

➜ ls ~/.ssh                                                                            
ls: cannot access '/home/codespace/.ssh': No such file or directory

➜  minikube docker-env --alsologtostderr
I0721 22:17:37.953472    8919 out.go:296] Setting OutFile to fd 1 ...
I0721 22:17:37.955057    8919 out.go:348] isatty.IsTerminal(1) = true
I0721 22:17:37.955074    8919 out.go:309] Setting ErrFile to fd 2...
I0721 22:17:37.955084    8919 out.go:348] isatty.IsTerminal(2) = true
I0721 22:17:37.956115    8919 root.go:338] Updating PATH: /home/codespace/.minikube/bin
[...]

PR #17039, opened 2023-08-10
remove ssh-add flag and fix calling appendKnownHelper when not needed

ComradeProgrammer:
Related #16933 #15452
1. appendKnownHelper  should only be called when sshHost is true. appendKnownHelper should not be run when user is trying to connect to minikube's docker daemon with HTTP
2. before #15452 , to run minikube docker-env with ssh connection, user have to manually start a ssh-agent, and use ssh-add flag. But in #15452 , we added a feature, which automatically help users to do these two things, and user don't need to specify ssh-add or start a ssh-agent. However, no automation for "docker runtime with ssh connection". This makes docker-env's behavior inconsistent
3. so I think perhaps we should add this automation for  "docker runtime with ssh connection" too. And if so, we no longer need ssh-add flag. As long as ssh-host is set, we will automatically do all these things for the user(start ssh-agent, run ssh-add). Continuing to use ssh-add flag will make things  confusing.

@spowelljr  @medyagh


cmd/minikube/cmd/docker-env.go
```diff
@@ -405,7 +405,7 @@ docker-cli install instructions: https://minikube.sigs.k8s.io/docs/tutorials/doc
 			exit.Error(reason.InternalDockerScript, "Error generating set output", err)
 		}
 
-		if sshAdd {
+		if sshHost {
```
spowelljr, line 408: We may want to think about this as this is a breaking change. Scripts that were set to use `--ssh-add` will no longer execute this. We should leave the flag in and mark it deprecated in the flag description and tell users to use `--ssh-host` but have this be a `sshHost || sshAdd` to not break it.

cmd/minikube/cmd/docker-env.go
```diff
@@ -324,9 +323,10 @@ docker-cli install instructions: https://minikube.sigs.k8s.io/docs/tutorials/doc
 			// docker-env on containerd depends on nerdctld (https://github.com/afbjorklund/nerdctld) as "docker" daeomn
 			// and nerdctld daemon must be used with ssh connection (it is set in kicbase image's Dockerfile)
 			// so directly set --ssh-host --ssh-add to true, even user didn't specify them
-			sshAdd = true
 			sshHost = true
+		}
 
+		if sshHost {
```
spowelljr, line 329: Instead of having two `if sshHost` blocks, could we move this down to the second block?

spowelljr (changes requested):

medyagh (commented):
lets keep things backward campatible

cmd/minikube/cmd/docker-env.go
ComradeProgrammer: Yes.
Now it has been done

cmd/minikube/cmd/docker-env.go
```diff
@@ -326,7 +326,11 @@ docker-cli install instructions: https://minikube.sigs.k8s.io/docs/tutorials/doc
 			// so directly set --ssh-host --ssh-add to true, even user didn't specify them
 			sshAdd = true
 			sshHost = true
+		}
+
+		d := co.CP.Host.Driver
 
+		if sshHost {
```
medyagh, line 333: this should be an OR so we dont break the old behavior

medyagh (changes requested):

cmd/minikube/cmd/docker-env.go
medyagh: so the question is @afbjorklund would user ever need to use ssh-add without Docker using SSH-HOST ? 
so do we need to some reason keep the old behviour or ssh-add to be working regardless if the user is using SSH or HTTPS

medyagh (changes requested):

medyagh:
lets go with most backward compatible approach

ComradeProgrammer:
@medyagh  The comments have been solved. Maybe you can have a look at it now?

medyagh:
@afbjorklund whats your opinion on this PR

cmd/minikube/cmd/docker-env.go
```diff
@@ -125,7 +125,7 @@ type DockerShellConfig struct {
 var (
 	noProxy              bool
 	sshHost              bool
-	sshAdd               bool
+	sshAdd               bool // depecrated
```
afbjorklund, line 128: deprecated

cmd/minikube/cmd/docker-env.go
```diff
@@ -692,7 +693,7 @@ func init() {
 	defaultNoProxyGetter = &EnvNoProxyGetter{}
 	dockerEnvCmd.Flags().BoolVar(&noProxy, "no-proxy", false, "Add machine IP to NO_PROXY environment variable")
 	dockerEnvCmd.Flags().BoolVar(&sshHost, "ssh-host", false, "Use SSH connection instead of HTTPS (port 2376)")
-	dockerEnvCmd.Flags().BoolVar(&sshAdd, "ssh-add", false, "Add SSH identity key to SSH authentication agent")
+	dockerEnvCmd.Flags().BoolVar(&sshAdd, "ssh-add", false, "Depecrated: Add SSH identity key to SSH authentication agent")
```
afbjorklund, line 696: Deprecated

afbjorklund (changes requested):
Some typos.

Not sure it actually has to be deprecated, if it is truly not needed anymore than maybe it should just be removed from the documentation and made it into a no-op instead.

afbjorklund:
But I never understood why we have to start a second `ssh-agent` anyway, usually it is started when you log in.

* #15452

site/content/en/docs/commands/docker-env.md
```diff
@@ -28,7 +28,7 @@ minikube docker-env [flags]
       --no-proxy        Add machine IP to NO_PROXY environment variable
   -o, --output string   One of 'text', 'yaml' or 'json'.
       --shell string    Force environment to be configured for a specified shell: [fish, cmd, powershell, tcsh, bash, zsh], default is auto-detect
-      --ssh-add         Add SSH identity key to SSH authentication agent
+      --ssh-add         Depecrated: Add SSH identity key to SSH authentication agent
```
medyagh, line 31: fix spelling

medyagh (changes requested):

PR #14846, opened 2022-08-23
Add new `minikube config update` sub command

presztak:
This PR: 
1. Adds new command `minikube config update` which fixes paths in machine config.json file after changing MINIKUBE_HOME location.
2. Adds warn message when path location change in config.json will be detected.

Above solves a problem when MINIKUBE_HOME path will be changed after cluster creation what can lead to following error:
```
* Restarting existing docker container for "design-time" ...
! StartHost failed, but will try again: provision: Error getting config for native Go SSH: open /Users/pre/.minikube/machines/design-time/id_rsa: no such file or directory
* Updating the running docker "design-time" container ...
* Failed to start docker container. Running "minikube delete -p design-time" may fix it: provision: Error getting config for native Go SSH: open /Users/pre/.minikube/machines/design-time/id_rsa: no such file or directory

X Exiting due to GUEST_SSH_CERT_NOT_FOUND: Failed to start host: provision: Error getting config for native Go SSH: open /Users/pre/.minikube/machines/design-time/id_rsa: no such file or directory
* Suggestion: minikube is missing files relating to your guest environment. This can be fixed by running 'minikube delete'
* Related issue: https://github.com/kubernetes/minikube/issues/9130
```

To make things works we need to:

- minikube stop
- mv ~/.minikube ~/something_else/.minikube
- export MINIKUBE_HOME="${HOME}/something_else/.minikube"
- minikube config update
- minikube start

Fixes: #14466

PR #17079, opened 2023-08-17
Allow importing of profile configurations

jelemux:
This PR adds an `import-profile` flag to the `minikube start` command.
It can be used to pass a path to a profile configuration file that should be imported.

I think this use-case is valid and important as configuration-as-code is a common practice nowadays.
It also coincides with minikubes principles of being user-friendly and developer-focused.


medyagh:
@jelemux do you mind adding output of minikube after this PR with a few examples ?

medyagh:
I am curious @jelemux could this problem be solved by this PR too ? https://github.com/kubernetes/minikube/pull/14846

Issue #16814, opened 2023-07-04
Error starting qemu2 VM


josetaylor06:
### What Happened?

Seems to be related to these dhcp connection attemps

0703 19:21:43.255376    5571 main.go:141] libmachine: STDERR: 
I0703 19:21:43.255402    5571 main.go:141] libmachine: DONE writing to /Users/jt/.minikube/machines/minikube/disk.qcow2.raw and /Users/jt/.minikube/machines/minikube/disk.qcow2
I0703 19:21:43.255407    5571 main.go:141] libmachine: Starting QEMU VM...
I0703 19:21:43.255474    5571 main.go:141] libmachine: executing: /usr/local/opt/socket_vmnet/bin/socket_vmnet_client /usr/local/var/run/socket_vmnet qemu-system-x86_64 -display none -accel hvf -m 4000 -smp 2 -boot d -cdrom /Users/jt/.minikube/machines/minikube/boot2docker.iso -qmp unix:/Users/jt/.minikube/machines/minikube/monitor,server,nowait -pidfile /Users/jt/.minikube/machines/minikube/qemu.pid -device virtio-net-pci,netdev=net0,mac=2a:f2:65:19:95:10 -netdev socket,id=net0,fd=3 -daemonize /Users/jt/.minikube/machines/minikube/disk.qcow2
I0703 19:21:43.875349    5571 main.go:141] libmachine: STDOUT: 
W0703 19:21:43.875375    5571 main.go:139] libmachine: STDERR: qemu-system-x86_64: warning: host doesn't support requested feature: CPUID.80000001H:ECX.svm [bit 2]
qemu-system-x86_64: warning: host doesn't support requested feature: CPUID.80000001H:ECX.svm [bit 2]

I0703 19:21:43.875388    5571 main.go:141] libmachine: Attempt 0
I0703 19:21:43.875418    5571 main.go:141] libmachine: Searching for 2a:f2:65:19:95:10 in /var/db/dhcpd_leases ...
I0703 19:21:43.875856    5571 main.go:141] libmachine: Found 1 entries in /var/db/dhcpd_leases!
I0703 19:21:43.875885    5571 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.2 HWAddress:ee:fc:78:b8:e9:2a ID:1,ee:fc:78:b8:e9:2a Lease:0x64a4b22d}
I0703 19:21:45.876815    5571 main.go:141] libmachine: Attempt 1
[...]

PR #17217, opened 2023-09-08
qemu: better cpu type and IP detection

raghavendra-talur:
This PR has two commits.

1. We increase the timeout for detecting the IP address of the qemu VM.
2. The microarchitecture level of the qemu VM is bumped up to enable RHEL9 and related variants. This is required to support container images that are built using ubi9.

### DIFF
--- a/pkg/minikube/registry/drvs/qemu2/qemu2.go
+++ b/pkg/minikube/registry/drvs/qemu2/qemu2.go
@@ -125,7 +125,9 @@ func configure(cc config.ClusterConfig, n config.Node) (interface{}, error) {
 	switch runtime.GOARCH {
 	case "amd64":
 		qemuMachine = "" // default
-		qemuCPU = ""     // default
+		// set cpu type to max to enable higher microarchitecture levels
+		// see https://lists.gnu.org/archive/html/qemu-devel/2022-08/msg04066.html for details
+		qemuCPU = "max"
 	case "arm64":
 		qemuMachine = "virt"
 		qemuCPU = "cortex-a72"
--- a/pkg/drivers/qemu/qemu.go
+++ b/pkg/drivers/qemu/qemu.go
@@ -510,7 +510,7 @@ func (d *Driver) Start() error {
 			return nil
 		}
 		// Implement a retry loop because IP address isn't added to dhcp leases file immediately
-		for i := 0; i < 30; i++ {
+		for i := 0; i < 60; i++ {
 			log.Debugf("Attempt %d", i)
 			err = getIP()
 			if err == nil {


raghavendra-talur:
Fixes: #16814

pkg/minikube/registry/drvs/qemu2/qemu2.go
```diff
@@ -125,7 +125,7 @@ func configure(cc config.ClusterConfig, n config.Node) (interface{}, error) {
 	switch runtime.GOARCH {
 	case "amd64":
 		qemuMachine = "" // default
-		qemuCPU = ""     // default
+		qemuCPU = "max"  // default
```
medyagh, line 128: Is there any documentation url we can read what is max and what are other options? And what are downsides of max vs others?

medyagh (changes requested):

pkg/minikube/registry/drvs/qemu2/qemu2.go
raghavendra-talur: I have added the documentation in the commit message. Happy to add it as a comment above this code if you think that is a better place.

medyagh: can u plz link the URL to the documentation I would like to evaluate the options and figure out what exactly is "max"

raghavendra-talur: RHEL9 and therefore the dependent distributions and base containers have
a minimum microarchitechture level of x86-64-v2.[1]

The default cpu level of the qemu driver on an Intel Mac is x86-64-v1
even if the host cpu has the x86-64-v2 capability. Using the cpu type of
max allows qemu to use all the host cpu capabilities that is also
supported by the hypervisor framework.[2]

This was fixed for linux host by changing the default within qemu but it
wasn't changed for MacOS.

[1] https://developers.redhat.com/blog/2021/01/05/building-red-hat-enterprise-linux-9-for-the-x86-64-v2-microarchitecture-level
[2] https://gitlab.com/qemu-project/qemu/-/issues/1686#note_1417714887
[3] https://lists.gnu.org/archive/html/qemu-devel/2022-08/msg04066.html

medyagh: thank you for this documentation link, I like the https://lists.gnu.org/archive/html/qemu-devel/2022-08/msg04066.html link you shared, would u plz put that in the comment of right befire 'max' in the code ? and then it would be good to go

raghavendra-talur: @medyagh Thanks for the review! I have added the link as a comment.

PR #17226, opened 2023-09-11
指定比给定更恰当

lilongfeng0902:
Assist in modifying English-Chinese translation documents.


jiafangtao:
"given cluster" vs "specified cluster", the original translation is good to me. If the word is the latter, I believe your translation will be more proper. Just my two cents.

lilongfeng0902:
> "given cluster" vs "specified cluster", the original translation is good to me. If the word is the latter, I believe your translation will be more proper. Just my two cents.

As you said, the original translation is better for you. When translating English into Chinese, the main thing is to use Chinese. I think it is better to express it more accurately. Let’s take a look at other people’s perspectives. Thanks for your review.

aiyijing:
'将节点添加到集群 [add nodes to the cluster]' is better for me. 😄Perhaps most people understand this way. but there is also a gap with your translation. So I believe it is best to maintain a literal translation. not causing misunderstandings is enough

lilongfeng0902:
> '将节点添加到集群 [add node to the cluster]' is better for me. 😄Perhaps most people understand this way. but there is also a gap with your translation. So I believe it is best to maintain a literal translation. not causing misunderstandings is enough

Maybe what you said is right. If there is another person who holds a literal translation view, I will close this PR.😄

uos-ljtian:
> '将节点添加到集群 [add node to the cluster]' is better for me. 😄Perhaps most people understand this way. but there is also a gap with your translation. So I believe it is best to maintain a literal translation. not causing misunderstandings is enough

我个人认为，只是直译的话，这个语言配置文件就没必要存在了。试想一下，这个配置文件真正的收益者是谁？如果是英语不好的中国人，他们更乐意看到什么样的翻译呢？如何能让他们以最低的成本了解内容信息，才是抉择的关键点。

我的观点为，不直译。

PR #17309, opened 2023-09-27
CI: Auto update kong image

spowelljr:


.github/workflows/update-kong-version.yml
```diff
@@ -0,0 +1,49 @@
+name: "update-kong-version"
+on:
+  workflow_dispatch:
+  schedule:
+    # every Monday at around 3 am pacific/10 am UTC
+    - cron: "0 10 * * 1"
```
medyagh, line 6: not for this PR but I think we can have the non-essential ones run less frequently we gotta come up with a solution for long term maintaince of all the auto-updater to be useful

PR #17287, opened 2023-09-20
Automate installing NVIDIA Container Toolkit --container-runtime

spowelljr:
```
$ minikube start --container-runtime nvidia-docker
😄  minikube v1.31.2 on Debian rodete
✨  Automatically selected the docker driver
📌  Using Docker driver with root privileges
👍  Starting control plane node minikube in cluster minikube
🚜  Pulling base image ...
🔥  Creating docker container (CPUs=2, Memory=32100MB) ...
🛠️   Installing the NVIDIA Container Toolkit...
🐳  Preparing Kubernetes v1.28.2 on Docker 24.0.6 ...
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔗  Configuring CNI (Container Networking Interface) ...
🔎  Verifying Kubernetes components...
    ▪ Using image nvcr.io/nvidia/k8s-device-plugin:v0.14.1
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: nvidia-device-plugin, storage-provisioner, default-storageclass
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default

$ cat << EOF | kubectl create -f -
apiVersion: v1
kind: Pod
metadata:
  name: nvidia-version-check
spec:
  restartPolicy: OnFailure
  containers:
  - name: nvidia-version-check
    image: "nvidia/cuda:11.0.3-base-ubuntu20.04"
    command: ["nvidia-smi"]
    resources:
      limits:
         nvidia.com/gpu: "1"
EOF
pod/nvidia-version-check created

$ kubectl logs nvidia-version-check
Fri Sep 22 18:45:31 2023       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 525.125.06   Driver Version: 525.125.06   CUDA Version: 12.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Quadro P1000        On   | 00000000:65:00.0 Off |                  N/A |
| 34%   26C    P8    N/A /  47W |     15MiB /  4096MiB |      0%      Default |
|                               |                      |                  N/A |
[...]

PR #17314, opened 2023-09-27
new flag -"-gpu" to enable Nvidia container runtime

spowelljr:
Rework of https://github.com/kubernetes/minikube/pull/17287 that removes the `nvidia-docker` container-runtime and uses the `--gpus` flag instead.

```
$ minikube start --gpus all
😄  minikube v1.31.2 on Debian rodete
🚜  Pulling base image ...
🔥  Creating docker container (CPUs=2, Memory=32100MB) ...
❗  Using GPUs with the Docker driver is experimental, if you experience any issues please report them at: https://github.com/kubernetes/minikube/issues/new/choose
🛠️   Installing the NVIDIA Container Toolkit...
🐳  Preparing Kubernetes v1.28.2 on Docker 24.0.6 ...
    ▪ Configuring RBAC rules ...
    ▪ Using image nvcr.io/nvidia/k8s-device-plugin:v0.14.1
🌟  Enabled addons: storage-provisioner, nvidia-device-plugin, default-storageclass

$ cat << EOF | kubectl create -f -
apiVersion: v1
kind: Pod
metadata:
  name: nvidia-version-check
spec:
  restartPolicy: OnFailure
  containers:
  - name: nvidia-version-check
    image: "nvidia/cuda:11.0.3-base-ubuntu20.04"
    command: ["nvidia-smi"]
    resources:
      limits:
         nvidia.com/gpu: "1"
EOF
pod/nvidia-version-check created

$ kubectl logs nvidia-version-check
Fri Sep 22 18:45:31 2023       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 525.125.06   Driver Version: 525.125.06   CUDA Version: 12.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Quadro P1000        On   | 00000000:65:00.0 Off |                  N/A |
| 34%   26C    P8    N/A /  47W |     15MiB /  4096MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
+-----------------------------------------------------------------------------+

$ minikube start --driver kvm --gpus all
😄  minikube v1.31.2 on Debian rodete (kvm/amd64)

❌  Exiting due to MK_USAGE: The gpus flag can only be used with the docker driver and docker container-runtime

$ minikube start --gpus cat
😄  minikube v1.31.2 on Debian rodete

❌  Exiting due to MK_USAGE: The gpus flag must be passed a value of "nvidia" or "all"
```

Screenshot 2023-09-27 at 1 40 51 PM

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -142,6 +142,7 @@ const (
 	socketVMnetPath         = "socket-vmnet-path"
 	staticIP                = "static-ip"
 	autoPauseInterval       = "auto-pause-interval"
+	enableNvidiaGPUs        = "enable-nvidia-gpus"
 )
 
 var (
@@ -204,6 +205,7 @@ func initMinikubeFlags() {
 	startCmd.Flags().Bool(disableMetrics, false, "If set, disables metrics reporting (CPU and memory usage), this can improve CPU usage. Defaults to false.")
 	startCmd.Flags().String(staticIP, "", "Set a static IP for the minikube cluster, the IP must be: private, IPv4, and the last octet must be between 2 and 254, for example 192.168.200.200 (Docker and Podman drivers only)")
 	startCmd.Flags().Duration(autoPauseInterval, time.Minute*1, "Duration of inactivity before the minikube VM is paused (default 1m0s).  To disable, set to 0s")
+	startCmd.Flags().Bool(enableNvidiaGPUs, false, "If set, allows pods to use your NVIDIA GPU(s) (Docker driver with Docker container-runtime only)")
 }
 
 // initKubernetesFlags inits the commandline flags for Kubernetes related options
@@ -595,6 +597,7 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 		},
 		MultiNodeRequested: viper.GetInt(nodes) > 1,
 		AutoPauseInterval:  viper.GetDuration(autoPauseInterval),
+		EnableNvidiaGPUs:   viper.GetBool(enableNvidiaGPUs),
 	}
 	cc.VerifyComponents = interpretWaitFlag(*cmd)
 	if viper.GetBool(createMount) && driver.IsKIC(drvName) {
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -205,7 +205,7 @@ func initMinikubeFlags() {
 	startCmd.Flags().Bool(disableMetrics, false, "If set, disables metrics reporting (CPU and memory usage), this can improve CPU usage. Defaults to false.")
 	startCmd.Flags().String(staticIP, "", "Set a static IP for the minikube cluster, the IP must be: private, IPv4, and the last octet must be between 2 and 254, for example 192.168.200.200 (Docker and Podman drivers only)")
 	startCmd.Flags().Duration(autoPauseInterval, time.Minute*1, "Duration of inactivity before the minikube VM is paused (default 1m0s).  To disable, set to 0s")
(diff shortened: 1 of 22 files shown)


cmd/minikube/cmd/start_flags.go
```diff
@@ -142,6 +142,7 @@ const (
 	socketVMnetPath         = "socket-vmnet-path"
 	staticIP                = "static-ip"
 	autoPauseInterval       = "auto-pause-interval"
+	enableNvidiaGPUs        = "enable-nvidia-gpus"
```
medyagh, line 145: seems to long word, I rather something shorter the user can type easier
how about --enable-nvidia

medyagh (changes requested):

cmd/minikube/cmd/start_flags.go
spowelljr: I think having something about GPUs would make it clear that it's GPU related. Especially if we add support for AMD, AMD makes CPUs and GPUs, `--enable-amd` would be broad and possibly confusing. `--nvidia-gpus` maybe?

medyagh: how about --gpus
with options such as 
nvidia, amd and all (if later we support both at same time)
and gives us room so in the future if we wanna have it Docker Style so the user can specify their own device

```
docker run --gpus device=GPU-3a23c669-1f69-c64e-cf85-44e9b07e7a2a
```

that would make users learn the syntax easier since it is already used in docker

https://docs.docker.com/engine/reference/commandline/run/#gpus

medyagh (changes requested):

cmd/minikube/cmd/start_flags.go
spowelljr: Renamed the `enable-nvidia-gpus` flag to `gpus` and made it a string flag that accepts `all` & `nvidia`.

cmd/minikube/cmd/start_flags.go
```diff
@@ -142,6 +142,7 @@ const (
 	socketVMnetPath         = "socket-vmnet-path"
 	staticIP                = "static-ip"
 	autoPauseInterval       = "auto-pause-interval"
+	gpus                    = "gpus"
```
medyagh, line 145: lets add -g for shorthand

medyagh (changes requested):

cmd/minikube/cmd/start_flags.go
spowelljr: Added the shorthand

medyagh (changes requested):
let also add a warning for the users that this feature is Beta and we like to get their feedback

site/content/en/docs/tutorials/nvidia.md
```diff
@@ -1,24 +1,73 @@
 ---
-title: "Using the Nvidia Addons"
-linkTitle: "Nvidia"
+title: "Using NVIDIA GPUs with minikube"
+linkTitle: "Using NVIDIA GPUs with minikube"
 weight: 1
 date: 2018-01-02
 ---
 
 ## Prerequisites
```
medyagh, line 8: Add minikube 1.32-beta or newer !

medyagh (changes requested):

site/content/en/docs/tutorials/nvidia.md
spowelljr: Done

medyagh (approved):
Thank you !

Issue #17242, opened 2023-09-13
Add memory validation for Hyper-V


paulomorgado:
### What Happened?

```
I0913 15:19:28.925045   14776 main.go:141] libmachine: Creating VHD
I0913 15:19:28.925045   14776 main.go:141] libmachine: [executing ==>] : C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -NonInteractive Hyper-V\New-VHD -Path 'C:\Users\***\.minikube\machines\hyperv3\fixed.vhd' -SizeBytes 10MB -Fixed
/ I0913 15:19:30.246678   14776 main.go:141] libmachine: [stdout =====>] :

ComputerName            : DESKTOP-001
Path                    : C:\Users\***\.minikube\machines\hyperv3\fixed.vhd
VhdFormat               : VHD
VhdType                 : Fixed
FileSize                : 10486272
Size                    : 10485760
MinimumSize             :
LogicalSectorSize       : 512
PhysicalSectorSize      : 512
BlockSize               : 0
ParentPath              :
DiskIdentifier          : 77350D06-FAC8-48F8-9461-70FAB8679688
FragmentationPercentage : 0
Alignment               : 1
Attached                : False
DiskNumber              :
IsPMEMCompatible        : False
AddressAbstractionType  : None
Number                  :




I0913 15:19:30.246678   14776 main.go:141] libmachine: [stderr =====>] :
I0913 15:19:30.246678   14776 main.go:141] libmachine: Writing magic tar header
I0913 15:19:30.246678   14776 main.go:141] libmachine: Writing SSH key tar header
I0913 15:19:30.249541   14776 main.go:141] libmachine: [executing ==>] : C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -NonInteractive Hyper-V\Convert-VHD -Path 'C:\Users\***\.minikube\machines\hyperv3\fixed.vhd' -DestinationPath 'C:\Users\***\.minikube\machines\hyperv3\disk.vhd' -VHDType Dynamic -DeleteSource
/ I0913 15:19:31.567001   14776 main.go:141] libmachine: [stdout =====>] :
I0913 15:19:31.567001   14776 main.go:141] libmachine: [stderr =====>] :
[...]

PR #17325, opened 2023-09-29
Add Hyper-V memory validation for odd numbers

spowelljr:
Closes https://github.com/kubernetes/minikube/issues/17242

Added logic to ensure that if `--memory max` is passed we don't generate an invalid memory value for Hyper-V and added validation to ensure odd MB memory value isn't passed to Hyper-V.

```
$ minikube start --driver hyperv --memory 6001
...

⛔  Exiting due to RSRC_INVALID_HYPERV_MEMORY: Hyper-V requires that memory MB be an even number, 6001MB was specified, try passing `--memory 6000`
```

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1180,7 +1180,7 @@ func validateRequestedMemorySize(req int, drvName string) {
 	}
 
 	if driver.IsHyperV(drvName) && req%2 == 1 {
-		exitIfNotForced(reason.RsrcInvalidHyperVMemory, "Hyper-V requires that memory MB be an even number, {{.memory}}MB was specified", out.V{"memory": req})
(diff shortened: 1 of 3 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -1178,6 +1178,10 @@ func validateRequestedMemorySize(req int, drvName string) {
 			`The requested memory allocation of {{.requested}}MiB does not leave room for system overhead (total system memory: {{.system_limit}}MiB). You may face stability issues.`,
 			out.V{"requested": req, "system_limit": sysLimit, "advised": advised})
 	}
+
+	if driver.IsHyperV(drvName) && req%2 == 1 {
+		exitIfNotForced(reason.RsrcInvalidHyperVMemory, "Hyper-V requires that memory MB be an even number, {{.memory}}MB was specified", out.V{"memory": req})
```
medyagh, line 1183: we could suggest, "--memory=req-1"

medyagh (changes requested):

cmd/minikube/cmd/start.go
spowelljr: Done

PR #17343, opened 2023-10-02
CI: Simplify build for arm64 functional tests

spowelljr:
We're running out of space while building our arm64 functional tests, but we're build binaries for every arch/OS but we only need Linux arm64, so simply the build.

Screenshot 2023-10-02 at 1 22 23 PM


medyagh (approved):

spowelljr:
> is cross build failing irrelevent?

Doesn't seem to be, the workflow file isn't used by Jenkins at all

PR #17394, opened 2023-10-10
CI: GitHub Actions only build required test binaries

spowelljr:
When building the binaries for the functional tests we build every possible combination, but we only run Linux & Darwin amd64 tests. To speed up the functional tests getting kicked off I made a new job that only builds Linux & Darwin amd64 test binaries and once it's done the functional tests start, it speed up from 19 minutes to 6.5 minutes. I still left the build_binaries job as we still want to make sure everything builds, but the functional tests don't rely on it before starting.

**Before 19 mins:**
Screenshot 2023-10-10 at 3 32 25 PM

**After: 6.5 mins**
Screenshot 2023-10-10 at 3 32 31 PM

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -19,8 +19,34 @@ permissions:
   contents: read
 
 jobs:
-  # Runs before all other jobs
-  # builds the minikube binaries
+  # Runs before before the functional tests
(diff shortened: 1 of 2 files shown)


medyagh:
nice

PR #17420, opened 2023-10-13
Remove unnecessary '-z' flag from gsutil tar extraction

chansuke:
This change removes the unnecessary `-z` flag from the 'tar' command used to extract the `gsutil.tar.gz` archive. 
The `-z` flag is typically used to specify gzip compression, but in this case, the archive is uncompressed. 
Removing the `-z` flag makes the command more concise and aligns it with the actual archive format.


afbjorklund:
If the file is uncompressed, it should not have a .tar.gz suffix?

chansuke:
@afbjorklund Yes, you're correct.  I've tested in local with several environment and realized that my change may not always be the case, and the -z flag is necessary for consistent behavior across different environments. I will close this PR.

PR #17487, opened 2023-10-25
oci.DaemonHost handle empty hostnames

### LINKED ISSUE #17452: Minikube commands fail with error `Exiting due to DRV_CP_ENDPOINT: Unable to get forwarded endpoint: failed to lookup ip for ""`
### What Happened?

I am following the tutorial in https://kubernetes.io/docs/tutorials/hello-minikube/ on Archlinux laptop with docker. I did `minikube start` and the docker container has started up. But when I run `minikube dashboard`, I get the error 

```
❌  Exiting due to DRV_CP_ENDPOINT: Unable to get forwarded endpoint: failed to lookup ip for ""
💡  Suggestion: 

    Recreate the cluster by running:
    minikube delete <no value>
    minikube start <no value>

│    Please also attach the following file to the GitHub issue:                             │
│    - /tmp/minikube_dashboard_e82743e0f55314733604a6150dcacabfbc842639_0.log               │
```

Deleting and restarting did not work, even after I restarted my entire machine.
When I run `minikube logs --file=logs.txt`, I get the same error. I have attached all the log files below.

Versions
* Linux: 6.5.7 
* Docker: 24.0.6
* minikube: 1.31.2

I'm not entirely sure what the issue is here since the only other reference I found to the error is on Windows. I can verify that the `minikube` container is running and I am able to log in to the container with `docker exec -it minikube bash` while the container is running.

### Attach the log file

[logs.txt](https://github.com/kubernetes/minikube/files/13058887/logs.txt)
[minikube_dashboard_e82743e0f55314733604a6150dcacabfbc842639_0.log](https://github.com/kubernetes/minikube/files/13058890/minikube_dashboard_e82743e0f55314733604a6150dcacabfbc842639_0.log)
[minikube_logs_273c584afb8f56625d862285474a6a21019a3238_0.log](https://github.com/kubernetes/minikube/files/13058891/minikube_logs_273c584afb8f56625d862285474a6a21019a3238_0.log)

### Operating System

Other

### Driver

Docker

pnasrat:
This handles DOCKER_HOST='tcp://:'

### DIFF
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -702,7 +702,9 @@ func DaemonHost(driver string) string {
 		if dh := os.Getenv(constants.DockerHostEnv); dh != "" {
 			if u, err := url.Parse(dh); err == nil {
 				if u.Host != "" {
-					return u.Hostname()
+					if hostname := u.Hostname(); hostname != "" {
+						return u.Hostname()
+					}
 				}
 			}
 		}
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -717,14 +717,14 @@ func IsExternalDaemonHost(driver string) bool {
 	if driver == Podman {
 		if dh := os.Getenv(constants.PodmanContainerHostEnv); dh != "" {
 			if u, err := url.Parse(dh); err == nil {
-				return u.Host != ""
+				return u.Hostname() != ""
 			}
 		}
 	}
 	if driver == Docker {
 		if dh := os.Getenv(constants.DockerHostEnv); dh != "" {
 			if u, err := url.Parse(dh); err == nil {
-				return u.Host != ""
+				return u.Hostname() != ""
 			}
 		}
 	}
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -701,10 +701,8 @@ func DaemonHost(driver string) string {
 	if driver == Docker {
 		if dh := os.Getenv(constants.DockerHostEnv); dh != "" {
 			if u, err := url.Parse(dh); err == nil {
-				if u.Host != "" {
-					if hostname := u.Hostname(); hostname != "" {
-						return u.Hostname()
-					}
+				if u.Hostname() != "" {
+					return u.Hostname()
 				}
 			}
 		}
--- a/pkg/drivers/kic/oci/oci_test.go
+++ b/pkg/drivers/kic/oci/oci_test.go
@@ -130,6 +130,7 @@ func TestDaemonHost(t *testing.T) {
 		{"docker", "", "unix:///var/run/something", "127.0.0.1", false},
 		{"docker", "", "tcp://127.0.0.1/foo", "127.0.0.1", true},
 		{"docker", "", "ssh://127.0.0.1/bar", "127.0.0.1", true},
+		{"docker", "", "tcp://:", "127.0.0.1", true},
 	}
 	for _, test := range tests {
 		t.Setenv("CONTAINER_HOST", test.containerHost)


medyagh:
thank you @pnasrat for this PR, do you mind specifying in what use cases this would happen ? is it a speicifc OS or if user have a custom env? is it for a corp setting?

pkg/drivers/kic/oci/oci.go
```diff
@@ -702,7 +702,9 @@ func DaemonHost(driver string) string {
 		if dh := os.Getenv(constants.DockerHostEnv); dh != "" {
 			if u, err := url.Parse(dh); err == nil {
 				if u.Host != "" {
-					return u.Hostname()
+					if hostname := u.Hostname(); hostname != "" {
```
spowelljr, line 705: The above `if u.Host != "" {` seems redundant since an empty `u.Host` returns an empty `u.Hostname()` as well. Can simplify it to just make sure the return value is not empty.

```
if u, err := url.Parse(dh); err == nil {
        if u.Hostname() != "" {
                return u.Hostname()
        }
}
```

spowelljr (changes requested):

pkg/drivers/kic/oci/oci.go
pnasrat: Done - should I also update the podman code above to just use u.Hostname?

pkg/drivers/kic/oci/oci_test.go
```diff
@@ -130,6 +130,7 @@ func TestDaemonHost(t *testing.T) {
 		{"docker", "", "unix:///var/run/something", "127.0.0.1", false},
 		{"docker", "", "tcp://127.0.0.1/foo", "127.0.0.1", true},
 		{"docker", "", "ssh://127.0.0.1/bar", "127.0.0.1", true},
+		{"docker", "", "tcp://:", "127.0.0.1", true},
```
medyagh, line 133: consider adding more unit tests from the upstream that you pointed it out
https://github.com/moby/moby/blob/1f096174d614276845527790e556a6a3911ae99c/opts/hosts_test.go#L26

pkg/drivers/kic/oci/oci.go
spowelljr: Yeah, it would likely suffer from the same issue

pkg/drivers/kic/oci/oci_test.go
pnasrat: @medyagh acknowledged but won't be until mid this week until I can get back to.

pnasrat: @medyagh I want to ensure I understand the intent of the code here and have a few clarifying questions around DaemonHost and IsExternalDaemonHost as I add tests 

When a docker host is set to a tcp://127.0.0.1 the current implementation considers that remote as `u.Hostname() != ""`, which on re-reading confused me. However thinking on it and comparing to ssh://127.0.0.1:1234 which may ssh onto a VM that is exposed locally. I'm fine with following that as is for tcp 127.0.0.1 (with default port) is considered remote particularly as it could be port forwarded 

Should all tcp and ssh URLs be considered remote?

I also have a question how compatible with the DOCKER_HOST or podman CONTAINER_HOST handling does the oci driver handling need to be here?

Doing this would probably require more specific handling here of supported url Scheme -  podman supports `ssh`, `unix`, and `tcp`, and `npipe` on windows. Docker cli supports the same.

For docker it treats DOCKER_HOST where no scheme is present (identified via ://)  docker defaults to `tcp://` ie should minikube handle DOCKER_HOST=1.1.1.1:2222 (url.Parse will raise an error here). 

https://github.com/docker/cli/blob/a6351d0cd3db3ff540fed4e50d7c35339c1742a0/opts/hosts.go#L75

Docker also supports `fd://` for dockerd that only makes sense running under systemd, generally for dockerd using `docker.socket` and probably doesn't make sense for minikube to support that but wanted to check that also

```go
{"docker", "", "tcp://127.0.0.1/foo", "127.0.0.1", true},
```

medyagh (approved):
thank you very much @pnasrat for this contribution

pkg/drivers/kic/oci/oci_test.go
medyagh: for podman we should do the same ! so lets change the line 695 too ! I believe both podman and docker should act the same, feel free to test to verify too. podman is free to install

medyagh (changes requested):
thank you for ur patience on this review, please take another look at the comment

Issue #14698, opened 2022-08-02
`--image-repository` unable to identify valid repository


vinnyzhaowenyu:
### What Happened?

When using a valid repository address, `--image-repository` will not work. For example: `https://harbor.cn-east-3.mydomain.com:443/google_containers/`

I find that the regular expression matching strategy is a problem

```
expression := regexp.MustCompile(`^(?:(\w+)\:\/\/)?([-a-zA-Z0-9]{1,}(?:\.[-a-zA-Z]{1,}){0,})(?:\:(\d+))?(\/.*)?$`)
```

The exact code location is as follows:
- [https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/start.go](https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/start.go)

```
// This function validates if the --image-repository
// args match the format of registry.cn-hangzhou.aliyuncs.com/google_containers
// also "<hostname>[:<port>]"
func validateImageRepository(imageRepo string) (validImageRepo string) {
	expression := regexp.MustCompile(`^(?:(\w+)\:\/\/)?([-a-zA-Z0-9]{1,}(?:\.[-a-zA-Z]{1,}){0,})(?:\:(\d+))?(\/.*)?$`)

	if strings.ToLower(imageRepo) == "auto" {
		imageRepo = "auto"
	}
```

So I took the following test

```golang
$ cat main.go 
package main

import (
  "fmt"
  "regexp"
)

func validateImageRepository(imageRepo string) {
  expression := regexp.MustCompile(`^(?:(\w+)\:\/\/)?([-a-zA-Z0-9]{1,}(?:\.[-a-zA-Z]{1,}){0,})(?:\:(\d+))?(\/.*)?$`)
  if expression.MatchString(imageRepo) {
    fmt.Println("+++++valid repo: " + imageRepo)
  } else {
    fmt.Println("---invalid repo: " + imageRepo)
  }
}

func main() {
    repo := make(map[int]string)
    repo[1] = "restry.cn-hangzhou.aliyuncs.com/google_containers"
    repo[2] = "https://registry.cn-hangzhou.aliyuncs.com/google_containers"
    repo[3] = "https://registry.cn-hangzhou.aliyuncs.com:443/google_containers"
    repo[4] = "https://registry.cn-hangzhou.aliyuncs.com:443/google_containers/"
    repo[5] = "harbor.mydomain.com/library"
[...]

PR #17496, opened 2023-10-26
fix(image-repository): allow subdomains with numbers

justinmchase:
My company blocks `docker.io` at the network level and there is no proxy solution. Instead they provide an Artifactory registry which mirrors all of docker.io.

Thus I am attempting to run:

```
minikube start --image-repository docker.repo1.uhc.com
```

Which fails with the error:
> E1025 20:07:14.125831   12982 start.go:1541] Provided repository is not a valid URL. Defaulting to "auto"

The URL is indeed valid and so I tracked the issue down to the line of code in this PR which appears to assume that only the bottom level subdomain may contain numbers, which is a bad assumption.

I used regexr.com to validate:

#### before
Screenshot 2023-10-25 at 8 20 37 PM

#### after
Screenshot 2023-10-25 at 8 20 46 PM

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1596,7 +1596,7 @@ func validateRegistryMirror() {
 // args match the format of registry.cn-hangzhou.aliyuncs.com/google_containers
 // also "<hostname>[:<port>]"
 func validateImageRepository(imageRepo string) (validImageRepo string) {
-	expression := regexp.MustCompile(`^(?:(\w+)\:\/\/)?([-a-zA-Z0-9]{1,}(?:\.[-a-zA-Z]{1,}){0,})(?:\:(\d+))?(\/.*)?$`)
+	expression := regexp.MustCompile(`^(?:(\w+)\:\/\/)?([-a-zA-Z0-9]{1,}(?:\.[-a-zA-Z0-9]{1,}){0,})(?:\:(\d+))?(\/.*)?$`)
 
 	if strings.ToLower(imageRepo) == "auto" {
 		imageRepo = "auto"
--- a/cmd/minikube/cmd/start_test.go
+++ b/cmd/minikube/cmd/start_test.go
@@ -384,6 +384,22 @@ func TestValidateImageRepository(t *testing.T) {
 			imageRepository:      "registry.test.com:6666/google_containers",
 			validImageRepository: "registry.test.com:6666/google_containers",
 		},
+		{
+			imageRepository:      "registry.1test.com:6666/google_containers",
+			validImageRepository: "registry.1test.com:6666/google_containers",
+		},
+		{
+			imageRepository:      "registry.t1est.com:6666/google_containers",
+			validImageRepository: "registry.t1est.com:6666/google_containers",
+		},
+		{
+			imageRepository:      "registry.test1.com:6666/google_containers",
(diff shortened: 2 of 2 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -1560,7 +1560,7 @@ func validateRegistryMirror() {
 // args match the format of registry.cn-hangzhou.aliyuncs.com/google_containers
 // also "<hostname>[:<port>]"
 func validateImageRepository(imageRepo string) (validImageRepo string) {
-	expression := regexp.MustCompile(`^(?:(\w+)\:\/\/)?([-a-zA-Z0-9]{1,}(?:\.[-a-zA-Z]{1,}){0,})(?:\:(\d+))?(\/.*)?$`)
+	expression := regexp.MustCompile(`^(?:(\w+)\:\/\/)?([-a-zA-Z0-9]{1,}(?:\.[-a-zA-Z0-9]{1,}){0,})(?:\:(\d+))?(\/.*)?$`)
```
medyagh, line 1563: @justinmchase thank you for this PR, how about we add a unit test for this function and add all the valid cases including the one you mentioned in the PR description

medyagh (changes requested):

cmd/minikube/cmd/start.go
justinmchase: Fortunately there already was a test with a list of test cases, I was able to easily just add the new domain to the list and it succeeded locally.

justinmchase:
Can we re-run the tests? I believe the requested change has been met.

justinmchase:
@medyagh can you re-review this one? I made the requested changes.

cmd/minikube/cmd/start.go
liangyuanpeng: Actually, it always match, also see https://github.com/kubernetes/minikube/issues/14698#issuecomment-1426707622

liangyuanpeng: In fact, I don't quite understand the role of this validateImageRepository. If the imageRepository I configure is wrong, then I just need it to fail to pull the image and know that I configured it correctly.

So + 1 from me.

liangyuanpeng:
I also need this PR. Please help to see what we are missing.

/cc @medyagh @spowelljr 
/kind bug

medyagh:
thank you @justinmchase for your patience, I saw there is also a newer PR 
https://github.com/kubernetes/minikube/pull/18290/files

but since your PR was first, if you include the test cases from the other PR int his PR we could merge yours!

this bug fix will be included in the next release

allenhaozi:
> thank you @justinmchase for your patience, I saw there is also a newer PR https://github.com/kubernetes/minikube/pull/18290/files
> 
> but since your PR was first, if you include the test cases from the other PR int his PR we could merge yours!
> 
> this bug fix will be included in the next release

I did not see this pr when I submitted. hi @justinmchase  iIf it is convenient for you to add more test cases
, I will close my pr 

/cc @liangyuanpeng @medyagh

justinmchase:
I'll get on that today then.

justinmchase:
I rebased my pr from upstream master and added the additional test cases from #18290.

justinmchase:
My company is currently evaluating alternatives to Docker desktop and this is blocking us from using minikube so I would definitely appreciate including this! Thank you.

spowelljr (approved):

PR #17552, opened 2023-11-04
German Translations

jklippel:
Added german translations to translation file.


spowelljr (approved):
Thanks for the translations!

PR #17585, opened 2023-11-07
Release v1.32.0

spowelljr:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,49 @@
 # Release Notes
 
+## Version 1.32.0 - 2023-11-07
+
+Features:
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,49 @@
 # Release Notes
 
+## Version 1.32.0 - 2023-11-07
+
+Features:
```
medyagh, line 5: lets add the Beta changs into the release notes for the 32 stable (but do not add it it to the changelog)
so only in release notes we have combined version of beta and stable release notes. so the users get the news about new features

spowelljr: Will do

PR #17601, opened 2023-11-11
Fix ingress-dns.md instructions

nyurik:
current instruction is incorrect because that file requires sudo, but `minikube ip` under sudo will not give anything meaningful


spowelljr (approved):
Thanks for the fix!

PR #17602, opened 2023-11-12
qemu2 default ssh user - point to host one to be able to ssh to colima without docker desktop

blind3dd:
When launching Colima with containerd and then trying to run Minikube with containerd and vmnet without docker desktop, the ssh fails to work when Minikube is about to start because username is hardcoded to docker. It cannot be edited or modified in config.json file, so in result Minikube hangs at 
`I1112 00:58:28.139890   46879 main.go:141] libmachine: Found match: a:71:bd:74:67:db
I1112 00:58:28.139897   46879 main.go:141] libmachine: IP: 127.0.0.1
I1112 00:58:28.139901   46879 main.go:141] libmachine: Waiting for VM to start (ssh -p 49590 **docker**@127.0.0.1)...`

setup: socket_vmnet / bridge
colima with lima and ForwardAgent Yes (ssh-add) and mount of ~/. to get *.sock files in place
vmnet command defined in Lima
colima with containerd
minikube with containerd
no docker installed

### DIFF
--- a/pkg/drivers/qemu/qemu.go
+++ b/pkg/drivers/qemu/qemu.go
@@ -18,6 +18,7 @@ package qemu
 
 import (
 	"archive/tar"
+	"bufio"
 	"bytes"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


afbjorklund:
It is not clear what this PR is trying to accomplish, typically the qemu driver would run the minikube.iso (with user "docker")

afbjorklund:
I commented in the issue.

PR #17659, opened 2023-11-22
Add journalctl logs to failed systemctl commands

spowelljr:
### Problem
Users (and our testing infra) occasionally run into flakey systemctl related failures when docker and cri-docker are being restarted. The error from systemctl doesn't include any useful information and outputs to run `journalctl -xeu <service>` to get the error logs. In our testing infra this is impossible as the clusters and already deleted when the logs complete, and due to the flakiness it's hard to reproduce a failure and users aren't able to generate use the logs. We don't know the cause of these errors so we can't further debug.

Example:
```
😄  minikube v1.31.2 on
💨  For improved Docker performance, enable the overlay Linux kernel module using 'modprobe overlay'
❗  docker is currently using the btrfs storage driver, setting preload=false
🚜  Pulling base image ...

❌  Exiting due to RUNTIME_ENABLE: Failed to enable container runtime: sudo systemctl restart docker: Process exited with status 1
stdout:

stderr:
Job for docker.service failed because the control process exited with error code.
See "systemctl status docker.service" and "journalctl -xeu docker.service" for details.


```

### Solution
I've wrapped the major systemctl commands with a function that checks if the output was successful. If it does there are no changes, but if the systemctl command did fail we call `sudo journalctl --no-pager -u <service>` to get the logs and then append them to the error so no extra work is needed to get the systemctl error logs.

### DIFF
--- a/pkg/minikube/sysinit/systemd.go
+++ b/pkg/minikube/sysinit/systemd.go
@@ -78,7 +78,7 @@ func (s *Systemd) Enable(svc string) error {
 		return errors.New("please don't enable kubelet as it creates a race condition; if it starts on systemd boot it will pick up /etc/hosts before we have time to configure /etc/hosts")
 	}
 	_, err := s.r.RunCmd(exec.Command("sudo", "systemctl", "enable", svc))
-	return err
+	return s.appendJournalctlLogsOnFailure(svc, err)
 }
 
 // EnableNow enables a service and then activates it too (not waiting for next start)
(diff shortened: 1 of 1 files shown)


medyagh (approved):

prezha:
@spowelljr i think this is a good call - at lest until we get to the bottom of these flakes
we should fix the unit tests, though, as they are [panic](https://github.com/kubernetes/minikube/actions/runs/6960935907/job/18941473690?pr=17659#step:6:71)-ing now b/c of the additional (unexpected) command (the `journalctl`)
i remember seeing some cr-related services giving up after three retries because the dependent service was not _yet_ up

Issue #17315, opened 2023-09-27
Failed to stop ssh-agent process: failed loading config: cluster "minikube" does not exist


spowelljr:
If you try deleting minikube with there's not a cluster you got the following error:
```
$ minikube delete
🙄  "minikube" profile does not exist, trying anyways.
❌  Failed to stop ssh-agent process: failed loading config: cluster "minikube" does not exist
💀  Removed all traces of the "minikube" cluster.
```

That error message should not be displayed, seems like we're trying to load a cluster that doesn't exist. If the cluster doesn't exist we shouldn't try loading it.

Songjf-ttk:
I will try to solve this issue

Songjf-ttk:
I think I can find the problem and fix it

tarishij17:
@spowelljr Is this issue fixed? Which minikube version is this cmd ran on? I'm not getting this error message in minikube version: v1.26.1

spowelljr:
@tarishij17 This issue is not fixed, it was introduced in v1.30 or v1.31

tarishij17:
PR for fixing this issue: https://github.com/kubernetes/minikube/pull/17713

Bharadwajshivam28:
@spowelljr Hey can i work on this?

mbelconis:
Any progress on this?  Ran into this issue myself today installing minikube on WSL2

roma2341:
Same on Ubuntu 22.04.4

PR #17713, opened 2023-12-02
suppress error msg incase of no cluster existence

tarishij17:
Fixes https://github.com/kubernetes/minikube/issues/17315

### DIFF
--- a/pkg/minikube/sshagent/sshagent.go
+++ b/pkg/minikube/sshagent/sshagent.go
@@ -106,6 +106,9 @@ func isRunning(cc *config.ClusterConfig) (bool, error) {
 func Stop(profile string) error {
 	cc, err := config.Load(profile)
 	if err != nil {
+		if strings.Contains(fmt.Sprint(err), fmt.Sprintf("cluster %q does not exist", profile)) {
+			return nil
+		}
 		return fmt.Errorf("failed loading config: %v", err)
 	}
 	running, err := isRunning(cc)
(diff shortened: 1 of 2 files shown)


dharmit (commented):
First timer here. Ignore me if I'm being naive. :)

pkg/minikube/sshagent/sshagent.go
```diff
@@ -106,6 +106,9 @@ func isRunning(cc *config.ClusterConfig) (bool, error) {
 func Stop(profile string) error {
 	cc, err := config.Load(profile)
 	if err != nil {
+		if strings.Contains(fmt.Sprint(err), fmt.Sprintf("cluster %q does not exist", profile)) {
+			return nil
+		}
 		return fmt.Errorf("failed loading config: %v", err)
 	}
```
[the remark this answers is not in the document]
spowelljr: You're both on the right track, @dharmit's usage of using `errors.Is(err, config.ErrNotExist)` is better than using the `string.Contains`. And @tarishij17 you have it right that if the cluster does not exist we should `return nil` as the profile doesn't exist so there's nothing to stop.

medyagh (approved):

pkg/minikube/sshagent/sshagent.go
```diff
@@ -106,6 +106,9 @@ func isRunning(cc *config.ClusterConfig) (bool, error) {
 func Stop(profile string) error {
 	cc, err := config.Load(profile)
 	if err != nil {
+		if config.IsNotExist(err) {
+			return nil
```
medyagh, line 110: ideally it would return an ErrorType here that the caller of "Stop" will check against it and decide what to do with it. but this is still acceptable

medyagh (approved):

pkg/minikube/sshagent/sshagent.go
spowelljr: Implemented

PR #17734, opened 2023-12-05
No style for logs delimiter output

### LINKED ISSUE #17733: minikube logs output is not  tail(1) format
The minikube logs  code states in a comment that it uses tail(1) format for multiple logs

https://github.com/kubernetes/minikube/blob/c56894f727f7c879d1f6fa2a61b0ec0b84ba6c52/pkg/minikube/logs/logs.go#L168

The [GNU tail invocation docs](https://www.gnu.org/software/coreutils/manual/html_node/tail-invocation.html ) specifies

> If more than one file is specified, tail prints a one-line header before the output for each file, consisting of:

```
==> file name <==
```

### Actual output

Logs have additional `* ` on the header line and around it:

```
minikube logs | head
* 
* ==> Audit <==
* |--------------|-----------------------|----------|---------|---------|---------------------|---------------------|
|   Command    |         Args          | Profile  |  User   | Version |     Start Time      |      End Time       |
|--------------|-----------------------|----------|---------|---------|---------------------|---------------------|
| start        | --alsologtostderr -v4 | minikube | pnasrat | v1.32.0 | 04 Dec 23 11:58 EST | 04 Dec 23 11:59 EST |
| help         |                       | minikube | pnasrat | v1.32.0 | 05 Dec 23 10:12 EST | 05 Dec 23 10:12 EST |
| update-check |                       | minikube | pnasrat | v1.32.0 | 05 Dec 23 10:38 EST | 05 Dec 23 10:38 EST |
|--------------|-----------------------|----------|---------|---------|---------------------|---------------------|
```

/kind bug
/assign

pnasrat:
Use None style for `minikube logs` output 

This change makes the logs output compatible with `tail(1)` removing leading `*` 

Fixes #17733

Tested using go run and inspecting logs output.

### DIFF
--- a/pkg/minikube/logs/logs.go
+++ b/pkg/minikube/logs/logs.go
@@ -184,9 +184,9 @@ func Output(r cruntime.Manager, bs bootstrapper.Bootstrapper, cfg config.Cluster
 	failed := []string{}
 	for i, name := range names {
 		if i > 0 {
-			out.Styled(style.Empty, "")
+			out.Styled(style.None, "")
 		}
-		out.Styled(style.Empty, "==> {{.name}} <==", out.V{"name": name})
+		out.Styled(style.None, "==> {{.name}} <==", out.V{"name": name})
 		var b bytes.Buffer
 		c := exec.Command("/bin/bash", "-c", cmds[name])
 		c.Stdout = &b
@@ -205,7 +205,7 @@ func Output(r cruntime.Manager, bs bootstrapper.Bootstrapper, cfg config.Cluster
 			klog.Errorf("failed to read output: %v", err)
 			failed = append(failed, name)
 		}
-		out.Styled(style.Empty, l)
+		out.Styled(style.None, l)
 	}
 
 	if len(failed) > 0 {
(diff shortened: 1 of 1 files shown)


pnasrat:
Stepping through this I am also wondering if the style output code needs to handle style.None or style.Empty without the low bullet explicitly as looking at it, it accesses the style config map which doesn't contain the none style 

https://github.com/kubernetes/minikube/blob/c56894f727f7c879d1f6fa2a61b0ec0b84ba6c52/pkg/minikube/out/out_style.go#L37

```go
	s, ok := style.Config[st]
	if !s.OmitNewline {
		format += "\n"
	}

	// Similar to CSS styles, if no style matches, output an unformatted string.
	if !ok || JSON {
		return format, s.Spinner
	}
```

medyagh (commented):
thank you for this PR @pnasrat would u mind pasting before after this PR output

pnasrat:
@medyagh so checking a bit further it might be happening when redirecting the output of `minikube logs` (eg with `>`  or. `| pbcopy` or whatever). and hitting the no color path of  so if you just type `minikube logs` on OS X and ubuntu you won't see it. It is also present with `minikube logs --file logs.txt` 

https://github.com/kubernetes/minikube/blob/f386f3f1e31af676c887fa171045cc8f530fe2a3/pkg/minikube/out/out_style.go#L47

However based upon downloading multiple users logs from issues from users this is definitely getting in the uploaded logs. 

But the below gist is is a diff of running minikube logs with and with out this at these versions and redirecting the output

```
minikube version && ./out/minikube version
minikube version: v1.32.0
commit: 8220a6eb95f0a4d75f7f2d7b14cef975f050512d
minikube version: v1.32.0
commit: 1d1c6f3c143e2d28fe63167ba90e3265538c6a3a
```

https://gist.github.com/pnasrat/b1482a82d01c4b34e1ecd12f895b096f

pnasrat:
Full logs for comparison:

[logs-before.txt](https://github.com/kubernetes/minikube/files/13591173/logs-before.txt)
[logs-after.txt](https://github.com/kubernetes/minikube/files/13591172/logs-after.txt)

spowelljr (approved):

PR #17779, opened 2023-12-13
Fix network retry check for subnet already in use for podman

thomasjm:
The code in the `tryCreateDockerNetwork` function (which is also used for Podman) tries to detect when network creation fails because the requested subnet is already in use, so that the caller can retry. It does this by checking for certain strings in the output of the `podman/docker network create` command. In the case of Podman, it fails to detect this correctly because the string Podman uses was not included.

(This manifested as the inability to create two Minikube clusters simultaneously on the same machine, because the second one would fail with an error about network creation.)

To see this, here's the error message when I use my machine's rootless `podman` (version 4.5.0) and try to create the same subnet twice:

```
λ podman network create --subnet 192.168.49.0/24
podman1
λ podman network create --subnet 192.168.49.0/24
Error: subnet 192.168.49.0/24 is already used on the host or by another config
```

This PR just adds an additional string to check in `tryCreateDockerNetwork`.

### DIFF
--- a/pkg/drivers/kic/oci/network_create.go
+++ b/pkg/drivers/kic/oci/network_create.go
@@ -156,6 +156,9 @@ func tryCreateDockerNetwork(ociBin string, subnet *network.Parameters, mtu int, 
 		if strings.Contains(rr.Output(), "is being used by a network interface") {
 			return nil, ErrNetworkGatewayTaken
 		}
+		if strings.Contains(rr.Output(), "is already used on the host or by another config") {
+			return nil, ErrNetworkGatewayTaken
+		}
 		return nil, fmt.Errorf("create %s network %s %s with gateway %s and MTU of %d: %w", ociBin, name, subnet.CIDR, subnet.Gateway, mtu, err)
 	}
 	return gateway, nil


pkg/drivers/kic/oci/network_create.go
```diff
@@ -156,6 +156,9 @@ func tryCreateDockerNetwork(ociBin string, subnet *network.Parameters, mtu int,
 		if strings.Contains(rr.Output(), "is being used by a network interface") {
 			return nil, ErrNetworkGatewayTaken
 		}
+		if strings.Contains(rr.Output(), "is already used on the host or by another config") {
```
samyuh, line 159: WYT of changing from `is already used on the host` to `is already in use on the host`?

thomasjm: Huh? This needs to match the output of Podman, it's not up to us to change.

medyagh:
@thomasjm thank you for improving the experience for the podman users !

btw our functional tests for podman has been failing for while (on github actions) I would love to see an extra eyes on that if you are interested, thank you for your contributions

PR #17801, opened 2023-12-15
Translate chainese doc zh-CN.json

mahmut-Abi:


mahmut-Abi:
/cc @prezha @spowelljr 
thanks

spowelljr (approved):
Thanks you for the translations @mahmut-Abi and the review @liangyuanpeng!

PR #17833, opened 2023-12-19
Correct typo from 'minukube' to 'minikube'

jongwooo:
Correct typo from 'minukube' to 'minikube'.


spowelljr (approved):
Thanks for the fix

PR #17842, opened 2023-12-20
Add translations for korean

jongwooo:
## Description

Add translations for korean.


spowelljr (approved):
Thanks for the translations!

PR #17861, opened 2023-12-26
CI: Fix auto updating registry addon to alpha or beta versions

spowelljr:


spowelljr:
Replaced with https://github.com/kubernetes/minikube/pull/17904

PR #17903, opened 2024-01-05
CI: Automate updating node version (for website)

spowelljr:


.github/workflows/update-node-version.yml
```diff
@@ -0,0 +1,48 @@
+name: "update-node-version"
```
medyagh, line 1: consider a more explicit name such as update-site-node-version

medyagh (changes requested):

.github/workflows/update-node-version.yml
spowelljr: Renamed

medyagh (approved):

PR #17905, opened 2024-01-06
docs: fixed broken nvidia link in Readme

MarcusDunn:
there is no https://minikube.sigs.k8s.io/docs/handbook/addons/nvidia/

so I changed to https://minikube.sigs.k8s.io/docs/tutorials/nvidia/ which seemed the closest.


spowelljr (approved):
Thanks for the fix!

PR #17866, opened 2023-12-28
addons: Fix auto-pause addon

spowelljr:
Reverting https://github.com/kubernetes/minikube/pull/11909 and https://github.com/kubernetes/minikube/pull/17070 as they both caused issues related to the auto-pause addon.

Starting with https://github.com/kubernetes/minikube/pull/17070, the PR added calling `viper.GetString(config.ProfileName)` in `cmd/auto-pause/auto-pause.go`. The problem is `cmd/auto-pause/auto-pause.go` is built and then ran inside the Kicbase/ISO, so  `mustload.Partial(profile)` always fails since we're inside the Kicbase/ISO and there is no `~/.minikube` folder, causing the the auto-pause binary to exit and breaking the addon and `kubectl` commands.

I'm not sure the exact cause of the failure related to https://github.com/kubernetes/minikube/pull/11909 (will revisit once addon is back in a working state) but after the pods are unpaused `kubectl` commands hang for a while until finally responding with `Unable to connect to the server: EOF`.

I also updated the auto-pause addon image since `v0.0.4` was only built for arm64 so other architectures fail with an exec error.

**Before:**
```
$ minikube start

$ minikube addons enable auto-pause

# kubectl commands fail
$ kubectl get pods -A
The connection to the server 192.168.49.2:32443 was refused - did you specify the right host or port?

# auto-pause-hook pod is exited
$ minikube ssh -- docker ps -a
CONTAINER ID   IMAGE                       COMMAND                  CREATED              STATUS                          PORTS     NAMES
cdad0268cc7d   a633c3ce5969                "/auto-pause-hook -t…"   21 seconds ago       Exited (1) 21 seconds ago                 k8s_webhook_env-inject-5d9dd8dcc4-wdxmj_auto-pause_187b3ee3-535b-415e-a280-efa86b5828c3_3

# pod logs
$ minikube ssh -- docker logs cdad0268cc7d
exec /auto-pause-hook: exec format error
```

**After updating auto-pause image:**
```
$ minikube start

$ minikube addons enable auto-pause

# kubectl commands fail
$ kubectl get pods -A
Unable to connect to the server: EOF

# auto-pause-hook pod is now running, but pods never pause
$ minikube ssh -- docker ps
CONTAINER ID   IMAGE                                 COMMAND                  CREATED         STATUS         PORTS     NAMES
[...]

PR #17070, opened 2023-08-16
configure auto-pause-interval addon

wzslr321:
fixes #10595

Make AutoPause interval configurable by both persistent config with addons and minikube start flag. Defaults to 1 minute.

Configurable addon
```
➜  minikube git:(auto-pause-time-config) ✗ ./out/minikube addons configure auto-pause-interval
-- Enter interval time of auto-pause-interval (in minutes): 3
✅  auto-pause-interval was successfully configured
➜  minikube git:(auto-pause-time-config) ✗ cat ~/.minikube/profiles/minikube/config.json | grep AutoPauseInterval
    "AutoPauseInterval": 180000000000
```

Configurable flag on `minikube start`
```
➜  minikube git:(auto-pause-time-config) ✗ ./out/minikube start --auto-pause-interval=2m0s
😄  minikube v1.31.1 on Darwin 13.0.1 (arm64)
🆕  Kubernetes 1.27.4 is now available. If you would like to upgrade, specify: --kubernetes-version=v1.27.4
✨  Using the docker driver based on existing profile
👍  Starting control plane node minikube in cluster minikube
🚜  Pulling base image ...
🔄  Restarting existing docker container for "minikube" ...
🐳  Preparing Kubernetes v1.27.3 on Docker 24.0.4 ...
🔗  Configuring bridge CNI (Container Networking Interface) ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: storage-provisioner, default-storageclass

❗  /usr/local/bin/kubectl is version 1.25.2, which may have incompatibilities with Kubernetes 1.27.3.
    ▪ Want kubectl v1.27.3? Try 'minikube kubectl -- get pods -A'
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default
```

PR #11909, opened 2021-07-06
Cleanup on ticker allocations.

vishjain:
fixes #10596

PR #17936, opened 2024-01-12
Addons auto-pause: Remove memory leak & add configurable interval

spowelljr:
What this PR accomplishes:
- Removes auto-pause memory leak related to multiple timers (https://github.com/kubernetes/minikube/issues/10596)
- Made auto-pause-interval configurable (https://github.com/kubernetes/minikube/issues/10595)

# Memory Leak
For details on the memory leak see the issue https://github.com/kubernetes/minikube/issues/10596.

A fix for the memory leak was originally implemented by @vishjain in https://github.com/kubernetes/minikube/pull/11909. The PR successful resolved the memory leak but introduced a bug where once the pods were paused `kubectl` commands failed with `Unable to connect to the server: EOF`, the PR was reverted in https://github.com/kubernetes/minikube/pull/17866.

I reimplemented the fix following @vishjain's main concept of using a [Ticker](https://pkg.go.dev/time#Ticker) in place of multiple [Timers](https://pkg.go.dev/time#Timer). Once the addon is enabled the ticker is started with a one minute interval, if any `kubectl` commands are executed before the one minute countdown concludes the ticker is reset and starts again. Once the one minute countdown concludes the ticker is indefinitely stopped and the pods are paused. If a `kubectl` command is executed while the pods are paused, the pods are unpaused, and then the ticker is reset and the cycle continues.

# Configurable Interval
A PR to implement a configurable interval for auto-pause was originally implemented by @wzslr321 in https://github.com/kubernetes/minikube/pull/17070. The PR had some issues related to loading configs that resulted in the auto-pause binary terminating and the addon not working, but all the logic around storing and configuring the interval value was well done and able to be reused. The PR was reverted in https://github.com/kubernetes/minikube/pull/17866 where more details on the config loading bug can be found.

For background, the auto-pause binary, which is where the ticker is, is its own separate binary from minikube. The binary exists inside the Kicbase/ISO itself and is ran via systemd. So in order the pass the interval value in I had to add a flag value to the auto-pause binary, and then when the auto-pause addon is enabled the interval flag is passed into the systemd unit file.
```
ExecStart=/bin/auto-pause --container-runtime={{.ContainerRuntime}} --interval={{.AutoPauseInterval}}
```
The interval can be configured in one of two ways, `--auto-pause-interval` during `minikube start` or `minikube addons configure auto-pause`. Both ways have validation to ensure a valid duration string is provided, and that the duration is greater than 0s, a negative or 0s duration will cause a panic in the Ticker.

# Follow up work
There's an issue with running `minikube addons configure auto-pause`, the value is successfully saved, but the value is not immediately updated in the auto-pause addon itself. The systemd unit file is updated, but then a `systemctl daemon-reload` & `systemctl restart auto-pause.service` both have to be run for the update to be reflected in the binary. Due to the already large size of this PR I'll leave it for a follow up.


spowelljr:
ok-to-build-iso

spowelljr:
ok-to-build-image

spowelljr:
ok-to-build-iso

PR #17937, opened 2024-01-12
Fix broken doc url

qlijin:
Fix broken doc url.
before:
```
https://github.com/kubernetes/enhancements/blob/master/keps/0001-kubernetes-enhancement-proposal-process.md
```

after:
```
https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/0000-kep-process/README.md
```


kundan2707 (commented):
/lgtm

spowelljr (approved):
Thanks for the fix!

PR #17944, opened 2024-01-12
bump legacyVersion to v1.26.0 from 2022-06-23

prezha:
move away from ancient minikube versions (ie, older than v1.26.0 that was released in june 2022, seven minor versions from the current one) for TestRunningBinaryUpgrade, TestStoppedBinaryUpgrade and TestMissingContainerUpgrade integration tests that otherwise consistently fail, providing no real value

### DIFF
--- a/test/integration/version_upgrade_test.go
+++ b/test/integration/version_upgrade_test.go
@@ -36,7 +36,12 @@ import (
 )
 
 func installRelease(version string) (f *os.File, err error) {
-	tf, err := os.CreateTemp("", fmt.Sprintf("minikube-%s.*.exe", version))
+	ext := ""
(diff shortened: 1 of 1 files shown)


spowelljr (approved):
Code looks good to me, just waiting for result of testing

afbjorklund:
I think there are cloud providers running older versions, and then there is the whole https://klts.io/ movement

prezha:
this is only for the _minikube_ version - for _kubernetes_, we still try to go back to v1.16:
https://github.com/kubernetes/minikube/blob/838ea0c7fb16de5a630512cb915c0e88a3dfeff2/pkg/minikube/constants/constants.go#L41-L42

PR #17949, opened 2024-01-14
treewide: Replace 'egrep' with 'grep -E'

travier:
Fixes the following warning:
```
egrep: warning: egrep is obsolescent; using grep -E
```


maximiliankolb:
I have reproduced the warning and fix in `Makefile` for at least compilation using `make` on Fedora 39 using "go1.21.6 linux/amd64".

spowelljr (approved):
I tried the PR on macOS and Linux and worked for me on both.

Thank you for the PR!

spowelljr (approved):

PR #17952, opened 2024-01-14
Add crun update script and update crun to 1.13

travier:
Makefile & hack: Add update-crun-version target & update script

---

crun: Update to 1.13

- Add a workaround for git-version.h build failure
- Also add hash for version 1.12

---

I've verified that this builds. Not fully tested yet.


travier:
Fixed build and updated to 1.13. Build tested only for now.

travier:
Included in https://github.com/kubernetes/minikube/pull/18020

PR #17989, opened 2024-01-19
Fix misleading error messages in TestCopyBinary

ph-ngn:
The error messages under func TestCopyBinary(t *testing.T) in cache_binaries_test.go for unexpected and expected error cases are mixed up.


spowelljr (approved):
Thanks for the fix!

PR #17991, opened 2024-01-19
minikube-iso: set CONFIG_QFMT_V2=y for aarch64

zjx20:
If the option `CONFIG_QFMT_V2=y` is not set, the kernel prompts an error when attempting to use the project quota feature of ext4 in a minikube VM or minikube k8s cluster:

```
[ 9289.548970] EXT4-fs (loop0): The kernel was not built with CONFIG_QUOTA and CONFIG_QFMT_V2
```

The reproduction steps are as follows:

```bash
minikube start --driver=qemu2
minikube ssh

# inside the minikube shell
sudo bash -c su
cd /root
touch loop.img
truncate -s 1g loop.img
losetup -fP loop.img
devname=$(losetup -a | grep /root/loop.img | awk -F":" '{print $1}')
echo $devname  # <-- should be /dev/loop0 or something similar
mkfs.ext4 -O project,quota $devname
mkdir -p /root/mountpoint
mount -o prjquota $devname /root/mountpoint  # <-- ERROR, cannot mount the device with the prjquota option
```

image

Then check `dmesg` and you will see the error message mentioned earlier:

image

Please also notice that this option has been enabled in the x86_64 image: https://github.com/kubernetes/minikube/blob/5af9a51aed18f68dc78b7ccc71d356cbc12fa7c6/deploy/iso/minikube-iso/board/minikube/x86_64/linux_x86_64_defconfig#L491


spowelljr:
ok-to-build-iso

spowelljr (approved):

PR #18080, opened 2024-02-02
Fix the links

zdxgs:
The link has been changed.
golang.org—>pkg.go.dev


spowelljr:
Hi @zdxgs, thank you for the PR! Unfortunately this file is generated, so your change would be lost the next time we generate this file, so it doesn't make sense for us to modify the file. For this reason I'm going to close this PR, but thank you for your contribution and please feel free to pick up other issues!

PR #18107, opened 2024-02-05
CI: Fix ingress auto update not getting newest version

spowelljr:
Related to https://github.com/kubernetes/minikube/pull/18098

Before we were just pulling the first `controller-v.*` tag, but now we iterate through the list of tags and store the latest one we find.

### DIFF
--- a/hack/update/ingress_version/update_ingress_version.go
+++ b/hack/update/ingress_version/update_ingress_version.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/google/go-github/v58/github"
+	"golang.org/x/mod/semver"
 	"k8s.io/klog/v2"
 
 	"k8s.io/minikube/hack/update"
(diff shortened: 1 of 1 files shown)


hack/update/ingress_version/update_ingress_version.go
```diff
@@ -84,7 +85,9 @@ func main() {
 }
 
 func LatestControllerTag(ctx context.Context) (string, error) {
```
medyagh, line 87: if this is a Docker Hub Specific Func, maybe it could be re-used across all automations ? hopefully this code is not duplicated?

spowelljr: We have a common version that the rest of the scripts use, however we need a one-off for this one because their tags follow the format `controller-vX.X.X`. Our common func checks if the tag matches semver and ignores ones that don't.

Issue #17614, opened 2023-11-13
Masking env passwords in start does not work when password includes encoded characters


dennyferra:
The pull request #17116 included changes to mask proxy passwords. However, the regex used in `pkg/minikube/node/start.go` (https://github.com/kubernetes/minikube/blob/master/pkg/minikube/node/start.go#L748) uses `\w+`. This will not match passwords with special characters (like @ = %40) that have been URL encoded, for example: `http://user:fake%21password@someproxy.com:80`

Example (proxy credentials faked):

```
😄  minikube v1.32.0 on Microsoft Windows 10 Enterprise 10.0.19045.3570 Build 19045.3570
✨  Automatically selected the hyperv driver
💿  Downloading VM boot image ...
    > minikube-v1.32.1-amd64.iso....:  65 B / 65 B [---------] 100.00% ? p/s 0s
    > minikube-v1.32.1-amd64.iso:  292.96 MiB / 292.96 MiB  100.00% 42.57 MiB p
👍  Starting control plane node minikube in cluster minikube
💾  Downloading Kubernetes v1.28.3 preload ...
    > preloaded-images-k8s-v18-v1...:  403.35 MiB / 403.35 MiB  100.00% 67.17 M
🔥  Creating hyperv VM (CPUs=4, Memory=8192MB, Disk=20000MB) ...
🌐  Found network options:
    ▪ HTTP_PROXY=http://fakeuser:fake%21pass@myproy.com:80
    ▪ HTTPS_PROXY=http://fakeuser:fake%21pass@myproy.com:80
🐳  Preparing Kubernetes v1.28.3 on Docker 24.0.7 ...
    ▪ env HTTP_PROXY=http://fakeuser:fake%21pass@myproxy.com:80
    ▪ env HTTPS_PROXY=http://fakeuser:fake%21pass@myproy.com:80
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔗  Configuring bridge CNI (Container Networking Interface) ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: storage-provisioner, default-storageclass
💡  kubectl not found. If you need it, try: 'minikube kubectl -- get pods -A'
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default
```

alessandrocapanna:
[...]

PR #18158, opened 2024-02-13
fix: updated regex on validateNetwork to support special characters.

alessandrocapanna:
Fix issue #17614 

Now the regex support special characters as mentioned on issue.

Test made with faked credential: "http://user:fake%21password@someproxy.com:80"

Before:
Schermata 2024-02-13 alle 01 16 07

After:

Schermata 2024-02-13 alle 00 58 13


alessandrocapanna:
fixed also if password have the '@'

spowelljr:
I agree with @Zhongyi-Lu, using his example string `http://jdoe:%n0tRe@alPassword!@minikube.sigs.k8s.io:8080`

Using `//(\w+):.*?@` matches `//jdoe:%n0tRe@`

While `//(\w+):.*@` matches `//jdoe:%n0tRe@alPassword!@`

alessandrocapanna:
Yeah... I also agree with @Zhongyi-Lu, I will modify the pull request, I will also insert strings like those suggested in the tests :D

spowelljr:
Hi @alessandrocapanna, I just realized we have a function somewhere else in our code that does this as well.

`maskProxyPassword` in `pkg/minikube/node/config.go`

Could you make the function pulic and move it to `pkg/util`, then update any refernces to, and then we can make the code you modified to use the function as well. There are also a test for the function in `pkg/minikube/node/config_test.go` you could move as well and add additional test cases if needed. Feel free to chance the function if it's not worked as expected.

alessandrocapanna:
Hi @spowelljr.
I've completed the refactor you suggested. I split the maskProxyPassword function into two separate functions (under `pkg/util`) and made the necessary corrections. Additionally, I updated the references accordingly and added tests for both functions.

spowelljr (approved):
Thanks for the PR!

PR #18233, opened 2024-02-22
calico: support kubeadm.pod-network-cidr

lollipopman:
Prior to this commit the Calico CNI ignored the setting of kubeadm.pod-network-cidr, because we set CALICO_IPV4POOL_CIDR to the DefaultPodCIDR. However, this is no longer necessary as Calico's pod cidr will be set to the default when we initialize kubeadm in pkg/minikube/bootstrapper/bsutil/kubeadm.go.

### DIFF
--- a/hack/update/calico_version/update_calico_version.go
+++ b/hack/update/calico_version/update_calico_version.go
@@ -72,9 +72,6 @@ func updateYAML(version string) {
 		`docker\.io\/calico\/cni:.*`:              "{{ .BinaryImageName }}",
 		`docker\.io\/calico\/node:.*`:             "{{ .DaemonSetImageName }}",
 		`docker\.io\/calico\/kube-controllers:.*`: "{{ .DeploymentImageName }}",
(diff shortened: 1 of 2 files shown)


medyagh:
@lollipopman thank you for this PR, can you please provide a before/after this PR with minikube output with an example?

medyagh:
@lollipopman can you plz take a look at the unit tests ? and also could you plz put an output of Before / After this PR of using this new feature using minikube in action?

medyagh:
thanks @lollipopman do you mind pull upstream to your PR, the unit test is fixed, but there are failed test on Docker Linux that I would like to see if it goes away 
https://storage.googleapis.com/minikube-builds/logs/18233/33262/Docker_Linux.html

medyagh (approved):
after the test are ok we could merge

lollipopman:
@prezha & @spowelljr can you take a look at this pull request when you have a moment?

lollipopman:
woohoo, thanks!

PR #18264, opened 2024-02-28
testing: Update oldest supported K8s version to v1.20

spowelljr:
Bumped `OldestKubernetesVersion` from `v1.16.0` to `v1.20.0`

Also deleted legacy ingress addon tests as requires K8s < v1.19


travier (commented):
LGTM

medyagh (approved):

medyagh:
Long due change ! minikube offically supports past 5 versions but this makes it past 9 versions which is very impressive

PR #18292, opened 2024-03-04
Path Changes in Documentation for Minikube Installation in Windows

### LINKED ISSUE #18291: minikube start (path specification in documentation)
I used the Windows installer (latest release) to install minikube as asked to do so in Step 1.
The installer doesn't have any option to set the path for minikube.exe in itself so I tried to use the command given in the documentation of step 2 under installation but it didnt work. 

image

I realised that's because the path given in the command ('C:\minikube') is not where the default installation is prompted to be done in by the windows installer.

The windows installer prompts you to install it in this path by default:
image

And, correct me if I'm wrong, but I think that's because of this code in the minikube.nsi file in the installers/windows directory of the github repository:
image

I think the documentation should be updated to propose the installation path as 'C:\Program Files\Kubernetes\Minikube' instead of 'C:\minikube' just to maintain uniformity in both the methods you have for installation in Windows and also so that setting the path actually works when someone installs minikube through the installer and then sets the path with the command you have given. It would help new users that are just starting to figure out Kubernetes and Minikube and don't have a lot knowledge about setting paths in Windows etc.

abhaymathur21:
Fixes #18291 

Changes the installation path given below from 'c:\minekube' to 'C:\Program Files\Minekube' because thats where the Windows installer installs it
image

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -442,8 +442,9 @@ choco install minikube
 <br>
     Or if using `PowerShell`, use this command:
     ```powershell
-    New-Item -Path 'c:\' -Name 'minikube' -ItemType Directory -Force
-    Invoke-WebRequest -OutFile 'c:\minikube\minikube.exe' -Uri 'https://github.com/kubernetes/minikube/releases/latest/download/minikube-windows-amd64.exe' -UseBasicParsing
+    New-Item -Path 'C:\Program Files' -Name 'Kubernetes' -ItemType Directory -Force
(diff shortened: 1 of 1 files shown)


xcarolan:
Hello @abhaymathur21, thank you for your contribution.
The minikube installer for Windows puts the executable in c:\minikube as expected from the -Outfile parameter. I don't think this change adds any real value. Is it causing problems?

xcarolan:
Hello @abhaymathur21, thank you for your contribution.
The minikube installer for Windows puts the executable in c:\minikube as expected from the -Outfile parameter. I don't think this change adds any real value.

PR #18346, opened 2024-03-10
ha tutorial

prezha:
add a tutorial for the ha - multi-control plane

bonus: reduce spam error messages for not having optional ip_vs kernel modules available


medyagh (changes requested):
Lets separate the Site from the Golang code, I would like to merge the Site after the release so the users dont be confused for trying the doc before the code is released

prezha:
> Lets separate the Site from the Golang code, I would like to merge the Site after the release so the users dont be confused for trying the doc before the code is released

done!

medyagh (approved):

Issue #18359, opened 2024-03-11
Add matching Kubernetes Version for crictl to the preload


medyagh:
### What Happened?

starting kubernetes v1.29 we need to have matching crictl and cni plugin

First we need extact preload
so we first check if crictl and kubernets version are same
then nothing
if not download and install matching crictl on the fly

so also need to add the crictl to the preload generation scriipt

### Attach the log file

n/a

### Operating System

None

### Driver

None

afbjorklund:
This needs to be added / moved to the bootstrapper, so that it also works for `--preload=false`.

But it is supposed to be able to install them, similar to how it installs kubeadm/kubelet/kubectl today.

afbjorklund:
https://github.com/kubernetes/release/blob/master/cmd/krel/templates/latest/metadata.yaml

k8s | cri | cni
--- | --- | ---
1.26.0 | >= 1.25.0 | >= 1.1.1
1.27.0 | >= 1.25.0 | >= 1.1.1
1.28.0 | >= 1.28.0 | >= 1.1.1
1.29.0 | >= 1.28.0 | >= 1.2.0
1.30 | >= 1.30.0 | >= 1.2.0

afbjorklund:
* https://github.com/kubernetes/minikube/issues/15495

PR #18294, opened 2024-03-04
kicbase: Fix install for cri-o

spowelljr:
**Problem:**
HEAD is failing with the following when trying to build kicbase
```
02:59:09  > [linux/amd64 kicbase 23/47] RUN export ARCH=$(dpkg --print-architecture | sed 's/ppc64el/ppc64le/' | sed 's/armhf/arm-v7/') &&     if [ "$ARCH" != "ppc64le" ] && [ "$ARCH" != "arm-v7" ]; then sh -c "echo 'deb https://downloadcontent.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable:/cri-o:/1.29/xUbuntu_22.04/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable:cri-o:1.29.list" &&     curl -LO https://downloadcontent.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable:/cri-o:/1.29/xUbuntu_22.04/Release.key &&     apt-key add - < Release.key &&     clean-install cri-o cri-o-runc; fi:
02:59:09 
100   146  100   146    0     0    334      0 --:--:-- --:--:-- --:--:--   335
02:59:09 0.736 Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
02:59:09 0.778 gpg: no valid OpenPGP data found.
```

**Cause:**
They moved cri-o 1.29+ to a new deb repo

cri-o < 1.29: `https://downloadcontent.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable:/cri-o:/${CRIO_VERSION}/xUbuntu_22.04/`

cri-o 1.29+: `https://pkgs.k8s.io/addons:/cri-o:/stable:/${CRIO_VERSION}/deb/`

**Update crictl repo**
We were previously using the old repo for crictl `https://downloadcontent.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_22.04/`

This repo only has crictl v1.25, so there's a version mismatch from our ISO.

Updated the use the new crictl repo `https://pkgs.k8s.io/core:/stable:/${CRICTL_VERSION}/deb/`

Also pin the crictl version in the kicbase now

PR #18367, opened 2024-03-12
Kicbase: Revert cri-o from v1.29 to v1.24

spowelljr:
As discussed during office hours and as evidenced in https://github.com/kubernetes/minikube/pull/18294, using newer crictl and cri-o versions results in incompatibility with previous Kubernetes versions. Rolling back until we implement a mechanism to use the matching cri-o and crictl versions to the chosen Kubernetes version (https://github.com/kubernetes/minikube/issues/18359).


spowelljr:
ok-to-build-image

PR #18421, opened 2024-03-17
add default sysctls to allow privileged ports with no capabilities

prezha:
this pr adds `net.ipv4.ip_unprivileged_port_start` sysctls as default to `containerd` and `cri-o` container runtimes to allow privileged ports to be used by containers run by a non-root user for `kubernetes v1.22+`

### background

during testing the bump of kubernetes default version to v1.29.2, we discovered that `coredns` would not come up if containerd or cri-o container runtimes are used with the docker driver (example: [conainerd](https://storage.googleapis.com/minikube-builds/logs/17786/33572/Docker_Linux_containerd.html) and [cri-o](https://storage.googleapis.com/minikube-builds/logs/17786/33572/Docker_Linux_crio.html))

the coredns errored with: `Listen: listen tcp :53: bind: permission denied`

kubernetes v1.29.0+ [uses](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.29.md#other-cleanup-or-flake-6) coredns v1.11.1 and from [v1.11.0](https://github.com/coredns/coredns/releases/tag/v1.11.0), coredns [runs as non-root user ](https://github.com/coredns/coredns/pull/5969)

some additional details about the issue also observed elsewhere can be found eg, [here](https://github.com/coredns/coredns/issues/6249) and [here](https://github.com/coredns/coredns/pull/6320)

since starting from kubernetes v1.22, the required net.ipv4.ip_unprivileged_port_start sysctl was [marked as safe](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md#feature-9), but it's up to the container runtime to implement it - [docker](https://github.com/moby/moby/pull/41030) already has it enabled by default, whereas other CRIs have that as option but it's not enabled by default ([containerd](https://github.com/containerd/containerd/pull/6170) and [cri-o](https://github.com/cri-o/cri-o/pull/1721)) and this pr tries to address those two and enable corresponding options

### DIFF
--- a/pkg/minikube/cruntime/crio.go
+++ b/pkg/minikube/cruntime/crio.go
@@ -90,7 +90,7 @@ func generateCRIOConfig(cr CommandRunner, imageRepository string, kv semver.Vers
 	}
 
 	// add 'net.ipv4.ip_unprivileged_port_start=0' sysctl so that containers that run with non-root user can bind to otherwise privilege ports (like coredns v1.11.0+)
-	// note: 'net.ipv4.ip_unprivileged_port_start' sysctl was marked as safe since kubernetes v1.22 (Aug 4, 2021) (ref: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md#feature-9)
+	// note: 'net.ipv4.ip_unprivileged_port_start' sysctl was marked as safe since Kubernetes v1.22 (Aug 4, 2021) (ref: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md#feature-9)
 	// note: cri-o supports 'default_sysctls' option since v1.12.0 (Oct 19, 2018) (ref: https://github.com/cri-o/cri-o/releases/tag/v1.12.0; https://github.com/cri-o/cri-o/pull/1721)
 	if kv.GTE(semver.Version{Major: 1, Minor: 22}) {
 		// remove any existing 'net.ipv4.ip_unprivileged_port_start' settings
--- a/pkg/minikube/cruntime/crio.go
+++ b/pkg/minikube/cruntime/crio.go
@@ -89,6 +89,24 @@ func generateCRIOConfig(cr CommandRunner, imageRepository string, kv semver.Vers
 		klog.Warningf("unable to remove /etc/cni/net.mk directory: %v", err)
 	}
 
+	// add 'net.ipv4.ip_unprivileged_port_start=0' sysctl so that containers that run with non-root user can bind to otherwise privilege ports (like coredns v1.11.0+)
+	// note: 'net.ipv4.ip_unprivileged_port_start' sysctl was marked as safe since kubernetes v1.22 (Aug 4, 2021) (ref: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md#feature-9)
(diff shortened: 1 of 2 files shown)


afbjorklund:
Seems like this is something that should at least be documented upstream, for changing the runtime settings...

afbjorklund:
Releated issue, about adding non-default config to kind:

* https://github.com/kubernetes-sigs/kind/pull/2748

prezha:
there's a notion to go even further - by k8s _requiring_ this from cri: https://github.com/kubernetes/kubernetes/issues/102612#issue-911968751

also, containerd plans to have it enabled by default in v2.0: https://github.com/containerd/containerd/issues/6924

afbjorklund:
I don't think that is a big issue, just wondering if minikube should follow rather than lead?

pkg/minikube/cruntime/crio.go
```diff
@@ -89,6 +89,24 @@ func generateCRIOConfig(cr CommandRunner, imageRepository string, kv semver.Vers
 		klog.Warningf("unable to remove /etc/cni/net.mk directory: %v", err)
 	}
 
+	// add 'net.ipv4.ip_unprivileged_port_start=0' sysctl so that containers that run with non-root user can bind to otherwise privilege ports (like coredns v1.11.0+)
+	// note: 'net.ipv4.ip_unprivileged_port_start' sysctl was marked as safe since kubernetes v1.22 (Aug 4, 2021) (ref: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md#feature-9)
+	// note: cri-o supports 'efault_sysctls' option since v1.12.0 (Oct 19, 2018) (ref: https://github.com/cri-o/cri-o/releases/tag/v1.12.0; https://github.com/cri-o/cri-o/pull/1721)
```
spowelljr, line 94:
```suggestion
	// note: cri-o supports 'default_sysctls' option since v1.12.0 (Oct 19, 2018) (ref: https://github.com/cri-o/cri-o/releases/tag/v1.12.0; https://github.com/cri-o/cri-o/pull/1721)
```

pkg/minikube/cruntime/crio.go
```diff
@@ -89,6 +89,24 @@ func generateCRIOConfig(cr CommandRunner, imageRepository string, kv semver.Vers
 		klog.Warningf("unable to remove /etc/cni/net.mk directory: %v", err)
 	}
 
+	// add 'net.ipv4.ip_unprivileged_port_start=0' sysctl so that containers that run with non-root user can bind to otherwise privilege ports (like coredns v1.11.0+)
+	// note: 'net.ipv4.ip_unprivileged_port_start' sysctl was marked as safe since kubernetes v1.22 (Aug 4, 2021) (ref: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md#feature-9)
```
spowelljr, line 93:
```suggestion
	// note: 'net.ipv4.ip_unprivileged_port_start' sysctl was marked as safe since Kubernetes v1.22 (Aug 4, 2021) (ref: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md#feature-9)
```

spowelljr (changes requested):
Just two super small things in the comments

prezha:
eagle eye 🙂
thanks @spowelljr
done!

spowelljr (approved):

PR #18434, opened 2024-03-18
CI: Auto update crun

spowelljr:


medyagh (commented):
does this need to to be added to do it ALL Job ?

medyagh (approved):

spowelljr:
> does this need to to be added to do it ALL Job ?

It is, that's the changes to `.github/workflows/update-iso-image-versions.yml` are

PR #18492, opened 2024-03-22
chore: fix function names

depthlending:

### DIFF
--- a/pkg/minikube/tunnel/kic/ssh_tunnel.go
+++ b/pkg/minikube/tunnel/kic/ssh_tunnel.go
@@ -187,7 +187,7 @@ func (t *SSHTunnel) stopMarkedConnections() {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/minikube/tunnel/kic/ssh_tunnel.go
```diff
@@ -187,7 +187,7 @@ func (t *SSHTunnel) stopMarkedConnections() {
 	}
 }
 
-// sshConnName creates a uniq name for the tunnel, using its name/clusterIP/ports.
+// sshConnUniqName creates an uniq name for the tunnel, using its name/clusterIP/ports.
```
spowelljr, line 190:
```suggestion
// sshConnUniqName creates a uniq name for the tunnel, using its name/clusterIP/ports.
```
Should stay as `a` since the `u` in `unique` has a `you` sound

spowelljr (changes requested):

spowelljr (approved):

PR #18493, opened 2024-03-23
site: VirtualBox driver NIC type alternative for performance enhancement

### LINKED ISSUE #7031: document work around for VirtualBox networking be slow
Hi,
we are using minikube to deploy local test installations of our stack, in particular for developers. In general, we are using the virtual box hypervisor (yes, even on win 10 machines, for *reasons* ;-)).

We note a general sluggishness of the minikube environment, even when running on very powerful host machines to which we assign lots of cores (<< #physical cores on the host, naturally) and RAM to the minikube VM with the "minikube start" command.
We are still in the early phase of pinpointing the bottlenecks, but one observation that we already made is that network performance seems to be a problem and it seems unrelated to our own software, because we can also observe it when doing something trivial like a docker build:

When building one of our images where we download a number of larger artifacts from our artifact repo (Nexus), the build takes ~8 secs on the Docker Toolbox "default" VM. The same build, on the same host machine, takes almost a minute on the minikube VM (which BTW has 6 cores vs. 2 for the docker toolbox VM). The main difference comes from all operations in our RUN commands that involve the network. For example, the download of a ~90MB artifact with curl command reaches ~70MB/s on Docker Toolbox, while it barely reaches 2,5MB/s with the minikube VM. 
The issue is the same when putting a number of large files into the Docker build context and letting Docker upload them.

Is this a known issue? Any ideas where to start looking?

JG

skoenig:
fixes #7031

### DIFF
--- a/site/content/en/docs/drivers/virtualbox.md
+++ b/site/content/en/docs/drivers/virtualbox.md
@@ -26,3 +26,8 @@ minikube start supports some VirtualBox specific flags:
 ## Troubleshooting
 
 * Run `minikube start --alsologtostderr -v=7` to debug crashes
+* If you experience slow network performance with the VirtualBox driver, changing the Network Interface Card (NIC) type may improve speed. Use the following command to start minikube with the AMD PCNet FAST III (Am79C973) for both NAT and host-only network interfaces:
+
+    ```shell
+    minikube start --vm-driver=virtualbox --nat-nic-type=Am79C973 --host-only-nic-type=Am79C973
+    ```


medyagh (approved):

medyagh:
@[skoenig](https://github.com/skoenig) thank you for this contribution :) hopefully this helps more users. btw I am curious are you using vbox on linux or macos ?

and have do you have any insight to this issue if you can help with it ? https://github.com/kubernetes/minikube/issues/15274

skoenig:
Hey @medyagh I am using Linux, so not sure how I could help with the issue you mentioned

PR #18514, opened 2024-03-26
Site: Disable Beta buttons in "Get started!" documentation if the release is older than the latest Stable release.

### LINKED ISSUE #18505: Minikube website should not show Beta Releases to download if they are Older than the Stable release
### What Happened?

in minikube wegsite 
we have a Beta Button that users can choose Beta releaes. 

it should NOT show anything or be grayed out if our latest Beta release is actually older than our stable release
Screenshot 2024-03-25 at 11 19 24 AM

### Attach the log file

/NA

### Operating System

None

### Driver

None

Sryther:
Disabling Beta buttons is not instantaneous. I'm not sure that's the most convenient method to do it.

First it looks for latest release, if it's not a Beta one it'll disable the Beta button and activate the Stable one instead.

fixes #18505

### DIFF
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -38,21 +38,11 @@ All you need is Docker (or similarly compatible) container or a Virtual Machine 
       if (releases && releases.length > 0 && releases[0] && releases[0].tag_name) {
         const isBetaMostRecent = releases[0].tag_name.includes("-beta");
 
-        if (!isBetaMostRecent) {
+        if (isBetaMostRecent) {
           for (architecture of architectures) {
             const betaElement = document.querySelector(`button[data-quiz-id="/${architecture}/Beta"]`);
-            const stableElement = document.querySelector(`button[data-quiz-id="/${architecture}/Stable"]`);
             if (betaElement) {
-              betaElement.disabled = true;
-
-              if (betaElement.classList.contains("active")) {
-                const stableElement = document.querySelector(`button[data-quiz-id="/${architecture}/Stable"]`);
-                if (stableElement) {
-                  stableElement.click();
-                  stableElement.classList.add("active");
-                }
-                betaElement.classList.remove("active");
-              }
+              betaElement.classList.remove("hide");
             }
           }
         }
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -38,7 +38,7 @@ All you need is Docker (or similarly compatible) container or a Virtual Machine 
       if (releases && releases.length > 0 && releases[0] && releases[0].tag_name) {
         const isBetaMostRecent = releases[0].tag_name.includes("-beta");
 
-        if (isBetaMostRecent) {
+        if (!isBetaMostRecent) {
           for (architecture of architectures) {
             const betaElement = document.querySelector(`button[data-quiz-id="/${architecture}/Beta"]`);
             const stableElement = document.querySelector(`button[data-quiz-id="/${architecture}/Stable"]`);
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -74,7 +74,7 @@ Click on the buttons that describe your target platform. For other architectures
 {{% /quiz_row %}}
 
 {{% quiz_row base="/Linux/x86-64" name="Release type" %}}
-{{% quiz_button option="Stable" %}} {{% quiz_button option="Beta" %}}
+{{% quiz_button option="Stable" %}} {{% quiz_button option="Beta" hide="true" %}}
 {{% /quiz_row %}}
 
 {{% quiz_row base="/Linux/x86-64/Stable" name="Installer type" %}}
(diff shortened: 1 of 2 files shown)


medyagh (changes requested):
hi @Sryther thanks for the quick PR ! however in the prview I see it is disabled

https://deploy-preview-18514--kubernetes-sigs-minikube.netlify.app/docs/start/

even though right now minikube does have a new beta release, (it should have NOT been grayed out)

Sryther:
> even though right now minikube does have a new beta release, (it should have NOT been grayed out)

You're totally right, I just forgot to invert the condition while I was testing...

spowelljr:
Hi @Sryther, thanks for your PR! I took a look at this and think we can simplify this. First off, 99% of the time we don't want to show the `Beta` button, so I think it's better if we start with it completely hidden and show it if there's a beta available. We have an existing `hide` class we can use, and for minimal code changes, we can check if `option == "Beta"` and if it is add the `hide` class by default (shown below).

```diff
diff --git a/site/layouts/shortcodes/quiz_button.html b/site/layouts/shortcodes/quiz_button.html
index 1100d1e92..3cf0b6433 100644
--- a/site/layouts/shortcodes/quiz_button.html
+++ b/site/layouts/shortcodes/quiz_button.html
@@ -1 +1 @@
-<button data-quiz-id="{{ .Parent.Get "base" }}/{{ .Get "option" }}" type="button" class="btn btn-outline-primary option-button">{{ .Get "option" }}</button>
+<button data-quiz-id="{{ .Parent.Get "base" }}/{{ .Get "option" }}" type="button" class="btn btn-outline-primary option-button{{ if eq (.Get "option") "Beta" }} hide{{ end }}">{{ .Get "option" }}</button>
```

Then this will simplify a lot of your code because you can remove your logic that changes the active button to `Stable` since the button won't be available if there's no beta, and if there is a beta we simply remove the `hide` class.

```diff
diff --git a/site/content/en/docs/start/_index.md b/site/content/en/docs/start/_index.md
index 71b0a86bc..569f9b0cf 100644
--- a/site/content/en/docs/start/_index.md
+++ b/site/content/en/docs/start/_index.md
@@ -38,21 +38,11 @@ All you need is Docker (or similarly compatible) container or a Virtual Machine
       if (releases && releases.length > 0 && releases[0] && releases[0].tag_name) {
         const isBetaMostRecent = releases[0].tag_name.includes("-beta");
 
-        if (!isBetaMostRecent) {
+        if (isBetaMostRecent) {
           for (architecture of architectures) {
             const betaElement = document.querySelector(`button[data-quiz-id="/${architecture}/Beta"]`);
-            const stableElement = document.querySelector(`button[data-quiz-id="/${architecture}/Stable"]`);
             if (betaElement) {
-              betaElement.disabled = true;
-
-              if (betaElement.classList.contains("active")) {
-                const stableElement = document.querySelector(`button[data-quiz-id="/${architecture}/Stable"]`);
-                if (stableElement) {
-                  stableElement.click();
-                  stableElement.classList.add("active");
-                }
-                betaElement.classList.remove("active");
-              }
+              betaElement.classList.remove("hide");
             }
           }
         }
```

Sryther:
I wasn't aware of the `.hide` class! Your proposal is clearly better with the arguments you're pointing at.

spowelljr:
Yeah, that's completely fine, I just used `Beta` to set `hide` because this component was specifically created for this so it's unlikely to be used anywhere else and would save from having to manually add `hide="true"` to all the beta ones. But making it generic is great too

site/content/en/docs/start/_index.md
```diff
@@ -20,6 +20,47 @@ All you need is Docker (or similarly compatible) container or a Virtual Machine
...
+    "Linux/ARM64",
+    "Linux/ppc64",
+    "Linux/S390x",
+    "Linux/ARMv7",
+    "macOS/x86-64",
+    "macOS/ARM64",
+    "Windows/x86-64"
+  ];
+
+  fetch('https://api.github.com/repos/kubernetes/minikube/releases')
```
medyagh, line 35: can we console out log how long this takes, so for reference ? I wanna know if it worth it to make our site slower by this call ..

medyagh (changes requested):

site/content/en/docs/start/_index.md
Sryther: I added a timer to measure the performance. Anyway it's asynchronous, it shouldn't affect page loading 🙂 .

site/content/en/docs/start/_index.md
```diff
@@ -33,7 +66,7 @@ Click on the buttons that describe your target platform. For other architectures
 {{% /quiz_row %}}
 
 {{% quiz_row base="/Linux/x86-64" name="Release type" %}}
-{{% quiz_button option="Stable" %}} {{% quiz_button option="Beta" %}}
+{{% quiz_button option="Stable" %}} {{% quiz_button option="Beta" hide="true" %}}
```
spowelljr, line 69: This will need to be added to all `Beta` buttons, should be 8 in total

spowelljr (changes requested):

site/content/en/docs/start/_index.md
Sryther: You're right! Sorry for that

spowelljr (approved):
Thanks for the PR!

PR #18591, opened 2024-04-07
Addition of German translations

jklippel:
Some strings are not translated to German yet.

This commit adds translations for these strings.


medyagh (approved):
danke schon !

PR #18602, opened 2024-04-08
addon: add addon for volcano

ComradeProgrammer:
Based on #13876
Add volcano addon, together with Invalid integration test fixed and incorrect manifest digest changed.
A auto updater is also added.


medyagh (changes requested):
https://github.com/kubernetes/minikube/actions/workflows/update-kong-version.yml

medyagh:
@ComradeProgrammer we could add the auto updater for the addon in a follow up PR

medyagh:
it seems like the arm64 test all fail so we shoudl tell users this is NOT supported for arm64 so they dont enable it
and also same for crio
we should skip the test and say doesnt work for crio yet

PR #18609, opened 2024-04-09
ci: bump macos version to 13

ComradeProgrammer:
ci: bump macos version to 13

related

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -527,8 +527,17 @@ jobs:
       JOB_NAME: "functional_virtualbox_macos"
       GOPOGH_RESULT: ""
       SHELL: "/bin/bash" # To prevent https://github.com/kubernetes/minikube/issues/6643
-    runs-on: macos-12
+    runs-on: macos-13
     steps:
+      - name: Check architecture
+        shell: bash
+        run: |
+          if [ $(uname -m) != "x86_64" ] 
+          then 
+              echo "$(uname -m) detected. This test must be run on x86_64"
+              exit -1
+          fi
+          exit 0
       - name: Install kubectl
         shell: bash
         run: |
@@ -547,6 +556,16 @@ jobs:
         run: |
           brew install docker-machine docker
           sudo docker --version
+      - name: Install vbox
+        shell: bash
+        run: |
+          brew update
+          brew search virtualbox
+          sudo spctl --master-disable
+          brew install --cask virtualbox 
+          curl -O -L "https://download.virtualbox.org/virtualbox/7.0.14/Oracle_VM_VirtualBox_Extension_Pack-7.0.14.vbox-extpack"
+          # Get SHAs from installing it manually once and then it will spit out the SHA for batch installation
+          echo "y" | sudo VBoxManage extpack install Oracle_VM_VirtualBox_Extension_Pack-7.0.14.vbox-extpack --accept-license=33d7284dc4a0ece381196fda3cfe2ed0e1e8e7ed7f27b9a9ebc4ee22e24bd23c
       - name: Info
         shell: bash
         run: |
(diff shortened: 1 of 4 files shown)


ComradeProgrammer:
@spowelljr  @medyagh could you please type an ok-to-test?
Now I changed the value of `runs-on` to `[macos-13,x64]` according to https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job 

I think it may works

spowelljr:
The additional tags are for self-hosted runners where you specify the tags on the runner itself, I don't think this is going to work.

ComradeProgrammer:
> The additional tags are for self-hosted runners where you specify the tags on the runner itself, I don't think this is going to work.

I see. I shall try something else later

ComradeProgrammer:
I think this time it might work.
Basically, what I did is that adding a new workflow into the github action. This workflow use PR workflow as its dependency, so that it will be triggered once this workflow is finished.

This workflow will try to use gh commandline to check the log of its depedent PR workflow to see if there are warnings about incorrect architecture(`gh run view ${{ github.event.workflow_run.id }} --log | grep ':type=>:arm'`) .

 If such error is found in the log, then it will try to fetch the jobid of `PR / functional_virtualbox_macos ` (` JOBID=`gh run view ${{ github.event.workflow_run.id }} --json jobs | jq '.jobs[] | select ( .name == "functional_virtualbox_macos") | .databaseId'``)and rerun it(`gh run rerun --job $JOBID`)

I created a test repo and used a simplified version of it, and I think it works. However, it seems that this change cannot take effect before it is merged.

.github/workflows/virtual_box_rerun.yml
```diff
@@ -0,0 +1,35 @@
+# This is a basic workflow to help you get started with Actions
```
medyagh, line 1: add better comments and explain what is this and why we need it

medyagh:
> mmandline to check the log of its depedent PR workflow to see if there are warnings about incorrect architecture(`gh run view ${{ github.event.workflow_run.id }} --log | grep ':type=>:arm'`) .
> 
> If such error is found in the log, then it will try to fetch the jobid of `PR / functional_vir`

cool, lets see if you can make the Test Pass too ?

.github/workflows/virtual_box_rerun.yml
```diff
@@ -0,0 +1,35 @@
...
+  check-and-rerun:
+    runs-on: ubuntu-latest
+    permissions: write-all
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: CheckArm64AndRerun
+        run: | 
+          set +e
+          gh run view ${{ github.event.workflow_run.id }} --log | grep ':type=>:arm'
```
medyagh, line 25: in the Original Job, we should have a Check that checks "step" for Architecture
and in the step it should fail Nice-er

.github/workflows/pr.yml
```diff
@@ -527,8 +527,17 @@ jobs:
...
     steps:
+      - name: Check architecture
+        shell: bash
+        run: |
+          if [ $(uname -m) != "x86_64" ] 
+          then 
+              echo "$(uname -m) detected. This test must be run on x86_64"
+              exit -1
+          fi
+          exit 0
```
medyagh, line 540: if this "Exit 0" will it still be triggered by the Re-Run job

medyagh: I also suggest that you create an issue in in minikube and also in Github Action itself, so they can fix this for everyone, they should not give users Arm Machines

medyagh (changes requested):

medyagh:
@ComradeProgrammer can we make sure we have create an issue in Github Action to make sure they know this issue exists?

spowelljr:
How often was this occurring? I just tried to replicate this, I ran 50 jobs using a `macos-13` runner and all of them gave me an x86 machine. Was this possibly resolved?

medyagh:
I think I would like to do this PR without the Re-Run logic that way at least we will have a nice message that it failed

spowelljr:
VirtualBox doesn't support macOS 13+ so I think we can close this PR.

PR #18666, opened 2024-04-17
clean up go mod

medyagh:
trying to remove two replaces that might not be needed anymore

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -239,11 +239,9 @@ require (
 )
 
 replace (
-	git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
 	github.com/Parallels/docker-machine-parallels/v2 => github.com/minikube-machine/machine-driver-parallels/v2 v2.0.1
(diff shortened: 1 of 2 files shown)


afbjorklund (changes requested):
Please don't revert the cpuid renaming that was done previously.

* https://github.com/minikube-machine/machine/pull/4

If it needs to be done, it should go in minikube-machine properly...

https://github.com/minikube-machine/machine/blob/main/drivers/virtualbox/vtx_intel.go

medyagh:
> Please don't revert the cpuid renaming that was done previously.
> 
> * [Organization of github.com/intel-go/cpuid changed minikube-machine/machine#4](https://github.com/minikube-machine/machine/pull/4)
> 
> If it needs to be done, it should go in minikube-machine properly...
> 
> https://github.com/minikube-machine/machine/blob/main/drivers/virtualbox/vtx_intel.go

good call thank you, yes I rather get rid of as many replaces as I can :)  I will do that on minikube-machine fist then

medyagh:
made a PR https://github.com/minikube-machine/machine/pull/5

PR #18707, opened 2024-04-22
[Multi-OS Cluster]: Windows Node Support in minikube Proposal

bobsira:


enhancements/proposed/20240405-multi-os-cluster/20240405-multi-os-cluster.md
```diff
@@ -0,0 +1,48 @@
...
+* Setting up the Windows node – Create a Windows Virtual Machine and install Windows Server 2022 unattended via [Autounattend XML file](https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/automate-windows-setup?view=windows-11). 
+* Create and configure a Minikube cluster with Flannel as CNI - Remote into the Virtual Machine (Windows node) and install containerd, NSSM, kubelet and kubeadm respectively. 
+* Join the Windows node into the Minikube cluster using kubeadm join command. 
+* Configuring Flannel CNI and Kube-Proxy - configure the networking settings as any Kubernetes cluster. 
+
+It is also worth noting that SIG Windows group exploring to bring Windows nodes on non-Windows host using KVM2 to create the nodes and as the minikube driver.  
+
+ Users will be able to start multi-os clusters by specifying a windows flag as in the example below, 
+
+minikube start –windows-22-node=1 or minikube start -windows-19-node=1 {windows-osversion-numberofwindowsnode}
```
spowelljr, line 44: I'm thinking if there is a better way to handle these flags, but not sure if there is. Having a bunch of `windows-<ver>-nodes` flags doesn't seem ideal, but if you want to support the user creating different node versions on start I don't see anything better. I was thinking of `--windows-node-version=22 --windows-nodes=1` but that would limit to only specifying a single node version per start, do we anticipate people wanting to have different Windows versions on a cluster?

iankingori: it would make sense to limit what we can do at `minikube start` and leverage the `node add` command instead to create special clusters.

bobsira: I've added a section in the proposal for non-goals. We could limit what we want to accomplish as a start. I do not see immediate case where folks would want different windows versions on a cluster but it is worth thinking about. Ian's suggestions might be a good way to start thinking of it.

enhancements/proposed/20240405-multi-os-cluster/20240405-multi-os-cluster.md
```diff
@@ -0,0 +1,55 @@
...
+*   Does this fit with minikube's [principles](https://minikube.sigs.k8s.io/docs/concepts/principles/)?
+*   Are there other approaches to consider?
+*   Could the implementation be made simpler?
+*   Are there usability, reliability, or technical debt concerns?
+
+Please leave the above text in your proposal as instructions to the reader.
+
+## Summary
+
+Struggling with comprehensive testing for applications across Linux and Windows environments? We are proposing to enhance Minikube's capabilities by introducing Windows node support. Minikube start should have the option of setting up a multi-OS cluster with a Linux control plane and both Linux and Windows worker nodes. This is [one of the most requested features](https://github.com/kubernetes/minikube/issues/2015) among the issues opened in the minikube repository. Developers will experience a significant boost in testing capabilities, streamlined workflows, and faster deployment, contributing to a more user-friendly and developer-focused environment
```
medyagh, line 19: minikube with lower case m everywhere :)

medyagh (commented):
this generally looks like a good idea to me to bring kubernetes to more operating systems,

we need to make sure we have a reliable testing infra for this.

enhancements/proposed/20240405-multi-os-cluster/20240405-multi-os-cluster.md
bobsira: Fixed this!

PR #18720, opened 2024-04-22
CI: Go auto update improvements

spowelljr:
Auto build kicbase image instead of ISO since we're not modifying the ISO in Go update PR anymore

Also remove updating `go.mod` because the potential formats of the Go version has changed, plus we don't need to bump it manually, modules that require it will auto update it.

### DIFF
--- a/hack/update/golang_version/update_golang_version.go
+++ b/hack/update/golang_version/update_golang_version.go
@@ -73,9 +73,8 @@ var (
 
 // Data holds stable Golang version - in full and in <major>.<minor> format
 type Data struct {
-	StableVersion   string
-	StableVersionMM string // go.mod wants go version in <major>.<minor> format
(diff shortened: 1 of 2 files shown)


hack/update/golang_version/update_golang_version.go
```diff
@@ -36,11 +36,6 @@ var (
 	}
 
 	schema = map[string]update.Item{
-		"go.mod": {
```
medyagh, line 39: for the go.mod  update, what does Kubernetes project do ? it would be nice to be in Par with them.

spowelljr: It seems they just let it auto update when the underlying packages require it, if an underlying package requires it and you run `go mod tidy` it will change the version automatically, we could follow that.

medyagh: that sounds reasonable

medyagh (approved):

PR #18723, opened 2024-04-22
Base ISO cache file name on the full ISO URL

albertofaria:
Currently, only the basename of the URL path is used, which can be the same for different URLs.

Base the cache file name on the full URL instead, encoding it using Go's base64.URLEncoding so it can be used as a file name on all platforms.

### DIFF
--- a/pkg/minikube/download/iso.go
+++ b/pkg/minikube/download/iso.go
@@ -17,6 +17,7 @@ limitations under the License.
 package download
 
 import (
+	"crypto/sha1"
 	"fmt"
 	"net/url"
 	"os"
@@ -77,7 +78,10 @@ func localISOPath(u *url.URL) string {
 		return u.String()
 	}
 
-	return filepath.Join(detect.ISOCacheDir(), path.Base(u.Path))
+	urlHash := sha1.Sum([]byte(u.String()))
+	fileName := fmt.Sprintf("%s-%040x", path.Base(u.Path), urlHash)
+
+	return filepath.Join(detect.ISOCacheDir(), fileName)
 }
 
 // ISO downloads and returns the path to the downloaded ISO


medyagh:
@albertofaria do you mind elaborating with an example what you are trying to improve ? maybe share with a output example

albertofaria:
A colleague reported that a stale cached ISO was being used after changing the value of the --iso-url option. This is because both the previous and new URLs given with --iso-url referenced GitLab package artifacts, which always end in a /download component.

They worked around this by manually deleting ~/.minikube/cache/iso/amd64/download.

albertofaria:
@medyagh I added more details to the commit message.

spowelljr:
Hi @albertofaria, I understand the point of the fix, but fully changing the name will make it hard to know the contents of the ISO cache dir for the average user.

ie. What version is `aHR0cHM6Ly9zdG9yYWdlLmdvb2dsZWFwaXMuY29tL21pbmlrdWJlLWJ1aWxkcy9pc28vMTg3MDYvbWluaWt1YmUtdjEuMzMuMC0xNzEzNzM2MjcxLTE4NzA2LWFybTY0Lmlzbw==`?

Could we maybe append it to the existing version so we can still identify the version for the default ISO file names
`minikube-v1.33.0-arm64.iso-<encoding>`

spowelljr:
That seems good, the 137 chars is a bit extreme, 40 sounds a lot better

pkg/minikube/download/iso.go
```diff
@@ -77,7 +78,10 @@ func localISOPath(u *url.URL) string {
 		return u.String()
 	}
 
-	return filepath.Join(detect.ISOCacheDir(), path.Base(u.Path))
+	urlHash := sha1.Sum([]byte(u.String()))
+	fileName := fmt.Sprintf("%s-%040x", path.Base(u.Path), urlHash)
```
nirs, line 82: This makes the iso file name a little bit ugly to support questionable API of a single application. The common way to publish downloads is to provide a URL ending with the filename, so the current behavior makes sense.

Since the image name always end with `.iso` or `.ISO`, and the file name is likely to be in the URL even in gitlab, can be search back the URL components and take the first component ending with `.(iso|ISO)` as the file name?

Or, maybe apply this behavior only if the URL does not end with `.(iso|ISO)`? This way this change helps gitlab users, without affecting other users.

Also did you open a bug for gitlab to provide a better URL? This /download URL will cause trouble for other tools, for example what do you get with `curl -O`?

nirs: Another way to solve this, add --iso-filename option. When set, using --iso-url will store the iso using the filename specified in --iso-filename. With this if you download the iso from a system that does not provide the common url with meaningful file name at the end, you can control the stored file name.

PR #18765, opened 2024-04-29
fix kubeflow docker image not available error for Mac M1(ARM 64)

BodhiHu:
Hello,

This PR will fix kubeflow docker image not available error for Mac M1(aarch64).

Thanks
Bodhi


spowelljr:
Fix is included in latest release of minikube: https://github.com/kubernetes/minikube/releases/tag/v1.33.1

Issue #18784, opened 2024-05-01
Some addon files are being templated when they shouldn't


spowelljr:
When adding addon files we call `MustBinAsset`:
https://github.com/kubernetes/minikube/blob/86fc9d54fca63f295d8737c8eacdbb7987e89c67/pkg/minikube/assets/addons.go#L105-L111

Which calls `NewBinAsset`:
https://github.com/kubernetes/minikube/blob/86fc9d54fca63f295d8737c8eacdbb7987e89c67/pkg/minikube/assets/vm_assets.go#L329-L342

Which calls `loadData`:
https://github.com/kubernetes/minikube/blob/86fc9d54fca63f295d8737c8eacdbb7987e89c67/pkg/minikube/assets/vm_assets.go#L355-L375

Which results in every addon file running through a templater. Only some files need templating, but up to now templating files that don't need it hasn't caused any issues.

But while trying to update the Kubeflow addon the YAML file has templating that essentially is used as validation during runtime to check for version compatibilities, but we're trying to template the file on load, resulting in:
```
panic: Failed to define asset kubeflow/kubeflow.yaml: template: kubeflow/kubeflow.yaml:118901: function "Template_Version_And_Istio_Version_Mismatched_Check_Installation" not defined

goroutine 1 [running]:
k8s.io/minikube/pkg/minikube/assets.MustBinAsset({0x10317f7e0?}, {0x101f2a8ec, 0x16}, {0x101f29d94?, 0x1?}, {0x101f1131e?, 0x14000593928?}, {0x101efea1b?, 0x10?})
	/Users/powellsteven/repo/minikube/pkg/minikube/assets/vm_assets.go:323 +0xb8
k8s.io/minikube/pkg/minikube/assets.init()
	/Users/powellsteven/repo/minikube/pkg/minikube/assets/addons.go:780 +0x687c
```

We have strictly been suffixing files that need templating with `.tmpl` so we can use that to decide what files need our templating. Everything else just load normally.

PR #18785, opened 2024-05-01
Addons: Only template `.tmpl` files

spowelljr:
Closes https://github.com/kubernetes/minikube/issues/18784

The full explanation for this can be found in https://github.com/kubernetes/minikube/issues/18784, but the short is that we're templating all addon files on load. This hasn't caused an issue until now, but when trying to update the kubeflow addon they added templating that acts as version validation on runtime, but we're templating on load, resulting in:
```
panic: Failed to define asset kubeflow/kubeflow.yaml: template: kubeflow/kubeflow.yaml:118901: function "Template_Version_And_Istio_Version_Mismatched_Check_Installation" not defined

goroutine 1 [running]:
k8s.io/minikube/pkg/minikube/assets.MustBinAsset({0x10317f7e0?}, {0x101f2a8ec, 0x16}, {0x101f29d94?, 0x1?}, {0x101f1131e?, 0x14000593928?}, {0x101efea1b?, 0x10?})
	/Users/powellsteven/repo/minikube/pkg/minikube/assets/vm_assets.go:323 +0xb8
k8s.io/minikube/pkg/minikube/assets.init()
	/Users/powellsteven/repo/minikube/pkg/minikube/assets/addons.go:780 +0x687c
```

Since we strictly add `.tmpl` to the filename of files that need templating, check for that suffix to decide if the file needs to be templated or not.

### DIFF
--- a/pkg/minikube/assets/vm_assets.go
+++ b/pkg/minikube/assets/vm_assets.go
@@ -26,6 +26,7 @@ import (
 	"path"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


spowelljr:
Blocked by https://github.com/kubernetes/minikube/issues/18784, will rebase after https://github.com/kubernetes/minikube/pull/18785 is merged

medyagh (approved):

PR #18814, opened 2024-05-06
18724

xcarolan:
Ref: Issue 18724
Exit gracefully  from mount of 9p file type when the host does not support 9p.
/sig cluster-lifecycle
/kind improvement
/area minikube


cmd/minikube/cmd/mount.go
```diff
@@ -239,11 +239,14 @@ var mountCmd = &cobra.Command{
...
 
 		err = cluster.Mount(co.CP.Runner, ip.String(), vmPath, cfg, pid)
+
 		if err != nil {
 			if rtErr, ok := err.(*cluster.MountError); ok && rtErr.ErrorType == cluster.MountErrorConnect {
 				exit.Error(reason.GuestMountCouldNotConnect, "mount could not connect", rtErr)
 			}
-			exit.Error(reason.GuestMount, "mount failed", err)
+			out.ErrT(style.Failure, "Failed mount: {{.error}}", out.V{"error": err})
+			exit.Message(reason.GuestMount, "mount failed, does host support filesystem type 9p?")
```
medyagh, line 248: @xcarolan I dont think this PR is actually doing any detection, or any improvement over the current situation,
if I am wrong, do you mind pasting the output Before and After this PR to prove?

there should be a mechanisim to Detect if this the host supports 9P or not, and if it doesn't it should detect it let the user know befrehand

medyagh (changes requested):
please see the comment

xcarolan:
yes, all I did was change the error type so instead of this:
image

you get this:

image

I was trying to find some way of checking the host system to see if it supported 9p, but could find anything. If you are aware of a way of checking, or can point me in the right direction, I'll go back and change it. As it stands now, i'm really not sure how useful (if at all) this change is.
Thanks for the feedback,

xcarolan:
Actually let me pull this back and rework it,,

xcarolan:
reworking...

PR #18858, opened 2024-05-12
Add AppArmor Support #8299

nnzv:
Implemented AppArmor support for Minikube on both aarch64 and x86_64 architectures. Testing pending, relying on documentation for guidance. Excluded unnecessary extras:

- `BR2_PACKAGE_APPARMOR_PROFILES`: Community profiles deemed non-essential for Minikube users.


nnzv:
I'm checking if the kernel image from Buildroot supports AppArmor. Following the [docs](https://minikube.sigs.k8s.io/docs/contrib/building/iso/#modifying-buildroot-components), I tried "linux-menuconfig" in the Makefile, but it says there's nothing to be done. Am I overlooking any essential steps? Also, for an overview of the kernel modifications required, here's a summary of the necessary options:

```
General setup --->
    -*- Auditing support

Security options  --->
    -*- Enable the securityfs filesystem
    -*- Socket and Networking Security Hooks
    [*] Enable different security models
    [*] AppArmor support
          [*]   Enable introspection of sha1 hashes for loaded profiles
          [*]      Enable policy hash introspection by default
    [ ] Build AppArmor with debug code
          First legacy 'major LSM' to be initialized (AppArmor)  --->
          "yama,apparmor" Ordered List of enabled LSMs
```

nnzv:
CC: @tstromberg

nnzv:
I've gone through the entire image building process to test out the new environment, following the documentation closely. It instructs to use the target `out/minikube-amd64.iso`, but it failed near the end. I fixed it by switching "amd64" to "x86_64". If this change gets merged into the master branch, we might need to update the documentation. Check the [logs](https://github.com/kubernetes/minikube/files/15289577/log.txt), especially after I switched to the "x86_64" target. The original log file is massive, around 56MB of plain text.


```
...
make[1]: Leaving directory '/mnt/out/buildroot'
# x86_64 ISO is still BIOS rather than EFI because of AppArmor issues for KVM, and Gen 2 issues for Hyper-V
if [ "x86_64" = "aarch64" ]; then \
                mv /mnt/out/buildroot/output-aarch64/images/boot.iso /mnt/out/minikube-arm64.iso; \
        else \
                mv /mnt/out/buildroot/output-x86_64/images/rootfs.iso9660 /mnt/out/minikube-amd64.iso; \
        fi;
Please enter a valid architecture. Choices are x86_64 and aarch64.
make: *** [Makefile:959: deploy/iso/minikube-iso/board/minikube/amd64/rootfs-overlay/usr/bin/auto-pause] Error 1
make: *** [Makefile:333: out/minikube-amd64.iso] Error 2
...

```

medyagh:
@nnzv do you mind sharing what use cases this would help with ? are there any issues that users are already asking for this ?

nnzv:
> Do you mind sharing what use cases this would help with ? are there any issues that users are already asking for this ?

User @acoulon99 requested this feature some time ago via an issue (https://github.com/kubernetes/minikube/issues/8299). For detailed information on its use cases, you can refer to the Kubernetes documentation at https://kubernetes.io/docs/tutorials/security/apparmor, particularly the [example](https://kubernetes.io/docs/tutorials/security/apparmor/#example) of a profile that blocks all file write operations on nodes. It's surprising that Minikube hasn't incorporated this feature, considering that Kubernetes has supported AppArmor since version `1.14`. Implementing it would be a logical next step.

medyagh:
ok-to-build-iso

medyagh:
> > Do you mind sharing what use cases this would help with ? are there any issues that users are already asking for this ?
> 
> User @acoulon99 requested this feature some time ago via an issue (#8299). For detailed information on its use cases, you can refer to the Kubernetes documentation at https://kubernetes.io/docs/tutorials/security/apparmor, particularly the [example](https://kubernetes.io/docs/tutorials/security/apparmor/#example) of a profile that blocks all file write operations on nodes. It's surprising that Minikube hasn't incorporated this feature, considering that Kubernetes has supported AppArmor since version `1.14`. Implementing it would be a logical next step.

is there a way we could Enable/Disable the feature on user's request ? I think we should have a flag to like --apparomor so if users face issues they could disable it

nnzv:
> If we I think we should have a flag to like --apparomor so if users face issues they could disable it. I am gonna build the ISO on the PR just to see initial tests results

You're welcome. Indeed, it does seem like a good idea. I wasn't aware that it was implemented long ago but disabled due to breaking changes. Regarding the new flag, I'm not sure how to dynamically ignore things by buildroot, specifically referring to the `iso/minikube-iso/configs/*` files depending on the architecture.

nnzv:
Additionally, _keep in mind that this is only partially implemented_. The remaining task involves configuring the kernel. You can find details in one of my [comments](https://github.com/kubernetes/minikube/pull/18858#issuecomment-2106096168) below.

medyagh:
yes I think there was issues with github action running with apparmor (for KIC - docker/podman drivers)
https://github.com/kubernetes/minikube/pull/7658

@nnzv the ISO build failed btw but not related to this PR (yet) the build failure is due to the changes in SHA by https://github.com/containernetworking/plugins/issues/1038 that they unexpectedly built the binary and overwrote the previous ones

for enable disabling it I am thinking if there should be a linux command to turn it on or off after the ISO is built. so if users start minikube --apparmor=true it would be enabled in the kernel module. that way we would not have a lot of new headaches for breaking people's existing workflows

medyagh:
@nnzv do you mind pulling upstream just merged https://github.com/kubernetes/minikube/pull/18896

nnzv:
> For enable disabling it I am thinking if there should be a linux command to turn it on or off after the ISO is built. so if users start minikube --apparmor=true it would be enabled in the kernel module. that way we would not have a lot of new headaches for breaking people's existing workflows.

Yeah, sounds like a [modprobe](https://man7.org/linux/man-pages/man8/modprobe.8.html). So, the AppArmor config needs to be `<M>` instead of `<*>`. I didn't check Minikube's code, how it handles flags, etc. But I guess it'll look something like this. Apologies for the simplistic code.

```go
...
if (*WantAppArmorSupport) {
    sshtunnel("sudo modprobe apparmor")
    // Additional logic can go here for further validation
}
...
```

deploy/iso/minikube-iso/configs/minikube_aarch64_defconfig
```diff
@@ -3,6 +3,7 @@ BR2_aarch64=y
 
 # Toolchain
 BR2_TOOLCHAIN_BUILDROOT_WCHAR=y
+BR2_TOOLCHAIN_BUILDROOT_CXX=y
```
medyagh, line 6: was enabling c++ libraries needed ?

medyagh:
@nnzv the ISO has been built succesfully, I think could you please  test it manually to see if it is enabled ?
and also how about we explore setting somethig like

BR2_TARGET_GENERIC_INITSCRIPT_APPARMOR=n

to prevent AppArmor from starting automatically at boot and then if the user wants it they can set it up

deploy/iso/minikube-iso/configs/minikube_aarch64_defconfig
nnzv: In menuconfig, under `Target Packages -> Security`, you'll find that you can't enable AppArmor until you meet some prerequisites, like enabling C++.

image

nnzv:
> and also how about we explore setting somethig like
>
> BR2_TARGET_GENERIC_INITSCRIPT_APPARMOR=n
>
> to prevent AppArmor from starting automatically at boot and then if the user wants it they can enable it

Is this config unique to Buildroot? Haven't found it. If you want to disable by default, AppArmor doesn't run as a service. You can try [apparmor_parser](http://manpages.ubuntu.com/manpages/precise/en/man8/apparmor_parser.8.html), but I haven't set up AppArmor in the kernel https://github.com/kubernetes/minikube/pull/18858#issuecomment-2106096168. Let me commit that change. Sorry for the delay.

image

medyagh:
@nnzv I didnt see any new change in your commit, if apparmor is added you should be able to verify it locally
(try running minikube start --iso-url) and provide local file path as the url to test it

medyagh:
ok-to-build-iso

nnzv (commented):
After researching, I found _that apparmor can't be activated as a kernel module_. So, with this setup, the kernel won't use apparmor LSM by default. To enable it dynamically, we may need to tweak [kernel parameters](https://www.kernel.org/doc/html/v4.15/admin-guide/LSM/apparmor.html#how-to-enable-disable) in the `grub.cfg`. So, after setting up a node, we'd SSH in, adjust the `grub.cfg`, or generate it with grub-mkconfig after editing `/etc/default/grub`.

```
minikube ssh
# ...
vi /etc/default/grub
# ...
grep CMDLINE_LINUX /etc/default/grub
# GRUB_CMDLINE_LINUX="... apparmor=0 security=apparmor ..."
grub-mkconfig -o /boot/grub/grub.cfg
# ...
```

Also we can manually edit `/boot/grub/grub.cfg` or its equivalent on the node to add the necessary kernel parameters. Afterward, a reboot is required for the kernel to recognize the new LSM. Of course, this explanation is simply to outline the process; ideally, **automation could be achieved** using a boolean flag variable like `--apparmor`, as you mentioned.

nnzv:
> @nnzv I didnt see any new change in your commit, if apparmor is added you should be able to verify it locally (try running minikube start --iso-url) and provide local file path as the url to test it.

No worries. It's enabled now. Since I don't have my ThinkPad (the one I used for development) with me, I'll wait for Minikube's CI to finish building the ISO. Once that's done, I'll test the POC I mentioned earlier, specifically about enabling apparmor with kernel parameters. Enabling it by default (adding `CONFIG_DEFAULT_SECURITY_APPARMOR`) in the kernel configuration could pose a problem. Take a look at this quote from a GitHub blog post: https://github.blog/2023-07-05-introduction-to-selinux/

> Please note that some Linux distributions have different MAC systems enabled, and running multiple exclusive LSMs at the same time will cause issues. For example, Ubuntu has AppArmor enabled by default, therefore enabling SELinux while AppArmor is enabled will prevent you from booting properly. Please check include/linux/lsm_hooks.h for the LSM_FLAG_EXCLUSIVE to see if your LSM is exclusive.

nnzv:
Hey @medyagh, could we give "ok-to-build-iso" another shot? I've checked the ISO created by the bot, but it seems my changes weren't taken into account (https://github.com/kubernetes/minikube/pull/18858/commits/494ee1dc10f3d6ffea8808e9f337bd347663465f and https://github.com/kubernetes/minikube/pull/18858/commits/e109ed0e07154022ee2b01beff17916a9ab59178). By the way, do you happen to know how we can tweak kernel parameters? I'm aware that each architecture in `deploy/iso/minikube-iso/board` has its own `grub.cfg`, but I couldn't find anything when I ran `find / -name "grub.cfg" -type f` from the root after SSHing into the minikube node.

medyagh:
@nnzv sure can you plz rebase your PR so I can trigger it one more time ?

nnzv:
Alright, so we got the utils binaries in place. But now, we need to activate apparmor using kernel parameters. It's disabled at boot by default to prevent [potential issues](https://github.com/kubernetes/minikube/pull/18858#issuecomment-2130420876) in case minikube decides to utilize another LSM like Selinux in the future.

```sh
minikube ssh
aa-enabled
# No - disabled at boot.
```

I see there's an `isolinux.cfg` for each board in `deploy/iso/minikube-iso`. It looks like you're using syslinux, though I couldn't confirm it in the docs. But the file suggests it. If that's the deal, you might want to create a [syslinux.cfg](https://wiki.archlinux.org/title/kernel_parameters#Syslinux) file in the `/boot/syslinux` directory. It allows you to adjust the [kernel parameters](https://www.kernel.org/doc/html/v4.15/admin-guide/LSM/apparmor.html#how-to-enable-disable) necessary to enable apparmor.

```sh
minikube ssh
sudo su -
mkdir /boot/syslinux
cat << EOF > /boot/syslinux/syslinux.cfg
APPEND apparmor=1 security=apparmor
EOF
reboot
````
Data isn't sticking around after a reboot, which is essential for loading "syslinux.cfg" to enable apparmor. From what I gather, there are specific directories that the iso keeps intact after a reboot, like `/var/tmp/minikube`, `/data`, `/tmp/hostpath-provisioner/` or `/tmp/hostpath_pv.` I'm not sure if there's a method like a hook that minikube uses to preserve symlinks, so we could symlink the syslinux directory under one of these special directories to `/boot`. Could you kindly see if any of the other maintainers might have insights on this matter? @medyagh 

```sh
minikube ssh
ls -l /boot
# total 21956
# -rw-r--r-- 1 root root 22482368 May 30 04:42 bzImage
````

mbana:
Hi folks, any updates on when this will get merged?

mhils:
My additions in https://github.com/mhils/minikube/commits/patch-apparmor/ make `aa-enabled` work, feel free to pull them in. This was based on a bunch of trial and error, so I'm not sure if every change is needed. I figured I'd share nonetheless, unfortunately I'm running out of time to investigate further. :)

mhils:
My additions in https://github.com/mhils/minikube/commits/patch-apparmor/ make `aa-enabled` work, feel free to pull them in. This was based on a bunch of trial and error, so I'm not sure if every change is needed. I figured I'd share nonetheless, unfortunately I'm running out of time to investigate further. :)

PR #18959, opened 2024-05-25
Issue 18724 Added check for host support of file system 9p before mounting

xcarolan:
…nting.

resolves issue <18724> 
Checks if host system supports file system 9p before attempting mount.


pkg/minikube/reason/exitcodes.go
```diff
@@ -78,7 +78,7 @@ const (
 	ExHostNotFound    = 35
 	ExHostUnsupported = 38
 	ExHostPermission  = 37
-	ExHostConfig      = 38
+	ExHostConfig      = 39
```
medyagh, line 81: how about instead of changing this to 39 
ExHostUnsupported = 36
that way they would be in order
```
	ExHostTimeout     = 32
	ExHostUsage       = 34
	ExHostNotFound    = 35
	ExHostUnsupported = 36
	ExHostPermission  = 37
	ExHostConfig      = 38
```

xcarolan:
Good catch - I missed that! Thanks

Issue #18956, opened 2024-05-24
site: make tabbed view linkable


medyagh:
### What Happened?

for example in these pages we have Tabs
https://minikube.sigs.k8s.io/docs/start/
for example we have linux mac windows tabs here
https://minikube.sigs.k8s.io/docs/handbook/addons/ingress-dns/

when we click on the Active Tab we should be able to link to the Tab we want in order to give to another person, lets say I wanna only link to the Windows there should be a way to link to it regardless of the user click

### Attach the log file

n/a

### Operating System

None

### Driver

None

rihib:
Hello, can I work on this issue?

rihib:
@medyagh

hritesh04:
can i work on this ?

medyagh:
@hritesh04 sorry had to revert your PR while it looked good when selecting a tab but it was broken on the index see the preview

https://deploy-preview-18977--kubernetes-sigs-minikube.netlify.app/docs/start/

spowelljr:
Hi @hritesh04, regarding the revert, I believe the issue is due to moving setting the default selection into an `else`:
```
const archParam = urlParams.get("arch");

//checks for query params and process each option one by one

if (archParam) {
    ...
} else {
    // auto-select the OS for user
    const btn = $(".option-button[data-quiz-id='/" + userOS + "']").first();
    btn.addClass("active");
    selectQuizOption(btn.attr("data-quiz-id"));
}
```

It seems that `archParam` is resulting to `true` when it shouldn't

hritesh04:
hey @spowelljr ,
I think the issue was created when the site was accessed from a Mac. if do there was a issue on line 59 of quiz.js which i fixed and opened a pr for. Please check that #18982

medyagh:
closed https://github.com/kubernetes/minikube/pull/18982

PR #18977, opened 2024-05-28
site: made tabs view linkable #18956

hritesh04:
Before :-
the options are auto selected based on user os on url - https://minikube.sigs.k8s.io/docs/start/
image

After :-
the options are selected based on the query-parameters and if there is no query-parameter options are selected according to os
url - http://localhost:1313/docs/start/?arch=%2Flinux%2Farmv7%2Fstable%2Fdebian+package
url format - http://site_name.io/docs/start/?arch=/operating_system/architecture/release_type/installer_type
' / ' are encoded to %2F
image

Before :-
option is auto selected according to user os on url - https://minikube.sigs.k8s.io/docs/handbook/addons/ingress-dns/
image

After :-
option is selected according to anchor in the url if none then selected based on user os
applicable to all other tabs on the website as well

url format - http://site_name.io/path#option

image

### Changes
data-quiz-id attribute on quiz_button, quiz_instruction, quiz-row converted to lower case.
added functionality to auto select options based on query-params
added href tag to each tab

### DIFF
--- a/site/layouts/shortcodes/quiz_button.html
+++ b/site/layouts/shortcodes/quiz_button.html
@@ -1,1 +1,1 @@
-<button data-quiz-id="{{ .Parent.Get "base" }}/{{ .Get "option" }}" type="button" class="btn btn-outline-primary option-button{{ if eq (.Get "hide") "true" }} hide{{ end }}">{{ .Get "option" }}</button>
+<button data-quiz-id="{{ .Parent.Get "base" |  lower }}/{{ .Get "option" | lower}}" type="button" class="btn btn-outline-primary option-button{{ if eq (.Get "hide") "true" }} hide{{ end }}">{{ .Get "option" }}</button>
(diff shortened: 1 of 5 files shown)


medyagh (approved):
@hritesh04 excellent ! thank you very much, this adds quality of life for us and our users, now we can link to a specific tab,

@hritesh04 do you mind checking if there are any other page in our website to add link for the tab?

additional, if we have a new tab in the future, could you add an instruction how to add new ones ? maybe as a follow up PR add a new section called "Website" under our
"Contributing" section
https://minikube.sigs.k8s.io/docs/contrib/

medyagh:
thank you I look forward to see more contributions from you @hritesh04 please feel free to suggest or pick up any website improvement tasks :)

hritesh04:
@medyagh hey sorry about that,
i would like to give it another try can you give more information like which os was this running on because its working fine on  my windows. 
Also other tabs should also be linkable but i will take another look and let you know

hritesh04:
i have created a pull request for the fix and the other tabs on the website are also linkable as long they use body-end.html file at \site\layouts\partials\hooks\body-end.html

PR #18982, opened 2024-05-28
Fix for linkable tabs ui breaking

hritesh04:
#18977 

fix for breaking changes 

### problem 

The issue was on macOS where the ui breaks when there is no parameters

### solution

in quiz file in line 59 the comparison should be with mac not Mac

This will fix the ui breaking if the issue was generated in macos

### DIFF
--- a/site/static/js/quiz.js
+++ b/site/static/js/quiz.js
@@ -24,19 +24,17 @@ function selectQuizOption(selectedId, autoselect = true) {
     ".option-button"
   );
   // auto-select the first option for the user, to reduce the number of clicks
-  if (buttons.length > 0) {
-    if (autoselect) {
-      const btn = buttons.first();
-      const dataContainerId = btn.attr("data-quiz-id");
-      btn.addClass("active");
-      const url = new URL(window.location);
-      url.searchParams.set("arch", dataContainerId); // Add selectedId as query parameter
+  if (buttons.length > 0 && autoselect) {
+    const btn = buttons.first();
+    const dataContainerId = btn.attr("data-quiz-id");
+    btn.addClass("active");
+    const url = new URL(window.location);
+    url.searchParams.set("arch", dataContainerId); // Add selectedId as query parameter
 
-      // Update the browser's location with the new URL
-      window.history.replaceState({}, document.title, url);
+    // Update the browser's location with the new URL
+    window.history.replaceState({}, document.title, url);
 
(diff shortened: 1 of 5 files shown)


site/static/js/quiz.js
```diff
@@ -1,57 +1,99 @@
...
+  });
+  // hide other answers
+  $(".quiz-instruction").addClass("hide");
+  // show the selected answer
+  $(".quiz-instruction[data-quiz-id='" + selectedId + "']").removeClass("hide");
+
+  const buttons = $(".option-row[data-quiz-id='" + selectedId + "']").find(
+    ".option-button"
+  );
+  //auto-select the first option for the user, to reduce the number of clicks
```
spowelljr, line 26:
```suggestion
  // auto-select the first option for the user, to reduce the number of clicks
```

site/static/js/quiz.js
```diff
@@ -1,57 +1,99 @@
...
+    if (userOS === "mac") {
+      // use the name "macos" to match the button
+      userOS = "macos";
+    }
+    $(".option-row[data-level=0]").removeClass("hide");
+
+    const urlParams = new URLSearchParams(window.location.search);
+    const archParam = urlParams.get("arch");
+
+    //checks for query params and process each option one by one
```
spowelljr, line 68:
```suggestion
    // checks for query params and process each option one by one
```

spowelljr (changes requested):
Just those two spaces and looks good to me

hritesh04:
done

site/static/js/quiz.js
```diff
@@ -1,57 +1,99 @@
...
+  $(".quiz-instruction").addClass("hide");
+  // show the selected answer
+  $(".quiz-instruction[data-quiz-id='" + selectedId + "']").removeClass("hide");
+
+  const buttons = $(".option-row[data-quiz-id='" + selectedId + "']").find(
+    ".option-button"
+  );
+  // auto-select the first option for the user, to reduce the number of clicks
+  if (buttons.length > 0) {
+    if (autoselect) {
```
spowelljr, line 28: Once last thing, this is unnecessarily double nested, could these two be `&&` together

spowelljr (changes requested):

hritesh04:
done

spowelljr (approved):

PR #19156, opened 2024-06-28
update translations/zh-CN.json

uos-ljtian:
Updated some translations


uos-ljtian:
Comparison of Changes in Line 591
Original:
image

result:

PR #19182, opened 2024-07-02
testing: Use multi-arch example app image

spowelljr:
Part of the fix for getting Ingress Addon tests passing on arm64, other changes are required also


spowelljr:
> is this still needed after #19198 ?

Yes, #19198 made the ingress addon image multi-arch, but this makes the images we use in our tests multi-arch also.

Issue #19173, opened 2024-07-02
❌  Exiting due to GUEST_START: failed to start node: wait 6m0s for node: wait for healthy API server: apiserver healthz never reported healthy: context deadline exceeded


rexsunny:
### What Happened?

I am using MacOS Chip: Apple M1
MacOS: 14.4.1 (23E224)

niyaaniyan@Niyas-MacBook-Pro Downloads % minikube start
😄  minikube v1.33.1 on Darwin 14.4.1 (arm64)
    ▪ MINIKUBE_ROOTLESS=true
✨  Using the qemu2 driver based on existing profile
👍  Starting "minikube" primary control-plane node in "minikube" cluster
🏃  Updating the running qemu2 "minikube" VM ...
🐳  Preparing Kubernetes v1.30.0 on Docker 26.0.2 ...
🤦  Unable to restart control-plane node(s), will reset cluster: <no value>
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔗  Configuring bridge CNI (Container Networking Interface) ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
❗  Enabling 'default-storageclass' returned an error: running callbacks: [Error making standard the default storage class: Error listing StorageClasses: Get "https://localhost:53680/apis/storage.k8s.io/v1/storageclasses": dial tcp [::1]:53680: connect: connection refused]
🌟  Enabled addons: storage-provisioner

❌  Exiting due to GUEST_START: failed to start node: wait 6m0s for node: wait for healthy API server: apiserver healthz never 
reported healthy: context deadline exceeded

### Attach the log file

[logs.txt](https://github.com/user-attachments/files/16061371/logs.txt)

### Operating System

macOS (Default)

### Driver

QEMU

spowelljr:
It works on v1.32.0, but broke on v1.33.0, I'm thinking it might have been caused by an ISO config change?

spowelljr:
Found the commit where the regression was introduced, but it's a big commit, so will need to figure out exactly what part cased the regression.

https://github.com/kubernetes/minikube/commit/ad7fb3992cbbbe5f1bc918adfc75588fab51bc22

spowelljr:
PR for the fix is up, I'll include binaries once it's built so those who are experiencing the issue can use it until the next release.

https://github.com/kubernetes/minikube/pull/19191

[...]

PR #19191, opened 2024-07-03
Fixed not creating API server tunnel

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/19173

**Before:**
```
$ minikube start --driver qemu --network builtin
😄  minikube v1.33.1 on Darwin 14.5 (arm64)
❗  You are using the QEMU driver without a dedicated network, which doesn't support `minikube service` & `minikube tunnel` commands.
To try the dedicated network see: https://minikube.sigs.k8s.io/docs/drivers/qemu/#networking
❗  This VM is having trouble accessing https://registry.k8s.io
❗  Due to DNS issues your cluster may have problems starting and you may not be able to pull images
More details available at: https://minikube.sigs.k8s.io/docs/drivers/qemu/#known-issues
🐳  Preparing Kubernetes v1.30.2 on Docker 27.0.3 ...
    ▪ Configuring RBAC rules ...
E0703 14:15:40.236604   71562 start.go:159] Unable to scale down deployment "coredns" in namespace "kube-system" to 1 replica: non-retryable failure while getting "coredns" deployment scale: Get "https://localhost:54274/apis/apps/v1/namespaces/kube-system/deployments/coredns/scale": dial tcp 127.0.0.1:54274: connect: connection refused
❗  Enabling 'default-storageclass' returned an error: running callbacks: [Error making standard the default storage class: Error listing StorageClasses: Get "https://localhost:54274/apis/storage.k8s.io/v1/storageclasses": dial tcp 127.0.0.1:54274: connect: connection refused]
🌟  Enabled addons: storage-provisioner

❌  Exiting due to GUEST_START: failed to start node: wait 6m0s for node: wait for healthy API server: apiserver healthz never reported healthy: context deadline exceeded

```

**After:**
```
$ minikube start --driver qemu --network builtin
😄  minikube v1.33.1 on Darwin 14.5 (arm64)
❗  You are using the QEMU driver without a dedicated network, which doesn't support `minikube service` & `minikube tunnel` commands.
To try the dedicated network see: https://minikube.sigs.k8s.io/docs/drivers/qemu/#networking
❗  This VM is having trouble accessing https://registry.k8s.io
❗  Due to DNS issues your cluster may have problems starting and you may not be able to pull images
More details available at: https://minikube.sigs.k8s.io/docs/drivers/qemu/#known-issues
🐳  Preparing Kubernetes v1.30.2 on Docker 27.0.3 ...
    ▪ Configuring RBAC rules ...
```

### DIFF
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -444,6 +444,8 @@ func (k *Bootstrapper) tunnelToAPIServer(cfg config.ClusterConfig) error {
 	if cfg.APIServerPort == 0 {
 		return fmt.Errorf("apiserver port not set")
 	}
+	// An API server tunnel is only needed for QEMU w/ builtin network, for
+	// everything else return
 	if !driver.IsQEMU(cfg.Driver) || !network.IsBuiltinQEMU(cfg.Network) {
 		return nil
 	}
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -440,7 +440,7 @@ func (k *Bootstrapper) StartCluster(cfg config.ClusterConfig) error {
 
 // tunnelToAPIServer creates ssh tunnel between apiserver:port inside control-plane node and host on port 8443.
 func (k *Bootstrapper) tunnelToAPIServer(cfg config.ClusterConfig) error {
-	if cfg.APIServerPort != 0 {
+	if cfg.APIServerPort == 0 {
 		return fmt.Errorf("apiserver port not set")
 	}
 
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -64,6 +64,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/style"
 	"k8s.io/minikube/pkg/minikube/sysinit"
 	"k8s.io/minikube/pkg/minikube/vmpath"
+	"k8s.io/minikube/pkg/network"
 	"k8s.io/minikube/pkg/util"
 	"k8s.io/minikube/pkg/util/retry"
 	"k8s.io/minikube/pkg/version"
@@ -442,6 +443,9 @@ func (k *Bootstrapper) StartCluster(cfg config.ClusterConfig) error {
 func (k *Bootstrapper) tunnelToAPIServer(cfg config.ClusterConfig) error {
 	if cfg.APIServerPort == 0 {
 		return fmt.Errorf("apiserver port not set")
+	}
+	if !driver.IsQEMU(cfg.Driver) || !network.IsBuiltinQEMU(cfg.Network) {
+		return nil
 	}
 
 	m, err := machine.NewAPIClient()


pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -440,7 +440,7 @@ func (k *Bootstrapper) StartCluster(cfg config.ClusterConfig) error {
 
 // tunnelToAPIServer creates ssh tunnel between apiserver:port inside control-plane node and host on port 8443.
 func (k *Bootstrapper) tunnelToAPIServer(cfg config.ClusterConfig) error {
-	if cfg.APIServerPort != 0 {
+	if cfg.APIServerPort == 0 {
```
medyagh, line 443: how is this not broken for everything else ? or since forever ? is this logic new and only affecting qemu builtin ?

spowelljr: Yeah, I was looking into this because our functional tests are failing with this change.

From what I saw this is what happened:

**Before HA PR:**
`tunnelToAPIServer` was only called when it was necessary, ie. QEMU builtin
If it wasn't necessary (docker driver) it wasn't called

**After HA PR but before this fix:**
`tunnelToAPIServer` is getting called for all drivers, but because the logic was broken, it wasn't getting called for any drivers (which broke QEMU w/ builtin)

**This PR**
`tunnelToAPIServer` is getting called for all drivers, and now with the logic fixed it's actually creating a tunnel for all drivers, which fixes QEMU w/ builtin, but Docker (and maybe others) are having issues due to the tunnel because created when it shouldn't be.

So I need to figure out how `tunnelToAPIServer` was getting called prior to the HA change and match it so it's only getting called when necessary.

spowelljr: Ah, I found it, before the HA changes `APIServerPort` was only used by QEMU:
```
          if driver.IsQEMU(host.Driver.DriverName()) && network.IsBuiltinQEMU(cfg.Network) {
                  apiServerPort, err := getPort()
                  if err != nil {
                          return runner, preExists, m, host, errors.Wrap(err, "Failed to find apiserver port")
                  }
                  cfg.APIServerPort = apiServerPort
          }
```

But now it's also setting it via a flag:
```
APIServerPort:           viper.GetInt(apiServerPort),
```

The very first test in our functional tests sets the `--apiserver-port` flag, which is starting the tunnel and causing issues.

spowelljr: Added the QEMU check below

medyagh: excellent thanks for the investigation

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -440,9 +441,12 @@ func (k *Bootstrapper) StartCluster(cfg config.ClusterConfig) error {
...
 		return fmt.Errorf("apiserver port not set")
 	}
+	if !driver.IsQEMU(cfg.Driver) || !network.IsBuiltinQEMU(cfg.Network) {
```
medyagh, line 447: would be nice to add an english comment before this if statement for 
// return nil for qemu with builtin

spowelljr: Done

Issue #18822, opened 2024-05-07
Minikube commands trigger Docker context error.


albertalvin8080:
### What Happened?

When I run any minikube command, it's logged that the "default" Docker CLI context cannot be found. The thing is, I'm not trying to use docker. It happens even when Docker is not running.

For example, when i use `minikube status`, the output is:

_W0506 23:45:30.171668     800 main.go:291] Unable to resolve the current Docker CLI context "default": context "default": context not found: open C:\Users\Albert\.docker\contexts\meta\37a8eec1ce19687d132fe29051dca629d164e2c4958ba141d5f4133a33f0688f\meta.json: The system cannot find the path specified.
minikube
type: Control Plane
host: Stopped
kubelet: Stopped
apiserver: Stopped
kubeconfig: Stopped_

As you can see, the status command is executed normaly. But the docker error is always triggered before the actual execution of the command.

For `minikube config get vm-driver` command:

_W0506 23:59:03.135165   10532 main.go:291] Unable to resolve the current Docker CLI context "default": context "default": context not found: open C:\Users\Albert\.docker\contexts\meta\37a8eec1ce19687d132fe29051dca629d164e2c4958ba141d5f4133a33f0688f\meta.json: The system cannot find the path specified.
hyperv_

Even if it's still harmless (so far), I'd like to understand why it happens.

### Operating System

Windows

### Driver

N/A

spowelljr:
Hi @albertalvin8080, if you run `docker context ls` it should show the contexts available, it might be that to the location of the docker context moved perhaps? I would try running `docker context use default` to set it again and maybe that will resolve your issue.

albertalvin8080:
It worked perfectly, thank you! So it really seems like that my docker context was off and Minikube was trying to interact with it.

spowelljr:
Glad that worked! I created a PR to include those instructions when that error is shown to help others with this in the future (https://github.com/kubernetes/minikube/pull/19197).

PR #19197, opened 2024-07-04
Add instructions to resolve docker context error

spowelljr:
Closes https://github.com/kubernetes/minikube/issues/18822

### DIFF
--- a/cmd/minikube/main.go
+++ b/cmd/minikube/main.go
@@ -289,6 +289,7 @@ func propagateDockerContextToEnv() {
 	md, err := st.GetMetadata(currentContext)
 	if err != nil {
 		klog.Warningf("Unable to resolve the current Docker CLI context %q: %v", currentContext, err)
+		klog.Warningf("Try running `docker context use %s` to resolve the above error", currentContext)
 		return
 	}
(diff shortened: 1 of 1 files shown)


cmd/minikube/main.go
```diff
@@ -289,6 +289,7 @@ func propagateDockerContextToEnv() {
 	md, err := st.GetMetadata(currentContext)
 	if err != nil {
 		klog.Warningf("Unable to resolve the current Docker CLI context %q: %v", currentContext, err)
+		klog.Warningf("Try running `docker context use %s` to resolve the above error", currentContext)
```
medyagh, line 292: this is not always gonna be the case !
it is true that if they set a wrong context it will not be able to resolve the docker,  but it would say the error if the docker host doesnt exist...

I think it would worth it if minikube check for multiple contexts in the begining...where we Cache the docker system info we can cache how many Contexts it has and keep it for each minikube run so we if there are Multiple Context we can make a note of it in the Logs and make it a possible source of error

medyagh (changes requested):

cmd/minikube/main.go
medyagh: never mind I think I mis-thought

Issue #19110, opened 2024-06-20
Minikube start does not work.


romu42:
### What Happened?

Minikube will not start despite multiple attempts. 
`	[ERROR KubeletVersion]: couldn't get kubelet version: cannot execute 'kubelet --version': executable file not found in $PATH`

I have tested the recommendations in:
https://github.com/kubernetes/minikube/issues/18939
and 
https://github.com/kubernetes/minikube/issues/18884

without success.

### Attach the log file

[logs.txt](https://github.com/user-attachments/files/15912852/logs.txt)

### Operating System

macOS (Default)

### Driver

Docker

drozd-A:
I am in the same boat. Followed the suggested issues and still unable to start mini-kube.

spowelljr:
Hi @romu42, thanks for reporting your issue with minikube. We have seen this occasionally but have so far been unable to reproduce this ourselves. The current theory is that the `kubectl` binary is either missing from the cluster, doesn't have executable permissions, or somehow makes it's way onto the machine too late in the start process.

After you attempt to start minikube and it fails, could you run `minikube ssh` it should hopefully allow you to do so, then could you run `ls -la /var/lib/minikube/binaries/v1.30.0` and upload the output. That will help us see if our theory is correct.

romu42:
ran `minikube delete --all --purge` ;  `minikube start` . 
and then:

```
❯ minikube ssh
docker@minikube:~$ ls -la /var/lib/minikube/binaries/v1.30.0
total 190780
drwxr-xr-x 2 root root     4096 Apr 18 05:50 .
drwxr-xr-x 3 root root     4096 Apr 18 05:50 ..
-rwxr-xr-x 1 1001 1001 48955544 Apr 18 05:50 kubeadm
-rwxr-xr-x 1 1001 1001 49938584 Apr 18 05:50 kubectl
-rwxr-xr-x 1 1001 1001 96446456 Apr 18 05:50 kubelet
docker@minikube:~$
```

spowelljr:
Interesting, so the binaries are there, and that correlates to your logs:
```
I0620 13:15:06.804729   14001 ssh_runner.go:195] Run: sudo ls /var/lib/minikube/binaries/v1.30.0
I0620 13:15:06.883552   14001 binaries.go:44] Found k8s binaries, skipping transfer
```

And immediately after that the kubectl service is started:
```
[...]

Issue #18895, opened 2024-05-15
How Do You Set the QEMU Driver Flags From Minikube


Dala0:
### What Happened?

I am trying to get Minikube to start up a QEMU based x86 emulator on an ARM CPU under macOS.

It would be useful to access flags such as 'qemu-program', 'qemu-machine', and 'qemu-cpu'. I have been unable to work out how to do this from 'minikube start'.

1. Is 'minikube start' the correct place to set these flags? If so, what is the syntax?
2. If 'minikube start' is not the correct place to set the flags, where do you set them?

@afbjorklund
            

### Attach the log file

[log.txt](https://github.com/kubernetes/minikube/files/15321310/log.txt)

### Operating System

macOS (Default)

### Driver

QEMU

spowelljr:
Hi @Dala0, currently those fields are not configurable in minikube, they are automatically depending on the architecture of the machine. However allowing those field to be configurable on start using flags would be a good feature addition.

spowelljr:
I created a quick 10 minute implementation: https://github.com/kubernetes/minikube/pull/18906

nnzv:
Looking forward to seeing the new features in the next release.

Dala0:
Wow, that was quick! Thank you.

@spowelljr

afbjorklund:
Using the docker driver and qemu-user is faster:

* #16530

But hopefully there will "soon" be a new tutorial:

* #18726

medyagh:
> > It would be useful to access flags such as 'qemu-program', 'qemu-machine', and 'qemu-cpu'. I have been unable to work out how to do this from 'minikube start'.
> 
> The flags are only available in `docker-machine`, minikube hardcodes all config in the driver "registry" - no plugins
> 
> > I am trying to get Minikube to start up a QEMU based x86 emulator on an ARM CPU under macOS.
> 
> This is not something that is supported in minikube, since it is too slow to run Kubernetes under emulation

@afbjorklund does that mean we need to modify the minikube-machine org code?

afbjorklund:
> does that mean we need to modify the minikube-machine org code?

Not really, minikube has its own fork of all the drivers (e.g. "qemu2")

So there are different ways of setting the Config, whether Flags or not

https://github.com/minikube-machine/machine/blob/main/drivers/qemu/qemu.go#L199

https://github.com/kubernetes/minikube/blob/master/pkg/minikube/registry/drvs/qemu2/qemu2.go#L117

[...]

PR #19205, opened 2024-07-06
Improvements for using x86 binary on arm64 machine

spowelljr:
Related https://github.com/kubernetes/minikube/issues/19110 https://github.com/kubernetes/minikube/issues/18895 https://github.com/kubernetes/minikube/issues/12274

## Improvement 1
**Issue:**
If a user desired to run an x86 cluster on an arm64 machine, they would use an x86 minikube binary. However, due to our logic we would use an x86 ISO, but we would then download arm64 preload and K8s binaries, resulting in arch mismatches between the cluster and the K8s binaries, resulting in minikube failing to start.

**Fix:**
We now download the preloads and K8s binary strictly based on the minikube binary arch, resulting in the proper K8s binaries being downloaded.

## Improvement 2

**Issue:**
When starting minikube using an x86 binary on an arm64 machine and using the Docker driver, we would incorrectly download both the x86 and arm64 kicbase images, and use the arm64 image for running the container. This was due to running a `docker pull` without the `--platform` flag, using the host machines default arch, being arm64.

**Fix:**
We now specify the `--platform` flag on the `docker pull`, resulting in correctly only pulling the x86 kicbase image.

## Improvement 3

**Issue:**
After resolving the above two issues I was able to attempt to to start an x86 kicbase with x86 K8s binaries on an arm64 machine, this resulted in the failure `could not detect clock speed from output` from `kubelet`. And as per https://github.com/kubernetes-sigs/kind/issues/3621#issuecomment-2120918280, `Docker's "cross platform images" only use partial emulation that cannot run a workload like Kubernetes.`

**Fix:**
Hard fail if a user attempts to start an x86 cluster on arm64 with the Docker driver.

```
$ minikube-darwin-amd64 start --driver docker
╭──────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                          │
│    You are trying to run the amd64 binary on an M1 system.                                               │
│    Please consider running the darwin/arm64 binary instead.                                              │
│    Download at https://github.com/kubernetes/minikube/releases/download/v1.33.1/minikube-darwin-arm64    │
│                                                                                                          │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────╯
😄  minikube v1.33.1 on Darwin 14.5

💣  Exiting due to PROVIDER_DOCKER_INCORRECT_ARCH: Cannot use amd64 minikube binary to start minikube cluster with Docker driver on arm64 machine
💡  Suggestion: Download and use arm64 version of the minikube binary
📘  Documentation: https://minikube.sigs.k8s.io/docs/start/
```

## Improvement 4

**Issue:**
Users have run into an issue where they try starting minikube with the x86 binary, which downloads the x86 kicbase, they then download and use the arm64 binary, but due to kicbase having identical tags uses the existing x86 version but then uses arm64 preload and K8s binaries, resulting in arch mismatches, and causing minikube to fail to start.

**Fix:**
Our previous check just did `docker images` and then checked for the existence of the expected tag. Improved the check to ensure the image matches the expected arch, if it doesn't it downloads the correct one.

### DIFF
--- a/pkg/minikube/download/image.go
+++ b/pkg/minikube/download/image.go
@@ -99,7 +99,8 @@ func ImageExistsInDaemon(img string) bool {
 	return true
 }
 
-// isImageCorrectArch is needed to resolve
+// isImageCorrectArch returns true if the image arch is the same as the binary
+// arch. This is needed to resolve
 // https://github.com/kubernetes/minikube/pull/19205
 func isImageCorrectArch(img string) (bool, error) {
 	ref, err := name.ParseReference(img)
--- a/pkg/minikube/download/image.go
+++ b/pkg/minikube/download/image.go
@@ -78,14 +78,35 @@ func ImageExistsInDaemon(img string) bool {
 	// Check if image exists locally
 	klog.Infof("Checking for %s in local docker daemon", img)
 	cmd := exec.Command("docker", "images", "--format", "{{.Repository}}:{{.Tag}}@{{.Digest}}")
-	if output, err := cmd.Output(); err == nil {
-		if strings.Contains(string(output), image.TrimDockerIO(img)) {
-			klog.Infof("Found %s in local docker daemon, skipping pull", img)
-			return true
-		}
+	output, err := cmd.Output()
+	if err != nil {
+		klog.Warningf("failed to list docker images: %v", err)
+		return false
 	}
-	// Else, pull it
-	return false
+	if !strings.Contains(string(output), image.TrimDockerIO(img)) {
+		return false
+	}
+	ref, err := name.ParseReference(img)
+	if err != nil {
+		klog.Warningf("failed to parse reference: %v", err)
+		return false
+	}
+	dImg, err := daemon.Image(ref)
+	if err != nil {
+		klog.Warningf("failed to get image from daemon: %v", err)
+		return false
+	}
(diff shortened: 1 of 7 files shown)


pkg/minikube/download/image.go
```diff
@@ -78,14 +78,35 @@ func ImageExistsInDaemon(img string) bool {
...
-			return true
-		}
+	output, err := cmd.Output()
+	if err != nil {
+		klog.Warningf("failed to list docker images: %v", err)
+		return false
 	}
-	// Else, pull it
-	return false
+	if !strings.Contains(string(output), image.TrimDockerIO(img)) {
```
medyagh, line 86: add link to the pr description and a comment why we need this block

medyagh (changes requested):

pkg/minikube/download/image.go
```diff
@@ -78,14 +78,35 @@ func ImageExistsInDaemon(img string) bool {
...
+		return false
+	}
+	ref, err := name.ParseReference(img)
```
medyagh, line 89: lets move this logic to a helper function too

medyagh (changes requested):

medyagh (approved):
one nit and then it will be good to go.

PR #19228, opened 2024-07-10
Support running x86 QEMU on arm64

spowelljr:
Closes https://github.com/kubernetes/minikube/issues/18895

Depends on https://github.com/kubernetes/minikube/pull/19205 being merged

Support running x86 QEMU cluster on arm64 with no additional flags.

### DIFF
--- a/pkg/drivers/qemu/qemu.go
+++ b/pkg/drivers/qemu/qemu.go
@@ -42,6 +42,7 @@ import (
 
 	"k8s.io/klog/v2"
 	pkgdrivers "k8s.io/minikube/pkg/drivers"
+	"k8s.io/minikube/pkg/minikube/detect"
 	"k8s.io/minikube/pkg/minikube/exit"
 	"k8s.io/minikube/pkg/minikube/firewall"
 	"k8s.io/minikube/pkg/minikube/out"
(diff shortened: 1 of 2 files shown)


pkg/drivers/qemu/qemu.go
```diff
@@ -412,14 +413,19 @@ func (d *Driver) Start() error {
 	}
 
 	// hardware acceleration is important, it increases performance by 10x
-	if runtime.GOOS == "darwin" {
+	var accel string
```
medyagh, line 416: lets move this logic to a helper function to address the complexity lint

something like SelectAccelMode()

medyagh (changes requested):

pkg/drivers/qemu/qemu.go
```diff
@@ -412,14 +413,10 @@ func (d *Driver) Start() error {
...
-		startCmd = append(startCmd,
-			"-accel", "hvf")
-	} else if _, err := os.Stat("/dev/kvm"); err == nil && runtime.GOOS == "linux" {
-		// On Linux, enable the Kernel Virtual Machine accelerator.
+	accel := hardwareAcceleration()
+	if accel != "" {
 		startCmd = append(startCmd,
-			"-accel", "kvm")
+			"-accel", accel)
 	}
```
medyagh, line 421:
```suggestion
	klog.infof("chose %s acceleration type", accell )
```

medyagh (approved):

PR #19299, opened 2024-07-18
WIP: Add QEMU functional tests

spowelljr:
Added functional test for QEMU that run on `macos-14` runners (arm64).

If using the QEMU driver on GitHub Action runners, doubled the number of attempts to get the VM IP from the `dhcpd_leases` file (from 60 to 120).

Set the `-cpu` QEMU flag to `max` & the `-accel` QEMU flag to `tcg` if running on an arm64 macOS GitHub Action runner.

### DIFF
--- a/pkg/drivers/qemu/qemu.go
+++ b/pkg/drivers/qemu/qemu.go
@@ -515,7 +515,12 @@ func (d *Driver) Start() error {
 			return nil
 		}
 		// Implement a retry loop because IP address isn't added to dhcp leases file immediately
-		for i := 0; i < 60; i++ {
+		retries := 60
+		// If running in GitHub Action runner double the retries
+		if detect.GithubActionRunner() {
+			retries = 120
+		}
+		for i := 0; i < retries; i++ {
 			log.Debugf("Attempt %d", i)
 			err = getIP()
 			if err == nil {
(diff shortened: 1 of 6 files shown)


pkg/drivers/qemu/qemu.go
```diff
@@ -545,18 +550,21 @@ func (d *Driver) Start() error {
...
 		return "kvm"
 	}
-	return ""
+	if runtime.GOOS != "darwin" {
+		return ""
+	}
+	if detect.IsAmd64M1Emulation() {
+		return "tcg"
+	}
+	if runtime.GOARCH == "arm64" && detect.GithubActionRunner() {
```
medyagh, line 563: Gemini tells me 
Flexibility: TCG allows QEMU to emulate a wide range of CPU architectures, even those not directly supported by the host system's hardware.   
 1.  Introduction — QEMU documentation 
[www.qemu.org](https://www.qemu.org/docs/master/system/introduction.html#:~:text=It%20supports%20a%20number%20of,capable%20of%20emulating%20many%20CPUs%2E)
Portability: Since TCG is software-based, QEMU can run on various platforms without requiring specialized hardware for each architecture.

so I am guessing this might mean we can use tcg on windows too

afbjorklund: Only for the slow path, which is **not** something that is recommended for running Kubernetes with

It should start with "whpx", and only use tcg if running on the wrong Architecture (like on Windows ARM)

PR #19310, opened 2024-07-21
Update required Go version in building docs

radeksm:
Sync with what is required by go.mod.


spowelljr (approved):
Thanks for the update!

Issue #19343, opened 2024-07-29
"minikube image ls" confusingly adds "docker.io/" prefix


thomasjm:
### What Happened?

When you load an image into Minikube, and then list the images using `minikube image ls`, it may sometimes add `docker.io/` or `docker.io/library/` as a prefix to your image.

I was scratching my head over why this was happening, until I found this code:

https://github.com/kubernetes/minikube/blob/master/pkg/minikube/cruntime/docker.go#L699-L719

These prefixes are added based on a pretty arbitrary criterion -- whether the image name contains a dot or not. Apparently the idea is to force a leading domain name to be on an image, so as to turn `busybox:latest` into `docker.io/library/busybox:latest`.

This behavior might have made sense when DockerHub was the only game in town, but I don't think Minikube should a) misrepresent the actual RepoTags coming from Docker or b) privilege the `docker.io` prefix in this way. After all, you can construct and load container images that have nothing to do with `docker.io` but will still get prefixed.

When the reported RepoTags are different from the one in the image I loaded, it makes it difficult to verify that an image was loaded correctly. This is especially vexing when combined with the fact that `minikube image load` sometimes [silently fails](https://github.com/kubernetes/minikube/issues/16032).

If there's a good reason for this behavior please let me know. I just experimented with removing it and it seems to work fine. PR incoming.

### Attach the log file

-

### Operating System

Ubuntu

### Driver

Docker

afbjorklund:
The implementation is buggy, but it needs to use long names for some of the runtimes to work properly.

* https://github.com/kubernetes/minikube/issues/19155

thomasjm:
Wow, thanks for the instant response!

I stripped out the `addDockerIO` function in the attached PR, and it was only used in one place -- in the `ListImages` function implementation in `cruntime/docker.go`. I assumed this function was only used to power the `minikube image ls` command. You're saying these results are also used internally?

afbjorklund:
I think it could use some refactoring (`minikube image` that is), and some better error reporting etc.

But the main idea was to provide a unified interface to all three container runtimes, for load and build.

I am not aware that other pieces of minikube should use it, but the code is quite old so that could be...

Upstream (k8s) uses this wrapper: https://github.com/distribution/reference/blob/main/normalize.go

afbjorklund:
Yeah, unfortunately Kubernetes (and e.g. CRI) does not have a position on how to specify images.

thomasjm:
/remove-lifecycle stale

[...]

PR #19344, opened 2024-07-29
Remove "docker.io/" prefixing when listing Docker images

thomasjm:
This PR removes the use of the `addDockerIO` / `TrimDockerIO` functions when working with Docker images, in order to address #19343.

closes https://github.com/kubernetes/minikube/issues/19343

### DIFF
--- a/pkg/minikube/cruntime/docker.go
+++ b/pkg/minikube/cruntime/docker.go
@@ -38,7 +38,6 @@ import (
 	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/minikube/docker"
 	"k8s.io/minikube/pkg/minikube/download"
-	"k8s.io/minikube/pkg/minikube/image"
 	"k8s.io/minikube/pkg/minikube/style"
 	"k8s.io/minikube/pkg/minikube/sysinit"
 )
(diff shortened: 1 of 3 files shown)


medyagh:
@thomasjm thank you for this PR, do you mind sharing the output an example, before/after this PR ?

thomasjm:
> do you mind sharing the output an example, before/after this PR ?

Before PR:

```
tom@desktop2 ~ » minikube image ls                                                        
...
registry.k8s.io/coredns/coredns:v1.11.1                                                                                                                                                                            
gcr.io/k8s-minikube/storage-provisioner:v5                                                                                                                                                                         
docker.io/library/busybox:1.36.1-musl
```

After PR:

```
tom@desktop2 ~ » ~/tools/minikube/out/minikube image ls                                                        
...
registry.k8s.io/coredns/coredns:v1.11.1
gcr.io/k8s-minikube/storage-provisioner:v5
busybox:1.36.1-musl
```

Notice how only the `busybox` image, which doesn't have a domain name, is affected.

I see there are some test failures, I'll look into these when I get a minute.

medyagh:
@thomasjm do you mind taking a look at the failed tests for Docker_Linux ?

[TestFunctional/parallel/ImageCommands/ImageLoadDaemon](https://storage.googleapis.com/minikube-builds/logs/19344/35558/Docker_Linux.html#fail_TestFunctional%2fparallel%2fImageCommands%2fImageLoadDaemon)

thomasjm:
Oh, I wasn't sure if this PR is still desired after the discussion on #19343.

I don't totally understand the full picture, but it sounds like it's believed that this PR might break things for more strict runtimes like containerd.

If we still want to move ahead with this I can look into it.

PR #19349, opened 2024-07-29
Addon headlamp: Update headlamp-k8s/headlamp image from v0.24.1 to v0.25.0

joaquimrocha:
Upgrade Headlamp version to 0.25.0
  cc: @joaquimrocha

### DIFF
--- a/pkg/minikube/assets/addons.go
+++ b/pkg/minikube/assets/addons.go
@@ -779,7 +779,7 @@ var Addons = map[string]*Addon{
 		MustBinAsset(addons.HeadlampAssets, "headlamp/headlamp-clusterrolebinding.yaml", vmpath.GuestAddonsDir, "headlamp-clusterrolebinding.yaml", "0640"),
 	}, false, "headlamp", "3rd party (kinvolk.io)", "yolossn", "https://minikube.sigs.k8s.io/docs/handbook/addons/headlamp/",
 		map[string]string{
-			"Headlamp": "headlamp-k8s/headlamp:v0.24.1@sha256:1c3f42aacd8eee1d3f1c63efb5a3b42da387ca1d87b77b0f486e8443201fcb37",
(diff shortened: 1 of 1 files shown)


pkg/minikube/assets/addons.go
```diff
@@ -779,7 +779,7 @@ var Addons = map[string]*Addon{
 		MustBinAsset(addons.HeadlampAssets, "headlamp/headlamp-clusterrolebinding.yaml", vmpath.GuestAddonsDir, "headlamp-clusterrolebinding.yaml", "0640"),
 	}, false, "headlamp", "3rd party (kinvolk.io)", "yolossn", "https://minikube.sigs.k8s.io/docs/handbook/addons/headlamp/",
 		map[string]string{
-			"Headlamp": "headlamp-k8s/headlamp:v0.24.1@sha256:1c3f42aacd8eee1d3f1c63efb5a3b42da387ca1d87b77b0f486e8443201fcb37",
+			"Headlamp": "headlamp-k8s/headlamp:v0.25.0@sha256:899d106eeb55b0afc4ee6e51c03bc4418de0bd0e79c39744d4d0d751aae6a971",
```
medyagh, line 782: @joaquimrocha please consider making this an automation job in github action
here is a simmilar job for kong addon.

https://github.com/kubernetes/minikube/blob/master/.github/workflows/update-kong-version.yml

you would need to make a new make target "make update-headlamp", let me know if you have any questions on it

medyagh (changes requested):

pkg/minikube/assets/addons.go
joaquimrocha: Hey @medyagh . Thanks, that sounds very cool. Our PRs are automated anyway, but I like the idea of having the action in the minikube repo instead.
If you agree, we could merge this PR so the addon is not delayed and we'll submit the new action before our next release. WDYT?

medyagh: sounds good ! btw I am curious thats good where do you run the automation ? look forward to see more contributions for you :)

medyagh (approved):

PR #19370, opened 2024-08-02
Allow privileged ports on WSL

daniel-iwaniec:
Using ports below 1024 should be allowed on WSL in the same way as on Linux.
It should be left for the OS to decide whether it is allowed, respecting Linux Capabilities (e.g. CAP_NET_BIND_SERVICE), etc.

Going through the git history I got the impression, that it was left out for WSL to err on the safe side, but I don't see any explicit reason it should stay. Correct me if I'm wrong, please.

It doesn't fix any known issue as far as I know, I just noticed it myself and decided to create a new PR

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1365,7 +1365,7 @@ func validateFlags(cmd *cobra.Command, drvName string) { //nolint:gocyclo
 	validateInsecureRegistry()
 }
 
-// validatePorts validates that the --ports are not below 1024 for the host and not outside range
+// validatePorts validates that the --ports are not outside range
 func validatePorts(ports []string) error {
 	var exposedPorts, hostPorts, portSpecs []string
 	for _, p := range ports {
@@ -1386,28 +1386,25 @@ func validatePorts(ports []string) error {
 		}
 	}
 	for _, p := range exposedPorts {
-		if err := validatePort(p, false); err != nil {
+		if err := validatePort(p); err != nil {
 			return err
 		}
 	}
 	for _, p := range hostPorts {
-		if err := validatePort(p, true); err != nil {
+		if err := validatePort(p); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
-func validatePort(port string, isHost bool) error {
+func validatePort(port string) error {
 	p, err := strconv.Atoi(port)
 	if err != nil {
 		return errors.Errorf("Sorry, one of the ports provided with --ports flag is not valid: %s", port)
 	}
 	if p > 65535 || p < 1 {
 		return errors.Errorf("Sorry, one of the ports provided with --ports flag is outside range: %s", port)
-	}
-	if isHost && detect.IsMicrosoftWSL() && p < 1024 {
-		return errors.Errorf("Sorry, you cannot use privileged ports on the host (below 1024): %s", port)
 	}
 	return nil
 }
--- a/cmd/minikube/cmd/start_test.go
+++ b/cmd/minikube/cmd/start_test.go
@@ -30,7 +30,6 @@ import (
 	cfg "k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/constants"
 	"k8s.io/minikube/pkg/minikube/cruntime"
-	"k8s.io/minikube/pkg/minikube/detect"
 	"k8s.io/minikube/pkg/minikube/driver"
 	"k8s.io/minikube/pkg/minikube/proxy"
 )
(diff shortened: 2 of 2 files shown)


daniel-iwaniec:
I tried to go through the documentation regarding the PR process and did what I could - let me know if there is anything more I can do :)

daniel-iwaniec:
I went through some of the WSL's GitHub issues and:

- here https://github.com/microsoft/WSL/issues/574 you can see that setcap was implemented
- here https://github.com/microsoft/WSL/issues/2007 you can see that people were able to use it and bind to ports lower than 1024

cmd/minikube/cmd/start.go
```diff
@@ -1386,29 +1386,26 @@ func validatePorts(ports []string) error {
...
-func validatePort(port string, isHost bool) error {
+func validatePort(port string) error {
 	p, err := strconv.Atoi(port)
 	if err != nil {
 		return errors.Errorf("Sorry, one of the ports provided with --ports flag is not valid: %s", port)
 	}
 	if p > 65535 || p < 1 {
 		return errors.Errorf("Sorry, one of the ports provided with --ports flag is outside range: %s", port)
 	}
-	if isHost && detect.IsMicrosoftWSL() && p < 1024 {
```
medyagh, line 1409: why isHost been removed from the logic ? doesnt seem related to this PR

medyagh:
@daniel-iwaniec thank you for this contribution do you mind sharing the Before/After this PR, with an example of using the WSL on windows?

cmd/minikube/cmd/start.go
daniel-iwaniec: The way I understand it, the current logic is "on WSL (and only on WSL), check if the host port is below 1024", and since the change is to allow those ports on WSL, we don't need any special checks for host ports.

Also if we remove this condition from function this argument is just not used anywhere anymore

daniel-iwaniec:
Current behavior on WSL
===
```
$ uname -a
Linux DESKTOP-2VAN5M9 5.15.153.1-microsoft-standard-WSL2 #1 SMP Fri Mar 29 23:14:13 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

$ minikube version
minikube version: v1.33.1
commit: 5883c09216182566a63dff4c326a6fc9ed2982ff

$ minikube start --ports=80:80
😄  minikube v1.33.1 on Ubuntu 22.04 (amd64)
💡  Alternatively you could install one of these drivers:
    ▪ kvm2: Not installed: exec: "virsh": executable file not found in $PATH
    ▪ qemu2: Not installed: exec: "qemu-system-x86_64": executable file not found in $PATH

```

daniel-iwaniec:
Current behavior on WSL
===
```
$ uname -a
Linux DESKTOP-2VAN5M9 5.15.153.1-microsoft-standard-WSL2 #1 SMP Fri Mar 29 23:14:13 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

$ minikube version
minikube version: v1.33.1
commit: 5883c09216182566a63dff4c326a6fc9ed2982ff

$ minikube start --ports=80:80
😄  minikube v1.33.1 on Ubuntu 22.04 (amd64)

❌  Exiting due to MK_USAGE: Sorry, you cannot use privileged ports on the host (below 1024): 80
```

daniel-iwaniec:
Expected behavior
===
Well, I obtained the output after building version with my changes, hence `-dirty` etc.
```
$ uname -a
Linux DESKTOP-2VAN5M9 5.15.153.1-microsoft-standard-WSL2 #1 SMP Fri Mar 29 23:14:13 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

$ minikube version
minikube version: v1.33.1
commit: adc0c841af400141e073e0e45061d84afa6c9617-dirty

$ minikube start --ports=80:80
😄  minikube v1.33.1 on Ubuntu 22.04 (amd64)
✨  Automatically selected the docker driver. Other choices: none, ssh
🚜  Pulling base image v0.0.44-1721902582-19326 ...
💾  Downloading Kubernetes v1.30.3 preload ...
    > preloaded-images-k8s-v18-v1...:  342.95 MiB / 342.95 MiB  100.00% 22.42 M
    > gcr.io/k8s-minikube/kicbase...:  487.17 MiB / 487.18 MiB  100.00% 22.29 M
🔥  Creating docker container (CPUs=2, Memory=2200MB) ...
🐳  Preparing Kubernetes v1.30.3 on Docker 27.1.1 ...
    ▪ Configuring RBAC rules ...
```

daniel-iwaniec:
@medyagh I replied to your comment and gave a before/after example
I'm not sure if I should do something about the failed tests or if they are just flaky
Let me know if everything's ok or if there's something else

Anyway, thanks for the quick response :)

medyagh:
thank you @daniel-iwaniec for the before/after do you mind doing an example of this on a on WSL ? I know this PR wont break that I  but I am suspecious that the code intended to be a" logical OR" and my be we meant to not allow ports bellow 1024 anywhere...it work on linux ?

daniel-iwaniec:
Unit tests should also be fixed now, although I'm not sure how to trigger them

medyagh:
> ring the Before/After this PR, with an example of using the WSL on windows?

yes I noticed that but I am wondering if this feature works on linux ?

I have a feeling this "if isHost" statement been there for a reason and maybe it was meant to not allow it on Linux either

daniel-iwaniec:
I'm using Minikube like that on Linux all the time for years, without any issues. I only stumbled upon this issue when I started working with developers using WSL, which prompted me to create this PR. So yeah, it works fine on Linux 😄 

Also, I already used binary with these changes on a few machines, and it is working fine

cmd/minikube/cmd/start_test.go
```diff
@@ -581,171 +580,98 @@ func TestIsTwoDigitSemver(t *testing.T) {
...
 		},
 		{
-			isTarget: isMicrosoftWSL,
-			ports:    []string{"80:80"},
-			errorMsg: "Sorry, you cannot use privileged ports on the host (below 1024): 80",
-		},
-		{
-			isTarget: isMicrosoftWSL,
-			ports:    []string{"1023-1025:8023-8025"},
-			errorMsg: "Sorry, you cannot use privileged ports on the host (below 1024): 1023",
```
medyagh, line 646: thank you @daniel-iwaniec for fixing the test, the Pr looks good just one last thing, can you plz emulate a secnario that the OS does NOT allow or can not create port bellow 1024 and see if minikube failed gracefuly ?

my fear is that for some new users this will blow with a large stack trace error if their OS does not allow this, it is good that we make sure minikube fails with a Nice one liner error that their OS didnt support it

medyagh (changes requested):

cmd/minikube/cmd/start_test.go
daniel-iwaniec: I would have to really go out of my way to make it not work :)

* This option (--ports) applies only to **docker** and **podman**
  * **docker** and **podman** use `CAP_NET_BIND_SERVICE` by default
  * which means binding ports below 1024 should work by default
* In specific case of **WSL**, I confirmed it works both with
  * docker engine
  * docker desktop
    * which means you could even install docker desktop on windows, then install minikube using WSL, and it still would work

PR #19395, opened 2024-08-08
site: add docker.io to all echo-server examples

medyagh:
on crio not specifying the FQDN for the image name makes crio not be able to pull the image, so better to have it FQDN everywhere.

on crio
### fails
$ kubectl create deployment hello-minikube --image=kicbase/echo-server:1.0
deployment.apps/hello-minikube created
```
default       hello-minikube-5c898d8489-8gh5t    0/1     ImageInspectError   0          15s
```

```
Failed to inspect image "kicbase/echo-server:1.0": rpc error: code = Unknown desc = short-name "kicbase/echo-server:1.0" did not resolve to an alias and no unqualified-search registries are defined in "/etc/containers/registries.conf"
```

### works
$ kubectl create deployment hello-minikube2 --image=docker.io/kicbase/echo-server:1.0
deployment.apps/hello-minikube2 created

### future works

we should consider adding docker.io to /etc/containers/registries.conf so other people dont have the same issue

https://github.com/kubernetes/minikube/issues/19396

### DIFF
--- a/site/content/en/docs/handbook/accessing.md
+++ b/site/content/en/docs/handbook/accessing.md
@@ -41,7 +41,7 @@ Services of type `NodePort` can be exposed via the `minikube service <service-na
 1. Create a Kubernetes deployment
 
     ```shell
-    kubectl create deployment hello-minikube1 --image=kicbase/echo-server:1.0
+    kubectl create deployment hello-minikube1 --image=docker.io/kicbase/echo-server:1.0
     ```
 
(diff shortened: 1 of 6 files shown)


afbjorklund:
Sounds like a broken cri-o configuration

spowelljr (approved):

afbjorklund (changes requested):
This change should not be needed, afaik kubernetes _requires_ docker.io shortnames?

PR #19423, opened 2024-08-13
Add new macOS vfkit driver, like hyperkit and qemu

### LINKED ISSUE #12826: Minikube with driver "new Virtualization Framework" on Apple M1
As more and more people are thinking of moving to Minikube to replace "Docker For Mac" some work to enable virtualization on it should be a priority to get them onboard. Hyperkit being deprecated going with the **new Virtualization Framework** which uses higher API and is supposed to be easier to use probably makes more sense

See https://github.com/Code-Hex/vz

afbjorklund:
It uses the new Virtualization.framework from macOS 11, instead of the older Hypervisor.framework (hvf) in QEMU.

Closes #12826

----

The "hyperkit" binary was bundled with **Docker for Mac**, but is available stand-alone:

https://github.com/moby/hyperkit

The "vfkit" binary is bundled with **Podman Desktop** (macOS), and is available stand-alone:

https://github.com/crc-org/vfkit

### DIFF
--- a/pkg/drivers/vfkit/vfkit.go
+++ b/pkg/drivers/vfkit/vfkit.go
@@ -246,6 +246,11 @@ func (d *Driver) Start() error {
 	log.Debugf("executing: vfkit %s", strings.Join(startCmd, " "))
 	os.Remove(d.sockfilePath())
 	cmd := exec.Command("vfkit", startCmd...)
+
+	// Create vfkit in a new process group, so minikube caller can use killpg
+	// to terminate the entire process group without harming the vfkit process.
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+
 	if err := cmd.Start(); err != nil {
 		return err
 	}
--- a/pkg/drivers/vfkit/vfkit.go
+++ b/pkg/drivers/vfkit/vfkit.go
@@ -1,0 +1,513 @@
+//go:build darwin
+
+/*
+Copyright 2024 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vfkit
+
+import (
+	"archive/tar"
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/docker/machine/libmachine/drivers"
+	"github.com/docker/machine/libmachine/log"
+	"github.com/docker/machine/libmachine/mcnutils"
+	"github.com/docker/machine/libmachine/ssh"
+	"github.com/docker/machine/libmachine/state"
+	"github.com/pkg/errors"
+
+	pkgdrivers "k8s.io/minikube/pkg/drivers"
+)
+
+const (
+	isoFilename    = "boot2docker.iso"
+	pidFileName    = "vfkit.pid"
+	sockFilename   = "vfkit.sock"
+	defaultSSHUser = "docker"
+)
+
+// Driver is the machine driver for vfkit (Virtualization.framework)
+type Driver struct {
+	*drivers.BaseDriver
+	*pkgdrivers.CommonDriver
+	Boot2DockerURL string
+	DiskSize       int
+	CPU            int
+	Memory         int
+	Cmdline        string
+	MACAddress     string
+	ExtraDisks     int
+}
+
+func NewDriver(hostName, storePath string) drivers.Driver {
+	return &Driver{
+		BaseDriver: &drivers.BaseDriver{
+			SSHUser:     defaultSSHUser,
+			MachineName: hostName,
+			StorePath:   storePath,
+		},
+		CommonDriver: &pkgdrivers.CommonDriver{},
+	}
+}
+
+func (d *Driver) PreCreateCheck() error {
+	return nil
+}
+
+func (d *Driver) GetMachineName() string {
+	return d.MachineName
+}
+
+func (d *Driver) DriverName() string {
+	return "vfkit"
+}
+
+func (d *Driver) GetSSHHostname() (string, error) {
+	return d.IPAddress, nil
+}
+
+func (d *Driver) GetSSHKeyPath() string {
+	return d.ResolveStorePath("id_rsa")
+}
+
+func (d *Driver) GetSSHPort() (int, error) {
+	if d.SSHPort == 0 {
+		d.SSHPort = 22
+	}
+	return d.SSHPort, nil
+}
+
+func (d *Driver) GetSSHUsername() string {
+	if d.SSHUser == "" {
+		d.SSHUser = defaultSSHUser
+	}
+
+	return d.SSHUser
+}
+
+func (d *Driver) GetURL() (string, error) {
+	if _, err := os.Stat(d.pidfilePath()); err != nil {
+		return "", nil
+	}
+	ip, err := d.GetIP()
+	if err != nil {
+		log.Warnf("Failed to get IP: %v", err)
+		return "", err
+	}
+	if ip == "" {
+		return "", nil
+	}
+	return fmt.Sprintf("tcp://%s:2376", ip), nil
+}
+
+func (d *Driver) GetIP() (string, error) {
+	return d.IPAddress, nil
+}
+
+func checkPid(pid int) error {
+	process, err := os.FindProcess(pid)
+	if err != nil {
+		return err
+	}
+	return process.Signal(syscall.Signal(0))
+}
+
+func (d *Driver) GetState() (state.State, error) {
+	if _, err := os.Stat(d.pidfilePath()); err != nil {
+		return state.Stopped, nil
+	}
+	p, err := os.ReadFile(d.pidfilePath())
+	if err != nil {
+		return state.Error, err
+	}
+	pid, err := strconv.Atoi(strings.TrimSpace(string(p)))
+	if err != nil {
+		return state.Error, err
+	}
+	if err := checkPid(pid); err != nil {
+		// No pid, remove pidfile
+		os.Remove(d.pidfilePath())
+		return state.Stopped, nil
+	}
+	ret, err := d.GetVFKitState()
+	if err != nil {
+		return state.Error, err
+	}
+	switch ret {
+	case "running", "VirtualMachineStateRunning":
+		return state.Running, nil
+	case "stopped", "VirtualMachineStateStopped":
+		return state.Stopped, nil
+	}
+	return state.None, nil
+}
+
+func (d *Driver) Create() error {
+	var err error
+	if d.SSHPort, err = d.GetSSHPort(); err != nil {
+		return err
+	}
+	b2dutils := mcnutils.NewB2dUtils(d.StorePath)
+	if err := b2dutils.CopyIsoToMachineDir(d.Boot2DockerURL, d.MachineName); err != nil {
+		return err
+	}
+	isoPath := d.ResolveStorePath(isoFilename)
+
+	log.Info("Extracting Kernel...")
+	if err := d.extractKernel(isoPath); err != nil {
+		return err
+	}
+
+	log.Info("Creating SSH key...")
+	if err := ssh.GenerateSSHKey(d.sshKeyPath()); err != nil {
+		return err
+	}
+
+	log.Info("Creating Disk image...")
+	if err := d.generateDiskImage(d.DiskSize); err != nil {
+		return err
+	}
+
+	if d.ExtraDisks > 0 {
+		log.Info("Creating extra disk images...")
+		for i := 0; i < d.ExtraDisks; i++ {
+			path := pkgdrivers.ExtraDiskPath(d.BaseDriver, i)
+			if err := pkgdrivers.CreateRawDisk(path, d.DiskSize); err != nil {
+				return err
+			}
+		}
+	}
+
+	log.Info("Starting vfkit VM...")
+	return d.Start()
+}
+
+func (d *Driver) Start() error {
+	machineDir := filepath.Join(d.StorePath, "machines", d.GetMachineName())
+
+	var startCmd []string
+
+	startCmd = append(startCmd,
+		"--memory", fmt.Sprintf("%d", d.Memory),
+		"--cpus", fmt.Sprintf("%d", d.CPU),
+		"--restful-uri", fmt.Sprintf("unix://%s", d.sockfilePath()))
+	var isoPath = filepath.Join(machineDir, isoFilename)
+	startCmd = append(startCmd,
+		"--device", fmt.Sprintf("virtio-blk,path=%s", isoPath))
+
+	var mac = d.MACAddress
+	startCmd = append(startCmd,
+		"--device", fmt.Sprintf("virtio-net,nat,mac=%s", mac))
+
+	startCmd = append(startCmd,
+		"--device", "virtio-rng")
+
+	startCmd = append(startCmd,
+		"--kernel", d.ResolveStorePath("bzimage"))
+	startCmd = append(startCmd,
+		"--kernel-cmdline", d.Cmdline)
+	startCmd = append(startCmd,
+		"--initrd", d.ResolveStorePath("initrd"))
+
+	for i := 0; i < d.ExtraDisks; i++ {
+		startCmd = append(startCmd,
+			"--device", fmt.Sprintf("virtio-blk,path=%s", pkgdrivers.ExtraDiskPath(d.BaseDriver, i)))
+	}
+
+	startCmd = append(startCmd,
+		"--device", fmt.Sprintf("virtio-blk,path=%s", d.diskPath()))
+
+	log.Debugf("executing: vfkit %s", strings.Join(startCmd, " "))
+	os.Remove(d.sockfilePath())
+	cmd := exec.Command("vfkit", startCmd...)
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	pid := cmd.Process.Pid
+	if err := os.WriteFile(d.pidfilePath(), []byte(fmt.Sprintf("%v", pid)), 0600); err != nil {
+		return err
+	}
+
+	// Need to strip 0's
+	mac = pkgdrivers.TrimMacAddress(mac)
+	if err := d.setupIP(mac); err != nil {
+		return err
+	}
+
+	log.Infof("Waiting for VM to start (ssh -p %d docker@%s)...", d.SSHPort, d.IPAddress)
+
+	return WaitForTCPWithDelay(fmt.Sprintf("%s:%d", d.IPAddress, d.SSHPort), time.Second)
+}
+
+func (d *Driver) setupIP(mac string) error {
+	var err error
+
+	getIP := func() error {
+		d.IPAddress, err = pkgdrivers.GetIPAddressByMACAddress(mac)
+		if err != nil {
+			return errors.Wrap(err, "failed to get IP address")
+		}
+		return nil
+	}
+
+	// Implement a retry loop because IP address isn't added to dhcp leases file immediately
+	for i := 0; i < 60; i++ {
+		log.Debugf("Attempt %d", i)
+		err = getIP()
+		if err == nil {
+			break
+		}
+		time.Sleep(2 * time.Second)
+	}
+
+	if err != nil {
+		return fmt.Errorf("IP address never found in dhcp leases file %v", err)
+	}
+	log.Debugf("IP: %s", d.IPAddress)
+
+	return nil
+}
+
+func (d *Driver) Stop() error {
+	if err := d.SetVFKitState("HardStop"); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d *Driver) Remove() error {
+	s, err := d.GetState()
+	if err != nil {
+		return errors.Wrap(err, "get state")
+	}
+	if s == state.Running {
+		if err := d.Kill(); err != nil {
+			return errors.Wrap(err, "kill")
+		}
+	}
+	if s != state.Stopped {
+		if err := d.SetVFKitState("Stop"); err != nil {
+			return errors.Wrap(err, "quit")
+		}
+	}
+	return nil
+}
+
+func (d *Driver) Restart() error {
+	s, err := d.GetState()
+	if err != nil {
+		return err
+	}
+
+	if s == state.Running {
+		if err := d.Stop(); err != nil {
+			return err
+		}
+	}
+	return d.Start()
+}
+
+func (d *Driver) extractKernel(isoPath string) error {
+	for _, f := range []struct {
+		pathInIso string
+		destPath  string
+	}{
+		{"/boot/bzimage", "bzimage"},
+		{"/boot/initrd", "initrd"},
+	} {
+		fullDestPath := d.ResolveStorePath(f.destPath)
+		if err := pkgdrivers.ExtractFile(isoPath, f.pathInIso, fullDestPath); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *Driver) Kill() error {
+	if err := d.SetVFKitState("HardStop"); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d *Driver) StartDocker() error {
+	return fmt.Errorf("hosts without a driver cannot start docker")
+}
+
+func (d *Driver) StopDocker() error {
+	return fmt.Errorf("hosts without a driver cannot stop docker")
+}
+
+func (d *Driver) GetDockerConfigDir() string {
+	return ""
+}
+
+func (d *Driver) Upgrade() error {
+	return fmt.Errorf("hosts without a driver cannot be upgraded")
+}
+
+func (d *Driver) sshKeyPath() string {
+	machineDir := filepath.Join(d.StorePath, "machines", d.GetMachineName())
+	return filepath.Join(machineDir, "id_rsa")
+}
+
+func (d *Driver) publicSSHKeyPath() string {
+	return d.sshKeyPath() + ".pub"
+}
+
+func (d *Driver) diskPath() string {
+	machineDir := filepath.Join(d.StorePath, "machines", d.GetMachineName())
+	return filepath.Join(machineDir, "disk.img")
+}
+
+func (d *Driver) sockfilePath() string {
+	machineDir := filepath.Join(d.StorePath, "machines", d.GetMachineName())
+	return filepath.Join(machineDir, sockFilename)
+}
+
+func (d *Driver) pidfilePath() string {
+	machineDir := filepath.Join(d.StorePath, "machines", d.GetMachineName())
+	return filepath.Join(machineDir, pidFileName)
+}
+
+// Make a boot2docker VM disk image.
+func (d *Driver) generateDiskImage(size int) error {
+	log.Debugf("Creating %d MB hard disk image...", size)
+
+	magicString := "boot2docker, please format-me"
+
+	buf := new(bytes.Buffer)
+	tw := tar.NewWriter(buf)
+
+	// magicString first so the automount script knows to format the disk
+	file := &tar.Header{Name: magicString, Size: int64(len(magicString))}
+	if err := tw.WriteHeader(file); err != nil {
+		return err
+	}
+	if _, err := tw.Write([]byte(magicString)); err != nil {
+		return err
+	}
+	// .ssh/key.pub => authorized_keys
+	file = &tar.Header{Name: ".ssh", Typeflag: tar.TypeDir, Mode: 0700}
+	if err := tw.WriteHeader(file); err != nil {
+		return err
+	}
+	pubKey, err := os.ReadFile(d.publicSSHKeyPath())
+	if err != nil {
+		return err
+	}
+	file = &tar.Header{Name: ".ssh/authorized_keys", Size: int64(len(pubKey)), Mode: 0644}
+	if err := tw.WriteHeader(file); err != nil {
+		return err
+	}
+	if _, err := tw.Write(pubKey); err != nil {
+		return err
+	}
+	file = &tar.Header{Name: ".ssh/authorized_keys2", Size: int64(len(pubKey)), Mode: 0644}
+	if err := tw.WriteHeader(file); err != nil {
+		return err
+	}
+	if _, err := tw.Write(pubKey); err != nil {
+		return err
+	}
+	if err := tw.Close(); err != nil {
+		return err
+	}
+	rawFile := d.diskPath()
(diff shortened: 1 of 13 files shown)


afbjorklund:
Some code is duplicated between pkg/drivers/hyperkit, pkg/drivers/qemu and pkg/drivers/vfkit.

It could be considered, to (later!) refactor this into the "common" part under pkg/drivers directly...

afbjorklund:
The "hyperkit" binary was bundled with Docker for Mac, but is available stand-alone:

https://github.com/moby/hyperkit

The "vfkit" binary is bundled with Podman Desktop (macOS), and is available stand-alone:

https://github.com/crc-org/vfkit

pkg/drivers/vfkit/iso.go
```diff
@@ -0,0 +1,65 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
```
medyagh, line 2: 2024 here and other files

medyagh (changes requested):

medyagh:
This is a very exciting PR ! thank you for thinking about it 
it is not working for me on macos M1 (arm64)
```
I0815 15:38:39.293989   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.2 HWAddress:ca:96:e9:43:b9:43 ID:1,ca:96:e9:43:b9:43 Lease:0x646e4c8e}
I0815 15:38:41.295145   63339 main.go:141] libmachine: Attempt 28
I0815 15:38:41.295195   63339 main.go:141] libmachine: Searching for 76:70:dd:1:99:e2 in /var/db/dhcpd_leases ...
I0815 15:38:41.295760   63339 main.go:141] libmachine: Found 34 entries in /var/db/dhcpd_leases!
I0815 15:38:41.295820   63339 main.go:141] libmachine: dhcp entry: {Name: IPAddress:192.168.106.2 HWAddress:7e:b:7b:f9:40:ae ID:1,7e:b:7b:f9:40:ae Lease:0x66aac1c1}
I0815 15:38:41.295843   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.34 HWAddress:8a:37:82:8d:50:f0 ID:1,8a:37:82:8d:50:f0 Lease:0x668d8331}
I0815 15:38:41.295863   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.33 HWAddress:a:94:4f:77:5f:83 ID:1,a:94:4f:77:5f:83 Lease:0x66649746}
I0815 15:38:41.295888   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.32 HWAddress:ea:cc:a6:94:f1:58 ID:1,ea:cc:a6:94:f1:58 Lease:0x66623678}
I0815 15:38:41.296047   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.31 HWAddress:f2:8f:83:8:6e:9e ID:1,f2:8f:83:8:6e:9e Lease:0x664e2690}
I0815 15:38:41.296081   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.30 HWAddress:2e:ee:ae:b:15:c5 ID:1,2e:ee:ae:b:15:c5 Lease:0x663eb6d8}
I0815 15:38:41.296119   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.29 HWAddress:6:8c:7d:21:84:7a ID:1,6:8c:7d:21:84:7a Lease:0x663eaece}
I0815 15:38:41.296142   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.28 HWAddress:a6:91:9b:ff:1a:c7 ID:1,a6:91:9b:ff:1a:c7 Lease:0x663eaddc}
I0815 15:38:41.296164   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.27 HWAddress:b6:15:58:c6:cf:c3 ID:1,b6:15:58:c6:cf:c3 Lease:0x663e9198}
I0815 15:38:41.296187   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.26 HWAddress:26:12:63:a2:cf:af ID:1,26:12:63:a2:cf:af Lease:0x6633c7b9}
I0815 15:38:41.296208   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.25 HWAddress:12:da:f7:50:c2:27 ID:1,12:da:f7:50:c2:27 Lease:0x66327ee3}
I0815 15:38:41.296230   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.24 HWAddress:82:1e:1:e2:6c:bb ID:1,82:1e:1:e2:6c:bb Lease:0x662c2929}
I0815 15:38:41.296272   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.23 HWAddress:ca:94:90:33:e1:19 ID:1,ca:94:90:33:e1:19 Lease:0x66280395}
I0815 15:38:41.296295   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.22 HWAddress:ce:1e:18:a6:2c:dc ID:1,ce:1e:18:a6:2c:dc Lease:0x6616bb16}
I0815 15:38:41.296316   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.21 HWAddress:fe:f9:df:2b:92:fd ID:1,fe:f9:df:2b:92:fd Lease:0x660dbec9}
I0815 15:38:41.296337   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.20 HWAddress:3a:5c:a4:60:23:dd ID:1,3a:5c:a4:60:23:dd Lease:0x660dbde6}
I0815 15:38:41.296359   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.19 HWAddress:9e:5:a1:b4:2:91 ID:1,9e:5:a1:b4:2:91 Lease:0x65678416}
I0815 15:38:41.296380   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.18 HWAddress:e:9d:a6:2:8:1d ID:1,e:9d:a6:2:8:1d Lease:0x65144f88}
I0815 15:38:41.296400   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.17 HWAddress:a6:2a:9b:f7:48:39 ID:1,a6:2a:9b:f7:48:39 Lease:0x65049be4}
I0815 15:38:41.296423   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.16 HWAddress:46:c9:95:80:5a:b8 ID:1,46:c9:95:80:5a:b8 Lease:0x64cb063b}
I0815 15:38:41.296444   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.15 HWAddress:6e:e7:c0:43:e5:e9 ID:1,6e:e7:c0:43:e5:e9 Lease:0x64c883fe}
I0815 15:38:41.296466   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.14 HWAddress:46:17:5b:20:e9:c1 ID:1,46:17:5b:20:e9:c1 Lease:0x64c883c3}
I0815 15:38:41.296486   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.13 HWAddress:be:4a:e3:61:22:87 ID:1,be:4a:e3:61:22:87 Lease:0x64c88180}
I0815 15:38:41.296506   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.12 HWAddress:ca:b6:91:14:9a:8f ID:1,ca:b6:91:14:9a:8f Lease:0x64c88095}
I0815 15:38:41.296527   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.11 HWAddress:92:c5:4a:cb:14:16 ID:1,92:c5:4a:cb:14:16 Lease:0x64add676}
I0815 15:38:41.296549   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.10 HWAddress:ea:dc:7e:e:52:e5 ID:1,ea:dc:7e:e:52:e5 Lease:0x649b2d96}
I0815 15:38:41.296570   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.9 HWAddress:3e:57:12:b8:5a:54 ID:1,3e:57:12:b8:5a:54 Lease:0x6491d686}
I0815 15:38:41.296591   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.8 HWAddress:82:cd:d8:88:50:40 ID:1,82:cd:d8:88:50:40 Lease:0x648c992a}
I0815 15:38:41.296612   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.7 HWAddress:fa:f3:26:f5:4:4d ID:1,fa:f3:26:f5:4:4d Lease:0x648a0ada}
I0815 15:38:41.296633   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.6 HWAddress:ae:62:32:86:d0:b ID:1,ae:62:32:86:d0:b Lease:0x648a0af2}
I0815 15:38:41.296653   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.5 HWAddress:ce:d5:f8:63:17:83 ID:1,ce:d5:f8:63:17:83 Lease:0x64821151}
I0815 15:38:41.296674   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.4 HWAddress:f6:b:8f:6e:42:1 ID:1,f6:b:8f:6e:42:1 Lease:0x646e4e12}
I0815 15:38:41.296695   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.3 HWAddress:1a:67:dd:6a:ac:b8 ID:1,1a:67:dd:6a:ac:b8 Lease:0x645a7ae8}
I0815 15:38:41.296718   63339 main.go:141] libmachine: dhcp entry: {Name:minikube IPAddress:192.168.105.2 HWAddress:ca:96:e9:43:b9:43 ID:1,ca:96:e9:43:b9:43 Lease:0x646e4c8e}
^C

```
here is the full log
[lastStart.txt](https://github.com/user-attachments/files/16630887/lastStart.txt)

medyagh:
update:
running same  known issue for Qemu fixes it https://minikube.sigs.k8s.io/docs/drivers/qemu/#socket_vmnet
```
sudo /usr/libexec/ApplicationFirewall/socketfilterfw --add /usr/libexec/bootpd
sudo /usr/libexec/ApplicationFirewall/socketfilterfw --unblock /usr/libexec/bootpd
```

medyagh (changes requested):
I think we need to add a case for cluster/ip.go
```
func HostIP(host *host.Host, clusterName string) (net.IP, error) 
```

https://github.com/medyagh/minikube/blob/1447d5b36358e3f202d17e9617d61252eb85fa8e/pkg/minikube/cluster/ip.go#L61

pkg/drivers/vfkit/iso.go
afbjorklund: We don't normally update old copyright years, but keep when it was written?

afbjorklund: i.e. the file was copied from pkg/hyperkit

afbjorklund:
> I do see one Log spam error though:
> 
> start.go:132] Unable to get host IP: HostIP not yet implemented for "vfkit" driver

These are hardcoded on driver name in minikube, so needs another case for vfkit

pkg/drivers/vfkit/iso.go
afbjorklund: Save this for the refactoring

afbjorklund:
/ok-to-test

There are no tests for vfkit, but nothing else should have broken either

pkg/drivers/vfkit/iso.go
```diff
@@ -0,0 +1,65 @@
...
+	defer dst.Close()
+
+	_, err = io.Copy(dst, f.Sys().(io.Reader))
+	return err
+}
+
+func findFile(r *iso9660.Reader, path string) (os.FileInfo, error) {
+	// Look through the ISO for a file with a matching path.
+	for f, err := r.Next(); err != io.EOF; f, err = r.Next() {
+		// For some reason file paths in the ISO sometimes contain a '.' character at the end, so strip that off.
```
spowelljr, line 59:
```suggestion
		if err != nil {
			klog.Warning(err)
			continue
		}
		// For some reason file paths in the ISO sometimes contain a '.' character at the end, so strip that off.
```
Log any errors

pkg/minikube/registry/drvs/vfkit/vfkit.go
```diff
@@ -0,0 +1,92 @@
...
+	"k8s.io/minikube/pkg/drivers/vfkit"
+	"k8s.io/minikube/pkg/minikube/config"
+	"k8s.io/minikube/pkg/minikube/download"
+	"k8s.io/minikube/pkg/minikube/driver"
+	"k8s.io/minikube/pkg/minikube/localpath"
+	"k8s.io/minikube/pkg/minikube/registry"
+)
+
+const (
+	docURL = "https://minikube.sigs.k8s.io/docs/reference/drivers/vfkit/"
```
spowelljr, line 37: Is this doc coming in a follow up PR?

spowelljr:
Would be great to add functional and integration tests for this in a follow up PR so we can find any bugs and accelerate this out of experimental

pkg/drivers/vfkit/vfkit.go
```diff
@@ -0,0 +1,513 @@
...
+	log.Debugf("executing: vfkit %s", strings.Join(startCmd, " "))
+	os.Remove(d.sockfilePath())
+	cmd := exec.Command("vfkit", startCmd...)
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	pid := cmd.Process.Pid
+	if err := os.WriteFile(d.pidfilePath(), []byte(fmt.Sprintf("%v", pid)), 0600); err != nil {
+		return err
+	}
```
nirs, line 261: Creating the pid file for vfkit is a bit racy, if we are killed after starting vfkit but before we create the pid file, we will never terminate vfkit.

I think we need to move this vfkit - it should provide a --pidfile option. It goes with the --daemonize option we also need.

nirs (commented):
Thanks @afbjorklund, this is awesome!

I'm testing this now with drenv - it works great except one issue with the way the vfkit "daemon" is running.

The missing stuff for our use case is enabling rosetta, since we don't have arm64 images for OCM yet.

afbjorklund:
I would like the refactoring of the hyperkit driver, including the iso "ExtractFile" to go in a separate PR.

But that can be done first, if you don't want the technical debt of simply copying the files to this driver.

```
diff -rs pkg/drivers/hyperkit/iso.go pkg/drivers/vfkit/iso.go
17c17
< package hyperkit
---
> package vfkit
diff -rs pkg/drivers/hyperkit/iso_test.go pkg/drivers/vfkit/iso_test.go
17c17
< package hyperkit
---
> package vfkit
Files pkg/drivers/hyperkit/iso_test.iso and pkg/drivers/vfkit/iso_test.iso are identical
```

It (extracting kernel) is only needed because the new "bootloader", that  didn't work with minikube.iso

With the EFI bootloader, it is _supposed_ to be able to read /boot/bzimage and /boot/initrd (on macOS 13)

----

The pidfile handling was mostly a quick hack when the `--pidfile`/`--daemonize` was missing (unlike qemu)

Those features can be driven to the CRC organization which is handling the "vfkit" (not Podman, actually)

https://github.com/crc-org/vfkit/issues

It is mostly developed by one developer

pkg/drivers/vfkit/vfkit.go
```diff
@@ -0,0 +1,513 @@
...
+	startCmd = append(startCmd,
+		"--initrd", d.ResolveStorePath("initrd"))
+
+	for i := 0; i < d.ExtraDisks; i++ {
+		startCmd = append(startCmd,
+			"--device", fmt.Sprintf("virtio-blk,path=%s", pkgdrivers.ExtraDiskPath(d.BaseDriver, i)))
+	}
+
+	startCmd = append(startCmd,
+		"--device", fmt.Sprintf("virtio-blk,path=%s", d.diskPath()))
```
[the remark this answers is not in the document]
afbjorklund: Most likely the order was just copied from another driver, whether it was hyperkit or qemu

pkg/drivers/vfkit/vfkit.go
```diff
@@ -0,0 +1,513 @@
...
+	}
+
+	if d.ExtraDisks > 0 {
+		log.Info("Creating extra disk images...")
+		for i := 0; i < d.ExtraDisks; i++ {
+			path := pkgdrivers.ExtraDiskPath(d.BaseDriver, i)
+			if err := pkgdrivers.CreateRawDisk(path, d.DiskSize); err != nil {
+				return err
+			}
+		}
```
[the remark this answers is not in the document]
afbjorklund: Dang it, another hardcoded driver name switch

pkg/drivers/vfkit/vfkit.go
```diff
@@ -0,0 +1,513 @@
...
+		// No pid, remove pidfile
+		os.Remove(d.pidfilePath())
+		return state.Stopped, nil
+	}
+	ret, err := d.GetVFKitState()
+	if err != nil {
+		return state.Error, err
+	}
+	switch ret {
+	case "running", "VirtualMachineStateRunning":
```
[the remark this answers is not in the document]
afbjorklund: Actually we can't, but the return value is not documented so it might change

pkg/drivers/vfkit/vfkit.go
```diff
@@ -0,0 +1,513 @@
...
+	p, err := os.ReadFile(d.pidfilePath())
+	if err != nil {
+		return state.Error, err
+	}
+	pid, err := strconv.Atoi(strings.TrimSpace(string(p)))
+	if err != nil {
+		return state.Error, err
+	}
+	if err := checkPid(pid); err != nil {
+		// No pid, remove pidfile
```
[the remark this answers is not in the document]
afbjorklund: This probably came straight from qemu

pkg/drivers/vfkit/vfkit.go
```diff
@@ -286,12 +290,23 @@ func (d *Driver) setupIP(mac string) error {
...
+	if unblockErr := firewall.UnblockBootpd(); unblockErr != nil {
+		klog.Errorf("failed unblocking bootpd from firewall: %v", unblockErr)
+		exit.Error(reason.IfBootpdFirewall, "ip not found", err)
+	}
+	out.Styled(style.Restarting, "Successfully unblocked bootpd process from firewall, retrying")
+	return fmt.Errorf("ip not found: %v", err)
+}
 
-	return nil
+func isBootpdError(err error) bool {
```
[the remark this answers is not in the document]
afbjorklund: Should probably go into common, to be shared with qemu

pkg/minikube/registry/drvs/vfkit/vfkit.go
afbjorklund: I can add a placeholder to this PR

pkg/minikube/cluster/ip.go
```diff
@@ -139,6 +139,10 @@ func HostIP(host *host.Host, clusterName string) (net.IP, error) {
 			return []byte{}, errors.Wrap(err, "Error converting VM IP address to IPv4 address")
 		}
 		return net.IPv4(vmIP[0], vmIP[1], vmIP[2], byte(1)), nil
+	case driver.VFKit:
+		vmIPString, _ := host.Driver.GetIP()
+		gatewayIPString := vmIPString[:strings.LastIndex(vmIPString, ".")+1] + "1"
+		return net.ParseIP(gatewayIPString), nil
```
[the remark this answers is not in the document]
afbjorklund: This code should really be handled by the driver directly, instead of this switch case poking at driver internals

Notice how hard it is to add extra networking options to the driver, without fiddling around in minikube code

Didn't look into what part of "vanilla" uses HostIP

But using a helper function for hyperkit sounds fair.

pkg/drivers/vfkit/vfkit.go
```diff
@@ -0,0 +1,513 @@
...
+			"--device", fmt.Sprintf("virtio-blk,path=%s", pkgdrivers.ExtraDiskPath(d.BaseDriver, i)))
+	}
+
+	startCmd = append(startCmd,
+		"--device", fmt.Sprintf("virtio-blk,path=%s", d.diskPath()))
+
+	log.Debugf("executing: vfkit %s", strings.Join(startCmd, " "))
+	os.Remove(d.sockfilePath())
+	cmd := exec.Command("vfkit", startCmd...)
+	if err := cmd.Start(); err != nil {
```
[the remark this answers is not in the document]
afbjorklund: The proposed changed looks good, I can cherry-pick it to this PR

afbjorklund:
Note that the "[libmachine](https://github.com/machine-drivers/docker.github.io/blob/173d3c65f8e7df2a8c0323594419c18086fc3a30/machine/DRIVER_SPEC.md)" drivers are (potentially) separate gRPC programs.

They were not _supposed_ to share too much code, beyond the libmachine API.

Now that it is all part of the monorepo (including the forked "external" drivers), there are possibilities to use more shared functions... The external binaries are now mostly used to separate runtime dependencies, like for hyperkit and kvm. For vfkit and qemu drivers, this size hit / build program is taken by the vfkit and the qemu-system program and the driver can do a simple exec (it doesn't have to link with system libraries, like you needed for xHyve and libvirt)

There are pros and cons, but there is something simple (KISS) about just forking a "VirtualBox"-esque program and let it handle all the system dependencies. Unfortunately we don't have such a universal program anymore, but c'est la vie. And in the odd case you were wondering, in the Lima framework it is the "hostagent" that is doing the system linking to `vz`. This is why it doesn't need the "vfkit" CLI, since it uses the "vz" API directly. With all upstream issues...

pkg/drivers/vfkit/vfkit.go
```diff
@@ -286,12 +290,23 @@ func (d *Driver) setupIP(mac string) error {
...
+		log.Debugf("IP: %s", d.IPAddress)
+		return nil
 	}
-	log.Debugf("IP: %s", d.IPAddress)
+	if !isBootpdError(err) {
+		return errors.Wrap(err, "IP address never found in dhcp leases file")
+	}
+	if unblockErr := firewall.UnblockBootpd(); unblockErr != nil {
+		klog.Errorf("failed unblocking bootpd from firewall: %v", unblockErr)
+		exit.Error(reason.IfBootpdFirewall, "ip not found", err)
```
nirs, line 302: This uses sudo with a password, blocking minikube for ever when running it by another program.

I don't think we should use such helpers. If minikube needs some setup as root, it should be done outside of minikube. If we want to make this easy, we cna provide a minikube setup command that will use sudo.

Another minor issue, this shows a bogus message about socket_vmnet which is not used by vfkit.

When this does not work, (e.g. I tried to change the virtio-net device to use gvproxy), this ends in very bad user experience - minikube blocks for long time on sudo, then try to restart the vm, finaly it fail after many minutes. It is better to do less and fail fast instead of trying to fix something we are not sure about.

I guess the same issue exist with qemu driver using socket_vmnet, so this is was not introduced by this driver.

afbjorklund: It was added (from qemu) after feedback about not working (dhcp not found):
https://github.com/kubernetes/minikube/pull/19423#issuecomment-2292431081

afbjorklund: I'm not a big fan of the "retry" myself, I would prefer it to just fail

pkg/drivers/vfkit/iso.go
afbjorklund: I would like this change to go into hyperkit driver first

afbjorklund:
The driver probably needs a "Network" setting, in preparation for future options such as vmnet or
gvproxy. Even if it starts out with only one option

pkg/drivers/vfkit/vfkit.go
```diff
@@ -0,0 +1,513 @@
...
+
+func (d *Driver) GetIP() (string, error) {
+	return d.IPAddress, nil
+}
+
+func checkPid(pid int) error {
+	process, err := os.FindProcess(pid)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
cfergeau: `crc` also has some code doing this, https://github.com/crc-org/crc/blob/7b042f76b82b854aabe3e25bc2831ffab987eedb/pkg/drivers/vfkit/driver_darwin.go#L397-L435

pkg/drivers/vfkit/vfkit.go
```diff
@@ -0,0 +1,513 @@
...
+	return d.Start()
+}
+
+func (d *Driver) extractKernel(isoPath string) error {
+	for _, f := range []struct {
+		pathInIso string
+		destPath  string
+	}{
+		{"/boot/bzimage", "bzimage"},
```
cfergeau, line 361: On arm64, vfkit requires an uncompressed kernel, I don't know what format `bzimage` is in.
Also be aware that there were some issues with direct kernel boot both on amd64 and arm64 on macOS 12 https://github.com/crc-org/vfkit/issues/11

If you can require macOS 13+, I'd recommend using uefi boot if possible, no need to extract kernel/initrd/...

pkg/drivers/vfkit/vfkit.go
cfergeau: Another race which can happen is that `vfkit` can exit after `cmd.Start()` succeeds, for example if Apple's virtualization framework fails to start the VM and returns an error. If this happens, you'd have a pid file but no corresponding process.
I added this code in vfkit's test suite to catch this https://github.com/crc-org/vfkit/commit/a38d84c167f97a520f859e17283c40c299e68e03
It probably would be useful to have a way to tell vfkit "do not exit by yourself, I'll explicitly stop/kill/.. you when I no longer need your process to be around"

pkg/drivers/vfkit/vfkit.go
```diff
@@ -0,0 +1,513 @@
...
+}
+
+func (d *Driver) pidfilePath() string {
+	machineDir := filepath.Join(d.StorePath, "machines", d.GetMachineName())
+	return filepath.Join(machineDir, pidFileName)
+}
+
+// Make a boot2docker VM disk image.
+func (d *Driver) generateDiskImage(size int) error {
+	log.Debugf("Creating %d MB hard disk image...", size)
```
cfergeau, line 421: Nit: the generated image will be `size` MiB, not MB.

afbjorklund: It wasn't so important back then (but it was always MiB)

Hopefully this can be decoupled at some future point...

https://github.com/minikube-machine/machine/blob/main/libmachine/mcnutils/b2d.go#L492

I think it's only doing some qemu-img convert / truncate?

afbjorklund: e.g.

```go
// ConvertMBToBytes converts MB to bytes
func ConvertMBToBytes(mbSize int) int64 {
        return int64(mbSize) * units.MiB
}
```

pkg/drivers/vfkit/vfkit.go
afbjorklund: For some reason it worked fine with boot2docker.iso, but failed to boot with minikube.iso

It might be the extra copy of the kernel, so might require some ISO cleanup before it works?

Dropping support for macOS 12 is not so much of an issue, I think? Had to upgrade anyway

afbjorklund: The kernel was _supposed_ to be compressed with lz4 (at least so I thought), but seems uncompressed

Could be some old bug with the aarch64 iso build perhaps, when it is copying the wrong file or something

afbjorklund: > Also be aware that there were some issues with direct kernel boot both on amd64 and arm64 on macOS 12 https://github.com/crc-org/vfkit/issues/11

Minikube still uses kernel 5.10, so that should be OK. Perhaps going to upgrade to 5.15, but still (< 5.18)

nirs: I think for new code we can assume current macOS version as requirement. The only reason to run older macOS is old hardware that cannot be upgraded (6+ years old), and such hardware is typically not usable for running minikube with virtual machines.

pkg/drivers/vfkit/vfkit.go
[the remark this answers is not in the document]
afbjorklund: The -pidfile seems simpler

pkg/minikube/cluster/ip.go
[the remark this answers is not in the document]
medyagh: I agree, i would love to see a follow up refactor if someone interested, to move this logic to the driver code not in minikube code

afbjorklund:
> Would be great to add functional and integration tests for this in a follow up PR so we can find any bugs and accelerate this out of experimental

Added the standard "hack" integration script, not sure if there is anything more that needs to be done for testing?

medyagh:
@afbjorklund https://github.com/kubernetes/minikube/pull/19468 is merged and could be pull into this PR

site/content/en/docs/drivers/vfkit.md
```diff
@@ -0,0 +1,33 @@
...
+title: "vfkit"
+weight: 2
+aliases:
+    - /docs/reference/drivers/vfkit
+---
+
+## Overview
+
+[VFKit](https://github.com/crc-org/vfkit) is an open-source program for macOS virtualization, optimized for lightweight virtual machines and container deployment.
+
```
spowelljr, line 11: To stay in line with the rest of the driver pages would be good to have a `Usage` section. I tried to do a suggestion comment but the closing backticks prevent it.

```
### Usage

```shell
minikube start --driver qemu --network socket_vmnet
```

site/content/en/docs/drivers/vfkit.md
```diff
@@ -0,0 +1,33 @@
...
+### Other
+
+* [Full list of open 'vfkit' driver issues](https://github.com/kubernetes/minikube/labels/co%2Fvfkit)
+
+## Troubleshooting
+
+### Run with logs
+
+Run `minikube start --alsologtostderr -v=7` to debug crashes
```
spowelljr, line 22:
```suggestion
Run `minikube start --driver vfkit --alsologtostderr -v=7` to debug crashes
```
Since it's still in experimental need to explicitly pass the flag.

spowelljr:
> > Would be great to add functional and integration tests for this in a follow up PR so we can find any bugs and accelerate this out of experimental
> 
> Added the standard "hack" integration script, not sure if there is anything more that needs to be done for testing?
> 
> @spowelljr
> 
> maybe `hack/jenkins/minikube_set_pending.sh`

Yeah, what you have should be good for integration, might be good to add functional tests as well, but could be done in a follow up PR.

medyagh (approved):
looks good to me ! this is gonna be exciting new feature for upcoming release, thank you for this contribution.

nirs:
This driver works beautifully for running disconnected clusters. The problem is how to make it work with shared network like socket_vmnet, or maybe getting minikube or vfkit a apple.netwroking entitlement, like [UTM](https://github.com/utmapp/UTM) has, so it can use bridged network.

I'm not sure if UTM built locally can use bridged network, or only UTM from Apple app store. But it would be awesome if we can get minikube or vfkit in the app store so it can use good networking - without adding any code.

PR #19464, opened 2024-08-17
feat: allow publish and download kicbase image in github release

ComradeProgrammer:
This PR include the following things:
1. include tarball of the kicbase image in github release
2. allow minkube download the kicbase from github release page if there is no available image from any registry

### DIFF
--- a/hack/jenkins/release_build_and_upload.sh
+++ b/hack/jenkins/release_build_and_upload.sh
@@ -39,7 +39,8 @@ grep -E "^VERSION_BUILD \\?=" Makefile | grep "${VERSION_BUILD}"
 
 # Force go packages to the Jekins home directory
 export GOPATH=$HOME/go
-
+# Make sure docker is installed and configured
+./hack/jenkins/installers/check_install_docker.sh
 # Verify ISO exists
 echo "Verifying ISO exists ..."
 make verify-iso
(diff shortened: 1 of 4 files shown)


hack/jenkins/release_build_and_upload.sh
```diff
@@ -111,5 +112,20 @@ fi
 #echo "Updating Docker images ..."
 #make push-gvisor-addon-image push-storage-provisioner-manifest
 
+echo "Generating tarball for docker images"
```
spowelljr, line 115:
```suggestion
echo "Generating tarballs for kicbase images"
```
nit: docker images is too broad

pkg/minikube/download/image.go
```diff
@@ -224,6 +226,50 @@ func ImageToCache(img string) error {
...
+	out.WarningT("minikube cannot pull kicbase image from any docker registry, and is trying to download kicbase tarball from github release page via HTTP.")
+	out.Ln("The url to the tarball is %s", downloadURL)
+	out.WarningT("It's very likely that you have an internet issue. Please ensure that you can access the internet at least via HTTP, directly or with proxy. Currently your proxy configure is:")
+	envs := []string{"HTTP_PROXY", "HTTPS_PROXY", "http_proxy", "https_proxy", "ALL_PROXY", "NO_PROXY"}
+	for _, env := range envs {
+		if v := os.Getenv(env); v != "" {
+			out.Infof("{{.env}}={{.value}}", out.V{"env": env, "value": v})
+		}
+	}
+	out.Ln("")
```
spowelljr, line 262: Can you move this to before we call the function in `cache.go`, we probably want to have it where if the user sets their country to China in the flag we call this directly and don't want to output this message for them.

spowelljr (changes requested):

ComradeProgrammer:
@spowelljr  @medyagh  updated

hack/jenkins/release_build_and_upload.sh
```diff
@@ -111,5 +112,20 @@ fi
...
 #make push-gvisor-addon-image push-storage-provisioner-manifest
 
+echo "Generating tarballs for kicbase images"
+# first get the correct tag of the kic base image
+KIC_VERSION=$(grep -E "Version =" pkg/drivers/kic/types.go | cut -d \" -f 2 | cut -d "-" -f 1)
+# then generate tarballs for all achitectures
+for ARCH in "amd64" "arm64" "arm/v7" "ppc64le" "s390x" 
+do
+  SUFFIX=$(echo $ARCH | sed 's/\///g')
+  IMAGE_NAME=kicbase/stable:${KIC_VERSION}
```
medyagh, line 122: can we ensure that this image is already pushed into the registry ...otherwise it wont find it...(maybe it is somewhere up in the script that pushes the image)

pkg/minikube/download/image.go
```diff
@@ -224,6 +226,35 @@ func ImageToCache(img string) error {
 	}
 }
 
+func ImageFromGitHubToCache(img, imgVersion string) (string, error) {
```
medyagh, line 229: this needs to have a comment that says what it does... and why is it exported? is it used somwhere else

medyagh, line 229: to not be confused with github packages or githbub image registyr, we should name it that is CLEAR that we are downloading a TAR ball and then converting it to Image (which shoudl be last resort becasue it is much slower)

maybe add in the comment that the USER should consdier this in Last effort

medyagh (changes requested):

hack/jenkins/release_build_and_upload.sh
ComradeProgrammer: @spowelljr is this the correct place to generate the tarball for new kicbase image? I guess this is a part of jenkins job, and the docker image should have been pushed into registry before this?

ComradeProgrammer:
https://github.com/kubernetes/minikube/blob/0dbe96ac5411b33140c50f7278a491748785a757/go.mod#L3
Does this go version make sense? Because if you run go list -m -json all , it is shown that [k8s.io/apimachinery](http://k8s.io/apimachinery) is using go version 1.23.0, so minikube's go version should not be less than 1.23.0

Issue #15709, opened 2023-01-25
Qemu on windows


medyagh:
### What Happened?

currently we do not return the Qemu Firmware path for windows

https://github.com/kubernetes/minikube/blob/1784105c66c9060eaa447efa3dab6ac2918167ca/pkg/minikube/registry/drvs/qemu2/qemu2.go#L69

I tried to add this 
	if runtime.GOOS == "windows" {
		return "C:\\Program Files\\qemu\\share\\edk2-x86_64-code.fd", nil
	}

but my current problem is I dont have a succesfull qemu installation on windows, since the exe binary seems to return nothing but blank space

### Attach the log file

```

PS C:\Users\M\Documents\GitHub\minikube> minikube delete --all --purge
🔥  Removing C:\Users\M\.minikube\machines\minikube ...
💀  Removed all traces of the "minikube" cluster.
🔥  Successfully deleted all profiles
💀  Successfully purged minikube directory located at - [C:\Users\M\.minikube]
PS C:\Users\M\Documents\GitHub\minikube> .\out\minikube.exe start --driver=qemu2 --alsologtostderr
I0125 12:45:35.162601   10756 out.go:296] Setting OutFile to fd 104 ...
I0125 12:45:35.163602   10756 out.go:348] isatty.IsTerminal(104) = true
I0125 12:45:35.163602   10756 out.go:309] Setting ErrFile to fd 108...
I0125 12:45:35.163602   10756 out.go:348] isatty.IsTerminal(108) = true
W0125 12:45:35.178708   10756 root.go:311] Error reading config file at C:\Users\M\.minikube\config\config.json: open C:\Users\M\.minikube\config\config.json: The system cannot find the path specified.
I0125 12:45:35.184407   10756 out.go:303] Setting JSON to false
I0125 12:45:35.195447   10756 start.go:125] hostinfo: {"hostname":"DESKTOP-CNIVAJV","uptime":325076,"bootTime":1674354459,"procs":308,"os":"windows","platform":"Microsoft Windows 11 Pro","platformFamily":"Standalone Workstation","platformVersion":"10.0.22621.1105 Build 22621.1105","kernelVersion":"10.0.22621.1105 Build 22621.1105","kernelArch":"x86_64","virtualizationSystem":"","virtualizationRole":"","hostId":"05a6b6f3-ae09-43f7-9085-ff1d48d12628"}
W0125 12:45:35.195979   10756 start.go:133] gopshost.Virtualization returned error: not implemented yet
[...]

PR #19467, opened 2024-08-18
qemu: Enable hardware acceleration on Windows

afbjorklund:
And document the default "tcg" fallback accel

Issue #15709


afbjorklund:
It needs to be `-accel whpx,kernel-irqchip=off` to avoid an issue when enabling the accel

And for some reason `-cpu max` doesn't work, so it needs to be `-cpu qemu64` instead

afbjorklund:
But there are other issues too, the QMP needs to use `tcp:`  rather than `unix:`

And the APIServer tunneling through SSH seems to be failing (exit code 255)

PR #19468, opened 2024-08-18
Move ExtractFile from hyperkit to drivers

afbjorklund:
So that it can be used in other drivers too

For instance the "vfkit" driver: #19423


nirs (commented):
Looks good

medyagh (approved):

Issue #19481, opened 2024-08-19
Resolved CrashLoopBackOff Issue (driver = docker)


asusikai:
### What Happened?

I was trying to use Jenkins with Helm, but the pod kept entering a CrashLoopBackOff state. The issue was resolved by changing the permissions of the directory `/tmp/hostpath-provisioner/default/[helm-chart-dir] ` on the pod from 755 to 777.

I tried this on both Mac and Windows.

I believe this issue could be resolved by modifying the directory creation permissions from 0755 to 0777 in the following file:

https://github.com/kubernetes/minikube/blob/master/cmd/storage-provisioner/main.go

### Attach the log file

dongjun@andongjun-ui-MacBookAir ~ % helm repo add jenkinsci https://charts.jenkins.io/
"jenkinsci" has been added to your repositories
dongjun@andongjun-ui-MacBookAir ~ % helm install my-jenkins jenkinsci/jenkins --version 5.5.6
NAME: my-jenkins
LAST DEPLOYED: Mon Aug 19 16:29:37 2024
NAMESPACE: default
STATUS: deployed
REVISION: 1
NOTES:
1. Get your 'admin' user password by running:
  kubectl exec --namespace default -it svc/my-jenkins -c jenkins -- /bin/cat /run/secrets/additional/chart-admin-password && echo
2. Get the Jenkins URL to visit by running these commands in the same shell:
  echo http://127.0.0.1:8080
  kubectl --namespace default port-forward svc/my-jenkins 8080:8080

3. Login with the password from step 1 and the username: admin
4. Configure security realm and authorization strategy
5. Use Jenkins Configuration as Code by specifying configScripts in your values.yaml file, see documentation: http://127.0.0.1:8080/configuration-as-code and examples: https://github.com/jenkinsci/configuration-as-code-plugin/tree/master/demos

For more information on running Jenkins on Kubernetes, visit:
https://cloud.google.com/solutions/jenkins-on-container-engine

For more information about Jenkins Configuration as Code, visit:
https://jenkins.io/projects/jcasc/

NOTE: Consider using a custom image with pre-installed plugins
dongjun@andongjun-ui-MacBookAir ~ % kubectl get nodes
NAME            STATUS   ROLES           AGE   VERSION
kube-test       Ready    control-plane   15m   v1.28.1
kube-test-m02   Ready    <none>          14m   v1.28.1
kube-test-m03   Ready    <none>          14m   v1.28.1
[...]

PR #19482, opened 2024-08-19
chmod hostpath-provisioner to 777

asusikai:
This might resolve CrashLoopBackoff when helm chart tries to use default pv made by minikube.
fixes https://github.com/kubernetes/minikube/issues/19481

### DIFF
--- a/cmd/storage-provisioner/main.go
+++ b/cmd/storage-provisioner/main.go
@@ -29,7 +29,7 @@ var pvDir = "/tmp/hostpath-provisioner"
 
 func main() {
 	// Glog requires that /tmp exists.
-	if err := os.MkdirAll("/tmp", 0755); err != nil {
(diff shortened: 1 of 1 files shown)


medyagh:
@asusikai this would require building a new docker image for the storage provisioner, do you mind trying it out first and see if it fixes the issue ? 
you could change the code to point the new storage provisioner image

asusikai:
After checking again, I realized I made a mistake. The directory is created in the form of `/tmp/hostpath-provisioner/{namespace}/{service-name}`, and even if I change the umask or modify the group of the parent path to docker, the subdirectories continue to be created with 755 permissions. It seems that I'll need to find another way to fix this.

medyagh:
ok sounds good, please consider closing this PR if not needed anymore

Issue #19284, opened 2024-07-17
--static-ip not respected when run with --network with docker driver


lbergesio:
### What Happened?

I have a docker network previously created:

```
$ docker network inspect minikube_net
[
    {
        "Name": "minikube_net",
        "Id": "92ce6fe7b171ecdfbdec5274d619f0d652d391d2bb86460d4a5ef2d1e51957ee",
        "Created": "2024-07-17T07:46:01.295275401-07:00",
        "Scope": "local",
        "Driver": "bridge",
        "EnableIPv6": false,
        "IPAM": {
            "Driver": "default",
            "Options": {},
            "Config": [
                {
                    "Subnet": "192.168.100.0/25"
                }
            ]
        },
        "Internal": false,
        "Attachable": false,
        "Ingress": false,
        "ConfigFrom": {
            "Network": ""
        },
        "ConfigOnly": false,
        "Containers": {},
        "Options": {
            "com.docker.network.bridge.name": "br-minikube_net"
        },
        "Labels": {}
    }
]
```

If I run the following command the ip is not repected:

```
$ minikube start -p leotest --static-ip 192.168.100.20 --network minikube_net
😄  [leotest] minikube v1.33.1 on Ubuntu 22.04 (kvm/amd64)
✨  Using the docker driver based on existing profile
❗  You cannot change the static IP of an existing minikube cluster. Please first delete the cluster.
👍  Starting "leotest" primary control-plane node in "leotest" cluster
🚜  Pulling base image v0.0.44 ...
🏃  Updating the running docker "leotest" container ...
🐳  Preparing Kubernetes v1.30.0 on Docker 26.1.1 ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: storage-provisioner, default-storageclass
🏄  Done! kubectl is now configured to use "leotest" cluster and "default" namespace by default
$ minikube ip -p leotest
172.17.0.6
```

But if I do not specify the network then, a new one is created and the ip is respected:

```
$ minikube start -p leotest --static-ip 192.168.200.20
😄  [leotest] minikube v1.33.1 on Ubuntu 22.04 (kvm/amd64)
✨  Automatically selected the docker driver. Other choices: none, ssh
📌  Using Docker driver with root privileges
👍  Starting "leotest" primary control-plane node in "leotest" cluster
🚜  Pulling base image v0.0.44 ...
[...]

PR #19494, opened 2024-08-23
Fix to use --static-ip when using a docker network with --network param

deenadayalans:
Issue #19284, Added a minor check to validate and use the static ip irrespective of the gateway when choosing a network.


spowelljr:
Recreated https://github.com/kubernetes/minikube/pull/19512

Issue #19396, opened 2024-08-08
CRIO: add docker.io to the `unqualified-search-registries`


medyagh:
after adding this line
```
unqualified-search-registries = ["docker.io"] 
```
to the 
sudo cat /etc/containers/registries.conf

it works
```
docker@minikube:~$ sudo cat /etc/containers/registries.conf
# For more information on this configuration file, see containers-registries.conf(5).
#
# NOTE: RISK OF USING UNQUALIFIED IMAGE NAMES
# We recommend always using fully qualified image names including the registry
# server (full dns name), namespace, image name, and tag
# (e.g., registry.redhat.io/ubi8/ubi:latest). Pulling by digest (i.e.,
# quay.io/repository/name@digest) further eliminates the ambiguity of tags.
# When using short names, there is always an inherent risk that the image being
# pulled could be spoofed. For example, a user wants to pull an image named
# `foobar` from a registry and expects it to come from myregistry.com. If
# myregistry.com is not first in the search list, an attacker could place a
# different `foobar` image at a registry earlier in the search list. The user
# would accidentally pull and run the attacker's image and code rather than the
# intended content. We recommend only adding registries which are completely
# trusted (i.e., registries which don't allow unknown or anonymous users to
# create accounts with arbitrary names). This will prevent an image from being
# spoofed, squatted or otherwise made insecure.  If it is necessary to use one
# of these registries, it should be added at the end of the list.
#
# # An array of host[:port] registries to try when pulling an unqualified image, in order.
unqualified-search-registries = ["docker.io"]
#
# [[registry]]
# # The "prefix" field is used to choose the relevant [[registry]] TOML table;
# # (only) the TOML table with the longest match for the input image name
# # (taking into account namespace/repo/tag/digest separators) is used.
# # 
# # The prefix can also be of the form: *.example.com for wildcard subdomain
# # matching.
# #
# # If the prefix field is missing, it defaults to be the same as the "location" field.
# prefix = "example.com/foo"
#
# # If true, unencrypted HTTP as well as TLS connections with untrusted
# # certificates are allowed.
# insecure = false
#
# # If true, pulling images with matching names is forbidden.
# blocked = false
#
# # The physical location of the "prefix"-rooted namespace.
# #
# # By default, this is equal to "prefix" (in which case "prefix" can be omitted
# # and the [[registry]] TOML table can only specify "location").
# #
# # Example: Given
# #   prefix = "example.com/foo"
[...]

PR #19504, opened 2024-08-25
Kicbase: update cri-o packaging sources

JanGutter:
* cri-o packages have moved to pkgs.k8s.io: https://kubernetes.io/blog/2023/10/10/cri-o-community-package-infrastructure/
* The new packaging format depends on some k8s distro packages. This makes the assumption that the most recent stable k8s is expected to work with the most recent stable cri-o.
* Since cri-o is also installed in the Kicbase/ISO via a different method, further work might still be needed for that case: https://github.com/kubernetes/minikube/issues/18359

### DIFF
--- a/deploy/kicbase/Dockerfile
+++ b/deploy/kicbase/Dockerfile
@@ -42,7 +42,10 @@ RUN if [ "$PREBUILT_AUTO_PAUSE" != "true" ]; then cd ./cmd/auto-pause/ && go bui
 FROM ${UBUNTU_JAMMY_IMAGE} as kicbase
 
 ARG BUILDKIT_VERSION="v0.15.2"
-ARG CRIO_VERSION="1.24"
+# cri-o depends on some kubernetes packages, the latest stable series
+# is a reasonable choice.
+ARG KUBERNETES_VERSION="v1.31.0"
+ARG CRIO_VERSION="v1.30"
 ARG CRI_DOCKERD_VERSION="v0.3.15"
 ARG CRI_DOCKERD_COMMIT="c1c566e0cc84abe6972f0bf857ecd8fe306258d9"
 ARG CNI_PLUGINS_VERSION="v1.4.0"
@@ -187,17 +190,6 @@ RUN clean-install podman && \
     echo "d /run/podman 0770 root podman" > /etc/tmpfiles.d/podman.conf && \
     systemd-tmpfiles --create
 
-# install cri-o dependencies:
-RUN export ARCH=$(dpkg --print-architecture | sed 's/ppc64el/ppc64le/') && \
-    sh -c "echo 'deb https://downloadcontent.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_22.04/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list" && \
-    curl -LO https://downloadcontent.opensuse.org/repositories/devel:kubic:libcontainers:stable/xUbuntu_22.04/Release.key && \
-    apt-key add - < Release.key && \
-    if [ "$ARCH" != "ppc64le" ]; then \
-        clean-install catatonit conmon cri-tools crun; \
-    else \
-        clean-install conmon crun; \
-    fi
-
 # install containernetworking-plugins
 RUN export ARCH=$(dpkg --print-architecture | sed 's/ppc64el/ppc64le/' | sed 's/armhf/arm/') && \
     curl -LO "https://github.com/containernetworking/plugins/releases/download/${CNI_PLUGINS_VERSION}/cni-plugins-linux-$ARCH-${CNI_PLUGINS_VERSION}.tgz" && \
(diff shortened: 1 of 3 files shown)


JanGutter:
I'd be happy to chat on slack, or open an issue, if needed.

afbjorklund:
Note that the new packages are only available for "new" versions of Kubernetes.

https://kubernetes.io/blog/2023/10/10/cri-o-community-package-infrastructure/

So versions 1.24-1.27 still use the old repo, versions 1.28-1.31 use the new repo(s).

New packages contain the **same** binaries as the github releases / GCS `cri-o/artifacts`.

---

Replacing the other packages such as podman will take more effort, and to fix the bugs...

* https://github.com/kubernetes/minikube/issues/19396

afbjorklund:
> I'd be happy to chat on slack, or open an issue, if needed.

Podman needs a way out of "experimental" in the roadmap.

This sort-of involves also the CRI-O container runtime, at least as long as they share image storage
(Similar to how docker and containerd are connected by optionally using the same image storage...)

So that is something for https://minikube.sigs.k8s.io/community/

There is also some discussion in the upstream Kubernetes project SIGs, if it is "included" in the k8s release.
The moving of the packages and the release script is one step into make it more standard, but it missed 1.31.

hack/update/kubernetes_version/update_kubernetes_version.go
```diff
@@ -101,6 +101,11 @@ var (
 				`kubernetesVersion:.*`: `kubernetesVersion: {{.LatestVersionP0}}`,
 			},
 		},
+		"deploy/kicbase/Dockerfile": {
+			Replace: map[string]string{
+				`ARG KUBERNETES_VERSION=.*`: `ARG KUBERNETES_VERSION="{{.StableVersion}}"`,
```
afbjorklund, line 106: The kicbase _cannot_ have a single kubernetes version, if it is going to be reused for multiple versions.

afbjorklund:
* https://github.com/kubernetes/minikube/pull/19456
* https://github.com/kubernetes/minikube/pull/19398

hack/update/kubernetes_version/update_kubernetes_version.go
JanGutter: Would it work for more cases if this kubernetes sources.list is then replaced by later tooling, when the specific version of k8s is installed?

afbjorklund: The easiest would probably be to install the CRI adapter with the Kubernetes components (like cri-tools/cni-plugins/kubeadm/kubelet/kubectl) - possibly split it out into per container runtime, like the "preload" is now?

afbjorklund: It seems like CRI-O now uses an internal bundling of crun and conmon, so it doesn't use the system installation of those anymore,,, Otherwise such shared things could go with the OS installation still, if not k8s-version specific.

afbjorklund:
> The new packaging format depends on some k8s distro packages.

That sounds like a bug, the cri-o package should **not** have any hard dependencies towards kubernetes packages.
It does depend on CRI and CNI during runtime, but those tools do not have to be provided by the other repository.

```
Depends: systemd-sysv,systemd,iptables
Recommends: kubernetes-cni
```

> This makes the assumption that the most recent stable k8s is expected to work with the most recent stable cri-o.

That is **not** currently the case, the stable k8s was tested with a prerelease of cri-o. Users are supposed to use stable.
This means that they will (for some weeks after the release) need to run the new k8s with the old (stable) cri-o.

```
KUBERNETES_VERSION=v1.31
CRIO_VERSION=v1.30
```

hack/update/kubernetes_version/update_kubernetes_version.go
JanGutter: Heh, the yak shaving continues (the path that led me here is quite convoluted already!), I'll check out the preload alternative.

JanGutter:
> > The new packaging format depends on some k8s distro packages.
> 
> That sounds like a bug, the cri-o package should **not** have any hard dependencies towards kubernetes packages. It does depend on CRI and CNI during runtime, but those tools do not have to be provided by the other repository.

Yeah, that's probably due to my misinterpretation of the instructions at https://github.com/cri-o/packaging

> > This makes the assumption that the most recent stable k8s is expected to work with the most recent stable cri-o.
> 
> That is **not** currently the case, the stable k8s was tested with a prerelease of cri-o. Users are supposed to use stable. This means that they will (for some weeks after the release) need to run the new k8s with the old (stable) cri-o.

My assumption stemmed from the same source: it _looks_ like they list the latest stable series of k8s (1.31) plus the latest stable series of cri-o (1.30) in the example, leading to my assumption that this is expected to work.

JanGutter:
> Note that the new packages are only available for "new" versions of Kubernetes.
> 
> https://kubernetes.io/blog/2023/10/10/cri-o-community-package-infrastructure/
> 
> So versions 1.24-1.27 still use the old repo, versions 1.28-1.31 use the new repo(s).

The old repos went through a bit of instability a month or so ago - incidentally, this was one of the reasons I started going down this rabbit hole.

afbjorklund:
> The old repos went through a bit of instability a month or so ago

There has been a lot of stability issues with OBS packages over the years.

hack/update/kubernetes_version/update_kubernetes_version.go
afbjorklund: Added CRI-O to the discuss items for today's meeting, since there was already WIP for it

afbjorklund:
Added this PR for discussion in the [community](https://minikube.sigs.k8s.io/community/) meeting and the office hours, if cri-o should be prioritized...

Last time, we added improved support for Podman to the minikube [roadmap](https://minikube.sigs.k8s.io/docs/contrib/roadmap/) - maybe also add CRI-O ?

----


2021 datadog: https://www.datadoghq.com/container-report-2021/



2023 datadog: https://www.datadoghq.com/container-report/




You can clearly see cri-o being utilized more, right ? Like the Linux desktop...

https://linux.slashdot.org/story/24/08/01/2029219/linux-hits-another-desktop-market-share-record

hack/update/kubernetes_version/update_kubernetes_version.go
afbjorklund: Experimented with building static podman with Nix, similar to the cri-o packaging... ([discussion](https://github.com/containers/podman/discussions/23743))

JanGutter:
Oh man, slashdot is still going!

This whole rabbithole started when I tried to get a minimal cluster running with centos 9, podman and an ingress. The nice feature of cri-o was that I could configure registry mirrors to rewrite both docker.io and quay.io fqdn paths, so with a CI job and no manifest rewrites, I could do a test of a kubernetes operator and a number of containers.

JanGutter:
(This open issue put me on the path to podman https://github.com/moby/moby/issues/18818 ), I'll have to check if containerd's rewrite options are similar. Honestly, if containerd for rootful and rootless podman works better, I'm much less interested in getting cri-o to work.

deploy/kicbase/Dockerfile
```diff
@@ -42,7 +42,10 @@ RUN if [ "$PREBUILT_AUTO_PAUSE" != "true" ]; then cd ./cmd/auto-pause/ && go bui
 FROM ${UBUNTU_JAMMY_IMAGE} as kicbase
 
 ARG BUILDKIT_VERSION="v0.15.2"
-ARG CRIO_VERSION="1.24"
+# cri-o depends on some kubernetes packages, the latest stable series
+# is a reasonable choice.
+ARG KUBERNETES_VERSION="v1.31.0"
+ARG CRIO_VERSION="v1.30"
```
medyagh, line 48: I dont think we are ready to bump to the latest crio version yet (till we make a preload tarbal to include the critctl and other binaries)

PR #19512, opened 2024-08-26
Fix to use --static-ip when using a docker network with --network param

deenadayalansrinivasan:
Issue #19284 - Use static ip with empty gateway info. Added a minor check to validate and use the static ip irrespective of the gateway when choosing a network.


pkg/drivers/kic/kic.go
```diff
@@ -109,6 +109,9 @@ func (d *Driver) Create() error {
 	} else if gateway != nil && staticIP != "" {
 		params.Network = networkName
 		params.IP = staticIP
+	} else if gateway == nil && staticIP != "" {
```
spowelljr, line 112: Instead of adding an additional if statement, the `gateway != nil` from the above if should be removed if we don't care about the state of the gateway.

spowelljr (changes requested):

pkg/drivers/kic/kic.go
deenadayalansrinivasan: yes, that was the original intention, however, added the condition separately to handle specific use cases that could arise based on the gateway. We could skip and just validate the staticIP check and use it.

PR #19552, opened 2024-09-02
Update linux command in pushing.md to use docker daemon inside minikube

amigueldiez:
## Description of change

This update corrects a command in the documentation related to using the Docker daemon with Minikube. The previous command was invalid, and it has been replaced with the correct one that properly connects to Minikube's Docker daemon.

## Details

By running the following command in the terminal:

```
$ minikube docker-env
```

The output is:

```
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.67.2:2376"
export DOCKER_CERT_PATH="/home/XXX/.minikube/certs"

```

The command shown at the end `(eval $(minikube -p minikube docker-env))` is the one that works and should be used, whereas the one previously found in the documentation is incorrect. This update replaces the invalid command with the correct one.

### DIFF
--- a/site/content/en/docs/handbook/pushing.md
+++ b/site/content/en/docs/handbook/pushing.md
@@ -57,7 +57,7 @@ To point your terminal to use the docker daemon inside minikube run this:
 {{% tabs %}}
 {{% linuxtab %}}
 ```shell
-eval $(minikube docker-env)
+eval $(minikube -p minikube docker-env)
 ```
 {{% /linuxtab %}}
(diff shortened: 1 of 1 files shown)


spowelljr:
> The previous command was invalid

`minikube docker-env` is a completely fine command, the command you replaced it with is just explicitly setting the profile flag, but if you don't supply the `-p` command it does `-p minikube` automatically (assuming you don't have a different profile set with `minikube profile <profile>`.

Issue #5950, opened 2019-11-19
kvm2: Cannot open boot2docker.iso: permission denied


lounicotra:
I have also tested this with a couple of previous releases, same results and also your latest beta. I own the .minikube dir under my home directory, so there should not be any permission issues downloading and placing files in the right dirs.

Seems that the boot2docker ISO never gets downloaded or placed in the correct place. Here's a listing of the dir structure after executing "minikube start":

lnicotra@tiger18 .minikube# du
16      ./certs
0       ./machines/minikube
0       ./machines
147212  ./cache/iso
341740  ./cache/images/k8s.gcr.io
500     ./cache/images/gcr.io/k8s-minikube
500     ./cache/images/gcr.io
342240  ./cache/images
489452  ./cache
0       ./config
0       ./addons
0       ./files
0       ./logs
14200   ./bin
4       ./profiles/minikube
4       ./profiles
503672  .

Deleting via "minikube delete" does not change the outcome...

Also, should you be providing a basic config.json file? That is also not being created.  Please provide a link to a downloadable one if necessary so the user can at least start minikube.

Thanks.

**The full output of the command that failed**:

lnicotra@tiger18 ~# minikube start
 minikube v1.5.2 on Centos 7.3.1611
 Automatically selected the 'kvm2' driver (alternates: [none])
 Downloading driver docker-machine-driver-kvm2:
    > docker-machine-driver-kvm2.sha256: 65 B / 65 B [-------] 100.00% ? p/s 0s
    > docker-machine-driver-kvm2: 13.87 MiB / 13.87 MiB [----] 100.00% ? p/s 0s
 Downloading VM boot image ...
    > minikube-v1.5.1.iso.sha256: 65 B / 65 B [--------------] 100.00% ? p/s 0s
    > minikube-v1.5.1.iso: 143.76 MiB / 143.76 MiB [-] 100.00% 42.13 MiB p/s 4s
 Creating kvm2 VM (CPUs=2, Memory=2000MB, Disk=20000MB) ...
[...]

PR #19557, opened 2024-09-02
libvirt-qemu error message should mention `$MINIKUBE_HOME`

peterbecich:
Error message was
"Ensure that the user listed in /etc/libvirt/qemu.conf has access to your home directory"

It is not always the home directory.

related issue: https://github.com/kubernetes/minikube/issues/5950

I have not translated this correctly.


peterbecich:
@medyagh please review

PR #19568, opened 2024-09-03
Addon headlamp: Update headlamp-k8s/headlamp image from v0.25.0 to v0.25.1

joaquimrocha:
Upgrade Headlamp version to 0.25.1
  cc: @illume


pkg/minikube/assets/addons.go
```diff
@@ -779,7 +779,7 @@ var Addons = map[string]*Addon{
 		MustBinAsset(addons.HeadlampAssets, "headlamp/headlamp-clusterrolebinding.yaml", vmpath.GuestAddonsDir, "headlamp-clusterrolebinding.yaml", "0640"),
 	}, false, "headlamp", "3rd party (kinvolk.io)", "yolossn", "https://minikube.sigs.k8s.io/docs/handbook/addons/headlamp/",
 		map[string]string{
-			"Headlamp": "headlamp-k8s/headlamp:v0.25.0@sha256:899d106eeb55b0afc4ee6e51c03bc4418de0bd0e79c39744d4d0d751aae6a971",
+			"Headlamp": "headlamp-k8s/headlamp:v0.25.1@",
```
illume, line 782: It seems that the hash is not here, because the image is not there yet... I will update this later.

PR #19580, opened 2024-09-08
Detect slow docker when deleting containers and profile

### LINKED ISSUE #19577: "warn if docker is slow" logic doesnt work
when trying to fix this issue https://github.com/kubernetes/minikube/pull/19576
that docker desktop Hangs on listing containers... I didnt get any warnninig from minikube that docker is slow and minikube hanged....

(only with a docker desktop Docker Desktop 4.34.0 with macos)
```
$ minikube delete --all --alsologtostderr
I0906 10:47:45.127199   28913 out.go:291] Setting OutFile to fd 1 ...
I0906 10:47:45.128263   28913 out.go:343] isatty.IsTerminal(1) = true
I0906 10:47:45.128272   28913 out.go:304] Setting ErrFile to fd 2...
I0906 10:47:45.128278   28913 out.go:343] isatty.IsTerminal(2) = true
I0906 10:47:45.128634   28913 root.go:338] Updating PATH: /Users/medya/.minikube/bin
I0906 10:47:45.130362   28913 out.go:298] Setting JSON to false
I0906 10:47:45.130901   28913 cli_runner.go:164] Run: docker ps -a --filter label=name.minikube.sigs.k8s.io --format {{.Names}}
```

I remembered that we had a logic that was supposed to Warn users but apparantly that logic doesnt work
here is the code that lists the containers in oci.go

```
func ListContainersByLabel(ctx context.Context, ociBin string, label string, warnSlow ...bool) ([]string, error) {
	rr, err := runCmd(exec.CommandContext(ctx, ociBin, "ps", "-a", "--filter", fmt.Sprintf("label=%s", label), "--format", "{{.Names}}"), warnSlow...)
	if err != nil {
		return nil, err
	}
	s := bufio.NewScanner(bytes.NewReader(rr.Stdout.Bytes()))
	var names []string
	for s.Scan() {
		n := strings.TrimSpace(s.Text())
		if n != "" {
			names = append(names, n)
		}
	}
	return names, s.Err()
```

and also here is the runCmd logic that was supposed to Warn users that the docker or podman is returning slow

the logic is in cli_runner.go

```
// runCmd runs a command exec.Command against docker daemon or podman
func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
	cmd = PrefixCmd(cmd)

	warn := false
	if len(warnSlow) > 0 {
		warn = warnSlow[0]
	}

	warnTime := 2 * time.Second

```

we need to fix this code so it can trigger the slowness

xcarolan:
Set warnSlow flag to try when deleting a profile and in Delete ContainersByLabel.

Fixes #19577

### DIFF
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -45,7 +45,8 @@ import (
 func DeleteContainersByLabel(ociBin string, label string) []error {
 	var deleteErrs []error
 	ctx := context.Background()
-	cs, err := ListContainersByLabel(ctx, ociBin, label)
+	//set warnSlow to true to give a warning for slow dockers
+	cs, err := ListContainersByLabel(ctx, ociBin, label, true)
 	if err != nil {
 		return []error{fmt.Errorf("listing containers by label %q", label)}
 	}
--- a/pkg/minikube/delete/delete.go
+++ b/pkg/minikube/delete/delete.go
@@ -47,7 +47,8 @@ func PossibleLeftOvers(ctx context.Context, cname string, driverName string) {
 
 	klog.Infof("deleting possible leftovers for %s (driver=%s) ...", cname, driverName)
 	delLabel := fmt.Sprintf("%s=%s", oci.ProfileLabelKey, cname)
-	cs, err := oci.ListContainersByLabel(ctx, bin, delLabel)
+	//set warnSlow to true to enable slow docker warning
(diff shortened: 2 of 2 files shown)


medyagh:
@xcarolan how is this PR going ? it is ready for being merged?

xcarolan:
> @xcarolan how is this PR going ? it is ready for being merged?

Not yet - i've been swamped at work.

PR #19677, opened 2024-09-20
Update "Using NVIDIA GPUs with minikube" docs

mattlgy:
Adding a note to delete existing minikube instance before starting one with the nvidia shim. This confused me for an good bit before I figured out why minikube couldn't see my GPUs.

Completely deleting the the instance and restarting from scratch seems to set everything up in a good state:
> 🌟  Enabled addons: nvidia-device-plugin, storage-provisioner, default-storageclass


medyagh (approved):
thank you @mattlgy for the contribution

PR #19729, opened 2024-09-29
Allow turning off the VBox accelerate3d flag

### LINKED ISSUE #19639: Can't start Minikube: VBOX_E_NOT_SUPPORTED
### What Happened?

If I try to start minikube with VirtualBox it won't start the VM with the error:
`😄  minikube v1.34.0 on Ubuntu 24.04
✨  Using the virtualbox driver based on existing profile
🤷  virtualbox "minikube" VM is missing, will recreate.
🔥  Creating virtualbox VM (CPUs=2, Memory=6000MB, Disk=20000MB) ...
🤦  StartHost failed, but will try again: recreate: creating host: create: creating: /usr/bin/VBoxManage modifyvm minikube --firmware bios --bioslogofadein off --bioslogofadeout off --bioslogodisplaytime 0 --biosbootmenu disabled --ostype Linux26_64 --cpus 2 --memory 6000 --acpi on --ioapic on --rtcuseutc on --natdnshostresolver1 on --natdnsproxy1 off --cpuhotplug off --pae on --hpet on --hwvirtex on --nestedpaging on --largepages on --vtxvpid on --accelerate3d off --boot1 dvd --natlocalhostreachable1 on failed:
VBoxManage: error: The graphics controller does not support the given feature
VBoxManage: error: Details: code VBOX_E_NOT_SUPPORTED (0x80bb0009), component GraphicsAdapterWrap, interface IGraphicsAdapter, callee nsISupports
VBoxManage: error: Context: "SetFeature(GraphicsFeature_Acceleration3D, ValueUnion.f)" at line 1108 of file VBoxManageModifyVM.cpp
`

The graphics driver is amdgpu and the CPU is a Ryzen 5 Pro on a ThinkPad L15 Gen 4 laptop

### Attach the log file

There are no generated logs as the VM can't start.

### Operating System

Ubuntu

### Driver

VirtualBox

afbjorklund:
Oracle VirtualBox version 7.1 seems to have some issues with
the `--accelerate3d` that was used by default with VirtualBox.

```
minikube start --driver=virtualbox --no-accelerate3d-off
```

Closes #19639

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -99,6 +99,7 @@ const (
 	vsockPorts              = "hyperkit-vsock-ports"
 	embedCerts              = "embed-certs"
 	noVTXCheck              = "no-vtx-check"
+	noAccelerate3DOff       = "no-accelerate3d-off"
 	downloadOnly            = "download-only"
 	dnsProxy                = "dns-proxy"
 	hostDNSResolver         = "host-dns-resolver"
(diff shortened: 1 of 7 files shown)


afbjorklund:
But minikube started just fine (on Linux) without this fix, as well.

When using 7.1.2 from https://www.virtualbox.org/wiki/Downloads

afbjorklund:
Added PR for the failing `functional_podman_ubuntu` test in CI:

* https://github.com/kubernetes/minikube/pull/19731

spowelljr (commented):
I completely agree with the idea of the PR, the flag being a double negative is a little confusing though, could we just rename it to `accelerate3d`?

ie. `--accelerate3d=true` vs `--no-accelerate3d-off=true`

afbjorklund:
> I completely agree with the idea of the PR, the flag being a double negative is a little confusing though, could we just rename it to `accelerate3d`?
> 
> ie. `--accelerate3d=true` vs `--no-accelerate3d-off=true`

Sure, but that is not what the flag does. It removes the parameter `--accelerate3d=off` from VBoxManage modifyvm.

Because in some builds, that flag is now throwing an error... But the 3D acceleration was never meant to be enabled.

spowelljr:
I see, if the `accelerate3d` flag is not specified what does it default to? If it defaults to true we could just remove it for everyone, but I'm assuming that's not the case.

medyagh:
@afbjorklund are you happy with the PR to be merged this way ? I think regardless of the naming of the flag it helps vbox users but I agree double negative flag seems kind of counter-intuitive and hard to read

Issue #19735, opened 2024-09-30
minikube checks for updates from the releases-v2.json file every time it starts.


nirajanpdl:
### What Happened?

In air gapped environment minikube is looking for   releases-v2.json file "https://storage.googleapis.com/minikube/releases-v2.json":  everry time we start minikube.

This issue is seen from minikube v1.33.0. how to fix it in air gapped environment ?

### Attach the log file

Running on machine: host
Binary: Built with gc go1.22.5 for linux/amd64
Log line format: [IWEF]mmdd hh:mm:ss.uuuuuu threadid file:line] msg
I0930 15:57:55.260968 1341420 out.go:345] Setting OutFile to fd 1 ...
I0930 15:57:55.261474 1341420 out.go:392] TERM=xterm,COLORTERM=, which probably does not support color
I0930 15:57:55.261478 1341420 out.go:358] Setting ErrFile to fd 2...
I0930 15:57:55.261483 1341420 out.go:392] TERM=xterm,COLORTERM=, which probably does not support color
I0930 15:57:55.261670 1341420 root.go:338] Updating PATH: /home/user/.minikube/bin
I0930 15:57:55.262075 1341420 out.go:352] Setting JSON to false
I0930 15:57:55.263823 1341420 start.go:129] hostinfo: {"hostname":"host-22-1-4-79","uptime":24780,"bootTime":1727667296,"procs":325,"os":"linux","platform":"ubuntu","platformFamily":"debian","platformVersion":"22.04","kernelVersion":"5.15.0-102-generic","kernelArch":"x86_64","virtualizationSystem":"kvm","virtualizationRole":"guest","hostId":"a63cc7ec-7f02-4eb4-a18a-4529f6bd13fa"}
I0930 15:57:55.263894 1341420 start.go:139] virtualization: kvm guest
I0930 15:57:55.265418 1341420 out.go:177] * minikube v1.34.0 on Ubuntu 22.04 (kvm/amd64)
I0930 15:57:55.266084 1341420 notify.go:220] Checking for updates...
I0930 15:57:55.266102 1341420 out.go:177]   - MINIKUBE_WANTREPORTERRORPROMPT=false
W0930 15:57:55.266079 1341420 preload.go:293] Failed to list preload files: open /home/user/.minikube/cache/preloaded-tarball: no such file or directory
I0930 15:57:55.266815 1341420 out.go:177]   - KUBECONFIG=/home/user/.kube/config
[...]

PR #19737, opened 2024-09-30
Check if should notify before making curl command

spowelljr:
Fixes https://github.com/kubernetes/minikube/issues/19735


medyagh:
closing since seems like this is the issue for kubenretes list version not related to the issue listed

PR #19748, opened 2024-10-03
fix --wait's failure to work on coredns pods

### LINKED ISSUE #19288: `--wait` flag sometimes misses pods
I was experimenting with the `--wait=all` command and found the that implementation grabs a list of the pods and then waits for them to be healthy, however, sometimes the list of pods is pulled before some of the pods have even been created, resulting in them not being in the waiting check. For example I had a run where the wait returned quickly, but after running `kubectl get pods -A` I could see that the coredns pod was just created and was not ready.

ComradeProgrammer:
FIX #19288 
Before: `minikube start --wait=all` may end when coredns is not ready
After: `minikube start --wait=all` will be able to wait until coredns is completly ready

The situation mentioned in #19288 was actually introduced by the HA-cluster PR.  By default coredns has a deployment consists 2 coredns pods. However in pkg/minikube/node/start.go:158 it was manyally scaled down to 1. This happens before we start to wait for those essential nodes(minikube waits for nodes at line 236).

When minikube waits for system pods, there were 2 checks  which will check the system pods's status:
- WaitExtra (pkg/minikube/bootstrapper/bsutil/kverify/pod_ready.go) will list all pods with the given labels, and check whether they are ready
- ExpectAppsRunning will list all the pods in kube-system namespace (pkg/minikube/bootstrapper/bsutil/kverify/system_pods.go:91 , in function ExpectAppsRunning), and check whether there are at least 1 running pod for some essential labels. **But the bug is that it only check the running state, and do not check the ready state**

After the HA-cluster PR was introduced, when minikube run WaitExtra funtion(the 1st check), one of the coredns pod's status can be Succeed. WaitExtra don't recognize this state and will print an error and break the checking loop. This logic is written at pkg/minikube/bootstrapper/bsutil/kverify/pod_ready.go line 99 and line 69.

The error I see is 
``` 
E1003 22:37:04.296076   17140 pod_ready.go:66] WaitExtra: waitPodCondition: pod "coredns-7db6d8ff4d-nljst" in "kube-system" namespace has status phase "Succeeded" (skipping!): {Phase:Succeeded Conditions:[{Type:PodReadyToStartContainers Status:False LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2024-10-03 22:37:04 +0200 CEST Reason: Message:} {Type:Initialized Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2024-10-03 22:36:53 +0200 CEST Reason:PodCompleted Message:} {Type:Ready Status:False LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2024-10-03 22:36:53 +0200 CEST Reason:PodCompleted Message:} {Type:ContainersReady Status:False LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2024-10-03 22:36:53 +0200 CEST Reason:PodCompleted Message:} {Type:PodScheduled Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2024-10-03 22:36:53 +0200 CEST Reason: Message:}] Message: Reason: NominatedNodeName: HostIP:192.168.49.2 HostIPs:[{IP:192.168.49.2}] PodIP: PodIPs:[] StartTime:2024-10-03 22:36:53 +0200 CEST InitContainerStatuses:[] ContainerStatuses:[{Name:coredns State:{Waiting:nil Running:nil Terminated:&ContainerStateTerminated{ExitCode:0,Signal:0,Reason:Completed,Message:,StartedAt:2024-10-03 22:36:54 +0200 CEST,FinishedAt:2024-10-03 22:37:04 +0200 CEST,ContainerID:docker://281a8c3106510cdc16a6d3f91ad2e9f5d7aa1609fac4f0e8f7494af67cb8b5d6,}} LastTerminationState:{Waiting:nil Running:nil Terminated:nil} Ready:false RestartCount:0 Image:registry.k8s.io/coredns/coredns:v1.11.1 ImageID:docker-pullable://registry.k8s.io/coredns/coredns@sha256:1eeb4c7316bacb1d4c8ead65571cd92dd21e27359f0d4917f1a5822a73b75db1 ContainerID:docker://281a8c3106510cdc16a6d3f91ad2e9f5d7aa1609fac4f0e8f7494af67cb8b5d6 Started:0x14001e58e00 AllocatedResources:map[] Resources:nil VolumeMounts:[]}] QOSClass:Burstable EphemeralContainerStatuses:[] Resize: ResourceClaimStatuses:[]}
```
The when minikube arrives at ExpectAppsRunning(the 2nd check), it doesn't check the ready state, so it believes that all pods are ok. This causes the #19288 

So the fix is to make ExpectAppsRunning(the 2nd check) check the ready state as well.

(The reason why I didn't make the 1st check function to recognized the Succeed state is that: if for some reason there is a job (e.g. init job for some containers) in kube-system namespace, and we change the WaitExtra's logic to reject Succeed state, there will be problems.)

### DIFF
--- a/pkg/minikube/bootstrapper/bsutil/kverify/system_pods.go
+++ b/pkg/minikube/bootstrapper/bsutil/kverify/system_pods.go
@@ -76,7 +76,8 @@ func WaitForSystemPods(r cruntime.Manager, bs bootstrapper.Bootstrapper, cfg con
 }
 
 // ExpectAppsRunning returns whether or not all expected k8s-apps are running. (without waiting for them)
-func ExpectAppsRunning(cs *kubernetes.Clientset, expected []string) error {
+func ExpectAppsRunning(cfg *config.ClusterConfig, cs *kubernetes.Clientset, expected []string) error {
+
 	found := map[string]bool{}
 
 	pods, err := cs.CoreV1().Pods("kube-system").List(context.Background(), meta.ListOptions{})
@@ -85,11 +86,30 @@ func ExpectAppsRunning(cs *kubernetes.Clientset, expected []string) error {
 	}
 	klog.Infof("%d kube-system pods found", len(pods.Items))
 
+	for !cfg.DisableOptimizations {
+		// when --disable-optimization is not specified
+		// core dns deployment has been scaled to 1 pods, wait until there is only 1 pod
+		corednsPods, err := cs.CoreV1().Pods("kube-system").List(context.Background(), meta.ListOptions{
+			LabelSelector: "k8s-app=kube-dns",
+		})
+		if err != nil {
+			return err
+		}
+		if len(corednsPods.Items) == 1 {
+			break
+		}
+	}
+
 	for _, pod := range pods.Items {
 		klog.Info(podStatusMsg(pod))
 
 		if pod.Status.Phase != core.PodRunning {
 			continue
+		}
+		for _, cs := range pod.Status.ContainerStatuses {
+			if !cs.Ready {
+				continue
+			}
 		}
 
 		for k, v := range pod.ObjectMeta.Labels {
@@ -112,12 +132,12 @@ func ExpectAppsRunning(cs *kubernetes.Clientset, expected []string) error {
 }
 
 // WaitForAppsRunning waits for expected Apps To be running
-func WaitForAppsRunning(cs *kubernetes.Clientset, expected []string, timeout time.Duration) error {
+func WaitForAppsRunning(cfg *config.ClusterConfig, cs *kubernetes.Clientset, expected []string, timeout time.Duration) error {
 	klog.Info("waiting for k8s-apps to be running ...")
 	start := time.Now()
 
 	checkRunning := func() error {
-		return ExpectAppsRunning(cs, expected)
+		return ExpectAppsRunning(cfg, cs, expected)
 	}
 
 	if err := retry.Local(checkRunning, timeout); err != nil {
--- a/pkg/minikube/bootstrapper/bsutil/kverify/system_pods.go
+++ b/pkg/minikube/bootstrapper/bsutil/kverify/system_pods.go
@@ -86,8 +86,9 @@ func ExpectAppsRunning(cfg *config.ClusterConfig, cs *kubernetes.Clientset, expe
 	}
 	klog.Infof("%d kube-system pods found", len(pods.Items))
 
-	for !cfg.DisableOptimizations {
+	for !config.IsHA(*cfg) && !cfg.DisableOptimizations {
 		// when --disable-optimization is not specified
+		// for non-HA cluster
 		// core dns deployment has been scaled to 1 pods, wait until there is only 1 pod
 		corednsPods, err := cs.CoreV1().Pods("kube-system").List(context.Background(), meta.ListOptions{
 			LabelSelector: "k8s-app=kube-dns",
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -567,7 +567,7 @@ func (k *Bootstrapper) WaitForNode(cfg config.ClusterConfig, n config.Node, time
 		}
 
 		if cfg.VerifyComponents[kverify.AppsRunningKey] {
-			if err := kverify.WaitForAppsRunning(client, kverify.AppsRunningList, timeout); err != nil {
+			if err := kverify.WaitForAppsRunning(&cfg, client, kverify.AppsRunningList, timeout); err != nil {
 				return errors.Wrap(err, "waiting for apps_running")
 			}
 		}


prezha:
hey @ComradeProgrammer thanks for looking into this

here's a bit more context that might help you:

wait for system-critical pods' _Ready_ condition implementation was intentionally made alongside the _Running_ checks, ie, the former would be called only if `wait` was explicitly requested and would add some startup delay, whereas the latter would almost always be called for quickest startup time

so, as the ExpectAppsRunning is called by `WaitForAppsRunning`, which, in turn, is called by `WaitForNode` that is always called (with the exception for the first ha control plane), by adding the _Ready_ check to the ExpectAppsRunning, we'd effectively always wait, which is not the intention

on the other hand, only if wait _is_ required, WaitExtra is called by `WaitForNode` (but before the `WaitForAppsRunning`) or by restartPrimaryControlPlane

also, for ha + wait, iirc the idea was that we'd be ok with waiting for at least one coredns pod to be ready, as the kube-dns service would take care of routing requests to the pod(s) that can process them

[example](https://github.com/kubernetes/minikube/issues/19288#issuecomment-2369106221) from #19288 shows only one coredns, so it's not a ha cluster, but makes a very good point:
>sometimes the list of pods is pulled before some of the pods have even been created, resulting in them not being in the waiting check

i think that the fix should be made in `WaitExtra`, and there we could eg, invert the logic so not to list all pods once and then loop through it waiting for each pod that's also on a system-critical list to become _Ready_ (as we do now), but instead to wait until all system-critical pods (which is a fixed list: kverify.CorePodsLabels) became _Ready_, re-fetching the pod's status as needed

as for `Succeed` status phase, that means that "All containers in the Pod have terminated in success, and will not be restarted", so it is handled - by skipping it (ie, it will never become _Ready_, so no point waiting for it)

spowelljr (changes requested):
I just tried this but `coredns` still wasn't ready.

```
$ minikube start --wait=all
😄  minikube v1.34.0 on Debian rodete (kvm/amd64)
🚜  Pulling base image v0.0.45-1727731891-master ...
🔥  Creating docker container (CPUs=2, Memory=26100MB) ...
🐳  Preparing Kubernetes v1.31.1 on Docker 27.3.1 ...
    ▪ Configuring RBAC rules ...

$ kubectl get pods -A
NAMESPACE     NAME                               READY   STATUS    RESTARTS   AGE
kube-system   coredns-7c65d6cfc9-qjrwh           0/1     Running   0          9s
kube-system   etcd-minikube                      1/1     Running   0          14s
kube-system   kube-apiserver-minikube            1/1     Running   0          14s
kube-system   kube-controller-manager-minikube   1/1     Running   0          15s
kube-system   kube-proxy-c8mdp                   1/1     Running   0          9s
kube-system   kube-scheduler-minikube            1/1     Running   0          16s
kube-system   storage-provisioner                1/1     Running   0          9s
```

pkg/minikube/bootstrapper/bsutil/kverify/system_pods.go
```diff
@@ -85,12 +86,30 @@ func ExpectAppsRunning(cs *kubernetes.Clientset, expected []string) error {
...
 
+	for {
+		// core dns deployment has been scaled to 1 pods, wait until there is only 1 pod
+		corednsPods, err := cs.CoreV1().Pods("kube-system").List(context.Background(), meta.ListOptions{
+			LabelSelector: "k8s-app=kube-dns",
+		})
+		if err != nil {
+			return err
+		}
+		if len(corednsPods.Items) == 1 {
```
medyagh, line 97: we have a flag that is called disable-optimizations ..or something like that, that would make the coredns 2 replicas
in that case it would be not working

medyagh (changes requested):

ComradeProgrammer:
> I just tried this but `coredns` still wasn't ready.
> 
> ```
> $ minikube start --wait=all
> 😄  minikube v1.34.0 on Debian rodete (kvm/amd64)
> ✨  Automatically selected the docker driver
> 📌  Using Docker driver with root privileges
> 👍  Starting "minikube" primary control-plane node in "minikube" cluster
> 🚜  Pulling base image v0.0.45-1727731891-master ...
> 🔥  Creating docker container (CPUs=2, Memory=26100MB) ...
> 🐳  Preparing Kubernetes v1.31.1 on Docker 27.3.1 ...
>     ▪ Generating certificates and keys ...
>     ▪ Booting up control plane ...
>     ▪ Configuring RBAC rules ...
> 🔗  Configuring bridge CNI (Container Networking Interface) ...
> 🔎  Verifying Kubernetes components...
>     ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
> 🌟  Enabled addons: storage-provisioner, default-storageclass
> 🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default
> 
> $ kubectl get pods -A
> NAMESPACE     NAME                               READY   STATUS    RESTARTS   AGE
> kube-system   coredns-7c65d6cfc9-qjrwh           0/1     Running   0          9s
> kube-system   etcd-minikube                      1/1     Running   0          14s
> kube-system   kube-apiserver-minikube            1/1     Running   0          14s
> kube-system   kube-controller-manager-minikube   1/1     Running   0          15s
> kube-system   kube-proxy-c8mdp                   1/1     Running   0          9s
> kube-system   kube-scheduler-minikube            1/1     Running   0          16s
> kube-system   storage-provisioner                1/1     Running   0          9s
> ```

could you plz have a try again to see if it works now? thx

medyagh:
@ComradeProgrammer Docker_Linux_containerd seems weird, there is no failure in gopogh but it exited as failure, could the None Zero code be related to waiting for DNS?   https://storage.googleapis.com/minikube-builds/logs/19748/37784/Docker_Linux_containerd.html

PR #19750, opened 2024-10-04
More robust MAC address matching

nirs:
On darwin bootp uses non-standard MAC address format[1]:
"8e:1:99:9c:54:b1" instead of "8e:01:99:9c:54:b1". We fixed this by
trimming leading "0" in the string before looking up the IP address.
    
There are several issues with the current code:
- Fragile, will break if bootp changes the format (unlikely)
- Fixing the wrong place, the drivers should not care about the MAC
  address format.
- The tests were confusing, showing that we can match standard MAC
  addresses while the actual code could match only non-standard bootp
  addresses.
- Logging wrong MAC address since we trimmed leading zeros before
  logging
    
This change replace trimming leading zeros with parsing MAC address
strings and comparing the bytes. The test includes now both standard and
non-standard MAC addresses.
    
[1] https://openradar.appspot.com/FB15382970

### DIFF
--- a/pkg/drivers/common.go
+++ b/pkg/drivers/common.go
@@ -18,11 +18,12 @@ package drivers
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"syscall"
 
@@ -39,8 +40,6 @@ import (
 
 // LeasesPath is the path to dhcpd leases
 const LeasesPath = "/var/db/dhcpd_leases"
-
-var leadingZeroRegexp = regexp.MustCompile(`0([A-Fa-f0-9](:|$))`)
 
 // This file is for common code shared among internal machine drivers
 // Code here should not be called from within minikube
@@ -190,7 +189,7 @@ func fixMachinePermissions(path string) error {
 type DHCPEntry struct {
 	Name      string
 	IPAddress string
-	HWAddress string
+	HWAddress net.HardwareAddr
 	ID        string
 	Lease     string
 }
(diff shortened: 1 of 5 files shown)


nirs:
I see that this is already fixed earlier:

```go
	// Need to strip 0's
	mac = pkgdrivers.TrimMacAddress(mac)
	if err := d.setupIP(mac); err != nil {
		return err
	}
```

Probably when this failed the reason was different (firewall or evil corp vpn software). When I came back to look at this, I found the trimmed addresses in the dhcp leases.

But the tests are confusing, not showing the actual entries and trimmed mac addresses, and the current code will break if bootp fixes this issue and start formatting mac addresses like other tools.

If we replace trimming with parsing, we don't need to care how bootp or other programs format hw_address.

pkg/drivers/common.go
```diff
@@ -264,6 +281,18 @@ func parseDHCPdLeasesFile(file io.Reader) ([]DHCPEntry, error) {
 	return dhcpEntries, scanner.Err()
 }
 
+func parseMACAddress(mac string) ([]byte, error) {
```
nirs, line 284: Found https://pkg.go.dev/net#ParseMAC now, I'll replace it if we continue with this change.

nirs:
@spowelljr can you review? you added this in 8c12c40853.

pkg/drivers/common.go
```diff
@@ -190,7 +189,7 @@ func fixMachinePermissions(path string) error {
 type DHCPEntry struct {
 	Name      string
 	IPAddress string
-	HWAddress string
+	HWAddress net.HardwareAddr
```
medyagh, line 192: thank you ! I like that

medyagh (approved):

PR #19758, opened 2024-10-07
fix timeout when stopping KVM machine with CRI-O container runtime

prezha:
we noticed in our integration tests that kvm machine with cri-o container runtime will sometime not gracefully shutdown waiting a given time of two minutes (eg: https://storage.googleapis.com/minikube-builds/logs/19736/36442/KVM_Linux_crio.html#fail_TestStartStop%2fgroup%2fembed-certs%2fserial%2fStop)

there is a known upstream [issue](https://github.com/cri-o/cri-o/issues/8609) that is tracking that

as we had similar issues in the past (and increased the timeout value several times), this pr aims to solve that problem by:
- waiting initial 90 seconds for the kvm machine to gracefully stop, and if that fails, instead of terminating with error:
- requesting a forceful shutdown (ie, "destroy" in libvirt) and waiting for another 30 seconds

in total, this should: self-resolve the issue in many cases, while reducing the total wait period

bonus:
- ensure any error from trying to close connection/domain is caught and logged but does not overwrite the underlying error
- more consistent error handling
- more consistent logging

### DIFF
--- a/pkg/drivers/kvm/kvm.go
+++ b/pkg/drivers/kvm/kvm.go
@@ -398,7 +398,7 @@ func (d *Driver) Create() error {
 		// Limiting the number of disks to 20 arbitrarily. If more disks are
 		// needed, the logical name generation has to changed to create them if
 		// the form hdaa, hdab, etc
-		return errors.New("creating more than 20 extra disks")
+		return errors.New("cannot create more than 20 extra disks")
 	}
 	for i := 0; i < d.ExtraDisks; i++ {
 		diskpath := pkgdrivers.ExtraDiskPath(d.BaseDriver, i)
--- a/pkg/drivers/kvm/kvm.go
+++ b/pkg/drivers/kvm/kvm.go
@@ -553,7 +553,7 @@ func (d *Driver) waitForStopState(maxsec int, method string) (state.State, error
 		time.Sleep(1 * time.Second)
 	}
 
-	return s, fmt.Errorf("timed out waitig for domain %s, current state is %q", method, s)
+	return s, fmt.Errorf("timed out waiting for domain %s, current state is %q", method, s)
 }
 
 // Remove a host
--- a/pkg/drivers/kvm/kvm.go
+++ b/pkg/drivers/kvm/kvm.go
@@ -299,7 +299,7 @@ func (d *Driver) Start() error {
 	log.Info("getting domain XML...")
 	dom, conn, err := d.getDomain()
 	if err != nil {
-		return errors.Wrap(err, "getting domain")
+		return errors.Wrap(err, "getting domain XML")
 	}
 	defer func() {
 		if err := closeDomain(dom, conn); err != nil {
--- a/pkg/drivers/kvm/kvm.go
+++ b/pkg/drivers/kvm/kvm.go
@@ -117,35 +117,19 @@ func NewDriver(hostName, storePath string) *Driver {
 	}
 }
 
-// PreCommandCheck checks the connection before issuing a command
-func (d *Driver) PreCommandCheck() error {
-	conn, err := getConnection(d.ConnectionURI)
-	if err != nil {
-		return errors.Wrap(err, "error connecting to libvirt socket. Have you added yourself to the libvirtd group?")
-	}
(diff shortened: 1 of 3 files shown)


spowelljr:
Hi @prezha, following up on this according to https://github.com/cri-o/cri-o/issues/8609#issuecomment-2408033958, this was fixed in [crun 1.17](https://github.com/containers/crun/releases/tag/1.17). We've updated to crun 1.17 in https://github.com/kubernetes/minikube/pull/19640, is this PR still needed?

prezha:
hey @spowelljr, i think we do as we've occasionally seen these timeouts even before (for reasons related to other/different issues), so it would also reduce the flakiness of our integration tests

pkg/drivers/kvm/kvm.go
```diff
@@ -236,57 +247,79 @@ func (d *Driver) DriverName() string {
...
+	if err := d.ensureNetwork(); err != nil {
 		return errors.Wrap(err, "ensuring active networks")
 	}
 
-	log.Info("Getting domain xml...")
+	log.Info("getting domain XML...")
 	dom, conn, err := d.getDomain()
 	if err != nil {
-		return errors.Wrap(err, "getting connection")
+		return errors.Wrap(err, "getting domain")
```
spowelljr, line 302:
```suggestion
		return errors.Wrap(err, "getting domain XML")
```

pkg/drivers/kvm/kvm.go
```diff
@@ -297,44 +330,48 @@ func (d *Driver) waitForStaticIP(conn *libvirt.Connect) error {
...
 		return errors.Wrap(err, "creating network")
 	}
+
 	if d.GPU {
-		log.Info("Creating devices...")
-		d.DevicesXML, err = getDevicesXML()
+		log.Info("creating devices...")
+		xml, err := getDevicesXML()
 		if err != nil {
 			return errors.Wrap(err, "creating devices")
```
spowelljr, line 372:
```suggestion
		log.Info("getting devices XML...")
		xml, err := getDevicesXML()
		if err != nil {
			return errors.Wrap(err, "getting devices XML")
```

pkg/drivers/kvm/kvm.go
```diff
@@ -346,22 +383,22 @@ func (d *Driver) Create() (err error) {
...
+	if err := pkgdrivers.MakeDiskImage(d.BaseDriver, d.Boot2DockerURL, d.DiskSize); err != nil {
+		return errors.Wrap(err, "creating disk")
 	}
 
 	if d.ExtraDisks > 20 {
 		// Limiting the number of disks to 20 arbitrarily. If more disks are
 		// needed, the logical name generation has to changed to create them if
 		// the form hdaa, hdab, etc
-		return errors.Wrap(err, "cannot create more than 20 extra disks")
+		return errors.New("creating more than 20 extra disks")
```
spowelljr, line 401: The before message is a better user facing error, `cannot create` vs `creating`, although we should have validation for this way before

spowelljr (changes requested):

pkg/drivers/kvm/kvm.go
prezha: agree, reworded it

pkg/drivers/kvm/kvm.go
```diff
@@ -427,100 +468,154 @@ func ensureDirPermissions(store string) error {
...
+		if s == state.Stopped {
+			return state.Stopped, nil
 		}
 
+		log.Infof("waiting for domain %s %d/%d", method, i, maxsec)
+		time.Sleep(1 * time.Second)
 	}
 
-	return fmt.Errorf("unable to stop vm, current state %q", s.String())
+	return s, fmt.Errorf("timed out waitig for domain %s, current state is %q", method, s)
```
spowelljr, line 556:
```suggestion
	return s, fmt.Errorf("timed out waiting for domain %s, current state is %q", method, s)
```

spowelljr (approved):
Found a typo, but the tests look good!

medyagh:
thank you @prezha for improving the KVM, its nice to see less failed tests

PR #19771, opened 2024-10-08
fixed accidental folder deletion bug when using `minikube delete --purge`

### LINKED ISSUE #19769: 1.34 - "minikube delete --purge" deletes entire MINIKUBE_HOME directory if env var does not end with "/.minikube"
### What Happened?

For legacy reasons, my environment had `MINIKUBE_HOME` explicitly specified. Prior to 1.34.0, if the `MINIKUBE_HOME` path did not end with `/.minikube`, all commands would functionally append `/.minikube` to the path before execution. If no `.minikube` directory existed in the directory specified by `MINIKUBE_HOME`, the command `minikube delete --purge` would have no effect.

Now, as of 1.34.0, if `MINIKUBE_HOME` is defined and does not end with `/.minikube`, the command `minikube delete --purge` deletes the full directory specified by `MINIKUBE_HOME` if no `.minikube` directory exists within the directory. This can be **extremely** destructive if `MINIKUBE_HOME` is specified to a location such as the user's home folder.

I verified that rolling back to 1.33.1 prevents the problem, and I reproduced the issue on Windows 10 and 11. I have not had a chance to test it on a Linux distribution.

[Configuration documentation](https://minikube.sigs.k8s.io/docs/handbook/config/#exclusive-environment-tunings) for `MINIKUBE_HOME` at the time of writing:
> **MINIKUBE_HOME** - (string) sets the path for the .minikube directory that minikube uses for state/configuration. If you specify it to /path/to/somewhere and somewhere is not equal to .minikube, the final MINIKUBE_HOME will be /path/to/somewhere/.minikube. Defaults to ~/.minikube if unspecified. *Please note: this is used only by minikube and does not affect anything related to Kubernetes tools such as kubectl.*

[Documentation for delete](https://minikube.sigs.k8s.io/docs/commands/delete/#options) `--purge` flag at the time of writing:
> Set this flag to delete the '.minikube' folder from your user directory.

To reproduce:
1) Install minikube 1.34.0.
2) Open up a new terminal.
    - Occurs in PowerShell and GitBash
    - Elevated permissions are not necessary, so a non-admin terminal will suffice
3) Set the `MINIKUBE_HOME` environment variable to a location to which the logged-in user has access.
4) Run `minikube delete --purge`. Note: You do not need to start minikube at all for this issue to occur.

The log files below were generated with `MINIKUBE_HOME` set to `%HOME%\Desktop\test-minikube`.

### Attach the log file

[1.34.0-with-.minikube-folder.txt](https://github.com/user-attachments/files/17284552/1.34.0-with-.minikube-folder.txt)
[1.34.0-without-.minikube-folder.txt](https://github.com/user-attachments/files/17284554/1.34.0-without-.minikube-folder.txt)

### Operating System

Windows

### Driver

N/A

johnmatthiggins:
fixes #19769

### DIFF
--- a/pkg/minikube/localpath/localpath.go
+++ b/pkg/minikube/localpath/localpath.go
@@ -47,12 +47,7 @@ func MiniPath() string {
 	if filepath.Base(minikubeHomeEnv) == ".minikube" {
 		return minikubeHomeEnv
 	}
-
-	legacyMinikubeHome := filepath.Join(minikubeHomeEnv, ".minikube")
-	if _, err := os.Stat(legacyMinikubeHome); !os.IsNotExist(err) {
-		return legacyMinikubeHome
-	}
-	return filepath.Clean(minikubeHomeEnv)
+	return filepath.Join(minikubeHomeEnv, ".minikube")
 }
 
 // MakeMiniPath is a utility to calculate a relative path to our directory.
--- a/pkg/minikube/localpath/localpath_test.go
+++ b/pkg/minikube/localpath/localpath_test.go
@@ -66,7 +66,7 @@ func TestMiniPath(t *testing.T) {
 		env, expectedPath string
 	}{
 		{"/tmp/.minikube", "/tmp/.minikube"},
-		{"/tmp", "/tmp"},
+		{"/tmp", "/tmp/.minikube"},
 		{"", filepath.Join(homedir.HomeDir(), ".minikube")},
 	}
 	for _, tc := range testCases {


pkg/minikube/localpath/localpath.go
```diff
@@ -47,12 +47,7 @@ func MiniPath() string {
 	if filepath.Base(minikubeHomeEnv) == ".minikube" {
 		return minikubeHomeEnv
 	}
-
```
medyagh, line 50: I think we kind of did a mistake in the previous release not to think about this, the rightthing to do is, 
in the delete --purge code. we need to Check What folders we are deleting...and if there is any folder other than minikube sub folders we should Warn the user if they wanna proceed.

medyagh (changes requested):

pkg/minikube/localpath/localpath.go
johnmatthiggins: I noticed you left a comment on issue #19769 that describes an alternative solution:
>  alternatvely we should enforce users to set their minikube home folder to something that is called .minikube, so in the purge we only delete .minikube folder

I [replied](https://github.com/kubernetes/minikube/issues/19769#issuecomment-2401111572) to the comment detailing why I think this alternative is the best approach and that adding a prompt while keeping the current MiniPath() behavior will actually be not aid in the overall goal of getting users to use a directory named ".minikube" as the config directory.

I'm a little confused about what solution to pursue of the two options presented in the comment. If you want a yes/no prompt that verifies whether the user wants to delete a directory when it's not named ".minikube" then I can rewrite the code. Just let me know, because I'm a little unclear on what direction I should go in.

johnmatthiggins:
Closing this because it has been replaced by PR #20045

PR #19802, opened 2024-10-14
testing: Remove requesting full CPU for nginx test pod

spowelljr:
Looking at Docker Linux container arm64 tests the Volcano addon test is consistently failing, looking at the logs I see the following:
```
Warning  FailedScheduling  2m59s  volcano  0/1 nodes are unavailable: 1 Insufficient cpu.
```

Looking into the pod, it's a nginx pod that's used in the Volcano addon test that requests a full CPU. All we do is check if the pod comes up before deleting it, so removing this high request.

### DIFF
--- a/test/integration/testdata/vcjob.yaml
+++ b/test/integration/testdata/vcjob.yaml
@@ -40,9 +40,4 @@ spec:
               - 10m
               image: nginx:latest
               name: nginx
-              resources:
-                requests:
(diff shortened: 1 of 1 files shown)


medyagh (approved):

medyagh:
intersting I wonder if any other addon "Requests" a full CPU

PR #19861, opened 2024-10-24
CI: Update GOPATH when building

spowelljr:


spowelljr:
Included in https://github.com/kubernetes/minikube/pull/19862

PR #19925, opened 2024-11-07
site: add step manage-docker-as-a-non-root-user in the requirements of docker driver

### LINKED ISSUE #19742: Installing step should have docker user and group creation step as well
### What Happened?

I was installing minikube and found that running minikube installation needs docker as a user and in group as well.

### Attach the log file

no such file for logs 

### Operating System

Ubuntu

### Driver

Docker

LuBingtan:
Close #19742


medyagh:
thank you @LuBingtan thats a good addition !

PR #20129, opened 2024-12-17
[20124] - Feature Request: Provide an option for passing a config file for addon configure command. 

omleteer:
Minor UI Change (Before and After described in the ticket) but tl;dr:

## BEFORE

Currently when configuring addons with the command:

```
minikube addons configure <addon>
```

eg:

```
minikube addons configure registry-creds
```

The user is prompted to input the values one by one, eg:

```
Do you want to enable AWS Elastic Container Registry? [y/n]:
```

and so on.    This makes it hard to store/load values in a scripted environemnt.   Some times it easier to provide a config file containing so that the values can be read from this.  

## AFTER

Add a "-c" local flag to the addon configure command, eg:

```
minikube addons registry-creds -c ~/.minikube/configs/registry-creds-config.json
minikube addons ingress -c ~/.minikube/configs/ingres-config.json
```

etc

### DIFF
--- a/cmd/minikube/cmd/config/configure.go
+++ b/cmd/minikube/cmd/config/configure.go
@@ -17,6 +17,7 @@ limitations under the License.
 package config
 
 import (
+	"log"
 	"net"
 	"os"
 	"regexp"
@@ -38,6 +39,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/sysinit"
 )
 
+var AddonConfigFile = ""
 var posResponses = []string{"yes", "y"}
 var negResponses = []string{"no", "n"}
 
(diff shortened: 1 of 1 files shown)


test/integration/testdata/addons_testconfig.json
```diff
@@ -0,0 +1,30 @@
+	{
```
medyagh, line 1: lets have top field called "addons:"

test/integration/configure_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	"k8s.io/minikube/pkg/kapi"
+	"k8s.io/minikube/pkg/minikube/constants"
+)
+
+// TestConfigure tests addons configure command to ensure that configurations are loaded/set correctly
+func TestAddonsConfigure(t *testing.T) {
```
medyagh, line 37: I would prefer have this in same test as other addons same cluster, 
so we pass the addon config file to the same cluster that enabled bunch of addons without config file,

medyagh (changes requested):

test/integration/configure_test.go
omleteer: +1. Ive have moved the test back to addons_test.

test/integration/testdata/addons_testconfig.json
omleteer: +1 Done

PR #20168, opened 2024-12-27
Add FAQ entry about `too many open files` errors when using a container-based driver

EmmanuelKasper:
This is based on https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files and was also mentioned in https://00formicapunk00.wordpress.com/2024/12/10/too-many-open-files-in-minikube-pod/

The problem will pop up on HA setups using a container driver.  I had this with Docker, but probably Podman will hit the same limit, as this is a Linux kernel limit independant of the container runtime.

### DIFF
--- a/site/content/en/docs/faq/_index.md
+++ b/site/content/en/docs/faq/_index.md
@@ -203,3 +203,34 @@ sudo mkdir -p "$CNI_PLUGIN_INSTALL_DIR"
 sudo tar -xf "$CNI_PLUGIN_TAR" -C "$CNI_PLUGIN_INSTALL_DIR"
 rm "$CNI_PLUGIN_TAR"
 ```
+
+## How to increase open files limit for minikube
+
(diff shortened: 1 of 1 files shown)


site/content/en/docs/handbook/troubleshooting.md
```diff
@@ -78,6 +78,37 @@ To view more detailed information about a pod, use:
 kubectl describe pod <name> -n <namespace>
 ```
 
+## Pods in error when using a container-based driver (docker, podman)
```
medyagh, line 81: this title could be more specific, and maybe a better place to put it would be FAQ ?
like "how to increase open file limits in minikube"
https://minikube.sigs.k8s.io/docs/faq/

medyagh (changes requested):

site/content/en/docs/handbook/troubleshooting.md
EmmanuelKasper: Agree that the title is too generic: there are plenty of reasons a Pod could err, we need to specify which one is addressed.
Also FAQ seem to have similar topics too. 
I will update the PR.

EmmanuelKasper: Marking this as resolved as last iteration should have addressed the concerns,

PR #20219, opened 2025-01-08
Site: fix slack link on minikube website

ComradeProgrammer:
fix: fix slack link on minikube website


medyagh (approved):

medyagh (changes requested):
we should still keep the Slack Link that Directly takes them to our Room, but also add a New link that says you need to First get invite here.

medyagh (approved):

medyagh (approved):

Issue #11068, opened 2021-04-11
Works only when I specify --base-image


aidagha:
When I start minikube without `--base-image=kicbase/stable:v0.0.20` I get this error:

```
minikube start --driver=docker
😄  minikube v1.19.0 on Darwin 10.15.7
✨  Using the docker driver based on user configuration
👍  Starting control plane node minikube in cluster minikube
🚜  Pulling base image ...
💾  Downloading Kubernetes v1.20.2 preload ...
    > gcr.io/k8s-minikube/kicbase...: 357.67 MiB / 357.67 MiB  100.00% 396.74 K
    > preloaded-images-k8s-v10-v1...: 491.71 MiB / 491.71 MiB  100.00% 411.47 K
    > index.docker.io/kicbase/sta...: 357.67 MiB / 357.67 MiB  100.00% 417.22 K
E0411 17:20:53.079237   47564 cache.go:180] Error downloading kic artifacts:  failed to download kic base image or any fallback image
🔥  Creating docker container (CPUs=2, Memory=1987MB) ...
🤦  StartHost failed, but will try again: creating host: create: creating: setting up container node: preparing volume for minikube container: docker run --rm --name minikube-preload-sidecar --label created_by.minikube.sigs.k8s.io=true --label name.minikube.sigs.k8s.io=minikube --entrypoint /usr/bin/test -v minikube:/var gcr.io/k8s-minikube/kicbase:v0.0.20@sha256:0250dab3644403384bd54f566921c6b57138eecffbb861f9392feef9b2ec44f6 -d /var/lib: exit status 125
stdout:

stderr:
Unable to find image 'gcr.io/k8s-minikube/kicbase:v0.0.20@sha256:0250dab3644403384bd54f566921c6b57138eecffbb861f9392feef9b2ec44f6' locally
docker: Error response from daemon: Get https://gcr.io/v2/: Service Unavailable.
See 'docker run --help'.

🤷  docker "minikube" container is missing, will recreate.
🔥  Creating docker container (CPUs=2, Memory=1987MB) ...
[...]

PR #20226, opened 2025-01-09
Fallback to trying kic image without SHA when using image registry

ghosind:
Like #11068, minikube with `--image-repository` option may fail to download kic image because it only tries to download the image with SHA. This patch will add the kic image without SHA to the fallback images list to avoid its failure.

### DIFF
--- a/pkg/minikube/node/cache.go
+++ b/pkg/minikube/node/cache.go
@@ -122,11 +122,19 @@ func beginDownloadKicBaseImage(g *errgroup.Group, cc *config.ClusterConfig, down
 	register.Reg.SetStep(register.PullingBaseImage)
 	out.Step(style.Pulling, "Pulling base image {{.kicVersion}} ...", out.V{"kicVersion": kic.Version})
 	g.Go(func() error {
+		images := make([]string, 0, 2+len(kic.FallbackImages))
 		baseImg := cc.KicBaseImage
+		baseFallbackImg := ""
 		if baseImg == kic.BaseImage && len(cc.KubernetesConfig.ImageRepository) != 0 {
-			baseImg = updateKicImageRepo(baseImg, cc.KubernetesConfig.ImageRepository)
+			baseImg, baseFallbackImg = updateKicImageRepo(baseImg, cc.KubernetesConfig.ImageRepository)
 			cc.KicBaseImage = baseImg
 		}
+		images = append(images, baseImg)
+		if baseFallbackImg != "" && baseFallbackImg != baseImg {
+			images = append(images, baseFallbackImg)
+		}
+		images = append(images, kic.FallbackImages...)
+
 		var finalImg string
 		// If we end up using a fallback image, notify the user
 		defer func() {
@@ -139,7 +147,7 @@ func beginDownloadKicBaseImage(g *errgroup.Group, cc *config.ClusterConfig, down
 		}()
 		// first we try to download the kicbase image (and fall back images) from docker registry
 		var err error
-		for _, img := range append([]string{baseImg}, kic.FallbackImages...) {
+		for _, img := range images {
 
 			if driver.IsDocker(cc.Driver) && download.ImageExistsInDaemon(img) && !downloadOnly {
 				klog.Infof("%s exists in daemon, skipping load", img)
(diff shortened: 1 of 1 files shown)


medyagh (changes requested):
hi @ghosind do you mind explaining the logic, how this doing it, some of the code is not clear for me.
and also I think if we have to do this, we might wanna output a warnning to user that we chose the image without SHA, since this could be a security issue

ghosind:
> hi @ghosind do you mind explaining the logic, how this doing it, some of the code is not clear for me. and also I think if we have to do this, we might wanna output a warnning to user that we chose the image without SHA, since this could be a security issue

Hi @medyagh,

minikube will try to download kic image with sha, and try the image without sha as fallback:

  https://github.com/kubernetes/minikube/blob/210b148df93a80eb872ecbeb7e35281b3c582c61/pkg/drivers/kic/types.go#L42

  https://github.com/kubernetes/minikube/blob/7a8a1c93ea94f8b5d128c5675ee870cce62d900d/pkg/minikube/node/cache.go#L142

However, it with `--image-registry` option only updates the base image's repo, but not the fallback image. So, if the image is unavailable, it never tries the repo updated image without sha.

  https://github.com/kubernetes/minikube/blob/7a8a1c93ea94f8b5d128c5675ee870cce62d900d/pkg/minikube/node/cache.go#L127

For example, start minikube v1.34.0 with `--image-registry=registry.cn-hangzhou.aliyuncs.com/google_containers` option, it will try to download the kic base image by the following sequence:

```sh
# update from gcr.io/k8s-minikube/kicbase:v0.0.45@sha256:81df288595202a317b1a4dc2506ca2e4ed5f22373c19a441b88cfbf4b9867c85
registry.cn-hangzhou.aliyuncs.com/google_containers/kicbase:v0.0.45@sha256:81df288595202a317b1a4dc2506ca2e4ed5f22373c19a441b88cfbf4b9867c85
docker.io/kicbase/stable:v0.0.45@sha256:81df288595202a317b1a4dc2506ca2e4ed5f22373c19a441b88cfbf4b9867c85
gcr.io/k8s-minikube/kicbase:v0.0.45
docker.io/kicbase/stable:v0.0.45
```

The image `registry.cn-.../kicbase:v0.0.45@sha256:81df...` is unavailable, and it will never try to download `registry.cn-.../kicbase:v0.0.45` like `gcr.io` one does. So, this patch will add `registry.cn-.../kicbase:v0.0.45` to the fallback list.

pkg/minikube/node/cache.go
```diff
@@ -122,11 +122,19 @@ func beginDownloadKicBaseImage(g *errgroup.Group, cc *config.ClusterConfig, down
...
 	g.Go(func() error {
+		images := make([]string, 0, 2+len(kic.FallbackImages))
 		baseImg := cc.KicBaseImage
+		baseFallbackImg := ""
 		if baseImg == kic.BaseImage && len(cc.KubernetesConfig.ImageRepository) != 0 {
-			baseImg = updateKicImageRepo(baseImg, cc.KubernetesConfig.ImageRepository)
+			baseImg, baseFallbackImg = updateKicImageRepo(baseImg, cc.KubernetesConfig.ImageRepository)
 			cc.KicBaseImage = baseImg
 		}
+		images = append(images, baseImg)
```
ghosind, line 132: if minikube does not start with `--image-registry` option, or specify a different kic image that without sha (with `--base-image` option), it will not add the image without sha into the fallback list

pkg/minikube/node/cache.go
```diff
@@ -278,13 +286,16 @@ func imagesInConfigFile() ([]string, error) {
...
 	if repo == constants.AliyunMirror {
 		// for aliyun registry must strip namespace from image name, e.g.
 		//   registry.cn-hangzhou.aliyuncs.com/google_containers/k8s-minikube/kicbase:v0.0.25 will not work
 		//   registry.cn-hangzhou.aliyuncs.com/google_containers/kicbase:v0.0.25 does work
 		image = strings.TrimPrefix(image, "k8s-minikube/")
 	}
-	return path.Join(repo, image)
+	baseImg := path.Join(repo, image)
+	// try a fallback image without sha, because #11068
+	fallbackImg := strings.Split(image, "@sha256:")[0]
```
ghosind, line 299: `pkg/driver/kic/types.go` did not export the image's sha, so I removed the sha part from the repo updated image.

ghosind:
As my last reply said, minikube is already trying to download no sha image as the fallback, is it a better idea to make another patch to add warning output? I think it is not for this patch only.

PR #20255, opened 2025-01-13
New flag to allow option for passing a config file for addon configure command.

### LINKED ISSUE #20124: Feature Request: Provide an option for passing a config file for addon configure command.
### What Happened?

Currently when configuring addons with the command:

```
minikube addons configure <addon>
```

eg:

```
minikube addons configure registry-creds
```

The user is prompted to input the values one by one, eg:

```
Do you want to enable AWS Elastic Container Registry? [y/n]:
```

and so on.

Some times it easier to provide a config file containing so that the values can be read from this.  

The request is for a "-c" local flag to the addon configure command, eg:

```
minikube addons configure registry-creds -c ~/.minikube/configs/registry-creds-config.json
```

An example of this file is:

```
{
  "enableAWSEcr": "enable",
  "awsEcrConfigs": {
     "awsAccessID": "changeme",
     "awsAccessKey": "changeme",
     "awsSessionToken": "changeme",
     "awsRegion": "changeme",
     "awsAccount": "changeme",
     "awsRole": "changeme"
  },
  "enableGCR": "disable",
  "gcrConfigs": {
     "gcrPath": "changeme",
     "gcrURL": "https://gcr.io"
  },
  "enableDockerRegistry": "disable",
  "dockerConfigs": {
     "dockerServer": "changeme",
     "dockerUser": "changeme",
     "dockerPass": "changeme"
  },
  "enableACR": "disable",
  "acrConfigs": {
     "acrURL": "changeme",
     "acrClientID": "changeme",
     "acrPassword": "changeme"
  }
}
```

The parsing of this config will be specific to the kind of addon.  (Here we have chosen a naming style consistent with the style within the code itself, eg`cmd/minikube/cmd/config/configure.go`, but we can adopt other stratgies here).

Also to consider:

* Whether to have a single options file for *all* addons
* [Preferred] addon specific options file - that MUST be passed to override default behavior always prompting.

### Attach the log file

This is not an issue but a feature request.

### Operating System

macOS (Default)

### Driver

Docker

panyam:
Minor UI Change (Before and After described in the ticket) but tl;dr:

## BEFORE

Currently when configuring addons with the command:

```
minikube addons configure <addon>
```

eg:

```
minikube addons configure registry-creds
```

The user is prompted to input the values one by one, eg:

```
Do you want to enable AWS Elastic Container Registry? [y/n]:
```

and so on.    This makes it hard to store/load values in a scripted environemnt.   Some times it easier to provide a config file containing so that the values can be read from this.  

## AFTER

Add a "-f" local flag to the addon configure command, eg:

```
minikube addons configure registry-creds -f ~/.minikube/configs/registry-creds-config.json
minikube addons configure ingress -f ~/.minikube/configs/ingres-config.json
```

etc

Fixes #20124


cmd/minikube/cmd/config/configure.go
```diff
@@ -339,5 +123,375 @@ func unpauseWholeCluster(co mustload.ClusterController) {
...
+		if err := addons.EnableOrDisableAddon(cfg, "registry-aliases", "true"); err != nil {
+			out.ErrT(style.Fatal, "Failed to configure registry-aliases {{.profile}}", out.V{"profile": profile})
+		}
+	}
+}
+
+// Processes registry-creds addon config from configFile if it exists otherwise resorts to default behavior
+func processRegistryCredsConfig(profile string, configFileData map[string]any) {
+	// Default values
+	awsAccessID := "changeme"
```
medyagh, line 303: is there reason for "changeme" default value? seems unconventional, I would prefer either "" or something very explicit that has no chance of clodiing with users real values, like "_MINIKUBE_DEFAULT_VALUE_"

medyagh (changes requested):

cmd/minikube/cmd/config/configure.go
panyam: Ah I like the MINIKUBE_DEFAULT_VALUE as a default.  Updated it.  PTAL

cmd/minikube/cmd/config/configure.go
```diff
@@ -38,6 +42,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/sysinit"
 )
 
+var AddonConfigFile = ""
```
medyagh: lets make this private

cmd/minikube/cmd/config/configure.go
```diff
@@ -339,5 +123,375 @@ func unpauseWholeCluster(co mustload.ClusterController) {
...
 }
+
+// Helper method to load a config file for addons
+func loadAddonConfigFile(addon, configFilePath string) (addonConfig map[string]any, err error) {
+	if configFilePath != "" {
+		configFileData := make(map[string]any)
+		out.Ln("Reading %s configs from %s", addon, configFilePath)
+		if confData, err := os.ReadFile(configFilePath); err != nil && errors.Is(err, os.ErrNotExist) {
+			exit.Message(reason.Usage, "config file does not exist")
+			return nil, err
```
medyagh, line 137: klog.warnf err before exis

cmd/minikube/cmd/config/configure.go
```diff
@@ -339,5 +123,375 @@ func unpauseWholeCluster(co mustload.ClusterController) {
...
+		} else if err != nil {
+			exit.Message(reason.Kind{ExitCode: reason.ExProgramConfig, Advice: "provide a valid config file"},
```
medyagh, line 139: klog.Errof(...)

cmd/minikube/cmd/config/configure.go
```diff
@@ -339,5 +123,375 @@ func unpauseWholeCluster(co mustload.ClusterController) {
...
+				fmt.Sprintf("error opening config file: %v", err))
+			return nil, err
```
medyagh, line 141: "wrap" the error

cmd/minikube/cmd/config/configure.go
```diff
@@ -339,5 +123,375 @@ func unpauseWholeCluster(co mustload.ClusterController) {
...
+		} else if err = json.Unmarshal(confData, &configFileData); err != nil {
+			exit.Message(reason.Kind{ExitCode: reason.ExProgramConfig, Advice: "provide a valid config file"},
```
medyagh, line 143: "ensure the json format of the config file is valid"

cmd/minikube/cmd/config/configure.go
```diff
@@ -339,5 +123,375 @@ func unpauseWholeCluster(co mustload.ClusterController) {
...
+				fmt.Sprintf("error opening config file: %v", err))
```
medyagh, line 144: error wrap

errors.Wrapf(err, "addon config %s" ... )

cmd/minikube/cmd/config/configure.go
```diff
@@ -339,5 +123,375 @@ func unpauseWholeCluster(co mustload.ClusterController) {
...
+			return nil, err
```
medyagh, line 145: if exit no need to return

cmd/minikube/cmd/config/configure.go
```diff
@@ -339,5 +123,375 @@ func unpauseWholeCluster(co mustload.ClusterController) {
...
+		}
+
+		// Make sure the addon specific config exists and it is a map
```
medyagh, line 148: make a helper function and put the nested ifs in there

test/integration/addons_test.go
```diff
@@ -304,6 +305,64 @@ func validateIngressAddon(ctx context.Context, t *testing.T, profile string) {
 	}
 }
 
+// validateRegistryCredsAddon tests the registry-creds addon by trying to load its configs
+func validateRegistryCredsAddon(ctx context.Context, t *testing.T, profile string) {
```
medyagh, line 309: lets use a struct for the regsitry cred addons so it can be typed checked

medyagh (changes requested):
this PR is exciting and its getting closer ! just a few more changes

cmd/minikube/cmd/config/configure.go
panyam: @medyagh - QQ - why do we need a klog AND exit.Message?   Also if we are exiting - there is no need to wrap the error right (since you called out earlier that we dont even have to return a nil)

panyam:
@medyagh Do you mind taking a look at the latest version?   I have a couple of cleanups to do but wanted to get your thoughts on the current "Typed configs" implementation to see if it looks ok.

cmd/minikube/cmd/config/configure.go
```diff
@@ -300,20 +300,20 @@ func processRegistryAliasesConfig(profile string, _ map[string]any) {
 // Processes registry-creds addon config from configFile if it exists otherwise resorts to default behavior
 func processRegistryCredsConfig(profile string, configFileData map[string]any) {
 	// Default values
-	awsAccessID := "changeme"
-	awsAccessKey := "changeme"
+	awsAccessID := "MINIKUBE_DEFAULT_VALUE"
```
nirs: Both "changeme" and "MINIKUBE_DEFAULT_VALUE" are not really defaults but a sentinel marking unset value. Can empty string be used instead or we these values are required?

nirs (commented):
Nice! To make perfect support both json and yaml configuration like k8s.

cmd/minikube/cmd/config/configure.go
```diff
@@ -339,5 +124,146 @@ func unpauseWholeCluster(co mustload.ClusterController) {
...
+			klog.Errorf("error opening config file (%s): %v", configFilePath, err)
+			// err = errors2.Wrapf(err, "config file (%s) does not exist", configFilePath)
+			exit.Message(reason.Kind{ExitCode: reason.ExProgramConfig, Advice: "provide a valid config file"},
+				fmt.Sprintf("error opening config file: %s", configFilePath))
+		} else if err = json.Unmarshal(confData, &config); err != nil {
+			// err = errors2.Wrapf(err, "error reading config file (%s)", configFilePath)
+			klog.Errorf("error reading config file (%s): %v", configFilePath, err)
+			exit.Message(reason.Kind{ExitCode: reason.ExProgramConfig, Advice: "provide a valid config file"},
+				fmt.Sprintf("error reading config file: %v", err))
+		}
```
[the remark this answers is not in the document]
panyam: Great idea.   I started with a simple option and planning to make this on a follow PR after discussion with @medyagh

cmd/minikube/cmd/config/configure_registry_creds.go
```diff
@@ -0,0 +1,251 @@
...
+import (
+	"os"
+
+	"k8s.io/minikube/pkg/minikube/exit"
+	"k8s.io/minikube/pkg/minikube/out"
+	"k8s.io/minikube/pkg/minikube/reason"
+	"k8s.io/minikube/pkg/minikube/service"
+)
+
+// Top level configs for RegistryCreds addons
```
panyam, line 12: Make these private.

cmd/minikube/cmd/config/configure.go
```diff
@@ -33,14 +37,19 @@ import (
...
 	"k8s.io/minikube/pkg/minikube/style"
 	"k8s.io/minikube/pkg/minikube/sysinit"
 )
 
+var addonConfigFile = ""
 var posResponses = []string{"yes", "y"}
 var negResponses = []string{"no", "n"}
 
+// Typed addon configs
+type AddonConfig struct {
```
panyam, line 49: This can also be private as not used outside this package.

cmd/minikube/cmd/config/configure.go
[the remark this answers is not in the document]
nirs: Sure, this great improvement as is and getting it the 1.36 will be nice.

cmd/minikube/cmd/config/configure.go
panyam: Done.

cmd/minikube/cmd/config/configure.go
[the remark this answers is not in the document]
panyam: Thanks heaps @nirs.  Appreciate it

cmd/minikube/cmd/config/configure.go
panyam: Initially the value was changeme but making it empty felt was a bit more riskier (so left it).

nirs: I would keep the previous value since we don't need to change this to support configuration file. We can consider a better way later.

panyam: Sounds good.  Ive changed this back to it.

cmd/minikube/cmd/config/configure_registry_creds.go
```diff
@@ -0,0 +1,257 @@
+package config
```
medyagh, line 1: this new files needs copyright boiler plate to be added to the header

medyagh (changes requested):

PR #20288, opened 2025-01-18
Added additional checks for 9p support

### LINKED ISSUE #19789: 9P detecting logic blocking when 9P is present
In https://github.com/kubernetes/minikube/pull/18995 we added a pre-check to check for the existence of 9P prior to starting the mount. This looks for the existence of `/lib/modules/<module>/kernel/fs/9p`. If it's not found we terminate with the following error: `❌  Exiting due to HOST_UNSUPPORTED: The host does not support filesystem 9p.`

However, it seems 9P can be located elsewhere on the system. For example, Cloud Shell, trying to use mount returns `❌  Exiting due to HOST_UNSUPPORTED: The host does not support filesystem 9p.`, but removing the check logic mounting works:
```
$ minikube mount ./thing123:/here
📁  Mounting host path ./thing123 into VM as /here ...
    ▪ Mount type:   9p
    ▪ User ID:      docker
    ▪ Group ID:     docker
    ▪ Version:      9p2000.L
    ▪ Message Size: 262144
    ▪ Options:      map[]
    ▪ Bind Address: 192.168.49.1:43445
✅  Successfully mounted ./thing123 to /here

```

I've verified that `/lib/modules/<module>/kernel/fs/9p` doesn't exist, but can see it in the following:
```
$ cat /proc/filesystems
nodev	sysfs
nodev	tmpfs
nodev	bdev
nodev	proc
nodev	cgroup
nodev	cgroup2
nodev	cpuset
nodev	devtmpfs
nodev	binfmt_misc
nodev	debugfs
nodev	tracefs
nodev	securityfs
nodev	sockfs
nodev	bpf
nodev	pipefs
nodev	ramfs
nodev	hugetlbfs
nodev	devpts
	ext3
	ext2
	ext4
nodev	autofs
nodev	overlay
	xfs
nodev	9p
nodev	efivarfs
nodev	mqueue
nodev	pstore
nodev	configfs
	fuseblk
nodev	fuse
nodev	fusectl
```

So we should do a different/additional check to ensure that we're not blocking on when 9P is in fact usable.

xcarolan:
fixes #19789

### DIFF
--- a/pkg/minikube/detect/detect_linux.go
+++ b/pkg/minikube/detect/detect_linux.go
@@ -20,8 +20,10 @@ package detect
 
 import (
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	"golang.org/x/sys/unix"
 )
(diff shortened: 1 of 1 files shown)


xcarolan:
@medyagh - I don't think these are valid error - are there issues with Jenkins?

medyagh (approved):
the failure looks like flake

ComradeProgrammer (approved):

PR #20291, opened 2025-01-19
release: fix missing kicbase asset in release

ComradeProgrammer:
fix: fix missing kicbase in release

### DIFF
--- a/hack/jenkins/release_build_and_upload.sh
+++ b/hack/jenkins/release_build_and_upload.sh
@@ -110,16 +110,7 @@ cp "out/minikube-${RPM_VERSION}-0.armv7hl.rpm" out/minikube-latest.armv7hl.rpm
 cp "out/minikube-${RPM_VERSION}-0.ppc64le.rpm" out/minikube-latest.ppc64le.rpm
 cp "out/minikube-${RPM_VERSION}-0.s390x.rpm" out/minikube-latest.s390x.rpm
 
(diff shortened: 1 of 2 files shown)


hack/jenkins/release_build_and_upload.sh
```diff
@@ -136,5 +127,19 @@ do
   docker rmi -f ${IMAGE_NAME}
 done
 
+
+# upload to google bucket
+gsutil -m cp out/* "gs://$BUCKET/releases/$TAGNAME/"
+
+# Update "latest" release for non-beta/non-alpha builds
```
medyagh, line 134: @ComradeProgrammer is this releated to this PR or is this something generally fixing ? what is exactly doing?

hack/jenkins/release_build_and_upload.sh
```diff
@@ -136,5 +127,19 @@ do
   docker rmi -f ${IMAGE_NAME}
 done
 
+
+# upload to google bucket
+gsutil -m cp out/* "gs://$BUCKET/releases/$TAGNAME/"
```
medyagh, line 132: there might be things in out folder that we dont want ? maybe like Linters binaries or somethign like that ? and why this is added to this PR? before this PR we didnt have artifacts in the GCS bucket?

medyagh (changes requested):

hack/jenkins/release_build_and_upload.sh
medyagh: nevermind we are just moving this code

PR #20400, opened 2025-02-10
Fix bootpd check on macOS >= 15

### LINKED ISSUE #20399: Wrong check for bootpd firewall configuration on macOS 15.3
### What Happened?

I'm using socket_vmnet installed as launched service, and [vmnet-helper](https://github.com/nirs/vmnet-helper). Both using bootpd for dhcp and they work fine.

When starting minkube, it complains that bootpd is blocked by the firewall and suggest to configure the firewall:

```
% minikube start --driver qemu --network socket_vmnet
😄  minikube v1.35.0 on Darwin 15.3 (arm64)
🔑  Your firewall is blocking bootpd which is required for this configuration. The following commands will be executed to unblock bootpd:

    $ sudo /usr/libexec/ApplicationFirewall/socketfilterfw --add /usr/libexec/bootpd 
    $ sudo /usr/libexec/ApplicationFirewall/socketfilterfw --unblock /usr/libexec/bootpd 


Password:
```

The check is wrong (I did not test why yet), since bootpd is working file on this system.

The suggested commands are also wrong, the right commands to suggest are:
https://github.com/lima-vm/socket_vmnet?tab=readme-ov-file#ip-address-is-not-assigned

But on macOS >= 15.2 (and maybe 15.1), bootdp is not listed in the firewall rules:

```
% sudo /usr/libexec/ApplicationFirewall/socketfilterfw --list | grep bootpd
(nothing)
```

I guess bootpd is allowed by default by some internal rules.

Fix:
- Find a more reliable way to detect if bootpd is configured properly
- Or remove the check and recommend a solution
- Don't run commands for the user, or let the user continue

### Attach the log file

[log.txt](https://github.com/user-attachments/files/18740718/log.txt)

### Operating System

macOS (Default)

### Driver

QEMU

nirs:
On macOS >= 15 bootpd is likely allowed to receive incoming connections as built-in software, and it will not be listed in the allowed applications. In this case we decide that bootpd is blocked and force the user to try to add and unblock it, which will never succeed.

Fixed using the new --getallowedsigned option. If the option is enabled, we know that bootpd is not blocked. If the option is not enabled, or the fails, we fallback to checking the list.

Tested on macOS 15.3.

fixes #20399

### DIFF
--- a/pkg/minikube/firewall/firewall.go
+++ b/pkg/minikube/firewall/firewall.go
@@ -46,6 +46,16 @@ func IsBootpdBlocked(cc config.ClusterConfig) bool {
 	if regexp.MustCompile(`Firewall is disabled`).Match(out) {
 		return false
 	}
+	out, err = exec.Command("/usr/libexec/ApplicationFirewall/socketfilterfw", "--getallowsigned").Output()
+	if err != nil {
+		// macOS < 15 or other issue: need to use --list.
+		klog.Warningf("failed to list firewall allowedsinged option: %v", err)
+	} else {
+		// macOS >= 15: bootpd may be allowed as builtin software.
+		if regexp.MustCompile(`Automatically allow built-in signed software ENABLED`).Match(out) {
+			return false
+		}
+	}
 	out, err = exec.Command("/usr/libexec/ApplicationFirewall/socketfilterfw", "--listapps").Output()
 	if err != nil {
 		klog.Warningf("failed to list firewall apps: %v", err)


medyagh (approved):
thank you for fixing this, if yiou have acess to macos 15+ do you mind pasting output of the minikube before anfter this PR?

nirs:
> do you mind pasting output of the minikube before anfter this PR?

Sure, I did not include it before because the output before it included in the issue, and the output after is just the normal output, but it is a good idea since you may not be able to test this.

## Before

```
%  minikube start --driver qemu --network socket_vmnet
😄  minikube v1.35.0 on Darwin 15.3 (arm64)
🔑  Your firewall is blocking bootpd which is required for this configuration. The following commands will be executed to unblock bootpd:

    $ sudo /usr/libexec/ApplicationFirewall/socketfilterfw --add /usr/libexec/bootpd 
    $ sudo /usr/libexec/ApplicationFirewall/socketfilterfw --unblock /usr/libexec/bootpd 


Password:
```

## After

```
% out/minikube start --driver qemu --network socket_vmnet
😄  minikube v1.35.0 on Darwin 15.3 (arm64)
    > minikube-v1.35.0-arm64.iso....:  65 B / 65 B [---------] 100.00% ? p/s 0s
    > minikube-v1.35.0-arm64.iso:  393.15 MiB / 393.15 MiB  100.00% 26.76 MiB p
💾  Downloading Kubernetes v1.32.1 preload ...
    > preloaded-images-k8s-v18-v1...:  314.96 MiB / 314.96 MiB  100.00% 22.64 M
🔥  Creating qemu2 VM (CPUs=2, Memory=6000MB, Disk=20000MB) ...
🐳  Preparing Kubernetes v1.32.1 on Docker 27.4.0 ...
    ▪ Configuring RBAC rules ...
```

Logs:
[log.txt](https://github.com/user-attachments/files/18751120/log.txt)

Note that I don't have any machine with macOS 14. We need to find someone with older version to test this.

PR #20406, opened 2025-02-11
fix lint issue

prezha:
fix lint complaint:
```
= make lint =============================================================
pkg/minikube/firewall/firewall.go:53:9: elseif: can replace 'else {if cond {}}' with 'else if cond {}' (gocritic)
	} else {
	       ^
make[1]: *** [Makefile:537: lint-ci] Error 1
```

### DIFF
--- a/pkg/minikube/firewall/firewall.go
+++ b/pkg/minikube/firewall/firewall.go
@@ -50,11 +50,9 @@ func IsBootpdBlocked(cc config.ClusterConfig) bool {
 	if err != nil {
 		// macOS < 15 or other issue: need to use --list.
 		klog.Warningf("failed to list firewall allowedsinged option: %v", err)
-	} else {
-		// macOS >= 15: bootpd may be allowed as builtin software.
-		if regexp.MustCompile(`Automatically allow built-in signed software ENABLED`).Match(out) {
-			return false
-		}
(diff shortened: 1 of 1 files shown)


pkg/minikube/firewall/firewall.go
```diff
@@ -50,11 +50,9 @@ func IsBootpdBlocked(cc config.ClusterConfig) bool {
...
 		// macOS < 15 or other issue: need to use --list.
 		klog.Warningf("failed to list firewall allowedsinged option: %v", err)
-	} else {
-		// macOS >= 15: bootpd may be allowed as builtin software.
-		if regexp.MustCompile(`Automatically allow built-in signed software ENABLED`).Match(out) {
-			return false
-		}
+		// macOS >= 15: bootpd may be allowed as builtin software
+	} else if regexp.MustCompile(`Automatically allow built-in signed software ENABLED`).Match(out) {
+		return false
```
nirs, line 55: The comment is misplaced now, it should be inside the condition, like the err != nil above.

But the new code is less clear and harder to change. The best way is to disable this linter. "elseif can replace else" is correct, but it does not mean it "should" replace else.

nirs: Sure we can merge as is, we can move the comment later.

PR #20440, opened 2025-02-21
docs: fix headlamp addon instructions to use new k8s way of generating a token

adwait-godbole:
Fixes headlamp-k8s/headlamp#2923

### DIFF
--- a/site/content/en/docs/handbook/addons/headlamp.md
+++ b/site/content/en/docs/handbook/addons/headlamp.md
@@ -24,15 +24,14 @@ minikube service headlamp -n headlamp
 To authenticate in Headlamp, fetch the Authentication Token using the following command:
 
 ```shell script
-export SECRET=$(kubectl get secrets --namespace headlamp -o custom-columns=":metadata.name" | grep "headlamp-token")
-kubectl get secret $SECRET --namespace headlamp --template=\{\{.data.token\}\} | base64 --decode
-``` 
+kubectl create token headlamp --duration 24h -n headlamp
+```
 
 Headlamp can display more detailed information when metrics-server is installed. To install it, run:
 
 ```shell script
-minikube addons enable metrics-server	
-```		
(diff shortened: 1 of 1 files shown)


site/content/en/docs/handbook/addons/headlamp.md
```diff
@@ -24,15 +24,14 @@ minikube service headlamp -n headlamp
 To authenticate in Headlamp, fetch the Authentication Token using the following command:
 
 ```shell script
-export SECRET=$(kubectl get secrets --namespace headlamp -o custom-columns=":metadata.name" | grep "headlamp-token")
-kubectl get secret $SECRET --namespace headlamp --template=\{\{.data.token\}\} | base64 --decode
-``` 
+kubectl create token headlamp --duration 24h -n headlamp
```
illume, line 27: I don't think it should be 24h. For a development system this is a bit tedious to have to create a new token every day.

```
kubectl create token headlamp-admin -n kube-system
```

If leaving it without an expiry is too much, what about a line telling the users about removing the duration if their minikube is just for development?

illume:
Hi.

We give similar instructions for creating a token for headlamp addon with Kubernetes 1.24+ here: https://headlamp.dev/docs/latest/installation/

adwait-godbole:
> Hi.

> We give similar instructions for creating a token for headlamp addon with Kubernetes 1.24+ here: https://headlamp.dev/docs/latest/installation/

Hi @illume. Should we give instructions for both Kubernetes 1.24+ and older versions as well?

illume:
@adwait-godbole 

> Hi @illume. Should we keep instructions for both Kubernetes 1.24+ and older versions as well? or just keep this new command assuming user is running the latest version of k8s?

As you have it is fine. This has been the way to do it for quite some time now.

joaquimrocha:
new friendly ping for @medyagh 😃

PR #20442, opened 2025-02-23
ui: clarify need for bidirectional firewall rules for mount

glerb:
Example for Ubuntu, where 192.168.49.1 is the host interface and 192.168.49.2 is the minikube IP address:

$ sudo ufw allow proto tcp from 192.168.49.1 to 192.168.49.2 port 22 
$ sudo ufw allow proto tcp from 192.168.49.2 to 192.168.49.1 port 2222 
$ minikube mount ./profiles:/var/lib/kubelet/seccomp/profiles --port=2222


medyagh (approved):

nirs:
@glerb can you rebase on master? We fixed 2 tests that used to fail in Prow. The next build will be merged automatically if we don't hit any flaky test.

PR #20470, opened 2025-02-26
Use preload image for worker nodes

Diff-fusion:
Currently when creating a worker node it will not be filled with any container images.
So it will pull the required images from the upstream registry.
This makes it impossible to use a multi node cluster in an offline environment.
This PR will changes the behavior to also copy the preload image to worker nodes.

Log before this patch: [without-preload.log](https://github.com/user-attachments/files/19521917/without-preload.log)
Log after this patch: [with-preload.log](https://github.com/user-attachments/files/19521918/with-preload.log)

### DIFF
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -163,6 +163,9 @@ func Start(starter Starter) (*kubeconfig.Settings, error) { // nolint:gocyclo
 			}()
 		}
 	} else {
+		if err := cr.Preload(*starter.Cfg); err != nil {
+			return nil, errors.Wrap(err, "preload failed")
+		}
 		bs, err = cluster.Bootstrapper(starter.MachineAPI, viper.GetString(cmdcfg.Bootstrapper), *starter.Cfg, starter.Runner)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


medyagh (commented):
good catch ! I didnt know our multinode was not using preload. 
thank you for this PR do you mind pasting in the PR description Before And After this PR, (--alsologtostderr  with verbose logs to see this is working) ?

and also we should also add this to our multi node integration test

medyagh:
@Diff-fusion would you be interested to add a small test to the multi node test?
https://github.com/kubernetes/minikube/blob/3332297bfe90dd26fc37e689284e8f2c00e533bd/test/integration/multinode_test.go#L39

you could check for the Log line in the worker nodes for using the preload ?

medyagh:
@ComradeProgrammer can you plz check this PR to verify it is correct

Diff-fusion:
I attached the logs from before and after this change to the initial comment. Also I changed the error behavior to just log a message instead of failing the start. This seems more reasonable here.
Regarding the test: Should I check the logs in the existing `validateMultiNodeStart` or create a new test case which adds another node to the cluster? Or is there some way to access the log of a previous case from a later one?

PR #20504, opened 2025-03-09
vfkit: Graceful shutdown on Stop()

### LINKED ISSUE #20503: minikube stop very slow with vfkit driver
### What Happened?

Stopping minikube driver is very slow due to unreliable implementation, incorrect error handling, and pointless backup attempts.

## To reproduce

1. Configuration

```console
minikube config set vm-driver vfkit
minikube config set container-runtime containerd
```

2. Start cluster

```console
minikube start
```
3. Stop cluster

```console
minikube stop --v=4 --alsologtostderr 2>stop.log
```

## Timeline

1. Stopping node started

```console
I0309 15:00:42.297876   13723 out.go:177] * Stopping node "minikube"  ...
```

2. Performing first backup

```console
I0309 15:00:42.307878   13723 machine.go:156] backing up vm config to /var/lib/minikube/backup: [/etc/cni /etc/kubernetes]
I0309 15:00:42.307953   13723 ssh_runner.go:195] Run: sudo mkdir -p /var/lib/minikube/backup
I0309 15:00:42.307982   13723 sshutil.go:53] new ssh client: &{IP:192.168.106.7 Port:22 SSHKeyPath:/Users/nir/.minikube/machines/minikube/id_rsa Username:docker}
I0309 15:00:42.339888   13723 ssh_runner.go:195] Run: sudo rsync --archive --relative /etc/cni /var/lib/minikube/backup
I0309 15:00:42.386662   13723 ssh_runner.go:195] Run: sudo rsync --archive --relative /etc/kubernetes /var/lib/minikube/backup
```

3. Stopping the driver

```console
I0309 15:00:42.433319   13723 main.go:141] libmachine: Stopping "minikube"...
I0309 15:00:42.433868   13723 main.go:141] libmachine: get state: {State:VirtualMachineStateRunning}
```

We got vfkit process via the HTTP API.

4. Trying to set state to `HardStop` via HTTP API

```console
I0309 15:00:42.452986   13723 stop.go:66] stop err: Post "http://_/vm/state": EOF
W0309 15:00:42.453070   13723 stop.go:165] stop host returned error: Temporary Error: stop: Post "http://_/vm/state": EOF
```

This always fails with EOF - maybe because vfkit is exiting immediately before returning the response.

5. Starting retry

```console
I0309 15:00:42.453102   13723 retry.go:31] will retry after 1.148164137s: Temporary Error: stop: Post "http://_/vm/state": EOF
```

6. Stopping the host again

```console
I0309 15:00:43.601796   13723 stop.go:39] StopHost: minikube
I0309 15:00:43.608726   13723 out.go:177] * Stopping node "minikube"  ...
I0309 15:00:43.616885   13723 machine.go:156] backing up vm config to /var/lib/minikube/backup: [/etc/cni /etc/kubernetes]
I0309 15:00:43.617271   13723 ssh_runner.go:195] Run: sudo mkdir -p /var/lib/minikube/backup
I0309 15:00:43.617329   13723 sshutil.go:53] new ssh client: &{IP:192.168.106.7 Port:22 SSHKeyPath:/Users/nir/.minikube/machines/minikube/id_rsa Username:docker}
W0309 15:01:58.617799   13723 sshutil.go:64] dial failure (will retry): dial tcp 192.168.106.7:22: connect: operation timed out
W0309 15:01:58.618132   13723 stop.go:55] failed to complete vm config backup (will continue): create dir: NewSession: new client: new client: dial tcp 192.168.106.7:22: connect: operation timed out
```

The retry attempt is in the wrong level, or maybe the backup is in the wrong level. We just backed up this host, and there is no reason to repeat the backup.

The backup attempt times out after 135 seconds, since the vfkit process is not running at this point.

7. Stopping the driver gain

```console
I0309 15:01:58.618315   13723 main.go:141] libmachine: Stopping "minikube"...
I0309 15:01:58.619098   13723 stop.go:66] stop err: Machine "minikube" is already stopped.
I0309 15:01:58.619145   13723 stop.go:69] host is already stopped
```

The stop attempt fails because the host is already stopped. It should succeed without bogus errors.

## Issues

- Using `HardStop` instead of `Stop` - does not do graceful shutdown of the host
- Using `HardStop` always fails, and we don't handle the error properly
- Retry is in the wrong level, trying to backup again after a successful backup
- Backup times out because the host is not running and previous error not handled properly

### Attach the log file

I tried several times and the behavior is same.

[stop.log](https://github.com/user-attachments/files/19149836/stop.log)

### Operating System

macOS (Default)

### Driver

N/A

nirs:
We use `HardStop` which seems to do a forced shutdown instead of graceful shutdown, and to make things worse, always fails with:

    I0309 15:00:42.452986   13723 stop.go:66] stop err: Post "http://_/vm/state": EOF

Which leads to unneeded retry and pointless backup attempt timing out after 135 seconds because vkfit was stopped.

With this change we do a graceful shutdown, and the time to stop the cluster decreased from 135 seconds to 3 seconds (45 times faster).

Example stop log:

    I0309 15:34:33.104429   14440 main.go:141] libmachine: Stopping "minikube"...
    I0309 15:34:33.105225   14440 main.go:141] libmachine: get state: {State:VirtualMachineStateRunning}
    I0309 15:34:33.105799   14440 main.go:141] libmachine: set state: {State:Stop}
    I0309 15:34:33.106099   14440 main.go:141] libmachine: get state: {State:VirtualMachineStateRunning}
    I0309 15:34:36.109380   14440 main.go:141] libmachine: Machine "minikube" was stopped.

Fixes #20503

### DIFF
--- a/pkg/drivers/vfkit/vfkit.go
+++ b/pkg/drivers/vfkit/vfkit.go
@@ -313,7 +313,7 @@ func isBootpdError(err error) bool {
 }
 
 func (d *Driver) Stop() error {
-	if err := d.SetVFKitState("HardStop"); err != nil {
+	if err := d.SetVFKitState("Stop"); err != nil {
 		return err
 	}
 	return nil


pkg/drivers/vfkit/vfkit.go
```diff
@@ -313,7 +313,7 @@ func isBootpdError(err error) bool {
 }
 
 func (d *Driver) Stop() error {
-	if err := d.SetVFKitState("HardStop"); err != nil {
+	if err := d.SetVFKitState("Stop"); err != nil {
```
cfergeau, line 316: Might be worth to try `HardStop` if  `Stop` fails?

cfergeau (commented):
Looks good to me. Unrelated, but there's a call to `d.SetVFKitState("Stop");` in `Remove` which could be changed to `HardStop`

pkg/drivers/vfkit/vfkit.go
nirs: Possible, but usually if `Stop` fails vfkit is likely not running, so `HadrStop` will also fail, and it typically fails.

medyagh: I agree we should do a HardStop if this fails, we have similar code 
we have a package called "retry" I think after a two times of retry with normal stop it should do a hard stop

medyagh (changes requested):

pkg/drivers/vfkit/vfkit.go
medyagh: here is an example of using it https://github.com/medyagh/minikube/blob/d510d46696d6e3588921d51d8daf41297e7fdc8d/pkg/minikube/machine/stop.go#L72

medyagh (approved):

pkg/drivers/vfkit/vfkit.go
nirs: The current code does not do retries in the driver, assuming that the caller doing the retries. Lets fix the issues first in the driver, and then we can evaluate if stop should retry in the driver, and apply it to more drivers.

PR #20506, opened 2025-03-09
vfkit: More robust state management

nirs:
We get and set vfkit state using the HTTP API. This is not very robust since the API is provided by the process we are terminating:

- Setting the state to `HardStop` typically fails when vfkit is terminated, leading to unneeded retries and delays.
- Setting the state to `Stop` may fail if vfkit is already stopped or shutting down, or does not listen to the socket.
- Getting the state is racy, since vfkit may be shutting down or terminate right after checking the pid (time of check, time of use).
- The pidfile was not removed in all cases, or removed too late.

This change fixes the issues by using vfkit API only for setting state, and falling back to process package pid management functions if the API fails.

### DIFF
--- a/pkg/drivers/vfkit/vfkit.go
+++ b/pkg/drivers/vfkit/vfkit.go
@@ -297,7 +297,29 @@ func isBootpdError(err error) bool {
 
 func (d *Driver) Stop() error {
 	if err := d.SetVFKitState("Stop"); err != nil {
-		return err
+		// vfkit may be already stopped, shutting down, or not listening.
+		// We don't fallback to "HardStop" since it typically fails due to
+		// https://github.com/crc-org/vfkit/issues/277.
+		log.Debugf("Failed to set vfkit state to 'Stop': %s", err)
+		pidfile := d.pidfilePath()
+		pid, err := process.ReadPidfile(pidfile)
+		if err != nil {
+			if !errors.Is(err, os.ErrNotExist) {
+				return err
+			}
+			// No pidfile.
+			return nil
+		}
+		if err := process.Terminate(pid, "vfkit"); err != nil {
+			if err != os.ErrProcessDone {
+				return err
+			}
+			// No process, stale pidfile.
+			if err := os.Remove(pidfile); err != nil {
+				log.Debugf("failed to remove %q: %s", pidfile, err)
+			}
+			return nil
+		}
 	}
 	return nil
 }
--- a/pkg/drivers/vfkit/vfkit.go
+++ b/pkg/drivers/vfkit/vfkit.go
@@ -30,7 +30,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -47,6 +46,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/exit"
 	"k8s.io/minikube/pkg/minikube/firewall"
 	"k8s.io/minikube/pkg/minikube/out"
+	"k8s.io/minikube/pkg/minikube/process"
 	"k8s.io/minikube/pkg/minikube/reason"
 	"k8s.io/minikube/pkg/minikube/style"
 )
@@ -145,20 +145,17 @@ func checkPid(pid int) error {
 }
 
 func (d *Driver) GetState() (state.State, error) {
-	if _, err := os.Stat(d.pidfilePath()); err != nil {
+	pidfile := d.pidfilePath()
+	pid, err := process.ReadPidfile(pidfile)
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return state.Error, err
+		}
 		return state.Stopped, nil
-	}
-	p, err := os.ReadFile(d.pidfilePath())
-	if err != nil {
-		return state.Error, err
-	}
-	pid, err := strconv.Atoi(strings.TrimSpace(string(p)))
-	if err != nil {
-		return state.Error, err
 	}
 	if err := checkPid(pid); err != nil {
 		// No pid, remove pidfile
-		os.Remove(d.pidfilePath())
+		os.Remove(pidfile)
 		return state.Stopped, nil
 	}
 	ret, err := d.GetVFKitState()
@@ -260,11 +257,9 @@ func (d *Driver) Start() error {
 	if err := cmd.Start(); err != nil {
 		return err
 	}
-	pid := cmd.Process.Pid
-	if err := os.WriteFile(d.pidfilePath(), []byte(fmt.Sprintf("%v", pid)), 0600); err != nil {
+	if err := process.WritePidfile(d.pidfilePath(), cmd.Process.Pid); err != nil {
 		return err
 	}
-
 	if err := d.setupIP(mac); err != nil {
 		return err
 	}
--- a/pkg/drivers/vfkit/vfkit.go
+++ b/pkg/drivers/vfkit/vfkit.go
@@ -158,17 +158,7 @@ func (d *Driver) GetState() (state.State, error) {
 		os.Remove(pidfile)
 		return state.Stopped, nil
 	}
-	ret, err := d.GetVFKitState()
-	if err != nil {
-		return state.Error, err
-	}
-	switch ret {
-	case "running", "VirtualMachineStateRunning":
-		return state.Running, nil
-	case "stopped", "VirtualMachineStateStopped":
-		return state.Stopped, nil
-	}
-	return state.None, nil
+	return state.Running, nil
 }
 
 func (d *Driver) Create() error {
@@ -322,11 +312,6 @@ func (d *Driver) Remove() error {
 	if s == state.Running {
 		if err := d.Kill(); err != nil {
 			return errors.Wrap(err, "kill")
-		}
-	}
-	if s != state.Stopped {
-		if err := d.SetVFKitState("Stop"); err != nil {
-			return errors.Wrap(err, "quit")
 		}
 	}
 	return nil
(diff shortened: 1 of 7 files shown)


pkg/drivers/vfkit/vfkit.go
```diff
@@ -136,6 +136,49 @@ func (d *Driver) GetIP() (string, error) {
...
+		}
+		// Already stopped.
+		os.Remove(pidfile)
+		return nil
+	}
+	process, err := os.FindProcess(pid)
+	if err != nil {
+		return err
+	}
+	log.Infof("Sending signal %q to pid %v", sig, pid)
```
cfergeau, line 170: `crc` checks that the process name contains `vfkit` in case the pid file has stale content, and the pid was reused by some unrelated process.

pkg/drivers/vfkit/vfkit.go
```diff
@@ -313,8 +341,10 @@ func isBootpdError(err error) bool {
 }
 
 func (d *Driver) Stop() error {
-	if err := d.SetVFKitState("HardStop"); err != nil {
-		return err
+	if err := d.SetVFKitState("Stop"); err != nil {
+		// vfkit may be already stopped, shutting down, or not listening.
+		log.Debugf("Failed to set vfkit state to 'Stop': %s", err)
+		return signalPidfile(d.pidfilePath(), syscall.SIGTERM)
 	}
```
cfergeau, line 348: maybe remove the pid file here so that `GetState` returns `Stopped` after `Stop` was called?

nirs: signalPidfile removes the pid when the process does not exist. But if the signal is sent the process still exists and the caller will discover it by polling on GetState(). Stop() is not responsible for waiting until the process is stopped, the caller is implementing this for all drivers.

cfergeau: What I mean is that if ` d.SetVFKitState("Stop")` succeeds, you have a small window for a race when `GetState` will return `running` while vfkit is `stopping`. It's probably not really important, as `GetState` will eventually get to a consistent state.

pkg/drivers/vfkit/vfkit.go
cfergeau: It is https://github.com/crc-org/crc/blob/4bdcc16077a84583ad85ea5b8b75eefab13ba82d/pkg/drivers/vfkit/driver_darwin.go#L420-L438

pkg/drivers/vfkit/vfkit.go
```diff
@@ -136,6 +136,49 @@ func (d *Driver) GetIP() (string, error) {
 	return d.IPAddress, nil
 }
 
+func writePidfile(pidfile string, pid int) error {
```
medyagh, line 139: there might be a util or lib that we are using in other parts of the code for writing or Locking PID files

pkg/drivers/vfkit/vfkit.go
```diff
@@ -313,8 +341,10 @@ func isBootpdError(err error) bool {
 }
 
 func (d *Driver) Stop() error {
-	if err := d.SetVFKitState("HardStop"); err != nil {
-		return err
+	if err := d.SetVFKitState("Stop"); err != nil {
+		// vfkit may be already stopped, shutting down, or not listening.
```
medyagh, line 345: I would still like to see we try to Force Stop using internal command first 
and if Vfkit force stop is not doing the right thing we should create an issue
and comment the link to the issue here
so when that issue is fixed we can clean up this code

medyagh (changes requested):

pkg/drivers/vfkit/vfkit.go
nirs: I'll create an issue for vfkit and add a link here.

I can add fallback using "HardStop", and then fallback to sending a signal. But the HardStop is very likely to fail so this will only add more failure logs to the process without improving anything. I can post another pr with this change later for testing.

pkg/drivers/vfkit/vfkit.go
```diff
@@ -153,6 +153,32 @@ func readPidfile(pidfile string) (int, error) {
...
+		os.Remove(pidfile)
+		return nil
+	}
+	process, err := os.FindProcess(pid)
+	if err != nil {
+		return err
+	}
+	log.Infof("Sending signal %q to pid %v", sig, pid)
+	if err := process.Signal(sig); err != nil {
+		if err != os.ErrProcessDone {
```
nirs, line 172: It may be more portable to use process.Kill() instead of process.Signal(SIGKILL).

cfergeau: NB: one issue with `Kill` / `SIGKILL` is that the process does not have the opportunity to catch the signal to do some cleanup on its own before exiting, see https://github.com/containers/gvisor-tap-vsock/issues/485 for an example of this, but vfkit could have exactly the same problem.
This ties back to your comment in https://github.com/crc-org/vfkit/issues/278, vfkit client library should provide code which does the right thing.

nirs: Yes, kill should be the last thing we try if stop did not work after some retries. Minikube use kill only when deleting a cluster, but maybe it should be better to try to stop gracefully and fallback to kill if there is no other way.

pkg/drivers/vfkit/vfkit.go
nirs: We have this issue in vfkit:
https://github.com/crc-org/vfkit/issues/277

pkg/drivers/vfkit/vfkit.go
nirs: This is implemented now using the process package #20528.

pkg/drivers/vfkit/vfkit.go
nirs: Using now the process package #20528, which uses the same method used in other packages for managing pids.

nirs:
@medyagh I think this version addresses all the comments. It includes the new process package added in #20528.

pkg/drivers/vfkit/vfkit.go
```diff
@@ -136,42 +136,24 @@ func (d *Driver) GetIP() (string, error) {
...
+	exists, err := process.Exists(pid, "vfkit")
 	if err != nil {
 		return state.Error, err
 	}
-	switch ret {
-	case "running", "VirtualMachineStateRunning":
-		return state.Running, nil
-	case "stopped", "VirtualMachineStateStopped":
+	if !exists {
+		os.Remove(pidfile)
```
medyagh, line 153: doesm't this return error ? if not wanna return error at least we could log it for the verbose log

pkg/drivers/vfkit/vfkit.go
```diff
@@ -314,7 +294,25 @@ func isBootpdError(err error) bool {
 
 func (d *Driver) Stop() error {
 	if err := d.SetVFKitState("Stop"); err != nil {
-		return err
+		// vfkit may be already stopped, shutting down, or not listening.
+		// We don't fallback to "HardStop" since it typically fails due to
+		// https://github.com/crc-org/vfkit/issues/277.
```
medyagh, line 299: thank you for liking the related issue :)

pkg/drivers/vfkit/vfkit.go
```diff
@@ -314,7 +294,25 @@ func isBootpdError(err error) bool {
...
+		log.Debugf("Failed to set vfkit state to 'Stop': %s", err)
+		pidfile := d.pidfilePath()
+		pid, err := process.ReadPidfile(pidfile)
+		if err != nil {
+			if !errors.Is(err, os.ErrNotExist) {
+				return err
+			}
+			return nil
```
medyagh, line 307: maybe add a comment saying why this is fine to return nil if errr is not Os.NotExists

pkg/drivers/vfkit/vfkit.go
nirs: If we treat this as an error, stop() or kill() may never complete, even when we know that the process is not running. So I think we should not fail the call, but we should log a debug message.

pkg/drivers/vfkit/vfkit.go
nirs: It looks obvious to me but I think a comment like

  // No pidfile
  return nil

Can be nice.

pkg/drivers/vfkit/vfkit.go
nirs: Resolved by logging the error. We have 3 instances of this, one existing (GetState()), and 2 new (Stop(), Kill()). Current version log the error in all instances.

pkg/drivers/vfkit/vfkit.go
nirs: Add comments in all branches in the 3 instances
- No pidfile
- No process, stale pidfile

nirs:
@medyagh current version addresses all the comments and should be ready for merge.

nirs:
@medyagh what is the next step here? are the failed test real failure or test flakes?

medyagh:
@nirs thanks for your patience, been a busy week, I will re-run the tests i doubt the flakes be related to your PR

medyagh:
These tests have been increasing to flake and not related to your PR


Rank	Test Name	Recent Flake Percentage	Growth (since last 15 days)
1	TestMultiControlPlane/serial/DeployApp	5.92% (10/169)	+5.34%
2	TestSkaffold	5.92% (10/169)	+3.61%
3	TestMultiControlPlane/serial/PingHostFromPods	5.92% (10/169)	+5.34%
4	TestMultiNode/serial/PingHostFrom2Pods	4.73% (8/169)	+4.73%
5	TestMultiNode/serial/DeployApp2Nodes

medyagh:
thank you @nirs I posted a comment about the flakes in slack https://kubernetes.slack.com/archives/C1F5CT6Q1/p1743612392713229

nirs:
@medyagh thanks for reviewing and merging this! rebasing hte vmnet pr...

PR #20520, opened 2025-03-11
feat: add known issue solution to K8S_API_SERVER_MISSING

### LINKED ISSUE #20518: Error message in K8S_APISERVER_MISSING
### What Happened?

Hi, after update of the Docker Desktop for MacOS (Apple Silicon) I came across following issue (img at the bottom). 
Even after multiple restarts (PC, Docker and minikube) error still occur. 

In the suggestions at the end of the error message I found information about probable problems with SELinux but that was not a issue. 
I tries many things and only one solved my issue, I had to delete cluster `minikube delete`. 

Maybe it would be a good idea to link this issue and add this solution to list at the bottom of the error message.  

### Attach the log file

I already solved this problem so every thing that I got is this error message from the terminal 

Image

### Operating System

macOS (Default)

### Driver

Docker

Ret2Me:
Add link to issue with solution to `K8S_API_SERVER_MISSING`.

fixes #20518

### DIFF
--- a/pkg/minikube/reason/known_issues.go
+++ b/pkg/minikube/reason/known_issues.go
@@ -1213,7 +1213,7 @@ var controlPlaneIssues = []match{
 			ID:       "K8S_APISERVER_MISSING",
 			ExitCode: ExControlPlaneNotFound,
 			Advice:   "Check that the provided apiserver flags are valid, and that SELinux is disabled",
-			Issues:   []int{4536, 6014},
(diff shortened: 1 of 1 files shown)


Ret2Me:
Hi, @medyagh would you mind review this really small change?

Ret2Me:
@ComradeProgrammer  Hi, would you mind to check out my MR? ^^

PR #20615, opened 2025-04-13
Update README.md

mrmonarch20:
Simple Name correction

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ See https://minikube.sigs.k8s.io/docs/
 
 ## More Examples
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,7 +52,7 @@ See https://minikube.sigs.k8s.io/docs/
 
 ## More Examples
 
-See minikube in action [here](https://minikube.sigs.k8s.io/docs/handbook/controls/)
+See Minikube in action [here](https://minikube.sigs.k8s.io/docs/handbook/controls/)
```
nirs, line 55: "minikube" is a name of a program, so it should not be called "Minikube".

PR #20636, opened 2025-04-20
Fix bug where cert files cannot have parenthesis in its name

### LINKED ISSUE #20080: Cert file naming makes Copy function in ssh_runner.go break
### What Happened?

The problem occurs when running `minikube start --embed-certs`

One of my certificates used on my system is called my_cert(1).crt. This naming does not work, since the Copy function in ssh_runner.go tries to run `scp += fmt.Sprintf(" && sudo touch -d \"%s\" %s", mtime.Format(layout), dst)`. This makes Bash run `sudo touch -d "2024-12-10 14:39:09.586983523 +0100" /usr/share/ca-certificates/my_cert(1).pem`, which is an invalid command, since the parenthesis needs to be preceded by a backslash like this:
`sudo touch -d "2024-12-10 14:39:09.586983523 +0100" /usr/share/ca-certificates/my_cert\(1\).pem`

This can of course be resolved by changing the name of the cert file.

I would attempt to fix this myself, but I am unfortunately not a go developer.

### Attach the log file

I don't think the entire log is necessary to paste here due to me finding out the cause of the bug myself, and I don't feel like sanitizing the output from secret environment variables and such. The line with the error goes as follows:

❌  Exiting due to GUEST_CERT: Failed to setup certs: Copy /home/chasar/.minikube/certs/my_cert(1).crt: sudo mkdir -p /usr/share/ca-certificates && sudo scp -t /usr/share/ca-certificates && sudo touch -d "2024-12-10 14:39:09.586983523 +0100" /usr/share/ca-certificates/my_cert(1).pem: Process exited with status 2
output: bash: -c: line 1: syntax error near unexpected token `('
bash: -c: line 1: `sudo mkdir -p /usr/share/ca-certificates && sudo scp -t /usr/share/ca-certificates && sudo touch -d "2024-12-10 14:39:09.586983523 +0100" /usr/share/ca-certificates/my_cert(1).pem'

### Operating System

Other

### Driver

None

Chasarr:
I have never once programmed in Go before, but I think this fix should work? I have to compile and run it on my work machine later next week to confirm that this works, but extra quotations can't hurt.
 
fixes #20080

### DIFF
--- a/pkg/minikube/command/ssh_runner.go
+++ b/pkg/minikube/command/ssh_runner.go
@@ -409,7 +409,7 @@ func (s *SSHRunner) Copy(f assets.CopyableFile) error {
 	if err != nil {
 		klog.Infof("error getting modtime for %s: %v", dst, err)
 	} else if mtime != (time.Time{}) {
-		scp += fmt.Sprintf(" && sudo touch -d \"%s\" %s", mtime.Format(layout), dst)
+		scp += fmt.Sprintf(" && sudo touch -d \"%s\" '%s'", mtime.Format(layout), dst)
 	}
 	out, err := sess.CombinedOutput(scp)
 	if err != nil {


medyagh:
@Chasarr thank you for this PR, do you mind putting before/after this PR in your PR description ?

medyagh:
hi [Chasarr](https://github.com/Chasarr) are you still interested to finish this PR

PR #20648, opened 2025-04-22
fix bug in parsing proxies with dashes

### LINKED ISSUE #20624: minikube fails to parse http_proxy with - in the address
### What Happened?

Hi all

I am behind a corporate proxy which address is: **internet-app.corp.thales:8000**
minikube fails to properly parse the variable to set in the minikube container:

```
  **- HTTP_PROXY=http://AN006498:*****@internet**
! You appear to be using a proxy, but your NO_PROXY environment does not include the minikube IP (192.168.49.2).
* Please see https://minikube.sigs.k8s.io/docs/handbook/vpn_and_proxy/ for more details
  - HTTPS_PROXY=http://AN006498:*****@internet
  - NO_PROXY=localhost,********
**! Failing to connect to https://registry.k8s.io/ from inside the minikube container**

```
As you can see the http_proxy is stopped at the -. I tried to set the - with ascii escape but:

`! Error parsing HTTPS_PROXY=http://AN006498:*********@internet%2dapp.corp.thales:8000, parse "http://AN006498:*********@internet%2dapp.corp.thales:8000": invalid URL escape "%2d"`

I tried to set an proxy hostname in the system, but of course the certificate is invalid due to different FQDN and minikube still fails to connect upstream for updating the container contents.

kubernetes environment seems to be ok:

```
* Preparing Kubernetes v1.32.0 on Docker 27.4.1 ...
  - env HTTP_PROXY=http://AN006498:********@internet-app.corp.thales:8000
  - env HTTPS_PROXY=http://AN006498:*********@internet-app.corp.thales:8000
  - env NO_PROXY=localhost,***********

```

### Attach the log file

no logs to be provided

### Operating System

Ubuntu

### Driver

Docker

ComradeProgrammer:
fix: fix bug in  MaskProxyPassword
FIX #20624


pkg/util/utils_test.go
```diff
@@ -250,6 +250,13 @@ func TestMaskProxyPassword(t *testing.T) {
 		{
 			input:  "https://fr@ank5t3in:an0th3erF@akeP@55word@minikube.sigs.k8s.io",
 			output: "https://fr@ank5t3in:*****@minikube.sigs.k8s.io",
+		}, {
+			input:  "http://aaa:bbb@internet-app.corp.thales:8080",
```
medyagh, line 254: @ComradeProgrammer I prefer we use psudo names not actual company names here, how about changing thales to something like "abcdcompany"

medyagh (changes requested):

pkg/util/utils_test.go
ComradeProgrammer: done

medyagh (approved):
/lgtm

PR #20657, opened 2025-04-28
small lint

LJTian:
Optimizing the code


third_party/go9p/clnt_clnt.go
```diff
@@ -221,10 +221,8 @@ func (clnt *Clnt) recv() {
 					r.Err = &Error{"invalid response", EINVAL}
 					log.Printf("TTT %v\n", r.Tc)
 					log.Printf("RRR %v\n", r.Rc)
-				} else {
```
medyagh, line 224: @LJTian we generally dont wanna touch or lint the Third_party code, it is basically like a fork for code that we can not import. but have to copy/paste

medyagh (changes requested):

third_party/go9p/clnt_clnt.go
LJTian: ok

Issue #20654, opened 2025-04-24
fix all lints for new linter


medyagh:
https://github.com/kubernetes/minikube/pull/20482

LJTian:
> /cc [@medyagh](https://github.com/medyagh) I borrowed the content of https://github.com/containerd/nerdctl/blob/main/.golangci.yml, but there are still some differences in the code style. I hope someone can discuss a solution and determine a golangci.yaml file. Then, according to the rules of golangci.yaml, the code can be gradually optimized.

"Please take a look at the issue on GitHub: https://github.com/kubernetes/minikube/issues/20654."

LJTian:
Keeping most of the existing rules and adding some new ones, you can gradually optimize the code and then relax the corresponding checking rules.
I think we should spend some time to determine the code style checking rules first, If there are too many things to fix, we should consider the order of repair.

LJTian:
Reference for file migration: https://golangci-lint.run/product/migration-guide/
Additional reference: https://github.com/golangci/golangci-lint/discussions/5618

medyagh:
I prefer we do one big PR to fix most lint issues that are easy than excluding

jaibhavaya:
I would work on this if it's decided you want to go the "one big pr to fix all the lint issues" route.

PR #20658, opened 2025-04-28
migrate golangci lint to v2

LJTian:
Update golint from v1.64.5 to v2.1.2 

add .golangci.yaml file

Fixes  https://github.com/kubernetes/minikube/issues/20654

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -132,43 +132,43 @@ linters:
         - elseif
         - regexpMust
         - deprecatedComment
-        # enabled-checks:
-        #   # Below used to be enabled, but we do not pass anymore
-        #   # - paramTypeCombine
-        #   # - octalLiteral
-        #   # - unnamedResult
-        #   # - equalFold
-        #   # - sloppyReassign
-        #   # - emptyStringTest
-        #   # - hugeParam
-        #   # - appendCombine
-        #   # - stringXbytes
-        #   # - ptrToRefParam
-        #   # - commentedOutCode
-        #   # - rangeValCopy
-        #   # - methodExprCall
-        #   # - yodaStyleExpr
-        #   # - typeUnparen
-  #   # We enabled these and we pass
-  #   - nilValReturn
-  #   - weakCond
-  #   - indexAlloc
-  #   - rangeExprCopy
-  #   - boolExprSimplify
-  #   - commentedOutImport
-  #   - docStub
-  #   - emptyFallthrough
-  #   - hexLiteral
-  #   - typeAssertChain
-  #   - unlabelStmt
-  #   - builtinShadow
-  #   - importShadow
-  #   - initClause
-  #   - nestingReduce
-  #   - unnecessaryBlock
+      enabled-checks:
+      # Below used to be enabled, but we do not pass anymore
+      # - paramTypeCombine
+      # - octalLiteral
+      # - unnamedResult
+      # - equalFold
+      # - sloppyReassign
+      # - emptyStringTest
+      # - hugeParam
+      # - appendCombine
+      # - stringXbytes
+      # - ptrToRefParam
+      # - commentedOutCode
+      # - rangeValCopy
+      # - methodExprCall
+      # - yodaStyleExpr
+      # - typeUnparen
 
-  ####### The content below is the content skipped by regular expression in the first version, 
-  ####### which has been gradually replaced by the detailed configuration above.
+      ##### TODO: fix and enable these
+      # We enabled these and we pass
+      - nilValReturn
+      # - weakCond # pkg/minikube/config/profile.go:61:9: weakCond: suspicious `cc.Nodes != nil && cc.Nodes[0].Name == node.Name`; nil check may not be enough, check for len (gocritic)
+      - indexAlloc
+      - rangeExprCopy
+      - boolExprSimplify
+      - commentedOutImport
+      # - docStub         # pkg/minikube/tunnel/kic/service_tunnel.go:51:1: docStub: silencing go lint doc-comment warnings is unadvised (gocritic)
+      - emptyFallthrough
+      - hexLiteral
+      - typeAssertChain
+      - unlabelStmt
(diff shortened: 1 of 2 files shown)


medyagh:
thank you @LJTian for picking this up ! are the Excludes in this PR new ? are they easy to fix in a follow up PR?

Makefile
```diff
@@ -525,7 +524,7 @@ out/linters/golangci-lint-$(GOLINT_VERSION):
 ifeq ($(MINIKUBE_BUILD_IN_DOCKER),y)
 lint:
 	docker run --rm -v $(pwd):/app -w /app golangci/golangci-lint:$(GOLINT_VERSION) \
-	golangci-lint run ${GOLINT_OPTIONS} --skip-dirs "cmd/drivers/kvm|cmd/drivers/hyperkit|pkg/drivers/kvm|pkg/drivers/hyperkit" ./...
+	golangci-lint run ${GOLINT_OPTIONS} --config .golangci-build-in-docker.yaml ./...
```
medyagh, line 527: why do we have a different file for build in docker? can we use the same file for simplicity ?

LJTian: Since version 2 removed parameters like --skip-dirs , the same functionality can only be achieved through configuration files.
I personally prefer to use a single configuration file to reduce maintenance costs, and I'd like to spend some time to investigate further to see if there is a better way to solve this problem.

LJTian:
> thank you @LJTian for picking this up ! are the Excludes in this PR new ? are they easy to fix in a follow up PR?

"Please take a look at the issue on GitHub: https://github.com/kubernetes/minikube/issues/20654."

.golangci.yaml
```diff
@@ -0,0 +1,212 @@
...
+          disabled: true
+        - name: unexported-naming
+          disabled: true
+        - name: struct-tag
+          disabled: true
+        - name: redundant-import-alias
+          disabled: true
+    gocritic:
+      #  See https://go-critic.com/overview.html
+      disabled-checks:
```
medyagh, line 121: are these skipped in v1 too ? can we create an issue to fix all these Skipped Lint Issues as a follow up PR

Makefile
```diff
@@ -524,8 +526,20 @@ out/linters/golangci-lint-$(GOLINT_VERSION):
 .PHONY: lint
 ifeq ($(MINIKUBE_BUILD_IN_DOCKER),y)
 lint:
-	docker run --rm -v $(pwd):/app -w /app golangci/golangci-lint:$(GOLINT_VERSION) \
-	golangci-lint run ${GOLINT_OPTIONS} --skip-dirs "cmd/drivers/kvm|cmd/drivers/hyperkit|pkg/drivers/kvm|pkg/drivers/hyperkit" ./...
+	mkdir -p ./out/linters
+	@if [ ! -f ./out/linters/yq ] ; then \
+		wget https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/${YQ_BINARY_NAME} -O ./out/linters/yq && chmod +x ./out/linters/yq; \
```
medyagh, line 531: this doesnt seem right to install things using wget, and we we need to install this ?

medyagh (changes requested):

Makefile
```diff
@@ -524,8 +526,20 @@ out/linters/golangci-lint-$(GOLINT_VERSION):
...
+	fi
+	
+	./out/linters/yq eval \
+	'.linters.exclusions.paths += [ "cmd/drivers/kvm", "cmd/drivers/hyperkit", "pkg/drivers/kvm", "pkg/drivers/hyperkit" ]' \
+	-i .golangci.yaml  &&  echo "yq add data ok"  
+	
+	docker run --rm -v `pwd`:/app -w /app golangci/golangci-lint:$(GOLINT_VERSION) \
+	bash -c "apt update && apt -y install libvirt-dev && golangci-lint run ${GOLINT_OPTIONS} ./..."
```
medyagh, line 539: did we really need to do apt update and install libvirt ?

medyagh (approved):
PR looks almost ready to merge, just need to add some comments in the code for why doing things that way
and also do we really need to apt update and install in docker run ? the image no longer has those packages by itself?

Makefile
LJTian: The main difference between these two configuration files is that some directories are skipped

.golangci.yaml
LJTian: old version info : 
# options for lint (golangci-lint)
GOLINT_OPTIONS = --timeout 7m \
	  --build-tags "${MINIKUBE_INTEGRATION_BUILD_TAGS}" \
	  --enable gofmt,goimports,gocritic,revive,gocyclo,misspell,nakedret,stylecheck,unconvert,unparam,dogsled \
	  --exclude 'variable on range scope.*in function literal|ifElseChain'

---
In our previous makfile, we mainly used regular expressions to skip some problems. In fact, we should use configuration to perform some modular checks.
We can add a new PR to fix the problems we think need to be fixed. What we need to know is that these fixes are only code style fixes and do not mean there are errors or vulnerabilities in the code.

Makefile
LJTian: Currently this code is still under debugging. I submit this PR mainly for the convenience of discussion. I will merge and submit it later. Then notify you again.

LJTian:
> PR looks almost ready to merge, just need to add some comments in the code for why doing things that way and also do we really need to apt update and install in docker run ? the image no longer has those packages by itself?

Currently this code is still under debugging. I submit this PR mainly for the convenience of discussion. I will merge and submit it later. Then notify you again.

LJTian:
> PR looks almost ready to merge, just need to add some comments in the code for why doing things that way and also do we really need to apt update and install in docker run ? the image no longer has those packages by itself?

I use the vscode space environment to verify the "build in docker" method, but it always times out. I am still looking for a new environment for verification. After the verification is completed, I will add corresponding comments and remove some redundant information.

medyagh:
> d that modifying the yq tool would cause the ".golangci.yaml

if our yaml format needs to be standardized so the yq doesnt mess it up, it is okay to change our yaml format


let me know when you are satisfied with the PR to be Ready for Review ! feel free to ping me on public minikube chat room on slack

PR #20685, opened 2025-05-01
Update qemu.md docs for homebrew/services decommission

keycodez:
## Decommissioned Tap
https://github.com/Homebrew/homebrew-services was decommissioned on March 20, 2025.

## Services Command Relocated
`brew services` still works since it is now included with brew by default https://github.com/Homebrew/brew/tree/master/Library/Homebrew/services.


medyagh (approved):

medyagh (commented):
[keycodez](https://github.com/keycodez) can you plz clarify that if this will break the users which are running older version of brew or not ?

PR #20746, opened 2025-05-13
build: Use jammy based xcgo image for hyperkit

nirs:
Image was built from:
https://github.com/nirs/xcgo

Changes from upstream:
https://github.com/neilotoole/xcgo/compare/master...nirs:xcgo:master

I also posted the change to upstream:
https://github.com/neilotoole/xcgo/pull/13

Maybe we need to add this repo under the minikube organization or
include the code in minikube repo.
    
With this I could build hyperkit using:

```console
$ make hyperkit_in_docker 
rm -f out/docker-machine-driver-hyperkit
make MINIKUBE_BUILD_IN_DOCKER=y out/docker-machine-driver-hyperkit
make[1]: Entering directory '/home/nsoffer/minikube'
docker run --rm -e GOCACHE=/app/.cache -e IN_DOCKER=1 \
	--user 1000:1000 -w /app \
	-v /home/nsoffer/minikube:/app:Z -v /home/nsoffer/go:/go:Z --init --entrypoint "" \
	quay.io/nirsof/xcgo:jammy /bin/bash -c 'CC=o64-clang CXX=o64-clang++ /usr/bin/make out/docker-machine-driver-hyperkit'
GOOS=darwin CGO_ENABLED=1 go build \
	-ldflags="-X k8s.io/minikube/pkg/drivers/hyperkit.version=v1.35.0 -X k8s.io/minikube/pkg/drivers/hyperkit.gitCommitID="ba04fc34350c9efde2c16709fc7b892ac944b623""   \
	-o out/docker-machine-driver-hyperkit k8s.io/minikube/cmd/drivers/hyperkit
make[1]: Leaving directory '/home/nsoffer/minikube'

$ file out/docker-machine-driver-hyperkit
out/docker-machine-driver-hyperkit: Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE>
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ $(shell mkdir -p $(BUILD_DIR))
 CURRENT_GIT_BRANCH ?= $(shell git branch | grep \* | cut -d ' ' -f2)
 
 # Use system python if it exists, otherwise use Docker.
-PYTHON := $(shell command -v python || echo "docker run --rm -it -v $(shell pwd):/minikube -w /minikube python python")
+PYTHON := $(shell command -v python || echo "docker run --rm -it -v $(shell pwd):/minikube:Z -w /minikube python python")
 BUILD_OS := $(shell uname -s)
 
 SHA512SUM=$(shell command -v sha512sum || echo "shasum -a 512")
@@ -189,7 +189,7 @@ endef
 
 # $(call DOCKER, image, command)
 define DOCKER
-	docker run --rm -e GOCACHE=/app/.cache -e IN_DOCKER=1 --user $(shell id -u):$(shell id -g) -w /app -v $(PWD):/app -v $(GOPATH):/go --init $(1) /bin/bash -c '$(2)'
+	docker run --rm -e GOCACHE=/app/.cache -e IN_DOCKER=1 --user $(shell id -u):$(shell id -g) -w /app -v $(PWD):/app:Z -v $(GOPATH):/go --init $(1) /bin/bash -c '$(2)'
 endef
 
 ifeq ($(BUILD_IN_DOCKER),y)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -103,7 +103,7 @@ $(shell mkdir -p $(BUILD_DIR))
 CURRENT_GIT_BRANCH ?= $(shell git branch | grep \* | cut -d ' ' -f2)
 
 # Use system python if it exists, otherwise use Docker.
-PYTHON := $(shell command -v python || echo "docker run --rm -it -v $(shell pwd):/minikube -w /minikube python python")
+PYTHON := $(shell command -v python || echo "docker run --rm -it -v $(shell pwd):/minikube:Z -w /minikube python python")
```
medyagh, line 106: curious what does :Z do here?

medyagh (commented):
~~@nirs in your example you ran, did you also set "MINIKUBE_BUILD_IN_DOCKER" to be 1  to ensure it is running in docker~~

oh never mind I see you are running it in docker! thank you

Makefile
nirs: Enable selinux for the mount, probably label it so root can access it.

nirs: From https://manpages.org/docker-run:

Labeling systems like SELinux require that proper labels are placed on volume content mounted into a container. Without a label, the security system might prevent the processes running inside the container from using the content. By default, Docker does not change the labels set by the OS.

To change a label in the container context, you can add either of two suffixes :z or :Z to the volume mount. These suffixes tell Docker to relabel file objects on the shared volumes. The z option tells Docker that two containers share the volume content. As a result, Docker labels the content with a shared content label. Shared volume labels allow all containers to read/write content. The Z option tells Docker to label the content with a private unshared label. Only the current container can use a private volume.

medyagh:
thank you @nirs for your amazing contributions you are making a difference here

Issue #20752, opened 2025-05-13
improve automation "update-cni-plugins-version" to update it everywhere


medyagh:
we have an automation that updates "cni-plugins-version" Example This PR: https://github.com/kubernetes/minikube/pull/20749

however I noticed it does NOT update a few places such as

-https://github.com/medyagh/minikube/blob/88e5a27313962b14bf811625d6ed9257c3bcc38e/hack/jenkins/installers/check_install_cni_plugins.sh#L19
- Github Action Yamls 
- https://github.com/medyagh/minikube/blob/88e5a27313962b14bf811625d6ed9257c3bcc38e/.github/workflows/pr.yml#L568
- https://github.com/medyagh/minikube/blob/88e5a27313962b14bf811625d6ed9257c3bcc38e/.github/workflows/master.yml#L570

the automation script lives here
https://github.com/kubernetes/minikube/blob/master/hack/update/cni_plugins_version/update_cni_plugins_version.go

## Expected Result
so when the automation PR runs make update-cni-plugins-version it should updated those files above too

medyagh:
@zvdy 
after merging the PR I noticed this failing in githu action
```
Run CNI_PLUGIN_VERSION=\"v1.7.1\"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100     9  100     9    0     0     43      0 --:--:-- --:--:-- --:--:--    43
tar: This does not look like a tar archive

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

example https://github.com/kubernetes/minikube/actions/runs/15037286364/job/42261436490?pr=20763

medyagh:
@zvdy we basically need to make sure instead of CNI_PLUGIN_VERSION=\"v1.7.1\" it should say CNI_PLUGIN_VERSION="v1.7.1" would you like to make a PR to fix it

zvdy:
Incorrect escaping, didn't think about this.
On it

medyagh:
was fixed ! thank you @zvdy for the quick response
https://github.com/kubernetes/minikube/actions/runs/15055064671/job/42319200866?pr=20771

PR #20769, opened 2025-05-15
CI: Fix cni escaping

zvdy:
# Properly handle regex 
Handles the regex differenciating sh and yaml files and removing `\` as discussed in #20752

# Tests

image
image
image

> cc: @medyagh


medyagh (approved):
thank you

PR #20779, opened 2025-05-18
Update SECURITY-INSIGHTS.yml

mrmonarch20:
Expiration date passed.Updated the expiration date


SECURITY-INSIGHTS.yml
```diff
@@ -1,6 +1,6 @@
 header:
   schema-version: 1.0.0
-  expiration-date: '2024-12-17T01:00:00.000Z'
+  expiration-date: '2026-06-18T01:00:00.000Z'
```
medyagh, line 3: is this something that needs to be updated manually ? is there any guidance or documentation we can follow

Issue #20803, opened 2025-05-19
Add krunkit driver


nirs:
### What Happened?

krunkit is very similar to vfkit, providing missing features like GPU passthrough. It is used by podman desktop on macOS to support GPU accelerated containers.

krunkit is mostly compatible with vkfit and works with vment-helper. It should be pretty easy to create a driver using krunkit based on the vfkit driver.

A krunkit based minikube cluster can use GPU passthrough to run GPU accelerated containers.

Links:
- [krunkit project](https://github.com/containers/krunkit)
- [GPU Accelerated Containers on Apple Silicon with libkrun and podman machine - DevConf.US 2024](https://www.youtube.com/watch?v=OyTJ8FtQaJ0)
- [RH Dan on Tech: Episode 12 - How libkrun enhances AI, Security, Containers, Virtualization, and more](https://www.youtube.com/watch?v=p8cdRVzETLY)

### Attach the log file

(not relevant)

### Operating System

macOS (Default)

### Driver

N/A

afbjorklund:
It seems like the GPU passthrough might neeed a newer kernel (6.x)

I also wonder if DRI is available in Docker, I think you need containerd?

PR #20808, opened 2025-05-21
vfkit: Bump to Preferred driver on macOs

### LINKED ISSUE #20807: minikube "auto select" driver fails to select vfkit when
on a fresh machine macos I only had vkfit installed and minikube auto driver select failed to choose it,
so when there is no other option like qemu or docker or anything else, why minikube didnt Pick Vfkit ??? thats a bug
``` 
medya-mac:~ medya$ minikube start
😄  minikube v1.35.0 on Darwin 15.5 (arm64)
💡  Alternatively you could install one of these drivers:
    ▪ hyperkit: Not installed: exec: "hyperkit": executable file not found in $PATH
    ▪ parallels: Not installed: exec: "prlctl": executable file not found in $PATH
    ▪ qemu2: Not installed: exec: "qemu-system-aarch64": executable file not found in $PATH

```

however when I specified it manually it worked

```
medya-mac:~ medya$ minikube start -d vfkit
😄  minikube v1.35.0 on Darwin 15.5 (arm64)
✨  Using the vfkit (experimental) driver based on user configuration
    > minikube-v1.35.0-arm64.iso....:  65 B / 65 B [---------] 100.00% ? p/s 0s
    > minikube-v1.35.0-arm64.iso:  393.15 MiB / 393.15 MiB  100.00% 26.78 MiB p
💾  Downloading Kubernetes v1.32.0 preload ...
    > preloaded-images-k8s-v18-v1...:  314.92 MiB / 314.92 MiB  100.00% 29.13 M
🔥  Creating vfkit VM (CPUs=2, Memory=6000MB, Disk=20000MB) ...
🐳  Preparing Kubernetes v1.32.0 on Docker 27.4.0 ...
    ▪ Configuring RBAC rules ...
```

nirs:
vfkit is using the native virtualization framework, provides good best performance and all the features needed by minikube. It is also well maintained and used by other projects like podman.

This fixes the automatic driver selection. When we start minikube on a system with both vfkit and qemu, vfkit is selected:

    % minikube start
    😄  minikube v1.35.0 on Darwin 15.5 (arm64)
    ✨  Automatically selected the vfkit driver. Other choices: qemu2, ssh, podman (experimental)
    🔥  Creating vfkit VM (CPUs=2, Memory=6000MB, Disk=20000MB) ...
    🐳  Preparing Kubernetes v1.33.1 on Docker 28.0.4 ...
        ▪ Configuring RBAC rules ...

Fixes #20807

### DIFF
--- a/pkg/minikube/registry/drvs/vfkit/vfkit.go
+++ b/pkg/minikube/registry/drvs/vfkit/vfkit.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 
 	"github.com/docker/machine/libmachine/drivers"
 	"github.com/google/uuid"
@@ -41,13 +42,17 @@ const (
 )
 
 func init() {
+	priority := registry.Unknown
+	if runtime.GOOS == "darwin" {
+		priority = registry.Preferred
+	}
 	if err := registry.Register(registry.DriverDef{
 		Name:     driver.VFKit,
 		Init:     func() drivers.Driver { return vfkit.NewDriver("", "") },
 		Config:   configure,
 		Status:   status,
 		Default:  true,
-		Priority: registry.Experimental,
+		Priority: priority,
 	}); err != nil {
 		panic(fmt.Sprintf("register failed: %v", err))
 	}
@@ -104,7 +109,7 @@ func configure(cfg config.ClusterConfig, n config.Node) (interface{}, error) {
 func status() registry.State {
 	_, err := exec.LookPath("vfkit")
 	if err != nil {
-		return registry.State{Error: err, Fix: "Run 'brew tap cfergeau/crc && brew install vfkit'", Doc: docURL}
+		return registry.State{Error: err, Fix: "Run 'brew install vfkit'", Doc: docURL}
 	}
 	return registry.State{Installed: true, Healthy: true, Running: true}
 }


pkg/minikube/registry/drvs/vfkit/vfkit.go
```diff
@@ -41,13 +42,17 @@ const (
 )
 
 func init() {
+	priority := registry.Unknown
+	if runtime.GOOS == "darwin" {
+		priority = registry.Preferred
```
medyagh, line 47: I think is better than preffered registry.Default so it be the same level as qemu,

medyagh (changes requested):

pkg/minikube/registry/drvs/vfkit/vfkit.go
```diff
@@ -104,7 +109,7 @@ func configure(cfg config.ClusterConfig, n config.Node) (interface{}, error) {
 func status() registry.State {
 	_, err := exec.LookPath("vfkit")
 	if err != nil {
-		return registry.State{Error: err, Fix: "Run 'brew tap cfergeau/crc && brew install vfkit'", Doc: docURL}
+		return registry.State{Error: err, Fix: "Run 'brew install vfkit'", Doc: docURL}
```
cfergeau, line 112: :+1: 
vfkit in cfergeau/crc will no longer be updated now that it’s in the main repo. I should remove it from this personal tap.

cfergeau (commented):
Looks good to me!

pkg/minikube/registry/drvs/vfkit/vfkit.go
medyagh: sure

cfergeau:
> Not long ago starting qemu on M4 was crashing and it took few month until they fixed the issue

For the record, this is not an isolated event, there was a similar problem around the time podman switched to vfkit, qemu hang on startup which took several weeks/months to be fixed.

> Unless we have some important features that does not work on vfkit and work on qemu. I'm not a ware of any.

vfkit has support for virtiofs and rosetta (important features off the top of my head), and the fact that Apple maintains most of the virtualization code is both a plus and a minus, but in my opinion mostly a plus. It overall works well, we had minimal issues over the years, but when there are issues, we are dependent on Apple to fix them, https://github.com/containers/podman/issues/25121 is a recent example of this. This bug only impacted some Intel macs, and is fixed with macOS 15.5
One big missing thing, but I’m not sure qemu has this either, is gpu acceleration support, which I guess could be useful to minikube (?). We’ll know at WWDC in a few weeks if they are going to add this to macOS 16 :)

nirs:
> One big missing thing, but I’m not sure qemu has this either, is gpu acceleration support, which I guess could be useful to minikube (?). We’ll know at WWDC in a few weeks if they are going to add this to macOS 16 :)

I started to work on #20803 for GPU acceleration.

afbjorklund:
The main feature of
QEMU is that it is available on all platforms, something that was previously done by VirtualBox.

Having to split minikube into different hypervisors and different runtimes (not just Docker) complicated things **a lot**.

pkg/minikube/registry/drvs/vfkit/vfkit.go
afbjorklund: From https://github.com/crc-org/vfkit/commit/1228143e04eee309db63d99da03cedcf54cc3ed3

PR #20826, opened 2025-05-23
Add krunkit driver supporting GPU acceleration on macOS

### LINKED ISSUE #20803: Add krunkit driver

nirs:
krunkit is a tool to launch configurable virtual machines using the libkrun platform, optimized for GPU accelerated virtual machines and AI workloads on Apple silicon.
    
It is mostly compatible with vfkit; the driver is a simplified copy of the vfkit driver.

To experiment with the krunkit driver see https://github.com/medyagh/ai-playground-minikube/tree/main/macos

[asciicast](https://asciinema.org/a/726193)

Benchmarking GPU compute workloads shows 2 orders of magnitude improvement with Virtio-GPU, and fully utilized GPU:

## Status

- [x] Tested using locally built iso
- [x] Start/delete clusters
- [x] GPU available in the guest
- [x] Run GPU benchmarks
- [x] Test unsupported platform (linux/amd64)
- [x] Test krunkit is not installed
- [x] Test vment-helper not installed
- [x] Test vment-helper not configured

## Integration test results

Created with iso from #21089 using:

```
env TEST_ARGS="-minikube-start-args='--driver=krunkit --iso-url=file:///Users/nir/src/minikube/minikube-arm64-minimal-k8s.iso'" make integration
```

[krunkit.tar.gz](https://github.com/user-attachments/files/21330310/krunkit.tar.gz)

Fixes #20803

### DIFF
--- a/site/content/en/docs/drivers/krunkit.md
+++ b/site/content/en/docs/drivers/krunkit.md
@@ -1,0 +1,102 @@
+---
+title: "krunkit"
+weight: 2
+aliases:
+    - /docs/reference/drivers/krunkit
+---
+
+## Overview
+
+[krunkit](https://github.com/containers/krunkit) is an open-source program for
+macOS virtualization, optimized for GPU accelerated virtual machines and AI
+workloads.
+
+## Requirements
+
+- Available only on Apple silicon.
+- Requires macOS 13 or later.
+- Requires minikube version 1.37.0 or later.
+- Requires krunkit version 0.2.2 or later.
+- Requires [vmnet-helper](https://github.com/nirs/vmnet-helper).
+
+## Installing krunkit
+
+To install krunkit run:
+
+```shell
+brew tap slp/krunkit
+brew install krunkit
+```
+
+## Networking
+
+To use the krunkit driver you must install
+[vmnet-helper](https://github.com/nirs/vmnet-helper), see installation
+instructions bellow.
+
+### Install vment-helper
+
+```shell
+machine="$(uname -m)"
+archive="vmnet-helper-$machine.tar.gz"
+curl -LOf "https://github.com/nirs/vmnet-helper/releases/latest/download/$archive"
+sudo tar xvf "$archive" -C / opt/vmnet-helper
+rm "$archive"
+```
+
+The command downloads the latest release from github and installs it to
+`/opt/vmnet-helper`.
+
+> [!IMPORTANT]
+> The vmnet-helper executable and the directory where it is installed
+> must be owned by root and may not be modifiable by unprivileged users.
+
+### Grant permission to run vmnet-helper
+
+The vment-helper process must run as root to create a vmnet interface.
+To allow users in the `staff` group to run the vmnet helper without a
+password, you can install the default sudoers rule:
+
+```shell
+sudo install -m 0640 /opt/vmnet-helper/share/doc/vmnet-helper/sudoers.d/vmnet-helper /etc/sudoers.d/
+```
+
+You can change the sudoers configuration to allow access to specific
+users or other groups.
+
+### Usage
+
+```shell
+minikube start --driver krunkit
+```
+
+## Issues
+
+### Other
+
+* [Full list of open 'krunkit' driver issues](https://github.com/kubernetes/minikube/labels/co%2Fkrunkit)
+
+## Troubleshooting
+
+### Run with logs
+
+Run `minikube start --driver krunkit --alsologtostderr -v=7` to debug crashes
+
+### Troubleshooting vmnet-helper
+
+Check for errors in vment-helper log:
+
+```shell
+$MINIKUBE_HOME/.minikube/machines/MACHINE-NAME/vmnet-helper.log
+```
+
+Check that the `vmnet-helper` process is running:
+
+```shell
+ps au | grep vmnet-helper | grep -v grep
+```
+
+If the helper is not running restart the minikube cluster.
+
+For help with vment-helper please use the
+[discussions](https://github.com/nirs/vmnet-helper/discussions).
--- a/site/content/en/docs/drivers/_index.md
+++ b/site/content/en/docs/drivers/_index.md
@@ -33,6 +33,7 @@ To do so, we use the [Docker Machine](https://github.com/docker/machine) library
 * [QEMU]({{<ref "qemu.md">}}) - VM
 * [Podman]({{<ref "podman.md">}}) - VM + Container (experimental)
 * [VFkit]({{<ref "vfkit.md">}}) - VM (preferred)
+* [Krunkit]({{<ref "krunkit.md">}}) - VM (experimental)
 * [SSH]({{<ref "ssh.md">}}) - remote ssh
 
 ## Windows


afbjorklund:
> So it seems that we need to move to newer kernel. Do you know why are stuck with 5.10?

I don't think it is stuck with anything, just that it was using the LTS versions.

nirs:
> It seems that it is only needed for this AI feature, though?

Yes, this is for AI on Apple silicon. I can try to get build an iso to play with it, and when we a working iso we can discuss how to proceed with the upgrade.

afbjorklund:
So you could still use vfkit for all other (non-AI) usage of Kubernetes.

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,561 @@
...
+		if err := os.Remove(pidfile); err != nil {
+			log.Debugf("failed to remove %q: %s", pidfile, err)
+		}
+		return state.Stopped, nil
+	}
+	return state.Running, nil
+}
+
+func (d *Driver) stopKrunkit() error {
+	if err := d.SetKrunkititState("Stop"); err != nil {
```
nirs, line 372: krunkit ignores this value, but it is documented, so we should keep it same as vfkit.

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,561 @@
...
+			return nil
+		}
+	}
+	return nil
+}
+
+func (d *Driver) killkrunkit() error {
+	if err := d.SetKrunkititState("HardStop"); err != nil {
+		// Typically fails with EOF due to https://github.com/crc-org/krunkit/issues/277.
+		log.Debugf("Failed to set krunkit state to 'HardStop': %s", err)
```
nirs, line 401: krunkit ignores the value and does not document this value, so we should not set the state here but just kill it.
https://github.com/containers/krunkit/blob/9421c0e1daca8af0eb84d6afc67bfb55b78214c4/src/status.rs#L101

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,561 @@
...
+	// https://github.com/containers/krunkit/issues/47
+	return "tcp://localhost:8081"
+}
+
+func (d *Driver) vmStateURI() string {
+	return d.restfulURI() + "/vm/state"
+}
+
+// Make a boot2docker VM disk image.
+func (d *Driver) generateDiskImage(size int) error {
```
nirs, line 460: Duplicate of vfkit version, which may duplicate of qemu?

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,561 @@
...
+
+func (d *Driver) GetDockerConfigDir() string {
+	return ""
+}
+
+func (d *Driver) Upgrade() error {
+	return fmt.Errorf("hosts without a driver cannot be upgraded")
+}
+
+func (d *Driver) getKrunkitState() (state.State, error) {
```
nirs, line 348: same as vfkit

nirs:
ok-to-build-iso

medyagh:
See the logs at:
https://storage.googleapis.com/minikube-builds/logs/PRNumber/7DigitSHA/iso_build.txt

for example for this PR 
https://storage.googleapis.com/minikube-builds/logs/20669/22ed44e/iso_build.txt

nirs:
> So you could still use vfkit for all other (non-AI) usage of Kubernetes.

Yes, the only interesting feature in krunkit is accelerated GPU, so if you don't need this I don't see a reason to use it.

> Or maybe even make it a single driver, if the syntax is close enough...

We can use a single driver that will use vfkit or krunkit, maybe if you use --enable-gpu-acceleration. But calling it vfkit and using krunkit will be confusing. Another issue is networking - for vfkit we have nat (default), or vmnet-helper (optional, for multi clusters or multi-node cluster). For krunkit we must use vmnet-helper so the --network option is not used. I think having 2 separate drivers is the best way for user experience and maintenance.

We have lot of duplicate code, but this can be solved by extracting shared pieces out of the drivers. Most of the duplicate code is also duplicated from other drivers (e.g. qemu).

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,561 @@
...
+	return d.ResolveStorePath(serialFileName)
+}
+
+func (d *Driver) isoPath() string {
+	return d.ResolveStorePath(isoFileName)
+}
+
+func (d *Driver) restfulURI() string {
+	// TODO: use unused port or a unix socket when supported.
+	// https://github.com/containers/krunkit/issues/47
```
nirs, line 451: We can use getAvailableTCPPortFromRange() from qemu driver.

nirs:
> /ok-to-test

This iso is broken for qemu, so the qemu tests will fail.

nirs:
ok-to-build-iso

pkg/drivers/krunkit/krunkit.go
nirs: krunkit should support unix:// socket soon via https://github.com/containers/krunkit/issues/47 so there is on point to waste time on this.

pkg/drivers/krunkit/krunkit.go
nirs: Fixed in current version

pkg/drivers/krunkit/krunkit.go
nirs: Does not work with minikube vm with kernel 6.6, we don't use restful api in current version.

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,561 @@
...
+	defer response.Body.Close()
+	err = json.NewDecoder(response.Body).Decode(&vmstate)
+	if err != nil {
+		return "", err
+	}
+	log.Debugf("get krunkit state: %+v", vmstate)
+	return vmstate.State, nil
+}
+
+func (d *Driver) SetKrunkititState(state string) error {
```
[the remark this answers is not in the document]
nirs: Removed in current version

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,561 @@
...
+	}
+	log.Debugf("DONE writing to %s and %s", rawFile, d.diskPath())
+	return nil
+}
+
+type VMState struct {
+	State string `json:"state"`
+}
+
+func (d *Driver) GetkrunkitState() (string, error) {
```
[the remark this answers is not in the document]
nirs: Remove, unused in current code.

pkg/drivers/krunkit/krunkit.go
nirs: Removed, unused in current code.

pkg/drivers/krunkit/krunkit.go
nirs: Using unix:// now.

nirs:
## Comparing start time: krunkit, vfkit, qemu

krunkit start a cluster with --no-kubernetes 1.24 times faster than vfkit and 1.91 faster than qemu. vfkit is 1.04 times faster when starting a full cluster.

### vm only

```console
% hyperfine -r 10 -C "krunkit/out/minikube delete" \
     "krunkit/out/minikube start --driver krunkit --no-kubernetes" \
     "krunkit/out/minikube start --driver vfkit --network vmnet-shared --no-kubernetes" \
     "krunkit/out/minikube start --driver qemu --no-kubernetes"
Benchmark 1: krunkit/out/minikube start --driver krunkit --no-kubernetes
  Time (mean ± σ):      8.355 s ±  1.112 s    [User: 0.283 s, System: 0.236 s]
  Range (min … max):    7.067 s … 10.269 s    10 runs
 
Benchmark 2: krunkit/out/minikube start --driver vfkit --network vmnet-shared --no-kubernetes
  Time (mean ± σ):     10.349 s ±  0.536 s    [User: 0.373 s, System: 0.243 s]
  Range (min … max):    9.511 s … 11.129 s    10 runs
 
Benchmark 3: krunkit/out/minikube start --driver qemu --no-kubernetes
  Time (mean ± σ):     15.977 s ±  0.433 s    [User: 0.507 s, System: 0.250 s]
  Range (min … max):   15.570 s … 16.787 s    10 runs
 
Summary
  krunkit/out/minikube start --driver krunkit --no-kubernetes ran
    1.24 ± 0.18 times faster than krunkit/out/minikube start --driver vfkit --network vmnet-shared --no-kubernetes
    1.91 ± 0.26 times faster than krunkit/out/minikube start --driver qemu --no-kubernetes
```

### kubernetes cluster

```console
% hyperfine -r 5 -C "krunkit/out/minikube delete" \
    "krunkit/out/minikube start --driver krunkit --container-runtime containerd" \
    "krunkit/out/minikube start --driver vfkit --container-runtime containerd --network vmnet-shared" \
    "krunkit/out/minikube start --driver qemu --container-runtime containerd"
Benchmark 1: krunkit/out/minikube start --driver krunkit --container-runtime containerd
  Time (mean ± σ):     20.958 s ±  1.098 s    [User: 0.978 s, System: 0.909 s]
  Range (min … max):   19.347 s … 21.887 s    5 runs
 
Benchmark 2: krunkit/out/minikube start --driver vfkit --container-runtime containerd --network vmnet-shared
  Time (mean ± σ):     20.108 s ±  0.774 s    [User: 1.057 s, System: 1.298 s]
  Range (min … max):   19.036 s … 21.078 s    5 runs
 
Benchmark 3: krunkit/out/minikube start --driver qemu --container-runtime containerd
  Time (mean ± σ):     27.047 s ±  1.633 s    [User: 0.932 s, System: 0.901 s]
  Range (min … max):   25.457 s … 28.802 s    5 runs
 
Summary
  krunkit/out/minikube start --driver vfkit --container-runtime containerd --network vmnet-shared ran
    1.04 ± 0.07 times faster than krunkit/out/minikube start --driver krunkit --container-runtime containerd
    1.35 ± 0.10 times faster than krunkit/out/minikube start --driver qemu --container-runtime containerd
```

nirs:
I rebase again on #20995 to consume #21027.

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,561 @@
...
+	}
+	return process.WritePidfile(d.pidfilePath(), cmd.Process.Pid)
+}
+
+func (d *Driver) openLogfile() (*os.File, error) {
+	logfile := d.ResolveStorePath(logFileName)
+	return os.OpenFile(logfile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600)
+}
+
+func (d *Driver) setupIP(mac string) error {
```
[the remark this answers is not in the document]
nirs: We can improve this later.

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+		if err := os.Remove(pidfile); err != nil {
+			log.Debugf("failed to remove %q: %s", pidfile, err)
+		}
+		return state.Stopped, nil
+	}
+	return state.Running, nil
+}
+
+func (d *Driver) stopKrunkit() error {
+	if err := d.SetKrunkitState("Stop"); err != nil {
```
nirs: This is not reliable enough. The guest can ignore the ACPI request, or it may not work in some cases.

When running the functional and integration tests I see many failures stop krunkit after many retries.

I think we should simplify this to terminate krunkit like we do with vmnet-heloper.
 
We can improve this later to try graceful shutdown dnd terminate after a timeout. This is also needed for vfkit and qemu, and ideally should be part of the driver API.

medyagh: sounds good to me, would you like to create an issue for this? we can comment TODO Iissue number)

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+	Boot2DockerURL string
+	DiskSize       int
+	CPU            int
+	Memory         int
+	ExtraDisks     int
+	MACAddress     string
+	VmnetHelper    vmnet.Helper
+}
+
+func NewDriver(hostName, storePath string) drivers.Driver {
```
medyagh, line 82: exported func, needs comment

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+		BaseDriver: &drivers.BaseDriver{
+			SSHUser:     defaultSSHUser,
+			MachineName: hostName,
+			StorePath:   storePath,
+		},
+		CommonDriver: &pkgdrivers.CommonDriver{},
+	}
+}
+
+func (d *Driver) PreCreateCheck() error {
```
medyagh, line 93: same here ones bellow, all exported funcs need comment, I am surprised the linter didnt yell at us

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+	}
+	return process.WritePidfile(d.pidfilePath(), cmd.Process.Pid)
+}
+
+func (d *Driver) openLogfile() (*os.File, error) {
+	logfile := d.ResolveStorePath(logFileName)
+	return os.OpenFile(logfile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600)
+}
+
+func (d *Driver) setupIP(mac string) error {
```
medyagh, line 261: add comment here too and if need improve please create issue and link the issue // TODO: issue number

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+		return nil
+	}
+	// Implement a retry loop because IP address isn't added to dhcp leases file immediately
+	for i := 0; i < 60; i++ {
+		log.Debugf("Attempt %d", i)
+		err = getIP()
+		if err == nil {
+			break
+		}
+		time.Sleep(2 * time.Second)
```
medyagh, line 277: curious how did we come up with 2 seconds, why not 1 ?

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+		return err
+	}
+
+	log.Infof("Waiting for VM to start (ssh -p %d docker@%s)...", d.SSHPort, d.IPAddress)
+
+	return WaitForTCPWithDelay(fmt.Sprintf("%s:%d", d.IPAddress, d.SSHPort), time.Second)
+}
+
+// startkrunkit starts the krunkit child process.
+func (d *Driver) startkrunkit(socketPath string) error {
```
medyagh: for easier reading the func name should be camle case (startKrunkit)

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+
+func (d *Driver) StopDocker() error {
+	return fmt.Errorf("hosts without a driver cannot stop docker")
+}
+
+func (d *Driver) GetDockerConfigDir() string {
+	return ""
+}
+
+func (d *Driver) Upgrade() error {
```
medyagh: I am not sure why there is this func at all, we dont seem to have it in other drivers, except qemu and vfkit... could be just un-used code?

pkg/drivers/krunkit/krunkit.go
medyagh: if this is an exact duplicate, lets add a TODO: issue to make them in a reusable code shared across 3 drirvers, we already have code in the Parent Driver package that is shared across all drivers

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+	return http.Client{
+		Transport: &http.Transport{
+			DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
+				return net.Dial("unix", path)
+			},
+		},
+	}
+}
+
+type VMState struct {
```
medyagh: does it need to be exported ? if yes comment

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+	State string `json:"state"`
+}
+
+func (d *Driver) SetKrunkitState(state string) error {
```
medyagh: does it need to be exported ? if yes comment

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,561 @@
...
+	}
+	_, err = http.Post(d.vmStateURI(), "application/json", bytes.NewReader(data))
+	if err != nil {
+		return err
+	}
+	log.Infof("Set krunkit state: %+v", vmstate)
+	return nil
+}
+
+func WaitForTCPWithDelay(addr string, duration time.Duration) error {
```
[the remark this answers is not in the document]
medyagh: yes can add TODO comment, can be moved to Parent Drivers Package

pkg/minikube/registry/drvs/krunkit/krunkit.go
```diff
@@ -0,0 +1,92 @@
...
+		VmnetHelper: vmnet.Helper{
+			MachineDir:  filepath.Join(storePath, "machines", machineName),
+			InterfaceID: u,
+			// Required Until https://github.com/containers/libkrun/issues/264 is fixed.
+			Offloading: true,
+		},
+	}, nil
+}
+
+func status() registry.State {
```
medyagh, line 88: in the status we can also check if vmnet-helper is installed and if Not we can set
the 
NeedsImprovement Bool in the 
```
type State struct {
	Installed        bool
	Healthy          bool
	Running          bool // it at least appears to be running
	NeedsImprovement bool // healthy but could be improved
	Error            error

	Reason  string // A reason ID, propagated to reason.Kind.ID
	Fix     string
	Doc     string
	Version string // 
```
that way we can inform user to install vmnet-helper as an improvement without yelling at them that somethign is wrong

medyagh (changes requested):

pkg/drivers/krunkit/krunkit.go
nirs: This was copied from vfkit, I guess this is missing in many places. I'll check how we document it in other places.

pkg/drivers/krunkit/krunkit.go
nirs: Again copied from vfkit, which was copied from qemu :-)

Maybe to limit cpu usage, reading and parsing /var/db/dhcpd_leases can be expensive. But this is pretty bad, causing unnecessary delays up to 2 seconds when waiting for ip.

But improving this is not in the scope of adding new driver, I'll open an issue for this.

pkg/drivers/krunkit/krunkit.go
nirs: No, this is internal. Will be removed since we want to terminate with a signal.

pkg/drivers/krunkit/krunkit.go
nirs: No, will be removed since we want to terminate with a signal.

pkg/minikube/registry/drvs/krunkit/krunkit.go
nirs: I'll check this. We check for vmnet-helper in configure(), but we can do the same check for status.

For krunkit we must use vmnet-helper so the status should not be healthy unless vment-helper is installed and we can run it.

pkg/drivers/krunkit/krunkit.go
nirs: Sure, I will simplify the code to use signals, and add an issue to support graceful shutdown.

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+func (d *Driver) isoPath() string {
+	return d.ResolveStorePath(isoFileName)
+}
+
+func (d *Driver) sockfilePath() string {
+	return d.ResolveStorePath(sockFileName)
+}
+
+func (d *Driver) restfulURI() string {
+	return fmt.Sprintf("unix://%s", d.sockfilePath())
```
nirs, line 461: Delete, not needed when using `--restful-url none://`

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+func (d *Driver) serialPath() string {
+	return d.ResolveStorePath(serialFileName)
+}
+
+func (d *Driver) isoPath() string {
+	return d.ResolveStorePath(isoFileName)
+}
+
+func (d *Driver) sockfilePath() string {
+	return d.ResolveStorePath(sockFileName)
```
nirs, line 457: Delete, not needed when using `--restful-url none://`

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+}
+
+func (d *Driver) restfulURI() string {
+	return fmt.Sprintf("unix://%s", d.sockfilePath())
+}
+
+func (d *Driver) vmStateURI() string {
+	return "http://_/vm/state"
+}
```
nirs, line 466: Delete, not needed when using `--restful-url none://`

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+		return nil
+	}
+	if err := os.Truncate(rawFile, int64(size)*int64(1024*1024)); err != nil {
+		return nil
+	}
+	log.Debugf("DONE writing to %s and %s", rawFile, d.diskPath())
+	return nil
+}
+
+func httpUnixClient(path string) http.Client {
```
nirs, line 523: Delete, not needed when using `--restful-url none://`

pkg/drivers/krunkit/krunkit.go
nirs: Delete, not needed when using `--restful-url none://`

pkg/drivers/krunkit/krunkit.go
nirs: Delete, not needed when using `--restful-url none://`

pkg/drivers/krunkit/krunkit.go
nirs: Maybe, @afbjorklund copied from qemu to vfkit, I copied from vfkit here.

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+
+	return WaitForTCPWithDelay(fmt.Sprintf("%s:%d", d.IPAddress, d.SSHPort), time.Second)
+}
+
+// startkrunkit starts the krunkit child process.
+func (d *Driver) startkrunkit(socketPath string) error {
+	var args = []string{
+		"--memory", fmt.Sprintf("%d", d.Memory),
+		"--cpus", fmt.Sprintf("%d", d.CPU),
+		"--restful-uri", d.restfulURI(),
```
nirs, line 219: Replace with:

```go
    "--restful-url", "none://",
```

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+	}
+	return fmt.Sprintf("tcp://%s:2376", ip), nil
+}
+
+func (d *Driver) GetIP() (string, error) {
+	return d.IPAddress, nil
+}
+
+// GetState returns driver state. Since krunkit driver may use 2 processes
+// (vmnet-helper, krunkit), this returns combined state of both processes.
```
nirs, line 148: krunkit always uses 2 processes

pkg/drivers/krunkit/krunkit.go
nirs: But commenting driver interface in all the implementations is not very helpful to users. We should comment only the interface, or if the driver implementation is special and require some docs.

Example for function that needs a comment: GetState()

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
...
+	isoFileName    = "boot2docker.iso"
+	pidFileName    = "krunkit.pid"
+	sockFileName   = "krunkit.sock"
+	logFileName    = "krunkit.log"
+	serialFileName = "serial.log"
+	logLevelInfo   = "3"
+	defaultSSHUser = "docker"
+)
+
+// Driver is the machine driver for krunkit (Virtualization.framework)
```
nirs: Krunkit does not use Virtualization.framework but libkrun. Add link to the project here.

pkg/drivers/krunkit/krunkit.go
medyagh: fair enough

pkg/drivers/krunkit/krunkit.go
medyagh: yeah that makes sense, the comment should be in the interface and only additional comment if there is a Need to say,  maybe thats why the linter is not yelling at us ! so I guess it is NOT needed to comment if there is comment in interface. no need to spam comments for sake of comments

pkg/drivers/krunkit/krunkit.go
```diff
@@ -0,0 +1,562 @@
+//go:build darwin
+
+/*
+Copyright 2024 The Kubernetes Authors All rights reserved.
```
nirs: The date is pointless but we can use 2025 now.

pkg/minikube/reason/reason.go
```diff
@@ -552,15 +552,11 @@ var (
 	NotFoundVmnetHelper = Kind{
 		ID:       "NOT_FOUND_VMNET_HELPER",
 		ExitCode: ExProgramNotFound,
-		Advice: translate.T(`vmnet-helper was not found on the system, resolve by:
+		Advice: translate.T(`vmnet-helper was not found on the system.
 
-		Option 1) Installing vmnet-helper:
+		Please install vmnet-helper using these instrunctions:
```
afbjorklund: "instructions"

pkg/minikube/reason/reason.go
```diff
@@ -552,15 +552,11 @@ var (
...
 
-		  https://github.com/nirs/vmnet-helper#installation
-
-		Option 2) Using the nat network:
-
-		  minikube start{{.profile}} --driver vfkit --network nat`),
+		https://github.com/nirs/vmnet-helper#installation`),
```
afbjorklund, line 559: You should probably fork this to the `minikube-machine` organization, or provide binaries in minikube itself

pkg/drivers/krunkit/krunkit.go
afbjorklund: The main reason was that qemu (and thus vfkit) were external drivers to libmachine, which didn't have helpers.

For qemu2 and minikube-machine, there are more opportunities to add it to the common driver SDK package

afbjorklund:
Looks good to me, I don't have anything further for vfkit or krunkit drivers so feel free to do whatever.

Thanks @nirs for picking it up, after the quick proof-of-concept of replacing parts in "qemu" with "vfkit"

* https://github.com/kubernetes/minikube/pull/19423

Eventually there should be virtio-gpu support in qemu, but until then this driver fills a niche on macOS

pkg/minikube/reason/reason.go
nirs: I don't think keeping a fork in minikube-machine is very useful, since minikube does not install vmnet-helper. This is a component that the user must install, just like socket_vmnet (used by qemu driver).

krunkit itself comes from Sergio Lopez personal GitHub:
https://github.com/slp/krunkit

So this driver depends on 2 other projects published by the authors.

It would be awesome if minikube can install the krunkit and vment-helper, but both projects are not trivial to install. krunkit is depending on libkrunt, which in turn depend son other libraries:

```console
% otool -L /opt/homebrew/bin/krunkit
/opt/homebrew/bin/krunkit:
        /opt/homebrew/opt/libkrun-efi/lib/libkrun-efi.1.dylib (compatibility version 1.0.0, current version 1.14.0)
        /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 3502.1.255)
        /usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
        /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
        /usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
```

```console
% otool -L /opt/homebrew/opt/libkrun-efi/lib/libkrun-efi.1.dylib
/opt/homebrew/opt/libkrun-efi/lib/libkrun-efi.1.dylib:
        /opt/homebrew/opt/libkrun-efi/lib/libkrun-efi.1.dylib (compatibility version 1.0.0, current version 1.14.0)
        /System/Library/Frameworks/Hypervisor.framework/Versions/A/Hypervisor (compatibility version 1.0.0, current version 210.6.2)
        /opt/homebrew/opt/libepoxy/lib/libepoxy.0.dylib (compatibility version 1.0.0, current version 1.0.0)
        /opt/homebrew/opt/virglrenderer/lib/libvirglrenderer.1.dylib (compatibility version 1.0.0, current version 1.0.0)
        /usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
```

```console
% otool -L /opt/homebrew/opt/virglrenderer/lib/libvirglrenderer.1.dylib
/opt/homebrew/opt/virglrenderer/lib/libvirglrenderer.1.dylib:
        /opt/homebrew/opt/virglrenderer/lib/libvirglrenderer.1.dylib (compatibility version 1.0.0, current version 1.0.0)
        /opt/homebrew/opt/molten-vk-krunkit/lib/libMoltenVK.dylib (compatibility version 1.0.0, current version 1.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
        /opt/homebrew/opt/libepoxy/lib/libepoxy.0.dylib (compatibility version 1.0.0, current version 1.0.0)
```

It is possible to package krunkit and its libraries as macOS bundle and provide an installer package, but nobody has packaged it yet. We can do this in minikube to provide a minikube owned package, but this requires additional work to update the package each time krunkit or its dependencies update.

vment-helper has no dependencies (except macOS vment framework), but it must be installed as root. Since minikube is is not installed as root, there is no safe way to install vment-helper as root. This is the same problem with socket_vmnet and socket_vmnet recommends installing from the GitHub binary release.

I want to make a package for vment-helper to allow easier installation at some point.

medyagh: deally if it can be installed using brew would be best, since Brew install things from source (I belive we install krunkit with brew), we dont have to "Trust" the author to publish the right Binary in the github assets, could we have it built by brew?

pkg/drivers/krunkit/krunkit.go
nirs: The issue was the kernel, not this code so we want to keep it.

Rename to vmState

pkg/drivers/krunkit/krunkit.go
nirs: The issue was the kernel, not this code so we want to keep it.

Rename to setKrunkitState()

pkg/drivers/krunkit/krunkit.go
nirs: Keeping it, the issue was in the kernel.

pkg/drivers/krunkit/krunkit.go
nirs: No change needed

pkg/drivers/krunkit/krunkit.go
nirs: No change needed

pkg/drivers/krunkit/krunkit.go
nirs: No change needed

pkg/drivers/krunkit/krunkit.go
nirs: No change needed

pkg/minikube/reason/reason.go
nirs: We will build and deliver binaries in https://github.com/minikube-machine/vmnet-helper

pkg/drivers/krunkit/krunkit.go
nirs: Added comment linking to #21090

pkg/drivers/krunkit/krunkit.go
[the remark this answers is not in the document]
nirs: Added TODO comments linking to https://github.com/kubernetes/minikube/issues/21091

pkg/drivers/krunkit/krunkit.go
nirs: The issue was the kernel, fixed in #21089.

Add TODO comment about the handling ignored termination request linking to https://github.com/kubernetes/minikube/issues/21092

pkg/drivers/krunkit/krunkit.go
nirs: Added comment like most drivers.

pkg/drivers/krunkit/krunkit.go
nirs: Add TODO comment linking to https://github.com/kubernetes/minikube/issues/21093

pkg/drivers/krunkit/krunkit.go
nirs: Renamed

pkg/drivers/krunkit/krunkit.go
nirs: Removed since it is unused and not part of the drivers.Driver interface.

pkg/drivers/krunkit/krunkit.go
nirs: Renamed

pkg/drivers/krunkit/krunkit.go
nirs: Renamed.

pkg/drivers/krunkit/krunkit.go
nirs: Updated

pkg/drivers/krunkit/krunkit.go
nirs: Removed mention of virtualization.framework

pkg/drivers/krunkit/krunkit.go
nirs: Updated to 2025

pkg/minikube/registry/drvs/krunkit/krunkit.go
nirs: Since vmnet-helper is required for krunkit, using NeedsImprovement is incorrect. We need to return hard error if vment-helper is not available.

nirs: Change to check:
- runtime.GOOS == "Darwin"
- runtime.GOARCH == "arm64"
- vment-helper validated

If any fail we return registry.State{Error: err, Fix: ..., Doc: ...}.

Removing network option validation for krunkit since we don't any any option. It was used for validating vmnent-helper but if we validate here is it not needed.

nirs: We can use NeedsImprovement for vfkit, since vmment-helpers enabled access between clusters and multi-nodes clusters.

pkg/minikube/reason/reason.go
nirs: Fixed

nirs:
@medyagh all comments addressed. Keeping as draft since it is rebased on #21089, required for vfkit and krunkit.

nirs:
Rebased on #21089 to consume the new iso.

pkg/minikube/driver/driver.go
```diff
@@ -178,6 +180,11 @@ func IsVFKit(name string) bool {
 	return name == VFKit
 }
 
+// IsKrunkit checks if the driver is vfkit
```
medyagh:
```suggestion
// IsKrunkit checks if the driver is krunkit
```

site/content/en/docs/drivers/krunkit.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Other
+
+* [Full list of open 'krunkit' driver issues](https://github.com/kubernetes/minikube/labels/co%2Fkrunkit)
+
+## Troubleshooting
+
+### Run with logs
+
+Run `minikube start --driver vfkit --alsologtostderr -v=7` to debug crashes
```
medyagh:
```suggestion
Run `minikube start --driver krunkit --alsologtostderr -v=7` to debug crashes
```

medyagh (changes requested):
just two small minor changes and then we can merge this

nirs:
> thank you @nirs just two small minor typos and then we can merge this

Fixed

nirs:
## Testing krunkit with unsupported platform

```console
$ out/minikube start --driver krunkit
😄  minikube v1.36.0 on Fedora 42 (kvm/amd64)

❌  Exiting due to DRV_UNSUPPORTED_OS: The driver 'krunkit' is not supported on linux/amd64
```

## Testing krunkit not installed

```console
% out/minikube start --driver krunkit
😄  minikube v1.36.0 on Darwin 15.5 (arm64)
✨  Using the krunkit (experimental) driver based on user configuration

🤷  Exiting due to PROVIDER_KRUNKIT_NOT_FOUND: The 'krunkit' provider was not found: exec: "krunkit": executable file not found in $PATH
💡  Suggestion: Run 'brew tap slp/krunkit && brew install krunkit'
📘  Documentation: https://minikube.sigs.k8s.io/docs/reference/drivers/krunkit/
```

## Testing vmnet-helper not installed

```console
% out/minikube start --driver krunkit
😄  minikube v1.36.0 on Darwin 15.5 (arm64)
✨  Using the krunkit (experimental) driver based on user configuration

🤷  Exiting due to PROVIDER_KRUNKIT_NOT_FOUND: The 'krunkit' provider was not found: stat /opt/vmnet-helper/bin/vmnet-helper: no such file or directory
💡  Suggestion: Install and configure vment-helper
📘  Documentation: https://minikube.sigs.k8s.io/docs/reference/drivers/krunkit/
```

## Testing vment-helper not configured

sudores rule not installed, password required during startup.

```console
% out/minikube start --driver krunkit
😄  minikube v1.36.0 on Darwin 15.5 (arm64)
💡  Unable to run vmnet-helper without a password
    To configure vment-helper to run without a password, please check the documentation:
    https://github.com/nirs/vmnet-helper/#granting-permission-to-run-vmnet-helper
Password:
✨  Using the krunkit (experimental) driver based on user configuration
🔥  Creating krunkit VM (CPUs=2, Memory=6144MB, Disk=20000MB) ...
🐳  Preparing Kubernetes v1.33.2 on Docker 28.3.2 ...
    ▪ Configuring RBAC rules ...
```

## Testing vment-helper configured

Using passwordless sudo.

```console
% sudo --reset-timestamp

% out/minikube start --driver krunkit                                                                 
😄  minikube v1.36.0 on Darwin 15.5 (arm64)
✨  Using the krunkit (experimental) driver based on user configuration
🔥  Creating krunkit VM (CPUs=2, Memory=6144MB, Disk=20000MB) ...
🐳  Preparing Kubernetes v1.33.2 on Docker 28.3.2 ...
    ▪ Configuring RBAC rules ...
```

PR #20830, opened 2025-05-24
docs: Fix instructions for building iso

nirs:
Testing shows that we need changes changes:
- x86_64 cpu
- Ubuntu 22.04
- docker is required even if building without docker
- python2 instead of python
- genisoimage (for mkisofs)
- Installing Go manually (Ubuntu 22.04 have only Go 1.18)
- Target should be minikube-iso-aarch64 or minikube-iso-x86_64. Using
  arm64 and amd64 fails.

I also cleaned up a little bit the formatting to make it easier to
maintain (one package per line).

Tested building:
- minikube-iso-aarch64
- minikube-iso-x86_64

I tests only the arm64 iso image.

### DIFF
--- a/site/content/en/docs/contrib/building/iso.md
+++ b/site/content/en/docs/contrib/building/iso.md
@@ -10,6 +10,9 @@ The minikube ISO is booted by each hypervisor to provide a stable minimal Linux 
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


medyagh (approved):
thank you for updating this doc, been a long time since touched

PR #20892, opened 2025-06-05
iso: Test flaky test with iso build

nirs:
Add a comment to grub.cfg forcing an iso rebuild. This will show the flaky tests failing without any change in the code.


nirs:
ok-to-build-iso

nirs:
iso build: https://storage.googleapis.com/minikube-builds/logs/20892/00ce030/iso_build.txt

nirs:
ok-to-test

PR #20894, opened 2025-06-05
qemu: Add serial log

nirs:
Log serial console to ~/.minikube/machines/NAME/serial.log to make it easy to debug boot issues.


nirs:
ok-to-test

nirs:
I checked the test failures:
- vulkano fails in 9 tests (the only failure) - known issue recently

Other failures are in container based drivers, not using the iso.

@medyagh @ComradeProgrammer needs your /lgtm.

PR #20900, opened 2025-06-07
krunkit: Use unix socket for --restul-uri

nirs:
This removes the limit of using a single machine, removing the need to
manage tcp ports. We use:
    
    --restful-uri unix://$MINIKUBE_HOKE/machines/NAME/krunkit.sock
    
The socket is created and deleted by krunkit.
    
NOTES:
 - Depends on https://github.com/containers/krunkit/pull/51.
   To test this change build and install krunkit from this PR.

Status:
- The API call works with krunkit using unix socket
- krunkit does not terminate since graceful shutdown does not work with minikube VM with kernel 6.6.

Based on #20826 for testing.


medyagh:
@nirs would you like ISO to be built for this PR ? (for you to test it ?)

nirs:
> @nirs would you like ISO to be built for this PR ? (for you to test it ?)

Not needed, this adds one commit to #20826 depending on krunkit PR, and it does not work.

nirs:
Squashed into #20826

PR #20921, opened 2025-06-10
Fix minikube load on windows (#20529)

### LINKED ISSUE #20529: minikube image load my-local-image:latest  ❌  Exiting due to GUEST_IMAGE_LOAD
### What Happened?

Exiting due to GUEST_IMAGE_LOAD: Failed to load image: save to dir: caching images: caching image "C:\\Users\\MyUserAcc\\.minikube\\cache\\images\\amd64\\my-local-image_latest": getting destination path: parsing docker archive dst ref: replace a Win drive letter to a volume name: exec: "wmic": executable file not found in %PATH%

Have tried to run `minikube delete --purge`, then reinstalling minikube, also reinstalling Docker desktop.

Running minikube using the Docker driver. Docker is using WSL.

### Attach the log file

[logs.txt](https://github.com/user-attachments/files/19260095/logs.txt)

### Operating System

Windows

### Driver

Docker

james-world:
fixes #20529

`getWindowsVolumeNameCmd` uses the deprecated wmic.exe, this breaks `minikube image load` on newer installs of windows. Whilst wmic.exe can be added, it will be removed from Windows 11 soon.

See [here](https://learn.microsoft.com/en-us/windows/win32/wmisdk/wmic) for details.

This fix replaces the use of `wmic.exe` with a call to the Powershell equivalent.


pkg/minikube/localpath/localpath.go
```diff
@@ -190,25 +191,20 @@ func replaceWinDriveLetterToVolumeName(s string) (string, error) {
 }
 
 func getWindowsVolumeNameCmd(d string) (string, error) {
-	cmd := exec.Command("wmic", "volume", "where", "DriveLetter = '"+d+":'", "get", "DeviceID")
+	psCommand := `Get-CimInstance -ClassName Win32_Volume -Filter "DriveLetter = '` + d + `:'" | Select-Object -ExpandProperty DeviceID`
```
medyagh: @james-world thank you for this PR do you mind putting the output of minikube Before/After this PR ? and also how about adding a Fail Over for users with Older Windows, so if this Method didnt work, minikube would use the older method?

and would this work in "git bash in windows"  or only in powershell ?

also I think this is an opportunity to add unit test here

james-world: Hi @medyagh,

Addressing your points in turn:

### 1. Output before and after

Before (on any windows system where wmic.exe is not in path, whether run from Git Bash / MingW64 / Cmd / Powershell):

```
.\minikube.exe image load employee-api:latest

❌  Exiting due to GUEST_IMAGE_LOAD: Failed to load image: save to dir: caching images: caching image "C:\\Users\\james.world\\.minikube\\cache\\images\\amd64\\employee-api_latest": getting destination path: parsing docker archive dst ref: replace a Win drive letter to a volume name: exec: "wmic": executable file not found in %PATH%

╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                               │
│    😿  If the above advice does not help, please let us know:                                                 │
│    👉  https://github.com/kubernetes/minikube/issues/new/choose                                               │
│                                                                                                               │
│    Please run `minikube logs --file=logs.txt` and attach logs.txt to the GitHub issue.                        │
│    Please also attach the following file to the GitHub issue:                                                 │
│    - C:\Users\JAMES~1.WOR\AppData\Local\Temp\minikube_image_1b5c8c33fb94831b9d273a970d695f2957a83a6b_0.log    │
│                                                                                                               │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```
Without this fix, this will be the default experience on latest Windows builds now because wmic.exe is no longer included by default; e.g. developers on my team with new laptops have run into this, existing developers with us more than a month ago are OK.

Output after fix applied (again, on any windows system where wmic.exe is not in path, whether run from Git Bash / MingW64 / Cmd / Powershell):

_There is no output!_ , as with the existing implementation the `minikube image load` command actually doesn't output anything when successful. This is a helper method executed by upstream functionality for image load - and that doesn't output anything on success bar the return code.

In terms of this specific function (`getWindowsVolumeNameCmd`), the output is of course identical to what it would have been using `wmic.exe`. Something like this:

```
\\?\Volume{ca87fc49-8366-4377-8c24-b7be4be5627b}\
```

### 2. Adding fallback behaviour

As for adding the failover - on the surface it sounds like a good idea. I can do it if you insist, however consider that as far as I know the Get-CimInstance was first added in Windows Powershell 3.0 and has been included in windows by default since Windows 8 and Windows Server 2012. Do we really need this fallback, do you think?

### 3. Would this work in "git bash in windows" or only in powershell ?

It works regardless of the windows shell from which minikube is executed - the go code is launching the Powershell process itself; it should always be in path. I've certainly tested from Git Bash, MingW64, Powershell and Cmd and all are fine.

### 4. I think this is an opportunity to add unit test here.

What would a unit test do beyond what's already in localpath_test.go? This feels more like integration testing territory to me? Because in this case, the function is dependent directly on Windows+Powershell in calling out to Powershell's Get-CimInstance command.

Honestly, I have very little go experience. It's been novel for me to build minikube on Windows; the docs aren't very informative for Windows platform builds. Because this is a drop in replacement for existing functionality, I had hoped the existing test coverage would be fine. If not, do you have a good example you can point in the codebase to for an existing integration test that is Windows specific that would serve as a good template for how to go about adding this idiomatically?

medyagh: thank you for the response, that sounds convincing, I think it wouldn't hurt to fall back to older vers way but I could be convinced either way.
 @bobsira do you mind commenting on this, what do you think ?

james-world: On further consideration @medyagh, it's pretty trivial to add the fallback, so I have.

james-world: So I made it so if both the PowerShell and the WMIC methods both fail, both errors are reported, to aid troubleshooting:

```
❌  Exiting due to GUEST_IMAGE_LOAD: Failed to load image: save to dir: caching images: caching image "C:\\Users\\james.world\\.minikube\\cache\\images\\amd64\\employee-api_latest": getting destination path: parsing docker archive dst ref: replace a Win drive letter to a volume name: PowerShell failed (exit status 1), WMIC also failed: exec: "wmic": executable file not found in %PATH%

╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                               │
│    😿  If the above advice does not help, please let us know:                                                 │
│    👉  https://github.com/kubernetes/minikube/issues/new/choose                                               │
│                                                                                                               │
│    Please run `minikube logs --file=logs.txt` and attach logs.txt to the GitHub issue.                        │
│    Please also attach the following file to the GitHub issue:                                                 │
│    - C:\Users\JAMES~1.WOR\AppData\Local\Temp\minikube_image_1b5c8c33fb94831b9d273a970d695f2957a83a6b_0.log    │
│                                                                                                               │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

gabrielgbs97: I believe the older method should not be used as a fallback. WMIC **has been deprecated** since the release of Windows 10 RTM, which was 10 years ago, and it is not included in the default installation of Windows 11 23H2. Therefore, it would not have been implemented using wmic.exe around 2016 - 2019 (?)

gabrielgbs97:
If powershell.exe is not available in PATH (it should be...) it can be discovered via `SystemRoot` env variable:
`'%SystemRoot%\system32\WindowsPowerShell\v1.0\powershell.exe`

pkg/minikube/localpath/localpath.go
james-world: I'm comfortable either way; it does no harm and may help someone that has disabled PowerShell access in some way - thus avoiding a breaking change. If an approver can let me know which way to jump please?

bobsira: WMIC is deprecated so we can get rid of the older method fallback bit.  Powershell is the way to go since it's recommended command-line utility on Windows

bobsira: https://github.com/minikube-machine/machine/blob/main/drivers/hyperv/powershell.go 

This currently exist on how we discover Powershell and execute PS commands, you can borrow from this. Your change looks alright to me.

james-world:
OK, sorry for the hiatus - been busy at work. @medyagh  I've removed the fallback as discussed and added code to robustly discover Powershell.exe, using the cited approaches.

medyagh:
thank you @james-world, let me ask @bobsira for last approve and then we can merge it
@bobsira can you plz test this on windows and see if it fixes the issue?

pkg/minikube/localpath/localpath.go
```diff
@@ -189,27 +191,55 @@ func replaceWinDriveLetterToVolumeName(s string) (string, error) {
...
+			return loc, nil
 		}
 	}
 
-	if vname == "" {
-		return "", errors.New("failed to get a volume GUID")
+	return "", fmt.Errorf("PowerShell not found in PATH or common locations")
+}
+
+func getWindowsVolumeNameCmd(d string) (string, error) {
```
medyagh, line 223: add comment here pleae

bobsira: The Get-CimInstance command works fine on my Windows machine!

bobsira (commented):
Tested this on Windows machine, works fine! Approved @medyagh 

Had a successful run with the changes,
`C:\dev\minikube\out>minikube.exe image load nginx:latest --alsologtostderr                                                                                                                                                                           I0714 02:30:21.918732   24816 out.go:345] Setting OutFile to fd 96 ...                                                                                                                                                                               I0714 02:30:21.949424   24816 out.go:392] TERM=,COLORTERM=, which probably does not support color                                                                                                                                                    I0714 02:30:21.949424   24816 out.go:358] Setting ErrFile to fd 100...                                                                                                                                                                               I0714 02:30:21.950091   24816 out.go:392] TERM=,COLORTERM=, which probably does not support color                                                                                                                                                    I0714 02:30:21.969639   24816 config.go:182] Loaded profile config "minikube": Driver=hyperv, ContainerRuntime=docker,
KubernetesVersion=v1.33.2                                                                                                     

=== some of the logs removed to have this fit in the comment
                                                                                                   
I0714 02:30:36.438734   24816 cache_images.go:123] Successfully loaded all cached images                                                                                                                                                             I0714 02:30:36.438734   24816 cache_images.go:92] duration metric: took 5.9631078s to LoadCachedImages                                                                                                                                               I0714 02:30:36.439366   24816 cache_images.go:262] succeeded pushing to: minikube  `

medyagh:
@james-world plz see last comments and then we can merge this

james-world:
@medyagh Added the requested comment.

pkg/minikube/localpath/localpath.go
james-world: Done.

medyagh:
thank you @james-world

rata:
The failures seem unrelated to this PR (failed to pull an image)

james-world:
@medyagh  I've started to see this kind of error at work; it's docker.io rate-limiting. Definitely nothing to do with this PR.

The rate limiting on anonymous logins is by IP - see [here](https://docs.docker.com/docker-hub/usage/). That means if you are using any shared IP on the build agent pool you can hit limits faster than you might think.

We've addressed it by using pull-through caches (we use ACR for this: https://learn.microsoft.com/en-us/azure/container-registry/artifact-cache-overview)

Guess minikube would need to cut a deal with docker.io, or do something similar.

james-world:
@medyagh Just following up on this - not sure what else I can do? I'm pretty sure the build failure here is an unrelated transient issue with throttled image pulls from docker registry. Anything I can do to help move it along?

medyagh:
> @medyagh I've started to see this kind of error at work; it's docker.io rate-limiting. Definitely nothing to do with this PR.
> 
> The rate limiting on anonymous logins is by IP - see [here](https://docs.docker.com/docker-hub/usage/). That means if you are using any shared IP on the build agent pool you can hit limits faster than you might think.
> 
> We've addressed it by using pull-through caches (we use ACR for this: https://learn.microsoft.com/en-us/azure/container-registry/artifact-cache-overview)
> 
> Guess minikube would need to cut a deal with docker.io, or do something similar.

thank you for sharing that I will check that out. we have two different rate limit issue,
on the host we login, but inside minikube, minikube itself pull images for addons and things like that that causes rate limit. one option would be login to docker inside minikube as well

I will checkout the caching you have linked

medyagh:
thank you very much @james-world for your patience and presistance and making minikube better for our windows users, I look forward to see more contributions from you, maybe you could join one of our Office Hours or Triage Parties
https://minikube.sigs.k8s.io/community/

PR #20923, opened 2025-06-10
WIP: Update kernel to 6.6

nirs:
Update kernel to longterm kernel 6.6.92.

Improve the makefile to make it easier to work on the iso.

Upgrading the kernel to 6.6 does not work since some configuration is missing, so I started with default kernel configuration.

The next steps:
- rebase on #20991 
- removing more configs that we don't need
- add configs that we need from the previous config
- enable virtio-gpu
- update kernel to 6.6.95 (2025-06-27)
- cleanup commits
- document the changes required on top of the default kernel configuration to make the next update easier

Status:
- arm64:
  - qemu, vfkit, krunkit can boot with --no-kubernetes
  - booting with Kubernetes fails in the preflight step since we are missing some configs
  - virtio-gpu not enabled yet
- x86_64:
  - not tested yet


deploy/iso/minikube-iso/board/minikube/aarch64/linux_aarch64_defconfig
```diff
@@ -34,49 +34,6 @@ CONFIG_PROFILING=y
 CONFIG_KEXEC=y
 CONFIG_KEXEC_FILE=y
 CONFIG_CRASH_DUMP=y
-CONFIG_ARCH_ACTIONS=y
```
medyagh, line 49: could this possibly affect emultation ? like running a X86 arch on Arm ?

Makefile
```diff
@@ -325,6 +316,18 @@ minikube-iso-%: deploy/iso/minikube-iso/board/minikube/%/rootfs-overlay/usr/bin/
                 mv $(BUILD_DIR)/buildroot/output-x86_64/images/rootfs.iso9660 $(BUILD_DIR)/minikube-amd64.iso; \
         fi;
 
+.PHONY: buildroot
```
medyagh, line 314: consider adding a a comment starting  with  "##"
 that will be shown in "make help"
 for example
```
 generate-docs: extract out/minikube ## Automatically generate commands documentation.
```

deploy/iso/minikube-iso/board/minikube/aarch64/linux_aarch64_defconfig
nirs: I don't think it is related. This was done via `make linux-menuconfig-aarch64`, entering platform support and disable all platforms, since we run in vm. If we have issues we can enable it later.

nirs:
@afbjorklund this is very early but you may want to review and suggest how we should do this in a better way.

nirs:
Replaced by #20995

Issue #20953, opened 2025-06-18
--addons accepts comma separated list, docs don't mention it.


ipochi:
### What Happened?

While I was trying to get ingress and ingress-dns working, I thought the below command when starting up the cluster would install the ingress and ingress dns addons

```
--addons ingress ingress-dns
```

It only installed ingress addon.

`--addons` flag does accept comma separated list, but the docs don't mention it. 

Output at the time of cluster creation

```
$ minikube start --driver vfkit --network vmnet-shared --memory 3072 --container-runtime containerd --addons ingress ingress-dns

😄  minikube v1.36.0 on Darwin 15.5 (arm64)
✨  Using the vfkit driver based on user configuration
👍  Starting "minikube" primary control-plane node in "minikube" cluster
🔥  Creating vfkit VM (CPUs=2, Memory=3072MB, Disk=20000MB) ...
❗  Image was not built for the current minikube version. To resolve this you can delete and recreate your minikube cluster using the latest images. Expected minikube version: v1.35.0 -> Actual minikube version: v1.36.0
📦  Preparing Kubernetes v1.33.1 on containerd 1.7.23 ...
    ▪ Generating certificates and keys ...
    ▪ Booting up control plane ...
    ▪ Configuring RBAC rules ...
🔗  Configuring bridge CNI (Container Networking Interface) ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
    ▪ Using image registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.5.3
    ▪ Using image registry.k8s.io/ingress-nginx/controller:v1.12.2
    ▪ Using image registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.5.3
🔎  Verifying ingress addon...
🌟  Enabled addons: storage-provisioner, default-storageclass, ingress
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default
```

### Attach the log file

[log.txt](https://github.com/user-attachments/files/20804494/log.txt)

### Operating System

macOS (Default)

### Driver

N/A

PR #9957, opened 2020-12-14
start: Support comma-delimited --addons

Aut0R3V:
Fixes #9319

Using slicing technique in the place of an Array for line 141.

PR #20954, opened 2025-06-18
docs: clarify --addons flag description

ipochi:
Support for comma separated list was added in this PR https://github.com/kubernetes/minikube/pull/9957, but docs aren't very clear about it.

Fixes: #20953 

follow up: IMO  `--addons ingress ingress-dns` is just as valid as`--addons ingress,ingress-dns`


medyagh (approved):
thank you

PR #20991, opened 2025-06-28
iso: Improve build and clean up configuration

nirs:
Previously making the iso-menuconfig-{aarch64,x86_64} or linux-menuconfig-{aarch64,x86_64} targets you had to build the entire iso, since the out/buildroot tree was created by the minikube-iso-{aarch64,x86_64} target.

This change fix the Makefile so you can run the iso-menuconfig-{aarch64,x86_64} or linux-menuconfig-{aarch64,x86_64} directly after checking the source. This makes it much easier to work on the iso configuration.

Previously the linux-menuconfig-{aarch64,x86_64} targets were not using the buildroot facilities, copying files manually using hard coded paths and kernel version. Fixing the target to use the buildroot way removed the unneeded KERNEL_VERSION variable from the Makefile, that made updating kernel version hard.

Running the iso-menuconfig-{aarch64,x86_64} and linux-menuconfig-{aarch64,x86_64} without any change revealed that we had manual changes in the configuration that were not created by config. I run all the targets to update the configuration to the output of kconfig, so future changes do not include unrelated changes.

For aarch64 the image is much larger (410 MiB vs 362 MiB). Turns out that removing unnecessary support for various Arm boards reduces the image to 392 MiB.

Status
- [x] Building iso locally
- [x] Testing with vfkit and qemu on arm64 using ramen rook testing environment https://github.com/RamenDR/ramen/blob/main/test/envs/rook.yaml 
- [x] Testing with kvm2 driver on amd64 using ramen regional dr environment https://github.com/RamenDR/ramen/blob/main/test/envs/regional-dr.yaml


nirs:
ok-to-build-iso

Makefile
```diff
@@ -302,10 +301,25 @@ out/e2e-windows-amd64.exe: out/e2e-windows-amd64
...
+	$(MAKE) -C $(BUILD_DIR)/buildroot $(BUILDROOT_OPTIONS) O=$(BUILD_DIR)/buildroot/output-$* host-python3
+	$(MAKE) -C $(BUILD_DIR)/buildroot $(BUILDROOT_OPTIONS) O=$(BUILD_DIR)/buildroot/output-$*
+	# x86_64 ISO is still BIOS rather than EFI because of AppArmor issues for KVM, and Gen 2 issues for Hyper-V
+	if [ "$*" = "aarch64" ]; then \
+                mv $(BUILD_DIR)/buildroot/output-aarch64/images/boot.iso $(BUILD_DIR)/minikube-arm64.iso; \
+        else \
+                mv $(BUILD_DIR)/buildroot/output-x86_64/images/rootfs.iso9660 $(BUILD_DIR)/minikube-amd64.iso; \
+        fi;
+
+.PHONY: iso-preapre-%
```
afbjorklund: typo

nirs: What is the typo? I'm trying to define both iso-preapre-aarch64 and iso-prepare-x86_64 as PHONY.

Maybe use this?

```
.PHONY iso-preapre-aarch64 iso-prepare-x86_64
```

nirs:
@afbjorklund can you add ok-to-test? the .PHONY change does not affect the build and fixed now.

Makefile
nirs: Fixed in current version.

nirs:
I added another commit removing unnecessary support for various Arm boards. This should not be needed since we run on in a vm.

nirs:
ok-to-build-iso

nirs:
I checked KVM errors:
- [KVM_Linux_containerd](https://storage.googleapis.com/minikube-builds/logs/20991/40237/KVM_Linux_containerd.html)
  - volcano (known issue)
  - toomanyrequests: You have reached your unauthenticated pull rate limit (93 instances)
- [KVM_Linux_crio](https://storage.googleapis.com/minikube-builds/logs/20991/40237/KVM_Linux_crio.html)
  - toomanyrequests: You have reached your unauthenticated pull rate limit (98 instances)

medyagh:
@nirs thank you for making it faster to iterate on the ISO build, do you mind sharing an example iteration of ISO build using this new target ? do you manually call make prepare-... ? or do you build the iso the way it was before and let it do the work underneath ?

if this new make target could help other developers, we could add it to the minikube website

nirs:
> @nirs thank you for making it faster to iterate on the ISO build, do you mind sharing an example iteration of ISO build using this new target ? 

Without this change this will fail:

```
git clone https://github.com/kubernetes/minikube.git
cd minikube
make iso-menuconfig-{arch}
```

Because the out/buildroot is created by `make minikube-iso-{arch}`, which an take 2 hours to build.

Same for `make linux-menuconfig-{arch}`.

With this change we can do this to update kernel version:

```
git clone https://github.com/kubernetes/minikube.git
cd minikube
```

Update kernel version and headers version via kconfig and exit:

```
make iso-menuconfig-{arch}
```

Run to update linux config for the new kernel:

```
make linux-menuconfig-{arch}
```

> do you manually call make prepare-... ? 

No

> or do you build the iso the way it was before and let it do the work underneath ?

Exactly, it is it internal target, this is also why it is not documented.

> if this new make target could help other developers, we could add it to the minikube website

It is better not to complicate the docs, this target is relevant only if you wan to modify the makefile.

I think that the current docs need update for multiple architectures. Maybe they worked when minikube had only single iso.
- https://minikube.sigs.k8s.io/docs/contrib/building/iso/#modifying-buildroot-components
- https://minikube.sigs.k8s.io/docs/contrib/building/iso/#adding-kernel-modules

I will post another PR to describing updating kernel flow.

PR #20994, opened 2025-06-29
iso: Disable all platforms for aarch64

nirs:
We run on qemu virt machine or apple virtualization so we don't need support for all kinds of embeded Arm boards. This reduces the arm64 iso size from 410 MiB to 392 MiB.

Status:
- [x] vfkit with vmnet-shared network: using ramen rook testing environment https://github.com/RamenDR/ramen/blob/main/test/envs/rook.yaml
- [x] qemu with socket_vmnet network: starting and deleting cluster

Based on #20991 for testing


nirs:
Squashed into #20991

PR #21017, opened 2025-07-02
build: Fix docker run when selinux is enabled

nirs:
When mouting volumes we need to use :Z suffix to allow access to the mounted directory. This fixes docker run when running on Fedora an other distros using selinux by default.

NOTE: This change was committed as part of #20746, but it was reverted by mistake by ##20852.


nirs:
@prezha can you confirm that this was reverted by mistake?

medyagh:
thats interesting...that was merged last month but the xcgo error happend today...I wonder why...

nirs:
@medyagh @ComradeProgrammer can you add /ok-to-test? build should pass now after #21027 was merged.

PR #20833, opened 2025-05-25
vfkit: Use EFI booloader

nirs:
With the fixed iso, we can simplify the driver using the EFI bootloader
option[1] instead of the legacy and deprecated --kernel, --kernel-cmdline,
and --initrd options[2].

Example run:

    % minikube start -p vfkit --driver vfkit --container-runtime containerd --network vmnet-shared
    😄  [vfkit] minikube v1.36.0 on Darwin 15.5 (arm64)
    ✨  Using the vfkit driver based on user configuration
    👍  Starting "vfkit" primary control-plane node in "vfkit" cluster
    🔥  Creating vfkit VM (CPUs=2, Memory=6000MB, Disk=20000MB) ... 
    📦  Preparing Kubernetes v1.33.1 on containerd 1.7.23 ... 
        ▪ Generating certificates and keys ... 
        ▪ Booting up control plane ... 
        ▪ Configuring RBAC rules ... 
    🔗  Configuring bridge CNI (Container Networking Interface) ... 
    🔎  Verifying Kubernetes components...
        ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
    🌟  Enabled addons: default-storageclass, storage-provisioner
    🏄  Done! kubectl is now configured to use "vfkit" cluster and "default" namespace by default

Comparing direct kernel boot and --bootloader efi shows that it is little bit faster and boot time is more consistent.

```shell
% hyperfine -r 10 -C "minikube delete" \
    "vfkit-efi/out/minikube start --driver vfkit --network vmnet-shared --container-runtime containerd --no-kubernetes" \
    "vfkit-direct/out/minikube start --driver vfkit --network vmnet-shared --container-runtime containerd --no-kubernetes"
Benchmark 1: vfkit-efi/out/minikube start --driver vfkit --network vmnet-shared --container-runtime containerd --no-kubernetes
  Time (mean ± σ):     10.205 s ±  0.656 s    [User: 0.381 s, System: 0.266 s]
  Range (min … max):    9.106 s … 11.254 s    10 runs
 
Benchmark 2: vfkit-direct/out/minikube start --driver vfkit --network vmnet-shared --container-runtime containerd --no-kubernetes
  Time (mean ± σ):     10.933 s ±  1.616 s    [User: 0.402 s, System: 0.406 s]
  Range (min … max):    9.155 s … 14.168 s    10 runs
 
Summary
[...]

PR #21021, opened 2025-07-03
site: vfkit: Fix requirements version to MacOS 12

illume:
- vfkit itself mentions it requires MacOS 12 (not 13)
- We did some basic tests of the vfkit minikube driver on a MacOS 12 laptop and it works fine

### DIFF
--- a/site/content/en/docs/drivers/vfkit.md
+++ b/site/content/en/docs/drivers/vfkit.md
@@ -13,7 +13,7 @@ container deployment.
 
 ## Requirements
 
-- Requires macOS 13 or later.
+- Requires macOS 12 or later.
 - Requires minikube version 1.36.0 or later.
 
 ## Networking


afbjorklund:
Some features like UEFI do require macOS 13, but in the end I choose to use the more compatible version (I think the framework was introduced with macOS 11?)

So it is ok to change this to macOS 12 for now, but it might be bumped again in the future if the vfkit driver changes. Now using macOS 15, so harder to test

afbjorklund:
Actually we _are_ using uefi now, so macOS 13 _is_ required

* https://github.com/kubernetes/minikube/pull/20833

illume:
How can I test something that will trigger this issue with EFI? Because we didn’t run into an issue yet with macOS 12.

Ok, the PR is not in a release yet (Last release was May 22, and the PR is from May 25).

So I'll close this PR.

afbjorklund:
The documentation is not branched, so there are no minikube docs available for "latest" but only for "main"

But there is also https://endoflife.date/macos

Issue #21012, opened 2025-07-02
Hyperkit build issues with xcgo


nirs:
### What Happened?

The hyperkit driver requires xcgo container image which is unmaintained for long time:
https://github.com/neilotoole/xcgo

The last release is from 2020:
https://github.com/neilotoole/xcgo/releases

and the last commit is 2 years ago:
https://github.com/neilotoole/xcgo/commit/044ab7c3f254b6418528b07734218b87240d72e7

We forked the project for building hyperkit for 1.35:
https://github.com/nirs/xcgo

But the image has many vulnerabilities again: 

- 4 Critical-level vulnerabilities.
- 15 High-level vulnerabilities.
- 190 Medium-level vulnerabilities.
- 322 Low-level vulnerabilities.
- 34 Negligible-level vulnerabilities.
- 59 Unknown-level vulnerabilities.

https://quay.io/repository/nirsof/xcgo/manifest/sha256:115720ae3f5f5b02326d083834610f809c0eeceacf994ddee15ca3ded3590b9c?tab=vulnerabilities

Since vfkit works on both arm64 and amd64, we don't need hyperkit and we can remove it in minikube 1.37.0.

We also don't have capacity to maintain the hyperkit driver. If someone want to maintain this driver they are invited to own it and fix the xcgo image.

### Operating System

macOS (Default)

### Driver

HyperKit

divysinghvi:
New contributor here — just checking if this is still being worked on or if any help is needed. Happy to assist.

nirs:
Welcome @divysinghvi!

This issue waits for discussion with the maintainers, so there is no work to do at this point.

If you are looking for stuff to work on, you can search for issue with "good first issue" label like this:
https://github.com/kubernetes/minikube/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22

You can also join [#minikube on Kubernetes Slack](https://kubernetes.slack.com/messages/minikube) and ask there about this.

divysinghvi:
Thank you! I’ll look into good first issues. If the discussion moves forward and there's any work to be done, I’d be happy to contribute.

medyagh:
the thing is the people on X86 Macos dont have too many options, and hystorically been a very stable driver.
I think what we can do is, depreciate the driver and warn the user to use other options such as Qemu 

simmilar to virtualbox we have some warninings for that

nirs:
> the thing is the people on X86 Macos dont have too many options

[...]

PR #21024, opened 2025-07-03
Deprecate HyperKit driver with warning

divysinghvi:
This PR deprecates the HyperKit driver by displaying a user-facing warning when it is selected via `--driver=hyperkit`.

- Warns users that HyperKit is deprecated and may be removed in future releases.
- Suggests using alternative drivers such as `docker`, `qemu`, or `vfkit`.
- Includes documentation links for migration.

Partially addresses #21012

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -406,7 +406,7 @@ func virtualBoxMacOS13PlusWarning(driverName string) {
 	if !driver.IsVirtualBox(driverName) || !detect.MacOS13Plus() {
 		return
 	}
-	suggestedDriver := driver.HyperKit
+	suggestedDriver := driver.VFKit
 	if runtime.GOARCH == "arm64" {
 		suggestedDriver = driver.QEMU
 	}
@@ -424,7 +424,7 @@ func hyperkitDeprecationWarning(driverName string) {
 		return
 	}
 	out.WarningT(`The 'hyperkit' driver is deprecated and will be removed in a future release.
-    Please consider using an alternative driver such as 'docker', 'qemu', or 'vfkit'.`)
+    Please consider using an alternative driver such as vfkit, qemu, or docker`)
 }
 
 func validateBuiltImageVersion(r command.Runner, driverName string) {
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -309,6 +309,7 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 	}
 
 	virtualBoxMacOS13PlusWarning(driverName)
+	hyperkitDeprecationWarning(driverName)
 	validateFlags(cmd, driverName)
 	validateUser(driverName)
 	if driverName == oci.Docker {
@@ -415,6 +416,15 @@ func virtualBoxMacOS13PlusWarning(driverName string) {
 
     For more details on the issue see: https://github.com/kubernetes/minikube/issues/15274
 `, out.V{"driver": suggestedDriver})
+}
+
+// hyperkitDeprecationWarning prints a deprecation warning for the hyperkit driver
+func hyperkitDeprecationWarning(driverName string) {
+	if !driver.IsHyperKit(driverName) {
+		return
+	}
+	out.WarningT(`The 'hyperkit' driver is deprecated and will be removed in a future release.
+    Please consider using an alternative driver such as 'docker', 'qemu', or 'vfkit'.`)
 }
 
 func validateBuiltImageVersion(r command.Runner, driverName string) {
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -406,16 +406,12 @@ func virtualBoxMacOS13PlusWarning(driverName string) {
 	if !driver.IsVirtualBox(driverName) || !detect.MacOS13Plus() {
 		return
 	}
-	suggestedDriver := driver.VFKit
-	if runtime.GOARCH == "arm64" {
-		suggestedDriver = driver.QEMU
-	}
-	out.WarningT(`Due to changes in macOS 13+ minikube doesn't currently support VirtualBox. You can use alternative drivers such as docker or {{.driver}}.
+	out.WarningT(`Due to changes in macOS 13+ minikube doesn't currently support VirtualBox. You can use alternative drivers such as 'vfkit', 'qemu', or 'docker'.
+    https://minikube.sigs.k8s.io/docs/drivers/vfkit/
+    https://minikube.sigs.k8s.io/docs/drivers/qemu/
     https://minikube.sigs.k8s.io/docs/drivers/docker/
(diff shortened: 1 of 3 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -417,6 +418,15 @@ func virtualBoxMacOS13PlusWarning(driverName string) {
 `, out.V{"driver": suggestedDriver})
 }
 
+// hyperkitDeprecationWarning prints a deprecation warning for the hyperkit driver
+func hyperkitDeprecationWarning(driverName string) {
+	if !driver.IsHyperKit(driverName) {
+		return
+	}
+	out.WarningT(`The 'hyperkit' driver is deprecated and will be removed in a future release.
+    Please consider using an alternative driver such as 'docker', 'qemu', or 'vfkit'.`)
```
nirs: The preferred driver for macOS is vfkit, and then qemu. docker requires docker desktop which is not open source, so we should not recommend it.

So let's use "such as vfkit, qemu, or docker"

cmd/minikube/cmd/start.go
```diff
@@ -309,6 +309,7 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 	}
 
 	virtualBoxMacOS13PlusWarning(driverName)
+	hyperkitDeprecationWarning(driverName)
```
nirs, line 312: In this function we recommend hyperkit as replacement for virtulabox - please update it to recommend the same drivers we recommend here.

divysinghvi: Okay i will do that and recommend vfkit, qemu instead of hyperkit

nirs:
@divysinghvi I think this change can fix #21012. We don't need to do any other change now. The driver build works, and we want to remove it later, but for now marking it as depracated and recommending other drivers is all we need to do.

We also need to update hyperkit docs:
site/content/en/docs/drivers/hyperkit.md

It should have as similar warning that the driver will be removed in future version and recommend other drivers, linking to the other drivers docs.

cmd/minikube/cmd/start.go
```diff
@@ -406,7 +406,7 @@ func virtualBoxMacOS13PlusWarning(driverName string) {
 	if !driver.IsVirtualBox(driverName) || !detect.MacOS13Plus() {
 		return
 	}
-	suggestedDriver := driver.HyperKit
+	suggestedDriver := driver.VFKit
 	if runtime.GOARCH == "arm64" {
 		suggestedDriver = driver.QEMU
```
nirs: This is wrong, vfkit is the preferred driver on both arm64 and amd64.

cmd/minikube/cmd/start.go
```diff
@@ -424,7 +424,7 @@ func hyperkitDeprecationWarning(driverName string) {
 		return
 	}
 	out.WarningT(`The 'hyperkit' driver is deprecated and will be removed in a future release.
-    Please consider using an alternative driver such as 'docker', 'qemu', or 'vfkit'.`)
+    Please consider using an alternative driver such as vfkit, qemu, or docker`)
```
nirs: I would keep the quoting, unless most other warnings do not quote driver names.

divysinghvi:
@nirs Thanks a lot for your help! 🙌
I've applied the suggested changes.

Regarding the HyperKit documentation — I’d be happy to work on it and add a similar warning to that too

divysinghvi:
@nirs While working on this, I noticed that the VirtualBox driver documentation page also does not mention that the driver is deprecated: https://minikube.sigs.k8s.io/docs/drivers/virtualbox/.

It might be helpful to update that page as well.

cmd/minikube/cmd/start.go
```diff
@@ -406,7 +406,7 @@ func virtualBoxMacOS13PlusWarning(driverName string) {
 	if !driver.IsVirtualBox(driverName) || !detect.MacOS13Plus() {
 		return
 	}
-	suggestedDriver := driver.HyperKit
+	suggestedDriver := driver.VFKit
 	if runtime.GOARCH == "arm64" {
 		suggestedDriver = driver.QEMU
 	}
```
nirs: We need to update the warning bellow - it says:

    You can use alternative drivers such as docker or {{.driver}}.
    https://minikube.sigs.k8s.io/docs/drivers/{{.driver}}/

It should say:

    You can use alternative drivers such as vfkit, qemu, or docker.

cmd/minikube/cmd/start.go
```diff
@@ -406,7 +406,7 @@ func virtualBoxMacOS13PlusWarning(driverName string) {
 	if !driver.IsVirtualBox(driverName) || !detect.MacOS13Plus() {
 		return
 	}
-	suggestedDriver := driver.HyperKit
+	suggestedDriver := driver.VFKit
```
nirs: We can also remove this variable and the {{.driver}} template, since we don't need to select a driver based on arch, and we have several recommendations.

divysinghvi: okay so i will change it to list it like you suggested
`You can use alternative drivers such as vfkit, qemu, or docker.
https://minikube.sigs.k8s.io/docs/drivers/docker/`
i will also add this to HyperKit warning too

nirs:
> The warning looks good, but we also need to mark the driver.Priority to Deprecated in pkg/minikube/registry/drvs/hyperkit/hyperkit.go.

@divysinghvi did you miss this comment?

divysinghvi:
@nirs You're right — I missed that, apologies! I'll update driver.Priority to Deprecated in pkg/minikube/registry/drvs/hyperkit/hyperkit.go and push the change shortly. Thanks for pointing it out!

cmd/minikube/cmd/start.go
afbjorklund: Kubernetes project does not have a problem with recommending using Docker in various places, though.

cmd/minikube/cmd/start.go
```diff
@@ -405,16 +406,25 @@ func virtualBoxMacOS13PlusWarning(driverName string) {
 	if !driver.IsVirtualBox(driverName) || !detect.MacOS13Plus() {
 		return
 	}
-	suggestedDriver := driver.HyperKit
-	if runtime.GOARCH == "arm64" {
-		suggestedDriver = driver.QEMU
-	}
-	out.WarningT(`Due to changes in macOS 13+ minikube doesn't currently support VirtualBox. You can use alternative drivers such as docker or {{.driver}}.
+	out.WarningT(`Due to changes in macOS 13+ minikube doesn't currently support VirtualBox. You can use alternative drivers such as 'vfkit', 'qemu', or 'docker'.
```
medyagh, line 409: is there a reason to do the virtuablox related changes in this PR ?

nirs: Yes, otherwise virtaulbox message recommends hyperkit an replacement.

medyagh:
@divysinghvi thank you for this contribution, look forward to see more :)

divysinghvi:
Thanks @medyagh @nirs for helping me with this PR looking forward to making a lot more contributions to minikube in future :)

PR #21027, opened 2025-07-03
ci: login to docker during build process

medyagh:

### DIFF
--- a/hack/jenkins/minikube_cross_build_and_upload.sh
+++ b/hack/jenkins/minikube_cross_build_and_upload.sh
@@ -41,7 +41,7 @@ declare -rx DEB_VER="$(make deb_version)"
 docker kill $(docker ps -q) || true
 docker rm $(docker ps -aq) || true
 docker system prune -a --volumes -f
-docker login -u minikubebot -p $DOCKERHUB_READONLY_TOKEN
+echo $DOCKERHUB_READONLY_TOKEN | docker login -u minikubebot --password-stdin
 make -j 16 \
   all \
   minikube-darwin-arm64 \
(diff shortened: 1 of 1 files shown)


hack/jenkins/minikube_cross_build_and_upload.sh
```diff
@@ -41,6 +41,7 @@ declare -rx DEB_VER="$(make deb_version)"
 docker kill $(docker ps -q) || true
 docker rm $(docker ps -aq) || true
 docker system prune -a --volumes -f
+docker login -u ${DOCKERHUB_USER} -p ${DOCKERHUB_PASS}
```
nirs: This exposes the password in the environment, and in ps output. Currently the environment is logged in the build log, accessible to the world.

Maybe store it in a user readable file?

medyagh: we use the jenkins credential plugin that masks the passwords, so even if we Echo the envs, it will show as "****"
it is used in the kicbae auto build as well, so I think that would be a bigger thing to change everywhere else

nirs: It seems safe enough for now, but I would do this:
- Use a secret file
- Use a Personal Access Token instead of password
  - If docker supports it, use personal access token with read only access

medyagh: I changed the password to Readonly Accesstoken

nirs: Can we make this more clear that this does not use a read/write password?


```
docker login -u ${DOCKERHUB_USER} -p ${DOCKERHUB_READONLY_TOKEN}
```

This will help the Jenkins admin to do the right thing.

nirs (commented):
Looks good

nirs (commented):
Looks good

PR #21054, opened 2025-07-11
fix: Remove podman binary detection from podman-env command

elasticdotventures:
Addresses feedback from reviewers by removing podman binary detection and clarifying the Docker client connectivity approach.

## Key Changes

- **Removed podman binary detection**: Eliminates `isPodmanAvailable()` function and `which podman` checks per reviewer feedback
- **Clarified service description**: Now explicitly refers to "Podman Docker-compatible service" for precision
- **Removed version specificity**: Changed from "podman 4.9.2" requirement to "recent Podman version with Docker API compatibility" 
- **Focused on Docker client only**: Documentation emphasizes only Docker client required, not podman binary

## Addresses Review Feedback

- ✅ **@afbjorklund**: Removed podman detection, focuses purely on Docker socket connectivity
- ✅ **Version specificity**: Removed specific version requirements as requested
- ✅ **Cleaner approach**: No podman binary dependency, only Docker client against Docker-compatible API

### Before
```bash
# Command would check for podman binary availability
eval $(minikube podman-env)  # Could fail if podman binary not found
```

### After  
```bash
# No podman binary check - only connects Docker client to Podman's Docker-compatible service
eval $(minikube podman-env)
export DOCKER_HOST="tcp://192.168.49.2:2376" 
docker images   # Works with Docker client against Podman service
```

## Foundation for Issue #9229

This implementation provides a clean foundation for the future SSH socket approach described in issue #9229, but keeps the current changes minimal and focused. The SSH socket implementation (#9229) would be a separate, more substantial project that can build on this foundation.

## Testing

- Updated tests to remove podman detection expectations
- All existing Docker client functionality preserved  
- Documentation updated to reflect new approach

This approach is more aligned with issue #9229's vision but represents a focused change that can be merged independently before tackling the larger SSH socket project.

### DIFF
--- a/site/content/en/docs/commands/podman-env.md
+++ b/site/content/en/docs/commands/podman-env.md
@@ -4,6 +4,14 @@ description: >
   Configure environment to use minikube's Podman service
 ---
 
+## Requirements
+
+- **Podman version 4.9.2 or newer is required.**
+- Support for Podman v3 and varlink-based communication has been removed. The `podman-env` command now configures your environment to use the Podman REST API socket, as required by Podman v4+.
+
+{{% pageinfo color="warning" %}}
+**Note:** If you are using an older version of Podman, please upgrade to at least v4.9.2 to use `minikube podman-env`. Legacy varlink-based workflows are no longer supported.
+{{% /pageinfo %}}
 
 ## minikube podman-env
 
@@ -30,7 +38,7 @@ minikube podman-env [flags]
       --add_dir_header                   If true, adds the file directory to the header of the log messages
       --alsologtostderr                  log to standard error as well as files (no effect when -logtostderr=true)
   -b, --bootstrapper string              The name of the cluster bootstrapper that will set up the Kubernetes cluster. (default "kubeadm")
-  -h, --help                             
+  -h, --help
       --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
       --log_dir string                   If non-empty, write log files in this directory (no effect when -logtostderr=true)
       --log_file string                  If non-empty, use this log file (no effect when -logtostderr=true)
--- a/site/content/en/docs/commands/podman-env.md
+++ b/site/content/en/docs/commands/podman-env.md
@@ -7,23 +7,40 @@ description: >
 ## Requirements
 
 - **Podman version 4.9.2 or newer is required.**
-- Support for Podman v3 and varlink-based communication has been removed. The `podman-env` command now configures your environment to use the Podman REST API socket, as required by Podman v4+.
+- **Docker client is required** - `podman-env` uses Docker's client to communicate with Podman's Docker-compatible API.
+- The `podman-env` command configures Docker client environment variables to connect to minikube's Podman service via its Docker-compatible API.
 
-{{% pageinfo color="warning" %}}
-**Note:** If you are using an older version of Podman, please upgrade to at least v4.9.2 to use `minikube podman-env`. Legacy varlink-based workflows are no longer supported.
+{{% pageinfo color="info" %}}
+**Note:** This command sets up standard Docker environment variables (`DOCKER_HOST`, `DOCKER_TLS_VERIFY`, `DOCKER_CERT_PATH`) to connect to Podman's Docker-compatible socket. Use the regular `docker` command-line tool to interact with minikube's Podman service.
 {{% /pageinfo %}}
 
 ## minikube podman-env
 
-Configure environment to use minikube's Podman service
+Configure environment to use minikube's Podman service via Docker API compatibility
 
 ### Synopsis
 
-Sets up podman env variables; similar to '$(podman-machine env)'.
+Sets up Docker client env variables to use minikube's Podman Docker-compatible service.
 
 ```shell
 minikube podman-env [flags]
 ```
+
+### Usage
+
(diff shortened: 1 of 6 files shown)


site/content/en/docs/commands/podman-env.md
```diff
@@ -4,6 +4,14 @@ description: >
   Configure environment to use minikube's Podman service
 ---
 
+## Requirements
+
+- **Podman version 4.9.2 or newer is required.**
+- Support for Podman v3 and varlink-based communication has been removed. The `podman-env` command now configures your environment to use the Podman REST API socket, as required by Podman v4+.
```
afbjorklund: I don't think that varlink is relevant here, as it was removed already in Podman v2:

https://podman.io/blogs/2020/06/29/podman-v2-announce.html

site/content/en/docs/commands/podman-env.md
```diff
@@ -4,6 +4,14 @@ description: >
...
+
+{{% pageinfo color="warning" %}}
+**Note:** If you are using an older version of Podman, please upgrade to at least v4.9.2 to use `minikube podman-env`. Legacy varlink-based workflows are no longer supported.
```
afbjorklund: The version 4.9.2 is quite specific, it might as well say 4.x (which is not supported anyway) and later (5.x)

https://endoflife.date/podman

nirs: Can we use latest podman (5.5) instead of a random old version?

medyagh (commented):
can you plz post before/after this PR in the description

site/content/en/docs/commands/podman-env.md
afbjorklund: I don't think 4.9.3 is random*, but upgrading the client is not an issue.
Some people have problems with _downgrading_, but that's different.

\* it is the "last" common v4 version, just like 1.9.3 was the last v1 version.
We could upgrade to v5, but it is not available in every distribution yet.

afbjorklund: There are no features in podman 5 (or even in podman 4, but anyway)
that are required by cri-o, as far as I know? Main issue is client compat.

For the podman **driver** things are different, there we can require 4.9
This was just for docker-env/podman-env, and what is installed on the VM

afbjorklund (commented):
I don't think we should look for podman, or suggest to people to use the `podman --remote` client. We should only give a path to a docker socket, for a docker client.

Podman Desktop does **not** use podman, and Podman Compose does **not** default to using podman-compose. Instead, they use the Docker API and docker-compose.

Using the podman-remote client, is what is causing these problems with using cri-o.

`Error: unable to connect to Podman socket: server API version is too old. Client "4.0.0" server "3.4.4"`

It is better to only use `docker` as a legacy client,  and `minikube image` for the rest.

That is why it is called "compatibility socket", while the libpod client/connection keeps breaking the backwards API...

elasticdotventures:
## Update: Added Docker Runtime Support

I've added an enhancement that extends podman-env compatibility to support both **crio** and **docker** container runtimes

### Changes in Latest Commit

- Extended runtime compatibility check to allow both `constants.CRIO` and `constants.Docker`
- Updated error message to reflect support for both runtimes
- Maintains the same Docker API compatibility approach for both runtime types

### New Supported Configurations

1. **Podman driver + crio runtime** (original support)
   ```bash
   minikube start --driver=podman --container-runtime=crio
   eval $(minikube podman-env)
   ```

2. **Podman driver + docker runtime** (new support)  
   ```bash
   minikube start --driver=podman --container-runtime=docker
   eval $(minikube podman-env)
   ```

Both configurations use the same Docker API compatibility approach that eliminates the SSH-based connectivity issues and API version mismatches.

### Testing

Verified that both runtime configurations work correctly:
- Docker client connects successfully via podman-env
- `docker version` reports minikube's Docker daemon version (28.3.3)
- `docker ps` lists containers running in the cluster
- No API version compatibility errors

elasticdotventures:
@afbjorklund I just wanted to say thanks for really great advice & feedback. 
I took it all in and I worked up this solution which I'm now running on my local system successfully!   You deserve all the credit.

I also submitted a separate issue in CRIO (my attempt to isolate the reviews)
https://github.com/kubernetes/minikube/pull/21252

afbjorklund:
I don't think that it should use the legacy tcp socket for Podman, if anything the ssh socket should be used for Docker...

* https://github.com/kubernetes/minikube/issues/9229

As far as I know there is no support for it upstream, and if there are issues with ssh it would be better to use local unix*.

\* probably tunneled over ssh, but anyway (like it is being done for Podman Desktop)

----

Future versions of minikube should not even **start** the docker daemon or the podman service, by default.

Instead they should be socket-activated on demand, and only containerd or crio needs to be running for CRI.

But it is better to fix the functionality in `minikube image`.

Kubernetes does not need to use Docker at all, any more.

elasticdotventures:
@afbjorklund Thanks for the feedback. I've updated the PR to remove podman binary detection and version specificity as requested.

Issue #9229 is separate work that shouldn't be muddled with this change. Hopefully this cleaner foundation makes that future SSH socket implementation a bit smaller.

elasticdotventures:
😖 still blocked (merge this will fix #21252) .  unsure how to proceed @afbjorklund @medyagh 🙏

PR #21064, opened 2025-07-14
Add a default cluster dns name config

wt:
Add a default cluster DNS name config

It is very handy to be able to store the default cluster DNS name so
that I can fire up and teardown clusters easily without losing that
setting. You can think of this as a default value for the
`start --dns-domain <val>` flag, but only when starting a new cluster.

In my particular case, I would like to set my default cluster dna domain
to a subdomain under a domain that I control. E.g.
`cluster.wt.user.dev.example.com`. I want to do this to make it easier
to manage TLS certs for my dev cluster.

For context, the `start --dns-domain <val>` flag works like this:
  If you fire up a new cluster with that flag, you will get a new cluster
  with the domain name base. If you start an existing cluster, the
  config will be updated to the new domain name base and then started.

This config will work a little differently. The change implements a
config that will only affect newly started cluster. Here are some
examples to show that difference:

Newly started cluster example:
```
$ minikube config set DefaultDNSDomain cluster2.local
$ minikube start
...
```

Stop and starting the cluster with `--dns-domain` flag:
```
$ minikube stop
...
$ minikub start --dns-domain cluster3.local
...
```

Stopping and starting a new cluster with a non-default domain name and
without the `--dns-domain` flag:
```
$ minikube stop
...
$ minikub start
...
```

The reason for this behavior is that I am not configuring the name of a
cluster. That is a cluster configuration option. I am setting the
default for that.

If I manually overrode the name of a cluster when started previously, I
don't want a default option to override my cluster configuration.

Before:
  User cannot cannot change the default DNS domain name for a cluster
  to something other than cluster.local.

  On cluster creation/start:
    If a user runs `minikube start` without a `--dnsDomain` flag, the
    cluster will have the dns domain "cluster.local". A cluster setting
    is persisted.

    If a user runs 'minikube start --dnsDomain cluster.example.com`, the
    cluster will hae the dns domain "cluster.example.com". A cluster
    setting is persisted.

  On cluster restart:
    If a user run `minikube start`, the cluster's existing dns domain
    (pulled from the persisted cluster config) dns domain will be used.

    If a user runs `minikube start --dnsDomain cluster.example.com`, the
    cluster's persisted dns name will be updated to the new name and the
    cluster will be started with the new domain ("cluster.example.com").

  Example command sequence:
  ```
  $ minikube start --dnsDomain=cluster.example.com
  # cluster now has dns domain "cluster.example.com"
  $ minikube delete
  $ minikube start
  # cluster now has dns domain "cluster.local"
  ```

After:
  The behavior of the minikube start flag `--dns-domain` does not
  change at all. When the new configuration value is not set, all of the
  behaviors from the "Before" section above will not change.

  User can set the default cluster domain name to whatever they want
  via the config option DefaultDNSDomain. E.g.:
  `minikube config set DefaultDNSDomain cluster.wt.users.dev.example.com`

  Since the case with no configuration is the same as before, here is a
  description of what happens only with the new configuration set like
  above.

  On cluster creation/start:
    If a user runs `minikube start` without a `--dnsDomain` flag, the
    cluster will have the dns domain "cluster.wt.users.dev.example.com".
    A cluster setting is persisted. This is the only thing that changes
    with the configuration option set.

    If a user runs 'minikube start --dnsDomain cluster.example.com`, the
    cluster will hae the dns domain "cluster.example.com". A cluster
    setting is persisted.

  On cluster restart:
    If a user run `minikube start`, the cluster's existing dns domain
    (pulled from the persisted cluster config not from the new config)
    dns domain will be used.

    If a user runs `minikube start --dnsDomain cluster.example.com`, the
    cluster's persisted dns name will be updated to the new name and the
    cluster will be started with the new domain ("cluster.example.com").

  Notice that the only case that changes with the new setting defined is
  the case of starting a new cluster. Setting the config will change the
  behavior of the "Before -> Example command sequence" section as
  described:

  Example command sequence:
  ```
  $ minikube config set DefaultDNSDomain cluster.wt.users.dev.example.com
  $ minikube start --dnsDomain=cluster.example.com
  # cluster now has dns domain "cluster.example.com"
  $ minikube delete
  $ minikube start
  # cluster now has dns domain "cluster.wt.users.dev.example.com"
  ```

  The important piece here is that I can now change the default
  dnsDomain so that I don't have to specify it explicitly. This is
  useful when I am starting and deleting a single cluster where I want
  the top-level domain default to be a subdomain of a domain I control.

### DIFF
--- a/cmd/minikube/cmd/config/config.go
+++ b/cmd/minikube/cmd/config/config.go
@@ -172,6 +172,10 @@ var settings = []Setting{
 		name: config.MaxAuditEntries,
 		set:  SetInt,
 	},
+	{
+		name: config.DefaultDNSDomain,
+		set:  SetString,
+	},
 }
 
 // ConfigCmd represents the config command
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -225,7 +225,7 @@ func initKubernetesFlags() {
 		Valid components are: kubelet, kubeadm, apiserver, controller-manager, etcd, proxy, scheduler
 		Valid kubeadm parameters: `+fmt.Sprintf("%s, %s", strings.Join(bsutil.KubeadmExtraArgsAllowed[bsutil.KubeadmCmdParam], ", "), strings.Join(bsutil.KubeadmExtraArgsAllowed[bsutil.KubeadmConfigParam], ",")))
 	startCmd.Flags().String(featureGates, "", "A set of key=value pairs that describe feature gates for alpha/experimental features.")
-	startCmd.Flags().String(dnsDomain, constants.ClusterDNSDomain, "The cluster dns domain name used in the Kubernetes cluster")
+	startCmd.Flags().String(dnsDomain, constants.DefaultDNSDomain, "The cluster dns domain name used in the Kubernetes cluster")
 	startCmd.Flags().Int(apiServerPort, constants.APIServerPort, "The apiserver listening port")
 	startCmd.Flags().String(apiServerName, constants.APIServerName, "The authoritative apiserver hostname for apiserver certificates and connectivity. This can be used if you want to make the apiserver available from outside the machine")
 	startCmd.Flags().StringSliceVar(&apiServerNames, "apiserver-names", nil, "A set of apiserver names which are used in the generated certificate for kubernetes.  This can be used if you want to make the apiserver available from outside the machine")
@@ -586,6 +586,9 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 		out.WarningT("--network flag is only valid with the docker/podman, qemu, kvm, and vfkit drivers, it will be ignored")
 	}
 
+	clusterDNSDomain := viper.GetString(config.DefaultDNSDomain)
+	updateStringFromFlag(cmd, &clusterDNSDomain, dnsDomain)
+
 	validateHANodeCount(cmd)
 
 	checkNumaCount(k8sVersion)
@@ -662,7 +665,7 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 			APIServerName:          viper.GetString(apiServerName),
 			APIServerNames:         apiServerNames,
 			APIServerIPs:           apiServerIPs,
-			DNSDomain:              viper.GetString(dnsDomain),
+			DNSDomain:              clusterDNSDomain,
 			FeatureGates:           viper.GetString(featureGates),
 			ContainerRuntime:       rtime,
 			CRISocket:              viper.GetString(criSocket),
--- a/pkg/minikube/config/config.go
+++ b/pkg/minikube/config/config.go
@@ -59,6 +59,9 @@ const (
 	EmbedCerts = "EmbedCerts"
 	// MaxAuditEntries is the maximum number of audit entries to retain
 	MaxAuditEntries = "MaxAuditEntries"
+	// DefaultDnsDomain is the key for the default dns domain name when creating
+	// a cluster
+	DefaultDNSDomain = "default-dns-domain"
 )
 
 var (
--- a/pkg/minikube/constants/constants.go
+++ b/pkg/minikube/constants/constants.go
@@ -79,7 +79,7 @@ const (
 	// APIServerName is the default API server name
 	APIServerName = "minikubeCA"
 	// ClusterDNSDomain is the default DNS domain
-	ClusterDNSDomain = "cluster.local"
+	DefaultDNSDomain = "cluster.local"
 	// DefaultServiceCIDR is The CIDR to be used for service cluster IPs
 	DefaultServiceCIDR = "10.96.0.0/12"
 	// HostAlias is a DNS alias to the container/VM host IP
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -331,6 +331,7 @@ func setupViper() {
 	viper.SetDefault(config.WantVirtualBoxDriverWarning, true)
 	viper.SetDefault(config.MaxAuditEntries, 1000)
 	viper.SetDefault(config.SkipAuditFlag, false)
+	viper.SetDefault(config.DefaultDNSDomain, constants.DefaultDNSDomain)
 }
 
 func addToPath(dir string) {
--- a/pkg/minikube/bootstrapper/certs_test.go
+++ b/pkg/minikube/bootstrapper/certs_test.go
@@ -37,7 +37,7 @@ func TestSetupCerts(t *testing.T) {
 		CertExpiration: constants.DefaultCertExpiration,
 		KubernetesConfig: config.KubernetesConfig{
 			APIServerName: constants.APIServerName,
-			DNSDomain:     constants.ClusterDNSDomain,
+			DNSDomain:     constants.DefaultDNSDomain,
 			ServiceCIDR:   constants.DefaultServiceCIDR,
 		},
 	}
--- a/site/content/en/docs/commands/config.md
+++ b/site/content/en/docs/commands/config.md
@@ -41,7 +41,7 @@ Configurable fields:
  * native-ssh
  * rootless
  * MaxAuditEntries
-
+ * default-dns-domain - the default cluster top-level domain name (default "cluster.local")
 ```shell
 minikube config SUBCOMMAND [flags]
 ```
@@ -49,8 +49,7 @@ minikube config SUBCOMMAND [flags]
 ### Options inherited from parent commands
 
 ```
-      --add_dir_header                   If true, adds the file directory to the header of the log messages
-      --alsologtostderr                  log to standard error as well as files (no effect when -logtostderr=true)
+      --add_dir_header                   If true, adds the file directory to the header of the lo      --alsologtostderr                  log to standard error as well as files (no effect when -logtostderr=true)
   -b, --bootstrapper string              The name of the cluster bootstrapper that will set up the Kubernetes cluster. (default "kubeadm")
   -h, --help                             
       --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
@@ -69,6 +68,17 @@ minikube config SUBCOMMAND [flags]
   -v, --v Level                          number for the log level verbosity
       --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
 ```
+
+### What the config fields mean
+
+#### default-dns-domain - the default cluster top-level domain name
+
+The default is "cluster.local". This configuration field will only affect a newly started cluster.
+When restarting a cluster with the `start` subcommand, this configuration field will not change a domain
+name for that cluster.
+
+You can override this setting by using the `--dns-domain` flagfor the `start` subcommand. The `--dns-domain`
+can also change the domain name when restarting a cluster.
 
 ## minikube config defaults
 
@@ -306,4 +316,3 @@ minikube config view [flags]
   -v, --v Level                          number for the log level verbosity
       --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
 ```
-
--- a/test/integration/domain_test.go
+++ b/test/integration/domain_test.go
@@ -1,0 +1,178 @@
+//go:build integration && linux
+
+/*
+Copyright 2019 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package integration
+
+import (
+	"context"
+	"os/exec"
+	"testing"
+)
+
+// Test to make sure the default domain works as intended
+func TestDefaultDomainName(t *testing.T) {
+	// MaybeParallel(t)
+
+	profile := UniqueProfileName("domain-default")
+	ctx, cancel := context.WithTimeout(context.Background(), Minutes(10))
+	defer CleanupWithLogs(t, profile, cancel)
+
+	startArgs := append([]string{Target(), "start", "-p", profile, "--wait=apiserver,system_pods,default_sa"}, StartArgs()...)
+	rr, err := Run(t, exec.CommandContext(ctx, "/usr/bin/env", startArgs...))
+	if err != nil {
+		t.Errorf("%s failed: %v", rr.Command(), err)
+	}
+
+	rr, err = Run(t, exec.CommandContext(ctx, KubectlBinary(), "--context", profile, "run", "busybox", "--rm", "--restart=Never", "--image=gcr.io/k8s-minikube/busybox", "-it", "--", "/bin/sh", "-c", "awk '$1 == \"search\" {print $2; if ($2 !~ /\\.cluster\\.local$/) { print $2;exit 1}}' /etc/resolv.conf"))
+	if err != nil {
+		t.Errorf("%s failed: %v", rr.Command(), err)
+		t.Errorf("Output: %s", rr.Output())
+	}
+}
+
+// Test to make sure using dns-domain flag with no default-dns-domain config works as intended.
+func TestDnsDomainFlag(t *testing.T) {
+	// MaybeParallel(t)
+
+	profile := UniqueProfileName("domain-flag")
+	ctx, cancel := context.WithTimeout(context.Background(), Minutes(10))
+	defer CleanupWithLogs(t, profile, cancel)
+
+	startArgs := append([]string{Target(), "start", "-p", profile, "--dns-domain=flagcluster.example.com", "--wait=apiserver,system_pods,default_sa"}, StartArgs()...)
+	rr, err := Run(t, exec.CommandContext(ctx, "/usr/bin/env", startArgs...))
+	if err != nil {
+		t.Errorf("%s failed: %v", rr.Command(), err)
+	}
+
+	rr, err = Run(t, exec.CommandContext(ctx, KubectlBinary(), "--context", profile, "run", "busybox", "--rm", "--restart=Never", "--image=gcr.io/k8s-minikube/busybox", "-it", "--", "/bin/sh", "-c", "awk '$1 == \"search\" {print $2; if ($2 !~ /\\.flagcluster\\.example\\.com$/) { print $2;exit 1}}' /etc/resolv.conf"))
+	if err != nil {
+		t.Errorf("%s failed: %v", rr.Command(), err)
+		t.Errorf("Output: %s", rr.Output())
+	}
+
+	StopCluster(t, ctx, profile)
+
+	// make sure the dns domain doesn't change when no flag is given on subsequent start
+	startArgs = append([]string{Target(), "start", "-p", profile}, StartArgs()...)
+	rr, err = Run(t, exec.CommandContext(ctx, "/usr/bin/env", startArgs...))
+	if err != nil {
+		t.Errorf("%s failed: %v", rr.Command(), err)
+	}
+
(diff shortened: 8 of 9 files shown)


wt:
I asked about this feature here: https://kubernetes.slack.com/archives/C1F5CT6Q1/p1752280310063959

cmd/minikube/cmd/start_flags.go
```diff
@@ -636,7 +643,7 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 			APIServerName:          viper.GetString(apiServerName),
 			APIServerNames:         apiServerNames,
 			APIServerIPs:           apiServerIPs,
-			DNSDomain:              viper.GetString(dnsDomain),
```
medyagh, line 665: what happens to "dnsDomain" variable, is it used anywhere else ? should that be updated too

wt: It's a constant.

It's also use in the function that generates the config from an existing config when a cluster is already started.

This is precisely how I avoid changing the config of a running cluster with this default. The default is only used when starting up a new cluster.

However, you can change the dns-domain of an existing cluster by using the commandline flag still. I didn't want to change that behavior.

See [here](https://github.com/kubernetes/minikube/blob/cb277f62bd041c88ddb560353decdb541be79ab9/cmd/minikube/cmd/start_flags.go#L866).

cmd/minikube/cmd/start_flags.go
```diff
@@ -221,7 +221,9 @@ func initKubernetesFlags() {
 		Valid components are: kubelet, kubeadm, apiserver, controller-manager, etcd, proxy, scheduler
 		Valid kubeadm parameters: `+fmt.Sprintf("%s, %s", strings.Join(bsutil.KubeadmExtraArgsAllowed[bsutil.KubeadmCmdParam], ", "), strings.Join(bsutil.KubeadmExtraArgsAllowed[bsutil.KubeadmConfigParam], ",")))
 	startCmd.Flags().String(featureGates, "", "A set of key=value pairs that describe feature gates for alpha/experimental features.")
-	startCmd.Flags().String(dnsDomain, constants.ClusterDNSDomain, "The cluster dns domain name used in the Kubernetes cluster")
+
+	startCmd.Flags().String(dnsDomain, constants.DefaultClusterDNSDomain, "The cluster dns domain name used in the Kubernetes cluster")
```
medyagh: to summerize the converation in slack, do you mind putting in the PR descroption how is this different than 
"--dns-domain"

wt: Sure thing.

wt:
I have updated the commit message with details about the flag vs the config option. I hope this makes sense.

wt:
I think I have fixed all the lints/tests. I added to the docs for the config command as well.

medyagh (changes requested):
thank you @wt could you please

Put in the PR description, Before/After this PR what changes for the user?

and   how is this different than setting "--dns-domain" ?

wt:
> thank you @wt could you please
> 
> Put in the PR description, Before/After this PR what changes for the user?
> 
> and how is this different than setting "--dns-domain" ?

I think I have addressed this. Please let me know if more work is needed.

wt:
I am very confused by these test failures. I don't see how much change could have caused these failures. Do you have any suggestions for moving forward.

nirs:
> I am very confused by these test failures. I don't see how much change could have caused these failures. Do you have any suggestions for moving forward.

We have flaky test that fail all the time regardless of the change. It is also very hard to inspect the failures due to the excessive logging.

nirs:
> It is very handy to be able to store the default cluster dns name so that I can fire up and teardown clusters easily without losing that setting.
> 
> It's important to note that this default will only be applied at cluster creation time. If you have an existing cluster config, the name default name will not be changed if you added the config after starting the cluster the first time.

Isn't this the normal behavior for configs? If you create a cluster it uses defaults from the minikube config (e.g. number of cpus, number of disks, driver, container runtime). If the cluster exists we use the existing clusters so none of the configs modify it.

I don't fully understand the problem you are trying to solve, and why it cannot be solved with the existing --dns-domain option. Do we have an issue showing your use case? If not please open one showing the commands you are using when creating the clusters, and what is the issue you are trying to solve. In this PR you can show the same flow using the new config.

wt:
@nirs I don't want to have to remember to specify the `--dns-domain` flag every time I start a cluster the first time. The allows me to set my default in my config. With this config, my normal case of starting up a cluster is just `minikube start`. Without this config, I have to remember to run `minikube start --dns-domain cluster.wt.users.dev.example.com`, but only in the case where I am starting a my normal cluster config. This is literally a setting a default at the minikube level.

And...it appears to have the exact same semantics as some other settings. For example, by my reading of the code, the config option `iso-url` works exactly the same way. If you set that config, you get a default at cluster creation time. Setting that setting does not update an existing cluster, but there is a corresponding flag `--iso-url` that you can give during a creation or just restart. This is exactly the semantics that my change has.

cmd/minikube/cmd/start.go
```diff
@@ -161,6 +161,8 @@ func runStart(cmd *cobra.Command, _ []string) {
 	}
 	defer pkgtrace.Cleanup()
 
+	viper.SetDefault(config.DefaultClusterDNSDomain, constants.DefaultClusterDNSDomain)
```
medyagh: this should be func setupViper()

cmd/minikube/cmd/start_flags.go
```diff
@@ -560,6 +560,11 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 		out.WarningT("--network flag is only valid with the docker/podman, qemu, kvm, and vfkit drivers, it will be ignored")
 	}
 
+	defaultClusterDNSDomain := viper.GetString(dnsDomain)
+	if defaultClusterDNSDomain == constants.DefaultClusterDNSDomain {
```
medyagh: what is this code doing ? please add a comment for the block

medyagh (changes requested):
please put before/after this PR with full example to point out what is changing in this PR exactly and why do we need this

cmd/minikube/cmd/start.go
wt: I was trying to set the default only in the subcommand (`start`) that uses it. The `setupViper` function applies to all the subcommands. I am happy to move it there. Is this your recommendation?

wt:
> please put before/after this PR with full example to point out what is changing in this PR exactly and why do we need this

I already add a Before/After section to the commit message of the PR. Can you please help me understand what is missing from the description? Do you want me to put a commandline sequence and describe the difference like I did above the Before/After sections?

cmd/minikube/cmd/start_flags.go
wt: This code only exists in the case that a new cluster is started.

This code gets the flag value. If that flag is set, that value is applied as the dns domain for the cluster. If the flag is not set, the DefaultClusterDNSDomain configuration is used. If no configuration value is specified, the default ("cluster.local") is used.

wt:
@medyagh I added a bunch of before/after context. I think this addresses your latest comment. Please let me know if more is needed.

medyagh:
> commit

the Before/After Description is most useful in the PR Description for the reviwer and also in the changelog for the users to click on PRs to see what a PR is about, we dont utilize commit messages to be Long format, I prefer commit messages to be short and relevant to the commit and any other extra info should be in the PR Description

nirs:
> > commit
> 
> the Before/After Description is most useful in the PR Description for the reviwer and also in the changelog for the users to click on PRs to see what a PR is about, we dont utilize commit messages to be Long format, I prefer commit messages to be short and relevant to the commit and any other extra info should be in the PR Description

The PR message is hard to find when inspecting git history later, so all important content should be accessible `git log`. The PR message can be a copy of the commit message or add more details when we have multiple commits.

In this case the commit message looks very good (but I did not ready it yet), but the PR mesage is not helpful. @wt you can copy the commit message to the PR message to help reviewers.

wt:
I updated the original PR message to match the current commit message. I didn't realize you wanted me to change that before. Thanks for spelling that out.

wt:
FWIW, I have had some time to think about this change a bit more. I realized that a more powerful version of this change's idea might allow one to config only the `.local` part of the domain name.

This can be added on top of this change. I do not think that the scope of this change should be expanded to include this. I just wanted to see if y'all had any thoughts on this.

cmd/minikube/cmd/start_flags.go
wt: I added the comment to the code. Forgot to push until now.

medyagh:
@wt I just tested minikube on HEAD, here are example of output of Before/After that would help with review of PR:

## Before
in this example in minikube head, I start a cluster with  --dns-domain="cluster.example.com"
and I stop the cluster and I start it again without ----dns-domain (just plain minikube start) and it does preserve the dns-domain


```
$ minikube delete --all
$ minikube start --dns-domain="cluster.example.com"
$ kubectl run -it --rm --restart=Never busybox --image=busybox:1.28 -- sh
If you don't see a command prompt, try pressing enter.
/ # cat /etc/resolv.conf
nameserver 10.96.0.10
search default.svc.cluster.example.com svc.cluster.example.com cluster.example.com
options ndots:5


### Second Start
$ minikube start
$ kubectl run -it --rm --restart=Never busybox --image=busybox:1.28 -- sh
 # cat /etc/resolv.conf 
nameserver 10.96.0.10
search default.svc.cluster.example.com svc.cluster.example.com cluster.example.com

### Stopping
$ minikube stop

### third Start without any flags
$ minikube start
$ kubectl run -it --rm --restart=Never busybox --image=busybox:1.28 -- sh
If you don't see a command prompt, try pressing enter.
/ # cat /etc/resolv.conf 
nameserver 10.96.0.10
search default.svc.cluster.example.com svc.cluster.example.com cluster.example.com
options ndots:5
/ # 
```

regarding things only affecting clusters After, that is okay as long as it is same spirit for other commands, I wouldn't want to it to be a special case.

so I think what your PR wants to do is, Allow setting the "dns-domain" globally for all clusters? (checking is that what your intend it?)
in that case, please make an output of 3 minikube clusters

```
minikube start -p p1
minikube config set DefaultClusterDNSDomain cluster2.local
minikube start -p p2
minikube start -p p3
```
and then verify p2 and p3 have the Desired dns-domain as I did in the above ^^

medyagh (commented):
lets remove the go.work files and also do a go mod tidy and then I think should be good to go
(and also check if the logic for global config is same as other Global Configs)

cmd/minikube/cmd/config/config.go
```diff
@@ -172,6 +172,10 @@ var settings = []Setting{
 		name: config.MaxAuditEntries,
 		set:  SetInt,
 	},
+	{
+		name: config.DefaultClusterDNSDomain,
```
nirs: config.DefaultDNSDomain?

cmd/minikube/cmd/start_flags.go
```diff
@@ -560,6 +560,13 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 		out.WarningT("--network flag is only valid with the docker/podman, qemu, kvm, and vfkit drivers, it will be ignored")
 	}
 
+	// Set the cluster dns name to the dnsDomain flag. If the flag is not set, set it to
+	// the configured default.
+	defaultClusterDNSDomain := viper.GetString(dnsDomain)
```
nirs: dnsDomain - this is not the default value but the actual value we want to use.

cmd/minikube/cmd/start_flags.go
```diff
@@ -560,6 +560,13 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
...
+	if defaultClusterDNSDomain == constants.DefaultClusterDNSDomain {
+		defaultClusterDNSDomain = viper.GetString(config.DefaultClusterDNSDomain)
+	}
```
nirs: The rename will make this much more clear:

```go
if dnsDomain == constants.DefaultClusterDNSDomain {
    dnsDomain = viper.GetString(config.DefaultClusterDNSDomain)
}
```

Using 2 variables called defaultXXX made this code so confusing that we added a comment.

cmd/minikube/cmd/start_flags.go
```diff
@@ -636,7 +643,7 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 			APIServerName:          viper.GetString(apiServerName),
 			APIServerNames:         apiServerNames,
 			APIServerIPs:           apiServerIPs,
-			DNSDomain:              viper.GetString(dnsDomain),
+			DNSDomain:              defaultClusterDNSDomain,
```
nirs: The rename will make this much more clear and correct:

```go
DNSDomain: dnsDomain
```

pkg/minikube/config/config.go
```diff
@@ -58,6 +58,9 @@ const (
 	EmbedCerts = "EmbedCerts"
 	// MaxAuditEntries is the maximum number of audit entries to retain
 	MaxAuditEntries = "MaxAuditEntries"
+	// DefaultDnsDomain is the key for the default dns domain name when creating
+	// a cluster
+	DefaultClusterDNSDomain = "DefaultClusterDNSDomain"
```
nirs: "default-dns-domain"

We want to make it easy to related to `--dns-domain`.

pkg/minikube/constants/constants.go
```diff
@@ -75,7 +75,7 @@ const (
 	// APIServerName is the default API server name
 	APIServerName = "minikubeCA"
 	// ClusterDNSDomain is the default DNS domain
-	ClusterDNSDomain = "cluster.local"
+	DefaultClusterDNSDomain = "cluster.local"
```
nirs: DefaultDNSDomain

nirs (commented):
Looks useful, but the names are very confusing.

- We use defaultClusterDNSDomain for value which is not the default
- We use the term "cluster dns domain" in the code but "dns-domain" in the options
- The config name does not match the option name CamelCase vs lower-case.

Config names are a mess of CamelCase and lower-case. Let's not add more CamelCase names. In Go code we don't have a choice, but users should not pay this.

cmd/minikube/cmd/config/config.go
wt: Sure. I can do that.

cmd/minikube/cmd/start_flags.go
wt: dnsDomain is a const. I didn't want to shadow the name.

I did change the name to something shorter, and, I hope, better.

cmd/minikube/cmd/start_flags.go
wt: Nope, it's a const that is the name of the flag. I don't want to shadow the name. I did change it. Hopefully, that works for you.

cmd/minikube/cmd/start_flags.go
wt: I changed this a bit to follow the pattern used in updateExistingConfigFromFlags. This also fixes a problem where manually setting a cluster name to `cluster.local` with the flag wouldn't have worked before. Now it does.

wt:
Addressed all of @nirs's comments. Need to write the test.

cmd/minikube/cmd/start_flags.go
```diff
@@ -560,6 +560,9 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 		out.WarningT("--network flag is only valid with the docker/podman, qemu, kvm, and vfkit drivers, it will be ignored")
 	}
 
+	clusterDNSDomain := viper.GetString(config.DefaultDNSDomain)
+	updateStringFromFlag(cmd, &clusterDNSDomain, dnsDomain)
```
nirs, line 590: Much nicer than what we had before.

go.mod
```diff
@@ -46,6 +46,8 @@ require (
 	github.com/moby/hyperkit v0.0.0-20210108224842-2f061e447e14
 	github.com/moby/patternmatcher v0.6.0
 	github.com/olekukonko/tablewriter v1.0.8
+	github.com/onsi/ginkgo v1.14.0
+	github.com/onsi/gomega v1.35.1
```
nirs: We don't use ginkgo and gamete, why they are added here? Maybe you tried to add a test using them? Please don't add ginkgo to minikube, we use go test and it is much better.

Check with `git status` that you don't have any untracked go files in the minikube checkout, and run `go mod tidy` to clean up the unused modules.

nirs (commented):
Awesome! but we need to remove the go.mod changes.

go.mod
medyagh: yes plz try go mod tidy

nirs:
@wt can you update the PR? It can be a nice improvement for minkiube 1.37.

go.mod
nirs: @wt Maybe you have untracked go files using ginkgo? Please remove all untracked files from the source tree and run `make gomodtidy`.

nirs: `git clean -dxf` will clean up everything, but note that it will delete any file not tracked by git.

wt:
I've been really busy with personal business over the past couple week. I rebased the PR.

I am looking into the test, and I am trying to figure out how to test this functionality. Given the scope of this change, I think this means just checking that the code paths for existing and non-existing configs are changed as expected with and without the config. Where would that kind of testing belong? The previous suggestions of testing that within the integration testing seem like the wrong place logically. This feels like a unit testable piece of functionality. @medyagh or @nirs Is there some convenient place where the config generation is tested so that I can just add in a new unit test for the functionality?

FWIW, I think there is also a reason to test that the dns domain switch functionality in general (like when using the `--dns-domain` flag with the start subcommand) works at all. However, that feels out of scope for this change. I'm happy to try to figure out how to test that as well, but I think that, at the very least, belongs in a different commit.

nirs:
> I've been really busy with personal business over the past couple week. I rebased the PR.

Thanks it looks good now.

> I am looking into the test, and I am trying to figure out how to test this functionality. Given the scope of this change, I think this means just checking that the code paths for existing and non-existing configs are changed as expected with and without the config. Where would that kind of testing belong? The previous suggestions of testing that within the integration testing seem like the wrong place logically. This feels like a unit testable piece of functionality. @medyagh or @nirs Is there some convenient place where the config generation is tested so that I can just add in a new unit test for the functionality?

Since this change is just adding config for a constant, we can test it by checking the current value when the config is set and when it is not set. If the config is set should override the builtin default. Otherwise this change should have no effect on the behavior of the system. I'm not sure there is good place to test that since the code does not merge the config and constants in a unified way.

I think we should test:
1. minikube config shows the new key
2. minikube config get return the key value
3. minikube config set set the value
4. starting cluster with the config set pass the value to k8s
5. starting cluster without the config pass the default value to k8s

These tests have low maintenance cost. We never have to change the test when changing the code. The worst tests are those that we must change each time we change the code.

I hope we can use existing clusters int the current tests to validate tests 4 and 5. If not we need to add new cluster for testing. Maybe this cluster can be used to test many config values.

> FWIW, I think there is also a reason to test that the dns domain switch functionality in general (like when using the `--dns-domain` flag with the start subcommand) works at all. However, that feels out of scope for this change. I'm happy to try to figure out how to test that as well, but I think that, at the very least, belongs in a different commit.

I agree it is not the scope of this PR, and we want this feature in the next release which may be next week.

medyagh:
@wt can you please take a look at the unit test for macos ? https://github.com/kubernetes/minikube/actions/runs/16997163024/job/49206179194?pr=21064

wt:
@medyagh I literally have no test changes included in this change right now, so I don't think this change would have caused test failures yet. I am working on adding a unit test to test the config generation for the flag/config combinations. I will update this change with it...hopefully soon.

wt:
Well, I definitely did not get that sent out as soon as I said. Sorry about that. I have included my initial tries to get some testing working. As you can see, the testing function doesn't work. The main reason is that, AFAICT, the defaults for the command line flags cannot be setup and manipulated during the test. I was trying to try all the combinations of the cli arg and domain configuration and check the generated configs (both from existing and from new).

Do you have any suggestions on this? I do not actually see any other testing that relies on default flag values. Is it possible that there is a wide open gap in the testing that doesn't test most of the flags?

wt:
I just pushed a version with the appropriate integration tests. Unfortunately, the integration test cannot be parallel since they involve configuration.

I will just remove the `cmd/minikube/cmd/start_test.go` changes unless anyone has any useful ideas about any way to unit test this change.

wt:
I think the current push has addressed all comments and added all appropriate testing.

wt:
I think the test failure for the `pull-minikube-docker-crio-linux-x86` test is not due to this change. Is that being worked on?

PR #21086, opened 2025-07-17
build: Tidy go files

nirs:
It seems that recent dependencies changes were missing `go mod tidy`. Running it modified go.* files.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/Xuanwo/go-locale v1.1.3
 	github.com/blang/semver/v4 v4.0.0
 	github.com/briandowns/spinner v1.11.1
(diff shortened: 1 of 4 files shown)


medyagh (changes requested):
its weird if I run go mod tidy in both folders it wont genearte a diff but if I do
`go work sync` it will do tht but the diff is in the hack/go.mod do u mind trying go work sync?


lets make this PR ina way that runs go work sync automatically

nirs:
I run go work sync, it updated hack/go.mod. Posted at a new commit.

I'm not sure what go work sync is doing, and when you don't want to run it.

I think we should run this in the CI to ensure that all PRs are tidy. This will tell the author when go mod tidy or do work sync should run.

nirs:
Replaced by #21140

PR #21095, opened 2025-07-19
ci: add smoke test to github action

medyagh:

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -20,6 +20,13 @@ jobs:
       matrix:
         os: [macos-13]
         driver: [qemu, vfkit]
+        include:
+          - driver: qemu
+            os: macos-13
+            network_flag: --network socket_vmnet
+          - driver: vfkit
+            os: macos-13
+            network_flag: --network vmnet-shared
     steps:
       - name: Info Block
         shell: bash
@@ -51,11 +58,17 @@ jobs:
           sudo $fw --remove /usr/libexec/bootpd
           sudo $fw --add /usr/libexec/bootpd
           sudo $fw --unblock /usr/libexec/bootpd
-      - name: Install vfkit
+      - name: Install vfkit and vmnet-helper
         if: matrix.driver == 'vfkit'
         run: |
           brew update
           brew install vfkit
+          machine="$(uname -m)"
+          archive="vmnet-helper-$machine.tar.gz"
+          curl -LOf "https://github.com/nirs/vmnet-helper/releases/latest/download/$archive"
+          sudo tar xvf "$archive" -C / opt/vmnet-helper
+          rm "$archive"
+          sudo install -m 0640 /opt/vmnet-helper/share/doc/vmnet-helper/sudoers.d/vmnet-helper /etc/sudoers.d/
       - name: Install qemu and socket_vmnet 
         if: matrix.driver == 'qemu'
         run: |
@@ -66,7 +79,7 @@ jobs:
       - name: Start minikube (1st boot)
         run: |
           set -ex
-          ./out/minikube start --driver=${{ matrix.driver }} --memory 8gb --no-kubernetes --alsologtostderr
+          ./out/minikube start --driver=${{ matrix.driver }} ${{ matrix.network_flag }} --memory 8gb --no-kubernetes --alsologtostderr
       - name: Stop minikube
         run: |
           set -ex
@@ -74,8 +87,8 @@ jobs:
       - name: Start minikube again (2nd boot)
         run: |
           set -ex
-          ./out/minikube start --driver=${{ matrix.driver }} --no-kubernetes --alsologtostderr
+          ./out/minikube start --driver=${{ matrix.driver }} ${{ matrix.network_flag }} --no-kubernetes --alsologtostderr
       - name: delete minikube 
         run: |
           set -ex
-          ./out/minikube --delete --alsologtostderr
+          ./out/minikube delete --alsologtostderr
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -39,6 +39,26 @@ jobs:
           system_profiler SPHardwareDataType
           sw_vers
           ifconfig
+      - name: Info Block (linux)
+        if: contains(matrix.os, 'ubuntu')
+        shell: bash
+        run: |
+            uname -a
+            grep MemTotal /proc/meminfo
+            awk '/MemTotal/ {printf "%.2f\n", $2 / 1024 / 1024}' /proc/meminfo
+            nproc
+            grep 'model name' /proc/cpuinfo | head -n1
+            # System model (requires `dmidecode`, may need sudo)
+            sudo dmidecode -s system-product-name
+            systemd-detect-virt
+            # Check if KVM is available (can show virtualization support)
+            egrep -c '(vmx|svm)' /proc/cpuinfo
+            # Detailed hardware info (like `system_profiler`)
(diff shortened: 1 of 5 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,81 @@
...
+          echo "$(sysctl -n hw.memsize) / 1024 / 1024 / 1024" | bc
+          sysctl -n hw.ncpu
+          sysctl -n machdep.cpu.brand_string
+          sysctl hw.model
+          sysctl -n kern.hv_vmm_present
+          sysctl -n kern.hv_support
+          system_profiler SPHardwareDataType
+          sw_vers
+          ifconfig
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
```
nirs, line 62: It helps to add a comment with the version of this commit.

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,81 @@
...
+      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5
+        with:
+          go-version: ${{env.GO_VERSION}}
```
nirs: We can take the go version from go.mod:
https://github.com/nirs/kubectl-gather/blob/35285b22ac027390b75d25b98369a7e73e436ea7/.github/workflows/test.yaml#L24

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,81 @@
...
+        run: |
+          set -x
+          fw=/usr/libexec/ApplicationFirewall/socketfilterfw
+          sudo $fw --remove /usr/libexec/bootpd
+          sudo $fw --add /usr/libexec/bootpd
+          sudo $fw --unblock /usr/libexec/bootpd
+      - name: Install vfkit
+        if: matrix.driver == 'vfkit'
+        run: |
+          brew update
```
nirs, line 83: We can move brew update to another step instead of calling it twice.

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,81 @@
...
+          brew install vfkit
+      - name: Install qemu and socket_vmnet 
+        if: matrix.driver == 'qemu'
+        run: |
+          brew update
+          brew install qemu
+          brew install socket_vmnet
```
nirs: Can one line:

    brew install qemu socket_vment

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,81 @@
...
+          HOMEBREW=$(which brew) && sudo ${HOMEBREW} services start socket_vmnet
```
nirs, line 88: This is unsafe for good better for the CI.

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,81 @@
...
+          set -ex
+          ./out/minikube stop
+      - name: Start minikube again (2nd boot)
+        run: |
+          set -ex
+          ./out/minikube start --driver=${{ matrix.driver }} --no-kubernetes --alsologtostderr
+      - name: delete minikube 
+        run: |
+          set -ex
+          ./out/minikube --delete --alsologtostderr
```
nirs: Nice and simple!

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,81 @@
...
+      - name: Install qemu and socket_vmnet 
+        if: matrix.driver == 'qemu'
+        run: |
+          brew update
+          brew install qemu
+          brew install socket_vmnet
+          HOMEBREW=$(which brew) && sudo ${HOMEBREW} services start socket_vmnet          
+      - name: Start minikube (1st boot)
+        run: |
+          set -ex
```
nirs: -e is the default, and this is one command so we don't really need -x. I would be nicer as:

    - name: ...
      run: out/minikube start ...

.github/workflows/smoke-test.yml
medyagh: this is auto-updated by dependabot I think so it will be useless to add version in comment

.github/workflows/smoke-test.yml
```diff
@@ -66,16 +79,16 @@ jobs:
...
+          ./out/minikube start --driver=${{ matrix.driver }} ${{ matrix.network_flag }} --memory 8gb --no-kubernetes --alsologtostderr
       - name: Stop minikube
         run: |
           set -ex
           ./out/minikube stop
       - name: Start minikube again (2nd boot)
         run: |
           set -ex
-          ./out/minikube start --driver=${{ matrix.driver }} --no-kubernetes --alsologtostderr
+          ./out/minikube start --driver=${{ matrix.driver }} ${{ matrix.network_flag }} --no-kubernetes --alsologtostderr
```
nirs: On the second start we don't need any argument. We know that these arguments are ignored (not sure about --no-kubernetes).

.github/workflows/smoke-test.yml
```diff
@@ -66,7 +66,7 @@ jobs:
       - name: Start minikube (1st boot)
         run: |
           set -ex
-          ./out/minikube start --driver=${{ matrix.driver }} --no-kubernetes --alsologtostderr
+          ./out/minikube start --driver=${{ matrix.driver }} --memory 8gb --no-kubernetes --alsologtostderr
```
nirs: For boot there is no reason to use 8g, 2g must work.

pkg/drivers/hyperkit/driver.go
```diff
@@ -296,7 +297,11 @@ func (d *Driver) setupIP(mac string) error {
 	var err error
 
 	// Implement a retry loop without calling any minikube code
-	for i := 0; i < 30; i++ {
+	multiplier := 1
+	if detect.NestedVM() {
```
nirs, line 301: I would not add this, instead make it configurable and let the CI set high value. This is not useful or real usage of minikube.

This can be set by setup-minikube action since it knows that it runs with slow nested runners.

pkg/minikube/detect/detect.go
```diff
@@ -212,3 +212,29 @@ func MacOS13Plus() bool {
 	}
 	return major >= 13
 }
+
+// NestedVM returns true if the current machine is running a nested VM (like in MacOs in Github Action)
+func NestedVM() bool {
+	if runtime.GOOS == "linux" {
+		c := exec.Command("systemd-detect-virt", "--quiet")
```
nirs, line 220: systemd is required? It will be best to do this outside of minikube.

.github/workflows/smoke-test.yml
medyagh: done thanks :)

pkg/drivers/hyperkit/driver.go
medyagh: I think it is helpful, I rather minikube do the right automatically thing than have Other third party tool setting things, pain of maitaining setup-minikube or even the tools that are not built by me, so better minikube does it for you !

making it retry more on a "nested Macos" seems reasonable, since we only bumping retry effectively on Macos nested, and wont break or make things worse

.github/workflows/smoke-test.yml
medyagh: yeah I had it without args at first I wanted to double check that is not casuing the failure

pkg/minikube/detect/detect.go
medyagh: yeah will need to find a way for non-systemd too (probably) not ever needed practically beyond macos, only put there as a Good to have

PR #21100, opened 2025-07-20
vfkit: Restore compatibility with x86_64

nirs:
In #20833 we switch to --bootloader=efi, breaking vfkit on x86_64. Switch to --bootloader=linux to restore support for older macs. This also enable running vfkit basic tests in github runners.

This change does not revert #20833 since it contained other important changes that we squashed during merge. We also use the new --bootloader=linux instead of the legacy --kernel, --initrd, and --cmdline flags.

Issues:
- On x86_64 using console=hvc0 breaks boot. Using console=ttyS0 works,
  but serial.log is always empty.

Tested with:
- [x] arm64 (locally)
- [x] x86_64 (#21102)

### DIFF
--- a/pkg/drivers/vfkit/vfkit.go
+++ b/pkg/drivers/vfkit/vfkit.go
@@ -30,6 +30,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


afbjorklund (approved):
Sounds good to me, didn't test

nirs:
Moved to draft since it seems that the machine does not boot in GitHub runner.

nirs:
@afbjorklund I changed to console=ttyS0 on x86_64. Using console=hvc0 breaks boot on GitHub runner:
https://github.com/kubernetes/minikube/compare/f1c86fe869a3cd7063cabf06a14f4db8d16030f8..d8a7522245a769838914f15cfdaaf8f555ed259e

I'm trying to fix this in #21101

nirs:
Latest push improves the comment for the cmdline argument.

PR #21101, opened 2025-07-20
iso: Enable Xen Hypervisor Console support for x86_64

nirs:
vfkit serial log is always empty on x86_64. With Ubuntu 24.10 we get full serial log when using console=hvc0. I think this is enabled by CONFIG_HVC_XEN. Adding CONFIG_XEN=y enables CONFIG_HVC_XEN.

This does not help.

I tested vfkit with:
- console=hvc0: breaks boot
- console=ttyS0 console=hvc0: boots but serial.log is still empty

Leaving open in case someone knows how to configure the kernel command line to make it work.


nirs:
ok-to-build-iso

nirs:
Probably the wrong change.

PR #21102, opened 2025-07-20
ci: Add vfkit smoke test

nirs:
Clean up the smoke test workflow a little bit and add a new vfkit smoke test.

The test is about much slower than running vfkit with Ubuntu 24.10 image in GitHub runners (as part of vmnet-helper integration tests), but it works most of the time, and it is much faster than the qemu test.

Example build: https://github.com/nirs/minikube/actions/runs/16405189904

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -88,9 +88,9 @@ jobs:
           HOMEBREW=$(which brew) && sudo ${HOMEBREW} services start socket_vmnet
       - name: Minikube Download only (iso/image)
         run:
-          ./out/minikube start --download-only --driver=${{ matrix.driver }}
+          ./out/minikube start --download-only --driver ${{ matrix.driver }}
       - name: Start minikube (1st boot)
-        run: ./out/minikube start --driver=${{ matrix.driver }} ${{ matrix.network_flag }} --memory 4gb --no-kubernetes --alsologtostderr
+        run: ./out/minikube start --driver ${{ matrix.driver }} ${{ matrix.network_flag }} --memory 4gb --no-kubernetes --alsologtostderr
       - name: Stop minikube
         run: ./out/minikube stop
       - name: Start minikube again (2nd boot)
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -68,8 +68,8 @@ jobs:
         run: go mod download
       - name: Build Binaries
         run: make
-      - name: Ensure bootpd is enabled
-        if: contains(matrix.os, 'macos')
+      - name: Ensure bootpd is enabled (macos)
+        if: matrix.os == 'macos-13'
         shell: bash
         run: |
           set -x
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -68,32 +71,53 @@ jobs:
...
       - name: Start minikube (1st boot)
-        run: ./out/minikube start --driver=${{ matrix.driver }} ${{ matrix.network_flag }} --memory 4gb --no-kubernetes --alsologtostderr 
+        run: ./out/minikube start --driver ${{ matrix.driver }} ${{ matrix.network_flag }} --memory 4gb --no-kubernetes --alsologtostderr
+      - name: Inspect machine
+        if: always()
+        run: |
+          tree -h ~/.minikube
+          machine="$HOME/.minikube/machines/minikube"
+          machine_files=$(find "$machine" -name "*.log")
+          for f in $machine_files /var/db/dhcpd_leases; do
```
medyagh: while we could also add file too
~/.minikube/logs/lastStart.txt

.github/workflows/smoke-test.yml
```diff
@@ -68,32 +71,53 @@ jobs:
...
-        if: contains(matrix.os, 'macos') && matrix.driver == 'qemu'
+        if: matrix.os == 'macos-13' && matrix.driver == 'qemu'
         run: |
           brew install qemu socket_vmnet
-          HOMEBREW=$(which brew) && sudo ${HOMEBREW} services start socket_vmnet          
+          HOMEBREW=$(which brew) && sudo ${HOMEBREW} services start socket_vmnet
       - name: Minikube Download only (iso/image)
         run:
-          ./out/minikube start --download-only --driver=${{ matrix.driver }} 
+          ./out/minikube start --download-only --driver ${{ matrix.driver }}
```
medyagh: lets add -v=8 to all minikube commands that will include the lib-machine super vebbose machine logs too

.github/workflows/smoke-test.yml
```diff
@@ -68,32 +71,53 @@ jobs:
...
       - name: Start minikube (1st boot)
-        run: ./out/minikube start --driver=${{ matrix.driver }} ${{ matrix.network_flag }} --memory 4gb --no-kubernetes --alsologtostderr 
+        run: ./out/minikube start --driver ${{ matrix.driver }} ${{ matrix.network_flag }} --memory 4gb --no-kubernetes --alsologtostderr
```
medyagh: -v=8 here too

.github/workflows/smoke-test.yml
```diff
@@ -68,32 +71,53 @@ jobs:
...
+          machine="$HOME/.minikube/machines/minikube"
+          machine_files=$(find "$machine" -name "*.log")
+          for f in $machine_files /var/db/dhcpd_leases; do
+            echo "==> $f <=="
+            head -n 1000 "$f" || true
+          done
       - name: Stop minikube
         run: ./out/minikube stop
       - name: Start minikube again (2nd boot)
         run: ./out/minikube start --alsologtostderr
```
medyagh: -v=8 here too

PR #21111, opened 2025-07-21
vfkit: Extract isoPath() helper

nirs:
For consistency with other path helpers.


nirs:
@medyagh should be ready, I rebased on master to run the vfkit smoke test.

medyagh (approved):

nirs:
@medyagh The smoke test failed (error in restarting docker), but this is not related to the change. We can run the job again to be sure.

PR #21135, opened 2025-07-24
refactor: use slices.Equal to simplify code

jishudashu:
In the Go 1.21 standard library, a new function has been introduced that enhances code conciseness and readability. It can be find  [here](https://pkg.go.dev/slices@go1.21.1#Equal).

### DIFF
--- a/pkg/drivers/hyperkit/driver_test.go
+++ b/pkg/drivers/hyperkit/driver_test.go
@@ -19,6 +19,7 @@ limitations under the License.
 package hyperkit
 
 import (
+	"slices"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/drivers/hyperkit/driver_test.go
```diff
@@ -72,15 +73,5 @@ func testEq(a, b []int) bool {
...
-	}
-
-	for i := range a {
-		if a[i] != b[i] {
-			return false
-		}
-	}
-
-	return true
+	return slices.Equal(a, b)
```
nirs, line 76: We can drop testEqual and use slices.Equal directly. But this driver is deprecated so it is not worth the time to improve the tests. We should focus on vfkit and krunkit drivers for macOS.

PR #21136, opened 2025-07-24
Update Headlamp's workflow go version

joaquimrocha:


.github/workflows/update-headlamp-version.yml
```diff
@@ -6,7 +6,7 @@ on:
     - cron: "0 10 * * 1"
 env:
   GOPROXY: https://proxy.golang.org
-  GO_VERSION: '1.23.4'
+  GO_VERSION: '1.24.0'
```
medyagh, line 9: please update the source of the issue, which is in the go-updater
check the script that is called by 
```
make update-golang-version
```

that one is missing this file to update

medyagh (changes requested):

.github/workflows/update-headlamp-version.yml
joaquimrocha: Left a comment at https://github.com/kubernetes/minikube/pull/21157#issuecomment-3133143409

Issue #21107, opened 2025-07-21
dont attempt to download kubeadm or kubelet when --no-kubernetes


medyagh:
in the smoke test that runs with --no-kubernetes I noticed this logs

```
I0721 19:21:51.018367    2366 ssh_runner.go:195] Run: /bin/bash -c "sudo mkdir -p /etc && printf %s "runtime-endpoint: unix:///var/run/dockershim.sock
" | sudo tee /etc/crictl.yaml"
I0721 19:21:51.035824    2366 command_runner.go:130] > runtime-endpoint: unix:///var/run/dockershim.sock
I0721 19:21:51.037250    2366 download.go:108] Downloading: https://dl.k8s.io/release/v0.0.0/bin/linux/amd64/kubeadm?checksum=file:https://dl.k8s.io/release/v0.0.0/bin/linux/amd64/kubeadm.sha1 -> /home/runner/.minikube/cache/linux/amd64/v0.0.0/kubeadm
```

our code should NOT even attempt to download "v0.0.0" stuff niether in 
minikube start --no-kubernetes
or 
minikube start --download-only --no-kubernetes
when there is --no-kubernetes flag is set

PR #21139, opened 2025-07-25
Skip Kubernetes preloads and binary downloads when --no-kubernetes is… #21107

divysinghvi:
This PR improves the Binary function in binary.go to better handle Kubernetes binary downloads. It adds logic to skip downloads when the --no-kubernetes flag is set, ensures not downloading unnecessary downloads with version v0.0.0
Before
Screenshot 2025-07-25 at 1 05 43 PM
After
Screenshot 2025-07-25 at 1 08 30 PM
This pr Fixes the issue #21107

### DIFF
--- a/pkg/minikube/download/binary.go
+++ b/pkg/minikube/download/binary.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/blang/semver/v4"
 	"github.com/pkg/errors"
+	"github.com/spf13/viper"
 	"k8s.io/klog/v2"
 	"k8s.io/minikube/pkg/minikube/localpath"
 )
@@ -53,6 +54,11 @@ func binaryWithChecksumURL(binaryName, version, osName, archName, binaryURL stri
 
 // Binary will download a binary onto the host
 func Binary(binary, version, osName, archName, binaryURL string) (string, error) {
+	// Prevent Kubernetes binary downloads in --no-kubernetes mode
+	if viper.GetBool("no-kubernetes") {
+		klog.Infof("Skipping Kubernetes binary download due to --no-kubernetes flag")
+		return "", nil
+	}
 	targetDir := localpath.MakeMiniPath("cache", osName, archName, version)
 	targetFilepath := path.Join(targetDir, binary)
 	targetLock := targetFilepath + ".lock"
--- a/pkg/minikube/download/preload.go
+++ b/pkg/minikube/download/preload.go
@@ -121,6 +121,11 @@ var checkRemotePreloadExists = func(k8sVersion, containerRuntime string) bool {
 
 // PreloadExists returns true if there is a preloaded tarball that can be used
 func PreloadExists(k8sVersion, containerRuntime, driverName string, forcePreload ...bool) bool {
+	// Prevent preload logic in --no-kubernetes mode
+	if viper.GetBool("no-kubernetes") {
+		klog.Infof("Skipping preload logic due to --no-kubernetes flag")
+		return false
+	}
 	// TODO (#8166): Get rid of the need for this and viper at all
 	force := false
 	if len(forcePreload) > 0 {


divysinghvi:
@medyagh Thanks for the suggestion! Just to clarify, for the integration test you mentioned, which exact Kubernetes binary path should I check for in the VM after running minikube with` --no-kubernetes --download-only`?
but i am confused what exact binary path should i check for

medyagh:
Lets also try not setting k8s version to 0.0.0 and see what breaks and why did we have to that

medyagh (changes requested):
lets add a sub test that that checks this folder should not exist
 ~/.minikube/cache/linux/amd64/v0.0.0/

medyagh:
ok-to-test

test/integration/no_kubernetes_test.go
```diff
@@ -74,6 +77,28 @@ func TestNoKubernetes(t *testing.T) {
 	})
 }
 
+// VerifyNoK8sDownloadCache verifies that starting minikube with --no-kubernetes does not create a download cache.
+func VerifyNoK8sDownloadCache(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+
+	// Start minikube with --no-kubernetes flag
+	args := append([]string{"start", "-p", profile, "--no-kubernetes", "--memory=2048"}, StartArgs()...)
```
medyagh: no need to start another minikube, use the Existing started minikube and simply add a sub test
like this
			{"StartNoK8sWithVersion", validateStartNoK8sWithVersion},
--->	    {"VerifyNok8sNoK8sDownloads", },

			{"StartWithK8s", validateStartWithK8S},
			{"StartWithStopK8s", validateStartWithStopK8s},

medyagh (changes requested):

test/integration/no_kubernetes_test.go
divysinghvi: since in StartNoK8sWithVersion we pass a Kubernetes version with --no-kubernetes isn't that suppose to exits with an error and that won't leave a running cluster to reuse i guess.

divysinghvi: maybe we are reuse this instance 	{"Start", validateStartNoK8S}, since it will be fresh working instance

test/integration/no_kubernetes_test.go
```diff
@@ -74,6 +77,21 @@ func TestNoKubernetes(t *testing.T) {
 	})
 }
 
+// VerifyNoK8sDownloadCache verifies that starting minikube with --no-kubernetes does not create a download cache.
+func VerifyNoK8sDownloadCache(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+
+	// Reuse the minikube instance started by validateStartNoK8S.
+	homeDir := os.Getenv("HOME")
```
medyagh: check other tests, see if we have already sometihng for minikube home

divysinghvi: In no_kubernetes_test.go no other has a home path, since they mostly interact was with minikube using via CLI commands, for the minikube home thing i found one instance in cache_binaries_test.go they use MINIKUBE_HOME , can you suggest any other way i can check this thanks

medyagh: @divysinghvi you shoud use localpath.MiniPath()
here is an example in test/integration/aaa_download_only_test.go

```
				// checking binaries downloaded (kubelet,kubeadm)
				for _, bin := range constants.KubernetesReleaseBinaries {
					fp := filepath.Join(localpath.MiniPath(), "cache", "linux", runtime.GOARCH, v, bin)
					_, err := os.Stat(fp)
					if err != nil {
						t.Errorf("expected the file for binary exist at %q but got error %v", fp, err)
					}
				}
```

divysinghvi: thanks for help i will look into it

test/integration/no_kubernetes_test.go
```diff
@@ -76,6 +81,20 @@ func TestNoKubernetes(t *testing.T) {
...
 
+// VerifyNoK8sDownloadCache verifies that starting minikube with --no-kubernetes does not create a download cache.
+func VerifyNoK8sDownloadCache(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+
+	// Reuse the minikube instance started by validateStartNoK8S.
+	cachePath := filepath.Join(localpath.MiniPath(), "cache", "linux", runtime.GOARCH, "v0.0.0")
+
+	if _, err := os.Stat(cachePath); err == nil {
+		t.Fatalf("Cache directory %s should not exist when using --no-kubernetes", cachePath)
```
medyagh: t.Errorf(...)

test/integration/no_kubernetes_test.go
```diff
@@ -76,6 +81,20 @@ func TestNoKubernetes(t *testing.T) {
...
+	} else if err != nil && !os.IsNotExist(err) {
+		t.Fatalf("Error checking cache directory %s: %v", cachePath, err)
```
medyagh: t.Errorf(

medyagh (changes requested):

test/integration/no_kubernetes_test.go
```diff
@@ -76,6 +81,32 @@ func TestNoKubernetes(t *testing.T) {
 	})
 }
 
+// VerifyNoK8sDownloadCache verifies that starting minikube with --no-kubernetes does not create a download cache.
+func VerifyNoK8sDownloadCache(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+
+	// Clean any existing cache from previous tests to ensure we test only --no-kubernetes behavior
+	cachePath := filepath.Join(localpath.MiniPath(), "cache", "linux", runtime.GOARCH, "v0.0.0")
```
medyagh: question is why Other tests generated any folder with "v0.0.0" ?
how about instead of this, When it fails we can list the minikube home and check why it is there in first place.

Before minikube test starts, we purge and have we have separte minikube home for the test

medyagh (changes requested):

test/integration/no_kubernetes_test.go
divysinghvi: that would be way better approach i had some issue with it so i was experimenting with thing i will look more into it thanks , yes the other tests should also not generate the folder with "v0.0.0" i was just experimenting with things if that fixes it and also one more issue i am facing is the tests are passing on my system since my last 3 commit 
WhatsApp Image 2025-09-26 at 14 58 12

test/integration/no_kubernetes_test.go
```diff
@@ -76,6 +81,32 @@ func TestNoKubernetes(t *testing.T) {
 	})
 }
 
+// VerifyNoK8sDownloadCache verifies that starting minikube with --no-kubernetes does not create a download cache.
+func VerifyNoK8sDownloadCache(ctx context.Context, t *testing.T, profile string) {
+	defer PostMortemLogs(t, profile)
+
+	cachePath := filepath.Join(localpath.MiniPath(), "cache", "linux", runtime.GOARCH, "v0.0.0")
```
medyagh: this should not be tested only on linux !
the reason it is passing on your machine it is beecause it checks to see if linux doesnt exist

you can use goos to get os

medyagh (changes requested):

test/integration/no_kubernetes_test.go
divysinghvi: Thanks medhya i have now used GOOS to get current os, and yes it passes on my system now also
`=== RUN   TestNoKubernetes/serial/VerifyNok8sNoK8sDownloads
    no_kubernetes_test.go:91: Checking cache directory: /Users/divysinghvi/.minikube/cache/darwin/arm64/v0.0.0
    no_kubernetes_test.go:104: No cache directory found (as expected)`
also found a function in binary.go TransferBinaries this maybe creating directories so testing it now

divysinghvi:
@medyagh i tired putting a lot of checks but was not able to identify what is causing the directory to get created but we are able to step the files being created, directory thing i am not able to verify what is causing that you also told me to check localpath.go but i was not able to identify specific for our purpose , now we have either one option we go with the only file approach and have directory check but the test fails only when we have the files in those or figure out a way for finding directory thing in linux

medyagh:
it is interesting on my mac and personal linux it does not create v0.0.0 folder but in CI machines it does ! this is a mystrie !

divysinghvi:
yes @medyagh i have tried everything i could find like i am still not able to figure how folder is getting created in the ci testing it is so weird , since now even you have verified it is not creating the v 0.0.0 when are testing it locally how should be proceed it next? easiest approach would be to check for files within the folder and let it pass when the directory is empty or anything other in mind ??

PR #21145, opened 2025-07-25
adding addons lisst if status enabled whole row green

pavansaikrishna78:
image

Updated the code and written the unittest unit test are passing


go.work.sum
```diff
@@ -1,6 +1,193 @@
+cloud.google.com/go/accessapproval v1.8.6/go.mod h1:FfmTs7Emex5UvfnnpMkhuNkRCP85URnBFt5ClLxhZaQ=
```
medyagh: you can remove this file from your PR, since we removed it from master

medyagh (changes requested):

medyagh:
@pavansaikrishna78 this what I see in your PR, what I meant, was if the Addons is Enabled, the word Enabled should be Green

Screenshot 2025-07-25 at 3 51 51 PM

pavansaikrishna78:
i had used go-pretty because i am not able to  change for the specific row using the previous package with go-pretty it is working fine 
image

pavansaikrishna78:
update the pr for the disabled one if it is disabled then it will show red 
image

PR #21174, opened 2025-07-28
improve docker service reliability, update docker systemd files

ComradeProgrammer:
Previous docker.service file generated by minikube is too old, and it does not wait for the containerd. This pr will update the docker.service according to the latest one.

### DIFF
--- a/pkg/provision/buildroot.go
+++ b/pkg/provision/buildroot.go
@@ -70,14 +70,16 @@ func (p *BuildrootProvisioner) GenerateDockerOptions(dockerPort int) (*provision
 		klog.Infof("root file system type: %s", fstype)
 		noPivot = fstype == "rootfs"
 	}
-
+	/* Template is copied from https://github.com/moby/moby/blob/44bca1adf3c806c4ed6688d67c6f995653b09b26/contrib/init/systemd/docker.service#L2 with the following changes:
+	   After: minikube-automount.service is added
+	   ExecStart: additional flags are added
+	*/
 	engineConfigTmpl := `[Unit]
 Description=Docker Application Container Engine
 Documentation=https://docs.docker.com
-After=network.target  minikube-automount.service docker.socket
-Requires= minikube-automount.service docker.socket 
-StartLimitBurst=3
-StartLimitIntervalSec=60
+After=network-online.target minikube-automount.service nss-lookup.target docker.socket firewalld.service containerd.service time-set.target
+Wants=network-online.target containerd.service
+Requires=docker.socket
 
 [Service]
 Type=notify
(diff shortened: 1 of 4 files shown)


pkg/provision/buildroot.go
```diff
@@ -117,13 +106,13 @@ LimitCORE=infinity
...
 # Only systemd 226 and above support this version.
 TasksMax=infinity
-TimeoutStartSec=0
 
 # set delegate yes so that systemd does not reset the cgroups of docker containers
 Delegate=yes
 
 # kill only the docker process, not all processes in the cgroup
 KillMode=process
+OOMScoreAdjust=-500
```
medyagh, line 126: how come we dont have this in ubuntu? btw I am curious is this in the standard one ?
it does seem helpful

pkg/provision/buildroot.go
```diff
@@ -74,10 +74,9 @@ func (p *BuildrootProvisioner) GenerateDockerOptions(dockerPort int) (*provision
 	engineConfigTmpl := `[Unit]
 Description=Docker Application Container Engine
 Documentation=https://docs.docker.com
-After=network.target  minikube-automount.service docker.socket
-Requires= minikube-automount.service docker.socket
```
medyagh, line 78: put back minikube-automount

medyagh (changes requested):

pkg/provision/buildroot.go
```diff
@@ -74,10 +74,9 @@ func (p *BuildrootProvisioner) GenerateDockerOptions(dockerPort int) (*provision
 	engineConfigTmpl := `[Unit]
 Description=Docker Application Container Engine
 Documentation=https://docs.docker.com
-After=network.target  minikube-automount.service docker.socket
-Requires= minikube-automount.service docker.socket 
-StartLimitBurst=3
-StartLimitIntervalSec=60
+After=network-online.target nss-lookup.target docker.socket firewalld.service containerd.service time-set.target
```
medyagh: -~please comment where the Service files are coming from, so in the the future we can automate that, add a TODO comment~- 
actually just commnt what is different

medyagh: also comment ANY line which is NOT from upstream

pkg/provision/buildroot.go
```diff
@@ -94,7 +94,7 @@ Environment=DOCKER_RAMDISK=yes
 {{end}}
 
 ExecStart=
-ExecStart=/usr/bin/dockerd -H fd:// --containerd=/run/containerd/containerd.sock -H unix:///var/run/docker.sock --default-ulimit=nofile=1048576:1048576 --tlsverify --tlscacert {{.AuthOptions.CaCertRemotePath}} --tlscert {{.AuthOptions.ServerCertRemotePath}} --tlskey {{.AuthOptions.ServerKeyRemotePath}} {{ range .EngineOptions.Labels }}--label {{.}} {{ end }}{{ range .EngineOptions.InsecureRegistry }}--insecure-registry {{.}} {{ end }}{{ range .EngineOptions.RegistryMirror }}--registry-mirror {{.}} {{ end }}{{ range .EngineOptions.ArbitraryFlags }}--{{.}} {{ end }}
+ExecStart=/usr/bin/dockerd -H tcp://0.0.0.0:2376 -H fd:// --containerd=/run/containerd/containerd.sock -H unix:///var/run/docker.sock --default-ulimit=nofile=1048576:1048576 --tlsverify --tlscacert {{.AuthOptions.CaCertRemotePath}} --tlscert {{.AuthOptions.ServerCertRemotePath}} --tlskey {{.AuthOptions.ServerKeyRemotePath}} {{ range .EngineOptions.Labels }}--label {{.}} {{ end }}{{ range .EngineOptions.InsecureRegistry }}--insecure-registry {{.}} {{ end }}{{ range .EngineOptions.RegistryMirror }}--registry-mirror {{.}} {{ end }}{{ range .EngineOptions.ArbitraryFlags }}--{{.}} {{ end }}
```
nirs: Not related to this change, but if we get format this command as multiline command it will be much more readable, and changes will be single line:

    ecStart=/usr/bin/dockerd -H tcp://0.0.0.0:2376 \
        -H fd:// --containerd=/run/containerd/containerd.sock \
        -H unix:///var/run/docker.sock \
        ...

And if we can do this, using long options when possible will make the command much more clear. My rule is use long options for code running another program, use short options when working in the shell.

Lets not make this change in this commit to keep the change small.

nirs (commented):
Nice to see the smoke tests green!

pkg/provision/buildroot.go
medyagh: -~@nirs I wonder if that 0.0.0.0:2376 is changing the listen binding ?
I think it was alreayd 0.00.0 No ? (since we can talk to the docker inside minikube using docker-env?) just thinking would that be a NEW seurity issue or was it a always like that~-

pkg/provision/ubuntu.go
```diff
@@ -75,12 +75,9 @@ func (p *UbuntuProvisioner) GenerateDockerOptions(dockerPort int) (*provision.Do
 	engineConfigTmpl := `[Unit]
 Description=Docker Application Container Engine
 Documentation=https://docs.docker.com
-BindsTo=containerd.service
-After=network-online.target firewalld.service containerd.service
-Wants=network-online.target
+After=network-online.target nss-lookup.target docker.socket firewalld.service containerd.service time-set.target
+Wants=network-online.target containerd.service
 Requires=docker.socket
-StartLimitBurst=3
```
medyagh, line 82: I think we should keep  these because this will prevent minikube's docker to restart too many times on no so powerful 
systems 


StartLimitBurst=3
→ The service can be restarted up to 3 times within the time window defined by StartLimitIntervalSec.

StartLimitIntervalSec=60
→ The time window is 60 seconds. If the service fails and restarts more than 3 times in 60 seconds, systemd will stop trying to restart it automatically.

medyagh (changes requested):

pkg/provision/buildroot.go
```diff
@@ -70,14 +70,16 @@ func (p *BuildrootProvisioner) GenerateDockerOptions(dockerPort int) (*provision
...
+	/* Template is copied from https://github.com/moby/moby/blob/44bca1adf3c806c4ed6688d67c6f995653b09b26/contrib/init/systemd/docker.service#L2 with the following changes:
+	   After: minikube-automount.service is added
+	   ExecStart: additional flags are added
+	*/
 	engineConfigTmpl := `[Unit]
 Description=Docker Application Container Engine
 Documentation=https://docs.docker.com
-After=network.target  minikube-automount.service docker.socket
-Requires= minikube-automount.service docker.socket 
-StartLimitBurst=3
```
medyagh, line 79: @ComradeProgrammer lets keep 
StartLimitBurst=3
StartLimitIntervalSec=60

those are useful it wont let minikube's docker to restart 100 times in 1 second and cause CPU spikes...specially in CI enviroments with limiited resources

medyagh (changes requested):

medyagh:
thank you @ComradeProgrammer I think this is gonna improve the experience of minikube and reduce flakes
I check the Docker_linux Test Duration didnt change still around 47 minutes, so it is probably not affecting performance either

medyagh (approved):

medyagh (approved):

PR #21175, opened 2025-07-28
WIP: Docker service test

ComradeProgrammer:

### DIFF
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -461,6 +461,7 @@ func configureRuntimes(runner cruntime.CommandRunner, cc config.ClusterConfig, k
 		if err != nil {
 			klog.Warningf("cannot ensure containerd is configured properly and reloaded for docker - cluster might be unstable: %v", err)
 		}
+		// wait for containerd socket is ready
 	}
 
 	disableOthers := !driver.BareMetal(cc.Driver)
--- a/pkg/provision/buildroot.go
+++ b/pkg/provision/buildroot.go
@@ -70,14 +70,16 @@ func (p *BuildrootProvisioner) GenerateDockerOptions(dockerPort int) (*provision
 		klog.Infof("root file system type: %s", fstype)
 		noPivot = fstype == "rootfs"
 	}
-
+	/* Template is copied from https://github.com/moby/moby/blob/44bca1adf3c806c4ed6688d67c6f995653b09b26/contrib/init/systemd/docker.service#L2 with the following changes:
+	   After: minikube-automount.service is added
+	   ExecStart: additional flags are added
+	*/
 	engineConfigTmpl := `[Unit]
 Description=Docker Application Container Engine
 Documentation=https://docs.docker.com
-After=network.target  minikube-automount.service docker.socket
-Requires= minikube-automount.service docker.socket 
-StartLimitBurst=3
-StartLimitIntervalSec=60
+After=network-online.target minikube-automount.service nss-lookup.target docker.socket firewalld.service containerd.service time-set.target
+Wants=network-online.target containerd.service
+Requires=docker.socket
 
 [Service]
 Type=notify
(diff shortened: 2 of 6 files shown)


pkg/provision/buildroot.go
```diff
@@ -70,18 +70,21 @@ func (p *BuildrootProvisioner) GenerateDockerOptions(dockerPort int) (*provision
...
+After=network-online.target minikube-automount.service nss-lookup.target docker.socket firewalld.service containerd.service time-set.target
+Wants=network-online.target containerd.service
+Requires=docker.socket
 StartLimitBurst=3
 StartLimitIntervalSec=60
-
 [Service]
 Type=notify
-Restart=on-failure
+Restart=always
```
medyagh, line 87: Restart=always doesnt seem right, what if user uses a different Container Runtime (like CRIO) will it restart it  ? question is, this restart it if it is Disabled? (I think we disable it when other container runtimes)

Restart=always
This setting tells systemd to always restart the service when it stops, regardless of how it was stopped. This includes:
When the service process exits with a clean exit code (typically 0).

Restart=on-failure
This setting tells systemd to only restart the service when it stops because of a failure. This includes:

When the service process exits with a non-zero exit code.

pkg/minikube/node/start.go
```diff
@@ -461,6 +461,7 @@ func configureRuntimes(runner cruntime.CommandRunner, cc config.ClusterConfig, k
 		if err != nil {
 			klog.Warningf("cannot ensure containerd is configured properly and reloaded for docker - cluster might be unstable: %v", err)
 		}
+		// wait for containerd socket is ready
```
medyagh, line 464: did you mean to add a code here?

pkg/provision/ubuntu.go
```diff
@@ -71,20 +71,25 @@ func (p *UbuntuProvisioner) GenerateDockerOptions(dockerPort int) (*provision.Do
...
+After=network-online.target nss-lookup.target docker.socket firewalld.service containerd.service time-set.target
+Wants=network-online.target containerd.service
 Requires=docker.socket
 StartLimitBurst=3
 StartLimitIntervalSec=60
 
 [Service]
 Type=notify
-Restart=on-failure
+Restart=always
```
medyagh, line 92: always seems wrong and dangrous, would this restart it if it is Disabled?

pkg/provision/buildroot.go
medyagh: maybe I am wrong and if you Stop it manually or if you disable it, it is okay and wont cause crazy restarts ? What do you think @nirs @prezha 

only thing I wanna guard against is in the past, we had times Docker would crazily restart and cause CPU hikes,
we already guard that with limiting 3 restarts a minute... but is that a good idea to expect docker to be running always? maybe it is

nirs: I don't understand the semantics of Restart=always and way we need it.

We should go with the default value used in other distros. If you want to change the default the reason should be documented in the config file.

@ComradeProgrammer can you explain this change?

pkg/provision/ubuntu.go
nirs: I think disabling or masking the service will deactivate the unit so the service will not start, but we should not change this unless we have a good reason and without documenting the change in the file.

pkg/provision/buildroot.go
ComradeProgrammer: Restart=always is copied from here https://github.com/moby/moby/blob/28df7815149571b3d3d32e26c32462a03b5b9d17/contrib/init/systemd/docker.service#L19

It seem that this is the default value for the config,

pkg/minikube/node/start.go
ComradeProgrammer: no, it is just a place holder which won't appear in real pr

pkg/provision/buildroot.go
nirs: This seems to be the dafault for 7 years. The change came from https://github.com/moby/moby/pull/42373

I think it is safe to use the default if we don't have any reason to change it.

Daemon typicaly never exit, so always or on-failure should not matter. If you stop the service the daemon should not be restarted since it conflicts with the ability to start and stop services.

pkg/provision/ubuntu.go
nirs: @ComradeProgrammer this came from same source or you check how docker is used on ubuntu?

pkg/provision/buildroot.go
medyagh: ok as long as our DisableOthers code (which disable other comtainer runtimes) can effecitvely stop/disable docker when it is not chosen, that should be fine

medyagh:
@ComradeProgrammer the test look good, should merge this into the main one and merge that ?

PR #21186, opened 2025-07-28
wip: make target Update all

medyagh:
add "make update-all"
and a github workflow to do that

and also rename the leaderboards not to be called "update-leaderboard" so it doesnt get triggered by update 
Proof of work https://github.com/medyagh/minikube/actions/runs/16582700589/job/46901976114

this one is kind of neat, I am trying to make a Github Workflow that would be much cleaner than our previous ISO-ALL workflow that we had to name every single job before/after
https://github.com/kubernetes/minikube/blob/a73390ae025383f1bd5ff2ced89cb4b6ac519557/.github/workflows/update-iso-image-versions.yml#L28


Makefile
```diff
@@ -1244,3 +1244,11 @@ update-kube-vip-version:
 .PHONY: get-dependency-verison
 get-dependency-version:
 	@(cd hack && go run update/get_version/get_version.go)
+
+.PHONY: make-update-all
+update-all: ## Run all update-* targets
+	@echo "Running all update-* targets..."
+	@targets=$(shell grep -Eo ^update-[^:]+ $(MAKEFILE_LIST)); \
+	for t in $$targets; do \
+		$(MAKE) $$t; \
```
nirs: I think we want to make update-all depend on all the update- targets. Running all targets like this does not have the same semantics. For example what happens when one target fail?

I don't think there is magic Makfile syntax to dot hat, just listing all the targets as dependencies.

Also should this update everything or test that all update targets do not fail? If this is a test let's cal this `test-updates`.

Makefile
```diff
@@ -1244,3 +1244,11 @@ update-kube-vip-version:
 .PHONY: get-dependency-verison
 get-dependency-version:
 	@(cd hack && go run update/get_version/get_version.go)
+
+.PHONY: make-update-all
```
nirs: should be same as target name

.github/workflows/update-all.yml
```diff
@@ -0,0 +1,68 @@
...
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5
+        with:
+          go-version: ${{env.GO_VERSION}}
+      - name: Bump all versions
+        id: bumpVersions
+        run: |
+          set -e
```
nirs: -e is the default

.github/workflows/update-all.yml
```diff
@@ -0,0 +1,68 @@
...
+  contents: read
+jobs:
+  update-all:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5
+        with:
+          go-version: ${{env.GO_VERSION}}
+      - name: Bump all versions
```
nirs, line 24: Bumping all version blindly is risky. If something fails, how can you tell which change caused the failure?

.github/workflows/update-all.yml
```diff
@@ -0,0 +1,68 @@
...
+        id: bumpVersions
+        run: |
+          set -e
+          deps=$(grep -oP '^\s*"([^"]+)":' hack/update/get_version/get_version.go | sed -E 's/^\s*"//;s/":$//')
+          for dep in $deps; do
+            var=$(echo $dep | tr 'a-z-' 'A-Z_')
+            old=$(DEP=$dep make get-dependency-version | tail -n1)
+            echo "OLD_${var}=$old" >> "$GITHUB_OUTPUT"
+            eval "OLD_${var}=$old"
```
nirs, line 37: What is the purpose of this loop? Same for the next loops.

.github/workflows/leaderboard.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Update Leaderboard
         id: leaderboard
         run: |
-          make update-leaderboard
+          make leaderboard
```
nirs, line 24: Nice, but not related to update-all. If you want to squash all commits of this pr (since they are not planned), we sill have mix of many unrelated changes.

Makefile
medyagh: Ideally this Target will never be Used in Reality,
this is useful for example
1- when we have 12 things to be updated but we just wanna quickly identify which ones will be triggered
2- check if all all update targets work
3- check right before release to see if we missed any individual updates
once we identify the indvidual updates, we should call that invidual target invidually (instead of looking through github actions and see which ones failed to upate)

so this is mostly for debug reasons.

.github/workflows/update-all.yml
medyagh: I think I should add a comment that this is only meant for debug and should not actually be used, in the github workflow it does NOT run as a schedule but as a At Will (work dispatch) but good call !

.github/workflows/leaderboard.yml
medyagh: just had to do that to Test if it works because it was running update-leaderboard First for some reason and faile don first

Makefile
nirs: So this is mostly a test which we need anyway, and it can also server as tool to get information about the update status of the code.

I think a better way would be a update_test.go that adds a test for every updater, doing basically the same thing:

    go run path/to/update-file.go

The test can do some validations to ensure that the update does the right thing, for example modify the source to use an old version before running the test.

To avoid changes in the current code, the test setup can create a new temporary branch, and the test cleanup can checkout back the original branch and delete the temporary branch.

The test can produce a yaml report for every update for use case 3.

For running all the tests we will go test:

    go test -run TestUpdate

For running specific test:

    go test -run TestUpdate/update-name

.github/workflows/update-all.yml
nirs: Using only workflow_dispatch is nice but we cannot run the tests locally which make it hard to maintain the update programs. We want to be able to run the programs both locally and in CI.

.github/workflows/leaderboard.yml
nirs: Lets split this change so we can merge it quickly.

medyagh: it is kind of related to this PR, because update-all calls all the targets that start with word "update"
and if we dont do that, it will break the "update-all"

.github/workflows/update-all.yml
medyagh: it is actually harder to do it locally, it is better to do it in a standardized env with installed docker and everything it needs, from practical maintainership prespective, it is nice to try once to run all jobs and see which what is missing Right before release and make a Dummy PR to see the diff as well, and it is pretty harmless and reduces a lot of manual tasks for me on my laptop, specially I wont have to remember to add new targets to update-all each time we will be adding a new addon or new update-able thing

.github/workflows/update-all.yml
medyagh: supposed to get the Older version of components before Update
and AFTER update we will check again what is the version,
if before/after versions are different we will print it in the PR description
so ti will be a list of 
bump X version form 1 to 2
bump Y version form 0.2 to 0.3

medyagh:
closing this and re-thinking it

PR #21187, opened 2025-07-28
gomod: remove replace for k8s.io/cluster-bootstrap

medyagh:
remove the replace 
```
k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.22.4
```
and go mod tidy in both root and hack


nirs (commented):
looks good

Issue #21195, opened 2025-07-29
fix "update-amd-gpu-device-plugin-version"


medyagh:
While testing https://github.com/kubernetes/minikube/pull/21186 I found out 
make update-amd-gpu-device-plugin-version
also not working 
```
$ make update-amd-gpu-device-plugin-version
cd hack && go run update/amd_device_plugin_version/update_amd_device_plugin_version.go
I0729 13:20:21.920950   34734 update_amd_device_plugin_version.go:50] Stable version: v1.25.2
F0729 13:20:22.899396   34734 update_amd_device_plugin_version.go:53] failed to get image SHA: failed to pull image: exit status 1: Error response from daemon: manifest for rocm/k8s-device-plugin:v1.25.2 not found: manifest unknown: manifest unknown
goroutine 1 [running]:
```
it tries to pul docker v1.25.2 
https://hub.docker.com/r/rocm/k8s-device-plugin/tags
which is not available but and 1.25.2.8  is availble

-~btw the latest tag is v1.31.07
https://github.com/ROCm/k8s-device-plugin/releases
why it is not trying to fetch that ??~-
never mind that is a pre-release

also could we make the    ghSearchLimit = 300  configurable for Non-Kubernetes Repos ?( we have been hitting the rate limit)

prezha:
@medyagh, good catch! i've also added a [comment](https://github.com/kubernetes/minikube/pull/21196) to your https://github.com/ROCm/k8s-device-plugin/issues/144 issue with an additional remark about discrepancy between the release and image tag (ie, [not] having the `v` prefix) - that's basically a separate issue with the same repo

as for avoiding the github rate limiting, i've created a [pr](https://github.com/kubernetes/minikube/pull/21196) to utilise the GITHUB_TOKEN and increase the quota to 1k req/h
here we're already using the max number of `per_page` (ie, 100) results for the [List repository tags](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repository-tags--parameters) github api endpoint), do we also need to increase the total number of tags to process (the  ghSearchLimit param)? - this would only increase a number of github api calls we're making (and thus more likely hit the rate limit)

nerdeveloper:
Can I look into this?

PR #21189, opened 2025-07-29
ci: use authenticated github when possible

medyagh:
I saw in github actions we hit Github's rate limit in update jobs sometimes:

```
F0728 23:38:04.704151   13024 update_docker_version.go:59] Unable to get docker stable version: GET https://api.github.com/repos/moby/moby/tags?per_page=100: 403 API rate limit exceeded for 172.184.173.240. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) [rate reset in 14m02s]
```
Example https://github.com/medyagh/minikube/actions/runs/16582700589/job/46901976114

I replaced all github.NewClient(nil) to an authenticated one (in hack folder and not minikube itself)

PR #21196, opened 2025-07-29
use GITHUB_TOKEN to avoid github rate-limiting during hack/update runs

prezha:
helps https://github.com/kubernetes/minikube/issues/21195

we noticed we're hitting the github rate limit in our github workflow runs with unauthenticated requests (ie, 60 requests per hour, ref: [Primary rate limit for unauthenticated users](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users))

with this pr, we try to use the auto-generated GITHUB_TOKEN to increase the rate limit to 1,000 requests per hour (ref: [Primary rate limit for GITHUB_TOKEN in GitHub Actions](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-github_token-in-github-actions))

### DIFF
--- a/hack/update/github.go
+++ b/hack/update/github.go
@@ -19,6 +19,7 @@ package update
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 
 	"golang.org/x/mod/semver"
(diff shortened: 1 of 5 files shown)


prezha:
closing as Medya already created a pr https://github.com/kubernetes/minikube/pull/21189 to address this issue

PR #21200, opened 2025-07-30
Improve user feedback for slow docker operations during delete

### LINKED ISSUE #19577: "warn if docker is slow" logic doesnt work

skartikey:
## **Description:**

Fixes #19577

## Problem
When docker daemon is slow or unresponsive, `minikube delete` operations hang indefinitely without providing feedback to users. This leads to confusion where users don't know if minikube is working or stuck.

## Solution
Enhanced docker command execution to provide immediate feedback and prevent infinite hangs:
- Show warning messages after 2-3 seconds when docker commands are slow
- Automatically timeout and terminate hanging processes after 19-30 seconds  
- Provide helpful guidance about restarting docker service
- Enable warning mode for all delete-related container operations

### Before:
```
$ minikube delete --all
🔥  Deleting "test-profile" in docker ...
🔥  Deleting container "test-profile" ...
[hangs indefinitely with no feedback when docker is slow]
```
### After:
```
$ minikube delete --all
🔥  Deleting "test-profile" in docker ...
🔥  Deleting container "test-profile" ...
❗  "docker ps -a --filter label=name.minikube.sigs.k8s.io=test-profile --format {{.Names}}" is taking an unusually long time to respond, please be patient.
💡  If this continues to hang, consider restarting the docker service.
❗  "docker ps -a --filter label=name.minikube.sigs.k8s.io=test-profile --format {{.Names}}" took too long to respond (>30s) and was terminated.
💡  Consider restarting the docker service if this problem persists.
💀  Removed all traces of the "test-profile" cluster.
```

### DIFF
--- a/pkg/drivers/kic/oci/cli_runner.go
+++ b/pkg/drivers/kic/oci/cli_runner.go
@@ -145,23 +145,37 @@ func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
 	killTime := 19 * time.Second // this will be applied only if warnSlow is true
 	warnTime := 2 * time.Second
 
-	if cmd.Args[1] == "volume" || cmd.Args[1] == "ps" { // volume and ps requires more time than inspect
-		killTime = 30 * time.Second
-		warnTime = 3 * time.Second
-	}
-
-	ctx, cancel := context.WithTimeout(context.Background(), killTime)
-	defer cancel()
-
-	if warn { // convert exec.Command to with context
-		cmdWithCtx := exec.CommandContext(ctx, cmd.Args[0], cmd.Args[1:]...)
-		cmdWithCtx.Stdout = cmd.Stdout // copying the original command
-		cmdWithCtx.Stderr = cmd.Stderr
-		cmd = cmdWithCtx
+	// Increase timeout for operations that typically take longer
+	if len(cmd.Args) > 1 {
+		switch cmd.Args[1] {
+		case "volume", "ps":
+			// volume and ps requires more time than inspect
+			killTime = 30 * time.Second
+			warnTime = 3 * time.Second
+		case "rm", "remove":
+			// Container/volume removal can take longer, especially if containers are stuck
+			killTime = 60 * time.Second // CHANGE: Increase from 45 to 60 seconds
+			warnTime = 5 * time.Second
+		case "stop", "kill":
+			// Stop/kill operations can take time if containers are unresponsive
+			killTime = 30 * time.Second
+			warnTime = 3 * time.Second
+		case "network":
+			// Network operations can be slow during cleanup
+			killTime = 30 * time.Second
+			warnTime = 3 * time.Second
+		case "prune":
+			// Prune operations can take a very long time
+			killTime = 120 * time.Second
+			warnTime = 10 * time.Second
+		}
 	}
 
 	rr := &RunResult{Args: cmd.Args}
 	klog.Infof("Run: %v", rr.Command())
+
+	// Determine the actual OCI binary name for user-facing messages
+	actualBinaryName := getActualBinaryName(cmd.Args)
 
 	var outb, errb io.Writer
 	if cmd.Stdout == nil {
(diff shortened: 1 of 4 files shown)


medyagh (commented):
thank you for this PR please take a look at the unit test 
"make test"

  cli_runner_test.go:44: runCmd does not print the correct log, instead print :! "sleep 3" is taking an unusually long time to respond, please be patient.

pkg/drivers/kic/oci/cli_runner.go
```diff
@@ -182,31 +172,84 @@ func runCmd(cmd *exec.Cmd, warnSlow ...bool) (*RunResult, error) {
...
+				warnLock.Lock()
+				_, ok := alreadyWarnedCmds[rr.Command()]
+				if !ok {
+					alreadyWarnedCmds[rr.Command()] = true
+					warnLock.Unlock()
+
+					// Show immediate warning that the operation is slow
+					out.WarningT(`"{{.command}}" is taking an unusually long time to respond, please be patient.`, out.V{"command": rr.Command()})
+					// Don't show any restarting hint, when running podman locally (on linux, with sudo). Only when having a service.
+					if cmd.Args[0] != "sudo" {
```
medyagh: this doesnt seem right, "sudo" could mean a lot of things, doesnt mean podman on linux, this check probably adds more harm than good

skartikey: @medyagh Thanks for the feedback! You're right that hardcoding "sudo" string comparison isn't ideal in general. However, I kept this approach because it's already consistently used throughout the codebase:

1. **Existing pattern in `PrefixCmd()`**: The code already explicitly adds `"sudo"` as a string literal:
   ```go
   cmdWithSudo := exec.Command("sudo", append(append([]string{"-n"}, o.sudoFlags...), cmd.Args...)...)
   ```

2. **Controlled context**: Since minikube controls the command construction through `PrefixCmd()`, we know exactly when and how `"sudo"` is added.

3. **Limited scope**: This helper function only processes commands that minikube itself constructs.

I've added a comment to clarify this:
```go
// If not using sudo, return the first argument
// Note: This checks for "sudo" specifically because PrefixCmd() 
// explicitly adds "sudo" as a string literal when needed
if cmdArgs[0] != "sudo" {
    return cmdArgs[0]
}
```

That said, if you have a preferred alternative approach for identifying the actual OCI binary name when sudo is involved, I'm happy to implement it. Some options could be:
- Using `filepath.Base()` to check just the binary name
- Passing a flag from `PrefixCmd()` indicating if sudo was added
- A different method altogether

What would you prefer?

PR #21206, opened 2025-07-30
download license from github assets

medyagh:
- **remove license binary from git repo**
- **copy licenses to gh assests as well**
- **download licneses from github**
- **try head first to make code more simple**
- **remove debug**


pkg/minikube/download/licenses.go
```diff
@@ -23,28 +23,60 @@ import (
...
 	}
-	f, err := os.CreateTemp("", "licenses")
+
+	tempFile, err := os.CreateTemp("", "licenses-*.tar.gz")
 	if err != nil {
-		return fmt.Errorf("failed to create file in tmp dir: %v", err)
+		return fmt.Errorf("failed to create temporary file: %v", err)
+	}
+	defer os.Remove(tempFile.Name())
+	defer tempFile.Close()
```
prezha: all three defer funcs could return error - i think we should check it and log as warning (ie, not fail) in case of an error

pkg/minikube/download/licenses.go
```diff
@@ -23,28 +23,60 @@ import (
...
+
+	if _, err := io.Copy(tempFile, resp.Body); err != nil {
+		return fmt.Errorf("failed to copy downloaded content: %v", err)
```
prezha:
```suggestion
		return fmt.Errorf("failed to copy downloaded content from %s: %v", url, err)
```

pkg/minikube/download/licenses.go
```diff
@@ -23,28 +23,60 @@ import (
...
+		return fmt.Errorf("download request to %s did not return a 200, received: %d", url, resp.StatusCode)
 	}
-	f, err := os.CreateTemp("", "licenses")
+
+	tempFile, err := os.CreateTemp("", "licenses-*.tar.gz")
 	if err != nil {
-		return fmt.Errorf("failed to create file in tmp dir: %v", err)
+		return fmt.Errorf("failed to create temporary file: %v", err)
+	}
+	defer os.Remove(tempFile.Name())
```
prezha:
```suggestion
	defer func() {
                if err := os.Remove(tempFile.Name()); err != nil {
                        klog.Warningf("Failed to remove temp file %s: %v", tempFile.Name(), err)
                }
        }()
```

pkg/minikube/download/licenses.go
prezha:
```suggestion
	defer func() {
                if err := tempFile.Close(); err != nil {
                        klog.Warningf("Failed to close temp file %s: %v", tempFile.Name(), err)
                }
        }()
```

pkg/minikube/download/licenses.go
```diff
@@ -23,28 +23,60 @@ import (
...
 func Licenses(dir string) error {
-	resp, err := http.Get(fmt.Sprintf("https://storage.googleapis.com/minikube/releases/%s/licenses.tar.gz", version.GetVersion()))
+	url := licensesTarballURL()
+
+	resp, err := http.Get(url)
 	if err != nil {
-		return fmt.Errorf("failed to download licenses: %v", err)
+		return fmt.Errorf("failed to download licenses from %s: %v", url, err)
 	}
 	defer resp.Body.Close()
```
prezha:
```suggestion
	defer func() {
		if err := resp.Body.Close(); err != nil {
			klog.Warningf("Failed to close response body: %v", err)
		}
	}()
```

pkg/minikube/download/licenses.go
prezha: let's also specify which source we used to download the tarball that didn't unpack successfully

prezha (commented):
couple of nits with suggestions for your consideration

prezha (approved):

PR #21212, opened 2025-07-31
CI: remove not needed  replaces in gomod and add automation

medyagh:
- go.mod: remove a few replaces that is no longer needed in 
- add a new target `make gomodtidy` to do go mod tidy both places
- update site contribution documents to remind contributors to use make clean, make gomodtidy and make test and make it more helpful
- add automation gh action to run go mod tidy after every push
- `make html_report`  install gopogh if it is not installed


Makefile
```diff
@@ -1026,6 +1026,12 @@ compare: out/mkcmp out/minikube
 generate-licenses:
 	./hack/generate_licenses.sh
 
+.PHONY: gomodtidy
+gomodtidy: ## run go mod tidy everywhere needed
```
nirs, line 1035: tidy?

Makefile
```diff
@@ -1026,6 +1026,12 @@ compare: out/mkcmp out/minikube
...
+	go mod tidy
+	(cd hack && go mod tidy)
```
nirs: We don't need to a sub shel since every command is using a sub shell.

I tried this:

```
check:
	pwd
	cd hack
	pwd

% make check
pwd
/Users/nsoffer/src/minikube
cd hack
pwd
/Users/nsoffer/src/minikube
``

File: site/content/en/docs/contrib/building/binaries.md [hunk 3]
```diff
@@ -31,9 +31,14 @@ You can also build platform specific executables like below:
     2. `make linux` will build the binary for Linux platform
     3. `make darwin` will build the binary for Darwin/Mac platform
 
-## Compiling minikube using Docker
+## Compiling minikube using
```
nirs: using?

site/content/en/docs/contrib/building/binaries.md
```diff
@@ -31,9 +31,14 @@ You can also build platform specific executables like below:
...
     3. `make darwin` will build the binary for Darwin/Mac platform
 
-## Compiling minikube using Docker
+## Compiling minikube using 
 
-To cross-compile to/from different operating systems:
+M
+
+```shell
+MINIKUBE_BUILD_IN_DOCKER=y make
```
nirs, line 46: Why use docker? you remove docker from the heading.

site/content/en/docs/contrib/building/binaries.md
```diff
@@ -49,6 +54,18 @@ Start the cluster using your built minikube with:
...
 
+## Unit Test and lint
+```shell
+make test
+```
+
+## clean and go mod tidy
+```shell
+make clean
+make gomodtidy
```
nirs, line 74: go mod tidy is not needed when cleaning up. It is needed when adding or removing dependencies.

Make clean is rarely needed (if the build system is correct). go mod tidy should be used before pushing changes to make sure the changes dependencies are updated. Since it is usually not needed running it manually for every push or even for every pr is waste of time. I think the CI should reject PRs that cause requires go mod tidy and users can fix the PR by running go mod tidy.

This is the way we use in ramen:
https://github.com/RamenDR/ramen/blob/main/.github/workflows/ramen.yaml#L56

We combine go mod tidy checks with checking that generated files are not modified which is not great, but the good thing is that PRs cannot be merged with untidy source, and authors can fix this easily.

.github/workflows/go-mod-tidy.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Tidy Go Modules
+        id: gmodtidy
+        run: |
+          make gomodtidy
+          c=$(git status --porcelain)
+          # The following is to support multiline with GITHUB_OUTPUT, see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings
+          echo "changes<<EOF" >> "$GITHUB_OUTPUT"
+          echo "$c" >> "$GITHUB_OUTPUT"
+          echo "EOF" >> "$GITHUB_OUTPUT"
+      - name: Create PR
```
nirs, line 31: So you want to push bad PR that introduces tidyness issues, and create another PR to fix the issues? I don't understand why this is needed. Just reject PRs which are not tidy.

Makefile
```diff
@@ -399,6 +399,11 @@ integration-functional-only: out/minikube$(IS_EXE) ## Trigger only functioanl te
 .PHONY: html_report
 html_report: ## Generate HTML  report out of the last ran integration test logs.
 	@go tool test2json -t < "./out/testout_$(COMMIT_SHORT).txt" > "./out/testout_$(COMMIT_SHORT).json"
+	# install gopogh if not already installed
+	@if ! command -v gopogh >/dev/null 2>&1; then \
+		echo "gopogh not found, installing..."; \
+		GOBIN=$(shell go env GOPATH)/bin go install github.com/medyagh/gopogh/cmd/gopogh@latest; \
```
nirs, line 405: Why do we need to set GOBIN? do you want to install gopogh locally only for the project or in the standard location?

.github/workflows/go-mod-tidy.yml
medyagh: ofcoruse if a PR messes up go.mod tidy on purpose or knowningly, we will catch it, 

this wont create any PR unless there is s change and it is better to use a "standardrized? go mod tidy (in both folders) vs relying your machine or my machine, I think past few weeks you and others created PRs claiming go mod needs to be updated, I reather we just let automation kick of.

again this wont do anything if nothing changes. plus this mo tidys Both Folders that most devs are gonna forget to do 

and historically most ppl forget to do go mod tidy both places and we have not been able to Detect that with our eyes, since everything looks perfectly fine in the PR and User Machine-dependecy of the go.mod is a scary thing

Makefile
medyagh: gopogh's installation instruction is using go install (easiest way to install)
GOPATH/bin is not by default added to the path for may developers, so this just ensures everyone gets right  Gopopgh version and uses that gopogh version (dont have to check where that other one is installed)

Makefile
medyagh: tidy could mean many other tidies could be node, site ,... I think it auto-completes very nicely and intuitively
you just type make gom..  and it auto completes itself, dont have to remember it was called Tidy

Makefile
medyagh: true but for other make Targets we use that (and it is needed there) while I can remove this ( ) I think it is  better to be in same syntax as others (we copy paste Make Targets all the time)

site/content/en/docs/contrib/building/binaries.md
medyagh: this Alternative ways, to group both docker and make cross in same heading

site/content/en/docs/contrib/building/binaries.md
medyagh: sounds good, we have a lint job on every PR, I think we can add soemthing to it to cause it to Fail if go mod changes in the PR

Makefile
medyagh: the install path is 
```/Users/medya/go/bin/gopogh```

some ppl dont have go/bin in their Path (the default) only have GOPATH, it is one way to remove all headaches and jsut give them the HTML_Report

File: Makefile (hunk 2 above)
Comment (medyagh) [reply Makefile]: I will remove this one :) thanks

File: site/content/en/docs/contrib/building/binaries.md (hunk 3 above)
Comment (medyagh) [reply site/content/en/docs/contrib/building/binaries.md]: fixed

File: site/content/en/docs/contrib/building/binaries.md (hunk 4 above)
Comment (medyagh) [reply site/content/en/docs/contrib/building/binaries.md:46]: changed heading

**Comment by prezha** [comment]
we already call `go mod tidy` as part of the [test.sh](https://github.com/kubernetes/minikube/blob/f18c9dc91bd68ab33bb4720fcb5a61511e0375c7/test.sh#L29-L36):
```
    if [[ "$CI" = "true" ]]
    then
        go mod tidy -v && git diff --quiet go.* && echo ok || (((exitcode += 2)) && echo ERROR: Please run go mod tidy)
        echo "= generate docs ========================================================="
        make generate-docs > /dev/null 2>&1 && git diff --quiet site && echo ok || (((exitcode += 3)) && echo ERROR: Please run make generate-docs)
    else
        go mod tidy -v && echo ok || ((exitcode += 2))
    fi
```
which is called by [make test](https://github.com/kubernetes/minikube/blob/f18c9dc91bd68ab33bb4720fcb5a61511e0375c7/Makefile#L414-L416) target, that is, in turn, used both locally and in ci:
- build.yml: [lint](https://github.com/kubernetes/minikube/blob/f18c9dc91bd68ab33bb4720fcb5a61511e0375c7/.github/workflows/build.yml#L62) and [unit_test](https://github.com/kubernetes/minikube/blob/f18c9dc91bd68ab33bb4720fcb5a61511e0375c7/.github/workflows/build.yml#L80)
- master.yml: [lint](https://github.com/kubernetes/minikube/blob/f18c9dc91bd68ab33bb4720fcb5a61511e0375c7/.github/workflows/master.yml#L78) and [unit_test](https://github.com/kubernetes/minikube/blob/f18c9dc91bd68ab33bb4720fcb5a61511e0375c7/.github/workflows/master.yml#L96)
- pr.yml: [lint](https://github.com/kubernetes/minikube/blob/f18c9dc91bd68ab33bb4720fcb5a61511e0375c7/.github/workflows/pr.yml#L76) and [unit_test](https://github.com/kubernetes/minikube/blob/f18c9dc91bd68ab33bb4720fcb5a61511e0375c7/.github/workflows/pr.yml#L94)
- translations.yml: [unit_test](https://github.com/kubernetes/minikube/blob/f18c9dc91bd68ab33bb4720fcb5a61511e0375c7/.github/workflows/translations.yml#L30)

we could simplify by just adapting the existing logic to also cover the hack/go.mod and also utilise the [go mod tidy `-diff`](https://github.com/golang/go/issues/27005) flag added last year:
```
$ go help mod tidy
...
The -diff flag causes tidy not to modify go.mod or go.sum but
instead print the necessary changes as a unified diff. It exits
with a non-zero code if the diff is not empty.
...
```
eg - replace the existing block in test.sh above and check for both `go mod tidy` and `make generate-docs` in both ci and local (as people, including myself, sometimes forget to run, only to fail in ci later on, which is an "expensive" waste of time):
```
    go mod tidy -diff > /dev/null 2>&1 && echo ok || (((exitcode += 2)) && echo ERROR: Please run go mod tidy)
    (cd hack && go mod tidy -diff > /dev/null 2>&1 && echo ok || (((exitcode += 2)) && echo ERROR: Please run go mod tidy in hack/))
    echo "= generate docs ========================================================="
    make generate-docs > /dev/null 2>&1 && git diff --quiet site && echo ok || (((exitcode += 3)) && echo ERROR: Please run make generate-docs)
```

ComradeProgrammer (approved):

PR #21213, opened 2025-07-31
ci: update all with go

medyagh:
- "make _update-all" to run all update targets with go
- standard name for all update-all (to be automated)
- update the gh workflow files to use the new standard path
- Identify update components to the ones that support Before/After 
- Identify "internal tools" currently living under update folder 
- identify not working updates and put in a black list

Tested on my Mac and works

TODO:
Follow Up PR will make get rid of all Make-Update Targets and make all of them use this new script
and there will be a central file with all update components and their types


hack/update/update_all/update_all.go
```diff
@@ -0,0 +1,94 @@
...
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+func getVersion(component string) (string, error) {
+	cmd := exec.Command("go", "run", "update/get_version/get_version.go")
```
nirs, line 61: Can we move the get version code to the update/version.go and consume it without running the get_version program?

hack/update/update_all/update_all.go
```diff
@@ -0,0 +1,94 @@
...
+	err := cmd.Run()
+	if err != nil {
+		log.Printf("failed to get version for %s: %v", component, err)
+		return "", err
+	}
+	return strings.TrimSpace(out.String()), nil
+}
+
+func main() {
+	updateDir := "update"
```
nirs: We assume that the current working directory is hack. It would be useful to change the directory when we start so it does not matter when we run the command from.

hack/update/update_all/update_all.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		log.Fatalf("failed to read directory %s: %v", updateDir, err)
+	}
+
+	var changes []string
+	i := 0
+	for _, d := range dirs {
+		i++
+		if i > 3 {
+			log.Println("debug - skipping more than 3 components")
```
nirs: Why do we need this?

hack/update/update_all/update_all.go
```diff
@@ -0,0 +1,94 @@
...
+			break
+		}
+		if !d.IsDir() {
+			continue
```
nirs, line 96: So we assume that all directories under update/ are update programs?

hack/update/update_all/update_all.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		component := d.Name()
+		if component == "get_version" || component == "update_all" || component == "k8s-lib" || component == "amd-gpu-device-plugin-version" {
+			continue
```
nirs, line 101: We probably need to keep components under `update/components/name` so we don't need to filter helper programs which are not update programs like get_version or this one.

It will help to comment why we need to skip the other components (k8s-libs, amd-gpu-device-plugin-version)

hack/update/update_all/update_all.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		fmt.Printf("Processing %s...\n", component)
+
+		oldVersion, err := getVersion(component)
+		if err != nil {
+			log.Fatalf("Could not get old version for %s: %v", component, err)
```
nirs: It wil leb more useful to log an error, update results lists and continue. This way we can report all the status at the end.

hack/update/update_all/update_all.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		updateCmd := exec.Command("go", "run", filepath.Join(updateDir, component, fmt.Sprintf("%s.go", component)))
```
nirs: Can we simplify and use main.go for all programs?

hack/update/update_all/update_all.go
```diff
@@ -0,0 +1,94 @@
...
+		} else {
+			fmt.Printf("No change for %s.\n", component)
+		}
+		fmt.Println()
+	}
+
+	fmt.Println("---")
+	fmt.Printf("Updated components summary:\n%s\n", strings.Join(changes, "\n"))
+
+	// Print a machine-readable summary for GitHub Actions
```
nirs: This is not very machine friendly. Why not generate json or yaml?

```go
type Update structure {
    Name string
    Updated bool
    From string
    To string
}
```

Makefile
```diff
@@ -1247,5 +1247,5 @@ get-dependency-version:
 
 # runs update on all hack/update/components only used for debugging purposes, not meant to be used regularly
 .PHONY: _update-all
-update-all:
+_update-all:
```
nirs, line 1249: It is very hard to review in this form. Squashing all the temporary fixes and splitting logic changes to one change per commit will make it much easier to review.

hack/update/get_version/get_version.go
```diff
@@ -101,3 +103,12 @@ func main() {
 	}
 	os.Stdout.Write(submatches[1])
 }
+
+// some components have _ or - in their names vs their make folders, standarizing for automation such as as update-all
+func standrizeComponentName(name string) string {
+	// Convert the component name to lowercase and replace underscores with hyphens
+	name = strings.ToLower(name)
+	name = strings.ReplaceAll(name, "_", "-")
+	name = strings.ReplaceAll(name, "-version", "")
```
nirs, line 112: Much nicer this way!

Can we rename the components directories and files instead of fixing the names?

hack/update/amd-gpu-device-plugin-version/amd-gpu-device-plugin-version.go
nirs, line 1: This was a shell script before? Why we don't delete the old version?

hack/update/buildkit_version/buildkit_version.go
nirs, line 1: Can we use `hack/update/components/buildkit/main.go` ?

It avoids the issues of filtering get_version and update_all, remove the repeated "_version" and remove the repeated component name.

hack/update/amd-gpu-device-plugin-version/amd_gpu_device_plugin_version.go
nirs, line 1: Is this the old version that the previous commit replaced? Why we did not modify the existing file?

Makefile
```diff
@@ -1073,172 +1073,172 @@ time-to-k8s-benchmark:
 
 .PHONY: update-gopogh-version
 update-gopogh-version: ## update gopogh version
-	cd hack && go run update/gopogh_version/update_gopogh_version.go
+	cd hack && go run update/gopogh_version/gopogh_version.go
```
nirs, line 1074: Much better like this!

Makefile
```diff
@@ -1203,9 +1203,10 @@ update-kong-ingress-controller-version:
 update-nvidia-device-plugin-version:
 	cd hack && go run update/nvidia_device_plugin_version/nvidia_device_plugin_version.go
 
-.PHONY: update-amd-gpu-device-plugin-version
-update-amd-gpu-device-plugin-version:
-	cd hack && go run update/amd-gpu-device-plugin-version/amd_gpu_device_plugin_version.go
+# for amd gpu 
+.PHONY: update-amd-gpu-device-plugin-version 
+update-amd-device-plugin-version:
```
nirs, line 1206: .PHONY and target are not the same

hack/update/update_all/update_all.go
```diff
@@ -65,8 +65,7 @@ func main() {
 		updateCmd.Stderr = os.Stderr
 
 		if err := updateCmd.Run(); err != nil {
-			log.Printf("Failed to update %s: %v", component, err)
-			continue
+			log.Fatalf("Failed to update %s: %v", component, err)
```
nirs, line 120: This stop on the first error, which is less useful for reporting multiple updates.

hack/update/cilium_version/cilium_version.go
```diff
@@ -45,14 +45,20 @@ func main() {
 	version := strings.TrimPrefix(stable.Tag, "v")
 
 	// Add the cilium repo to helm
-	if err := exec.Command("helm", "repo", "add", "cilium", "https://helm.cilium.io/").Run(); err != nil {
-		klog.Fatal(err)
+	addRepoCmd := exec.Command("helm", "repo", "add", "cilium", "https://helm.cilium.io/")
+	if err := addRepoCmd.Run(); err != nil {
+		klog.Fatalf("failed to run command %v: %v", addRepoCmd.Args, err)
+	}
```
nirs, line 51: Why is this better? We could use Fatalf() without adding a variable for the command.

Since we run lot of commands in the same way, it will help to use a common Run() function so we can have:

```go
if err := Run("helm", "repo", "add", "cilium", "https://helm.cilium.io/"); err != nil {
    handle it ...
}
```

This ensures that we run everything with the same way (stdout, stderr, env). It also can return better errors - the error we get from command.Run() is usually not helpful (command exited with error code 1). We typically want to get the ExitError.Stderr to report meaningful errors.

hack/update/cilium_version/cilium_version.go
```diff
@@ -45,14 +45,20 @@ func main() {
...
+
+	if err := exec.Command("helm", "repo", "update").Run(); err != nil {
+		klog.Fatalf("failed to update help repo %v", err)
 	}
 
 	// Generate the cilium YAML
-	yamlBytes, err := exec.Command("helm", "template", "cilium", "cilium/cilium", "--version", version, "--namespace", "kube-system").Output()
+	templateCmd := exec.Command("helm", "template", "cilium", "cilium/cilium", "--version", version, "--namespace", "kube-system")
+	yamlBytes, err := templateCmd.Output()
```
nirs, line 59: For long commands with many flags it is easier to work with line line per argument:

```go
if err := exec.Command(
    "helm",
    "template",
    "cilium",
    "cilium/cilium",
    "--version", version,
    "--namespace", "kube-system",
); err != nil {
    ...
}
```

hack/update/cri_o_version/cri_o_version.go
nirs, line 1: Forgot to delete the previous file?

hack/update/update_all/update_all.go
```diff
@@ -38,10 +38,11 @@ func main() {
 
 	var changes []string
 	i := 0
+	max := 100
```
nirs: Should be a constant but I don't think we need it.

Instead we should have a summary at the end like:

    updated: 42, skipped: 7, failed: 1

hack/update/update_all/update_all.go
```diff
@@ -38,10 +38,11 @@ func main() {
...
 	for _, d := range dirs {
 		i++
-		if i > 3 {
-			log.Println("debug - skipping more than 3 components")
+		if i > max {
+			log.Println("debug - skipping more than 100 components")
```
nirs: why not format max into the message?

hack/update/update_all/update_all.go
```diff
@@ -50,7 +51,14 @@ func main() {
 		}
 
 		component := d.Name()
-		if component == "get_version" || component == "update_all" || component == "k8s-lib" || component == "amd_device_gpu_plugin_version" {
+		blackList := map[string]bool{
```
nirs: Few years ago the entire world replaced terms like this. Why introduce them again?

The main issue with blacklist is that it does not describe well the purpose. It would be more helpful to call this "skip", so we can have code like:

```go
if skip[component] {
    continue
}

File: hack/update/update.go [hunk 19]
```diff
@@ -93,11 +93,13 @@ func Apply(schema map[string]Item, data interface{}) {
 	changed, err := fsUpdate(FSRoot, schema, data)
 	if err != nil {
 		klog.Errorf("Unable to update local repo: %v", err)
+		return err
```
nirs, line 96: If we return an error, why log the error here? This will lead to double logging for the same error which makes it much harder to debug issues.

We can do:

    return fmt.Errorf("Unable to update local repo: %w", err")

hack/update/get_version/get_version.go
```diff
@@ -50,7 +50,7 @@ var dependencies = map[string]dependency{
 	"flannel":                 {"pkg/minikube/cni/flannel.yaml", `flannel:(.*)`},
 	"gcp-auth":                {addonsFile, `k8s-minikube/gcp-auth-webhook:(.*)@`},
 	"gh":                      {"hack/jenkins/installers/check_install_gh.sh", `GH_VERSION="(.*)"`},
-	"go":                      {"Makefile", `\nGO_VERSION \?= (.*)`},
+	"golang":                  {"Makefile", `\nGO_VERSION \?= (.*)`},
```
nirs, line 53: Later we ned to get the version from go.mod:

```console
% go list -m -f '{{.GoVersion}}'   
1.24.0
```

hack/update/update_all/update_all.go
```diff
@@ -81,20 +90,22 @@ func main() {
 			log.Fatalf("Failed to update %s: %v", component, err)
 		}
 
-		newVersion, err := getVersion(component)
-		if err != nil {
-			log.Printf("Could not get new version for %s: %v", component, err)
-			continue
+		if !notSupportBeforeAfterVersion[component] {
```
nirs: This is confusing name (double negative)

Makefile
```diff
@@ -35,12 +35,12 @@ RPM_REVISION ?= 0
...
 # used by hack/jenkins/release_build_and_upload.sh and KVM_BUILD_IMAGE, see also BUILD_IMAGE below
 # update this only by running `make update-golang-version`
-GO_VERSION ?= 1.24.0
+GO_VERSION ?= 1.24.5
 # set GOTOOLCHAIN to GO_VERSION to override any toolchain version specified in
 # go.mod (ref: https://go.dev/doc/toolchain#GOTOOLCHAIN)
 export GOTOOLCHAIN := go$(GO_VERSION)
 # update this only by running `make update-golang-version`
-GO_K8S_VERSION_PREFIX ?= v1.33.0
+GO_K8S_VERSION_PREFIX ?= v1.34.0
```
nirs, line 43: Is this intended?

nirs (commented):
I reviewed 26 commits. I'll look again when the fixup commits are squashed.

hack/update/update_all/update_all.go
medyagh: This can be considered in a follow up PR, this one is already too big

hack/update/update_all/update_all.go
medyagh: this was old commits debug, need to squash those commits

hack/update/update_all/update_all.go
medyagh: in this iteration yes, (we have skip list) in the next iteration they will be in a struct

hack/update/update_all/update_all.go
medyagh: agree, commented in the other list, this will be refactored in a follow up PR

Makefile
medyagh: was old ocmmit

hack/update/update_all/update_all.go
medyagh: for now that is more practical, with tons components, it will produce too much to consume (including rate limits), I rather fix one component at a time.

PR #21214, opened 2025-08-01
ci: skip boilerplate for git submodules

medyagh:
## Before this PR (would try to add boiler plate to submodules)
```

$ git s
On branch master
Your branch is up to date with 'upstream/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
  (commit or discard the untracked or modified content in submodules)
        modified:   hack/benchmark/image-build/minikube-image-benchmark (modified content)
        modified:   hack/benchmark/time-to-k8s/time-to-k8s-repo (modified content)

```
## after this PR
skip git submodules listed in in our .gitmouldes

### DIFF
--- a/hack/boilerplate/boilerplate.go
+++ b/hack/boilerplate/boilerplate.go
@@ -31,7 +31,7 @@ var (
 	boilerplatedir = flag.String("boilerplate-dir", ".", "Boilerplate directory for boilerplate files")
 	rootdir        = flag.String("rootdir", "../../", "Root directory to examine")
 	verbose        = flag.Bool("v", false, "Verbose")
-	skippedPaths   = regexp.MustCompile(`Godeps|third_party|_gopath|_output|\.git|cluster/env.sh|vendor|test/e2e/generated/bindata.go|site/themes/docsy|test/integration/testdata`)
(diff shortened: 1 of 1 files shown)


prezha:
with this change we will modify the skippedPaths even if flag is set, by always appending all git submodules?

we already have "site/themes/docsy" git submodule in the [skippedPaths](https://github.com/medyagh/minikube/blob/45e708b6e26f69212be7937eae299b26e9cc0764/hack/boilerplate/boilerplate.go#L34) flag's default value, which is the reason we don't see it in the "Before this PR" example you gave above - would we just add those remaining two to the default skippedPaths value?

medyagh:
> with this change we will modify the skippedPaths even if flag is set, by always appending all git submodules?
> 
> we already have "site/themes/docsy" git submodule in the [skippedPaths](https://github.com/medyagh/minikube/blob/45e708b6e26f69212be7937eae299b26e9cc0764/hack/boilerplate/boilerplate.go#L34) flag's default value, which is the reason we don't see it in the "Before this PR" example you gave above - would we just add those remaining two to the default skippedPaths value?

thats a good point ! I was trying to future proof it for all possible coming git modules, but maybe better it is just add another folder there

Issue #21148, opened 2025-07-25
fix spinner


medyagh:
since this PR https://github.com/kubernetes/minikube/pull/21115

the Sub Steps that shows kubernetes Creating Shows for 1 miliseconds and vanishes

medyagh:
@bobsira also mentioned issues with windows https://kubernetes.slack.com/archives/C1F5CT6Q1/p1753714579472059?thread_ts=1753137617.531989&cid=C1F5CT6Q1

PR #21215, opened 2025-08-01
Refactor spinner library & hide sub steps after spinning

medyagh:
This PR fixes the broken spinner 
## Before
Some Spining steps were over-written and didnt show
spinner-before

## After 
mini-spinner

###  New Feature: Hide Sub Steps (details after showing them once during spinning)

and also After Showing the Sub Steps for 
🐳  Preparing Kubernetes v1.33.2 on Docker 28.3.2 ...
it will hide the following after Showing them Once :
```
    ▪ Generating certificates and keys ...(hide after showing)
    ▪ Booting up control plane ...(hide after showing)
    ▪ Configuring RBAC rules ...(hide after showing)
```

###  Enable Spinning on a second start

Before this PR minikube second start would not spin on "updating the running VM"

### refactor code
did some refactor such as rename boolean "spinner" to say "shoudSpin" since we already had a spinner object ! the could should be more readable now 

closes http://github.com/kubernetes/minikube/issues/21148

### Ideas for Follow up Todos
now that we have a Hiding Sub Step Mechanisim, we can break down the "🔥  Creating krunkit VM " Step to sub steps and Show and Hide while it is creaiting

### DIFF
--- a/pkg/minikube/out/out.go
+++ b/pkg/minikube/out/out.go
@@ -91,7 +91,7 @@ func Step(st style.Enum, format string, a ...V) {
 		Infof(format, a...)
 		return
 	}
-	outStyled, _ := stylized(st, useColor, format, a...)
+	outStyled, _, _ := stylized(st, useColor, format, a...)
 	if JSON {
 		register.PrintStep(outStyled)
 		klog.Info(outStyled)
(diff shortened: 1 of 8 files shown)


medyagh:
mini

nirs:
Hiding sub steps is confusing, and using progress bar for indeterminte progress is wrong. When we cannot tell the progress value we should use a spinner and not animated progress bar going forward and backward. I think Windows got this wrong in some versions bug we should not copy bad UI.

Can we separate the spinner fix without changing the behavior? Just restore the behavior of minikube 1.36. Do we have any issues about the previous behavior?

nirs (commented):
Added comment

medyagh:
> Hiding sub steps is confusing, and using progress bar for indeterminte progress is wrong. When we cannot tell the progress value we should use a spinner and not animated progress bar going forward and backward. I think Windows got this wrong in some versions bug we should not copy bad UI.
> 
> Can we separate the spinner fix without changing the behavior? Just restore the behavior of minikube 1.36. Do we have any issues about the previous behavior?

changed progress bar to simpler spining, the sub steps in original design were always meant to be hidden, we had never made it work because were not using the lib

PR #21250, opened 2025-08-05
start: Simpler and safer mount on start

nirs:
The `--mount-string` argument defaults to `/Users` on darwin, and homedir.Homedir() on other platforms (e.g. `$HOME` on unix).

This is wrong in many ways:

- `/Users` is not `$HOME` on darwin (the right path is `/Users/$USER`). Using the default mount we cannot access anything inside the guest in the user home directory.  We can access the special `/Users/Shared` directory, but this should not be a default mount.

- Mounting the user home directory inside the guest in read-write mode is a horrible default. This exposes the users private keys in .ssh/ to the guest, any sensitive files in the user home directory, and allows the guest to change any file on the host.

- Using the `--mount` option mounts the default mount directory silently. This is unexpected, surprising, and not documented in the minikube handbook[1].

Example access to user private key from the guest with the default mount:

    $ minikube start --mount

    $ minikube ssh cat /minikube-host/.ssh/id_ed25519
    -----BEGIN OPENSSH PRIVATE KEY-----
    ...
    -----END OPENSSH PRIVATE KEY-----

Fixed by removing the default mount directory and changing mount logic to check for non-empty mount-string instead of the mount flag.

The mount flag is kept for backward compatibility, but its value is ignored. In the next release we want to use this flag for supporting multiple mounts.

Example usage before:

    minikube start --mount --mount-string ~/models:/mnt/models

Example usage after:

    minikube start --mount-string ~/models:/mnt/models

Breaking changes:

User depending the default mount will have to replace the command:

    minikube start --mount

With:

    minikube start  --mount-string $HOME:/minikube-host

[1] https://minikube.sigs.k8s.io/docs/handbook/mount/


pkg/minikube/constants/constants_darwin.go
```diff
@@ -1,22 +0,0 @@
-//go:build (darwin && ignore) || !gendocs
```
medyagh, line 1: Why are these files deleted ?

medyagh:
I agree, we should not mount the home folder to ppl's minikubes by default that was a horrible default of some drivers. I am okay with breaking change if we can get rid of it by default

pkg/minikube/constants/constants_darwin.go
nirs: The only constant in this file was the one I removed.

nirs:
Changes in latest version:
- Fix TestMountStart to use --mount-string instead of --mount
- Rebase on master

cmd/minikube/cmd/start_flags.go
```diff
@@ -173,8 +173,8 @@ func initMinikubeFlags() {
 	startCmd.Flags().Bool(keepContext, false, "This will keep the existing kubectl context and will create a minikube context.")
 	startCmd.Flags().Bool(embedCerts, false, "if true, will embed the certs in kubeconfig.")
 	startCmd.Flags().StringP(containerRuntime, "c", constants.DefaultContainerRuntime, fmt.Sprintf("The container runtime to be used. Valid options: %s (default: auto)", strings.Join(cruntime.ValidRuntimes(), ", ")))
-	startCmd.Flags().Bool(createMount, false, "This will start the mount daemon and automatically mount files into minikube.")
-	startCmd.Flags().String(mountString, constants.DefaultMountDir+":/minikube-host", "The argument to pass the minikube mount command on start.")
+	startCmd.Flags().Bool(createMount, false, "Kept for backward compatibility, value is ignored.")
```
medyagh, line 176: we can consider hiding this flag (we have a few flags that hidden since we dont want ppl to use them, but left for backward campatiblitiy

medyagh (commented):
Question: does this mean we need to update this site page about "built-in host folder sharing"?
Driver mounts
Some hypervisors, have built-in host folder sharing. Driver mounts are reliable with good performance, but the paths are not predictable across operating systems or hypervis


https://minikube.sigs.k8s.io/docs/handbook/mount/

nirs:
> Question: does this mean we need to update this site page about "built-in host folder sharing"? Driver mounts Some hypervisors, have built-in host folder sharing. Driver mounts are reliable with good performance, but the paths are not predictable across operating systems or hypervis
> 
> https://minikube.sigs.k8s.io/docs/handbook/mount/

This change does not effect built in mounts - these are not done using 9p so they should not be affected by ignoring the --mount flag. These should be disable by --disable-driver-mounts which we did not touch.

So I don't think any update is needed, the only change is the detail of using --mount flag which is not documented in the handbook.

test/integration/mount_start_test.go
```diff
@@ -87,13 +93,23 @@ func TestMountStart(t *testing.T) {
 }
 
 // validateStartWithMount starts a cluster with mount enabled
-func validateStartWithMount(ctx context.Context, t *testing.T, profile string) {
+func validateStartWithMount(ctx context.Context, t *testing.T, profile string, hostPath string) {
```
medyagh, line 96: not a problem in this PR but wanted to let you know 
functions that start with vadliate...in integraiton test will be auto-published to the site with their comment, to show what things are being tests in minikube
https://minikube.sigs.k8s.io/docs/contrib/tests/

cmd/minikube/cmd/start_flags.go
nirs: Hiding the flag will hide the useful help message. I think we should keep if for now. If we don't use it in the next releases we can hide it.

nirs:
Rebased to consume the docker service fix

cmd/minikube/cmd/start_flags.go
medyagh, line 176: help text, this will be depricated and merged with --mount-string in the next version

nirs: Opened #21291, I'll try to get this into 1.37.

PR #21266, opened 2025-08-08
drivers: Extract drivers/common package

nirs:
Move all files and packages which are not a driver to the drivers/common package. This helps to understand the structure of the code.

While moving, move the iso_test.iso into testdata/test.iso to separate code and test data.

While fixing the test iso path fix few bugs in the tests using wrong iso path. The test failed because the iso path was wrong instead of issue with source or destination path.

New directory structure:

    % tree -L1 pkg/drivers
    pkg/drivers
    ├── common
    ├── hyperkit
    ├── kic
    ├── krunkit
    ├── kvm
    ├── none
    ├── qemu
    ├── ssh
    └── vfkit


medyagh:
@nirs nice! I like that, should I merge this after the other ones ?

nirs:
> @nirs nice! I like that, should I merge this after the other ones ?

Before, I need this for #21149

PR #21274, opened 2025-08-09
ci: add qemu functional test to github action on macos13

medyagh:
- adding the qemu functional test to minikube github action intel -macos13

I am hoping to make a follow up PR to use matrixes for these jobs (would be a bigger refactor)

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -415,7 +415,121 @@ jobs:
           if [ "$numPass" -eq 0 ];then echo "*** 0 Passed! ***";exit 2;fi
           if [ "$numPass" -lt 36 ];then echo "*** Failed to pass at least 36 ! ***";exit 2;fi
           if [ "$numPass" -eq 0 ];then echo "*** Passed! ***";exit 0;fi
-
+  functional_qemu_macos13_intel:
+    permissions:
+      contents: none
+    needs: [build_minikube_test_binaries]
+    env:
+      TIME_ELAPSED: time
+      JOB_NAME: "functional_qemu_macos13_intel"
+      GOPOGH_RESULT: ""
+      SHELL: "/bin/bash" # To prevent https://github.com/kubernetes/minikube/issues/6643
+    runs-on: macos-13
+    steps:
+      - name: Install kubectl and docker-cli
+        shell: bash
+        run: |
+          brew install kubectl
+          brew install docker
+          kubectl version --client=true
+      - name: Install Qemu and socket_vmnet
+        shell: bash
+        run: |
+          brew update
+          brew install qemu socket_vmnet
+          HOMEBREW=$(which brew) && sudo ${HOMEBREW} services start socket_vmnet
+          fw=/usr/libexec/ApplicationFirewall/socketfilterfw
+          sudo $fw --remove /usr/libexec/bootpd
+          sudo $fw --add /usr/libexec/bootpd
+          sudo $fw --unblock /usr/libexec/bootpd
+      - name: Info Block
+        shell: bash
+        run: |
+          set -x
+          uname -a
+          sysctl -n hw.memsize
+          echo "$(sysctl -n hw.memsize) / 1024 / 1024 / 1024" | bc
+          sysctl -n hw.ncpu
+          sysctl -n machdep.cpu.brand_string
+          sysctl hw.model
+          sysctl -n kern.hv_vmm_present
+          sysctl -n kern.hv_support
+          system_profiler SPHardwareDataType
+          docker version || true
+          docker info || true
+          echo "macOS version:"
+          sw_vers
+          ifconfig
+      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5
+        with:
+          cache: true
+          go-version: ${{env.GO_VERSION}}
+      - name: Install gopogh
+        shell: bash
+        run: |
+          go install github.com/medyagh/gopogh/cmd/gopogh@v0.29.0
+      - name: Download Binaries
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093
+        with:
+          name: minikube_binaries
+          path: minikube_binaries
+      - name: Run Integration Test
+        continue-on-error: true
+        # bash {0} to allow test to continue to next step. in case of
+        shell: bash {0}
+        run: |
+          set -x
+          pwd
+          cd minikube_binaries
+          chmod a+x e2e-*
+          chmod a+x minikube-*
+          mkdir -p report
+          # test home
+          mkdir -p /tmp/th
+          MINIKUBE_HOME=/tmp/th ./minikube-darwin-amd64 delete --all --purge
+          START_TIME=$(date -u +%s)
+          KUBECONFIG=$/tmp/th/kubeconfig MINIKUBE_HOME=/tmp/th ./e2e-darwin-amd64 -minikube-start-args="--vm-driver=qemu --wait-timeout=15m --network socket_vmnet"  -test.run TestFunctional -test.timeout=40m -test.v -timeout-multiplier=1.5 -binary=./minikube-darwin-amd64 2>&1 | tee ./report/testout.txt
+          END_TIME=$(date -u +%s)
+          TIME_ELAPSED=$(($END_TIME-$START_TIME))
+          min=$((${TIME_ELAPSED}/60))
+          sec=$((${TIME_ELAPSED}%60))
+          TIME_ELAPSED="${min} min $sec seconds "
+          echo "TIME_ELAPSED=${TIME_ELAPSED}" >> $GITHUB_ENV
+      - name: Generate HTML Report
+        shell: bash
+        run: |
+          cd minikube_binaries
+          export PATH=${PATH}:`go env GOPATH`/bin
+          go tool test2json -t < ./report/testout.txt > ./report/testout.json || true
+          STAT=$(gopogh -in ./report/testout.json -out_html ./report/testout.html -out_summary ./report/testout_summary.json -name "${JOB_NAME} ${GITHUB_REF}" -repo "${GITHUB_REPOSITORY}"  -details "${GITHUB_SHA}")  || true
+          echo status: ${STAT}
+          FailNum=$(echo $STAT | jq '.NumberOfFail')
+          TestsNum=$(echo $STAT | jq '.NumberOfTests')
+          GOPOGH_RESULT="${JOB_NAME} : completed with ${FailNum} / ${TestsNum} failures in ${TIME_ELAPSED}"
+          echo "GOPOGH_RESULT=${GOPOGH_RESULT}" >> $GITHUB_ENV
+          echo 'STAT<<EOF' >> $GITHUB_ENV
+          echo "${STAT}" >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/master.yml
```diff
@@ -415,7 +415,121 @@ jobs:
...
+      JOB_NAME: "functional_qemu_macos13_intel"
+      GOPOGH_RESULT: ""
+      SHELL: "/bin/bash" # To prevent https://github.com/kubernetes/minikube/issues/6643
+    runs-on: macos-13
+    steps:
+      - name: Install kubectl and docker-cli
+        shell: bash
+        run: |
+          brew install kubectl
+          brew install docker
```
nirs, line 433: Better have one brew command for both and add a comment why we need docker.

.github/workflows/master.yml
```diff
@@ -415,7 +415,121 @@ jobs:
...
+      contents: none
+    needs: [build_minikube_test_binaries]
+    env:
+      TIME_ELAPSED: time
+      JOB_NAME: "functional_qemu_macos13_intel"
+      GOPOGH_RESULT: ""
+      SHELL: "/bin/bash" # To prevent https://github.com/kubernetes/minikube/issues/6643
+    runs-on: macos-13
+    steps:
+      - name: Install kubectl and docker-cli
```
nirs, line 429: We will have to maintain this line if we change the content, better have a generic title that does not need maintenance like "Install required packages"

.github/workflows/master.yml
```diff
@@ -415,7 +415,121 @@ jobs:
...
-
+  functional_qemu_macos13_intel:
+    permissions:
+      contents: none
+    needs: [build_minikube_test_binaries]
+    env:
+      TIME_ELAPSED: time
+      JOB_NAME: "functional_qemu_macos13_intel"
+      GOPOGH_RESULT: ""
+      SHELL: "/bin/bash" # To prevent https://github.com/kubernetes/minikube/issues/6643
```
nirs, line 426: Not clear to me why this is needed, I looked at the issue but I don't understand it. Smells like issue in the code we test that should be fixed there instead of complicating the workflow. But if we already use this in the current workflow we can continue to use it.

.github/workflows/master.yml
```diff
@@ -415,7 +415,121 @@ jobs:
...
           if [ "$numPass" -lt 36 ];then echo "*** Failed to pass at least 36 ! ***";exit 2;fi
           if [ "$numPass" -eq 0 ];then echo "*** Passed! ***";exit 0;fi
-
+  functional_qemu_macos13_intel:
+    permissions:
+      contents: none
+    needs: [build_minikube_test_binaries]
+    env:
+      TIME_ELAPSED: time
+      JOB_NAME: "functional_qemu_macos13_intel"
```
nirs, line 424: Seems unnecessary, same as the job name.

.github/workflows/master.yml
```diff
@@ -415,7 +415,121 @@ jobs:
           if [ "$numPass" -eq 0 ];then echo "*** 0 Passed! ***";exit 2;fi
           if [ "$numPass" -lt 36 ];then echo "*** Failed to pass at least 36 ! ***";exit 2;fi
           if [ "$numPass" -eq 0 ];then echo "*** Passed! ***";exit 0;fi
-
+  functional_qemu_macos13_intel:
+    permissions:
+      contents: none
+    needs: [build_minikube_test_binaries]
+    env:
+      TIME_ELAPSED: time
```
nirs, line 423: Why this is needed?

.github/workflows/master.yml
```diff
@@ -415,7 +415,121 @@ jobs:
...
+      - name: Install kubectl and docker-cli
+        shell: bash
+        run: |
+          brew install kubectl
+          brew install docker
+          kubectl version --client=true
+      - name: Install Qemu and socket_vmnet
+        shell: bash
+        run: |
+          brew update
```
nirs, line 438: This should be done once per job. Can we install all packages in the same step?

.github/workflows/master.yml
```diff
@@ -415,7 +415,121 @@ jobs:
...
+          brew install qemu socket_vmnet
+          HOMEBREW=$(which brew) && sudo ${HOMEBREW} services start socket_vmnet
```
nirs, line 440: This can be in start socket_vmnet service step

.github/workflows/master.yml
```diff
@@ -415,7 +415,121 @@ jobs:
...
+          fw=/usr/libexec/ApplicationFirewall/socketfilterfw
+          sudo $fw --remove /usr/libexec/bootpd
+          sudo $fw --add /usr/libexec/bootpd
+          sudo $fw --unblock /usr/libexec/bootpd
```
nirs, line 444: this should be separate step like "fixup bootpd"

We have yaml for installing and configuring docker_vment in 2 workflows now, maybe we can convert it to local action, so we can use:

    use: actions/socket_vmnet.yaml

.github/workflows/master.yml
```diff
@@ -415,7 +415,121 @@ jobs:
...
+        run: |
+          set -x
+          pwd
+          cd minikube_binaries
+          chmod a+x e2e-*
+          chmod a+x minikube-*
+          mkdir -p report
+          # test home
+          mkdir -p /tmp/th
+          MINIKUBE_HOME=/tmp/th ./minikube-darwin-amd64 delete --all --purge
```
nirs, line 489: Why do we need a temporary minikube home in the CI?

Maybe define this in the env? this way we don't need to repeat it and there is no way to forget the use the temporary home.

.github/workflows/master.yml
```diff
@@ -415,7 +415,121 @@ jobs:
...
+          START_TIME=$(date -u +%s)
+          KUBECONFIG=$/tmp/th/kubeconfig MINIKUBE_HOME=/tmp/th ./e2e-darwin-amd64 -minikube-start-args="--vm-driver=qemu --wait-timeout=15m --network socket_vmnet"  -test.run TestFunctional -test.timeout=40m -test.v -timeout-multiplier=1.5 -binary=./minikube-darwin-amd64 2>&1 | tee ./report/testout.txt
```
nirs, line 491: Why do we need to specify KUBECONFIG? if this is required for everything in this job it should be in the job ENV.

.github/workflows/master.yml
```diff
@@ -415,7 +415,121 @@ jobs:
...
+          END_TIME=$(date -u +%s)
+          TIME_ELAPSED=$(($END_TIME-$START_TIME))
```
nirs, line 493: We defined TIME_ELAPSED=time in the env, maybe it is not needed?

.github/workflows/master.yml
```diff
@@ -415,7 +415,121 @@ jobs:
...
+          go tool test2json -t < ./report/testout.txt > ./report/testout.json || true
+          STAT=$(gopogh -in ./report/testout.json -out_html ./report/testout.html -out_summary ./report/testout_summary.json -name "${JOB_NAME} ${GITHUB_REF}" -repo "${GITHUB_REPOSITORY}"  -details "${GITHUB_SHA}")  || true
+          echo status: ${STAT}
+          FailNum=$(echo $STAT | jq '.NumberOfFail')
+          TestsNum=$(echo $STAT | jq '.NumberOfTests')
+          GOPOGH_RESULT="${JOB_NAME} : completed with ${FailNum} / ${TestsNum} failures in ${TIME_ELAPSED}"
+          echo "GOPOGH_RESULT=${GOPOGH_RESULT}" >> $GITHUB_ENV
+          echo 'STAT<<EOF' >> $GITHUB_ENV
+          echo "${STAT}" >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
```
nirs, line 512: Why we need al this here? This should not be part of the yaml, if we need so much babysitting for goppgh it should be in a shell scripts usable in all gopogh calls, and better inside gopogh so users do not need this to repeat this.

.github/workflows/master.yml
```diff
@@ -415,7 +415,121 @@ jobs:
...
+          echo "*******************${numPass} Passes :) *******************"
+          echo $STAT | jq '.PassedTests' || true
+          echo "*******************************************************"
+          echo "---------------- ${numFail} Failures :( ----------------------------"
+          echo $STAT | jq '.FailedTests' || true
+          echo "-------------------------------------------------------"
+          if [ "$numFail" -gt 0 ];then echo "*** $numFail Failed ***";exit 2;fi
+          if [ "$numPass" -eq 0 ];then echo "*** 0 Passed! ***";exit 2;fi
+          if [ "$numPass" -lt 56 ];then echo "*** Failed to pass at least 56 ! ***";exit 2;fi
+          if [ "$numPass" -eq 0 ];then echo "*** Passed! ***";exit 0;fi
```
nirs, line 532: This should not be in the GitHub action yaml, it should be part of gopogh or a script we use for all calls.

.github/workflows/pr.yml
```diff
@@ -413,7 +413,121 @@ jobs:
           if [ "$numPass" -eq 0 ];then echo "*** 0 Passed! ***";exit 2;fi
           if [ "$numPass" -lt 36 ];then echo "*** Failed to pass at least 36 ! ***";exit 2;fi
           if [ "$numPass" -eq 0 ];then echo "*** Passed! ***";exit 0;fi
-
+  functional_qemu_macos13_intel:
```
nirs, line 416: This looks like a copy of the previous yaml. We should not do this.

.github/workflows/master.yml
medyagh: those are used for gopogh report file names that are produced in the end

.github/workflows/master.yml
medyagh: sounds good, these are copy/paste of all jobs, these comments will be used in the follow up massive refactor

.github/workflows/master.yml
medyagh: will do in follow up matrifying

.github/workflows/master.yml
medyagh: I find those local actions annyoing, since you will have to look what they are doing, for such a small step, better to just do the comment

.github/workflows/master.yml
medyagh: this is because Qemu has a limitation of number of characters in machine folder, and action runners is in a nested folder longer than than max char

.github/workflows/master.yml
medyagh: some runners would come with their own kubeconfig

.github/workflows/master.yml
medyagh: I think when we define it there it makes it useable across the script, otherwise it would be local variable that goes away in next step

.github/workflows/master.yml
nirs: OK, this is the issue with the tests using bad minikube home directory. There is no limit in qemu, the issue is unix socket name length, and it is the same for all drivers using unix socket (qemu, vfkit, krunkit).

We need to fix the tests to use a better home directory (.e.g /tmp/minikube-integration-xxxyyy) instead of patching the qemu tests.

.github/workflows/pr.yml
medyagh: originally this was made this way so we can Filter all the jobs that Master, and look at the health of functional/build test on master branch and we even have a Status Badge for it on the minikube readme that refrences it. so thats why we sperated them.( this is probably 5 years old) and I am sure we can merge them in to a matrix in the coming refactor.

PR #21275, opened 2025-08-10
Use a well known 9p port range instead of random port

nirs:
Using a random port is not practical when using a firewall blocking access to the host on the libvirt network, which is the default in Fedora (for good reason).

We want to switch to well known port range so users can configure their firewall. User that want to reproduce the old way we configure the port range to the dynamic port range (49152 to 65535).

Minikube should manage the mount ports by keeping a pool of ports and picking the next available port from the pool.

Based on #21250 for testing


nirs:
This should really be an issue and much more work is needed. Will reopen when I have time to work on this and we have agreement on the design.

Issue #21301, opened 2025-08-12
Fix Unit Tests Windows


medyagh:
https://github.com/kubernetes/minikube/actions/runs/16902511147/job/47884533728?pr=21290

PR #21329, opened 2025-08-12
skip tests not relevant for windows

prezha:
fixes https://github.com/kubernetes/minikube/issues/21301

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -56,9 +56,8 @@ jobs:
           run: choco install make -y
         # TODO: add gopogh reports for unit tests too
         - name: unit test
-          timeout-minutes: 5
+          timeout-minutes: 15
           env:
             TESTSUITE: unittest
           run: make test
           continue-on-error: false
-
--- a/cmd/minikube/cmd/delete_test.go
+++ b/cmd/minikube/cmd/delete_test.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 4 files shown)


medyagh (changes requested):
@prezha seems like you checcked in a file, please rebase :) so it doesnt stay in history

cmd/minikube/cmd/delete_test.go
```diff
@@ -225,6 +226,9 @@ func TestDeleteAllProfiles(t *testing.T) {
 // then tries to execute the tryKillOne function on it;
 // if after tryKillOne the process still exists, we consider it a failure
 func TestTryKillOne(t *testing.T) {
```
medyagh, line 228: is it possible to make our Killing Func to work on windows?

.github/workflows/unit-test.yml
```diff
@@ -56,9 +56,8 @@ jobs:
           run: choco install make -y
         # TODO: add gopogh reports for unit tests too
         - name: unit test
-          timeout-minutes: 5
+          timeout-minutes: 15
```
medyagh, line 59: was it really needed 15 mins?

cmd/minikube/cmd/delete_test.go
bobsira: Yes.The test relied on platform-specific process/Wait() behavior and/or error text ("killed"). Windows and POSIX terminate processes differently: POSIX signals (SIGHUP/SIGKILL) and their wait/exit semantics differ from Windows TerminateProcess. The test assumed Windows Wait() would return a non-nil error that contains "killed", but in your environment Wait() returned nil. That mismatch made the assertion fail. Pushing a fix for this in a bit

medyagh (approved):

PR #21399, opened 2025-08-22
lint: Improve go mod tidy lint step

nirs:
- Fix incorrect usage of git diff
- Use git diff-index to detect changes
- Show full diff to show the required change
- Move the fix suggestion to the end to make it easier to find
- Group log line with ::group:: workflow command[1]
- Use ::notice workflow command[2] for the fix suggestion so we don't get multiple errors annotations

## Example failed job

Screenshot 2025-08-22 at 19 46 29

[1] https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#grouping-log-lines
[2] https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-a-notice-message

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -60,17 +60,17 @@ jobs:
           with:
             go-version: ${{env.GO_VERSION}}
             cache: true  
-        - name: Verify gomod tidy
+        - name: Verify go mod tidy
           timeout-minutes: 2
           continue-on-error: false
           run: |
+            echo "::group::go mod tidy"
             make gomodtidy
-            if ! git diff --quiet; then
-              echo "::error::Please run 'make gomodtidy' and commit the changes"
-              echo "=== files ==="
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -60,17 +60,17 @@ jobs:
...
           run: |
+            echo "::group::go mod tidy"
             make gomodtidy
-            if ! git diff --quiet; then
-              echo "::error::Please run 'make gomodtidy' and commit the changes"
-              echo "=== files ==="
-              git diff --name-only
-              echo "=== diff ==="
-              git diff -
+            echo "::endgroup::"
```
medyagh, line 69: oh I love this !!! that is so great !!! we need to also add that to the info block, so we can Fold and Unfold each info category

nirs: I can add an issue for using groups. Lets fix this first.

PR #21400, opened 2025-08-22
test: Fix service test on arm64

nirs:
The echoserver-arm:1.8 image is not an arm64 image:

    % kubectl create deployment echoserver --image registry.k8s.io/echoserver-arm:1.8
    deployment.apps/echoserver created

    % kubectl logs deploy/echoserver
    exec /usr/sbin/nginx: exec format error

    % minikube ssh -- sudo nerdctl -n k8s.io image ls registry.k8s.io/echoserver-arm:1.8
    REPOSITORY                        TAG    IMAGE ID        CREATED               PLATFORM       SIZE        BLOB SIZE
    registry.k8s.io/echoserver-arm    1.8    b33d4cdf6ed0    About a minute ago    linux/amd64    90.1 MiB    43.2 MiB 

Replace with kickbase/echo-server we used in other tests.

With this change ServerCmdConnect pass. The other test using the broken
arm image was not failing, maybe it was not checking the deployment
status properly.

Functional tests time reduced from 230 seconds to 200 seconds:

    --- FAIL: TestFunctional (200.87s)

### DIFF
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -1442,18 +1442,13 @@ func validateServiceCmd(ctx context.Context, t *testing.T, profile string) {
 	validateServiceCmdURL(ctx, t, profile)
 }
 
-// validateServiceCmdDeployApp Create a new `registry.k8s.io/echoserver` deployment
+// validateServiceCmdDeployApp Create a new `kickbase/echo_server` deployment
 func validateServiceCmdDeployApp(ctx context.Context, t *testing.T, profile string) {
 	t.Run("DeployApp", func(t *testing.T) {
 		var rr *RunResult
 		var err error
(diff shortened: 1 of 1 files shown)


medyagh (changes requested):
how about we push a Arm64 echo-server image somewhere and use that ? we could use one of your quai.io images? this is an important test to see if service works

nirs:
> how about we push a Arm64 echo-server image somewhere and use that ? we could use one of your quai.io images? this is an important test to see if service works

We can but this requires more work so not in scope of this PR.

nirs:
Trying to fix the image commands I see that we use `docker.io/kicbase/echo-server`:

```
% podman inspect docker.io/kicbase/echo-server | jq '.[0].Architecture'
"arm64"
```

Trying to use it in the service test...

Issue #21398, opened 2025-08-22
Kubernetes version default: v1.33.4 tests time out on kvm-linux


medyagh:
as seen in this PR (I repeated the PR multiple times ) and I forced to re-generate the preloads just to ensure the proeloads were not bad and that didnt make a difference.
https://github.com/kubernetes/minikube/pull/21397

### the tests that timed out

- Functional Test Docker Containerd Linux (this one never flakes or time out)
https://github.com/kubernetes/minikube/actions/runs/17141932774/job/48631038305?pr=21397
- Baremetal fucntional test timed out https://github.com/kubernetes/minikube/actions/runs/17141932774/job/48631038292?pr=21397
- [KVM Linux (full test - docker runtime)](https://github.com/kubernetes/minikube/pull/21397)

Docker linux does not time out but it takes 4-5 minutes longer than usual the usual is (48 minutes) acording to [gopogh server](https://gopogh-server-tts3vkcpgq-uc.a.run.app/)
```
[Docker_Linux](https://storage.googleapis.com/minikube-builds/logs/21397/41071/Docker_Linux.html) — Jenkins: completed with success in 53.20 minutes.
```

medyagh:
[#37219 - PR #21397 @ 41071.txt](https://github.com/user-attachments/files/21943679/37219.-.PR.21397.%40.41071.txt)

medyagh:
I also checked [KVM_Linux_containerd](https://storage.googleapis.com/minikube-builds/logs/21397/41071/KVM_Linux_containerd.html)
KVM_Linux_containerd — Jenkins: completed with 9 / 416 failures in 101.53 minutes.

KVM_Linux_containerd usually takes 77 minutes https://gopogh-server-tts3vkcpgq-uc.a.run.app/

so it is probably not a cri-dockerd related

prezha:
i found that kube-proxy fails to start with kvm driver and k8s v1.34.*, which leads to the integration tests timeouts
reported the issue upstream:
https://kubernetes.slack.com/archives/C09QYUH5W/p1756131900526569?thread_ts=1754135835.803739&cid=C09QYUH5W

medyagh:
Copy Pasting @prezha's Message in minikube Slack for reference
https://kubernetes.slack.com/archives/C1F5CT6Q1/p1756132251012779?thread_ts=1755801359.036009&cid=C1F5CT6Q1

```
[...]

PR #21402, opened 2025-08-22
debug: bump k8s and enable ipv6 in test

medyagh:
blind attempt to debug https://github.com/kubernetes/minikube/issues/21398
raw time out logs here https://github.com/kubernetes/minikube/issues/21398#issuecomment-3215358001 (saw some stuff about kube-prxy)
attmepting to load ipv6 maybe that will un-time out kvm test


medyagh:
BTW this one also timed out for kvm linux

PR #21405, opened 2025-08-23
test: Fix image tests on macOS

nirs:
Fix image tests on macOS:

- Rewrite the tests that do not need docker daemon to use the container runtime in minikube.
- Check saved image, revealing #21408
- Add test for `image tag`
- Skip tests that require docker daemon if one is not available on the host.
- Improve image list tests to do proper parsing and image name lookup
- Improve image existence checks

Example run:

```
--- PASS: TestFunctional/parallel/ImageCommands (1.62s)
    --- PASS: TestFunctional/parallel/ImageCommands/ImageListShort (0.07s)
    --- PASS: TestFunctional/parallel/ImageCommands/ImageListTable (0.07s)
    --- PASS: TestFunctional/parallel/ImageCommands/ImageListJSON (0.07s)
    --- PASS: TestFunctional/parallel/ImageCommands/ImageListYAML (0.07s)
    --- SKIP: TestFunctional/parallel/ImageCommands/SetupDaemon (0.00s)
    --- SKIP: TestFunctional/parallel/ImageCommands/ImageLoadFromDaemon (0.00s)
    --- SKIP: TestFunctional/parallel/ImageCommands/ImageReloadFromDaemon (0.00s)
    --- SKIP: TestFunctional/parallel/ImageCommands/ImageTagAndLoadFromDaemon (0.00s)
    --- SKIP: TestFunctional/parallel/ImageCommands/ImageSaveToDaemon (0.00s)
    --- PASS: TestFunctional/parallel/ImageCommands/ImageTag (0.30s)
    --- PASS: TestFunctional/parallel/ImageCommands/ImageRemove (0.29s)
    --- PASS: TestFunctional/parallel/ImageCommands/ImageSaveToFile (0.17s)
    --- PASS: TestFunctional/parallel/ImageCommands/ImageLoadFromFile (0.58s)
    --- PASS: TestFunctional/parallel/ImageCommands/ImageBuild (4.21s)
```

## Issues

- Testing with pause image does not work because of #21408
- Test with "docker.io/kicbase/echo-server:1.0" instead so we don't have to skip tests
- `image pull` not tested
- `image push` not tested but we need local registry to test it

### DIFF
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -19,6 +19,7 @@ limitations under the License.
 package integration
 
 import (
+	"archive/tar"
 	"bufio"
 	"bytes"
 	"context"
@@ -457,10 +458,7 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
 		testPath := filepath.Join(t.TempDir(), "test.tar")
 
 		saveImageToFile(ctx, t, profile, pauseImage, testPath)
-
-		if _, err := os.Stat(testPath); err != nil {
-			t.Fatalf("expected %q to exist after `image save`, but doesn't exist", testPath)
-		}
+		checkSavedImage(t, testPath)
 	})
 
 	// docs: Make sure image loading from file works by `minikube image load`
@@ -475,6 +473,8 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
 		})
 
 		saveImageToFile(ctx, t, profile, testImage, testPath)
+		checkSavedImage(t, testPath)
+
 		removeImage(ctx, t, profile, testImage)
 		loadImageFromFile(ctx, t, profile, testPath)
 
@@ -493,6 +493,36 @@ func checkImageNotExists(ctx context.Context, t *testing.T, profile string, imag
 	images := listImagesShort(ctx, t, profile)
 	if slices.Contains(images, image) {
 		t.Fatalf("expected %q to not exist in minikube but the image is there", image)
+	}
+}
+
+// checkSavedImage validates a saved image. Validating the contents is not easy
+// without depending on docker or podmamn, but we can check that we have a non
+// empty tar file. We log the tar contents for debugging.
+func checkSavedImage(t *testing.T, tarPath string) {
+	f, err := os.Open(tarPath)
+	if err != nil {
+		t.Fatalf("failed to open file %q: %v", tarPath, err)
+	}
+	defer f.Close()
+
+	reader := tar.NewReader(f)
+	count := 0
+
+	for {
+		header, err := reader.Next()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			t.Fatalf("failed to read tar header: %s", err)
+		}
+		t.Logf("tar: %s (%d bytes)", header.Name, header.Size)
+		count++
+	}
+
+	if count == 0 {
+		t.Fatalf("tar file is empty: %q", tarPath)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


test/integration/main_test.go
```diff
@@ -175,6 +175,17 @@ func VMDriver() bool {
 	return !KicDriver() && !NoneDriver()
 }
 
+// ContainerCommand retrurn the first available container command.
+func ContainerCommand() (string, error) {
```
medyagh: this assumes it is okay with if "docker" is in the path but it is only cli and not connected to any docker-server, right?

medyagh:
@nirs please rebase

test/integration/functional_test.go
```diff
@@ -343,32 +346,39 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+	taggedImage := fmt.Sprintf("%s:%s", echoServerImage, profile)
 
 	t.Run("Setup", func(t *testing.T) {
 		var err error
-		imagePath, err = filepath.Abs(imageFile)
+
+		containerCmd, err := ContainerCommand()
 		if err != nil {
-			t.Fatalf("failed to get absolute path of file %q: %v", imageFile, err)
+			t.Skip(err)
```
medyagh: Consider adding a Log Scaring the Tester that We Skipped but you are not off the hook, you better test this and not feel so good about it 

t.Log(....something With that message ....)

test/integration/functional_test.go
```diff
@@ -343,32 +346,39 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
 
 	// docs: Make sure image loading from Docker daemon works by `minikube image load --daemon`
-	t.Run("ImageLoadDaemon", func(t *testing.T) {
+	t.Run("ImageLoadFromDaemon", func(t *testing.T) {
+		containerCmd, err := ContainerCommand()
+		if err != nil {
+			t.Skip(err)
+		}
+
+		if containerCmd != "docker" {
```
medyagh: does this assume Existance of Docker-cli means existance of Docker-server ?

you can install brew install docker, which gives them only docker cli
https://minikube.sigs.k8s.io/docs/tutorials/docker_desktop_replacement/

test/integration/functional_test.go
```diff
@@ -378,7 +388,16 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
 
 	// docs: Try to load image already loaded and make sure `minikube image load --daemon` works
-	t.Run("ImageReloadDaemon", func(t *testing.T) {
+	t.Run("ImageReloadFromDaemon", func(t *testing.T) {
+		containerCmd, err := ContainerCommand()
+		if err != nil {
+			t.Skip(err)
+		}
+
+		if containerCmd != "docker" {
```
medyagh: maybe the check should be checking existance of Docker server?
we have a func in minikube that Caches the docker Info
https://github.com/kubernetes/minikube/blob/7d4604a3681029e26a363922952cd51da9e65623/pkg/drivers/kic/oci/info.go#L60

which Caches it once and lets the minikube run uses it (without querying it over and over since it takes time 1-2 seconds for docker info to come back)

nirs: oci.CachedDaemonInfo is what we need.

nirs:
I'm working on v2, thanks for reviewing.

test/integration/functional_test.go
nirs: We skip now only when docker daemon is not available, and I think he message is clear enough. We can add a helper to log a more verbose message and reuse it in all tests that require a docker daemon.

PR #21410, opened 2025-08-24
iso: update kernel lts version to 6.6.102

prezha:


nirs:
Can you update the title to match the actual version?

medyagh:
ok-to-test

prezha:
rebased to current master head and squashed commits

note that this was only for testing - for ruling out the kernel version being the problem for the issue with kube-proxy, which proved to be correct assumption, ie, i haven't looked into if we need some additional kernel config changes as well

PR #21426, opened 2025-08-26
test: make chown tests cross-platform and relax TryKill test on Windows

bobsira:
- Made chown-related tests portable: removed compile-time Unix-only assumptions, closed temp files, and used runtime checks + reflection so ownership assertions run only on platforms that expose Uid/Gid. This prevents Windows build failures and TempDir cleanup errors.
- Adjusted behavior expectations for Windows SIDs (non-numeric Uid) so tests don't call strconv.Atoi on a SID.
- Consolidated duplicated test loops to run each case once and avoid flaky/misleading failures.
- Relaxed the process-kill test (TestTryKillOne) to treat any process termination as success and stop asserting on platform-specific Wait() error strings (Windows vs POSIX differences).

### DIFF
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -22,8 +22,10 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/docker/machine/libmachine/mcnerror"
@@ -674,9 +676,22 @@ func trySigKillProcess(pid int) error {
 	}
 
 	klog.Infof("Killing pid %d ...", pid)
-	if err := proc.Kill(); err != nil {
-		klog.Infof("Kill failed with %v - removing probably stale pid...", err)
-		return errors.Wrapf(err, "removing likely stale unkillable pid: %d", pid)
+	// On non-Windows try to send SIGUP first (graceful), fallback to kill if signaling fails.
+	// Use numeric syscall.Signal(1) (SIGUP) to avoid referencing platform-specific constant names.
+	if runtime.GOOS != "windows" {
+		if err := proc.Signal(syscall.Signal(1)); err != nil {
+			klog.Infof("SIGUP failed with %v - falling back to kill...", err)
+			if err := proc.Kill(); err != nil {
+				klog.Infof("Kill failed with %v - removing probably stale pid...", err)
+				return errors.Wrapf(err, "removing likely stale unkillable pid: %d", pid)
+			}
+		}
+	} else {
+		// On Windows, Signal is not meaningful, so we just kill the process
+		if err := proc.Kill(); err != nil {
+			klog.Infof("Kill failed with %v - removing probably stale pid...", err)
+			return errors.Wrapf(err, "removing likely stale unkillable pid: %d", pid)
+		}
 	}
 
 	return nil
--- a/cmd/minikube/cmd/delete_test.go
+++ b/cmd/minikube/cmd/delete_test.go
@@ -21,7 +21,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strings"
+	"runtime"
 	"testing"
 
 	"github.com/docker/machine/libmachine"
@@ -277,7 +277,17 @@ func main() {
 	}
 
 	// waiting for process to exit
-	if err := processToKill.Wait(); !strings.Contains(err.Error(), "killed") {
-		t.Fatalf("unable to kill process: %v\n", err)
+	waitErr := processToKill.Wait()
+	if runtime.GOOS == "windows" {
+		// Windows Wait commonly returns a non-nil error (exit status).
+		// Expect a non-nil Wait error on Windows; fail if we got nil.
+		if waitErr == nil {
+			t.Fatalf("expected non-nil Wait error on windows, got nil")
+		}
+	} else {
+		// On POSIX we signal SIGHUP — child should exit cleanly.
+		if waitErr != nil {
+			t.Fatalf("unable to kill process: %v\n", waitErr)
+		}
 	}
 }
--- a/pkg/util/utils_test.go
+++ b/pkg/util/utils_test.go
@@ -19,7 +19,9 @@ package util
 import (
 	"os"
 	"os/user"
-	"syscall"
+	"reflect"
+	"runtime"
+	"strconv"
 	"testing"
 
 	"github.com/blang/semver/v4"
@@ -81,9 +83,28 @@ func TestParseKubernetesVersion(t *testing.T) {
 
 func TestChownR(t *testing.T) {
 	testDir := t.TempDir()
-	if _, err := os.Create(testDir + "/TestChownR"); err != nil {
+	f, err := os.Create(testDir + "/TestChownR")
+	if err != nil {
 		return
 	}
+	_ = f.Close()
+
+	curUID := -1
+	curGID := -1
+	if u, err := user.Current(); err == nil {
+		if u.Uid != "" {
+			if v, err := strconv.Atoi(u.Uid); err == nil {
+				curUID = v
+			}
+		}
+		if u.Gid != "" {
+			if v, err := strconv.Atoi(u.Gid); err == nil {
+				curGID = v
+			}
+		}
+	}
+
+	normalExpectError := runtime.GOOS == "windows"
 
 	cases := []struct {
 		name          string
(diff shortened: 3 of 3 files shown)


cmd/minikube/cmd/delete.go
```diff
@@ -674,9 +676,22 @@ func trySigKillProcess(pid int) error {
...
+				klog.Infof("Kill failed with %v - removing probably stale pid...", err)
+				return errors.Wrapf(err, "removing likely stale unkillable pid: %d", pid)
+			}
+		}
+	} else {
+		// On Windows, Signal is not meaningful, so we just kill the process
+		if err := proc.Kill(); err != nil {
+			klog.Infof("Kill failed with %v - removing probably stale pid...", err)
+			return errors.Wrapf(err, "removing likely stale unkillable pid: %d", pid)
+		}
```
nirs: If we want to do graceful termination,we should not add platform specific code in this file. We have a process package supporting both posix and windows. Can we use it here?

cmd/minikube/cmd/delete.go
```diff
@@ -674,9 +676,22 @@ func trySigKillProcess(pid int) error {
 	}
 
 	klog.Infof("Killing pid %d ...", pid)
-	if err := proc.Kill(); err != nil {
-		klog.Infof("Kill failed with %v - removing probably stale pid...", err)
-		return errors.Wrapf(err, "removing likely stale unkillable pid: %d", pid)
+	// On non-Windows try to send SIGUP first (graceful), fallback to kill if signaling fails.
+	// Use numeric syscall.Signal(1) (SIGUP) to avoid referencing platform-specific constant names.
```
nirs: This is not graceful shutdown on posix. Most processes do not handle SIGHUP but they expect SIGTERM as gracefull termination. Previous code did not do graceful termination, why do we ned to do this now?

cmd/minikube/cmd/delete_test.go
```diff
@@ -277,7 +277,17 @@ func main() {
...
+		// Windows Wait commonly returns a non-nil error (exit status).
+		// Expect a non-nil Wait error on Windows; fail if we got nil.
+		if waitErr == nil {
+			t.Fatalf("expected non-nil Wait error on windows, got nil")
+		}
+	} else {
+		// On POSIX we signal SIGHUP — child should exit cleanly.
+		if waitErr != nil {
+			t.Fatalf("unable to kill process: %v\n", waitErr)
+		}
```
nirs: We should not have platform specific code in this file. If we need different code for different platforms it should be implemented once in minikube. Can we add this code to the process package?

cmd/minikube/cmd/delete_test.go
```diff
@@ -277,7 +277,17 @@ func main() {
...
-		t.Fatalf("unable to kill process: %v\n", err)
+	waitErr := processToKill.Wait()
+	if runtime.GOOS == "windows" {
+		// Windows Wait commonly returns a non-nil error (exit status).
+		// Expect a non-nil Wait error on Windows; fail if we got nil.
+		if waitErr == nil {
+			t.Fatalf("expected non-nil Wait error on windows, got nil")
+		}
+	} else {
+		// On POSIX we signal SIGHUP — child should exit cleanly.
```
nirs: SIGHUP is not graceful termination, we expect SIGTERM.

pkg/util/utils_test.go
```diff
@@ -81,9 +83,28 @@ func TestParseKubernetesVersion(t *testing.T) {
...
+		if u.Uid != "" {
+			if v, err := strconv.Atoi(u.Uid); err == nil {
+				curUID = v
+			}
+		}
+		if u.Gid != "" {
+			if v, err := strconv.Atoi(u.Gid); err == nil {
+				curGID = v
+			}
+		}
```
nirs: How errors from strconv.Atoi(u.Uid) and strconv.Atoi(u.Gid) are handled?

pkg/util/utils_test.go
```diff
@@ -93,53 +114,97 @@ func TestChownR(t *testing.T) {
...
-			gid:           os.Getgid(),
+			gid:           curGID,
 			expectedError: true,
 		},
 		{
 			name:          "invalid gid",
-			uid:           os.Getuid(),
+			uid:           curUID,
 			gid:           2147483647,
 			expectedError: true,
```
nirs, line 113: This table test is too complicated. We should group the happy path cases in table test that must succeed, and the negative cases in another test that must fail with some error.

pkg/util/utils_test.go
```diff
@@ -93,53 +114,97 @@ func TestChownR(t *testing.T) {
...
-			fileInfo, _ := os.Stat(testDir + "/TestChownR")
+			// Basic expectation: error vs no error
+			if (err != nil) != c.expectedError {
+				t.Fatalf("case %q: expectedError=%v, got err=%v", c.name, c.expectedError, err)
+			}
+
+			// Skip strict ownership assertions on Windows or if the operation errored.
+			if runtime.GOOS == "windows" || c.expectedError {
+				return
+			}
```
nirs: Instead of complicating all tests, create a new test for windows since it has different behaviour.

pkg/util/utils_test.go
```diff
@@ -93,53 +114,97 @@ func TestChownR(t *testing.T) {
...
+			gidField := v.FieldByName("Gid")
+			if uidField.IsValid() && gidField.IsValid() {
+				uid := int(uidField.Convert(reflect.TypeOf(uint32(0))).Uint())
+				gid := int(gidField.Convert(reflect.TypeOf(uint32(0))).Uint())
+				if curUID != -1 && uid != curUID {
+					t.Fatalf("ownership uid mismatch: expected %d, got %d", curUID, uid)
+				}
+				if curGID != -1 && gid != curGID {
+					t.Fatalf("ownership gid mismatch: expected %d, got %d", curGID, gid)
+				}
```
nirs: This seems very complicated compared to the old code. What are the new behaviors that this add?

pkg/util/utils_test.go
```diff
@@ -93,53 +114,97 @@ func TestChownR(t *testing.T) {
...
+	// Ensure CHANGE_MINIKUBE_NONE_USER is set for the test.
 	if os.Getenv("CHANGE_MINIKUBE_NONE_USER") == "" {
 		t.Setenv("CHANGE_MINIKUBE_NONE_USER", "1")
 	}
 
-	if os.Getenv("SUDO_USER") == "" {
-		user, err := user.Current()
-		if nil != err {
-			t.Error("fail to get user")
+	// Determine if the current user's UID string is numeric (Unix) or not (Windows SIDs).
```
nirs: This looks like code to put in a sudo helper. Putting it here means we will duplicate it in many tests quickly.

bobsira:
creating an issue to track all the windows failures here -> https://github.com/kubernetes/minikube/issues/21427

cmd/minikube/cmd/delete.go
```diff
@@ -674,9 +676,22 @@ func trySigKillProcess(pid int) error {
...
+	if runtime.GOOS != "windows" {
```
djeada: you can just call `err := process.Terminate(pid, "")` and then handle it with `if err != nil { ... }`. the package is `k8s.io/minikube/pkg/minikube/process`

nirs: I would not use the process package if we don't have pid file. But maybe we an extend it to handle child processes where you know the pid and we don't need to verify that the pid exists.

pkg/util/utils_test.go
bobsira: Moving the fix for TestMaybeChownDirRecursiveToMinikubeUser and TestChownR unit tests to a different PR as recommended by @medyagh

pkg/util/utils_test.go
bobsira: ditto

pkg/util/utils_test.go
bobsira: ditto

pkg/util/utils_test.go
bobsira: ditto

pkg/util/utils_test.go
bobsira: ditto

medyagh:
@bobsira plz see nirs comments and also rebase

cmd/minikube/cmd/delete_test.go
bobsira: Removed this!

cmd/minikube/cmd/delete.go
bobsira: removed this!

cmd/minikube/cmd/delete_test.go
bobsira: I've added ExistsPID functionality in the process package

cmd/minikube/cmd/delete.go
bobsira: cc: @djeada @nirs 

TestTryKillOne now starts a small child Go program that listens for SIGTERM and then invokes trySigKillProcess(pid) to request termination; the test calls processToKill.Wait() only to consume the child’s exit for diagnostics, and then uses the centralized, platform-aware helper process.ExistsPID(pid) to check the OS process table—if ExistsPID reports the PID still exists the test fails, otherwise it passes, avoiding brittle runtime.GOOS branching or string-parsing of Wait() errors.

cmd/minikube/cmd/delete.go
bobsira: yes. I've removed this bit of code.

pkg/minikube/process/process.go
```diff
@@ -73,6 +73,20 @@ func Exists(pid int, executable string) (bool, error) {
 	return entry.Executable() == executable, nil
 }
 
+// ExistsPID reports whether a process with the given pid exists.
+// This is a PID-only check (no executable name matching).
+func ExistsPID(pid int) (bool, error) {
```
nirs: When do we call this? if we have a child process we know that the pid exist and we don't need to look it up since the pid cannot be reused until we wait() for the child.

cmd/minikube/cmd/delete_test.go
```diff
@@ -277,7 +277,12 @@ func main() {
...
 	// waiting for process to exit
-	if err := processToKill.Wait(); !strings.Contains(err.Error(), "killed") {
-		t.Fatalf("unable to kill process: %v\n", err)
+	waitErr := processToKill.Wait()
+	exists, err := process.ExistsPID(pid)
+	if err != nil {
+		t.Fatalf("error checking process existence for pid %d: %v", pid, err)
+	}
+	if exists {
+		t.Fatalf("process %d still exists after trySigKillProcess; waitErr=%v", pid, waitErr)
```
nirs: sending SIGKILL (or any other signal is asynchronous. It can return before the process was killed.

So checking if the pid exist in the processes table is racy and incorrect. The way to check the process status and detect if it was killed is to wait() for the process:
https://pkg.go.dev/os#Process.Wait

and get the system dependent exist status:
https://pkg.go.dev/os#ProcessState.Sys

On posix we can get check if the process was terminated by signal and get the signal from
https://pkg.go.dev/syscall#WaitStatus

In process_test.go we test Kill() little differently:
1. Kill the process (async)
2. We start a timer to fail test after a short timeout
3. Wait of the process
4. Check that Exists() return false

We don't verify the exit status since our test process can be configured to block SIGTERM, and we know that the only reason for termination can be SIGKILL.

If we need ugly platform specific code to check the exit status we can put it in the process package to make this easy to use everywhere. The package already have build tags to make it easy to do the right thing on windows and other platforms.

bobsira: In the current change,

1. I've kill the process with trySigKillProcess(PID)
2. Then I've created a channel and spawn goroutine to Wait()
3. I then wait for either process exit or timeout 
4. Finally I check process table for PID presence.

I've avoided adding any platform-dependent code checks for exit status to make a short and clean change.

any thoughts, @nirs

nirs: Waiting on for Wait() with t timeout looks correct and better than the way we do this in process_test.go, since we know we will stop waiting on timeout. I'm not sure the wait is intrrupted in current code in process_test.go.

The check for pid exists seems unneeded - if the the wait completed we know the process terminated so there is nothing to check. And if there is nothing to check there is no need to add the process.ExistPID().

I agree that we don't have to verify that the process was terminated by signal, but this seems like a useful utility to have.

pkg/minikube/process/process.go
```diff
@@ -73,6 +73,20 @@ func Exists(pid int, executable string) (bool, error) {
...
+	if pid <= 0 {
+		return false, nil
+	}
```
nirs: This check seems unneeded - why would you call with invalid pid?

pkg/minikube/process/process.go
```diff
@@ -73,6 +73,20 @@ func Exists(pid int, executable string) (bool, error) {
...
+
+	entry, err := ps.FindProcess(pid)
+	if err != nil {
+		return true, err
+	}
+	return entry != nil, nil
```
nirs: This is incorrect - we have pidExist() helper that does the right thing on windows and posix.

pkg/minikube/process/process.go
nirs: I would call this PIDExist() - since it is a little bit confusing with Exists() which is about a process, using a pidfile. We already have a function pidExists() so this function can simply expose it, or we can renamse pidExists() to PIDExists() to make it useful outside of the package.

pkg/minikube/process/process.go
bobsira: @nirs I've centralize the pid-existence logic and exposed PIDExists functionality

cmd/minikube/cmd/delete_test.go
bobsira: @nirs is there anything in specific that you'd want me to further address in this PR?

bobsira: @medyagh I believe this PR can be checked in now.

pkg/minikube/process/process.go
```diff
@@ -73,6 +73,15 @@ func Exists(pid int, executable string) (bool, error) {
 	return entry.Executable() == executable, nil
 }
 
+// PIDExists reports whether a process with the given pid exists (PID-only).
+func PIDExists(pid int) (bool, error) {
```
nirs: I don't think this is the way. Previously we have 2 simple implementations using the os package without dependency on any library. Now both of them depend on a 3rd party library which is archived and not maintained for new years. We should use this library only when we cannot use the standard library.

The right change would be to remove this function, and rename pidExists() to PIDExists().

pkg/minikube/process/process_other.go
```diff
@@ -24,16 +24,7 @@ import (
...
-	// os.ErrProcessDone from Signal or Wait.
-	process, err := os.FindProcess(pid)
-	if err != nil {
-		return true, err
-	}
-	if process.Signal(syscall.Signal(0)) == os.ErrProcessDone {
-		return false, nil
-	}
-	return true, nil
+	return PIDExists(pid)
```
nirs: Keep this function as is, and rename it to PIDExists()

pkg/minikube/process/process_windows.go
```diff
@@ -21,10 +21,7 @@ import (
 )
 
 func pidExists(pid int) (bool, error) {
-	// Fails with "OpenProcess: The parameter is incorrect" if the process does
-	// not exist.
-	_, err := os.FindProcess(pid)
-	return err == nil, nil
+	return PIDExists(pid)
```
nirs: Keep this function as is, and rename it to PIDExists()

nirs:
> @bobsira, this is not a comment about the actual diff and nature of the changes but an observation about the commits in this PR:
...
> Would you consider cleaning the series of commits?

I agree. Unfortunately clean commit history is not popular in this project. I hope we can improve in this area.

I would split it to 2 commits:
- Exporting process.PIDExist
- Fixing TestTryKillOne on windows

pkg/minikube/process/process_windows.go
medyagh: @bobsira I agree please just rename the func to be  PIDExists

no need for the private func to call the public func

pkg/minikube/process/process_other.go
medyagh: @bobsira I agree please just rename the func to be PIDExists

no need for the private func to call the public func

cmd/minikube/cmd/delete_test.go
medyagh: @bob plz check my comment bellow

pkg/minikube/process/process.go
bobsira: Removed this!

pkg/minikube/process/process.go
bobsira: Got rid of pidExists logic in this PR. If we need it we can have it in another PR.

pkg/minikube/process/process_windows.go
bobsira: Got rid of pidExists logic in this PR. If we need it we can have it in another PR.

bobsira:
> > @bobsira, this is not a comment about the actual diff and nature of the changes but an observation about the commits in this PR:
> > ...
> > Would you consider cleaning the series of commits?
> 
> I agree. Unfortunately clean commit history is not popular in this project. I hope we can improve in this area.
> 
> I would split it to 2 commits:
> 
> * Exporting process.PIDExist
> * Fixing TestTryKillOne on windows

Thanks for this, I've squash committed this PR

pkg/minikube/process/process.go
bobsira: We can address this pidExisits logic in a different PR in we need it. Keeping this PR simple for now.

medyagh (approved):

PR #21456, opened 2025-09-01
Testing github.com/ulikunitz/x build failure

nirs:
Reproducing https://storage.googleapis.com/minikube-builds/logs/21409/build.txt with a minimal change.

```
# github.com/ulikunitz/xz/lzma
/go/pkg/mod/github.com/ulikunitz/xz@v0.5.14/lzma/reader.go:33:15: cannot use 1 << 31 (untyped int constant 2147483648) as int value in assignment (overflows)
# github.com/ulikunitz/xz/lzma
/go/pkg/mod/github.com/ulikunitz/xz@v0.5.14/lzma/reader.go:33:15: cannot use 1 << 31 (untyped int constant 2147483648) as int value in assignment (overflows)
```

This seems to be related to #21446:


nirs:
Resolved in #21466

PR #21409, opened 2025-08-24
iso: bump to containerd/nerdctl 2.x

nirs:
This is #21368 with the following changes:
- Rebase on master
- Fix whitespace in hash files
- Unify containerd-bin makefiles
- Update contained-bin Config.in
- Remove containerd.conf.d directory
- Add missing _AARCH64 to variable names
- Bump nerdctl to 2.1.3 (it must be upgraded with containerd)
- Update containerd update command for installing binaries

The update command will update containerd to v2.2.0 on the next update, so I'm not spending time on testing v2.1.5 now.

## Notes
- nerdctl update must be applied before updating containerd since nerdctl 2.1.x does not support containerd 2.2.x
  - https://github.com/containerd/nerdctl/releases/tag/v2.1.6
  - https://github.com/containerd/nerdctl/releases/tag/v2.2.0

Fixes #20497
Fixes #21986

PR #21466, opened 2025-09-01
deps: Update github.com/ulikunitz/xz to v0.5.15

nirs:
v0.5.14 broke 32 bit builds:
https://github.com/ulikunitz/xz/issues/65
    
The issue is fixed in v0.5.15:
https://github.com/ulikunitz/xz/releases/tag/v0.5.15

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -205,7 +205,7 @@ require (
 	github.com/subosito/gotenv v1.6.0 // indirect
 	github.com/tklauser/go-sysconf v0.3.12 // indirect
 	github.com/tklauser/numcpus v0.6.1 // indirect
-	github.com/ulikunitz/xz v0.5.14 // indirect
+	github.com/ulikunitz/xz v0.5.15 // indirect
 	github.com/vbatts/tar-split v0.12.1 // indirect
 	github.com/x448/float16 v0.8.4 // indirect
 	github.com/xo/terminfo v0.0.0-20220910002029-abceb7e1c41e // indirect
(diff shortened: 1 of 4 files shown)


medyagh:
how did it pass on https://github.com/kubernetes/minikube/pull/21446 ??
https://github.com/kubernetes/minikube/actions/runs/17306180270/job/49129544915

nirs:
> how did it pass on #21446 ?? https://github.com/kubernetes/minikube/actions/runs/17306180270/job/49129544915

I don't see a Build Cross test for this PR. It seems that the Jenkins jobs are not reflected properly in the test results. We really need to use GitHub actions for all tests to have proper tracking of tests results after a PR was merged.

nirs:
I rebased #21409 on the commit before we updated xz to 0.5.14, but it does not help, the Cross Build job still fails with the same issue:
https://storage.googleapis.com/minikube-builds/logs/21409/build.txt

So we get this version from other other component even if we don't require it, and it seems to be broken now.

```console
% go mod why -m github.com/ulikunitz/xz
# github.com/ulikunitz/xz
k8s.io/minikube/pkg/minikube/download
github.com/hashicorp/go-getter
github.com/ulikunitz/xz
```

nirs:
I see that go-getter is requiring 0.5.10:
https://github.com/hashicorp/go-getter/blob/66b90c505ff8898bbc5b93fef0156943067e9feb/go.mod#L17C2-L17C33

Why are we updating an indirect dependency? We need a good reason to do such update, for example CVE or performance improvement. Project releasing a new version is not a good reason.

nirs:
Looks like 0.5.14 did fix a CVE:
https://github.com/ulikunitz/xz/commit/88ddf1d0d98d688db65de034f48960b2760d2ae2

But the fix introduced incompatibility with 32 bit build. Fixed here:
https://github.com/ulikunitz/xz/commit/235be8df4f86c943c154112d1abb3c951c86babb

nirs:
The issue was already fixed in xz and we have a new release:
https://github.com/ulikunitz/xz/releases/tag/v0.5.15

So the fix on our side is to update to the new release.

PR #21490, opened 2025-09-04
Update Kubernetes versions

nirs:
We use constants.OldestKubernetesVersion for testing the oldest version and limiting --kubernetes-version when starting the clusters. Our tradition is testing 6 releases back from current version, but we were testing 14 release back (1.20.0).

For upgrading containerd to latest version (v2.1.4) we need to upgrade to a newer release. Upgrade constants.OldestKubernetesVersion to 1.28.0 which seems to pass all tests.

legacyVersion() used in version_upgrade_test.go was 1.26.0. The comment in file mention that this should be release from the last 6 month. We do see failures in the relevant tests (TestRunningBinaryUpgrade) in many builds so I bumped it as well to 1.32.0 (2 releases back from current).

In preload_test.go we tested --kubernetes-version=1.24.4 which is not compatible with containerd v2. Use legacyVersion() instead so we don't need to maintain another version.

We had many example of --kubernetes-version in the docs using older version which are not supported. Replace all example with current version to minimize future maintenance. We need to automated this later so updating the version in minikube will also update the examples.

With this change we have 2 places to update kubernetes versions:
- constants.*KubernetesVersion
- legacyVersion()

### DIFF
--- a/test/integration/preload_test.go
+++ b/test/integration/preload_test.go
@@ -38,8 +38,7 @@ func TestPreload(t *testing.T) {
 
 	startArgs := []string{"start", "-p", profile, "--memory=3072", "--alsologtostderr", "--wait=true", "--preload=false"}
 	startArgs = append(startArgs, StartArgs()...)
-	k8sVersion := "v1.24.4"
-	startArgs = append(startArgs, fmt.Sprintf("--kubernetes-version=%s", k8sVersion))
+	startArgs = append(startArgs, fmt.Sprintf("--kubernetes-version=%s", legacyVersion()))
 
 	rr, err := Run(t, exec.CommandContext(ctx, Target(), startArgs...))
 	if err != nil {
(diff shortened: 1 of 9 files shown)


test/integration/preload_test.go
```diff
@@ -38,8 +38,7 @@ func TestPreload(t *testing.T) {
 
 	startArgs := []string{"start", "-p", profile, "--memory=3072", "--alsologtostderr", "--wait=true", "--preload=false"}
 	startArgs = append(startArgs, StartArgs()...)
-	k8sVersion := "v1.24.4"
-	startArgs = append(startArgs, fmt.Sprintf("--kubernetes-version=%s", k8sVersion))
+	startArgs = append(startArgs, fmt.Sprintf("--kubernetes-version=%s", legacyVersion()))
```
medyagh, line 41: I think this one is a test that we specifically do NOT have a preload for this kubernetes version and we are testing that no preload for this...this might break the test

I might be wrong but I am feeling we chose ""v1.24.4" as one that wont ever get a preload for testing purposes

medyagh (changes requested):

test/integration/preload_test.go
nirs: It may be possible, but we cannot do this. The tested version must be dynamic and change each time we update k8s version, on every release. If this test depend on not having a preload we need to find another way to cause to make preload lookup to fail, or delete the test.

nirs: 1.24.4 is is too old to be compatible with containerd v2. Even 1.28 is not supported.

PR #21499, opened 2025-09-07
Add helper function to generate unique Kubernetes versions for testing

### LINKED ISSUE #21483: test: Remove duplicate kubernetes versions in tests
Some tests use multiple kubernetes versions:

```go
	versions := []string{
		constants.DefaultKubernetesVersion,
		constants.NewestKubernetesVersion,
	}
```

Currently DefaultKubernetesVersion and constants.NewestKubernetesVersion are the same ("1.34.0") so we run the same test twice.

Fix:
- remove duplicate versions in the list

shubhanshu-02:
### What does this PR do?
This PR cleans up Kubernetes version array handling in the integration tests by introducing a helper function (`getKubernetesVersionsForTesting`) in `helpers_test.go`. This function builds a unique list of Kubernetes versions for use in tests, ensuring that DefaultKubernetesVersion and NewestKubernetesVersion are not both added if they are equal. The test `aaa_download_only_test.go` is updated to use this helper.

### Why is this needed?
Previously, duplicate versions could be present in test arrays if DefaultKubernetesVersion and NewestKubernetesVersion are the same, causing redundant test runs. 

### What did you change?
- Added `getKubernetesVersionsForTesting()` to `test/integration/helpers_test.go`.
- Updated `test/integration/aaa_download_only_test.go` to use the new helper.

### Reference
Fixes #21483.


shubhanshu-02:
Hey @nirs, did you check this PR?

test/integration/helpers_test.go
```diff
@@ -704,3 +704,17 @@ func CopyDir(src, dst string) error {
 
 	return nil
 }
+
+// getKubernetesVersionsForTesting returns unique Kubernetes versions to test against
+func getKubernetesVersionsForTesting() []string {
```
nirs: The get prefix is not needed. It should be used if this preform a Get request or something similar. KubernetesVersionsForTesting is good but we are in the context of the tests, so we can remove the ForTesting suffix.

Rename to KubernetesVersions()?

test/integration/aaa_download_only_test.go
```diff
@@ -58,16 +58,7 @@ func TestDownloadOnly(t *testing.T) { // nolint:gocyclo
...
-		constants.NewestKubernetesVersion,
-	}
-
-	// Small optimization, don't run the exact same set of tests twice
-	if constants.DefaultKubernetesVersion == constants.NewestKubernetesVersion {
-		versions = versions[:len(versions)-1]
-	}
+	versions := getKubernetesVersionsForTesting()
 
 	for _, v := range versions {
```
nirs: It seems that we don't use versions, so we can just use:

```go
for _, v := range KubernetesVersions() {
```

test/integration/helpers_test.go
shubhanshu-02: Sure!

test/integration/aaa_download_only_test.go
shubhanshu-02: Made the changes 😄

nirs (commented):
Looks good, but it would be more useful to squash the commits into one, since this is a small and simple change. We can squash the commits during merge, but this create ugly commit message that makes it header to inspect git history later.

nirs:
> Follows up on issue #21483.

This should be:

    Fixes #21483

Which will link this PR to the issue. When we merge the issue will be closed automatically.

nirs (approved):

nirs:
@shubhanshu-02 thanks!

@medyagh merge?

DeepDN (commented):
- Looks good to me

PR #21500, opened 2025-09-07
[don't merge]CI: add gcloud yaml config for minikube prow test

ComradeProgrammer:


upodroid:
Hi, we don't need this image. Please use https://github.com/kubernetes/test-infra/tree/master/images/kubekins-e2e-v2, its used widely in prow. Look at this job as an example. https://github.com/kubernetes/test-infra/blob/b41395b4a2cad696cbf242c984cb8098f6839c43/config/jobs/kubernetes/sig-testing/kubernetes-kind-ci.yaml#L315

PR #21515, opened 2025-09-08
UI: improve config flag long description

developowl:
### comments 🥝
I've addressed the issue where the `Short` and `Long` descriptions for the `configViewCmd` in `view.go` were nearly identical, differing only by a single period. I have improved the `Long` description by adding more specific details about how to use the command, including how to customize the output with the `--format` flag and the typical location of the configuration file. This change helps to clearly differentiate the `Long` description from the `Short` one, making the command's help text more useful for users.
thank you :)


developowl:
@medyagh I mistakenly used double quotes instead of backticks due to `Golang` syntax, which resulted in a build error. 
I've fixed that. I'm sorry..

medyagh:
thank you @developowl

PR #21541, opened 2025-09-11
Pull calico images from quay.io by default

Alphadelta14:
Calico supports quay.io images in its manifests, yet we end up using docker.io (which does happen to be published in its manifests on given tagged branches).
As many are familiar, docker.io rate limits pulls when not authenticated. Authentication is challenging to configure correctly when setting up minikube and registry mirrors are not supported for some container runtimes (like containerd).

The referred manifest is here: https://github.com/projectcalico/calico/blob/master/manifests/calico.yaml.
Tags can be found at https://quay.io/repository/calico/kube-controllers?tab=tags
Related upstream issue: https://github.com/projectcalico/calico/issues/4833


medyagh:
thank you @Alphadelta14 that is a great call ! it is much better to use images that dont get rate limitted in our testinng env

Alphadelta14:
@medyagh no changes, but figured I'd spin tests some more. Currently looking like there is another side for https://github.com/kubernetes/test-infra/pull/35521 that isn't implemented.

PR #21542, opened 2025-09-11
debug: update ssh-keygen algorithm

medyagh:

### DIFF
--- a/cmd/minikube/cmd/cp.go
+++ b/cmd/minikube/cmd/cp.go
@@ -49,9 +49,9 @@ var cpCmd = &cobra.Command{
 	Long: `Copy the specified file into minikube, it will be saved at path <target file absolute path> in your minikube.
 Default target node controlplane and If <source node name> is omitted, It will trying to copy from host.
 
-Example Command : "minikube cp a.txt /home/docker/b.txt" +
-                  "minikube cp a.txt minikube-m02:/home/docker/b.txt"
-                  "minikube cp minikube-m01:a.txt minikube-m02:/home/docker/b.txt"`,
+Example Command : "minikube cp a.txt /root/b.txt" +
+                  "minikube cp a.txt minikube-m02:/root/b.txt"
+                  "minikube cp minikube-m01:a.txt minikube-m02:/root/b.txt"`,
 	Run: func(_ *cobra.Command, args []string) {
 		if len(args) != 2 {
 			exit.Message(reason.Usage, `Please specify the path to copy: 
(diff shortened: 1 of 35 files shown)


cmd/minikube/cmd/mount.go
```diff
@@ -51,7 +51,7 @@ const (
 	nineP                     = "9p"
 	defaultMount9PVersion     = "9p2000.L"
 	mount9PVersionDescription = "Specify the 9p version that the mount should use"
-	defaultMountGID           = "docker"
+       defaultMountGID           = "root"
```
nirs, line 54: Why do we change mount command user? how is this related to changing ssh key algorithm?

cmd/minikube/cmd/cp.go
```diff
@@ -210,6 +210,6 @@ func validateArgs(src, dst *remotePath) {
 	// if node name not explicitly specified in both of source and target,
 	// consider target node is control-plane for backward compatibility.
 	if src.node == "" && dst.node == "" && !strings.HasPrefix(dst.path, "/") {
-		exit.Message(reason.Usage, `Target <remote file path> must be an absolute Path. Relative Path is not allowed (example: "minikube:/home/docker/copied.txt")`)
+           exit.Message(reason.Usage, `Target <remote file path> must be an absolute Path. Relative Path is not allowed (example: "minikube:/root/copied.txt")`)
```
nirs, line 213: Why change the location of the file?

cmd/minikube/cmd/service.go
```diff
@@ -208,7 +208,7 @@ func startKicServiceTunnel(services service.URLs, configName, driverName string)
 			exit.Error(reason.DrvPortForward, "error getting ssh port", err)
 		}
 		sshPort := strconv.Itoa(port)
-		sshKey := filepath.Join(localpath.MiniPath(), "machines", configName, "id_rsa")
+		sshKey := filepath.Join(localpath.MiniPath(), "machines", configName, "id_ed25519")
```
nirs, line 211: This looks fine, but why do we have to hard code the algorithm? why not use the default algorithm on the host? With this code we will have to change again when the default changes next time.

cmd/minikube/cmd/ssh-host.go
```diff
@@ -67,7 +67,7 @@ func appendKnownHelper(nodeName string, appendKnown bool) {
 		}
 	}
 
-	scanArgs := []string{"-t", "rsa"}
+	scanArgs := []string{"-t", "ed25519"}
```
nirs, line 70: We use this in too many places - add a constant so we can have the algorithm exactly once in the project.

deploy/kicbase/Dockerfile
```diff
@@ -262,6 +262,8 @@ RUN echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
 USER docker
 RUN mkdir /home/docker/.ssh
 USER root
+# prepare root ssh directory for minikube access
+RUN mkdir /root/.ssh
```
nirs, line 266: Why it has to be in /root?

nirs (commented):
This seems to do 2 changes:
- use ed25519 instead of rsa - looks good, matching the default algorithm in Fedora and macOS
- Keep the key in /root/.ssh instead of /docker.ssh - seems untreated to this change.
 
It is nice to keep the key in a common user that will never change instead of docker user that seems to be the wrong name for minikube vm user, but not clear why this is mixed with the algorithm change. Mixing unrelated changes make it harder to review and manage, for example reverting the user change if it turns out problematic.

PR #21544, opened 2025-09-13
ci: fix failed ISO not commenting on PR

medyagh:
we were checking exit status of "tee" which is always 0 ... !

### DIFF
--- a/hack/jenkins/build_iso.sh
+++ b/hack/jenkins/build_iso.sh
@@ -65,16 +65,20 @@ else
 	export ISO_BUCKET
 fi
 
-make release-iso | tee iso-logs.txt
-# Abort with error message if above command failed
-ec=$?
-if [ $ec -gt 0 ]; then
-	if [ "$release" = false ]; then
-		gh pr comment ${ghprbPullId} --body "Hi ${ghprbPullAuthorLoginMention}, building a new ISO failed.
-		See the logs at: https://storage.cloud.google.com/minikube-builds/logs/${ghprbPullId}/${ghprbActualCommit::7}/iso_build.txt
-		"
-	fi
-	exit $ec
+if ! make release-iso 2>&1 | tee iso-logs.txt; then
+  # Exit of `make` (PIPESTATUS[0]); fallback to 1 if unavailable
+  ec=${PIPESTATUS[0]:-1}
+
+  # Only comment on non-release; default release=false if unset
+  if [[ ${release:-false} != "true" ]]; then
+    gh pr comment "${ghprbPullId}" --body "$(cat <<'MSG'
+Hi ${ghprbPullAuthorLoginMention}, building a new ISO failed for Commit ${ghprbActualCommit}
+See the logs at:
+https://storage.cloud.google.com/minikube-builds/logs/${ghprbPullId}/${ghprbActualCommit::7}/iso_build.txt
+MSG
+)"
+  fi
+  exit "$ec"
 fi
 
 git config user.name "minikube-bot"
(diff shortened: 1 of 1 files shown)


hack/jenkins/build_iso.sh
```diff
@@ -65,16 +65,20 @@ else
...
-# Abort with error message if above command failed
-ec=$?
-if [ $ec -gt 0 ]; then
-	if [ "$release" = false ]; then
-		gh pr comment ${ghprbPullId} --body "Hi ${ghprbPullAuthorLoginMention}, building a new ISO failed.
-		See the logs at: https://storage.cloud.google.com/minikube-builds/logs/${ghprbPullId}/${ghprbActualCommit::7}/iso_build.txt
-		"
-	fi
-	exit $ec
+if ! make release-iso 2>&1 | tee iso-logs.txt; then
```
nirs, line 68: Why do you redirect stderr to stdout in this change? This looks like unrelated change. This look like a fix for the logs that previously contained only stdout, which makes sense but should not be hidden in another change.

I did not try this but since we use `-o pipefail` the if block looks right.

hack/jenkins/build_iso.sh
```diff
@@ -65,16 +65,20 @@ else
...
+  # Exit of `make` (PIPESTATUS[0]); fallback to 1 if unavailable
+  ec=${PIPESTATUS[0]:-1}
```
nirs, line 70: Do we care about the exit code of the makefile? it will be easier to maintain if we simply exit with exit code 1 in all cases.

hack/jenkins/build_iso.sh
```diff
@@ -65,16 +65,20 @@ else
...
+
+  # Only comment on non-release; default release=false if unset
+  if [[ ${release:-false} != "true" ]]; then
```
nirs, line 73: Preview code was simpler but could be better with quoting.

`#{release:-false}` is not needed since we define the release variable in line 56. If it is not defined in all cases we should fix the issue by always defining it and avoiding expansion like this. This makes the code less readable and promote bad code using undefined variables.

hack/jenkins/build_iso.sh
```diff
@@ -65,16 +65,20 @@ else
...
+    gh pr comment "${ghprbPullId}" --body "$(cat <<'MSG'
+Hi ${ghprbPullAuthorLoginMention}, building a new ISO failed for Commit ${ghprbActualCommit}
+See the logs at:
+https://storage.cloud.google.com/minikube-builds/logs/${ghprbPullId}/${ghprbActualCommit::7}/iso_build.txt
+MSG
+)"
```
nirs: This probably works but hard to follow. It will be better to create the body in a variable and use it in the gh pr call:

```shell
body="$(cat <<'EOF'
Hi ${ghprbPullAuthorLoginMention}, building a new ISO failed for Commit ${ghprbActualCommit}
See the logs at:
https://storage.cloud.google.com/minikube-builds/logs/${ghprbPullId}/${ghprbActualCommit::7}/iso_build.txt
EOF
"

gh pr comment "${ghprbPullId}" --body "$body"
```

hack/jenkins/build_iso.sh
medyagh: i think it only gets defined when it is true, and when it is not a release job, it is not defined

the previous code was not working after 20 tries I found out this is how it would work !
we never set it to false and it was somehow never being triggered

hack/jenkins/build_iso.sh
medyagh: fair enoug

hack/jenkins/build_iso.sh
medyagh: yeah any info is good if you have a suggested better code let me see :) I this is what worked for me

hack/jenkins/build_iso.sh
medyagh: it is just a lot of try and error

PR #21569, opened 2025-09-15
add auto to --log-colors because llama-server will crashloop in ai playground tutorial

volatilemolotov:
Llama-server will crash with
```error while handling argument "--log-colors": expected value for argument

usage:
--log-colors [on|off|auto]              Set colored logging ('on', 'off', or 'auto', default: 'auto')
                                        'auto' enables colors when output is to a terminal
                                        (env: LLAMA_LOG_COLORS)
```

Adding `auto` makes the command valid

### DIFF
--- a/site/content/en/docs/tutorials/ai-playground.md
+++ b/site/content/en/docs/tutorials/ai-playground.md
@@ -210,6 +210,7 @@ spec:
           --threads, "6",
           --no-warmup,
           --log-colors,
+          auto,
         ]
         resources:
           limits:
@@ -292,6 +293,7 @@ spec:
           --threads, "6",
           --no-warmup,
           --log-colors,
+          auto,
         ]
         resources:
           limits:


nirs:
@volatilemolotov we tested the tutorial few 1-2 weeks ago and --log-colors worked. Can you share the ramalama image details you are using?

```
minikube image ls --format table
```

site/content/en/docs/tutorials/ai-playground.md
```diff
@@ -210,6 +210,7 @@ spec:
           --threads, "6",
           --no-warmup,
           --log-colors,
+          auto,
```
nirs, line 213: I'm not sure auto is needed (since they have a default value), but maybe the behavior changed since we tested it. However all other options use `--key, value,` style, let's keep this style.

nirs:
Reproduced with current ramalama image:

```
% kubectl logs deploy/granite
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = Virtio-GPU Venus (Apple M2 Max) (venus) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 32 | shared memory: 32768 | int dot: 0 | matrix cores: none
error while handling argument "--log-colors": expected value for argument

usage:
--log-colors [on|off|auto]              Set colored logging ('on', 'off', or 'auto', default: 'auto')
                                        'auto' enables colors when output is to a terminal
                                        (env: LLAMA_LOG_COLORS)
% minikube image ls --format table | grep ramalama
│ quay.io/ramalama/ramalama               │ latest  │ e1d309388e19d │ 1.57GB │
```

PR #21577, opened 2025-09-15
test: No auto update aux drivers in integration tests

medyagh:

### DIFF
--- a/test/integration/main_test.go
+++ b/test/integration/main_test.go
@@ -140,6 +140,11 @@ func HyperVDriver() bool {
 	return strings.Contains(*startArgs, "--driver=hyperv") || strings.Contains(*startArgs, "--vm-driver=hyperv")
 }
 
+// KVM returns true is is KVM driver
+func KVMDriver() bool {
+	return strings.Contains(*startArgs, "--driver=kvm") || strings.Contains(*startArgs, "--vm-driver=kvm") || strings.Contains(*startArgs, "--driver=kvm2") || strings.Contains(*startArgs, "--vm-driver=kvm2")
+}
+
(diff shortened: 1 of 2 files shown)


test/integration/main_test.go
```diff
@@ -57,7 +57,9 @@ const (
 func TestMain(m *testing.M) {
 	flag.Parse()
 	setMaxParallelism()
-
+	if NeedsAuxDriver() {
+		*startArgs += " --auto-update-drivers=false"
```
nirs, line 61: It makes sense but it can break tests for auto update.

test/integration/main_test.go
```diff
@@ -140,6 +142,11 @@ func HyperVDriver() bool {
 	return strings.Contains(*startArgs, "--driver=hyperv") || strings.Contains(*startArgs, "--vm-driver=hyperv")
 }
 
+// KVM returns true is is KVM driver
+func KVMDriver() bool {
+	return strings.Contains(*startArgs, "--driver=kvm") || strings.Contains(*startArgs, "--vm-driver=kvm") || strings.Contains(*startArgs, "--driver=kvm2") || strings.Contains(*startArgs, "--vm-driver=kvm2")
```
nirs, line 147: I know we use this for other drivers, but this will not work with "--driver kvm". We cannot fix this now, but we should change the way we run the tests, test configuration should come from a configuration file:

    $ cat test.yaml
    configs:
    - name: kvm
      os: linux
      driver: kvm
      auto-update-drivers: false
    - name: vfkit
      os: darwin
      driver: vfkit
      network: vmnet-shared  

    $ make integration CONFIG=kvm

medyagh: good call

PR #21590, opened 2025-09-17
Detect and log Docker Hub rate limit errors in image pulls

### LINKED ISSUE #21442: integration tests: detect and Log Docker Hub Rate limit Failures
it should Log something simple instead of we search through logs for the docker rate limit

Example this Gopogh Report
https://storage.googleapis.com/minikube-builds/logs/21439/41161/Docker_Linux_containerd.html

```
Warning  Failed     7m14s (x2 over 7m57s)   kubelet            Failed to pull image "docker.io/nginx:alpine": failed to pull and unpack image "docker.io/library/nginx:alpine": failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/library/nginx/manifests/sha256:60e48a050b6408d0c5dd59b98b6e36bf0937a0bbe99304e3e9c0e63b7563443a: 429 Too Many Requests - Server message: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
```

also there is probably a WAY to check how many Docker API calls we have left before Banned, so we can log that in the beggninign of the test and potentially even Wait for it before running the test and wassting time

henry3260:
This PR improves error visibility when pulling images fails due to Docker Hub rate limits.

### Changes
- Updated `PullImages` in `cache_images.go` to detect `429 Too Many Requests` errors and return them to the caller.
- Added a warning log with a helpful link: https://www.docker.com/increase-rate-limit
- Ensures users are clearly informed when image pulls fail due to Docker Hub rate limiting.

### Impact
- Users will now see a clear log message if image pulls fail because of Docker Hub rate limits, instead of only a generic failure message.
- Callers of `PullImages` can now handle the `429` error explicitly.

### Testing
- Added a unit test in `cache_image_test.go` that simulates a `429 Too Many Requests` error in `pullImages`.
- Verified the following log output appears:

Docker Hub rate limit reached for profile minikube. See: https://www.docker.com/increase-rate-limit

### Before: 
image

### After: 
image

Fixes: #21442


pkg/minikube/machine/cache_images.go
```diff
@@ -596,6 +596,10 @@ func PullImages(images []string, profile *config.Profile) error {
 			err = pullImages(crMgr, images)
 			if err != nil {
 				failed = append(failed, m)
+				// Rate limit check
+				if strings.Contains(err.Error(), "429 Too Many Requests") {
+					klog.Warningf("Docker Hub rate limit reached for profile %s. See: https://www.docker.com/increase-rate-limit", pName)
```
medyagh: we should use 
out.Styled(style.Notice, `Kicbase images have not been deleted. To delete images run:`)

I supsect there are other places we could that

do you putting a Before/Aeter this PR, 

you can simulate the rate limit by making a Dummy docker binary that always returns  with rate limit...
howevever we also have pulling images using go library , you mock those libraries for full coverage of everywhere we pull image


https://github.com/google/go-containerregistry

medyagh (changes requested):

pkg/minikube/machine/cache_images.go
medyagh: also in our addons we need to check if addons which docker images if they are rate limitted or not.

so we should add code to where we enable the addons (and verify that the image is not rate limitted) and if it is we should tell the users nicely

pkg/minikube/machine/cache_images.go
```diff
@@ -596,6 +596,10 @@ func PullImages(images []string, profile *config.Profile) error {
 			err = pullImages(crMgr, images)
 			if err != nil {
 				failed = append(failed, m)
+				// Rate limit check
+				if strings.Contains(err.Error(), "429 Too Many Requests") {
```
nirs: Using strings.Contains() means we are not using a public API. This is too fragile to depend on. Do we have an error code field or error name we can check instead?

pkg/minikube/machine/cache_images.go
henry3260: OK, thanks for the clarification!  
For this PR, I can switch the logging in `cache_images.go` from `klog` to `out.Styled(style.Notice, …)` where appropriate.  
Do you suggest I update all `klog` calls in that file, or just the ones related to rate limit messages?  

Also, I’ll look into adding tests by mocking docker/go-containerregistry to simulate the rate limit scenario.

henry3260: Got it!  
So for addons, I’ll add a check when enabling them to verify the image is not rate-limited, and if it is, we’ll surface a friendly notice to the user.

pkg/minikube/machine/cache_images.go
henry3260: > Using `strings.Contains()` means we are not using a public API.  
> This is too fragile to depend on.  
> Do we have an error code field or error name we can check instead?  

Thanks for the feedback! 🙏  
I agree that relying on `strings.Contains()` is fragile.  

In my opinion, we could consider using [go-containerregistry](https://github.com/google/go-containerregistry), since its API exposes structured errors. For example:

```go
if terr, ok := err.(*transport.Error); ok {
    if terr.StatusCode == http.StatusTooManyRequests {
        // This confirms it's a 429 rate limit error
    }
}
```
This way we can explicitly check the HTTP status code instead of depending on error strings.
Do you think this approach would be acceptable for minikube?

nirs: This looks good, this is way to handle errors.

pkg/minikube/machine/cache_images.go
henry3260: > we should use out.Styled(style.Notice, `Kicbase images have not been deleted. To delete images run:`)
> 
> I supsect there are other places we could that
> 
> do you putting a Before/Aeter this PR,
> 
> you can simulate the rate limit by making a Dummy docker binary that always returns with rate limit... howevever we also have pulling images using go library , you mock those libraries for full coverage of everywhere we pull image
> 
> https://github.com/google/go-containerregistry

Thanks for the suggestion! I'm not sure how to properly mock `go-containerregistry`'s `remote.Image` in our current test setup. Could you give me a pointer or a small example of how you'd recommend doing that?

So far, I tried:
- Using a dummy Docker binary (which worked for simulating errors).
- Replacing the function with a variable to trigger the error path, but it didn't enter the `if` block as expected.

At this point, I'm not sure how to apply `go-containerregistry` to simulate the rate-limit case in tests. Any guidance or example would be really helpful.

If you'd prefer, I can open a follow-up issue for this discussion.

nirs:
@henry3260 can you rebase to resolve the conflicts?

henry3260:
> @henry3260 can you rebase to resolve the conflicts?

Is there anything I can do to help get it merged?

nirs:
> > @henry3260 can you rebase to resolve the conflicts?
> 
> Is there anything I can do to help get it merged?

Yes, rebase on master and resolve the conflicts reported by GitHub:

Screenshot 2025-12-08 at 20 52 56

Otherwise I think this need a review. Several reviewers are added, need to wait until they have time to review this.

Issue #21582, opened 2025-09-16
Running functional or integration tests should use the driver built from source instead of the downloaded driver


obnoxxx:
Third spin-off of #21551 

currently, the functional and integration test runs use the downloaded driver instead of one locally built from sources.

This is a possible cause of  deceiving false positive test results.

Just like for manual invocation (see #21581 ), the test runs should use the locally built driver instead of a  downloaded one.

depends on #21580 as the default build should first produce the driver to be tested.

/kind bug

medyagh:
@obnoxxx I think you created three issues plz close other ones, and do you mind checking if this has fixed the issue ? (verify)

https://github.com/kubernetes/minikube/pull/21577

obnoxxx:
@medyagh  wrote:

> [@obnoxxx](https://github.com/obnoxxx) I think you created three issues 
> plz close other ones,

I am not sure why only opening of a single issue is alllowed.

All three issues serve different purposes as sub-aspects of a different issue:

https://github.com/kubernetes/minikube/issues/21551#issuecomment-3297720881

In fact, two more issues would be needed to complete the tracking list:

https://github.com/kubernetes/minikube/issues/21551#issuecomment-3293184588







> and do you mind checking if this has fixed the issue ? (verify)
> [#21577](https://github.com/kubernetes/minikube/pull/21577)


I am testing and will report back.

obnoxxx:
> > and do you mind checking if this has fixed the issue ? (verify)
> > [#21577](https://github.com/kubernetes/minikube/pull/21577)
> 
> I am testing and will report back.

@medyagh I do in fact have a question how to properly test this, i. e. how can I tell
definitively whether a driver in `˜/.minikube/bin` is downloaded or a. local build?

I guess looking at the timestamp is not sufficient,  right?

medyagh:
MINIKUBE_HOME=/tmp/test1

"make functional"


check the logs
docker-machine-kvm... --version

obnoxxx:
@medyagh  wrote:

> MINIKUBE_HOME=/tmp/test1
> 
> "make functional"
> 
> check the logs docker-machine-kvm... --version

I tried this  but I am not quite clear yet about the conclusion:

I ran the following from a current master checkout:

````console
$ ls -l $HOME/.minikube/bin/docker-machine-driver-kvm2 
[...]

PR #21594, opened 2025-09-19
Require valid aux driver version

nirs:
This change makes driver version checking more strict. Previously we failed only if the driver was not found or running "driver version" failed. For all other errors we logged a warning and use the currently installed driver. With this change all the errors are fatal and you must have a valid driver to use the kvm or hyperkit driver.

Checking the driver version is more correct and strict. We separate driver version stdout and stderr, parse the driver output using yaml parser, and have more detailed logging and error messages.

The tests for extracting driver version from "driver version" output were replaced with test running driver version command and parsing the version.

This change makes it easy to validate the driver commit hash for addressing #21582.

Based on #21597 for testing to avoid the vfkit test failures.


nirs:
## Testing with function tests

Started functional tests in one shell

```
env MINIKUBE_HOME=/tmp TEST_ARGS="'-minikube-start-args=--driver=kvm'" make functional
```

Looking at the logs in another shell:

```
$ MINIKUBE_HOME=/tmp minikube logs
...
I0920 00:57:34.743726  141622 start.go:304] selected driver: kvm2
I0920 00:57:34.743738  141622 start.go:918] validating driver "kvm2" against <nil>
I0920 00:57:34.743765  141622 start.go:929] status for kvm2: {Installed:true Healthy:true Running:true NeedsImprovement:false Error:<nil> Reason: Fix: Doc: Version:}
I0920 00:57:34.746915  141622 install.go:51] acquiring lock: {Name:mk900956b073697a4aa6c80a27c6bb0742a99a53 Clock:{} Delay:500ms Timeout:10m0s Cancel:<nil>}
I0920 00:57:34.747205  141622 install.go:123] Validating docker-machine-driver-kvm2, PATH=/tmp/.minikube/bin:/home/nsoffer/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/bin:/home/nsoffer/bin:/home/nsoffer/.krew/bin:/home/nsoffer/sdk/go1.23.2/bin:/home/nsoffer/go/bin:/home/nsoffer/.local/bin:/home/nsoffer/bin:/usr/local/bin:/usr/bin
W0920 00:57:34.747320  141622 install.go:61] docker-machine-driver-kvm2: failed to find driver "docker-machine-driver-kvm2": exec: "docker-machine-driver-kvm2": executable file not found in $PATH
I0920 00:57:34.747443  141622 out.go:179] * Downloading driver docker-machine-driver-kvm2:
I0920 00:57:34.747571  141622 download.go:108] Downloading: https://github.com/kubernetes/minikube/releases/download/v1.37.0/docker-machine-driver-kvm2-amd64?checksum=file:https://github.com/kubernetes/minikube/releases/download/v1.37.0/docker-machine-driver-kvm2-amd64.sha256 -> /tmp/.minikube/bin/docker-machine-driver-kvm2
I0920 00:57:38.135430  141622 install.go:123] Validating docker-machine-driver-kvm2, PATH=/tmp/.minikube/bin:/home/nsoffer/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/bin:/home/nsoffer/bin:/home/nsoffer/.krew/bin:/home/nsoffer/sdk/go1.23.2/bin:/home/nsoffer/go/bin:/home/nsoffer/.local/bin:/home/nsoffer/bin:/usr/local/bin:/usr/bin
I0920 00:57:38.166004  141622 install.go:134] /tmp/.minikube/bin/docker-machine-driver-kvm2 version is {Version:v1.37.0 Commit:1af8bdc072232de4b1fec3b6cc0e8337e118bc83}
```

We used:

```
out/minikube start -p functional-945141 --memory=4096 --apiserver-port=8441 --wait=all --driver=kvm --auto-update-drivers=false
```

But the driver was downloaded during the test!

```
$ tree /tmp/.minikube/bin/
/tmp/.minikube/bin/
└── docker-machine-driver-kvm2
```

I would expect that with --auto-update-drivers=false the driver will not be downloaded and the tests will fail very early.

This seems to be the issue - we install the driver if it does not exist even if autoUpdate is false:

```go
	if !exists || (err != nil && autoUpdate) {
		klog.Warningf("%s: %v", executable, err)
		path = filepath.Join(directory, executable)
		if err := download.Driver(executable, path, v); err != nil {
			return err
		}
	}
```

The driver must not be installed when autoUpdate is false.

cmd/minikube/cmd/start.go
```diff
@@ -534,13 +534,7 @@ func updateDriver(driverName string) {
...
 		out.WarningT("Error parsing minikube version: {{.error}}", out.V{"error": err})
 	} else if err := auxdriver.InstallOrUpdate(driverName, localpath.MakeMiniPath("bin"), v, viper.GetBool(interactive), viper.GetBool(autoUpdate)); err != nil {
-		if errors.Is(err, auxdriver.ErrAuxDriverVersionCommandFailed) {
-			exit.Error(reason.DrvAuxNotHealthy, "Aux driver "+driverName, err)
-		}
-		if errors.Is(err, auxdriver.ErrAuxDriverVersionNotinPath) {
-			exit.Error(reason.DrvAuxNotHealthy, "Aux driver"+driverName, err)
-		} //if failed to update but not a fatal error, log it and continue (old version might still work)
-		out.WarningT("Unable to update {{.driver}} driver: {{.error}}", out.V{"driver": driverName, "error": err})
+		exit.Error(reason.DrvAuxNotHealthy, fmt.Sprintf("Auxiliary driver %q", driverName), err)
```
medyagh, line 537: we barely ever make a new kvm aux driver, may releases can just work with old one,
plus many Embedded users (like cloud code users for example) download minikube aux drivers ones, and some of them need root (for hyperkit) and they Prefer to give this root permission once to that binary and not be forced to get a new aux version on every single minikube update.

it is too much to exit when we can run. warning is fine. no need to disrupt ppl when not needed.

nirs: The kvm driver is built from minikube source. The docker-machine-driver-kvm2 is just a small wrapper around minikube code. It does not make sense to run an old driver from previous release with new minikube since we don't know if it will work or not. Hoping that an old driver will work is a not a valid release engineering strategy.

On system with a package manager (dnf, apt) this driver should be installed by the package manager in the standard location in the same way minikube is installed. This will ensure that we always run with the right driver tested by the CI. There is no need to install a driver dynamically when software is managed by a package manager.

When minikube is installed by downloading the minikube executable, installing the driver on the first use is a nice feature, but we want to make sure the driver is the driver released with minikube.

If the driver version command does not produce the expected output, something is very wrong and it is better to fail early and loudly, helping the user to fix the issue with minimal debugging.

If you have an old driver and minkube fail to download the driver, the user can download it manually in the same way they downloaded minikube itself. If we think that downloading the driver is not reliable enough, we can provide a tarball with minikube and the driver to avoid the unreliable automatic install. If we think the automatic install is reliable enough, there is no reason to continue if minikube cannot install the right driver.

Hyperkit is depcated and should be removed in minikube 1.38 (#21601) so we can ignore it.

afbjorklund: You could consider converting the `kvm` driver from an external driver into an internal driver, then when `hyperkit` is gone there wouldn't be any more "aux" drivers and you wouldn't need the extra build... The `virsh` command is already required by the driver, and most communication with libvirt is done through XML anyway...

That is, replace the `libvirt-go` calls with the matching exec.Command - like with all other minikube drivers

nirs: Using virsh instead of the libvirt api is not great since it is not designed for machines. But given the complexity and trouble caused by auxiliary drivers it sounds like a good plan.

afbjorklund: * https://github.com/kubernetes/minikube/issues/21618

nirs:
I think this is the wrong direction - fixing the aux driver is hard, and we don't have a reason to use aux driver for libvirt. I'll work instead on making the #21618.

nirs:
Replaced by #21625

PR #21599, opened 2025-09-19
set sysctl params for inotify, to avoid 'too many open files' errors

prezha:
we noticed that [TestMultiControlPlane/serial/AddWorkerNode](https://gopogh-server-tts3vkcpgq-uc.a.run.app/?env=Docker_Linux&test=TestMultiControlPlane/serial/AddWorkerNode) has started failing persistently since we switched the ci runners to use ubuntu 22.04 ([gopogh flake chart](https://gopogh-server-tts3vkcpgq-uc.a.run.app/?env=Docker_Linux&test=TestMultiControlPlane/serial/AddWorkerNode))

example ci job log: https://storage.googleapis.com/minikube-builds/logs/master/41465/Docker_Linux.html

the issue is: `Failed to create control group inotify object: Too many open files`, and this pr sets the sysctl params for inotify on the ci runner, to avoid these errors

### DIFF
--- a/hack/jenkins/common.sh
+++ b/hack/jenkins/common.sh
@@ -89,6 +89,13 @@ if [ "$(uname)" = "Darwin" ]; then
   fi
 fi
 
+## set sysctl params for inotify, to avoid "too many open files" errors
+## ref: https://cloud.google.com/kubernetes-engine/distributed-cloud/bare-metal/docs/installing/configure-os/ubuntu
+if [ "$OS" == "linux" ]; then
+  sudo sysctl -w fs.inotify.max_user_instances=8192
+  sudo sysctl -w fs.inotify.max_user_watches=524288
+fi
+
 # We need pstree for the restart cronjobs
 if [ "$(uname)" != "Darwin" ]; then
   sudo apt-get -y install lsof psmisc dnsutils


hack/jenkins/common.sh
```diff
@@ -89,6 +89,27 @@ if [ "$(uname)" = "Darwin" ]; then
   fi
 fi
 
+## set sysctl params for inotify, to avoid "too many open files" errors
+## ref: https://cloud.google.com/kubernetes-engine/distributed-cloud/bare-metal/docs/installing/configure-os/ubuntu
+if [ "$OS" == "linux" ]; then
```
medyagh, line 94: I think that should be "Linux"

medyagh (changes requested):

hack/jenkins/common.sh
```diff
@@ -89,6 +89,27 @@ if [ "$(uname)" = "Darwin" ]; then
...
+  SYSCTL=/etc/sysctl.conf
+
+  MAX_USER_INSTANCES=8192
+  MAX_USER_WATCHES=524288
+
+  if grep -q "^[[:blank:]]*fs.inotify.max_user_instances=" "${SYSCTL}"; then
+    sed -i -E "s/fs.inotify.max_user_instances=.*/fs.inotify.max_user_instances=${MAX_USER_INSTANCES}/g" "${SYSCTL}"
```
nirs: If we set the value it overrides what we have in /etc/sysctl.conf, so we don't really care about it.

This is also not correct since we can have drop in configs in SYSCTL=/etc/sysctl.conf.d. It does not make sense to look in all of them for the settings and replacing the configuration in files created by other components is not a good idea.

Typically you add a new config file with higher number so it would be processed last and override other configs like /etc/sysctl.conf/99-override.conf.

nirs:
Add an issue to upgrade the arm64 hosts to Ubuntu 22.04 (#21602)

PR #21603, opened 2025-09-20
Remove the deprecated hyperkit driver

### LINKED ISSUE #21601: Remove the hyperkit driver
The hyperkit driver does not support Apple silicon machines introduced in June 2020. There are no plans to add support for Apple silicon and it was replaced by the virtualization framework in the docker project.

Building hyperkit requires the xcgo container with is not maintained for several years, and has many vulnerabilities (#21012).

The docker-machine-driver-hyperkit requires root permissions. Currently minikube add the setuid bit to make this driver runable by root by any user, obtaining user password with sudo. This is insecure and wrong in many ways.

The minikube project does not have enough contributors to support legacy drivers. We want to spend time on making modern drivers better, providing value to most users.

The driver is mark deprecated in minikube 1.37 and we want to remove it in minikube 1.38.

macOS users with intel based machines can use other drivers: vfkit (preferred), qemu, or docker. macOS user with Apple silicon machine can also use the new krunkit driver (experimental).

nirs:
The hyperkit driver does not support Apple silicon machines introduced
in June 2020. There are no plans to add support for Apple silicon and it
was replaced by the virtualization framework in the docker project.

Building hyperkit requires the xcgo container with is not maintained for 
several years, and has many vulnerabilities.

The docker-machine-driver-hyperkit requires root permissions. Currently
minikube add the setuid bit to make this driver runnable by root by any 
user, obtaining user password with sudo. This is insecure and wrong in
many ways.

The minikube project does not have enough contributors to support legacy
drivers. We want to spend time on making modern drivers better,
providing value to most users.

The driver is mark deprecated in minikube 1.37 and we want to remove it
in minikube 1.38.

macOS users with intel based machines can use other drivers: vfkit                                                                                                                         
(preferred), qemu, or docker. macOS user with Apple silicon machine can 
also use the new krunkit driver (experimental).

Fixes #21601


medyagh (changes requested):
I dont think there is urgency in removing this driver, we can continue warn and also remove it from being listed or advertised.
I actually do have mac device from 2010 and it still running and I do like to run minikube on it once in a while. so I dont think there is an urgency in removing it, since there are still x86 macbooks.

nirs:
> I dont think there is urgency in removing this driver, we can continue warn and also remove it from being listed or advertised. I actually do have mac device from 2010 and it still running and I do like to run minikube on it once in a while. so I dont think there is an urgency in removing it, since there are still x86 macbooks.

Keeping the driver make it harder to work on minikube. We are slowed down by having to deal with the hacks added to support it.

macOS 26 is the last version supporting intel macs. We don’t have the capacity to support older machine that Apple does not.

nirs:
Replaced by #23320

Issue #20887, opened 2025-06-04
[Parent] migration of "minikube  infrastructure"  to public


medyagh:
# Please See Sub Issues

Currently minikube uses internal to google GCP project to build/test/release minikube, this issue tracks the requirements needs and related issues in one place

- Requirement 1: Ability Build ISOs on the PR and test the ISO on multiple environments (example Linux KVM, Windows HyperV, Qemu Macos...) and Host the ISO after merged to be tested on the head 
- Requirement 2: Ability to build/push docker images PR PR
- Requirement 3: Test Linux VM with support for Nested Virtualizations and host build artifacts. 
- Requirement 4: Host Minikube Binaries/Driver  for released miniube versions
- Requirement 5: Ability to host HTTP server with Cloud SQL (for generating flake charts/HTML readable) [Example Gopogh Server](https://gopogh-server-tts3vkcpgq-uc.a.run.app/?env=KVM_Linux) and [Triage-Party](https://teaparty-tts3vkcpgq-uc.a.run.app/s/daily-triage) 
- Requirement 6: Host Tutorial Material (Example Kubernetes 101 https://github.com/kubernetes/minikube/issues/20886)  , Essential components (storage provisioner), Maintained Addons Images (auto-pause-hook, gvisor, gcp-auth webhook),   Required tools abandoned by Kubernetes  (example "kube-registry-proxy"), 

Examples:

```
 auto-pause-hook	—	Aug 27, 2024	Aug 27, 2024	
 gcp-auth-webhook	—	Aug 27, 2024	Oct 10, 2024	
 gvisor-addon	—	Aug 27, 2024	Oct 9, 2024	
 kicbase	—	Aug 27, 2024	May 21, 2025	
 kicbase-builds	—	Aug 27, 2024	7 hours ago	
 kube-registry-proxy	—	Aug 27, 2024	11 days ago	
 kubernetes-bootcamp	—	Aug 27, 2024	Apr 14, 2025	
 kvm-build-image_arm64	—	Oct 18, 2024	Oct 18, 2024	
 minikube-ingress-dns	—	Aug 27, 2024	Apr 14, 2025	
 storage-provisioner	—	Aug 27, 2024	Apr 14, 2025	
 triage-party	—	Aug 27, 2024	Apr 14, 2025	
 xcgo
```

medyagh:
I am going to request a GCP project from CNCF
https://github.com/cncf/cluster/issues/282

medyagh:
[...]

PR #21607, opened 2025-09-20
run the initial set of e2e prow job

upodroid:
Part of #20887

### DIFF
--- a/tests/e2e/conformance.sh
+++ b/tests/e2e/conformance.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+OS=$(go env GOOS)
+ARCH=$(go env GOARCH)
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+make # build the binary first
+LATEST_RELEASE=$(curl -sSfL https://dl.k8s.io/release/stable.txt)
+"${REPO_ROOT}"/out/minikube start --nodes=2 --driver=docker --cpus=no-limit --memory=no-limit --force --kubernetes-version=$LATEST_RELEASE
+
+kubetest2-tester-ginkgo --test-package-marker stable.txt \
+    --focus-regex='\[Conformance\]'
--- a/tests/e2e/e2e.mk
+++ b/tests/e2e/e2e.mk
@@ -1,0 +1,12 @@
+GOOS ?= $(shell go env GOOS)
+GOARCH ?= $(shell go env GOARCH)
+
+.PHONY: minikube-e2e-fast
+minikube-e2e-fast:
+	./tests/e2e/fast.sh
+
+minikube-e2e-conformance:
+	./tests/e2e/conformance.sh
+
+integration-kvm-prow: #temp for prow
+	./tests/e2e/fast.sh
--- a/tests/e2e/fast.sh
+++ b/tests/e2e/fast.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+OS=$(go env GOOS)
+ARCH=$(go env GOARCH)
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+make # build the binary first
+LATEST_RELEASE=$(curl -sSfL https://dl.k8s.io/release/stable.txt)
+"${REPO_ROOT}"/out/minikube start --nodes=2 --driver=docker --cpus=no-limit --memory=no-limit --force --kubernetes-version=$LATEST_RELEASE
+
+kubetest2-tester-ginkgo --test-package-marker stable.txt \
+    --parallel=30 \
+    --skip-regex='\[Driver:.gcepd\]|\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]'
--- a/Makefile
+++ b/Makefile
@@ -1260,3 +1260,5 @@ get-dependency-version:
 .PHONY: _update-all
 _update-all:
 	@(cd hack && go run update/update_all/update_all.go)
+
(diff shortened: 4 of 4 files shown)


tests/e2e/conformance.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+OS=$(go env GOOS)
+ARCH=$(go env GOARCH)
+REPO_ROOT="$(git rev-parse --show-toplevel)"
```
upodroid, line 4: This script replaces https://github.com/kubernetes/minikube/blob/master/hack/conformance_tests.sh and we'll add minikube conformance(HEAD + supported branches) jobs to the main conformance board on testgrid at https://testgrid.k8s.io/conformance

tests/e2e/e2e.mk
```diff
@@ -0,0 +1,12 @@
+GOOS ?= $(shell go env GOOS)
+GOARCH ?= $(shell go env GOARCH)
+
+.PHONY: minikube-e2e-fast
+minikube-e2e-fast:
```
medyagh, line 5: to avoid confusion how about name this folder test/conformance

since we already have test/integration that we also call it e2e
and we have "functional test" and "smoke test"

so I suggest using a name that makes it not overlap with others

medyagh (changes requested):

tests/e2e/e2e.mk
upodroid: I'm thinking of splitting the tests like this:

https://github.com/kubernetes/kops/tree/master/tests

- integration is getting a successful minikube cluster up and running on all the permutations of arch/os/driver. Its purely specific to minikube
- e2e test is running the kubernetes test suite on a combination of minikube setups with various kubernetes versions
   - minikube defaults with conformance + fast suite
   - different CNI/Addons/hardware combinations against the fast suite
- smoke test which is compile minikube + launch it on docker, a base test scenario that runs on presubmits all the time

I looked at common.sh and that harness can't be used in Prow.  https://github.com/kubernetes/minikube/tree/master/hack/jenkins This needs to be removed entirely so lets start it from scratch using the sig-testing toolchain.

upodroid:
The fast test suite is failing for the following reasons:
- FeatureGate:InPlacePodVerticalScaling is turned off
- conntrack config in kube-proxy needs to be tweaked

tests/e2e/e2e.mk
medyagh: we call our integration tests e2e in makefiles https://github.com/kubernetes/minikube/blob/b374745bd42da46be83c953c9d1aaa6c95a192fc/Makefile#L288

the e2e integration tests does also include testing different kubernetes versions. but I do like having , so for not making more confusing I suggest limiting this test to only confromacne test

tests/e2e/fast.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+OS=$(go env GOOS)
+ARCH=$(go env GOARCH)
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+make # build the binary first
+LATEST_RELEASE=$(curl -sSfL https://dl.k8s.io/release/stable.txt)
+"${REPO_ROOT}"/out/minikube start --nodes=2 --driver=docker --cpus=no-limit --memory=no-limit --force --kubernetes-version=$LATEST_RELEASE
```
medyagh, line 7: are two nodes required for this test "kubetest2-tester-ginkgo " ?if yes please comment.

tests/e2e/e2e.mk
medyagh: you can name the folder anything else and it would be better
for example
conformance, kubetest-e2e, ginkgo-e2e...

tests/e2e/conformance.sh
medyagh: thank you ! btw I am curious does this has to be shell script ? does prow accept go scripts ?

we try to convert as much as scripts to golang as possible

tests/e2e/e2e.mk
upodroid: > we call our integration tests e2e in makefiles
> 
> https://github.com/kubernetes/minikube/blob/b374745bd42da46be83c953c9d1aaa6c95a192fc/Makefile#L288
> 
> the e2e integration tests does also include testing different kubernetes versions. but I do like having , so for not making more confusing I suggest limiting this test to only confromacne test

I looked at the tests defined at https://github.com/kubernetes/minikube/pull/21607#discussion_r2369682245 and they are not kubernetes e2e tests. The kubernetes e2e tests are defined over [here](https://github.com/kubernetes/kubernetes/tree/master/test/e2e). Minikube needs to run these tests to verify kubernetes functionality. The default minikube config is already failing the simple e2e scenario we run.

tests/e2e/fast.sh
upodroid: Yes, the default cluster size for testing is 1 control plane node and 3 worker nodes. 2 workers for kind e2e tests

https://testgrid.k8s.io/sig-release-master-blocking#kind-master
https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-default

tests/e2e/conformance.sh
upodroid: We can write a kubetest2-minikube-deployer similar to this https://github.com/kubernetes/kops/tree/master/tests/e2e/kubetest2-kops
 and its invoked like this: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/kops/kops-presubmits-e2e.yaml#L348

PR #21624, opened 2025-09-23
UI: Added Ukrainian language translation

Andygol:
This PR proposes a Ukrainian locale for minikube. You can see the locale in action here 👇.

[asciicast](https://asciinema.org/a/AHFKwHCbHB2vo8fD7Sk2ybRzO)


Andygol:
@medyagh TY for approving and merging this PR. 🙇‍♂️ 

I would be very grateful if you could tell when we can expect the next release that will include this translation.

Issue #21626, opened 2025-09-24
Not all strings extracted to "strings.txt" and added to respective language *.json files


Andygol:
While working on the [Ukrainian translation of the minikube UI](https://github.com/kubernetes/minikube/pull/21624), I encountered an issue: that not all strings that need to be localized are present in [strings.txt](https://github.com/kubernetes/minikube/blob/4d64786b338cdd1b1e328e98c66de9f93e6a6c30/translations/strings.txt) and the corresponding language *.json files. Some of them can be added to *.json manually, for example, as shown here.

https://github.com/kubernetes/minikube/blob/71227b25cd999e4333dc9e85b986d6fb50b350cf/translations/de.json#L48

source: 

https://github.com/kubernetes/minikube/blob/71227b25cd999e4333dc9e85b986d6fb50b350cf/cmd/minikube/cmd/start_flags.go#L220-L224

I would be very grateful if "make extract" would extract all strings that are subject to localization.

Andygol:
One possible reason is that the text is not properly decorated for translation, and the import section is missing `"k8s.io/minikube/pkg/minikube/translate"`.  

Many files do not contain a "minikube/translate" package in the import section, and not all strings are formatted for further translation and use.

PR #21629, opened 2025-09-24
UI: Manually added some lines for Ukrainian translation

Andygol:
Some lines with Ukrainian translations that were not extracted by "make extract" have been added manually.

RE Issue: #21626


medyagh (changes requested):
@Andygol are you sure make extract is missing this ? this is part of automation and we need to fix the automation.
because if make extract is doing wrong, then it will overwrite the manual added strings

Andygol:
> @Andygol are you sure make extract is missing this ?

Yes, I do. I run `make clean && make extract && make` and than `out/minikube <something>` to visually check if something really changed.

> then it will overwrite the manual added strings

@medyagh As far as I know, make extract does not overwrite existing lines.

UPD. I found a similar approach in other languages as well, when lines were added manually. (see Issue #21626)

Andygol:
@kubernetes/minikube-maintainers  Can we proceed with this PR for now, while the issue https://github.com/kubernetes/minikube/issues/21626 is being resolved? This PR contains only a few lines of translation (including typo fixes), which can be restored fairly quickly if the updated process for extracting lines for translation is implemented and these changes are removed.

PR #21631, opened 2025-09-24
Deprecated VirtualBox with warning

### LINKED ISSUE #21608: Marking VirtualBox as deprecated
Currently VirtualBox is marked as registry.Fallback  and Virtualbox driver does not work for several years with minikube. So this could be deprecated like we did with the hyperkit.

divysinghvi:
This PR adds a deprecation warning for the VirtualBox driver, similar to the existing hyperkit deprecation warning. The VirtualBox driver will display a clear warning message to users indicating it will be removed in a future release and suggesting alternative drivers.
Before 

## Before  

```bash
😄 minikube v1.37.0 on Darwin 15.6.1 (arm64)
✨ Using the virtualbox driver based on user configuration
👍 Starting "minikube" primary control-plane node in "minikube" cluster 
```

## After
```bash
divysinghvi@Divys-MacBook-Air minikube % ./out/minikube start --driver=virtualbox
😄  minikube v1.37.0 on Darwin 15.6.1 (arm64)
❗  The 'virtualbox' driver is deprecated and will be removed in a future release.
    You can use alternative drivers such as 'vfkit', 'qemu', or 'docker'.
```

After suggested change
```bash
❗  The 'virtualbox' driver is deprecated and will be removed in a future release.
    You can use alternative drivers: qemu2, vfkit, krunkit, parallels, docker, podman, ssh.
```

Fixes #21608

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -308,7 +308,7 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 		klog.Errorf("Error autoSetOptions : %v", err)
 	}
 
-	virtualBoxMacOS13PlusWarning(driverName)
+	virtualBoxDeprecationWarning(driverName)
 	hyperkitDeprecationWarning(driverName)
 	validateFlags(cmd, driverName)
 	validateUser(driverName)
@@ -402,16 +402,17 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 	}, nil
 }
 
-func virtualBoxMacOS13PlusWarning(driverName string) {
-	if !driver.IsVirtualBox(driverName) || !detect.MacOS13Plus() {
+// virtualBoxDeprecationWarning prints a deprecation warning for the virtualbox driver
+func virtualBoxDeprecationWarning(driverName string) {
+	if !driver.IsVirtualBox(driverName) {
 		return
 	}
-	out.WarningT(`Due to changes in macOS 13+ minikube doesn't currently support VirtualBox. You can use alternative drivers such as 'vfkit', 'qemu', or 'docker'.
+	out.WarningT(`The 'virtualbox' driver is deprecated and will be removed in a future release.
+    You can use alternative drivers such as 'vfkit', 'qemu', or 'docker'.
     https://minikube.sigs.k8s.io/docs/drivers/vfkit/
     https://minikube.sigs.k8s.io/docs/drivers/qemu/
     https://minikube.sigs.k8s.io/docs/drivers/docker/
-    For more details on the issue see: https://github.com/kubernetes/minikube/issues/15274
-`)
+	`)
 }
 
 // hyperkitDeprecationWarning prints a deprecation warning for the hyperkit driver
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -408,11 +408,7 @@ func virtualBoxDeprecationWarning(driverName string) {
 		return
 	}
 	out.WarningT(`The 'virtualbox' driver is deprecated and will be removed in a future release.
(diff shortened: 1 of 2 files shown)


divysinghvi:
@nirs  https://minikube.sigs.k8s.io/docs/drivers/ i was thinking since we say stuff like preferred , experimental on this docs page shouldn't we mention the deprecated one's also this could be useful for people who are new so that then focus on right drivers .

cmd/minikube/cmd/start.go
```diff
@@ -402,16 +402,17 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
...
-func virtualBoxMacOS13PlusWarning(driverName string) {
-	if !driver.IsVirtualBox(driverName) || !detect.MacOS13Plus() {
+// virtualBoxDeprecationWarning prints a deprecation warning for the virtualbox driver
+func virtualBoxDeprecationWarning(driverName string) {
+	if !driver.IsVirtualBox(driverName) {
 		return
 	}
-	out.WarningT(`Due to changes in macOS 13+ minikube doesn't currently support VirtualBox. You can use alternative drivers such as 'vfkit', 'qemu', or 'docker'.
+	out.WarningT(`The 'virtualbox' driver is deprecated and will be removed in a future release.
+    You can use alternative drivers such as 'vfkit', 'qemu', or 'docker'.
```
medyagh: we should list the drivers available Per Platform,
I believe regstiry package might have a func that gives you Available Drviers

medyagh (changes requested):

cmd/minikube/cmd/start.go
divysinghvi: I have done the change this is how it looks now 
```bash
❗  The 'virtualbox' driver is deprecated and will be removed in a future release.
    You can use alternative drivers: qemu2, vfkit, krunkit, parallels, docker, podman, ssh.
```

afbjorklund: You can go back and make a similar fix to hyperkit warning (in a new PR), to make them look more similar:
* #21024

divysinghvi: Yes i have to change the hyperkit one in a new pr i was waiting for this one to be confirmed if this looks good i can make the pr for the hyperkit one also thanks

medyagh (approved):

medyagh:
thank you @divysinghvi that looks better, yes you can create a follow up PR to make hyperkit and others do the same

I suggest the follow up Pr have a general func called

warnDriverDepricated

and that can have for loop that goes ever every Depricatred driverd and checks if that matches the selected driver.
and only differen thing there would be the message if needs to be different

divysinghvi:
yes that could be really useful in future cause since we have a lot of upcoming deprecated drivers as nirs and abjorklund were discussing like parallals etc actually i did make the hyperkit message change also but it's better to have a common function for this task will make change in that too

divysinghvi:
@medyagh why is integration test failing on this pr any clue??

cmd/minikube/cmd/start.go
```diff
@@ -408,11 +408,7 @@ func virtualBoxDeprecationWarning(driverName string) {
 		return
 	}
 	out.WarningT(`The 'virtualbox' driver is deprecated and will be removed in a future release.
-    You can use alternative drivers such as 'vfkit', 'qemu', or 'docker'.
-    https://minikube.sigs.k8s.io/docs/drivers/vfkit/
-    https://minikube.sigs.k8s.io/docs/drivers/qemu/
-    https://minikube.sigs.k8s.io/docs/drivers/docker/
-	`)
+    You can use alternative drivers: {{.drivers}}.`, out.V{"drivers": strings.Join(driver.SupportedDrivers(), ", ")})
```
nirs, line 411: Why did you remove this? please keep the links to the docs. This make it very easy to find more info by clicking on the link in the shell, and it is not required for this change.

nirs (changes requested):

PR #21635, opened 2025-09-25
vmnet: Skip validation in download-only mode

### LINKED ISSUE #21634: vmnet-helper should not be asking for password in --download-only mode
````
$ mk delete --all --purge && mk start --download-only
💀  Successfully purged minikube directory located at - [/Users/medya/.minikube]
😄  minikube v1.37.0 on Darwin 15.6.1 (arm64)
💡  Unable to run vmnet-helper without a password
    To configure vment-helper to run without a password, please check the documentation:
    https://github.com/nirs/vmnet-helper/#granting-permission-to-run-vmnet-helper
Password: 

```

vmnt-helper shoudl had not been invoked at all during --download--only

nirs:
In this mode we don't start anything so we don't need to validate vmnet-helper. This avoids unwanted interaction if the user does not have sudoers configuration.

Test with vfkit:

    % out/minikube start --driver vfkit --network vmnet-shared --download-only
    😄  minikube v1.37.0 on Darwin 15.7 (arm64)
    ✨  Using the vfkit driver based on user configuration
    ✅  Download complete!

    % out/minikube logs | grep vmnet-helper
    I0925 16:52:08.601139   78267 main.go:141] libmachine: Skipping vmnet-helper validation in download-only mode

Test with krunkit:

    % out/minikube start --driver krunkit --download-only
    😄  minikube v1.37.0 on Darwin 15.7 (arm64)
    ✨  Using the krunkit (experimental) driver based on user configuration
    ✅  Download complete!

    % out/minikube logs | grep vmnet-helper
    I0925 16:52:49.570566   78405 main.go:141] libmachine: Skipping vmnet-helper validation in download-only mode

Fixes #21634

### DIFF
--- a/pkg/drivers/common/vmnet/vmnet.go
+++ b/pkg/drivers/common/vmnet/vmnet.go
@@ -72,6 +72,14 @@ type interfaceInfo struct {
 // The returned error.Kind can be used to provide a suggestion for resolving the
 // issue.
 func ValidateHelper() error {
+	// Ideally minikube will not try to validate in download-only mode, but this
+	// is called from different places in different drivers, so the easier way
+	// to skip validation is to skip it here.
+	if viper.GetBool("download-only") {
+		log.Debug("Skipping vmnet-helper validation in download-only mode")
+		return nil
+	}
+
 	// Is it installed?
 	if _, err := os.Stat(executablePath); err != nil {
 		if errors.Is(err, os.ErrNotExist) {


nirs:
/cc @medyagh 

No need to run integration tests since this change only vmnet.Validate() used only in vfkit and krunkit.

pkg/drivers/common/vmnet/vmnet.go
```diff
@@ -72,6 +72,14 @@ type interfaceInfo struct {
 // The returned error.Kind can be used to provide a suggestion for resolving the
 // issue.
 func ValidateHelper() error {
+	// Ideally minikube will not try to validate in download-only mode, but this
+	// is called from different places in different drivers, so the easier way
+	// to skip validation is to skip it here.
+	if viper.GetBool("download-only") {
```
medyagh, line 78: I have been trying to discourge checking the viper config in the packages deep inside. the guideline is all the viper stuff needs to be in cmd, how about make an optional arg for this func that if passed Skip to skip it.

there are instances we violate that we have TODO comments to fix them (like preload) that I think is causing some issues with @divysinghvi 's PR testing

also when we rename the flag variables, in this case we have a const for "download-only" in cmd package that shoud had been there. there is another viper in this vmnet package code that should also be moved out

medyagh (changes requested):

pkg/drivers/common/vmnet/vmnet.go
medyagh: it should be in https://github.com/kubernetes/minikube/blob/71227b25cd999e4333dc9e85b986d6fb50b350cf/cmd/minikube/cmd/start_flags.go#L538

btw I also noticed func validateVfkitNetwork()   is not called for Krunkit

```func getNetwork(driverName string) string {
	n := viper.GetString(network)
	if driver.IsQEMU(driverName) {
		return validateQemuNetwork(n)
	} else if driver.IsVFKit(driverName) {
		return validateVfkitNetwork(n)
	}
	return n
}
``` 

that shoudl also be sometihng like validateVmnet-HelperNetwork
and only run on krunkit or vkfit

medyagh: one of the reaons for avoid using Viper in the non-cmd packages is testing issues, when we do parallel testing and configs overriding each other.

nirs: We already use viper for checking if we run with --interactive=false, so using download-only matches the current code.

The reason validateVfkitNetwork() is not used for krunkit is that it does not have a --network flag so there is nothing to validate. In krunkit case we validate in the driver Startus(), same we validate the kvm driver configuration.

I can improve it later to pass options to validate, since the interactive check must be done in validate.

medyagh: interactive flag is not as bad as "download-only flag" since that one is used more in the test. I think since we are already here, we should move all the viper ones here to cmd packages

PR #21642, opened 2025-09-26
Refactor Ivoking Kubeadm command to prepare for debian 12

medyagh:
- **bootstrapper: Export bsutil.binRoot**
- **bootstraper: Fix kubeadm init PATH on Debian 12**
- **bootstrapper: rename func to  KubeadmCmdWithPath**
- **move sudo before bash -c everywhere we invoke kubeadms to ensure gets the correct path**

### DIFF
--- a/pkg/minikube/bootstrapper/bsutil/binaries.go
+++ b/pkg/minikube/bootstrapper/bsutil/binaries.go
@@ -46,7 +46,7 @@ func TransferBinaries(cfg config.KubernetesConfig, c command.Runner, sm sysinit.
 	}
 	klog.Infof("Didn't find k8s binaries: %v\nInitiating transfer...", err)
 
-	dir := binRoot(cfg.KubernetesVersion)
+	dir := BinRoot(cfg.KubernetesVersion)
 	_, err = c.RunCmd(exec.Command("sudo", "mkdir", "-p", dir))
 	if err != nil {
 		return err
@@ -79,7 +79,7 @@ func TransferBinaries(cfg config.KubernetesConfig, c command.Runner, sm sysinit.
 
 // binariesExist returns true if the binaries already exist
 func binariesExist(cfg config.KubernetesConfig, c command.Runner) (bool, error) {
-	dir := binRoot(cfg.KubernetesVersion)
+	dir := BinRoot(cfg.KubernetesVersion)
 	rr, err := c.RunCmd(exec.Command("sudo", "ls", dir))
 	if err != nil {
 		return false, err
@@ -97,7 +97,7 @@ func binariesExist(cfg config.KubernetesConfig, c command.Runner) (bool, error) 
 	return true, nil
 }
 
-// binRoot returns the persistent path binaries are stored in
-func binRoot(version string) string {
+// BinRoot returns the persistent path binaries are stored in
+func BinRoot(version string) string {
 	return path.Join(vmpath.GuestPersistentDir, "binaries", version)
 }
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -228,8 +228,19 @@ func (k *Bootstrapper) init(cfg config.ClusterConfig) error {
 	ctx, cancel := context.WithTimeout(context.Background(), initTimeoutMinutes*time.Minute)
 	defer cancel()
 	kr, kw := io.Pipe()
-	c := exec.CommandContext(ctx, "/bin/bash", "-c", fmt.Sprintf("%s init --config %s %s --ignore-preflight-errors=%s",
-		bsutil.InvokeKubeadm(cfg.KubernetesConfig.KubernetesVersion), conf, extraFlags, strings.Join(ignore, ",")))
+
+	// NOTE: The command must run with the a root shell to expand PATH to the
+	// root PATH. On Debian 12 user PATH does not contain /usr/sbin which breaks
+	// kubeadm since https://github.com/kubernetes/kubernetes/pull/129450.
+	cmd := fmt.Sprintf(
+		"env PATH=\"%s:$PATH\" kubeadm init --config %s %s --ignore-preflight-errors=%s",
+		bsutil.BinRoot(cfg.KubernetesConfig.KubernetesVersion),
+		conf,
+		extraFlags,
+		strings.Join(ignore, ","),
+	)
+	c := exec.CommandContext(ctx, "sudo", "bash", "-c", cmd)
+
 	c.Stdout = kw
 	c.Stderr = kw
 	var wg sync.WaitGroup
--- a/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
+++ b/pkg/minikube/bootstrapper/kubeadm/kubeadm.go
@@ -229,17 +229,14 @@ func (k *Bootstrapper) init(cfg config.ClusterConfig) error {
 	defer cancel()
 	kr, kw := io.Pipe()
 
-	// NOTE: The command must run with the a root shell to expand PATH to the
-	// root PATH. On Debian 12 user PATH does not contain /usr/sbin which breaks
-	// kubeadm since https://github.com/kubernetes/kubernetes/pull/129450.
 	cmd := fmt.Sprintf(
-		"env PATH=\"%s:$PATH\" kubeadm init --config %s %s --ignore-preflight-errors=%s",
-		bsutil.BinRoot(cfg.KubernetesConfig.KubernetesVersion),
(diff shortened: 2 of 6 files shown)


pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -672,10 +669,11 @@ func (k *Bootstrapper) restartPrimaryControlPlane(cfg config.ClusterConfig) erro
 
 	// Run commands one at a time so that it is easier to root cause failures.
 	for _, c := range cmds {
-		if _, err := k.c.RunCmd(exec.Command("/bin/bash", "-c", c)); err != nil {
+		// c := exec.CommandContext(ctx, "sudo", "bash", "-c", cmd)
```
nirs: can be removed now

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -672,10 +669,11 @@ func (k *Bootstrapper) restartPrimaryControlPlane(cfg config.ClusterConfig) erro
...
+		if _, err := k.c.RunCmd(exec.Command("sudo", "/bin/bash", "-c", c)); err != nil {
 			klog.Errorf("%s failed - will try once more: %v", c, err)
 
-			if _, err := k.c.RunCmd(exec.Command("/bin/bash", "-c", c)); err != nil {
+			if _, err := k.c.RunCmd(exec.Command("sudo", "/bin/bash", "-c", c)); err != nil {
```
nirs, line 675: We should reuse the same command instead of rebuilding it, but it was already like this so it can be improve later.

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -784,14 +782,14 @@ func (k *Bootstrapper) JoinCluster(cc config.ClusterConfig, n config.Node, joinC
...
-	tokenCmd := exec.Command("/bin/bash", "-c", fmt.Sprintf("%s token create --print-join-command --ttl=0", bsutil.InvokeKubeadm(cc.KubernetesConfig.KubernetesVersion)))
+	tokenCmd := exec.Command("sudo", "/bin/bash", "-c", fmt.Sprintf("%s token create --print-join-command --ttl=0", bsutil.KubeadmCmdWithPath(cc.KubernetesConfig.KubernetesVersion)))
 	r, err := k.c.RunCmd(tokenCmd)
 	if err != nil {
 		return "", errors.Wrap(err, "generating join command")
 	}
 
 	joinCmd := r.Stdout.String()
-	joinCmd = strings.Replace(joinCmd, "kubeadm", bsutil.InvokeKubeadm(cc.KubernetesConfig.KubernetesVersion), 1)
+	joinCmd = strings.Replace(joinCmd, "kubeadm", bsutil.KubeadmCmdWithPath(cc.KubernetesConfig.KubernetesVersion), 1)
```
nirs, line 791: This returns joinCmd, but the caller need to run it with `sudo bash -c`

medyagh: the caller is  785, no ?
already converted
```	tokenCmd := exec.Command("sudo", "/bin/bash", "-c", fmt.Sprintf("%
```

nirs: The caller is this - missing sudo.

```
	if _, err := k.c.RunCmd(exec.Command("/bin/bash", "-c", joinCmd)); err != nil {
		return errors.Wrapf(err, "kubeadm join")
	}
```

medyagh: thanks ! good catch

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
```diff
@@ -768,11 +768,11 @@ func (k *Bootstrapper) JoinCluster(cc config.ClusterConfig, n config.Node, joinC
 			" --apiserver-bind-port=" + strconv.Itoa(n.Port)
 	}
 
-	if _, err := k.c.RunCmd(exec.Command("/bin/bash", "-c", joinCmd)); err != nil {
+	if _, err := k.c.RunCmd(exec.Command("sudo", "/bin/bash", "-c", joinCmd)); err != nil {
 		return errors.Wrapf(err, "kubeadm join")
 	}
 
-	if _, err := k.c.RunCmd(exec.Command("/bin/bash", "-c", "sudo systemctl daemon-reload && sudo systemctl enable kubelet && sudo systemctl start kubelet")); err != nil {
+	if _, err := k.c.RunCmd(exec.Command("sudo", "/bin/bash", "-c", "systemctl daemon-reload && systemctl enable kubelet && systemctl start kubelet")); err != nil {
```
nirs, line 774: This change is not needed since we don't use `env PATH=` in the command.

pkg/minikube/bootstrapper/certs.go
```diff
@@ -418,8 +419,7 @@ func renewExpiredKubeadmCerts(cmd command.Runner, cc config.ClusterConfig) error
 		return nil
 	}
 	out.WarningT("kubeadm certificates have expired. Generating new ones...")
-	kubeadmPath := path.Join(vmpath.GuestPersistentDir, "binaries", cc.KubernetesConfig.KubernetesVersion)
-	bashCmd := fmt.Sprintf("sudo env PATH=\"%s:$PATH\" kubeadm certs renew all --config %s", kubeadmPath, constants.KubeadmYamlPath)
+	bashCmd := fmt.Sprintf("%s certs renew all --config %s", bsutil.KubeadmCmdWithPath(cc.KubernetesConfig.KubernetesVersion), constants.KubeadmYamlPath)
 	if _, err := cmd.RunCmd(exec.Command("/bin/bash", "-c", bashCmd)); err != nil {
```
nirs: Missing sudo

pkg/minikube/bootstrapper/bsutil/binaries.go
```diff
@@ -101,3 +101,21 @@ func binariesExist(cfg config.KubernetesConfig, c command.Runner) (bool, error)
...
+// copyBinary copies a locally cached binary to the guest VM
+func copyBinary(cr command.Runner, src, dest string) error {
+	f, err := assets.NewFileAsset(src, path.Dir(dest), path.Base(dest), "0755")
+	if err != nil {
+		return errors.Wrap(err, "new file asset")
+	}
+	defer func() {
+		if err := f.Close(); err != nil {
+			klog.Warningf("error closing the file %s: %v", f.GetSourcePath(), err)
+		}
```
nirs, line 114: The asset is opened for reading, and close error is not very interesting. But the issue was not introduced by your change.

nirs:
@medyagh can you squash the fixup commits so we can have clean git history?

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
medyagh: doenst hurt to be in same format for all, does it ? I prefer they all look same

nirs: Good point, but we should avoid running command with bash when bash is not needed. Using the a shell is a security risk and it should be used only when we must use the shell. In this case we run:

    bash -c "systemctl daemon-reload && systemctl enable kubelet && systemctl start kubelet"

But it is better to run 2 separate commands without a shell:

     systemctl daemon-reload
     systemctl enable kubelet --now

`systemctl daemon-reload` is needed only if we added or modified the unit file. If we did not, it should not be called, so this should be really:

    systemctl enable kubelet --now

This is in not the scope of this change. The changes in scope to not add unneeded change in this command.

pkg/minikube/bootstrapper/certs.go
medyagh: good catch ! thannk you solved @nirs

pkg/minikube/bootstrapper/kubeadm/kubeadm.go
medyagh: I think there some minikube things that Might change kubelet (maybe --registry-mirror)
one reason for combining the commands is our CommandRunners
we have many command runners like ExecRunner, SSHrunner, KicRunner (docker exec -it ) and some of them are Slow (like kicrunner) so we try to make one session to make it faster.
the slowness comes from initiating a session

nirs: Running a script to speed up make sense, but I think in this case it is not needed. `systemctl daemon-reload` should be called by the command adding or modifying the unit file - not here. And enabling and starting the service is one command (enable --now). But this is not in the scope of this change.

PR #21643, opened 2025-09-26
Switch ubuntu to debian 

medyagh:
- **bootstrapper: Export bsutil.binRoot**
- **bootstraper: Fix kubeadm init PATH on Debian 12**
- **bootstrapper: rename func to  KubeadmCmdWithPath**
- **move sudo before bash -c everywhere we invoke kubeadms to ensure gets the correct path**
- **make binRoot private again**
- **change ubuntu name to debian in code base**
- **update automation update tool for kicbase**
- **update comment**
- **capitalize AS in Docker instructions**
- **update comment**
- **enable RSA1 support in sshd config in debian dockerfile**
- **config Docker user for Passwordless login and add to corect groups**
- **install latst version of crio available in  debian repos**
- **update docker installation for debian based on docker docs**
- **add crictl installation seprately since it is no longer shipped with crio**
- **remove old workarrounds**


nirs:
This must be merged together with https://github.com/kubernetes/minikube/pull/21409 since the preloads created with containerd 1.7 may not be correct for containerd 2.

medyagh:
> This must be merged together with #21409 since the preloads created with containerd 1.7 may not be correct for containerd 2.

this doesnt touch iso, Containerd2 work can be done After this PR

nirs:
> > This must be merged together with #21409 since the preloads created with containerd 1.7 may not be correct for containerd 2.
> 
> this doesnt touch iso, Containerd2 work can be done After this PR

When we update the base image we get a new version of containerd, so we ned to update also nerdctl. And since we share the preloads, we may need the same version of containerd in kicbase and iso. If this change does not bring contianerd v2 we are fine.

medyagh:
Docker Linux Passes
Docker Linux Containerd Failures are Docker hub Rate limits it passed in [previous commit](https://storage.googleapis.com/minikube-builds/logs/21643/41724/Docker_Linux_containerd.html)

Docker Linux Crio is meant to be fixed by community or droped from minikube

PR #21625, opened 2025-09-23
kvm: Convert to internal driver

nirs:
Remove the docker-machine-driver-kvm2 wrapper and use the kvm driver as internal driver. This fixes all the build and test issues related to building and testing the docker-machine-driver-kvm2, and simplify the kvm driver runtime flow.

To avoid dependency on libvirt shared library on Linux, we build now with libvirt_dlopen tag. With this we don't need libvirt devel libraies during build, and in runtime we will fail if libvirt shared library is not installed.

Building minikube for arm64 with libvirt does not work yet. This build was already skipped for docker-machine-driver-kvm2 external driver. Since now the kvm driver is internal, the kvm driver was modified to build only for linux/amd64.

## Example run

```
$ make
go build  -tags "libvirt_dlopen" -ldflags="-X k8s.io/minikube/pkg/version.version=v1.37.0 -X k8s.io/minikube/pkg/version.isoVersion=v1.37.0 -X k8s.io/minikube/pkg/version.gitCommitID="b25c816e2321bfd20593b48d1f9359079818c20b" -X k8s.io/minikube/pkg/version.storageProvisionerVersion=v5" -o out/minikube k8s.io/minikube/cmd/minikube

$ ldd out//minikube 
	linux-vdso.so.1 (0x00007f2808c43000)
	libresolv.so.2 => /lib64/libresolv.so.2 (0x00007f2808c1a000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f2808a28000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f2808c45000)

$ rm -f ~/.minikube/bin/docker*

$ out/minikube start -d kvm
😄  minikube v1.37.0 on Fedora 42 (kvm/amd64)
✨  Using the kvm2 driver based on user configuration
👍  Starting "minikube" primary control-plane node in "minikube" cluster
🔥  Creating kvm2 VM (CPUs=2, Memory=6144MB, Disk=20000MB) ...
🐳  Preparing Kubernetes v1.34.0 on Docker 28.4.0 ...
🔗  Configuring bridge CNI (Container Networking Interface) ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: default-storageclass, storage-provisioner
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default

$ ls ~/.minikube/bin/
```

## Testing
[...]

PR #21644, opened 2025-09-27
test: fix test helper for AUX and fix SELinux build

nirs:
Fix issues found when working #21625. Separated to another PR since these are trivial changes and simple fixes that we can merge quickly, and to verify the changes without the big kvm driver changes.

- build: Fix docker command when selinux is enforced
- Remove trailing whitespace
- test: Fix NeedsAuxDriver() to check for hyperkit
- test: De-duplicate driver detection

### DIFF
--- a/test/integration/main_test.go
+++ b/test/integration/main_test.go
@@ -172,9 +172,13 @@ func KicDriver() bool {
 	return DockerDriver() || PodmanDriver()
 }
 
+func HyperkitDriver() bool {
+	return strings.Contains(*startArgs, "--driver=hyperkit") || strings.Contains(*startArgs, "--vm-driver=hyperkit")
+}
+
 // NeedsAuxDriver Returns true if the driver needs an auxiliary driver (kvm, hyperkit,..)
 func NeedsAuxDriver() bool {
-	return HyperVDriver() || KVMDriver()
+	return HyperkitDriver() || KVMDriver()
 }
 
 // VMDriver checks if the driver is a VM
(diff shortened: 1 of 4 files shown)


test/integration/main_test.go
```diff
@@ -228,3 +232,15 @@ func Seconds(n int) time.Duration {
 func TestingKicBaseImage() bool {
 	return strings.Contains(*startArgs, "base-image")
 }
+
+func matchDriverFlag(names ...string) bool {
+	args := *startArgs
+	for _, name := range names {
+		for _, prefix := range []string{"--driver=", "--vm-driver="} {
```
medyagh, line 239: since you are adding this general helper, lets add --driver without =
--driver kvm2 should be correct too

and also why not add "-d" also the shorthand for the --driver, so it will be future proofed.

nirs: Sur we can handle now all cases, but I'm not sure we support `--driver name` in the tests, since we accept the arguments as string, and we split the string by space. So the helper can detect:

- `--driver=foo`
- `--driver foo`
- `--vm-driver=foo`
- `--vm-driver foo`
- `-dfoo`
- `-d foo`

But the rest of the code may not work correctly with the `--driver foo` form. If we support all forms in the helper, we tell developers that they can use these forms, and it may not work. I think we should keep this simple and do not claim that the other forms work.

I think it is better open an issue for supporting the other forms, or replacing this mechanism with a configuration file. This is a minimal change to avoid the ugly duplication.

medyagh: I cant imagine why the test wouldn't work if the test was created with -d docker instaed of --driver docker (this args are usually not used in the test itself but rather jsut to start the cluster with that drveri for test env)

this is good enough, thats just a nit ! we could merge this

medyagh (approved):

nirs:
@medyagh tests finished, can you merge? this is required for #21625

PR #21645, opened 2025-09-27
docs: Improve testing docs

nirs:
- Simplify the example commands to pass TEST_ARGS as a command line variable. Variables are pass to the target shell as environment variable.
- Show an example with 2 minikube start args with correct quoting.
- Add a note about quoting multiple minikube start arguments
- Add a note about values with spaces
- Fix headings levels for conformance tests

### DIFF
--- a/site/content/en/docs/contrib/testing.en.md
+++ b/site/content/en/docs/contrib/testing.en.md
@@ -45,15 +45,19 @@ make integration
 You may find it useful to set various options to test only a particular test against a non-default driver. For instance:
 
 ```shell
- env TEST_ARGS="-minikube-start-args=--driver=hyperkit -test.run TestStartStop" make integration
+make integration TEST_ARGS="-minikube-start-args='--driver=vfkit --network=vmnet-shared' -test.run TestStartStop"
 ```
+
+IMPORTANT:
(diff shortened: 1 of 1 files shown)


site/content/en/docs/contrib/testing.en.md
```diff
@@ -45,15 +45,19 @@ make integration
...
+- To pass multiple flags to `-minikube-start-args` you must quote the value
+- flag values cannot contain space since the string is split by space
+
 ### Quickly iterating on a single test
 
 Run a single test on an active cluster:
 
 ```shell
-make integration -e TEST_ARGS="-test.run TestFunctional/parallel/MountCmd --profile=minikube --cleanup=false"
+make integration TEST_ARGS="-test.run TestFunctional/parallel/MountCmd --profile=minikube --cleanup=false"
```
medyagh, line 60: do you wanna choose a more meaninful sub test ? I think once you said MountCmd is not testable by itself... ?

nirs: I kept the existing example - only changing the way we define the environment varaible. I don't know it it works. I had trouble with the image commands tests, depending on state created by the parent and/or previous tests.

PR #21664, opened 2025-09-29
Move linux/darwin binaries into cache/binaries folder (#21179)

### LINKED ISSUE #21179: move binaries in cache folder under folder called "binaries"
right now it looks like this

```
$ ls ~/.minikube/cache/
darwin            images            iso               kic               linux             preloaded-tarball

```

here the darwin and linux (and if windows) 
it should be moved a folder

henry3260:
This PR addresses issue #21179 by reorganizing the cache folder structure for binaries.

### Before
- Binaries for Linux and Darwin were placed directly under:
  - `~/.minikube/cache/linux`
  - `~/.minikube/cache/darwin`

### After
- They are now moved into a dedicated `binaries` subfolder:
  - `~/.minikube/cache/bin/linux`
  - `~/.minikube/cache/bin/darwin`

### Why
- This improves organization of the cache directory.
- Aligns binaries with how other cached items (e.g., images, ISO files) are stored.
- No functional behavior is changed, only the storage path for binaries.

Fixes #21179

### Screenshot (Before)
image

### Screenshot (After)
image

### DIFF
--- a/pkg/minikube/download/binary.go
+++ b/pkg/minikube/download/binary.go
@@ -19,7 +19,7 @@ package download
 import (
 	"fmt"
 	"os"
-	"path"
+	"path/filepath"
 	"runtime"
 
 	"github.com/blang/semver/v4"
@@ -53,8 +53,12 @@ func binaryWithChecksumURL(binaryName, version, osName, archName, binaryURL stri
 
 // Binary will download a binary onto the host
 func Binary(binary, version, osName, archName, binaryURL string) (string, error) {
-	targetDir := localpath.MakeMiniPath("cache", osName, archName, version)
-	targetFilepath := path.Join(targetDir, binary)
+	targetFilepath := localpath.CachedBinaryPath(binary, version, osName, archName)
+	targetDir := filepath.Dir(targetFilepath)
+
+	if err := os.MkdirAll(targetDir, 0755); err != nil {
+		return "", errors.Wrapf(err, "failed to create target dir %s", targetDir)
+	}
 	targetLock := targetFilepath + ".lock"
 
 	url, err := binaryWithChecksumURL(binary, version, osName, archName, binaryURL)
--- a/pkg/minikube/download/download_test.go
+++ b/pkg/minikube/download/download_test.go
@@ -20,11 +20,13 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
+	"path/filepath"
 	"sync"
 	"testing"
 	"time"
 
 	"k8s.io/minikube/pkg/minikube/constants"
+	"k8s.io/minikube/pkg/minikube/localpath"
 )
 
 // Force download tests to run in serial.
@@ -240,5 +242,26 @@ func testPreloadWithCachedSizeZero(t *testing.T) {
 
 	if downloadNum != 1 {
 		t.Errorf("Expected only 1 download attempt but got %v!", downloadNum)
+	}
+}
+
+func TestBinaryCreatesBinariesDir(t *testing.T) {
+	// mock download function that creates a fake binary file
+	DownloadMock = func(src, dst string) error {
+		// create a fake binary file at dst
+		return os.WriteFile(dst, []byte("fake-binary"), 0755)
+	}
+
+	_, err := Binary("kubectl", "v1.31.0", "linux", "amd64", "")
+	if err != nil {
+		t.Fatalf("Binary() failed: %v", err)
+	}
+
+	// check that the bin directory was created
+	targetDir := filepath.Join(
+		localpath.MiniPath(), "cache", "bin", "linux", "amd64", "v1.31.0",
+	)
+	if _, err := os.Stat(targetDir); os.IsNotExist(err) {
+		t.Errorf("Expected bin dir %s to exist, but it does not", targetDir)
 	}
 }
--- a/pkg/minikube/localpath/localpath.go
+++ b/pkg/minikube/localpath/localpath.go
@@ -247,3 +247,7 @@ func getWindowsVolumeNameCmd(d string) (string, error) {
 }
 
 var getWindowsVolumeName = getWindowsVolumeNameCmd
+
+func CachedBinaryPath(binary, version, goos, goarch string) string {
+	return filepath.Join(MiniPath(), "cache", "bin", goos, goarch, version, binary)
+}
--- a/test/integration/aaa_download_only_test.go
+++ b/test/integration/aaa_download_only_test.go
@@ -152,7 +152,7 @@ func TestDownloadOnly(t *testing.T) { // nolint:gocyclo
 				}
 				// checking binaries downloaded (kubelet,kubeadm)
 				for _, bin := range constants.KubernetesReleaseBinaries {
-					fp := filepath.Join(localpath.MiniPath(), "cache", "linux", runtime.GOARCH, v, bin)
+					fp := localpath.CachedBinaryPath(bin, v, "linux", runtime.GOARCH)
 					_, err := os.Stat(fp)
 					if err != nil {
 						t.Errorf("expected the file for binary exist at %q but got error %v", fp, err)


pkg/minikube/download/binary.go
```diff
@@ -53,7 +53,11 @@ func binaryWithChecksumURL(binaryName, version, osName, archName, binaryURL stri
 
 // Binary will download a binary onto the host
 func Binary(binary, version, osName, archName, binaryURL string) (string, error) {
-	targetDir := localpath.MakeMiniPath("cache", osName, archName, version)
+	// targetDir := localpath.MakeMiniPath("cache", osName, archName, version)
+	targetDir := localpath.MakeMiniPath("cache", "binaries", osName, archName, version)
```
nirs: I think we should use "bin", it is more consistent with ~/.minikube/bin and the rest of the world.

We are using "binaries" internally which would nice to change also to bin but it may be harder since we may code using it. Since this cache directory is new we don't have backward compatibility concerns.

henry3260: > I think we should use "bin", it is more consistent with ~/.minikube/bin and the rest of the world.
> 
> We are using "binaries" internally which would nice to change also to bin but it may be harder since we may code using it. Since this cache directory is new we don't have backward compatibility concerns.

Thanks for the suggestion! That makes sense, I’ll update it to use bin for consistency.

nirs (approved):

pkg/minikube/download/download_test.go
```diff
@@ -242,3 +243,24 @@ func testPreloadWithCachedSizeZero(t *testing.T) {
...
+	}
+
+	_, err := Binary("kubectl", "v1.31.0", "linux", "amd64", "")
+	if err != nil {
+		t.Fatalf("Binary() failed: %v", err)
+	}
+
+	// check that the bin directory was created
+	targetDir := filepath.Join(
+		os.Getenv("HOME"), ".minikube", "cache", "bin", "linux", "amd64", "v1.31.0",
```
nirs: ~/.minikube is the default, but if MINIKUBE_HOME is defined it will be used. There is a helper function to return the right path (but I don't remember the function name). Check what other tests are doing.

pkg/minikube/download/download_test.go
```diff
@@ -258,7 +259,7 @@ func TestBinaryCreatesBinariesDir(t *testing.T) {
 
 	// check that the bin directory was created
 	targetDir := filepath.Join(
-		os.Getenv("HOME"), ".minikube", "cache", "bin", "linux", "amd64", "v1.31.0",
+		localpath.MiniPath(), "cache", "bin", "linux", "amd64", "v1.31.0",
```
nirs, line 262: Looks good, to make it even better squash this commit into the first one. Having this in git history is not helpful.

nirs (approved):

medyagh (changes requested):
thanks for the PR, please update the before/after (it seems to be outdated)

henry3260:
> thanks for the PR, please update the before/after (it seems to be outdated)

updated!

nirs:
> > thanks for the PR, please update the before/after (it seems to be outdated)
> 
> updated!

It is still outdated:

> They are now moved into a dedicated binaries subfolder:
> ~/.minikube/cache/binaries/linux
> ~/.minikube/cache/binaries/darwin

henry3260:
> > > thanks for the PR, please update the before/after (it seems to be outdated)
> > 
> > 
> > updated!
> 
> It is still outdated:
> 
> > They are now moved into a dedicated binaries subfolder:
> > ~/.minikube/cache/binaries/linux
> > ~/.minikube/cache/binaries/darwin

Thanks for pointing that out! I've corrected it now.

nirs:
@henry3260 this change broke few test:

```
=== RUN   TestDownloadOnly/v1.28.0/binaries
aaa_download_only_test.go:157: expected the file for binary exist at "/home/jenkins/minikube-integration/21664-14293/.minikube/cache/linux/amd64/v1.28.0/kubelet" but got error stat /home/jenkins/minikube-integration/21664-14293/.minikube/cache/linux/amd64/v1.28.0/kubelet: no such file or directory
aaa_download_only_test.go:157: expected the file for binary exist at "/home/jenkins/minikube-integration/21664-14293/.minikube/cache/linux/amd64/v1.28.0/kubeadm" but got error stat /home/jenkins/minikube-integration/21664-14293/.minikube/cache/linux/amd64/v1.28.0/kubeadm: no such file or directory
aaa_download_only_test.go:157: expected the file for binary exist at "/home/jenkins/minikube-integration/21664-14293/.minikube/cache/linux/amd64/v1.28.0/kubectl" but got error stat /home/jenkins/minikube-integration/21664-14293/.minikube/cache/linux/amd64/v1.28.0/kubectl: no such file or directory
```

Please run integration tests locally and fix the failing tests.
See https://minikube.sigs.k8s.io/docs/contrib/testing/

nirs (changes requested):
Need to fix the tests assuming the old location.

The issue is we compute the path manually in the test and in the code. We need a helper to locate the binaries that can be used by the code and the tests.

henry3260:
> @henry3260 this change broke few test:
> 
> ```
> === RUN   TestDownloadOnly/v1.28.0/binaries
> aaa_download_only_test.go:157: expected the file for binary exist at "/home/jenkins/minikube-integration/21664-14293/.minikube/cache/linux/amd64/v1.28.0/kubelet" but got error stat /home/jenkins/minikube-integration/21664-14293/.minikube/cache/linux/amd64/v1.28.0/kubelet: no such file or directory
> aaa_download_only_test.go:157: expected the file for binary exist at "/home/jenkins/minikube-integration/21664-14293/.minikube/cache/linux/amd64/v1.28.0/kubeadm" but got error stat /home/jenkins/minikube-integration/21664-14293/.minikube/cache/linux/amd64/v1.28.0/kubeadm: no such file or directory
> aaa_download_only_test.go:157: expected the file for binary exist at "/home/jenkins/minikube-integration/21664-14293/.minikube/cache/linux/amd64/v1.28.0/kubectl" but got error stat /home/jenkins/minikube-integration/21664-14293/.minikube/cache/linux/amd64/v1.28.0/kubectl: no such file or directory
> ```
> 
> Please run integration tests locally and fix the failing tests. See https://minikube.sigs.k8s.io/docs/contrib/testing/

It looks like the current logic no longer downloads all KubernetesReleaseBinaries locally. Before I made any changes, I ran the integration tests on main and saw similar errors:

```
aaa_download_only_test.go:157: expected the file for binary exist at "/home/jenkins/minikube-integration/.../kubeadm" but got error stat: no such file or directory
aaa_download_only_test.go:157: expected the file for binary exist at "/home/jenkins/minikube-integration/.../kubectl" but got error stat: no such file or directory

```

So the previous behavior already failed to download kubeadm and kubelet locally. Only kubectl seems to be downloaded on the host.

This means the test assumption that all binaries are always cached locally is outdated. I think we need to update the tests to only check for binaries that Minikube actually downloads on the host or to conditionally check depending on the runtime driver.

henry3260:
Do you think we should introduce a helper function like `CachedBinaryPath` inside `t.Run("cached-images", ...)` to keep things consistent?
If so, should we open a separate issue for that improvement, or would you prefer that we include it in this PR?

nirs:
It is possible that the test was broken before this change. To be sure please share complete test command and output with master showing the failures.

We meed to fix the test before we change the code. Otherwise we may break things.

I think kubeadm is cached only when disabling preloads. By default preloads are enabled and kubeadm is onstalled from the preloaded tarball. Maybe the test does not use the right flags.

Let’s open another issue for this test and work on fixing it in another PR. This PR can wait until we fix the test.

nirs:
@henry3260 looking in the test is does not disable preloads, and it checks if reload exists after starting minikube. If the preload tarball exists it does check for binaries. If preload tarball does not exit, it fallback to checking the binaries.

This test has few problems:
- it is too complicated
- it does not test the same case in all runs, depending on external state. If the preload tarball exist in the external system it test one case, and if not another case.

```go
			preloadExists := false
			t.Run("preload-exists", func(t *testing.T) {
				// skip for none, as none driver does not have preload feature.
				if NoneDriver() {
					t.Skip("None driver does not have preload")
				}
				// Driver does not matter here, since the only exception is none driver,
				// which cannot occur here.
				if !download.PreloadExists(v, containerRuntime, "docker", true) {
					t.Skip("No preload image")
				}
				// Just make sure the tarball path exists
				if _, err := os.Stat(download.TarballPath(v, containerRuntime)); err != nil {
					t.Errorf("failed to verify preloaded tarball file exists: %v", err)
				}
				preloadExists = true
			})

			t.Run("cached-images", func(t *testing.T) {
				// skip verify for cache images if --driver=none
				if NoneDriver() {
					t.Skip("None driver has no cache")
				}
				if preloadExists {
					t.Skip("Preload exists, images won't be cached")
				}
				imgs, err := images.Kubeadm("", v)
				if err != nil {
					t.Errorf("failed to get kubeadm images for %v: %+v", v, err)
				}

				for _, img := range imgs {
					pathToImage := []string{localpath.MiniPath(), "cache", "images", runtime.GOARCH}
					img = strings.Replace(img, ":", "_", 1) // for example kube-scheduler:v1.15.2 --> kube-scheduler_v1.15.2
					imagePath := strings.Split(img, "/")    // changes "gcr.io/k8s-minikube/storage-provisioner_v5" into ["gcr.io", "k8s-minikube", "storage-provisioner_v5"] to match cache folder structure
					pathToImage = append(pathToImage, imagePath...)
					fp := filepath.Join(pathToImage...)
					_, err := os.Stat(fp)
					if err != nil {
						t.Errorf("expected image file exist at %q but got error: %v", fp, err)
					}
				}
			})
```

We need is to test all cases in all runs:
- Preload does not exist, falling back to downloaded images - change preload url/name so it cannot be downloaded from the system publishing the preloads. The best way to test it is to add a --preload-url flag. We can make this flag hidden if we don't think it is useful or users and we need it only for testing.
- Preload exists, no fallback - must use a preload that always exists. Using latest release is a good bet. Since preloads are important minikube feature, we can assume that we always have a preload for the latest release. If it does not work it is likely to be fixed.
- Preload disabled (--preload=false), fallback to downloading the binaries.

Note that we also have preload_test.go, which tests --preload=false. But it does not test downloaded binaries. It seem to be focused on something else, not sure how it is related to preloads.

henry3260:
> @henry3260 looking in the test is does not disable preloads, and it checks if reload exists after starting minikube. If the preload tarball exists it does check for binaries. If preload tarball does not exit, it fallback to checking the binaries.
> 
> This test has few problems:
> 
> * it is too complicated
> * it does not test the same case in all runs, depending on external state. If the preload tarball exist in the external system it test one case, and if not another case.
> 
> ```go
> 			preloadExists := false
> 			t.Run("preload-exists", func(t *testing.T) {
> 				// skip for none, as none driver does not have preload feature.
> 				if NoneDriver() {
> 					t.Skip("None driver does not have preload")
> 				}
> 				// Driver does not matter here, since the only exception is none driver,
> 				// which cannot occur here.
> 				if !download.PreloadExists(v, containerRuntime, "docker", true) {
> 					t.Skip("No preload image")
> 				}
> 				// Just make sure the tarball path exists
> 				if _, err := os.Stat(download.TarballPath(v, containerRuntime)); err != nil {
> 					t.Errorf("failed to verify preloaded tarball file exists: %v", err)
> 				}
> 				preloadExists = true
> 			})
> 
> 			t.Run("cached-images", func(t *testing.T) {
> 				// skip verify for cache images if --driver=none
> 				if NoneDriver() {
> 					t.Skip("None driver has no cache")
> 				}
> 				if preloadExists {
> 					t.Skip("Preload exists, images won't be cached")
> 				}
> 				imgs, err := images.Kubeadm("", v)
> 				if err != nil {
> 					t.Errorf("failed to get kubeadm images for %v: %+v", v, err)
> 				}
> 
> 				for _, img := range imgs {
> 					pathToImage := []string{localpath.MiniPath(), "cache", "images", runtime.GOARCH}
> 					img = strings.Replace(img, ":", "_", 1) // for example kube-scheduler:v1.15.2 --> kube-scheduler_v1.15.2
> 					imagePath := strings.Split(img, "/")    // changes "gcr.io/k8s-minikube/storage-provisioner_v5" into ["gcr.io", "k8s-minikube", "storage-provisioner_v5"] to match cache folder structure
> 					pathToImage = append(pathToImage, imagePath...)
> 					fp := filepath.Join(pathToImage...)
> 					_, err := os.Stat(fp)
> 					if err != nil {
> 						t.Errorf("expected image file exist at %q but got error: %v", fp, err)
> 					}
> 				}
> 			})
> ```
> 
> We need is to test all cases in all runs:
> 
> * Preload does not exist, falling back to downloaded images - change preload url/name so it cannot be downloaded from the system publishing the preloads. The best way to test it is to add a --preload-url flag. We can make this flag hidden if we don't think it is useful or users and we need it only for testing.
> * Preload exists, no fallback - must use a preload that always exists. Using latest release is a good bet. Since preloads are important minikube feature, we can assume that we always have a preload for the latest release. If it does not work it is likely to be fixed.
> * Preload disabled (--preload=false), fallback to downloading the binaries.
> 
> Note that we also have preload_test.go, which tests --preload=false. But it does not test downloaded binaries. It seem to be focused on something else, not sure how it is related to preloads.

I opened the issue for this #21805

PR #21671, opened 2025-09-30
hack: Fix review comment for failed iso build

nirs:
When we fixed the way we detect failure #21544, we broke the heredoc. The script fails now with[1]:

    ./hack/jenkins/build_iso.sh: line 74: unexpected EOF while looking for matching `)'
    Build step 'Execute shell' marked build as failure

Fixes:
- Use unquoted delimiter (EOF, not 'EOF') to allow variable expansion in the content.
- Fix content indentation to avoid unwanted leading whitespace
- Fix location of the EOF - it must be start of the last line
- Add the missing closing ")"
- Remove the unneeded double quote
- Unify indentation to 4 spaces for consistency with the rest of the file.

[1] https://storage.googleapis.com/minikube-builds/logs/21409/e4af2e4/iso_build.txt

### DIFF
--- a/hack/jenkins/build_iso.sh
+++ b/hack/jenkins/build_iso.sh
@@ -66,22 +66,20 @@ else
 fi
 
 if ! make release-iso 2>&1 | tee iso-logs.txt; then
-  # Exit of `make` (PIPESTATUS[0]); fallback to 1 if unavailable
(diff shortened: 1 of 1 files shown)


medyagh (approved):
thank you

Issue #21670, opened 2025-09-30
interactive flag is not as bad as "download-only flag" since that one is used more in the test. I think since we are already here, we should move all the viper ones here to cmd packages


medyagh:
interactive flag is not as bad as "download-only flag" since that one is used more in the test. I think since we are already here, we should move all the viper ones here to cmd packages

_Originally posted by @medyagh in https://github.com/kubernetes/minikube/pull/21635#discussion_r2392461773_

nirs:
The viper calls have another big issue - they use literal strings with the flag names (e.g. "download-only"). We cannot use the actual names since the constant are not exported.

nirs:
#21682 remove the download-only check form the vmnet package. This restore the state we had before fixing #21634.

Moving all viper calls from minikube is very hard since we need to get the flags through multiple packages, down to the Driver.SetupIP(), calling firewall.Unlock() which needs the --interactive flag.

- cmd/minikube/cmd
  - minikube/node
    - minikube/machine
      - minikube/registry/drvs
        - drivers/*

nirs:
#21683 is an attempt to remove viper calls for --interactive, --download-only, and --delete-on-failure.

PR #21682, opened 2025-10-02
vmnet: Move download-only check to start command

nirs:
Skipping validation in vment.ValidateHelper() was the simplest way to avoid validation when starting with --download-only flag, but it requires accessing viper outside of the commands using strings. This make the code fragile and makes testing harder since we depend on global state.

Fixed by moving vment-helper validation to validateVmnetNetwork() helper in the minikube/cmd package. We skip the call to vment.ValidateHelper() in download-only mode. The helper is called for vfkit and krunkit when validating the --network flag.

For krunkit the validation was part of Driver.Status check, which is good place to validate, but it does not have access to start command flags. Validating in the start command is little bit ugly since krunkit does not use the --network option, and it moves driver logic to the start command, but at least it is consistent with vfkit and qemu, and it avoids the trouble of passing start flags deep inside minikube.

Part-of #21670

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -490,6 +490,8 @@ func getNetwork(driverName string) string {
 		return validateQemuNetwork(n)
 	} else if driver.IsVFKit(driverName) {
 		return validateVfkitNetwork(n)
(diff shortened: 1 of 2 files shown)


nirs:
Replaced by #21683

PR #21687, opened 2025-10-03
Lower priority for NeedsSudo drivers when no sudo available (#21633)

### LINKED ISSUE #21633: Check if we have Sudo Permission, if not Lower The priority of the drivers that Need Sudo
That also means we need to add a field to the drivers in registry package if they need sudo.

henry3260:
Fixes #21633

### What this PR does
When sudo is not available on the host, drivers that require sudo (`NeedsSudo = true`) will have their priority lowered.  
Drivers that do not require sudo remain unaffected. This ensures the `Available()` function returns a properly prioritized driver list depending on the environment.

### How I tested
- Manually verified that a "NeedsSudo" driver gets a lower priority when sudo is unavailable.
- Existing unit tests for driver registration and availability remain unaffected.

### DIFF
--- a/pkg/minikube/driver/driver.go
+++ b/pkg/minikube/driver/driver.go
@@ -19,6 +19,7 @@ package driver
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"runtime"
 	"sort"
 	"strconv"
@@ -79,6 +80,7 @@ const (
 var (
 	// systemdResolvConf is path to systemd's DNS configuration. https://github.com/kubernetes/minikube/issues/3511
 	systemdResolvConf = "/run/systemd/resolve/resolv.conf"
+	HasSudo           = realHasSudo
 )
 
 // SupportedDrivers returns a list of supported drivers
@@ -319,6 +321,14 @@ func FlagDefaults(name string) FlagHints {
 func Choices(vm bool) []registry.DriverState {
 	options := registry.Available(vm)
 
+	if !HasSudo() {
+		for i := range options {
+			if options[i].NeedsSudo {
+				options[i].Priority = registry.Discouraged
+			}
+		}
+	}
+
 	// Descending priority for predictability and appearance
 	sort.Slice(options, func(i, j int) bool {
 		return options[i].Priority > options[j].Priority
@@ -392,10 +402,11 @@ func Status(name string) registry.DriverState {
 	select {
 	case s := <-stateChannel:
 		return registry.DriverState{
-			Name:     d.Name,
-			Default:  d.Default,
-			Priority: d.Priority,
-			State:    s,
+			Name:      d.Name,
+			Default:   d.Default,
+			Priority:  d.Priority,
+			State:     s,
+			NeedsSudo: d.NeedsSudo,
 		}
 	case <-timeoutChannel:
 		klog.Infof("time out when checking for status of %s driver", name)
@@ -437,4 +448,9 @@ func IndexFromMachineName(machineName string) int {
 		return i // minion node
 	}
 	return 1 // master node
+}
+
+func realHasSudo() bool {
+	err := exec.Command("sudo", "-n", "true").Run()
+	return err == nil
 }
--- a/pkg/minikube/registry/drvs/docker/docker.go
+++ b/pkg/minikube/registry/drvs/docker/docker.go
@@ -53,6 +53,8 @@ func init() {
 		Status:   status,
 		Default:  true,
 		Priority: registry.HighlyPreferred,
+		// Docker driver does not require sudo
+		NeedsSudo: false,
 	}); err != nil {
 		panic(fmt.Sprintf("register failed: %v", err))
 	}
--- a/pkg/minikube/registry/drvs/hyperkit/hyperkit.go
+++ b/pkg/minikube/registry/drvs/hyperkit/hyperkit.go
@@ -54,6 +54,8 @@ func init() {
 		Status:   status,
 		Default:  true,
 		Priority: registry.Deprecated,
+		// Hyperkit driver requires sudo
+		NeedsSudo: true,
 	}); err != nil {
 		panic(fmt.Sprintf("register: %v", err))
 	}
--- a/pkg/minikube/registry/drvs/krunkit/krunkit.go
+++ b/pkg/minikube/registry/drvs/krunkit/krunkit.go
@@ -50,6 +50,8 @@ func init() {
 		Status:   status,
 		Default:  true,
 		Priority: registry.Experimental,
+		// Krunkit driver requires sudo
+		NeedsSudo: true,
 	}); err != nil {
 		panic(fmt.Sprintf("register failed: %v", err))
 	}
--- a/pkg/minikube/registry/drvs/kvm2/kvm2.go
+++ b/pkg/minikube/registry/drvs/kvm2/kvm2.go
@@ -50,6 +50,8 @@ func init() {
 		Status:   status,
 		Default:  true,
 		Priority: registry.Preferred,
+		// KVM2 driver does not require sudo
+		NeedsSudo: false,
 	}); err != nil {
 		panic(fmt.Sprintf("register failed: %v", err))
 	}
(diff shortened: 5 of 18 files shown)


pkg/minikube/driver/driver.go
```diff
@@ -319,6 +321,14 @@ func FlagDefaults(name string) FlagHints {
 func Choices(vm bool) []registry.DriverState {
 	options := registry.Available(vm)
 
+	if !HasSudo() {
+		for i := range options {
+			if options[i].NeedsSudo {
+				options[i].Priority = registry.Discouraged
```
nirs, line 327: We must not do this. Driver priority must not change based on sudo availability.

We can add more info on the driver status, but the priority must not change.

pkg/minikube/driver/driver.go
```diff
@@ -392,10 +402,11 @@ func Status(name string) registry.DriverState {
...
 		return registry.DriverState{
-			Name:     d.Name,
-			Default:  d.Default,
-			Priority: d.Priority,
-			State:    s,
+			Name:      d.Name,
+			Default:   d.Default,
+			Priority:  d.Priority,
+			State:     s,
+			NeedsSudo: d.NeedsSudo,
```
nirs, line 409: This is problematic since this info is dynamic. For example vfkit does not need sudo with the default network (nat), but need sudo for setup with vmnet-shared network. This should be done in the driver documentation and not here.

pkg/minikube/driver/driver.go
```diff
@@ -438,3 +449,8 @@ func IndexFromMachineName(machineName string) int {
 	}
 	return 1 // master node
 }
+
+func realHasSudo() bool {
+	err := exec.Command("sudo", "-n", "true").Run()
+	return err == nil
+}
```
nirs, line 456: This is incorrect check. Nobody configures a system so you can run any command as root without a password. This configuration is use only in throwaway systems like CI VM that is deleted after the CI run, or inside minikube VM where you are not expected to modify the host.

pkg/minikube/registry/drvs/docker/docker.go
```diff
@@ -53,6 +53,8 @@ func init() {
 		Status:   status,
 		Default:  true,
 		Priority: registry.HighlyPreferred,
+		// Docker driver does not require sudo
+		NeedsSudo: false,
```
nirs, line 57: You must have sudo to install docker on Linux or add your self to docker group. This is same as kvm.

pkg/minikube/registry/drvs/hyperkit/hyperkit.go
```diff
@@ -54,6 +54,8 @@ func init() {
 		Status:   status,
 		Default:  true,
 		Priority: registry.Deprecated,
+		// Hyperkit driver requires sudo
+		NeedsSudo: true,
```
nirs, line 58: Minikube has code to configure hyperkit to run without a password, but hyperkit is deprecated so it is not interesting to add this info.

pkg/minikube/registry/drvs/krunkit/krunkit.go
```diff
@@ -50,6 +50,8 @@ func init() {
 		Status:   status,
 		Default:  true,
 		Priority: registry.Experimental,
+		// Krunkit driver requires sudo
+		NeedsSudo: true,
```
nirs, line 54: False, sudo required only to installl vment-helper, same as docker.

pkg/minikube/registry/drvs/kvm2/kvm2.go
```diff
@@ -50,6 +50,8 @@ func init() {
 		Status:   status,
 		Default:  true,
 		Priority: registry.Preferred,
+		// KVM2 driver does not require sudo
+		NeedsSudo: false,
```
nirs, line 54: Requires sudo to install libvirt and add yourself to the libvirt group.

pkg/minikube/registry/drvs/podman/podman.go
```diff
@@ -61,6 +61,8 @@ func init() {
 		Status:   status,
 		Default:  true,
 		Priority: priority,
+		// Podman driver may require sudo (on linux)
+		NeedsSudo: true,
```
nirs, line 65: This depends on the os:
- Linux: requires sudo to install podman. Using rootless mode by default so no need to configure special group
- macOS: does not require sudo to install and run podman (using vfkit and gvproxy)

The minikube driver may have limitations and issues with using podman, it is not not well maintained.

pkg/minikube/registry/drvs/qemu2/qemu2.go
```diff
@@ -53,6 +53,8 @@ func init() {
 		Status:   status,
 		Default:  true,
 		Priority: priority,
+		// QEMU2 driver does not require sudo
+		NeedsSudo: false,
```
nirs, line 57: Wrong, qemu can be installed via brew, but to use socket_vment (e.g to support multiple clusters accessing each other, or multi-node clusters), you need root - same as vmnet-helper.

This also depends on the system:
- Linux: sudo required to install qemu
- macOS: sudo not required to install qemu (e.g. brew install qemu)

pkg/minikube/registry/drvs/ssh/ssh.go
```diff
@@ -41,6 +41,8 @@ func init() {
 		Default:  false, // requires external VM
 		Priority: registry.Discouraged,
 		Init:     func() drivers.Driver { return ssh.NewDriver(ssh.Config{}) },
+		// SSH driver does not require sudo
+		NeedsSudo: false,
```
nirs, line 45: I think you will need sudo on the host you connect to.

pkg/minikube/registry/drvs/vfkit/vfkit.go
```diff
@@ -54,6 +54,8 @@ func init() {
 		Status:   status,
 		Default:  true,
 		Priority: priority,
+		// VFKit driver requires sudo
+		NeedsSudo: true,
```
nirs, line 58: sudo required only for the initial setup to install vment-helper.

nirs (changes requested):
This change is wrong in many ways:
- The idea to change the priority based on sudo requirement is wrong
- The check if sudo is available is wrong
- The information about which drivers requires sudo are mostly wrong

In general almost anything worth using requires sudo , and we should document the requirements. We should not change driver priority based on sudo requirements for setting up the system for the driver.

pkg/minikube/registry/drvs/qemu2/qemu2.go
afbjorklund: It didn't need sudo for the "user" networking, and you don't need sudo if you install in an odd location

pkg/minikube/registry/drvs/podman/podman.go
afbjorklund: The minikube driver for podman was using "sudo podman", and it now has a rootless config.

Since it was so buggy with rootless in the beginning, the default has always been to use rootful

pkg/minikube/registry/drvs/kvm2/kvm2.go
afbjorklund: Those can be done by a sysadmin, without having to use `sudo`

pkg/minikube/driver/driver.go
afbjorklund: We configured minikube to use `sudo -n podman`, since there was no other way (no "podman" group)

So this check seems to match what the driver is using, even if there are some details that is differing.

pkg/minikube/driver/driver.go
afbjorklund: It should be in the "healthy", I think?

afbjorklund:
> In general almost anything worth using requires sudo , and we should document the requirements. We should not change driver priority based on sudo requirements for setting up the system for the driver.

I think there is a difference if minikube itself is calling `sudo`, or if some setup is needed that normally requires root or sudo during the installation...

But if the driver is using sudo, then it is reasonable to check if sudo has been setup. And I was under the impression that was already done, for health?

PR #21689, opened 2025-10-03
Automated generic deprecation warning

divysinghvi:
We were previously were creating warnings for each driver deprecated driver which was causing too much repetative code 
warnDriverDeprecated() function automatically warns if the current driver is marked deprecated
## Before  

```bash
func hyperkitDeprecationWarning(driverName string) {
	if !driver.IsHyperKit(driverName) {
		return
	}
	out.WarningT(`The 'hyperkit' driver is deprecated and will be removed in a future release.
    You can use alternative drivers such as 'vfkit', 'qemu', or 'docker'.
	`)
}
```
we had to create new function for each deprecated driver 

## After
```bash
❗  The 'hyperkit' driver is deprecated and will be removed in a future release.
    You can use alternative drivers: qemu2, vfkit, krunkit, parallels, docker, podman, ssh, hyperkit, virtualbox.
```
automated warning for deprecated drivers 
adds to #21631

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -308,6 +308,7 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 		klog.Errorf("Error autoSetOptions : %v", err)
 	}
 
+	virtualBoxMacOS13PlusWarning(driverName)
 	warnDriverDeprecated(driverName)
 	validateFlags(cmd, driverName)
 	validateUser(driverName)
@@ -399,6 +400,18 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 		Cfg:            &cc,
 		Node:           &n,
 	}, nil
+}
+
+func virtualBoxMacOS13PlusWarning(driverName string) {
+	if !driver.IsVirtualBox(driverName) || !detect.MacOS13Plus() {
+		return
+	}
+	out.WarningT(`Due to changes in macOS 13+ minikube doesn't currently support VirtualBox. You can use alternative drivers such as 'vfkit', 'qemu', or 'docker'.
+    https://minikube.sigs.k8s.io/docs/drivers/vfkit/
+    https://minikube.sigs.k8s.io/docs/drivers/qemu/
+    https://minikube.sigs.k8s.io/docs/drivers/docker/
+    For more details on the issue see: https://github.com/kubernetes/minikube/issues/15274
+`)
 }
 
 // warnDriverDeprecated prints a small deprecation warning if the selected driver is deprecated
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -308,8 +308,7 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 		klog.Errorf("Error autoSetOptions : %v", err)
 	}
 
-	virtualBoxMacOS13PlusWarning(driverName)
-	hyperkitDeprecationWarning(driverName)
+	warnDriverDeprecated(driverName)
 	validateFlags(cmd, driverName)
 	validateUser(driverName)
 	if driverName == oci.Docker {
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -402,29 +401,31 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 	}, nil
 }
 
-func virtualBoxMacOS13PlusWarning(driverName string) {
-	if !driver.IsVirtualBox(driverName) || !detect.MacOS13Plus() {
+// warnDriverDeprecated prints a small deprecation warning if the selected driver is deprecated
+func warnDriverDeprecated(driverName string) {
+	if driverName == "" {
 		return
 	}
```
nirs, line 421: empty driver means the driver was not selected by the user. Do we want to warn only about driver selected by the user?

cmd/minikube/cmd/start.go
```diff
@@ -402,29 +401,31 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
...
-    https://minikube.sigs.k8s.io/docs/drivers/docker/
-    For more details on the issue see: https://github.com/kubernetes/minikube/issues/15274
-`)
-}
 
-// hyperkitDeprecationWarning prints a deprecation warning for the hyperkit driver
-func hyperkitDeprecationWarning(driverName string) {
-	if !driver.IsHyperKit(driverName) {
-		return
+	for _, d := range registry.List() {
```
nirs, line 423: Looping over all drivers is not needed, since we have the driver name. I think we better get the driver from the registry and check if the driver is deprecated.

Can we done using:

```
d := registry.Driver(driverName)
if d.Name == "" {
    // fail, no such driver
}

If d.Priority == registry.Deprecated {
    // show warning
}
```

cmd/minikube/cmd/start.go
```diff
@@ -402,29 +401,31 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
...
+			out.WarningT("The '{{.name}}' driver is deprecated and will be removed in a future release.\n    You can use alternative drivers: {{.drivers}}.", out.V{"name": d.Name, "drivers": strings.Join(driver.SupportedDrivers(), ", ")})
+			return
+		}
+
+		// match any alias
+		for _, a := range d.Alias {
+			if a == driverName {
+				out.WarningT("The '{{.name}}' driver is deprecated and will be removed in a future release.\n    You can use alternative drivers: {{.drivers}}.", out.V{"name": d.Name, "drivers": strings.Join(driver.SupportedDrivers(), ", ")})
+				return
+			}
```
nirs, line 439: It does not make sense to iterate over all drivers and check both driver name and all aliases. The registry keeps a map of drivers by name and alias, so getting a driver should be a single map lookup using the name and if not found a second map lookup using the alias.

cmd/minikube/cmd/start.go
```diff
@@ -402,29 +401,31 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
...
-	if !driver.IsHyperKit(driverName) {
-		return
+	for _, d := range registry.List() {
+		if d.Priority != registry.Deprecated {
+			continue
+		}
+
+		// match driver name
+		if d.Name == driverName {
+			out.WarningT("The '{{.name}}' driver is deprecated and will be removed in a future release.\n    You can use alternative drivers: {{.drivers}}.", out.V{"name": d.Name, "drivers": strings.Join(driver.SupportedDrivers(), ", ")})
```
nirs, line 430: SupportedDrivers() return results based on current OS? Is it sorted by priority?

Previously we used: 

   You can use alternative drivers such as 'vfkit', 'qemu', or 'docker'

This is the wanted order, using driver priority (vfkit is the preferred driver) and preferring vm drivers since on macOS all drivers need a vm. We want to keep this order. If SupportedDrivers() does not return the drivers in the right order for the current OS, we need to improve it before using it here.

nirs (commented):
I like the idea!

cmd/minikube/cmd/start.go
divysinghvi: is it is priority sorted  The 'hyperkit' driver is deprecated and will be removed in a future release.
You can use alternative drivers: qemu2, vfkit, krunkit, parallels, docker, podman, ssh, hyperkit, virtualbox. this is a real result i did added hyperkit and virtualbox to it for testing on my local system rest they won't be visible since they will be deprecated

nirs:
> ## Before
> ```shell
> func hyperkitDeprecationWarning(driverName string) {
> 	if !driver.IsHyperKit(driverName) {
> 		return
> 	}
> 	out.WarningT(`The 'hyperkit' driver is deprecated and will be removed in a future release.
>     You can use alternative drivers such as 'vfkit', 'qemu', or 'docker'.
>     https://minikube.sigs.k8s.io/docs/drivers/vfkit/
>     https://minikube.sigs.k8s.io/docs/drivers/qemu/
>     https://minikube.sigs.k8s.io/docs/drivers/docker/
> 	`)
> }
> ```
> ## After
> ```shell
> ❗  The 'hyperkit' driver is deprecated and will be removed in a future release.
>     You can use alternative drivers: qemu2, vfkit, krunkit, parallels, docker, podman, ssh, hyperkit, virtualbox.
> ```

The new warning does not include links to the alternative drivers docs. Why did you remove the links? This is not required change for adding automated deprecated warnings. Please do not mix unrelated changes.

If you want to remove the links please open an issue and we can discuss this there.

cmd/minikube/cmd/start.go
divysinghvi: My bad i don't think warning for empty driver should be part of this function

cmd/minikube/cmd/start.go
nirs: The order is wrong - vfkit is the preferred driver and should be first, not qemu. This is the native driver for macOS. qemu on macOS is poorly maintained and should not be the preferred driver.

Also the hyperkit driver appears in the list of alternatives which is wrong - the hyperkit driver is not an alternative to itself.

We need to create AlternativeDrivers() function - SupportedDrivers() is not the right way.

cmd/minikube/cmd/start.go
nirs: sure, we should not warn about empty driverName, and the check achieve this. But can get this for free by getting the driver and return on missing driver.

cmd/minikube/cmd/start.go
nirs: Note that with driverName "" we will get a DriverDef with empty name, and this is not an error, so we should not fail, just return.

The check for valid driver should be done before we call this function, for example if a user makes a typo in the driver name, or the driver is not available on this OS.

divysinghvi:
@nirs thanks for the comment will keep in mind not to mix unrelated things , but i don't understand like   https://minikube.sigs.k8s.io/docs/drivers/vfkit/
since these are mainly valid for the mac os only right but when we are trying to be do  generalised warning won't it be bad to include specific driver link that might be unrelated to that OS like links will vary for other OS if i deprecated a windows driver? what do you think

nirs:
> @nirs thanks for the comment will keep in mind not to mix unrelated things , but i don't understand like https://minikube.sigs.k8s.io/docs/drivers/vfkit/ https://minikube.sigs.k8s.io/docs/drivers/qemu/ https://minikube.sigs.k8s.io/docs/drivers/docker/ 

What do you mean by "I don't understand like .."?

> since these are mainly valid for the mac os only right but when we are trying to be do generalised warning won't it be bad to include specific driver link that might be unrelated to that OS like links will vary for other OS if i deprecated a windows driver? what do you think

The warning is generated - it is not static, so it should never include drivers which are not available on the current OS, and all the links to the docs will also be generated, so they will always be relevant to the current os.

A DriverDef should include Name and DocURL. In this warning we should get list of alternative drivers, which should be drivers which are not deprecated or discouraged, and available on the current OS and architecture.

For example when running on darwin/arm64 we should not recommend drivers which available only on amd64. So we should not have such a long list of alternative as you show in your example.

We can generate the links form the Driver.DockURLs - in the same order of the alternative drivers.

divysinghvi:
Thanks for explaining now i understand it way better i was a bit confused , and also the long list is also useless also since have the Driver.DockURLs  it would be way easier to create docs link did not knew about that

nirs:
@divysinghvi to finish this change with minimal changes, I think we can do this:

- Add DocURL fields to drivers in the registry
- Look if we already have function returning alternative drivers (may be Suggests()?). If we have use it to create the list of alternative driver names.
- If we don't have such function, add AlternativeDrivers(driverName) function build a list from Available() by removing driverName
- Update the hard coded drivers list in the driver package to match the driver priority in the registry.

divysinghvi:
Thanks @nirs for detailed path i will look into it and start working on it function by function and will ask for help if any needed :)

PR #21717, opened 2025-10-06
deb automation

medyagh:
- **add debian get_version automation**
- **fix: update  gh workflow automation to use debian to get version**
- **fix: fix automtion to track get most recent  dated tag**

### DIFF
--- a/hack/update/debian_version/debian_version.go
+++ b/hack/update/debian_version/debian_version.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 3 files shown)


hack/update/debian_version/debian_version.go
```diff
@@ -40,13 +41,36 @@ type Data struct {
...
+// resulting image digest remains stable and predictable between runs.
 func latestBookwormSlimTag(tags []string) (string, error) {
+	var newestDateTag string
 	for _, tag := range tags {
-		if strings.Contains(tag, "bookworm-slim") {
-			return tag, nil
+		// Skip anything that isn't a bookworm slim tag to avoid matching other
+		// Debian variants.
+		if !strings.HasPrefix(tag, "bookworm-") || !strings.HasSuffix(tag, "-slim") {
+			continue
```
nirs, line 57: Why do we need this? we match the tag using regex bellow.

hack/update/debian_version/debian_version.go
```diff
@@ -40,13 +41,36 @@ type Data struct {
...
+		// the most recent date stamp provided by Debian.
+		if bookwormDateTag.MatchString(tag) {
+			if newestDateTag == "" || tag > newestDateTag {
+				newestDateTag = tag
+			}
 		}
 	}
-	return "", fmt.Errorf("no tag found that matches: bookworm-slim")
+
+	// Prefer a dated tag when available.
```
nirs: This seems outdated - we fail if we did not find a dated tag.

hack/update/debian_version/debian_version.go
medyagh: to avoid weird situations, it add a layer of Skip if it is not remotely correct.

PR #21737, opened 2025-10-13
test: Fix TestGetCoreClient on Windows

bobsira:
In this simple fix:

- t.TempDir() returns a directory guaranteed to exist and writable (e.g., %TEMP%\... on Windows, typically /tmp/... on Linux/macOS).
- filepath.Join produces the correct path separator for the current OS.
- No hard-coded absolute Unix path remains.

### DIFF
--- a/pkg/minikube/service/service_test.go
+++ b/pkg/minikube/service/service_test.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path/filepath"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


bobsira:
@nirs @medyagh can I get eyes on this PR

pkg/minikube/service/service_test.go
```diff
@@ -712,50 +713,49 @@ preferences: {}
...
-			err := os.WriteFile(mockK8sConfigPath, mockK8sConfigByte, 0644)
-			defer os.Remove(mockK8sConfigPath)
-			if err != nil {
-				t.Fatalf("Unexpected error when writing to file %v. Error: %v", test.kubeconfigPath, err)
+			// Each sub-test gets its own temp dir & file for isolation and portability.
+			tmpDir := t.TempDir()
+			mockK8sConfigPath := filepath.Join(tmpDir, "kube_config")
+
+			// Only write the file if we have any content; an empty string still creates a 0-byte file,
+			// which is what we want to simulate "empty config".
```
nirs: This comment contradicts the code, we always write a file, even if we don't have any content.

pkg/minikube/service/service_test.go
```diff
@@ -712,50 +713,49 @@ preferences: {}
...
 
 	for _, test := range tests {
 		t.Run(test.description, func(t *testing.T) {
-			mockK8sConfigByte := []byte(test.config)
-			mockK8sConfigPath := test.kubeconfigPath
-			err := os.WriteFile(mockK8sConfigPath, mockK8sConfigByte, 0644)
-			defer os.Remove(mockK8sConfigPath)
-			if err != nil {
-				t.Fatalf("Unexpected error when writing to file %v. Error: %v", test.kubeconfigPath, err)
+			// Each sub-test gets its own temp dir & file for isolation and portability.
```
nirs: This is standard practice, I don't think we ned to justify it. All tests should work like this.

pkg/minikube/service/service_test.go
```diff
@@ -712,50 +713,49 @@ preferences: {}
...
+			tmpDir := t.TempDir()
+			mockK8sConfigPath := filepath.Join(tmpDir, "kube_config")
+
+			// Only write the file if we have any content; an empty string still creates a 0-byte file,
+			// which is what we want to simulate "empty config".
+			if err := os.WriteFile(mockK8sConfigPath, []byte(test.config), 0644); err != nil {
```
nirs: There is no reason to create the kubeconfig world readable. This is the default kubeconfig permissions on macOS:

```
% ls -lh ~/.kube/config
-rw-------  1 nir  staff   106B Oct 10 11:37 /Users/nir/.kube/config
```

It will be better to use real world permissions. This will break the code if it has bad assumptions about being able to read the config from another user in the same group.

pkg/minikube/service/service_test.go
```diff
@@ -712,50 +713,49 @@ preferences: {}
...
 			k8s := K8sClientGetter{}
-			_, err = k8s.GetCoreClient("minikube")
+			_, err := k8s.GetCoreClient("minikube")
+
 			if err != nil && !test.err {
 				t.Fatalf("GetCoreClient returned unexpected error: %v", err)
 			}
 			if err == nil && test.err {
-				t.Fatal("GetCoreClient expected to return error but got nil")
+				t.Fatalf("GetCoreClient expected to return error but got nil")
```
nirs: Why t.Fatalf() if we don't format anything?

pkg/minikube/service/service_test.go
bobsira: Removing this comment!

pkg/minikube/service/service_test.go
bobsira: Removing this comment as well!

pkg/minikube/service/service_test.go
bobsira: Replacing with 

```
if err := os.WriteFile(mockK8sConfigPath, []byte(test.config), 0600); err != nil {
				t.Fatalf("failed to write kubeconfig: %v", err)
			}
```

 0600 is better because

- Aligns with kubectl’s default and common distro behavior (owner read/write only).
- Avoids tests silently passing due to overly permissive access that won’t exist in production.
- Makes permission-related bugs more visible (e.g. code accidentally running as another user expecting read access).
- Reduces the risk of accidental leakage of credentials in environments where test artifacts persist.

pkg/minikube/service/service_test.go
bobsira: Nice catch, replaced this with t.Fatal()

bobsira:
@nirs  @medyagh this PR is ready to approval and merge!

PR #21738, opened 2025-10-13
test: Fix TestGetStoragev1 unit test on windows

bobsira:
Summary of fix

- Removed hard-coded /tmp usage that broke on Windows.
- Each subtest in TestGetStoragev1 now creates its own temp directory via t.TempDir() and builds kubeconfig path with filepath.Join (cross-platform).
- Helper setK8SConfig ensures the parent dir exists (os.MkdirAll) and writes the config, then sets KUBECONFIG.
- Per-subtest isolation eliminates shared file/state, avoids cleanup code, and works on all OSes.

### DIFF
--- a/pkg/minikube/storageclass/storageclass_test.go
+++ b/pkg/minikube/storageclass/storageclass_test.go
@@ -252,11 +252,8 @@ func TestGetStoragev1(t *testing.T) {
 }
 
 func setK8SConfig(t *testing.T, config, kubeconfigPath string) error {
-	if err := os.MkdirAll(filepath.Dir(kubeconfigPath), 0o755); err != nil {
-		return fmt.Errorf("failed to create kubeconfig directory: %w", err)
-	}
 	if err := os.WriteFile(kubeconfigPath, []byte(config), 0o644); err != nil {
-		return fmt.Errorf("unexpected error when writing to %v: %w", kubeconfigPath, err)
+		t.Fatalf("unexpected error when writing to %v: %v", kubeconfigPath, err)
 	}
 	t.Setenv("KUBECONFIG", kubeconfigPath)
 	return nil


pkg/minikube/storageclass/storageclass_test.go
```diff
@@ -252,12 +252,12 @@ func TestGetStoragev1(t *testing.T) {
...
 
 func setK8SConfig(t *testing.T, config, kubeconfigPath string) error {
-	mockK8sConfigByte := []byte(config)
-	mockK8sConfigPath := kubeconfigPath
-	err := os.WriteFile(mockK8sConfigPath, mockK8sConfigByte, 0644)
-	if err != nil {
-		return fmt.Errorf("Unexpected error when writing to file %v. Error: %v", kubeconfigPath, err)
+	if err := os.MkdirAll(filepath.Dir(kubeconfigPath), 0o755); err != nil {
+		return fmt.Errorf("failed to create kubeconfig directory: %w", err)
+	}
```
nirs: We don't need to create a directory since we use t.TempDir(), which creates a new directory on every call.

pkg/minikube/storageclass/storageclass_test.go
```diff
@@ -252,12 +252,12 @@ func TestGetStoragev1(t *testing.T) {
...
+	if err := os.WriteFile(kubeconfigPath, []byte(config), 0o644); err != nil {
+		return fmt.Errorf("unexpected error when writing to %v: %w", kubeconfigPath, err)
 	}
-	t.Setenv("KUBECONFIG", mockK8sConfigPath)
+	t.Setenv("KUBECONFIG", kubeconfigPath)
```
nirs, line 258: This is really bad and not thread safe. We should never set the env since it is process level value and we may have several tests running in the same time that may use the wrong KUBECONFIG. Since this issue was not introduced by your change we can improve it later.

pkg/minikube/storageclass/storageclass_test.go
```diff
@@ -252,12 +252,12 @@ func TestGetStoragev1(t *testing.T) {
...
-	mockK8sConfigByte := []byte(config)
-	mockK8sConfigPath := kubeconfigPath
-	err := os.WriteFile(mockK8sConfigPath, mockK8sConfigByte, 0644)
-	if err != nil {
-		return fmt.Errorf("Unexpected error when writing to file %v. Error: %v", kubeconfigPath, err)
+	if err := os.MkdirAll(filepath.Dir(kubeconfigPath), 0o755); err != nil {
+		return fmt.Errorf("failed to create kubeconfig directory: %w", err)
+	}
+	if err := os.WriteFile(kubeconfigPath, []byte(config), 0o644); err != nil {
+		return fmt.Errorf("unexpected error when writing to %v: %w", kubeconfigPath, err)
```
nirs: We should really call t.Fatalf() here instead of returning an error, but since this issue was not introduced by your change we can improve it later.

pkg/minikube/storageclass/storageclass_test.go
bobsira: Dropping the redundant directory creation

pkg/minikube/storageclass/storageclass_test.go
bobsira: Noted! Made this change.

nirs (approved):

bobsira:
@medyagh this PR is ready to be merged!

PR #21739, opened 2025-10-13
test: Fix TestMiniPath unit test on windows

bobsira:

Running tool: C:\Program Files\Go\bin\go.exe test -timeout 30s -run ^TestMiniPath$ k8s.io/minikube/pkg/minikube/localpath

```
=== RUN   TestMiniPath
=== RUN   TestMiniPath//tmp/.minikube
    c:\dev\minikube\pkg\minikube\localpath\localpath_test.go:78: MiniPath expected to return '\tmp\.minikube', but got '/tmp/.minikube'
--- FAIL: TestMiniPath//tmp/.minikube (0.00s)
=== RUN   TestMiniPath//tmp/
--- PASS: TestMiniPath//tmp/ (0.00s)
=== RUN   TestMiniPath/#00
--- PASS: TestMiniPath/#00 (0.00s)
--- FAIL: TestMiniPath (0.00s)
FAIL
FAIL    k8s.io/minikube/pkg/minikube/localpath  4.151s
```

**Test Run with the fix**

```
Running tool: C:\Program Files\Go\bin\go.exe test -timeout 30s -run ^TestMiniPath$ k8s.io/minikube/pkg/minikube/localpath

=== RUN   TestMiniPath
=== RUN   TestMiniPath//tmp/.minikube
--- PASS: TestMiniPath//tmp/.minikube (0.00s)
=== RUN   TestMiniPath//tmp/
--- PASS: TestMiniPath//tmp/ (0.00s)
=== RUN   TestMiniPath/#00
--- PASS: TestMiniPath/#00 (0.00s)
--- PASS: TestMiniPath (0.00s)
PASS
ok      k8s.io/minikube/pkg/minikube/localpath  (cached)
```

### DIFF
--- a/pkg/minikube/localpath/localpath.go
+++ b/pkg/minikube/localpath/localpath.go
@@ -44,13 +44,15 @@ func ConfigFile() string {
 func MiniPath() string {
 	minikubeHomeEnv := os.Getenv(MinikubeHome)
 	if minikubeHomeEnv == "" {
-		return filepath.Join(homedir.HomeDir(), ".minikube")
+		return filepath.Join(filepath.Clean(homedir.HomeDir()), ".minikube")
 	}
-	if filepath.Base(minikubeHomeEnv) == ".minikube" {
-		// Normalize to platform-specific separators for consistency on Windows.
-		return filepath.Clean(filepath.FromSlash(minikubeHomeEnv))
+
+	norm := filepath.Clean(minikubeHomeEnv)
+
+	if filepath.Base(norm) == ".minikube" {
+		return norm
 	}
-	return filepath.Join(minikubeHomeEnv, ".minikube")
+	return filepath.Join(norm, ".minikube")
 }
 
 // MakeMiniPath is a utility to calculate a relative path to our directory.


pkg/minikube/localpath/localpath.go
```diff
@@ -47,7 +47,8 @@ func MiniPath() string {
 		return filepath.Join(homedir.HomeDir(), ".minikube")
 	}
 	if filepath.Base(minikubeHomeEnv) == ".minikube" {
-		return minikubeHomeEnv
+		// Normalize to platform-specific separators for consistency on Windows.
+		return filepath.Clean(filepath.FromSlash(minikubeHomeEnv))
```
nirs: This is value from the MINIKUBE_HOME environment variable. Why do we need to fix value set by the user? they should set the right value, and it the path is invalid the minikube command should fail.

If the issue comes from a test using wrong path, we should fix the test, not the code.

pkg/minikube/localpath/localpath.go
```diff
@@ -47,7 +47,8 @@ func MiniPath() string {
...
 	}
 	return filepath.Join(minikubeHomeEnv, ".minikube")
```
nirs: However if we need to fix the value for the user, why we don't need to fix it in this case?

pkg/minikube/localpath/localpath.go
bobsira: We’re not correcting a bad user value; we’re standardizing a valid path so it has one canonical string form. Windows accepts forward slashes, but raw string comparisons (tests, map keys, logs) treat “C:/Users/Alice/.minikube” and “C:\Users\Alice.minikube” as different. Without normalization this causes flaky tests, cache/key mismatches, and forces every caller to remember to clean paths. Centralizing normalization in MiniPath() creates a consistent invariant (clean, native separators, no trailing slash), reduces maintenance, and doesn’t hide truly invalid paths (bad paths still fail later). So the change solves a real cross‑platform consistency problem, not just a test artifact.

pkg/minikube/localpath/localpath.go
bobsira: Good point—the most consistent approach is to canonicalize the final result unconditionally, not just one branch. The branch-specific change was a minimal surgical patch targeting the exact failing scenario (raw env already points at `.minikube`). A cleaner follow-up is:

````go
func MiniPath() string {
	minikubeHomeEnv := os.Getenv(MinikubeHome)
	var p string
	switch {
	case minikubeHomeEnv == "":
		p = filepath.Join(homedir.HomeDir(), ".minikube")
	case filepath.Base(minikubeHomeEnv) == ".minikube":
		p = minikubeHomeEnv
	default:
		p = filepath.Join(minikubeHomeEnv, ".minikube")
	}
	return filepath.Clean(filepath.FromSlash(p))
}
````

**Comment by bobsira** [comment]
@nirs 

```
func TestMiniPath(t *testing.T) {
	var testCases = []struct {
		env, basePath string
	}{
		{"/tmp/.minikube", "/tmp/"},
		{"/tmp/", "/tmp"},
		{"", homedir.HomeDir()},
	}
		t.Run(tc.env, func(t *testing.T) {
			expectedPath := filepath.Join(tc.basePath, ".minikube")
			t.Setenv(MinikubeHome, tc.env)
			path := MiniPath()
			if path != expectedPath {
				t.Errorf("MiniPath expected to return '%s', but got '%s'", expectedPath, path)
			}
		})
	}
}
```


Looking at TestMiniPath it fails on Windows because users often set MINIKUBE_HOME with forward slashes (e.g. C:/Users/Alice/.minikube), and the current implementation of MiniPath()  returned that raw value unchanged when it already ended in .minikube, while other branches build the path using filepath.Join, which on Windows emits backslashes; the filesystem accepts both, but our test performs a strict string equality, so the expected path (backslashes) and actual path (forward slashes) diverge despite referring to the same directory. By normalizing the chosen path with filepath.Clean(filepath.FromSlash(...)), we enforce one canonical, native-separator, no-trailing-slash representation, eliminating flaky equality failures, preventing subtle cache/map key mismatches, and centralizing path cleanup so every caller gets a consistent result without masking truly invalid paths (bad values still fail later during actual file operations).

**Comment by nirs** [comment]
@bobsira So this is not a test issue, but issue with using both unix paths (/path) and windows paths (C:\path). Normalizing all input paths makes sense but it should be done early as possible.

**Comment by bobsira** [comment]
> @bobsira So this is not a test issue, but issue with using both unix paths (/path) and windows paths (C:\path). Normalizing all input paths makes sense but it should be done early as possible.

@nirs Exactly, I propose the following modification. If you are alright with it, I'll push the change.


```
func MiniPath() string {
    minikubeHomeEnv := os.Getenv(MinikubeHome)
    if minikubeHomeEnv == "" {
        return filepath.Join(homedir.HomeDir(), ".minikube")
    }

    // Early normalization: collapse any forward slashes and clean dot segments.
    minikubeHomeEnv = filepath.Clean(filepath.FromSlash(minikubeHomeEnv))

    if filepath.Base(minikubeHomeEnv) == ".minikube" {
        return minikubeHomeEnv
    }
    return filepath.Join(minikubeHomeEnv, ".minikube")
}
```

@medyagh we will have this as a product fix instead of a test issue

**Comment by nirs** [comment]
> > @bobsira So this is not a test issue, but issue with using both unix paths (/path) and windows paths (C:\path). Normalizing all input paths makes sense but it should be done early as possible.
> 
> @nirs Exactly, I propose the following modification. If you are alright with it, I'll push the change.
> 
> ```
> func MiniPath() string {
>     minikubeHomeEnv := os.Getenv(MinikubeHome)
>     if minikubeHomeEnv == "" {
>         return filepath.Join(homedir.HomeDir(), ".minikube")

homedir.HomeDir() returns a normalized path?

>     }
> 
>     // Early normalization: collapse any forward slashes and clean dot segments.
>     minikubeHomeEnv = filepath.Clean(filepath.FromSlash(minikubeHomeEnv))

Are you sure we need both FromSlash and Clean?

The docs says:

> On Windows, Clean does not modify the volume name other than to replace occurrences of "/" with `\`. For example, Clean("//host/share/../x") returns `\\host\share\x`.

So it seems that Clean is enough. I did not test this on Windows.

>     if filepath.Base(minikubeHomeEnv) == ".minikube" {
>         return minikubeHomeEnv
>     }
>     return filepath.Join(minikubeHomeEnv, ".minikube")
> }
> ```

Do we have a test convering this code? Let's add one if not. The test will tell us if Clean is enough or we need also FromSlash.

**Comment by nirs** [comment]
@bobsira what about other paths? for example --mount-string. Ideally we normalize all paths on input, so inside minikube all paths are clean.

**Comment by bobsira** [comment]
> > > @bobsira So this is not a test issue, but issue with using both unix paths (/path) and windows paths (C:\path). Normalizing all input paths makes sense but it should be done early as possible.
> > 
> > 
> > @nirs Exactly, I propose the following modification. If you are alright with it, I'll push the change.
> > ```
> > func MiniPath() string {
> >     minikubeHomeEnv := os.Getenv(MinikubeHome)
> >     if minikubeHomeEnv == "" {
> >         return filepath.Join(homedir.HomeDir(), ".minikube")
> > ```
> 
> homedir.HomeDir() returns a normalized path?
>
> > >  @nirs 
> > > Great question, no, homedir.HomeDir() just returns the raw home directory environment variables; it does not actively 
> > > normalize. So if we need a normalized invariant for subsequent joins or comparisons, we will have to do it ourselves. I'll  have  a revised MiniPath() for this.
>
>We can safely drop filepath.FromSlash here; filepath.Clean alone is enough for the normalization we need on Windows. 
>
>
>
> > ```
> > }
> > 
> > // Early normalization: collapse any forward slashes and clean dot segments.
> > minikubeHomeEnv = filepath.Clean(filepath.FromSlash(minikubeHomeEnv))
> > ```
> 
> Are you sure we need both FromSlash and Clean?
> 
> The docs says:
> 
> > On Windows, Clean does not modify the volume name other than to replace occurrences of "/" with `\`. For example, Clean("//host/share/../x") returns `\\host\share\x`.
> 
> So it seems that Clean is enough. I did not test this on Windows.
> 
> > ```
> > if filepath.Base(minikubeHomeEnv) == ".minikube" {
> >     return minikubeHomeEnv
> > }
> > return filepath.Join(minikubeHomeEnv, ".minikube")
>> }
> > ```
> > 
> > > @nirs 
> > > We can safely drop filepath.FromSlash here; filepath.Clean alone is enough for the normalization we need on Windows.
> > >
> > >What filepath.Clean does:
> > >
> > > - Removes redundant separators.
> > > - Eliminates . segments and resolves .. within the path string (without hitting the filesystem).
> > > - Converts all / to \ on Windows (while preserving UNC/volume semantics).
> > > - Removes a trailing separator (except for root / or drive roots like ).
> > > - Collapses mixed slash usage (e.g. C:/foo\\bar//baz → C:\foo\bar\baz).
> > > 
> > > What filepath.FromSlash does:
> > > 
> > > - Replaces every / with os.PathSeparator (on Windows: \; on Unix: stays /).
> > > - It does not clean dot segments or collapse duplicates.

> >   
> > 
> 
> Do we have a test covering this code? Let's add one if not. The test will tell us if Clean is enough or we need also FromSlash.
> > > I'll get rid of FromSlash

**Comment by bobsira** [comment]
> @bobsira what about other paths? for example --mount-string. Ideally we normalize all paths on input, so inside minikube all paths are clean.

>> @nirs This PR limits scope to normalizing MINIKUBE_HOME inside MiniPath(). I agree other user inputs (e.g. --mount-string, addon dirs, image/cache paths) should be normalized early for consistency. We should open a follow-up issue to inventory and standardize normalization at parse time so internal code always works with canonical paths. Keeping this change focused reduces risk and unblocks windows test stability now.

File: pkg/minikube/localpath/localpath.go [hunk 3]
```diff
@@ -44,12 +44,15 @@ func ConfigFile() string {
 func MiniPath() string {
```
medyagh, line 44: not sure what this is doing, but if you add unit tests for the Exact Failed Tests that we have now, we can see in in the unit test better.

plz add unit test

medyagh, line 44: please add multiple unit test cases for this func

pkg/minikube/localpath/localpath.go
```diff
@@ -44,12 +44,15 @@ func ConfigFile() string {
...
 	minikubeHomeEnv := os.Getenv(MinikubeHome)
 	if minikubeHomeEnv == "" {
-		return filepath.Join(homedir.HomeDir(), ".minikube")
+		return filepath.Join(filepath.Clean(homedir.HomeDir()), ".minikube")
```
mloskot, line 47: Slim chance, but given [current implementation of HomeDir](https://github.com/kubernetes/client-go/blob/9bcb69436287b966d0c5c195efef00aed921fb1b/util/homedir/homedir.go#L88-L89) and the fact that [`filepath.Clean("") == "."`](https://cs.opensource.google/go/go/+/refs/tags/go1.25.5:src/internal/filepathlite/path.go;drc=89af77deef7e554dff5ca21ff2c2aaf0a2d253ac;l=74)  this can return `.\.minikube` on Windows. It may be worth a comment or a even a fallback to `C:\minikube`.

mloskot:
@nirs 
> homedir.HomeDir() returns a normalized path?

Although current implementation does not normalize it, chances seem slim for having the OS-controlled `HOME` or `USERPROFILE` environment variables set with non-normalised or illformed paths.

pkg/minikube/localpath/localpath.go
mloskot: For the records, I [discussed that on slack](https://kubernetes.slack.com/archives/C1F5CT6Q1/p1767914908161729) saying:
> @bobsira  is fixing an issue in `MiniPath` that has already been revealed by `TestMiniPath`, so I'm unclear what unit tests coverage is missing there.
> IMHO, this PR is ready to merge as it seems to correctly address one of the failing unit tests.
> There may be room for improvements in the `MiniPath` and handling of paths in other places, what @bobsira explains in https://github.com/kubernetes/minikube/pull/21739#issuecomment-3470266542, but that seems like deserving separate PRs.

mloskot: see https://github.com/kubernetes/minikube/pull/21739#discussion_r2676357089

pkg/minikube/localpath/localpath.go
medyagh, line 47: lets use path.Join(

mloskot: Later during the OH, we decided that, before we move to exploring `path.Join` blast radius, we should try a minimal fix for the tests. I've opened a new PR #22447

medyagh:
replaced by https://github.com/kubernetes/minikube/pull/22447

Issue #21178, opened 2025-07-28
Merge --mount and --mount-string flags


medyagh:
we dont need both flags, if you provide --mount-string, it shoud also mean you want --mount to be true

and we should Deprecate --mount-string
and for users that provide the old way, we should warning and TRY to do be our best to translate

also seems like we also have mount-options (super confusing) too many mount related things
```
      --mount                             This will start the mount daemon and automatically mount files into minikube.
      --mount-9p-version string           Specify the 9p version that the mount should use (default "9p2000.L")
      --mount-gid string                  Default group id used for the mount (default "docker")
      --mount-ip string                   Specify the ip that the mount should be setup on
      --mount-msize int                   The number of bytes to use for 9p packet payload (default 262144)
      --mount-options strings             Additional mount options, such as cache=fscache
      --mount-port uint16                 Specify the port that the mount should be setup on, where 0 means any free port.
      --mount-string string               The argument to pass the minikube mount command on start.
      --mount-type string                 Specify the mount filesystem type (supported types: 9p) (default "9p")
      --mount-uid string                  Default user id used for the mount (default "docker")
```

seems like we had added mount-string in 2017 even before @tstromberg and I were in minikube

https://github.com/kubernetes/minikube/blob/master/CHANGELOG.md#version-0191---5302017

nirs:
## Plan for unified mounts as discussed in slack

### minikube 1.36

Support single mount using 9p protocol:

```
minikube start --mount=true --mount-string /host-path:/guest-path --mount-options cache=fscache
```

This will not work since --mount expects an argument:

```
minikube start --mount --mount-string /host-path:/guest-path
```

Maybe we can find a way to treat --mount and --mount=true in the same way. We want to use StringArrayVar to support multiple arguments for the new way.

### minikube 1.37

Support single mount using 9p protocol or virtiofs (depending on the driver):

```
minikube start --mount-string /host-path:/guest-path --mount-options cache=fscache
```

The `--mount` flag is silently ignored - nothing is mounted since we don't have a default value for `--mount-string`

```
minikube start --mount
```

### New way - minikube 1.38

Drivers supporting virtiofs can accept multiple mount arguments. Other drivers can accept single mount and use 9p.

Example multiple mounts for vfkit and krunkit (using virtiofs):

```
minikube start --mount /host-path1:/guest-path1 --mount /host-path2:/guest-path2
```

Example single mount for all drivers (using virtiofs or fallback to 9p):

```
[...]

PR #21741, opened 2025-10-13
Add warn about using the ignored --mount option

### LINKED ISSUE #21291: Warn about using the ignored --mount option
Since #21250 the `--mount` flag is ignored and --mount-string default was removed. Users that used to run:

    minikube start --mount

Should run now:

    minikube start --mount-string $HOME/shared/dir

We want to help the user use the new way, and warn them about exposing the home directory to the guest.

## Fix

- Detect usage of --mount with unchanged --mount-string and recommend the new way, recommending a specific shared directory instead of $HOME
- Detect usage of --mount with changed --mount-string and warn that the --mount flag is ignored and not needed

henry3260:
fix #21291

We have 2 cases:

- `--mount`: user expects to have HOME mounted in the guest. This does nothing and we want to show an example how to mount a specific directory in the guest.
- `--mount --mount-string ...`: warn that the `--mount` flag is ignored and can be removed

### After:
First Case:
image

Second Case:
image

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -269,6 +269,10 @@ func runStart(cmd *cobra.Command, _ []string) {
 
 	validateBuiltImageVersion(starter.Runner, ds.Name)
 
+	if viper.GetBool(createMount) {
+		warnAboutMountFlag()
+	}
+
 	if existing != nil && driver.IsKIC(existing.Driver) && viper.GetString(mountString) != "" {
 		old := ""
 		if len(existing.ContainerVolumeMounts) > 0 {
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -269,6 +269,10 @@ func runStart(cmd *cobra.Command, _ []string) {
```
medyagh, line 269: I think instead of warnining, we should merge the 
--mount 
and 
--mount-string flags into

--mount

and if ppl enter --mount true --mount-string /blahblah:/blahblah

we could still support it
with good unit testing we could just transition into "--mount " flag without breaking old syntax users

henry3260:
@nirs Thanks for reviewing! I’d like to ask for your guidance here.

I noticed an unexpected behavior:

When running:
    `./out/minikube start --mount="/home/henry:/mnt" --alsologtostderr --v=2 2>&1 | grep -i mount
`
the boolean/string parsing logic in `start.go` is triggered correctly.
However, the same flag condition inside `start_flags.go` (where `interpretMountFlag(cmd)` is called) does **not** execute.

Both conditions check `cmd.Flags().Changed(createMount)`, but they produce different results at runtime.

Do you have any hints why the flag state behaves differently between these two code paths? I suspect the config is initialized too early, but I’m not fully sure where the state gets overridden. Any suggestions or directions would be appreciated!

henry3260:
> Please do not merge --mount and mount-string in this PR. This is not related to the issue.
> 
> We have another issue for having the new --mount flags supporting multiple mounts. We don't want to introduce the new behavior supporting only one mount.
> 
> See [#21178 (comment)](https://github.com/kubernetes/minikube/issues/21178#issuecomment-3152280321)

So do we still need to add a warning about using the --mount option in this pr?

nirs:
> > Please do not merge --mount and mount-string in this PR. This is not related to the issue.
> > We have another issue for having the new --mount flags supporting multiple mounts. We don't want to introduce the new behavior supporting only one mount.
> > See [#21178 (comment)](https://github.com/kubernetes/minikube/issues/21178#issuecomment-3152280321)
> 
> So do we still need to add a warning about using the --mount option in this pr?

If we introduce the new --mount flag we don't need a warning. But this PR is fine for now since we don't have the new --mount flag yet. I would remove the second commit and keep it as is in case we cannot finish the new --mount flag in time of the next release.

PR #21770, opened 2025-10-18
Add initial image commands to smoke test (partial implementation)

### LINKED ISSUE #21154: tests: Test image operations in smoke test
Currently we test only starting, stopping and deleting cluster, so we don't catch regression like #21103. This makes it harder to contribute and maintain minikube.

We have image tests the functional test but they assume docker daemon and fail on macOS. We can test using minikube cluster and external repo with a tiny image (such as the alpine image).

The image commands do not require kubernetes and are very lightweight, so we can test them in GitHub runners.

Add image tests to the smoke test:
- minikube image load docker.io/library/alpine
- minikube image load image.tar
- minikube image build
- minikube image ls
- minikube image pull
- minikube image push
- minikube image rm
- minikube image save
- minikube image tag

henry3260:
Fixes #21154 

Currently, it adds partial support for image-related smoke tests. 
The following commands are implemented so far:

- minikube image load pause.tar
- minikube image build
- minikube image ls
- minikube image pull registry.k8s.io/pause
- minikube image rm
- minikube image save
- minikube image tag

These commands can run in GitHub runners and do not require Kubernetes. 

Should we include the image push operation in the smoke test, even though we can't truly validate its success?

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -126,6 +126,34 @@ jobs:
               --driver ${{ matrix.driver }} \
               ${{ matrix.network_flag }} \
               ${{ env.LOG_ARGS }}
+      - name: Copy testdata for image build
+        run: |
+          cp -r test/integration/testdata ./out
+      - name: minikube image pull
+        run: |
+          ./out/minikube image pull registry.k8s.io/pause ${{ env.LOG_ARGS }}
+      - name: minikube image save
+        run: |
+          echo ">>> Saving registry.k8s.io/pause image to pause.tar"
+          ./out/minikube image save registry.k8s.io/pause pause.tar ${{ env.LOG_ARGS }}
+      - name: minikube image build
+        run: |
+          ./out/minikube image build -t test-image ./out/testdata/docker-env ${{ env.LOG_ARGS }}
+      - name: minikube image tag
+        run: |
+          echo ">>> Tagging test-image as test-image:smoke"
+          ./out/minikube image tag test-image test-image:smoke ${{ env.LOG_ARGS }}
+      - name: minikube image ls
+        run: |
+          ./out/minikube image ls ${{ env.LOG_ARGS }}
+      - name: minikube image rm
+        run: |
+          echo ">>> Removing test-image"
+          ./out/minikube image rm test-image ${{ env.LOG_ARGS }}
+      - name: minikube image load pause.tar
+        run: |
+          echo ">>> Loading image from pause.tar into minikube"
+          ./out/minikube image load pause.tar ${{ env.LOG_ARGS }}
       - name: Inspect minikube
         if: always()
         run: |


.github/workflows/smoke-test.yml
```diff
@@ -126,6 +126,13 @@ jobs:
               --driver ${{ matrix.driver }} \
               ${{ matrix.network_flag }} \
               ${{ env.LOG_ARGS }}
+      - name: minikube image load
+        run: |
+          ./out/minikube image load docker.io/library/alpine ${{ env.LOG_ARGS }}
+      - name: minikube image build
+        run: |
+          ./out/minikube image build -t test-image . ${{ env.LOG_ARGS }}
```
nirs: image build needs a Containerfile. You can create one in this test.

.github/workflows/smoke-test.yml
```diff
@@ -126,13 +126,24 @@ jobs:
...
+          ./out/minikube image save docker.io/library/alpine alpine.tar ${{ env.LOG_ARGS }}
+      - name: minikube image load alpine.tar
+        run: |
+          echo ">>> Loading image from alpine.tar into minikube"
+          ./out/minikube image load alpine.tar ${{ env.LOG_ARGS }}
       - name: minikube image build
         run: |
-          ./out/minikube image build -t test-image . ${{ env.LOG_ARGS }}
-          
+          ./out/minikube image build -t test-image ./out/testdata/docker-env ${{ env.LOG_ARGS }}
```
nirs, line 141: All the changes here belongs to previous commit, please squash.

.github/workflows/smoke-test.yml
```diff
@@ -126,13 +126,24 @@ jobs:
               --driver ${{ matrix.driver }} \
               ${{ matrix.network_flag }} \
               ${{ env.LOG_ARGS }}
+      - name: Copy testdata for image build
+        run: |
+          echo ">>> Copying testdata to ./out"
```
nirs: This log does not add any value, and it hides the actual command if the step command is folded. Please remove it.

nirs (commented):
This tests do not have any validations, so they will pass even if the command is broken, but they are very simple.

Another issue is that we cannot run these smoke test locally. We have image tests in the functional test suite that we could run there but this is much more work, so adding these trivial tests seems like a good idea if they are really trivial.

.github/workflows/smoke-test.yml
henry3260: Can we just use the Dockerfile located in `test/integration/testdata/docker-env`?

.github/workflows/smoke-test.yml
```diff
@@ -126,6 +126,24 @@ jobs:
...
+          cp -r test/integration/testdata ./out
+      - name: minikube image load
+        run: |
+          ./out/minikube image load docker.io/library/alpine ${{ env.LOG_ARGS }}
```
medyagh: lets use an image in github or anywhere else to avoid rate limit

medyagh: you can use the gcr image for pause

henry3260: > you can use the gcr image for pause

Can I continue using Alpine in the `image build`?

PR #21773, opened 2025-10-19
Remove old obsolete Podman kubic repository

afbjorklund:
The podman installation is now coming from Debian instead, and the cri-o package also includes conmon and crun now.

https://podman.io/blogs/2021/03/02/podman-support-for-older-distros

### DIFF
--- a/hack/kicbase_version/os-package-list.txt
+++ b/hack/kicbase_version/os-package-list.txt
@@ -20,8 +20,7 @@ ii  containerd.io                   1.7.28-1~debian.12~bookworm    amd64        
 ii  containernetworking-plugins     1.1.1+ds1-3+b5                 amd64        standard networking plugins - binaries
 ii  coreutils                       9.1-1                          amd64        GNU core utilities
 ii  cri-o                           1.34.1-1.1                     amd64        Open Container Initiative-based implementation of Kubernetes Container Runtime Interface
-ii  criu                            4.1.1-3                        amd64        checkpoint and restore in userspace
-ii  crun                            100:1.2-2                      amd64        OCI runtime written in C
+ii  crun                            1.8.1-1+deb12u1                amd64        lightweight OCI runtime for running containers
 ii  curl                            7.88.1-10+deb12u14             amd64        command line tool for transferring data with URL syntax
 ii  dash                            0.5.12-2                       amd64        POSIX-compliant shell
 ii  dbus                            1.14.10-1~deb12u1              amd64        simple interprocess messaging system (system message bus)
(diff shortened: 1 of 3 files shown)


afbjorklund:
ok-to-build-image

afbjorklund:
ok-to-build-image

nirs:
Nice to see the installation simplified.

/lgtm

afbjorklund:
> Nice to see the installation simplified

The main downside is the older version in system 

But in this case (podman), the upstream does not provide any ~~binaries~~ packages anyway (unlike docker). So the only alternative is building from source, or finding a third party package repository that is properly maintained...

nirs:
> But in this case (podman), the upstream does not provide any binaries anyway (unlike docker).

We have these upstream releases:
- https://github.com/openSUSE/catatonit/releases
- https://github.com/containers/crun/releases
- https://github.com/containers/podman/releases

afbjorklund:
Note that Podman is source-only, ever since the Nix static package was removed...

Only the client is provided, not the server

nirs:
@afbjorklund The tests looks good (the normal failures with crio). Does this change affect podman? Since we don't test podman in the CI, can you share integration tests results with podman before and after this change?

nirs:
> Note that Podman is source-only, ever since the Nix static package was removed...
> 
> Only the client is provided, not the server

What do you mean by server?

The [podman-remote-static tarball](https://github.com/containers/podman/releases/download/v5.6.2/podman-remote-static-linux_amd64.tar.gz) does not include all the components installed by the podman rpm and deb packages?

afbjorklund:
> The [podman-remote-static tarball](https://github.com/containers/podman/releases/download/v5.6.2/podman-remote-static-linux_amd64.tar.gz) does not include all the components installed by the podman rpm and deb packages?

It includes all the components installed by the `podman-remote` package, and same as the Mac and Win client...

On the other platforms, the `podman-remote` program is renamed to just `podman` since there is no daemon there.

By server I mean daemon or "podman.service", like `dockerd` (it is started with `podman system service`)

In order to have a functional setup, you would need to use `podman machine` (or get a "podman-full" tarball)

----

Server, as in:

```
$ podman --remote version
Client:       Podman Engine
Version:      4.9.3
API Version:  4.9.3
...

Server:       Podman Engine
Version:      4.9.3
API Version:  4.9.3
...
```

When run locally, there is no daemon but it reads the storage directly. But the podman-remote binary cannot.

afbjorklund:
> Does this change affect podman?

I think the only actual change, besides removing an old repository, is that the `crun` version was upgraded.

The old crun package had an "epoch", that was shadowing the newer version from the debian repository.

```diff
-ii  crun                            100:1.2-2                      amd64        OCI runtime written in C
+ii  crun                            1.8.1-1+deb12u1                amd64        lightweight OCI runtime for running containers
```

From 1.2 to 1.8.1

Note that the new cri-o package now bundles everything, under libexec. So it doesn't use crun or conmon (deb).

```
/usr/libexec/crio/conmon
/usr/libexec/crio/conmonrs
/usr/libexec/crio/crun
/usr/libexec/crio/runc
```

PR #21847, opened 2025-11-04
Addons: helm based addons

volatilemolotov:
This PR adds support for addons based on helm.

Adds the following:
 - helmChart struct to addon. If this is defined code will go to helmCommand . If set to nil regular kubectl addon install happens
 - helmInstall is triggered which installs the helm binary based on arch. This was done to not add bloat to the binary/iso
 - uses values and value files to set --set and --values respectively. This way one can ship a value file to configure via embed.FS
 
 
 Example impelentation:
 
 In addons list in pkg/minikube/assets/addons.go
 ```
 		"testaddon": NewAddon([]*BinAsset{
		MustBinAsset(addons.TestAddonAssets, "test-addon/test-addon.yaml", vmpath.GuestAddonsDir, "test-addon-values.yaml", "0640"),
		}, false, "testaddon", "3rd party (akvelon.com)", "volatilemolotov", "", nil, nil,
		&HelmChart{
			Name:      "testaddon",
			Repo:      "oci://linktooci/chart",
			Namespace: "testaddon",
			Values:    []string{"service.type=NodePort"},
			ValueFiles: []string{"/etc/kubernetes/addons/test-addon-values.yaml",},
		},
```
Provides foundation for #21257

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -450,20 +450,18 @@ func enableOrDisableAddonInternal(cc *config.ClusterConfig, addon *assets.Addon,
 		}
 	}
 
-if addon.HelmChart != nil {
-	err := helmInstallBinary(addon,runner)
-	if err != nil {
+	if addon.HelmChart != nil {
+		err := helmInstallBinary(addon, runner)
+		if err != nil {
+			return err
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+		defer cancel()
+		cmd := helmUninstallOrInstall(ctx, addon.HelmChart, enable)
+		_, err = runner.RunCmd(cmd)
 		return err
 	}
-	
-	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
-	defer cancel()
-	cmd := helmUninstallOrInstall(ctx, addon.HelmChart, enable)
-	_, err = runner.RunCmd(cmd)
-	return err
-		}
-  
-
 
 	// on the first attempt try without force, but on subsequent attempts use force
 	force := false
--- a/pkg/addons/helm.go
+++ b/pkg/addons/helm.go
@@ -29,41 +29,41 @@ import (
 )
 
 // runs a helm install within the minikube vm or container based on the contents of chart *assets.HelmChart
-func installHelmChart (ctx context.Context, chart *assets.HelmChart) *exec.Cmd {
+func installHelmChart(ctx context.Context, chart *assets.HelmChart) *exec.Cmd {
 	args := []string{
 		fmt.Sprintf("KUBECONFIG=%s", path.Join(vmpath.GuestPersistentDir, "kubeconfig")),
 		"helm", "upgrade", "--install", chart.Name, chart.Repo, "--create-namespace",
 	}
-			if chart.Namespace != "" {
-			args = append(args, "--namespace", chart.Namespace)
-		}
-
-			if chart.Values != nil {
-				for _, value := range chart.Values {
-					args = append(args, "--set", value)
-					}
-				}
-
-			if chart.ValueFiles != nil {
-				for _, value := range chart.ValueFiles {
-					args = append(args, "--values", value)
-					}
-				}
-				
-		return exec.CommandContext(ctx, "sudo", args...)
+	if chart.Namespace != "" {
+		args = append(args, "--namespace", chart.Namespace)
 	}
 
+	if chart.Values != nil {
+		for _, value := range chart.Values {
+			args = append(args, "--set", value)
+		}
+	}
+
+	if chart.ValueFiles != nil {
+		for _, value := range chart.ValueFiles {
+			args = append(args, "--values", value)
+		}
+	}
+
+	return exec.CommandContext(ctx, "sudo", args...)
+}
+
 // runs a helm uninstall based on the contents of chart *assets.HelmChart
-func uninstalllHelmChart (ctx context.Context, chart *assets.HelmChart) *exec.Cmd {
+func uninstalllHelmChart(ctx context.Context, chart *assets.HelmChart) *exec.Cmd {
 	args := []string{
 		fmt.Sprintf("KUBECONFIG=%s", path.Join(vmpath.GuestPersistentDir, "kubeconfig")),
 		"helm", "uninstall", chart.Name,
-		}
-			if chart.Namespace != "" {
-			args = append(args, "--namespace", chart.Namespace)
-		}
-		return exec.CommandContext(ctx, "sudo", args...)
 	}
+	if chart.Namespace != "" {
+		args = append(args, "--namespace", chart.Namespace)
+	}
+	return exec.CommandContext(ctx, "sudo", args...)
+}
 
 // based on enable will execute installHelmChart or uninstallHelmChart
 func helmUninstallOrInstall(ctx context.Context, chart *assets.HelmChart, enable bool) *exec.Cmd {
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: pkg/addons/helm.go [hunk 1]
```diff
@@ -0,0 +1,90 @@
...
+	"path"
+	"strings"
+
+	"github.com/pkg/errors"
+	"k8s.io/minikube/pkg/minikube/assets"
+	"k8s.io/minikube/pkg/minikube/command"
+	"k8s.io/minikube/pkg/minikube/vmpath"
+)
+
+func helmCommand(ctx context.Context, chart *assets.HelmChart, enable bool) *exec.Cmd {
```
medyagh: add a comment explain in short what this func does?

pkg/addons/helm.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	if chart.ValueFiles != nil {
+		for _, value := range chart.ValueFiles {
+			args = append(args, "--values", value)
+		}
+	}
+	return exec.CommandContext(ctx, "sudo", args...)
+}
+
+func helmInstall(addon *assets.Addon, runner command.Runner) error {
```
medyagh: better name would be installHelmBinary, so we know we are talking about

pkg/addons/helm.go
```diff
@@ -0,0 +1,90 @@
...
+	var helmArch string
+	switch arch {
+		case "x86_64":
+			helmArch = "amd64"
+		case "aarch64", "arm64":
+			helmArch = "arm64"
+		default:
+			return fmt.Errorf("failure to detect architecture or unsupported architecture: %s", arch)
+	}
+	helmURL := fmt.Sprintf("https://get.helm.sh/helm-v3.19.0-linux-%s.tar.gz", helmArch)
```
medyagh: is there a "latest" helm url that we could use ?

pkg/addons/helm.go
```diff
@@ -0,0 +1,90 @@
...
+			args = append(args, "--values", value)
+		}
+	}
+	return exec.CommandContext(ctx, "sudo", args...)
+}
+
+func helmInstall(addon *assets.Addon, runner command.Runner) error {
+	_, err := runner.RunCmd(exec.Command("test", "-f", "/usr/bin/helm"))
+	if err != nil {
+		// If not, install it
```
medyagh: how about installing it through their Script ?
https://helm.sh/docs/intro/install#from-script

medyagh (changes requested):

pkg/addons/helm.go
medyagh: this will install latest by script and no need for us to parse archs

pkg/addons/helm.go
volatilemolotov: Not aware of any available latest target.

pkg/addons/helm.go
volatilemolotov: Did not want to shoehorn a whole script into this but we could do that. It would require a chmod call and similar but doable. Ill see if its works

volatilemolotov: Actually not possible without changing either ISO or the script. Script will try to put helm into `/usr/local/bin` which the VM does not have. Issue encountered on macos with `krunkit` driver. Same happens on linux with `kvm` driver

Output from `minikube ssh` and `$ ls /usr/ ` 
```
bin  lib  lib64  libexec  sbin  share
```

pkg/addons/helm.go
volatilemolotov: Yeah, short description added

pkg/addons/helm.go
volatilemolotov: Agree, changed

pkg/addons/addons.go
```diff
@@ -450,14 +450,31 @@ func enableOrDisableAddonInternal(cc *config.ClusterConfig, addon *assets.Addon,
...
 
+if addon.HelmChart != nil {
+			err := helmInstallBinary(addon,runner)
+			if err != nil {
+				return err
+			}
+			
+			ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+			defer cancel()
+			cmd := helmCommand(ctx, addon.HelmChart, enable)
```
medyagh: would this ever be called with disable?

pkg/addons/helm.go
medyagh: how about make this a Generic func, and have two other helpers
called
insatallHelmChart
uninstallHelmChart

so we dont have to pass the enable to it

pkg/addons/helm.go
volatilemolotov: Ill revisit this one. Probably we can make the dir and copy the binary where it needs to be and satisfy both environments

pkg/addons/helm.go
volatilemolotov: Added installHelmChart and uninstallHelmChart along with helmUninstallOrInstall

pkg/addons/helm.go
volatilemolotov: changed this to get_helm.sh

pkg/addons/helm.go
volatilemolotov: Decided to go with checking if the folder exists and create it if it does not. Then copy helm binary to the PATH.

pkg/addons/addons.go
volatilemolotov: Yes

pkg/addons/addons.go
```diff
@@ -450,14 +450,31 @@ func enableOrDisableAddonInternal(cc *config.ClusterConfig, addon *assets.Addon,
 		}
```
medyagh, line 450: we can add a documentation to the website https://minikube.sigs.k8s.io/docs/contrib/addons/

we can add a block (how to add helm based addons)

volatilemolotov: Pushed a new site doc with some explanation. As we work on adding a actual addon we will update the doc for the helm based addon

pkg/addons/helm.go
```diff
@@ -0,0 +1,89 @@
...
+	"os/exec"
+	"path"
+
+	"github.com/pkg/errors"
+	"k8s.io/minikube/pkg/minikube/assets"
+	"k8s.io/minikube/pkg/minikube/command"
+	"k8s.io/minikube/pkg/minikube/vmpath"
+)
+
+// runs a helm install based on the contents of chart *assets.HelmChart
```
medyagh: clarify in the comment that this runs inside minikube vm/container

pkg/addons/helm.go
```diff
@@ -0,0 +1,89 @@
...
+	_, err := runner.RunCmd(exec.Command("test", "-f", "/usr/bin/helm"))
+	if err != nil {
+		_, err = runner.RunCmd(exec.Command("test", "-d", "/usr/local/bin"))
+		if err != nil {
+			_, err = runner.RunCmd(exec.Command("sudo", "mkdir", "-p", "/usr/local/bin"))
+		}
+
+		installCmd := fmt.Sprint("curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 && chmod 700 get_helm.sh && ./get_helm.sh")
+		_, err = runner.RunCmd(exec.Command("sudo", "bash", "-c", installCmd))
+		_, err = runner.RunCmd(exec.Command("sudo", "mv", "/usr/local/bin/helm", "/usr/bin/helm"))
```
medyagh, line 87: add comment explain the reason we do this is /usr/loca/bin/helm is not in PATH on both iso/kicbase

site/content/en/docs/contrib/helm-addons.md
```diff
@@ -0,0 +1,87 @@
+---
+title: "Helm Based Addons"
+weight: 5
+description: >
+  How to develop minikube addons using Helm charts.
```
medyagh, line 5: add a link to this in the main Addon Doc (use relative links for docs)

pkg/addons/helm.go
volatilemolotov: Added

site/content/en/docs/contrib/helm-addons.md
volatilemolotov: Added

pkg/addons/helm.go
volatilemolotov: added

PR #21903, opened 2025-11-16
feat: Add custom CRI-O storage paths with automatic bind mount

### LINKED ISSUE #17415: Allow to reuse podman images of the host when using podman driver
### What Happened?

Minikube allows to run a kubernetes cluster on top of a container engine like docker or podman.

We can also specify a runtime with minikube like docker, containerd or cri-o.

In case of cri-o engine, it's nice as podman is also available inside the container

image

so with `minikube podman-env` there is a way to use podman engine inside minikube container and then share images with cri-o engine as it's sharing /var/lib/containers path

Minikube explains different ways of pushing images to the cluster
https://minikube.sigs.k8s.io/docs/handbook/pushing/

I think there is room for another method. As a podman user, I would like to continue to build using the default podman running inside the podman machine and cherry on the cake, see the images I'm building in cri-o (so it's close of podman-env method but I shouldn't need to run podman inside podman, just use the podman on the podman machine/host.

so if we could bind the podman host /var/lib/containers to the minikube container and that cri-o on that container use this mount folder, then as a user, I could do

`podman build -t my-image .` and then reference in a k8s yaml this image and doing `kubectl apply -f mypod.yaml` quickly spin a pod

image

it requires to add inside crio.conf the path of the mounted folder of the podman machine host rather than defaulting to the `/var/lib`

so, to sum up:

start minikube with an extra mount path

something like

```bash
minikube start --driver=podman --container-runtime=cri-o --mount --mount-string=/var/lib/containers:/host-containers
```

and then add a way in kicbase image to add the extra crio configuration

```toml
[crio]

# Path to the "root directory". CRI-O stores all of its data, including
# containers images, in this directory.
# root = "/var/lib/containers/storage"
root = "/host-containers/storage"
```

maybe the mount folder could be the rootless host path (so instead of mounting `/var/lib/containers`, mounting `/var/home/core/.local/share/containers`)

How do you see how we could have a toggle for changing/updating crio configuration to use the underlying images folder.

on a side issue, could cri-o runtime and podman runtime could be updated as well inside the kicbase image ?

### Attach the log file

N/A

### Operating System

macOS (Default)

### Driver

Podman

vtri950:
Add support for custom CRI-O storage paths with automatic bind mounts

Implements automatic sharing of Podman host images with CRI-O in minikube
by allowing users to specify custom storage paths that are automatically
bind-mounted into the container.

This enhancement adds two new flags:
- --container-storage-root: Specifies custom CRI-O storage root path
- --container-storage-runroot: Specifies custom CRI-O runtime root path

When these flags are used with KIC drivers (Podman/Docker) and CRI-O runtime,
the specified directories are automatically:
1. Bind-mounted from host to container at the same paths
2. Configured in CRI-O's configuration file
3. Used by CRI-O for image storage

This eliminates the need for manual kicbase modifications or image loading
when using Podman as the driver with CRI-O runtime.

Fixes #17415

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -146,6 +146,8 @@ const (
 	staticIP                = "static-ip"
 	gpus                    = "gpus"
 	autoPauseInterval       = "auto-pause-interval"
+	containerStorageRoot    = "container-storage-root"
+	containerStorageRunRoot = "container-storage-runroot"
 )
 
 var (
@@ -211,6 +213,8 @@ func initMinikubeFlags() {
 	startCmd.Flags().String(staticIP, "", "Set a static IP for the minikube cluster, the IP must be: private, IPv4, and the last octet must be between 2 and 254, for example 192.168.200.200 (Docker and Podman drivers only)")
 	startCmd.Flags().StringP(gpus, "g", "", "Allow pods to use your GPUs. Options include: [all,nvidia,amd] (Docker driver with Docker container-runtime only)")
 	startCmd.Flags().Duration(autoPauseInterval, time.Minute*1, "Duration of inactivity before the minikube VM is paused (default 1m0s)")
+	startCmd.Flags().String(containerStorageRoot, "", "Absolute path for container runtime storage root directory (e.g. /host-containers/storage for CRI-O). Works with CRI-O and other compatible runtimes.")
+	startCmd.Flags().String(containerStorageRunRoot, "", "Absolute path for container runtime storage state directory (e.g. /host-containers/storage for CRI-O). Works with CRI-O and other compatible runtimes.")
 }
 
 // initKubernetesFlags inits the commandline flags for Kubernetes related options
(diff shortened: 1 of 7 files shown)


medyagh (commented):
@vtri950 
to help me undrestand this, can you provide an example of running this new feature ? and how does this help crio users?

vtri950:
@medyagh absolutely!

it starts with this https://github.com/podman-desktop/podman-desktop/issues/2866
the current setup of podman with kind  has a manual step of loading images in kind container after podman builds them
we wanted to automate that process

so  @benoitf and @afbjorklund came up with soln using minikube to do that
https://gist.github.com/benoitf/3e45effb48e27791282eb227410f5950

but this required to run the cluster on kicbase image and to avoid that they openend this issue 
https://github.com/kubernetes/minikube/issues/17415

and my pr fixes that

i have tested it on my wsl(ubuntu) for podman with rootless and in windows with a podman vm(rootful) and it works as expected 

you can build your images with podman and directly run your container inside minikubbe cluster without having that extra manual load step

afbjorklund:
> can you provide an example of running this new feature ? 

There was a blog post about it at the time, but just like the minikube preloads it increases the coupling...
i.e. it assumes that you can mount the internals of the images from another incarnation of the runtime

https://podman-desktop.io/blog/sharing-podman-images-with-kubernetes-cluster

I know it speeds things up, especially with large images, but using the cache or save/load is much easier.
Currently the preferred approach is to access the cluster with minikube docker-env and build from there.

The flags look harmless enough though, since by default they don't do anything.

afbjorklund:
I don't remember why it was better to configure the runtime, versus just mounting a docker volume over it.

```
# CRI-O reads its storage defaults from the containers-storage.conf(5) file
# located at /etc/containers/storage.conf. Modify this storage configuration if
# you want to change the system's defaults. If you want to modify storage just
# for CRI-O, you can change the storage configuration options here.
[crio]

# Path to the "root directory". CRI-O stores all of its data, including
# containers images, in this directory.
# root = "/var/lib/containers/storage"

# Path to the "run directory". CRI-O stores all of its state in this directory.
# runroot = "/run/containers/storage"

# Path to the "imagestore". If CRI-O stores all of its images in this directory differently than Root.
# imagestore = ""
```

Maybe there were some clues in https://github.com/kubernetes/minikube/issues/17415

PR #21904, opened 2025-11-16
[ci]: migrate  kubernetes bootcamp image to prow

ComradeProgrammer:
[ci]: migrate  kubernetes bootcamp image to prow

### DIFF
--- a/hack/prow/prow.mk
+++ b/hack/prow/prow.mk
@@ -22,3 +22,11 @@ setup-prow-gcp-ssh-keys: # set up ssh keys for gcloud cli. These env vars are se
 	mkdir -p -m 0700 ~/.ssh
 	cp -f "${GCE_SSH_PRIVATE_KEY_FILE}" ~/.ssh/google_compute_engine
 	cp -f "${GCE_SSH_PUBLIC_KEY_FILE}" ~/.ssh/google_compute_engine.pub
+	
+.PHONY: push-kubernetes-bootcamp
+push-kubernetes-bootcamp:
+	docker run --rm --privileged tonistiigi/binfmt:latest --install all
+	docker buildx create --name multiarch --bootstrap
+	docker buildx build --builder multiarch --push --platform  linux/amd64,linux/arm64 \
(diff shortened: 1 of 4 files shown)


hack/prow/prow.mk
```diff
@@ -1,5 +1,4 @@
-.PHONY: integration-prow-kvm-docker-linux-x86-64
-
+.PHONY: integration-prow-kvm-docker-linux-x86-64 integration-prow-docker-docker-linux-x86-64  push-kubernetes-bootcamp
```
medyagh: how about add this new phony right before the target to match other ones.  Instead of putting all .phonies in one place

hack/prow/prow.mk
```diff
@@ -16,4 +15,11 @@ integration-prow-kvm-docker-linux-x86-64:
 	cp -f "${GCE_SSH_PRIVATE_KEY_FILE}" ~/.ssh/google_compute_engine
 	cp -f "${GCE_SSH_PUBLIC_KEY_FILE}" ~/.ssh/google_compute_engine.pub
 	GOTOOLCHAIN=auto go build -C ./hack/prow/minitest -o $(PWD)/out/minitest .
-	./out/minitest  --deployer boskos --tester kvm-docker-linux-amd64-integration --config hack/prow/kvm.json
\ No newline at end of file
+	./out/minitest  --deployer boskos --tester kvm-docker-linux-amd64-integration --config hack/prow/kvm.json
+
+push-kubernetes-bootcamp:
```
medyagh, line 27: how about naming it prow so all prow things have a prow in their name


push-prow-kubernetes-bootcamp:

hack/prow/prow.mk
```diff
@@ -16,4 +15,11 @@ integration-prow-kvm-docker-linux-x86-64:
...
+	docker run --rm --privileged tonistiigi/binfmt:latest --install all
+	docker buildx create --name multiarch --bootstrap
+	docker buildx build --builder multiarch --push --platform  linux/amd64,linux/arm64 \
+		-t gcr.io/minikube/kubernetes-bootcamp:$(_GIT_TAG) -t gcr.io/minikube/kubernetes-bootcamp:latest deploy/image/kubernetes-bootcamp
```
medyagh: do we have access to gcr.io/minikube in prow ?

ComradeProgrammer: no, I just updated the right registry name

hack/prow/prow.mk
ComradeProgrammer: we can't change it here because it is referenced in test-infra

PR #21934, opened 2025-11-19
build: bump golang manually to 1.25.3

medyagh:
bumps golang to 1.25
and also removes hyperkit from cross build till we decide if to remove hyperkit fully or continue allowing it https://github.com/kubernetes/minikube/issues/21940


medyagh:
OK to test

PR #21964, opened 2025-11-22
WIP: test merging functional_verified

divysinghvi:
test pr 
issue : #21915

### DIFF
--- a/.github/workflows/functional_test.yml
+++ b/.github/workflows/functional_test.yml
@@ -43,7 +43,7 @@ jobs:
         run: go mod download
       - name: Build minikube and e2e test binaries
         run: |
-          make e2e-linux-amd64 e2e-darwin-amd64
+          make e2e-linux-amd64 e2e-darwin-amd64 e2e-linux-arm64
           cp -r test/integration/testdata ./out
       - name: Upload Test Binaries
         uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4
@@ -53,7 +53,7 @@ jobs:
   functional-test:
     name: ${{ matrix.name }}
     needs: build-test-binaries
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.runner }}
     permissions:
       contents: none
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/functional_test.yml
```diff
@@ -64,42 +64,55 @@ jobs:
...
             cruntime: docker
             os: macos-13
+            runner: macos-13
             extra-start-args: --network vmnet-shared
             test-timeout: 50m            
+          - name: docker-docker-ubuntu-arm64
+            driver: docker
+            cruntime: docker
+            os: ubuntu-22.04
+            runner: [self-hosted, arm64]
```
medyagh, line 110: now that we are adding "Runner" to the matrix, maybe we can remove "os" if it is not used anywhere ? or make them use runner.

divysinghvi: yes we can remove the os

divysinghvi:
@medyagh right now didn't add the check for the ok-to-test also since i could not trigger that this was more of the test pr to check if it works , actually one thing that i was getting confused with is why here i am not seeing the functional test results am i missing something ?

.github/workflows/functional_test.yml
medyagh, line 110: we would need to see if we can make that Only one of the Matrix items be required to have ok-to-test label and the test be  as it is, can you see if thats possible?


feel free to make a Dummy Repo just to check that in a Dummy repo with a dummy workflow on your own repo

divysinghvi: Medhya i looked more into it thw core issue having is that GitHub Actions doesn't support conditional logic at the matrix definition level so we can't add a check in the include section , if that would have been possible it would be easier but it's not but we can add a check in the steps portion 
`      - name: Check if arm64 job should run
        if: matrix.name == 'docker-docker-ubuntu-arm64' && github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'ok-to-test')
        run: |
          echo "Skipping arm64 job - 'ok-to-test' label not found on PR"
          exit 1`
also please ignore the commits i am experimenting for the main pr i will close this on and create a fresh one once i figure this thing out thanks

PR #21971, opened 2025-11-22
ci: fix upater get version

### LINKED ISSUE #21970: Update Automation PR:The "after" version is wrong in PR title
This Automation PR 
https://github.com/kubernetes/minikube/pull/21950/files

https://github.com/kubernetes/minikube/tree/master/.github/workflows/update-go-github-version.yml

The PR Title is:
CI: update go-github from v74.0.0 to v74.0.0

While the PR itself is changing it to v79.0,0

anyone taking this issue I suggest looking at the yaml file above and also the hack/update folder

I noticed a couple other PRs titles were wrong and I fixed manually

Caesarsage:
Fixes #21970

The get_version.go tool was using FindSubmatch which returns only the
first match. When go.mod contains multiple versions of the same
dependency (e.g., v74 and v79 for go-github), it would always return
the first (older) version.

This caused automated update PRs to show incorrect titles like:
'CI: update go-github from v74.0.0 to v74.0.0'

### DIFF
--- a/hack/update/get_version/get_version.go
+++ b/hack/update/get_version/get_version.go
@@ -135,7 +135,15 @@ func main() {
 	if err != nil {
 		log.Fatalf("failed to read file: %v", err)
 	}
-	submatches := re.FindSubmatch(data)
+
+  // this handles cases where multiple versions exist (e.g., old and new versions in go.mod)
+	allMatches := re.FindAllSubmatch(data, -1)
+	if len(allMatches) == 0 {
+		log.Fatalf("no matches found")
+	}
+
+	// Take the last match (most recent version)
+	submatches := allMatches[len(allMatches)-1]
 	if len(submatches) < 2 {
 		log.Fatalf("less than 2 submatches found")
 	}


Caesarsage:
> @Caesarsage can you plz post an example of running the code with getting version of something with multiple

Hello @medyagh 

Here's the exampl showing the fix working with multiple versions:

### Test Scenario
Using the current `go.mod` with multiple go-github entries:
```
github.com/google/go-github/v74 v74.0.0
github.com/google/go-github/v79 v79.0.0  // new as a result of `make update-go-github-version`:
```

### Before Fix (using `FindSubmatch`)

The original code uses `FindSubmatch()` which returns only the **first match** (even after the automation runs `make update-go-github-version`):
```bash
$ DEP=go-github make get-dependency-version
v74.0.0  # Returns the FIRST/OLD version
```

### After Fix (using `FindAllSubmatch`)

The fixed code uses `FindAllSubmatch()` and takes the **last match**, which is what is needed after the automation runs `make update-go-github-version`:
```bash
$ DEP=go-github make get-dependency-version
v79.0.0  # Returns the LAST/CURRENT version
```

When the automated workflow runs for futture reference on new versions:
1. **OLD_VERSION** captures the current version → `v79.0.0` 
2. `make update-go-github-version` runs and updates to new version → `v80.0.0`
3. **NEW_VERSION** captures the updated version → `v80.0.0`

PR #21972, opened 2025-11-22
ci: add none driver tests to prow

medyagh:
- renamed kvm.json to boskos-nested.json since it is being used by other tests to created nested vms
- added name make targets in prow makefile to be re-used 
- bumped legacy minikube version to upgrade from

### DIFF
--- a/test/integration/preload_test.go
+++ b/test/integration/preload_test.go
@@ -57,8 +57,8 @@ func TestPreload(t *testing.T) {
 		t.Fatalf("%s failed: %v", rr.Command(), err)
 	}
 
-	// re-start the cluster and check if image is preserved
-	startArgs = []string{"start", "-p", profile, "--memory=3072", "--alsologtostderr", "-v=1", "--wait=true"}
+	// re-start the cluster and check if image is preserved with enabled preload
+	startArgs = []string{"start", "-p", profile, "--preload=true", "--alsologtostderr", "-v=1", "--wait=true"}
 	startArgs = append(startArgs, StartArgs()...)
 	rr, err = Run(t, exec.CommandContext(ctx, Target(), startArgs...))
 	if err != nil {


hack/prow/minitest/tester/none_docker_linux_amd64_integration.go
```diff
@@ -0,0 +1,60 @@
...
+package tester
+
+import (
+	"fmt"
+	"os"
+
+	"k8s.io/klog/v2"
+)
+
+var _ MiniTestTester = &KVMDockerLinuxAmd64IntegrationTester{}
```
divysinghvi, line 26: Should this be var _ MiniTestTester = &NoneDockerLinuxAmd64IntegrationTester{} ?

hack/prow/minitest/tester/none_docker_linux_amd64_integration.go
```diff
@@ -0,0 +1,60 @@
...
+		klog.Errorf("tester: deployed environment is not up: %v", err)
+	}
+	if err := runner.SyncToRemote(".", "~/minikube", []string{".cache"}); err != nil {
+		klog.Errorf("failed to sync file in docker deployer: %v", err)
+	}
+	pr := os.Getenv("PULL_NUMBER")
+
+	var testErr error
+	// install docker and libvirtd first then run the test in a new shell
+	if err := runner.Execute("cd minikube && ./hack/prow/integration_kvm_docker_linux_x86-64_pre.sh"); err != nil {
```
divysinghvi, line 45: is there any specific reason ./hack/prow/integration_kvm_docker_linux_x86-64_pre.sh is referenced here

medyagh: yes because after docker is installed we need to Exit the shell (and restart the shell) so we run that in a separate script

hack/prow/minitest/tester/none_docker_linux_amd64_integration.go
medyagh: I think this is just setting the default (it gets overwritten after I think)

divysinghvi: oh understood

hack/prow/minitest/tester/none_docker_linux_amd64_integration.go
divysinghvi: okay got it

Issue #21916, opened 2025-11-17
add more Prow Full Tests


medyagh:
Jimaing added KVM Docker Linux

Example with Needing a VM
https://github.com/kubernetes/minikube/pull/21556

Example not needing GCP VM and just run in a Coaniner (example Docker Driver)
https://github.com/kubernetes/minikube/pull/21807

we need to add the rest of them which are not implemented 

## Remaining ones:

[integration-docker-containerd-linux-x86-64](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/21913/integration-docker-containerd-linux-x86-64/1990495555268120576)
Required
[integration-docker-crio-linux-x86-64](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/21913/integration-docker-crio-linux-x86-64/1990495555356200960)
[integration-kvm-containerd-linux-x86-64](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/21913/integration-kvm-containerd-linux-x86-64/1990495554840301568)
[integration-kvm-crio-linux-x86-64](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/21913/integration-kvm-crio-linux-x86-64/1990495555125514240)
[integration-none-docker-linux-x86-64](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/21913/integration-none-docker-linux-x86-64/1990495555456864256)
integration-none-docker-linux-x86-64 — Job failed.

divysinghvi:
[integration-docker-containerd-linux-x86-64](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/21913/integration-docker-containerd-linux-x86-64/1990495555268120576) i will be working on this prow test

PR #21974, opened 2025-11-23
ci: Added docker-containerd-linux-x86-64 tests to prow

divysinghvi:
Added integration-prow-docker-containerd-linux-x86-64 prow test 
issue : #21916

### DIFF
--- a/deploy/kicbase/Dockerfile
+++ b/deploy/kicbase/Dockerfile
@@ -19,7 +19,7 @@
 
 
 # this ARG needs to be global to use it in `FROM` & is updated for new versions of debian:bookworm-slim-*
-ARG KICBASE_IMAGE="debian:bookworm-20251020-slim"
+ARG KICBASE_IMAGE="debian:bookworm-20251117-slim"
 # multi-stage docker build so we can build auto-pause for arm64
 FROM golang:1.25.3 AS auto-pause
 WORKDIR /src
--- a/deploy/image/kubernetes-bootcamp/Dockerfile
+++ b/deploy/image/kubernetes-bootcamp/Dockerfile
@@ -1,0 +1,5 @@
+FROM node:slim
+RUN apt-get update && apt-get install curl -y
+EXPOSE 8080
+COPY server.js .
+CMD [ "node", "server.js" ]
--- a/deploy/image/kubernetes-bootcamp/server.js
+++ b/deploy/image/kubernetes-bootcamp/server.js
@@ -1,0 +1,19 @@
+var http = require('http');
+var requests = 0;
+var podname = process.env.HOSTNAME;
+var startTime;
+var host;
+var handleRequest = function(request, response) {
+  response.setHeader('Content-Type', 'text/plain');
+  response.writeHead(200);
+  response.write('Hello Kubernetes bootcamp! | Hostname: ');
+  response.write(host);
+  response.end(' | v=2\n');
+  console.log('Hostname:', host, '| Total Requests:', ++requests, '| App Uptime:', (new Date() - startTime)/1000 , 'seconds', '| Log Time:', new Date());
+}
+var www = http.createServer(handleRequest);
+www.listen(8080, function () {
+    startTime = new Date();
+    host = process.env.HOSTNAME;
+    console.log('Kubernetes Bootcamp App Started At:', startTime, '| Running On:', host, '\n');
+});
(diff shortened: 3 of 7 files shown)


test/integration/start_stop_delete_test.go
```diff
@@ -339,9 +339,15 @@ func testPodScheduling(ctx context.Context, t *testing.T, profile string) {
...
+	// Default for docker runtime: 1048576
+	// Containerd with docker driver: 1024
+	// Containerd with KVM driver: 1048576
 	expected := int64(1048576)
+	if ContainerRuntime() == "containerd" && DockerDriver() {
+		expected = int64(1024)
+	}
 	if got != expected {
-		t.Errorf("'ulimit -n' returned %d, expected %d", got, expected)
+		t.Errorf("'ulimit -n' returned %d, expected %d for %s runtime with driver", got, expected, ContainerRuntime())
```
medyagh: is this just to debug ? doesnt seem related to this PR

hack/prow/integration_docker_containerd_linux_x86-64.sh
```diff
@@ -0,0 +1,37 @@
...
+set -e
+set -x
+
+OS="linux"
+ARCH="amd64"
+DRIVER="docker"
+CONTAINER_RUNTIME="containerd"
+EXTRA_START_ARGS="" 
+EXTRA_TEST_ARGS=""
+JOB_NAME="Docker_Containerd_Linux"
```
medyagh: add _x86-64 suffix here

divysinghvi: okay

test/integration/start_stop_delete_test.go
divysinghvi: sorry i was trying to debug something i forgot to remove it is not related to this pr i will remove

ComradeProgrammer:
Great Job!
But this ulimit change sounds a bit weird becaue I think the value 1048576 comes from minikube (pkg/provision) with something like `--default-ulimit=nofile=1048576` so it is pretty weird why it is 1024 here 

@medyagh Do you have any idea? or perhaps we should investigate why it is 1024...

test/integration/start_stop_delete_test.go
```diff
@@ -339,7 +339,13 @@ func testPodScheduling(ctx context.Context, t *testing.T, profile string) {
 	}
 
 	// Arbitrary value set by some container runtimes. If higher, apps like MySQL may make bad decisions.
+	// Default for docker runtime: 1048576
+	// Containerd with docker driver: 1024
+	// Containerd with KVM driver: 1048576
 	expected := int64(1048576)
+	if ContainerRuntime() == "containerd" && DockerDriver() {
```
medyagh: would this fail in Jenkins ? if this is only prow-sepecifc we might wanna add an evironment detection to detect package and if detect.IsProw() is true expect different ulimit

divysinghvi: @medyagh @ComradeProgrammer  I might have found the root cause. The issue is that containerd doesn't have ulimit configuration in its config.toml.

When using docker driver:
- docker runtime: inherits dockerd's --default-ulimit=nofile=1048576 → gets 1048576 
- containerd runtime: no ulimit config, falls back to system default for minikube user → gets 1024 
could this be causing this issue ??

medyagh: yes that could be the issue

pkg/minikube/cruntime/containerd.go
```diff
@@ -224,6 +224,14 @@ func generateContainerdConfig(cr CommandRunner, imageRepository string, kv semve
 	return nil
 }
 
+// setContainerdUlimit sets LimitNOFILE for containerd to match docker's default ulimit
+func setContainerdUlimit(cr CommandRunner) error {
```
medyagh: if we are adding this to the kicbase this code wont be needed.

deploy/kicbase/Dockerfile
```diff
@@ -172,6 +172,10 @@ RUN install -m 0755 -d /etc/apt/keyrings \
     tee /etc/apt/sources.list.d/docker.list > /dev/null \
     && clean-install docker-ce docker-ce-cli containerd.io docker-buildx-plugin
 
+# Set ulimit for containerd to match docker's default-ulimit setting
+# This ensures consistent file descriptor limits (1048576) across container runtimes
+RUN sed -i '/^\[Service\]/a LimitNOFILE=1048576' /lib/systemd/system/containerd.service
```
medyagh, line 177: if you want this change plz ask one of the maintainers to run
ok-to-build-image

pkg/minikube/cruntime/containerd.go
divysinghvi: yes we won't need if we are updating the kicbase image i did not knew how the kickbase image update worked

deploy/kicbase/Dockerfile
divysinghvi: can you run ok-to-build-image sorry i am still confused about the process

medyagh:
ok-to-build-image

PR #21932, opened 2025-11-19
wip: Fix image loading with podman driver 

obnoxxx:
This fixes the image loading for the podman driver removing the annoying message

Resolves: #8426 

## without this patch

```console
$ git branch --show-current
master
$ git fetch origin
...
$  git reset --hard origin/master
HEAD is now at bac7021992 Merge pull request #21933 from medyagh/fix_update_golang
$ make clean && make
$ ./out/minikube  profile list
┌───────────────────┬────────┬─────────┬──────────────┬─────────┬─────────┬───────┬────────────────┬────────────────────┐
│      PROFILE      │ DRIVER │ RUNTIME │      IP      │ VERSION │ STATUS  │ NODES │ ACTIVE PROFILE │ ACTIVE KUBECONTEXT │
├───────────────────┼────────┼─────────┼──────────────┼─────────┼─────────┼───────┼────────────────┼────────────────────┤
│ functional-865055 │ docker │ crio    │ 192.168.49.2 │ v1.34.1 │ Stopped │ 1     │                │                    │
│ functional-981737 │ docker │ crio    │ 192.168.58.2 │ v1.34.1 │ Stopped │ 1     │                │                    │
│ minikube          │ kvm2   │ docker  │              │ v1.33.0 │ Stopped │ 1     │ *              │                    │
└───────────────────┴────────┴─────────┴──────────────┴─────────┴─────────┴───────┴────────────────┴────────────────────┘
11:21:42 obnox@cone minikube ±|master ✗|→ ./out/minikube delete --all
🔥  Deleting "functional-865055" in docker ...
🔥  Removing /home/obnox/.minikube/machines/functional-865055 ...
💀  Removed all traces of the "functional-865055" cluster.
🔥  Deleting "functional-981737" in docker ...
🔥  Removing /home/obnox/.minikube/machines/functional-981737 ...
💀  Removed all traces of the "functional-981737" cluster.
🔥  Deleting "minikube" in kvm2 ...
💀  Removed all traces of the "minikube" cluster.
🔥  Successfully deleted all profiles
 $ time ./out/minikube start -d podman 
😄  minikube v1.37.0 on Fedora 43
✨  Using the podman driver based on user configuration
📌  Using Podman driver with root privileges
👍  Starting "minikube" primary control-plane node in "minikube" cluster
🚜  Pulling base image v0.0.48-1763507788-21924 ...
💾  Downloading Kubernetes v1.34.1 preload ...
[...]

Issue #8426, opened 2020-06-09
podman: load kic base image from cache if available for offline mode


priyawadhwa:
This should be easier to implement once https://github.com/kubernetes/minikube/pull/8417 is merged

cdbattags:
Any update on this?

crfrolik:
Why is this issue closed?

This line of code still exists: https://github.com/kubernetes/minikube/blob/master/pkg/minikube/node/cache.go#L158

It prints when using podman:

> cache.go:203] Error downloading kic artifacts:  not yet implemented, see issue #8426

iapicca:
since this issue is closed I'm opening a [new one](https://github.com/kubernetes/minikube/issues/14664)

terjr938:
Still receiving the error, even though 8417 has been merged, or so it seems.  I'm attaching the requested logs.txt, so hopefully someone could take a look and let me know if it's a bug, because it sure seems like one.
[logs.txt](https://github.com/user-attachments/files/18378701/logs.txt)

jogerj:
still not resolved but closed

https://github.com/kubernetes/minikube/blob/b973335449467dc12e23956be18336c42ff6b3ce/pkg/minikube/node/cache.go#L159-L161

afbjorklund:
It got reverted, since #11063

laundry-96:
Running into this issue now with Podman Desktop :(

@afbjorklund Sorry if you're the wrong person, but do you know if this will get re-fixed?

EDIT: Running `minikube start --driver podman` helped. I also set the default connection to the root one (not sure if that helped or not either)

obnoxxx:
can this be reopened?

PR #22024, opened 2025-12-02
UI: remove  confusing message  related to podman

obnoxxx:
Image loading from chache  is not implemented and  `minikube start -d podman`always issued a confusing and unhelpful error message. 

This change removes  the confusing and unhelpful message, replacing it with a log message. 

This is intended as a replacement for PR #21932  without pretending to implement the cache handling.

The outputs  of `minikube start -d podman` on current Fedora Linux 43 are provided with and without this patch illustrate the changed behavior:

# without the patch

```console

$ # on current master:
$ ./out/minikube delete --all
🔥  Deleting "minikube" in podman ...
🔥  Removing /home/obnox/.minikube/machines/minikube ...


$ time ./out/minikube start -d podman 
😄  minikube v1.37.0 on Fedora 43
✨  Using the podman driver based on user configuration
📌  Using Podman driver with root privileges
🚜  Pulling base image v0.0.48-1764169655-21974 ...
E1203 11:02:21.355429   86335 cache.go:238] Error downloading kic artifacts:  not yet implemented, see issue #8426
🐳  Preparing Kubernetes v1.34.2 on Docker 29.0.4 ...

real    0m29.764s
user    0m1.816s
sys     0m0.963s
$ echo $?
0
$


```
# with the patch

```console
$ # on obnoxxx/kicbase-simplify-image-load :
$ make clean && make
$ ./out/minikube  delete --all --purge
$ time ./out/minikube start -d podman 😄  minikube v1.37.0 on Fedora 43
✨  Using the podman driver based on user configuration
📌  Using Podman driver with root privileges
🚜  Pulling base image v0.0.48-1765275396-22083 ...
🐳  Preparing Kubernetes v1.34.2 on Docker 29.1.2 ...

real    0m30.899s
user    0m1.452s
sys     0m1.009s
$ echo $?
0
$ 
```

### DIFF
--- a/pkg/minikube/node/cache.go
+++ b/pkg/minikube/node/cache.go
@@ -170,7 +170,8 @@ func beginDownloadKicBaseImage(g *errgroup.Group, cc *config.ClusterConfig, down
 			}
 
 			if cc.Driver == driver.Podman {
-				return fmt.Errorf("not yet implemented, see issue #8426")
+				klog.Warningf("skipping image load from cache: not implemented for podman. see issue #8426")
+				return nil
 			}
 			if driver.IsDocker(cc.Driver) && err == nil {
 				klog.Infof("Loading %s from local cache", img)


pkg/minikube/node/cache.go
```diff
@@ -165,14 +165,14 @@ func beginDownloadKicBaseImage(g *errgroup.Group, cc *config.ClusterConfig, down
 			if err == nil {
 				klog.Infof("successfully saved %s as a tarball", img)
 			}
-			if downloadOnly && err == nil {
+			if downloadOnly {
```
nirs: If we run with --download-only for downloading required resource, I think we expect to fail if the download fails. Here we ignore the download error and return success to caller.

pkg/minikube/node/cache.go
```diff
@@ -165,14 +165,14 @@ func beginDownloadKicBaseImage(g *errgroup.Group, cc *config.ClusterConfig, down
...
 				return nil
 			}
 
 			if cc.Driver == driver.Podman {
 				return fmt.Errorf("not yet implemented, see issue #8426")
 			}
-			if driver.IsDocker(cc.Driver) && err == nil {
+			if driver.IsDocker(cc.Driver) {
```
nirs: If the download failed, why are we loading the image from cache?

pkg/minikube/node/cache.go
```diff
@@ -154,25 +154,22 @@ func beginDownloadKicBaseImage(g *errgroup.Group, cc *config.ClusterConfig, down
...
-				klog.Infof("%s exists in daemon, skipping load", img)
-				finalImg = img
-				return nil
-			}
+			if driver.IsDocker(cc.Driver) {
+				if download.ImageExistsInDaemon(img) {
+					klog.Infof("%s exists in daemon, skipping load", img)
+					finalImg = img
+					return nil
+				}
```
nirs: This is simplied since we already handled downloadOnly, but this handling seems wrong, so we need to revisit this change after error handling is fixed.

pkg/minikube/node/cache.go
```diff
@@ -154,25 +154,22 @@ func beginDownloadKicBaseImage(g *errgroup.Group, cc *config.ClusterConfig, down
...
+				if err == nil {
+					klog.Infof("successfully saved %s as a tarball", img)
+				}
+				if downloadOnly {
+					return nil
+				}
 
-			if cc.Driver == driver.Podman {
-				return fmt.Errorf("not yet implemented, see issue #8426")
-			}
```
nirs, line 174: This is the only correct change - removing the bad fmt.Errorf. But it would be more correct to convert this to debug log like:

    Skipping <operation description>: not implemented, see issue #8426

This will improve the user experience but make it easier to debug podman driver, in case skipping has negative effects. If this operation is not needed for podman removing the log is a better change, but I'm not sure what this code is doing and wha is the unimplemented operation that we skipped.

pkg/minikube/node/cache.go
obnoxxx: This is not new code. The  diff just makes it look new, but it is presented this way as the convoluted if statement above is broken apart for greater clarity in the code.

fwiw, the downloadOnly case was treated here without this patch. Now it is treated further below (see line 169).

pkg/minikube/node/cache.go
obnoxxx: good point. I will change the patch accordingly.

obnoxxx: done, fwiw.

pkg/minikube/node/cache.go
obnoxxx: The `if downloadOnly` block is now within a bigger `if err == nil` block.
So this should be OK, right?

obnoxxx:
@medyagh , @afbjorklund FYI.

afbjorklund:
You can just skip the message, there is no need for such error output among the emojis?

There is also no need to pull the kicbase to the cache, if it is not going to be used?

obnoxxx:
@afbjorklund  wrote:

> You can just skip the message, there is no need for such error output among the emojis?

Good point. Removed.

 
> There is also no need to pull the kicbase to the cache, if it is not going to be used?

With the patch, pulling to cache is only done for the docker driver anyways.


new output with patch (also updated the description):

```console
$ make clean && make
$ ./out/minikube  delete --all
🔥  Deleting "minikube" in podman ...
🔥  Removing /home/obnox/.minikube/machines/minikube ...
$ time ./out/minikube start -d podman 
😄  minikube v1.37.0 on Fedora 43
✨  Using the podman driver based on user configuration
📌  Using Podman driver with root privileges
🚜  Pulling base image v0.0.48-1764169655-21974 ...
❗  minikube cannot pull kicbase image from any docker registry, and is trying to download kicbase tarball from github release page via HTTP.
❗  It's very likely that you have an internet issue. Please ensure that you can access the internet at least via HTTP, directly or with proxy. Currently your proxy configuration is:

    > kicbase-v0.0.48-amd64.tar:  1.22 GiB / 1.22 GiB  100.00% 18.34 MiB p/s 1m
🐳  Preparing Kubernetes v1.34.2 on Docker 29.0.4 ...

real    1m42.655s
user    0m6.445s
sys     0m6.654s
$ echo $?
0
$
```

afbjorklund:
If you don't pull anything to the cache, then there should not be any such output either.

```
🚜  Pulling base image v0.0.48-1764169655-21974 ...
❗  minikube cannot pull kicbase image from any docker registry, and is trying to download kicbase tarball from github release page via HTTP.
❗  It's very likely that you have an internet issue. Please ensure that you can access the internet at least via HTTP, directly or with proxy. Currently your proxy configuration is:

    > kicbase-v0.0.48-amd64.tar:  1.22 GiB / 1.22 GiB  100.00% 18.34 MiB p/s 1m
```

The first line could remain, as an information. But it will be podman doing the pulling.

obnoxxx:
@afbjorklund  wrote:

> If you don't pull anything to the cache, then there should not be any such output either.
> 
> ```
> 🚜  Pulling base image v0.0.48-1764169655-21974 ...
> ❗  minikube cannot pull kicbase image from any docker registry, and is trying to download kicbase tarball from github release page via HTTP.
> ❗  It's very likely that you have an internet issue. Please ensure that you can access the internet at least via HTTP, directly or with proxy. Currently your proxy configuration is:
> 
>     > kicbase-v0.0.48-amd64.tar:  1.22 GiB / 1.22 GiB  100.00% 18.34 MiB p/s 1m
> ```
> 
> The first line could remain, as an information. But it will be podman doing the pulling.

Good points, thanks!

Updated accordingly.

pkg/minikube/node/cache.go
```diff
@@ -170,7 +170,8 @@ func beginDownloadKicBaseImage(g *errgroup.Group, cc *config.ClusterConfig, down
 			}
 
 			if cc.Driver == driver.Podman {
-				return fmt.Errorf("not yet implemented, see issue #8426")
+				klog.Infof("iskipping image load from cache: not implemented for podman. see issue #8426")
```
medyagh: you can change this to warnningf

medyagh (changes requested):

pkg/minikube/node/cache.go
nirs: klog.Warningf is fine, but it should go only to the log (unless --alsologtostderr was used).

obnoxxx: @medyagh  wrote:
> you can change this to warnningf

done.

pkg/minikube/node/cache.go
```diff
@@ -170,7 +170,8 @@ func beginDownloadKicBaseImage(g *errgroup.Group, cc *config.ClusterConfig, down
 			}
 
 			if cc.Driver == driver.Podman {
-				return fmt.Errorf("not yet implemented, see issue #8426")
+				klog.Warningf("iskipping image load from cache: not implemented for podman. see issue #8426")
```
medyagh: typo!

pkg/minikube/node/cache.go
medyagh, line 236: add an errors.Is here for podman and that specific error. and if that is the case only log it with warnning.

pkg/minikube/node/cache.go
obnoxxx: Thanks for spotting! fixed.

pkg/minikube/node/cache.go
obnoxxx: @medyagh  wrote:

> add an errors.Is here for podman and that specific error. and if that is the case only log it with warnning.

Thanks for the suggestion. looking into it.

nirs (approved):

pkg/minikube/node/cache.go
obnoxxx: @medyagh  qrote:

> add an errors.Is here for podman and that specific error. and if that is the case only log it with warnning.

I am afraid I did not quite get your request yet:
This a different function.

Are you saying to log only in this function (`waitDownloadKicBaseImage`)  and  not in the other function (`beginDownloadKicBaseImage`) ?

obnoxxx: clarification by @medyagh  from chat: 

> show that you investigated the place that we checked the Value of returned error, and instead of returning nil we could check for the Type of Error there and Log it in the caller

PR #22025, opened 2025-12-02
Fix/ci windows newline mismatch

bobsira:


test/integration/testdata/hyperkit-driver-older-version/docker-machine-driver-hyperkit
nirs, line 1: This is not a text file:

```
$ file test/integration/testdata/hyperkit-driver-older-version/docker-machine-driver-hyperkit
test/integration/testdata/hyperkit-driver-older-version/docker-machine-driver-hyperkit: Mach-O 64-bit executable x86_64
```

This change should modify this file and other files in this commit.

.gitattributes
```diff
@@ -1,2 +1,2 @@
 # Prevent git from normalizing or converting any newline or encoding inside test data folders
-**/testdata/** text eol=lf
+**/testdata/** text eol=lf
```
nirs, line 3: I think you are missing a newline at the end of the file.

.gitattributes
```diff
@@ -1,2 +1,3 @@
 # Prevent git from normalizing or converting any newline or encoding inside test data folders
+
```
nirs, line 2: Why do we need this blank line? the comment is for single rule.

.gitattributes
nirs: To avoid changes in binary files (seen in the first commit), let's make the rule more specific, only for *.yaml, *.json, and *.txt files.

bobsira:
closing this PR and tracking this work here -> https://github.com/kubernetes/minikube/pull/22026

PR #22027, opened 2025-12-02
Dropping macOS 13 support

p172913:
What type of PR is this?
/kind enhancement

What this PR does / why we need it:

macOS 13 support removed and updated macOS 13 to macOS 15 along with updated docs to macOS 14

Which issue(s) this PR fixes:
Fixes https://github.com/kubernetes/minikube/issues/21666

Special notes for your reviewer:
None

Does this PR introduce a user-facing change?
None

Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:
None

PR #22034, opened 2025-12-04
prow: run macos functional tests in prow

ComradeProgrammer:

### DIFF
--- a/pkg/minikube/detect/detect.go
+++ b/pkg/minikube/detect/detect.go
@@ -111,11 +111,6 @@ func GithubActionRunner() bool {
 	return os.Getenv("GITHUB_ACTIONS") == "true"
 }
 
-func ProwCIEnvironment() bool {
-	_, ok := os.LookupEnv("PULL_NUMBER")
-	return ok
-}
-
 // ImageCacheDir returns the path in the minikube home directory to the container image cache for the current architecture
 func ImageCacheDir() string {
 	return filepath.Join(localpath.MakeMiniPath("cache", "images"), runtime.GOARCH)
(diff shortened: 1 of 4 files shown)


.github/workflows/functional_test.yml
```diff
@@ -88,16 +88,16 @@ jobs:
             cruntime: docker
             os: ubuntu-22.04
             test-timeout: 7m
-          - name: qemu-docker-macos-13-x86_64
+          - name: qemu-docker-macos-15-x86_64
             driver: qemu
             cruntime: docker
-            os: macos-13
+            os: macos-15
```
nirs: This is already handled in https://github.com/kubernetes/minikube/pull/22027
But the author is not responsive. Maybe you can take the correct changes to this commit.

Makefile
```diff
@@ -1085,7 +1085,6 @@ _update-all:
 	@(cd hack && go run update/update_all/update_all.go)
 
 
-
```
nirs: Removing this line triggers all actions?

.github/workflows/functional_test.yml
```diff
@@ -106,6 +106,7 @@ jobs:
         shell: bash
         run: |
           set -x
+          set +e
```
nirs: Since we set the shell to "bash" and not "bash -e", this should not be needed, since -e is not the default.

pkg/minikube/detect/detect.go
```diff
@@ -111,6 +111,11 @@ func GithubActionRunner() bool {
 	return os.Getenv("GITHUB_ACTIONS") == "true"
 }
 
+func ProwCIEnvironment() bool {
+	_, ok := os.LookupEnv("PULL_NUMBER")
```
medyagh: change this in follow up PR

PR #22035, opened 2025-12-04
ci: fix incorrect golang download url

ComradeProgrammer:
prow: fix incorrect golang download url


hack/prow/installer/check_install_golang.sh
```diff
@@ -89,7 +89,7 @@ function install_golang() {
 
   # using sudo because previously installed versions might have been installed by a different user.
   sudo rm -rf "$GO_TGZ"
-  curl -qL -O "https://storage.googleapis.com/golang/$GO_TGZ"
+  curl -qL -O "https://go.dev/dl/$GO_TGZ"
```
medyagh, line 92: plz change in jenkins copy too

PR #22054, opened 2025-12-07
WIP: Fix KVM + CRI-O test failures: use cni0 bridge and skip preload test 

divysinghvi:
Fix KVM + CRI-O test failures: use cni0 bridge and skip preload test


test/integration/preload_test.go
```diff
@@ -31,6 +31,10 @@ func TestPreload(t *testing.T) {
 		t.Skipf("skipping %s - incompatible with none driver", t.Name())
 	}
 
+	if ContainerRuntime() == "crio" {
+		t.Skipf("skipping %s - user-pulled images not persisted across restarts with crio", t.Name())
```
medyagh, line 35: add link to documentation

pkg/minikube/cni/bridge.go
```diff
@@ -41,7 +41,7 @@ var bridgeConf = template.Must(template.New("bridge").Parse(`
   "plugins": [
     {
       "type": "bridge",
-      "bridge": "bridge",
+      "bridge": "cni0",
```
medyagh, line 44: seems like we are changing this for everyeone all runtimes ? can you add more contexts maybe Before/AFter this PR 
and also maybe this could be Two Prs for two different test cases?

medyagh (changes requested):

pkg/minikube/cni/bridge.go
divysinghvi: I was experimenting with something it did not work will remove this

test/integration/preload_test.go
divysinghvi: will do

Issue #22071, opened 2025-12-08
flaky test TestFunctional/parallel/ImageCommands/ImageSaveDaemon on Github Action


medyagh:
The Functional Tests on PRs are being Red for on Docker Contained for one Test , (TestFunctional/parallel/ImageCommands/ImageSaveDaemon)
https://github.com/kubernetes/minikube/actions/workflows/functional_test.yml

https://github.com/kubernetes/minikube/actions/runs/19810406200/job/56751869820)

✗ TestFunctional/parallel/ImageCommands/ImageSaveDaemon
 X Exiting due to GUEST_IMAGE_SAVE: Failed to save image: failed to determine the image tag from tarball, err: unexpected EOF

here is a full log

```
=== RUN TestFunctional/parallel/ImageCommands/ImageSaveDaemon
functional_test.go:434: (dbg) Run: docker rmi kicbase/echo-server:functional-534732
functional_test.go:439: (dbg) Run: ./minikube-linux-amd64 -p functional-534732 image save --daemon kicbase/echo-server:functional-534732 --alsologtostderr
functional_test.go:439: (dbg) Non-zero exit: ./minikube-linux-amd64 -p functional-534732 image save --daemon kicbase/echo-server:functional-534732 --alsologtostderr: exit status 80 (297.392302ms)
-- stdout --

-- /stdout --
** stderr **
I1201 03:49:02.388151 16596 out.go:360] Setting OutFile to fd 1 ...
I1201 03:49:02.388595 16596 out.go:408] TERM=,COLORTERM=, which probably does not support color
I1201 03:49:02.388606 16596 out.go:374] Setting ErrFile to fd 2...
I1201 03:49:02.388612 16596 out.go:408] TERM=,COLORTERM=, which probably does not support color
I1201 03:49:02.388822 16596 root.go:338] Updating PATH: /home/runner/.minikube/bin
I1201 03:49:02.389367 16596 config.go:182] Loaded profile config "functional-534732": Driver=docker, ContainerRuntime=containerd, KubernetesVersion=v1.34.1
I1201 03:49:02.389401 16596 cache_images.go:404] Save images: ["kicbase/echo-server:functional-534732"]
I1201 03:49:02.389507 16596 config.go:182] Loaded profile config "functional-534732": Driver=docker, ContainerRuntime=containerd, KubernetesVersion=v1.34.1
[...]

PR #22072, opened 2025-12-08
image save: Add retry logic for local image in case of EOF

medyagh:
to help improve flakes like this https://github.com/kubernetes/minikube/issues/22071

retry if  the error indicates an unexpected or premature EOF that can be retried.

### DIFF
--- a/.github/workflows/functional_test.yml
+++ b/.github/workflows/functional_test.yml
@@ -87,7 +87,7 @@ jobs:
             driver: none
             cruntime: docker
             os: ubuntu-22.04
-            test-timeout: 9m
+            test-timeout: 7m
           - name: qemu-docker-macos-13-x86_64
             driver: qemu
             cruntime: docker


pkg/minikube/image/image.go
```diff
@@ -271,6 +267,52 @@ func uploadRemote(ref name.Reference, img v1.Image, p v1.Platform) error {
...
+			return nil, lastErr
+		}
+		klog.Infof("retrying manifest read for %s due to EOF (%d/%d)", path, i+1, maxAttempts)
+	}
+	return nil, lastErr
+}
+
+func isRetryableEOF(err error) bool {
+	return errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, io.EOF) || strings.Contains(err.Error(), "unexpected EOF")
+}
```
nirs, line 314: How EOF can succeed after retry?

medyagh: I think this is because the image file saved needs a few miliseconds in slow read/write disks to be available and when it is not that it is repots as EOF.

Issue #22067, opened 2025-12-08
tests: Too long timeout - test fails after 6+ minutes


nirs:
I'm not sure what is the purpose of `TestFunctionalNewestKubernetes/Versionv1.35.0-beta.0/parallel/PersistentVolumeClaim` but there is no reason why the test should fail after 6 minutes.

Checking the logs shows that it failed because the known rate-limit issue, but it is not clear why we need to pull nginx image for testing a PVC.

We need to minimize the test timeout so it fails quickly.

```
	Events:
	  Type     Reason     Age                  From               Message
	  ----     ------     ----                 ----               -------
	  Normal   Scheduled  6m3s                 default-scheduler  Successfully assigned default/sp-pod to functional-940895
	  Warning  Failed     5m16s                kubelet            Failed to pull image "docker.io/nginx": copying system image from manifest list: determining manifest MIME type for docker://nginx:latest: reading manifest sha256:5c733364e9a8f7e6d7289ceaad623c6600479fe95c3ab5534f07bfd7416d9541 in docker.io/library/nginx: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
	  Normal   Pulling    47s (x5 over 6m)     kubelet            Pulling image "docker.io/nginx"
	  Warning  Failed     15s (x5 over 5m16s)  kubelet            Error: ErrImagePull
	  Warning  Failed     15s (x4 over 4m28s)  kubelet            Failed to pull image "docker.io/nginx": reading manifest latest in docker.io/library/nginx: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
	  Normal   BackOff    1s (x11 over 5m16s)  kubelet            Back-off pulling image "docker.io/nginx"
	  Warning  Failed     1s (x11 over 5m16s)  kubelet            Error: ImagePullBackOff
-- /stdout --
helpers_test.go:293: <<< TestFunctionalNewestKubernetes/Versionv1.35.0-beta.0/parallel/PersistentVolumeClaim FAILED: end of post-mortem logs <<<
helpers_test.go:294: ---------------------/post-mortem---------------------------------
--- FAIL: TestFunctionalNewestKubernetes/Versionv1.35.0-beta.0/parallel/PersistentVolumeClaim (369.63s)
```

Full test log:
https://storage.googleapis.com/minikube-builds/logs/21409/42683/KVM_Linux_crio.html

[...]

PR #22081, opened 2025-12-09
test: Fail Fast if docker hub rate limited 

medyagh:
fixes https://github.com/kubernetes/minikube/issues/22067


test/integration/addons_test.go
```diff
@@ -46,8 +47,12 @@ import (
 	"k8s.io/minikube/pkg/util/retry"
 )
 
+const dockerHubRateLimitURL = "https://registry-1.docker.io/v2/"
+
 // TestAddons tests addons that require no special environment in parallel
 func TestAddons(t *testing.T) {
+	skipAddonsIfDockerHubRateLimited(t)
```
nirs: This will skip the test if we detected a rate limit, but what if the rate limit is detected on the next request?

The right way to handle this is to check the error from image pull, and abort the test with a well known error that will abort all other tests that depends on docker hub.

Instead we add one check before starting one test that may not be relevant to the test once we move all images out of dockerhub.

Also dockerhub related code should be in dockerhub package and not in addon test file. We may want to use this in any test that depend on dockerhub.

test/integration/addons_test.go
```diff
@@ -189,6 +194,61 @@ func TestAddons(t *testing.T) {
 	})
 }
 
+// skipAddonsIfDockerHubRateLimited proactively stops the addon suite when Docker Hub is rate limited.
+func skipAddonsIfDockerHubRateLimited(t *testing.T) {
+	t.Helper()
+
+	client := &http.Client{Timeout: 10 * time.Second}
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, dockerHubRateLimitURL, nil)
```
nirs: Use the t.Context()

test/integration/addons_test.go
```diff
@@ -46,8 +47,12 @@ import (
 	"k8s.io/minikube/pkg/util/retry"
 )
 
+const dockerHubRateLimitURL = "https://registry-1.docker.io/v2/"
```
nirs: Are you sure that rate-limit is applied to this URL in the same way it is applied to image pulls?

test/integration/addons_test.go
```diff
@@ -189,6 +194,61 @@ func TestAddons(t *testing.T) {
...
+		t.Logf("unable to check Docker Hub rate limit (continuing): %v", err)
+		return
+	}
+	resp.Body.Close()
+
+	if resp.StatusCode == http.StatusTooManyRequests {
+		t.Fatalf("failing TestAddons: Docker Hub rate limit exceeded (HTTP 429)")
+	}
+
+	remainingHeader := resp.Header.Get("RateLimit-Remaining")
```
nirs: Where is this documented?

nirs (commented):
Function for checking dockerhub rate limit looks useful for it should be be in this test file.

test/integration/addons_test.go
medyagh: seems like it, based on internet

test/integration/addons_test.go
medyagh: minikube's own code does not depend on docker.io just addons do. I think once we implement something that uses then it can be moved to that package. meanwhile could be a humble helper func

medyagh: all minikube code that would use docker.io images woudl be in test data or they would be in the addons ... so I think this is the big one before we can force all addons use non docker.io images

test/integration/addons_test.go
medyagh: added documentation

PR #21273, opened 2025-08-09
spelling: Fix spelling errors with codespell

nirs:
Fix trivial spelling errors using codespell[1]:

    codespell --skip '*.yaml,*.tmpl,*.json,*.html,*.patch,go.sum' -w

And rejecting some false positives fixes:

    ./CHANGELOG.md:907: fliter ==> filter
    ./third_party/go9p/clnt_write.go:48: Writen ==> Written
    ./third_party/kubeadm/app/features/features.go:69: AtLeast ==> at least
    ./site/content/en/docs/contrib/translations.md:106: certificats ==> certificates
    ./site/content/en/docs/contrib/translations.md:113: espace ==> escape
    ./site/content/en/docs/tutorials/amd.md:75: HSA ==> HAS
    ./site/content/en/docs/tutorials/amd.md:87: HSA ==> HAS
    ./pkg/minikube/config/extra_options_test.go:143: expRes ==> express
    ./pkg/minikube/config/extra_options_test.go:151: expRes ==> express
    ./pkg/minikube/config/extra_options_test.go:152: expRes ==> express
    ./pkg/minikube/config/extra_options_test.go:168: expRes ==> express
    ./pkg/minikube/config/extra_options_test.go:177: expRes ==> express
    ./pkg/minikube/config/extra_options_test.go:178: expRes ==> express

There are more spelling errors that need manual selection:

    ./CHANGELOG.md:234: issuse ==> issue, issues
    ./CHANGELOG.md:543: Pris ==> Prise, Prism
    ./hack/benchmark/time-to-k8s/page.go:73: readin ==> reading, read in
    ./hack/benchmark/image-build/generate-chart.go:82: INTERATIVE ==> INTERACTIVE, ITERATIVE
    ./hack/benchmark/image-build/generate-chart.go:87: INTERATIVE ==> INTERACTIVE, ITERATIVE
    ./hack/benchmark/image-build/generate-chart.go:137: INTERATIVE ==> INTERACTIVE, ITERATIVE
    ./hack/benchmark/image-build/generate-chart.go:162: interative ==> interactive, iterative
    ./hack/benchmark/image-build/generate-chart.go:195: INTERATIVE ==> INTERACTIVE, ITERATIVE
    ./third_party/go9p/fmt.go:132: Tread ==> Thread, Treat
    ./third_party/go9p/fmt.go:133: Tread ==> Thread, Treat
    ./third_party/go9p/p9.go:33: Tread ==> Thread, Treat
[...]

PR #22086, opened 2025-12-09
docs: Fix spelling errors with codespell

Varsius:
This PR introduces another round of spellchecking, similar to #21273.

While #21273 proposes adding a codespell step to CI, I would currently advise against doing so. The number of false positives is still too high, and there is no clean way to generically exclude them. Additionally, the majority of the identified issues require manual decisions between multiple valid alternatives.

For reference, these are the commands I used to identify most of the changes in this PR:

```
codespell --skip '*.yaml,*.tmpl,*.json,*.html,*.patch,go.sum,./third_party' -i 3 -w

typos   --exclude '*.yaml'   --exclude '*.tmpl'   --exclude '*.json'   --exclude '*.html'   --exclude '*.patch'   --exclude 'go.sum'   --exclude 'third_party/**'
```

### DIFF
--- a/hack/benchmark/image-build/generate-chart.go
+++ b/hack/benchmark/image-build/generate-chart.go
@@ -79,14 +79,14 @@ var RuntimeMethods = map[string][]string{
 }
 
 const (
-	INTERATIVE    = "Iterative"
-	NONINTERATIVE = "NonIterative"
+	ITERATIVE    = "Iterative"
+	NONITERATIVE = "NonIterative"
 )
 
 var Itrs = []string{
-	INTERATIVE,
-	// to simplify the output, non-interative is omitted
-	// NONINTERATIVE,
+	ITERATIVE,
+	// to simplify the output, non-iterative is omitted
+	// NONITERATIVE,
 }
 
 // method name-> test result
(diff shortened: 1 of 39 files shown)


hack/benchmark/image-build/generate-chart.go
```diff
@@ -79,14 +79,14 @@ var RuntimeMethods = map[string][]string{
 }
 
 const (
-	INTERATIVE    = "Iterative"
```
medyagh, line 82: these are environment variables that are used by other scripts, plz revert these changes

medyagh (changes requested):

hack/benchmark/image-build/generate-chart.go
Varsius: Done 👍🏻

Varsius:
@medyagh please take another look

Varsius:
@prezha maybe you could take a look? Please let me know if there are any actions points left from my side

nirs:
@Varsius can you rebase on master to fix the conflicts?

nirs:
> For reference, these are the commands I used to identify most of the changes in this PR:
> 
> ```
> codespell --skip '*.yaml,*.tmpl,*.json,*.html,*.patch,go.sum,./third_party' -i 3 -w
> 
> typos   --exclude '*.yaml'   --exclude '*.tmpl'   --exclude '*.json'   --exclude '*.html'   --exclude '*.patch'   --exclude 'go.sum'   --exclude 'third_party/**'
> ```

@Varsius This is very useful info! Can  you add a make target running this to help developers check spelling?

Varsius:
@nirs I rebased and added the Makefile target

PR #22097, opened 2025-12-10
replace mkisofs command with native xorriso commands

vtri950:
When building the aarch64 ISO on Fedora, the build failed with:
`xorriso : FAILURE : -as genisofs: Unrecognized option '-eltorito-platform=efi'`

On Fedora, the mkisofs command is actually implemented by xorriso using -as genisofs emulation mode. However, newer versions of xorriso don't support the -eltorito-platform=efi flag when emulating genisofs.

### The Fix
I replaced the mkisofs command in [post-image.sh](vscode-file://vscode-app/c:/Users/vtripathi/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with native xorriso commands.

fixes #22056

PR #22098, opened 2025-12-10
add iso build instructions for fedora os

### LINKED ISSUE #22055: iso: Support build on Fedora - x86_64
Since we removed gluster, we don't need the legacy python2 dependency and we can support building on more modern distros like Fedora.

Currently the x86_64 build works on Fedora 43 by installing these tools:

```console
sudo dnf install \
    7zip \
    automake \
    bc \
    bzip2 \
    cpio \
    gcc \
    gcc-c++ \
    git \
    gnupg2 \
    patch \
    perl-core \
    libtool \
    make \
    ncurses-devel \
    unzip \
    wget \
    which \
```

The aarch64 version has more issues but we can start with the simple change.

The docs need to describe 2 systems for building (Ubuntu 22.04, Fedora 43). Each system need its own instructions for installing require tools. The actual build instructions should be shared since they must be the same.

vtri950:
Added separate installation commands for Ubuntu and Fedora, including the complete list of packages used (with xorriso which is critical for the aarch64 build)

fixes #22055

### DIFF
--- a/site/content/en/docs/contrib/building/iso.md
+++ b/site/content/en/docs/contrib/building/iso.md
@@ -11,7 +11,7 @@ The minikube ISO is booted by each hypervisor to provide a stable minimal Linux 
 ## Prerequisites
 
 * Machine with x86\_64 CPU
-* Ubuntu 22.04.5 LTS (Jammy Jellyfish)
+* Ubuntu 22.04.5 LTS (Jammy Jellyfish) or Fedora 43
 * docker
 * A recent GNU Make distribution (>=4.0)
 * A recent Go distribution (>=1.22.0)
@@ -50,6 +50,11 @@ The bootable ISO image will be available in `out/minikube-<arch>.iso`.
 
 Install required tools:
 
+Install Go using these instructions:
+https://go.dev/doc/install
+
+### On Ubuntu
+
 ```shell
 sudo apt-get install \
     automake \
@@ -68,8 +73,31 @@ sudo apt-get install \
     wget \
 ```
 
-Install Go using these instructions:
-https://go.dev/doc/install
+### On Fedora
+
+```shell
+sudo dnf install -y \
+    7zip \
+    automake \
+    bc \
+    cpio \
+    gcc \
+    gcc-c++ \
+    genisoimage \
+    git \
+    glibc-devel \
+    glibc-devel.i686 \
+    glibc-langpack-en \
+    gnupg2 \
+    kernel-devel \
+    libtool \
+    make \
+    patch \
+    perl \
+    unzip \
+    wget \
+    xorriso
+```
 
 To build without docker run:
 


site/content/en/docs/contrib/building/iso.md
```diff
@@ -68,6 +70,34 @@ sudo apt-get install \
...
+    glibc-devel.i686 \
+    glibc-devel \
+    cpio \
+    genisoimage \
+    git \
+    gnupg2 \
+    libtool \
+    glibc-langpack-en \
+    p7zip \
+    p7zip-plugins \
```
nirs: This does not work on Fedora 43, we need the `7zip` package (see the example in the issue).

I would include only instructions for Fedora 43, since this is the current version. Fedora 42 will be EOL in a year or so there is no point to complicate the instructions.

site/content/en/docs/contrib/building/iso.md
```diff
@@ -11,7 +11,7 @@ The minikube ISO is booted by each hypervisor to provide a stable minimal Linux
 ## Prerequisites
 
 * Machine with x86\_64 CPU
-* Ubuntu 22.04.5 LTS (Jammy Jellyfish)
+* Ubuntu 22.04.5 LTS (Jammy Jellyfish) or Fedora 42/43
```
nirs: Let's use only Fedora 43 to simplify.

site/content/en/docs/contrib/building/iso.md
vtri950: done

site/content/en/docs/contrib/building/iso.md
vtri950: done

site/content/en/docs/contrib/building/iso.md
```diff
@@ -50,14 +50,16 @@ The bootable ISO image will be available in `out/minikube-<arch>.iso`.
 
 Install required tools:
 
+**On Ubuntu:**
```
nirs: This should be a sub heading:

```markdown
### On Ubuntu
```

site/content/en/docs/contrib/building/iso.md
```diff
@@ -68,6 +70,33 @@ sudo apt-get install \
     wget \
 ```
 
+**On Fedora:**
```
nirs: Use sub heading like ubuntu

site/content/en/docs/contrib/building/iso.md
```diff
@@ -68,6 +70,33 @@ sudo apt-get install \
...
+    wget \
+    make \
+    kernel-devel \
+    patch \
+    perl \
+    xorriso \
+    golang
+```
+
 Install Go using these instructions:
```
nirs: This should be under common sub heading - same on all distros.

site/content/en/docs/contrib/building/iso.md
```diff
@@ -68,6 +70,33 @@ sudo apt-get install \
...
+    gcc-c++ \
+    glibc-devel.i686 \
+    glibc-devel \
+    cpio \
+    genisoimage \
+    git \
+    gnupg2 \
+    libtool \
+    glibc-langpack-en \
+    7zip \
```
nirs: This is not sorted - sort the packages

site/content/en/docs/contrib/building/iso.md
```diff
@@ -68,6 +70,33 @@ sudo apt-get install \
...
+    unzip \
+    wget \
+    make \
+    kernel-devel \
+    patch \
+    perl \
+    xorriso \
+    golang
```
nirs: If you install golang here, we don't need the common Install go comment on the bottom.

I think we can remove the golang package here and keep the common golang section bellow. Minikube always uses latest golang and using the bistro golang will usually provide old version.

site/content/en/docs/contrib/building/iso.md
```diff
@@ -50,14 +50,16 @@ The bootable ISO image will be available in `out/minikube-<arch>.iso`.
...
+
 ```shell
 sudo apt-get install \
     automake \
     bc \
     build-essential \
     cpio \
     gcc-multilib \
-    genisoimage \
+    xorriso \
```
nirs: Need sorting

nirs (commented):
The instruction should include a note that building arm64 on Fedora does not work yet. It can be something like:

> [!WARNING]
> Building arm64 iso on Fedora does not work yet

site/content/en/docs/contrib/building/iso.md
nirs: This also depends on #22097 - so it should be part of that PR.

vtri950:
@nirs is this pr good to merge or is there anything else that needs fixing?

PR #22109, opened 2025-12-11
ci: remove free gha macos functioanl tests

medyagh:
also bump all ubuntu22.04 to ubuntu24.04 for functional tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1096,4 +1096,3 @@ _update-all:
 
 # targets for tests on prow
 include ./hack/prow/prow.mk
-


.github/workflows/functional_test.yml
```diff
@@ -32,7 +32,7 @@ permissions:
 jobs:
   # build-test-binaries job runs before all other jobs and produces binaries/test-data to be shared by all following jobs
   build-test-binaries:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
nirs, line 35: This change is not related to removing the macos functional tests

nirs (commented):
"also" means we mix unrelated changes, would be nice to split it to another commit.

.github/workflows/functional_test.yml
medyagh: fair enough

PR #22110, opened 2025-12-11
ci: refactor functional test to build in x86 and armt64 in parallel

medyagh:

### DIFF
--- a/.github/workflows/functional_test.yml
+++ b/.github/workflows/functional_test.yml
@@ -41,7 +41,7 @@ jobs:
             arch: amd64
             runs-on: ubuntu-24.04
             make-targets: e2e-linux-amd64
-          - name: build-test-binaries-arm
+          - name: build-test-binaries-arm64
             arch: arm64
             runs-on: ubuntu-24.04-arm
             make-targets: e2e-linux-arm64


.github/workflows/functional_test.yml
```diff
@@ -43,16 +43,37 @@ jobs:
...
-          make e2e-linux-arm64 e2e-linux-amd64 e2e-darwin-amd64
+          make e2e-linux-amd64
           cp -r test/integration/testdata ./out
       - name: Upload Test Binaries
         uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4
         with:
-          name: binaries
+          name: binaries-amd64
+          path: out
+  build-test-binaries-arm64:
```
nirs: Use matrix?

.github/workflows/functional_test.yml
```diff
@@ -43,16 +43,37 @@ jobs:
...
+        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4
+        with:
+          name: binaries-arm64
           path: out
   functional-test:
     name: ${{ matrix.name }}
-    needs: build-test-binaries
+    needs:
+      - build-test-binaries-amd64
+      - build-test-binaries-arm64
```
nirs: If we use matrix, maybe we can depend on single job name? Will make it easier to build more binaries in parallel in the future (darwin-arm64, windows-amd64, windows-arm64).

.github/workflows/functional_test.yml
medyagh: done

.github/workflows/functional_test.yml
medyagh: done thanks

nirs (approved):
Nice

PR #22111, opened 2025-12-12
ci: adjust cpu/memory for smoke tests on macos

medyagh:

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -126,7 +126,7 @@ jobs:
               --no-kubernetes \
               --memory 1200mb \
               --force \
-              --cpus 2 \
+              --cpus 1 \
               --wait-timeout=13m \
               --driver ${{ matrix.driver }} \
               ${{ matrix.network_flag }} \


.github/workflows/smoke-test.yml
```diff
@@ -124,7 +124,7 @@ jobs:
         run: |
           ./out/minikube start \
               --no-kubernetes \
-              --memory 2gb \
+              --memory 1200mb \
```
nirs, line 127: 1200mb may be too low because our image need about 800mb for the k8s components, so I would use 1800mb with is also our minimal size (there is a warning in the log).

I did not test small memory with --no-kubernetes. We can test it locally to find the real limit.

.github/workflows/smoke-test.yml
```diff
@@ -124,7 +124,7 @@ jobs:
...
               --cpus 2 \
```
nirs: In my tests on the intel runners, 1 cpu is most reliable configuration. But I did not test more than 1 cpu for long time, so maybe it works fine now.

.github/workflows/smoke-test.yml
```diff
@@ -124,7 +124,7 @@ jobs:
...
               --wait-timeout=13m \
```
nirs, line 130: Starting without kubernetes should take 2-3 minutes. No reason to wait 13 minutes. Waiting 5 minutes should be good enough - if it did not work we need to fix the issue with the configuration.

.github/workflows/smoke-test.yml
```diff
@@ -125,6 +125,7 @@ jobs:
           ./out/minikube start \
               --no-kubernetes \
               --memory 1200mb \
+              --force \
```
nirs, line 128: Why dod we ned force?

medyagh: force because minikube wont let you do it ! minikube forces you to have at least 1800mb (because required by kubernetes and we NOW it wont work)

we can actually change minikube to not have that requirement for no-kubernetes if we prove it works...but basiclalyl this is to help us Force ppl give them enough memeory that wont cause flakes that will make many github issues that we will need to investiage.

nirs: ok, I would use 1800 and remove the force.

PR #22116, opened 2025-12-12
iso: Fix go.hash override

### LINKED ISSUE #22104: Upstream GO_HASH_FILE variable for buildroot go package
We define GO_HASH_FILE= to override buildroot go.hash file:

```Makefile
# the go version on the line below is for the ISO
GOLANG_OPTIONS = GOWORK=off GO_VERSION=1.23.4 GO_HASH_FILE=$(PWD)/deploy/iso/minikube-iso/go.hash
BUILDROOT_OPTIONS = BR2_EXTERNAL=../../deploy/iso/minikube-iso $(GOLANG_OPTIONS)
```

but there is no such variable in the buildroot go package, 
So we just replace the existing one:

```Makeile
cp deploy/iso/minikube-iso/go.hash $(BUILD_DIR)/buildroot/package/go/go.hash
```

Probably should be upstreamed properly...

The real files for the "go" package are in https://github.com/buildroot/buildroot/tree/2025.02.x/package/go

We can upstream this change to the buildroot package so we don't need to replace the go.hash file.

_Originally posted by @afbjorklund in https://github.com/kubernetes/minikube/issues/22101#issuecomment-3641539986_

nirs:
We used `GO_HASH_FILE` which is static, but since buildroot 2024.02.x we
can use `GO_HASH_FILES`.
    
This avoids the need to clobber the buildroot go.hash file.

Fixes #22104

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ GO_K8S_VERSION_PREFIX ?= v1.35.0
 
 INSTALL_SIZE ?= $(shell du out/minikube-windows-amd64.exe | cut -f1)
 BUILDROOT_BRANCH ?= 2025.02
-GOLANG_OPTIONS = GOWORK=off GO_VERSION=$(GO_VERSION) GO_HASH_FILES=$(PWD)/deploy/iso/minikube-iso/go.hash
+GOLANG_OPTIONS = GO_VERSION=$(GO_VERSION) GO_HASH_FILES=$(PWD)/deploy/iso/minikube-iso/go.hash
 BUILDROOT_OPTIONS = BR2_EXTERNAL=../../deploy/iso/minikube-iso $(GOLANG_OPTIONS)
 REGISTRY ?= gcr.io/k8s-minikube
 
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ GO_K8S_VERSION_PREFIX ?= v1.35.0
 
 INSTALL_SIZE ?= $(shell du out/minikube-windows-amd64.exe | cut -f1)
 BUILDROOT_BRANCH ?= 2025.02
-GOLANG_OPTIONS = GOWORK=off GO_VERSION=$(GO_VERSION) GO_HASH_FILE=$(PWD)/deploy/iso/minikube-iso/go.hash
+GOLANG_OPTIONS = GOWORK=off GO_VERSION=$(GO_VERSION) GO_HASH_FILES=$(PWD)/deploy/iso/minikube-iso/go.hash
 BUILDROOT_OPTIONS = BR2_EXTERNAL=../../deploy/iso/minikube-iso $(GOLANG_OPTIONS)
 REGISTRY ?= gcr.io/k8s-minikube
 
@@ -314,7 +314,6 @@ buildroot:
 		git clone --depth=1 --branch=$(BUILDROOT_BRANCH) https://github.com/buildroot/buildroot $(BUILD_DIR)/buildroot; \
 		perl -pi -e 's@\s+source "package/sysdig/Config\.in"\n@@;' $(BUILD_DIR)/buildroot/package/Config.in; \
 		rm -r $(BUILD_DIR)/buildroot/package/sysdig; \
-		cp deploy/iso/minikube-iso/go.hash $(BUILD_DIR)/buildroot/package/go/go.hash; \
 		git --git-dir=$(BUILD_DIR)/buildroot/.git config user.email "dev@random.com"; \
 		git --git-dir=$(BUILD_DIR)/buildroot/.git config user.name "Random developer"; \
 	fi;


nirs:
@afbjorklund Using GO_HASH_FILES does not work, maybe I missed something, or maybe we need to upgrade to newer buildroot?

```
2025-12-12 22:08:52 (122 MB/s) - ‘/home/jenkins/workspace/iso-pr-build/out/buildroot/output-aarch64/build/.go1.25.5.linux-amd64.tar.gz.3souPK/output’ saved [59768009/59768009]

ERROR: No hash found for go1.25.5.linux-amd64.tar.gz
```

Tested also locally on top of change to update to 2025.02.9, and it fails in the same way:

```console
>>> host-go-bin 1.25.5 Downloading
wget -nd -t 3 --connect-timeout=10 -O '/home/nsoffer/minikube/out/buildroot/output-x86_64/build/.go1.25.5.linux-amd64.tar.gz.s29vsG/output' 'https://go.dev/dl/go1.25.5.linux-amd64.tar.gz' 
Failed to connect (hostname='go.dev', errno=101)
Failed to connect (hostname='go.dev', errno=101)
Failed to connect (hostname='go.dev', errno=101)
Failed to connect (hostname='go.dev', errno=101)
HTTP response 302  [https://go.dev/dl/go1.25.5.linux-amd64.tar.gz]
Adding URL: https://dl.google.com/go/go1.25.5.linux-amd64.tar.gz
Adding URL: https://dl.google.com/go/go1.25.5.linux-amd64.tar.gz
Failed to connect (hostname='dl.google.com', errno=101)
Failed to connect (hostname='dl.google.com', errno=101)
Failed to connect (hostname='dl.google.com', errno=101)
Failed to connect (hostname='dl.google.com', errno=101)
Saving '/home/nsoffer/minikube/out/buildroot/output-x86_64/build/.go1.25.5.linux-amd64.tar.gz.s29vsG/output'
HTTP response 200  [https://dl.google.com/go/go1.25.5.linux-amd64.tar.gz]
/home/nsoffer/miniku 100% [========================================================================================================================================================>]   56.99M   69.95MB/s
                          [Files: 1  Bytes: 56.99M [31.43MB/s] Redirects: 1  Todo: 0  Errors: 0                                                                                     ]
ERROR: No hash found for go1.25.5.linux-amd64.tar.gz
```

So seems that overriding does not work for GO_HASH_FILES, or I did not use it correctly.

afbjorklund:
See https://github.com/kubernetes/minikube/issues/22104#issuecomment-3642371308, there is no variable...

You just need to move the go.hash, from the current location into the patch directory. And then remove the cp.

EDIT: You probably need **two** copies, though. Might even want to split the binary files, for each patch directory?

deploy/iso/minikube-iso/configs/minikube_aarch64_defconfig:BR2_GLOBAL_PATCH_DIR="$(BR2_EXTERNAL_MINIKUBE_PATH)/board/minikube/aarch64/patches"

deploy/iso/minikube-iso/board/minikube/aarch64/patches/go/go.hash

sha256  22a5fd0a91efcd28a1b0537106b9959b2804b61f59c3758b51e8e5429c1a954f  go1.25.5.src.tar.gz
sha256  b00b694903d126c588c378e72d3545549935d3982635ba3f7a964c9fa23fe3b9  go1.25.5.linux-arm64.tar.gz

deploy/iso/minikube-iso/configs/minikube_x86_64_defconfig:BR2_GLOBAL_PATCH_DIR="$(BR2_EXTERNAL_MINIKUBE_PATH)/board/minikube/x86_64/patches"

deploy/iso/minikube-iso/board/minikube/x86_64/patches/go/go.hash

sha256  22a5fd0a91efcd28a1b0537106b9959b2804b61f59c3758b51e8e5429c1a954f  go1.25.5.src.tar.gz
sha256  9e9b755d63b36acf30c12a9a3fc379243714c1c6d3dd72861da637f336ebb35b  go1.25.5.linux-amd64.tar.gz

It would be easier to not have two directories (one for each arch), but to try to collect them into a single tree.

nirs:
> See [#22104 (comment)](https://github.com/kubernetes/minikube/issues/22104#issuecomment-3642371308), there is no variable...
> 
> You just need to move the go.hash, from the current location into the patch directory. And then remove the cp.

So we cannot override GO_HASH_FILES just like the old GO_HASH_FILE, but instead we can add a patches directory with out go.hash file, and it will override the buildroot hash file?

> EDIT: You probably need **two** copies, though. Might even want to split the binary files, for each patch directory?

I need 2 patch directories because we have one directory for each arch?

> It would be easier to not have two directories (one for each arch), but to try to collect them into a single tree.

We cannot have a single patches directory used by both arches?

afbjorklund:
> So we cannot override GO_HASH_FILES just like the old GO_HASH_FILE,

The old variable couldn't be overridden either, it was just left in "for docs" (or something)

So previously we replaced the "go.hash" file in buildroot, and now that process is cleaner

As explained in https://github.com/buildroot/buildroot/commit/5d36710e36fc4698c8fae71675bcff7395246006

afbjorklund:
> We cannot have a single patches directory used by both arches?

I moved the arch cleanup to a separate issue: https://github.com/kubernetes/minikube/issues/22135

----

I think historically there was a problem with the docker downloads, that cause there to be two?

deploy/iso/minikube-iso/arch/aarch64/package/docker-bin-aarch64/docker-bin.hash

`sha256 9e4f82996ab790724094475ebed33a736434bfe5d45231b676fef22ffb80044d  docker-28.5.2.tgz`

deploy/iso/minikube-iso/arch/x86_64/package/docker-bin/docker-bin.hash

`sha256 ea90cfd12e1eeb12aa1c971741adb8bd4ed88e2a574eaac13f5029a1dbc6300d  docker-28.5.2.tgz`

But I'm sure there is a better fix for that, and it is definitely not needed by the rest of the packages.

EDIT: seems that you just define a package SOURCE (with arch in it)

nirs:
> As explained in [buildroot/buildroot@5d36710](https://github.com/buildroot/buildroot/commit/5d36710e36fc4698c8fae71675bcff7395246006)

The commit message explain the problem and the solution very well.

I need an example how to configure a global patch directory.

afbjorklund:
> I need an example how to configure a global patch directory.

It is in the defconfig, but it will have to wait until arch is merged?

deploy/iso/minikube-iso/board/minikube/x86_64/patches
deploy/iso/minikube-iso/board/minikube/aarch64/patches

=> deploy/iso/minikube-iso/patches

deploy/iso/minikube-iso/arch/x86_64/package
deploy/iso/minikube-iso/arch/aarch64/package

=> deploy/iso/minikube-iso/package

nirs:
> > I need an example how to configure a global patch directory.
> 
> It is in the defconfig, but it will have to wait until arch is merged?

Do you mean until #22135 is fixed?

> deploy/iso/minikube-iso/board/minikube/x86_64/patches deploy/iso/minikube-iso/board/minikube/aarch64/patches

Keep the one go.hash file, and add a symlink in `deploy/iso/minikube-iso/board/minikube/*/patches` ?

afbjorklund:
I cleaned all of it, in a new PR (#22116) - you might still want to remove the `GOWORK`

nirs:
Not needed now

PR #22121, opened 2025-12-13
ci: replace macOS and Linux info blocks with a reusable action

medyagh:


Makefile
```diff
@@ -1092,5 +1092,6 @@ get-dependency-version:
 _update-all:
 	@(cd hack && go run update/update_all/update_all.go)
 
+
```
nirs, line 1095: I don't think this is needed, keeping one blank line between items is better.

PR #22140, opened 2025-12-13
vfkit: Add Rosetta support for deploying amd64 images on Apple silicon

### LINKED ISSUE #20559: vfkit: Rosetta support - run amd64 containers on Apple silicon macs
### What Happened?

Some projects do not build container images for arm64 (aarch64 on linux). When using minikube on Apple silicon macs, the container will fail to start.

vfkit already support Rosetta:
https://github.com/crc-org/vfkit/blob/main/doc/usage.md#rosetta

Enabling the feature requires:

1. Adding --device rosetta:

```
--device rosetta,mountTag=rosetta-share
```

2. Mounting the roseta share in the guest:

```
mount -t virtiofs vfkitTag /mnt
```

3. Configure [binfmt](https://docs.kernel.org/admin-guide/binfmt-misc.html) to use this rosetta binary for x86_64 executables 

The minimal change is to add a --device rosetta flag to the vfkit command. Since this may not wanted in all cases, we need a minikube start flag to enable this feature.
The rest of the configuration can be done manually by the user.

Ideally enable rosetta in minikube start command will also handle the mount in the guest and configure binfmt.

## See also

- [lima](https://lima-vm.io/docs/config/multi-arch/#fast-mode-2-rosetta-intel-containers-on-arm-vm-on-arm-hostfast-mode-2) have full support for rosetta. You only need to add the --rosetta flag and it handles the rest.

### Attach the log file

(not relevant)

### Operating System

macOS (Default)

### Driver

None

nirs:
Add --rosetta flag enabling Rosetta[1] for running apps built for Intel processor on Mac with Apple silicon. An example use case is deploying open-cluster-management that do not provide arm64 builds yet.

If Rosetta is not installed on the host, it will be installed on the first time starting a cluster. When running in non-interactive mode automatic install is disabled and if Rosetta is not installed start will fail.

The --rosetta flag is ignored with a warning if enabled on Mac with Intel processor.

## Testing amd64 executable on Apple silicon

```console
% cat rosetta/test.go 
package main

import (
        "bytes"
        "fmt"
        "os/exec"
        "runtime"
)

func main() {
        uname, _ := exec.Command("uname", "-a").Output()
        fmt.Printf("✅ Running %s binary on %s\n", runtime.GOARCH, bytes.TrimSpace(uname))
}

% GOARCH=amd64 GOOS=linux go build -o rosetta/test rosetta/test.go

% file rosetta/test
rosetta/test: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, BuildID[sha1]=9743ccad9a20c7d535a9794cc7b22ed1147c89d8, with debug_info, not stripped

% out/minikube start --driver vfkit --rosetta --mount-string $PWD/rosetta:/mnt/rosetta --no-kubernetes
😄  minikube v1.37.0 on Darwin 26.1 (arm64)
✨  Using the vfkit driver based on user configuration
👍  Starting minikube without Kubernetes in cluster minikube
🔥  Creating vfkit VM (CPUs=2, Memory=6144MB, Disk=20000MB) ...
🐳  Preparing Docker 28.5.2 ...
🏄  Done! minikube is ready without Kubernetes!

% minikube ssh /mnt/rosetta/test                                  
✅ Running amd64 binary on Linux minikube 6.6.95 #1 SMP PREEMPT Thu Dec 11 21:09:52 UTC 2025 aarch64 GNU/Linux
```

## Testing with container images

I tested the PR with ramen regional DR environment, using containerd runtime. We deploy several adm64 only container images like open-cluster-management and ramen.
https://github.com/RamenDR/ramen/pull/2356

---

[1] https://developer.apple.com/documentation/virtualization/running-intel-binaries-in-linux-vms-with-rosetta

Fixes #20559

### DIFF
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -147,6 +147,7 @@ const (
 	gpus                    = "gpus"
 	autoPauseInterval       = "auto-pause-interval"
 	preloadSrc              = "preload-source"
+	rosetta                 = "rosetta"
 )
 
 var (
@@ -287,6 +288,9 @@ func initDriverFlags() {
 
 	// qemu
 	startCmd.Flags().String(qemuFirmwarePath, "", "Path to the qemu firmware file. Defaults: For Linux, the default firmware location. For macOS, the brew installation location. For Windows, C:\\Program Files\\qemu\\share")
+
+	// vfkit
+	startCmd.Flags().Bool(rosetta, false, "Enable Rosetta to support apps built for Intel processor on a Mac with Apple silicon (vfkit driver only)")
 }
 
 // initNetworkingFlags inits the commandline flags for connectivity related flags for start
@@ -551,6 +555,24 @@ func validateVfkitNetwork(n string, options *run.CommandOptions) string {
 	return n
 }
 
+func getRosetta(driverName string) bool {
+	enabled := viper.GetBool(rosetta)
+	if enabled {
+		if !driver.IsVFKit(driverName) {
+			out.WarningT("--rosetta flag is only valid with the vfkit driver, it will be ignored")
+			return false
+		}
+		if runtime.GOARCH != "arm64" {
+			out.WarningT("--rosetta flag is only valid on Apple silicon, it will be ignored")
+			return false
+		}
+		if !viper.GetBool(flags.Interactive) {
+			out.Styled(style.Warning, "Skipping Rosetta automatic install in non-interactive mode")
+		}
+	}
+	return enabled
+}
+
 // generateNewConfigFromFlags generate a config.ClusterConfig based on flags
 func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime string, drvName string, options *run.CommandOptions) config.ClusterConfig {
 	var cc config.ClusterConfig
@@ -655,6 +677,7 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 		MultiNodeRequested: viper.GetInt(nodes) > 1 || viper.GetBool(ha),
 		GPUs:               viper.GetString(gpus),
 		AutoPauseInterval:  viper.GetDuration(autoPauseInterval),
+		Rosetta:            getRosetta(drvName),
 	}
 	cc.VerifyComponents = interpretWaitFlag(*cmd)
 
@@ -885,6 +908,7 @@ func updateExistingConfigFromFlags(cmd *cobra.Command, existing *config.ClusterC
 	updateStringFromFlag(cmd, &cc.SocketVMnetClientPath, socketVMnetClientPath)
 	updateStringFromFlag(cmd, &cc.SocketVMnetPath, socketVMnetPath)
 	updateDurationFromFlag(cmd, &cc.AutoPauseInterval, autoPauseInterval)
+	updateBoolFromFlag(cmd, &cc.Rosetta, rosetta)
 
 	if cmd.Flags().Changed(kubernetesVersion) {
 		kubeVer, err := getKubernetesVersion(existing)
--- a/pkg/drivers/vfkit/vfkit.go
+++ b/pkg/drivers/vfkit/vfkit.go
@@ -63,6 +63,10 @@ const (
 	logFileName    = "vfkit.log"
 	serialFileName = "serial.log"
 	defaultSSHUser = "docker"
+
+	// Rosetta mount in the guest.
+	rosettaMountTag   = "minikube-rosetta"
+	rosettaMountPoint = "/mnt/minikube-rosetta"
 )
 
 // Driver is the machine driver for vfkit (Virtualization.framework)
@@ -78,6 +82,7 @@ type Driver struct {
 	Network        string        // "", "nat", "vmnet-shared"
 	MACAddress     string        // For network=nat, network=""
 	VmnetHelper    *vmnet.Helper // For network=vmnet-shared
+	Rosetta        bool          // Enable rosetta support
 }
 
 func NewDriver(hostName, storePath string, options *run.CommandOptions) drivers.Driver {
@@ -259,6 +264,12 @@ func (d *Driver) Start() error {
 		return err
 	}
 
+	if d.Rosetta {
+		if err := d.setupRosetta(); err != nil {
+			return err
+		}
+	}
+
 	if len(d.VirtiofsMounts) > 0 {
 		log.Infof("Setup virtiofs mounts ...")
 		if err := virtiofs.SetupMounts(d, d.VirtiofsMounts); err != nil {
@@ -329,6 +340,10 @@ func (d *Driver) startVfkit(socketPath string) error {
 
 	}
 
+	if d.Rosetta {
+		startCmd = append(startCmd, d.rosettaOptions()...)
+	}
+
 	log.Debugf("executing: vfkit %s", strings.Join(startCmd, " "))
 	os.Remove(d.sockfilePath())
 	cmd := exec.Command("vfkit", startCmd...)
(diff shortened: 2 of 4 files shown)


nirs:
@cfergeau can you review?

cmd/minikube/cmd/start_flags.go
```diff
@@ -548,6 +552,24 @@ func validateVfkitNetwork(n string, options *run.CommandOptions) string {
 	return n
 }
 
+func getRosetta(driverName string) bool {
+	enabled := viper.GetBool(rosetta)
+	if enabled {
+		if !driver.IsVFKit(driverName) {
+			out.Styled(style.Warning, "Rosetta only supported on for the vfkit driver")
```
cfergeau: Keep only `on` or `for`, but not both (`on for`) in the warning message.

pkg/drivers/vfkit/vfkit.go
```diff
@@ -350,6 +365,57 @@ func (d *Driver) startVfkit(socketPath string) error {
...
 
+func (d *Driver) rosettaOptions() []string {
+	options := []string{
+		"rosetta",
+		"mountTag=" + rosettaMountTag,
+	}
+
+	// Try to install rosetta automatically for best user experience. The
+	// installation requires user interaction so we must skip it in
+	// non-interactive mode. If Rosetta is not instaled vfkit will fail.
```
cfergeau: `installed`

pkg/drivers/vfkit/vfkit.go
```diff
@@ -350,6 +365,57 @@ func (d *Driver) startVfkit(socketPath string) error {
...
+		// flags: F (fix binary), C (credentials), or O (open binary)
+		":F",
+	}, "")
+
+	var b strings.Builder
+
+	fmt.Fprintf(&b, "set -e\n")
+	fmt.Fprintf(&b, "sudo mount binfmt_misc -t binfmt_misc /proc/sys/fs/binfmt_misc\n")
+	fmt.Fprintf(&b, "sudo mkdir -p %s\n", rosettaMountPoint)
+	fmt.Fprintf(&b, "sudo mount -t virtiofs %s %s\n", rosettaMountTag, rosettaMountPoint)
```
cfergeau, line 410: This could use `virtiofs.SetupMounts`

pkg/drivers/vfkit/vfkit.go
```diff
@@ -350,6 +365,57 @@ func (d *Driver) startVfkit(socketPath string) error {
...
+	fmt.Fprintf(&b, "echo '%s' | sudo tee /proc/sys/fs/binfmt_misc/register\n", binfmt)
```
cfergeau, line 411: I don’t know what distro this is running on, but an alternative to writing to `/proc` directly is to use `systemd-binfmt` https://github.com/crc-org/vfkit/blob/main/doc/usage.md#rosetta
Either way is fine with me.

cmd/minikube/cmd/start_flags.go
nirs: Thanks, fixing

pkg/drivers/vfkit/vfkit.go
nirs: Yes, we can split it to:
- virtiofs.SetupMounts() for mounting the rosetta share
- run script to mount binfmt_misc and registry rosetta

But this is more complicated and harder to follow, and introduce dependency on the virtiofs package in a flow which does not really depend on this package. I think it will be easier to mainain like this.

pkg/drivers/vfkit/vfkit.go
nirs: I don't think that systemd-binmft.service will work for minikube use case. When the service starts very early, the rosetta mount is not available, and /etc/binfmt.d/rosetta.conf may not be available since minikube is immutable OS - any change that you make in runtime is lost when you restart the vm. So the only reliable way to register rosetta is to run this in the driver on every start.

Even if systemd-binfmt can work (with some tweaks), this service is not available in minikube. We can enable it in buildroot systemd configuration and it requires building a new iso with this feature.

The service seems like a good solution for standard server when you configure it once and systemd handles the next boots. For minikube use case when we have to run code during start, the systemd-binfmt service is not helpful and it is simpler to do this in minikube.

lima is using similar solution - a boot script is mounting and registering rosetta.

nirs:
Latest push rebased on master to split the rebase from the actual changes.

nirs:
@cfergeau I addressed the comments:
- Typos in warnings fixed
- Warnings change to match better similar warnings around this code
- I kept setupRosetta() as is for simplicity and ease of maintainability

nirs:
This KVM_Linux fail with the known issue when TestISO are run after the context deadline was exceeded:
[KVM_Linux](https://storage.googleapis.com/minikube-builds/logs/22140/42814/KVM_Linux.html)
KVM_Linux — Jenkins: completed with 10 / 447 failures in 65.73 minutes.

The interesting part is that this test did not time out after 90 minutes - so this is not an issue with the global test timeout. Maybe this an issue with the context we use in these tests:

```
=== RUN   TestISOImage/PersistentMounts//data
=== PAUSE TestISOImage/PersistentMounts//data
=== CONT  TestISOImage/PersistentMounts//data
iso_test.go:97: (dbg) Run:  out/minikube-linux-amd64 -p guest-625557 ssh "df -t ext4 /data | grep /data"
iso_test.go:97: (dbg) Non-zero exit: out/minikube-linux-amd64 -p guest-625557 ssh "df -t ext4 /data | grep /data": context deadline exceeded (3.039µs)
```

obnoxxx:
@nirs , I was trying yout patch locally, but I can't get `minikube start` with the commandline as in the PR description to complete successfully on my machine.

I am seeing this: 
```console

$ minikube delete --all --purge

$ minikube start ... --rosetta ..
...
🔄  Successfully unblocked bootpd process from firewall, retrying
🔥  Creating vfkit VM (CPUs=2, Memory=6144MB, Disk=20000MB) ...
😿  Failed to start vfkit VM. Running "minikube delete" may fix it: creating host: create host timed out in 360.000000 seconds

❌  Exiting due to DRV_CREATE_TIMEOUT: Failed to start host: creating host: create host timed out in 360.000000 seconds
💡  Suggestion: Try 'minikube delete', and disable any conflicting VPN or firewall software
🍿  Related issue: https://github.com/kubernetes/minikube/issues/7072

$ echo $?
52
$

```

nirs:
> @nirs , I was trying yout patch locally, but I can't get `minikube start` with the commandline as in the PR description to complete successfully on my machine.
> 
> I am seeing this:
> 
> ```
> $ minikube delete --all --purge
> 
> $ minikube start ... --rosetta ..
> ...
> 🔄  Successfully unblocked bootpd process from firewall, retrying
> 🔥  Creating vfkit VM (CPUs=2, Memory=6144MB, Disk=20000MB) ...
> 😿  Failed to start vfkit VM. Running "minikube delete" may fix it: creating host: create host timed out in 360.000000 seconds
> 
> ❌  Exiting due to DRV_CREATE_TIMEOUT: Failed to start host: creating host: create host timed out in 360.000000 seconds
> 💡  Suggestion: Try 'minikube delete', and disable any conflicting VPN or firewall software
> 🍿  Related issue: https://github.com/kubernetes/minikube/issues/7072
> ```

You hide the details, so I must guess from the partial output. You are using vment-helper on managed Mac, and did not get IP address. minikube wrongly try to "unblock" bootpd which does not have any affect and fail with the wrong error.

If you try without --rosetta you will probably get the same result.

One known issue is connecting to your company VPN. For me after connecting to the VPN and disconnecting, the vment network is always broken and the only way to recover is to reboot.

If you did not connect to the VPN since boot, this workaround that usually works:

```
minikube delete
sudo killall socketfilterfw bootpd
minikube start ...
```

It this does not work reboot usually fix the issue.

obnoxxx:
@nirs, with your hint, I got minikube start to complete now, and running the test binary in the vm succeeds:

```console
% ./out/minikube ssh
$ /mnt/rosetta/test
✅ Running amd64 binary on Linux minikube 6.6.95 #1 SMP PREEMPT Tue Dec 16 02:19:13 UTC 2025 aarch64 GNU/Linux
$ 
```

so this seems all fine.

cmd/minikube/cmd/start_flags.go
```diff
@@ -285,6 +286,9 @@ func initDriverFlags() {
 
 	// qemu
 	startCmd.Flags().String(qemuFirmwarePath, "", "Path to the qemu firmware file. Defaults: For Linux, the default firmware location. For macOS, the brew installation location. For Windows, C:\\Program Files\\qemu\\share")
+
+	// vfkit
+	startCmd.Flags().Bool(rosetta, false, "Enable Rosetta to support apps built for Intel processor on a Mac with Apple silicon (vfkit driver only)")
```
medyagh, line 293: does it also support running images built for x86 on all container runtimes ?
or is it only for binaries ? if also images, should mention in the help text.

it would be nice to check or verify (
suggestion, add an integration test that pulls an x86 image and runs it on silicon)
and test be skipped on anything but macos arm64
(could be a subset of functional test)

if dont wanna add ftest, can also just add manual verification and add to the PR description.

pkg/drivers/vfkit/vfkit.go
```diff
@@ -350,6 +365,57 @@ func (d *Driver) startVfkit(socketPath string) error {
 	return process.WritePidfile(d.pidfilePath(), cmd.Process.Pid)
 }
 
+func (d *Driver) rosettaOptions() []string {
```
medyagh, line 369: could add a short general comment for easy code skimming

cmd/minikube/cmd/start_flags.go
nirs: > does it also support running images built for x86 on all container runtimes ? or is it only for binaries ? if also images, should mention in the help text.

It works for running any executable, on the host or in a container. I'm using the same language used by apple to describe the feature (apps built for Intel processors). Should we replace "apps" with container images since minikube is about containers?
 
> it would be nice to check or verify

I tested running executable in the guest (example in the message), and running ramen testing clusters (containerd runtime) which include several amd64 images (ocm, ramen).
The ramen PR https://github.com/RamenDR/ramen/pull/2356 shows example run with this PR.

> ( suggestion, add an integration test that pulls an x86 image and runs it on silicon) and test be skipped on anything but macos arm64 (could be a subset of functional test)
> 
> if dont wanna add ftest, can also just add manual verification and add to the PR description.

I want to add a test that will do this:
- created executable like the example as part of the test
- run it using the virtiofs mount
- create a container image including the executable by building inside minikube, or in podman
- load the image into minikube
- start a pod with the image

I'll open an issue for working on the test, since it more work than the actual feature.

Issue #22056, opened 2025-12-07
iso: Support build on Fedora - aarch64


nirs:
Since we removed gluster, we don't need the legacy python2 dependency and we can support building on more modern distros like Fedora.

Building the aarch64 version on Fedora 42 and 43 has 2 issues:

## selinux issue when running chown via a container

Error:

```console
if ! grep -sq 'docker\|lxc' /proc/1/cgroup; then \                                                                                                                                                         
    docker run --rm -v /home/nsoffer/minikube/out/buildroot/output-aarch64/build/cri-dockerd-aarch64-55d6e1a1d6f2ee58949e13a0c66afe7d779ac942/packaging/static/build/win:/v -w /v alpine sh -c 'apk update 
&& apk add zip && zip -r cri-dockerd-0.4.1.win.amd64.zip cri-dockerd'; \                                                                                                                                   
    docker run --rm -v /home/nsoffer/minikube/out/buildroot/output-aarch64/build/cri-dockerd-aarch64-55d6e1a1d6f2ee58949e13a0c66afe7d779ac942/packaging/static:/v -w /v alpine chown -R 1000:1000 build; \ 
fi                                                                                                                                                                                                         
Unable to find image 'alpine:latest' locally                                                                                                                                                               
latest: Pulling from library/alpine                                                                                                                                                                        
014e56e61396: Pull complete                                                                                                                                                                                
Digest: sha256:51183f2cfa6320055da30872f211093f9ff1d3cf06f39a0bdb212314c5dc7375                                                                                                                            
Status: Downloaded newer image for alpine:latest                                                                                                                                                           
v3.23.0_rc2-34-g319b82182d1 [https://dl-cdn.alpinelinux.org/alpine/v3.23/main]                                                                                                                             
v3.23.0-1-gbad45407d26 [https://dl-cdn.alpinelinux.org/alpine/v3.23/community]                                                                                                                             
OK: 27555 distinct packages available                                                                                                                                                                      
(1/2) Installing unzip (6.0-r16)                                                                                                                                                                           
(2/2) Installing zip (3.0-r13)                                                                                                                                                                             
Executing busybox-1.37.0-r29.trigger                                                                                                                                                                       
OK: 8 MiB in 18 packages                                                                                                                                                                                   
        zip warning: name not matched: cri-dockerd                                                                                                                                                         
                                                                                                                                                                                                           
[...]

PR #22158, opened 2025-12-15
iso: Stop using packaging scripts for cri-dockerd

### LINKED ISSUE #22146: Remake the packaging of cri-dockerd in the ISO buildroot
Currently using some upstream "packaging" scripts with docker and whatnot.

It should use the standard buildroot mechanism, and just build a golang package...

Update the documentation, after fixing the package:

~~Some external projects will try to use docker even when building without docker.~~
~~You must install docker on the build host.~~

*  #20830

afbjorklund:
The upstream packaging scripts do a lot of _insane_ things like start a docker container to run chown and use strange variables.

Also restore the package version to normal, now that it is released. Should change the tarballs and hash for next time, with refactoring.

Fixes #22146

_Eventually this should just use golang-package.mk and then it would be really really small. But that can come later...._

### DIFF
--- a/pkg/minikube/download/iso.go
+++ b/pkg/minikube/download/iso.go
@@ -43,7 +43,6 @@ func DefaultISOURLs() []string {
 	v := version.GetISOVersion()
 	isoBucket := "minikube-builds/iso/22158"
 
-
 	return []string{
 		fmt.Sprintf("https://storage.googleapis.com/%s/minikube-%s-%s.iso", isoBucket, v, runtime.GOARCH),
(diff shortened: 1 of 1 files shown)


afbjorklund:
ok-to-build-iso

nirs:
Tested on top of #21997 on Fedora 43 aarch64 vm: 
- minikube-iso-x86+64:
  - previously the build failed since it was built as arm64
  - iso created!
- minikube-iso-aarch64:
  - previously failing because of running docker without wrong --volume arguments. Now we don't run docker so the issue is fixed.
  - fails in post-image script (#22056)

nirs:
Tests looks good. I'll test this iso locally with vfkit later. Since the build works, it is unlikely to be broken on macOS.

/cc @medyagh

Issue #22144, opened 2025-12-14
Upgrade the build image from ubuntu to debian for ISO buildroot


afbjorklund:
ubuntu 20.04:
https://github.com/kubernetes/minikube/blob/master/deploy/iso/minikube-iso/Dockerfile

debian 12:
https://github.com/buildroot/buildroot/blob/2025.02.x/support/docker/Dockerfile

----

This should also make it multi-arch, and the documentation can updated accordingly.

~~Machine with x86_64 CPU~~
~~Ubuntu 22.04.5 LTS (Jammy Jellyfish)~~

* https://github.com/kubernetes/minikube/pull/20830

afbjorklund:
Fixed the old image, meanwhile:

* https://github.com/kubernetes/minikube/pull/22164

Might want to upgrade it to `ubuntu:22.04`, as well

* https://github.com/kubernetes/minikube/pull/22187

---

There are some funky bugs in `sed` that breaks build on virtiofs

`/usr/bin/sed: couldn't open temporary file /mnt/out/buildroot/output-aarch64/host/bin/sed4hn88z: Permission denied`

```
* Noteworthy changes in release 4.8 (2020-01-14) [stable]

** Bug fixes

  "sed -i" now creates temporary files with correct umask (limited to u=rwx).
  Previously sed would incorrectly set umask on temporary files, resulting
  in problems under certain fuse-like file systems.
  [bug introduced in sed 4.2.1]

```

Unfortunately it still has issues with symlinks, so virtiofs is a no-go...

`/mnt/out/buildroot/output-aarch64/host/bin/tar: b3sum/LICENSE_A2: Cannot open: Permission denied`

* #22166

afbjorklund:
The new buildroot also needs a newer version of tar, version 1.35 or up

commit https://github.com/buildroot/buildroot/commit/b11956fb6690e0083eab9f30b5dc146fc40c2d5f

Both Debian 12 and Ubuntu 22.04 unfortunately have version 1.34 (close!)

Currently buildroot builds it from source:

```Dockerfile
# We need tar >= 1.35
ARG TAR_VERSION="1.35"

# Build host-tar
RUN curl -sfL https://ftpmirror.gnu.org/tar/tar-${TAR_VERSION}.tar.xz | \
    tar -Jx -C /tmp && \
    cd /tmp/tar-${TAR_VERSION} && \
    FORCE_UNSAFE_CONFIGURE=1 ./configure \
        --disable-year2038 && \
[...]

PR #22164, opened 2025-12-16
ci: iso image Only install gcc-multilib on x86_64 arch

afbjorklund:
Use regular gcc instead on aarch64 arch

It doesn't need the i386 compatibility.

----

`make buildroot-image`

https://github.com/buildroot/buildroot/commit/3bc94471ca8a90f331366d9593630952b0e7bb10

### DIFF
--- a/deploy/iso/minikube-iso/Dockerfile
+++ b/deploy/iso/minikube-iso/Dockerfile
@@ -18,6 +18,7 @@ RUN apt-get update \
 	&& apt-get install -y apt dpkg apt-utils ca-certificates software-properties-common \
 	&& add-apt-repository -y ppa:longsleep/golang-backports \
 	&& apt-get upgrade -y \
+	&& if [ "$(uname -ms)" = "Linux x86_64" ]; then apt-get install -y gcc-multilib; fi \
 	&& apt-get install -y \
 		build-essential \
 		git \
@@ -26,7 +27,7 @@ RUN apt-get update \
 		python \
 		unzip \
 		bc \
-		gcc-multilib \
+		gcc \
 		automake \
 		libtool \
 		gnupg2 \


nirs:
Why do we need x86 32 bit support in 2026? Who will run minikube on ancient 32 bit systems?

afbjorklund:
> Why do we need x86 32 bit support in 2026? Who will run minikube on ancient 32 bit systems?

We do not, but buildroot still supports 32-bit targets as far as I know (or they just haven't updated the builder perhaps)

https://github.com/buildroot/buildroot/blob/master/support/docker/Dockerfile

Note that it is not about 32-bit OS

nirs (approved):
Can we use buildroot Dockerfile as base?

medyagh (commented):
@afbjorklund 

I think this image is not used to build iso ( I think the iso is built on bare-linux for faster speed)

jenkins calls build_iso.sh script
https://github.com/kubernetes/minikube/blob/21d5a9e6516426cd37ee5590ba21e77747d5ef86/hack/jenkins/build_iso.sh#L68C6-L68C22


however, I created an issue so we add a automation to move this to K8s.registry too anyways
https://github.com/kubernetes/minikube/issues/22167

but I dont think this is used ?

afbjorklund:
> I dont think this is used ?

But it should be used, since otherwise you need to fix all these host dependencies that is breaking the build.

afbjorklund:
> Can we use buildroot Dockerfile as base?

That is the long-term plan, but it is similar to the situation with kind - we forked the Dockerfile and made it our own...

* https://github.com/kubernetes/minikube/issues/22144

But it should at least be synced a bit.

medyagh:
> > Can we use buildroot Dockerfile as base?
> 
> That is the long-term plan, but it is similar to the situation with kind - we forked the Dockerfile and made it our own...

fair enough, 

currently we have dedicated machines in jenkins (limitted number) but yes we will need to know the receipie we need to recreate those once moving to minikube-os

do you mind building the image locally or pushing it github container registry (if needed can be pushed to minikube-os repo packages) so we know this image builds at all

afbjorklund:
The image builds now, but _running_ it is another story... Pushing and pulling the image would need more details?

New location needs to go in the documentation and have bandwidth and so, so should be planned more carefully.

medyagh:
> The image builds now, but _running_ it is another story... Pushing and pulling the image would need more details?
> 
> New location needs to go in the documentation and have bandwidth and so, so should be planned more carefully.

we can add them to be built in prow
I am purposing making a Branch for minikube called "images" that each time we push there it should trigger

https://github.com/kubernetes/minikube/issues/22172

Staging Image build in Prow K8s.io registry
and then we can promote that image


right now it runs on master all the time.

PR #22179, opened 2025-12-17
drivers: Introduce WaitForSSHAccess and fix macos smoke tests

### LINKED ISSUE #21091: cleanup: WaitForTCPWithDelay() duplicated in multiple drivers
### What Happened?

This function is duplicated in:
- qemu: https://github.com/kubernetes/minikube/blob/b973335449467dc12e23956be18336c42ff6b3ce/pkg/drivers/qemu/qemu.go#L869
- vfkit: https://github.com/kubernetes/minikube/blob/b973335449467dc12e23956be18336c42ff6b3ce/pkg/drivers/vfkit/vfkit.go#L591
- krunkit: https://github.com/nirs/minikube/blob/83299339bbba3449e6f92243266cc5a15a590fb5/pkg/drivers/krunkit/krunkit.go#L548

Should move to common location and reused.

### LINKED ISSUE #22151: Improve WaitForTCPWithDelay
We have this in vfkit and (identical?) copies in krunkit and qemu.

```go
func WaitForTCPWithDelay(addr string, duration time.Duration) error {
	for {
		conn, err := net.Dial("tcp", addr)
		if err != nil {
			continue
		}
		defer conn.Close()
		if _, err := conn.Read(make([]byte, 1)); err != nil && err != io.EOF {
			time.Sleep(duration)
			continue
		}
		break
	}
	return nil
}
```

Problems:
- If connect fails, we are try again without a delay - busy loop
- We wait for the server identification string without sending our identification string. In a typical SSH session the client talk first. A server may never send the identification string since we don't send anything
- No logging, impossible to debug issues
- No timeout - we can be in busy dial loop for 6 minutes until minikube timeout expires.

For reference, here is how ssh client session looks like:

```console
$ ssh -v localhost
...
debug1: Connecting to localhost port 22.
debug1: Connection established.
debug1: Local version string SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.6 
debug1: Remote protocol version 2.0, remote software version OpenSSH_8.9p1 Ubuntu-3ubuntu0.6
```

We need similar debug logs:
- connecting to <address>
- sending local version
- received remote version: OpenSSH_8.9p1 Ubuntu-3ubuntu0.6

/kind improvement

nirs:
This change fixes the smoke tests on macOS, makes waiting for SSH more robust, and makes the workflow easier to debug.

Changes:
- Run minikube as root on macOS runners - this is the actual fix, escaping local network privacy limits introduced in macOS 15.
- Disable spotlight indexing eliminating high load right before we start minikube
- Make the workflow easier to debug, removing unnecessary skip when the runner load is high
- Deduplicate code wait for SSH access, extracting `common.WaitForSSHAccess()`
- Improve logging and error handling when waiting for SSH access
- Limit the time waiting for SSH access

## Before

Screenshot 2025-12-20 at 22 49 13

## After

Screenshot 2025-12-20 at 22 45 28

---

Fixes #22151
Fixes #21091

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -44,19 +44,23 @@ jobs:
           - name: qemu-docker-macos-15-x86
             driver: qemu
             os: macos-15-intel
-            network_flag: --network socket_vmnet
+            start_flags: --network socket_vmnet --force
+            sudo: "sudo"
           - name: vfkit-docker-macos-15-x86
             driver: vfkit
             os: macos-15-intel
-            network_flag: --network vmnet-shared
+            start_flags: --network vmnet-shared --force
+            sudo: "sudo"
           - name: docker-docker-ubuntu-24.04-x86
             driver: docker
             os: ubuntu-24.04
-            network_flag: ""
+            start_flags: ""
+            sudo: ""
           - name: docker-docker-ubuntu-24.04-arm
             driver: docker
             os: ubuntu-24.04-arm
-            network_flag: ""
+            start_flags: ""
+            sudo: ""
     steps:
       - id: info-block
         uses: medyagh/info-block@main
@@ -105,13 +109,13 @@ jobs:
           sudo pfctl -s all || true
       - name: Start minikube (1st boot)
         run: |
-          ./out/minikube start \
+          ${{ matrix.sudo }} ./out/minikube start \
+              ${{ env.LOG_ARGS }} \
               --no-kubernetes \
               --cpus 2 \
               --memory 3gb \
               --driver ${{ matrix.driver }} \
-              ${{ matrix.network_flag }} \
-              ${{ env.LOG_ARGS }}
+              ${{ matrix.start_flags }}
       - name: Inspect host after test (macos)
         if: always() && contains(matrix.os, 'macos')
         run: |
@@ -126,17 +130,17 @@ jobs:
       - name: Inspect minikube
         if: always()
         run: |
-          tree -h ~/.minikube
+          ${{ matrix.sudo }} tree -h ~/.minikube
           machine="$HOME/.minikube/machines/minikube"
-          machine_logs=$(find "$machine" -name "*.log")
+          machine_logs=$(${{ matrix.sudo }} find "$machine" -name "*.log")
           minikube_logs="$HOME/.minikube/logs/lastStart.txt"
           for f in $machine_logs $minikube_logs /var/db/dhcpd_leases; do
             echo "==> $f <=="
-            head -n 1000 "$f" || true
+            ${{ matrix.sudo }} head -n 1000 "$f" || true
           done
       - name: Stop minikube
-        run: ./out/minikube stop ${{ env.LOG_ARGS }}
+        run: ${{ matrix.sudo }} ./out/minikube stop ${{ env.LOG_ARGS }}
       - name: Start minikube again (2nd boot)
-        run: ./out/minikube start ${{ env.LOG_ARGS }}
+        run: ${{ matrix.sudo }} ./out/minikube start ${{ env.LOG_ARGS }} ${{ matrix.start_flags }}
       - name: Delete minikube
-        run: ./out/minikube delete ${{ env.LOG_ARGS }}
+        run: ${{ matrix.sudo }} ./out/minikube delete ${{ env.LOG_ARGS }}


.github/workflows/smoke-test.yml
```diff
@@ -93,14 +81,20 @@ jobs:
         run: |
           brew install qemu socket_vmnet
           HOMEBREW=$(which brew) && sudo ${HOMEBREW} services start socket_vmnet
+      - name: Info (before test)
+        uses: medyagh/info-block@main
```
nirs: This and the next one should run only the dynamic checks (top, uptime, ps)

medyagh: it takes 1 seconds to run why matters not to include other info anyways?

medyagh: once this feature is added you can do it https://github.com/medyagh/info-block/issues/15

nirs: replaced with simple top for now.

nirs: That break the linux jobs, but it was very useful to fix indexing.

.github/workflows/smoke-test.yml
```diff
@@ -54,20 +54,8 @@ jobs:
             os: ubuntu-24.04-arm
             network_flag: ""
     steps:
-      - id: info-block
+      - name: Info block
```
nirs: Keep the id, add name for nicer display

.github/workflows/smoke-test.yml
```diff
@@ -54,20 +54,8 @@ jobs:
...
-          echo "memory_gb=${{ steps.info-block.outputs.memory_gb }}"
-          echo "cpu_cores=${{ steps.info-block.outputs.cpu_cores }}"
-          echo "load_average=${{ steps.info-block.outputs.load_average }}"
-          echo "free_mem=${{ steps.info-block.outputs.free_mem }}"
-          load_avg=${{ steps.info-block.outputs.load_average }}
-          cores=${{ steps.info-block.outputs.cpu_cores }}
-          if (( $(echo "$load_avg / $cores > 2" | bc -l) )); then
-            echo "Load average per core is above 2; stopping job early."
-            exit 1
-          fi
```
nirs, line 70: This is not helpful but the change is not required for fixing the test.

.github/workflows/smoke-test.yml
```diff
@@ -93,14 +81,22 @@ jobs:
         run: |
           brew install qemu socket_vmnet
           HOMEBREW=$(which brew) && sudo ${HOMEBREW} services start socket_vmnet
+      - name: Top (before test)
+        # 60 seconds cool down before starting the test.
+        run: top -l 30 -s 2 -n 20 -stats pid,command,cpu,time,mem
```
nirs: The cooldown was useful for debugging but we don't need it now.

.github/workflows/smoke-test.yml
```diff
@@ -93,14 +81,22 @@ jobs:
...
       - name: Start minikube (1st boot)
         run: |
           ./out/minikube start \
               --no-kubernetes \
-              --memory 4gb \
+              --cpus 1 \
```
nirs: Need to test with 2 cpus later, the issue was not the number of cpus.

.github/workflows/smoke-test.yml
```diff
@@ -93,14 +81,22 @@ jobs:
...
+              --memory 2600m \
```
nirs: 2600m is too close the minimum (2400m). 4g is too much since we don't have a lot of memory on these runners, about 5-6g unused memory. We can use the standard 3g used in other tests.

.github/workflows/smoke-test.yml
```diff
@@ -93,14 +81,22 @@ jobs:
...
               --driver ${{ matrix.driver }} \
               ${{ matrix.network_flag }} \
               ${{ env.LOG_ARGS }}
+      - name: Top (after test)
+        if: always()
+        # We need at leat 2 samples to get meanigful cpu info.
+        run: top -l 5 -s 2 -n 20 -stats pid,command,cpu,time,mem
```
nirs: No need to run 5 samples after the test.

.github/workflows/smoke-test.yml
```diff
@@ -112,9 +108,3 @@ jobs:
             echo "==> $f <=="
             head -n 1000 "$f" || true
           done
-      - name: Stop minikube
-        run: ./out/minikube stop ${{ env.LOG_ARGS }}
-      - name: Start minikube again (2nd boot)
-        run: ./out/minikube start ${{ env.LOG_ARGS }}
-      - name: Delete minikube
-        run: ./out/minikube delete ${{ env.LOG_ARGS }}
```
nirs, line 120: This should work now (with the new sudo and start_flags.

.github/workflows/smoke-test.yml
```diff
@@ -56,6 +56,13 @@ jobs:
     steps:
       - name: Info block
         uses: medyagh/info-block@main
+      - name: Disable spotlight indexing (macos)
+        if: contains(matrix.os, 'macos')
+        run: |
+          # Disable indexing globally
+          sudo mdutil -a -i off
+          # Erase Existing Indexes
+          sudo mdutil -E /
```
nirs, line 73: This is not required to fix the test but a good change, it keeps the load minimal before test.

.github/workflows/smoke-test.yml
```diff
@@ -63,6 +63,11 @@ jobs:
           sudo mdutil -a -i off
           # Erase Existing Indexes
           sudo mdutil -E /
+      - name: Disable macOS firewall (macos)
+        if: contains(matrix.os, 'macos')
+        run: |
+          sudo /usr/libexec/ApplicationFirewall/socketfilterfw --setglobalstate off
+          sudo /usr/libexec/ApplicationFirewall/socketfilterfw --setstealthmode off
```
nirs: This should not be needed. It is better to test with the firewall so the test can detect issues that real user can have with the firewall.

.github/workflows/smoke-test.yml
```diff
@@ -43,16 +43,16 @@ jobs:
         include:
           - driver: qemu
             os: macos-15-intel
-            network_flag: --network socket_vmnet
+            start_flags: --network socket_vmnet --native-ssh=false
```
nirs: Does not work with current code - the flag is not respected in provisioning code.

Change to start_flags is good since we have more flags to add in later commit.

pkg/drivers/common/ssh.go
```diff
@@ -34,43 +31,31 @@ var (
...
-
-	log.Infof("Waiting until SSH server %q is accessible", addr)
+	log.Infof("Waiting until SSH server is accessible")
 	start := time.Now()
 
-	for {
-		log.Debugf("Dialing to %q", addr)
-		conn, err := net.Dial("tcp", addr)
+	maxAttempts := 60
+	for i := 1; i <= maxAttempts+1; i++ {
```
nirs: Need to split the limit on number of retries from switch to ssh.

pkg/minikube/command/ssh_runner.go
```diff
@@ -107,6 +107,7 @@ func (s *SSHRunner) client() (*ssh.Client, error) {
 		return s.c, nil
 	}
 
+	// XXX Should respect --native-ssh=false, does not work in github macos amd64 runner
```
nirs: Open new issue for this and link to it here.

.github/workflows/smoke-test.yml
```diff
@@ -43,31 +43,30 @@ jobs:
...
+            start_flags: ""
+            sudo: ""
           - driver: docker
             os: ubuntu-24.04-arm
-            network_flag: ""
+            start_flags: ""
+            sudo: ""
     steps:
-      - id: info-block
+      - name: Runner info
```
medyagh: changing the block name is not related to this PR
 ! the author perferd this name and I agree with him :) please change this or if you want we can create issue for dicuss in a separete issue/PR

nirs: The author did not add a `name:` - but I can remove it and post this in a separate PR.

pkg/drivers/common/ssh.go
```diff
@@ -0,0 +1,101 @@
...
+		return err
+	}
+	addr := net.JoinHostPort(ip, strconv.Itoa(port))
+	log.Infof("Waiting until SSH server %q is accessible", addr)
+
+	start := time.Now()
+	deadline := start.Add(timeout)
+	dialer := net.Dialer{Deadline: deadline}
+
+	for {
```
medyagh, line 63: should we use the retry package here while we are refactoring it ?

https://github.com/kubernetes/minikube/blob/f6e9269466b163855eb96770d8773b7051bcafa5/pkg/util/retry/retry.go#L101

nirs: We need a deadline on the dialer and connection. How the retry package can handle this?

medyagh: could be its own dedicated PR to refactor to use retry.package

.github/workflows/smoke-test.yml
```diff
@@ -43,31 +43,30 @@ jobs:
         include:
           - driver: qemu
             os: macos-15-intel
-            network_flag: --network socket_vmnet
+            start_flags: --network socket_vmnet --force
+            sudo: "sudo"
```
medyagh, line 48: add a Job Name here
so it doesnt become too long that wont fit How long it took in the PR Check Status in one look (see bellow we can see how long it took for docker but not macos smoke tests) becasue it adds --force --sudo

Smoke Test / Boot (docker, ubuntu-24.04-arm) (pull_request)Successful in 1m

Smoke Test / Boot (docker, ubuntu-24.04) (pull_request)Successful in 1m

Smoke Test / Boot (qemu, macos-15-intel, --network socket_vmnet --force, sudo) (pull_request)Successful in 11m

Smoke Test / Boot (vfkit, macos-15-intel, --network vmnet-shared --force, sudo) (pull_request)Successful i

nirs:
> ## Pull request overview
> This PR refactors SSH access waiting logic across multiple VM drivers and improves macOS smoke test reliability. The key change is running minikube as root on macOS runners to bypass local network privacy restrictions introduced in macOS 15.

@Copilot this is NOT a refactoring. Refactoring is improving the design without changing the behavior. This PR changes the behavior - add timeout, fix busy dial loop, add logs and more. Do not use the "refactoring" again when for such changes.

pkg/drivers/common/ssh.go
```diff
@@ -0,0 +1,101 @@
...
+			log.Debugf("Failed to read: %v", err)
+			conn.Close()
+			time.Sleep(retryDelay)
+			continue
+		}
+
+		log.Infof("SSH server %q is accessible in %.3f seconds", addr, time.Since(start).Seconds())
+		return nil
+	}
+}
```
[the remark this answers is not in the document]
nirs: This code is covered by the functional and integration tests. Testing timeouts and other errors requires intrusive mocking which couples the test to the implementation, and make the tests harmful instead of helpful.

PR #22181, opened 2025-12-17
lint: gocritic

medyagh:
fixing gocritic lint warnnings at once

```
	         ^
cmd/minikube/cmd/start.go:796:3: ifElseChain: rewrite if-else to switch statement (gocritic)
		if foundStoppedDocker {
		^
cmd/minikube/cmd/start.go:1242:3: ifElseChain: rewrite if-else to switch statement (gocritic)
		if drvName == oci.Docker && runtime.GOOS == "darwin" {
		^
cmd/minikube/cmd/start.go:1562:2: ifElseChain: rewrite if-else to switch statement (gocritic)
	if memString == constants.NoLimit && driver.IsKIC(drvName) {
	^
cmd/minikube/cmd/start_flags.go:395:3: ifElseChain: rewrite if-else to switch statement (gocritic)
		if memString == constants.NoLimit && driver.IsKIC(drvName) {
		^
cmd/minikube/main.go:136:2: ifElseChain: rewrite if-else to switch statement (gocritic)
	if machineLogEnvironmentRe.Match(b) {
	^
pkg/minikube/tunnel/route.go:80:3: ifElseChain: rewrite if-else to switch statement (gocritic)
		if route.Equal(tableLine.route) {
		^
pkg/minikube/tunnel/route_linux.go:97:4: ifElseChain: rewrite if-else to switch statement (gocritic)
			if err != nil {
			^
pkg/addons/helm.go:76:24: helmInstallBinary - addon is unused (unparam)
func helmInstallBinary(addon *assets.Addon, runner command.Runner) error {
                       ^
pkg/minikube/storageclass/storageclass_test.go:254:64: setK8SConfig - result 0 (error) is always nil (unparam)
func setK8SConfig(t *testing.T, config, kubeconfigPath string) error {
```

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -451,7 +451,7 @@ func enableOrDisableAddonInternal(cc *config.ClusterConfig, addon *assets.Addon,
 	}
 
 	if addon.HelmChart != nil {
-		err := helmInstallBinary(runner)
+		err := helmInstallBinary(addon, runner)
 		if err != nil {
 			return err
 		}
--- a/pkg/addons/helm.go
+++ b/pkg/addons/helm.go
@@ -73,7 +73,7 @@ func helmUninstallOrInstall(ctx context.Context, chart *assets.HelmChart, enable
 	return uninstalllHelmChart(ctx, chart)
 }
 
-func helmInstallBinary(runner command.Runner) error {
+func helmInstallBinary(addon *assets.Addon, runner command.Runner) error {
 	_, err := runner.RunCmd(exec.Command("test", "-f", "/usr/bin/helm"))
 	if err != nil {
 		_, err = runner.RunCmd(exec.Command("test", "-d", "/usr/local/bin"))


cmd/minikube/cmd/start.go
```diff
@@ -1238,12 +1239,14 @@ func validateCPUCount(drvName string) {
...
-	if availableCPUs < 2 {
-		if drvName == oci.Docker && runtime.GOOS == "darwin" {
+	switch {
+	case availableCPUs < 2:
+		switch {
+		case drvName == oci.Docker && runtime.GOOS == "darwin":
 			exitIfNotForced(reason.RsrcInsufficientDarwinDockerCores, "Docker Desktop has less than 2 CPUs configured, but Kubernetes requires at least 2 to be available")
-		} else if drvName == oci.Docker && runtime.GOOS == "windows" {
+		case drvName == oci.Docker && runtime.GOOS == "windows":
 			exitIfNotForced(reason.RsrcInsufficientWindowsDockerCores, "Docker Desktop has less than 2 CPUs configured, but Kubernetes requires at least 2 to be available")
```
nirs, line 1248: Why do we need 2 different exit code and messages for docker on mac or windows?

Also not clear why we need special case for docker desktop.   We spend too much code and effort on this warning. Using one case for all is better.

medyagh: my gut feeling says it is about sudo or root permissions, on windows it would popup ask for permission maybe... not sure but this PR is meant to not change or improve anything but just fix the lint

medyagh: @nirs 


You are right that the error message text passed in that block is identical for both Mac and Windows: "Docker Desktop has less than 2 CPUs configured..."

However, the first argument (reason.RsrcInsufficientDarwinDockerCores vs reason.RsrcInsufficientWindowsDockerCores) is what matters.

The 
exitIfNotForced
 function passes that reason object to Minikube's exit handler, which uses it to print distinct Advice (instructions) to the user.

Why the if/else is needed:
Minikube wants to show the user exactly how to fix the problem. The instructions are different for Mac vs Windows:

If Mac (darwin): pass RsrcInsufficientDarwinDockerCores
Advice shown: Click on "Docker for Desktop" menu icon -> "Preferences"
If Windows: pass RsrcInsufficientWindowsDockerCores
Advice shown: Open the "Docker Desktop" menu -> "Settings"
If we didn't have this if/else block, we would have to show generic instructions (e.g. "Check your Docker settings") instead of the helpful step-by-step guide.

Here is the underlying definition from

PR #22207, opened 2025-12-18
feat: add --image-repository support for Helm-based addons

### LINKED ISSUE #21257: provide "helm"  based addons
we should have addons that delegate installation of their yaml files to helm, instead of embedding all their yaml files into minikube binary,

minikube addons could should handle Go Part of the Code 
for example, gcp-auth addon, looks up credentials on users host and handles the creds inside minikube code

but majority of other addons they only deploy a yaml file, that can be delegated to helm, and we can simply bump the version of pinned version when we wanna bump an addon that works for minikube.

only down side this that I might see, is our templatization uses Mirror Image Repositories for users that pass their mirror image, we should see if we can pass that mirror image repo to helm

PhantomInTheWire:
#### Summary

Adds `--image-repository` flag to `minikube addons enable` for Helm-based addons. Supports format `registry/org[:tag]` to override container images.

#### Usage

```bash
minikube addons enable volcano --image-repository=ghcr.io/volcano-sh:latest
minikube addons enable volcano # or use default
```

#### Changes
- New --image-repository flag that parses registry, org, and tag
- Added TagSetKey and ImageNameKeys to  HelmChart struct
- Updated volcano addon to use Helm (removed yaml templates)
- Unit tests for parsing and command generation

Fixes #21257


medyagh:
@PhantomInTheWire the issue you linked talks about doing one exmaple of helm based addon, we have not added one yet

PhantomInTheWire:
@medyagh i initially wanted to migrate an existing addon that but but figured adding this first would be needed for a full migration. i was under the impression that we want to migrate existing addons, is that not the case? i thought not being able to use --image-repository flag was the blocker, if not that what exactly would we need to cloase #21257

PhantomInTheWire:
@medyagh thanks for the feedback earlier!

Just wanted to follow up and clarify intent: this PR only adds the plumbing needed for `--image-repository` to work with Helm-based addons, so that existing addons can be migrated incrementally afterward.

If the expectation is that this PR should also migrate at least one addon as an example, I’m happy to do that, just wanted to confirm the preferred direction before proceeding.

Appreciate your guidance.

medyagh:
I undrstood the intent of the PR I just meant that we currently dont have any Addon that uses helm, you could try with adding a helm-based addon, such as converting "Volcano" addon to be helm based and test your PR with ti
it would be easier to review

pkg/minikube/assets/addons.go
```diff
@@ -36,11 +36,13 @@ import (
...
-	Namespace  string
-	Values     []string
-	ValueFiles []string
+	Name        string
+	Repo        string
+	RepoURL     string
+	Namespace   string
+	Values      []string
+	ValueFiles  []string
+	ImageSetKey string
```
medyagh, line 56: @PhantomInTheWire 
why is this called ImageSetKey is that same Name used in helm flags? can you please put an example of using helm itself with image respostiry 

also add comment to this field

medyagh (changes requested):
@PhantomInTheWire can you plz test this addon by moving volano addon to 

https://github.com/volcano-sh/volcano to helm charts and then try the --image-repostiry flag with it

volcano is now on ghcr.io
https://github.com/volcano-sh/volcano/pull/4846

pkg/minikube/assets/addons.go
PhantomInTheWire: helm doesn't actually give us a clean way to do this and we have to use --set to manually set fields, i have added comments to explain this data structure better, also i was initially testing  by migrating metrics server but working on volcano(on both dockerio and ghcrio) revealed that there were some bad assumptions i made regarding versions and how different registeries handle names, i have changed the pr up a little bit PTAL again

PhantomInTheWire:
@medyagh sure i have tested and pushed the migration for volcano

with image repository it pulls from where we ask
```bash
./out/minikube addons enable volcano --image-repository=ghcr.io/volcano-sh:latest
kubectl get pods -n volcano-system -o jsonpath='{range .items[*]}{.spec.containers[*].image}{"\n"}{end}'
# Expected: ghcr.io/volcano-sh/vc-*:latest
```

by default it pulls in v1.13 from docker.io that was in the codebase earlier(cant pull this exact version from ghcr it seems it only has the latest version)
```bash
./out/minikube addons enable volcano
kubectl get pods -n volcano-system -o jsonpath='{range .items[*]}{.spec.containers[*].image}{"\n"}{end}'
# Expected: docker.io/volcanosh/vc-*:v1.13.1
```

medyagh:
> @medyagh sure i have tested and pushed the migration for volcano
> 
> with image repository it pulls from where we ask
> 
> ```shell
> ./out/minikube addons enable volcano --image-repository=ghcr.io/volcano-sh:latest
> kubectl get pods -n volcano-system -o jsonpath='{range .items[*]}{.spec.containers[*].image}{"\n"}{end}'
> # Expected: ghcr.io/volcano-sh/vc-*:latest
> ```
> 
> by default it pulls in v1.13 from docker.io that was in the codebase earlier(cant pull this exact version from ghcr it seems it only has the latest version)
> 
> ```shell
> ./out/minikube addons enable volcano
> kubectl get pods -n volcano-system -o jsonpath='{range .items[*]}{.spec.containers[*].image}{"\n"}{end}'
> # Expected: docker.io/volcanosh/vc-*:v1.13.1
> ```

thank you so much @PhantomInTheWire for the update, I am curious is "latest" needed there? what happens if user doesnt specificy ghcr.io/volcano-sh:latest


also I am curious, instead of creating a new flag, why not use the existing one

is there a good reason not to use --registries ?

cmd/minikube/cmd/config/enable.go
```diff
@@ -86,6 +86,9 @@ You can view the list of minikube maintainers at: https://github.com/kubernetes/
 		if registries != "" {
 			viper.Set(config.AddonRegistries, registries)
 		}
+		if imageRepository != "" {
```
medyagh, line 89: instead of creating a new flag, why not use the existing one 

is there a good reason not to use --registries ?

PhantomInTheWire: there were some semantic and ux reasons as well but the major reason is:
missing org/tag support: --registries only handles the registry portion, but the Helm use case also needs to override the org path and optionally the tag (e.g., switching from docker.io/volcanosh:v1.13.0 to ghcr.io/volcano-sh:latest)

PhantomInTheWire:
thankyou so much for the quick review @medyagh 

>thank you so much @PhantomInTheWire for the update, I am curious is "latest" needed there? what happens if user doesnt specificy ghcr.io/volcano-sh:latest

it defaults to latest if nothing is specified


>is there a good reason not to use --registries ?

it was semantically difficult to use and i needed a lot more than just overriding registry(tag and org path)

cmd/minikube/cmd/config/enable.go
```diff
@@ -104,13 +107,15 @@ func isOfficialMaintainer(maintainer string) bool {
...
-	images     string
-	registries string
+	images          string
+	registries      string
+	imageRepository string
 )
 
 func init() {
 	addonsEnableCmd.Flags().StringVar(&images, "images", "", "Images used by this addon. Separated by commas.")
 	addonsEnableCmd.Flags().StringVar(&registries, "registries", "", "Registries used by this addon. Separated by commas.")
```
medyagh, line 117: @nirs @afbjorklund what do you think

PhantomInTheWire:
rebased to master

cmd/minikube/cmd/config/enable.go
medyagh: in helm world do they Have to specify the tag for the image respository or is that optional ?

medyagh: I am mor ein favor of making things more simple and not add one more flag, I rather if we can just have one flag for both, (we can add an alias for a flag to match helm)

what is helms exact name and documetnation

medyagh (changes requested):

cmd/minikube/cmd/config/enable.go
PhantomInTheWire: in helm tag is optional and default is latest, but tags are widely used for getting specific version

PhantomInTheWire: Helm doesn't have a direct equivalent flag, it uses --set flag with chart-specific keys.

medyagh (changes requested):
@PhantomInTheWire thanks for updating the PR,
if we cant merge the --registries and --image-repository
 how about we add a safe guard that if user tried to use the --image-repostiry with addons that are not helm based it should Either try to do the right thing or tell the user to use --registries https://minikube.sigs.k8s.io/docs/handbook/addons/custom-images/


@PhantomInTheWire I am curious is there anway way you could try to make this One Flag for both ?

like if it is Not Helm addon, it should try to add the image repo and if it is helm based use the new logic ?

and make sure we dont break this tutorial ?
https://minikube.sigs.k8s.io/docs/handbook/addons/custom-images/

medyagh:
@PhantomInTheWire u stil want to take a look at my last comment?

PhantomInTheWire:
@medyagh yeah I am still taking a look at it, sorry for the delay, I've been caught up in some uni stuff, I think it can be done in a satisfactory way, I'll push the fix by weekend if that's okay

medyagh:
> @medyagh yeah I am still taking a look at it, sorry for the delay, I've been caught up in some uni stuff, I think it can be done in a satisfactory way, I'll push the fix by weekend if that's okay

sure

PhantomInTheWire:
hey @medyagh, so I looked into unifying them again, but merging --registries and --image-repository is very tricky because they operate on fundamentally different logic:

--registries does a simple string replacement on the registry domain (e.g. docker.io -> ghcr.io). It doesn't support changing the organization/path structure(eg: docker.io/volcanosh vs ghcr.io/volcano-sh), we could change the nature of the flag itself to a new structure that would allow for this sort of thing(and the legacy stuff) but that seems very confusing from a ux point of view.

--image-repository (my addition) parses the input to extract registry, org, and tag separately (e.g. ghcr.io/volcano-sh:latest), which is often required for Helm charts that split these into different values (like basic.image_registry vs basic.controller_image_name).

Attempting to apply the complex parsing logic of --image-repository to non-Helm addons would likely break them since their templates expect simple registry strings.

I have implemented the safeguard as suggested. If a user tries to use --image-repository with a non-Helm addon (like dashboard), minikube will now explicitly error and guide them to use the correct flag. This is to ensure that the user is educated on the correct usage without breaking existing behavior.

PhantomInTheWire:
ping @medyagh just in case this got lost

mloskot:
@PhantomInTheWire FIY, this PR was briefly discussed during the latest [Office Hours](https://tinyurl.com/minikube-oh), I'd consider joining the OH next week.

PhantomInTheWire:
@mloskot sure I'll be there

PhantomInTheWire:
also the calendar invite is of august 2019, we should probably fix that

Issue #21754, opened 2025-10-15
Fix Docker Crio Test Image Tests


medyagh:
The following Tests are failing on Docker CRIO Linux, we need to figoure out why image tests are failing,
the image tests are supposed to run simple commands such as 
minikube image ls

is that actually failing on our CIRO ?

98 | TestFunctional/parallel/ServiceCmdConnect | 602.99
-- | -- | --
115 | TestFunctional/parallel/ServiceCmd/DeployApp | 600.65
118 | TestFunctional/parallel/ImageCommands/ImageListShort | 2.27
124 | TestFunctional/parallel/ImageCommands/ImageLoadDaemon | 0.93
127 | TestFunctional/parallel/ImageCommands/ImageReloadDaemon | 0.94
131 | TestFunctional/parallel/ImageCommands/ImageTagAndLoadDaemon | 1.52
132 | TestFunctional/parallel/ImageCommands/ImageSaveToFile | 0.47
134 | TestFunctional/parallel/ImageCommands/ImageLoadFromFile | 0.19
135 | TestFunctional/parallel/ImageCommands/ImageSaveDaemon | 0.37
153 | TestFunctional/parallel/ServiceCmd/HTTPS | 0.53
154 | TestFunctional/parallel/ServiceCmd/Format | 0.53
155 | TestFunctional/parallel/ServiceCmd/URL | 0.53

98	[TestFunctional/parallel/ServiceCmdConnect](https://storage.googleapis.com/minikube-builds/logs/21724/41893/Docker_Linux_crio.html#fail_TestFunctional%2fparallel%2fServiceCmdConnect)	602.99
115	[TestFunctional/parallel/ServiceCmd/DeployApp](https://storage.googleapis.com/minikube-builds/logs/21724/41893/Docker_Linux_crio.html#fail_TestFunctional%2fparallel%2fServiceCmd%2fDeployApp)	600.65
118	[TestFunctional/parallel/ImageCommands/ImageListShort](https://storage.googleapis.com/minikube-builds/logs/21724/41893/Docker_Linux_crio.html#fail_TestFunctional%2fparallel%2fImageCommands%2fImageListShort)	2.27
124	[TestFunctional/parallel/ImageCommands/ImageLoadDaemon](https://storage.googleapis.com/minikube-builds/logs/21724/41893/Docker_Linux_crio.html#fail_TestFunctional%2fparallel%2fImageCommands%2fImageLoadDaemon)	0.93
[...]

PR #22214, opened 2025-12-18
cri: increase default 2s timeout for crictl commands

medyagh:
attempt to fix this in KVM CRIO Linux Tests
where "image ls" hits the limit

```
W1217 22:03:37.118496   21489 cache_images.go:736] Failed to list images for profile functional-683881 crictl images: sudo crictl images --output json: Process exited with status 1
stdout:
stderr:
E1217 22:03:37.106555   10313 log.go:32] "ListImages with filter from image service failed" err="rpc error: code = DeadlineExceeded desc = context deadline exceeded" filter="image:{}"
```

https://github.com/kubernetes/minikube/issues/21754#issuecomment-3671740283

### DIFF
--- a/pkg/minikube/cruntime/cri.go
+++ b/pkg/minikube/cruntime/cri.go
@@ -289,7 +289,7 @@ func populateCRIConfig(cr CommandRunner, socket string) error {
 
 // getCRIInfo returns current information
 func getCRIInfo(cr CommandRunner) (map[string]interface{}, error) {
-	args := []string{"crictl",timeoutOverrideFlag, "info"}
+	args := []string{"crictl", timeoutOverrideFlag, "info"}
 	c := exec.Command("sudo", args...)
 	rr, err := cr.RunCmd(c)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


afbjorklund:
You shouldn't have special code for docker, that seems like a legacy from the dockershim?

i.e. it should also call `listCRIImages`

Need to remove docker from the code...

medyagh:
> You shouldn't have special code for docker, that seems like a legacy from the dockershim?
> 
> i.e. it should also call `listCRIImages`
> 
> Need to remove docker from the code...

thats a good idea, I can do that in a separate PR I think I wanna limit this PR to crictl only

afbjorklund:
Was just reflecting on the "containerd and crio" subject, while the change is related to CRI (i.e. everything)

medyagh:
> Was just reflecting on the "containerd and crio" subject, while the change is related to CRI (i.e. everything)

thanks, changed the title, also created an issue for removing docker calls for CRI calls in the code base, will be tested in a sepaerate PR

Issue #22148, opened 2025-12-14
minikube panic when starting with --kubernetes-version v1.34.3


nirs:
### What Happened?

To reproduce run:

```console
minikube start --driver vfkit --kubernetes-version v1.34.3 --v=4 --alsologtostderr
```

Based on the logs, we find a preload:

```
I1214 18:13:35.140368   32301 preload.go:188] Checking if preload exists for k8s version v1.34.3 and runtime docker
W1214 18:13:35.332462   32301 preload.go:144] https://storage.googleapis.com/minikube-preloaded-volume-tarballs/v18/v1.34.3/preloaded-images-k8s-v18-v1.34.3-docker-overlay2-arm64.tar.lz4 status code: 404
I1214 18:13:35.526174   32301 preload.go:148] Found remote preload: https://github.com/kubernetes-sigs/minikube-preloads/releases/download/v18/preloaded-images-k8s-v18-v1.34.3-docker-overlay2-arm64.tar.lz4
```

But we fail to find the checksum in the GitHub API:

```
I1214 18:13:35.537801   32301 preload.go:329] getting checksum for preloaded-images-k8s-v18-v1.34.3-docker-overlay2-arm64.tar.lz4 from github api...
panic: runtime error: invalid memory address or nil pointer dereference
```

There is no issue with --kubernetes-version v1.34.2.

Tested with master (commit ab27a62312f816494976f44834dde5dcf95facfe): broken

## Testing with release 1.37.0

It works in latest release - so this is a regression in master.

```
% /opt/homebrew/bin/minikube start --driver vfkit --kubernetes-version v1.34.3
😄  minikube v1.37.0 on Darwin 26.1 (arm64)
❗  Specified Kubernetes version 1.34.3 is newer than the newest supported version: v1.34.0. Use `minikube config defaults kubernetes-version` for details.
❗  Specified Kubernetes version 1.34.3 not found in Kubernetes version list
🤔  Searching the internet for Kubernetes version...
✅  Kubernetes version 1.34.3 found in GitHub version list
✨  Using the vfkit driver based on user configuration
👍  Starting "minikube" primary control-plane node in "minikube" cluster
🔥  Creating vfkit VM (CPUs=2, Memory=6144MB, Disk=20000MB) ...
🐳  Preparing Kubernetes v1.34.3 on Docker 28.4.0 ...
🔗  Configuring bridge CNI (Container Networking Interface) ...
🔎  Verifying Kubernetes components...
[...]

PR #22221, opened 2025-12-18
 add --preload-source flag to enfroce preload download src

medyagh:
- add --preload-src flag to enfroce preload download src (default auto, options gcs, github)
- add integration tests to ensure we can download from both

fixes https://github.com/kubernetes/minikube/issues/22148

### DIFF
--- a/test/integration/preload_test.go
+++ b/test/integration/preload_test.go
@@ -83,9 +83,10 @@ func TestPreload(t *testing.T) {
 		}
 	})
 	// PreloadSrc verifies that downloading preload from github and gcs works using --preload-src and --download-only
+	// "auto" is the default preload source (tries both gcs and github); here we explicitly verify each source
 	t.Run("PreloadSrc", func(t *testing.T) {
 		MaybeParallel(t)
-		// "gcs" is the default source, so we can verify it by default
+		
 		tests := []struct {
 			name              string
 			source            string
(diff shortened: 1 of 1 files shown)


cmd/minikube/cmd/start_flags.go
```diff
@@ -146,6 +146,7 @@ const (
 	staticIP                = "static-ip"
 	gpus                    = "gpus"
 	autoPauseInterval       = "auto-pause-interval"
+	preloadSrc              = "preload-src"
```
nirs, line 149: --preload-src is not a good name for users. We want to use complete words to make the flag more clear. For common flags we provide short option to make it easy to type, but flags that will almost never be used should user clear names.

Change to --preload-source or --preload-provider?

cmd/minikube/cmd/start_flags.go
```diff
@@ -212,7 +212,7 @@ func initMinikubeFlags() {
 	startCmd.Flags().String(staticIP, "", "Set a static IP for the minikube cluster, the IP must be: private, IPv4, and the last octet must be between 2 and 254, for example 192.168.200.200 (Docker and Podman drivers only)")
 	startCmd.Flags().StringP(gpus, "g", "", "Allow pods to use your GPUs. Options include: [all,nvidia,amd] (Docker driver with Docker container-runtime only)")
 	startCmd.Flags().Duration(autoPauseInterval, time.Minute*1, "Duration of inactivity before the minikube VM is paused (default 1m0s)")
-	startCmd.Flags().String(preloadSrc, "gcs", "Which source to download the preload from (valid options: gcs, github)")
+	startCmd.Flags().String(preloadSrc, "auto", "Which source to download the preload from (valid options: gcs, github, auto). Defaults to auto (try both).")
```
nirs, line 215: Lets squash this commit to the previous, there is no point in suggesting a change in one commit and changing it in the next commit.

nirs (commented):
I reviewed only the first commits, I don't have time to review more now.

cmd/minikube/cmd/start_flags.go
medyagh: thats a good idea !

medyagh, line 149: @copilot  do you prefer --preload-source or --preload-provider?

PR #22263, opened 2025-12-20
docker runtime: Use CRI for ListImages function

### LINKED ISSUE #22215: use "crictl" and cri for docker runtime
You shouldn't have special code for docker, that seems like a legacy from the dockershim?

i.e. it should also call `listCRIImages`

Need to remove docker from the code...

_Originally posted by @afbjorklund in https://github.com/kubernetes/minikube/issues/22214#issuecomment-3671867709_

afbjorklund:
Closes #22215

### DIFF
--- a/pkg/minikube/cruntime/docker.go
+++ b/pkg/minikube/cruntime/docker.go
@@ -262,6 +262,9 @@ func (r *Docker) ImageExists(name string, sha string) bool {
 
 // ListImages returns a list of images managed by this container runtime
 func (r *Docker) ListImages(ListImagesOptions) ([]ListImage, error) {
+	if r.UseCRI {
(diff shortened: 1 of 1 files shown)


afbjorklund:
The "UseCRI" is always true, but the cleanup and refactoring can go in a separate PR.

And it might not always be true either, there might be a "only between 1.24 and 1.32")

afbjorklund:
> curious is this not in contradict with [https://github.com/kubernetes/minikube/pull/22263 ?](https://github.com/kubernetes/minikube/issues/22264) ?

Not really, that is more related to the setting of the "UseCRI" variable.

This change makes it more consistent, all functions working the same?

You could argue the fix is a couple of years late, and you would be right...

Issue #22105, opened 2025-12-11
Rename the "sysdig" and "podman" packages in ISO buildroot


afbjorklund:
The package names are conflicting with the default packages:

* [x] ~~sysdig (since 2015.05)~~

* [ ] podman (since 2025.05)

Renaming them would avoid having to patch/remove buildroot...

```shell
                perl -pi -e 's@\s+source "package/sysdig/Config\.in"\n@@;' $(BUILD_DIR)/buildroot/package/Config.in; \
                rm -r $(BUILD_DIR)/buildroot/package/sysdig; \
```

https://github.com/kubernetes/minikube/pull/21997#issuecomment-3592937690

Similar to other packages, that have been duplicated from default:

crun-latest 
runc-master
cni-plugins-latest
containerd-bin
docker-bin
docker-buildx
nerdctl-bin

Eventually it would be nice to use upstream OS as much as possible.

afbjorklund:
A side quest is to merge as many of the "arch" packages as possible:

package/
├── automount
├── cni
├── crio-bin
├── crun-latest
├── podman
├── runc-master
├── scheduled-stop
└── sysdig
arch/x86_64/package/
├── buildkit-bin
├── cni-plugins-latest
├── containerd-bin
├── crictl-bin
├── cri-dockerd
├── docker-bin
├── docker-buildx
├── hyperv-daemons
├── nerdctl-bin
└── vbox-mount-service
arch/aarch64/package/
├── buildkit-bin-aarch64
├── cni-plugins-latest-aarch64
├── containerd-bin-aarch64
├── crictl-bin-aarch64
├── cri-dockerd-aarch64
├── docker-bin-aarch64
├── docker-buildx-aarch64
└── nerdctl-bin-aarch64

They only need a couple of ARCH variables, and multiple hash entries...

Then there is no need to duplicate the _entire package_, under "arch"?

It is also possible to have arch-constraints, in the regular package dir.

EDIT:

* #22135

afbjorklund:
minikube

BUILDKIT_BIN_VERSION = v0.18.1
CNI_PLUGINS_LATEST_VERSION = v1.8.0
CNI_VERSION = v0.7.1
CONTAINERD_BIN_VERSION = 2.1.4
CRICTL_BIN_VERSION = v1.34.0
CRI_DOCKERD_VERSION = 55d6e1a1d6f2ee58949e13a0c66afe7d779ac942
CRIO_BIN_VERSION = v1.29.1
[...]

PR #22290, opened 2025-12-22
iso: Remove unused sysdig package from image

afbjorklund:
A newer version 0.29 is available upstream, if needed... But the package was *not* enabled by default in minikube.

Remove old patches from previous version 0.23 (now 0.27), remove dependency package luajit that was added manually.

Issue #22105

```
    BR2_PACKAGE_LUAJIT
        Make sure to build lua required for sysdig
```

commit f91a11192ed1b319ab476434110cc6734107e332


afbjorklund:
ok-to-build-iso

medyagh:
I guess we removed falco for good in this PR https://github.com/kubernetes/minikube/pull/20720/changes/4ef9e5160cfffeff678e8c285f42d64142ea8c73

maybe we should create an issue and let community to take that if they wanna re-add it

PR #22312, opened 2025-12-23
Fix misleading Docker Desktop references in error messages

syscl:
The problem is I didn't have docker desktop installed on my macOS, but it shows this confused information. The problem is the function assume people use docker desktop on macOS / Windows unconditionally, which is inaccurate and confused for debugging. This improves this logic handling. 

The fix detects actual Docker runtime instead of assuming
Docker Desktop on macOS/Windows. Updates `IsDockerDesktop()` to query daemon
info and `FullName()` to return the actual runtime name for accurate error
messages.


medyagh (commented):
@syscl can you plz put an example output in the PR description. both Before and After this PR?

syscl:
Hi @medyagh wish all is well! Apology for the late reply due to super busy recently, will put. Basically it is the log for the memory that is not accurate if I do not use Docker Desktop.

Issue #22231, opened 2025-12-19
"minikube service" should use "https" when the forwarded port is 443


medyagh:
example when trying to install dashbord

crio_optional ✓
$ mk service kubernetes-dashboard-kong-proxy -n kubernetes-dashboard
│      NAMESPACE       │              NAME               │    TARGET PORT     │             URL              │
│ kubernetes-dashboard │ kubernetes-dashboard-kong-proxy │ kong-proxy-tls/443 │ http://192.168.105.143:32726 │

🎉  Opening service kubernetes-dashboard/kubernetes-dashboard-kong-proxy in default browser...

openning http://192.168.105.143:32726 will not work in the browser and make the user think it is broken
it is better give https:// in the first place

p172913:
tested using https protocol when a Service is using port 443. while testing, found that service is not TLS-enabled. Because of that generated HTTPS URL  in browser having issues with SSL/protocol. Since Minikube donot provide any TLS certificates for services, using HTTPS purely because of the port leads to issues.

**Conclusion:** using HTTP by default is safer unless TLS certificate is configured by the user.

p172913:
@medyagh Can you close this Issue. I don't have access to close this. So, please close this as you mention in the PR. Thanks...

PR #22317, opened 2025-12-24
wip: make minikube to use https when port in use is 443

p172913:
What type of PR is this?
/kind enhancement

What this PR does / why we need it:

Made minikube to use https protocol when using forwarded port as 443

Which issue(s) this PR fixes:
Fixes https://github.com/kubernetes/minikube/issues/22231

Special notes for your reviewer:
None

Does this PR introduce a user-facing change?
None

Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:
None

### DIFF
--- a/cmd/minikube/cmd/service.go
+++ b/cmd/minikube/cmd/service.go
@@ -126,13 +126,25 @@ You may select another namespace by using 'minikube service {{.service}} -n <nam
 		var noNodePortServices service.URLs
 
 		for _, svc := range services {
-			openUrls, err := service.WaitForService(co.API, co.Config.Name, namespace, svc.Name, serviceURLTemplate, serviceURLMode, https, wait, interval)
+			// Check if port is 443
+			isHTTPSIntent := false
+			for _, portName := range svc.PortNames {
+				lowered := strings.ToLower(portName)
+				if strings.Contains(lowered, "https") || strings.Contains(lowered, "443") {
+					isHTTPSIntent = true
+					break
+				}
+			}
+
+			shouldUseHTTPS := https || isHTTPSIntent
+
+			openUrls, err := service.WaitForService(co.API, co.Config.Name, namespace, svc.Name, serviceURLTemplate, serviceURLMode, shouldUseHTTPS, wait, interval)
 
 			if err != nil {
 				var s *service.SVCNotFoundError
 				if errors.As(err, &s) {
 					exit.Message(reason.SvcNotFound, `Service '{{.service}}' was not found in '{{.namespace}}' namespace.
-You may select another namespace by using 'minikube service {{.service}} -n <namespace>'. Or list out all the services using 'minikube service list'`, out.V{"service": svc, "namespace": namespace})
+You may select another namespace by using 'minikube service {{.service}} -n <namespace>'. Or list out all the services using 'minikube service list'`, out.V{"service": svc.Name, "namespace": namespace})
 				}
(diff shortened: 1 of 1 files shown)


medyagh:
@p172913 also make sure gofmt your code

make gomft

p172913:
> @p172913 thank you for this PR, can you make a Before/After this PR?

### Before this PR
- Services exposed to port 443 did not use HTTPS usage when generating service URLs.

### After this PR
- Services exposed to port 443 now correctly uses HTTPS by setting the `UseHTTPS` flag when URL is generated.

medyagh:
> * osed to port 443 did not use HTTPS usage when generating service URL

@p172913 by before/after this PR I mean an example of using this feature and copy/paste the output before/after this PR

p172913:
> > * osed to port 443 did not use HTTPS usage when generating service URL
> 
> @p172913 by before/after this PR I mean an example of using this feature and copy/paste the output before/after this PR

@medyagh One small confirmation. Should this changes also reflect in cli also...

p172913:
> > * osed to port 443 did not use HTTPS usage when generating service URL
> 
> @p172913 by before/after this PR I mean an example of using this feature and copy/paste the output before/after this PR

First ran:
kubectl create deployment https-demo --image=nginx
kubectl expose deployment https-demo --type=NodePort --port=443
Before PR::
image

After PR::
image

medyagh:
@p172913 can you check in case of Docker Driver on Windows, will the port be available in https ?? becuase on windows we will be tunneling, if you hit that url htps are you 100% sure it works ? can you screenshot the url hitting too

p172913:
> @p172913 can you check in case of Docker Driver on Windows, will the port be available in https ?? becuase on windows we will be tunneling, if you hit that url htps are you 100% sure it works ? can you screenshot the url hitting too

Verified that the https:// prefix working correctly for the Docker driver on Windows.
code correctly finds port 443 and using the https protocol.
ERR_SSL_PROTOCOL_ERROR in the browser confirms that the encrypted request successfully reached the backend. In this test I used a standard Nginx pod without a certificate
image
image

medyagh:
@p172913
can you screenshot without https? what happens then?

seems like chrome does NOT let you see the url with https in this case, so maybe http is better?

p172913:
> @p172913 can you screenshot without https? what happens then?
> 
> seems like chrome does NOT let you see the url with https in this case, so maybe http is better?

If I am using http protocol( not with 443 port):
image
image
It will be same with https if we provide TLS certificate.

medyagh:
@p172913 in the example you provided if we give them https it wont work, it will take them to the Warnning page by chrome that the cert is not valid, so maybe this is not a good idea to do ! at all

p172913:
> @p172913 in the example you provided if we give them https it wont work, it will take them to the Warnning page by chrome that the cert is not valid, so maybe this is not a good idea to do ! at all

@medyagh I agree with you. After testing, it is clear that switching to HTTPS purely based on port 443 leads to browser errors when the service does not have a  TLS cert. 
Since HTTPS cannot work without TLS and Minikube doesn’t provide any default certificates, using HTTP here seems good idea here. Please correct me if I’m missing something.

medyagh:
thank you for help in investigating this issue, please close and feel free add it as a comment to the issue

p172913:
> thank you for help in investigating this issue, please close and feel free add it as a comment to the issue

@medyagh Happy to help.

PR #22329, opened 2025-12-26
Add Falco as optional Linux-only Minikube addon

### LINKED ISSUE #22298: add falco back to the new iso
we removed falco in https://github.com/kubernetes/minikube/pull/20720/changes/4ef9e5160cfffeff678e8c285f42d64142ea8c73
also we removed sysdig needed by falco here
https://github.com/kubernetes/minikube/pull/22290

if any developer search for falco on minikube, this task is available to work on as Nice to have !

ThankaBharathi:
This PR reintroduces Falco as an optional Minikube addon (DaemonSet-based),
enabled via `minikube addons enable falco`.

Falco was previously removed from the ISO during buildroot updates.
This approach avoids bundling Falco into the ISO while still allowing
users to experiment with runtime security locally.

The addon is marked as `supported_os: linux`, so it is intentionally
hidden on non-Linux hosts, which is expected behavior.

Fixes #22298

### Example usage

```bash
# Enable Falco addon
minikube addons enable falco

# Check if Falco pod is running
kubectl get pods -n kube-system | grep falco

# View Falco logs
kubectl logs -n kube-system -l app=falco | head
```

**Expected output:**
```
falco is an experimental addon
Verifying addon image integrity …
The 'falco' addon is enabled
falco-7k9x2   1/1   Running   0   25s
Falco version: 0.38.1
Loading rules from file /etc/falco/falco_rules.yaml
Loading rules from file /etc/falco/falco_rules.local.yaml
Starting internal webserver, listening on port 8765
```

### DIFF
--- a/deploy/addons/falco/falco-addon.yaml
+++ b/deploy/addons/falco/falco-addon.yaml
@@ -1,0 +1,13 @@
+apiVersion: addons.minikube.sigs.k8s.io/v1
+kind: Addon
+metadata:
+  name: falco
+spec:
+  description: Runtime security monitoring using Falco
+  supported_os:
+    - linux
+  default: false
+  requires:
+    - kubernetes
+  manifests:
+    - falco-daemonset.yaml.tmpl
--- a/deploy/addons/falco/falco-daemonset.yaml.tmpl
+++ b/deploy/addons/falco/falco-daemonset.yaml.tmpl
@@ -1,0 +1,38 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: falco
+  namespace: kube-system
+  labels:
+    app: falco
+spec:
+  selector:
+    matchLabels:
+      app: falco
+  template:
+    metadata:
+      labels:
+        app: falco
+    spec:
+      hostPID: true
+      containers:
+        - name: falco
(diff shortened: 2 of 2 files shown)


medyagh:
@ThankaBharathi would plz share an example output of using this addon in the PR description

ThankaBharathi:
Thanks for the suggestion! I’ve added an example usage section with expected output to the PR description.

afbjorklund:
I don't see why you would need this addon, and not just use helm:

https://falco.org/docs/getting-started/falco-kubernetes-quickstart/

> Falco was previously removed from the ISO during buildroot updates.

Falco was **not** included in the ISO, only the kernel module was prebuilt.

ThankaBharathi:
Helm is definitely the recommended and most flexible way to deploy Falco, especially for production or highly customized setups.

The addon is intended for a different audience: Minikube users who want a simple, discoverable, one-command way to try Falco locally, similar to other built-in addons like ingress or metrics-server, even though those can also beinstalled via Helm.

This PR intentionally avoids reintroducing Falco into the ISO or kernel module path, and does not replace the Helm-based workflow. Helm remains the preferred approach for advanced use cases, while the addon targets quick experimentation,
learning, and demos.

Issue #22137, opened 2025-12-13
Refactor the names and versions of packages in ISO buildroot


afbjorklund:
Binary packages end with `-bin`, while the regular (i.e. source) packages do **not**

Versions do **not** have a leading `v`, even though git tags and downloads might

See https://buildroot.org/downloads/manual/manual.html#_developer_guide

This is to make it more similar with upstream, so that packages can be upstreamed.

----

minikube

BUILDKIT_BIN_VERSION = v0.18.1
CNI_PLUGINS_LATEST_VERSION = v1.8.0
CNI_VERSION = v0.7.1
CONTAINERD_BIN_VERSION = 2.1.4
CRICTL_BIN_VERSION = v1.34.0
CRI_DOCKERD_VERSION = 55d6e1a1d6f2ee58949e13a0c66afe7d779ac942
CRIO_BIN_VERSION = v1.29.1
CRUN_LATEST_VERSION = 1.23.1
DOCKER_BIN_VERSION = 28.5.2
DOCKER_BUILDX_VERSION = v0.17.1
NERDCTL_BIN_VERSION = 2.1.3
PODMAN_VERSION = v3.4.7
RUNC_MASTER_VERSION = v1.4.0
SYSDIG_VERSION = 0.27.1
VBOX_MOUNT_SERVICE_VERSION = 5.2.42

buildroot (2025.02)

MOBY_BUILDKIT_VERSION = 0.11
CNI_PLUGINS_VERSION = 1.3.0
CONTAINERD_VERSION = 2.0.2
CRUN_VERSION = 1.18.2
DOCKER_ENGINE_VERSION = 27.5.1
DOCKER_CLI_BUILDX_VERSION = 0.16.1
NERDCTL_VERSION = 1.7.3
RUNC_VERSION = 1.2.4
SYSDIG_VERSION = 0.29.3

PR #22348, opened 2025-12-27
ISO: clean up the "patches" directory and newer go.hash version

afbjorklund:
*    Merge the patches directory separated by arch
*    Remove patches for obsolete package versions

*    Use patches for go.hash instead of copying
*    Clean up the go.hash file to match upstream
 
Closes: #22104

Replaces:

* https://github.com/kubernetes/minikube/pull/22116


afbjorklund:
When this patch is in, it is possible to add a checksum also for the custom kernel in the `linux.hash`

```
# From https://www.kernel.org/pub/linux/kernel/v6.x/sha256sums.asc
sha256  18640ae9a2c952f8f8868324378df1a9e138ad29e6b93a9c38155764ea67ea85  linux-6.6.95.tar.xz
```

Then it will be possible to verify _all_ the downloads, which is good thing - especially when caching.

```
config BR2_DOWNLOAD_FORCE_CHECK_HASHES
        bool "Force all downloads to have a valid hash"
        help
          Say 'y' here to enforce downloads to have at least one valid
          hash (and of course, that all hashes be valid).

          By default, Buildroot checks hashes of all packages
          downloaded, except those for which a custom version is
          used.

          With this option turned on, Buildroot will check hashes of
          all packages, including those that use a custom version. In
          order to provide hashes for such packages, place additional
          hash files in BR2_GLOBAL_PATCH_DIR directories.

```

afbjorklund:
It will also be possible to upgrade _other_ packages, not just host-go, in the same way:

* https://github.com/kubernetes/minikube/issues/22137

That is, only change the ***_VERSION and add the new checksum to the patches dir.

PR #22350, opened 2025-12-27
ci: split iso makefile targets into host and docker

### LINKED ISSUE #22145: Mount the ccache directory in the docker image for ISO buildroot
Using [ccache](https://buildroot.org/downloads/manual/manual.html#ccache) is recommended, but needs to be added to the docker image to be persisted:

```Dockerfile
RUN useradd -ms /bin/bash br-user && \
    mkdir -p /home/br-user/.buildroot-ccache && \
    chown -R br-user:br-user /home/br-user

# Persist the build cache between builds
VOLUME ["/home/br-user/.buildroot-ccache"]

USER br-user
WORKDIR /home/br-user
ENV HOME=/home/br-user
ENV LC_ALL=en_US.UTF-8
```

Otherwise the cache will be deleted with the build container. And that would defeat the purpose.

`BR2_USE_CCACHE=y`

### LINKED ISSUE #22165: The minikube iso build target in make doesn't work anymore
It is trying to parse the iso target as a minikube target, and fails.

`go: unsupported GOOS/GOARCH pair aarch64.iso/aarch64.iso`

```Makefile
out/minikube-%.iso: $(shell find "deploy/iso/minikube-iso" -type f)
ifeq ($(IN_DOCKER),1)
        $(MAKE) minikube-iso-$*
else
        docker run --rm --workdir /mnt --volume $(CURDIR):/mnt:Z $(ISO_DOCKER_EXTRA_ARGS) \
                --user $(shell id -u):$(shell id -g) --env HOME=/tmp --env IN_DOCKER=1 \
                $(ISO_BUILD_IMAGE) /bin/bash -lc '/usr/bin/make minikube-iso-$*'
endif
```

### LINKED ISSUE #22227: Build auto-pause on the host and not in the builder image
Then it can use the regular go compiler and cache, and just `GOOS=linux`

Like ./deploy/kicbase/build_auto_pause.sh and PREBUILT_AUTO_PAUSE

Previously it was building auto-pause in emulation (the docker --platform)

Now it is cross-compiling instead, and copying the result to the build context

afbjorklund:
Make the iso target higher priority
Closes #22165

Split iso targets into host and docker
`IN_DOCKER`

Build auto-pause program on the host
Closes #22227

Mount the ccache and tmp volumes
Closes #22145

### DIFF
--- a/hack/jenkins/build_iso.sh
+++ b/hack/jenkins/build_iso.sh
@@ -72,7 +72,7 @@ else
 	export ISO_BUCKET
 fi
 
-if ! make release-iso 2>&1 | tee iso-logs.txt; then
+if ! IN_DOCKER=1 make release-iso 2>&1 | tee iso-logs.txt; then
     # Exit of `make` (PIPESTATUS[0]); fallback to 1 if unavailable
     ec=${PIPESTATUS[0]:-1}
 


afbjorklund:
ok-to-build-iso

Makefile
```diff
@@ -287,10 +291,8 @@ out/e2e-%: out/minikube-%
 out/e2e-windows-amd64.exe: out/e2e-windows-amd64
 	cp $< $@
 
-minikube-iso-amd64: minikube-iso-x86_64
-minikube-iso-arm64: minikube-iso-aarch64
-
-minikube-iso-%: iso-prepare-% deploy/iso/minikube-iso/board/minikube/%/rootfs-overlay/usr/bin/auto-pause # build minikube iso
+minikube-iso-%: # build minikube iso
+	$(MAKE) iso-prepare-$* deploy/iso/minikube-iso/board/minikube/%/rootfs-overlay/usr/bin/auto-pause
```
nirs, line 295: Why not use the dependency for iso-prepare-%?

We need this target to run iso-menuconfig-% and linux-menuconfig-% or minikube-iso-% on a clean checkout. Using the dependency for all the targets make this clear and consistent.

Makefile
```diff
@@ -301,11 +308,18 @@ minikube-iso-%: # build minikube iso
...
-.PHONY: iso-prepare-%
+# iso-prepare runs on the host, it sets up the buildroot rootfs-overlay content
+
+.PHONY: iso-prepare-x86_64 iso-prepare-aarch64
 iso-prepare-%: buildroot
 	echo $(VERSION_JSON) > deploy/iso/minikube-iso/board/minikube/$*/rootfs-overlay/version.json
 	echo $(ISO_VERSION) > deploy/iso/minikube-iso/board/minikube/$*/rootfs-overlay/etc/VERSION
 	cp deploy/iso/minikube-iso/arch/$*/Config.in.tmpl deploy/iso/minikube-iso/Config.in
+
+# iso-config runs in the builder, it creates the buildroot .config file to use
```
nirs, line 319: maybe use iso-prepare-host and iso-prepare-container? The comment is very helpful but the name can explain the purpose of the target.

Why do need to separate the setup steps on the host and in the container?

Makefile
```diff
@@ -349,6 +349,7 @@ ifeq ($(IN_DOCKER),1)
 	$(MAKE) iso-build-$*
 else
 	docker run --rm --workdir /mnt --volume $(CURDIR):/mnt:Z $(ISO_DOCKER_EXTRA_ARGS) \
+		--volume buildroot-ccache:/var/cache/buildroot --volume /tmp \
```
nirs, line 352: This does not work on Fedora/RHEL (or other distros using selinux) - we must use ":Z" suffix for volumes, as used in all other docker commands in this makefile.

afbjorklund: Do you still need to relabel, when not sharing files with the host? Didn't think selinux cared about volumes

Makefile
afbjorklund: Otherwise we would need to redesign the targets some other way, so that the pattern rules work the way wanted. But some things got a bit out of sync, when trying to split the big change into commits

afbjorklund: The second part split up the buildroot and the output directories, and did all the writing into `$O` only.

Makefile
afbjorklund: `$(CURDIR):/mnt:Z` is not a volume, it is a bind mount. But wondering if you still need to use :z on volumes?

afbjorklund: No, it seems to work just fine without relabeling (unless someone else previously relabled the volume private)

Makefile
```diff
@@ -287,10 +291,15 @@ out/e2e-%: out/minikube-%
 out/e2e-windows-amd64.exe: out/e2e-windows-amd64
 	cp $< $@
 
-minikube-iso-amd64: minikube-iso-x86_64
-minikube-iso-arm64: minikube-iso-aarch64
+out/minikube-amd64.iso:
+	$(MAKE) minikube-iso-x86_64
+out/minikube-arm64.iso:
+	$(MAKE) minikube-iso-aarch64
```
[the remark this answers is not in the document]
afbjorklund: The minikube-iso-x86_64 _does_ create the out/minikube-amd64.iso

afbjorklund:
Will redo as smaller changes.

PR #22354, opened 2025-12-29
Always build the auto-pause binary on the host

### LINKED ISSUE #22227: Build auto-pause on the host and not in the builder image

afbjorklund:
Build the "auto-pause" binary on the host, just like the "minikube" program.

Then just copy the output into `deploy/kicbase` or `deploy/iso/minikube-iso`

This avoids needing the entire minikube code base, for building OS images.

In the future it would be possible to use binaries, from "kicbase-artifacts"

Closes #22227

`PREBUILT_AUTO_PAUSE=true`, always

----

For some reason, auto-pause is built differently for each of the OS targets...

For KIC it is static and in `/bin`, but for ISO it is dynamic and in `/usr/bin`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ KIC_VERSION ?= $(shell grep -E "Version =" pkg/drivers/kic/types.go | cut -d \" 
 HUGO_VERSION ?= $(shell grep -E "HUGO_VERSION = \"" netlify.toml | cut -d \" -f2)
 
 # Default to .0 for higher cache hit rates, as build increments typically don't require new ISO versions
-ISO_VERSION ?= v1.37.0-1766979747-22353
+ISO_VERSION ?= v1.37.0-1767060032-22354
 
 # Dashes are valid in semver, but not Linux packaging. Use ~ to delimit alpha/beta
 DEB_VERSION ?= $(subst -,~,$(RAW_VERSION))
--- a/pkg/minikube/download/iso.go
+++ b/pkg/minikube/download/iso.go
@@ -41,7 +41,7 @@ const fileScheme = "file"
 // DefaultISOURLs returns a list of ISO URL's to consult by default, in priority order
 func DefaultISOURLs() []string {
 	v := version.GetISOVersion()
-	isoBucket := "minikube-builds/iso/22353"
+	isoBucket := "minikube-builds/iso/22354"
 
 	return []string{
 		fmt.Sprintf("https://storage.googleapis.com/%s/minikube-%s-%s.iso", isoBucket, v, runtime.GOARCH),


deploy/kicbase/Dockerfile
```diff
@@ -52,17 +35,17 @@ ARG NERDCTL_VERSION="2.2.0"
...
-COPY deploy/kicbase/containerd.toml /etc/containerd/config.toml
-COPY deploy/kicbase/containerd_docker_io_hosts.toml /etc/containerd/certs.d/docker.io/hosts.toml
-COPY deploy/kicbase/clean-install /usr/local/bin/clean-install
-COPY deploy/kicbase/entrypoint /usr/local/bin/entrypoint
-COPY deploy/kicbase/CHANGELOG ./CHANGELOG
-COPY deploy/kicbase/nerdctld/nerdctld.socket  /etc/systemd/system/nerdctld.socket
-COPY deploy/kicbase/nerdctld/nerdctld.service  /etc/systemd/system/nerdctld.service
-COPY --from=auto-pause /src/cmd/auto-pause/auto-pause-${TARGETARCH} /bin/auto-pause
+COPY 10-network-security.conf /etc/sysctl.d/10-network-security.conf
+COPY 11-tcp-mtu-probing.conf /etc/sysctl.d/11-tcp-mtu-probing.conf
```
medyagh, line 39: curious why we are removing the path prefix ?
we are removing the relative path "deploy/kicbase/" woulnd't that break some docker build ? if the context expects it in a relative way?

afbjorklund: Because we don't want to leak things from minikube, into the stand-alone kicbase. Also makes it simpler.

Makefile
```diff
@@ -702,8 +705,8 @@ endif
 ifndef CIBUILD
 	$(call user_confirm, 'Are you sure you want to push $(KICBASE_IMAGE_REGISTRIES) ?')
 endif
-	./deploy/kicbase/build_auto_pause.sh $(KICBASE_ARCH)
-	docker buildx build -f ./deploy/kicbase/Dockerfile --platform $(KICBASE_ARCH) $(addprefix -t ,$(KICBASE_IMAGE_REGISTRIES)) --push --build-arg VERSION_JSON=$(VERSION_JSON) --build-arg COMMIT_SHA=${VERSION}-$(COMMIT_NOQUOTES) --build-arg PREBUILT_AUTO_PAUSE=true .
+	./hack/build_auto_pause.sh $(KICBASE_ARCH) deploy/kicbase
```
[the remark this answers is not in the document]
afbjorklund: Nice try, but they were **both** wrong. The build script did a `cd`, so need to expand the relative path first.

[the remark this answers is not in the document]
afbjorklund: The error with the relative path was actually caught above, but suppressed "due to low confidence"

deploy/kicbase/Dockerfile
afbjorklund: Was trying to do the same thing for the buildroot deploy/iso, it should copy out everything it needs _first_

PR #22357, opened 2025-12-29
Warn user when host is using cgroup v1 in KIC

### LINKED ISSUE #22321: detect and warn user if the "Host" is using cgroup v1
https://github.com/kubernetes/minikube/issues/22318#issue-3761005528

and maybe add a check here to not use cgroup v2 in that case https://github.com/kubernetes/minikube/pull/22320

ThankaBharathi:
### What this PR does
- Detects when the host system is using cgroup v1
- Warns users during KIC (Kubernetes in Docker) node creation

### Why
Minikube may assume cgroup v2 support even when the host is using cgroup v1, which can lead to confusing runtime or test failures. This change adds an early, clear warning so users understand potential limitations before encountering issues.

### Example
On a Linux host using cgroup v1 (for example, older distributions or systems with legacy cgroup configuration), starting Minikube with the KIC driver will now print the following warning during node creation:

WARNING: Host system is using cgroup v1.
Minikube will not enable cgroup v2.
Some features may not work as expected.

This improves user visibility and avoids silent misconfiguration.

### Scope
- Linux only
- KIC driver only

Fixes #22321

### DIFF
--- a/pkg/drivers/kic/oci/cgroups_linux.go
+++ b/pkg/drivers/kic/oci/cgroups_linux.go
@@ -80,3 +80,21 @@ func hasMemorySwapCgroup() bool {
 	}
 	return true
 }
+
+// IsCgroupV2 returns true if the host is using cgroup v2
+func IsCgroupV2() bool {
+	_, err := os.Stat("/sys/fs/cgroup/cgroup.controllers")
+	return err == nil
+}
+
+// WarnIfCgroupV1 prints a warning if host is using cgroup v1
+func WarnIfCgroupV1() {
+	if !IsCgroupV2() {
+		klog.Warning(
+			"Host system is using cgroup v1. " +
+				"Minikube will not enable cgroup v2. " +
+				"Some features may not work as expected.",
+		)
+	}
+}
+
--- a/pkg/drivers/kic/oci/cgroups_other.go
+++ b/pkg/drivers/kic/oci/cgroups_other.go
@@ -25,3 +25,8 @@ func HasMemoryCgroup() bool {
 func hasMemorySwapCgroup() bool {
 	return true
 }
+
+// WarnIfCgroupV1 is a no-op on non-Linux systems
+func WarnIfCgroupV1() {
+	// no-op
+}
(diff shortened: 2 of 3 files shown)


afbjorklund:
@ThankaBharathi you need to rebase, there is some Falco commit attached here

ThankaBharathi:
Thanks for pointing that out! I’ve rebased the branch and removed the unrelated Falco commit.

pkg/drivers/kic/oci/cgroups_linux.go
```diff
@@ -80,3 +80,21 @@ func hasMemorySwapCgroup() bool {
 	}
 	return true
 }
+
+// IsCgroupV2 returns true if the host is using cgroup v2
+func IsCgroupV2() bool {
```
medyagh, line 85: @ThankaBharathi I am curious was this generated by AI?

we already have a func I think that checks the cgroups, I dont think, I think this PR might need to be on hold for more surgical or tactical

ThankaBharathi:
Thanks for this question 

No this is was not add by an ai. The intent was to add a very small, explicit check to distinguish cgroup v1 vs v2, since the existing helper in this file focus on controller availability (memory/swap) rather than identification whether the host is running a unified cgroup v2 hierarchy.

That said, I completely agree that it would be better to reuse or extend an existing helper if there is one that aldready capture this distinction. I'm happy to adjust the implementation to take it more surgical/tactical. Please let me know which exsiting function you'd prefer to build on, or the direction you'd like this to take.

pkg/drivers/kic/oci/cgroups_linux.go
ThankaBharathi: Thanks for the clarification.

I double checked this by grepping the repo:

While there is a cgroup v2 detection in deploy/kicbase/entrypoint (inside the node/container) and in various runtime/bootstrap path.

I couldn't find an existing go helper at the KIC host layer that explicitly distinguishes cgroup v1 vs v2 during container node creation. The existing helper in pkg/drivers/kic/oci focus on controller availability (memory/swap), Which is why added the small IsCgroupV2 function check here for host side warn. 

If you would prefer this logic to live else to be refactored to align with an existing pattern , I'm happy to adjust.

PR #22365, opened 2025-12-31
site: community page add button for joining dev-group

Wilson-Duncan:
Relaced dev mailing list bullet with RSVP button to link users to event and meeting sign up:

- Newlines between param link definitions for readability
- Added toml table for calendar rsvp button related variables
- Created partial for rsvp
- CSS styling for new partial

(Screenshot doesn't show mouse cursor)
Screenshot 2025-12-31 at 10 55 02 AM

### DIFF
--- a/site/config.toml
+++ b/site/config.toml
@@ -145,17 +145,14 @@ no = 'Sorry to hear that. Please <a href="https://github.com/kubernetes/minikube
 	name = "GitHub"
 	url = "https://github.com/kubernetes/minikube"
 	icon = "fab fa-github"
-        desc = "Development takes place here!"
-[[params.links.developer]]
-	name = "minikube-dev mailing list"
-	url = "https://groups.google.com/forum/#!forum/minikube-dev"
-	icon = "fa fa-envelope"
-	desc = "Contact the minikube developers here"
+    desc = "Development takes place here!"
+
 [[params.links.developer]]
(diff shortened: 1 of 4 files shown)


medyagh (commented):
In some websites I have seen that there is a way to add "an event" to your calendar. this would take them to the group which is okay, RSVP button seems not very helpful here, we need to figure out something that makes it more simple for users looking at the page, not sure whats the solution but I apperciate the exploration

Wilson-Duncan:
@medyagh 

> In some websites I have seen that there is a way to add "an event" to your calendar. this would take them to the group which is okay, RSVP button seems not very helpful here, we need to figure out something that makes it more simple for users looking at the page, not sure whats the solution but I apperciate the exploration

I see, if I understand correctly and the intention is to redirect the user to a calendar invite for the Monday Minikube SIG then simply updating the href to invite should do the trick - https://calendar.google.com/event?action=TEMPLATE&tmeid=NGxhdGFndGY5dW1sMjJnOW50ZDlrMnZjY3RfMjAxOTA0MDhUMTcwMDAwWiBoZWxpeGJsdWVAbQ&tmsrc=helixblue%40gmail.com&scp=ALL

Do you have the link to the calendar invite? The above was taken from the google doc but doesn't seem to work. Also, if that is not what your're looking for could you send an example website.

Issue #21619, opened 2025-09-22
Remove old references to docker machine


afbjorklund:
Docker Machine is not really a _thing_ anymore, and the code was archived more than 5 years ago:

https://github.com/docker-archive-public/docker.machine

```specfile
%description
Minikube uses Docker Machine to manage the Kubernetes VM so it benefits
from the driver plugin architecture that Docker Machine uses to provide
a consistent way to manage various VM providers.
```

Now it is part of minikube, and can be referred to as either libmachine or perhaps minikube-machine:

https://github.com/minikube-machine/machine

`require github.com/docker/machine v0.16.2`

`replace github.com/docker/machine => github.com/minikube-machine/machine`

Provisioning a machine with Docker on does not really help with running modern Kubernetes*.

Since it uses different container runtimes, such as containerd (with nerdctl) or cri-o (with podman).

Providing a Docker API for the node is better done on **top** of containerd, as an optional add-on:
https://docs.docker.com/engine/storage/containerd/ (from /var/lib/docker to /var/lib/containerd)

Similar to how you can access the CRI-O images with Podman, but the containers are kept separate?
Normally you would not run any containers that way, but only use it for loading and building images.

Eventually libmachine could be refactored to just be an internal implementation detail of minikube...

And then the minikube-machine organization could be retired, along with the no longer used drivers?

----

\* well, you could use Docker to run Kubernetes in. But that's not really the primary project goal?

I meant docker as a container runtime for Kubernetes, which has been fading out ever since 1.24.

afbjorklund:
Here is a list on the **unused** repositories that can be archived, they have been forked to minikube and not used:

* https://github.com/minikube-machine/machine-driver-hyperkit

* https://github.com/minikube-machine/machine-driver-kvm

* https://github.com/minikube-machine/machine-driver-qemu

Here are the unsupported repositories, that should be archived - **after** they are removed from the minikube code:

* https://github.com/minikube-machine/machine-driver-parallels

* https://github.com/minikube-machine/machine-driver-vmware

And then finally archive, after removing commands/mcndirs and the unused drivers (generic, none), and moving it:

* https://github.com/minikube-machine/machine / drivers
 
* https://github.com/minikube-machine/machine / libmachine

afbjorklund:
These are the remaining drivers, to be moved with libmachine:

* errdriver (for tests)
* fakedriver (for tests)
* hyperv
* virtualbox

The "commands" and "version" directories should not be needed.

[...]

PR #22369, opened 2026-01-02
Update minikube-machine drivers

afbjorklund:
This moves the driver helpers to inside libmachine, so that they don't conflict with minikube drivers.

i.e. avoids a conflict between none and none2 drivers

The libmachine drivers and provisioners are not used in minikube anyway, since they have been forked...

And avoids a naming conflict between net/rpc and rpc

----

* errdriver
* fakedriver
* nodriver
* rpcdriver

```
drivers/errdriver/error.go => libmachine/drivers/errdriver/errdriver.go
{drivers => libmachine/drivers}/fakedriver/fakedriver.go
drivers/none/driver.go +> libmachine/drivers/nodriver/nodriver.go
libmachine/drivers/plugin/register_driver.go
libmachine/drivers/{rpc => rpcdriver}/client_driver.go
libmachine/drivers/{rpc => rpcdriver}/server_driver.go
libmachine/drivers/{rpc => rpcdriver}/server_driver_test.go
```

Preparation from:

* https://github.com/kubernetes/minikube/pull/21647

* fixes LICENSE (https://github.com/kubernetes/minikube/issues/21619#issuecomment-3699398622)

* fixes migration (none driver was used as the base for importing all driver configuration)

----

NOTE: None of these are real drivers, they are used as part of the testing and implementation.

That is why they are moved out of the "drivers" directory, which is where all the real drivers are.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -259,6 +259,6 @@ require (
 
 replace (
 	github.com/Parallels/docker-machine-parallels/v2 => github.com/minikube-machine/machine-driver-parallels/v2 v2.0.2-0.20240730142131-ada9375ea417
-	github.com/docker/machine => github.com/minikube-machine/machine v0.0.0-20251212115451-c0ae03d5e398
+	github.com/docker/machine => github.com/minikube-machine/machine v0.0.0-20260102125217-0c3490c1962a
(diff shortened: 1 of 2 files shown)


medyagh:
thank you for this PR !

PR #22383, opened 2026-01-04
fix: make html_report gopogh path

### LINKED ISSUE #21914: install gopogh if not installed , when you run make htm_report
we have a make target that runs Gopogh and opens the html report

"make html_report"

Market should check if gopogh is not in PATH and if it is not it should Tell user that is not in PATH and try to install it .

bhavyaBeliever:
Fixes #21914


bhavyaBeliever:
This issue can be solved if the users have added there ~/go/bin/ into PATH environment variable.

PR #22392, opened 2026-01-07
translations: add "make prune-translations" to remove stale translations

### LINKED ISSUE #21468: some duplicate contexts are exist(/translations)
Hi 🥝 . I'm contributing to the `translation` folder of Minikube. As a Korean Kubernetes user, I'm working on the `ko.json` file sequentially (L1-100, L101-200, ~ing). While working on this, I encountered <duplicate sentences> and a sentence that I'm unsure of <why they were separated>. 

1. (Based on ko.json) I think sentences `L79 and 80` have the same meaning, except that "a" is added.
> L79: "Amount of time to wait for a service in seconds"
> L80: "Amount of time to wait for service in seconds"

This sentence is also found in json files of several other languages.

2. (Based on ko.json) `L165 - L167` / `L166 - L168` are each duplicates of each other.
> L165: "Deletes a local Kubernetes cluster"
> L167: "Deletes a local Kubernetes cluster"

> L166: "Deletes a local Kubernetes cluster.
> L168: "Deletes a local Kubernetes cluster.

This sentence is also found in json files of several other languages.

3. (Based on ko.json) `L184 and 185` are identical sentences except for the period (.).
> L184: "Display values currently set in the minikube config file"
> L185: "Display values currently set in the minikube config file."

This sentence is also found in json files of several other languages.

---

## Comment(for maintainer)

If you could give me feedback on each number, I'd be happy to edit the JSON files for all languages ​​to align with those guidelines. For number 1, you could probably choose between L79 and 80. For number 2, since it's redundant, you could simply delete one of them. For number 3, since the preceding sentences in a similar context don't use periods (.), I think L184 would be a good choice. 

Your feedback would be greatly appreciated! Thank you!! :)

developowl:
fixes #21468 

# Problem(Issue)
Each language.json contained duplicate and orphan keys (keys that were deleted from strings.txt but still persisted in the JSON files).

# Proposed Solution
- **Validation**: Compare `strings.txt` with each language JSON file to identify duplicates and orphan keys.
- **Data Flow Audit**: Verified the synchronization process from `strings.txt` to individual translation files.
- **Data Integrity**: Ensured that existing, valid `string: translation` pairs are preserved during the cleanup process.
- **Strict Matching**: Only 100% exact duplicates are targeted for removal to avoid false positives from minor differences (e.g., punctuation or casing).
- **Implementation**: 
  - Developed the pruning logic under the `cmd/` directory.
  - Added a new `Makefile` target for easy execution.
- **Verification**: Completed local testing to ensure the logic works as expected without regressions.

# Implementation Details
## Analyzing the current system architecture.

- `extract.go` : extract logic file
- `writeStringsToFiles`: strings to file function

### 🚨 Root Cause
```
// Remove translations from the file that are empty and were not extracted
for k, v := range currentTranslations {
			if _, ok := e.translations[k]; !ok && len(v.(string)) == 0 {
				delete(currentTranslations, k)
			}
		}
```

- A key is only deleted when both of the following conditions are met:
1. `!ok` : The key is missing from e.translations (it is no longer used in the current source code).
2. `len == 0` : The translation value is an empty string.

### Data Retention Strategy
To ensure that we don't lose any existing translations, I've implemented a strict pruning logic. As shown in the table below, even if a string is no longer detected in the source code (!ok), it is **preserved** if it already has a translated value. Only unused strings with empty values are removed.

Screenshot 2026-01-07 at 19 03 22

# Solution 1) Easy version (I don't choose.)

```
// Remove translations from the file that are empty and were not extracted
		
		for k, v := range currentTranslations {
			if _, ok := e.translations[k]; !ok {
				delete(currentTranslations, k)
			}
		}
```

- Immediate deletion if the key is missing from e.translations.
- Disadvantage: Requires caution when running extract (potential safety/data loss issues).
- A single command performs too many roles (violates the Single Responsibility Principle).

# Solution 2) create `prune-translations` & `update-translations` also, removed the deletion logic from the extract

```
extract: 
prune-translations:   
update-translations:  
    $(MAKE) extract
    $(MAKE) prune-translations
```

Adhering to the Single Responsibility Principle (SRP):
- `extract`: Handles only string extraction from the source code.
- `prune-translations`: Handles only the cleanup (pruning) of obsolete or orphan keys.
- `update-translations`: A wrapper command that performs the entire workflow (extraction and pruning) at once.

## prune-translations Job

1. Extract keys from strings.txt.
2. Compare the extracted keys with the keys in each locale-specific JSON file.
3. Remove duplicate or obsolete (deleted) keys from each JSON file.
4. Compare the final `key counts` of each JSON file with strings.txt after the cleanup.

## Add `prune-translations`, `update-translations` target into Makefile

```
.PHONY: prune-translations
prune-translations: ## remove stale translations that no longer exist in strings.txt
	go run cmd/prune-translations/prune-translations.go

.PHONY: update-translations
update-translations: extract prune-translations ## extract strings and remove stale translations

```

# Test!
before test(ko json)

- Before test
- Example: ko.json
- Duplicate key: L168 & L170

## Test Case 1) `extract` -> `prune-translations`
extract and prune

## Test Case 2) Only `update-translations`
update

### After Test
after test(ko json)

- Example: `ko.json`
- Duplicate key is not availabler.

---

### To maintainer.

@medyagh 
My apologies for the delay due to personal reasons. I have remained committed to this issue and have finally completed the task. While I considered Solution 1) as a simpler alternative, I felt it assigned too many responsibilities to the extract command alone. I would appreciate it if you could positively review the `prune-translations` and `update-translations` commands I have implemented. Thank you for your patience.🥝

### DIFF
--- a/cmd/prune-translations/prune-translations.go
+++ b/cmd/prune-translations/prune-translations.go
@@ -93,7 +93,11 @@ func pruneFile(path string, validKeys map[string]interface{}) (PruneResult, erro
 		if err != nil {
 			return PruneResult{}, err
 		}
-		if err := os.WriteFile(path, append(output, '\n'), 0644); err != nil {
+		info, err := os.Stat(path)
+		if err != nil {
+			return PruneResult{}, err
+		}
+		if err := os.WriteFile(path, append(output, '\n'), info.Mode()); err != nil {
 			return PruneResult{}, err
 		}
 	}


developowl:
@medyagh Thank you for the feedback!🥝 I've reflected all your reviews.
Please check this [commit](https://github.com/kubernetes/minikube/pull/22392/changes/529612081f4d1425b69d08158b5457208f174ba8), which changes `generate-docs` dependency from `extract` to `update-translations`. So it now include both extracting strings(`extract`) and pruning(`prune-translations`) 
Thank you :)

cmd/prune-translations/prune-translations.go
```diff
@@ -0,0 +1,71 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// This tool removes stale translations (keys not in strings.txt) from language JSON files.
+// Usage: go run cmd/prune-translations/prune-translations.go (from minikube root)
+
+package main
```
medyagh, line 20: how about add a unit test for this , and you can make dummy test data and expect it to to do the right thing, and make sure we dont ever delete real translations by mistake

developowl: @medyagh 
I wrote a unit test. I added dummy data to each unit test function. Is this correct? Please review my unit test. :)

developowl (commented):
I wrote a unit test. I added dummy data to each unit test function. Is this correct? Please review my unit test. :)

developowl:
@medyagh 
More information of unit test. 🥝

**Test cases:**
- `TestPruneTranslations_RemovesStaleKeys` - Verifies stale keys are correctly removed
- `TestPruneTranslations_PreservesValidKeys` - Ensures valid keys are never deleted
- `TestPruneTranslations_MultipleFiles` - Tests processing of multiple translation files
- `TestPruneTranslations_EmptyTranslationFile` - Handles empty JSON files
- `TestPruneTranslations_MissingStringsFile` - Error handling for missing strings.txt
- `TestPruneTranslations_InvalidStringsJSON` - Error handling for invalid JSON
- `TestPruneTranslations_DoesNotTouchRealTranslations` - Safety check to ensure tests use isolated temp directories

All tests use `t.TempDir()` with dummy test data, ensuring real translation files are never modified.

**Run tests:**
```sh
go test -v ./cmd/prune-translations/...

File: cmd/prune-translations/prune-translations.go [hunk 2]
```diff
@@ -0,0 +1,120 @@
...
+		Original: original,
+		Removed:  removed,
+		Kept:     len(translations),
+	}, nil
+}
+
+func main() {
+	results, err := PruneTranslations("translations")
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
medyagh: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/kubernetes/minikube/pull/22392#discussion_r2693115381)

medyagh:
thank you @developowl

PR #22400, opened 2026-01-08
ui: fix correct formatting in tunnel service output

### LINKED ISSUE #22133: CLI Formatting: Missing newline and stray characters in "minikube service" output
### What Happened?

When running minikube service <name>, the output formatting is broken. The status message "Starting tunnel for service..." does not include a newline at the end. As a result, the table header (NAMESPACE, NAME, etc.) is printed on the same line, making it hard to read. There is also a stray / character (likely from the spinner animation) stuck at the end of the message.

minikube version: v1.37.0
Shell: zsh

### Attach the log file

Image

### Operating System

macOS (Default)

### Driver

None

XiaWuSharve:
## Fix formatting issue in `minikube service` tunnel output

**fixes #22133**

### Problem Description
When running `minikube service <service-name>`, as reported in issue #22133, before this fix, the output `Starting tunnel for service <service-name>.` was incorrectly followed by a `/` character without a line break. The table would start printing immediately on the same line. Furthermore, an extra duplicate line `Starting tunnel for service <service-name>.` would appear after the table. The root cause was a race condition: the spinner attached to the status message via `style.Running` used in `pkg/minikube/tunnel/kic/ssh_conn.go` attempted to refresh in place, but the concurrent output of the table moved the cursor, preventing the spinner from overwriting itself. This resulted in the spinner's final state being appended to the end of the output stream.

before

### Solution in This PR
This PR removes the unintended `/` character and ensures proper line breaks. The fix works by changing the output style to one without a spinner (`style.StartingSSH`). This guarantees that the status message is printed completely (ending with a newline `\n`) before the table is rendered, eliminating the cursor conflict and the duplicate line. This directly resolves the readability issue described in #22133.

after

### Verification
- Built and tested locally using: `out/minikube service hello-node`
- The output now correctly shows
- `make test` passed.

### Notes on Future UI Considerations
While this PR is aiming at closing #22133 by fixing the immediate formatting bug, it highlights a potential deeper UI concern: printing the final table *while* tunnels are still starting (a process that can sometimes take over a second) might be suboptimal. A more robust approach could involve either:
1.  Printing the table only after all tunnels have been confirmed established, or
2.  Printing the table first and then dynamically updating it with the connection status of each tunnel.
These enhancements are noted here for future consideration but are outside the scope of this specific formatting fix.

---
**Note to Reviewers**

Hi, this is my first PR to the minikube project. I've tried to follow the contribution guidelines, but I'm sure there might be areas I've overlooked or conventions I'm not yet familiar with. I'd be very grateful for any feedback on the fix itself, the code style, or the PR process. Please don't hesitate to point out anything that could be improved. Thank you!

### DIFF
--- a/pkg/minikube/tunnel/kic/ssh_conn.go
+++ b/pkg/minikube/tunnel/kic/ssh_conn.go
@@ -159,7 +159,7 @@ func createSSHConnWithRandomPorts(name, sshPort, sshKey string, svc *v1.Service)
 
 func (c *sshConn) startAndWait() error {
 	if !c.suppressStdOut {
-		out.Step(style.Running, "Starting tunnel for service {{.service}}.", out.V{"service": c.service})
+		out.Step(style.StartingSSH, "Starting tunnel for service {{.service}}.", out.V{"service": c.service})
 	}
 
 	r, w := io.Pipe()


medyagh:
@XiaWuSharve can you please put Before This PR as well as after to see the difference?

XiaWuSharve:
Hi @medyagh ,

Thank you for the feedback. I have updated the PR description as suggested, adding a clear “Before” and “After” output comparison to showcase the fix for the formatting issue.

Please take a look when you have a moment.

medyagh:
thank you @XiaWuSharve

PR #22401, opened 2026-01-08
Remove minikube support for 32-bit architectures

### LINKED ISSUE #22378: Remove minikube support for legacy 32-bit architectures
Kubernetes dropped support for `arm` (armv7) in version 1.27, released in 2023.

https://kubernetes.io/blog/2023/04/11/kubernetes-v1-27-release/

Raspberry Pi OS (raspbian) changed to the 64-bit OS version by default, in 2022.

https://www.raspberrypi.com/news/raspberry-pi-os-64-bit/

---

So it is no longer needed to build minikube for _any_ arm32, only for arm64.

We can also remove the experimental client-only builds for x86 and armv6.
(these were only used to interact with a remote cluster using the ssh driver)

The mainframe targets still remain upstream, but they're less supported...
(the deal is that the binaries are built for the architectures, but **not** tested)

OBS is making packages for architectures: amd64, arm64, ppc64le, s390x

afbjorklund:
Require 64-bit architecture

* Remove retro remote 32-bit arch targets

* Remove legacy 32-bit arm arch targets

Closes #22378

### DIFF
--- a/pkg/drivers/kic/types.go
+++ b/pkg/drivers/kic/types.go
@@ -24,10 +24,10 @@ import (
 
 const (
 	// Version is the current version of kic
-	Version = "v0.0.48-1767429414-22376"
+	Version = "v0.0.48-1767944074-22401"
 
 	// SHA of the kic base image
-	baseImageSHA = "8522127cc955c7239f09dcfbdedbbaf07caea2e2fc8a14cb55fe332ed7496532"
+	baseImageSHA = "5af296c365892fa7c4c61cd02bf3cdb33e2c362939e717d7686924b3b3f07773"
 	// The name of the GCR kicbase repository
 	gcrRepo = "gcr.io/k8s-minikube/kicbase-builds"
 	// The name of the Dockerhub kicbase repository
--- a/site/content/en/docs/commands/start.md
+++ b/site/content/en/docs/commands/start.md
@@ -27,7 +27,7 @@ minikube start [flags]
       --apiserver-port int                The apiserver listening port (default 8443)
       --auto-pause-interval duration      Duration of inactivity before the minikube VM is paused (default 1m0s) (default 1m0s)
       --auto-update-drivers               If set, automatically updates drivers to the latest version. Defaults to true. (default true)
-      --base-image string                 The base image to use for docker/podman drivers. Intended for local development. (default "gcr.io/k8s-minikube/kicbase-builds:v0.0.48-1767429414-22376@sha256:8522127cc955c7239f09dcfbdedbbaf07caea2e2fc8a14cb55fe332ed7496532")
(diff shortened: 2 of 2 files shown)


pkg/minikube/detect/detect.go
```diff
@@ -41,18 +41,7 @@ func RuntimeOS() string {
...
-		if !cpu.ARM.HasVFP {
-			return "arm/v5"
-		}
-		if !cpu.ARM.HasVFPv3 {
-			return "arm/v6"
-		}
-		// "arm" (== "arm/v7")
-	}
-	return arch
+	return runtime.GOARCH
```
nirs, line 43: We can change all calls to use runtime.GOARCH

nirs (approved):
Good idea, minikube makefile looked like a computer museum. I don't see any used for minikube on legacy machines.

Need to fix the lint issues.

pkg/minikube/detect/detect.go
medyagh: plz fix this lint 

Error: pkg/minikube/detect/detect.go:31:2: "golang.org/x/sys/cpu" imported and not used (typecheck)
/*

and other lints

afbjorklund:
> Good idea, minikube makefile looked like a computer museum.

The Raspberry Pi OS was defaulting to 32-bit for a bit longer than other platforms, because they wanted one OS for all

So even the newer models that were capable of 64-bit came with 32-bit OS, but eventually the old models faded out...

> I don't see any used for minikube on legacy machines.

Unfortunately Kubernetes is now too big for the Raspberry Pi, unless you go for the more PC-like versions (with fans, etc)

You can still build a cluster with them, but you are better off using k3s and something else than minikube to do it

pkg/minikube/detect/detect.go
afbjorklund: Sure, but that refactoring can come separately

If so might want to do the same for RuntimeOS

site/content/en/docs/start/_index.md
```diff
@@ -66,7 +65,7 @@ Click on the buttons that describe your target platform. For other architectures
 {{% /quiz_row %}}
 
 {{% quiz_row base="/Linux" name="Architecture" %}}
-{{% quiz_button option="x86-64" %}} {{% quiz_button option="ARM64" %}} {{% quiz_button option="ARMv7" %}} {{% quiz_button option="ppc64" %}} {{% quiz_button option="S390x" %}}
+{{% quiz_button option="x86-64" %}} {{% quiz_button option="ARM64" %}} {{% quiz_button option="ppc64" %}} {{% quiz_button option="S390x" %}}
```
afbjorklund, line 68: Wonder why this says "ARM64" when all the other architectures are lowercase?

PR #22407, opened 2026-01-08
fix/etcd flaky test

### LINKED ISSUE #22408: fix(integration): Flaky test in OLM addon for etcd readiness
The 'validateOlmAddon' integration test can be flaky when checking for the etcd operator's readiness. It uses a custom retry loop that can fail intermittently. This should be replaced with a more robust 'kubectl wait' command.

aka1976mb:
The 'validateOlmAddon' integration test was using a custom retry loop
with 'retry.Expo' to check if the etcd operator's CSV had been
successfully installed. This could lead to intermittent failures (flakes)
in CI.

This change replaces the custom Go retry loop with a single, more
idiomatic 'kubectl wait' command. The command waits for the CSV's
status phase to become 'Succeeded', using 'kubectl''s built-in polling
mechanism.

This makes the test more robust and less prone to flakiness.

---

Fixes #22408


nirs:
@aka1976mb Can you rebase on master and remove the merge commits?

Issue #22360, opened 2025-12-29
Allow minikube ssh login as root instead of as docker


afbjorklund:
Currently we are changing the permissions or the group of the "docker" socket in minikube.

This is to match the old "docker" group, that used to control `/var/run/docker.sock` access.

But it is **not** the default setting of Podman or containerd, that prefer to connect as root instead...

When no longer using the legacy TCP socket, it would be possible to ssh root@ instead of docker@

* https://github.com/kubernetes/minikube/issues/9229

It is mostly a matter of adding /root/.ssh/authorized_keys in addition to /home/docker/.ssh/authorized_keys

Then the docker (group) and podman configuration could back to the default, which requires sudo or root.

https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user

"The Docker daemon binds to a Unix socket, not a TCP port. **By default it's the root user that owns the Unix socket, and other users can only access it using sudo.** The Docker daemon always runs as the root user." (This is assuming kubernetes and not usernetes, in the future it might be possible to run rootless...)

You can still use `sudo podman` or `sudo nerdctl`, if not using podman system service or nerdctld.

It is just that if using a tunneled socket instead of a command line, then there is no place for "sudo"

----

The minikube-automount does this: (when mounting the disk image)

https://github.com/kubernetes/minikube/blob/master/deploy/iso/minikube-iso/package/automount/minikube-automount#L150 (from https://github.com/boot2docker/boot2docker/blob/master/files/init.d/autoformat#L147)

```bash
    if [ -e "/userdata.tar" ]; then
        mv /userdata.tar /var/lib/boot2docker/
    fi

    tar xf /var/lib/boot2docker/userdata.tar -C /home/docker/
    chown -R docker:docker /home/docker/.ssh
```

Where userdata.tar contains the keys. (when creating the disk image)

https://github.com/minikube-machine/machine/blob/main/libmachine/mcnutils/b2d.go#L498
(similar code has been copied/pasted into various libmachine drivers, to the same effect)

```go
        // .ssh/key.pub => authorized_keys
        file = &tar.Header{Name: ".ssh", Typeflag: tar.TypeDir, Mode: 0700}
        if err := tw.WriteHeader(file); err != nil {
                return err
        }
        pubKey, err := os.ReadFile(d.publicSSHKeyPath())
        if err != nil {
                return err
        }
[...]

PR #22434, opened 2026-01-11
ISO: Use the minikube version of crun buildroot package

afbjorklund:
The package was renamed from "crun" to "crun-latest", but the dependencies were never updated to match...

commit 0191b983e35d9e05e975e178b3fff39ec478cf9f

So it was using crun 1.18.2 from buildroot 2025.02.x, instead of crun 1.26 in the minikube external packages.

```
BR2_PACKAGE_CRUN=y
# BR2_PACKAGE_CRUN_LATEST is not set
```

* https://github.com/buildroot/buildroot/commit/7bb88682e8 package/crun: bump to version 1.18.2
* https://github.com/buildroot/buildroot/commit/531fc0a60e package/crun: bump to version 1.15
* https://github.com/buildroot/buildroot/commit/ea59fd8827 package/crun: bump to version 1.12
* https://github.com/buildroot/buildroot/commit/0f1d33b6c8 package/crun: bump to version 1.8.4
* https://github.com/buildroot/buildroot/commit/6987b92da5 package/crun: bump to version 1.7.2
* https://github.com/buildroot/buildroot/commit/530d6f661e package/crun: new package

EDIT: Apparently crun-latest was never tested after the rename either, since the package failed to build

It was trying to download https://github.com/containers/crun/releases/download/1.26/crun-latest-1.26.tar.gz

* 64de71e74 Kicbase/ISO: Update crun from 1.23.1 to 1.26
* 03a3c91ab Kicbase/ISO: Update crun from 1.23 to 1.23.1
* 0afbdc537 Kicbase/ISO: Update crun from 1.19 to 1.23
* 7e1077f88 Kicbase/ISO: Update crun from 1.18.2 to 1.19
* 7cf70409c Kicbase/ISO: Update crun from 1.18 to 1.18.2
* 79ad8801c Kicbase/ISO: Update crun from 1.17 to 1.18
* b0b1a5d3f Kicbase/ISO: Update crun from 1.16.1 to 1.17
* b7534896e Kicbase/ISO: Update crun from 1.16 to 1.16.1
* 820e51520 Kicbase/ISO: Update crun from 1.15 to 1.16
* 3afdff0f3 Kicbase/ISO: Update crun from 1.14.4 to 1.15
* f0f00dbda Kicbase/ISO: Update crun from 1.14 to 1.14.4
* 84691f8d1 Kicbase/ISO: Update crun from v1.2 to v1.14
* 0191b983e iso/package: Move crun to crun-latest
[...]

PR #22435, opened 2026-01-11
ISO: Rename podman buildroot package to podman-legacy

### LINKED ISSUE #22105: Rename the "sysdig" and "podman" packages in ISO buildroot

afbjorklund:
To avoid conflict with the new buildroot 2025.05 podman package
 It could have been called podman-master, which would be ironic.

But it _should_ be updated from v3.4 to v4.9*, possibly with CNI
support enabled to avoid having to build netavark (with rust).

Closes #22105

\* it would still be podman-legacy, since both 3.4 and 4.9 are EOL

https://endoflife.date/podman

But it doesn't have to _run_ things, only work with the cri-o images?

So 5.x (quadlet) is not so important.

https://www.redhat.com/en/blog/podman-new-network-stack

* https://github.com/containers/podman/issues/25730

----

In order to switch to BR "podman", the socket permissions need fixing:

* https://github.com/kubernetes/minikube/issues/22360

This package adds a "podman" group that works like the "docker" group:

* https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user


afbjorklund:
ok-to-build-iso

afbjorklund:
Same jenkins issue as in https://github.com/kubernetes/minikube/pull/22434#issuecomment-3763674794

Issue #21841, opened 2025-11-03
Deprecate the Docker container runtime and merge with containerd


afbjorklund:
The "docker" container runtime does not have full support for new Kubernetes features.

The dockershim successor since v1.24, **cri-dockerd**, is not fully maintained and supported.

https://github.com/Mirantis/cri-dockerd

But the "containerd" container runtime does not have support for the Docker REST API.

Instead there is a small middleware, **nerdctld**, that is translating API calls to the nerdctl CLI.

https://github.com/afbjorklund/nerdctld

---

The best would be if the supported container runtime could run the supported daemon.

Then you wouldn't need to have cri-dockerd and nerdctld, but only containerd and docker.

The dockerd (daemon) would only be started on-demand ([socket activated](https://github.com/moby/moby/tree/master/contrib/init/systemd)), not by default.

When using `minikube image`, there is no need for docker - it only needs containerd/buildkit.

---

The docker snapshotter is replaced with containerd, and docker build is replaced with buildkit.

This change has [already happened in Docker Desktop](https://docs.docker.com/desktop/features/containerd/), but it is still optional in Docker Engine.

https://docs.docker.com/engine/storage/containerd/ (overlay2 -> io.containerd.snapshotter.v1)

The "docker buildx" plugin uses BuildKit, and it has replaced the legacy "docker build" builder.

Currently minikube is reverting to the legacy "overlay2" graph driver, in the Docker configuration.

* https://github.com/kubernetes/minikube/issues/22057

---

By configuring Docker to use the "k8s.io" namespace instead of "moby", it sees the _same_ images.

This is similar to how Podman sees the same images as cri-o, since it only has a single namespace.

This way, the feature of `minikube docker-env` building images for Kubernetes is still preserved...

```json
{
  "features": {
    "containerd-snapshotter": true
  },
  "containerd-namespace": "k8s.io"
}
```

NOTE: This changes the storage directory used, from `/var/lib/docker` to `/var/lib/containerd`

medyagh:
that sounds like a great plan ! and we should have that in our road map

afbjorklund:
[...]

PR #18530, opened 2024-03-28
Clean up the nerdctld installation

afbjorklund:
Use the regular upstream names for socket and service.

Don't use world-writable permissions for the binaries.

The socket starts the service when it is activated.

The units go in the installation and not in config.

----

Cleanup from

508adea11f2d68c4fd3a55f62c5b8d975369d938

PR #22438, opened 2026-01-11
kicbase: Clean up the nerdctld installation

afbjorklund:
Use a "nerdctl" group and restore the units.

* Use the regular upstream names for socket and service.
    
* Don't use world-writable permissions for the binaries.

This is an updated version of the previous:

* #18530

* #15452

---

Eventually the "nerdctl" and "podman" groups should be removed, and the commands run as root instead:

* #22360

The current setup is a legacy from docker, while containerd and crio require the use of `sudo` and not a group.

Since you can't run sudo over ssh, the solution is to login as root instead of as the regular user (with the group).

Using `nerdctld` is now deprecated, eventually the same approach ("k8s.io" namespace) will be used for `dockerd`:

* #21841

### DIFF
--- a/deploy/kicbase/Dockerfile
+++ b/deploy/kicbase/Dockerfile
@@ -136,6 +136,12 @@ EOF
 RUN export ARCH=$(dpkg --print-architecture | sed 's/ppc64el/ppc64le/') \
     echo "Installing nerdctl and nerdctld ..." &&\
     addgroup --system nerdctl &&\
+    mkdir -p /etc/systemd/system/nerdctl.socket.d && \
+    printf "[Socket]\nSocketMode=0660\nSocketGroup=nerdctl\n" \
+           > /etc/systemd/system/nerdctl.socket.d/override.conf && \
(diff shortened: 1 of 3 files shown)


afbjorklund:
Jenkins is out of free disk space:

```
xorriso : FAILURE : Image size 189600s exceeds free space on media 46416s
libisofs: MISHAP : Image write cancelled
xorriso : NOTE : -return_with SORRY 32 triggered by problem severity FAILURE
make[2]: *** [fs/iso9660/iso9660.mk:216: /home/jenkins/workspace/iso-pr-build/out/buildroot/output-x86_64/images/rootfs.iso9660] Error 32
make[1]: *** [Makefile:83: _all] Error 2
```

But there were no ISO changes?

There is no installation of nerdctld in the buildroot environment yet, afaik.

afbjorklund:
ok-to-build-image

afbjorklund:
The installation only ran for amd64 and arm64, so the setup was failing on ppc64le and s390x.

* https://github.com/kubernetes/minikube/pull/19730

PR #22447, opened 2026-01-13
test: Minimal fix for TestMiniPath on Windows

mloskot:
Following up the discussion during latest office hours,
here is a minimal fix as alternative to PR #21739.

The premise is this:
- Both, backslash `\` and slash `/`, are valid path separators on Windows
- Current implementation of `MiniPath` function is correct
- Path normalisation is needed only at point of direct (string) comparison
  in order to avoid OS-specific separators

Before this fix, the `TestMiniPath` was failing with:

```
--- FAIL: TestMiniPath (0.00s)
    --- FAIL: TestMiniPath//tmp/.minikube (0.00s)
        localpath_test.go:78: MiniPath expected to return '\tmp\.minikube', but got '/tmp/.minikube'
```

That is, before the fix, `expectedPath` was returned as `\tmp\.minikube` as it was turned into Windows-native path by `filepath.Join`, whereas `MiniPath` returned `/tmp/.minikube` as it returned vanilla value of `MINIKUBE_HOME`.

### DIFF
--- a/pkg/minikube/localpath/localpath_test.go
+++ b/pkg/minikube/localpath/localpath_test.go
@@ -71,9 +71,9 @@ func TestMiniPath(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		t.Run(tc.env, func(t *testing.T) {
-			expectedPath := filepath.Join(tc.basePath, ".minikube")
 			t.Setenv(MinikubeHome, tc.env)
-			path := MiniPath()
+			expectedPath := filepath.Clean(filepath.Join(tc.basePath, ".minikube"))
+			path := filepath.Clean(MiniPath())
 			if path != expectedPath {
 				t.Errorf("MiniPath expected to return '%s', but got '%s'", expectedPath, path)
 			}


pkg/minikube/localpath/localpath_test.go
```diff
@@ -71,9 +71,9 @@ func TestMiniPath(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		t.Run(tc.env, func(t *testing.T) {
-			expectedPath := filepath.Join(tc.basePath, ".minikube")
 			t.Setenv(MinikubeHome, tc.env)
-			path := MiniPath()
+			expectedPath := filepath.Clean(filepath.Join(tc.basePath, ".minikube"))
+			path := filepath.Clean(MiniPath())
```
nirs, line 76: So we don't care how minikube use the path internally, only that the the "cleaned" version of the path matches?

Can you share the values before and after cleaning? Adding an example in the commit message will make this clear.

pkg/minikube/localpath/localpath_test.go
```diff
@@ -71,9 +71,9 @@ func TestMiniPath(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		t.Run(tc.env, func(t *testing.T) {
-			expectedPath := filepath.Join(tc.basePath, ".minikube")
 			t.Setenv(MinikubeHome, tc.env)
-			path := MiniPath()
+			expectedPath := filepath.Clean(filepath.Join(tc.basePath, ".minikube"))
```
medyagh, line 75: please share example values ?

pkg/minikube/localpath/localpath_test.go
mloskot: Yes, AFAIU, that is the current consensus. We don't care about format of paths Minikube uses to pass around internally. However, we may decide on further unification and improvements, for example, using OS-agnostic `path` module.

Sure, I've just updated the commit and the PR description.

pkg/minikube/localpath/localpath_test.go
mloskot: Before, `expectedPath` was returned as `\tmp\.minikube` as it was turned into Windows-native path by `filepath.Join`, whereas `MiniPath` returned `/tmp/.minikube` as it returned vanilla value of `MINIKUBE_HOME`. 

I've also updated the PR description with example of the test output before the fix.

medyagh (approved):
looks good to me

PR #22467, opened 2026-01-16
test: Fix virtualbox driver unit tests failures on windows

bobsira:
**Cause**

- **Mismatch**: On Windows the driver builds OS-native paths with backslashes (e.g. `path\machines\default\id_rsa`), but the test expected signatures use forward slashes. The mock compared raw strings, so the equality check failed and `TestCreateVM `errored with "Unexpected call".

**Fix**

- What: Normalize both the actual and expected call signatures to POSIX-style paths before comparing using `filepath.ToSlash`.
- Result: Test comparisons are platform-agnostic; expected strings remain unchanged and tests pass on Windows and Unix.

This fix also addresses these 4 other virtual box unit test failures:

1. `TestCreateVMWithSpecificNatNicType`
2. `TestCreateVMWithoutAccelerate3D`
3. `TestStart`
4. `TestStartWithHostOnlyAdapterCreationBug`

before

before3

before2

before1

before

**Test Run with the fix**
after

after3

after2

after1

after

### DIFF
--- a/pkg/drivers/virtualbox/virtualbox_test.go
+++ b/pkg/drivers/virtualbox/virtualbox_test.go
@@ -499,8 +499,8 @@ func (v *MockCreateOperations) doCall(callSignature string) (string, error) {
 
 	call := v.expectedCalls[v.call]
 	if call.signature != "IGNORE CALL" {
-		actual := filepath.ToSlash(callSignature)
-		expected := filepath.ToSlash(call.signature)
+		actual := filepath.Clean(callSignature)
+		expected := filepath.Clean(call.signature)
 		if actual != expected {
 			v.test.Fatal("Unexpected call", callSignature)
 		}


pkg/drivers/virtualbox/virtualbox_test.go
```diff
@@ -497,8 +498,12 @@ func (v *MockCreateOperations) doCall(callSignature string) (string, error) {
 	}
 
 	call := v.expectedCalls[v.call]
-	if call.signature != "IGNORE CALL" && (callSignature != call.signature) {
-		v.test.Fatal("Unexpected call", callSignature)
+	if call.signature != "IGNORE CALL" {
+		actual := filepath.ToSlash(callSignature)
+		expected := filepath.ToSlash(call.signature)
```
mloskot: In PR #22447, I applied similar kind of fix but with `filepath.Clean`. Although both, `filepath.Clean` and `filepath.ToSlash` are valid tools here, I think, it is a good idea to unify the 'technique' across the source code.

What do you think @bobsira @medyagh ..., is there a point in following a common convention?

If that is relevant, then either I will refine the PR #22447 with `filepath.ToSlash` or @bobsira `s/ToSlash/Clean/` :) and then all other similar places can follow.

medyagh: @bobsira I think we should follow the example we did in @mloskot PR for a unified approach please read the comments in that PR

bobsira: Sound! I'll use filepath.Clean here as well

bobsira: Done!

Issue #22390, opened 2026-01-06
Can't minikube mount a folder to a VM on macos using podman as a driver


eddienubes:
### What Happened?

I followed through the initial setup page with
```bash
podman machine init --cpus 2 --memory 2048 --disk-size 20
podman machine set --rootful
podman machine start

minikube start --driver=podman --container-runtime=cri-o
```

Then tried to mount a folder with
```bash
minikube mount ./:/app
```
and got 
```
❌  Exiting due to IF_HOST_IP: Error getting the host IP address to use from within the VM: RoutableHostIPFromInside is currently only implemented for linux
```
The error is pretty much self-explanatory, however I didn't find similar opened issues/feature requests of people using podman.

### Attach the log file

[logs.txt](https://github.com/user-attachments/files/24455913/logs.txt)

### Operating System

macOS (Default)

### Driver

Podman

afbjorklund:
The container (KIC) drivers can bind-mount directories from the (VM) host, using another syntax (`--mount` parameter)

Then you can use your containers.conf (from the Podman Machine settings) to control what is mounted from the host

By default, `podman machine init` mounts the entire home folder (that is, under `/Users`):

`  -v, --volume stringArray     Volumes to mount, source:target (default [$HOME:$HOME])`

EDIT: This feature could probably need some documenting at https://minikube.sigs.k8s.io/docs/handbook/mount/

I think the Fedora OS doesn't support the 9p kernel driver, so `minikube mount` probably doesn't work anyway?

`minikube start`
```
    --mount-string='':
	Directory to mount in the guest using format '/host-path:/guest-path'.
```

Abhigyan-Shekhar:
Thanks for the context — this is helpful.

You’re right that KIC drivers can bind-mount directories via the `--mount-string` flag, and that Podman Machine mounts `$HOME` by default via `containers.conf`. However, this issue specifically affects the `minikube mount` command, which relies on resolving the host IP from inside the container.

Currently on macOS with the Podman driver, `minikube mount` fails early because `RoutableHostIPFromInside` is not implemented for Podman on darwin, resulting in the error:
> RoutableHostIPFromInside is currently only implemented for linux

The proposed fix does not change mount mechanisms or Podman Machine defaults — it only enables correct host IP resolution on macOS using `host.containers.internal`, mirroring Docker’s macOS behavior. This allows `minikube mount` to function as intended when users choose to use it.

[...]

Issue #21973, opened 2025-11-23
[Parent]: Make containerd default container runtime


medyagh:
This is to track all the sub issues for all the work needed to change minikube default runtime from cri-dockerd to containerd.

afbjorklund:
Steps:
1) Change the "--container-runtime" default, from docker to containerd
2) ~~Replace the current nerdctld installation with a dockerd installation~~
3) ~~Deprecate the docker container runtime option and the docker preload~~
4) Clean up all the dockershim (<1.24) code that is littering the code base
5) ~~Remove the option to use docker, or alias it to use containerd instead~~
6) ~~Don't have to maintain cri-dockerd and nerdctld in minikube, anymore~~

EDIT: The other optional steps, are detailed in this issue:

* https://github.com/kubernetes/minikube/issues/21841

All references to podman here are just examples, it is used for the **CRI-O** runtime.
The containerd runtime will still keep the (low-level) `ctr` and `buildctl` tools too...

The main difference is that `docker-env` works remotely, the others have to use `ssh`:
`minikube ssh sudo nerdctl` (need to change the containerd namespace to "k8s.io")

----

But changing the default runtime for minikube is a one-line change:

```go
// defaultRuntime returns the default container runtime
func defaultRuntime() string {
        // minikube default
        return constants.Docker
}
```

Then we can remove Docker container runtime, but keep Docker driver.

```go
        // Containerd is the default name and spelling for the containerd container runtime
        Containerd = "containerd"
        // CRIO is the default name and spelling for the cri-o container runtime
        CRIO = "crio"
        // Docker is the default name and spelling for the docker container runtime
        Docker = "docker"
        // DefaultContainerRuntime is our default container runtime
        DefaultContainerRuntime = ""
```

The default is just that, and it can reference _either_ of docker or containerd

With this issue, it will change (but existing clusters will keep their current one)

afbjorklund:
You will also need to update the contributor guide and the process:

https://minikube.sigs.k8s.io/docs/contrib/guide/#contributing-larger-changes

Specifically it is referencing the "MEP", that is no longer being used:

https://github.com/kubernetes/minikube/tree/master/enhancements

medyagh:
@afbjorklund 
I dont think we need to remove cri-dockerd from minikube, it can be one of our available runtimes, just like crio.
changing the default shouldnt have to mean excluding cri-dockerd.

medyagh:
[...]

PR #22495, opened 2026-01-17
docs: Clarify container runtime choice

mloskot:
Improve the originally laconic guidance clarifying the default runtime
and explaining that specific drivers may require or recommend alternative.

The updated guidance should also help users understand the choice of
container runtime is not clear-cut and may require consideration
depending on the use case.

----

Digression: As a still inexperienced Minikube user, I was looking into #22390 and `minikube start --driver=podman --container-runtime=cri-o` made me wonder, why `cri-o`? Can/Shall I try different runtime? Based on what, OS of my host, driver? ;)

The runtimes pages do not seem to mention conditions of their choice, but some drivers do, like podman:

> It’s recommended to run minikube with the podman driver and CRI-O container runtime (except when using Rootless Podman): (...) For Rootless Podman, it is recommended to set --container-runtime to containerd

but the original section updated in this PR did not seem to suggest users should look for answers on pages of drivers.

### DIFF
--- a/site/content/en/docs/handbook/config.md
+++ b/site/content/en/docs/handbook/config.md
@@ -89,13 +89,17 @@ minikube start --extra-config=kubeadm.ignore-preflight-errors=SystemVerification
 
 ## Runtime configuration
 
-The default container runtime in minikube varies. You can select one explicitly by using:
+The default container runtime in minikube is [Docker]({{<ref "/docs/runtimes/docker">}}).
+
+Depending on specific factors like the driver you choose, there may be recommendations of a preferred container runtime, see [drivers page]({{<ref "/docs/drivers">}}) for details.
+
+You can select container runtime explicitly by using:
 
 ```shell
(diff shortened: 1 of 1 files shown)


medyagh (approved):

site/content/en/docs/handbook/config.md
```diff
@@ -89,13 +89,17 @@ minikube start --extra-config=kubeadm.ignore-preflight-errors=SystemVerification
 
 ## Runtime configuration
 
-The default container runtime in minikube varies. You can select one explicitly by using:
+The default container runtime in minikube is [Docker]({{<ref "/docs/runtimes/docker">}}).
```
medyagh, line 92: starting in version after next we will recommend containerd

mloskot: I see, looks like it's tracked here
- #21973

Can I get a sub-issue there to remember about updating the docs?

PR #22498, opened 2026-01-17
[kic] fix syntax error in docker file

ComradeProgrammer:
[kic] fix syntax error in docker file


afbjorklund:
It was merged with build failure:

* https://github.com/kubernetes/minikube/pull/22438#issuecomment-3761544270

Issue #22483, opened 2026-01-17
Window Integreation: Generate Script to Create Azure VM


medyagh:
we have a Azure account through Sig-Windows the images there are not touched in years and not connecting to jenkins anmymore properly

we should have a Script that Can Create Windows VMs from a Base Image Windows Image and installs everything we need

- docker desktop
- nosleep (or anything we need that prevents monitor to go to sleep_
- latest golang
- choco and other package managers needed

our instances Currently Uses

export AZ_IMAGE=MicrosoftWindowsDesktop:Windows-10:19h2-entn-g2:latest
export AZ_SIZE=Standard_D8s_v3

chhavx1618:
@medyagh do you require a specific tool (terraform, azure cli, etc.) ?

SamiDesan:
@medyagh In my opinion the best decision would be Azure Bicep. It's like Terraform, but specifically for Azure

mloskot:
Before bidding on which tool will be the best to complete this task, I have some open questions and issues to clarify of what needs to be done, and only Minikube team has the knowledge to evaluate the options, I think. (_This does not seem to be a typical 'good first issue' task._)

### What 

VM images (quote: "images there are not touched in years") and then VM or just VM (quote: "Create Windows VMs from a Base Image").

The basic options are:

1. Create VM(s) based one of Azure vanilla images and provision it with the required config/tools on-fly every time a VM is needed (integration tests are triggered).
2. First pipeline periodically builds the golden VM image including the required config/tools and publishes it to Shared Image Gallery (SIG) on Azure. Next, the tests pipeline creates VM(s) from the golden image.

Both options have their pros and cons, to list the basic ones:
- Creating VM and provisioning it on fly allows to use cutting edge versions of tools, but takes much longer; may fail quite often.
- Creating VM from golden image is much quicker and hardly never fails as there is no post-boot config/setup involved, but golden images (their content) may become outdated depending on build schedule.  

### How

Create and provision VM:
- Azure CLI
- Bicep
- PowerShell
- Terraform

Build VM image using:
- Azure CLI from VM
- Azure Image Builder [built on HashiCorp Packer](https://learn.microsoft.com/en-us/azure/virtual-machines/image-builder-overview#features)) with Packer or JSON template using Azure CLI, PowerShell, Bicep, Terraform
- Packer performs image build on Azure
- PowerShell
[...]

PR #22536, opened 2026-01-24
ci: Build VM image with Windows 11 for CI on Azure

mloskot:
An ongoing work in progress from https://github.com/mloskot/minikube-windows-image with my part of collaboration with @medyagh related these items
- #22483 
- #22534

## Idea

The `README.md` explains more, but the grand idea is to build golden image based on Windows 11 as a common Windows flavour used by Minikube users. The golden image allows to create VM with reproducible environment fast, in 15-20 seconds on Azure. 

## Solution

This solution is dedicated to Azure infrastructure, because it targets Azure space provided by SIG-Cluster-Lifecycle/SIG-Windows, so this PR includes Bicep templates for the four GHA workflows:
1. Deploy Azure Shared Image Gallery (ASIG)
2. Build VM image on Azure Image Builder (AIB) using Packer, and publish it to Azure Shared Image Gallery
3. Deploy Azure Virtual Network required to run VMs
4. Deploy a test Azure Virtual Machine as a test of the whole setup

The idea is to configure GHA for at least the workflows 1-3 and run them on schedule.

The VM image is provisioned with the following
- [x] Hyper-V
- [x] Chocolatey
- [x] Docker Engine (fully non-interactive installation and use of Docker Desktop seems impossible)
- [x] Kubernetes CLI
- [ ] ...

TODO: Complete content of the VM image with Minikube-specific tools that are missing

## Future

- Build common **base golden image** with Windows 11 optimizations applied and bloatware removed
- Based on the base golden image, build two flavours of the VM image: `*-hyperv` and `*-wsl`
- Export the VM image outside Azure, as VHD or Vagrant box, to make it usable with other hypervisors.
- ...

------

# Showcasing this work

This work can be seen in action in my personal:

- https://github.com/mloskot/minikube-windows-image
- https://github.com/mloskot/minikube-windows-image/actions

### DIFF
--- a/hack/windows-ci-image/README.md
+++ b/hack/windows-ci-image/README.md
@@ -1,0 +1,134 @@
+# Build VM image with Windows 11 for CI on Azure
+
+> [!CAUTION]
+> Minikube was granted access to Azure resource group `SIG-CLUSTER-LIFECYCLE-MINIKUBE` managed by SIG-Cluster-Lifecycle/SIG-Windows.
+> We must **NOT** attempt to delete this resource group!
+>
+> Unfortunately, locking the resource group would not help, becuase it also locks any contained resources:
+>
+>    ```bash
+>    az group lock create --lock-type CanNotDelete \
+>       --name "CanNotDelete-${MINIKUBE_AZ_RESOURCE_GROUP}" \
+>       --resource-group "${MINIKUBE_AZ_RESOURCE_GROUP}" \
+>       --notes "Group managed by SIG-Windows"
+>    ```
+>
+
+## Features
+
+The goal of the project is to build Windows-based golden image for Minikube in order
+to create create short-lived ephemeral Azure Virtual Machines used for testing.
+
+The project provides:
+
+1. Bicep templates to create Shared Image Gallery on Azure in given resource group.
+2. Packer templates to build single Virtual Machine image:
+    - based on Windows 11 with Windows settings optimized and bloatware removed
+    - [specialized, not generalized](https://learn.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries#generalized-and-specialized-images), and [shallowly replicated](https://learn.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries#shallow-replication)
+    - provisioned with local administrator (user-provided username (default: `minikubeadmin`) and password)
+    - provisioned with user-provided public SSH key for streamlined authentication
+    - provisioned with [tools required to run Minikube tests](packer/provisioners/))
+
+3. Packer configuration to publish it to the Azure Shared Image Gallery.
+4. Bicep templates to create Azure Virtual Machine from the published image to verify the output of the image build.
+
+## Usage
+
+Help: `make help`
+
+### Configure
+
+1. Set up Azure CLI:
+
+    ```bash
+    az login
+    ```
+
+    or keep the Azure CLI profile in project-specific location (recommended? safety?):
+
+    ```bash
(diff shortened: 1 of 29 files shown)


hack/windows-ci-image/README.md
```diff
@@ -0,0 +1,134 @@
...
+Connect to PowerShell on VM using the insecure SSH key:
+
+```bash
+ssh -i ~/.ssh/minikube-ci.id_ecdsa  -o StrictHostKeyChecking=no minikubeadmin@vm-minikube-ci.${AZURE_DEFAULTS_LOCATION}.cloudapp.azure.com
+```
+
+Connect to PowerShell on VM using password:
+
+```bash
+export SSHPASS='M!n!kub34dm!n'
```
medyagh: lets not have any password at all (this will trigger CVE hunting ppl) even though this is local for testing, we can replace this 
export SSHPASS=$ENV_NAME...

medyagh (changes requested):

medyagh:
@mloskot also lets run hack/boilerplate/fix.sh

hack/windows-ci-image/README.md
mloskot: Done

mloskot:
@bobsira I have just looked at your provisioning scripts from https://github.com/bobsira/windows-node-image-builder and I think the only piece that we may reuse in future is the containerd installation, when we decide on building the Windows 11-based image flavour with different container runtime than Docker, which we may, of course. Your image has a bit different purpose: it is image with vanilla Kubernetes, while the goal of this PR is to build image for Minikube. Please, correct me if I'm wrong.

mloskot:
https://github.com/kubernetes/minikube/pull/22536#issuecomment-3795817610
> also lets run hack/boilerplate/fix.sh

@medyagh I will need to learn more. From reading the script, I'm not sure how it applies here. If it does, then it will have to be ported to PowerShell (I can do it, of course).

medyagh (changes requested):
@mloskot I think have 4 different workflows is a bit confusing to build the image, I would prefer that we have One Workflow that we call and those others be steps of the Main Workflow, 

could we merge all 4 workflows into windows-ci-image-build.yaml

and also have a one Line Env that we can easily Bump the version of it

mloskot:
> I think have 4 different workflows is a bit confusing to build the image
> could we merge all 4 workflows into windows-ci-image-build.yaml

@medyagh Yes. No. :) I am not able to design the final GHA workflow without mutual understanding how you you want it to work and how it can be done. That is why I included separate workflows in this PR to make it clearer during review w.r.t. what Azure resources are required, what are their lifecycle  and what are the dependencies between the resources:

1. SIG + image definition are a prerequisite for image build using Packer. Those are created once and persist.
2. Actual image versions are output of Packer build and are published to pre-existing SIG.
3. Virtual network is required to create VM-s and, ideally, it should persists and not be re-created for every VM

and so on.

mloskot:
@medyagh I've refactored the four GHA workflows into a single workflow. The jobs creating SIG and VNet are idempotent i.e. Azure resources are created only if they do not exist. We may want to adjust some names of the Azure resources, etc. but it works, and ready for further tweaks, if any necessary:

image

You can see it in action here https://github.com/mloskot/minikube-windows-image/actions 

Once that DGA of the workflow above completes, here are the Azure resources that get created and persist i.e. SIG + VM images + VNet, but the test VM is gone:

image

----

> [!NOTE]
> The **image is still quite useless** as it is provisioned with the Docker Engine.  I will work on replacing it with the Docker in WSL over next days.

Issue #22509, opened 2026-01-21
make "docker-env" on containerd stable


medyagh:
currently we warn users
		out.WarningT("Using the docker-env command with the containerd runtime is a highly experimental feature, please provide feedback or contribute to make it better")

afbjorklund:
This needs more details, on what is actually required before it is considered non-experimental?

As long as it is using nerdctld, it will _always_ be experimental... The alternative is to use dockerd:

* https://github.com/kubernetes/minikube/issues/21841

Then it will run the real docker daemon with containerd storage, and continue to be supported.

----

In order of preference:

1) `minikube image`

2) `minikube ssh`

3) `minikube docker-env` (legacy)

Recently we saw Docker v29 client breaking the support for Podman servers, after an API upgrade...

It has been upgraded to Docker v25 LTS now, but it can very well break again with Docker v30 and up.

afbjorklund:
I think it will be possible to make `load` and `build` still work, assuming DOCKER_BUILDKIT=0 that is...

But the support for running Docker in minikube (no Kubernetes) is something that should be removed.
It is useful as a testing target when developing the VM, but nothing that needs to be supported for users.

https://minikube.sigs.k8s.io/docs/tutorials/docker_desktop_replacement/ - there are other replacements

----

Here is the default response, when building: (this is a known issue, there is no support for running a builder)

```
 => ERROR [internal] booting buildkit                                                                                                                                              10.9s
 => => pulling image moby/buildkit:buildx-stable-1                                                                                                                                 10.7s
 => => creating container buildx_buildkit_default                                                                                                                                   0.1s
------
 > [internal] booting buildkit:
------
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
ERROR: Error response from daemon: 404 page not found
```

Here is the DOCKER_BUILDKIT=0 response: (this is the recommended workaround, for nerdctld and podman)

```
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            BuildKit is currently disabled; enable it by removing the DOCKER_BUILDKIT=0
            environment-variable.

Sending build context to Docker daemon  2.048kB
#1 [internal] load build definition from Dockerfile
...
#5 DONE 0.2s
#6 exporting to image
#6 exporting layers 0.0s done
[...]

Issue #15461, opened 2022-12-03
Deprecate "podman-env", in favor of using "docker-env" with cri-o


afbjorklund:
The podman socket has compatibility support for the Docker API, so prefer using `docker` as the client.

* #15389

No major changes otherwise, just that the variable names are different and more workarounds are needed.

podman-env

```
export CONTAINER_HOST="ssh://docker@127.0.0.1:49162/run/podman/podman.sock"
export CONTAINER_SSHKEY="/home/anders/.minikube/machines/minikube/id_rsa"
export MINIKUBE_ACTIVE_PODMAN="minikube"

# To point your shell to minikube's podman service, run:
# eval $(minikube -p minikube podman-env)
```

docker-env

```
export DOCKER_HOST="ssh://docker@127.0.0.1:49162"
export MINIKUBE_ACTIVE_PODMAN="minikube"

# To point your shell to minikube's docker-daemon, run:
# eval $(minikube -p minikube docker-env)
```

The socket path and identity key now needs to be configured on the node and on the host, explicitly.

afbjorklund:
For this to work, the $DOCKER_HOST must be set on the **remote** end:

```
DOCKER_HOST=unix:///run/podman/podman.sock
```

And as mentioned above, known_hosts and ssh-agent needs updating.

```
minikube ssh-host --append-known
minikube docker-env --ssh-host --ssh-add
```

afbjorklund:
Note that using `minikube image` works already today, without any clients (except `minikube`) or sockets

PR #22547, opened 2026-01-25
Support docker-env with cri-o runtime

afbjorklund:
It is very similar to containerd, but with podman for cri-o instead of nerdctl for containerd. Both using ssh sockets.

* https://github.com/kubernetes/minikube/issues/15461

There are plenty of refactoring opportunities (this is just copy/paste), for merging the nerdctl and podman support.

* https://github.com/kubernetes/minikube/issues/22509

----

This will use BuildKit in a podman container, by default:

```
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
...
 => => pulling image moby/buildkit:buildx-stable-1
 => => creating container buildx_buildkit_default   
```

To switch to the legacy buildah, use DOCKER_BUILDKIT=0:

```
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            BuildKit is currently disabled; enable it by removing the DOCKER_BUILDKIT=0
            environment-variable.
....
```

### DIFF
--- a/cmd/minikube/cmd/docker-env.go
+++ b/cmd/minikube/cmd/docker-env.go
@@ -314,14 +314,20 @@ docker-cli install instructions: https://minikube.sigs.k8s.io/docs/tutorials/doc
 			exit.Message(reason.Usage, err.Error())
 		}
 
-		// for the sake of docker-env command, start nerdctl and nerdctld
-		if cr == constants.Containerd {
-			out.WarningT("Using the docker-env command with the containerd runtime is a highly experimental feature, please provide feedback or contribute to make it better")
+		// for the sake of docker-env command, start nerdctl and nerdctld (or the podman system service, "podmand")
+		if cr == constants.Containerd || cr == constants.CRIO {
+			if cr == constants.Containerd {
+				out.WarningT("Using the docker-env command with the containerd runtime is a highly experimental feature, please provide feedback or contribute to make it better")
 
-			startNerdctld(options)
+				startNerdctld(options)
+			} else if cr == constants.CRIO {
+				out.WarningT("Using the docker-env command with the cri-o runtime is a highly experimental feature, please provide feedback or contribute to make it better")
 
-			// docker-env on containerd depends on nerdctld (https://github.com/afbjorklund/nerdctld) as "docker" daeomn
-			// and nerdctld daemon must be used with ssh connection (it is set in kicbase image's Dockerfile)
+				startPodmand(options)
+			}
+
+			// docker-env on containerd depends on nerdctld (https://github.com/afbjorklund/nerdctld) as "docker" daemon
+			// and nerdctld (or podman) daemon must be used with ssh connection (it is set in kicbase image's Dockerfile)
 			// so directly set --ssh-host --ssh-add to true, even user didn't specify them
 			sshAdd = true
 			sshHost = true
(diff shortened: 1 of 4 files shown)


afbjorklund:
This message is bogus by the way, it is supported also with the podman driver of course:

"the docker-env command only supports the containerd runtime with the docker driver"

cmd/minikube/cmd/docker-env.go
```diff
@@ -315,10 +315,16 @@ docker-cli install instructions: https://minikube.sigs.k8s.io/docs/tutorials/doc
 		}
```
medyagh, line 316: are there are any test we are skipping on crio we can allow now?

afbjorklund: Probably, it seems to be skipped now (hardcoded to docker only)

```go
        if ContainerRuntime() != "docker" {
                t.Skipf("skipping: only runs with docker container runtime, currently testing %s", ContainerRuntime())
        }
```

cmd/minikube/cmd/docker-env.go
```diff
@@ -315,10 +315,16 @@ docker-cli install instructions: https://minikube.sigs.k8s.io/docs/tutorials/doc
...
+
+				startNerdctld(options)
+			} else if cr == constants.CRIO {
+				out.WarningT("Using the docker-env command with the cri-o runtime is a highly experimental feature, please provide feedback or contribute to make it better")
 
-			startNerdctld(options)
+				startPodmand(options)
+			}
 
 			// docker-env on containerd depends on nerdctld (https://github.com/afbjorklund/nerdctld) as "docker" daeomn
```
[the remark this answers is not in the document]
afbjorklund: The comment was not in this commit, though.

cmd/minikube/cmd/docker-env.go
afbjorklund: Oh, the entire test was duplicated for containerd (and only tested with docker driver, not podman)

afbjorklund:
The test is failing, because the old host key has been stored in the image unfortunately.

image

TestDockerEnvCrio

`error during connect: Get "http://docker.example.com/v1.51/version": command [ssh -l docker -p 32773 -o ConnectTimeout=30 -T -- 127.0.0.1 docker system dial-stdio] has exited with exit status 255, make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=Host key verification failed.`

----

EDIT: No idea why it didn't fail for containerd already? Looks the test is missing, for docker-containerd

TestDockerEnvContainerd

afbjorklund:
Aargh, yet another runtime-specific hack:

```go
                        // TODO: refactor to work with docker, temp fix to resolve regression
                        if cr == constants.Containerd {
                                // eventually, run something similar to ssh --append-known
                                appendKnownHelper(nodeName, true)
                        }
```

cmd/minikube/cmd/start.go
```diff
@@ -2100,3 +2100,20 @@ func startNerdctld(options *run.CommandOptions) {
...
+	runner := co.CP.Runner
+
+	// sudo systemctl start podman.socket
+	if rest, err := runner.RunCmd(exec.Command("sudo", "systemctl", "start", "podman.socket")); err != nil {
+		exit.Error(reason.StartPodmand, fmt.Sprintf("Failed to enable podman.socket: %s", rest.Output()), err)
+	}
+
+	// set up environment variable on remote machine. docker client uses 'non-login & non-interactive shell' therefore the only way is to modify .bashrc file of user 'docker'
+	// insert this at 4th line
+	envSetupCommand := exec.Command("/bin/bash", "-c", "sed -i '4i export DOCKER_HOST=unix:///run/podman/podman.sock' .bashrc")
```
bhavyaBeliever, line 2115: @afbjorklund This line always adds `export DOCKER_HOST=unix:///run/podman/podman.sock` in the .bashrc file everytime when `minikube docker-env` gets executed. Can we check by first by `echo $DOCKER_HOST`

bhavyaBeliever: in the PR #22577 for the function startNerdctld we have implemented the above suggestion.
Link for the implementation: [link](https://github.com/kubernetes/minikube/pull/22577/changes#diff-785e0c3520557a4a553e977cac218c88514f688e6f2e01e387236c7ca6dd492fR2098)

afbjorklund: Yeah. as I mentioned it was just copied. Probably would be a good thing to refactor these helpers

Issue #22135, opened 2025-12-13
Merge the packages duplicated by arch in the ISO buildroot


afbjorklund:
The goal is to merge as many of the "arch" packages as possible:

package/
├── automount
├── cni
├── crio-bin
├── crun-latest
├── podman
├── runc-master
├── scheduled-stop
└── sysdig
arch/x86_64/package/
├── buildkit-bin
├── cni-plugins-latest
├── containerd-bin
├── crictl-bin
├── cri-dockerd
├── docker-bin
├── docker-buildx
├── hyperv-daemons
├── nerdctl-bin
└── vbox-mount-service
arch/aarch64/package/
├── buildkit-bin-aarch64
├── cni-plugins-latest-aarch64
├── containerd-bin-aarch64
├── crictl-bin-aarch64
├── cri-dockerd-aarch64
├── docker-bin-aarch64
├── docker-buildx-aarch64
└── nerdctl-bin-aarch64

They only need a couple of ARCH variables, and multiple hash entries...

Then there is no need to duplicate the _entire package_, under "arch"?

It is also possible to have arch-constraints, in the regular package dir.

PR #22548, opened 2026-01-25
ISO: add nerdctld

afbjorklund:
Matching the kicbase installation, and the "podman" package.

The systemd units are from upstream git, but not in tarball.

----

Using the same duplicated install as nerdctl, to be fixed later:

* https://github.com/kubernetes/minikube/issues/22135


afbjorklund:
ok-to-build-iso

afbjorklund:
ok-to-build-iso

PR #22557, opened 2026-01-25
fix: kubevirt addon by removing curl dependency

### LINKED ISSUE #15918: KubeVirt Addon broken due to missing curl in Pod
### What Happened?

Tried `minikube start --addons=kubevirt`

No kubevirt resources are created.

I will attempt a fix at https://github.com/kubernetes/minikube/blob/master/deploy/addons/kubevirt/pod.yaml.tmpl

### Attach the log file

Not using the standard logging, found problem in:

kubectl -n kube-system logs kubevirt-install-manager

/kubevirt-scripts/install.sh: line 3: curl: command not found
Installing KubeVirt version: 
/kubevirt-scripts/install.sh: line 6: curl: command not found
error: the path "/manifests/kubevirt-operator.yaml" does not exist
error: resource mapping not found for name: "kubevirt" namespace: "kubevirt" from "STDIN": no matches for kind "KubeVirt" in version "kubevirt.io/v1"
ensure CRDs are installed first

### Operating System

Redhat/Fedora

### Driver

KVM2

adilwiser:
fixes #15918

reference: https://github.com/kubernetes/minikube/issues/15918#issuecomment-3797366305

so basically the kubevirt addon was broken and the [limitation was also on the official docs here](https://github.com/kubevirt/kubevirt.github.io/blob/0532d8fa1ef7ee31ba38ce7c91bdf335ddec4ae0/pages/quickstart_minikube.md?plain=1#L78) because the bitnami/kubectl image no longer includes curl. i tried fixing stuffs and it works perfect now :D

now this PR improves how KubeVirt is managed by applying manifests directly from URLs using kubectl’s native support, eliminating the need to download them first. It pins the KubeVirt version to v1.7.0 for better reliability and consistency, adds an automated script to update the pinned version, and introduces a Makefile target to run that update easily. The PR also integrates KubeVirt into dependency version tracking and re-enables kubectl version handling that was previously disabled due to an issue.

@medyagh here is the output of minikube with the addon as requested:

before - 
image

after -
image


hack/update/kubevirt_version/kubevirt_version.go
```diff
@@ -0,0 +1,86 @@
+/*
```
medyagh, line 1: @adilwiser thank you for this fix please also add an github action simmilar to this so it bumps it automatically
https://github.com/kubernetes/minikube/blob/master/.github/workflows/update-yakd-version.yml

and please verify manually that updater works

adilwiser: thanks for pointing me in that direction! I'll address it asap!

adilwiser:
@medyagh  the updater works !!

image

PR #22623, opened 2026-02-03
Fix lock file regression by appending UID to the lock driectory

### LINKED ISSUE #22619: HOST_HOME_PERMISSION "unable to acquire lock" in 1.38.0
### Summary
  Minikube v1.38.0 introduces a regression that causes "permission denied" errors when trying to acquire locks in CI environments (particularly GitHub Actions). The issue prevents `minikube image load` and other
  commands from working.

  ### Error Message
```
 Exiting due to HOST_HOME_PERMISSION: save to dir: caching images: caching image "/home/runner/work/_temp/.minikube/cache/images/<image_path>": unable to acquire lock for
  {Name:mk<hash> Timeout:10m0s Delay:500ms}: acquiring lock: open /tmp/minikube-locks/mk<hash>.lock: permission denied
  Error: unable to acquire lock for /tmp/minikube-locks/mk.lock: acquiring lock: open /tmp/minikube-locks/mk.lock: permission denied
```

  ### Environment
  - **Minikube version**: v1.38.0
  - **Operating System**: Linux (GitHub Actions runners)
  - **Driver**: docker
  - **Kubernetes version**: Various (reproduced on v1.30.0, v1.31.0, v1.32.0, v1.33.0, v1.34.0)

  ### Reproduction Steps
  1. Set up concurrent GitHub Actions workflow with minikube v1.38.0
  2. Start minikube: `minikube start --driver=docker`
  3. Run any command that requires locking: `minikube image load <image>`
  4. Observe permission denied error on lock file acquisition

  ### Root Cause Analysis

  **Breaking Change in PR #22460**

  PR #22460 (merged Jan 16, 2026, commit `98364c31346ee162afe2ff886beb4262d18875d3`) replaced the `juju/mutex` library with `gofrs/flock` and centralized all lock files in a single directory.

  **Code Comparison:**

  **v1.37.0** (Working):
  ```go
  // Uses juju/mutex - lock files handled internally by the library
  releaser, err := mutex.Acquire(spec)

  v1.38.0 (Broken in CI):
  // New Acquire() function creates centralized lock directory
  func Acquire(spec Spec) (Releaser, error) {
      tmpDir := os.TempDir()
      lockDir := filepath.Join(tmpDir, "minikube-locks")
      if err := os.MkdirAll(lockDir, 0755); err != nil {  // ← Problem here
          return nil, fmt.Errorf("creating lock dir: %w", err)
      }

      lockPath := filepath.Join(lockDir, spec.Name+".lock")
      f := flock.New(lockPath)
      // ... rest of lock acquisition
  }
```

  The Problem:

  In CI environments with concurrent minikube operations:
  1. The /tmp/minikube-locks directory is created with mode 0755
  2. If multiple processes run as different users (or in different security contexts), subsequent processes cannot create lock files in this directory
  3. The directory permissions (0755) don't allow write access for other users/groups

  Affected Commands
  In a concurrent setting - 
  - minikube image load
  - minikube image build
  - Any operation that calls lock.WriteFile() or lock.AppendToFile()

  Workaround

  Revert to minikube v1.37.0 until this is fixed.

  Impact

  - Breaks GitHub Actions workflows using minikube v1.38.0
  - Prevents image loading, which is critical for testing custom images

  Additional Context

  This issue was discovered when all helm chart lint tests started failing after Renovate automatically updated from v1.37.0 to v1.38.0. The error manifests specifically in the lock acquisition code introduced in
  the migration from juju/mutex to gofrs/flock.

  References:
  - Source file v1.37.0: https://github.com/kubernetes/minikube/blob/v1.37.0/pkg/util/lock/lock.go
  - Source file v1.38.0: https://github.com/kubernetes/minikube/blob/v1.38.0/pkg/util/lock/lock.go
  - Breaking PR: https://github.com/kubernetes/minikube/pull/22460

### Attach the log file

Running `minikube logs` in CI context risks exposing secrets.

### Operating System

Ubuntu

### Driver

Docker

medyagh:
fixes a regression introduced in v1.38.0 where the shared lock directory `/tmp/minikube-locks` was created with `0755` permissions. This prevented different users (or CI jobs running as different users) from acquiring locks on the same machine, resulting in `HOST_HOME_PERMISSION` errors.
Changes:
- Append the current user's UID to the lock directory name (e.g., `/tmp/minikube-locks-1000`).
- This ensures each user has a dedicated, writable directory for their lock files.
- Added `TestLockDirectoryStructure` to verify the directory naming convention.
Fixes #22619

fix https://github.com/kubernetes/minikube/issues/22619

### DIFF
--- a/pkg/util/lock/lock.go
+++ b/pkg/util/lock/lock.go
@@ -107,8 +107,11 @@ func getUserSpecificDirName() string {
 			// Fallback if home dir cannot be obtained
 			return "minikube-locks-windows-default"
 		}
+		// On Windows, os.Getuid() returns -1, so we cannot verify the user ID directly.
+		// Instead, we use the SHA1 hash of the user's home directory as a unique identifier.
+		// This ensures per-user isolation in shared temporary directories.
 		hash := sha1.Sum([]byte(homeDir))
-		return fmt.Sprintf("minikube-locks-windows-%x", hash[:8]) // Use first 8 bytes of hash
+		return fmt.Sprintf("minikube-locks-windows-%x", hash)
 	}
 	return fmt.Sprintf("minikube-locks-%d", os.Getuid())
 }
--- a/pkg/util/lock/lock_test.go
+++ b/pkg/util/lock/lock_test.go
@@ -17,12 +17,18 @@ limitations under the License.
 package lock
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
+	"strings"
 	"testing"
 	"time"
 )
 
+// TestUserMutexSpec verifies that the MutexSpec generation is deterministic and collision-resistant.
+// It ensures that different paths generate unique 40-character SHA1 based names, which is crucial
+// for avoiding lock collisions between different projects or resources.
 func TestUserMutexSpec(t *testing.T) {
 	var tests = []struct {
 		description string
@@ -174,6 +180,11 @@ func TestMutexConcurrency(t *testing.T) {
 	r2.Release()
 }
 
+// TestLockDirectoryStructure verifies that the lock directory is created with correct permissions and ownership.
+// It addresses security concerns by ensuring:
+// 1. The lock directory is created in a user-specific path to avoid permission denial in multi-user environments.
+// 2. The directory is writable by the current user, ensuring locks can be acquired.
+// 3. On Unix systems, the directory name contains the user's UID for clear isolation (e.g. minikube-locks-1000).
 func TestLockDirectoryStructure(t *testing.T) {
 	// 1. Acquire a lock
 	spec := PathMutexSpec("test-lock-structure")
@@ -195,4 +206,19 @@ func TestLockDirectoryStructure(t *testing.T) {
 	if !info.IsDir() {
 		t.Errorf("Expected %s to be a directory", lockDir)
 	}
+
+	// 3. Verify that the directory matches likely expectation for current user
+	if runtime.GOOS != "windows" {
+		uid := os.Getuid()
+		if !strings.Contains(expectedDir, fmt.Sprintf("%d", uid)) {
+			t.Errorf("Expected dir name %s to contain user UID %d", expectedDir, uid)
+		}
+	}
+
+	// 4. Verify that the directory is writable by the current user
+	testFile := filepath.Join(lockDir, "write-test")
+	if err := os.WriteFile(testFile, []byte("test"), 0600); err != nil {
+		t.Errorf("Expected to be able to write to %s: %v", lockDir, err)
+	}
+	os.Remove(testFile)
 }


medyagh:
here is the link to the binary from a PR that I think might fix this issue, do you mind trying it out ?

https://storage.googleapis.com/minikube-builds/22623/minikube-linux-amd64

nirs (commented):
This works, but it will be better keep locks the recommended location for the platform:

| Operating System | Primary Location (User-level) | Specification / Documentation |
| :--- | :--- | :--- |
| **Linux** | `$XDG_RUNTIME_DIR` (e.g., `/run/user/1000/`) | [XDG Base Directory Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) |
| **macOS** | `~/Library/Application Support/<app>/` | [Apple File System Guide](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html) |
| **Windows** | `%LOCALAPPDATA%\<app>\` | [Windows App Data Storage](https://learn.microsoft.com/en-us/windows/uwp/design/app-settings/store-and-retrieve-app-data) |

medyagh:
> This works, but it will be better keep locks the recommended location for the platform:
> 
> Operating System	Primary Location (User-level)	Specification / Documentation
> **Linux**	`$XDG_RUNTIME_DIR` (e.g., `/run/user/1000/`)	[XDG Base Directory Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
> **macOS**	`~/Library/Application Support/<app>/`	[Apple File System Guide](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html)
> **Windows**	`%LOCALAPPDATA%\<app>\`	[Windows App Data Storage](https://learn.microsoft.com/en-us/windows/uwp/design/app-settings/store-and-retrieve-app-data)

that sounds like a good idea ! however I think for a patch for 1.38.1 we should stick to something low risk and more fimilar with what we have

Summary: While the suggestion follows OS standards (which is generally good), switching to these paths for lock files introduces significant risks for Minikube's specific use cases (CI, containers, enterprise setups).

Analysis:

Reliability (The Main Risk):
CI & Containers: XDG_RUNTIME_DIR (Linux) is often not set in Docker containers or minimal CI environments (root/headless). We would need a fallback to /tmp anyway.
Networked Home Dirs (NFS): On some enterprise Linux/macOS setups, the home directory (~/...) is mounted via NFS. File locking (flock) on NFS is historically unreliable and can lead to deadlocks or ignored locks. /tmp is almost always a local disk or RAM disk (tmpfs), making it much safer for locking coordination.
Complexity:
We would need to implement robust path detection for all 3 OSes, plus the fallback logic for when those paths (like XDG_RUNTIME_DIR) are missing.
The current solution (UID/Hash in /tmp) effectively implements a "custom runtime dir" that is guaranteed to be on a local, writable filesystem.
Recommendation: Stick with the current approach (/tmp with unique names).

It solves the original "permission denied" and collision issues.
It works reliably in CI, containers, and weird machine setups where standard paths might be missing or network-mounted.
It avoids the "flock on NFS" trap

wt:
This does not seem to comply with recommendations for generating temp files/dirs [here](https://systemd.io/TEMPORARY_DIRECTORIES/). The old name and the suggested change is unfortunately guessable. Would it be better to have directory that is created with [this golang api](https://pkg.go.dev/os#MkdirTemp). It doesn't even need to have the username in it...something like '/tmp/minikube-${random_val}'?

Is there any reason that the directory needs to be the same for each run of the tool? Would it be reasonable to have `minikube {stop|delete}` clean them up?

wt:
And I don't know what the multiple lock files in that directory are for, but it would be possible to name them something more understandable if the directory itself was unique.

wt:
Also, the change seems to do what is described on a Fedora machine. I created a /tmp/minikube-locks dir and set the perms to 600. I then started a minikube cluster and saw that the `/tmp/minikube-locks-$my_uid` directory was created and populated with lock files. Start and stop seems to work fine.

nirs:
> This does not seem to comply with recommendations for generating temp files/dirs [here](https://systemd.io/TEMPORARY_DIRECTORIES/). The old name and the suggested change is unfortunately guessable. Would it be better to have directory that is created with [this golang api](https://pkg.go.dev/os#MkdirTemp). It doesn't even need to have the username in it...something like '/tmp/minikube-${random_val}'?
> 
> Is there any reason that the directory needs to be the same for each run of the tool? Would it be reasonable to have `minikube {stop|delete}` clean them up?

The lock files in the directory will be accessed by multiple minikube instances, so we cannot use random names. This is not a directory or temporary data but directory synchronizing multiple processes.

Creating the directory is safe since creating a directory is atomic operation. If the directory already exits the call will fail. If 2 minikube instance try to create the directory in the same time one will succeed and the other will with ErrExist.

The directory must be created with permissions 0o700 so only the user can access it.

When creating a lock file we can use O_CREATE|O_NOFOLLOW to protect forms symlinks attacks.

See #22624 for discussion on location of the runtime data.

wt:
> > This does not seem to comply with recommendations for generating temp files/dirs [here](https://systemd.io/TEMPORARY_DIRECTORIES/). The old name and the suggested change is unfortunately guessable. Would it be better to have directory that is created with [this golang api](https://pkg.go.dev/os#MkdirTemp). It doesn't even need to have the username in it...something like '/tmp/minikube-${random_val}'?
> > Is there any reason that the directory needs to be the same for each run of the tool? Would it be reasonable to have `minikube {stop|delete}` clean them up?
> 
> The lock files in the directory will be accessed by multiple minikube instances, so we cannot use random names. This is not a directory or temporary data but directory synchronizing multiple processes.

I assume that the lock files directory needs to be used for multiple invocations of the minikube binary. If that is true, couldn't minikube store that in the minikube profile config? Is there some problem with having a different dir for each profile?
 
> Creating the directory is safe since creating a directory is atomic operation. If the directory already exits the call will fail. If 2 minikube instance try to create the directory in the same time one will succeed and the other will with ErrExist.

The problem is that another user could create the directory and prevent my user from running minikube successfully. There are other failure modes that I can think of.

> The directory must be created with permissions 0o700 so only the user can access it.

For the test, I created the old name (/tmp/minikube-locks) with 0o600 to make sure that minikube wasn't writing to it with the change. I didn't create the new directory manually. The directory wsa created with 0o700, as expected.

> When creating a lock file we can use O_CREATE|O_NOFOLLOW to protect forms symlinks attacks.

I am less worried about symlink attacks. I am more worried about just making sure that best practices around temp dir creation/usage are followed. Also, using os.mkTempDir will make sure that the proper temp directories are used on different platforms following their conventions. For example, the TMPDIR env var on Linux and using the equivalent appropriate env vars on Windows.

> See #22624 for discussion on location of the runtime data.

I can also post a summary of this there as well, if that is useful.

wt:
I read #22624. That's a really good idea. I made a small comment, but it really sounds like the right direction.

medyagh:
> I read #22624. That's a really good idea. I made a small comment, but it really sounds like the right direction.

> I can also post a summary of this there as well, if that is useful

thanks @wt :) yes a summery would be nice, please feel free to add it to the issue I created so we can tackle it in next minor version (not patch version)

PR #22659, opened 2026-02-13
fix: PowerShell curl alias on Windows to check resgistry.k8s.io connectivity

bobsira:
**Issue**: On Windows the host-side `curl `call could invoke PowerShell's curl alias `(Invoke-WebRequest`) which uses different CLI flags. That made the host registry check fail and produced spurious "Failing to connect to h_ttps://registry.k8s.io/_" messages.

**Fix**: In tryRegistry the code now:

- Runs the real curl binary inside the guest for the in-VM check (curl).
- Runs curl.exe on Windows for the host-side check (curl.exe) to avoid the PowerShell alias. Minimal variable renaming kept; behavior otherwise unchanged.

**Result**: Host checks on Windows use the real curl executable, avoiding false failures and removing the noisy registry connection error.

_Before the fix:_

before

_After the fix:_

after

### DIFF
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -872,11 +872,17 @@ func tryRegistry(r command.Runner, driverName, imageRepository, ip string) {
 
 	curlTarget := fmt.Sprintf("https://%s/", imageRepository)
 	opts = append(opts, curlTarget)
-	exe := "curl"
+
+	// Always use the 'curl' binary inside the guest. On Windows hosts,
+	// PowerShell exposes an alias which can interfere with the host check,
+	// so use 'curl.exe' for the host-run check only.
+	cmd := exec.Command("curl", opts...)
+	hostExe := "curl"
 	if runtime.GOOS == "windows" {
-		exe = "curl.exe"
+		hostExe = "curl.exe"
 	}
-	cmd := exec.Command(exe, opts...)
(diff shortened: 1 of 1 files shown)


pkg/minikube/node/start.go
```diff
@@ -872,11 +872,17 @@ func tryRegistry(r command.Runner, driverName, imageRepository, ip string) {
 
 	curlTarget := fmt.Sprintf("https://%s/", imageRepository)
 	opts = append(opts, curlTarget)
-	exe := "curl"
+
+	// Always use the 'curl' binary inside the guest. On Windows hosts,
+	// PowerShell exposes an alias which can interfere with the host check,
+	// so use 'curl.exe' for the host-run check only.
+	cmd := exec.Command("curl", opts...)
```
mloskot, line 879: This line seems to superfluous, replaced by line 884, isn't it?

```suggestion
```

bobsira: `cmd `is the in-guest check via` r.RunCmd(cmd)` (executes inside the VM/container)

`cmdHost `( line 884 is the host-side check run via `cmdHost.Run()` (executes on the developer machine)

We can also construct cmdHost only when the host check will run.

mloskot: My bad, I got confused  and that the `cmdHost` is used way down in the function.

Issue #22269, opened 2025-12-20
discuss: if Containerd needs Backup user images "Before Preload"


medyagh:
as seen in this PR https://github.com/kubernetes/minikube/pull/22246

This failing minikube Test (TestPreload/Restart-With-Preload-Check-User-Image) 
Checks the following:
- start minikube "without" preload
- pull an user image (like busybox)
- start minikube again "WITH" preload

without Backup user images "user image (busybox) gone" in CRIO and Docker
and both Crio and docker we have custom code that Stores the Current images images before preload

- Docker https://github.com/kubernetes/minikube/blob/239b3119fe72b925701d0652c7480fd579bfe496/pkg/minikube/cruntime/docker.go#L626
- CRIO https://github.com/kubernetes/minikube/pull/22246

The question is why we dont need it in Containerd ?

Based on my analysis of pkg/minikube/cruntime/containerd.go and hack/preload-images/generate.go, Containerd should theoretically be broken essentially the same way as CRIO.

Both runtimes: 3. Have their storage in /var (/var/lib/containerd vs /var/lib/containers). 4. Have their preload tarballs generated as filesystem dumps of these directories. 5. Have their Preload function execute a destructive tar -C /var -xf extraction.

If tar overwrites the metadata database (e.g., meta.db for containerd), existing images should be lost.

If it appears "not broken", it might be due to:

- Luck/Race Condition: Maybe the test environment behaves differently or the metadata file isn't overwritten in a way that corrupts the state immediately?
- Unverified: Maybe we just assume it's working because we haven't seen the bug report yet

We should add more integration tests and if it fails, We should likely apply the same Backup/Restore logic to Containerd as well to be safe and consistent. Docker is the only one "safe" by default because of its specific refStore code.

afbjorklund:
I don't think containerd will notice the wiped/edited database, until you **restart** the daemon?

I installed alpine, created an archive of /var, pulled busybox, restored the archive:

```console
anders@lima-default:~$ sudo nerdctl images
REPOSITORY    TAG       IMAGE ID        CREATED           PLATFORM       SIZE       BLOB SIZE
busybox       latest    d80cd694d3e9    13 seconds ago    linux/amd64    4.563MB    2.215MB
alpine        latest    865b95f46d98    51 seconds ago    linux/amd64    9.105MB    3.862MB
anders@lima-default:~$ sudo systemctl restart containerd
anders@lima-default:~$ sudo nerdctl images
[...]

PR #22670, opened 2026-02-17
docker-runtime: make storage-driver overlayfs for version greater than v29

### LINKED ISSUE #22057: migrate preloads from "overlay2" to "overlayfs"
docker is moving to overlayfs as storage driver, 

https://kubernetes.slack.com/archives/C1F5CT6Q1/p1764980767008509

we currently set the docker config and set the storage driver to overlay2

our current preload scripts generate for overlay2 for docker, we need to create a new preload generation that will generate preload for ovleryafs for docker runtime and also changes our config to use it

bhavyaBeliever:
This implementation generates preload image according to docker version inside the KIC driver (for docker runtime). If the version is greater than v29, docker is configured to use storage driver as overlayfs else it uses overlay2 and respectively builds preload image. 

The changes are tested by generating a new preload image locally. 

Before:
image

After generating a new preload image and start a new cluster:

image

fixes #22057

### DIFF
--- a/pkg/minikube/cruntime/docker.go
+++ b/pkg/minikube/cruntime/docker.go
@@ -133,6 +133,7 @@ func (r *Docker) Active() bool {
 
 // Enable idempotently enables Docker on a host
 func (r *Docker) Enable(disOthers bool, cgroupDriver string, inUserNamespace bool) error {
+	storageDriver := r.selectStorageDriver()
 	if inUserNamespace {
 		if err := CheckKernelCompatibility(r.Runner, 5, 11); err != nil {
 			// For using overlayfs
@@ -163,8 +164,9 @@ func (r *Docker) Enable(disOthers bool, cgroupDriver string, inUserNamespace boo
 	if err := r.Init.Enable("docker.socket"); err != nil {
 		klog.ErrorS(err, "Failed to enable", "service", "docker.socket")
 	}
+	klog.Infof("Configuring docker service ...")
 
-	if err := r.configureDocker(cgroupDriver); err != nil {
+	if err := r.configureDocker(cgroupDriver, storageDriver); err != nil {
 		return err
 	}
 
@@ -551,15 +553,21 @@ func (r *Docker) SystemLogCmd(length int) string {
 }
 
 type dockerDaemonConfig struct {
-	ExecOpts       []string              `json:"exec-opts"`
-	LogDriver      string                `json:"log-driver"`
-	LogOpts        dockerDaemonLogOpts   `json:"log-opts"`
-	StorageDriver  string                `json:"storage-driver"`
-	DefaultRuntime string                `json:"default-runtime,omitempty"`
-	Runtimes       *dockerDaemonRuntimes `json:"runtimes,omitempty"`
+	ExecOpts            []string              `json:"exec-opts"`
+	LogDriver           string                `json:"log-driver"`
+	LogOpts             dockerDaemonLogOpts   `json:"log-opts"`
+	StorageDriver       string                `json:"storage-driver"`
+	DefaultRuntime      string                `json:"default-runtime,omitempty"`
+	Runtimes            *dockerDaemonRuntimes `json:"runtimes,omitempty"`
+	Features            dockerDaemonFeatures  `json:"features"`
+	ContainerdNamespace string                `json:"containerd-namespace,omitempty"`
 }
 type dockerDaemonLogOpts struct {
 	MaxSize string `json:"max-size"`
+}
+type dockerDaemonFeatures struct {
+	ContainerdSnapshotter bool `json:"containerd-snapshotter"`
+	ContainerdMigration   bool `json:"containerd-migration"`
 }
 type dockerDaemonRuntimes struct {
 	Nvidia struct {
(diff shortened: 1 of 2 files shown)


bhavyaBeliever:
**Outstanding Issues in the Current Implementation**

1. Docker Version and Storage Driver Mismatch
When the preload tar file is created using a KIC driver that has Docker v29 or newer, the storage driver used is `overlayfs`. However, if the cluster is later started with an older Docker version (< v29), Docker falls back to using the `overlay2` storage driver.
This mismatch causes the preload tar file to be incompatible and unusable in the older Docker environment.

2. Misleading Preload Tar File Naming
The name of the preload tar file is misleading. Even though the tar file name implies a specific storage driver, the storage driver field is hard-coded to `overlay2` for the Docker runtime.
As a result, the tar file name does not accurately reflect the actual storage driver used during its creation.

medyagh:
we will need to ensure to have a new generation of preloads after this https://github.com/kubernetes-sigs/minikube-preloads/

medyagh:
@bhavyaBeliever you mentioned we can use preload for containerd for docker runtime (if 29>) as well, 
can you hard code an example and proof that it works

with minikube start --alsologtostderr 

time minikube start --alsologtostderr

bhavyaBeliever:
[logs_docker_runtime_without_preload_overlayfs.txt](https://github.com/user-attachments/files/25412426/logs_docker_runtime_without_network_overlayfs.txt)
[logs_docker_runtime_with_preload_overlayfs.txt](https://github.com/user-attachments/files/25412428/logs_docker_runtime_with_network_overlayfs.txt)

@medyagh please find the attached logs. Also here is the patch file to reproduce the solution on your machine. 
[installing_containerd_preload_for_docker.patch](https://github.com/user-attachments/files/25413274/installing_containerd_preload_for_docker.patch)

What I observed is that containerd-based preload images work correctly, but the preloading process is significantly slower (roughly 3× slower).

EDIT:

@medyagh please check the below log file which uses containerd's preload image for container-runtime docker.
[logs_containerd_preload_images_for_docker_runtime.txt](https://github.com/user-attachments/files/25733765/logs_containerd_preload_images_for_docker_runtime.txt)

It takes about 20 seconds to start which is almost same with docker's preload image.

To reproduce this case please use the below patch file which can be applied on master's head. Please make sure that the KIC base image version is >= 0.0.49. 
[migrate_preloads.patch](https://github.com/user-attachments/files/25733748/migrate_preloads.patch)

bhavyaBeliever:
After a little research I found the that even though snapshotter used here is overlayfs which stores layers inside `/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/` the docker access this containers through `/var/lib/docker/rootfs/overlayfs/` (similar to `/var/lib/docker/overlay2/`)

Hence we need to generate new preload images of docker-runtime for overlayfs storage driver. 

EDIT:
When using the containerd runtime, preloaded images are stored in the `k8s.io` namespace. However, when Docker is configured to use the containerd snapshotter, it stores images in the `moby` namespace by default. As a result, Kubernetes components cannot access these images since they expect them to be present in the `k8s.io` namespace.

The `migrate_preloads.patch` resolves this issue by configuring Docker to use the k8s.io namespace instead of moby, ensuring that the preloaded images are visible and usable by Kubernetes.

SaurabhD18:
@medyagh, @ComradeProgrammer — the diagram below illustrates the current high-level code flow used to configure the runtime and extract the preload image.

MinikubeStartTraceTillPreload drawio
 
&nbsp;

To download a valid preload image aligned with the storage driver, we first need to determine the Docker version. There are two possible approaches:
- Infer the Docker version from the KIC base / ISO image being downloaded.
- Defer the preload image download until after runtime configuration.

While the second approach is simpler conceptually, it comes at a cost: we lose the parallelism that currently helps optimize startup time, leading to less effective utilization of network resources.
Given this trade-off, we believe the first approach is preferable, as it preserves parallel execution and minimizes startup latency.

We’d appreciate your thoughts or any alternative suggestions you may have.

bhavyaBeliever:
@medyagh @ComradeProgrammer Please review the changes.

Summary of updates:

1. The Docker daemon is configured to use the `overlayfs` storage driver and the `containerd-snapshotter`. This assumes the KIC/ISO image contains Docker version > v29.
2. Docker is configured to use the containerd namespace `k8s.io`.
3. While extracting preload images for the Docker runtime, the containerd preload images are downloaded and extracted.
4. Preload images are **not** generated separately for the Docker runtime, since Docker now uses containerd’s images.

bhavyaBeliever:
The code doesn't work for:
1. The cluster with VM based driver since ISO currently uses a stable [docker version(v28.5.2)](https://github.com/kubernetes/minikube/blob/d5796e6cc57812c49c2a81510a429bef96a4ae65/deploy/iso/minikube-iso/arch/x86_64/package/docker-bin/docker-bin.mk#L7)
2. If the user updates minikube by this binary, the old cluster's **user pods might not work** since the image was previously stored in `/var/lib/docker/images/overlay2` . They need to rebuild the image inside the cluster if the images aren't present in docker-hub.
3. Changing of kubernetes version after restarting the cluster with preload=true doesn't guarantee old about user images. [#22269](http://github.com/kubernetes/minikube/issues/22269) 

Things which are pending:

- [x] 1. Update the testcases that checks storage driver
- [ ] 2. Raise a PR  in [minikube-preloads](https://github.com/kubernetes-sigs/minikube-preloads)

hack/preload-images/generate.go
```diff
@@ -68,9 +68,8 @@ func generateTarball(kubernetesVersion, containerRuntime, tarballFilename string
 		return fmt.Errorf("kubeadm images: %w", err)
 	}
 
-	if containerRuntime != "docker" { // kic overlay image is only needed by containerd and cri-o https://github.com/kubernetes/minikube/issues/7428
-		imgs = append(imgs, images.KindNet(""))
```
medyagh: docker usually does not need a CNI, why in this PR we are making cri-ddocker runtime to use cni and how is it that related to this overlayfs PR?

medyagh (changes requested):

hack/preload-images/generate.go
bhavyaBeliever: The removal code in `/hack/preload_images` is to stop making docker preload images since now the docker-runtime **uses** containerd images.

bhavyaBeliever: Wait, if we stop to make preload images for docker runtime, The people using older minikube versions won't be able use new kubernetes version for docker runtime. Right?

bhavyaBeliever:
@nirs This PR is intended to configure docker-runtime's storage driver based on the version. I think this PR is still valid with the new Proposal you have mention in #23018.

nirs:
> @nirs This PR is intended to configure docker-runtime's storage driver based on the version. I think this PR is still valid with the new Proposal you have mention in #23018.

If this is not about preloads but optimizing docker runtime storage driver you are probably correct. But why do we change docker runtime defaults? We need a good reason to change the default.

bhavyaBeliever:
> But why do we change docker runtime defaults? We need a good reason to change the default.

After the release of docker v29 the default storage driver is shifted to `overlayfs` with containerd image store (https://docs.docker.com/engine/storage/containerd/). It means that the image layers will be now managed by containerd. But for older cluster we need to configure storage driver as `overlay2`.

nirs:
> > But why do we change docker runtime defaults? We need a good reason to change the default.
> 
> After the release of docker v29 the default storage driver is shifted to `overlayfs` with containerd image store (https://docs.docker.com/engine/storage/containerd/). It means that the image layers will be now managed by containerd. But for older cluster we need to configure storage driver as `overlay2`.

What do you mean by older cluster? What it we remove the configuration and let docker use the default?

bhavyaBeliever:
> What do you mean by older cluster?

A cluster started with older docker version and with new minikube release.
> What it we remove the configuration and let docker use the default?

Should work, but if you look at the PR implementation for docker v29+ we are keeping the docker images in `k8.io` containerd-namespace  whereas the default is `moby`. The `k8s.io` is the namespace where minikube too stores images for containerd.

Issue #22314, opened 2025-12-24
investigate cgroup v2 is used by all container runtimes


medyagh:
some integration test failures on Prow hint me that maybe we are not optimized for cgroup v2 (maybe different than jenkins)

in one minikube test log I see warrnings like
https://storage.googleapis.com/minikube-builds/logs/22301/42940/Docker_Linux.html#fail_TestFunctionalNewestKubernetes%2fVersionv1.35.0-rc.1%2fparallel%2fDockerEnv%2fbash

```
ControlAvailable:false Error: RemoteManagers:<nil>} LiveRestoreEnabled:false Isolation: InitBinary:docker-init ContainerdCommit:{ID:dea7da592f5d1d2b7755e3a161be07f43fad8f75 Expected:} RuncCommit:{ID:v1.3.4-0-gd6d73eb8 Expected:} InitCommit:{ID:de40ad0 Expected:} SecurityOptions:[name=apparmor name=seccomp,profile=builtin] ProductLicense: Warnings:[WARNING: Support for cgroup v1 is deprecated and planned to be removed


	==> Docker <==

	Dec 22 22:57:39 functional-384766 dockerd[17979]: time="2025-12-22T22:57:39.427378701Z" level=info msg="Docker daemon" commit=fbf3ed2 containerd-snapshotter=false storage-driver=overlay2 version=29.1.3
```

related
https://github.com/moby/moby/issues/51111

PR #22695, opened 2026-02-21
ci: Fixes for cgroup compilation

p172913:
What type of PR is this?
/kind enhancement

What this PR does / why we need it:

Added implementation of isCgroupV2() in cgroups_other.go for non-Linux platforms (always returns `false`). Also added HasMemoryCgroup() for consistent behavior across platforms.

Which issue(s) this PR fixes:

https://github.com/kubernetes/minikube/issues/22314

Special notes for your reviewer:
None

Does this PR introduce a user-facing change?
None

Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:
None


pkg/drivers/kic/oci/cgroups_linux.go
```diff
@@ -59,8 +64,46 @@ func HasMemoryCgroup() bool {
 	return true
 }
 
+func isCgroupV2() bool {
```
medyagh, line 67: can you check if we already have a func that detects cgroupv2 ?

p172913: I found a function cgroupVersion inside minikube\pkg\minikube\detect\detect_linux.go to check if this cgroup is whether v1 or v2.

Issue #22705, opened 2026-02-23
build minikube iso in Free GH runners


medyagh:
Try to use cache

medyagh:
@afbjorklund

medyagh:
```
'/usr/lib/systemd/system/systemd-networkd.service'.
2026-02-24T01:43:34.8090232Z ln -snf /home/runner/work/minikube/minikube/out/buildroot/output-x86_64/host/x86_64-minikube-linux-gnu/sysroot /home/runner/work/minikube/minikube/out/buildroot/output-x86_64/staging
2026-02-24T01:43:34.8318456Z make[1]: Leaving directory '/home/runner/work/minikube/minikube/out/buildroot'
2026-02-24T01:43:34.8329955Z # x86_64 ISO is still BIOS rather than EFI because of AppArmor issues for KVM, and Gen 2 issues for Hyper-V
2026-02-24T01:43:34.8335870Z if [ "x86_64" = "aarch64" ]; then \
2026-02-24T01:43:34.8337429Z                 mv /home/runner/work/minikube/minikube/out/buildroot/output-aarch64/images/boot.iso /home/runner/work/minikube/minikube/out/minikube-arm64.iso; \
2026-02-24T01:43:34.8338236Z         else \
2026-02-24T01:43:34.8339794Z                 mv /home/runner/work/minikube/minikube/out/buildroot/output-x86_64/images/rootfs.iso9660 /home/runner/work/minikube/minikube/out/minikube-amd64.iso; \
2026-02-24T01:43:34.8340475Z         fi;
2026-02-24T01:43:34.8849595Z echo "{\"iso_version\": \"v1.38.0-1771889789-22708\", \"kicbase_version\": \"v0.0.50\", \"minikube_version\": \"v1.38.1\", \"commit\": \"3a1f73851ec4bf0e37e6226bfdf2c5bc38e949bc\"}" > deploy/iso/minikube-iso/board/minikube/amd64/rootfs-overlay/version.json
2026-02-24T01:43:35.0917686Z /bin/sh: 1: cannot create deploy/iso/minikube-iso/board/minikube/amd64/rootfs-overlay/version.json: Directory nonexistent
2026-02-24T01:43:35.0919266Z make: *** [Makefile:285: iso-prepare-amd64] Error 2
2026-02-24T01:43:35.0940278Z ##[error]Process completed with exit code 2.
```
[...]

Issue #22167, opened 2025-12-16
move ISO_BUILD_IMAGE to registry.k8s.io


medyagh:
in our make file we have a build image for ISO 
 $(REGISTRY)/buildroot-image

this needs to be moved to registry.k8s.io

afbjorklund:
I think "moved" implies that it actually exists:

`Error response from daemon: failed to resolve reference "gcr.io/k8s-minikube/buildroot-image:latest": gcr.io/k8s-minikube/buildroot-image:latest: not found`

But you could change the REGISTRY perhaps, to use another imaginary name for it... Or did you mean to host it?

```Makefile
REGISTRY ?= gcr.io/k8s-minikube

ISO_BUILD_IMAGE ?= $(REGISTRY)/buildroot-image
STORAGE_PROVISIONER_MANIFEST ?= $(REGISTRY)/storage-provisioner:$(STORAGE_PROVISIONER_TAG)
STORAGE_PROVISIONER_IMAGE ?= $(REGISTRY)/storage-provisioner-$(GOARCH):$(STORAGE_PROVISIONER_TAG)
KICBASE_IMAGE_GCR ?= $(REGISTRY)/kicbase:$(KIC_VERSION)
```

medyagh:
yeah this image doesnt get used at all I think, we build ISO without docker If i am not wrong, here is an example logs
https://storage.cloud.google.com/minikube-builds/logs/22101/6836e9f/iso_build.txt

afbjorklund:
It is only used as a local name, some other runtimes use other imaginary prefixes like `localhost` (or `docker.io`....)

But it does get used, by `make buildroot-image` which is still in the docs. Even if image was broken on aarch64.

PR #22708, opened 2026-02-23
WIP: Build iso in gha

medyagh:
part of https://github.com/kubernetes/minikube/issues/22167

https://github.com/kubernetes/minikube/issues/20887

https://github.com/kubernetes/minikube/issues/22705


.github/workflows/iso-build.yml
```diff
@@ -46,9 +46,24 @@ jobs:
...
           fi
           ./hack/jenkins/build_changelog.sh deploy/iso/minikube-iso/board/minikube/${ARCH_DIR}/rootfs-overlay/CHANGELOG
+      - name: Cache build directories
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.buildroot-ccache
+            ~/.cache/go-build
+            out/buildroot/output-${{ matrix.arch }}/dl
+            out/buildroot/output-${{ matrix.arch }}/share/host-go-cache
```
nirs, line 65: These caches are huge, I would cache each directory in separate cache. It will make it easier to manage by deleting single cache.

.github/workflows/iso-build.yml
```diff
@@ -46,9 +46,24 @@ jobs:
...
+          key: ${{ runner.os }}-${{ matrix.arch }}-iso-build-${{ hashFiles('deploy/iso/minikube-iso/package/Config.in') }}
```
nirs: Remove the config hash - this will invalidate the entire cache on every config change. Practically every PR does at least one config change so the cache will not be useful.

.github/workflows/iso-build.yml
```diff
@@ -46,9 +46,24 @@ jobs:
...
+          restore-keys: |
+            ${{ runner.os }}-${{ matrix.arch }}-iso-build-
       - name: Build ISO
         run: |
-          make minikube-iso-${{ matrix.arch }}
+          if [ "${{ matrix.arch }}" == "arm64" ]; then
```
nirs: We can keep the arch in the build matrix instead of this if block.

.github/workflows/iso-build.yml
```diff
@@ -10,11 +10,16 @@ env:
...
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        arch: [amd64, arm64]
+        include:
+          - arch: amd64
+            os: ubuntu-latest
+          - arch: arm64
+            os: ubuntu-24.04-arm
```
nirs: To simplify we can use:

```yaml
- arch: aarch64
  goarch: arm64
  os: ubuntu-24.04-arm
- arch: x86_64
  goarch: amd64
  os: ubuntu-24.04
```

And later use:

```
make minikube-iso-${{ matrix.arch }} 
```

medyagh: good idea

.github/workflows/iso-build.yml
medyagh: actually the config.in barely gets touched but I agree that we dont need a key here

.github/workflows/iso-build.yml
medyagh: yes! good idea

PR #22713, opened 2026-02-26
Document that `minikube ssh` can run a single command on the host

pfif:
This is simply expanding documentation for `minikube ssh`.

I struggled to find a way to run a command on nodes programmatically. Delightfully, this feature has already been implemented in Minikube!

I expand the documentation to hopefully make it clearer for future users. I based the documentation on the style of `minikube kubectl`'s. It can probably be improved

# Before
```
$ minikube ssh --help
Log into or run a command on a machine with SSH; similar to 'docker-machine ssh'.

Options:
    --native-ssh=true:
        Use native Golang SSH client (default true). Set to 'false' to use the command line 'ssh' command when
        accessing the docker machine. Useful for the machine drivers when they will not start with 'Waiting for SSH'.

    -n, --node='':
        The node to ssh into. Defaults to the primary control plane.

Usage:
  minikube ssh [flags] [options]

```

# After
```bash
$ minikube ssh --help
Log into or run a command (remember -- after minikube ssh!) on a machine with SSH; similar to 'docker-machine ssh'.

Examples:
minikube ssh -- pwd
minikube ssh -- df -h

Options:
    --native-ssh=true:
        Use native Golang SSH client (default true). Set to 'false' to use the command line 'ssh' command when
        accessing the docker machine. Useful for the machine drivers when they will not start with 'Waiting for SSH'.

    -n, --node='':
        The node to ssh into. Defaults to the primary control plane.

Usage:
  minikube ssh [flags] [options]

```

### DIFF
--- a/cmd/minikube/cmd/ssh.go
+++ b/cmd/minikube/cmd/ssh.go
@@ -36,9 +36,10 @@ var nativeSSHClient bool
 
 // sshCmd represents the docker-ssh command
 var sshCmd = &cobra.Command{
-	Use:   "ssh",
-	Short: "Log into the minikube environment (for debugging)",
-	Long:  "Log into or run a command on a machine with SSH; similar to 'docker-machine ssh'.",
+	Use:     "ssh",
+	Short:   "Log into the minikube environment (for debugging)",
+	Long:    "Log into or run a command (remember, -- after minikube ssh!) on a machine with SSH; similar to 'docker-machine ssh'.",
+	Example: "minikube ssh -- pwd\nminikube ssh -- df -h",
 	Run: func(_ *cobra.Command, args []string) {
 		options := flags.CommandOptions()
 		cname := ClusterFlagValue()


cmd/minikube/cmd/ssh.go
```diff
@@ -36,9 +36,10 @@ var nativeSSHClient bool
 
 // sshCmd represents the docker-ssh command
 var sshCmd = &cobra.Command{
-	Use:   "ssh",
-	Short: "Log into the minikube environment (for debugging)",
-	Long:  "Log into or run a command on a machine with SSH; similar to 'docker-machine ssh'.",
+	Use:     "ssh",
+	Short:   "Log into the minikube environment (for debugging)",
+	Long:    "Log into or run a command (remember, -- after minikube ssh!) on a machine with SSH; similar to 'docker-machine ssh'.",
```
nirs, line 41: "(remember -- after minikube ssh!)" is not needed.

cmd/minikube/cmd/ssh.go
```diff
@@ -36,9 +36,10 @@ var nativeSSHClient bool
 
 // sshCmd represents the docker-ssh command
 var sshCmd = &cobra.Command{
-	Use:   "ssh",
-	Short: "Log into the minikube environment (for debugging)",
-	Long:  "Log into or run a command on a machine with SSH; similar to 'docker-machine ssh'.",
+	Use:     "ssh",
+	Short:   "Log into the minikube environment (for debugging)",
```
nirs, line 40: If we touch this, we need remove "(for debugging)". minikube ssh is not for debugging.

cmd/minikube/cmd/ssh.go
```diff
@@ -36,9 +36,10 @@ var nativeSSHClient bool
...
+	Long:    "Log into or run a command (remember, -- after minikube ssh!) on a machine with SSH; similar to 'docker-machine ssh'.",
+	Example: "minikube ssh -- pwd\nminikube ssh -- df -h",
```
nirs, line 42: One example is enough - using an example with some command line arguments is good, since -- is required only when using command line arguments.

nirs:
We need to change the usage message to:

````
Usage:
  minikube ssh [flags] [options] -- command [arg ...]
  minikube ssh [flags] [options] -- "command arg ..."
```

pfif:
Sorry, work is pretty intense these days, but I will try to get to this this week

PR #22740, opened 2026-03-07
Authenticate GitHub API call in K8s version discovery

manusa:
## Summary

`IsInGitHubKubernetesVersions` creates an unauthenticated GitHub client, which causes
rate-limit errors (HTTP 403) when running minikube within a GitHub Action where
`GITHUB_TOKEN` is available.

This change reuses the existing oauth2 pattern from `pkg/minikube/download/gh/gh.go`
so that the client respects the `GITHUB_TOKEN` environment variable, falling back to
an unauthenticated client when no token is set.

## Changes

- Extract GitHub client creation into a `newGitHubClient` package-level variable
  for testability (same pattern used in `cmd/minikube/cmd/delete.go`,
  `cmd/minikube/cmd/start.go`, etc.)
- Use `GITHUB_TOKEN` with oauth2 when available (same wiring as `gh.go`)
- Add unit tests for both the release lookup logic and the auth wiring

## Follow-up

A cleaner refactor could export a shared `NewClient` helper from
`pkg/minikube/download/gh` to avoid duplicating the oauth2 wiring across packages.
Would this be a welcome change? Happy to send a follow-up PR if so.

### DIFF
--- a/cmd/minikube/cmd/config/kubernetes_version.go
+++ b/cmd/minikube/cmd/config/kubernetes_version.go
@@ -19,9 +19,11 @@ package config
 import (
 	"context"
 	"net/http"
+	"os"
 
 	"github.com/google/go-github/v84/github"
 	"golang.org/x/mod/semver"
+	"golang.org/x/oauth2"
 	"k8s.io/minikube/pkg/minikube/constants"
 )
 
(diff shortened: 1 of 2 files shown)


nirs:
> ## Follow-up
> A cleaner refactor could export a shared `NewClient` helper from `pkg/minikube/download/gh` to avoid duplicating the oauth2 wiring across packages. Would this be a welcome change? Happy to send a follow-up PR if so.

@manusa  I did not have time to look at this yet, but this sounds like the right way. Have single client that we can use in any code accessing GitHub to avoid rate limits.

But before you spend time on it, let's create an issue for this, and collect example builds that failed because of this issue. I don't remember any build failing with rate limits in the last weeks.

manusa:
> > ## Follow-up
> > A cleaner refactor could export a shared `NewClient` helper from `pkg/minikube/download/gh` to avoid duplicating the oauth2 wiring across packages. Would this be a welcome change? Happy to send a follow-up PR if so.
> 
> @manusa I did not have time to look at this yet, but this sounds like the right way. Have single client that we can use in any code accessing GitHub to avoid rate limits.
> 
> But before you spend time on it, let's create an issue for this, and collect example builds that failed because of this issue. I don't remember any build failing with rate limits in the last weeks.

Let me gather some data for your inspection.

> I don't remember any build failing with rate limits in the last weeks.

This is not a problem within the minikube repo itself, but within repos that consume minikube to start a Minikube-Kubernetes cluster in a GitHub action runner.

PR #22775, opened 2026-03-23
Fix usage string for minikube node commands.

toddliebenschutz-jones:
The delete, start and stop commands require the name of the node. This PR fixes the usage strings and adds an example invocation for each of these three commands.

I'm new to the project. This is how I made the change:
1. Edit `cmd/minikube/cmd/node_<action>.go` directly.
2. Run `make generate-docs` to create the changes to the `site/content/en/docs/commands/node.md` file.
3. Run `make` on my machine to generate the diff below.

The style is consistent with the `minikube addons enable` command.

### Help diff

Before:

```
$ out/minikube node delete --help
Deletes a node from a cluster.

Usage:
  minikube node delete [flags] [options]

```

After:

```
$ out/minikube node delete --help
Deletes a node from a cluster.

Examples:
minikube node delete m02

Usage:
  minikube node delete NODE_NAME [flags] [options]

```

### DIFF
--- a/cmd/minikube/cmd/node_delete.go
+++ b/cmd/minikube/cmd/node_delete.go
@@ -34,9 +34,10 @@ import (
 )
 
 var nodeDeleteCmd = &cobra.Command{
-	Use:   "delete",
-	Short: "Deletes a node from a cluster.",
-	Long:  "Deletes a node from a cluster.",
+	Use:     "delete NODE_NAME",
+	Short:   "Deletes a node from a cluster.",
+	Long:    "Deletes a node from a cluster.",
+	Example: "minikube node delete m02",
 	Run: func(_ *cobra.Command, args []string) {
 		if len(args) == 0 {
 			exit.Message(reason.Usage, "Usage: minikube node delete [name]")
--- a/cmd/minikube/cmd/node_delete.go
+++ b/cmd/minikube/cmd/node_delete.go
@@ -34,12 +34,13 @@ import (
 )
 
 var nodeDeleteCmd = &cobra.Command{
-	Use:   "delete",
-	Short: "Deletes a node from a cluster.",
-	Long:  "Deletes a node from a cluster.",
+	Use:     "delete NODE_NAME",
+	Short:   "Deletes a node from a cluster.",
+	Long:    "Deletes a node from a cluster.",
+	Example: "minikube node delete m02",
 	Run: func(_ *cobra.Command, args []string) {
 		if len(args) == 0 {
-			exit.Message(reason.Usage, "Usage: minikube node delete [name]")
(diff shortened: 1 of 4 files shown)


cmd/minikube/cmd/node_delete.go
```diff
@@ -34,9 +34,10 @@ import (
...
-	Use:   "delete",
-	Short: "Deletes a node from a cluster.",
-	Long:  "Deletes a node from a cluster.",
+	Use:     "delete NODE_NAME",
+	Short:   "Deletes a node from a cluster.",
+	Long:    "Deletes a node from a cluster.",
+	Example: "minikube node delete m02",
 	Run: func(_ *cobra.Command, args []string) {
 		if len(args) == 0 {
 			exit.Message(reason.Usage, "Usage: minikube node delete [name]")
```
nirs: name is shown here as optional argument - the exit message should be fixed.

Same issue exists in the other commands.

site/content/en/docs/commands/node.md
nirs, line 1: I think this file is auto generated and you don't ned to do change it. It will be updated automatically.

nirs (changes requested):

nirs:
I verified the issue, without node name the command fails with wrong usage message:

```console
% minikube node start    

❌  Exiting due to MK_USAGE: Usage: minikube node start [name]

% minikube node start -h
Starts an existing stopped node in a cluster.

Options:
    --delete-on-failure=false:
        If set, delete the current cluster if start fails and try again. Defaults to false.

Usage:
  minikube node start [flags] [options]

```

cmd/minikube/cmd/node_delete.go
toddliebenschutz-jones: Fixed in https://github.com/kubernetes/minikube/pull/22775/changes/582a1b0e647a46b64aed7381d7fc071d71e55a6b

site/content/en/docs/commands/node.md
toddliebenschutz-jones: I removed my changes to the file in https://github.com/kubernetes/minikube/pull/22775/changes/df8641cca2fde72d71a779c25e3a7200e1777c94

nirs:
@toddliebenschutz-jones Thanks! The change look good.

Can you squash the commits? We like clean git history without fixup commits.

toddliebenschutz-jones:
> @toddliebenschutz-jones Thanks! The change look good.
> 
> Can you squash the commits? We like clean git history without fixup commits.

@nirs Squashed

nirs (approved):
@toddliebenschutz-jones thank!

/lgtm
/ok-to-test

PR #22778, opened 2026-03-24
ci: skip containerd update when incompatible with current nerdctl version

### LINKED ISSUE #22084: Update nerdctl and containerd together
nerdctl 2.1.x does not support containerd 2.2.x, so we should update containerd with nerdctl, or update containerd after nerdctl is updated.

- https://github.com/containerd/nerdctl/releases/tag/v2.2.0 - This release of nerdctl is expected to be used with containerd v1.7, v2.0, v2.1, or v2.2.
- https://github.com/containerd/nerdctl/releases/tag/v2.1.6 - This release of nerdctl is expected to be used with containerd v1.7, v2.0, or v2.1.

I think the simplest way to avoid issues is to run both updates in the same PR, always updating to latest stable version of nerdctl and containerd.

nitishagar:
**Fixes #22084**

## What does this PR do?

Prevents the containerd auto-update workflow from bumping containerd to a version that is incompatible with the currently pinned nerdctl version.

### Root cause

The `update-nerdctl-version.yml` and `update-containerd-version.yml` workflows run independently on separate schedules. This allows containerd to be bumped to e.g. `2.2.x` while nerdctl is still on `2.1.x` — an incompatible combination since nerdctl `2.1.x` only supports containerd `≤ 2.1.x`.

### Fix

Added a `compatCheck` step to `update-containerd-version.yml` that runs after the version bump but before PR creation:

1. Reads the current nerdctl version via `DEP=nerdctl make get-dependency-version`
2. Extracts `major.minor` from both the proposed containerd version and the current nerdctl version
3. Compares them using `sort -V` — if containerd's `major.minor` exceeds nerdctl's, runs `git restore .` to revert changes and sets `skip=true`
4. The PR creation and bot comment steps are gated on `skip != 'true'`

**Behavior:**
- Same `major.minor` (e.g. both on `2.2.x`) → update proceeds normally
- containerd bumps to `2.3.x` while nerdctl is `2.2.x` → silently skipped; the next nerdctl update run will bring nerdctl up, then the following containerd run will succeed

### Files changed
- `.github/workflows/update-containerd-version.yml` — added compatibility check step + gating conditions

## Testing

- Workflow YAML validated with `python3 -c "import yaml; yaml.safe_load(open(...))"` — valid
- `make lint` — passes
- No Go code changes required


.github/workflows/update-containerd-version.yml
```diff
@@ -28,9 +28,27 @@ jobs:
           echo "changes<<EOF" >> "$GITHUB_OUTPUT"
           echo "$(git status --porcelain)" >> "$GITHUB_OUTPUT"
           echo "EOF" >> "$GITHUB_OUTPUT"
+      - name: Check nerdctl compatibility
+        id: compatCheck
+        run: |
+          NERDCTL_VERSION=$(DEP=nerdctl make get-dependency-version)
```
medyagh: @nitishagar is there any documentation for this requirment or any link to any issue ? 
and if it is a requirement, how about we add that to the actual go code that updates it instead of the workflow? in the hack/update folder?

medyagh (changes requested):

.github/workflows/update-containerd-version.yml
nitishagar: I tagged the original issue in the description: https://github.com/kubernetes/minikube/issues/22084

nirs:
@nitishagar The idea to check current nerdctl version in the containerd update go code sounds cleaner.

It is beter to have the logic in the update programs that can run locally and tested, instead of in the workflow. Go code is easy to maintain and understand, and runs everywhere. GitHub yaml is ugly, hard to work with and runs only in GitHub actions.

We already have the code to get nerdctl version in another go program - maybe we can share it?

nitishagar:
Thanks @nirs, agreed that's the better home for it. I moved the check into the containerd update program — it now reads the pinned nerdctl version from the kicbase Dockerfile (same place get-dependency-version looks) and skips the update when the new containerd major.minor is ahead of nerdctl's. The workflow is back to its original shape.

Added a few unit tests for the comparison logic while at it. That needed a small fix first: hack/update's shared init calls flag.Parse(), which made these programs crash under go test. Let me know what you think.

PR #22808, opened 2026-04-07
docs: add docker-env prompt tip

### LINKED ISSUE #6489: make it obvious in terminal for user if using docker-env
I was thinking as part of the eval $(minikube docker-env) we could pass something that changes the terminal Prompt so it is obvious for the user they are usinf minikube's docker not their own docker.

maybe:
something like:
```
medya@~ $ PS1="$PS1""(using minikube dockerd) $:"
```
that will try turn into
```
medya@~ $ (using minikube dockerd) $:
```

the minikube word would be profile name, this will also make it clear which profile's docker-env we are using.

helps with https://github.com/kubernetes/minikube/issues/6462
helps with https://github.com/kubernetes/minikube/issues/6437

also we could explicity set a ENV that says

```
USES_MINIKUBE_DOCKERD=PROFILE_NAME
```

iacker:
## Summary
- add a new tip to the image pushing handbook showing how to surface `MINIKUBE_ACTIVE_DOCKERD` in the shell prompt
- document a simple `bash` / `zsh` example for making the active minikube Docker target visible
- note that users with custom prompts should adapt the snippet rather than replacing their prompt configuration verbatim

## Validation
- `cd site && npm ci`
- independent content review against issue discussion and existing docs
- site build target was not available via `make -C site build` in this checkout

Fixes #6489

### DIFF
--- a/site/content/en/docs/handbook/pushing.md
+++ b/site/content/en/docs/handbook/pushing.md
@@ -107,8 +107,28 @@ By closing the terminal, you will go back to using your own system's docker daem
 
 {{% pageinfo color="info" %}}
 Tip 3:
+You can use the `MINIKUBE_ACTIVE_DOCKERD` environment variable in your shell prompt to remind yourself when `docker` commands are targeting the Docker daemon inside minikube.
+
+For example, in `bash` or `zsh`:
+
+```shell
+__active_minikube() {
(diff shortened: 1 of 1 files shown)


iacker:
Closing due to prolonged inactivity and triage-bot stale marking. The docs improvements can be re-proposed if the areas are still relevant and actively maintained.

Issue #22606, opened 2026-02-02
Can't start multi-node cluster with Rootless Podman and Containerd runtime


up9t:
### What Happened?

It won't start.

```bash
minikube start --driver=podman --nodes=2 --container-runtime=containerd

😄  minikube v1.38.0 on Fedora 43
    ▪ MINIKUBE_ROOTLESS=true
✨  Using the podman driver based on user configuration
📌  Using rootless Podman driver
👍  Starting "minikube" primary control-plane node in "minikube" cluster
🚜  Pulling base image v0.0.49 ...
E0202 13:40:40.320728  452129 cache.go:239] Error downloading kic artifacts:  not yet implemented, see issue #8426
🔥  Creating podman container (CPUs=2, Memory=3072MB) ...
📦  Preparing Kubernetes v1.35.0 on containerd 2.2.1 ...
🔗  Configuring CNI (Container Networking Interface) ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: storage-provisioner, default-storageclass

👍  Starting "minikube-m02" worker node in "minikube" cluster
🚜  Pulling base image v0.0.49 ...
E0202 13:40:54.108148  452129 cache.go:239] Error downloading kic artifacts:  not yet implemented, see issue #8426
🔥  Creating podman container (CPUs=2, Memory=3072MB) ...
🌐  Found network options:
    ▪ NO_PROXY=192.168.49.2
❗  Failing to connect to https://registry.k8s.io/ from inside the minikube container
💡  To pull new external images, you may need to configure a proxy: https://minikube.sigs.k8s.io/docs/reference/networking/proxy/
📦  Preparing Kubernetes v1.35.0 on containerd 2.2.1 ...
    ▪ env NO_PROXY=192.168.49.2



E0202 13:46:01.563777  452129 start.go:347] worker node failed to join cluster, will retry: kubeadm join: sudo /bin/bash -c "env PATH="/var/lib/minikube/binaries/v1.35.0:$PATH" kubeadm join control-plane.minikube.internal:8443 --token a129i0.wzh061d8nc6nrhha --discovery-token-ca-cert-hash sha256:76df01fa1b2eb181b87723a2a57f45621713523051ec34d1c06973f14e2ef03e --ignore-preflight-errors=all --cri-socket unix:///run/containerd/containerd.sock --node-name=minikube-m02": Process exited with status 1
stdout:
[...]

Issue #16962, opened 2023-07-29
pod networking fails if net.ipv4.ip_forward=0 running on rootless podman


haodeon:
### What Happened?

I am doing some testing with kind and minikube. I have run into a k8s networking issue which I am hoping someone with more experience can help explain.

I am running rootless podman on Pop_OS! which is based off Ubuntu. When I create a cluster using minikube, pod networking is not working. I initially thought this was a DNS issue but then when I looked at the coredns pod logs it tells me that it cannot reach `192.168.49.1:53: i/o timeout`.

After further debugging I discover that by enabling `net.ipv4.ip_forward` it all works after recreating the cluster.

What is confusing to me is when I use kind to create a cluster it works just fine without `net.ipv4.ip_forward` being enabled.

From what I can tell minikube and kind are both using kindnet by default. I tried the bridge cni in minikube and it wouldn't work regardless of IP forwarding being enabled or disabled.

Is it intended to require `net.ipv4.ip_forward` to be enabled for minikube with rootless podman to work?

How is kind able to work with `net.ipv4.ip_forward` disabled?

### Attach the log file

[log.txt](https://github.com/kubernetes/minikube/files/12206965/log.txt)

### Operating System

Ubuntu

### Driver

Podman

haodeon:
Haven’t found an answer yet but issue #17107 appears to be a duplicate of this one.

Have tested @juparog’s workaround by adding another name server, like `8.8.8.8` to `/etc/resolv.conf`. After minikube stop + start, CoreDNS is able to resolve lookups.

SanjayVas:
It looks like this may affect rootless Docker as well. See https://github.com/kubernetes/minikube/issues/18667#issuecomment-2121495029

Angelin01:
I'm going to let https://github.com/kubernetes/minikube/issues/18667 be closed automatically in favor of this one. The problem is also present with the docker driver.

Angelin01:
At this point it's been so long I'm tempted to open a Pull Request updating the documentation instead. It might as well be a feature.

PR #22823, opened 2026-04-10
kicbase: fix deterministic machine-id breaking MAC addresses in multi-node Podman rootless clusters

RobinMcCorkell:
## Problem

`/var/lib/dbus/machine-id` is baked into the kicbase container image at build time. When the entrypoint's `fix_machine_id` runs `systemd-machine-id-setup`, it finds that file and derives `/etc/machine-id` from it — producing the **same machine ID in every container**.

This breaks anything that depends on the machine ID being unique per node. The most visible symptom is in **multi-node minikube clusters using Podman rootless mode**: a `veth` interface is placed into each Podman container, and systemd configures it according to `MACAddressPolicy=persistent`. That policy selects a MAC address based on the machine ID (via `systemd-machine-id-setup`, which reads from the D-Bus machine ID: https://www.freedesktop.org/software/systemd/man/latest/systemd-machine-id-setup.html). Because all containers share the same machine ID derived from the baked-in D-Bus ID, **all nodes get identical MAC addresses on `eth0`**, causing network failures.

## Fix

Per https://systemd.io/CONTAINER_INTERFACE/, add a `RUN` step to the Dockerfile (before the final squash) that:
- **Truncates `/etc/machine-id` to an empty file** — the spec requires the file to be present but uninitialized so systemd can fill it on boot.
- **Deletes `/var/lib/dbus/machine-id`** — removes the baked-in D-Bus ID that was the source of the deterministic machine ID.

With these changes, systemd generates a fresh random machine ID on every container boot without any entrypoint assistance. This makes the existing `fix_machine_id` function in the entrypoint redundant — it has been **removed**.

## Testing

Tested against the real kicbase image (`gcr.io/k8s-minikube/kicbase-builds:v0.0.50-1772266598-22719`) with Podman rootless.

### Reproducing the bug

Both files contain the same baked-in value in the current image:

```
$ KICBASE=gcr.io/k8s-minikube/kicbase-builds:v0.0.50-1772266598-22719
$ podman run --rm --entrypoint bash $KICBASE -c '
    echo "/etc/machine-id:          $(cat /etc/machine-id)"
    echo "/var/lib/dbus/machine-id: $(cat /var/lib/dbus/machine-id)"
  '
/etc/machine-id:          636135c8833b3f1430bfc1ec69a2a4d1
/var/lib/dbus/machine-id: 636135c8833b3f1430bfc1ec69a2a4d1
```

Running `systemd-machine-id-setup` (as `fix_machine_id` does) produces the same ID every time:

```
$ for i in 1 2 3; do
    podman run --rm --entrypoint bash $KICBASE -c '
      rm -f /etc/machine-id; systemd-machine-id-setup 2>/dev/null; cat /etc/machine-id'
  done
636135c8833b3f1430bfc1ec69a2a4d1
636135c8833b3f1430bfc1ec69a2a4d1
636135c8833b3f1430bfc1ec69a2a4d1
```

### Verifying the fix

Build a patched image applying the fix on top of the current kicbase:

```
$ podman build -t kicbase-patched - <<EOF
FROM $KICBASE
RUN truncate -s 0 /etc/machine-id && rm -f /var/lib/dbus/machine-id
EOF
```

Confirm the image state:

```
$ podman run --rm --entrypoint bash kicbase-patched -c '
    echo "/etc/machine-id size: $(wc -c < /etc/machine-id) bytes"
    ls /var/lib/dbus/machine-id 2>/dev/null || echo "/var/lib/dbus/machine-id: (removed)"
  '
/etc/machine-id size: 0 bytes
/var/lib/dbus/machine-id: (removed)
```

Each container now gets a unique machine ID:

```
$ for i in 1 2 3; do
    podman run --rm --entrypoint bash kicbase-patched -c '
      systemd-machine-id-setup 2>/dev/null; cat /etc/machine-id'
  done
4c971ecb6dc34159ad1da13a867a51c3
2d717cadb4804b1b9863f7b8b2da998d
b25726fead9246f691e97b3a63834de2
```

### DIFF
--- a/deploy/kicbase/Dockerfile
+++ b/deploy/kicbase/Dockerfile
@@ -290,6 +290,12 @@ RUN rm -rf \
   /usr/share/doc/* \
   /usr/share/man/* \
   /usr/share/local/*
+
+# Per https://systemd.io/CONTAINER_INTERFACE/, /etc/machine-id must be present
+# but empty in a container image so it is freshly initialized on each boot.
+# Remove /var/lib/dbus/machine-id so that systemd-machine-id-setup cannot fall
+# back to a baked-in value, which would make every container share the same ID.
+RUN truncate -s 0 /etc/machine-id && rm -f /var/lib/dbus/machine-id
 RUN echo "kic! Build: ${COMMIT_SHA} Time :$(date)" > "/kic.txt"
 
 # squash all layers into one
--- a/deploy/kicbase/entrypoint
+++ b/deploy/kicbase/entrypoint
@@ -357,14 +357,6 @@ retryable_fix_cgroup() {
     exit 31
 }
 
-fix_machine_id() {
-  # Deletes the machine-id embedded in the node image and generates a new one.
-  # This is necessary because both kubelet and other components like weave net
-  # use machine-id internally to distinguish nodes.
-  echo 'INFO: clearing and regenerating /etc/machine-id' >&2
-  rm -f /etc/machine-id
-  systemd-machine-id-setup
-}
 
 fix_product_name() {
   # this is a small fix to hide the underlying hardware and fix issue #426
@@ -545,7 +537,6 @@ configure_containerd
 configure_proxy
 fix_mount
 retryable_fix_cgroup
-fix_machine_id
 fix_product_name
 fix_product_uuid
 select_iptables


RobinMcCorkell:
~The CLA signing page (and the support link) take me to a blank page on https://sso.linuxfoundation.org/login, looks like there's some CORS issues preventing the page from loading. I can't sign the CLA without that.~ this is now fixed

Also I don't think I can sign the CLA on behalf of Copilot :see_no_evil:

medyagh:
@RobinMcCorkell this code is common with Kind, do you know if this affects kind too ? do we have any users complaining about this in the issues?

RobinMcCorkell:
I believe this is the cause of #22606, 

It is also mentioned in #16962 - "I tried the bridge cni in minikube and it wouldn't work regardless of IP forwarding being enabled or disabled"

It's possible these Kind bug reports also refer to this issue: https://github.com/kubernetes-sigs/kind/issues/2996 https://github.com/kubernetes-sigs/kind/issues/3412

afbjorklund (approved):
LGTM

nirs:
@RobinMcCorkell Sorry for the long delay. Can you rebase and test on current master? We switch to debian trixie recently and we need to validate this fix.

I think we also need an integration test, verifying that we get unique machine id for every node.

RobinMcCorkell:
Rebased and re-verified - the dbus ID in /var/lib/dbus/machine-id causes this deterministic result during machine ID generation:

```
root@a7697dbac726:/# systemd-machine-id-setup
Initializing machine ID from D-Bus machine ID.
```

Which when removed per this PR changes to:

```
root@a7697dbac726:/# systemd-machine-id-setup
```

RobinMcCorkell:
I think the whole reason this bug exists (vs being fixed in the base image) is that the systemd package is installed in the kicbase Dockerfile, and that package installation generates a dbus and machine ID which needs to be reset.

PR #22826, opened 2026-04-11
Addon: run Rook-Ceph inside Minikube

### LINKED ISSUE #21527: feature: add ceph storage provisioner addon
with the gluster addon being removed (see PR #20370 ), now might be a good time to consider adding a rook-ceph addon as a substitute.

[ceph](https://github.com/ceph/ceph) is a mature SDS system and [rook](https://github.com/rook/rook) an operator to deploy and manage ceph in kubernetes.

:Adavantages:
*  the  rook project  already uses minikube for testing (both CI and local developer testing).
* rook-ceph is much more modern than gluster with a CSI driver (while gluster only had legacy k8s i-tree drivers). and the fully-fledged rook operator.

Ppssible challenge: rook usally expects a few extra disks (typically of 40GiB) for setting up ceph's object storage deamons (OSDs).

bhavyaBeliever:
Steps to run:
1. Start a minikube cluster with disk-size atleast 40gb, 4 cpus and 8gb of memory:
``` 
./out/minikube start --disk-size=40g  --memory=8000 --cpus=4
```
2. Configure the rook-ceph addon with: (Optional)
```
./out/minikube addons configure rook-ceph
```
Two things that can be configured.
*  The default (loop) disk size which should be atleast 6 GiB 
* Device that can be used by Ceph (e.g. /dev/vdb), this needs to be created at the start time with flag `extra-disks`   

3. Enable the addon, it takes few minutes. (majority of the time in pulling images)
```
./out/minikube addons enable rook-ceph
```

4. Verify the addon with:
```
minikube kubectl -- get cephcluster -n rook-ceph
```
Check if OSD pod is running with:
```
minikube kubectl -- get pods -n rook-ceph
```
## Current Implementation Details and configuration of Ceph cluster:

1. There is only one disk that ceph cluster uses. The default disk is a loop device inside the VM. 
2. No replication, all components like mon, mgr, osd are of count 1. 
3. Two storage interfaces: RBD(works for VM-based drivers) and CephFS(works for all drivers).
4. Ceph Dashboard is enabled.
5. Two storage classes: `rook-cephfs`, `rook-ceph-block`.
6. Ceph runs only on one node since it runs only on one disk.
7. The `dataDirHostPath` is kept `/data/rook` instead of default `/var/lib/rook`, [supports persistant storage](https://minikube.sigs.k8s.io/docs/handbook/persistent_volumes/#a-note-on-mounts-persistence-and-minikube-hosts).
8. `minikube addons disable rook-ceph` is implemented, mapped custom teardown callback, used steps that were mentioned [here](https://rook.io/docs/rook/latest-release/Getting-Started/ceph-teardown/).

## Pending Work:

- [ ]  **Multiple Disks / OSDs Support**
- [ ]  **Object Storage (S3 API) Support**
- [ ]  **Image Caching Strategy**
- [ ]  **Support RBD storage in KIC-based driver**
- [ ]  **Adding Testcases**
- [ ]  **Enable CSIAddons & VolumeReplication**
- [ ]  **Use New version of Rook-ceph**: Currently using v16.7

fixes #21527

### DIFF
--- a/cmd/minikube/cmd/config/configure.go
+++ b/cmd/minikube/cmd/config/configure.go
@@ -23,6 +23,7 @@ import (
 	"net"
 	"os"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -82,6 +83,9 @@ var addonsConfigureCmd = &cobra.Command{
 
 		case "auto-pause":
 			processAutoPauseConfig(profile, addonConfig, options)
+
+		case "rook-ceph":
+			processRookCephConfig(profile, addonConfig, options)
 
 		default:
 			out.FailureT("{{.name}} has no available configuration options", out.V{"name": addon})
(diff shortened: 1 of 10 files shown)


bhavyaBeliever:
@medyagh @prezha @nirs @obnoxxx would like to get feedback from your side!!

nirs:
@bhavyaBeliever nobody can review 17k lines change, and the discussion on the issue barely started. We need to finish planning first before spending time on implementation.

bhavyaBeliever:
@nirs Out of 17k lines much of this is yaml files that are copied from the rook repository([crds.yaml](https://github.com/rook/rook/blob/master/deploy/examples/crds.yaml), [common.yaml](https://github.com/rook/rook/blob/master/deploy/examples/common.yaml), [operator.yaml](https://github.com/rook/rook/blob/master/deploy/examples/operator.yaml)). The [rook-ceph-crds.yaml](https://github.com/kubernetes/minikube/pull/22826/changes#diff-8de8ad182e6f13497b9855f6fe6c3d0c7426dabcc15325e6799a711c642889f8) and [rook-ceph-operator.yaml](https://github.com/kubernetes/minikube/pull/22826/changes#diff-0507b279b9a6b7d37d0c7d3c9e400d969d8a113cf3c9ff553e2eac017485a1f6) are almost 16k

bhavyaBeliever:
Also, as we have taken this yaml files from rook repo what should be the license for such files?

nirs:
> @nirs Out of 17k lines much of this is yaml files that are copied from the rook repository

1k is much better than 17k but still a lot when we do't have any plan to discuss.

I would not add 16k of rook resources to minikube. It does not make sense to keep this in minikube. We depend on rook images from the registry so we can depend on pulling rook resources from GitHub in the same way.

Think abut updating rook version - the addon must update when rook updates. Pulling the resources from GitHub using kustomization makes update easier - you change the URL  and maybe update the kustomization for new options added in the release.

I think the modern way to implement addons is helm, which should be supported now in rook. This will keep the minikube code minimal.

But the big question is not how to implement the addon, but why. Before we discuss the use cases we want to support there is no point to work on the implementation.

bhavyaBeliever:
> But the big question is not how to implement the addon, but why. Before we discuss the use cases we want to support there is no point to work on the implementation.

Agreed, The whole point of making this PR is to test different configuration we might be looking forward with use cases we wanted inside minikube cluster by maybe just modify the rook-ceph-cluster.yaml.

Issue #22326, opened 2025-12-25
crio on in KIC uses different runtime-root than ISO ("open /run/runc: no such file or directory")


medyagh:
we need to investigate why on docker driver with crio runtime many tests fail with 
"msg="open /run/runc: no such file or directory"

The same tests passes on VM drivers (ISO)

example test log
https://storage.googleapis.com/minikube-builds/logs/22301/42985/Docker_Linux_crio.html#fail_TestStartStop%2fgroup%2fold-k8s-version%2fserial%2fEnableAddonWhileActive

```
-- /stdout --
** stderr ** 
	X Exiting due to MK_ADDON_ENABLE_PAUSED: enabled failed: check paused: list paused: runc: sudo runc list -f json: Process exited with status 1
	stdout:
	
	stderr:
	time="2025-12-25T19:01:33Z" level=error msg="open /run/runc: no such file or directory"

```

Maybe there is a Path Isseu in Kicbase?

medyagh:
#KIC 

$ crio config | grep default_runtime
# default_runtime = "crun"
docker@minikube:~$ crio version
Version:        1.34.3



# VM

$ crio config | grep default_runtime
# default_runtime = "runc"

$ crio version
INFO[2025-12-26 00:39:33.461096541Z] Starting CRI-O, version: 1.29.1, git: unknown(clean) 
Version:        1.29.1



We might just wanna make the ISO and KIC have matching CRIO version.

afbjorklund:
It is just that cri-o has not been updated, the default was changed in v1.31.0

commit https://github.com/cri-o/cri-o/commit/5e21d495c8ac96866fa3e5b469b7303e0a9b2c7d

Calling "runc" is also hardcoded in cruntime "cri", for the pause functionality. So should be separated out from "cri".

medyagh:
> It is just that cri-o has not been updated, the default was changed in v1.31.0
> 
> commit [cri-o/cri-o@5e21d49](https://github.com/cri-o/cri-o/commit/5e21d495c8ac96866fa3e5b469b7303e0a9b2c7d)
> 
> Calling "runc" is also hardcoded in cruntime "cri", for the pause functionality. So should be separated out from "cri".

I updated the crio to 1.35 and still CRO tests on docker driver only fails I think the problem is our code is Expecting to call "runc" while it should do "crun"
(our iso config is outdated and still uses runc)

rafaelmfried:
/remove-lifecycle rotten

[...]

Issue #20993, opened 2025-06-29
iso: Update buildroot to 2025.05


nirs:
### What Happened?

Update buildroot to consume new packages versions and fixes.
https://gitlab.com/buildroot.org/buildroot/-/blob/2025.05/CHANGES

### Attach the log file

-

### Operating System

None

### Driver

None

afbjorklund:
Historically we used the LTS releases, since it was trouble keeping up with the updates...

i.e. 2025.02 (supported to March 2026), rather than 2025.05 (supported until 2025.08)

nirs:
> Historically we used the LTS releases, since it was trouble keeping up with the updates...
> 
> i.e. 2025.02 (supported to March 2026), rather than 2025.05 (supported until 2025.08)

Good point, we need to add a comment for buildroot version about this.

Updating issue to upgrade to next LTS (2026.02)

nirs:
@afbjorklund what about the bugfix releases like:
https://gitlab.com/buildroot.org/buildroot/-/blob/2025.02.4/CHANGES

The changeling lists many CVEs.

afbjorklund:
I meant 2025.02.x, including all bugfix releases (for the LTS release)

Note that 2026.02 will be out _next_ year, i.e. before 2025.02 goes EOL

----

https://buildroot.org/news.html

* 2025.02.01 released
* 2024.02.13 released, 2024.02.x series EOL
* 2024.02.12 released, migrate to 2025.02.x
* 2025.02 released
* 2023.02.11 released, 2023.02.x series EOL
* 2024.02 released
* 2022.02.12 released, 2022.02.x series EOL
* 2023.02 released

afbjorklund:
Missing 6 months:

2025.02.6 released
9 September 2025

2025.02.5 released
11 August 2025

2025.02.4 released
23 June 2025

2025.02.3 released
19 May 2025

2025.02.2 broken

2025.02.01 released
22 April 2025

2025.02 released
17 March 2025

* #20720

afbjorklund:
When upgrading the Buildroot version, you also want to update the Linux minor version:

BUILDROOT_BRANCH ?= 2025.02.9

BR2_LINUX_KERNEL_CUSTOM_VERSION_VALUE="6.6.119"

To stick with the default version from the upstream Buildroot release, for the current branch:

https://github.com/buildroot/buildroot/blob/2025.02.x/linux/linux.hash

```
[...]

PR #22862, opened 2026-04-19
iso: support building the arm64 ISO on an arm64 host

### LINKED ISSUE #22866: iso: building arm64 ISO on an aarch64 host drops Go-target packages
### What Happened?

Building the arm64 minikube ISO from an aarch64 Linux host (e.g. Colima on an Apple Silicon Mac, or a Linux arm64 machine) silently produces a broken ISO. `make minikube-iso-aarch64` completes with exit 0 but the result is ~319 MB instead of the expected ~403 MB and `minikube start` fails during provisioning on every arm64 driver (vfkit, qemu2, krunkit, virtualbox) with:

```
diff: can't stat '/lib/systemd/system/docker.service': No such file or directory
Created symlink '/etc/systemd/system/multi-user.target.wants/docker.service' -> '/usr/lib/systemd/system/docker.service'.
Job for docker.service failed because the control process exited with error code.
```

Not driver-specific: any driver that boots the broken arm64 ISO fails the same way.

## Root cause

Buildroot's `BR2_PACKAGE_HOST_GO_BOOTSTRAP_STAGE1_ARCH_SUPPORTS` in `package/go/go-bootstrap-stage1/Config.in.host` only defaults `y` for `BR2_HOSTARCH` of `x86`, `x86_64`, `arm`:

```
config BR2_PACKAGE_HOST_GO_BOOTSTRAP_STAGE1_ARCH_SUPPORTS
    bool
    # See src/cmd/dist/unix.c for the list of supported architectures
    default y if BR2_HOSTARCH = "x86"
    default y if BR2_HOSTARCH = "x86_64"
    default y if BR2_HOSTARCH = "arm"
```

On an aarch64 host the symbol stays unset, which cascades to `BR2_PACKAGE_HOST_GO_TARGET_ARCH_SUPPORTS` staying unset, which silently drops every Go-target package from the build: `cri-dockerd-aarch64`, `cni-plugins-latest-aarch64`, `podman`, `crio-bin`, plus their transitive dependencies. `docker-bin-aarch64` has no host-go dependency and still installs, which is why the `docker`/`dockerd` binaries end up in the ISO but none of the CRI service glue does.

Verified in the build's `.config`:

```
BR2_PACKAGE_DOCKER_BIN_AARCH64=y
BR2_PACKAGE_CONTAINERD_BIN_AARCH64=y
# BR2_PACKAGE_CRI_DOCKERD_AARCH64 not set
# (no BR2_PACKAGE_PODMAN, BR2_PACKAGE_CRIO_BIN, BR2_PACKAGE_CNI_PLUGINS_LATEST_AARCH64 either)
```

The build log never mentions `cri-dockerd`, `crio`, or `podman`.

## Reproducer

1. Apple Silicon Mac with Colima installed:
   ```
   colima start --vm-type vz --arch aarch64 --cpu 6 --memory 8 --disk 80
   docker context use colima
   ```
2. From a minikube checkout: `make buildroot-image && make out/minikube-arm64.iso`.
3. Observe the resulting ISO is 319 MB and `minikube start` fails on the `docker.service` restart step for any driver.

Also reproduces on native aarch64 Linux hosts (e.g. Debian arm64, Fedora aarch64); it isn't macOS-specific but macOS users are the main population affected since most contributors' minikube arm64 ISO builds happen on Apple Silicon.

## Fix

Add `aarch64` to the stage1 default-y list:

```diff
--- a/package/go/go-bootstrap-stage1/Config.in.host
+++ b/package/go/go-bootstrap-stage1/Config.in.host
@@ -4,3 +4,4 @@ config BR2_PACKAGE_HOST_GO_BOOTSTRAP_STAGE1_ARCH_SUPPORTS
     default y if BR2_HOSTARCH = "x86"
     default y if BR2_HOSTARCH = "x86_64"
     default y if BR2_HOSTARCH = "arm"
+    default y if BR2_HOSTARCH = "aarch64"
```

Go upstream ships prebuilt aarch64 binaries (`go*.linux-arm64.tar.gz`) and `host-go-bin` already declares aarch64 support via `BR2_PACKAGE_HOST_GO_BIN_HOST_ARCH_SUPPORTS`. Go has also supported source bootstrap on aarch64 since Go 1.5, so allowing aarch64 in the stage1 gate is safe for both prebuilt and source Go providers.

With the patch applied the ISO is 403 MB (matches the published one) and boots cleanly on vfkit, qemu2, krunkit, and virtualbox.

## Related

- PR #22862 lands this as a local buildroot patch in `deploy/iso/minikube-iso/patches/buildroot/`.
- The proper fix belongs upstream in buildroot; patch for submission is prepared against buildroot master.

### Attach the log file

Log shows the successful-looking build with missing packages; happy to attach if useful.

### Operating System

macOS (Default)

### Driver

N/A

abedegno:
Fix for arm64-host buildroot builds of the minikube arm64 ISO.

## Problem

On an arm64 Linux host (e.g. Colima on Apple Silicon), `make minikube-iso-aarch64` completes with exit 0 but silently produces a **~319 MB** ISO instead of the expected **~403 MB**. The ISO boots fine, but `minikube start` fails during provisioning:

```
diff: can't stat '/lib/systemd/system/docker.service': No such file or directory
Created symlink '/etc/systemd/system/multi-user.target.wants/docker.service' -> '/usr/lib/systemd/system/docker.service'.
Job for docker.service failed because the control process exited with error code.
```

Same failure mode on vfkit, qemu2, and virtualbox (so it isn't driver-specific).

## Root cause

buildroot's `BR2_PACKAGE_HOST_GO_BOOTSTRAP_STAGE1_ARCH_SUPPORTS` only default-y's for x86, x86_64 and arm hosts. When the buildroot host is aarch64 the symbol stays unset, which leaves `BR2_PACKAGE_HOST_GO_TARGET_ARCH_SUPPORTS` unset, which silently drops every Go-target package: `cri-dockerd-aarch64`, `cni-plugins-latest-aarch64`, `podman`, `crio-bin`, plus transitive deps. `docker-bin-aarch64` has no host-go dependency and still installs, which is why the docker binary ends up in the ISO but none of the CRI service glue does.

The gate is the right shape for `host-go-src` (source bootstrap of Go has real host arch restrictions), but minikube resolves host-go via `host-go-bin` which has a prebuilt Go binary for aarch64 and works fine. The stage1 gate is overly strict for the `host-go-bin` case.

## Fix

Adds a buildroot patch at `deploy/iso/minikube-iso/patches/buildroot/0001-host-go-aarch64-stage1.patch`:

```diff
--- a/package/go/go-bootstrap-stage1/Config.in.host
+++ b/package/go/go-bootstrap-stage1/Config.in.host
@@ -4,3 +4,4 @@ config BR2_PACKAGE_HOST_GO_BOOTSTRAP_STAGE1_ARCH_SUPPORTS
 	default y if BR2_HOSTARCH = "x86"
 	default y if BR2_HOSTARCH = "x86_64"
 	default y if BR2_HOSTARCH = "arm"
+	default y if BR2_HOSTARCH = "aarch64"
```

Extends the `buildroot:` target in the top-level Makefile to apply any `*.patch` files dropped into that directory right after the initial `git clone`. Any future per-host fixes can be added there without further Makefile work.

The proper long-term fix belongs upstream in buildroot; this is a pragmatic near-term workaround so local arm64-host builds produce the same artifact as CI.

## Test

On darwin/arm64 (M-series Mac, Colima `--vm-type vz --arch aarch64`):

- Before this PR: ISO 319 MB, provisioning fails on all drivers at docker.service restart.
- After this PR: ISO 403 MB (size-matches the published `minikube-v1.38.0-arm64.iso`), `cri-docker.service`, `crio.service`, `podman.service`, `cri-dockerd`/`crio`/`podman` binaries all present.
- `minikube start --driver=vfkit|qemu2|virtualbox` all produce a `Ready` control-plane node on the newly-built ISO.
- `make minikube-iso-x86_64` unaffected (patch only activates on aarch64 hosts).

Upstream CI (x86_64 builders) is unaffected.

## Related
- Fixes #22866 (the bug this PR works around).
- Found while validating PR #22861 (darwin/arm64 VirtualBox support); the two PRs are independent.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -292,6 +292,11 @@ buildroot:
 	if [ ! -d $(BUILD_DIR)/buildroot ]; then \
 		mkdir -p $(BUILD_DIR); \
 		git clone --depth=1 --branch=$(BUILDROOT_BRANCH) https://github.com/buildroot/buildroot $(BUILD_DIR)/buildroot; \
+		for p in $(CURDIR)/deploy/iso/minikube-iso/patches/buildroot/*.patch; do \
+			[ -f "$$p" ] || continue; \
+			echo "Applying buildroot patch: $$p"; \
+			(cd $(BUILD_DIR)/buildroot && patch -p1 -i "$$p") || exit 1; \
+		done; \
 	fi;
 
 # Change buildroot configuration for the minikube ISO
--- a/deploy/iso/minikube-iso/patches/buildroot/0001-host-go-aarch64-stage1.patch
+++ b/deploy/iso/minikube-iso/patches/buildroot/0001-host-go-aarch64-stage1.patch
@@ -1,0 +1,29 @@
+Add aarch64 as a supported host arch for the Go bootstrap stage 1.
+
+The stage1 gate in the upstream buildroot config only default-y's for
+x86, x86_64 and arm hosts. On an aarch64 host this leaves
+BR2_PACKAGE_HOST_GO_BOOTSTRAP_STAGE1_ARCH_SUPPORTS unset, which in
+turn leaves BR2_PACKAGE_HOST_GO_TARGET_ARCH_SUPPORTS unset, which
+silently drops every Go-target package (cri-dockerd, cni-plugins,
+podman, crio-bin, ...) from the build. The resulting arm64 minikube
+ISO is ~80 MB short and the provisioner fails at the docker.service
+restart step because no docker.service file was ever installed.
+
+The gate is correct for host-go-src (source bootstrap of Go has real
+host arch restrictions), but minikube resolves host-go via
+host-go-bin (PROVIDES_HOST_GO="host-go-bin"), and prebuilt Go binaries
+support aarch64 natively. Allow aarch64 in the stage1 default list so
+Go-target packages build correctly when the buildroot host is aarch64
+(e.g. Colima on an Apple Silicon Mac).
+
+This is a local patch; the proper fix belongs upstream in buildroot.
+
+Signed-off-by: Jon Williams <jon@jonwilliams.org.uk>
+
+--- a/package/go/go-bootstrap-stage1/Config.in.host
++++ b/package/go/go-bootstrap-stage1/Config.in.host
+@@ -4,3 +4,4 @@ config BR2_PACKAGE_HOST_GO_BOOTSTRAP_STAGE1_ARCH_SUPPORTS
+ 	default y if BR2_HOSTARCH = "x86"
+ 	default y if BR2_HOSTARCH = "x86_64"
+ 	default y if BR2_HOSTARCH = "arm"
++	default y if BR2_HOSTARCH = "aarch64"


nirs (commented):
Changes looks good to me, but the best way to validate the change is to post a patch to buildroot. The fix is like much simpler if we don't have to keep a patch file.

nirs:
@abedegno maybe open an issue for building on macOS? the existing issue is about building on Fedora - but the issue fixed here is not related to Fedora.

abedegno:
Quick triage of the failed lanes - they all look environmental to me, and I don't think anything this PR changes can affect them. Flagging so we can kick a retest rather than chase them.

afbjorklund (changes requested):
You don't have to apply patches manually, if you put the patch in the proper subdirectory it will be applied auomatically by the buildroot system.

EDIT: Hmm, maybe that is only for code changes - not for changes to the makefile itself.

afbjorklund (approved):

afbjorklund:
It will be hard to get patches from upstream, if never updating the operating system...

* https://github.com/kubernetes/minikube/issues/20993

(this bug is on master too, though. buildroot doesn't really try to build on non-x86)

abedegno:
Two failed lanes from the previous retest - appear to be environmental again.

Retesting both lanes individually:

/test pull-minikube-kvm-docker-linux-x86
/test pull-minikube-docker-crio-linux-x86

abedegno:
> @abedegno maybe open an issue for building on macOS? the existing issue is about building on Fedora - but the issue fixed here is not related to Fedora.

@nirs Done https://github.com/kubernetes/minikube/issues/22866

nirs:
> It will be hard to get patches from upstream, if never updating the operating system...
> 
> * [iso: Update buildroot to bugfix release 2025.02.x #20993](https://github.com/kubernetes/minikube/issues/20993)

Sure, this is another gap we need to close.

> (this bug is on master too, though. buildroot doesn't really try to build on non-x86)

We also build on x86_64 in CI, but we want to build on aarch64 in GitHub actions for faster build.

abedegno:
All required checks green:

- `pull-minikube-kvm-docker-linux-x86` passed on retest. The previous `TestMultiNode/serial/FreshStart2Nodes` failure was a flake as suspected (pods mid-startup when the test snapshotted, then gone by post-mortem).
- `pull-minikube-docker-crio-linux-x86` (not required) is still hitting the known `runc: open /run/runc: no such file or directory` pattern tracked at #22326. Not PR-related.

PR #22908, opened 2026-04-28
Use authenticated GitHub client for Kubernetes version check

ushitora-anqou:
Switch IsInGitHubKubernetesVersions from an unauthenticated github.NewClient to update.GHClient, which uses GITHUB_TOKEN when set. This avoids hitting the GitHub API anonymous rate limit when checking whether a given Kubernetes version exists upstream.

Add k8s.io/minikube/hack to go.mod to expose update.GHClient.


cmd/minikube/cmd/config/kubernetes_version.go
```diff
@@ -20,8 +20,8 @@ import (
 	"context"
 	"net/http"
 
-	"github.com/google/go-github/v84/github"
 	"golang.org/x/mod/semver"
+	"k8s.io/minikube/hack/update"
```
medyagh, line 24: I dont think we wanna use code from our hack module into minikube main module

medyagh (changes requested):
is there any reason for this PR ? anyone has created issue ?

ushitora-anqou:
Sorry, I misread the Contributor Guide. I'll close this PR and read it more carefully.

Issue #22901, opened 2026-04-27
add "docker deprecation notice" to allow list in TestErrorSpam


medyagh:
https://storage.googleapis.com/minikube-builds/logs/22895/44566/Docker_Linux.html#fail_TestErrorSpam%2fsetup

abedegno:
The deprecation notice is already accepted by `TestErrorSpam` and isn't what fails it. 

The deprecation notice matches via the `regexp.QuoteMeta(constants.DefaultContainerRuntimeChangeWarning)` entry in `stderrAllow`, and the log line confirms the `acceptable stderr:`

The unexpected stderr is the `proxy.go:152` `Errorf` call, which now fires because client-go 0.36 wraps the RoundTripper in `*transport.trackedTransport` and the cast at `proxy.go:148` no longer succeeds. Adding the deprecation message to the allow-list wouldn't change the test outcome since the cast error lines aren't covered by it.

Issue #22896 covers this with the cast site and a fix sketch (either demote the log, walk the wrapper via `WrappedRoundTripper()`, or use `rest.Config.Proxy`). Probably worth closing this as a duplicate and tracking the fix there?

medyagh:
@abedegno would you like to make a PR to fix that?

abedegno:
@medyagh - your wish is my command!
https://github.com/kubernetes/minikube/pull/22910

nirs:
Fixed by #22910

Issue #22900, opened 2026-04-27
fix flaky functional test in x86


medyagh:
recently seeing more failure of Functional Test (github action ones) of this test
✗ TestFunctional/serial/StartWithCustomCerts
example
https://github.com/kubernetes/minikube/actions/runs/24980982828/job/73143214581?pr=22899

this test spins up a mitmproxy to emulate coprorate proxy in the GH action...I wonder if the mitim is failing or needs to be bumped

```
	I0427 07:09:42.045937    3335 round_trippers.go:632] "Response" verb="GET" url="https://192.168.49.2:8441/api/v1/nodes/functional-330595" status="" milliseconds=48498
	W0427 07:09:42.046001    3335 node_ready.go:55] error getting node "functional-330595" condition "Ready" status (will retry): Get "https://192.168.49.2:8441/api/v1/nodes/functional-330595": context deadline exceeded
	I0427 07:09:42.046025    3335 node_ready.go:38] duration metric: took 6m0.001031172s for node "functional-330595" to be "Ready" ...
	I0427 07:09:42.046564    3335 out.go:203] 
	W0427 07:09:42.046867    3335 out.go:285] X Exiting due to GUEST_START: failed to start node: wait 6m0s for node: waiting for node to be ready: WaitNodeCondition: context deadline exceeded
	X Exiting due to GUEST_START: failed to start node: wait 6m0s for node: waiting for node to be ready: WaitNodeCondition: context deadline exceeded
	W0427 07:09:42.046877    3335 out.go:285] * 

```

full log
https://github.com/kubernetes/minikube/actions/runs/24980982828/artifacts/6655068598

oglego:
I'm interested in trying to work on this!

Just to double check, it looks like mitmproxy can be bumped up to version 12.2.2, and then would it be beneficial to also add in a TCP readiness check that polls port 8080 for up to 30 seconds after cert generation, ensuring mitmdump is fully ready before Minikube starts?

medyagh:
I just bumped the mimproxy and doesnt seem to help
I also noticed it only fails on X86 agents... I think thouse machiens are slower

PR #22910, opened 2026-04-29
pkg/proxy: walk RoundTripper wrappers to strip proxy on inner *http.Transport

### LINKED ISSUE #22896: TestErrorSpam/setup fails on every PR after k8s-dependencies bump (proxy.go RoundTripper cast)
### What Happened?

`TestErrorSpam/setup` is failing on every PR's prow integration tests since the k8s-dependencies bump in [#22876](https://github.com/kubernetes/minikube/pull/22876) (commit `f711fb429`) merged on 2026-04-24. The failure is reproducible on at minimum docker, containerd, kvm and crio runtimes across both x86 and arm.

The test asserts that `minikube start` produces no unexpected stderr. After the dep bump, every `minikube start` now emits:

```
E0425 20:01:48.250329   63697 proxy.go:152] Error while casting RoundTripper (of type *transport.trackedTransport) to *http.Transport : false
E0425 20:01:48.368989   63697 proxy.go:152] Error while casting RoundTripper (of type *transport.trackedTransport) to *http.Transport : false
E0425 20:01:48.369055   63697 proxy.go:152] Error while casting RoundTripper (of type *transport.trackedTransport) to *http.Transport : false
```

`TestErrorSpam` flags the lines as unexpected stderr and fails:

```
--- FAIL: TestErrorSpam/setup (14.80s)
    error_spam_test.go:100: unexpected stderr: "E0425 20:01:48.250329   63697 proxy.go:152] Error while casting RoundTripper (of type *transport.trackedTransport) to *http.Transport : false"
    ...
```

### Attach the log file

Any PR opened against current `master` reproduces. Verified across at least:

- #22862 (this is what surfaced the regression): https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/22862/pull-minikube-docker-docker-linux-x86/2048128180010618880
- #22895 (an unrelated kubevirt addon bump, only TestErrorSpam/setup fails): https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/22895/pull-minikube-docker-containerd-linux-x86/2047970412591058944

The failure is the sole `Fail` in test_summary.json on most lanes, with `Pass` numbers in the high 300s, so the rest of the integration suite is healthy. Master itself almost certainly fails the same way; this is blocking PR merges across the repo.

### Operating System

None

### Driver

None

abedegno:
## Summary

After [#22876](https://github.com/kubernetes/minikube/pull/22876) bumped `k8s.io/client-go` from 0.35.1 to 0.36.0, [`kubernetes/client-go@f5fc1e5f`](https://github.com/kubernetes/client-go/commit/f5fc1e5f67e9f487e6e1b96384c16788e1a9db9a) wraps the cached HTTP transport in `*trackedTransport` to anchor garbage collection of the TLS transport cache. `proxy.UpdateTransport` was downcasting directly to `*http.Transport`; that assertion now fails on every minikube start, logging three `Errorf` lines and tripping `TestErrorSpam/setup` on every PR against current master.

This PR walks `RoundTripperWrapper` layers via `WrappedRoundTripper()` until the underlying `*http.Transport` is reached, then nils `Proxy` in place. The outer wrapper is returned unchanged so client-go's TLS cache GC anchoring continues to work.

## Changes

- Extracts the proxy-disable logic into `disableProxyOnTransport` so it can be unit-tested without `UpdateTransport`'s closure scaffolding.
- `UpdateTransport`'s `WrapTransport` closure delegates to the new helper; behavior is unchanged on the no-wrapping path that worked before client-go 0.36.
- Adds `TestUpdateTransport/wrapped`, the regression test (wraps `*http.Transport` in a fake `RoundTripperWrapper`, asserts inner `Proxy` is nilled).
- Adds `TestUpdateTransport/proxy-stripping`, pinning the no-wrap branch so it does not silently regress in future refactors.

## Test plan

- [x] `go test ./pkg/minikube/proxy/...` passes locally
- [x] `go build ./...` clean
- [x] `gocyclo` clean on `pkg/minikube/proxy/proxy.go`
- [x] End-to-end: `minikube start` produces zero `Error while casting RoundTripper` lines on stderr (was three per start)
- [ ] CI green

## Related

Fixes #22896. #22901 reports the same `TestErrorSpam/setup` symptom; both should resolve once this lands.

### DIFF
--- a/pkg/minikube/proxy/proxy.go
+++ b/pkg/minikube/proxy/proxy.go
@@ -27,6 +27,7 @@ import (
 
 	"errors"
 
+	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 	"k8s.io/minikube/pkg/minikube/config"
@@ -145,15 +146,32 @@ func UpdateTransport(cfg *rest.Config) *rest.Config {
 		if wt != nil {
 			rt = wt(rt)
 		}
-		if ht, ok := rt.(*http.Transport); ok {
-			ht.Proxy = nil
-			rt = ht
-		} else {
-			klog.Errorf("Error while casting RoundTripper (of type %T) to *http.Transport : %v", rt, ok)
+		if t := wrappedHTTPTransport(rt); t != nil {
+			t.Proxy = nil
+		} else if rt != nil {
+			klog.Errorf("transport %T could not be unwrapped to *http.Transport; proxy stripping was skipped", rt)
 		}
 		return rt
 	}
 	return cfg
+}
+
+// wrappedHTTPTransport returns the *http.Transport at the bottom of any
+// RoundTripperWrapper chain wrapping rt, or nil if the chain does not end
+// in one (or rt itself is nil). See
+// https://github.com/kubernetes/minikube/issues/22896 for context.
+func wrappedHTTPTransport(rt http.RoundTripper) *http.Transport {
+	cur := rt
+	for {
+		switch t := cur.(type) {
+		case *http.Transport:
+			return t
+		case utilnet.RoundTripperWrapper:
+			cur = t.WrappedRoundTripper()
+		default:
+			return nil
+		}
+	}
 }
 
 // SetDockerEnv sets the proxy environment variables in the docker environment.
--- a/pkg/minikube/proxy/proxy_test.go
+++ b/pkg/minikube/proxy/proxy_test.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"testing"
 
+	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/client-go/rest"
 )
 
@@ -188,6 +189,24 @@ func TestExcludeIP(t *testing.T) {
 	}
 }
 
+// fakeRoundTripperWrapper satisfies utilnet.RoundTripperWrapper for tests
+// that need to simulate the way client-go 0.36's *trackedTransport (or any
+// future wrapping layer) interposes between WrapTransport callers and the
+// underlying *http.Transport.
+type fakeRoundTripperWrapper struct{ rt http.RoundTripper }
+
+func (f *fakeRoundTripperWrapper) RoundTrip(req *http.Request) (*http.Response, error) {
+	return f.rt.RoundTrip(req)
+}
+
+func (f *fakeRoundTripperWrapper) WrappedRoundTripper() http.RoundTripper { return f.rt }
+
+// Compile-time assertion: fakeRoundTripperWrapper must satisfy
+// utilnet.RoundTripperWrapper so the test exercises the same interface the
+// helper walks at runtime. Future client-go interface changes will surface
+// here as a build failure rather than a silently irrelevant test.
+var _ utilnet.RoundTripperWrapper = (*fakeRoundTripperWrapper)(nil)
+
 func TestUpdateTransport(t *testing.T) {
 	t.Run("new", func(t *testing.T) {
 		rcBefore := rest.Config{}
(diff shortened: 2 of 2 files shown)


pkg/minikube/proxy/proxy.go
```diff
@@ -145,17 +146,38 @@ func UpdateTransport(cfg *rest.Config) *rest.Config {
...
+// returned value is rt itself; mutation happens in place on the inner
+// transport, and the wrapper holds the inner by reference so the change
+// is visible through it. Returning the outer wrapper preserves whatever
+// reference the caller intends to hold (notably, client-go 0.36 anchors
+// its TLS transport cache on a *trackedTransport wrapper). If the chain
+// does not bottom out at a *http.Transport the proxy is not disabled and
+// the failure is logged at Errorf so any future client-go change that
+// re-breaks this surface is loudly visible rather than silently disabling
+// proxy stripping for users behind corporate proxies.
+func disableProxyOnTransport(rt http.RoundTripper) http.RoundTripper {
```
medyagh: consider adding the link to the Github issue and also a more simplified comment on what does this func do (maybe add a summery ? )

medyagh (commented):
thank you @abedegno please see comment and also copilot's comment

pkg/minikube/proxy/proxy.go
abedegno: @medyagh - done.

pkg/minikube/proxy/proxy.go
```diff
@@ -145,17 +146,38 @@ func UpdateTransport(cfg *rest.Config) *rest.Config {
...
+	for {
+		switch t := cur.(type) {
+		case *http.Transport:
+			t.Proxy = nil
+			return rt
+		case utilnet.RoundTripperWrapper:
+			cur = t.WrappedRoundTripper()
+		default:
+			klog.Errorf("transport %T could not be unwrapped to *http.Transport; proxy stripping was skipped", cur)
+			return rt
```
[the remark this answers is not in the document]
abedegno: Added an early `if rt == nil { return rt }` at the top of `disableProxyOnTransport`. `TestUpdateTransport/nil` now runs silently (it was previously logging the spurious Errorf even though the test passed).

pkg/minikube/proxy/proxy_test.go
```diff
@@ -188,6 +188,18 @@ func TestExcludeIP(t *testing.T) {
...
+// future wrapping layer) interposes between WrapTransport callers and the
+// underlying *http.Transport.
+type fakeRoundTripperWrapper struct{ rt http.RoundTripper }
+
+func (f *fakeRoundTripperWrapper) RoundTrip(req *http.Request) (*http.Response, error) {
+	return f.rt.RoundTrip(req)
+}
+
+func (f *fakeRoundTripperWrapper) WrappedRoundTripper() http.RoundTripper { return f.rt }
+
```
nirs, line 203: Faking this means that when the real client will change, the test will continue to pass while the real code will break. We need an integration test using a real client and detecting changes in the client.

pkg/minikube/proxy/proxy.go
```diff
@@ -145,17 +146,35 @@ func UpdateTransport(cfg *rest.Config) *rest.Config {
...
 	}
 	return cfg
 }
 
+// disableProxyOnTransport finds the *http.Transport at the bottom of any
+// RoundTripperWrapper chain wrapping rt and sets its Proxy to nil. The
+// outer rt is returned unchanged (mutation happens in place on the inner
+// transport). A nil rt is a no-op. See
+// https://github.com/kubernetes/minikube/issues/22896 for context.
+func disableProxyOnTransport(rt http.RoundTripper) http.RoundTripper {
```
nirs: returning rt from disableProxyOnTransport(rt) looks strange. Why this need to return anything?

pkg/minikube/proxy/proxy.go
```diff
@@ -145,17 +146,35 @@ func UpdateTransport(cfg *rest.Config) *rest.Config {
...
+	for {
+		switch t := cur.(type) {
+		case *http.Transport:
+			t.Proxy = nil
+			return rt
+		case utilnet.RoundTripperWrapper:
+			cur = t.WrappedRoundTripper()
+		default:
+			klog.Errorf("transport %T could not be unwrapped to *http.Transport; proxy stripping was skipped", cur)
+			return rt
```
nirs: Did consider simplifying the function to return the *http.Transport, replacing the type assertion in line 148?

```go
t := wrappedHTTPTransport(rt)
t.Proxy = nil
```

medyagh:
@abedegno thank you for fixing it ! I tink this also fixed the StartWithProxy test https://github.com/kubernetes/minikube/issues/22900

pkg/minikube/proxy/proxy.go
medyagh: @abedegno plz also check nir's comments and see if they are valid

pkg/minikube/proxy/proxy_test.go
abedegno: I think the integration coverage you're describing already exists - it's `TestErrorSpam/setup`, the test that surfaced this regression in the first place?

medyagh:
thank you @abedegno this actually fixes the failure on startwithproxy test too
https://github.com/kubernetes/minikube/issues/22900

medyagh:
@abedegno btw if you have slack feel free to join our minikube channel on slack 
https://kubernetes.slack.com/archives/C1F5CT6Q1/p1777572942023289

PR #22935, opened 2026-05-04
qemu2: enable WHPX on Windows and fix Windows startup path

### LINKED ISSUE #21529: Impossibility to add confing files resolv.conf and envirorment and the flag "-accel whpx" to qemu driver
### What Happened?

I'm using the qemu driver on Windows and I need to pass the proxy and DNS settings to the VM but it doesn't seem to work.
I added the files in the "$USER\.minikube\files\etc\environment and "$USER\.minikube\files\etc\resolv.conf" folder as well as setting the environment variables for the proxy, but the VM doesn't detect the configuration.
I also haven't found a way to pass the "-accel whpx" parameter to the qemu driver to get a faster machine since I'm using that parameter when I'm running qemu directly and it makes the machine much faster.
I ran this command:
` minikube start --driver=qemu2 --qemu-firmware-path="$USER\scoop\apps\qemu\current\share\edk2-x86_64-code.fd" --cpus=4 --memory=4096 --disk-size=20g --dns-domain=CUSTOM_DNS --container-runtime=docker --kubernetes-version=stable --image-repository=https://registry.k8s.io --docker-env HTTP_PROXY=$env:HTTP_PROXY --docker-env HTTPS_PROXY=$env:HTTPS_PROXY --docker-env NO_PROXY=$env:NO_PROXY --extra-config=kubelet.resolv-conf=/etc/resolv.conf --alsologtostderr `
Could someone tell me if there is anything I'm doing wrong?
Is it a real problem or is it me who didn't configure it right?

### Attach the log file

[logs.txt](https://github.com/user-attachments/files/22237583/logs.txt)

### Operating System

Windows

### Driver

QEMU

GiuseppeCapaldo93:
## Summary

This PR enables the qemu2 driver to work on Windows 10/11 (no admin rights required) by adding WHPX hardware acceleration support and fixing two Windows-specific bugs that caused `minikube start --driver=qemu2` to hang until the 360s StartHost timeout.

## Background

I ran into this while trying to get a local Kubernetes dev environment on an enterprise Windows 11 where virtualization is disabled. The qemu2 driver looked like the only option — QEMU itself works fine on Windows without admin — but the driver consistently timed out at the SSH wait step.

Investigation showed three separate problems compounding:

1. **No hardware acceleration on Windows.** The driver already detects `hvf` on macOS and `/dev/kvm` on Linux but has no Windows branch, so QEMU fell back to TCG and VM boot was prohibitively slow.
2. **CPU feature incompatibility with WHPX.** Once WHPX was enabled, the default `-cpu max` crashed the guest with "Unexpected VP exit code 4" because newer features like APX/MPX aren't supported under WHPX. This may be more visible on enterprise hardware configurations, but `qemu64` is a safe baseline that works everywhere.
3. **QEMU network port-forwarding fails on Windows with `-nic`.** The host side listens but forwarded packets never reach the guest — SSH from the Windows host gets "Connection refused" despite the port being in `LISTEN` state. Switching to the explicit `-device`/`-netdev` split form fixes this and matches the configuration used by working standalone QEMU+WHPX VMs on Windows.
4. **`os.Process.Signal(Signal(0))` is not implemented on Windows.** The Windows-specific `cmdStart` helper used this syscall to detect QEMU crashing immediately after spawn. On Windows it always returns `"not supported by windows"`, so every successful QEMU start was misreported as a failure and `Create()` aborted before `WaitForSSHAccess` ever ran. This was the actual root cause of the hang — the earlier network-level fixes were necessary but not sufficient.

With all three changes applied, `minikube start --driver=qemu2` completes in ~3 minutes on Windows 11 + WHPX and `kubectl get nodes` shows the cluster Ready.

## Changes

* **Commit 1 — Add WHPX hardware acceleration support for Windows.** Mirror the existing hvf/kvm pattern. Falls back to no acceleration if WHPX isn't available.

* **Commit 2 — qemu2: switch to qemu64 CPU when WHPX is enabled.** When `accel == "whpx"` and the requested CPU is `max` or `host`, override to `qemu64`.

* **Commit 3 — qemu2: fix Windows startup path (netdev + cmdStart detection).** Replace `-nic` with `-device virtio-net-pci,netdev=net0` + `-netdev user,...`; replace the broken `Signal(0)` early-crash check with a `select` over a non-blocking `cmd.Wait()` goroutine and a 2s timer; log the allocated host SSH/Engine ports for debugging.

## Test plan

- [x] `minikube start --driver=qemu2` on Windows 11 Enterprise 24H2 + QEMU 9.2.0 + WHPX, `kubectl get nodes` shows Ready
- [x] Fresh spin-up (after `minikube delete`) succeeds end-to-end
- [x] Core pods (etcd, apiserver, controller-manager, scheduler, kube-proxy, coredns) all `Running`
- [x] Each of the 3 commits builds individually (bisect-safe)
- [ ] Tested on other platforms (macOS / Linux) — **I could not test this; reviewers please confirm no regressions on hvf/kvm paths.**

## Related

- Closes #21529

### DIFF
--- a/pkg/drivers/qemu/qemu.go
+++ b/pkg/drivers/qemu/qemu.go
@@ -428,7 +428,6 @@ func (d *Driver) Start() error {
 	}
 
 	// hardware acceleration is important, it increases performance by 10x
-	accel = hardwareAcceleration()
 	if accel != "" {
 		klog.Infof("Using %s for hardware acceleration", accel)
 		startCmd = append(startCmd,
--- a/pkg/drivers/qemu/qemu.go
+++ b/pkg/drivers/qemu/qemu.go
@@ -564,6 +564,12 @@ func hardwareAcceleration() string {
 		// On Linux, enable the Kernel Virtual Machine accelerator.
 		return "kvm"
 	}
+	if runtime.GOOS == "windows" {
+		// On Windows, enable the Windows Hypervisor Platform accelerator (WHPX).
+		// This requires Windows 10/11 with the "Windows Hypervisor Platform" feature enabled.
+		// Users can enable it with: Enable-WindowsOptionalFeature -Online -FeatureName HypervisorPlatform
+		return "whpx"
+	}
 	return ""
 }
 
--- a/pkg/drivers/qemu/qemu.go
+++ b/pkg/drivers/qemu/qemu.go
@@ -247,6 +247,7 @@ func (d *Driver) Create() error {
 		if err != nil {
 			return err
 		}
+		log.Infof("Allocated host SSH port: %d (forwards to guest port 22)", d.SSHPort)
 
 		for {
 			d.EnginePort, err = getAvailableTCPPortFromRange(minPort, maxPort)
@@ -259,6 +260,7 @@ func (d *Driver) Create() error {
 			}
 			break
 		}
+		log.Infof("Allocated host Docker engine port: %d (forwards to guest port 2376)", d.EnginePort)
 	case "socket_vmnet":
 		d.SSHPort, err = d.GetSSHPort()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


medyagh:
@GiuseppeCapaldo93 plz sign the CLA

medyagh:
@bobsira would u plz try this PR and see if it works

medyagh (changes requested):
can you plz add Before/After this PR in the output ?

GiuseppeCapaldo93:
@medyagh thanks for the info, I was not aware of the need of the CLA, last time that I contributed was a GNU library it was not necessary, I think now it is done. let me know if there are problems

GiuseppeCapaldo93:
Thanks, CLA is fixed now. I added the requested Before/After section to the PR description. Could you please re-review when you have a chance? Also, could a Kubernetes member please run /ok-to-test so CI can start?

GiuseppeCapaldo93:
Hi @medyagh  — following up on this PR. I've addressed your feedback by adding the Before/After output section to the description (July 9), and the CLA issue is also resolved.

Since it's been a couple of months without further review, could you (or another maintainer) take another look when you have a chance? It would also help to get `/ok-to-test` so CI can run.

Thanks for your patience!

PR #22957, opened 2026-05-09
start: allow 1-CPU nodes when --no-kubernetes is used

### LINKED ISSUE #22152: minikube should allow 1 cpu --no-kubernetes
currently we dont allow it unless by --force

Pittu-Sharma:
#### **Summary**
This PR resolves a regression where `minikube` enforced a strict 2-CPU minimum requirement even when the `--no-kubernetes` flag was provided. 

The 2-CPU minimum is a requirement for a stable Kubernetes control plane; however, when a user opts to start minikube without bootstrapping Kubernetes (e.g., to run plain Docker/Containerd workloads), this restriction is unnecessary and prevents minikube from running on resource-constrained environments like single-CPU CI runners or small cloud instances.

#### **Changes**
- Modified `validateCPUCount` in `cmd/minikube/cmd/start.go` to bypass both the hardware availability check and the requested-CPU minimum check when `no-kubernetes` is active.
- Refactored `validateCPUCount` to use the `cobra.Command` object, allowing the system to distinguish between the default CPU count (2) and an explicit user request (e.g., `--cpus 1`).
- Ensures that minikube will automatically proceed with 1 CPU on 1-CPU systems in `--no-kubernetes` mode if no specific CPU count was requested.

#### **Verification**
- **Unit Tests**: Added and ran a targeted test suite verifying that the validation is correctly skipped only when `--no-kubernetes` is set.
- **Build**: Verified that the project builds successfully (`go build ./cmd/minikube`).
- **Dry-run**: Verified flag parsing and initialization logic with `minikube start --dry-run`.

**Fixes #22152**

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1230,7 +1230,7 @@ func validateRequestedMemorySize(req int, drvName string) {
 }
 
 // validateCPUCount validates the cpu count matches the minimum recommended & not exceeding the available cpu count
-func validateCPUCount(drvName string) {
+func validateCPUCount(cmd *cobra.Command, drvName string) {
 	var availableCPUs int
 
 	cpuCount := getCPUCount(drvName)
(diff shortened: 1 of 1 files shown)


nirs:
> #### **Summary**
> This PR resolves a regression where `minikube` enforced a strict 2-CPU minimum requirement even when the `--no-kubernetes` flag was provided.
> 
> The 2-CPU minimum is a requirement for a stable Kubernetes control plane; however, when a user opts to start minikube without bootstrapping Kubernetes (e.g., to run plain Docker/Containerd workloads), this restriction is unnecessary and prevents minikube from running on resource-constrained environments like single-CPU CI runners or small cloud instances.
> 
> #### **Changes**
> * Modified `validateCPUCount` in `cmd/minikube/cmd/start.go` to bypass both the hardware availability check and the requested-CPU minimum check when `no-kubernetes` is active.

Looks right

> * Refactored `validateCPUCount` to use the `cobra.Command` object, allowing the system to distinguish between the default CPU count (2) and an explicit user request (e.g., `--cpus 1`).

Not needed - the value can come from minikube config or environment variables. We don't care if --cpus was used for the change of skipping the 2 cpus limit when --no-kubernetes is used.

If this change is for improving the check check when not using --no-kubernetes than it does not belong to this PR. You can post another PR for this, but I'm not sure the flag check is correct.

> * Ensures that minikube will automatically proceed with 1 CPU on 1-CPU systems in `--no-kubernetes` mode if no specific CPU count was requested.

What is the existing behavior when the system has only one cpu and using --no-kubernetes?  Show an example run before an after this change.

PR #22959, opened 2026-05-09
fix: don't warn against virtualbox when minikube auto-selected it

### LINKED ISSUE #15456: minikube should not warn you to NOT use virtualbox if it auto-selected it

rafaelmfried:
Fixes #15456

## Problem

When minikube auto-selects the virtualbox driver (because no other driver was viable on the host), it still emits the steering-away-from-virtualbox warning:

```
* Automatically selected the virtualbox driver. Other choices: ssh, qemu2 (experimental)
...
╭───────────────────────────────────────────────────────────────────────╮
│ You have selected "virtualbox" driver, but there are better options ! │
│ For better performance and support consider using a different driver: │
│         - qemu2                                                       │
╰───────────────────────────────────────────────────────────────────────╯
```

This is contradictory: minikube has just told the user it picked virtualbox itself because the alternatives weren't viable, then immediately suggests using one of those alternatives.

## Fix

`selectDriver` already returns a `specified bool` indicating whether the user explicitly chose the driver (via `--driver`, `--vm-driver`, or an existing persisted profile). This patch threads that flag through `provisionWithDriver` into `node.Starter` (new `DriverSpecified` field) and gates the warning on it.

- User-specified virtualbox + warning enabled → warn (unchanged behavior)
- Auto-selected virtualbox → suppress warning (fixes the bug)
- Driver picked as a fallback after the primary driver failed → suppress (also not a user choice)
- Adding nodes to an existing cluster (`node start`) → suppress (user already saw or opted out at cluster creation)

## Tests

New table-driven unit test `TestShouldWarnVirtualBox` in `pkg/minikube/node/start_test.go` covers six combinations:

```
=== RUN   TestShouldWarnVirtualBox
--- PASS: TestShouldWarnVirtualBox/user-specified_virtualbox_with_warning_enabled
--- PASS: TestShouldWarnVirtualBox/auto-selected_virtualbox_(#15456)
--- PASS: TestShouldWarnVirtualBox/user-specified_virtualbox_with_warning_opted_out
--- PASS: TestShouldWarnVirtualBox/auto-selected_virtualbox_with_warning_opted_out
--- PASS: TestShouldWarnVirtualBox/non-virtualbox_driver_user-specified
--- PASS: TestShouldWarnVirtualBox/non-virtualbox_driver_auto-selected
PASS
```

The condition was extracted into a small pure function `shouldWarnVirtualBox(driverName, driverSpecified, wantWarning)` so the suppression logic is testable without spinning up a real cluster.

## Manual validation

| Scenario | Driver | Specified | Want warning | Before | After |
|---|---|---|---|---|---|
| User typed `--driver=virtualbox` | virtualbox | true | true | warn | warn |
| Auto-selected virtualbox (#15456) | virtualbox | false | true | **warn (bug)** | **silent (fix)** |
| User typed `--driver=virtualbox`, opted out | virtualbox | true | false | silent | silent |
| Non-virtualbox driver | docker | true | true | silent | silent |

End-to-end validation against a real virtualbox cluster on Linux is planned as a follow-up in the PR comments once a host is available; the local function-level test covers the regression boundary.

## Notes

- Two prior attempts (#15485 by @DiptoChakrabarty and #21088 by @vamshi1188) were closed by the lifecycle bot due to inactivity. This patch follows the same general direction but with a smaller, less invasive change: instead of persisting `DriverAutoSelected` on `ClusterConfig` (which would touch the on-disk profile schema), it stays on the in-memory `node.Starter` struct, which is the only consumer of the flag.
- /assign

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -231,7 +231,7 @@ func runStart(cmd *cobra.Command, _ []string) {
 
 	useForce := viper.GetBool(force)
 
-	starter, err := provisionWithDriver(cmd, ds, existing, options)
+	starter, err := provisionWithDriver(cmd, ds, existing, specified, options)
 	if err != nil {
 		node.ExitIfFatal(err, useForce)
 		machine.MaybeDisplayAdvice(err, ds.Name)
@@ -258,7 +258,9 @@ func runStart(cmd *cobra.Command, _ []string) {
 				if err != nil {
 					out.WarningT("Failed to delete cluster {{.name}}, proceeding with retry anyway.", out.V{"name": ClusterFlagValue()})
 				}
-				starter, err = provisionWithDriver(cmd, ds, existing, options)
+				// the alternate driver was picked by minikube as a fallback, not
+				// specified by the user
+				starter, err = provisionWithDriver(cmd, ds, existing, false, options)
 				if err != nil {
 					continue
 				}
@@ -305,7 +307,7 @@ func runStart(cmd *cobra.Command, _ []string) {
 	}
 }
 
-func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *config.ClusterConfig, options *run.CommandOptions) (node.Starter, error) {
+func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *config.ClusterConfig, driverSpecified bool, options *run.CommandOptions) (node.Starter, error) {
 	driverName := ds.Name
 	klog.Infof("selected driver: %s", driverName)
 	validateDriver(ds, existing)
@@ -401,14 +403,15 @@ func provisionWithDriver(cmd *cobra.Command, ds registry.DriverState, existing *
 	}
 
 	return node.Starter{
-		Runner:         mRunner,
-		PreExists:      preExists,
-		StopK8s:        stopk8s,
-		MachineAPI:     mAPI,
-		Host:           host,
-		ExistingAddons: existingAddons,
-		Cfg:            &cc,
-		Node:           &n,
+		Runner:          mRunner,
+		PreExists:       preExists,
+		StopK8s:         stopk8s,
+		MachineAPI:      mAPI,
+		Host:            host,
+		ExistingAddons:  existingAddons,
+		Cfg:             &cc,
+		Node:            &n,
+		DriverSpecified: driverSpecified,
 	}, nil
 }
 
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -224,7 +224,7 @@ func Start(starter Starter, options *run.CommandOptions) (*kubeconfig.Settings, 
 	// discourage use of the virtualbox driver, but only when the user actually
 	// asked for it — auto-selected virtualbox means no other driver was viable,
 	// so suggesting alternatives is misleading. See #15456.
-	if starter.Cfg.Driver == driver.VirtualBox && starter.DriverSpecified && viper.GetBool(config.WantVirtualBoxDriverWarning) {
+	if shouldWarnVirtualBox(starter.Cfg.Driver, starter.DriverSpecified, viper.GetBool(config.WantVirtualBoxDriverWarning)) {
 		warnVirtualBox(options)
 	}
 
@@ -998,6 +998,15 @@ func addCoreDNSEntry(runner command.Runner, name, ip string, cc config.ClusterCo
 	klog.Infof("{%q: %s} host record injected into CoreDNS's ConfigMap", name, ip)
 
 	return nil
+}
+
+// shouldWarnVirtualBox returns true when the steering-away-from-virtualbox
+// warning is appropriate: the active driver is virtualbox, the user opted
+// into the warning (WantVirtualBoxDriverWarning) AND they explicitly asked
+// for virtualbox. Auto-selection means minikube already considered (and
+// rejected) the alternatives, so there is no better option to point at.
+func shouldWarnVirtualBox(driverName string, driverSpecified, wantWarning bool) bool {
+	return driverName == driver.VirtualBox && driverSpecified && wantWarning
 }
 
 // prints a warning to the console against the use of the 'virtualbox' driver, if alternatives are available and healthy
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -83,14 +83,15 @@ var (
 
 // Starter is a struct with all the necessary information to start a node
 type Starter struct {
-	Runner         command.Runner
-	PreExists      bool
-	StopK8s        bool
-	MachineAPI     libmachine.API
-	Host           *host.Host
-	Cfg            *config.ClusterConfig
-	Node           *config.Node
-	ExistingAddons map[string]bool
+	Runner          command.Runner
+	PreExists       bool
+	StopK8s         bool
+	MachineAPI      libmachine.API
+	Host            *host.Host
+	Cfg             *config.ClusterConfig
+	Node            *config.Node
+	ExistingAddons  map[string]bool
+	DriverSpecified bool
 }
 
 // Start spins up a guest and starts the Kubernetes node.
@@ -220,8 +221,10 @@ func Start(starter Starter, options *run.CommandOptions) (*kubeconfig.Settings, 
 		go addons.Enable(&wg, starter.Cfg, list, enabledAddons, options)
 	}
 
-	// discourage use of the virtualbox driver
-	if starter.Cfg.Driver == driver.VirtualBox && viper.GetBool(config.WantVirtualBoxDriverWarning) {
+	// discourage use of the virtualbox driver, but only when the user actually
+	// asked for it — auto-selected virtualbox means no other driver was viable,
+	// so suggesting alternatives is misleading. See #15456.
+	if starter.Cfg.Driver == driver.VirtualBox && starter.DriverSpecified && viper.GetBool(config.WantVirtualBoxDriverWarning) {
 		warnVirtualBox(options)
 	}
 
--- a/pkg/minikube/node/start_test.go
+++ b/pkg/minikube/node/start_test.go
@@ -1,0 +1,86 @@
+/*
+Copyright 2026 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package node
+
+import (
+	"testing"
+
+	"k8s.io/minikube/pkg/minikube/driver"
+)
+
+func TestShouldWarnVirtualBox(t *testing.T) {
+	tests := []struct {
+		name            string
+		driverName      string
+		driverSpecified bool
+		wantWarning     bool
+		want            bool
+	}{
+		{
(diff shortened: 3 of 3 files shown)


rafaelmfried:
## Local validation

Reporting test results from my local macOS Sequoia (Apple Silicon) box.

### Targeted unit test

\`\`\`
$ go test ./pkg/minikube/node/ -run TestShouldWarnVirtualBox -v

=== RUN   TestShouldWarnVirtualBox
=== RUN   TestShouldWarnVirtualBox/✅_user-specified_virtualbox_with_warning_enabled_→_warn
=== RUN   TestShouldWarnVirtualBox/❌_auto-selected_virtualbox_→_suppress_(#15456)
=== RUN   TestShouldWarnVirtualBox/❌_user-specified_virtualbox_with_warning_opted_out_→_suppress
=== RUN   TestShouldWarnVirtualBox/❌_auto-selected_virtualbox_with_warning_opted_out_→_suppress
=== RUN   TestShouldWarnVirtualBox/❌_non-virtualbox_driver,_user-specified_→_no_warning
=== RUN   TestShouldWarnVirtualBox/❌_non-virtualbox_driver,_auto-selected_→_no_warning
--- PASS: TestShouldWarnVirtualBox (0.00s)
    --- PASS: TestShouldWarnVirtualBox/[all 6 subtests]
PASS
ok  	k8s.io/minikube/pkg/minikube/node
\`\`\`

The bug-fixing case is the second one (auto-selected virtualbox → suppress).

### Full unit test suite

Ran \`MINIKUBE_LDFLAGS="" TESTSUITE=unittest ./test.sh\` covering \`./cmd/...\` and \`./pkg/...\` (~80 packages). All packages \`ok\`, no \`FAIL\`. \`pkg/minikube/node\` shown below for emphasis:

\`\`\`
ok  	k8s.io/minikube/pkg/minikube/node    2.761s   coverage: 0.1% of statements
\`\`\`

(Coverage is low for this package overall because it has no other tests yet — this PR adds the first one.)

### Note on local lint

Running \`make test\` (full suite including \`make lint-ci\`) on master without my changes already fails with a pre-existing govet warning in \`pkg/drivers/parallels/parallels_darwin.go:425\` (printf %d with a \`*go-version.Version\` arg). Not introduced by this PR; happy to file a separate cleanup PR if useful.

End-to-end validation against a real virtualbox cluster on Linux is planned as a follow-up; the function-level test above covers the regression boundary.

nirs:
> Fixes #15456
> 
> ## Problem
> When minikube auto-selects the virtualbox driver (because no other driver was viable on the host), it still emits the steering-away-from-virtualbox warning:
> 
> ```
> * Automatically selected the virtualbox driver. Other choices: ssh, qemu2 (experimental)
> ...
> ╭───────────────────────────────────────────────────────────────────────╮
> │ You have selected "virtualbox" driver, but there are better options ! │
> │ For better performance and support consider using a different driver: │
> │         - qemu2                                                       │
> ╰───────────────────────────────────────────────────────────────────────╯
> ```
> 
> This is contradictory: minikube has just told the user it picked virtualbox itself because the alternatives weren't viable, then immediately suggests using one of those alternatives.

The issue is the text, not the warning. The purpose of the warning is to recommend a better driver when a user selects virtaulbox, since the driver has a lower priority.

Lets start with an issue about this, and discuss what it the wanted behavior there.

To it looks like "You have selected "virtualbox" driver, but there are better options !" is the problem. The user did not select the driver. It can make sense to show this warning:

```
╭───────────────────────────────────────────────────────────────────────╮
│ For better performance and support consider using a different driver: │
│         - qemu2                                                       │
╰───────────────────────────────────────────────────────────────────────╯
```

cmd/minikube/cmd/start.go
```diff
@@ -231,7 +231,7 @@ func runStart(cmd *cobra.Command, _ []string) {
 
 	useForce := viper.GetBool(force)
 
-	starter, err := provisionWithDriver(cmd, ds, existing, options)
+	starter, err := provisionWithDriver(cmd, ds, existing, specified, options)
```
nirs, line 234: This makes provision code more complicate, passing more options down. We could use options for this, but I think we to move the warning earlier - before we select the driver, or better into start_flags.go. The warning about selecting a driver belong to argument validation phase, which must be done before we actually start the node.

pkg/minikube/node/start.go
```diff
@@ -220,8 +221,10 @@ func Start(starter Starter, options *run.CommandOptions) (*kubeconfig.Settings,
 		go addons.Enable(&wg, starter.Cfg, list, enabledAddons, options)
 	}
 
-	// discourage use of the virtualbox driver
-	if starter.Cfg.Driver == driver.VirtualBox && viper.GetBool(config.WantVirtualBoxDriverWarning) {
+	// discourage use of the virtualbox driver, but only when the user actually
+	// asked for it — auto-selected virtualbox means no other driver was viable,
+	// so suggesting alternatives is misleading. See #15456.
+	if starter.Cfg.Driver == driver.VirtualBox && starter.DriverSpecified && viper.GetBool(config.WantVirtualBoxDriverWarning) {
 		warnVirtualBox(options)
```
nirs, line 228: Move this warning to start_flags. This is not the place to warn about the driver.

nirs (commented):
The table showing when the warning is wanted looks good, but the real fix is to move the warning to argument validation phase. This will automatically fix the issue by handling only the case of `--driver virtualbox` - before minikube selects the driver.

rafaelmfried:
Thanks for the careful review @nirs — the argument-validation framing makes sense, and you're right that pushing `DriverSpecified` through `provisionWithDriver`/`Starter` adds plumbing that wouldn't be needed if the warning lived in `start_flags.go`.

A few quick notes before I pivot, since the rework is non-trivial and I'd like to make sure we're aiming at the same target:

1. **Equivalence question.** Moving the warning to argument validation (gated on `viper.GetString("driver") == "virtualbox"` or `"vm-driver"`) covers the same set of cases the current patch covers, with two small differences:
   - **Existing profile reuse.** With the current patch, a profile that was originally created with vbox still triggers the warning on every subsequent `minikube start` (because `selectDriver` returns `specified=true` for the existing-profile path). Moving to `start_flags.go` means the warning would only fire when the user passes `--driver` *on this specific invocation* — quieter, arguably better.
   - **Output position.** The warning would print before the `* Automatically selected ...` / `* Using the X driver based on existing profile` line instead of after node provisioning. I think that's actually an improvement.
2. **Wording.** Independent of where the check lives, your suggestion to drop `You have selected \"virtualbox\" driver, but there are better options !` and just keep the `For better performance and support consider using a different driver: …` block is cleaner. Happy to fold that in too.
3. **@medyagh** — since you opened the issue, before I rewrite this could you confirm the desired behavior? Two options on the table:
   - **A (this PR's current shape):** suppress the warning when minikube auto-selects vbox; keep it when the user explicitly picks vbox or reuses an existing vbox profile.
   - **B (@nirs's proposal):** move the warning to argument validation in `start_flags.go`, fire only when `--driver=virtualbox` is on the current invocation; reword the message.

If B is the direction you want, I'll close this and open a fresh PR against `start_flags.go`. If A is closer to what you had in mind in the original issue, I'll address @nirs's concerns about the plumbing in a different way (e.g. plumbing the flag through a smaller surface).

Either way I'd rather not guess and end up with another PR rotting in `lifecycle/rotten` — so a quick steer from you would help a lot.

rafaelmfried:
While we wait for direction, I went ahead and prototyped @nirs's suggestion as a separate draft PR for side-by-side comparison: #22960. Same end-user behavior on the bug case in this issue, but with the warning living in argument-validation and a reworded message (drops the contradictory "You have selected …" opener).

Either PR is mergeable; the other will be closed once we know which shape is preferred. No need to review #22960 in depth yet — it's there so the choice is concrete rather than hypothetical.

nirs:
> Thanks for the careful review @nirs — the argument-validation framing makes sense, and you're right that pushing `DriverSpecified` through `provisionWithDriver`/`Starter` adds plumbing that wouldn't be needed if the warning lived in `start_flags.go`.
> 
> A few quick notes before I pivot, since the rework is non-trivial and I'd like to make sure we're aiming at the same target:
> 
> 1. **Equivalence question.** Moving the warning to argument validation (gated on `viper.GetString("driver") == "virtualbox"` or `"vm-driver"`) covers the same set of cases the current patch covers, with two small differences:
>    
>    * **Existing profile reuse.** With the current patch, a profile that was originally created with vbox still triggers the warning on every subsequent `minikube start` (because `selectDriver` returns `specified=true` for the existing-profile path). Moving to `start_flags.go` means the warning would only fire when the user passes `--driver` _on this specific invocation_ — quieter, arguably better.

The flag `--driver virtualbox` is ignored when starting existing cluster, and user has no reason to specify the flag in this case, but if we validate the flag very early we will warn again about the driver. So this check should be right after we checked if we have an existing cluster, and we should not warn when starting an existing cluster.

Unfortunately we don't have `create` command. We have `start`, `stop`, and `delete`. We should really have `create`, `start`, `stop`, and `delete`. And the warning about driver is relevant only to the `create` command.

>    * **Output position.** The warning would print before the `* Automatically selected ...` / `* Using the X driver based on existing profile` line instead of after node provisioning. I think that's actually an improvement.

I agree, this is the right place to warn about the user select argument. Also selecting a driver when we have an existing cluster does not make sense. If the cluster exists (stopped or started) there is nothing to select.

> 2. **Wording.** Independent of where the check lives, your suggestion to drop `You have selected \"virtualbox\" driver, but there are better options !` and just keep the `For better performance and support consider using a different driver: …` block is cleaner. Happy to fold that in too.
> 3. **@medyagh** — since you opened the issue, before I rewrite this could you confirm the desired behavior? Two options on the table:
>    
>    * **A (this PR's current shape):** suppress the warning when minikube auto-selects vbox; keep it when the user explicitly picks vbox or reuses an existing vbox profile.
>    * **B (@nirs's proposal):** move the warning to argument validation in `start_flags.go`, fire only when `--driver=virtualbox` is on the current invocation; reword the message.
> 
> If B is the direction you want, I'll close this and open a fresh PR against `start_flags.go`. If A is closer to what you had in mind in the original issue, I'll address @nirs's concerns about the plumbing in a different way (e.g. plumbing the flag through a smaller surface).
> 
> Either way I'd rather not guess and end up with another PR rotting in `lifecycle/rotten` — so a quick steer from you would help a lot.

I suggested to remove the text about select augment to warn about better driver when the user did not select the driver. But since we the warning we have now is only for user selection, we can consider this change later.

nirs:
Already fixed by #22960

Issue #22919, opened 2026-04-30
flake in smoke test docker arm64 in github action: "The driver 'docker' is not supported on linux/arm64"


medyagh:
```
I0430 18:52:19.678950    2825 out.go:360] Setting OutFile to fd 1 ...
I0430 18:52:19.679052    2825 out.go:408] TERM=,COLORTERM=, which probably does not support color
I0430 18:52:19.679058    2825 out.go:374] Setting ErrFile to fd 2...
I0430 18:52:19.679062    2825 out.go:408] TERM=,COLORTERM=, which probably does not support color
I0430 18:52:19.679189    2825 root.go:338] Updating PATH: /home/runner/.minikube/bin
W0430 18:52:19.679308    2825 root.go:314] Error reading config file at /home/runner/.minikube/config/config.json: open /home/runner/.minikube/config/config.json: no such file or directory
I0430 18:52:19.679731    2825 out.go:368] Setting JSON to false
I0430 18:52:19.680677    2825 start.go:134] hostinfo: {"hostname":"runnervm6gd1v","uptime":147,"bootTime":1777574992,"procs":163,"os":"linux","platform":"ubuntu","platformFamily":"debian","platformVersion":"24.04","kernelVersion":"6.14.0-1017-azure","kernelArch":"aarch64","virtualizationSystem":"","virtualizationRole":"","hostId":"3503f128-d5b0-4f65-b473-0081260290f4"}
I0430 18:52:19.680726    2825 start.go:144] virtualization:  
I0430 18:52:19.693365    2825 out.go:179] * minikube v1.38.1 on Ubuntu 24.04 (arm64)
* minikube v1.38.1 on Ubuntu 24.04 (arm64)
W0430 18:52:19.701178    2825 preload.go:371] Failed to list preload files: open /home/runner/.minikube/cache/preloaded-tarball: no such file or directory
I0430 18:52:19.701214    2825 notify.go:220] Checking for updates...
I0430 18:52:19.701282    2825 start.go:1919] No Kubernetes flag is set, setting Kubernetes version to v0.0.0
I0430 18:52:19.701336    2825 driver.go:422] Setting default libvirt URI to qemu:///system
I0430 18:52:24.961816    2825 docker.go:125] docker version: linux-28.0.4:Docker Engine - Community
I0430 18:52:24.961949    2825 cli_runner.go:164] Run: docker system info --format "{{json .}}"
[...]

Issue #22913, opened 2026-04-29
investigate test failures for kubernetes v1.36.0


medyagh:
logs for the failed tests
- https://storage.googleapis.com/minikube-builds/logs/22883/44555/Docker_Linux.html
- https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/22883/pull-minikube-docker-docker-linux-x86/2047800736380620800

medyagh:
@abedegno here is another failure tests that might interest you

rafaelmfried:
I went through one of the failures in this bucket — `TestAddons/parallel/Registry`
on `pull-minikube-docker-docker-linux-x86` — and wanted to share what I found, in
case it helps narrow it down.

The docker-docker run linked in the issue (PR #22883) fails on
`TestAddons/parallel/Registry`, and the same failure is currently deterministic
on the presubmits — on #22960 it failed on all 3 runs I looked at, every time at
the same step:

```
addons_test.go:405: expected curl response be "HTTP/1.1 200", but got *Connecting to registry.kube-system.svc.cluster.local (10.110.224.128:80)
```

The registry addon itself comes up fine — the `registry` and `registry-proxy`
pods both reach `Running` (`healthy within ~5s` each). The test then runs an
in-cluster probe (`wget --spider -S http://registry.kube-system.svc.cluster.local`),
and that's where it fails: `wget` resolves the Service DNS and gets as far as
`Connecting to ...:80`, but no HTTP response comes back — no `HTTP/1.1 ...` line
— so the check for `HTTP/1.1 200` fails.

Worth flagging that this is a different failure from #22608 ("Registry addon
doesn't work"): there the `kube-registry-proxy` image pull fails (`manifest
unknown`) and the addon never enables. Here the addon enables and both pods are
`Running` — the failure is purely at the in-cluster HTTP probe.

I haven't been able to pin the root cause from the Prow build logs alone — the
`registry` pod's container logs, the Service endpoints and the full `wget`
exchange aren't in the build log, so confirming why the Service returns nothing
would need reproducing the addon locally. Sharing this as triage, not a
diagnosis — just wanted to isolate this one cleanly from the rest of the
failures in that run. Happy to dig further if it helps.

rafaelmfried:
[...]

Issue #21087, opened 2025-07-17
VM driver wont work if user path is too long


medyagh:
when tryin to run qemu in github action it is not happy that 
'/Users/runner/work/minikube/minikube/minikube_binaries/testhome/.minikube/machines/functional-405000/monitor' is too long

we should Detect this and if this happening the LEAST we could do is warn ppl so it doesnt blow up in an ugly way
and most we could do is actually find a way to fix that 

```
=== RUN   TestFunctional/serial/StartWithProxy
    functional_test.go:2251: (dbg) Run:  ./minikube-darwin-amd64 start -p functional-405000 --memory=4096 --apiserver-port=8441 --wait=all --vm-driver=qemu --wait-timeout=15m --network socket_vmnet
    functional_test.go:2251: (dbg) Non-zero exit: ./minikube-darwin-amd64 start -p functional-405000 --memory=4096 --apiserver-port=8441 --wait=all --vm-driver=qemu --wait-timeout=15m --network socket_vmnet: exit status 80 (23.564294248s)
        
        -- stdout --
        	* [functional-405000] minikube v1.36.0 on Darwin 13.7.6
        	  - MINIKUBE_HOME=/Users/runner/work/minikube/minikube/minikube_binaries/testhome
        	  - KUBECONFIG=/Users/runner/work/minikube/minikube/minikube_binaries/testhome/kubeconfig
        	* Using the qemu2 driver based on user configuration
        	* Downloading VM boot image ...
        	* Starting "functional-405000" primary control-plane node in "functional-405000" cluster
        	* Downloading Kubernetes v1.33.2 preload ...
        	* Creating qemu2 VM (CPUs=2, Memory=4096MB, Disk=20000MB) ...
        	OUTPUT: 
        	ERROR: qemu-system-x86_64: -qmp unix:/Users/runner/work/minikube/minikube/minikube_binaries/testhome/.minikube/machines/functional-405000/monitor,server,nowait: UNIX socket path '/Users/runner/work/minikube/minikube/minikube_binaries/testhome/.minikube/machines/functional-405000/monitor' is too long
        	Path must be less than 104 bytes
        	
        	* Deleting "functional-405000" in qemu2 ...
        	* Creating qemu2 VM (CPUs=2, Memory=4096MB, Disk=20000MB) ...
        	OUTPUT: 
[...]

PR #22960, opened 2026-05-09
fix: warn against virtualbox only when the flag is explicit

### LINKED ISSUE #15456: minikube should not warn you to NOT use virtualbox if it auto-selected it

rafaelmfried:
**Fixes #15456** — moves the "don't use VirtualBox if better options exist" warning out of the post-provisioning path (`node.Start`) into `minikube start` argument-validation, so it only fires when the user actually selected VirtualBox, not when minikube auto-selected it.

Implements the structure @nirs suggested in #22959 (companion PR, to be closed in favor of this one): no `DriverSpecified` flag threaded through `Starter` — the warning lives in `cmd/minikube/cmd/start_flags.go::warnVirtualBoxDriver` and consults `viper.GetString("driver")` directly. Skipped when reusing an existing cluster (`--driver` is ignored on restart).

## Behavior matrix

| Invocation | Before any fix | This PR |
|---|---|---|
| `minikube start` (auto-pick lands on virtualbox) | warns (the bug) | silent ✅ |
| `minikube start --driver=virtualbox` (alternatives healthy) | warns | warns |
| `minikube start --vm-driver=virtualbox` (legacy) | warns | warns |
| `--driver=docker`, docker fails, fallback to virtualbox | warns | silent (flag is "docker") |
| reusing an existing virtualbox profile | warns | silent (existing cluster) |
| `--driver=virtualbox` reusing an existing profile | warns | silent (`--driver` ignored on restart) |
| opted out via `WantVirtualBoxDriverWarning=false` | silent | silent |
| `--driver=docker` | silent | silent |

## Tests

Per @nirs's review, the original table-driven unit test was removed: it monkey-patched global viper/stdout state, which is fragile and unsafe under `t.Parallel()`. The remaining logic is a handful of trivial short-circuit guards; the alternative-driver discovery branch (`driver.Choices` over a populated registry) is covered by the integration/e2e path, where a unit test would need a populated registry to add little signal.

Single DCO-signed commit.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -209,6 +209,11 @@ func runStart(cmd *cobra.Command, _ []string) {
 	}
 
 	validateSpecifiedDriver(existing, options)
+	if existing == nil {
+		// driver-selection warnings are only meaningful for fresh clusters;
+		// an existing cluster keeps its original driver
+		warnVirtualBoxDriver(options)
+	}
 	validateKubernetesVersion(existing)
 	validateContainerRuntime(existing)
 
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -44,6 +44,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/out"
 	"k8s.io/minikube/pkg/minikube/proxy"
 	"k8s.io/minikube/pkg/minikube/reason"
+	"k8s.io/minikube/pkg/minikube/registry"
 	"k8s.io/minikube/pkg/minikube/run"
 	"k8s.io/minikube/pkg/minikube/style"
 	pkgutil "k8s.io/minikube/pkg/util"
@@ -1068,4 +1069,48 @@ func checkExtraDiskOptions(cmd *cobra.Command, driverName string) {
 			out.WarningT("Specifying extra disks is currently only supported for the following drivers: {{.supported_drivers}}. If you can contribute to add this feature, please create a PR.", out.V{"supported_drivers": supportedDrivers})
 		}
 	}
+}
+
+// warnVirtualBoxDriver warns the user when the configured driver is
+// virtualbox (--driver/--vm-driver flag or MINIKUBE_DRIVER env) and better
+// alternatives are installed and healthy. This intentionally lives in the argument-validation
+// phase so the warning never fires when minikube auto-selects virtualbox
+// (auto-selection already implies the alternatives were rejected). Callers
+// must skip it when reusing an existing cluster, which keeps its original
+// driver.
+func warnVirtualBoxDriver(options *run.CommandOptions) {
+	if !viper.GetBool(config.WantVirtualBoxDriverWarning) {
+		return
+	}
+
+	requested := viper.GetString("driver")
+	if requested == "" {
+		requested = viper.GetString("vm-driver")
+	}
+	if !driver.IsVirtualBox(requested) {
+		return
+	}
+
+	var altDriverList strings.Builder
+	for _, choice := range driver.Choices(true, options) {
+		if !driver.IsVirtualBox(choice.Name) && choice.Priority != registry.Discouraged && choice.State.Installed && choice.State.Healthy {
+			fmt.Fprintf(&altDriverList, "\n\t- %s", choice.Name)
+		}
+	}
+	if altDriverList.Len() == 0 {
+		return
+	}
+
+	out.Boxed(`You have selected "virtualbox" driver, but there are better options !
+For better performance and support consider using a different driver: {{.drivers}}
+
+To turn off this warning run:
+
+	$ minikube config set WantVirtualBoxDriverWarning false
+
+
+To learn more about on minikube drivers checkout https://minikube.sigs.k8s.io/docs/drivers/
+To see benchmarks checkout https://minikube.sigs.k8s.io/docs/benchmarks/cpuusage/
+
+`, out.V{"drivers": altDriverList.String()})
 }
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -64,7 +64,6 @@ import (
 	"k8s.io/minikube/pkg/minikube/out/register"
 	"k8s.io/minikube/pkg/minikube/proxy"
 	"k8s.io/minikube/pkg/minikube/reason"
-	"k8s.io/minikube/pkg/minikube/registry"
 	"k8s.io/minikube/pkg/minikube/run"
 	"k8s.io/minikube/pkg/minikube/style"
 	"k8s.io/minikube/pkg/minikube/vmpath"
@@ -218,11 +217,6 @@ func Start(starter Starter, options *run.CommandOptions) (*kubeconfig.Settings, 
 		list := addons.ToEnable(starter.Cfg, starter.ExistingAddons, addonList)
 		wg.Add(1)
 		go addons.Enable(&wg, starter.Cfg, list, enabledAddons, options)
-	}
-
-	// discourage use of the virtualbox driver
-	if starter.Cfg.Driver == driver.VirtualBox && viper.GetBool(config.WantVirtualBoxDriverWarning) {
-		warnVirtualBox(options)
 	}
 
 	// special ops for "none" driver on control-plane node, like change minikube directory
@@ -995,29 +989,4 @@ func addCoreDNSEntry(runner command.Runner, name, ip string, cc config.ClusterCo
 	klog.Infof("{%q: %s} host record injected into CoreDNS's ConfigMap", name, ip)
 
 	return nil
-}
-
-// prints a warning to the console against the use of the 'virtualbox' driver, if alternatives are available and healthy
-func warnVirtualBox(options *run.CommandOptions) {
-	var altDriverList strings.Builder
-	for _, choice := range driver.Choices(true, options) {
-		if !driver.IsVirtualBox(choice.Name) && choice.Priority != registry.Discouraged && choice.State.Installed && choice.State.Healthy {
-			fmt.Fprintf(&altDriverList, "\n\t- %s", choice.Name)
-		}
-	}
-
-	if altDriverList.Len() != 0 {
-		out.Boxed(`You have selected "virtualbox" driver, but there are better options !
-For better performance and support consider using a different driver: {{.drivers}}
-
-To turn off this warning run:
-
-	$ minikube config set WantVirtualBoxDriverWarning false
-
-
-To learn more about on minikube drivers checkout https://minikube.sigs.k8s.io/docs/drivers/
-To see benchmarks checkout https://minikube.sigs.k8s.io/docs/benchmarks/cpuusage/
-
-`, out.V{"drivers": altDriverList.String()})
-	}
 }


rafaelmfried:
Pushed two adjustments per @nirs's last comment on #22959:

1. **Skip when reusing an existing cluster** (`existing != nil` short-circuit). `--driver` is ignored on cluster restart, so warning every time would be noise. Added a regression test.
2. **Restored the original \"You have selected virtualbox driver, but there are better options !\" opener.** Since the warning now only fires on explicit user selection, the original wording is appropriate again. The broader rewording (for the auto-selected path) is no longer needed because that path is silent — out of scope for this PR.

Updated the description and behavior matrix to reflect both changes. New test case `reusing_existing_cluster_—_silent_even_if_--driver=virtualbox` covers the new branch.

rafaelmfried:
@nirs since you wrote the original review, would you mind taking a quick look at this version? It implements the structure you suggested (warning in argument-validation phase) and the two follow-up adjustments from your last comment on #22959 are already in:

- `existing != nil` short-circuit so the warning doesn't fire when reusing a cluster
- original "You have selected virtualbox..." wording kept intact, since the warning now only triggers on explicit user selection

If this matches what you had in mind, a `/lgtm` or any further nits would help move this forward — currently waiting on `/ok-to-test` from a member.

cmd/minikube/cmd/start_flags_test.go
```diff
@@ -0,0 +1,120 @@
...
+			wantOutput: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			viper.Reset()
+			viper.Set("driver", tt.driverFlag)
+			viper.Set("vm-driver", tt.vmDriver)
+			viper.Set(config.WantVirtualBoxDriverWarning, tt.wantOptIn)
```
nirs: This changes a global - will fail if we run tests in parallel. If we we want to test this, we need to pass the function the user selection so we don't need to change viper.

cmd/minikube/cmd/start_flags_test.go
```diff
@@ -0,0 +1,120 @@
...
+			os.Stdout = w
+			os.Stderr = w
+			out.SetOutFile(w)
+			out.SetErrFile(w)
+			defer func() {
+				os.Stdout = oldOut
+				os.Stderr = oldErr
+				out.SetOutFile(oldOut)
+				out.SetErrFile(oldErr)
+			}()
```
nirs: These are globals that we should not depend on for the test. This again will break if we run tests in parallel.

cmd/minikube/cmd/start_flags_test.go
```diff
@@ -0,0 +1,120 @@
...
+			validateVirtualBoxDriverFlag(&run.CommandOptions{})
+			w.Close()
+			got := <-done
+			if tt.wantOutput && !strings.Contains(got, "consider using a different driver") {
+				t.Errorf("expected warning, got none. output=%q", got)
+			}
+			if !tt.wantOutput && strings.Contains(got, "consider using a different driver") {
+				t.Errorf("did not expect warning, got it. output=%q", got)
+			}
+		})
```
nirs: I think testing this code is not worth the effort. If we want to test the warning create a pure function that does not depend on any global and return a boolean. Then we can test all combinations without having to monkeypatch global state.

Or, modify the code so it does not depend on globals using dependency injection. If options and output comes from arguments to the function showing the warning, we can test it without mocking global state.

For this fix, I don't think we need any tests. The value added by the tests is very low and the using global state makes the test harmful by breaking parallel tests. The actual logic is trivial.

cmd/minikube/cmd/start_flags_test.go
```diff
@@ -106,7 +115,11 @@ func TestValidateVirtualBoxDriverFlag(t *testing.T) {
 				done <- string(b)
 			}()
 
-			validateVirtualBoxDriverFlag(&run.CommandOptions{})
+			var existing *config.ClusterConfig
+			if tt.hasExisting {
+				existing = &config.ClusterConfig{Name: "existing", Driver: "virtualbox"}
+			}
```
nirs: Tests should be trivial - the existing driver should be part of the table. We don't want extra setup code in the test.

cmd/minikube/cmd/start.go
```diff
@@ -209,7 +209,7 @@ func runStart(cmd *cobra.Command, _ []string) {
 	}
 
 	validateSpecifiedDriver(existing, options)
-	validateVirtualBoxDriverFlag(options)
+	validateVirtualBoxDriverFlag(existing, options)
```
nirs: We can pass existing to validateVirtualBoxDriverFlag, but since this validation is not relevant for existing driver it will be better to not call it:

```go
if !existing {
    validateVirtualBoxDriverFlag(options)
}
```

This makes it hard to test, but this test does not add much value.

nirs (commented):
The fix looks good, handling all cases correctly. The tests are too complicated trying to test global state. We don't want such tests.

If you remove the test and squash the fix commits we can merge this now. If we find a better way to test this avoiding global state we can add tests later.

rafaelmfried:
Done — addressed all the feedback in a single squashed commit (c068f9507):

- Moved the `existing == nil` guard to the call site in `start.go` per your suggestion, so `validateVirtualBoxDriverFlag` no longer needs to know about cluster state.
- Removed `start_flags_test.go` entirely. You're right that mocking viper and `os.Stdout` makes the test fragile and hostile to parallelism — not worth shipping for logic this small.
- Squashed the three fix commits into one.

Happy to follow up in a separate PR with a small refactor that extracts the decision into a stateless helper (something like `shouldWarnVirtualBoxDriver(driverFlag, vmDriver string, optedIn bool) bool`) so future regressions can be caught with a trivial table-driven test — no globals, parallel-safe. Let me know if that would be welcome and I'll open it after this lands.

rafaelmfried:
@nirs — for context, all your feedback from 2026-05-09 is in the squashed commit `c068f9507`:

- `validateVirtualBoxDriverFlag` no longer touches cluster state (guard moved to the call site in `start.go`)
- `start_flags_test.go` removed entirely per the global-state concern

CI is held by `needs-ok-to-test` since I'm not an org member. Would you mind `/ok-to-test` + `/lgtm` when you have a minute?

Separately: I'm ready to open the #21087 PR (the cross-driver socket path refactor we scoped together) — holding it back so this one can land first.

nirs:
@rafaelmfried thanks! Can you add example runs output showing the main use cases?

nirs:
@rafaelmfried I cleaned up the title to make it better for release notes.

rafaelmfried:
Thanks @nirs — appreciate the title cleanup and the `/ok-to-test`.

Force-pushed `6dbb8db9e` with a small `gofmt` fix on `pkg/minikube/node/start.go` (a stray trailing blank line was tripping the Lint job). Same single squashed commit, same content otherwise.

Example runs as requested — three scenarios on macOS 15.5 (arm64), built from this branch:

**1. `--driver=virtualbox` explicit → warning shown**



**2. No `--driver` flag → driver auto-selected, no warning**



`vfkit` auto-selected; nothing about virtualbox shown.

**3. Restart of an existing profile → warning suppressed by `existing != nil` guard**



(The profile in this print actually fell back to `vfkit` on first start because no real VirtualBox runtime exists on Apple silicon — but the code path that short-circuits when `existing != nil` is the one being exercised here, and it correctly suppresses the warning regardless of the underlying driver choice).

The other failing jobs in the previous run are flakes hitting unrelated PRs:

- `pull-minikube-none-docker-linux-x86` — failing on PRs #22974, #22978, #22970, #22973
- `pull-minikube-docker-crio-linux-x86` — same flake pattern
- `qemu-docker-macos-15-x86` — SSH dial timeout on the 2nd boot (`dial tcp 192.168.105.2:22: connect: host is down`), known qemu/networking flake on the GHA macOS runner

`/retest` once the new pipeline finishes.

rafaelmfried:
Per the bot's flake-tracking request — the failing jobs map to known open issues:

- `pull-minikube-docker-docker-linux-x86`, `docker-containerd-linux-x86`, `docker-crio-linux-x86` → #22913
- `pull-minikube-docker-docker-linux-arm` → #22919
- `pull-minikube-none-docker-linux-x86` → recurring on every recent PR (#22974, #22978, #22970) without a dedicated tracker that I could find

rafaelmfried:
`/retest` would just re-trigger the same broken jobs, so holding off until master clears.

rafaelmfried:
Following up on the red presubmits here — I re-checked them after the `/retest`,
and I should correct my earlier comments in this thread: they aren't flakes, and
they won't clear on their own. They are four separate, pre-existing CI
breakages, none of them caused by this PR (which only changes the
virtualbox-warning validation):

- `pull-minikube-none-docker-linux-x86` — `hack/prow/integration_none_docker_linux_x86.sh`
  pins a `cri-dockerd` version that does not exist; the install 404s. Filed as
  #22991, fix in #22992.
- `pull-minikube-docker-containerd-linux-x86` — `TestPreload/Restart-With-Preload-Check-User-Image`;
  already tracked by #22269.
- `pull-minikube-docker-crio-linux-x86` — `runc list` fails under crio-in-KIC
  (`MK_ADDON_*_PAUSED`); already tracked by #22326.
- `pull-minikube-docker-docker-linux-x86` — `TestAddons/parallel/Registry`, the
  registry Service probe gets no HTTP response; covered by #22913.

`build` and the KVM jobs are green here, and none of the four red jobs touches
the code this PR changes. Happy to `/retest` again whenever it is useful.

rafaelmfried:
Hi @nirs — small nudge on this one. Everything from your review is addressed:

- global-state test removed, fix commits squashed (`c068f9507`)
- the three example runs you asked for, added on 2026-05-13

It's `BLOCKED` purely on labels now — `ok-to-test` is set, just no `lgtm`/`approved`. A `/lgtm` whenever you have a minute would help.

The red presubmits are four pre-existing CI breakages unrelated to this PR (it only touches the virtualbox-warning validation) — I mapped each to its tracker in my 2026-05-17 comment above. `none-docker` specifically is fixed by #22992. Happy to follow whatever path you prefer to clear those required contexts.

cmd/minikube/cmd/start_flags.go
```diff
@@ -1069,3 +1070,47 @@ func checkExtraDiskOptions(cmd *cobra.Command, driverName string) {
...
+// in that case.
+func validateVirtualBoxDriverFlag(options *run.CommandOptions) {
+	if !viper.GetBool(config.WantVirtualBoxDriverWarning) {
+		return
+	}
+
+	requested := viper.GetString("driver")
+	if requested == "" {
+		requested = viper.GetString("vm-driver")
+	}
```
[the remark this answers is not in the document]
rafaelmfried: Good catch. `start_flags.go` sets `viper.SetEnvPrefix` + `viper.AutomaticEnv()`, so `viper.GetString("driver")` also resolves `MINIKUBE_DRIVER` (not just the flag). Reworded the doc comment to say the trigger is the *configured* driver (`--driver`/`--vm-driver` flag or `MINIKUBE_DRIVER` env). Behavior stays the same — an env/config-set driver is still a deliberate choice, unlike auto-selection. Fixed in d5bce92.

cmd/minikube/cmd/start_flags.go
```diff
@@ -1069,3 +1070,47 @@ func checkExtraDiskOptions(cmd *cobra.Command, driverName string) {
...
+	if !driver.IsVirtualBox(requested) {
+		return
+	}
```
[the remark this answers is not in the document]
rafaelmfried: The missing test is intentional, not an oversight. An earlier revision had `TestValidateVirtualBoxDriverFlag`, but @nirs asked to drop it during review: it monkey-patched global viper/stdout state (fragile, unsafe under `t.Parallel()`) and the remaining logic is a few trivial short-circuit guards. The PR description still referenced that test — that was stale, now corrected. If a maintainer would prefer unit coverage, the clean way (per @nirs's own suggestion) would be to refactor the helper to take its inputs as arguments instead of reading globals, then table-test it — happy to do that.

cmd/minikube/cmd/start_flags.go
```diff
@@ -1069,3 +1070,47 @@ func checkExtraDiskOptions(cmd *cobra.Command, driverName string) {
...
+
+	var altDriverList strings.Builder
+	for _, choice := range driver.Choices(true, options) {
+		if !driver.IsVirtualBox(choice.Name) && choice.Priority != registry.Discouraged && choice.State.Installed && choice.State.Healthy {
+			altDriverList.WriteString(fmt.Sprintf("\n\t- %s", choice.Name))
```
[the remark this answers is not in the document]
rafaelmfried: Done — switched to `fmt.Fprintf(&altDriverList, ...)`, which also lines up with the original `node.Start` code this logic was moved from. Fixed in d5bce92.

cmd/minikube/cmd/start.go
```diff
@@ -209,6 +209,11 @@ func runStart(cmd *cobra.Command, _ []string) {
 	}
 
 	validateSpecifiedDriver(existing, options)
+	if existing == nil {
+		// driver-selection warnings are only meaningful for fresh clusters;
+		// when reusing an existing cluster --driver is ignored
```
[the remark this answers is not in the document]
rafaelmfried: Reworded to "an existing cluster keeps its original driver" — accurate regardless of the enforcement path. You're right that `--driver` isn't silently ignored (`validateSpecifiedDriver` runs first); I went with the user-visible outcome rather than restating the internals. Same wording applied to the helper's doc comment. Fixed in d5bce92.

cmd/minikube/cmd/start_flags.go
```diff
@@ -1072,3 +1070,47 @@ func checkExtraDiskOptions(cmd *cobra.Command, driverName string) {
...
 }
+
+// validateVirtualBoxDriverFlag warns the user when the configured driver is
+// virtualbox (--driver/--vm-driver flag or MINIKUBE_DRIVER env) and better
+// alternatives are installed and healthy. This intentionally lives in the argument-validation
+// phase so the warning never fires when minikube auto-selects virtualbox
+// (auto-selection already implies the alternatives were rejected). Callers
+// must skip it when reusing an existing cluster, which keeps its original
+// driver.
+func validateVirtualBoxDriverFlag(options *run.CommandOptions) {
```
nirs: This uses viper.GetString which reads the value from minikube config, environment variables and flags. So the name validateVirtualBoxDriverFlag is too specific. We can use warnVirtualBoxDriver() which does not imply using flags, and make it clear that the purpose of this function is to warn about virtualbox driver. This also matches the config option to skip the warning (WantVirtualBoxDriverWarning)

nirs (commented):
@rafaelmfried Thanks, looks good, only one minor naming issue remains.

cmd/minikube/cmd/start_flags.go
rafaelmfried: Renamed to `warnVirtualBoxDriver` in 24e45c1 — agreed, since it reads from `viper` (config/env/flags) the "Flag" suffix was misleading, and `warn…` lines up with the `WantVirtualBoxDriverWarning` config option. Thanks for the catch.

nirs (approved):

PR #22964, opened 2026-05-11
start: add --mdns flag to enable mDNS on every VM start

### LINKED ISSUE #22804: Enable mDNS in cluster
When using mDNS vms on the host can be accessed using the vm hostname:

```console
% ssh krunkit.local -- hostname
krunkit
```

But inside minikube cluster the name cannot be resolved:

```console
% minikube ssh -- resolvectl query krunkit.local
krunkit.local: resolve call failed: All attempts to contact name servers or networks failed
```

Because mDNS is disabled by default:

```console
Link 2 (eth0)
    Current Scopes: DNS LLMNR/IPv4
         Protocols: +DefaultRoute +LLMNR -mDNS DNSOverTLS=opportunistic DNSSEC=no/unsupported
Current DNS Server: 192.168.105.1
       DNS Servers: 192.168.105.1
```

If we enable mDNS, resolving works:

```console
% minikube ssh -- sudo resolvectl mdns eth0 yes 

% minikube ssh -- resolvectl query krunkit.local
krunkit.local: 192.168.105.5                   -- link: eth0

-- Information acquired via protocol mDNS/IPv4 in 2.1ms.
-- Data is authenticated: no; Data was acquired via local or encrypted transport: no
-- Data from: network
```

## Possible fix

- enable mDNS by default in minikube iso
- add --enble-mdns start flag and config

### Expansion of start flags and node configuration

The current proliferation of flags in minikube start is a result of the tool's role as an all-in-one appliance managing the guest operating system.

- **Limitations of extra-config**: The --extra-config flag is technically scoped to kubernetes components like the kubelet or apiserver. Extending it to cover arbitrary node-level systemd settings would require a significant refactor of the provisioning logic.
- **Benefits of a structured file**: A dedicated --dns-config or --node-config yaml file provides a declarative way to manage nested settings that are difficult to express as flat key-value pairs in a cli flag.

### Defaulting mDNS in minikube

Enabling mdns by default involves a trade-off between user convenience and network hygiene.

- **Pros**: It enables seamless discovery of .local addresses for users on macos and linux without manual configuration. This is particularly useful for tools like vfkit that use virtual bridges.
- **Cons**: Multicast traffic can be problematic in restricted corporate environments or public networks. It also creates a dependency on the host having a functional mdns responder listening on the virtual network interface.

### Historical reasons for current defaults

Minikube defaults are generally optimized for a predictable, minimal environment that mirrors cloud behavior.

- **Cloud consistency**: Standard cloud environments do not use mdns for discovery, relying instead on unicast dns and service meshes.
- **Driver compatibility**: Multicast support varies across drivers. While it works with vfkit, it can be inconsistent on older nat implementations, leading to a fragmented user experience.
- **Resource constraints**: The buildroot-based iso aims for the smallest possible footprint. Disabling non-essential protocols by default reduces background noise and potential attack vectors.

### Comparison of mDNS defaults in Linux distributions

The default state of mdns in 2026 depends heavily on the distribution's philosophy regarding systemd-resolved.

- **Fedora 43**: Typically enables mdns by default as part of its standard systemd-resolved integration. [Fedora Changes: systemd-resolved](https://fedoraproject.org/wiki/Changes/systemd-resolved)
- **Debian 13 (trixie)**: Often disables mdns in systemd-resolved by default to avoid conflicts with the avahi-daemon, which remains the preferred provider for the project. [Debian Release Notes](https://www.debian.org/releases/trixie/releasenotes)
- **General systemd-resolved behavior**: The upstream default for systemd-resolved is often opportunistic, but distributions frequently apply their own overrides. [systemd-resolved documentation](https://www.freedesktop.org/software/systemd/man/latest/resolved.conf.html)

carterpewpew:
Add an opt-in `--mdns` start flag and `minikube config set mdns true` option that enables mDNS (.local address resolution) inside the cluster node by running `resolvectl mdns eth0 yes` on every start.

The resolvectl call lives in `configureMDNS()` in `pkg/minikube/node/dns.go` and is invoked from `startMachine` (not provisioning) so it persists across stop/start cycles, since resolvectl is a runtime setting that does not survive a VM reboot.

- Flag is create-time only: baked into cluster config at first start, no need to repeat on subsequent starts
- Flag defaults to `false`, no behavior change for existing users
- Non-VM drivers (docker, podman) silently skipped via `driver.IsVM()` guard
- Failures logged as user-visible warnings via `out.WarningT`, start continues

Fixes #22804

### DIFF
--- a/cmd/minikube/cmd/config/config.go
+++ b/cmd/minikube/cmd/config/config.go
@@ -177,6 +177,10 @@ var settings = []Setting{
 		set:         SetStringSlice,
 		validations: []setFn{IsValidIPAddressSlice},
 	},
+	{
+		name: config.MDNS,
+		set:  SetBool,
+	},
 }
 
 // ConfigCmd represents the config command
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -152,6 +152,7 @@ const (
 	rosetta                 = "rosetta"
 	vmnetOffloading         = "vmnet-offloading"
 	dnsServers              = config.DNSServers
+	mdns                    = config.MDNS
 )
 
 var (
@@ -318,6 +319,7 @@ func initNetworkingFlags() {
 
 	// dns
 	startCmd.Flags().StringSlice(dnsServers, nil, "Static DNS server IP addresses for the VM (VM drivers only)")
+	startCmd.Flags().Bool(mdns, false, "Enable mDNS (.local address resolution) by configuring systemd-resolved inside the node (VM drivers only)")
 
 	// socket vmnet
 	startCmd.Flags().String(socketVMnetClientPath, "", "Path to the socket vmnet client binary (QEMU driver only)")
@@ -636,6 +638,23 @@ func getDNSServers(cmd *cobra.Command, driverName string) []netip.Addr {
 	return addrs
 }
 
+// getMDNS returns true if mDNS should be enabled for the given driver.
+// For non-VM drivers the value is ignored since mDNS configuration via
+// systemd-resolved is not applicable.
+func getMDNS(cmd *cobra.Command, driverName string) bool {
+	enabled := viper.GetBool(mdns)
+	if !enabled {
+		return false
+	}
+	if !driver.IsVM(driverName) || driver.IsSSH(driverName) {
+		if cmd.Flags().Changed(mdns) {
+			out.WarningT("--mdns flag is only valid with VM drivers, it will be ignored")
+		}
+		return false
+	}
+	return enabled
+}
+
 // generateNewConfigFromFlags generate a config.ClusterConfig based on flags
 func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime string, drvName string, options *run.CommandOptions) config.ClusterConfig {
 	var cc config.ClusterConfig
@@ -743,6 +762,7 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 		Rosetta:            getRosetta(drvName),
 		VmnetOffloading:    getVmnetOffloading(drvName),
 		DNSServers:         getDNSServers(cmd, drvName),
+		MDNS:               getMDNS(cmd, drvName),
 	}
 	cc.VerifyComponents = interpretWaitFlag(*cmd)
 
--- a/pkg/minikube/node/dns.go
+++ b/pkg/minikube/node/dns.go
@@ -27,6 +27,10 @@ import (
 	"k8s.io/minikube/pkg/minikube/out"
 )
 
+// primaryInterface is the main network interface in the buildroot-based minikube ISO.
+// Update this if the ISO switches to a different distro or naming scheme.
+const primaryInterface = "eth0"
+
 // configureDNS configures static DNS servers on the VM, overriding DHCP-provided
 // DNS settings. This fixes DNS resolution on managed Macs where network extensions
 // block DNS traffic from the VM bridge to the host resolver.
@@ -54,10 +58,10 @@ func configureDNS(runner command.Runner, servers []netip.Addr) {
 	dnsServers := strings.Join(values, " ")
 
 	script := fmt.Sprintf(`
-resolvectl dns eth0 %s
-resolvectl domain eth0 "~."
+resolvectl dns %s %s
+resolvectl domain %s "~."
 resolvectl flush-caches
-`, dnsServers)
+`, primaryInterface, dnsServers, primaryInterface)
 
 	cmd := exec.Command("sudo", "bash", "-o", "errexit", "-c", script)
 	if _, err := runner.RunCmd(cmd); err != nil {
@@ -67,4 +71,20 @@ resolvectl flush-caches
 	}
 
 	klog.Infof("Configured static DNS servers: %s", dnsServers)
+}
+
+// configureMDNS enables mDNS (.local address resolution) on the VM by configuring
+// systemd-resolved. This allows the guest to resolve other machines on the local
+// network that advertise via mDNS.
+func configureMDNS(runner command.Runner, enabled bool) {
+	if !enabled {
+		return
+	}
+	cmd := exec.Command("sudo", "resolvectl", "mdns", primaryInterface, "yes")
+	if _, err := runner.RunCmd(cmd); err != nil {
+		klog.Warningf("Failed to enable mDNS on %s: %v", primaryInterface, err)
+		out.WarningT("Failed to enable mDNS on {{.iface}}", out.V{"iface": primaryInterface})
+		return
+	}
+	klog.Infof("Enabled mDNS on %s", primaryInterface)
 }
--- a/pkg/minikube/node/start.go
+++ b/pkg/minikube/node/start.go
@@ -671,6 +671,7 @@ func startMachine(cfg *config.ClusterConfig, node *config.Node, delOnFail bool, 
 	}
 
 	configureDNS(runner, cfg.DNSServers)
+	configureMDNS(runner, cfg.MDNS)
 
 	ip, err := validateNetwork(hostInfo, runner, cfg.KubernetesConfig.ImageRepository)
 	if err != nil {
--- a/pkg/minikube/config/config.go
+++ b/pkg/minikube/config/config.go
@@ -61,6 +61,8 @@ const (
 	MaxAuditEntries = "MaxAuditEntries"
 	// DNSServers is the key for static DNS server addresses for VM drivers
 	DNSServers = "dns-servers"
+	// MDNS is the key for the mDNS parameter (boolean)
+	MDNS = "mdns"
 )
 
 var (
--- a/pkg/minikube/config/types.go
+++ b/pkg/minikube/config/types.go
@@ -113,6 +113,7 @@ type ClusterConfig struct {
 	Rosetta                 bool          // Only used by vfkit driver
 	VmnetOffloading         bool          // Only used by krunkit driver
 	DNSServers              []netip.Addr  // Static DNS servers for the VM (VM drivers only)
+	MDNS                    bool          // Enable mDNS (.local) resolution via systemd-resolved
 }
 
 // KubernetesConfig contains the parameters used to configure the VM Kubernetes.


nirs:
Thanks @carterpewpew! This is very exciting change that will make my life much nicer. I'll review this later this week.

pkg/provision/buildroot.go
```diff
@@ -192,5 +192,9 @@ func (p *BuildrootProvisioner) Provision(swarmOptions swarm.Options, authOptions
 		return err
 	}
 
+	if err := setMDNSOptions(p.clusterName, p); err != nil {
+		klog.Warningf("failed to configure mDNS during provisioning: %v", err)
```
nirs: Bug: `setMDNSOptions` is called here during provisioning, but `resolvectl mdns eth0 yes` is a runtime setting that does not survive a VM reboot. After `minikube stop && minikube start`, mDNS is no longer enabled.

This should be moved out of provisioning and into the start path (e.g. in `node.Start` in `pkg/minikube/node/start.go`, after `AddHostAlias` and before `startPrimaryControlPlane`) so it runs on every start.

Tested locally:
```
$ minikube start --driver vfkit --network vmnet-shared --mdns-enable  # works
$ minikube ssh -- ping foo.local  # resolves OK
$ minikube stop && minikube start
$ minikube ssh -- ping foo.local  # "bad address"
```

cmd/minikube/cmd/start_test.go
```diff
@@ -860,3 +860,20 @@ func TestValidateAutoPause(t *testing.T) {
 		}
 	}
 }
+
+func TestGenerateConfigIncludesMDNSEnable(t *testing.T) {
+	viper.SetDefault(humanReadableDiskSize, defaultDiskSize)
+	viper.SetDefault(kvmNUMACount, 1)
```
nirs: This test only verifies a single-line `viper.GetBool` → struct field assignment, which doesn't add meaningful coverage.

More importantly, it mutates global viper state (`viper.Set`/`viper.SetDefault`) which prevents parallel test execution. We should avoid adding more tests with this pattern — it makes the existing problem harder to fix.

Suggest removing this and instead adding an integration test that starts a cluster with the flag on a VM driver and verifies mDNS resolution actually works (including after a stop/start cycle). That would catch real regressions — like the persistence bug above — which this unit test cannot.

pkg/provision/buildroot.go
nirs: Bug: `setMDNSOptions` is called here during provisioning, but `resolvectl mdns eth0 yes` is a runtime setting that does not survive a VM reboot. After `minikube stop && minikube start`, mDNS is no longer enabled.

This should be moved out of provisioning and into the node.StartMachine so it runs on every start before we validate the network:
https://github.com/kubernetes/minikube/blob/cb53ec0dcd058cf9a617d403fa2e147ad4de5695/pkg/minikube/node/start.go#L681-L683

Tested locally:
```
$ minikube start --driver vfkit --network vmnet-shared --mdns-enable  # works
$ minikube ssh -- ping foo.local  # resolves OK
$ minikube stop && minikube start
$ minikube ssh -- ping foo.local  # "bad address"
```

cmd/minikube/cmd/start_test.go
nirs: This test only verifies a single-line `viper.GetBool` → struct field assignment, which doesn't add meaningful coverage.

More importantly, it mutates global viper state (`viper.Set`/`viper.SetDefault`) which prevents parallel test execution. We should avoid adding more tests with this pattern — it makes the existing problem harder to fix.

Suggest removing this and instead adding an integration test that starts a cluster with the flag on a VM driver and verifies mDNS resolution actually works (including after a stop/start cycle). That would catch real regressions — like the persistence bug above — which this unit test cannot.

cmd/minikube/cmd/start_flags.go
```diff
@@ -147,6 +147,7 @@ const (
 	autoPauseInterval       = "auto-pause-interval"
 	preloadSrc              = "preload-source"
 	rosetta                 = "rosetta"
+	mdnsEnable              = "mdns-enable"
```
nirs: Boolean flags in minikube are named as the feature itself: `--rosetta`, `--ha`, `--preload`, `--embed-certs`, `--delete-on-failure`. The `-enable` suffix is inconsistent.

Suggest renaming both the constant and the flag value: `mdnsEnable = "mdns-enable"` → `mdns = "mdns"`, to match the existing pattern where they always correspond (e.g. `rosetta = "rosetta"`, `ha = "ha"`). The struct field, config key, and other references should be updated accordingly.

I think this is my fault, specifying the flag name in the issue. The idea was that we may have more mDNS flags, so all can have --mdns prefix, but I think we can still do this if needed.

cmd/minikube/cmd/config/config.go
```diff
@@ -172,6 +172,11 @@ var settings = []Setting{
 		name: config.MaxAuditEntries,
 		set:  SetInt,
 	},
+	{
+		name:      "mdns-enable",
```
nirs: Should be config.MDNS like other options.

cmd/minikube/cmd/config/config.go
```diff
@@ -172,6 +172,11 @@ var settings = []Setting{
...
+		set:       SetBool,
+		callbacks: []setFn{RequiresRestartMsg},
```
nirs: Remove `RequiresRestartMsg`. `minikube config set` only sets default values for new clusters — it does not modify the config of an existing cluster. The existing cluster's config is baked in at creation time. A restart will not pick up the new value, so this message is misleading.

We want to keep `config set` simple: it sets defaults. To change an existing cluster, delete it and recreate. We don't have a `minikube edit` command and many settings are baked in and cannot be changed after creation.

nirs (commented):
Thanks for picking this up and putting together such a clean PR! The code is well-organized, the commit message is excellent, and the defensive error handling (warning instead of aborting) is right for an optional feature like this.

I have a few suggestions before we merge:

1. The mDNS setting is lost after `minikube stop && minikube start` — it needs to move from provisioning to the start path so it runs on every start (see inline comment with reproduction steps).
2. The unit test should be replaced with an integration test that verifies mDNS actually works, including after a stop/start cycle.
3. The flag should be `--mdns` to match the naming convention of other boolean flags.
4. Use a config constant instead of a string literal, and remove the `RequiresRestartMsg` callback.

Sorry for the long list — the PR is really close, these are just refinements to make it solid.

carterpewpew:
@nirs All four addressed: --mdns naming, moved to startMachine so it survives reboots, removed the misleading restart callback, and added an integration test that validates persistence across stop/start. PTAL, Thanks again for the thorough review and feedback.

test/integration/mdns_test.go
```diff
@@ -0,0 +1,75 @@
...
+			t.Fatalf("failed to stop minikube: %v\n%s", err, rr.Output())
+		}
+
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), startArgs...))
+		if err != nil {
+			t.Fatalf("failed to start minikube after stop: %v\n%s", err, rr.Output())
+		}
+
+		assertMDNSEnabled(ctx, t, profile)
+	})
```
nirs: Using sub tests for this is not useful. This is the way existing tests are written, but this is wrong.

We want to test this flow:
1. start cluster with --mdns
2. validate that mdns is configured
3. stop cluster
4. start cluster without arguments
5. validate that mdns is configured

There is no way to split this to sub tests - if any step fails, we cannot run the later steps - this means that all steps must be part of a single test.

Using sub tests makes sense only when each test is standalone and does not depend on any other sub test.

pkg/minikube/node/start.go
```diff
@@ -680,6 +680,12 @@ func startMachine(cfg *config.ClusterConfig, node *config.Node, delOnFail bool,
 		return runner, preExists, m, hostInfo, fmt.Errorf("Failed to get command runner: %w", err)
 	}
 
+	if cfg.MDNS && driver.IsVM(hostInfo.Driver.DriverName()) {
+		if _, err := runner.RunCmd(exec.Command("sudo", "resolvectl", "mdns", "eth0", "yes")); err != nil {
+			klog.Warningf("failed to enable mDNS on eth0: %v", err)
```
nirs: klog.Warning logs a warning to the log - we want to show warning in minikube output. Check how we issue warnings to the console. I think validateNetwork issues such warning.

test/integration/mdns_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"context"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+// TestMDNS verifies that --mdns enables mDNS on the node and persists across stop/start.
+func TestMDNS(t *testing.T) {
```
nirs: Adding separate test is the cleanest way o test this, but is very slow - this takes at least 35 seconds locally with vfkit, and can table 90 seconds with kvm on linux.

What we try to do usually is find another test that already start cluster twice, and change the test to use also this feature. I'm not sure that this is possible since we need to have one test doing start+stop+start, and most tests are split to sub tests and cannot be used. Please check if we have a test that can be used. If not adding this test is fine.

cmd/minikube/cmd/start_flags.go
```diff
@@ -213,6 +214,7 @@ func initMinikubeFlags() {
 	startCmd.Flags().StringP(gpus, "g", "", "Allow pods to use your GPUs. Options include: [all,nvidia,amd] (Docker driver with Docker container-runtime only)")
 	startCmd.Flags().Duration(autoPauseInterval, time.Minute*1, "Duration of inactivity before the minikube VM is paused (default 1m0s)")
 	startCmd.Flags().String(preloadSrc, "auto", "Which source to download the preload from (valid options: gcs, github, auto). Defaults to auto (try both).")
+	startCmd.Flags().Bool(mdns, false, "If set, enable mDNS (.local address resolution) by configuring systemd-resolved inside the node (VM drivers only).")
```
nirs: Here we use false as the default value, but the default value should come form the config, or we should use the config value later when we validate the option value.

Check how other config options are used - I think we have helper functions for application the config value. The best way to apply the config is to set the value at the point we initialize the flags, so the using --help we see the actual default value.

For example if you set the config to true:

    $ minikube config set mdns true

The minikube start --help will show:

    $ minikube start --help
    ...
    --mdns=true ...

nirs (commented):
@carterpewpew Thanks! please the see the comments about using the config value and the integration test.

test/integration/mdns_test.go
```diff
@@ -0,0 +1,75 @@
...
+	if !VMDriver() {
+		t.Skip("--mdns only applies to VM drivers")
+	}
+
+	profile := UniqueProfileName("mdns")
+	ctx, cancel := context.WithTimeout(context.Background(), Minutes(10))
```
nirs: Stating a cluster does not need 10 minutes. The cluster should start in 40-80 seconds, so we can use 5 minutes timeout. If we cannot start in 5 minutes the system is too slow for testing.

This context is used for first start, stop, and second start, so it enforces a deadline for the entire flow which is reasonable. But it is better to have a timeout for each step:
- first start: 5m
- stop: 1m
- second start: 5m

With this we can fail fast if something is broken, instead of waiting 10 minutes for stuck cluster.

test/integration/mdns_test.go
```diff
@@ -0,0 +1,75 @@
...
+	defer CleanupWithLogs(t, profile, cancel)
+
+	startArgs := append([]string{"start", "-p", profile, "--mdns", "--memory=2048"}, StartArgs()...)
```
nirs: --memory 2048 is too low. Minikube cannot star with less than 2400mb. Check how other tests specify memory size, we should have a standard way to do this. I think in most tests we use 3072mb as the default, we should have a standard way to add the memory so we don't need to duplicate the size everywhere.

You can in the build log that the test failed, and if you run this locally you will see the same failure.

Check the docs to learn how to run tests locally:
https://minikube.sigs.k8s.io/docs/contrib/testing/

nirs (commented):
Added more comments on the new test.

medyagh:
thanks for this PR @carterpewpew please rebase

medyagh:
lets have a tutorial page in our site for it and make sure we unrestand and tell users how it is different than
https://minikube.sigs.k8s.io/docs/handbook/addons/ingress-dns/#Mac

pkg/minikube/node/start.go
```diff
@@ -680,6 +680,12 @@ func startMachine(cfg *config.ClusterConfig, node *config.Node, delOnFail bool,
...
+		}
+	}
```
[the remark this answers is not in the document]
nirs: We use buildroot for minikube iso, so using eth0 is fine.

test/integration/mdns_test.go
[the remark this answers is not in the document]
nirs: The second start does not apply the flag since the cluster exists and the setting is baked into the cluster config. Start flags are ignored when using existing cluster. But regardless there is no reason to use the flag in the second start.

test/integration/mdns_test.go
```diff
@@ -0,0 +1,75 @@
...
+func assertMDNSEnabled(ctx context.Context, t *testing.T, profile string) {
+	t.Helper()
+	rr, err := Run(t, exec.CommandContext(ctx, Target(), "ssh", "-p", profile, "--", "sudo", "resolvectl", "mdns", "eth0"))
+	if err != nil {
+		if strings.Contains(rr.Output(), "not found") || strings.Contains(rr.Output(), "No such file") {
+			t.Skipf("resolvectl not available in this VM image, skipping: %s", rr.Output())
+		}
+		t.Fatalf("failed to check mDNS status: %v\n%s", err, rr.Output())
+	}
+	if !strings.Contains(rr.Output(), "yes") {
```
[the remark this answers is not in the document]
nirs: Good idea to check for more reliable output.

test/integration/mdns_test.go
carterpewpew: Addressed, switched to per phase contexts with independent deadlines and removed the shared context approach.

test/integration/mdns_test.go
carterpewpew: Sorry, I missed that earlier. Updated this test to use --memory=3072 to match the integration test baseline across the suite. Also ran locally and verified it passes with vfkit.

Screenshot 2026-05-19 at 08 59 49

test/integration/mdns_test.go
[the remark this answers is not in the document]
carterpewpew: Addressed. Removed --mdns from the second start and kept restart args minimal. The second start now relies on the persisted cluster config only.

test/integration/mdns_test.go
```diff
@@ -0,0 +1,75 @@
...
+		if err != nil {
+			t.Fatalf("failed to stop minikube: %v\n%s", err, rr.Output())
+		}
+
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), startArgs...))
+		if err != nil {
+			t.Fatalf("failed to start minikube after stop: %v\n%s", err, rr.Output())
+		}
+
+		assertMDNSEnabled(ctx, t, profile)
```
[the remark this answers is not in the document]
carterpewpew: Addressed by moving to separate contexts per phase instead of reusing one shared context, so each step has an independent timeout and we fail fast on the specific stuck phase.

test/integration/mdns_test.go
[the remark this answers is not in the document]
carterpewpew: Addressed, switched the check to resolvectl status eth0 and assert on +mDNS, which matches actual output on the minikube ISO and is more deterministic than the previous loose yes match.

carterpewpew:
@nirs @medyagh Thanks for the detailed reviews and insights (again 😄). Addressed all points in the latest push and rebased onto latest master. PTAL

cmd/minikube/cmd/start_flags.go
```diff
@@ -214,6 +216,7 @@ func initMinikubeFlags() {
 	startCmd.Flags().StringP(gpus, "g", "", "Allow pods to use your GPUs. Options include: [all,nvidia,amd] (Docker driver with Docker container-runtime only)")
 	startCmd.Flags().Duration(autoPauseInterval, time.Minute*1, "Duration of inactivity before the minikube VM is paused (default 1m0s)")
 	startCmd.Flags().String(preloadSrc, "auto", "Which source to download the preload from (valid options: gcs, github, auto). Defaults to auto (try both).")
+	startCmd.Flags().Bool(mdns, mdnsDefault(), "If set, enable mDNS (.local address resolution) by configuring systemd-resolved inside the node (VM drivers only).")
```
nirs: This provides best user experience - we will see the right default in --help. But this is not consistent with other code, and it read the config twice.

Minikube configures viper to read minikube config, so we can use here viper.GetBool(mdns) to get the default.

The issue is that it does not work now, since minikube call viper.ReadConfig() too late - after the flags were created. This should be fixed to create read config before flags are created.

So we have 2 options - keep the "false" hardcoded value, and replace with viper.GetBool(mdns) when viper integration is fixed, or use now viper.GetBool(mdns) and fix the viper issue properly. I'm fine with both options.

cmd/minikube/cmd/start_flags.go
```diff
@@ -558,6 +561,15 @@ func validateVfkitNetwork(n string, options *run.CommandOptions) string {
 	return n
 }
 
+func mdnsDefault() bool {
```
nirs: Adding special helper to read the config for every flag does not scale and not efficient. If we do this for every flag we will read the config many times during start.

cmd/minikube/cmd/start_flags.go
```diff
@@ -923,6 +936,7 @@ func updateExistingConfigFromFlags(cmd *cobra.Command, existing *config.ClusterC
 	updateDurationFromFlag(cmd, &cc.AutoPauseInterval, autoPauseInterval)
 	updateBoolFromFlag(cmd, &cc.Rosetta, rosetta)
 	updateBoolFromFlag(cmd, &cc.VmnetOffloading, vmnetOffloading)
+	updateBoolFromFlag(cmd, &cc.MDNS, mdns)
```
nirs: This works and allow changing the flag when starting an existing machine:

```
minikube start                    # mdns disabled
minikube stop
minikube start --mdns             # mdns enabled
minikube stop
minikube start                    # mdns enabled
minikube stop
minikube start --mdns=false       # mdns disable
```

This is helps the user to fix mistakes 0 they forgot to enable mdns when the cluster was created.

But it is not fully baked - this will not work:

```
minikube start --mdns             # mdns enabled
minikube start --mdns=false       # mdns disabled in the running vm
minikube stop
minikube start                    # mdns disabled
```

I think the callback you add in the first version to warn that the user needs to restart is the way to handle this case, but this is a very poor user exprience.

The right way to support enable and disabling the flag dynamically is to disable mdns when on a running vm. But for this we need to keep the value in the config as a pointer so it can present:
- nil: mdns was not specified
- true: mdns enabled
- false: mdns disabled

But we don't have the infra to do this - the current code supports only bool values.

So we have 2 options:
1. Support partial working flag that does not work running cluster
   - Requires adding test coverage for dynamic changes - we need to cover disabling mdns on a cluster created with --mdns and enabling it back - that's 2 more start and stop test steps.
2. Disable support for dynamic modifications - make --mdns create time flag. You need to recreate the cluster if you want to change it
   - Current test are good enough of this.

I suggest we remove the update here to make the flag create time flag to keep the tests simpler. We can add dynamic enable/disable and support changing running vms later if users will request this feature. I think this is lower priority before we have a request, and we have much more important work.

BTW, rosetta and ventOffloading options are bugs - they do not work every for starting a stopped vm, since these options use ~/.minikube/machines/NAME/config.json and not ~/.minikube/profiles/NAME/config.json - which is the file modified in this function.

pkg/minikube/node/start.go
```diff
@@ -680,6 +680,12 @@ func startMachine(cfg *config.ClusterConfig, node *config.Node, delOnFail bool,
 		return runner, preExists, m, hostInfo, fmt.Errorf("Failed to get command runner: %w", err)
 	}
 
+	if cfg.MDNS && driver.IsVM(hostInfo.Driver.DriverName()) {
+		if _, err := runner.RunCmd(exec.Command("sudo", "resolvectl", "mdns", "eth0", "yes")); err != nil {
```
nirs: It will help to add a comment about eth0, or add a constant. AI agents complain that this depends the distro. This is right since we use buildroot. The comment can explain why and help to fix this is we switch to another distro later.

A more robust way would be to use a constant from the code assuming buildroot, so it is clear that code depend on the distro. I'm not sure we have a good place of this, so adding a constant here should be good enough for now.

test/integration/mdns_test.go
```diff
@@ -0,0 +1,82 @@
...
+	restartCtx, restartCancel := context.WithTimeout(context.Background(), Minutes(5))
+	defer restartCancel()
+	restartArgs := append([]string{"start", "-p", profile, "--memory=3072"}, StartArgs()...)
+	rr, err = Run(t, exec.CommandContext(restartCtx, Target(), restartArgs...))
+	if err != nil {
+		t.Fatalf("failed to restart minikube without --mdns: %v\n%s", err, rr.Output())
+	}
+
+	// 5. Validate mDNS is still configured
+	assertMDNSEnabled(restartCtx, t, profile)
```
nirs: The test looks correct now, we run the entire flow so we will not have false negative like the other tests using sub tests. We will also get good logs when the test fails since we see the entire flow in test. This is another issue with tests trying to use very small sub tests.

One issue, we don't don't test that starting without --mdns does not enable mdns - if the code is broken and we always enable it the tests will succeed.

If we add a test it may fail if systemd changes the default and enable mdns by default, but in this case we want to detect the issue quickly and simplify the code.

Adding a test for disable --mdns should be skipped if minikube is configured to enable mdns by default.

So we are missing another test that start minikube without the flag and test that mdns is not set. This can be a test that run in parallel with this test. We can create 2 sub tests with MaybeParallel() so they we wait only for the slower test.

This again can be adde to another test that start a cluster, but it does not make sense to split tests like this. We want to have clean tests that are easy to modify and debug, and handle performance issue by running them in parallel.

test/integration/mdns_test.go
```diff
@@ -0,0 +1,82 @@
...
+func TestMDNS(t *testing.T) {
+	if !VMDriver() {
+		t.Skip("--mdns only applies to VM drivers")
+	}
+
+	profile := UniqueProfileName("mdns")
+	defer CleanupWithLogs(t, profile, func() {})
+
+	// 1. Start with --mdns
+	startCtx, startCancel := context.WithTimeout(context.Background(), Minutes(5))
```
nirs: We need to use t.Context() and not context.Background(). This context is canceled when we reach the global test timeout.

test/integration/mdns_test.go
```diff
@@ -0,0 +1,82 @@
...
+	defer stopCancel()
+	rr, err = Run(t, exec.CommandContext(stopCtx, Target(), "stop", "-p", profile))
+	if err != nil {
+		t.Fatalf("failed to stop minikube: %v\n%s", err, rr.Output())
+	}
+
+	// 4. Start without --mdns (setting is persisted in cluster config)
+	restartCtx, restartCancel := context.WithTimeout(context.Background(), Minutes(5))
+	defer restartCancel()
+	restartArgs := append([]string{"start", "-p", profile, "--memory=3072"}, StartArgs()...)
```
nirs: This should run only "minkube start -p profile" no other argument is needed.

nirs (commented):
@carterpewpew Thanks for the updates!

The new tests works nicely and will be easy to debug.

```
% make integration TEST_ARGS="-minikube-start-args='--driver=vfkit --network=vmnet-shared' -test.run TestMDNS"     
go test -ldflags="-X k8s.io/minikube/pkg/version.version=v1.38.1 -X k8s.io/minikube/pkg/version.isoVersion="v1.38.0"-1778539887-22974 -X k8s.io/minikube/pkg/version.gitCommitID="746cbc6fadc5d03d9ebea9bb06a7cb21fc978d72" -X k8s.io/minikube/pkg/version.storageProvisionerVersion=v5" -v -test.timeout=90m ./test/integration --tags="integration libvirt_dlopen" -minikube-start-args='--driver=vfkit --network=vmnet-shared' -test.run TestMDNS 2>&1 | tee "./out/testout_746cbc6fa.txt"
Found 12 cores, limiting parallelism with --test.parallel=6
=== RUN   TestMDNS
    mdns_test.go:41: (dbg) Run:  out/minikube start -p mdns-727000 --mdns --memory=3072 --driver=vfkit --network=vmnet-shared
    mdns_test.go:41: (dbg) Done: out/minikube start -p mdns-727000 --mdns --memory=3072 --driver=vfkit --network=vmnet-shared: (16.311147667s)
    mdns_test.go:47: (dbg) Run:  out/minikube ssh -p mdns-727000 -- sudo resolvectl status eth0
    mdns_test.go:52: (dbg) Run:  out/minikube stop -p mdns-727000
    mdns_test.go:52: (dbg) Done: out/minikube stop -p mdns-727000: (12.203785333s)
    mdns_test.go:61: (dbg) Run:  out/minikube start -p mdns-727000 --memory=3072 --driver=vfkit --network=vmnet-shared
    mdns_test.go:61: (dbg) Done: out/minikube start -p mdns-727000 --memory=3072 --driver=vfkit --network=vmnet-shared: (9.453992375s)
    mdns_test.go:67: (dbg) Run:  out/minikube ssh -p mdns-727000 -- sudo resolvectl status eth0
    helpers_test.go:185: Cleaning up "mdns-727000" profile ...
    helpers_test.go:188: (dbg) Run:  out/minikube delete -p mdns-727000
--- PASS: TestMDNS (38.20s)
PASS
Tests completed in 38.200416542s (result code 0)
ok  	k8s.io/minikube/test/integration	39.219s
```

We still have some issues - minikube is old and complicated and not easy to work with.

test/integration/mdns_test.go
```diff
@@ -0,0 +1,82 @@
...
+		t.Skip("--mdns only applies to VM drivers")
+	}
+
+	profile := UniqueProfileName("mdns")
+	defer CleanupWithLogs(t, profile, func() {})
+
+	// 1. Start with --mdns
+	startCtx, startCancel := context.WithTimeout(context.Background(), Minutes(5))
+	defer startCancel()
+	startArgs := append([]string{"start", "-p", profile, "--mdns", "--memory=3072"}, StartArgs()...)
```
nirs: We can cover the config here by checking if minikube is configured to enable mdns by default - in this case we don't need to add "--mdns" to the command and the test should succeed.

nirs:
@carterpewpew Please check #23069 - very similar problem, it would be nice to have configureMDNS() in node/dns.go, and we can test both using the same dns_test.go.

Since writing integration tests is the hard part and we are close to the release, we can split the integration test to another pr and merge the fix now.

carterpewpew:
Thank you @nirs for the inputs and apologies for the delay. I will take a look at this this week.

carterpewpew:
> @carterpewpew Please check #23069 - very similar problem, it would be nice to have configureMDNS() in node/dns.go, and we can test both using the same dns_test.go.
> 
> Since writing integration tests is the hard part and we are close to the release, we can split the integration test to another pr and merge the fix now.

@nirs Sounds good, split the integration test out and kept this PR to just the core feature. Moved the resolvectl logic into pkg/minikube/node/dns.go as suggested. PTAL

cmd/minikube/cmd/start_flags.go
carterpewpew: Kept the hardcoded false default for now. Will swap to viper.GetBool(mdns) once the viper config read ordering is fixed upstream.

cmd/minikube/cmd/start_flags.go
carterpewpew: Agreed, removed the per-flag config reader. Using hardcoded false now.

cmd/minikube/cmd/start_flags.go
carterpewpew: Went with option 2, --mdns is create-time only. No update in updateExistingConfigFromFlags, users recreate to change it.

pkg/minikube/node/start.go
carterpewpew: Extracted eth0 into a primaryInterface constant with a comment noting it's buildroot-specific.

cmd/minikube/cmd/start_flags.go
```diff
@@ -694,6 +693,7 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 		AutoPauseInterval:  viper.GetDuration(autoPauseInterval),
 		Rosetta:            getRosetta(drvName),
 		VmnetOffloading:    getVmnetOffloading(drvName),
+		MDNS:               viper.GetBool(mdns),
```
nirs: Lets add getMDNS() helper like in #23069, so we filter non vm drivers (!driver.IsVM(), driver.IsSSH()) and warn that the flag will be ignored if the flag changed.
https://github.com/kubernetes/minikube/pull/23069/changes/a2544acd36644ce47ac1480452c44eae97a5729e

This will allow simplifying the code later, since cfg.MDNS will be set only if we need to use it, and user will get better feedback if they use an option that does not work with the current driver.

pkg/minikube/node/dns.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"os/exec"
+
+	"k8s.io/minikube/pkg/minikube/command"
+	"k8s.io/minikube/pkg/minikube/out"
+)
+
+// primaryInterface is the main network interface in the buildroot-based minikube ISO.
+// Update this if the ISO switches to a different distro or naming scheme.
+const primaryInterface = "eth0"
```
nirs, line 32: Good, I need to adopt this in https://github.com/kubernetes/minikube/pull/23069/changes/781cab17d259bf876e093f319bdde1c98b11dbb0#diff-be4fac8e21b0e76d789b6aad73dceccc3118e0cbc88c1345e8aab848722e5e14R56

Check how we use "eth0" in other code, maybe we can have this constant in a common location, maybe part of the buildroot provisioner.

pkg/minikube/node/dns.go
```diff
@@ -0,0 +1,37 @@
...
+
+// configureMDNS enables mDNS (.local address resolution) on the VM by configuring
+// systemd-resolved. This allows the guest to resolve other machines on the local
+// network that advertise via mDNS.
+func configureMDNS(runner command.Runner) {
+	if _, err := runner.RunCmd(exec.Command("sudo", "resolvectl", "mdns", primaryInterface, "yes")); err != nil {
+		out.WarningT("Failed to enable mDNS on {{.iface}}: {{.error}}", out.V{"iface": primaryInterface, "error": err})
+	}
```
nirs, line 88: Lets keep this style to make the code easier to follow and change.

```go
cmd := exec.Command("sudo", "resolvectl", "mdns", primaryInterface, "yes")
if _, err := runner.RunCmd(cmd); err != nil {
    out.WarningT("Failed to enable mDNS on {{.iface}}: {{.error}}", out.V{"iface": primaryInterface, "error": err})
}
```

This is easier to read and easier to modify, for example if we want to set cmd fields later.

pkg/minikube/node/start.go
```diff
@@ -676,6 +680,10 @@ func startMachine(cfg *config.ClusterConfig, node *config.Node, delOnFail bool,
 		return runner, preExists, m, hostInfo, fmt.Errorf("Failed to get command runner: %w", err)
 	}
 
+	if cfg.MDNS && driver.IsVM(hostInfo.Driver.DriverName()) {
+		configureMDNS(runner)
+	}
```
nirs: If we do the check early in start_flags.go we can use here:

```go
configureMDNS(runner, cfg.MDNS)
```

Same as #23069

pkg/minikube/node/dns.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// primaryInterface is the main network interface in the buildroot-based minikube ISO.
+// Update this if the ISO switches to a different distro or naming scheme.
+const primaryInterface = "eth0"
+
+// configureMDNS enables mDNS (.local address resolution) on the VM by configuring
+// systemd-resolved. This allows the guest to resolve other machines on the local
+// network that advertise via mDNS.
+func configureMDNS(runner command.Runner) {
```
nirs: Lets pass the cfg.MDNS value and skip if it is not enabled. Keeping the logic of when to configure here keeps the caller code clean and make future changes easier. For exmple if we move to three-sate (true, false, nil) we need to change only this function.

nirs (commented):
@carterpewpew thanks, deferring the test is the right way. Let's make this more like #23069.

Feel free to review #23069.

cmd/minikube/cmd/start_flags.go
carterpewpew: Done. getMDNS(cmd, driverName) with !driver.IsVM || driver.IsSSH guard, warns only if cmd.Flags().Changed(mdns)

pkg/minikube/node/dns.go
carterpewpew: Done. primaryInterface constant added, configureDNS updated to use it

pkg/minikube/node/dns.go
carterpewpew: Done. cmd := exec.Command(...) on its own line

pkg/minikube/node/start.go
carterpewpew: Done. No outer if/driver check in start.go

pkg/minikube/node/dns.go
carterpewpew: Done. configureMDNS(runner command.Runner, enabled bool) with if !enabled { return }

cmd/minikube/cmd/start_flags.go
```diff
@@ -636,6 +638,23 @@ func getDNSServers(cmd *cobra.Command, driverName string) []netip.Addr {
...
+	if !enabled {
+		return false
+	}
+	if !driver.IsVM(driverName) || driver.IsSSH(driverName) {
+		if cmd.Flags().Changed(mdns) {
+			out.WarningT("--mdns flag is only valid with VM drivers, it will be ignored")
+		}
+		return false
+	}
+	return enabled
```
nirs, line 655: This would be little more clear as:

```go
enabled := viper.GetBool(mdns)
if enabled && (!driver.IsVM(driverName) || driver.IsSSH(driverName)) {
    if cmd.Flags().Changed(mdns) {
        out.WarningT("--mdns flag is only valid with VM drivers, it will be ignored")
    }
    return false
}
return enabled
```

If we want to go with early return like the current code we don't need the enabled flag:

```go
if !viper.GetBool(mdns) {
    return false
}
if !driver.IsVM(driverName) || driver.IsSSH(driverName) {
    if cmd.Flags().Changed(mdns) {
    out.WarningT("--mdns flag is only valid with VM drivers, it will be ignored")
    }
    return false
 }
 return true
```

But It think the first option reveals the intent better.

pkg/minikube/node/dns.go
```diff
@@ -54,10 +58,10 @@ func configureDNS(runner command.Runner, servers []netip.Addr) {
 	dnsServers := strings.Join(values, " ")
 
 	script := fmt.Sprintf(`
-resolvectl dns eth0 %s
-resolvectl domain eth0 "~."
+resolvectl dns %s %s
+resolvectl domain %s "~."
 resolvectl flush-caches
-`, dnsServers)
+`, primaryInterface, dnsServers, primaryInterface)
```
nirs, line 64: Nice - but best done in a separate commit. Not critical but if we need to revert this change, we will lose the nice primaryInterface constant. The clean way to do this is 2 commits:
- add primaryInterface constants and use it in configureDNS
- add --mdns option and configureMDNS

nirs (commented):
@carterpewpew Thanks! Looks good, but see the comment for possible minor teaks.

nirs (approved):

PR #22971, opened 2026-05-11
Fix flaky node addition race condition when applying labels

medyagh:
currently 10% flake rate on KVM linux


jacisko:
Facing same issue in our environment.

PR #22983, opened 2026-05-15
kic: fix inspect to read only stdout, ignoring stderr warnings

### LINKED ISSUE #22537: Minikube fails to detect Podman container IP when BoltDB deprecation warning is emitted (Fedora 42, Podman 5.7.0, boltdb)
### Cosa è successo?

[log.txt](https://github.com/user-attachments/files/24841002/log.txt)

Minikube fails to start when using the Podman driver if Podman emits the BoltDB deprecation warning.  
The warning interferes with Minikube’s parsing of `podman inspect` output, causing Minikube to incorrectly fall back to the `lo` interface (127.0.0.1) and fail with:

unable to select an IP from lo network interface

This appears to be a parsing/robustness issue in Minikube when handling non-empty stderr output from Podman.

---

### Environment
- OS: Fedora 42
- Minikube: v1.36.0
- Driver: podman
- Podman: 5.7.0
- Podman database backend: BoltDB (legacy)
- Runtime: rootful (via sudo)

---

### Steps to Reproduce
1. Install Fedora 42 with Podman 5.7.0 still configured to use BoltDB.
2. Ensure the BoltDB deprecation warning is **not** suppressed.
3. Run:
   ```bash
   minikube start --driver=podman
   ```
   or
   ```bash
   minikube -p minikube-new start --driver=podman
   ```

### Observed Behavior

- Minikube startup fails with:

   ```log
  unable to select an IP from lo network interface
   ```
- Debug logs (--alsologtostderr --v=8) show Minikube retrieving network data using:

 ```bash
podman container inspect -f {{.NetworkSettings.IPAddress}} minikube
podman container inspect -f "{{range .NetworkSettings.Networks}}{{.IPAddress}},{{.GlobalIPv6Address}}{{end}}" minikube
 ```

- Podman emits the following warning on stderr:

 ```log
WARN[0000] The deprecated BoltDB database driver is in use. This driver will be removed in the upcoming Podman 6.0 release...
 ```
### Expected Behavior

Minikube should correctly detect the Podman container IP regardless of non-fatal warnings printed to stderr by Podman.

### Workarounds

Any of the following restore correct behavior:

1. Suppress the BoltDB warning for sudo commands and Wait for safe db migration

 ```
echo "SUPPRESS_BOLTDB_WARNING=true" >> /etc/environment
 ```

2. **DISRUPTIVE WAY** migrate BoltDB, but lost your current minikube profile 

### Allegare il file di log

[log.txt](https://github.com/user-attachments/files/24841004/log.txt)

### Sistema operativo

Redhat/Fedora

### Driver

Podman

carterpewpew:
**Problem:**

The `inspect()` function in `pkg/drivers/kic/oci/oci.go` merged stdout and stderr into a single buffer before passing the command to `runCmd()`. When Podman 5.7.0+ emits warnings to stderr (e.g. BoltDB deprecation: `WARN[0000] The deprecated BoltDB database driver is in use...`), those lines end up in the parsed output. 

**Fix:**

Remove the manual buffer that merged both streams. Let `runCmd()` handle stdout/stderr separation via its default path (which already writes them to separate `rr.Stdout` and `rr.Stderr` buffers), then scan only `rr.Stdout`.

**Testing:**

Added `TestInspectIgnoresStderr` which creates a fake OCI binary that emits to both stdout and stderr, confirming only stdout is parsed.

Fixes #22537

### DIFF
--- a/pkg/drivers/kic/oci/oci.go
+++ b/pkg/drivers/kic/oci/oci.go
@@ -409,11 +409,8 @@ func inspect(ociBin string, containerNameOrID, format string) ([]string, error) 
 	cmd := exec.Command(ociBin, "container", "inspect",
 		"-f", format,
 		containerNameOrID) // ... against the "node" container
-	var buff bytes.Buffer
-	cmd.Stdout = &buff
-	cmd.Stderr = &buff
-	_, err := runCmd(cmd)
-	scanner := bufio.NewScanner(&buff)
+	rr, err := runCmd(cmd)
+	scanner := bufio.NewScanner(&rr.Stdout)
 	var lines []string
 	for scanner.Scan() {
 		lines = append(lines, scanner.Text())
--- a/pkg/drivers/kic/oci/oci_test.go
+++ b/pkg/drivers/kic/oci/oci_test.go
@@ -18,6 +18,7 @@ package oci
 
 import (
 	"os"
+	"path/filepath"
 	"testing"
 )
 
@@ -110,6 +111,23 @@ func TestPointToHostPodman(t *testing.T) {
 		if v := os.Getenv(exp.key); v != exp.value {
 			t.Errorf("invalid %v env variable. got: %v, want: %v", exp.value, v, exp.value)
 		}
+	}
+}
+
+func TestInspectIgnoresStderr(t *testing.T) {
+	script := filepath.Join(t.TempDir(), "fake-oci")
+	content := "#!/bin/sh\necho '192.168.49.2'\necho 'WARN[0000] The BoltDB backend is deprecated' >&2\n"
+	if err := os.WriteFile(script, []byte(content), 0700); err != nil {
+		t.Fatal(err)
+	}
+
+	lines, err := inspect(script, "test-container", "{{.NetworkSettings.IPAddress}}")
+	if err != nil {
+		t.Fatalf("inspect failed: %v", err)
+	}
+
+	if len(lines) != 1 || lines[0] != "192.168.49.2" {
+		t.Errorf("inspect returned %v, want [\"192.168.49.2\"]", lines)
 	}
 }
 


pkg/drivers/kic/oci/oci.go
```diff
@@ -410,11 +410,8 @@ func inspect(ociBin string, containerNameOrID, format string) ([]string, error)
 	cmd := exec.Command(ociBin, "container", "inspect",
 		"-f", format,
 		containerNameOrID) // ... against the "node" container
-	var buff bytes.Buffer
-	cmd.Stdout = &buff
-	cmd.Stderr = &buff
-	_, err := runCmd(cmd)
-	scanner := bufio.NewScanner(&buff)
+	rr, err := runCmd(cmd)
+	scanner := bufio.NewScanner(&rr.Stdout)
```
Roslaan001, line 414: If `runCmd(cmd)` fails (e.g. container does not exist or daemon is unreachable), we shouldn't continue scanning `rr.Stdout` or return a partially populated `lines` slice alongside `err`.

Could we return early if `runCmd` fails?
```suggestion
	rr, err := runCmd(cmd)
	if err != nil {
		return nil, err
	}
	scanner := bufio.NewScanner(&rr.Stdout)
	var lines []string
	for scanner.Scan() {
		lines = append(lines, scanner.Text())
	}
	if err := scanner.Err(); err != nil {
		klog.Warningf("failed to read output: %v", err)
		return lines, err
	}
	return lines, nil

File: pkg/drivers/kic/oci/oci_test.go [hunk 2]
```diff
@@ -113,6 +114,37 @@ func TestPointToHostPodman(t *testing.T) {
 	}
 }
 
+// fakeOCIEnv makes the test binary act as a fake OCI binary, see TestMain.
+const fakeOCIEnv = "MINIKUBE_TEST_FAKE_OCI"
+
+func TestMain(m *testing.M) {
+	if os.Getenv(fakeOCIEnv) == "1" {
```
Roslaan001, line 121: In `TestMain`, the fake OCI currently returns success regardless of the arguments passed. To make sure `inspect()` is actually constructing the CLI invocation correctly (e.g. passing `container inspect`), could we add a basic check on `os.Args`?

```go
	if os.Getenv(fakeOCIEnv) == "1" {
		if len(os.Args) < 3 || os.Args[1] != "container" || os.Args[2] != "inspect" {
			fmt.Fprintf(os.Stderr, "unexpected args: %v\n", os.Args)
			os.Exit(1)
		}
		fmt.Println("192.168.49.2")
		fmt.Fprintln(os.Stderr, "WARN[0000] The BoltDB backend is deprecated")
		os.Exit(0)
	}

Roslaan001 (commented):
Thanks @carterpewpew!

Removing the merged buffer in `inspect()` directly fixes #22537 and aligns `inspect()` with the rest of the OCI runner package.

i left two suggestions inline:
1. Adding an early return `if err != nil { return nil, err }` in `inspect()` so we fail fast and don't scan stdout when the command itself returned an error.
2. Adding a quick `os.Args` validation in `TestMain` so the test verifies the arguments passed to the CLI.

nirs:
@carterpewpew The issue you link to was reported as fixed in minikube 1.38. Regardless the change makes sense.

Roslaan001:
> @carterpewpew The issue you link to was reported as fixed in minikube 1.38. Regardless the change makes sense.

Yeah, I also saw the issue was closed but since this PR doesn't get closed, I thought it's still needed

PR #23034, opened 2026-05-23
image load: Fix silent failure when loading an image used by a container

### LINKED ISSUE #23023: `image load` silently fails when the image is in use by a running container
## Problem

`minikube image load <image:tag>` silently fails to update an image when a container using that image is already running inside minikube.

This is a common scenario during development: rebuild a Docker image with the same tag (e.g., `myapp:0.1.0-SNAPSHOT`), then reload it into minikube while the previous deployment is still running. The command exits successfully with no error, but the image inside minikube is not updated.

Related to #11276 (originally fixed in #11366).

### Root cause

`transferAndLoadImage` calls `removeExistingImage` before loading the new image ([ref: v1.38.1 source][1]). `removeExistingImage` runs the runtime's remove command (e.g., `docker rmi <image:tag>`) inside minikube, which fails when a running container references the image:

```
conflict: unable to remove repository reference "myapp:0.1.0" (must force)
- container abc123 is using its referenced image def456
```

The error-handling in `removeExistingImage` uses ad-hoc string matching against a hard-coded list of "acceptable" error messages ([ref: v1.38.1 source][2]):

- `"no such image"`
- `"unable to remove the image"` (added in #19312)

Any error that does not match is treated as fatal. This includes Docker's `"unable to remove repository reference"` error shown above, as well as any new error messages that container runtimes may introduce in the future. When the error is treated as fatal, `LoadImage` is never called. Additionally, `DoLoadImages` swallows the error (returns nil), so the user sees no indication of failure.

Note: with older versions of `crictl rmi` (cri-tools ≤ v1.31), the error message for any removal failure was a catch-all `"unable to remove the image(s)"` ([ref: cri-tools v1.29.0 source][3]), which happened to match the
hard-coded pattern. Since cri-tools v1.32.0 ([ref: cri-tools PR #1639][4]), image removal was parallelized and the catch-all was replaced with per-image error messages, meaning the pattern match no longer works for newer CRI runtimes either. This further demonstrates that ad-hoc string matching is not a sustainable approach.

More fundamentally, removal of the old image cannot be a precondition for loading the new one. A running container holds a reference to the image it was started from, and that reference is not released until the container is stopped - not even after a new image is loaded with the same tag. This means a "remove-then-load" sequence will always fail when a container is running with the target image.

[1]: https://github.com/kubernetes/minikube/blob/v1.38.1/pkg/minikube/machine/cache_images.go#L280-L288
[2]: https://github.com/kubernetes/minikube/blob/v1.38.1/pkg/minikube/machine/cache_images.go#L338-L341
[3]: https://github.com/kubernetes-sigs/cri-tools/blob/v1.29.0/cmd/crictl/image.go#L458-L460
[4]: https://github.com/kubernetes-sigs/cri-tools/pull/1639

## Proposed fix

Two approaches are possible:

### Option 1: Remove the `removeExistingImage` step entirely (preferred)

Remove the `removeExistingImage` call from `transferAndLoadImage` altogether, as suggested in the review of #22993 ([ref: review comment][10]). This makes `minikube image load` semantics consistent with the underlying container runtime, i.e., users get the same behavior they would expect from `docker load` / `ctr images import` / `podman load` directly.

As shown in the verification below, all three container runtimes can overwrite an existing tag without prior removal, even when a container is using the old image. Removing this step eliminates minikube-specific logic that has been a recurring source of bugs (#11276, #11992, #19312).

[10]: https://github.com/kubernetes/minikube/pull/22993#issuecomment-4520716527

### Option 2: Make `removeExistingImage` best-effort

Keep the removal attempt but treat all errors as non-fatal (log and continue to `LoadImage`). It preserves the "try to clean up first" behavior while ensuring that `LoadImage` is always reached.

It looks like this approach was originally suggested during the review in #11366 ([ref: review comment][5]):

> or alternatively we could do
>
> `klog.Warniningf("failed to remove image (which might be okay if it doesn't exists")`

The initial commit of that PR ([`32a91d6`][6]) implemented this as a warning, but a follow-up commit within the same PR ([`16dbac1f`][7]) changed it to a fatal error to address performance concerns around `ListImages`. See the [full commit history of #11366][8] for context.

[5]: https://github.com/kubernetes/minikube/pull/11366#discussion_r631437500
[6]: https://github.com/kubernetes/minikube/commit/32a91d6eb29c477e4f197b4387f5bbee5592b2af
[7]: https://github.com/kubernetes/minikube/commit/16dbac1f66e00ab11b33d1e0fc10ae835ed60ae6
[8]: https://github.com/kubernetes/minikube/pull/11366/commits

### Comparison

| | Option 1 (remove step) | Option 2 (best-effort) |
|---|---|---|
| Behavior | Same as underlying container runtime | minikube-specific "try remove first" |
| Code change | Delete `removeExistingImage` and its call | Change error handling to non-fatal |
| Dangling images | Handled by container runtime natively | Same, but may remove some before load |
| Future maintenance | None | None (no more pattern matching) |

## Verification: container runtimes can overwrite tags without prior removal

We verified that all three container runtimes supported by minikube can overwrite an existing image tag while a container is using the old image. This proves that the `removeExistingImage` step is not a precondition for `LoadImage`.

Note: minikube uses `podman load` for the CRI-O runtime ([ref: v1.38.1 source][9]).

[9]: https://github.com/kubernetes/minikube/blob/v1.38.1/pkg/minikube/cruntime/crio.go#L273-L280

### Test environment

| Runtime | Version |
|---|---|
| Docker | 29.5.1 |
| containerd | 2.2.3 |
| Podman | 4.9.3 |

### Test setup

```dockerfile
# Dockerfile.v1
FROM busybox:latest
RUN echo "version1" > /version.txt
```

```dockerfile
# Dockerfile.v2
FROM busybox:latest
RUN echo "version2" > /version.txt
RUN echo "sample content" > /sample.txt
```

```bash
# Build and save tar files
docker buildx build -f Dockerfile.v1 -t repro-test:0.1.0 .
docker save repro-test:0.1.0 -o /tmp/v1.tar
docker buildx build -f Dockerfile.v2 -t repro-test:0.1.0 .
docker save repro-test:0.1.0 -o /tmp/v2.tar
```

### Test procedure

1. Load v1 image
2. Verify v1 image is loaded
3. Start a container using v1 (simulates a running deployment)
4. Verify the container is running
5. Load v2 image with the same tag while the container is running
6. Verify v2 image is loaded
7. Verify original container is still running
8. Check old image status

### Docker (`docker load`)

```bash
# 1. Load v1
docker load -i /tmp/v1.tar
# Output: Loaded image: repro-test:0.1.0

docker images --no-trunc --format '{{.ID}}' repro-test:0.1.0
# Output: sha256:27e2f051ea8985616e8f9291c09fd35727122ab0155cfd890923b6e0106a2400

# 2. Verify v1
docker run --rm repro-test:0.1.0 cat /version.txt
# Output: version1

# 3. Start container using v1
docker run -d --name repro-running repro-test:0.1.0 sleep 3600

# 4. Verify container is running
docker ps --filter name=repro-running --format '{{.Names}} {{.Status}}'
# Output: repro-running Up Less than a second

# 5. Load v2 (same tag, while container is running)
docker load -i /tmp/v2.tar
# Output:
#   The image repro-test:0.1.0 already exists, renaming the old one with
#   ID sha256:27e2... to empty string
#   Loaded image: repro-test:0.1.0

# 6. Verify v2 is loaded
docker run --rm repro-test:0.1.0 cat /version.txt
# Output: version2
docker run --rm repro-test:0.1.0 cat /sample.txt
# Output: sample content

# 7. Verify original container is still running
docker ps --filter name=repro-running --format '{{.Names}} {{.Status}}'
# Output: repro-running Up Less than a second

# 8. Check old image status
docker images --filter dangling=true --no-trunc --format '{{.ID}}'
# Output includes: sha256:27e2f051ea89... (v1 is now dangling)

# Cleanup
docker rm -f repro-running
docker rmi repro-test:0.1.0
```

Docker explicitly reports the tag reassignment in step 5: "renaming the old one ... to empty string". Old image becomes dangling (`<none>:<none>`) and can be cleaned up with `docker image prune`.

### containerd (`ctr images import`)

```bash
NS="test"
IMAGE="docker.io/library/repro-test:0.1.0"

# 1. Load v1
sudo ctr -n="$NS" images import /tmp/v1.tar
# Output: docker.io/library/repro-test:0.1.0 saved
#   sha256:0ba0043d1df615956e2528c841713ea3f3bec0a38d8ce0ece26982b93ef01793

sudo ctr -n="$NS" images ls "name==$IMAGE" | grep -o 'sha256:[a-f0-9]*'
# Output: sha256:0ba0043d1df615956e2528c841713ea3f3bec0a38d8ce0ece26982b93ef01793

# 2. Verify v1
sudo ctr -n="$NS" run --rm "$IMAGE" test-v1 cat /version.txt
# Output: version1

# 3. Start container using v1
sudo ctr -n="$NS" run -d "$IMAGE" repro-running /bin/sh -c 'sleep 3600'

# 4. Verify container is running
sudo ctr -n="$NS" tasks ls | grep repro-running
# Output: repro-running    2257496    RUNNING

# 4a. Record v1 snapshots (before v2 load)
sudo ctr -n="$NS" snapshots ls
# Output:
#   repro-running                  sha256:10dd19a4...  Active
#   sha256:10dd19a4...             sha256:7f74ca72...  Committed
#   sha256:7f74ca72...                                 Committed

# 5. Load v2 (same tag, while container is running)
sudo ctr -n="$NS" images import /tmp/v2.tar
# Output: docker.io/library/repro-test:0.1.0 saved
#   sha256:66fe5e60b722e9596232a64a99a79b8feb10bd95eb29b14ba1094d03339b2a4e

# 6. Verify v2 is loaded
sudo ctr -n="$NS" run --rm "$IMAGE" test-v2 cat /version.txt
# Output: version2
sudo ctr -n="$NS" run --rm "$IMAGE" test-v2b cat /sample.txt
# Output: sample content

# 7. Verify original container is still running
sudo ctr -n="$NS" tasks ls | grep repro-running
# Output: repro-running    2257496    RUNNING

# 8. Check old image status
sudo ctr -n="$NS" images ls "name==$IMAGE" | grep -o 'sha256:[a-f0-9]*'
# Output: sha256:66fe5e60b722... (tag now points to v2)

# v1 manifest may or may not be in content store (GC timing varies)
sudo ctr -n="$NS" content ls | grep sha256:0ba0043d1df6
# Output: present or absent depending on GC timing

# v1 container snapshot is retained (filesystem layers protected)
sudo ctr -n="$NS" containers info repro-running | grep SnapshotKey
# Output: "SnapshotKey": "repro-running"
sudo ctr -n="$NS" snapshots info repro-running
# Output: confirms snapshot exists

# All v1 snapshots from step 4a are retained while container is running
sudo ctr -n="$NS" snapshots info sha256:10dd19a4547510a810302be61ab179dc42af8565f8b3519688e21b2d5a217ed1
# Output: retained
sudo ctr -n="$NS" snapshots info sha256:7f74ca72855668453fc4207c1e5c50bb5653e72bb9939eb1d5512c001624e927
# Output: retained

# Cleanup
sudo ctr -n="$NS" tasks kill -s SIGKILL repro-running
sudo ctr -n="$NS" tasks rm repro-running
sudo ctr -n="$NS" containers rm repro-running
sudo ctr -n="$NS" images rm "$IMAGE"
```

containerd replaces the tag reference with the new digest. The v1 manifest (metadata) may be garbage-collected, but all v1 filesystem layers (snapshots) recorded in step 4a are confirmed retained after v2 load - the running container's snapshot chain is protected. Unlike Docker, containerd does not keep dangling images as `<none>:<none>` - old content is handled automatically by GC, while layers referenced by running containers are preserved.

### Podman (`podman load` - used by minikube for CRI-O runtime)

```bash
# 1. Load v1
podman load -i /tmp/v1.tar
# Output: Loaded image: docker.io/library/repro-test:0.1.0

podman images --no-trunc --format '{{.ID}}' repro-test:0.1.0
# Output: sha256:27e2f051ea8985616e8f9291c09fd35727122ab0155cfd890923b6e0106a2400

# 2. Verify v1
podman run --rm repro-test:0.1.0 cat /version.txt
# Output: version1

# 3. Start container using v1
podman run -d --name repro-running repro-test:0.1.0 sleep 3600

# 4. Verify container is running
podman ps --filter name=repro-running --format '{{.Names}} {{.Status}}'
# Output: repro-running Up Less than a second

# 5. Load v2 (same tag, while container is running)
podman load -i /tmp/v2.tar
# Output:
#   Copying blob 7f74ca728556 skipped: already exists
#   Copying blob 419389ad29ef done
#   Loaded image: docker.io/library/repro-test:0.1.0

# 6. Verify v2 is loaded
podman run --rm repro-test:0.1.0 cat /version.txt
# Output: version2
podman run --rm repro-test:0.1.0 cat /sample.txt
# Output: sample content

# 7. Verify original container is still running
podman ps --filter name=repro-running --format '{{.Names}} {{.Status}}'
# Output: repro-running Up 1 second

# 8. Check old image status
podman images --filter dangling=true --no-trunc --format '{{.ID}}'
# Output includes: sha256:27e2f051ea89... (v1 is now dangling)

# Cleanup
podman rm -f repro-running
podman rmi repro-test:0.1.0
```

Podman reuses existing layers ("skipped: already exists") and reassigns the tag. Old image becomes dangling, same as Docker.

Note: minikube uses `podman load` for the CRI-O runtime ([ref: v1.38.1 source][9]).

### Summary

| | Docker | containerd | Podman (CRI-O) |
|---|---|---|---|
| Tag overwrite | Yes | Yes | Yes |
| Running container affected | No | No | No |
| Old image layers (while container runs) | Retained as dangling (`<none>`) | Snapshots retained; manifest may be GC'd | Retained as dangling (`<none>`) |
| Prior removal required | No | No | No |

## Reproduction steps (minikube)

```bash
# Build v1 and v2
docker buildx build -f Dockerfile.v1 -t repro-test:0.1.0 .
docker buildx build -f Dockerfile.v2 -t repro-test:0.1.0 .

# Load v1 into minikube
minikube image load repro-test:0.1.0

# Start a container using v1
minikube ssh -- docker run -d --name repro-running repro-test:0.1.0 sleep 3600

# Rebuild as v2 and try to reload
docker buildx build -f Dockerfile.v2 -t repro-test:0.1.0 .
minikube image load --overwrite repro-test:0.1.0

# Verify: still v1 (silent failure)
minikube ssh -- docker run --rm repro-test:0.1.0 cat /version.txt
# Output: "version1" (expected: "version2")
```

monkey-mas:
## Summary

Remove the `removeExistingImage` step from `transferAndLoadImage` to fix `minikube image load` silently failing when the image is in use by a running container.

Container runtimes (`docker load`, `ctr images import`, `podman load`) can overwrite an existing tag without prior removal. Removing this step aligns `minikube image load` semantics with the underlying container runtime.

fixes #23023 
See the issue for detailed root cause analysis and verification across all three runtimes (Docker, containerd, Podman/CRI-O).

## Changes

1. Add `ImageLoadWhileInUse` integration test to `validateImageCommands` that verifies `minikube image load` can overwrite an existing image tag while a container is using the old image.
2. Remove `removeExistingImage` function and its call in `transferAndLoadImage`.

## Test results

The integration test (`ImageLoadWhileInUse`) was run at each commit:

| Commit | `removeExistingImage` | Test result |
|---|---|---|
| initial test-only commit | Present | **FAIL** - `expected 'version2' but got "version1" - image was not overwritten while in use` |
| fix commit | Removed | **PASS** |

```bash
# Run the test
go test -tags integration -v \
  -run "TestFunctional/parallel/ImageCommands/ImageLoadWhileInUse" \
  -timeout 10m ./test/integration/ \
  --binary="$(pwd)/out/minikube" --profile=minikube --cleanup=false
```

## Manual integration test

**Setup**: minikube running with Docker runtime.

**Dockerfiles**:

```dockerfile
# Dockerfile.v1
FROM busybox:latest
RUN echo "version1" > /version.txt
```

```dockerfile
# Dockerfile.v2
FROM busybox:latest
RUN echo "version2" > /version.txt
```

**Steps**:
```bash
# 1. Build and load v1
docker buildx build -f Dockerfile.v1 -t repro-test:0.1.0 .
minikube image load repro-test:0.1.0

# 2. Verify v1 is loaded
minikube ssh -- docker run --rm repro-test:0.1.0 cat /version.txt
# Expected: version1

# 3. Start a container using the image (simulates a running deployment)
minikube ssh -- docker run -d --name repro-running repro-test:0.1.0 sleep 3600

# 4. Build v2 with the same tag
docker buildx build -f Dockerfile.v2 -t repro-test:0.1.0 .

# 5. Load v2 with --overwrite
minikube image load --overwrite repro-test:0.1.0

# 6. Verify v2 is loaded
minikube ssh -- docker run --rm repro-test:0.1.0 cat /version.txt
# Before fix: "version1" (silent failure)
# After fix:  "version2"

minikube ssh -- docker run --rm repro-test:0.1.0 cat /sample.txt
# Before fix: "cat: can't open '/sample.txt': No such file or directory"
# After fix:  "sample content"

# 7. Cleanup
minikube ssh -- docker rm -f repro-running
minikube ssh -- docker rmi repro-test:0.1.0
docker rmi repro-test:0.1.0
```

### DIFF
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -319,10 +319,6 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
 	t.Run("ImageBuild", func(t *testing.T) {
 		MaybeParallel(t)
 
-		if _, err := Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "pgrep", "buildkitd")); err == nil {
-			t.Errorf("buildkitd process is running, should not be running until `minikube image build` is ran")
-		}
-
 		newImage := fmt.Sprintf("localhost/my-image:%s", profile)
 
 		// try to build the new image with minikube
--- a/test/integration/functional_test.go
+++ b/test/integration/functional_test.go
@@ -380,6 +380,44 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
 		checkImageExists(ctx, t, profile, taggedImage)
 	})
 
+	// docs: Make sure image loading overwrites existing image even when a container is using it (#23023)
+	t.Run("ImageLoadInUse", func(t *testing.T) {
+		// Loads v1 image, starts a pod with it, then loads v2 with the same
+		// tag and verifies that a new pod picks up v2.
+		testImage := "localhost/image-load-in-use:test"
+		testDir := filepath.Join(*testdataDir, "image-load-in-use")
+		tmpDir := t.TempDir()
+
+		// Build images and store tar files.
+		// t.Cleanup ensures the image is removed from the cluster even if
+		// saveImageToTarfile or subsequent steps fail. In the normal case,
+		// the image is explicitly removed after saving, then reloaded via
+		// loadImageFromTarfile — t.Cleanup handles the final removal.
+		buildImage(ctx, t, profile, filepath.Join(testDir, "v1"), testImage)
+		t.Cleanup(func() { silentRemoveImage(ctx, t, profile, testImage) })
+		v1TarPath := filepath.Join(tmpDir, "image-load-in-use-v1.tar")
+		saveImageToTarfile(ctx, t, profile, testImage, v1TarPath)
+		removeImage(ctx, t, profile, testImage)
+		buildImage(ctx, t, profile, filepath.Join(testDir, "v2"), testImage)
+		v2TarPath := filepath.Join(tmpDir, "image-load-in-use-v2.tar")
+		saveImageToTarfile(ctx, t, profile, testImage, v2TarPath)
+		removeImage(ctx, t, profile, testImage)
+
+		// Load image and start a pod using it
+		loadImageFromTarfile(ctx, t, profile, v1TarPath)
+		startTempPod(ctx, t, profile, filepath.Join(testDir, "image-load-in-use-v1.yaml"), "image-load-in-use-v1")
+		if got := readPodFile(ctx, t, profile, "image-load-in-use-v1", "/version.txt"); strings.TrimSpace(got) != "version1" {
+			t.Fatalf("expected 'version1' but got %q", got)
+		}
+
+		// Load image overwriting image in use
+		loadImageFromTarfile(ctx, t, profile, v2TarPath)
+		startTempPod(ctx, t, profile, filepath.Join(testDir, "image-load-in-use-v2.yaml"), "image-load-in-use-v2")
+		if got := readPodFile(ctx, t, profile, "image-load-in-use-v2", "/version.txt"); strings.TrimSpace(got) != "version2" {
+			t.Fatalf("expected 'version2' but got %q - image was not overwritten while in use", got)
+		}
+	})
+
 	// docs: Make sure a new updated tag works by `minikube image load --daemon`
 	t.Run("ImageTagAndLoadDaemon", func(t *testing.T) {
 		tagAndLoadImage(ctx, t, profile, taggedImage)
@@ -459,6 +497,73 @@ func checkImageExists(ctx context.Context, t *testing.T, profile string, image s
 
 func listImages(ctx context.Context, t *testing.T, profile string) (*RunResult, error) {
 	return Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "image", "ls"))
+}
+
+func buildImage(ctx context.Context, t *testing.T, profile, buildDir, imageName string) {
+	t.Helper()
+	cmd := exec.CommandContext(ctx, Target(), "-p", profile, "image", "build", "-t", imageName, buildDir, "--alsologtostderr")
+	if rr, err := Run(t, cmd); err != nil {
+		t.Fatalf("failed to build image %s (%s): %v\n%s", imageName, buildDir, err, rr.Output())
+	}
+}
+
+func removeImage(ctx context.Context, t *testing.T, profile, imageName string) {
+	t.Helper()
+	cmd := exec.CommandContext(ctx, Target(), "-p", profile, "image", "rm", imageName)
+	if rr, err := Run(t, cmd); err != nil {
+		t.Fatalf("failed to remove image %s: %v\n%s", imageName, err, rr.Output())
+	}
+}
+
+// silentRemoveImage is an idempotent variant of removeImage for use in
+// t.Cleanup. It ignores errors because minikube image rm does not provide
+// specific error codes to distinguish "image not found" from other failures.
+// Run() already logs errors, so we don't need additional logging here.
+func silentRemoveImage(ctx context.Context, t *testing.T, profile, imageName string) {
+	t.Helper()
+	cmd := exec.CommandContext(ctx, Target(), "-p", profile, "image", "rm", imageName)
+	Run(t, cmd)
+}
+
+func saveImageToTarfile(ctx context.Context, t *testing.T, profile, imageName, tarPath string) {
+	t.Helper()
+	cmd := exec.CommandContext(ctx, Target(), "-p", profile, "image", "save", imageName, tarPath, "--alsologtostderr")
+	if rr, err := Run(t, cmd); err != nil {
+		t.Fatalf("failed to save image %s to %s: %v\n%s", imageName, tarPath, err, rr.Output())
+	}
+}
+
+func loadImageFromTarfile(ctx context.Context, t *testing.T, profile, tarPath string) {
+	t.Helper()
+	cmd := exec.CommandContext(ctx, Target(), "-p", profile, "image", "load", tarPath, "--alsologtostderr")
+	if rr, err := Run(t, cmd); err != nil {
+		t.Fatalf("failed to load image from %s: %v\n%s", tarPath, err, rr.Output())
+	}
+}
+
+func startTempPod(ctx context.Context, t *testing.T, profile, podYAMLPath, podName string) {
+	t.Helper()
+	cmd := exec.CommandContext(ctx, Target(), "-p", profile, "kubectl", "--", "apply", "-f", podYAMLPath)
+	if rr, err := Run(t, cmd); err != nil {
+		t.Fatalf("failed to create pod %s: %v\n%s", podName, err, rr.Output())
+	}
+	t.Cleanup(func() {
+		cmd := exec.CommandContext(ctx, Target(), "-p", profile, "kubectl", "--", "delete", "pod", podName, "--ignore-not-found")
+		Run(t, cmd)
+	})
+	if _, err := PodWait(ctx, t, profile, "default", "run="+podName, Minutes(3)); err != nil {
+		t.Fatalf("failed waiting for pod %s: %v", podName, err)
+	}
+}
+
+func readPodFile(ctx context.Context, t *testing.T, profile, podName, filePath string) string {
+	t.Helper()
+	cmd := exec.CommandContext(ctx, Target(), "-p", profile, "kubectl", "--", "exec", podName, "--", "cat", filePath)
+	rr, err := Run(t, cmd)
+	if err != nil {
+		t.Fatalf("failed to read %s from pod %s: %v\n%s", filePath, podName, err, rr.Output())
+	}
+	return rr.Stdout.String()
 }
 
 // check functionality of minikube after evaluating docker-env
--- a/test/integration/testdata/image-load-in-use/image-load-in-use-v1.yaml
+++ b/test/integration/testdata/image-load-in-use/image-load-in-use-v1.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: image-load-in-use-v1
+  labels:
+    run: image-load-in-use-v1
+spec:
+  containers:
+  - name: test
+    image: localhost/image-load-in-use:test
+    imagePullPolicy: Never
+    command: ["sleep", "infinity"]
+  restartPolicy: Never
--- a/test/integration/testdata/image-load-in-use/image-load-in-use-v2.yaml
+++ b/test/integration/testdata/image-load-in-use/image-load-in-use-v2.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: image-load-in-use-v2
+  labels:
+    run: image-load-in-use-v2
+spec:
+  containers:
+  - name: test
+    image: localhost/image-load-in-use:test
+    imagePullPolicy: Never
+    command: ["sleep", "infinity"]
+  restartPolicy: Never
--- a/pkg/minikube/machine/cache_images.go
+++ b/pkg/minikube/machine/cache_images.go
@@ -284,10 +284,6 @@ func transferAndLoadImage(cr command.Runner, k8s config.KubernetesConfig, src st
 		return fmt.Errorf("runtime: %w", err)
 	}
 
-	if err := removeExistingImage(r, src, imgName); err != nil {
-		return err
-	}
-
 	klog.Infof("Loading image from: %s", src)
 	filename := filepath.Base(src)
 	if _, err := os.Stat(src); err != nil {
@@ -321,26 +317,6 @@ func transferAndLoadImage(cr command.Runner, k8s config.KubernetesConfig, src st
 	}
 
 	klog.Infof("Transferred and loaded %s from cache", src)
-	return nil
-}
-
-func removeExistingImage(r cruntime.Manager, src string, imgName string) error {
-	// if loading an image from tar, skip deleting as we don't have the actual image name
-	// ie. imgName = "C:\this_is_a_dir\image.tar.gz"
-	if src == imgName {
-		return nil
-	}
-
-	err := r.RemoveImage(imgName)
-	if err == nil {
-		return nil
-	}
-
-	errStr := strings.ToLower(err.Error())
-	if !strings.Contains(errStr, "no such image") && !strings.Contains(errStr, "unable to remove the image") {
-		return fmt.Errorf("removing image: %w", err)
-	}
-
 	return nil
 }
 
--- a/test/integration/testdata/image-load-in-use/v1/Dockerfile
+++ b/test/integration/testdata/image-load-in-use/v1/Dockerfile
@@ -1,0 +1,2 @@
+FROM busybox:latest
+RUN echo "version1" > /version.txt
--- a/test/integration/testdata/image-load-in-use/v2/Dockerfile
+++ b/test/integration/testdata/image-load-in-use/v2/Dockerfile
@@ -1,0 +1,2 @@
+FROM busybox:latest
+RUN echo "version2" > /version.txt


nirs (commented):
@monkey-mas Change looks good, but it will be better to start with a failing test. The fact that image tests do not fail with the current code means that we don't covert this use case, or the test is missing verification that would fail with current code.

Current image tests are not great shape. I hope you can add a simple test reproducing this issue. Then add the fix in the next commit. Without the test this will break again easily.

test/integration/testdata/image-load-overwrite/Dockerfile.v2
```diff
@@ -0,0 +1,3 @@
+FROM busybox:latest
+RUN echo "version2" > /version.txt
+RUN echo "sample content" > /sample.txt
```
nirs: Do we need this layer? in't one layer enough to tell the version of the image?

test/integration/functional_test.go
```diff
@@ -384,6 +384,68 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
 		checkImageExists(ctx, t, profile, taggedImage)
 	})
 
+	// docs: Make sure image loading overwrites existing image even when a container is using it (#23023)
+	t.Run("ImageLoadWhileInUse", func(t *testing.T) {
+		testImage := "localhost/load-overwrite-test:0.1.0"
+		testDir := filepath.Join(*testdataDir, "image-load-overwrite")
+
+		// Build and load v1
+		rr, err := Run(t, exec.CommandContext(ctx, "docker", "build", "-t", testImage, "-f", filepath.Join(testDir, "Dockerfile.v1"), testDir))
```
nirs: This requires docker so it will not work on macOS. We can use podman which is open source and available everywhere, or build the image inside minikube which avoid the external dependency.

But this is not a new issue - existing tests already have this problem so this should not block this change.

test/integration/functional_test.go
```diff
@@ -384,6 +384,68 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+		if err != nil {
+			t.Fatalf("failed to build v1 image: %v\n%s", err, rr.Output())
+		}
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "image", "load", "--daemon", testImage, "--alsologtostderr"))
```
nirs: Same dependency issue. Can build solve by saving the image from minikube to tar file on the host and loading the tar file.

test/integration/functional_test.go
```diff
@@ -384,6 +384,68 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+		}
+
+		// Start a container using v1 (simulates a running deployment)
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "--", "docker", "run", "-d", "--name", "overwrite-test", testImage, "sleep", "3600"))
+		if err != nil {
+			t.Fatalf("failed to start container: %v\n%s", err, rr.Output())
+		}
+		defer func() {
+			Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "--", "docker", "rm", "-f", "overwrite-test"))
+		}()
```
nirs: This works, but in tests it is better to use t.Cleanup()

test/integration/functional_test.go
```diff
@@ -384,6 +384,68 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+		// Verify v2 is loaded (not v1)
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "--", "docker", "run", "--rm", testImage, "cat", "/version.txt"))
+		if err != nil {
+			t.Fatalf("failed to verify v2: %v\n%s", err, rr.Output())
+		}
+		if !strings.Contains(rr.Stdout.String(), "version2") {
+			t.Fatalf("expected 'version2' but got %q — image was not overwritten while in use", rr.Stdout.String())
+		}
+
+		// Verify sample.txt exists (only in v2)
```
nirs: Why do we need this? we already verified that the current image was updated.

nirs:
@monkey-mas I modified the PR title to make it more useful in the release notes.

This will help users identify a useful fix. Removing the removeExistingImage is the way this was fixed, but the user visible effect is fixing the issue where image load command does not fail but the image is not replaced.

Fill free to fix the title if I got this wrong.

test/integration/testdata/image-load-overwrite/Dockerfile.v2
monkey-mas: Yeah, so true. Let me drop this unnecessary testing item...
fixed in: https://github.com/kubernetes/minikube/pull/23034/commits/5363f4a7f750b8bde5190bfa13ed85a5268b5072

test/integration/functional_test.go
monkey-mas: Removed it here as well!
https://github.com/kubernetes/minikube/pull/23034#discussion_r3293377264

test/integration/functional_test.go
monkey-mas: Thanks for sharing the background. Just wanted to double-check that we don't have to fix this in this PR? :eyes:

test/integration/functional_test.go
monkey-mas: Thanks, done!
https://github.com/kubernetes/minikube/pull/23034/commits/5363f4a7f750b8bde5190bfa13ed85a5268b5072

monkey-mas:
> I modified the PR title to make it more useful in the release notes.

@nirs Thank you! Yep, the updated title is more informative :)

test/integration/functional_test.go
nirs: No need to fix this in this PR, we want to fix the silent failure and have a test to protect us from future failure. We cannot fix the entire world at once.

test/integration/testdata/image-load-overwrite/Dockerfile.v2
```diff
@@ -1,3 +1,2 @@
 FROM busybox:latest
 RUN echo "version2" > /version.txt
-RUN echo "sample content" > /sample.txt
```
nirs: Let's fixup this commit into the test commit for cleaner git history. We can squash all the commits during merge but is manual step, error prone and creates ugly commit message.

This also makes it harder to review for new reviewers. GitHub supports comparing commits so there is no need to push review fixes as separate commit.

nirs (commented):
@monkey-mas Looks ready after fixing up the last commit into the first.

test/integration/functional_test.go
nirs: I tried to rewrite these tests here:
https://github.com/kubernetes/minikube/pull/21405

Still requires lot of cleanup and I did not have time to go back to this. Maybe you want to continue this work?

test/integration/testdata/image-load-overwrite/Dockerfile.v2
monkey-mas: Oops, sorry about! It should be ok for now :pray:

test/integration/functional_test.go
monkey-mas: Ah, got it - let me take a look at it :) (Not 100% for sure if I can fix them properly though 😅)

test/integration/functional_test.go
```diff
@@ -384,6 +384,59 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+
+		// Build and load v1
+		rr, err := Run(t, exec.CommandContext(ctx, "docker", "build", "-t", testImage, "-f", filepath.Join(testDir, "Dockerfile.v1"), testDir))
+		if err != nil {
+			t.Fatalf("failed to build v1 image: %v\n%s", err, rr.Output())
+		}
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "image", "load", "--daemon", testImage, "--alsologtostderr"))
+		if err != nil {
+			t.Fatalf("failed to load v1 image: %v\n%s", err, rr.Output())
+		}
```
nirs: We want to clean up this image when the test is completed so other tests are not affected by this image.

nirs (commented):
One issue I missed in the previous review - we need to remove the image loaded by the tests so other tests do not break by finding unexpected image.

test/integration/functional_test.go
monkey-mas: Oops - added the cleanup logic here in the first testing commit.
```diff
@@ -415,6 +415,8 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
                }
                t.Cleanup(func() {
                        Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "--", "docker", "rm", "-f", "overwrite-test"))
+                       Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "image", "rm", testImage))
+                       Run(t, exec.CommandContext(ctx, "docker", "rmi", testImage))
                })

```

test/integration/functional_test.go
```diff
@@ -384,6 +384,61 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+		// Start a container using v1 (simulates a running deployment)
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "--", "docker", "run", "-d", "--name", "overwrite-test", testImage, "sleep", "3600"))
+		if err != nil {
+			t.Fatalf("failed to start container: %v\n%s", err, rr.Output())
+		}
+		t.Cleanup(func() {
+			Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "--", "docker", "rm", "-f", "overwrite-test"))
+			Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "image", "rm", testImage))
+			Run(t, exec.CommandContext(ctx, "docker", "rmi", testImage))
+		})
```
nirs: This will not clean up if we fail before this call. The best way to clean up is to add a t.Cleanup() call for each step that modified the cluster or the environment. The test will run the cleanup function in reverse order.
https://pkg.go.dev/testing#T.Cleanup

monkey-mas: Ah, I see. IIUC, with a single `t.Cleanup()` block like I did, if the test fails before the `t.Cleanup()` is registered (e.g., right after loading v1), no cleanup runs at all. Conversely, if it fails after registration but before the container is started, the cleanup tries to remove a container that doesn't exist (harmless but unnecessary).

I've updated the code to register `t.Cleanup()` after each step that modifies the cluster or the host environment. This way, only the resources created so far are cleaned up, in reverse order.

```diff
@@ -389,15 +389,23 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
                testImage := "localhost/load-overwrite-test:0.1.0"
                testDir := filepath.Join(*testdataDir, "image-load-overwrite")
 
-               // Build and load v1
+               // Build v1 on the host
                rr, err := Run(t, exec.CommandContext(ctx, "docker", "build", "-t", testImage, "-f", filepath.Join(testDir, "Dockerfile.v1"), testDir))
                if err != nil {
                        t.Fatalf("failed to build v1 image: %v\n%s", err, rr.Output())
                }
+               t.Cleanup(func() {
+                       Run(t, exec.CommandContext(ctx, "docker", "rmi", testImage))
+               })
+
+               // Load v1 into minikube
                rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "image", "load", "--daemon", testImage, "--alsologtostderr"))
                if err != nil {
                        t.Fatalf("failed to load v1 image: %v\n%s", err, rr.Output())
                }
+               t.Cleanup(func() {
+                       Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "image", "rm", testImage))
+               })
 
                // Verify v1 is loaded
                rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "--", "docker", "run", "--rm", testImage, "cat", "/version.txt"))
@@ -415,8 +423,6 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
                }
                t.Cleanup(func() {
                        Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "--", "docker", "rm", "-f", "overwrite-test"))
-                       Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "image", "rm", testImage))
-                       Run(t, exec.CommandContext(ctx, "docker", "rmi", testImage))
                })
```

nirs (dismissed):
@monkey-mas Thanks!

test/integration/functional_test.go
```diff
@@ -384,6 +384,67 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+		}
+		t.Cleanup(func() {
+			Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "image", "rm", testImage))
+		})
+
+		// Verify v1 is loaded
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "--", "docker", "run", "--rm", testImage, "cat", "/version.txt"))
+		if err != nil {
+			t.Fatalf("failed to verify v1: %v\n%s", err, rr.Output())
+		}
```
nirs: This works only with docker runtime. We need to find another way to detect the image.

```
=== NAME  TestFunctional/parallel/ImageCommands/ImageLoadWhileInUse
    functional_test.go:393: (dbg) Done: docker build -t localhost/load-overwrite-test:0.1.0 -f testdata/image-load-overwrite/Dockerfile.v1 testdata/image-load-overwrite: (1.164897666s)
    functional_test.go:402: (dbg) Run:  ./minikube-linux-amd64 -p functional-538876 image load --daemon localhost/load-overwrite-test:0.1.0 --alsologtostderr
    functional_test.go:411: (dbg) Run:  ./minikube-linux-amd64 -p functional-538876 ssh -- docker run --rm localhost/load-overwrite-test:0.1.0 cat /version.txt
    functional_test.go:411: (dbg) Non-zero exit: ./minikube-linux-amd64 -p functional-538876 ssh -- docker run --rm localhost/load-overwrite-test:0.1.0 cat /version.txt: exit status 1 (257.242336ms)
        
        -- stdout --
        	Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
        
        -- /stdout --
        ** stderr ** 
```

The portable way is to run a pod using this image - we already do this in the integration tests, and use kubectl exec to get the version.

nirs (commented):
The new tests works only with docker runtime. It should work with all container runtimes.

test/integration/functional_test.go
monkey-mas: Thank you :pray: Updated the test (to use `kubectl`)

```diff
@@ -386,6 +386,8 @@
 
 	// docs: Make sure image loading overwrites existing image even when a container is using it (#23023)
 	t.Run("ImageLoadWhileInUse", func(t *testing.T) {
+		// Use kubectl to run pods instead of "minikube ssh -- docker run" to
+		// work with all container runtimes (docker, containerd, cri-o).
 		testImage := "localhost/load-overwrite-test:0.1.0"
 		testDir := filepath.Join(*testdataDir, "image-load-overwrite")
 
@@ -407,25 +409,26 @@
 			Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "image", "rm", testImage))
 		})
 
-		// Verify v1 is loaded
-		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "--", "docker", "run", "--rm", testImage, "cat", "/version.txt"))
+		// Start a pod using v1 (simulates a running deployment) and verify v1
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "kubectl", "--", "run", "overwrite-test", "--image="+testImage, "--image-pull-policy=Never", "--restart=Never", "--", "sleep", "3600"))
 		if err != nil {
-			t.Fatalf("failed to verify v1: %v\n%s", err, rr.Output())
+			t.Fatalf("failed to start pod: %v\n%s", err, rr.Output())
 		}
-		if !strings.Contains(rr.Stdout.String(), "version1") {
-			t.Fatalf("expected 'version1' but got %q", rr.Stdout.String())
+		t.Cleanup(func() {
+			Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "kubectl", "--", "delete", "pod", "overwrite-test", "--ignore-not-found"))
+		})
+		if _, err := PodWait(ctx, t, profile, "default", "run=overwrite-test", Minutes(3)); err != nil {
+			t.Fatalf("failed waiting for overwrite-test pod: %v", err)
 		}
-
-		// Start a container using v1 (simulates a running deployment)
-		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "--", "docker", "run", "-d", "--name", "overwrite-test", testImage, "sleep", "3600"))
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "kubectl", "--", "exec", "overwrite-test", "--", "cat", "/version.txt"))
 		if err != nil {
-			t.Fatalf("failed to start container: %v\n%s", err, rr.Output())
+			t.Fatalf("failed to get output of v1 image: %v\n%s", err, rr.Output())
+		}
+		if !strings.Contains(rr.Stdout.String(), "version1") {
+			t.Fatalf("expected 'version1' but got %q", rr.Stdout.String())
 		}
-		t.Cleanup(func() {
-			Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "--", "docker", "rm", "-f", "overwrite-test"))
-		})
 
-		// Build and load v2 with the same tag while container is running
+		// Build and load v2 with the same tag while pod is running
 		rr, err = Run(t, exec.CommandContext(ctx, "docker", "build", "-t", testImage, "-f", filepath.Join(testDir, "Dockerfile.v2"), testDir))
 		if err != nil {
 			t.Fatalf("failed to build v2 image: %v\n%s", err, rr.Output())
@@ -435,10 +438,20 @@
 			t.Fatalf("failed to load v2 image: %v\n%s", err, rr.Output())
 		}
 
-		// Verify v2 is loaded (not v1)
-		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "--", "docker", "run", "--rm", testImage, "cat", "/version.txt"))
+		// Verify v2 is loaded by running a new pod
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "kubectl", "--", "run", "verify-v2", "--image="+testImage, "--image-pull-policy=Never", "--restart=Never", "--", "sleep", "3600"))
+		if err != nil {
+			t.Fatalf("failed to create v2 verify pod: %v\n%s", err, rr.Output())
+		}
+		t.Cleanup(func() {
+			Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "kubectl", "--", "delete", "pod", "verify-v2", "--ignore-not-found"))
+		})
+		if _, err := PodWait(ctx, t, profile, "default", "run=verify-v2", Minutes(3)); err != nil {
+			t.Fatalf("failed waiting for v2 verify pod: %v", err)
+		}
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "kubectl", "--", "exec", "verify-v2", "--", "cat", "/version.txt"))
 		if err != nil {
-			t.Fatalf("failed to verify v2: %v\n%s", err, rr.Output())
+			t.Fatalf("failed to get output of v2 image: %v\n%s", err, rr.Output())
 		}
 		if !strings.Contains(rr.Stdout.String(), "version2") {
 			t.Fatalf("expected 'version2' but got %q - image was not overwritten while in use", rr.Stdout.String())
```

test/integration/functional_test.go
```diff
@@ -384,6 +384,80 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
 
+	// docs: Make sure image loading overwrites existing image even when a container is using it (#23023)
+	t.Run("ImageLoadWhileInUse", func(t *testing.T) {
+		// Use kubectl to run pods instead of "minikube ssh -- docker run" to
+		// work with all container runtimes (docker, containerd, cri-o).
+		testImage := "localhost/load-overwrite-test:0.1.0"
+		testDir := filepath.Join(*testdataDir, "image-load-overwrite")
+
+		// Build v1 on the host
+		rr, err := Run(t, exec.CommandContext(ctx, "docker", "build", "-t", testImage, "-f", filepath.Join(testDir, "Dockerfile.v1"), testDir))
```
nirs: This fails with using rootless option:
https://github.com/kubernetes/minikube/actions/runs/26358537109/job/77590833918?pr=23034

But TestFunctional/parallel/ImageCommands/ImageBuild pass - building inside minikube. So we should switch to build inside minikube and use image save to save to image to a tar file. Then we can test image load with the tar file.

We actually need to test both load from daemon and load from file, but loading from daemon is harder to test since docker is not available everywhere. We can test using podman but I'm not sure minikube support loading images from podman (e.g it may depend on docker socket).

So the fix should be:
- build using minikube - see ImageBuild test
- save image to tar file
- remove image from minikube - we want to load image once without replacing, and then with replacing.
- load the tar file instead of container image

test/integration/functional_test.go
```diff
@@ -384,6 +384,80 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+		if err != nil {
+			t.Fatalf("failed to build v1 image: %v\n%s", err, rr.Output())
+		}
+		t.Cleanup(func() {
+			Run(t, exec.CommandContext(ctx, "docker", "rmi", testImage))
+		})
```
nirs: Not needed when we move to build inside minikube.

test/integration/functional_test.go
```diff
@@ -384,6 +384,80 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+
+		// Load v1 into minikube
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "image", "load", "--daemon", testImage, "--alsologtostderr"))
```
nirs: remove --daemon, use path to the tar file.

test/integration/functional_test.go
```diff
@@ -384,6 +384,80 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+		if err != nil {
+			t.Fatalf("failed to load v1 image: %v\n%s", err, rr.Output())
+		}
+		t.Cleanup(func() {
+			Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "image", "rm", testImage))
+		})
+
+		// Start a pod using v1 (simulates a running deployment) and verify v1
+		rr, err = Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "kubectl", "--", "run", "overwrite-test", "--image="+testImage, "--image-pull-policy=Never", "--restart=Never", "--", "sleep", "3600"))
```
nirs: This is too long and hard to follow. Format in multiple lines, one arguemment and value per line.

```go
exec.CommandContext(
    ctx,
    Target(),
    "-p", profile,
    "kubectl",
    "--",
    "run",
    "overwrite-test",
    "--image=", testImage,
    "--image-pull-policy", "Never",
    "--restart", "Never",
    "--",
    "sleep",
    "infinity",
)
```

Note we can sleep forever.

But consider adding a pod yaml to testdata instead to make this cleaner.

nirs (commented):
Test sill broken with rootless - we can fix by building in minikube.

test/integration/functional_test.go
monkey-mas: Thanks for the detailed feedback :pray:

> But TestFunctional/parallel/ImageCommands/ImageBuild pass - building inside minikube. So we should switch to build inside minikube and use image save to save to image to a tar file. Then we can test image load with the tar file.

For this PR, the test now builds images inside minikube using `minikube image build`, saves them as tar files, and loads from tar. This avoids the dependency on host docker. 

https://github.com/kubernetes/minikube/pull/23034/commits/84d859237259a1f8386fed26222904d7f180f4ad: I've added a new commit for this change and will squash it into the test commit once the CI integration tests pass.


> We actually need to test both load from daemon and load from file, but loading from daemon is harder to test since docker is not available everywhere. We can test using podman but I'm not sure minikube support loading images from podman (e.g it may depend on docker socket).

I did some investigation and it looks like we can test loading from a Podman daemon - Podman implements a Docker-compatible API, so pointing `DOCKER_HOST` to a Podman socket should work with `minikube image load --daemon`. (I haven't verified this end-to-end yet though..)
                                            
That said, it might expand the scope of this PR a bit, so it could also be a follow-up. Let me know if you'd like to include it in this PR or keep it separate. Happy to work on it either way :)

test/integration/functional_test.go
monkey-mas: Removed - no longer needed since we build inside minikube.

test/integration/functional_test.go
monkey-mas: Switched to loading from tar file path instead of `--daemon`.

test/integration/functional_test.go
monkey-mas: Formatted long commands in multiple lines. Also added pod YAML files in `testdata/image-load-overwrite/` for the overwrite-test and verify-v2 pods, replacing `kubectl run` with `kubectl apply -f`.

test/integration/testdata/image-load-overwrite/overwrite-test-pod.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: overwrite-test
+  labels:
+    run: overwrite-test
```
nirs: Not needed

test/integration/testdata/image-load-overwrite/verify-v2-pod.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: verify-v2
+  labels:
+    run: verify-v2
```
nirs: not needed

test/integration/testdata/image-load-overwrite/overwrite-test-pod.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: overwrite-test
```
nirs: name: image-load-in-use-v1

test/integration/testdata/image-load-overwrite/verify-v2-pod.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: verify-v2
```
nirs: name: image-load-in-use-v2

test/integration/testdata/image-load-in-use/image-load-in-use-v1.yaml
nirs, line 1: use pod name?

test/integration/testdata/image-load-in-use/image-load-in-use-v2.yaml
nirs, line 1: use pod name?

test/integration/testdata/image-load-overwrite/verify-v2-pod.yaml
```diff
@@ -0,0 +1,13 @@
...
+  labels:
+    run: verify-v2
+spec:
+  containers:
+  - name: verify-v2
+    image: load-overwrite-test:0.1.0
```
nirs: image-load-in-use:tag?

test/integration/testdata/image-load-overwrite/overwrite-test-pod.yaml
```diff
@@ -0,0 +1,13 @@
...
+  labels:
+    run: overwrite-test
+spec:
+  containers:
+  - name: overwrite-test
+    image: load-overwrite-test:0.1.0
```
nirs: image-load-in-use:tag?

test/integration/testdata/image-load-overwrite/overwrite-test-pod.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: overwrite-test
+  labels:
+    run: overwrite-test
+spec:
+  containers:
+  - name: overwrite-test
```
nirs: The container name does not matter much, no need to repeat the test name here.

test/integration/testdata/image-load-overwrite/verify-v2-pod.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: verify-v2
+  labels:
+    run: verify-v2
+spec:
+  containers:
+  - name: verify-v2
```
nirs: container name does not matter, we can use the same name

test/integration/functional_test.go
```diff
@@ -388,68 +388,121 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
 	t.Run("ImageLoadWhileInUse", func(t *testing.T) {
```
nirs: ImageLoadInUse? it will help if the test and all files and resources use the same name - makes it easy to debug later.

test/integration/functional_test.go
```diff
@@ -388,68 +388,121 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
 		// Use kubectl to run pods instead of "minikube ssh -- docker run" to
 		// work with all container runtimes (docker, containerd, cri-o).
-		testImage := "localhost/load-overwrite-test:0.1.0"
+		testImage := "load-overwrite-test:0.1.0"
```
nirs: image-load-in-use:test

test/integration/functional_test.go
```diff
@@ -388,68 +388,121 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+			if err != nil {
+				t.Fatalf("failed to save image (%s): %v\n%s", buildDir, err, rr.Output())
+			}
+			rr, err = Run(t, exec.CommandContext(ctx,
+				Target(), "-p", profile,
+				"image", "rm", testImage))
+			if err != nil {
+				t.Fatalf("failed to remove image (%s): %v\n%s", buildDir, err, rr.Output())
+			}
+			return tarPath
```
nirs: Adding a helper make sense but this is very specific helper and not very reusable. It will be more reusable if we add these helpers:

```
buildImage(t *testing.T, containerFile, imageName string)
removeImage(t *testing.T, imageName string)
saveImageToTarfile(t *testing.T, imageName string)
loadImageFromTarfile(t *testing.T, imagePath string)
readPodFile(t *testing.T, podName, fileName string) string
```

With these helpers we can write the entire test in a much cleaner way:

```go
    // Build images and store tar files
    buildImage(t, containerFile1, testImage)
    saveImageToTarfile(t, testImage, tarPath1)
    removeImage(t, testImage)
    buildImage(t, containerFile2, testImage)
    saveImageToTarfile(t, testImage, tarPath2)
    removeImage(t, testImage)

    // Load image and start a pod using it.
    loadImageFromTarfile(t, tarPath1)
    startPod(t, podName1)
    if readPodFile(t, podName1, "/version.txt") != "version1" {
        t.Fatal(...)
    }
    
    // Load image overwriting image in use.
    loadImageFromTarfile(t, tarPath2)
    startPod(t, podName2)
    if readPodFile(t, podName2, "/version.txt") != "version1" {
        t.Fatal(...)
    } 
```

And we can use the helpers later in other tests to avoid the boileplate we have now.

Since you already create helper functions, lets make them more useful.

nirs:
Now the new test pass, but ImageBuild test fails:
https://github.com/kubernetes/minikube/actions/runs/26366586240/job/77704913642?pr=23034

ImageBuild assumes that it is the only test build images - need to remove this wrong check:

```go
		if _, err := Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "pgrep", "buildkitd")); err == nil {
			t.Errorf("buildkitd process is running, should not be running until `minikube image build` is ran")
		}
```

@monkey-mas: The clean way to fix this is to add another commit before the new test removing this check.

monkey-mas:
> Now the new test pass, but ImageBuild test fails:
https://github.com/kubernetes/minikube/actions/runs/26366586240/job/77704913642?pr=23034
>
> ImageBuild assumes that it is the only test build images - need to remove this wrong check:
>
>		if _, err := Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", "pgrep", "buildkitd")); > err == nil {
> 			t.Errorf("buildkitd process is running, should not be running until `minikube image build` > is ran")
> 		}
> @monkey-mas: The clean way to fix this is to add another commit before the new test removing this check.

@nirs  Thanks for confirming. I've added a commit to remove the pre-check before the test commit. (https://github.com/kubernetes/minikube/pull/23034/changes/f51f5c98efd4c9559933f66345abb26cacdb3c79)

For context & self-note: as you might know, on containerd runtime, `minikube image build` uses `buildctl` which triggers systemd socket activation to start `buildkitd`. Once started, `buildkitd` remains running for the lifetime of the node. Since `ImageBuild` uses `MaybeParallel(t)`, it gets paused until all sequential subtests complete. If any sequential subtest(eg `ImageLoadInUse`) calls `minikube image build` before `ImageBuild` resumes, the pre-check finds `buildkitd` already running and fails.

This is containerd-specific since Docker runtime uses `docker build` and CRI-O uses `podman build`, neither of which starts `buildkitd`.

The pre-check was added in PR #12081 to verify socket activation (buildkitd starts on-demand, not at boot), which is still the case, but the check is incompatible with other tests that trigger image builds.

test/integration/functional_test.go
monkey-mas: That's amazing! :rocket: Your feedback is much appreciated :pray: I've refactored the helper methods and test as a whole (https://github.com/kubernetes/minikube/pull/23034/changes/94a4f4397ccad4a5bece52ffe3e5c2ab86d362f5) - please let me know if I got it right.  (Like I said before, I'll squash all the integration commits into one once tests passe)

test/integration/functional_test.go
monkey-mas: Done: https://github.com/kubernetes/minikube/pull/23034/changes/94a4f4397ccad4a5bece52ffe3e5c2ab86d362f5

test/integration/testdata/image-load-overwrite/overwrite-test-pod.yaml
monkey-mas: quick question: I kept the labels in the pod YAML because the test uses `PodWait()` which requires a label selector to find pods. Should we switch to `kubectl wait pod/<name>` instead to drop the labels? One concern might be `PodWait` provides richer debug output on failure (pod status logs, stability checks) which `kubectl wait` doesn't. What do you think?

nirs:
> > @monkey-mas: The clean way to fix this is to add another commit before the new test removing this check.
> 
> @nirs Thanks for confirming. I've added a commit to remove the pre-check before the test commit. ([f51f5c9](https://github.com/kubernetes/minikube/commit/f51f5c98efd4c9559933f66345abb26cacdb3c79))
> 
> For context & self-note: as you might know, on containerd runtime, `minikube image build` uses `buildctl` which triggers systemd socket activation to start `buildkitd`. Once started, `buildkitd` remains running for the lifetime of the node. Since `ImageBuild` uses `MaybeParallel(t)`, it gets paused until all sequential subtests complete. If any sequential subtest(eg `ImageLoadInUse`) calls `minikube image build` before `ImageBuild` resumes, the pre-check finds `buildkitd` already running and fails.
> 
> This is containerd-specific since Docker runtime uses `docker build` and CRI-O uses `podman build`, neither of which starts `buildkitd`.
> 
> The pre-check was added in PR #12081 to verify socket activation (buildkitd starts on-demand, not at boot), which is still the case, but the check is incompatible with other tests that trigger image builds.

Thanks for checking this. I think this is implementation details that we should not test. The tests should verify minikube functionality, not checking that we have specific performance optimizations. The optimization is likey related to containerd configuration.

If you think that it is important to test this we can open an issue to add another test verifying the behavior. This is tricky since it depends on all previous tests with the same cluster, so it must be part of a test starting a new cluster.

nirs:
@monkey-mas it is hard to review when I need to consider changes from the original test. Please squash the commit for adding a new test to single commit, unless there is a reason to split it.

The helpers should be simple functions (like listImage()) so we can reuse them later in other tests, not nested function inside the validateImageCommands test.

test/integration/testdata/image-load-overwrite/overwrite-test-pod.yaml
nirs: Keeping the labels is fine, I was not aware that they are used by PodWait().

test/integration/functional_test.go
nirs: Loading from podman works but not in scope of this PR. We can open a new issue for using podman in the tests to ensure we can test loading and saving images to/trom podman everywhere.

Testing docker is good idea since docker engine is open source and available on Linux, but it cannot be the only solution since docker desktop is non-free. Supporting docker desktop is nice to have feature.

monkey-mas:
> it is hard to review when I need to consider changes from the original test. Please squash the commit for adding a new test to single commit, unless there is a reason to split it.

@nirs So sorry about it :facepalm:

Now the commits are in the following order:
- drop buildkitd pre-check: 4a9dea9d9
- test commit: 7af8d94b5
- drop removal image method: 3c95fa57f

> The helpers should be simple functions (like listImage()) so we can reuse them later in other tests, not nested function inside the validateImageCommands test.

Got it - they are included in the test commit (7af8d94b5).

test/integration/functional_test.go
```diff
@@ -380,6 +380,37 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+			t.Fatalf("expected 'version1' but got %q", got)
+		}
+
+		// Load image overwriting image in use
+		loadImageFromTarfile(ctx, t, profile, v2TarPath)
+		startPod(ctx, t, profile, filepath.Join(testDir, "image-load-in-use-v2.yaml"), "image-load-in-use-v2")
+		if got := readPodFile(ctx, t, profile, "image-load-in-use-v2", "/version.txt"); got != "version2" {
+			t.Fatalf("expected 'version2' but got %q - image was not overwritten while in use", got)
+		}
+	})
```
nirs, line 419: So much better now, you can understand the flow in 10 seconds and all the uninteresting details are hidden in the helpers.

test/integration/functional_test.go
```diff
@@ -380,6 +380,37 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+	// docs: Make sure image loading overwrites existing image even when a container is using it (#23023)
+	t.Run("ImageLoadInUse", func(t *testing.T) {
+		// Loads v1 image, starts a pod with it, then loads v2 with the same
+		// tag and verifies that a new pod picks up v2.
+		testImage := "image-load-in-use:test"
+		testDir := filepath.Join(*testdataDir, "image-load-in-use")
+
+		// Build images and store tar files
+		buildImage(ctx, t, profile, filepath.Join(testDir, "v1"), testImage)
+		v1TarPath := saveImageToTarfile(ctx, t, profile, testImage, "image-load-in-use-v1.tar")
```
nirs: If saving the image fails, we will leave the built image in the cluster. We need to clean up after we build.

Where the tar is stored? we probably need a temporary directory for this test, and pass a path inside the temporary directory.

test/integration/functional_test.go
```diff
@@ -380,6 +380,37 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+		removeImage(ctx, t, profile, testImage)
+		buildImage(ctx, t, profile, filepath.Join(testDir, "v2"), testImage)
```
nirs, line 401: Since this build the same image we don't need to clean up after this.

test/integration/functional_test.go
```diff
@@ -380,6 +380,37 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+		v2TarPath := saveImageToTarfile(ctx, t, profile, testImage, "image-load-in-use-v2.tar")
+		removeImage(ctx, t, profile, testImage)
+
+		// Load image and start a pod using it
+		loadImageFromTarfile(ctx, t, profile, v1TarPath)
+		t.Cleanup(func() { removeImage(ctx, t, profile, testImage) })
```
nirs: This cleanup can move after there first build.

test/integration/functional_test.go
```diff
@@ -461,6 +492,89 @@ func listImages(ctx context.Context, t *testing.T, profile string) (*RunResult,
...
+		t.Fatalf("failed to build image %s (%s): %v\n%s", imageName, buildDir, err, rr.Output())
+	}
+}
+
+func removeImage(ctx context.Context, t *testing.T, profile, imageName string) {
+	t.Helper()
+	rr, err := Run(t, exec.CommandContext(ctx,
+		Target(), "-p", profile,
+		"image", "rm", imageName))
+	if err != nil {
```
nirs: This is short enough for single line but keeping the commands consistent is fine.

But now that we have a small helper we can format the code in more clear way:

```go
cmd = exec.CommandContext(ctx, Target(), "-p", profile, "image", "rm", imageName))
if rr, err := Run(t, cmd); err != nil {
    t.Fatalf(...)
}
```

This can be applied to all helpers.

test/integration/functional_test.go
```diff
@@ -461,6 +492,89 @@ func listImages(ctx context.Context, t *testing.T, profile string) (*RunResult,
...
+		t.Fatalf("failed to remove image %s: %v\n%s", imageName, err, rr.Output())
+	}
+}
+
+func saveImageToTarfile(ctx context.Context, t *testing.T, profile, imageName, tarFile string) string {
+	t.Helper()
+	tarPath, err := filepath.Abs(tarFile)
```
nirs: This uses the current directory - leaving leftover test files around. We need to get a path in the a test temporary directory from the user and we don't need to use filepath.Abs here.

test/integration/functional_test.go
```diff
@@ -461,6 +492,89 @@ func listImages(ctx context.Context, t *testing.T, profile string) (*RunResult,
...
+	if err != nil {
+		t.Fatalf("failed to get absolute path for %s: %v", tarFile, err)
+	}
+	t.Cleanup(func() { os.Remove(tarPath) })
```
nirs: Cleanup up automatically is nice but it is not consistent - in other helper we don't cleanup automatically, and and it not needed in all cases (e.g. helpers using the same test image).

If the caller pass a file in a temporary directory the caller is responsible for the cleanup.

test/integration/functional_test.go
```diff
@@ -461,6 +492,89 @@ func listImages(ctx context.Context, t *testing.T, profile string) (*RunResult,
...
+	rr, err := Run(t, exec.CommandContext(ctx,
+		Target(), "-p", profile,
+		"image", "save", imageName, tarPath,
+		"--alsologtostderr"))
+	if err != nil {
+		t.Fatalf("failed to save image %s to %s: %v\n%s", imageName, tarFile, err, rr.Output())
+	}
+	return tarPath
```
nirs: The caller should own this path, we don't need to return it here.

test/integration/functional_test.go
```diff
@@ -461,6 +492,89 @@ func listImages(ctx context.Context, t *testing.T, profile string) (*RunResult,
...
+		"apply", "-f", podYAMLPath))
+	if err != nil {
+		t.Fatalf("failed to create pod %s: %v\n%s", podName, err, rr.Output())
+	}
+	t.Cleanup(func() {
+		Run(t, exec.CommandContext(ctx,
+			Target(), "-p", profile,
+			"kubectl", "--",
+			"delete", "pod", podName, "--ignore-not-found"))
+	})
```
nirs, line 553: Cleaning up here is good, it is makes it easy to use correctly, but it is not consistent with other helpers that should not clean up automatically. However since we want to wait for the pod here, cleaning up in the caller is not possible (this may fail after creating the pod), so lets keep this cleanup.

test/integration/functional_test.go
```diff
@@ -461,6 +492,89 @@ func listImages(ctx context.Context, t *testing.T, profile string) (*RunResult,
...
+func readPodFile(ctx context.Context, t *testing.T, profile, podName, filePath string) string {
+	t.Helper()
+	rr, err := Run(t, exec.CommandContext(ctx,
+		Target(), "-p", profile,
+		"kubectl", "--",
+		"exec", podName, "--", "cat", filePath))
+	if err != nil {
+		t.Fatalf("failed to read %s from pod %s: %v\n%s", filePath, podName, err, rr.Output())
+	}
+	return strings.TrimSpace(rr.Stdout.String())
```
nirs: trimming space here is niece for our test, but may make it hard for other tests. We should return the exact string we read here. The caller can trim space if they need.

test/integration/functional_test.go
```diff
@@ -461,6 +492,89 @@ func listImages(ctx context.Context, t *testing.T, profile string) (*RunResult,
...
+	rr, err := Run(t, exec.CommandContext(ctx,
+		Target(), "-p", profile,
+		"image", "load", tarPath,
+		"--alsologtostderr"))
+	if err != nil {
+		t.Fatalf("failed to load image from %s: %v\n%s", tarPath, err, rr.Output())
+	}
+}
+
+func startPod(ctx context.Context, t *testing.T, profile, podYAMLPath, podName string) {
```
nirs: Since this start a pod and clean up, maybe we can rename the function to reflect the auto cleanup? `startTempPod()`?

test/integration/functional_test.go
```diff
@@ -384,6 +380,44 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+		// loadImageFromTarfile — t.Cleanup handles the final removal.
+		buildImage(ctx, t, profile, filepath.Join(testDir, "v1"), testImage)
+		t.Cleanup(func() { removeImage(ctx, t, profile, testImage) })
+		v1TarPath := filepath.Join(tmpDir, "image-load-in-use-v1.tar")
+		saveImageToTarfile(ctx, t, profile, testImage, v1TarPath)
+		removeImage(ctx, t, profile, testImage)
+		buildImage(ctx, t, profile, filepath.Join(testDir, "v2"), testImage)
+		v2TarPath := filepath.Join(tmpDir, "image-load-in-use-v2.tar")
+		saveImageToTarfile(ctx, t, profile, testImage, v2TarPath)
+		removeImage(ctx, t, profile, testImage)
```
monkey-mas, line 404: @nirs One thing I wanted to confirm about the cleanup behavior:

In the test flow, `t.Cleanup(func() { removeImage(...) })` is registered right after the first `buildImage()` as a safety net. In the normal case, the image will be loaded at test end, so `removeImage` in cleanup works fine.

However, if the test fails midway (e.g., after an explicit `removeImage()` call but before `loadImageFromTarfile()`), the cleanup would try to remove an image that doesn't exist, and `removeImage` would call `t.Fatalf`. I'm not sure if `t.Fatalf` inside `t.Cleanup` could interrupt other cleanup functions.

A few options:

1. Keep `removeImage` as-is with `t.Fatalf`: the cleanup error only happens when the test already failed, so it's practically harmless.
2. Add a best-effort variant for cleanup use:
   ```go
     func removeImageBestEffort(ctx context.Context, t *testing.T, profile, imageName string) {
         // same as removeImage but uses t.Logf instead of t.Fatalf
   }
   ```
3. Make removeImage accept a log function so it can be used for both:
e.g.,
  - `removeImage(ctx, t, profile, testImage, t.Fatalf)` for normal useage
  - `t.Cleanup(func() { removeImage(ctx, t, profile, testImage, t.Logf) })` for cleanup

What do you think? :eyes:

monkey-mas:
@nir Thanks for your review! I  applied all the feedback from you. Had one question about `removeImage` logging - PTAL when you get a chance :pray:
https://github.com/kubernetes/minikube/pull/23034/changes#r3312853651

test/integration/functional_test.go
nirs: removing image during cleanup should be idempotent - it should treat missing image as expected condition. Logging debug message like "image was already removed" is nice to make debugging easier, but logging errors and warnings is not.

nirs: If we want to use `removeImage()` for validation and fail on missing image, we can add a variant for cleanup like `silentRemoveImage()` that does not fail on missing image.

test/integration/functional_test.go
```diff
@@ -380,6 +380,44 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+		testDir := filepath.Join(*testdataDir, "image-load-in-use")
+		tmpDir := t.TempDir()
+
+		// Build images and store tar files.
+		// t.Cleanup ensures the image is removed from the cluster even if
+		// saveImageToTarfile or subsequent steps fail. In the normal case,
+		// the image is explicitly removed after saving, then reloaded via
+		// loadImageFromTarfile — t.Cleanup handles the final removal.
+		buildImage(ctx, t, profile, filepath.Join(testDir, "v1"), testImage)
+		t.Cleanup(func() { removeImage(ctx, t, profile, testImage) })
```
nirs: What happens when this run during cleanup an fails since the image is not found?

I think we need to call a silentRemoveImage() here that does not use *testing.T and does not fail.

nirs (commented):
Looks, good!

minor issue: use silent cleanup function for the image.

test/integration/functional_test.go
monkey-mas: > I think we need to call a silentRemoveImage() here that does not use *testing.T and does not fail.

Yes, I was planning to use `silentRemoveImage()` for `t.Cleanup`. Just to clarify - I thought `silentRemoveImage` still takes `*testing.T` for `t.Logf` debug output like below. Should it avoid `*testing.T` entirely (e.g., use `fmt.Printf` or `klog` instead)?

> that does not use *testing.T and does not fail.

Or did you mean that we should avoid using `t.Fatalf` like we discussed?

[`silentRemoveImage()` impl in my mind:]
```go
// silentRemoveImage is an idempotent variant of removeImage for use in
// t.Cleanup. It treats a missing image as an expected condition.
func silentRemoveImage(ctx context.Context, t *testing.T, profile, imageName string) {
       t.Helper()
       cmd := exec.CommandContext(ctx, Target(), "-p", profile, "image", "rm", imageName)
       if rr, err := Run(t, cmd); err != nil {
               t.Logf("image %s was already removed or not found: %v\n%s", imageName, err, rr.Output())
       }
}
```

nirs: If we always use t.Logf() for logging, Wes should take *testing.T and use it.

monkey-mas: Sounds great! Thank you for confirming :)

test/integration/functional_test.go
```diff
@@ -461,6 +499,73 @@ func listImages(ctx context.Context, t *testing.T, profile string) (*RunResult,
...
+	}
+}
+
+// silentRemoveImage is an idempotent variant of removeImage for use in
+// t.Cleanup. It treats a missing image as an expected condition.
+func silentRemoveImage(ctx context.Context, t *testing.T, profile, imageName string) {
+	t.Helper()
+	cmd := exec.CommandContext(ctx, Target(), "-p", profile, "image", "rm", imageName)
+	if rr, err := Run(t, cmd); err != nil {
+		t.Logf("image %s was already removed or not found: %v\n%s", imageName, err, rr.Output())
```
nirs: This log is correct only if we find that the image does not exist, but `minikube image rm` can fail for many reasons and it does not provide a way to get the specific error. Since we already log errors in Run(), I think this version can simply call Run() and ignore the result.

To do real silent remove we need to modify the image rm command to return specific error code or json response exposing the underlying error. Since we ton't have have this, we cannot do a real silent remove, and we may see failures log when the image does not exist.

nirs:
@monkey-mas please also rebase on master, we fixed some tests that always failed on Prow. Rebasing will allow automatic merge when this is approved.

test/integration/functional_test.go
monkey-mas: Got it, thanks for the feedback. I've updated the method so that it just executes `Run(...)` :+1:

monkey-mas:
> please also rebase on master, we fixed some tests that always failed on Prow. Rebasing will allow automatic merge when this is approved.

@nirs Yup, just rebased the latest commits :+1: 

Looks like the `ImageLoadInUse` test fails on arm64 containerd (pull-minikube-docker-containerd-linux-arm) with `ErrImageNeverPull`, while containerd-linux-x86 and other jobs pass.
- [Failing job (arm64 containerd)][1]

Unfortunately, I was unable to reproduce this locally since I'm on amd64. That said, I'm adding some debug logging after `image load` to gather more information(https://github.com/kubernetes/minikube/pull/23034/commits/89cddd8ae120c963701ff056acd7d3aad81ddb45):
1. Whether the image is visible via `minikube image ls`                                                                                                                   
2. `ctr -n=k8s.io images ls` output for platform info (containerd only)
3. The tar manifest/index for platform metadata

Please let me know if there's a better way to debug this on arm64, or any known (flaky) issues.

[1]: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/23034/pull-minikube-docker-containerd-linux-arm/2059709065012973568

test/integration/testdata/image-load-in-use/image-load-in-use-v1.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: image-load-in-use-v1
+  labels:
+    run: image-load-in-use-v1
+spec:
+  containers:
+  - name: test
+    image: image-load-in-use:test
```
nirs: try the full name `localhost/image-load-in-use:test`

test/integration/testdata/image-load-in-use/image-load-in-use-v2.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: image-load-in-use-v2
+  labels:
+    run: image-load-in-use-v2
+spec:
+  containers:
+  - name: test
+    image: image-load-in-use:test
```
nirs: full name

test/integration/functional_test.go
```diff
@@ -380,6 +380,44 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
 		checkImageExists(ctx, t, profile, taggedImage)
 	})
 
+	// docs: Make sure image loading overwrites existing image even when a container is using it (#23023)
+	t.Run("ImageLoadInUse", func(t *testing.T) {
+		// Loads v1 image, starts a pod with it, then loads v2 with the same
+		// tag and verifies that a new pod picks up v2.
+		testImage := "image-load-in-use:test"
```
nirs: use full name: `localhost/image-load-in-use:test`

nirs (commented):
The failures may be related to using image without the registry. Maybe it is handled differently in containerd on arm?

nirs:
@monkey-mas check my last review

monkey-mas:
@nirs Thanks for the suggestion! I've updated the image name to use `localhost/image-load-in-use:test` in all three places (test code + both pod YAMLs), removed the debug logging, and rebased on latest master.

Regarding the arm64 containerd failure, it looks like it was flaky. The [previous CI run][1] on containerd-linux-arm already passed, and I don't see `ImageBuild` or `ImageLoadInUse` failing in any of the current CI environments either. That said, using the full name with `localhost/` prefix should make it more robust across runtimes. If the arm64 issue recurs, I can add the debug logging back.

[1]: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/23034/pull-minikube-docker-containerd-linux-arm/2060088070832656384

nirs (approved):
@monkey-mas Thanks for this hard work!

nirs:
Functional tests pass now: https://github.com/kubernetes/minikube/actions/runs/26652633253?pr=23034

/lgtm

test/integration/functional_test.go
```diff
@@ -384,6 +380,44 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
 		checkImageExists(ctx, t, profile, taggedImage)
 	})
 
+	// docs: Make sure image loading overwrites existing image even when a container is using it (#23023)
+	t.Run("ImageLoadInUse", func(t *testing.T) {
+		// Loads v1 image, starts a pod with it, then loads v2 with the same
```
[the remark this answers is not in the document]
nirs: The test name does not have to match the pr description. The PR description is stale, and the test name is good and does not need to change.

test/integration/functional_test.go
```diff
@@ -384,6 +380,44 @@ func validateImageCommands(ctx context.Context, t *testing.T, profile string) {
...
+		startTempPod(ctx, t, profile, filepath.Join(testDir, "image-load-in-use-v1.yaml"), "image-load-in-use-v1")
+		if got := readPodFile(ctx, t, profile, "image-load-in-use-v1", "/version.txt"); strings.TrimSpace(got) != "version1" {
+			t.Fatalf("expected 'version1' but got %q", got)
+		}
+
+		// Load image overwriting image in use
+		loadImageFromTarfile(ctx, t, profile, v2TarPath)
+		startTempPod(ctx, t, profile, filepath.Join(testDir, "image-load-in-use-v2.yaml"), "image-load-in-use-v2")
+		if got := readPodFile(ctx, t, profile, "image-load-in-use-v2", "/version.txt"); strings.TrimSpace(got) != "version2" {
+			t.Fatalf("expected 'version2' but got %q - image was not overwritten while in use", got)
```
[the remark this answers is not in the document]
nirs: We cannot test with --daemon since it depends on docker which is not available on some platforms. We can add another test that will be skipped if docker is not available.

Issue #23018, opened 2026-05-22
Proposal: Simple preloads


nirs:
## Problem with current preloads

Current preloads are runtime-specific snapshots of `/var`, fragile, expensive
to build, and require 656 assets (221.5 GiB) on GitHub.
See [#23016](https://github.com/kubernetes/minikube/issues/23016) for full details.

## Goals

- **Robust**: cannot break when a container runtime changes its internal storage format
- **Easy to build**: no KIC container, no Docker daemon — just pull images and package
- **Easy to maintain**: trivial code, minimal moving parts
- **Low storage and bandwidth**: smaller tarballs, fewer assets
- **Minimal builds**: build once per k8s version per arch, never rebuild
- **Fast enough**: acceptable cluster start time compared to current preloads
- **Easy to understand**: accessible to new contributors and maintainers
- **Community accessible**: all infrastructure on GitHub, no Google-only dependencies

## Simple preloads design

Package OCI image archives and k8s binaries into a single archive. Load them via
the runtime's standard import API at cluster creation time. No Docker daemon
needed to build.

### What's in the tarball

Name: `minikube-preload-<arch>-<k8s-version>-<release>.tar.zst`

- `<arch>`: target architecture (`amd64`, `arm64`)
- `<k8s-version>`: kubernetes version (`1.36.1`)
- `<release>`: preload release, starting at `1`. Bumped if the tarball needs
  to be rebuilt for the same k8s version (e.g. corrupt build, auxiliary image
  update)

Example: `minikube-preload-arm64-1.36.1-1.tar.zst`

```
bin/
├── kubeadm
├── kubelet
└── kubectl
images/
├── registry.k8s.io/
│   ├── kube-apiserver_v1.36.1
│   ├── kube-controller-manager_v1.36.1
│   ├── kube-scheduler_v1.36.1
│   ├── kube-proxy_v1.36.1
│   ├── etcd_3.6.8-0
│   ├── coredns/
│   │   └── coredns_v1.14.2
│   └── pause_3.10.2
└── gcr.io/
    └── k8s-minikube/
        └── storage-provisioner_v5
```

### Key properties

- **One tarball per (k8s version × arch)**: no runtime or filesystem dimension
  (overlay2, overlay, etc.) → ~3x fewer assets
- **Runtime-agnostic**: OCI tarballs are a standard format, loaded via the runtime's
  own import command — never touches internal storage directories
- **Robust**: can never break due to runtime storage layout changes
[...]

PR #23036, opened 2026-05-23
containerd: skip TestPreload user-image check

### LINKED ISSUE #23035: containerd: Skip TestPreload/Restart-With-Preload-Check-User-Image
The test fails in Prow but for some reason does not fail in Jenkins.

The failure is expected since we don't have mechanism for preserving user images when loading preloads for containerd.

Until we add this or simplify preloads so they don't overwrite user images (#23018) we need to skip the test.

High priority since this break required tests, preventing merge of other PRs. Currently merge requires admin rights because of this issue.

Example failure:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/23031/pull-minikube-docker-containerd-linux-x86/2058215384078618624

Related issues:
- #22269

Plan:
- Skip the test with a message linking to this issue

thc1006:
TestPreload/Restart-With-Preload-Check-User-Image has been failing on the required pull-minikube-docker-containerd-linux-x86 job, which blocks other PRs from merging without an admin override.

The test pulls a user image, restarts with --preload=true, and checks the image is still there. On containerd the preload is a filesystem dump that gets untarred over /var, and that overwrites the runtime metadata, taking the user's image with it. docker and cri-o save and restore images around that step, but containerd doesn't do that yet (#22269), so the check can't pass there.

For now this just skips the check when the runtime is containerd, with a message linking back to this issue. docker and cri-o are untouched and keep running it.

Tested locally with the docker driver: on containerd the subtest skips (the setup still runs and passes), and on docker it runs and passes.

The real fix (backup/restore on containerd) is #22269, and #23018 would remove the /var overwrite altogether.

Fixes #23035

### DIFF
--- a/test/integration/preload_test.go
+++ b/test/integration/preload_test.go
@@ -24,6 +24,8 @@ import (
 	"os/exec"
 	"strings"
 	"testing"
+
+	"k8s.io/minikube/pkg/minikube/constants"
 )
 
 // TestPreload verifies that disabling the initial preload, pulling a specific image,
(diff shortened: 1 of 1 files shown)


nirs (approved):
@thc1006 Thanks! excellent commit message and the test is skipped correctly with useful link.

nirs:
@thc1006 the other required tests fail for other unrelated reasons. 

This change fixed the following builds:
- required: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/23036/pull-minikube-docker-containerd-linux-x86/2058528978062282752
- optional: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/minikube/23036/pull-minikube-docker-containerd-linux-arm/2058528977642852352

/lgtm

nirs:
@medyagh ready to merge, but unrelated tests failure block the merge.

/assign @medyagh

PR #23037, opened 2026-05-23
pkg/minikube/runtimedir: move per-machine sockets out of MINIKUBE_HOME

### LINKED ISSUE #21087: VM driver wont work if user path is too long

rafaelmfried:
### Summary

AF_UNIX sockets are limited to `sun_path[108]` on Linux and `sun_path[104]` on macOS/BSD. minikube currently stores all per-machine sockets under `$MINIKUBE_HOME/.minikube/machines/<name>/`, which can exceed the limit when the home path, username, machine name or socket name are long. This is the root cause reported in #21087 for the QEMU `qmp` socket, but as @nirs pointed out in the issue thread the same trap exists for vfkit, krunkit and vmnet-helper.

This PR introduces a small `pkg/minikube/runtimedir/` helper that resolves per-machine socket paths to a short, platform-specific runtime directory, and migrates every driver socket to it. Non-socket per-machine state (disk images, logs, configs, pidfiles) stays under `MINIKUBE_HOME`.

Fixes #21087

> **Draft.** Opening this as a draft to surface the design decisions and open points below before polishing. Happy to take direction on any of them — and #22960 lands first.

### Layout

```
Linux:    $XDG_RUNTIME_DIR/minikube/<hash>/<sock-name>
          fallback /tmp/<uid>/minikube/<hash>/<sock-name>   (when XDG is unset or missing)
macOS:    /tmp/<uid>/minikube/<hash>/<sock-name>
Windows:  <os.TempDir()>/minikube/<hash>/<sock-name>        (qemu Experimental on Windows)
```

`<hash>` is the first 32 hex characters of `sha256(machineName)` (128 bits — collision-free across realistic profile names).

Per @nirs's budget calculation in the issue:

|       | base              | uid      | `minikube/` | hash | overhead | sun_path budget for sock name |
| ----- | ----------------- | -------- | ----------- | ---- | -------: | ----------------------------: |
| Linux | `/run/user/` (10) | up to 11 | 9           | 33   |       63 |             108 − 63 = **45** |
| macOS | `/tmp/` (5)       | up to 11 | 9           | 33   |       58 |             104 − 58 = **46** |

The longest socket name currently used in the tree is `vmnet-helper.sock-krun.sock` (27 chars), so the budget is comfortable.

### What changed

- New package `pkg/minikube/runtimedir/` exposing two entry points so callers can opt in to filesystem side effects:
  - `SocketPath(machineName, sockName string) (string, error)` — pure resolver, no filesystem writes. Used by callers that dial a socket the server already bound (state polls).
  - `EnsureSocketPath(machineName, sockName string) (string, error)` — resolves the path and creates the parent directory with mode `0700`. Used by callers that bind a new socket.
  - `Cleanup(machineName string) error` — idempotent removal of the per-machine runtime directory.
- Migrated socket call sites. Each driver exposes a `*Path` / `ensure*Path` pair so binding paths use the side-effecting variant and read paths stay pure:
  - `pkg/drivers/qemu` — QMP `monitor` socket. `Start` uses Ensure; `RunQMPCommand` uses the pure variant. Best-effort cleanup of any stale socket before bind.
  - `pkg/drivers/vfkit` — `vfkit.sock`. `startVfkit` uses Ensure; `GetVFKitState` / `SetVFKitState` stay pure. Existing best-effort stale-socket removal preserved.
  - `pkg/drivers/krunkit` — `krunkit.sock`. `restfulURI` uses Ensure; `setKrunkitState` stays pure.
  - `pkg/drivers/common/vmnet` — `vmnet-helper.sock`. `Helper.EnsureSocketPath` resolves the path and creates the dir (called from driver Start before the helper binds). The helper's pidfile and logfile remain in `MachineDir` (regular files, no `sun_path` limit).
- `vmnet.Helper` gains a `MachineName` field so the runtime path can be resolved without reaching into `MachineDir`'s filesystem layout. Populated at the two construction sites: `pkg/minikube/registry/drvs/vfkit` and `pkg/minikube/registry/drvs/krunkit`.
- All three drivers (`qemu`, `vfkit`, `krunkit`) now call `runtimedir.Cleanup` from `Remove()` so the directory is reclaimed on `minikube delete`.

### Tests

Table-driven coverage in `pkg/minikube/runtimedir/runtimedir_test.go`:

- `baseDir` picks the correct branch on darwin/linux/windows, including `$XDG_RUNTIME_DIR` present, empty, and pointing to a missing directory.
- `SocketPath` (pure) is deterministic across calls and asserts no filesystem side effect.
- `EnsureSocketPath` creates the parent dir with mode `0700` on the linux xdg-present and windows branches; the /tmp-rooted darwin/linux-fallback branches are exercised with a synthetic high uid and cleaned up afterwards so test runs don't pollute `/tmp/<real-uid>`.
- Both entry points reject empty machine or socket names.
- `Cleanup` removes the per-machine directory and is idempotent.
- `machineHash` has the expected length and avoids collisions across realistic profile names.
- `TestPathBudget` asserts the resolved path fits within `sun_path[108]` on Linux and `sun_path[104]` on macOS even for the worst-case machine name (64 chars) + socket name (`vmnet-helper.sock-krun.sock`, 27 chars).

Cross-compile checks pass for `linux/amd64`, `darwin/arm64`, `darwin/amd64`, and `windows/amd64` (qemu is supported on Windows with Experimental priority).

### Design notes / open points

A few decisions are surfaced here so reviewers can redirect early:

1. **Two entry points (pure vs ensure).** `SocketPath` and `EnsureSocketPath` differ only in whether the parent directory is materialised. The pure variant exists because state-poll callers (`GetVFKitState`, `RunQMPCommand`, `setKrunkitState`) should not lazily create directories on the filesystem for a machine that may not even be running. Drivers wrap the package functions in thin `*Path` / `ensure*Path` helpers for symmetry. The internal `resolveSocketPath(... ensure bool)` keeps both paths on one implementation.
2. **Single PR vs staged.** The helper is small but only meaningful with all four sockets migrated; splitting would force re-reviewing the same design 2–4 times. Commits are organized one per driver, so reverting a single migration if needed is trivial. Happy to stage if preferred.
3. **`MachineName` field on `vmnet.Helper`.** The alternative was deriving the machine name from `filepath.Base(MachineDir)`, which couples to a path convention. The explicit field is ~5 LOC across the two construction sites and easier to reason about.
4. **Permissions.** Runtime directories are created `0700`. When vmnet-helper runs under `sudo` (`NeedsSudo == true`), the socket is bound by root inside a user-owned `0700` directory; the helper receives the path via `--socket` from the caller (uid=user), so client and server agree on the location. This ownership/permission relationship is unchanged by this PR — vmnet-helper already bound the socket (as root) inside a user-owned `0700` directory under `MachineDir`; only the parent path is now shorter. Worth confirming on your end whether the helper already `chmod`s the socket post-bind; if not, we may need a small change in the external `/opt/vmnet-helper/bin/vmnet-helper` repo too. Happy to follow up there separately.
5. **Hash determinism.** `sha256(machineName)` only — `uid` is already in the path and `profile` is already baked into `machineName` in minikube's naming. I can include either if preferred.
6. **No build tags on the helper package.** The `runtimedir` package is cross-platform; the platform switch is done at runtime via `runtime.GOOS`. This keeps a single, easily testable implementation, and includes a Windows branch since qemu is supported there with Experimental priority.
7. **Stale socket on the old path.** After upgrading to a binary with this change, any leftover socket under `~/.minikube/machines/<name>/` becomes orphaned. It does not block the new socket and is removed by `minikube delete`, so I left an explicit one-off cleanup out of scope. Happy to add it if you'd rather sweep on `Start`.

### How to reproduce / validate locally

```bash
# Force a long MINIKUBE_HOME and observe start succeeding with this change.
export MINIKUBE_HOME=/Users/$(whoami)/.really-long-path-to-force-the-bug-aaaaaaaaaaaaaa
minikube start --driver=qemu2 -p test-long-path-bbbbbbbbbbb
# The QMP socket now lives at /tmp/<uid>/minikube/<hash>/monitor
```

Without the change, `minikube start` fails with "UNIX socket path too long" on macOS.

### Empirical validation

Reproduced end-to-end on macOS 15.5 (arm64) with `qemu-system-aarch64` 11.0.0. Without the fix, `minikube start` fails on a long `MINIKUBE_HOME` with the verbatim `UNIX socket path ... is too long` error from #21087. With the fix, the same path and profile name succeed: the QMP socket binds under the runtime directory and the rest of the per-machine state stays in `MINIKUBE_HOME`. The full `stop`/`start`/`delete` lifecycle, the default `$HOME/.minikube` path, and a sha256 check of the resolved directory name all pass — no regression. This run exercised the qemu2 driver end-to-end; the vfkit, krunkit and vmnet-helper migrations apply the identical `runtimedir` call-site pattern and were not separately e2e-tested locally.

Full repro logs, command output and a consolidated `SUMMARY.md` are attached as a separate comment, so the diff and this description stay readable.

A standalone, containerized **proof of concept** — [`unix-socket-path-limit`](https://github.com/rafaelmfried/unix-socket-path-limit) — runs the real `qemu` binary in Testcontainers to reproduce the bug and prove the fix deterministically, independent of the host's `$MINIKUBE_HOME` length.

### Commits

A single DCO-signed commit, per the squash-before-merge norm signalled on
the prior PR (#22960).

### DIFF
--- a/pkg/drivers/common/vmnet/vmnet.go
+++ b/pkg/drivers/common/vmnet/vmnet.go
@@ -60,6 +60,10 @@ type Helper struct {
 	// The pidfile and log are stored here.
 	MachineDir string
 
+	// The socket is stored here. The caller creates the directory before
+	// starting the helper. See pkg/minikube/runtimedir.
+	MachineSocketDir string
+
 	// InterfaceID is an optional UUID for the vmnet interface. When set,
 	// vmnet-helper passes --interface-id to obtain a deterministic MAC address
 	// from vmnet.
@@ -339,7 +343,7 @@ func (h *Helper) GetState() (state.State, error) {
 }
 
 func (h *Helper) SocketPath() string {
-	return filepath.Join(h.MachineDir, sockfileName)
+	return filepath.Join(h.MachineSocketDir, sockfileName)
 }
 
 func (h *Helper) openLogfile() (*os.File, error) {
--- a/pkg/drivers/krunkit/krunkit.go
+++ b/pkg/drivers/krunkit/krunkit.go
@@ -51,6 +51,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/process"
 	"k8s.io/minikube/pkg/minikube/reason"
 	"k8s.io/minikube/pkg/minikube/run"
+	"k8s.io/minikube/pkg/minikube/runtimedir"
 	"k8s.io/minikube/pkg/minikube/style"
 )
 
@@ -197,6 +198,11 @@ func (d *Driver) Create() error {
 }
 
 func (d *Driver) Start() error {
+	socketDir, err := d.SocketDir()
+	if err != nil {
+		return err
+	}
+	d.VmnetHelper.MachineSocketDir = socketDir
 	socketPath := d.VmnetHelper.SocketPath()
 	if err := d.VmnetHelper.Start(socketPath); err != nil {
 		return err
@@ -226,10 +232,14 @@ func (d *Driver) Start() error {
 
 // startKrunkit starts the krunkit child process.
 func (d *Driver) startKrunkit(socketPath string) error {
+	restURI, err := d.restfulURI()
+	if err != nil {
+		return err
+	}
 	var args = []string{
 		"--memory", fmt.Sprintf("%d", d.Memory),
 		"--cpus", fmt.Sprintf("%d", d.CPU),
-		"--restful-uri", d.restfulURI(),
+		"--restful-uri", restURI,
 		"--device", fmt.Sprintf("virtio-net,type=unixgram,path=%s,mac=%s,offloading=%t",
 			socketPath, d.MACAddress, d.VmnetHelper.Offloading),
 		"--device", fmt.Sprintf("virtio-serial,logFilePath=%s", d.serialPath()),
@@ -319,6 +329,9 @@ func (d *Driver) Remove() error {
 		if err := d.Kill(); err != nil {
 			return fmt.Errorf("kill: %w", err)
 		}
+	}
+	if err := runtimedir.RemoveSocketDir(d.GetMachineName()); err != nil {
+		log.Debugf("Failed to remove socket dir: %v", err)
 	}
 	return nil
 }
@@ -449,12 +462,16 @@ func (d *Driver) isoPath() string {
 	return d.ResolveStorePath(isoFileName)
 }
 
-func (d *Driver) sockfilePath() string {
-	return d.ResolveStorePath(sockFileName)
+func (d *Driver) sockfilePath() (string, error) {
+	return d.ResolveSocketPath(sockFileName)
 }
 
-func (d *Driver) restfulURI() string {
-	return fmt.Sprintf("unix://%s", d.sockfilePath())
+func (d *Driver) restfulURI() (string, error) {
+	sock, err := d.sockfilePath()
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("unix://%s", sock), nil
 }
 
 func (d *Driver) vmStateURI() string {
@@ -538,7 +555,11 @@ func (d *Driver) setKrunkitState(desired string) error {
 	if err != nil {
 		return err
 	}
-	httpc := httpUnixClient(d.sockfilePath())
+	sock, err := d.sockfilePath()
+	if err != nil {
+		return err
+	}
+	httpc := httpUnixClient(sock)
 	_, err = httpc.Post(d.vmStateURI(), "application/json", bytes.NewReader(data))
 	if err != nil {
 		return err
--- a/pkg/drivers/qemu/qemu.go
+++ b/pkg/drivers/qemu/qemu.go
@@ -49,6 +49,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/out"
 	"k8s.io/minikube/pkg/minikube/reason"
 	"k8s.io/minikube/pkg/minikube/run"
+	"k8s.io/minikube/pkg/minikube/runtimedir"
 	"k8s.io/minikube/pkg/minikube/style"
 	"k8s.io/minikube/pkg/network"
 	"k8s.io/minikube/pkg/util/retry"
@@ -438,8 +439,12 @@ func (d *Driver) Start() error {
 		startCmd = append(startCmd,
 			"-cdrom", isoPath)
 	}
+	monitorPath, err := d.monitorPath()
+	if err != nil {
+		return err
+	}
 	startCmd = append(startCmd,
-		"-qmp", fmt.Sprintf("unix:%s,server,nowait", d.monitorPath()),
+		"-qmp", fmt.Sprintf("unix:%s,server,nowait", monitorPath),
 		"-pidfile", d.pidfilePath(),
 	)
 
@@ -608,6 +613,9 @@ func (d *Driver) Remove() error {
 			return fmt.Errorf("quit: %w", err)
 		}
 	}
+	if err := runtimedir.RemoveSocketDir(d.GetMachineName()); err != nil {
+		log.Debugf("Failed to remove socket dir: %v", err)
+	}
 	return nil
 }
 
@@ -662,9 +670,8 @@ func (d *Driver) diskPath() string {
 	return filepath.Join(machineDir, "disk.qcow2")
 }
 
-func (d *Driver) monitorPath() string {
-	machineDir := filepath.Join(d.StorePath, "machines", d.GetMachineName())
-	return filepath.Join(machineDir, "monitor")
+func (d *Driver) monitorPath() (string, error) {
+	return d.ResolveSocketPath("monitor")
 }
 
 func (d *Driver) pidfilePath() string {
@@ -763,7 +770,11 @@ func (d *Driver) generateUserdataDisk(userdataFile string) (string, error) {
 
 func (d *Driver) RunQMPCommand(command string) (map[string]interface{}, error) {
 	// connect to monitor
-	conn, err := net.Dial("unix", d.monitorPath())
+	monitorPath, err := d.monitorPath()
+	if err != nil {
+		return nil, err
+	}
+	conn, err := net.Dial("unix", monitorPath)
 	if err != nil {
 		return nil, fmt.Errorf("connect: %w", err)
 	}
--- a/pkg/minikube/runtimedir/runtimedir.go
+++ b/pkg/minikube/runtimedir/runtimedir.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2026 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package runtimedir resolves the per-machine directory for AF_UNIX sockets.
+//
+// Sockets created under MINIKUBE_HOME (default $HOME/.minikube/machines/<name>/)
+// can exceed sockaddr_un.sun_path (108 bytes on Linux, 104 on macOS) when the
+// home path, user name or machine name are long. This package keeps only the
+// sockets in a short runtime directory; the rest of the per-machine state
+// (disk images, logs, configs, pidfiles) stays under MINIKUBE_HOME.
+//
+// Layout:
+//
+//	Linux:   $XDG_RUNTIME_DIR/minikube/<machine>
+//	         /tmp/user/<uid>/minikube/<machine> when XDG_RUNTIME_DIR is unset
+//	macOS:   /tmp/user/<uid>/minikube/<machine>
+//	Windows: %TEMP%\minikube\<machine>
+package runtimedir
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strconv"
+
+	"k8s.io/klog/v2"
+)
+
+// dirMode is the mode of every directory created by this package. Sockets
+// bound inside need to be reachable only by the owning user.
+const dirMode = 0o700
+
+// Indirections for testing, so tests can run without touching the environment
+// or the real temporary directory.
+var (
+	// baseDir is the root of the fallback runtime directory.
+	baseDir = func() string { return "/tmp" }
+
+	// xdgRuntimeDir is the user runtime directory on Linux, typically
+	// /run/user/<uid>.
+	xdgRuntimeDir = func() string { return os.Getenv("XDG_RUNTIME_DIR") }
+
+	// tempDir is the per-user temporary directory on Windows, typically
+	// C:\Users\<username>\AppData\Local\Temp.
+	tempDir = os.TempDir
+)
+
+// EnsureSocketDir returns the directory holding the AF_UNIX sockets of the
+// given machine, creating it with mode 0700 if needed.
+func EnsureSocketDir(machineName string) (string, error) {
+	dir, err := resolve(machineName)
+	if err != nil {
+		return "", err
+	}
+	if err := os.MkdirAll(dir, dirMode); err != nil {
+		return "", fmt.Errorf("mkdir %q: %w", dir, err)
+	}
+	// MkdirAll applies the umask, and the directory may already exist with
+	// a more permissive mode.
+	if err := os.Chmod(dir, dirMode); err != nil {
+		return "", fmt.Errorf("chmod %q: %w", dir, err)
+	}
+	return dir, nil
+}
+
+// RemoveSocketDir removes the socket directory of the given machine. It is
+// idempotent: removing a directory that does not exist is not an error.
+func RemoveSocketDir(machineName string) error {
+	dir, err := resolve(machineName)
+	if err != nil {
+		return err
+	}
+	if err := os.RemoveAll(dir); err != nil {
+		return fmt.Errorf("remove %q: %w", dir, err)
+	}
+	return nil
+}
+
+// resolve returns the socket directory of the given machine without touching
+// the filesystem.
+func resolve(machineName string) (string, error) {
+	if machineName == "" {
+		return "", errors.New("machine name is empty")
+	}
+	switch runtime.GOOS {
+	case "windows":
+		// User names are limited to 20 characters, so the per-user temporary
+		// directory is short enough.
+		// https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/net-user
+		return filepath.Join(tempDir(), "minikube", machineName), nil
+	case "linux":
+		xdg := xdgRuntimeDir()
+		if xdg == "" {
+			break
+		}
+		// We cannot create the user runtime directory ourselves, since /run
+		// and /run/user are owned by root.
+		switch _, err := os.Stat(xdg); {
+		case err == nil:
+			return filepath.Join(xdg, "minikube", machineName), nil
+		case errors.Is(err, os.ErrNotExist):
+			klog.Warningf("XDG_RUNTIME_DIR %q does not exist, falling back to %s", xdg, baseDir())
+		default:
+			return "", fmt.Errorf("stat %q: %w", xdg, err)
+		}
+	}
+	// Fallback mirroring the XDG layout (/run/user/<uid>) under the temporary
+	// directory. Used on macOS, and on Linux when XDG_RUNTIME_DIR is unset or
+	// points to a missing directory.
+	return filepath.Join(baseDir(), "user", strconv.Itoa(os.Getuid()), "minikube", machineName), nil
+}
--- a/pkg/drivers/vfkit/vfkit.go
+++ b/pkg/drivers/vfkit/vfkit.go
@@ -53,6 +53,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/process"
 	"k8s.io/minikube/pkg/minikube/reason"
 	"k8s.io/minikube/pkg/minikube/run"
+	"k8s.io/minikube/pkg/minikube/runtimedir"
 	"k8s.io/minikube/pkg/minikube/style"
 )
 
@@ -243,6 +244,11 @@ func (d *Driver) Start() error {
 	var socketPath string
 
 	if d.VmnetHelper != nil {
+		socketDir, err := d.SocketDir()
+		if err != nil {
+			return err
+		}
+		d.VmnetHelper.MachineSocketDir = socketDir
 		socketPath = d.VmnetHelper.SocketPath()
 		if err := d.VmnetHelper.Start(socketPath); err != nil {
 			return err
@@ -282,10 +288,14 @@ func (d *Driver) Start() error {
 func (d *Driver) startVfkit(socketPath string) error {
 	var startCmd []string
 
+	vfkitSock, err := d.sockfilePath()
+	if err != nil {
+		return err
+	}
 	startCmd = append(startCmd,
 		"--memory", fmt.Sprintf("%d", d.Memory),
 		"--cpus", fmt.Sprintf("%d", d.CPU),
-		"--restful-uri", fmt.Sprintf("unix://%s", d.sockfilePath()),
+		"--restful-uri", fmt.Sprintf("unix://%s", vfkitSock),
 		"--log-level", "debug")
 
 	// On arm64 console= is required get boot messages in serial.log. On x86_64
@@ -342,7 +352,7 @@ func (d *Driver) startVfkit(socketPath string) error {
 	}
 
 	log.Debugf("executing: vfkit %s", strings.Join(startCmd, " "))
-	os.Remove(d.sockfilePath())
+	os.Remove(vfkitSock)
 	cmd := exec.Command("vfkit", startCmd...)
 
 	// Create vfkit in a new process group, so minikube caller can use killpg
@@ -493,6 +503,9 @@ func (d *Driver) Remove() error {
 			return fmt.Errorf("kill: %w", err)
 		}
 	}
+	if err := runtimedir.RemoveSocketDir(d.GetMachineName()); err != nil {
+		log.Debugf("Failed to remove socket dir: %v", err)
+	}
 	return nil
 }
 
@@ -593,9 +606,8 @@ func (d *Driver) diskPath() string {
 	return filepath.Join(machineDir, "disk.img")
 }
 
-func (d *Driver) sockfilePath() string {
-	machineDir := filepath.Join(d.StorePath, "machines", d.GetMachineName())
-	return filepath.Join(machineDir, sockFilename)
+func (d *Driver) sockfilePath() (string, error) {
+	return d.ResolveSocketPath(sockFilename)
 }
 
 func (d *Driver) pidfilePath() string {
@@ -672,7 +684,11 @@ type VMState struct {
 }
 
 func (d *Driver) GetVFKitState() (string, error) {
-	httpc := httpUnixClient(d.sockfilePath())
+	sock, err := d.sockfilePath()
+	if err != nil {
+		return "", err
+	}
+	httpc := httpUnixClient(sock)
 	var vmstate VMState
 	response, err := httpc.Get("http://_/vm/state")
 	if err != nil {
@@ -689,7 +705,11 @@ func (d *Driver) GetVFKitState() (string, error) {
 
 // SetVFKitState sets the state of the vfkit VM, (s is the state)
 func (d *Driver) SetVFKitState(s string) error {
-	httpc := httpUnixClient(d.sockfilePath())
+	sock, err := d.sockfilePath()
+	if err != nil {
+		return err
+	}
+	httpc := httpUnixClient(sock)
 	var vmstate VMState
 	vmstate.State = s
 	data, err := json.Marshal(&vmstate)
--- a/pkg/libmachine/drivers/base.go
+++ b/pkg/libmachine/drivers/base.go
@@ -19,6 +19,8 @@ package drivers
 import (
 	"errors"
 	"path/filepath"
+
+	"k8s.io/minikube/pkg/minikube/runtimedir"
 )
 
 const (
@@ -39,6 +41,11 @@ type BaseDriver struct {
 	SwarmMaster    bool
 	SwarmHost      string
 	SwarmDiscovery string
+
+	// socketDir is the per-machine socket directory, resolved and created on
+	// first use. It is not persisted: the runtime directory may not survive a
+	// reboot.
+	socketDir string
 }
 
 // DriverName returns the name of the driver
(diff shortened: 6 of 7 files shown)


rafaelmfried:
**Empirical validation for this change.**

_Note on commits: these runs were performed on the original implementation (`515c9a4d`); the current PR HEAD (`90142df9`) is that work rebased onto current master plus naming/doc-only changes (e.g. the `EnsureSocketPath` rename) — no behavioral difference from what's validated below._

## Environment

```
macOS:        15.5 (24F74) — Sequoia
Arch:         arm64 (Apple Silicon)
Go:           go1.24.0 darwin/arm64
QEMU:         11.0.0 (brew)
vfkit:        0.6.3  (brew)
minikube:
  - master:   v1.36.0  (brew, pre-fix — used to reproduce the bug)
  - branch:   built from this PR
```

---

## Phase A — Bug reproduction (without the fix)

minikube `v1.36.0` (brew) + `--driver=qemu2` + long `MINIKUBE_HOME`:

```
export MINIKUBE_HOME="/Users/rafaelfriederick/.really-long-path-that-forces-bug-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
minikube start --driver=qemu2 --network=user -p test-long-bug-aaaaaaaaaaaaa
```

Error captured:

```
qemu-system-aarch64: -qmp unix:/Users/.../monitor,server,nowait:
  UNIX socket path '/Users/.../machines/test-long-bug-aaaaaaaaaaaaa/monitor' is too long
  Path must be less than 104 bytes

X Exiting due to GUEST_PROVISION: error provisioning guest: Failed to start host
```

This is the verbatim error from the issue report.

---

## Phase B — Fix validation (with the fix)

Built from this branch, same long `MINIKUBE_HOME`:

### B.2 — start succeeds, socket lives in the runtime directory

```
* Done! kubectl is now configured to use "test-long-fix-aaaaaaaaaaaaa" cluster
```

Socket location + hash verification:

```
socket: /tmp/501/minikube/be4a8a3f4d7bb581655549054f8f4ccb/monitor

$ echo -n "test-long-fix-aaaaaaaaaaaaa" | shasum -a 256 | cut -c1-32
be4a8a3f4d7bb581655549054f8f4ccb        # matches the directory name
```

Non-socket per-machine state stays under `MINIKUBE_HOME` as designed (disk image, configs, `qemu.pid`, `serial.log` remain; only `monitor` moved out). `grep -i "too long"` on the start log → empty (bug eliminated).

### B.3 — golden path

- `kubectl get nodes` → control-plane `Ready`
- `minikube status` → host/kubelet/apiserver Running
- `minikube stop` + `minikube start` → socket recreated at the same path on restart, which proves the pure `SocketPath` resolver in `RunQMPCommand` agrees with the `EnsureSocketPath` bind path

### B.4 — delete cleans the runtime dir

```
* Removed all traces of the "test-long-fix-aaaaaaaaaaaaa" cluster.
$ ls /tmp/501/minikube/be4a8a3f4d7bb581655549054f8f4ccb
ls: No such file or directory   # runtimedir.Cleanup() worked
```

### B.5 — default MINIKUBE_HOME still works (no regression)

```
$ echo -n "test-default" | shasum -a 256 | cut -c1-32
7985dded0b0113415535c36476463060        # matches socket dir under /tmp/<uid>/minikube/
```

---

## Summary

| | Without the fix (v1.36.0) | With the fix |
|---|---|---|
| Long `MINIKUBE_HOME` + qemu2 start | ❌ `sun_path too long`, GUEST_PROVISION | ✅ Done, node Ready |
| Socket location (long home) | `$MINIKUBE_HOME/.minikube/machines/<n>/monitor` | `/tmp/<uid>/minikube/<sha256>/monitor` |
| Restart after stop | n/a (start fails) | ✅ same socket path (deterministic) |
| Delete cleans runtime | n/a | ✅ runtime hash dir removed |
| Default `MINIKUBE_HOME` | ✅ works | ✅ works (no regression) |
| Non-socket state in `MINIKUBE_HOME` | unchanged | ✅ unchanged |

---

The deterministic, containerized reproduction (runs the real `qemu` binary in Testcontainers to reproduce the bug and prove the fix) is in the PoC linked in the description; full per-step logs available on request.

nirs:
@rafaelmfried thanks for the details PR message!

> ### What changed
> * New package `pkg/minikube/runtimedir/` exposing two entry points so callers can opt in to filesystem side effects:
>   
>   * `SocketPath(machineName, sockName string) (string, error)` — pure resolver, no filesystem writes. Used by callers that dial a socket the server already bound (state polls).
>   * `EnsureSocketPath(machineName, sockName string) (string, error)` — resolves the path and creates the parent directory with mode `0700`. Used by callers that bind a new socket.

I think we can have only `EnsureSocketDir(machineName)`. This returns the path to the per-machine socket directory using XDG_RUNTIME_DIR if this directory exists or fallback to /tmp/user/uid. If the directory does not exist we should not use it since we cannot create a directory in `/run/` or `/run/user`. If XDG_RUNTIME_DIR is defined but the directory is missing we are running in a broken system but we can recover by using `/tmp/user/uid/`.

Current code is using the machine directory to build the socket path. We need to replace the machine directory the runtimedir socket dir - the rest of the code should not change.

The socket name remains in the caller, runtimedir should not care about it.

>   * `pkg/drivers/common/vmnet` — `vmnet-helper.sock`. `Helper.EnsureSocketPath` resolves the path and creates the dir (called from driver Start before the helper binds). The helper's pidfile and logfile remain in `MachineDir` (regular files, no `sun_path` limit).

Not sure why we need `Helper.EnsureSocketPath` - the helper should get the machine name and call runtimedir.EnsureSocketDir(machineName) to get the path.

> * `vmnet.Helper` gains a `MachineName` field so the runtime path can be resolved without reaching into `MachineDir`'s filesystem layout. Populated at the two construction sites: `pkg/minikube/registry/drvs/vfkit` and `pkg/minikube/registry/drvs/krunkit`.

Makes sense.

pkg/drivers/krunkit/krunkit.go
```diff
@@ -469,12 +480,25 @@ func (d *Driver) isoPath() string {
 	return d.ResolveStorePath(isoFileName)
 }
 
-func (d *Driver) sockfilePath() string {
-	return d.ResolveStorePath(sockFileName)
+// sockfilePath returns the krunkit REST socket path without touching the
+// filesystem. Used by state read/write callers that dial an existing socket.
+func (d *Driver) sockfilePath() (string, error) {
+	return runtimedir.SocketPath(d.GetMachineName(), sockFileName)
```
nirs: Since use this path multiple times in many drivers. The best way to handle this is to create the socket dir and cache the path in common.Driver, and add a helper like ResolveStorePath - ResolveSocketPath(name).

This way we create the socket directory once in all drivers and we minimize the changes in the drivers - replacing ResolveStorePath(filename) with ResolveSocketPath(filename).

pkg/drivers/qemu/qemu.go
```diff
@@ -437,8 +438,18 @@ func (d *Driver) Start() error {
...
+	monitorSock, err := d.ensureMonitorPath()
+	if err != nil {
+		return fmt.Errorf("resolve QMP socket path: %w", err)
+	}
+	// Best-effort: clear a stale socket from a previous run so QEMU can bind.
+	// Failures (e.g. socket owned by root from a prior sudo run) are logged
+	// but do not block start — QEMU will surface a clearer error if needed.
+	if err := os.Remove(monitorSock); err != nil && !errors.Is(err, os.ErrNotExist) {
+		log.Debugf("Failed to remove stale QMP socket %q: %v", monitorSock, err)
+	}
```
nirs, line 445: Change not needed - we can keep the existing behavior - if we did not remove the socket before we don't need to remove it now, and there is no reason why the socket will be owned by root.

pkg/drivers/common/vmnet/vmnet.go
```diff
@@ -60,6 +61,10 @@ type Helper struct {
 	// The pidfile and log are stored here.
 	MachineDir string
 
+	// MachineName identifies the per-machine runtime directory used for
+	// the AF_UNIX socket. See pkg/minikube/runtimedir.
+	MachineName string
```
nirs: Fo consistency with MachineDir, pass here MachineSocketDir. The call will handle creation of the directory and we can keep the simple SocketPath() semantics.

pkg/minikube/runtimedir/runtimedir.go
```diff
@@ -0,0 +1,165 @@
...
+// name are long. This package moves only the sockets to a short runtime
+// directory while leaving other per-machine state (disk images, logs,
+// configs) under MINIKUBE_HOME.
+//
+// Layout:
+//
+//	Linux:    $XDG_RUNTIME_DIR/minikube/<hash>/<name>
+//	          fallback /tmp/<uid>/minikube/<hash>/<name>
+//	macOS:    /tmp/<uid>/minikube/<hash>/<name>
+//	Windows:  <os.TempDir()>/minikube/<hash>/<name>
```
nirs: Need to document windows case:
- Max user name is 20 characters https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/net-user
- So the standard per-user temporary directory is good enough: `C:\Users\<username>\AppData\Local\Temp`

pkg/minikube/runtimedir/runtimedir.go
```diff
@@ -0,0 +1,165 @@
...
+	"path/filepath"
+	"runtime"
+	"strconv"
+)
+
+const (
+	// runtimeDirMode is the permission applied to every directory created
+	// by this package. AF_UNIX sockets bound inside need only be reachable
+	// by the owning user.
+	runtimeDirMode = 0o700
```
nirs: dirMode - we are in the context of runtimedir. Both the machine directory and user directory need the same mode.

pkg/minikube/runtimedir/runtimedir.go
```diff
@@ -0,0 +1,165 @@
...
+
+// SocketPath returns the AF_UNIX-safe path for the given per-machine socket
+// without writing to the filesystem. Use this when dialing a socket the server
+// already bound (e.g. reading VM state).
+//
+// The returned path is deterministic for the same (machineName, sockName)
+// pair on the same host, so the server and the client agree on the location.
+func SocketPath(machineName, sockName string) (string, error) {
+	return resolveSocketPath(runtime.GOOS, os.Getuid(), os.Getenv("XDG_RUNTIME_DIR"), os.TempDir(), machineName, sockName, false)
+}
```
nirs: Not needed

pkg/minikube/runtimedir/runtimedir.go
```diff
@@ -0,0 +1,165 @@
...
+
+// EnsureSocketPath behaves like SocketPath but also creates the parent
+// directory with mode 0700. Use this when binding a new socket.
+func EnsureSocketPath(machineName, sockName string) (string, error) {
```
nirs: EnsureSocketDir(machineName) (string, error)

pkg/minikube/runtimedir/runtimedir.go
```diff
@@ -0,0 +1,165 @@
...
+// Cleanup removes the per-machine runtime directory. It is idempotent and
+// safe to call when the directory does not exist.
+func Cleanup(machineName string) error {
+	return cleanup(runtime.GOOS, os.Getuid(), os.Getenv("XDG_RUNTIME_DIR"), os.TempDir(), machineName)
+}
+
+// resolveSocketPath is the env-free variant used by the public entry points
+// and by tests. When ensure is true, the parent directory is created with
+// mode 0700; otherwise the filesystem is left untouched.
+func resolveSocketPath(goos string, uid int, xdg, winTmp, machineName, sockName string, ensure bool) (string, error) {
```
nirs: Rename to resolve and remove all the arguments. Lets start with simple interface that make sense.

For example to make the code testable, we can add a function returning the the base dir "/tmp":

var baseDir = func() string {
    ...
}

We can change this in the tests to ensure that tests are isolated and do not change the environment. Now the uid does not matter since each test can use its own base tmpdir.

We can do the same for xdgRuntimeDir() - the function will return the env variable and the tests can modify it to return empty value or non existing directory.

The code remain sane, using baseDir() and xdgRuntimeDir(), not passing any argument we don't need and we can test all edge cases.

pkg/minikube/runtimedir/runtimedir.go
```diff
@@ -0,0 +1,165 @@
...
+//     enough.
+func baseDir(goos string, uid int, xdg, winTmp string) string {
+	if goos == "darwin" {
+		return filepath.Join("/tmp", strconv.Itoa(uid), "minikube")
+	}
+	if goos == "windows" {
+		return filepath.Join(winTmp, "minikube")
+	}
+	if xdg != "" {
+		fi, err := os.Stat(xdg)
```
nirs: Need to check stat errors:
- os.ErrNotExists: log warning and fallback to /tmp (broken system)
- other: fail, we cannot handle this

pkg/minikube/runtimedir/runtimedir.go
```diff
@@ -0,0 +1,165 @@
...
+//
+//   - macOS always uses /tmp/<uid>/minikube. os.TempDir() on macOS returns a
+//     long path under /var/folders that would defeat the whole point.
+//   - Linux prefers $XDG_RUNTIME_DIR when it exists and is a directory,
+//     falling back to /tmp/<uid>/minikube.
+//   - Windows uses os.TempDir() (i.e. %TEMP%). os.Getuid() returns -1 on
+//     Windows, so /tmp/<uid> is not usable. AF_UNIX support on Windows 10+
+//     has the same sun_path[108] limit; %TEMP% paths are typically short
+//     enough.
+func baseDir(goos string, uid int, xdg, winTmp string) string {
```
nirs: Passing goos, uid, xdd, winTemp does not make sense. Help the callers by accepting nothing and returning the right location.

pkg/minikube/runtimedir/runtimedir.go
```diff
@@ -0,0 +1,165 @@
...
+	if goos == "darwin" {
+		return filepath.Join("/tmp", strconv.Itoa(uid), "minikube")
+	}
+	if goos == "windows" {
+		return filepath.Join(winTmp, "minikube")
+	}
+	if xdg != "" {
```
nirs:
```go
case "linux":
```

And get XDG_RUNTIME_DIR here

pkg/minikube/runtimedir/runtimedir.go
```diff
@@ -0,0 +1,165 @@
...
+//   - macOS always uses /tmp/<uid>/minikube. os.TempDir() on macOS returns a
+//     long path under /var/folders that would defeat the whole point.
+//   - Linux prefers $XDG_RUNTIME_DIR when it exists and is a directory,
+//     falling back to /tmp/<uid>/minikube.
+//   - Windows uses os.TempDir() (i.e. %TEMP%). os.Getuid() returns -1 on
+//     Windows, so /tmp/<uid> is not usable. AF_UNIX support on Windows 10+
+//     has the same sun_path[108] limit; %TEMP% paths are typically short
+//     enough.
+func baseDir(goos string, uid int, xdg, winTmp string) string {
+	if goos == "darwin" {
```
nirs:
```go
switch runtime.GOOS {
case "darwin":
    ...
```

pkg/minikube/runtimedir/runtimedir.go
```diff
@@ -0,0 +1,165 @@
...
+		return filepath.Join("/tmp", strconv.Itoa(uid), "minikube")
+	}
+	if goos == "windows" {
```
nirs:
```
case "windows":
```

And use os.TempDir() here

pkg/minikube/runtimedir/runtimedir.go
```diff
@@ -0,0 +1,165 @@
...
+		fi, err := os.Stat(xdg)
+		if err == nil && fi.IsDir() {
+			return filepath.Join(xdg, "minikube")
+		}
+	}
+	return filepath.Join("/tmp", strconv.Itoa(uid), "minikube")
+}
+
+// ensureDir creates dir (and parents) with mode 0700, then forces the mode
+// with Chmod to bypass any umask that would relax the permission.
```
nirs: the umask cannot relax the permissions only make it more strict.

The chmod is needed if the directory already exists and was created by with the wrong permissions.

pkg/minikube/runtimedir/runtimedir.go
```diff
@@ -0,0 +1,165 @@
...
+// EnsureSocketPath behaves like SocketPath but also creates the parent
+// directory with mode 0700. Use this when binding a new socket.
+func EnsureSocketPath(machineName, sockName string) (string, error) {
+	return resolveSocketPath(runtime.GOOS, os.Getuid(), os.Getenv("XDG_RUNTIME_DIR"), os.TempDir(), machineName, sockName, true)
+}
+
+// Cleanup removes the per-machine runtime directory. It is idempotent and
+// safe to call when the directory does not exist.
+func Cleanup(machineName string) error {
+	return cleanup(runtime.GOOS, os.Getuid(), os.Getenv("XDG_RUNTIME_DIR"), os.TempDir(), machineName)
```
nirs: Should be:

   path, err := resolve(machineName)
   ...
   remove it ..

nirs (commented):
I think we can simplify this using:

1. Add `rumtimedir.EnsureSocketDir(machineName)` - create and return the socket directory from machine name
1. Remove `runtimedir.ResolveSocketPath()` - not needed 
1. Add `libmachine/driver.BaseDriver.ResolveSocketPath()` - like ResolveStorePath() but using `runtimedir.EnsureSocketDir(d.machineName)`
1. Use d.ResolveSocketPath(filename) in the drivers when resolving socket path. We will have to handle errors since `runtimedir.EnsureSocketDir(machineName)` can fail.
1. Simplify runtimedir implementation by adding baseDir() and xdgRuntimeDir() variables functions to allow easy testing without passing so many argument around

I would also separate it to 2 smaller commits:
- runtimedir package with tests
- add libmachine/driver.BaseDriver.ResolveSocketPath() and use it the drivers

The changes in the drivers can be minimal - no other change needed except calling ResolveSocketPath instead of ResoreStorePath

rafaelmfried:
Thanks @nirs — this all makes sense; I'll rework it along these lines (single `EnsureSocketDir` + a driver-level `ResolveSocketPath` like `ResolveStorePath`) over the next few days, a bit tied up at work this week.

nirs:
> Thanks @nirs — this all makes sense; I'll rework it along these lines (single `EnsureSocketDir` + a driver-level `ResolveSocketPath` like `ResolveStorePath`) over the next few days, a bit tied up at work this week.

Sure, thanks for working on this.

pkg/minikube/runtimedir/runtimedir.go
```diff
@@ -0,0 +1,165 @@
...
+		return filepath.Join("/tmp", strconv.Itoa(uid), "minikube")
+	}
+	if goos == "windows" {
+		return filepath.Join(winTmp, "minikube")
+	}
+	if xdg != "" {
+		fi, err := os.Stat(xdg)
+		if err == nil && fi.IsDir() {
+			return filepath.Join(xdg, "minikube")
+		}
```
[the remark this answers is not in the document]
nirs: @rafaelmfried  No need to handle long XDG_RUNTIME_DIR - this is the responsibility of the system or the user overriding the value. If an admin breaks the system they can keep the pieces.

pkg/drivers/common/vmnet/vmnet.go
```diff
@@ -325,8 +330,13 @@ func (h *Helper) GetState() (state.State, error) {
...
 
-func (h *Helper) SocketPath() string {
-	return filepath.Join(h.MachineDir, sockfileName)
+// EnsureSocketPath returns the AF_UNIX path used by vmnet-helper and ensures the
+// parent directory exists. It lives in a short runtime directory (see
+// pkg/minikube/runtimedir) so the full path fits within sockaddr_un.sun_path
+// regardless of $MINIKUBE_HOME length. Called during driver Start, before
+// vmnet-helper binds the socket.
+func (h *Helper) EnsureSocketPath() (string, error) {
+	return runtimedir.EnsureSocketPath(h.MachineName, sockfileName)
```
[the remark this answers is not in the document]
nirs: @rafaelmfried MachineName should never be empty. I think this filed exists since vment was released in 1.36 or 1.37. You can check git history to verify.

nirs:
@rafaelmfried any update?

rafaelmfried:
Sorry @nirs, I've been busy with a DevOps project here at the company and got a bit stuck these past few days. However, I'll take a look this week and give you some direction on the PR.

rafaelmfried:
@nirs sorry for the long delay — personal reasons, thanks for your patience. Pushed a rewrite following your review (single squashed commit).

**One entry point.** `SocketPath()`/`EnsureSocketPath()` are gone. The package now exposes `EnsureSocketDir(machineName) (string, error)` and `RemoveSocketDir(machineName) error`. The socket name stays in the caller.

**`resolve()` takes no arguments** and switches on `runtime.GOOS`:

```go
case "windows":  // %TEMP%\minikube\<machine>, user names are limited to 20 chars
case "linux":    // $XDG_RUNTIME_DIR/minikube/<machine>
default:         // /tmp/user/<uid>/minikube/<machine>
```

`XDG_RUNTIME_DIR` is read inside the `linux` case. If it does not exist we log a warning and fall back to the temporary directory, since we cannot create a directory under `/run/user`; any other stat error fails. Long `XDG_RUNTIME_DIR` is not handled, as you suggested.

**Testability.** The package uses `baseDir()`, `xdgRuntimeDir()` and `tempDir()` indirections, replaced in the tests. Each test gets its own `t.TempDir()`, so the uid does not matter and no test touches the environment.

**`ResolveSocketPath()` in the driver.** Added to `BaseDriver` next to `ResolveStorePath()`, since that is where `MachineName` lives:

```go
func (d *BaseDriver) SocketDir() (string, error)          // ensures + caches
func (d *BaseDriver) ResolveSocketPath(file string) (string, error)
```

The socket directory is created and cached on the first call, so drivers create it once, and the change in each driver is mechanical: `ResolveStorePath(name)` → `ResolveSocketPath(name)`. The cache is a private field, so it is not persisted — the runtime directory may not survive a reboot.

**vmnet.** `Helper.EnsureSocketPath()` is gone. The helper stores the socket directory in `MachineSocketDir`, for consistency with `MachineDir`, and `SocketPath()` is a plain join again. The driver creates the directory and sets the field in `Start()`. `MachineName` is no longer needed on the helper.

**qemu.** The socket removal is reverted, as you asked — only `monitorPath()` moved to the socket directory.

Two things worth flagging, since they are not literally what you asked for:

1. **The machine name hash is gone.** Previously the directory was `sha256(machineName)[:32]`. The `EnsureSocketDir(machineName)` shape makes the plain name the natural choice, and it is far easier to debug (`/tmp/user/501/minikube/minikube/`). Budget on macOS, the tightest case: `/tmp/user/<uid>/minikube/<machine>/vmnet-helper.sock` is 41 chars + the machine name, leaving ~60 chars for the machine name within `sun_path[104]`. Happy to bring the hash back if you prefer it.
2. **State polls now create the socket directory.** With a single entry point, callers that only dial an existing socket (`GetState`, `RunQMPCommand`, `GetVFKitState`) go through `ResolveSocketPath()` and therefore `MkdirAll`. It is idempotent and the directory already exists whenever the VM runs, but it is a behavior change from a pure resolver.

PTAL.

pkg/drivers/krunkit/krunkit.go
rafaelmfried: Done — added `BaseDriver.ResolveSocketPath()` next to `ResolveStorePath()`. It creates and caches the socket dir on first use, so each driver just swaps `ResolveStorePath(name)` for `ResolveSocketPath(name)`.

pkg/drivers/qemu/qemu.go
rafaelmfried: Done — reverted, qemu keeps the existing behavior.

pkg/drivers/common/vmnet/vmnet.go
rafaelmfried: Done — the helper now has `MachineSocketDir` and `SocketPath()` is a plain join again. The driver creates the directory in `Start()`.

pkg/minikube/runtimedir/runtimedir.go
rafaelmfried: Done — documented in the `windows` case of `resolve()`.

pkg/minikube/runtimedir/runtimedir.go
rafaelmfried: Done — renamed to `dirMode` and used for every directory the package creates.

pkg/minikube/runtimedir/runtimedir.go
rafaelmfried: Done — removed. `EnsureSocketDir()` is the only entry point now.

pkg/minikube/runtimedir/runtimedir.go
rafaelmfried: Done — that is the signature now.

pkg/minikube/runtimedir/runtimedir.go
rafaelmfried: Done — `resolve(machineName)` takes nothing else and uses the `baseDir()`, `xdgRuntimeDir()` and `tempDir()` indirections. The tests replace them with `t.TempDir()`, so no test touches the environment and the uid does not matter.

pkg/minikube/runtimedir/runtimedir.go
rafaelmfried: Done — `os.ErrNotExist` logs a warning and falls back to the temporary directory; any other stat error fails.

pkg/minikube/runtimedir/runtimedir.go
rafaelmfried: Done — `resolve()` accepts only the machine name.

pkg/minikube/runtimedir/runtimedir.go
rafaelmfried: Done — `XDG_RUNTIME_DIR` is read inside `case "linux"`.

pkg/minikube/runtimedir/runtimedir.go
rafaelmfried: Done — `switch runtime.GOOS` now. macOS is the `default` case, shared with the other unixes; happy to make it an explicit `case "darwin"` if you prefer.

pkg/minikube/runtimedir/runtimedir.go
rafaelmfried: Done — `case "windows"` uses `os.TempDir()`.

pkg/minikube/runtimedir/runtimedir.go
rafaelmfried: Right — kept the chmod for exactly that case. Covered by `TestEnsureSocketDirExistingMode`.

pkg/minikube/runtimedir/runtimedir.go
rafaelmfried: Done — `RemoveSocketDir()` resolves and removes.

pkg/minikube/runtimedir/runtimedir.go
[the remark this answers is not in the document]
rafaelmfried: Done — dropped, no special handling for a long `XDG_RUNTIME_DIR`.

pkg/drivers/common/vmnet/vmnet.go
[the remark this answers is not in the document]
rafaelmfried: Moot now — the helper no longer takes `MachineName`. It gets `MachineSocketDir` from the driver, which resolves it from the machine name.

Issue #23041, opened 2026-05-25
ci: Ensure workflow timeouts - runs can get stuck for hours


nirs:
This run was stuck for 6 hours! in the apt-get install step
https://github.com/kubernetes/minikube/actions/runs/26373054430/job/77628725110?pr=23040

## Fix

- [ ] Set job level timeout
- [ ] Set step level timeout

### Job level timeout

```yaml
jobs:
  build:
    runs-on: ubuntu-latest
    timeout-minutes: 30
```

### Step level timeout

```yaml
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - name: Run complex tests
        run: ./run-tests.sh
        timeout-minutes: 10
```

For the next step see #23043.

nirs:
@SzymonIwaniuk adding timeouts is easy - finding the right timeout is not.

The best way to do this is to analyze the logs for at least 10 runs and find the typical time for each step. Multiple the maximum time by 3 and round to minutes.

Start with one workflow, when we have working configuration it will be easier to apply the same change to other workflows.

nirs:
@SzymonIwaniuk I create a tool for computing steps stats. Here are the stats for successful jobs in the last 30 days (241 runs):

| Step | N | Min | Avg | P50 | P90 | P95 | Max | Timeout |
|---|--:|--:|--:|--:|--:|--:|--:|--:|
| Run Functional Test | 1499 | 2m48s | 4m19s | 4m00s | 5m17s | 6m02s | 15m18s | 19m00s |
| Build minikube and e2e test binaries | 369 | 1m07s | 1m39s | 1m37s | 2m08s | 2m10s | 2m29s | 7m00s |
| Set up Rootless Docker (rootless) | 161 | 42s | 50s | 50s | 55s | 57s | 1m28s | 3m00s |
| Run actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c | 1713 | 7s | 21s | 21s | 26s | 30s | 49s | 2m00s |
| Run actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 | 155 | 6s | 19s | 21s | 26s | 30s | 40s | 2m00s |
| Update apt-get package index (ubuntu) | 733 | 5s | 14s | 11s | 21s | 26s | 46s | 2m00s |
| Upload Test Binaries | 369 | 8s | 10s | 10s | 13s | 13s | 14s | 1m00s |
[...]

PR #23042, opened 2026-05-25
hack: add workflow-stats tool for analyzing CI step durations

nirs:
Analyzes GitHub Actions workflow step durations to help set per-step timeouts based on historical data. Computes min, avg, P50, P90, P95, max, and a suggested timeout (3x P95, rounded to minutes) across completed workflow runs.

Features:
- SQLite cache at ~/.cache/workflow-stats/<owner>/<repo>/stats.db avoids redundant API calls; only new runs are fetched
- Incremental updates using the latest cached run date
- Filter by job name (-job), conclusion (-conclusion), branch (-branch)
- Output as table (default), markdown, CSV, or JSON

Dependencies:
- google/go-github/v85: GitHub Actions API client for fetching workflow runs and job details
- modernc.org/sqlite: pure-Go SQLite driver (transpiled from C), chosen over mattn/go-sqlite3 to avoid CGO build dependencies

## Example usage

```console
$ go run workflow-stats/workflow_stats.go -workflow "Functional Test"
Fetching runs since 2026-05-24 ... 6 runs (0.7s)

  Step                                                N     Min     Avg     P95     Max  Timeout
  Run Functional Test                               728   2m50s   4m05s   5m21s   9m27s   17m00s
  Build minikube and e2e test binaries              156   1m07s   1m39s   2m09s   2m16s    7m00s
  Set up Rootless Docker (rootless)                  67     42s     48s     56s   1m21s    3m00s
  Update apt-get package index (ubuntu)             373      5s     12s     23s     42s    2m00s
  ...
```

Related-to #23041 
Related-to #23043


hack/workflow-stats/workflow_stats.go
```diff
@@ -0,0 +1,589 @@
+/*
```
medyagh, line 1: how about add a readme to this folder how to use this and what this thing is...or what uses it

medyagh (changes requested):

hack/workflow-stats/workflow_stats.go
nirs: Good idea - we have good content in the package comment that can be more useful as readme.

nirs: README.md added with example usage.

hack/workflow-stats/README.md
```diff
@@ -0,0 +1,109 @@
...
+Analyze GitHub Actions workflow step durations for the minikube project.
+
+Fetches completed workflow runs via the GitHub API, caches results in a
+local SQLite database, and reports per-step timing statistics (min, avg,
+p50, p90, p95, max) along with a suggested timeout.
+
+## Building
+
+```
+cd hack && go build -o workflow-stats ./workflow-stats/
```
medyagh: can this be a go run ? if not how about add the binary to .gitignore?

medyagh (commented):
@nirs this is actually super cool,
some idea (not for this PR ofc )

but it would be cool if we have an automation job that every 14 days gathers this logs and makes a graphic chart our of them and adds them to our site

https://minikube.sigs.k8s.io/docs/benchmarks/

we could potenitally have something like this for example for "Functional Test on Docker"
we export the data that we get into a csv

- (how long Run integration Step Took)
- Add to the CSV in our hack folder (timestamp, env name,value)
- Generate some chart with Google Charts like this 
- https://minikube.sigs.k8s.io/docs/benchmarks/timetok8s/weekly_benchmark/

and make a PR to add it to our site once a month (update functional test benchmarking)

hack/workflow-stats/README.md
nirs: Sure, we show go run in the next example. Do you want to always use go run so we don't need to deal with ignoring the executables?

nirs:
@medyagh Comment addressed:
- Removed "Building" section to we don't need to handle ignoring tools
- Use `go run` in all the examples
- Unified flag usage (`-workflow` instead of `--workflow`)

hack/workflow-stats/workflow_stats.go
```diff
@@ -0,0 +1,589 @@
...
+	requestedSince := time.Now().UTC().AddDate(0, 0, -opts.Since)
+	if fetchSince.Before(requestedSince) {
+		fetchSince = requestedSince
+	}
+
+	fmt.Fprintf(os.Stderr, "Fetching runs since %s ...", fetchSince.Format("2006-01-02"))
+	t := time.Now()
+	wfID := findWorkflowID(ctx, client, opts.Owner, opts.Repo, opts.Workflow)
+	runs := fetchRuns(ctx, client, opts.Owner, opts.Repo, wfID, opts.Branch, fetchSince)
+	fmt.Fprintf(os.Stderr, " %d runs (%.1fs)\n", len(runs), time.Since(t).Seconds())
```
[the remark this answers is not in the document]
nirs: Need to consider the branch flag, I think we can remove it. This tool should be used on the master branch. We can add branch option later if we have a real need.

hack/workflow-stats/README.md
```diff
@@ -0,0 +1,109 @@
...
+Fetching runs since 2026-05-24 ... 50 runs (2.6s)
+Fetching 44 new runs ... (46.7s)
+
+Step                                                    N       Min       Avg       P50       P90       P95       Max    Timeout
+────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
+Run Functional Test                                   932     2m47s     4m05s     3m58s     5m09s     5m23s     9m27s     17m00s
+Build minikube and e2e test binaries                  200     1m07s     1m39s     1m38s     2m07s     2m09s     2m21s      7m00s
+Set up Rootless Docker (rootless)                      87       42s       47s       46s       53s       55s     1m01s      3m00s
+Run actions/setup-go@4a3601121dd01d1626a1e23e3721…   1132        7s       20s       21s       26s       29s     1m01s      2m00s
+Update apt-get package index (ubuntu)                 475        5s       11s        8s       20s       23s       48s      2m00s
```
nirs, line 52: We need to add job name to the table, so we can see the best timeout for each job/step combination.

When using matrix, we may be able to set the timeout in the matrix, and use:

```yaml
timeout-minutes: ${{ matrix.timeout-minutes }}
```

The code updating the timeouts can search timeout-minutes in the matrix when it finds `timeout-minutes: ${{ matrix.timeout-minutes }}`. This way we can fail fast jobs (ubuntu) quickly and wait longer only for slow builds (macos-15-intel).

PR #23058, opened 2026-05-27
golangci-lint: Fix all lint errors with latest version

nirs:
Tested with golangci-lint 2.12.2 (go1.26.2, c0d3ddc, 2026-05-06).

- Exclude pkg/drivers/parallels/ from linting: deprecated driver that no one can test, so fixing lint errors there is not worthwhile.

- Disable revive if-return: the explicit "if err != nil { return err }" pattern is more maintainable than "return f()" because adding a step between the call and the return is a one-line diff.

- Disable revive unsecure-url-scheme: the flagged URLs are unix socket addresses (http://_/vm/state), not real HTTP endpoints. HTTPS is nonsensical here.

- Disable revive error-strings: redundant with staticcheck ST1005, which catches significantly more cases (170+ vs 4).

- Enable revive bare-return globally, with a per-file exclusion for pkg/minikube/tunnel/route_darwin.go to avoid fixing existing code. New code will be checked.

- Move revive package-naming exclusion for pkg/libmachine/log/ from //nolint:revive comments in 5 source files to a single rule in .golangci.yaml. Centralizing suppressions in the config keeps them visible in one place and keeps source code free of linter noise.

## Before

```console
% make lint
./out/linters/golangci-lint-v2.12.2 run --max-issues-per-linter 0 --max-same-issues 0 --build-tags "integration libvirt_dlopen" --config .golangci.yaml  ./...
pkg/drivers/parallels/parallels_darwin.go:425:47: printf: k8s.io/minikube/pkg/libmachine/log.Debugf format %d has arg ver of wrong type *github.com/hashicorp/go-version.Version (govet)
        log.Debugf("Found Parallels Desktop version: %d, edition: %s", ver, edit)
                                                     ^
pkg/drivers/parallels/parallels_darwin.go:633:17: inline: Call of ioutil.ReadFile should be inlined (govet)
        leases, err := ioutil.ReadFile(DHCPLeaseFile)
                       ^
pkg/drivers/krunkit/krunkit.go:481:9: unsecure-url-scheme: prefer secure protocol https over http in "http://_/vm/state" (revive)
        return "http://_/vm/state"
               ^
pkg/drivers/parallels/parallels_darwin.go:322:2: if-return: redundant if ...; err != nil check, just return error instead. (revive)
        if err := d.Start(); err != nil {
                return err
        }
pkg/drivers/parallels/parallels_darwin.go:406:10: unnecessary-format: unnecessary use of formatting function "fmt.Errorf", you can replace it with "errors.New" (revive)
                return fmt.Errorf("Driver \"parallels\" works only on macOS!")
                       ^
pkg/drivers/parallels/parallels_darwin.go:416:21: error-strings: error strings should not be capitalized or end with punctuation or a newline (revive)
                return fmt.Errorf("Driver \"parallels\" supports only Parallels Desktop 11 and higher. You use: Parallels Desktop %s.", ver)
                                  ^
pkg/drivers/parallels/parallels_darwin.go:429:3: useless-break: useless break in case clause (revive)
                break
                ^
pkg/drivers/parallels/parallels_darwin.go:442:2: if-return: redundant if ...; err != nil check, just return error instead. (revive)
        if err := b2dutils.UpdateISOCache(d.Boot2DockerURL); err != nil {
                return err
        }
pkg/drivers/parallels/parallels_darwin.go:563:3: useless-break: useless break in case clause (revive)
                break
                ^
pkg/drivers/parallels/parallels_darwin.go:625:25: error-strings: error strings should not be capitalized or end with punctuation or a newline (revive)
                return "", fmt.Errorf("MAC address for NIC: nic0 on Virtual Machine: %s not found!\n", d.MachineName)
                                      ^
pkg/drivers/parallels/parallels_darwin.go:630:25: error-strings: error strings should not be capitalized or end with punctuation or a newline (revive)
                return "", fmt.Errorf("Not a valid MAC address: %s. It should be exactly 12 digits.", mac)
                                      ^
pkg/drivers/parallels/parallels_darwin.go:653:25: error-strings: error strings should not be capitalized or end with punctuation or a newline (revive)
                return "", fmt.Errorf("IP lease not found for MAC address %s in: %s\n", mac, DHCPLeaseFile)
                                      ^
pkg/drivers/parallels/parallels_darwin.go:759:2: if-return: redundant if ...; err != nil check, just return error instead. (revive)
        if err := prldisktool("resize",
                "--hdd", d.diskPath(),
                "--size", fmt.Sprintf("%d", size)); err != nil {
                return err
        }
pkg/drivers/parallels/parallels_darwin.go:803:14: unnecessary-format: unnecessary use of formatting function "fmt.Errorf", you can replace it with "errors.New" (revive)
                return "", fmt.Errorf("Driver \"parallels\" requires Parallels Desktop license to be activated. More info: https://kb.parallels.com/en/124225")
                           ^
pkg/drivers/vfkit/vfkit.go:406:2: unnecessary-format: unnecessary use of formatting function "fmt.Fprintf", you can replace it with "fmt.Fprint" (revive)
        fmt.Fprintf(&b, "set -e\n")
        ^
pkg/drivers/vfkit/vfkit.go:407:2: unnecessary-format: unnecessary use of formatting function "fmt.Fprintf", you can replace it with "fmt.Fprint" (revive)
        fmt.Fprintf(&b, "sudo mount binfmt_misc -t binfmt_misc /proc/sys/fs/binfmt_misc\n")
        ^
pkg/drivers/vfkit/vfkit.go:700:29: unsecure-url-scheme: prefer secure protocol https over http in "http://_/vm/state" (revive)
        response, err := httpc.Get("http://_/vm/state")
                                   ^
pkg/drivers/vfkit/vfkit.go:722:22: unsecure-url-scheme: prefer secure protocol https over http in "http://_/vm/state" (revive)
        _, err = httpc.Post("http://_/vm/state", "application/json", bytes.NewReader(data))
                            ^
pkg/minikube/registry/drvs/krunkit/krunkit.go:106:15: unchecked-type-assertion: type cast result is unchecked in err.(*vmnet.Error) - type assertion will panic if not matched (revive)
                vmnetErr := err.(*vmnet.Error)
                            ^
pkg/minikube/registry/drvs/parallels/parallels.go:52:7: unchecked-type-assertion: type cast result is unchecked in parallels.NewDriver(config.MachineName(cfg, n), localpath.MiniPath()).(*parallels.Driver) - type assertion will panic if not matched (revive)
        d := parallels.NewDriver(config.MachineName(cfg, n), localpath.MiniPath()).(*parallels.Driver)
             ^
pkg/minikube/tunnel/route_darwin.go:66:3: bare-return: avoid using bare returns, please add return expressions (revive)
                return
                ^
pkg/minikube/tunnel/route_darwin.go:73:2: bare-return: avoid using bare returns, please add return expressions (revive)
        return
        ^
pkg/drivers/parallels/parallels_darwin.go:44:2: SA1019: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package [io] or package [os], and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
        "io/ioutil"
        ^
pkg/drivers/parallels/parallels_darwin.go:746:14: SA1019: os.SEEK_SET has been deprecated since Go 1.7: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd. (staticcheck)
        hds.Seek(0, os.SEEK_SET)
                    ^
24 issues:
* govet: 2
* revive: 20
* staticcheck: 2
make: *** [lint] Error 1
```

## After

```console
% make lint
./out/linters/golangci-lint-v2.12.2 run --max-issues-per-linter 0 --max-same-issues 0 --build-tags "integration libvirt_dlopen" --config .golangci.yaml  ./...
0 issues.
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -30,7 +30,6 @@ linters:
         - "-ST1005"
         # Omit embedded fields from selector expression
         - "-QF1008"
-        
     revive:
       enable-all-rules: true
       rules:
@@ -89,6 +88,14 @@ linters:
         - name: unused-receiver
           disabled: true
         - name: add-constant
+          disabled: true
+        ###### Redundant with other linters.
+        - name: error-strings
+          disabled: true # staticcheck ST1005 handles this and catches more cases
+        ###### Opinionated style rules - permanently disabled.
+        - name: if-return
+          disabled: true
+        - name: unsecure-url-scheme
           disabled: true
         ###### To be determined, this is a rule with differences.
         - name: import-alias-naming
@@ -156,6 +163,8 @@ linters:
       # we generally dont wanna touch or lint the Third_party code, it is basically like a fork for code that we can not import.
       # but have to copy/paste
       - third_party
+      # Deprecated driver, no one can test changes.
+      - pkg/drivers/parallels/
       # Skip test files
       - '(.+)_test\.go'
     rules:
@@ -187,6 +196,10 @@ linters:
       # Ignore package-naming rule in pkg/drivers/common/ because renaming it would create massive code churn and refactoring.
       - path: 'pkg/drivers/common/'
         text: "package-naming:"
+        linter: revive
+      # Existing bare returns in route_darwin.go - fix later.
+      - path: 'pkg/minikube/tunnel/route_darwin.go'
+        text: "bare-return:"
         linter: revive
 formatters:
   enable:
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -197,6 +197,10 @@ linters:
       - path: 'pkg/drivers/common/'
         text: "package-naming:"
         linter: revive
+      # pkg/libmachine/log shadows the standard library "log" package, but renaming it would touch the entire codebase.
+      - path: 'pkg/libmachine/log/'
+        text: "package-naming:"
+        linter: revive
       # Existing bare returns in route_darwin.go - fix later.
       - path: 'pkg/minikube/tunnel/route_darwin.go'
         text: "bare-return:"
--- a/pkg/libmachine/log/fmt_machine_logger.go
+++ b/pkg/libmachine/log/fmt_machine_logger.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package log //nolint:revive
+package log
 
 import (
 	"fmt"
--- a/pkg/libmachine/log/fmt_machine_logger_test.go
+++ b/pkg/libmachine/log/fmt_machine_logger_test.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package log //nolint:revive
+package log
 
 import (
 	"bufio"
--- a/pkg/libmachine/log/history_recorder_test.go
+++ b/pkg/libmachine/log/history_recorder_test.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package log //nolint:revive
+package log
 
 import (
 	"testing"
--- a/pkg/libmachine/log/log.go
+++ b/pkg/libmachine/log/log.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package log //nolint:revive
+package log
 
 import (
 	"io"
--- a/pkg/libmachine/log/machine_logger.go
+++ b/pkg/libmachine/log/machine_logger.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package log //nolint:revive
+package log
 
 import "io"
 


.golangci.yaml
```diff
@@ -90,6 +89,14 @@ linters:
           disabled: true
         - name: add-constant
           disabled: true
+        ###### Redundant with other linters.
+        - name: error-strings
```
medyagh, line 93: if redundant why are we adding a an additional block into the config?

.golangci.yaml
```diff
@@ -90,6 +89,14 @@ linters:
...
+          disabled: true # staticcheck ST1005 handles this and catches more cases
+        ###### Opinionated style rules - permanently disabled.
+        - name: if-return
+          disabled: true
+        - name: unsecure-url-scheme
```
medyagh, line 98: why are we adding this back ? 
it is nice to catch all non-http traffic and force ourselves to ignore them one by one? the lint was not failing before this

medyagh (changes requested):

.golangci.yaml
nirs: We add a block to disable this option - otherwise it is enabled and fails locally.

.golangci.yaml
nirs: http is fine in the case we use it with unix socket. There is nothing to improve - this is just noise.

This lint fails locally.

medyagh (approved):

Issue #17945, opened 2024-01-12
Fix auto-pause interval value not immediately being taking affect


spowelljr:
In https://github.com/kubernetes/minikube/pull/17936 I implemented making the auto-pause interval configurable, from the PR:
> There's an issue with running `minikube addons configure auto-pause`, the value is successfully saved, but the value is not immediately updated in the auto-pause addon itself. The systemd unit file is updated, but then a `systemctl daemon-reload` & `systemctl restart auto-pause.service` both have to be run for the update to be reflected in the binary. Due to the already large size of this PR I'll leave it for a follow up.

As stated, when the user executes `minikube addons configure auto-pause` the change doesn't immediately take effect and `systemctl daemon-reload` & `systemctl restart auto-pause.service` need to be run. Calling `sysinit.New(co.CP.Runner).Restart("auto-pause")` would fix it, but the messy part is getting `co` which you would normally get via `co := mustload.Running(profile)`. If we do implement it that way then `minikube addons configure auto-pause` with return an error when trying to execute the command when the cluster is stopped. But it gets confusing as the new interval value would actually be updated in the config, it's just the `mustload.Running(profile)` that failing.

We should only call `mustload.Running(profile)` after we confirm that the cluster is running, otherwise just update the value in the config.

PR #23083, opened 2026-06-01
fix: stream daemon images to disk to avoid OOM on image load

### LINKED ISSUE #17785: Excessive memory usage in `minikube image load`
### What Happened?

```console
$ docker pull cvat/server:v2.9.2
[... snipped ...]
$ minikube --alsologtostderr image load cvat/server:v2.9.2
I1213 17:52:14.641171 1485897 out.go:296] Setting OutFile to fd 1 ...
I1213 17:52:14.647657 1485897 out.go:309] Setting ErrFile to fd 2...
I1213 17:52:14.647834 1485897 root.go:338] Updating PATH: /home/dpb/.minikube/bin
I1213 17:52:14.648351 1485897 config.go:182] Loaded profile config "minikube": Driver=docker, ContainerRuntime=docker, KubernetesVersion=v1.28.3
I1213 17:52:14.648471 1485897 cache.go:107] acquiring lock: {Name:mke937cece9923eb59ace0008a0ae80e0c33e4c8 Clock:{} Delay:500ms Timeout:10m0s Cancel:<nil>}
I1213 17:52:14.648854 1485897 image.go:134] retrieving image: cvat/server:v2.9.2
I1213 17:52:14.648908 1485897 image.go:140] checking repository: index.docker.io/cvat/server
I1213 17:52:15.770377 1485897 image.go:144] canonical name: docker.io/cvat/server:v2.9.2
I1213 17:52:15.784553 1485897 image.go:173] found index.docker.io/cvat/server:v2.9.2 locally: &{ref:{Repository:{Registry:{insecure:false registry:index.docker.io} repository:cvat/server} tag:v2.9.2 original:cvat/server:v2.9.2} opener:0xc0003fd810 tarballImage:<nil> computed:false id:0xc0023cfcc0 configFile:<nil> once:{done:0 m:{state:0 sema:0}} err:<nil>}
I1213 17:52:15.784656 1485897 cache.go:162] opening:  /home/dpb/.minikube/cache/images/amd64/cvat/server_v2.9.2
Killed
```

Looking at dmesg output, I see this:

```
[246684.160979] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/,task=minikube,pid=1463487,uid=1000
[246684.161006] Out of memory: Killed process 1463487 (minikube) total-vm:11663440kB, anon-rss:5821500kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:11772kB oom_score_adj:0
```

It seems like `minikube` is using an excessive amount of memory, more even than the size of the image itself (which is under 3 GB).

### Attach the log file

[log.txt](https://github.com/kubernetes/minikube/files/13662883/log.txt)

### Operating System

Ubuntu

### Driver

Docker

Victorthedev:
minikube image load <image> was buffering the entire docker save output into a []byte via go-containerregistry's buffered daemon opener (buffered: true by default), causing peak heap usage of 6x or more the image size. A 3 GB image could consume 18+ GB of RAM, triggering OOM kills.

Added streamImageFromDaemon which pipes docker save directly to a temp file on disk via io.Copy and atomically renames it into the cache. saveToTarFile now takes this path first when useDaemon=true, falling through to the existing remote retrieval path only if the image is not found locally.

Measured on redis:7 (47 MB tar):
  before: 289.8 MB heap, 23.6 s
  after:    0.2 MB heap,  9.4 s

The workaround reported in the issue (docker save | minikube image load file.tar) worked for the same reason, this fix makes that the default behavior for the daemon path.

fixes #17785

### DIFF
--- a/pkg/minikube/image/cache.go
+++ b/pkg/minikube/image/cache.go
@@ -129,6 +129,18 @@ func saveToTarFile(iname, rawDest string, overwrite bool) error {
 		return fmt.Errorf("nil reference for %s: %w", iname, err)
 	}
 
+	// Stream directly from the local daemon to disk, avoiding loading the full image into RAM. Falls through to the normal retrieval path if the image is not present in the daemon.
+	if useDaemon {
+		if err := streamImageFromDaemon(ref, dst); err == nil {
+			return nil
+		} else {
+			klog.Infof("failed to stream image %q from daemon: %v", ref, err)
+			if !useRemote {
+				return errCacheImageDoesntExist
+			}
+		}
+	}
+
 	img, cname, err := retrieveImage(ref, iname)
 	if err != nil {
 		klog.V(2).ErrorS(err, "an error while retrieving the image")
--- a/pkg/minikube/image/cache.go
+++ b/pkg/minikube/image/cache.go
@@ -135,6 +135,9 @@ func saveToTarFile(iname, rawDest string, overwrite bool) error {
 		if streamErr == nil {
 			return nil
 		}
+		// errNotInDaemon means the image is simply not local; fall through to the remote path.
+		// Any other error means the image was confirmed present but streaming failed (e.g. a
+		// disk full or rename error), so surface it rather than silently retrying via the remote registry.
 		klog.Infof("failed to stream image %q from daemon: %v", ref, streamErr)
 		if !errors.Is(streamErr, errNotInDaemon) {
 			// Real I/O failure after the image was confirmed present — surface it rather than
--- a/pkg/minikube/image/cache.go
+++ b/pkg/minikube/image/cache.go
@@ -129,15 +129,20 @@ func saveToTarFile(iname, rawDest string, overwrite bool) error {
 		return fmt.Errorf("nil reference for %s: %w", iname, err)
 	}
 
-	// Stream directly from the local daemon to disk, avoiding loading the full image into RAM. Falls through to the normal retrieval path if the image is not present in the daemon.
+	// Stream directly from the local daemon to disk, avoiding loading the full image into RAM. Falls through to the remote path only when the image is not found in the daemon.
 	if useDaemon {
-		if err := streamImageFromDaemon(ref, dst); err == nil {
+		streamErr := streamImageFromDaemon(ref, dst)
+		if streamErr == nil {
 			return nil
-		} else {
-			klog.Infof("failed to stream image %q from daemon: %v", ref, err)
-			if !useRemote {
-				return errCacheImageDoesntExist
-			}
+		}
+		klog.Infof("failed to stream image %q from daemon: %v", ref, streamErr)
+		if !errors.Is(streamErr, errNotInDaemon) {
+			// Real I/O failure after the image was confirmed present — surface it rather than
+			// pretending the image doesn't exist.
+			return streamErr
+		}
+		if !useRemote {
+			return errCacheImageDoesntExist
 		}
 	}
 
--- a/pkg/minikube/image/daemon_memory_test.go
+++ b/pkg/minikube/image/daemon_memory_test.go
@@ -1,0 +1,146 @@
+/*
+Copyright 2024 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package image
+
+import (
+	"io"
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+)
+
+// TestSaveToTarFileMemory measures peak heap usage when saving a daemon image to cache.
+// Before the fix, this allocates roughly the full image size in RAM (buffered daemon read).
+// After the fix, allocations should be near zero (streaming directly to disk).
+//
+// Requires Docker daemon with redis:7 pulled locally:
+//
+//	docker pull redis:7
+func TestSaveToTarFileMemory(t *testing.T) {
+	if _, err := os.Stat("/var/run/docker.sock"); err != nil {
+		if _, err2 := os.Stat(os.ExpandEnv("$HOME/.docker/run/docker.sock")); err2 != nil {
+			t.Skip("Docker daemon not available")
+		}
+	}
+
+	UseDaemon(true)
+	UseRemote(false)
+	t.Cleanup(func() {
+		UseDaemon(true)
+		UseRemote(true)
+	})
+
+	tmpDir := t.TempDir()
+	dst := filepath.Join(tmpDir, "redis_7")
+
+	runtime.GC()
+	var before, after runtime.MemStats
+	runtime.ReadMemStats(&before)
+
+	if err := saveToTarFile("redis:7", dst, true); err != nil {
+		t.Fatalf("saveToTarFile: %v", err)
+	}
+
+	runtime.GC()
+	runtime.ReadMemStats(&after)
+
+	fi, err := os.Stat(dst)
+	if err != nil {
+		t.Fatalf("output file missing: %v", err)
+	}
+
+	heapAllocMB := float64(after.TotalAlloc-before.TotalAlloc) / 1024 / 1024
+	fileSizeMB := float64(fi.Size()) / 1024 / 1024
+
+	t.Logf("Image size on disk:  %.1f MB", fileSizeMB)
+	t.Logf("Heap allocated:      %.1f MB", heapAllocMB)
+	t.Logf("Ratio (heap/disk):   %.2fx", heapAllocMB/fileSizeMB)
+
+	// After the fix the heap allocation should be a small fraction of the image size.
+	// Before the fix it will be >= image size (the entire image is buffered in []byte).
+	if heapAllocMB >= fileSizeMB*0.5 {
+		t.Errorf("EXCESSIVE MEMORY: allocated %.1f MB for a %.1f MB image — full image is being buffered in RAM (bug #17945)", heapAllocMB, fileSizeMB)
+	}
+}
+
+// TestStreamedTarIsValidDockerFormat verifies that the output of streamImageFromDaemon
+// (raw docker save output) is a valid Docker image tar that minikube's downstream
+// consumers — tarball.LoadManifest and transferAndLoadImage — can read correctly.
+// This matters because the temp-dir path skips go-containerregistry's tarball.Write
+// and writes docker save output directly.
+//
+// Requires Docker daemon with redis:7 pulled locally.
+func TestStreamedTarIsValidDockerFormat(t *testing.T) {
+	if _, err := os.Stat("/var/run/docker.sock"); err != nil {
+		if _, err2 := os.Stat(os.ExpandEnv("$HOME/.docker/run/docker.sock")); err2 != nil {
+			t.Skip("Docker daemon not available")
+		}
+	}
+
+	UseDaemon(true)
+	UseRemote(false)
+	t.Cleanup(func() {
+		UseDaemon(true)
+		UseRemote(true)
+	})
+
+	tmpDir := t.TempDir()
+	dst := filepath.Join(tmpDir, "redis_7")
+
+	if err := saveToTarFile("redis:7", dst, true); err != nil {
+		t.Fatalf("saveToTarFile: %v", err)
+	}
+
+	manifest, err := tarball.LoadManifest(func() (io.ReadCloser, error) {
+		return os.Open(dst)
+	})
+	if err != nil {
+		t.Fatalf("output is not a valid Docker tar: %v", err)
+	}
+	if len(manifest) == 0 {
+		t.Fatal("manifest is empty")
+	}
+	if len(manifest[0].RepoTags) == 0 {
+		t.Fatal("manifest has no repo tags")
+	}
+	t.Logf("manifest repo tags: %v", manifest[0].RepoTags)
+}
+
+func BenchmarkSaveToTarFile_Daemon(b *testing.B) {
+	if _, err := os.Stat("/var/run/docker.sock"); err != nil {
+		if _, err2 := os.Stat(os.ExpandEnv("$HOME/.docker/run/docker.sock")); err2 != nil {
+			b.Skip("Docker daemon not available")
+		}
+	}
+
+	UseDaemon(true)
+	UseRemote(false)
+	b.Cleanup(func() {
+		UseDaemon(true)
+		UseRemote(true)
+	})
+
+	for b.Loop() {
+		dst := filepath.Join(b.TempDir(), "redis_7")
+		if err := saveToTarFile("redis:7", dst, true); err != nil {
+			b.Fatalf("saveToTarFile: %v", err)
+		}
+	}
+}
--- a/pkg/minikube/image/daemon_memory_test.go
+++ b/pkg/minikube/image/daemon_memory_test.go
@@ -83,7 +83,7 @@ func TestSaveToTarFileMemory(t *testing.T) {
 	// After the fix the heap allocation should be a small fraction of the image size.
 	// Before the fix it will be >= image size (the entire image is buffered in []byte).
 	if heapAllocMB >= fileSizeMB*0.5 {
-		t.Errorf("EXCESSIVE MEMORY: allocated %.1f MB for a %.1f MB image — full image is being buffered in RAM (issue #17785)", heapAllocMB, fileSizeMB)
+		t.Errorf("EXCESSIVE MEMORY: allocated %.1f MB for a %.1f MB image — full image is being buffered in RAM (https://github.com/kubernetes/minikube/issues/17785)", heapAllocMB, fileSizeMB)
 	}
 }
 
--- a/pkg/minikube/image/daemon_memory_test.go
+++ b/pkg/minikube/image/daemon_memory_test.go
@@ -83,7 +83,7 @@ func TestSaveToTarFileMemory(t *testing.T) {
 	// After the fix the heap allocation should be a small fraction of the image size.
 	// Before the fix it will be >= image size (the entire image is buffered in []byte).
 	if heapAllocMB >= fileSizeMB*0.5 {
-		t.Errorf("EXCESSIVE MEMORY: allocated %.1f MB for a %.1f MB image — full image is being buffered in RAM (issues #17785/#17945)", heapAllocMB, fileSizeMB)
+		t.Errorf("EXCESSIVE MEMORY: allocated %.1f MB for a %.1f MB image — full image is being buffered in RAM (issue #17785)", heapAllocMB, fileSizeMB)
 	}
 }
 
--- a/pkg/minikube/image/daemon_memory_test.go
+++ b/pkg/minikube/image/daemon_memory_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2024 The Kubernetes Authors All rights reserved.
+Copyright 2026 The Kubernetes Authors All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -19,12 +19,23 @@ package image
 import (
 	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"testing"
 
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 )
+
+// requireDockerImage skips the test if the named image is not present in the local daemon.
+// This handles all Docker configurations (regular, rootless, remote via $DOCKER_HOST) because
+// it delegates to the docker CLI which reads $DOCKER_HOST and other env vars.
+func requireDockerImage(t *testing.T, img string) {
+	t.Helper()
+	if err := exec.Command("docker", "inspect", "--type=image", img).Run(); err != nil {
+		t.Skipf("image %s not in local daemon (run: docker pull %s)", img, img)
+	}
+}
 
 // TestSaveToTarFileMemory measures peak heap usage when saving a daemon image to cache.
 // Before the fix, this allocates roughly the full image size in RAM (buffered daemon read).
@@ -34,11 +45,7 @@ import (
 //
 //	docker pull redis:7
 func TestSaveToTarFileMemory(t *testing.T) {
-	if _, err := os.Stat("/var/run/docker.sock"); err != nil {
-		if _, err2 := os.Stat(os.ExpandEnv("$HOME/.docker/run/docker.sock")); err2 != nil {
-			t.Skip("Docker daemon not available")
-		}
-	}
+	requireDockerImage(t, "redis:7")
 
 	UseDaemon(true)
 	UseRemote(false)
@@ -76,7 +83,7 @@ func TestSaveToTarFileMemory(t *testing.T) {
 	// After the fix the heap allocation should be a small fraction of the image size.
 	// Before the fix it will be >= image size (the entire image is buffered in []byte).
 	if heapAllocMB >= fileSizeMB*0.5 {
-		t.Errorf("EXCESSIVE MEMORY: allocated %.1f MB for a %.1f MB image — full image is being buffered in RAM (bug #17945)", heapAllocMB, fileSizeMB)
+		t.Errorf("EXCESSIVE MEMORY: allocated %.1f MB for a %.1f MB image — full image is being buffered in RAM (issues #17785/#17945)", heapAllocMB, fileSizeMB)
 	}
 }
 
@@ -88,11 +95,7 @@ func TestSaveToTarFileMemory(t *testing.T) {
 //
 // Requires Docker daemon with redis:7 pulled locally.
 func TestStreamedTarIsValidDockerFormat(t *testing.T) {
-	if _, err := os.Stat("/var/run/docker.sock"); err != nil {
-		if _, err2 := os.Stat(os.ExpandEnv("$HOME/.docker/run/docker.sock")); err2 != nil {
-			t.Skip("Docker daemon not available")
-		}
-	}
+	requireDockerImage(t, "redis:7")
 
 	UseDaemon(true)
 	UseRemote(false)
@@ -124,10 +127,8 @@ func TestStreamedTarIsValidDockerFormat(t *testing.T) {
 }
 
 func BenchmarkSaveToTarFile_Daemon(b *testing.B) {
-	if _, err := os.Stat("/var/run/docker.sock"); err != nil {
-		if _, err2 := os.Stat(os.ExpandEnv("$HOME/.docker/run/docker.sock")); err2 != nil {
-			b.Skip("Docker daemon not available")
-		}
+	if err := exec.Command("docker", "inspect", "--type=image", "redis:7").Run(); err != nil {
+		b.Skipf("image redis:7 not in local daemon (run: docker pull redis:7)")
 	}
 
 	UseDaemon(true)
--- a/pkg/minikube/image/image.go
+++ b/pkg/minikube/image/image.go
@@ -189,6 +189,52 @@ func retrieveDaemon(ref name.Reference) (v1.Image, error) {
 	return img, err
 }
 
+// streamImageFromDaemon saves a local Docker daemon image directly to dst by piping docker save output to disk, avoiding buffering the entire image in RAM.
+// Returns an error if the daemon is unavailable or the image does not exist locally.
+func streamImageFromDaemon(ref name.Reference, dst string) error {
+	cl, err := client.NewClientWithOpts(client.FromEnv)
+	if err != nil {
+		return err
+	}
+	ctx := context.Background()
+	cl.NegotiateAPIVersion(ctx)
+
+	// Confirm the image exists before creating temp files.
+	if _, _, err := cl.ImageInspectWithRaw(ctx, ref.Name()); err != nil {
+		return err
+	}
+
+	rc, err := cl.ImageSave(ctx, []string{ref.Name()})
+	if err != nil {
+		return err
+	}
+	defer rc.Close()
+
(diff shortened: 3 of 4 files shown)


nirs:
@Victorthedev Thanks for contributing!

Saving the image to file when loading image from daemon works and may be good enough - we basically do the same around minkkube by saving to file and loading the file. But I'm worried about "saving to cache" - loading an image should not add it to the minikube cache to avoid loading the image later to unrelated clusters.

If we want to go in this way, we should:

- Create temporary directory
- Save image to temporary directory
- Load image from file
- Delete the temporary directory

But it will be even better to not use temporary files. Do we have a way to write the image as tar file to stdout? Then pipe data from the daemon to minikube - something like:

    docker save | minikube ssh -- bash -c "cat > /tmp/image.tar"

This should perform little better and avoid the need for temporary files on the host.

If this is not possible or require much more effort we can go with the simple way.

Another issue - what about remote images - not pull to the docker/podman daemon. Do we buffer these images in memory? Ideally we should solve this issue for all image types. We can also open another issue of this, we cannot fix the entire world at once.

Victorthedev:
> @Victorthedev Thanks for contributing!
> 
> Saving the image to file when loading image from daemon works and may be good enough - we basically do the same around minkkube by saving to file and loading the file. But I'm worried about "saving to cache" - loading an image should not add it to the minikube cache to avoid loading the image later to unrelated clusters.
> 
> If we want to go in this way, we should:
> 
> * Create temporary directory
> * Save image to temporary directory
> * Load image from file
> * Delete the temporary directory
> 
> But it will be even better to not use temporary files. Do we have a way to write the image as tar file to stdout? Then pipe data from the daemon to minikube - something like:
> 
> ```
> docker save | minikube ssh -- bash -c "cat > /tmp/image.tar"
> ```
> 
> This should perform little better and avoid the need for temporary files on the host.
> 
> If this is not possible or require much more effort we can go with the simple way.
> 
> Another issue - what about remote images - not pull to the docker/podman daemon. Do we buffer these images in memory? Ideally we should solve this issue for all image types. We can also open another issue of this, we cannot fix the entire world at once.

Thanks for the review

I see your point, minikube image load shouldn't persist to `~/.minikube/cache/images/` which may affect future clusters. I'll switch to the temp dir approach which saves to a temporary directory, loads into the cluster then delete. The `streamImageFromDaemon` streaming function stays the same, just to change where it saves.

But about piping, I looked into this and `SSHRunner.RunCmd` currently rejects stdin explicitly (SSHRunner does not support stdin) so `docker save | ssh ... docker load` would require adding stdin support to the SSH runner plus updating all three container runtime `LoadImage` implementations (Docker, containerd and CRI-O). I think that should be a separate issue. I could open the issue for it if you agree.

On remote image, yes, they're also buffered, `retrieveRemote` + `tarball.Write` pulls the whole image through go-containerregistry into memory the same way. I agree this is a separate issue too

I'll update the PR with the temp dir approach.

Victorthedev:
> > @Victorthedev Thanks for contributing!
> > Saving the image to file when loading image from daemon works and may be good enough - we basically do the same around minkkube by saving to file and loading the file. But I'm worried about "saving to cache" - loading an image should not add it to the minikube cache to avoid loading the image later to unrelated clusters.
> > If we want to go in this way, we should:
> > 
> > * Create temporary directory
> > * Save image to temporary directory
> > * Load image from file
> > * Delete the temporary directory
> > 
> > But it will be even better to not use temporary files. Do we have a way to write the image as tar file to stdout? Then pipe data from the daemon to minikube - something like:
> > ```
> > docker save | minikube ssh -- bash -c "cat > /tmp/image.tar"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This should perform little better and avoid the need for temporary files on the host.
> > If this is not possible or require much more effort we can go with the simple way.
> > Another issue - what about remote images - not pull to the docker/podman daemon. Do we buffer these images in memory? Ideally we should solve this issue for all image types. We can also open another issue of this, we cannot fix the entire world at once.
> 
> Thanks for the review
> 
> I see your point, minikube image load shouldn't persist to `~/.minikube/cache/images/` which may affect future clusters. I'll switch to the temp dir approach which saves to a temporary directory, loads into the cluster then delete. The `streamImageFromDaemon` streaming function stays the same, just to change where it saves.
> 
> But about piping, I looked into this and `SSHRunner.RunCmd` currently rejects stdin explicitly (SSHRunner does not support stdin) so `docker save | ssh ... docker load` would require adding stdin support to the SSH runner plus updating all three container runtime `LoadImage` implementations (Docker, containerd and CRI-O). I think that should be a separate issue. I could open the issue for it if you agree.
> 
> On remote image, yes, they're also buffered, `retrieveRemote` + `tarball.Write` pulls the whole image through go-containerregistry into memory the same way. I agree this is a separate issue too
> 
> I'll update the PR with the temp dir approach.

@nirs

nirs:
@Victorthedev I'll try to review this weekend. Please squash the commits unless there reason to have separate commits (more than one logical change, splitting refactoring from actual fix, etc).

pkg/minikube/image/cache.go
```diff
@@ -130,6 +130,19 @@ func saveToTarFile(iname, rawDest string, overwrite bool) error {
 		return fmt.Errorf("nil reference for %s: %w", iname, err)
 	}
 
+	// Stream directly from the local daemon to disk, avoiding loading the full image into RAM. Falls through to the normal retrieval path if the image is not present in the daemon.
+	if useDaemon {
+		if err := streamImageFromDaemon(ref, dst); err == nil {
+			klog.Infof("%s exists", dst)
```
nirs: What is the purpose of this log? We don't log successful operation in streamImageFromDaemon()?

pkg/minikube/image/cache.go
```diff
@@ -130,6 +130,19 @@ func saveToTarFile(iname, rawDest string, overwrite bool) error {
...
+			return nil
+		} else {
+			klog.Infof("daemon stream not available for %q: %v", ref, err)
```
nirs: The error is not clear - what is "daemon stream not available"? Maybe "failed to stream image %q from daemon: %v"?

nirs (commented):
Partial review, will continue later.

nirs:
Preparing review...

nirs:
Preparing review...

pkg/minikube/image/cache.go
Victorthedev: The `%s exists` log follows the existing pattern in `saveToTarFile` where the same message is logged after `writeImage` completes successfully. I carried that pattern over for the daemon streaming path. I see now that it's inconsistent to log success in the caller when the work happens inside `streamImageFromDaemon` so I'll move the success log into `streamImageFromDaemon` itself so it sits next to the work it confirms and remove it from the caller.

pkg/minikube/image/cache.go
Victorthedev: Okay, I'll update it now

Victorthedev:
> @Victorthedev I'll try to review this weekend. Please squash the commits unless there reason to have separate commits (more than one logical change, splitting refactoring from actual fix, etc).

Doing this with the next commit

pkg/minikube/image/cache.go
```diff
@@ -129,6 +130,18 @@ func saveToTarFile(iname, rawDest string, overwrite bool) error {
...
 	}
 
+	// Stream directly from the local daemon to disk, avoiding loading the full image into RAM. Falls through to the normal retrieval path if the image is not present in the daemon.
+	if useDaemon {
+		if err := streamImageFromDaemon(ref, dst); err == nil {
+			return nil
+		} else {
+			klog.Infof("failed to stream image %q from daemon: %v", ref, err)
+			if !useRemote {
+				return errCacheImageDoesntExist
```
nirs: This returns an error only if useRemote is false. If useRemote is true we ignore the failure silently and continue. Can you explain this behavior?

What is the expected behavior when useDaemon and useRemote are true? Are these compatible options?

In general error handling should use the happy path pattern:

```go
if err := streamImageFromDaemon(ref, dst); err != nil {
    return err
}

// normal code path...
```

If we need to return early, the code should be refactored so we can continue to use the happy path pattern.

Can you explain why this change belongs to saveTarFile? The comments says:

```
// saveToTarFile caches an image
```

Which is alarming - how saving tar file is related to caching? Is this used in `minikube image load` command?

pkg/minikube/image/daemon_memory_test.go
```diff
@@ -0,0 +1,146 @@
+/*
+Copyright 2024 The Kubernetes Authors All rights reserved.
```
Roslaan001: Isn't the copyright year supposed to be 2026?

pkg/minikube/image/daemon_memory_test.go
```diff
@@ -0,0 +1,146 @@
...
+	fileSizeMB := float64(fi.Size()) / 1024 / 1024
+
+	t.Logf("Image size on disk:  %.1f MB", fileSizeMB)
+	t.Logf("Heap allocated:      %.1f MB", heapAllocMB)
+	t.Logf("Ratio (heap/disk):   %.2fx", heapAllocMB/fileSizeMB)
+
+	// After the fix the heap allocation should be a small fraction of the image size.
+	// Before the fix it will be >= image size (the entire image is buffered in []byte).
+	if heapAllocMB >= fileSizeMB*0.5 {
+		t.Errorf("EXCESSIVE MEMORY: allocated %.1f MB for a %.1f MB image — full image is being buffered in RAM (bug #17945)", heapAllocMB, fileSizeMB)
```
Roslaan001: The message says "bug #17945", but this PR fixes #17785, right?

pkg/minikube/image/daemon_memory_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+// TestSaveToTarFileMemory measures peak heap usage when saving a daemon image to cache.
+// Before the fix, this allocates roughly the full image size in RAM (buffered daemon read).
+// After the fix, allocations should be near zero (streaming directly to disk).
+//
+// Requires Docker daemon with redis:7 pulled locally:
+//
+//	docker pull redis:7
+func TestSaveToTarFileMemory(t *testing.T) {
+	if _, err := os.Stat("/var/run/docker.sock"); err != nil {
```
Roslaan001: I believe these tests skip when Docker is absent but fail on machines that have Docker without redis:7 pulled, since UseRemote(false) prevents a pull. Maybe pull the image in the test setup, or skip on not-found. (This check also misses rootless Docker at $XDG_RUNTIME_DIR/docker.sock.)

cc @nirs

pkg/minikube/image/daemon_memory_test.go
```diff
@@ -0,0 +1,146 @@
...
+	heapAllocMB := float64(after.TotalAlloc-before.TotalAlloc) / 1024 / 1024
+	fileSizeMB := float64(fi.Size()) / 1024 / 1024
+
+	t.Logf("Image size on disk:  %.1f MB", fileSizeMB)
+	t.Logf("Heap allocated:      %.1f MB", heapAllocMB)
+	t.Logf("Ratio (heap/disk):   %.2fx", heapAllocMB/fileSizeMB)
+
+	// After the fix the heap allocation should be a small fraction of the image size.
+	// Before the fix it will be >= image size (the entire image is buffered in []byte).
+	if heapAllocMB >= fileSizeMB*0.5 {
```
Roslaan001, line 85: Since heap-ratio thresholds can vary due to GC timing and runtime changes, im not sure it's a reliable unit test assertion. Maybe consider recording the measurement with t.Logf or as a benchmark, and probably use `TestStreamedTarIsValidDockerFormat` as the gating test instead.

cc @nirs

pkg/minikube/image/cache.go
```diff
@@ -129,6 +130,18 @@ func saveToTarFile(iname, rawDest string, overwrite bool) error {
 		return fmt.Errorf("nil reference for %s: %w", iname, err)
 	}
 
+	// Stream directly from the local daemon to disk, avoiding loading the full image into RAM. Falls through to the normal retrieval path if the image is not present in the daemon.
+	if useDaemon {
+		if err := streamImageFromDaemon(ref, dst); err == nil {
+			return nil
+		} else {
+			klog.Infof("failed to stream image %q from daemon: %v", ref, err)
```
Roslaan001: If streaming fails with the `useRemote` enabled, the current fallback invokes `retrieveImage`, which retries the daemon through the buffered `go-containerregistry` path. That actually reintroduces the OOM path this PR is trying to fix. It might be better to just bypass the daemon entirely in this scenario and fall back directly to the remote registry. 

cc @nirs

pkg/minikube/image/cache.go
```diff
@@ -129,6 +130,18 @@ func saveToTarFile(iname, rawDest string, overwrite bool) error {
 		return fmt.Errorf("nil reference for %s: %w", iname, err)
 	}
 
+	// Stream directly from the local daemon to disk, avoiding loading the full image into RAM. Falls through to the normal retrieval path if the image is not present in the daemon.
+	if useDaemon {
+		if err := streamImageFromDaemon(ref, dst); err == nil {
+			return nil
+		} else {
```
Roslaan001: The `else` after a `return` may be unnecessary and will likely be flagged by the linter. Remove the else and dedent (reducing the indentation level) the subsequent code

pkg/minikube/image/image.go
```diff
@@ -183,6 +183,52 @@ func retrieveDaemon(ref name.Reference) (v1.Image, error) {
 	return img, err
 }
 
+// streamImageFromDaemon saves a local Docker daemon image directly to dst by piping docker save output to disk, avoiding buffering the entire image in RAM.
+// Returns an error if the daemon is unavailable or the image does not exist locally.
+func streamImageFromDaemon(ref name.Reference, dst string) error {
+	cl, err := client.NewClientWithOpts(client.FromEnv)
```
Roslaan001, line 200: It looks like the client isn't being closed. Adding `defer cl.Close()` right after the error check would ensure it's cleaned up before the function exits.

cc @nirs

pkg/minikube/image/image.go
```diff
@@ -183,6 +183,52 @@ func retrieveDaemon(ref name.Reference) (v1.Image, error) {
...
+	if err != nil {
+		return err
+	}
+	ctx := context.Background()
+	cl.NegotiateAPIVersion(ctx)
+
+	// Confirm the image exists before creating temp files.
+	if _, _, err := cl.ImageInspectWithRaw(ctx, ref.Name()); err != nil {
```
Roslaan001: `ImageInspectWithRaw` is deprecated and has been fully removed from the official Docker Go SDK. This file already uses `ImageInspect` elsewhere, so it would be good to keep things consistent. Also, `context.Background()` without a timeout may cause the inspect call to hang for minutes when offline. A short timeout (e.g., the 10-second timeout used in `DigestByDockerLib`) will match the existing pattern, while still allowing the save operation to run without a timeout.

cc @nirs

pkg/minikube/machine/cache_images.go
```diff
@@ -191,12 +193,17 @@ func CacheAndLoadImages(images []string, profiles []*config.Profile, overwrite b
 		return nil
 	}
 
-	// This is the most important thing
-	if err := image.SaveToDir(images, detect.ImageCacheDir(), overwrite); err != nil {
+	tmpDir, err := os.MkdirTemp("", "minikube-image-*")
```
Roslaan001: This change appears to me to break the persistent `image cache` and should **likely** be **reverted**. `CacheAndLoadImagesInConfig` (used by `minikube start` and `cache reload`) flows through this code path. Today, it short-circuits when images already exist in `detect.ImageCacheDir()`, but using a fresh temporary directory causes every configured image to be `re-fetched` on each start and prevents offline loads from a populated cache directory. It also ignores the overwrite parameter by always passing `true`. The memory fix may not require this change `streamImageFromDaemon` already streams directly to the provided destination, including the cache path.

@nirs Please check this out

pkg/minikube/image/image.go
```diff
@@ -183,6 +183,52 @@ func retrieveDaemon(ref name.Reference) (v1.Image, error) {
...
+		return err
+	}
+
+	rc, err := cl.ImageSave(ctx, []string{ref.Name()})
```
Roslaan001: This changes the tag semantics compared to the previous path.` tarball.Write `preserved the canonical image name (e.g. `docker.io/library/redis:7`), while docker save writes Docker's normalised `RepoTags` (e.g. redis:7). For digest references, `RepoTags` is `null`, so the image is loaded untagged. Have the `containerd``/CRI-O `load paths and digest-reference scenarios been tested with this change?

Roslaan001:
@Victorthedev, Very impressive work on this one , #17785 has been open for a while now and streaming Docker save straight to disk is such a very good way to solve it. The 289 MB → 0.2 MB numbers are really impressive, and I liked that you reused the temp-file + rename pattern from `writeImage` and added a test that checks the tar format, not just the memory.

I was reading through the diff and left some comments as I went, mostly questions and small things I noticed. The one I'm most unsure about is the `CacheAndLoadImages` `temp-dir` change: it looks like it might affect the persistent `cache` and offline starts, but maybe I'm missing the reason for it? Curious what you think. Thank you

Roslaan001:
@nirs Review done. Please also check it out

Roslaan001:
@nirs, you might have missed this too. I reviewed this PR; there is another I reviewed. I will tag you to it

pkg/minikube/image/cache.go
```diff
@@ -130,6 +130,18 @@ func saveToTarFile(iname, rawDest string, overwrite bool) error {
...
+			klog.Infof("failed to stream image %q from daemon: %v", ref, err)
+			if !useRemote {
+				return errCacheImageDoesntExist
```
nitishagar: When `streamImageFromDaemon` fails *after* the image was confirmed present (e.g. `os.Rename` fails, or `io.Copy` hits a disk error), the daemon-only path translates any streaming failure into `errCacheImageDoesntExist`, telling the user the image is absent when it is actually present in the daemon. It may be worth distinguishing a not-found inspect error from a later I/O/rename error: only treat a not-found from `ImageInspectWithRaw` as `errCacheImageDoesntExist`, and surface `io.Copy`/`os.Rename` errors as a real failure so the user sees the actual cause.

pkg/minikube/image/image.go
```diff
@@ -183,6 +183,52 @@ func retrieveDaemon(ref name.Reference) (v1.Image, error) {
...
+		tmp.Close()
+		return err
+	}
+	if err = tmp.Close(); err != nil {
+		return err
+	}
+	if err = os.Rename(tmpName, dst); err != nil {
+		return err
+	}
+	klog.Infof("streamed %s from daemon to %s", ref.Name(), dst)
```
nitishagar, line 247: On the `minikube image load` path, the streamed-to path logged here is a file inside the temp dir that `CacheAndLoadImages` removes as soon as the function returns, so logs reference a path that no longer exists by the time anyone reads them. Consider logging the image ref and the fact that it was streamed (e.g. `streamed %s from daemon to temp file for loading`) instead of printing the ephemeral temp-dir destination, or moving this log into `CacheAndLoadImages` after the load completes so it reflects a durable outcome.

nitishagar (commented):
The streaming approach is sound: the temp file is created in the same dir as the destination (so `os.Rename` is same-filesystem/atomic), partial writes leave a temp file that is removed on error (never a corrupt destination), and the daemon path no longer buffers the image in RAM. The highest-impact risks here (the `CacheAndLoadImages` temp-dir change regressing offline cache reuse, the deprecated `ImageInspect` API, client close, `RepoTags` semantics, test reliability) are already covered by another reviewer — these three are the remaining error-handling/observability gaps.

pkg/minikube/image/image.go
```diff
@@ -183,6 +183,52 @@ func retrieveDaemon(ref name.Reference) (v1.Image, error) {
...
+		return err
+	}
+	tmpName := tmp.Name()
+	defer func() {
+		if _, statErr := os.Stat(tmpName); statErr == nil {
+			os.Remove(tmpName)
+		}
+	}()
+
+	if _, err = io.Copy(tmp, rc); err != nil {
```
nitishagar, line 235: On a mid-stream `io.Copy` error, the daemon's `ImageSave` response body is dropped via `defer rc.Close()` without being read to completion (the success path drains to EOF via `io.Copy` before close). This can leave the daemon-side export cancelled mid-flight and produce daemon-side errors for what is otherwise a handled client condition. On the error branch it may be worth draining the remainder of `rc` (e.g. `io.Copy(io.Discard, rc)`) before returning so the daemon-side stream completes cleanly.

pkg/minikube/image/daemon_memory_test.go
Victorthedev: Fixed, thanks for pointing out

pkg/minikube/image/daemon_memory_test.go
Victorthedev: Fixed, updated to reference #17785 only.

pkg/minikube/image/daemon_memory_test.go
Victorthedev: Fixed. Added a `requireDockerImage` helper that calls `docker inspect --type=image ` and skips with a clear message if the image is absent. Using the docker CLI directly means it respects `$DOCKER_HOST`, so it works for regular, rootless and remote Docker configurations without needing to enumerate socket paths.

pkg/minikube/image/daemon_memory_test.go
```diff
@@ -83,7 +83,7 @@ func TestSaveToTarFileMemory(t *testing.T) {
 	// After the fix the heap allocation should be a small fraction of the image size.
 	// Before the fix it will be >= image size (the entire image is buffered in []byte).
 	if heapAllocMB >= fileSizeMB*0.5 {
-		t.Errorf("EXCESSIVE MEMORY: allocated %.1f MB for a %.1f MB image — full image is being buffered in RAM (issues #17785/#17945)", heapAllocMB, fileSizeMB)
+		t.Errorf("EXCESSIVE MEMORY: allocated %.1f MB for a %.1f MB image — full image is being buffered in RAM (issue #17785)", heapAllocMB, fileSizeMB)
```
nirs: Use a full URL - editors and terminals do not how to open `#17785`. The shortcut is useful only in github comments.

nirs (commented):
I did not have time to review the change yet. Add comment about the error message.

pkg/minikube/image/daemon_memory_test.go
Victorthedev: I kept the assertion and want to explain why it's reliable here. The threshold is 50% of image size, for redis:7 (47MB) that's ~23.5MB. After the fix it measured 0.2MB. The regression being guarded against is `daemon.Image()` with `buffered: true`, which buffers the entire image into RAM. For redis:7, that's ~280MB. The gap between the bug being present (~280MB) and the bug absent (~0.2MB) is three orders of magnitude. GC timing would need to add ~100x overhead on top of the fixed path to trigger a false positive, which isn't realistic. The threshold is deliberately wide precisely to make it immune to runtime variance while still catching the actual regression. The raw numbers are logged via `t.Logf` and `BenchmarkSaveToTarFile_Daemon` is available for microbenchmarking. We should keep the assertion as the regression guard.

pkg/minikube/image/cache.go
Victorthedev: Fixed, removed the else and dedented.

pkg/minikube/image/cache.go
Victorthedev: Thank you for pointing out and fixed. I introduced an `errNotInDaemon` sentinel: `streamImageFromDaemon` returns it when the image is absent or the daemon is unreachable. `saveToTarFile` checks `errors.Is(streamErr, errNotInDaemon)` only this sentinel triggers fallback to the remote path. Any I/O failure that occurs after the image is confirmed present (disk full, rename failure, etc.) is returned as a real error and never falls through to `retrieveImage`. So the buffered daemon path can no longer be re-entered on the streaming fallback.

pkg/minikube/image/image.go
Victorthedev: Fixed, added defer `cl.Close()` immediately after the nil check.

pkg/minikube/image/image.go
Victorthedev: Fixed both. Replaced with `ImageInspect` and a 10-second timeout context for the inspect call, matching the pattern already used in `DigestByDockerLib`. `ImageSave` gets its own `context.Background()` with no timeout because large images can legitimately take minutes to export and we don't want to cut that off.

pkg/minikube/machine/cache_images.go
Victorthedev: It's reverted. `CacheAndLoadImages` is called from three places: `minikube image load`, `minikube cache add` and `CacheAndLoadImagesInConfig` (which runs on `minikube start`). Using a temp dir broke the persistent cache for all three and silently hardcoded overwrite=true, ignoring the caller's parameter entirely. `streamImageFromDaemon` already streams directly to whatever destination it's given, including the persistent cache path at `detect.ImageCacheDir()` so the fix did not need thus change after all.

pkg/minikube/image/image.go
Victorthedev: Good point. For tagged images, the common case for `minikube image load`. Docker's normalized name and the canonical name are equivalent from the downstream consumers' perspective (`docker load`, `ctr import`, `podman load` all read RepoTags and load successfully). The genuine edge case is a pure digest reference (e.g., `redis@sha256:abc`) with no tag associated in the local daemon: `docker save` would produce a tar with empty RepoTags and the image would load untagged into the runtime. I think a follow-up issue should be opened to track this and add a test for the digest ref path. It's not a regression introduced by this PR (the prior path had its own nuances for digest refs) but it deserves explicit coverage.

pkg/minikube/image/cache.go
Victorthedev: Fixed. `streamImageFromDaemon` now returns `errNotInDaemon` specifically for inspect failures (image absent or daemon unreachable). Any failure that happens after the inspect succeeds for e.g., `io.Copy` errors, temp file creation, rename, returns the real error. `saveToTarFile` uses `errors.Is(streamErr, errNotInDaemon)` to decide: only that sentinel triggers `errCacheImageDoesntExist` or remote fallback. Everything else surfaces as-is.

pkg/minikube/image/image.go
Victorthedev: Fixed. Added `io.Copy(io.Discard, rc)` on the error branch so the daemon-side export stream drains to completion before it returns, rather than being cancelled mid-flight.

pkg/minikube/image/image.go
Victorthedev: Resolved. Reverted the` CacheAndLoadImages` temp-dir change. The `dst` passed to `streamImageFromDaemon` is now the persistent cache path at `detect.ImageCacheDir()`, so the log message "streamed %s from daemon to %s" correctly reflects a durable write.

pkg/minikube/image/daemon_memory_test.go
Victorthedev: done

pkg/minikube/image/cache.go
Victorthedev: They are compatible and the behavior is intentional but the old code was silent about it, fair enough. With the `errNotInDaemon` sentinel introduced in this update, the logic is now explicit:

- `streamImageFromDaemon` returns `errNotInDaemon` when the image is not in the local daemon or the daemon is unreachable. We fall through to `retrieveImage` for the remote path, same intent as the existing `retrieveImage` -> `retrieveDaemon` -> `retrieveRemote` fallback: try local first, use remote if not found locally.
Any failure that occurs after the image is confirmed present (disk write error, rename failure) returns the real error immediately and never falls through.
So we are not silently ignoring failures, we only fall through on the genuine "not found locally" case.



On if general error handling should use happy path pattern, agree on the principle. The complication here is that `streamImageFromDaemon` has three meaningful outcomes; success, image-not-in-daemon and real I/O error and only one of them should fall through to the remote path. A plain `if err != nil { return err }` would prevent us from ever trying remote when the image is simply not local. The `errNotInDaemon` sentinel is the minimal way to express this three-way branch without restructuring `retrieveImage` itself. If you have a preferred shape for this we can discuss it.


`saveToTarFile` is the cache write. The call chain for `minikube image load` is:

[cmd/minikube/cmd/image.go:142](https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/image.go#L142), image load handler calls `CacheAndLoadImages`
[machine/cache_images.go:194](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/machine/cache_images.go#L194), calls `image.SaveToDir(images, detect.ImageCacheDir(), overwrite)`
[image/cache.go:73](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/image/cache.go#L73), `SaveToDir` calls `saveToTarFile` for each image, writing the tar to `detect.ImageCacheDir()`
[machine/cache_images.go:198](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/machine/cache_images.go#L198), then calls `DoLoadImages` with that same cache dir
[machine/cache_images.go:272](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/machine/cache_images.go#L272), `transferAndLoadCachedImage` does `src := filepath.Join(cacheDir, imgName)` and transfers it into the VM runtime
The tar that `saveToTarFile` writes is the cache, it is what gets read back and transferred into the VM. The OOM happened here because` retrieveImage` ([image.go:162](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/image/image.go#L162)) calls `retrieveDaemon` ([image.go:182](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/image/image.go#L182)), which calls `daemon.Image(ref)` with no options and go-containerregistry's makeOptions defaults to [buffered: true](https://github.com/google/go-containerregistry/blob/main/pkg/v1/daemon/options.go#L44), loading the full docker save output into a `[]byte` before anything hits disk. The fix belongs at this level, intercept before `retrieveImage` is called, stream directly to `dst`via `io.Copy` and return early.

Victorthedev:
> I did not have time to review the change yet. Add comment about the error message.

Added a comment above the log line explaining the two-path behaviour. `errNotInDaemon` means the image is simply not local so we fall through to remote as expected. Any other error means streaming failed after the image was confirmed present (e.g. a disk full or rename error) so we surface it rather than silently retrying via the registry. Let me know if you would like it worded differently once you get a chance to review.

PR #23125, opened 2026-06-13
ssh: prefer AES-GCM ciphers for external SSH client

### LINKED ISSUE #23121: Use hardware-accelerated AES-GCM cipher for SSH connections
## Summary

On modern CPUs with hardware AES acceleration, `aes128-gcm@openssh.com` is significantly faster than `chacha20-poly1305@openssh.com` (the OpenSSH default). minikube's external SSH client path should prefer AES-GCM ciphers to speed up all SSH operations.

## Benchmark Results

Tested with the macOS `ssh` command (OpenSSH 10.2) and a 184 MB image (`quay.io/ramendr/ramen-operator`) on minikube with vfkit driver (Apple Silicon M3):

| Test                                                | chacha20 (default)  | aes128-gcm              | Speedup    |
|-----------------------------------------------------|---------------------|-------------------------|------------|
| Raw SSH pipe (184 MB → /dev/null)                   | 788 ms — 234 MB/s   | **446 ms — 413 MB/s**   | **1.77x**  |
| SCP file to guest                                   | 1.10 s — 167 MB/s   | **781 ms — 236 MB/s**   | **1.41x**  |
| Stream image load (cat file \| ssh docker load)     | 1.48 s — 124 MB/s   | **1.16 s — 159 MB/s**   | **1.28x**  |
| Streaming pipeline (podman save \| ssh docker load) | 2.08 s —  88 MB/s   | **1.87 s —  98 MB/s**   | **1.11x**  |

The raw SSH throughput nearly doubles: **234 MB/s → 413 MB/s**.

## Why AES-GCM is faster

`chacha20-poly1305` is a software-only cipher, designed for devices without hardware AES support (old phones, IoT). `aes128-gcm` leverages hardware AES instructions that are available on all CPUs that can realistically run minikube:

- Apple Silicon (M1+): dedicated AES unit
- Intel/AMD (2016+): AES-NI instruction set (universal since Skylake/Zen)
- ARMv8 with Crypto Extensions

OpenSSH defaults to chacha20 first regardless of the hardware, which leaves performance on the table. Lima [does the same fix](https://github.com/lima-vm/lima/blob/master/pkg/sshutil/sshutil.go) — detecting AES hardware at runtime and prioritizing AES-GCM when available.

## Proposed Change

minikube already fully controls the `ssh` command — it passes `-F /dev/null` (ignoring the user's `~/.ssh/config`) and configures all options via `baseSSHArgs` in `pkg/libmachine/ssh/client.go`. The fix is to add a cipher preference there:

```go
var baseSSHArgs = []string{
    "-F", "/dev/null",
    "-o", "Ciphers=^aes128-gcm@openssh.com,aes256-gcm@openssh.com",
    // ... existing options ...
}
```

The `^` prefix (OpenSSH 7.6+) prepends these ciphers to the front of OpenSSH's default cipher list. All default ciphers remain as implicit fallbacks, so if the server doesn't support AES-GCM, the connection will still succeed using the next available cipher.

minikube also controls the guest VM (buildroot-based Linux with OpenSSH), so AES-GCM is always available on the server side.

## Native SSH client — no change needed

minikube defaults to the Go native SSH client (`NativeClient` via `crypto/ssh`). The library's default cipher list already prefers AES-GCM:

```go
// golang.org/x/crypto/ssh defaultCiphers (v0.52.0)
[]string{
    "aes128-gcm@openssh.com",
    "aes256-gcm@openssh.com",
    "chacha20-poly1305@openssh.com",
    "aes128-ctr",
    "aes192-ctr",
    "aes256-ctr",
}
```

No change is needed for this path. The fix only affects the external SSH client used with `--native-ssh=false`.

## Runtime AES detection — not needed

Lima [detects AES hardware at runtime](https://github.com/lima-vm/lima/blob/master/pkg/sshutil/sshutil.go) using `golang.org/x/sys/cpu` and only prioritizes AES-GCM when acceleration is available. We don't think this is necessary for minikube:

- The `^` prefix only changes the **preference order** — it doesn't remove any ciphers. On a CPU without AES-NI, OpenSSH will still negotiate `aes128-gcm`, but run it in software. Performance will be roughly the same as chacha20 — no regression, just no speedup.
- Go's `crypto/ssh` also uses a static cipher list with AES-GCM first, with no runtime CPU detection. The Go AES-GCM implementation internally uses hardware instructions when available, but the preference order is the same on all hardware.
- Every CPU that can realistically run minikube and a VM has AES acceleration (Intel Skylake+ 2016, AMD Zen+ 2016, all Apple Silicon, ARMv8 with Crypto Extensions).

Adding runtime detection would add complexity (a new dependency on `golang.org/x/sys/cpu`) for no practical benefit.

## Future work

minikube could generate a per-machine `ssh_config` file (e.g. in `~/.minikube/machines/<name>/ssh_config`) with the cipher preference and other options. This would let users SSH to the VM directly using `ssh -F ~/.minikube/machines/minikube/ssh_config minikube` with optimal settings. Out of scope for this issue.

## References

- [Tuning TLS: AES-256 Beats ChaCha20 on Every CPU (2025)](https://ashvardanian.com/posts/chacha-vs-aes-2025/) — AES-GCM is 134-236% faster than ChaCha20 on Apple M2 Pro
- [Benchmarking AES on Apple Silicon](https://notonbluray.com/blog/benchmarking-aes-256-sha256-and-crc-64-on-apple-silicon/) — M2: 9.4 GiB/s, M4 Pro: 14.7 GiB/s for AES-256
- [Benchmarking SSH Ciphers](https://devops.aibit.im/article/benchmarking-ssh-ciphers-fastest-encryption) — recommends aes128-gcm for SSH on CPUs with AES-NI
- [Lima `detectAESAcceleration()`](https://github.com/lima-vm/lima/blob/master/pkg/sshutil/sshutil.go) — runtime AES detection, prioritizes AES-GCM when available

nirs:
Prepend aes128-gcm and aes256-gcm to the cipher list for the external SSH client (--native-ssh=false). On CPUs with hardware AES acceleration (all Apple Silicon, Intel Skylake+, AMD Zen+), this more than doubles SSH throughput vs the OpenSSH default:

    % hyperfine --warmup 1 \
        "dd if=/dev/zero bs=1M count=1024 2>/dev/null | out/minikube.before ssh --native-ssh=false -- 'cat > /dev/null'" \
        "dd if=/dev/zero bs=1M count=1024 2>/dev/null | out/minikube.after ssh --native-ssh=false -- 'cat > /dev/null'"
    Benchmark 1: dd if=/dev/zero bs=1M count=1024 2>/dev/null | out/minikube.before ssh --native-ssh=false -- 'cat > /dev/null'
      Time (mean ± σ):      3.430 s ±  0.008 s    [User: 2.863 s, System: 0.664 s]
      Range (min … max):    3.418 s …  3.443 s    10 runs

    Benchmark 2: dd if=/dev/zero bs=1M count=1024 2>/dev/null | out/minikube.after ssh --native-ssh=false -- 'cat > /dev/null'
      Time (mean ± σ):      1.554 s ±  0.012 s    [User: 0.302 s, System: 0.659 s]
      Range (min … max):    1.535 s …  1.570 s    10 runs

    Summary
      dd if=/dev/zero bs=1M count=1024 2>/dev/null | out/minikube.after ssh --native-ssh=false -- 'cat > /dev/null' ran
        2.21 ± 0.02 times faster than dd if=/dev/zero bs=1M count=1024 2>/dev/null | out/minikube.before ssh --native-ssh=false -- 'cat > /dev/null'

This measures raw SSH pipe throughput. In real-world use cases like image loading, the bottleneck is unpacking images and writing to disk on the guest, so the effective speedup will be much lower.

Fixes #23121

### DIFF
--- a/pkg/libmachine/ssh/client.go
+++ b/pkg/libmachine/ssh/client.go
@@ -79,6 +79,8 @@ const (
 var (
 	baseSSHArgs = []string{
 		"-F", "/dev/null",
+		// Prefer hardware-accelerated AES-GCM over ChaCha20 (https://github.com/kubernetes/minikube/issues/23121)
+		"-o", "Ciphers=^aes128-gcm@openssh.com,aes256-gcm@openssh.com",
 		"-o", "ConnectionAttempts=3", // retry 3 times if SSH connection fails
 		"-o", "ConnectTimeout=10", // timeout after 10 seconds
 		"-o", "ControlMaster=no", // disable ssh multiplexing


afbjorklund (approved):

nirs:
@afbjorklund Thanks for reviewing!

Issue #23043, opened 2026-05-25
Auto-tune workflow step timeouts based on historical CI performance


nirs:
## Summary

Create a weekly GitHub Actions workflow that automatically adjusts per-step timeout-minutes values in all CI workflow files based on recent performance data.

## Background

The workflow-stats tool (#23042) analyzes GitHub Actions step durations and suggests timeouts based on P95 x 3, rounded up to the nearest minute (minimum 1 minute). Currently this is a manual process - run the tool, review the output, update the YAML files.

Example output:

```
$ workflow-stats -workflow "Functional Test" -since 30

Step                                                N     Min     Avg     P95     Max  Timeout
Run Functional Test                               728   2m50s   4m05s   5m21s   9m27s   17m00s
Build minikube and e2e test binaries              156   1m07s   1m39s   2m09s   2m16s    7m00s
Set up Rootless Docker (rootless)                  67     42s     48s     56s   1m21s    3m00s
...
```

## Proposal

Add a scheduled workflow that:

1. Runs workflow-stats -o json -since 30 for each CI workflow
2. Parses the suggested timeouts per step from the JSON output
3. Matches step names to their definitions in .github/workflows/*.yml
4. Updates timeout-minutes values where they differ from the suggestion
5. Opens a PR if any changes were made

### Considerations

- YAML parsing must preserve comments and formatting
- Step name matching between tool output and workflow YAML needs to handle truncation and special characters
- Matrix jobs share step definitions across variants with different performance profiles - use the worst-case (max across variants) timeout
- The PR should include a summary of what changed and why
- Should run weekly (e.g., Sunday night) to avoid noisy PRs

## Related

- #23042 - workflow-stats CLI tool for analyzing step durations

medyagh:
thats a really cool idea ! I love that

baohuy1303:
Hi I'm working on #23041 but would also love to take on this issue as what I'm working on for that issue can be automated with this!

PR #23168, opened 2026-06-16
ci: add step-level timeouts to remaining CI workflows

### LINKED ISSUE #23041: ci: Ensure workflow timeouts - runs can get stuck for hours

baohuy1303:
Fixes #23041.

Adds `timeout-minutes` to every step in 66 `.github/workflows/*.yml` files. Two files were skipped (see below). One commit per file.

Each commit message has the exact `workflow-stats` (#23042) command used and its output table, so the numbers can be reproduced and reviewed per file. Two files were left unmodified (see below).

### Workflows with computed timeouts (42 files)

Every step's timeout comes straight from that file's own `workflow-stats` output (`P95 x 3`, rounded up, floor 1 minute). No defaults, no overrides.

`dependabot-gomodtidy.yml`, `docs.yml`, `leaderboard.yml`, `sbom.yml`, `smoke-test.yml`, `sync-minikube.yml`, `translations.yml`, `twitter-bot.yml`, `winget.yml`, `yearly-leaderboard.yml`, `update-calico-version.yml`, `update-cilium-version.yml`, `update-cloud-spanner-emulator-version.yml`, `update-cni-plugins-version.yml`, `update-containerd-version.yml`, `update-cri-dockerd-version.yml`, `update-debian-version.yml`, `update-docker-buildx-version.yml`, `update-flannel-version.yml`, `update-gh-version.yml`, `update-go-github-version.yml`, `update-golang-version.yml`, `update-golint-version.yml`, `update-headlamp-version.yml`, `update-hugo-version.yml`, `update-ingress-version.yml`, `update-inspektor-gadget-version.yml`, `update-k8s-versions.yml`, `update-kindnetd-version.yml`, `update-kong-ingress-controller-version.yml`, `update-kong-version.yml`, `update-kube-vip-version.yml`, `update-kubeadm-constants.yml`, `update-kubernetes-versions-list.yml`, `update-kubevirt-version.yml`, `update-metrics-server-version.yml`, `update-nerdctl-version.yml`, `update-nvidia-device-plugin-version.yml`, `update-registry-version.yml`, `update-site-node-version.yml`, `update-volcano-version.yml`, `update-yakd-version.yml`

### Workflows needing special treatment (24 files)

Each of these has at least one step where the timeout did not come directly from that file's own data. The exact reasoning is in each commit message; short version here:

- **Pre-existing timeouts kept**: `lint.yml` (`Lint`=8m, `Boilerplate check`=2m), `unit-test.yml` (`unit test`=15m).

- **0 run samples for a step, defaulted by analogy**: `time-to-k8s.yml`, `update-all.yml`, `update-amd-gpu-device-plugin-version.yml`, `update-docsy-version.yml`, `update-gcp-auth-version.yml`, `update-gopogh-version.yml`, `update-gotestsum-version.yml`, `update-istio-operator.yml`, `update-kubectl-version.yml`, `update-portainer-version.yml`, `update-nerdctld-version.yml` (Create PR step, mostly; usually defaulted to 2m).

- **0 run samples for every step**: `update-cri-o-version.yml`, `update-crictl-version.yml`, `update-iso-image-versions.yml` (manual-trigger workflow), all timeouts defaulted by analogy with similar update workflows.

- **`go-housekeeping.yml`**: `Create PR` has 0 samples; cross-checked against real `Create PR` timings in the other 65 files in this PR (highest P95 anywhere is 13s) before keeping the 2m default.

- **Non-standard command flags**: `hide-minikube-bot-comments.yml` triggers on every `issue_comment` repo-wide and returned 1000+ runs in 30 days (fetch did not finish). Used `-since 2` (163 runs) instead. `vex.yml` (no top-level `name:`, so `-workflow` uses the file path).

- `minikube-image-benchmark.yml` got a timeout despite always failing. The steps ran to completion every time before the job fails so we still have data for the timeouts. That said, the job has failed on every single run for the last 30 days, so we've noted that in this issue #23195 

### Skipped files

**`functional_extra.yml`**

Zero runs in 30 days. Only 4 runs in 180 days, and those runs have different step names than the current file, so the data is stale and cannot be used to compute timeouts. Left unchanged. Noted in #23196 

**`time-to-k8s-public-chart.yml`**

7 out of 10 steps have zero run data because of two pre-existing failures: the Docker-runtime benchmark fails on every run, and the virtualbox job is cancelled on every run. Left unchanged. Noted in #23195

### Discussions

The `P95 x 3` formula gets very loose for long-running steps. `minikube-image-benchmark.yml`'s `Run Benchmark` step has P95 = 22m49s, giving a 69m timeout. I kept it consistent in this PR, but it's worth considering a multiplier as a follow-up to the `workflow-stats` tool itself and use that for the later automation that I'll be working on (I'm thinking of `x3` under 5m, `x2` up to 10m, `x1.5` beyond).

### Notes

- [x] Open separate issues #23195 #23196 
- [ ] Open a follow-up issue to add explicit `name:` to unnamed `uses:` steps (e.g. in the `update-*-version.yml` family). Reason: without a name, GitHub (and `workflow-stats`) reports the step by its action SHA, so a routine Dependabot SHA bump splits one logical step's run history across multiple names.


nirs:
@baohuy1303 Thanks for this great work!

> ### Skipped files (need input)
> **`functional_extra.yml`**
> 
> Zero runs in 30 days. Only 4 runs in 180 days, and those runs have different step names than the current file, so the data is stale and cannot be used to compute timeouts. Left unchanged.

Looks good, until we have some runs, we cannot set the timeout. The author of this workflow should handle this.

> **`time-to-k8s-public-chart.yml`**
> 
> 7 out of 10 steps have zero run data because of two pre-existing failures: the Docker-runtime benchmark fails on every run, and the virtualbox job is cancelled on every run. Left unchanged, so same questions as above: follow-up after the failures are fixed, or do we have any defaults to apply?

I think this is broken for a long time and not relevant. No need to change it.

Please open an issue for the workflow failures. If it always fails we should not run it.

> ### Steps with zero run data (defaults applied)
> Some steps had no run history at any window size. A default was applied by analogy and is documented here.
> 
> File	Step	Reason	Applied
> `update-iso-image-versions.yml`	All steps	Manual trigger, 0 runs in 365 days	checkout=1m, setup-go=2m, Bump versions=2m (all these update workflows have essentially the same time, so I just derived from those)

Sounds good - but the right place to document the decision is the commit message - not here.

> `go-housekeeping.yml`	`Create PR`	Conditional step, 0 triggers in 90 days	2m (following the other `Create PR` timeouts)

We can check how much time `Create PR` job took in other workflows and use the same timeout.

This reason for the timeout should be documented in the commit message.

> `sbom.yml`, `twitter-bot.yml`, `winget.yml`, `time-to-k8s.yml`	All steps	Release-triggered, fewer than 5 total runs	Setup steps=1m, longer steps=based on max observed or 5m

Sounds good, document in the commit message.

> Note: `hide-minikube-bot-comments.yml` triggers on every `issue_comment` repo-wide and returned 1000+ runs in 30 days (fetch did not finish). Used `-since 2` (163 runs) instead.]

Sounds good, document in the commit message.

> ### Discussions
> The `P95 x 3` formula gets very loose for long-running steps. `minikube-image-benchmark.yml`'s `Run Benchmark` step has P95 = 22m49s, giving a 69m timeout. I kept it consistent in this PR, but it's worth considering a multiplier as a follow-up to the `workflow-stats` tool itself and use that for the later automation that I'll be working on (I'm thinking of `x3` under 5m, `x2` up to 10m, `x1.5` beyond).

We may have smarter function that works like you describe - it can be implemented in the workflow-stats tool. Maybe the factor can decrease slowly as the time increases. But for now this is good enough, preventing stuck jobs for days, while allowing successful runs when the runners are overloaded.

To complete this work we need:
- Add the workflow-stats command and output to every commit message - this makes it easy to review and understand later why we used certain timeouts.
- For every timeout that does not come from the command output explain why it was used in the commit message.
- Update the pr message - the commit messages provide all the details about specific workflows, the pr message can show a summary of the changes like:
  - workflows applying computed timeouts from workflow-stats
  - workflows that need special treatment
  - unmodified workflows (no runs, always fail)

nirs:
@baohuy1303 We can use `Fixes #23041` in this PR. You can open a new issue for handling special cases that we cannot handle now, like new workflows that have no runs yet.

baohuy1303:
Hi @nirs I've updated all the commits to have commands, tables, and justification if needed for the message like you said. I fixed the PR body as well! You can review it

.github/workflows/lint.yml
nirs: Thanks for commenting on this in the commit message. I think we should apply the computed timeout here as well. Using special values make it hard to automated this (#23043).

nirs: Same here, use the computed value. History shows clearly that we don't need 2 minutes here.

.github/workflows/unit-test.yml
```diff
@@ -49,24 +49,29 @@ jobs:
...
               sudo apt-get update
               sudo apt-get install -y libvirt-dev
         - name: Install make (Windows)
+          timeout-minutes: 1
           if: runner.os == 'Windows'
           run: choco install make -y
         # TODO: add gopogh reports for unit tests too
         - name: unit test
-          timeout-minutes: 15
+          timeout-minutes: 12
```
nirs, line 74: Good change - but you wrote:
> unit test already had timeout-minutes: 15m and was left as-is;

nirs (commented):
Partial review - thanks the great commit messages!

next: https://github.com/kubernetes/minikube/pull/23168/changes/a69fd8b2abdd57ae7b87163ea3e8b618aec372dd

baohuy1303:
> Partial review - thanks the great commit messages!
> 
> next: [a69fd8b](https://github.com/kubernetes/minikube/commit/a69fd8b2abdd57ae7b87163ea3e8b618aec372dd)

Awesome, once I get the full review I'll fix everything! I'll also start checking out #23043 and see what I can do!

.github/workflows/update-kong-version.yml
```diff
@@ -15,11 +15,14 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10
+        timeout-minutes: 1
```
nirs, line 18: We need to add name for every step to avoid the unhelpful runs per sha256sum. This is also nicer for working with the workflow.

We can do this later, since adding name will not help to analyze the current history. It is already messy because we don't have names.

nirs:
@baohuy1303 Reviewed all now. All the special cases will have to be dropped when we automate this to keep the automated code simple.

baohuy1303:
> @baohuy1303 Reviewed all now. All the special cases will have to be dropped when we automate this to keep the automated code simple.

Thanks nirs, I'll change to the correct computed timeouts for all the workflow you mentioned. For the steps without name, do you want me to just fix the computed timeouts, and handle the missing names in another PR or on the PR that I'm working on for #23043 ? So that this PR can be done for now

Also I've already coded a denylist/droplist for all the special cases (insufficient data, failing workflows) too :)

nirs:
> > @baohuy1303 Reviewed all now. All the special cases will have to be dropped when we automate this to keep the automated code simple.
> 
> For the steps without name, do you want me to just fix the computed timeouts, and handle the missing names in another PR or on the PR that I'm working on for #23043 ? 

Adding names will not help us now, it will be useful in 30-90 days when we have enough runs to compute new timeouts from. So lets open an issue to add this and limit the scope this PR - it is huge as is.

> Also I've already coded a denylist/droplist for all the special cases (insufficient data, failing workflows) too :)

I would not add denylist since we will need to maintain is. The code automating this should handle missing data in a smart way limiting manual work.

baohuy1303:
> I would not add denylist since we will need to maintain is. The code automating this should handle missing data in a smart way limiting manual work.

For the tool, I'm planning to have it open a PR every time the cron-job is ran and finished.

What I'm planning to do is flag these types of workflow (insufficient data, failing) to accumulate a list. Then on the PR body (which would have a summary of all the changes), we will put the flagged workflows on its own section as well.

> We may have smarter function that works like you describe - it can be implemented in the workflow-stats tool. Maybe the factor can decrease slowly as the time increases. But for now this is good enough, preventing stuck jobs for days, while allowing successful runs when the runners are overloaded.

I also forgot to mention, would you like me to implement this in the full automation as well

nirs:
> > I would not add denylist since we will need to maintain is. The code automating this should handle missing data in a smart way limiting manual work.
> 
> For the tool, I'm planning to have it open a PR every time the cron-job is ran and finished.
> 
> What I'm planning to do is flag these types of workflow (insufficient data, failing) to accumulate a list. Then on the PR body (which would have a summary of all the changes), we will put the flagged workflows on its own section as well.
> 
> > We may have smarter function that works like you describe - it can be implemented in the workflow-stats tool. Maybe the factor can decrease slowly as the time increases. But for now this is good enough, preventing stuck jobs for days, while allowing successful runs when the runners are overloaded.
> 
> I also forgot to mention, would you like me to implement this in the full automation as well

I would start with the simplest possible change - update command that get the stats from all other workflows, compute the timeout and update the workflows. Then open a PR with the changes.

This likely to create PR very rarely since we use at least 30 days of data, and p95x3, which is unlikely to change a lot.

Smarter timeout computation and other nice features can be added later when we have the basic automation in place.

baohuy1303:
> I would start with the simplest possible change - update command that get the stats from all other workflows, compute the timeout and update the workflows. Then open a PR with the changes.
> 
> This likely to create PR very rarely since we use at least 30 days of data, and p95x3, which is unlikely to change a lot.
> 
> Smarter timeout computation and other nice features can be added later when we have the basic automation in place.

Got it, I'm working on it right now!

I've also fixed all the commit messages and make the timeout of some files to be exactly the same as the precomputed table. I think it should be good now!

nirs (commented):
@baohuy1303 Great work, thanks!

PR #23170, opened 2026-06-16
Enhance ingress-dns.md with DNS setup notes

zaro:
Added a note about the limitations of the systemd-resolved system level DNS approach and provided a better method for setting link-level DNS.

### DIFF
--- a/site/content/en/docs/handbook/addons/ingress-dns.md
+++ b/site/content/en/docs/handbook/addons/ingress-dns.md
@@ -144,6 +144,23 @@ EOF
 sudo systemctl restart systemd-resolved
 ```
 
+NOTE: the above apporach makes the system DNS unusable, as the first request to any noncached domain will always fail!
+
+Better approach is to register it as link level DNS ( unformtinately doesn't persist restarts ):
+```
+MINIKUBE_IP=$(minikube ip)
+MINIKUBE_IP_PREFIX=$(echo "$MINIKUBE_IP" | sed 's/\.[0-9]\{1,3\}$//')
+MINIKUBE_IF=$(ip route | grep ${MINIKUBE_IP_PREFIX} | awk '{print $3}' | head -1)
+echo "Setting DNS for .test on $MINIKUBE_IF(ip:$MINIKUBE_IP)"
+
+# Set DNS scoped to just that interface
+sudo resolvectl dns $MINIKUBE_IF "${MINIKUBE_IP}"
+sudo resolvectl domain $MINIKUBE_IF "~test"
+
+sudo systemctl restart systemd-resolved
+```
+
+
 {{% /linuxtab %}}
 
 {{% mactab %}}


site/content/en/docs/handbook/addons/ingress-dns.md
```diff
@@ -144,6 +144,23 @@ EOF
...
+Better approach is to register it as link level DNS ( unformtinately doesn't persist restarts ):
+```
+MINIKUBE_IP=$(minikube ip)
+MINIKUBE_IP_PREFIX=$(echo "$MINIKUBE_IP" | sed 's/\.[0-9]\{1,3\}$//')
+MINIKUBE_IF=$(ip route | grep ${MINIKUBE_IP_PREFIX} | awk '{print $3}' | head -1)
+echo "Setting DNS for .test on $MINIKUBE_IF(ip:$MINIKUBE_IP)"
+
+# Set DNS scoped to just that interface
+sudo resolvectl dns $MINIKUBE_IF "${MINIKUBE_IP}"
+sudo resolvectl domain $MINIKUBE_IF "~test"
```
nirs, line 158: This should run on a Linux host using systemd-resolved - right?

It will help to add a section for this method, and add a note about portability - on which system it works.

site/content/en/docs/handbook/addons/ingress-dns.md
```diff
@@ -144,6 +144,23 @@ EOF
 sudo systemctl restart systemd-resolved
 ```
 
+NOTE: the above apporach makes the system DNS unusable, as the first request to any noncached domain will always fail!
```
nirs, line 147: If the approach above does not work, why keep it? Let's replace it with a better one?

zaro: I am not sure why is it there in the first place. Also it kind of works, only the first lookup of domain fails, on retry it succeeds. 

I have tested only on my system, but it's kind of clear that this issue will happen on any modern systemd ( i have no clue about the older ones).

I can replace the whole section, not an issue. I am simply not sure :)

site/content/en/docs/handbook/addons/ingress-dns.md
zaro: yes, the system where minikube is being used. It is already in teh right section, it works for any system with systemd-resolved.

site/content/en/docs/handbook/addons/ingress-dns.md
nirs: I don't think it makes sense to have docs that have an issue and suggest alternative way. We should have one good way.

Not clear to me why the current way does not work. We use standard drop-in file, which makes the configuration persistent. The configuration is also much simpler than new way, and it does not depend on cryptic and unclear commands.

What is the issue with this configuration? Why the temporary configuration does not have the same issue?

Can we improve the current way to avoid the failure on the first access?

site/content/en/docs/handbook/addons/ingress-dns.md
```diff
@@ -144,6 +144,23 @@ EOF
...
+
+Better approach is to register it as link level DNS ( unformtinately doesn't persist restarts ):
+```
+MINIKUBE_IP=$(minikube ip)
+MINIKUBE_IP_PREFIX=$(echo "$MINIKUBE_IP" | sed 's/\.[0-9]\{1,3\}$//')
```
nirs, line 152: This is not clear without example value. You can add a comment to make this clear.

site/content/en/docs/handbook/addons/ingress-dns.md
```diff
@@ -144,6 +144,23 @@ EOF
...
+MINIKUBE_IF=$(ip route | grep ${MINIKUBE_IP_PREFIX} | awk '{print $3}' | head -1)
```
nirs, line 153: Same issue - comment with example value will make this clear.

nirs:
@zaro please check https://github.com/kubernetes/minikube/pull/23172 - this changing or deleting a cluster (with the same profile name). With this you can add the cluster to /etc/hosts and ingress-dns and setting up split dns configuration is not needed.

zaro:
@nirs Yes, this is good. Actually I am using /etc/hosts and the ip doesn't change even when deleting and recreating the cluster. My issue is that I have few apps I develop in minikube and each one of them has subdomains. And the list is getting longer :)

But  guess with this change, I can try one of the other approaches like the Network manager one.

PR #23176, opened 2026-06-19
docs: Update krunkit.md with latest installation instructions

JoKrefting:
Replace deprecated installation instructions with new tap slp/krun and note about how to uninstall previous tap.

### DIFF
--- a/site/content/en/docs/drivers/krunkit.md
+++ b/site/content/en/docs/drivers/krunkit.md
@@ -7,7 +7,7 @@ aliases:
 
 ## Overview
 
-[krunkit](https://github.com/containers/krunkit) is an open-source program for
+[krunkit](https://github.com/libkrun/krunkit) is an open-source program for
 macOS virtualization, optimized for GPU accelerated virtual machines and AI
 workloads.
 
@@ -24,9 +24,14 @@ workloads.
 To install krunkit run:
 
 ```shell
-brew tap slp/krunkit
+brew tap libkrun/krun
+brew trust libkrun/krun # trust the tap only required if using Homebrew 6.0 or later
 brew install krunkit
 ```
+
+{{% alert title="Note" color="primary" %}}
+If you've ever installed krunkit from the old tap `slp/krunkit` or `slp/krun`, to upgrade to the latest version you'll need to follow the [Removing the old Homebrew tap](https://github.com/libkrun/krunkit#removing-the-old-homebrew-tap) instructions first before installing as shown above.
+{{% /alert %}}
 
 ## Networking
 


site/content/en/docs/drivers/krunkit.md
```diff
@@ -24,10 +24,14 @@ workloads.
 To install krunkit run:
 
 ```shell
-brew tap slp/krunkit
+brew tap slp/krun
+brew trust slp/krun
```
nirs: Unfortunately slp/krun is also deprecated now - use libkrun/krun:
https://github.com/libkrun/homebrew-krun

Worth a note that brew trust is required only on Homebrew >= 6.0

site/content/en/docs/drivers/krunkit.md
```diff
@@ -24,10 +24,14 @@ workloads.
...
 brew install krunkit
 ```
 
+> [!IMPORTANT]
+> If you've ever installed krunkit from the old tap `slp/krunkit`, to upgrade to the latest version you'll need to follow the [Removing the old Homebrew tap](https://github.com/libkrun/krunkit#removing-the-old-homebrew-tap) instructions first, and then the ones in the [Installing from Homebrew](https://github.com/libkrun/krunkit#installing-from-homebrew) section (as shown above).
```
nirs: This note is needed for both `slp/krunkit` and `slp/krun`, and the current docs are stale.

See the upstream pull request fixing "Removing the old Homebrew tab" section:
https://github.com/libkrun/krunkit/pull/105

So both suggest links should not use use at this point. We can include here the correct commands to remove both sections or wait until upstream fixed the instructions.

site/content/en/docs/drivers/krunkit.md
```diff
@@ -24,10 +24,14 @@ workloads.
 To install krunkit run:
 
 ```shell
-brew tap slp/krunkit
+brew tap slp/krun
+brew trust slp/krun
 brew install krunkit
 ```
 
+> [!IMPORTANT]
```
nirs: We use different syntax for alerts, check other docs for examples.

site/content/en/docs/drivers/krunkit.md
JoKrefting: i replaced the slp with the new libkrun but i didn't include the note regarding trust as this is going to be default regardless for everyone who is keeping the software up to date, right?

site/content/en/docs/drivers/krunkit.md
JoKrefting: I replaced this with updated url for brew repo and changed the notation to match the note notation

site/content/en/docs/drivers/krunkit.md
nirs: For people updating brew regularly the comment is not needed, they already have Homebrew 6.0+. The comment can help people that do not update Homebrew. I don't know how users may be affected but adding a temporary note that we can remove in the next release does not cost much.

site/content/en/docs/drivers/krunkit.md
nirs: The first part looks good - you can learn how to remove the old tap from https://github.com/libkrun/krunkit#removing-the-old-homebrew-tap.

The second part is not needed - we show how to install krunkit from latest tap in this page. There is no point to go to another repo to find the same instrcutions.

nirs (commented):
Pleas squash the commits - there is no reason to have more than one commit for this change.

site/content/en/docs/drivers/krunkit.md
JoKrefting: i added a small inline comment @nirs

nirs (approved):
@JoKrefting Thanks!

PR #23208, opened 2026-06-23
start: simplify getMDNS() control flow

carterpewpew:
Combine the two early returns in `getMDNS()` into a single conditional that reads: "if enabled but unsupported driver, warn and reject." This reveals the intent better than separate guard clauses.

Follow-up cleanup from #22964 per review feedback.

cc @nirs


nirs (approved):
@carterpewpew Thanks!

PR #23215, opened 2026-06-24
fix: use buffered channels to prevent goroutine leak on status timeout

carterpewpew:
Fixes goroutine leak when driver status check exceeds the 20s timeout. Both `driver.Status()` and `registry.Available()` spawn a goroutine that sends on an unbuffered channel. If the timeout fires first, no receiver remains and the goroutine blocks on send forever. Changing to a buffered channel (capacity 1) lets the sender complete and exit.

Same pattern and fix as #11247

### DIFF
--- a/pkg/minikube/registry/global.go
+++ b/pkg/minikube/registry/global.go
@@ -121,7 +121,7 @@ func Available(vm bool, options *run.CommandOptions) []DriverState {
 			klog.Errorf("%q does not implement Status", d.Name)
 			continue
 		}
-		stateChannel := make(chan State)
+		stateChannel := make(chan State, 1)
 		timeoutChannel := time.After(20 * time.Second)
 		go func() {
 			stateChannel <- d.Status(options)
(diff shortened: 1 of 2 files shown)


nirs (approved):
Thanks

pkg/minikube/registry/global.go
```diff
@@ -121,7 +121,7 @@ func Available(vm bool, options *run.CommandOptions) []DriverState {
 			klog.Errorf("%q does not implement Status", d.Name)
 			continue
 		}
-		stateChannel := make(chan State)
+		stateChannel := make(chan State, 1)
```
nirs, line 124: Note that this function duplicates the timeout mechanism and configuration (20s timeout). It also use 20s timeout per driver instead of 20s timeout per Available() call.

We can extract a helper to call d.Status() with a timeout and use it in both sites, but this will be wrong since we want to have timeout for the entire call, not 20s * number of drivers.

Let's open an issue about this for future change.

PR #23219, opened 2026-06-24
iso: Add build progress status

nirs:
Extract retry_github_status from common.sh into github.sh so it can be reused without pulling in test-job dependencies. Both common.sh and build_iso.sh source it.

Add status updates to build_iso.sh so the PR check description shows build progress:

    ISO PR Build Push — pending — "Cleaning up build machine"
    ISO PR Build Push — pending — "Preparing build machine"
    ISO PR Build Push — pending — "Building ISO"
    ISO PR Build Push — success — (set by Jenkins plugin)

Previously the status showed only "Build started for merge commit" until the job finished, with no visibility into which stage was running or where it got stuck.

### DIFF
--- a/hack/jenkins/build_iso.sh
+++ b/hack/jenkins/build_iso.sh
@@ -21,6 +21,14 @@
 
 set -x -o pipefail
 
+source ./hack/jenkins/github.sh
+
+set_status() {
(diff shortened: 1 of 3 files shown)


nirs:
Not needed, the progress is not helpful.

PR #23221, opened 2026-06-24
refactor: use slices.Contains to simplify code

daixihegu:
There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.

### DIFF
--- a/pkg/minikube/reason/k8s.go
+++ b/pkg/minikube/reason/k8s.go
@@ -16,7 +16,11 @@ limitations under the License.
 
 package reason
 
-import "github.com/blang/semver/v4"
+import (
+	"slices"
+
+	"github.com/blang/semver/v4"
+)
 
 // K8sIssue represents a known issue with a particular version of Kubernetes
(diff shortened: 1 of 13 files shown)


daixihegu:
signed.

pkg/minikube/reason/k8s.go
```diff
@@ -16,6 +16,8 @@ limitations under the License.
 
 package reason
 
+import "slices"
+
 import "github.com/blang/semver/v4"
```
nirs: Please group the imports

nirs (changes requested):
@daixihegu thanks!

pkg/minikube/reason/k8s.go
daixihegu: Thanks. Modified. Please review it again.  @nirs

nirs (approved):
Thanks

PR #23223, opened 2026-06-25
Iso build cleanup test

nirs:
Test iso build push commit logic: The iso build skips make release, mark the build as success, and use a commit updating the iso details.

If the push fail we should get a commit with manual instruction on how to create the ISO commit.


nirs:
Not needed now

PR #23228, opened 2026-06-26
fix: preserve tag for registry refs with a port

### LINKED ISSUE #23227: --image-repository with a registry port breaks kicbase pull - offline clusters won't start
### What Happened?

For anyone using minikube in air-gapped setup w/ a local img registry that has a port number, minikube fails to pull the kicbase img from the mirror, because the port makes it misparse the img ref, name.NewTag() then rejects the ref, so minikube ignores the mirror and falls back to the public registry. In offline envs this fallback also fails, cluster fails to start. I proved this e2e against a real registry.

Steps:                                                                                                                                                  
 1. Run in no internet env;                                                                        
 2. Point at a local img mirror w/ --image-repository localhost:5000/.... This is what docs say on how to do offline setups;                             
 3. minikube tries to pull the kicbase base img from the mirror (localhost:5000/.../kicbase-builds:vX@sha256:...);                                       
 4. Bug is here, because the mirror url has a port, :5000, image.Tag() miscounts colons and does not strip the digest. name.NewTag() then rejects the ref w/ "repository can only contain ... sha256";                                                                                        
 5. minikube gives up on the mirror and falls back to the public registry. Offline, the fallback also fails = cluster fails to start. 

### Attach the log file

[log.txt](https://github.com/user-attachments/files/29392862/log.txt)

### Operating System

macOS (Default)

### Driver

Docker

vladlearns:
fixes #23227                                                                                                                            

image.Tag() guesses whether a ref has a digest by counting colons (3 = has digest). That breaks for registries with a port - host:port/repo:tag also has 3 parts. So --image-repository localhost:5000/... (standard air-gapped/offline setup) keeps the sha256 on the kicbase ref, name.NewTag() rejects it, minikube falls back to docker.io instead of the mirror. Offline, the cluster won't start.

Reproduced on master:                                                                                                                                   
                                                                                                                                                         
 ```                                                                                                                                                     
cache.go:183] failed to download localhost:5050/mirror/k8s-minikube/kicbase-builds:v0.0.50-...@sha256:...,                                            
will try fallback image if available: parsing tag: repository can only contain ... @sha256                                                            
 ```                                                                                                                                                     

fix: don't count colons - strip from the last @ (a digest is always @-delimited); treat the part after the last / as the tag (a registry port's : sits before the /).                                                                                                                                          

Before/after:                                                                                                                                           
                                                                                                                                                         
 ```                                                                                                                                                     
Tag("localhost:5000/repo/kicbase-builds:v0.0.50@sha256:abc")
 before: "...kicbase-builds:v0.0.50@sha256:abc"  # NewTag fails
 after:  "...kicbase-builds:v0.0.50"

Tag("localhost:5000/repo/img:v1")
 before: "localhost:5000/repo/img"   # tag dropped
 after:  "localhost:5000/repo/img:v1"
 ```                                                                                                                                                     

only the kicbase path (download.ImageToCache/CacheToDaemon) calls Tag(). k8s component img use image.SaveToDir and aren't affected.                
                                                                                                                                                         
Tested w/ go test ./pkg/minikube/image/ (added 3 port cases), re-ran the repro against the fix - mirror ref parses, fallback no longer triggers.

### DIFF
--- a/pkg/minikube/image/image.go
+++ b/pkg/minikube/image/image.go
@@ -107,20 +107,22 @@ func DigestByGoLib(imgName string) string {
 	return cf.Hex
 }
 
-// Tag returns just the image with the tag
-// eg image:tag@sha256:digest -> image:tag if there is an associated tag
-// if not possible, just return the initial img
+// Tag returns the image ref with any digest stripped
+//
+//	image:tag@sha256:digest	-> image:tag
+//	image@sha256:digest	-> image
+//	localhost:5000/myimage:v1@sha256:digest	-> localhost:5000/myimage:v1
+//
+// refs without a digest are returned unchanged:
+//
+//	image:tag -> image:tag
+//	image	-> image
 func Tag(img string) string {
 	at := strings.LastIndex(img, "@")
 	if at == -1 {
 		return img
 	}
-	base := img[:at]
-	slash := strings.LastIndex(base, "/")
-	if strings.IndexByte(base[slash+1:], ':') == -1 {
-		return img
-	}
-	return base
+	return img[:at]
 }
 
 func canonicalName(ref name.Reference) string {
--- a/pkg/minikube/image/image.go
+++ b/pkg/minikube/image/image.go
@@ -111,12 +111,16 @@ func DigestByGoLib(imgName string) string {
 // eg image:tag@sha256:digest -> image:tag if there is an associated tag
 // if not possible, just return the initial img
 func Tag(img string) string {
-	split := strings.Split(img, ":")
-	if len(split) == 3 {
-		tag := strings.Split(split[1], "@")[0]
-		return fmt.Sprintf("%s:%s", split[0], tag)
+	at := strings.LastIndex(img, "@")
+	if at == -1 {
+		return img
 	}
-	return img
+	base := img[:at]
+	slash := strings.LastIndex(base, "/")
+	if strings.IndexByte(base[slash+1:], ':') == -1 {
+		return img
+	}
+	return base
 }
 
 func canonicalName(ref name.Reference) string {
--- a/pkg/minikube/image/image_test.go
+++ b/pkg/minikube/image/image_test.go
@@ -31,7 +31,7 @@ func TestTag(t *testing.T) {
 			expected: "image:tag",
 		}, {
 			image:    "image@sha256:digest",
-			expected: "image@sha256:digest",
+			expected: "image",
 		}, {
 			image:    "image",
 			expected: "image",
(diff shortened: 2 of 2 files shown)


vladlearns:
looks like a known cri-o flake unrelated to this pr

pkg/minikube/image/image.go
```diff
@@ -111,12 +111,16 @@ func DigestByGoLib(imgName string) string {
 // eg image:tag@sha256:digest -> image:tag if there is an associated tag
 // if not possible, just return the initial img
```
nirs: Can you update the comment to show the cases we handle?
- image:tag@sha256:digest -> image:tag (existing)
- image:tag -> image:tag     ?
- image -> image             ?

Do we have any other cases?

pkg/minikube/image/image_test.go
```diff
@@ -36,6 +36,16 @@ func TestTag(t *testing.T) {
...
 		},
+		{
+			image:    "localhost:5000/myimage:v1",
+			expected: "localhost:5000/myimage:v1",
+		}, {
+			image:    "localhost:5000/mymirror/kube-apiserver:v1.30.0",
+			expected: "localhost:5000/mymirror/kube-apiserver:v1.30.0",
+		}, {
+			image:    "registry.io:5000/path/img:v1@sha256:digest",
+			expected: "registry.io:5000/path/img:v1",
```
nirs, line 47: Can be useful to test `@digest` with all new cases. It should not matter based on the code, but the test should not care about the implementation, and these tests are very quick and easy to add.

pkg/minikube/image/image.go
```diff
@@ -111,12 +111,16 @@ func DigestByGoLib(imgName string) string {
...
+	if at == -1 {
+		return img
 	}
-	return img
+	base := img[:at]
+	slash := strings.LastIndex(base, "/")
+	if strings.IndexByte(base[slash+1:], ':') == -1 {
+		return img
+	}
+	return base
```
nirs: This seems inconsistent - if we return the entire image from image@digest, why we need to split the image by slash? I also don't see any test case requiring this code.

To satisfy the tests, this should work:

```go
at := strings.LastIndex(img, "@")
if at == -1 {
	return img
}
return img[:at]
```

Maybe the test do not describe the wanted behavior? In this case need to update the tests. For this function, we should not have code that is not required by the tests.

nirs (changes requested):
@vladlearns Thanks for the fix!

pkg/minikube/image/image.go
vladlearns: Yes, the test from 2020 was expecting the wrong thing. 
I checked all 7 callers - we can cut the slash.
Thank you!

pkg/minikube/image/image.go
vladlearns: Yes and yes, added 1 more case. Thank you!

pkg/minikube/image/image_test.go
vladlearns: Done. Thank you!

pkg/minikube/image/image_test.go
```diff
@@ -31,11 +31,27 @@ func TestTag(t *testing.T) {
 			expected: "image:tag",
 		}, {
 			image:    "image@sha256:digest",
-			expected: "image@sha256:digest",
+			expected: "image",
```
nirs, line 34: Nice to fix 6 years old bug.

nirs (approved):
@vladlearns thanks!

/lgtm
/ok-to-test

nirs:
@vladlearns failed test are not related #23323.

/retest-required

PR #23230, opened 2026-06-27
fix: honor --static-ip flag when used with --network on pre-existing networks

### LINKED ISSUE #19284: --static-ip not respected when run with --network with docker driver

carterpewpew:
When `--static-ip` is used with `--network` pointing to a pre-existing Docker network (created without an explicit `--gateway`), the static IP is silently ignored. The container ends up on the default bridge with a
random IP.

**Root cause:** `CreateNetwork` calls `dockerNetworkInspect` on existing networks. If Docker's IPAM config has no explicit gateway, `net.ParseIP("")` returns `nil`. Back in `Driver.Create()`, the branch that applies the static IP was guarded by `gateway != nil && staticIP != ""`, so a nil gateway caused the static IP to be skipped entirely.

**Fix:** Remove the `gateway != nil` guard from the `staticIP` branch.

The user explicitly requested a static IP; whether the network reports a gateway is irrelevant to honoring that request.
Only the previously-broken code path (nil gateway + non-empty staticIP) changes behavior. All other combinations are unaffected.

**Tested locally:**
- Created a Docker network without `--gateway`: `docker network create --driver bridge --subnet=192.168.150.0/24 test_net`
- Stock minikube: `minikube ip` returns `172.17.0.6` (bug, static IP ignored)
- Patched minikube: `minikube ip` returns `192.168.150.20` (fix confirmed)

Fixes : #19284

### DIFF
--- a/pkg/drivers/kic/kic.go
+++ b/pkg/drivers/kic/kic.go
@@ -106,7 +106,7 @@ func (d *Driver) Create() error {
 			exit.Message(reason.IfDedicatedNetwork, msg, args)
 		}
 		out.WarningT(msg, args)
-	} else if gateway != nil && staticIP != "" {
+	} else if staticIP != "" {
 		params.Network = networkName
 		params.IP = staticIP
 	} else if gateway != nil {


nirs:
@carterpewpew Can you add a test reproducing this issue?

nirs:
@carterpewpew We are short of reviewers. Please look at the pending PR for 1.39 and help reviewing them:
https://github.com/kubernetes/minikube/milestone/53

pkg/drivers/kic/kic.go
```diff
@@ -73,6 +73,28 @@ func NewDriver(c Config) *Driver {
...
+// containerNetwork returns the network and IP for the container. The network
+// is empty when the container should use the default network. A static IP
+// requested by the user is used even when the network has no gateway, for
+// example when it is a pre-existing network created without --gateway.
+func containerNetwork(gateway net.IP, networkName, staticIP, machineName string) (string, string, error) {
+	if staticIP != "" {
+		return networkName, staticIP, nil
+	}
+	if gateway == nil {
+		return "", "", nil
```
Roslaan001, line 85: If `gateway` is an IPv6 address (e.g. on an IPv6-only Docker network), `gateway.To4()` returns `nil`. Accessing `ip[3]` on line 88 will cause a runtime panic (`nil pointer / slice bounds out of range`).

Could we add a nil check here?

Roslaan001, line 85: If `gateway` is an IPv6 address (e.g. on an IPv6-only Docker network), `gateway.To4()` returns `nil`. Accessing `ip[3]` on line 88 will cause a runtime panic (`nil pointer / slice bounds out of range`).
Could we add a nil check here?

pkg/drivers/kic/kic_test.go
```diff
@@ -0,0 +1,104 @@
...
+			gateway:     nil,
+			networkName: "bridge",
+			machineName: "minikube",
+		},
+		{
+			name:        "too many machines",
+			gateway:     net.ParseIP("192.168.49.253"),
+			networkName: "minikube",
+			machineName: "minikube",
+			wantErr:     true,
```
Roslaan001, line 81: Could we add a test case here verifying that an IPv6 gateway returns an error rather than panicking?

```go
		{
			name:        "IPv6 gateway returns error",
			gateway:     net.ParseIP("2001:db8::1"),
			networkName: "minikube",
			machineName: "minikube",
			wantErr:     true,
		},

Roslaan001 (commented):
Thanks @carterpewpew!

This fixes a nasty silent failure where `--static-ip` on pre-existing networks without an explicit gateway was dropped, silently putting the node on the default bridge (#19284).

Extracting `containerNetwork` makes this clean and easily testable. ileft two quick inline suggestions:
1. Add a check for `if ip == nil` after `gateway.To4()` in `containerNetwork` to guard against nil-slice panics on IPv6 gateways.
2. Add a matching IPv6 test case in `kic_test.go`.

PR #23262, opened 2026-06-29
Add .hugo_build.lock to .gitignore

Wilson-Duncan:
Hugo creates `.hugo_build.lock` during `hugo serve` and `hugo build` to prevent
concurrent builds. If Hugo is interrupted the lock file may be left behind in
the working tree. Add it to `.gitignore` to prevent accidental commits.


nirs:
@Wilson-Duncan Thanks for the fix. I added a PR description to make it more clear. It will help if you add similar description to the commit message to make it easier to review.

Issue #23013, opened 2026-05-21
consider renaming package "log"


medyagh:
consider renaming package "log" 

pkg/libmachine/log/log.go

maybe to mlog (for machine log)

nirs:
Let's close this issue. `log` is the best name for a logger and shadowing the standard library logger is not an issue since we don't use it. Using worse name to make the linter happy does not make sense.

The lint check is disabled for the relevant files and should remain disabled.

PR #23263, opened 2026-06-29
Rename log package to diagnostics

Wilson-Duncan:
This PR is for this issue https://github.com/kubernetes/minikube/issues/23013

### DIFF
--- a/pkg/drivers/common/dhcp/lease.go
+++ b/pkg/drivers/common/dhcp/lease.go
@@ -26,7 +26,7 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/minikube/pkg/libmachine/log"
+	"k8s.io/minikube/pkg/libmachine/diagnostics"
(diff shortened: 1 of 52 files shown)


pkg/drivers/common/dhcp/lease.go
```diff
@@ -65,18 +65,18 @@ func WaitForLease(mac string, timeout time.Duration) (string, error) {
 		return "", err
 	}
 	if detect.NestedVM() {
-		log.Debugf("Nested VM detected, increasing timeout from %s to %s", timeout, timeout*3)
+		diagnostics.Debugf("Nested VM detected, increasing timeout from %s to %s", timeout, timeout*3)
```
nirs, line 68: This is not a good change. Logs should use log, and we should close the issue - it is invalid. There is no issue with shadowing the standard go package since we don't use it.

nirs (changes requested):

PR #23270, opened 2026-07-02
tool: add workflow timeout auto-tuning

### LINKED ISSUE #23043: Auto-tune workflow step timeouts based on historical CI performance

baohuy1303:
Fixes #23043.
Builds on #23042.

This PR adds automation for keeping GitHub Actions step `timeout-minutes` values up to date.

The new workflow runs `hack/workflow-stats` weekly, computes suggested step timeouts from recent successful workflow history, updates matching `timeout-minutes` values, and opens one bot PR with the generated summary.

### What this does

- Adds batch apply mode to `hack/workflow-stats`.
- Reads each workflow's top-level `name:` and fetches matching GitHub Actions history.
- Computes timeouts using the existing `P95 x 3` rule, rounded up to whole minutes with a 1 minute floor.
- Edits only `timeout-minutes` lines, without reserializing workflow YAML.
- Leaves a step unchanged when no matching stats data is available.
- Writes a generated PR body that groups updated workflows, already-up-to-date workflows, and steps left unchanged because stats were unavailable.
- Adds `.github/workflows/auto-tune-timeouts.yml` to run this weekly and open one bot PR per run.

### Missing stats behavior

This intentionally does not use a hard-coded denylist or per-workflow policy.

The tool attempts every workflow with the same 30-day successful-run window. If a step has matching data, it can be updated. If a step has no matching data, it is left unchanged and listed in the generated PR body.

This can happen when:

- The workflow has no successful runs in the last 30 days.
- The workflow is new or only exists on this PR, such as `auto-tune-timeouts.yml`.
- The workflow is rarely triggered and only has useful history outside the 30-day window.
- The step name changed, so old GitHub Actions history no longer matches the current YAML.
- The step only runs in failed/cancelled jobs or conditional paths excluded by the default successful-job filter.

### Scheduled workflow

The scheduled workflow uses the existing minikube bot pattern:

- `secrets.MINIKUBE_BOT_PAT`
- `minikube-bot <minikube-bot@google.com>`
- `peter-evans/create-pull-request`
- `push-to-fork: minikube-bot/minikube`

It also persists the `workflow-stats` SQLite cache with `actions/cache`.

The cache is important because a cold 30-day fetch can be slow. In local validation, `hide-minikube-bot-comments.yml` alone had 1000 runs in the 30-day window and took several minutes to fetch from an empty cache. With the cache restored, future scheduled runs should only fetch new runs incrementally.

### Local validation

Validated with the same command the scheduled workflow uses:

```console
go run ./workflow-stats -since 30 -apply-dir ../.github/workflows -write -pr-body "$RUNNER_TEMP/pr-body.md"
```

Result:

- Ran locally against a temporary copy of `.github/workflows`.
- 4 workflow files were already up to date.
- 44 workflow files were changed in the temporary copy.
- 33 workflow files had one or more steps left unchanged because matching stats were unavailable.
- The diff changed only `timeout-minutes` lines.
- No non-timeout workflow lines changed.
- Compared against the manual timeout PR: the automation produced the same timeout coverage for shared steps.


nirs:
@baohuy1303 Please rebase on master to get rid of the flaky tests failures. I'll try to review this next week.

nirs (changes requested):
Looks you parse and update the yaml manually - this is not the way. This is an also a hug change. We want small and simple changes that we can review quickly.

The foundation for this work is creating yaml helper for parsing and modfying yaml preserving comments.

Using the yaml parser, we should be able to iterate over the workflow and update the timeouts. This should be simple and easy to follow.

Let's start with parser and test proving that we can update or add timeout values without breaking the file.

We will work on the updater and workflow when this is done.

## Parsing yaml preserving comments

1. parse the yaml
2. iterate on the yaml nodes, finding the jobs and steps
3. if a step value needs update, update the yaml node
4. write the yaml to the target file

Here is a draft by Gemini for this:

```go
package main

import (
	"bytes"
	"fmt"
	"os"

	"gopkg.in/yaml.v3"
)

// MapEntry represents a key-value pair to facilitate ranging over mappings.
type MapEntry struct {
	Key   string
	Value *yaml.Node
}

// Entries returns an ordered slice of key-value pairs from a mapping node.
func Entries(node *yaml.Node) []MapEntry {
	if node == nil || node.Kind != yaml.MappingNode {
		return nil
	}
	var entries []MapEntry
	for i := 0; i < len(node.Content); i += 2 {
		entries = append(entries, MapEntry{
			Key:   node.Content[i].Value,
			Value: node.Content[i+1],
		})
	}
	return entries
}

// Elements returns a slice of nodes from a sequence node.
func Elements(node *yaml.Node) []*yaml.Node {
	if node == nil || node.Kind != yaml.SequenceNode {
		return nil
	}
	return node.Content
}

// Get retrieves a child node by key from a mapping node.
func Get(node *yaml.Node, key string) *yaml.Node {
	if node == nil || node.Kind != yaml.MappingNode {
		return nil
	}
	for i := 0; i < len(node.Content); i += 2 {
		if node.Content[i].Value == key {
			return node.Content[i+1]
		}
	}
	return nil
}

// GetString retrieves the string value of a child node.
func GetString(node *yaml.Node, key string) string {
	valNode := Get(node, key)
	if valNode != nil && valNode.Kind == yaml.ScalarNode {
		return valNode.Value
	}
	return ""
}

// Set updates or appends a string key-value pair in a mapping node.
func Set(node *yaml.Node, key, value string) {
	if node == nil || node.Kind != yaml.MappingNode {
		return
	}
	for i := 0; i < len(node.Content); i += 2 {
		if node.Content[i].Value == key {
			node.Content[i+1].Value = value
			return
		}
	}
	node.Content = append(node.Content,
		&yaml.Node{Kind: yaml.ScalarNode, Value: key},
		&yaml.Node{Kind: yaml.ScalarNode, Value: value},
	)
}
```

Using the helpers:

```go
func main() {
	data := []byte(`
name: CI
jobs:
  build:
    steps:
      - name: Checkout
        uses: actions/checkout@v3
      - name: Long running step
        run: sleep 300
`)

	var root yaml.Node
	if err := yaml.Unmarshal(data, &root); err != nil {
		fmt.Fprintf(os.Stderr, "error: %v\n", err)
		os.Exit(1)
	}

	computed := map[string]map[string]string{
		"build": {
			"Checkout":          "2",
			"Long running step": "45",
		},
	}

	// Assuming the document node is the first child of the root node
	if len(root.Content) > 0 {
		doc := root.Content[0]

		for _, job := range Entries(Get(doc, "jobs")) {
			jobTimeouts, hasComputedJob := computed[job.Key]
			if !hasComputedJob {
				continue
			}

			for _, step := range Elements(Get(job.Value, "steps")) {
				stepName := GetString(step, "name")
				
				if newTimeout, ok := jobTimeouts[stepName]; ok {
					Set(step, "timeout-minutes", newTimeout)
				}
			}
		}
	}

	var buf bytes.Buffer
	enc := yaml.NewEncoder(&buf)
	enc.SetIndent(2)
	enc.Encode(&root)
	enc.Close()

	fmt.Print(buf.String())
}
```

baohuy1303:
> Looks you parse and update the yaml manually - this is not the way. This is an also a hug change. We want small and simple changes that we can review quickly.
> 
> The foundation for this work is creating yaml helper for parsing and modfying yaml preserving comments.
> 
> Using the yaml parser, we should be able to iterate over the workflow and update the timeouts. This should be simple and easy to follow.
> 
> Let's start with parser and test proving that we can update or add timeout values without breaking the file.
> 
> We will work on the updater and workflow when this is done.

Hi nirs, thank you for the feedback. I've opened a PR #23361 just for parsing YAML with testing like you requested! I intentionally kept it small for your review

PR #22758, opened 2026-03-14
feat: enable taskstats in ISO

ancoron:
This small change set simply enables the following for the Linux kernel:
* `CONFIG_TASKSTATS`: https://www.kernel.org/doc/html/latest/accounting/taskstats.html
* `CONFIG_TASK_DELAY_ACCT`: https://www.kernel.org/doc/html/latest/accounting/delay-accounting.html (e.g. I/O delay accounting)
* `CONFIG_TASK_XACCT`: extended statistics (e.g. for memory reclaim delays)

This is required for modern observability tooling, often using eBPF to generically capture runtime details.

fixes #22134

PR #23271, opened 2026-07-03
iso/x86_64: enable CONFIG_TASK_IO_ACCOUNTING and related kernel options

### LINKED ISSUE #22737: Enable CONFIG_TASK_IO_ACCOUNTING in Linux kernel config
### What Happened?

[OpenTelemetry process metrics](https://opentelemetry.io/docs/specs/semconv/system/process-metrics/) use a file `/proc/<pid>/io` (via [gopsutil](https://github.com/shirou/gopsutil)) for reading IO stats (see [Host Metrics Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/hostmetricsreceiver)). However, the file is not present on Minikube nodes because the kernel config `CONFIG_TASK_IO_ACCOUNTING` isn't enabled.

Same with [Grafana Beyla](https://github.com/grafana/beyla)'s process metrics, they don't work on Minikube nodes either.

Could you enable it the `CONFIG_TASK_IO_ACCOUNTING`  kernel config?

Maybe related: https://github.com/kubernetes/minikube/issues/22134

### Attach the log file

```
open /proc/1852/io: no such file or directory
```

### Operating System

Redhat/Fedora

### Driver

KVM2

nitishagar:
## What this PR does

Enables the task I/O accounting Kconfig chain in the x86_64 ISO kernel configuration, mirroring what aarch64 already has:

```
CONFIG_TASKSTATS=y
CONFIG_TASK_XACCT=y
CONFIG_TASK_IO_ACCOUNTING=y
```

All three lines are required: `TASK_IO_ACCOUNTING` depends on `TASK_XACCT`, which depends on `TASKSTATS` (init/Kconfig). Without all three, `olddefconfig` silently drops the option.

This makes `/proc/<pid>/io` available on x86_64 ISO nodes, fixing the arch parity gap reported in #22737.

## Changes

- `deploy/iso/minikube-iso/board/minikube/x86_64/linux_x86_64_defconfig`: insert three lines after `CONFIG_BSD_PROCESS_ACCT=y`, byte-identical to aarch64's lines 12–14
- `test/integration/iso_test.go`: add `procIOAccounting` subtest to `TestISOImage` that asserts `/proc/1/io` exists via `minikube ssh`, modeled on the existing `eBPFSupport` subtest

## Notes

- End-to-end verification requires a maintainer to trigger `ok-to-build-iso` (the ISO build is not run in standard CI)
- Prior art: PR #22758 (open, inactive since 2026-03-14, same three options plus `CONFIG_TASK_DELAY_ACCT`). This PR intentionally stays minimal — no `TASK_DELAY_ACCT`, no aarch64 changes. Happy to close in favour of #22758 if it revives.

Fixes #22737


test/integration/iso_test.go
```diff
@@ -139,4 +139,16 @@ func TestISOImage(t *testing.T) {
 			t.Errorf("failed to get info for kernel module %s: args %q: %v", kmod, rr.Command(), err)
 		}
 	})
+
+	t.Run("procIOAccounting", func(t *testing.T) {
+		// Ensure per-task IO accounting is available (https://github.com/kubernetes/minikube/issues/22737)
+		ioFile := "/proc/1/io"
+		rr, err := Run(t, exec.CommandContext(ctx, Target(), "-p", profile, "ssh", fmt.Sprintf("test -f %s && echo 'OK' || echo 'NOT FOUND'", ioFile)))
+		if err != nil {
+			t.Errorf("failed to verify existence of %q file: args %q: %v", ioFile, rr.Command(), err)
```
nirs: If this check fails, the test will continue the next line and log a misleading error. Use t.Fatalf() in both checks.

We have the same issue in the ebpf test. Please don't fix the epbf test in this commit. You can add another commit before or after this one to fix that test.

test/integration/iso_test.go
```diff
@@ -139,4 +139,16 @@ func TestISOImage(t *testing.T) {
...
+		}
+		if !strings.Contains(rr.Stdout.String(), "OK") {
```
nirs: strings.Contains() works but too permissive. We should really compare:

```go
if strigns.TrimSpace(rr.Stdout.String()) != "OK" {
```

Again we have the same issue with the ebpf test, can be part of another commit.

nirs (changes requested):
@nitishagar Thanks for the fix!

nitishagar:
Thanks @nirs! Addressed both points `t.Fatalf` and the exact `strings.TrimSpace(...) != "OK"` compare are now in for `procIOAccounting`, and the matching `eBPFSupport` fix lands in a separate commit as mentioned.

nirs (commented):
@nitishagar Thanks!

/ok-to-build-iso

nirs:
@nitishagar you can assign yourself for the issue using `/assign`

nirs (approved):

PR #23273, opened 2026-07-03
drivers/vfkit,krunkit: add --vmnet-* flags for vmnet network pinning

### LINKED ISSUE #22828: Support --network-cidr for vfkit and krunkit drivers
## The problem

vmnet-helper v0.12.0 will change the default network, matching the system defaults. If `--start-address`, `--end-address`, and `--subnet-mask` are not specified, you get the default network 192.168.64.0/24 instead of the legacy network 192.168.105.0/24.

If the default network is already reserved by another program (e.g Apple container), we will get the next available network (typically 192.168.65.0/24).

This does not affect minikube since we discover the VM IP address. It may affect users that hard code the network address in configuration, or depend on other programs (e.g. lima) that use the legacy network.

See https://github.com/nirs/vmnet-helper/issues/123

## Adding a network CIDR option

If users need to control the network used by minikube, they do not have a way to configure minikube. We can add a single `--network-cidr` flag:

### vfkit driver

```console
minikube start --driver vfkit --network vmnet-shared --network-cidr 192.168.105.0/24
```

### krunkit driver

```console
minikube start --driver krunkit --network-cidr 192.168.105.0/24
```

### Config

To make it easy to configure many clusters, we want to add the option to minikube config.

```console
minikube config set network-cidr 192.168.105.0/24
```

### Implemention

Minikube will parse the CIDR notation using Go's standard library and pass
the right flags to vmnet-helper:

```go
_, network, err := net.ParseCIDR("192.168.105.0/24")

if len(network.IP) != net.IPv4len {
	return fmt.Errorf("vmnet requires an IPv4 CIDR, got %q", cidr)
}

// Start address (gateway): 192.168.105.1
start := make(net.IP, net.IPv4len)
copy(start, network.IP)
start[3] |= 1

// End address: 192.168.105.254
end := make(net.IP, net.IPv4len)
for i := range end {
	end[i] = network.IP[i] | ^network.Mask[i]
}
end[3] &= 0xfe

// Subnet mask: 255.255.255.0
mask := net.IP(network.Mask)
```

```yaml
start-address: "192.168.105.1"
end-address: "192.168.105.254"
subnet-mask: "255.255.255.0"
```

vmnet requires specifying the host address 192.168.105.1 and rejects the network
address 192.168.105.0. Using CIDR notation we improve the user experience by
hiding these details.

### Notes

- The krunkit driver uses --network vmnet-shared implicitly - we may want to change it to accept --network vmnet-shared and reject other values.
- The `--network-cidr` option is generic and could be supported by other drivers.
- The option is not compatible with vmnet-helper `--network` option, that we want to use on macOS 26 if vmnet-broker is installed.
- If the network is reserved by another program (e.g. Apple container) minikube start will fail. Letting vmnet pick the network is the most reliable way.

nitishagar:
## Summary

Add `--vmnet-start-address`, `--vmnet-end-address`, and `--vmnet-subnet-mask` start flags (and matching `minikube config set` keys) that pin the vmnet network for the `vfkit` (with `--network vmnet-shared`) and `krunkit` drivers, mirroring vmnet-helper's own three options directly.

Fixes #22828

## Changes

- **`pkg/drivers/common/vmnet/options.go`** (new, no build tag): per-value validation (netip-parsed IPv4, RFC 1918, contiguous mask) via `NormalizeAddress` / `NormalizeSubnetMask`, and cross-field validation (`validateSemantics`): all-or-none, same subnet, start not network/broadcast, end above start. Also `AddrPtr`/`AddrOrZero`/`AddrString` helpers for the optional-address juggling.
- **`pkg/drivers/common/vmnet/vmnet.go`**: `Helper` gains `StartAddress`/`EndAddress`/`SubnetMask` as `netip.Addr` values (zero = unset); `Helper.Validate()` runs the cross-field checks and is called from `Start()`, so every path that starts a VM validates at the point of use — no command carries its own call.
- **`pkg/minikube/config/types.go`**: `VmnetStartAddress`/`VmnetEndAddress`/`VmnetSubnetMask` as `*netip.Addr` (nil = unset, like `DNSServers`).
- **`cmd/minikube/cmd/start_flags.go`**: `validateVmnetOptions` does driver/network scoping first (warn+clear for unsupported drivers, usage error for vfkit without vmnet-shared), then parses each provided value to `netip.Addr` once at the flag boundary; the typed values flow into the config.
- **`cmd/minikube/cmd/config/config.go`** + **`util.go`** + **`validations.go`**: three `config set` keys using a new `SetIPAddr` setter (stores the canonical spelling); per-key validators reject empty values — `minikube config unset` clears a key.
- **`pkg/minikube/registry/drvs/vfkit/vfkit.go`** + **`krunkit.go`**: populate the `Helper` fields from config (pure passthrough).
- **Docs**: the pinning section lives in a single shared shortcode (`vmnet-pinning`) included by the vfkit and krunkit driver pages, ready for a future qemu include; `start.md`/`config.md` regenerated (`make generate-docs` shape).

## Example run

```sh
minikube start --driver vfkit --network vmnet-shared \
  --vmnet-start-address 192.168.200.1 \
  --vmnet-end-address 192.168.200.127 \
  --vmnet-subnet-mask 255.255.255.0
```

vmnet-helper receives `--start-address 192.168.200.1 --end-address 192.168.200.127 --subnet-mask 255.255.255.0`; with all three unset the argv is byte-identical to today (the vmnet framework allocates the network).

The same values can be persisted with `minikube config set vmnet-start-address ...`, `vmnet-end-address`, and `vmnet-subnet-mask`.

## Review notes (updated per feedback)

- Values are `netip.Addr` end-to-end (Helper, ClusterConfig, and the start path), parsed once at the flag/config boundary, with a `SetIPAddr` config setter — invalid configurations are unrepresentable.
- Validation is at the driver boundary (`Helper.Validate` from `Start()`); per-value checks happen at flag parsing / `config set`; empty values are not storable.
- Machine configuration no longer updates the profile; only the cluster config carries the options.

### DIFF
--- a/cmd/minikube/cmd/config/config.go
+++ b/cmd/minikube/cmd/config/config.go
@@ -91,6 +91,21 @@ var settings = []Setting{
 		validations: []setFn{IsValidCIDR},
 	},
 	{
+		name:        "vmnet-start-address",
+		set:         SetString,
+		validations: []setFn{IsValidVmnetAddress},
+	},
+	{
+		name:        "vmnet-end-address",
+		set:         SetString,
+		validations: []setFn{IsValidVmnetAddress},
+	},
+	{
+		name:        "vmnet-subnet-mask",
+		set:         SetString,
+		validations: []setFn{IsValidVmnetSubnetMask},
+	},
+	{
 		name:        "memory",
 		set:         SetString,
 		validations: []setFn{IsValidMemory},
--- a/cmd/minikube/cmd/config/config.go
+++ b/cmd/minikube/cmd/config/config.go
@@ -91,6 +91,21 @@ var settings = []Setting{
 		validations: []setFn{IsValidCIDR},
 	},
 	{
+		name:        "vmnet-start-address",
+		set:         SetIPAddr,
+		validations: []setFn{IsValidVmnetAddress},
+	},
+	{
+		name:        "vmnet-end-address",
+		set:         SetIPAddr,
+		validations: []setFn{IsValidVmnetAddress},
+	},
+	{
+		name:        "vmnet-subnet-mask",
+		set:         SetIPAddr,
+		validations: []setFn{IsValidVmnetSubnetMask},
+	},
+	{
 		name:        "memory",
 		set:         SetString,
 		validations: []setFn{IsValidMemory},
--- a/cmd/minikube/cmd/node_start.go
+++ b/cmd/minikube/cmd/node_start.go
@@ -22,6 +22,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"k8s.io/minikube/cmd/minikube/cmd/flags"
+	"k8s.io/minikube/pkg/drivers/common/vmnet"
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/exit"
 	"k8s.io/minikube/pkg/minikube/machine"
@@ -46,6 +47,14 @@ var nodeStartCmd = &cobra.Command{
 
 		api, cc := mustload.Partial(ClusterFlagValue(), options)
 		name := args[0]
+
+		// Validate vmnet options right after loading the config (INV-C): any value
+		// handed to the pkg/minikube layer must already be valid. A persisted partial
+		// config (an interrupted `config set` or a hand-edited config) is caught here
+		// before it reaches node.Provision->configure() (R7).
+		if err := vmnet.ValidateOptions(cc.VmnetStartAddress, cc.VmnetEndAddress, cc.VmnetSubnetMask); err != nil {
+			exit.Message(reason.Usage, "invalid vmnet options: {{.err}}", out.V{"err": err})
+		}
 
 		n, _, err := node.Retrieve(*cc, name)
 		if err != nil {
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -152,6 +152,9 @@ const (
 	preloadSrc              = "preload-source"
 	rosetta                 = "rosetta"
 	vmnetOffloading         = "vmnet-offloading"
+	vmnetStartAddress       = "vmnet-start-address"
+	vmnetEndAddress         = "vmnet-end-address"
+	vmnetSubnetMask         = "vmnet-subnet-mask"
 	dnsServers              = config.DNSServers
 	mdns                    = config.MDNS
 )
@@ -300,6 +303,13 @@ func initDriverFlags() {
 
 	// krunkit
 	startCmd.Flags().Bool(vmnetOffloading, false, "Enable vmnet checksum and TSO offloading. See krunkit driver documentation for known limitations (krunkit driver only)")
+
+	// vfkit (vmnet-shared) and krunkit: pin the vmnet network. All three must be
+	// set together, or none (empty keeps current behavior, letting the vmnet
+	// framework allocate the network). Help text mirrors vmnet-helper's run -h.
+	startCmd.Flags().String(vmnetStartAddress, "", "The starting IPv4 address for the vmnet interface, used as the gateway address. The subsequent address up to and including --vmnet-end-address are placed in the DHCP pool. All other addresses are available for static assignment. The address must be in the private IP range (RFC 1918). Must be set together with --vmnet-end-address and --vmnet-subnet-mask (vfkit with --network vmnet-shared, and krunkit drivers only)")
+	startCmd.Flags().String(vmnetEndAddress, "", "The DHCP IPv4 range end address for the vmnet interface. Must be in the private IP range (RFC 1918) and in the same subnet as --vmnet-start-address. Must be set together with --vmnet-start-address and --vmnet-subnet-mask (vfkit with --network vmnet-shared, and krunkit drivers only)")
+	startCmd.Flags().String(vmnetSubnetMask, "", "The IPv4 subnet mask to use on the vmnet interface. Must be set together with --vmnet-start-address and --vmnet-end-address (vfkit with --network vmnet-shared, and krunkit drivers only)")
 }
 
 // initNetworkingFlags inits the commandline flags for connectivity related flags for start
@@ -610,6 +620,36 @@ func getVmnetOffloading(driverName string) bool {
 	return enabled
 }
 
+// validateVmnetOptions is the start-time validator for the three vmnet network
+// flags. It does driver/network scoping (R9) first, then delegates the
+// all-or-none (R5), per-value (R3), and cross-field (R4) checks to
+// vmnet.ValidateOptions (R6: the single shared validator).
+//
+// Scoping is done first so an unsupported driver is warned+cleared without
+// surfacing value errors for flags that are ignored. It returns the three
+// values, possibly cleared to "" when the driver cannot honor them (so an
+// unsupported-driver drop is reflected in the assembled config). On a usage
+// violation it exits with reason.Usage (R7).
+func validateVmnetOptions(drvName, network, start, end, mask string) (string, string, string) {
+	// R9: driver/network scoping (cmd-specific). Done first so unsupported drivers
+	// are warned+cleared without surfacing value errors for flags that are ignored.
+	if !driver.IsVFKit(drvName) && !driver.IsKrunkit(drvName) {
+		if start != "" || end != "" || mask != "" {
+			out.WarningT("--vmnet-* flags are only valid with the vfkit and krunkit drivers, they will be ignored")
+			return "", "", ""
+		}
+		return start, end, mask
+	}
+	// krunkit always uses vmnet, so only vfkit is gated on the effective network.
+	if driver.IsVFKit(drvName) && network != "vmnet-shared" {
+		exit.Message(reason.Usage, "--vmnet-* flags with the vfkit driver require --network vmnet-shared")
+	}
+	if err := vmnet.ValidateOptions(start, end, mask); err != nil {
+		exit.Message(reason.Usage, "invalid vmnet options: {{.err}}", out.V{"err": err})
+	}
+	return start, end, mask
+}
+
 // getDNSServers returns the configured DNS servers for VM drivers.
 // For non-VM drivers the value is ignored since DNS configuration via
 // systemd-resolved is not applicable.
@@ -673,13 +713,19 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 
 	checkExtraDiskOptions(cmd, drvName)
 
+	// Validate the three vmnet options against the effective network + driver
+	// (R5/R9). Done once here so the populated config already reflects any
+	// unsupported-driver drop. May exit with reason.Usage (R7).
+	network := getNetwork(drvName, options)
+	vmnetStart, vmnetEnd, vmnetMask := validateVmnetOptions(drvName, network, viper.GetString(vmnetStartAddress), viper.GetString(vmnetEndAddress), viper.GetString(vmnetSubnetMask))
+
 	cc = config.ClusterConfig{
 		Name:                    ClusterFlagValue(),
 		KeepContext:             viper.GetBool(keepContext),
 		EmbedCerts:              viper.GetBool(embedCerts),
 		MinikubeISO:             viper.GetString(isoURL),
 		KicBaseImage:            viper.GetString(kicBaseImage),
-		Network:                 getNetwork(drvName, options),
+		Network:                 network,
 		Subnet:                  viper.GetString(subnet),
 		Memory:                  getMemorySize(cmd, drvName),
 		CPUs:                    getCPUCount(drvName),
@@ -759,6 +805,9 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 		AutoPauseInterval:  viper.GetDuration(autoPauseInterval),
 		Rosetta:            getRosetta(drvName),
 		VmnetOffloading:    getVmnetOffloading(drvName),
+		VmnetStartAddress:  vmnetStart,
+		VmnetEndAddress:    vmnetEnd,
+		VmnetSubnetMask:    vmnetMask,
 		DNSServers:         getDNSServers(cmd, drvName),
 		MDNS:               getMDNS(cmd, drvName),
 	}
@@ -993,6 +1042,9 @@ func updateExistingConfigFromFlags(cmd *cobra.Command, existing *config.ClusterC
 	updateDurationFromFlag(cmd, &cc.AutoPauseInterval, autoPauseInterval)
 	updateBoolFromFlag(cmd, &cc.Rosetta, rosetta)
 	updateBoolFromFlag(cmd, &cc.VmnetOffloading, vmnetOffloading)
+	updateStringFromFlag(cmd, &cc.VmnetStartAddress, vmnetStartAddress)
+	updateStringFromFlag(cmd, &cc.VmnetEndAddress, vmnetEndAddress)
+	updateStringFromFlag(cmd, &cc.VmnetSubnetMask, vmnetSubnetMask)
 
 	if cmd.Flags().Changed(kubernetesVersion) {
 		kubeVer, err := getKubernetesVersion(existing)
@@ -1032,6 +1084,11 @@ func updateExistingConfigFromFlags(cmd *cobra.Command, existing *config.ClusterC
 	if cc.ScheduledStop != nil && time.Until(time.Unix(cc.ScheduledStop.InitiationTime, 0).Add(cc.ScheduledStop.Duration)) <= 0 {
 		cc.ScheduledStop = nil
 	}
+
+	// Validate the merged vmnet options on restart (R5): a partial set left in
+	// the persisted config, or a flag overriding one of three, must be caught
+	// before start (R7). Re-assigns in case of an unsupported-driver drop.
+	cc.VmnetStartAddress, cc.VmnetEndAddress, cc.VmnetSubnetMask = validateVmnetOptions(cc.Driver, cc.Network, cc.VmnetStartAddress, cc.VmnetEndAddress, cc.VmnetSubnetMask)
 
 	return cc
 }
--- a/cmd/minikube/cmd/start_flags.go
+++ b/cmd/minikube/cmd/start_flags.go
@@ -152,6 +152,9 @@ const (
 	preloadSrc              = "preload-source"
 	rosetta                 = "rosetta"
 	vmnetOffloading         = "vmnet-offloading"
+	vmnetStartAddress       = "vmnet-start-address"
+	vmnetEndAddress         = "vmnet-end-address"
+	vmnetSubnetMask         = "vmnet-subnet-mask"
 	dnsServers              = config.DNSServers
 	mdns                    = config.MDNS
 )
@@ -300,6 +303,13 @@ func initDriverFlags() {
 
 	// krunkit
 	startCmd.Flags().Bool(vmnetOffloading, false, "Enable vmnet checksum and TSO offloading. See krunkit driver documentation for known limitations (krunkit driver only)")
+
+	// vfkit (vmnet-shared) and krunkit: pin the vmnet network. All three must be
+	// set together, or none (empty keeps current behavior, letting the vmnet
+	// framework allocate the network). Help text mirrors vmnet-helper's run -h.
+	startCmd.Flags().String(vmnetStartAddress, "", "The starting IPv4 address for the vmnet interface, used as the gateway address. The subsequent address up to and including --vmnet-end-address are placed in the DHCP pool. All other addresses are available for static assignment. The address must be in the private IP range (RFC 1918). Must be set together with --vmnet-end-address and --vmnet-subnet-mask (vfkit with --network vmnet-shared, and krunkit drivers only)")
+	startCmd.Flags().String(vmnetEndAddress, "", "The DHCP IPv4 range end address for the vmnet interface. Must be in the private IP range (RFC 1918) and in the same subnet as --vmnet-start-address. Must be set together with --vmnet-start-address and --vmnet-subnet-mask (vfkit with --network vmnet-shared, and krunkit drivers only)")
+	startCmd.Flags().String(vmnetSubnetMask, "", "The IPv4 subnet mask to use on the vmnet interface. Must be set together with --vmnet-start-address and --vmnet-end-address (vfkit with --network vmnet-shared, and krunkit drivers only)")
 }
 
 // initNetworkingFlags inits the commandline flags for connectivity related flags for start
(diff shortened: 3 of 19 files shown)


nirs:
@nitishagar Thanks for this change, but need to revise the way it works. See https://github.com/kubernetes/minikube/issues/22828#issuecomment-4876941470

nitishagar:
Thanks @nirs. I've revised the PR to mirror vmnet-helper's three options directly. The old single `--network-cidr` surface is fully removed.

nirs:
> Thanks @nirs. I've revised the PR to mirror vmnet-helper's three options directly. The old single `--network-cidr` surface is fully removed.

Thanks! can you update the description to describe the current change and add example run with the the address options?

translations/de.json
nirs, line 1: You don't need to generate docs, this is done automatically after merge.

pkg/minikube/node/vmnet_test.go
```diff
@@ -0,0 +1,144 @@
...
+package node
+
+import (
+	"strings"
+	"testing"
+
+	"k8s.io/minikube/pkg/minikube/config"
+)
+
+func TestValidateVmnetConfig(t *testing.T) {
```
nirs: Please split valid and invalid configuration to different tests. Valid tests are simpler (no error). Invalid tests are simpler - all have an error.

pkg/drivers/common/vmnet/options_test.go
```diff
@@ -0,0 +1,250 @@
...
+			}
+		})
+	}
+}
+
+func TestIsValidVmnetAddress(t *testing.T) {
+	tests := []struct {
+		val      string
+		wantErr  bool
+		errMatch string
```
nirs, line 293: Split valid and invalidate tests

pkg/drivers/common/vmnet/options_test.go
```diff
@@ -0,0 +1,250 @@
...
+			t.Errorf("IsValidVmnetAddress(%q): unexpected error: %v", tt.val, err)
+		}
+	}
+}
+
+func TestIsValidVmnetSubnetMask(t *testing.T) {
+	tests := []struct {
+		val      string
+		wantErr  bool
+		errMatch string
```
nirs, line 334: Split valida and invalid tests

pkg/minikube/node/start.go
```diff
@@ -655,8 +656,52 @@ func setupKubeconfig(h host.Host, cc config.ClusterConfig, n config.Node, cluste
 	return kcs
 }
 
+// ValidateVmnetConfig is the defense-in-depth chokepoint for the vmnet options.
+// It enforces all-or-none (R5) and the full cross-field checks (R3/R4) by
+// delegating to vmnet.ValidateOptions (R6: the single shared validator).
```
nirs: node is not the right place to validate options. Options must be validated right when we load them in the cmd package. Any value passed to the minikube package must be already valid.

Ideally all vmnet validation is part of the vment package. Spreading validation code in multiple packages and duplicating the tests is not maintainable.

nirs (commented):
I only looked quickly but this adds too much code and tests in multiple locations.

We should have:
- One place validation all flags and config options related to vmnet.
- One test for validation flags and config options
- Commands should use the same code to validation flags, options, config

nirs:
@nitishagar We are short of reviewers. Please look at the pending PR for 1.39 and help reviewing them:
https://github.com/kubernetes/minikube/milestone/53

site/content/en/docs/drivers/krunkit.md
```diff
@@ -77,6 +77,27 @@ users or other groups.
 minikube start --driver krunkit
 ```
 
+### Pinning the vmnet network
```
nirs: Duplicating this section in vfkit and krunit is too much. We want to enable vmnet-shared for qemu, and we already have duplicate content.

Lets use include file to have single vment-shared documentation and include it in vfkit and krunkit (and later in qemu).

pkg/drivers/common/vmnet/options_test.go
```diff
@@ -0,0 +1,294 @@
...
+*/
+
+package vmnet
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestValidateOptions_Valid(t *testing.T) {
```
nirs: Use sub tests for nicer structure:

```go
func TestValidateOptions(t *testing.T) {
    t.Run("valid", func(t *testing.T) {
        ...
    }
    t.Run("invalid", func(t *testing.T) {
        ...
    }
}

File: pkg/drivers/common/vmnet/options_test.go [hunk 7]
```diff
@@ -0,0 +1,294 @@
...
+	tests := []struct {
+		name  string
+		start string
+		end   string
+		mask  string
+	}{
+		// --- valid triples (all three set, all checks pass) ---
+		{
+			name:  "valid 192.168 range (maintainer example)",
```
nirs: We can remove "valid" from the name since the test name include "valid"

pkg/drivers/common/vmnet/options_test.go
```diff
@@ -0,0 +1,294 @@
...
+			errMatch: "not a valid IPv4",
+		},
+		{
+			name: "IPv6 mask rejected", start: "192.168.1.1", end: "192.168.1.10", mask: "ffff:ffff:ffff:ffff::",
+			errMatch: "not a valid IPv4",
+		},
+
+		// --- public / non-RFC-1918 IPv4 rejected (R3, OQ2) ---
+		{
+			name: "public start rejected", start: "8.8.8.8", end: "192.168.1.10", mask: "255.255.255.0",
```
nirs: We have 2 issues here:
- start address is not in private ragne
- end address not in same network

So we don't know if we rejected the value because of the first or second issues. We need one test with valid public range for first issue, and another with 2 private addresses not in the same network.

pkg/drivers/common/vmnet/options.go
```diff
@@ -0,0 +1,216 @@
...
+// networks are constrained to RFC 1918 by vmnet-helper, so non-private
+// addresses are rejected early. net.IP.IsPrivate() is deliberately not used:
+// it also matches CGN (100.64/10), loopback, link-local and IPv6 ULA, none of
+// which are valid vmnet addresses.
+var rfc1918Nets = mustParseCIDRs("10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16")
+
+func mustParseCIDRs(cidrs ...string) []*net.IPNet {
+	nets := make([]*net.IPNet, len(cidrs))
+	for i, cidr := range cidrs {
+		_, n, err := net.ParseCIDR(cidr)
```
nirs: `n` is net.IPNet, use `net` or `network`

pkg/drivers/common/vmnet/options.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+		nets[i] = n
+	}
+	return nets
+}
+
+func isRFC1918(ip net.IP) bool {
+	for _, n := range rfc1918Nets {
+		if n.Contains(ip) {
+			return true
```
nirs, line 56: `network.Contains(ip)`?

Code should be joy to read.

pkg/drivers/common/vmnet/options.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+	}
+	return false
+}
+
+// ValidateOptions validates the three vmnet-helper options together. It is the
+// single shared validator (R6): the start-flag path (via validateVmnetOptions)
```
nirs: What is R6?

pkg/drivers/common/vmnet/vmnet.go
```diff
@@ -68,6 +68,14 @@ type Helper struct {
 	// Offloading is required for krunkit, does not work with vfkit.
 	Offloading bool
 
+	// StartAddress/EndAddress/SubnetMask pin the vmnet network by passing
+	// --start-address/--end-address/--subnet-mask to vmnet-helper. Empty means
+	// vmnet-helper picks the network (current behavior). Only used by the vfkit
+	// (vmnet-shared) and krunkit drivers.
+	StartAddress string
+	EndAddress   string
+	SubnetMask   string
```
nirs: Why use string and not a specific type? Using a string means that it is possible to create invalid configuration. If we use netip.Addr we make many errors impossible.

This type can also be use for the config - like dns-servers. When parsing flags, we can convert the string to netip.Addr.

We can do similar change later for other options (like InterfaceID).

pkg/drivers/common/vmnet/vmnet.go
```diff
@@ -191,6 +199,13 @@ func (h *Helper) Start(socketPath string) error {
 		args = append(args, "--interface-id", h.InterfaceID)
 	}
 
+	// Pin the vmnet network when configured. Guarded on StartAddress so an
+	// all-empty config produces byte-identical argv to today (vmnet framework
+	// allocates the network). The three are validated together before start.
```
nirs, line 221: Depending on client code validation seems reasonable, but we can also do the validation here. The advantage is that we have single place validation the values, and this will fail the flow early enough - before we start the helper and vm process.

Let's add:

```go
func (h *Helper) Validate() error
```

This will replace ValidateOptions() or call it if there is a reason to keep it.

This will be called from Start() to ensure we cannot start the helper or vm with invalid options, and we get helpful error in the command.

This is early enough to give good user experience, without duplicating validation calls in all flows creating vms:

- start
- node add
- node start
- future commands

pkg/drivers/common/vmnet/options_test.go
```diff
@@ -0,0 +1,294 @@
...
+		{
+			name:  "end below subnet broadcast leaves room for static assignment",
+			start: "192.168.1.1", end: "192.168.1.10", mask: "255.255.255.0",
+		},
+		{
+			name:  "valid /24 with end two below broadcast",
+			start: "192.168.0.1", end: "192.168.0.253", mask: "255.255.255.0",
+		},
+		{
+			name:  "valid /16",
```
nirs: "valid" is part of test name, maybe "172.16.0.1/16"?

pkg/drivers/common/vmnet/options.go
```diff
@@ -0,0 +1,216 @@
...
+// and the node-command config-load path both delegate to it.
+//
+// It enforces the all-or-none gate (R5), the per-value IPv4/RFC 1918/mask
+// checks (R3), and the cross-field same-subnet/ordering/broadcast checks
+// (R4/OQ3/OQ6). An all-empty triple short-circuits to nil (R8: default path
+// stays behavior-identical). The per-value setFn wrappers (IsValidVmnetAddress
+// /IsValidVmnetSubnetMask, used by `config set`, which sees one key at a time)
+// call validateAddress/validateMask directly and are unaffected by the
+// all-or-none gate here.
```
nirs: Let document here what is R3, R6 and other cryptic values. I guess these refer to the RFC, so we can have something like:

```go
// Rules:
// - R3 - description URL
...
```

pkg/drivers/common/vmnet/options.go
```diff
@@ -0,0 +1,216 @@
...
+	if mask != "" {
+		if err := validateMask(mask); err != nil {
+			return err
+		}
+	}
+
+	// Cross-field checks require all three values (R4, OQ3, OQ6).
+	if start == "" || end == "" || mask == "" {
+		return nil
+	}
```
nirs, line 102: Move to start - if the values are all empty, we are done.

pkg/drivers/common/vmnet/options.go
```diff
@@ -0,0 +1,216 @@
...
+// and the node-command config-load path both delegate to it.
+//
+// It enforces the all-or-none gate (R5), the per-value IPv4/RFC 1918/mask
+// checks (R3), and the cross-field same-subnet/ordering/broadcast checks
+// (R4/OQ3/OQ6). An all-empty triple short-circuits to nil (R8: default path
+// stays behavior-identical). The per-value setFn wrappers (IsValidVmnetAddress
+// /IsValidVmnetSubnetMask, used by `config set`, which sees one key at a time)
+// call validateAddress/validateMask directly and are unaffected by the
+// all-or-none gate here.
+func ValidateOptions(start, end, mask string) error {
```
nirs: return here id all are unset

pkg/drivers/common/vmnet/options.go
```diff
@@ -0,0 +1,216 @@
...
+	// All-or-none (R5). The three options must be set together; an all-empty
+	// triple is valid (R8, the default path), but a partial triple is not. The
+	// message is context-neutral since the validator is called from both the
+	// flag and config-load contexts.
+	if n := nonEmptyCount(start, end, mask); n != 0 && n != 3 {
+		return errors.New("vmnet-start-address, vmnet-end-address, and vmnet-subnet-mask must be set together")
+	}
+
```
nirs, line 98: The following checks will be simper if we already handled the empty case.

pkg/drivers/common/vmnet/options.go
```diff
@@ -0,0 +1,216 @@
...
+	// Cross-field checks require all three values (R4, OQ3, OQ6).
+	if start == "" || end == "" || mask == "" {
+		return nil
+	}
+
+	sip := net.ParseIP(start).To4()
+	eip := net.ParseIP(end).To4()
+	m := net.IPMask(net.ParseIP(mask).To4())
```
nirs: This convert the address to IPv4 twice - once in validateAddress(), and again here. Instead we should return the valid address from valdiateAddress().

pkg/drivers/common/vmnet/options.go
```diff
@@ -0,0 +1,216 @@
...
+		if v != "" {
+			n++
+		}
+	}
+	return n
+}
+
+// validateAddress runs the per-value checks for a start/end address: valid IPv4
+// and within an RFC 1918 private range (R3).
+func validateAddress(val, label string) error {
```
nirs: return the address so we don't need to do this gain. But we should really get netip.Addr - the value is already parsed by flag and config code, so here we just ned to convert to IPv4.

pkg/drivers/common/vmnet/options.go
```diff
@@ -0,0 +1,216 @@
...
+		return fmt.Errorf("%s %q is not in the RFC 1918 private range (10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16)", label, val)
+	}
+	return nil
+}
+
+// validateMask runs the per-value checks for the subnet mask: valid IPv4 and a
+// contiguous netmask (R3, OQ4). A non-contiguous mask (e.g. 255.255.255.1) is
+// rejected because net.IPMask.Size reports it with bits != 32.
+func validateMask(val string) error {
+	ip := net.ParseIP(val).To4()
```
nirs: Use the modern netip.ParseAddr().

pkg/drivers/common/vmnet/options.go
```diff
@@ -0,0 +1,216 @@
...
+	return nil
+}
+
+// IsValidVmnetAddress is a per-value validator for the `config set` path, which
+// can see only one key's value at a time. It runs only the per-value IPv4 + RFC
+// 1918 checks (no cross-field); the same-subnet, ordering and all-or-none checks
+// are enforced at `minikube start` via ValidateOptions. An empty value is valid
+// here so that an in-progress config sequence (one of three keys set) can be
+// persisted and completed.
+func IsValidVmnetAddress(_, val string) error {
```
nirs, line 185: Looks good to reuse the validation at config time. Will catch one errors early.

cmd/minikube/cmd/config/config.go
```diff
@@ -90,6 +90,21 @@ var settings = []Setting{
...
 		validations: []setFn{IsValidCIDR},
 	},
+	{
+		name:        "vmnet-start-address",
+		set:         SetString,
+		validations: []setFn{IsValidVmnetAddress},
+	},
+	{
+		name:        "vmnet-end-address",
+		set:         SetString,
```
nirs: We need to add SetIPAddr - we don't want to save values as string.

pkg/drivers/common/vmnet/options.go
```diff
@@ -0,0 +1,216 @@
...
+	if val == "" {
+		return nil
+	}
```
nirs: why empty is valid? there is reason to save an empty value in the config.

pkg/drivers/common/vmnet/options.go
```diff
@@ -0,0 +1,216 @@
...
+	return validateAddress(val, "vmnet address")
+}
+
+// IsValidVmnetSubnetMask is a per-value validator for the `config set` path,
+// running only the per-value IPv4 + contiguous-mask checks.
+func IsValidVmnetSubnetMask(_, val string) error {
+	if val == "" {
+		return nil
+	}
```
nirs: no need to allow empty value

pkg/minikube/config/types.go
```diff
@@ -112,6 +112,9 @@ type ClusterConfig struct {
 	AutoPauseInterval       time.Duration // Specifies interval of time to wait before checking if cluster should be paused
 	Rosetta                 bool          // Only used by vfkit driver
 	VmnetOffloading         bool          // Only used by krunkit driver
+	VmnetStartAddress       string        // Only used by the vfkit (vmnet-shared) and krunkit drivers.
+	VmnetEndAddress         string        // Only used by the vfkit (vmnet-shared) and krunkit drivers.
+	VmnetSubnetMask         string        // Only used by the vfkit (vmnet-shared) and krunkit drivers.
```
nirs: Use netip.Addr like DNSServers

cmd/minikube/cmd/node_start.go
```diff
@@ -47,6 +48,14 @@ var nodeStartCmd = &cobra.Command{
 		api, cc := mustload.Partial(ClusterFlagValue(), options)
 		name := args[0]
 
+		// Validate vmnet options right after loading the config (INV-C): any value
+		// handed to the pkg/minikube layer must already be valid. A persisted partial
+		// config (an interrupted `config set` or a hand-edited config) is caught here
+		// before it reaches node.Provision->configure() (R7).
+		if err := vmnet.ValidateOptions(cc.VmnetStartAddress, cc.VmnetEndAddress, cc.VmnetSubnetMask); err != nil {
+			exit.Message(reason.Usage, "invalid vmnet options: {{.err}}", out.V{"err": err})
+		}
```
nirs: This works, but it requires duplicating the validation in 3 commands, remembering to add this to future commands that start machines.

It also add validation for vment options to all drivers when vment is used only by 2 drivers on macOS. We will have to add vment.ValidateOptions() stub (it is compiled only on drawing).

Lets move validation to vfkit and krunkit Start() method. 

We will have 2 kinds of validations:
- per value validation - at the flag/config level
- semantic validation - at the driver level, calling vmnet Helper.Validate()

cmd/minikube/cmd/start_flags.go
```diff
@@ -610,6 +620,36 @@ func getVmnetOffloading(driverName string) bool {
...
+		if start != "" || end != "" || mask != "" {
+			out.WarningT("--vmnet-* flags are only valid with the vfkit and krunkit drivers, they will be ignored")
+			return "", "", ""
+		}
+		return start, end, mask
+	}
+	// krunkit always uses vmnet, so only vfkit is gated on the effective network.
+	if driver.IsVFKit(drvName) && network != "vmnet-shared" {
+		exit.Message(reason.Usage, "--vmnet-* flags with the vfkit driver require --network vmnet-shared")
+	}
```
nirs, line 650: We select vmnet-shared by if possible in visit - so this validation must come after selected the network.

cmd/minikube/cmd/start_flags.go
```diff
@@ -610,6 +620,36 @@ func getVmnetOffloading(driverName string) bool {
...
+	if err := vmnet.ValidateOptions(start, end, mask); err != nil {
+		exit.Message(reason.Usage, "invalid vmnet options: {{.err}}", out.V{"err": err})
+	}
```
nirs, line 672: This good - but validation at the driver level will simplify this code.

For example if if we don't need to validated all values here, we can use a simpler function to get private ip address from viper:

```go
func getVmnetAddr(key, network string) *netip.Addr {
    warn if driver does not use vmnet...
    warn if driver does not use vmnet-shared network...

    addr, err := netip.ParseAddr(viper.Get(key))
    ...
    return vmnet.ValidedVmnetAddress(addr)
}
```

cmd/minikube/cmd/start_flags.go
```diff
@@ -610,6 +620,36 @@ func getVmnetOffloading(driverName string) bool {
...
+// flags. It does driver/network scoping (R9) first, then delegates the
+// all-or-none (R5), per-value (R3), and cross-field (R4) checks to
+// vmnet.ValidateOptions (R6: the single shared validator).
+//
+// Scoping is done first so an unsupported driver is warned+cleared without
+// surfacing value errors for flags that are ignored. It returns the three
+// values, possibly cleared to "" when the driver cannot honor them (so an
+// unsupported-driver drop is reflected in the assembled config). On a usage
+// violation it exits with reason.Usage (R7).
+func validateVmnetOptions(drvName, network, start, end, mask string) (string, string, string) {
```
nirs: return netip.Addr so once the value is parsed from the flags the application use specific type.

cmd/minikube/cmd/start_flags.go
```diff
@@ -673,13 +713,19 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 
 	checkExtraDiskOptions(cmd, drvName)
 
+	// Validate the three vmnet options against the effective network + driver
+	// (R5/R9). Done once here so the populated config already reflects any
+	// unsupported-driver drop. May exit with reason.Usage (R7).
+	network := getNetwork(drvName, options)
+	vmnetStart, vmnetEnd, vmnetMask := validateVmnetOptions(drvName, network, viper.GetString(vmnetStartAddress), viper.GetString(vmnetEndAddress), viper.GetString(vmnetSubnetMask))
```
nirs, line 743: If we switch to per-value validation here we don't need this call.

cmd/minikube/cmd/start_flags.go
```diff
@@ -759,6 +805,9 @@ func generateNewConfigFromFlags(cmd *cobra.Command, k8sVersion string, rtime str
 		AutoPauseInterval:  viper.GetDuration(autoPauseInterval),
 		Rosetta:            getRosetta(drvName),
 		VmnetOffloading:    getVmnetOffloading(drvName),
+		VmnetStartAddress:  vmnetStart,
```
nirs: With per-value validation will be:

```go
    VmnetStartAddress: getVmnetAdddr(vmnetStartAddress, network)
```

cmd/minikube/cmd/start_flags.go
```diff
@@ -993,6 +1042,9 @@ func updateExistingConfigFromFlags(cmd *cobra.Command, existing *config.ClusterC
 	updateDurationFromFlag(cmd, &cc.AutoPauseInterval, autoPauseInterval)
 	updateBoolFromFlag(cmd, &cc.Rosetta, rosetta)
 	updateBoolFromFlag(cmd, &cc.VmnetOffloading, vmnetOffloading)
+	updateStringFromFlag(cmd, &cc.VmnetStartAddress, vmnetStartAddress)
+	updateStringFromFlag(cmd, &cc.VmnetEndAddress, vmnetEndAddress)
+	updateStringFromFlag(cmd, &cc.VmnetSubnetMask, vmnetSubnetMask)
```
nirs: Not needed - machine configuration should not update the profile. These updates are not effective - the machine configuration cannot be modified after the machine is created.

The existing for VmentOffloading and Rosetta are a bug we need to fix later.

cmd/minikube/cmd/start_flags.go
```diff
@@ -1033,6 +1085,11 @@ func updateExistingConfigFromFlags(cmd *cobra.Command, existing *config.ClusterC
 		cc.ScheduledStop = nil
 	}
 
+	// Validate the merged vmnet options on restart (R5): a partial set left in
+	// the persisted config, or a flag overriding one of three, must be caught
+	// before start (R7). Re-assigns in case of an unsupported-driver drop.
+	cc.VmnetStartAddress, cc.VmnetEndAddress, cc.VmnetSubnetMask = validateVmnetOptions(cc.Driver, cc.Network, cc.VmnetStartAddress, cc.VmnetEndAddress, cc.VmnetSubnetMask)
```
nirs: Not needed - the values in the profile json never used after the machine is created. Moving validation to Driver.Start() avoid this.

nirs (changes requested):
@nitishagar Thanks, this is awesome work - well documented and tested.

Having single function validation the options is great, but we still to call this 5 times from 3 commands.

I suggest to simplify to:

- per value validation in flags parsing and config set. Parsing and setting netip.Addr instead of strings.
- semantic validation in vmnet.Helper.Validate() - called from vfkit and krunkit Start()

Start() is early enough to give good user experience, and will be easier to maintain. For example if we add new command starting vm, no change will be needed.

nitishagar:
> @nitishagar We are short of reviewers. Please look at the pending PR for 1.39 and help reviewing them: https://github.com/kubernetes/minikube/milestone/53

Thanks @nirs;  picking these up. I scanned the open queue for 1.39-bound work. Planning to review: 
- #23083 (image-load OOM, fixes #17785)
- #23366 (minikube doctor, no reviews yet)
- #23319 (NodeProvisioner refactor)

I'll skip the ones already in a changes-requested loop with you unless you'd like a second pass. Shout if you'd rather I prioritize differently.

nirs:
@nitishagar Thanks! this will be very helpful!

> * [feat: implement minikube doctor command #23366](https://github.com/kubernetes/minikube/pull/23366) (minikube doctor, no reviews yet)

This one is not included in minikube 1.39, and has no issue explaining why we need it. This kind of changes need discussion before spend time on reviewing big changes. But of course if this looks interesting to you are invited to review.

nirs:
@nitishagar Turns out we need this now for CI on macOS 26: #23477

nitishagar:
Thanks a lot for the thorough review @nirs, and sorry it took me a while to circle back to this one.

I've pushed a rework along the lines you suggested. The addresses are now netip.Addr all the way through — Helper and the cluster config both hold the typed value, parsed once when the flags are read, so an invalid configuration can't really be constructed anymore. config set goes through a new SetIPAddr setter and no longer accepts empty values, minikube config unset is the way to clear a key.

Validation stayed at the driver level like you preferred (Helper.Validate from Start()), so every path that starts a VM is covered without each command carrying its own call. I also dropped the profile updates entirely — only the cluster config carries the options now.

The docs section is a single shared include between the vfkit and krunkit pages now, so we can reuse it for qemu whenever we get there. Test names lost the redundant "valid" prefixes too. Description is updated with an example run. Let me know if this looks closer to what you had in mind.

PR #23309, opened 2026-07-06
pod-test

alessandrostefano27:


nirs:
@alessandrostefano27 Please start with an issue describing the problem you want to solve, and describe the solution.

You can reopen this PR after we have an agreement on the proposed solution.

PR #23253, opened 2026-06-28
addons: fix Helm verification failure inside guest VMs

Roslaan001:
## Description

This PR fixes a bug where minikube fails to enable Helm-based addons (such as Traefik) on virtualised environments (e.g. KVM) where `helm` is not pre-installed on the host. 

When the host lacks Helm, minikube attempts to install Helm inside the guest VM using the official `get-helm-3` script. However, this script executes a post-install verification check (`hash helm`). Since `/usr/local/bin` is not on the guest's default non-interactive `PATH` when executing commands under `sudo bash -c`, the script fails and exits with status 1.

By prepending `export PATH=$PATH:/usr/local/bin` to the installation command, the validation succeeds, allowing minikube to successfully complete the installation and move the binary to `/usr/bin/helm`.

Fixes https://github.com/kubernetes/minikube/issues/23252

## Section/Component affected

* `pkg/addons/helm.go`

## Verification Results

Verified locally by starting a minikube cluster and simulating the restricted non-interactive guest `PATH` (`/usr/bin:/bin`):

### Without this fix (this fails):

```bash
minikube ssh -p addons-033061  -- "sudo PATH=/usr/bin:/bin bash -c 'curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 && chmod 700 get_helm.sh && ./get_helm.sh'"
```

**Output:**

```
[WARNING] Could not find git. It is required for plugin installation.

Helm v3.21.2 is already latest
helm not found. Is /usr/local/bin on your $PATH?
Failed to install helm
```

### With this fix (this works):

```bash
minikube ssh -p addons-033061  -- "sudo PATH=/usr/bin:/bin bash -c 'export PATH=\$PATH:/usr/local/bin && curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 && chmod 700 get_helm.sh && ./get_helm.sh'"
```

**Output:**

```
[WARNING] Could not find git. It is required for plugin installation.

Helm v3.21.2 is already latest (Succeeds)
```

PR #23326, opened 2026-07-10
   Fix flaky TestHelmInstall: use minimum version floor

Roslaan001:
**What:** Replace exact latest version matching in `TestHelmInstall` with a `>=` semver comparison against a minimum expected version floor.
  
**Why:** During Helm release rollouts, CDN propagation delays can cause the test's HTTP query to `helm3-latest-version` and the `get_helm.sh` install script to see different versions (e.g., `v3.21.3` vs `v3.21.2`), leading to flaky
  test failures.
  
**How:**
- Add a `minExpectedHelmVersion` const (`v3.20.0`), bumped occasionally (not on every release), to catch the CDN serving something wildly stale or broken without needing to track the exact current latest
- Replace exact version checks with `semver.Compare >= minExpectedHelmVersion`
- Remove `getLatestHelmVersion()` function and its unused imports (`io`, `net/http`, `strings`)
- Add `t.Logf` for all version retrievals to aid debugging

Replace exact latest version matching with a semver >= comparison against a minimum expected version floor (v3.20.0). This avoids flaky failures during Helm release rollouts when CDN propagation causes the test's HTTP query and get_helm.sh to see different versions.

Changes:
- Add minExpectedHelmVersion const (bumped occasionally, not every release)
- Replace exact version checks with semver.Compare >= minimum
- Add t.Logf for all version retrievals to aid debugging
- Remove getLatestHelmVersion() and its unused imports

Fixes #23323

PR #23329, opened 2026-07-10
addons: use blang/semver for helm version validation

### LINKED ISSUE #23327: Replace "golang.org/x/mod/semver" with "github.com/blang/semver/v4"
The project is using "github.com/blang/semver/v4", which provides more ergonomic interface.

We added "golang.org/x/mod/semver" by mistake in #23253, instead of using the existing dependency.

Notes:
- "github.com/blang/semver/v4" requires stripping the "v" prefix

@Roslaan001 do you want to take it?

Roslaan001:
Replace `golang.org/x/mod/semver` with the existing `github.com/blang/semver/v4` dependency in `pkg/addons/helm.go`, as raised in #23327.

`golang.org/x/mod/semver` was introduced in #23253 for Helm version validation, but the project already depends on `github.com/blang/semver/v4`, which provides a more ergonomic interface.

`blang/semver` requires a strict `x.y.z` format (`vx.y.z` is not valid semver), so the leading `v` is stripped before parsing, matching how existing code in the repo uses the package (e.g. `pkg/drivers/common/vmnet/vmnet.go`).

fixes https://github.com/kubernetes/minikube/issues/23327

### DIFF
--- a/pkg/addons/helm.go
+++ b/pkg/addons/helm.go
@@ -23,7 +23,7 @@ import (
 	"path"
 	"strings"
 
-	"golang.org/x/mod/semver"
+	"github.com/blang/semver/v4"
 	"k8s.io/klog/v2"
 	"k8s.io/minikube/pkg/minikube/assets"
 	"k8s.io/minikube/pkg/minikube/command"
@@ -99,11 +99,13 @@ func InstallHelm(runner command.Runner, opts HelmOptions) error {
 		HELM_INSTALL_DIR=/usr/bin ./get_helm.sh`
 
 	if opts.Version != "" {
-		if !semver.IsValid(opts.Version) {
-			return fmt.Errorf("invalid helm version %q: must be a valid semver (e.g. v3.12.0)", opts.Version)
+		// semver.Parse does not support "v" prefix.
+		parsed, err := semver.Parse(strings.TrimPrefix(opts.Version, "v"))
+		if err != nil {
+			return fmt.Errorf("invalid helm version %q: %w", opts.Version, err)
 		}
 		klog.Infof("Installing helm version %s", opts.Version)
-		script += " --version " + opts.Version
+		script += " --version v" + parsed.String()
 	} else {
 		klog.Info("Installing helm latest version")
 	}
--- a/test/integration/helm_test.go
+++ b/test/integration/helm_test.go
@@ -21,10 +21,11 @@ package integration
 import (
 	"context"
 	"os/exec"
+	"strings"
 	"testing"
 	"time"
 
-	"golang.org/x/mod/semver"
+	"github.com/blang/semver/v4"
 	"k8s.io/minikube/pkg/addons"
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/machine"
@@ -39,7 +40,7 @@ import (
 // See https://github.com/kubernetes/minikube/issues/23323
 // Bumped occasionally (not on every release) to catch the CDN serving something
 // wildly stale or broken, without needing to track the exact current latest.
-const minExpectedHelmVersion = "v3.20.0"
+var minExpectedHelmVersion = semver.Version{Major: 3, Minor: 20}
 
 // TestHelmInstall integration test verifies helm installation, upgrade, and no-change behavior inside a live guest VM.
 // We test this flow because installing the latest helm allows self-healing and updating the cluster when Helm is missing, outdated, or broken.
@@ -106,9 +107,13 @@ func TestHelmInstall(t *testing.T) {
 		}
 
 		version := addons.HelmVersion(runner)
-		t.Logf("installed helm version: %s", version)
-		if semver.Compare(version, minExpectedHelmVersion) < 0 {
-			t.Fatalf("installed helm version %q is older than minimum expected %q", version, minExpectedHelmVersion)
+		t.Logf("installed helm version: %q", version)
+		parsed, err := semver.Parse(strings.TrimPrefix(version, "v"))
+		if err != nil {
+			t.Fatalf("failed to parse helm version: %q: %v", version, err)
+		}
+		if parsed.LT(minExpectedHelmVersion) {
+			t.Fatalf("installed helm version %q is older than minimum expected %q", parsed, minExpectedHelmVersion)
 		}
 	})
 
@@ -117,7 +122,7 @@ func TestHelmInstall(t *testing.T) {
 	// calling InstallHelm with HelmOptions{} (defaults to latest) will correctly upgrade it to the latest version.
 	t.Run("Upgrade", func(t *testing.T) {
 		// Install minExpectedHelmVersion directly to /usr/bin/helm
-		err := addons.InstallHelm(runner, addons.HelmOptions{Version: minExpectedHelmVersion})
+		err := addons.InstallHelm(runner, addons.HelmOptions{Version: minExpectedHelmVersion.String()})
 		if err != nil {
 			t.Fatalf("failed to install helm %s: %v", minExpectedHelmVersion, err)
 		}
@@ -125,7 +130,11 @@ func TestHelmInstall(t *testing.T) {
 		// Verify the pinned helm version was installed and is executable
 		versionOld := addons.HelmVersion(runner)
 		t.Logf("installed helm version: %s (expected %s)", versionOld, minExpectedHelmVersion)
-		if versionOld != minExpectedHelmVersion {
+		parsedOld, err := semver.Parse(strings.TrimPrefix(versionOld, "v"))
+		if err != nil {
+			t.Fatalf("failed to parse helm version: %q: %v", versionOld, err)
+		}
+		if parsedOld.NE(minExpectedHelmVersion) {
 			t.Fatalf("helm version mismatch: expected %q, got %q", minExpectedHelmVersion, versionOld)
 		}
 
@@ -138,11 +147,12 @@ func TestHelmInstall(t *testing.T) {
 		// Verify that a newer version of helm has been installed in /usr/bin/helm
 		versionNew := addons.HelmVersion(runner)
 		t.Logf("upgraded helm version: %s (from %s)", versionNew, versionOld)
-		if versionNew == versionOld {
-			t.Fatalf("helm version was not upgraded: still %q", versionNew)
+		parsedNew, err := semver.Parse(strings.TrimPrefix(versionNew, "v"))
+		if err != nil {
+			t.Fatalf("failed to parse helm version: %q: %v", versionNew, err)
 		}
-		if semver.Compare(versionNew, minExpectedHelmVersion) < 0 {
-			t.Fatalf("upgraded helm version %q is older than minimum expected %q", versionNew, minExpectedHelmVersion)
+		if parsedNew.LTE(minExpectedHelmVersion) {
+			t.Fatalf("installed version %q not newer than older version %q", parsedNew, minExpectedHelmVersion)
 		}
 	})
 
@@ -151,7 +161,7 @@ func TestHelmInstall(t *testing.T) {
 	// running InstallHelm again is a no-op and does not modify or reinstall it.
 	t.Run("NoChange", func(t *testing.T) {
 		// Run InstallHelm to ensure the pinned helm version is present
-		err := addons.InstallHelm(runner, addons.HelmOptions{Version: minExpectedHelmVersion})
+		err := addons.InstallHelm(runner, addons.HelmOptions{Version: minExpectedHelmVersion.String()})
 		if err != nil {
 			t.Fatalf("InstallHelm failed: %v", err)
 		}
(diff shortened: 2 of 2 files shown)


nirs:
@Roslaan001 lets wait until #23326 is merged, than replace the semver package in that change.

Roslaan001:
@nirs, I tested with a small code;

```bash
package addons

import (
  "strings"
  "testing"

  "github.com/blang/semver/v4"
)

func TestSemverCheckDemo(t *testing.T) {
  for _, v := range []string{"v3.12.0", "3.12.0", "v3.12", "3", "latest", "", "not-a-version", "v3.12.0-rc.1"} {
    _, err := semver.Parse(strings.TrimPrefix(v, "v"))
    if err != nil {
      t.Logf("%-16q REJECTED  (%v)", v, err)
    } else {
      t.Logf("%-16q ACCEPTED", v)
    }
  }
}
```

and the output is;
```bash
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ v3.12.0       │ ✅ ACCEPTED                                                                                                        │
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 3.12.0        │ ✅ ACCEPTED                                                                                                        │
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ v3.12.0-rc.1  │ ✅ ACCEPTED (pre-release ok)                                                                                       │
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ v3.12         │ ❌ REJECTED — no patch                                                                                             │
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 3             │ ❌ REJECTED                                                                                                        │
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ latest        │ ❌ REJECTED                                                                                                        │
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ ""            │ ❌ REJECTED (empty), although InstallHelm skips validation when version is empty, since "" means "install the latest" │
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ not-a-version │ ❌ REJECTED

**Comment by nirs** [comment]
> @nirs, I tested with a small code;

Nice, expected behavior.

But testing without stripping will be better, it will show that "v3.12.0" is rejected.

**Review by nirs — COMMENTED** [review]
Wait for #23326

**Comment by Roslaan001** [comment]
> > @nirs, I tested with a small code;
> 
> Nice, expected behavior.
> 
> But testing without stripping will be better, it will show that "v3.12.0" is rejected.

Okay, should i test without stripping then?

**Comment by nirs** [comment]
> > But testing without stripping will be better, it will show that "v3.12.0" is rejected.
> 
> Okay, should i test without stripping then?

If you want to see the real behavior

**Comment by Roslaan001** [comment]
> But testing without stripping will be better, it will show that "v3.12.0" is rejected.

@nirs , Tested it with this now;

```bash
package addons

import (
  "strings"
  "testing"

  "github.com/blang/semver/v4"
)

func TestSemverStripVsNoStrip(t *testing.T) {
  for _, v := range []string{"v3.12.0", "3.12.0", "v3.12.0-rc.1"} {
    // Without stripping the leading "v".
    _, errNoStrip := semver.Parse(v)
    // With stripping (what helm.go does).
    _, errStrip := semver.Parse(strings.TrimPrefix(v, "v"))

    t.Logf("%-16q  no-strip: %-40v  strip: %v", v, result(errNoStrip), result(errStrip))
  }
}

func result(err error) string {
  if err != nil {
    return "REJECTED (" + err.Error() + ")"
  }
  return "ACCEPTED"
}
```

and the output;

```bash
┌──────────────┬───────────────────────────────────────────────────────────────┬─────────────────────────────────────────────┐
│    Input     │                   No strip: semver.Parse(v)                   │ With strip: semver.Parse(TrimPrefix(v,"v")) │
├──────────────┼───────────────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
│ v3.12.0      │ ❌ REJECTED — Invalid character(s) found in major number "v3" │ ✅ ACCEPTED                                 │
├──────────────┼───────────────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
│ 3.12.0       │ ✅ ACCEPTED                                                   │ ✅ ACCEPTED                                 │
├──────────────┼───────────────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
│ v3.12.0-rc.1 │ ❌ REJECTED — Invalid character(s) found in major number "v3" │ ✅ ACCEPTED                                 │
└──────────────┴───────────────────────────────────────────────────────────────┴───────────────

**Comment by nirs** [comment]
@Roslaan001 please rebase on master and replace semver also in helm_test.go

**Comment by Roslaan001** [comment]
> please rebase on master and replace semver also in helm_test.go

Hi @nirs, there is no semver in helm_test.go. Or did you mean that the semver validation added to InstallHelm should also have test coverage in helm_test.go?

**Comment by Roslaan001** [comment]
@nirs, Rebase on master is done

**Comment by nirs** [comment]
> > please rebase on master and replace semver also in helm_test.go
> 
> Hi @nirs, there is no semver in helm_test.go. 

https://github.com/kubernetes/minikube/blob/85d8639e1273a105a666f01b92101a409ef90eb4/test/integration/helm_test.go#L27

File: pkg/addons/helm.go [hunk 1]
```diff
@@ -99,7 +99,8 @@ func InstallHelm(runner command.Runner, opts HelmOptions) error {
 		HELM_INSTALL_DIR=/usr/bin ./get_helm.sh`
 
 	if opts.Version != "" {
-		if !semver.IsValid(opts.Version) {
+		// semver requires x.y.z, so strip the leading "v" from versions like "v3.12.0".
```
nirs: Comment not needed. Look at other code using semver and make this code similar.

pkg/drivers/common/vmnet is a good example.

pkg/addons/helm.go
```diff
@@ -99,7 +99,8 @@ func InstallHelm(runner command.Runner, opts HelmOptions) error {
...
+		if _, err := semver.Parse(strings.TrimPrefix(opts.Version, "v")); err != nil {
 			return fmt.Errorf("invalid helm version %q: must be a valid semver (e.g. v3.12.0)", opts.Version)
```
nirs: We accept both valid semver (3.12.0) and semver with v prefix (v3.12.0).

Check how other code reports invalid semver.

Check if get-helm-3 works with valid semver without the v prefix. If it requires the prefix, we should require it here.

nirs (changes requested):
Replace also in helm_test.go

pkg/addons/helm.go
Roslaan001: > Check if get-helm-3 works with valid semver without the v prefix. If it requires the prefix, we should require it here.

 get-helm-3 actually accepts a missing v prefix (auto-prepends it), which was the justification for accepting both 3.12.0 and v3.12.0 here

test/integration/helm_test.go
nirs: We want to have semver.Version to make it easy to compare in the tests.

```go
var minExpectedHelmVersion = semver.Version{Major: 3, Minor: 20}
```

This is better than MustParse, eliminating runtime error and avoiding the issue of "v" prefix.

test/integration/helm_test.go
```diff
@@ -107,7 +108,7 @@ func TestHelmInstall(t *testing.T) {
 
 		version := addons.HelmVersion(runner)
 		t.Logf("installed helm version: %s", version)
-		if semver.Compare(version, minExpectedHelmVersion) < 0 {
+		if semver.MustParse(strings.TrimPrefix(version, "v")).LT(semver.MustParse(strings.TrimPrefix(minExpectedHelmVersion, "v"))) {
 			t.Fatalf("installed helm version %q is older than minimum expected %q", version, minExpectedHelmVersion)
```
nirs: MustParse on minExpectedHelmVersion is not needed since we can use a semver.Version.

MustParse on HelmVersion() wil panic instead of failing the test.

Use:

```go
		version := addons.HelmVersion(runner)
		t.Logf("installed helm version: %q", version)
        parsed, err := semver.Parse(strings.TrimPrefix(version, "v"))
        if err != nil {
            t.Fatalf("failed to parse helm version: %q: %v", version, err)
        }
```

This is little bit verbose, but we can improve it later.

Now the assert become much more clear:

```go
        if parsed.LT(minExpectedHelmVersion) {
            t.Fatalf("installed helm version %q is older than minimum expected %q", parsed, minExpectedHelmVersion)
        }
```

test/integration/helm_test.go
```diff
@@ -141,7 +142,7 @@ func TestHelmInstall(t *testing.T) {
 		if versionNew == versionOld {
 			t.Fatalf("helm version was not upgraded: still %q", versionNew)
 		}
-		if semver.Compare(versionNew, minExpectedHelmVersion) < 0 {
+		if semver.MustParse(strings.TrimPrefix(versionNew, "v")).LT(semver.MustParse(strings.TrimPrefix(minExpectedHelmVersion, "v"))) {
```
nirs: Handle like the preivous test

nirs (changes requested):

Roslaan001:
@nirs, check it out now

test/integration/helm_test.go
```diff
@@ -117,15 +122,19 @@ func TestHelmInstall(t *testing.T) {
...
 
 		// Verify the pinned helm version was installed and is executable
 		versionOld := addons.HelmVersion(runner)
 		t.Logf("installed helm version: %s (expected %s)", versionOld, minExpectedHelmVersion)
-		if versionOld != minExpectedHelmVersion {
+		parsedOld, err := semver.Parse(strings.TrimPrefix(versionOld, "v"))
+		if err != nil {
+			t.Fatalf("failed to parse helm version: %q: %v", versionOld, err)
+		}
+		if !parsedOld.EQ(minExpectedHelmVersion) {
```
nirs:
```suggestion
		if parsedOld.NE(minExpectedHelmVersion) {
```

test/integration/helm_test.go
```diff
@@ -141,8 +150,12 @@ func TestHelmInstall(t *testing.T) {
 		if versionNew == versionOld {
 			t.Fatalf("helm version was not upgraded: still %q", versionNew)
 		}
```
nirs: We can delete tis check - it can be covered in the next check.

test/integration/helm_test.go
```diff
@@ -141,8 +150,12 @@ func TestHelmInstall(t *testing.T) {
...
-		if semver.Compare(versionNew, minExpectedHelmVersion) < 0 {
-			t.Fatalf("upgraded helm version %q is older than minimum expected %q", versionNew, minExpectedHelmVersion)
+		parsedNew, err := semver.Parse(strings.TrimPrefix(versionNew, "v"))
+		if err != nil {
+			t.Fatalf("failed to parse helm version: %q: %v", versionNew, err)
+		}
+		if parsedNew.LT(minExpectedHelmVersion) {
+			t.Fatalf("upgraded helm version %q is older than minimum expected %q", parsedNew, minExpectedHelmVersion)
```
nirs: This can cover both cases (same or older version).

```suggestion
		if parsedNew.LTE(minExpectedHelmVersion) {
		        t.Fatalf("installed version %q not newer than older version %q", parsedNew, minExpectedHelmVersion)
```

pkg/addons/helm.go
```diff
@@ -99,8 +99,9 @@ func InstallHelm(runner command.Runner, opts HelmOptions) error {
 		HELM_INSTALL_DIR=/usr/bin ./get_helm.sh`
 
 	if opts.Version != "" {
-		if !semver.IsValid(opts.Version) {
-			return fmt.Errorf("invalid helm version %q: must be a valid semver (e.g. v3.12.0)", opts.Version)
+		// semver.Parse does not support "v" prefix.
+		if _, err := semver.Parse(strings.TrimPrefix(opts.Version, "v")); err != nil {
```
nirs: keep the parsed version here, and add the "v" prefix when we add it to the config. We should always add the official version and should not trust get-helm-3 script. This is not a public documented API but implementation detail.

pkg/addons/helm.go
```diff
@@ -99,8 +99,9 @@ func InstallHelm(runner command.Runner, opts HelmOptions) error {
...
+			return fmt.Errorf("invalid helm version %q: %w", opts.Version, err)
 		}
 		klog.Infof("Installing helm version %s", opts.Version)
 		script += " --version " + opts.Version
```
nirs:
```suggestion
		script += " --version v" + parsed.String()
```

nirs (changes requested):

Roslaan001:
@nirs, I have fixed it now

nirs (approved):
Thanks!

/lgtm

nirs (approved):

nirs:
@Roslaan001 no need to retest, this job is not required.

PR #23361, opened 2026-07-14
tool: add YAML node helpers for workflow timeout updates

baohuy1303:
## Summary
A part of #23043 

Adds a small YAML helper foundation for future workflow timeout updates.

This PR parses GitHub Actions workflow YAML through `yaml.Node`, iterates workflow jobs and steps, updates matching `timeout-minutes` values, and writes the YAML back while preserving comments. It does not add GitHub workflow stats collection, PR body generation, or scheduled automation.

## What changed

- Added YAML node helpers for ordered mappings, sequences, scalar lookup, scalar updates, and encoding.
- Added error handling for missing keys and wrong YAML node types instead of silently returning empty values.
- Added a small workflow timeout update flow:
  - parse workflow YAML
  - iterate `jobs`
  - iterate each job's `steps`
  - match steps by `name`, unnamed `uses`, or unnamed `run`
  - add or update `timeout-minutes`
  - encode the workflow YAML back
- Added tests comparing full input and expected output workflow YAML.
- Added tests for malformed or unexpected workflow YAML shapes.
- Moved `gopkg.in/yaml.v3` to a direct dependency in the `hack` module.

## Tests

```sh
cd hack
go test ./workflow-stats
```


hack/workflow-stats/yaml_test.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	steps := Elements(Get(job, "steps"))
+	if len(steps) == 0 {
+		t.Fatalf("job %q has no steps", jobName)
+	}
+	return steps[0]
+}
+
+func TestYAMLNodeHelpers(t *testing.T) {
+	root := parseYAML(t, `
```
nirs: This should test the real interface - get rid of all the test helpers accepting testing.T.

The most important purpose of this test is to show the client code, checking that the yaml helper provides an easy to use an clear API.

hack/workflow-stats/yaml_test.go
```diff
@@ -0,0 +1,216 @@
...
+name: CI
+jobs:
+  build:
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+`)
+	doc := documentNode(t, root)
```
nirs: same here, use the actual function/type. No test helpers when we want to evaluate the client code.

hack/workflow-stats/yaml_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+	if got := GetString(doc, "name"); got != "CI" {
+		t.Fatalf("GetString(name) = %q, want CI", got)
+	}
```
nirs, line 58: Looks good - but this cannot detect wrong type (e.g. name is a list or object).

If we return value, err we can handle missing key or wrong type. The value is used only if we found it, it can be zero value on errors.

hack/workflow-stats/yaml_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+	jobs := Entries(Get(doc, "jobs"))
```
nirs: This suspicious - what do we get when there is not "jobs" key or the return value is not a list?

I don't think we want to use such code in the updater, we will use:

```go
jobs := Get(doc, "jobs")
handle errors...

for job := range Entries(jobs) {

}
```

hack/workflow-stats/yaml_test.go
```diff
@@ -0,0 +1,216 @@
...
+func TestSetScalarAddsTimeoutMinutes(t *testing.T) {
+	root := parseYAML(t, `
+name: CI
+jobs:
+  build:
+    steps:
+      - name: Long running step
+        run: sleep 300
+`)
+	step := firstWorkflowStep(t, documentNode(t, root), "build")
```
nirs: Don't hide the API with these helpers. We want to see how real code will look like.

I think the issue you are trying to avoid here is how to not repeat the code checking the parser yaml. But the real updater code will be very simple - same code for every workflow - iterating on all steps in all jobs.

So these tests should be arranged in a different way.

Write the code to iterated over steps and update the timeout once:

```go
updateTimeouts([]byte, desired map[string]int) ([]byte, err)
```

This will be become later the real update function.

All tests will have input and output yaml. The test will compare the updated yaml with the expected yaml, and show a unified diff if the yaml do not match.

We can keep the yaml as yaml files in test/data to make the tests easier to work with.

You can write small unit tests to verify that various helpers work correctly, but the most important test is validating the input and output yaml. This validates that can update the timeouts, and we don't change anything else.

nirs (commented):
@baohuy1303 Thanks for focusing on a smaller problem first. This is much easier to review. However I reviewed only some of the tests, and I think need to change the way we test to focus on the actual update flow, and avoid lot of unneeded test helpers that hide the actual code.

Also commented on the API that does not have great error handling.

I'll spend on this more time this weekend.

baohuy1303:
> @baohuy1303 Thanks for focusing on a smaller problem first. This is much easier to review. However I reviewed only some of the tests, and I think need to change the way we test to focus on the actual update flow, and avoid lot of unneeded test helpers that hide the actual code.
> 
> Also commented on the API that does not have great error handling.
> 
> I'll spend on this more time this weekend.

Thanks for the feedback nirs, I've rewritten the code so that it handles file testing and better error handling. What I plan to do next is actually extract the timeout-minutes from our workflow-stats tool to update actual YAML files

hack/workflow-stats/workflow_test.go
```diff
@@ -0,0 +1,128 @@
...
+			if _, err := UpdateTimeouts([]byte(tt.input), StepTimeouts{}); err == nil {
+				t.Fatalf("UpdateTimeouts succeeded, want error containing %q", tt.wantErr)
+			} else if !strings.Contains(err.Error(), tt.wantErr) {
+				t.Fatalf("UpdateTimeouts error = %q, want containing %q", err, tt.wantErr)
+			}
+		})
+	}
+}
+
+func unifiedDiff(want, got string) string {
```
nirs: We already use a library that does unified diff. If what we use is not good enough we can add github.com/aymanbagabas/go-udiff.

hack/workflow-stats/testdata/workflow_expected.yml
```diff
@@ -0,0 +1,33 @@
...
+        timeout-minutes: 3
+      - run: |
+          make build
+          make test
+        timeout-minutes: 8
+      - name: Test
+        timeout-minutes: 15 # timeout comment
+        run: make test
+      - name: Already tuned
+        timeout-minutes: 3
```
nirs: Lets check also comment before timeout-minutes and comment after timeout-minutes.

Keeping tiny files, each testing one aspect will make it easier to debug issues.

hack/workflow-stats/yaml.go
```diff
@@ -109,3 +121,30 @@ func EncodeYAML(node *yaml.Node) ([]byte, error) {
 	}
 	return buf.Bytes(), nil
 }
+
+func wantKind(node *yaml.Node, kind yaml.Kind) error {
```
nirs: wantKind() is surprising name for function validating node kind. Maybe `validateKind()`?

hack/workflow-stats/yaml.go
```diff
@@ -18,20 +18,25 @@ package main
 
 import (
 	"bytes"
+	"errors"
+	"fmt"
```
nirs, line 22: Please move change sin this commit to the first commit. It is hard to review and not useful history when we fix issues in the first commit in the second commit.

hack/workflow-stats/yaml_test.go
nirs, line 1: Why not keep trivial unit tests of the yaml helpers? I want to finish with the yaml helpers before reviewing on the workflow.

nirs (commented):
I like the split to yaml and workflow! But let's finish the yaml commit first.

baohuy1303:
> I like the split to yaml and workflow! But let's finish the yaml commit first.

Hi @nirs I've split the commits and added the tests back in! Sorry for taking so long, I was kinda packed with school :)

baohuy1303:
hi @nirs can you review this so i could move to the next ask:)

PR #23363, opened 2026-07-14
Validate memory requirements for VM drivers

### LINKED ISSUE #23317: mikube start fail when --memory is lower than 2500M - missing input validation
### What Happened?

🔥  Creating kvm2 VM (CPUs=2, Memory=1726MB, Disk=20000MB) ...
😿  Failed to start kvm2 VM. Running "minikube delete" may fix it: creating host: create: creating: starting domain: waiting for IP: domain minikube didn't return IP after 1m30s

❌  Exiting due to GUEST_PROVISION: error provisioning guest: Failed to start host: creating host: create: creating: starting domain: waiting for IP: domain minikube didn't return IP after 1m30s

### Attach the log file

🔥  Creating kvm2 VM (CPUs=2, Memory=1726MB, Disk=20000MB) ...
😿  Failed to start kvm2 VM. Running "minikube delete" may fix it: creating host: create: creating: starting domain: waiting for IP: domain minikube didn't return IP after 1m30s

❌  Exiting due to GUEST_PROVISION: error provisioning guest: Failed to start host: creating host: create: creating: starting domain: waiting for IP: domain minikube didn't return IP after 1m30s

### Operating System

None

### Driver

None

coyaSONG:
## What changed

- use a 2500 MB minimum for VM drivers while preserving the existing 1800 MB minimum for container, bare-metal, and SSH drivers
- warn VM-driver users below the tested 3072 MB recommendation
- add focused coverage for VM and non-VM requirement selection

## Why

VM drivers currently reuse the general 1800/1900 MB thresholds. That allows configurations such as KVM2 with 2048 MB to reach VM provisioning even though they cannot start reliably, which surfaces later as an unrelated-looking IP timeout.

Before:

```text
minikube start --driver=kvm2 --memory=2048
...
waiting for IP: domain minikube didn't return IP after 1m30s
```

After this change, the same request exits during resource validation because 2048 MB is below the 2500 MB VM minimum. Requests from 2500 MB through 3071 MB continue with the existing low-memory warning, now using the 3072 MB VM recommendation.

Fixes #23317

## Validation

- `go test ./cmd/minikube/cmd -count=1`
- `go test ./pkg/minikube/driver -count=1`
- `TESTSUITE=unittest make test` with the repository-selected Go 1.26.4 toolchain
- `go vet ./cmd/minikube/cmd`
- `make gofmt`
- `TESTSUITE=boilerplate make test`
- `make out/minikube`
- repository-configured `golangci-lint` on `./cmd/minikube/cmd`: 0 issues

The repository-wide lint invocation reports four existing `revive` findings in untouched vfkit, krunkit, and parallels files. No changed file is involved.

This contribution was assisted by OpenAI Codex and manually reviewed and validated by the contributor.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -1170,6 +1170,13 @@ func suggestMemoryAllocation(sysLimit, containerLimit, nodes int) int {
 	return suggested
 }
 
+func memoryRequirements(drvName string) (minimum, recommended int) {
+	if driver.IsVM(drvName) && !driver.IsSSH(drvName) {
+		return minVMUsableMem, minVMRecommendedMem
+	}
+	return minUsableMem, minRecommendedMem
+}
+
 // validateRequestedMemorySize validates the memory size matches the minimum recommended
 func validateRequestedMemorySize(req int, drvName string) {
 	// TODO: Fix MB vs MiB confusion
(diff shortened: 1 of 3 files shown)


nirs:
@coyaSONG Thanks! You need to sign the CLA to continue.
https://github.com/kubernetes/minikube/pull/23363#issuecomment-4972685253

cmd/minikube/cmd/start_test.go
```diff
@@ -304,6 +304,30 @@ func TestSuggestMemoryAllocation(t *testing.T) {
 	}
 }
 
+func TestMemoryRequirements(t *testing.T) {
```
nirs: This tests a trivial unexported two-branch function. The cost/value tradeoff doesn't look right:

- Hardcodes driver names. Adding or removing a driver silently leaves the test
  without coverage or breaks it.
- Duplicates memory constants (2500, 3072, 1800, 1900). If the constants change,
  the test fails with an opaque mismatch rather than catching a real bug.
- Doesn't cover all drivers
- Doesn't test actual behavior. The real behavior lives in
  `validateRequestedMemorySize`, which rejects or warns based on these
  thresholds. This test only checks that the helper returns the right numbers,
  not that those numbers actually gate anything.

The meaningful test would be against the validation logic with synthetic inputs
(e.g. "VM driver requests 2000 MiB on an 8 GiB system, rejected"), but that
requires decoupling `validateRequestedMemorySize` from `memoryLimits` first,
which is out of scope here.

Until then, this test is mostly maintenance burden. Please remove it.

cmd/minikube/cmd/start_flags.go
```diff
@@ -114,6 +114,8 @@ const (
 	nativeSSH               = "native-ssh"
 	minUsableMem            = 1800 // Kubernetes (kubeadm) will not start with less
 	minRecommendedMem       = 1900 // Warn at no lower than existing configurations
+	minVMUsableMem          = 2500 // VM drivers cannot start reliably with less
+	minVMRecommendedMem     = 3072 // Default and tested VM allocation
```
nirs: Good comment, but would be useful to mention that we warn about lower limit like minRecommendedMem.

cmd/minikube/cmd/start_flags.go
```diff
@@ -114,6 +114,8 @@ const (
 	nativeSSH               = "native-ssh"
 	minUsableMem            = 1800 // Kubernetes (kubeadm) will not start with less
 	minRecommendedMem       = 1900 // Warn at no lower than existing configurations
+	minVMUsableMem          = 2500 // VM drivers cannot start reliably with less
```
nirs: Linking to the issue here will be useful. This helps to explain why this value exists.

cmd/minikube/cmd/start.go
nirs, line 1: Please add proper commit message explaining this change. The PR description is not a replacement for commit message.

Include free -h output from vm with minikube memory (2500) and recommended memory (3072).

nirs (changes requested):
@coyaSONG Thank you for this fix.

nirs:
@coyaSONG are you planning to complete this PR? If not someone else can pick it up.

coyaSONG:
@nirs I addressed the review in `9c9ed1ae` and have now resolved the four review threads:

- removed the maintenance-heavy helper test
- documented both the 2500 MiB VM minimum and 3072 MiB warning threshold
- linked the threshold to #23317
- included the requested 2500/3072 MiB `free -h` measurements in the commit message

The focused packages, full unit suite, vet, formatting, boilerplate, binary build, and changed-scope golangci-lint validation are recorded in the PR description. When convenient, could you re-review the updated draft so the stale `CHANGES_REQUESTED` state can be refreshed?

coyaSONG:
Thanks for the review and follow-up. On re-audit, #23507 is now the maintainer-selected active implementation for #23317 and is receiving the requested VM validation, so I am closing this older parallel PR to avoid splitting review effort. The exact final head remains preserved at coyaSONG/minikube@9c9ed1aebeefeb0dbf50047f207b81588e5a3bae for reference or reuse.

PR #23348, opened 2026-07-12
helm: improve error handling, type safety, and test clarity

nirs:
Clean up the helm addon code and integration test to improve error
handling, type safety, and test readability.

## Consistent exit codes across all runners

All runners now return exit code 127 for missing binaries. SSHRunner
and kicRunner already did this naturally (via shell / docker exec), but
execRunner did not — it calls execve directly with no shell, so a
missing binary returned ExitCode 0. This broke `HelmVersion` on the
none driver.

## HelmVersion error handling

`HelmVersion` now returns `(semver.Version, error)` instead of a raw
string. When helm is not installed (the common case — `/usr/bin/` is on
a non-persistent filesystem and lost on every cluster stop/start), the
error is short and clear:

    helm is not installed: bash: /usr/bin/helm: No such file or directory

Unexpected failures return the full runner error with stdout/stderr for
debugging:

    failed to run helm version: /usr/bin/helm version ...: exit status 1
    stdout:
    ...
    stderr:
    ...

Previously, errors were silently dropped and `"minikube logs"` showed
no evidence of why helm was reinstalled.

## Type safety

`HelmOptions.Version` changed from `string` to `*semver.Version`:
- `nil` clearly means "install latest"
- Non-nil means "install exactly this version"
- No redundant string/parse/string round-trips

## Test improvements

- Use `t.Cleanup` instead of `defer` for robustness with parallel subtests
- Extract `runnerForProfile` helper to reduce boilerplate
- Add `t.Log` progress lines for easier CI debugging and parallel output filtering
- Verify `HelmVersion` returns `ErrHelmNotInstalled` when helm is missing
- Rewrite comments to document real use cases instead of narrating code

## Example test run

The new test logs make it easy to filter test logs when running tests in parallel.

```console
% timestamp make integration TEST_ARGS="-minikube-start-args='--driver=vfkit' -test.run 'TestHelmInstall'" | grep -E '(TestHelmInstall|helm_test.go)'
[    0.000] make integration "TEST_ARGS=-minikube-start-args='--driver=vfkit' -test.run 'TestHelmInstall'"
[...]

PR #23364, opened 2026-07-15
Fix SSHRunner ExitCode always returning 0 for SSH commands

nirs:
## Summary

SSHRunner.ExitCode was always 0 because RunCmd and WaitCmd used
`*exec.ExitError` (os/exec) instead of `*ssh.ExitError` (crypto/ssh).
The type assertion silently failed, so every SSH command appeared to
succeed regardless of its actual exit code.

## Fix

- Use `*ssh.ExitError` and `ExitStatus()` instead of `*exec.ExitError`
  and `ExitCode()`
- Extract `sshExitCode` helper to avoid duplication and document the
  behavior for non-exit errors (e.g. connection drop, SSH protocol
  error): ExitCode remains 0 to match ExecRunner and KICRunner behavior

## SSH mock server rewrite

The previous mock server always returned exit code 0, had no stderr
support, accepted unauthenticated connections, and was never used by
any test.

Bug fixes:
- Use atomic.Bool for quit flag (thread-safe)
- Guard Start against double call (leaked listener and goroutine)

New features:
- Support configurable stdout, stderr, and exit code per command
- Unknown commands return exit code 127, simulating shell behavior
- Require public key authentication using ed25519 keys
- Provide Dial() method for simple test setup
- Make Stop idempotent (safe to call multiple times)

Refactoring:
- Simplify public interface: Start returns only error, port is internal
- Commands map cloned at construction, preventing data races
- Extract request handlers for clean error handling
- Use ListenTCP for type-safe listener setup
- No OS dependencies: works on any platform without a shell

## Tests

SSHRunner was never tested. Tests use the rewritten mock server to
verify that SSHRunner correctly reports stdout, stderr, exit code,
and errors for every combination:

| Test          | ExitCode | Stdout | Stderr | Error |
|---------------|----------|--------|--------|-------|
| ok            | 0        |        |        | no    |
| ok-out        | 0        | yes    |        | no    |
| ok-err        | 0        |        | yes    | no    |
| ok-out-err    | 0        | yes    | yes    | no    |
| fail          | 1        |        |        | yes   |
| fail-out-err  | 42       | yes    | yes    | yes   |
| no-such-cmd   | 127      |        | yes    | yes   |

### DIFF
--- a/pkg/minikube/command/ssh_runner.go
+++ b/pkg/minikube/command/ssh_runner.go
@@ -154,6 +154,16 @@ func (s *SSHRunner) Remove(f assets.CopyableFile) error {
 	return sess.Run(fmt.Sprintf("sudo rm %s", dst))
 }
 
+// sshExitCode extracts the exit code from an SSH error.
+// For non-exit errors (e.g. network failures), returns 0 to match
+// ExecRunner and KICRunner behavior.
+func sshExitCode(err error) int {
+	if exitError, ok := err.(*ssh.ExitError); ok {
+		return exitError.ExitStatus()
+	}
+	return 0
+}
+
 // teeSSH runs an SSH command, streaming stdout, stderr to logs
 func teeSSH(s *ssh.Session, cmd string, outB io.Writer, errB io.Writer) error {
 	outPipe, err := s.StdoutPipe()
@@ -223,9 +233,8 @@ func (s *SSHRunner) RunCmd(cmd *exec.Cmd) (*RunResult, error) {
 	err = teeSSH(sess, shellquote.Join(cmd.Args...), outb, errb)
 	elapsed := time.Since(start)
 
-	if exitError, ok := err.(*exec.ExitError); ok {
-		rr.ExitCode = exitError.ExitCode()
-	}
+	rr.ExitCode = sshExitCode(err)
+
 	// Decrease log spam
 	if elapsed > (1 * time.Second) {
 		klog.Infof("Completed: %s: (%s)", rr.Command(), elapsed)
@@ -315,9 +324,7 @@ func (s *SSHRunner) WaitCmd(sc *StartedCmd) (*RunResult, error) {
 	rr := sc.rr
 
 	err := s.s.Wait()
-	if exitError, ok := err.(*exec.ExitError); ok {
-		rr.ExitCode = exitError.ExitCode()
-	}
+	rr.ExitCode = sshExitCode(err)
 
 	sc.wg.Wait()
 
--- a/pkg/minikube/command/ssh_runner_test.go
+++ b/pkg/minikube/command/ssh_runner_test.go
@@ -19,9 +19,12 @@ package command
 import (
 	"bytes"
 	"fmt"
+	"os/exec"
 	"strings"
 	"sync"
 	"testing"
+
+	"k8s.io/minikube/pkg/minikube/tests"
 )
 
 func TestTeePrefix(t *testing.T) {
@@ -60,4 +63,62 @@ func TestTeePrefix(t *testing.T) {
 	if gotLog != wantLog {
 		t.Errorf("log=%q, want: %q", gotLog, wantLog)
 	}
+}
+
+func TestSSHRunner(t *testing.T) {
+	commands := map[string]tests.CommandResult{
+		"ok":           {},
+		"ok-out":       {Stdout: "out"},
+		"ok-err":       {Stderr: "err"},
+		"ok-out-err":   {Stdout: "out", Stderr: "err"},
+		"fail":         {ExitCode: 1},
+		"fail-out-err": {Stdout: "out", Stderr: "err", ExitCode: 42},
+	}
+
+	s, err := tests.NewSSHServer(t, commands)
+	if err != nil {
+		t.Fatalf("NewSSHServer: %v", err)
+	}
+
+	if err := s.Start(); err != nil {
+		t.Fatalf("Start: %v", err)
+	}
+	defer s.Stop()
+
+	client, err := s.Dial()
+	if err != nil {
+		t.Fatalf("Dial: %v", err)
+	}
+	defer client.Close()
+
+	runner := &SSHRunner{c: client}
+
+	for cmd, result := range commands {
+		t.Run(cmd, func(t *testing.T) {
+			rr, err := runner.RunCmd(exec.Command(cmd))
+			wantErr := result.ExitCode != 0
+			if (err != nil) != wantErr {
+				t.Errorf("err = %v, want error: %v", err, wantErr)
+			}
+			if rr.ExitCode != result.ExitCode {
+				t.Errorf("ExitCode = %d, want %d", rr.ExitCode, result.ExitCode)
+			}
+			if rr.Stdout.String() != result.Stdout {
+				t.Errorf("Stdout = %q, want %q", rr.Stdout.String(), result.Stdout)
+			}
+			if rr.Stderr.String() != result.Stderr {
+				t.Errorf("Stderr = %q, want %q", rr.Stderr.String(), result.Stderr)
+			}
+		})
+	}
+
+	t.Run("no-such-command", func(t *testing.T) {
+		rr, err := runner.RunCmd(exec.Command("no-such-command"))
+		if err == nil {
+			t.Errorf("err = nil, want error")
+		}
+		if rr.ExitCode != 127 {
+			t.Errorf("ExitCode = %d, want 127", rr.ExitCode)
+		}
+	})
 }
--- a/pkg/minikube/tests/ssh_mock.go
+++ b/pkg/minikube/tests/ssh_mock.go
@@ -18,222 +18,223 @@ package tests
 
 import (
 	"bytes"
+	"crypto/ed25519"
 	"crypto/rand"
-	"crypto/rsa"
+	"encoding/binary"
 	"errors"
 	"fmt"
-	"io"
+	"maps"
 	"net"
-	"strconv"
-	"sync"
 	"sync/atomic"
 	"testing"
+	"time"
 
 	"golang.org/x/crypto/ssh"
 )
 
-// SSHServer provides a mock SSH Server for testing. Commands are stored, not executed.
-type SSHServer struct {
-	Config *ssh.ServerConfig
-	// Commands stores the raw commands executed against the server.
-	Commands  map[string]int
-	Connected bool
-	Transfers *bytes.Buffer
-	// Only access this with atomic ops
-	hadASessionRequested int32
-	// commandToOutput can be used to mock what the SSHServer returns for a given command
-	// Only access this with atomic ops
-	commandToOutput atomic.Value
+// CommandResult defines the mock response for a command.
+type CommandResult struct {
+	Stdout   string
+	Stderr   string
+	ExitCode int
+}
 
-	quit     bool
+// SSHServer is a test SSH server returning preconfigured responses.
+// Runs one session at a time, matching minikube's SSHRunner.
+type SSHServer struct {
+	Config    *ssh.ServerConfig
+	ClientKey ssh.Signer
+
+	commands map[string]CommandResult
+	quit     atomic.Bool
+	port     int
 	listener net.Listener
 	t        *testing.T
 }
 
-// NewSSHServer returns a NewSSHServer instance, ready for use.
-func NewSSHServer(t *testing.T) (*SSHServer, error) {
+// NewSSHServer returns a new SSHServer instance with ed25519 key auth
+// and the given command responses.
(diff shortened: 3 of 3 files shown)


nirs:
@Roslaan001 want to review? With this we can have better logs in #23348.

Roslaan001:
> @Roslaan001 want to review? With this we can have better logs in #23348.

Yeah, I will check it out

pkg/minikube/tests/ssh_mock.go
```diff
@@ -151,20 +154,24 @@ func (s *SSHServer) handleRequest(channel ssh.Channel, req *ssh.Request, wg *syn
...
 		}
-		s.Commands[cmd.Command] = 1
 
-		s.t.Logf("returning output for %s ...", cmd.Command)
-		// Write specified command output as mocked ssh output
-		if val, err := s.GetCommandToOutput(cmd.Command); err == nil {
-			if _, err := channel.Write([]byte(val)); err != nil {
-				s.t.Errorf("Write failed: %v", err)
+		s.t.Logf("running command: %s", cmd.Command)
+		c := exec.Command("sh", "-c", cmd.Command)
```
Roslaan001: Since this mock SSH server runs during unit tests (which are executed on Windows hosts too), wouldn't hardcoding  `sh` `-c ` cause failures on Windows environments where a POSIX shell is not available in the PATH?                       

Should we just add a check/fallback for Windows, or does the minikube development/test environment assume a POSIX shell is always available?

pkg/minikube/tests/ssh_mock.go
```diff
@@ -151,20 +154,24 @@ func (s *SSHServer) handleRequest(channel ssh.Channel, req *ssh.Request, wg *syn
...
+		c.Stdout = channel
+		c.Stderr = channel.Stderr()
+		exitCode := 0
+		if err := c.Run(); err != nil {
+			if exitError, ok := err.(*exec.ExitError); ok {
+				exitCode = exitError.ExitCode()
 			}
 		}
```
Roslaan001: If  `c.Run()`  fails with a `non- ExitError ` (e.g., if the command binary itself is not found, or there are host resource issues),  `exitCode` will default to  `0`, signalling success to the caller.                                      
                                                                                                                                                                                                                                    
Can we  probably log the error and set  exitCode  to a non-zero value (like  255 ) to indicate execution failure:                                                                                                                    

```bash
  if err := c.Run(); err != nil {                                                                                                                                                                                                      
         if exitError, ok := err.(*exec.ExitError); ok {                                                                                                                                                                                     
             exitCode = exitError.ExitCode()                                                                                                                                                                                                     
         } else {                                                                                                                                                                                                                            
             s.t.Errorf("failed to run command: %v", err)                                                                                                                                                                                        
             exitCode = 255                                                                                                                                                                                                                      
         }                                                                                                                                                                                                                                   
     }

File: pkg/minikube/command/ssh_runner_test.go [hunk 3]
```diff
@@ -61,3 +65,85 @@ func TestTeePrefix(t *testing.T) {
...
+	port, err := s.Start()
+	if err != nil {
+		t.Fatalf("Start: %v", err)
+	}
+	defer s.Stop()
+
+	client, err := ssh.Dial("tcp", fmt.Sprintf("127.0.0.1:%d", port), &ssh.ClientConfig{
+		Auth:            []ssh.AuthMethod{ssh.PublicKeys(s.ClientKey)},
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+	})
```
Roslaan001: I think it's better to add a defensive connection timeout to the client configuration (e.g.,  Timeout: 5 * time.Second). What if the mock SSH server fails to start or accept connections for any reason? Won't `ssh.Dial` without a timeout hang indefinitely and block the entire test suite?

pkg/minikube/tests/ssh_mock.go
```diff
@@ -53,32 +49,48 @@ type SSHServer struct {
 // NewSSHServer returns a NewSSHServer instance, ready for use.
 func NewSSHServer(t *testing.T) (*SSHServer, error) {
 	t.Helper()
+
+	_, clientKey, err := ed25519.GenerateKey(rand.Reader)
+	if err != nil {
+		return nil, fmt.Errorf("Error generating client key: %w", err)
```
Roslaan001: According to the Go style guide (and standard lint tools), error messages should not start with capitalised letters or end with punctuation since they are usually wrapped.
                                                    
Maybe we should change  "Error generating..."  to  "error generating..." or "failed to generate..."

pkg/minikube/tests/ssh_mock.go
```diff
@@ -53,32 +49,48 @@ type SSHServer struct {
...
+		},
+		ClientKey: clientSigner,
 		t:         t,
 	}
 
-	private, err := rsa.GenerateKey(rand.Reader, 2014)
+	_, hostKey, err := ed25519.GenerateKey(rand.Reader)
 	if err != nil {
-		return nil, fmt.Errorf("Error generating RSA key: %w", err)
+		return nil, fmt.Errorf("Error generating host key: %w", err)
```
Roslaan001: According to the Go style guide (and standard lint tools), error messages should not start with capitalised letters or end with punctuation since they are usually wrapped.
                                                    
Maybe we should change  "Error generating..."  to  "error generating..." or "failed to generate..."

pkg/minikube/tests/ssh_mock.go
```diff
@@ -53,32 +49,48 @@ type SSHServer struct {
...
 	}
-	signer, err := ssh.NewSignerFromKey(private)
+	hostSigner, err := ssh.NewSignerFromKey(hostKey)
 	if err != nil {
 		return nil, fmt.Errorf("Error creating signer from key: %w", err)
```
Roslaan001: According to the Go style guide (and standard lint tools), error messages should not start with capitalised letters or end with punctuation since they are usually wrapped.
                                                    
Maybe we should change  "Error creating..."  to  "error creating..." or "failed to create..."

pkg/minikube/tests/ssh_mock.go
```diff
@@ -53,32 +49,48 @@ type SSHServer struct {
...
 func NewSSHServer(t *testing.T) (*SSHServer, error) {
 	t.Helper()
+
+	_, clientKey, err := ed25519.GenerateKey(rand.Reader)
+	if err != nil {
+		return nil, fmt.Errorf("Error generating client key: %w", err)
+	}
+	clientSigner, err := ssh.NewSignerFromKey(clientKey)
+	if err != nil {
+		return nil, fmt.Errorf("Error creating client signer: %w", err)
```
Roslaan001: According to the Go style guide (and standard lint tools), error messages should not start with capitalised letters or end with punctuation since they are usually wrapped.

Maybe we should change "Error creating..." to "error creating..." or "failed to create..."

pkg/minikube/tests/ssh_mock.go
Roslaan001: I think there is a data race here on  `s.quit`  between this read and the write in  `Stop() ` since they are executed in different goroutines without synchronisation.  Since  `sync/atomic`  is already imported in this file, why don't we use `s.quit`  to use atomic operations?

Roslaan001 (commented):
hi @nirs , I made some comments based on what I noticed, I would love you enlighten me if any is wrong or why it's not usable in this case. I believe it'll aid my understanding too

pkg/minikube/tests/ssh_mock.go
nirs: Yes this is an issue, the current tests will not run on Windows.

To keep this platform independent we have 2 ways:
- create separate test cases for windows and posix (ugly)
- change the mock to fake stdout, stderr, exit code

pkg/minikube/tests/ssh_mock.go
nirs: The ssh command always run a shell - if the command is not found, the shell returns  exit code 127.

We should always get ssh.ExitError - but if this assumption is wrong we need a special error code for internal error. Real exit code is 7 bits (0-127). 128-255 is used when the command was terminated by signal (e.g. kill -15 -> exit code 128+15).

Need to check how other tools running sub commands handle this and use a special exit code.

pkg/minikube/command/ssh_runner_test.go
nirs: If the server fail to start we will fail in line 74. If we reach this line, the server is listening on the port. But it is true that we need a timeout, so if the server has a bug, the test will fail quickly instead of getting stuck forever.

The best way would be to DialContext with the a context created from the test context with a short timeout.

pkg/minikube/tests/ssh_mock.go
nirs: Correct, need to use the right style. I think the linter should fail with this code.

pkg/minikube/tests/ssh_mock.go
nirs: No need to repeat the same comment, the author is responsible to address all the issues of the same type.

pkg/minikube/tests/ssh_mock.go
nirs: This is a bug in existing code - not in scope for this PR. Open an issue for this.

nirs:
@Roslaan001 Thanks for reviewing! this is very helpful review.

nirs:
@bobsira can you explain how the Windows unit tests pass, when this change added tests that can work only on posix? (e.g. running true, false, and sh -c "...").
https://github.com/kubernetes/minikube/actions/runs/29382613659/job/87249294306?pr=23364

pkg/minikube/tests/ssh_mock.go
nirs: Fixed in curent version

pkg/minikube/tests/ssh_mock.go
nirs: Fixed in current version

pkg/minikube/command/ssh_runner_test.go
nirs: Fixed in current version. Using timeout and not context to keep the code simpler. The actual test is very fast (10 milliseconds) so there no need to work hard on respecting the test context deadline.

pkg/minikube/tests/ssh_mock.go
nirs: We cannot do this since we need to match other runners behavior. We can open an issue to return invalid exit code when we don't get a complete info from the server. Using -1 seems to be a good value that can never be a real exit code (uint8).

pkg/minikube/tests/ssh_mock.go
nirs: Fixed by using a simpler mock that does not run commands, not it is portable and faster.

nirs:
> @bobsira can you explain how the Windows unit tests pass, when this change added tests that can work only on posix? (e.g. running true, false, and sh -c "..."). https://github.com/kubernetes/minikube/actions/runs/29382613659/job/87249294306?pr=23364

Cursor suggests that this works since git for windows is installed on the GitHub runners. It comes from sh, true, false and other tools, and installed in the PATH if user selected the option.

Current versions avoids this issue by using a portable mock that does not run anything.

nirs:
@Roslaan001 All comments are addressed, please check again.

pkg/minikube/command/ssh_runner.go
```diff
@@ -154,6 +154,16 @@ func (s *SSHRunner) Remove(f assets.CopyableFile) error {
...
 
+// sshExitCode extracts the exit code from an SSH error.
+// For non-exit errors (e.g. network failures), returns 0 to match
+// ExecRunner and KICRunner behavior.
+func sshExitCode(err error) int {
+	if exitError, ok := err.(*ssh.ExitError); ok {
+		return exitError.ExitStatus()
+	}
+	return 0
+}
```
Roslaan001, line 165: Is it not better to use `errors.As` instead of direct type assertion? Won't this type assertion fail if `err` is wrapped in the future?

Roslaan001 (commented):
@nirs, This is highly solid, no further comments from me

pkg/minikube/command/ssh_runner.go
nirs: The SSH library [documents](https://pkg.go.dev/golang.org/x/crypto/ssh#Session.Wait) that Wait() returns an error "of type *ExitError" -- this is a type identity contract, not a wrapping guarantee.

Using errors.As would be defensive against a hypothetical API-breaking change that would also break every other consumer of the library. The direct type assertion matches what the library promises and is consistent with how os/exec.ExitError is handled throughout Go codebases (including this one).

Roslaan001: Ohhh! Thank you so much for the clarification

PR #23368, opened 2026-07-16
Skip Docker client initialization for non-Docker drivers

### LINKED ISSUE #23349: `minikube cache` config path queries host Docker daemon, adding 10-30s per image on non-Docker drivers
## Summary

When images are registered via `minikube cache add`, they are persisted in
`~/.minikube/config/config.json` under the `cache` key. On every subsequent
`minikube start`, minikube reads this config and runs two phases per image:

1. **Save phase** — verify/download the image tarball to `~/.minikube/cache/images/`
2. **Load phase** — transfer the tarball to the guest and load it

Both phases attempt to query the **host's** Docker daemon for image digest
optimization. On non-Docker drivers (vfkit, hyperkit, QEMU, KVM), no host Docker
daemon exists. Each query blocks for a full 10-second timeout before falling back.

This adds **10-30 seconds per cached image** to every `minikube start`.

### With `cache` config entry (image file already exists on disk)

```
% minikube config view
- cache: map[nginx:latest:<nil>]

% timestamp minikube start -d vfkit --log_file start-after.log
[   14.560]   - Using image gcr.io/k8s-minikube/storage-provisioner:v5
[   14.928] * Enabled addons: default-storageclass, storage-provisioner
[   27.009] * Done!    ← 27s total
```

Log shows the save phase is instant (file exists), but the load phase hits the
10-second timeout:

```
cache.go:96]  cache image "nginx:latest" -> ".../nginx_latest" took 80.458µs
cache.go:80]  save to tar file nginx:latest -> .../nginx_latest succeeded

cache_images.go:90]  LoadCachedImages start: [nginx:latest]
cache_images.go:117] "nginx:latest" needs transfer: needs transfer timed out in 10.000000 seconds
image.go:80]          couldn't find image digest nginx:latest from local daemon:
                      Get "http://root%40docker.local/v1.51/images/nginx:latest/json": context deadline exceeded
cache_images.go:94]  duration metric: took 12.036926292s to LoadCachedImages
```

### With `cache` config entry (image file deleted from disk)

```
% rm ~/.minikube/cache/images/arm64/nginx_latest
% timestamp minikube start -d vfkit --log_file start-no-cached-image.log
[   13.273] * Enabled addons: default-storageclass, storage-provisioner
[   53.699] * Done!    ← 53s total
```

Now BOTH phases hit the Docker daemon timeout:

```
image.go:188] daemon lookup for nginx:latest: failed to connect to the docker API
              at ssh://root@docker.local: lookup root@docker.local: no such host
cache.go:96]  cache image "nginx:latest" -> ".../nginx_latest" took 28.570988208s  ← 25s daemon timeout + download

cache_images.go:90]  LoadCachedImages start: [nginx:latest]
cache_images.go:117] "nginx:latest" needs transfer: needs transfer timed out in 10.000000 seconds
image.go:80]          couldn't find image digest nginx:latest from local daemon:
                      Get "http://root%40docker.local/v1.51/images/nginx:latest/json": context deadline exceeded
cache_images.go:94]  duration metric: took 11.996759708s to LoadCachedImages
```

### Without `cache` config entry (image file still on disk)

```
% minikube cache delete nginx:latest
% timestamp minikube start -d vfkit --log_file start-no-cached-image.log
[   14.609] * Enabled addons: default-storageclass, storage-provisioner
[   15.092] * Done!    ← 15s total
```

No `LoadCachedImages` in the log at all. The image file in
`~/.minikube/cache/images/` is inert without the config entry.

## Root Cause

`minikube cache add` persists images to config:

```go
// cmd/minikube/cmd/cache.go
if err := cmdConfig.AddToConfigMap(cacheImageConfigKey, args); err != nil { ... }
```

On every `minikube start`, `CacheAndLoadImagesInConfig` reads this config and calls
`machine.CacheAndLoadImages`, which:

1. Calls `image.SaveToDir` → `saveToTarFile` → tries the Docker daemon to get the
   image (via `go-containerregistry`'s daemon client which resolves `DOCKER_HOST`)
2. Calls `LoadCachedImages` → creates a Docker `client.Client` via
   `client.FromEnv` → calls `DigestByDockerLib` with a 10-second timeout

In `pkg/minikube/machine/cache_images.go`:

```go
var imgClient *client.Client
if cr.Name() == "Docker" {
    imgClient, err = client.NewClientWithOpts(client.FromEnv)
}
```

The condition `cr.Name() == "Docker"` checks the **guest** container runtime, not
whether a Docker daemon is available on the **host**. For VM-based drivers (vfkit,
hyperkit, QEMU, KVM), the host Docker daemon is irrelevant — images must be
transferred via SCP regardless.

In `pkg/minikube/image/image.go`:

```go
func DigestByDockerLib(imgClient *client.Client, imgName string) string {
    ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
    defer cancel()
    img, err := imgClient.ImageInspect(ctx, imgName)
    // ... blocks for 10s if daemon is unreachable
}
```

## Impact

- **+10s per cached image** (load phase) on every `minikube start` for VM drivers
- **+25-30s per cached image** (save + load) if the tarball file was deleted
- Multiplies with number of cached images
- Silent — no user-visible error or warning, just slow startup
- Users who followed `minikube cache add` docs are penalized on every start

## Proposed Fix

The Docker daemon optimization is only valid when minikube uses the **Docker
driver** (KIC mode), where the host daemon IS the container runtime.

```go
if cc.Driver == driver.Docker {
    imgClient, err = client.NewClientWithOpts(client.FromEnv)
}
```

For the save phase, skip the daemon source when the driver is not Docker.

Minikube has no dependency on the host Docker daemon unless using the Docker
driver. Any code path that attempts to access it for other drivers is broken by
design. On macOS vmnet networks, connecting to a non-existing Docker VM blocks
indefinitely instead of failing with "connection refused" (as on Linux), making
timeout-based workarounds impractical.

## Environment

- macOS 26.5.2 (arm64)
- minikube v1.38.1
- Driver: vfkit
- Container runtime in guest: docker
- No Docker Desktop on host

## Logs

- [start-with-cached-image.log](https://github.com/user-attachments/files/29942517/start-with-cached-image.log)
- [start-no-cached-image.log](https://github.com/user-attachments/files/29942521/start-no-cached-image.log)

OchukoWH:
## Summary

`minikube start` could initialize a client for the host Docker daemon whenever the guest container runtime was Docker, even when Minikube was using a non-Docker driver such as `vfkit`.

For VM drivers, access to the host Docker daemon is unnecessary. When Docker Desktop is stopped or unavailable, attempting to initialize the Docker client can introduce unnecessary connection attempts and startup delays.

This PR guards Docker client initialization so it only occurs when Minikube is actually running with a Docker-based driver.

---

## Root Cause

The Docker client initialization was previously guarded only by the guest container runtime:

```go
if cr.Name() == "Docker" {
	imgClient, err = client.NewClientWithOpts(client.FromEnv)
	if err != nil {
		klog.Infof("couldn't get a local image daemon which might be ok: %v", err)
	}
}
```

This correctly checks whether Docker is the **guest container runtime**, but it does **not** verify that the selected Minikube driver is Docker.

As a result, VM drivers such as `vfkit` could still attempt to initialize a Docker client for the host even though the host Docker daemon is not required.

Example:

```
Driver: vfkit
Guest container runtime: Docker
```

In this configuration, Docker runs **inside the VM**, not on the host, so attempting to initialize a host Docker client is unnecessary.

---

## What this PR changes

Docker client initialization is now guarded by both:

- the selected Minikube driver
- the guest container runtime

```go
if driver.IsDocker(cc.Driver) && cr.Name() == "Docker" {
	imgClient, err = client.NewClientWithOpts(client.FromEnv)
	if err != nil {
		klog.Infof("couldn't get a local image daemon which might be ok: %v", err)
	}
}
```

The Docker client is now initialized **only when**:

- the selected driver is Docker, and
- the guest container runtime is Docker.

For non-Docker drivers (`vfkit`, `hyperkit`, `qemu`, `kvm`, etc.), the Docker client initialization is skipped and the existing image-loading path continues unchanged.

---

## Scope

This PR only changes the condition controlling Docker client initialization during `minikube start`.

---

### Environment

```
OS: macOS 26.5.1
Architecture: amd64
Minikube: v1.38.1 (development build)
Driver: vfkit
Guest Runtime: Docker 28.5.2
Kubernetes: v1.36.2
```

---

#### 1. Non-Docker driver

Docker Desktop stopped.

```bash
./out/minikube start \
  --profile=minikube-dev-tester \
  -d vfkit
```

Result:

- Cluster starts successfully
- No dependency on the host Docker daemon
- Existing image-loading behavior preserved

---

#### 2. Docker driver

Docker Desktop running.

```bash
./out/minikube start \
  --profile=minikube-dev-docker \
  -d docker
```

Result:

- Cluster starts successfully
- Existing Docker-driver behavior unchanged

---

## Startup timing

Startup timing was measured using both the previous build and the modified build. I noticed the new build starts up the cluster about 10 - 15 seconds earlier than the older the build.

Startup timing varied between runs because of factors outside this change, including:

- VM provisioning
- DHCP allocation
- Registry connectivity
- Host machine load

The purpose of this change is **not** to optimize startup time directly, but to eliminate unnecessary host Docker daemon initialization for non-Docker drivers.

---

### Before this change

Stop Docker Desktop.

```bash
time ./out/minikube start \
  --profile=minikube-dev-tester \
  -d vfkit \
  --container-runtime=docker \
  -v=9
```

Observe that Minikube initializes a Docker client even though the selected driver is not Docker.

---

### After this change

Stop Docker Desktop.

```bash
time ./out/minikube start \
  --profile=minikube-dev-tester \
  -d vfkit \
  --container-runtime=docker \
  -v=9
```

The Docker client initialization path is skipped for non-Docker drivers while preserving existing startup behavior.

---

Fixes #23349

### DIFF
--- a/pkg/minikube/machine/cache_images.go
+++ b/pkg/minikube/machine/cache_images.go
@@ -44,6 +44,7 @@ import (
 	"k8s.io/minikube/pkg/minikube/config"
 	"k8s.io/minikube/pkg/minikube/cruntime"
 	"k8s.io/minikube/pkg/minikube/detect"
+	"k8s.io/minikube/pkg/minikube/driver"
 	"k8s.io/minikube/pkg/minikube/image"
 	"k8s.io/minikube/pkg/minikube/localpath"
 	"k8s.io/minikube/pkg/minikube/out"
@@ -97,7 +98,7 @@ func LoadCachedImages(cc *config.ClusterConfig, runner command.Runner, imgs []st
 	var g errgroup.Group
 
 	var imgClient *client.Client
-	if cr.Name() == "Docker" {
+	if driver.IsDocker(cc.Driver) && cr.Name() == "Docker" {
 		imgClient, err = client.NewClientWithOpts(client.FromEnv) // image client
 		if err != nil {
 			klog.Infof("couldn't get a local image daemon which might be ok: %v", err)


OchukoWH:
Hi @nirs, just a friendly ping on this PR whenever you have some time.

The implementation is ready, and from what I can tell the remaining failing checks appear to be unrelated CI/infrastructure issues. Happy to make any additional changes if needed. Thanks!

nirs:
> ### Functional verification
> #### 1. Non-Docker driver
> Docker Desktop stopped.
> 
> ```shell
> ./out/minikube start \
>   --profile=minikube-dev-tester \
>   -d vfkit
> ```
> 
> Result:
> 
> * Cluster starts successfully
> * No dependency on the host Docker daemon
> * Existing image-loading behavior preserved

This worked before this change. To compare the results you need to run before and after the change with time, and share the complete output.

> #### 2. Docker driver
> Docker Desktop running.
> 
> ```shell
> ./out/minikube start \
>   --profile=minikube-dev-docker \
>   -d docker
> ```
> 
> Result:
> 
> * Cluster starts successfully
> * Existing Docker-driver behavior unchanged

This looks good to show that we don't have regressions. But it will be more helpful to measure the time to see if this change has any bad effect.

> ## Startup timing
> Startup timing was measured using both the previous build and the modified build. I noticed the new build starts up the cluster about 10 - 15 seconds earlier than the older the build.
> 
> Startup timing varied between runs because of factors outside this change, including:
> 
> * VM provisioning
> * DHCP allocation
> * Registry connectivity
> * Host machine load
> 
> The purpose of this change is **not** to optimize startup time directly, but to eliminate unnecessary host Docker daemon initialization for non-Docker drivers.

You did not share any results. In my tests start time is consistent.

To test this you want to restart, don't run anything else, and use hyperfine to do at leat 10 runs.

```
hyperfine --warmup 1 --conclude "out/minikube delete -p test" "out/minikube start -p test -d vfkit --log_file=start.log"
```

Note also that the issue is about delay after adding an image using a cache command. Please repeat the step in #23349 to reproduce the 15 seconds delay when starting a cluster.

pkg/minikube/machine/cache_images.go
```diff
@@ -97,7 +98,7 @@ func LoadCachedImages(cc *config.ClusterConfig, runner command.Runner, imgs []st
 	var g errgroup.Group
 
 	var imgClient *client.Client
-	if cr.Name() == "Docker" {
+	if driver.IsDocker(cc.Driver) && cr.Name() == "Docker" {
```
nirs, line 101: This looks correct - but is this the only place we need to change this?

We need to search the code checking `cr.Name() = "Docker"` and evaluate if they they are missing docker driver check.

nirs:
@OchukoWH please check the comments.

OchukoWH:
> @OchukoWH please check the comments.

This skipped my mind
Please I will work on this tomorrow

nirs:
@OchukoWH any update?

OchukoWH:
> @OchukoWH any update?

Hi @nirs 
I actually ran the functional verification tests but I wasn't able to observe consistent startup times, there was a fluctuation in the startup times. This may be due to my PC being a little bit noisy and a little much older.

pkg/minikube/machine/cache_images.go
OchukoWH: > This looks correct - but is this the only place we need to change this?
> 
> We need to search the code checking `cr.Name() = "Docker"` and evaluate if they they are missing docker driver check.

about this, I checked other places where `cr.Name() = "Docker"` is used  and they don't directly affect startup

